PR #3397, opened 2014-12-06
(maint) Remove 'create' key for masterhttplog setting

kylog:
This is the only setting that uses this key, and yet the
webrick setup_logger step already creates the file if needed.
So removing this will allow for some code cleanup in
FileSetting and derived classes.


joshcooper:
ping @kylog

kylog:
Thanks for the reminder. I created https://tickets.puppetlabs.com/browse/PUP-3817. Not sure if I'm going to circle back to this one super soon (although it should be simple), so I'll just close this PR for now.

PR #3398, opened 2014-12-06
(PUP-3274) Remove obsolete parser setting

kylog:


joshcooper:
@kylog I'm following up on this

PR #3399, opened 2014-12-06
(PUP-1085) Fix handling of package groups in pacman provider.

peterhuene:
These commits fix the pacman package provider's handling of groups.

Previously, the pacman provider would return a version of '1' for package groups if a single package of the group was installed. This meant that groups that were partially installed were considered to be installed.  Additionally, if a package in a group was updated, the group was still considered to be the latest and no action would be taken to update the group.

This fixes the above issue by considering a package group to be installed only if all the packages in the group are installed.  Additionally, package group versions are now a concatenation of all the versions of the packages comprising the groups.  This allows the group to install packages in the group that have been updated.

Also fixed:

* Pass the `--needed` option to pacman to prevent needless reinstallation of packages.
* Return :absent instead of :purged if a package is missing
* Implement the virtual_packages feature to disable handling package groups.
* Improve errors and warnings in the provider.

### DIFF
--- a/lib/puppet/provider/package/pacman.rb
+++ b/lib/puppet/provider/package/pacman.rb
@@ -55,42 +55,22 @@ Puppet::Type.type(:package).provide :pacman, :parent => Puppet::Provider::Packag
   # If a group removes packages over time, nothing will happen. This is intended.
   def self.instances
     instances = []
+
+    # Get the installed packages
     installed_packages = get_installed_packages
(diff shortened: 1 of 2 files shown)


kylog:
This needs a rebase now, possibly due to the rspec 3 support, or perhaps other changes.

peterhuene:
@kylog I've rebased.  The conflicts came from the rspec 3 changes.

ScottGarman:
Triage notes: we do intend to merge this after the Puppet 4.0.0 merge freeze.

PR #3401, opened 2014-12-06
(PUP-3750) Support :include: in mailalias type

dspinellis:
The mailalias type does not allow the specification of included files,
which contain an alias's elements. Such alias entries are commonly
specified in the an MTA's aliases file through the following syntax.

aliasname: :include: /path/to/file

Specifying the rest of an ":include:" line as a recipient does not
work, because the existing mailalias type will quote the entry.

This change adds a new parameter named "file" to the mailalias type,
which allows the indirect specification of a file containing the
recipients.  The parameter parses and generates ":include:" alias entries.

### DIFF
--- a/lib/puppet/type/mailalias.rb
+++ b/lib/puppet/type/mailalias.rb
@@ -33,6 +33,12 @@ module Puppet
     newproperty(:file) do
       desc "A file containing the alias's contents.  The file and the
         recipient entries are mutually exclusive."
+
+      validate do |value|
+	unless Puppet::Util.absolute_path?(value)
+	  fail Puppet::Error, "File paths must be fully qualified, not '#{value}'"
+	end
+      end
     end
 
     newproperty(:target) do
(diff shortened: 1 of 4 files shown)


lib/puppet/type/mailalias.rb
```diff
@@ -33,6 +30,11 @@ def should_to_s(value)
       end
     end
 
+    newproperty(:file) do
```
MikaelSmith: Was that a conclusion that we can leave it as-is for this PR and someone (probably me) can try switching them in the instances @adrienthebo identified, or that the PR should be changed to use `newparam`?

adrienthebo: @MikaelSmith my inclination is to be consistent and use `newproperty` for now and get the PR merged; we can try changing this value to a parameter in another pass.

MikaelSmith: Then this looks ready.

MikaelSmith:
Can you squash commits to remove the intermediate work?

dspinellis:
OK. To avoid rewriting history on a publicly visible repo, I have created [a new pull request](https://github.com/puppetlabs/puppet/pull/3447).

MikaelSmith:
Merged the new request, thanks.

PR #3402, opened 2014-12-08
(PUP-3313) Remove marker method puppet_lambda

hlindberg:
Now when the 3x AST implementation has been removed the marker
method :puppet_lambda is no longer needed. This commit replaces
the check if an argument is callable by using the type calculator
instead of checking the marker method.

The marker method was only used by the lookup function. All other such
calls have already been cleaned up.


zaphod42:
:+1:

PR #3428, opened 2014-12-19
(PUP-3656) Prefetch only rescue LoadError or MissingCommand

nanliu:
Currently prefetch rescue all StandardError, and this results in puppet
creating resources unecessarily when prefetch fails. This patch limits
the exception handling to only rescue LoadError for rubygems, and
Puppet::MissingCommand when software packages have not been installed
yet.

### DIFF
--- a/lib/puppet/error.rb
+++ b/lib/puppet/error.rb
@@ -59,4 +59,8 @@ module Puppet
   class DevError < Puppet::Error
     include ExternalFileError
   end
+
+  class MissingCommand < Puppet::Error
+  end
+
 end
--- a/lib/puppet/provider/command.rb
+++ b/lib/puppet/provider/command.rb
@@ -19,7 +19,7 @@ class Puppet::Provider::Command
   # @param args [Array<String>] Any command line arguments to pass to the executable
   # @return The output from the command
   def execute(*args)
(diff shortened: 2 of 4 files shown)


joshcooper:
Not sure... it has the potential to break a lot of modules, and a lot of modules on the forge define a prefetch method. On the other hand, the "break" might be welcome, because previously exceptions were being silently caught, and puppet acted like everything was ok. /cc @hunner

joshcooper:
@nanliu @kylog I agree that the change is the right thing. I'm not 100% convinced that most providers will work without modification simply because this behavior was been this way for so long. But it is puppet 4, so :+1:

kylog:
My thoughts exactly :)

kylog:
@nanliu and/or @joshcooper take a look at (and please improve) the verbiage I added to the description in https://tickets.puppetlabs.com/browse/PUP-3656, as that will be input to docs for the puppet 4 documentation.

nanliu:
@kylog, the documentation needs some tweaking. In Puppet 4, the internal transaction system will rescue LoadError and Puppet::MissingCommand. The module author can raise Puppet::Error so resources will fail correctly, and they can throw other exceptions but they are responsible for handling them appropriately.

```ruby
def prefetch
   data = api_call
   raise(Puppet::Error, 'Unable to prefetch resource') unless data # this should work now
   ...
rescue API::Error
   ...
end
```

PR #3433, opened 2014-12-22
(maint) Change spec test to pass on ruby 2.2

kylog:
Prior to this change, this spec test failed on ruby 2.2.0-rc1 with:

     Failure/Error: Time.stubs(:now).returns(Time.parse("2011-01-06 12:00:00 UTC"))
     NoMethodError:
       undefined method `utc_offset' for nil:NilClass

I didn't track down the root cause of this failure, but just
normalized this stub to use Time.utc, which (along with Time.local)
is the more common way of stubbing Time objects in the codebase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ script:
 notifications:
   email: false
 rvm:
+  - 2.2.0
   - 2.1.5
   - 2.0.0
   - 1.9.3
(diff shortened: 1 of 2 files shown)


joshcooper:
I verified the important thing is that `Time.parse` works:

<pre>
$ ruby --version
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-darwin12.0]
$ ruby -rtime -e 'puts Time.parse("2011-01-06 12:00:00 UTC")'
2011-01-06 12:00:00 UTC
</pre>

just not in tests for some reason...

PR #3434, opened 2014-12-22
(PUP-3719) Group resource non-authoritative by default

ferventcoder:
The `auth_membership` group parameter controls whether puppet ensures the
group contains _exactly_ the members specified, and no more, or contains _at
least_ the members specified. By default, the group parameter defaults to the
former, which is opposite to the `auth_membership` user parameter. This makes
the group parameter difficult to use in practice, because you need to know what
the complete set of group members should be. For example, on Windows the local
`Administrators` group may contain a combination of local and domain
user/group accounts, and that may vary across different types of machines.

This switches the default to false to correct the confusion and needing to 
know/specify every member of a group.

### DIFF
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -329,7 +329,7 @@ module Puppet::Util::Windows::ADSI
     end
 
     def self.name_sid_hash(names)
-      return [] if names.nil? or names.empty?
+      return {} if names.nil? or names.empty?
 
       sids = names.map do |name|
         sid = Puppet::Util::Windows::SID.name_to_sid_object(name)
--- a/spec/unit/provider/group/windows_adsi_spec.rb
+++ b/spec/unit/provider/group/windows_adsi_spec.rb
@@ -51,12 +51,6 @@ describe Puppet::Type.type(:group).provider(:windows_adsi), :if => Puppet.featur
         provider.members_insync?(['user1'], nil).should be_false
       end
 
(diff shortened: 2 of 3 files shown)


spec/unit/provider/group/windows_adsi_spec.rb
```diff
@@ -151,7 +173,8 @@
 
       create = sequence('create')
       group.expects(:commit).in_sequence(create)
-      group.expects(:set_members).with(['user1', 'user2'], true).in_sequence(create)
+      # it seems a bit strange that defaultto false returns nil
+      group.expects(:set_members).with(['user1', 'user2'], nil).in_sequence(create)
```
joshcooper: It appears puppet does not handle default false values correctly due to a long standing bug PUP-1967. In https://github.com/puppetlabs/puppet/blob/3.7.3/lib/puppet/type.rb#L813-L817, if the default value is `false` then it's not truthy, so `value = parameter.default` is false, preventing the default `parameter.value` from being initialized correctly.

lib/puppet/util/windows/adsi.rb
```diff
@@ -329,7 +329,7 @@ def commit
     end
 
     def self.name_sid_hash(names)
-      return [] if names.nil? or names.empty?
+      return {} if names.nil? or names.empty?
```
joshcooper: nice catch, I wonder how that ever worked?

ferventcoder: Not sure it did... :+1:

PR #3435, opened 2014-12-23
(do not merge) (PUP-3011) johnduarte pmt install acceptance test changes for PE 3.7

er0ck:
this should supersede and close https://github.com/puppetlabs/puppet/pull/3047


er0ck:
don't merge this.  investigating moving much of this to spec tests.

PR #3439, opened 2014-12-23
(PUP-1343) Use chkconfig's --add and --del options

ScottGarman:
These options ensure that the appropriate S\* and K\* symlinks are
created and removed when a service is enabled or disabled.


mckern:
:+1:

ahpook:
LGTM, this looks like it does not cause the problems we saw in https://projects.puppetlabs.com/issues/15797  :+1:

PR #3441, opened 2014-12-25
(PUP-3793) Check for agent's process owner instead of master's one for user property validation in exec type

mcanevet:
Should fix that kind of stuffs : https://github.com/puppetlabs/puppetlabs-postgresql/pull/394

### DIFF
--- a/lib/puppet/type/exec.rb
+++ b/lib/puppet/type/exec.rb
@@ -204,7 +204,7 @@ module Puppet
       validate do |user|
         if Puppet.features.microsoft_windows?
           self.fail "Unable to execute commands as other users on Windows"
-        elsif !Puppet.features.root? && resource.current_username() != user
+        elsif Facter.value(:id) != 'root' && Facter.value(:id) != user
           self.fail "Only root can execute commands as other users"
         end
       end
@@ -583,10 +583,6 @@ module Puppet
           self.property(:returns).sync
         end
       end
-    end
-
-    def current_username
-      Etc.getpwuid(Process.uid).name
     end
   end
 end
--- a/spec/unit/type/exec_spec.rb
+++ b/spec/unit/type/exec_spec.rb
@@ -198,15 +198,14 @@ describe Puppet::Type.type(:exec) do
   describe "when setting user" do
     describe "on POSIX systems", :as_platform => :posix do
       it "should fail if we are not root" do
-        Puppet.features.stubs(:root?).returns(false)
+        Facter.fact(:id).stubs(:value).returns('foo')
         expect {
           Puppet::Type.type(:exec).new(:name => '/bin/true whatever', :user => 'input')
         }.to raise_error Puppet::Error, /Parameter user failed/
       end
 
       it "accepts the current user" do
-        Puppet.features.stubs(:root?).returns(false)
-        Etc.stubs(:getpwuid).returns(Struct::Passwd.new('input'))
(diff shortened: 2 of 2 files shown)


mcanevet:
@kylog @MikaelSmith after much thought I think the problem comes from the fact that we are currently testing upon the master's process owner instead of the agent one.
What do you think of this new commit?

mcanevet:
How can I stub `id` fact in unit tests so that it passes?

MikaelSmith:
I'm still evaluating the impact of this.

You can stub the `id` fact with something like
`@Facter.stubs(:value).with(:id).returns("a user")`

lib/puppet/type/exec.rb
```diff
@@ -204,7 +204,7 @@ def value=(*values)
       validate do |user|
         if Puppet.features.microsoft_windows?
           self.fail "Unable to execute commands as other users on Windows"
-        elsif !Puppet.features.root? && resource.current_username() != user
+        elsif Facter.value(:id) != 'root' && Facter.value(:id) != user
```
joshcooper: @mcanevet I'm not understanding why this is necessary. The validate block is executed on the agent when it converts the resource catalog (transmitted via REST) to the RAL catalog. Here is a trace leading up to this:

<pre>
/Users/josh/work/puppet/lib/puppet/provider/exec.rb:89:in `validatecmd'
/Users/josh/work/puppet/lib/puppet/type/exec.rb:500:in `block (2 levels) in <module:Puppet>'
/Users/josh/work/puppet/lib/puppet/type.rb:2311:in `initialize'
/Users/josh/work/puppet/lib/puppet/resource.rb:377:in `new'
/Users/josh/work/puppet/lib/puppet/resource.rb:377:in `to_ral'
/Users/josh/work/puppet/lib/puppet/resource/catalog.rb:501:in `block in to_catalog'
/Users/josh/work/puppet/lib/puppet/resource/catalog.rb:493:in `each'
/Users/josh/work/puppet/lib/puppet/resource/catalog.rb:493:in `to_catalog'
/Users/josh/work/puppet/lib/puppet/resource/catalog.rb:396:in `to_ral'
/Users/josh/work/puppet/lib/puppet/configurer.rb:76:in `convert_catalog'
</pre>

I suspect that you are seeing an issue because your test is both the agent and the master, and is not stubbing `Puppet.features.stubs(:root?).returns(true)`.

mcanevet: @joshcooper to summarize, rspec-puppet's `it { should compile }` does not work as regular user when using `exec { 'foo': user => 'bar' }` in your manifests without using fakeroot.

joshcooper: @mcanevet so `it { should compile }` actually does two things, it compiles the "resource" catalog (which normally happens on the master) and it converts it to a "ral" catalog (which normally happens on the agent). In your case, both steps are happening together, and the same is true when doing `puppet apply`.

The second step is what triggers the validate method in the `user` parameter of the `exec` type. Since you are not running the test as root, the test fails as expected.

I believe the fix should be done in your test, e.g.

<pre>
before :each do
  Puppet.features.stubs(:root?).returns(true)
end

it { should compile }
</pre>

mcanevet:
@joshcooper OK that works fine. I don't need this PR anymore.
Maybe we should put this in puppetlabs_spec_helper

PR #3442, opened 2014-12-30
Port puppet specs to rspec 3

kylog:
This is WIP so please don't merge.

At this point specs look good on my laptop (mavericks) and travis (ubuntu 12.04), but windows has 334 spec fails. I'm taking a look at the windows spec failures (which look largely mechanical), but meanwhile, if anyone wants to review the changes so far, go for it!


kylog:
@MikaelSmith or @joshcooper or @hlindberg can any of you comment on this, not so much from the perspective of the individual commits (which may be dauntingly vast from a line count perspective) but from the "do we even want this" perspective?

I ask because it's proving to be a fair bit of work to keep rebasing it against other incoming spec changes (and I still haven't arrived at a 100% reliable windows spec environment - next stop will be to use vmpooler).

So if the collective feeling is 'meh', I'll just drop this PR (but keep the branch around if (my/our) interest ever revives). 

Thanks for any comments!

hlindberg:
Rspec 3 has composable matchers which is something that will make it possible to write better spec tests. There are several such cases in the puppet language part of our test suite. While you can typically make it work with other mechanisms it is hard to provide meaningful failure messages. (We even have a ticket to improve one such implementation). So yeah, it is worth it.

joshcooper:
@kylog overall I'm +1 with the direction. As to be expected, I'm getting a few windows failures: https://gist.github.com/joshcooper/01981d9ec46517db82f6 I can help wrap those up tomorrow.

kylog:
@joshcooper cool, I've been pushing up some windows fixes as I go this afternoon and might have some of that gist fixed. But not all. It is *really* close though, so I'd very much appreciate the help tomorrow.

One puzzler though: I don't have the file `spec/ext/windows/service/daemon_spec.rb` in my tree at all.

kylog:
@joshcooper or @hlindberg I've removed the WIP prefix on the PR because there are no remaining spec failures on travis, my mavericks laptop, or my windows 2012 vm (except for a handful of specs I see on puppet/master as well - artifacts of my dev environment that I won't bore you with at the moment ... ;)).

spec/integration/util/feature_spec.rb
```diff
@@ -1,53 +0,0 @@
...
-    end
-
-    yep = File.join(@path, "yep.rb")
-    File.open(yep, "w") { |f|
-      f.puts "$features.add(:yep, :libs => %w{puppet})"
-    }
-
-    assert(@features.yep?, "false 'yep' is apparently cached or feature could not be loaded")
-  end
-end
```
joshcooper: Was this test replaced with something else?

spec/monkey_patches/alias_should_to_must.rb
```diff
@@ -1,21 +0,0 @@
...
-require 'puppet/type'
-class Puppet::Type
-  alias :should_native :should
-  def should(value)
-    unless value.is_a? String or value.is_a? Symbol
-      raise "you need to call .must rather than .should on Puppet::Type instances"
-    end
-    should_native(value)
-  end
-end
```
joshcooper: yay!

spec/unit/daemon_spec.rb
```diff
@@ -27,7 +27,7 @@ def run_loop(jobs)
     end
   end
 
-  let(:server) { stub("Server", :start => nil, :wait_for_shutdown => nil) }
+  let(:server) { double("Server", :start => nil, :wait_for_shutdown => nil) }
```
joshcooper: is `double` ok here?

spec/integration/util/feature_spec.rb
kylog: No, it wasn't. Worth discussing.

There are two things in this file:
* the `def test_dynamic_loading` method which is just orphaned (or never implemented) code AFAICT
* the `"should be able to load features from disk"` test, which is flat-out removed (see https://github.com/kylog/puppet/commit/2d722b7e62534a01fd71474cdfe37c862a5f056a for my reasoning at the time)

I'm not wedded to removing it, but I wasn't sure at the time (or now tbh) how to test this with rspec 3, and it seemed well tested by virtue of use of features. But could be worth revisiting now?

spec/unit/daemon_spec.rb
kylog: Hmm, good catch. I need to look at that.

spec/integration/util/feature_spec.rb
joshcooper: Makes sense, I think we can kill it. We know other features can load from disk, especially all of the windows gems that we load.

joshcooper:
PR passes for me also on 2008r2.

spec/unit/daemon_spec.rb
kylog: It was dead code. There was one other dead stub (transpec'd to double) as well. I just added a commit removing those stubs/doubles.

kylog:
> PR passes for me also on 2008r2.

Woot!

kylog:
@joshcooper and @hlindberg: any additional thoughts on this? I know Josh got a chance to review part of it.

And two logistics questions:

1) Should I file a ticket for this? Currently it's all "(maint)" which is often what we use where there's no functional change, as here. But worth checking.

2) Should I squash? As-is, some of the intermediate commits will fail spec runs, so would be a black hole for a git bisect. I usually don't like this, but in this case, the individual commits may have some value where they were shedding light on particular fixups needed. And if all squashed, the more interesting bits, will be lost in the sea of mechanical changes. I could go either way on this.

hlindberg:
1. No ticket really needed unless there is documentation / discussion we want to have with puppet developers. (And something for people to click on in the release notes).  (I am ambivalent).

2. Ambivalent here as well - choose your poison.

kylog:
Rebased.

joshcooper:
I'm thinking we should email puppet-dev about the move to rspec3 and what this will mean for submitted PRs, e.g. using `expect(..).to eq(42)`, etc. Having a ticket to point people to might be helpful, but I'm fine either way...

PR #3443, opened 2014-12-31
(PUP-3801) Fixed issue with HP-UX Trusted Computing where password

jjperry69:
expirations were not being updated.

When I modified the HP-UX Password Provider in 2013, the issue was not
found until months later when the accounts used for testing exceeded the
system defined 90 day password expiration. The password hash for the new
password was getting pushed. When logging in with the new password,
HP-UX Trusted Computing was showing a successful password change, but it
was not resetting the password age back to 0 days. 

This fix corrects that behavior.

### DIFF
--- a/lib/puppet/provider/user/hpux.rb
+++ b/lib/puppet/provider/user/hpux.rb
@@ -2,17 +2,20 @@ Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
   desc "User management for HP-UX. This provider uses the undocumented `-F`
     switch to HP-UX's special `usermod` binary to work around the fact that
     its standard `usermod` cannot make changes while the user is logged in.
+<<<<<<< HEAD
+    New functionality provides for changing trusted computing passwords and
+=======
     New functionality provides for changing trusted computing passwords and 
+>>>>>>> refs/remotes/import/fix/master/PUP-3801-HP-UX-User-Provider
     resetting password expirations under trusted computing"
 
   defaultfor :operatingsystem => "hp-ux"
   confine :operatingsystem => "hp-ux"
 
-  commands :modify => "/usr/sam/lbin/usermod.sam", :delete => "/usr/sam/lbin/userdel.sam", :add => "/usr/sam/lbin/useradd.sam", :password => '/usr/lbin/modprpw', :expiry => '/usr/lbin/modprpw'
+  commands :modify => "/usr/sam/lbin/usermod.sam", :delete => "/usr/sam/lbin/userdel.sam", :add => "/usr/sam/lbin/useradd.sam"
   options :comment, :method => :gecos
   options :groups, :flag => "-G"
   options :home, :flag => "-d", :method => :dir
-  options :expiry, :method => :expire
 
   verify :gid, "GID must be an integer" do |value|
     value.is_a? Integer
@@ -22,7 +25,8 @@ Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
     value !~ /\s/
   end
 
-  has_features :manages_homedir, :allows_duplicates, :manages_passwords, :manages_password_age, :manages_expiry
+
+  has_features :manages_homedir, :allows_duplicates, :manages_passwords
 
   def deletecmd
     super.insert(1,"-F")
@@ -31,7 +35,11 @@ Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
   def modifycmd(param,value)
      cmd = super(param, value)
      cmd << "-F"
-     if self.trusted == "Trusted"
+     trusted_sys = trusted
+     if(trusted_sys == :true ) then
+     # JP - Append on additional command to reset password age to 0
+     # until work around with expiry module can be found for trusted
+     # computing.
         cmd << ";"
         cmd << "/usr/lbin/modprpw"
         cmd << "-v"
@@ -77,41 +85,18 @@ Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
   end
 
   def trusted
-      trusted_sys = %x(/usr/lbin/getprpw root 2>&1)
-      if trusted_sys.chomp == "System is not trusted."
-         "NotTrusted"
+  # JP - Check to see if the HP-UX box is a running in trusted computer mode
+  # UID for root should always be 0
+      trusted_sys = exec_getprpw('root','-m uid')
+      if trusted_sys.chomp == "uid=0"
+        return :true
       else
-         "Trusted"
+        return :false
       end
+      nil
   end
 
-  def trust2
-      trusted_sys = %x(/usr/lbin/getprpw root 2>&1)
-      if trusted_sys.chomp == "System is not trusted."
-         false
-      else
-         true
-      end
-  end
-
-  def password_min_age
-    ent= Etc.getpwnam(resource.name)
-    temp = %x( /usr/lbin/getprdef -m mintm ).chomp
-    mintm = temp[/mintm=(.*)/, 1]
-    if mintm == ""
-       return nil
-    else
-       return mintm
-    end
-  end
-
-  def password_max_age
-    temp = %x( /usr/lbin/getprdef -m exptm ).chomp
-    maxtm = temp[/exptm=(.*)/,1 ]
-    if maxtm == ""
-       return nil
-    else
-       return maxtm
-    end
+  def exec_getprpw(user,opts)
+        %x(/usr/lbin/getprpw #{opts} #{user} 2>&1)
   end
 end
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/user/hpux.rb
```diff
@@ -1,15 +1,18 @@
 Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
   desc "User management for HP-UX. This provider uses the undocumented `-F`
     switch to HP-UX's special `usermod` binary to work around the fact that
-    its standard `usermod` cannot make changes while the user is logged in."
+    its standard `usermod` cannot make changes while the user is logged in.
+    New functionality provides for changing trusted computing passwords and 
+    resetting password exipirations under trusted computing"
```
kylog: nit: "expirations" is misspelled.

MikaelSmith:
It appears your spec tests haven't been appropriately mocked; i.e. they're trying to actually call system binaries. That's why the Travis build failed. Please look into fixing the failing tests.

lib/puppet/provider/user/hpux.rb
```diff
@@ -63,4 +75,43 @@ def password
...
+  def trusted
+      trusted_sys = %x(/usr/lbin/getprpw root 2>&1)
+      if trusted_sys.chomp == "System is not trusted."
+         "NotTrusted"
+      else
+         "Trusted"
+      end
+  end
+
+  def trust2
```
MikaelSmith: This appears to be nearly identical to #trusted, and unused. Why is it included?

lib/puppet/provider/user/hpux.rb
```diff
@@ -63,4 +75,43 @@ def password
       return ent.passwd
     end
   end
+
+  def trusted
+      trusted_sys = %x(/usr/lbin/getprpw root 2>&1)
+      if trusted_sys.chomp == "System is not trusted."
```
MikaelSmith: The default behavior if getprpw isn't found or returns an error is to label the system "Trusted". Is that the desired default?

lib/puppet/provider/user/hpux.rb
```diff
@@ -19,14 +22,23 @@
...
   def deletecmd
     super.insert(1,"-F")
   end
 
   def modifycmd(param,value)
-    super.insert(1,"-F")
+     cmd = super(param, value)
+     cmd << "-F"
+     if self.trusted == "Trusted"
+        cmd << ";"
```
MikaelSmith: What is this doing? Why use a semi-colon - which appears to just ignore the 1st command - instead of just returning the modprpw call?

jjperry69:
This is because I have no clue how to mock anything and the only guide I 
have the rspec book that really has not details that i can comprehend on 
proper mocking.

I am a system admin trying to fix a deficiency for the HP-UX support.

If you have any better references or clearer details on proper mocking, 
I am all ears :)

lib/puppet/provider/user/hpux.rb
jjperry69: The trusted returns a string, the trust2 returns a boolean.

This is all used for the mocks / rspec code, not the actual functional 
coding required to make it work.

Again I have not found any good references on writing rspec 
specifications to test this code or proper mocking, so I have had to 
hack the code to make the testing properly return as I expected.

Any suggestions to references would be greatly appreciated.

lib/puppet/provider/user/hpux.rb
jjperry69: Good catch!  I will go back to look over this and test if changing the 
test case order breaks anything.

lib/puppet/provider/user/hpux.rb
jjperry69: I could not find any code references on the best way to make it append 
the command to have
/usr/sbin/lbin/usermod.sam ... -F  --> Which changes the password hash 
for the user.

and, if trusted,

Add in a second command to reset the password aging to 0 days via 
/usr/lbin/modprpw -v -k <user name>.

This is the command I run manually when I am pushing a user password 
hash to a system and then resetting their expiration back to 0 days.

Suggestions on a better way to implement this routine?  I am still 
learning Ruby and trying to find the best way to get this code to 
properly change the passwords as this is part of the base Puppet code, 
versus having been a module.

Thanks!

jjperry69: Question on mocking this for testing...

Instead of
        if provider.trust2
        resource.stubs(:allowdupe?).returns true
        provider.expects(:execute).with() { |args| 
args.include?('/usr/lbin/modprpw') and args.include?("-v") and 
args.include?("-l") }
        provider.uid = 1000

Would the correct be to do:

        let (:provider.trusted) { "Trusted" }
        resource.stubs(:allowdupe?).returns true
        provider.expects(:execute).with() { |args| 
args.include?('/usr/lbin/modprpw') and args.include?("-v") and 
args.include?("-l") }
        provider.uid = 1000

To test the below command works properly?

As said before I am learning so I am trying to figure this all out and 
clean things up where required.

I have not had much success inspecting the :provider / 
:resource.provider  to see what is contained within when I am trying to 
troubleshoot what is going on. It seems that the puppet wrapper around 
things is causing some difficulty getting useful debug output. I know 
this is just me learning Ruby, but it is frustrating all the same.

jjperry69:
It will not permit me to sign the agreement saying my jjperry69@gmail.com
address is already in use since I am using that on my GitHub account.

Do I have to use another email address than the one I use for GitHub?

kylog:
@jjperry69 you don't need to use a separate email. Sounds like the CLA bot is confused somehow. I'll ask the folks who manage that bot.

geekygirldawn:
@jjperry69 @kylog OK, I think I fixed this on the back end. It looks like @jjperry69 signed a CLA from the old Redmine system, but the GitHub account wasn't properly associated. Future pull requests should be fine.

@kylog you can consider this pull request approved from a the human CLAbot :)

In the future, @jeffmccune is actually responsible for the CLAbot now, but this was an easy fix, so I thought I would save Jeff some time.

lib/puppet/provider/user/hpux.rb
MikaelSmith: If trust2 is purely for testing, I would move it's implementation to the spec test. Simpler interfaces are easier to maintain.

I would look at https://github.com/puppetlabs/puppet/blob/master/docs/rspec_tutorial.md for an introduction to how we write specs. There are many examples in spec/unit/* once you've read the tutorial. The provider/user/hpux_spec.rb that you modified has examples of using the #stub method, which are what you want to emulate.

jjperry69: I will look it over again. I am just not able to get any useful data out of
the ruby and spec scripts as to what is going on under the covers.  The
only way to know if the box is trusted is to go out and check for it. I
have tried multiple ways to get it to work without having the command there
and nothing seems to work.

lib/puppet/provider/user/hpux.rb
MikaelSmith: I'll see about coming up with specs that work for this.

jjperry69: Thanks.  The self.trusted is the part I have not been able to mock out in a
spec file.

MikaelSmith: I think I just addressed that in
https://groups.google.com/d/msg/puppet-dev/MG-zwQYaKYM/IlXueb9pjbUJ.

jjperry69:
Coding issues and logic were corrected and rspec tests correctly mocked now to check for Trusted Computing and untrusted systems.

lib/puppet/provider/user/hpux.rb
```diff
@@ -1,7 +1,13 @@
 Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
   desc "User management for HP-UX. This provider uses the undocumented `-F`
     switch to HP-UX's special `usermod` binary to work around the fact that
-    its standard `usermod` cannot make changes while the user is logged in."
+    its standard `usermod` cannot make changes while the user is logged in.
+<<<<<<< HEAD
```
MikaelSmith: This is a merge conflict that needs to be resolved. Git adds these when it can't automatically merge based on commit history. You should be able to remove lines 5 and 7-9.

ScottGarman:
@jjperry69 thanks for your contribution. I'm happy to assist by cherry-picking your changes and doing the final cleanups. I should be able to do this in such a way so as to keep your name in the commit.

jjperry69:
Thanks.  I am more interested in getting them out there for others to 
use than I am with getting the credit for the work.  It would be nice to 
have it for bragging rights, but it is not required.

ScottGarman:
I've done some minor cleanups and started a new PR:

https://github.com/puppetlabs/puppet/pull/3492

In the meantime we'll close this one.

PR #3444, opened 2015-01-01
(PUP-3645) Add support for configurable url prefix

rlinehan:
This commit support for a configurable setting `master_url_prefix`, which
defaults to `/puppet`, that the master's http endpoints are mounted at. All
requests generated by the agent honor this prefix. This will allow us to mount
the Puppet HTTP endpoints in a namespaced URL space alongside other apps
running on the same webserver.

### DIFF
--- a/lib/puppet/network/http/webrick/rest.rb
+++ b/lib/puppet/network/http/webrick/rest.rb
@@ -15,8 +15,11 @@ class Puppet::Network::HTTP::WEBrickREST < WEBrick::HTTPServlet::AbstractServlet
 
   def initialize(server)
     raise ArgumentError, "server is required" unless server
-    register([Puppet::Network::HTTP::API::V3.routes,
-              Puppet::Network::HTTP::API::V2.routes])
+    prefix = Regexp.new("^#{Puppet[:master_url_prefix]}")
+    register([Puppet::Network::HTTP::Route.path(prefix).
+                  any.
+                  chain(Puppet::Network::HTTP::API::V3.routes,
+                        Puppet::Network::HTTP::API::V2.routes)])
     super(server)
   end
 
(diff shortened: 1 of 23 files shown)


lib/puppet/network/http/webrick/rest.rb
```diff
@@ -15,8 +15,12 @@ def self.mutex
 
   def initialize(server)
     raise ArgumentError, "server is required" unless server
-    register([Puppet::Network::HTTP::API::V3.routes,
-              Puppet::Network::HTTP::API::V2.routes])
+    prefix = Regexp.new("^#{Puppet[:master_url_prefix]}")
```
cprice404: We chatted at some point about how we might need to include a trailing slash in the regex in order to make sure it falls through to the 404 properly.  Is that something that only affects the subsequent PR?  Or did you find a different solution?

cprice404:
+1 ; will leave this open for a few more days in case other folks want to review it as well, but will plan on merging by mid-week unless anyone raises concerns.

lib/puppet/network/http/webrick/rest.rb
rlinehan: @cprice404 yup, that only affects the subsequent PR, and I've made that change there.

nwolfe:
@rlinehan I was surprised when I discovered that I didn't need an exact match with the URL prefix.  For example, I started my master with a prefix of `foo` and then did an agent run with a prefix of `foo1` expecting it to fail but it worked.  I suppose it's still a "prefix" and not an exact match...

Is this desirable? Or is it unavoidable? Is there a real-world scenario in which this could cause problems, like if there were multiple masters with prefixes like `ps`, `ps1`, `ps2` etc.?

rlinehan:
@nwolfe hah yep, you have found the hidden easter egg, or something. This is fixed/worked around in the next PR (#3464) by adding a slash to the end of the prefix when the routes are defined, because this was causing problems when the CA routes were separated out with a default prefix of `/puppet-ca`.

nwolfe:
:+1: looks good! Verified `master_url_prefix` is honored by agent and CLI tools (`certificate`, `certificate_revocation_list`, etc).

fpringvaldsen:
:+1: Verified the prefix is honored by the agent.

PR #3445, opened 2015-01-02
(maint) Update deep_merge to 1.0.1 on x86

MikaelSmith:
Use the updated puppet-win32-ruby including deep_merge 1.0.1 for x86
builds. x64 builds already used 1.0.1, so this makes the gem versions
consistent.


melissaanne:
:+1:  the only additional commit this update pulls into the build is `ae68cf2 (maint) Update deep_merge to 1.0.1`, which updates the deep_merge gem here to match the one provided in the x64 installer

PR #3449, opened 2015-01-06
(PUP-3750) Fix specs on Windows

MikaelSmith:
Used Linux-style paths while requiring that they be absolute paths. Fixed
so that Windows testing uses Windows-style paths that are correctly
interpreted as absolute paths.

### DIFF
--- a/spec/unit/type/mailalias_spec.rb
+++ b/spec/unit/type/mailalias_spec.rb
@@ -4,13 +4,19 @@ require 'spec_helper'
 describe Puppet::Type.type(:mailalias) do
   include PuppetSpec::Files
 
+  if Puppet.features.microsoft_windows?
+    tmpfile_path = 'C:\\afile'
+  else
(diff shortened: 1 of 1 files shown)


cyberious:
:+1:

spec/unit/type/mailalias_spec.rb
```diff
@@ -4,13 +4,19 @@
 describe Puppet::Type.type(:mailalias) do
   include PuppetSpec::Files
 
+  if Puppet.features.microsoft_windows?
```
Iristyle: Puppet already has helpers for temporary files and directories that abstract these details (and keep track of the generated paths to clean up later):

See the definition of `tmpfile` here:
https://github.com/puppetlabs/puppet/blob/c157bc9c2ee6422457bcb2bda70eadbf465c7c6c/spec/lib/puppet_spec/files.rb#L27

Grep on `tmpfile` and you'll see it's usage all over the place in the codebase.

PR #3451, opened 2015-01-06
(PUP-1343) Use chkconfig --add to ensure K-symlinks get installed

ScottGarman:
On RHEL/CentOS 5 systems, the use of chkconfig on to enable a service
does not create the K-symlinks in /etc/rc.d to shut down the service.
Running chkconfig --add before chkconfig on fixes this issue, and
should be harmless on the unaffected OS versions.

The rspec unit test has been updated, and a new acceptance test
included to validate the bugfix.


joshcooper:
I was going to backport the fixes in master, but I had a question about whether it is safe to execute `chkconfig --add <name>` on rhel6 and up. If it is, then we should update the acceptance test so it runs on all rhel platforms. If not, then we need to change the provider. @ScottGarman I'm going to close this PR, please refile with your updated version.

PR #3453, opened 2015-01-07
(docs) Clarify meaning of `schedule` key in scheduled_task trigger hashes

nfagerlund:
When this is merged, please resolve DOC-1509.


joshcooper:
:+1: /cc @binford2k @llowder

PR #3454, opened 2015-01-07
(maint) Add Puppet::Indirector::Request#description for better errors

rlinehan:
The `to_s` method on Puppet::Indirector::Request was removed in
fd7c4f912d3f26977863faaec91e010c9bc48abb, without realizing that it was used
to make two error messages more useful. This adds back this functionality,
renamed to `description`.


cprice404:
+1 @joshcooper this should hopefully address the error messages that I borked when I got rid of that `to_s`.  Let us know if you think of any other reasons why it would need to be `to_s`.

PR #3459, opened 2015-01-08
(WIP DO NOT MERGE)(PUP-3837) Ensure Ruby 2.1.5 on Windows

ferventcoder:
This provides the necessary changes so that Puppet will run with Ruby 2.1.5 on Windows.

* Confined Hpux.password spec not to run on Windows. See commit for specifics.

### DIFF
--- a/lib/puppet/util/windows/registry.rb
+++ b/lib/puppet/util/windows/registry.rb
@@ -32,15 +32,20 @@ module Puppet::Util::Windows
 
     def values(subkey)
       values = {}
-      subkey.each_value do |name, type, data|
-        case type
-        when Win32::Registry::REG_MULTI_SZ
-          data.each { |str| force_encoding(str) }
-        when Win32::Registry::REG_SZ, Win32::Registry::REG_EXPAND_SZ
-          force_encoding(data)
+      begin
+        subkey.each_value do |name, type, data|
+          case type
+            when Win32::Registry::REG_MULTI_SZ
+              data.each { |str| force_encoding(str) }
+            when Win32::Registry::REG_SZ, Win32::Registry::REG_EXPAND_SZ
+              force_encoding(data)
+          end
+          values[name] = data
         end
-        values[name] = data
+      rescue Encoding::UndefinedConversionError => error
+        raise Puppet::Error.new("Failed to get registry key values for '#{subkey.name}'.\n  #{error.to_s}")
       end
+
       values
     end
 
--- a/lib/puppet/util/windows/registry.rb
+++ b/lib/puppet/util/windows/registry.rb
@@ -32,48 +32,21 @@ module Puppet::Util::Windows
 
     def values(subkey)
       values = {}
+      orig_enc = Encoding.default_internal
+
       begin
+        Encoding.default_internal = Encoding::UTF_8
         subkey.each_value do |name, type, data|
-          case type
-            when Win32::Registry::REG_MULTI_SZ
-              data.each { |str| force_encoding(str) }
-            when Win32::Registry::REG_SZ, Win32::Registry::REG_EXPAND_SZ
-              force_encoding(data)
-          end
           values[name] = data
         end
-      rescue Encoding::UndefinedConversionError => error
+      rescue Encoding::UndefinedConversionError, Encoding::CompatibilityError => error
         raise Puppet::Error.new("Failed to get registry key values for '#{subkey.name}'.\n  #{error.to_s}")
+      ensure
+        Encoding.default_internal = orig_enc
       end
 
       values
     end
-
-    if defined?(Encoding)
-      def force_encoding(str)
-        if @encoding.nil?
-          # See https://bugs.ruby-lang.org/issues/8943
-          # Ruby uses ANSI versions of Win32 APIs to read values from the
-          # registry. The encoding of these strings depends on the active
-          # code page. However, ruby incorrectly sets the string
-          # encoding to US-ASCII. So we must force the encoding to the
-          # correct value.
-          begin
-            cp = GetACP()
-            @encoding = Encoding.const_get("CP#{cp}")
-          rescue
-            @encoding = Encoding.default_external
-          end
-        end
-
-        str.force_encoding(@encoding)
-      end
-    else
-      def force_encoding(str, enc)
-      end
-    end
-    private :force_encoding
-
 
     ffi_convention :stdcall
 
--- a/spec/unit/provider/user/hpux_spec.rb
+++ b/spec/unit/provider/user/hpux_spec.rb
@@ -4,7 +4,7 @@ require 'etc'
 
 provider_class = Puppet::Type.type(:user).provider(:hpuxuseradd)
 
-describe provider_class do
+describe provider_class, :unless => Puppet.features.microsoft_windows? do
   let :resource do
     Puppet::Type.type(:user).new(
       :title => 'testuser',


kylog:
Wild. I'm a little surprised that this is the only spec that happens to tickle `Etc::Password.passwd`. Two thoughts:

1) are we building ruby 2.1.5 wrong somehow? I'm just wondering if this is some kind of canary wrt the ruby 2.1.5 build for windows.

2) if not, then one suggestion here is to confine not just this one spec test, but the entire spec context ("managing passwords") not to run on windows.

My reasoning is that the value of testing the hpux specs on windows is low, and that a slightly higher level confine would insulate windows specs from future changes in hpux land. E.g. there's a PR open now around hpux password management, so it would be a shame to add this confine now and then have to do the same change for some upcoming (unrelated) PRs. (I haven't audited that PR for this issue - my point is that such an audit isn't a great use of our time.)

joshcooper:
I'm also curious what this means for our 2.1.5 ruby builds on *nix platforms. Are we using the right compile flags there? Sadly we would have to run specs on the ruby we build to see if there are problems, or rely entirely on acceptance. It also means Travis CI won't tell us if there are problems.

ferventcoder:
@kylog This is not with our vendored Ruby, this is with vanilla RubyInstaller Ruby.

ferventcoder:
>2) if not, then one suggestion here is to confine not just this one spec test, but the entire spec context ("managing passwords") not to run on windows.

>My reasoning is that the value of testing the hpux specs on windows is low, and that a slightly higher level confine would insulate windows specs from future changes in hpux land. E.g. there's a PR open now around hpux password management, so it would be a shame to add this confine now and then have to do the same change for some upcoming (unrelated) PRs. (I haven't audited that PR for this issue - my point is that such an audit isn't a great use of our time.)

Good point, I was thinking about that but didn't want to jump the gun on the whole thing right off. Sounds like folks are in agreement about confining the entire thing, so I will push that back up.

kylog:
> @kylog This is not with our vendored Ruby, this is with vanilla RubyInstaller Ruby.

Ah okay. Well, same question wrt rubyinstaller's ruby :)

Basically, it seems like ideally we would audit any ruby we're going to ship for what fields may be missing. Not sure how practical that is.

ferventcoder:
> Basically, it seems like ideally we would audit any ruby we're going to ship for what fields may be missing. Not sure how practical that is.

Isn't this what we already do in part with our spec / integration / acceptance testing? We are validating the world works for Puppet on the Ruby platform it is running on. Fields that are used that are missing should come up just like it did here.

kylog:
>> Basically, it seems like ideally we would audit any ruby we're going to ship for what fields may be missing. Not sure how practical that is.

> Isn't this what we already do in part with our spec / integration / acceptance testing? We are validating the world works for Puppet on the Ruby platform it is running on. Fields that are used that are missing should come up just like it did here.

In part, yes. Puppet isn't the only thing that uses the ruby we ship. So *if* we could audit this, that would still be ideal.

lib/puppet/util/windows/registry.rb
```diff
@@ -31,16 +31,24 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
     end
 
     def values(subkey)
+      orig_enc = Encoding.default_internal
+      Encoding.default_internal = Encoding::UTF_8
```
ferventcoder: @Iristyle @kylog @joshcooper is doing this a good idea? I don't see another way around it. I could go with another encoding here if we think this will be the method but not sufficient. Looks like it removed the endash in favor of just a dash (but I'd need to inspect better to be sure).

joshcooper: Sorry but I'm not following why this is necessary. Ruby uses the wide versions to read data from the registry. Strings should have UTF-16LE encoding at that point. When [ruby's encode method](https://github.com/ruby/ruby/blob/v2_1_5/ext/win32/lib/win32/registry.rb#L894-L896) is called, it should convert from UTF-16LE to the native codepage encoding, e.g. CP1252, and yield to us a CP1252 encoded string.

<pre>
subkey.each_value do |name, type, data|
  # if type == REG_SZ, then data should be CP1252 encoded
end
</pre>

We should then be able to convert that string to UTF-8, `data.encode('UTF-8')`, and we should be able to skip the non-sense that we were doing earlier in https://github.com/ferventcoder/puppet/blob/ca76c70d06bcbb379b28f4549a38f66a43176623/lib/puppet/util/windows/registry.rb#L55-L78

ferventcoder: You would think that is the case, but sadly it is not. Have you read the commit this is related to?

ferventcoder: When it converts to the native locale_charmap (in my case that is `IBM437`), instead of a nice successful conversion, you get this nice fun instead:

```
Encoding::UndefinedConversionError:
       U+2013 to IBM437 in conversion from UTF-16LE to UTF-8 to IBM437
```

`Encoding.default_internal` is `nil` by default. Falling back to rely on whatever the native encoding is doesn't handle converting strings correctly. It's probably better to go with whatever the registry itself adheres to. And stay there. 

Sounds like UTF-16LE?

ferventcoder: For more context it errors in the `each_value` when it calls `export_string` because it is trying to convert to the system locale.

The package in question that can make it happen every time is `Windows Azure Libraries for .NET 1.7 â?" June 2012` (although I have somewhere between 4 and 6 that fail) without this change. Our handling code works fine on Ruby 1.9.3 and Ruby 2.0.0 because they don't call the export_string trying to handle encoding for us.

ferventcoder: Seems dangerous to rely on the code page since we have no control over what it is.

[1] pry(main)> Encoding.locale_charmap
=> "CP437"
[2] pry(main)> .chcp 1252
Active code page: 1252
[3] pry(main)> Encoding.locale_charmap
=> "CP1252"


It seems that the ruby developers should have set the locale to Encoding.default_external, not internal (which is `nil` by default).

ferventcoder: I forgot to mention that setting `Encoding.default_internal = Encoding::UTF_8` will run it through https://github.com/ruby/ruby/blob/v2_1_5/ext/win32/lib/win32/registry.rb#L894-L896 and encode it as UTF-8 handed back to us. So it's likely we could remove the force_encoding() although I would be hesitant to remove the method strictly for converting each line of the `REG_MULTI_SZ`.

Iristyle: Withholding comment until I can test locally.

Iristyle: So the problem here @joshcooper is that inside of Ruby's registry code, they're doing some puzzling stuff.

In their `each_key` and `each_value` methods they're using the [RegEnumKeyEx](http://msdn.microsoft.com/en-us/library/windows/desktop/ms724862\(v=vs.85\).aspx) and [RegEnumValueW](http://msdn.microsoft.com/en-us/library/windows/desktop/ms724865\(v=vs.85\).aspx) API calls respectively.

As expected, the values returned from these calls include some `lpName [out]` which are `UTF-16LE`.  So far, so good.  Verified during 2.1.5 debugging, the string has the appropriate encoding after the API call returns.

Unfortunately, https://github.com/ruby/ruby/commit/79e880c90df6938e8226336a883d1973031da3df added the `export_string` function.  (For reference, see the other "interesting" commits following that one via https://github.com/ruby/ruby/commits/802d4f9fb2a8a5f2be1eb6be13dc4f286ad221af/ext/win32/lib/win32/registry.rb):

```ruby
def export_string(str, enc = Encoding.default_internal || LOCALE) # :nodoc:
  str.encode(enc)
end
```

As @ferventcoder pointed out, `Encoding.default_internal` will typically be `nil` by default, and Ruby hardcodes `LOCALE` with `LOCALE = Encoding.find(Encoding.locale_charmap)`.  The `LOCALE` here is the fallback, and is set based on current code page - which will be system dependent.  On my system, default is `1252` and everything is OK, but on other systems it could be `437` which means certain `UTF-16LE` characters in reg key or value names (like endash) might not translate properly.

From `each_value`, we can see that we have a `UTF-16LE` conversion to `LOCALE`:

```ruby
begin
  #### Wrapper for RegEnumValueEx that returns UTF-16LE
  subkey = API.EnumValue(@hkey, index)
rescue Error
  break
end
##### TRANSLATE out of UTF-16LE to LOCALE
subkey = export_string(subkey)
begin
  type, data = read(subkey)
rescue Error
  next
end
```

And then of course, when we look at what [read](https://github.com/ruby/ruby/blob/802d4f9fb2a8a5f2be1eb6be13dc4f286ad221af/ext/win32/lib/win32/registry.rb#L641) and [QueryValue](https://github.com/ruby/ruby/blob/802d4f9fb2a8a5f2be1eb6be13dc4f286ad221af/ext/win32/lib/win32/registry.rb#L315-L323) (wrapper over [RegQueryValueEx](http://msdn.microsoft.com/en-us/library/windows/desktop/ms724911%28v=vs.85%29.aspx)) do, it's more evident why this is problematic.  As you might expect, `make_wstr` calls `str.encode(WCHAR)` where `WCHAR = Encoding::UTF_16LE`.

```ruby
def read(name, *rtype)
  #### now call RegQueryValueEx with a LOCALE encoded string
  type, data = API.QueryValue(@hkey, name)
  unless rtype.empty? or rtype.include?(type)
    raise TypeError, "Type mismatch (expect #{rtype.inspect} but #{type} present)"
  end
  case type
  when REG_SZ, REG_EXPAND_SZ
    #### encode accepts dst_encoding, src_encoding
    #### so this translates UTF-16LE data back to LOCALE
    [ type, data.encode(name.encoding, WCHAR).chop ]
  when REG_MULTI_SZ
    [ type, data.encode(name.encoding, WCHAR).split(/\0/) ]

#### SNIP

def QueryValue(hkey, name)
  type = packdw(0)
  size = packdw(0)
  #### LOCALE encoded string turned back into UTF-16LE
  name = make_wstr(name)
  check RegQueryValueExW.call(hkey, name, 0, type, 0, size)
  data = "\0".force_encoding('ASCII-8BIT') * unpackdw(size)
  check RegQueryValueExW.call(hkey, name, 0, type, data, size)
  [ unpackdw(type), data[0, unpackdw(size)] ]
end
```

As you can see, Ruby has:

* Read the value names via `RegEnumValueW` as `UTF-16LE`
* Translated those to `LOCALE`, whatever that may be
* Round-tripped back to `UTF-16LE` prior to calling `RegQueryValueExW`
* Depending on the type of the `data` returned, a `UTF-16LE` string value is translated back to `LOCALE`

Clearly the right thing to do here is to leave the `UTF-16LE` string alone prior to calling `RegQueryValueExW`, and it looks like the only good way to leave that string intact and not introduce unnecessary roundtrips is:

- Set `Encoding.default_internal` to `UTF-16LE`
- Set `LOCALE` to `UTF-16LE`
- Monkey patch the Ruby code so that `export_string` isn't re-encoding strings unnecessarily as this appears to be the only affected call-site

These can be done at a global level, which seems rather dangerous, or they can be done at a more local level when calling into these Ruby methods.  The safer approach is clearly (as long as there's a proper try/catch/finally style mechanism) to do this in a more local level.  It's possible Ruby will realize the error of this approach and fix in a later version, so patching `export_string` seems dangerous as well.

Also note, that there is another spot where `LOCALE` is forced that could be problematic (and requires additional investigation), [self.expand_environ](https://github.com/ruby/ruby/blob/802d4f9fb2a8a5f2be1eb6be13dc4f286ad221af/ext/win32/lib/win32/registry.rb#L372-L378) which is used whenever `REG_EXPAND_SZ` values are encountered.

```ruby
def self.expand_environ(str)
  str.gsub(Regexp.compile("%([^%]+)%".encode(str.encoding))) {
    v = $1.encode(LOCALE)
    (e = ENV[v] || ENV[v.upcase]; e.encode(str.encoding) if e) ||
    $&
  }
end
```

Sigh, Ruby.

ferventcoder: Thanks for summing that up. +1. Better explanation than I gave.

lib/puppet/util/windows/registry.rb
```diff
@@ -32,43 +32,21 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
...
+      rescue Encoding::UndefinedConversionError, Encoding::CompatibilityError => error
+        raise Puppet::Error.new("Failed to get registry key values for '#{subkey.name}'.\n  #{error.to_s}")
+      ensure
+        Encoding.default_internal = orig_enc
       end
-    end
-    private :force_encoding
 
+      values
+    end
```
ferventcoder: That diff looks horrible. But the premise is that `force_encoding` went away.

ferventcoder:
When this gets picked up, registry specs are failing. Should be able to fixup by removing the strip ansi values part of the test and evaluate those tests more explicitly for value.

lib/puppet/util/windows/registry.rb
ferventcoder: Also en-dash gets converted to a dash now that we don't have force_encoding... so yay for that.

Iristyle:
Let's open a new PR to include just the first commit - 0cd462d

Then I'm :+1: on merging that to fix CI.  We can continue the registry discussion on this issue.

Iristyle:
Updated PR in #3476 - we can strip that commit from this guy and continue discussion about registry updates.

ferventcoder:
Superseded by https://github.com/puppetlabs/puppet/pull/3540

PR #3460, opened 2015-01-08
(QENG-1562) All-In-One Agent Beaker usage added

kevpl:


acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -54,6 +54,15 @@ def install_packages_on(hosts, package_hash, options = {})
         return true
       end
 
+      def install_aio_on(hosts)
+        check_if_exists = options[:check_if_exists]
+        hosts = [hosts] unless hosts.kind_of?(Array)
+        hosts.each do |host|
+          install_puppetagent_dev_repo host, options[:aio]
```
kylog: Does this require a newer version of beaker than what puppet is specifying now?

acceptance/setup/packages/pre-suite/010_Install.rb
```diff
@@ -46,4 +46,8 @@
 }
 
 install_packages_on(master, MASTER_PACKAGES)
-install_packages_on(agents, AGENT_PACKAGES)
+if ENV['AIO_AGENT_INSTALL'].nil?
```
kylog: Not a functional change, but for readability i'd reverse the if/else clauses to drop the nil?. i.e.

```
if ENV['AIO_AGENT_INSTALL']
  install_aio_on(agents)
else
end
```

One less negative to process and the "if aio, do aio stuff" reads cleanly.

haus:
@kevpl There are repos available for the AIO just like for puppet or facter, so we shouldn't need to resort to using rpm or dpkg directly (apt-get and yum should be just fine). The repos live at http://builds.puppetlabs.lan/puppet-agent/e5922492cce4c236e1d850c8a9a4649f3b367e47/repos/ and repo configs live at http://builds.puppetlabs.lan/puppet-agent/e5922492cce4c236e1d850c8a9a4649f3b367e47/repo_configs/.

I would expect beaker to already have methods for adding a builds.pl.lan repo to the test machine and installing a package from it, but I could be crazy.

acceptance/setup/packages/pre-suite/010_Install.rb
haus: Along with the repo comment i left below, it seems like this could be...

```
if ENV['AIO_AGENT_INSTALL']
  install_packages_on(agents, 'puppet-agent')
else
end
```

That assumes the correct repo has already been added for the puppet-agent.

kevpl:
@kylog to your comment, yes.  It requires Beaker 2.2.0 to get the new paths required for the All-In-One agent, even if we were able to use `install_packages_on`.  Let me know if because of this I should be doing something differently here.  I'm not familiar with who I should be talking to about which version of Beaker you guys use, would that be Roger? 

@haus, `install_packages_on` doesn't do any package proxying that I'm aware of, so unless `sudo apt-get install puppet-agent` will work, we'll need to use something else.  Now that you have the hash directories on the build server, we can switch over to using `install_puppetlabs_dev_repo`, which did already exist though.  I've done this in my latest version of the change.

Let me know if you guys have any questions about this, or need me to do anything for you to get this moving.

Thanks,
Kevin

kylog:
@kevpl sorry, lost track of this PR. Is this still live or has it been overtaken by @justinstoller's work?

If still live, it looks like it needs a rebase, but otherwise I'm :+1: 

/cc @haus

kylog:
Ping @kevpl on questions above.

kevpl:
sorry @kylog about not getting back to you sooner on this.  I've spoken with Justin about it, and we're actually coming at these as separate passes.  

He's doing what he can to make AIO work in the Jenkins environments with what he has in the projects as a quick & dirty solution, and I'm coming around as a second pass with the Beaker supported way of installing these packages from the dev repos.

Unfortunately due to new work for AIO issues that have come up in testing, this code needs something that has been merged but not released in Beaker yet ([#638](https://github.com/puppetlabs/beaker/pull/638)).  Would you rather I just close this PR until I have something released and can have the conversation with Justin and Roger about getting that version of Beaker into testing for this project?

kylog:
@kevpl sounds great and thanks for the update.

Yeah I'd close it for now and re-open it once the dependencies are resolved and it's ready for review.

Thanks!

PR #3461, opened 2015-01-08
(PUP-3488) Stop overwriting agent's facts timestamp

whopper:
This commit undoes a change made in commit fb5f859cf4 that
causes the master to alter the agent's facts timestamp when
they are received by the catalog compiler.

With this change, the facts timestamp is always based on the
agent's time.

### DIFF
--- a/spec/unit/indirector/catalog/compiler_spec.rb
+++ b/spec/unit/indirector/catalog/compiler_spec.rb
@@ -150,15 +150,12 @@ describe Puppet::Resource::Catalog::Compiler do
       @compiler.extract_facts_from_request(request).should be_nil
     end
 
-    it "deserializes the facts and timestamps them" do
-      @facts.timestamp = Time.parse('2010-11-01')
+    it "should deserialize the facts without changing the timestamp" do
+      time = Time.now
+      @facts.timestamp = time
       request = a_request_that_contains(@facts)
-      now = Time.parse('2010-11-02')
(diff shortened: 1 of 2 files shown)


spec/unit/indirector/catalog/compiler_spec.rb
```diff
@@ -150,15 +150,9 @@ def a_request_that_contains(facts)
...
-    it "deserializes the facts and timestamps them" do
-      @facts.timestamp = Time.parse('2010-11-01')
+    it "should deserialize the facts" do
       request = a_request_that_contains(@facts)
-      now = Time.parse('2010-11-02')
-      Time.stubs(:now).returns(now)
-
       facts = @compiler.extract_facts_from_request(request)
-
-      facts.timestamp.should == now
```
joshcooper: It doesn't look like we are "expecting" anything? Perhaps we should set an explicit timestamp on the facts, generate the request, extract the facts, and verify the timestamp is the same?

<pre>
time = Time.now
@facts.timestamp = time
request = a_request_that_contains(@facts)
facts = @compiler.extract_facts_from_request(request)
expect(facts.timestamp).to eq(time)
</pre>

whopper: @joshcooper ah yeah, I was going off of the original test, which essentially just did this (I had the same thought :)) I also toyed with the timestamp idea. I can add that in.

PR #3462, opened 2015-01-09
(PUP-3755) Make compile for a node use compiler's context for transform

hlindberg:
Without this fix, the final step of compiling a catalog for a node will
transform the produced catalog to resource form in a context that is not
configured for the environment used to compule the catalog. This can
lead to a wide variety of problems since there is a high chans that
unrelated code gets loaded and this code may have errors, may cause side
effects and cause the catalog to silently contain a result different
from the intended.

The fix is simple - the compile method now accepts a block that it
yields the produced catalog to if the block is given in a call to
compile. This is then used instead of simply calling to_resource on the
produced result.

As a conseuence of the change mocking tests were failing. As it was
found that these tests were meaningless and testing nothing they were
removed.

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -31,7 +31,7 @@ class Puppet::Parser::Compiler
       raise(Puppet::Error, errmsg.join(' '))
     end
 
-    new(node).compile.to_resource
+    new(node).compile {|resulting_catalog| resulting_catalog.to_resource }
   rescue => detail
     message = "#{detail} on node #{node.name}"
     Puppet.log_exception(detail, message)
@@ -140,7 +140,11 @@ class Puppet::Parser::Compiler
 
       fail_on_unevaluated
 
-      @catalog
+      if block_given?
+        yield @catalog
+      else
+        @catalog
+      end
     end
   end
 
--- a/spec/integration/parser/compiler_spec.rb
+++ b/spec/integration/parser/compiler_spec.rb
@@ -522,4 +522,20 @@ describe "Puppet::Parser::Compiler" do
       expect(catalog).to have_resource("Notify[nbr2]").with_parameter(:message, 'overridden')
     end
   end
+
+
+  context "when converting catalog to resource" do
+    it "the same environment is used for compilation as for transformation to resource form" do
+        Puppet[:code] = <<-MANIFEST
+          notify { 'dummy':
+          }
+        MANIFEST
+
+      Puppet::Parser::Resource::Catalog.any_instance.expects(:to_resource).with do |catalog|
+        Puppet.lookup(:current_environment).name == :production
+      end
+
+      Puppet::Parser::Compiler.compile(Puppet::Node.new("mynode"))
+    end
+  end
 end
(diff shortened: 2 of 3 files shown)


lib/puppet/parser/compiler.rb
```diff
@@ -31,7 +31,7 @@ def self.compile(node)
       raise(Puppet::Error, errmsg.join(' '))
     end
 
-    new(node).compile.to_resource
+    new(node).compile {|resulting_catalog| resulting_catalog.to_resource }
```
joshcooper: @hlindberg could you add a test to confirm that `to_resource` was performed in the context of the *correct* environment? Perhaps check that the compiler generated resources have the correct environment set on them?

hlindberg:
@joshcooper love to add a test, but it is not as easy as checking that the generated resources have the correct environment (they are created and environment set when in context also before the change afaict). It is a much more subtle thing; if something requires access to things in the environment when performing the `to_resource` transformation...

spec/integration/parser/compiler_spec.rb
```diff
@@ -522,4 +522,31 @@ class xport_test {
...
+
+          def active?
+            Puppet.lookup(:current_environment).name == :production
+          end
+
+          def to_s
+            # displayed if the expectation about this state fails
+            "in production environment"
+          end
+        end.new
```
joshcooper: @hlindberg nicely done, I forgot about mocha state machines. This is a bit magical as we're using a generic Class instead of a `Mocha::StateMachine::State` object. Can this be reworked to use the standard mocha APIs? Also I'm wondering if we should instead test that the `to_resource` takes place in non-default environment?

hlindberg: if I understood the StateMachine::State class it is intended to describe a set of static states, and there are methods to set/get and transition between such states. Basically, our need is to return an instance of an object that responds to `active` - the same can be achieved through other less contorted means than the anonymous class I came up with (I thought I needed to implement additional methods) - OTOH, such an implementation will probably be as hard to read as the anonymous class. So - in short; no, I don't see how the standard StateMachine::State can be used - I did not see an apparent way of doing it at least.

I do not think it matters if we test the default env, or some other env. It starts with the *root* env in both cases. I contemplated adding an assertion in the test that the current_environment is not `production` before the test - is that good enough?

joshcooper: @hlindberg I think it's fine as is. I thought it started in production environment. As for the mocha bits, it's a little concerning that knows about mocha internals like that. Let me take a look and see if I can come up with something...

hlindberg: afaikt, the when(state) is the only option to get dynamic code into the equation. Trying to specify it as a mock return and using a block to compute the return did not work at all (it seems to want to bind the value when the mock is created). Maybe something clever can be done with the StateMachine class as opposed to only State... Appreciate you taking a look.

joshcooper: @hlindberg how about https://github.com/joshcooper/puppet/commit/ed9ee3aaf9f4b8095900e6f051123ad1f975430d. Also, btw it looks like one of your commits as PUP-3775 instead of PUP-3755

hlindberg: yeah, that achieves the same, will change to using that, and fixup the commit with the wrong reference and update this PR. Thanks!

hlindberg:
Rebased with corrected commit message and with simplified way of checking environment state

PR #3464, opened 2015-01-09
(PUP-3526) Add configurable prefix for CA routes

rlinehan:
Restructure the CA endpoints to allow for a configurable url prefix, separate from the master url prefix.

### DIFF
--- a/acceptance/tests/environment/directory_environment_production_created_master.rb
+++ b/acceptance/tests/environment/directory_environment_production_created_master.rb
@@ -30,7 +30,7 @@ master_opts = {
 step 'run master; ensure production environment created'
 with_puppet_running_on(master, master_opts, testdir) do
   if master.is_using_passenger?
-    on(master, "curl -k --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} https://localhost:8140/v2.0/environments")
+    on(master, "curl -k --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} https://localhost:8140/puppet/v2.0/environments")
   end
   on(master, "test -d '#{testdir}/environments/production'")
 
--- a/lib/puppet/network/authconfig.rb
+++ b/lib/puppet/network/authconfig.rb
@@ -1,34 +1,39 @@
 require 'puppet/network/rights'
+require 'puppet/network/http'
 
 module Puppet
   class ConfigurationError < Puppet::Error; end
   class Network::AuthConfig
     attr_accessor :rights
 
-    def self.url_prefix
-      Puppet[:master_url_prefix]
+    def self.master_url_prefix
+      Puppet::Network::HTTP::MASTER_URL_PREFIX
+    end
+
+    def self.ca_url_prefix
+      Puppet::Network::HTTP::CA_URL_PREFIX
     end
 
     def self.default_acl
       [
       # API V2.0
-      { :acl => "#{url_prefix}/v2.0/environments", :method => :find, :allow => '*', :authenticated => true },
+      { :acl => "#{master_url_prefix}/v2.0/environments", :method => :find, :allow => '*', :authenticated => true },
 
       # API V3
-      { :acl => "~ ^#{url_prefix}\/v3\/catalog\/([^\/]+)$", :method => :find, :allow => '$1', :authenticated => true },
-      { :acl => "~ ^#{url_prefix}\/v3\/node\/([^\/]+)$", :method => :find, :allow => '$1', :authenticated => true },
+      { :acl => "~ ^#{master_url_prefix}\/v3\/catalog\/([^\/]+)$", :method => :find, :allow => '$1', :authenticated => true },
+      { :acl => "~ ^#{master_url_prefix}\/v3\/node\/([^\/]+)$", :method => :find, :allow => '$1', :authenticated => true },
       # this one will allow all file access, and thus delegate
       # to fileserver.conf
-      { :acl => "#{url_prefix}/v3/file" },
-      { :acl => "#{url_prefix}/v3/certificate_revocation_list/ca", :method => :find, :authenticated => true },
-      { :acl => "~ ^#{url_prefix}\/v3\/report\/([^\/]+)$", :method => :save, :allow => '$1', :authenticated => true },
+      { :acl => "#{master_url_prefix}/v3/file" },
+      { :acl => "#{ca_url_prefix}/v1/certificate_revocation_list/ca", :method => :find, :authenticated => true },
+      { :acl => "~ ^#{master_url_prefix}\/v3\/report\/([^\/]+)$", :method => :save, :allow => '$1', :authenticated => true },
       # These allow `auth any`, because if you can do them anonymously you
       # should probably also be able to do them when trusted.
-      { :acl => "#{url_prefix}/v3/certificate/ca", :method => :find, :authenticated => :any },
-      { :acl => "#{url_prefix}/v3/certificate/", :method => :find, :authenticated => :any },
-      { :acl => "#{url_prefix}/v3/certificate_request", :method => [:find, :save], :authenticated => :any },
-      { :acl => "#{url_prefix}/v3/status", :method => [:find], :authenticated => true },
-      { :acl => "#{url_prefix}/v3/environments", :method => :find, :allow => '*', :authenticated => true },
+      { :acl => "#{ca_url_prefix}/v1/certificate/ca", :method => :find, :authenticated => :any },
+      { :acl => "#{ca_url_prefix}/v1/certificate/", :method => :find, :authenticated => :any },
+      { :acl => "#{ca_url_prefix}/v1/certificate_request", :method => [:find, :save], :authenticated => :any },
+      { :acl => "#{master_url_prefix}/v3/status", :method => [:find], :authenticated => true },
+      { :acl => "#{master_url_prefix}/v3/environments", :method => :find, :allow => '*', :authenticated => true },
       ]
       end
 
--- a/lib/puppet/network/authconfig.rb
+++ b/lib/puppet/network/authconfig.rb
@@ -16,24 +16,30 @@ module Puppet
 
     def self.default_acl
       [
-      # API V2.0
+      # Master API V2.0
       { :acl => "#{master_url_prefix}/v2.0/environments", :method => :find, :allow => '*', :authenticated => true },
 
-      # API V3
+      # Master API V3
+      { :acl => "#{master_url_prefix}/v3/environments", :method => :find, :allow => '*', :authenticated => true },
+
       { :acl => "~ ^#{master_url_prefix}\/v3\/catalog\/([^\/]+)$", :method => :find, :allow => '$1', :authenticated => true },
       { :acl => "~ ^#{master_url_prefix}\/v3\/node\/([^\/]+)$", :method => :find, :allow => '$1', :authenticated => true },
+      { :acl => "~ ^#{master_url_prefix}\/v3\/report\/([^\/]+)$", :method => :save, :allow => '$1', :authenticated => true },
+
       # this one will allow all file access, and thus delegate
       # to fileserver.conf
       { :acl => "#{master_url_prefix}/v3/file" },
+
+      { :acl => "#{master_url_prefix}/v3/status", :method => [:find], :authenticated => true },
+
+      # CA API V1
       { :acl => "#{ca_url_prefix}/v1/certificate_revocation_list/ca", :method => :find, :authenticated => true },
-      { :acl => "~ ^#{master_url_prefix}\/v3\/report\/([^\/]+)$", :method => :save, :allow => '$1', :authenticated => true },
+
       # These allow `auth any`, because if you can do them anonymously you
       # should probably also be able to do them when trusted.
       { :acl => "#{ca_url_prefix}/v1/certificate/ca", :method => :find, :authenticated => :any },
       { :acl => "#{ca_url_prefix}/v1/certificate/", :method => :find, :authenticated => :any },
       { :acl => "#{ca_url_prefix}/v1/certificate_request", :method => [:find, :save], :authenticated => :any },
-      { :acl => "#{master_url_prefix}/v3/status", :method => [:find], :authenticated => true },
-      { :acl => "#{master_url_prefix}/v3/environments", :method => :find, :allow => '*', :authenticated => true },
       ]
       end
 
--- a/lib/puppet/network/http/api.rb
+++ b/lib/puppet/network/http/api.rb
@@ -8,7 +8,7 @@ class Puppet::Network::HTTP::API
   end
 
   def self.master_routes
-    master_prefix = Regexp.new("^#{Puppet::Network::HTTP::MASTER_URL_PREFIX}")
+    master_prefix = Regexp.new("^#{Puppet::Network::HTTP::MASTER_URL_PREFIX}/")
     Puppet::Network::HTTP::Route.path(master_prefix).
       any.
       chain(Puppet::Network::HTTP::API::Master::V3.routes,
@@ -17,7 +17,7 @@ class Puppet::Network::HTTP::API
   end
 
   def self.ca_routes
-    ca_prefix = Regexp.new("^#{Puppet::Network::HTTP::CA_URL_PREFIX}")
+    ca_prefix = Regexp.new("^#{Puppet::Network::HTTP::CA_URL_PREFIX}/")
     Puppet::Network::HTTP::Route.path(ca_prefix).
       any.
       chain(Puppet::Network::HTTP::API::CA::V1.routes,
--- a/lib/puppet/network/http/api.rb
+++ b/lib/puppet/network/http/api.rb
@@ -6,4 +6,21 @@ class Puppet::Network::HTTP::API
         raise Puppet::Network::HTTP::Error::HTTPNotFoundError.new("No route for #{req.method} #{req.path}", Puppet::Network::HTTP::Issues::HANDLER_NOT_FOUND)
       end)
   end
+
+  def self.master_routes
+    master_prefix = Regexp.new("^#{Puppet::Network::HTTP::MASTER_URL_PREFIX}")
+    Puppet::Network::HTTP::Route.path(master_prefix).
+      any.
+      chain(Puppet::Network::HTTP::API::Master::V3.routes,
+            Puppet::Network::HTTP::API::Master::V2.routes,
+            Puppet::Network::HTTP::API.not_found)
+  end
+
+  def self.ca_routes
+    ca_prefix = Regexp.new("^#{Puppet::Network::HTTP::CA_URL_PREFIX}")
+    Puppet::Network::HTTP::Route.path(ca_prefix).
+      any.
+      chain(Puppet::Network::HTTP::API::CA::V1.routes,
+            Puppet::Network::HTTP::API.not_found)
+  end
 end
--- a/spec/unit/network/http/api_spec.rb
+++ b/spec/unit/network/http/api_spec.rb
@@ -1,6 +1,7 @@
 #! /usr/bin/env ruby
 
 require 'spec_helper'
+require 'puppet_spec/handler'
 require 'puppet/network/http'
 
 describe Puppet::Network::HTTP::API do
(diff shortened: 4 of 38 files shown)


rlinehan:
This is blocked by https://github.com/puppetlabs/puppet/pull/3444 and will need to be rebased once that goes in.

lib/puppet/network/http/api/v3/indirection_type.rb
```diff
@@ -0,0 +1,13 @@
+class Puppet::Network::HTTP::API::V3::IndirectionType
+
+  INDIRECTION_TYPE_MAP = {
+    "certificate" => :ca,
+    "certificate_request" => :ca,
+    "certificate_revocation_list" => :ca,
+    "certificate_status" => :ca
```
nwolfe: We need the plural of this too right? `"certificate_statuses"`

lib/puppet/network/http/api/v3/indirection_type.rb
```diff
@@ -0,0 +1,13 @@
+class Puppet::Network::HTTP::API::V3::IndirectionType
+
+  INDIRECTION_TYPE_MAP = {
+    "certificate" => :ca,
+    "certificate_request" => :ca,
```
nwolfe: We need the plural of this too right? `"certificate_requests"`

rlinehan: I don't think so, because when this is used (https://github.com/puppetlabs/puppet/pull/3464/files#diff-20dc5688c93b62d47f42bf0a835001ecR83 and https://github.com/puppetlabs/puppet/pull/3464/files#diff-20dc5688c93b62d47f42bf0a835001ecR237), the indirection name passed in is singular.

rlinehan:
Rebased on top of merged-to-master PUP-3645. This should be ready to go now.

lib/puppet/network/http/api/v3.rb
```diff
@@ -14,9 +14,13 @@ class Puppet::Network::HTTP::API::V3
...
-  def self.routes
-    Puppet::Network::HTTP::Route.path(%r{/v3}).
+  def self.master_routes
+    Puppet::Network::HTTP::Route.path(%r{v3}).
         any.
         chain(ENVIRONMENTS, INDIRECTED)
   end
+
+  def self.ca_routes
+    Puppet::Network::HTTP::Route.path(%r{v1}).any.chain(INDIRECTED)
```
nwolfe: I'm wondering if it's going to be problematic at all to have these coupled together in a file named after the version of the master's (but not the CA's) API.

Maybe I'm overthinking it but what happens if we need to bump CA to v2?  Would we make the change in this file and leave it as `http/api/v3`?

kylog:
A fix for the spec failure has been merged in. Kicking travis to confirm.

lib/puppet/network/http/api/v3.rb
rlinehan: That's a good question. I think it's sort of weird right now because both the CA prefix at v1 and the master prefix at v3 end up in the same place - `IndirectedRoutes`. Other than here, the only place I can think that makes sense to define these would be directly in `http/api.rb`.

If we were bumping the CA routes to v2, then presumably it would be because we were separating them out to no longer use `IndirectedRoutes`, in which case it seems like it would make more sense to define them in their own file.

lib/puppet/network/authconfig.rb
```diff
@@ -5,30 +5,34 @@ class ConfigurationError < Puppet::Error; end
...
-      { :acl => "#{url_prefix}/v3/certificate/ca", :method => :find, :authenticated => :any },
-      { :acl => "#{url_prefix}/v3/certificate/", :method => :find, :authenticated => :any },
-      { :acl => "#{url_prefix}/v3/certificate_request", :method => [:find, :save], :authenticated => :any },
-      { :acl => "#{url_prefix}/v3/status", :method => [:find], :authenticated => true },
-      { :acl => "#{url_prefix}/v3/environments", :method => :find, :allow => '*', :authenticated => true },
+      { :acl => "#{ca_url_prefix}/v1/certificate/ca", :method => :find, :authenticated => :any },
+      { :acl => "#{ca_url_prefix}/v1/certificate/", :method => :find, :authenticated => :any },
+      { :acl => "#{ca_url_prefix}/v1/certificate_request", :method => [:find, :save], :authenticated => :any },
+      { :acl => "#{master_url_prefix}/v3/status", :method => [:find], :authenticated => true },
+      { :acl => "#{master_url_prefix}/v3/environments", :method => :find, :allow => '*', :authenticated => true },
```
cprice404: Do you think it might be worth sorting these lines such that the CA ones are all grouped together, separate from the master ones?

@haus if we were to do that would it cause any more or fewer packaging/upgrade concerns?

lib/puppet/network/http/api.rb
```diff
@@ -6,4 +6,27 @@ def self.not_found
         raise Puppet::Network::HTTP::Error::HTTPNotFoundError.new("No route for #{req.method} #{req.path}", Puppet::Network::HTTP::Issues::HANDLER_NOT_FOUND)
       end)
   end
+
+  def self.master_routes
+    master_prefix = Puppet[:master_url_prefix]
+    if !master_prefix.end_with?("/")
```
cprice404: Do you think maybe we should do this in the setting hook?  Since we already have some slash manipulation going on there?  Or are there cases where we need that setting to not have a trailing slash?

lib/puppet/network/http/api/v3.rb
cprice404: This is going to sound ridiculous but I wonder if it wouldn't make sense to go ahead and have `api/master/v3.rb` and `api/ca/v1.rb` now.  It's ridiculous because those namespaces would only have like 3 lines of code in them, but I kind of think it might still be worth it because otherwise the `v3.rb` file doesn't seem like an intuitive place to look for the v1 CA routes... :(  I could survive with it like this but I think I might have a slight preference for going ahead and breaking it out now.

lib/puppet/network/http/api.rb
rlinehan: There are cases (such as when we insert it into the default ACLs, and in all the tests) where we don't want it to have a trailing slash. It wouldn't be that much work to set it in the setting hook and change all the other instances, but it was less work to do this.

lib/puppet/network/authconfig.rb
rlinehan: Sorting them such that CA ones are grouped together makes sense to me; I didn't know how much it mattered and figured I'd just leave it as is with the comment about `auth any`.

lib/puppet/defaults.rb
```diff
@@ -1010,6 +1010,15 @@ def self.default_diffargs
           end
         end
     },
+    :ca_url_prefix => {
+        :default => "/puppet-ca",
+        :desc    => "The prefix at which the puppet certificate authority API is mounted.",
+        :hook    => proc do |value|
+          if !value.start_with?("/")
+            Puppet[:ca_url_prefix] = "/#{value}"
```
rlinehan: A conversation with @nwolfe about the previous PR brought up the potential issue with this and the `master_url_prefix` that if a user specifies a prefix with a slash in it, they may end up with really confusing errors because `IndirectedRoutes` does a split on `/` (https://github.com/puppetlabs/puppet/pull/3464/files#diff-20dc5688c93b62d47f42bf0a835001ecR70). Would it make sense to raise an error for `ca_url_prefix` and `master_url_prefix` if a prefix contains a slash or other illegal characters?

lib/puppet/network/http/api/v3.rb
rlinehan: This makes sense to me... my only question is where should the `indirected_routes.rb` file live?

lib/puppet/network/authconfig.rb
haus: @cprice404 since we're just scrapping and replacing this file during 3 => 4 upgrades, i'm not concerned here. Future upgrades within the 4 series will go fine, with any changes being handled by the package manager and user.

lib/puppet/network/http/api/v3/indirected_routes.rb
```diff
@@ -50,21 +53,38 @@ def call(request, response)
...
+  def self.master_url_prefix
     "#{Puppet[:master_url_prefix]}/v3"
   end
 
+  def self.ca_url_prefix
+    "#{Puppet[:ca_url_prefix]}/v1"
+  end
+
+  def self.url_prefix_for_indirection(indirection_name)
+    IndirectionType.type_for(indirection_name) == :ca ? ca_url_prefix : master_url_prefix
```
cprice404: This is very nitpicky, but... :)

From this line of code it's not necessarily obvious what other values `type_for` might return besides :ca.  If it were changed in the future, this line might no longer be correct, but there wouldn't necessarily be any obvious indicator of that pitfall for the person who was updating the `type_for` method.

It might be better to have a case statement that falls through to a default behavior of throwing an error in case we get an unrecognized value.  Or you could provide a map that contained the mappings from the `type` keyword to the URL string, and then throw an error if `type_for` returns something that we don't have an entry for in the map.

Or, if we're going to try to separate the ca/v1 and master/v3 references so that they're not in the same file, maybe we could move the url mapping into `IndirectionType` as well?  And just add a `url_prefix_for(indirection_name)` method?

lib/puppet/network/http/api/v3/indirected_routes.rb
```diff
@@ -50,21 +53,38 @@ def call(request, response)
...
 
+    # this also depluralizes the indirection_name if it is a search
     method = indirection_method(http_method, indirection_name)
-    check_authorization(method, "#{self.class.url_prefix}/#{indirection_name}/#{key}", params)
+
+    # check whether this indirection matches the prefix and version in the
+    # request
+    if url_prefix != self.class.url_prefix_for_indirection(indirection_name)
+      raise ArgumentError, "Indirection '#{indirection_name}' does not match url prefix '#{url_prefix}'"
+    end
```
cprice404: Nice catch on this.

lib/puppet/network/http/api/v3/indirection_type.rb
cprice404: @camlow325 totally unrelated to this PR but I think this ^^ explains the plural/singular discrepancy that you were asking me about the other day.

cprice404:
I'm done with initial review; will pull the code down and play with it a bit.  Looking good, just a few minor organizational questions / comments.

lib/puppet/network/http/api/v3.rb
cprice404: I think that with the minor refactors we discussed at various points in the PR comments, we could end up in a state where IndirectedRoutes doesn't contain any references to any version strings.  If that's the case then we could put it in the `api` dir or even the parent dir of that, whichever seems like a better fit.

lib/puppet/network/authconfig.rb
cprice404: @rlinehan ah, that makes sense about grouping them by the comment / auth level.  If we do that we might as well change it so that for each one of the different comment sections, the ca routes are always at the top or the bottom of the individual section maybe?  I'm good w/whatever you think is best, not a big deal.

lib/puppet/network/http/api.rb
cprice404: OK, this seems fine then.

nwolfe:
:+1: verified with agent and CLI tools using both/neither/combinations of the prefixes.

spec/unit/network/http/api/v3/indirected_routes_spec.rb
```diff
@@ -198,7 +207,12 @@ def a_request_that_searches(key, request = {})
     end
 
     it "should include the environment in the query string of the URI" do
-      handler.class.request_to_uri(request).should == "#{url_prefix}/foo/with%20spaces?environment=myenv&foo=bar"
+      handler.class.request_to_uri(request).should == "#{master_url_prefix}/foo/with%20spaces?environment=myenv&foo=bar"
+    end
+
+    it "should include the correct url prefix if it is a ca request" do
+      request.stubs(:indirection_name).returns("certificate")
+      handler.class.request_to_uri(request).should == "#{ca_url_prefix}/certificate/with%20spaces?environment=myenv&foo=bar"
```
camlow325: Probably not a big deal but "environment" is invalid for "certificate", right?  But then again so is "foo" I guess and I suppose the validation isn't intended to work at the level of URI parameters anyway...

spec/unit/network/http/api/v3_spec.rb
```diff
@@ -4,34 +4,53 @@
...
+    master_routes.process(request, response)
+
+    expect(response.code).to eq(200)
+  end
+
+  it "mounts ca routes" do
+    Puppet::SSL::Certificate.indirection.stubs(:find).returns "foo"
+    request = Puppet::Network::HTTP::Request.
+        from_hash(:path => "#{ca_url_prefix}/certificate/foo",
+                  :params => {:environment => "production"},
```
camlow325: Same question here as above - `:environment` is probably irrelevant here, right?

spec/unit/network/http/api_spec.rb
```diff
@@ -8,30 +9,107 @@ def respond(text)
...
+        expect(res[:status]).to eq(404)
+        expect(JSON.parse(res[:body])['issue_kind']).to eq("HANDLER_NOT_FOUND")
+      end
+    end
+
+    describe "when processing CA routes" do
+      it "responds to v1 indirector requests" do
+        Puppet::SSL::Certificate.indirection.stubs(:find).returns "foo"
+        req = Puppet::Network::HTTP::Request.from_hash(:path => "#{ca_prefix}/v1/certificate/foo",
+                                                       :params => {:environment => "production"},
```
camlow325: And just to be pedantic, same question here on `:environment`.  :smile:

spec/unit/network/http/api/v3/indirected_routes_spec.rb
cprice404: No, environment is required for all indirector requests.  Even the ones that don't use it for anything.  (Spoiler alert: we should not carry that over to v2 of the CA API :) )

camlow325: Does this mean that external HTTP API callers to the CA v1 certificate* endpoints will be required to include "?environment=<whatever>" or, if it is absent at that level, will some default be substituted to make the indirector happy?

cprice404: Requests originating from the agent will include the environment query parameter, because all agent URLs are constructed via the indirector.  For the Rack/Webrick masters, the requests will fail if this parameter is not provided.

For the Puppet Server CA endpoints, I suspect that it would be acceptable to ignore this query parameter rather than returning a failure response, but that is a question for product, and is orthogonal to this PR.

We should not expect the agent to stop sending this parameter until there is a significant rework of the client-side code that involves removing the indirector from the process of constructing the URLs.

camlow325: That's fine.  I don't see a problem there, mostly curious about what the expected behavior would be for CA v1.  I agree it is orthogonal to this PR, but I agree with you that having Puppet Server's CA just "ignore" the environment for the CA v1 endpoints seems reasonable.

joshcooper: :+1: to ignoring the parameter

rlinehan:
@cprice404 added some commits based on your comments.

cprice404:
+1 from me.

I'm going to pull the code down and play with it a bit, but unless anything jumps out at me from that, I'll plan on merging this on Friday or so.  ping @joshcooper @hlindberg @kylog in case you guys have any interest in further review before we merge.

rlinehan:
Should I squash some of these commits, or is it fine as is?

cprice404:
@rlinehan maybe let's wait and squash on Friday before we merge?

rlinehan:
@cprice404 sounds good.

cprice404:
I pulled the latest code down and attacked it with curl for a while; lgtm

joshcooper:
@rlinehan @cprice404 Thinking out loud here... why do we need a prefix? As in, doesn't trapperkeeper know when a `catalog` request comes in, it should be dispatched to the puppetmaster, etc? Why do we need the additional prefix and the added configuration?

joshcooper:
@rlinehan @cprice404 ok, I read the ticket, that helps, but I am still surprised that we're exposing this data to the agent as it seems like an implementation detail for **how** to retrieve something. For example, suppose the agent makes a request for a node information, and with this PR, it'd be under the `puppet` prefix. But in a future release, we move that to a different trapperkeeper app. Now we need to update all REST clients...

hlindberg:
@rlinehan @cprice404 @joshcooper Agree with Josh - the indirection is better done on the server side - no changed URLs on agents. If we want to also be able to redirect agents wouldn't using HTTP 30x responses be the way to do that? (But that would be for other reasons than the original "do not squat on the server's '/' namespace")

camlow325:
My understanding has been that we're trying to move to a model where each service has a mount point that can be set via configuration -- not requiring code changes -- like what is afforded by  https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/webrouting-service.md.  We could treat legacy Puppet as a special case that avoids this pattern for the sake of maintaining shorter-term compatibility (along with not addressing the aforementioned problems around the top-level environment in URL paths and squatting on the server's '/' namespace).  This wouldn't seem like a good long-term solution, though, and essentially what I had thought had led to this whole set of work around redoing the URL paths for Puppet 4.0.

One of the major benefits that I see to having the top-level service mount point be configurable both from the server and client side is the increased level of control given to users.  For example, I could see where  some users might want to take advantage of the configurable prefix as part of their load-balancing scheme.  For example, users might want to include a top-level "region" path in the mount point which is set for agents as appropriate for the region in which they reside, e.g., some would make requests to "/americas/puppet/..." whereas others would make requests to "/europe/puppet/...", and the load-balancer upstream would use that context from the prefix to redirect to the appropriate masters.  In this model, it would be possible for the `master_url_prefix` on agents to be different as compared to the master.

Hardcoding the mount points into the implementation with no external configurability wouldn't seem to be progressing the implementation forward.

One could ask the question about why we're not going further by making each of the endpoints in legacy Puppet independently configurable - e.g., by codifying a `catalog_url_prefix`, `report_url_prefix`, and others in anticipation that these may become standalone services at some point.  This would probably be premature, though, if the strategy around further decomposing Puppet into more discrete services hasn't been worked out.  Although it may imply that further service decomposition in the future would necessitate a new round of breaking configuration changes for the server and clients.

In short, I'm good with the overall direction of the implementation captured in this current set of PRs.

cprice404:
@joshcooper @hlindberg fair questions.  Here's what I'd say:

* We can probably make a guarantee that the default values for these prefixes will never change from the values we're populating them with now.

* I would expect that in 99.99% of cases a user would not modify these settings.

* On the Puppet Server side, all web apps will always be mounted at a configurable prefix, so that we have maximum flexibility for combining web apps in a single web server without risking URL collisions.  Therefore, it seems like it would be short-sighted to not at least allow the configuration to be made on the client.  I can imagine a future world where we support something like mounting two versions of puppet on a single server to help make it possible for people to upgrade their agents to the latest puppet version incrementally, in which case someone might choose to mount the different versions at, e.g., `/puppet4` and `/puppet5`.  The agents would need to be able to be configured to hit the correct web apps in a case like that.  (It's also possible that for that specific use case, we could just have a single `/puppet` app on the server and have it use an HTTP header from the agent to deal with routing the request accordingly, so I'm just using this as a general example.)

In short, I think that making the prefixes configurable on the agent gives us the maximum amount of flexibility going forward, even if these settings end up almost never being used.

I'd be willing to entertain a conversation about hard-coding the prefixes on the agent, but my current opinion is that the tradeoff could theoretically paint us into a corner in the future that I feel that this solution avoids.

cprice404:
@joshcooper also, since we will now have versioned URLs, we can make sure that in the future we still support things like `/puppet/v3/catalog` (via a redirect or whatever other mechanism we choose), even if the canonical endpoint has moved to `/puppet-compiler/v1/catalog`, etc.

joshcooper:
I think we're all in violent agreement that the server should allow REST endpoints to be mapped to services in configuration as outlined in https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/webrouting-service.md

I'm not 100% sold on the argument that the load-balancer might segregate traffic to geographically different locations based on a URL prefix. I think more likely is that clients are configured to talk to their "local" master either by changing the `server` property and/or using SRV records.

If we were to allow REST endpoints to be configured on the client, then I'm not a fan of having a prefix setting, as it seems very specific to our implementation of how REST URLs are constructed. I'd prefer having an explicit URL, e.g. `catalog_url=puppet:///puppet/v3/catalog`. That way everything about the path is configurable. Also, it makes it possible for the client to use a "newer" versioned endpoint without changing code.

cprice404:
If there were a newer version of an endpoint available, it would probably be because there were API changes, in which case the client code would most likely need to be modified anyway.  That also seems like a significantly more invasive change to me, and seems like it would jeopardize the Puppet 4.0 target dates.

Given that what we're roughly shooting for here is a way to split out the monolithic HTTP API into more service-specific namespaces, the prefix approach seems like a fairly simple / standard way to handle it... perhaps we should schedule a meeting of interested stakeholders to hash this out, though... it seems like it's going to be challenging to come to a consensus over github comment threads...

cprice404:
Also, just to clarify, I don't necessarily think that the goal is "to allow REST endpoints to be configured on the client".  It's more like "allowing REST endpoints to be namespaced".  I know that's not necessarily a black-and-white distinction, but I'm just saying that I'm not advocating for 100% configurability of the full URL... I think that the majority of the URL construction should be hard-coded for a given client version.

cprice404:
One other random thought... in the world of today, we are using these settings to determine the URL namespaces for the Rack/Webrick server implementations, as well as for the client.  So, even if we decided to change how we expose this stuff for the client, we will still need these for the server (unless we decide to just hard-code them).

cprice404:
we just had a meeting to discuss this comment thread.  I'm going to update the jira ticket with my interpretation of the outcome of that meeting if anyone is interested, but the tl;dr is that we're basically OK with this going in as-is.  I'll leave it open for another day or so in case anyone wants to raise any last-minute objections, but, failing that, we'll merge it.

rlinehan:
Per continued discussion on PUP-3526, we decided that it wasn't worth the effort to support settings for the master and ca url prefixes, and it would be better to remove them. I've updated this so that he url prefixes are now constants. I've also squashed some of the commits together, so that this is a bit cleaner. It should be ready to go.

cprice404:
@joshcooper @kylog @hlindberg I'm going to go ahead and get this merged in because it's blocking some things.  Happy to revisit any of the details if any questions come up.

acceptance/tests/environment/directory_environment_production_created_master.rb
```diff
@@ -30,7 +30,7 @@
 step 'run master; ensure production environment created'
 with_puppet_running_on(master, master_opts, testdir) do
   if master.is_using_passenger?
-    on(master, "curl -k --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} https://localhost:8140/v2.0/environments")
+    on(master, "curl -k --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} https://localhost:8140/puppet/v2.0/environments")
```
Jhesiah: "Curl -k --cert # {cert_path}  --key #{key_path }  --cacert #{cacert_path}  https://localhost:8140/puppet /v2.0/enviroments"

PR #3466, opened 2015-01-09
(PUP-1343) Use chkconfig --add to ensure K-symlinks get installed

ScottGarman:
On RHEL/CentOS 5 systems, the use of chkconfig on to enable a service
does not create the K-symlinks in /etc/rc.d to shut down the service.
Running chkconfig --add before chkconfig on fixes this issue, and
should be harmless on the unaffected OS versions.

The rspec unit test has been updated, and a new acceptance test
included to validate the bugfix.

### DIFF
--- a/lib/puppet/provider/service/redhat.rb
+++ b/lib/puppet/provider/service/redhat.rb
@@ -15,9 +15,11 @@ Puppet::Type.type(:service).provide :redhat, :parent => :init, :source => :init 
     # The off method operates on run levels 2,3,4 and 5 by default We ensure
     # all run levels are turned off because the reset method may turn on the
     # service in run levels 0, 1 and/or 6
-    output = chkconfig("--level", "0123456", @resource[:name], :off)
-  rescue Puppet::ExecutionFailure
-    raise Puppet::Error, "Could not disable #{self.name}: #{output}", $!.backtrace
+    # We're not using --del here because we want to disable the service only,
+    # and --del removes the service from chkconfig management
+    chkconfig("--level", "0123456", @resource[:name], :off)
+  rescue Puppet::ExecutionFailure => detail
+    raise Puppet::Error, "Could not disable #{self.name}: #{detail}", detail.backtrace
   end
 
   def enabled?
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/service/redhat.rb
```diff
@@ -39,6 +41,7 @@ def enabled?
   # Don't support them specifying runlevels; always use the runlevels
   # in the init scripts.
   def enable
+      chkconfig("--add", @resource[:name])
```
joshcooper: The commit message says "Running chkconfig --add before chkconfig on fixes this issue, and
should be harmless on the unaffected OS versions." However, we're skipping the acceptance test on `rhel7`, so do we know that it really is harmless?

For example, a user could explicitly specify the `redhat` provider (as `systemd` is the default on `rhel7`): 

<pre>
service { '...':
  ...
  provider => redhat
}
</pre>

And since we're skipping the acceptance test on `rhel7` we won't know if we've regressed. But perhaps using the `redhat` provider on `rhel7` doesn't make any sense?

ScottGarman: @joshcooper My note about the --add being harmless on unaffected OS versions refers to the fact that the bug exists on CentOS/RHEL 5 but not v6, where the acceptance test is also being run. I did give some thought to running it on v7 as well, but I haven't found a good service to use for testing that uses chkconfig and not systemd on v7. My search hasn't been 100% exhaustive, and I can look further into this on Monday.

ScottGarman: @joshcooper It looks like the network service is one of the only services left on RHEL 7 that you can control with chkconfig (and exists on v5 and v6 also). So I can refactor the acceptance test to use the network service instead of httpd (and be careful not to disable the service during the test!) :) Then we can run it on v5-v7.

ScottGarman: @joshcooper This has been refactored to use the netconsole service,and to force provider => redhat which was needed to ensure the redhat provider was used on RHEL v7. I think this should take care of the issues you've raised.

PR #3467, opened 2015-01-09
(PUP-3632) Update paths and defaults for the unified agent layout

whopper:
This commit updates Puppet's confdir, vardir, logdir and rundir to
adhere to the all in one agent layout. Unit and acceptance tests have also been
updated to reflect these changes.

The new paths (for Unix) are:
* confdir: /etc/puppetlabs/agent
* vardir: /opt/puppetlabs/agent/cache
* logdir: /var/log/puppetlabs/agent
* rundir: /var/run/puppetlabs

Windows paths remain unchanged.


whopper:
@haus mentioned that we will also need to figure out how we want to deal with Puppet's logdir and rundir, which with the way they are configured now will become: `/opt/puppetlabs/agent/run` and `/opt/puppetlabs/agent/log` by default. These aren't the correct directories according to the layout, but it seems that the default has always been wrong, and we just override them by default in `puppet.conf`. Do we want to keep doing this or make the defaults sane?

kylog:
>These aren't the correct directories according to the AIO layout, but it seems that the default has always been wrong, and we just override them in puppet.conf. For example, right now Puppet's default rundir is /var/lib/puppet/run, but we change that to /var/run/puppet in puppet.conf.
>
>Do we want to keep doing this or make the defaults sane?

Ah good catch @whopper. Let's make the defaults sane. Now is the perfect time.

whopper:
@kylog right on - will do and update PR!

whopper:
Update: currently this PR makes the following path changes for Unix:

confdir: `/etc/puppet` becomes `/etc/puppetlabs/agent`
vardir: `/var/lib/puppet` becomes `/opt/puppetlabs/agent/cache`
rundir: `/var/lib/puppet/run` becomes `/var/lib/puppetlabs/agent/run`
logdir: `var/lib/puppet/log` becomes `/var/log/puppetlabs/agent`

The only change I'm not totally sure about is the `rundir`, as it wasn't specified in the list of new AIO paths. Does `/var/lib/puppetlabs/agent/run` seem right?

Also, I'm currently trying to figure out how to make `install.rb` create the correct directories so acceptance  passes. Does anyone have any advice on that?

install.rb
```diff
@@ -299,11 +299,17 @@ def prepare_installation
   bindir = join(destdir, bindir)
   mandir = join(destdir, mandir)
   sitelibdir = join(destdir, sitelibdir)
+  vardir = "/opt/puppetlabs/agent/cache"
+  rundir = "/var/lib/puppetlabs/agent/run"
+  logdir = "/var/log/puppetlabs/agent"
```
whopper: Note: this can't be the right way to do this, but I'm doing it now for testing purposes...

haus:
`/var/lib/puppetlabs/agent` doesn't seem right. It should be `/var/run/puppetlabs/agent`.

For install.rb, you can throw flags like `--configdir`, `--mandir`, `--bindir`. Logdir won't be created on install.

whopper:
As an update: I've still got some acceptance failures that I'm trying to work through on this.

install.rb
```diff
@@ -299,11 +299,17 @@ def prepare_installation
...
+  rundir = "/var/run/puppetlabs/agent"
+  logdir = "/var/log/puppetlabs/agent"
 
   FileUtils.makedirs(configdir) if InstallOptions.configs
   FileUtils.makedirs(bindir)
   FileUtils.makedirs(mandir)
   FileUtils.makedirs(sitelibdir)
+  FileUtils.makedirs(vardir)
+  FileUtils.makedirs(rundir)
+  FileUtils.makedirs(logdir)
```
whopper: I'm not sure if this bit is the right way to get these directories. Since the new paths require us to create multiple sub-directories, it seems like it might be necessary.

haus: puppet makes some (all?) of these directories on startup.

whopper:
Haus suggested we may want to update the `puppet.conf`s we ship by removing each of them in ext and just symlinking them from a single empty `puppet.conf` in `conf`, since the default paths should now be sane. Should this new `puppet.conf` just be an empty file, or have any documentation?

haus:
@whopper when discussing this, it seemed arduous without an automated solution, so leaving it empty initially was the proposal.

install.rb
joshcooper: puppet will create any file/directory based setting in its defaults.rb, so even if it's not specified in puppet.conf. IOW, I don't think creating rundir or logdir is necessary.

whopper:
General update: in this PR's current state, I've got all of the acceptance tests passing! \o/. The only weirdness is the bit about creating the new var and conf dirs in install.rb, which I tried removing. That caused immediate failures due to the directories not existing:

    Test Case setup/git/pre-suite/020_PuppetUserAndGroup.rb reported: #    
    <Beaker::Host::CommandFailure: Host 'gluak8yaywikjh6.delivery.puppetlabs.net' exited with 1 running:
    puppet resource user puppet ensure=present
    ...
    Error: Could not run: Got 3 failure(s) while initializing: File[/opt/puppetlabs/agent/cache]: change from 
    absent to directory failed: Cannot create /opt/puppetlabs/agent/cache; parent directory 
    /opt/puppetlabs/agent does not exist; File[/var/log/puppetlabs/agent]: change from absent to directory 
    failed: Cannot create /var/log/puppetlabs/agent; parent directory /var/log/puppetlabs does not exist; 
    File[/var/run/puppetlabs/agent]: change from absent to directory failed: Cannot create 
    /var/run/puppetlabs/agent; parent directory /var/run/puppetlabs does not exist>

install.rb
haus: It seems to create the immediate directory, but not any parent directories. Not sure if that's a bug or not.

whopper:
@haus and I discussed a possible solution for this, which is to add additional, configurable settings for `vardir`, `rundir` and `logdir` into `install.rb`. I've updated the PR with this implementation, and am running tests again to verify all is still well.

acceptance/setup/git/pre-suite/030_PuppetMasterSanity.rb
```diff
@@ -1,6 +1,6 @@
 test_name "Puppet Master sanity checks: PID file and SSL dir creation"
 
-pidfile = '/var/lib/puppet/run/master.pid'
+pidfile = '/var/run/puppetlabs/agent/master.pid'
```
joshcooper: Why is the master's pid file under the `agent` directory?

install.rb
joshcooper: Ah I was still thinking `rundir` and `logdir` were still under `vardir`, but of course they're not anymore. And that's correct that puppet won't create parent directories, so :+1:

acceptance/setup/git/pre-suite/030_PuppetMasterSanity.rb
whopper: @joshcooper doh! This should be under just /var/run/puppetlabs, right?

joshcooper:
I'm sure this has been discussed, but I'm curious why `vardir` isn't in `/var` anymore? I assume puppet has been doing the wrong thing for a long time now. But that raises a second question of whether we should continue calling it `vardir`?

acceptance/setup/git/pre-suite/030_PuppetMasterSanity.rb
kylog: Fair point - the naming is confusing. Glad we won't be living with this in puppet 5 :)

https://confluence.puppetlabs.com/display/DEL/Unified+Puppet+Agent+Layout only specifies a single pidir, but if the naming will be too confusing perhaps we should either: a) have 2 directories, 1 for master, 1 for agent? or b) have one but at /var/run/puppetlabs?

/cc @haus

kylog:
> I'm sure this has been discussed, but I'm curious why vardir isn't in /var anymore? 

IIRC @stahnma said that many people want an enterprise app nicely segregated into /opt (which might be symlinked to /var on some systems).

>  But that raises a second question of whether we should continue calling it vardir?

Hmm, interesting. "cachedir" ? Naming is fun!

acceptance/setup/git/pre-suite/030_PuppetMasterSanity.rb
joshcooper: ah, so this is just the webrick pid file? what about `/var/run/puppetlabs/agent/puppet.pid` for the agent and `/var/run/puppetlabs/master/puppet.pid` for the master?

whopper: So we'll split the `pidfile` in defaults.rb to `agent_pidfile` and `master_pidfile`? I also just realized that we'll need to update every init script in ext as well.

whopper: Also, I'm a bit unfamiliar with Puppet's pidfile management, so how would we distinguish between using the master or agent pidfile? We seem to only currently define one path without distinguishing whether it's a master running or an agent run, and we use that single path in all of our init scripts.

joshcooper: Also puppet defaults the pidfile based on run_mode, which is horrible because it requires that run-mode be set before settings are initialized:

<pre>
    :default  => "$rundir/${run_mode}.pid",
</pre>

whopper: Ah right. So we could change `rundir` to `/var/run/puppetlabs` and just set the default path for the pidfile to `$rundir/${run_mode}/${run_mode}.pid`? We'd have to also be sure that install.rb creates both `/var/run/puppetlabs/agent` and `/var/run/puppetlabs/master`, and that AIO packaging updates reflect this.

Thoughts?

whopper:
As per a quick discussion with haus, I've moved `rundir` to `/var/run/puppetlabs` for now.

kylog:
This may need a rebase after the rspec 3 changes.

cprice404:
@nwolfe: @kylog said he mentioned this to you; it sounds like your PR needs to be based on this branch?

whopper:
Rebased on the rspec changes from yesterday. I'm in class all day today, but I'll try to make sure everything is good to go with this.

kylog:
Will this break our current acceptance pipeline?

whopper:
@kylog I ran acceptance against this in Linux and Windows and everything was passing, but that was before the last few changes, so I will definitely run again tomorrow. (Unless you meant break it in a different way?)

kylog:
@whopper thanks! Was that a test against packages or against source?

It may be that it won't pass with current packages (which would be fine) and that this can only be merged when we have AIO pipelines as the definitive CI pipeline - that's basically the issue I was trying to explore.

/cc @haus

whopper:
@kylog aha! Yeah, passing with source. Unless the packaging has been updated it will definitely fail, since the source is only passing with all of the changes to `install.rb that specifically update the paths.

lib/puppet/util/run_mode.rb
```diff
@@ -55,11 +55,19 @@ def which_dir( system, user )
...
-        which_dir("/var/lib/puppet", "~/.puppet/var")
+        which_dir("/opt/puppetlabs/agent/cache", "~/.puppet/var")
+      end
+
+      def run_dir
+        which_dir("/var/run/puppetlabs", "~/.puppet/var/run")
+      end
+
+      def log_dir
+        which_dir("/var/log/puppetlabs/agent", "~/.puppet/var/log/")
```
nwolfe: These changes mean that I now have to specify --logdir and --rundir when starting the master because they no longer properly land under --vardir since this UnixRunMode subclass overrides the parent default value of $vardir/{log,run}.

Is this behavior desirable?

I see this topic has come up a bit already on this PR...

spec/unit/util/run_mode_spec.rb
```diff
@@ -10,14 +10,18 @@
     expect(@run_mode.run_dir).to eq('$vardir/run')
   end
 
+  it "has logdir depend on vardir" do
+    expect(@run_mode.log_dir).to eq('$vardir/log')
```
nwolfe: Is this accurate?  Judging from some of the other comments about rundir and logdir it sounds like we no longer expect this to be the case.  The spec is testing the implementation of RunMode#log_dir, but at runtime either UnixRunMode or WindowsRunMode is used, and they both override the implementation of RunMode#log_dir so it's not under vardir.

Maybe I'm too deep in the weeds and confusing myself, but it feels like we have some inconsistencies?

lib/puppet/util/run_mode.rb
whopper: @nwolfe the AIO path changes have `logdir` and `rundir` no longer falling under `vardir`, with the specified directories being the new defaults. Is there a reason you don't want to use the new defaults? In the case of `logdir` being `/var/log/puppetlabs/agent`, from what I've heard we know it's a bit odd, but it's still the decided path for AIO.

spec/unit/util/run_mode_spec.rb
whopper: @nwolfe `WindowsRunMode` still uses the base implementation of `RunMode#log_dir` and `RunMode#run_dir`:

    class WindowsRunMode < RunMode
      def conf_dir
        which_dir(File.join(windows_common_base("etc")), "~/.puppet")
      end

      def var_dir
        which_dir(File.join(windows_common_base("var")), "~/.puppet/var")
      end
      ...
    end

nwolfe: You're right about Windows, but my question still applies to UnixRunMode

nwolfe: Based on your response to my other comment...this test is wrong, isn't it?

ext/redhat/puppet.conf
```diff
@@ -0,0 +1 @@
+puppet.conf
```
haus: So for the master, an empty puppet.conf won't work currently. I ran into this with @rlinehan. The puppet master process will run as the puppet user to reduce privilege, which means that the defaults it will see are ~/.puppet/var and friends.
I think it probably makes sense to set vardir, logdir and rundir in puppet.conf for the master run mode. This seems less than ideal because really the master shouldn't need config overriding to run properly, but the only alternative I can think of is to update the which_dir method to do something special if running in the master run mode and the puppet user (as masters will be), which seems weird as well.

lib/puppet/util/run_mode.rb
whopper: Talked with @nwolfe and we discovered that the problem now is that if you specify `--vardir` as an argument, you now also need to specifiy `rundir` and `logdir`, as they no longer fall under `vardir` dynamically.

nwolfe: The `logdir` and `rundir` fall under `vardir` if I don't specify `vardir` (when starting the master as user nwolfe via `puppet master --no-daemonize ...`), but only because they happen to be hard-coded under `~/.puppet/var`.

It'd be nice if we could interpolate these to be under $vardir.  This would only apply when we're non-root since the paths *don't* follow this structure when run as root.

nwolfe: Played around with this locally a bit and I'm having success with the following:
```ruby
which_dir("/var/log/puppetlabs/agent", Puppet[:vardir] + "/log")
```
Spec tests pass except for the run_mode_spec which would be updated accordingly.

lib/puppet/util/run_mode.rb
```diff
@@ -55,11 +55,22 @@ def which_dir( system, user )
...
       def conf_dir
-        which_dir("/etc/puppet", "~/.puppet")
+        which_dir("/etc/puppetlabs/agent", "~/.puppet")
       end
 
       def var_dir
-        which_dir("/var/lib/puppet", "~/.puppet/var")
+        # If Puppet is run as a non-root user and vardir is specified via the commandline,
+        # we'll need to use this user defined value when creating paths for rundir and logdir
+        user_vardir = if Puppet[:vardir] then Puppet[:vardir] else "~/.puppet/var" end
```
whopper: This is the solution @nwolfe and I came up with to handle a user running with the `--vardir` option. Since `rundir` and `logdir` are no longer set to `$vardir/run` or `$vardir/log`, we have to make sure to dynamically set them to whatever was passed for `vardir`. Does this look sane?

whopper:
@haus I went ahead and put back `vardir`, `rundir` and `logdir` into our new default `puppet.conf` to match how we were doing it before. You're right though.. depending on a conf file for the master to work seems a bit goofy.

lib/puppet/util/run_mode.rb
haus: @whopper Style-wise, this is weird. Why not `user_vardir = Puppet[:vardir] || "~/.puppet/var"` or `user_vardir = Puppet[:vardir] ? Puppet[:vardir] : "~/.puppet/var"`? Not a blocker, just weird.

PR #3468, opened 2015-01-10
(PUP-3829) Allow pip provider to work when osfamily is 'RedHat'

ghoneycutt:
Without this patch, the pip package provider is broken on EL systems.


ScottGarman:
Thanks @ghoneycutt for the submission, we appreciate it. However these changes don't take into account that the pip changes have only happened for EL v7 systems, and remain the same for previous versions. 

I am currently testing another pull request to fix this issue and expect we'll end up merging this one:

https://github.com/puppetlabs/puppet/pull/3470

PR #3357, opened 2014-11-29
(PUP-2531) add epp validate and render face

hlindberg:
This adds a new command called puppet-epp that can validate
one or several epp templates, render them to text with
various options for specifying facts and variables in scope,
giving variables to the template etc.

The command also has a dump action that is useful for debugging
epp parsing. This command is similar to the puppet parser dump
action.

For more details run 'puppet man epp' on this branch.

PR #3470, opened 2015-01-12
(PUP-3829) pip package provider : on RHEL7, with latest python-pip >= 1.5 from EPEL...

gillesgagniard:
..., 'pip-python' is no more and the command to use is back to 'pip' !

Currently, on an up-to-date RHEL7 or CentOS 7 install, it is not possible anymore to use the pip package provider : it fails with a "Could not locate the pip command." error.

This issue happens since 2014-12-18, when the python-pip package on EPEL has been updated to 1.5.6.

### DIFF
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -36,11 +36,10 @@ Puppet::Type.type(:package).provide :pip,
   end
 
   def self.cmd
-    case Facter.value(:osfamily)
-      when "RedHat"
-        "pip-python"
-      else
-        "pip"
+    if Facter.value(:osfamily) == "RedHat" and Facter.value(:operatingsystemmajrelease).to_i < 7
+      "pip-python"
+    else
+      "pip"
     end
   end
 
--- a/spec/unit/provider/package/pip_spec.rb
+++ b/spec/unit/provider/package/pip_spec.rb
@@ -2,7 +2,7 @@
 require 'spec_helper'
 
 provider_class = Puppet::Type.type(:package).provider(:pip)
-osfamilies = { 'RedHat' => 'pip-python', 'Not RedHat' => 'pip' }
+osfamilies = { ['RedHat', '6'] => 'pip-python', ['RedHat', '7'] => 'pip', ['Not RedHat', nil] => 'pip' }
 
 describe provider_class do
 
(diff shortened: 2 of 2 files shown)


gillesgagniard:
The unit test failure reported by CI is about the mount provider (and only with ruby 1.9.3), which I haven't touched with this PR ...

ScottGarman:
Thanks @gillesgagniard for this contribution. I'm going to run some internal tests against this first, but a visual inspection of this looks good so far.

kylog:
Worth noting that #3468 addresses the same issue (I think).

Also, please make sure your commit message follows our commit guidelines: first line should reference a ticket and be in imperative form. E.g.

"(PUP-1234) Fix broken assumption about pip's name on rhel7"

ScottGarman:
Passes my testing - :+1: if the commit message is updated to reflect @kylog's request.

MikaelSmith:
PR Triage Notes:
* We're mortally afraid of pip and python paths on RHEL.
* We suggest changing self.cmd to try to find pip, and if not found try pip-python (because Fedora).
* The Travis failures are likely due to calling a new Facter fact with the spec/integration/provider/mount_spec not properly stubbed out.

lib/puppet/provider/package/pip.rb
```diff
@@ -36,11 +36,10 @@ def self.instances
   end
 
   def self.cmd
-    case Facter.value(:osfamily)
-      when "RedHat"
-        "pip-python"
-      else
-        "pip"
+    if Facter.value(:osfamily) == "RedHat" and Facter.value(:operatingsystemmajrelease).to_i < 7
+      "pip-python"
```
MikaelSmith: This doesn't necessarily cover Fedora. A more robust solution would be to try to find pip (which), and if not found use pip-python.

kylog:
Re the travis fail: I don't think it's this PR. We saw the same thing on #3357. But it's not all PRs so ... we're wondering if it's some travis funkiness.

kylog:
A fix for the spec failure has been merged in. Kicking travis to confirm.

gillesgagniard:
The commit message has been updated to match guidelines.

Also, this patch covers Fedora too, since both currently supported versions (20 and 21) both have the 'pip' command available, and they are covered by the test on version.

MikaelSmith:
Thanks for your contribution!

PR #3471, opened 2015-01-13
(PUP-3684) Remove deprecated settings and module methods

hlindberg:


lib/puppet/settings.rb
```diff
@@ -839,20 +779,20 @@ def set_by_cli?(param)
...
-  def set_value(param, value, type, options = {})
-    Puppet.deprecation_warning("Puppet.settings.set_value is deprecated. Use Puppet[]= instead.")
+  # Patches the value for a param in a section.
+  # This method is required to support the use case of unifying --dns-alt-names and
+  # --dns_alt_names in the certificate face. Ideally this should be cleaned up.
+  # See PUP-3684 for more information.
+  # For regular use of setting a value, the method `[]=` should be used.
+  # @api private
+  #
+  def patch_value(param, value, type, options = {})
```
thallgren: What is the purpose of the options parameter here?

hlindberg: excellent question, I just renamed the method and did not notice the default dead options. Will remove it.

PR #3472, opened 2015-01-14
(PUP-3838) Change warning console message to yellow

kylog:


MikaelSmith:
Does this get used by both client and server?

kylog:
Any application which calls Puppet.warning (and logs to the console) would use this path, so that include `puppet master`, `puppet agent`, `puppet apply`, etc.

PR #3473, opened 2015-01-14
(PUP-3665) Make tagging arguments case insensitive

thallgren:
Tags are stored internally as downcased strings. All objects passed
as tags to the tagged? method where converted to strings prior to
comparison but they were not downcased. As a result, a query for
tags containing uppercase letters would always fail.

This commit adds the needed downcasing of the tags parameter. It
also introduces a new raw_tagged? method that assumes that whatever
tags that are passed to it already have been converted to strings
and downcased. This to allow the CollectorTransformer to do this
conversion once and then ask multiple resources if they are tagged.

### DIFF
--- a/lib/puppet/parser/resource.rb
+++ b/lib/puppet/parser/resource.rb
@@ -181,15 +181,21 @@ class Puppet::Parser::Resource < Puppet::Resource
     copy_as_resource.to_ral
   end
 
-  # Is the receiver tagged with the given tags?
+  # Faster variant of the tagged method that does no conversion of its
+  # arguments. Instead it's assumed that the arguments already are
+  # downcased strings.
+  #
   # This match takes into account the tags that a resource will inherit from its container
   # but have not been set yet.
   # It does *not* take tags set via resource defaults as these will *never* be set on
   # the resource itself since all resources always have tags that are automatically
   # assigned.
   #
-  def tagged?(*tags)
-    super || ((scope_resource = scope.resource) && scope_resource != self && scope_resource.tagged?(tags))
+  # @param tag_array [Array[String]] list tags to look for
+  # @return [Boolean] true if this instance is tagged with at least one of the provided tags
+  #
+  def raw_tagged?(tag_array)
+    super || ((scope_resource = scope.resource) && !scope_resource.equal?(self) && scope_resource.raw_tagged?(tag_array))
   end
 
   private
--- a/lib/puppet/parser/resource.rb
+++ b/lib/puppet/parser/resource.rb
@@ -181,11 +181,12 @@ class Puppet::Parser::Resource < Puppet::Resource
     copy_as_resource.to_ral
   end
 
-  # Faster variant of the tagged method that does no conversion of its
-  # arguments. Instead it's assumed that the arguments already are
-  # downcased strings.
+  # Answers if this resource is tagged with at least one of the tags given in downcased string form.
   #
-  # This match takes into account the tags that a resource will inherit from its container
+  # The method is a faster variant of the tagged? method that does no conversion of its
+  # arguments.
+  #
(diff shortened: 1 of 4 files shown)


lib/puppet/parser/resource.rb
```diff
@@ -181,15 +181,21 @@ def to_ral
     copy_as_resource.to_ral
   end
 
-  # Is the receiver tagged with the given tags?
+  # Faster variant of the tagged method that does no conversion of its
```
hlindberg: We want the first line of the comment to be on the form "Returns...", "Answers...", "Computes...", "Adds, "... etc.

Suggest saying "Answers if this resource is tagged with one of the tags given in downcased string form.", and then adjust the rest of the documentation of this method accordingly.

lib/puppet/util/tagging.rb
```diff
@@ -47,9 +47,25 @@ def tag_if_valid(name)
     end
   end
 
-  # Is the receiver tagged with the given tags?
+  # Is the receiver tagged with at least one of the given tags?
```
hlindberg: When rephrasing it is worth to update the text to say "Answers if the receiver is tagged with at least one of the given tags".

lib/puppet/util/tagging.rb
```diff
@@ -47,9 +47,25 @@ def tag_if_valid(name)
...
+  #
+  # @param *tags [String] splat of tags to look for
+  # @return [Boolean] true if this instance is tagged with at least one of the provided tags
+  #
   def tagged?(*tags)
-    not ( self.tags & tags.flatten.collect { |t| t.to_s } ).empty?
+    raw_tagged?(tags.collect {|t| t.to_s.downcase})
+  end
+
+  # Faster variant of the tagged method that does no conversion of its
```
hlindberg: same as above

lib/puppet/util/tagging.rb
```diff
@@ -47,9 +47,25 @@ def tag_if_valid(name)
...
+  # arguments. Instead it's assumed that the arguments already are
+  # downcased strings.
+  #
+  # @param tag_array [Array] array of tags to look for
+  # @return [Boolean] true if this instance is tagged with at least one of the provided tags
+  #
+  def raw_tagged?(tag_array)
+    my_tags = self.tags
+    not tag_array.index { |t| my_tags.include?(t) }.nil?
```
hlindberg: we prefer the operators `!`, `&&` and `||` since the keyword alternatives have precedence that sometimes screws up the logic. I.e. change `not` to `!`.

spec/unit/util/tagging_spec.rb
```diff
@@ -88,6 +88,13 @@
       expect(tagger).to be_tagged("goodbye")
     end
 
+    it "downcases tag arguments" do
+      tagger.tag("hello")
+      tagger.tag("goodbye")
+      expect(tagger).to be_tagged(:HEllO)
+      expect(tagger).to be_tagged("GooDByE")
+    end
+
```
hlindberg: think test should cover what is expected if tagging is done with uc word as well

hlindberg:
one with review, only a couple of small things to fix before this can be merged

spec/unit/util/tagging_spec.rb
thallgren: I can add that but please note that the test that precedes this one tests the opposite case, i.e. that uc tags are downcased. Also in this test, the GooDByE contains uc characters. So what would an additional uc word test add?

hlindberg: ok, then it is not needed.

hlindberg:
This PR should target stable, not master since it is for 3.7.4

thallgren:
I guess that means creating a new PR? Or is it possible to change the target branch?

PR #3481, opened 2015-01-15
(PUP-3839) Fix calling gem on Windows

MikaelSmith:
Fixes calling the gem command on Windows, which requires that Ruby is
installed first and that it be called via cmd /c.


MikaelSmith:
I've run this through setting up acceptance runs with: win2012-rubyx64 and rhel7, both from git and packages (Windows packages doesn't work because we don't support install from packages this way).

PR #3482, opened 2015-01-15
(PUP-3853) Add cfacter to the msi build

melissa:
This commit will allow us to include a specific cfacter build in the
puppet-agent msi package. This update is dependent on updates to the
automation in puppet_for_the_win to be able to encorporate this
addition.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -22,6 +22,11 @@ build_msi:
   facter:
     ref: 'refs/tags/2.3.0'
     repo: 'git://github.com/puppetlabs/facter.git'
+  cfacter:
+    archive:
+      x86: 'cfacter-0.3.0-x86.zip'
+      x64: 'cfacter-0.3.0-x64.zip'
+    path: 'http://builds.puppetlabs.lan/cfacter/0.3.0/artifacts/windows'
   hiera:
     ref: 'refs/tags/1.3.4'
     repo: 'git://github.com/puppetlabs/hiera.git'


melissaanne:
This pull request is dependent on https://github.com/puppetlabs/puppet_for_the_win/pull/85

joshcooper:
It looks like this is 64-bit cfacter, which won't work on 32-bit Windows. Are we going to compile two versions of cfacter, and selectively include the appropriate build in `puppet_for_the_win`?

melissaanne:
Hmm, I hadn't thought about that, unfortunately. But yes, that's probably the goal here. I'll create two different cfacter builds for x64 and x84, and put them into the appropriate msi

ext/build_defaults.yaml
```diff
@@ -21,6 +21,9 @@ build_msi:
   facter:
     ref: 'refs/tags/2.3.0'
     repo: 'git://github.com/puppetlabs/facter.git'
+  cfacter:
+    ref: 'cfacter-553c5dd5eaeda6ee6c8591ab536e776390c58f8b.zip'
+    repo: 'http://builds.puppetlabs.lan/cfacter/553c5dd5eaeda6ee6c8591ab536e776390c58f8b/artifacts/windows/cfacter-553c5dd5eaeda6ee6c8591ab536e776390c58f8b.zip'
```
ferventcoder: repo is a pretty loosely defined term here. Have we considered naming these fields differently for binaries?

melissaanne: It's a lot easier with the template we're using to generate the job if we keep these consistent. It's definitely do-able, but it'll take a little more thought

ferventcoder: Yes, but the way we pull cfacter is different, and the others will start moving towards this. I don't want to one day have someone go "Why do we use ref and repo for our built assemblies?"

ferventcoder: If that makes sense. :)

melissaanne: Yes, that is very true. I'll look into what I can do to change this. I do like having the name and url separate, so I'll probably maintain that, and remove the package name from the end of the path

melissaanne: I also have to change this to have the distinction between x86 and x64 builds

ferventcoder: :+1:

ext/build_defaults.yaml
```diff
@@ -21,6 +21,11 @@ build_msi:
   facter:
     ref: 'refs/tags/2.3.0'
     repo: 'git://github.com/puppetlabs/facter.git'
+  cfacter:
+    archive:
+      x86: 'cfacter-0.2.0-377-g4fa6175-x86.zip'
+      x64: 'cfacter-0.2.0-377-g4fa6175-x64.zip'
+    path: 'http://builds.puppetlabs.lan/cfacter/4fa61758e3f4f2bf909da5eb5007f197c04707f5/artifacts/windows/'
```
melissa: @ferventcoder I changed these to `archive` and `path`. Thoughts?

ferventcoder: Love!

Iristyle: Do we expect there to be a lot of churn here... or is this a good version to pin to at the moment?

melissa: I'm expecting there'll be churn. I'd suggest waiting until the cfacter release next week CFACT-182

PR #3484, opened 2015-01-15
(PUP-3842) Include MCO 2.7.0 in the puppet msi

melissaanne:


ext/build_defaults.yaml
```diff
@@ -24,6 +24,9 @@ build_msi:
   hiera:
     ref: 'refs/tags/1.3.4'
     repo: 'git://github.com/puppetlabs/hiera.git'
+  mcollective:
+    ref: 'refs/tags/2.7.0'
+    repo: 'git@github.com:puppetlabs/marionette-collective.git'
```
joshcooper: should that be `git://github.com/...`?

melissaanne: It's a key based way of cloning, so it works, but you're right in that it should be `git://github.com/...`

melissaanne:
http://jenkins-release.delivery.puppetlabs.net/job/puppet-packaging-2015-01-15-16-26-55-88e69dc53b5ddfd62a5c3873cc20284406ffed34/ includes mco in the msi, and also has the msi rename updates

haus:
:+1:

ext/build_defaults.yaml
```diff
@@ -24,6 +24,9 @@ build_msi:
   hiera:
```
melissaanne: I can't comment on the line I want to comment on, but the puppet_for_the_win ref needs to be updated for this to not fail

melissaanne:
Closed while we get the rest of this in. Will reopen when all bits are ready

PR #3485, opened 2015-01-16
(PUP-3844) Rename the MSI

melissaanne:
This change is necessary because we are renaming the puppet installer
on windows. Since we want to maintain the ability to build older
versions of the puppet installer with the old name, this new var will
allow us to know when we expect the new name, and when we expect the old
name.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -14,6 +14,7 @@ yum_host: 'yum.puppetlabs.com'
 yum_repo_path: '/opt/repository/yum/'
 build_gem: TRUE
 build_dmg: TRUE
+msi_name: 'puppet-agent'
 build_msi:
   puppet_for_the_win:
     ref: '2476cd007b1b52fb1b9f22202b56218629626b5c'


melissaanne:
http://jenkins-release.delivery.puppetlabs.net/job/puppet-packaging-2015-01-15-16-26-55-88e69dc53b5ddfd62a5c3873cc20284406ffed34/ has this change, plus includes mco in the msi build

haus:
:+1:

melissaanne:
This shouldn't get merged until we've brought in the updates to puppet_for_the_win

ext/build_defaults.yaml
```diff
@@ -14,6 +14,7 @@ yum_host: 'yum.puppetlabs.com'
 yum_repo_path: '/opt/repository/yum/'
 build_gem: TRUE
 build_dmg: TRUE
+msi_name: 'puppet-agent'
 build_msi:
   puppet_for_the_win:
     ref: '2476cd007b1b52fb1b9f22202b56218629626b5c'
```
melissaanne: This ref needs to be updated to latest for this to not fail, just fyi, And I believe we want to wait until https://github.com/puppetlabs/puppet_for_the_win/pull/84 is merged for that.

melissaanne:
Closed while we get the rest of this in. Will reopen when all bits are ready

PR #3486, opened 2015-01-16
(PUP-2402) add sshkey spec tests

er0ck:
This commit adds tests for the sshkey type. They can be automated in
rspec at much lower cost than acceptance.
Convert John's earlier closed PR to spec tests. Add tests for the full
range of encryption types.  Add a negative test for unknown encryption
types.
Move the test to provider directory as it's testing provider
functionality.

This commit also fixes a deprecation warning for testing file modes.  It
used to use should, use expect instead.


spec/integration/type/sshkey_spec.rb
```diff
@@ -1,22 +0,0 @@
-#! /usr/bin/env ruby
```
MikaelSmith: Did this essentially do nothing?

er0ck: good question.  i don't even remember taking this out.

looks like 525 tests call ruby out of 752 spec tests.
rspec obviously calls the ruby interpreter.

MikaelSmith: I meant the whole file. You removed the 'type' test and added one in providers.

er0ck: hah.  sorry.  
this is what i meant by "Move the test to provider directory as it's testing provider
functionality."

there was just the single test in the type dir that checked that the sshkey target file was created with the correct mode.

peterhuene: :+1: on moving this test to provider.

PR #3487, opened 2015-01-16
(maint) Ensure tests can run in isolation

joshcooper:
Ensure tests can be run in isolation:

<pre>
$ find spec/{unit,integration} -type f -name "*_spec.rb" | while read spec; do echo $spec; bundle exec rspec $spec; done
</pre>


kylog:
Killer. Glad you tried this @joshcooper. I wonder if we should add this as a low-frequency (e.g. nightly, not per-commit) Jenkins job.

joshcooper:
:+1:

kylog:
https://tickets.puppetlabs.com/browse/QENG-1682. Not sure when we'll get to it but it'll be nice to have.

PR #3488, opened 2015-01-16
(PUP-3850) Run specs on fedora as root

joshcooper:


joshcooper:
<pre>
[root@dwz8lmi10zm5i7h ~]# yum install git gcc ruby-devel
[root@dwz8lmi10zm5i7h ~]# gem install bundler
[root@dwz8lmi10zm5i7h ~]# git clone git://github.com/puppetlabs/puppet
[root@dwz8lmi10zm5i7h ~]# cd puppet
[root@dwz8lmi10zm5i7h puppet]# bundle install --path .bundle
[root@dwz8lmi10zm5i7h puppet]# bundle exec rake parallel:spec
[root@dwz8lmi10zm5i7h puppet]# git remote add joshcooper git://gitub.com/joshcooper/puppet
[root@dwz8lmi10zm5i7h puppet]# git fetch joshcooper
[root@dwz8lmi10zm5i7h puppet]# git merge joshcooper/ticket/master/PUP-3850-fedora-specs
[root@dwz8lmi10zm5i7h puppet]# bundle exec parallel:spec
Finished in 1 minute 43.75 seconds
18884 examples, 0 failures, 43 pending
</pre>

kylog:
@ahpook take a look.

This is not a real mention. If this were a real mention, um, more thoughts would appear here.

spec/integration/util/windows/security_spec.rb
```diff
@@ -368,7 +368,7 @@ def grant_everyone_full_access(path)
         end
       end
 
-      describe "for an administrator", :if => Puppet.features.root? do
+      describe "for an administrator", :if => (Puppet.features.root? && Puppet.features.microsoft_windows?) do
```
MikaelSmith: Why is this necessary, shouldn't line 11 have it covered?

joshcooper: I know, it seems completely unnecessary, right? But rspec doesn't AND nested include/exclude filters, it ORs them. So the following `inner` example group will be executed even though the `outer `one is excluded.

<pre>
describe "outer", :if => false do
  describe "inner", :if => true do
    it "executes" do
      puts "hello"
    end
  end
end
</pre>

which when executed:

<pre>
$ bundle exec rspec spec/unit/crazy_spec.rb
Hello
.

Finished in 0.00035 seconds
1 example, 0 failures
</pre>

MikaelSmith: Crazy's right.

joshcooper:
@kylog I updated the useradd spec like we talked about https://github.com/joshcooper/puppet/commit/6969b4bff6a662452ea70c6cb1e22c51b627f988#diff-22c3e68afbbef4f5831d712199056791R54

MikaelSmith:
:+1:

PR #3489, opened 2015-01-16
Don't color numbers in class names

sathieu:
This makes the digit in 'include apache2' use the same color.

i.e.: digits are colored only when starting a word.


sathieu:
My patch is trivial, it can be merged without CLA.

kylog:
@sathieu can you prepend "(maint) " to the commit message. Our release scripts look through the commit messages for first lines starting with "(maint)" or "(PUP-1234)"-style beginning-of-lines, and croak if the commit message doesn't match.

sathieu:
@kylog done

kylog:
Thanks for the contribution @sathieu!

PR #3491, opened 2015-01-16
(PUP-3461) Make Variant assignable when contained types are assignable

thallgren:
This commit changes the assignable?(t1, t2) so that A Variant t2 is
assignable to t1 if all types contained in t2 are assignable to t1. A
test is also added to verify this behavior.

The commit also adds the method instance_of_PIntegerType(t, o) as
a performance improvement. There were several places where an integer
instance was checked against an existing PIntegerType. Creating an
interim type for that instance just to perform the check is expensive
and puts an unecessary load on the garbage collector.


hlindberg:
looks good, merging when travis says ok

PR #3492, opened 2015-01-16
(PUP-3801) HP-UX Trusted Computing password age fix

ScottGarman:
HP-UX Trusted Computing does not change the password age when doing an
update to the password hash using /usr/sam/lbin/usermod.sam. This change
corrects that behavior and sets the password age back to 0 days old
after it has been modified.

If the HP-UX box is not running Trusted Computing, it only changes the
password.

### DIFF
--- a/lib/puppet/provider/user/hpux.rb
+++ b/lib/puppet/provider/user/hpux.rb
@@ -1,7 +1,9 @@
 Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
   desc "User management for HP-UX. This provider uses the undocumented `-F`
     switch to HP-UX's special `usermod` binary to work around the fact that
-    its standard `usermod` cannot make changes while the user is logged in."
+    its standard `usermod` cannot make changes while the user is logged in.
+    New functionality provides for changing trusted computing passwords and
+    resetting password expirations under trusted computing"
 
   defaultfor :operatingsystem => "hp-ux"
   confine :operatingsystem => "hp-ux"
@@ -26,7 +28,19 @@ Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
   end
 
   def modifycmd(param,value)
-    super.insert(1,"-F")
+     cmd = super(param, value)
+     cmd << "-F"
+     if trusted then
+       # Append an additional command to reset the password age to 0
+       # until a workaround with expiry module can be found for trusted
+       # computing.
+       cmd << ";"
+       cmd << "/usr/lbin/modprpw"
+       cmd << "-v"
+       cmd << "-l"
+       cmd << "#{resource.name}"
+     end
+     cmd
   end
 
   def password
@@ -62,5 +76,20 @@ Puppet::Type.type(:user).provide :hpuxuseradd, :parent => :useradd do
     else
       return ent.passwd
     end
+  end
+
+  def trusted
+    # Check to see if the HP-UX box is running in trusted compute mode
+    # UID for root should always be 0
+    trusted_sys = exec_getprpw('root','-m uid')
+    if trusted_sys.chomp == "uid=0"
+      return true
+    else
+      return false
+    end
+  end
+
+  def exec_getprpw(user,opts)
+    Puppet::Util::Execution.execute("/usr/lbin/getprpw #{opts} #{user}", { :combine => true })
   end
 end
--- a/spec/unit/provider/user/hpux_spec.rb
+++ b/spec/unit/provider/user/hpux_spec.rb
@@ -34,6 +34,7 @@ describe provider_class, :unless => Puppet.features.microsoft_windows? do
     before :each do
       Etc.stubs(:getpwent).returns(pwent)
       Etc.stubs(:getpwnam).returns(pwent)
+      resource.stubs(:command).with(:modify).returns '/usr/sam/lbin/usermod.sam'
     end
 
     it "should have feature manages_passwords" do
@@ -47,6 +48,26 @@ describe provider_class, :unless => Puppet.features.microsoft_windows? do
 
     it "should return password entry if exists" do
       provider.password.must == "foopassword"
+    end
+  end
+
+  context "check for trusted computing" do
+    before :each do
+      provider.stubs(:command).with(:modify).returns '/usr/sam/lbin/usermod.sam'
+    end
+
+    it "should add modprpw to modifycmd if Trusted System" do
+      resource.stubs(:allowdupe?).returns true
+      provider.expects(:exec_getprpw).with('root','-m uid').returns('uid=0')
+      provider.expects(:execute).with(['/usr/sam/lbin/usermod.sam', '-u', 1000, '-o', 'testuser', '-F', ';', '/usr/lbin/modprpw', '-v', '-l', 'testuser'])
+      provider.uid = 1000
+    end
+
+    it "should not add modprpw if not Trusted System" do
+      resource.stubs(:allowdupe?).returns true
+      provider.expects(:exec_getprpw).with('root','-m uid').returns('System is not trusted')
+      provider.expects(:execute).with(['/usr/sam/lbin/usermod.sam', '-u', 1000, '-o', 'testuser', '-F'])
+      provider.uid = 1000
     end
   end
 end


lib/puppet/provider/user/hpux.rb
```diff
@@ -19,6 +21,7 @@
     value !~ /\s/
   end
 
+
```
MikaelSmith: I have a couple pedantic comments, starting with the unnecessary newline here.

lib/puppet/provider/user/hpux.rb
```diff
@@ -26,7 +29,20 @@ def deletecmd
   end
 
   def modifycmd(param,value)
-    super.insert(1,"-F")
+     cmd = super(param, value)
+     cmd << "-F"
+     trusted_sys = trusted
+     if(trusted_sys == :true ) then
```
MikaelSmith: Just squash these two lines to
`if trusted`

lib/puppet/provider/user/hpux.rb
```diff
@@ -26,7 +29,20 @@ def deletecmd
...
+       # Append on additional command to reset the password age to 0
+       # until a work around with expiry module can be found for trusted
+       # computing.
+        cmd << ";"
```
MikaelSmith: Use 2 spaces instead of 3. The comments and code should line up.

lib/puppet/provider/user/hpux.rb
```diff
@@ -63,4 +79,21 @@ def password
...
+    # Check to see if the HP-UX box is running in trusted compute mode
+    # UID for root should always be 0
+    trusted_sys = exec_getprpw('root','-m uid')
+    if trusted_sys.chomp == "uid=0"
+      return :true
+    else
+      return :false
+    end
+
+    nil
```
MikaelSmith: newline and nil are unnecessary here, never reached.

spec/unit/provider/user/hpux_spec.rb
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env ruby
 require 'spec_helper'
 require 'etc'
+require 'pp'
```
MikaelSmith: Pretty sure this module isn't used, it doesn't need to be added.

spec/unit/provider/user/hpux_spec.rb
```diff
@@ -13,7 +14,6 @@
     )
   end
   let(:provider) { resource.provider }
-
```
MikaelSmith: Unnecessary deletion.

lib/puppet/provider/user/hpux.rb
```diff
@@ -63,4 +79,21 @@ def password
...
+  end
+
+  def exec_getprpw(user,opts)
+    %x(/usr/lbin/getprpw #{opts} #{user} 2>&1)
```
haus: You *probably* want to use Puppet::Util::Execution.execute here instead.

spec/unit/provider/user/hpux_spec.rb
haus: This isn't actually used, except I presume in debugging test failures :smile:

spec/unit/provider/user/hpux_spec.rb
```diff
@@ -49,4 +50,23 @@
       provider.password.must == "foopassword"
     end
   end
+
+  context "Check for trusted computing" do
+  before :each do
```
haus: Whitespace nitpick:

Spacing should be something like...

```
  context....
    before...
    end

    it...
    end
    it...
    end
  end
```

spec/unit/provider/user/hpux_spec.rb
```diff
@@ -49,4 +50,23 @@
...
+    provider.stubs(:command).with(:modify).returns '/usr/sam/lbin/usermod.sam'
+  end
+
+  it "Should add modprpw to modifycmd if Trusted System" do
+    resource.stubs(:allowdupe?).returns true
+    provider.stubs(:exec_getprpw).with('root','-m uid').returns('uid=0')
+    provider.expects(:execute).with { |args| args.include?("/usr/lbin/modprpw") }
```
MikaelSmith: I'd be more comfortable with this asserting the whole command that will be executed.

ScottGarman:
Thanks for the review, I should be able to take care of these this afternoon.

spec/unit/provider/user/hpux_spec.rb
```diff
@@ -49,4 +50,23 @@
...
   end
+
+  context "Check for trusted computing" do
+  before :each do
+    provider.stubs(:command).with(:modify).returns '/usr/sam/lbin/usermod.sam'
+  end
+
+  it "Should add modprpw to modifycmd if Trusted System" do
+    resource.stubs(:allowdupe?).returns true
+    provider.stubs(:exec_getprpw).with('root','-m uid').returns('uid=0')
```
MikaelSmith: @haus pointed out that we should use provider.expects(:exec_getprpw)... instead of stubs, as it adds the expectation that the method will be called. It'll still stub the method.

ScottGarman:
@jjperry69 I'm having issues getting the the following tests to pass:

it "Should add modprpw to modifycmd if Trusted System" and
it "Should not add modprpw if not Trusted System"

I'll re-add the last two comments that also still need to be resolved.

spec/unit/provider/user/hpux_spec.rb
```diff
@@ -49,4 +50,23 @@
...
+    before :each do
+      provider.stubs(:command).with(:modify).returns '/usr/sam/lbin/usermod.sam'
+    end
+
+    it "Should add modprpw to modifycmd if Trusted System" do
+      resource.stubs(:allowdupe?).returns true
+      provider.stubs(:exec_getprpw).with('root','-m uid').returns('uid=0')
+      provider.expects(:execute).with { |args| args.include?("/usr/lbin/modprpw") }
+      provider.uid = 1000
+    end
```
ScottGarman: Mikael suggested to assert the entire command being executed, not just checking for an argument in the list.

jjperry69:
I wonder what may be going on here then. On the closed PR previously it tested fine on the Travis CI build.

jjperry69:
I have updated the spec file with the requested changes and committed them.  What is the process to have the Travis CI build retest the changes?

jjperry69:
I found the problem when I pulled Scott's version from his jjperry-fix fork.

The issue I noted was that in the def trusted, The code is returning :true versus just true.  When I made that change to hpux.rb, the tests passed without any additional modifications. 

The error seems to be generated in relation to the change to the modify command where I had 2 lines for checking trusted and it was compacted into "if trusted then".  It seems the :true was not resolving the same as "true" and thus causing incorrect results. 

Can you please make these changes to the hpux.rb in your jjperry-fix branch to see if that allows the tests to pass with the modified hpux.rb and hpux_rspec.rb please?

ScottGarman:
Thanks @jjperry69, that does resolve the problems I was seeing here. @MikaelSmith - I think this now addresses all the feedback raised so far.

jjperry69:
Great! Thanks for all of the assistance getting this ready to be merged into the code base.

lib/puppet/provider/user/hpux.rb
```diff
@@ -26,7 +28,19 @@ def deletecmd
   end
 
   def modifycmd(param,value)
-    super.insert(1,"-F")
+     cmd = super(param, value)
+     cmd << "-F"
+     if trusted then
+       # Append on additional command to reset the password age to 0
```
MikaelSmith: Just noticed the typo in 'on'.

MikaelSmith:
I don't see any other issues, so after the typo is addressed I'll merge it.

lib/puppet/provider/user/hpux.rb
jjperry69: I have looked at it five different times now. What is the typo?

haus: 'on' should be 'an'

ScottGarman:
@MikaelSmith - typo fixed.

kylog:
For consistency, I would lower-case the first work of those context and should descriptions. E.g. see


```
101]  puppet git:(ScottGarman-jjperry-fix) be rspec spec/unit/provider/user/hpux_spec.rb -fd

Puppet::Type::User::ProviderHpuxuseradd
  should add -F when modifying a user
  should add -F when deleting a user
  managing passwords
    should have feature manages_passwords
    should return nil if user does not exist
    should return password entry if exists
  Check for trusted computing
    Should add modprpw to modifycmd if Trusted System
    Should not add modprpw if not Trusted System
```

Note that those upper case lines.

ScottGarman:
@kylog - done

PR #3495, opened 2015-01-20
(PUP-3548) Allow 4x functions to call 3x functions

thallgren:
The call_function() method of the Puppet::Pops::Functions::Function
class did not make any attempt to find functions that had been
declared using the 3x function API. This commit adds the following:

Proper load and execute:
The call_function() method will now consult the 3x Functions class
to find (and optionally load) the function from there when it fails
to find it using the 4x loader.

Other improvements:
The initialize method in Puppet::Pops::Evaluator::Runtime3Support
was removed in favor of accessors that creates the needed visitors.
This was done to get rid of code in the EvaluatorImpl initializer
that otherwize would need to be copied into the ArgsConverter class
as well.

### DIFF
--- a/lib/puppet/pops/functions/function.rb
+++ b/lib/puppet/pops/functions/function.rb
@@ -47,18 +47,46 @@ class Puppet::Pops::Functions::Function
   # Allows the implementation of a function to call other functions by name. The callable functions
   # are those visible to the same loader that loaded this function (the calling function).
   #
+  # @param function_name [String] The name of the function
+  # @param args [Any] splat of arguments
+  # @return [Any] The result returned by the called function
+  #
   # @api public
   def call_function(function_name, *args)
-    if the_loader = loader
-      func = the_loader.load(:function, function_name)
-      if func
-        return func.call(closure_scope, *args)
-      end
-    end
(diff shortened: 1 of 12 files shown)


lib/puppet/pops/functions/function.rb
```diff
@@ -49,16 +49,24 @@ def call(scope, *args)
   #
   # @api public
   def call_function(function_name, *args)
```
hlindberg: should take scope as an argument

lib/puppet/pops/functions/function.rb
```diff
@@ -49,16 +49,24 @@ def call(scope, *args)
...
-      if func
-        return func.call(closure_scope, *args)
-      end
-    end
-    # Raise a generic error to allow upper layers to fill in the details about where in a puppet manifest this
-    # error originates. (Such information is not available here).
-    #
-    raise ArgumentError, "Function #{self.class.name}(): cannot call function '#{function_name}' - not found"
+    the_loader = loader
+    scope = closure_scope
```
hlindberg: should use the given scope, (or not try if scope is nil)

PR #3497, opened 2015-01-20
(maint) Stub execute when testing debug logging

MikaelSmith:
Execute was trying to call 'echo hello', which is invalid on Windows. Stub
execution subcalls so no process is called, making it platform independent.


MikaelSmith:
Manually tested specs on Windows. Fixes https://jenkins.puppetlabs.com/view/Puppet/view/Master/job/platform_puppet_specs_windows-master/734/.

MikaelSmith:
@joshcooper

PR #3498, opened 2015-01-21
(PUP-2699) Remove as_platform shared context

joshcooper:
Commit 772f7566 added the ability to emulate posix or windows by
stubbing the corresponding platform facts, and file & path separators.
While a good idea in practice, it causes problems. For example, when
executing `as_platform => :posix` tests on a Windows host, the path
separator is changed to ':'. However, if any code calls into facter, it
will use the default fact search path of
`C:\ProgramData\PuppetLabs\facter\facts.d` which is interpreted to mean
two directories, 'C' and '\ProgramData\...' emitting warnings of the form:

    Excluding C from search path. Fact file paths must be an absolute directory

Another source of problems is when trying to use a file-based log
destination in a test. For example, if you create a log destination
using a posix path, but try to execute an `as_platform => :windows` test,
the file separator will be changed from '/' to '\\', which causes the
log destination to report that it is not a valid absolute path.

This commit eliminates the as_platform shared context and modifies the
respective tests to only execute on the corresponding platforms. It
means some unit tests will only execute on certain platforms, but that
is a small subset, and leads to more predictable results.


ferventcoder:
:+1:

PR #3499, opened 2015-01-21
(maint) Rename let `example` instance variable

joshcooper:
Previously, the test created an instance variable named `example`, for
which rspec created a method that memoized the instance variable. The
method shadowed the `example` method on RSpec::Core::ExampleGroup, which
can lead to surprising results. For example if you add an rspec before
hook:

    RSpec.configure do |config|
      config.before :each do |test|
        ...
      end
    end

Then `test` will be derived from `RSpec::Core::ExampleGroup`:

    (rdb:1) test.is_a? RSpec::Core::ExampleGroup
    true

But if you call `test.example` it will return the memoized variable,
instead of the RSpec::Core::Example:

    (rdb:2) test.example.class
    Puppet::Type::Ssh_authorized_key

This commit simply renames the instance variable/method.


ferventcoder:
:+1:

PR #3500, opened 2015-01-21
(PUP-3894) Update default license in metadata so that it matches string from http:/...

thlapin:
Module metadata default for license is still 'Apache 2.0'. Default should be in line with http://spdx.org/licenses/.
https://docs.puppetlabs.com/puppet/latest/reference/modules_publishing.html#write-a-metadatajson-file

### DIFF
--- a/lib/puppet/module_tool/metadata.rb
+++ b/lib/puppet/module_tool/metadata.rb
@@ -19,7 +19,7 @@ module Puppet::ModuleTool
       'version'      => nil,
       'author'       => nil,
       'summary'      => nil,
-      'license'      => 'Apache 2.0',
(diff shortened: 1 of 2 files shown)


joshcooper:
Hi @thlapin, thank you for your contribution. Could you open a jira ticket for this, and update your commit subject line to reference the ticket, e.g. (PUP-XXX) Update default license in metadata. The [CONTRIBUTING.md](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md) doc has more details. Thanks again!

/cc @andersonmills

andersonmills:
:+1:

PR #3501, opened 2015-01-21
(PUP-3548) Allow 4x function to call 3x function

thallgren:
The call_function() method of the Puppet::Pops::Functions::Function
class did not make any attempt to find functions that had been
declared using the 3x function API. This commit adds the following:

Proper 3x load and execute:
The call_function() method will now consult the 3x Functions class
to find (and optionally load) the function from there when it fails
to find it using the 4x loader. The load is always performed using
the closure scope of the caller. The 3x load will fail unless the
closure_scope is of type Puppet::Parser::Scope. The scope of the
caller (which might be different from closure_scope) will be the
instance that holds the method that corresponds to the 3x function.

Propagation of caller scope:
The call_function() will pass the closure_scope as that scope of the
caller. In some situations this is not the desired behavior. The
user may want to inject the caller scope, alter it in some way, and
the pass it on as a new caller scope. This commit adds a new API
method to facilitate this:

call_function_with_scope(scope, function_name, *args)

The approach of adding a new method was chosen in favor of adding
an extra scope argument to the existing call_function because in
general, functions should not modify the scope.

### DIFF
--- a/lib/puppet/pops/evaluator/runtime3_converter.rb
+++ b/lib/puppet/pops/evaluator/runtime3_converter.rb
@@ -1,0 +1,175 @@
+module Puppet::Pops::Evaluator
+# Converts nested 4x supported values to 3x values. This is required because
+# resources and other objects do not know about the new type system, and does not support
+# regular expressions. Unfortunately this has to be done for array and hash as well.
+# A complication is that catalog types needs to be resolved against the scope.
+#
+# Users should not create instances of this class. Instead the class methods {Runtime3Converter.convert},
+# {Runtime3Converter.convert_args}, or {Runtime3Converter.instance} should be used
+class Runtime3Converter
+  # Converts 4x supported values to a 3x values using the singleton instance.
+  #
+  # @param args [Array] Array of values to convert
+  # @param scope [Puppet::Parser::Scope] The scope to use when converting
+  # @param undef_value [Object] The value that nil is converted to
+  # @return [Array] The converted values
+  #
+  def self.convert_args(args, scope, undef_value)
+    @@instance.convert_args(args, scope, undef_value)
+  end
+
+  # Converts a 4x supported value to a 3x value using the singleton instance.
+  #
+  # @param o [Object]The value to convert
+  # @param scope [Puppet::Parser::Scope] The scope to use when converting
+  # @param undef_value [Object] The value that nil is converted to
+  # @return [Object] The converted value
+  #
+  def self.convert(o, scope, undef_value)
+    @@instance.convert(o, scope, undef_value)
+  end
+
+  # Returns the singleton instance of this class.
+  # @return [Runtime3Converter] The singleton instance
+  def self.instance
+    @@instance
+  end
+
+  # Converts 4x supported values to a 3x values.
+  #
+  # @param args [Array] Array of values to convert
+  # @param scope [Puppet::Parser::Scope] The scope to use when converting
+  # @param undef_value [Object] The value that nil is converted to
+  # @return [Array] The converted values
+  #
+  def convert_args(args, scope, undef_value)
+    args.map {|a| convert(a, scope, undef_value) }
+  end
+
+  # Converts a 4x supported value to a 3x value.
+  #
+  # @param o [Object]The value to convert
+  # @param scope [Puppet::Parser::Scope] The scope to use when converting
+  # @param undef_value [Object] The value that nil is converted to
+  # @return [Object] The converted value
+  #
+  def convert(o, scope, undef_value)
+    @convert_visitor.visit_this_2(self, o, scope, undef_value)
+  end
+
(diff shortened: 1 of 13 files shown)


lib/puppet/pops/evaluator/runtime3_converter.rb
```diff
@@ -0,0 +1,175 @@
...
+  # @param o [Object]The value to convert
+  # @param scope [Puppet::Parser::Scope] The scope to use when converting
+  # @param undef_value [Object] The value that nil is converted to
+  # @return [Object] The converted value
+  #
+  def convert2(o, scope, undef_value)
+    @convert2_visitor.visit_this_2(self, o, scope, undef_value)
+  end
+end
+end
```
hlindberg: missing end of line

lib/puppet/pops/evaluator/runtime3_converter.rb
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  # Converts 4x supported values to a 3x values.
+  #
+  # @param args [Array] Array of values to convert
+  # @param scope [Puppet::Parser::Scope] The scope to use when converting
+  # @param undef_value [Object] The value that nil is converted to
+  # @return [Array] The converted values
+  #
+  def convert_args(args, scope, undef_value)
```
hlindberg: name of this method is a bit confusing as it looks like one of the polymorphic dispatch methods. (and yes you can have a class that returns a lower case class name if you really want to, so there is a small but possible clash). Maybe call it 'map_args' ?

lib/puppet/pops/functions/function.rb
```diff
@@ -47,18 +47,46 @@ def call(scope, *args)
   # Allows the implementation of a function to call other functions by name. The callable functions
   # are those visible to the same loader that loaded this function (the calling function).
   #
```
hlindberg: It is worth mentioning that this calls the referenced function with the calling functions closure scope as the caller's scope (typically "global scope").

lib/puppet/pops/evaluator/runtime3_converter.rb
```diff
@@ -0,0 +1,175 @@
+module Puppet::Pops::Evaluator
+# Converts nested 4x supported values to 3x values. This is required because
+# resources and other objects do not know about the new type system, and does not support
+# regular expressions. Unfortunately this has to be done for array and hash as well.
+# A complication is that catalog types needs to be resolved against the scope.
+#
+# Users should not create instances of this class. Instead the class methods {Runtime3Converter.convert},
+# {Runtime3Converter.convert_args}, or {Runtime3Converter.instance} should be used
```
hlindberg: I suggested a name change of this method below

lib/puppet/pops/evaluator/runtime3_converter.rb
```diff
@@ -0,0 +1,175 @@
...
+class Runtime3Converter
+  # Converts 4x supported values to a 3x values using the singleton instance.
+  #
```
hlindberg: the implementation detail "using a singleton detail" is not of interest to the reader

lib/puppet/pops/evaluator/runtime3_converter.rb
```diff
@@ -0,0 +1,175 @@
...
+  # @param scope [Puppet::Parser::Scope] The scope to use when converting
+  # @param undef_value [Object] The value that nil is converted to
+  # @return [Array] The converted values
+  #
+  def self.convert_args(args, scope, undef_value)
+    @@instance.convert_args(args, scope, undef_value)
+  end
+
+  # Converts a 4x supported value to a 3x value using the singleton instance.
+  #
```
hlindberg: same as above

lib/puppet/pops/evaluator/runtime3_converter.rb
```diff
@@ -0,0 +1,175 @@
...
+# {Runtime3Converter.convert_args}, or {Runtime3Converter.instance} should be used
+class Runtime3Converter
+  # Converts 4x supported values to a 3x values using the singleton instance.
+  #
+  # @param args [Array] Array of values to convert
+  # @param scope [Puppet::Parser::Scope] The scope to use when converting
+  # @param undef_value [Object] The value that nil is converted to
+  # @return [Array] The converted values
+  #
+  def self.convert_args(args, scope, undef_value)
```
hlindberg: this name is fine as there are no polymorphic methods on the class, but should probably be renamed to the same as what the instance method is renamed to.

lib/puppet/pops/functions/function.rb
```diff
@@ -74,4 +69,40 @@ def self.dispatcher
   def self.signatures
     @dispatcher.signatures
   end
+
+  private
+
```
hlindberg: change to protected

PR #3506, opened 2015-01-22
(PUP-3897) Add acceptance: ensure new environment picked up by agent

er0ck:
in PUP-3755 there are two scenarios that lead to a newly classified
agent not picking up its new environment.  This test should catch both
of the cases that aren't tested by spec.
This test does not use the environment helpers as they circumvent most
of the functionality in this test such as hard wiring to test
environment and specifying agent environment on commandline.  We need to
classify a node's environment from master.

### DIFF
--- a/acceptance/tests/environment/environment_scenario-new.rb
+++ b/acceptance/tests/environment/environment_scenario-new.rb
@@ -1,0 +1,54 @@
+test_name "Test a new environment, unknown to agent"
+require 'pry'
+
+step "setup environments"
+
+testdir = create_tmpdir_for_user master, "confdir"
+manifest = <<-MANIFEST
+  File {
+    ensure => directory,
+    owner => #{master['user']},
+    group => #{master['group']},
+    mode => "0750",
+  }
+
+  file { "#{testdir}":;
+    "#{testdir}/environments":;
+    "#{testdir}/environments/production":;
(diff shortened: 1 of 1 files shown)


acceptance/tests/environment/environment_scenario-new.rb
```diff
@@ -0,0 +1,54 @@
+test_name "Test a new environment, unknown to agent"
+require 'pry'
```
joshcooper: @er0ck please remove pry

joshcooper:
Can you retarget at stable?

acceptance/tests/environment/environment_scenario-new.rb
```diff
@@ -0,0 +1,54 @@
...
+    'environmentpath' => "#{testdir}/environments",
+  },
+  'agent' => {
+    'environment' => 'debug'
+  }
+}
+
+with_puppet_running_on(master, master_opts, testdir) do
+  agents.each do |agent|
+    on(agent, puppet("agent", "--test"), :acceptable_exit_codes => (0..255) ) do
```
joshcooper: By default the puppet agent will try to connect to `puppet` as the server, so we usually specify the `--server #{master}` args explicitly. Ideally the agent would be configured once during setup and each acceptance tests wouldn't need to specifying tath.

PR #3507, opened 2015-01-22
(PUP-3887) Fix checksum and source_permissions option in file servers

MikaelSmith:
The file_server uses the requested checksum_type to calculate a
file checksum, rather than always defaulting to md5.

### DIFF
--- a/lib/puppet/file_serving/content.rb
+++ b/lib/puppet/file_serving/content.rb
@@ -21,11 +21,9 @@ class Puppet::FileServing::Content < Puppet::FileServing::Base
     instance
   end
 
-  # BF: we used to fetch the file content here, but this is counter-productive
-  # for puppetmaster streaming of file content. So collect just returns itself
+  # This is no longer used, but is still called by the file server implementations when interacting
+  # with their model abstraction.
   def collect(source_permissions = nil)
-    return if stat.ftype == "directory"
-    self
   end
 
   # Read the content of our file in.
--- a/lib/puppet/indirector/direct_file_server.rb
+++ b/lib/puppet/indirector/direct_file_server.rb
@@ -7,9 +7,7 @@ class Puppet::Indirector::DirectFileServer < Puppet::Indirector::Terminus
 
   def find(request)
     return nil unless Puppet::FileSystem.exist?(request.key)
-    instance = model.new(request.key)
-    instance.links = request.options[:links] if request.options[:links]
-    instance
+    path2instance(request, request.key)
   end
 
   def search(request)
--- a/lib/puppet/indirector/file_metadata/file.rb
+++ b/lib/puppet/indirector/file_metadata/file.rb
@@ -4,19 +4,4 @@ require 'puppet/indirector/direct_file_server'
 
 class Puppet::Indirector::FileMetadata::File < Puppet::Indirector::DirectFileServer
   desc "Retrieve file metadata directly from the local filesystem."
-
-  def find(request)
-    return unless data = super
-    data.collect(request.options[:source_permissions])
-
-    data
-  end
-
-  def search(request)
-    return unless result = super
-
-    result.each { |instance| instance.collect }
-
-    result
-  end
 end
--- a/lib/puppet/indirector/file_server.rb
+++ b/lib/puppet/indirector/file_server.rb
@@ -27,10 +27,7 @@ class Puppet::Indirector::FileServer < Puppet::Indirector::Terminus
     # The mount checks to see if the file exists, and returns nil
     # if not.
     return nil unless path = mount.find(relative_path, request)
-    result = model.new(path)
-    result.links = request.options[:links] if request.options[:links]
-    result.collect(request.options[:source_permissions])
-    result
+    path2instance(request, path)
   end
 
   # Search for files.  This returns an array rather than a single
@@ -42,18 +39,7 @@ class Puppet::Indirector::FileServer < Puppet::Indirector::Terminus
       Puppet.info "Could not find filesystem info for file '#{request.key}' in environment #{request.environment}"
       return nil
     end
-
-    filesets = paths.collect do |path|
-      # Filesets support indirector requests as an options collection
-      Puppet::FileServing::Fileset.new(path, request)
-    end
-
-    Puppet::FileServing::Fileset.merge(*filesets).collect do |file, base_path|
-      inst = model.new(base_path, :relative_path => file)
-      inst.links = request.options[:links] if request.options[:links]
-      inst.collect
-      inst
-    end
+    path2instances(request, *paths)
   end
 
   private
--- a/spec/integration/indirector/direct_file_server_spec.rb
+++ b/spec/integration/indirector/direct_file_server_spec.rb
@@ -2,6 +2,7 @@ require 'spec_helper'
 require 'matchers/include'
 
 require 'puppet/indirector/file_content/file'
+require 'puppet/indirector/file_metadata/file'
 
 describe Puppet::Indirector::DirectFileServer, " when interacting with the filesystem and the model" do
   include PuppetSpec::Files
@@ -9,30 +10,21 @@ describe Puppet::Indirector::DirectFileServer, " when interacting with the files
   before do
     # We just test a subclass, since it's close enough.
     @terminus = Puppet::Indirector::FileContent::File.new
-
-    @filepath = make_absolute("/path/to/my/file")
   end
 
   it "should return an instance of the model" do
-    if Puppet.features.microsoft_windows?
-      skip("porting to Windows")
-    else
-      Puppet::FileSystem.expects(:exist?).with(@filepath).returns(true)
+    filepath = make_absolute("/path/to/my/file")
+    Puppet::FileSystem.expects(:exist?).with(filepath).returns(true)
 
-      expect(@terminus.find(@terminus.indirection.request(:find, "file://host#{@filepath}", nil))).to be_instance_of(Puppet::FileServing::Content)
-    end
+    expect(@terminus.find(@terminus.indirection.request(:find, Puppet::Util.path_to_uri(filepath).to_s, nil))).to be_instance_of(Puppet::FileServing::Content)
   end
 
   it "should return an instance capable of returning its content" do
-    if Puppet.features.microsoft_windows?
-      skip("porting to Windows")
-    else
-      filename = file_containing("testfile", "my content")
+    filename = file_containing("testfile", "my content")
 
-      instance = @terminus.find(@terminus.indirection.request(:find, "file://host#{filename}", nil))
+    instance = @terminus.find(@terminus.indirection.request(:find, Puppet::Util.path_to_uri(filename).to_s, nil))
 
-      expect(instance.content).to eq("my content")
-    end
+    expect(instance.content).to eq("my content")
   end
 end
 
@@ -58,11 +50,43 @@ describe Puppet::Indirector::DirectFileServer, " when interacting with FileServi
     File.open(File.join(path, "two"), "w") { |f| f.print "two content" }
 
     terminus = Puppet::Indirector::FileContent::File.new
-    request = terminus.indirection.request(:search, "file:///#{path}", nil, :recurse => true)
+    request = terminus.indirection.request(:search, Puppet::Util.path_to_uri(path).to_s, nil, :recurse => true)
 
     expect(terminus.search(request)).to include_in_any_order(
       file_with_content(File.join(path, "one"), "one content"),
       file_with_content(File.join(path, "two"), "two content"),
       directory_named(path))
+  end
+end
+
+describe Puppet::Indirector::DirectFileServer, " when interacting with filesystem metadata" do
+  include PuppetSpec::Files
+  include_context 'with supported checksum types'
+
+  before do
+    @terminus = Puppet::Indirector::FileMetadata::File.new
+  end
+
+  with_checksum_types("file_metadata", "testfile") do
+    it "should return the correct metadata" do
+      request = @terminus.indirection.request(:find, Puppet::Util.path_to_uri(checksum_file).to_s, nil, :checksum_type => checksum_type)
+      result = @terminus.find(request)
+      expect_correct_checksum(result, checksum_type, checksum, Puppet::FileServing::Metadata)
+    end
+  end
+
+  with_checksum_types("direct_file_server_testing", "testfile") do
+    it "search of FileServing::Fileset should return the correct metadata" do
+      request = @terminus.indirection.request(:search, Puppet::Util.path_to_uri(env_path).to_s, nil, :recurse => true, :checksum_type => checksum_type)
+      result = @terminus.search(request)
+
+      expect(result).to_not be_nil
+      expect(result.length).to eq(2)
+      file, dir = result.partition {|x| x.relative_path == 'testfile'}
+      expect(file.length).to eq(1)
+      expect(dir.length).to eq(1)
+      expect_correct_checksum(dir[0], 'ctime', "#{CHECKSUM_STAT_TIME}", Puppet::FileServing::Metadata)
+      expect_correct_checksum(file[0], checksum_type, checksum, Puppet::FileServing::Metadata)
+    end
   end
 end
--- a/spec/integration/indirector/file_metadata/file_server_spec.rb
+++ b/spec/integration/indirector/file_metadata/file_server_spec.rb
@@ -9,6 +9,7 @@ require 'puppet_spec/files'
 describe Puppet::Indirector::FileMetadata::FileServer, " when finding files" do
   it_should_behave_like "Puppet::Indirector::FileServerTerminus"
   include PuppetSpec::Files
+  include_context 'with supported checksum types'
 
   before do
     @terminus = Puppet::Indirector::FileMetadata::FileServer.new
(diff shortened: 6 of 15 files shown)


MikaelSmith:
I'm exploring ways to improve the spec testing on checksums, so there may be a bit more added to this.

MikaelSmith:
Done. Any suggestions on how to use rspec to simplify file_server_spec would be great.

peterhuene:
My only suggestion would be to merge it with the spec above it by using a mapping of "property to validate" to "value to set".  They seem to be exactly the same save for which property is being validated.

spec/integration/indirector/file_metadata/file_server_spec.rb
```diff
@@ -4,11 +4,147 @@
...
+
+    CHECKSUMS.each do |checksum_type, checksum|
+      it "should find the expected checksum with checksum_type #{checksum_type} specified" do
+        result = Puppet::FileServing::Metadata.indirection.search("plugins", :environment => @env, :checksum_type => checksum_type, :recurse => true)
+
+        result.should_not be_nil
+        result.length.should == 2
+        result.map {|x| x.should be_instance_of(Puppet::FileServing::Metadata) }
+        result.find {|x| x.relative_path == 'file.rb' }.checksum_type.should == checksum_type
+        result.find {|x| x.relative_path == 'file.rb' }.checksum.should == checksum
```
joshcooper: @kylog has PR up to switch to rspec-3, it would be great to use the new syntax for new PRs, e.g. `expect(result.length).to eq(2)`

spec/unit/file_serving/metadata_spec.rb
```diff
@@ -108,14 +108,29 @@
...
               metadata.collect
-              metadata.checksum.should == "{mtime}#{@time}"
+              metadata.checksum.should == "{#{digest_algorithm}}#{checksum}"
             end
           end
+
+          it "should give a mtime checksum when checksum_type is set" do
+            time = Time.now
+            metadata.checksum_type = "mtime"
+            metadata.expects(:mtime_file).returns(@time)
```
joshcooper: should this be returning `time` as opposed to `@time`?

spec/unit/file_serving/metadata_spec.rb
```diff
@@ -108,14 +108,29 @@
...
+            metadata.collect
+            metadata.checksum.should == "{mtime}#{@time}"
+          end
+
+          it "should give a ctime checksum when checksum_type is set" do
+            time = Time.now
+            metadata.checksum_type = "ctime"
+            metadata.expects(:ctime_file).returns(@time)
```
joshcooper: same comment here

lib/puppet/indirector/file_server.rb
```diff
@@ -51,6 +52,7 @@ def search(request)
     Puppet::FileServing::Fileset.merge(*filesets).collect do |file, base_path|
       inst = model.new(base_path, :relative_path => file)
       inst.links = request.options[:links] if request.options[:links]
+      inst.checksum_type = request.options[:checksum_type] if request.options[:checksum_type]
       inst.collect
```
joshcooper: unrelated to your PR, but should we be doing `inst.collect(request.options[:source_permissions])` like above?

MikaelSmith: It kind of seems like we should.

spec/integration/indirector/file_metadata/file_server_spec.rb
```diff
@@ -4,11 +4,147 @@
...
+  describe "with a plugin environment specified in the request" do
+    before do
+      path = tmpfile("file_content_with_env")
+
+      Dir.mkdir(path)
+
+      modpath = File.join(path, "mod")
+      FileUtils.mkdir_p(File.join(modpath, "lib"))
+      file = File.join(modpath, "lib", "file.rb")
+      File.open(file, "wb") { |f| f.write "1\r\n" }
```
joshcooper: :+1: to including "\r" to test binary mode.

I'm thinking we should increase the length of this file, so that we can test that the "lite" checksums are different than their non-lite counterparts.

spec/integration/indirector/file_metadata/file_server_spec.rb
MikaelSmith: That replaces all uses of `.should ==`?

spec/integration/indirector/file_metadata/file_server_spec.rb
MikaelSmith: Yup, going to add a test for longer file lengths.

spec/integration/indirector/file_metadata/file_server_spec.rb
```diff
@@ -4,11 +4,147 @@
...
+
+      @env = Puppet::Node::Environment.create(:foo, [path])
+    end
+
+    CHECKSUMS.each do |checksum_type, checksum|
+      it "should find the expected checksum with checksum_type #{checksum_type} specified" do
+        result = Puppet::FileServing::Metadata.indirection.search("plugins", :environment => @env, :checksum_type => checksum_type, :recurse => true)
+
+        result.should_not be_nil
+        result.length.should == 2
```
joshcooper: why does the search request return two entries? Is there an entry for '.' in there?

lib/puppet/indirector/file_server.rb
MikaelSmith: I'll open a new ticket. Adding it doesn't break any tests, so we'll need some test coverage for it.

spec/unit/indirector/file_server_spec.rb
```diff
@@ -206,6 +206,28 @@ module Testing; end
...
     end
 
+    describe "should set 'checksum_type' on the instances if it is set in the request options" do
+      [:md5, :md5lite, :sha256, :sha256lite, :mtime, :ctime].each do |checksum_type|
+        it "to #{checksum_type}" do
+          @configuration.expects(:split_path).with(@request).returns([@mount, "rel/path"])
+
+          @mount.expects(:search).with { |key, request| key == "rel/path" }.returns []
+
+          Puppet::FileSystem.stubs(:exist?).returns true
```
joshcooper: You may want to restrict this to a specific path, otherwise the file-related settings code called during test setup/teardown could be affected.

spec/integration/indirector/file_metadata/file_server_spec.rb
MikaelSmith: Yes, that's what it's finding.

spec/integration/indirector/file_metadata/file_server_spec.rb
joshcooper: yeah, plus some things like `expect(..).to be_truthy`. There's a good writeup of rspec3 in http://myronmars.to/n/dev-blog/2014/05/notable-changes-in-rspec-3

spec/unit/indirector/file_server_spec.rb
MikaelSmith: This was pulled from file_content/file_server_spec.rb. I'll take a look at whether both can be cleaned up.

peterhuene:
@MikaelSmith Just noticed that one of your commits is from "root".

MikaelSmith:
I think I've addressed all the comments.

lib/puppet/indirector/file_server.rb
```diff
@@ -29,6 +29,7 @@ def find(request)
     return nil unless path = mount.find(relative_path, request)
     result = model.new(path)
     result.links = request.options[:links] if request.options[:links]
+    result.checksum_type = request.options[:checksum_type] if request.options[:checksum_type]
```
peterhuene: After reviewing this fix and the fix for PUP-3888, would it be possible to set these options only once in a helper method, perhaps in `Puppet::FileServing::TerminusHelper`?  Seem like there's four places where we're setting both `links` and `checksum_type` (the two in this PR, the one in the fix for PUP-3888, and one already present in `Puppet::FileServing::TerminusHelper`).

MikaelSmith: I see that too, I'll start looking at it.

peterhuene: Also, seems like just passing the options hash along to the constructor of the model rather than setting each option explicitly should do the trick, although there's one method in `Puppet::FileServing::Fileset` we don't set in the indirector (`ignore=`).  Not sure if we want explicitness where we instantiate the model or just pass through the request options.

MikaelSmith: That doesn't work, because there are some options (source_permissions)
don't exist and it throws an exception. I gave that a try already.

peterhuene: Ah, rats.  Worth a shot at least :)

MikaelSmith:
@Iristyle @ferventcoder this now covers changes related to PUP-1450 as well, applying source_permissions (if specified) on calls to the direct_file_server and searching for files via the file_server. It'd be great if one of you could review those changes.

MikaelSmith:
Is source_permissions even needed anymore, now that we defaulted to ignore on Windows?

MikaelSmith:
Planning to rewrite file_server_spec and direct_file_server_spec based off a single shared_context.

MikaelSmith:
I believe this PR's complete now.

lib/puppet/indirector/direct_file_server.rb
```diff
@@ -7,9 +7,7 @@ class Puppet::Indirector::DirectFileServer < Puppet::Indirector::Terminus
 
   def find(request)
     return nil unless Puppet::FileSystem.exist?(request.key)
-    instance = model.new(request.key)
-    instance.links = request.options[:links] if request.options[:links]
-    instance
+    path2instance(request, request.key)
```
peterhuene: Unlike the other calls to `path2instance`, this one was previously lacking a call to `collect`. This appears safe to do for the direct file server, but is it an operation (stating the source file) that needs to be done (although it is nice to be consistent with the other calls, for sure)?

MikaelSmith: I'm going to talk with @Iristyle about that this afternoon.

MikaelSmith: The resolution after talking with @Iristyle and @joshcooper this afternoon was two things:
* It's not clear whether ignore_source_permissions should be passed to #collect in path2instances; it wasn't in the past, but that could still be a bug. Tracking in PUP-3906.
* The collect method is only defined for file_serving/content.rb and file_serving/metadata.rb, and now only used by path2instance. It was previously called for this code path by file_metadata/file.rb (corresponding to the metadata.rb model) and not called by file_content/file.rb (the content.rb model). The return value of file_serving/content.rb's #collect was never used, so I've removed the implementation and combined call paths to simplify the DirectFileServer and FileServer logic.

lib/puppet/indirector/file_metadata/file.rb
```diff
@@ -4,19 +4,4 @@
 
 class Puppet::Indirector::FileMetadata::File < Puppet::Indirector::DirectFileServer
   desc "Retrieve file metadata directly from the local filesystem."
-
-  def find(request)
-    return unless data = super
-    data.collect(request.options[:source_permissions])
```
MikaelSmith: Implementation removed because it's now redundant, collect is called by the DirectFileServer.

lib/puppet/file_serving/content.rb
```diff
@@ -21,11 +21,9 @@ def self.from_raw(content)
     instance
   end
 
-  # BF: we used to fetch the file content here, but this is counter-productive
-  # for puppetmaster streaming of file content. So collect just returns itself
+  # This is no longer used, but is still called by the file server implementations when interacting
+  # with their model abstraction.
   def collect(source_permissions = nil)
-    return if stat.ftype == "directory"
```
MikaelSmith: Implementation removed because it's dead code, the results aren't used anywhere. The only call to this method now comes from the inherited #find/#search calls in indirector/file_content/{file.rb, file_server.rb}.

spec/integration/indirector/direct_file_server_spec.rb
```diff
@@ -2,6 +2,15 @@
 require 'matchers/include'
 
 require 'puppet/indirector/file_content/file'
+require 'puppet/indirector/file_metadata/file'
```
MikaelSmith: Rewritten to make Windows work.

spec/unit/indirector/file_metadata/file_spec.rb
```diff
@@ -43,7 +43,7 @@
 
     it "should collect the attributes of the instances returned" do
       Puppet::FileSystem.expects(:exist?).with(@path).returns true
-      @metadata.expects(:path2instances).returns( [mock("one", :collect => nil), mock("two", :collect => nil)] )
```
MikaelSmith: Going to rewrite this to be a stronger assertion.

MikaelSmith: Rewrote it and had to mock up file searching quite a bit.

ferventcoder:
@MikaelSmith I'm just coming back up to speed. What is the current state of this versus PUP-1450?

joshcooper:
nit, commit 59237458e4b756b47160ec069ffefdfed5b44eda, still references PUP-3887 and PUP-3888

MikaelSmith:
No acceptance failures.

@ferventcoder The portion related to PUP-1450: "Also discovered that source_permissions set to :ignore was not properly followed when examining file metadata." I believe searching for files (via recursive source?) could still be affected by whatever that problem is, but don't have an example, so filed PUP-3906 to follow up on it.

I decided to leave the behavior changes related to PUP-1450 untouched, with a comment about it at https://github.com/puppetlabs/puppet/pull/3507/files#diff-7f963a19de967fc91a97c00c11dde29fR12.

MikaelSmith:
Squashed commits and fixed the commit title. @joshcooper

lib/puppet/indirector/direct_file_server.rb
MikaelSmith: DirectFileServer#search had the opposite problem, it called collect twice: once in path2instances, and again in file_metadata/file.rb.

joshcooper:
I'm seeing an issue when using a file `source` whose URI scheme is `puppet` and setting a global `digest_algorithm` to `sha256`. The agent is always sending md5 in the (singular) `file_metadata` request.

I installed `puppetlabs-stdlib` on my master, and created a file in `.../modules/stdlib/files/source.txt`. Then created a manifest containing:

<pre>
file { '/tmp/files':
  ensure => directory
}

file { '/tmp/files/puppet.txt':
  ensure   => file,
  source   => 'puppet:///modules/stdlib/source.txt',
}
</pre>

When I run the agent multiple times, the file "changes". Note I'm not using `--test`, because of PUP-3916, and instead specifying the other options that it implies minus `show_diff`:

<pre>
$ bundle exec puppet agent --no-daemonize --no-show_diff --onetime --verbose  --digest_algorithm sha256
...
Notice: /Stage[main]/Main/File[/tmp/files/puppet.txt]/content: content changed '{sha256}fd7b35a699bafda0b74e2af0b816a22583bb89803e1be3b2677758399109cca0' to '{md5}0ef8258a06349161b34c1083603090a5'
</pre>

In looking at the HTTP requests, the `checksum_type` parameter is present during pluginsync (for modules and external facts), but not for file resources:

<pre>
$ bundle exec puppet agent --no-daemonize --no-show_diff --onetime --http_debug --digest_algorithm sha256 2>&1 >/dev/null | grep '/puppet/v3'
...
&lt;- "GET /puppet/v3/file_metadatas/pluginfacts?environment=production&links=manage&recurse=true&ignore=.svn&ignore=CVS&ignore=.git&checksum_type=sha256 HTTP/1.1\r\nAccept: pson, msgpack, yaml, raw\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Ruby\r\nHost: arcturus.local:8140\r\n\r\n"
&lt;- "GET /puppet/v3/file_metadatas/plugins?environment=production&links=manage&recurse=true&ignore=.svn&ignore=CVS&ignore=.git&checksum_type=sha256 HTTP/1.1\r\nAccept: pson, msgpack, yaml, raw\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Ruby\r\nHost: arcturus.local:8140\r\n\r\n"
...
&lt;- "GET /puppet/v3/file_metadata/modules/stdlib/source.txt?environment=production&links=manage&source_permissions=ignore HTTP/1.1\r\nAccept: pson, msgpack, yaml, raw\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Ruby\r\nHost: arcturus.local:8140\r\n\r\n"
&lt;- "GET /puppet/v3/file_content/modules/stdlib/source.txt?environment=production& HTTP/1.1\r\nAccept: raw\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Ruby\r\nHost: arcturus.local:8140\r\n\r\n"
&lt;- "GET /puppet/v3/file_metadata/modules/stdlib/source.txt?environment=production&links=manage&source_permissions=ignore HTTP/1.1\r\nAccept: pson, msgpack, yaml, raw\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Ruby\r\nHost: arcturus.local:8140\r\n\r\n"
&lt;- "GET /puppet/v3/file_content/modules/stdlib/source.txt?environment=production& HTTP/1.1\r\nAccept: raw\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Ruby\r\nHost: arcturus.local:8140\r\n\r\n"
...
</pre>

If I specify the checksum in the file resource:

<pre>
file { '/tmp/files/puppet.txt':
  ensure   => file,
  source   => 'puppet:///modules/stdlib/source.txt',
  checksum => sha256,
}
</pre>

Puppet does not report the file as being changed:

<pre>
$ bundle exec puppet agent --no-daemonize --no-show_diff --onetime --verbose
...
Info: Applying configuration version '1422407509'
</pre>

But it's because it's actually comparing using md5, note `checksum_type` is missing:

<pre>
$ bundle exec puppet agent --no-daemonize --no-show_diff --onetime --http_debug 2>&1 >/dev/null | grep '/puppet/v3'
&lt;- "GET /puppet/v3/file_metadata/modules/stdlib/source.txt?environment=production&links=manage&source_permissions=ignore HTTP/1.1\r\nAccept: pson, msgpack, yaml, raw\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Ruby\r\nHost: arcturus.local:8140\r\n\r\n"
...
</pre>

And if I modify the local file, it shows the change message in terms of md5:

<pre>
$ bundle exec puppet agent --no-daemonize --no-show_diff --onetime --verbose
...
Notice: /Stage[main]/Main/File[/tmp/files/puppet.txt]/content: content changed '{md5}85796b70bce3801a94ea40b61bddb765' to '{md5}0ef8258a06349161b34c1083603090a5'
</pre>

MikaelSmith:
Yes, that's what https://github.com/MikaelSmith/puppet/commit/51395c71ab6b67144599f560ddf059a629ecdcdb resolves, and should be my next PR.

I broke the problems up because sufficient testing is going to be complicated, and I wanted to be able to isolate the problems. I'd suggest testing with curl to simulate correct agent requests.

spec/integration/indirector/direct_file_server_spec.rb
```diff
@@ -2,6 +2,15 @@
...
+
+def make_uri(path)
+  if Puppet.features.microsoft_windows?
+    "file:///#{path}"
+  else
+    "file://host#{path}"
+  end
```
joshcooper: I'm confused as to why the old version of the tests specify `file://host`. The unit tests for direct_file_server use `Puppet::Util.path_to_uri` to generate a URI, and it doesn't contain an authority component on either windows or posix:

<pre>
irb(main):003:0> Puppet::Util.path_to_uri('/foo').to_s
=> "file:/foo"
</pre>

<pre>
irb(main):018:0* Puppet::Util.path_to_uri('C:/foo').to_s
=> "file:/C:/foo"
</pre>

I added the `path_to_uri` method in 0ce60a569604105e5e7ef47431d239bcd0d46b27, and what I found is that `file://<host>/C:/foo` is used to express UNC paths on windows (even though `C:/foo` is local). In order to refer to a local file, it is better to use either `file:/C:/foo` or `file:///C:/foo`, where the former is allowed because the authority component is optional (https://tools.ietf.org/html/rfc3986#section-3.2)

But I'm wondering why windows and posix aren't consistent about the authority component. Is the test modeling the current behavior (posix agents include it, but windows ones don't)? I'm guessing the current tests don't match reality and that we should remove the authority component in the posix case. If so, we can just call `path_to_uri` instead.

spec/integration/indirector/direct_file_server_spec.rb
```diff
@@ -66,3 +66,35 @@
...
+
+describe Puppet::Indirector::DirectFileServer, " when interacting with filesystem metadata" do
+  include PuppetSpec::Files
+  include_context 'with supported checksum types'
+
+  before do
+    @terminus = Puppet::Indirector::FileMetadata::File.new
+  end
+
+  with_checksum_types("file_content", "testfile") do
```
joshcooper: use `"file_metadata"` instead to match terminus class?

spec/shared_contexts/checksum.rb
```diff
@@ -0,0 +1,48 @@
...
+CHECKSUM_TYPES_TO_TRY = [
+  ['md5', 'a7a169ac84bb863b30484d0aa03139c1'],
+  ['md5lite', '22b4182363e81b326e98231fde616782'],
+  ['sha256', '47fcae62967db2fb5cba2fc0d9cf3e6767035d763d825ecda535a7b1928b9746'],
+  ['sha256lite', 'fd50217a2b0286ba25121bf2297bbe6c197933992de67e4e568f19861444ecf8'],
+  ['ctime', "#{CHECKSUM_STAT_TIME}"],
+  ['mtime', "#{CHECKSUM_STAT_TIME}"]
+]
+
+shared_context('with supported checksum types') do
```
joshcooper: I'm not a huge fan of shared contexts, because data (really memoized methods) are magically available inside of your test, like [`checksum`](https://github.com/puppetlabs/puppet/pull/3507/files#diff-ff71306e757bf24c8e36adb4bc7da8cfR33) below. It gets really confusing if your test also has a locally defined `let(:checksum)`.

We've been moving more towards things like `/spec/lib/puppet_spec/compiler.rb`. It's more explicit and less magical. That way you can include the module in your test, call the helper methods, passing all of the required data.

PR #3509, opened 2015-01-23
(PUP-3812) Change file_bucket content type to application/octet-stream

dankreek:
Previously the MIME type that the filebucket endpoints used to transfer
data was `text/plain`. This leads to potential problems transfering
binary data. This commit adds a new format called `:binary` which uses
`application/octet-stream` as the MIME type, which is the correct type
to specify arbitrary binary data.

### DIFF
--- a/lib/puppet/file_bucket/file.rb
+++ b/lib/puppet/file_bucket/file.rb
@@ -14,16 +14,11 @@ class Puppet::FileBucket::File
   attr :bucket_path
 
   def self.supported_formats
-    [:s]
+    [:binary]
   end
 
   def self.default_format
-    # This should really be :raw, like is done for Puppet::FileServing::Content
-    # but this class hasn't historically supported `from_raw`, so switching
-    # would break compatibility between newer 3.x agents talking to older 3.x
-    # masters. However, to/from_s has been supported and achieves the desired
-    # result without breaking compatibility.
-    :s
+    :binary
   end
 
   def initialize(contents, options = {})
@@ -67,6 +62,10 @@ class Puppet::FileBucket::File
     @contents.to_s
   end
 
+  def to_binary
+    @contents.to_s
+  end
+
   def contents
     to_s
   end
@@ -76,6 +75,10 @@ class Puppet::FileBucket::File
   end
 
   def self.from_s(contents)
+    self.new(contents)
+  end
+
+  def self.from_binary(contents)
     self.new(contents)
   end
 
--- a/lib/puppet/file_bucket/file.rb
+++ b/lib/puppet/file_bucket/file.rb
@@ -59,23 +59,19 @@ class Puppet::FileBucket::File
   end
 
   def to_s
-    @contents.to_s
+    to_binary
   end
 
   def to_binary
-    @contents.to_s
+    @contents.to_binary
   end
 
   def contents
-    to_s
+    to_binary
   end
 
   def name
     "#{checksum_type}/#{checksum_data}"
-  end
-
-  def self.from_s(contents)
-    self.new(contents)
   end
 
   def self.from_binary(contents)
@@ -84,7 +80,7 @@ class Puppet::FileBucket::File
 
   def to_data_hash
     # Note that this serializes the entire data to a string and places it in a hash.
-    { "contents" => contents.to_s }
+    { "contents" => contents.to_binary }
   end
 
   def self.from_data_hash(data)
@@ -116,7 +112,7 @@ class Puppet::FileBucket::File
       Puppet::Util::Checksums.method(base_method).call(@contents)
     end
 
-    def to_s
+    def to_binary
       # This is not so horrible as for FileContent, but still possible to mutate the content that the
       # checksum is based on... so semi horrible...
       return @contents;
@@ -141,7 +137,7 @@ class Puppet::FileBucket::File
       Puppet::Util::Checksums.method(:"#{base_method}_file").call(@path)
     end
 
-    def to_s
+    def to_binary
       Puppet::FileSystem::binread(@path)
     end
   end
--- a/lib/puppet/file_serving/content.rb
+++ b/lib/puppet/file_serving/content.rb
@@ -12,10 +12,10 @@ class Puppet::FileServing::Content < Puppet::FileServing::Base
   attr_writer :content
 
   def self.supported_formats
-    [:raw]
+    [:binary]
   end
 
-  def self.from_raw(content)
+  def self.from_binary(content)
     instance = new("/this/is/a/fake/path")
     instance.content = content
     instance
@@ -37,7 +37,7 @@ class Puppet::FileServing::Content < Puppet::FileServing::Base
     @content
   end
 
-  def to_raw
+  def to_binary
     File.new(full_path, "rb")
   end
 end
--- a/lib/puppet/network/formats.rb
+++ b/lib/puppet/network/formats.rb
@@ -64,10 +64,7 @@ end
 
 Puppet::Network::FormatHandler.create(:s, :mime => "text/plain", :extension => "txt")
 
-Puppet::Network::FormatHandler.create(:binary, :mime => "application/octet-stream")
-
-# A very low-weight format so it'll never get chosen automatically.
-Puppet::Network::FormatHandler.create(:raw, :mime => "application/x-raw", :weight => 1) do
+Puppet::Network::FormatHandler.create(:binary, :mime => "application/octet-stream", :weight => 1) do
   def intern_multiple(klass, text)
     raise NotImplementedError
   end
--- a/lib/puppet/network/formats.rb
+++ b/lib/puppet/network/formats.rb
@@ -64,6 +64,8 @@ end
 
 Puppet::Network::FormatHandler.create(:s, :mime => "text/plain", :extension => "txt")
 
+Puppet::Network::FormatHandler.create(:binary, :mime => "application/octet-stream")
+
 # A very low-weight format so it'll never get chosen automatically.
 Puppet::Network::FormatHandler.create(:raw, :mime => "application/x-raw", :weight => 1) do
   def intern_multiple(klass, text)
--- a/spec/unit/file_bucket/file_spec.rb
+++ b/spec/unit/file_bucket/file_spec.rb
@@ -9,19 +9,19 @@ describe Puppet::FileBucket::File, :uses_checksums => true do
   # this is the default from spec_helper, but it keeps getting reset at odd times
   let(:bucketdir) { Puppet[:bucketdir] = tmpdir('bucket') }
 
-  it "defaults to serializing to `:s`" do
-    expect(Puppet::FileBucket::File.default_format).to eq(:s)
+  it "defaults to serializing to `:binary`" do
+    expect(Puppet::FileBucket::File.default_format).to eq(:binary)
   end
 
   it "accepts s" do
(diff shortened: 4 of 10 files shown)


lib/puppet/file_bucket/file.rb
```diff
@@ -79,6 +78,10 @@ def self.from_s(contents)
     self.new(contents)
   end
 
+  def self.from_binary(contents)
+    self.new(contents)
+  end
+
```
joshcooper: since we're removing `:s` as a supported format, I think we should remove the `from_s` and `to_s` methods as well, and rename the `to_s` method to `to_binary` for `StringContents` and `FileContents`

lib/puppet/network/formats.rb
```diff
@@ -64,6 +64,8 @@ def supported?(klass)
 
 Puppet::Network::FormatHandler.create(:s, :mime => "text/plain", :extension => "txt")
 
+Puppet::Network::FormatHandler.create(:binary, :mime => "application/octet-stream")
+
 # A very low-weight format so it'll never get chosen automatically.
 Puppet::Network::FormatHandler.create(:raw, :mime => "application/x-raw", :weight => 1) do
   def intern_multiple(klass, text)
```
joshcooper: I think we should also remove `:raw` format, and change `Puppet::FileServing::Content` to use `:binary` as its supported format, and rename its `from_raw` & `to_raw` methods to `from_binary` & `to_binary`.

lib/puppet/file_bucket/file.rb
dankreek: Removing `from_s` makes sense to me, but `to_s` is core Ruby functionality, and a ton of tests seem to rely on Ruby automagically turning the `Puppet::FileBucket::File` object into a string.

joshcooper: yeah, sounds good. Maybe make a comment above the to_s methods that they are deprecated, `# @deprecated` so we can remove another day

dankreek: Cool, I also reworked `.to_s` to directly call `to_binary` to make it a little more obvious

spec/unit/file_bucket/file_spec.rb
```diff
@@ -9,19 +9,19 @@
...
   end
 
   it "accepts s" do
-    expect(Puppet::FileBucket::File.supported_formats).to include(:s)
+    expect(Puppet::FileBucket::File.supported_formats).to include(:binary)
   end
 
   describe "making round trips through network formats" do
     with_digest_algorithms do
       it "can make a round trip through `s`" do
```
camlow325: Should this say *...through `binary`* now?

dankreek:
All comments have been addressed so far but I would to squash things together into a more logical ordering/grouping before merging if everyone's ok with it now.

lib/puppet/file_serving/content.rb
```diff
@@ -39,7 +39,7 @@ def content
     @content
   end
 
-  def to_raw
+  def to_binary
     File.new(full_path, "rb")
```
joshcooper: we should probably change this to `Puppet::FileSystem::binread` like the file_bucket/file above.

lib/puppet/file_bucket/file.rb
```diff
@@ -138,7 +137,7 @@ def checksum_data(base_method)
       Puppet::Util::Checksums.method(:"#{base_method}_file").call(@path)
     end
 
-    def to_s
+    def to_binary
       Puppet::FileSystem::binread(@path)
```
joshcooper: One side effect of calling `binread` or `File.new(path, 'rb')` is that the encoding of the returned string will be `ASCII-8BIT`. But I don't know if any of our tests actually verify it. For example, if `binread` started returning a string in `UTF-8`, then ruby's net/http code might (not sure) corrupt the data when it serializes it.

lib/puppet/file_serving/content.rb
dankreek: Hmm, according to the corresponding spec test `spec/unit/file_serving/content_spec.rb` the expected return value is an open file and not a binary string.

joshcooper: Doh, sorry. The file handle is necessary so that we can stream file content.

spec/unit/network/formats_spec.rb
```diff
@@ -154,36 +164,6 @@ def to_pson(*args)
...
-      @format.should be_supported(String)
-    end
-
-    it "should fail if its multiple_render method is used" do
-      lambda { @format.render_multiple("foo") }.should raise_error(NotImplementedError)
-    end
-
-    it "should fail if its multiple_intern method is used" do
-      lambda { @format.intern_multiple(String, "foo") }.should raise_error(NotImplementedError)
-    end
```
joshcooper: Do we want to preserve these two tests for the `binary` format?

dankreek: I went ahead and worked those in.

spec/unit/network/formats_spec.rb
```diff
@@ -130,6 +130,24 @@ def to_pson(*args)
...
   end
 
+  describe "binary" do
+    before do
+      @binary = Puppet::Network::FormatHandler.format(:binary)
+    end
+
+    it "should have its mimetype set to application/octet-stream" do
+      @binary.mime.should == "application/octet-stream"
+    end
```
kylog: Can we switch these to rspec 3 style, e.g.
```
expect(@binary.mime).to eq("application/octet-stream")
```

And similarly for the spec changes below.

spec/unit/file_serving/content_spec.rb
```diff
@@ -15,7 +15,7 @@
   end
 
   it "should only support the raw format" do
```
camlow325: s/raw/binary

spec/unit/file_serving/content_spec.rb
```diff
@@ -56,7 +56,7 @@
   end
 
   it "should be able to create a content instance from raw file contents" do
```
camlow325: s/raw/binary

spec/unit/file_serving/content_spec.rb
```diff
@@ -56,7 +56,7 @@
...
-    Puppet::FileServing::Content.should respond_to(:from_raw)
+    Puppet::FileServing::Content.should respond_to(:from_binary)
   end
 
   it "should create an instance with a fake file name and correct content when converting from raw" do
```
camlow325: s/raw/binary

spec/unit/network/formats_spec.rb
```diff
@@ -130,6 +130,24 @@ def to_pson(*args)
...
+
+    it "should fail if its multiple_render method is used" do
```
camlow325: I noticed that the "should always be supported" test wasn't ported over from the `:raw` block to here.  Does that test still make sense to include for `:binary`?

camlow325: Same question for the "should have a weight of 1" test.

camlow325:
I noticed with HTTP tracing on an agent run that some requests that formerly included "raw" in the Accept header now include "binary".  I suppose the way Puppet uses the Accept header is kinda bogus anyway - see "s", for example.  Since "binary" doesn't reasonably approximate a known MIME type, though, would this change be problematic?  Should we be trying to get this to have "application/octet-stream" instead?

camlow325:
I did a few manual file_bucket requests with the code in this branch and everything looked good for the Content-Type header - using "application/octet-stream" in the PUT request and response for a GET to pull the file back down.  Also, the binary content of the file was preserved as stored in the file bucket and as retrieved back to the local file system via a GET.  Also, the md5sum on the response and in the file stored in the bucket looked good.

camlow325:
I ran into one problem in local testing around the file_content endpoint.  I created a simple module which included a file resource with a "source" parameter pointing back to a file stored in the module:

~~~puppet
class catalog_zero1 {
  file { "/tmp/catalog-zero1-impl24.txt":
    ensure => file,
    source => "puppet:///modules/catalog_zero1/catalog-zero1-impl24.txt",
  }
}
~~~

With http_debug set to true for the agent run, I saw that the agent was including "raw" in the Accept header rather than the "binary" that I'd expected:

~~~
GET /puppet/v3/file_content/modules/catalog_zero1/catalog-zero1-impl24.txt?environment=production& HTTP/1.1\r\nAccept: raw\r\n...
~~~

This caused the request to the master to fail with:

~~~
HTTP/1.1 406 Not Acceptable \r\n"
...
"Not Acceptable: No supported formats are acceptable (Accept: raw)"
~~~

Do we need to include some other change in order to cause the agent to include "binary" in the Accept header for file_content?

dankreek:
Oof, good catch, @camlow325 . I'll investigate the `file_content` endpoint here as well.

spec/unit/network/formats_spec.rb
dankreek: I removed this test and restored the old `:raw` test and just updated it to support the `:binary` format instead.

dankreek:
@jb I fixed that problem you found with the `accept: raw` header and it's fixed now.

camlow325:
@dankreek Confirmed locally that your latest fix addressed the file_content / accept: raw problem.

@joshcooper The only remaining question I'd had on this was whether the conversion from `accept: raw` to `accept: binary` in the HTTP request is acceptable.  I'm inclined to just go with this since it doesn't appear that we've maintained strict adherence to using industry-standard MIME types here in the past.  Do you agree?

@joshcooper @kylog Any last concerns on this PR overall?  If we don't hear anything from you today, we'll plan on merging this and can follow-up on new tickets for any additional feedback that needs to be addressed.

joshcooper:
@dankreek @camlow325 So when sending an HTTP request, we send an `Accept: binary` header due to https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/rest.rb#L46. But the server doesn't actually use the header when deciding what format to return the result, presumably because the `file_content` endpoint is hardwired to always return binary content?

If so, I agree it is a different issue than this PR. But I'd like to make sure we have a ticket filed to fix this behavior. I'm not as concerned about the `file_content` case, but for other endpoints were we might have multiple supported formats (and the server needs to choose one that is acceptable to the client).

camlow325:
> But the server doesn't actually use the header when deciding what format to return the result, presumably because the file_content endpoint is hardwired to always return binary content?

I believe the server does use it at least to the extent of validating that the client has included the only supported format that the file_* endpoints would claim to support, binary.  If the client doesn't include this as one entry in its Accept header, an HTTP 406 (Not Acceptable) is returned.  That doesn't necessarily seem like a bad behavior - although arguably the server could just return the payload in the one format it does support, tag the Content-Type as such, and let the consumer decide if the Content-Type is one it knows how to handle.

> I'm not as concerned about the file_content case, but for other endpoints were we might have multiple supported formats (and the server needs to choose one that is acceptable to the client).

Are there cases today that this applies to that you know of?  If we were going to file a new ticket about this, I think it would be good to file it in the context of an endpoint that has to make a decision about different types of formats it needs to return and whether or not there is a problem with doing so.

---

So it sounds like you are okay with the new behavior this PR would introduce where a client would need to include "binary" in the Accept header instead of "raw" for any of the file_* endpoints.  Or did I read that wrong?

camlow325:
Since it sounds like we're basically okay with the changes here, I'm going to go ahead and merge this.  We can sort out any follow-up work in later tickets, as needed.

joshcooper:
:+1:

PR #3514, opened 2015-01-23
(PUP-3902) Remove duplicate key from main settings

mcarpenter:
Redundant stanza causes warning under ruby 2.2.0:

    # puppet
    /opt/puppet/lib/puppet/defaults.rb:465: warning: duplicated key at line 466 ignored: :queue_type

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -463,10 +463,6 @@ module Puppet
       :default    => "stomp",
       :desc       => "Which type of queue to use for asynchronous processing.",
     },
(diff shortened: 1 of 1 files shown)


MikaelSmith:
You've targeted this change against puppetlabs:master rather than puppetlabs:stable. I don't think that was your intention.

mcarpenter:
Arg, apparently the target branch cannot be changed https://github.com/isaacs/github/issues/18. And the original still doesn't reflected my forced push (and I can't reopen it in any case). 3rd time lucky maybe...

PR #3515, opened 2015-01-23
(PUP-3855) Restore ability for auth.conf to restrict on environment

camlow325:
The PR for PUP-3644 moved the `check_authorization` call for
indirected routes to a spot before the environment for a request is
translated into a `Puppet::Node::Environment` instance in the request
params.  This caused any restrictions on environment made in an
auth.conf file to inappropriately reject all requests matching the
rule path, even ones that should have been allowed.

This PR moves the `check_authorization` call back to a spot in the
code after the translation to an `Puppet::Node:::Environment` instance
occurs.  This change restores the ability for an environment value in
the auth.conf to appropriately allow/deny access for requests.

### DIFF
--- a/lib/puppet/network/http/api/indirected_routes.rb
+++ b/lib/puppet/network/http/api/indirected_routes.rb
@@ -87,14 +87,16 @@ class Puppet::Network::HTTP::API::IndirectedRoutes
     end
 
     configured_environment = Puppet.lookup(:environments).get(environment)
-    if configured_environment.nil?
-      raise ArgumentError, "Could not find environment '#{environment}'"
-    else
+    unless configured_environment.nil?
       configured_environment = configured_environment.override_from_commandline(Puppet.settings)
       params[:environment] = configured_environment
     end
 
     check_authorization(method, "#{url_prefix}/#{indirection_name}/#{key}", params)
+
+    if configured_environment.nil?
+      raise ArgumentError, "Could not find environment '#{environment}'"
+    end
 
     params.delete(:bucket_path)
 
--- a/lib/puppet/network/http/api/indirected_routes.rb
+++ b/lib/puppet/network/http/api/indirected_routes.rb
@@ -72,8 +72,6 @@ class Puppet::Network::HTTP::API::IndirectedRoutes
       raise ArgumentError, "Indirection '#{indirection_name}' does not match url prefix '#{url_prefix}'"
     end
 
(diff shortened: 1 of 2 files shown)


lib/puppet/network/http/api/v3/indirected_routes.rb
```diff
@@ -88,6 +87,8 @@ def uri2indirection(http_method, uri, params)
       params[:environment] = configured_environment
     end
 
+    check_authorization(method, "#{self.class.url_prefix}/#{indirection_name}/#{key}", params)
+
```
joshcooper: Previously we checked authorization fairly early on, but now we have work to do to lookup the environment object. If there is a problem we raise an exception, e.g. ["Could not find environment"](https://github.com/camlow325/puppet/blob/bug/master/PUP-3855-auth-conf-not-honoring-environment/lib/puppet/network/http/api/v3/indirected_routes.rb#L84) which would allow an unauthenticated user the ability to enumerate environment names. Should we always raise not authorized instead?

camlow325: This doesn't appear to be a new issue since the same code path appears to be in place on "stable" -- https://github.com/puppetlabs/puppet/blob/stable/lib/puppet/network/http/api/v1.rb#L68 -- but I agree this would be a good idea.  The information disclosure we'd allow for otherwise seems inappropriate.  I'll pursue that if there are no other objections...

camlow325:
I rebased this PR due to conflicting changes that had been merged.

I added a new commit, 367a728, which attempts to address the concern that @joshcooper raised about the inappropriate information disclosure which would occur if an unauthenticated/unauthorized user were to reference a non-existent environment in a request.  I modified the code to go through the `check_authorization` path so that the failure message for an invalid environment would be identical to the "Forbidden request" that `check_authorization` raises for any sort of failure.

I could have mimicked that message inline in `uri2indirection` but that didn't seem right because even a subtle difference in the message content between the two paths could tip off an untrusted user about the validity of individual environments.  Also, this ensures that as any changes to the content of the message that the `check_authorization` path uses will be picked up for the bad environment case.

rlinehan:
+1 This looks good to me, and I tested it locally and it fixes the bug. I agree that going through `check_authorization` even if the environment is not present is the best solution to the issue of potentially leaking environment names.

cprice404:
+1 looks fine to me.  @joshcooper do you want to take another quick look before I hit the button?

joshcooper:
sorry just saw this, :+1:

PR #3517, opened 2015-01-23
(PUP-3897) Add acceptance: ensure new environment picked up by agent

er0ck:
in PUP-3755 there are two scenarios that lead to a newly classified
agent not picking up its new environment.  This test will catch the case
in which a broken but non-assigned environment is parsed and causes
errors or warnings.

### DIFF
--- a/acceptance/tests/environment/broken_unassigned_environment_handled_gracefully.rb
+++ b/acceptance/tests/environment/broken_unassigned_environment_handled_gracefully.rb
@@ -1,0 +1,54 @@
+test_name 'PUP-3755 Test an un-assigned broken environment'
+
+step 'setup environments'
+
+testdir = create_tmpdir_for_user(master, 'confdir')
+environment = 'debug'
+manifest = <<-MANIFEST
+  File {
+    ensure => directory,
+    owner => #{master['user']},
+    group => #{master['group']},
+    mode => "0750",
+  }
+
+  file { "#{testdir}":;
+    "#{testdir}/environments":;
+    "#{testdir}/environments/production":;
+    "#{testdir}/environments/production/manifests":;
+    "#{testdir}/environments/production/modules":;
+    "#{testdir}/environments/#{environment}":;
+    "#{testdir}/environments/#{environment}/manifests":;
+    "#{testdir}/environments/#{environment}/modules":;
+  }
+  # broken envioronment
+  file { "#{testdir}/environments/production/manifests/site.pp":
+    ensure  => file,
+    content => 'import "/tmp/bogus/*.pp"'
+  }
+  file { "#{testdir}/environments/#{environment}/manifests/site.pp":
+    ensure  => file,
+    content => 'node default{\nnotify{"you win":}\n}'
+  }
+MANIFEST
+
+apply_manifest_on(master, manifest, :catch_failures => true)
+
+step 'run agents, ensure no one complains about the other environment'
+
+master_opts = {
+  'main' => {
+    'environmentpath' => "#{testdir}/environments"
+  }
+}
+
+with_puppet_running_on(master, master_opts, testdir) do
+  agents.each do |agent|
+    on(agent, puppet('agent',
+                     "--test --server #{master} --environment #{environment}"),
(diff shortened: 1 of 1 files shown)


joshcooper:
@er0ck looks good, on commit 5818ce69f5205f4022f5a7641c35c75cdd9173df, without the second fix, the test detects the issue:

<pre>
jc7y049n7hvlmeh.delivery.puppetlabs.net (master) 16:04:46$  env PATH="/usr/bin:/opt/puppet-git-repos/hiera/bin:${PATH}" RUBYLIB="/opt/puppet-git-repos/hiera/lib:/opt/puppet-git-repos/hiera-puppet/lib:${RUBYLIB}" puppet agent  --test --server jc7y049n7hvlmeh.delivery.puppetlabs.net --environment debug
...
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not parse for environment production: No file(s) found for import of '/tmp/bogus/*.pp' at /tmp/confdir.0pH9zM/environments/production/manifests/site.pp:1
</pre>

On commit 3e8e8420095ed6f0c399301cd54f745b83bae8fd, it passes:

<pre>
gj0o30tal9m1itu.delivery.puppetlabs.net (agent) 16:02:48$  env PATH="/usr/bin:/opt/puppet-git-repos/hiera/bin:${PATH}" RUBYLIB="/opt/puppet-git-repos/hiera/lib:/opt/puppet-git-repos/hiera-puppet/lib:${RUBYLIB}" puppet agent  --test --server hgyg2bg8ojylt0l.delivery.puppetlabs.net --environment debug
...
Notice: /Stage[main]/Main/Node[default]/Notify[you win]/message: defined 'message' as 'you win'
</pre>

My only question is whether this should be merged to stable or master?

er0ck:
yeah, PR'd against master because i didn't think it was worth folks up-merging separately to master. and we're fairly safe with recent manual tests.  only exposure in stable is 3.7.5 which has a few unrelated tickets?

joshcooper:
sounds good

joshcooper:
The test passes in master, but I'm having trouble finding an earlier revision where it fails. @hlindberg do you know of a commit in master where this problem would have shown up?

hlindberg:
@joshcooper before PUP-3436 was added? (It added optimization that stopped calling logic that required access to details in types and thus after the fix did not trigger the faulty behavior. The merge commit of PUP-3436 is 2d0e0a6b97ceec23b217304226c2e7fbc848cefa

hlindberg:
Have not looked at this test logic again but at first read through it did not seem to be able to observe the problem. The issue was not that it ended up with the wrong environment in the resulting catalog - the problem was that it loads the default environment, and if that could not be loaded due to errors in manifests say, the operation mysteriously would fail. Other potential problems would be that the referenced type is not found, or that the found type would be incompatible with the type in the wanted environment (same for anything loaded actually).

joshcooper:
It looks like this issue existed in the master branch up to and including 30b2b5d:

<pre>
$ bundle exec rake ci:test:packages SHA=30b2b5dea2ce7854bd610f2fa8ac3dbd3c7348d6 CONFIG=config/nodes/rhel7.yaml TESTS=tests/environment/broken_unassigned_environment_handled_gracefully.rb
...
mhm3mtvvme9a896.delivery.puppetlabs.net (master) 10:51:16$  env PATH="/usr/bin:/opt/puppet-git-repos/hiera/bin:${PATH}" RUBYLIB="/opt/puppet-git-repos/hiera/lib:/opt/puppet-git-repos/hiera-puppet/lib:${RUBYLIB}" puppet agent  --test --server mhm3mtvvme9a896.delivery.puppetlabs.net --environment debug
...
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not parse for environment production: No file(s) found for import of '/tmp/bogus/*.pp' at /tmp/confdir.E3C5TP/environments/production/manifests/site.pp:1 on node mhm3mtvvme9a896.delivery.puppetlabs.net
</pre>

The issue was fixed in bf80651, which removed the 3x AST for PUP-3274 and PUP-3638.

PR #3519, opened 2015-01-26
(PUP-3894) Update license in PMT generate example to match SPDX string

andersonmills:
Before this commit, the "examples" string in the generate face still
    read "Apache 2.0". After this commit, it reads "Apache-2.0" to be in
    line with the behavior of puppet module generate.


adrienthebo:
:+1:

adrienthebo:
Merged in 215b5d2.

PR #3039, opened 2014-09-02
(PUP-3168) Fix all file permissions on puppet module install

vholer:
If directory permissions are fixed to 755, file permissions
should be fixed in same way. Otherwise module files with 600
permissions (e.g. ceritsc-psacct) aren't available to real
Puppet user (pe-apache?) different from current user.

See
https://tickets.puppetlabs.com/browse/PUP-3168
https://github.com/CERIT-SC/puppet-psacct/issues/1

PR #3520, opened 2015-01-26
(PUP-3168) Fix file permissions on puppet module install

andersonmills:
Before this commit, file permissions on installing a module only removed
the setuid and sticky bit permissions. After this commit, file
permissions are added for read and write for the executor of the puppet
module install and read for group in addition.


andersonmills:
This is supplants PR 3039 - https://github.com/puppetlabs/puppet/pull/3039.

adrienthebo:
Merged in cc51668.

PR #3521, opened 2015-01-26
maint/add-legacy-formatters

kylog:


kylog:
This passes at my desk, and I've pushed the gem to the mirror. Let's see if travis likes it.

/cc @joshcooper @MikaelSmith

PR #3523, opened 2015-01-27
(PUP-3851) Remove v2.0 endpoints

rlinehan:
Remove the v2.0 endpoints (the only one of which is `/v2.0/environmnets`),
since there is now a prefix before all routes, so it is not backwards
compatible.


cprice404:
+1 from me.  @hlindberg any interest in reviewing this before we merge it?

hlindberg:
+1 I see what you are doing - looks clear.

PR #3524, opened 2015-01-27
maint/update-yarjuf-for-rspec3-cleanliness

kylog:


joshcooper:
I updated stickler with both 2.0.0 and 1.0.6

PR #3526, opened 2015-01-27
(PUP-3861) Parse multi-line yum version information

joshcooper:
Commit 101d1deb11 removed the python bindings to interact with yum, but
did not retain the previous behavior of handling long package names,
because `yum check-update` can return output like the following:

    libnl.x86_64                           1.1.4-2.el6               base
    libpcap.x86_64                         14:1.4.0-1.20130826git2dbcaa1.el6
                                                                     base
    libproxy.x86_64                        0.3.0-4.el6_3             base

This commit modifies the yum provider to parse either single or
multiline output, and adds corresponding tests.


shrug:
:+1:

ScottGarman:
:+1:

PR #3528, opened 2015-01-28
(PUP-3889) Fix checksums writing files

MikaelSmith:
When checking whether the local file must be modified, ensure the metadata
includes the requested checksum, rather than defaulting to md5.

Streaming checksums are compared to file checksums, so they must behave
the same way on a stream that is being written to a file.

### DIFF
--- a/lib/puppet/util/checksums.rb
+++ b/lib/puppet/util/checksums.rb
@@ -56,11 +56,10 @@ module Puppet::Util::Checksums
     sha256_file(filename, true)
   end
 
-  def sha256_stream(&block)
+  def sha256_stream(lite = false, &block)
     require 'digest/sha2'
     digest = Digest::SHA256.new
-    yield digest
-    digest.hexdigest
+    checksum_stream(digest, block, lite)
   end
 
   def sha256_hex_length
@@ -68,7 +67,7 @@ module Puppet::Util::Checksums
   end
 
   def sha256lite_stream(&block)
-    sha256_stream(&block)
+    sha256_stream(true, &block)
   end
 
   def sha256lite_hex_length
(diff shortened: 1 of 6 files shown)


MikaelSmith:
@joshcooper another part of PUP-1208

lib/puppet/util/checksums.rb
```diff
@@ -198,6 +195,29 @@ def none(content)
...
+    # Provide an interface for digests. If lite, only digest the first 512 bytes
+    def <<(str)
+      if @lite
+        if (str.length + @bytes) < 512
+          @digest << str
+          @bytes += str.length
+        elsif @bytes < 512
+          @digest << str[0, 512-@bytes]
+          @bytes = 512
+        end
```
joshcooper: Something doesn't look right with the `elsif` case. If we're doing md5lite, I think there would be an else case that ignored `str` once `@bytes == 512`. Perhaps something like this?

```
if @bytes < 512
  if (str.length + @bytes) <= 512
    buf = str
  else
    buf = str[0, 512 - @bytes]
  end

  @digest << buf
  @bytes += buf.length
end
```

In the first `if` case, if the `str.length` + current position is less than or *equal* to the max length, then we can process the entire `str`. Else we'd be exceeding the max length, and would need to substring (copy) the buffer.

Alternatively, we could always substring (as it's only the lite case):

```
if @bytes < 512
  buf = str[0, 512 - @bytes]
  @digest << buf
  @bytes += buf.length
end
```

MikaelSmith: I think my code and yours are roughly equivalent (if str.length + @bytes == 512 and @bytes < 512 in my case, it falls through to the elsif and does an unnecessary substring (which should be a constant-time operation).

I like your last option though, that should work as well and be simpler to understand.

PR #3534, opened 2015-01-28
(PUP-3915) Remove win32console

ferventcoder:
With Ruby 2.1.5, all consoles support ansicolor, so it removes a need
to determine `console_has_color?` and further removes the need for
win32console. Simplify the logic assuming that all consoles support
color.

This is likely to fail, plus it blocked on Windows movement to Ruby 2.1.5. This code needs to be held until we can move to 2.1.5 on both x86 and x64.


Iristyle:
#3540 should be merged first

Iristyle:
We're also going to want to merge #3562 prior to this as well

ferventcoder:
\o/

PR #3535, opened 2015-01-29
(PUP-3890) Fix mtime/ctime checksum type

MikaelSmith:
Makes file contents that use mtime or ctime for checksum_type
compare to ensure the target mtime/ctime is newer than the source.

### DIFF
--- a/lib/puppet/type/file/content.rb
+++ b/lib/puppet/type/file/content.rb
@@ -1,6 +1,7 @@
 require 'net/http'
 require 'uri'
 require 'tempfile'
+require 'date'
 
 require 'puppet/util/checksums'
 require 'puppet/network/http'
@@ -73,19 +74,6 @@ module Puppet
       end
     end
 
-    def checksum_type
-      if source = resource.parameter(:source)
-        result = source.checksum
-      else
-        result = resource[:checksum]
-      end
-      if result =~ /^\{(\w+)\}.+/
-        return $1.to_sym
-      else
-        return result
-      end
-    end
-
     def length
       (actual_content and actual_content.length) || 0
     end
@@ -117,6 +105,21 @@ module Puppet
         end
       end
       result
+    end
+
+    def property_matches?(current, desired)
+      basic = super
+      # The inherited equality is always accepted, so use it if valid.
+      time_types = [:mtime, :ctime]
+      checksum_type = resource.parameter(:checksum).value
+      return basic if basic || !time_types.include?(checksum_type)
+      return false unless current && desired
+      begin
+        raise if !time_types.include?(sumtype(current).to_sym) || !time_types.include?(sumtype(desired).to_sym)
+        DateTime.parse(sumdata(current)) >= DateTime.parse(sumdata(desired))
+      rescue => detail
+        self.fail Puppet::Error, "Resource with checksum_type #{checksum_type} didn't contain a date in #{current} or #{desired}", detail.backtrace
+      end
     end
 
     def retrieve
--- a/spec/integration/type/file_spec.rb
+++ b/spec/integration/type/file_spec.rb
@@ -686,8 +686,10 @@ describe Puppet::Type.type(:file), :uses_checksums => true do
     describe "when recursing remote directories" do
       describe "for the 2nd time" do
         with_checksum_types "one", "x" do
-          it "should not update the target directory" do
-            options = {
+          let(:target_file) { File.join(path, 'x') }
+          let(:second_catalog) { Puppet::Resource::Catalog.new }
+          before(:each) do
+            @options = {
               :path => path,
               :ensure => :directory,
               :backup => false,
@@ -695,21 +697,44 @@ describe Puppet::Type.type(:file), :uses_checksums => true do
               :checksum => checksum_type,
               :source => env_path
             }
-            target_file = File.join(path, 'x')
+          end
 
-            first_catalog = Puppet::Resource::Catalog.new
-            first_catalog.add_resource Puppet::Type.send(:newfile, options)
-            first_catalog.apply
+          it "should not update the target directory" do
+            # Ensure the test believes the source file was written in the past.
+            FileUtils.touch checksum_file, :mtime => Time.now - 20
+            catalog.add_resource Puppet::Type.send(:newfile, @options)
+            catalog.apply
             expect(File).to be_directory(path)
             expect(Puppet::FileSystem.exist?(target_file)).to be_truthy
 
             # The 2nd time the resource should not change.
-            second_catalog = Puppet::Resource::Catalog.new
-            second_catalog.add_resource Puppet::Type.send(:newfile, options)
+            second_catalog.add_resource Puppet::Type.send(:newfile, @options)
             result = second_catalog.apply
             status = result.report.resource_statuses["File[#{target_file}]"]
             expect(status).not_to be_failed
             expect(status).not_to be_changed
+          end
+
+          it "should update the target directory if contents change" do
+            pending "a way to appropriately mock ctime checks for a particular file" if checksum_type == 'ctime'
+
+            catalog.add_resource Puppet::Type.send(:newfile, @options)
+            catalog.apply
+            expect(File).to be_directory(path)
+            expect(Puppet::FileSystem.exist?(target_file)).to be_truthy
+
+            # Change the source file.
+            File.open(checksum_file, "wb") { |f| f.write "some content" }
+            File::Stat.any_instance.unstub(:mtime)
+            File::Stat.any_instance.unstub(:ctime)
+            FileUtils.touch target_file, :mtime => Time.now - 20
+
+            # The 2nd time should update the resource.
+            second_catalog.add_resource Puppet::Type.send(:newfile, @options)
+            result = second_catalog.apply
+            status = result.report.resource_statuses["File[#{target_file}]"]
+            expect(status).not_to be_failed
+            expect(status).to be_changed
           end
         end
       end
--- a/spec/integration/type/file_spec.rb
+++ b/spec/integration/type/file_spec.rb
@@ -725,8 +725,6 @@ describe Puppet::Type.type(:file), :uses_checksums => true do
 
             # Change the source file.
             File.open(checksum_file, "wb") { |f| f.write "some content" }
-            File::Stat.any_instance.unstub(:mtime)
-            File::Stat.any_instance.unstub(:ctime)
             FileUtils.touch target_file, :mtime => Time.now - 20
 
             # The 2nd time should update the resource.
--- a/spec/unit/type/file/content_spec.rb
+++ b/spec/unit/type/file/content_spec.rb
@@ -24,29 +24,6 @@ describe Puppet::Type.type(:file).attrclass(:content), :uses_checksums => true d
     end
   end
 
-  describe "when determining the checksum type" do
-    let(:content) { described_class.new(:resource => resource) }
-
-    it "should use the type specified in the source checksum if a source is set" do
-      resource[:source] = File.expand_path("/foo")
-      resource.parameter(:source).expects(:checksum).returns "{md5lite}eh"
-
-      expect(content.checksum_type).to eq(:md5lite)
-    end
-
-    it "should use the type specified by the checksum parameter if no source is set" do
-      resource[:checksum] = :md5lite
-
-      expect(content.checksum_type).to eq(:md5lite)
-    end
-
-    with_digest_algorithms do
-      it "should use the type specified by digest_algorithm by default" do
-        expect(content.checksum_type).to eq(digest_algorithm.intern)
-      end
-    end
-  end
-
   describe "when determining the actual content to write" do
     let(:content) { described_class.new(:resource => resource) }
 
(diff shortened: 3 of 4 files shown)


joshcooper:
Does this PR depend on the server-side PR being merged first? I ask because when puppet creates a file, it does so using the default `md5`, and when it updates the file, it thinks the `current` checksum is `{ctime}2015-01-29 08:04:38 -0800'`

<pre>
0 ~/work/puppet (pr/3535)  $ git rev-parse HEAD
ea56969c70d7380d2637c30f0568449fb4f4ea83
0 ~/work/puppet (pr/3535)  $ rm /tmp/foo /tmp/source.txt
0 ~/work/puppet (pr/3535)  $ touch /tmp/source.txt
0 ~/work/puppet (pr/3535)  $ be puppet apply -e "file { '/tmp/foo': ensure => file, checksum => ctime, source => '/tmp/source.txt' }"
Notice: /Stage[main]/Main/File[/tmp/foo]/ensure: defined content as '{md5}d41d8cd98f00b204e9800998ecf8427e'
0 ~/work/puppet (pr/3535)  $ ls > /tmp/source.txt
0 ~/work/puppet (pr/3535)  $ be puppet apply -e "file { '/tmp/foo': ensure => file, checksum => ctime, source => '/tmp/source.txt' }"
Notice: /Stage[main]/Main/File[/tmp/foo]/content: content changed '{ctime}2015-01-29 08:04:38 -0800' to '{md5}4cc2c5f3a375af73bae0ffc6c279d8a4'
</pre>

lib/puppet/type/file/content.rb
```diff
@@ -119,6 +120,11 @@ def insync?(is)
       result
     end
 
+    def property_matches?(current, desired)
+      return super unless [:mtime, :ctime].include?(checksum_type)
+      DateTime.parse(current) >= DateTime.parse(desired)
```
joshcooper: I think `current` can be `nil` and/or `:absent`. The file `mode` property has some similar logic: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/mode.rb#L117-L122

I wonder what happens if you say `ensure => absent`, does the `property_matches?` method get called? For a "normal" type I would say no, I would expect only the providers' `destroy` method to be called. But the file type has intertwined properties and parameters, so I'm not sure.

Do we know that `current` and `desired` are always in a specific date-time format? I think these strings are generated via `File.stat(path).mtime`, with an implicit `Time#to_s` somewhere along the way. The `current` time is generated on the local system and `desired` on the (potentially) remote system. Hopefully ruby is consistent across platforms (windows agent & linux master) and versions (1.9.3 vs 2.2.0)?

Our `file_metadata` schema only says the checksum value is a string: https://github.com/puppetlabs/puppet/blob/master/api/schemas/file_metadata.json#L31-L40. Also it looks like the schema is missing some checksum types (`ctime` and `*lite`).

MikaelSmith: I'll look at this more. I haven't checked whether Windows and Linux mtimes match.

When they're not convertible to a time an exception is thrown, so I should probably catch that and have a fallback super comparison. The exception gets caught higher up, but will always assume the match failed.

MikaelSmith:
This PR depends on https://github.com/puppetlabs/puppet/pull/3528/ to be usable with either puppet apply or puppet agent. I had to force the specs pretty hard.

MikaelSmith:
Oh, it'll also need the fix at https://github.com/MikaelSmith/puppet/commit/93a2d2131eb83057aae190198fd475293e0fb79a, which I haven't written specs for yet. I could probably put all 3 of those into a single PR.

lib/puppet/type/file/content.rb
MikaelSmith: I don't think the super would ever return true however. If either of them doesn't have a time-based checksum, but checksum_type is mtime/ctime, then they're out of sync. That's the behavior you'll get right now.

MikaelSmith: mtime/ctime/Time.now all return the format "2015-01-29 11:40:47 -0800" on both Windows and Linux. DateTime.parse is also pretty flexible about parsing.

joshcooper:
I have the same comment on this PR too, about what was the old behavior and how this PR fixes it. Were we comparing current and desired as string objects? or were we comparing md5 against ctime/mtime?

lib/puppet/type/file/content.rb
```diff
@@ -119,6 +120,16 @@ def insync?(is)
       result
     end
 
+    def property_matches?(current, desired)
+      return super unless [:mtime, :ctime].include?(checksum_type)
+      begin
+        DateTime.parse(current) >= DateTime.parse(desired)
+      rescue => detail
+        Puppet.debug "resource with checksum_type #{checksum_type} didn't contain a date in #{current} or #{desired}"
+        super
```
joshcooper: I think we should explicitly check for `current` or `desired` being `:absent` or `nil` (whichever is the one that this method can be called with), and short-circuit the method. FWIW, `lib/puppet/type/file/mode.rb` only seems to handle `nil`, not `:absent`

I don't like calling into `DateTime.parse` and rescuing because there could be other reasons for failing to parse. Also, I think `DateTime.parse` exceptions should fail the resource.

spec/unit/type/file/content_spec.rb
```diff
@@ -227,6 +227,30 @@
           end
         end
       end
+
+      SAVED_TIME = Time.now
+      [:ctime, :mtime].each do |time_stat|
+        [["older", SAVED_TIME-1, false], ["same", SAVED_TIME, true], ["newer", SAVED_TIME+1, true]].each do
+          |compare, target_time, success|
```
joshcooper: it would be good to also test what happens when ctime/mtime checksum types are used when a file is created and deleted.

joshcooper: Also wondering if we need to account for ctime/mtime for directories and symlinks? Do they have the same meaning as files?

MikaelSmith: I believe so, but I'll check. Checksum may not be allowed on directories.

MikaelSmith: Directories and symlinks specifically don't cause an issue. However, testing that with recursive directory sourcing found another issue with the checksum_type not propagating correctly. I'll have to re-open or start a new subtask.

MikaelSmith: Hmm, may not need a new subtask. This task makes it an error because it throws when current isn't a datetime.

lib/puppet/type/file/content.rb
```diff
@@ -119,6 +120,18 @@ def insync?(is)
       result
     end
 
+    def property_matches?(current, desired)
+      basic = super
+      # The inherited equality is always accepted, so use it if valid.
+      return basic if basic || ![:mtime, :ctime].include?(checksum_type)
+      return false unless current && current != :absent && desired && desired != :absent
+      begin
+        DateTime.parse(current) >= DateTime.parse(desired)
```
MikaelSmith: This accepts a md5 SHA (why?) as a valid date, so I'm modifying it a bit. I'm also trying to add tests that have positive validation that an update will occur when ctime/mtime are different, but finding it difficult to stub ctime/mtime the way I need to.

MikaelSmith:
@joshcooper Updated.

lib/puppet/type/file/content.rb
```diff
@@ -119,6 +120,20 @@ def insync?(is)
       result
     end
 
+    def property_matches?(current, desired)
+      basic = super
+      # The inherited equality is always accepted, so use it if valid.
+      time_types = ["mtime", "ctime"]
+      return basic if basic || !time_types.include?(checksum_type.to_s)
```
joshcooper: It turns out that the `#checksum_type` is essentially dead code. It's only called here and in `content_spec`, and I think can be deleted. I think it's existence is due to the fact that the checksum type was not always set correctly due to all of the issues you've fixed.

If the file resource has a `source`, then the `checksum_type` and `checksum` are copied from the metadata to the `content` and `checksum` parameter values, and become the `desired` value:

```
(rdb:1) current
"{mtime}2015-02-03 23:34:26 -0800"
(rdb:1) desired
"{mtime}2015-02-03 21:58:49 -0800"
(rdb:1) resource.parameter(:content).value
"{mtime}2015-02-03 21:58:49 -0800"
(rdb:1) resource.parameter(:source).checksum
"{mtime}2015-02-03 21:58:49 -0800"
```

And the `checksum` parameter already contains what the `checksum_type` method was trying to do:

```
(rdb:1) resource.parameter(:checksum).value
:mtime
```

If the file resource has `content`, then the checksum is stored as the value of the `content` parameter, and it becomes the `desired` value:

```
(rdb:1) current
"{sha256}ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad"
(rdb:1) desired
"{sha256}88d4266fd4e6338d13b845fcf289579d209c897823b9217da3e161936f031589"
(rdb:1) resource.parameter(:content).value
"{sha256}88d4266fd4e6338d13b845fcf289579d209c897823b9217da3e161936f031589"
(rdb:1) resource.parameter(:source)
nil
```

And again the `checksum` parameter contains the expected value:

```
(rdb:1) resource.parameter(:checksum).value
:sha256
```

So I think we can delete `checksum_type`, and just call `resource.parameter(:checksum).value` directly.

lib/puppet/type/file/content.rb
```diff
@@ -119,6 +120,20 @@ def insync?(is)
...
+      return false unless current && desired
+      begin
+        raise "not a time checksum" if !time_types.include?(sumtype(current)) || !time_types.include?(sumtype(desired))
+        DateTime.parse(current) >= DateTime.parse(desired)
```
joshcooper: So `current` and `desired` are of the form `{mtime}2015-02-03 23:02:42 -0800`, and I have no idea how `DateTime.parse` is working, but that seems super suprising. It might be better to only parse the `sumdata`: 

```
self.fail "Not a time checksum #{current}" if !time_types.include?(sumtype(current))
self.fail "Not a time checksum #{desired}" if !time_types.include?(sumtype(desired))
begin
  DateTime.parse(sumdata(current)) >= DateTime.parse(sumdata(desired))
rescue => detail
```

Also self.fail (is mixed into each type) and provides context about the resource that generated the errored (both on the console and in the report's logs).

joshcooper:
I tested a bunch of scenarios and it seems to be working well:

https://gist.github.com/joshcooper/3a6ef632fcf35641a8ca
https://gist.github.com/joshcooper/44f7c6fe38551572e0c9
https://gist.github.com/joshcooper/0e7afb7bf720d0413408

I haven't tried remote sources or those with symlinks yet.

spec/integration/type/file_spec.rb
```diff
@@ -695,22 +697,43 @@ def build_path(dir)
...
+
+            catalog.add_resource Puppet::Type.send(:newfile, @options)
+            catalog.apply
+            expect(File).to be_directory(path)
+            expect(Puppet::FileSystem.exist?(target_file)).to be_truthy
+
+            # Change the source file.
+            File.open(checksum_file, "wb") { |f| f.write "some content" }
+            File::Stat.any_instance.unstub(:mtime)
+            File::Stat.any_instance.unstub(:ctime)
```
joshcooper: It would be good to eliminate the stubbing in the `with_checksum_types` shared context as that is only used in integration tests (so we should be able to create the necessary files). That might also fix the ctime issue (since we can delete the source file and recreate it, generating a new `ctime`)

lib/puppet/type/file/content.rb
MikaelSmith: self.fail instead of raising an error?

Yeah, just sumdata makes sense, parse is a bit rogue.

joshcooper: yep `self.fail` calls https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/errors.rb#L104, which will raise.

spec/integration/type/file_spec.rb
MikaelSmith: That's a little difficult to do correctly in a spec test, because in many cases ctime only has a resolution of 1 second: http://en.wikipedia.org/wiki/Stat_%28system_call%29#Granularity_of_mtime.2C_etc.

joshcooper: I'm thinking there has to be a way to write the integration tests so that they don't need to stub ctime, eg. to test the case where there should be no changes, we can retrieve the ctime from the file on disk. to test the case where there should be changed, delete and create the file with the same content.

MikaelSmith: I see what you mean, I have an idea I'll try.

MikaelSmith:
Updated to avoid needing to stub ctime/mtime.

lib/puppet/type/file/content.rb
```diff
@@ -119,6 +107,21 @@ def insync?(is)
       result
     end
 
+    def property_matches?(current, desired)
+      basic = super
+      # The inherited equality is always accepted, so use it if valid.
+      time_types = ["mtime", "ctime"]
+      checksum_type = resource.parameter(:checksum).value.to_s
+      return basic if basic || !time_types.include?(checksum_type)
```
joshcooper: nit, can we just compare symbols instead of converting to strings to do the comparison? In ruby, all occurrences of a symbol refer to the same instance, so it doesn't generate new objects.

MikaelSmith: Done.

PR #3536, opened 2015-01-29
(HI-328) Test against Hiera 2.x source

hlindberg:
Tests that the additonal recurse_guard argument added to the
Hiera backend lookup mehtod by HI-328 is propagated correctly.

This commit assumes that 2x hiera source is clone adjacent
to the puppet clone (Gemfile appoints ../hiera)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,8 @@ end
 
 gem "puppet", :path => File.dirname(__FILE__), :require => false
 gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 1.6', '< 3'])
-gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '~> 1.0')
+# gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '~> 1.0')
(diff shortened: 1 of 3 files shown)


hlindberg:
Hm, the commit needs to be relabeled PUP-3920

Gemfile
```diff
@@ -25,7 +25,8 @@ end
 
 gem "puppet", :path => File.dirname(__FILE__), :require => false
 gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 1.6', '< 3'])
-gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '~> 1.0')
+# gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '~> 1.0')
+gem "hiera", :path => "#{File.dirname(File.dirname(__FILE__))}/hiera", :require => false
```
kylog: Why not leave it the way it was (where HIERA_LOCATION could override a gem), but expand the gem range to be hiera 1.x or 2.x?

hlindberg: yeah ! Actually - we will not need these changes as we are instead going to delete the puppet_backend - it is a horrible thing anyway and creates a circular dpendency between puppet and hiera repos.

PR #3540, opened 2015-01-29
(PUP-3837) FFI Registry calls to work around Ruby 2.1.5 Windows bugs

Iristyle:
Ready to go.

### DIFF
--- a/lib/puppet/util/windows/api_types.rb
+++ b/lib/puppet/util/windows/api_types.rb
@@ -42,6 +42,7 @@ module Puppet::Util::Windows::APITypes
 
     alias_method :read_dword, :read_uint32
     alias_method :read_win32_ulong, :read_uint32
+    alias_method :read_qword, :read_uint64
 
     alias_method :read_hresult, :read_int32
 
@@ -52,10 +53,10 @@ module Puppet::Util::Windows::APITypes
     alias_method :read_wchar, :read_uint16
     alias_method :read_word,  :read_uint16
 
-    def read_wide_string(char_length)
+    def read_wide_string(char_length, dst_encoding = Encoding.default_external)
       # char_length is number of wide chars (typically excluding NULLs), *not* bytes
       str = get_bytes(0, char_length * 2).force_encoding('UTF-16LE')
-      str.encode(Encoding.default_external)
+      str.encode(dst_encoding)
     end
 
     def read_arbitrary_wide_string_up_to(max_char_length = 512)
@@ -139,6 +140,7 @@ module Puppet::Util::Windows::APITypes
   FFI.typedef :pointer, :lpcvoid
   FFI.typedef :pointer, :lpvoid
   FFI.typedef :pointer, :lpword
+  FFI.typedef :pointer, :lpbyte
   FFI.typedef :pointer, :lpdword
   FFI.typedef :pointer, :pdword
   FFI.typedef :pointer, :phandle
@@ -227,6 +229,18 @@ module Puppet::Util::Windows::APITypes
         Time.local(self[:wYear], self[:wMonth], self[:wDay],
           self[:wHour], self[:wMinute], self[:wSecond], self[:wMilliseconds] * 1000)
       end
+    end
+
+    # https://msdn.microsoft.com/en-us/library/windows/desktop/ms724284(v=vs.85).aspx
+    # Contains a 64-bit value representing the number of 100-nanosecond
+    # intervals since January 1, 1601 (UTC).
+    # typedef struct _FILETIME {
+    #   DWORD dwLowDateTime;
+    #   DWORD dwHighDateTime;
+    # } FILETIME, *PFILETIME;
+    class FILETIME < FFI::Struct
+      layout :dwLowDateTime, :dword,
+             :dwHighDateTime, :dword
     end
 
     ffi_convention :stdcall
--- a/lib/puppet/util/windows/registry.rb
+++ b/lib/puppet/util/windows/registry.rb
@@ -44,8 +44,10 @@ module Puppet::Util::Windows
       index = 0
       subkey = nil
 
+      subkey_max_len, value_max_len = reg_query_info_key_max_lengths(key.hkey)
+
       begin
-        subkey, filetime = reg_enum_key(key.hkey, index)
+        subkey, filetime = reg_enum_key(key.hkey, index, subkey_max_len)
         yield subkey, filetime if !subkey.nil?
         index += 1
       end while !subkey.nil?
@@ -67,8 +69,10 @@ module Puppet::Util::Windows
       index = 0
       subkey = nil
 
+      subkey_max_len, value_max_len = reg_query_info_key_max_lengths(key.hkey)
+
       begin
-        subkey, type, data = reg_enum_value(key.hkey, index)
+        subkey, type, data = reg_enum_value(key.hkey, index, value_max_len)
         yield subkey, type, data if !subkey.nil?
         index += 1
       end while !subkey.nil?
@@ -140,6 +144,36 @@ module Puppet::Util::Windows
       end
 
       [subkey, type, data]
+    end
+
+    def reg_query_info_key_max_lengths(hkey)
+      result = nil
+
+      FFI::MemoryPointer.new(:dword) do |max_subkey_name_length_ptr|
+        FFI::MemoryPointer.new(:dword) do |max_value_name_length_ptr|
+
+          status = RegQueryInfoKeyW(hkey,
+            FFI::MemoryPointer::NULL, FFI::MemoryPointer::NULL,
+            FFI::MemoryPointer::NULL, FFI::MemoryPointer::NULL,
+            max_subkey_name_length_ptr, FFI::MemoryPointer::NULL,
+            FFI::MemoryPointer::NULL, max_value_name_length_ptr,
+            FFI::MemoryPointer::NULL, FFI::MemoryPointer::NULL,
+            FFI::MemoryPointer::NULL
+          )
+
+          if status != FFI::ERROR_SUCCESS
+            raise Puppet::Util::Windows::Error.new('Failed to query registry key info')
+          end
+
+          result = [
+            # Unicode characters *not* including trailing NULL
+            max_subkey_name_length_ptr.read_dword + 1,
+            max_value_name_length_ptr.read_dword + 1,
+          ]
+        end
+      end
+
+      result
     end
 
     # Read a registry value named name and return array of
@@ -304,5 +338,25 @@ module Puppet::Util::Windows
     ffi_lib :advapi32
     attach_function_private :RegDeleteKeyExW,
       [:handle, :lpcwstr, :win32_ulong, :dword], :win32_long
+
+    # https://msdn.microsoft.com/en-us/library/windows/desktop/ms724902(v=vs.85).aspx
+    # LONG WINAPI RegQueryInfoKey(
+    #   _In_         HKEY hKey,
+    #   _Out_opt_    LPTSTR lpClass,
+    #   _Inout_opt_  LPDWORD lpcClass,
+    #   _Reserved_   LPDWORD lpReserved,
+    #   _Out_opt_    LPDWORD lpcSubKeys,
+    #   _Out_opt_    LPDWORD lpcMaxSubKeyLen,
+    #   _Out_opt_    LPDWORD lpcMaxClassLen,
+    #   _Out_opt_    LPDWORD lpcValues,
+    #   _Out_opt_    LPDWORD lpcMaxValueNameLen,
+    #   _Out_opt_    LPDWORD lpcMaxValueLen,
+    #   _Out_opt_    LPDWORD lpcbSecurityDescriptor,
+    #   _Out_opt_    PFILETIME lpftLastWriteTime
+    # );
+    ffi_lib :advapi32
+    attach_function_private :RegQueryInfoKeyW,
+      [:handle, :lpwstr, :lpdword, :lpdword, :lpdword, :lpdword, :lpdword,
+        :lpdword, :lpdword, :lpdword, :lpdword, :pointer], :win32_long
   end
 end
--- a/lib/puppet/util/windows/registry.rb
+++ b/lib/puppet/util/windows/registry.rb
@@ -13,6 +13,8 @@ module Puppet::Util::Windows
     KEY_WRITE      = 0x20006
     KEY_ALL_ACCESS = 0x2003f
 
+    ERROR_NO_MORE_ITEMS = 259
+
     def root(name)
       Win32::Registry.const_get(name)
     rescue NameError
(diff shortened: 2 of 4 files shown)


Iristyle:
FYI - conversion from `UTF-16LE` to `UTF-8` is considered lossless in practical terms.

https://msdn.microsoft.com/en-us/library/windows/desktop/dd374081%28v=vs.85%29.aspx
http://en.wikipedia.org/wiki/UTF-8#Invalid_byte_sequences

spec/integration/provider/package_spec.rb
```diff
@@ -15,6 +15,13 @@
         expect { pkg.provider.install }.to raise_error
       end
 
+      # TODO: this fails on Windows if the active code page is IBM437 with
```
Iristyle: This no longer fails after last commit that translates registry strings (key names, value names and actual values) to UTF8... but still need some sort of test here.

lib/puppet/util/windows/registry.rb
```diff
@@ -30,51 +32,276 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
...
+          FFI::MemoryPointer.new(:wchar, max_key_length) do |subkey_ptr|
+            subkey_length_ptr.write_dword(max_key_length)
+
+            # RegEnumKeyEx cannot be called twice to properly size the buffer
+            result = RegEnumKeyExW(hkey, index,
+              subkey_ptr, subkey_length_ptr,
+              FFI::Pointer::NULL, FFI::Pointer::NULL,
+              FFI::Pointer::NULL, filetime_ptr)
+
+            break if result == ERROR_NO_MORE_ITEMS
```
joshcooper: Does this break out of this method? or does it break out of the caller's loop in https://github.com/puppetlabs/puppet/pull/3540/files#diff-85725d2700dc7373579506c97d84a83eR60? In either case, it might be clearer to do:

<pre>
if result != ERROR_NO_MORE_ITEMS
  if result !=FFI::ERROR_SUCCESS
    raise ...
  end
  ...
end

[subkey, filetime]
</pre>

lib/puppet/util/windows/registry.rb
```diff
@@ -30,51 +32,276 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
...
+
+            if result != FFI::ERROR_SUCCESS
+              raise Puppet::Util::Windows::Error.new('Failed to enumerate registry keys')
```
joshcooper: Can we provide context about which key name & index we failed to enumerate?

Iristyle:
A note about current behavior of this PR, under different runtimes, with different codepages set.

#### CHCP 437

##### Ruby 2.0.0 + PR

```
package { 'Windows Azure Libraries for .NET – v2.2':
  ensure => '2.2.0924.200',
}
```

##### Ruby 2.0.0 + existing code

```
package { 'Windows Azure Libraries for .NET – v2.2':
  ensure => '2.2.0924.200',
}
```

##### Ruby 2.1.5 + PR

```
package { 'Windows Azure Libraries for .NET â?" v2.2':
  ensure => '2.2.0924.200',
}
```

I actually believe the new PR to be correct behavior, even though the strings look incorrect.  If you look at [IBM 437](http://en.wikipedia.org/wiki/Code_page_437) there is no equivalent to U+2013.

##### Ruby 2.1.5 + existing code

```
** CRASH **
Error: Could not run: U+2013 to IBM437 in conversion from UTF-16LE to UTF-8 to IBM437
```

#### CHCP 1252

##### Ruby 2.0.0 + PR

```
package { 'Windows Azure Libraries for .NET – v2.2':
  ensure => '2.2.0924.200',
}
```

##### Ruby 2.0.0 + existing code

```
package { 'Windows Azure Libraries for .NET – v2.2':
  ensure => '2.2.0924.200',
}
```

##### Ruby 2.1.5 + PR

```
package { 'Windows Azure Libraries for .NET â€“ v2.2':
  ensure => '2.2.0924.200',
}
```

[Windows 1252](http://en.wikipedia.org/wiki/Windows-1252) should have an equivalent to the [Unicode En Dash](http://www.fileformat.info/info/unicode/char/2013/index.htm) U+2013 at 96 if you take a look at the [Microsoft conversion chart](https://msdn.microsoft.com/en-us/goglobal/cc305145.aspx)... yet Ruby did not convert it.

##### Ruby 2.1.5 + existing code

```
package { 'Windows Azure Libraries for .NET – v2.2':
  ensure => '2.2.0924.200',
}
```

##### Summary

The behavior is still a bit confusing because the new 437 conversion seems appropriate given no equivalent character, but UTF8 -> 1252 should translate properly.

lib/puppet/util/windows/registry.rb
Iristyle: I had written that way initially, but opted to use `break` for the sake of less indentation.

Happy to change it back if you think its clearer the other way.

lib/puppet/util/windows/registry.rb
```diff
@@ -30,51 +32,276 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
...
+      subkey, filetime = nil, nil
+
+      FFI::MemoryPointer.new(:dword) do |subkey_length_ptr|
+        FFI::MemoryPointer.new(FFI::WIN32::FILETIME.size) do |filetime_ptr|
+          FFI::MemoryPointer.new(:wchar, max_key_length) do |subkey_ptr|
+            subkey_length_ptr.write_dword(max_key_length)
+
+            # RegEnumKeyEx cannot be called twice to properly size the buffer
+            result = RegEnumKeyExW(hkey, index,
+              subkey_ptr, subkey_length_ptr,
```
joshcooper: This API is crazy. As an input parameter `subkey_length_ptr` is the number of characters including NULL wide char. But as the output parameter, it's the number of characters excluding the NULL. It looks like you've handled it correctly.

lib/puppet/util/windows/registry.rb
Iristyle: Sure, sounds good.

lib/puppet/util/windows/registry.rb
Iristyle: Yes, agree that this API is very bizarre with the lack of / presence of trailing NULLs in the count.  I read the docs and debugged quite a bit to make sure I have it right (though should have some more tests for regression protection).

lib/puppet/util/windows/registry.rb
joshcooper: k, sounds good as is

lib/puppet/util/windows/registry.rb
```diff
@@ -30,51 +32,276 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
...
+
+        string_length = 0
+        # buffer is raw bytes, *not* chars - less a NULL terminator
+        string_length = (byte_length / FFI.type_size(:wchar)) - 1 if byte_length > 0
+
+        case type
+        when Win32::Registry::REG_SZ, Win32::Registry::REG_EXPAND_SZ
+          result = [ type, data_ptr.read_wide_string(string_length, Encoding::UTF_8) ]
+        when Win32::Registry::REG_MULTI_SZ
+          result = [ type, data_ptr.read_wide_string(string_length, Encoding::UTF_8).split(/\0/) ]
```
joshcooper: Just to make sure I follow this part, `REG_MULTI_SZ` are special because they should have an extra null terminator, so as characters it'd look like:

<pre>
foo\0bar\0\0
</pre>
 
so `string_length` should be 8, and `byte_length` would be 18? So then `read_wide_string` reads 8 characters and converts it to UTF-8, which generates an 8 byte long string (excluding NULL). Then we split on the embedded NULL (single byte), and get back two strings each 3 bytes long (excluding NULL).

lib/puppet/util/windows/registry.rb
```diff
@@ -30,51 +32,276 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
...
+        when Win32::Registry::REG_BINARY
+          result = [ type, data.read_bytes(0, byte_length) ]
+        when Win32::Registry::REG_DWORD, Win32::Registry::REG_DWORD_BIG_ENDIAN
+          result = [ type, data_ptr.read_dword ]
```
joshcooper: Is that correct for BIG_ENDIAN?

joshcooper:
<quote>
but UTF8 -> 1252 should translate properly.
</quote>

Can you check that the UTF-8 encoded string returned by this PR correctly encodes U+2013 as `0xE2 0x80 0x93`?

lib/puppet/util/windows/registry.rb
ferventcoder: Yes :+1:

lib/puppet/util/windows/api_types.rb
```diff
@@ -52,10 +53,10 @@ def read_handle
     alias_method :read_wchar, :read_uint16
     alias_method :read_word,  :read_uint16
 
-    def read_wide_string(char_length)
+    def read_wide_string(char_length, dst_encoding = Encoding.default_external)
```
ferventcoder: :+1:

ferventcoder:
IIRC, this can be removed and tests fixed up https://github.com/Iristyle/puppet/blob/ticket/master/PUP-3837-FFI-registry-calls-Ruby-215-bug/spec/unit/util/windows/registry_spec.rb#L112-L114 @joshcooper is that correct?

ferventcoder:
@Iristyle this is my output on azure with 437:

```puppet
package { 'Windows Azure Libraries for .NET 1.7 – June 2012':
  ensure => '1.7',
}
```

lib/puppet/util/windows/registry.rb
Iristyle: Odd.  I had verified this in the docs earlier, but now I've found conflicting docs (which seem to correlate what Ruby has done).

Or I could have just been really tired and misread.  I will take a look at the headers to verify the definitions, thanks.

Iristyle: I think I just misread...

https://msdn.microsoft.com/en-us/library/windows/desktop/ms724884%28v=vs.85%29.aspx

```
REG_DWORD_LITTLE_ENDIAN

A 32-bit number in little-endian format.

Windows is designed to run on little-endian computer architectures. Therefore, this value is defined as REG_DWORD in the Windows header files.
```

Iristyle:
@joshcooper - Ruby can indeed handle conversion of endash properly to 1252 from UTF8 (and barfs going to IBM437):

```
C:\source\puppet [master...origin/master +11 ~0 -0 !]> ruby --version
ruby 2.1.5p273 (2014-11-13 revision 48405) [x64-mingw32]


[19] pry(main)> endash_bytes = [0xE2, 0x80, 0x93]
=> [226, 128, 147]
[20] pry(main)> endash_str = endash_bytes.pack('c*').force_encoding('utf-8')
=> "\u2013"
[21] pry(main)> endash_str.encode(Encoding::Windows_1252)
=> "\x96"



[3] pry(main)> endash_str.encode(Encoding::IBM437)
Encoding::UndefinedConversionError: U+2013 from UTF-8 to IBM437
from (pry):3:in `encode'
```

Given this trivial repo case barfs going to IBM437, that casts a lot of doubt around how / why we're getting a string out of Ruby when its something that cannot be converted.

ferventcoder:
@Iristyle why does it work for me?

ferventcoder:
So your example fails for me as expected. However the code seems to be handling this situation correctly?! I agree with you, it seems there is more going on in Ruby somewhere.

joshcooper:
So the simple example confirms (as expected) ruby can't transcode emdash to IBM437, but it does appear to transcode to CP1252, so something is wrong with the PR I think. Given the ruby emdash.rb:

<pre>
endash_bytes = [0xE2, 0x80, 0x93]
endash_str = endash_bytes.pack('c*').force_encoding('utf-8')

puts "CP1252 " + endash_str.encode(Encoding::CP1252)
puts "IBM437 " + endash_str.encode(Encoding::IBM437)
</pre>

I get the same results for all rubies:

<pre>
C:\Ruby193p545\bin>ruby --version
ruby 1.9.3p545 (2014-02-24) [i386-mingw32]
C:\Ruby193p545\bin>.\ruby c:\emdash.rb
CP1252 –
c:/emdash.rb:5:in `encode': U+2013 from UTF-8 to IBM437 (Encoding::UndefinedConversionError)
        from c:/emdash.rb:5:in `<main>'
</pre>

<pre>
C:\Ruby200p481-x64\bin>ruby --version
ruby 2.0.0p481 (2014-05-08) [x64-mingw32]
C:\Ruby200p481-x64\bin>.\ruby c:\emdash.rb
CP1252 –
c:/emdash.rb:5:in `encode': U+2013 from UTF-8 to IBM437 (Encoding::UndefinedConversionError)
        from c:/emdash.rb:5:in `<main>'
</pre>

<pre>
C:\Ruby215\bin>ruby --version
ruby 2.1.5p273 (2014-11-13 revision 48405) [i386-mingw32]
C:\Ruby215\bin>.\ruby c:\emdash.rb
CP1252 –
c:/emdash.rb:5:in `encode': U+2013 from UTF-8 to IBM437 (Encoding::UndefinedConversionError)
        from c:/emdash.rb:5:in `<main>'
</pre>

ferventcoder:
Yeah, so try this:
```
endash_bytes = [0x2013]
endash_str = endash_bytes.pack('s*').force_encoding(Encoding::UTF_16)

puts "UTF-8 " + endash_str.encode(Encoding::UTF_8)
```

And watch that fail fantastically as well. Are we sure that packing and forcing the encoding is properly working? 


Here's some interesting data:

```
[40] pry(main)> endash_str = endash_bytes.pack('s*').force_encoding(Encoding::UTF_16)
=> "\x13\x20"
[45] pry(main)> endash_str.encoding
=> #<Encoding:UTF-16 (dummy)>
```

What is the `(dummy)` part coming back?

ferventcoder:
But this works:

```ruby
endash_bytes = [0x1320]
endash_str = endash_bytes.pack('s*').force_encoding(Encoding::UTF_16LE)

puts "UTF-8 " + endash_str.encode(Encoding::UTF_8)
```


and no dummy.... :+1: 

```
[53] pry(main)> endash_str.encoding
=> #<Encoding:UTF-16LE>
```

lib/puppet/util/windows/api_types.rb
```diff
@@ -52,10 +53,10 @@ def read_handle
...
       # char_length is number of wide chars (typically excluding NULLs), *not* bytes
       str = get_bytes(0, char_length * 2).force_encoding('UTF-16LE')
-      str.encode(Encoding.default_external)
+      str.encode(dst_encoding)
```
ferventcoder: I wonder if we should revisit 
 * `ADSI.self.computername` adsi.rb:44
 * `User.current_user_name` adsi.rb:240
 * `Error.format_error_code` error.rb:51
 * `File.resolve_symlink` file.rb:285

And make sure all of them come back correctly with weird default external encodings?

Iristyle: We actually want to use UTF-8 everywhere if we can.

ferventcoder: Sweet, so does that mean you are going to update all of them to call this an encoding?

lib/puppet/util/windows/registry.rb
Iristyle: So that would actually change how `reg_enum_value` functions and friends are defined (i.e. accepting a `Win32::Registry` class instance instead of a simple `handle` value).

I can certainly do this, but would generally lean toward passing the smallest value possible.

@ferventcoder / @joshcooper - thoughts on expanding this out so that we can grab name off the object?

lib/puppet/util/windows/api_types.rb
Iristyle: Let's worry about that bit later.

lib/puppet/util/windows/registry.rb
joshcooper: I'd really, really like to see the name in the error message :)

Iristyle:
OK @joshcooper / @ferventcoder - I've verified that the following work correctly under Ruby 2.1.5

* `bundle exec puppet resource package`
* `bundle exec rspec .\spec\unit\util\windows\registry_spec.rb .\spec\integration\provider\package_spec.rb`

Presently verifying against Ruby 2 / Ruby 1.9.3 (for the sake of checking impact against CI)

Iristyle:
So it looks like the new test that I've added works fine under Ruby 2.1.5, but fails on 2.0.0 and 1.9.3 due to:

```
Encoding::CompatibilityError:
       incompatible character encodings: UTF-16LE and UTF-8
```

Given I pass a `UTF-16LE` string along to `Win32::Registry#write` and Ruby then tries to plop a `UTF-8` `\0` to the end, I see the issue.  Part of the point of that test is to *not* convert the string at all prior to trying to write it to the registry, so I may have to branch the test based on Ruby version.

spec/unit/util/windows/registry_spec.rb
```diff
@@ -71,70 +71,94 @@ class TestRegistry
...
 
-      it "passes REG_DWORD through" do
-        reg_value = ['dword', Win32::Registry::REG_DWORD, '1']
-
-        value = expects_registry_value(reg_value)
-
-        expect(Integer(value)).to eq(1)
+      # proof that local encodings (such as IBM437 are no longer relevant)
+      it "will return a UTF-8 string from a REG_SZ registry value (written as UTF-16LE)",
+        :if => RUBY_VERSION >= '2.1' do
```
Iristyle: No great way of making this test run correctly on `2.0.0` or `1.9.3` (which would be a temporary measure for the sake of CI), so instead just drop a guard in here to make sure it doesn't cause CI failures.

Iristyle:
:shipit:

lib/puppet/util/windows/api_types.rb
ferventcoder: Fair enough, file a ticket? I can do it or you can do it. :)

Iristyle: Go for it

ferventcoder: https://tickets.puppetlabs.com/browse/PUP-3951

lib/puppet/util/windows/registry.rb
```diff
@@ -30,51 +32,337 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
...
+          msg = "Failed to delete registry value #{name} at #{key.keyname}"
+          raise Puppet::Util::Windows::Error.new(msg)
+        end
       end
+
+      result
     end
-    private :force_encoding
 
+    def reg_delete_key_ex(key, name, regsam = KEY64)
```
ferventcoder: I take this is here as a reimplementation of the win32/registry (I don't really see it used anywhere though).

ferventcoder: I ended up here b/c I was curious abut the `regsam = KEY64` versus  `mode = KEY64` used above in `delete_key` (which also isn't used anywhere).

ferventcoder: This is just observations, no action necessary

ferventcoder:
No spec failures. Code looks clean... I'm pretty happy to merge this in. :+1: 

Just doing a final once over

lib/puppet/util/windows/registry.rb
ferventcoder: Thanks for including the keyname! Will help immensely when errors are thrown! :)

lib/puppet/util/windows/registry.rb
```diff
@@ -30,51 +32,337 @@ def open(name, path, mode = KEY_READ | KEY64, &block)
...
+    # :REG_MULTI_SZ
+    #    Array of String
+    # :REG_DWORD, REG_DWORD_BIG_ENDIAN, REG_QWORD
+    #    Integer
+    # :REG_BINARY
+    #    String (contains binary data)
+    #
+    # When rtype is specified, the value type must be included by
+    # rtype array, or TypeError is raised.
+    def read(key, name_ptr, *rtype)
```
ferventcoder: Interesting this fell under private. No worries, I don't think folks were using it. Just more if a module ever wanted to. At least they have ways into here from above.

lib/puppet/util/windows/registry.rb
Iristyle: This was added for testing / is called via:

```ruby
subject.delete_key(reg, subkey_name)
```

`reg_delete_value` however is unused at the moment.

I added it only for parity - so that both methods were available given internal Ruby methods are completely busted.  I could have understood arguments to cut it based on YAGNI.

lib/puppet/util/windows/registry.rb
Iristyle: Yeah, it was already called by things higher in the call stack, so I don't think we needed to expose it.  Also note that I changed the Ruby implementation a little to optimized for taking an `FFI::MemoryPointer` instead of a Ruby `String`, so that we could avoid an extra few unnecessary allocations / round-tripping through Ruby's UTF-8 process.  That was another +1 for the `keep it private` argument.

lib/puppet/util/windows/registry.rb
Iristyle: To your point about `regsam` vs `mode` - the accepted values, while sharing the same basic access mask enumeration values, have different acceptable values for the called APIs in question:

* [RegDeleteKeyEx](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724847\(v=vs.85\).aspx) REGSAM
* [RegOpenKeyEx](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724897%28v=vs.85%29.aspx) samRequired

lib/puppet/util/windows/registry.rb
ferventcoder: :+1:

PR #3541, opened 2015-01-30
(PUP-3898) Fix writing content with checksum

MikaelSmith:
Fixes writing a file from content for a non-default checksum type when
the content is initialized as part of creating a file resource.
Fix applies to both puppet apply and puppet agent runs.

### DIFF
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -379,6 +379,11 @@ Puppet::Type.newtype(:file) do
 
     self.warning "Possible error: recurselimit is set but not recurse, no recursion will happen" if !self[:recurse] and self[:recurselimit]
 
+    if @parameters[:content] && @parameters[:content].actual_content
+      # Now that we know the checksum, update content (in case it was created before checksum was known).
+      @parameters[:content].value = @parameters[:checksum].sum(@parameters[:content].actual_content)
+    end
+
     provider.validate if provider.respond_to?(:validate)
   end
 
--- a/spec/shared_contexts/checksum.rb
+++ b/spec/shared_contexts/checksum.rb
@@ -6,13 +6,15 @@
 # create a new example group for each types and will run the given block
 # in each example group.
 
-CHECKSUM_STAT_TIME = Time.now
+CHECKSUM_PLAINTEXT = "1\r\n"*4000
 CHECKSUM_TYPES_TO_TRY = [
   ['md5', 'a7a169ac84bb863b30484d0aa03139c1'],
   ['md5lite', '22b4182363e81b326e98231fde616782'],
   ['sha256', '47fcae62967db2fb5cba2fc0d9cf3e6767035d763d825ecda535a7b1928b9746'],
-  ['sha256lite', 'fd50217a2b0286ba25121bf2297bbe6c197933992de67e4e568f19861444ecf8'],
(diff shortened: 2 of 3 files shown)


MikaelSmith:
@joshcooper: last set of checksum fixes.

joshcooper:
@MikaelSmith looks like travis failed

lib/puppet/type/file.rb
```diff
@@ -379,6 +379,11 @@ def self.title_patterns
 
     self.warning "Possible error: recurselimit is set but not recurse, no recursion will happen" if !self[:recurse] and self[:recurselimit]
 
+    if @parameters[:content] and @parameters[:content].actual_content
+      # Now that we know the checksum, update content (in case it was created before checksum was known).
+      @parameters[:content].value = @parameters[:checksum].sum(@parameters[:content].actual_content)
```
MikaelSmith: The value attribute is only made to be updated once on Content. That makes this awkward, but I don't have great ideas how to improve it without some substantial changes: either forced ordering to attribute initialization, or coming up with a pattern for re-initializing Content when it's checksum type changes.

MikaelSmith: On more consideration, I'll explore moving re-initialization to the Content class.

MikaelSmith:
No errors in acceptance tests on Windows or Linux with the 2nd commit.

lib/puppet/type.rb
```diff
@@ -107,13 +107,14 @@ class << self
...
-  # the {parameters} and {metaparams},
+  # The {key_attributes} come first, then the {provider} and {checksum}, then the {properties},
+  # and finally the {parameters} and {metaparams},
   # all in the order they were specified in the respective files.
   # @return [Array<String>] all type attribute names in a defined order.
   #
   def self.allattrs
-    key_attributes | (parameters & [:provider]) | properties.collect { |property| property.name } | parameters | metaparams
+    # Checksum has no side-effects, and should be set early for properties (like content) that use it.
+    key_attributes | (parameters & [:provider, :checksum]) | properties.collect { |property| property.name } | parameters | metaparams
```
joshcooper: I forgot this was in `type.rb`. I don't think we can do this here, as who knows what other types have been created with a `checksum` parameter.

MikaelSmith: Oh, as in external types. Ok, I'll just drop that commit.

lib/puppet/type/file.rb
MikaelSmith: This led to a dead-end; making a method of the Content class that did the same thing interacts with the class in strange ways.

lib/puppet/type/file.rb
```diff
@@ -379,6 +379,11 @@ def self.title_patterns
 
     self.warning "Possible error: recurselimit is set but not recurse, no recursion will happen" if !self[:recurse] and self[:recurselimit]
 
+    if @parameters[:content] and @parameters[:content].actual_content
```
joshcooper: could you change to `&&`?

MikaelSmith: Yes, I'll get to that soon.

joshcooper:
@MikaelSmith changes look good, I just had a small nit on the commit message. Could you describe what the old behavior was and how this commit fixes it? I assume we used to compare the old "content" using the default checksum type, e.g. `{md5}xxxx`, against the non-default checksum type, e.g. `{sha256}yyyy`?

PR #3542, opened 2015-01-30
(PUP-1601) Refactor split and regsubst to 4x API

thallgren:
Refactors the 3x functions to 4x API. Also adds more stringent
type checks and permits patterns to be sent as strings, regexps,
or regexp types.

A parameter restriciton is imposed when using precompiled regexps
or regexps types so that the flags are restricted to 'G' (global
substitution). Flags affectingn the actual regexp are not permitted
since that would mean recompiling the regexp. This approach was
chosen based on the assumption that regexp flags will be properly
implemented in the Puppet Parser in the future.


hlindberg:
Reminder: since this guts the 3x versions, and people may be calling these functions from other 3x functions via scope, we should add a note to documentation about how to call them e.g.
```closure_scope().function_foo()```

PR #3543, opened 2015-01-31
(PUP-3678) Correct and unify module name validation

andersonmills:
Before this commit, we were checking the validity of a module name in
    multiple places. We were also only excluding . and .. from the list
    of modules to create. This permitted directories in the modulepath like
    .git, .svn, and .DS_store to cause errors and warnings. This commit
    consolidates the module name checking and excludes those types of
    directories. It also enforces a name validity check more in line with the Forge validator.

### DIFF
--- a/lib/puppet/module.rb
+++ b/lib/puppet/module.rb
@@ -33,6 +33,25 @@ class Puppet::Module
     env.module(modname)
   end
 
+  def self.is_module_directory?(name, path)
+    # it must be a directory
+    fullpath = File.join(path, name)
+    return false unless Puppet::FileSystem.directory?(fullpath)
+    return is_module_directory_name?(name)
+  end
+
+  def self.is_module_directory_name?(name)
+    # it must match an installed module name according to forge validator
+    return true if name =~ /^[a-z][a-z0-9_]*$/
+    return false
+  end
+
+  def self.is_module_namespaced_name?(name)
+    # it must match the full module name according to forge validator
+    return true if name =~ /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/   
+    return false
+  end
+
   attr_reader :name, :environment, :path, :metadata
   attr_writer :environment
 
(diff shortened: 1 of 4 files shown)


lib/puppet/node/environment.rb
```diff
@@ -271,10 +271,11 @@ def modules
       seen_modules = {}
       modulepath.each do |path|
         Dir.entries(path).each do |name|
-          next if name == "." || name == ".."
+          next unless Puppet::Module.is_module?(name, path)
```
adreyer: Modules in a namespaced(puppetlabs-apache vs apache) path won't load properly. Is this what we want here?

andersonmills: I think that is not the problem trying to be solved by PUP-3678, but should probably go into a new ticket.

andersonmills:
I will note here that this unification of module name checking does mean that where different metrics had been applied in different use cases, the same metric is now being applied in the different cases. This could cause some of those cases to break now especially for users who have written their own unpublished modules.

lib/puppet/module.rb
```diff
@@ -33,6 +33,25 @@ def self.find(modname, environment = nil)
...
+  def self.is_module_directory_name?(name)
+    # it must match an installed module name according to forge validator
+    return true if name =~ /^[a-z][a-z0-9_]*$/
+    return false
+  end
+
+  def self.is_module_namespaced_name?(name)
+    # it must match the full module name according to forge validator
+    return true if name =~ /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/   
+    return false
```
kylog: Can these two lines just be replaced with:

```
name =~ /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/
```

kylog: @adreyer pointed out =~ doesn't actually return a bool. And I found http://stackoverflow.com/questions/2215303/ruby-whats-the-proper-syntax-for-a-boolean-regex-method

But tldr, I'm okay with this as is.

andersonmills: =~ doesn't return true for matches. It returns a number of matches, so we'll leave it as is for now.

adreyer:
The first commit isn't used anymore and should probably be removed

andersonmills:
Removed.

PR #3544, opened 2015-01-31
(PUP-3927) Ensure Fedora uses the systemd service provider

ScottGarman:
Recent changes to the redhat (chkconfig) service provider requires
features that the chkconfig shim doesn't support (specifically, --add).
Besides, the systemd provider should be the default for Fedora anyway.

### DIFF
--- a/acceptance/tests/resource/service/ticket_3927_redhat_systemd_vs_service_provider.rb
+++ b/acceptance/tests/resource/service/ticket_3927_redhat_systemd_vs_service_provider.rb
@@ -1,0 +1,46 @@
+test_name 'RedHat Service vs. Systemd Provider Validation'
+
+# A simple acceptance test to ensure basic usage of the service
+# provider works for a mix of sysV and systemd RedHat Linux platforms
+confine :to, :platform => /el|centos|fedora/
+
+manifest_install_httpd = %Q{
+  package { 'httpd':
+    ensure => present,
+  }
+}
+
+manifest_httpd_enabled = %Q{
+  service { 'httpd':
+    enable => true,
+  }
+}
+
+manifest_httpd_disabled = %Q{
+  service { 'httpd':
+    enable => false,
+  }
+}
+
+agents.each do |agent|
+  distro = on(agent, facter('operatingsystem')).stdout.chomp
+  if distro == 'fedora'
+    majrelease = on(agent, facter('operatingsystemmajrelease')).stdout.chomp.to_i
+    if majrelease < 17
+      skip_test "Test not applicable to Fedora #{majrelease}"
+    elsif majrelease > 21
+      # This is a reminder so we update the provider's defaultfor when new
+      # versions of Fedora are released (then update this test)
+      fail_test "Provider needs manual update to support Fedora #{majrelease}"
+    end
+  end
+
+  step "installing httpd"
+  apply_manifest_on(agent, manifest_install_httpd, :catch_failures => true)
+
+  step "enabling httpd"
+  apply_manifest_on(agent, manifest_httpd_enabled, :catch_failures => true)
+
+  step "disabling httpd"
+  apply_manifest_on(agent, manifest_httpd_disabled, :catch_failures => true)
+end
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -7,6 +7,7 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
 
   defaultfor :osfamily => [:archlinux]
   defaultfor :osfamily => :redhat, :operatingsystemmajrelease => "7"
+  defaultfor :osfamily => :redhat, :operatingsystem => :fedora, :operatingsystemmajrelease => ["17", "18", "19", "20", "21"]
 
   def self.instances
     i = []
--- a/spec/unit/provider/service/systemd_spec.rb
+++ b/spec/unit/provider/service/systemd_spec.rb
@@ -30,10 +30,22 @@ describe Puppet::Type.type(:service).provider(:systemd) do
     described_class.default?.should be_true
   end
 
-  it "should not be the default provider on rhel6" do
-    Facter.expects(:value).with(:osfamily).at_least_once.returns(:redhat)
-    Facter.expects(:value).with(:operatingsystemmajrelease).returns("6")
-    described_class.default?.should_not be_true
+  [ 4, 5, 6 ].each do |ver|
+    it "should not be the default provider on rhel#{ver}" do
+      Facter.expects(:value).with(:osfamily).at_least_once.returns(:redhat)
+      Facter.expects(:value).with(:operatingsystem).returns(:redhat)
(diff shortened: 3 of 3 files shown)


ScottGarman:
I'm not sure if the acceptance test is overkill - perhaps the spec tests should be sufficient?

lib/puppet/provider/service/systemd.rb
```diff
@@ -7,6 +7,7 @@
 
   defaultfor :osfamily => [:archlinux]
   defaultfor :osfamily => :redhat, :operatingsystemmajrelease => "7"
+  defaultfor :osfamily => :redhat, :operatingsystem => :fedora, :operatingsystemmajrelease => ["17", "18", "19", "20", "21"]
```
joshcooper: I'm worried we are going to forget to increment these as new versions are released, especially because fedora releases are more frequent than rhel.

One way may be to confine the acceptance tests to `/el\-|centos\-|fedora\-/` and blacklist older versions, and fail versions that we don't support yet:

```
agents.each do |agent|
  case on(agent, facter('osfamily')).stdout.chomp
  when 'fedora'
    majrelease = on(agent, facter('operatingssystemmajrelease')).stdout.chomp.to_i
    if majrelease < 17
      skip_test "Test not applicable to fedora #{majrelease}"
    elsif majrelease > 21
      fail_test "Provider does not support fedora #{majrelease}"
    end
  when 'redhat'
   ...
```

Actually now that I write this, I think we could do this check in specs to handle the fedora 22, etc cases.

acceptance/tests/resource/service/ticket_3927_redhat_systemd_vs_service_provider.rb
```diff
@@ -0,0 +1,34 @@
+test_name 'RedHat Service vs. Systemd Provider Validation'
+
+# A simple acceptance test to ensure basic usage of the service
+# provider works for a mix of sysV and systemd RedHat Linux platforms
+confine :to, :platform => /el\-5|el\-6|el\-7|centos\-5|centos\-6|centos\-7|fedora\-19|fedora\-20/
+
+manifest_install_httpd = %Q{
+  package { 'httpd':
+    ensure => present,
```
joshcooper: is this package available for all platforms we confine to and test on? Is it always in the base VM from the pooler, or is the acceptance test installing it?

ScottGarman: @joshcooper yes, I made sure that the httpd service is available in the default repos for all the redhat platforms we're testing here.

lib/puppet/provider/service/systemd.rb
ScottGarman: Sure, I can make these improvements

joshcooper:
@ScottGarman it looks like some extra commits made it into your PR, could you repush with PUP-3947 commits removed?

ScottGarman:
@joshcooper I've fixed the rebase, and the acceptance test has been fail and warn about the need to update the provider. I'm not sure if you had something else in mind regarding the spec test, though.

ScottGarman:
@joshcooper I think this is merge-able pending your final review.

joshcooper:
@peterhuene @whopper @kylog could you guys review this PR, merge to stable for 3.7.5 and merge up to master?

lib/puppet/provider/service/systemd.rb
peterhuene: On that same note, I'd certainly prefer something like:

```ruby
defaultfor do
  Facter.value(:operatingsystem) == 'Fedora' && Facter.value(:operatingsystemmajrelease).to_i >= 17
end
```

Unfortunately `defaultfor` isn't expressive enough to enable this as it only allows for feature or fact value comparisons.  We could certainly extend it to support blocks, though.

The reason I prefer it this way is that we would only have to fix the provider when Fedora moves away from systemd, which I'm sure will happen less frequently than Fedora doing a release.

ScottGarman: @peterhuene that would be a nicer way to deal with this. Given that we don't have this capability yet, is that your only feedback? If so, I'd like to get this merged.

acceptance/tests/resource/service/ticket_3927_redhat_systemd_vs_service_provider.rb
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+agents.each do |agent|
+  distro = on(agent, facter('operatingsystem')).stdout.chomp
+  if distro == 'fedora'
+    majrelease = on(agent, facter('operatingsystemmajrelease')).stdout.chomp.to_i
+    if majrelease < 17
+      skip_test "Test not applicable to Fedora #{majrelease}"
+    elsif majrelease > 20
```
peterhuene: Isn't 21 supported?

lib/puppet/provider/service/systemd.rb
peterhuene: Other than the comment I just added, no.  I can live with having explicit version number checks for now.

acceptance/tests/resource/service/ticket_3927_redhat_systemd_vs_service_provider.rb
ScottGarman: @peterhuene I tried testing F21 with our internal acceptance test beaker setup and we don't offer a repo file to set it up. (It's looking in http://builds.puppetlabs.lan/puppet/3.7.3/repo_configs/rpm/). So I assumed I needed to skip F21 to avoid causing errors in our CI testing. Let me know if that assumption is wrong.

peterhuene: I don't see why checking for `majrelease > 21` here would cause CI failures, unless I'm misunderstanding the purpose of this code.  When we do stand up a F21 config, won't it unnecessarily fail if we leave it `> 20`?

peterhuene:
Once we're in agreement about that last comment, I'll get this merged through to master.

peterhuene:
@kylog @whopper Do you guys have any comments before I merge?

whopper:
@peterhuene looks sane to me! :+1:

ScottGarman:
@peterhuene I've updated the acceptance test to include Fedora 21. Thanks for the feedback.

whopper:
:+1: on running the acceptance test locally on Fedora 20:

       Total Suite Time: 20.78 seconds
      Average Test Time: 20.78 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1

      - Specific Test Case Status -

I also ran the test against master *without* the `defaultfor` change, and it failed as we'd expect it to:

    Error: Could not enable httpd: Execution of '/sbin/chkconfig --add httpd' returned 1: error reading information on service httpd: No such file or directory
    Error: /Stage[main]/Main/Service[httpd]/enable: change from false to true failed: Could not enable httpd: Execution of '/sbin/chkconfig --add httpd' returned 1: error reading information on service httpd: No such file or directory

PR #3545, opened 2015-01-31
(PUP-3826) Update API docs for Puppet 4 URL changes

rlinehan:
For Puppet 4, a number of changes have been made to Puppet's API. These
include splitting it into a master and CA API, complete with separate url
prefixes and explicit versioning, and moving the environment out of the path
and into a query parameter. This commit updates the API docs accordingly.


api/docs/http_api_index.md
```diff
@@ -1,13 +1,43 @@
 V1 API Services
 ---------------
 
+The V1 API has been removed. All previous routes can now be found under the
+[Master V3](#Master-V3-HTTP-API) API or the [CA V1](#CA-V1-HTTP-API).
```
rlinehan: It seems like these docs eventually end up on our docs.puppetlabs.com site (https://docs.puppetlabs.com/references/3.7.latest/developer/file.http_api_index.html), and I don't know what kinds of markdown is allowed, so this might not be the right thing to do to link to other sections on the same page.

api/docs/http_api_index.md
```diff
@@ -1,13 +1,43 @@
...
+
+V2 HTTP API
+-----------
+The V2 API has been removed. All previous routes can now be found under the
+[Master V3](#V3-HTTP-API).
```
rlinehan: Should this and the V1 section be totally removed - as I have done - or should they be marked as deprecated but left otherwise intact?

api/docs/http_api_index.md
```diff
@@ -1,13 +1,43 @@
...
+one for the certificate authority (CA).
+
+All master endpoints are prefixed with `/puppet`, while all CA endpoints are
+prefixed with `/puppet-ca`. All endpoints are explicitly versioned.
+
+Authorization for these endpoints is still controlled with the `auth.conf`
+authorization system in puppet. When specifying the authorization in
+`auth.conf` the prefix (either `/puppet` or `/puppet-ca`) and the version
+mumber on the paths must be retained; the full request path is used.
+
```
kylog: s/mumber/number/

api/docs/http_api_index.md
```diff
@@ -1,13 +1,43 @@
...
+The V3 API contains endpoints of two types: those based off of dispatching to
+puppet's internal "indirector" framework and those that are not (namely the
+[environments endpoint](#Environments-Endpoint)).
+
+Every HTTP endpoint that dispatches to the indirector follows the form:
+`/puppet/v3/:indirection/:key?environent=:environment` where
+
   * `:environment` is the name of the environment that should be in effect for
     the request. Not all endpoints need an environment, but the path component
     must always be specified.
```
kylog: Maybe update "the path component" to "the environment query parameter"?

But actually is this true? Does the caller need to specify `?environment=:environment` in all cases now?

api/docs/http_api_index.md
```diff
@@ -95,6 +107,33 @@ error responses will uniformly be a JSON object with the following properties:
...
+CA V1 HTTP API
+--------------
+
+The CA API contains all of the endpoints used in support of Puppet's PKI
+system.
+
+The CA V1 endpoints share the same basic format as the Master V3 API, since
+they are also based off of puppet's internal "indirector". However, they have
+a different prefix and version. The endpoints thus follow the form:
+`/puppet-ca/v1/:indirection/:key?environent=:environment` where
```
kylog: s/environent/environment/

api/docs/http_api_index.md
```diff
@@ -95,6 +107,33 @@ error responses will uniformly be a JSON object with the following properties:
...
+
+  * `:environment` is the name of the environment that should be in effect for
+    the request. Not all endpoints need an environment, but the path component
+    must always be specified.
```
kylog: Same question here about whether to refer to it as a "path component" or as a query parameter or some such.

api/docs/http_api_index.md
```diff
@@ -82,10 +98,6 @@ use standard HTTP response code to signify those errors.
 * When the server is unable to find an endpoint handler for the request that
   starts with `/v2.0`, it will return a 404 Not Found response
```
kylog: Should this clause be removed or tweaked? The reference to "/v2.0" doesn't make sense any more.

api/docs/http_api_index.md
kylog: Ping @nfagerlund re what markdown can be used (and general review).

api/docs/http_api_index.md
kylog: My $.02: I'd totally remove them, because they are removed and not deprecated. That said, it might still make sense on the docs site to retain docs for the 3.7.x-era endpoints. Not sure if that's worth it though, and also not sure what it would take from a markdown/etc perspective.

Related: I'd lean toward not having separate "V1 API Services" and "V2 HTTP API" sections with the repeated verbiage, but just a single "V1/V2 HTTP API" with the verbiage. And that might be where to have a link to the 3.7.x API docs if we go that route.

kylog: Also, once the above questions get sorted out, will this PR cover updates to the individual endpoint docs (`api/docs/http_certificate.md` and so forth)?

api/docs/http_api_index.md
nfagerlund: Hello!

1. The way they currently end up on the site sucks, and I'm gonna replace it. In fact, I'll go ahead and promise it'll be different for Puppet 4 -- if I don't have time to make the automated process better, I'll cart them over by hand. 
2. Hmm, I should probably write a confluence page or something about writing markdown that works on both docs site and github. Well, here's the capsule version:

* Fenced code blocks are cool, but you need to fence them with triple-tildes instead of triple-backticks.
* It's okay to make relative links to .markdown files, the way you would to link to a file in the same github repo. On the docs site, we munge those to .html with RedirectMatch (or nginx's equiv). 
* You can usually link to a header inside a document so it'll work on both our site and Github. We both auto-convert header text to element IDs, and while we don't use the exact same algorithm, it's close enough that they're the same for most English text. To convert a header to a linkable ID:
    * If there's a run of non-alphabetic characters at the _start_ of the header text (including numbers), delete them.
    * Delete any characters that don't match `/[a-zA-Z0-9 -]/`. (Note that underscores get the axe.)
    * Replace all spaces with hyphens. 
    * Lowercase all letters. 

So those two would be `#master-v3-http-api` and `#ca-v1-http-api`. 

I'll take a peek at the content in a sec!

api/docs/http_api_index.md
rlinehan: @kylog Whoops, totally missed that the individual endpoint docs need to have changes. Those will be part of this PR - I'll push that up shortly.

Re: V1 and V2 sections - I think I agree that since we are removing those APIs it makes sense to just have a single "V1/V2 HTTP API" section. I like the idea of linking to 3.7.x docs there, but I don't know what exactly to link to - would it just be to https://docs.puppetlabs.com/references/3.7.latest/developer/file.http_api_index.html?

api/docs/http_api_index.md
rlinehan: Yeah, the environment query parameter must always be present for any routes that go through the indirector. This is a little weird, but making that not the case would have required a lot more changes to the indirector than seemed worth it.

This is not the case for the CA routes when running puppet server, but I don't know how we're handling puppet server docs or whether these docs should mention puppet server at all.

api/docs/http_api_index.md
rlinehan: Should definitely be `query parameter`.

api/docs/http_api_index.md
rlinehan: Perhaps "When the server is unable to find an endpoint handler for the request to the endpoint, it will return a 404 Not Found response"?

kylog: Yeah that works for me. Or just ""When the server is unable to find an endpoint handler for an http request, it will return a 404 Not Found response"?

api/docs/http_api_index.md
kylog: Okay that makes sense. I'm all for reducing work on the indirector :)

And, yeah, I'd think we don't need to mention puppet server in *this* set of docs, at least for puppet 4.

api/docs/http_api_index.md
kylog: /cc @nfagerlund for that last question. I have no idea ...

rlinehan:
@kylog updated for all of your comments. Still an open question on what the best thing to do for `V1/V2 HTTP API` section is, and if we want to link to 3.7 docs, how do we do that.

api/docs/http_catalog.md
```diff
@@ -42,7 +42,7 @@ Three parameters should be provided to the POST or GET:
 
 #### Catalog found
 
-    POST /env/catalog/elmo.mydomain.com
+    POST /puppet/v3/catalog/elmo.mydomain.com?environment=:environment
```
fpringvaldsen: Since this is an example, should we have the environment set to `production` instead of `:environment`, since that seems to be the environment based on the example response?

fpringvaldsen:
:+1: barring my minor comment and the resolution of the `V1/V2 HTTP API` question.

api/docs/http_catalog.md
rlinehan: Whoops, should have had that be `?environment=env`. I was just trying to keep all the environments/urls the same, just updated for the new structure.

kylog:
:+1:

kylog:
@nfagerlund any reservations on this?

nfagerlund:
Ruth and I were talking in chat, but I wanted a slightly wider gut-check on some things. (These won't all be addressed by this one 1-point ticket! But I want to make sure I'm clear to take action on them later.)

- The version of these files readable on GitHub is the canonical version. 
    - In the future, a version on docs.puppetlabs.com that resembles the PuppetDB API docs will be co-canonical with the GitHub version. 
    - The way we display these files in the yarddoc output is not canonical, and we should feel free to remove it or break it slightly more than it's already broken. 
- We should harmonize the Puppet API docs with the PuppetDB and Puppet Server API docs. 
    - We should use the same placeholder tokens for both. The docs team supports using `<PLACEHOLDER>` instead of `:placeholder`. 
    - We should split the docs into directories the same way PuppetDB does, to facilitate multiple API versions coexisting.

Does any of that seem controversial?

api/docs/http_api_index.md
nfagerlund: Hmm. I should probably import the old versions as real docs pages. I can do that by hand...

kylog:
@nfagerlund none of that is controversial to me, but I could use more details around the puppetdb model.

E.g. re directory structure, is it just the idea of `vN` directory per supported api versions? Or mirror the full pdb directory structure?

And what (if anything) should we do today, where we *don't* support multiple API versions in puppet but an older version of puppet supports an older version of the API?

(No super strong opinions on any of the above, btw, just kicking ideas around.)

nfagerlund:
Mostly I'm just talking about the /v3, /v4 directories. We might take the chance to re-arrange the top-level directories while we're at it, and do something like:

~~~
documentation
    puppet-api
        v4
            ...md files...
            schemas
                ...json files...
    puppet-ca-api
        ....
~~~

As for dealing with the old API versions... I think the thing to do is add a manually cleaned-up version of the API docs to the Puppet 3.7 reference manual, and link to those as appropriate.

api/docs/http_file_metadata.md
```diff
@@ -12,7 +12,7 @@ the following three types:
 * symbolic link
 
 Note that an `:environment` must be specified in the endpoint, but is actually ignored since the puppet file server
-is not environment-specific.  (In fact, the specified `:environment` does even need to be valid.)
+is not environment-specific.
```
nfagerlund: Hey, is this actually true? Because if you're asking for anything under `/file_metadata/<ENVIRONMENT>/modules/`, it should matter a lot, and I believe I wrote docs assuming it does.

nfagerlund:
With the comment on http_file_metadata addressed and the commits from https://github.com/nfagerlund/puppet/tree/feb2015_ruth_api amending the wording, I'm :+1: on this.

api/docs/http_file_metadata.md
kylog: Ah right. It ignores `:environment` *except* for the magical `/file_metadata/:environment/modules/` and `/file_metadata/:environment/plugins` endpoints, where it is . Sigh.

kylog:
@nfagerlund I like the idea for reorg'ing the docs directory structure and commonizing it across projects, but I had more questions and I'm afraid the convo may get lost in this PR. Maybe we should kick off an email thread.

nfagerlund:
@kylog Yeah good call. Will do.

api/docs/http_file_metadata.md
rlinehan: So it sounds like I should probably just remove this paragraph since it's not really true at all, or I could try to say something about where `:environment` is ignored and where it isn't? But trying to explain that seems like it might be more confusing than helpful....

nfagerlund: I favor removal.

kylog: Yeah I favor removal.

nfagerlund:
...wait, I just noticed http_file_content.md didn't get updated. Oversight, or did that endpoint get left unmodified for some reason?

rlinehan:
@nfagerlund Argh, total oversight on my part; will update.

rlinehan:
@nfagerlund @kylog Removed the file_metada environment note, updated the file_content doc, and merged in Nick's edits to my branch. This should be good to go.

kylog:
:+1:

nfagerlund:
:+1:

PR #3546, opened 2015-01-31
(PUP-3934) Propagate environment to Resource initializer

thallgren:
Prior to this commit, if a Resource was created as 'Bar[instance]',
it would fail an attemt to fetch the resource type from its
environment. The environment would not be set and thus fallback
to whatever the :current_environment referred to. This made
the test spec/integration/parser/ruby_manifest_spec.rb fail.

Here, the environment is passed from the catalog to the Resource
initialize methdo in the attributes hash and properly assigned
before used.

### DIFF
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -211,6 +211,7 @@ class Puppet::Resource
   # @api public
   def initialize(type, title = nil, attributes = {})
     @parameters = {}
+    environment = attributes[:environment]
     if type.is_a?(Class) && type < Puppet::Type
       # Set the resource type to avoid an expensive `known_resource_types`
       # lookup.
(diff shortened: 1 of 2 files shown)


hlindberg:
@kylog @joshcooper I am good with this change, but want your eyes on it as well. Basically this gives logic the ability to set the environment of a resource that is created via the catalog (to get the same environment as the catalog is in) instead of later when the resource is asked for its environment and it tries to look it up from its catalog (which is sometimes does not have), then from current_environment in context (which some tests do not establish), and then in the environment none.  You could argue that tests should set things up correctly - but it is better if we instead could get rid of the behavior where resource goes on a hunt for its environment; the change in this PR is a step in the right direction. (I am reluctant to make a bigger change to clean this up completely this late in the game for 3.7.5/4.0.0). 

Hence... second opinions? (if you are +1 on this, go ahead and merge).

PR #3547, opened 2015-02-01
(PUP-3935) Fix pluginsync errors when dealing with symlinks

whopper:
Currently, Puppet produces one of two pluginsync errors during an
agent run if the root of a modulepath directory is a symlink
and no modules contain a facts.d or lib directory.

In the normal case, when the base module path is a directory
and none if its modules include facts.d or lib directories,
pluginsync returns the path to the module path directory itself,
at which point the file type disables recursion in its
`recurse_remote` method and stops the agent from attempting to
download plugins.

However, if the base module path is not a directory (in this case
a link), `recurse_remote` returns early, causing the agent-side
facts.d or lib directory to not have its 'source' metadata set,
leading to a failing query back to the master.

This commit forces pluginsync to follow symlinks rather than
manage them. To do this, the :links attribute is set to :follow
in the #default_arguments method of the
Puppet::Configurer::Downloader class, which is only used
for pluginsync. Doing this ensures that what pluginsync
returns will be a directory rather than a link, allowing the
resource to have its source metadata correctly set.

Paired with Peter Huene <peter.huene@puppetlabs.com>

### DIFF
--- a/acceptance/tests/pluginsync/3935_pluginsync_should_follow_symlinks.rb
+++ b/acceptance/tests/pluginsync/3935_pluginsync_should_follow_symlinks.rb
@@ -1,0 +1,44 @@
+test_name "pluginsync should not error when modulepath is a symlink and no modules have plugin directories"
+
+step "Create a modulepath directory which is a symlink and includes a module without facts.d or lib directories"
+basedir = master.tmpdir("symlink_modulepath")
+
+target           =  "#{basedir}/target_dir"
+test_module_dir  =  "#{target}/module1"
+link_dest        =  "#{basedir}/link_dest"
+modulepath       =  "#{link_dest}"
+modulepath << "#{master['sitemoduledir']}" if master.is_pe?
+
+apply_manifest_on(master, <<MANIFEST, :catch_failures => true)
+File {
+  ensure => directory,
+  mode   => "0750",
+  owner  => #{master.puppet['user']},
+  group  => #{master.puppet['group']},
+}
+
+file {
+  '#{basedir}':;
+  '#{target}':;
+  '#{test_module_dir}':;
+}
+
+file { '#{link_dest}':
+  ensure => link,
+  target => '#{target}',
+}
+MANIFEST
+
+master_opts = {
+  'main' => {
+    'basemodulepath' => "#{modulepath}"
+  }
+}
+
+with_puppet_running_on master, master_opts, basedir do
+  agents.each do |agent|
+    on(agent, puppet('agent', "-t --server #{master}"))
+      assert_no_match(/Could not retrieve information from environment production source\(s\) puppet:\/\/\/pluginfacts/, stderr)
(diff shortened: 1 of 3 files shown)


whopper:
I need to clarify in https://tickets.puppetlabs.com/browse/PUP-3324 whether the issue there was that symlinks were involved. If so we can use that ticket.

acceptance/tests/pluginsync/yyzz_pluginsync_should_follow_symlinks.rb
```diff
@@ -0,0 +1,44 @@
+test_name "pluginsync should follow symlinks for facts.d and lib directories in the modulepath"
```
whopper: I'll rename this test file once I have a ticket number :)

whopper:
I dunno if the acceptance test is necessary, but it's there if we want it.

joshcooper:
Without the fix:


```
 $ bundle exec rake ci:test:packages TESTS=tests/pluginsync/3935_pluginsync_should_follow_symlinks.rb SHA=f94e84467d6efba95b274eaaded4945e156a9ac6 CONFIG=config/nodes/rhel7.yaml
...
lv21f4zvd7fouwe.delivery.puppetlabs.net (master) 14:15:51$ puppet agent -t --server lv21f4zvd7fouwe.delivery.puppetlabs.net
Info: Retrieving pluginfacts
Error: /File[/var/lib/puppet/facts.d]: Could not evaluate: Could not retrieve information from environment production source(s) puppet:///pluginfacts
Info: Retrieving plugin
Error: /File[/var/lib/puppet/lib]: Could not evaluate: Could not retrieve information from environment production source(s) puppet:///plugins
Info: Caching catalog for lv21f4zvd7fouwe.delivery.puppetlabs.net
Info: Applying configuration version '1422915354'
```

With the fix:

```
$ bundle exec rake ci:test:packages TESTS=tests/pluginsync/3935_pluginsync_should_follow_symlinks.rb SHA=16dada13538a3074925ca6b8b75631cd5db582e5 CONFIG=config/nodes/rhel7.yaml
...
bymm8hcumou6gv5.delivery.puppetlabs.net (master) 14:27:45$ puppet agent -t --server bymm8hcumou6gv5.delivery.puppetlabs.net
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Caching catalog for bymm8hcumou6gv5.delivery.puppetlabs.net
Info: Applying configuration version '1422916068'

bymm8hcumou6gv5.delivery.puppetlabs.net (master) executed in 2.68 seconds

cvt0touq06arilk.delivery.puppetlabs.net (agent) 14:27:48$ puppet agent -t --server bymm8hcumou6gv5.delivery.puppetlabs.net
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Caching catalog for cvt0touq06arilk.delivery.puppetlabs.net
Info: Applying configuration version '1422916068'
```

:+1:

PR #3548, opened 2015-02-01
(PUP-3863) Move 3x hiera functions to 4x API

thallgren:
This commit moves all four hiera functions from 3x to 4x. It also moves the
parameter validation from the HieraPuppet class into the 4x function dispatch.

Tests are added to verify the validity of the new functions (and invalidity of
the old).

Since three functions are identical in all aspects except the merge type
argument sent to hiera, and since the merge type can be derived from the name
of the function, an approach was taken where these three functions share the
same class definition block. The hiera_include is separate though, since its
logic differs.

### DIFF
--- a/lib/puppet/functions/hiera.rb
+++ b/lib/puppet/functions/hiera.rb
@@ -1,0 +1,26 @@
+require 'puppet/functions/hiera_function'
+
+# Performs a standard priority lookup and returns the most specific value for a given key.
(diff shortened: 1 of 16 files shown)


thallgren:
Some rewrite needed but PR will have to wait for [PR-3549](https://github.com/puppetlabs/puppet/pull/3549)

PR #3549, opened 2015-02-01
(PUP-3936) Ensure that weaving logic is aware of block

thallgren:
Previously, a block parameter was not added to the weaving array
which resulted in it being omitted by the weaving logic. This
commit ensures two things.

1. No parameters can be added after a block parameter.
2. The index of the block parameter (size of the names array)
   is added to the weaving array.

A test is also added that tests combining injection with a required block
parameter.


hlindberg:
nice catch ! looks good

PR #3550, opened 2015-02-01
(PUP-777) Remove 'versionRequirement' for dependencies

andersonmills:
Prior to this commit, we had undocumented support for
'versionRequirement' instead of 'version_requirement' for a dependency
in the metadata.json. This support was created because an old version
of Gepetto inadvertently gave this as the field key. With this commit,
we remove the support for that undocumented field.


ScottGarman:
:+1:

PR #3551, opened 2015-02-01
(PUP-3863) Move 3x hiera functions to 4x API

thallgren:
This commit moves all four hiera functions from 3x to 4x. It also moves the
parameter validation from the HieraPuppet class into the 4x function dispatch.

Tests are added to verify the validity of the new functions (and invalidity of
the old).

Since three functions are identical in all aspects except the merge type
argument sent to hiera, and since the merge type can be derived from the name
of the function, an approach was taken where these three functions share the
same class definition block. The hiera_include is separate though, since its
logic differs.


lib/puppet/pops/binder/lookup.rb
```diff
@@ -137,7 +137,14 @@ def self.search_for(scope, type, name, options)
    end
   end
 
-  # This is the method called from the puppet/parser/functions/lookup.rb
```
hlindberg: I don't think we want to give the Lookup class this public method since it means it can be used elsewhere to call any function in any scope. Isn't this only used to call `hiera` btw?

lib/puppet/functions/hiera_common.rb
```diff
@@ -0,0 +1,67 @@
+require 'hiera_puppet'
+
+# Provides the block that defines the hiera, hiera_array, and hiera_hash functions. The only difference
+# between those functions is the merge type (:priority, :array, or :hash) passed to the Hiera lookup. The
+# definition contained here derives the merge type from the name of the function.
+#
+class Puppet::Functions::HieraCommon
```
hlindberg: We probably do not want to have this functionality in this namespace as it will fail mysteriously if someone tries to call the function "hiera_common" from puppet, as the loader will try to load this file. The error will be very different that if the file did not exist in this directory. I think it should move to some other suitable place (any subnamespace of lib/puppet/functions/ is also subject to the same issue - so move it out of the way; I think we have a lib/puppet/hiera namespace - move it there instead).

hlindberg:
The documentation does not seem to mention that blocks can be used to deal with missing lookup results and how that works. (And when documentation that, the docs needs to also be on the 3x function stubs to get documentation generation for 4.0.0...)

hlindberg:
have not reviewed the spec tests yet...

lib/puppet/pops/binder/lookup.rb
thallgren: This method doesn't provide access to anything that wasn't public before. The reason I broke it out into a separate method was purely stylistic. I guess {{search_path}} and some other methods are broken out for the same reason? Perhaps they all should be private?

lib/puppet/functions/hiera_common.rb
thallgren: We have no lib/puppet/hiera namespace. We do have a lib/hiera namespace though and we also have a HieraPuppet class in the /lib namespace. Is it OK if I place it in the lib/hiera namespace rather than creating a third namespace for hiera related stuff?

lib/puppet/pops/binder/lookup.rb
hlindberg: Please make it private, and also specific to only do the hiera call (the Lookup does not need a general way to call any function). Other helper methods should also be private if they are not.

thallgren:
The docs do mention that a block can be used to provide a default (#2 in the ways to call). I'll copy that to the 3x methods as well.

hkenney:
So after talking to @hlindberg, we agreed that the 3x functionality for these functions needs to stay (since this is targeted at stable and will go into 3.7.5). I'm going to go ahead and close this pull request, and I'll reopen one that includes both the 3x and 4x versions that we can merge into stable.

PR #3554, opened 2015-02-02
(PUP-3921) Add 404 error suggesting an upgrade

fpringvaldsen:
Add a 404 error that is triggered when requests are made to
routes without the /puppet or /puppet-ca prefixes to indicate
to the user that they may need to upgrade Puppet on their
agents.

This is a new PR addressing the same ticket as https://github.com/puppetlabs/puppet/pull/3532, which I am closing in favor of this one.

### DIFF
--- a/lib/puppet/network/http.rb
+++ b/lib/puppet/network/http.rb
@@ -3,10 +3,10 @@ module Puppet::Network::HTTP
   HEADER_PUPPET_VERSION = "X-Puppet-Version"
 
   MASTER_URL_PREFIX = "/puppet"
-  CA_URL_PREFIX = "/puppet-ca"
+  MASTER_URL_VERSIONS = "v3"
 
-  PUPPET_API_VERSIONS = "v3"
-  PUPPET_CA_API_VERSIONS = "v1"
+  CA_URL_PREFIX = "/puppet-ca"
+  CA_URL_VERSIONS = "v1"
 
   require 'puppet/network/authorization'
   require 'puppet/network/http/issues'
--- a/lib/puppet/network/http.rb
+++ b/lib/puppet/network/http.rb
@@ -5,6 +5,9 @@ module Puppet::Network::HTTP
   MASTER_URL_PREFIX = "/puppet"
   CA_URL_PREFIX = "/puppet-ca"
 
+  PUPPET_API_VERSIONS = "v3"
+  PUPPET_CA_API_VERSIONS = "v1"
+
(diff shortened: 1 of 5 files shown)


lib/puppet/network/http.rb
```diff
@@ -5,6 +5,9 @@ module Puppet::Network::HTTP
   MASTER_URL_PREFIX = "/puppet"
   CA_URL_PREFIX = "/puppet-ca"
 
+  PUPPET_API_VERSIONS = "v3, v2"
```
kylog: I could be behind the times but I thought puppet 4 was `v3`only for the `puppet` part of the API?

rlinehan: Yep, this is correct. See PUP-3581.

fpringvaldsen:
@jeffmccune @rlinehan I've addressed your feedback on https://github.com/puppetlabs/puppet/pull/3532 with this PR.

lib/puppet/network/http.rb
fpringvaldsen: Ah, I see. I knew v2 was getting removed but wasn't sure if that was going to be removed for Puppet 4 or not. I've fixed it to only say v3.

lib/puppet/network/http.rb
```diff
@@ -5,6 +5,9 @@ module Puppet::Network::HTTP
   MASTER_URL_PREFIX = "/puppet"
   CA_URL_PREFIX = "/puppet-ca"
 
+  PUPPET_API_VERSIONS = "v3"
+  PUPPET_CA_API_VERSIONS = "v1"
+
```
kylog: Would it make sense to name these two VERSIONS constants to match up with the corresponding `_URL_PREFIX` constants above? I.e. to make it more obvious that they go with those.

E.g.

```
MASTER_URL_PREFIX = "/puppet"
MASTER_URL_VERSIONS = "v3"

CA_URL_PREFIX = "/puppet-ca"
CA_URL_VERSIONS = "v1"
```

Naming is hard, so please feel free to ignore me :)

nfagerlund: +1 to that.

fpringvaldsen:
@kylog I've renamed the constants as per your suggestion

kylog:
Oops, I meant to circle back to this one. I'm :+1: 

@jeffmccune or @rlinehan or @nfagerlund you all had comments on the first PR. Thoughts?

rlinehan:
:+1:

PR #3556, opened 2015-02-02
(DO NOT MERGE) (PUP-3809) Separate code and config directories

nwolfe:
This PR introduces the new $puppetdir directory setting that will be the parent of the existing $confdir and the new $codedir.

The following settings have moved out from under $confdir entirely and are now under $codedir:
* $environmentpath (note that manifests and modules are rooted at this, so they've effectively moved)
* $hiera_config
* $basemodulepath

In addition, the $ssldir has been moved out from under $confdir and is now directly under $puppetdir.

The directory structure now looks like:
```bash
$puppetdir
  $codedir
  $confdir
  $ssldir
```

These changes pass spec tests, but I have not ran acceptance tests yet.

Happy to squash commits if desired.


lib/puppet/defaults.rb
```diff
@@ -34,6 +34,11 @@ def self.default_diffargs
   )
 
   define_settings(:main,
+    :puppetdir => {
+        :default  => nil,
+        :type     => :directory,
+        :desc     => "TODO ncw"
```
nwolfe: @nfagerlund Not sure how we want to describe these directories now.  Suggestions welcome :smiley:

lib/puppet/defaults.rb
```diff
@@ -43,6 +48,11 @@ def self.default_diffargs
           directory, but if it's running as any other user, it defaults to being
           in the user's home directory.",
     },
+    :codedir  => {
+        :default  => "$puppetdir/code",
+        :type     => :directory,
+        :desc     => "TODO ncw",
```
nwolfe: @nfagerlund Not sure how we want to describe these directories now.  Suggestions welcome :smiley:

nwolfe:
Pinging @kylog, @joshcooper, @hlindberg, @haus, @jpartlow, @cprice404, and @jeffmccune.

CC @pcarlisle and @kbarber - FYI this is the work that I pinged you about on the sub-tasks of PUP-3809.

nwolfe:
This PR is based on @whopper's PUP-3632 AIO changes and will need to be rebased once the fallout from those changes is addressed.

whopper:
While looking into an erroring acceptance test, I noticed that confdir is now `/etc/puppetlabs/agent/config` and puppetdir is `/etc/puppetlabs/agent`, but puppet.conf was still under `/etc/puppetlabs/agent`:

    root@pjeqlxxyc4m8t9l:/tmp/file_source_attr.idQs4e# puppet config print confdir -- environment=production
    /etc/puppetlabs/agent/config
    root@pjeqlxxyc4m8t9l:/tmp/file_source_attr.idQs4e# puppet config print puppetdir --environment=production
    /etc/puppetlabs/agent
    root@pjeqlxxyc4m8t9l:/tmp/file_source_attr.idQs4e# ls /etc/puppetlabs/agent/
    auth.conf  code  config  puppet.conf  ssl

roidelapluie:
$puppetdir does not get expanded:


# strace -e trace=file /opt/puppetlabs/agent/bin/puppet agent --test --server localhost 2>&1|grep puppet\\.conf
stat("$puppetdir/config/puppet.conf", 0x7fff13614080) = -1 ENOENT (No such file or directory)


EL7, puppet-agent-0.2.2-1.rpm

lib/puppet/util/run_mode.rb
```diff
@@ -38,6 +38,10 @@ def log_dir
         "$vardir/log"
       end
 
+      def conf_dir
+        "$puppetdir/config"
+      end
+
```
roidelapluie: We have a problem here because in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/settings.rb#L583 we do not extrapolate the variable ( it is actually a chicken-and-egg problem because there is no settings to extrapolate yet.

roidelapluie: A solution would be to use `"#{puppet_dir}/config"` instead of `"$puppetdir/config"`

KevinCorcoran: @justinstoller @ahpook - Here is the answer to `$puppetdir` not expanding - turns out @roidelapluie already solved it for us.

whopper:
@roidelapluie Aha! Thanks a ton for catching that - hugely helpful! @nwolfe now we just need to figure out why puppet.conf isn't ending up in the new config directory.

lib/puppet/util/run_mode.rb
cprice404: I think before we make the change that is being proposed here, we need to understand why the logdir setting a few lines above is not being affected by this.

ahpook: IIRC there is some extra magic about variable interpolation in settings, where only some settings which are "known" at this stage of the initialization can be interpolated... IIRC @jpartlow got down deep into this at one point, maybe as a result of https://tickets.puppetlabs.com/browse/PUP-1015  ??

roidelapluie: On Wed, Feb 04, 2015 at 10:08:09AM -0800, Eric Sorenson wrote:
> > @@ -38,6 +38,10 @@ def log_dir
> >          "$vardir/log"
> >        end
> >  
> > +      def conf_dir
> > +        "$puppetdir/config"
> > +      end
> > +
> 
> IIRC there is some extra magic about variable interpolation in settings, where only some settings which are "known" at this stage of the initialization can be interpolated... IIRC @jpartlow got down deep into this at one point, maybe as a result of https://tickets.puppetlabs.com/browse/PUP-1015  ?? 

This is easy: at any other location of the code the variable is correct,
except in the file I have added to the comment because it is called in
plain ruby, without any parser/etc option. It also means that *at that
particular stage* you can not overwrite --puppetdir with the command line.

> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/puppetlabs/puppet/pull/3556/files#r24104469

-- 
 (o-    Julien Pivotto
 //\    Open-Source Consultant
 V_/_   Inuits - https://www.inuits.eu

fpringvaldsen:
I've functionally reviewed @nwolfe's commits and have verified they're working.

lib/puppet/util/run_mode.rb
cprice404: Yeah, I think I agree with that statement, but it seems like it would imply that the log_dir value a few lines earlier should be equally broken.

roidelapluie: On Wed, Feb 04, 2015 at 11:06:47AM -0800, Chris Price wrote:
> > @@ -38,6 +38,10 @@ def log_dir
> >          "$vardir/log"
> >        end
> >  
> > +      def conf_dir
> > +        "$puppetdir/config"
> > +      end
> > +
> 
> Yeah, I think I agree with that statement, but it seems like it would imply that the log_dir value a few lines earlier should be equally broken.
> 

I don't think so

we NEVER use Puppet::Util::RunMode[:master].log_dir directly. However it
is Puppet::Util::RunMode[:master].conf_dir that is bypassing the
extrapolate stuff.

cprice404: Then we should probably just remove the `log_dir` one, because if it were ever used, it would have the same problem... right?

roidelapluie: On Wed, Feb 04, 2015 at 12:09:16PM -0800, Chris Price wrote:
> > @@ -38,6 +38,10 @@ def log_dir
> >          "$vardir/log"
> >        end
> >  
> > +      def conf_dir
> > +        "$puppetdir/config"
> > +      end
> > +
> 
> Then we should probably just remove the `log_dir` one, because if it were ever used, it would have the same problem... right?

It is used but not directly like this :-)


-- 
 (o-    Julien Pivotto
 //\    Open-Source Consultant
 V_/_   Inuits - https://www.inuits.eu

ferventcoder:
It is my understanding that some of the Windows pathing changes are not quite ironed out yet, is that right?

Iristyle:
Yes, testing Puppet acceptance runs through Beaker via git does not work.

Firstly, it dies on the master as Puppet is not in the PATH.  I've hacked a temporary fix into my fork of Puppet to tackle the PATH issue.  Note I am running against Beaker 2.3.0.
https://github.com/Iristyle/puppet/commit/b140fd9f22049f83c9eabe24c5b416726e333909

After that change is made, it dies in `setup/git/pre-suite/020_PuppetUserAndGroup.rb` on Windows with

```
cmd.exe /c puppet resource user puppet ensure=present
Last 10 lines of output were:
	Error: /File[C:/opt/puppetlabs/agent/cache]/ensure: change from absent to directory failed: Cannot create C:/opt/puppetlabs/agent/cache; parent directory C:/opt/puppetlabs/agent does not exist
	Error: Cannot create C:/var/log/puppetlabs/agent; parent directory C:/var/log/puppetlabs does not exist
	Error: /File[C:/var/log/puppetlabs/agent]/ensure: change from absent to directory failed: Cannot create C:/var/log/puppetlabs/agent; parent directory C:/var/log/puppetlabs does not exist
	Warning: /File[C:/opt/puppetlabs/agent/cache/state]: Skipping because of failed dependencies
	Error: Cannot create C:/var/run/puppetlabs; parent directory C:/var/run does not exist
	Error: /File[C:/var/run/puppetlabs]/ensure: change from absent to directory failed: Cannot create C:/var/run/puppetlabs; parent directory C:/var/run does not exist
	Warning: /File[C:/opt/puppetlabs/agent/cache/lib]: Skipping because of failed dependencies
	Notice: /File[C:/opt/puppetlabs/agent/cache/facts.d]: Dependency File[C:/opt/puppetlabs/agent/cache] has failures: true
	Warning: /File[C:/opt/puppetlabs/agent/cache/facts.d]: Skipping because of failed dependencies
	Error: Could not run: Got 3 failure(s) while initializing: File[C:/opt/puppetlabs/agent/cache]: change from absent to directory failed: Cannot create C:/opt/puppetlabs/agent/cache; parent directory C:/opt/puppetlabs/agent does not exist; File[C:/var/log/puppetlabs/agent]: change from absent to directory failed: Cannot create C:/var/log/puppetlabs/agent; parent directory C:/var/log/puppetlabs does not exist; File[C:/var/run/puppetlabs]: change from absent to directory failed: Cannot create C:/var/run/puppetlabs; parent directory C:/var/run does not exist>
```

Some settings are mangled on the Windows machine - currently investigating.


```
cmd.exe /c puppet agent --configprint vardir
C:/opt/puppetlabs/agent/cache

Administrator@dmiyp7sx7q0zbhu ~
$ cmd.exe /c puppet agent --configprint codedir
C:/ProgramData/PuppetLabs/puppet/etc/code

Administrator@dmiyp7sx7q0zbhu ~
$ cmd.exe /c puppet agent --configprint puppetdir
C:/ProgramData/PuppetLabs/puppet/etc

Administrator@dmiyp7sx7q0zbhu ~
$ cmd.exe /c puppet agent --configprint rundir
C:/var/run/puppetlabs

Administrator@dmiyp7sx7q0zbhu ~
$ cmd.exe /c puppet agent --configprint configdir
invalid setting: configdir

Administrator@dmiyp7sx7q0zbhu ~
$ cmd.exe /c puppet agent --configprint logdir
C:/var/log/puppetlabs/agent

Administrator@dmiyp7sx7q0zbhu ~
$ cmd.exe /c puppet agent --configprint bindir
invalid setting: bindir
```

ferventcoder:
@Iristyle - Will (@whopper) had hit me up about this having some fun with Windows and changes. I just wanted to drop a note here so this didn't get merged in until those changes were also validated against Windows.

ext/debian/puppet.conf
```diff
@@ -0,0 +1 @@
+../../conf/puppet.conf
```
ferventcoder: Symlinks do not work well with Windows. Like at all. Sorry.

ext/gentoo/puppet/puppet.conf
```diff
@@ -0,0 +1 @@
+../../../conf/puppet.conf
```
ferventcoder: Same comment on symlinks as before. Git for Windows doesn't know how to deal with Symlinks.

ext/ips/puppet.conf
```diff
@@ -1,23 +0,0 @@
-
```
ferventcoder: Symlink

ext/redhat/puppet.conf
```diff
@@ -0,0 +1 @@
+../../conf/puppet.conf
```
ferventcoder: and here as well.

conf/puppet.conf
```diff
@@ -0,0 +1,9 @@
+[main]
+    # Where Puppet stores dynamic and growing data.
+    vardir=/opt/puppetlabs/agent/cache
+
+    # The Puppet log directory.
+    logdir=/var/log/puppetlabs/agent
+
+    # Where Puppet PID files are kept.
+    rundir=/var/run/puppetlabs
```
ferventcoder: You may want to create one of these specifically for Windows (unless it is already there).

ext/debian/puppet.conf
ferventcoder: Git Symlinks do not work well with Windows. Like at all. Sorry.

We don't like to see them in any cross platform repository.

ext/gentoo/puppet/puppet.conf
Iristyle: I don't think this is a problem since it only affects non-Windows

ext/debian/puppet.conf
ferventcoder: It's possible that these are never addressed in any context related to Windows, which might be okay. We had issues at one point @jpartlow may remember related to files that were called even during Windows acceptances (setup steps).

Iristyle: I don't think `ext/OS` are relevant to Windows

jpartlow: We pulled symlinks from the acceptance repo in: https://github.com/puppetlabs/puppet/commit/959e0cc8559afb879a10d62425788e729b5b2503 They were also causing problems for @hlindberg in Eclipse, just by being in the repository, as I recall.  I don't know if their being in ext makes either point less of an issue.

lib/puppet/util/run_mode.rb
nwolfe: | A solution would be to use "#{puppet_dir}/config" instead of "$puppetdir/config"

The implementation `"#{puppet_dir}/config"` doesn't honor the explicit `--puppetdir=/foo/bar` when ran from source because it's just calling the `puppet_dir` method which will just return the hard-coded defaults - it won't interpolate my --puppetdir setting.

Trying to come up with a solution that works for both packages and source...

conf/puppet.conf
Iristyle: Yeah, it would appear that on Windows acceptance runs from git, these configuration values are getting picked up and used by Windows.

What necessitated the addition of this file to the PR?

ext/debian/puppet.conf
hlindberg: Yeah. Do not put symlinks anywhere in the repository please.

ferventcoder: Will someone followup with the changes made to hiera that also introduced symlinks? @MikaelSmith pointed them out.

install.rb
```diff
@@ -410,7 +485,7 @@ def install_binfile(from, op_file, target)
 # Change directory into the puppet root so we don't get the wrong files for install.
 FileUtils.cd File.dirname(__FILE__) do
   # Set these values to what you want installed.
-  configs = glob(%w{conf/auth.conf})
+  configs = glob(%w{conf/auth.conf conf/puppet.conf})
```
Iristyle: This breaks Windows

haus: How does this break windows?

Iristyle: By adding `puppet.conf` to Windows, we get errant paths injected into the environment for `vardir`, `logdir` and `rundir`.  As soon as `puppet.conf` is gone, things are back to normal.

See https://github.com/puppetlabs/puppet/pull/3556#issuecomment-73300673

conf/puppet.conf
haus: Can we just use [master] instead of main. Master should be the only runmode that needs overriding.

haus: https://github.com/nwolfe/puppet/pull/17 opened to address.

nwolfe:
Sorry for the force-push updates - trying to stay on top of origin/master like a hawk

joshcooper:
I have pushed the commits in this PR to the https://github.com/puppetlabs/puppet/tree/aio branch and we'll be running the pipeline against it. Going to close this PR and resubmit once it passes on rhel7.

PR #3557, opened 2015-02-02
(PUP-3632) Update RPM packaging spec for AIO changes

whopper:
This commit updates the RPM packaging specfile for pathing
changes brought in by the all-in-one agent switchover.

### DIFF
--- a/ext/redhat/puppet.spec.erb
+++ b/ext/redhat/puppet.spec.erb
@@ -114,17 +114,18 @@ find examples/ -type f | xargs --no-run-if-empty chmod a-x
 rm -rf %{buildroot}
 ruby install.rb --destdir=%{buildroot} --quick --no-rdoc --sitelibdir=%{puppet_libdir}
 
-install -d -m0755 %{buildroot}%{_sysconfdir}/puppet/environments/example_env/manifests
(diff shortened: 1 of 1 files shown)


ext/redhat/puppet.spec.erb
```diff
@@ -336,8 +338,8 @@ fi
 if [ "$1" -ge 1 ]; then
   # The pidfile changed from 0.25.x to 2.6.x, handle upgrades without leaving
   # the old process running.
-  oldpid="%{_localstatedir}/run/puppet/puppetmasterd.pid"
-  newpid="%{_localstatedir}/run/puppet/master.pid"
+  oldpid="%{_localstatedir}/run/puppetlabs/puppetmasterd.pid"
```
kylog: The oldpid is probably in the same old place?

And do we even need this bit anymore? /cc @haus

haus:
Nothing stands out as crazy. This should work.

ext/redhat/puppet.spec.erb
haus: No, we don't need this anymore.

haus: I don't know how much cleanup we should be doing here.

kylog:
:+1:

PR #3559, opened 2015-02-03
Revert "(PUP-3632) Update paths and defaults for the unified agent layou...

whopper:
Reverts puppetlabs/puppet#3467 until the AIO pipeline is ready to go.


haus:
:+1:

PR #3561, opened 2015-02-03
(PUP-3947) Let infer_set of Hash return a Struct

thallgren:
Changes the infer_set_Hash method so that it returns a PStructType
rather than a PHashType when all keys of the hash are found to be
non empty strings (assignable to String[1]).


hlindberg:
needs a test for the change that fixes the problem of Struct not being assignable to Hash to avoid future regressions

PR #3562, opened 2015-02-03
(PUP-3818) Windows acceptance on Ruby 2.1.5

Iristyle:
- Update the pointers to the puppet-win32-ruby to the appropriate
   SHAs so that Windows acceptance tests can run against Ruby 2.1.5


Iristyle:
Depends on #3540

PR #3563, opened 2015-02-03
(CODEMGMT-69) Update Acceptance Test

cowofevil:
PMT is no longer using the Modulefile and the acceptance test needed to be
updated to use a "metadata.json" file instead. Also, the test has been
updated with current Beaker syntax in increase readability.

### DIFF
--- a/acceptance/tests/modules/build/build_basic.rb
+++ b/acceptance/tests/modules/build/build_basic.rb
@@ -1,47 +1,70 @@
-begin test_name "puppet module build (basic)"
+test_name 'CODEMGMT-69 - Build a Module Using "metadata.json" Only'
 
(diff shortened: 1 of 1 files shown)


andersonmills:
Confirmed acceptance test passes on current master.

kylog:
@cowofevil and @andersonmills too late for this PR, but please reference PUP tickets in the commit messages for PRs against the puppet repository. It helps with release scripts, and ensures that we are always referencing tickets in open jira projects from our open github projects.

andersonmills:
@kylog That appears to mean that all PMT tickets have to have a PUP clone if they don't originate as PUP tickets. Is that accurate?

PR #3565, opened 2015-02-03
(PUP-3848) Bump Facter/Hiera dependencies

ferventcoder:
Bump Facter up to at least 2.0 but less than 4 to allow for Facter
3, which is expected to be out during Puppet 4.x series. Bump Hiera
to allow it to take a dependency from 1.0 but less than 3. Without
this change Facter 3.x and Hiera 2.x will not be able to be
resolved as dependencies.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,8 +24,8 @@ platforms :ruby do
 end
 
 gem "puppet", :path => File.dirname(__FILE__), :require => false
-gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 1.6', '< 3'])
-gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '~> 1.0')
+gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 2.0', '< 4'])
+gem "hiera", *location_for(ENV['HIERA_LOCATION'] || ['> 1.0', '< 3'])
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -24,8 +24,8 @@ platforms :ruby do
 end
 
 gem "puppet", :path => File.dirname(__FILE__), :require => false
-gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 1.6', '< 3'])
-gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '~> 1.0')
+gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 2.0', '< 3'])
```
kylog: Oh right, totally forgot about this change. Thanks!

I think facter should be `'< 4'` so that puppet 4 can be run against facter 3.

ferventcoder: Good point. Will adjust.

ferventcoder: Done.

kylog:
:+1:

ferventcoder:
To my knowledge there is nothing holding this up...

ferventcoder:
OH wait!

ferventcoder:
:white_check_mark:

ferventcoder:
@haus good here?

haus:
@ferventcoder looks fine to me.

PR #3574, opened 2015-02-04
(PUP-3962) Test for '.' module directories

andersonmills:
Before this test, the spec tests don't include negative tests for module
directories starting with '.'. After this commit, we test for those
types of directories also. This tests the original intention of
PUP-3678.


adreyer:
+1

PR #3575, opened 2015-02-04
(PUP-3960) Verify rpms installed for rpm_util helper

johnduarte:
This commit adds a 'setup' method to the Puppet::Acceptance::RpmUtils module.
This method installs the rpms required by the other methods in the module.
Currently, these rpms are createrepo and rpm-build.

The acceptable_exit_codes have been expanded in clean_rpm to allow this method
to gracefully continue if the given rpms are not on the system when the
clean_rpm method is called.

The new 'setup' method is also added to the beginning of the setup_rpm and
send_rpm methods to ensure that their requirements are satisfied.


MikaelSmith:
This fixes the failure for puppet-server acceptance runs?

It looks like this could completely replace the addition of 001_PkgBuildSetup.rb in https://github.com/puppetlabs/puppet/commit/6a97df1ed4a7665dd2a9d1aad958f0fb27092872

johnduarte:
@MikaelSmith alas, this does not fix the test failure. I then get separate failures.

{code}
Failed Tests Cases:
  Test Case tests/resource/package/common_package_name_in_different_providers.rb reported: #<Minitest::Assertion: Expected /Package\[guid\]\/ensure:\ created/ to match "clzvjxiw3o4mahl.delivery.puppetlabs.net: \e[mNotice: Compiled catalog for clzvjxiw3o4mahl.delivery.puppetlabs.net in environment production in 0.81 seconds\e[0m\n\e[0;32mInfo: Applying configuration version '1423092706'\e[0m\n\e[mNotice: /Stage[main]/Main/Package[gem-guid]/ensure: created\e[0m\n\e[
0;32mInfo: Class[Main]: Unscheduling all events on Class[Main]\e[0m\n\e[mNotice: Applied catalog in 2.82 seconds\e[0m\n".>
{code}

Then, upon additional runs on the same system, I receive the following:
{code}
Failed Tests Cases:
  Test Case tests/resource/package/common_package_name_in_different_providers.rb reported: #<Minitest::Assertion: </^guid /> expected to not match
<"bigdecimal (1.2.4)\ndeep_merge (1.0.0)\nffi (1.9.6)\nguid (0.1.1)\nio-console (0.4.2)\njson (1.8.1)\nminitest (4.7.5)\npsych (2.0.5)\nrake (10.1.0)\nrdoc (4.1.0)\nstomp (1.3.3)\ntest-unit (2.1.5.0)\n">.>
{code}

jeffmccune:
FWIW I submitted a similar fix to this issue (before Ruth informed me about PUP-3960) at https://github.com/puppetlabs/puppet/pull/3601

camlow325:
@MikaelSmith and @kylog - even if this doesn't ultimately get this test passing, this seems like a good change.  Any reason we shouldn't commit this?  I'd like to avoid creating any workarounds for this in the Puppet Server acceptance pre-suite if this is ultimately the right change to make.

kylog:
My only qs are: 
* How does this compare with Jeff's change at #3601? 
* Does this allow us to remove the step that Michael suggested above?

I haven't studied enough to answer either of those questions.

camlow325:
Seems like these are all similar, but I'm not familiar with the details either.  I'd be comfortable letting @johnduarte make the call on the best approach.  @johnduarte, what do you think?

johnduarte:
@kylog 

I believe this PR makes @jeffmccune #3601 PR unnecessary and addresses the concerns he pointed out in https://github.com/puppetlabs/puppet-server/pull/398.
It also allows for the removal of the setup step as @MikaelSmith noted.

I have pushed a new commit removing the setup step.

I still get the same errors as noted above. I would appreciate it if someone can verify my results.

cc @joshcooper , @camlow325

kylog:
@johnduarte I can't make complete sense of the failures you snipped. We should sync up on how you're reproducing those failures.

However, this change looks like a standalone Good Thing, so I'm merging away.

kylog:
Doh! Well that error reproduces in CI. I'll take a look.

acceptance/lib/puppet/acceptance/rpm_util.rb
```diff
@@ -4,14 +4,24 @@ module RpmUtils
       # Utilities for creating a basic rpm package and using it in tests
       @@defaults = {:repo => '/tmp/rpmrepo', :pkg => 'mypkg', :publisher => 'tstpub.lan', :version => '1.0'}
 
+      def setup(agent)
+        required_packages = ['createrepo', 'rpm-build']
+        required_packages.each do |pkg|
+          unless ((on agent, "#{pkg} --version", :acceptable_exit_codes => (0..255)).exit_code == 0) then
```
kylog: On reflection, this should perhaps be using `yum list installed #{pkg}`.  E.g. the rpm-build package doesn't install a tool called rpm-build, so `rpm-build --version` will always be an error.

johnduarte: @kylog I have added https://github.com/puppetlabs/puppet/pull/3664 to fix this . thanks

camlow325: @kylog, @johnduarte Does the test now pass with that change or still more work to do?

kylog: @camlow325 not sure yet. Will keep you posted.

johnduarte: FYI: The test now passes for me

cc @kylog @camlow325

PR #3577, opened 2015-02-05
(maint) fix tests that incorrectly use host['user']

er0ck:
Beaker uses these for user/group on test-runner.  Ensure the tests use
the output of puppet agent/master --configprint so manifests get created
and run correctly using the puppet user/group.
This is required for the AIO pipelines.


kurtwall:
LGTM

PR #3579, opened 2015-02-05
(PUP-3821) Don't check Windows admin? with Facter

Iristyle:
- The previous method for checking whether a user is an admin? on
   Windows can cause issues loading Puppet when the --cfacter switch is
   applied.  The error message is:

   Error: Could not initialize global default settings: facter has
   already evaluated facts.

   Detect that the Windows major version is at least 6 (Vista+) as a
   shortcut to whether or not TokenElevation is available on the current
   operating system.

   An alternative approach would rewrite supports_elevated_security to
   execute similarly to elevated_security?, but watch only for
   ERROR_NO_SUCH_PRIVILEGE.  This would be less code, but several
   additional API calls to clone the current process token and
   call GetTokenInformation against it.

### DIFF
--- a/lib/puppet/util/windows/process.rb
+++ b/lib/puppet/util/windows/process.rb
@@ -204,6 +204,36 @@ module Puppet::Util::Windows::Process
   end
   module_function :elevated_security?
 
+  def windows_major_version
+    ver = 0
+
+    FFI::MemoryPointer.new(OSVERSIONINFO.size) do |os_version_ptr|
+      os_version = OSVERSIONINFO.new(os_version_ptr)
+      os_version[:dwOSVersionInfoSize] = OSVERSIONINFO.size
+
+      result = GetVersionExW(os_version_ptr)
+
+      if result == FFI::WIN32_FALSE
+        raise Puppet::Util::Windows::Error.new("GetVersionEx failed")
+      end
+
+      ver = os_version[:dwMajorVersion]
+    end
+
+    ver
+  end
+  module_function :windows_major_version
+
+  # Returns whether or not the OS has the ability to set elevated
+  # token information.
+  #
+  # Returns true on Windows Vista or later, otherwise false
+  #
+  def supports_elevated_security?
+    windows_major_version >= 6
+  end
+  module_function :supports_elevated_security?
+
   ABOVE_NORMAL_PRIORITY_CLASS = 0x0008000
   BELOW_NORMAL_PRIORITY_CLASS = 0x0004000
   HIGH_PRIORITY_CLASS         = 0x0000080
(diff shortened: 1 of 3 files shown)


lib/puppet/util/windows/process.rb
```diff
@@ -351,4 +382,32 @@ class TOKEN_ELEVATION < FFI::Struct
...
+      :szCSDVersion, [:wchar, 128]
+    )
+  end
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/ms724451(v=vs.85).aspx
+  # BOOL WINAPI GetVersionEx(
+  #   _Inout_  LPOSVERSIONINFO lpVersionInfo
+  # );
+  ffi_lib :kernel32
+  attach_function_private :GetVersionExW,
```
Iristyle: I thought about yanking this out into `os.rb` or similar, but given that would require more touches, I moved away from that solution.

MikaelSmith:
:+1:

peterhuene:
:+1: Thanks for tackling this.

lib/puppet/util/windows/process.rb
```diff
@@ -204,6 +204,37 @@ def elevated_security?
...
+
+    FFI::MemoryPointer.new(OSVERSIONINFO.size) do |os_version_ptr|
+      os_version = OSVERSIONINFO.new(os_version_ptr)
+      os_version[:dwOSVersionInfoSize] = OSVERSIONINFO.size
+
+      result = GetVersionExW(os_version_ptr)
+
+      if result == FFI::WIN32_FALSE
+        raise Puppet::Util::Windows::Error.new(
+          "GetVersionEx(#{os_version_ptr})")
```
peterhuene: I'm curious, what does the interpolation of `os_version_ptr` give?

lib/puppet/util/windows/process.rb
kylog: +1 on that reasoning. It can be factored out if/when there's another need.

lib/puppet/util/windows/process.rb
Iristyle: It's not super useful, aside from letting you know that it's a live Ruby object.  There's not a default `to_s` for an FFI struct, for reasons I'm sure you can understand.

```
#<Puppet::Util::Windows::Process::OSVERSIONINFO:0x0000000314dc98>
```

Iristyle:
Hold tight on merging this - some mocha expectations need cleaning up.

lib/puppet/util/windows/process.rb
peterhuene: Just wondering at its utility in the error message, then.  Does `Windows::Error` have a way at getting at the last error code?  I figure that'd be more use than `os_version_ptr`, no?

joshcooper:
## 2008r2

### elevated

```
C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin>irb
=> true
irb(main):002:0> Puppet::Util::Windows::User.admin?
=> true
irb(main):003:0> Puppet::Util::Windows::Process.supports_elevated_security?
=> true
```

### not elevated

```
C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin>irb
=> true
irb(main):002:0> Puppet::Util::Windows::User.admin?
=> false
irb(main):003:0> Puppet::Util::Windows::Process.supports_elevated_security?
=> true
```

lib/puppet/util/windows/process.rb
peterhuene: Ah, I see that `Windows::Error` does get the last error from FFI.

Iristyle: Yup, we always call `GetLastError` as part of `Windows::Error` (by way of `FFI.errno`)

We also dump the pointers in our other error messages.  It *might* help to let you know that an allocation failed in a low memory situation.

joshcooper: I don't know.. I think I'd kill the ptr in the message

Iristyle: Updated to remove ptr from message.  Note we still have at least these guys kicking around that use ptrs in error messages.

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/file.rb#L172-L173

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/process.rb#L51-L52

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/process.rb#L89-L90

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/process.rb#L108-L109

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/process.rb#L119-L120

PR #3580, opened 2015-02-05
(PUP-3679) Drop use of facter's 'ps' fact.

kylog:
Facter 3 is planning to drop the 'ps' fact (which is only used here
as far as anyone can tell), so this commit simply moves that logic
for different ps versions into the one provider that relies on it.

### DIFF
--- a/lib/puppet/provider/service/base.rb
+++ b/lib/puppet/provider/service/base.rb
@@ -11,15 +11,28 @@ Puppet::Type.type(:service).provide :base, :parent => :service do
 
   commands :kill => "kill"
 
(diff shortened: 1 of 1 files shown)


kylog:
Oops stand by on an amend of the commit message :)

kylog:
This is just a straight port of https://github.com/puppetlabs/facter/blob/2.4.0/lib/facter/ps.rb#L14-L26.

Note that the Windows resolution is omitted since the provider isn't applicable on Windows.

peterhuene:
:+1:

PR #3584, opened 2015-02-06
DO NOT MERGE (QENG-1800) Install MSI during package acceptance

joshcooper:
Previously it was not possible to run windows acceptance against FOSS
MSI packages. This commit adds a step to install the puppet-agent MSI. It
depends on https://github.com/puppetlabs/beaker/pull/674

With this commit you can do:

    $ bundle exec rake ci:test:packages \
        CONFIG=config/nodes/win2008r2-rubyx64.yaml \
        SHA=<SHA> VERSION=3.7.4-1303

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -232,6 +232,20 @@ module Puppet
           on host, "#{gem} source --add #{gem_source}"
         end
       end
+
+      def install_puppet_from_msi( host, opts )
+        if not link_exists?(opts[:url])
+          raise "Puppet does not exist at #{opts[:url]}!"
+        end
+
+        # `start /w` blocks until installation is complete, but needs to be wrapped in `cmd.exe /c`
+        on host, "cmd.exe /c start /w msiexec /qn /i #{opts[:url]}"
+
+        # make sure install is sane, beaker has already added puppet and ruby
+        # to PATH in ~/.ssh/environment
+        on host, puppet('--version')
+        on host, 'ruby --version'
+      end
     end
   end
 end
--- a/acceptance/setup/packages/pre-suite/010_Install.rb
+++ b/acceptance/setup/packages/pre-suite/010_Install.rb
@@ -48,5 +48,15 @@ AGENT_PACKAGES = {
 install_packages_on(master, MASTER_PACKAGES)
 install_packages_on(agents, AGENT_PACKAGES)
 
+agents.each do |agent|
+  if agent['platform'] =~ /windows/
+    arch = agent[:ruby_arch] || 'x86'
+    base_url = "http://builds.puppetlabs.lan/puppet/#{ENV['SHA']}/artifacts/windows"
+    filename = "puppet-agent-#{ENV['VERSION']}-#{arch}.msi"
+
+    install_puppet_from_msi(agent, :url => "#{base_url}/#{filename}")
+  end
+end
+
 configure_gem_mirror(hosts)
 


Iristyle:
Out of curiosity, is there not a way to correlate a `VERSION` to a `SHA` without having to supply both?

I had to look at the code to figure out what `SHA` is used for.... and as it turns out, it looks like there's only ever a single x86 and single x64 MSI in the artifacts directory for a given `SHA`

acceptance/setup/packages/pre-suite/010_Install.rb
```diff
@@ -48,5 +48,15 @@
 install_packages_on(master, MASTER_PACKAGES)
 install_packages_on(agents, AGENT_PACKAGES)
 
+agents.each do |agent|
+  if agent['platform'] =~ /windows/
+    arch = agent[:ruby_arch] || 'x86'
+    base_url = "http://builds.puppetlabs.lan/puppet/#{ENV['SHA']}/artifacts/windows"
+    filename = "puppet-agent-#{ENV['VERSION']}-#{arch}.msi"
```
Iristyle: We only recently added `-#{arch}` to x86.

So you can see builds from 2/5 have it:
http://builds.puppetlabs.lan/puppet/09b0ece5f6f8a5c836f0fa518ca7455126825293/artifacts/windows/

But not from 1/29:
http://builds.puppetlabs.lan/puppet/d75d56332b53e7a252cd1550736c2b9964ec9b14/artifacts/windows/

Might want to add some safeguards here for when `-x86` wasn't part of the package name, for the sake of backward compat.

joshcooper: I thought about that, but this change would be in puppet/master for 4.0 where we know the MSI will have `-x86` (as opposed to being able to acceptance test earlier versions of puppet)

joshcooper:
It is possible to do `git describe <SHA>` and then strip out the short git:

```
$ git describe HEAD^1
3.7.4-1307-ge91e31f
```
But that requires that you have the commit in your current branch... We must do this somewhere else? /cc @haus @melissa

acceptance/setup/packages/pre-suite/010_Install.rb
Iristyle: Fair enough.

justinstoller:
You may want to target this work also at the acceptance/setup/aio folder.

joshcooper:
@justinstoller good point. I was hoping to get testing against packages working in puppet/master before the uber puppet PR is merged. Then look at AIO as a second step.

joshcooper:
@Iristyle I talked with @haus about parsing the YAML to get the `version` and he said the AIO pipeline is not creating this file currently. And it would probably be easier for the pipeline to do a `git describe` at the beginning, strip the commit, and pass the version through. I'm thinking we come back to this issue, so the caller doesn't need to specify `VERSION`

acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -232,6 +232,15 @@ def configure_gem_mirror(hosts)
...
         end
       end
+
+      def install_puppet_from_msi( host, opts )
+        if not link_exists?(opts[:url])
+          raise "Puppet does not exist at #{opts[:url]}!"
+        end
+
+        # beaker has already added puppetbindir to PATH in ~/.ssh/environment
+        on host, "cmd.exe /c start /w msiexec /qn /i #{opts[:url]}"
```
joshcooper: After installing the MSI, we should execute `puppet --version` and `ruby --version` to ensure it was installed correctly, and so that later steps don't fail in obscure ways.

acceptance/setup/packages/pre-suite/010_Install.rb
joshcooper: @melissa @haus @Iristyle I'm wondering if it would instead be better for the MSI build automation to create an MSI whose version is full `git describe`, e.g. `3.7.4-1303-ge96f55b`, instead of stripping the short sha, e.g. `3.7.4-1303`? I ask because I imagine the latter is something we're already passing through the AIO pipeline, maybe?

ferventcoder: The way it is done now, if you pass in a name, it will take it wholesale without attempting to change it. https://github.com/puppetlabs/puppet_for_the_win/blob/master/tasks/windows/windows.rake#L116

joshcooper:
Plan is to:

1. Modify the MSI packaging job to use `git describe` of `puppet-agent` in the package name, see https://github.com/puppetlabs/packaging/pull/374
2. Modify the AIO pipeline to include the `PROJECT_VERSION` as a build parameter, which will match the above.
3. Modify this PR to use the `PROJECT_VERSION` by default, but allow an alternate version to be specified on the command line.

Iristyle:
I think we want to retarget this at the `aio` branch?

PR #3587, opened 2015-02-09
(PUP-3981) Add Acceptance Test for Modulefile

cowofevil:
Create a new acceptance test for the PMT tool. Verify that the user will
receive an appropriate warning message if a "Modulefile" is present when
building a module. (The "Modulefile" support has been removed from PMT.)


andersonmills:
Reviewed and tested this acceptance test, and ran the full suite as well. :+1:

PR #3588, opened 2015-02-09
(PUP-3274) Change Closure.call to not take scope as parameter

thallgren:
This commit was merged to master 3 months ago. It is also needed on stable since it changes the way blocks are called from puppet functions written in Ruby.

### DIFF
--- a/lib/puppet/functions/each.rb
+++ b/lib/puppet/functions/each.rb
@@ -62,7 +62,7 @@ Puppet::Functions.create_function(:each) do
   def foreach_Hash_1(hash, pblock)
     enumerator = hash.each_pair
     hash.size.times do
-      pblock.call(nil, enumerator.next)
+      pblock.call(enumerator.next)
     end
     # produces the receiver
     hash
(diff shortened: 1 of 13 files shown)


kylog:
I'm confused. By the ticket number, this looks like "remove old parser" was just merged to puppet/stable?

joshcooper:
@thallgren won't this break existing functions in 3.7? I don't think we can make this change here.

hlindberg:
This will not break functions in 3.7 because the change is only for calling lambdas. The functions in puppet that use iterations etc are updated in this PR. The fix was merged to master originally but should have been made to stable to ensure that the block/lambda API is the same on 3x and 4x.

hlindberg:
@thallgren I added the PUP-3987 ticket for the backport/cherry-pick. Unfortunately we now have a commit on stable with the wrong ticket number that will be a small problem for the person that lines up ticket with commits since PUP-3274 is the mega "remove old parser" ticket. I updated the title of this PR though to help, and I made a comment on the ticket about the mislabeled commit.

PR #3592, opened 2015-02-10
(PUP-3978) Pass block using Ruby semantics, not as normal parameter

thallgren:
Passing a block as a normal parameter messed up the notion of
optional parameters since the last parameter could be a required
block. It also made the calling convention somewhat unorthodox
from a ruby developers point of view.

This commit introduces a new class 'PuppetProc' derived from the Ruby
'Proc'. It wrapps the normal closure that would otherwise be passed in
a parameter and enables it to instead be passed as a block, thus
making it possible to use standard calls like 'block_given?' and
'yield' to execute a Puppet lambda from Ruby.

The 'TypeCalculator' will infer a 'PuppetProc' to the type of its
closure. It is also capable of inferring a normal ruby 'Proc' using the
information returned by its 'parameters' method so that all parameters
are of type 'Any' and the min/max is controlled by the :req, :opt, and
:rest indicators of the parameter info arrays.

### DIFF
--- a/lib/puppet/functions.rb
+++ b/lib/puppet/functions.rb
@@ -314,12 +314,8 @@ module Puppet::Functions
       if @block_type.nil?
         @block_type = type
         @block_name = name
-
-        # mark what should be picked for this position when dispatching. This is the size of
-        # the @names array since the block is required to appear last
-        @weaving << @names.size()
       else
-        raise ArgumentError, "Attempt to redefine block"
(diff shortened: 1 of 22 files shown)


hlindberg:
Please write the first line of the commit messages in imperative form "Add the...", "Change..." etc. That is to complete the sentence "This commit will..." (not blocking merge because of this, but if there are other changes, that should be done as well).

hlindberg:
Rest looks fine, merging.

PR #3593, opened 2015-02-10
(PUP-3991) Implement scanf function for classic parser

DavidS:
See https://tickets.puppetlabs.com/browse/PUP-3991

### DIFF
--- a/lib/puppet/parser/functions/scanf.rb
+++ b/lib/puppet/parser/functions/scanf.rb
@@ -1,3 +1,5 @@
+require 'scanf'
+
 Puppet::Parser::Functions::newfunction(
   :scanf,
   :type => :rvalue,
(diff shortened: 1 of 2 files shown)


lib/puppet/parser/functions/scanf.rb
```diff
@@ -8,27 +10,17 @@
 are similar but not identical to the formats used in Puppet's `sprintf` function.
 
 This function takes two mandatory arguments: the first is the String to convert, and the second
-the format String. A parameterized block may optionally be given, which is called with the result
```
hlindberg: Unfortunately the documentation extraction extracts the text for the function from the 3x implementation at the moment. The documentation text must describe both the future parser (4x) version and the 3x version. Suggest rephrasing to language such as "When used with the future parser ..." for the things that are not available on 3x classic.

hlindberg:
This was merged to master unintentionally - it should have gone to stable. I cherry picked it over to stable.

PR #3594, opened 2015-02-11
DO NOT MERGE (QENG-1800) Install MSI during package acceptance

joshcooper:
Previously it was not possible to run windows acceptance against FOSS
MSI packages. This commit adds a step to install the puppet-agent MSI. It
depends on https://github.com/puppetlabs/beaker/pull/674

With this commit you can do:

    $ bundle exec rake ci:test:packages \
        CONFIG=config/nodes/win2008r2-rubyx64.yaml \
        SHA=<SHA> VERSION=3.7.4-1303

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -232,6 +232,20 @@ module Puppet
           on host, "#{gem} source --add #{gem_source}"
         end
       end
+
+      def install_puppet_from_msi( host, opts )
+        if not link_exists?(opts[:url])
+          raise "Puppet does not exist at #{opts[:url]}!"
+        end
+
+        # `start /w` blocks until installation is complete, but needs to be wrapped in `cmd.exe /c`
+        on host, "cmd.exe /c start /w msiexec /qn /i #{opts[:url]}"
+
+        # make sure install is sane, beaker has already added puppet and ruby
+        # to PATH in ~/.ssh/environment
+        on host, puppet('--version')
+        on host, 'ruby --version'
+      end
     end
   end
 end
--- a/acceptance/setup/packages/pre-suite/010_Install.rb
+++ b/acceptance/setup/packages/pre-suite/010_Install.rb
@@ -48,5 +48,15 @@ AGENT_PACKAGES = {
 install_packages_on(master, MASTER_PACKAGES)
 install_packages_on(agents, AGENT_PACKAGES)
 
+agents.each do |agent|
+  if agent['platform'] =~ /windows/
+    arch = agent[:ruby_arch] || 'x86'
+    base_url = "http://builds.puppetlabs.lan/puppet/#{ENV['SHA']}/artifacts/windows"
+    filename = "puppet-agent-#{ENV['VERSION']}-#{arch}.msi"
+
+    install_puppet_from_msi(agent, :url => "#{base_url}/#{filename}")
+  end
+end
+
 configure_gem_mirror(hosts)
 


joshcooper:
Supercedes #3584

joshcooper:
Still need to verify that the environment variable matches what the AIO pipeline will pass through. Note it is the SHA/version of the `puppet-agent` package, not for the `puppet` component. See https://github.com/puppetlabs/puppet/pull/3584#issuecomment-73562557

joshcooper:
Also I'm not sure this was the correct thing to do... This commit adds to the setup step for `packages` type acceptance, not `aio`...

acceptance/setup/packages/pre-suite/010_Install.rb
```diff
@@ -48,5 +48,15 @@
 install_packages_on(master, MASTER_PACKAGES)
 install_packages_on(agents, AGENT_PACKAGES)
 
+agents.each do |agent|
+  if agent['platform'] =~ /windows/
+    arch = agent[:ruby_arch] || 'x86'
+    base_url = "http://builds.puppetlabs.lan/puppet/#{ENV['SHA']}/artifacts/windows"
+    filename = "puppet-agent-#{ENV['VERSION']}-#{arch}.msi"
```
Iristyle: Make sure that this follows the convention of puppet-agent, which should be something like:

`puppet-agent-0.2.4.8.g7cdf605-x86.msi`

acceptance/setup/packages/pre-suite/010_Install.rb
```diff
@@ -48,5 +48,15 @@
 install_packages_on(master, MASTER_PACKAGES)
 install_packages_on(agents, AGENT_PACKAGES)
 
+agents.each do |agent|
+  if agent['platform'] =~ /windows/
+    arch = agent[:ruby_arch] || 'x86'
+    base_url = "http://builds.puppetlabs.lan/puppet/#{ENV['SHA']}/artifacts/windows"
```
joshcooper: This needs to be `puppet-agent` in the URL not `puppet`

acceptance/setup/packages/pre-suite/010_Install.rb
Iristyle: Note that `LATEST-win` previously went with a convention like

`4.0.0-rc2-8-gcb8b8d5`

See
https://enterprise.delivery.puppetlabs.net/4.0/ci-ready/LATEST-win

acceptance/setup/packages/pre-suite/010_Install.rb
Iristyle: I have modified this locally as 

```ruby
base_url = ENV['MSI_BASE_URL'] || "http://builds.puppetlabs.lan/puppet/#{ENV['SHA']}/artifacts/windows"
```

acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -232,6 +232,20 @@ def configure_gem_mirror(hosts)
...
         end
       end
+
+      def install_puppet_from_msi( host, opts )
+        if not link_exists?(opts[:url])
+          raise "Puppet does not exist at #{opts[:url]}!"
+        end
+
+        # `start /w` blocks until installation is complete, but needs to be wrapped in `cmd.exe /c`
+        on host, "cmd.exe /c start /w msiexec /qn /i #{opts[:url]}"
```
Iristyle: Missing `/L*V C:\\Windows\\Temp\\Puppet-Install.log`

PR #3595, opened 2015-02-11
(PUP-4000) Make optional Struct value imply optional key

thallgren:
This commit changes the TypeCalculator so that a Struct with optional
values becomes assignable from a Struct where the keys corresponding
to those values are missing. It also changes the instance_of logic so
that a Hash missing such keys is considered an instance of the struct.

### DIFF
--- a/lib/puppet/pops/types/type_calculator.rb
+++ b/lib/puppet/pops/types/type_calculator.rb
@@ -1200,7 +1200,7 @@ class Puppet::Pops::Types::TypeCalculator
       size_t2 = t2.size_type || @collection_default_size_t
       size_t = Types::PIntegerType.new
       elements = t.elements
-      size_t.from = elements.reduce(0) { |c, e| assignable?(e.type, @nil_t) ? c : c + 1 }
+      size_t.from = elements.count {|e| !assignable?(e.type, @nil_t) }
       size_t.to = elements.size
       # compatible size
       # hash key type must be string of min 1 size
--- a/lib/puppet/pops/types/type_calculator.rb
+++ b/lib/puppet/pops/types/type_calculator.rb
@@ -1192,22 +1192,23 @@ class Puppet::Pops::Types::TypeCalculator
   # @api private
   #
   def assignable_PStructType(t, t2)
-    return true if t == t2 || t.elements.empty? && (t2.is_a?(Types::PHashType))
-    h = t.hashed_elements
     if t2.is_a?(Types::PStructType)
+      h = t.hashed_elements
       h2 = t2.hashed_elements
-      h.size == h2.size && h.all? {|k, v| assignable?(v, h2[k]) }
+      (h2.keys - h.keys).empty? && h.all? {|k, v| v2 = h2[k]; assignable?(v, v2.nil? ? @nil_t : v2) }
     elsif t2.is_a?(Types::PHashType)
       size_t2 = t2.size_type || @collection_default_size_t
       size_t = Types::PIntegerType.new
-      size_t.from = size_t.to = h.size
+      elements = t.elements
+      size_t.from = elements.reduce(0) { |c, e| assignable?(e.type, @nil_t) ? c : c + 1 }
+      size_t.to = elements.size
       # compatible size
       # hash key type must be string of min 1 size
       # hash value t must be assignable to each key
       element_type = t2.element_type
       assignable_PIntegerType(size_t, size_t2) &&
-        assignable?(@non_empty_string_t, t2.key_type) &&
-        h.all? {|k,v| assignable?(v, element_type) }
+        (size_t2.to == 0 || assignable?(@non_empty_string_t, t2.key_type)) &&
+          elements.all? {|e| assignable?(e.type, element_type) }
     else
       false
     end
--- a/spec/unit/pops/types/type_calculator_spec.rb
+++ b/spec/unit/pops/types/type_calculator_spec.rb
@@ -85,6 +85,14 @@ describe 'The type calculator' do
     Puppet::Pops::Types::TypeFactory.any()
   end
 
+  def optional_t(t)
+    Puppet::Pops::Types::TypeFactory.optional(t)
+  end
+
+  def undef_t
+    Puppet::Pops::Types::TypeFactory.undef
+  end
+
   def unit_t
     # Cannot be created via factory, the type is private to the type system
     Puppet::Pops::Types::PUnitType.new
@@ -1068,6 +1076,30 @@ describe 'The type calculator' do
         calculator.assignable?(struct2, struct1).should == true
       end
 
+      it 'should accept matching structs with less elements when unmatched elements are optional' do
+        struct1 = struct_t({'a'=>Integer, 'b'=>Integer, 'c'=>optional_t(Integer)})
+        struct2 = struct_t({'a'=>Integer, 'b'=>Integer})
+        calculator.assignable?(struct1, struct2).should == true
+      end
+
+      it 'should reject matching structs with more elements even if excess elements are optional' do
+        struct1 = struct_t({'a'=>Integer, 'b'=>Integer})
+        struct2 = struct_t({'a'=>Integer, 'b'=>Integer, 'c'=>optional_t(Integer)})
+        calculator.assignable?(struct1, struct2).should == false
+      end
+
+      it 'should accept matching structs where one is more general than the other with respect to optional' do
+        struct1 = struct_t({'a'=>Integer, 'b'=>Integer, 'c'=>optional_t(Integer)})
+        struct2 = struct_t({'a'=>Integer, 'b'=>Integer, 'c'=>Integer})
+        calculator.assignable?(struct1, struct2).should == true
+      end
+
+      it 'should reject matching structs where one is more special than the other with respect to optional' do
+        struct1 = struct_t({'a'=>Integer, 'b'=>Integer, 'c'=>Integer})
+        struct2 = struct_t({'a'=>Integer, 'b'=>Integer, 'c'=>optional_t(Integer)})
+        calculator.assignable?(struct1, struct2).should == false
+      end
+
       it 'should accept matching structs where one is more general than the other' do
         struct1 = struct_t({'a'=>Integer, 'b'=>Integer})
         struct2 = struct_t({'a'=>Numeric, 'b'=>Numeric})
@@ -1083,6 +1115,13 @@ describe 'The type calculator' do
         factory.constrain_size(hsh, 2, 2)
         calculator.assignable?(struct1, hsh).should == true
         calculator.assignable?(hsh, struct1).should == true
+      end
+
+      it 'should accept empty hash with key_type undef' do
+        struct1 = struct_t({'a'=>optional_t(Integer)})
+        hsh = hash_t(undef_t, undef_t)
+        factory.constrain_size(hsh, 0, 0)
+        calculator.assignable?(struct1, hsh).should == true
       end
     end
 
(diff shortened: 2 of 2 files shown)


lib/puppet/pops/types/type_calculator.rb
```diff
@@ -452,9 +452,18 @@ def instance_of_PTupleType(t, o)
 
   def instance_of_PStructType(t, o)
     return false unless o.is_a?(Hash)
-    h = t.hashed_elements
```
hlindberg: this gives a cached hash of key to value in the struct - why not use it ?

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -452,9 +452,18 @@ def instance_of_PTupleType(t, o)
...
-    # all keys must be present and have a value (even if nil/undef)
-    (o.keys - h.keys).empty? && h.all? { |k,v| instance_of(v, o[k]) }
```
hlindberg: I don't get why this did not work. The first part (o.keys - h.keys) should be empty if o has fewer keys than h. The second part where instance_of(v, o[k]) should end up checking if nil is instance of `Optional[T]` which it should be.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -452,9 +452,18 @@ def instance_of_PTupleType(t, o)
...
+    match_count = 0
+    ok = t.elements.reduce(true) do |memo,e|
+      break false unless memo
+      v = o[e.name]
+      if v.nil?
+        e.type.is_a?(Types::POptionalType)
```
hlindberg: This is not correct - there are other types that accept Optional, such as a Variant that includes an Optional, or Data.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1192,22 +1201,34 @@ def tuple_entry_at(tuple_t, from, to, index)
   # @api private
   #
   def assignable_PStructType(t, t2)
-    return true if t == t2 || t.elements.empty? && (t2.is_a?(Types::PHashType))
-    h = t.hashed_elements
+    els = t.elements
```
hlindberg: 'els' as short for elements ? (not such a great name, looks like misspelled 'else'

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1192,22 +1201,34 @@ def tuple_entry_at(tuple_t, from, to, index)
...
+    return true if t == t2 || els.empty? && (t2.is_a?(Types::PHashType))
     if t2.is_a?(Types::PStructType)
       h2 = t2.hashed_elements
-      h.size == h2.size && h.all? {|k, v| assignable?(v, h2[k]) }
+      match_count = 0
+      ok = els.reduce(true) do |memo, e|
+        break false unless memo
+        t2e = h2[e.name]
+        if t2e.nil?
+          e.type.is_a?(Types::POptionalType)
```
hlindberg: same problem here - not a good enough check.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1192,22 +1201,34 @@ def tuple_entry_at(tuple_t, from, to, index)
   # @api private
   #
   def assignable_PStructType(t, t2)
-    return true if t == t2 || t.elements.empty? && (t2.is_a?(Types::PHashType))
-    h = t.hashed_elements
+    els = t.elements
+    return true if t == t2 || els.empty? && (t2.is_a?(Types::PHashType))
```
hlindberg: This is odd - accepting any Hash if empty is fine, since if the t Struct has no elements it means any struct, and thus any hashtype, but that is also true of any struct type.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1192,22 +1201,34 @@ def tuple_entry_at(tuple_t, from, to, index)
...
     if t2.is_a?(Types::PStructType)
       h2 = t2.hashed_elements
-      h.size == h2.size && h.all? {|k, v| assignable?(v, h2[k]) }
```
hlindberg: this line should be changed to subtract keys, and then check assignability of either the corresponding type in the t2 struct, or PNilType if that key did not have a type.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1192,22 +1201,34 @@ def tuple_entry_at(tuple_t, from, to, index)
...
+          match_count += 1
+          assignable?(e.type, t2e)
+        end
+      end
+      ok && match_count == h2.size
     elsif t2.is_a?(Types::PHashType)
       size_t2 = t2.size_type || @collection_default_size_t
       size_t = Types::PIntegerType.new
-      size_t.from = size_t.to = h.size
+      size_t.from = els.reduce(0) { |c, e| e.type.is_a?(Types::POptionalType) ? c : c + 1 }
```
hlindberg: this is not good enough since other types that Optional may accept Optional

hlindberg:
Sorry, back to the drawing board on this one.

thallgren:
Repoening this since I think the approach taken has some merits that I want to discuss.

lib/puppet/pops/types/type_calculator.rb
thallgren: Because a Struct may well be very short-lived (a Hash is inferred to a Struct) and the cache in that case causes more harm than good. Reverse the question, why use it when it's not needed here?

lib/puppet/pops/types/type_calculator.rb
thallgren: No need to change unless you want to limit the use of copies in the code. The new approach creates no hash, nor does it cache it. Only one side needs a hash to maximize efficiency during matching.

lib/puppet/pops/types/type_calculator.rb
thallgren: I removed the special test since even if t2 is a Hash, it's key_type needs to be assignable to the Struct key_type (e.g. String). The normal test is quick enough anyway.

lib/puppet/pops/types/type_calculator.rb
hlindberg: oh, so this was an optimization - not a fix. I am fine with that, but put it in a separate commit please.

thallgren: To be honest, it didn't start as an optimization. I didn't realize that nil was assignable to Optional, hence the change. I can move it into a separate commit, but doesn't that also imply creating a separate issue and a separate PR? If so, then let's just forget about it. It's not worth all that.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1192,27 +1200,44 @@ def tuple_entry_at(tuple_t, from, to, index)
...
+        break false unless memo
+        t2e = h2[e.name]
+        if t2e.nil?
+          is_type_optional?(e.type)
+        else
+          match_count += 1
+          assignable?(e.type, t2e)
+        end
+      end
+      ok && match_count == h2.size
```
hlindberg: same applies to this optimization

thallgren: Although this is not an optimization. The old code tested if {{nil}} is a {{Type}} that is assignable to an {{Optional}}. I assume that isn't the case since nil is not a type at all.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1192,27 +1200,44 @@ def tuple_entry_at(tuple_t, from, to, index)
...
+        (size_t2.to == 0 || assignable?(@non_empty_string_t, t2.key_type)) &&
+          elements.all? {|e| assignable?(e.type, element_type) }
     else
       false
     end
   end
 
+  # api private
+  #
+  def is_type_optional?(t)
```
hlindberg: Data and Any also accepts nil. Not a fan of special methods for checking type. We have one or so already. I would rather it went in the other direction with fewer ways to check if something has a particular type.

thallgren: Accepting nil is very different from not supplying a value at all. Please note that the is_type_optional? is only called when there is a value (i.e. when _hash.include?_ responds *true*). It is not called for existing keys with nil values.

lib/puppet/pops/types/type_calculator.rb
hlindberg: That was the real problem then? instead of using h2[k] directly, it should have used a PNilType. The rest is the same change as above that does not use all?

thallgren: Again, not providing a value at all is quite different from providing a nil value. Don't you agree with that?

hlindberg:
I want to see the problem fixed clearly without any optimizations. We have tickets to later measure and optimize - lets to such things then.

thallgren:
I'd argue that the "optimization" is a natural part of this PR since it was made possible by the introduction of the _is_optional_type?_ method. It also makes the _instance_of_PStructType_ more similar to the _assignable_PStructType_ since they use the same approach.

hlindberg:
I don't like the introduction of is_optional_type - unnecessary specialized check. It is hard enough to read and reason about the type system methods without optimizations - if we can use straight forward code like: 
```
(o.keys - h.keys).empty? && h.all? { |k,v| instance_of(v, o[k]) }
```
instead of over > 10 lines of code with conditionals and counters and stuff - then I very much prefer
the former.

I like the tests though :-)

lib/puppet/pops/types/type_calculator.rb
hlindberg: yes, please keep the original version - there was nothing wrong with it

lib/puppet/pops/types/type_calculator.rb
thallgren: This test demostrates where just using ```PNilType``` (and also introducing the necessary but slow ```(h2.keys - h.keys).empty?```) would go wrong:

      it 'should reject matching structs with less elements when unmatched elements are data' do
        struct1 = struct_t({'a'=>Integer, 'b'=>Integer, 'c'=>data_t})
        struct2 = struct_t({'a'=>Integer, 'b'=>Integer})
        calculator.assignable?(struct1, struct2).should == false
      end

Would you consider that test invalid?

hlindberg: In general that is different - but not here. If the StructType T says key 'a' is Optional, then it is assignable with a StructType that does not have an 'a' at all, or a StructType that has a type for its 'a' that is assignable to T's 'a' type.

lib/puppet/pops/types/type_calculator.rb
hlindberg: It is specified that an Optional entry in a Struct means that the key may be missing, or it may have a nil (i.e. undef) value.

thallgren:
Code simplified and tests retained :-)

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1192,22 +1192,23 @@ def tuple_entry_at(tuple_t, from, to, index)
...
+      h = t.hashed_elements
       h2 = t2.hashed_elements
-      h.size == h2.size && h.all? {|k, v| assignable?(v, h2[k]) }
+      (h2.keys - h.keys).empty? && h.all? {|k, v| v2 = h2[k]; assignable?(v, v2.nil? ? @nil_t : v2) }
     elsif t2.is_a?(Types::PHashType)
       size_t2 = t2.size_type || @collection_default_size_t
       size_t = Types::PIntegerType.new
-      size_t.from = size_t.to = h.size
+      elements = t.elements
+      size_t.from = elements.reduce(0) { |c, e| assignable?(e.type, @nil_t) ? c : c + 1 }
```
hlindberg: this can be written
```
size_t.from = elements.count {|e| !assignable?(e.type @nil_t) }
```
That is "the count of required keys"

PR #3596, opened 2015-02-11
(PUP-3998) Fix interpolation of "${0[...]}"

hlindberg:
When interpolating a short form numeric variable name (e.g. ${0}) and
continuing with an access operation (to take a substring e.g.
${0[2,4]}) this failed because the validation of a Variable expected the
name of the variable to always be a NAME, not an integer).

This fixes the validation logic, and adds tests.

Also fixed is the label provider which now provides the label "Undef
Value" instead of "Nil Object".


thallgren:
Looks good.

PR #3597, opened 2015-02-12
(PUP-3986) Adds new OID names to Puppet

mrzarquon:
Introduces 13 new possible named OID references that can be used when
provisioning a node's certificate, allowing for easy reference in the
trusted hash in the Puppet dsl.

Pushing to the right origin this time.

### DIFF
--- a/lib/puppet/ssl/oids.rb
+++ b/lib/puppet/ssl/oids.rb
@@ -35,7 +35,20 @@ module Puppet::SSL::Oids
     ["1.3.6.1.4.1.34380.1.1.2", 'pp_instance_id', 'Puppet Node Instance ID'],
     ["1.3.6.1.4.1.34380.1.1.3", 'pp_image_name', 'Puppet Node Image Name'],
     ["1.3.6.1.4.1.34380.1.1.4", 'pp_preshared_key', 'Puppet Node Preshared Key'],
-
+    ["1.3.6.1.4.1.34380.1.1.5", 'pp_cost_center', 'Puppet Node Cost Center Name'],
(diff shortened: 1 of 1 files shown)


mrzarquon:
@kylog I remembered to check where my origin was set on my repo this time.

kylog:
Heh, thanks @mrzarquon!  Btw, since you have the branch on your fork now, I'm going to delete the branch on upstream.

kylog:
@camlow325 or @cprice404 I tagged this as Server team, but let me know if you think that's incorrect.

dankreek:
This looks pretty straightforward to me, but I believe some work on the JVM side may need to be done to mirror this. I will look into this.

dankreek:
Okay, I've submitted a Puppet server ticket for mirroring this work (SERVER-383). @mrzarquon are there any tests we can update with these new names?

mrzarquon:
@dankreek tests around what OIDs can be mapped to appear to live here: https://github.com/puppetlabs/puppet/blob/master/spec/integration/ssl/autosign_spec.rb

dankreek:
Hmm, it looks like that test is intended for testing autosign behavior. If the previous short name => OID mapping isn't tested then I'm thinking we don't need to create a new one. Does this sound good, @kylog ?

kylog:
There are low-level tests of the known OIDs at https://github.com/puppetlabs/puppet/blob/25eb0b279defa6ec5f684a505f91c294f69259f5/spec/unit/ssl/oids_spec.rb#L11-L14

There are also some tests now that decode from a .pem fixture - could be worth adding these OIDs to one of those (or a new one)? I could also be convinced that would be overkill, so just tossing the idea out there.

kylog:
Closing in favor of #3647.

PR #3598, opened 2015-02-12
(PUP-3863) Add 4x API hiera functions

hkenney:
This pull request is the new version of #3551 which includes both the 3x and 4x versions of the hiera functions.


hkenney:
Accidentally opened this against master :T 

Closing and reopening against stable

PR #3599, opened 2015-02-12
(PUP-3863) Add 4x API hiera functions

hkenney:
This pull request is the new version of #3551 which includes both the 3x and 4x versions of the hiera functions.


spec/unit/parser/functions/lookup_spec.rb
```diff
@@ -42,7 +42,7 @@
   end
 
   it "hiera is called to lookup if value is not bound" do
-    Puppet::Parser::Scope.any_instance.stubs(:function_hiera).returns('from_hiera')
+    Hiera.any_instance.stubs(:lookup).returns('from_hiera')
```
hkenney: I wasn't sure whether or not this should be changed back so I left it

thallgren:
All looks good to me. Merging.

PR #3600, opened 2015-02-12
(DOCUMENT-280) Adds note about compatibility

petems:
This is a 3.5+ feature, but someone looking at the main Puppet doc page on and older version might not realise.

### DIFF
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -295,7 +295,9 @@ Puppet::Type.newtype(:file) do
   end
 
   newparam(:validate_cmd) do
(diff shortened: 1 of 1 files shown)


kylog:
IIRC here are a number of examples where params/props/providers/etc have been added in y releases and not sure what the pattern has been for capturing that in docs.

/cc @nfagerlund can you comment?

nfagerlund:
The docs are already versioned: if you're looking at 3.7 or "latest," you should assume it only applies to that version, and should look at older docs for an older version. 

I'm against adding notes like this to the new version of versioned docs — they'll basically pile up forever with no limit, and become less and less relevant.

PR #3601, opened 2015-02-12
(SERVER-360) Add deps for "common package name" test

jeffmccune:
Without this patch the `common_package_name_in_different_providers` acceptance
test is failing in puppet-server with a `createrepo: command not found` error
and exit code of 127.

This patch addresses the problem by ensuring the necessary dependencies of
createrepo and rpm-utils are installed on the system under test.

### DIFF
--- a/acceptance/tests/resource/package/common_package_name_in_different_providers.rb
+++ b/acceptance/tests/resource/package/common_package_name_in_different_providers.rb
@@ -40,6 +40,9 @@ end
 
 # Setup repo and package
 hosts_to_test.each do |agent|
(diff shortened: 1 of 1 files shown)


jeffmccune:
Please don't feel the need to merge this, we're working around the issue in puppet-server by installing dependencies in the pre-suite.  We'll remove that behavior once this gets merged, but no rush.

kylog:
I'm closing this in favor of #3575, which seems like it *should* address the need addressed here but at library level where other tests could benefit.

That said, I could *easily* be missing something with the exact state of testing of puppet-agent vs puppet-server, etc, so if this is still needed, even only temporarily, please just re-open.

PR #3602, opened 2015-02-13
(WIP) Add migration script for FOSS 3.7 -> AIO puppet-agent

branan:
This is a first pass at a migration script for the AIO
installation. It copies certificates and does some basic configuration
cleanup on the agent.

There's a bunch of stuff that's still TODO, and flagged as such in the code. 

It may be possible to break this into pre/postinst steps and make it part of the actual packaging, instead of a standalone script. This version assumes that it owns the entire migration process including the actual package installation.

So far tested on RHEL 7 with a minimal migration:

* install puppet and mcollective
* enable puppet-agent repo
* run the migration script

### DIFF
--- a/ext/migrate-puppet.rb
+++ b/ext/migrate-puppet.rb
@@ -1,0 +1,234 @@
+#!/usr/bin/env ruby
+
+require 'tmpdir'
+require 'optparse'
+
+PUP_FILES = %w(routes.yaml hiera.yaml puppetdb.conf puppet.conf)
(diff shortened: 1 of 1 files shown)


branan:
Big changes that still need to be made:

* Debian support
* pathing updates when that's all finalized

I figured both of those are worth holding off on until we know for sure the way we want to handle migrations.

branan:
We've decided to go with a module for this work, so closing this.

PR #3603, opened 2015-02-13
(docs) Add example error message for boolean CLI options to configuration reference

nfagerlund:
This should make the error more googleable and less baffling.


kylog:
Thanks @nfagerlund!

PR #3606, opened 2015-02-13
(PUP-4011) systemd spec test fix for Ruby 1.8.7

ScottGarman:
In Ruby 1.8.7, the order of hash elements differs from 1.9+ and
caused short-circuiting of the logic used by default.all? in the
provider. Thanks to Peter Huene for diagnosing this issue and
proposing the fix.


peterhuene:
:+1: will merge when Travis is happy.

PR #3474, opened 2015-01-14
(PUP-3560) Properly wrap trusted fact extensions

dankreek:
Reopening this PR in an attempt to deal with travis-ci issues.

PR #3607, opened 2015-02-13
(PUP-4005) Revert "Merge pull request #3474 from dankreek/PUP-3560/wrap-trusted-facts-in-printablestrings

whopper:
This reverts commit 82b7a84bd017c77c95ab02a4e3547e228d7234e1.

Currently, the certificate_extensions acceptance test fails in the AIO
pipeline due to corrupted extensions in the CSR that is generated for
the test. Reverting this commit causes the agent to generate the CSR
correctly.

The certificate_extensions test is also re-enabled in the commit.

Conflicts:
	spec/unit/ssl/certificate_factory_spec.rb
	spec/unit/ssl/certificate_request_spec.rb


kylog:
:+1: 

/cc @joshcooper should we go ahead and merge this?

/cc @dankreek

joshcooper:
@kylog we need to sync up with @dankreek to see if this actually needs to be reverted.

kylog:
I'm going to close this for now until @dankreek gets a chance to catch up and opine. This isn't blocking anything in CI right now with the test disabled so I think we can wait for the real fix. (Of course, if it's hairy/involved we might end up needing to revert after all, but ... fingers crossed.)

PR #3608, opened 2015-02-13
(maint) Fix rspec deprecation warning caused by merge from stable.

peterhuene:
Stable still uses the previous rspec "should" calls that cause rspec
deprecation warnings in master.  Converting to "expect" calls.


joshcooper:
@peterhuene I ran into this earlier and pushed straight to master, not realizing you had a PR up, sorry about that.

peterhuene:
No worries.

PR #3610, opened 2015-02-13
(PUP-1291) Scheduled Tasks - Repeat Triggers

mrdima:
(PUP-1291) Add support for repeat triggers using minutes_interval and minutes_duration.

### DIFF
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -102,6 +102,9 @@ Puppet::Type.newtype(:scheduled_task) do
           * `start_date` ---  The date when the trigger should first become active.
             Defaults to the current date. You should format dates as YYYY-MM-DD,
             although other date formats may work. (Under the hood, this uses `Date.parse`.)
+            `minutes_interval` --- The repeat interval in minutes.
+            `minutes_duration` --- The duration in minutes, needs to be greater or equal to
+             minutes_interval.
       * For `daily` triggers:
           * `every` --- How often the task should run, as a number of days. Defaults
             to 1. ("2" means every other day, "3" means every three days, etc.)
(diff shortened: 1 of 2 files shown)


Iristyle:
Thanks for the contribution @mrdima!  Would you mind squashing down the commit history into a single commit and repushing?

You can do this using interactive rebase with `git rebase -i head~5` and then changing `pick` to `s` (for `squash`) on each of the commits beyond the first.  Git will also combine all the commit messages into one, and you can edit that into the final commit message.

Thanks!

lib/puppet/type/scheduled_task.rb
```diff
@@ -102,6 +102,9 @@ def insync?(current)
           * `start_date` ---  The date when the trigger should first become active.
             Defaults to the current date. You should format dates as YYYY-MM-DD,
             although other date formats may work. (Under the hood, this uses `Date.parse`.)
+			`minutes_interval` --- The repeat interval in minutes.
```
Iristyle: Looks like we have a mix of tabs and spaces here.  We use spaces for indentation in the Puppet codebase.

lib/puppet/type/scheduled_task.rb
```diff
@@ -140,7 +143,7 @@ def insync?(current)
           }
 
           # Run at 8am on the first Monday of the month for January, March, and May,
-          # starting after August 31st, 2011.
+          # starting after August 31st, 2011. Repeat every 10 minutes for 60 minutes.
```
Iristyle: I think it might be better to add an additional example that focuses on `minutes_interval` and `minutes_duration`, rather than changing the existing example.

Iristyle:
I think that we're also going to want some specs added to https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb to cover the new `minutes_interval` and `minutes_duration` configuration values.

mrdima:
Yes, you're totally right, I was already reading into it and saw that 9 spec tests are failing (because of the new "unexpected" properties being returned and initialized to nil there), need to fix that, add new specs and I'll take care of the other things you mentioned as well, but bare with me as I'm pretty busy this week.

mrdima:
Sorry, but I couldn't get things to re-push with everything squashed into 1 commit. I will hopefully have some time tomorrow to create a new one.

kylog:
@mrdima for future reference, you would need to force push after squashing.

mrdima:
@kylog. Thanks, I tried that as well (but maybe too late, but also after rebasing several times), but kept getting everything is uptodate.

mrdima:
Please ignore this specific pull request, it's 3637 that does the job

kylog:
@mrdima got it. This PR is closed so we'll continue review on #3637. Thanks!

PR #3611, opened 2015-02-14
update all linices(?) for temp aio testing

justinstoller:
This allows us to get feedback for what is breaking on each supported linux OS for the all in one agent.


joshcooper:
This PR isn't using the final AIO paths, so probably best to close this out, and submit a new PR for the second wave of platforms. We should also chat with @kylog about what platforms that should be.

PR #3613, opened 2015-02-14
(PUP-4008) Consider empty hash assignable to any Hash with size 0

thallgren:
This commit adds the method 'is_the_empty_hash?' to the PHashType.
It responds true when both the key and value types are PNilType and
the size_type is 0,0. This method is then called by the
TypeCalculator.assignable_PHashType when the first parameter accepts
a hash type of size zero.

### DIFF
--- a/lib/puppet/pops/types/type_calculator.rb
+++ b/lib/puppet/pops/types/type_calculator.rb
@@ -1429,7 +1429,8 @@ class Puppet::Pops::Types::TypeCalculator
   # @api private
   def assignable_PHashType(t, t2)
     case t2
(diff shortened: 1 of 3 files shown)


lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1429,7 +1429,8 @@ def assignable_PArrayType(t, t2)
   # @api private
   def assignable_PHashType(t, t2)
     case t2
-    when Types::PHashType
+      when Types::PHashType
```
hlindberg: indent is wrong, when should be aligned with case

PR #3615, opened 2015-02-17
Allow command symlinks

asharpe:
Some commands may be symlinked into the "standard" path (eg. using the alternatives system)

### DIFF
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -145,7 +145,7 @@ module Util
   # @return [String] the absolute path to the found executable.
   def which(bin)
     if absolute_path?(bin)
-      return bin if FileTest.file? bin and FileTest.executable? bin
+      return bin if (FileTest.file? bin or FileTest.symlink? bin) and FileTest.executable? bin
     else
       ENV['PATH'].split(File::PATH_SEPARATOR).each do |dir|
         begin
@@ -173,7 +173,7 @@ module Util
               return destext if FileTest.file? destext and FileTest.executable? destext
             end
           end
-          return dest if FileTest.file? dest and FileTest.executable? dest
+          return dest if (FileTest.file? dest or FileTest.symlink? bin) and FileTest.executable? dest
         end
       end
     end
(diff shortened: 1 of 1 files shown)


asharpe:
This is a trivial patch

ferventcoder:
Please squash your commits and update your commit message with the prefix `(maint)`. We may also need a ticket filed for this. If so, we'll ask that you prefix your commit message with that instead.

kylog:
@asharpe thanks for the contribution! This is actually a behavioral change IMHO so this should be ticketed in jira and not a 'maint' commit. Thoughts? /cc @ferventcoder

ferventcoder:
@kylog agreed. Many things look trivial when in fact they are not based on the impact that they carry. :)

ferventcoder:
I would have stated so earlier on the ticket but I was in here looking to see if there was an impact on Windows or not.

lib/puppet/util.rb
```diff
@@ -145,7 +145,7 @@ def benchmark(*args)
   # @return [String] the absolute path to the found executable.
   def which(bin)
     if absolute_path?(bin)
-      return bin if FileTest.file? bin and FileTest.executable? bin
+      return bin if (FileTest.file? bin or FileTest.symlink? bin) and FileTest.executable? bin
```
stschulte: this would also match for symlinks to directories (which have the executable bit set in most cases). AFAIK symlinks are followed anyway, so `FileTest.file? "symlink_to_file"` does return `true` and `FileTest.file? "symlink to directory"` does return `false`. So do you have a real use case that breaks without the change?

kylog:
@asharpe have you had a chance to review the comments above?

kylog:
@asharpe ping.

asharpe:
@kylog apologies.

I don't agree with all comments (eg. that "This is actually a behavioral change") however the issue can be resolved by altering the PATH env var before running puppet.

I couldn't find a simple test to replicate FileTest.file? failing however I had /usr/bin/php symlinked via the alternatives system, and the patch provided resolved the issue.

PR #3616, opened 2015-02-17

ferventcoder:

~~Add `Puppet::FileSystem.read_preserve_line_endings` for cases where line
endings are important and use `File.readlines` method, which is known to
preserve original line endings (from reading the source, read has defined
calls for Windows where readlines does not).~~

~~References:~~

 * ~~IO.Read - https://github.com/ruby/ruby/blob/v2_1_5/io.c#L9638~~
 * ~~IO.Readlines - https://github.com/ruby/ruby/blob/v2_1_5/io.c#L3324~~
 * ~~IO.Read Windows - https://github.com/ruby/ruby/blob/v2_1_5/io.c#L2819-L2828~~
 * ~~https://www.ruby-forum.com/topic/132198~~

Supersedes https://github.com/puppetlabs/puppet/pull/2397

### DIFF
--- a/lib/puppet/file_system/file_impl.rb
+++ b/lib/puppet/file_system/file_impl.rb
@@ -75,6 +75,14 @@ class Puppet::FileSystem::FileImpl
     path.read
   end
 
+  def read_preserve_line_endings(path)
+    contents = path.read( :mode => 'rb', :encoding => Encoding::UTF_8)
+    contents = path.read( :mode => 'rb', :encoding => Encoding::default_external) unless contents.valid_encoding?
+    contents = path.read unless contents.valid_encoding?
+
+    contents
+  end
+
   def binread(path)
     raise NotImplementedError
   end
--- a/spec/unit/file_system_spec.rb
+++ b/spec/unit/file_system_spec.rb
@@ -5,6 +5,14 @@ require 'puppet/util/platform'
 describe "Puppet::FileSystem" do
   include PuppetSpec::Files
 
+  def with_file_content(content)
+    path = tmpfile('file-system')
+    file = File.new(path, 'wb')
+    file.sync = true
+    file.print content
+    yield path
+  end
+
   context "#exclusive_open" do
     it "opens ands allows updating of an existing file" do
       file = file_containing("file_to_update", "the contents")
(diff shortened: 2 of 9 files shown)


ferventcoder:
Strangely enough in Ruby 1.9.3 and Ruby 2.1.5 I couldn't get File.read to return incorrect line endings. So this may not quite be the solution.

ferventcoder:
So I have a repro proving this is not sufficient. Will look at switching to binary.

spec/unit/file_system_spec.rb
```diff
@@ -5,6 +5,14 @@
 describe "Puppet::FileSystem" do
   include PuppetSpec::Files
 
+  def with_file_content(content)
+    path = tmpfile('file-system')
+    file = File.new(path, 'w')
```
joshcooper: I think the mode needs to be `'wb'`, otherwise ruby will strip '\r' while it's writing the test file that it tries to read back in later.

lib/puppet/file_system/file_impl.rb
```diff
@@ -75,6 +75,15 @@ def read(path)
     path.read
   end
 
+  def read_preserve_line_endings(path)
+    contents = path.read( :mode => 'rb', :encoding => Encoding::UTF_8)
+    contents = path.read( :mode => 'rb', :encoding => Encoding::default_external) unless contents.valid_encoding?
+    contents = path.read unless contents.valid_encoding?
+
+    contents
+    #::File.readlines(path).join('')
```
joshcooper: remove

spec/unit/file_system_spec.rb
ferventcoder: Done.

lib/puppet/file_system/file_impl.rb
ferventcoder: Done

ferventcoder:
The question I have is should we just switch Puppet::FileSystem.read to use this method and not have a separate method? I wasn't trying to be too intrusive and wanted to catch the specific cases where this was occurring. Or maybe that should just wait until [PUP-2564](https://tickets.puppetlabs.com/browse/PUP-2564?focusedCommentId=128361&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-128361)?

ferventcoder:
This is ready to go again.

Iristyle:
Test repo from @ferventcoder at https://github.com/ferventcoder/puppet-simple_templates (note it's missing `.gitattributes` so you're subject to your global Git settings)

ferventcoder:
@Iristyle I take pull requests! :+1:

Iristyle:
@ferventcoder let's retarget this at the new `3.x` branch.  Thanks!

ferventcoder:
Closing, will be retargeting soon.

PR #3617, opened 2015-02-17
(PUP-3982) Enable condrestart for puppet agent init script

branan:
Previously, condrestart was supported only on redhat. This brings the
support to Debian. Condrestart is useful for packaging scripts that
want to bounce the service if it's running, but not start it if a user
chose to keep it off.

### DIFF
--- a/ext/debian/puppet.init
+++ b/ext/debian/puppet.init
@@ -49,6 +49,13 @@ stop_puppet_agent() {
     rm -f "$PIDFILE"
 }
 
+restart_puppet_agent() {
+    log_begin_msg "Restarting $DESC"
+    stop_puppet_agent
+    start_puppet_agent
+    log_end_msg $?
+}
+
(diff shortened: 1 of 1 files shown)


kylog:
Ping @haus

ext/debian/puppet.init
```diff
@@ -107,10 +114,12 @@ case "$1" in
...
-        log_end_msg $?
+        restart_puppet_agent
+    ;;
+    condrestart)
+        if status_puppet_agent >/dev/null 2>&1; then
+            restart_puppet_agent
+        fi
     ;;
     *)
         echo "Usage: $0 {start|stop|status|restart|force-reload|reload}" >&2
```
haus: condrestart should be added here as well.

branan:
@haus updated

haus:
:+1:

PR #3612, opened 2015-02-14
(DO NOT MERGE)(QENG-1800) Install MSI during package acceptance

Iristyle:
Previously it was not possible to run Windows acceptance against FOSS
MSI packages. This commit adds a step to both the package and aio
rake tasks that can install the puppet-agent MSI.
It depends on https://github.com/puppetlabs/beaker/pull/674

With this commit you can do:

    $ bundle exec rake ci:test:packages \
        CONFIG=config/nodes/win2008r2-rubyx64.yaml \
        SHA=<SHA> VERSION=3.7.4-1303

    $ bundle exec rake ci:test:aio \
        CONFIG=config/nodes/win2012r2-rubyx64.yaml \
        SHA=<SHA> VERSION=1.0.0-o-ga354e25 \
        MSI_BASE_URL=http://int-resources.corp.puppetlabs.net/aio

Note the VERSION string must match the filename of the MSI which
includes the output from `git describe`. The host definition file needs
to specify which `ruby_arch` it wants to install, either x86 or x64.

Given one off builds may wish to be tested that are not products
of the build pipeline, the MSI url and filename may be overridden
completely with two environment variables:

* MSI_BASE_URL may optionally override the URL that Puppet can be
retrieved from, which normally defaults to:

`http://builds.puppetlabs.lan/puppet(-agent)/<SHA>/artifacts/windows`

* MSI_FILENAME may optionally override the filename that Puppet pulls
the MSI from, which normally defaults to:

`puppet-agent-<VERSION>-<arch>.msi`

PR #3618, opened 2015-02-17
(QENG-1800) Install MSI for AIO package acceptance

Iristyle:
Previously it was not possible to run Windows acceptance against FOSS
MSI packages. This commit adds a step to the aio rake tasks that can
install the puppet-agent MSI.
It depends on https://github.com/puppetlabs/beaker/pull/674

With this commit you can do:

    $ bundle exec rake ci:test:aio \
        CONFIG=config/nodes/win2012r2-rubyx64.yaml \
        SHA=<SHA> VERSION=1.0.0-o-ga354e25 \
        MSI_BASE_URL=http://int-resources.corp.puppetlabs.net/aio

Given one off builds may wish to be tested that are not products
of the build pipeline, the MSI url and filename may be overridden
completely with two environment variables:

* MSI_BASE_URL may optionally override the URL that Puppet can be
retrieved from, which normally defaults to:

http://builds.puppetlabs.lan/puppet(-agent)/<SHA>/artifacts/windows

* MSI_FILENAME may optionally override the filename that Puppet pulls
the MSI from, which normally defaults to:

puppet-agent-<VERSION>-<arch>.msi

### DIFF
--- a/acceptance/config/nodes/win2012r2-rubyx64.yaml
+++ b/acceptance/config/nodes/win2012r2-rubyx64.yaml
@@ -5,6 +5,14 @@ HOSTS:
     platform: el-7-x86_64
     hypervisor: vcloud
     template: redhat-7-x86_64
+    puppetcodedir: /etc/puppetlabs/code
+    distmoduledir: /etc/puppetlabs/code/modules
+    hieraconf:     /etc/puppetlabs/code/hiera.yaml
+    puppetconfdir: /etc/puppetlabs/puppet
+    puppetserver-confdir: /etc/puppetserver/conf.d/
+    puppetbindir:  /opt/puppetlabs/puppet/bin
+    puppetvardir:  /opt/puppetlabs/puppet/cache
+    sitemoduledir: /opt/puppetlabs/puppet/modules
   agent-2012r2-x86_64-rubyx64:
     roles:
       - agent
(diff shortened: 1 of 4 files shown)


Iristyle:
Updated per comments in #3612

Iristyle:
Supercedes #3612 (which was targeted at `aio` branch rather than `aio2` branch)

acceptance/config/nodes/win2012r2-rubyx64.yaml
```diff
@@ -12,6 +20,12 @@ HOSTS:
     ruby_arch: x64
     hypervisor: vcloud
     template: win-2012r2-x86_64
+    puppetconfdir: C:/ProgramData/PuppetLabs/puppet/etc
+    puppetcodedir: C:/ProgramData/PuppetLabs/puppet/code
+    puppetvardir: C:/ProgramData/PuppetLabs/puppet/var
+    distmoduledir: C:/ProgramData/PuppetLabs/puppet/code/modules
+    sitemoduledir: C:/opt/puppetlabs/puppet/modules
+    hieraconf: C:/ProgramData/PuppetLabs/puppet/etc/hiera.yaml
```
joshcooper: I don't think this file is actually used on the windows agents (perhaps during `puppet apply`), but for consistency this should be `C:/ProgramData/PuppetLabs/puppet/code/hiera.yaml`

Iristyle: Sorry... shouldn't have missed that one - good catch.

PR #3619, opened 2015-02-17
(docs) Revise description of file type's mode attribute

nfagerlund:
* Modes MUST be strings, not should
* Add more info about setuid/setgid/sticky
* Break up description of numeric modes to make it more scannable
* Remove reference to "four-digit," since that's a style issue and not technically required
* Docs standards: uppercase text in `<PLACEHOLDER>` tokens

### DIFF
--- a/lib/puppet/type/file/mode.rb
+++ b/lib/puppet/type/file/mode.rb
@@ -10,7 +10,7 @@ module Puppet
 
     desc <<-'EOT'
       The desired permissions mode for the file, in symbolic or numeric
-      notation. This value should be specified as a quoted string; do not use
+      notation. This value **must** be specified as a string; do not use
       un-quoted numbers to represent file modes.
 
       The `file` type uses traditional Unix permission schemes and translates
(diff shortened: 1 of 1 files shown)


lib/puppet/type/file/mode.rb
```diff
@@ -19,16 +19,22 @@ module Puppet
...
 
-      Numeric modes should use the standard four-digit octal notation of
-      `<setuid/setgid/sticky><owner><group><other>` (e.g. 0644). Each of the
-      "owner," "group," and "other" digits should be a sum of the
-      permissions for that class of users, where read = 4, write = 2, and
-      execute/search = 1. When setting numeric permissions for
-      directories, Puppet sets the search permission wherever the read
-      permission is set.
+      Numeric modes should use the standard octal notation of
+      `<SETUID/SETGID/STICKY><OWNER><GROUP><OTHER>` (e.g. '0644').
```
kylog: Should we drop the 'standard' here since that was a source of confusion?

nfagerlund: Nah. There was disagreement over whether using specifically _four_ digits should be considered "standard" (answer: talk to your local style guide), but the order in which those digits go is really the standard numerical way to express permissions.

PR #3620, opened 2015-02-18
(DO NOT MERGE)(QENG-1800) Add <ARCH> to location of Puppet MSI

Iristyle:
- As it turns out, the build process produces MSIs at a path that
   includes /puppet-agent/<SHA>/artifacts/windows/<ARCH> and not
   /puppet-agent/<SHA>/artifacts/windows/


Iristyle:
I think we should close this and modify how the MSI is placed on builds.puppetlabs.lan, so that `<ARCH>` is not in the path.

Iristyle:
Commit 322464b of https://github.com/puppetlabs/puppet-agent/pull/55 is what I believe is currently responsible for the unneeded nesting of `<ARCH>` and the preference would be to modify puppet-agent instead of merging this PR.

Iristyle:
Not going with this approach... closing

PR #3621, opened 2015-02-18
(DO NOT MERGE)(QENG-1800) Use SUITE_VERSION to identify packages

Iristyle:
- On other platforms, repo configs are used to find actual package
   names that include a .x.y.z.XX-<SHA> type version.  The previous
   assumption was that there would be an ENV['VERSION'] passed through
   the pipeline, but that assumption proved incorrect as it's not
   necessary for other platforms.  In the existing pipelines, <SHA> is
   the only thing necessary to identify a package.

   So a new parameter called SUITE_VERSION was added to the environment
   to address this.


Iristyle:
Still under active discussion if this approach is the best.

joshcooper:
:+1: to this approach and PR

joshcooper:
This was what @justinstoller and I had discussed earlier: https://tickets.puppetlabs.com/browse/QENG-1805

Iristyle:
Going with a third approach using symlinks in #3626 so closing this PR.

PR #3622, opened 2015-02-18
(DO NOT MERGE) (QENG-1800) VERSION is not available ENV var

Iristyle:
- Previously, the Windows package installation assumed that a
   VERSION environment variable is available during the Jenkins pipeline
   However, other platforms use only a SHA + repo configs to identify
   the actual package names.

   Therefore, the Windows packaging jobs are using a rough equivalent to
   repo configs by creating a LATEST-x64 and LATEST-x86 in
   builds.puppetlabs.lan/puppet-agent/<SHA>/artifacts/windows

   With this approach, the file is read for the actual version string
   in the x.y.z.XX.g<SHA> format and injected into the MSI package name

### DIFF
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -54,7 +54,14 @@ agents.each do |agent|
   if agent['platform'] =~ /windows/
     arch = agent[:ruby_arch] || 'x86'
     base_url = ENV['MSI_BASE_URL'] || "http://builds.puppetlabs.lan/puppet-agent/#{ENV['SHA']}/artifacts/windows"
-    filename = ENV['MSI_FILENAME'] || "puppet-agent-#{ENV['VERSION']}-#{arch}.msi"
+
+    filename = ENV['MSI_FILENAME']
+    if !filename
+      Dir.mktmpdir do |tmp_dir|
+        latest = fetch(base_url, "LATEST-#{ARCH}", tmp_dir)
+        filename = File.read(latest).strip
+      end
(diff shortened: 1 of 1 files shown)


Iristyle:
This approach is under active discussion.  Either #3621 or this PR should be merged, but *NOT BOTH*.

Iristyle:
This is dependent on https://github.com/Iristyle/puppet-agent/commit/e1485ae689530c14f1eefbca642c06d805c1ee44 from https://github.com/puppetlabs/puppet-agent/pull/55

acceptance/setup/aio/pre-suite/010_Install.rb
```diff
@@ -54,7 +54,14 @@
   if agent['platform'] =~ /windows/
     arch = agent[:ruby_arch] || 'x86'
     base_url = ENV['MSI_BASE_URL'] || "http://builds.puppetlabs.lan/puppet-agent/#{ENV['SHA']}/artifacts/windows"
-    filename = ENV['MSI_FILENAME'] || "puppet-agent-#{ENV['VERSION']}-#{arch}.msi"
+
+    filename = ENV['MSI_FILENAME']
+    if !filename
+      Dir.mktmpdir do |tmp_dir|
+        latest = fetch(base_url, "LATEST-#{ARCH}", tmp_dir)
```
Iristyle: AFAIK, we don't have a `fetch_as_string` sort of method, which would be preferable to writing a temp file here.

Iristyle:
Going with a third approach using symlinks in #3626 so closing this PR.

PR #3623, opened 2015-02-18
(PUP-2929) Optimize the TypedName for use as hash key

thallgren:
This commit will optimize the TypedName in two ways.

1. The initialize method now executes more than twice as fast. The
optimization consists of three improvements:
 - Use a string instead of a regexp when splitting the name
 - Never join the parts together. Instead use name or substring of name
 - Use direct match with NUMERIC. Avoids redundant relativization

2. A compound frozen name is created when the instance is initialized.
This instance is then used when comparing and when doing to_s.

The cached @qualified attribute and associated attr_reader was
replaced with an 'qualified?' method because there's almost no
noticable gain by caching it (it's extremely fast with or without a
cache). Also, 'qualified?' (with a questionmark) is a better name.

The commit does not contain the actual performance tests since a) I
wasn't sure where to put them and b) Not sure how useful they will be
once this commit has been merged. The number are:

Benchmark using old TypeName
```
Create              6.850000   0.000000   6.850000 (  6.859569)
Create ::global     6.800000   0.000000   6.800000 (  6.802436)
Hash lookup         0.590000   0.000000   0.590000 (  0.586534)
qualified           0.050000   0.000000   0.050000 (  0.052163)
to_s                0.380000   0.000000   0.380000 (  0.386374)
```
Benchmark using new TypeName
```
Create              2.400000   0.000000   2.400000 (  2.393237)
Create ::global     2.940000   0.000000   2.940000 (  2.947483)
Hash lookup         0.430000   0.000000   0.430000 (  0.427165)
qualified?          0.070000   0.000000   0.070000 (  0.067695)
to_s                0.060000   0.000000   0.060000 (  0.062700)
```

### DIFF
--- a/lib/puppet/pops/loader/dependency_loader.rb
+++ b/lib/puppet/pops/loader/dependency_loader.rb
@@ -26,7 +26,7 @@ class Puppet::Pops::Loader::DependencyLoader < Puppet::Pops::Loader::BaseLoader
   # Finds name in a loader this loader depends on / can see
   #
   def find(typed_name)
-    if typed_name.qualified
+    if typed_name.qualified?
       if l = index()[typed_name.name_parts[0]]
         l.load_typed(typed_name)
       else
--- a/lib/puppet/pops/loader/loader.rb
+++ b/lib/puppet/pops/loader/loader.rb
@@ -137,43 +137,49 @@ class Puppet::Pops::Loader::Loader
   # A name/type combination that can be used as a compound hash key
   #
   class TypedName
+    attr_reader :hash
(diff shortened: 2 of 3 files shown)


hlindberg:
Looks fine. We want to wait with a merge of this until there is a benchmark in the puppet repo that measures function loading and calls. (It is a lot more difficult to compare results if the optimization is merged first). Also want the benchmark on stable to be able to measure 3x with future parser.

hlindberg:
Closing in favor of #3644

PR #3624, opened 2015-02-18
(PUP-4027) Fix package docs for PUP-1073

MikaelSmith:
Updates the package type docs to account for changes in PUP-1073
allowing multiple package resources that share the same name if they
come from different providers.

### DIFF
--- a/lib/puppet/type/package.rb
+++ b/lib/puppet/type/package.rb
@@ -20,6 +20,13 @@ module Puppet
       requires in order to function, and you must meet those requirements
       to use a given provider.
 
+      You can declare multiple package resources with the same `name`, as long
+      as they specify different providers and have unique titles.
+
+      Note that you must use the _title_ to make a reference to a package
+      resource; `Package[<NAME>]` is not a synonym for `Package[<TITLE>]` like
+      it is for many other resource types.
+
       **Autorequires:** If Puppet is managing the files specified as a
       package's `adminfile`, `responsefile`, or `source`, the package
       resource will autorequire those files."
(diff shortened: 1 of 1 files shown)


MikaelSmith:
ping @nfagerlund

lib/puppet/type/package.rb
```diff
@@ -213,11 +218,9 @@ def should_to_s(newvalue = @should)
             default => openssl
           }
 
-          # It is not an error to set an alias to the same value as the
-          # object name.
-          package { $ssl:
-            ensure => installed,
-            alias  => openssl
```
MikaelSmith: Do we care about keeping examples of using alias?

lib/puppet/type/package.rb
```diff
@@ -20,6 +20,11 @@ module Puppet
       requires in order to function, and you must meet those requirements
       to use a given provider.
 
+      Multiple resources for packages from different providers that share
+      the same name can be specified, but they must have unique titles.
+      Resources referenced by Package[name] are no longer recognized as
+      they would be ambiguous, and must now be referenced by Package[title].
```
nfagerlund: We can tighten this up a bit by removing passive voice and focusing on the action and its consequences:

    You can declare multiple package resources with the same `name`, as long as they
    specify different providers and have unique titles.

    Note that you must use the _title_ to make a reference to a package resource;
    `Package[<NAME>]` is not a synonym for `Package[<TITLE>]` like it is for many
    other resource types.

lib/puppet/type/package.rb
nfagerlund: Definitely not!

nfagerlund:
Other than that, I'm good with this.

MikaelSmith:
Reworded the bit about aliases.

PR #3633, opened 2015-02-20
(PUP-4032) Fix exitcodes on puppet device

hunner:
Puppet device does one catalog run per device in device.conf, however it
was then throwing away the return codes of the catalog compile and
apply. The man page says that puppet device supports
--detailed-exitcodes, so this fixes it.

Since there are multiple catalog runs in one command, exitcode masking
is unavoidable, so I selected to take the path of returning the single
most important exit code, in decending order with the exception of exit
1 taking precedence.

### DIFF
--- a/lib/puppet/application/device.rb
+++ b/lib/puppet/application/device.rb
@@ -115,9 +115,10 @@ parameter, so you can specify '--server <servername>' as an argument.
 
 * --detailed-exitcodes:
   Provide transaction information via exit codes. If this is enabled, an exit
-  code of '2' means there were changes, an exit code of '4' means there were
-  failures during the transaction, and an exit code of '6' means there were both
-  changes and failures.
+  code of '1' means at least one device had a compile failure, an exit code of
+  '2' means at least one device had resource changes, and an exit code of '4'
+  means at least one device had resource failures. Exit codes of '3', '5', '6',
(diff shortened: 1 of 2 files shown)


lib/puppet/application/device.rb
```diff
@@ -199,6 +200,15 @@ def main
         Puppet::SSL::Host.reset
       end
     end
+    if ! returns or ! returns.max
```
hunner: Doesn't work on 1.8.7 when returns is an array of `nil`

kylog:
@hunner can you file a jira ticket? And then reference it in the commit message, e.g. "(PUP-1234) Fix etc etc".

Also, I see you targeted this at stable. Do you have a need for this in the 3.x series? If not, you can target it at master (and then you don't need to worry about fixing the 1.8.7 thing :))

hunner:
@kylog We depend on detailed exitcodes for testing. I can work around it by grepping output for the time being, but it would be nice to have in stable if it's no trouble.

Otherwise, I can rebase to master if you'd rather retire stable.

delllinux:
dPlease do what is best
thank you
m.

melissa:
@hunner any updates on this?

hunner:
Updated PR to do bitwise exit code munging and document the new exit codes.

branan:
I'm happy with this with the orred exit codes.

PR #3637, opened 2015-02-20
(PUP-1291) Scheduled Tasks Support for Repeat Triggers

mrdima:
Add minutes_interval and minutes_duration for supporting repeat triggers.
Also added support for the comment field as a bonus.

### DIFF
--- a/lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
+++ b/lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
@@ -47,6 +47,10 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
     task.application_name
   end
 
+  def comment
+    task.comment
+  end
+  
   def arguments
     task.parameters
   end
@@ -76,7 +80,6 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
                   nil
                 end
       next unless trigger and scheduler_trigger_types.include?(trigger['trigger_type'])
-
       puppet_trigger = {}
       case trigger['trigger_type']
       when Win32::TaskScheduler::TASK_TIME_TRIGGER_DAILY
@@ -101,6 +104,8 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
       puppet_trigger['start_date'] = self.class.normalized_date("#{trigger['start_year']}-#{trigger['start_month']}-#{trigger['start_day']}")
       puppet_trigger['start_time'] = self.class.normalized_time("#{trigger['start_hour']}:#{trigger['start_minute']}")
       puppet_trigger['enabled']    = trigger['flags'] & Win32::TaskScheduler::TASK_TRIGGER_FLAG_DISABLED == 0
+      puppet_trigger['minutes_interval'] = trigger['minutes_interval'] ||= 0
+      puppet_trigger['minutes_duration'] = trigger['minutes_duration'] ||= 0
       puppet_trigger['index']      = i
 
       @triggers << puppet_trigger
@@ -139,6 +144,10 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
 
   def command=(value)
     task.application_name = value
+  end
+
+  def comment=(value)
+    task.comment = value
   end
 
   def arguments=(value)
@@ -209,10 +218,9 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
   def create
     clear_task
     @task = Win32::TaskScheduler.new(resource[:name], dummy_time_trigger)
-
     self.command = resource[:command]
 
-    [:arguments, :working_dir, :enabled, :trigger, :user].each do |prop|
+    [:arguments, :working_dir, :enabled, :trigger, :user, :comment].each do |prop|
       send("#{prop}=", resource[prop]) if resource[prop]
     end
   end
@@ -239,6 +247,8 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
     desired['months']      ||= current_trigger['months']      if current_trigger.has_key?('months')
     desired['on']          ||= current_trigger['on']          if current_trigger.has_key?('on')
     desired['day_of_week'] ||= current_trigger['day_of_week'] if current_trigger.has_key?('day_of_week')
+    desired['minutes_interval'] ||= current_trigger['minutes_interval'] if current_trigger.has_key?('minutes_interval')
+    desired['minutes_duration'] ||= current_trigger['minutes_duration'] if current_trigger.has_key?('minutes_duration')
 
     translate_hash_to_trigger(current_trigger) == translate_hash_to_trigger(desired)
   end
@@ -258,10 +268,10 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
       'flags'                   => 0,
       'random_minutes_interval' => 0,
       'end_day'                 => 0,
-      "end_year"                => 0,
-      "minutes_interval"        => 0,
-      "end_month"               => 0,
-      "minutes_duration"        => 0,
+      'end_year'                => 0,
+      'minutes_interval'        => 0,
+      'end_month'               => 0,
+      'minutes_duration'        => 0,
       'start_year'              => now.year,
       'start_month'             => now.month,
       'start_day'               => now.day,
@@ -280,7 +290,7 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
       trigger['flags'] &= ~Win32::TaskScheduler::TASK_TRIGGER_FLAG_DISABLED
     end
 
-    extra_keys = puppet_trigger.keys.sort - ['index', 'enabled', 'schedule', 'start_date', 'start_time', 'every', 'months', 'on', 'which_occurrence', 'day_of_week']
+    extra_keys = puppet_trigger.keys.sort - ['index', 'enabled', 'schedule', 'start_date', 'start_time', 'every', 'months', 'on', 'which_occurrence', 'day_of_week', 'minutes_interval', 'minutes_duration']
     self.fail "Unknown trigger option(s): #{Puppet::Parameter.format_value_for_display(extra_keys)}" unless extra_keys.empty?
     self.fail "Must specify 'start_time' when defining a trigger" unless puppet_trigger['start_time']
 
@@ -331,6 +341,17 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
       trigger['trigger_type'] = Win32::TaskScheduler::ONCE
     else
       self.fail "Unknown schedule type: #{puppet_trigger["schedule"].inspect}"
+    end
+
+    if puppet_trigger['minutes_interval']
+      integer_interval = Integer(puppet_trigger['minutes_interval'])
+      self.fail 'minutes_interval must be an integer greater or equal to 0' unless integer_interval >= 0
+      trigger['minutes_interval'] = integer_interval
+    end
+    if puppet_trigger['minutes_duration']
+      integer_duration = Integer(puppet_trigger['minutes_duration'])
+      self.fail 'minutes_duration must be an integer greater than minutes_interval' unless integer_duration > integer_interval || (integer_duration == 0 && integer_interval == 0)
+      trigger['minutes_duration'] = integer_duration
     end
 
     if start_date = puppet_trigger['start_date']
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -96,8 +96,10 @@ Puppet::Type.newtype(:scheduled_task) do
       A trigger can contain the following keys:
 
       * For all triggers:
-          * `schedule` **(Required)** --- The schedule type. Valid values are
-            `daily`, `weekly`, `monthly`, or `once`.
+          * `schedule` **(Required)** --- What kind of trigger this is.
+            Valid values are `daily`, `weekly`, `monthly`, or `once`. Each kind
+            of trigger is configured with a different set of keys; see the
+            sections below. (`once` triggers only need a start time/date.)
           * `start_time` **(Required)** --- The time of day when the trigger should
             first become active. Several time formats will work, but we
             suggest 24-hour time formatted as HH:MM.
@@ -105,24 +107,24 @@ Puppet::Type.newtype(:scheduled_task) do
             Defaults to the current date. You should format dates as YYYY-MM-DD,
             although other date formats may work. (Under the hood, this uses `Date.parse`.)
           * `minutes_interval` --- The repeat interval in minutes.
-          * `minutes_duration` --- The duration in minutes, needs to be greater or equal to
+          * `minutes_duration` --- The duration in minutes, needs to be greater than the
             minutes_interval.
-      * For daily triggers:
+      * For `daily` triggers:
           * `every` --- How often the task should run, as a number of days. Defaults
             to 1. ("2" means every other day, "3" means every three days, etc.)
-      * For weekly triggers:
+      * For `weekly` triggers:
           * `every` --- How often the task should run, as a number of weeks. Defaults
             to 1. ("2" means every other week, "3" means every three weeks, etc.)
           * `day_of_week` --- Which days of the week the task should run, as an array.
             Defaults to all days. Each day must be one of `mon`, `tues`,
             `wed`, `thurs`, `fri`, `sat`, `sun`, or `all`.
-      * For monthly-by-date triggers:
+      * For `monthly` (by date) triggers:
           * `months` --- Which months the task should run, as an array. Defaults to
             all months. Each month must be an integer between 1 and 12.
           * `on` **(Required)** --- Which days of the month the task should run,
             as an array. Each day must beeither an integer between 1 and 31,
             or the special value `last,` which is always the last day of the month.
-      * For monthly-by-weekday triggers:
+      * For `monthly` (by weekday) triggers:
           * `months` --- Which months the task should run, as an array. Defaults to
             all months. Each month must be an integer between 1 and 12.
           * `day_of_week` **(Required)** --- Which day of the week the task should
@@ -131,6 +133,7 @@ Puppet::Type.newtype(:scheduled_task) do
           * `which_occurrence` **(Required)** --- The occurrence of the chosen weekday
             when the task should run. Must be one of `first`, `second`, `third`,
             `fourth`, `fifth`, or `last`.
+
 
       Examples:
 
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -27,6 +27,10 @@ Puppet::Type.newtype(:scheduled_task) do
     isnamevar
   end
 
+  newproperty(:comment) do
+    desc "The comment for the scheduled task."
+  end
+  
   newproperty(:command) do
     desc "The full path to the application to run, without any arguments."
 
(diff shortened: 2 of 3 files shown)


mrdima:
This is the improved version of my old pull request (https://github.com/puppetlabs/puppet/pull/3610) which I closed. Looking forward to your comments.

kylog:
@Iristyle ping

Iristyle:
Thanks for the updates @mrdima.  This particular issue is being actively discussed internally and will likely see additional follow-up from the Puppet team very shortly as we chart out the next releases.

Thanks for all the work on this.

kylog:
@ferventcoder or @Iristyle any status update?

branan:
@ferventcoder @Iristyle Another triage, another ping from the triage team :)

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -333,6 +343,17 @@ def translate_hash_to_trigger(puppet_trigger)
       self.fail "Unknown schedule type: #{puppet_trigger["schedule"].inspect}"
     end
 
+    if puppet_trigger['minutes_interval']
+      integer_interval = Integer(puppet_trigger['minutes_interval'])
+      self.fail 'minutes_interval must be an integer greater or equal to 0' unless integer_interval >= 0
```
joshcooper: I prefer to avoid double-negatives, so just `if integer_interval < 0`

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -333,6 +343,17 @@ def translate_hash_to_trigger(puppet_trigger)
...
+      trigger['minutes_interval'] = integer_interval
+    end
+    if puppet_trigger['minutes_duration']
+      integer_duration = Integer(puppet_trigger['minutes_duration'])
+      self.fail 'minutes_duration must be an integer greater than minutes_interval' unless integer_duration > integer_interval || (integer_duration == 0 && integer_interval == 0)
```
joshcooper: similar comment here, and since there are many comparisons being made, I prefer to do:

```ruby
if integer_duration <= integer_interval && (integer_duration != 0 || integer_interval != 0)
  self.fail ...
end
```

And now that I write it that may, I'm not sure the logic checks are correct.

lib/puppet/type/scheduled_task.rb
```diff
@@ -92,32 +96,33 @@ def insync?(current)
       A trigger can contain the following keys:
 
       * For all triggers:
-          * `schedule` **(Required)** --- What kind of trigger this is.
-            Valid values are `daily`, `weekly`, `monthly`, or `once`. Each kind
-            of trigger is configured with a different set of keys; see the
-            sections below. (`once` triggers only need a start time/date.)
+          * `schedule` **(Required)** --- The schedule type. Valid values are
+            `daily`, `weekly`, `monthly`, or `once`.
```
joshcooper: I don't think we want to make these changes, beyond just adding the `daily` trigger below. @nfagerlund and I spent some time coming up with "kind of trigger" because "type" is an overloaded term in puppet.

spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -68,6 +68,8 @@ def date_component
     end
   end
 
+
+
```
joshcooper: extra white space

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -141,6 +146,10 @@ def command=(value)
     task.application_name = value
   end
 
+  def comment=(value)
+    task.comment = value
```
joshcooper: Is there a maximum comment length? I ask because we have a bug currently (https://tickets.puppetlabs.com/browse/PUP-3768) where we only handle 256 characters in the `argument` property, and don't want to make the situation worse with comment.

spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -1022,7 +1151,7 @@ def time_component
       end
     end
   end
-
+
```
joshcooper: extra white space

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -101,6 +104,8 @@ def trigger
       puppet_trigger['start_date'] = self.class.normalized_date("#{trigger['start_year']}-#{trigger['start_month']}-#{trigger['start_day']}")
       puppet_trigger['start_time'] = self.class.normalized_time("#{trigger['start_hour']}:#{trigger['start_minute']}")
       puppet_trigger['enabled']    = trigger['flags'] & Win32::TaskScheduler::TASK_TRIGGER_FLAG_DISABLED == 0
+      puppet_trigger['minutes_interval'] = trigger['minutes_interval'] ||= 0
+      puppet_trigger['minutes_duration'] = trigger['minutes_duration'] ||= 0
```
joshcooper: IIUC, `minutes_interval` and `minutes_duration` are only meaningful for `daily` triggers? In which case, I'd prefer to move this logic up to the `when Win32::TaskScheduler::TASK_TIME_TRIGGER_DAILY` block

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -333,6 +343,17 @@ def translate_hash_to_trigger(puppet_trigger)
       self.fail "Unknown schedule type: #{puppet_trigger["schedule"].inspect}"
     end
 
+    if puppet_trigger['minutes_interval']
```
joshcooper: Similar comment as https://github.com/puppetlabs/puppet/pull/3637/files#r26366163, I'd prefer to move this into the `daily` specific logic.

joshcooper:
@mrdima thank you for your contribution on this! This is a highly requested feature, so it's great to see this.

We've had issues in the past with puppet continually reapplying scheduled_tasks because it doesn't correctly read the *current* state, e.g. https://tickets.puppetlabs.com/browse/PUP-1165, https://tickets.puppetlabs.com/browse/PUP-3429

Could you verify that if you run `puppet` with a `daily` task, that it doesn't keep "changing" the system every time it runs? Also what happens if you create the scheduled task, change your system clock to one day in the future, and re-run puppet, does puppet think the task needs to be updated unnecessarily?

lib/puppet/type/scheduled_task.rb
nfagerlund: Yeah exactly. +1 to stick with kind.

lib/puppet/type/scheduled_task.rb
```diff
@@ -92,32 +96,33 @@ def insync?(current)
...
           * `start_time` **(Required)** --- The time of day when the trigger should
             first become active. Several time formats will work, but we
             suggest 24-hour time formatted as HH:MM.
           * `start_date` ---  The date when the trigger should first become active.
             Defaults to the current date. You should format dates as YYYY-MM-DD,
             although other date formats may work. (Under the hood, this uses `Date.parse`.)
-      * For `daily` triggers:
```
nfagerlund: These were wrapped in code spans on purpose. The reason was to reinforce the connection between these groups of keys and whichever allowed value you fed to `schedule`, thus making it easier to read. Likewise with the extra text you deleted above in the `schedule` key. 

Is there a good reason to make these relationships less transparent?

Iristyle:
@mrdima we've decided that this is something we want to incorporate into our upcoming `3.8` release.  Would you be able to address the issues mentioned above and resubmit the pull request against the `3.x` branch?


Thanks!

(Also FYI - we've pulled this issue into our current sprint, so would hope to get this merged in within the next week or so)

mrdima:
I will have try to address the issues / comments tomorrow (tuesday 17th of march) afternoon (CET) or wednesday at latest.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: I would like to leave it as it is, because it matches the coding style of the rest of the checking in this file like above and below with things like self.fail "Must specify 'start_date' when defining a one-time trigger" unless puppet_trigger['start_date'] and self.fail "start_date must be on or after 1753-01-01" unless start_date >= Date.new(1753, 1, 1). Would that be ok?

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: Similar comment, I would prefer to keep using the unless variant because that's what's used as coding style in the rest of the module. The logic with the unless is correct and has been tested. The duration needs to be higher than the interval, except when not using repeat triggers, then they're both 0.

lib/puppet/type/scheduled_task.rb
mrdima: Agree, I probably copy/pasted text from an older version. Corrected in new patch

lib/puppet/type/scheduled_task.rb
mrdima: No good reason, I don't know why these disappeared. I'll correct it on the next patch (if git doesn't play tricks)

spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
mrdima: Corrected in next patch

spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
mrdima: Corrected in next patch

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: There's not a specific maximum length according the API, but we can put a limit;
In the Task Scheduler GUI, 260 chars is the maximum input for a comment which seems to represent MAX_PATH. Everything longer will be truncated when you press Apply/OK if you can actually get it in there (you could using the API! but as said it will be truncated if you make a manual change anyway).
Same goes for command/arguments (and probably the others like account)..also 260 max in gui when starting fresh, but if you got a bigger amount in it seems to support adding even more...! No true limit there really.
I propose changing puppet/util/windows/taskscheduler.rb#702 to 260 (for comment) and #408 to 512 (for arguments) for now which should be enough for most cases(?). And all others using 256 as max should be changed to 260 as well (like working dir etc), because I think it's logical to use MAX_PATH (https://msdn.microsoft.com/en-us/library/aa370993%28v=vs.85%29.aspx)
We could add a check on the length of all these parameters and fail if user supplies a longer argument...do you want that? (and kind of workaround for PUP-3768 on that). I did not change anything regarding this, awaiting your comments.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: No, they are supported for other triggers too (like weekly, monthly, dow, once etc). There might be people to have a use case to have something done each sunday from 08:00am to 09:00am (1 hour duration) for every 10 minutes.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: Same comment, it is supported for non-daily as well, there is no need to enforce a limited use to daily triggers only.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
joshcooper: Our coding style at puppet has evolved since this code was originally written, and our current preference is to avoid `unless`, especially as this is double-negative logic.

I'd be fine if you left this as is, and added another commit that changed the double-negative logic here and `unless start_date >= Date.new(1753, 1, 1)` as well.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
joshcooper: If we're not using repeat triggers and they're both supposed to be zero, but aren't, we'll issue a misleading error message, "minutes_duration must be an integer greater than minutes_interval` I think we should break the error checking to handle the two cases separately.

lib/puppet/type/scheduled_task.rb
mrdima: @Josh,

I tested with daily triggers, changing the clock to the future, everything kept working fine without re-applying. I do want to repeat a caveat from the old pull request (which is the same feature as the current provider has):
There is a small, minor issue though (which I don't care about too much but which should take into account). If you have a scheduled task with repeat triggers and for example a start time. Then you start managing this scheduled task with puppet, leave out the minutes_interval and minutes_duration definition it won't change the repeat triggers when nothing changes (which is what you want because by not defining those optional properties you leave them unmanaged), but when for example the start time changes, the repeat triggers will dissapear because the provider uses a dummy_trigger definition in which they initialize the minutes_interval and minutes_duration to 0. They should take the current values in the scheduled task instead if repeat triggers aren't managed. Actually the current provider has the same issue I guess.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
joshcooper: I think adding length checks to the various properties is worthy of a separate PR (and should be tagged with PUP-3768). Also I do think you want to `fail` if the length is exceeded. If you only `warn` and the value is truncated, then puppet will keep thinking the resource is out-of-sync.

The limits your proposed sound good. Perhaps for command, we follow what CreateProcess does for its lpCommandLine parameter, "The maximum length of this string is 32,768 characters, including the Unicode terminating null character."

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -47,6 +47,10 @@ def command
     task.application_name
   end
 
+  def comment
+    task.comment
+  end
+
```
joshcooper: I'm thinking managing `comment` should be its own JIRA ticket and PR.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: I'm not sure what you mean, I would like to leave it as it is and you can modify the whole file completely into the new coding style someday. The unless start_date >= Date.new(1753, 1, 1) was just an example of what's already in there, I don't want to change anything on that old code.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: Ok, so we'll leave the length checks to a seperate PR (but I cannot promise that I'm volunteering for that, if you expected that anyway ;-).

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: Ok, I thought it was really trivial, but if you want I'm happy to take it out and resubmit.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: I don't see how this could ever happen. If you're not using repeat triggers, they're either nil (not filled in in the manifest) and will be set to 0 by the dummy_trigger (if anything else of the trigger changes) as it always was even without repeat triggers. The API doesn't support putting a greater interval than duration so it won't be coming from current repeat triggers either.

mrdima:
@Iristyle, I can create a new pull request for the 3.8 branch but I'll await Josh's (or yours) renewed comments on taking out managing the comment field and the "unless" stuff (although I got the feeling we're not understanding each other well there. I think I prefer to leave it and if you want to merge just change it to whatever you feel it needs to be).

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: Yes, let's make that part of a separate ticket.

ferventcoder:
I've rebased this onto 3.8 and squashed the commits down. Also removed comments That's over at #3747.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: Removed from updated PR

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: :+1:

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: Comment was removed from the updated PR, so when the other comes up, the validation can come with it.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: Actually would never had made it here in some cases if the Integer call fails

ferventcoder: Updated in new pr

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: image

ferventcoder: One thing I've noticed, it's called `repetition interval` in Windows. Too bad the naming is confusing here.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: Windows allows the values to be equal, so the logic becomes much simpler.
image

ferventcoder: Note that would mean it would repeat exactly one time.

mrdima: That doesn't work on Windows 2003, that's why I put it like that...also the API didn't allow me to put it like that on 2003.
image

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: Yes, it fails if you put something like 'abc' instead of an integer, which is what would be expected.  If you get passed that the integer it should be higher or equal to 0

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: I guess the guy who originally wrote the gem was inspired from the TASK_TRIGGER structure name https://msdn.microsoft.com/en-us/library/windows/desktop/aa383618(v=vs.85).aspx and I thought it would be ok to keep it like that.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: Makes sense, I will switch it back.

ferventcoder: This is all the logic check really needs to do:

`if integer_duration <= integer_interval && integer_duration != 0`

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
ferventcoder: It's likely required.

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
mrdima: That indeed looks much easier to read, I think there's a few rspecs around this..

PR #3640, opened 2015-02-20
(PUP-3090) Set always_cache_features to true for masters

haus:
While for agents caching features would not be good as they can have new
features delivered during the course of a run, for masters this poses no
problems and decreases overhead during catalog compilation. Without
this, masters will reset and reinterrogate features during the course of
operation which is expensive.

### DIFF
--- a/conf/puppet.conf
+++ b/conf/puppet.conf
@@ -7,3 +7,7 @@
 
     # Where Puppet PID files are kept.
     rundir=/var/run/puppetlabs
+
+    # Masters won't get new features while compiling catalogs for agents, so it
+    # is safe to set this to true
+    always_cache_features=true


shrug:
:+1:

joshcooper:
yep, the expensive part is https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/rubygems.rb#L50. We clear the gem directory cache as a resource may install a new gem, which won't be visible until the cache is cleared.

It feels wrong to me to add this to puppet.conf that puppet-agent lays down... I was thinking puppet-server packaging would do this, somewhere in https://github.com/puppetlabs/puppet-server/blob/master/resources/ext/config/conf.d/puppetserver.conf#L2-L15, though it does require more server side work to plumb the setting through.

Also changing puppet-server won't help webrick or passenger, but I don't know that I care.

haus:
@joshcooper yea I'm all for all of these master settings being added or handled by puppet-server, i just don't know that it's ready today or in scope for the current release. @nwolfe @cprice404 thoughts?

kylog:
Is it wrong to add this one setting to puppet-agent's puppet.conf, or is it wrong in general that puppet-agent's puppet.conf is all about master settings?

haus:
@kylog yes

joshcooper:
It seems wrong that puppet-agent is specifying how puppet-server should behave, when puppet-server is the "application" using the puppet "library"

cprice404:
I agree with @kylog ; the fact that there is a `[master]` section at all in the `puppet.conf` file created by the `puppet-agent` packaging seems hokey.  That  said, if there's going to be such a section, adding this setting to it doesn't seem like we're making things any worse than they already are, IMO.

Ideally if we're going to try to move the responsibility of configuring the `[master]` section over to the Puppet Server packaging, we'd do the whole section all at once?  If so, then it's maybe a question of whether now is the right time to address that, and I honestly don't know :)

@haus when you say "I'm all for these settings being handled by puppet-server", do you mean by putting them in our own config files and no longer supporting the `[master]` section in `puppet.conf`?  Or do you mean by having the packaging for Puppet Server be responsible for adding the appropriate settings to the shared `puppet.conf` file?  If the former, we're probably not ready yet; I think we'd need to put some deprecation warnings in to Puppet for a while first.  If the latter, it's just a matter of whether there's a sane way to do that during our package installation... if there is, it's fine by me.

haus:
@cprice404 mostly i mean the former (or some variation on the vardir override that puppet-server already has), but i get that it isn't in the cards today. the latter is pretty hard to do in a sane way in packaging, because puppet-agent owns the config file, and having another package change the file breaks assumptions there.

joshcooper:
So puppet-server already changes puppet.conf during post-install: https://github.com/puppetlabs/puppet-server/pull/411/files#r25198384

cprice404:
I was under the impression that a lot of that stuff in puppet server's postinst was considered to be 'hacks' that @haus wanted to get rid of.  But I'm fine with this being solved there if @haus is.

KaysLounge:
Stop

haus:
Calling config set as is being done in this PR https://github.com/puppetlabs/puppet-server/pull/411#discussion-diff-25198384R6 is a way of getting rid of this section, and it's probably fine. The only weird case is if a user has updated puppet.conf's master section to change the defaults being set by puppetserver, there isn't a good way of knowing.

joshcooper:
If puppet-server postinst changes the value, then it's going to run `puppet config set ...` and overwrite whatever may have been in `puppet.conf`.

If we go with how this PR is currently implemented (in puppet-agent), will it apply the new puppet.conf? Or does it only create the file if it doesn't exist?

haus:
@joshcooper It depends a little on how a user's package manager is configured. In the general case, the puppet.conf as packaged will be laid down and updated as long as the user has not modified the file. puppetserver running `puppet config set` in the post will update the file, so the immediate consequence is that future updates would not be applied automatically to the file. However, I imagine most master puppet.conf files are updated for various customizations, so this might be a moot point.

camlow325:
> If puppet-server postinst changes the value, then it's going to run puppet config set ... and overwrite whatever may have been in puppet.conf.

Yeah, this is the part that would concern me:

1) At package installation time, puppet-server postinst config sets the value to true.

2) User edits the file, changing the value to false.

3) At package upgrade time, puppet-server postinst config automatically resets the value to true.

I suppose the user might have other conflicts to manage if the puppet.conf had changed from one version to the next.  But if it didn't, it seems like we might have automatically reverted the user's selection without any warning, which seems bad.  Unless I'm not understanding how this all works...

If the scenario I mentioned above is plausible and we don't have a better story for how to sanely encapsulate the master settings in master-specific packaging for now, then I'd be in favor of managing the setting in puppet-agent's default puppet.conf - even though I agree that this would be architecturally distasteful for the long term.

haus:
@camlow325 I think the best course in the immediate term is probably to pull these 4 settings into puppetserver's postinst, and just do a check to see if there is a master setting for that already, and if not, set it. That way, a user can change the setting and not have it be overwritten by puppetserver.

camlow325:
>  just do a check to see if there is a master setting for that already

What's the best way to do that?  If you ask for the value of the setting from the CLI config tool, I think you'll get a value back whether or not one was explicitly set in the config file -- not sure you can tell the difference.  Suppose we could scrape the puppet.conf file directly to see if it is set, but that seems hacky.

haus:
@camlow325 yea, you're right. despite the description of `puppet config print` (This subcommand can inspect and modify settings from Puppet's 'puppet.conf' configuration file.), it makes no distinction between a config value that is the default and a config value that is not set in puppet.conf. So it would be some sort of scraping, which i agree is hacky and fragile.

joshcooper:
So @camlow325 @haus the short term fix is to merge this as is, and someday change puppetserver package to use its own `puppet.conf`?

camlow325:
That's my suggestion for the short-term change - happy to be overruled.  :smile: 

Not sure what the best long-term solution is.  Having the master at least manage the settings that have historically been in the `master` section of the "puppet.conf" file seems like the right direction.  Not sure if that would end up being a new file, "master.conf"?, or a wholesale replacement for "puppet.conf", or something else?  Would be nice, in the process, to land on a solution that eliminates the duplication we're already doing between "puppetserver.conf" and "puppet.conf" for the master's confdir and vardir settings.

haus:
That seems fine to me.
If someday `puppet config print` can distinguish between a setting being the default and a setting not being present in puppet.conf, at which point, my other suggestion becomes more tenable.

cprice404:
I think that the long-term goal should that in Puppet 5, anything that someone might have set in [master] (that is actually related to the master) would just go into the existing puppetserver.conf file.  We need to get rid of run_mode and that would be a step towards it.  run_mode is the only reason we have a duplicate vardir setting.

kylog:
:+1: on merge as-is.

And yeah in puppet 5, this goes away.

joshcooper:
And this has the added benefit of working for webrick and passenger, so merging

joshcooper:
Reverted in https://github.com/puppetlabs/puppet/commit/3dcafadbc208dc216d8b5fd44a8eb19211e9b5db

PR #3643, opened 2015-02-24
(maint) allow puppet server installs from version

er0ck:
Without this change the pre-suite installs puppet-server from
    nightlies/puppetserver-latest which can sometimes be an older version.
    this fix allows one to specify SERVER_VERSION environment variable to
    pull a specific build.
    This change also adds puppetserver-confdir to the aio options file.

### DIFF
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -7,13 +7,14 @@ test_name "Install Packages"
 step "Install repositories on target machines..." do
 
   sha = ENV['SHA']
+  server_version = ENV['SERVER_VERSION'] ||= 'nightly'
   repo_configs_dir = 'repo-configs'
 
   hosts.each do |host|
     install_repos_on(host, 'puppet-agent', sha, repo_configs_dir)
   end
 
-  install_repos_on(master, 'puppetserver', 'nightly', repo_configs_dir)
+  install_repos_on(master, 'puppetserver', server_version, repo_configs_dir)
 end
 
 
@@ -61,4 +62,3 @@ agents.each do |agent|
 end
 
 configure_gem_mirror(hosts)
-


acceptance/setup/aio/pre-suite/020_AIO_Workarounds.rb
```diff
@@ -4,10 +4,10 @@
 
 step "(PUP-4004) Set permissions on puppetserver directories that currently live in the agent cache dir"
 %w[reports server_data yaml bucket].each do |dir|
-  on master, "install --directory /opt/puppetlabs/puppet/cache/#{dir}"
+  on master, "install --directory #{options[:HOSTS][:master][:puppetvardir]}/#{dir}"
```
joshcooper: Ideally, I hope we can do:

```
vardir = master.puppet('master')['vardir']
```
This will effectively run `puppet master --configprint vardir` on the master node, so that we don't need to rely on `vardir` being set in the host yaml files (or beaker). This assumes that the master is setting the correct values in the shared `puppet.conf` and not in some other config file that puppet (ruby) can't see.

acceptance/setup/aio/pre-suite/020_AIO_Workarounds.rb
```diff
@@ -34,10 +34,10 @@
 # of the default master-var-dir in puppetserver, and the user
 # based codedir.
 step "(SERVER-347) Set required codedir setting on puppetserver"
-on master, puppet("config set codedir /etc/puppetlabs/code --section master")
+on master, puppet("config set codedir #{options[:HOSTS][:master][:puppetcodedir]} --section master")
```
joshcooper: same comment here about `codedir`

acceptance/setup/aio/pre-suite/020_AIO_Workarounds.rb
```diff
@@ -34,10 +34,10 @@
...
 
 step "(SERVER-370) overwrite ruby-load-path"
-create_remote_file(master, '/etc/puppetserver/conf.d/os-settings.conf', <<-EOF)
+create_remote_file(master, "#{options[:puppetserver_confdir]}/os-settings.conf", <<-EOF)
```
joshcooper: :+1: Unfortunately I don't think this setting is accessible via (ruby) puppet

er0ck:
thanks @joshcooper.  fixed-up

acceptance/config/aio/options.rb
```diff
@@ -1,6 +1,7 @@
 {
   :type => 'aio',
   :is_puppetserver => true,
+  :puppetserver_confdir => '/etc/puppetlabs/puppetserver/conf.d',
```
joshcooper: If we merge this now, then AIO acceptance will fail running against the *current* nightly puppet-server, until builds with https://tickets.puppetlabs.com/browse/SERVER-387 are promoted to nightly. Can we leave this as `/etc/puppetserver/conf.d` for now? Then when the new puppet-server is promoted, we'll encounter failures on our end telling us to change `puppetserver_confdir`?

joshcooper: Also I just remembered that beaker uses `puppetserver_confdir` when bouncing the puppetserver: https://github.com/puppetlabs/beaker/blob/beaker2.5.0/lib/beaker/dsl/helpers.rb#L551. Will moving this to the options file instead of the host files cause beaker to fall back to its hardcoded (and incorrect) AIO defaults?

joshcooper:
@er0ck can you rebase your PR on master? We recently removed the workaround that overrode `puppetserver_confdir` as the puppetserver packages are doing the right thing now, so you'll be able to remove the change to `options.rb` and various host config files.

acceptance/config/aio/options.rb
er0ck: tested this against various tests that use with_puppet_running_on, and they all work fine.  the options hash is the options hash if it comes from any of the 6! places you can set options

er0ck:
@joshcooper rebased against master and tested against rhel7 and win2012

acceptance/setup/aio/pre-suite/020_AIO_Workarounds.rb
```diff
@@ -1,9 +1,10 @@
 step "(PUP-4004) Set permissions on puppetserver directories that currently live in the agent cache dir"
+vardir = master.puppet('master')['vardir']
 %w[reports server_data yaml bucket].each do |dir|
-  on master, "install --directory /opt/puppetlabs/puppet/cache/#{dir}"
+  on master, "install --directory #{vardir}/#{dir}"
 end
-on master, "chown -R puppet:puppet /opt/puppetlabs/puppet/cache"
-on master, "chmod -R 750 /opt/puppetlabs/puppet/cache"
+on master, "chown -R puppet:puppet #{vardir}"
+on master, "chmod -R 750 #{vardir}"
```
joshcooper: @er0ck could you remove this part, this workaround was removed in  https://github.com/puppetlabs/puppet/commit/482250f2c3032d483565755566ad1e695521d4f1

acceptance/setup/aio/pre-suite/020_AIO_Workarounds.rb
```diff
@@ -1,10 +1,3 @@
-step "(PUP-4004) Set permissions on puppetserver directories that currently live in the agent cache dir"
-%w[reports server_data yaml bucket].each do |dir|
-  on master, "install --directory /opt/puppetlabs/puppet/cache/#{dir}"
-end
-on master, "chown -R puppet:puppet /opt/puppetlabs/puppet/cache"
-on master, "chmod -R 750 /opt/puppetlabs/puppet/cache"
-
```
joshcooper: @er0ck can you undo this change (as it's already been done in https://github.com/puppetlabs/puppet/commit/0c138daad9ba0ed595956a2ba046d0cab08756ef), and update the commit message?

acceptance/config/aio/options.rb
joshcooper: ah, this actually need to be  `:'puppetserver-confdir' => '/etc/puppetlabs/puppetserver/conf.d',` note the dash, which has to be quoted since it's a symbol. I think it was passing for you because the version of beaker we are pinned to already set that as the AIO default.

Let's hold off on this until we get puppet testing against beaker#master

er0ck: weird. i thought beaker converted the dashes in arguments and env vars to underscores for use in its options hash. but this is just an inconsistent convention...

joshcooper:
@er0ck the options.rb and node definition files have been overtaken by events leading up to rc1. Could you update the commit to only contain the `010_Install.rb` change?

er0ck:
@joshcooper do we really want to revert to getting puppetserver-confdir from beaker?  
i suppose that setting really needs to live with the puppet-server code and not with puppet, so i suppose i should remove it from options.rb for that reason alone.
is there a way to query puppet-server for its confdir after its installed?  not sure if that will work in all cases.

joshcooper:
looks good, will merge after rc1 freeze

PR #3644, opened 2015-02-24
(PUP-2929) Optimize function loader performance

thallgren:
This PR contains a series of performance improvements. Each commit covers one specific improvement. The details of each improvement can be found in the commit message.

The 'Use setXXX instead of repeated addXXX ...' commit includes some changes to RGen. A PR with those changes has been submitted here: https://github.com/mthiede/rgen/pull/22

### DIFF
--- a/lib/puppet/pops/adaptable.rb
+++ b/lib/puppet/pops/adaptable.rb
@@ -67,7 +67,7 @@ module Puppet::Pops::Adaptable
     # @raise [ArgumentError] if the object is not adaptable
     #
     def self.get(o)
-      attr_name = :"@#{instance_var_name(self.name)}"
+      attr_name = self_attr_name
       if o.instance_variable_defined?(attr_name)
         o.instance_variable_get(attr_name)
       else
@@ -92,7 +92,7 @@ module Puppet::Pops::Adaptable
     # @raise [ArgumentError] if the given object o is not adaptable
     #
     def self.adapt(o, &block)
-      attr_name = :"@#{instance_var_name(self.name)}"
+      attr_name = self_attr_name
       adapter = if o.instance_variable_defined?(attr_name) && value = o.instance_variable_get(attr_name)
         value
       else
@@ -145,7 +145,7 @@ module Puppet::Pops::Adaptable
     # @return [nil] if the adapter has not been set
     #
     def self.clear(o)
-      attr_name = :"@#{instance_var_name(self.name)}"
+      attr_name = self_attr_name
       if o.instance_variable_defined?(attr_name)
         o.send(:remove_instance_variable, attr_name)
       else
@@ -175,16 +175,27 @@ module Puppet::Pops::Adaptable
       adapter
     end
 
+    DOUBLE_COLON = '::'
+    USCORE = '_'
+
     # Returns a suitable instance variable name given a class name.
     # The returned string is the fully qualified name of a class with '::' replaced by '_' since
     # '::' is not allowed in an instance variable name.
     # @param name [String] the fully qualified name of a class
     # @return [String] the name with all '::' replaced by '_'
     # @api private
-    # @private
     #
     def self.instance_var_name(name)
-      name.split("::").join('_')
+      name.split(DOUBLE_COLON).join(USCORE)
+    end
+
+    # Returns a suitable instance variable name for the _name_ of this instance. The name is created by calling
+    # Adapter#instance_var_name and then cached.
+    # @return [String] the instance variable name for _name_
+    # @api private
+    #
+    def self.self_attr_name
+      @attr_name_sym ||= :"@#{instance_var_name(self.name)}"
     end
   end
 end
--- a/lib/puppet/pops/loader/loader.rb
+++ b/lib/puppet/pops/loader/loader.rb
@@ -137,43 +137,44 @@ class Puppet::Pops::Loader::Loader
   # A name/type combination that can be used as a compound hash key
   #
   class TypedName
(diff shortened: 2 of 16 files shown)


lib/puppet/pops/loader/loader.rb
```diff
@@ -137,43 +137,51 @@ def initialize(typed_name, value, origin)
...
+        @name = name
       end
+      @name_parts = parts
 
-      freeze()
-    end
+      # Not allowed to have numeric names - 0, 010, 0x10, 1.2 etc.
+      # At this point it is known that @name is a relativized string so it's
+      # safe to do a direct regexp match
+      unless Puppet::Pops::Patterns::NUMERIC.match(@name).nil?
```
hlindberg: I wonder if the loader/typed-name really needs to care about numeric names - I don't think there is any ambiguity in the loaders. Names of things should be checked at the source (i.e. "illegal function name", "illegal class name" etc.) as there are far fewer of those than instances of TypedName.

That was a long way of saying: I think we can simply skip this assertion.

(sorry, I first commented on the commit - does not show up in the conversation).

lib/puppet/pops/adaptable.rb
```diff
@@ -184,7 +187,11 @@ def self.associate_adapter(adapter, o)
     # @private
     #
     def self.instance_var_name(name)
-      name.split("::").join('_')
+      name.split(DOUBLE_COLON).join(USCORE)
+    end
+
+    def self.self_attr_name
```
hlindberg: Add documentation and mark it as api private

lib/puppet/vendor/rgen/lib/rgen/metamodel_builder/builder_extensions.rb
```diff
@@ -395,7 +395,7 @@ def get<%= firstToUpper(name) %>
```
hlindberg: Never modify the vendored version directly! Users can install other versions of a vendored library - they are not forced to use what we ship. Patching has to be done through normal coding or worst case monkey patching just as if the library was not vendored.

hlindberg: We have some other patches / additions to rgen - see parser_factory.rb. There is also the vendored code loading that could be used to patch the vendored version IIRC.

lib/puppet/pops/visitor.rb
```diff
@@ -24,11 +24,14 @@ def initialize(receiver, message, min_args=0, max_args=nil)
 
   # Visit the configured receiver
   def visit(thing, *args)
-    visit_this(@receiver, thing, *args)
+    visit_this(@receiver, thing, args)
   end
 
+  DOUBLE_COLON = '::'
+  NO_ARGS = []
```
hlindberg: should be frozen since it is passed on in calls, and faulty code could add values to it (which would be very bad).

lib/puppet/pops/types/types.rb
```diff
@@ -23,18 +23,7 @@ class PAnyType < TypeModelObject
       module ClassModule
         # Produce a deep copy of the type
         def copy
-          return @clone unless @clone.nil?
```
hlindberg: Ideally this revert should go into its own commit

hlindberg:
Are there any benchmark results that show how much this speeds things up?

hlindberg:
Are benchmarks added to the repo yet? They should go in first to enable before/after comparisson.

hlindberg:
Found the relevant benchmarks. Looks like a 4% improvement on type inference and a whopping 3.6x improvement on function loading.

thallgren:
Benchmarks have been added. I mention that in the JIRA.

thallgren:
The 4% improvement on type inference is actually a lot more (about 50%) since the previous code was based on the flawed deep copy. Try without the optimization and with the marshal/unmarshal version of the copy. That will give you the real numbers.

lib/puppet/vendor/rgen_patch.rb
```diff
@@ -0,0 +1,129 @@
+# Patch to improve generated addXXX and setXXX method performance
+# Submitted to rgen as PR: https://github.com/mthiede/rgen/pull/22
+#
+# The current implementation of the generated addXXX method will iterate
+# over the contained array where each iteration will then test for object_id
+# equality. Once the iteration is done and such an entry is found it
+# checks it the added value is an instance of MMBase or MMGeneric. If it
```
hlindberg: typo "it"... "it", the second should be an "if" I think.

lib/puppet/vendor/rgen_patch.rb
```diff
@@ -0,0 +1,129 @@
...
+#
+# This patch changes the implementation so that the test for MMBase/
+# MMGeneric is made prior to the iteration, thus avoiding the iteration
+# alltogether for all non model objects
+#
+# The patch also changes the setXXX method to allow it to optimize the way
+# it ensures that the added elements are unique. Calling the addXXX method
+# to do an increasingly expensive sequential scan is very expensive when
+# large sets of values are assigned.
+#
```
hlindberg: worth mentioning and reference the PR to RGen and that this patch can then be removed.

hlindberg:
Nice, this has an 11% speedup on mixed evaluation as well!

lib/puppet/vendor/rgen_patch.rb
thallgren: It is mentioned. Second line from the top.

PR #3645, opened 2015-02-24
(PUP-4036) Autorequire parents to home directory

raphink:


kylog:
@raphink we chatted about this briefly during PR triage. Two thoughts:
1) this will probably be a breaking change so will need to be aligned on a major version boundary. And it's pretty late in the puppet 4 release cycle, but the time gap from puppet 4 to puppet 5 should be *nowhere near* the 3->4 gap. So now is a good time to engage on this.
2) there was some concern that this approach might mean magical/unexpected behavior. Can you bring this up for discussion on the puppet-dev google group?

Also, the spec failures are interesting. I'd be curious what is going on there.

raphink:
@kylog: thanks for the comment. To be clear, I came upon something like this in our code recently:

```puppet
file { '/srv/home': ensure => directory }
file { '/home': ensure => link, target => '/srv/home' } -> User <| |>
```

It seemed wrong to have to specify manually that `/home` should come before all users when it's used as a parent directory to the user's home dir. I took the code in this PR from the `file` type where it is used for `autorequire` of nearest parent directory, and it made sense to use the same code.

I don't really see how this would be a break change, as it makes a lot of sense imo that parent dirs to the home dir should be managed before the user. However, I do not see the reason why the specs fail, it'll be interesting indeed to figure that out.

I'll be on vacation for the next week, but I'm open to discussing this on puppet-dev at some point.

raphink:
I started a thread on puppet-dev at https://groups.google.com/forum/#!topic/puppet-dev/C5nrwu2fRdI

branan:
Based on the discussion in the thread, and especially the concern over potential resource loops, I'm going to close this.

branan:
If we can find a way to do this in a more focused way that can avoid resource loops we could revisit it, though

PR #3646, opened 2015-02-24
(PUP-4037) Update Windows paths for AIO

joshcooper:
This commit moves `codedir` out of the `confdir` directory, so it
changes from `C:\ProgramData\PuppetLabs\puppet\etc\code` to
`C:\ProgramData\PuppetLabs\code`.

It also renames `vardir` from `C:\ProgramData\PuppetLabs\puppet\var` to
`C:\ProgramData\PuppetLabs\puppet\cache` to match *nix.

This commit also adds tests for `rundir` and `logdir`, similar to what
is done for *nix. Note we have to undefine `HOMEDRIVE` and `USERPROFILE`
as Ruby will fall back to those when trying to expand `~`.

### DIFF
--- a/lib/puppet/util/run_mode.rb
+++ b/lib/puppet/util/run_mode.rb
@@ -77,7 +77,7 @@ module Puppet
 
     class WindowsRunMode < RunMode
       def conf_dir
-        which_dir(File.join(windows_common_base("etc")), "~/.puppet")
+        which_dir(File.join(windows_common_base("puppet/etc")), "~/.puppet")
       end
 
       def code_dir
@@ -85,21 +85,21 @@ module Puppet
       end
 
       def var_dir
-        which_dir(File.join(windows_common_base("var")), "~/.puppet/var")
+        which_dir(File.join(windows_common_base("puppet/cache")), "~/.puppet/var")
       end
 
       def run_dir
-        which_dir(File.join(windows_common_base("var/run")), "~/.puppet/var/run")
+        which_dir(File.join(windows_common_base("puppet/var/run")), "~/.puppet/var/run")
       end
 
       def log_dir
-        which_dir(File.join(windows_common_base("var/log")), "~/.puppet/var/log")
+        which_dir(File.join(windows_common_base("puppet/var/log")), "~/.puppet/var/log")
       end
 
     private
 
       def windows_common_base(*extra)
-        [Dir::COMMON_APPDATA, "PuppetLabs", "puppet"] + extra
+        [Dir::COMMON_APPDATA, "PuppetLabs"] + extra
       end
     end
   end
--- a/spec/unit/util/run_mode_spec.rb
+++ b/spec/unit/util/run_mode_spec.rb
@@ -167,8 +167,8 @@ describe Puppet::Util::RunMode do
     end
 
     describe "#code_dir" do
-      it "has codedir ending in PuppetLabs/puppet/code when run as root" do
-        as_root { expect(@run_mode.code_dir).to eq(File.expand_path(File.join(Dir::COMMON_APPDATA, "PuppetLabs", "puppet", "code"))) }
+      it "has codedir ending in PuppetLabs/code when run as root" do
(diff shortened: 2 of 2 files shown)


lib/puppet/util/run_mode.rb
```diff
@@ -85,21 +85,21 @@ def code_dir
       end
 
       def var_dir
-        which_dir(File.join(windows_common_base("var")), "~/.puppet/var")
+        which_dir(File.join(windows_common_base("puppet/cache")), "~/.puppet/var")
```
ferventcoder: Interesting.

spec/unit/util/run_mode_spec.rb
```diff
@@ -168,7 +168,7 @@
 
     describe "#code_dir" do
       it "has codedir ending in PuppetLabs/puppet/code when run as root" do
-        as_root { expect(@run_mode.code_dir).to eq(File.expand_path(File.join(Dir::COMMON_APPDATA, "PuppetLabs", "puppet", "code"))) }
+        as_root { expect(@run_mode.code_dir).to eq(File.expand_path(File.join(Dir::COMMON_APPDATA, "PuppetLabs", "code"))) }
```
ferventcoder: Update the test description as well.

lib/puppet/util/run_mode.rb
```diff
@@ -85,21 +85,21 @@ def code_dir
...
       end
 
       def run_dir
-        which_dir(File.join(windows_common_base("var/run")), "~/.puppet/var/run")
+        which_dir(File.join(windows_common_base("puppet/var/run")), "~/.puppet/var/run")
```
ferventcoder: Is var sticking around?

lib/puppet/util/run_mode.rb
```diff
@@ -85,21 +85,21 @@ def code_dir
...
       end
 
       def log_dir
-        which_dir(File.join(windows_common_base("var/log")), "~/.puppet/var/log")
+        which_dir(File.join(windows_common_base("puppet/var/log")), "~/.puppet/var/log")
```
ferventcoder: Same question here.

lib/puppet/util/run_mode.rb
ferventcoder: If `var` directory sticks around, and we switch `var_dir` to point to `cache`, that could get somewhat misleading/confusing.

Iristyle: Yeah, I don't think either @joshcooper or I are in disagreement here that this seems a little peculiar.  

However, this is how settings are mapped on the Linux side of the fence and we're ultimately shooting for consistency (even if it doesn't seem to make total sense).

See the full documentation at
https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppet-agent-nix

The relevant / important bits you can see here:

| Setting  | Linux  | Windows |
| :------------ |:---------------|:-----|
| :vardir      | /var/lib/puppet/cache | C:\ProgramData\PuppetLabs\puppet\cache |
| :logdir      | /var/log/puppetlabs        |  C:\ProgramData\PuppetLabs\puppet\var\log |
| :rundir | /var/run/puppetlabs        |    C:\ProgramData\PuppetLabs\puppet\var\run |

spec/unit/util/run_mode_spec.rb
joshcooper: updated

lib/puppet/util/run_mode.rb
ferventcoder: Should the MSI create the cache directory as well?

joshcooper: yes it will need to create both `cache` and `var`. The `:logdir` and `:rundir` will be created by puppet as a result of it being directory settings.

joshcooper: Puppet will create any directory that is a puppet directory-based setting, which includes `vardir` (cache), `confdir`, `logdir`, and `rundir`.

However, the MSI must create directories that are not puppet settings, `C:\ProgramData\PuppetLabs\puppet\var`, including parent directories of things that are settings, e.g. `C:\ProgramData\PuppetLabs\puppet`

Iristyle:
:shipit:

PR #3648, opened 2015-02-24
(PUP-4017) Make parser an environment specific setting

hkenney:
Prior to this pull request, the 'parser' setting was not environment
specific. In order to make it easier for users to migrate to using
the future parser, make 'parser' environment specific.

The environment parameter is optional, and if missing will use
what is configured for all environments.

This mechanism can only be used with directory environments.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -258,6 +258,18 @@ module Puppet
 
   # The single instance used for normal operation
   @context = Puppet::Context.new(bootstrap_context)
+
+  def self.future_parser?
+    env = Puppet.lookup(:current_environment) { return Puppet[:parser] == 'future' }
+    env_conf = Puppet.lookup(:environments).get_conf(env.name)
+
+    if env_conf.nil?
+      # Case for non-directory environments
+      Puppet[:parser] == 'future'
+    else
+      env_conf.parser == 'future'
+    end
+  end
 end
 
 # This feels weird to me; I would really like for us to get to a state where there is never a "require" statement
--- a/spec/integration/parser/environment_spec.rb
+++ b/spec/integration/parser/environment_spec.rb
@@ -1,0 +1,46 @@
+require 'spec_helper'
+
+describe "A parser environment setting" do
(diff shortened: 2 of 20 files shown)


lib/puppet.rb
```diff
@@ -258,6 +258,11 @@ def self.rollback_context(name)
 
   # The single instance used for normal operation
   @context = Puppet::Context.new(bootstrap_context)
+
+  def self.future_parser?
+    env = Puppet.lookup(:current_environment) { return Puppet[:parser] == 'future' }
+    Puppet.lookup(:environments).get_conf(env.name).parser == 'future'
```
hlindberg: get_conf returns nil if it is not a directory env. In that case, return Puppet[:parser]

spec/integration/parser/environment_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'spec_helper'
+
+describe "A parser environment setting" do
+
+  let(:confdir) { Puppet[:confdir] }
```
hlindberg: I think this needs to point to a tmp directory in order for the tests to not clutter the file system, then all the files can be regular files under that tmp directory (they will all then be cleaned up).

spec/integration/parser/environment_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  let(:testingdir) { File.join(environmentpath, "testing") }
+
+  before(:each) do
+    FileUtils.mkdir_p(testingdir)
+  end
+
+  it "selects the given parser when compiling" do
+    manifestsdir = File.expand_path("manifests", confdir)
+    FileUtils.mkdir_p(manifestsdir)
+
```
hlindberg: You also want to write out a puppet.conf that has parser set to 'current' ( to the tmp directory)

hlindberg:
Will merge this and then the merge up to master manually (at the same time) as there will be merge conflicts and fixups that needs to take place when merged to master.

PR #3649, opened 2015-02-25
(PUP-1085) Be more specific about expected exception

joshcooper:
Previously, the pacman unit test "should raise a Puppet::Error when
trying to install a group and allow_virtual is false" assumed
`allow_virtual` was false, and tried to install a package, assuming the
provider would raise a Puppet::Error.

However, in Puppet 4, `allow_virtual` is true, so the test was not
failing early as expected, and instead would actually try to execute:

    /usr/bin/pacman --noconfirm --needed --noprogressbar -Sy package'

Puppet raises Puppet::ExecutionFailure on systems without the `pacman`
executable. However, this went unnoticed on *nix, because the test was
only expecting Puppet::Error, which is the superclass for
Puppet::ExecutionFailure. This did not go unnoticed on Windows, because
puppet raises a different exception:

    Errno::ENOENT: No such file or directory - CreateProcess

This commit modifies the unit test to expect specific messages. It also
sets state directly on the resource object, e.g.
`resource[:allow_virtual] = false`. The `resource` is defined in a
`let`, so a new instance is created for each example as needed. Also
note, since `:allow_virtual` specifies `:boolean => true`, puppet
automatically creates a predicate method of the same name, e.g.
`allow_virtual?`.

### DIFF
--- a/spec/unit/provider/package/pacman_spec.rb
+++ b/spec/unit/provider/package/pacman_spec.rb
@@ -43,17 +43,22 @@ describe Puppet::Type.type(:package).provider(:pacman) do
       executor.stubs(:execute).returns("")
       provider.expects(:query).returns(nil)
 
-      expect { provider.install }.to raise_exception(Puppet::Error)
+      expect {
+        provider.install
+      }.to raise_exception(Puppet::Error, /Could not find package/)
     end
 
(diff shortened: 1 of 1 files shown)


kylog:
Thanks for fixing this up!

PR #3650, opened 2015-02-25
(PUP-1806) Add varargs support to 4x function API

thallgren:
This commit adds additional methods to the 4x Function disptach to
allow declarations of required, optional, or repeated parameters. The
new methods replaces the current way of defining this using arg_count()
and last_captures_rest() and those methods are therefore removed. 

NOTE: Care must be taken when merging this to master since new 4x
functions have been added there that makes use of the arg_count method.

### DIFF
--- a/lib/puppet/functions.rb
+++ b/lib/puppet/functions.rb
@@ -57,13 +57,13 @@
 # @example Dispatching to different methods by type
 #   Puppet::Functions.create_function('math::min') do
 #     dispatch :numeric_min do
-#       param 'Numeric', 'a'
-#       param 'Numeric', 'b'
+#       param 'Numeric', :a
+#       param 'Numeric', :b
 #     end
 #
 #     dispatch :string_min do
-#       param 'String', 'a'
-#       param 'String', 'b'
+#       param 'String', :a
+#       param 'String', :b
 #     end
 #
 #     def numeric_min(a, b)
@@ -82,28 +82,52 @@
 # be the same as the number of parameters, and all of the parameters are of
 # type 'Any'.
 #
-# To express that the last parameter captures the rest, the method
-# `last_captures_rest` can be called. This indicates that the last parameter is
-# a varargs parameter and will be passed to the implementing method as an array
-# of the given type.
+# The following methods can be used for declaring a parameter
 #
-# When defining a dispatch for a function, the resulting dispatch matches
-# against the specified argument types and min/max occurrence of optional
-# entries. When the dispatch makes the call to the implementation method the
-# arguments are simply passed and it is the responsibility of the method's
-# implementor to ensure it can handle those arguments (i.e. there is no check
-# that what was declared as optional actually has a default value, and that
-# a "captures rest" is declared using a `*`).
+#  - _param_ - the parameter must be given in a call.
+#  - _optional_param_ - the parameter may be missing in the call. May not be used on a parameter that is followed by a required parameter
+#  - _repeated_param_ - the type specifies a repeating type that occurs 0 to "infinite" number of times. It may only appear last or just before a block parameter.
+#  - _block_param_ - a block must be given in the call. May only appear last.
+#  - _optional_block_param_ - a block may be given in the call. May only appear last.
 #
-# @example Varargs
+# The method name _required_param_ is an alias for _param_ and _required_block_param_ is an alias for _block_param_
+#
+# A parameter declaration takes 2 arguments:
+#  - _type_ A string that must conform to a type in the puppet language
+#  - _name_ A symbol denoting the parameter name
+#
+# Both arguments are optional when declaring a block parameter. The _type_ defaults to "Callable"
+# and the _name_ to :block.
+#
+# When defining a dispatch for a function, the resulting dispatch matches against the specified
+# argument types. When the dispatch makes the call to  the implementation method the arguments are
+# simply passed and it is the responsibility of the method's implementor to ensure it can handle
+# those arguments.
+#
+# @example Variable number of arguments
 #   Puppet::Functions.create_function('foo') do
 #     dispatch :foo do
-#       param 'Numeric', 'first'
-#       param 'Numeric', 'values'
-#       last_captures_rest
+#       param 'Numeric', :first
+#       repeated_param 'Numeric', :values
 #     end
 #
 #     def foo(first, *values)
+#       # do something
+#     end
+#   end
+#
+# There is no requirement for direct mapping between parameter declarations and the parameters in the receiver so
+# the following example is also legal. Here the dispatch will ensure that *values in the receiver will
+# be an array with at least one entry of type String and that any remaining entries are of type Numeric:
+#
+# @example Inexact mapping or parameters
+#   Puppet::Functions.create_function('foo') do
+#     dispatch :foo do
+#       param 'String', :first
+#       repeated_param 'Numeric', :values
+#     end
+#
+#     def foo(*values)
 #       # do something
 #     end
 #   end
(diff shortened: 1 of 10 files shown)


hlindberg:
We decided to use specific methods `param`, `optional_param` etc. instead of passing an extra argument. Please update the PR.

lib/puppet/functions.rb
```diff
@@ -82,25 +82,33 @@
 # be the same as the number of parameters, and all of the parameters are of
 # type 'Any'.
 #
-# To express that the last parameter captures the rest, the method
-# `last_captures_rest` can be called. This indicates that the last parameter is
-# a varargs parameter and will be passed to the implementing method as an array
-# of the given type.
+# The following methods can be used for declaring a parameter
```
hlindberg: "The following methods can be used to declare a parameter"  (or is it define a parameter?)

lib/puppet/functions.rb
```diff
@@ -82,25 +82,33 @@
...
 #
-# When defining a dispatch for a function, the resulting dispatch matches
-# against the specified argument types and min/max occurrence of optional
-# entries. When the dispatch makes the call to the implementation method the
-# arguments are simply passed and it is the responsibility of the method's
-# implementor to ensure it can handle those arguments (i.e. there is no check
-# that what was declared as optional actually has a default value, and that
-# a "captures rest" is declared using a `*`).
+#  - _param_ - the parameter must be given in a call.
```
hlindberg: "in a call" -> "in the call"

lib/puppet/functions.rb
```diff
@@ -82,25 +82,33 @@
...
+#  - _optional_param_ - the parameter may be missing in the call. May not be used on a parameter that is followed by a required parameter
+#  - _repeated_param_ - the type specifies a repeating type that occurs 0 to "infinite" number of times. It may only appear last or just before a block parameter.
+#  - _block_param_ - a block must be given in the call. May only appear last.
+#  - _optional_block_param_ - a block may be given in the call. May only appear last.
```
hlindberg: for all of the above, it is an "argument" that may be missing - the "parameter" is what receives the argument - please rephrase

lib/puppet/functions.rb
```diff
@@ -82,25 +82,33 @@
...
 #
-# @example Varargs
+# The method name _required_param_ is an alias for _param_ and _required_block_param_ is an alias for _block_param_
+#
+# A parameter declaration takes 2 arguments:
```
hlindberg: A parameter definition...

lib/puppet/functions.rb
```diff
@@ -82,25 +82,33 @@
...
+#  - _type_ A string that must conform to a type in the puppet language
+#  - _name_ A symbol denoting the parameter name
+#
+# Both arguments are optional when declaring a block parameter. The _type_ defaults to "Callable"
```
hlindberg: defining

lib/puppet/functions.rb
```diff
@@ -82,25 +82,33 @@
...
+# and the _name_ to :block.
+#
+# When defining a dispatch for a function, the resulting dispatch matches against the specified
+# argument types. When the dispatch makes the call to  the implementation method the arguments are
```
hlindberg: eh - the argument types are not specified (types of arguments are inferred), did you mean the parameter definitions?

lib/puppet/functions.rb
```diff
@@ -108,6 +116,22 @@
 #     end
 #   end
 #
+# There is no requirement for direct mapping between parameter declarations and the parameters in the receiver so
```
hlindberg: definitions instead of declarations. "receiver" is also referred to as "imlementation method" (maybe explain that separately).

lib/puppet/functions.rb
```diff
@@ -108,6 +116,22 @@
...
+# the following example is also legal. Here the dispatch will ensure that *values in the receiver will
```
hlindberg: should be `*values` (in code style, otherwise hard to read)

lib/puppet/functions.rb
```diff
@@ -260,25 +284,51 @@ def initialize(dispatcher, type_parser, all_callables)
       @dispatcher = dispatcher
     end
 
-    # Defines a positional parameter with type and name
+    # Defines a required positional parameter with a _type_ and a _name_.
```
hlindberg: remove the two "a"

lib/puppet/functions.rb
```diff
@@ -260,25 +284,51 @@ def initialize(dispatcher, type_parser, all_callables)
...
-        raise ArgumentError, "Type signature argument must be a String reference to a Puppet Data Type. Got #{type.class}"
-      end
+      internal_param(type, name)
+      raise ArgumentError, 'A required parameter cannot be added after an optional parameter' if @min != @max
+      @min += 1
+      @max += 1
+    end
+    alias required_param param
+
+    # Defines an optional positional parameter with a _type_ and a _name_.
```
hlindberg: Drop the two "a"

lib/puppet/functions.rb
```diff
@@ -260,25 +284,51 @@ def initialize(dispatcher, type_parser, all_callables)
...
+    # May not be used on a parameter that is followed by a required parameter
```
hlindberg: "May not be followed by a required_parameter"

lib/puppet/functions.rb
```diff
@@ -260,25 +284,51 @@ def initialize(dispatcher, type_parser, all_callables)
...
+    #
+    # @param type [String] The type specification for the parameter.
+    # @param name [Symbol] The name of the parameter. This is primarily used
+    #   for error message output and does not have to match the name of the
+    #   parameter on the implementation method.
```
hlindberg: "and does not have to match an implementation method parameter" (I don't think of parameters as being "on" a method)

lib/puppet/functions.rb
```diff
@@ -260,25 +284,51 @@ def initialize(dispatcher, type_parser, all_callables)
...
+    # @return [Void]
+    #
+    # @api public
+    def optional_param(type, name)
+      internal_param(type, name)
+      @max += 1
+    end
+
+    # Defines a repeated positional parameter with a _type_ and a _name_ that occurs 0 to "infinite" number of times.
```
hlindberg: Drop the two "a"s

lib/puppet/functions.rb
```diff
@@ -260,25 +284,51 @@ def initialize(dispatcher, type_parser, all_callables)
...
+    # It may only appear last or just before a block parameter.
+    #
+    # @param type [String] The type specification for the parameter.
+    # @param name [Symbol] The name of the parameter. This is primarily used
+    #   for error message output and does not have to match the name of the
+    #   parameter on the implementation method.
```
hlindberg: same comment as for similar text above

hlindberg:
Looks good - only have comments about documentation comments

lib/puppet/functions.rb
thallgren: You tell me. This is your wording, not mine.

lib/puppet/functions.rb
thallgren: What markup is used for code style here?

lib/puppet/functions.rb
thallgren: It must be either spelled out as I did, or 'param' or it's alias 'required_param'. There's no such thing as 'required_parameter'. I chose to spell it out since 'required_param' is an alias and not often used. I still think that's the best option.

lib/puppet/functions.rb
thallgren: Again, your wording, not mine. I just copied what was there before.

lib/puppet/functions.rb
hlindberg: well, then I wrote something that was confusing - was this somewhere else in the code before or is it from the spec? (i.e. do I need to fix this someplace else?).

hlindberg: "Note that the dispatch definition is used to match arguments given in a call to the function with the defined parameters. It then dispatches the call to the implementation method simply passing the given arguments on to that method without any further processing and it is the responsibility of that method's implementor to ensure that it can handle those arguments."

lib/puppet/functions.rb
hlindberg: here in github comments? It is the same as in markdown - backtick before and after (klick on edit to see how it is done). You can create a block by using three backticks before and after like this:
```
this is code
```

lib/puppet/functions.rb
hlindberg: ok, that was unintentional, I meant "May not be followed by a required parameter"  it is the "used on a.." that is bugging me.

lib/puppet/functions.rb
hlindberg: yea, so I am not always brilliant :-)

hlindberg:
Oy, what happened with the branch? (Did you want to trigger travis? - you can do that on the travis page)

lib/puppet/functions.rb
thallgren: No need to fix elsewhere. It was on line 90 in this file.

thallgren:
I got a bit too ambitious when cleaning up old branches.

hlindberg:
I am merging this manually as it requires fixing up things on master before it is also merged there.

PR #3655, opened 2015-02-26
PUP-4046 correct require in hiera_include

richardc:
Here we correct the include of hiera_puppet to one of hiera/puppet_function,
allowing Hiera::PuppetFunction to be resolved.


richardc:
As noted in PUP-4064 this is only a partial solution to the wider problem of using hiera_include with puppet apply and just moves the problem along.

PR #3656, opened 2015-02-26
(PUP-4033) Test for correct puppet configuration

MikaelSmith:
Tests that the results of `puppet agent --genconfig` and `puppet agent
--configprint` match puppet-specifications, and that files that are
expected to exist from the installer and agent runs are in the locations
specified in said specification.

### DIFF
--- a/acceptance/tests/ensure_puppet-agent_paths.rb
+++ b/acceptance/tests/ensure_puppet-agent_paths.rb
@@ -1,0 +1,86 @@
+# ensure installs and code honor new puppet-agent path spec:
+# https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md
+test_name 'PUP-4033: Ensure aio path spec is honored'
+
+# include file_exists?
+require 'puppet/acceptance/temp_file_utils'
+extend Puppet::Acceptance::TempFileUtils
+
+config_options = [
+  # code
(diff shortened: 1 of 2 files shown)


acceptance/tests/ensure_puppet-agent_paths.rb
```diff
@@ -0,0 +1,91 @@
...
+      end
+    when :file
+      if !file_exists?(agent, path)
+        fail_test("Failed to find expected file '#{path}' on agent '#{agent}'")
+      end
+    end
+  end
+end
+
+step 'test puppet binaries exist'
```
MikaelSmith: @er0ck Thoughts on testing this? Binaries for AIO vs source builds will be in different places.

MikaelSmith:
Passes ci:test:git and ci:test:aio on Linux, ci:test:git on Windows. Still pending ci:test:aio run with Windows.

acceptance/tests/ensure_puppet-agent_paths.rb
er0ck: yeah.  i was concerned about that.  is there a way we can tell if we're in an AIO or source produced environment?
oh: duh  option[:type]  should tell us if we're aio, git, package, etc

MikaelSmith:
Passes Windows AIO run with the test for a file existing at :rest_authconfig removed. That change makes sense, since it's not setup by installation.

er0ck:
(y) if this is ready to go

MikaelSmith:
Should be ready.

PR #3348, opened 2014-11-26
(PUP-3706) console format should pretty print nested structures

dalen:
This makes the console format actually pretty print nested structures.
It already said it would, but didn't in fact.

PR #3658, opened 2015-02-26
(PUP-3706) console format should pretty print nested structures

ScottGarman:
This is a rebase and trivial rspec update to @dalen's [PR-3348](https://github.com/puppetlabs/puppet/pull/3348):

This makes the console format actually pretty print nested structures.
It already said it would, but didn't in fact.

### DIFF
--- a/lib/puppet/network/formats.rb
+++ b/lib/puppet/network/formats.rb
@@ -151,7 +151,7 @@ Puppet::Network::FormatHandler.create(:console,
     end
 
     # ...or pretty-print the inspect outcome.
-    return json.render(datum)
+    return JSON.pretty_generate(datum)
   end
 
   def render_multiple(data)
--- a/lib/puppet/network/format_support.rb
+++ b/lib/puppet/network/format_support.rb
@@ -92,6 +92,10 @@ module Puppet::Network::FormatSupport
     to_data_hash.to_pson(*args)
   end
 
+  def to_json(*args)
+    to_data_hash.to_json(*args)
+  end
+
   def render(format = nil)
     format ||= self.class.default_format
 
--- a/spec/unit/application/face_base_spec.rb
+++ b/spec/unit/application/face_base_spec.rb
@@ -325,9 +325,9 @@ describe Puppet::Application::FaceBase do
         end
       end
 
-      it "should render a non-trivially-keyed Hash with using JSON" do
+      it "should render a non-trivially-keyed Hash with using pretty printed JSON" do
         hash = { [1,2] => 3, [2,3] => 5, [3,4] => 7 }
-        expect(app.render(hash, {})).to eq(hash.to_pson.chomp)
+        expect(app.render(hash, {})).to eq(JSON.pretty_generate(hash).chomp)
       end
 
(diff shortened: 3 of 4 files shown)


dalen:
The JSON format handler should say PUP-3540 instead of PUP-3355, I made a mistake when writing the commit message.

ScottGarman:
Thanks for catching that @dalen, I've fixed that commit now.

lib/puppet/network/formats.rb
```diff
@@ -151,7 +151,7 @@ def render(datum)
     end
 
     # ...or pretty-print the inspect outcome.
-    return json.render(datum)
+    return JSON.pretty_generate(datum)
```
joshcooper: This is causing a problem in CI when `datum` is `true`, because our `json.render` method accepts datum that JSON.pretty_generate does not:

```
(rdb:1) datum
true
(rdb:1) datum.class
TrueClass
(rdb:1) json.render(datum)
"true"
(rdb:1) JSON.pretty_generate(datum)
*** JSON::GeneratorError Exception: only generation of JSON objects or arrays allowed
```

joshcooper: Puppet's `json.render` method uses PSON which monkey patches `to_pson` on a number of classes: `Object`, `Hash`, `Array`, `Integer`, `Float`, `String`, `TrueClass`, `FalseClass`, `NilClass`. The `:console` renderer short-circuits evaluation of [`String` and `Numeric`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/formats.rb#L127-L128), but not `Object`, `TrueClass`, `FalseClass`, `NilClass`. Also when converting Hash, the renderer is still calling `json.render` for its [nested values](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/formats.rb#L136).

I'm thinking we either need to switch everything over to using `JSON.pretty_generate` and ensure the console renderer short-circuits the other classes that PSON monkey patches, e.g. `TrueClass`.

Or we go back to `json.render(datum)` and deal with eliminating PSON another day.

As much as I don't like PSON, I think it would be cleaner and less risky to do the latter, and then remove all of PSON in one PR at a future date.

dalen: Yeah, the first version of this just used `PSON.pretty_generate`. I think it makes sense to deal with this patch and the PSON -> JSON transition as two separate things.

But short-circuit evaluation of things not responding to `to_json` seems like a reasonable workaround otherwise. This renderer already has a bunch of short circuit renders, so wouldn't really hurt to add a few more.

hlindberg:
having "to_this", and "to_that" on classes is a general nuisance from a separation of concerns perspective - avoid such constructs - we already have too many...

lib/puppet/network/formats.rb
joshcooper: I opened a PR https://github.com/puppetlabs/puppet/pull/3672

@dalen sorry that we led you down a dark alley way to switch from PSON.pretty_generate to JSON.pretty_generate, only to find out it doesn't work.

PR #3660, opened 2015-02-27
(PUP-2455) Update service provider to use 'ctrun' on Solaris

geoffnichols:
Prior to this commit, non-SMF Puppet-managed services on Solaris
are not started in their own SMF contracts. If started
by Puppet running as a daemon, the services end up in Puppet's
SMF contract.

There are several negative implications from this scenario:

(1) If Puppet is stopped or restarted, the non-SMF Puppet-managed
services in Puppet's SMF contract are killed (by SMF).

(2) If Puppet-managed services are in Puppet's SMF contract, certain
events impacting any one of the managed services can lead to SMF killing
killing Puppet's contract, also killing all Puppet-managed services
that have ended up in Puppet's SMF contract.

In both scenarios, SMF also places Puppet in the SMF "maintenance"
state.

This commit updates the Puppet service provider to use the Solaris
'ctrun' facility to start Puppet-managed services in their own
SMF contracts.

### DIFF
--- a/acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
+++ b/acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
@@ -1,0 +1,98 @@
+test_name "(PUP-2455) Service provider should start Solaris init service in its own SMF contract"
+
+step "Setup fixture service manifest on master"
+
+  testdir = master.tmpdir('solaris_services_in_own_smf_contract')
+
+  test_manifest = <<MANIFEST
+File {
+  ensure => directory,
+  mode => "0750",
+  owner => #{master.puppet['user']},
+  group => #{master.puppet['group']},
+}
+file { '#{testdir}': }
+file { '#{testdir}/environments': }
+file { '#{testdir}/environments/production': }
+file { '#{testdir}/environments/production/manifests': }
+file { '#{testdir}/environments/production/manifests/site.pp':
+  ensure  => file,
+  content => '
+    node default {
+      service { "vmware-tools":
+      provider => "init",
+      enable   => "true",
+      ensure   => "running",
+      status   => "/etc/init.d/vmware-tools status",
+      }
+    }
+  ',
+}
+MANIFEST
+
+  apply_manifest_on master, test_manifest
+
+step "Start master"
+  
+  if master.is_pe?
+    master['puppetservice'] = 'pe-puppetserver'
+    agent_service           = 'pe-puppet'
+  else
+    master['puppetservice'] = 'puppetserver'
+    agent_service           = 'puppet'
+  end
+ 
+  master_opts = {
+    'main' => {
+    'environmentpath' => "#{testdir}/environments",
+    }
+  }
+
+  with_puppet_running_on master, master_opts, testdir do
+
+    agents.each do |agent|
+
+    fixture_process = 'vmtoolsd'
+    fixture_resource = 'vmware-tools'
+    fixture_service_stop = '/etc/init.d/vmware-tools stop'
+
+      arch = on(agent, facter('architecture')).stdout.chomp
+      plat = on(agent, facter('osfamily')).stdout.chomp
+
+      skip_test unless arch == "i86pc" and plat == "Solaris" 
+
+      step "Start the fixture service on #{agent} "
+        on(agent, "puppet resource service #{fixture_resource} provider=init ensure=stopped")
+        on(agent, "puppet resource service #{fixture_resource} provider=init ensure=running")
+
+        assert_match(/ensure changed 'stopped' to 'running'/, stdout, "The fixture service #{fixture_process} is not in a testable state on #{agent}.")
+
+      step "Verify whether the fixture process is alone in its SMF contract on #{agent}"
+        service_ctid = on(agent, "sleep 20;ps -eo ctid,args | grep #{fixture_process} | grep -v grep | awk '{print $1}'").stdout.chomp.to_i
+        number_in_contract = on(agent, "pgrep -c #{service_ctid} | wc -l").stdout.chomp.to_i
+
+        assert(number_in_contract == 1, "The fixture process #{fixture_process} is not alone in its SMF contract on #{agent}.")
+
+      if agent.is_pe? 
+
+        step "Stop puppet on #{agent}"
+          on(agent, "svcadm disable #{agent_service};sleep 70;svcadm disable #{agent_service}")
+
+        step "Stop fixture service on #{agent}"
+          on(agent, "#{fixture_service_stop}")
+
+        step "Enable puppet service on #{agent}"
+          on(agent, "svcadm enable #{agent_service};sleep 10") do
+            puppet_ctid = on(agent, "svcs -Ho CTID #{agent_service} | awk '{print $1}'").stdout.chomp.to_i
+            service_ctid = on(agent, "ps -eo ctid,args | grep #{fixture_process} | grep -v grep | awk '{print $1}'").stdout.chomp.to_i
+
+          step "Compare SMF contract ids for puppet and #{fixture_process} on #{agent}"
+            unless ( puppet_ctid != "0" and service_ctid != "0" ) then
+              fail_test("SMF contract ids should not equal zero.")
+            end
+
+            assert(service_ctid != puppet_ctid, "Service is in the same SMF contract as puppet on #{agent}.")
+         end 
+       end
+    end
+  end
--- a/acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
+++ b/acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
@@ -1,5 +1,45 @@
 test_name "(PUP-2455) Service provider should start Solaris init service in its own SMF contract"
 
+sleepy_daemon_script = <<SCRIPT
+#!/usr/bin/env ruby
+while true
+  sleep (2)
+end
+SCRIPT
+
+sleepy_daemon_initscript = <<INITSCRIPT
+#!/bin/sh
+FIXTURESERVICE="/tmp/sleepy_daemon"
+start(){
+    $FIXTURESERVICE &
+}
+
+stop(){
+    FIXTUREPID=`ps -ef | grep "$FIXTURESERVICE" | grep -v grep | awk '{print $2}'`
+    kill -9 $FIXTUREPID
+}
+
+status(){
+    FIXTUREPID=`ps -ef | grep "$FIXTURESERVICE" | grep -v grep | awk '{print $2}'`
+    if [ "x$FIXTUREPID" = "x" ]; then
+      exit 1
+    else
+      exit 0
+    fi
+}
+case "$1" in
+    start)
+        start
+        ;;
+    stop)
+        stop
+        ;;
+    status)
+        status
+        ;;
+esac
+INITSCRIPT
+
 step "Setup fixture service manifest on master"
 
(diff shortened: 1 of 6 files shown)


lib/puppet/provider/service/service.rb
```diff
@@ -36,6 +36,9 @@ def ucommand(type, fof = true)
     else
       cmd = [send("#{type}cmd")].flatten
     end
+    if (( Facter.value(:operatingsystem) == "Solaris" ) and ( type =~ /^start/ ))
```
stahnma: Is this the case only on Solaris, or could this happen on other Solaris derived platforms? (Illuminos, OpenIndiana, Nextenta, etc?)

acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
```diff
@@ -0,0 +1,98 @@
...
+}
+file { '#{testdir}': }
+file { '#{testdir}/environments': }
+file { '#{testdir}/environments/production': }
+file { '#{testdir}/environments/production/manifests': }
+file { '#{testdir}/environments/production/manifests/site.pp':
+  ensure  => file,
+  content => '
+    node default {
+      service { "vmware-tools":
```
stahnma: I have a concern, not sure it block a merge, but if somebody is running the tests somewhere that isn't a VMware shop? Is there any daemon that's more universal to use for an example?

lib/puppet/provider/service/service.rb
geoffnichols: I'll need to do some additional investigation to confirm, but I suspect this would apply to other Solaris-derived platforms. It looks like spec tests fail if we changen```Facter.value(:operatingsystem)``` to ```Facter.value(:osfamily)```. If this is the intended design, the Solaris-derived platforms may need to be enumerated explicitly here.

lib/puppet/provider/service/service.rb
```diff
@@ -36,6 +36,9 @@ def ucommand(type, fof = true)
     else
       cmd = [send("#{type}cmd")].flatten
     end
+    if (( type =~ /^start/ ) && ( Facter.value(:osfamily) == "Solaris" ))
+      cmd = ["/usr/bin/ctrun -l none", cmd].flatten.join(" ")
+    end
```
joshcooper: It feels wrong to put this logic in the base `service` provider. Since we already have an `smf` service provider, how about if `smf` just overrides the `texecute` method:

```
def texecute(type, command, fof = true, squelch = false, combine = true)
  if type =~ /^start/
    command =  ["/usr/bin/ctrun -l none", command].flatten.join(" ")
  end
  super(type, command, fof, squelch, combine)
end
```

geoffnichols: @joshcooper Thanks, that approach makes more sense.

I'm not sure if I'm asking the right question, but will this work to override a method in a "grandparent" class? (The ```texecute``` method is in ```service```, which is the parent of ```base```, which is the parent of ```smf```.)

joshcooper: @geoffnichols yep, it should work fine

```
0 ~/work/puppet (master)  $ be irb
=> true
irb(main):002:0> resource = Puppet::Type.type(:service).new(:name => 'foo', :provider => :smf)
=> Service[foo]
irb(main):003:0> resource.provider.method(:texecute)
=> #<Method: Puppet::Type::Service::ProviderSmf(Puppet::Type::Service::ProviderService)#texecute>
```

geoffnichols: Updated PR with the ```texecute``` method override in ```init```. (Since management of Solaris init.d services on apparently requires explicitly declaring "provider=init", it looks like this needed to be done in ```init``` instead of ```smf```.)

lib/puppet/provider/service/init.rb
```diff
@@ -152,6 +152,15 @@ def restartcmd
...
 
+  if Facter.value(:osfamily) == "Solaris"
+    def texecute(type, command, fof = true, squelch = false, combine = true)
+      if type =~ /^start/
+        command =  ["/usr/bin/ctrun -l none", command].flatten.join(" ")
+      end
+      super(type, command, fof, squelch, combine)
+    end
+  end
+
```
joshcooper: I wouldn't conditionally override the `texecute` method as it leads to subtle issues, e.g. the call to `Facter.value` will execute on the master when it loads the provider (in the ruby `require` sense). It also makes it hard to test because you can't stub the behavior. It'd be better to always override the `texecute` method, and conditionally wrap the command if it's Solaris and type starts with `start`.

geoffnichols:
Some of the spec tests are failing here with "unexpected invocation: Facter.value(:osfamily)".

acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
geoffnichols: Agreed, a more universal daemon would be better. It looks like the few scripts that are in /etc/init.d in a default Solaris install are actually wrappers for SMF 'svcadm' commands. My first thought is to try to pick something from the OpenCSW packages that are installed by our full acceptance suite.

melissa:
@geoffnichols do you want me to bring this up in the community triage meeting?

acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
branan: Could we just have a fixture that runs a dummy service? If puppet is there we know we have ruby. A dumb daemon is only a couple lines of code...

lib/puppet/provider/service/init.rb
```diff
@@ -152,6 +152,15 @@ def restartcmd
     (@resource[:hasrestart] == :true) && [initscript, :restart]
   end
 
+  def texecute(type, command, fof = true, squelch = false, combine = true)
+    if type =~ /^start/
+      if Facter.value(:operatingsystem) == "Solaris"
+        command =  ["/usr/bin/ctrun -l none", command].flatten.join(" ")
```
branan: Are we sure this `flatten.join()` will play nice with all the possible inputs here?

branan:
I like this in theory. Simple change that fixes a real problem, has tests, etc. Just a couple comments above.

lib/puppet/provider/service/init.rb
joshcooper: I think this is good. We know we're calling flatten on a non-nil Array, and flatten produces a single array with at least one element (since we hardcoded `/usr/bin/ctrun ...`):

```
irb(main):009:0> [].join(' ')
=> ""
irb(main):010:0> ['foo'].join(' ')
=> "foo"
irb(main):011:0> ['foo', 'bar'].join(' ')
=> "foo bar"
```

I think the possible bad input would be if `command` included something that wasn't a String or Array, e.g.

```
irb(main):013:0> ['foo', { 1 => 2 } ].join(' ')
=> "foo {1=>2}"
```

But I'm pretty sure that wouldn't happen (since `command` gets passed to `Puppet::Util::Execution.execute` and that at least documents it as an array of strings https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/execution.rb#L112-L114

geoffnichols:
Updated the acceptance test to use a simple ruby daemon as the fixture service.

geoffnichols:
If the update to the acceptance test looks ok, I think next step here is to squash the commits and retarget against the puppet 3.x branch?

melissa:
@geoffnichols yeah, I think so. @joshcooper would you have time to make sure all your comments have been addressed? Would it be a good idea to get someone from QA to take a look at the acceptance tests you've added here? @johnduarte ?

lib/puppet/provider/service/init.rb
joshcooper: minor nit, we could collapse the `if's` instead of nesting:

```
if type =~ /^start/ && Facter.value(:operatingsystem) == "Solaris"
```

joshcooper: Now that I say that, I'm wondering why we don't have new service provider, whose parent is `:init` and is confined to operatingsystem `Solaris` and then the override logic can live there?

Also do we want to be using osfamily instead of operatingsystem?

acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  with_puppet_running_on master, master_opts, testdir do
+
+    agents.each do |agent|
+
+    fixture_service = 'sleepy_daemon'
+    fixture_service_stop = '/etc/init.d/sleepy_daemon stop'
+
+      arch = on(agent, facter('architecture')).stdout.chomp
```
johnduarte: Is this variable used?

acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
```diff
@@ -0,0 +1,144 @@
...
+      plat = on(agent, facter('osfamily')).stdout.chomp
+
+      skip_test unless plat == "Solaris"
```
johnduarte: nit pick. this value is available from beaker via `agent['platform']` so this can be abbreviated and save the facter call to

    skip_test unless agent['platform'] =~ /solaris/

or you can use the `confine` beaker helper

    confine :to, :platform => 'solaris'

acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
```diff
@@ -0,0 +1,144 @@
...
+      step "Setup fixture service on #{agent}"
+        sleepy_daemon_path = "/tmp/sleepy_daemon"
+        sleepy_daemon_initscript_path = "/etc/init.d/sleepy_daemon"
+        create_remote_file(agent, sleepy_daemon_path, sleepy_daemon_script)
+        create_remote_file(agent, sleepy_daemon_initscript_path, sleepy_daemon_initscript)
+        on(agent, "chmod +x #{sleepy_daemon_path} #{sleepy_daemon_initscript_path}")
+
+      step "Start the fixture service on #{agent} "
+        on(agent, "puppet resource service #{fixture_service} provider=init ensure=stopped")
+        on(agent, "puppet resource service #{fixture_service} provider=init ensure=running")
```
johnduarte: nit pick. it is recommended to use the `puppet` beaker method for making calls to puppet in case there are global settings that affect the desired behavior (e.g. future parser enabled).

    on(agent, puppet("resource service #{fixture_service} provider=init ensure=running"))

This should not be an issue with this code specifically. Just thought I would point it out.

acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
geoffnichols: Ah, thanks, good catch. It was used, but not any more. I'll update.

geoffnichols:
@melissa With your spec test patch, looks like this is now passing Travis when using 'Facter.value(:osfamily) == "Solaris"'. 

OK to squash these commits and retarget this against 3.x?

melissa:
@geoffnichols yeah, I think we're about ready for this. Just making sure, but the values I put in for `:osfamily` are completely arbitrary, and I have no idea if they're correct or not...

geoffnichols:
@melissa Based on my quick validation, those values appear to be correct.

geoffnichols:
Closing this PR. Opened https://github.com/puppetlabs/puppet/pull/3790 to retarget against 3.x.

PR #3661, opened 2015-02-27
(maint) Update manpages

joshcooper:
Update man pages to reflect reality.


haus:
@joshcooper you probably want to rm the entire man directory before generating them. some pages like https://github.com/joshcooper/puppet/blob/manpages/man/man8/puppet-kick.8 are still present, even though they probably shouldn't be.

haus:
:+1:

PR #3665, opened 2015-03-02
(DO NOT MERGE) (PUP-4063) Remove ext/build_defaults.yaml

Iristyle:
- Prior to the AIO pipeline, Windows MSI build dependencies were
   defined in build_defaults.yaml.  This included:

   - puppet_for_the_win repo that had MSI package details
   - puppet-win32-ruby that had vendored Ruby + gems
   - facter, cfacter, hiera, mcollective versions

   However, with the introduction of the AIO pipeline and Vanagon
   for building packages for all platforms, the above dependencies
   are now defined in the puppet-agent repository via JSON files.


kylog:
Can we put it on a barge, set it aflame, and let it float into the Arctic Ocean while we drink whiskey on the shore and tell tall tales about the great things it once did?

Iristyle:
I know the MSI bits won't be used any longer, but just want some verification from @haus that none of the info for the other platforms is needed either.

/cc @ferventcoder

ferventcoder:
Whoa... :eyes:

melissa:
:-1: I think we really should keep this file around in case someone from the community is using the packaging automation to build puppet

kylog:
If we remove this file, there are references in `ext/debian/rules`, `ext/redhat/puppet.spec.erb`, and `Rakefile` that should be cleaned up as well.

If we keep it but we aren't using it at all (and it would be nice to confirm that) should we add a big "no longer maintained, you're on your own" comment?

Iristyle:
@kylog - yeah, I assumed there would be other carnage related to this - hence the `DO NOT MERGE`.

FWIW - I think we should torch it if it's not actively being used by us, because it adds a source of confusion. It will remain alive and well on the stable branch.  If an external 3rd party wants to build packages with another toolchain that we no longer support on this branch, then let them maintain their own copy of that metadata.

If we are using it, then that's obviously a different story.

kylog:
Left to my own devices I'd be :+1: on removing this because I think it will bitrot. 

But I'll defer to @melissa and @haus. If they're :-1: let's just close this.

kylog:
Given #3802  and #3801, we're clearly planning to continue to maintain ext/build_defaults.yaml. We can always drop it at a later date.

PR #3672, opened 2015-03-03
(PUP-3706) Render console output using PSON

joshcooper:
Previously, rendering console output would fail if `datum` was true,
false, nil, or Object. This regressed in commit ac56d115ac9, which
switched from `json.render(datum)` to `JSON.pretty_generate(datum)`. It
turns out that the latter will render values that are not valid JSON,
e.g. `json.render(true) => 'true'`, while the latter raises an
exception.

This commit modifies the console formatter to use
`PSON.pretty_generate(datum)`. This preserves the old behavior of
accepting non-JSON values, while adding the pretty printing we need for
nested datum values.

We will be switching everything from PSON to JSON in PUP-3524

### DIFF
--- a/lib/puppet/network/formats.rb
+++ b/lib/puppet/network/formats.rb
@@ -151,7 +151,7 @@ Puppet::Network::FormatHandler.create(:console,
     end
 
     # ...or pretty-print the inspect outcome.
-    return JSON.pretty_generate(datum)
+    return PSON.pretty_generate(datum)
   end
 
   def render_multiple(data)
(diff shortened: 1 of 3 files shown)


joshcooper:
Verified the command that the acceptance test executes now passes:

```
$ be puppet certificate generate edm3pndp0rngyd6.delivery.puppetlabs.net --ca-location remote --server $(bundle exec facter fqdn)
true

kylog:
:+1: 

Waiting on travis, then will merge.

PR #3674, opened 2015-03-04
(PUP-3974) test source attribute, all checksums

er0ck:
Prior to this change the source_attributes test did not cover the puppet
apply cases for all checksums.  This change adds these tests.
This change also removes the sleep(1) in favor of a loop which checks to
ensure we're ready to move on, saving up to a second.

### DIFF
--- a/acceptance/tests/resource/file/source_attribute.rb
+++ b/acceptance/tests/resource/file/source_attribute.rb
@@ -136,7 +136,7 @@ apply_manifest_on(master, <<-MANIFEST, :catch_failures => true)
   }
 MANIFEST
 
-step "when using a puppet:/// URI with a master/agent setup"
+step "When using a puppet:/// URI with a master/agent setup"
 master_opts = {
   'main' => {
     'environmentpath' => "#{env_dir}",
@@ -152,11 +152,11 @@ with_puppet_running_on(master, master_opts, testdir) do
 
       checksums.each do |checksum_type|
         on agent, "cat #{file_to_check}#{checksum_type}" do
-          assert_match(/the content is present/, stdout, "Result file not created")
+          assert_match(/the content is present/, stdout, "Result file not created #{checksum_type}")
         end
 
         on agent, "cat #{dir_to_check}#{checksum_type}/source_dir_file" do
-          assert_match(/the content is present/, stdout, "Result file not created")
+          assert_match(/the content is present/, stdout, "Result file not created #{checksum_type}")
         end
       end
     end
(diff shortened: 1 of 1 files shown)


acceptance/tests/resource/file/source_attribute.rb
```diff
@@ -182,43 +189,51 @@ class source_test_module {
...
-  on agent, "mkdir -p #{File.dirname(source)}"
-  create_remote_file agent, source, 'Yay, this is the local file.'
+  checksums.each do |checksum_type|
+    target = "#{a_testdir}/target#{checksum_type}"
+    checksum = if checksum_type then "checksum => #{checksum_type}," else "" end
+
+    on agent, "mkdir -p #{File.dirname(source)}"
+    create_remote_file agent, source, 'Yay, this is the local file.'
+
+    step "Using a local file path (#{checksum})"
```
joshcooper: it was surprising to me that `checksum` included the trailing comma, which led to unintuitive log messages:

>  Using a local file path (checksum => ctime,)

Can we move the comma to the manifest where the `checksum` is used, or just remove entirely (since no properties/parameters come after `checksum` in the manifests below)?

acceptance/tests/resource/file/source_attribute.rb
```diff
@@ -182,43 +189,51 @@ class source_test_module {
...
+    manifest = %(file { '#{target}': source => '#{source}', ensure => present, #{checksum} })
+    apply_manifest_on agent, manifest
+    on agent, "cat #{target}" do
+      assert_match(/Yay, this is the local file./, stdout, "FIRST: File contents not matched on #{agent}")
```
joshcooper: The `lite` checksum types only compare the first 512 bytes of input, which is bigger than the file content we're managing. So in this test, `md5lite` is effectively the same as `md5`, etc. I think there is value in testing this combination, but I think we also want to test the case where the file content > 512. In that case, the `lite` checksums should only compare the first 512 bytes, ignoring whatever data comes after. You could even have the test modify `content[512]` and verify the `lite` checksums don't detect the change.

joshcooper:
This test passed for rhel7 and win2012r2-rubyx64, but it takes ~ 10 mins on Windows because it runs `puppet apply` twice for each checksum type, where each run takes ~ 16 seconds.

```
be rake ci:test:aio CONFIG=config/nodes/win2012r2-rubyx64.yaml  SHA=215e36f962dfd7080947d78a1a1340b0aadcfa62 TESTS=tests/resource/file/source_attribute.rb
...
seo2bu30c2m5h80.delivery.puppetlabs.net (agent-2012r2-x86_64-rubyx64) 00:55:22$ cmd.exe /c puppet apply --modulepath=C:/cygwin64/tmp/local_source_file_test.XP46nu -e "file { 'C:/cygwin64/tmp/local_source_file_test.XP46nu/targetmtime': source => 'puppet:///modules/source_mod/source', ensure => present, checksum => mtime, }"
...
seo2bu30c2m5h80.delivery.puppetlabs.net (agent-2012r2-x86_64-rubyx64) executed in 16.06 seconds
...
       Total Suite Time: 613.05 seconds
      Average Test Time: 613.05 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```

Can we rework this test to create a single manifest with a number of file resources, run `puppet apply` once to create the initial files, then run `puppet apply` again and verify no changes are necessary?

We have the same issue with the symbolic file mode test and I'd prefer not to add another 10mins to the test suite time.

er0ck:
good comments @joshcooper 
windows runtime down to ~100 seconds \o/
added the <checksum>lite tests beyond 512 bytes as well.  nice

PR #3675, opened 2015-03-04
WIP: (PUP-4080) accept s3 URLs as file source

pjfoley:
Hi Guys,

Hoping to get some feedback at an initial stab at building in support for s3 URLs as a source for the puppet file type resource.  I have modeled the code heavily on PR-3372 by @ffrank.

I am yet to add spec tests and will do some work on it.  Is this something that would likely be merged into Puppet?  I am new to ruby programming and originally tried to add the code to the puppetlabs-aws module (@garethr @Iristyle) however I could not work through the issues accessing internal puppet classes??

Thoughts, suggestions?

Even though it does not have any spec tests yet it is in a usable state.  To test:

1. Create an S3 bucket, file in the bucket and setup access permissions

(On your development machine)
2. Pull down the code
3. Install the aws-sdk-core ruby gem
4. Setup your s3 credentials (see the aws-sdk-core documentation)
5. Create a puppet manifest eg:

file { '/tmp/example.txt':
      source => 's3://foo-bucket/bar.file',
      region => 'ap-southeast-2',
}

6. bundle exec puppet apply manifest.pp


lib/puppet/file_serving/s3_metadata.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize(s3_response, path = '/dev/null')
+    super(path)
+
+    # ignore options that do not apply to s3 metadata
+    @owner = @group = @mode = nil
+
+    if s3_response.etag
+      @checksum_type = 'md5'
+      # Strip the double quotes from the S3 etag response
+      @checksum = "{md5}#{s3_response.etag.tr('"','')}"
```
ffrank: So does the `Etag` header as used by S3 consist of the actual MD5 hash of the file's content? Otherwise this won't work.

Perhaps @garethr or @Iristyle can elaborate?

pjfoley: In the majority of instances the etag will be the md5 hash of the file's content.

However there are certain instances where the etag will not be the md5 hash.  These instances are when:
1. multipart uploads;
2. files are over a certain size (5GB's?);
3. using customer-provided encryption keys during the upload process.

References:
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html
http://stackoverflow.com/questions/6591047/etag-definition-changed-in-amazon-s3
http://stackoverflow.com/questions/1775816/how-to-get-the-md5sum-of-a-file-on-amazons-s3

lib/puppet/type/file/content.rb
```diff
@@ -185,6 +185,8 @@ def each_chunk_from(source_or_content)
         yield source_or_content
       elsif content_is_really_a_checksum? && source_or_content.nil?
         yield read_file_from_filebucket
+      elsif source_or_content.respond_to?(:metadata) && source_or_content.metadata.respond_to?(:source) && source_or_content.metadata.source =~ /^s3?:/
```
pjfoley: Is there a better way to run this check?

My original check on the object was causing an exception due to not all source_or_content objects passed to the function having the property I needed to check.

lib/puppet/indirector/s3.rb
```diff
@@ -0,0 +1,47 @@
+require 'puppet/file_serving/terminus_helper'
+require 'puppet/indirector/terminus'
+require 'aws-sdk-core' if Puppet.features.aws?
```
pjfoley: Is there a better way to restrict the S3 file source than placing it here?

Looking at some other parts of the puppet code them seem to use a confine statement?

ffrank: I'm pretty sure that `confine` is specific to providers.
I rather like the approach of querying for a feature.

Iristyle:
Hi @pjfoley - thanks for the contribution!

I'm a little short on time at the moment, but I can tell you that is something that is unlikely to be merged into core Puppet.

As part of the 4 release, we have been trying to remove any non-essential types and providers, and place them into modules. For one, this makes it easier for us to ship updated modules easily out of band from releases of Puppet itself. It also simplifies or dependency trees and packaging efforts.

That said, the AWS module is *probably* where we would extend with something like this, though the implementation might get a little tricky.  I'll discuss with my colleagues some more when I get to the office.

pjfoley:
Hi @Iristyle thanks for the feedback, I originally was trying to work out how to add the capability to the puppetlabs-aws module so happy for it to go there (it just makes sense (tm)).

I understand that you are busy with the 4 release but hoping you might have a couple of minutes to provide some comments.  This will help me to work out how to try to build it into the puppetlabs-aws module.

Do you envisage it working in a similar way to how puppet manages hiera backends?

So the module would implement (similar to how file sources are managed):
* Metadata function
* Return the content function

How would the user signal to use a module provided source?  Something like how a user can choose a provider for the package type? ie something like the below:

file { '/tmp/example.txt':
      source   => 's3://foo-bucket/bar.file',
      provider => 'aws',
      region    => 'ap-southeast-2',
}

This would allow a puppet user to specify different providers per-file type.

ffrank:
FWIW, there was a lengthy discussion during the PR hangout. Sadly, the
recording at [YouTube](https://www.youtube.com/watch?v=toTsWoN4cI4)
doesn't seem to work.

pjfoley:
I was just reading through http://docs.puppetlabs.com/guides/custom_functions.html and I thought I would flag something for your discussion internally.

Whatever solution developed must *allow the agent* to be the machine that pulls file source.

This is to allow for the use of amazon IAM roles to protect access to the s3 bucket.  Ie only a certain machine running the puppet agent with the right credentials allocated can pull the file out of the s3 bucket.

I assume that both hiera and the package provider code only runs on the puppet master?

pjfoley:
Thanks for the heads up @ffrank I will keep an eye on the youtube link to watch it when it is posted.

ffrank:
On 03/09/2015 01:07 PM, pjfoley wrote:
> I assume that both hiera and the package provider code only runs on the
> puppet master?

The problem is that we have *no* file provider at this time. The type
code does everything. Both type and provider code runs on the agent side.

Parser functions will not come into play, I reckon.

Implementing such support in a module will make it necessary to ship
some type and/or provider code that ties in seamlessly with the file
resource type.

melissa:
@ffrank @pjfoley The link should be https://www.youtube.com/watch?v=IRwEVuX_7GY for the PR triage from 2015-03-4

ffrank:
Thanks @melissa I had found a wrong entry through youtube search, apparently.
Pertaining discussion is [towards the end](https://www.youtube.com/watch?v=IRwEVuX_7GY#t=95m05s).

pjfoley:
Hi Guys,

Just watching the latest triage video that Melissa posted.  Regarding the region extension to the puppet type (+ 30 minutes into the video) I was following the convention used in the puppetlabs-aws module (see https://github.com/puppetlabs/puppetlabs-aws#ec2_instance).  Happy for it to be built into the source property if that makes it easier.

ffrank:
Well, that discussion belongs on the mailing list, I believe, but FWIW, I think that the `file` resource will need a generic parameter akin to `package`'s `install_options` for passing information that pertains to certain files only (e.g., such that are to be retrieved from `s3`).

pjfoley:
That sounds ideal! Obviously each source provider may provide different options

pjfoley:
@branan just following up on your reference to your s3file module and how via a defined type it fetches files from s3 (+ 30 minutes in the triage video).  When originally trying to work out how to pull protected files out of s3 I looked at your module.  

One thing that was not clear to me (I admit I could not work out how to do it with your module) was accessing files from secured s3 buckets (ie non-public).

By using the Amazon supported gem (aws-sdk-core) it automatically sources credentials available to it (see http://docs.aws.amazon.com/sdkforruby/api/#Credentials).

For my example in response to @ffrank I want to use the instance credential allocated against a IAM role for the ec2 instance (unique for each ec2 instance) which the aws-sdk-core gem will automatically source from the metadata service.

Gemfile
```diff
@@ -27,6 +27,7 @@ gem "puppet", :path => File.dirname(__FILE__), :require => false
 gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 2.0', '< 4'])
 gem "hiera", *location_for(ENV['HIERA_LOCATION'] || ['> 1.0', '< 3'])
 gem "rake", "10.1.1", :require => false
+gem "aws-sdk-core", "2.0.5"
```
pjfoley: Just following up a comment from the triage video on the 4th of March.  I didn't mean to imply that the aws-sdk-core gem should be built into the standard puppet suite.  I only added it here to signal that it was required to have my code work.

I agree it needs to be moved into a development group.

In my view those users of puppet who wish to use the amazon cloud should install the gem if required.

branan:
@pjfoley I didn't realize pulling content from protected buckets was a requirement in your design, so yeah - using the gem is better in that case.

branan:
Given your PR against puppetlabs-aws, should we close this one?

pjfoley:
@branan in my opinion no.  Unfortunately the PR against puppetlabs-aws is majorly inefficient.  It does not check the metadata to see if it needs to download the content of a file.

For my use case it would be managable (a couple of Kb per private key) but for those who use it to download larger files it would soon add up in costs.

Also as noted in the PR message, I am not sure if it is by design, a bug or a limitation in the puppet language.  If in the file resource I set "replace => false" it will still call the function and download the content.

 solution I have

Gemfile
MikaelSmith: That can be indicated by using `:require => false`.

lib/puppet/feature/aws.rb
```diff
@@ -0,0 +1,3 @@
+require 'puppet/util/feature'
+
+Puppet.features.add(:aws, libs: 'aws-sdk-core')
```
MikaelSmith: This will likely break anything pre Ruby-2.0, please use syntax found in the other feature files: `:libs => ['aws-sdk-core']`

branan:
I'm going to close this out. I'm leaving the ticket open, and have also filed [PUP-4368](https://tickets.puppetlabs.com/browse/PUP-4368) to track adding the appropriate extension points so that this can be implemented in a module.

PR #3681, opened 2015-03-04
(PUP-4076) Remove 'allow_virtual' deprecation warning in the 3.x series

whopper:
In Puppet 4, the package type's allow_virtual setting will be true by
default. In commit 3f6d7cad, this change was made and the deprecation
warning for allow_virtual was removed in master. However, many have
found the deprecation warning to be disruptive in the 3.x series, so
this commit removes it. We will deprecate in docs instead.

### DIFF
--- a/lib/puppet/type/package.rb
+++ b/lib/puppet/type/package.rb
@@ -438,11 +438,8 @@ module Puppet
     newparam(:allow_virtual, :boolean => true, :parent => Puppet::Parameter::Boolean, :required_features => :virtual_packages) do
       desc 'Specifies if virtual package names are allowed for install and uninstall.'
 
-      # In a future release, this should be defaulted to true and the below deprecation warning removed
-      defaultto do
-        Puppet.deprecation_warning('The package type\'s allow_virtual parameter will be changing its default value from false to true in a future release. If you do not want to allow virtual packages, please explicitly set allow_virtual to false.') unless value
-        false
-      end
+      # In a future release, this should be defaulted to true
+      defaultto false
     end
 
     autorequire(:file) do
--- a/spec/unit/type/package_spec.rb
+++ b/spec/unit/type/package_spec.rb
@@ -118,26 +118,6 @@ describe Puppet::Type.type(:package) do
         Puppet::Type.type(:package).new(:name => ["error"])
       end.to raise_error(Puppet::ResourceError, /Name must be a String/)
     end
-
-    it "should issue deprecation warning for default allow_virtual for a provider that supports virtual packages" do
-      Puppet.expects(:deprecation_warning).with('The package type\'s allow_virtual parameter will be changing its default value from false to true in a future release. If you do not want to allow virtual packages, please explicitly set allow_virtual to false.')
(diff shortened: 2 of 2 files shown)


lib/puppet/type/package.rb
```diff
@@ -437,12 +437,6 @@ def change_to_s(currentvalue, newvalue)
 
     newparam(:allow_virtual, :boolean => true, :parent => Puppet::Parameter::Boolean, :required_features => :virtual_packages) do
       desc 'Specifies if virtual package names are allowed for install and uninstall.'
-
-      # In a future release, this should be defaulted to true and the below deprecation warning removed
-      defaultto do
-        Puppet.deprecation_warning('The package type\'s allow_virtual parameter will be changing its default value from false to true in a future release. If you do not want to allow virtual packages, please explicitly set allow_virtual to false.') unless value
-        false
```
joshcooper: I think we want to keep the default to `false`

peterhuene: The point of the deprecation warning was that it would eventually default to true so that rpm providers like yum can (in theory) behave like they do on the command line with respect to virtual packages.  If it defaults to false here, we added this (annoying) deprecation for no reason.  Removing the warning is definitely the correct thing to do, but I could go either way with respect to the default.

kylog: For stable (3.x) I think we want to retain the default value of false here. The reason for the default of false was that we saw the `allow_virtual=true` behavior as making more sense long-term, but as a breaking change, so we were planning to not change behavior in the 3.x series. I think that logic still holds. But I may be missing sth since it's been a few months.

Assuming we do keep this as `false`, we do need to take care on the merge-up to master that we keep the default at `true` on master.

whopper:
@joshcooper @kylog @peterhuene cool, I went ahead and put the default back to false

lib/puppet/type/package.rb
peterhuene: Whoops, didn't see this going into stable :/.  Yeah makes sense, sorry for the unnecessary comment.

branan:
Can we instead make it easier to squelch the warning via config of some sort? These warnings about compatibility changes are important, and will be useful when people start actually caring about porting their code.

Something like Python 2's `-3` option.

Deprecating in the docs is good (again see Python as a surprisingly good example here), but when our tools can know and alert users about compatibility we should let them opt-in (or out) easily to that information.

ahpook:
@branan  we *have* a warning squelch, but it's hard to find, defaults to off, and turns off a whole category of errors rather than just this one. In _this specific case_ the error message is so spammy and the effects of the deprecation so narrow it makes more sense just to release-note it.

branan:
@ahpook Word. With that context I totally agree. That was missing from the ticket.

branan:
:+1: given the history from ahpook

kylog:
For additional context, there are some modules on the forge that include custom facts that require puppet and invoke the package providers to report custom facts. So squelching gets extra-gymnastic in that case.

This message has been a lot of fun, and I'm pretty sure I was the one that suggested it in the first place, so I feel sad inside.

PR #3682, opened 2015-03-04
(PUP-4082) Make unfold of undef produce nothing

hlindberg:
Before this commit, an unfold of undef would be treated like unfold of
any other value - i.e. wrapping it in an array, and then depending on
context unfold the result into that context.

It is more natural that an unfold of undef leads to nothing - i.e that
undef is turned into an empty array, that when unfolded in context adds
nothing.

Thus *undef and [] has the same result.
In options for case and selector expressions, if a *undef is used as an
option, that option will not match any value, not even undef or an empty
array.


hlindberg:
Travis hiccup - restarted job

PR #3683, opened 2015-03-04
(PUP-4083) Add sample environment.conf to conf/

nfagerlund:
The file is all comments, so we can include it in a default production
environment. It includes a link to docs, a brief explanation, and a list of
allowed settings with default values.


haus:
:+1:

kylog:
I started to write a 'what about per-environment parser settings' comment, but then remembered that's 3.x only.

So :+1:

PR #3684, opened 2015-03-04
(PUP-1253) Add systemd service masking support

trevor-vaughan:
This updates the systemd service type to add support for masking. If a
service is masked, it is deemed to also be disabled. If a service is
masked and changed to enabled, it will first be unmasked since the
standard 'systemctl enable' command does not properly unmask the command
first.

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -1,4 +1,4 @@
-# Manage systemd services using /bin/systemctl
+# Manage systemd services using systemctl
 
 Puppet::Type.type(:service).provide :systemd, :parent => :base do
   desc "Manages `systemd` services using `systemctl`."
@@ -12,7 +12,7 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
   def self.instances
     i = []
     output = systemctl('list-unit-files', '--type', 'service', '--full', '--all',  '--no-pager')
-    output.scan(/^(\S+)\s+(disabled|enabled)\s*$/i).each do |m|
+    output.scan(/^(\S+)\s+(disabled|enabled|masked)\s*$/i).each do |m|
       i << new(:name => m[0])
     end
     return i
@@ -28,12 +28,35 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
 
   def enabled?
     begin
-      systemctl("is-enabled", @resource[:name])
+      systemctl_info = systemctl(
+         'show',
+         @resource[:name],
+         '--property', 'LoadState',
+         '--property', 'UnitFileState',
+         '--no-pager'
+      )
+
+      svc_info = Hash.new
+      systemctl_info.split.each do |svc|
+        entry_pair = svc.split('=')
+        svc_info[entry_pair.first.to_sym] = entry_pair.last
+      end
+
+      # The masked state is equivalent to the disabled state in terms of
+      # comparison so we only care to check if it is masked if we want to keep
+      # it masked.
+      #
+      # We only return :mask if we're trying to mask the service. This prevents
+      # flapping when simply trying to disable a masked service.
+      return :mask if (@resource[:enable] == :mask) && (svc_info[:LoadState] == 'masked')
+      return :true if svc_info[:UnitFileState] == 'enabled'
     rescue Puppet::ExecutionFailure
-      return :false
+      # The execution of the systemd command can fail for quite a few reasons.
+      # In all of these cases, the failure of the query indicates that the
+      # service is disabled and therefore we simply return :false.
     end
 
-    :true
+    return :false
   end
 
   def status
@@ -46,9 +69,20 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
   end
 
   def enable
+    output = systemctl("unmask", @resource[:name])
     output = systemctl("enable", @resource[:name])
   rescue Puppet::ExecutionFailure
     raise Puppet::Error, "Could not enable #{self.name}: #{output}", $!.backtrace
+  end
+
+  def mask
+    self.disable
+
+    begin
+      output = systemctl("mask", @resource[:name])
+    rescue Puppet::ExecutionFailure
+      raise Puppet::Error, "Could not mask #{self.name}: #{output}", $!.backtrace
+    end
   end
 
   def restartcmd
--- a/lib/puppet/type/service.rb
+++ b/lib/puppet/type/service.rb
@@ -42,6 +42,9 @@ module Puppet
 
     feature :flaggable, "The provider can pass flags to the service."
 
+    feature :maskable, "The provider can 'mask' the service.",
+      :methods => [:mask]
+
     newproperty(:enable, :required_features => :enableable) do
       desc "Whether a service should be enabled to start at boot.
         This property behaves quite differently depending on the platform;
@@ -58,6 +61,12 @@ module Puppet
 
       newvalue(:manual, :event => :service_manual_start) do
         provider.manual_start
+      end
+
+      # This only makes sense on systemd systems. Otherwise, it just defaults
+      # to disable.
+      newvalue(:mask, :event => :service_disabled, :required_features => :maskable) do
+        provider.mask
       end
 
       def retrieve
--- a/spec/unit/provider/service/systemd_spec.rb
+++ b/spec/unit/provider/service/systemd_spec.rb
@@ -108,20 +108,57 @@ describe Puppet::Type.type(:service).provider(:systemd) do
   describe "#enabled?" do
     it "should return :true if the service is enabled" do
       provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service'))
-      provider.expects(:systemctl).with('is-enabled', 'sshd.service').returns 'enabled'
+      provider.expects(:systemctl).with(
+        'show',
+        'sshd.service',
+        '--property', 'LoadState',
+        '--property', 'UnitFileState',
+        '--no-pager'
+      ).returns "LoadState=loaded\nUnitFileState=enabled\n"
       expect(provider.enabled?).to eq(:true)
     end
 
     it "should return :false if the service is disabled" do
       provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service'))
-      provider.expects(:systemctl).with('is-enabled', 'sshd.service').raises Puppet::ExecutionFailure, "Execution of '/bin/systemctl is-enabled sshd.service' returned 1: disabled"
+      provider.expects(:systemctl).with(
+        'show',
+        'sshd.service',
+        '--property', 'LoadState',
+        '--property', 'UnitFileState',
+        '--no-pager'
+      ).returns "LoadState=loaded\nUnitFileState=disabled\n"
       expect(provider.enabled?).to eq(:false)
+    end
+
+    it "should return :false if the service is masked and the resource is attempting to be disabled" do
+      provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service', :enable => false))
+      provider.expects(:systemctl).with(
+        'show',
+        'sshd.service',
+        '--property', 'LoadState',
+        '--property', 'UnitFileState',
+        '--no-pager'
+      ).returns "LoadState=masked\nUnitFileState=\n"
+      expect(provider.enabled?).to eq(:false)
+    end
+
+    it "should return :mask if the service is masked and the resource is attempting to be masked" do
+      provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service', :enable => 'mask'))
+      provider.expects(:systemctl).with(
+        'show',
+        'sshd.service',
+        '--property', 'LoadState',
+        '--property', 'UnitFileState',
+        '--no-pager'
+      ).returns "LoadState=masked\nUnitFileState=\n"
+      expect(provider.enabled?).to eq(:mask)
     end
   end
 
   describe "#enable" do
     it "should run systemctl enable to enable a service" do
       provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service'))
+      provider.expects(:systemctl).with('unmask', 'sshd.service')
       provider.expects(:systemctl).with('enable', 'sshd.service')
       provider.enable
     end
@@ -132,6 +169,18 @@ describe Puppet::Type.type(:service).provider(:systemd) do
       provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service'))
       provider.expects(:systemctl).with(:disable, 'sshd.service')
       provider.disable
+    end
+  end
+  
+  describe "#mask" do
+    it "should run systemctl to disable and mask a service" do
+      provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service'))
+      # :disable is the only call in the provider that uses a symbol instead of
+      # a string.
+      # This should be made consistent in the future and all tests updated.
+      provider.expects(:systemctl).with(:disable, 'sshd.service')
+      provider.expects(:systemctl).with('mask', 'sshd.service')
+      provider.mask
     end
   end
 
--- a/spec/unit/type/service_spec.rb
+++ b/spec/unit/type/service_spec.rb
@@ -63,6 +63,11 @@ describe Puppet::Type.type(:service), "when validating attribute values" do
       expect(srv.should(:enable)).to eq(:false)
     end
 
+    it "should support :mask as a value" do
+      srv = Puppet::Type.type(:service).new(:name => "yay", :enable => :mask)
+      expect(srv.should(:enable)).to eq(:mask)
+    end
+
     it "should support :manual as a value on Windows" do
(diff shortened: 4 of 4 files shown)


peiriannydd:
I can change this to a feature, that's easy enough and appropriate.

feature :maskable, :service_mask, something else?

kylog:
Ooh naming ... I'd go with 'maskable' because the rest of the service features are 'xxxable's.

/cc @joshcooper both for the naming and because I'm not sure if I completely captured his concerns.

kylog:
Ping @joshcooper

lib/puppet/type/service.rb
```diff
@@ -60,6 +63,16 @@ module Puppet
         provider.manual_start
       end
 
+      # This only makes sense on systemd systems. Otherwise, it just defaults
+      # to disable.
+      newvalue(:mask, :event => :service_disabled) do
+        if provider.declared_feature?(:maskable)
```
joshcooper: Puppet will automagically create predicate methods based on provider features declared in the type. So you can just do:

```
irb(main):002:0> require 'puppet'
=> true
irb(main):003:0> resource = Puppet::Type.type(:service).new(:name => 'foo')
=> ...
irb(main):004:0> resource.provider.maskable?
=> false
```

lib/puppet/type/service.rb
```diff
@@ -60,6 +63,16 @@ module Puppet
         provider.manual_start
       end
 
+      # This only makes sense on systemd systems. Otherwise, it just defaults
+      # to disable.
+      newvalue(:mask, :event => :service_disabled) do
```
joshcooper: Sorry about the merge, revert, and new PR mess, that was my fault.

In the previous PR, we were talking about whether it makes sense to write `ensure => mask` in cases where the provider isn't maskable. Should puppet reject that as an error or interpret that to mean `disable`? The former is easier for manifest authors, because you don't have to know *a priori* if the service provider is maskable or not. The latter providers cleaner semantics about the user's intent -- if you requested `mask` but the system doesn't support it, we fail.

I don't have a good sense for whether the latter case is important. Perhaps maskable and disabled are so similar the distinction isn't meaningful?

For comparison, puppet's package type confines the `ensure` property based on `purgeable` and `holdable`: features: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/package.rb#L89-L95.

/cc @nfagerlund @kylog @ScottGarman @adrienthebo

kylog:
My $.02: it is surprising to allow `mask` on systems that don't have the `maskable` feature.

nfagerlund:
Mine: this is conceptually identical to ensure => purged, and we should treat them identically. I can see an argument for going either route, but from an explainability standpoint, I want it to be the _same_ route. Inconsistency harms user brains, and there needs to be a compelling reason to introduce it.

peiriannydd:
Well, as surprising as it may be, here are your options:

Allowing Mask:

    service { 'foo': enable => 'mask' }

Masks the service on systems that support mask, disables it otherwise because that's the obvious fallback (and documented!).

Not-Allowing Mask:

    service { 'foo': enable => '.......Wait, no

    if $::operatingsystem in ['RedHat','CentOS'] and $::operatingsystemversion in....Wait, no

    <create custom fact, has_systemd>

    if $has_systemd {
      service { 'foo': enable => 'mask' }
    }
    else {
      service { 'foo': enable => false }
    }

I know where my vote lies.....

lib/puppet/type/service.rb
peiriannydd: Sorry that I missed this last time, could have sworn that I pushed it in.

peiriannydd:
Ok, so now it's documented! I updated the documentation header to describe what happens when you use mask on systems without service masking.

Hopefully I didn't butcher Nick's expectations for formatting too much :-D

kylog:
But my guess is that 99+% of sites won't need that special-case logic at all - they'll just keep saying `ensure => stopped` and it will behave as expected on systemd/upstart/etc.

And the rare site that *wants* masked behavior on systemd systems (but has a mix and wants the same code to work "the same" on all) might do something like the above. Optimizing for that corner case isn't making sense to me atm (and as Nick points out isn't consistent with how we've treated other provider-specific features).

I'll have to sleep on this :)

adrienthebo:
:-1: to treat `mask` equivalent to `disabled` where not supported; if we do that we're lying to the user and silently doing the wrong thing. If given the option between failing hard and silently doing the wrong thing, I've found you'll always regret going with the latter.

peiriannydd:
Well, of course I woke up and had some thoughts.

1) My implementation of ensure => 'mask' is incorrect! Thinking about this, the type should *not* be making implementation decisions for the provider. Should we have another provider that implements :maskable, my current implementation may not be what it should do. I will be changing this to be more appropriate.

2) @kylog and @adrienthebo are correct. Allowing :mask to happen on systems that do not support masking does not accurately reflect the state of the system.

3) I should change 'mask' to 'masked' since 'mask' does not fit with 'enabled|disabled'.

PR update coming soon.

peiriannydd:
As a side note, can we get FACT-853 pushed through since it will be necessary for figuring out whether or not we have systemd running on a system? I suppose that a more accurate fact would be has_maskable_services, but that's a bit overkill.

peiriannydd:
PR Updated with the adjustments mentioned above.

melissa:
@kylog or @joshcooper have you had a chance to look this over?

branan:
I'm :+1: on this in its current state, but I'd like one of the earlier reviewers to weight in

kylog:
I haven't yet, but I would like to take another pass through before merging.

lib/puppet/provider/service/systemd.rb
```diff
@@ -28,12 +28,33 @@ def disable
...
+      # comparison so we only care to check if it is masked if we want to keep
+      # it masked.
+      #
+      # We only return :mask if we're trying to mask the service. This prevents
+      # flapping when simply trying to disable a masked service.
+      return :mask if (@resource[:enable] == :mask) && (svc_info[:LoadState] == 'masked')
+      return :true if svc_info[:UnitFileState] == 'enabled'
     rescue Puppet::ExecutionFailure
-      return :false
+      # Don't worry about this failing, just return :false if it does.
```
adrienthebo: What does a failure indicate in this case?

lib/puppet/provider/service/systemd.rb
```diff
@@ -28,12 +28,33 @@ def disable
...
+        svc_info[entry_pair.first.to_sym] = entry_pair.last
+      }
+
+      # The masked state is equivalent to the disabled state in terms of
+      # comparison so we only care to check if it is masked if we want to keep
+      # it masked.
+      #
+      # We only return :mask if we're trying to mask the service. This prevents
+      # flapping when simply trying to disable a masked service.
+      return :mask if (@resource[:enable] == :mask) && (svc_info[:LoadState] == 'masked')
```
adrienthebo: In general methods suffixed with `?` return booleans or boolean values; this return means that `#enabled?` returns three values which is unexpected and reminiscent of a [daily wtf article](http://thedailywtf.com/articles/What_Is_Truth_0x3f_). Does calling code know to expect three possible values?

lib/puppet/provider/service/systemd.rb
```diff
@@ -46,11 +67,22 @@ def status
   end
 
   def enable
+    output = systemctl("unmask", @resource[:name])
```
adrienthebo: We discussed whether `enable => true` should unmask and enable a service; this will hide the fact that a service was masked and may make it easier to unintentionally unmask a service. Is this acceptable?

spec/unit/type/service_spec.rb
```diff
@@ -63,6 +63,16 @@
       expect(srv.should(:enable)).to eq(:false)
     end
 
+    it "should support :mask as a value" do
```
adrienthebo: How do these two tests differ?

lib/puppet/provider/service/systemd.rb
```diff
@@ -28,12 +28,33 @@ def disable
...
+      systemctl_info = systemctl(
+         'show',
+         @resource[:name],
+         '--property', 'LoadState',
+         '--property', 'UnitFileState',
+         '--no-pager'
+      )
+
+      svc_info = Hash.new
+      systemctl_info.split.map{|svc|
```
adrienthebo: `#map` is used to transform one enumerable object to another, but in this case we're ignoring the return value and modifying `svc_info` instead. Would `#each` be more appropriate?

lib/puppet/provider/service/systemd.rb
peiriannydd: There are several failure cases that could happen when calling the systemctl command. This ignores those cases since they are not actually relevant to the state of the resource.

adrienthebo: Ah, okay. It might be more helpful to indicate something along those lines in that the failure to query the status of the service can be treated like the service is disabled - that would reveal the intent more clearly.

peiriannydd: Ok, easy enough.

lib/puppet/provider/service/systemd.rb
peiriannydd: Sure, I can change that. Probably a holdover from an earlier implementation.

lib/puppet/provider/service/systemd.rb
peiriannydd: In the case of Puppet enabled? returns a symbol and, in this particular case, it is *already* not a boolean.

Valid values are :true, :false, :manual, :mask (new)

If :true then true, else false. In this case :mask is equivalent to :false since it is not :true.

Speaking of unexpected :true != true and :false != false has caused no end of confusion with writing custom types.

Honestly, to do anything else would make the maintainability of the code MUCH more difficult.

lib/puppet/provider/service/systemd.rb
peiriannydd: Leaving a service masked and trying to start it is not a logical path. Either you want it masked and stopped or started. Masked and started is not a valid case.

So, we can either throw an error and ask the user what they actually want to do, or we can just unmask it. Either is fine with me.

spec/unit/type/service_spec.rb
peiriannydd: The most definitely don't. I'll kill the dup

lib/puppet/provider/service/systemd.rb
adrienthebo: I don't know if one way or the other is more correct; throwing an error in this case might not provide a way to unmask a masked service, so this might be the only viable approach as-is. I think this approach will do.

lib/puppet/provider/service/systemd.rb
adrienthebo: Fair enough, let's not make more work for ourselves. Might be good to add method documentation explicitly calling this out.

lib/puppet/provider/service/systemd.rb
kylog: I'm +1 on this approach. You asked for it to be enabled, puppet makes it so regardless of what state it was in previously.

spec/unit/provider/service/systemd_spec.rb
```diff
@@ -134,6 +171,17 @@
       provider.disable
     end
   end
+  
+  describe "#mask" do
+    it "should run systemctl to disable and mask a service" do
+      provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service'))
+      # I think that this is a symbol method due to the direct call from inside
+      # the provider. I'm not 100% sure.
+      provider.expects(:systemctl).with(:disable, 'sshd.service')
```
kylog: Hmm, me either. Why is `:disable` different?

adrienthebo: We invoke `systemctl` with `"unmask"` as a string when unmasking (https://github.com/puppetlabs/puppet/pull/3684/files#diff-dafbe8211057bc10ff2cfe788a6c4ebbR70) but invoke `systemctl` with `:disable` as a symbol when disabling (https://github.com/puppetlabs/puppet/pull/3684/files#diff-dafbe8211057bc10ff2cfe788a6c4ebbR24); I believe that the underlying execution code converts all args to strings when executing. The end result is the same, but the code varies - nothing inherently harmful, just superficially inconsistent.

peiriannydd: :disable is an outlier in the provider. I'm not changing it due to potential other ramifications that have nothing to do with my patch.

lib/puppet/provider/service/systemd.rb
peiriannydd: Went to document this and it's pretty clear from the 'newvalue' statements what it can return so didn't add anything extra.

spec/unit/provider/service/systemd_spec.rb
kylog: Yep that makes sense, and yeah unrelated to this PR except that you could lose/tweak the comment. 

Meanwhile, a future cleanup opportunity :)

peiriannydd: Comment updated to explain what's going on.

melissa:
@adrienthebo @kylog it looks like all your comments have been addressed on this PR. If you could take a look to give either a final thumbs up or additional comments, that would be great!

adrienthebo:
I've run out of things to complain about, :+1:

melissa:
@kylog it looks like everything is good to go on this PR. We'll give you a few more days to approve or deny this before we merge it.

melissa:
@kylog you mentioned you wanted to take another look at this

peterhuene:
@kylog Any update on this PR?  Sorry for the repeated pings.

kylog:
Damn. My apologies for becoming such a bottleneck on this PR.

:+1:

peterhuene:
@kylog Are we good to merge this now or are we waiting for post 4.1?

peterhuene:
@trevor-vaughan One last thing before we can merge this: could you amend the commit messages to be prefixed with "(PUP-XXX)" for consistency with our contribution guidelines.  The current commit messages are missing the parenthesis (it's a bit pedantic, I know, but consistency is a good thing when viewing git logs).  Thanks!

trevor-vaughan:
Weird, I thought I did this. I'll check again.

On Thu, Apr 23, 2015 at 2:55 PM, Peter Huene <notifications@github.com>
wrote:

> @trevor-vaughan <https://github.com/trevor-vaughan> One last thing before
> we can merge this: could you amend the commit messages to be prefixed with
> "(PUP-XXX)" for consistency with our contribution guidelines. The current
> commit messages are missing the parenthesis (it's a bit pedantic, I know,
> but consistency is a good thing when viewing git logs). Thanks!
>
> —
> <https://github.com/puppetlabs/puppet/pull/3684#issuecomment-95684881>.
>



-- 
Trevor Vaughan
Vice President, Onyx Point, Inc
(410) 541-6699
tvaughan@onyxpoint.com

-- This account not approved for unencrypted proprietary information --

trevor-vaughan:
Rebased and done.

On Thu, Apr 23, 2015 at 2:58 PM, Trevor Vaughan <tvaughan@onyxpoint.com>
wrote:

> Weird, I thought I did this. I'll check again.
>
> On Thu, Apr 23, 2015 at 2:55 PM, Peter Huene <notifications@github.com>
> wrote:
>
>> @trevor-vaughan <https://github.com/trevor-vaughan> One last thing
>> before we can merge this: could you amend the commit messages to be
>> prefixed with "(PUP-XXX)" for consistency with our contribution guidelines.
>> The current commit messages are missing the parenthesis (it's a bit
>> pedantic, I know, but consistency is a good thing when viewing git logs).
>> Thanks!
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/puppetlabs/puppet/pull/3684#issuecomment-95684881>.
>>
>
>
>
> --
> Trevor Vaughan
> Vice President, Onyx Point, Inc
> (410) 541-6699
> tvaughan@onyxpoint.com
>
> -- This account not approved for unencrypted proprietary information --
>



-- 
Trevor Vaughan
Vice President, Onyx Point, Inc
(410) 541-6699
tvaughan@onyxpoint.com

-- This account not approved for unencrypted proprietary information --

kylog:
Thanks @trevor-vaughan! This will be released in puppet 4.1.

lib/puppet/provider/service/systemd.rb
```diff
@@ -12,7 +12,7 @@
   def self.instances
     i = []
     output = systemctl('list-unit-files', '--type', 'service', '--full', '--all',  '--no-pager')
-    output.scan(/^(\S+)\s+(disabled|enabled)\s*$/i).each do |m|
+    output.scan(/^(\S+)\s+(disabled|enabled|masked)\s*$/i).each do |m|
```
kurtwall: The list to scan should include `manual` per the docs in `puppet describe service`:
```
$ puppet describe service
...
- **enable**
    Whether a service should be enabled to start at boot.
    This property behaves quite differently depending on the platform;
    wherever possible, it relies on local tools to enable or disable
    a given service.
    Valid values are `true`, `false`, `manual`, `mask`.
    Requires features enableable.
...
```

PR #3688, opened 2015-03-05
(DO NOT MERGE)(PUP-4088) Fix new path acceptance for some Windows

MikaelSmith:
Without this change, the new ensure_puppet-agent_paths acceptance test
fails on Windows Server 2003, and when testing 32-bit puppet-agent
installs on 64-bit Windows.

Update the test to use the correct paths for both cases.

### DIFF
--- a/acceptance/tests/ensure_puppet-agent_paths.rb
+++ b/acceptance/tests/ensure_puppet-agent_paths.rb
@@ -75,9 +75,9 @@ end
 
 step 'test configprint outputs'
 agents.each do |agent|
-  config_options(agent[:platform]).each do |config_option|
-    on(agent, puppet_agent('--configprint ' "#{config_option[:name]}")) do
-      assert_match(config_option[:expected], stdout)
+  on(agent, puppet_agent('--configprint all')) do
+    config_options(agent[:platform]).each do |config_option|
+      assert_match("#{config_option[:name]} = #{config_option[:expected]}", stdout)
     end
   end
 end
--- a/acceptance/tests/ensure_puppet-agent_paths.rb
+++ b/acceptance/tests/ensure_puppet-agent_paths.rb
@@ -6,53 +6,78 @@ test_name 'PUP-4033: Ensure aio path spec is honored'
 require 'puppet/acceptance/temp_file_utils'
 extend Puppet::Acceptance::TempFileUtils
 
-config_options = [
-  # code
-  {:name => :codedir, :posix_expected => '/etc/puppetlabs/code', :win_expected => 'C:/ProgramData/PuppetLabs/code', :installed => :dir},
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Running through acceptance tests for this.

acceptance/tests/ensure_puppet-agent_paths.rb
```diff
@@ -90,17 +114,37 @@
...
+    return '/usr/bin'
+  end
+
+  case platform
+  when /windows/
+    # If undefined, ruby_arch defaults to x86
+    if ruby_arch == 'x64'
+      ruby_arch = /-64/
+    else
+      ruby_arch = /-32/
```
joshcooper: does `-` need to be escaped?

MikaelSmith: Quite probably... I'm surprised that hasn't caused a test issue.

MikaelSmith: It's valid, quick test: `'windows-2003r2-64' =~ /-2003/`.

It's only a keyword in a square-bracket expression, such as /[2-3]/.

MikaelSmith:
Tested with MSI from SHA 215e36f962dfd7080947d78a1a1340b0aadcfa62, using x86 build on Windows Server 2003 (x86/x64), and x86 and x64 builds on Windows Server 2012 x64.

Also tested with git on all those platforms; a minor fix is needed to run with git that looks unrelated, so I'll post it as a separate PR.

MikaelSmith:
This should be ready to go, passed all my testing.

acceptance/tests/ensure_puppet-agent_paths.rb
```diff
@@ -6,53 +6,78 @@
...
-]
+def config_options(platform)
+  case platform
+  when /windows/
+    if platform =~ /2003/
+      common_app_data = 'C:/Documents and Settings/All Users/Application Data'
+    else
+      common_app_data = 'C:/ProgramData'
+    end
+    puppet_install = "#{common_app_data}/PuppetLabs"
```
Iristyle: Minor, but this really `puppet_data`, not `puppet_install`

acceptance/tests/ensure_puppet-agent_paths.rb
Iristyle: Right, `-` should only be meaningful in character classes...

Iristyle:
Now that I'm not on the VPN (over a DSL connection), I'm going to run this through acceptance locally to make sure we're good.

Shouldn't take too long to verify / merge.

Iristyle:
Note that the test is running very slowly querying for settings individually.  @MikaelSmith is changing to `puppet agent --configprint all`

MikaelSmith:
Pushed a change to do configprint all.

PR #3690, opened 2015-03-06
(PUP-4086) Patch ParseErrors with location if missing

hlindberg:
Before this commit, ParseErrors that occured in a call from the
evaluator was passed on verbatim to the higher powers. The intent is
that ParseError is used when the file and line in a .pp file is known.
The use of raised ParseErrors without such information does however
permeate throughput the puppet code base and in modules.

This was bad because many errors would then not be associated with
a location in a manifest.

Further complications were that a) some errors were already preformatted
with location information being baked into the message text when
reaching the evaluator (these should pass verbatim), and b) the way
safeevaluate of AST objects work meant that all errors would end up with
a wrapped 'original' exception which leads to duplicated output of the
message and the text 'Wrapped Exception'.

This commit fixes these problems by:

* Introducing a specialized PreformattedError which is a specialized
variant of ParseError (that enables the evaluator to no yet again add
location information). A choice was made to derive it from ParseError to
ensure that all that rescues ParseError would still recognize it.
* Making the evaluator do more extensive introspection and rewrite of
the different kinds of errors that occur while evaluating one
instruction.
* Making the Errors.adderrorcontext method strip away the original error
since it has already effectively made it superflous by poking its
backtrace into its parent error.

Summary - solution means:
* no superflous Wrapped Exception output
* location information for ParseErrors "in the wild" that do not provide
this.

A possible downside is the dropped original error may cause loss of
information in corner cases (non covered by spec tests at least) if
someone explicitly depends on the original error being kept after a
safeevaluate, or a call path that passes through the evaluator. (No such
cases where found in the puppet code base though).

### DIFF
--- a/lib/puppet/pops/evaluator/evaluator_impl.rb
+++ b/lib/puppet/pops/evaluator/evaluator_impl.rb
@@ -74,14 +74,43 @@ class Puppet::Pops::Evaluator::EvaluatorImpl
       @@eval_visitor.visit_this_1(self, target, scope)
 
     rescue Puppet::Pops::SemanticError => e
-      # a raised issue may not know the semantic target
+      # A raised issue may not know the semantic target, use errors call stack, but fill in the 
+      # rest from a supplied semantic object, or the target instruction if there is not semantic
+      # object.
+      #
       fail(e.issue, e.semantic || target, e.options, e)
 
-    rescue StandardError => e
-      if e.is_a? Puppet::ParseError
-        # ParseError's are supposed to be fully configured with location information
+    rescue Puppet::PreformattedError => e
+      # Already formatted with location information, and with the wanted call stack.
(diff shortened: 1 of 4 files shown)


lib/puppet/pops/evaluator/evaluator_impl.rb
```diff
@@ -74,14 +74,43 @@ def evaluate(target, scope)
...
+      # Accept a ParseError that has file or line information available
+      # as an error that should be used verbatim. (Tests typically run without
+      # setting a file name).
+      # ParseError can supply an original - it is impossible to determine which
+      # call stack that should be propagated, using the ParseError's backtrace.
+      #
+      if e.file || e.line
         raise e
+      else
+        # Since it had no location information, threat it as user intended a general purpose
```
hkenney: Tiny typo here (threat -> treat), not sure if it's worth fixing

PR #3518, opened 2015-01-24
(PUP-3908) Tidy type is too verbose when working of directories containing lot of files

amuraru:

PR #3693, opened 2015-03-09
(PUP-3908) Make tidy type messages less verbose

op-ct:
This patch reduces the amount of notifications tidy generates when working
with a large number of files.  Tidy now reports the number of changed files
using 'notify,' and reports the list of files (the old behavior) with 'debug.'

### DIFF
--- a/lib/puppet/type/tidy.rb
+++ b/lib/puppet/type/tidy.rb
@@ -253,7 +253,9 @@ Puppet::Type.newtype(:tidy) do
     else
       files = [self[:path]]
     end
-    result = files.find_all { |path| tidy?(path) }.collect { |path| mkfile(path) }.each { |file| notice "Tidying #{file.ref}" }.sort { |a,b| b[:path] <=> a[:path] }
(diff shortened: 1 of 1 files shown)


op-ct:
This incorporates improvements suggested at #3518.

kylog:
:+1: 

Let's merge after code unfreeze.

bittner:
Tidy now always reports what is has done. When no change happens (i.e. "Tidying 0 files") I expect that nothing is reported. Kind of, because there is nothing to do.

It should be a bit like `unless (no files files to remove)` or `onlyif (files to remove)`. Otherwise I have no chance to get zero noise in our Puppet logs when nothing changes in a Puppet run.

Is there a reason why this hasn't been implemented like this? Is an additional PR welcome?

joshcooper:
I think it was just an oversight. I agree puppet should only emit a notice if there are 1 or more files to tidy. PRs are definitely welcome. Please see the CONTRIBUTORS.md for more details.

PR #3694, opened 2015-03-09
(PUP-4131) Add insync? to the gem provider

raphink:


lib/puppet/provider/package/gem.rb
```diff
@@ -108,6 +108,15 @@ def install(useversion = true)
     self.fail "Could not install: #{output.chomp}" if output.include?("ERROR")
   end
 
+  # Allow Gem version expressions in ensure
+  def insync?(is)
+    begin
+      Gem::Dependency.new('', resource[:ensure]).match?('', is)
+    rescue
+      false
```
raphink: Note: this is necessary in case `is` is not a valid version number (e.g. "absent").

raphink:
@jeffmccune @kylog what do you think of this?

lib/puppet/provider/package/gem.rb
branan: Looking at this, will it work correctly with "present" or "absent" in the ensure? That would cause the `Gem::Dependency.new` to throw an exception, which would always be false... but I also know that ensure absest/present have some weird special code paths.

Basically I have no idea what would happen here in that case. I'm going to play with this code before next week's PR triage meeting so we can decide what to do with this.

branan: It looks like this behaves correctly with ensure absent/present.

branan:
Now that I've finally had a chance to test this and  confirm it behaves correctly, I'm pretty happy with it.

I would like to see tests added before we bring this in, though

lib/puppet/provider/package/gem.rb
raphink: This method is not a standard location for the ensure parameter, so it is not called by default. It is only called by `insync?` in the type when all other options have failed. This is why this works with present/absent.

raphink: Note: this code won't be called in puppet 3, it only works with puppet 4.

branan: That makes sense

branan:
It looks like the only other provider that specifies its own `insync?` is the `pkg` provider. There may be some clues for testing your implementation over there.

raphink:
I didn't really see a lot of existing tests for this provider iirc.

raphink:
Also, I don't really view this as a new feature, rather as a bugfix. As a matter of fact, the syntax works without this patch (even in puppet 3), but the change is not idempotent, so this patch only makes sure that things like `ensure => "~> 1.2" ` are idempotent.

raphink:
@branan any news about this?

peterhuene:
Hi @raphink.  @branan is out this week and I was supposed to look into adding specs for this PR.  Unfortunately I haven't had the time in the past few days, but I'll be looking into it later this afternoon.

raphink:
OK thanks.

peterhuene:
@branan How did you verify this?  For me, `is` is an array (see: `gemsplit` method) and the call to `Gem::Dependency#match?` always throws an `ArgumentError` with 'Malformed version number string ["xxxx"]`, thus never "in sync" when using a version specifier.  I tested this with something like:

`package { 'yard': ensure => '>= 0.8.7.6' }`

With version 0.8.7.6 installed, we're still reinstalling the gem with every invocation. By testing each element of `is` in turn, we get the correct behavior.

peterhuene:
Also, I'd like to see us catch `Gem::Requirement::BadRequirementError` to guard against malformed requirement strings, rather than catching all exceptions (which inadvertently hide exceptions like the above).

branan:
@peterhuene I just created a simple manifest that depended on a `>= foo.bar` gem version. If that's not working now it's possibly related to different behavior in different ruby/gem versions, which would be truly awesome and not super surprising.

raphink:
I also tested like Branan and it worked, so indeed it must be a version
issue...

peterhuene:
@branan @raphink What versions of ruby and rubygems are you using to verify, so I can take a look at the discrepency?  I don't see any indication in the rubygems source that arrays were ever supported for `match?`.

peterhuene:
@raphink @branan I'm still unable to verify the fix as described in this PR on different versions of ruby and rubygems.

For example, with the haml gem already installed (4.0.6):

    $ puppet apply -d -e "package { 'haml': ensure => '>= 4', provider => gem }"
    ...
    Debug: Executing: '/usr/bin/gem list --local'
    Debug: Executing: '/usr/bin/gem install -v >= 4 --no-rdoc --no-ri haml'
    Notice: /Stage[main]/Main/Package[haml]/ensure: ensure changed '["4.0.6"]' to '>= 4'
    ...

Note the attempt to reinstall the gem.  This is the same behavior as without the patch applied. 

With my fix that respecst `is` as an array instead:

    $ puppet apply -d -e "package { 'haml': ensure => '>= 4', provider => gem }"
    ...
    Debug: Executing: '/usr/bin/gem list --local'
    ...

`Gem::Dependency.new` will not take an array for a version specifier, so I still fail to see how the original fix could work.

If either of you could give me the exact steps you did to verify the fix as originally implemented, that would be great.  Otherwise, I think we should either amend this fix or close it in favor of the other one.

peterhuene:
@raphink @branan Last ping before I close this PR in favor of one that contains my fix above :)

raphink:
Which one is the other PR?

peterhuene:
I haven't put it up yet, but it would contain the proposed fix described here: https://github.com/puppetlabs/puppet/pull/3694#issuecomment-98827319 (along with some specs).

peterhuene:
Hi @raphink.  As discussed, I've opened #3959 with the proposed fix above (plus some specs).  I'm going to close this PR in favor of that one.  If you believe this is in error, please reopen this PR and we can discuss.  Thanks!

raphink:
Cool, looking forward to the feature

PR #3695, opened 2015-03-10
(PUP-4056) Use privatebindir to resolve ruby

joshcooper:
* Prepends privatebindir to the PATH environment used to execute `ruby` and `gem` on all hosts.
* For ruby-based ENC scripts executing on the master, resolve `ruby` relative to `privatebindir`.
* Moves the beaker pin to the version that contains the AIO pathing changes
* Removes the AIO path overrides from node config files

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -12,7 +12,7 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem "beaker", *location_for(ENV['BEAKER_VERSION'] || 'git://github.com/puppetlabs/beaker#31bc7aa4405ac4fa0a2896be1c644dee34a0ea3e')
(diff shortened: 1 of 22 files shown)


joshcooper:
```
$ be rake ci:test:aio SHA=58fbddcf1973262126a3f6699fdb929e4e78a6a5 CONFIG=config/nodes/rhel7.yaml TESTS=tests
...
       Total Suite Time: 7076.47 seconds
      Average Test Time: 26.70 seconds
              Attempted: 265
                 Passed: 212
                 Failed: 0
                Errored: 0
                Skipped: 52
                Pending: 1
                  Total: 265
```

joshcooper:
This should be merged after PR #3698

kylog:
:+1: 

Not following why the PR ordering matters though. Maybe it will all make sense with coffee :)

joshcooper:
> Maybe it will all make sense with coffee :)

Hmm I think maybe I need more coffee... I was thinking moving the beaker pin had to be the last thing, but I don't think that's strictly true.

PR #3696, opened 2015-03-10
(PUP-3997) Run puppet apply to reset logdir owner

johnduarte:
This commit adds a teardown step in the
tests/loader/func4x_loadable_from_modules.rb acceptance test to
run puppet apply without a puppet.conf specified in order to
restore the ownership of the logdir.

Prior to this commit, the test left ownership of the logdir set to
the user and group specified in the test, which may not be the
proper standard ownership.

### DIFF
--- a/acceptance/tests/loader/func4x_loadable_from_modules.rb
+++ b/acceptance/tests/loader/func4x_loadable_from_modules.rb
@@ -80,4 +80,9 @@ SOURCE
     assert_match(/^Generated, 1 => 10, 2 => 20, 3 => 30$/, stdout, "Generated the wrong content")
   end
 
+  teardown do
+    # Run apply against standard puppet.conf to restore ownership of logdir
+    on agent, puppet('apply', '-e', "'notify {\"hello\":}'")
+  end
+
 end


acceptance/tests/loader/func4x_loadable_from_modules.rb
```diff
@@ -80,4 +80,9 @@ class helloworld {
     assert_match(/^Generated, 1 => 10, 2 => 20, 3 => 30$/, stdout, "Generated the wrong content")
   end
 
+  teardown do
```
joshcooper: If the test raises an exception after running `puppet apply` but before we get to the teardown, then the permissions won't be reset correctly. We could register the teardown earlier, but I don't like letting the test "corrupt" the permissions and then trying to patch them up. I'd rather redo e6e0ae02dcc41746d8196b4a6e0e11248af51006 so the workaround is specific to PE:

```
puppetconf = get_test_file_path(agent, 'puppet.conf')
if agent.is_pe?
  user = agent.puppet['user']
  group = agent.puppet['group']
  on agent, puppet("config", "set", "user", user, "--section", "main", "--config", puppetconf)
  on agent, puppet("config", "set", "group", group, "--section", "main", "--config", puppetconf)
end
```

We could also just revert e6e0ae02dcc41746d8196b4a6e0e11248af51006 since PE4 will be using AIO agents and there won't be a `puppet` vs `pe-puppet` issue.

kylog: I'd favor reverting e6e0ae0. It *seems* like that should just work, but I might be missing something.

joshcooper: I reverted the commit in PR #3698, added in the other puppet user & group commits, and acceptance passed.

PR #3697, opened 2015-03-10
(PUP-4132) Pass merge strategy to Hiera

thallgren:
(PUP-4132) Pass merge strategy to Hiera

Hiera version 2.x will recognize a resolution_type parameter value
that is a Hash to mean a `:hash` _resolution_type_ with configured deep
merge options. This commit ensures that the lookup _merge_ parameter
is correctly converted into such a Hash when needed. The conversion
also translates the `'unique'` merge type into the Hiera `:array`
equivalence.

### DIFF
--- a/lib/puppet/indirector/hiera.rb
+++ b/lib/puppet/indirector/hiera.rb
@@ -17,7 +17,8 @@ class Puppet::Indirector::Hiera < Puppet::Indirector::Terminus
 
   def find(request)
     not_found = Object.new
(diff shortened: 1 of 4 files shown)


hlindberg:
@ripienaar any comments?

hlindberg:
Looks good to me - will wait to see if R.I has any thoughts.

ripienaar:
seems ok to me too

hlindberg:
Thanks @ripienaar for the extra set of eyes on this feature.

PR #3699, opened 2015-03-10
(PUP-3931) Change systemd KillMode to process for agent

shanemadden:
Currently, the entire control group is sent a kill; if a puppet agent run under the control group is running an upgrade of the puppet package, both the running agent process and the yum process doing the upgrade are killed as well as the daemon agent process, leaving the puppet installation in a bad state.  This brings the behavior in line with the behavior under the SysV init script.

### DIFF
--- a/ext/systemd/puppet.service
+++ b/ext/systemd/puppet.service
@@ -7,6 +7,7 @@ After=basic.target network.target puppetmaster.service
 EnvironmentFile=-/etc/sysconfig/puppetagent
 EnvironmentFile=-/etc/sysconfig/puppet
 ExecStart=/opt/puppetlabs/puppet/bin/puppet agent $PUPPET_EXTRA_OPTS --no-daemonize
+KillMode=process
 
 [Install]
 WantedBy=multi-user.target


kylog:
@haus ping.

haus:
:+1: /cc @shrug

kylog:
@shanemadden btw I'd target puppet#stable because that's where the next 3.x release will come from, whereas puppet#master will be 4.x and will be released with different (all-in-one) packaging.

shanemadden:
@kylog Good call, I'll close this one out and make a new PR against stable.

shanemadden:
https://github.com/puppetlabs/puppet/pull/3700

shanemadden:
@kylog Actually, should I leave this one open as well?  There will probably be some heavy refactoring of the package related stuff for all-in-one but the service unit stuff will probably be re-used, right?  Might as well have the fix present in master as well?

haus:
@shanemadden yes, the service file is being re-used, so this should stay open/get merged into master as well. good catch!

kylog:
@shanemadden we'll just pick this up when we merge stable -> master after #3700 is merged, so closing this one.

Thanks for the fix!!

PR #3700, opened 2015-03-10
(PUP-3931) Change systemd KillMode to process for agent

shanemadden:
Currently, the entire control group is sent a kill; if a puppet agent run under the control group is running an upgrade of the puppet package, both the running agent process and the yum process doing the upgrade are killed as well as the daemon agent process, leaving the puppet installation in a bad state.  This brings the behavior in line with the behavior under the SysV init script.

### DIFF
--- a/ext/systemd/puppet.service
+++ b/ext/systemd/puppet.service
@@ -7,6 +7,7 @@ After=basic.target network.target puppetmaster.service
 EnvironmentFile=-/etc/sysconfig/puppetagent
 EnvironmentFile=-/etc/sysconfig/puppet
 ExecStart=/usr/bin/puppet agent ${PUPPET_EXTRA_OPTS} --no-daemonize
+KillMode=process
 
 [Install]
 WantedBy=multi-user.target


shanemadden:
See https://github.com/puppetlabs/puppet/pull/3699 - re-targeted to stable.

haus:
:+1:

shrug:
:+1:

melissa:
@kylog since this is slated for the 3.7.5 release, are we ready to merge this into stable?

opoplawski:
The problem with this now is that a simple systemctl restart puppet fails because the child process refuses to die:
~~~~~
# ps -fe | grep puppet
root     21421     1  0 16:45 ?        00:00:00 /bin/sh /usr/bin/start-puppet-agent agent  --no-daemonize
root     21424 21421  0 16:45 ?        00:00:01 /usr/bin/ruby-mri /usr/bin/puppet agent --no-daemonize
root     22536  2094  0 16:51 pts/4    00:00:00 grep --color=auto puppet
# systemctl restart puppet
# ps -fe | grep puppet
root     21424     1  0 16:45 ?        00:00:01 /usr/bin/ruby-mri /usr/bin/puppet agent --no-daemonize
~~~~~
At least for me on Fedora 21.

PR #3701, opened 2015-03-11
(PUP-4081) lookup in module when looking up qualified names

thallgren:
This commit changes the data provider logic so that when a qualified
name is used in the lookup and no value is found globally or in the
environment, then the first segment of the qualified name will be used
to determine what module the lookup should be perfomed in.

No attempt will be made to do a module lookup unless the name is
qualified. This applies even if the lookup is made from within a
module.

This commit also adds a validator for the function provided module
data hash that checks that all keys are qualified with the name of
the module.

### DIFF
--- a/lib/puppet/data_providers.rb
+++ b/lib/puppet/data_providers.rb
@@ -17,12 +17,15 @@ module Puppet::DataProviders
   MODULE_NAME = 'module_name'.freeze
 
   def self.lookup_in_module(name, scope, merge)
-    # Do not attempt to do a lookup in a module if evaluated code is not in a module
-    # which is detected by checking if "MODULE_NAME" exists in scope
-    return nil unless scope.exist?(MODULE_NAME)
+    # Do not attempt to do a lookup in a module unless the name is qualified.
+    qual_index = name.index('::')
+    throw :no_such_key if qual_index.nil?
+    module_name = name[0..qual_index-1]
 
     assert_loaded()
     adapter = Puppet::DataProviders::DataAdapter.adapt(Puppet.lookup(:current_environment))
-    adapter.module_provider(scope[MODULE_NAME]).lookup(name, scope, merge)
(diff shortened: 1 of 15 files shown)


lib/puppet/data_providers.rb
```diff
@@ -17,12 +17,15 @@ def self.lookup_in_environment(name, scope, merge)
...
+    throw :no_such_key if qual_index.nil?
+    module_name = name[0..qual_index-1]
 
     assert_loaded()
     adapter = Puppet::DataProviders::DataAdapter.adapt(Puppet.lookup(:current_environment))
-    adapter.module_provider(scope[MODULE_NAME]).lookup(name, scope, merge)
+    data_provider = adapter.module_provider(module_name, true)
+    throw :no_such_key if data_provider.nil?
+    data_provider.lookup(name, scope, merge)
   end
```
hlindberg: Don't think the above should fail when doing a merge; you may ask for merges of keys and search continues into a module for which there may not be a data provider, or no key bound.

thallgren: Not sure what you mean. Throwing a :no_such_key exception doesn't imply failure. It just meanst that no value was found for the given key. The alternative is to say that the key was found and its value is `nil` which is incorrect IMO.

hlindberg: yes, got it...

lib/puppet/data_providers/data_adapter.rb
```diff
@@ -28,15 +28,20 @@ def env_provider
...
+    end
+    @data[module_name]
   end
 
   def self.create_adapter(environment)
     new(environment)
   end
 
-  def initialize_module_provider(module_name)
+  def initialize_module_provider(module_name, nil_if_no_service)
```
hlindberg: nil_if_no_service is not needed - there is a 'none' provider for both modules and environments

PR #3704, opened 2015-03-11
(PUP-400) Remove systemcomponent check from package

cyberious:


spec/unit/provider/package/windows/exe_package_spec.rb
```diff
@@ -34,7 +34,6 @@
     {
       'DisplayName'      => ['My App', ''],
       'UninstallString'  => ['E:\uninstall.exe', ''],
-      'SystemComponent'  => [nil, 1],
       'WindowsInstaller' => [nil, 1],
       'ParentKeyName'    => [nil, 'Uber Product'],
       'Security Update'  => [nil, 'KB890830'],
```
cyberious: Should we also remove the check if security update as some might want to do this as well.

joshcooper: I'm not sure about Security Updates or Hotfixes for that matter, e.g. a user might try to `ensure => absent` and I don't know that that would work in all cases.

joshcooper:
@cyberious can you change the commit summary to `(PUP-400) Remove systemcomponent ..`. We have automation in place to reconcile git commits with jira tickets and the subject line needs to fit the pattern. Otherwise LGTM.

spec/unit/provider/package/windows/exe_package_spec.rb
joshcooper: I'm actually on the fence about this.. I can see reasons for wanting puppet to report on all installed packages, e.g. `puppet resource package` even if it's not something that puppet can uninstall.

melissa:
@Iristyle @ferventcoder since this is a windows change, would you mind taking a look at this pull request?

ferventcoder:
@melissa I'm +1 with the same comment that @joshcooper has - the commit message needs fixed up.

ferventcoder:
@cyberious can you address the issues with the commit message so we can move forward with this one? 

Please read over https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md#making-changes (Making Changes) - see the commit message information. Make sure the first line of the commit message is `(issue#) summary <= 50 char`

cyberious:
@ferventcoder @joshcooper updated commit message

ferventcoder:
@cyberious And then in the commit body, please explain what the change does, why it is necessary, etc.

cyberious:
@ferventcoder added

ferventcoder:
Thanks. We can get a bit dogmatic about the commit messages, but we have good reasons for doing so e.g. maintainability.

ferventcoder:
@cyberious we are in a holding pattern waiting for a green status on CI before I'll merge this in. I want to know if we've resolved @joshcooper 's concerns as well.

spec/unit/provider/package/windows/exe_package_spec.rb
ferventcoder: Any decision here? I think I'm also on the fence about not seeing security updates and hotfixes by default. 

It's too bad `package` doesn't have `query_options`

cyberious: I see this as a separate provider perhaps for package. Thoughts?

ferventcoder: Are you talking about security updates / hotfixes?

cyberious: Yes, a package {... provider => 'windows_updates' } for example

joshcooper:
@ferventcoder CI is green, free to merge.

Changes look good to me, :+1:

PR #3706, opened 2015-03-11
(PUP-4190) Don't display credentials in output

hunner:
The url contains the credentials for a device. Don't print them in plaintext.

### DIFF
--- a/lib/puppet/application/device.rb
+++ b/lib/puppet/application/device.rb
@@ -169,7 +169,11 @@ Licensed under the Apache 2.0 License
     end
     devices.each_value do |device|
       begin
-        Puppet.info "starting applying configuration to #{device.name} at #{device.url}"
+        device_url = URI.parse(device.url)
+        # Handle nil scheme & port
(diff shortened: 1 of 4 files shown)


lib/puppet/application/device.rb
```diff
@@ -169,7 +169,7 @@ def main
     end
     devices.each_value do |device|
       begin
-        Puppet.info "starting applying configuration to #{device.name} at #{device.url}"
+        Puppet.info "starting applying configuration to #{device.name}"
```
joshcooper: If you care about losing the debug context and `device.uri` is a ruby URI, you could use `#scheme`, `#host`, etc to extract out the public bits. You could also create a new URI, and maybe clear the user part with `#userinfo=`?

branan:
Looks sane to me :+1:

branan:
Actually, wait. Is it possible for device URLs to have path components, not just protocol/host/port? If so we might want to include those in the output too.

hunner:
Updated for path.

PR #3707, opened 2015-03-12
(PUP-4133) Fix interpolation for variables with leading underscore

hkenney:
(PUP-4133) Fix interpolation for variables with leading underscore

Prior to this commit, if string interpolation was done on a variable
that had a leading underscore in the name (i.e. ${bar::_foo}), puppet
would fail with an "illegal fully qualified name" error from
the lexer.

In order to ensure that such variable names can be interpolated
properly, update the lexer so that it will create WORD tokens for
variable names with leading underscores, rather than exiting with
an "illegal fully qualified name" error.


lib/puppet/pops/parser/egrammar.ra
```diff
@@ -581,6 +581,7 @@ node_definition_expression
 #   TODO: The error that "class" is not a valid classname is bad - classname rule is also used for other things
 classname
   : NAME
+  | WORD
```
hlindberg: The grammar was changed, but not regenerated. Did you forget to check it in? (Alternatively, if you did not regenerate the parser, this change was not needed since it only takes effect after the parser has been regenerated).

hlindberg:
Looks good, will merge when stable is again open for merges.

PR #3708, opened 2015-03-12
(PUP-4191,PUP-4203) Use execute function to run gem uninstall

queeno:
When attempting to uninstall a gem using a custom gem provider inheriting gem.rb, the wrong gemcmd is issued.

For example, the [sensu-puppet](https://github.com/sensu/sensu-puppet) module introduces embedded ruby gem support through a custom [sensu_gem](https://github.com/sensu/sensu-puppet/blob/master/lib/puppet/provider/package/sensu_gem.rb) provider, inheriting puppet's `gem.rb`.

In sensu-gem, the `:gemcmd` is declared to match the embedded sensu gem path:
`commands :gemcmd => "/opt/sensu/embedded/bin/gem"`

When the following resource is run:
```
package { 'sensu-plugin':
  ensure   => absent,
  provider => 'sensu_gem',
}
```

`/usr/bin/gem uninstall -x -a sensu-plugin` is executed rather than
`/opt/sensu/embedded/bin/gem uninstall -x -a sensu-plugin`.

I noticed that in `gem.rb` the install and self.gemlist methods build a command array and pass it to the execute function. This results in using the right gemcmd.

This PR makes the uninstall method behave just like the self.gemlist and install methods.
An array of arguments is built and passed to the execute function, which runs the right command and returns the output. Since some versions of gem may return exit code 0 on error, a hard failure is triggered if the output contains the word 'ERROR'. This is compliant with the `install` method behaviour.

### DIFF
--- a/lib/puppet/provider/package/gem.rb
+++ b/lib/puppet/provider/package/gem.rb
@@ -122,7 +122,12 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
   end
 
   def uninstall
-    gemcmd "uninstall", "-x", "-a", resource[:name]
+    command = [command(:gemcmd), "uninstall"]
+    command << "-x" << "-a" << resource[:name]
+    output = execute(command)
+    
+    # Apparently some stupid gem versions don't exit non-0 on failure
+    self.fail "Could not uninstall: #{output.chomp}" if output.include?("ERROR")
   end
 
   def update
--- a/lib/puppet/provider/package/gem.rb
+++ b/lib/puppet/provider/package/gem.rb
@@ -8,11 +8,12 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
     interpreted as the path to a local gem file.  If source is not present at all,
     the gem will be installed from the default gem repositories.
 
-    This provider supports the `install_options` attribute, which allows command-line flags to be passed to the gem command.
+    This provider supports the `install_options` and `uninstall_options` attributes,
+    which allow command-line flags to be passed to the gem command.
     These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
     or an array where each element is either a string or a hash."
 
-  has_feature :versionable, :install_options
+  has_feature :versionable, :install_options, :uninstall_options
 
   commands :gemcmd => "gem"
 
@@ -124,6 +125,9 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
   def uninstall
     command = [command(:gemcmd), "uninstall"]
     command << "--executables" << "--all" << resource[:name]
+
+    command += uninstall_options if resource[:uninstall_options]
+ 
     output = execute(command)
     
     # Apparently some stupid gem versions don't exit non-0 on failure
@@ -136,5 +140,9 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
 
   def install_options
     join_options(resource[:install_options])
+  end
+
+  def uninstall_options
+    join_options(resource[:uninstall_options])
   end
 end
--- a/lib/puppet/provider/package/gem.rb
+++ b/lib/puppet/provider/package/gem.rb
@@ -123,7 +123,7 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
 
   def uninstall
     command = [command(:gemcmd), "uninstall"]
-    command << "-x" << "-a" << resource[:name]
+    command << "--executables" << "--all" << resource[:name]
     output = execute(command)
     
     # Apparently some stupid gem versions don't exit non-0 on failure
--- a/spec/unit/provider/package/gem_spec.rb
+++ b/spec/unit/provider/package/gem_spec.rb
@@ -187,3 +187,52 @@ context 'installing myresource' do
     end
   end
 end
+
+context 'uninstalling myresource' do
+  describe provider_class do
+    let(:resource) do
+      Puppet::Type.type(:package).new(
+        :name     => 'myresource',
+        :ensure   => :absent
+      )
+    end
+
+    let(:provider) do
+      provider = provider_class.new
+      provider.resource = resource
+      provider
+    end
+
+    before :each do
+      resource.provider = provider
+    end
+
+    describe "when uninstalling" do
+      it "should use the path to the gem" do
+        provider_class.stubs(:command).with(:gemcmd).returns "/my/gem"
+        provider.expects(:execute).with { |args| args[0] == "/my/gem" }.returns ""
+        provider.uninstall
+      end
+
+      it "should specify that the gem is being uninstalled" do
+        provider.expects(:execute).with { |args| args[1] == "uninstall" }.returns ""
+        provider.uninstall
+      end
+
+      it "should specify that the relevant executables should be removed without confirmation" do
+        provider.expects(:execute).with { |args| args[2] == "--executables" }.returns ""
+        provider.uninstall
+      end
+
+      it "should specify that all the matching versions should be removed" do
+        provider.expects(:execute).with { |args| args[3] == "--all" }.returns ""
+        provider.uninstall
+      end
+
+      it "should specify the package name" do
+        provider.expects(:execute).with { |args| args[4] == "myresource" }.returns ""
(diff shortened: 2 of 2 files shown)


queeno:
@puppetcla CLA signed.

lib/puppet/provider/package/gem.rb
```diff
@@ -122,7 +122,12 @@ def query
   end
 
   def uninstall
-    gemcmd "uninstall", "-x", "-a", resource[:name]
+    command = [command(:gemcmd), "uninstall"]
+    command << "-x" << "-a" << resource[:name]
```
melissa: Could you expand these flags to make the command more clear? (`--executables` and `--all`)

melissa:
@queeno thank you for your contribution!

Could you add tests to verify this behavior?

melissa:
I'm also not sure if this is the place to bring this up, but this might be a good opportunity to add in `uninstall_options` to this provider. Something along the lines of
```diff
diff --git a/lib/puppet/provider/package/gem.rb b/lib/puppet/provider/package/gem.rb
index 301cfd2..ebb88f0 100644
--- a/lib/puppet/provider/package/gem.rb
+++ b/lib/puppet/provider/package/gem.rb
@@ -8,11 +8,12 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
     interpreted as the path to a local gem file.  If source is not present at all,
     the gem will be installed from the default gem repositories.

-    This provider supports the `install_options` attribute, which allows command-line flags to be passed to the gem command.
+    This provider supports the `install_options` and `uninstall_options` attribute, which
+    allows command-line flags to be passed to the gem command.
     These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
     or an array where each element is either a string or a hash."

-  has_feature :versionable, :install_options
+  has_feature :versionable, :install_options, :uninstall_options

   commands :gemcmd => "gem"

@@ -124,8 +125,10 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
   def uninstall
     command = [command(:gemcmd), "uninstall"]
     command << "-x" << "-a" << resource[:name]
+
+    command += uninstall_options if resource[:uninstall_options]
     output = execute(command)
-
+
     # Apparently some stupid gem versions don't exit non-0 on failure
     self.fail "Could not uninstall: #{output.chomp}" if output.include?("ERROR")
   end
@@ -137,4 +140,8 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
   def install_options
     join_options(resource[:install_options])
   end
+
+  def uninstall_options
+    join_options(resource[:uninstall_options])
+  end
 end
```

queeno:
@melissa No problem Melissa, it's a pleasure contributing to such an amazing product.

I think all your suggestions make sense, will work on them tomorrow morning :smile:

lib/puppet/provider/package/gem.rb
```diff
@@ -122,7 +122,12 @@ def query
   end
 
   def uninstall
-    gemcmd "uninstall", "-x", "-a", resource[:name]
+    command = [command(:gemcmd), "uninstall"]
```
joshcooper: It is unfortunate that you have to do this... When you declare a command in your provider, puppet automatically creates a method, e.g. `gemcmd`, which ultimately calls into `Puppet::Util::Execution.execute` to execute the command. Unfortunately, it appears `sensu_gem` is able to override the value of `command(:gem)`, but not the path used in `gemcmd`. So it's surprising when you use one vs the other:

```
0 ~/work/puppet (master)  $ bundle exec irb
...
irb(main):012:0> require 'puppet'
=> true
irb(main):013:0> Puppet::Type.type(:package).provide :sensu_gem, :parent => :gem do
irb(main):014:1*   desc "Sensu Embedded Ruby Gem support. If a URL is passed via `source`, then
irb(main):015:1"     that URL is used as the remote gem repository; if a source is present but is
irb(main):016:1"     not a valid URL, it will be interpreted as the path to a local gem file.  If
irb(main):017:1"     source is not present at all, the gem will be installed from the default gem
irb(main):018:1"     repositories."
irb(main):019:1>
irb(main):020:1*   has_feature :versionable, :install_options
irb(main):021:1>
irb(main):022:1*   commands :gemcmd => "/opt/sensu/embedded/bin/gem"
irb(main):023:1> end
=> Puppet::Type::Package::ProviderSensu_gem
...
irb(main):028:0> provider = Puppet::Type.type(:package).provider(:sensu_gem)
=> Puppet::Type::Package::ProviderSensu_gem
irb(main):029:0> provider.command(:gemcmd)
=> nil
```

This is because I don't have `/opt/sensu/embedded/bin/gem`, so this is returning the overridden value as expected. But

```
irb(main):030:0> provider.gemcmd
Puppet::ExecutionFailure: Execution of '/Users/josh/.rbenv/versions/2.1.5/bin/gem' returned 1: RubyGems is a sophisticated package manager for Ruby.
```

Is executing using the gem command from the base class.

So I'm :+1: with the changes, just sad puppet is not consistent in how commands are overridden.

queeno: @joshcooper I totally agree with you.

queeno:
@melissa I believe I addressed all the issues raised.
I introduced `uninstalled_options` and I expanded the gem.rb unit tests, so we now test for uninstalling gems as well.

kylog:
This all seems like goodness to me (thanks @queeno!), but this feels like two different tickets/PRs:
1) fix uninstall
2) add uninstall_options
I think that separation will be clearer to someone who bumps into the 'gem uninstall' bug, since they would only be interested in the bugfix addressed in the original commit of this PR.

Thoughts?

melissa:
@kylog would separate commits be acceptable? It's a nice addition to have, even if it isn't necessarily related to the original topic of this PR.

kylog:
Yes, definitely! 

Sorry I wasn't clear. Both the bug fix and the feature add of install_options are definitely nice additions.

However, IMHO it's clearer to a future reader (e.g. someone who is interested in *just* the bugfix or *just* the addition of `install_options`) if there are two tickets and if each commit maps to just one of the two tickets. (I don't really care if it's separate PRs although that's what I suggested above.)

Does that make any sense? It's Friday afternoon so I may be uttering gibberish ;)

melissa:
Cool! My bad for tacking things on. @queeno apologies for giving you the extra work. Feel free to take the `:uninstall_options` work out if you want to. I can add that in later in that case. Otherwise, if you want to reference PUP-4203 for that work, that'd be awesome!

queeno:
@kylog I know exactly how that feels. It's Friday night here in London and I've already gone into the next phase: I'm totally out of control :stuck_out_tongue: 

I wouldn't mind opening a new PR for the `install_options` addition, however I would then need to rebase it once this PR is merged. I quite agree with @melissa, let's keep this nice and simple in this PR.

Thanks @melissa for opening a new issue in Puppet Jira for me. Will make the necessary changes in my branch and report back.

queeno:
@kylog @melissa all is well :+1: 
I've updated the commits and the relative jira issue [PUP-4203](https://tickets.puppetlabs.com/browse/PUP-4203).

Fingers crossed for the Travis build!

If you can't spot anything else, this PR should be good to go :smile:

kylog:
Last nit: can you put your commit messages in the imperative? (https://github.com/puppetlabs/puppet/blame/master/CONTRIBUTING.md#L53 for example/rationale.)

Other than that I'm :+1: 

Thanks!!

queeno:
No problem @kylog!
I've just rewritten my commit messages, hope they're compliant to the guidelines now :smile:

kylog:
One last imperative: can you amend 78f8308 so it's first line reads "(PUP-4203) Add unit tests for gem.rb uninstall_options"? Thanks!

Btw, timing wise I should mention that we're holding off on merges for a couple days while we tidy up our initial 4.0.0 release candidate. But after we open master back up for merges, this looks good. Thanks!!

queeno:
Ooops... no problem! :smile: 
All done and Travis passed! Please let me know if you spot anything else.

Thanks @kylog for your time reviewing this PR!

kylog:
:+1:

melissa:
Merged in at https://github.com/puppetlabs/puppet/commit/a01803875ae8f4659794d610c07cf7a086f8de12

MikaelSmith:
We manually merged with `git rebase master --whitespace=fix` to clean up some trailing whitespaces.

PR #3711, opened 2015-03-13
(PUP-4198) Add install_options feature to pip provider

butlern:
Add support for passing in additional options to pip install. The
current provider has no way to send additional options to pip.

For example, there are cases where you may want to connect to a
different package index, or add an extra index url to look for packages.
In my use case, we maintained our own pre-built wheels and wanted to
connect only to our pypi index and not to the general index at all, so
--find-links and --no-index achieve this.


branan:
@butlern this looks great. As soon as we're finished with the puppet 4.0 release cycle we'll get this merged

PR #3712, opened 2015-03-13
(PUP-4199) Add codedir to ARGV in config.ru example

stahnma:
While our config.ru was mostly modified to work with the AIO and Puppet
4, we missed adding the codedir option. This commit adds that with the
default of /etc/puppetlabs/code as the value.


joshcooper:
:+1: will merge after master branch is open for merges, likely Monday

PR #3714, opened 2015-03-13
(maint) Remove unused master_url_prefix setting

rlinehan:
When the changes were made to the Puppet 4 API to add a url prefix to all
endpoints, a `master_url_prefix` setting was added as this prefix was
initially going to be configurable. However, it was then determined that we
did not want the prefix to be configurable. While changes were made to the
code to set the prefix as a constant (see
540ec57817a7cfc6562fa725c60036210391a3d6), rendering this setting useless, the
setting was never removed. This commit removes this now-unused setting.


cprice404:
+1
ping @joshcooper

joshcooper:
@rlinehan makes sense, could you also rerun `bundle exec rake gen_manpages` (as it regenerates the puppet.conf man page)? You'll need to create Gemfile.local in the root of the project, add `gem "ronn"` to it, then run `bundle update && bundle exec rake gen_manpages`?

Also, we're holding on merges to master right now until after rc1 is packaged and ready.

rlinehan:
@joshcooper added a commit updating the manpages

man/man5/puppet.conf.5
```diff
@@ -589,7 +589,7 @@ The name of a registered environment data provider\. The two built in and regist
 The time to live for a cached environment\. This setting can be a time interval in seconds (30 or 30s), minutes (30m), hours (6h), days (2d), or years (5y)\. This setting can also be set to \fBunlimited\fR, which causes the environment to be cached until the master is restarted\.
 .
 .IP "\(bu" 4
-\fIDefault\fR: unlimited
+\fIDefault\fR: 0
```
joshcooper: ah good, this is the environment cache timeout change

PR #3715, opened 2015-03-14
(PUP-4205) Fix syntax error in heredoc support.

peterhuene:
Missing a '@' when referencing lexing_context.  This prevented multiple
heredocs on the same line from being parsed correctly.

### DIFF
--- a/lib/puppet/pops/parser/heredoc_support.rb
+++ b/lib/puppet/pops/parser/heredoc_support.rb
@@ -64,7 +64,7 @@ module Puppet::Pops::Parser::HeredocSupport
     # and it should start scanning after the first found \n (or if not found == error).
 
     if ctx[:newline_jump]
-      scn.pos = lexing_context[:newline_jump]
+      scn.pos = @lexing_context[:newline_jump]
     else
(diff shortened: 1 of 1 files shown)


peterhuene:
Need to add some lexer tests.

hlindberg:
Great catch, please target stable branch when updating the PR.

lib/puppet/pops/parser/heredoc_support.rb
```diff
@@ -64,7 +64,7 @@ def heredoc
     # and it should start scanning after the first found \n (or if not found == error).
 
     if ctx[:newline_jump]
-      scn.pos = lexing_context[:newline_jump]
+      scn.pos = @lexing_context[:newline_jump]
```
hlindberg: it should actually be `ctx[:newline_jump]` because it is much faster to access the local variable than the instance variable. At one point the code was refactored to use the shorter `ctx` instead of `lexing_context`. Apparently, this location was missed.

peterhuene: Good point, will fix, add a spec, and target stable.

PR #3716, opened 2015-03-14
issue/pup-4206-fedora-package-names

kylog:
Puppet acceptance assumes `f20` instead of `20` in various places, e.g. fedora repo configs are named like `...fedora-f20-...` but the actual packages are named like `...fedora-20-...`.

This could be fixed in packaging or fixed in acceptance. This PR assumes the latter is what we want to do but I don't care much. 

/cc @justinstoller @haus for comments.

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -126,7 +126,6 @@ module Puppet
         case platform
           when /^(fedora|el|centos)-(\d+)-(.+)$/
             variant = (($1 == 'centos') ? 'el' : $1)
-            fedora_prefix = ((variant == 'fedora') ? 'f' : '')
             version = $2
             arch = $3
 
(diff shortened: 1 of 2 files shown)


kylog:
Note that this fixes a spec test under the acceptance directory. I didn't realize we had any spec tests under acceptance, but there are a few. However we don't test them regularly and there are several existing failures. That's not addressed in this PR but I did file PUP-4207.

joshcooper:
@kylog fixing acceptance makes sense to me. Could you update the commit summary?

kylog:
I tweaked the message but defer to @justinstoller and @haus as to where to fix this mismatch issue.

haus:
@kylog @justinstoller I think I'd prefer to revert to the old behavior in vanagon for the moment as it is hardcoded in other RE tools. I'll open tickets to track the work and we can try to coordinate the change when things are more stable.

kylog:
Closing per @haus's comments.

PR #3717, opened 2015-03-15
(PUP-4209) Introduce fail attribute in notify resource

queeno:
Notify resources can output messages, but cannot alert the users about errors or fail the current puppet run.

This commit introduces the fail attribute to notify resources, which, if set to true, will output the message as error and fail the puppet run. The fail attribute defaults to false and this doesn't modify the current behaviour of the notify resource.

Example:

```puppet
notify { 'fail_puppet_run':
  message => 'Puppet is going to fail... Bye!',
  fail => true,
}
```

Output:

```shell
Notice: Compiled catalog for gds3562.local in environment production in 0.30 seconds
Error: Puppet is going to fail... Bye!
Error: /Stage[main]/Main/Notify[fail_puppet_run]/message: change from absent to Puppet is going to fail... Bye! failed: Puppet is going to fail... Bye!
Notice: Applied catalog in 0.06 seconds
```

### DIFF
--- a/lib/puppet/type/notify.rb
+++ b/lib/puppet/type/notify.rb
@@ -8,7 +8,14 @@ module Puppet
 
     newproperty(:message) do
       desc "The message to be sent to the log."
+
       def sync
+
+        case @resource["fail"]
+        when :true
+          self.fail self.should
+        end
+
         case @resource["withpath"]
         when :true
           send(@resource[:loglevel], self.should)
@@ -31,6 +38,13 @@ module Puppet
 
     newparam(:withpath) do
       desc "Whether to show the full object path. Defaults to false."
+      defaultto :false
+
+      newvalues(:true, :false)
+    end
+
+    newparam(:fail) do
+      desc "Triggers failure of the current puppet run. Defaults to false."
       defaultto :false
 
       newvalues(:true, :false)
--- a/spec/unit/type_spec.rb
+++ b/spec/unit/type_spec.rb
@@ -79,8 +79,8 @@ describe Puppet::Type, :unless => Puppet.features.microsoft_windows? do
   end
 
   it "can return an iterator for all set parameters" do
-    resource = Puppet::Type.type(:notify).new(:name=>'foo',:message=>'bar',:tag=>'baz',:require=> "File['foo']")
-    params = [:name, :message, :withpath, :loglevel, :tag, :require]
+    resource = Puppet::Type.type(:notify).new(:name=>'foo',:message=>'bar',:tag=>'baz',:fail=>false,:require=> "File['foo']")
+    params = [:name, :message, :withpath, :loglevel, :tag, :fail, :require]
     resource.eachparameter { |param|
       expect(params).to be_include(param.to_s.to_sym)
     }


joshcooper:
@queeno thank you for your contribution. We do already have a [`fail`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/parser/functions/fail.rb) function which is evaluated during catalog compilation on the master. Is this sufficient? If not, can you describe why failing the `notify` is required?

queeno:
@joshcooper 
The idea behind introducing a `fail` attribute to the notify resource is to hard fail the puppet agent run, as opposed to the catalog compilation performed by the puppet master.

In my use case, I'd like to trigger a puppet agent failure if some (manually deployed) files are missing from the disk or their permissions are not set correctly.
I'm also considering to write a patch so that notify supports a `refreshonly` attribute, similar to `exec`. 

What do you think? Does it sound sensible to you?

joshcooper:
@queeno Presumably, you'll need logic in your manifest to determine whether to add the `fail` parameter to the `notify` resource. So during catalog compilation, you'll need to check if the agent is in a good state (has the required manually deployed files and permissions). But that makes me think you should add a custom fact that checks the agent's state, and then have your manifest check the value of the fact:

```
unless $::files_present {
  fail("agent is in a bad state")
}
```

If you truly need to fail the run *during* catalog application, then I don't think adding the `fail` parameter to the `notify` resource will help. Instead, you'd need to create a custom type & provider, to evaluate the system state, and `fail` the resource.

queeno:
@joshcooper Using custom facter facts is definitely a fascinating option, which I didn't consider before. Thank you for suggesting that.

I was also reconsidering the rationale behind this PR. Causing a puppet agent failure is definitely something nasty to do. I guess it would leave the system in a random and unmanaged state due to out of order execution.

How would you feel introducing a `warning` attribute rather than a hard `fail` into the notify resource? The idea behind this PR was to raise user awareness when some critical resource was changed or was absent from the system.

Again, taking into account the scenario where Puppet needs to revert private key permissions, it could be nice if the file resources managing the keys could refresh a notify resource (see PR #3720) which would then output a clear warning.

The final result (including PR #3720) would look like this:

```shell
bundle exec puppet apply -e "notify{ 'change_private_key_permissions': message => 'The permissions have changed!', warn => true, refreshonly => true } file { '/tmp/private.pem': content => 'PRIVATE_KEY', mode=>'0400',notify=>Notify['change_private_key_permissions']}"
```

```puppet
Notice: Compiled catalog for simons-macbook-pro.local in environment production in 0.80 seconds
Notice: /Stage[main]/Main/File[/tmp/private.pem]/mode: mode changed '0777' to '0400'
Warning: /Stage[main]/Main/Notify[change_private_key_permissions]/message: The permissions have changed!
Notice: /Stage[main]/Main/Notify[change_private_key_permissions]: Triggered 'refresh' from 1 events
```

Would you agree this is a useful feature to have?

ffrank:
@queeno would it not be more straight forward to add a trivial custom type (in a module) that invariably botches its `validate` check?

Personally, I think failing the compilation is plenty alarming, so the `fail` function is sufficient.

ffrank:
Concerning a `warning` from `notify` resources that would be triggered through `refresh` events - you can have that simpler (and more accurate) through the `loglevel` metaparameter.

    file { '/tmp/private.pem': content => 'PRIVATE_KEY', mode=>'0400', loglevel => 'warning' }

branan:
So when I look at this along with #3720, it looks to me that you are trying to cause puppet to fail when certain resources change.

I don't think the Notify type should be responsible for this. Either a new type that is responsible for handling this should be created (and probably kept in a module, isntead of core) or you should write a report processor that can handle the requirements of your environment.

Using notify for this sort of compliance check seems like a "when you have a hammer everything looks like a nail" sort of situation

PR #3720, opened 2015-03-16
(PUP-4213) Add refreshonly attribute to notify resource

queeno:
Notify resources are always added to the catalog and run all the time, producing a notification message.

It would be useful to add a `refreshonly` attribute that mimics exec's `refreshonly`.
This attribute should trigger the execution of notify on refresh events only.

For example, it'd be useful to notify the user when some critical files are being changed.

Example:

```puppet
file { '/tmp/key.pem':
  content => 'PRIVATE_KEY',
  notify => Notify['private_key_content_change'],
}
notify { 'private_key_content_change':
  message => 'One of your private key\'s content has just changed!',
  refreshonly => true,
}
```

This results into the following output on the first run:

```shell
Notice: Compiled catalog for gds3562.local in environment production in 0.50 seconds
Notice: /Stage[main]/Main/File[/tmp/key.pem]/ensure: defined content as '{md5}ec6532e2d9d099e9b67c97dea5cd05fd'
Notice: One of your private key's content has just changed!
Notice: /Stage[main]/Main/Notify[private_key_content_change]: Triggered 'refresh' from 1 events
```

And on the second run:

```shell
Notice: Compiled catalog for gds3562.local in environment production in 0.51 seconds
Notice: Applied catalog in 0.06 seconds
```

### DIFF
--- a/lib/puppet/type/notify.rb
+++ b/lib/puppet/type/notify.rb
@@ -6,6 +6,10 @@ module Puppet
   Type.newtype(:notify) do
     @doc = "Sends an arbitrary message to the agent run-time log."
 
+    def refresh
+      self.property(:message).sync
+    end
+
     newproperty(:message) do
       desc "The message to be sent to the log."
       def sync
@@ -23,6 +27,7 @@ module Puppet
       end
 
       def insync?(is)
+        return true if @resource["refreshonly"] == :true
         false
       end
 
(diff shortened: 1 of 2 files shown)


melissa:
@joshcooper have you had an opportunity to look over this?

branan:
So when I look at this along with #3717, it looks to me that you are trying to cause puppet to fail when certain resources change.

I don't think the Notify type should be responsible for this. Either a new type that is responsible for handling this should be created (and probably kept in a module, isntead of core) or you should write a report processor that can handle the requirements of your environment.

Using notify for this sort of compliance check seems like a "when you have a hammer everything looks like a nail" sort of situation

PR #3722, opened 2015-03-16

ferventcoder:

Supersedes #3616.

### DIFF
--- a/lib/puppet/file_system/file_impl.rb
+++ b/lib/puppet/file_system/file_impl.rb
@@ -75,6 +75,10 @@ class Puppet::FileSystem::FileImpl
     path.read
   end
 
+  def read_preserve_line_endings(path)
+    read(path)
+  end
+
   def binread(path)
     raise NotImplementedError
   end
--- a/spec/unit/file_system_spec.rb
+++ b/spec/unit/file_system_spec.rb
@@ -5,6 +5,18 @@ require 'puppet/util/platform'
 describe "Puppet::FileSystem" do
   include PuppetSpec::Files
 
+  def with_file_content(content)
+    path = tmpfile('file-system')
+    file = File.new(path, 'wb')
+    file.sync = true
+    file.print content
+
+    yield path
+
+  ensure
+    file.close
+  end
+
   context "#exclusive_open" do
     it "opens ands allows updating of an existing file" do
       file = file_containing("file_to_update", "the contents")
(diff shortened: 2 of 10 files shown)


spec/unit/file_system_spec.rb
```diff
@@ -5,6 +5,14 @@
 describe "Puppet::FileSystem" do
   include PuppetSpec::Files
 
+  def with_file_content(content)
+    path = tmpfile('file-system')
+    file = File.new(path, 'wb')
+    file.sync = true
+    file.print content
+    yield path
```
joshcooper: we should explicitly close the file handle so we don't rely on finalizers, or alternatively use the block form of `File.open`

ferventcoder: Fair enough - I picked this up from another area of specs.

Iristyle:
Travis + Ruby 1.8.7 is not happy about

```
  NameError:
       uninitialized constant Puppet::FileSystem::FileImpl::Encoding
     # ./lib/puppet/file_system/file_impl.rb:79:in `read_preserve_line_endings'
```

lib/puppet/file_system/file_impl.rb
```diff
@@ -75,6 +75,14 @@ def read(path)
     path.read
   end
 
+  def read_preserve_line_endings(path)
+    contents = path.read( :mode => 'rb', :encoding => Encoding::UTF_8)
```
Iristyle: This is what Travis is not happy about ^

Iristyle: So thinking on this a little bit more, I think that we should change this method to look like 

```ruby
def read_preserve_line_endings(path)
    File.read(path)
end
```

That way we can reduce the surface area of this change and continue to preserve the behavior on non-Windows machines in the callsites that have changed in 

https://github.com/puppetlabs/puppet/pull/3722/files#diff-286f148216c044806e2a9d4beafe0925R29

https://github.com/puppetlabs/puppet/pull/3722/files#diff-965778ed2c6767ce9918080492dc0b37R101

Then for Windows specifically we could override the behavior to implement the method the way you have done here.  That way we also subvert the Ruby 1.8.7 / TravisCI issue as well (since we don't care about 1.8.7 on Windows)

ferventcoder: Fixed that up.

spec/unit/file_system_spec.rb
ferventcoder: Okay, file handle is closed

PR #3723, opened 2015-03-16
(PUP-4195) Update node configuration for release

MikaelSmith:
Updates the node configuration to use rhel7 for all masters - other
master platforms are tested as part of puppetserver testing - and ensure
all 32-bit and 64-bit platforms for release are tested.

Dropping testing for Fedora 18 and 19.


MikaelSmith:
Acceptance results at: https://gist.github.com/MikaelSmith/44b2670c8d77fc2a97c7

MikaelSmith:
The common failures were
* `env` missing on Solaris
* package conflicts on centos5, rhel5
```
	Running rpm_check_debug
	ERROR with rpm_check_debug vs depsolve:
	puppet conflicts with puppet-agent-0.3.2.158.g924a559-1.el5.x86_64
	facter conflicts with puppet-agent-0.3.2.158.g924a559-1.el5.x86_64
	hiera conflicts with puppet-agent-0.3.2.158.g924a559-1.el5.x86_64
	mcollective conflicts with puppet-agent-0.3.2.158.g924a559-1.el5.x86_64
	mcollective-common conflicts with puppet-agent-0.3.2.158.g924a559-1.el5.x86_64
	mcollective-client conflicts with puppet-agent-0.3.2.158.g924a559-1.el5.x86_64
	Complete!
```

joshcooper:
Thanks @MikaelSmith, the rhel5 issue is https://tickets.puppetlabs.com/browse/RE-4159

kylog:
Some comments:
* I *think* that some configs are obsolete and can now be deleted: `all*` `foss*` `standalone`
* We could move the `solaris` nodes to `pe`. (Or just delete `pe` until we get to the point of doing tests against PE-only packages.)
* If we don't *need* the `master` host to include the `agent` role (and I think we fixed this), then we should consider where we *want* to test agent-on-master functionality. I'd suggest *only* doing so on the server platforms we test against, i.e. for now rhel7.
* Re file naming conventions, could we match those in the puppet-agent repo (or v-v)? That way, less ad hoc translation is required.

MikaelSmith:
I created a Ruby script at https://gist.github.com/MikaelSmith/b8c4a25a1869907fb0a4 to create the updated versions. I'm half tempted to put that in Rake and just generate configs on the fly.
* I don't see a reason for `foss*` or `standalone*` anymore. `all*` is kind of convenient to smoke test all platforms, but not hard to create manually.
* Moving `solaris` to `pe` or dropping `pe` sounds fine. Who would have a stake in that?
* Dropping `agent` role for the `master` host sounds great; the rhel7.yaml covers `agent` role for `rhel7`, and that would potentially speed up tests.
* Matching naming sounds good and is easy to do... is there automation that uses these right now?

er0ck:
*please* do not generate configs from rake.  most of us don't run from rake to bring up environments or locally running acceptance when developing tests. 
agreed that we need a better config file generator, but that's for outside of the Rakefiles, IMO, and can be used across projects, etc, etc

acceptance/config/nodes/all.yaml
```diff
@@ -3,9 +3,9 @@ HOSTS:
     roles:
```
er0ck: +1 to delete this file, as kylo suggested.

acceptance/config/nodes/lucidx64.yaml
```diff
@@ -3,15 +3,15 @@ HOSTS:
     roles:
```
er0ck: if we're renaming these anyway, can we please name them after the os name? we shouldn't have to know OS release codenames to run tests.  
actually, since we're renaming them all, why don't we follow the vmpooler template naming conventions?
so this one would be "ubuntu-lucid-x86_64"

MikaelSmith: I think that's in-line with Kylo's suggestion to use puppet-agent platform names; however, it's slightly different.

Puppet-Agent uses `ubuntu-10.04-amd64`, vmpooler uses ubuntu-1004-x86_64. So some naming sanity would be nice here; I'm tempted to follow the vmpooler format, because it's likely harder to change than the puppet-agent configuration.

er0ck: i'm cool with either one as long as they are sane and semantic.

MikaelSmith:
Manually running acceptance to test that removing the `agent` role works.

MikaelSmith:
acceptance run on el-7-x86_64 passed.

acceptance/config/nodes/ubuntu-lucid-i386.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - master
+    platform: el-7-x86_64
+    hypervisor: vcloud
+    template: redhat-7-x86_64
+  agent:
+    roles:
+      - agent
+    platform: ubuntu-lucid-i386
+    hypervisor: vcloud
+    template: ubuntu-1004-i386
```
joshcooper: Unfortunately we're not consistent in our naming. The VM template is `ubuntu-1004-i386`. The BUILD_TARGETS as specified in jenkins AIO pipeline and puppet-agent is [ubuntu-10.04-i386](https://github.com/puppetlabs/puppet-agent/blob/master/configs/platforms/ubuntu-10.04-i386.rb). This yaml file and the `platform` are `ubuntu-lucid-i386`. It also looks like we sometimes use `amd64` vs `x86_64`... And the TEST_TARGET in AIO are something else...

MikaelSmith: Yup... I decided to make the file names and platform name match.

kylog:
In terms of naming conventions, I *think* that:
* BUILD_TARGETS in the Jenkins job matches the platform names found in puppet-agent's `config/platforms`
* TEST_TARGETS in the Jenkins job matches the config file names in puppet's `acceptance/config/nodes`

And then in the latter config files:
* `platform` is arbitrary but must match confines in puppet's acceptance tests
* 'template` matches the names used in vmpooler

Also note that puppet-agent has to map *its* naming convention to the pooler's with lines like: https://github.com/puppetlabs/puppet-agent/blob/215aaf7e13191c2d7af2309148c68f0eb91adb24/configs/platforms/debian-6-i386.rb#L10

It seems like we could make this all simpler if we consolidated on one naming convention. I nominate the pooler's because it's probably most ubiquitous. So what if:
a) puppet-agent used vmpooler's naming for its own platforms (it could then generate `vcloud_name` from the platform filename rather than mapping it per-file)
b) puppet (and other projects) switches to beaker config file names following the same convention - the template field would then always match the file name
c) puppet (and other projects) could make the `platform` field match the same naming convention (which would require touching some acceptance tests as well)

Thoughts? /cc @haus @joshcooper @mikaelsmith

MikaelSmith:
That sounds fine. Separating `platform` and `template` allows some flexibility if we have to use funny template names, like minor version bumps, so I'm not sure I'd drop one of the fields.

kylog:
Agreed on flexibility: how about we keep both `platform` and `template` but default to having them match? (I don't think the flexibility has come up often - in fact I can't remember an instance. Knock on wood.)

MikaelSmith:
Platform use is scattered around, so I'm going to handle changing the `platform` entries for a new ticket.

kylog:
+1 on the change to `platform` being another ticket. It's all intra puppet so more self-contained in any event.

@haus, thoughts on making puppet-agent platform names match pooler platform names?

haus:
@kylog i'm not in favor of that. i really don't like that part of the pooler. I'd rather add some sort of linking to the pooler so that the more correct platform name would also work in addition to the name the pooler expects today.
amd64 is what debian and ubuntu use, x64 is what windows uses, and x86_64 is what el and fedora use. I'd rather be consistent with the OS in public packaging (or soon to be public).

kylog:
Okay so I'm going to have to throw some of my dreams of standardizing in the dream incinerator, but that's okay.

But I *think* there still might be benefit to having puppet's acceptance/config/node filenames match with puppet-agent's config/platform filenames. This would allow BUILD_TARGETS == TEST_TARGETS (so maybe we could toss one and just have TARGETS).

There would still be 3 or 4 naming conventions to be aware of:
* beaker
* pooler
* puppet-agent
* package names (it might be that nothing in the puppet repo has this knowledge)

Thoughts on this?

haus:
@kylog I love having 1 standard and using it everywhere. There are more places than just that, but that is a good list to start with (PE has more places, but lets leave that can on the shelf for now).

joshcooper:
windows breaks the BUILD_TARGETS == TEST_TARGETS because we only build 32 and 64 bit packages (win-x64 & win-x86), but then test on a number of different variations (on 64-bit OS, test 32-bit and 64-bit puppet).

kylog:
Okay final offer: cryptopgraphic hashes of random cat gifs on the internet for our platform names.

er0ck:
1.yaml, 2.yaml, 3.yaml, 4

On Wed, Mar 18, 2015 at 3:38 PM, Kylo Ginsberg <notifications@github.com>
wrote:

> Okay final offer: cryptopgraphic hashes of random cat gifs on the internet
> for our platform names.
>
> —
> <https://github.com/puppetlabs/puppet/pull/3723#issuecomment-83213270>.
>



-- 
Eric Thompson
eric.thompson@puppetlabs.com <erict@puppetlabs.com>
Senior Software Quality Assurance Engineer

MikaelSmith:
One further note to add: having consistency between Beaker config and Puppet-Agent would allow us to have a simpler command to build a package and run acceptance tests against it.

kylog:
Okay so we need to retain BUILD_TARGETS and TEST_TARGETS in general b/c of (at least) Windows. But for the unixes, any downside to a match between `puppet/acceptance/config/nodes` and `puppet-agent/config/platforms`?

er0ck:
pretty happy about the filenames considering all the options and trade-offs.  thanks!

PR #3724, opened 2015-03-16
(maint) Update Fedora package location

MikaelSmith:
Packaging dropped the 'f' prefix that was previously added to the Fedora
version number in URLs, such as 'f20'. Without this change, acceptance
tests from packages (AIO) fail because the URL is wrong.

Update acceptance setup to use the new URLs.


joshcooper:
Looks like this is a dup of #3716

MikaelSmith:
Yup, he fixes the spec test too, so closing mine.

PR #3725, opened 2015-03-16
(PUP-4231) Log file content hash when using ensure=>present

elyscape:
When `ensure => present` and file content is provided via `content` or
`source`, the message logged was that the file was created. This is in
contrast to when `ensure => file` with content provided via `content` or
`source`, in which scenario the message logged contains the file's hash.
This makes both values log the file's hash.

### DIFF
--- a/lib/puppet/type/file/ensure.rb
+++ b/lib/puppet/type/file/ensure.rb
@@ -115,7 +115,7 @@ module Puppet
     end
 
     def change_to_s(currentvalue, newvalue)
-      return super unless newvalue.to_s == "file"
+      return super unless ["file", "present"].include?(newvalue.to_s)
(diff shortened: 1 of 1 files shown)


elyscape:
This is targeted at the stable branch.  I can make a separate PR that targets the master branch if that would be preferable.

branan:
I'm not sure exactly where we want to target this. I'll get a consensus from the group at tomorrows PR Triage meeting.

branan:
@elyscape We just discussed this in the triage meeting. We would like you to retarget at master.

Thanks!

elyscape:
@branan Done!  See #3732.

branan:
Thanks!

PR #3727, opened 2015-03-17
(PUP-4094) Update description of environment_timeout in defaults.rb

nfagerlund:
The defaults.rb file didn't have any advice on how to use this setting, so this
commit adds our current best practices.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -454,10 +454,33 @@ module Puppet
     :environment_timeout => {
       :default    => "0",
       :type       => :ttl,
-      :desc       => "The time to live for a cached environment.
+      :desc       => "How long the Puppet master should cache data it loads from an
+      environment.
       #{AS_DURATION}
-      This setting can also be set to `unlimited`, which causes the environment to
-      be cached until the master is restarted."
+      A value of `0` will disable caching. This setting can also be set to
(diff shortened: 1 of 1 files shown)


lib/puppet/defaults.rb
```diff
@@ -454,10 +454,30 @@ def self.default_diffargs
...
+      users update their code without any extra steps, but it lowers the
+      performance of your Puppet master.
+
+      We recommend setting this to `unlimited` and explicitly refreshing your
+      Puppet master as part of your code deployment process.
+
+      * With Puppet Server, you should refresh environments by calling the
+        `environment-cache` API endpoint. See the docs for the Puppet Server
+        administrative API.
+      * With a Rack Puppet master like Passenger, you should restart the web server.
```
ahpook: Better, you can restart the *application* server, not the whole web server -- in Passenger this means touching `tmp/restart.txt` under the PassengerRoot directory: https://www.phusionpassenger.com/documentation/Users%20guide%20Nginx.html#_redeploying_restarting_the_rack_application

ahpook:
:+1: , you can take or leave my line comment about passenger's restart.txt.

lib/puppet/defaults.rb
nfagerlund: OH MAN, I forgot about restart.txt. Yeah, good call.

PR #3728, opened 2015-03-18
add-puppet-gem-provider-for-puppet-agent

kylog:


kylog:
Note that the second commit here assumes the `uninstall_options` feature added in #3708, so it will fail spec tests until that is merged.

lib/puppet/provider/package/puppet_gem.rb
```diff
@@ -0,0 +1,10 @@
+require 'puppet/provider/package'
+
+Puppet::Type.type(:package).provide :puppet_gem, :parent => :gem do
+  desc "Puppet Ruby Gem support. This provider is useful for managing
+        gems needed by the ruby provided in the puppet-agent package."
+
+  has_feature :versionable, :install_options, :uninstall_options
+
+  commands :gemcmd => "/opt/puppetlabs/puppet/bin/gem"
```
joshcooper: support windows too?

branan:
@kylog I didn't expect us to be getting #3708 in until after 4.0 final, which would also block this to 4.1... Is that intended? Do yoy plan to bring the other pull into 4.0?

kylog:
@branan ah, good question. Yes, I think this will ease 4.0 adoption so, yes, I'd vote that we should pull this (and #3708) into 4.0.

branan:
@kylog got it. We'll move ahead with merging #3708 this week so we can get this in.

MikaelSmith:
#3708 has been merged.

branan:
@kylog can you refresh this commit somehow so we can get travis to re-run?

branan:
Also, shoot. I think we merged #3708 into 4.1 (master). We'll probably want to cherry-pick that back to stable and re-target this PR.

branan:
Assuming we still want this in 4.0, that is (and I think we do)

branan:
With the number of meetings @kylog is in today, I'll take care of all that. Closing this for now, and I'll open a new one targeted at stable with #3708 cherry-picked.

PR #3730, opened 2015-03-18
(PUP-4232) add acceptance: lookup function

er0ck:
This change adds acceptance tests for the agnostic data lookup()
function provided for puppet 4x.
The tests include baseline functionality for testing round-tripping of
data from environments, modules, and hiera through implied class
parameter lookups and calling lookup() directly.
This change does not test the various data types that are enabled in
lookup(), hiera, etc.  These are well covered in the unit tests.

### DIFF
--- a/acceptance/tests/parser_functions/lookup.rb
+++ b/acceptance/tests/parser_functions/lookup.rb
@@ -7,6 +7,7 @@ testdir = master.tmpdir('lookup')
 step 'Setup'
 
 module_name                     = "data_module"
+module_name2                    = "other_module"
 
 env_data_implied_key            = "env_data_implied"
 env_data_implied_value          = "env_implied_a"
@@ -17,6 +18,7 @@ module_data_implied_key         = "module_data_implied"
 module_data_implied_value       = "module_implied_b"
 module_data_key                 = "module_data"
 module_data_value               = "module_b"
+module_data_value_other         = "other_module_b"
 
 env_data_override_implied_key   = "env_data_override_implied"
 env_data_override_implied_value = "env_override_implied_c"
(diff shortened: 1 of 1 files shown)


er0ck:
@hlindberg should we add tests here for including classes via lookup()?

hlindberg:
The referenced source line is in a helper method that constructs the arguments passed to the lookup implmentation Puppet::Pops::Lookup - it is not relevant for class defaults being looked up via data binding. (Yes, that should be tested separately, including the full stack of global hiera, environment, and module data). The `hiera_include` function is simple, it looks up 'classes' and then loops over them to include those. If an acceptance test correctly looks up (global, env, module) then hiera_include should work (tested with an integration test). Important to test the kind of lookup it performs though.

er0ck:
@thallgren please take a look at the upstream ticket for this.  test is now failing due to merge of pup-4081.  see my questions in pup-4232

er0ck:
@thallgren fixed the test to use fully qualified names when required.  i also added a test that grabs data from another function.

er0ck:
@thallgren a fully qualified name is used there, it's just matching the output on the keyname followed by the value.  on this line it's matching on the other module's data.
the notify and lookups actually occur in the class above.  all the module data uses lookups that are fully qualified.
is that your concern?

thallgren:
No, my concern probably stems from pure ignorance. I don't understand what happens here:

    assert_match("#{module_data_key} #{module_data_value_other}", stdout)

Exactly what does `assert_match(x, stdout)` do with x?

er0ck:
assert_match(x, stdout)
asserts that x has a regex match in stdout.
it's a beaker method that is more or less inherited from minitest

On Mon, Mar 23, 2015 at 8:38 AM, Thomas Hallgren <notifications@github.com>
wrote:

> No, my concern probably stems from pure ignorance. I don't understand what
> happens here:
>
> assert_match("#{module_data_key} #{module_data_value_other}", stdout)
>
> Exactly what does assert_match(x, stdout) do with x?
>
> —
> <https://github.com/puppetlabs/puppet/pull/3730#issuecomment-85053324>.
>



-- 
Eric Thompson
eric.thompson@puppetlabs.com <erict@puppetlabs.com>
Senior Software Quality Assurance Engineer

PR #3731, opened 2015-03-18

thallgren:
This pull-request removes the responsibility for formatting the log messages
from the `IssueReporter` and the `ExternalFilerError` classes. Instead,
the formatting is deferred and performed by the `Log.to_s` method (if
such formatting is at all needed).

Some instance variables were added, both to the `ExternalFileError` class
and to the Log class to accommodate detailed information about
issue_code, environment, node, and position.

A new Log destination _'json_file'_ was added. It will not use the
formatted message but instead output all details in a JSON style
Map. Exception backtrace is output as a JSON Array.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -89,15 +89,9 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
 
     benchmark(:notice, str) do
       Puppet::Util::Profiler.profile(str, [:compiler, :compile, node.environment, node.name]) do
-        begin
-          config = Puppet::Parser::Compiler.compile(node)
-        rescue Puppet::Error => detail
-          Puppet.err(detail.to_s) if networked?
-          raise
-        end
+        config = Puppet::Parser::Compiler.compile(node)
       end
     end
-
     config
   end
 
--- a/lib/puppet/util/log.rb
+++ b/lib/puppet/util/log.rb
@@ -80,9 +80,20 @@ class Puppet::Util::Log
   # Create a new log message.  The primary role of this method is to
   # avoid creating log messages below the loglevel.
   def Log.create(hash)
-    raise Puppet::DevError, "Logs require a level" unless hash.include?(:level)
-    raise Puppet::DevError, "Invalid log level #{hash[:level]}" unless @levels.index(hash[:level])
-    @levels.index(hash[:level]) >= @loglevel ? Puppet::Util::Log.new(hash) : nil
+    level = hash[:level]
+    raise Puppet::DevError, "Logs require a level which is a symbol or string" unless level.respond_to? "to_sym"
+
+    level = level.to_sym
+    index = @levels.index(level)
+    raise Puppet::DevError, "Invalid log level #{level}" if index.nil?
+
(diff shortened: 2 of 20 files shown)


cprice404:
Has this been tested with Puppet Server, or has any design work been put into the compatibility with the Puppet Server logging framework?

thallgren:
It has not been tested with Puppet Server but care has been taken to not change the current log output.

lib/puppet/util/log.rb
```diff
@@ -238,44 +249,41 @@ def self.validlevel?(level)
...
   def self.from_data_hash(data)
-    obj = allocate
-    obj.initialize_from_hash(data)
-    obj
+    symkeyed_data = {}
+    data.each_pair { |k,v| symkeyed_data[k.to_sym] = v }
+    new(symkeyed_data)
   end
 
-  attr_accessor :time, :remote, :file, :line, :source
```
hlindberg: was :remote never used? (it is missing in the new version).

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -89,15 +89,9 @@ def compile(node)
 
     benchmark(:notice, str) do
       Puppet::Util::Profiler.profile(str, [:compiler, :compile, node.environment, node.name]) do
-        begin
-          config = Puppet::Parser::Compiler.compile(node)
-        rescue Puppet::Error => detail
-          Puppet.err(detail.to_s) if networked?
-          raise
-        end
```
hlindberg: where is the above done now? (seems like it logged only if networked (an agent request? or something else?). understand that new version does not catch anything (better), but want to make sure that the implication of not logging if networking is clear.

lib/puppet/util/log.rb
```diff
@@ -18,11 +18,11 @@ class Puppet::Util::Log
   @desttypes = {}
 
   # Create a new destination type.
-  def self.newdesttype(name, options = {}, &block)
+  def self.newdesttype(name, options = {}, parent = nil, &block)
```
hlindberg: so this should be backwards compatible so this logdest works: https://github.com/puppetlabs/puppet-server/blob/master/src/ruby/puppet-server-lib/puppet/server/logger.rb

hlindberg:
this should target the branch 3.8, not master

lib/puppet/util/log.rb
thallgren: I couldn't find any usage of it anywhere in the code.

lib/puppet/indirector/catalog/compiler.rb
thallgren: The only statement within begin/rescue here was a call to `Puppet::Parser::Compile.compile(node)` (the class method). That method will _always_ rescue, log, and re-raise all exceptions. So what this change does is avoiding a duplicate from being logged when networked.

lib/puppet/util/log.rb
thallgren: Yes, it's backward compatible. The only dest_type that make use of this argument at present is the _json_file_ (passes the _file_ as it's parent). All others remain unchanged.

thallgren:
This PR is replaced by [PR-3737](3737) that targets branch 3.8.

PR #3733, opened 2015-03-19
(PUP-4190) Catch exit in device spec

hunner:
The PR for PUP-4032 added exits to the code, and the PR for PUP-4190
added tests that didn't catch the exits. This should dovetail the two
and make specs pass again.


kylog:
Travis is really backed up right now. This passes at my desk. Merging.

PR #3735, opened 2015-03-19
(PUP-4230) Log content hash when removing a file

elyscape:
When removing a file with `ensure => absent`, the message logged is that
the file was removed. By contrast, when files are created or modified,
the message logged contains a hash of current and, if applicable,
previous contents. This changes file removal log entries to also contain
the hash of the removed file.

### DIFF
--- a/lib/puppet/type/file/ensure.rb
+++ b/lib/puppet/type/file/ensure.rb
@@ -114,8 +114,14 @@ module Puppet
       value
     end
 
+    @old_checksum = nil
+
     def change_to_s(currentvalue, newvalue)
-      return super unless [:file, :present].include?(newvalue)
+      return super unless [:file, :present, :absent].include?(newvalue)
+
+      # The change message for removal of directories
+      # and links is handled by the parent class.
+      return super if newvalue == :absent && @old_checksum.nil?
 
       return super unless property = @resource.property(:content)
 
@@ -127,7 +133,9 @@ module Puppet
         should = property.should
       end
       if should == :absent
-        is = property.retrieve
+        # When this method is run, the file has already been
+        # deleted, so we need to manually cache the checksum.
+        is = @old_checksum
       else
         is = :absent
       end
@@ -175,6 +183,19 @@ module Puppet
     end
 
     def sync
+      if self.should == :absent && !@resource.bucket.nil?
+        # Make a dummy content property so that Puppet
+        # will generate a content object, which we can
+        # use to the checksum of the current file.
+        @resource[:content] = :absent
+        # Make sure we're using the same hash as the
+        # bucket, so that our log is actually useful.
+        # This may need to change when PUP-4329 is fixed.
+        @resource[:checksum] = Puppet[:digest_algorithm].to_sym
+        # Note that content.retrieve will be `nil` unless
+        # the current file is a normal file.
+        @old_checksum = @resource.property(:content).retrieve
+      end
       @resource.remove_existing(self.should)
       if self.should == :absent
         return :file_removed
--- a/spec/integration/type/file_spec.rb
+++ b/spec/integration/type/file_spec.rb
@@ -1158,6 +1158,38 @@ describe Puppet::Type.type(:file), :uses_checksums => true do
           expect(logs).to be_empty
         end
       end
+
+      context "ensure is absent" do
+        let(:expected_checksum) do
+          expected_type = Puppet[:digest_algorithm]
+          expected_value = CHECKSUM_TYPES_TO_TRY.find { |type, value| type == expected_type }[1]
+          "{#{expected_type}}#{expected_value}"
+        end
+
+        # This may need to change when PUP-4329 is fixed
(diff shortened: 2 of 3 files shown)


elyscape:
Note that this requires #3732.

branan:
I would really like to see tests for this, since it pokes some hairy parts of the file type

elyscape:
@branan Any suggestions? I'm open to writing tests but there don't appear to be any preexisting tests for `Ensure#change_to_s` and the tests for `Ensure#sync` are all related to directory functionality.

lib/puppet/type/file/ensure.rb
```diff
@@ -175,6 +183,14 @@ def retrieve
     end
 
     def sync
+      if self.should == :absent
+        # Make a dummy content property so that we can
```
MikaelSmith: I think I understand this, but the comment could be clearer. What's the content property that exists here otherwise, and how is it set?

elyscape: > What's the content property that exists here otherwise, and how is it set?

Before we do this, it's `nil`.  When you put `:absent` into `@resource[:content]`, Puppet generates the correct object to go in there.

> It seems like you have to set this so Puppet will read the checksum of the file.

Pretty much.

MikaelSmith:
Some consensus on this is that we'd like to see a better message. The 'undefined content' message is used elsewhere, so we should probably special-case this change_to_s call for :absent.

MikaelSmith:
I think the other major concern is that recalculating the file hash can be expensive. If you don't need this functionality, why do it? So we might want to have this enabled only when using file buckets, or as a separate configuration option.

elyscape:
> I think the other major concern is that recalculating the file hash can be expensive. If you don't need this functionality, why do it? So we might want to have this enabled only when using file buckets, or as a separate configuration option.

That's an easy change.  I'll make it happen.

elyscape:
> Some consensus on this is that we'd like to see a better message. The 'undefined content' message is used elsewhere, so we should probably special-case this change_to_s call for :absent.

Any suggestions?  Maybe `removed '{md5}hash'` or `removed file '{md5}hash'`?

elyscape:
Message changed to `removed file with content '{md5}hash'`.

elyscape:
Calculating file hash is now done only when a filebucket is enabled for the resource.

MikaelSmith:
I have no further objections, other than testing (waiting on https://github.com/puppetlabs/puppet/pull/3752).

elyscape:
Added some tests.  Pinging @MikaelSmith for review.  If positive, I'll squash it down.

lib/puppet/type/file/ensure.rb
```diff
@@ -175,6 +183,19 @@ def retrieve
...
     def sync
+      if self.should == :absent && !@resource.bucket.nil?
+        # Make a dummy content property so that Puppet
+        # will generate a content object, which we can
+        # use to the checksum of the current file.
+        @resource[:content] = :absent
+        # Make sure we're using the same hash as the
+        # bucket, so that our log is actually useful.
+        # This may need to change when PUP-4329 is fixed.
+        @resource[:checksum] = Puppet[:digest_algorithm].to_sym
```
MikaelSmith: This means the `removed file` message won't match up with the `defined content` message if `Puppet[:digest_algorithm]` (global config) differs from `@resource[:checksum]` (resource property), but this was my suggested approach to fulfill the desired behavior of mapping to the Filebucket contents until PUP-4329 is fixed. The checksums will still be correct and self-consistent, as they include the checksum type in the message.

spec/integration/type/file_spec.rb
```diff
@@ -1091,6 +1091,30 @@ def build_path(dir)
           expect(logs).to be_empty
         end
       end
+
+      context "ensure is absent" do
+        let(:expected_checksum) do
```
MikaelSmith: This could have been done with the with_digest_algorithms shared context, but since it's a temporary behavior I'm happier with this approach.

MikaelSmith:
I think this looks good. I want to pass it by @joshcooper, and it'll likely need a rebase after #3787 is merged.

elyscape:
Rebased and squashed now that #3787 is merged.

elyscape:
Added a check that a second apply of the resource doesn't log anything, in keeping with similar logging tests.

melissa:
@joshcooper @MikaelSmith any updates on this?

melissa:
ping @joshcooper and @MikaelSmith

MikaelSmith:
I've responded on the ticket; I feel it's blocked by PUP-4329. That may mean we should close this PR for now.

peterhuene:
Hi @elyscape.  After discussing this with Josh and Michael, we feel that we cannot go forward with this PR while having to work around PUP-4329, along with concerns over performance recalculating checksums and resyncing content.  Please see PUP-4230 for feedback from @MikaelSmith.  We may be able to revisit this feature in the future, but for now we're closing the PR.  Thank you again for the contribution!

PR #3736, opened 2015-03-19

thallgren:
This commit removes the responsibility for formatting the log messages
from the `IssueReporter` and the `ExternalFileError` classes. Instead,
the formatting is deferred and performed by the `Log.to_s` method (if
such formatting is at all needed).

Some instance variables were added, both to the `ExternalFileError` class
and to the `Log` class to accommodate detailed information about
issue_code, environment, node, and position.

This commit also unifies the logic for a `Log` instance is created and
how its arguments are validated. All attributes except the source are
made read-only. The source can be changed because the `Puppet::Reports`
for `:log` alters it to include the name of the host.

A new `Log` destination _'json_file'_ was added. It will not use the
formatted message but instead output all details in a JSON style
Map. Exception backtrace is output as a JSON Array.

The merge to master will not be trivial. New tests have been added
that assumes that the file:line information is embedded in the
exception message (it no longer is since the formatting is handled
by the `Log.to_s`).

### DIFF
--- a/lib/puppet/error.rb
+++ b/lib/puppet/error.rb
@@ -12,37 +12,40 @@ module Puppet
     # This module implements logging with a filename and line number. Use this
     # for errors that need to report a location in a non-ruby file that we
     # parse.
-    attr_accessor :line, :file, :pos
+    attr_accessor :file, :issue_code, :environment, :node
+    attr_reader :line, :pos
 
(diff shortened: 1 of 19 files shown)


thallgren:
Gahh, used branch stable, not 3.8. Closing...

PR #3737, opened 2015-03-19

thallgren:
The main objective for this commit is to retain backward compatility
with the 3.x stream. It only touches the logged messages that stems
from the Puppet::Pops::IssueReporter.

The commit removes the responsibility for formatting the log messages
from the IssueReporter. Instead it passes all details to the Log
instance that is created, either by direct logging och by raising an
error. A new class named ParseErrorWithIssue (a subclass of ParseError)
that keeps track of the basic message and the issue_code was needed to
accomplish this. This class is also has the attributes 'environment'
and 'node' that will be assigned in rescue blocks before the exception
is re-thrown. Prior to this change, the rescue block would create a
new error instance with additional information in the actual message
but loose the detailed information.

The 'file' log destination was altered slightly so that if the path
ends with '.json', it will output each Log entry in JSON format. The
list of entries will be prefixed by a '[' (added when the file is
first created) and delimited with ','. The file will contain almost
valid JSON at all times but the last ']' will be missing.

### DIFF
--- a/lib/puppet/error.rb
+++ b/lib/puppet/error.rb
@@ -54,8 +54,34 @@ module Puppet
     include ExternalFileError
   end
 
+  # Contains an issue code and can be annotated with an environment and a node
+  class ParseErrorWithIssue < Puppet::ParseError
+    attr_reader :issue_code, :basic_message
+    attr_accessor :environment, :node
+
+    # @param message [String] The error message
+    # @param file [String] The path to the file where the error was found
+    # @param line [Integer] The line in the file
+    # @param pos [Integer] The position on the line
+    # @param original [Exception] Original exception
+    # @param issue_code [Symbol] The issue code
+    #
+    def initialize(message, file=nil, line=nil, pos=nil, original=nil, issue_code= nil)
+      super(message, file, line, pos, original)
+      @issue_code = issue_code
+      @basic_message = message
+    end
+
+    def to_s
+      msg = super
+      msg = "Could not parse for environment #{environment}: #{msg}" if environment
+      msg = "#{msg} on node #{node}" if node
+      msg
+    end
+  end
+
   # An error that already contains location information in the message text
-  class PreformattedError < Puppet::ParseError
+  class PreformattedError < Puppet::ParseErrorWithIssue
   end
 
   # An error class for when I don't know what happened.  Automatically
--- a/lib/puppet/pops/issue_reporter.rb
+++ b/lib/puppet/pops/issue_reporter.rb
@@ -23,7 +23,7 @@ class Puppet::Pops::IssueReporter
     emit_warnings = options[:emit_warnings] || false
     emit_errors = options[:emit_errors].nil? ? true : !!options[:emit_errors]
     emit_message = options[:message]
-    emit_exception = options[:exception_class] || Puppet::ParseError
+    emit_exception = options[:exception_class] || Puppet::ParseErrorWithIssue
 
     # If there are warnings output them
     warnings = acceptor.warnings
@@ -37,10 +37,10 @@ class Puppet::Pops::IssueReporter
           # deprecation of constructs in manifests! (It is not designed for that purpose even if
           # used throughout the code base).
           #
-          Puppet.warning(formatter.format(w)) if emitted_dw < max_deprecations
+          log_message(:warning, formatter, w) if emitted_dw < max_deprecations
           emitted_dw += 1
         else
-          Puppet.warning(formatter.format(w)) if emitted_w < max_warnings
+          log_message(:warning, formatter, w) if emitted_w < max_warnings
           emitted_w += 1
         end
         break if emitted_w >= max_warnings && emitted_dw >= max_deprecations # but only then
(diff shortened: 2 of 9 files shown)


lib/puppet/error.rb
```diff
@@ -12,37 +12,40 @@ module ExternalFileError
...
+    def initialize(message, file=nil, line=nil, pos=nil, original=nil, issue_code=nil)
       if pos.kind_of? Exception
         original = pos
         pos = nil
       end
       super(message, original)
+      @issue_code = issue_code
       @file = file unless (file.is_a?(String) && file.empty?)
+      self.line = line if line
+      self.pos = pos if pos
```
hlindberg: why are these two (line, pos) set conditionally? it was required for line earlier since it was converted to integer if being a string (which would fail if it was nil). Don't think the condition makes any sense now, except if we want a boolean false to have the same meaning as nil ??

hlindberg: line= transforms input to integer, shouldn't that be done here as well ?

thallgren: The reason for calling self.line and self.pos instead of just assign to the attributes directly is to ensure that they are converted to integers if needed and keep that conversion in one place.

The reason for having the calls conditional is to not define the attributes at all unless they are given.

lib/puppet/util/log.rb
```diff
@@ -244,39 +255,36 @@ def self.from_pson(data)
...
-    [:file, :line].each do |attr|
-      next unless value = args[attr]
-      send(attr.to_s + "=", value)
+    # Avoid setting these instance variables. We don't want them defined unless they exist
+    [:backtrace, :environment, :node, :issue_code, :file, :line, :pos].each do |attr|
+      value = args[attr]
+      instance_variable_set("@#{attr}", value) unless value.nil?
     end
 
-    Log.newmessage(self)
```
hlindberg: This removal changes the API/semantics of Log. If there is code that logs by calling Log.new (and not create) the result is silently dropping those entries. The original construct is horrible as it means it is logged immediately when creating the Log instance. I wonder if we need to make this conditional - when coming from create it passes an additional arg :do_not_log - thereby preserving the old API ? (+ log issue to clean this up asap on major release boundary - 4.0 if we can)

lib/puppet/util/log.rb
```diff
@@ -244,39 +255,36 @@ def self.from_pson(data)
...
-  end
+    self.source = args[:source]
+    tags = args[:tags]
+    tags.each { |t| tag(t) } unless tags.nil?
 
-  def initialize_from_hash(data)
```
hlindberg: This also alters the API

lib/puppet/util/log.rb
thallgren: I didn't find any calls to Log.new where this mattered. My preference would be to change the behavior because it's not just that it's logged, it's also first created with all bells and whistles and then discarded when log level is below the threshold. The create method will check log level first and then create the log entry.

lib/puppet/util/log.rb
```diff
@@ -284,47 +292,34 @@ def to_hash
...
+      hash[attr.to_s] = instance_variable_get(iv) if instance_variable_defined?(iv)
+    end
+    hash
   end
 
   def to_pson(*args)
     to_data_hash.to_pson(*args)
   end
 
-  def message=(msg)
```
hlindberg: API change

lib/puppet/util/log.rb
```diff
@@ -284,47 +292,34 @@ def to_hash
...
-    raise ArgumentError, "Puppet::Util::Log requires a message" unless msg
-    @message = msg.to_s
-  end
-
-  def level=(level)
```
hlindberg: API change (and this one is nasty since it means you can log with high level, then lower it and still get it in the output even if it is flltered on that level)

lib/puppet/util/log.rb
thallgren: The only place where this was called was from the class method `from_data_hash(data)` which first did an allocate and then this initialize_from_hash. The class method is still there with the same API.

lib/puppet/util/log.rb
thallgren: On the other hand, there are not calls that does that. Not in our code anyway and that's where the filtering on level is performed. Do we want to support possible third monkey patching?

hlindberg:
I don't think we can make these "API" changes (horrible as it is). We don't know all the code that may be calling these methods in the wild (they are not private). We can only see what is inside puppet. Searching trough all modules on the forge is required before even considering changing the API, and even then we are not certain we are not breaking anyone. We can do the break on a major boundary but not for 3.8. I think "logging" should be a theme for Puppet 5 as we have then also more firmly moved over to Puppet Server and we then do not need two levels of log configurations (= good opportunity to clean up).

lib/puppet/util/log.rb
thallgren: I didn't find any place where the message is changed.

lib/puppet/pops/issue_reporter.rb
```diff
@@ -84,4 +84,35 @@ def self.format_with_prefix(prefix, message)
...
+        :level => severity,
+        :message => formatter.format_message(diagnostic),
+        :issue_code => diagnostic.issue.issue_code,
+        :file => file,
+        :line => line,
+        :pos => pos,
+      })
+  end
+  private_class_method :log_message
+end
```
hlindberg: missing end of line

thallgren: Just out of curiosity, why does that matter?

kylog:
Travis missed this PR for some reason. Closing and reopening to kick travis.

thallgren:
@kylog I've noticed this too. In this particular case I'm 100% sure that I saw a travis build kick in shortly after I submitted the PR. Perhaps the problem is related to not reporting back once it's done?

PR #3740, opened 2015-03-19
(PUP-1628) Fix mount provider for AIX

hunner:
The way that AIX tracks its mounted filesystems is different than other
unix or linux operatingsystems. AIX uses /etc/filesystems and has
entirely different filesystem options.

This commit takes advantage of two parsedfile hooks, `post_parse` and
`to_line`, to read & generate /etc/filesystem entries. Parsedfile was
not created with arbitrary-ordering in mind, and any /etc/filesystem
entries will have their attributes reordered by `to_line`

Options in AIX without related properties in Puppet's mount type are
placed under the "options" property.

### DIFF
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -51,9 +51,28 @@ Puppet::Type.type(:mount).provide(
     # Override lines and use scan instead of split, because we DON'T want to
     # remove the separators
     def self.lines(text)
-      # Separate base on comments, blank lines, or a mountpoint followed by
-      # anything that isn't a mountpoint, comment, or blank line.
-      ret = text.scan(%r{(^\s*\*.*?$|^\s*?$|^\S+:(?:(?!(?:^\S+:|^\s*\*|^\s*$)).)*)}m).flatten
+      lines = text.split("\n")
+      filesystem_stanza = false
+      filesystem_index = 0
+      ret = Array.new
+      lines.each_with_index do |line,i|
+        if line.match(%r{^\S+:})
+          # Begin new filesystem stanza and save the index
+          ret[filesystem_index] = filesystem_stanza.join("\n") if filesystem_stanza
+          filesystem_stanza = Array(line)
+          filesystem_index = i
+          nil
+        elsif line.match(%r{^(\s*\*.*|\s*)$})
+          # Just a comment or blank line; add in place
+          ret[i] = line
+        else
+          # Non-comments or blank lines must be part of a stanza
+          filesystem_stanza << line
+        end
+      end
+      # Add the final stanza to the return
+      ret[filesystem_index] = filesystem_stanza.join("\n") if filesystem_stanza
+      ret = ret.compact.flatten
       ret.reject { |line| line.match(/^\* HEADER/) }
     end
     def self.header
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -61,6 +61,8 @@ Puppet::Type.type(:mount).provide(
           ret[filesystem_index] = filesystem_stanza.join("\n") if filesystem_stanza
           filesystem_stanza = Array(line)
           filesystem_index = i
+          # Eat the preceeding blank line
+          ret[i-1] = nil if i > 0 and ret[i-1] and ret[i-1].match(%r{^\s*$})
           nil
         elsif line.match(%r{^(\s*\*.*|\s*)$})
           # Just a comment or blank line; add in place
@@ -100,7 +102,7 @@ Puppet::Type.type(:mount).provide(
         end.compact.join("\n")
         result[:record_type] = memo[:record_type]
         special_options = Array.new
-        result[:name] = memo[:name].sub(%r{:$},'')
+        result[:name] = memo[:name].sub(%r{:\s*$},'').strip
         memo.each do |_,k_v|
           if k_v and k_v.is_a?(String) and k_v.match("=")
             attr_name, attr_value = k_v.split("=",2).map(&:strip)
@@ -134,6 +136,11 @@ Puppet::Type.type(:mount).provide(
         if result[:device] and result[:device].match(%r{^/})
           output << "\tdev\t\t= #{result[:device]}"
         elsif result[:device] and result[:device] != :absent
+          if ! result[:device].match(%{^.+:/})
+            # Just skip this entry; it was malformed to begin with
+            Puppet.err "Mount[#{result[:name]}]: Field 'device' must be in the format of <absolute path> or <host>:<absolute path>"
+            return result[:line]
+          end
           nodename, path = result[:device].split(":")
           output << "\tdev\t\t= #{path}"
           output << "\tnodename\t= #{nodename}"
@@ -164,9 +171,9 @@ Puppet::Type.type(:mount).provide(
           output << "\toptions\t\t= #{options.sort.join(",")}" unless options.empty?
         end
         if result[:line] and result[:line].split("\n").sort == output.sort
-          return result[:line]
+          return "\n#{result[:line]}"
         else
-          return output.join("\n")
+          return "\n#{output.join("\n")}"
         end
       end
     end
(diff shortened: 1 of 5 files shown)


ahpook:
WOW!! Awesome @hunner -- I'm super happy to see this addressed.

One comment, can you change the commit message to be prefixed with the public-facing ticket (PUP-1628) rather than the internal-only PE ticket? That way anyone can see the history not just PL folks.

stahnma:
All the :+1: s.  :smile:

hunner:
Limitations:
- It will reorder the filesystem attributes to a specific order (dev, nodename, vfs, then the rest alphabetic)
- Comments embedded in a filesystem attribute list will cause the following ones to be dropped (but between filesystem entries is fine)
- The `options` property is comma separated (fine) and must be in the order from the first bullet point, otherwise it will always try to update the value
- The `options` property is overloaded with all of AIX's extra attributes; perhaps adding new properties to the type would be better

helperton:
Hello Hunner, I've patched my parsed.rb and mount.rb with these changes.  What I'm seeing is that for "nodename" in /etc/filesystems the value is the entire string which is defined in the manifest "hostname:device" where it should just be "hostname".  So our /etc/filesystems look like this now:

/nfs/depot/someplace:
        dev             = /nfs/depot/someplace <-- should be right side of a split on ":" of "nodename")
        nodename  = nfsserver:/oracle_depot/mnt  <- should be left side of a split on ":" of "nodename")
        vfs             = nfs
        options      = nosuid,ro,soft

Should look like:
/nfs/depot/someplace: 
        dev = "/oracle_depot/mnt"  < -- correct
        vfs = nfs
        nodename = nfsserver <--- correct
        mount = true
        options = bg,soft,intr
        account = false

lib/puppet/provider/mount/parsed.rb
```diff
@@ -36,7 +42,118 @@
...
+    # The only field that is actually ordered is :name. See `man filesystems` on AIX
+    @fields = [:name, :account, :boot, :check, :dev, :free, :mount, :nodename,
+               :options, :quota, :size, :type, :vfs, :vol, :log]
+    self.line_separator = "\n"
+    # Override lines and use scan instead of split, because we DON'T want to
+    # remove the separators
+    def self.lines(text)
+      # Separate base on comments, blank lines, or a mountpoint followed by
+      # anything that isn't a mountpoint, comment, or blank line.
+      ret = text.scan(%r{(^\s*\*.*?$|^\s*?$|^\S+:(?:(?!(?:^\S+:|^\s*\*|^\s*$)).)*)}m).flatten
```
hunner: This should be turned into logic, not a scary regex.

helperton:
Ok, so this works great.  Could you make one cosmetic change and put in one blank before each entry you put in /etc/filesystems?

hunner:
@helperton Blank lines and comments are preserved across updates, so if you have blank lines before each entry, they should stay there. I'd rather not enforce blank lines before every entry, as some users may not prefer that style. Is that okay?

helperton:
I think it's a requirement by IBM

https://www.ibm.com/developerworks/community/wikis/home?lang=en#!/wiki/Power+Systems/page/How+to+change+order+of+fs+mounts+on+AIX+V5.3

Filesystems are mounted at AIX boot time in the order they appear in /etc/filesystems. Filesystems are added to /etc/filesystems in the order they are defined. The easiest way to change the order in which filesystems are mounted is to edit /etc/filesystems and change the ordering of the filesystem stanzas. (But be careful to maintain a blank line between each stanza. If blank lines are removed, some filesystems won't get mounted at boot time!)

kylog:
@helperton I read that three times and I'm still somewhat baffled. What do the blank lines signify? That advice says not to remove blank lines but not what they mean. It sounds like this is *not* a cosmetic change so I'm curious how it actually works.

helperton:
Ya, I thought it was cosmetic until I found that article.  I was suspicious that all of our /etc/filesystems have that blank line, then when I looked at examples of /etc/filesystems on the internet I found those too, have the blank line.  Then I found the IBM article.

helperton:
Here's another article which talks about the blank line (still doesn't answer your question).

http://bio.gsi.de/DOCS/AIX/2x/faqssoft.html

kylog:
I can't find anything about it, but the examples I found have a blank line between stanzas. So probably better safe than sorry ....

Also in the surprising-to-a-non-AIX-admin department:

> Note: The asterisk (*) is the comment character used in the /etc/filesystems file.

which I found in http://www-01.ibm.com/support/knowledgecenter/ssw_aix_71/com.ibm.aix.files/filesystems.htm

hunner:
@helperton Oooo, okay. I'll update the PR with blank lines then. Thanks!

hunner:
Newlines updated. The only issue I've seen still occurring is the `mount -o remount <nfs mountpoint>` causing failures after updating an NFS entry. It seems that it must be either `remount,ro` or `remount,rw` to work correctly.

The mount type previously just ran an unmount & mount instead of a remount, but this PR changes the default to do remount... How should this be solved?
- Change back to unmount/remount
- Leave with remount; this error is PEBKAC
- Make the option default to unmount/mount for NFS volumes and remount for non-NFS volumes
- Other

helperton:
Ok, so it did fix the newline issue, but we now seem to have some issue with the option ordering and it saying that it's updated the options, but it's not actually reflected in /etc/filesystems.

During the puppet run:

Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Info: Caching catalog for somehost.domain.com
Info: Applying configuration version '0cfe3f36a46cb4ddc81f790626ffbf793245f2d3'
Notice: /Stage[main]/Puppet_nfs/Mount[/nfs/infra]/options: options changed 'bg,intr,soft' to 'bg,soft,intr'
Info: Computing checksum on file /etc/filesystems
Info: /Stage[main]/Puppet_nfs/Mount[/nfs/infra]: Scheduling refresh of Mount[/nfs/infra]
Info: Mount[/nfs/infra](provider=parsed): Remounting
Notice: /Stage[main]/Puppet_nfs/Mount[/nfs/infra]: Triggered 'refresh' from 1 events
Info: /Stage[main]/Puppet_nfs/Mount[/nfs/infra]: Scheduling refresh of Mount[/nfs/infra]
Notice: /Stage[main]/Puppet_nfs/Mount[/nfs/depot/aix]/options: options changed 'bg,intr,soft' to 'bg,soft,intr'
Info: /Stage[main]/Puppet_nfs/Mount[/nfs/depot/aix]: Scheduling refresh of Mount[/nfs/depot/aix]
Info: Mount[/nfs/depot/aix](provider=parsed): Remounting
Notice: /Stage[main]/Puppet_nfs/Mount[/nfs/depot/aix]: Triggered 'refresh' from 1 events
Info: /Stage[main]/Puppet_nfs/Mount[/nfs/depot/aix]: Scheduling refresh of Mount[/nfs/depot/aix]
Notice: /Stage[main]/Puppet_nfs/Mount[/nfs/depot/oracle]/options: options changed 'bg,intr,soft' to 'bg,soft,intr'
Info: /Stage[main]/Puppet_nfs/Mount[/nfs/depot/oracle]: Scheduling refresh of Mount[/nfs/depot/oracle]
Info: Mount[/nfs/depot/oracle](provider=parsed): Remounting
Notice: /Stage[main]/Puppet_nfs/Mount[/nfs/depot/oracle]: Triggered 'refresh' from 1 events
Info: /Stage[main]/Puppet_nfs/Mount[/nfs/depot/oracle]: Scheduling refresh of Mount[/nfs/depot/oracle]
Notice: Finished catalog run in 1.17 seconds

It keeps saying that it's changing the options, but it doesn't actually change them in the file, so the next puppet run outputs the exact same thing again.

hunner:
@helperton Okay. I was sorting the list of options returned from /etc/filesystem (which is why it returned `bg,intr,soft` even though that wasn't the actual order). I removed that sort.

However, some "options" (account, boot, check, free, mount, size, type, vol, log, and quota) are passed to the mount type via the `options` parameter before being separated into their respective lines, and must be alphabetically sorted at the end of the comma-separated list to avoid this.

helperton:
Bravo!  Works.

stahnma:
After another review, I'm still happy with this. :+1:

melissa:
@hunner if this is going to get into Puppet 3.8, could you retarget this pull request against the 3.x branch? The stable branch now corresponds to 4.0 and master to 4.1.

PR #3741, opened 2015-03-20
(maint) Introduce VCR to the bundle, have nicer HTTP stubbing

ffrank:
This adds more lines than it removes, but most of them are in fixtures.

The code itself is currently about +/- 0, but future HTTP tests can be implemented with very little overhead.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -43,6 +43,13 @@ Pathname.glob("#{dir}/shared_behaviours/**/*.rb") do |behaviour|
   require behaviour.relative_path_from(Pathname.new(dir))
 end
 
+require 'vcr'
+VCR.configure do |vcr|
+  vcr.cassette_library_dir = File.expand_path('vcr/cassettes', PuppetSpec::FIXTURE_DIR)
+  vcr.hook_into :webmock
+  vcr.configure_rspec_metadata!
+end
+
 RSpec.configure do |config|
   include PuppetSpec::Fixtures
 
--- a/spec/unit/network/http/connection_spec.rb
+++ b/spec/unit/network/http/connection_spec.rb
@@ -105,6 +105,14 @@ describe Puppet::Network::HTTP::Connection do
     let (:host) { "my_server" }
     let (:port) { 8140 }
 
+    before :all do
+      WebMock.disable!
+    end
+
+    after :all do
+      WebMock.enable!
+    end
+
     it "should provide a useful error message when one is available and certificate validation fails", :unless => Puppet.features.microsoft_windows? do
       connection = Puppet::Network::HTTP::Connection.new(
         host, port,
--- a/spec/unit/network/http/connection_spec.rb
+++ b/spec/unit/network/http/connection_spec.rb
@@ -42,31 +42,18 @@ describe Puppet::Network::HTTP::Connection do
     end
   end
 
-  context "when methods that accept a block are called with a block" do
-    let (:host) { "my_server" }
+  context "when methods that accept a block are called with a block", :vcr do
+    let (:host) { "my-server" }
     let (:port) { 8140 }
     let (:subject) { Puppet::Network::HTTP::Connection.new(host, port, :use_ssl => false, :verify => Puppet::SSL::Validator.no_validator) }
 
-    before :each do
-      httpok.stubs(:body).returns ""
-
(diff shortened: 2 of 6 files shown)


ffrank:
Huh. Does Travis not care about `maint` branches anymore?

kylog:
Not sure what's up with travis. It just didn't trigger. Same with #3737.

I'm also not sure how to trigger a run on a PR that travis hasn't noticed. I'm going to try and close/reopen to see if that kicks it.

kylog:
That seemed to do it. I see it's pending in travis now.

ffrank:
Thanks @kylog.

spec/spec_helper.rb
```diff
@@ -43,6 +43,15 @@ module PuppetSpec
   require behaviour.relative_path_from(Pathname.new(dir))
 end
 
+PROJECT_ROOT = File.expand_path('..', File.dirname(__FILE__))
+
+require 'vcr'
+VCR.configure do |vcr|
+  vcr.cassette_library_dir = File.expand_path('spec/fixtures/vcr/cassettes', PROJECT_ROOT)
```
ffrank: Note to self: use `FIXTURE_DIR` from above.

peterhuene:
I'm marking this PR blocked until we decide what is going to happen to PR #3748.  We don't think that this change is necessary without moving forward on the other PR.

peterhuene:
Given @ahpook's approval on PR #3748, I'm removing the block status.  I'll take one more look at this tomorrow, but I think the time has finally come to merge.

kylog:
I see that this was originally inspired by use of vcr in r10k, and also that r10k removed use of vcr in https://github.com/puppetlabs/r10k/pull/414, so it'd be nice to see if there were lessons learned there that we should consider here.  /cc @adrienthebo

spec/unit/network/http/connection_spec.rb
```diff
@@ -42,31 +42,18 @@
     end
   end
 
-  context "when methods that accept a block are called with a block" do
-    let (:host) { "my_server" }
+  context "when methods that accept a block are called with a block", :vcr => true do
```
kylog: Per https://www.relishapp.com/vcr/vcr/v/2-9-3/docs/test-frameworks/usage-with-rspec-metadata, with rspec 3 this can be just `:vcr` rather than `:vcr => true`. (It's a bit of a nit, but it seems to be the pattern that shows up in the reference docs, so may seem less different to vcr newbies like myself who go poking around on relishapp.)

adrienthebo:
@kylog I removed VCR from r10k for several reasons. First off, VCR 2.9.3 doesn't support Faraday 0.9.x which meant that we couldn't record new cassettes without using a bunch of workarounds. Secondly, Faraday has a test adapter built in, which makes it very easy to test HTTP request/responses without having to manually stub out the entire world - VCR makes it a bit nicer but it's not necessary. Lastly VCR hasn't had a release in a while and currently needs maintainers, and the fact that VCR 2.9.4 wasn't getting released in a timely manner coupled with the fact that Faraday has a good testing system built in meant that it was easier for me to drop VCR.

For your case where you're testing with net/http VCR might be much more useful, and the problems that affected me may not affect you.There's nothing wrong with VCR so much as there are issues with VCR and Faraday.

ffrank:
Worth noting that I picked `WebMock` over Faraday because it looked easier.

Granted, I didn't put much research effort into the pros and cons of the different stubbing solutions that are available here. So far, `WebMock` looks perfectly adequate to me.

kylog:
To be clear, I wasn't advocating *against* vcr per se (I honestly haven't looked into this area at all), just noting that bit of history and wanted to foster some discussion around tradeoffs.

peterhuene:
For what it's worth, it does look like there are some new maintainers to VCR that are starting to pick development back up.  I'm :+1: with keeping VCR as we're not using Faraday in puppet currently.

adrienthebo:
Yep, if you're not using Faraday then VCR should be perfectly adequate. It's a really nice library and I am quite fond of it.

spec/unit/network/http/connection_spec.rb
ffrank: That's valuable feedback, thanks Kylo. I'm all :+1: for being more in keeping with common documentation.

PR #3742, opened 2015-03-20
(PUP-4194) Move `logdir` default to /var/log/puppetlabs/puppet

joshcooper:
Previously, puppet would set its `logdir` permissions to `root:root:750`
if run on a host without the `puppet` service user and group. If the
`puppetserver` package was installed later, it would create the `puppet`
user and group accounts, try to start, but fail to write to its log
directory `/var/log/puppetlabs/puppetserver`, because of the parent
directory permissions.

This commit moves puppet's logdir to `/var/log/puppetlabs/puppet`,
isolated from both `puppetserver` and `puppetmaster` packages, the latter
being the passenger-based compatibility packages.

This commit affects both the `puppetdlog` setting, which is pretty much
dead, and the puppet master webrick `masterhttplog` setting, which is
soon to be deprecated.

### DIFF
--- a/acceptance/tests/ensure_puppet-agent_paths.rb
+++ b/acceptance/tests/ensure_puppet-agent_paths.rb
@@ -26,7 +26,7 @@ def config_options(platform)
     codedir = '/etc/puppetlabs/code'
     confdir = '/etc/puppetlabs/puppet'
     vardir = '/opt/puppetlabs/puppet/cache'
-    logdir = '/var/log/puppetlabs'
+    logdir = '/var/log/puppetlabs/puppet'
     rundir = '/var/run/puppetlabs'
     sep = ":"
   end
--- a/ext/redhat/client.sysconfig
+++ b/ext/redhat/client.sysconfig
@@ -5,7 +5,7 @@
 #PUPPET_PORT=8140
 
 # Where to log to. Specify syslog to send log messages to the system log.
-#PUPPET_LOG=/var/log/puppet/puppet.log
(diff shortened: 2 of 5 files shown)


joshcooper:
Btw, I didn't change the following packaging related files:

```
0 ~/work/puppet (ticket/master/PUP-4194-relocate-puppet-logdir)  $ git grep '/var/log/puppet[^l]'
ext/debian/changelog.erb:    /var/log/puppet on purge (closes: #571130)
ext/debian/changelog.erb:  * Add puppetmaster.postrm to remove /var/log/puppet on purge
ext/debian/puppet-common.postrm:                rm -rf /var/log/puppet/
ext/ips/transforms:# make sure /var/log/puppet and /var/lib/puppet are owned by puppet
ext/redhat/client.sysconfig:#PUPPET_LOG=/var/log/puppet/puppet.log
ext/redhat/puppet.spec.erb:- Fix permissions on /var/log/puppet (#495096)
spec/unit/provider/service/src_spec.rb:myservice::--no-daemonize:/usr/sbin/puppetd:0:0:/dev/null:/var/log/puppet.log:/var/log/puppet.log:-O:-Q:-S:0:0:20:15:9:-d:20::"
```

joshcooper:
Supercedes PR #3710

joshcooper:
Updated `ext/redhat/client.sysconfig:#PUPPET_LOG=/var/log/puppetlabs/puppet/puppet.log` as that is added to puppet-agent packages.

joshcooper:
The acceptance test will fail until the corresponding packaging changes are made.

PR #3746, opened 2015-03-23
(PUP-4090) Zypper provider improvements

melissa:
Prior to this commit, our zypper provider didn't work. When installing a
package with zypper, we pass in the `--name` flag, which is not
available in earlier versions of zypper that we account for. It also
breaks when you use `install_options`. It previously dropped in the
extra install options between `--name` and the package name to be
installed. This cuased it to fail.

This pull request reorders the options that we pass to the zypper
command. If the user has passed in the global option `--no-gpg-check`,
we pull that flag out, and pass it in where zypper expects it (after
`zypper` and before `install`).

### DIFF
--- a/lib/puppet/provider/package/zypper.rb
+++ b/lib/puppet/provider/package/zypper.rb
@@ -35,10 +35,9 @@ Puppet::Type.type(:package).provide :zypper, :parent => :rpm do
     end
 
     #This has been tested with following zypper versions
-    #SLE 10.2: 0.6.104
-    #SLE 11.0: 1.0.8
-    #OpenSuse 10.2: 0.6.13
-    #OpenSuse 11.2: 1.2.8
+    #SLE 10.4: 0.6.201
+    #SLE 11.3: 1.6.307
+    #SLE 12.0: 1.11.14
     #Assume that this will work on newer zypper versions
 
     #extract version numbers and convert to integers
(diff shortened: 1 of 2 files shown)


haus:
@melissa would it be worth adding rspec tests that exercise the allow_virtual => true and maj > 1 combination? I don't see that present in current testing.

melissa:
@haus updated.

haus:
This looks sane to me. I can't speak to zypper itself, but the tests and change make sense.

shrug:
:+1:

melissa:
@kylog if you want to take a look at this, that would be awesome. It should get in once the master branch has been unfrozen

shrug:
Still :+1:

melissa:
Closed in favor of https://github.com/puppetlabs/puppet/pull/3774 targeted at puppet/3.x

melissa:
Since this also needs to be in Puppet 4.1.0, I've reopened this PR.

haus:
@melissa are we not merging up from 3.x to stable/master?

melissa:
I honestly have no idea. From earlier discussion, it seemed like we weren't. Even if we are though, this should be safe to merge, since the commit has been cherry-picked.

melissa:
just got confirmation this will be merged up, so will be available in 4.x

PR #3747, opened 2015-03-24
[DO NOT MERGE](PUP-1291) Scheduled Tasks Support for Repeat Triggers

ferventcoder:
Previously, the scheduled_task resource did not support repetition for anything
less than 1 day. This is limiting as the cron resource can support repetition
in minutes. Task scheduler on Windows can support repeating tasks every x
number of minutes, and the win32-taskscheduler gem also allows this.

Allow Puppet's scheduled_task resource to support repetition every x number of
minutes. Without this change the schedule resource continues to be more limited
than on other platforms.

This supersedes #3637

### DIFF
--- a/lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
+++ b/lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
@@ -76,7 +76,6 @@ Puppet::Type.type(:scheduled_task).provide(:win32_taskscheduler) do
                   nil
                 end
       next unless trigger and scheduler_trigger_types.include?(trigger['trigger_type'])
-
       puppet_trigger = {}
       case trigger['trigger_type']
       when Win32::TaskScheduler::TASK_TIME_TRIGGER_DAILY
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -239,6 +239,8 @@ def triggers_same?(current_trigger, desired_trigger)
     desired['months']      ||= current_trigger['months']      if current_trigger.has_key?('months')
     desired['on']          ||= current_trigger['on']          if current_trigger.has_key?('on')
     desired['day_of_week'] ||= current_trigger['day_of_week'] if current_trigger.has_key?('day_of_week')
+    desired['minutes_interval'] ||= current_trigger['minutes_interval'] if current_trigger.has_key?('minutes_interval')
+    desired['minutes_duration'] ||= current_trigger['minutes_duration'] if current_trigger.has_key?('minutes_duration')
```
ferventcoder: @mrdima this is a bit hard to follow for me. Does the if apply for the entire line of code or just `current_trigger['minutes_duration']`?

mrdima: I just copied those lines from the above lines (desired['months'] as they seem to have been working fine. In short, the if statement only applies to the current_trigger['minutes_duration'] and not the whole line. It should mean something like:
if desired['minutes_interval'] is unset, make it equal to the current_trigger['minutes_interval'] if the current_trigger['minutes_interval'] exists.

ferventcoder: Pretty sure it is the entire line.

PR #3372, opened 2014-12-03
WIP: (PUP-1072) accept HTTP URLs as file source and retrieve file data via HTTP

ffrank:
Opening for early review.

I would have liked to add new indirector termini, but there are issues. During the very creation of the indirector [request](https://github.com/puppetlabs/puppet/blob/6df5fd9250affd62a2d18fc2cfb8bafb5a144f92/lib/puppet/indirector/request.rb#L75-84), URLs get special treatment. The indirector assumes that they are in a format that complies with the REST API.

While it would certainly be possible to teach the indirector to accept these file URLs as is, it would add complexity that is currently required for only this one obscure use case. I don't think we should.

PR #3748, opened 2015-03-25
(PUP-1072) HTTP(S) support for the file type

ffrank:
Uses two commits from #3741 which will hopefully get merged first. These `maint` commits will vanish upon rebasing then.


lib/puppet/type/file/content.rb
```diff
@@ -220,6 +221,19 @@ def get_from_source(source_or_content, &block)
       end
     end
 
+    def get_from_http_source(source, &block)
```
branan: I feel like I'd rather see this go through the indirector and avoid having the file type require special knowledge of how files are gotten. What prevents that from working here?

melissa:
@cprice404 it looks like you did some review for the previous iteration of this PR. Do you have some time to take a look at this one as well?

lib/puppet/type/file/content.rb
ffrank: This is the equivalent to the pre-existing `get_from_source` (now `get_from_puppet_source`) above, which will bypass the indirector for `puppet://` URLs.

The reason for this is optimizing the handling of large files (shoutout to @masterzen for pointing this out to me in the old PR). These methods here take a block and stream the contents. The indirector, on the other hand, will store the whole file content in a variable and return it for processing. This implies considerable memory consumption for large files.

ffrank:
To perhaps save @cprice404 some time, here's the [summary](https://github.com/puppetlabs/puppet/pull/3372#issuecomment-76295520) from the old PR #3372:

> I'm confident that we can solve all of @cprice404 's comments, except of course for the static compiler bit. That is really interesting. As a cop-out, I hope we can refuse static-compiling manifests that make use of this feature before implementing actual support. Thoughts anyone?

I had replied to all of Chris's comments, I believe. This doesn't necessarily mean that they are all resolved. It's likely a good idea to review them. They were quite thoughtful :)

lib/puppet/type/file/content.rb
branan: ah, that makes sense. Extending the indirector to handle streaming results is definitely outside the scope of this PR

melissa:
@cprice404, @kylog, @ahpook, @joshcooper, @ffrank, we're going to try to focus on this pull request during the Community PR Triage meeting tomorrow. Your input here would be greatly valued.

cprice404:
@melissa I think other folks are more up-to-date on this PR than I am at this point.  My general stance remains the same: I think this is a desirable (perhaps necessary) feature, but I'm worried about what it will mean for upcoming goals that we have related to static compilation.  In an ideal world, I think it would be better to wait until we'd done some thorough design work on that before we moved forward with this.  I'm not entirely sure how easy it will be (or how tolerable of a UX it will be) to try to preclude this functionality when we're running in a static-compile mode, even as an interim solution until we figure out how this should fit in the long-term plans.  My concern is elevated a bit by the fact that I keep getting the impression that moving to a 100%-static-compilation world is really an implicit requirement of a lot of the features that seem to be appearing on mid-to-long-term roadmaps.

That said, I know that this is a very useful feature and has been oft requested, and I'm extremely grateful to @ffrank for all the hard work on it... that, combined with the fact that we're probably not going to get to the requisite design work that I'm alluding to for several months yet, makes me reticent to continue to be a bottleneck :)

At the end of the day, @ahpook is probably going to have to make the call :)

lib/puppet/file_serving/terminus_helper.rb
```diff
@@ -29,4 +29,16 @@ def path2instances(request, *paths)
       path2instance(request, base_path, :ignore_source_permissions => true, :relative_path => file)
     end
   end
+
+  def unescape_url(url)
```
MikaelSmith: Why the inconsistency in defining escape_url and unescape_url?

lib/puppet/file_serving/http_metadata.rb
```diff
@@ -0,0 +1,29 @@
...
+  def initialize(http_response, path = '/dev/null')
+    super(path)
+
+    # ignore options that do not apply to HTTP metadata
+    @owner = @group = @mode = nil
+
+    if checksum = http_response['content-md5']
+      # convert base64 digest to hex
+      checksum = checksum.unpack("m0").first.unpack("H*").first
+      @checksum_type = 'md5'
```
MikaelSmith: If the checksum type doesn't match the system config or resource's checksum type, this will always resync the resource. That's not a great behavior, but I don't immediately have a suggestion how to fix it.

lib/puppet/type/file/content.rb
```diff
@@ -220,6 +221,19 @@ def get_from_source(source_or_content, &block)
...
+      Puppet::Util::HttpProxy.request_with_redirects(URI(source), :get, &block)
```
MikaelSmith: No plan to provide a way to override the redirects limit?

lib/puppet/file_serving/http_metadata.rb
```diff
@@ -0,0 +1,29 @@
...
+      @checksum = "{md5}#{checksum}"
+    elsif mtime = DateTime.httpdate(http_response['last-modified']).to_time
```
MikaelSmith: This check fails if you try to source a directory-like page, such as http://nightlies.puppetlabs.com/facter-latest/repo_configs/rpm/. `http_response['last-modified']` returns `nil`, which causes `DateTime` throw an exception.

Once you check whether `http_response['last-modified']` is `nil`, you find that `PuppetError` is not defined, so spec testing isn't hitting it.

MikaelSmith: Also, it might be worth dumping http responses like we do in other places when the `http_debug` feature is enabled.

lib/puppet/file_serving/http_metadata.rb
```diff
@@ -0,0 +1,29 @@
...
+      @checksum_type = 'mtime'
+      @checksum = "{mtime}#{mtime}"
+    else
+      raise PuppetError, "HTTP response contained no usable checksum equivalent"
+    end
+
+    @ftype = 'file'
```
MikaelSmith: I lean towards preferring that all other types are invalid, rather than hard-coding the type.

lib/puppet/file_serving/http_metadata.rb
ffrank: Yes, this is rather terrible. I was thinking of raising a followup issue to allow file resources to pick their checksum type dynamically based on what the indirector hands back. I.e., just do the right thing here, unless the user specified something else.

lib/puppet/file_serving/http_metadata.rb
ffrank: Exceptions: Good point, code should be less deflated.

`http_debug` - also good point. Wasn't aware of that at all. Will look into it.

lib/puppet/file_serving/http_metadata.rb
ffrank: This was a question of semantics to me.

A HTTP source can only ever offer a flat file (unless we implement a spider in Puppet that can download directory trees), never a link or a directory.

lib/puppet/type/file/content.rb
ffrank: Come to think of it, it would appear that I should just rely on the [option](https://github.com/puppetlabs/puppet/blob/a9a0e9abdee14931966fe036fc723062dca66dd5/lib/puppet/network/http/connection.rb#L59) for that in the first place, and not expose it as a parameter at all.

lib/puppet/file_serving/http_metadata.rb
MikaelSmith: It's that "unless we implemented x" that I'd like to avoid breaking or complicating in the future. I'd rather prevent anyone ever saying `ensure => directory`, rather than having it silently interpret that as `file` - until it doesn't.

lib/puppet/type/file/content.rb
MikaelSmith: That looks like it could require some plumbing, but :+1:

lib/puppet/file_serving/http_metadata.rb
MikaelSmith: This factors into how we might make a better extension point for creating modules to extend the File type for arbitrary URIs, so I'll try to spend some time on figuring out how to do it better.

lib/puppet/file_serving/http_metadata.rb
ffrank: I see what you mean now. But I believe you are mistaken.

This piece of code determines what the indirector reports to find upstream. For all intents and purposes, this is invariably a flat `file` here. Even if a manifest specifies `ensure => 'directory'`, the agent will sync a flat file. This is in keeping with the behavior when using `puppet://` or `file://` URLs.

Even if it would make sense to error out here, there is no way to catch that. The indirector must report the type of the entity it finds under the given key. It does not know about the specifics of the file resource on the other end.

MikaelSmith: Ok, you're right that I was mistaken. That makes sense.

ffrank: @MikaelSmith is this reasoning sensible to you?

lib/puppet/file_serving/http_metadata.rb
ffrank: Oh neat - I just realized that it supports `http_debug` [already](https://github.com/puppetlabs/puppet/pull/3748/files#diff-cc443622f47b7470d9bd7e356e32eb43R90)

lib/puppet/type/file/content.rb
ffrank: I think I see what you mean. Turns out that is no agent option after all. In fact, I disbelieve that there is *any* code currently in place to override this for any HTTP request.

I believe that this is fine. The above line can be patched later to pass either an agent option or a resource parameter value to the method to pick a custom limit.

lib/puppet/file_serving/terminus_helper.rb
ffrank: Yikes! Adressed in the upcoming rebase.

ffrank:
PR note to self:

 * rebase, resolve `no_proxy` conflict
 * double check the `url=` prefix logic, i.e. need for that

lib/puppet/file_serving/terminus_helper.rb
```diff
@@ -29,4 +29,16 @@ def path2instances(request, *paths)
...
+  def self.unescape_url(url)
+    url.sub(/^url=/,'')
+  end
+
+  def self.escape_url(url)
+    if url =~ /^https?:/
+      "url=#{url}"
+    else
+      url
+    end
```
joshcooper: Adding a comment from the PR triage. Since puppet's source parameter already supports other URL types, e.g. file, puppet, it would ideally not be necessary to wrap/unwrap http as a special case. It would be good to understand where puppet goes haywire when given an http URL.

ffrank: Ack. So here's the full story:

    $ bundle exec puppet apply -e 'file { "/tmp/httptest": source => "http://cdn.sstatic.net/stackoverflow/img/newsletter-ad.png" }' --trace
    Notice: Compiled catalog for geras.fritz.box in environment production in 1.28 seconds
    Error: /Stage[main]/Main/File[/tmp/httptest]: Could not evaluate: Could not retrieve file metadata for http://cdn.sstatic.net/stackoverflow/img/newsletter-ad.png: Could not find a directory environment named 'stackoverflow' anywhere in the path: /home/ffrank/.puppetlabs/etc/code/environments. Does the directory exist?

This is the top of the stack:

    /home/ffrank/git/puppet/lib/puppet/environments.rb:38:in `get!'
    /home/ffrank/git/puppet/lib/puppet/indirector/request.rb:40:in `environment='
    /home/ffrank/git/puppet/lib/puppet/indirector/request.rb:255:in `set_uri_key'
    /home/ffrank/git/puppet/lib/puppet/indirector/request.rb:81:in `initialize'

The indirector tries to analyze the URI [here](https://github.com/puppetlabs/puppet/blob/6df5fd9250affd62a2d18fc2cfb8bafb5a144f92/lib/puppet/indirector/request.rb#L253). `file://` and `puppet://` URL do not suffer this issue because there is special handling for those just above that line.

I believe that the indirector API is too simple for us to do a clean implementation of "special" URIs that should not be treated this way (i.e., we have no way to tell the indirector that one request is different from another in this way).

joshcooper: Ah so the line https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/request.rb#L253 is completely wrong! We no longer send the environment in the URI path, it's now sent as a query parameter.

I suspect you'd have to do something like https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/request.rb#L248-L251 to short-circuit evaluation for http(s) protocols.

Ideally puppet should not mutate the request key like the way it does. I'd prefer to see the request preserve the original URL, and extract the parts it needs, e.g. 

```
def protocol
  @uri.scheme
end
```

That would be more work than in this PR though.

ffrank: Hmm, the mutation looks kind of neat. But that's just me :)

So if the indicated line is broken, does that mean that this code path is not in active use by anything anymore? I kind of feel like just replacing it with a "don't touch HTTP URLs" behavior.

ffrank:
I just pushed a commit that optimistically drops the logic that @joshcooper has identified as obsolete. In the best case, nothing is using that anymore.

FWIW, the `indrector` integration tests still seem to pass for one. I'll let Travis sort out the rest.
#copout

If this looks good, I'll open a separate maint PR to clean this up first with more review.

ffrank:
Apparently all spec failures originate with `Puppet::FileBucket::Dipper`. I'll dig around and try to see whether these tests are still relevant to actual operation.

ffrank:
Whoo, it's working without funky URL prefixing shenanigans.

I'm breaking [a test](https://github.com/puppetlabs/puppet/blob/c7b1fdb63497015f0ebf7b894a33a08ef3a96d0d/spec/unit/indirector/request_spec.rb#L139-141) that is spurious now, right @joshcooper ?

peterhuene:
@ffrank Yes, that error is a known failure and can be ignored.

peterhuene:
@ahpook We've discussed this a few times during PR triage and it seems that there is general consensus that this would be a useful addition, but we're not sure if the product team would like us to move forward on this.  Would you mind looking things over and making a call on whether or not we should proceed on this?  If so, we will merge the PRs that this PR depends on and Felix will clean up this PR to make it easier to review for next week's triage.  Thanks!

ffrank:
We were greenlit on Jira \o/

Something occured to me in the meantime. I should mimic the original semantics of the `file` type's `checksum` parameter. This is now [done](https://github.com/ffrank/puppet/commit/a6bc53716d14f76f0996d824f6d09856f03f4248). Doesn't even need squashing, really, I feel. Should have a test for the exception, I guess.

peterhuene:
This is now unblocked, but final review will take place once #3741 and #3910 are merged and this has been rebased and squashed.

kylog:
Wow, this PR is finally unblocked! Woot!

MikaelSmith:
Merged #3741

joshcooper:
@ffrank both dependent PRs have been merged and this is ready for a rebase.

joshcooper:
PR Triage Notes

* @joshcooper and @peterhuene to give one last review
* Pull into next sprint and merge

ffrank:
...aaand it broke again during rebase. So tired

ffrank:
It turns out that it was actually broken even before the rebase. Should be fixed now.

lib/puppet/file_serving/http_metadata.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    self
+  end
+
+  # Override of the parent class method. Does not call super!
+  # We can only return metadata that was extracted from the
+  # HTTP headers during #initialize.
+  def collect
+    # Prepare a fallback if no actual sums were among the headers
+    @checksums[:none] = none(:content => '')
```
MikaelSmith: This seems like it results in an empty file, as in the manifest
```
file { 'facter.zip':
    ensure => file,
    source => 'https://github.com/puppetlabs/facter/archive/master.zip',
    path   => '/tmp/facter.zip',
}
```

MikaelSmith: Maybe we don't follow redirects?

MikaelSmith: Rather than using `:none`, you could always set an mtime hash, using `DateTime.now` if none provided by the server. That would force it to always re-download.

spec/unit/file_serving/http_metadata_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+require 'digest'
+
+describe Puppet::FileServing::HttpMetadata do
+  let(:foobar) { File.expand_path('/foo/bar') }
+
+  it "should be a subclass of Metadata" do
+    expect( described_class.superclass ).to be Puppet::FileServing::Metadata
+  end
+
+  describe "when initializing" do
```
MikaelSmith: Should add a test when the server returns neither `Content-MD5` nor `Last-Modified`.

MikaelSmith: It might be helpful to have an integration test that confirms `sync` is called for http resources.

MikaelSmith:
The commit message needs updating.

ffrank:
I settled for following your other proposal and use `:mtime` with `Time.now` as a default. That seems much cleaner. Yes, the commit message will need updating.

MikaelSmith:
@ffrank This PR is waiting on you to update the commit message. Think you'll get to that soon?

ffrank:
I have, but not yet pushed. I'm still pondering how to go about [better test coverage](https://github.com/puppetlabs/puppet/pull/3748#discussion_r35054020).

MikaelSmith:
Ah, forgot about extending tests. Thanks for the reminder!

spec/unit/file_serving/http_metadata_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      let(:time) { Time.now.utc }
+      before do
+        http_response.stubs(:[]).with('content-md5').returns nil
+      end
+
+      it "should use :mtime as the checksum type, based on Last-Modified" do
+        # HTTP uses "GMT" not "UTC"
+        http_response.stubs(:[]).with('last-modified').returns(time.strftime("%a, %d %b %Y %T GMT"))
+        metadata = described_class.new(http_response)
+	metadata.collect
```
MikaelSmith: This is using tabs, would be nice to change to spaces.

spec/unit/file_serving/http_metadata_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      let(:base64) { Digest::MD5.new.base64digest input }
+      let(:hex) { Digest::MD5.new.hexdigest input }
+      before do
+        http_response.stubs(:[]).with('last-modified').returns nil
+        http_response.stubs(:[]).with('content-md5').returns base64
+      end
+
+      it "should use the md5 checksum" do
+        metadata = described_class.new(http_response)
+	metadata.collect
```
MikaelSmith: This is using tabs, would be nice to change to spaces.

MikaelSmith:
I'm working on a set of suggested spec tests that I think will cover the areas that need more coverage.

lib/puppet/indirector/file_content/http.rb
```diff
@@ -0,0 +1,15 @@
+require 'puppet/file_serving/metadata'
+require 'puppet/indirector/generic_http'
+
+class Puppet::Indirector::FileContent::Http < Puppet::Indirector::GenericHttp
```
MikaelSmith: I don't see this class ever used or invoked. Do you think it's still needed?

MikaelSmith:
Found what I think is another bug
```
file { '/tmp/README.md':
  ensure => file,
  source => 'https://raw.githubusercontent.com/puppetlabs/facter/master/README.md',
  backup => false,
}
```
When run using an agent/master, produces
```
Error: Could not set 'file' on ensure: incorrect header check at 3:/Users/michaelsmith/puppetlabs/puppet/master/code/environments/production/manifests/site.pp
Error: Could not set 'file' on ensure: incorrect header check at 3:/Users/michaelsmith/puppetlabs/puppet/master/code/environments/production/manifests/site.pp
Wrapped exception:
incorrect header check
Error: /Stage[main]/Main/File[/tmp/README.md]/ensure: change from absent to file failed: Could not set 'file' on ensure: incorrect header check at 3:/Users/michaelsmith/puppetlabs/puppet/master/code/environments/production/manifests/site.pp
```

MikaelSmith:
I put some ideas for spec tests in https://github.com/MikaelSmith/puppet/commit/045eed3a2d0c584ba90340ee57b651c91b1a14a5. Still working on understanding the bug I mentioned above.

ffrank:
I will look at the issue next chance I get - likely not before PuppetConf.

The patches look very good, do you feel that I could likely just pull them into this branch, @MikaelSmith ?

MikaelSmith:
I think so, minus the `byebug` line.

spec/unit/file_serving/http_metadata_spec.rb
ffrank: Fixed now.

ffrank:
It's weird, I currently cannot reproduce the bug described [above](https://github.com/puppetlabs/puppet/pull/3748#issuecomment-136860306).

It Just Works with both Ruby `2.1` and `1.9` apparently.

MikaelSmith:
I'll give it another try. It may be time to merge this and fix up bugs as they're discovered.

spec/integration/type/file_spec.rb
```diff
@@ -1260,6 +1260,27 @@ def build_path(dir)
     end
 
     let(:source) { tmpfile_with_contents("source_default_values", "yay") }
+
```
MikaelSmith: Didn't care for my spec tests at https://github.com/MikaelSmith/puppet/commit/045eed3a2d0c584ba90340ee57b651c91b1a14a5#diff-1a4c70e8a84e98497af8fcdbc8403474R1283?

ffrank:
Just pushed that @MikaelSmith thanks again. Will likely pull the headers fix as well, although I'd really like to reproduce as well. I'll keep you posted.

ffrank:
I just took the reverse approach and tried to find out more about the error. This [SO question](http://stackoverflow.com/questions/31284988/zlibdataerror-incorrect-header-check) looks suspicious. The source of the error message does seem to be zlib, at least that's what a quick skimming of the Google results suggest.

What strikes me about the SO question is that the issue is directly caused by an attempt to `Base64.decode`, which is what my new code does **if** a `Content-MD5` header is found. I wonder why it's popping up here now.

ffrank:
Believe it or not, but I think we're finally good.

What's happened:
 * pulled in the spec tests that @MikaelSmith contributed
 * reworked those to match current behavior
 * amended them to cover some more scenarios
 * pulled in Mikael's fix for the `incorrect header check` bug

I have consistently failed to reproduce that issue ever since PuppetConf. Seriously, no idea what's up. Perhaps I should :fireworks: download a fresh bundle in another directory.

Anyway, I chose to just use it as "the right thing to do" and not bother with a test. Do we need a test? I say we don't need a test. :-)

MikaelSmith:
So close, Windows failure. I can dig in Monday.

ffrank:
Thanks! That is a weird failure. Conceivable though - Windows has not seen any smoke testing yet.

spec/integration/type/file_spec.rb
```diff
@@ -1260,6 +1260,102 @@ def build_path(dir)
...
+        it "should update the file if content differs on disk" do
+          File.open(httppath, "w") { |f| f.puts "Content originally on disk\n" }
+          catalog.add_resource resource
+          catalog.apply
+          expect(Puppet::FileSystem.exist?(httppath)).to be_truthy
+          expect(File.read(httppath)).to eq "Content via HTTP\n"
+        end
+
+        it "should not update the file if content on disk is up-to-date" do
+          File.open(httppath, "w") { |f| f.puts "Content via HTTP\n" }
```
MikaelSmith: Use "wb" instead of "w" for this and those above. That fixes the issue on Windows.

ffrank: That makes sense. Thanks once more to @MikaelSmith!

MikaelSmith:
Travis is unhappy with your last commit message. I'm hopeful Windows specs will pass though. If they do we can squash the changes up.

kylog:
confetti1

ahpook:
PARTY

petems:
YESSSSSSSSSSSS

PR #3757, opened 2015-03-26

ferventcoder:

Supersedes #3722

### DIFF
--- a/spec/unit/file_system_spec.rb
+++ b/spec/unit/file_system_spec.rb
@@ -5,6 +5,18 @@ require 'puppet/util/platform'
 describe "Puppet::FileSystem" do
   include PuppetSpec::Files
 
+  def with_file_content(content)
+    path = tmpfile('file-system')
+    file = File.new(path, 'wb')
+    file.sync = true
+    file.print content
+
+    yield path
+
+  ensure
+    file.close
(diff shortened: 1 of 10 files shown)


ferventcoder:
Supersedes #3722 - switched due to new branch target

Iristyle:
Merged to master in https://github.com/puppetlabs/puppet/commit/a7b5072437a0a30e067d16d86144a44b293b4d91

spec/unit/file_system_spec.rb
```diff
@@ -5,6 +5,18 @@
...
+  def with_file_content(content)
+    path = tmpfile('file-system')
+    file = File.new(path, 'wb')
+    file.sync = true
+    file.print content
+
+    yield path
+
+  ensure
+    file.close
```
joshcooper: One thing to watch out for is that `file` could be nil, e.g. if `tmpfile` or `File.new` raise an error. I generally nest the begin/rescue at the point where you know you have a valid file to close, e.g.

```
path = tmpfile('file-system')
file = File.new(path, 'wb')
begin
  ...
ensure
  file.close
end
```

ferventcoder: Right on.

PR #3759, opened 2015-03-26
(PUP-4122) Update future_parser? to accept environment

hkenney:
Prior to this commit the Puppet.future_parser? method only worked
with the current environment. So that we can do validation around
the parser setting in puppet preview, update the method so it can
accept an environment as an argument and check the parser setting
for that specific environment.


hkenney:
Closing since this change is no longer necessary for preview

PR #3762, opened 2015-03-29
(PUP-4277) Make case/selector match be recursive

hlindberg:


lib/puppet/pops/evaluator/compare_operator.rb
```diff
@@ -149,4 +155,64 @@ def include_Array(a, b, scope)
...
+# This is the type of matching performed in a case option, using == for every type
+# of value except regular expression where a match is performed.
+#
+def is_match? left, right, o, scope
+  if right.is_a?(Regexp)
+    return false unless left.is_a? String
+    matched = right.match(left)
+    set_match_data(matched, scope) # creates or clears ephemeral
+    !!matched # convert to boolean
+  elsif right.is_a?(Puppet::Pops::Types::PAnyType)
```
thallgren: The above four lines can be replaced with a call to match_Regexp(right, left, scope)

lib/puppet/pops/evaluator/compare_operator.rb
```diff
@@ -149,4 +155,64 @@ def include_Array(a, b, scope)
...
+    # right is a type and left is not - check if left is an instance of the given type
+    # (The reverse is not terribly meaningful - computing which of the case options that first produces
+    # an instance of a given type).
+    #
+    @@type_calculator.instance?(right, left)
+  else
```
thallgren: Same here. Call `match_PAnyType(right, left, scope)` instead of repeating its comment and implementation.

PR #3764, opened 2015-03-29
(PUP-4336) Fix issues with trace and structured logging

thallgren:
This commit ensures that trace is output correctly for errors
that origins from the future parser. It also adds four new issues
related to the lexer and ensures that the errors and warnings related
to them are generated using the improved structured logging.

### DIFF
--- a/lib/puppet/pops/issues.rb
+++ b/lib/puppet/pops/issues.rb
@@ -594,10 +594,6 @@ module Puppet::Pops::Issues
     "Unrecognized escape sequence '\\#{ch}'"
   end
 
-  UNRECOGNIZED_QUOTE = hard_issue :UNRECOGNIZED_QUOTE, :followed_by do
-    "Unclosed quote after \"'\" followed by '#{followed_by}'"
-  end
-
   UNCLOSED_QUOTE = hard_issue :UNCLOSED_QUOTE, :after, :followed_by do
     "Unclosed quote after #{after} followed by '#{followed_by}'"
   end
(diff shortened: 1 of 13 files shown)


lib/puppet/pops/issues.rb
```diff
@@ -545,4 +545,21 @@ def self.hard_issue(issue_code, *args, &block)
...
     "Unfolding of attributes from Hash can only be used once per resource body"
   end
+
+  SYNTAX_ERROR = hard_issue :SYNTAX_ERROR, :where do
+    "Syntax error at #{where}"
+  end
+
+  LEX_ERROR = hard_issue :LEX_ERROR do
+    # Error here for completeness. It's never printed
+    "Unable to create lexical token stream"
```
hlindberg: So, if this is never used, why does it have the message text "Unable to create lexical token stream" ?? Why does it not say something like "Internal Error: LEX_ERROR used the wrong way" ?

thallgren: Matter of preference I guess. I chose a message that a) makes sense in the context where it's found, and b) is descriptive for the issue.

thallgren: A better solution would probably be to not have a LEX_ERROR at all but instead have a unique issue for every problem that can occur.

hlindberg: Yes, that is much better - and removes the error messages from the source in preparation of I18N. It also provides better feedback in structured output as the code is more meaningful. There are not that many different lex errors, so worth doing.

hlindberg:
Since none of the existing tests caught the regression, it is worth adding a couple integration tests for the structured and unstructured cases. Say one each for lex error, syntax error, and validation error. The "puppet apply --trace" case is more difficult - I think that requires an acceptance test (ping @er0ck - separate ticket?)

hlindberg:
See comments on the commit cd7033d

er0ck:
@hlindberg i think, if it's high-risk we should sub-task and create the acceptance and the ticket doesn't fully resolve without it.  this way it's done quickly, and the ticket doesn't resolve without full testing.  the regression implies high-risk.

hlindberg:
@er0ck yes, an acceptance test that `puppet apply -e 'something that errors' --trace` errors with a backtrace in the log is useful. Agree a Sub Task to PUP 4336 is suitable for writing an acceptance test.

PR #3765, opened 2015-03-30
(PUP-4362) fix portage to list all installed packages

chrigl:
In commit 9e7bf07580d79e5b68d28cc594162959cf2a8d7d, portage.rb was
refactored, but from this on, puppet resource package only lists
packages installed by puppet, but no packages installed by hand.
The other providers apt/dpkg and yum/rpm list _all_ installed packages.
The format string in portage.rb:L123 misses a tailing \n.

### DIFF
--- a/lib/puppet/provider/package/portage.rb
+++ b/lib/puppet/provider/package/portage.rb
@@ -132,7 +132,7 @@ Puppet::Type.type(:package).provide :portage, :parent => Puppet::Provider::Packa
 
   private
   def self.eix_search_format
-    "'<category> <name> [<installedversions:LASTVERSION>] [<bestversion:LASTVERSION>] <homepage> <description>'"
+    "'<category> <name> [<installedversions:LASTVERSION>] [<bestversion:LASTVERSION>] <homepage> <description>\n'"
   end
 
   def self.eix_result_format


chrigl:
Although this is a trivial patch, i signed the CLA

melissa:
@chrigl thanks for the contribution! Also, thank you for signing the CLA even though it's a trivial patch. That is definitely something that's good to have.

In your commit message you say that `puppet resource package` no longer lists all installed packages the update made in https://github.com/puppetlabs/puppet/commit/9e7bf07580d79e5b68d28cc594162959cf2a8d7d. Does the change in your commit fix that issue? Or is it for a separate issue? If it does fix it, could you expand on how your change fixes the problem? I'd also say, if it does fix the problem, this isn't necessarily a trivial change. If this causes `puppet resource package` to list all installed packages, not just those installed with puppet, this should have a ticket associated with it.

You can find more information about contributing guidelines at https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md

chrigl:
Hi @melissa.

This commit fixes the issue of `puppet resource package` not listing all the gentoo packages.
I opened PUP-4362 for this issue.

branan:
Thanks for your contribution. It's been a while since I've setup a Gentoo box, but I'll try to get a VM spun up this week so I can validate this PR.

branan:
:+1:

PR #3767, opened 2015-03-30
(PUP-4330) Vendor plist into Puppet from Facter

MikaelSmith:
We no longer use the plist utility in Facter, but several Puppet providers use it. Vendor the code into puppet/vendor for use in the providers.

Also reapply patch https://github.com/puppetlabs/facter/commit/b6a0a65a34192a1e8b2755e528cbffd7d00a251c from Facter.


MikaelSmith:
Following the vendoring guidelines was slightly tricky, because several untracked patches had been applied in the Facter repo.

MikaelSmith:
The Ruby 1.9 fix could also be done by updating the commit SHA we track. The change to use Emit.dump hasn't been made on the source.

lib/puppet/vendor/plist/README
```diff
@@ -0,0 +1,36 @@
...
+Other folks who have helped along the way:
+
+[<b>Martin Dittus</b>] who pointed out that +Time+ wasn't enough for plist <tt>Dates</tt>, especially those in <tt>~/Library/Cookies/Cookies.plist</tt>
+[<b>Chuck Remes</b>] who pushed Patrick towards implementing <tt>#to_plist</tt>
+[<b>Mat Schaffer</b>] who supplied code and test cases for <tt><data></tt> elements
+[<b>Michael Granger</b>] for encouragement and help
+
+== License and Copyright
+
+plist is released under the MIT License.
```
joshcooper: Can we vendor MIT given that puppet is ASL 2.0 licensed? /cc @haus

MikaelSmith: We already have for the last 8 years, so that's a fun time if not. See https://github.com/puppetlabs/facter/blob/46d9bedc8b3f894fc8fab64955775daf94211814/lib/facter/util/plist.rb#L6.

branan: I've always understood the answer to be "yes", and a quick google agrees with me. Would be nice to get a confirmation from RE

joshcooper: ha, yeah, good point @MikaelSmith!

stahnma: Yes. Just keep the MIT license bits in tact where you vendor.

haus: As long as we leave the license intact in tree and files, this is fine.

kylog:
Can you amend the commit that came from fact-81 to reference pup-4330? That will be friendlier on the ticketmatch.rb script?

Other than that :+1:

MikaelSmith:
Done.

PR #3768, opened 2015-03-30
(PUP-4239) add array and hash types to lookup acceptance

er0ck:
This change adds merged arrays and hashes to the lookup acceptance test
from across data providers.
This implicitly tests including classes via lookup from arrays and
hashes and is not interesting to test at this layer.


thallgren:
All values used for 'unique' lookups are strings. This is OK and such tests are needed but perhaps an additional test is needed where the original values are arrays so that we verify that the arrays are concatenated and the result is unique.

er0ck:
@thallgren good idea.  will update

er0ck:
@thallgren done.

PR #3769, opened 2015-03-30
(PUP-1628) Fix mount provider for AIX

hunner:
The way that AIX tracks its mounted filesystems is different than other
unix or linux operatingsystems. AIX uses /etc/filesystems and has
entirely different filesystem options.

This commit takes advantage of two parsedfile hooks, post_parse and
to_line, to read & generate /etc/filesystem entries. Parsedfile was
not created with arbitrary-ordering in mind, and any /etc/filesystem
entries will have their attributes reordered by to_line

Options in AIX without related properties in Puppet's mount type are
placed under the "options" property.

### DIFF
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -4,6 +4,7 @@ require 'puppet/provider/mount'
 fstab = nil
 case Facter.value(:osfamily)
 when "Solaris"; fstab = "/etc/vfstab"
+when "AIX"; fstab = "/etc/filesystems"
 else
   fstab = "/etc/fstab"
 end
@@ -25,7 +26,12 @@ Puppet::Type.type(:mount).provide(
     @fields = [:device, :name, :fstype, :options, :dump, :pass]
   end
 
-  text_line :comment, :match => /^\s*#/
+  if Facter.value(:osfamily) == "AIX"
+    # * is the comment character on AIX /etc/filesystems
+    text_line :comment, :match => /^\s*\*/
+  else
+    text_line :comment, :match => /^\s*#/
+  end
   text_line :blank, :match => /^\s*$/
 
   optional_fields  = @fields - [:device, :name, :blockdevice]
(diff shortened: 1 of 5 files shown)


hunner:
Replaces #3740

geoffnichols:
Thanks for retargeting this. @melissa and I spent some time this afternoon trying to validate some of the concerns you mentioned with the remount behavior on AIX. A couple of findings of note:

- As best as we can tell, 'remount' for NFS filesystems is not supported in AIX 5.3 and supported only in later releases of AIX 6.1.**

- On AIX releases where remount is supported, it looks like there are some notable differences (vs Linux) in the effect of `mount -o remount`.

On AIX, for NFS filesystems, it looks like `mount -o remount /foo` removes existing mount options for the filesystem. Similarly, `mount -o remount,rw /foo` removes any existing mount options and remounts with 'rw' as the only mount option.

Here's a comparison of CentOS 6.5 vs AIX 7.1:

CentOS 6.5:

```
# mount | grep "nfs "
server:/srv/test on /mnt type nfs (rw,vers=3,wsize=65536)

# mount -o remount,ro /mnt
# mount | grep "nfs "
server:/srv/test on /mnt type nfs (ro,vers=3,wsize=65536)

# mount -o remount /mnt
# mount | grep "nfs "
server:/srv/test on /mnt type nfs (rw,vers=3,wsize=65536)
```

AIX 7.1:
```
# mount | grep nfs
server /srv/test         /mnt         nfs3   Mar 30 16:13 rw,wsize=65536

# mount -o remount,ro /srv/aix
# mount | grep nfs
server /srv/test        /mnt         nfs3   Mar 30 16:13 ro

# mount -o remount /srv/aix
# mount | grep nfs
server /srv/test         /mnt         nfs3   Mar 30 16:13
```
**Example output and a couple of references indicating remount is not supported for NFS filesystems on all releases of AIX:

On AIX 5.3:
```
# mount -o remount,rw /srv/test
nfsmnthelp: invalid option remount ignored
```
http://www-01.ibm.com/support/docview.wss?uid=isg1fixinfo108448
http://www-01.ibm.com/support/docview.wss?uid=isg1IZ32019

melissa:
@hunner I think in terms of next steps to address the issue @geoffnichols described above would be to have the `mount -o remount` command pass in all options in a comma separated list (i.e, `mount -o remount,ro,vers=3`), except on AIX 5.3, where unmount/mount should be used.  Maybe defaulting to `mount -o remount,rw` might not be a bad idea unless the user specifies `ro`. It seemed like that's what was happening anyway on AIX. @geoffnichols was that what you were thinking in terms of what should be done?

geoffnichols:
@melissa Yes, I think that seems sane.

hunner:
Updated: Does unmount/mount for both local and nfs mounts on AIX 5.3. On other AIX platforms, it does unmount/mount for nfs mounts and remounts for normal mounts.

geoffnichols:
This looks good to me in manual validation after the update. I'm :+1: on this.

melissa:
@helperton I know you helped to review this pull request already, but I was wondering if you might be willing to take another pass at it? It's been updated to address some additional concerns, and I'd like to get your confirmation that it still works for you with these changes. Thanks for your help!

helperton:
@melissa Hello, I've just downloaded the latest and things still seem to work as before during my testing.

melissa:
Awesome, @helperton, thanks for your help! I'm hoping to merge this as soon as the 3.x pipeline goes green

melissa:
This has been merged into puppet/3.x, and will be merged up into puppet/master. That means that things are going to be a little weird in that this will be available in Puppet 3.8 and Puppet 4.1, but not Puppet 4.0. If that's cause for concern, this work should have an additional pull request opened against the puppet/stable branch. Otherwise, we should be fine.

PR #3770, opened 2015-03-30
(PUP-1279) ADSI - Explicitly set user/group deleted

ferventcoder:
Previously, flush would only check against whether the
user/group variable was set before deciding to commit. This is
erroneous as the variable could be set (and sometimes is) by other
means. Explicitly set a  flag when a user/group is deleted if the
user/group variable is set.

Without this change there are cases when the user/group variable will
be set and flush will fail, throwing an error even though the deletion
is successful.

### DIFF
--- a/lib/puppet/provider/group/windows_adsi.rb
+++ b/lib/puppet/provider/group/windows_adsi.rb
@@ -9,6 +9,13 @@ Puppet::Type.type(:group).provide :windows_adsi do
 
   has_features :manages_members
 
+  attr_accessor :deleted
+
+  def initialize(value={})
+    super(value)
+    @deleted = false
+  end
+
   def members_insync?(current, should)
     return false unless current
 
@@ -72,11 +79,13 @@ Puppet::Type.type(:group).provide :windows_adsi do
 
   def delete
     Puppet::Util::Windows::ADSI::Group.delete(@resource[:name])
+
+    @deleted = true
   end
 
   # Only flush if we created or modified a group, not deleted
   def flush
-    @group.commit if @group
+    @group.commit if @group && !@deleted
   end
 
   def gid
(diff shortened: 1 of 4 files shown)


Iristyle:
@ferventcoder can we drop some tests in for this?

https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/user/windows_adsi_spec.rb
https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/group/windows_adsi_spec.rb

ferventcoder:
It's a refactor - the current tests should cover it. The only thing I could test that we may not have been testing previously is that flush doesn't error.

ferventcoder:
will add one for running flush as well

lib/puppet/provider/group/windows_adsi.rb
```diff
@@ -72,11 +72,12 @@ def exists?
 
   def delete
     Puppet::Util::Windows::ADSI::Group.delete(@resource[:name])
+    @group.deleted=true if @group
```
joshcooper: nit, can you add spaces, e.g. `deleted = true`

joshcooper: I think it might be easier to keep the deleted flag in the provider and not on the adsi object, e.g. `@deleted = true`, and then down below:

```
@group.commit if @group && !@deleted
```

ferventcoder: That could work as well.

ferventcoder:
@Iristyle @joshcooper updated.

lib/puppet/provider/group/windows_adsi.rb
```diff
@@ -9,6 +9,13 @@
 
   has_features :manages_members
 
+  attr_accessor :deleted
```
joshcooper: I don't think we need an accessor as this is private state to the provider. Same with user provider below.

Iristyle: :+1:

Iristyle:
I've updated the PR locally and will be closing shortly via merge commit.

Iristyle:
Merged to 3.x in https://github.com/puppetlabs/puppet/commit/d0288c80dfe95d46016b8d663308654956c74447 - GH commit message closing failing me again :(

PR #3771, opened 2015-03-30
(PUP-4340) Fix intermittent failure in test case

MikaelSmith:
The source_attribute test was modified to wait for time to pass
sufficient that touching source files on the master would cause puppet
agent runs to resync the files. However, the test only checked
mtime/ctime on the master - which is unrelated to ctime/mtime of the
target files on the agents - and the agent target files are what will be
compared to the ctime/mtime of the source file to see if they need to be
resynced.

We could fix this by iterating over the agents - or assuming they'll
always be done in order and only querying the last agent - to check the
ctime/mtime stats of the target files using `puppet resource file
<filename>`, but that's more complicated and unlikely to be
significantly less than the 1 second sleep we used to have.

Revert to the previous behavior of waiting 1 second to touch the files
and try resyncing.

### DIFF
--- a/acceptance/tests/resource/file/source_attribute.rb
+++ b/acceptance/tests/resource/file/source_attribute.rb
@@ -169,14 +169,11 @@ with_puppet_running_on(master, master_opts, testdir) do
   end
 
   step "touch files and verify they're updated with ctime/mtime"
-  # wait until we're not at the file's mtime
-  #   mod_dir_source_file is the last file modified
-  #   This stat will not work on most BSDs
-  dir_file_mtime = on(master, "stat --format '%Y' #{mod_source_dir_file}").stdout.chomp
-  master_time    = on(master, 'date +%s').stdout.chomp
-  until master_time > dir_file_mtime do
-    master_time  = on(master, 'date +%s').stdout.chomp
-  end
+  # wait until we're not at the mtime of files on the agents
+  # this could be done cross-platform using Puppet, but a single puppet query is unlikely to be less than a second,
+  # and iterating over all agents would be much slower
+  sleep(1)
+
   on master, "touch #{mod_source_file} #{mod_source_dir_file}"
   agents.each do |agent|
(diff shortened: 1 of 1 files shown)


acceptance/tests/resource/file/source_attribute.rb
```diff
@@ -169,14 +169,11 @@ class source_test_module {
...
-  #   This stat will not work on most BSDs
-  dir_file_mtime = on(master, "stat --format '%Y' #{mod_source_dir_file}").stdout.chomp
-  master_time    = on(master, 'date +%s').stdout.chomp
-  until master_time > dir_file_mtime do
-    master_time  = on(master, 'date +%s').stdout.chomp
-  end
+  # wait until we're not at the mtime of files on the agents
+  # this could be done cross-platform using Puppet, but a single puppet query is unlikely to be less than a second,
+  # and iterating over all agents would be much slower
+  sleep(1)
```
joshcooper: I'm assuming mtime/ctime resolution is always less than 1 second on all supported platforms (so the 1 second sleep is always long enough)?

MikaelSmith: Yes, resolution seems to be 1 second.

MikaelSmith:
I think the quickest alternative would be running a Ruby script to query mtime/ctime of the last file specified on the last agent.

joshcooper:
```
$ be rake ci:test:aio SHA=a8cc7cbd0d3581f19e40477a8cd1ba9452708a7d CONFIG=config/nodes/redhat-7-x86_64.yaml TESTS=tests/resource/file/source_attribute.rb
...
tests/resource/file/source_attribute.rb passed in 96.78 seconds
      Test Suite: tests @ 2015-03-31 08:47:52 -0700

       Total Suite Time: 96.78 seconds
      Average Test Time: 96.78 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```

PR #3772, opened 2015-03-30
(PUP-3804) User resource active directory handling

ferventcoder:
Error messages on user/group resources attempting to create
Active Directory users or groups are not very sufficient. Provide
an actionable error to the user so it is clear that Puppet cannot
handle creation and deletion of those types of users/groups.

### DIFF
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -161,7 +161,14 @@ module Puppet::Util::Windows::ADSI
     def commit
       begin
         native_user.SetInfo unless native_user.nil?
-      rescue Exception => e
+      rescue WIN32OLERuntimeError => e
+        if e.message =~ /8007089A/m
+          raise Puppet::Error.new(
+           "Puppet is not able to create/delete domain users with the user resource.\nPlease see http://links.puppetlabs.com/windows_usersgroups for details.",
+           e
+          )
+        end
+
         raise Puppet::Error.new( "User update failed: #{e}", e )
       end
       self
@@ -328,7 +335,14 @@ module Puppet::Util::Windows::ADSI
     def commit
       begin
         native_group.SetInfo unless native_group.nil?
-      rescue Exception => e
+      rescue WIN32OLERuntimeError => e
+        if e.message =~ /8007089A/m
+          raise Puppet::Error.new(
+            "Puppet is not able to create/delete domain groups with the group resource.\nPlease see http://links.puppetlabs.com/windows_usersgroups for details.",
+            e
+          )
+        end
+
         raise Puppet::Error.new( "Group update failed: #{e}", e )
       end
       self
--- a/spec/unit/provider/group/windows_adsi_spec.rb
+++ b/spec/unit/provider/group/windows_adsi_spec.rb
@@ -162,6 +162,18 @@ describe Puppet::Type.type(:group).provider(:windows_adsi), :if => Puppet.featur
         /Cannot create group if user 'testers' exists./ )
     end
 
+    it "should fail with an actionable message when trying to create an active directory group" do
+      resource[:name] = 'DOMAIN\testdomaingroup'
+      Puppet::Util::Windows::ADSI::User.expects(:exists?).with(resource[:name]).returns(false)
+      connection.expects(:Create)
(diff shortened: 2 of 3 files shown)


lib/puppet/util/windows/adsi.rb
```diff
@@ -162,6 +162,7 @@ def commit
       begin
         native_user.SetInfo unless native_user.nil?
       rescue Exception => e
+        raise Puppet::Error.new( "Puppet is not able to create or delete Active Directory users with the user resource", e ) if /8007089A in Active Directory/ =~ e.message
```
joshcooper: The error message is likely localized and may not match the regex. Can you GetLastError instead? Or just match the error code?

ferventcoder: I think I will just match the error code. :)

Iristyle: Do we want to catch the error, or catch the problem earlier (i.e. we know that `DOMAIN\User` is invalid input)

ferventcoder: Hard to know for sure. Catching it earlier could set us up for more validation of what the intent was. There are known scenarios you can work with a domain user in, just not ensure => present / absent.

joshcooper: The trouble is that `DOMAIN` could be the local computer, `.`, or `NT AUTHORITY` I think, and those we do support. You might be able to resolve the name to a SID, and ask the object if it's a built-in account or not. If not, check the domain and against the list of well known domains... but that gets a bit kludgy...

ferventcoder: What Josh just said. :)

Iristyle: Yeah, those are all fair points.

lib/puppet/util/windows/adsi.rb
```diff
@@ -162,6 +162,13 @@ def commit
       begin
         native_user.SetInfo unless native_user.nil?
       rescue Exception => e
+        if e.message =~ /8007089A/m
```
Iristyle: So in debugging, I see that this error is actually a `WIN32OLERuntimeError` type, so I would rather us be a bit more explicit about catching that type.

ferventcoder: Let's do that. DO we get any other type of error? Or should I say, should we capture any other type of error?

Iristyle: AFAIK, it's always a `WIN32OLERuntimeError`

ferventcoder: I'll just let the error bubble up if it is not a WIN32 OLE Runtime Error.

lib/puppet/util/windows/adsi.rb
```diff
@@ -328,7 +335,14 @@ def sid
...
         native_group.SetInfo unless native_group.nil?
-      rescue Exception => e
+      rescue WIN32OLERuntimeError => e
+        if e.message =~ /8007089A/m
+          raise Puppet::Error.new(
+            "Puppet is not able to create/delete domain groups with the group resource.\nPlease see http://links.puppetlabs.com/windows_usersgroups for details.",
+            e
+          )
+        end
+
```
ferventcoder: At some point we really should stop the duplication here. :|

Iristyle: This actually changes the original code so that now we only capture `WIN32OLERuntimeError` instead of all exceptions... is that what we want?

lib/puppet/util/windows/adsi.rb
Iristyle: Ah I see why you did this now.  I think we got our lines crossed... let me think about this.

Perhaps we do only want to catch `WIN32OLERuntimeError` and not *all* exceptions.

lib/puppet/util/windows/adsi.rb
ferventcoder: That is exactly what I asked before.

ferventcoder: https://github.com/puppetlabs/puppet/pull/3772#discussion_r27666189

ferventcoder: Nevermind, I see your comments at the end of that thread. Non-linear conversations FTL. haaha

PR #3774, opened 2015-03-31
(PUP-4090) Zypper provider improvements

melissa:
Prior to this commit, our zypper provider didn't work. When installing a
package with zypper, we pass in the `--name` flag, which is not
available in earlier versions of zypper that we account for. It also
breaks when you use `install_options`. It previously dropped in the
extra install options between `--name` and the package name to be
installed. This cuased it to fail.

This pull request reorders the options that we pass to the zypper
command. If the user has passed in the global option `--no-gpg-check`,
we pull that flag out, and pass it in where zypper expects it (after
`zypper` and before `install`).

### DIFF
--- a/lib/puppet/provider/package/zypper.rb
+++ b/lib/puppet/provider/package/zypper.rb
@@ -34,10 +34,9 @@ Puppet::Type.type(:package).provide :zypper, :parent => :rpm do
     end
 
     #This has been tested with following zypper versions
-    #SLE 10.2: 0.6.104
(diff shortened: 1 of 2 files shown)


melissa:
@kylog This pr was originally targeted at puppet/master, but it looks like we actually want to get into puppet 3.8.0. Because of this, I've retargeted this work at puppet/3.x. It would be awesome if you could take a look at it so we can get it in. @shrug gave his approval for the previous pr. Nothing has changed code-wise since then.

PR #3777, opened 2015-03-31
(PUP-4237) Add puppet-agent gem provider

branan:
This is a rebase of #3728 against the stable branch so it can be included in Puppet 4.0

It includes cherry-picks of PUP-4191 and PUP-4203 (from #3708), since this code was originally written assuming they would be merged into 4.0 but they ended up in 4.1

### DIFF
--- a/lib/puppet/provider/package/gem.rb
+++ b/lib/puppet/provider/package/gem.rb
@@ -122,7 +122,12 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
   end
 
   def uninstall
-    gemcmd "uninstall", "-x", "-a", resource[:name]
+    command = [command(:gemcmd), "uninstall"]
+    command << "--executables" << "--all" << resource[:name]
+    output = execute(command)
+
+    # Apparently some stupid gem versions don't exit non-0 on failure
+    self.fail "Could not uninstall: #{output.chomp}" if output.include?("ERROR")
   end
 
   def update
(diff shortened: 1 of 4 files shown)


branan:
To be more specific, I cherry-picked the fixed up versions of the #3708 commits that @melissa merged into master.

branan:
@joshcooper on the original PR you had asked about windows support. What is needed to make that happen?

branan:
Trivially it's "add another provider", but that really stinks. (having `puppet_gem_win` in addition to this one). I'm not aware of a way to declare a command is used/needed based on a fact. We could flag both the windows and linux commands as optional and implement the failure logic ourselves if neither is found, but that also feels kinda hacky to me :/

Is there an existing way of handling this sort of situation?

branan:
I still need to verify proper functioning on windows, just pushed up the code to my working branch

MikaelSmith:
:+1: This looks like it should work.

MikaelSmith:
@branan Let me know when you've verified on Windows?

branan:
@MikaelSmith looks like it works.

Windows firewall blocked Gem from the network, but that's a different problem that I'll talk to either RE or the windows team about.

jhoblitt:
Is there any chance of also getting this functionality in a 3.x release?

kylog:
@jhoblitt puppet 3.x doesn't have the same packages that bundle ruby, so the straight gem provider should work.

ferventcoder:
@kylog I wonder if @jhoblitt is specifically interested in the uninstall_options that was introduced to the gem provider at https://github.com/branan/puppet/commit/70a7e2870fa684a9642c6f110d407d20c47f1224

PR #3780, opened 2015-04-01
(PUP-4363) Support splay in apply as well

dalen:
For masterless deploys it would be good if puppet apply supported splay just like agent does to avoid thundering herd issues (without the need of wrapping it in a shell script with a sleep).

The `--test` option should imply `--no-splay` just like it does in the agent.

### DIFF
--- a/lib/puppet/agent.rb
+++ b/lib/puppet/agent.rb
@@ -9,11 +9,12 @@ class Puppet::Agent
   require 'puppet/agent/disabler'
   include Puppet::Agent::Disabler
 
-  attr_reader :client_class, :client, :splayed, :should_fork
+  require 'puppet/util/splayer'
+  include Puppet::Util::Splayer
+
+  attr_reader :client_class, :client, :should_fork
 
   def initialize(client_class, should_fork=true)
-    @splayed = false
-
     @should_fork = can_fork? && should_fork
     @client_class = client_class
   end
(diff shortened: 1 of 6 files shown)


branan:
The reason to have splay in normal daemonized agent mode is that in that case Puppet is responsible for scheduling its own runs. When running masterless, Puppet isn't doing that scheduling - some other process is telling puppet "run now." I feel like in that case it's the responsibility of the external scheduling process to handle splaying the agent.

Can you explain your use-case a bit better and why you feel Puppet needs to handle the splay here?

ahpook:
(from JIRA) This seems OK to me, I guess, as long as it defaults to false and needs to be enabled.

peterhuene:
Hi @dalen, any updates on this?  Branan posed some questions that we feel need to be answered before merging this PR.  Thanks!

dalen:
Well, even with the agent it seems like a fairly common pattern to run it from cron with a `--onetime` option. An issue with cron though is that it only supports minutes, so you ended up having to sleep fqdn_rand(60) before running puppet apply (and setting the minute to another fqdn_rand value). Which looks a bit ugly, so I thought it would make sense to just have the same feature as in the agent and use that to splay the runs.

dalen:
But yeah, this is a very minor feature. But on the other hand it reduces the diff between agent and apply slightly which is good IMHO.

Adding locking support in apply is likely next on my todo list in this direction :)

peterhuene:
@dalen I think we'd like to discuss this one tomorrow in the weekly PR triage if you're available (10 AM -07:00).  Thanks!

dalen:
sure, that should work for me this week

peterhuene:
@branan Ping for Erik's response above.  Does that make sense?  The code changes look good to me.

branan:
@peterhuene @dalen that sounds reasonable enough to me (too bad about Cron's lameness - I don't use it enough to know it's limits, I guess)

peterhuene:
@dalen Thanks for the contribution.  This has been merged into the next release of Puppet.

PR #3781, opened 2015-04-01
(PUP-4355) Update windows tests for ~/.puppetlabs

joshcooper:
Commit 4ae5aca changed the per-user directories from ~/.puppet to ~/.puppetlabs
but did not update the windows specs that only run on Windows.

This commit fixes the tests and fixes an old reference to ~/.puppet.


kylog:
:+1: 

Travis has kicked off; I'll merge once it goes green.

PR #3784, opened 2015-04-02
(PUP-4237) Add puppet_gem provider (take 2)

branan:
The previous version of this had an issue where the `provider` resource parameter would be set to `:gem` in the `instances` method of `puppet_gem`. This fixes the provider to be set properly.


kylog:
Does this need a rebase? I thought we'd already merged the PUP-4191 and PUP-4203 commits?

branan:
@kylog those ended up in master for 4.1

PR #3785, opened 2015-04-02
(maint) fix install repos on to use PC1 urls

er0ck:
this change helps the install utilities find repo_configs from our new
PC1 URLS on build.puppetlabs.lan.  this WILL break again (PC2?) and we need to
refactor and repackage all this code elsewhere, sometime, RSN.

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -162,6 +162,17 @@ module Puppet
             ]
 
             if not link_exists?(link)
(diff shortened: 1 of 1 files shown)


joshcooper:
This change was made in stable, but not master (because we're not merging stable to master any time soon). Does this really need to go master?

Also the version in stable checks PC1 -> products -> devel, but this one does products -> PC1 -> devel.

er0ck:
this or something similar should go in when master pipelines are setup again
i have no clue of what order of precedence the dirs should be in.

PR #3786, opened 2015-04-02
(PUP-3653) Create/set empty Windows groups

ferventcoder:
Allow setting a Windows group to no users. Previously it was not possible to set
Windows groups with no members due to the way the checking of should versus
current was done. Without this change Puppet will not be able to set Windows
groups to empty lists.

- `group/windows_adsi.members_insync?` - Remove the short circuit check on `current.empty?` versus
  `should_empty` - this results in incorrect behavior when non-authoritative
- `group/windows_adsi.members_insync?` - Return true when non-athoritative and should is empty.
  This prevents a check on keys of an empty hash, which will result in an error.
- `adsi.set_members` - return early only if `desired_members` is nil
- `adsi.set_members` - skip adding members when `desired_members` is empty
- `adsi.set_members` - if `desired_members` is empty, remove all current members

### DIFF
--- a/lib/puppet/provider/group/windows_adsi.rb
+++ b/lib/puppet/provider/group/windows_adsi.rb
@@ -22,9 +22,6 @@ Puppet::Type.type(:group).provide :windows_adsi do
 
     # Cannot use munge of the group property to canonicalize @should
     # since the default array_matching comparison is not commutative
-    should_empty = should.nil? or should.empty?
-
-    return false if current.empty? != should_empty
 
     # dupes automatically weeded out when hashes built
     current_users = Puppet::Util::Windows::ADSI::Group.name_sid_hash(current)
@@ -33,6 +30,7 @@ Puppet::Type.type(:group).provide :windows_adsi do
     if @resource[:auth_membership]
       current_users == specified_users
     else
+      return true if specified_users.empty?
       (specified_users.keys.to_a & current_users.keys.to_a) == specified_users.keys.to_a
     end
   end
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -388,19 +388,27 @@ module Puppet::Util::Windows::ADSI
     end
 
     def set_members(desired_members, inclusive = true)
-      return if desired_members.nil? or desired_members.empty?
+      return if desired_members.nil?
 
       current_hash = Hash[ self.member_sids.map { |sid| [sid.to_s, sid] } ]
       desired_hash = self.class.name_sid_hash(desired_members)
 
       # First we add all missing members
-      members_to_add = (desired_hash.keys - current_hash.keys).map { |sid| desired_hash[sid] }
-      add_member_sids(*members_to_add)
+      if !desired_members.empty?
+        members_to_add = (desired_hash.keys - current_hash.keys).map { |sid| desired_hash[sid] }
+        add_member_sids(*members_to_add)
+      end
 
-      # Then we remove all extra members
-      members_to_remove = (current_hash.keys - desired_hash.keys).map { |sid| current_hash[sid] }
+      # Then we remove all extra members if inclusive
+      if inclusive
+        if desired_members.empty?
+          members_to_remove = (current_hash.keys).map { |sid| current_hash[sid] }
+        else
+          members_to_remove = (current_hash.keys - desired_hash.keys).map { |sid| current_hash[sid] }
+        end
 
-      remove_member_sids(*members_to_remove) if inclusive
+        remove_member_sids(*members_to_remove)
+      end
     end
 
     def self.create(name)
--- a/spec/unit/provider/group/windows_adsi_spec.rb
+++ b/spec/unit/provider/group/windows_adsi_spec.rb
@@ -69,6 +69,10 @@ describe Puppet::Type.type(:group).provider(:windows_adsi), :if => Puppet.featur
         provider.members_insync?(['user1', 'user2', 'user2'], ['user2', 'user1', 'user1']).should be_true
       end
 
+      it "should return true when current user(s) and should user(s) are empty lists" do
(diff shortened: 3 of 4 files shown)


lib/puppet/util/windows/adsi.rb
```diff
@@ -388,19 +388,27 @@ def member_sids
...
-      add_member_sids(*members_to_add)
+      if !desired_members.empty?
+        members_to_add = (desired_hash.keys - current_hash.keys).map { |sid| desired_hash[sid] }
+        add_member_sids(*members_to_add)
+      end
 
-      # Then we remove all extra members
-      members_to_remove = (current_hash.keys - desired_hash.keys).map { |sid| current_hash[sid] }
+      # Then we remove all extra members if inclusive
+      if inclusive
```
ferventcoder: I bumped this up to keep this entire section from running unless inclusive.

lib/puppet/provider/group/windows_adsi.rb
```diff
@@ -33,6 +33,7 @@ def members_insync?(current, should)
     if @resource[:auth_membership]
       current_users == specified_users
     else
+      return true if specified_users.empty?
```
Iristyle:
```ruby
elsif specified_users.empty?
  true
else
```

?

ferventcoder: could do that as well, just not a fan of branching in that way and if it ever gets reordered, don't want the meaning to be lost.

ferventcoder: It would need specified like this: `elsif specified_users.empty? && !@resource[:auth_membership]`

lib/puppet/provider/group/windows_adsi.rb
```diff
@@ -22,9 +22,6 @@ def members_insync?(current, should)
 
     # Cannot use munge of the group property to canonicalize @should
     # since the default array_matching comparison is not commutative
-    should_empty = should.nil? or should.empty?
-
-    return false if current.empty? != should_empty
```
ferventcoder: Yanked. If we put this back for any reason, it also has a bug in should_empty set because assignment to logical boolean.

So it would need to come back as 
`should_empty = should.nil? || should.empty?`

ferventcoder: This changes what was brought in with https://github.com/puppetlabs/puppet/commit/455834e7f21f213e9d0085510d5b47c0c3e2a1ab

lib/puppet/provider/group/windows_adsi.rb
Iristyle: Yup, you're right....

lib/puppet/provider/group/windows_adsi.rb
Iristyle: The guy who worked on https://github.com/puppetlabs/puppet/commit/455834e7f21f213e9d0085510d5b47c0c3e2a1ab is a total :monkey:

ferventcoder: ha

lib/puppet/util/windows/adsi.rb
```diff
@@ -388,19 +388,27 @@ def member_sids
...
+      return if desired_members.nil?
 
       current_hash = Hash[ self.member_sids.map { |sid| [sid.to_s, sid] } ]
       desired_hash = self.class.name_sid_hash(desired_members)
 
       # First we add all missing members
-      members_to_add = (desired_hash.keys - current_hash.keys).map { |sid| desired_hash[sid] }
-      add_member_sids(*members_to_add)
+      if !desired_members.empty?
+        members_to_add = (desired_hash.keys - current_hash.keys).map { |sid| desired_hash[sid] }
```
ferventcoder: @Iristyle I'm curious what happens when the current hash is empty. I believe I tested this but still curious.

ferventcoder: Actually I guess the specs cover it.

lib/puppet/util/windows/adsi.rb
```diff
@@ -388,19 +388,27 @@ def member_sids
...
+        add_member_sids(*members_to_add)
+      end
 
-      # Then we remove all extra members
-      members_to_remove = (current_hash.keys - desired_hash.keys).map { |sid| current_hash[sid] }
+      # Then we remove all extra members if inclusive
+      if inclusive
+        if desired_members.empty?
+          members_to_remove = (current_hash.keys).map { |sid| current_hash[sid] }
```
Iristyle: Changing to 

```ruby
members_to_add = current_hash.values
```

lib/puppet/util/windows/adsi.rb
```diff
@@ -388,19 +388,27 @@ def member_sids
...
-      return if desired_members.nil? or desired_members.empty?
+      return if desired_members.nil?
 
       current_hash = Hash[ self.member_sids.map { |sid| [sid.to_s, sid] } ]
       desired_hash = self.class.name_sid_hash(desired_members)
 
       # First we add all missing members
-      members_to_add = (desired_hash.keys - current_hash.keys).map { |sid| desired_hash[sid] }
-      add_member_sids(*members_to_add)
+      if !desired_members.empty?
```
Iristyle: Going to perform the empty checks on `desired_hash` since that's what we're using after 394.

lib/puppet/util/windows/adsi.rb
ferventcoder: :+1:

lib/puppet/util/windows/adsi.rb
ferventcoder: Works for me - long as it is is empty and not nil

ferventcoder: Actually it doesn't matter...

Iristyle:
Merged to 3.x in 6b8d187

PR #3752, opened 2015-03-25
(maint) Cover logs generated by ensure file type

MikaelSmith:
Without this commit, the File type has no tests that verify the logs
generated when the state is changed.

Add tests that verify the logs generated when
- creating a new file
- overwriting an existing file
- removing a file

PR #3787, opened 2015-04-06
(PUP-4231) Cover logs generated by ensure => present in file type

whopper:
This commit adds spec tests to cover 'ensure => present' for
the file type.

Note: This PR is rebased on top of https://github.com/puppetlabs/puppet/pull/3752, and should be merged afterwards.

### DIFF
--- a/spec/integration/type/file_spec.rb
+++ b/spec/integration/type/file_spec.rb
@@ -93,6 +93,43 @@ describe Puppet::Type.type(:file), :uses_checksums => true do
     expect(Puppet::FileSystem.exist?(source)).to be_falsey
   end
 
+  describe "when ensure is present using an empty file" do
+    before(:each) do
+      catalog.add_resource(described_class.new(:path => path, :ensure => :present, :backup => :false))
+    end
+
+    context "file is present" do
+      before(:each) do
+        FileUtils.touch(path)
+      end
+
+      it "should do nothing" do
+        report = catalog.apply.report
+        expect(report.resource_statuses["File[#{path}]"]).not_to be_failed
+        expect(Puppet::FileSystem.exist?(path)).to be_truthy
+      end
+
+      it "should log nothing" do
+        logs = catalog.apply.report.logs
+        expect(logs).to be_empty
+      end
+    end
+
+    context "file is not present" do
+      it "should create the file" do
+        report = catalog.apply.report
+        expect(report.resource_statuses["File[#{path}]"]).not_to be_failed
+        expect(Puppet::FileSystem.exist?(path)).to be_truthy
+      end
+
+      it "should log that the file was created" do
+        logs = catalog.apply.report.logs
+        expect(logs.first.source).to eq("/File[#{path}]/ensure")
+        expect(logs.first.message).to eq("created")
+      end
+    end
+  end
+
   describe "when ensure is absent" do
     before(:each) do
       catalog.add_resource(described_class.new(:path => path, :ensure => :absent, :backup => :false))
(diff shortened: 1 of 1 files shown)


MikaelSmith:
You can rebase now.

whopper:
@MikaelSmith rebased!

MikaelSmith:
PUP-4231 added writing the content hash when `ensure => present`. Please add a test that covers when the file has content defined, and verifies that the log includes `defined content as '...'`.

spec/integration/type/file_spec.rb
```diff
@@ -93,6 +93,43 @@ def get_aces_for_path_by_sid(path, sid)
...
+      it "should create the file" do
+        report = catalog.apply.report
+        expect(report.resource_statuses["File[#{path}]"]).not_to be_failed
+        expect(Puppet::FileSystem.exist?(path)).to be_truthy
+      end
+
+      it "should log that the file was created" do
+        logs = catalog.apply.report.logs
+        expect(logs.first.source).to eq("/File[#{path}]/ensure")
+        expect(logs.first.message).to eq("created")
```
elyscape: This should be:
```ruby
expect(logs.first.message).to match(/defined content/)
```

spec/integration/type/file_spec.rb
```diff
@@ -93,6 +93,43 @@ def get_aces_for_path_by_sid(path, sid)
...
+        expect(Puppet::FileSystem.exist?(path)).to be_truthy
+      end
+
+      it "should log nothing" do
+        logs = catalog.apply.report.logs
+        expect(logs).to be_empty
+      end
+    end
+
+    context "file is not present" do
```
elyscape: Might be a good idea to ensure the file is missing first with:
```ruby
before(:each) do
  FileUtils.rm(path)
end
```

MikaelSmith: `path` will be guaranteed to be recreated, because it uses `let(:path)` to create a `tmpfile`.

elyscape: Oh.  In that case, never mind!

whopper:
@MikaelSmith updated to add the actual content hash log test (which I somehow missed was the entire original point of the ticket :)) Are those tests what you had in mind?

MikaelSmith:
:+1:

spec/integration/type/file_spec.rb
whopper: @elyscape heh, I had that same thought. Thanks for the help reviewing!

spec/integration/type/file_spec.rb
```diff
@@ -93,6 +93,43 @@ def get_aces_for_path_by_sid(path, sid)
...
+      it "should create the file" do
+        report = catalog.apply.report
+        expect(report.resource_statuses["File[#{path}]"]).not_to be_failed
+        expect(Puppet::FileSystem.exist?(path)).to be_truthy
+      end
+
+      it "should log that the file was created" do
+        logs = catalog.apply.report.logs
+        expect(logs.first.source).to eq("/File[#{path}]/ensure")
+        expect(logs.first.message).to eq("created")
```
whopper: @elyscape your comment here might have gotten blown away by my rebase (or you just deleted it:)). If no content is specified, it looks like we still just log "created" with your change:

    Notice: /Stage[main]/Main/File[/tmp/test1]/ensure: created

And then of course, `/defined content.../` when we provide `content => "foo"`:

    Notice: /Stage[main]/Main/File[/tmp/test2]/ensure: defined content as '{md5}877639120b17f7a4dc07783b667f13cf'

Does that all sound correct to you and for this test?

elyscape: Yeah, if all we have is an `ensure` parameter but nothing to specify what should be in the file, it'll say `created`, but if you actually specify what the file should be then it'll say `defined content as 'whatever'`.

PR #3788, opened 2015-04-06
(PUP-4373) Windows ADSI User groups property should behave similarly to Groups members property

ferventcoder:
Previously, when specifying a user's groups, the names were compared
as strings and needed to be an exact match. The group resource uses
SIDs when comparing the users, so the user resource should mimic that
behavior.

Without this change casing and names must match exactly and can 
produce erroneous results.

With this change, the comparison will be by SIDs so it will allow for
the following when a user is a member of Administrators and Users
groups:

```puppet
user { 'tim':
  ensure => present,
  groups => ['BUILTIN\Administrators','users'],
}
```

Groups members can also be set to an empty list and the user's groups will
report that it does but will not make the changes. Allow that the user groups to
be set to an empty list so that group's members and user's groups behavior
is the same. Without this change an existing user with groups modeled in the
following resource will continue to give the message that it is removing the 
groups without actually making any changes:

```puppet
user { 'tim':
  ensure => present,
  groups => [],
}
```

There is also shared behavior between `Puppet::Util::Windows::ADSI::User`
and `Puppet::Util::Windows::ADSI::Group` so it makes sense to start
combining shared functionality to a common class named
`Puppet::Util::Windows::ADSI::Shared`.

### DIFF
--- a/lib/puppet/provider/user/windows_adsi.rb
+++ b/lib/puppet/provider/user/windows_adsi.rb
@@ -41,7 +41,6 @@ Puppet::Type.type(:user).provide :windows_adsi do
     if @resource[:membership] == :inclusive
       current_users == specified_users
     else
-      return true if specified_users.empty?
       (specified_users.keys.to_a & current_users.keys.to_a) == specified_users.keys.to_a
     end
   end
--- a/lib/puppet/provider/user/windows_adsi.rb
+++ b/lib/puppet/provider/user/windows_adsi.rb
@@ -25,6 +25,27 @@ Puppet::Type.type(:user).provide :windows_adsi do
     user.set_groups(groups, @resource[:membership] == :minimum)
   end
 
+  def groups_insync?(current, should)
+    return false unless current
+
+    # By comparing account SIDs we don't have to worry about case
+    # sensitivity, or canonicalization of account names.
+
+    # Cannot use munge of the group property to canonicalize @should
+    # since the default array_matching comparison is not commutative
+
+    # dupes automatically weeded out when hashes built
+    current_users = Puppet::Util::Windows::ADSI::Group.name_sid_hash(current)
+    specified_users = Puppet::Util::Windows::ADSI::Group.name_sid_hash(should)
+
+    if @resource[:membership] == :inclusive
+      current_users == specified_users
+    else
+      return true if specified_users.empty?
+      (specified_users.keys.to_a & current_users.keys.to_a) == specified_users.keys.to_a
+    end
+  end
+
   def create
     @user = Puppet::Util::Windows::ADSI::User.create(@resource[:name])
     @user.password = @resource[:password]
--- a/lib/puppet/type/user.rb
+++ b/lib/puppet/type/user.rb
@@ -274,6 +274,14 @@ module Puppet
         raise ArgumentError, "Group names must be provided as an array, not a comma-separated list." if value.include?(",")
         raise ArgumentError, "Group names must not be empty. If you want to specify \"no groups\" pass an empty array" if value.empty?
       end
+
+      def insync?(current)
+        if provider.respond_to?(:groups_insync?)
+          return provider.groups_insync?(current, @should)
+        end
+
+        super(current)
+      end
     end
 
     newparam(:name) do
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -265,8 +265,7 @@ module Puppet::Util::Windows::ADSI
     end
 
     def set_groups(desired_groups, minimum = true)
-      return if desired_groups.nil? or desired_groups.empty?
-      # this needs fixed up once PUP-3653 is merged up
+      return if desired_groups.nil?
 
       desired_groups = desired_groups.split(',').map(&:strip)
 
@@ -282,7 +281,12 @@ module Puppet::Util::Windows::ADSI
       # Then we remove the user from all groups it is in but shouldn't be, if
       # that's been requested
       if !minimum
-        groups_to_remove = (current_hash.keys - desired_hash.keys).map { |sid| current_hash[sid] }
+        if desired_hash.empty?
+          groups_to_remove = current_hash.values
+        else
+          groups_to_remove = (current_hash.keys - desired_hash.keys).map { |sid| current_hash[sid] }
+        end
+
         remove_group_sids(*groups_to_remove)
       end
     end
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -97,18 +97,18 @@ module Puppet::Util::Windows::ADSI
       [:lpwstr, :lpdword], :win32_bool
   end
 
-  class User
-    extend Enumerable
-    extend FFI::Library
+  module Shared
+    def uri(name, host = '.')
+      if sid_uri = Puppet::Util::Windows::ADSI.sid_uri_safe(name) then return sid_uri end
 
-    attr_accessor :native_user
-    attr_reader :name, :sid
-    def initialize(name, native_user = nil)
-      @name = name
-      @native_user = native_user
+      host = '.' if ['NT AUTHORITY', 'BUILTIN', Socket.gethostname].include?(host)
+
+      account_type = self.name.split('::').last.downcase
+
+      Puppet::Util::Windows::ADSI.uri(name, account_type, host)
     end
 
-    def self.parse_name(name)
+    def parse_name(name)
       if name =~ /\//
         raise Puppet::Error.new( "Value must be in DOMAIN\\user style syntax" )
       end
@@ -120,20 +120,51 @@ module Puppet::Util::Windows::ADSI
       return account, domain
     end
 
+    def get_sids(adsi_child_collection)
+      sids = []
+      adsi_child_collection.each do |m|
+        sids << Puppet::Util::Windows::SID.octet_string_to_sid_object(m.objectSID)
+      end
+
+      sids
+    end
+
+    def name_sid_hash(names)
+      return {} if names.nil? || names.empty?
+
+      sids = names.map do |name|
+        sid = Puppet::Util::Windows::SID.name_to_sid_object(name)
+        raise Puppet::Error.new( "Could not resolve name: #{name}" ) if !sid
+        [sid.to_s, sid]
+      end
+
+      Hash[ sids ]
+    end
+  end
+
+  class User
+    extend Enumerable
+    extend Puppet::Util::Windows::ADSI::Shared
+    extend FFI::Library
+
+    # https://msdn.microsoft.com/en-us/library/aa746340.aspx
+    # IADsUser interface
+
+    require 'puppet/util/windows/sid'
+
+    attr_accessor :native_user
+    attr_reader :name, :sid
+    def initialize(name, native_user = nil)
+      @name = name
+      @native_user = native_user
+    end
+
     def native_user
       @native_user ||= Puppet::Util::Windows::ADSI.connect(self.class.uri(*self.class.parse_name(@name)))
     end
 
     def sid
       @sid ||= Puppet::Util::Windows::SID.octet_string_to_sid_object(native_user.objectSID)
-    end
-
-    def self.uri(name, host = '.')
-      if sid_uri = Puppet::Util::Windows::ADSI.sid_uri_safe(name) then return sid_uri end
-
-      host = '.' if ['NT AUTHORITY', 'BUILTIN', Socket.gethostname].include?(host)
-
-      Puppet::Util::Windows::ADSI.uri(name, 'user', host)
     end
 
     def uri
@@ -189,6 +220,7 @@ module Puppet::Util::Windows::ADSI
     end
 
     def groups
+      # https://msdn.microsoft.com/en-us/library/aa746342.aspx
       # WIN32OLE objects aren't enumerable, so no map
       groups = []
       native_user.Groups.each {|g| groups << g.Name} rescue nil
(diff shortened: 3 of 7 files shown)


ferventcoder:
Not sure if this should have gone against Stable or not.

lib/puppet/util/windows/adsi.rb
```diff
@@ -201,21 +231,50 @@ def remove_from_groups(*group_names)
...
 
       # Then we remove the user from all groups it is in but shouldn't be, if
       # that's been requested
-      groups_to_remove = current_groups - desired_groups
-      remove_from_groups(*groups_to_remove) unless minimum
+      if !minimum
+        #groups_to_remove = (current_hash.keys - desired_hash.keys).map { |sid| current_hash[sid] }
+        groups_to_remove = current_groups - desired_groups
+        remove_from_groups(*groups_to_remove)
+        #remove_group_sids(*groups_to_remove)
```
ferventcoder: WIP, ignore this bit

lib/puppet/util/windows/adsi.rb
```diff
@@ -302,7 +365,7 @@ def initialize(name, native_group = nil)
     end
 
     def uri
-      self.class.uri(name)
+      self.class.uri(*self.class.parse_name(name))
```
ferventcoder: Brings the same behavior for user domain\account checking that was present in users.

ferventcoder:
Unfortunately with these changes right now it fails because it converts `Adminstrators` to `BUILTIN\Administrators`.

Create a user and then try to put them into groups:

```puppet
user { 'tim':
  ensure => present,
  password => '123Adc!dd',
  groups => ['Administrators','Users'],
}
```
and you run into this:
```
Error: /Stage[main]/Main/User[tim]/groups: change from Users to Administrators,Users failed: ADSI connection error: failed to parse display name of moniker `WinNT://BUILTIN/Administrators,group'
    HRESULT error code:0x80070035
      The network path was not found.
```

ferventcoder:
So I think we need some additional work on well-known accounts and how they work with ADSI, because the account name doesn't match the host machine.

lib/puppet/util/windows/adsi.rb
```diff
@@ -312,7 +376,7 @@ def self.uri(name, host = '.')
     end
 
     def native_group
-      @native_group ||= Puppet::Util::Windows::ADSI.connect(uri)
+      @native_group ||= Puppet::Util::Windows::ADSI.connect(self.class.uri(*self.class.parse_name(name)))
```
ferventcoder: Brings the same behavior for user domain\account checking that is present in users.

ferventcoder:
@Iristyle @joshcooper any ideas how we want to handle Well-known accounts and their domains? e.g. `NT AUTHORITY`, `BUILTIN`, etc

lib/puppet/util/windows/adsi.rb
ferventcoder: And also brings the problems related to well-known group accounts...

ferventcoder: Nevermind on this.

ferventcoder:
@Iristyle @joshcooper I missed something from the original. So nevermind,

lib/puppet/util/windows/adsi.rb
```diff
@@ -97,18 +97,18 @@ def execquery(query)
...
+      if sid_uri = Puppet::Util::Windows::ADSI.sid_uri_safe(name) then return sid_uri end
 
-    attr_accessor :native_user
-    attr_reader :name, :sid
-    def initialize(name, native_user = nil)
-      @name = name
-      @native_user = native_user
+      host = '.' if ['NT AUTHORITY', 'BUILTIN', Socket.gethostname].include?(host)
+
+      account_type = self.name.split('::').last.downcase
```
ferventcoder: I started to introduce a new variable but moved to this instead due to some of the specs bringing the appearance that something like `Puppet::Util::Windows::ADSI::Group.uri('somegroup')` - and if they don't specify class type it was either null (or in my case I set up a default), so it could come out with an error or be wrong.

lib/puppet/util/windows/adsi.rb
```diff
@@ -201,21 +233,50 @@ def remove_from_groups(*group_names)
...
 
+
+    def add_group_sids(*sids)
+      group_names = []
+      sids.each do |sid|
+        group_names << Puppet::Util::Windows::SID.sid_to_name(sid)
+      end
+
+      add_to_groups(*group_names)
+    end
```
ferventcoder: Introduced instead of changing up `add_to_groups`. With the aliases I went the safe route but don't necessarily need to.

lib/puppet/util/windows/adsi.rb
```diff
@@ -201,21 +233,50 @@ def remove_from_groups(*group_names)
...
+
+    def remove_group_sids(*sids)
+      group_names = []
+      sids.each do |sid|
+        group_names << Puppet::Util::Windows::SID.sid_to_name(sid)
+      end
+
+      remove_from_groups(*group_names)
+    end
```
ferventcoder: Introduced instead of changing up the signature of `remove_from_groups`. With the aliases I went the safe route but we don't necessarily need to.

lib/puppet/util/windows/adsi.rb
```diff
@@ -201,21 +233,50 @@ def remove_from_groups(*group_names)
...
+
+    def group_sids
+      self.class.get_sids(native_user.Groups)
+    end
+
     def set_groups(desired_groups, minimum = true)
       return if desired_groups.nil? or desired_groups.empty?
+      # this needs fixed up once PUP-3653 is merged up
```
ferventcoder: Once PUP-3653 is merged up to master, we should fix this up. Doesn't need to be part of this PR though.

ferventcoder: That is this commit https://github.com/puppetlabs/puppet/commit/6b8d187cd151f01ddcaf43f26354ee6755f6c4c6 (with at the current time is only in 3.x)

lib/puppet/util/windows/adsi.rb
```diff
@@ -201,21 +233,50 @@ def remove_from_groups(*group_names)
...
+      end
 
       # Then we remove the user from all groups it is in but shouldn't be, if
       # that's been requested
-      groups_to_remove = current_groups - desired_groups
-      remove_from_groups(*groups_to_remove) unless minimum
+      if !minimum
+        groups_to_remove = (current_hash.keys - desired_hash.keys).map { |sid| current_hash[sid] }
+        remove_group_sids(*groups_to_remove)
+      end
```
ferventcoder: Introduced the same guard to not do any work if not inclusive.

spec/unit/util/windows/adsi_spec.rb
```diff
@@ -189,24 +189,21 @@
...
-            group1.expects(:Add).with("WinNT://testcomputername/#{username},user")
-
-            group3 = stub 'group1'
-            group3.expects(:Remove).with("WinNT://testcomputername/#{username},user")
-
-            Puppet::Util::Windows::ADSI.expects(:sid_uri).with(sid).returns("WinNT://testcomputername/#{username},user").twice
-            Puppet::Util::Windows::ADSI.expects(:connect).with('WinNT://./group1,group').returns group1
-            Puppet::Util::Windows::ADSI.expects(:connect).with('WinNT://./group3,group').returns group3
+            user.expects(:add_group_sids).with([-1])
+            user.expects(:remove_group_sids).with([1])
```
ferventcoder: I really simplified this. Instead of trying to work through each of the calls (I had about 20 expects going) and try to get the right combination, I switched over to this.

spec/unit/util/windows/adsi_spec.rb
```diff
@@ -301,11 +295,14 @@
         expect {
           adsi_group.expects(:Members).returns []
           group.set_members(['foobar'])
-        }.to raise_error(Puppet::Error, /Could not resolve username: foobar/)
+        }.to raise_error(Puppet::Error, /Could not resolve name: foobar/)
```
ferventcoder: No longer username since it works for both group and user

lib/puppet/type/user.rb
```diff
@@ -274,6 +274,14 @@ def should_to_s( newvalue )
...
         raise ArgumentError, "Group names must not be empty. If you want to specify \"no groups\" pass an empty array" if value.empty?
       end
+
+      def insync?(current)
+        if provider.respond_to?(:groups_insync?)
+          return provider.groups_insync?(current, @should)
+        end
+
+        super(current)
+      end
```
ferventcoder: Fun!

Iristyle: But make senses given we had to perform the same specialization to `group.rb`

lib/puppet/util/windows/adsi.rb
```diff
@@ -120,20 +120,51 @@ def self.parse_name(name)
...
+
+      Hash[ sids ]
     end
+  end
 
-    def self.uri(name, host = '.')
-      if sid_uri = Puppet::Util::Windows::ADSI.sid_uri_safe(name) then return sid_uri end
+  class User
+    extend Enumerable
+    extend Puppet::Util::Windows::ADSI::Shared
```
Iristyle: Maybe we can just use `module_function` instead of a mix-in here for `Shared`

lib/puppet/type/user.rb
ferventcoder: Yup.

lib/puppet/util/windows/adsi.rb
ferventcoder: Perhaps, but I didn't want to break any existing calls surrounding those things. That's why I went the mix-in route.

Iristyle: Ah right... hmm.

It's too bad we haven't been better at marking things `@private`.  Realistically, I think if we don't have callers internally to Puppet that we're OK.... but... you're technically correct wrt semver.

lib/puppet/provider/user/windows_adsi.rb
```diff
@@ -20,6 +20,27 @@ def groups=(groups)
...
+    # since the default array_matching comparison is not commutative
+
+    # dupes automatically weeded out when hashes built
+    current_users = Puppet::Util::Windows::ADSI::Group.name_sid_hash(current)
+    specified_users = Puppet::Util::Windows::ADSI::Group.name_sid_hash(should)
+
+    if @resource[:membership] == :inclusive
+      current_users == specified_users
+    else
+      return true if specified_users.empty?
```
joshcooper: This seems unnecessary, and I noticed there's similar logic in the group provider:

```
irb(main):001:0> a = {}
=> {}
irb(main):002:0> b = {:foo => 'bar'}
=> {:foo=>"bar"}
irb(main):003:0> (a.keys.to_a & b.keys.to_a) == a.keys.to_a
=> true
```

lib/puppet/util/windows/adsi.rb
```diff
@@ -120,20 +120,51 @@ def self.parse_name(name)
...
+        sids << Puppet::Util::Windows::SID.octet_string_to_sid_object(m.objectSID)
+      end
+
+      sids
     end
 
-    def sid
-      @sid ||= Puppet::Util::Windows::SID.octet_string_to_sid_object(native_user.objectSID)
+    def name_sid_hash(names)
+      return {} if names.nil? or names.empty?
```
joshcooper: Please let's not use `or` flow control here, instead `return {} if names.nil? || names.empty?`

spec/unit/provider/user/windows_adsi_spec.rb
```diff
@@ -60,6 +60,88 @@
...
+    let(:group3) { stub(:account => 'group3', :domain => '.', :to_s => 'group3sid') }
+
+    before :each do
+      Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with('group1').returns(group1)
+      Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with('group2').returns(group2)
+      Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with('group3').returns(group3)
+    end
+
+    it "should return false when current is nil" do
+      expect(provider.groups_insync?(nil, ['group2'])).to be_falsey
```
joshcooper: style comment, if this fails, rspec will just say true doesn't equal false, which isn't super helpful. You can get better error messages by flipping things around:

```
expect(provider).to be_groups_insync(nil, ['group2'])
```

See https://www.relishapp.com/rspec/rspec-expectations/v/3-2/docs/built-in-matchers/predicate-matchers#matcher-arguments-are-passed-on-to-the-predicate-method

lib/puppet/util/windows/adsi.rb
ferventcoder: Good catch. This was code moved up, I didn't refactor that change.

lib/puppet/provider/user/windows_adsi.rb
ferventcoder: It actually errors because specified_users doesn't have any keys. It could be that name_sid_hash(should) is returning an empty array though.

ferventcoder: It may be fine with the changes to the logical boolean below.

spec/unit/provider/user/windows_adsi_spec.rb
ferventcoder: The output is pretty cool.

```
 1) Puppet::Type::Group::ProviderWindows_adsi group type :members property helpers #members_insync? should return true when current user(s) and should user(s) are nil lists
     Failure/Error: expect(provider.members_insync?(nil, nil)).to be_truthy
       expected: truthy value
            got: false
     # ./spec/unit/provider/group/windows_adsi_spec.rb:63:in `block (4 levels) in <top (required)>'

  2) Puppet::Type::User::ProviderWindows_adsi groups_insync? when membership => inclusive should return true when should is nil
     Failure/Error: expect(provider).to be_groups_insync(['group1'], nil)
       expected `User[testuser](provider=windows_adsi).groups_insync?(["group1"], nil)` to return true, got false
     # ./spec/unit/provider/user/windows_adsi_spec.rb:97:in `block (4 levels) in <top (required)>'
```

However the code reader's experience is really really bad IMHO. I don't immediately know what it going on looking at that code and if I have similarly named methods e.g. `in_sync` vs `in_sync?`, which one does it match on? It's not immediately obvious unfortunately. My method named example isn't one we use but I have seen them from time to time. The actual test name is really helpful in determining the error, IMHO.

`Puppet::Type::Group::ProviderWindows_adsi group type :members property helpers #members_insync? should return true when current user(s) and should user(s) are nil lists` tells me immediately what is wrong regardless of the rest of the test output if the test fails.

ferventcoder:
I'm going to fix up the tests we talked about on Wednesday though.

ferventcoder:
This is ready @joshcooper  / @Iristyle

ferventcoder:
This adds additional behavior for setting user groups to empty now that https://github.com/puppetlabs/puppet/pull/3786 has made it's way to master.

joshcooper:
This is currently targeting master to be included in 4.2.0 (as part of SG). If you need it fixed sooner, please retarget to stable and update ticket fix version to be 4.1.0.

ferventcoder:
Retargeting

PR #3789, opened 2015-04-06
(PUP-1931) Fix mount provider failure w/o options

melissa:
Previously, when not specifying options to a mount resource, puppet
would fail with an opaque resource. It would create a bogus fstab entry
without an options field.

This commit adds in the `default` option to ensure that the options
field is created and valid for all mount resources.

Contribution originally from Mark Bainter <mbainter+github@gmail.com>

### DIFF
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -178,7 +178,21 @@ Puppet::Type.type(:mount).provide(
       end
     end
   else
-    record_line self.name, :fields => @fields, :separator => /\s+/, :joiner => "\t", :optional => optional_fields
+    record_line self.name, :fields => @fields, :separator => /\s+/, :joiner => "\t", :optional => optional_fields, :block_eval => :instance do
+      def pre_gen(record)
+        if !record[:options] || record[:options].empty?
+          if Facter.value(:kernel) == 'Linux'
+            record[:options] = 'defaults'
+          else
+            raise Puppet::Error, "Mount[#{record[:name]}]: Field 'options' is required"
+          end
+        end
+        if !record[:fstype] || record[:fstype].empty?
+          raise Puppet::Error, "Mount[#{record[:name]}]: Field 'fstype' is required"
+        end
+        record
+      end
+    end
   end
 
   # Every entry in fstab is :unmounted until we can prove different
--- a/spec/integration/provider/mount_spec.rb
+++ b/spec/integration/provider/mount_spec.rb
@@ -22,9 +22,9 @@ describe "mount provider (integration)", :unless => Puppet.features.microsoft_wi
     Puppet::Type.type(:mount).defaultprovider.stubs(:default_target).returns(@fake_fstab)
     Facter.stubs(:value).with(:hostname).returns('some_host')
     Facter.stubs(:value).with(:domain).returns('some_domain')
-    Facter.stubs(:value).with(:kernel).returns('Darwin')
-    Facter.stubs(:value).with(:operatingsystem).returns('Darwin')
-    Facter.stubs(:value).with(:osfamily).returns('Darwin')
+    Facter.stubs(:value).with(:kernel).returns('Linux')
+    Facter.stubs(:value).with(:operatingsystem).returns('RedHat')
+    Facter.stubs(:value).with(:osfamily).returns('RedHat')
     Puppet::Util::ExecutionStub.set do |command, options|
       case command[0]
       when %r{/s?bin/mount}
@@ -35,24 +35,7 @@ describe "mount provider (integration)", :unless => Puppet.features.microsoft_wi
             ''
           end
         else
-          expect(command.length).to eq(4)
-          expect(command[1]).to eq('-o')
-
-          # update is a special option, used on bsd's
-          # strip it out and track as a local bool here
-          update = false
-          tmp_options = command[2].split(",")
-
-          if tmp_options.include?("update")
-            update = true
-            tmp_options.delete("update")
-          end
-          @current_options = tmp_options.join(",")
-
-          if !update
-            expect(@mounted).to eq(false) # verify that we don't try to call "mount" redundantly
-          end
-          expect(command[3]).to eq('/Volumes/foo_disk')
+          expect(command.last).to eq('/Volumes/foo_disk')
           @current_device = check_fstab(true)
           @mounted = true
           ''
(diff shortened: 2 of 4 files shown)


branan:
I'd like to see tests that verify it's still possible to get an empty string (no `defaults`) if one so chooses/desires in their manifest.

melissa:
@branan I updated the spec tests to include the empty string situation. I'm not sure how effective they are though

branan:
Looks good to me.

melissa:
@branan who would be a good person from CS to get to look at this change?

branan:
@Sharpie maybe? He should at least know who to forward it to :)

The question: Does switching the default to "defaults" (while still allowing an empty string when it's specified explicitly) constitute a crazy enough change that it could piss off users/custoemrs?

I suspect "no, as long as we document it", since it's an obviously sane thing for us to be doing :)

Sharpie:
I actually haven't spent a lot of time mounting volumes with my bare hands. On the surface, this change seems reasonable --- but the best source of input would be the community and Professional Services.

melissa:
@Sharpie is there someone in Professional Services that would be a good person to ask about this update?

stschulte:
One thing you should keep in mind: A default does not only apply when creating the resource, but everytime `options` is left blank. So let's just say I want to mount my dvd with `puppet resource /mnt/dvd ensure=mounted` and I already have an `/etc/fstab` entry for `/mnt/dvd`. Since I have not specified `options` on the command line, the new default of `defaults` will apply and effectively change my `/etc/fstab`. Without the change, puppet will ignore the current options with the drawback that puppet also leaves options blank when making a new `fstab` entry.

Having said that I am in favour of sane defaults for the `options` property (not only for linux) but this should be implemented at a provider level when actually writing `/etc/fstab` (IIRC the parsed provider can implement a `pre_gen` hook that could check if the `options` field is empty) or treated as a breaking change. At least that would be my opinion on that :-)

Sharpie:
@melissa Unfortunately, I can't pick a `fstab` expert off the top of my head. The best source of input for this change may be the puppet-dev mailing list.

I agree with @stschulte: we should avoid overwriting any on-disk `options` if a resource declaration doesn't specify a value for this parameter.

melissa:
Unfortunately, I'm not an `fstab` expert either :(

@stschulte I'm trying to wrap my brain around what you said. Since I have such little experience with fstab, please correct me if I'm wrong:

### As it stands
  - Creation of a new entry
    * `:options => ""`; `fstab` options will be set to an empty string
    * options not set; `fstab` options will be set to an empty string
    * `:options => "something,valid"`; `fstab` options will be set to `something,valid`
  - Updating an existing entry
    * existing options blank
      - `:options => ""`; `fstab` options will remain an empty string
      - options not set; `fstab` options will remain an empty string
      - `:options => "something,valid"`; `fstab` options will be set to `something,valid`
    * existing options set to `existing,stuff`
      - `:options => ""`; `fstab` options will be reset to an empty string
      - options not set; `fstab` options will remain set to `existing,stuff`
      - `:options => "something,valid"`; `fstab` options will be reset to `something,valid`

### With my change
  - Creation of a new entry
    * `:options => ""`; `fstab` options will be set to an empty string
    * options not set; `fstab` options will be set to `defaults`
    * `:options => "something,valid"`; `fstab` options will be set to `something,valid`
  - Updating an existing entry
    * existing options blank
      - `:options => ""`; `fstab` options will remain an empty string
      - options not set; `fstab` options will be set to `defaults`
      - `:options => "something,valid"`; `fstab` options will be set to `something,valid`
    * existing options set to `existing,stuff`
      - `:options => ""`; `fstab` options will be reset to an empty string
      - options not set; `fstab` options will be reset to `defaults`
      - `:options => "something,valid"`; `fstab` options will be reset to `something,valid`

So, if someone has an existed `fstab` entry that has options set to an empty string, this entry is technically unmountable, from what I understand. I agree we shouldn't overwrite options if the fstab is already set to something valid, but what if it isn't? Like the case where it's an empty string? Since that's an existing entry, should we leave it as is, or set it to `defaults`?

stschulte:
@melissa you are absolutly correct about the current behaviour and the expected behaviour once your patch is applied.

If puppet currently sets options to an empty string (either because `options` is explicitly set to `""` or because puppet creates the resource while `options` was not set) we are left with an invalid `/etc/fstab` entry. Because the entry is invalid, Puppet will not be able to parse the entry correctly on any subsequent run. This happens because the field seperator in `/etc/fstab` is any number of whitespace, so puppet will interpret the `dump` (the one next to `options` field) as the options

Example:
Create an fstab entry with no option:

    # puppet resource mount /mnt/test device=/dev/test fstype=tmpfs ensure=present
    Notice: /Mount[/mnt/test]/ensure: defined 'ensure' as 'defined'

Reread the `fstab` entry. Notice that `options` is read as `0` which is actually the `dump` field:
````puppet
# puppet resource mount /mnt/test                                             
mount { '/mnt/test':
  ensure  => 'unmounted',
  device  => '/dev/test',
  dump    => '0',
  fstype  => 'tmpfs',
  options => '0',
  target  => '/etc/fstab',
}
````

The parsing itself is correct, because man fstab states: *Each filesystem is described on a separate line.  Fields on  each  line are separated by tabs or spaces.*

That beeing said:
*Since that's an existing entry, should we leave it as is or set it to defaults?* - I'd argue that we should not try to fix an incorrect entry because we wouldn't know how to fix it. If an fstab entry has to have **x** fields seperated by whitespace and we only find **x-1** fields, we cannot say which one is which.

So in short:
- setting `options` explicitly to an empty string should not be allowed (by the validate method of the property). But not setting `options` at all should be allowed.
- setting `fstype` explicitly to an empty string should not be allowed (by the validate method of the property). Not setting `fstype` at all should be allowed
- when actually writing an entry to `fstab` we should check that all mandatory fields are set (according to `man fstab` on Linux, only `dump` and `pass` (the last two fields) are allowed to be absent) and either set default values (like your proposal for the `options` property) or fail.

Does that make sense?

**Comment by melissa** [comment]
@stschulte Yes, it does. Thank for your the clarification and explanation here! I'll get started on updatting this pull request

**Comment by melissa** [comment]
Hmm, I'm running into some trouble figuring out the best way to do that. Do you have any advice for me @stschulte ?

**Comment by melissa** [comment]
Thanks to @haus, I think this is what we want to be doing. I still have to update the tests though.

File: lib/puppet/provider/mount/parsed.rb [hunk 1]
```diff
@@ -178,7 +178,21 @@ def to_line(result)
...
+      def pre_gen(record)
+        if !record[:options] || record[:options].empty?
+          if Facter.value(:kernel) == 'Linux'
+            record[:options] = 'defaults'
+          else
+            raise RuntimeError, "Mount[#{record[:name]}]: Field 'options' is required"
+          end
+        end
+        if !record[:fstype] || record[:fstype].empty?
+          raise RuntimeError, "Mount[#{record[:name]}]: Field 'fstype' is required"
```
stschulte: I am not sure about this one, but I guess `self.fail` is the best way to raise and error here because it will automatically add context like classname and failing resource. But haven't tried it

melissa: `self.fail` gave me `Error: /Stage[main]/Main/Mount[/vol0]: Could not evaluate: private method `fail' called for #<Puppet::Util::FileParsing::FileRecord:0x0000000413a638>`, `Puppet.err` did error out, but also updated the contents of `/etc/fstab`. `fail` does what I would expect, as does `raise RuntimeError`, in that it throws an error, and doesn't update `/etc/fstab`

lib/puppet/provider/mount/parsed.rb
```diff
@@ -178,7 +178,21 @@ def to_line(result)
       end
     end
   else
-    record_line self.name, :fields => @fields, :separator => /\s+/, :joiner => "\t", :optional => optional_fields
+    record_line self.name, :fields => @fields, :separator => /\s+/, :joiner => "\t", :optional => optional_fields, :block_eval => :instance do
```
stschulte: BTW: You can also pass `:pre_gen => proc {} ` here as in `provider/ssh_authorized_key/parsed.rb`. I guess just a matter of taste :)

melissa:
Alright, I'm running into some trouble with these tests. @joshcooper @Sharpie @stschulte any advice for me on the best way to test this change with rspec?

haus:
@melissa I'm :+1: here once the commit messages are in order

melissa:
@joshcooper and @kylog if either of you have a change, it would be very helpful to get your opinion on this. I believe it's just about wrapped up

peterhuene:
I'm :+1:

melissa:
@stschulte thank you for your guidance on this so far! I'd very much appreciate if you could look over the code, and let us know if you approve or not, and if not, where any changes/improvements should be made.

@mbainter I believe you were the one who originally started this work. Could you take a look at this fix and let me know what you think?

geoffnichols:
This looks good to me. I'm :+1: .

mbainter:
Looks good here as well.  :+1:

lib/puppet/provider/mount/parsed.rb
```diff
@@ -178,7 +178,21 @@ def to_line(result)
...
+      def pre_gen(record)
+        if !record[:options] || record[:options].empty?
+          if Facter.value(:kernel) == 'Linux'
+            record[:options] = 'defaults'
+          else
+            raise Puppet::Error, "Mount[#{record[:name]}]: Field 'options' is required"
```
MikaelSmith: We should have spec tests for error conditions as well.

melissa: @MikaelSmith check out line 116 in spec/integration/provider/mount_spec.rb

spec/integration/provider/mount_spec.rb
```diff
@@ -123,27 +106,31 @@ def run_in_catalog(settings)
...
                     if family == "Solaris"
                       skip("Solaris: The mock :operatingsystem value does not get changed in lib/puppet/provider/mount/parsed.rb")
                     else
-                      @desired_options = options_setting
-                      run_in_catalog(:ensure=>ensure_setting, :options => options_setting)
-                      expect(@mounted).to eq(expected_final_state)
-                      if expected_fstab_data
-                        expect(check_fstab(expected_fstab_data)).to eq("/dev/disk1s1")
+                      if options_setting && options_setting.empty?
+                        expect { run_in_catalog(:ensure=>ensure_setting, :options => options_setting) }.to raise_error Puppet::ResourceError
```
MikaelSmith: I don't see where a ResourceError is being raised. It also doesn't appear that we test the case where `:kernel != "Linux"`.

melissa: @haus @Sharpie and I decided it wasn't worth the additional nesting and magic to add that test case. Before these changes, this test really wasn't doing much at all that was helpful, since it was only testing darwin. The resource error here does catch the error raised on the line you commented above.

Sharpie: The essential issue is that this entire integration test is a quagmire of sadness and pain and it badly needs to be re-written so that _all of the tests_ are not living together inside of 12 nested loops and conditionals.

However, we decided that this task is outside the scope of "implementing a change to the mount provider".

MikaelSmith: Ok, I'll accept that.

PR #3790, opened 2015-04-07
(PUP-2455) Update service provider to use 'ctrun' on Solaris

geoffnichols:
Prior to this commit, non-SMF Puppet-managed services on Solaris
are not started in their own SMF contracts. If started
by Puppet running as a daemon, the services end up in Puppet's
SMF contract.

There are several negative implications from this scenario:

(1) If Puppet is stopped or restarted, the non-SMF Puppet-managed
services in Puppet's SMF contract are killed (by SMF).

(2) If Puppet-managed services are in Puppet's SMF contract, certain
events impacting any one of the managed services can lead to SMF killing
killing Puppet's contract, also killing all Puppet-managed services
that have ended up in Puppet's SMF contract.

In both scenarios, SMF also places Puppet in the SMF "maintenance"
state.

This commit updates the Puppet service provider to use the Solaris
'ctrun' facility to start Puppet-managed services in their own
SMF contracts.

We also add an acceptance test and update spec tests to accomodate
use of :osfamily in the texecute method override.

### DIFF
--- a/acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
+++ b/acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
@@ -1,0 +1,132 @@
+test_name "(PUP-2455) Service provider should start Solaris init service in its own SMF contract"
+
+sleepy_daemon_script = <<SCRIPT
+#!/usr/bin/env ruby
+while true
+  sleep (2)
+end
+SCRIPT
+
+sleepy_daemon_initscript = <<INITSCRIPT
+#!/bin/sh
+FIXTURESERVICE="/tmp/sleepy_daemon"
+start(){
+    $FIXTURESERVICE &
+}
+
+stop(){
+    FIXTUREPID=`ps -ef | grep "$FIXTURESERVICE" | grep -v grep | awk '{print $2}'`
+    kill -9 $FIXTUREPID
+}
+
+status(){
+    FIXTUREPID=`ps -ef | grep "$FIXTURESERVICE" | grep -v grep | awk '{print $2}'`
+    if [ "x$FIXTUREPID" = "x" ]; then
+      exit 1
+    else
+      exit 0
+    fi
+}
+case "$1" in
+    start)
+        start
+        ;;
+    stop)
+        stop
+        ;;
+    status)
+        status
+        ;;
+esac
+INITSCRIPT
+
+step "Setup fixture service manifest on master"
+
+  testdir = master.tmpdir('solaris_services_in_own_smf_contract')
+
+  test_manifest = <<MANIFEST
+File {
+  ensure => directory,
+  mode => "0750",
+  owner => #{master.puppet['user']},
+  group => #{master.puppet['group']},
+}
+file { '#{testdir}': }
+file { '#{testdir}/environments': }
+file { '#{testdir}/environments/production': }
+file { '#{testdir}/environments/production/manifests': }
+file { '#{testdir}/environments/production/manifests/site.pp':
+  ensure  => file,
+  content => '
+    node default {
+      service { "sleepy_daemon":
+      provider => "init",
+      enable   => "true",
+      ensure   => "running",
+      status   => "/etc/init.d/sleepy_daemon status",
+      }
+    }
+  ',
+}
+MANIFEST
+
+  apply_manifest_on master, test_manifest
+
+step "Start master"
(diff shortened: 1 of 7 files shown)


geoffnichols:
Replaces https://github.com/puppetlabs/puppet/pull/3660.

lib/puppet/provider/service/init.rb
```diff
@@ -152,6 +152,13 @@ def restartcmd
     (@resource[:hasrestart] == :true) && [initscript, :restart]
   end
 
+  def texecute(type, command, fof = true, squelch = false, combine = true)
+    if type =~ /^start/ && Facter.value(:osfamily) == "Solaris"
```
joshcooper: Ideally this logic would live in a Solaris specific provider that inherited from `init`. But it appears the `init` provider has lots of platform specific tweaks already so probably ok as is.

melissa:
:+1: @branan @kylog if you want to take a look at this, that would be awesome. I'd like to get it merged in sooner rather than later as it's a support escalation.

acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
```diff
@@ -0,0 +1,140 @@
...
+    }
+  ',
+}
+MANIFEST
+
+  apply_manifest_on master, test_manifest
+
+step "Start master"
+
+  if master.is_pe?
```
branan: You shouldn't have to set this stuff - it's part of beaker setup (typically)

geoffnichols: @branan Thanks, I updated to remove that.

haus:
Do we want new spec tests for this behavior in addition to the acceptance tests?

geoffnichols:
@haus Yes, thanks. 

 I'm struggling a bit with with the spec test.

I added this to ```spec/unit/provider/service/init_spec.rb```:

```
describe "when starting a service on Solaris" do
  it "should use ctrun" do
    Facter.stubs(:value).with(:osfamily).returns 'Solaris'
    provider.expects(:texecute).with(:start, ['/usr/bin/ctrun -l none', '/service/path/myservice', :start], true).returns("")
    provider.start
  end
end
```
When I execute, I'm seeing this error:

```
# be rspec spec/unit/provider/service/init_spec.rb
..................................F

Failures:

  1) Puppet::Type::Service::ProviderInit if the init script is present when starting a service on Solaris should use ctrun
     Failure/Error: provider.start
     Mocha::ExpectationError:
       unexpected invocation: Service[myservice](provider=init).texecute(:start, ['/service/path/myservice', :start], true)
       unsatisfied expectations:
       - expected exactly once, not yet invoked: Service[myservice](provider=init).texecute(:start, ['/usr/bin/ctrun -l none', '/service/path/myservice', :start], true)
       satisfied expectations:
       - allowed any number of times, invoked once: File.directory?('/alt/service/path')
       - allowed any number of times, invoked once: File.directory?('/service/path')
       - allowed any number of times, invoked once: Puppet::FileSystem.exist?('/service/path/myservice')
       - allowed any number of times, not yet invoked: Facter.value(:osfamily)
     # ./lib/puppet/provider/service/service.rb:39:in `ucommand'
     # ./lib/puppet/provider/service/base.rb:65:in `start'
     # ./spec/unit/provider/service/init_spec.rb:216:in `block (4 levels) in <top (required)>'

Finished in 0.1179 seconds
35 examples, 1 failure

```
Since we're overriding ```texecute``` based on the Facter returning "Solaris" for ```osfamily```,  it seems like ```allowed any number of times, not yet invoked: Facter.value(:osfamily)``` may be part of the issue.

geoffnichols:
@branan @melissa Any ideas regarding the spec test referenced in the previous comment?

geoffnichols:
Updated commit with a couple of simple spec tests.

geoffnichols:
Thanks @branan for your help with this! 

The just-added spec tests found a valid issue on Ruby 1.8.7, where we can't use regex match on a symbol. 

Updated the commit to change 
```
if type =~ /^start/ && Facter.value(:osfamily) == "Solaris"
```
to
```
 if type == :start && Facter.value(:osfamily) == "Solaris"
```

branan:
:+1: from me here

haus:
:+1:

melissa:
This requires a merge up into pe-puppet

PR #3791, opened 2015-04-07
(maint) Change old rspec 'should' to 'expect().to' syntax

thallgren:
One of the tests were found to be broken. It's now commented out and
the logged as issue PUP-4382.


joshcooper:
I thought we had checks in Travis to ensure old syntax didn't creep in? /cc @kylog

thallgren:
The code with the old syntax came in when I merged the 3.x branch.

kylog:
In addition, since this one is a deprecation warning, nothing would fail in travis. I'm not aware of a 'treat warnings as errors' flag for rspec, but I also haven't looked very hard :)

PR #3792, opened 2015-04-07
(PUP-4382) Fix test for retrieval of 'latest' property

branan:
There were two issues with this test:
* It was using a resource for which the fixture data didn't actually
exist for this query
* It failed to call `prefetch`, so the `property_hash` was never
  populated anyway

This fixes up the test to use a different resource and appropriately
prefetches the data.

### DIFF
--- a/lib/puppet/provider/package/pkgng.rb
+++ b/lib/puppet/provider/package/pkgng.rb
@@ -127,4 +127,8 @@ Puppet::Type.type(:package).provide :pkgng, :parent => Puppet::Provider::Package
     @property_hash[:latest]
   end
 
+  def origin
+    @property_hash[:origin]
+  end
+
 end
--- a/spec/unit/provider/package/pkgng_spec.rb
+++ b/spec/unit/provider/package/pkgng_spec.rb
@@ -6,6 +6,7 @@ provider_class = Puppet::Type.type(:package).provider(:pkgng)
 
 describe provider_class do
   let(:name) { 'bash' }
+  let(:installed_name) { 'zsh' }
   let(:pkgng) { 'pkgng' }
 
   let(:resource) do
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/package/pkgng.rb
```diff
@@ -127,4 +127,8 @@ def latest
     @property_hash[:latest]
   end
 
+  def origin
```
branan: I needed to add this to get `prefetch` to work correctly

spec/unit/provider/package/pkgng_spec.rb
```diff
@@ -1,11 +1,13 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
 require 'puppet/provider/package/pkgng'
+require 'pry'
```
kylog: You can drop pry :)

branan: Whoops, thought I skipped all of those with `add -p`

spec/unit/provider/package/pkgng_spec.rb
```diff
@@ -140,9 +142,10 @@ def run_in_catalog(*resources)
     # but neither of them works because provider.latest returns nil. This
     # is logged as issue PUP-4382
     #
-    # it("should retrieve the correct version of the latest package") {
-    #  provider.latest.should_not nil
-    #}
+    it("should retrieve the correct version of the latest package") {
+      provider_class.prefetch( { installed_name => installed_resource })
+      expect(installed_resource.provider.latest).not_to be_nil
+    }
```
kylog: I'd drop the entire comment block above.

Also, for consistency with the rest of the file, I'd drop the parens on the `it`, and use `do ... end` rather than `{ ... }`.  Not sure where it ended up inconsistent, but might as well clean it up while we're in the neighborhood.

branan: fixed

branan:
@kylog I just pushed up another commit to uncomment the remaining disabled test. LMK if you want me to ticket that or put it up into another PR, but it seemed reasonably related and was an easy fix.

melissa:
:+1:

kylog:
Boy, there's some refactoring that could go into these spec tests too. 

But meanwhile :+1: on these changes.

PR #3793, opened 2015-04-07
(PUP-4356) - DO NOT MERGE - Remove puppetversion from PMT metadata

andersonmills:
Before this commit, there was an undocumented key, puppetversion, which was tracked by the Puppet::Module class. This did nothing but throw an error if the metadata value was an exact match to Puppet.version. This commit removes the recognition and checking of that metadata key:value
pair.

DO NOT MERGE - As of 2015-04-07 we're still waiting on CI for 4.1.0, so acceptance tests and spec tests are not passing, yet. DO NOT MERGE this until all of those are set up or passing.

### DIFF
--- a/lib/puppet/module.rb
+++ b/lib/puppet/module.rb
@@ -56,7 +56,7 @@ class Puppet::Module
   attr_writer :environment
 
   attr_accessor :dependencies, :forge_name
-  attr_accessor :source, :author, :version, :license, :puppetversion, :summary, :description, :project_page
+  attr_accessor :source, :author, :version, :license, :summary, :description, :project_page
 
   def initialize(name, path, environment)
     @name = name
@@ -66,8 +66,6 @@ class Puppet::Module
     assert_validity
 
     load_metadata if has_metadata?
-
-    validate_puppet_version
 
     @absolute_path_to_manifests = Puppet::FileSystem::PathPattern.absolute(manifests)
   end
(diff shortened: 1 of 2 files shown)


lib/puppet/module.rb
```diff
@@ -56,7 +56,7 @@ def self.is_module_namespaced_name?(name)
   attr_writer :environment
 
   attr_accessor :dependencies, :forge_name
-  attr_accessor :source, :author, :version, :license, :puppetversion, :summary, :description, :project_page
+  attr_accessor :source, :author, :version, :license, :summary, :description, :project_page
```
joshcooper: Removing the accessor is a breaking change. I think you want to add explicit getter/setters that return nil. May want to issue a deprecation warning (though I could be convinced to make it a docs only deprecation).

lib/puppet/module.rb
```diff
@@ -307,11 +303,6 @@ def unmet_dependencies
     unmet_dependencies
   end
 
-  def validate_puppet_version
```
joshcooper: similar breaking change note here.

joshcooper:
Side note, we're about to reorg the stable & master branches. If this change needs to be in 4.1.0, it should target stable. If the change needs to be in 4.2.0, it should target master, with the caveat that 4.2.0 will be in SG. So if you just need it in SG, then can leave it targeted at master.

lib/puppet/module.rb
andersonmills: Will add the explicit getter/setter. 

From what I understand, this was a never-documented metadata key (aside from the source, of course), and affects 6 modules on the Forge with values ranging from 0.25 to 2.6 to 2.6.4. Since the validation is a string match, I think it would only allow one of those to actually work. So, I'd say just a release note about deprecation should be fine.

andersonmills:
In an ideal world, it would've gone in 4.0.0, but that didn't happen, so if I can get it in 4.1.0, that'd be great. I'll target stable, which means I'll close this one and reopen a new one.

PR #3795, opened 2015-04-07
(maint) Pin to a released beaker version

joshcooper:
Previously we were tracking on the master branch of beaker, which makes
it hard to know which version of beaker was compatible with puppet 4
when it was released. For example, as we continue to support puppet
4.0.z releases while beaker development moves forward.

This commit pins us to beaker ~> 2.8, which includes 2.8.1 and 2.9.z
releases, but not 3.0.

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -12,7 +12,7 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem "beaker", *location_for(ENV['BEAKER_VERSION'] || 'git://github.com/puppetlabs/beaker#7d0cccf22cc7289113de862fe4a4494872f574bb')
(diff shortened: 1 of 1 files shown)


haus:
:+1: I'm in.

joshcooper:
Since only beaker 2.8.0 has been released, this will cause puppet to test against an earlier beaker version: https://github.com/puppetlabs/beaker/compare/beaker2.8.0...master.

However, the new 2.8.1 beaker is supposed to be tagged end of day tomorrow (where master currently is), so worst case, we'll be back where we started tomorrow.

joshcooper:
Also I sanity tested rhel7 and win2012r2, made sure the presuite passed and a handful of `with_puppet_running_on` acceptance tests passed, and agent commands, e.g. `puppet agent -td`, etc

johnduarte:
+1 for me.

PR #3797, opened 2015-04-07
(PUP-4385) Add support for 1-6 hex digits in \u escape

hlindberg:
Before this it was not possible to enter \u escapes unless the
char was expressed with exactly 4 hex digits.
1
This commit makes it possible to use 1-6 (max 10ffff value) hex
digits by enclosing them in braces. Thus "\u{1f452}" results in
an emoji hat icon.


hlindberg:
Hold off on merging this until a decision has been made if this should only go to master (and not 3.x).

hlindberg:
This can go into 3.x now and then merged to master

PR #3799, opened 2015-04-08
(DOCUMENT-318) Add note about unless taking arrays

petems:


kylog:
@peterhuene @melissa: I just noticed this PR. It hasn't been covered in PR triage because it has the 'PL' label, but @petems isn't on a team that merges to puppet despite working at PL :)

Should we drop the PL label from this (and #3800) and/or treat these as part of PR triage?

melissa:
@kylog oops, that was my mistake. Thanks for catching it! I say we do what you suggest.

peterhuene:
@petems Would you mind terribly amending the commit message to be prefixed with `(docs)` for this one?  Sorry this fell off our radar.  I'd like to get this merged in.

petems:
@peterhuene Done! :+1:

joshcooper:
for future reference, just made sense it was "all" commands need to return non-zero exit codes:

```
0 ~/work/puppet (stable)  $ be puppet apply -e 'exec { "echo oh no": path => $path, unless => ["/usr/bin/false","/usr/bin/true"] }'
Notice: Compiled catalog for arcturus.corp.puppetlabs.net in environment production in 0.43 seconds
Notice: Applied catalog in 0.04 seconds
```

vs

```
0 ~/work/puppet (stable)  $ be puppet apply -e 'exec { "echo oh no": path => $path, unless => ["/usr/bin/false","/usr/bin/false"] }'
Notice: Compiled catalog for arcturus.corp.puppetlabs.net in environment production in 0.43 seconds
Notice: /Stage[main]/Main/Exec[echo oh no]/returns: executed successfully
```

PR #3800, opened 2015-04-08
(DOCUMENT-319) Clarification on unless and only

petems:
Could probably clearer that the `unless` and `only` parameters have all the same parameters as the main exec, not just the path.

### DIFF
--- a/lib/puppet/type/exec.rb
+++ b/lib/puppet/type/exec.rb
@@ -413,7 +413,8 @@ module Puppet
         `grep` determines it's already there.
 
         Note that this command follows the same rules as the main command,
-        which is to say that it must be fully qualified if the path is not set.
+        such as which user and group it's run as.
+        This also means it must be fully qualified if the path is not set.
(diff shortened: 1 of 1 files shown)


joshcooper:
Thanks @petems, I think these days we are prefixing with (docs) in the commit as opposed to the DOCUMENT ticket? /cc @kylog @nfagerlund

kylog:
Yes please. You can still keep the ticket if it's helpful like: "(docs) (DOCUMENT-1234) Fix the thing".

nfagerlund:
Yeah, several of the people governing this repo didn't like foreign ticket numbers being referenced in commit messages. My practice has been to prefix with (docs), and add a note in the PR requesting that whoever merges the PR resolve the DOC or DOCUMENT ticket#.

nfagerlund:
As for the change itself, +1 assuming someone has tested and proven this! Are there specs or acceptance tests that show the intended behavior? I remember those sub-execs being squirrelly somehow, but I can't remember the details.

PR #3801, opened 2015-04-08
(maint) decrease build targets

melissa:
With the switch to puppet 4, we are no longer building and shipping
individual package. Instead, we're shipping puppet as a part of the
puppet-agent package. This means we can reduce the number of cow and
mock build targets to only LTS releases (for community members who are
using the automation in the packaging repo to build puppet). We can also
stop building the dmg and the msi, as puppet will be delivered to those
platforms also with puppet-agent.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -2,46 +2,19 @@
 packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
 packaging_repo: 'packaging'
 default_cow: 'base-squeeze-i386.cow'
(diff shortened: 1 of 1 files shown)


melissa:
@joshcooper I'm not sure if this is taking what we talked about too far, but this PR is pulling out a lot of stuff from build targets that we should no longer need. I'm not sure if this change will break things (ci, etc) or not... @haus also should participate in this. 

Once https://github.com/puppetlabs/packaging/pull/388 is merged, we can update this file again to make sure that we never build deb/rpm packages when an uber_build job is kicked off. That will leave us with only building the gem and tar

melissa:
Also I opened https://github.com/puppetlabs/puppet/pull/3802 against the stable branch

kylog:
@melissa we generally target stable and merge up, rather than open separate PRs.

Also I just closed #3665, since this pretty much answers the question asked there :)

melissa:
targeting stable for this

PR #3802, opened 2015-04-08
(PUP-4410) decrease build targets

melissa:
With the switch to puppet 4, we are no longer building and shipping
individual package. Instead, we're shipping puppet as a part of the
puppet-agent package. This means we can reduce the number of cow and
mock build targets to only LTS releases (for community members who are
using the automation in the packaging repo to build puppet). We can also
stop building the dmg and the msi, as puppet will be delivered to those
platforms also with puppet-agent.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -1,47 +1,20 @@
 ---
 packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
 packaging_repo: 'packaging'
-default_cow: 'base-squeeze-i386.cow'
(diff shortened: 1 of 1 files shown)


kylog:
:+1:

melissa:
From the PR against master:

@joshcooper I'm not sure if this is taking what we talked about too far, but this PR is pulling out a lot of stuff from build targets that we should no longer need. I'm not sure if this change will break things (ci, etc) or not... @haus also should participate in this.

Once puppetlabs/packaging#388 is merged, we can update this file again to make sure that we never build deb/rpm packages when an uber_build job is kicked off. That will leave us with only building the gem and tar

melissa:
@haus we should be able to merge this. Since my packaging PR has been merged, this shouldn't break anything. Though, it looks like we're not even running the uber_build job anywhere for this branch of puppet

haus:
:+1:

PR #3803, opened 2015-04-08
(maint) Attach pe classifier groups to Production environment group

jpartlow:
Runs of PE Puppet Acceptance for 3.8.x began failing several environment
tests after the new "Production environment" group was added to the
default pe classifications in PE-8444.  This new "Production
environment" group set the production environment for all pe nodes.  The
failing environment tests were making us of the helper method
lib/puppet/acceptance/classifier_utils#create_group_for_nodes to
create test groups in the classifier setting environment for nodes
under test.  But the method was setting the parent to the default root
group, and this was causing the classifier to return a conflict error
due to two separate groups attempting to override environment for the
same nodes.

Error: Could not retrieve catalog from remote server: Error 400 on
SERVER: Failed when searching for node
g4rlqrppmmi75av.delivery.puppetlabs.net: Classification of
g4rlqrppmmi75av.delivery.puppetlabs.net failed due to a classification
conflict: The node was classified into groups named "PE Certificate
Authority", "PE Console", "PE PuppetDB", "Production environment", "PE
ActiveMQ Broker", "PE Agent", "PE MCollective", "Special Environment",
and "PE Master" that defined conflicting values for the environment.

(In this case the conflict was between the groups "Production
environment" and "Special Environment" created by the test)

The fix was to get the new "Production environment"'s uuid and have it
be the parent of the test groups being created.

### DIFF
--- a/acceptance/lib/puppet/acceptance/classifier_utils.rb
+++ b/acceptance/lib/puppet/acceptance/classifier_utils.rb
@@ -17,6 +17,24 @@ module Puppet
         @classifier_utils_tmpdirs ||= []
       end
 
+      # PE creates a "Production environment" group during installation which
+      # all nodes are a member of by default.  This method just looks up this
+      # group and returns its uuid so that other methods may reference it.
+      def get_production_environment_group_uuid
+        step "Get classifier groups so we can locate the 'Production environment' group"
+        response = classifier_handle.get("/v1/groups")
+        assert_equal(200, response.code, "Unable to get classifer groups: #{response.body}")
+
+        groups_json = response.body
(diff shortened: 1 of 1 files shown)


jpartlow:
So the first question for this is whether it should be merged to master.

It will be needed in the 3.7 branch, which will then need to be merged to pe-puppet/3.8.x so that the PE Puppet Acceptance can pass.

So I've got this backwards, this should merge to 3.7, and then be merged up through stable/master as needed.  I'll close and reopen

jpartlow:
Closed in favor of #3804

PR #3805, opened 2015-04-09
(PUP-2455) Fix logic in ticket_2455 acceptance test

geoffnichols:
Prior to this commit, the ticket_2455 acceptance test
fails if there are no Solaris agents.

With no valid agents, the interaction of 'skip_test'
after the setup of 'with_puppet_running_on' causes
the whole test to fail.

This commit updates the acceptance test to skip the setup
if there are no Solaris agents.


justinstoller:
So I think you will still want to do a
```ruby
next unless agent['platform'] =~ /solaris/
```
when iterating over agents.

The guard you've put in the test will skip the whole thing if there are no valid agents, but if there are a mix of agent platforms (like if the master also has the agent role), then this will fail trying to execute solaris specific commands on agents that are not solaris.

geoffnichols:
Thanks @justinstoller. This is updated. What's the best way to validate this?

geoffnichols:
@branan @kylog I think this should fix the issue we saw with the acceptance test last night. /cc @melissa @justinstoller

branan:
This looks right to me

justinstoller:
lgtm :+1:

PR #3806, opened 2015-04-09
(PUP-4343) Add test(s) for file resource links attribute

kurtwall:
The File resource is unevenly tested in acceptance and spec. This adds tests for the links attribute.

### DIFF
--- a/spec/integration/type/file_spec.rb
+++ b/spec/integration/type/file_spec.rb
@@ -1528,4 +1528,134 @@ describe Puppet::Type.type(:file), :uses_checksums => true do
 
     raise "Got #{status_failures.length} failure(s) while applying: #{status_fail_msg}"
   end
+
+  describe "copying a file that is a link to a file", :if => Puppet.features.manages_symlinks? do
+    let(:target) { tmpfile('target') }
+    let(:link) { tmpfile('link') }
+    let(:copy) { tmpfile('copy') }
+    it "should copy the target of the link if :links => follow" do
+      catalog.add_resource described_class.new(
+        :name => target,
+        :ensure => "present",
+        :content => "Jenny I got your number / I need to make you mine")
+      catalog.add_resource described_class.new(
+        :name => link,
+        :ensure => "link",
+        :target => target)
+      catalog.add_resource described_class.new(
+        :name => copy,
+        :ensure => "present",
+        :source => link,
+        :links => "follow")
+      catalog.apply
+      expect(Puppet::FileSystem.file?(copy))
+      expect(File.read(target)).to eq(File.read(copy))
+    end
+
+    it "should copy the link itself if :links => manage" do
+      catalog.add_resource described_class.new(
+        :name => target,
+        :ensure => "present",
+        :content => "Jenny I got your number / I need to make you mine")
+      catalog.add_resource described_class.new(
+        :name => link,
+        :ensure => "link",
+        :target => target)
+      # FIXME: do I need to clean up from the previous test?
+      catalog.add_resource described_class.new(
+        :name => copy,
+        :ensure => "present",
+        :source => link,
+        :links => "manage")
+      catalog.apply
+      expect(Puppet::FileSystem.symlink?(copy))
+      expect(File.read(link)).to eq(File.read(copy))
+    end
+  end
+
+  describe "copying a file that is a link to a directory", :if => Puppet.features.manages_symlinks? do
+    let(:target) { tmpdir('target') }
+    let(:link) { tmpfile('link') }
+    let(:copy) { tmpfile('copy') }
+    context "when the recurse attribute is false" do
+      it "should copy the top-level directory if :links => follow" do
+        catalog.add_resource described_class.new(
+          :name => target,
+          :ensure => "directory")
+        catalog.add_resource described_class.new(
+          :name => link,
+          :ensure => "link",
+          :target => target)
+        catalog.add_resource described_class.new(
+          :name => copy,
+          :ensure => "present",
+          :source => link,
+          :recurse => false,
+          :links => "follow")
+        catalog.apply
+        expect(Puppet::FileSystem.directory?(copy))
+        expect(File).to be_directory(copy)
+      end
+    
+      it "should copy the link itself if :links => manage" do
+        catalog.add_resource described_class.new(
+          :name => target,
+          :ensure => "directory")
+        catalog.add_resource described_class.new(
+          :name => link,
+          :ensure => "link",
+          :target => target)
+        catalog.add_resource described_class.new(
+          :name => copy,
+          :ensure => "present",
+          :source => link,
+          :recurse => false,
+          :links => "manage")
+        catalog.apply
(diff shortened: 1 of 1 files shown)


spec/integration/type/file_spec.rb
```diff
@@ -1528,4 +1528,161 @@ def pretty_transaction_error(transaction)
...
+      # FIXME: do I need to clean up from the previous test?
+      catalog.add_resource described_class.new(
+        :name => copy,
+        :ensure => "present",
+        :source => link,
+        :links => "manage")
+      catalog.apply
+      expect(File.ftype(link)).to eq(File.ftype(copy))
+      expect(File.read(link)).to eq(File.read(copy))
+      expect(File.symlink?(copy))
```
joshcooper: Does this assert that `symlink?` returns true (I'm not sure what the default behavior is when you don't specify `.to eq(true)`.

Also since it's a predicate method, if the expectation fails, it will just say false doesn't equal true. You can output better error messages using this form:

```
expect(File).to be_symlink(copy)
```

spec/integration/type/file_spec.rb
```diff
@@ -1528,4 +1528,161 @@ def pretty_transaction_error(transaction)
 
     raise "Got #{status_failures.length} failure(s) while applying: #{status_fail_msg}"
   end
+
+  describe "copying a file that is a link to a file" do
```
joshcooper: This should be guarded to platforms that support `manages_symlinks` since 2003 does not (and newer versions don't if you're not running with privileges):

```
it "does not package with a symlink", :if => Puppet.features.manages_symlinks? do
```

spec/integration/type/file_spec.rb
```diff
@@ -1528,4 +1528,161 @@ def pretty_transaction_error(transaction)
...
+        :name => link,
+        :ensure => "link",
+        :target => target)
+      catalog.add_resource described_class.new(
+        :name => copy,
+        :ensure => "present",
+        :source => link,
+        :links => "follow")
+      catalog.apply
+      expect(File.ftype(target)).to eq(File.ftype(copy))
```
joshcooper: I'm not sure that ruby's `File.ftype` will return the correct thing for a symlink on Windows. I'd recommend using `Puppet::FileSystem.file?` or `Puppet::FileSystem.symlink?` instead.

spec/integration/type/file_spec.rb
```diff
@@ -1528,4 +1528,161 @@ def pretty_transaction_error(transaction)
...
+        :ensure => "link",
+        :target => target)
+      # FIXME: do I need to clean up from the previous test?
+      catalog.add_resource described_class.new(
+        :name => copy,
+        :ensure => "present",
+        :source => link,
+        :links => "manage")
+      catalog.apply
+      expect(File.ftype(link)).to eq(File.ftype(copy))
```
joshcooper: similar comment about ftype

spec/integration/type/file_spec.rb
```diff
@@ -1528,4 +1528,161 @@ def pretty_transaction_error(transaction)
...
+      expect(File.read(link)).to eq(File.read(copy))
+      expect(File.symlink?(copy))
+    end
+  end
+
+  describe "copying a file that is a link to a directory" do
```
joshcooper: Similar comments here about `manages_symlinks`, `ftype` and setting expects on predicate methods.

spec/integration/type/file_spec.rb
kurtwall: Done

spec/integration/type/file_spec.rb
kurtwall: Done.

spec/integration/type/file_spec.rb
kurtwall: If I guard for `Puppet.features.manages_symlinks?`, is it still necessary to use `Puppet::Filesystem.file?` or `Puppet::FileSystem.symlink?`? That is, if Windows _does_ support symlinks, does `File.ftype()` still do the wrong thing?

joshcooper: Yep, `File.ftype` on Windows works for files and directories, but not symlinks on systems that support them:

```
C:\work\puppet> bundle exec puppet resource file C:\work\puppet\link ensure=link target=c:\work\puppet\target
Notice: /File[C:\work\puppet\link]/ensure: created
file { 'C:\work\puppet\link':
  ensure => 'link',
  target => 'c:\work\puppet\target',
}
C:\work\puppet> bundle exec irb
irb(main):001:0> File.ftype('c:\work\puppet\link')
=> "file"
irb(main):002:0> require 'puppet'
=> true
irb(main):003:0> Puppet::FileSystem.symlink?('c:\work\puppet\link')
=> true
```

And actually a better way of doing this would be to use:

```
irb(main):008:0> Puppet::FileSystem.lstat('c:\work\puppet\link').ftype
=> "link"
```

spec/integration/type/file_spec.rb
```diff
@@ -1528,4 +1528,135 @@ def pretty_transaction_error(transaction)
 
     raise "Got #{status_failures.length} failure(s) while applying: #{status_fail_msg}"
   end
+
+  describe "copying a file that is a link to a file" do
+    let(:target) { tmpfile('target') }
+    let(:link) { tmpfile('link') }
+    let(:copy) { tmpfile('copy') }
+    #it "should copy the target of the link if :links => follow", :if => Puppet.features.manages_symlinks? do
```
er0ck: probably should remove this comment

spec/integration/type/file_spec.rb
```diff
@@ -1528,4 +1528,134 @@ def pretty_transaction_error(transaction)
 
     raise "Got #{status_failures.length} failure(s) while applying: #{status_fail_msg}"
   end
+
+  describe "copying a file that is a link to a file", :if => Puppet.features.manages_symlinks? do
+    let(:target) { tmpfile('target') }
+    let(:link) { tmpfile('link') }
+    let(:copy) { tmpfile('copy') }
+    it "should copy the target of the link", :if => Puppet.features.manages_symlinks? do
```
er0ck: i believe you only need the confines on the describe blocks

kurtwall: Done.

spec/integration/type/file_spec.rb
```diff
@@ -1528,4 +1528,134 @@ def pretty_transaction_error(transaction)
 
     raise "Got #{status_failures.length} failure(s) while applying: #{status_fail_msg}"
   end
+
+  describe "copying a file that is a link to a file", :if => Puppet.features.manages_symlinks? do
+    let(:target) { tmpfile('target') }
+    let(:link) { tmpfile('link') }
+    let(:copy) { tmpfile('copy') }
+    it "should copy the target of the link if :links => follow" do
```
er0ck: in general, ruby likes strings without interpolation in single quotes, but we don't really enforce that here, so i'm cool with it.

PR #3778, opened 2015-03-31
(PUP-4351) Default service type and package provider not set on CumulusL...

LeslieCarr:
...inux

This fix adds CumulusLinux to the defaultfor for its package and service
provider.

Thanks also to Barry Peddycord (@isharacomix) for work on this patch

PR #3807, opened 2015-04-10
(PUP-4351) Cumulus Linux should set defaults

melissa:
We've recently updated Cumulus Linux to default to the apt package
provider and debian service type. This commit adds in spec tests for
those changes to verify that those defaults are assigned correctly.

### DIFF
--- a/spec/unit/provider/package/apt_spec.rb
+++ b/spec/unit/provider/package/apt_spec.rb
@@ -19,6 +19,11 @@ describe provider_class do
     provider
   end
 
+  it "should be the default provider on :osfamily => Debian" do
+    Facter.expects(:value).with(:osfamily).returns("Debian")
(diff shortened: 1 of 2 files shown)


melissa:
This pull request is dependent on https://github.com/puppetlabs/puppet/pull/3778

whopper:
We thought these tests might be nice to have after running into a case where we weren't sure if `defaultfor` was handling symbols case-insensitively. These tests verify `defaultfor` is correct for these two providers, so :+1: from me, if there are no objections.

branan:
@melissa it's failing hard in travis with your new tests.

melissa:
@branan yes, like I mentioned earlier, that is expected until #3778 is merged.

branan:
@melissa whoops, I scanned right past that comment in my pre-coffee daze

PR #3808, opened 2015-04-12
(PUP-3480) Python 3 package provider (pip3).

karmux:
Based on Python 2 package provider (pip).

### DIFF
--- a/lib/puppet/provider/package/pip3.rb
+++ b/lib/puppet/provider/package/pip3.rb
@@ -1,0 +1,18 @@
+# Puppet package provider for Python's `pip3` package management frontend.
+# <http://pip.openplans.org/>
+
+require 'puppet/provider/package/pip'
+
+Puppet::Type.type(:package).provide :pip3,
+  :parent => :pip do
+
+  desc "Python packages via `pip3`.
+
+  This provider supports the `install_options` attribute, which allows command-line flags to be passed to pip3.
+  These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
+  or an array where each element is either a string or a hash."
+
+  def self.cmd
+    "pip3"
+  end
+end
--- a/spec/unit/provider/package/pip3_spec.rb
+++ b/spec/unit/provider/package/pip3_spec.rb
@@ -1,8 +1,8 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
 
-provider_class = Puppet::Type.type(:package).provider(:pip)
-osfamilies = { ['RedHat', '6'] => 'pip-python', ['RedHat', '7'] => 'pip', ['Not RedHat', nil] => 'pip' }
+provider_class = Puppet::Type.type(:package).provider(:pip3)
+osfamilies = { ['All', nil] => 'pip3' }
 
 describe provider_class do
 
@@ -21,7 +21,7 @@ describe provider_class do
       expect(provider_class.parse("real_package==1.2.5")).to eq({
         :ensure   => "1.2.5",
         :name     => "real_package",
-        :provider => :pip,
+        :provider => :pip3,
       })
     end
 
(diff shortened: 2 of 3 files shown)


karmux:
CLA has been signed.

branan:
@karmux Could this be done as a subclass of the pip provider that simply overrides the command to execute? See the `puppet_gem` provider for how this is already done.

karmux:
@branan , refactored it.

lib/puppet/provider/package/pip3.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  # Execute a `pip3` command.  If Puppet doesn't yet know how to do so,
+  # try to teach it and if even that fails, raise the error.
+  private
+  def lazy_pip(*args)
+    pip3 *args
+  rescue NoMethodError => e
+    if pathname = which(self.class.cmd)
+      self.class.commands :pip3 => pathname
+      pip3 *args
```
melissa: Is there any reason that this can't be ` #{self.class.cmd}`? If that were the case, we could avoid the duplication of this block

branan: It looks like the original `lazy_pip` is already using `self.class.cmd`, so I think it already should just work correctly without a duplicated method.

peterhuene: @karmux Any update on the feedback here?  It seems that this method isn't necessary for the implementation and the base method can be used instead.

karmux: I have to fix 10 tests after removing this method. It seems to actually work without overriding it in pip3, so the tests must be wrong.

lib/puppet/provider/package/pip3.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  desc "Python packages via `pip3`.
+
+  This provider supports the `install_options` attribute, which allows command-line flags to be passed to pip3.
+  These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
+  or an array where each element is either a string or a hash."
+
+  def self.cmd
+    if Facter.value(:osfamily) == "RedHat" and Facter.value(:operatingsystemmajrelease).to_i < 7
+      "pip-python3"
```
melissa: I'm not convinced there is a command called `pip-python3`. It could easily be that I just wasn't digging in the right places, but poking around on a centos 6 machine, there aren't any python3 packages readily available, but you can build from source which gives you `pip3`, not `pip-python3`. I know this logic is also in the pip provider, but I'm not convinced it's what we want here. Thoughts?

melissa: For clarification, I was thinking we should be checking if `pip-python3` is available, if not, check if `pip3` is available, rather than just looking for `pip-python3` for el < 7

melissa:
Ping @karmux have you had a chance to look over my suggestion?

karmux:
@melissa I installed CentOS 6 and I can confirm there is no pip-python3 in Red Hat 6 and older. Pushed a fix for that. I kept osfamilies in unit test so somebody can add new OSes in case of pip3 is not pip3.

melissa:
:+1: from me! Thanks for all your work on this PR, @karmux . Hopefully we can get it merged soon!

peterhuene:
Hi @karmux.  This looks good to merge pending some fixups to the commits.  Would you mind squashing the commits into a single commit like `(PUP-3480) Implement Python 3 package provider (pip3).`?  There are three commits that don't have ticket number prefixes and these changes make sense as a single commit.  Thanks!

spec/unit/provider/package/pip3_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+
+describe provider_class do
+
+  before do
+    @resource = Puppet::Resource.new(:package, "fake_package")
+    @provider = provider_class.new(@resource)
+    @client = stub_everything('client')
+    @client.stubs(:call).with('package_releases', 'real_package').returns(["1.3", "1.2.5", "1.2.4"])
+    @client.stubs(:call).with('package_releases', 'fake_package').returns([])
+    XMLRPC::Client.stubs(:new2).returns(@client)
```
joshcooper: The XMLRPC line can be removed from `pip_spec.rb` and `pip3_spec.rb`

kylog: I get spec fails if I remove that stub from `pip_spec.rb`.

joshcooper: My bad, I didn't realize the pip(3) providers used XMLRPC. I thought that was left over from when puppet did.

peterhuene:
@joshcooper Anything else that needs to be addressed before I merge this?

joshcooper:
:+1:

peterhuene:
Hi @karmux.  Thanks again for the contribution!  This has been merged into master and should appear in the Puppet 4.2 release.

karmux:
@peterhuene , I'm happy to see it merged :)

PR #3810, opened 2015-04-13
(maint) Remove Fedora 19 from build defaults

melissa:
Fedora 19 went EOL on 2015-01-06. We should not longer be providing
packages for this platform.


kylog:
:+1: 

/cc @justinstoller to verify we're not testing f19 anywhere. (Pretty sure we're not.)

PR #3812, opened 2015-04-13
(maint) Checksum in binary mode

joshcooper:


MikaelSmith:
:+1: Both commits look good.

PR #3813, opened 2015-04-13
(PUP-585, PUP-4404) Add AppVeyor CI configuration, fix specs expecting Administrator user

MikaelSmith:
Windows spec tests cover behavior that's platform-specific and not
exposed by running spec tests on other CI platforms. This sometimes
leads to missing spec failures on pull requests that primarily affect
Windows.

Add AppVeyor configuration to run spec tests on Windows.

- (PUP-4404) Use current user for SID spec test
- (PUP-4404) Use current user for security specs

### DIFF
--- a/spec/integration/util/windows/security_spec.rb
+++ b/spec/integration/util/windows/security_spec.rb
@@ -15,7 +15,6 @@ describe "Puppet::Util::Windows::Security", :if => Puppet.features.microsoft_win
     @sids = {
       :current_user => Puppet::Util::Windows::SID.name_to_sid(Puppet::Util::Windows::ADSI::User.current_user_name),
       :system => Win32::Security::SID::LocalSystem,
-      :admin => Puppet::Util::Windows::SID.name_to_sid("Administrator"),
       :administrators => Win32::Security::SID::BuiltinAdministrators,
       :guest => Puppet::Util::Windows::SID.name_to_sid("Guest"),
       :users => Win32::Security::SID::BuiltinUsers,
@@ -46,6 +45,14 @@ describe "Puppet::Util::Windows::Security", :if => Puppet.features.microsoft_win
       winsec.set_mode(mode & ~WindowsSecurityTester::S_IRWXG, path)
     else
       winsec.set_group(sids[:guest], path)
+    end
+  end
+
+  def check_group_depending_on_current_user(path)
+    if sids[:current_user] == sids[:system]
+      expect(winsec.get_group(path)).to eq(sids[:system])
+    else
+      expect(winsec.get_group(path)).to eq(sids[:guest])
     end
   end
 
@@ -387,9 +394,13 @@ describe "Puppet::Util::Windows::Security", :if => Puppet.features.microsoft_win
         end
 
         describe "#owner=" do
+          it "should accept the guest sid" do
+            expect(winsec.get_owner(path)).to eq(sids[:guest])
+          end
+
           it "should accept a user sid" do
-            winsec.set_owner(sids[:admin], path)
-            expect(winsec.get_owner(path)).to eq(sids[:admin])
+            winsec.set_owner(sids[:current_user], path)
+            expect(winsec.get_owner(path)).to eq(sids[:current_user])
           end
 
           it "should accept a group sid" do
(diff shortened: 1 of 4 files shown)


peterhuene:
:+1:

MikaelSmith:
Depends on @joshcooper's spec fixes in #3812 and an upcoming PR from @Iristyle.

kylog:
Latest fail looks legit! Woot!

kylog:
Since travis was getting backed up, I killed the travis jobs generated for intermediate pushes of this PR.

Also I merged #3812. So once that last PR comes in, we can re-run this.

joshcooper:
awesome, good stuff! :+1:

MikaelSmith:
Several spec tests expect an Administrator user to exist; it doesn't on the AppVeyor VMs.

MikaelSmith:
This definitely needs sign-off from @joshcooper, @Iristyle, and @ferventcoder. The conclusion I came to is we were expecting an Administrator user to exist, but didn't have any reason to require it - the tests that used it didn't expect it to be an actual member of the Administrators group.

spec/unit/util/windows/sid_spec.rb
```diff
@@ -15,7 +15,7 @@
   context "#octet_string_to_sid_object" do
     it "should properly convert an array of bytes for the local Administrator SID" do
       host = '.'
-      username = 'Administrator'
+      username = ENV['USERNAME']
```
Iristyle: You should be able to use our helper method here, which calls the `[GetUserName](http://msdn.microsoft.com/en-us/library/windows/desktop/ms724432(v=vs.85).aspx)` Win32 API call under the hood.  Should be a bit more reliable than relying on ENV vars, which I don't know if AppVeyor monkeys with or not....

```ruby
Puppet::Util::Windows::ADSI.current_user_name
```

spec/integration/util/windows/security_spec.rb
```diff
@@ -388,8 +387,8 @@ def grant_everyone_full_access(path)
 
         describe "#owner=" do
           it "should accept a user sid" do
-            winsec.set_owner(sids[:admin], path)
-            expect(winsec.get_owner(path)).to eq(sids[:admin])
+            winsec.set_owner(sids[:current_user], path)
```
Iristyle: I think the only potential problem here is that previously we created files with one owner, then used `set_owner` to change the owner (on dev boxes it *probably* went from `Administrator` to `Administrator`, but in Jenkins it should be from the Jenkins service account -> `Administrator`).

By leaving the user the same we're actually not quite testing the same thing.  I wonder if we can switch this to `:guest` instead of `:current_user` and continue to test the previous mutating behavior of the test.

MikaelSmith: I'll give it a try.

spec/integration/util/windows/security_spec.rb
```diff
@@ -388,8 +387,8 @@ def grant_everyone_full_access(path)
...
+            expect(winsec.get_owner(path)).to eq(sids[:current_user])
```
joshcooper: One potential issue is that if `set_owner` fails, then the file/dir may be owned by the current user, so `get_owner` may continue to pass. I think it's more robust to set the owner to an account that we know isn't the current owner.

Unfortunately, the current owner will depend on the default DACL for the user that is executing the tests. For regular user accounts (that are not running with elevated privileges), the default DACL will assign the owner to be the current user. For accounts like LocalSystem (which is what jenkins CI runs as), the owner will be `Administrators`:

```
irb(main):012:0* FileUtils.touch('c:\work\puppet\footest')
=> ["c:\\work\\puppet\\footest"]
irb(main):013:0> Puppet::Util::Windows::Security.get_security_descriptor('c:\work\puppet\footest')
=> BUILTIN\Administrators
WIN-QP47VOHA2P4\None
  BUILTIN\Users:(I)                             0x1f01ff
  NT AUTHORITY\SYSTEM:(I)                       0x1f01ff
  BUILTIN\Administrators:(I)                    0x1f01ff
```

Here the owner is `Administrators`, group is `None`, and there are 3 inherited access control entries.

MikaelSmith: Got it; I'll look more into why we `set_owner(:guest, path)` before the tests. Maybe we can change that and have the test verify it works.

MikaelSmith: This test category seems to be designed to test the Administrator changing permissions away from :guest, so I left that part and added tests to confirm that ownership was actually set to guest.

joshcooper:
Can we retarget for stable (4.0.1)?

MikaelSmith:
Yup, opened #3821

PR #3814, opened 2015-04-13
(PUP-4111) environment scenario tests should assert success

er0ck:
This change ensures all the environment tests assert on the proper
outputs to finalize the tests that gather results into one step.
Prior to this change, some tests could result in false passes.
This change also fixes the "bad" scenario test, which tests for a bad
environmentpath.

### DIFF
--- a/acceptance/tests/environment/environment_scenario-non_existent.rb
+++ b/acceptance/tests/environment/environment_scenario-non_existent.rb
@@ -46,9 +46,9 @@ expectations = {
   },
   :puppet_agent => {
     :exit_code => 1,
-    :matches => [%r{Warning.*404.*Could not find environment '#{env}'},
+    :matches => [%r{(Warning|Error).*(404|400).*Could not find environment '#{env}'},
                  %r{Could not retrieve catalog; skipping run}],
   }
 }
 
-review_results(results,expectations)
+assert_review(review_results(results,expectations))
--- a/acceptance/tests/environment/environment_scenario-bad.rb
+++ b/acceptance/tests/environment/environment_scenario-bad.rb
@@ -1,4 +1,4 @@
-test_name "Test behavior of directory environments when environmentpath is set to a non-existent directory"
+test_name 'Test behavior of directory environments when environmentpath is set to a non-existent directory'
 require 'puppet/acceptance/environment_utils'
 extend Puppet::Acceptance::EnvironmentUtils
 require 'puppet/acceptance/classifier_utils'
@@ -6,26 +6,25 @@ extend Puppet::Acceptance::ClassifierUtils
 
 classify_nodes_as_agent_specified_if_classifer_present
 
-step "setup environments"
+step 'setup environments'
 
 stub_forge_on(master)
 
-testdir = create_tmpdir_for_user master, "confdir"
+testdir = create_tmpdir_for_user master, 'confdir'
 puppet_conf_backup_dir = create_tmpdir_for_user(master, "puppet-conf-backup-dir")
 
 apply_manifest_on(master, environment_manifest(testdir), :catch_failures => true)
 
-step  "Test"
+step  'Test'
 master_opts = {
   'main' => {
     'environmentpath' => '/doesnotexist',
   }
 }
-general = [ master_opts, testdir, puppet_conf_backup_dir, { :directory_environments => true } ]
-env = 'doesnotexist'
+env = 'testing'
 path = master.puppet('master')['codedir']
 
-results = use_an_environment("testing", "bad environmentpath", master_opts, testdir, puppet_conf_backup_dir, :directory_environments => true)
+results = use_an_environment(env, 'bad environmentpath', master_opts, testdir, puppet_conf_backup_dir, :directory_environments => true)
 
 expectations = {
   :puppet_config => {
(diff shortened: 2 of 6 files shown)


acceptance/tests/environment/environment_scenario-non_existent.rb
```diff
@@ -46,9 +46,9 @@
   },
   :puppet_agent => {
     :exit_code => 1,
-    :matches => [%r{Warning.*404.*Could not find environment '#{env}'},
+    :matches => [%r{(Warning|Error).*(404|400).*Could not find environment '#{env}'},
```
jpartlow: I'm a little surprised that we sometimes get a 400 here instead of a 404?  Do you know why, Eric?

jpartlow:
@er0ck I've looked through the changes and I have some questions.  By passing /doesnotexist into use_an_environment, you're not changing the master's environmentpath, you're just changing where the code would look to find files to copy into the default $confdir.  The safely_shadow_directory_contents_and_yield() nastiness was put in place to work around being unable to point to a new confdir (because we could not pass commandline args to puppetserver), which is why the tests go through the rigmarole of creating an env structure in the testdir and then backing up the real puppetserver confdir, copying in the test code, testing, and then restoring.

So I think there are more changes here than are necessary, and at its heart you just to check that you are asking for a real directory env in the default environmentpath, when the server's $environmentpath setting is set to a non existent directory.

Or perhaps drop the test altogether if it doesn't seem useful.

er0ck:
@jpartlow use_an_environment specifies the environmentpath on the commandline to the agent run inside with_puppet_running_on:
```
 args << "--environmentpath='$confdir/environments'" if directory_environments && agent != master
              on(agent, puppet("agent", *args), :acceptable_exit_codes => (0..255)) do
                agent_results[:puppet_agent] = result
              end
```
i still think it's a useful test, but this was the first way i thought of to make this work without breaking the other env tests.   i believe the commandline switch overrides the master config file.  the test was certainly behaving as if it does (it was using the temporary environmentpath, which exists, and not checking /doesnotexist)

jpartlow:
@er0ck specifying the commandline switch overrides the environmentpath on the agent.  It won't have any effect on the master's environmentpath.

Setting the --environmentpath on the agent is checking that an agent with an environmentpath set does not mess with the agent's environment/code loading process in any unfortunate ways.

acceptance/tests/environment/environment_scenario-non_existent.rb
er0ck: yeah, i was surprised too. maybe it's a "bad request" because we are asking for an environment that doesn't exist?

er0ck:
yeah.  that makes sense, in retrospect.   cool, that means we can change it back...

acceptance/tests/environment/environment_scenario-non_existent.rb
jpartlow: Not sure; do you have the error output in context?

er0ck: here is the output after switching the test back to using the master's environmentpath:
```
huybl681i0j3mx6.delivery.puppetlabs.net (agent) 10:12:05$ puppet agent -t --server tna1l5h77tzypm7.delivery.puppetlabs.net --environment testing --environmentpath='$confdir/environments'
Warning: Unable to fetch my node definition, but the agent run will continue:
Warning: Error 400 on SERVER: Could not find environment 'testing'
Info: Retrieving pluginfacts
Error: /File[/opt/puppetlabs/puppet/cache/facts.d]: Failed to generate additional resources using 'eval_generate': Error 400 on SERVER: Could not find environment 'testing'
Error: /File[/opt/puppetlabs/puppet/cache/facts.d]: Could not evaluate: Could not retrieve file metadata for puppet:///pluginfacts: Error 400 on SERVER: Could not find environment 'testing'
Info: Retrieving plugin
Error: /File[/opt/puppetlabs/puppet/cache/lib]: Failed to generate additional resources using 'eval_generate': Error 400 on SERVER: Could not find environment 'testing'
Error: /File[/opt/puppetlabs/puppet/cache/lib]: Could not evaluate: Could not retrieve file metadata for puppet:///plugins: Error 400 on SERVER: Could not find environment 'testing'
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Could not find environment 'testing'
Error: Could not send report: Error 400 on SERVER: Could not find environment 'testing'
```

jpartlow: Maybe puppet switched from a 404 to a 400.  @hlindberg?

jpartlow:
Anyway, other than the open question about 400/404 response codes, this looks good to me @er0ck :)

acceptance/tests/environment/environment_scenario-non_existent.rb
hlindberg: I am not aware of any recent changes/tickets regarding http status changes - not since we worked on messages propagating to the client - many months ago.

Not sure that 400 is an appropriate response for a missing environment, 404 seems right. Could it be a difference between different kinds of missing (no environment path directory vs. can not find an environment on the path)?

er0ck: I confirmed that both tests (*-bad* and *-non-existent*) produce 400 error.  in the "bad" test, the environmentpath does not exist on the master (nor agent).  in the non-existent test, the environmentpath exists on the master but the environment itself does not.

er0ck:
@jpartlow i believe this is ready to go

jpartlow:
:+1: from me @er0ck

er0ck:
@jpartlow, i think maybe @hlindberg wants you to hit "go" :-)

jpartlow:
@er0ck oh, sure; is stable the right branch?

er0ck:
yeah, i think stable is the right branch for eventual merge up to master.

hlindberg:
Thanks @jpartlow

jpartlow:
now it's got my fingerprints all over it :)

PR #3815, opened 2015-04-13
(PUP-4404) Use current user for SID spec test

MikaelSmith:
The SID generation spec needs the name of a user that exists; that account
doesn't need to have any particular permissions. Previously the test would
expect an 'Administrator' account to exist and hard-coded that name; on
systems where no user named 'Administrator' exists the spec would fail.

Change the user name used in the test to be the current account - accessed
via the USERNAME environment variable - instead of hard-coding
'Administrator'.


kylog:
:+1: 

Although: do you want to fold this into the appveyor PR?

MikaelSmith:
Hmm, guess I could.

PR #3816, opened 2015-04-13
(PUP-4336) add test for puppet apply trace

er0ck:
This change adds a test for --trace using puppet apply.
Other tests use --trace with apply but don't check the output.
Create a dedicated test to protect against possible regressions.

### DIFF
--- a/acceptance/tests/apply/puppet_apply_trace.rb
+++ b/acceptance/tests/apply/puppet_apply_trace.rb
@@ -1,0 +1,7 @@
+test_name 'puppet apply --trace should provide a stack trace'
+
+agents.each do |agent|
+  on(agent, puppet('apply --trace -e "blue < 2"'), :acceptable_exit_codes => 1) do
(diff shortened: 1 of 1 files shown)


kylog:
America Works!

er0ck:
heh.  i need to add Underwood quotes to my PRs.

kylog:
I think FU says it all ;)

acceptance/tests/apply/puppet_apply_trace.rb
```diff
@@ -0,0 +1,7 @@
+test_name 'puppet apply --trace should provide a stack trace'
+
+agents.each do |agent|
+  on(agent, puppet('apply --trace -e "blue < 2"'), :acceptable_exit_codes => [0,1]) do
```
kylog: Is '0' actually an acceptable exit code here?

er0ck: probably not.

er0ck:
@hlindberg fixed the exit code and made a better stack trace matcher (i hope)

PR #3820, opened 2015-04-14
(maint) Fix Solaris ticket 2455 PE acceptance test.

geoffnichols:
Prior to this commit, PE Puppet acceptance testing for
ticket 2455 fails with an "undefined local variable or
method 'agent_service'" error.

This commit updates the acceptance test to fix this issue.

We also update the acceptance test fixture to more reliably
stop the fixture service.

### DIFF
--- a/acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
+++ b/acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
@@ -10,7 +10,7 @@ end
 SCRIPT
 
 sleepy_daemon_initscript = <<INITSCRIPT
-#!/bin/sh
+#!/usr/bin/bash
 FIXTURESERVICE="/tmp/sleepy_daemon"
 start(){
     $FIXTURESERVICE &
@@ -18,7 +18,9 @@ start(){
 
 stop(){
     FIXTUREPID=`ps -ef | grep "$FIXTURESERVICE" | grep -v grep | awk '{print $2}'`
-    kill -9 $FIXTUREPID
+    if [ "x$FIXTUREPID" != "x" ]; then
+      kill -9 ${FIXTUREPID}
+    fi
 }
 
 status(){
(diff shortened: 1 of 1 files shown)


acceptance/tests/ticket_2455_on_solaris_init_provider_should_start_service_in_own_smf_contract.rb
```diff
@@ -112,15 +114,15 @@
...
         step "Stop fixture service on #{agent}"
           on(agent, "#{fixture_service_stop}")
 
         step "Enable puppet service on #{agent}"
-          on(agent, "svcadm enable #{agent_service};sleep 10") do
-            puppet_ctid = on(agent, "svcs -Ho CTID #{agent_service} | awk '{print $1}'").stdout.chomp.to_i
-            service_ctid = on(agent, "ps -eo ctid,args | grep #{fixture_service} | grep -v grep | awk '{print $1}'").stdout.chomp.to_i
+          on(agent, "svcadm enable pe-puppet;sleep 10") do
+            puppet_ctid = on(agent, "svcs -Ho CTID pe-puppet | awk '{print $1}'").stdout.chomp.to_i
+            service_ctid = on(agent, "ps -eo ctid,args | grep pe-puppet | grep -v grep | awk '{print $1}'").stdout.chomp.to_i
```
kylog: This line, to extract `service_ctid`, changed from grepping for `#{fixture_service}` to grepping for `pe-puppet` but it seems like the step below is still assuming `#{fixture_service}` for the comparison. Was this correct?

I don't fully get the context for this test, so apologies if I'm just missing the boat :)

geoffnichols: @kylog You're correct. That's a mistake. I'll validate and update.

geoffnichols:
Retesting this now.

geoffnichols:
In my validation, this test fails (as expected) when run against PE 3.7.2, and passes when run against PE ci-ready builds.

geoffnichols:
@branan If you have a chance, could you please take a peek at this?

shrug:
I can't note the line, it's not from this PR - but the test on line 26 is not testing equality - it's an assignment

shrug:
Never mind, it's bash, that's a valid test. Kind of ambiguous but it will work.

geoffnichols:
I may be confused, but I think that the single =  (with whitespace) is the POSIX-compliant version of the string equality test, but ```==``` is more idiomatic in bash. 

I'm happy to update if needed.

This is from the man page for ```test``` on Solaris:
```
     string1 = string2           True if the strings string1  and
                                 string2 are identical.

     string1 != string2          True if the strings string1  and
                                 string2 are not identical.
```

ScottGarman:
:+1:  I think you'd only use == if you also use double brackets around the test

melissa:
@geoffnichols just confirming, but this works both with foss and pe?

geoffnichols:
@melissa As best as I can tell, yes.

PR #3821, opened 2015-04-15
(PUP-585, PUP-4404) Add AppVeyor, fix specs expecting Administrator user

MikaelSmith:
Windows spec tests cover behavior that's platform-specific and not
exposed by running spec tests on other CI platforms. This sometimes
leads to missing spec failures on pull requests that primarily affect
Windows.

Add AppVeyor configuration to run spec tests on Windows.

(PUP-4404) Use current user for SID spec test
(PUP-4404) Use current user for security specs

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,10 @@
+version: 4.0.0.{build}
+platform:
+    - x64
+install:
+    - SET PATH=C:\Ruby215\bin;%PATH%
+    - gem install bundler --quiet --no-ri --no-rdoc
+    - bundle install --without development
+build: off
+test_script:
+    - bundle exec rspec spec
--- a/spec/integration/util/windows/security_spec.rb
+++ b/spec/integration/util/windows/security_spec.rb
@@ -15,7 +15,6 @@ describe "Puppet::Util::Windows::Security", :if => Puppet.features.microsoft_win
     @sids = {
       :current_user => Puppet::Util::Windows::SID.name_to_sid(Puppet::Util::Windows::ADSI::User.current_user_name),
       :system => Win32::Security::SID::LocalSystem,
-      :admin => Puppet::Util::Windows::SID.name_to_sid("Administrator"),
       :administrators => Win32::Security::SID::BuiltinAdministrators,
       :guest => Puppet::Util::Windows::SID.name_to_sid("Guest"),
       :users => Win32::Security::SID::BuiltinUsers,
@@ -387,9 +386,14 @@ describe "Puppet::Util::Windows::Security", :if => Puppet.features.microsoft_win
         end
 
         describe "#owner=" do
+          it "should accept the guest sid" do
+            winsec.set_owner(sids[:guest], path)
+            expect(winsec.get_owner(path)).to eq(sids[:guest])
+          end
+
           it "should accept a user sid" do
-            winsec.set_owner(sids[:admin], path)
-            expect(winsec.get_owner(path)).to eq(sids[:admin])
+            winsec.set_owner(sids[:current_user], path)
+            expect(winsec.get_owner(path)).to eq(sids[:current_user])
           end
 
           it "should accept a group sid" do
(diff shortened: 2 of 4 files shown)


appveyor.yml
```diff
@@ -0,0 +1,10 @@
+version: 4.0.0.{build}
+platform:
+    - x64
+install:
+    - SET PATH=C:\Ruby215\bin;%PATH%
+    - gem install bundler --quiet --no-ri --no-rdoc
+    - bundle install --without development
+build: off
+test_script:
+    - bundle exec rspec spec
```
ferventcoder: Is there any chance you can set email notifications to maybe our ci email list? or would that not be desirable?

ferventcoder:
I'm +1 to merge on the rest of this. Anecdotally, I use puppet apply to put back the administrator account specifically for running these specs on my vagrant boxes. Now I don't need to do that. :)

appveyor.yml
MikaelSmith: Do we do that for Travis?

MikaelSmith: It's certainly possible, see http://www.appveyor.com/docs/notifications.

ferventcoder: I use if, I'm just curious if this is something that folks would be okay with. 

Here's how I set it up for choco - https://github.com/chocolatey/choco/blob/master/appveyor.yml#L36-L43

spec/integration/util/windows/security_spec.rb
```diff
@@ -387,9 +394,13 @@ def grant_everyone_full_access(path)
...
         describe "#owner=" do
+          it "should accept the guest sid" do
+            expect(winsec.get_owner(path)).to eq(sids[:guest])
+          end
+
           it "should accept a user sid" do
-            winsec.set_owner(sids[:admin], path)
-            expect(winsec.get_owner(path)).to eq(sids[:admin])
+            winsec.set_owner(sids[:current_user], path)
+            expect(winsec.get_owner(path)).to eq(sids[:current_user])
```
joshcooper: I'm a little concerned that if `set_owner` fails to perform the operation, and doesn't raise an exception, then the test will pass. Can we instead set the owner to `Guest`?

spec/integration/util/windows/security_spec.rb
```diff
@@ -407,14 +418,18 @@ def grant_everyone_full_access(path)
...
           it "should accept a group sid" do
             winsec.set_group(sids[:power_users], path)
             expect(winsec.get_group(path)).to eq(sids[:power_users])
           end
 
           it "should accept a user sid" do
-            winsec.set_group(sids[:admin], path)
-            expect(winsec.get_group(path)).to eq(sids[:admin])
+            winsec.set_group(sids[:current_user], path)
+            expect(winsec.get_group(path)).to eq(sids[:current_user])
```
joshcooper: same comment here about setting group to `Guest`

appveyor.yml
MikaelSmith: The feedback I've gotten is generally disable e-mail notifications, but the ci list is a special case. It'd probably get a lot more volume than it does now though, and I'm not sure who watches it.

spec/integration/util/windows/security_spec.rb
MikaelSmith: It starts as Guest, which is why I added the test at line 398 (`expect(winsec.get_owner(path)).to eq(sids[:guest])`) to ensure line 384 (`winsec.set_owner(sids[:guest], path)`) succeeds.

appveyor.yml
ferventcoder: It's more that I get email to my personal account and I would like to self select whether I get that or not.

spec/integration/util/windows/security_spec.rb
MikaelSmith: Line 383 (`set_group_depending_on_current_user(path)`) will set_group to guest (or a restricted permission set). I added line 422 (`check_group_depending_on_current_user(path)`) to verify it succeeded.

spec/integration/util/windows/security_spec.rb
```diff
@@ -407,14 +418,18 @@ def grant_everyone_full_access(path)
         end
 
         describe "#group=" do
+          it "should accept the test group" do
+            check_group_depending_on_current_user(path)
```
MikaelSmith: Actually, if the user is :system then this doesn't help, I'll make this set group to :guest to ensure there's a positive test.

appveyor.yml
ferventcoder: Maybe I can do that per project in appveyor. Last I looked it was a global setting unless it was in the appveyor files themselves.

spec/integration/util/windows/security_spec.rb
joshcooper: Ah, didn't see `set_owner` in the `before :each`, :+1:

ferventcoder:
\o/

PR #3822, opened 2015-04-15
(PUP-4388) Remove START option from /etc/default/puppet on Debian

whopper:
In the new puppet-agent AIO package, services no longer
start on package installation, as there are a few required
configuration steps which must be addressed before the
service can be started.

This means we can remove the agent service START option from
/etc/default/puppet, a relic of old packages where debhelper
started services upon package installation.

### DIFF
--- a/ext/debian/puppet.default
+++ b/ext/debian/puppet.default
@@ -1,9 +1,4 @@
 # Defaults for puppet - sourced by /etc/init.d/puppet
 
-# Enable puppet agent service?
-# Setting this to "yes" allows the puppet agent service to run.
-# Setting this to "no" keeps the puppet agent service from running.
-START=no
-
 # Startup options
 DAEMON_OPTS=""
--- a/ext/debian/puppet.init
+++ b/ext/debian/puppet.init
@@ -22,26 +22,12 @@ test -x $DAEMON || exit 0
 
 . /lib/lsb/init-functions
 
-is_true() {
-    if [ "x$1" = "xtrue" -o "x$1" = "xyes" -o "x$1" = "x0" ] ; then
-       return 0
-    else
-        return 1
-    fi
-}
-
 reload_puppet_agent() {
     start-stop-daemon --stop --quiet --signal HUP --pidfile $PIDFILE --name $PROCNAME
 }
 
 start_puppet_agent() {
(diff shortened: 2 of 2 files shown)


MikaelSmith:
This looks good, but I don't feel super familiar with it. I'd feel more comfortable with @melissa or @haus checking it.

geoffnichols:
I have't looked, but what happens on upgrades here? If Puppet is running, and we upgrade the package, what is the status of the Puppet service post-upgrade (still running, stopped, restarted)?

whopper:
@geoffnichols I'm not sure if upgrades will have an effect here - I'll see about running through a few tests to see what happens.

/cc @haus (we talked about this ticket briefly)

whopper:
@geoffnichols alright, just tested out an upgrade from current puppet-agent -> puppet-agent with this change:

1) In current puppet-agent, set START option to true in `/etc/default/puppet` and run `service puppet start`
`service puppet status` => `agent is running`

2) Upgrade to the newly built package with this change. Dpkg gives us a helpful warning:

    Configuration file `/etc/default/puppet'
     ==> Modified (by you or by a script) since installation.
     ==> Package distributor has shipped an updated version.
       What would you like to do about it ?  Your options are:
        Y or I  : install the package maintainer's version
        N or O  : keep your currently-installed version
          D     : show the differences between the versions
          Z     : start a shell to examine the situation
     The default action is to keep your current version.
    *** puppet (Y/I/N/O/D/Z) [default=N] ? Y

We say 'Y', and the result is:

    Installing new version of config file /etc/default/puppet ...
    Restarting puppet agent.
    root@bqyw08kav2k5avh:~/new# service puppet status
    agent is running.
    cat /etc/default/puppet

    # Defaults for puppet - sourced by /etc/init.d/puppet

    # Startup options
    DAEMON_OPTS=""

So in the end, the service is restarted and then continues to run.

haus:
@whopper you can safely delete the is_true function in the init script as well

whopper:
@haus removed the function and tested again - all is well!

whopper:
Also, I'm retargeting this at stable since the ticket is targeted at 4.0.1

haus:
I'm :+1: here.

PR #3828, opened 2015-04-16
(docs) Unshorten the word 'repository' in the yumrepo type descriptions

juniorsysadmin:
Unshorten 'repository' in the yumrepo type descriptions and be consistent with using the full word.

### DIFF
--- a/lib/puppet/type/yumrepo.rb
+++ b/lib/puppet/type/yumrepo.rb
@@ -167,7 +167,7 @@ Puppet::Type.newtype(:yumrepo) do
   newproperty(:includepkgs) do
     desc "List of shell globs. If this is set, only packages
       matching one of the globs will be considered for
-      update or install from this repo. #{ABSENT_DOC}"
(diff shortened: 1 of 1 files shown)


peterhuene:
Hi @juniorsysadmin.  Thank you for the contribution.  Could you please amend your commit message to be prefixed with "(doc)" per our [contributing guidelines](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md)?  Also, a short explanation of the change in the commit message would be appreciated (we like a title and a message, even when the title is mostly self-explanatory).

Also, it appears there are other instances of using the word "repo" in this file, while "repository" does seem to be the majority.  Would it be possible to fix those other instances as well to make the documentation consistent?  Thanks.

juniorsysadmin:
Sorry, this PR was er a bit sloppy and done in a rush. Should be a bit better now.

peterhuene:
@juniorsysadmin Thanks again for the contribution!

PR #3830, opened 2015-04-16
(PUP-2862) Make assignment in parameter default expr invalid

hlindberg:
Before this it was possible to perform assignment of variables in
a parameter default expression. This is bad because the variable is
set in the calling context!

```
define bad_boy($a = ($b = 10)) { }
```

This commit adds validation and an error for such assignments.

### DIFF
--- a/lib/puppet/pops/validation/checker4_0.rb
+++ b/lib/puppet/pops/validation/checker4_0.rb
@@ -413,13 +413,12 @@ class Puppet::Pops::Validation::Checker4_0
       acceptor.accept(Issues::ILLEGAL_NUMERIC_PARAMETER, o, :name => o.name)
     end
     return unless o.value
-    assignments =
+
     if o.value.is_a?(Puppet::Pops::Model::AssignmentExpression)
       [o.value]
-    else o.value
+    else
       o.value.eAllContents.select {|model| model.is_a? Puppet::Pops::Model::AssignmentExpression }
-    end
-    assignments.each do |assignment|
+    end.each do |assignment|
       acceptor.accept(Issues::ILLEGAL_ASSIGNMENT_CONTEXT, assignment)
     end
   end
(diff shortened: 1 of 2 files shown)


igalic:
it would, imo still be nice to do something like:
```
deine foo (
  $prefix = '/usr/local',
  $libdir = "${prefix}/lib",
) {}
```

lib/puppet/pops/validation/checker4_0.rb
```diff
@@ -412,6 +412,15 @@ def check_Parameter(o)
     if o.name =~ /^(?:0x)?[0-9]+$/
       acceptor.accept(Issues::ILLEGAL_NUMERIC_PARAMETER, o, :name => o.name)
     end
+    assignments =
+    if o.value.is_a?(Puppet::Pops::Model::AssignmentExpression)
+      [o.value]
+    else
+      o.value.eAllContents.select {|model| model.is_a? Puppet::Pops::Model::AssignmentExpression }
```
thallgren: o.value can apparently be nil at this point (this is why the tests are failing).

hlindberg:
@igalic the ability to reference parameter values on the left should be specified and ensured that it works; now it is by chance - no intention of removing that ability.

hlindberg:
@thallgren fix pushed to this PR

igalic:
:+1: @hlindberg.

lib/puppet/pops/validation/checker4_0.rb
```diff
@@ -412,6 +412,16 @@ def check_Parameter(o)
     if o.name =~ /^(?:0x)?[0-9]+$/
       acceptor.accept(Issues::ILLEGAL_NUMERIC_PARAMETER, o, :name => o.name)
     end
+    return unless o.value
+    assignments =
+    if o.value.is_a?(Puppet::Pops::Model::AssignmentExpression)
```
thallgren: This if statement should be indented (this is probably why RuboCop is failing).

hlindberg: no, it is 3 lines down, where an "elsif" should have been used

hlindberg:
Thanks RubyCop. Problem fixed (dead code removed).

PR #3832, opened 2015-04-16
(PUP-4394) Improve erroring for WEBrick server when Puppet user doesn't exist

whopper:
The puppet-agent package does not create the puppet user or group
by default, meaning that attempting to run the webrick server without
manually creating the user results in Puppet failing hard.

This commit updates Puppet's webrick code to first check it the
puppet user exists, and only attempt to switch to it if it does.
Otherwise, we emit an error and exit. Previously, Puppet would
actually try to switch to the non-existent user and throw several
errors. We now only throw one.

### DIFF
--- a/lib/puppet/application/master.rb
+++ b/lib/puppet/application/master.rb
@@ -200,6 +200,7 @@ Copyright (c) 2012 Puppet Labs, LLC Licensed under the Apache 2.0 License
     if options[:rack]
       start_rack_master
     else
+      Puppet.deprecation_warning("The WEBrick Puppet master server is deprecated and will be removed in a future release.")
       start_webrick_master
     end
   end
--- a/lib/puppet/application/master.rb
+++ b/lib/puppet/application/master.rb
@@ -185,11 +185,15 @@ Copyright (c) 2012 Puppet Labs, LLC Licensed under the Apache 2.0 License
     Puppet::SSL::Host.ca_location = :only if Puppet::SSL::CertificateAuthority.ca?
 
     if Puppet.features.root?
-      begin
-        Puppet::Util.chuser
-      rescue => detail
-        Puppet.log_exception(detail, "Could not change user to #{Puppet[:user]}: #{detail}")
-        exit(39)
+      if Puppet::Type.type(:user).new(:name => Puppet[:user]).exists?
+        begin
+          Puppet::Util.chuser
+        rescue => detail
+          Puppet.log_exception(detail, "Could not change user to #{Puppet[:user]}: #{detail}")
+          exit(39)
+        end
+      else
+        raise Puppet::Error.new("Could not change user to #{Puppet[:user]}. User does not exist and is required to continue.")
       end
     end
 
--- a/spec/unit/application/master_spec.rb
+++ b/spec/unit/application/master_spec.rb
@@ -338,6 +338,16 @@ describe Puppet::Application::Master, :unless => Puppet.features.microsoft_windo
         end
       end
 
+      it "should log a deprecation notice when running a WEBrick server" do
+        Puppet.features.stubs(:root?).returns true
+        a_user_type_for("puppet").expects(:exists?).returns true
+        Puppet::Util.expects(:chuser)
+
+        Puppet.expects(:deprecation_warning).with("The WEBrick Puppet master server is deprecated and will be removed in a future release.")
+
+        @master.main
+      end
+
       it "should daemonize if needed" do
         Puppet[:daemonize] = true
 
--- a/spec/unit/application/master_spec.rb
+++ b/spec/unit/application/master_spec.rb
@@ -314,12 +314,28 @@ describe Puppet::Application::Master, :unless => Puppet.features.microsoft_windo
         @master.main
       end
 
-      it "should drop privileges if running as root" do
-        Puppet.features.stubs(:root?).returns true
+      def a_user_type_for(username)
+        user = mock 'user'
+        Puppet::Type.type(:user).expects(:new).with { |args| args[:name] == username }.returns user
+        user
+      end
 
-        Puppet::Util.expects(:chuser)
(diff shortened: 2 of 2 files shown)


lib/puppet/application/master.rb
```diff
@@ -186,10 +186,11 @@ def main
 
     if Puppet.features.root?
       begin
-        Puppet::Util.chuser
+        if Etc.getpwnam(Puppet[:user])
```
whopper: First approach: use `getpwnam` which throws an ArgumentError if the user doesn't exist, and emit a warning if it does.

whopper:
I'd like to run this through acceptance manually, but I'm not sure how to do a puppet-agent run using a local vanagon build (or just using this branch). Any advice?

lib/puppet/application/master.rb
joshcooper: You could test for the user as:

```
irb(main):004:0> Puppet::Type.type(:user).new(:name => Puppet[:user]).exists?
=> true
```

whopper: @joshcooper magic! I wish I'd found this earlier! :)

lib/puppet/application/master.rb
```diff
@@ -186,10 +186,11 @@ def main
...
+          Puppet::Util.chuser
+        end
       rescue => detail
-        Puppet.log_exception(detail, "Could not change user to #{Puppet[:user]}: #{detail}")
-        exit(39)
+        Puppet.warning "Could not change user to #{Puppet[:user]}: #{detail}. Continuing as root."
```
joshcooper: Since we only `chuser` when the puppet user exists, I think we want to keep the fail behavior.

whopper: @joshcooper right, that was my mistake. If we use `Puppet::Type.type(:user)...` like you mentioned above to check, we probably won't have to deal with an extra level of exceptions, so I'll try that.

whopper:
@joshcooper updated - feels cleaner now! Thanks for the suggestions.

peterhuene:
Are we really going to make it easier for users to run a webrick master as root? :/  Just seems like we're making it easier for inadvertent master invocations from a puppet-agent install to be insecure, rather than erroring and informing users that they need to create a puppet user to run a master?

joshcooper:
@peterhuene good points, I was on the fence about it, but yeah you're right. We should preserve the existing of running webrick as non-root, and if the accounts don't exist, then fail securely rather than carry on.

whopper:
@peterhuene @joshcooper sounds good! Do we want to clean up the error message(s) for that (doing something similar to what's in this PR) or just leave it the way it is and close this PR? Currently it'll try to `chuser` and throw a couple of exceptions leading to three messages in stderr. If we check that the user exists beforehand here, we can clean that up.

joshcooper:
I'd check if the user exists, if not, log an error and exit. Otherwise do what we did before.

kylog:
Agree with that approach, and thanks @peterhuene for the sanity check.

I'd like to add a deprecation notice at the same time, since the plan to deprecate, and later drop, webrick is in part what has motivated the changes that lead to this issue. Objections?

The deprecation notice should show regardless of whether you're running as root or not.

whopper:
@kylog cool, I'll add that in as part of this PR and ticket the deprecation.

spec/unit/application/master_spec.rb
```diff
@@ -314,11 +314,38 @@
         @master.main
       end
 
-      it "should drop privileges if running as root" do
-        Puppet.features.stubs(:root?).returns true
+      context "user privileges" do
+        before do
+          @user = Puppet::Type.type(:user).new(:name => 'puppet')
+          Puppet::Type.type(:user).expects(:new).returns @user
```
joshcooper: I'm thinking we want to restrict this to only the cases where `.new` is called with `:name => 'puppet'`. We have some code that does something similar: https://github.com/puppetlabs/puppet/blob/master/spec/unit/settings_spec.rb#L1591-L1595. It also returns a mock `user` object, which I think is preferred when testing the master application.

spec/unit/application/master_spec.rb
```diff
@@ -314,11 +314,38 @@
...
+      end
+
+      it "should log a deprecation notice when running a WEBrick server" do
+        @user = Puppet::Type.type(:user).new(:name => 'puppet')
+        Puppet::Type.type(:user).expects(:new).returns @user
+        Puppet.features.stubs(:root?).returns true
+        @user.expects(:exists?).returns true
         Puppet::Util.expects(:chuser)
 
+        Puppet.expects(:deprecation_warning).with("The WEBrick Puppet Master is deprecated and will be removed in a future release.")
```
joshcooper: Based on https://docs.puppetlabs.com/puppet/latest/reference/services_master_webrick.html, I'm thinking this should be "The WEBrick Puppet master server".

joshcooper: Is this something we want to deprecate in code, or just in documentation? It's going to be awhile before Puppet 5 comes out, and I imagine this warning could become annoying.

whopper: @kylog I was wondering the same thing. If we do deprecate in code, would it be going out in 4.1.0?

whopper:
@joshcooper updated the specs to use a mock `user` object. Also corrected "WEBrick Puppet master server" message, if we don't end up making that a docs-only deprecation.

joshcooper:
@whopper sorry I didn't notice earlier, could you retarget for stable so we can release in 4.0.1?

Also, I'm a little hesitant to put in a deprecation warning that can't be turned off. What do you think about splitting this into two PRs? We can merge the one for PUP-4394 now, and discuss the other one when @kylog is back?

whopper:
@joshcooper no problem, I'll split this into 2 PRs and target both at stable

PR #3835, opened 2015-04-16
(PUP-4044) curl directly to host in  install_repos_on

er0ck:
Download directly to SUT in install_utils

    This commit modifies the helpers in
    lib/puppet/acceptance/install_utils.rb to download files directly
    onto the target test host rather than downloading them locally and
    scp'ing them to the host.

    This reduces the test execution time by removing the duplication of
    downloading the target files twice (once from the origin point to
    the test runner, and then again from the test runner to the SUT).

This change eliminates the curls of the rpm files prior to installing them by
giving the URL directly to rpm. This should save a small amount of time.

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -55,26 +55,8 @@ module Puppet
         return true
       end
 
-      def fetch(base_url, file_name, dst_dir)
-        FileUtils.makedirs(dst_dir)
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Is it to pedantic to ask that the commit messages start capitalized, and sentences are formatted (the double space and lower case start for the last sentence)?

er0ck:
Thanks @MikaelSmith.  I should have been more careful.

PR #3836, opened 2015-04-17
(PUP-4359) Add the puppetversion fact for cfacter

MikaelSmith:
cfacter - and in the future, Facter 3 - don't define the puppetversion
fact. This is done to avoid having the current cyclical dependency in
Facter.

Puppet already adds several facts to the node's facts. Add puppetversion
prior to loading custom facts, so the custom facts can access it.

### DIFF
--- a/lib/puppet/indirector/facts/facter.rb
+++ b/lib/puppet/indirector/facts/facter.rb
@@ -21,6 +21,13 @@ class Puppet::Node::Facts::Facter < Puppet::Indirector::Code
   # Lookup a host's facts up in Facter.
   def find(request)
     Facter.reset
+
+    # Add the puppetversion fact; this is done before generating the hash so it is
+    # accessible to custom facts.
+    Facter.add(:puppetversion) do
+      setcode { Puppet.version.to_s }
+    end
+
     self.class.setup_external_search_paths(request) if Puppet.features.external_facts?
     self.class.setup_search_paths(request)
 
--- a/spec/integration/indirector/facts/facter_spec.rb
+++ b/spec/integration/indirector/facts/facter_spec.rb
@@ -19,4 +19,12 @@ describe Puppet::Node::Facts::Facter do
                              Puppet::Node.indirection.find('foo'))
     expect(cat.resource("Notify[AaBbCc]")).to be
   end
+
+  it "adds the puppetversion fact" do
+    Facter.stubs(:reset)
+
+    cat = compile_to_catalog('notify { $::puppetversion: }',
+                             Puppet::Node.indirection.find('foo'))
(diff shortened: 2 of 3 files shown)


lib/puppet/indirector/facts/facter.rb
```diff
@@ -21,6 +21,13 @@ def save(facts)
   # Lookup a host's facts up in Facter.
   def find(request)
     Facter.reset
+
+    # Add the puppetversion fact; this is done before generating the hash so it is
+    # accessible to custom facts.
+    Facter.add(:puppetversion) do
+      setcode { Puppet.version.to_s }
+    end
+
```
joshcooper: We're capturing `puppetversion` here, but `Puppet::Node::Facts#add_local_facts`, called below, also captures the [`clientversion`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/node/facts.rb#L31) fact (which has the same value as `puppetversion`).

```
$ be puppet apply -e 'notice($clientversion)'
Notice: Scope(Class[main]): 4.0.0
```

It would be good to consolidate the two code paths. On one hand, we could just add `values[:puppetversion] = Puppet.version.to_s` to the `add_local_facts` and be done.

On the other hand, if we want these facts to be visible to custom facts, then we have to set them before `Facter.to_hash` is called. That would take more work to unwind, e.g. the network_device also has logic for constructing facts objects: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/facts/network_device.rb#L13-L17

Also we'll want to update the docs about Puppet Agent Facts: https://docs.puppetlabs.com/puppet/latest/reference/lang_facts_and_builtin_vars.html#puppet-agent-facts

MikaelSmith: The need for `puppetversion` comes from a custom fact in `stdlib`, so we definitely need to do it before `Facter.to_hash`. I chose this implementation route to preserve the previous behavior, where the local facts don't show up for custom facts, and the `puppetversion` fact doesn't show up for network devices.

Agreed that we should update the docs somehow; I assumed `puppetversion` was already in docs somewhere and usage hasn't really changed.

MikaelSmith: The only route that seems to make sense is to separate what fact/value pairs we want to add from how they're added, then do it before to_hash for factor facts and after we create the hash for networking facts.

kylog: Re the facts in `add_local_facts` let's treat those as a separate concern. I created https://tickets.puppetlabs.com/browse/PUP-4457 for that, and we can revisit at a later date.

Meanwhile this looks good.

PR #3837, opened 2015-04-17
(PUP-4404) Fix SID lookup when running as SYSTEM

MikaelSmith:
When running as the SYSTEM account, doing lookup of the current user via
the WIN32OLE interface fails, because SYSTEM isn't recognized as a user.

We'd like to test looking up an Administrator's SID, so query WIN32OLE for
the members of the Administrators group and get the SID of the first
member.

### DIFF
--- a/spec/unit/util/windows/sid_spec.rb
+++ b/spec/unit/util/windows/sid_spec.rb
@@ -13,18 +13,6 @@ describe "Puppet::Util::Windows::SID", :if => Puppet.features.microsoft_windows?
   let(:unknown_name) { 'chewbacca' }
 
   context "#octet_string_to_sid_object" do
-    it "should properly convert an array of bytes for the local Administrator SID" do
-      host = '.'
-      # Bringing the ADSI module in isn't ideal, but we're leaving re-arranging the modules
-      # for a future breaking release.
-      username = Puppet::Util::Windows::ADSI::User.current_user_name
-      admin = WIN32OLE.connect("WinNT://#{host}/#{username},user")
-      converted = subject.octet_string_to_sid_object(admin.objectSID)
-
-      expect(converted).to eq(Win32::Security::SID.new(username, host))
-      expect(converted).to be_an_instance_of Win32::Security::SID
-    end
-
     it "should properly convert an array of bytes for a well-known SID" do
       bytes = [1, 1, 0, 0, 0, 0, 0, 5, 18, 0, 0, 0]
       converted = subject.octet_string_to_sid_object(bytes)


MikaelSmith:
Should fix the failure at https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Stable/view/Puppet/job/platform_aio-puppet_unit-ruby_stable-windows/46/.

So far I've manually tested on Windows 2008r2-x64 running as the SYSTEM account and as a member of the Administrators group (account named `vagrant`).

MikaelSmith:
I'm also open to special-casing when running as SYSTEM, but I think this should cover the use case.

joshcooper:
Sweet https://ci.appveyor.com/project/puppetlabs/puppet/build/4.0.0.15

spec/unit/util/windows/sid_spec.rb
```diff
@@ -15,10 +15,15 @@
...
-      admin = WIN32OLE.connect("WinNT://#{host}/#{username},user")
+      # Get the first user in the Administrators group
+      admins = WIN32OLE.connect("WinNT://#{host}/Administrators,group")
+      admin = nil
+      admins.members.each do |member|
+        admin = member
+        break
+      end
+      username = admin.name
+
```
joshcooper: It's sad that WIN32OLE objects behave like enumerable objects but don't implement Enumerable, so you can't just do `admins.members.first`. That said this test is just making sure `octet_string_to_sid_object` works, so how about if we just specify a byte array for an account we know exists, e.g.

```
C:\work\puppet [stable +9 ~0 -0 !]> bundle exec irb
=> true
irb(main):002:0> sid = Win32::Security::SID.new('SYSTEM').sid.unpack('C*')
=> [1, 1, 0, 0, 0, 0, 0, 5, 18, 0, 0, 0]
irb(main):003:0> Puppet::Util::Windows::SID.octet_string_to_sid_object(sid)
=> #<Win32::Security::SID:0x39750f0 @sid="\x01\x01\x00\x00\x00\x00\x00\x05\x12\x00\x00\x00", @account="SYSTEM", @host="WIN-QP47VOHA2P4", @domain="NT AUTHORITY", @account_type="well known group">
```

So we could simplify the test to:

```
sid =  [1, 1, 0, 0, 0, 0, 0, 5, 18, 0, 0, 0]
converted = subject.octet_string_to_sid_object(sid)
expect(converted).to eq(Win32::Security::SID.new('SYSTEM', host))
```

ferventcoder: The only issue that may be here is when the first person in the administrators group is the SYSTEM account. :/

ferventcoder: It could cause intermittent issues that may be hard to detect. I would prefer a known user.

MikaelSmith: Sounds good. I didn't see the SYSTEM account in the Administrators group when I queried it (it only had one member). The problem in Jenkins was that querying WinNT://./SYSTEM,user failed, so once we had the OLE object we can generate an SID from it. But I like simpler too.

MikaelSmith: That's actually exactly what the next test does.

joshcooper: ha! yeah, let's drop the test.

PR #3838, opened 2015-04-17
(DOC-1887) Fix indirection reference comments for Puppet 4 url changes

rlinehan:
These comments get generated and posted onto the docs site. This commit
updates them for the Puppet 4 url changes.


kylog:
Oh right! Good catch.

PR #3840, opened 2015-04-17
(PUP-4178) Move defined function to 4.x to correct behavior

hlindberg:
Before this, the 3x defined() function beahaves in a strange way when
giving undef and empty string as input. A user may want to check if a
variable is defined and calls:
```
defined($not_defined)
```
and becomes surprised when true is returned even if the variable is not
defined. The user should have asked for:
```
defined('$not_defined')
```
This anomaly is caused by the 3.x functions inability to distinquish
between undef and an empty string, and the strange naming of the
Class[main] which is called '' (the empty string) internally, and 'main'
externally.

This commit fixes this problem by adding a 4.x implementation of the
defined() function. This version:
* does not accept undef input (protects users from the mistake of
  defined($foo)).
* the empty string produces false
* the string 'main' produces true (note that there is always a main
class)

The documentation for the 3x defined function is also updated to reflect
this (to avoid confusion).

This also fixes the ticket PUP-4331 since what it suggest (giving an
error when given undef) is implmented by typing the arguments to the
function.

### DIFF
--- a/lib/puppet/functions/defined.rb
+++ b/lib/puppet/functions/defined.rb
@@ -70,8 +70,7 @@ Puppet::Functions.create_function(:'defined', Puppet::Functions::InternalFunctio
 
   dispatch :is_defined do
     scope_param
-    param          ARG_TYPE, 'first_arg'
-    repeated_param ARG_TYPE, 'additional_args'
+    optional_repeated_param ARG_TYPE, 'additional_args'
   end
 
   def is_defined(scope, *vals)
--- a/lib/puppet/functions/defined.rb
+++ b/lib/puppet/functions/defined.rb
@@ -65,15 +65,14 @@ Puppet::Functions.create_function(:'defined', Puppet::Functions::InternalFunctio
         if m = /^\$(.+)$/.match(val)
           scope.exist?(m[1])
         else
-          val = case val
+          case val
           when ''
             next nil
           when 'main'
-            ''
+            scope.compiler.findresource(:class, '') # scope.find_hostclass('')
           else
-            val
+            scope.find_resource_type(val) || scope.find_definition(val) || scope.compiler.findresource(:class, val)
           end
-          scope.find_resource_type(val) || scope.find_definition(val) || scope.find_hostclass(val)
         end
       when Puppet::Resource
         scope.compiler.findresource(val.type, val.title)
--- a/lib/puppet/functions/defined.rb
+++ b/lib/puppet/functions/defined.rb
@@ -16,7 +16,8 @@
 #     defined('$name')
 #
 # Resource declarations are checked using resource references, e.g.
-# `defined( File['/tmp/myfile'] )`. Checking whether a given resource
+# `defined( File['/tmp/myfile'] )`, or `defined( Class[myclass] )`.
+# Checking whether a given resource
 # has been declared is, unfortunately, dependent on the evaluation order of
 # the configuration, and the following code will not work:
 #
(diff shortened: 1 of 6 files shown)


hlindberg:
Hm problems with Ruby 1.8.7. And this problem actually unveiled a different issue... brb

thallgren:
Suggest deferring this until [PUP-4438](https://tickets.puppetlabs.com/browse/PUP-4438) has been fixed.

hlindberg:
This PR was rebased and the new functionality from pup 4438 was used. A problem was found with weaving and repeated parameter. I added a fix for that (e9787bf) and could then use the new feature.

lib/puppet/functions/defined.rb
```diff
@@ -0,0 +1,130 @@
...
+# affect the behavior of `defined`.
+#
+# You may also search using types:
+#
+#     defined(Resource['file','/some/file'])
+#     defined(File['/some/file'])
+#     defined(Class['foo'])
+#
+# The `defined` function does not answer if data types (e.g. `Integer`) are defined.
+#
```
thallgren: What does it answer on datatypes then? false or true?

lib/puppet/functions/defined.rb
```diff
@@ -0,0 +1,130 @@
...
+# @since 3.8.1 type specific requests with future parser
+#
+Puppet::Functions.create_function(:'defined', Puppet::Functions::InternalFunction) do
+
+  ARG_TYPE = 'Variant[String,Type[CatalogEntry], Type[Type[CatalogEntry]]]'
+
+  dispatch :is_defined do
+    scope_param
+    optional_repeated_param ARG_TYPE, 'additional_args'
+  end
```
thallgren: Why not `required_repeated_param ARG_TYPE args`?

lib/puppet/functions/defined.rb
```diff
@@ -0,0 +1,130 @@
...
+
+  def is_defined(scope, *vals)
+    vals.any? do |val|
+      case val
+      when String
+        if m = /^\$(.+)$/.match(val)
+          scope.exist?(m[1])
+        else
```
thallgren: It's cleaner to write:

    if val =~ /^\$(.+)$/
      scope.exists?($1)

and it avoids use of assignment operator in conditional

lib/puppet/functions/defined.rb
```diff
@@ -0,0 +1,130 @@
...
+            scope.find_resource_type(val) || scope.find_definition(val) || scope.find_hostclass(val)
+            #scope.compiler.findresource(:class, val)
+          end
+        end
+      when Puppet::Resource
+        # Find instance of given resource type and title that is in the catalog
+        scope.compiler.findresource(val.type, val.title)
+
+      when Puppet::Pops::Types::PResourceType
+        raise ArgumentError, "The given resource type is a reference to all kind of types" if val.type_name.nil?
```
thallgren: Use single quoted string

lib/puppet/functions/defined.rb
```diff
@@ -0,0 +1,130 @@
...
+        if val.title.nil?
+          scope.find_builtin_resource_type(val.type_name) || scope.find_definition(val.type_name)
+        else
+          scope.compiler.findresource(val.type_name, val.title)
+        end
+
+      when Puppet::Pops::Types::PHostClassType
+        raise  ArgumentError, "The given class type is a reference to all classes" if val.class_name.nil?
```
thallgren: single quoted string

lib/puppet/functions/defined.rb
```diff
@@ -0,0 +1,130 @@
...
+        when Puppet::Pops::Types::PResourceType
+          # It is most reasonable to take Type[File] and Type[File[foo]] to mean the same as if not wrapped in a Type
+          # Since the difference between File and File[foo] already captures the distinction of type vs instance.
+          is_defined(scope, val.type)
+
+        when Puppet::Pops::Types::PHostClassType
+          # Interpreted as asking if a class (and nothing else) is defined without having to be included in the catalog
+          # (this is the same as asking for just the class' name, but with the added certainty that it cannot be a defined type.
+          #
+          raise  ArgumentError, "The given class type is a reference to all classes" if val.type.class_name.nil?
```
thallgren: single quoted string

lib/puppet/parser/functions/defined.rb
```diff
@@ -38,8 +38,30 @@
       defined(File[\'/some/file\'])
       defined(Class[\'foo\'])
 
+  When used with the future parser (4.x), the `defined` function does not answer if data
+  types (e.g. `Integer`) are defined, and the rules for asking for undef, empty strings, and
+  the main class are different:
```
thallgren: "does not answer" doesn't really say what the answer is and there's always an answer unless it raises an exception.

spec/unit/functions/defined_spec.rb
```diff
@@ -0,0 +1,291 @@
...
+
+  def newresource(type, title)
+    resource = Puppet::Resource.new(type, title)
+    @compiler.add_resource(@scope, resource)
+    resource
+  end
+
+  #--- CLASS
+  #
+  context "can determine if a class" do
```
thallgren: Please use single quoted strings everywhere unless there's a need for double quotes.

lib/puppet/functions/defined.rb
hlindberg: That was exactly why this PR was updated - only it did not work at first and my experiment if it was required or optional that caused it made me forget to change it back. Good catch.

lib/puppet/functions/defined.rb
hlindberg: it also avoids one method call (match), the operators are faster than method calls.

hlindberg:
@thallgren review-comments addressed, rebased on top of change to check that required does not follow optional

lib/puppet/functions/defined.rb
```diff
@@ -0,0 +1,132 @@
...
+      when String
+        if val =~ /^\$(.+)$/
+          scope.exist?($1)
+        else
+          case val
+          when ''
+            next nil
+          when 'main'
+            # Find the main class (known as ''), it does not have to be in the catalog
+            scope.find_hostclass('') # scope.find_hostclass('')
```
thallgren: The comment on the same line seems somewhat redundant.

PR #3842, opened 2015-04-17
Maint/master/merge in stable

melissa:


joshcooper:
Everything looks good except I have no idea about build_defaults.yaml. I presume all of those deleted lines are not needed now that we have puppet-agent.

I'm going to wait for the appveyor run to validate windows specs, then merge

melissa:
@joshcooper all of those deleted lines come from https://github.com/melissa/puppet/commit/96ba5cfae137bb81699a57241ba1c768bb8badca

Since we're no longer using the packaging repo to build puppet, we no longer need all of those specifications. We're keeping that file around in case anyone from the community is using the packaging repo to build puppet packages on their own, but we no longer use that file

MikaelSmith:
Looks consistent with the investigation I did yesterday :+1:

PR #3843, opened 2015-04-17
(PUP-4233, PUP-4234, PUP-4235) Enable Puppet Passenger Acceptance

MikaelSmith:
* (PUP-4233) Add node configs for puppet-passenger
* (PUP-4234) Modify AIO pre-suite to use passenger

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -281,6 +281,14 @@ EOS
     end
 
     desc <<-EOS
+Run the acceptance tests through Beaker and install packages as part of the AIO puppet-agent installation, testing against puppet-master-passenger.
+#{USAGE}
+EOS
+    task :passenger => 'ci:check_env' do
+      beaker_test(:passenger)
+    end
+
+    desc <<-EOS
 Run the acceptance tests through Beaker and install from git on the configuration targets.
 #{USAGE}
 EOS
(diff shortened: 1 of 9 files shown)


MikaelSmith:
That makes sense, and makes the description in PUP-4233 make sense too.

MikaelSmith:
This works, but I'm still cleaning it up and testing all the new configs.

acceptance/config/nodes/debian-7-passenger.yaml
```diff
@@ -0,0 +1,20 @@
+---
+HOSTS:
+  master:
+    roles:
+    - master
+    platform: debian-wheezy-x86_64
+    hypervisor: vcloud
+    template: debian-7-x86_64
+    passenger: true
```
MikaelSmith: Trying to override is_puppetserver here doesn't appear to work, so I'm going to make these generic server configs and do all the passenger vs puppetserver selection in options files.

MikaelSmith:
Ran the configs, they all seem to work and get puppet-master-passenger packages.

justinstoller:
I think that `server` is an ambiguous term for the host files. Not saying that I love `ubuntu-1404-passenger`, but I don't think I would understand why these configs are special from their name if I wasn't involved in these tickets.

MikaelSmith:
Well, there's nothing specific to passenger in those configs. I named them to differentiate varying the server OS instead of the client.

joshcooper:
Justin and I were working on updates to this test so we don't have to create s new rake target

MikaelSmith:
Ah, ok. I played around with it, but felt a new options file was the simplest.

joshcooper:
Superceded by PR #3859

PR #3845, opened 2015-04-17
(PUP-4044) Restore curl directly to host in install_repos_on

MikaelSmith:
Prior commit was reverted because it didn't work on Debian and Ubuntu;
it downloaded the repo_configs dir to the wrong location on the host.

Restore those changes with a fix to download to the correct location.

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -55,26 +55,8 @@ module Puppet
         return true
       end
 
-      def fetch(base_url, file_name, dst_dir)
-        FileUtils.makedirs(dst_dir)
-        src = "#{base_url}/#{file_name}"
-        dst = File.join(dst_dir, file_name)
-        if File.exists?(dst)
-          logger.notify "Already fetched #{dst}"
-        else
-          logger.notify "Fetching: #{src}"
-          logger.notify "  and saving to #{dst}"
-          open(src) do |remote|
-            File.open(dst, "w") do |file|
-              FileUtils.copy_stream(remote, file)
-            end
-          end
-        end
-        return dst
-      end
-
-      def fetch_remote_dir(url, dst_dir)
-        logger.notify "fetch_remote_dir (url: #{url}, dst_dir #{dst_dir})"
+      def fetch_remote_dir(host, url, dst_dir)
+        logger.notify "fetch_remote_dir (host #{host}, url #{url}, dst_dir #{dst_dir})"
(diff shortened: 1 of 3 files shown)


acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -118,7 +101,6 @@ def stop_firewall_on(host)
 
       def install_repos_on(host, project, sha, repo_configs_dir)
```
MikaelSmith: `repo_configs_dir` is no longer used, and could be removed (and all the call sites updated)

MikaelSmith:
This looks like a trade-off between performing well for small numbers of machines over VPN vs performing well for lots of the same platforms over a local network.

acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -212,27 +184,18 @@ def install_repos_on(host, project, sha, repo_configs_dir)
...
+
             repo_loc = "/root/#{project}"
 
             on host, "rm -rf #{repo_loc}"
             on host, "mkdir -p #{repo_loc}"
 
-            scp_to host, deb, repo_loc
-            scp_to host, list, repo_loc
-            scp_to host, repo_dir, repo_loc
+            repo_dir = fetch_remote_dir(host, "http://%s/%s/%s/repos/apt/%s" % [tld, project, sha, version], repo_loc)
```
MikaelSmith: This line was fixed from the last commits; instead of downloading to `platform_configs_dir` (which resolved to `~/#{platform_configs_dir}` on the host) we just download the repo directly to `~/#{project}`.

MikaelSmith:
Tested Ubuntu 14.04 and RHEL 7.

acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -91,7 +74,7 @@ def fetch_remote_dir(url, dst_dir)
         logger.notify "  using command: #{wget_command}"
 
         #in ruby 1.9+ we can upgrade this to popen3 to gain access to the subprocess pid
-        result = `#{wget_command} 2>&1`
+        result = on(host, "#{wget_command} 2>&1").stdout
```
joshcooper: Can this be shortened to just `on(host, wget_command).stdout`?

joshcooper:
@MikaelSmith I have some ideas for how to shorten this code quite a bit. I'll submit a PR later today, once I verify on additional systems.

MikaelSmith:
Awesome, thanks.

PR #3846, opened 2015-04-18
(PUP-3988) Remove unused hostname checking feature

hlindberg:
Before this, the validator had the ability to pass a single-feature
name when checking hostnames. This was used to check the parent of a
node. This is no longer needed since having a parent is now illegal.

This removes this feature. Tests that checked validation of parent
host name matches were removed a long time ago.

### DIFF
--- a/lib/puppet/pops/validation/checker4_0.rb
+++ b/lib/puppet/pops/validation/checker4_0.rb
@@ -52,8 +52,8 @@ class Puppet::Pops::Validation::Checker4_0
 
   # Performs check if this is a vaid hostname expression
   # @param single_feature_name [String, nil] the name of a single valued hostname feature of the value's container. e.g. 'parent'
-  def hostname(o, semantic, single_feature_name = nil)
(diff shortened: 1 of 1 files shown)


lib/puppet/pops/validation/checker4_0.rb
```diff
@@ -537,31 +533,31 @@ def hostname_ConcatenatedString(o, semantic, single_feature_name)
...
   end
 
-  def hostname_LiteralRegularExpression(o, semantic, single_feature_name)
+  def hostname_LiteralRegularExpression(o, semantic)
     # always ok
   end
 
-  def hostname_Object(o, semantic, single_feature_name)
+  def hostname_Object(o, semantic)
     acceptor.accept(Issues::ILLEGAL_EXPRESSION, o, {:feature=> single_feature_name || 'hostname', :container=>semantic})
```
thallgren: I guess a change to `:feature => 'hostname` here is in order?

hlindberg:
@thallgren update (removing dead branch of logic) and now using "hostname" at all times pushed to this PR.

PR #3847, opened 2015-04-18
fix service_name for not include description

takumin:
FreeBSD service provider for not include description.

Ex.
  dns/bind99 package rc script

```
$ /usr/local/etc/rc.d/named rcvar
# named : named BIND startup script
#
named_enable="NO"
#   (default: "")

```

before fix:
 create file "/etc/rc.conf.d/named : named BIND startup script"
after fix:
 create file "/etc/rc.conf.d/named"

thx!


takumin:
sry, again post JIRA.

PR #3848, opened 2015-04-20
(PUP-4438) Add required_repeated_param to 4.x function API

thallgren:
Adds the ability to declare a repeated parameter as being required
using the method `required_repeated_parameter` with the meaning that
at least one argument must be passed to it.

For consistency, this commit also adds an `optional_repeated_parameter`
alias for the already existing `repeated_parameter` method.


hlindberg:
Please retarget this to the 3.x. branch.

PR #3854, opened 2015-04-20
(PUP-3814) Only log compilation errors once

hlindberg:
Before this all compilation errors were logged both by the compiler, and
then again by the apply application.

This splits up the error handling so that those errors that are logged
and processed by the compiler are caught and then causing an immediate
exit. Other errors (such as errors after the compiler has finished
compiling, or before being in a state where errors are rescued and
logged) are caught by the general logic at the end of the compile/apply
block.

This fix is not a guarantee there will never be duplicate output since
different compilation logic can be used (and it may not behave the same
way).

### DIFF
--- a/lib/puppet/application/apply.rb
+++ b/lib/puppet/application/apply.rb
@@ -212,23 +212,35 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
       end
 
       begin
-        # Compile our catalog
+        # Compile the catalog
         starttime = Time.now
-        catalog = Puppet::Resource::Catalog.indirection.find(node.name, :use_node => node)
 
-        # Translate it to a RAL catalog
-        catalog = catalog.to_ral
+        # When compiling, the compiler traps and logs certain errors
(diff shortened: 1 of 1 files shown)


lib/puppet/application/apply.rb
```diff
@@ -212,23 +212,35 @@ def main
...
-          catalog.write_class_file
-          catalog.write_resource_file
+          catalog.retrieval_duration = Time.now - starttime
+
+          if options[:write_catalog_summary]
+            catalog.write_class_file
+            catalog.write_resource_file
+          end
+          catalog
+        rescue Puppet::ParseErrorWithIssue, Puppet::Error
```
thallgren: There's no need to explicitly declare `Puppet::ParserErrorWithIssue` in the rescue since it inherits from `Puppet::ParseError` which in turn inherits from `Puppet::Error`

hlindberg: understood, but I wanted to be explicit, those two exact types are raised by the compiler; and if the hierarchy ever changes it would do it wrong. Suggest keeping the explicit catch of both.

PR #3857, opened 2015-04-21
(PUP-4044) Curl repo config directly to SUT

joshcooper:
Previously, we copied the puppetlabs-release package from
{yum,apt}.puppetlabs.com, the per-commit repo_config from
builds.puppetlabs.lan or nightlies.puppetlabs.com, and the actual repos
from either PC1, main or devel to the coordinator node, which then
scp'ed the files to each SUT.

This commit modifies the `install_repos_on` method to curl the
repo_config for the specified SHA to the SUT's system location e.g.
/etc/yum.repos.d. Once installed the repo_config specifies the URL from
which to install the package for the specified SHA, e.g

    baseurl=http://builds.puppetlabs.lan/puppet-agent/<SHA>/repos/el/7/PC1/x86_64/

This commit also skips installing the puppetlabs-release package.
In puppet 3.x, the package is required, because it provides dependencies
for puppet, e.g. facter, hiera. In the AIO world, this is no longer
necessary.

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -124,127 +124,39 @@ module Puppet
         sha     = sha == 'nightly' ? nil                        :  sha
 
         case platform
-          when /^(fedora|el|centos)-(\d+)-(.+)$/
-            variant = (($1 == 'centos') ? 'el' : $1)
-            fedora_prefix = ((variant == 'fedora') ? 'f' : '')
-            version = $2
-            arch = $3
+        when /^(fedora|el|centos)-(\d+)-(.+)$/
+          variant = (($1 == 'centos') ? 'el' : $1)
+          fedora_prefix = ((variant == 'fedora') ? 'f' : '')
+          version = $2
+          arch = $3
 
-            rpm = fetch(
-              "http://yum.puppetlabs.com",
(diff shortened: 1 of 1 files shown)


joshcooper:
Supercedes PR #3845

acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -124,127 +124,39 @@ def install_repos_on(host, project, sha, repo_configs_dir)
...
+
+          list_filename = "pl-%s%s-%s.list" % [
+            project,
+            sha ? '-' + sha : '',
+            version
+          ]
+          list_url = "http://%s/%s/%s/repo_configs/deb/%s" % [tld, project, sha, list_filename]
+
+          on host, "curl -o /etc/apt/sources.list.d/#{list_filename} #{list_url}"
+          on host, "apt-get update"
```
joshcooper: @justinstoller mentioned that we're not calling `yum makecache`, and likely don't need to since we're installing a new repo config, but something to consider.

joshcooper:
Tested with redhat-7-x86_64, debian-7-x86_64, ubuntu-1404-x86_64, fedora-21-x86_64.

joshcooper:
I've talked this over with @er0ck and @johnduarte, and paired with @justinstoller on this, merging now because I can't stand the 5 minute delay running acceptance tests over the VPN.

PR #3860, opened 2015-04-22
(PUP-4388) Remove START option from /etc/default/puppet on Debian

whopper:
In the new puppet-agent AIO package, services no longer
start on package installation, as there are a few required
configuration steps which must be addressed before the
service can be started.

This means we can remove the agent service START option from
/etc/default/puppet, a relic of old packages where debhelper
started services upon package installation.

### DIFF
--- a/ext/debian/puppet.default
+++ b/ext/debian/puppet.default
@@ -1,9 +1,4 @@
 # Defaults for puppet - sourced by /etc/init.d/puppet
 
-# Enable puppet agent service?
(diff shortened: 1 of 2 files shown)


whopper:
This is the retargeted equivalent of https://github.com/puppetlabs/puppet/pull/3822

whopper:
Doh, nevermind, the ticket was just updated to target 4.1.0! Closing and reopening the original....

PR #3861, opened 2015-04-22
(maint) install_puppetlabs_dependency_repo_in_install_repos_on

er0ck:
This change adds the .rpm and .deb repo configurations so installs can
resolve puppetlabs dependencies, when required.  This is required by
package install method, and in the deployer acceptance setup when
installing from packages (this change is mirrored there).

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -10,6 +10,7 @@ module Puppet
         :redhat        => /fedora|el|centos/,
         :debian        => /debian|ubuntu/,
         :debian_ruby18 => /debian|ubuntu-lucid|ubuntu-precise/,
+        :sles          => /sles/,
         :solaris_10    => /solaris-10/,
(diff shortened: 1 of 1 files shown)


er0ck:
FYI, @joshcooper.

joshcooper:
I'm wondering why we care about `packages` style acceptance tests? Moving forward, I would expect consumers of puppet packages to be using `aio` always, and I'd like to deprecate/remove the `ci:test:packages` rake task and supporting code.

If we need to support `packages` for the time being, then I would be in favor of limiting the puppetlabs-release repo when `@options[:type] == packages`, because installing puppetlabs-release repo during AIO testing can introduce/mask problems where a dependency is resolved outside of the AIO package: https://tickets.puppetlabs.com/browse/PUP-4392.

er0ck:
agreed all around.  i use package install method because i'm lazy and i have helpers that find the latest sha for me, which i never fixed for puppet_agent.  i'll kill this PR and i'll restrict on :type for deployer setup.

PR #3865, opened 2015-04-24
(PUP-3814) Fix problem with apply swallowing post compilation errors

hlindberg:
Before this, errors during the post processing of a compiled catalog
(to_ral, etc) were swallowed by the apply command.

This fixes the problem by not including the post processing step in the
block that prevented errors from being logged twice (since the compiler
does its own logging).


joshcooper:
Merged in #3866

PR #3867, opened 2015-04-25
(PUP-4462) Fix problem with heredoc interpolation turning on escapes

hlindberg:
Before this, heredoc with interpolation would also turn on escapes. This
means input like \$name would not interpolate, and \\ would be required
to get a single \. It should be required to be explici about allowing
escapes by using /$ (to be allowed to escape interpolation).

The prolem was caused by the pattern used to find the start of an
interpolation. This pattern included the escapes that would make the
pattern skip the interpolation start - it would simply scan past the
interpolation start if escaped.

The solution was to create a new pattern for finding interpolation
starts when a $ is not one of the wanted escapes.

### DIFF
--- a/lib/puppet/pops/parser/lexer2.rb
+++ b/lib/puppet/pops/parser/lexer2.rb
@@ -213,7 +213,7 @@ class Puppet::Pops::Parser::Lexer2
     @scanner = StringScanner.new(string)
     @locator = locator || Puppet::Pops::Parser::Locator.locator(string, '')
     @lexing_context[:escapes] = escapes || UQ_ESCAPES
-    @lexing_context[:uq_slurp_pattern] = (interpolate || !escapes.empty?) ? SLURP_UQ_PATTERN : SLURP_ALL_PATTERN
+    @lexing_context[:uq_slurp_pattern] = interpolate ? (!escapes.include?('$') ? SLURP_UQNE_PATTERN : SLURP_UQ_PATTERN) : SLURP_ALL_PATTERN
   end
 
   # Convenience method, and for compatibility with older lexer. Use the lex_file instead.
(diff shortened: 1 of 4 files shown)


lib/puppet/pops/parser/slurp_support.rb
```diff
@@ -12,6 +12,8 @@ module Puppet::Pops::Parser::SlurpSupport
   SLURP_SQ_PATTERN  = /(?:[^\\]|^|[^\\])(?:[\\]{2})*[']/
   SLURP_DQ_PATTERN  = /(?:[^\\]|^|[^\\])(?:[\\]{2})*(["]|[$]\{?)/
   SLURP_UQ_PATTERN  = /(?:[^\\]|^|[^\\])(?:[\\]{2})*([$]\{?|\z)/
+  # unquoted, no escapes
+  SLURP_UQNE_PATTERN  = /([$]\{?|\z)/m
```
thallgren: It's more consistent to use `\$\{` instead of `[$]\{`.

lib/puppet/pops/parser/lexer2.rb
```diff
@@ -213,7 +213,7 @@ def lex_unquoted_string(string, locator, escapes, interpolate)
     @scanner = StringScanner.new(string)
     @locator = locator || Puppet::Pops::Parser::Locator.locator(string, '')
     @lexing_context[:escapes] = escapes || UQ_ESCAPES
-    @lexing_context[:uq_slurp_pattern] = (interpolate || !escapes.empty?) ? SLURP_UQ_PATTERN : SLURP_ALL_PATTERN
+    @lexing_context[:uq_slurp_pattern] = interpolate ? (!escapes.include?('$') ? SLURP_UQNE_PATTERN : SLURP_UQ_PATTERN) : SLURP_ALL_PATTERN
```
thallgren: Inner condition can be simplified to `(escapes.include?('$') ? SLURP_UQ_PATTERN : SLURP_UQNE_PATTERN)`

PR #3871, opened 2015-04-27
(PUP-4413) Add acceptance test that 3x forbidden env names work in 4x

er0ck:
This change adds an acceptance test to ensure that the 3x forbidden
environment names now work, and continue to work in 4x.
There are 4 previously forbidden names, which means this test has to run
puppet 5 times, unfortunately.


kurtwall:
Shipzit.

PR #3872, opened 2015-04-28
(PUP-1295) Erase only installed pkgs when using yum purge

scenek:


scenek:
What do you think? Is it ok even all providers doesn't support purgeable?

MikaelSmith:
Thank you for attempting to solve this ticket. However, this type of fix is one that would require at least some spec tests demonstrating that it fixes the problem, and likely acceptance tests (https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/resource/package/common_package_name_in_different_providers.rb would provide a foundation for testing rpm package install).

Also, please review our commit message [guidelines](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md).

Last, this code change is a no-op. The `||` operator, when used in this context, will only evaluate the right-hand side if the left-hand side is `nil`, but since it's already set to `{:ensure => :absent}` you will never evaluate `{:ensure => purged}`; `||` does not merge hashes.

peterhuene:
Hi @scenek.  Even if the above comments made my Michael didn't hold true, this is probably not the correct fix for this issue.  I believe the correct fix is to treat `ensure => purged` as in sync when the current status is `absent`.

Thus, something like this:

```ruby
def insync?(is)
...
when :absent
  return true if is == :absent or is == :purged
when :purged
  return true if is == :purged
...        
```

should become:

```ruby
def insync?(is)
...
when :absent, :purged
  return true if is == :absent || is == :purged
...        
```

The problem is that the yum provider supports purging, but never returns "purged" resources, only absent; thus it is always out of sync and the package is always purged even when not present.

peterhuene:
The above will break the `dpkg` provider since it can report on `:purged` vs. `:absent` packages.  Still researching the right fix.

peterhuene:
Given that `purged` is the "most uninstalled" state a package can be, I think the fix should probably be along the lines of what this PR attempted to do.  That is to say, the default hash should have `:ensure => :purged` if the query doesn't return a resource (which yum doesn't if the package isn't installed).  That way ensuring absent or purged will be a no-op due to the check in `insync?` above.

peterhuene:
Hi @scenek.  Thanks again for the contribution!  I've put up a PR that supersedes this one: https://github.com/puppetlabs/puppet/pull/3921 so I am closing this PR for now.  If you believe that the other PR does not address this issue, please reopen. Thanks!

PR #3873, opened 2015-04-28
(PUP-4465) Update description of --logdesc option

thallgren:
This commit updates the descripton of the --logdesc option for
the applications agent, apply, device, and master with information
about JSON structured logging output when the path ends with '.json'.


hlindberg:
It is logdest (not logdesc), but the typo is only in the commit message. Merged anyway.

PR #3874, opened 2015-04-28
Merge 3.x

hlindberg:


lib/puppet/application/agent.rb
```diff
@@ -231,6 +231,11 @@ def help
   file. If debugging or verbosity is enabled, this defaults to 'console'.
   Otherwise, it defaults to 'syslog' on POSIX systems and 'eventlog' on Windows.
 
+  A path ending with '.json' will receive structured output in JSON format. The
+  log file will not have an ending ']' automatically written to it due to the
+  appending nature of logging. It must be appended manually to make the content
+  valid JSON.
+
```
joshcooper: The log destination should override the close method and ensure the trailing `]` is added

PR #3876, opened 2015-04-28
(PUP-4373) Windows ADSI User groups property should behave similarly to Groups members property

ferventcoder:
This supersedes https://github.com/puppetlabs/puppet/pull/3788

Previously, when specifying a user's groups, the names were compared
as strings and needed to be an exact match. The group resource uses
SIDs when comparing the users, so the user resource should mimic that
behavior.

Without this change casing and names must match exactly and can 
produce erroneous results.

With this change, the comparison will be by SIDs so it will allow for
the following when a user is a member of Administrators and Users
groups:

```puppet
user { 'tim':
  ensure => present,
  groups => ['BUILTIN\Administrators','users'],
}
```

Groups members can also be set to an empty list and the user's groups will
report that it does but will not make the changes. Allow that the user groups to
be set to an empty list so that group's members and user's groups behavior
is the same. Without this change an existing user with groups modeled in the
following resource will continue to give the message that it is removing the 
groups without actually making any changes:

```puppet
user { 'tim':
  ensure => present,
  groups => [],
}
```

There is also shared behavior between `Puppet::Util::Windows::ADSI::User`
and `Puppet::Util::Windows::ADSI::Group` so it makes sense to start
combining shared functionality to a common class named
`Puppet::Util::Windows::ADSI::Shared`.

### DIFF
--- a/lib/puppet/provider/group/windows_adsi.rb
+++ b/lib/puppet/provider/group/windows_adsi.rb
@@ -30,7 +30,6 @@ Puppet::Type.type(:group).provide :windows_adsi do
     if @resource[:auth_membership]
       current_users == specified_users
     else
-      return true if specified_users.empty?
       (specified_users.keys.to_a & current_users.keys.to_a) == specified_users.keys.to_a
     end
   end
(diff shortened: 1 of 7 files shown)


ferventcoder:
@joshcooper this has been retargeted.

Iristyle:
Work for this continues in my branch since there were some things that required tweaking:

https://github.com/Iristyle/puppet/commits/ticket/master/PUP-4373-user-group-sids

Almost done....

PR #3877, opened 2015-04-29
(maint) skip upstart test if there are no ubuntu agents

joshcooper:
The handle_upstart acceptance test fails on ubuntu systems when using
passenger, because the master is ubuntu, but the agent is not. So the
confine would pass, but `agent` was nil, causing the test to fail later
with:

    undefined method `[]' for nil:NilClass

This commit skips the test if there are no ubuntu agents.


joshcooper:
```
$ be rake ci:test:passenger SHA=nightly CONFIG=config/nodes/debian-7-passenger.yaml TESTS=tests/resource/service/ticket_14297_handle_upstart.rb
...
Skipped Tests Cases:
  Test Case tests/resource/service/ticket_14297_handle_upstart.rb skip
```

PR #3879, opened 2015-04-29
(PUP-2630) Ensure it is possible to get trusted server facts that cannot be overwritten

hkenney:
Prior to this change, there was no way for module authors to ensure
that they were getting trusted server facts due to the fact that
server side global variables can be overwritten by client side
facts.

Add the $server_facts global variable which will contain a hash
of server side facts which cannot be overwritten. Add the
corresponding trusted_server_facts setting. The setting is false
by default, but if set to true by the user then the $server_facts
variable will be populated.

Additionally, if merge attempted to set a node parameter
that already existed, it would quietly fail to update the value.
This meant that if a client fact value was being used for one of
the server side facts, it would not be overwritten with the correct
value and the user would never know.

Since we cannot change the behavior without breaking API, simply
add a warning so that the user will be aware if merge failed to
change the value of a node parameter.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -518,6 +518,12 @@ module Puppet
       :desc    => "Stores trusted node data in a hash called $trusted.
         When true also prevents $trusted from being overridden in any scope.",
     },
+    :trusted_server_facts => {
+      :default => false,
+      :type    => :boolean,
+      :desc    => "Stores a trusted set of server-side global variables in a hash
+        called $server_facts, which cannot be overridden by client facts.",
+    },
     :immutable_node_data => {
       :default => '$trusted_node_data',
       :type    => :boolean,
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -522,7 +522,9 @@ module Puppet
       :default => false,
       :type    => :boolean,
       :desc    => "Stores a trusted set of server-side global variables in a hash
-        called $server_facts, which cannot be overridden by client facts.",
+        called $server_facts, which cannot be cannot be overridden by client_facts 
+        or logic in manifests. Makes it illegal to assign to the variable $server_facts 
+        in any scope.",
     },
     :immutable_node_data => {
       :default => '$trusted_node_data',
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -16,6 +16,8 @@ class Puppet::Node
   attr_accessor :name, :classes, :source, :ipaddress, :parameters, :trusted_data, :environment_name
   attr_reader :time, :facts
 
+  attr_reader :server_facts
+
   def self.from_data_hash(data)
     raise ArgumentError, "No name provided in serialized data" unless name = data['name']
 
@@ -86,6 +88,8 @@ class Puppet::Node
 
     @facts = options[:facts]
 
+    @server_facts = {}
+
     if env = options[:environment]
       self.environment = env
     end
@@ -109,7 +113,7 @@ class Puppet::Node
   def merge(params)
     params.each do |name, value|
       if @parameters.include?(name)
-        Puppet::Util::Warnings.warnonce("The node paramters #{name} for node #{@parameters["hostname"]} was already set to #{@parameters[name]}. It could not be set to #{value}")
+        Puppet::Util::Warnings.warnonce("The node paramter #{name} for node #{@parameters["hostname"]} was already set to #{@parameters[name]}. It could not be set to #{value}")
       else
         @parameters[name] = value
       end
@@ -118,17 +122,12 @@ class Puppet::Node
     @parameters["environment"] ||= self.environment.name.to_s
   end
 
-  def add_server_facts(server_facts)
-   # Append the current environment to the list of server facts
-    complete_server_facts = server_facts.merge({ "environment" => self.environment.name.to_s})
-
-   # Set the top scope variable $server_facts if :trusted_server_facts is true
-   if Puppet[:trusted_server_facts]
-     @topscope.set_trusted(complete_server_facts)
-   end
+  def add_server_facts(facts)
+    # Append the current environment to the list of server facts
+    @server_facts = facts.merge({ "environment" => self.environment.name.to_s})
 
     # Merge the server facts into the parameters for the node
-    merge(server_facts)
+    merge(facts)
   end
 
   # Calculate the list of names we might use for looking
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -114,6 +114,18 @@ class Puppet::Node
     @parameters["environment"] ||= self.environment.name.to_s
   end
 
+  def add_server_facts(server_facts)
+   # Complete server facts
+
+   # Set the top scope variable $server facts if :trusted_server_facts is true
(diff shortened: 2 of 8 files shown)


lib/puppet/node.rb
```diff
@@ -108,12 +108,28 @@ def fact_merge
...
+        @parameters[name] = value
+      end
     end
 
     @parameters["environment"] ||= self.environment.name.to_s
   end
 
+  def add_server_facts(server_facts)
+   # Complete server facts
+
```
hkenney: Shoot, this is where I needed to append the environment to the hash, I forgot to add that. I'll fix this.

kylog:
Seems like the first commit message is referencing the wrong Jira ticket (2639 vs 2630).

hkenney:
@kylog Thanks for catching all my typos! :+1: I'll fix all of those up.

foo.pp
```diff
@@ -0,0 +1 @@
+notify {"from global: ${::trusted_server_facts}":}
```
hlindberg: Debug / manual test file... should not be here

lib/puppet/defaults.rb
```diff
@@ -518,6 +518,12 @@ def self.default_diffargs
       :desc    => "Stores trusted node data in a hash called $trusted.
         When true also prevents $trusted from being overridden in any scope.",
     },
+    :trusted_server_facts => {
+      :default => false,
+      :type    => :boolean,
+      :desc    => "Stores a trusted set of server-side global variables in a hash
+        called $server_facts, which cannot be overridden by client facts.",
```
hlindberg: "...cannot be overridden by client_facts or logic in manifests. Makes it illegal to assign to the variable $server_facts in any scope."

lib/puppet/node.rb
```diff
@@ -108,12 +108,29 @@ def fact_merge
   # Merge any random parameters into our parameter list.
   def merge(params)
     params.each do |name, value|
-      @parameters[name] = value unless @parameters.include?(name)
+      if @parameters.include?(name)
+        Puppet::Util::Warnings.warnonce("The node paramters #{name} for node #{@parameters["hostname"]} was already set to #{@parameters[name]}. It could not be set to #{value}")
```
hlindberg: It should also set the parameter - now it only warns about something (that it does not do). I.e. line 114 should be here as well.

lib/puppet/node.rb
```diff
@@ -108,12 +108,29 @@ def fact_merge
...
+   # Append the current environment to the list of server facts
+    complete_server_facts = server_facts.merge({ "environment" => self.environment.name.to_s})
+
+   # Set the top scope variable $server_facts if :trusted_server_facts is true
+   if Puppet[:trusted_server_facts]
+     @topscope.set_trusted(complete_server_facts)
+   end
+
+    # Merge the server facts into the parameters for the node
+    merge(server_facts)
```
hlindberg: this could be `merge(complete_server_facts)` and line 118 that also sets the same thing in parameters can be removed.

lib/puppet/parser/scope.rb
```diff
@@ -587,7 +587,7 @@ def define_settings(type, params)
     }
   end
 
-  RESERVED_VARIABLE_NAMES = ['trusted', 'facts'].freeze
+  RESERVED_VARIABLE_NAMES = ['trusted', 'facts', 'server_facts'].freeze
```
hlindberg: This is not correct since 'trusted' and 'facts' are always reserved and 'server_facts' are optional.
This means that the test (where RESERVED_VARIABLE_NAMES) is used must also have a conditional check that it is not 'server_facts'. Once this is no longer optional (in puppet 5.0.0), it can be done as above (but not now).

lib/puppet/node.rb
hkenney: I was hesitant to do this because `merge` also adds the environment to the parameters and is called from places other than `add_server_facts`. So if I remove the addition of environment from `merge` it might not be added when it should be, but obviously when I leave it then it will be added twice. Alternately I could check in merge if it's already there before adding it. Not sure what's best.

lib/puppet/parser/scope.rb
hkenney: Should server_facts be a reserved name when the :trusted_server_facts setting is true?

lib/puppet/node.rb
hkenney: Are we changing the behavior of this? Because if I'm understanding the original code correctly then the parameter would only be updated if it was not already there:

`@parameters[name] = value unless @parameters.include?(name)`

hlindberg: Ah, true - my mistake. Fooled by the "overwriting" going the other direction server_facts on top of facts (and it is the already "written" facts that are "overwriting". You are correct, the values should not be set.

lib/puppet/node.rb
hlindberg: You can leave it the way it was. I was concerned that it may warn about overwrite for environment, but it depends on how many time `merge` is called; it may only be done once. OTOH, it may be expected that the parameter for environment is set after a merge. As long as it is not set more than once it should be fine (and if set more than once, there should be no warning about an "overwrite"

lib/puppet/parser/scope.rb
hlindberg: yes, exactly. (adding it to RESERVED_VARIABLE_NAMES makes always be reserved, which we cannot do since this is an opt in).

lib/puppet/node.rb
```diff
@@ -108,12 +112,24 @@ def fact_merge
   # Merge any random parameters into our parameter list.
   def merge(params)
     params.each do |name, value|
-      @parameters[name] = value unless @parameters.include?(name)
+      if @parameters.include?(name)
+        Puppet::Util::Warnings.warnonce("The node parameter #{name} for node #{@parameters["hostname"]} was already set to #{@parameters[name]}. It could not be set to #{value}")
```
hlindberg: I think there should be single quotes around `#{name}` to make it easier to read. Also noticed that the test did not output anything for the part `#{@parameters["hostname"]}` - maybe that is because of how the test is done. There should be quotes there as well since it looks like the hostname is the empty string.

spec/unit/node_spec.rb
```diff
@@ -219,6 +219,12 @@ def from_json(json)
     expect(@node.parameters["two"]).to eq("b")
   end
 
+  it "warns when a parameter value is not updated" do
+    @node = Puppet::Node.new("testnode", :parameters => {"one" => "a"})
+    Puppet.expects(:warning).with('The node parameter one for node  was already set to a. It could not be set to b')
```
hlindberg: there is no node name in the message, and the name of the parameter should be in single quotes (as commented on above).

MikaelSmith:
I'd really like to know why AppVeyor is failing before merging this. The prior merge to stable had no problems, and I've spent awhile looking at it and I don't see any obvious explanations.

hkenney:
Hmmm... I don't recall any failures until I pushed my most recent change.

MikaelSmith:
I haven't been able to reproduce the failure, so it may be transitory. Trying a re-run, and if it is I'll file a separate ticket for digging into it.

MikaelSmith:
Passed on re-run, I'll file a ticket.

PR #3882, opened 2015-04-30
(PUP-4483) - DO NOT MERGE - Add NotUndef type to the Puppet type system

thallgren:
This commit adds a new NotUndef type. The purpose of this type is
to constrain its contained type so that it is not assignable from
the Undef type. The contained type is optional and defaults to
Any.


lib/puppet/pops/types/type_factory.rb
```diff
@@ -351,6 +351,22 @@ def self.hash_of_data()
...
+  # the type String[inst_type].
+  #
+  # @param inst_type [Type,String] the type to qualify
+  # @return [Puppet::Pops::Types::PNotUndefType] the NotUndef type
+  #
+  # @api public
+  #
+  def self.not_undef(inst_type = nil)
+    type = Types::PNotUndefType.new()
+    inst_type = string(inst_type) if inst_type.instance_of?(String)
```
hlindberg: We typically use .is_a? everywhere instead of instance_of? which requires that the object is an instance of exactly that class, and not a subclass. This can something screw things up when running tests and doing mocking.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -1467,7 +1476,15 @@ def assignable_PResourceType(t1, t2)
   # Data is assignable by other Data and by Array[Data] and Hash[Scalar, Data]
   # @api private
   def assignable_PDataType(t, t2)
-    t2.is_a?(Types::PDataType) || assignable?(@data_variant_t, t2)
+    # We cannot put the NotUndefType[Data] in the @data_variant_t since that causes an endless recursion
+    case t2
+    when Types::PDataType
+      true
+    when Types::PNotUndefType
+      assignable?(t, t2.type)
```
hlindberg: t2.type can be nil, is that handled elsewhere?

lib/puppet/pops/types/type_factory.rb
thallgren: OK, but would anyone mock a String? The reason instance_of? is used here is to avoid is_a? (slow has hell in rgen).

spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -584,6 +590,42 @@ class Foo
...
+          t2 = c.new
+          if tc.assignable?(t.type, t2) && !tc.assignable?(t2, undef_t)
+            expect(t2).to be_assignable_to(t)
+          else
+            expect(t2).not_to be_assignable_to(t)
+          end
+        end
+      end
+
+      it 'NotUndef is not assignable to anything but NotUndef or Any' do
```
hlindberg: it must be possible to assign T to NotUndef[T] (except if T is undef) and NotUndef[T] to T if the corresponding T's are assignable.

lib/puppet/pops/types/type_factory.rb
hlindberg: is `kind_of?` also slow? (it may be just an alias).
My comment was that whenever you see `is_instance?` there is chance it is a typo and should be `is_a?`. ok to use `is_instance?`, but requires a comment so someone else does not change it, or has to figure out why it is used.

spec/unit/pops/types/type_calculator_spec.rb
thallgren: That's not entirely true. Consider:

    assignable?(NotUndef, NotUndef[Any]) == true
    assignable?(NotUndef, Any) == false

The rule for the RHS type must be that `NotUndef[T]` becomes `T` only when `assignable?(T, Undef) == false`

hlindberg: yes

thallgren:
Retargeting for 3.x

PR #3883, opened 2015-04-30
(PUP-4420) Ensure pluginfacts don't ignore source permissions

whopper:
Prior to this commit, the :source_permissions attribute was
not being properly propagated into the Posix MetaStat class,
and so source permissions were ignored for all resources.

External facts are required to use source permissions so
that the facts are executable when they are downloaded onto
the agent. This means that due to this regression, external
facts were being pluginsync'd to the client without execution
permissions.

This commit re-enables using source permissions for fact
pluginsync, while continuing to ignore them otherwise.

See commit 21edcd6 for the first fix for of a similar, earlier
regression.

### DIFF
--- a/acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
+++ b/acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
@@ -1,0 +1,115 @@
+test_name "Pluginsync'ed external facts should be resolvable on the agent"
+
+#
+# This test is intended to ensure that external facts downloaded onto an agent via
+# pluginsync are resolvable. In Linux, the external fact should have the same
+# permissions as its source on the master.
+#
+
+step "Create a codedir with a manifest and test module with external fact"
+codedir = master.tmpdir('4420-codedir')
+
+site_manifest_content =<<EOM
+node default {
+  include mymodule
+  notify { "foo is ${foo}": }
+}
+EOM
+
+unix_fact =<<EOM
+#!/bin/sh
+echo "foo=bar"
+EOM
+
+win_fact =<<EOM
+@echo off
+echo foo=bar
+EOM
+
+if agent['platform'] =~ /windows/
+  fact_content = win_fact
+  extension = '.bat'
+else
+  fact_content = unix_fact
+  extension = '.sh'
+end
+
+apply_manifest_on(master, <<MANIFEST, :catch_failures => true)
+File {
+  ensure => directory,
+  mode   => "0755",
+  owner  => #{master.puppet['user']},
+  group  => #{master.puppet['group']},
+}
+
+file {
+  '#{codedir}':;
+  '#{codedir}/environments':;
+  '#{codedir}/environments/production':;
+  '#{codedir}/environments/production/manifests':;
+  '#{codedir}/environments/production/modules':;
+  '#{codedir}/environments/production/modules/mymodule':;
+  '#{codedir}/environments/production/modules/mymodule/manifests':;
+  '#{codedir}/environments/production/modules/mymodule/facts.d':;
+}
+
+file { '#{codedir}/environments/production/manifests/site.pp':
+  ensure => file,
+  content => '#{site_manifest_content}',
+}
+
+file { '#{codedir}/environments/production/modules/mymodule/manifests/init.pp':
+  ensure => file,
+  content => 'class mymodule {}',
+}
+
+file { '#{codedir}/environments/production/modules/mymodule/facts.d/external_fact#{extension}':
+  ensure  => file,
+  content => '#{fact_content}',
+}
+MANIFEST
+
+master_opts = {
+  'main' => {
+    'environmentpath' => "#{codedir}/environments"
+  }
+}
+
+with_puppet_running_on master, master_opts, codedir do
+  agents.each do |agent|
+    factsd   = agent.tmpdir('facts.d')
+    tmpdir   = agent.tmpdir('tmpdir')
+    testfile = File.join(tmpdir, 'testfile')
+
+    teardown do
+      on(master, "rm -rf #{codedir}")
+      on(agent, "rm -rf #{factsd}")
+    end
+
+    step "Pluginsync the external fact to the agent and ensure it resolves correctly"
+    on(agent, puppet('agent', "-t --server #{master} --pluginfactdest #{factsd}"), :acceptable_exit_codes => [2])
+    assert_match(/foo is bar/, stdout)
+
+    # Linux specific tests
+    next if agent['platform'] =~ /windows/
+
+    step "In Linux, ensure the pluginsync'ed external fact has the same permissions as its source"
+    on(agent, puppet('resource', "file #{factsd}/external_fact#{extension}"))
+    assert_match(/0755/, stdout)
+
+    step "In Linux, ensure puppet apply uses the correct permissions"
+    test_source = File.join('/', 'tmp', 'test')
+    on(agent, puppet('apply', "-e \"file { '#{test_source}': ensure => file, mode => '0456' }\""))
+
+    {'source_permissions => use,'    => /0456/,
+     'source_permissions => ignore,' => /0644/,
+     ''                              => /0644/
+    }.each do |source_permissions, mode|
+      on(agent, puppet('apply', "-e \"file { '/tmp/test_target': ensure => file, #{source_permissions} source => '#{test_source}' }\""))
+      on(agent, puppet('resource', "file /tmp/test_target"))
+      assert_match(mode, stdout)
+
+      on(agent, "rm /tmp/test_target")
+    end
+  end
+end
--- a/lib/puppet/file_serving/metadata.rb
+++ b/lib/puppet/file_serving/metadata.rb
@@ -12,7 +12,7 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
   extend Puppet::Indirector
   indirects :file_metadata, :terminus_class => :selector
 
-  attr_reader :path, :owner, :group, :mode, :checksum_type, :checksum, :ftype, :destination
+  attr_reader :path, :owner, :group, :mode, :checksum_type, :checksum, :ftype, :destination, :source_permissions
 
   PARAM_ORDER = [:mode, :ftype, :owner, :group]
 
@@ -22,10 +22,16 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
     @checksum_type = type
   end
 
+  def source_permissions=(source_permissions)
+    raise(ArgumentError, "Unsupported source_permission #{source_permissions}") unless [:use, :use_when_creating, :ignore].include?(source_permissions.intern)
+
+    @source_permissions = source_permissions.intern
+  end
+
   class MetaStat
     extend Forwardable
 
-    def initialize(stat, source_permissions = nil)
+    def initialize(stat, source_permissions)
       @stat = stat
       @source_permissions_ignore = (!source_permissions || source_permissions == :ignore)
     end
@@ -50,7 +56,7 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
       require 'puppet/util/windows/security'
     end
 
-    def initialize(stat, path, source_permissions = nil)
+    def initialize(stat, path, source_permissions)
       super(stat, source_permissions)
       @path = path
       raise(ArgumentError, "Unsupported Windows source permissions option #{source_permissions}") unless @source_permissions_ignore
@@ -66,13 +72,13 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
     end
   end
 
-  def collect_stat(path, source_permissions)
+  def collect_stat(path)
     stat = stat()
 
     if Puppet.features.microsoft_windows?
-      WindowsStat.new(stat, path, source_permissions)
+      WindowsStat.new(stat, path, @source_permissions)
     else
-      MetaStat.new(stat, source_permissions)
+      MetaStat.new(stat, @source_permissions)
     end
   end
 
@@ -82,7 +88,7 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
   def collect(source_permissions = nil)
     real_path = full_path
 
-    stat = collect_stat(real_path, source_permissions)
+    stat = collect_stat(real_path)
     @owner = stat.owner
     @group = stat.group
     @ftype = stat.ftype
@@ -130,7 +136,6 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
         },
         'type'         => ftype,
         'destination'  => destination,
-
       }
     )
   end
--- a/lib/puppet/file_serving/terminus_helper.rb
+++ b/lib/puppet/file_serving/terminus_helper.rb
@@ -26,7 +26,8 @@ module Puppet::FileServing::TerminusHelper
     end
 
     Puppet::FileServing::Fileset.merge(*filesets).collect do |file, base_path|
-      path2instance(request, base_path, :ignore_source_permissions => true, :relative_path => file)
+      ignore_source_permissions = (!request.options[:source_permissions] || request.options[:source_permissions] == :ignore)
+      path2instance(request, base_path, :ignore_source_permissions => ignore_source_permissions, :relative_path => file)
     end
   end
 end
--- a/lib/puppet/file_serving/terminus_helper.rb
+++ b/lib/puppet/file_serving/terminus_helper.rb
@@ -2,19 +2,18 @@ require 'puppet/file_serving'
 require 'puppet/file_serving/fileset'
 
 # Define some common methods for FileServing termini.
+
 module Puppet::FileServing::TerminusHelper
   # Create model instance for a file in a file server.
   def path2instance(request, path, options = {})
     result = model.new(path, :relative_path => options[:relative_path])
-    result.checksum_type = request.options[:checksum_type] if request.options[:checksum_type]
     result.links = request.options[:links] if request.options[:links]
 
-    # :ignore_source_permissions is here pending investigation in PUP-3906.
-    if options[:ignore_source_permissions]
-      result.collect
-    else
-      result.collect(request.options[:source_permissions])
-    end
+    result.checksum_type = request.options[:checksum_type] if request.options[:checksum_type]
+    result.source_permissions = request.options[:source_permissions] if request.options[:source_permissions]
+
+    result.collect
+
     result
   end
 
@@ -26,8 +25,7 @@ module Puppet::FileServing::TerminusHelper
     end
 
     Puppet::FileServing::Fileset.merge(*filesets).collect do |file, base_path|
-      ignore_source_permissions = (!request.options[:source_permissions] || request.options[:source_permissions] == :ignore)
-      path2instance(request, base_path, :ignore_source_permissions => ignore_source_permissions, :relative_path => file)
+      path2instance(request, base_path, :relative_path => file)
     end
   end
 end
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -668,6 +668,7 @@ Puppet::Type.newtype(:file) do
       :links => self[:links],
       :recurse => (self[:recurse] == :remote ? true : self[:recurse]),
       :recurselimit => self[:recurselimit],
+      :source_permissions => self[:source_permissions],
       :ignore => self[:ignore],
       :checksum_type => (self[:source] || self[:content]) ? self[:checksum] : :none,
       :environment => catalog.environment_instance
--- a/spec/unit/file_serving/metadata_spec.rb
+++ b/spec/unit/file_serving/metadata_spec.rb
@@ -179,7 +179,7 @@ describe Puppet::FileServing::Metadata, :uses_checksums => true do
 
       stat = Puppet::FileSystem.stat(path)
 
-      win_stat = Puppet::FileServing::Metadata::WindowsStat.new(stat, path)
+      win_stat = Puppet::FileServing::Metadata::WindowsStat.new(stat, path, :ignore)
 
       expect(win_stat.owner).to eq('S-1-5-32-544')
       expect(win_stat.group).to eq('S-1-0-0')
--- a/spec/unit/file_serving/terminus_helper_spec.rb
+++ b/spec/unit/file_serving/terminus_helper_spec.rb
@@ -26,12 +26,11 @@ describe Puppet::FileServing::TerminusHelper do
 
   it "should pass through links, checksum_type, and source_permissions" do
     file = stub 'file', :checksum_type= => nil, :links= => nil, :collect => nil
-    [[:checksum_type, :sha256], [:links, true]].each {|k, v|
+    [[:checksum_type, :sha256], [:links, true], [:source_permissions, :use]].each {|k, v|
       file.expects(k.to_s+'=').with(v)
       @request.options[k] = v
     }
-    @request.options[:source_permissions] = :yes
-    file.expects(:collect).with(:yes)
+    file.expects(:collect)
     @model.expects(:new).with("/my/file", {:relative_path => :file}).returns(file)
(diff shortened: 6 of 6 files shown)


whopper:
Putting this up now as I work on an acceptance test to make sure I'm headed in the right direction. @joshcooper given your fix for the exact same regression in June, does this look sane?

lib/puppet/type/file.rb
```diff
@@ -668,6 +668,7 @@ def perform_recursion(path)
       :links => self[:links],
       :recurse => (self[:recurse] == :remote ? true : self[:recurse]),
       :recurselimit => self[:recurselimit],
+      :source_permissions => self[:source_permissions],
```
joshcooper: This looks correct to me. The agent needs to specify whether it wants to use them (or not), and the server should process the request based on the query parameters specified in the `file_metadatas` search request.

But this makes me wonder how it ever worked? I'm guessing the ruby implementation of the puppet file serving code mistakenly had a hardcoded assumption that `file_metadatas` search requests should always use source permissions?

lib/puppet/file_serving/terminus_helper.rb
```diff
@@ -26,7 +26,8 @@ def path2instances(request, *paths)
     end
 
     Puppet::FileServing::Fileset.merge(*filesets).collect do |file, base_path|
-      path2instance(request, base_path, :ignore_source_permissions => true, :relative_path => file)
+      ignore_source_permissions = (!request.options[:source_permissions] || request.options[:source_permissions] == :ignore)
```
joshcooper: The Metadata class also has logic about how to interpret the default `source_permissions` value: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_serving/metadata.rb#L30. I'm thinking we should keep the logic in the Metadata class, and have the terminus_helper treat `source_permissions` like it does `checksum_type`, have the terminus_helper call the `#source_permissions=` method on the model class if specified in `request.options[:source_permissions]`.

I've pushed up a commit https://github.com/joshcooper/puppet/commit/07eb06f7f1517866314fab5459c1abef22959686 to describe what I mean. Amazingly only one test failed (because the test was passing an invalid source_permissions value).

whopper: @joshcooper cool, I'll see about integrating that into this PR!

lib/puppet/type/file.rb
whopper: @joshcooper as far as I could tell, it was working previously because we only ignored source permissions if the option was explicitly set to `:ignore`, and since we weren't actually passing any options in, `source_permissions` was always `nil` (except when we actually were passing `:ignore` in).

`@source_permissions_ignore = source_permissions == :ignore      # resulted in :use when source_permissions was nil`

The change that caused the regression essentially broke this weird assumption that `nil` should use source permissions by saying that if `source_permissions` is `nil`, then we should ignore source permissions. This is why we now need the agent to specify `:use` if it actually needs to use source permissions. Anything that was depending on simply not setting `source_permissions` to `:use` them would be broken. I believe that's just external facts, however.

`@source_permissions_ignore = (!source_permissions || source_permissions == :ignore)`

lib/puppet/file_serving/metadata.rb
```diff
@@ -22,10 +22,16 @@ def checksum_type=(type)
     @checksum_type = type
   end
 
+  def source_permissions=(source_permissions)
+    raise(ArgumentError, "Unsupported source_permission #{source_permissions}") unless [:use, :use_when_creating, :ignore].include?(source_permissions.intern)
```
whopper: @joshcooper ran into some weirdness here. `source_permissions` was being compared to the symbol array as a string, and so Puppet was throwing exceptions.. `intern'd` it for now, but I'm not sure I understand why that's happening (I verified `source_permissions` is a symbol itself).

acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
```diff
@@ -0,0 +1,91 @@
...
+master_opts = {
+  'main' => {
+    'environmentpath' => "#{codedir}/environments"
+  }
+}
+
+step "Pluginsync the external fact to the agent and ensure it resolves correctly"
+with_puppet_running_on master, master_opts, codedir do
+  agents.each do |agent|
+    on(agent, puppet('agent', "-t --server #{master}"), :acceptable_exit_codes => [2])
```
whopper: Second note: the `notify` in the test manifest used here causes Puppet to exit with `2`, so I made that the acceptable exit code.

whopper:
So far I've tested these changes and the acceptance test in linux by running against patched preserved hosts. Next up is Windows.

lib/puppet/file_serving/metadata.rb
```diff
@@ -50,7 +56,7 @@ class WindowsStat < MetaStat
       require 'puppet/util/windows/security'
     end
 
-    def initialize(stat, path, source_permissions = nil)
```
whopper: @joshcooper one windows spec test is failing because it was depending on `source_permissions` being `nil` by default here. Could that be a problem elsewhere, or do we expect to always pass something in? (If `nil` or `ignore` isn't explicitly passed in here, `@source_permissions_ignore` will not be set. Shouldn't it be true by default now?)

lib/puppet/file_serving/content.rb
```diff
@@ -23,7 +23,7 @@ def self.from_binary(content)
 
   # This is no longer used, but is still called by the file server implementations when interacting
   # with their model abstraction.
-  def collect(source_permissions = nil)
+  def collect
```
joshcooper: I probably shouldn't have done this, as it's a breaking change.

lib/puppet/file_serving/metadata.rb
```diff
@@ -66,23 +72,23 @@ def initialize(stat, path, source_permissions = nil)
...
-      MetaStat.new(stat, source_permissions)
+      MetaStat.new(stat, @source_permissions)
     end
   end
 
   # Retrieve the attributes for this file, relative to a base directory.
   # Note that Puppet::FileSystem.stat(path) raises Errno::ENOENT
   # if the file is absent and this method does not catch that exception.
-  def collect(source_permissions = nil)
+  def collect
```
joshcooper: same breaking change comment here

lib/puppet/file_serving/metadata.rb
joshcooper: I would expect a `puppet agent` run to serialize/deserialize the `source_permissions` option as a String, and then call the setter with a string. It may be that `puppet apply` or running via tests is calling with a symbol? I'm thinking we should be consistent about what we store in the instance variable, so that the comparison below `source_permissions == :ignore` work as expected.

whopper:
As far as I can tell, acceptance looks to be in a good state with this PR. It should now be ready to go pending any further review.

acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
```diff
@@ -0,0 +1,115 @@
...
+    testfile = File.join(tmpdir, 'testfile')
+
+    teardown do
+      on(master, "rm -rf #{codedir}")
+      on(agent, "rm -rf #{factsd}")
+    end
+
+    step "Pluginsync the external fact to the agent and ensure it resolves correctly"
+    on(agent, puppet('agent', "-t --server #{master} --pluginfactdest #{factsd}"), :acceptable_exit_codes => [2])
+      assert_match(/foo is bar/, stdout)
```
MikaelSmith: Why the extra space here?

whopper: @MikaelSmith ah, that's just a bit of style I picked up from a few other tests a while back, presumably to make it clear that this assert is connected to the command run above. I can definitely de-indent it.

lib/puppet/file_serving/content.rb
MikaelSmith: Are you proposing to revert the changes to `metadata.rb` and most of the changes to `terminus_helper.rb`? I'm not sure `Puppet::FileServing::Metadata#collect` and `Puppet::FileServing::Content#collect` are a public interface we need to preserve; they're not even part of `Puppet::FileServing::Base`. Putting this in `4.1.0` seems reasonable to me.

acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
MikaelSmith: I hadn't seen it before, and in other languages it implies actual control-flow. Just confused me, if we do it in other tests I don't need it changed.

lib/puppet/file_serving/content.rb
whopper: @MikaelSmith: in reference to removing the default argument here, @joshcooper mentioned he was on the fence about doing it. It'd be easy enough to remove it again (might require a spec update). @joshcooper thoughts?

acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
whopper: @MikaelSmith a cursory glance around other tests tells me I might just be crazy and may have made that up myself :). I'll unindent them.

lib/puppet/file_serving/content.rb
MikaelSmith: Oh, I missed that you put it back.

whopper: Aha, I see! I was a bit confused there :)

joshcooper:
@whopper overall looks good.

One thing I noticed is that we are not including the `source_permission` when serializing/deserializing the metadata in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_serving/metadata.rb#L107-L136. This doesn't affect the overall functionality, but it could result in inconsistent state. For example, during a `pluginfacts` request, the agent requests `source_permissions=use`, and the server will collect metadata including stat'ing the file owner/group/mode. However, the server will call `Metadata#to_data_hash` to serialize the request (omitting `source_permissions` in the process).

The agent will use `Metadata.from_data_hash` to deserialize the request, but since `source_permissions` aren't specified, the state of the `@source_permissions` field will depend on the default (for that version of puppet agent).

It would probably be better to include `source_permissions` in serialization, so that the state of the Metadata object matches the various inputs that went into their collection.

joshcooper:
When sourcing pluginfacts, `use` and `use_when_creating` return the source owner, group, and mode `0755` (in decimal `493`) And, `ignore` and the default return `Process.euid`, `Process.egid`, `0644` (in decimal `420`). When an invalid permission is specified the server generates an HTTP 400.

```
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadatas/pluginfacts?environment=production&links=follow&recurse=true&ignore=.svn&ignore=CVS&ignore=.git&checksum_type=sha256&source_permissions=use' | jq 'map(.relative_path, .owner, .group, .mode)'
[
  ".",
  0,
  0,
  493,
  "bash.sh",
  0,
  0,
  493,
  "text.txt",
  0,
  0,
  420
]
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadatas/pluginfacts?environment=production&links=follow&recurse=true&ignore=.svn&ignore=CVS&ignore=.git&checksum_type=sha256&source_permissions=use_when_creating' | jq 'map(.relative_path, .owner, .group, .mode)'
[
  ".",
  0,
  0,
  493,
  "bash.sh",
  0,
  0,
  493,
  "text.txt",
  0,
  0,
  420
]
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadatas/pluginfacts?environment=production&links=follow&recurse=true&ignore=.svn&ignore=CVS&ignore=.git&checksum_type=sha256&source_permissions=ignore' | jq 'map(.relative_path, .owner, .group, .mode)'
[
  ".",
  501,
  20,
  420,
  "bash.sh",
  501,
  20,
  420,
  "text.txt",
  501,
  20,
  420
]
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadatas/pluginfacts?environment=production&links=follow&recurse=true&ignore=.svn&ignore=CVS&ignore=.git&checksum_type=sha256' | jq 'map(.relative_path, .owner, .group, .mode)'
[
  ".",
  501,
  20,
  420,
  "bash.sh",
  501,
  20,
  420,
  "text.txt",
  501,
  20,
  420
]
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadatas/pluginfacts?environment=production&links=follow&recurse=true&ignore=.svn&ignore=CVS&ignore=.git&checksum_type=sha256&source_permissions=foobar' -w '\n%{http_code}\n'
Unsupported source_permission foobar
400
```

joshcooper:
When sourcing a file (with a puppet URI), I see similar behavior, with mode 0777 (or 511 in decimal):

```
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadata/modules/mymodule/somecontent.txt?environment=production&links=manage&checksum_type=md5&source_permissions=use' | jq '{path, owner, group, mode}'
{
  "path": "/Users/josh/test/master/source_perms/code/environments/production/modules/mymodule/files/somecontent.txt",
  "owner": 0,
  "group": 0,
  "mode": 511
}
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadata/modules/mymodule/somecontent.txt?environment=production&links=manage&checksum_type=md5&source_permissions=use_when_creating' | jq '{path, owner, group, mode}'
{
  "path": "/Users/josh/test/master/source_perms/code/environments/production/modules/mymodule/files/somecontent.txt",
  "owner": 0,
  "group": 0,
  "mode": 511
}
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadata/modules/mymodule/somecontent.txt?environment=production&links=manage&checksum_type=md5&source_permissions=ignore' | jq '{path, owner, group, mode}'
{
  "path": "/Users/josh/test/master/source_perms/code/environments/production/modules/mymodule/files/somecontent.txt",
  "owner": 501,
  "group": 20,
  "mode": 420
}
0 ~/work/puppet (pr/3883)  $ curl -sk -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadata/modules/mymodule/somecontent.txt?environment=production&links=manage&checksum_type=md5' | jq '{path, owner, group, mode}'{
  "path": "/Users/josh/test/master/source_perms/code/environments/production/modules/mymodule/files/somecontent.txt",
  "owner": 501,
  "group": 20,
  "mode": 420
}
0 ~/work/puppet (pr/3883)  $ curl -k -E ./arcturus.p12:password -H "Accept: pson" 'https://arcturus:8140/puppet/v3/file_metadata/modules/mymodule/somecontent.txt?environment=production&links=manage&checksum_type=md5&source_permissions=foobar' -w "\n%{http_code}\n"
Unsupported source_permission foobar
400
```

whopper:
@joshcooper makes sense. I'll get this updated as soon as I finish up another (unrelated) PR

whopper:
@joshcooper pushed your suggested change but I won't have a chance to test it / see if there are other needed changes until tomorrow D:

whopper:
@joshcooper update: ran this through acceptance and some manual tests, and everything looks to be in order. Also updated a few specs and the metadata schema.

whopper:
@joshcooper doh, not sure how I missed that :P. Updated

lib/puppet/file_serving/metadata.rb
```diff
@@ -128,9 +135,9 @@ def to_data_hash
           'type'   => checksum_type,
           'value'  => checksum
         },
-        'type'         => ftype,
-        'destination'  => destination,
-
+        'type'               => ftype,
+        'destination'        => destination,
+        'source_permissions' => source_permissions,
```
joshcooper: @nwolfe previously, puppet's ruby implementation of the `file_metadata` endpoint did not include `source_pemissions` when serializing the response. This lead to inconsistent state on the agent if the agent requested `source_permissions` that were different from its default. For example, a 4.0 agent could request `use`, but the deserialized Metadata objects would have an unspecified `source_permissions` value, and the default value in 4 is `ignore`.

We've decided to include `source_permissions` in the response, similar to `checksum_type`. This should not affect compatibility, because the agent doesn't require or use the `source_permissions` value of objects it gets back. But we likely should modify the clojure endpoint to be consistent. Should I file a SERVER ticket?

whopper:
@joshcooper @kylog Added `source_permissions` to the list of required properties, and sadly blew away your comments in the process :(. For posterity:

Josh:
kylog we're modifying the file_metadata rest endpoint to include the source_permissions that were "used" to generate the Metadata object. Do we also need to include source_permissions in the list of requiredProperties? I'm not super familiar with the api schemas.

I also don't want to break compatibility, e.g. a 4.0.0 agent should be able to talk to 4.1.0 server and vice-versa. A 4.1.0 server (including this info) will not break older clients, because old clients don't try to read in the property, and ignore values they don't recognize.

A 4.1.0 client talking to 4.0.0 server will try to read the value, but default to ignore if it is not specified in the serialized data hash.

That said, I'll definitely try that out to make sure.

Kylo:

Ah, yes, good catch. I would add source_permissions to the list of required properties, since (with this PR in place) it would be an error if the generated json did not include source_permissions. Adding it to required ensures that the specs that check against the schema, e.g. https://github.com/puppetlabs/puppet/blob/e3d4a73c4e950452ed532e779e1094aa8ad54068/spec/unit/file_serving/metadata_spec.rb#L135-L137, will enforce that behavior.

Independent of that, I agree on doing some manual verification for both compatibility directions. Strictly speaking, we require updating the server first, but as a matter of practice many people don't, so it'd be nice not to introduce a newer-client-older-server incompatibility if we don't have to.

kylog:
<shakes fist at github comment support>

lib/puppet/file_serving/metadata.rb
nwolfe: @joshcooper Thanks for the explanation, that fix sounds kosher.

> But we likely should modify the clojure endpoint to be consistent

We don't yet have a Clojure endpoint for this, we just forward the request into Puppet via JRuby.  I suspect that upgrading our Puppet dependency will be enough for Puppet Server.

> Should I file a SERVER ticket?

Yes, please!  I don't expect much, if any, work to come out of this, but we do have some acceptance test(s) that we can clean up (un-skip) once this fix comes in so it would be good to have a ticket for it.

joshcooper:
Unfortunately this breaks 4.0.0 agents talking to 4.1.0 masters. The agent tries to deserialize the response, and because of the MethodHelper tries to call the `Metadata#source_permissions=` method, which doesn't exist in 4.0.0...

```
[root@hwvft1j1a4531gm environments]# puppet agent -td --server arcturus.delivery.puppetlabs.net --trace
...
Error: /File[/opt/puppetlabs/puppet/cache/lib]: Failed to generate additional resources using 'eval_generate': Could not intern_multiple from text/pson: Invalid parameter source_permissions to object class Puppet::FileServing::Metadata
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/methodhelper.rb:16:in `block in set_options'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/methodhelper.rb:11:in `each'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/methodhelper.rb:11:in `set_options'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/file_serving/base.rb:41:in `initialize'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/file_serving/metadata.rb:118:in `initialize'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/file_serving/metadata.rb:139:in `new'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/file_serving/metadata.rb:139:in `from_data_hash'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/network/formats.rb:110:in `data_to_instance'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/network/formats.rb:94:in `block in intern_multiple'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/network/formats.rb:93:in `collect'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/network/formats.rb:93:in `intern_multiple'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/network/format_support.rb:18:in `convert_from_multiple'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/indirector/rest.rb:237:in `deserialize_search'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/indirector/rest.rb:147:in `search'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/indirector/file_metadata/selector.rb:19:in `search'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/indirector/indirection.rb:260:in `search'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type/file.rb:666:in `perform_recursion'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type/file.rb:635:in `block in recurse_remote'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type/file.rb:634:in `collect'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type/file.rb:634:in `recurse_remote'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type/file.rb:562:in `recurse'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type/file.rb:463:in `eval_generate'
```

whopper:
@joshcooper well shoot.. shall we sync up tomorrow and figure out how to work around this?

joshcooper:
So it looks like the server can't include `source_permissions` without breaking existing clients. Although `Puppet::FileServing::Metadata.from_data_hash` ignores keys it doesn't know about, it calls the super method `Puppet::FileServing::Base.from_data_hash` which relies on `MethodHelper` (and will blindly call setters for each key-value in the hash). And if the class doesn't respond to a particular setter, it raises an exception: `raise ArgumentError, "Invalid parameter #{param} to object class #{self.class}"`

So I'm thinking we should:

1. Drop commit 6a5e81d74 and merge this PR.
2. File a new ticket for 4.1.0 to future-proof `Base.from_data_hash` so it will ignore parameters it doesn't know about. 
3. File a new ticket for 5.0 to essentially add commit 6a5e81d74 back in.

We could also consider modifying the two places where we call `Metadata.indirection.{find,search}` and explicitly set `source_permissions=` on the returned objects (similar to how we call `Metadata#source=`)? On one hand it would ensure the returned Metadata objects are in a consistent state. But on the other hand, it feels like a bandaid that would only confuse things (some data is specified on the server, other data is specified on the client). And if intend to support 4.x agents talking to 4.x and 5.x masters, things get really confusing.

whopper:
@joshcooper I went ahead and dropped 6a5e81d74 for the time being.

MikaelSmith:
Merged #3933, can you rebase and verify the acceptance test works?

whopper:
@MikaelSmith rebased, and the test indeed passes (also verified that it failed *before* rebasing)

PR #3884, opened 2015-05-01
(PUP-4511) Ensure that unspecified contained types default to Any

thallgren:
The `Array`, `Hash`, and `Optional` all have contained types that may
be `nil`. The `TypeCalculator.assignable?` method didn't handle this
well (it passed `nil` in a recursive call to `assignable?`, an
operation that is permitted but will yield a `false` result).

This commit adds checks to ensure that if the contained type is `nil`,
then it will be considered to be of type `Any`.

Some tests were changed to accommodate the fact that `Optional[?]` now
considers everything to be assignable whereas nothing was assignable
before.

### DIFF
--- a/spec/unit/pops/types/type_calculator_spec.rb
+++ b/spec/unit/pops/types/type_calculator_spec.rb
@@ -553,6 +553,13 @@ describe 'The type calculator' do
   context 'computes assignability' do
     include_context "types_setup"
 
+    it 'such that all types are assignable to themselves' do
+      all_types.each do |tc|
+        t = tc.new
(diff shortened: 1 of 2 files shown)


spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -578,7 +585,7 @@ class Foo
       end
 
       it 'Any is not assignable to anything but Any' do
```
hlindberg: seems to test not assignable to anything but Any and Optional (implied Optional[Any]) - change the title of the test?

PR #3885, opened 2015-05-01
(maint) Fix/update AppVeyor settings

ferventcoder:
The appveyor script gives the appearance of setting to Ruby 2.1.5,
but that particular folder doesn't actually exist, so it is actually running with Ruby 
1.9.3. When the build runs it should share the environment variables, the 
Gemfile.lock file and the Ruby version it is set on. ~~Additionally it should test both 
x86 and x64 architectures.~~

Additionally we should test against the x64 architecture, which should catch a 
reasonable amount of failures related to Windows. Adding a second architecture 
slows the build down quite a bit (to more than double the amount of time),
so we won't go the route of adding a matrix for now.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,22 @@
-version: 4.0.0.{build}
-platform:
-    - x64
+version: 4.1.0.{build}
+skip_commits:
+  message: /(^\(?doc\)?.*|.*[A|a]cceptance [T|t]est.*)/
+clone_depth: 10
+init:
+  - SET
 install:
-    - SET PATH=C:\Ruby215\bin;%PATH%
-    - gem install bundler --quiet --no-ri --no-rdoc
-    - bundle install --without development
+  - SET PATH=C:\Ruby21-x64\bin;%PATH%
+  - gem install bundler --quiet --no-ri --no-rdoc
+  - bundle install --jobs 4 --retry 2 --without development
+  - type Gemfile.lock
+  - ruby -v
 build: off
 test_script:
-    - bundle exec rspec spec
+  - bundle exec rspec spec
+notifications:
+  - provider: Email
+    to:
+    - nobody@nowhere.com
+    on_build_success: false
+    on_build_failure: false
+    on_build_status_changed: false


ferventcoder:
@MikaelSmith @joshcooper here are the fixes for appveyor to ensure that it is running against 2.1.x instead of 1.9.3 (the default Ruby on the build workers). I verified that the Ruby215 folder doesn't exist, the documentation states it is somewhere else http://www.appveyor.com/docs/installed-software#languages-libraries-frameworks, and I verified that with http://help.appveyor.com/discussions/suggestions/664-ruby-216-support

Here's a build that shows that we don't switch to ruby 2.1.x - https://ci.appveyor.com/project/puppetlabs/puppetlabs-acl/build/1.1.x.14/job/2tidt88gcfd8wetk#L154

ferventcoder:
And it looks like there is a failure against x64 ruby 2.1.x - Rekicking appveyor

appveyor.yml
```diff
@@ -1,10 +1,19 @@
...
+  matrix:
+  - RUBY_VER: 21
+  - RUBY_VER: 21-x64
 install:
-    - SET PATH=C:\Ruby215\bin;%PATH%
-    - gem install bundler --quiet --no-ri --no-rdoc
-    - bundle install --without development
+  - SET PATH=C:\Ruby%RUBY_VER%\bin;%PATH%
+  - gem install bundler --quiet --no-ri --no-rdoc
+  - bundle install --jobs 4 --retry 2 --without development
```
ferventcoder: This should speed up bundler a bit.

appveyor.yml
```diff
@@ -1,10 +1,19 @@
...
+  - SET
+environment:
+  matrix:
+  - RUBY_VER: 21
+  - RUBY_VER: 21-x64
 install:
-    - SET PATH=C:\Ruby215\bin;%PATH%
-    - gem install bundler --quiet --no-ri --no-rdoc
-    - bundle install --without development
+  - SET PATH=C:\Ruby%RUBY_VER%\bin;%PATH%
```
ferventcoder: Setting the ruby folder appropriately.

appveyor.yml
```diff
@@ -1,10 +1,19 @@
...
+  - gem install bundler --quiet --no-ri --no-rdoc
+  - bundle install --jobs 4 --retry 2 --without development
+  - type Gemfile.lock
+  - ruby -v
```
ferventcoder: Report on the ruby version that we are about to use for spec testing.

appveyor.yml
```diff
@@ -1,10 +1,19 @@
...
 build: off
 test_script:
-    - bundle exec rspec spec
+  - bundle exec rspec spec -fd -b
```
ferventcoder: Let's format with a full backtrace. Thoughts @joshcooper @MikaelSmith

appveyor.yml
```diff
@@ -1,10 +1,19 @@
-version: 4.0.0.{build}
+version: 4.1.0.{build}
 platform:
-    - x64
+  - x64
```
ferventcoder: I'm not sure that using platform here gives us any value. Where did you find this?

appveyor.yml
```diff
@@ -1,10 +1,19 @@
...
+clone_depth: 10
+init:
+  - SET
```
ferventcoder: Report on environment variables.

appveyor.yml
```diff
@@ -1,10 +1,19 @@
-version: 4.0.0.{build}
+version: 4.1.0.{build}
 platform:
-    - x64
+  - x64
+clone_depth: 10
```
ferventcoder: Reduce clone depth to speed up checkout.

appveyor.yml
```diff
@@ -1,10 +1,19 @@
...
+init:
+  - SET
+environment:
+  matrix:
+  - RUBY_VER: 21
+  - RUBY_VER: 21-x64
```
ferventcoder: Introduce a matrix so we are testing both x86 and x64.

appveyor.yml
```diff
@@ -1,10 +1,19 @@
-version: 4.0.0.{build}
+version: 4.1.0.{build}
```
ferventcoder: Updated the build number here to match current stable.

appveyor.yml
ferventcoder: Does this slow down the run some?

ferventcoder:
It does appear that the addition of a matrix has taken the run from <15 minutes to around 40 minutes. We may want to choose one of the two platforms. We can reasonably assume that we'd get about the same amount of benefit from just one of the two. I'd suggest x64 ruby.

appveyor.yml
```diff
@@ -1,10 +1,15 @@
...
+clone_depth: 10
+init:
+  - SET
 install:
-    - SET PATH=C:\Ruby215\bin;%PATH%
-    - gem install bundler --quiet --no-ri --no-rdoc
-    - bundle install --without development
+  - SET PATH=C:\Ruby21-x64\bin;%PATH%
+  - gem install bundler --quiet --no-ri --no-rdoc
+  - bundle install --jobs 4 --retry 2 --without development
```
ferventcoder: This should speed up the bundle install and protect against network glitches

appveyor.yml
ferventcoder: I removed the matrix due to the large amount of runtime.

appveyor.yml
ferventcoder: I remove this as it appears to slow down the build.

appveyor.yml
MikaelSmith: http://www.appveyor.com/docs/appveyor-yml, when I was thinking about testing 32-bit and 64-bit Windows. Didn't end up using it.

ferventcoder: That looks .NET / Visual Studio specific.

MikaelSmith: Oh yeah, it does. Can drop it then.

ferventcoder: I'll do that.

ferventcoder: Done.

appveyor.yml
MikaelSmith: The one thing I'd add is `--without extra`.

MikaelSmith: AppVeyor speed seems comparable to my Windows VM.

MikaelSmith: I think Ruby 2.1.5 is just slower running specs.

MikaelSmith:
As far as I can tell this is the best we can do, so merging.

PR #3886, opened 2015-05-01
(PUP-4466) Add acceptance for functions in the puppet language

kurtwall:
Validate functions written in the Puppet DSL. They are under `.functions` relative to either a module or an environment. Puppet functions are handled by the 4.x function API, so they follow the visibility rules set based on a module's dependencies:
* An environment sees everything in every module on the module path
* A module with dependencies only sees what it is the environment and in modules it has a dependency on
* A module without dependencies (no declaration at all) can see all other modules

### DIFF
--- a/acceptance/tests/language/invoke_pp_function_from_dependent_module.rb
+++ b/acceptance/tests/language/invoke_pp_function_from_dependent_module.rb
@@ -1,0 +1,67 @@
+test_name 'Puppet-language functions can be called from inside an another module' do
+
+  mini_func = 'function jenny::mini($a, $b) {if $a <= $b {$a} else {$b}}'
+  maxi_func = 'function jenny::nested::maxi($a, $b) {if $a >= $b {$a} else {$b}}'
+  
+  step 'Create PP functions on master' do
+    manifest = <<-EOM.gsub(/^ {6}/,'')
+      File {
+        ensure => present,
+        owner => 'puppet',
+        group => 'puppet',
+        mode => '0644',
+      }
+      file {['/etc/puppetlabs',
+          '/etc/puppetlabs/code',
+          '/etc/puppetlabs/code/modules',
+          '/etc/puppetlabs/code/modules/jenny',
+          '/etc/puppetlabs/code/modules/jenny/functions',
+          '/etc/puppetlabs/code/modules/jenny/functions/nested',
+          '/etc/puppetlabs/code/modules/tutone',
+          '/etc/puppetlabs/code/modules/tutone/manifests']:
+        ensure => directory,
+        owner => 'puppet',
+        group => 'puppet',
+        mode => '0755',
+      }  
+      file {'/etc/puppetlabs/code/modules/jenny/functions/mini.pp':
+        content => '#{mini_func}',
+      }
+      file {'/etc/puppetlabs/code/modules/jenny/functions/nested/maxi.pp':
+        content => '#{maxi_func}',
+      }
+      file {'/etc/puppetlabs/code/modules/tutone/manifests/testmini.pp':
+        content => 'notify {"mini": message => jenny::mini(1,2)}'
+      }
+      file {'/etc/puppetlabs/code/modules/tutone/manifests/testmaxi.pp':
+        content => 'notify {"maxi": message => jenny::nested::maxi(1,2)}'
+      }
+    EOM
+    apply_manifest_on(master, manifest)
+  end
+  
+  step 'Call a simple PP function from another module' do
+    rc = on master, puppet('apply /etc/puppetlabs/code/modules/tutone/manifests/testmini.pp')
+    fail_test "Dependent simple PP function failed; returned #{rc.stdout}" \
+      unless rc.stdout.include? "defined 'message' as '1'"
+  end
+
+  step 'Call a nested PP function from another module' do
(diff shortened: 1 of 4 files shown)


kurtwall:
Tested on RHEL7, Ubuntu 14.04, Fedora 21.

hlindberg:
I can't see a test for a module that has dependencies, but no dependency on the module that defines a function - it should not find that function. (i.e. bullet 2 in the description above).

acceptance/tests/language/invoke_pp_function_from_dependent_module.rb
```diff
@@ -0,0 +1,81 @@
+test_name 'Puppet-language functions can be called from inside an another module' do
+
+  mini_func = 'function jenny::mini($a, $b) {if $a <= $b {$a} else {$b}}'
+  maxi_func = 'function jenny::nested::maxi($a, $b) {if $a >= $b {$a} else {$b}}'
```
er0ck: just an idea, but if we have two functions, maybe the second should do some string operations?

acceptance/tests/language/invoke_pp_function_from_independent_module.rb
```diff
@@ -0,0 +1,50 @@
...
+          '/etc/puppetlabs/code/modules/jenny',
+          '/etc/puppetlabs/code/modules/jenny/functions',
+          '/etc/puppetlabs/code/modules/jenny/functions/nested']:
+        ensure => directory,
+        owner => 'puppet',
+        group => 'puppet',
+        mode => '0755',
+      }
+      
+      file {'/etc/puppetlabs/code/modules/jenny/functions/mini.pp':
```
er0ck: IMO, this test file should be merged with the one above.    acceptance runtimes are an issue and this one repeats a bunch of the setup steps as well as the test for mini function.

acceptance/tests/language/invoke_pp_function_from_multiple_environments.rb
```diff
@@ -0,0 +1,61 @@
...
+  applyme = <<-EOM
+    class f {
+      notify { 'mini': message => jenny::mini(1,2), }
+      notify { 'maxi': message => jenny::nested::maxi(1,2), }
+    }
+    include 'f'
+  EOM
+
+  step 'Call a PP function from two environments' do
+    env_1 = on master, puppet("apply -e \"#{applyme}\" --environment frick")
```
er0ck: we should probably eliminate as many puppet apply calls as possible (see runtime of this on windows).  applyme manifest could be merged with the setup manifest above it.  if you don't want to repeat yourself with the two environments you could abstract it into a function to replace the environment name in the two environment sections.
thoughts?

acceptance/tests/language/invoke_pp_function_from_nondependent_module.rb
```diff
@@ -0,0 +1,58 @@
...
+      file {'/etc/puppetlabs/code/modules/jenny/functions/mini.pp':
+        content => '#{mini_func}',
+      }
+      file {'/etc/puppetlabs/code/modules/jenny/functions/nested/maxi.pp':
+        content => '#{maxi_func}',
+      }
+      file {'/etc/puppetlabs/code/modules/tommy/manifests/init.pp':
+        content => 'class tommy { notify { "tommy says FOO": } }',
+      }
+      file {'/etc/puppetlabs/code/modules/tutone/manifests/testmini.pp':
```
kurtwall: @hlindberg I don't think I'm getting the setup correct to call a puppet function from a module that has dependencies, but not on the module defining the function. I expect lines 36-41 to fail, but they succeed.

acceptance/tests/language/invoke_pp_function_from_multiple_environments.rb
```diff
@@ -0,0 +1,61 @@
...
+    fail_test "Calling PP function failed in environment 'frick'; returned #{env_1.stdout}" \
+      unless env_1.stdout.include? "frick"
+    fail_test "Simple PP function failed; returned #{env_1.stdout}" \
+      unless env_1.stdout.include? "defined 'message' as '1'"
+    fail_test "Nested PP function failed; returned #{env_1.stdout}" \
+      unless env_1.stdout.include? "defined 'message' as '2'"
+
+    env_2 = on master, puppet("apply -e '#{applyme}' --environment frack")
```
er0ck: this other environment test block is probably not useful as the functions are defined in the global function namespace.

acceptance/tests/language/invoke_pp_function_from_nondependent_module.rb
```diff
@@ -0,0 +1,58 @@
...
+      file {'/etc/puppetlabs/code/modules/tutone/manifests/testmaxi.pp':
+        content => 'require "tommy"; notify {"maxi": message => jenny::nested::maxi(1,2)}'
+      }
+    EOM
+    apply_manifest_on(master, manifest)
+  end
+
+  step 'Call a simple PP function from another module' do
+    rc = on master, puppet('apply /etc/puppetlabs/code/modules/tutone/manifests/testmini.pp')
+    fail_test "Dependent simple PP function failed; returned #{rc.stdout}" \
```
er0ck: i don't think there are module dependencies here?

PR #3887, opened 2015-05-01
(maint) Enable version module if it's not already loaded

joshcooper:
The apache `version` module must be loaded to use the <IfVersion>
syntax. On debian 7 and ubuntu 14.04, the module is static, and it is
an error to try to enable it. On ubuntu 12.04, the module is not static,
and must be explicitly enabled, see RE-4465.

This commit enables the `version` module only if it is not already
loaded.


haus:
:+1:

stahnma:
LGTM.

stahnma:
Should we wait for travis and such to finish?

PR #3889, opened 2015-05-04
(PUP-4483) Add NotUndef type to the Puppet type system

thallgren:
This PR adds a new `NotUndef` type. The purpose of this type is to constrain its contained type so that it is not assignable from the `Undef` type. The contained type is optional and defaults to `Any`.

The PR also changes the behavior of the `Struct` type by allowing its members to be types. This change enables declaration of required members who's values may be optional.

### DIFF
--- a/spec/unit/pops/types/type_calculator_spec.rb
+++ b/spec/unit/pops/types/type_calculator_spec.rb
@@ -395,7 +395,7 @@ describe 'The type calculator' do
           t = calculator.infer_set({ 'mode' => 'read', 'path' => ['foo', 'fee' ] })
           expect(t.class).to eq(Puppet::Pops::Types::PStructType)
           expect(t.elements.size).to eq(2)
-          els = t.elements.map { |e| e.type }.sort {|a,b| a.to_s <=> b.to_s }
+          els = t.elements.map { |e| e.value_type }.sort {|a,b| a.to_s <=> b.to_s }
           els[0].class.should == Puppet::Pops::Types::PStringType
           els[1].class.should == Puppet::Pops::Types::PTupleType
         end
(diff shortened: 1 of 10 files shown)


thallgren:
Please note that this PR is supposed to fail due to [PUP-4511](https://tickets.puppetlabs.com/browse/PUP-4511).

spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -832,10 +885,40 @@ class Foo
         tested_types = all_types - [
           Puppet::Pops::Types::PAnyType,
           Puppet::Pops::Types::POptionalType,
+          Puppet::Pops::Types::PNotUndefType,
           Puppet::Pops::Types::PDataType] - collection_types
         t = Puppet::Pops::Types::PStructType.new()
         tested_types.each {|t2| t.should_not be_assignable_to(t2.new) }
       end
+
+      it 'Default key optionality is controlled by value assignability to undef' do
```
hlindberg: This example title does not seem to describe what the test does as it checks if the constructed type has the expected shape.

spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -832,10 +885,40 @@ class Foo
...
+        expect(t1.elements[0].key_type).to eq(optional_t(string_t('member')))
+      end
+
+      it 'Optional members are not required' do
+        t1 = struct_t({optional_t('optional_member') => string_t, not_undef_t('other_member') => string_t})
+        t2 = struct_t({not_undef_t('other_member') => string_t})
+        expect(t2).to be_assignable_to(t1)
+      end
+
+      it 'Required members not optional even when value is' do
```
hlindberg: "Required member is not optional even if value is"

spec/unit/pops/types/type_calculator_spec.rb
thallgren: Not sure I agree. The test verifies that if the value is not assignable to undef, then the key is not wrapped in an Optional and if the value is assignable to undef, the key is wrapped in an Optional. I.e. "the key optionality is controlled by value assignability to undef". Can you find a better short and concise wording for it?

hlindberg: ok, that makes sense - it was difficult to match up the description and the test.

spec/unit/pops/types/type_calculator_spec.rb
hlindberg: This is so minor, or even not a problem when I read it again... and the other comment I made was ok as well. Merging.

PR #3890, opened 2015-05-04
(PUP-4517) Add type name to Puppet.newtype deprecation warning

roman-mueller:
Add type name to Puppet.newtype deprecation warning so it is easier to find and fix those occurrences.


peterhuene:
@roman-mueller Thanks for the contribution! This has been merged into master and will appear in the next release of Puppet.

PR #3891, opened 2015-05-04
(PUP-4532) Add launchd plist for use with OSX AIO

heathseals:
This commit adds a launchd plist for use with OSX AIO packaging.

### DIFF
--- a/ext/osx/puppet.plist
+++ b/ext/osx/puppet.plist
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+        <key>EnvironmentVariables</key>
+        <dict>
+                <key>PATH</key>
+                <string>/opt/puppetlabs/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+                <key>LANG</key>
+                <string>en_US.UTF-8</string>
+        </dict>
+        <key>Label</key>
+        <string>puppet</string>
+        <key>KeepAlive</key>
+        <true/>
+        <key>ProgramArguments</key>
+        <array>
+                <string>/opt/puppetlabs/bin/puppet</string>
+                <string>agent</string>
+                <string>--verbose</string>
+                <string>--no-daemonize</string>
+                <string>--logdest</string>
+                <string>console</string>
+        </array>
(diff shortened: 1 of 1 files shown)


ext/osx/puppet.plist
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+        <key>EnvironmentVariables</key>
+        <dict>
+                <key>PATH</key>
+                <string>/opt/puppetlabs/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+                <key>RUBYLIB</key>
+		<string>/opt/puppetlabs/puppet/lib/ruby/site_ruby/2.1.0/</string>
```
melissa: hmm, is there anyway we could make this an erb and process it as we're installing it, so that we don't have to hardcode the ruby version here?

haus: @heathseals did you try using this without RUBYLIB? it shouldn't be needed with AIO ruby.

heathseals: I'm not certain that RUBYLIB is required at all.  @haus?

heathseals:
@melissa @haus Removed RUBYLIB.    It doesn't appear to be required by AIO ruby.

haus:
excellent. i'm :+1: here.

ext/osx/puppet.plist
```diff
@@ -0,0 +1,32 @@
...
+        <string>puppet</string>
+        <key>KeepAlive</key>
+        <true/>
+        <key>ProgramArguments</key>
+        <array>
+                <string>/opt/puppetlabs/bin/puppet</string>
+                <string>agent</string>
+                <string>--verbose</string>
+                <string>--no-daemonize</string>
+                <string>--logdest</string>
```
melissa: I noticed that we have the mcollective pidfile listed here in the mcollective plist file. We don't need that for puppet?

melissa: That's probably a silly question, but I just wanted to check and make sure

haus: Puppet gets rundir and pidfile from defaults.rb and run_mode.rb, mcollective does not.

haus: mcollective doesn't have a default pidfile location.

melissa: Cool, thanks for the clarification!

PR #3895, opened 2015-05-05
(PUP-735) Ensure reports are finalized when catalog.apply fails

branan:
After the initial work on PUP-735, we were left with a situation where
certain catalog apply errors (including resource cycles) could casue a
report to not be finalized. An easy symptom of this to see is that the
metrics section of the report is empty.

This fixes the configurer to always finalize the report, and adds a
test.

### DIFF
--- a/acceptance/tests/reports/finalized_on_cycle.rb
+++ b/acceptance/tests/reports/finalized_on_cycle.rb
@@ -1,0 +1,37 @@
+test_name "Reports are finalized on resource cycles"
+
+require 'puppet/acceptance/common_utils'
+extend Puppet::Acceptance::CommandUtils
+
+check_script = <<CHECK
+require 'yaml'
+require 'puppet'
+
+exit YAML.load_file(ARGV[0]).metrics.empty? ? 1 : 0
+CHECK
+
+cyclic_manifest = <<MANIFEST
+notify { 'foo':
+  require => Notify['bar']
+}
+
+notify { 'bar':
+  require => Notify['foo']
+}
+MANIFEST
+
+agents.each do |agent|
+  tmpdir = agent.tmpdir('report_finalized')
+  check = "#{tmpdir}/check_report.rb"
+  manifest = "#{tmpdir}/manifest.pp"
+  report = "#{agent.puppet['vardir']}/state/last_run_report.yaml"
+
+  create_remote_file(agent, check, check_script)
+
+  # We can't use apply_manifest_on here because we can't tell it not
+  # to fail the test when it encounters a cyclic manifest.
+  create_remote_file(agent, manifest, cyclic_manifest)
+  on(agent, puppet("apply", manifest), :acceptable_exit_codes => [1])
+  result = on(agent, "#{ruby_command(agent)} #{check} #{report}", :acceptable_exit_codes => [0,1])
+  fail_test("Report was not finalized") if result.exit_code == 1
+end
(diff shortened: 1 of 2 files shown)


acceptance/tests/reports/finalized_on_cycle.rb
```diff
@@ -0,0 +1,48 @@
...
+
+master_opts = {
+  'main' => {
+    'environmentpath' => "#{testdir}/environments",
+    'reports'         => 'store',
+    'reportdir'       => "#{testdir}/reports"
+  },
+}
+
+ruby = '/opt/puppetlabs/puppet/bin/ruby'
```
branan: This is totally not gonna work on Windows. Need to fix this up before merge

branan: `agent['privatebindir']` looks right

acceptance/tests/reports/finalized_on_cycle.rb
```diff
@@ -0,0 +1,48 @@
...
+reports = "#{testdir}/reports"
+check = "#{testdir}/check_report.rb"
+
+with_puppet_running_on(master, master_opts) do
+  agents.each do |agent|
+    # We expect the agent to fail here (because of the cycle in the manifest above)
+    on(agent, puppet('agent', "-t --server #{master}"), :acceptable_exit_codes => [1])
+    on(master, "#{ruby} #{check} #{reports}/#{agent}/*.yaml")
```
joshcooper: I'm wondering if we can shorten the test time by just running `puppet apply` and parsing the report on the agent?

branan: @joshcooper TIL that apply generates reports. I'll refactor

branan:
@joshcooper updated to use apply.

acceptance/tests/reports/finalized_on_cycle.rb
```diff
@@ -0,0 +1,34 @@
...
+}
+
+notify { 'bar':
+  require => Notify['foo']
+}
+MANIFEST
+
+agents.each do |agent|
+  tmpdir = agent.tmpdir('report_finalized')
+  ruby = "#{agent['privatebindir']}/ruby"
```
MikaelSmith: lib/puppet/acceptance/common_utils.rb has a util for getting the local ruby, `Puppet::Acceptance::CommandUtils.ruby_command`. See tests/windows/eventlog.rb for usage.

acceptance/tests/reports/finalized_on_cycle.rb
```diff
@@ -0,0 +1,34 @@
...
+  check = "#{tmpdir}/check_report.rb"
+  manifest = "#{tmpdir}/manifest.pp"
+  report = "#{agent.puppet['vardir']}/state/last_run_report.yaml"
+
+  create_remote_file(agent, check, check_script)
```
MikaelSmith: Do you ever use this script?

branan: I can make the variable names longer/clearer if you want

acceptance/tests/reports/finalized_on_cycle.rb
branan: OK, that looks like a reasonably easy fix.

acceptance/tests/reports/finalized_on_cycle.rb
MikaelSmith: No, they're fine. I just wasn't thinking straight.

branan:
Updated to use the ruby helper, and to fail the test explicitly with a useful error message.

branan:
AppVeyor failure appears to be PUP-4537

MikaelSmith:
Hmm, that's suddenly happening frequently.

branan:
Yeah, it might be worth more investigation sooner rather than later. I probably have time to poke it this afternoon

PR #3896, opened 2015-05-05
(PUP-4030) Add missing require for Puppet::Util::HttpProxy.

peterhuene:
A require for `puppet/util/http_proxy` was missing for
`puppet/network/http/factory` after code that uses
`Puppet::Util::HttpProxy` was added.

This originally passed specs because the spec required the file before
using the HTTP factory.


MikaelSmith:
:+1: waiting on Travis.

PR #3899, opened 2015-05-06
(PUP-4473) Extend ENC tests to Windows

MikaelSmith:
The ENC integration tests using the exec terminus were originally only
implemented with shell scripts. Extend the testing to create Windows batch
scripts and run on Windows.

Also create a new spec_helper script_containing to generate a script on
the appropriate platform. Since script languages also differ, usage
requires that you provide both Windows and shell versions of the script.

### DIFF
--- a/spec/lib/puppet_spec/files.rb
+++ b/spec/lib/puppet_spec/files.rb
@@ -43,6 +43,20 @@ module PuppetSpec::Files
     file
   end
 
+  def script_containing(name, contents) PuppetSpec::Files.script_containing(name, contents) end
+  def self.script_containing(name, contents)
+    file = tmpfile(name)
+    if Puppet.features.microsoft_windows?
+      file += '.bat'
+      text = contents[:windows]
+    else
+      text = contents[:posix]
+    end
+    File.open(file, 'wb') { |f| f.write(text) }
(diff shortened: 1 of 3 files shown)


spec/lib/puppet_spec/files.rb
```diff
@@ -43,6 +43,20 @@ def self.file_containing(name, contents)
...
+  def script_containing(name, contents) PuppetSpec::Files.script_containing(name, contents) end
+  def self.script_containing(name, contents)
+    file = tmpfile(name)
+    if Puppet.features.microsoft_windows?
+      file += '.bat'
+      text = contents[:windows]
+    else
+      text = contents[:posix]
+    end
+    File.open(file, 'w') { |f| f.write(text) }
```
joshcooper: Since ruby defaults to text mode on windows, it will translate "\n" to "\r\n", which I think is ok for scripts. But if this helper is ever used (in the future) to create a file whose checksum we care about, then we might be surprised. Maybe default to 'wb' always, or allow the caller to specify the text/binary mode?

MikaelSmith: Batch scripts seem to be ok with just "\n", so defaulting to 'wb' would probably be safe.

PR #3900, opened 2015-05-07
(PUP-4503, PUP-2023) Debian for systemd

sathieu:


sathieu:
Note that this is not sufficient, because both "debian" dans 'systemd' provider are default. How can I do an "confine :noexists => "/run/systemd/system" in the debian provider ?

sathieu:
The double default provider is fixed in revision 2

sathieu:
NB: there is still this bug to be solved: https://bugs.debian.org/751638

leepa:
This change won't work for the reasons stated on PUP-4503 and the version attached in the Gist on that ticket is probably best until the Puppetlabs guys solve it.

The Debian bug is unlikely to be fixed anytime soon. Shown by the fact it's a year old and during the RCs they didn't fix it.

Please don't merge this change.

melissa:
Closed in favor of https://github.com/puppetlabs/puppet/pull/3916

PR #3904, opened 2015-05-07
(PUP-4524) Stub command execution in base service provider spec.

peterhuene:
Rather than rely on executing ruby to manipulate a file to "emulate" a
service start/status/stop command, stub command execution itself to
ensure the base service provider is executing the commands as expected.


MikaelSmith:
:+1:

PR #3905, opened 2015-05-08
(PUP-4546) Fix retrieving package version in #instances

jasperla:
The output for 'pkgin list' has been changed to separate fields with ';'
for non-interactive terminals, so adjust the regex accordingly. To ensure
we'll always get this output pass '-p' to the list command (this is what pkgin
does internally for non-interactive terminals).

### DIFF
--- a/lib/puppet/provider/package/pkgin.rb
+++ b/lib/puppet/provider/package/pkgin.rb
@@ -5,15 +5,15 @@ Puppet::Type.type(:package).provide :pkgin, :parent => Puppet::Provider::Package
 
   commands :pkgin => "pkgin"
 
-  defaultfor :operatingsystem => [ :dragonfly , :smartos ]
+  defaultfor :operatingsystem => [ :dragonfly , :smartos, :netbsd ]
 
   has_feature :installable, :uninstallable, :upgradeable, :versionable
 
   def self.parse_pkgin_line(package)
 
     # e.g.
-    #   vim-7.2.446 =        Vim editor (vi clone) without GUI
-    match, name, version, status = *package.match(/(\S+)-(\S+)(?: (=|>|<))?\s+.+$/)
+    #   vim-7.2.446;Vim editor (vi clone) without GUI
+    match, name, version, status = *package.match(/([^\s;]+)-([^\s;]+)[;\s](=|>|<)?.+$/)
     if match
       {
         :name     => name,
--- a/spec/unit/provider/package/pkgin_spec.rb
+++ b/spec/unit/provider/package/pkgin_spec.rb
@@ -41,7 +41,7 @@ describe provider_class do
 
   describe "#instances" do
     let(:pkgin_ls_output) do
-      "zlib-1.2.3           General purpose data compression library\nzziplib-0.13.59      Library for ZIP archive handling\n"
+      "zlib-1.2.3;General purpose data compression library\nzziplib-0.13.59;Library for ZIP archive handling\n"
     end
 
     before do
@@ -72,7 +72,7 @@ describe provider_class do
 
     context "when the package is installed" do
       let(:pkgin_search_output) do
-        "vim-7.2.446 =        Vim editor (vi clone) without GUI\nvim-share-7.2.446 =  Data files for the vim editor (vi clone)\n\n=: package is installed and up-to-date\n<: package is installed but newer version is available\n>: installed package has a greater version than available package\n"
+        "vim-7.2.446;=;Vim editor (vi clone) without GUI\nvim-share-7.2.446;=;Data files for the vim editor (vi clone)\n\n=: package is installed and up-to-date\n<: package is installed but newer version is available\n>: installed package has a greater version than available package\n"
       end
 
       it "returns installed version" do
@@ -83,7 +83,7 @@ describe provider_class do
 
     context "when the package is out of date" do
       let(:pkgin_search_output) do
-        "vim-7.2.447 <        Vim editor (vi clone) without GUI\nvim-share-7.2.447 <  Data files for the vim editor (vi clone)\n\n=: package is installed and up-to-date\n<: package is installed but newer version is available\n>: installed package has a greater version than available package\n"
+        "vim-7.2.447;<;Vim editor (vi clone) without GUI\nvim-share-7.2.447;<;Data files for the vim editor (vi clone)\n\n=: package is installed and up-to-date\n<: package is installed but newer version is available\n>: installed package has a greater version than available package\n"
       end
 
       it "returns the version to be installed" do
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/package/pkgin.rb
```diff
@@ -12,8 +12,8 @@
   def self.parse_pkgin_line(package)
 
     # e.g.
-    #   vim-7.2.446 =        Vim editor (vi clone) without GUI
-    match, name, version, status = *package.match(/(\S+)-(\S+)(?: (=|>|<))?\s+.+$/)
+    #   vim-7.2.446;Vim editor (vi clone) without GUI
+    match, name, version, status = *package.match(/(\S+)-(\S+?);(?:(=|>|<))?.+$/)
```
peterhuene: Is this a backwards compatible change?  That is to say, was this always broken or did pkgin's format change?

lib/puppet/provider/package/pkgin.rb
```diff
@@ -30,7 +30,7 @@ def self.prefetch(packages)
   end
 
   def self.instances
-    pkgin(:list).split("\n").map do |package|
+    pkgin("-p", :list).split("\n").map do |package|
```
peterhuene: Likewise, is `-p` backwards compatible?  This wasn't a recent feature added to pkgin, right?

lib/puppet/provider/package/pkgin.rb
jasperla: pkgin's format changed.

lib/puppet/provider/package/pkgin.rb
jasperla: -p was added in 0.7.0 which was released end of last year and is part of pkgsrc 2015Q1. There is no backwards compat for pkgsrc so it's a safe assumption that people using pkgin are running with at least 0.7.0. Otherwise the rest of their software isn't "supported" either.

lib/puppet/provider/package/pkgin.rb
peterhuene: Should we have the regex then support either format so that we don't break users running an older pkgin?  Something like: `/(\S+)-([^\s;=><]+)(?: ?[;=><])?\s*.+$/`?

Also, I don't see a capture group for `status`. Is that always nil?

lib/puppet/provider/package/pkgin.rb
peterhuene: I just meant in terms of the provider being backward compatible with older versions of `pkgin` (i.e. can Puppet users use this pkgin provider with a < 0.7.0 pkgin?).  If that's not the case, then this should be fine.

jasperla: Puppet users of this pkgin provider cannot use it with pkgin < 0.7.0. So perhaps we should remove the explicit `-p` and rely on the implied behaviour on pre and post 0.7.0 versions, while using the new format in the spec tests. Of course with the adjusted regex on line 16.

lib/puppet/provider/package/pkgin.rb
jasperla: It may be best to add some backwards compat for the time being.

As for `status` it's captured as `=|>|<`.

peterhuene: Whoops, you're right.  I keep forgetting that a capture group inside a non-capture group still captures.  That makes sense with regards to status.

Perhaps `/(\S+)-(\S+?)[;\s](=|>|<)?.+$/` then?

lib/puppet/provider/package/pkgin.rb
peterhuene: Makes sense, thanks.

lib/puppet/provider/package/pkgin.rb
jasperla: Not quite, it seems there are some issues with that: http://rubular.com/r/EjqNx1xzWY

peterhuene: `/([^\s\-]+)-([^;\s]+)[;\s](=|>|<)?.+$/`? Regex whack-a-mole! :)

jasperla: Bingo! I'll push the new batch in a few.

jasperla: Too soon, it seems it's breaking on `ruby18-puppet-0.25.5nb1`; the irony...

`$1 == ruby18`
`$2 == puppet-0.25.5nb1`

The hyphen in `ruby18-puppet` shouldn't be a breaking character.

peterhuene: Package names can have hypens? :-1:

lib/puppet/provider/package/pkgin.rb
```diff
@@ -5,15 +5,15 @@
...
 
   has_feature :installable, :uninstallable, :upgradeable, :versionable
 
   def self.parse_pkgin_line(package)
 
     # e.g.
-    #   vim-7.2.446 =        Vim editor (vi clone) without GUI
-    match, name, version, status = *package.match(/(\S+)-(\S+)(?: (=|>|<))?\s+.+$/)
+    #   vim-7.2.446;Vim editor (vi clone) without GUI
+    match, name, version, status = *package.match(/([^\s\-]+)-([^;\s]+)[;\s](=|>|<)?.+$/)
```
peterhuene: So just `/(\S+)-([^\s;]+)[;\s](=|>|<)?.+$/` then?  Should greedily match everything up to the last `-`.

lib/puppet/provider/package/pkgin.rb
jasperla: Oh yes, even on redhat ;-)

I'm starting to wonder if it's actually making sense to add the backwards compat. SmartOS (another consumer of pkgin) has pkgin > 0.7.0 and given the sorry state of providers on NetBSD I doubt many people (if anyone) has used them somewhat recently...

lib/puppet/provider/package/pkgin.rb
jasperla: That breaks when there's a hyphen in the description: `ruby200-facter-2.4.3nb1;Cross-platform Ruby library for retrieving facts from OS`

peterhuene: `/([^\s;]+)-([^\s;]+)[;\s](=|>|<)?.+$/` :sob:

jasperla: But it works! Thanks for your persistence in this.

peterhuene:
@jasperla Thanks again for the contribution!  This looks good to merge.  Is there anything else you'd like to push up or are we all set?

jasperla:
We're all good. I've got nothing pending anymore for this provider. Thanks!

peterhuene:
@jasperla Thanks!  We've merged this into the upcoming Puppet 4.2 release.

PR #3908, opened 2015-05-08
(PUP-735) Disable finalized_on_cycle on Windows

MikaelSmith:
With our Windows MSI package, Puppet isn't installed into normal Ruby
library locations. For puppet/facter/etc we provide wrappers that setup
RUBYLIB, but the same isn't done for ruby/irb. Disable the test on
Windows until PUP-4547 can be fixed.


joshcooper:
No spec changes, merging

PR #3909, opened 2015-05-08
(PUP-4491) Add a proper confine for the openbsd user provider

kylog:
Note that an openbsd-specific user provider was added in
https://github.com/puppetlabs/puppet/commit/986281e1d. However,
that commit neglected to include a confine, so the provider
became suitable for all platforms (although it isn't actually
suitable). Since puppet resolves suitability ties alphabetically,
this meant that providers alphabetically after 'openbsd' would
lose out to this provider.

### DIFF
--- a/lib/puppet/provider/user/openbsd.rb
+++ b/lib/puppet/provider/user/openbsd.rb
@@ -11,6 +11,7 @@ Puppet::Type.type(:user).provide :openbsd, :parent => :useradd do
            :password => "passwd"
 
   defaultfor :operatingsystem => :openbsd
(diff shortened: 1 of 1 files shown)


kylog:
I thought about adding specs ensuring that this is suitable for openbsd and not suitable for other platforms, but we really don't have precedent for such spec tests. So if we do something like that, it makes most sense to me to do it systematically.

Meanwhile, this is the change that should have been part of the original PR and matches the pattern of other platform-specific user providers (like aix, hpux, windows, ...).

peterhuene:
:+1:

PR #3910, opened 2015-05-10
(PUP-4653) stop using hardcoded REST URLs around the filebucket

ffrank:
Network communication with filebuckets used to be implemented by

 1. building a REST URL using logic that duplicated that from
    Puppet::Network::HTTP::API::IndirectedRoutes
 2. deconstructing the URL when handling this special case
    of Puppet::Indirector::Request

This is a relic from API v1 (release 0.25.5) and has not been
necessary for a while, apparently.

We can just stop doing both of those things and the filebucket
Just Works.

### DIFF
--- a/lib/puppet/file_bucket/dipper.rb
+++ b/lib/puppet/file_bucket/dipper.rb
@@ -22,7 +22,7 @@ class Puppet::FileBucket::Dipper
       @rest_path  = nil
     else
       @local_path = nil
-      @rest_path = "https://#{server}:#{port}/#{environment}/file_bucket_file/"
+      @rest_path = "filebucket://#{server}:#{port}/"
     end
     @checksum_type = Puppet[:digest_algorithm].to_sym
     @digest = method(@checksum_type)
--- a/lib/puppet/indirector/request.rb
+++ b/lib/puppet/indirector/request.rb
@@ -243,15 +243,14 @@ class Puppet::Indirector::Request
       @port = uri.port.to_i
     end
 
-    @protocol = uri.scheme
-
-    if uri.scheme == 'puppet'
-      @key = URI.unescape(uri.path.sub(/^\//, ''))
-      return
+    # filebucket:// is only used internally to pass request details
+    # from Dipper objects to the indirector. The wire always uses HTTPS.
+    if uri.scheme == 'filebucket'
+      @protocol = 'https'
+    else
+      @protocol = uri.scheme
     end
 
-    env, indirector, @key = URI.unescape(uri.path.sub(/^\//, '')).split('/',3)
-    @key ||= ''
-    self.environment = env unless env == ''
+    @key = URI.unescape(uri.path.sub(/^\//, ''))
   end
 end
--- a/spec/integration/file_bucket/file_spec.rb
+++ b/spec/integration/file_bucket/file_spec.rb
@@ -18,7 +18,7 @@ describe Puppet::FileBucket::File do
 
       {
         "md5/d41d8cd98f00b204e9800998ecf8427e" => :file,
-        "https://puppetmaster:8140/production/file_bucket_file/md5/d41d8cd98f00b204e9800998ecf8427e" => :file,
+        "filebucket://puppetmaster:8140/md5/d41d8cd98f00b204e9800998ecf8427e" => :file,
       }.each do |key, terminus|
         it "should use the #{terminus} terminus when requesting #{key.inspect}" do
           described_class.indirection.terminus(terminus).class.any_instance.expects(:find)
@@ -31,7 +31,7 @@ describe Puppet::FileBucket::File do
     describe "when running another application" do
       {
         "md5/d41d8cd98f00b204e9800998ecf8427e" => :file,
-        "https://puppetmaster:8140/production/file_bucket_file/md5/d41d8cd98f00b204e9800998ecf8427e" => :rest,
+        "filebucket://puppetmaster:8140/md5/d41d8cd98f00b204e9800998ecf8427e" => :rest,
       }.each do |key, terminus|
         it "should use the #{terminus} terminus when requesting #{key.inspect}" do
           described_class.indirection.terminus(terminus).class.any_instance.expects(:find)
--- a/spec/unit/indirector/request_spec.rb
+++ b/spec/unit/indirector/request_spec.rb
@@ -116,32 +116,32 @@ describe Puppet::Indirector::Request do
       end
 
       it "should set the protocol to the URI scheme" do
-        expect(Puppet::Indirector::Request.new(:ind, :method, "http://host/an_environment", nil).protocol).to eq("http")
+        expect(Puppet::Indirector::Request.new(:ind, :method, "http://host/", nil).protocol).to eq("http")
       end
 
       it "should set the server if a server is provided" do
-        expect(Puppet::Indirector::Request.new(:ind, :method, "http://host/an_environment", nil).server).to eq("host")
+        expect(Puppet::Indirector::Request.new(:ind, :method, "http://host/", nil).server).to eq("host")
       end
 
       it "should set the server and port if both are provided" do
-        expect(Puppet::Indirector::Request.new(:ind, :method, "http://host:543/an_environment", nil).port).to eq(543)
+        expect(Puppet::Indirector::Request.new(:ind, :method, "http://host:543/", nil).port).to eq(543)
       end
 
       it "should default to the masterport if the URI scheme is 'puppet'" do
         Puppet[:masterport] = "321"
-        expect(Puppet::Indirector::Request.new(:ind, :method, "puppet://host/an_environment", nil).port).to eq(321)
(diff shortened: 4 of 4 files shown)


ffrank:
This might be a way to save us some trouble with the patch in #3748 (PUP-1072).

It breaks a number of tests, but I'm not sure if those should just be updated/removed.

I couldn't really get a full picture of what's going on and what the ideas of the existing code were. Feedback is appreciated.

peterhuene:
@ffrank Looks like CI failed (Travis is down at the moment, so I can't figure out the failures there).

lib/puppet/indirector/request.rb
```diff
@@ -250,8 +250,6 @@ def set_uri_key(key)
       return
     end
 
-    env, indirector, @key = URI.unescape(uri.path.sub(/^\//, '')).split('/',3)
-    @key ||= ''
-    self.environment = env unless env == ''
+    @key = key
```
kylog: @joshcooper: you and I talked about this also. One idea was to throw an exception here since IIUC this should only happen if a new URI scheme is used.

lib/puppet/file_bucket/dipper.rb
```diff
@@ -22,7 +22,7 @@ def initialize(hash = {})
       @rest_path  = nil
     else
       @local_path = nil
-      @rest_path = "https://#{server}:#{port}/#{environment}/file_bucket_file/"
+      @rest_path = ""
```
joshcooper: Hi @ffrank, this "works" because the puppet will use the local filebucket instead of the remote one.

With this line changed, when I backup an `empty` file, the request object is:

```
$ bundle exec puppet filebucket backup empty --server hwvft1j1a4531gm.delivery.puppetlabs.net --http_debug
...
(rdb:1) request
#<Puppet::Indirector::Request:0x007fa919a6e770 @instance=nil, @indirection_name=:file_bucket_file, @method=:head, @options={}, @key="md5/d41d8cd98f00b204e9800998ecf8427e//Users/josh/work/puppet/empty">
(rdb:2) request.protocol
nil
```

The default terminus for the file_bucket is the [`:selector`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/file_bucket_file/selector.rb#L7-L13). When the `request.protocol` is nil it selects the `:file` terminus (the local one).

After puppet runs, you'll see the backed up file in the local filebucket, even though I've specified a remote server.

```
$ find /Users/josh/.puppetlabs/opt/puppet/cache/bucket/ -name paths -type f -exec cat {} \; -print
/Users/josh/work/puppet/empty
/Users/josh/.puppetlabs/opt/puppet/cache/bucket//d/4/1/d/8/c/d/9/d41d8cd98f00b204e9800998ecf8427e/paths
```

lib/puppet/indirector/request.rb
joshcooper: Ah, this is actually not dead code. When using a remote file bucket, the uri is of the form:

```
(rdb:2) uri.to_s
"https://hwvft1j1a4531gm.delivery.puppetlabs.net:8140/production/file_bucket_file/md5/d41d8cd98f00b204e9800998ecf8427e//Users/josh/work/puppet/empty"
```

but due to the code above, the request key is rewritten, and later the indirected routes transforms the HTTP request to:

```
$ bundle exec puppet filebucket backup empty --server hwvft1j1a4531gm.delivery.puppetlabs.net --http_debug
...
<- "HEAD /puppet/v3/file_bucket_file/md5/d41d8cd98f00b204e9800998ecf8427e//Users/josh/work/puppet/empty?environment=production& HTTP/1.1
...
```

So this raises some interesting questions:

1. We haven't specified what a valid file_bucket request/response look like in api/schemas. We should do that before making any changes.
2. The Dipper gets the environment (as a string) from `Puppet[:environment]`, includes it in the URI prefix, and the Request strips it out. That is confusing and unnecessary. If an environment is specified in defaults.rb, as a setting, or command-line, then puppet will make that the `Puppet.lookup(:current_environment)`. And since the request will use that environment, there is no reason for the Dipper to specify it when calling the indirected methods, e.g. `Puppet::FileBucket::File.indirection.head(file_bucket_path)`. IOW, the Dipper shouldn't need to specify the enviroment.

lib/puppet/file_bucket/dipper.rb
ffrank: Weird. My smoke tests indicated valid REST interactions. But I may have been coding tired :-)

lib/puppet/indirector/request.rb
ffrank: Good call about specification. Quite the yak behind this corner. Begs the question whether this should keep blocking #3748 (PUP-1072).

joshcooper: For prosperity, adding a note about what I was saying at PR triage. Currently, `Dipper` calls `Puppet::FileBucket::File.indirector.head` specifying a `file_bucket_path` of `https://hwvft1j1a4531gm.delivery.puppetlabs.net:8140/production/file_bucket_file/md5/d41d8cd98f00b204e9800998ecf8427e//Users/josh/work/puppet/empty`, and `Request` has special logic, rewriting the key since it's neither a puppet nor file based scheme.

The more "standard" spooky-action-at-a-distance approach would be for the `filebucket` application to configure the termini based on whether a local or remote file bucket is being used (and `inspect` application too). For example, the newer face-based `file` download & store actions do this already.

Then the Dipper can just call `Puppet::FileBucket::File.indirector.head("md5/d41d8cd...")`, and puppet should route the request to the current terminus. We might then be able to able to move forward with the change to `Request` in this PR, and perhaps simplify/remove `lib/puppet/indirector/file_bucket_file/selector.rb`.

ffrank: Also worth noting that this code is not only used by faces. The agent will use it as well when `file` resources are used with the `filebucket` type.

ffrank: @joshcooper I think I don't follow after all. The `file` face and resources work because files *are* referenced through URLs and there is a [terminus selector](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_serving/terminus_selector.rb) to make sure the indirector does the right thing, correct? Either way, I don't think we can tackle this through application specific configuration, because of the `agent` use case.

As such, I think we might want to hold on to the selector for `FileBucket::File` after all, so that the same code can keep interacting with local and remote buckets.

The most simple approach might be to just switch to a new and arbitrary URL scheme such as `filebucket://` and move the current catch-all in `Indirector::Request` to a special handling for these.

joshcooper:
I was hoping the dipper could be modified to essentially do:

```
    if hash.include?(:Path)
      @local_path = hash[:Path]
      Puppet::FileBucket::File.indirection.terminus_class = :file
    else
      @local_path = nil
      Puppet::FileBucket::File.indirection.terminus_class = :rest
    end
```

and then use the unique filebucket name, e.g. `Puppet::FileBucket::File.indirection.find('md5/...')`. This would provide several benefits:

1. `Puppet::Indirection::Request` doesn't have to know about filebuckets.
1. We could eliminate the Puppet::FileBucket::Selector terminus.
1. We could delete the lines https://github.com/puppetlabs/puppet/pull/3910/files#diff-5bfe8ae5ebd5d25f936221932f87aab5R262, and always raise
1. Filebucket requests could use SRV records. Currently they can't, because we always specify the request's host and port.

But this only works if the filebucket server happens to be the puppetmaster (`Puppet[:server]`). For example, it doesn't work if the `filebucket` resource specifies an alternate `server` parameter:

```
filebucket { 'main':
  server => 'myother.server.com'
}
```

Also doesn't work if the `inspect` application specifies an alternate `Puppet[:archive_file_server]`.

So long story, short, I think this PR makes sense. Dipper needs to continue specifying the server and port in the indirected methods (find, head, save, etc). But I have two comments relating to environments:

1. Puppet no longer specifies `environment` as the URI path prefix, so it's kind of goofy to have the Dipper add it to the prefix, and then have the Request strip it off.

2. Currently the Dipper always assumes `Puppet[:environment]` is the correct environment. But that's not true if `puppet agent` is applying a catalog from an ENC-specified environment and making use of `filebucket` resources. I think instead, we want the application, e.g agent, filebucket, etc to [push the current environment on the context stack](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/application_support.rb#L34-L35) (as it already does), and have the Dipper omit the environment altogether:

```
Puppet::FileBucket::File.indirection.head('filebucket://server:port/md5/...')`
```

The request will then just use the [current environment](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/request.rb#L32).

Also note I've removed the `file_bucket_file` from the URI path as it's repetitive.

peterhuene:
@ffrank Looks like we just need a new ticket and amended commit message for this one.  Thanks!

ffrank:
I kept the branch name because I didn't want to open a new PR.

Also took the liberty to touch up the indirector unit test as mentioned by @joshcooper durin triage.

joshcooper:
Will merge after VCR change goes through

ffrank:
Great. If you can manage this before triage, I shall try and rebase #3748 in time.

PR #3911, opened 2015-05-11
(doc) Tweak how some features are set to fix display in docs

jasperla:
Otherwise the provider support table available at:
http://docs.puppetlabs.com/references/latest/type.html#user shows this
provider as not supporting these features.

### DIFF
--- a/lib/puppet/provider/user/openbsd.rb
+++ b/lib/puppet/provider/user/openbsd.rb
@@ -56,9 +56,7 @@ Puppet::Type.type(:user).provide :openbsd, :parent => :useradd do
 
   has_features :manages_homedir, :manages_expiry, :system_users
   has_features :manages_shell
-  if Puppet.features.libshadow?
-    has_features :manages_passwords, :manages_loginclass
-  end
+  has_features :manages_passwords, :manages_loginclass if Puppet.features.libshadow?
 
   def loginclass=(value)
     set("loginclass", value)


peterhuene:
Hi @jasperla.  Given that this is related to a doc-change, could we change from "maint" to "doc" in the commit message.  Also, I'm not familiar with how these docs get generated; what's the best way for me to verify that this change will fix said table?  Thanks.

jasperla:
Hi @peterhuene , I don't know how these docs are generated. It doesn't seem to be done from one of the rake tasks in the repository. I based it on the fact that `useradd.rb` does it like this with a post-notated conditional and it works out.

kylog:
@nfagerlund can you answer the question above about how the docs for this are generated?

nfagerlund:
Sorry to say, these will have an identical result in docs, and this PR should be closed without merging. These docs are generated by actually loading and evaluating all of the relevant Ruby code, and then iterating through the providers and calling methods on them to discover their features. If the docs are generated on a Mac (they are), libshadow won't be present and it will show those features as unavailable. This was why I filed https://projects.puppetlabs.com/issues/18426, which never really gained any traction. 

Anyway though, we can fix it elsewhere. For the other providers that do this, I had to make an explicit whitelist in the references generator. Which is maintained in the puppet-docs repo, since I abandoned the type reference generator in the Puppet codebase in favor of one that could be tweaked out of band of Puppet releases. 

The relevant code to tweak is here: https://github.com/puppetlabs/puppet-docs/blob/master/lib/puppet_docs/quarantine/get_typedocs.rb#L85 Are these the only provider features that need to get tweaked for OpenBSD stuff, or are there others that aren't displaying for the same reason?

jasperla:
Hi @nfagerlund , thanks for the explanation on this. As far as I'm aware it's just the OpenBSD user provider that is affected. So we'd need to add something like the following?

```ruby
 docobject[:providers][:openbsd][:features] << :manages_passwords << :manages_password_age << :manages_loginclass
```

nfagerlund:
@jasperla Yeah exactly. I can take a PR, or a Jira ticket, and we can try and get the updated references into the puppet 4.1 docs.

jasperla:
@nfagerlund perfect; I'll open a PR tomorrow.

PR #3915, opened 2015-05-11
(PUP-4503,PUP-4540,PUP-2023) Update service provider default for newe…

melissa:
…r debian/ubuntu releases

With the release of Debian 8 Jessie and Ubuntu 15.04 (Vivid Vervet),
ubuntu and debian are noew defaulting the available service manager to
systemd, rather than init or upstart. This commit updates Puppet to
use systemd as the default service provider on these two systems, while
also defaulting back to other service providers if necessary.

### DIFF
--- a/lib/puppet/provider/service/debian.rb
+++ b/lib/puppet/provider/service/debian.rb
@@ -16,7 +16,8 @@ Puppet::Type.type(:service).provide :debian, :parent => :init do
   # is resolved.
   commands :invoke_rc => "/usr/sbin/invoke-rc.d"
 
-  defaultfor :operatingsystem => [:debian, :cumuluslinux]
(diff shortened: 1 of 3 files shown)


kylog:
FYI https://github.com/puppetlabs/ticketmatch/blob/master/ticketmatch.rb chokes on the format "(PROJ-1,PROJ-2)" format. Can this be split into a commit per ticket? Or, if that's too messy, then can it be amended to "(PROJ-1)(PROJ-2)"? Or updating ticketmatch.rb would work too :)

kylog:
Also, should we point this at puppet#stable for 4.1.0?

melissa:
closed in favor of https://github.com/puppetlabs/puppet/pull/3916

PR #3916, opened 2015-05-11
(PUP-4503) Enable systemd on debian jessie and ubuntu vivid

melissa:


melissa:
@kylog this pulls in the work that mine builds off. PUP-1253 is also marked for release in Puppet 4.1, which is why I felt fine about pulling it in as well

sathieu:
@melissa: The test on existence of /run/systemd/system is the canonical check for systemd. Please use it, as some sysadmin may use another init under Debian (sysvinit or upstart).

Once you use this check, you can use a less stric confine for Debian and Ubuntu (which should cover Debian wheezy users with systemd).

kylog:
@melissa okay, sorry for the churn, but I think I was overly optimistic before. Given that we're trying to tag 4.1.0 in the next 2-3 days, and that there are still open issues to suss out here, we should probably target this at 4.2.0 (meaning puppet#master today). 

That said, we can still discuss in this PR and re-target once review settles down.

ahpook:
@kylog @melissa FWIW, I think it'd be bad to release jessie AIOs that don't correctly pick the service provider

melissa:
Closing this and reopening the previous PR against master

melissa:
Needed to open a new pr https://github.com/puppetlabs/puppet/pull/3932

kylog:
Agreed. I'm sad that Jessie didn't make the cut for 4.1.0.

PR #3919, opened 2015-05-12
(PUP-4436) Generate error message in eventlog differently

joshcooper:
Prior to commit 88ae8a9, it was an error to specify a `confdir` whose
parent directory didn't exist. After the change, puppet would create the
ancestor directories `C:/does/not`, enabling puppet's settings catalog
to create the `confdir` named `C:/does/not/exist`.

This commit changes the test to no longer rely on setting an invalid
confdir.  Instead it just runs `puppet agent` trying to connect to a
non-existent master, as it did before, and it changes the expectation to
match the error we generate.

The commit also changes from the deprecated `puppet_agent` beaker
method to `puppet('agent', ...)`.

It would have been nice to test this behavior using:

    puppet apply -e "err('some message')"

However, `apply` uses the console as the default log destination. It is
possible to specify `puppet apply --logdest eventlog ...` but that
doesn't test the default `agent` behavior of always writing to the
eventlog.

### DIFF
--- a/acceptance/tests/windows/eventlog.rb
+++ b/acceptance/tests/windows/eventlog.rb
@@ -12,16 +12,15 @@ agents.each do |agent|
   # get remote time
   now = on(agent, "#{ruby_command(agent)} -e \"puts Time.now.utc.strftime('%m/%d/%Y %H:%M:%S')\"").stdout.chomp
 
-  # it should fail to start since parent directories don't exist
-  confdir = "/does/not/exist"
-
   # generate an error, no master on windows boxes
-  on agent, puppet_agent('--server', '127.0.0.1', '--test', '--confdir', confdir), :acceptable_exit_codes => [1]
+  # we use `agent` because it creates an eventlog log destination by default,
+  # whereas `apply` does not.
+  on agent, puppet('agent', '--server', '127.0.0.1', '--test'), :acceptable_exit_codes => [1]
 
   # make sure there's a Puppet error message in the log
   # cygwin + ssh + wmic hangs trying to read stdin, so echo '' |
   on agent, "cmd /c echo '' | wmic ntevent where \"LogFile='Application' and SourceName='Puppet' and TimeWritten >= '#{now}'\"  get Message,Type /format:csv" do
     fail_test "Event not found in Application event log" unless
-      stdout =~ /Cannot create [a-z]:\/does\/not\/exist.*,Error/mi
+      stdout =~ /Could not retrieve catalog.*skipping run,Error/mi
   end
 end


kylog:
:+1:

acceptance/tests/windows/eventlog.rb
```diff
@@ -12,16 +12,16 @@
...
 
   # make sure there's a Puppet error message in the log
   # cygwin + ssh + wmic hangs trying to read stdin, so echo '' |
   on agent, "cmd /c echo '' | wmic ntevent where \"LogFile='Application' and SourceName='Puppet' and TimeWritten >= '#{now}'\"  get Message,Type /format:csv" do
-    fail_test "Event not found in Application event log" unless
-      stdout =~ /Cannot create [a-z]:\/does\/not\/exist.*,Error/mi
+    if stdout !~ /Could not retrieve catalog.*skipping run,Error/mi
+      fail_test "Event not found in Application event log"
+    end
   end
```
er0ck: do we prefer the stdout !~ //  pattern over the previous fail.. unless?

also, we should probably be specifying the result.stdout to use, otherwise i'm not sure what stdout this is...
``` on agent, "blah" do |result|
  assert_match('blah 2', result.stdout, 'omg')
end
```

joshcooper: I prefer `if something fail_test`, but there are lots of tests that use `fail_test unless`. I'll put that part back.

Historically beaker has supported "global" `stdout`, `stderr` methods that are mixed into the test case. While I too dislike that (and prefer for the `result` to be yielded to the block`), there are many tests that still follow the old convention, and I would rather fix all of them at once at a different time:

```
$ git grep ' stdout' tests
tests/aix/aix_package_provider.rb:    actual_version = stdout.chomp
tests/aix/aix_package_provider.rb:  assert_match(/aix package provider is unable to downgrade packages/, stdout, "Didn't get an error about downgrading packages")
tests/aix/nim_package_provider.rb:    actual_version = stdout.chomp
tests/aix/nim_package_provider.rb:        assert_match(/NIM package provider is unable to downgrade packages/, stdout, "Didn't get an error about downgrading packages")
tests/allow_arbitrary_node_name_fact_for_agent.rb:  node_names << stdout.chomp
tests/allow_arbitrary_node_name_fact_for_agent.rb:    assert_match(/defined 'message'.*#{success_message}/, stdout)
tests/allow_arbitrary_node_name_fact_for_apply.rb:  node_names << stdout.chomp
tests/allow_arbitrary_node_name_fact_for_apply.rb:  assert_match(/defined 'message'.*#{success_message}/, stdout)
tests/allow_arbitrary_node_name_for_agent.rb:    assert_match(/defined 'message'.*#{success_message}/, stdout)
tests/allow_arbitrary_node_name_for_apply.rb:  assert_match(/defined 'message'.*#{success_message}/, stdout)
tests/allow_symlinks_as_config_directories.rb:    assert_match(/My certname is awesome_certname[^\w]/, stdout)
tests/apply/classes/parameterized_classes.rb:    fail_test "inclusion after parameterization failed" unless stdout.include? "1-2"
...
```

joshcooper:
```
$ be rake ci:test:aio SHA=6b1de4c672a620cef7b66f5c556f2865e3b96e98 TESTS=tests/windows/eventlog.rb CONFIG=config/nodes/win2012r2-rubyx64.yaml
...
tests/windows/eventlog.rb passed in 23.44 seconds
      Test Suite: tests @ 2015-05-12 22:10:09 -0700

       Total Suite Time: 23.44 seconds
      Average Test Time: 23.44 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```

PR #3917, opened 2015-05-12
(PUP-4342) test calling all parser functions

er0ck:
This change adds a test for all parser functions.
This test calls all the parser functions that ship with puppet.  It
ensures sane output in most cases.  The remainder of the coverage for
functions lies in the spec tests.

PR #3920, opened 2015-05-12
DO NOT MERGE (PUP-4342) add acceptance to call all parser functions

er0ck:
This change adds a test for all parser functions.
This test calls all the parser functions that ship with puppet. It
ensures sane output in most cases. The remainder of the coverage for
functions lies in the spec tests.

### DIFF
--- a/acceptance/tests/parser_functions/calling_all_functions.rb
+++ b/acceptance/tests/parser_functions/calling_all_functions.rb
@@ -1,0 +1,180 @@
+test_name 'Calling all functions.. test in progress!'
+
+# create list of 3x functions and args
+# notes: hiera functions are well tested elsewhere, included for completeness
+#   special cases: contain (call this from call_em_all)
+#   do fail last because it errors out
+testdir   = ''
+generator = ''
+agents.each do |agent|
+  testdir = agent.tmpdir('calling_all_functions')
+  if agent["platform"] =~ /win/
+    generator = {:args => '"c:/windows/system32/tasklist.exe"', :expected => /\nImage Name/}
+  else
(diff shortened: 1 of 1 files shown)


joshcooper:
@er0ck is this still needed given https://github.com/puppetlabs/puppet/pull/3917 was merged to master? Or is this a backport?

er0ck:
it's a backport.  some folks (including @hlindberg, IIRC) thought it might be important to get into 3.x.
i guess we all forgot about the second PR on the ticket... :-\

joshcooper:
@er0ck @hlindberg do you want to add a ticket to your sprint so this work gets backported?

er0ck:
what work needs to be done to merge?  as far as i can tell it's ready to go.

joshcooper:
@er0ck just that someone needs to own reviewing and merging it, and updating it if there are any 3.x differences that shake out in CI. Also, since this PR is from May, were any other changes made to puppet#master since that time, and this PR for 3.x should be updated accordingly? I'd prefer someone from the language team do this, since you guys did the original work.

hlindberg:
+1 on merging - @er0ck will there be merge problems (merging 3.x -> stable -> master) because of this?

er0ck:
i'll add a ticket to the sprint starting today, in the doing column so i can investigate any merge issues prior to forcing someone to review.  thanks josh

er0ck:
yeah, this will cause issues when merging up to stable/master. we should divide the test in two parts, 3.x only and 4.x only.  4.x only test-file won't live in 3.x branch.  this PR will be for merging to 3.x.  i'll create a separate PR for removing the 3.x portion from stable.  later merges from 3.x to stable will re-enable testing of the 3.x functions in 4+ branches.

PR #3921, opened 2015-05-12
(PUP-1295) Fix always purging package resources that are not installed.

peterhuene:
For package providers, the base package provider was defaulting to a
property hash with a status of :absent when the package was missing.

When the resource is `ensure => purged`, an :absent status causes it to
be purged.  This is because an absent package could represent a
"partially installed" package that is not truly installed and can
therefore be purged (see the dpkg provider).  If the base provider
always returns a status of :absent for missing packages then purgeable
providers will always attempt to purge a missing resource.

The fix is to have the base provider default to a :purged status when
the resource is missing and the provider supports purging.  Otherwise,
it defaults to :absent like before.

### DIFF
--- a/lib/puppet/property/ensure.rb
+++ b/lib/puppet/property/ensure.rb
@@ -49,7 +49,7 @@ class Puppet::Property::Ensure < Puppet::Property
 
   def change_to_s(currentvalue, newvalue)
     begin
-      if currentvalue == :absent or currentvalue.nil?
+      if currentvalue == :absent || currentvalue.nil?
         return "created"
       elsif newvalue == :absent
         return "removed"
--- a/spec/integration/type/package_spec.rb
+++ b/spec/integration/type/package_spec.rb
@@ -125,3 +125,68 @@ describe Puppet::Type.type(:package), "when packages with the same name are sour
   end
 end
 
+describe Puppet::Type.type(:package), 'logging package state transitions' do
+
+  let(:catalog) { Puppet::Resource::Catalog.new }
+  let(:provider) { stub('provider', :class => Puppet::Type.type(:package).defaultprovider, :clear => nil, :validate_source => nil) }
+
+  before :each do
+    provider.stubs(:satisfies?).with([:purgeable]).returns(true)
+    provider.class.stubs(:instances).returns([])
+    provider.stubs(:install).returns nil
+    provider.stubs(:uninstall).returns nil
+    provider.stubs(:purge).returns nil
+    Puppet::Type.type(:package).defaultprovider.stubs(:new).returns(provider)
+  end
+
+  after :each do
+    Puppet::Type.type(:package).defaultprovider = nil
+  end
+
+  # Map of old state -> {new state -> change}
+  states = {
+    # 'installed' transitions to 'removed' or 'purged'
+    :installed => {
+      :installed => nil,
+      :absent    => 'removed',
+      :purged    => 'purged'
+    },
+    # 'absent' transitions to 'created' or 'purged'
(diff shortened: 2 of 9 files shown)


peterhuene:
Review note: CI failure is from spurious rdoc failure on Windows and can be ignored.

MikaelSmith:
The notices are a little funny now.
When installing: `Notice: /Stage[main]/Main/Package[sudo]/ensure: ensure changed 'purged' to 'present'`
When purging: `Notice: /Stage[main]/Main/Package[sudo]/ensure: ensure changed '1.8.6p7-13.el7' to 'purged'`
When absenting: `Notice: /Stage[main]/Main/Package[sudo]/ensure: removed`

MikaelSmith:
This seems like it's worth running through some acceptance tests; have you done so? I'll kick off running  all the acceptance tests that manage packages.

peterhuene:
To fix:

* First notice: transitioning from purged to present should be treated the same as absent (i.e. `created`), so will fix that to say `created`.

* Second notice: transitioning from present to purged should be treated as present to absent, but used `purged` instead of `removed`.

Does that make sense to do?

MikaelSmith:
Yeah, those make sense.

MikaelSmith:
The acceptance coverage is pretty sparse.

MikaelSmith:
It looks like this may impact the acceptance test: [tests/resource/package/common_package_name_in_different_providers.rb](https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/resource/package/common_package_name_in_different_providers.rb)

peterhuene:
Nice catch, I'll look through the acceptance tests and make sure there aren't any others.

MikaelSmith:
I've tried out scenarios with `gem`, `dpkg`, `apt`, `yum`, and `apple` providers. No more observations, so just the previously mentioned cleanup should be sufficient.

lib/puppet/property/ensure.rb
```diff
@@ -49,7 +49,7 @@ def self.inherited(sub)
 
   def change_to_s(currentvalue, newvalue)
     begin
-      if currentvalue == :absent or currentvalue.nil?
```
MikaelSmith: :facepunch:

MikaelSmith:
:+1:

MikaelSmith:
Can you add integration tests for the reports produced?

peterhuene:
Will do.

peterhuene:
@MikaelSmith Test added.

peterhuene:
@MikaelSmith Looks like the new test failed travis (runs locally :-1: ).  Looking into the failure.

MikaelSmith:
Appears to still fail.

peterhuene:
@MikaelSmith Stubbed the instance `instances` instead of the class method.  Should work now.

spec/integration/type/package_spec.rb
```diff
@@ -125,3 +125,68 @@ def provider_name(os)
...
+
+  states.each do |old, new_states|
+    describe "#{old} package" do
+
+      before :each do
+        provider.stubs(:properties).returns(:ensure => old)
+      end
+
+      new_states.each do |new, status|
+        it "ensure => #{new} should log #{status ? status : 'nothing'}" do
```
MikaelSmith: Using a keyword name makes highlighting funny, but I'll let it pass.

peterhuene: `new` isn't a keyword in ruby :/

MikaelSmith: Hmm, github fail then.

PR #3925, opened 2015-05-12
(PUP-4601) Add ruby version requirement to .gemspec.

peterhuene:
This will prevent gem upgrade on environments with ruby 1.8.7


peterhuene:
Triage note: original contributor has indicated that this meets the trivial patch exemption (change to packaging) and does not require a CLA signature.

PR #3930, opened 2015-05-13
(docs) Minor updates to API docs

nfagerlund:
Hey @dankreek, a note about this showed up in an otherwise unrelated docs ticket. Could you verify that this is true, and Puppet Server CA doesn't support the pluralized version of this? 

When this PR gets merged, please resolve https://tickets.puppetlabs.com/browse/DOC-864!

### DIFF
--- a/api/docs/http_api_index.md
+++ b/api/docs/http_api_index.md
@@ -11,28 +11,39 @@ V1/V2 HTTP APIs (Removed)
 ---------------
 
 The V1 and V2 APIs were removed in Puppet 4.0.0. The routes that were previously
-under `/` or `/v2.0` can now be found under the [master V3](#puppet-v3-http-api)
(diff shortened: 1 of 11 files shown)


dankreek:
That is correct, this endpoint no longer works.

nfagerlund:
Thanks!

joshcooper:
@dankreek by "no longer works" I assume that means "no longer supported by design" (and should be documented as such), as opposed to "no longer works due to a bug and will be fixed"?

dankreek:
That is correct, @joshcooper. Sorry for the poor wording.

nfagerlund:
Ping. I'm pretty sure the test failures here are not a problem with these commits. :)

joshcooper:
Will merge once CI clears

PR #3931, opened 2015-05-13
(PUP-4602) Add rcng service provider

jasperla:
Here's a new provider for service functionality through rcng, while it shares some traits with the `bsd` provider it differs enough to warrant a new one due to the different paths and content of files that need to be enabled.

### DIFF
--- a/lib/puppet/provider/service/bsd.rb
+++ b/lib/puppet/provider/service/bsd.rb
@@ -5,7 +5,7 @@ Puppet::Type.type(:service).provide :bsd, :parent => :init do
     Uses `rc.conf.d` for service enabling and disabling.
   EOT
 
-  confine :operatingsystem => [:freebsd, :netbsd, :dragonfly]
+  confine :operatingsystem => [:freebsd, :dragonfly]
 
   def rcconf_dir
     '/etc/rc.conf.d'
--- a/lib/puppet/provider/service/rcng.rb
+++ b/lib/puppet/provider/service/rcng.rb
@@ -1,0 +1,51 @@
+Puppet::Type.type(:service).provide :rcng, :parent => :bsd do
+  desc <<-EOT
+    RCng service management with rc.d
+  EOT
+
+  defaultfor :operatingsystem => [:netbsd, :cargos]
+  confine :operatingsystem => [:netbsd, :cargos]
+
+  def self.defpath
+    "/etc/rc.d"
+  end
+
+  # if the service file exists in rc.conf.d AND matches an expected pattern
+  # then it's already enabled
+  def enabled?
+    rcfile = File.join(rcconf_dir, @resource[:name])
+    if Puppet::FileSystem.exist?(rcfile)
+      File.open(rcfile).readlines.each do |line|
+        # Now look for something that looks like "service=${service:=YES}" or "service=YES"
+        if line.match(/^\s*#{@resource[:name]}=(?:YES|\${#{@resource[:name]}:=YES})/)
+          return :true
+        end
+      end
+    end
+
+    :false
+  end
+
+  # enable service by creating a service file under rc.conf.d with the
+  # proper contents, or by modifying it's contents to to enable the service.
+  def enable
+    Dir.mkdir(rcconf_dir) if not Puppet::FileSystem.exist?(rcconf_dir)
+    rcfile = File.join(rcconf_dir, @resource[:name])
+    if Puppet::FileSystem.exist?(rcfile)
+      newcontents = []
+      File.open(rcfile).readlines.each do |line|
+        if line.match(/^\s*#{@resource[:name]}=(NO|\$\{#{@resource[:name]}:NO\})/)
+          line = "#{@resource[:name]}=${#{@resource[:name]}:=YES}"
+        end
+        newcontents.push(line)
+      end
+      Puppet::Util.replace_file(rcfile, 0644) do |f|
+        f.puts newcontents
+      end
+    else
+      Puppet::Util.replace_file(rcfile, 0644) do |f|
+        f.puts "%s=${%s:=YES}\n" % [@resource[:name], @resource[:name]]
+      end
+    end
+  end
+end
--- a/spec/unit/provider/service/rcng_spec.rb
+++ b/spec/unit/provider/service/rcng_spec.rb
@@ -1,0 +1,41 @@
+#!/usr/bin/env ruby
+
+require 'spec_helper'
+
+provider_class = Puppet::Type.type(:service).provider(:rcng)
+
+describe provider_class, :unless => Puppet.features.microsoft_windows? do
+  before :each do
+    Puppet::Type.type(:service).stubs(:defaultprovider).returns described_class
+    Facter.stubs(:value).with(:operatingsystem).returns :netbsd
+    Facter.stubs(:value).with(:osfamily).returns 'NetBSD'
+    described_class.stubs(:defpath).returns('/etc/rc.d')
+    @provider = provider_class.new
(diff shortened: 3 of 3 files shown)


lib/puppet/provider/service/rcng.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  # if the service file exists in rc.conf.d AND matches an expected pattern
+  # then it's already enabled
+  def enabled?
+    rcfile = File.join(rcconf_dir, @resource[:name])
+    if Puppet::FileSystem.exist?(rcfile)
+      File.open(rcfile).readlines.each do |line|
+        # Now look for something that looks like "service=${service:=YES}" or "service=YES"
+        if matchdata = line.match(/^#{@resource[:name]}=(?:YES|\${#{@resource[:name]}:=YES})/)
```
peterhuene: Is there guaranteed to be no leading whitespace on the line (i.e. use of `^` here)?

Also, `matchdata` appears to be unused.

lib/puppet/provider/service/rcng.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  # enable service by creating a service file under rc.conf.d with the
+  # proper contents, or by modifying it's contents to to enable the service.
+  def enable
+    Dir.mkdir(rcconf_dir) if not Puppet::FileSystem.exist?(rcconf_dir)
+    rcfile = File.join(rcconf_dir, @resource[:name])
+    if Puppet::FileSystem.exist?(rcfile)
+      newcontents = []
+      File.open(rcfile).readlines.each do |line|
+        if matchdata = line.match(/^#{@resource[:name]}=(NO|\$\{#{@resource[:name]}:NO\})/)
```
peterhuene: Same comment as above, along with `matchdata` being unused.

peterhuene:
Hi @jasperla.  Thanks for the contribution.  Besides the minor comments above, this looks good to me.

Note to reviewers: AppVeyor failure is the result of the spurious rdoc spec.

lib/puppet/provider/service/rcng.rb
jasperla: As verified by putting junk into the rc file, it makes no difference to rcng if there is leading whitespace. So I've adjusted the regex in the provider too.

lib/puppet/provider/service/rcng.rb
jasperla: That's fixed too now.

lib/puppet/provider/service/rcng.rb
```diff
@@ -0,0 +1,48 @@
...
+    rcfile = File.join(rcconf_dir, @resource[:name])
+    if Puppet::FileSystem.exist?(rcfile)
+      newcontents = []
+      File.open(rcfile).readlines.each do |line|
+        if line.match(/^\s*#{@resource[:name]}=(NO|\$\{#{@resource[:name]}:NO\})/)
+          line = "#{@resource[:name]}=${#{@resource[:name]}:=YES}"
+        end
+        newcontents.push(line)
+      end
+      File.open(rcfile, 'w') { |file| file.puts newcontents }
```
joshcooper: Could you change this and the one below to use `Puppet::Util.replace_file`? The reason is that if puppet crashes (or is Ctrl-C) while writing the file, it will leave it in a corrupt state. The `replace_file` helper writes to a temp file (in the same directory) and does an atomic rename.

lib/puppet/provider/service/rcng.rb
```diff
@@ -0,0 +1,48 @@
...
+    else
+      File.open(rcfile, File::WRONLY | File::CREAT, 0644) { |f|
+        f << "%s=${%s:=YES}\n" % [@resource[:name], @resource[:name]]
+      }
```
joshcooper: Same comment here

peterhuene:
Hi @jasperla.  Have you had a chance to review @joshcooper's feedback?  Thanks!

jasperla:
Hi, I've been too busy recently to handle this. I should be able to incorporate the feedback later this week or latest early next week.

jasperla:
Hi @peterhuene , @joshcooper 's feedback has been incorporated now.

lib/puppet/provider/service/rcng.rb
```diff
@@ -0,0 +1,50 @@
+Puppet::Type.type(:service).provide :rcng, :parent => :bsd do
+  desc <<-EOT
+    RCng service management with rc.d
+  EOT
+
+  defaultfor :operatingsystem => [:netbsd, :cargos]
```
joshcooper: note to self, need to review and make sure the confine of the parent applies to the `rcng` provider.

lib/puppet/provider/service/bsd.rb
```diff
@@ -5,7 +5,7 @@
     Uses `rc.conf.d` for service enabling and disabling.
   EOT
 
-  confine :operatingsystem => [:freebsd, :netbsd, :dragonfly]
+  confine :operatingsystem => [:freebsd, :dragonfly]
```
joshcooper: note to self, need to confine to `netbsd` somewhere

lib/puppet/provider/service/rcng.rb
peterhuene: Should we confine for `:netbsd` or `:cargos` here as well?

lib/puppet/provider/service/rcng.rb
```diff
@@ -0,0 +1,50 @@
...
+    rcfile = File.join(rcconf_dir, @resource[:name])
+    if Puppet::FileSystem.exist?(rcfile)
+      newcontents = []
+      File.open(rcfile).readlines.each do |line|
+        if line.match(/^\s*#{@resource[:name]}=(NO|\$\{#{@resource[:name]}:NO\})/)
+          line = "#{@resource[:name]}=${#{@resource[:name]}:=YES}"
+        end
+        newcontents.push(line)
+      end
+      Puppet::Util.replace_file(rcfile, 0644) do |f|
```
peterhuene: Given that this is similar to the base provider's permission for the rc file, could we perhaps have a base method that returns the mode for the file and make use of it from the  derived provider?  That way the mode of `0644` is only defined in one place.

lib/puppet/provider/service/rcng.rb
joshcooper: @jasperla could you update this PR to `confine` to `netbsd` and `cargos`, otherwise, this provider may be suitable, and chosen on other platforms.

joshcooper: ping @jasperla, wondering if you've had a chance to look at this?

jasperla: i'm currently unavailable, but ill apple the suggested fixes when im back. 
-- 
jasper

hkenney:
@peterhuene Could you take another look at this now that it's updated?

peterhuene:
@jasperla This looks good, other than my above comment above about the mode '0644' being specified in multiple places and is perhaps a candidate for extracting into a `self.default_mode` into the base BSD service provider method like `lib/puppet/provider/sshkey/parsed.rb` has.  Thoughts on extracting that out or are we happy with having the mode specified in multiple places?

jasperla:
@peterhuene I'm fine with having it specified in multiple locations; it feels like a piece of data too trivial for it's own method.

peterhuene:
Good enough for me then. :+1:

peterhuene:
@jasperla Thanks again for the contribution!  This has been merged to master and will appear in Puppet 4.3.0.

PR #3932, opened 2015-05-13
(PUP-4503)(PUP-2023) Update service provider default for newer debian…

melissa:
…/ubuntu releases

With the release of Debian 8 (Jessie) and Ubuntu 15.04 (Vivid Vervet),
ubuntu and debian now default to systemd as the main service manager,
rather than init or upstart. This commit updates Puppet to
use systemd as the default service provider on these two systems, while
also defaulting back to other service providers if necessary.

Additionally, there is a bug on debian systemd where it is unable to
correctly detect SysV style services[1]. This commit also adds in a
workaround to correctly detect `is-enabled` for these services.

[1] - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=751638

### DIFF
--- a/lib/puppet/provider/service/debian.rb
+++ b/lib/puppet/provider/service/debian.rb
@@ -16,7 +16,8 @@ Puppet::Type.type(:service).provide :debian, :parent => :init do
   # is resolved.
   commands :invoke_rc => "/usr/sbin/invoke-rc.d"
 
-  defaultfor :operatingsystem => [:debian, :cumuluslinux]
+  defaultfor :operatingsystem => :cumuluslinux
+  defaultfor :operatingsystem => :debian, :operatingsystemmajrelease => ['5','6','7']
 
   # Remove the symlinks
   def disable
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -5,10 +5,14 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
 
   commands :systemctl => "systemctl"
 
+  confine :exists => "/run/systemd/system"
+
   defaultfor :osfamily => [:archlinux]
   defaultfor :osfamily => :redhat, :operatingsystemmajrelease => "7"
   defaultfor :osfamily => :redhat, :operatingsystem => :fedora, :operatingsystemmajrelease => ["17", "18", "19", "20", "21"]
   defaultfor :osfamily => :suse, :operatingsystemmajrelease => ["12", "13"]
+  defaultfor :operatingsystem => :debian, :operatingsystemmajrelease => "8"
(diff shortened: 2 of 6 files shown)


melissa:
AppVeyor failed on the rdoc test, a known issue

lib/puppet/provider/service/debian.rb
```diff
@@ -16,7 +16,8 @@
   # is resolved.
   commands :invoke_rc => "/usr/sbin/invoke-rc.d"
```
sathieu: Why not add "confine :false => Puppet::FileSystem.exist?("/run/systemd/system")" here?

melissa: If a user wants to continue managing a service with this debian flavor of init, they would be unable to. Especially because of the bug in systemd on jessie when managing services with sysv profile (see PUP-4503 for more details), it would be disadvantageous to users to completely prevent them from using this provider.

sathieu: OK.

peterhuene:
I've verified that we now default to systemd for Jessie and start, stop, and report on services correctly.

peterhuene:
@melissa Hmm.  If I run `puppet resource service` I get systemd service entries and the init services together in the output (e.g. for cron I get `cron` and `cron.service`).  Is this desired?

haus:
@peterhuene As i understand it, `puppet resource $thing` will go through all of the suitable providers for $thing on $platform and report on all of them. I may be mistaken.

haus:
Additionally, systemd has more than just .service files, so simply stripping .service for systemd instances won't fly either.

peterhuene:
Right, wasn't advocating removing the `.server` from the names.  Just wasn't expecting seeing the init.d scripts in there (my distro uses systemd, but has for years and thus has no init.d scripts anymore).

peterhuene:
I'm :+1:

sathieu:
Any chance to have this in the 3.x branch?

j-vizcaino:
Just backported this fix to a Puppet 3.7.5 installation: it introduces another bug.
The line 62 from `systemd.rb` needs to be updated from

```ruby
if Facter.value(:osfamily) == 'debian'
```

to
 
```ruby
if Facter.value(:osfamily) == 'Debian'
```

otherwise Puppet thinks it has to enable the service everytime it runs...
After fixing this, our services status seem to be detected correctly.

Thanks

kylog:
@j-vizcaino I think this came up in #4092 as well. Do you mind taking a look at that PR?

j-vizcaino:
@kylog: actually that's where I got the hint, as @raphink applied the exact same fix in his PR. When looking for problem regarding Puppet services on Debian Jessie, you end up here. Just wanted to let readers know that applying this fix was not sufficient.

PR #3933, opened 2015-05-13
(PUP-4607) Fix loading external facts

whopper:
In Facter 2.x, upon the first Facter.add, the directory loaders
are created and cannot be changed.

PUP-4359 added the puppetversion fact before setting the external
fact directories, meaning that no other external facts were being
found by Puppet in puppet specific external fact directories.

This commit causes Puppet to first set its external fact directories,
and then add the puppetversion fact.


whopper:
@kylog good call on the comment - added.

PR #3934, opened 2015-05-14
(PUP-2573) Make agent locker raise LockError if we fail to aquire lock

dalen:
This fixes a possible raise condition in the puppet agent where it first
checked if the lock was taken and then later on aquired it, possibly
having some other agent taking the lock in between with no handling of
that situation.

Also it updates the description of the lock method to actually match
what it does, namely return the value of the block itself. Not
true/false based on if lock was acquired. That was in fact impossible to
find out previously.

### DIFF
--- a/lib/puppet/agent.rb
+++ b/lib/puppet/agent.rb
@@ -1,4 +1,5 @@
 require 'puppet/application'
+require 'puppet/error'
 
 # A general class for triggering a run of another
 # class.
@@ -29,10 +30,6 @@ class Puppet::Agent
 
   # Perform a run with our client.
   def run(client_options = {})
-    if running?
-      Puppet.notice "Run of #{client_class} already in progress; skipping  (#{lockfile_path} exists)"
-      return
-    end
     if disabled?
       Puppet.notice "Skipping run of #{client_class}; administratively disabled (Reason: '#{disable_message}');\nUse 'puppet agent --enable' to re-enable."
       return
@@ -46,6 +43,9 @@ class Puppet::Agent
           begin
             client_args = client_options.merge(:pluginsync => Puppet[:pluginsync])
             lock { client.run(client_args) }
+          rescue Puppet::LockError
+            Puppet.notice "Run of #{client_class} already in progress; skipping  (#{lockfile_path} exists)"
+            return
           rescue StandardError => detail
             Puppet.log_exception(detail, "Could not run #{client_class}: #{detail}")
           end
--- a/spec/unit/agent_spec.rb
+++ b/spec/unit/agent_spec.rb
@@ -58,7 +58,6 @@ describe Puppet::Agent do
 
     client.expects(:run)
 
-    @agent.stubs(:running?).returns false
     @agent.stubs(:disabled?).returns false
     @agent.run
   end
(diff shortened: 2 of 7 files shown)


peterhuene:
Hi @dalen.  Thanks for the contribution!  Nice catch on the race condition.

This looks good to me.  I checked any other places we might be calling `Agent::Locker#lock` and not expecting the new exception being raised and couldn't find any.  We'll get some more eyes on this in the meantime before merging it in.

lib/puppet/agent.rb
```diff
@@ -29,10 +30,6 @@ def needing_restart?
 
   # Perform a run with our client.
   def run(client_options = {})
-    if running?
-      Puppet.notice "Run of #{client_class} already in progress; skipping  (#{lockfile_path} exists)"
-      return
-    end
```
joshcooper: I think we can also remove the call to [`agent.running?`](https://github.com/puppetlabs/puppet/blob/4.0.0/lib/puppet/daemon.rb#L90-L93), and instead have it do:

```
def reload
  return unless agent
  agent.run({:splay => false})
rescue Puppet::LockError
  Puppet.notice "Not triggering already-running agent"
end
```

I'm thinking we should also deprecate the [`Puppet::Agent::Locker#locked?`](https://github.com/puppetlabs/puppet/blob/4.0.0/lib/puppet/util/lockfile.rb#L45-L48) method as it is inherently unsafe. The only safe way to know if the lock is locked and perform some action based on the result, is to try to lock it, and then handle the LockError that may result.

Puppet::Util::Lockfile has similar problems, but I think there's already a ticket about it containing race conditions.

joshcooper:
minor nit, typo in commit subject, should be "acquire"

joshcooper:
@richardc will raising an exception break mcollective's puppet-agent plugin? I didn't see the method referenced in https://github.com/puppetlabs/mcollective-puppet-agent/blob/master/util/puppet_agent_mgr/mgr_v3.rb so I think it's ok, but wanted to get your thoughts.

richardc:
@joshcooper though the plugin loads the puppet library, it triggers runs by sending signals or invoking the cli, so these internal changes shouldn't impact the mcollective agent's operation.

spec/unit/agent_spec.rb
```diff
@@ -85,12 +83,6 @@ def controlled_run(&block)
       @agent.run
     end
 
-    it "should do nothing if already running" do
-      @agent.expects(:running?).returns true
-      AgentTestClient.expects(:new).never
-      @agent.run
-    end
-
```
joshcooper: Should we check if the lock raises `Puppet::LockError` then we expect `client.expects(:run).never` and a notice to be logged, e.g. `Puppet.expects(:notice).with(regexp_matches(/not sure what goes here../))`

peterhuene:
@dalen Looks like this will be good to merge once Josh's latest feedback is addressed.  Thanks!

peterhuene:
Hi @dalen.  Just checking in to see if you saw @joshcooper's feedback.  Thanks!

peterhuene:
Hi @dalen.  Just checking in again from the PR triage.  We're hoping to merge this after we unfreeze (we have an upcoming release).  Thanks!

spec/unit/agent_spec.rb
dalen: I didn't manage to get such a test working, it seemed like the AgentTestClient was stubbing a bit too much to actually test it. But I felt the code is covered by enough other tests.

But adding a similar one might be good, I likely won't have time for it short term though.

dalen:
Fixed the deprecation of Puppet::Agent::Locker.running? and the call to that from the daemon reload method.

joshcooper:
Thanks @dalen! We're getting ready to release 4.2 and are holding off on PRs until after the release. Once that happens we'll get this merged. Thanks again!

peterhuene:
@dalen Thanks again for the contribution.  This has been merged into master and should appear in Puppet 4.3.

PR #3935, opened 2015-05-14
(maint) Update service to work on debian

melissa:
Unlike redhat, debian stores env files in '/etc/default'. We need to be
able to search there in order to access them. This commit adds that
search path to the systemd service file. Note that this works because
the service doesn't depend on the existance of those files. So, on
redhat, neither '/etc/default/puppet' nor '/etc/default/puppetagent'
will exist, but the service will continue to run regardless.


ext/systemd/puppet.service
```diff
@@ -6,6 +6,8 @@ After=basic.target network.target puppetmaster.service
 [Service]
 EnvironmentFile=-/etc/sysconfig/puppetagent
 EnvironmentFile=-/etc/sysconfig/puppet
+EnvironmentFile=-/etc/default/puppetagent
```
haus: The puppetagent variations are for compatibility with older fedora packages. It shouldn't be needed for debian.

ext/systemd/puppetmaster.service
```diff
@@ -5,7 +5,8 @@ After=basic.target network.target
 
 [Service]
 EnvironmentFile=-/etc/sysconfig/puppetmaster
-ExecStart=/usr/bin/puppet master $PUPPETMASTER_EXTRA_OPTS --no-daemonize
+EnvironmentFile=-/etc/default/puppetmaster
+ExecStart=/opt/puppetlabs/puppet/bin/puppet master $PUPPETMASTER_EXTRA_OPTS --no-daemonize
```
melissa: So, I noticed we also have the puppetmaster.service file, which is referenced in the specfile in the puppet repo, as well as the puppet.service file, but no where in puppet-agent. I'm not sure if we even need this file, but it should probably be updated according to the new layout

haus:
:+1:

ext/systemd/puppet.service
```diff
@@ -6,6 +6,7 @@ After=basic.target network.target puppetmaster.service
 [Service]
 EnvironmentFile=-/etc/sysconfig/puppetagent
 EnvironmentFile=-/etc/sysconfig/puppet
+EnvironmentFile=-/etc/default/puppet
```
shrug: I feel like we should either template this, or have a separate debian file... but that's kind of terrible also

melissa: I'd prefer to keep it like this. Although it would make sense if we were dealing with more paths that just `/etc/default` and `/etc/sysconfig`, I'd agree with you, but at the moment, the extra logic doesn't seem worth it.

That, and templating this with vanagon is painful (according to @haus)

haus: We could template it during build using sed (that's what works right now), but that leaves the file basically useless to anyone else because we'd need to use some sort of token in place of the environment file.

ext/systemd/puppetmaster.service
stahnma: Seems like we should just kill the puppetmaster service file all together. It's not really a thing anymore.

PR #3937, opened 2015-05-14
(PUP-4552) Merge from 3.x

thallgren:
This commit contains the merge of PUT-4552 from 3.x to stable.
In addition to what's covered in that issue, it also copies a lot
of tests from the future_compiler_spec (which doesn't exist in
stable) to compiler_spec.

### DIFF
--- a/lib/puppet/pops/evaluator/access_operator.rb
+++ b/lib/puppet/pops/evaluator/access_operator.rb
@@ -251,12 +251,11 @@ class Puppet::Pops::Evaluator::AccessOperator
   def access_POptionalType(o, scope, keys)
     keys.flatten!
     if keys.size == 1
-      unless keys[0].is_a?(Puppet::Pops::Types::PAnyType)
-        fail(Puppet::Pops::Issues::BAD_TYPE_SLICE_TYPE, @semantic.keys[0], {:base_type => 'Optional-Type', :actual => keys[0].class})
+      type = keys[0]
+      unless type.is_a?(Puppet::Pops::Types::PAnyType) || type.is_a?(String)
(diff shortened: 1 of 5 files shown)


kylog:
@thallgren re the AppVeyor failure, there are 3 spec failures.

The two rdoc ones can be ignored -- see https://tickets.puppetlabs.com/browse/PUP-4537.

The remaining one however may be legit, snipped here for easier parsing:
```
 1) Puppet::Parser::Compiler the compiler when using future parser and evaluator should be able to determine the configuration version from a local version control repository
     Failure/Error: pending("Bug #14071 about semantics of Puppet::Util::Execute on Windows") do
     ArgumentError:
       The semantics of `RSpec::Core::Pending#pending` have changed in
       RSpec 3. In RSpec 2.x, it caused the example to be skipped. In
       RSpec 3, the rest of the example is still run but is expected to
       fail, and will be marked as a failure (rather than as pending) if
       the example passes.
       
       Passing a block within an example is now deprecated. Marking the
       example as pending provides the same behavior in RSpec 3 which was
       provided only by the block in RSpec 2.x.
       
       Move the code in the block provided to `pending` into the rest of
       the example body.
       
       Called from C:/projects/puppet/spec/integration/parser/compiler_spec.rb:1035:in `block (3 levels) in <top (required)>'.
     # ./spec/integration/parser/compiler_spec.rb:1035:in `block (3 levels) in <top (required)>'
 
```

thallgren:
Thanks @kylog. I fixed and pushed. I saw the two rdoc features too and concluded that the file in question wasn't affected by this PR so that was a bit strange. Good to know that they can be ignored.

PR #3938, opened 2015-05-15
(docs) Show more useful defaults for special directories in configuration reference

nfagerlund:
The configuration reference was showing blank defaults for codedir, rundir, and
logdir, and it was showing root Unix paths only for confdir and vardir. Since
we're overriding these with simple strings when we generate the config reference
anyway, we can just make those strings more useful.

### DIFF
--- a/lib/puppet/reference/configuration.rb
+++ b/lib/puppet/reference/configuration.rb
@@ -22,10 +22,16 @@ config = Puppet::Util::Reference.newreference(:configuration, :depth => 1, :doc 
 
     # Now print the data about the item.
     val = object.default
-    if name.to_s == "vardir"
-      val = "/opt/puppetlabs/puppet/cache"
-    elsif name.to_s == "confdir"
-      val = "/etc/puppetlabs/puppet"
+    if name.to_s == 'vardir'
+      val = 'Unix/Linux: /opt/puppetlabs/puppet/cache -- Windows: C:\ProgramData\PuppetLabs\puppet\cache -- Non-root user: ~/.puppetlabs/opt/puppet/cache'
+    elsif name.to_s == 'confdir'
+      val = 'Unix/Linux: /etc/puppetlabs/puppet -- Windows: C:\ProgramData\PuppetLabs\puppet\etc -- Non-root user: ~/.puppetlabs/etc/puppet'
+    elsif name.to_s == 'codedir'
+      val = 'Unix/Linux: /etc/puppetlabs/code -- Windows: C:\ProgramData\PuppetLabs\code -- Non-root user: ~/.puppetlabs/etc/code'
+    elsif name.to_s == 'rundir'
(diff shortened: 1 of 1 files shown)


nfagerlund:
oh whoops, that string should definitely not be double quoted. Hold on.

nfagerlund:
There we go.

lib/puppet/reference/configuration.rb
```diff
@@ -22,10 +22,16 @@
 
     # Now print the data about the item.
     val = object.default
-    if name.to_s == "vardir"
-      val = "/opt/puppetlabs/puppet/cache"
-    elsif name.to_s == "confdir"
-      val = "/etc/puppetlabs/puppet"
+    if name.to_s == 'vardir'
+      val = 'Unix/Linux: /opt/puppetlabs/puppet/cache -- Windows: C:\ProgramData\PuppetLabs\puppet\cache -- Non-root user: ~/.puppetlabs/opt/puppet/cache'
```
joshcooper: Windows 2003 uses a different prefix, but I don't know that I care. /cc @Iristyle

nfagerlund: Since it's formally deprecated now, I'm inclined to leave it behind. We have it in the main docs about these directories; that's good enough.

Iristyle: One thing we could do is use `%ALLUSERSPROFILE%` in place of `C:\ProgramData` everywhere.  I think most Windows users understand that just fine.

nfagerlund: I've been avoiding doing that, on the assumption that most Windows users DON'T understand that just fine... but that assumption might have been bad. I would kind of want to check with support or something, unless the whole windows team agrees on this?

joshcooper: :+1: to %ALLUSERSPROFILE%, Windows users will know that is

joshcooper: Technically, it's `%ALLUSERSPROFILE%\Puppet Labs\puppet\cache` on Vista and up, but `%ALLUSERSPROFILE%\Application Data\Puppet Labs\puppet\cache` in 2003, but only 5 more days, so not too worried about that.

nfagerlund: So... changing it to %ALLUSERSPROFILE% wouldn't fix the concern about win 2003 in the first place, right? And we're still fine with leaving it behind in this particular place, right? 

Should we change it to %ALLUSERSPROFILE% regardless? I consider that more opaque, but I don't have any idea how many windows installs have that directory in a place other than C:\ProgramData; if it's a bunch, then sure?

hkenney:
@Iristyle Could you please take a look at this so we can get it into 4.2.1?

kylog:
Probably not making it into 4.2.1 per se because we're about to tag.

lib/puppet/reference/configuration.rb
MikaelSmith: @joshcooper @Iristyle this discussion stalled out, any more comments?

PR #3940, opened 2015-05-15
(docs) Revise description of trusted_server_facts setting

nfagerlund:


joshcooper:
After merging, please regenerate man pages, eg https://github.com/puppetlabs/puppet/pull/3947

kylog:
@hkenney or @hlindberg can you review?

hkenney:
:+1:  Looks good! One thought:

I'm not sure whether or not we need to explicitly make the distinction between what changes in an agent run vs apply. With an apply only the environment part of the hash will be present. That may be too low level for the user to care about, not sure.

kylog:
@nfagerlund thoughts on ^ ??

nfagerlund:
@hkenney @kylog I think the right place for that level of detail is probably in the language reference where we talk about these pre-populated variables, not in the list of settings. I'm cool with leaving the setting description pretty high-level.

hkenney:
That seem's totally reasonable to me. I'm :+1: on merging this as is.

ferventcoder:
Note that Travis CI ran a build but appveyor did not. :)

ferventcoder:
This means the AppVeyor filters are working appropriately.

PR #3941, opened 2015-05-16
(QENG-1453) Make use of Beaker's epel_url option

justinstoller:
Previously we were depending on the "use-el-extras" feature of Beaker
but we were not taking full advantage of its ability to point at an EPEL
mirror.  We have sporadically seen failures based on the availability of
the mirrors at kernel.org.

This commit infers if we are running within the internal network by
looking for a well known environment variable and, if so, adds the
appropriate beaker configuration to use our internal epel mirror.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -9,9 +9,13 @@ ONE_DAY_IN_SECS = 24 * 60 * 60
 REPO_CONFIGS_DIR = "repo-configs"
 CLEAN.include('*.tar', REPO_CONFIGS_DIR, 'merged_options.rb')
 
+# If elsewhere we're depending on internal network resources
+# then assume we can depend on them here
+EPEL_MIRROR = ENV['BEAKER_EPEL_MIRROR']
+
 module HarnessOptions
 
-  DEFAULTS = {
+  defaults = {
     :type => 'git',
     :helper => ['lib/helper.rb'],
     :tests  => ['tests'],
@@ -27,8 +31,10 @@ module HarnessOptions
     :add_el_extras => true,
     :preserve_hosts => 'onfail',
     :forge_host => 'forge-aio01-petest.puppetlabs.com',
-    :'master-start-curl-retries' => 30,
+    :'master-start-curl-retries' => 30
   }
+
+  DEFAULTS = EPEL_MIRROR ? defaults.merge(:epel_url => EPEL_MIRROR) : defaults
 
   class Aggregator
     attr_reader :mode


justinstoller:
I'll try to get this tested locally today.

kylog:
Nit: can you preface the commit line with "(maint)"? The release script looks in that leading element of the commit summary for a PUP- ticket or maint or docs.

acceptance/Rakefile
```diff
@@ -9,9 +9,13 @@ ONE_DAY_IN_SECS = 24 * 60 * 60
 REPO_CONFIGS_DIR = "repo-configs"
 CLEAN.include('*.tar', REPO_CONFIGS_DIR, 'merged_options.rb')
 
+# If elsewhere we're depending on internal network resources
+# then assume we can depend on them here
+INTERNAL_NETWORK = ENV['GEM_SOURCE'] && ENV['GEM_SOURCE'] =~ /puppetlabs/
```
kylog: My $.02: I'd prefer an environment variable per mirror. This solution makes me worry that 6 months from now someone will tweak `GEM_SOURCE` and then we'll all be scratching our heads about why all the other mirrors got turned off.

acceptance/Rakefile
```diff
@@ -26,10 +30,16 @@ module HarnessOptions
...
     :add_el_extras => true,
     :preserve_hosts => 'onfail',
-    :forge_host => 'forge-aio01-petest.puppetlabs.com',
-    :'master-start-curl-retries' => 30,
+    :'master-start-curl-retries' => 30
+  }
+
+  site_specific_info = {
+    :epel_url   => 'http://osmirror.delivery.puppetlabs.net/fedora-epel',
+    :forge_host => 'forge-aio01-petest.puppetlabs.com'
```
kylog: `forge_host` could be controlled by its own env var also.

justinstoller: The forge host is controlled by the env var `BEAKER_FORGE_HOST` with the above as a default. 

Looking at it, I realize it's that way because there's no valid value for it outside of our network.

Since the forge host is needed for site specific tests and there's a Beaker env var already, on second thought I don't actually think I should be changing around the forge_host configuration.

acceptance/Rakefile
justinstoller: Updated with `BEAKER_EPEL_MIRROR`

That follows the `BEAKER_*` env var scheme, which may be good or bad depending on whether you want to have all configuration for Beaker stored in `BEAKER_` or you want to differentiate env vars that are consumed by the rake task vs those consumed directly via Beaker (if so, perhaps a `PUPPET_*` pattern?).

justinstoller:
Tested locally with an example test case
```
16:06:13 hobo:acceptance justin (maint/3.x/QENG-1453_use_internal_epel_mirror) ⭑ bundle exec rake ci:test:packages
./config/packages/options.rb
./local_options.rb
No options file found at /Users/justin/Backup/code/work/puppet/acceptance/local_options.rb
beaker --options-file merged_options.rb --hosts=config/nodes/rhel6.yaml --tests=tests/ticket_4233_resource_with_a_newline.rb

....

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) 16:06:32$  rpm -qa | grep epel-release

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) executed in 0.70 seconds
Exited: 1

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) 16:06:33$  rpm -i http://osmirror.delivery.puppetlabs.net/fedora-epel/6/i386/epel-release-6-8.noarch.rpm
warning: /var/tmp/rpm-tmp.az0FbC: Header V3 RSA/SHA256 Signature, key ID 0608b895: NOKEY

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) executed in 0.41 seconds

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) 16:06:33$  sed -i -e 's;#baseurl.*$;baseurl=http://osmirror\.delivery\.puppetlabs\.net/fedora\-epel/6/$basearch;' /etc/yum.repos.d/epel.repo

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) executed in 0.03 seconds

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) 16:06:33$  sed -i -e '/mirrorlist/d' /etc/yum.repos.d/epel.repo

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) executed in 0.02 seconds

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) 16:06:33$  yum clean all && yum makecache
Loaded plugins: product-id, subscription-manager
This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
Cleaning repos: epel redhat-6-base redhat-6-optional redhat-6-updates
Cleaning up Everything
Loaded plugins: product-id, subscription-manager
This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
Metadata Cache Created

gn04dhhx1sm71lo.delivery.puppetlabs.net (master) executed in 22.48 seconds
....
```

acceptance/Rakefile
```diff
@@ -9,9 +9,13 @@ ONE_DAY_IN_SECS = 24 * 60 * 60
 REPO_CONFIGS_DIR = "repo-configs"
 CLEAN.include('*.tar', REPO_CONFIGS_DIR, 'merged_options.rb')
 
+# If elsewhere we're depending on internal network resources
+# then assume we can depend on them here
```
kylog: This comment dates to the previous implementation but doesn't make sense anymore.

acceptance/Rakefile
```diff
@@ -27,9 +31,11 @@ module HarnessOptions
     :add_el_extras => true,
     :preserve_hosts => 'onfail',
     :forge_host => 'forge-aio01-petest.puppetlabs.com',
-    :'master-start-curl-retries' => 30,
+    :'master-start-curl-retries' => 30
```
kylog: FWIW I'm fond of the trailing comma on the last item in a multi-line construct (because of shorter diffs when extending the construct).

But regardless, I'd prefer to keep this commit's diff just the functional changes, and then if we want to do a style scrub for trailing commas across the code base that could be a separate PR (ideally rubocop-backed so that it would  be enforced).

PR #3942, opened 2015-05-18
(PUP-2630) Populate $server_facts during puppet apply

hkenney:
Prior to this commit, $server_facts would only be populated during
an agent run. This it made it difficult for users to use
`$server_facts[environment]` in their puppet code since they could
not test it via a puppet apply.

Now that variable will be populated during an apply if
trusted_server_facts is true. It will only contain the environment
portion of the hash.


hkenney:
Re-running AppVeyor, the failures didn't look like they should be related to this change (something about hostnames)

hkenney:
AppVeyor passed the second time! :+1:

kylog:
Makes sense. :+1:

PR #3944, opened 2015-05-18
(PUP-4193) Revert earlier addition of lambda case option

hlindberg:
Before this, it was (for a while) possible to use a lambda as an
advanced case option where case value was splat-applied to a lambda and
the result of the lambda being used as the outcome of option match.

It was decided that this should undergo the Puppet RFC process instead
of being directly added to the language.

This commit reverts the previously added (and unreleased) functionality.


kylog:
The appveyor failure can be ignored -- it's a known transient in the rdoc specs.

PR #3945, opened 2015-05-18
(PUP-4538) Use source permissions when using plugin application

joshcooper:
Puppet has a `plugin` face application whose `download` action will perform a pluginsync. However, it did not `use` source permissions when downloading external facts. As a result, executable external facts could not be executed on the agent. This is similar to PUP-4420, but is a different code path that triggers the same bad behavior.

This commit modifies the `download` action to use the same code paths that the configurer uses to perform pluginsyncing of external facts and modules. It also adds spec and acceptance tests.

### DIFF
--- a/acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
+++ b/acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
@@ -78,16 +78,28 @@ master_opts = {
 with_puppet_running_on master, master_opts, codedir do
   agents.each do |agent|
     factsd   = agent.tmpdir('facts.d')
+    pluginfactdest = agent.tmpdir('facts.d')
     tmpdir   = agent.tmpdir('tmpdir')
     testfile = File.join(tmpdir, 'testfile')
 
     teardown do
       on(master, "rm -rf #{codedir}")
       on(agent, "rm -rf #{factsd}")
+      on(agent, "rm -rf #{pluginfactdest}")
     end
 
     step "Pluginsync the external fact to the agent and ensure it resolves correctly"
-    on(agent, puppet('agent', "-t --server #{master} --pluginfactdest #{factsd}"), :acceptable_exit_codes => [2])
+    on(agent, puppet('agent', '-t', '--server', master, '--pluginfactdest', factsd), :acceptable_exit_codes => [2])
+    assert_match(/foo is bar/, stdout)
+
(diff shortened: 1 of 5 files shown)


joshcooper:
Hold off on merging until after 4.1.0 is released, and target for 4.1.1

joshcooper:
appveyor failures are due to https://tickets.puppetlabs.com/browse/PUP-4537

joshcooper:
rebased to pick up appveyor change to archive spec_order.txt failures.

spec/unit/face/plugin_spec.rb
```diff
@@ -2,9 +2,38 @@
...
+  let(:pluginface) { described_class }
+  let(:action) { pluginface.get_action(:download) }
+
+  def render(result)
+    action.when_rendering(:console).call(result)
+  end
+
+  context "download" do
+    before :each do
+      Puppet.features.stubs(:external_facts?).returns(true)
```
joshcooper: I lots @MikaelSmith comments on an earlier commit, see https://github.com/joshcooper/puppet/commit/54746e7#commitcomment-11389424

I'm proposing we don't add tests for when external_facts? is false, because it never is in practice, and I think we should remove the conditional logic. But do that in a different ticket/PR

acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
```diff
@@ -78,16 +78,28 @@
...
     teardown do
       on(master, "rm -rf #{codedir}")
       on(agent, "rm -rf #{factsd}")
+      on(agent, "rm -rf #{pluginfactdest}")
     end
 
     step "Pluginsync the external fact to the agent and ensure it resolves correctly"
-    on(agent, puppet('agent', "-t --server #{master} --pluginfactdest #{factsd}"), :acceptable_exit_codes => [2])
+    on(agent, puppet('agent', '-t', '--server', master, '--pluginfactdest', factsd), :acceptable_exit_codes => [2])
+    assert_match(/foo is bar/, stdout)
```
joshcooper: I lost @MikaelSmith comments on an earlier commit, see https://github.com/joshcooper/puppet/commit/54746e7#commitcomment-11389335

spec/unit/face/plugin_spec.rb
MikaelSmith: Sounds appropriate.

PR #3950, opened 2015-05-19
fix-spec-warnings-part-deux

kylog:


hlindberg:
Hm, that is a real problem. The constant becomes a constant in the function instantiator! That is unintended. This fixes a symptom - the real problem is that functions cannot use constants and this will cause hard to find bugs when users later try to have different versions.

hlindberg:
I have updated the 4x function API spec with the rule that a 4x function may not introduce a constant. (It already specified that it is illegal to introduce nested classes and modules; for the same reason - the Ruby compiler must know them a priori to creating the anonymous class representing the function.

hlindberg:
also merged to stable and master

kylog:
@hlindberg I wasn't clear from your two comments above whether you concluded that "functions cannot use constants" is a bug or not?

hlindberg:
It is not a bug - it is a constraint that users have to understand - if anything it was a bug in the spec and my blocg post - it appears to work, but it creates code that will overwrite existing constants in the class that defines the function instead of in the anonymous class representing the function when doing a load - constants (just like subclasses and sub modules which were already specified as illegal) should simply not be used. (I did try to come up with a way to make this work; but it is actually impossible when you think about it - you cannot have something that is both constant and variable at the same time. If a function wants to use a constant, it must be defined in code that is constant (e.g. you can use Float::PI). The closest thing to a constant is to set an instance variable in the function (that is allowed and works). An instance variable value in a function remains the same for the life of the function instance (i.e. as long as the environment is cached).

Using constants for types (as in the define) is however tricky since the logic in the dispatcher is not running as a method on the function class. Therefore, any reuse of a type specification string in the dispatcher logic is to use a local variable.

PR #3951, opened 2015-05-19
(PUP-4431) Fix getpid usage in base service provider 

kylog:


lib/puppet/provider/service/base.rb
```diff
@@ -31,8 +31,9 @@ def getps
   def getpid
     @resource.fail "Either stop/status commands or a pattern must be specified" unless @resource[:pattern]
     regex = Regexp.new(@resource[:pattern])
+    ps = getps
     self.debug "Executing '#{ps}'"
-    IO.popen(getps) { |table|
+    IO.popen(ps) { |table|
```
peterhuene: It's so easy to miss string interpolation references :/  For a second I saw this and thought why was this change necessary.

PR #3952, opened 2015-05-19
(PUP-4629) Return values instead of augeas paths during the match operation

mmarod:
This fixes PUP-4629 by using aug.get to return the values of the paths returned by the aug.match operation.

### DIFF
--- a/lib/puppet/provider/augeas/augeas.rb
+++ b/lib/puppet/provider/augeas/augeas.rb
@@ -44,6 +44,7 @@ Puppet::Type.type(:augeas).provide(:augeas) do
     "rename" => [ :path, :string ],
     "insert" => [ :string, :string, :path ],
     "get" => [ :path, :comparator, :string ],
+    "values" => [ :path, :glob ],
     "defvar" => [ :string, :path ],
     "defnode" => [ :string, :path, :string ],
     "match" => [ :path, :glob ],
@@ -243,6 +244,54 @@ Puppet::Type.type(:augeas).provide(:augeas) do
     !!return_value
   end
 
+  # Used by the need_to_run? method to process values filters. Returns
+  # true if there is a matched value, false if otherwise
+  def process_values(cmd_array)
+    return_value = false
+
+    #validate and tear apart the command
+    fail("Invalid command: #{cmd_array.join(" ")}") if cmd_array.length < 3
+    cmd = cmd_array.shift
+    path = cmd_array.shift
+
+    # Need to break apart the clause
+    clause_array = parse_commands(cmd_array.shift)[0]
+    verb = clause_array.shift
+
+    #Get the match paths from augeas
+    result = @aug.match(path) || []
+    fail("Error trying to get path '#{path}'") if (result == -1)
+
+    #Get the values of the match paths from augeas
+    values = result.collect{|r| @aug.get(r)}
+
+    case verb
+    when "include"
+      arg = clause_array.shift
+      return_value = values.include?(arg)
+    when "not_include"
+      arg = clause_array.shift
+      return_value = !values.include?(arg)
+    when "=="
+      begin
+        arg = clause_array.shift
+        new_array = eval arg
+        return_value = (values == new_array)
+      rescue
+        fail("Invalid array in command: #{cmd_array.join(" ")}")
+      end
+    when "!="
+      begin
+        arg = clause_array.shift
+        new_array = eval arg
+        return_value = (values != new_array)
+      rescue
+        fail("Invalid array in command: #{cmd_array.join(" ")}")
+      end
+    end
+    !!return_value
+  end
+
   # Used by the need_to_run? method to process match filters. Returns
   # true if there is a match, false if otherwise
   def process_match(cmd_array)
@@ -367,6 +416,7 @@ Puppet::Type.type(:augeas).provide(:augeas) do
         begin
           case command
           when "get"; return_value = process_get(cmd_array)
+          when "values"; return_value = process_values(cmd_array)
           when "match"; return_value = process_match(cmd_array)
           end
         rescue StandardError => e
--- a/spec/unit/provider/augeas/augeas_spec.rb
+++ b/spec/unit/provider/augeas/augeas_spec.rb
@@ -195,6 +195,56 @@ describe provider_class do
     end
   end
 
+  describe "values filters" do
+    before do
+      augeas = stub("augeas", :match => ["set", "of", "values"])
+      augeas.stubs(:get).returns('set').then.returns('of').then.returns('values')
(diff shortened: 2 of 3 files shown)


peterhuene:
Hi @mmarod.  Thanks for the contribution! 

@lutter and @domcleal: as you're both more knowledgeable about the augeas provider than I am, does this change make sense to either of you? 

Note to reviewers: the AppVeyor failure is the known rdoc failure and can be ignored.

domcleal:
I can see where you're going with this, but changing the behaviour of `match` isn't a good idea as it could easily break existing manifests.

I would suggest adding a `get` command with this behaviour instead (although in Augeas' API, get won't work for multiple nodes, but having it do so here might be OK).

domcleal:
> I would suggest adding a get command with this behaviour instead

Oh, we do have it, apologies.

Maybe call it something else then if you'd like support for returning an array of values, e.g. `getm` (to emulate `setm`, `clearm` etc.)

mmarod:
Adding a `getm` seems reasonable but it seems like it might be unnecessary. It boils down to what the intent of the `match <MATCH_PATH> == <AN_ARRAY>` is really meant to do. Is it supposed to compare labels or values?

I would argue that the `match <MATCH_PATH> != <AN_ARRAY>` and `match <MATCH_PATH> == <AN_ARRAY>` controls in `onlyif` do not currently do what the Puppet type reference says they should do. `AN_ARRAY` is described as `['a string, 'another']` which could never actually be returned when calling `aug.match`. It would instead return `['/files/something/some_param[1]', '/files/something/some_param[2]']` which is not useful to match against.

If we are trying to make the controls in `onlyif` mimic exacty what augeas does, then we should add a `getm` parameter. However, it seems to me (according to the type ref) that `match <MATCH_PATH> == <AN_ARRAY>` was meant to solve this problem in the first place.

domcleal:
I'd say that's a documentation bug and that changing the provider behaviour is a) going to be incompatible, so would have to live in master for some time, and b) would differ from Augeas.

I think mimicking Augeas' API is quite important, as many users will test how Augeas responds via augtool, so to change the return value from a standard API call or command is surprising.

This is a very useful feature, particularly so the `==` and `include` operators work more naturally, but I don't think changing an existing command is the way to do it.

mmarod:
Fair enough

mmarod:
If we are trying to mimic augeas commands exactly -- how should getm be implemented? There does not seem to be a built in getm command. It seems running a `match` and then passing each result into a `get` is the only way to retrieve all of the values in an augeas path.

domcleal:
Yes, that'd be the way to do it.

peterhuene:
Hi @mmarod Thanks for pushing up a new commit!  Would you mind squashing the two commits into one (basically amending the original commit)?

spec/unit/provider/augeas/augeas_spec.rb
```diff
@@ -195,6 +195,36 @@
...
+    before do
+      augeas = stub("augeas", :match => ["set", "of", "values"])
+      augeas.stubs(:get).returns('set').then.returns('of').then.returns('values')
+      augeas.stubs("close")
+      @provider = provider_class.new(@resource)
+      @provider.aug = augeas
+    end
+
+    it "should return true for an array match" do
+      command = ["match", "fake value", "== ['set', 'of', 'values']"]
```
domcleal: It doesn't affect the code, but for clarity, change `match` to `getm`.

domcleal:
Looks good to me.  Would you also mind adding the `include` and `not_include` operators to getm while we're here?  I think they're likely to be useful.

domcleal:
:+1: looks good, and tests well.

domcleal:
Oh, the only thing I'd add is that the commit message doesn't quite reflect what it does any more, since the `match` behaviour's been left unchanged.  Would you mind updating that?

lutter:
I am a little late to the party, but how about calling that new operator `values` ? That seems to express pretty well what it's supposed to do, and there won't be any way to confuse it with existing augtool commands.

mmarod:
`values` seems a bit less confusing to me. Since `getm` isn't a true augeas API call anyway it shouldn't make a difference whether we call it `getm` or `values`. What do you think @domcleal ?

domcleal:
Good suggestion, works for me!

peterhuene:
@domcleal and @lutter: could either one of you review @mmarod's changes one last time to make sure this looks good (as I'm no augeas provider expert) before we merge this?  Thanks!

domcleal:
Looks fine to me, although the commit message is incorrect again, s/getm/values/.  Could be fixed on merge.

mmarod:
Fixed the commit message again

domcleal:
:+1: thanks @mmarod!

lib/puppet/provider/augeas/augeas.rb
```diff
@@ -243,6 +244,57 @@ def process_get(cmd_array)
...
+    verb = clause_array.shift
+
+    #Get the match paths from augeas
+    result = @aug.match(path) || []
+    fail("Error trying to get path '#{path}'") if (result == -1)
+
+    #Get the values of the match paths from augeas
+    values = []
+    result.each do |r|
+      values.push(@aug.get(r))
```
ffrank: Could this loop be replaced by an invocation of `Array#map` aka `#collect`?

mmarod: Good idea... I updated it

joshcooper:
The appveyor failures are a known issue https://tickets.puppetlabs.com/browse/PUP-4537 and can be ignored.

mmarod:
So it's been a while since this is seemingly ready to go -- What is the hold up?

joshcooper:
@mmarod We've had to cancel the last three community PR triage sessions while preparing for the Puppet 4.2.0 release. We'll resume next 10am PDT, Wednesday, June 1, so it should be merged soon.

joshcooper:
@mmarod one minor nit, could you update the commit message to describe how `values` can be used as a result of this change? A real-world example would be really useful. Thank you!

joshcooper:
@mmarod thanks for updating that, the example is very helpful. Could you shorten the subject line to follow our convention as described in https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md:

```
(PUP-1234) Make the example in CONTRIBUTING imperative and concrete

Without this patch applied the example commit message in the CONTRIBUTING
document is not a concrete example.
```

I'm happy to fix that up if you're busy. Thank you again!

mmarod:
No problem -- Let me know how it looks.

joshcooper:
Thank you @mmarod for your contribution! This will be released in puppet 4.3.0.

PR #3953, opened 2015-05-19
(PUP-4619) sort fstab entries

ffrank:
Does not yet work, but will solve an annoying fstab effect.

### DIFF
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -272,6 +272,48 @@ Puppet::Type.type(:mount).provide(
     instances
   end
 
+  # Sort fstab entries so that
+  # a mount point is located inside another mount point appears later and
+  # a device that is located inside a mount point appears later.
+  #
+  # "Stable opportunistic insertion sort"
+  # Copy unsorted list A to sorted B by repeatedly
+  # * removing the first element from A
+  # * compare with each element in B, from first to last
+  # * insert before current element in B if smaller
+  # * insert at the end of B otherwise
+  # * finish if A is empty
+  # * repeat
+  #
+  # Stabilizes itself by inserting as late as possible.
+  #
+  # @note Does not really care about directory names and does substring
+  #   comparison only. So /media will be put before /mediastuff. This is
+  #   not necessary but will be rare and usually cause no harm, either.
+  #
+  # @api private
+  def self.order(records)
+    return records if records.empty?
+    # insert first record first
+    result = [ ]
+    # iterate over the rest
+    records.each do |a|
+      inserted = false
+      result.each_index do |i|
+        b = result[i]
+        # is a a leading substring of b?
+        if b[:name].index(a[:name]) == 0
+          result.insert(i, a)
+          inserted = true
+          break
+        end
+      end
+      next if inserted
+      result.push(a)
+    end
+    result
+  end
+
   def flush
     needs_mount = @property_hash.delete(:needs_mount)
     super
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -195,6 +195,10 @@ Puppet::Type.type(:mount).provide(
     end
   end
 
+  # Class variable to store the catalog's choice about whether
+  # fstab entries should be sorted.
+  @sort_output = false
+
   # Every entry in fstab is :unmounted until we can prove different
   def self.prefetch_hook(target_records)
     target_records.collect do |record|
@@ -239,6 +243,15 @@ Puppet::Type.type(:mount).provide(
           mount.provider.set(:ensure => :mounted)
         end
       end
+    end
+    # Examine the catalog to determine whether the fstab should be ordered.
+    # Default is no sorting
+    @sort_output = false
+    return if !resources || resources.empty?
+    meta = resources.values[0].catalog.resource('resources', 'mount')
+    return unless meta
+    if meta[:sort_output]
+      @sort_output = true
     end
   end
 
@@ -298,6 +311,7 @@ Puppet::Type.type(:mount).provide(
   # @api private
   def self.order(records)
     return records if records.empty?
+    return records unless @sort_output;
     # insert first record first
     result = [ ]
     # iterate over the rest
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -287,6 +287,10 @@ Puppet::Type.type(:mount).provide(
   #
   # Stabilizes itself by inserting as late as possible.
   #
+  # A record is "smaller" than another if
+  # * the mount point of A is a leading substring of B's mount point
+  # * the mount point of A is a leading substring of B's device (B is a bind mount)
+  #
   # @note Does not really care about directory names and does substring
   #   comparison only. So /media will be put before /mediastuff. This is
   #   not necessary but will be rare and usually cause no harm, either.
@@ -302,7 +306,7 @@ Puppet::Type.type(:mount).provide(
       result.each_index do |i|
         b = result[i]
         # is a a leading substring of b?
-        if b[:name].index(a[:name]) == 0
+        if b[:name].index(a[:name]) == 0 || b[:device].index(a[:name])
           result.insert(i, a)
           inserted = true
           break
--- a/lib/puppet/provider/parsedfile.rb
+++ b/lib/puppet/provider/parsedfile.rb
@@ -391,6 +391,10 @@ class Puppet::Provider::ParsedFile < Puppet::Provider
     header + text
   end
 
+  def self.order(records)
+    return records
+  end
+
   def create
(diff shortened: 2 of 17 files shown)


ffrank:
Fixed up with a custom sorting algorithm. Seems to work now.

Completed tests and updated all commit messages.

lib/puppet/provider/mount/parsed.rb
```diff
@@ -272,6 +272,49 @@ def self.mountinstances
...
+  # A record is "smaller" than another if
+  # * the mount point of A is a leading substring of B's mount point
+  # * the mount point of A is a leading substring of B's device (B is a bind mount)
+  #
+  # @note Does not really care about directory names and does substring
+  #   comparison only. So /media will be put before /mediastuff. This is
+  #   not necessary but will be rare and usually cause no harm, either.
+  #
+  # @api private
+  def self.order(records)
```
peterhuene: Rather than an O(n^2) sort with a nested catch/throw (I would imagine that the Ruby VM as it needs to setup an exception handler in the frame, which is an expensive operation on some platforms), would this work?

```ruby
def self.order(records)
  # Create an array of indexes into the given records array
  indexes = [*0..(records.size - 1)]
  indexes.sort! do |a, b|
    # First sort by mount point
    result = records[a][:name] <=> records[b][:name]
    # If b's mountpoint should come before a's, sort by device
    result = records[a][:device] <=> records[b][:device] if result >= 0
    # If equal, sort by index to create a stable sort
    result = a <=> b if result == 0
    result
  end.map! do |index|
    # Replace the index with the original entry
    records[index]
  end
end
```

ffrank: Great point about complexity, but at first glance, this looks to me as though it's basically a clever trick to stabilize Ruby's own sort. That's helpful, but we're still running the risk of critical comparisons never taking place.

Yes, this is a total order, but only because of the arbitrary "keep original order" ruling. It still does not address the issue that for the primary comparator, most elements will be equal.

In other words, if the quicksort happens to choose convenient pivot elements, this will work, but not in the average case. Unless I'm wrong :-) I'll pass it by the existing tests in any case.

Thanks for looking into this @peterhuene !

peterhuene: Yeah, this doesn't look correct results :/

peterhuene: My original proposal wasn't correct.  I've updated the above with what I think may work, but I haven't tested it yet.

peterhuene: Hmm, the above still isn't quite right. At any rate, I'm not terribly concerned about the O(N^2) runtime.  Could we at least do away with the catch/throw for an inserted flag?

ffrank: As I feared: Passes some tests, but does break one of them.

Agree that fstabs are usually not large. Even if, this is an "at most once" kind of deal. The agent will do this only after all `mount` resources are synced and **only** if any of them changed.

I wasn't aware of the catch/throw penalty. If elegance costs performance, I will absolutely change to a flag as you suggested.

ffrank: ...pushed that as an unsquashed commit for posterity. Will tidy up before triage :)

peterhuene: Yeah, the above simply won't function as a topological sort.  The insertion sort will so :+1:

lib/puppet/provider/mount/parsed.rb
```diff
@@ -272,6 +272,51 @@ def self.mountinstances
...
+    # insert first record first
+    result = [ records[0] ]
+    # iterate over the rest
+    records.drop(1).each do |a|
+      short_circuited = false
```
peterhuene: Perhaps `inserted` to make it a little clearer?  The break is what's technically short circuiting the inner most loop, the flag is just preventing it from being inserted again.

lib/puppet/provider/mount/parsed.rb
```diff
@@ -272,6 +272,51 @@ def self.mountinstances
...
+  # @note Does not really care about directory names and does substring
+  #   comparison only. So /media will be put before /mediastuff. This is
+  #   not necessary but will be rare and usually cause no harm, either.
+  #
+  # @api private
+  def self.order(records)
+    # insert first record first
+    result = [ records[0] ]
+    # iterate over the rest
+    records.drop(1).each do |a|
```
peterhuene: Nitpick: `drop(1)` will allocate a new array.  Can we avoid doing so just for skipping the first element in the enumeration?

Insertion sort could also easily be done in-place rather than creating a resulting array as well.  We probably don't want `records` mutated (haven't looked into what we do with the array after writing out the entries), so no big deal.

lib/puppet/provider/mount/parsed.rb
```diff
@@ -272,6 +272,51 @@ def self.mountinstances
...
+  # * the mount point of A is a leading substring of B's device (B is a bind mount)
+  #
+  # @note Does not really care about directory names and does substring
+  #   comparison only. So /media will be put before /mediastuff. This is
+  #   not necessary but will be rare and usually cause no harm, either.
+  #
+  # @api private
+  def self.order(records)
+    # insert first record first
+    result = [ records[0] ]
```
peterhuene: Can `records` be empty?

lib/puppet/provider/parsedfile.rb
```diff
@@ -391,6 +391,10 @@ def self.to_file(records)
     header + text
   end
 
+  def self.order(records)
+    return records
```
peterhuene: Could we comment why the `order` call is a no-op here?

lib/puppet/provider/mount/parsed.rb
ffrank: Good call - this is what broke some tests after the previous push.

lib/puppet/provider/mount/parsed.rb
ffrank: Oh my - the skip wasn't even necessary. Just got rid of it.

I can't quite picture the in-place variant. I'll just hope that the copy approach has a readability advantage ;-)

ffrank:
Well, what do you know. Making sorting optional was [easier](https://github.com/ffrank/puppet/commit/9c216083f2f88558bd99473c71bcdb5ab9e4ae1c) than I feared.

I don't *like* burdening the `resources` type further, but it might be the cleanest approach after all.

Perhaps @Sharpie has more thoughts on that?

Sharpie:
@ffrank Hmm, that's a tricky one. I would expect a type-level default:

```puppet
Mount {
  sort_output => true,
}
```

Over a default coming from the `resources` meta-type:

```puppet
resource { "mount":
  sort_output => true,
}
```

The former is visible in the catalog as an attribute of a given `Mount` resource. The latter introduces some action at a distance.

Sharpie:
And, of course, as soon as I wrote that I realize that ordering isn a property that makes sense for `Mount` at the individual resource level --- it has to be applied at the whole file level by the provider.

In that case, I can't think of a better place than the `resources` metatype.

joshcooper:
Discuss current state at PR triage tomorrow

joshcooper:
PR Triage Notes

* @joshcooper to add to next sprint

MikaelSmith:
AppVeyor failure is rdoc, can ignore.

joshcooper:
@ffrank sorry this is taking so long. We're currently reviewing PR #3748 and will try to bring this into a future sprint.

lib/puppet/provider/mount/parsed.rb
```diff
@@ -272,6 +285,53 @@ def self.mountinstances
...
+    return records unless @sort_output;
+    # insert first record first
+    result = [ ]
+    # iterate over the rest
+    records.each do |a|
+      inserted = false
+      result.each_index do |i|
+        b = result[i]
+        # is a a leading substring of b?
+        if b[:name].index(a[:name]) == 0 || b[:device].index(a[:name])
```
MikaelSmith: Based on the comment above, shouldn't it be `b[:device].index(a[:name]) == 0`?

lib/puppet/provider/mount/parsed.rb
```diff
@@ -240,6 +244,15 @@ def self.prefetch(resources = nil)
         end
       end
     end
+    # Examine the catalog to determine whether the fstab should be ordered.
+    # Default is no sorting
+    @sort_output = false
```
MikaelSmith: This seems odd, but I don't see any consistent patterns about using the resources metatype.

MikaelSmith:
I think sorting needs a fix for the device name, otherwise it'll match any substring and not just the leading substring. When you do, can you also rebase on master?

MosesMendoza:
This PR hasn't been updated in some time, and I'm afraid the team won't have time to review it in the near future. Let's track future conversation in the ticket, PUP-4619, and we can re-open this PR if/when there's bandwidth to support it. Closing for now.

PR #3955, opened 2015-05-20
(maint) Archive spec_order.txt on appveyor failures

joshcooper:
Previously, tests executed in a platform-specific order in appveyor, and
if an order dependent failure happened, we didn't know what the order
was.

This commit sets the `LOG_SPEC_ORDER` environment variable which causes
puppet's spec helper to write each spec to `spec_order.txt` in the
project root.

It also adds an `on_failure` handler to archive the file, which can then
be downloaded from the job's Artifacts tab.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,7 @@ init:
   - SET
 install:
   - SET PATH=C:\Ruby21-x64\bin;%PATH%
+  - SET LOG_SPEC_ORDER=true
   - gem install bundler --quiet --no-ri --no-rdoc
   - bundle install --jobs 4 --retry 2 --without development
   - type Gemfile.lock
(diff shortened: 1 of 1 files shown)


joshcooper:
Alternatively, we could execute tests using rspec's order option and either specify a seed `rspec --order random:<seed>` (so the run is always in a consistent order), or let rspec generate the seed and write it to the console, e.g.

```
Randomized with seed 29492
```

and then replay the run using that seed.

That said, it could mask order dependent failures, and is not how most people run puppet specs.

joshcooper:
Of course, this PR passes appveyor, but you can see an example of what this would do if the specs failed in https://ci.appveyor.com/project/puppetlabs/puppet/build/4.1.0.342/artifacts

appveyor.yml
```diff
@@ -13,6 +14,8 @@ install:
 build: off
 test_script:
   - bundle exec rspec spec
+on_failure:
+  - appveyor PushArtifact .\spec_order.txt
```
ferventcoder: nice!

PR #3957, opened 2015-05-20
(maint) remove extra gems from appveyor

Iristyle:
- When running specs in our local Jenkins environment, the
   'extra' gems, which include 'msgpack' are not installed.  We do
   this because:

   - 'msgpack' requires building native code, which unnecessarily
     adds to the feedback loop
   - we intentionally try to maintain a clean environment without
     the Ruby DevKit installed

   This has previously worked on AppVeyor because they have the
   Ruby DevKit installed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ install:
   - SET PATH=C:\Ruby21-x64\bin;%PATH%
   - SET LOG_SPEC_ORDER=true
   - gem install bundler --quiet --no-ri --no-rdoc
-  - bundle install --jobs 4 --retry 2 --without development
+  - bundle install --jobs 4 --retry 2 --without development extra
   - type Gemfile.lock
   - ruby -v
 build: off


ferventcoder:
makes sense. Curious if spec_helper is necessary though?

joshcooper:
For reference, [list of appveyor preinstalled software](http://www.appveyor.com/docs/installed-software)

ferventcoder:
Waiting for Travis/AppVeyor...

joshcooper:
Pretty sure `puppetlabs_spec_helper` is not required. It was added in 0350afd5, but it introduces a cyclic dependency as `puppetlabs_spec_helper` depends on puppet. Also `tzinfo` is not used in `puppet` from what I can tell:

```
$ git grep -i tzinfo
Gemfile:  gem "tzinfo", :require => false
```

That said cleaning up the Gemfile can be a separate PR

ferventcoder:
:+1:

MikaelSmith:
Looks no worse than before.

ferventcoder:
The fun rdoc issue that just won't go away. Possibly made worse now. Perhaps we ensure those don't run as part of this fix?

ferventcoder:
I rekicked the PR

joshcooper:
The rdoc issue is PUP-4537, I've updated with my notes recently...

ferventcoder:
:+1:

PR #3958, opened 2015-05-20
(PUP-4636) Fix problem with environment.conf parser=future

hlindberg:
Before this, when the compiler was instantiated from an environment that
was configured to use parser=current, the initialization of the context
for compilation would do this for parser=current. This led to not having
loaded the ruby code for Puppet::Pops::Loaders, and because of this, the
first function call made in a puppet manifest would fail.

This fixes the problem by making the compiler check if the environment
it is going to *set* as the current environment has parser=future set
(instead of checking in the current environment).

This also moves the requirement to load the ruby logic to the loading of
'puppet/pops' since the loaders since quite some time back are lazilily
initialized.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -259,8 +259,11 @@ module Puppet
   # The single instance used for normal operation
   @context = Puppet::Context.new(bootstrap_context)
 
-  def self.future_parser?
-    env = Puppet.lookup(:current_environment) { return Puppet[:parser] == 'future' }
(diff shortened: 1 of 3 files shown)


jpartlow:
@hlindberg the future_parser? method is used pretty widely; are there any other cases that might require the environment to be passed in?

hlindberg:
@jpartlow I could not come up with one; it would be code that does what compile does - I am not sure we have any that does not use the compiler to uh... compile. Tests that do not use the compiler are forced to do all required setup on their own.

PR #3960, opened 2015-05-21
(PUP-4497) Ignore obsoleted packages on update check

melissa:
Prior to this commit, if there were any packages that had been
obsoleted, the check to see if any updates were available would fail.
This was because of the way the output from `yum check-update` was being
parsed. We take the output, split it into sets of three, and then take
the first two entries for each set, assuming they correspond to the
package name and the package version. When there are packages that have
been obsoleted, this method doesn't hold true. We find ourselves with a
tuplet consisting of something like ["Obsoleting", "Packages",
"NetworkManager.x86_64"], which then threw off all the following package
tuples. From here on out, what we are assuming in the package name is
actually the package version, and what we are assuming is the package
version is actually the repo where the package is located. When we then
go to parse out the details of the package version (by splitting out
elements based on a regex), we don't get the matches we expect and
Puppet then goes and vomits on your nice shoes.

This commit simple breaks off the process of parsing the output of `yum
check-updates` once the line with "Obsoleting" is reached. Everything
below here is a list of packages along with the package that is
obsoleted. i.e.,
```
Obsoleting Packages
NetworkManager.x86_64     1:1.0.0-14.git20150121.b4ea599c.el7     base
    NetworkManager.x86_64 1:0.9.9.1-25.git20140326.4dba720.el7_0  @updates
```

We could parse this out and return the first of the pair (in this case,
`NetworkManager.x86_64     1:1.0.0-14.git20150121.b4ea599c.el7`).
However, we currently are not doing anything with these obsoleted
packages, so this commit simple ignores the packages that obsolete and
are obsoleted.


kylog:
You had me at "vomits on your nice shoes" ;)

haus:
@melissa the fix seems sane. is there a simple spec test that would verify this?

kylog:
Can we add a fixture with the "Obsoleting" line and it to the spec coverage in `spec/unit/provider/package/yum_spec.rb`?

haus:
@melissa seems like @kylog and I had the same thought

melissa:
With the addition of these spec changes, but without the code changes, we get
```
Failures:

  1) Puppet::Type::Package::ProviderYum parsing the output of check-update with multiline entries parses multi-line values as a single package tuple
     Failure/Error: let(:output) { described_class.parse_updates(check_update) }
     NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./lib/puppet/provider/package/yum.rb:111:in `update_to_hash'
     # ./lib/puppet/provider/package/yum.rb:94:in `block in parse_updates'
     # ./lib/puppet/provider/package/yum.rb:93:in `each'
     # ./lib/puppet/provider/package/yum.rb:93:in `each_slice'
     # ./lib/puppet/provider/package/yum.rb:93:in `parse_updates'
     # ./spec/unit/provider/package/yum_spec.rb:468:in `block (4 levels) in <top (required)>'
     # ./spec/unit/provider/package/yum_spec.rb:471:in `block (4 levels) in <top (required)>'
```
which is the same error that was seen in production.

melissa:
@haus @kylog better?

kylog:
:+1: 

Thanks for jumping on this one so fast @melissa !

spec/unit/provider/package/yum_spec.rb
```diff
@@ -471,8 +471,20 @@
...
+
+      it "ignores all entries including and after 'Obsoleting Packages'" do
+        expect(output).not_to include("Obsoleting")
+        expect(output).not_to include("NetworkManager-bluetooth.x86_64")
+        expect(output).not_to include("1:1.0.0-14.git20150121.b4ea599c.el7")
+      end
+    end
   end
 
+
```
haus: want to remove this extra whitespace?

haus:
I'm :+1: as well, just a nitpick about whitespace.

PR #3961, opened 2015-05-22
(PUP-4622) Add a diff command in Puppet filebucket

fihuer:
This patch adds a diff subcommand to the filebucket command.
It implements a way to diff filebucket files. Either between
them or between a filebucket file and the local one.

### DIFF
--- a/lib/puppet/application/filebucket.rb
+++ b/lib/puppet/application/filebucket.rb
@@ -43,6 +43,9 @@ restore:
   path to this argument; you are not restricted to restoring the content
   to its original location.
 
+diff:
+  Print a diff in unified format between two checksums in the filebucket or between a checksum and
+  its matching file.
 
 DESCRIPTION
 -----------
@@ -97,7 +100,14 @@ EXAMPLE
     $ puppet filebucket backup /etc/passwd
     /etc/passwd: 429b225650b912a2ee067b0a4cf1e949
     $ puppet filebucket restore /tmp/passwd 429b225650b912a2ee067b0a4cf1e949
-
+    ## Diff between two files in the filebucket
+    $ puppet filebucket -l diff d43a6ecaa892a1962398ac9170ea9bf2 7ae322f5791217e031dc60188f4521ef
+    1a2
+    > again
+    ## Diff between the file in the filebucket and the current one
+    $ puppet filebucket -l diff d43a6ecaa892a1962398ac9170ea9bf2 /tmp/testFile
+    1a2
+    > again
 
 AUTHOR
 ------
@@ -115,7 +125,7 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
   def run_command
     @args = command_line.args
     command = args.shift
-    return send(command) if %w{get backup restore}.include? command
+    return send(command) if %w{get backup restore diff}.include? command
     help
   end
 
@@ -146,6 +156,34 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
     file = args.shift
     md5 = args.shift
     @client.restore(file, md5)
+  end
+
+  def diff
+    raise Puppet::Error, "Need exactly two arguments: filebucket diff <file_a> <file_b>" unless args.count == 2
+    left = args.shift
+    right = args.shift
+    if Puppet::FileSystem.exist?(left)
+      # It's a file
+      file_a = left
+      checksum_a = nil
+    else
+      file_a = nil
+      checksum_a = left
+    end
+    if Puppet::FileSystem.exist?(right)
+      # It's a file
+      file_b = right
+      checksum_b = nil
+    else
+      file_b = nil
+      checksum_b = right
+    end
+    if (checksum_a || file_a) && (checksum_b || file_b)
+      Puppet.info("Comparing #{checksum_a} #{checksum_b} #{file_a} #{file_b}")
+      print @client.diff(checksum_a, checksum_b, file_a, file_b)
+    else
+      raise Puppet::Error, "Need exactly two arguments: filebucket diff <file_a> <file_b>"
+    end
   end
 
   def setup
@@ -181,4 +219,3 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
     end
   end
 end
-
--- a/lib/puppet/file_bucket/dipper.rb
+++ b/lib/puppet/file_bucket/dipper.rb
@@ -2,6 +2,8 @@ require 'pathname'
 require 'puppet/file_bucket'
 require 'puppet/file_bucket/file'
 require 'puppet/indirector/request'
+require 'puppet/util/diff'
+require 'tempfile'
 
 class Puppet::FileBucket::Dipper
   include Puppet::Util::Checksums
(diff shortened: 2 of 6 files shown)


fihuer:
Appveyor failures are due to https://tickets.puppetlabs.com/browse/PUP-4537

lib/puppet/file_bucket/dipper.rb
```diff
@@ -56,6 +58,38 @@ def backup(file)
     end
   end
 
+  # Diffs two filebucket files identified by their sums
+  def diff(suma, sumb)
+    raise RuntimeError, "Diff is not supported on windows platforms" if Puppet.features.microsoft_windows?
+    source_path = "#{@rest_path}#{@checksum_type}/#{suma}"
+    if sumb and sumb != ''
```
peterhuene: Could we use `&&` here?  We try to avoid using `and` and `or` as logical operators.

lib/puppet/application/filebucket.rb
```diff
@@ -148,6 +158,17 @@ def restore
     @client.restore(file, md5)
   end
 
+  def diff
+    cksm_a = args.shift
+    cksm_b = args.shift || ''
+    if cksm_a
```
ffrank: Should this be `cksm_b`?

Can we have longer variable names? :)

Also, is the empty string a good default?

lib/puppet/file_bucket/dipper.rb
```diff
@@ -56,6 +58,38 @@ def backup(file)
     end
   end
 
+  # Diffs two filebucket files identified by their sums
+  def diff(suma, sumb)
+    raise RuntimeError, "Diff is not supported on windows platforms" if Puppet.features.microsoft_windows?
```
MikaelSmith: It'd be preferable to disable this only when `Puppet[:diff]` isn't set (which is the default on Windows). If a diff method is configured, it should be used. That requires modifying the indirector to use `Puppet[:diff]` instead of `diff` below.

lib/puppet/indirector/file_bucket_file/file.rb
```diff
@@ -19,6 +19,8 @@ def find(request)
           other_contents_file = path_for(request.options[:bucket_path], request.options[:diff_with], 'contents')
           raise "could not find diff_with #{request.options[:diff_with]}" unless Puppet::FileSystem.exist?(other_contents_file)
           return `diff #{Puppet::FileSystem.path_string(contents_file).inspect} #{Puppet::FileSystem.path_string(other_contents_file).inspect}`
```
MikaelSmith: Use `Puppet[:diff]` here?

lib/puppet/application/filebucket.rb
```diff
@@ -43,6 +43,9 @@ def help
   path to this argument; you are not restricted to restoring the content
   to its original location.
 
+diff:
+  Print a diff between two md5 in the filebucket or between a md5 and
+  its matching file
```
joshcooper: The comment says `md5` but implementation seems to use the "current" checksum type: `"#{@rest_path}#{@checksum_type}/#{suma}"`, for example if `Puppet[:digest_algorithm] = 'sha256'`

lib/puppet/file_bucket/dipper.rb
```diff
@@ -56,6 +58,38 @@ def backup(file)
...
+        :diff_with => sumb)
+    else
+      original_paths = Puppet::FileBucket::File.indirection.find(
+        source_path,
+        :bucket_path => @local_path,
+        :original_path => true)
+      raise Puppet::Error, "Couldn't retrieve file's original path" unless original_paths
+      original_paths = original_paths.to_s.split(":")
+      file_diff = ""
+      original_paths.each do |original_path|
```
peterhuene: So this will diff against every path where the file was added to the file bucket, correct?  So if the file's location (but not its contents) differs on a lot of agents, this would diff against all those locations?  I assume the diff will just show the "right side" of the diff for files that are missing.  It'll also download the file N number of times to perform each diff.  While it's nice that the user doesn't have to specify a path to diff against, I assume this could potentially be confusing to users if additional locations not relevant to the current node are displayed.

If users were required to specify the file to diff against it would eliminate the need for this special "original path" handling.  The ordering of the two (one a hash and the other a path) could also potentially be used to control the diff side (whichever is specified first is the left side, for example).  I don't know how much of an annoyance it would be to users to have to specify the file to diff against, but to me it makes a cleaner "diff" interface; i.e. it functions like diff on the command line, only it'll diff against file bucket contents instead of a file when given a hash.

peterhuene:
Hi @fihuer.  Thanks for the contribution!  Overall the added functionality looks useful.  We have a few questions and comments above that need to be addressed before we can merge this though.

lib/puppet/application/filebucket.rb
fihuer: It could be `nil` instead. You're right.

lib/puppet/application/filebucket.rb
fihuer: Yep, it should be
```
Print a diff between two checkums in the filebucket or between a checksum and
its matching file
```

lib/puppet/file_bucket/dipper.rb
fihuer: So you're suggesting to add an another optionnal parameter `--diff-with` (or whatever parameter pointing to a file) which will be diffed with the file in the filebucket. And, by default, it will try to find the "original path" using the `paths` file and if he finds multiple paths it will just fail. Is that right ?

lib/puppet/indirector/file_bucket_file/file.rb
fihuer: Yep it should be `Puppet[:diff]` or `Puppet::Util::Diff.diff` like in `dipper.rb`, but it was already here (Maybe for some obscure reason ...)

fihuer:
Hi @peterhuene @ffrank @joshcooper @MikaelSmith 
Thanks for the review !
I will try to give you another commit with what you've said asap.

lib/puppet/file_bucket/dipper.rb
peterhuene: I was actually suggesting not doing any of the magic for automatically diffing against where the file may be located on the current system and requiring the path to the file to diff against be passed like it would if you were diffing two hashes.  I prefer explicitness over magic, personally, but that doesn't mean we should go that route.  I just wanted to throw that suggestion out there as I think it gives a consistent "diff"-like interface to the command.

fihuer:
Rebased and squashed commits

ffrank:
Thanks. But I fear some notes did not make it in here after the previous triage sessions.

IIRC, everyone was very skeptical that the `original_paths` request option for the `FileBucketFile::File` terminus was a sound design.

Does anyone have the link to the recording floating around? I'd volunteer to review that and summarize.

MikaelSmith:
It might be in https://github.com/puppet-community/community-triage/blob/master/core/notes/2015-05-26.md.

ffrank:
Thanks - it was the following one, but notes just say that Peter meant to comment ;-)

ffrank:
There was one good and lengthy discussion [during that meeting](https://www.youtube.com/watch?v=7LyLtL9c7DU#t=17m30). I recommend checking that out @fihuer.

I believe we went even more in depth about the indirector details the week before that. I may look that up later.

fihuer:
I've removed the magic in the indirector. It now requires 2 args : checksum & checksum or checksum & file.

joshcooper:
@peterhuene to review

joshcooper:
PR Triage Notes

* @peterhuene to review
* Can the order of the left and right options be switched? Can they both be files?
* What the option looks like a SHA, but actually refers to a local file?

fihuer:
Now the args can be switched (it changes the diff order) and both args can be files. It checks if both args refers to an existent file with a `Puppet::FileSystem.exists?` . So non-existent file diffs are not handled.

lib/puppet/application/filebucket.rb
```diff
@@ -148,6 +158,29 @@ def restore
...
+    if checksum_b && Puppet::FileSystem.exist?(checksum_b)
+      # It's a file
+      file_b = checksum_b
+      checksum_b = nil
+    end
+    if (checksum_a || file_a) && (checksum_b || file_b)
+      Puppet.info("Comparing #{checksum_a} #{checksum_b} #{file_a} #{file_b}")
+      print @client.diff(checksum_a, checksum_b, file_a, file_b)
+    else
+      raise Puppet::Error, "Need at least two arguments: filebucket diff #{ARGV} <file_a> <file_b>"
```
peterhuene: The `ARGV` here prints out a weird error message, in my opinion:

`Error: Could not run: Need at least two arguments: filebucket diff ["filebucket", "diff"] <file_a> <file_b>`

I think we could remove `#{ARGV}` here as it's confusing.

lib/puppet/application/filebucket.rb
```diff
@@ -148,6 +158,29 @@ def restore
     @client.restore(file, md5)
   end
 
+  def diff
+    checksum_a = args.shift
+    checksum_b = args.shift
```
peterhuene: We should probably check for `args.count > 2` and report an error so that users can't do `puppet diff foo bar baz` and have it silently ignore baz.

lib/puppet/file_bucket/dipper.rb
```diff
@@ -56,6 +58,46 @@ def backup(file)
     end
   end
 
+  # Diffs two filebucket files identified by their sums
+  def diff(suma, sumb, filea, fileb)
```
peterhuene: Nitpick: the locals being passed to this method I think weren't as confusing, named `checksum_a`, `checksum_b`, `file_a` and `file_b` respectively.  `sum_a` and `sum_b` is also fine, but I think we should make it slightly more readable with an underscore.

lib/puppet/application/filebucket.rb
```diff
@@ -148,6 +158,29 @@ def restore
...
+    file_a = nil
+    file_b = nil
+    if checksum_a && Puppet::FileSystem.exist?(checksum_a)
```
peterhuene: Rather than `Dipper#diff` taking four arguments and having a somewhat complicated implementation, I think it should take only two (e.g. left/right) and these checks for file existence should be moved there.  If either does not exist, download the file to a temp file and use the temp file as the argument to `Puppet::Util::Diff.diff` (i.e. only one call).

peterhuene:
@fihuer Thanks for your patience with this PR.  Everything looks good functionally, so I've merged this into the upcoming Puppet 4.3 release.  Thanks again for the contribution!

PR #3963, opened 2015-05-22
(doc) Clarify auth_membership

ferventcoder:
Auth_membership plays a role in how a provider can set a group to
empty based on whether it is considered authoritative or not. The
documentation should provide an indication that setting a group
empty is possible.

### DIFF
--- a/lib/puppet/type/group.rb
+++ b/lib/puppet/type/group.rb
@@ -80,7 +80,9 @@ module Puppet
 
     newproperty(:members, :array_matching => :all, :required_features => :manages_members) do
       desc "The members of the group. For directory services where group
-      membership is stored in the group objects, not the users."
+      membership is stored in the group objects, not the users. Use
+      with auth_membership to determine whether the specified members
+      are inclusive or the minimum."
 
       def change_to_s(currentvalue, newvalue)
         currentvalue = currentvalue.join(",") if currentvalue != :absent
@@ -108,7 +110,9 @@ module Puppet
     end
 
     newparam(:auth_membership, :boolean => true, :parent => Puppet::Parameter::Boolean) do
-      desc "whether the provider is authoritative for group membership."
+      desc "Whether the provider is authoritative for group membership. This
+        must be set to true to allow setting the group to no members with
+        `members => [],`."
       defaultto true
     end
 


joshcooper:
@nfagerlund @gguillotte can you review this PR?

lib/puppet/type/group.rb
```diff
@@ -108,7 +110,9 @@ def is_to_s(currentvalue)
     end
 
     newparam(:auth_membership, :boolean => true, :parent => Puppet::Parameter::Boolean) do
-      desc "whether the provider is authoritative for group membership."
+      desc "Whether the provider is authoritative for group membership. This
+        must be set to true to allow setting the group to empty e.g.
+        `members => [],`."
```
gguillotte: Per the style guide, avoid e.g.

It might also help to be specific ("This must be set to true to allow setting the group to an empty value, such as `members => [],`.").

lib/puppet/type/group.rb
```diff
@@ -80,7 +80,9 @@ def sync
 
     newproperty(:members, :array_matching => :all, :required_features => :manages_members) do
       desc "The members of the group. For directory services where group
```
gguillotte: We can avoid a sentence fragment with "The group's members in directory services...".

lib/puppet/type/group.rb
```diff
@@ -80,7 +80,9 @@ def sync
...
-      membership is stored in the group objects, not the users."
+      membership is stored in the group objects, not the users. Combine
+      with auth_membership to determine whether the specified members
```
gguillotte: Would "Use auth_membership to..." work? Combine suggests something more complex than just setting a parameter.

gguillotte:
:+1: aside from some nit picks.

lib/puppet/type/group.rb
ferventcoder: I didn't write that sentence and it doesn't even make sense to me at all.

ferventcoder: The point of that is that there are groups and users and the only difference is that one is shown as a group.

lib/puppet/type/group.rb
ferventcoder: It's not such as, it is the only way of doing it. Would we recommend "with `members => []`" instead?

gguillotte: Yes.

ferventcoder: "empty **value**" doesn't really make sense in the context of Groups. It's a bucket, not an object.

lib/puppet/type/group.rb
gguillotte: If we can make that clear here, we should.

lib/puppet/type/group.rb
ferventcoder: :+1: "Use with" to suggest there is another provider value that can be set. :)

lib/puppet/type/group.rb
ferventcoder: The fragment is "For directory services where group membership is stored in the group objects, not the users." It's a bit outside the context of this particular ticket, but if you have a good suggestion, I can make the change.

lib/puppet/type/group.rb
gguillotte: Empty bucket, then? Anything more specific than "setting the group to empty" would be clearer.

ferventcoder: It's actually very clear if you have the context of what Groups are in the sense of Windows. Setting a group to have no members then?

lib/puppet/type/group.rb
gguillotte: I'm fine with leaving it for a more comprehensive style pass through these. :+1:

lib/puppet/type/group.rb
ferventcoder: Updted.

lib/puppet/type/group.rb
ferventcoder: I don't like the fragment either FWIW. :) I just don't have a good suggestion on changing it currently.

nfagerlund:
I'm cool with this.

ferventcoder:
@joshcooper :shipit:

PR #3964, opened 2015-05-27
(PUP-4652) Don't rely on beaker's `agent` method

joshcooper:
Previously, the test passed when run with a single agent node, but
failed when run with multiple agent nodes, with the error:

    TypeError Exception: no implicit conversion of String into Integer

The root cause is that Beaker metaprograms methods for each role,
where the method returns either a single host or an array of hosts
with that role. And in the latter case, we were calling `['platform']`
on an array causing the TypeError.

This commit modifies the test to not assume there is a single agent.
Instead we always create unix and windows external facts on the master,
and for each agent, verify the external fact is resolved.

### DIFF
--- a/acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
+++ b/acceptance/tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb
@@ -26,14 +26,6 @@ win_fact =<<EOM
 echo foo=bar
 EOM
 
-if agent['platform'] =~ /windows/
-  fact_content = win_fact
-  extension = '.bat'
-else
-  fact_content = unix_fact
-  extension = '.sh'
-end
-
 apply_manifest_on(master, <<MANIFEST, :catch_failures => true)
 File {
   ensure => directory,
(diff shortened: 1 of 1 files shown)


joshcooper:
```
$ be rake ci:test:aio SHA=5f82aa3d36972c619e2b6989e98a5d851258417e TESTS=tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb CONFIG=config/nodes/win2012r2-rubyx64.yaml
...
dwayj4bzdejwh2g.delivery.puppetlabs.net (agent-2012r2-x86_64-rubyx64) executed in 0.15 seconds
tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb passed in 82.53 seconds
```

```
$ be rake ci:test:aio SHA=5f82aa3d36972c619e2b6989e98a5d851258417e TESTS=tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb CONFIG=config/nodes/redhat-7-x86_64.yaml
...
it4ozjc6riaedh6.delivery.puppetlabs.net (agent) executed in 0.05 seconds
tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb passed in 80.78 seconds
```

joshcooper:
@camlow325 this should resolve the puppetserver acceptance failure

camlow325:
:+1:

PR #3966, opened 2015-05-27
(PUP-2728) zypper should always be the default provider for Suse (backport).

kylog:


joshcooper:
Isn't changing the default provider and removing `rug` a breaking change, and so can't be backported to 3.x? Or were things horribly broken and could never work?

kylog:
That's a good question, and I have to offer up a caveat that my understanding of this is all second hand, based on comments in #2716, comments in https://tickets.puppetlabs.com/browse/PUP-2728 and a little bit of web searching. I'll cc @m4ce and @adrienthebo who were involved in the initial review, because they may have more context.

That said, my understanding is that zypper has been the preferred/recommended package provider since  suse 10 (perhaps in one of the SPs). Suse 9 doesn't show up on https://www.suse.com/lifecycle/, so I assume it's EOL and can be disregarded.

If the above is correct, this would only be a breaking change for someone who had, say, both yum and zypper installed (so the providers would "tie" on specificity) and would be relying on the alphabetization tie-breaker to prefer 'yum'. If *that* is the case, then this is arguably a simple bug-fix and only breaking from a semver sense in that it provides deterministic behavior where the previous behavior was accidental.

All that said, I don't use suse myself so ... not sure.

adrienthebo:
SLES 10 SP1 was the first release that used zypper (http://www.ucs.cam.ac.uk/support/unix-support/updates/suse/sles-local); if we don't have to support SLES 10 SP0 then we can do this. It's a small breaking change, but a breaking change regardless - so how closely do we want to adhere to semver?

kylog:
Thanks for the context on when zypper was introduced in SLES 10. Also in case anyone else wonders about dates: per http://en.wikipedia.org/wiki/SUSE_Linux_Enterprise_Server, SLES 10 SP1 was released 2007-06-18, so systems updated since then would benefit from this.

That inclines me to include this, but I think the strictest reading of semver would exclude it. 

Also keep in mind that this came up as a PE request so we could certainly re-target this PE-only (where the SLES 10 support is for SP4 and up). I targeted foss puppet because it seems of general utility (and there's charm to minimizing the foss/PE differences). But I'm not aware of any foss users asking for the backport.

@joshcooper thoughts?

joshcooper:
If the `rug` provider is broken and unusable in 3.x, then anyone running SLES 10 SP1 and up would have had to specify `provider => zypper` in their manifests, and won't be affected by this PR.

If anyone is using the `rug` provider, then this will break anyone that hasn't specified a default provider. That may be a very small number of people, and is perhaps 0, but it feels wrong to me to make that change in 3.x.

kylog:
I'll close this, untarget the ticket from 3.8.2, and open a PR against pe-puppet.

bigon:
Has 199b52e01351eb4574a52fe43b4d5696681b704b been commited to the 3.x branch? I don't see anything

MosesMendoza:
Hi @bigon, thanks for reaching out. I think the conclusion was that users in the 3.x series may be relying on `rug` as the default provider and would/could be surprised by this change, so it was not targeted at 3.x. Additionally, the 3.x branch is now in security-only maintenance mode, so I don't think it would be a candidate for changes of this nature.

bigon:
Well it's not only rug, I've added pe_gem module to my puppet code and all the packages resources (on sles) are using it instead of zypper

PR #3967, opened 2015-05-27
(PUP-4635) Fix pacman provider to handle failure on group queries.

vdloo:
Installing packages with pacman failed with
Error: /Stage[main]/Main/Package[git]: Could not evaluate: Error while getting installed groups
This is fixed by allowing a non-zero exit code for some of the pacman calls.

https://tickets.puppetlabs.com/browse/PUP-4635

### DIFF
--- a/lib/puppet/provider/package/pacman.rb
+++ b/lib/puppet/provider/package/pacman.rb
@@ -28,7 +28,9 @@ Puppet::Type.type(:package).provide :pacman, :parent => Puppet::Provider::Packag
   # Checks if a given name is a group
   def self.group?(name)
     begin
-      !pacman("-Sg", name).empty?
+      !execpipe([command(:pacman), "-Sg"], false).empty? do |process|
+        process.read.chomp!
+      end
     rescue Puppet::ExecutionFailure
       fail("Error while determining if '#{@resource[:name]}' is a group")
     end
@@ -98,7 +100,7 @@ Puppet::Type.type(:package).provide :pacman, :parent => Puppet::Provider::Packag
       # Build a hash of group name => list of packages
       command = [command(:pacman), "-Sgg"]
       command << filter if filter
-      execpipe(command) do |process|
+      execpipe(command, false) do |process|
         process.each_line do |line|
           name, package = line.split
           packages = (groups[name] ||= [])
--- a/spec/unit/provider/package/pacman_spec.rb
+++ b/spec/unit/provider/package/pacman_spec.rb
@@ -199,7 +199,7 @@ describe Puppet::Type.type(:package).provider(:pacman) do
   describe "when querying" do
     it "should query pacman" do
       executor.expects(:execpipe).with(["/usr/bin/pacman", '-Q'])
-      executor.expects(:execpipe).with(["/usr/bin/pacman", '-Sgg', 'package'])
+      executor.expects(:execpipe).with(["/usr/bin/pacman", '-Sgg', 'package'], false)
       provider.query
     end
 
@@ -211,7 +211,7 @@ otherpackage 1.2.3.4
 package 1.01.3-2
 yetanotherpackage 1.2.3.4
 EOF
-      executor.expects(:execpipe).with(['/usr/bin/pacman', '-Sgg', 'package']).yields('')
+      executor.expects(:execpipe).with(['/usr/bin/pacman', '-Sgg', 'package'], false).yields('')
 
       expect(provider.query).to eq({ :name => 'package', :ensure => '1.01.3-2', :provider => :pacman,  })
     end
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/package/pacman.rb
```diff
@@ -8,7 +8,6 @@
   This provider supports the `install_options` attribute, which allows command-line flags to be passed to pacman.
   These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
   or an array where each element is either a string or a hash."
-
```
peterhuene: Can you restore this line break so there's some separation between the doc comment and the `yaourt?` method?

lib/puppet/provider/package/pacman.rb
```diff
@@ -28,7 +27,10 @@ def self.yaourt?
   # Checks if a given name is a group
   def self.group?(name)
     begin
-      !pacman("-Sg", name).empty?
+      execpipe([command(:pacman), "-Sg"], failonfail=false) do |process|
```
peterhuene: This can just be `execpipe([command(:pacman), '-Sg'], false)`.  `failonfail=false` sets a local called `failonfail` to false and then passes the result of the expression.  Same result with just `false`, only no local is set.

lib/puppet/provider/package/pacman.rb
```diff
@@ -98,7 +100,7 @@ def self.get_installed_groups(installed_packages, filter = nil)
       # Build a hash of group name => list of packages
       command = [command(:pacman), "-Sgg"]
       command << filter if filter
-      execpipe(command) do |process|
+      execpipe(command, failonfail = false) do |process|
```
peterhuene: Same comment as above.

lib/puppet/provider/package/pacman.rb
```diff
@@ -28,7 +27,10 @@ def self.yaourt?
   # Checks if a given name is a group
   def self.group?(name)
     begin
-      !pacman("-Sg", name).empty?
+      execpipe([command(:pacman), "-Sg"], failonfail=false) do |process|
+        output = process.readlines.join('').chomp!
```
peterhuene: I think this should probably be simply `process.read.chomp!` and move the `empty?` to the result of `execpipe` (the blocks are generally expected to return strings not booleans).

peterhuene:
Hi @vdloo.  Thanks so much for taking the time to put up a PR to fix this!  Generally it looks like your fix is good.  Other than the above comments, I think we should add a test for the condition where pacman returns 1 for the group query.  The rspec code can be a little confusing, so if you're unable to figure out adding that test let me know and I can add one.

peterhuene:
@vdloo Also, could you amend your commit message per our [contributing guidelines](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md)?  We generally prefix the commit message to reference the ticket, like so: `(PUP-4635) Fix pacman provider to handle failure on group queries.` or some such.  Thanks!

peterhuene:
Review note: the AppVeyor failures are known issues and can be ignored.

peterhuene:
@vdloo Also, I have a feeling that we'll want to target this fix at stable.  If it's not too difficult would you mind applying the fix to the stable branch with the above corrections and putting up a new PR against that branch?  That way the fix can go into the next 4.1.x release and then we'll merge it up into the 4.2.x release branch (which is currently master).  Please let me know if you would like me to do this for you instead.  Thanks!

vdloo:
I tried adding the following to pacman_spec.rb:229

```
     it "should return 1 as exit code" do
       executor.expects(:execpipe).with(["/usr/bin/pacman", '-Q'])
       executor.expects(:execpipe).with(["/usr/bin/pacman", '-Sgg', 'package'], false).stubs(:exitstatus).returns 1
       provider.query
     end
```

but that seemed to have no effect (when changing the 1 to 0). 
I'm not familiar with ruby at all so I'm afraid that figuring this out will take me a lot of time + I am in a train with limited internet all day unfortunately. If you could fix the test and make the pr for the stable branch that would be fantastic, thanks for the quick response!

peterhuene:
@vdloo Thanks again for putting this forward.  I will get this fixed up and re-target against stable.

peterhuene:
@vdloo After discussion with @kylog, we don't need to target this against stable and can keep it at master.

lib/puppet/provider/package/pacman.rb
```diff
@@ -28,7 +28,9 @@ def self.yaourt?
   # Checks if a given name is a group
   def self.group?(name)
     begin
-      !pacman("-Sg", name).empty?
+      !execpipe([command(:pacman), "-Sg"], false).empty? do |process|
```
peterhuene: This is applying `!` to the string, which will always be false.

We need to add a `.empty?` to the closing `end` of the block.

peterhuene: Also, I just realized we're no longer passing name like we were before.

peterhuene:
Hi @vdloo.  I've put up a PR that includes your commit that was amended with the above fixes along with some additional tests that we were missing for `group?` and `get_installed_groups`: https://github.com/puppetlabs/puppet/pull/3974.  As such, I'm going to close this PR in favor of that one.  When you have a chance, can you please test the changes out to your satisfaction?  Cheers.

lib/puppet/provider/package/pacman.rb
vdloo: Whoops sorry, my ruby skills are really lacking. Thanks for all the effort!

PR #3969, opened 2015-05-28
(PUP-2400) Load loop kernel module for mount tests

johnduarte:
This commit adds a setup step to the mount resource tests to make
sure that the `loop` kernel module is loaded in order to allow the
system under test to mount loopback devices.

### DIFF
--- a/acceptance/tests/resource/mount/destroy.rb
+++ b/acceptance/tests/resource/mount/destroy.rb
@@ -28,6 +28,9 @@ agents.each do |agent|
   step "(setup) create mount point"
   on(agent, "mkdir /#{name}", :acceptable_exit_codes => [0,1])
 
+  step "(setup) ensure loop kernel module is installed"
+  on(agent, "modprobe loop", :acceptable_exit_codes => (0..254))
+
   step "(setup) add entry to filesystem table"
   on(agent, "echo '/tmp/#{name}  /#{name}  ext3  loop  0  0' >> #{fstab}")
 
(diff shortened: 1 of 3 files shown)


acceptance/tests/resource/mount/destroy.rb
```diff
@@ -28,6 +28,9 @@
   step "(setup) create mount point"
   on(agent, "mkdir /#{name}", :acceptable_exit_codes => [0,1])
 
+  step "(setup) ensure loop kernel module is installed"
```
er0ck: should probably be using single quotes here, but i'm inclined to let it go O:)

acceptance/tests/resource/mount/modify.rb
```diff
@@ -29,6 +29,9 @@
   step "(setup) create mount point"
   on(agent, "mkdir /#{name}", :acceptable_exit_codes => [0,1])
 
+  step "(setup) ensure loop kernel module is installed"
+  on(agent, "modprobe loop", :acceptable_exit_codes => (0..254))
```
kurtwall: `mount` returns a respectable number of meaningful error codes; checking and displaying those might eas debugging in the future.

kurtwall:
Spec failure is unrelated.

kurtwall:
I'm not convinced that coding around a VM image problem is the right solution here.

johnduarte:
This PR is being redacted in favor of ensuring that the `loop` kernel module is loaded on failing platforms.

PR #3970, opened 2015-05-28
(PUP-4666) Create unique tmpdir for 'build_ignore_module_file' test

camlow325:
This commit changes the 'build_ignore_module_file' acceptance test to
create its own unique tmpdir for use while the test is running.  In the
previous commit, the test would try to populate the tmpdir at a
hard-coded location, '/tmp/test', which would fail if that directory were
pre-existing.

### DIFF
--- a/acceptance/tests/modules/build/build_ignore_module_file.rb
+++ b/acceptance/tests/modules/build/build_ignore_module_file.rb
@@ -1,7 +1,7 @@
 test_name 'PUP-3981 - C63215 - Build Module Should Ignore Module File'
 
 #Init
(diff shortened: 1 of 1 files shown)


joshcooper:
:+1: 

```
$ be rake ci:test:aio SHA=540c4550c1604c2e87e28d43151e4c21195198c7 TESTS=tests/modules/build/build_ignore_module_file.rb CONFIG=config/nodes/redhat-7-x86_64.yaml
...
tests/modules/build/build_ignore_module_file.rb passed in 0.78 seconds
```

PR #3972, opened 2015-05-28
(maint) Add acceptance configs for OSX

justinstoller:
Note: We currently aren't building packages for Vivid yet so I don't have an artifact to test these configs with. Having said that, I copied existing configs and updated them, this should be a very low risk change.

I used a "code name" approach to the platform as is done with Debian. I don't know if this is how you folks prefer the OSX testing or if you'd want to refer to their platform as numeric.


kylog:
Can you add `[skip ci]` to the commit messages of acceptance-only PRs?

justinstoller:
updated

PR #3973, opened 2015-05-28
(PUP-4665) Add inspect function to scope class

fpringvaldsen:
Add an inspect function to the Puppet::Parser::Scope class that
aliases to the to_s method so that calling inspect on a scope no
longer generates a massive, unused string.


hlindberg:
Travis had an intermittent problem (same as in other PRs) - retriggered one of the travis builds...

PR #3977, opened 2015-05-29
(PUP-4633) fix non-ASCII user comment with ruby >= 2.1

yastupin:
I submit this PR as a trivial fix for PUP-4633.

More details are available in the debian bug report :
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=782494

PS: I did not sign the CLA but it looks like it complies with the Trivial Patch Exemption Policy.

### DIFF
--- a/spec/unit/type/user_spec.rb
+++ b/spec/unit/type/user_spec.rb
@@ -344,13 +344,20 @@ describe Puppet::Type.type(:user) do
     end
   end
 
-  describe "when managing comment on Ruby 1.9", :if => String.method_defined?(:encode) do
-    it "should force value encoding to ASCII-8BIT" do
-      value = 'abcd™'
-      expect(value.encoding).to eq(Encoding::UTF_8)
-      user = described_class.new(:name => 'foo', :comment => value)
-      expect(user[:comment].encoding).to eq(Encoding::ASCII_8BIT)
-      expect(user[:comment]).to eq(value.force_encoding(Encoding::ASCII_8BIT))
+  describe "when managing comment" do
+    before :each do
+      @value = 'abcd™'
+      expect(@value.encoding).to eq(Encoding::UTF_8)
+      @user = described_class.new(:name => 'foo', :comment => @value)
+    end
+
+    it "should be converted to ASCII_8BIT for ruby 1.9 / 2.0", :if => RUBY_VERSION < "2.1.0" && String.method_defined?(:encode) do
+      expect(@user[:comment].encoding).to eq(Encoding::ASCII_8BIT)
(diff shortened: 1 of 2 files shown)


spec/unit/type/user_spec.rb
```diff
@@ -344,13 +344,15 @@ def self.instances; []; end
...
   end
 
-  describe "when managing comment on Ruby 1.9", :if => String.method_defined?(:encode) do
-    it "should force value encoding to ASCII-8BIT" do
-      value = 'abcd™'
-      expect(value.encoding).to eq(Encoding::UTF_8)
-      user = described_class.new(:name => 'foo', :comment => value)
-      expect(user[:comment].encoding).to eq(Encoding::ASCII_8BIT)
-      expect(user[:comment]).to eq(value.force_encoding(Encoding::ASCII_8BIT))
+  if RUBY_VERSION < "2.1.0"
```
peterhuene: Can we merge this clause with the `:if` to the spec?

peterhuene:
Hi @yastupin.  Thanks for the contribution!  I think this makes sense, although having explicit ruby checks in puppet source seems a little unwieldy to me, although I can't think of a way to fix this otherwise.  I'm going to mark this as up for discussion for tomorrow's Community PR Triage.

peterhuene:
Also, I think we should add a spec example that stubs `RUBY_VERSION` to be `>= 2.1` and ensures `force_encoding` is not called.

yastupin:
Hi @peterhuene, I agree that testing RUBY_VERSION like this is not pretty, but ruby/ruby@4c009414119347e87d9bb2402617a4585dd3f341 does not provide an easy way to check for the fix...

I've updated the spec according to your comments, thanks.

joshcooper:
@yastupin could you sign the CLA?

@peterhuene to review

peterhuene:
I'm reviewing this today.  Sorry for the delay.

peterhuene:
Hi @yastupin.  This has been merged to master and will be available in an upcoming release of Puppet (4.3.0).  Thanks again for the contribution!

KlavsKlavsen:
This bug has cropped up again in puppet-4.6.1 (puppet-agent 1.6.1) - since we upgraded from puppet-3.7.1 on an agent. it only happens when we run puppet in cron - and the workaround LANG=en_US.utf8 in front of puppet agent -t - works.

er0ck:
@KlavsKlavsen can you open a new ticket for this regression, please?

KlavsKlavsen:
@er0ck Ticket opened :) https://tickets.puppetlabs.com/browse/PUP-6777

PR #3979, opened 2015-05-29
(maint) Remove rubygems_version from gemspec.

peterhuene:
Removing `rubygems_version` from gemspec as it breaks under the most
recently released version of bundler.

From the documentation this should not be set and is instead generated
by rubygems at packaging.


kylog:
Should we point this all the way back to 3.x?

peterhuene:
Yeah, probably should :/ Okay, I'll close this one too.

PR #3984, opened 2015-05-30
(PUP-4673) Use proxy when fetching latest for pip packages

nibalizer:
Puppet uses an xmlrpc request to ask pypi what the latest possible
version of the package is. This adds some inspection of the environment
to optionally set proxy information for the xmlrpc request. Future work
could refactor this to use pip search.

### DIFF
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -2,7 +2,7 @@
 # <http://pip.openplans.org/>
 
 require 'puppet/provider/package'
-require 'xmlrpc/client'
+require 'open-uri'
 
 Puppet::Type.type(:package).provide :pip,
   :parent => ::Puppet::Provider::Package do
@@ -56,11 +56,8 @@ Puppet::Type.type(:package).provide :pip,
   # cache of PyPI's package list so this operation will always have to
   # ask the web service.
   def latest
-    client = XMLRPC::Client.new2("http://pypi.python.org/pypi")
-    client.http_header_extra = {"Content-Type" => "text/xml"}
-    client.timeout = 10
-    result = client.call("package_releases", @resource[:name])
-    result.first
+    result = JSON.parse(open("https://pypi.python.org/pypi/#{@resource[:name]}/json").read)
+    result['releases'].first
   rescue Timeout::Error => detail
     raise Puppet::Error, "Timeout while contacting pypi.python.org: #{detail}", detail.backtrace
   end
--- a/spec/unit/provider/package/pip_spec.rb
+++ b/spec/unit/provider/package/pip_spec.rb
@@ -12,7 +12,6 @@ describe provider_class do
     @client = stub_everything('client')
     @client.stubs(:call).with('package_releases', 'real_package').returns(["1.3", "1.2.5", "1.2.4"])
     @client.stubs(:call).with('package_releases', 'fake_package').returns([])
-    XMLRPC::Client.stubs(:new2).returns(@client)
   end
 
   describe "parse" do


nibalizer:
Cancelled?

kylog:
Wtf? I think appveyor maybe got confused by the open/close/re-open of the PR? I can retrigger.

nibalizer:
Haha, I guess github is smart enough to preserve numbers when the 'same' pr shows up against a diff branch?

Also is this the right place for this? I don't know what procedure is.

kylog:
Yeah it was confused. I clicked through to the link on *this* PR, hit re-build and it said #3983 is closed. Which is true, but this is #3984. So instead I clicked the "build again with latest commit on this branch" and let's see if it figures it all out.

Appveyor has been *super* handy, but there are some rough edges.

kylog:
Maybe? I feel like we do that type of thing all the time, and I haven't seen this before. But ... dunno.

As for branch, I'm okay targeting 3.x, but can you open a ticket and reference it in the commit message like "(PUP-1234) Use proxy ...."?

Thanks! (Btw, haven't looked at the *content* of the commit yet and about to head out. Just getting the paperwork out of the way :))

peterhuene:
Hi @nibalizer.  Thanks for the contribution.  The fix looks good, but appears to be missing a test.  Could we add an example that simply stubs the environment variable and ensures the proper overload of ` XMLRPC::Client.new2` gets called with the right hostname and port?

lib/puppet/provider/package/pip.rb
```diff
@@ -56,7 +56,13 @@ def query
   # cache of PyPI's package list so this operation will always have to
   # ask the web service.
   def latest
-    client = XMLRPC::Client.new2("http://pypi.python.org/pypi")
+    proxy = ENV["http_proxy"]
+    if proxy.nil? then
+      client = XMLRPC::Client.new2("http://pypi.python.org/pypi")
+    else
+      host, port = proxy[7..-1].split(':')
+      client = XMLRPC::Client.new2("http://pypi.python.org/pypi", proxy_host=host, proxy_port=port)
```
peterhuene: It appears that `proxy_host` and `proxy_port` are unused.

nibalizer: I don't follow. Can you clarify your concern?

peterhuene: This is setting two local variables, `proxy_host` and `proxy_port` in the assignments, but those aren't used anywhere.

nibalizer: Oh no, those are arguments to the instantiation of teh XMLRPC::Client class.

nibalizer: Wait, I actually don't know ruby. Let me fix this up a bit.

peterhuene: Right, but the names here have no bearing on the arguments:

```ruby
def foo(a, b)
  puts a, b
end

foo(c = 1, d = 2)
puts c
puts d
```

Results in:
```
1
2
1
2
```

The above is simply assigning `c` to `1` and `d` to `2` and passing the result of the expressions (the current values of `c` and `d` respectively) to the function.

peterhuene: Also, it appears that `XMLRPC::Client#new2` takes the argument in a “host:port" format, so splitting is probably not necessary.

peterhuene:
We discussed this PR in triage this morning.  In Puppet stable (4.2) we currently have an implementation of the utility module `Puppet::Util::HttpProxy` that can be used to respect `http_proxy` and `no_proxy` environment variables, along with user and password settings for the proxy from settings.  I don't know if we want to keep this PR's implementation for 3.x when a common implementation already exists in 4.x.

It doesn't seem like the `XMLRPC::Client` class can pass a user or password to the proxy (I assume the user and password arguments to `new3` are for the request itself, not for the proxy) though :/

lib/puppet/provider/package/pip.rb
peterhuene: As such, this could simply become `client = XMLRPC::Client.new2("http://pypi.python.org/pypi", ENV['http_proxy'] || ENV['HTTP_PROXY'])` without having to split or check for existence.

nibalizer:
@peterhuene I believe I have written code that is actually ruby now, and not python. :)

As for the correct way to solve this, I think the /real/ way to do this it to change the provider to shell out to the pip search function so that pip can evaluate if the package is latest or not. I'm not sure what the life of 3.x is and if changes like that would be landable on 3.x. Certainly on 4.x I think its the way to go.

I am a bit biased on where things land because a 3.x change is going to consumable by me, and a 4.x change won't help me solve problems for some time.

Please let me know how you'd like me to proceed. I can also show up to the triage next week if that helps.

nibalizer:
I chatted a bit with the pip maintainers. There isn't a good way right now to ask pip what pending upgrades are available. pip list --outdated will tell us everything that has optional upgrades, however its format is explicitly not stable right now.

The pip toolmasters are open to commits to lock down or add behaviour to make management with CM easier, but those features would only land on new pip, leaving us in a weird spot with which pip versions puppet supports.

One thing that did come out of the conversation is that curling the way we are doing is more or less right. They do request that we use the json api instead of the xml api because the xml hits their server every time and generates load, whereas the jsonapi stuff can be cached in fastly.

nibalizer:
@peterhuene I took a stab at writing the test... failed pretty hard. Can you point me to a place in the code where this kind of testing is already in place? That was I can copy from that.

joshcooper:
@nibalizer Would you be up for changing the pip provider to use the json api? It would make it easier to backport this to 3.x and merge up cleanly to stable & master, and allow us to use puppet's http connection handling, e.g. authenticating proxies.

nibalizer:
@joshcooper sure thing. I'll take a stab at that in the next few days.

hkenney:
@nibalizer Still working on this? Just checking in

nibalizer:
@HAIL9000 Ya poking at it now. Hoping to be able to spend some time on this during puppethack. I'll reach out when I have intelligent questions.

nibalizer:
Well here is a json-functional component. I'm still (as usual) murky on what testing to do where, maybe someone can assist me during #puppethack on that.

HAIL9000:
@nibalizer Hey, saw there was an update on this but it looks like Travis is still failing. Have you had a chance to take a look at the spec failures?

nibalizer:
@HAIL9000 only a bit. I'm a super novice when it comes to rspec/spec. I'll try to bang on that when I get a chance. If someone could point me in the right direction that would be awesome!

lib/puppet/provider/package/pip.rb
```diff
@@ -56,11 +56,8 @@ def query
   # cache of PyPI's package list so this operation will always have to
   # ask the web service.
   def latest
-    client = XMLRPC::Client.new2("http://pypi.python.org/pypi")
-    client.http_header_extra = {"Content-Type" => "text/xml"}
-    client.timeout = 10
-    result = client.call("package_releases", @resource[:name])
-    result.first
+    result = JSON.parse(open("https://pypi.python.org/pypi/#{@resource[:name]}/json").read)
```
branan: open-uri will use the proxy environment variables but will ignore puppets `http_proxy_*` settings. The right way to do HTTP requests from inside puppet is with our own Net:HTTP wrapper in `puppet/network/http/factory.rb`.

branan: Which... is apparently marked API private. As is `Puppet::Network::HTTP::Pool`.

I suppose since this is a core provider either method is acceptable, but it feels like bad form to be using private APIs in our core providers without exposing similar functionality to 3rd-party ones :/

joshcooper: Ideally you could do something like:

```
http = Puppet::Network::HttpPool.http_instance('pypi.python.org', '443', true)
response = http.get("/pypi/cufflinks/json")
case response
when Net::HTTPSuccess
  puts JSON.parse(response.body)
else
  raise "Failed to request pypi metadata: #{response.code}, #{response.message}"
end
```

However, puppet will only accept the puppet CA resulting in:

```
certificate verify failed: [unable to get local issuer certificate for /C=US/O=DigiCert Inc/OU=www.digicert.com/CN=DigiCert SHA2 Extended Validation Server CA] (Puppet::Error)
```

If you disable ssl verification (for testing):

```
http = Puppet::Network::HttpPool.http_instance('pypi.python.org', '443', true, false)
```

It works as expected:

```
{"info"=>{"maintainer"=>nil, "docs_url"=>nil, "requires_python"=>nil, "maintainer_email"=>nil, "cheesecake_code_kwalitee_id"=>nil, "keywords"=>"pandas,plotly,plotting", "package_url"=>"http://pypi.python.org/pypi/cufflinks", "author"=>"Jorge Santos", "author_email"=>"santos.jorge@gmail.com", "download_url"=>"UNKNOWN", "platform"=>"UNKNOWN", "version"=>"0.6.1", "cheesecake_documentation_id"=>nil, "_pypi_hidden"=>false, "description"=>"UNKNOWN", "release_url"=>"http://pypi.python.org/pypi/cufflinks/0.6.1", "downloads"=>{"last_month"=>2443, "last_week"=>635, "last_day"=>74}, "_pypi_ordering"=>7, "classifiers"=>[], "name"=>"cufflinks", "bugtrack_url"=>nil, "license"=>"MIT", "summary"=>"Productivity Tools for Plotly + Pandas", "home_page"=>"https://github.com/santosjorge/cufflinks", "stable_version"=>nil, "cheesecake_installability_id"=>nil}, "releases"=>{"0.3.2"=>[{"has_sig"=>false, "upload_time"=>"2015-06-08T11:52:13", "comment_text"=>"", "python_version"=>"source", "url"=>"https://pypi.python.org/packages/source/c/cufflinks/cufflinks-0.3.2.tar.gz", "md5_digest"=>"818dd8967156698fdc8b62da02900aab", "downloads"=>3050, "filename"=>"cufflinks-0.3.2.tar.gz", "packagetype"=>"sdist", "size"=>15136}], "0.6.1"=>[{"has_sig"=>false, "upload_time"=>"2015-08-17T20:54:05", "comment_text"=>"", "python_version"=>"source", "url"=>"https://pypi.python.org/packages/source/c/cufflinks/cufflinks-0.6.1.tar.gz", "md5_digest"=>"19bd13762a5b06d1c115eb03b7d7397d", "downloads"=>0, "filename"=>"cufflinks-0.6.1.tar.gz", "packagetype"=>"sdist", "size"=>28528}], "0.2"=>[{"has_sig"=>false, "upload_time"=>"2015-02-14T19:18:40", "comment_text"=>"", "python_version"=>"source", "url"=>"https://pypi.python.org/packages/source/c/cufflinks/cufflinks-0.2.tar.gz", "md5_digest"=>"68e60ef6f95c83d9b9be14f1bb345d52", "downloads"=>612, "filename"=>"cufflinks-0.2.tar.gz", "packagetype"=>"sdist", "size"=>6932}], "0.5"=>[{"has_sig"=>false, "upload_time"=>"2015-07-24T19:55:46", "comment_text"=>"", "python_version"=>"source", "url"=>"https://pypi.python.org/packages/source/c/cufflinks/cufflinks-0.5.tar.gz", "md5_digest"=>"0667054349dc75031a324cab641cc7f2", "downloads"=>393, "filename"=>"cufflinks-0.5.tar.gz", "packagetype"=>"sdist", "size"=>22902}], "0.4"=>[{"has_sig"=>false, "upload_time"=>"2015-07-18T21:09:10", "comment_text"=>"", "python_version"=>"source", "url"=>"https://pypi.python.org/packages/source/c/cufflinks/cufflinks-0.4.tar.gz", "md5_digest"=>"d389e2ecbe655bf3bfd538f4176240c2", "downloads"=>322, "filename"=>"cufflinks-0.4.tar.gz", "packagetype"=>"sdist", "size"=>21465}], "0.6"=>[{"has_sig"=>false, "upload_time"=>"2015-08-17T20:19:48", "comment_text"=>"", "python_version"=>"source", "url"=>"https://pypi.python.org/packages/source/c/cufflinks/cufflinks-0.6.tar.gz", "md5_digest"=>"813b4c9842e483d48870945d38dfee5f", "downloads"=>0, "filename"=>"cufflinks-0.6.tar.gz", "packagetype"=>"sdist", "size"=>28499}], "0.1"=>[], "0.3"=>[{"has_sig"=>false, "upload_time"=>"2015-05-05T01:38:10", "comment_text"=>"", "python_version"=>"source", "url"=>"https://pypi.python.org/packages/source/c/cufflinks/cufflinks-0.3.tar.gz", "md5_digest"=>"bcfa36336c29170a31b4dc7c3928e4ba", "downloads"=>946, "filename"=>"cufflinks-0.3.tar.gz", "packagetype"=>"sdist", "size"=>13552}]}, "urls"=>[{"has_sig"=>false, "upload_time"=>"2015-08-17T20:54:05", "comment_text"=>"", "python_version"=>"source", "url"=>"https://pypi.python.org/packages/source/c/cufflinks/cufflinks-0.6.1.tar.gz", "md5_digest"=>"19bd13762a5b06d1c115eb03b7d7397d", "downloads"=>0, "filename"=>"cufflinks-0.6.1.tar.gz", "packagetype"=>"sdist", "size"=>28528}]}
```

The trouble seems to be that puppet's http API doesn't allow the caller to specify they want to accept the default set of cacerts. To do that you need access to the underlying ruby connection `Net::HTTP` so you can do:

```
http.cert_store.set_default_paths
```

But callers to puppet API don't have access to the `Net::HTTP` object, as it's only in scope once we've created a new connection (or borrowed one from the pool).

We should probably file a ticket to make that possible, so that the call to `OpenSSL::X509::Store#set_default_paths` happens automatically.

branan: The third option is to use net::http directly (without puppet's HTTP pools) and pass in the values from the `puppet::util::http_proxy` class, just as `network::http::factory` does. That avoids fighting with puppet's CA mess (at the expense of some duplicate code).

But realistically we already have several providers that need to convert the proxy info into new formats for the programs they execute, so it doesn't seem that evil to do it here.

branan: Trying to see what the PMT does might be a good idea - it's the place in-repo where we use both the proxy and have to use the system certs instead of the puppet CA

joshcooper: I filed https://tickets.puppetlabs.com/browse/PUP-5069 to update the HTTP API. In the meantime, I agree with @branan that following the [PMT example](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/forge/repository.rb#L135-L142) might be the way to go for now. It avoids the issue because it creates its own `Net::HTTP` objects and calls `OpenSSL::X509::Store#set_default_paths` directly. You'll also want to apply the http connect and read timeouts.

branan:
@nibalizer Do you think this is something you can get into shape per our comments above? If not we'd probably be willing to take it over, since this is clearly a Real Bug(tm)

nibalizer:
@branan yea if you guys want to run with it please be my guest. Sorry I haven't been able to get the spec tests into shape, although it sounds like the implementation is going to get reworked. How is the best way to give you push access to the PR? Do you wanna just close this and(you) create a new one?

whopper:
@nibalizer cool - we'll add this to our triage bucket and talk about it at our next planning meeting. Closing this PR for now.

PR #3986, opened 2015-06-01
(PUP-4648) Fix EPP trimming/indentation problems

hlindberg:
Before this, a right trim would replace the whitespace on each following
line up to the next tag instead of just trimming the space on the same
line as the ending tag.

A similar problem occured for a left trim, as it removed the line ending
of the previous line.

Fixing these issues means more strictness wrt which whitespace is
trimmed by each rule.
* A right trim rule trims at most to (including) a line break, stopping
on the first non whitespace.
* A left trim rule troms at most to the beginning of the same line as
the opening (left trimming tag).

This has consequences for comment tags (as they cannot start with a left
trim). Instead, they are now always made left trimming. This is the most
reasonable thing to do since they otherwise needs to be placed
immediately adjacent to a tag or text on a line, cannot be indented etc.

### DIFF
--- a/spec/unit/pops/parser/lexer2_spec.rb
+++ b/spec/unit/pops/parser/lexer2_spec.rb
@@ -535,7 +535,7 @@ describe 'Lexer2' do
       [:VARIABLE, "x"],
       :EQUALS,
       [:NUMBER, "10"],
-      [:RENDER_STRING, "just text\n"]
+      [:RENDER_STRING, "      just text\n"]
       )
     end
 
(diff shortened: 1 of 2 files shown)


spec/unit/pops/parser/lexer2_spec.rb
```diff
@@ -551,7 +551,39 @@ def expect_issue(code, issue)
...
+      )
+    end
+
+    it 'epp comments strips left whitespace when preceding is right trim' do
+      code = <<-CODE
+      This is <% $x=10 -%>
+      space-before-me-but-not-after   <%# This is an epp comment %>
+      just text
+      CODE
+      epp_tokens_scanned_from(code).should match_tokens2(
```
thallgren: Please use 'expect' syntax and not 'should' on new tests as the latter creates deprecation warnings when merged to stable and master.

spec/unit/pops/parser/lexer2_spec.rb
```diff
@@ -551,7 +551,39 @@ def expect_issue(code, issue)
...
+      )
+    end
+
+    it 'epp comments strips left whitespace on same line when preceding is not right trim' do
+      code = <<-CODE
+      This is <% $x=10 %>
+      <%# This is an epp comment -%>
+      just text
+      CODE
+      epp_tokens_scanned_from(code).should match_tokens2(
```
thallgren: 'expect', not 'should'

PR #3987, opened 2015-06-01
(PUP-3774) Only apply necessary group membership changes

branan:
once using usermod. This would cause problems when a user had existing
groups coming from AD, which would not be valid as arguments to
usermod.

existing groups when adding a user to a new group.

valid operation.


branan:
I probably need to retarget this at stable now, but I'll let the specs finish running first

branan:
AppVeyor failure is the rdoc stuff

PR #3990, opened 2015-06-01
(PUP-4356) Remove puppetversion from PMT metadata

andersonmills:
Before this commit, there was an undocumented field, puppetversion, which was tracked by the Puppet::Module class.  An error would be thrown if the metadata value was present and was not an exact string match to Puppet.version.  This commit removes the recognition and checking of that metadata key:value pair.

Note: Accessors and validation for puppetversion still exist in order to not cause a breaking change, but they return nil.

### DIFF
--- a/lib/puppet/module.rb
+++ b/lib/puppet/module.rb
@@ -56,7 +56,7 @@ class Puppet::Module
   attr_writer :environment
 
   attr_accessor :dependencies, :forge_name
-  attr_accessor :source, :author, :version, :license, :puppetversion, :summary, :description, :project_page
+  attr_accessor :source, :author, :version, :license, :summary, :description, :project_page
 
   def initialize(name, path, environment)
     @name = name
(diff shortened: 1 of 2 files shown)


lib/puppet/module.rb
```diff
@@ -67,11 +67,23 @@ def initialize(name, path, environment)
 
     load_metadata if has_metadata?
 
-    validate_puppet_version
-
     @absolute_path_to_manifests = Puppet::FileSystem::PathPattern.absolute(manifests)
   end
 
+  # This is here because puppetversion as module data is deprecated.
+  def puppetversion
```
adrienthebo: Since this is nonfunctional and should probably be removed in the next breaking release, this should be documented as deprecated with something like `# @deprecated The puppetversion module metadata field is no longer used`

andersonmills:
This PR replaces PR3793.

adrienthebo:
:+1:

PR #3992, opened 2015-06-02
(PUP-4521) Add configured_environment to Node Facts find

MikaelSmith:
Prior to this commit, there was no way to write an external node
classifier that allows some agents - based on Facts - to override the
node classifier with its own requested environment.

The first request from the agent for the environment will not use the
correct Facts (that might need to include custom facts), which may end
up classifying it in the incorrect environment. Subsequent requests
will send that environment instead of the configured environment.

Fix by passing the configured environment to the Node::Facts find
request as a new option.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -161,6 +161,7 @@ class Puppet::Configurer
         begin
           if node = Puppet::Node.indirection.find(Puppet[:node_name_value],
               :environment => Puppet::Node::Environment.remote(@environment),
+              :configured_environment => Puppet[:environment],
               :ignore_cache => true,
               :transaction_uuid => @transaction_uuid,
               :fail_on_404 => true)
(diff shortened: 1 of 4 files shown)


kylog:
:+1: 

/cc @pcarlisle

lib/puppet/configurer.rb
```diff
@@ -161,6 +161,7 @@ def run_internal(options)
         begin
           if node = Puppet::Node.indirection.find(Puppet[:node_name_value],
               :environment => Puppet::Node::Environment.remote(@environment),
+              :configured_environment => Puppet[:environment],
```
joshcooper: IIUC, `environment` and `configured_environment` are always equal at this point? I say because the Configurer initializer sets `@environment = Puppet[:environment]`. That said, I think it's good to be explicit and always send the param when making the node request.

MikaelSmith: Yes, they should be equal here.

joshcooper: If the `environment` and `configured_environment` parameters are always the same, then I don't understand why we're doing this?

MikaelSmith: They're the same here, later calls they aren't. I added it because I was testing with different logic in the node terminus, and it made sense to me to be explicit, but in this case we could drop it and reduce network activity slightly.

joshcooper: Ah, yep, I forgot the compiler terminus also calls `Puppet::Node.indirection.find`, and in that case, the environments can be different. :+1: for being explicit.

PR #3993, opened 2015-06-03
(PUP-4663) Ensure Fedora uses the systemd service provider

strider:
- All supported releases of Fedora have systemd and the latest release
  with upstart has been EOLed three years ago. So we could drop the
  version restriction now.

Signed-off-by: Gael Chamoulaud <gchamoul@redhat.com>


strider:
@kylo, @peterhuene Done into 4.2.0!

kylog:
Thanks @strider !

PR #3995, opened 2015-06-03
(PUP-3774) Only apply minimal membership changes in useradd provider

branan:
once using usermod. This would cause problems when a user had existing
groups coming from AD, which would not be valid as arguments to
usermod.

existing groups when adding a user to a new group.

valid operation.

### DIFF
--- a/lib/puppet/property/list.rb
+++ b/lib/puppet/property/list.rb
@@ -1,4 +1,5 @@
 require 'puppet/property'
+require 'set'
 
 module Puppet
   class Property
@@ -46,6 +47,20 @@ module Puppet
         members = add_should_with_current(members, retrieve) if ! inclusive?
 
         dearrayify(members)
+      end
+
+      # Returns any values from the list that were returned by
+      # retrieve but are not set as being managed.
+      def is_but_shouldnt
+        # We know that if inclusive is not set, we won't have to
+        # remove any values from the system.
+        inclusive? ? Set.new(retrieve) - Set.new(@should) : Set.new()
+      end
+
+      # Returns any values from the list that we are managing but that
+      # were not returned by retrieve
+      def should_but_isnt
+        Set.new(@should) - Set.new(retrieve)
       end
 
       def delimiter
--- a/lib/puppet/provider/nameservice.rb
+++ b/lib/puppet/provider/nameservice.rb
@@ -89,19 +89,11 @@ class Puppet::Provider::NameService < Puppet::Provider
       end
     end
 
-    # This awful hack of passing `myself` is brought to you by the
-    # fact that this method isn't marked private, so I can't move it
-    # to be an instance method like I'd rather do - BER
-    def validate(name, value, myself=nil)
+    def validate(name, value)
       name = name.intern if name.is_a? String
       if @checks.include? name
         block = @checks[name][:block]
-        if (myself)
-          result = myself.instance_exec(value, &block)
-        else
-          result = block.call(value)
-        end
-        raise ArgumentError, "Invalid value #{value}: #{@checks[name][:error]}" unless result
+        raise ArgumentError, "Invalid value #{value}: #{@checks[name][:error]}" unless block.call(value)
       end
     end
 
--- a/lib/puppet/provider/nameservice.rb
+++ b/lib/puppet/provider/nameservice.rb
@@ -89,11 +89,19 @@ class Puppet::Provider::NameService < Puppet::Provider
       end
     end
 
-    def validate(name, value)
+    # This awful hack of passing `myself` is brought to you by the
+    # fact that this method isn't marked private, so I can't move it
+    # to be an instance method like I'd rather do - BER
+    def validate(name, value, myself=nil)
       name = name.intern if name.is_a? String
       if @checks.include? name
         block = @checks[name][:block]
-        raise ArgumentError, "Invalid value #{value}: #{@checks[name][:error]}" unless block.call(value)
+        if (myself)
+          result = myself.instance_exec(value, &block)
+        else
+          result = block.call(value)
+        end
+        raise ArgumentError, "Invalid value #{value}: #{@checks[name][:error]}" unless result
       end
     end
 
--- a/lib/puppet/provider/user/useradd.rb
+++ b/lib/puppet/provider/user/useradd.rb
@@ -9,7 +9,11 @@ Puppet::Type.type(:user).provide :useradd, :parent => Puppet::Provider::NameServ
     install Ruby's shadow password library (often known as `ruby-libshadow`)
     if you wish to manage user passwords."
 
+  # These commands have special names for use with the base ObjectAdd provider
   commands :add => "useradd", :delete => "userdel", :modify => "usermod", :password => "chage"
+
+  # This is used in our local implementation and has no special name
+  commands :gpasswd => "gpasswd"
 
   options :home, :flag => "-d", :method => :dir
   options :comment, :method => :gecos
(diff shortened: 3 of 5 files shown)


lib/puppet/provider/nameservice.rb
```diff
@@ -261,6 +269,10 @@ def groups
     groups.join(",")
   end
 
+  def groups
+    @groups ||= fetch_groups
```
MikaelSmith: Is this just a performance optimization?

branan: See commit 0c8da753e0909720be84a3b423dd1eb0d8e402b4. The list property can end up querying the provider several times now.

I didn't want to add the caching to the list property, since other property types don't do caching, and caching in providers isn't all that unusual.

lib/puppet/provider/nameservice.rb
```diff
@@ -89,11 +89,19 @@ def section
       end
     end
 
-    def validate(name, value)
+    # This awful hack of passing `myself` is brought to you by the
+    # fact that this method isn't marked private, so I can't move it
+    # to be an instance method like I'd rather do - BER
+    def validate(name, value, myself=nil)
```
MikaelSmith: Would it be terrible to copy it as an instance method, with a note to remove this for Puppet 5?

MikaelSmith:
Are there acceptance tests covering this?

MikaelSmith:
I'd like to see a test that tries to resemble the ticket issue; i.e. a unit test that stubs returning existing groups with spaces.

lib/puppet/property/list.rb
```diff
@@ -38,14 +38,34 @@ def dearrayify(array)
...
+        # We know that if inclusive is not set, we won't have to
+        # remove any values from the system.
+        inclusive? ? Set.new(retrieve) - Set.new(@should) : Set.new()
+      end
 
-        dearrayify(members)
+      # Returns any values from the list that we are managing but that
+      # were not returned by retrieve
+      def should_but_isnt
+        Set.new(@should) - Set.new(retrieve)
```
joshcooper: I haven't looked closely, but have we changed the semantics of the list property? I ask because it's used by several types.

spec/unit/provider/user/useradd_spec.rb
```diff
@@ -443,4 +444,13 @@
     end
   end
 
+  describe "#groups=" do
```
branan: @MikaelSmith Here's the test you asked for

lib/puppet/property/list.rb
branan: Nope. All the existing methods continue to behave as they used to. The new methods are the only ones that provider the full-blown Set semantics

spec/unit/provider/user/useradd_spec.rb
MikaelSmith: :+1:

MikaelSmith:
This should be targeted at stable.

lib/puppet/property/list.rb
MikaelSmith: Deconstructing #should into another method makes this confusing to review.

branan: I don't think that's used at this point (it was for an earlier version of the code), so I'll back that out to how it was.

lib/puppet/provider/nameservice.rb
MikaelSmith: Other than this question - whether it makes sense to create a new instance method in preparation for deprecating the class method - this PR makes sense to me.

joshcooper: This is a bit too clever for my liking, more comments below.

lib/puppet/provider/user/useradd.rb
```diff
@@ -9,8 +9,12 @@
     install Ruby's shadow password library (often known as `ruby-libshadow`)
     if you wish to manage user passwords."
 
+  # These commands have special names for use with the base ObjectAdd provider
   commands :add => "useradd", :delete => "userdel", :modify => "usermod", :password => "chage"
 
+  # This is used in our local implementation and has no special name
+  commands :gpasswd => "gpasswd"
```
joshcooper: This will make the provider unsuitable on any system that doesn't have gpasswd in the path, will that be a problem?

lib/puppet/provider/user/useradd.rb
```diff
@@ -89,8 +93,39 @@ def shell=(value)
...
+  verify :groups, "Groups must not contain spaces" do |value|
+    prop = @resource.parameters[:groups]
+    changes = prop.should_but_isnt + prop.is_but_shouldnt
+    changes.all? do |change|
+      change !~ /\s/
+    end
+  end
+
+  def groups=(value)
+    self.class.validate(:groups, value, self)
```
joshcooper: The `verify` check is yielded a `value`, but the thing we verify is a bit different (the list of groups split on ','). Also if the check fails, we say "Invalid value group,another group: Groups must not contain spaces" instead of the individual group name that had the issue.

I think it would be cleaner to check for whitespace when the `groups=` method is called. Since the method is already collecting the set of current and desired values, and enumerating them, we could issue an error message for the specific group:

```
def groups=(value)
  prop = @resource.parameters[:groups]
  ...
  changes.each do |group|
    self.fail("Invalid value '#{group[:name]': Group must not contain spaces") if group[:name] =~ /\s/
    md = [ command(:gpasswd),
      group[:name],
      group[:action],
     @resource[:name] ]
    ...
  end
```

lib/puppet/provider/user/useradd.rb
branan: AFAIK gpasswd should exist on all systems that have useradd/usermod, but I'll go through some really new and really old ones to see if it's an optional binary anywhere.

If so I can fixup `groups=` to fall back to the old usermod behavior if it doesn't exist (with the caveat that the bug will still exist in that case)

lib/puppet/provider/user/useradd.rb
branan: All of this is a mess. I'm bypassing the `value` param pretty much everywhere because it's the comma-joined `should` that comes from the list property.

As for the verify/validate bit, I was on the fence between doing what you suggested and trying to map my changes back to use as much of the nameservice base provider's existing functionality. I stuck with this because I wanted to make sure we kept the same error message format. I'm happy to rip out the verify/validate indirection and just throw an exception here if you don't think that's a concern.

lib/puppet/provider/nameservice.rb
branan: This is the part I really hate.

The more I look at this the more I think we can toss it - it made way more sense when I was working on a different approach for this (where we were still going through the nameservice-generated setter). Now that I own the whole set/validation process in the useradd provider, I don't really need this anymore.

lib/puppet/provider/user/useradd.rb
branan: gpasswd exists on centos 4 and on ubuntu 1504. I'm inclined to believe it's universal.

branan:
@MikaelSmith I'll retarget at stable once this is ready to merge, so we don't lose existing conversations.

branan:
@joshcooper My implementation for group error handling is slightly different from your suggestion - it skips ALL group changes if any single one would fail validation, which matches prior behavior.

lib/puppet/property/list.rb
```diff
@@ -48,6 +48,20 @@ def should
...
+      def is_but_shouldnt
+        # We know that if inclusive is not set, we won't have to
+        # remove any values from the system.
+        inclusive? ? Set.new(retrieve) - Set.new(@should) : Set.new()
+      end
+
+      # Returns any values from the list that we are managing but that
+      # were not returned by retrieve
+      def should_but_isnt
+        Set.new(@should) - Set.new(retrieve)
```
joshcooper: Probably should explicitly `require 'set'`.

```
irb(main):001:0> Set.new
NameError: uninitialized constant Set
	from (irb):1
	from :0
irb(main):002:0> require 'set'
=> true
irb(main):003:0> Set.new
=> #<Set: {}>
```

branan: Done

lib/puppet/provider/nameservice.rb
joshcooper: I'd rather not add caching to the provider at this late stage. The original ticket is about groups from AD containing spaces, and I'm thinking the least risky thing is to only validate that.

branan:
@joshcooper squashed the requested commit

lib/puppet/provider/user/useradd.rb
joshcooper: Only took 6 years to implement: https://projects.puppetlabs.com/issues/1942

branan: Should I add that ticket number to the commit? ;)

lib/puppet/provider/nameservice.rb
branan: I was concerned that retrieving groups thrice would end up being slow, especially in the case when a direcotry service is involved... We can definitely pull out the caching commit and see if anybody complains, though :)

lib/puppet/provider/user/useradd.rb
joshcooper: And now that I write that I don't understand why we're switching from usermod to gpasswd as part of this ticket?

branan: For the exact same reason as ticket 1942 - the user in the case of the ticket (PUP-3774) is coming from AD, so doesn't have an entry in /etc/passwd. We need to use gpasswd in this case.

I can expand the commit message with that note.

joshcooper: Ah yeah that makes sense, can you change the title of the PUP-3774 ticket too? There's definitely more to this than "puppet fails to manipulate user's group membership if group name has space in it"

joshcooper:
I'd prefer to not cache and see if there's an issue later. Could you yank https://github.com/branan/puppet/commit/20235f176fb8363ed591f918d2eaaacb4087830b, and then retarget for stable?

branan:
New PR at https://github.com/puppetlabs/puppet/pull/4021

PR #3996, opened 2015-06-03
(PUP-2638) Use settings during puppet apply setup

joshcooper:
Previously, if you installed puppet, and never executed `puppet agent`,
but immediately executed `puppet apply` (such as with vagrant
workflows), and `graph` was true (as is the default in PE), then
puppet would fail to write the graph. This occurred because `puppet
apply` tries to write the graph (during `Catalog#to_ral`) before
creating the necessary settings related directories (as a side-effect
of calling `Puppet::Configurer.new`).

The `agent` and `device` applications also use the Configurer to apply
the catalog, but they explicitly use the `main`, `ssl`, and `agent`
sections during application setup, and before calling `Catalog#to_ral`.

This commit modifies `apply` to match the other two. Alternatively, the
setup logic could be added to the base `Puppet::Application#setup`, but
it's more complex due to run_mode.

This commit adds an acceptance test, which uses a temporary vardir to
ensure `statedir` and `graphdir` do not exist. It also adds a debug
message to print which settings-related sections are applied.

### DIFF
--- a/acceptance/tests/apply/puppet_apply_graph.rb
+++ b/acceptance/tests/apply/puppet_apply_graph.rb
@@ -1,0 +1,15 @@
+test_name 'puppet apply should generate a graph'
+
+agents.each do |agent|
+  step "Create var temp directory"
+  vardir = agent.tmpdir('vardir')
+  graphdir = File.join(vardir, 'state', 'graphs')
+
+  step "Ensure it creates parent directories and generates the graph"
+  on(agent, puppet("apply --graph --graphdir #{graphdir} --vardir #{vardir} -e 'notify { \"hi\": }'")) do
+    resources_dot = File.join(graphdir, 'resources.dot')
+    if !agent.file_exist?(resources_dot)
+      fail_test("Failed to create graph: #{resources_dot}")
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


MikaelSmith:
:+1:

MikaelSmith:
I see
```
xnwreh0kv3ub1vo.delivery.puppetlabs.net (agent) 09:30:08$ puppet apply --graph --graphdir /tmp/vardir.IF0b2x/state/graphs --vardir /tmp/vardir.IF0b2x -e 'notify { "hi": }'
Could not retrieve macaddress: undefined method `each_line' for nil:NilClass
Could not retrieve macaddress: undefined method `each_line' for nil:NilClass
Could not retrieve macaddress: undefined method `each_line' for nil:NilClass
Could not retrieve ipaddress6: undefined method `scan' for nil:NilClass
Notice: Compiled catalog for xnwreh0kv3ub1vo.delivery.puppetlabs.net in environment production in 0.43 seconds
Notice: hi
Notice: /Stage[main]/Main/Notify[hi]/message: defined 'message' as 'hi'
```
running on RedHat 7. Going to look into that a little bit.

MikaelSmith:
Somehow I'm getting `Facter 2.0.0-rc4` installed in acceptance tests.

MikaelSmith:
Oh, because I used FORK=joshcooper. Looks like a non-issue.

PR #3997, opened 2015-06-03
(maint) Default to facter#2.x for git based acceptance

joshcooper:
Previously, git based acceptance (rake ci:test:git) defaulted to
facter#stable, but that branch was recently promoted to native facter,
so acceptance would fail trying to install facter `ruby install.rb`

This commit changes the default facter branch to 2.x, containing the
legacy ruby facter implementation. It also relaxes the regex to match
digits and a literal dot, when translating short branch names,
facter#2.x, to git URLs.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -85,7 +85,7 @@ def beaker_test(mode = :packages, options = {})
   if mode == :git
     # Build up project git urls based on git server and fork env variables or defaults
     final_options[:install].map! do |install|
-      if md = /^(\w+)#(\w+)$/.match(install)
+      if md = /^(\w+)#([\d.\w]+)$/.match(install)
         project, project_sha = md.captures
         "#{build_giturl(project)}##{project_sha}"
       elsif md = /^(\w+)$/.match(install)
--- a/acceptance/config/git/options.rb
+++ b/acceptance/config/git/options.rb
@@ -1,6 +1,6 @@
 {
   :install => [
-    'facter#stable',
+    'facter#2.x',
     'hiera#stable',
     'puppet',
   ],


joshcooper:
```
$ be rake ci:test:git SHA=3.8.1 TESTS=tests/ticket_7101_template_compile.rb CONFIG=config/nodes/rhel7.yaml
...
tests/ticket_7101_template_compile.rb passed in 2.98 seconds
...
0 ~/work/puppet/acceptance (maint/3.x/git-acceptance-facter-2.x)  $ cat merged_options.rb
...
 :install=>
  ["git://github.com/puppetlabs/facter.git#2.x",
   "git://github.com/puppetlabs/hiera.git#stable",
   "git://github.com/puppetlabs/puppet.git#3.8.1"],
```

acceptance/Rakefile
```diff
@@ -85,7 +85,7 @@ def beaker_test(mode = :packages, options = {})
   if mode == :git
     # Build up project git urls based on git server and fork env variables or defaults
     final_options[:install].map! do |install|
-      if md = /^(\w+)#(\w+)$/.match(install)
+      if md = /^(\w+)#([\d.\w]+)$/.match(install)
```
MikaelSmith: Will this continue to work if we move back to `stable`? Can we make a regex that'll match any branch name?

joshcooper: > Will this continue to work if we move back to stable?

Yep.

```
irb(main):001:0> /^(\w+)#([\d.\w]+)$/.match('facter#stable')
=> #<MatchData "facter#stable" 1:"facter" 2:"stable">
```

> Can we make a regex that'll match any branch name?

Based on https://www.kernel.org/pub/software/scm/git/docs/git-check-ref-format.html, we will incorrectly match some things, e.g. `..`, `foo.`, and won't match when we should, e.g. `a+b`, `foo/bar`. The authoritative way seems to be to shell out to `git check-ref-format --branch <name>`, but that seems like overkill.

MikaelSmith: Ok. Also missed the square brackets, so this looks fine.

PR #3998, opened 2015-06-03
(PUP-4522) Set agent_specified_environment

MikaelSmith:
Creates an `agent_specified_environment` fact with the value being the
environment name if the environment was set on the agent by the user.

### DIFF
--- a/lib/puppet/settings.rb
+++ b/lib/puppet/settings.rb
@@ -749,13 +749,13 @@ class Puppet::Settings
 
   class SearchPathElement < Struct.new(:name, :type); end
 
-  # The order in which to search for values.
+  # The order in which to search for values, without defaults.
   #
   # @param environment [String,Symbol] symbolic reference to an environment name
   # @param run_mode [Symbol] symbolic reference to a Puppet run mode
   # @return [Array<SearchPathElement>]
   # @api private
-  def searchpath(environment = nil, run_mode = preferred_run_mode)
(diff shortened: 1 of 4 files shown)


spec/integration/indirector/facts/facter_spec.rb
```diff
@@ -78,4 +78,32 @@
...
+                         Puppet::Node.indirection.find('foo'))
+    end.to raise_error(Puppet::PreformattedError)
+  end
+
+  it "the agent_specified_environment fact is nil when set in puppet.conf" do
+    FileUtils.mkdir_p(Puppet[:confdir])
+    File.open(File.join(Puppet[:confdir], 'puppet.conf'), 'w') do |f|
+      f.puts("environment=bar")
+    end
+    Puppet.settings.initialize_global_settings
```
MikaelSmith: This looks concerning to me, but I'm not sure what other way to do it.

lib/puppet/settings.rb
```diff
@@ -799,6 +809,27 @@ def set_by_cli?(param)
     !@value_sets[:cli].lookup(param).nil?
   end
 
+  # Allow later inspection to determine if the setting was set by user
+  # config, rather than a default setting.
+  def set_by_config?(param)
```
MikaelSmith: The horror...

MikaelSmith: I plan to clean it up a little so we don't duplicate the `source` handling.

PR #4001, opened 2015-06-04
(maint) Ensure acceptance fixtures are removed

joshcooper:
Previously, if the functions_in_puppet_language test was run prior to
the module acceptance tests, the former would leave behind modules that
caused the latter to fail.

This commit adds a teardown to remove acceptance fixtures.


kylog:
Dup of #4005

PR #4003, opened 2015-06-04
(PUP-4693) Confine the pw user and group providers to BSD (Puppet 3.x)

Sharpie:
These providers were designed to function on systems similar to FreeBSD. If a
`pw` binary shows up on Linux, the providers will be activated and Puppet's
ability to manage users and groups will break.


joshcooper:
looks good, will look at next sprint

PR #4004, opened 2015-06-04
(PUP-4693) Confine the pw user and group providers to BSD (Puppet 4.x)

Sharpie:
These providers were designed to function on systems similar to FreeBSD. If a
`pw` binary shows up on Linux, the providers will be activated and Puppet's
ability to manage users and groups will break.


joshcooper:
@Sharpie thanks for the PRs! A few weeks or so ago, we straightened out our puppet branches to follow the normal 3.x->stable->master flow. So I'll merge PR #4003 targeted at 3.x and close this one.

Sharpie:
:+1: Thanks for the update on the workflow!

PR #4007, opened 2015-06-05
(PUP-4702) Replace RGEN based Type system with immutable classes

thallgren:
This commit changes all RGEN based classes in the Puppet Type
system into plain Ruby classes. Each class is immutable which means
that the instance is completely initialized by its initializer and
that only attr_readers exists to read the properties of the instance.

Each class has a `DEFAULT` constant defined which is the singleton
default for the type that the class represents. Some types have
additional constants such as `PCollectionType::DEFAULT_SIZE` or
`PStringType::NON_EMPTY`.

In addition to creating immutable classes, some performance critical
methods were also moved out of the polymorphic dispatch. This includes
methods like `assignable?`, `instance?` and `generalize`. The methods
are still available on the TypeCalculator to ensure backward
compatibility.


hlindberg:
I am fine with having a runtime implementation of types that is optimized like this. The intent though is to be able to model types and serialize them. We currently have no need to do so in the AST (we build types by evaluating expressions that mean "construct a type"), but we will most likely need this when we also allow users to define type aliases, and to define complete types.

In order to support that, we need to either have a model of the types and serialize that, or make it possible to use types as a data type in RGen models.

My original idea was to build one type model out of all types in a module; the server and client would share this model. 

With that in mind; I don't know how valuable the type_meta model is - if it should be kept and evolved, if it is meaningful to be able to go from runtime object to model, and vice versa.

When we get to the implementation of resource types (and other types described by users), they have to also fit into the runtime model.

A general note is that it is a shame that we cannot get the performance out of an RGen model and that we instead start to hand code model after model. There will be more models. There will be polyglot implementations (representing models in different implementation languages).

At the same time, this is a tactical performance optimization (and work around for deficiencies in RGen that are more time consuming to correct).

lib/puppet/pops/binder/bindings_model_meta.rb
```diff
@@ -94,7 +94,7 @@ class ProducerProducerDescriptor < ProducerDescriptor
   # @api public
   #
   class LookupProducerDescriptor < ProducerDescriptor
-    contains_one_uni 'type', Puppet::Pops::Types::PAnyType
+    has_attr 'type', Object
```
hlindberg: While this is ok for now, an Object type cannot be serialized/deserialized. Eventually it needs to be fixed or it will be impossible to for a non Ruby model creator to produce an instance of this model. It does not matter at the moment because there are no such use-cases.

It is a technical debt though and deserves a comment like '# TODO: Use real type, not Object'

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -313,240 +225,111 @@ def equals(left, right)
...
 
-  # Generalizes value specific types. The given type is mutated and returned.
+  # Generalizes value specific types. The generalized type is returned.
   # @api public
-  def generalize!(o)
-    @@generalize_visitor.visit_this_0(self, o)
-    o.eAllContents.each { |x| @@generalize_visitor.visit_this_0(self, x) }
+  def generalize(o)
+    o = o.generalize if o.is_a?(Types::PAnyType)
     o
```
hlindberg: Slightly more efficient to do: `o.is_a?(Type::PAnyType) ? o.generalize : o` as it avoids the assignment to o

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -313,240 +225,111 @@ def equals(left, right)
...
-  end
-
   # Answers 'what is the single common Puppet Type describing o', or if o is an Array or Hash, what is the
   # single common type of the elements (or keys and elements for a Hash).
   # @api public
   #
   def infer(o)
-    @@infer_visitor.visit_this_0(self, o)
+    case o
+      when String
```
hlindberg: Suggest comment that this is an optimization for most common cases (at least I think this is why it avoids calling the visitor)

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -582,115 +365,86 @@ def common_type(t1, t2)
...
+      return Types::PHostClassType::DEFAULT
     end
 
     # when both are resources, reduce to Resource[T] or Resource[] (since one was not assignable to the other)
     if t1.is_a?(Types::PResourceType) && t2.is_a?(Types::PResourceType)
-      result = Types::PResourceType.new()
       # only Resource[] unless the type name is the same
-      if t1.type_name == t2.type_name then result.type_name = t1.type_name end
+      return t1.type_name == t2.type_name ?  Types::PResourceType.new(t1.type_name, nil) : Types::PResourceType::DEFAULT
       # the cross assignability test above has already determined that they do not have the same type and title
```
hlindberg: dangling comment?

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -582,115 +365,86 @@ def common_type(t1, t2)
...
     if t1.is_a?(Types::PFloatType) && t2.is_a?(Types::PFloatType)
-      t1range = from_to_ordered(t1.from, t1.to)
-      t2range = from_to_ordered(t2.from, t2.to)
-      t = Types::PFloatType.new()
-      from = [t1range[0], t2range[0]].min
-      to = [t1range[1], t2range[1]].max
-      t.from = from unless from == Float::INFINITY
-      t.to = to unless to == Float::INFINITY
-      return t
+      return Types::PFloatType.new([t1.from, t2.from].min, [t1.to, t2.to].max)
```
hlindberg: Both this and the corresponding Integer have a flaw (and so did the original). They both expand the range to include every value between min and max. That is not great. It would be better to have a Variant of integer ranges, and to replace a range with the merged if the found range is adjacent, otherwise it is added as a another range in the variant. Finally removing the variant if only containing one integer range. Thus, the result of [1,3,5, 2, 4] would be Integer[1,4], after having been Variant[Integer[1]], ... Variant[Integer[1], Integer[3], Integer[5]]... until '2' is merged, Variant[Integer[1,3], Integer[5]], then '4'.

Not worth doing this in this PR, but worth a ticket.
(OTOH, I could be reading this diff wrong - and what I am talking about is irrelevant because we do want generalization to reduce it to a wider (and "incorrect" range)).

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -756,19 +514,19 @@ def reduce_type(enumerable)
   # @api public
   #
   def infer_and_reduce_type(enumerable)
-    reduce_type(enumerable.collect() {|o| infer(o) })
+    reduce_type(enumerable.collect {|o| infer(o) })
```
hlindberg: prefer use of `map` over `collect`

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -875,64 +618,50 @@ def infer_Symbol(o)
 
   # @api private
   def infer_TrueClass(o)
-    Types::PBooleanType.new()
+    Types::PBooleanType::DEFAULT
```
hlindberg: Note: we should consider making boolean parameterized. (We should log a ticket for that). Reason being you cannot precisely compare booleans with the type system and instead have to use a test on value.

lib/puppet/pops/types/types.rb
```diff
@@ -1,415 +1,1482 @@
...
+      def _assignable?(o)
+        o.is_a?(PAnyType)
+      end
+
+      NAME_SEGMENT_SEPARATOR = '::'.freeze
+
+      # @api private
+      def class_from_string(str)
+        begin
+          str.split(NAME_SEGMENT_SEPARATOR).inject(Object) do |memo, name_segment|
```
hlindberg: prefer ´reduce´ over ´inject´ (i.e. we use map/reduce instead of collect/inject)

lib/puppet/pops/types/types.rb
```diff
@@ -1,415 +1,1482 @@
...
+    # @api public
+    #
     class PEnumType < PScalarType
-      module ClassModule
-        def hash
-          [self.class, Set.new(self.values)].hash
+      attr_reader :values
+
+      def initialize(values)
+        @values = values.sort.freeze
```
hlindberg: this sort makes the source and string form of the type differ - it also makes a difference if an enum is mapped to numeric form as additions can be placed in the middle. I think we an live with those "problems" though.

lib/puppet/pops/types/types.rb
```diff
@@ -1,415 +1,1482 @@
...
+
+      def ==(o)
+        self.class == o.class && @from == o.numeric_from && @to == o.numeric_to
+      end
+
+      def instance?(o)
+        o.is_a?(Numeric) && o >= @from && o <= @to
+      end
+
+      def unbounded?
```
hlindberg: IIRC, elsewhere it is checked that enumeration can take place if value is !unbounded? - an open range at either end is not acceptable for enumeration. Do the numeric type need to have an additional method for being open_ended?

thallgren: Would it make sense to do the opposite and instead have a method named enumerable? and then move that method to PAnyType?

hlindberg: IIRC, enumerable? is the only use case for unbounded? in the first place. It would also be neat to be able to iterate over type parameters - e.g. Tuple[Integer, String] would yield Integer, String, and index would yield 0, Integer, 1, String. An Integer[1,2] would yield something like 'min', 1 and 'max',2. Basically representing themselves as either Array or Hash and then iterating over them. -- All of that is worth a separate ticket.

Regarding enumerable? (I am not sure if that is part of a Ruby protocol, and I am not sure we want to be able to iterate over everything that Ruby says is enumerable? true. 

Seems reasonable to keep unbounded? to check lower and upper bounds at the same time, but I like the idea of enumerable?

thallgren: The enumerable? method is not part of the Ruby protocol. We will use it to help out in case `is_a?(Enumerable)` (which always returns true for a `PIntegerType` instance) doesn't tell the whole truth.

thallgren:
Pushed a rebased commit that I think addresses all issues that we've agreed should be addressed in this PR.

PR #4009, opened 2015-06-05
(PUP-3539) add specs globbing ACL in fileserver.conf

er0ck:
ENTERPRISE-361 reported on regression in fileserver.conf which did not
allow globbing subdomains.  Allow in
file_serving/configuration/parser.rb sets simple comma separated allows
in mount.allow, which inherits from network/authconf allow, where most
of the testing lives. This change adds tests that ensure we are sending
the correct patterns to mount.allow.

### DIFF
--- a/spec/unit/file_serving/configuration/parser_spec.rb
+++ b/spec/unit/file_serving/configuration/parser_spec.rb
@@ -33,16 +33,6 @@ describe Puppet::FileServing::Configuration::Parser do
       expect { @parser.parse }.not_to raise_error
     end
 
-    it "should create a new mount for each section in the configuration" do
-      mount1 = mock 'one', :validate => true
-      mount2 = mock 'two', :validate => true
-      Puppet::FileServing::Mount::File.expects(:new).with("one").returns(mount1)
-      Puppet::FileServing::Mount::File.expects(:new).with("two").returns(mount2)
-      write_config_file "[one]\n[two]\n"
-      @parser.parse
-    end
-
-    # This test is almost the exact same as the previous one.
     it "should return a hash of the created mounts" do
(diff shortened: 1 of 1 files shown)


kurtwall:
Passed on stable and master:
```
[kwall@breve puppet]$ bundle exec rspec -fd spec//unit/file_serving/configuration/parser_spec.rb

Puppet::FileServing::Configuration::Parser
  Puppet::FileServing::Configuration::Parser  when parsing
    should allow comments
    should allow blank lines
    should create a new mount for each section in the configuration
    should return a hash of the created mounts
    should only allow mount names that are alphanumeric plus dashes
    should fail if the value for path/allow/deny starts with an equals sign
    should validate each created mount
    should fail if any mount does not pass validation
    should return comprehensible error message, if invalid line detected
  Puppet::FileServing::Configuration::Parser  when parsing mount attributes
    should set the mount path to the path attribute from that section
    should tell the mount to allow any allow values from the section
    should support inline comments
    should tell the mount to deny any deny values from the section
    should return comprehensible error message, if failed on invalid attribute
  Puppet::FileServing::Configuration::Parser  when parsing the modules mount
    should create an instance of the Modules Mount class
    should warn if a path is set
  Puppet::FileServing::Configuration::Parser  when parsing the plugins mount
    should create an instance of the Plugins Mount class
    should warn if a path is set

Finished in 0.03696 seconds (files took 0.49008 seconds to load)
18 examples, 0 failures
```

ferventcoder:
AppVeyor failures are the rdoc parser

kurtwall:
Merging. Failures are unrelated.

kurtwall:
Merging. Failures are unrelated.

PR #4010, opened 2015-06-08
(WIP) (PUP-4386) Windows group invalid user error msgs

Iristyle:

```puppet
group { 'guests':
  ensure  => 'present',
  gid     => 'S-1-5-32-546',
}
```

```puppet
group { 'guests':
  ensure => present,
}
```

```
```

```
Error: /Stage[main]/Main/Group[Guests]/members: change from VAGRANT-2008R2\Guest,VAGRANT-2008R2\Administrator to ["Guest", "Administrator", ""] failed: Could not resolve username:
```

### DIFF
--- a/lib/puppet/type/group.rb
+++ b/lib/puppet/type/group.rb
@@ -99,7 +99,12 @@ module Puppet
       def is_to_s(currentvalue)
         if provider.respond_to?(:members_to_s)
           currentvalue = '' if currentvalue.nil?
-          return provider.members_to_s(currentvalue.split(','))
+          currentvalue = currentvalue.is_a?(Array) ? currentvalue : currentvalue.split(',')
+
+          begin
+            return provider.members_to_s(currentvalue)
+          rescue Puppet::Util::Windows::Error
+          end
         end
 
         super(currentvalue)
--- a/spec/unit/provider/group/windows_adsi_spec.rb
+++ b/spec/unit/provider/group/windows_adsi_spec.rb
@@ -35,11 +35,13 @@ describe Puppet::Type.type(:group).provider(:windows_adsi), :if => Puppet.featur
     let(:user1) { stub(:account => 'user1', :domain => '.', :to_s => 'user1sid') }
     let(:user2) { stub(:account => 'user2', :domain => '.', :to_s => 'user2sid') }
     let(:user3) { stub(:account => 'user3', :domain => '.', :to_s => 'user3sid') }
+    let(:invalid_user) { SecureRandom.uuid }
 
     before :each do
       Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with('user1').returns(user1)
       Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with('user2').returns(user2)
       Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with('user3').returns(user3)
+      Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with(invalid_user).returns(nil)
     end
 
     describe "#members_insync?" do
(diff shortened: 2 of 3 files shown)


lib/puppet/type/group.rb
```diff
@@ -99,7 +99,12 @@ def insync?(current)
       def is_to_s(currentvalue)
         if provider.respond_to?(:members_to_s)
           currentvalue = '' if currentvalue.nil?
-          return provider.members_to_s(currentvalue.split(','))
+          currentvalue = currentvalue.is_a?(Array) ? currentvalue : currentvalue.split(',')
+
+          begin
+            return provider.members_to_s(currentvalue)
+          rescue Puppet::Util::Windows::Error
+          end
```
ferventcoder: I'm hoping this part is part of the WIP.

spec/unit/provider/group/windows_adsi_spec.rb
```diff
@@ -268,4 +268,10 @@
   it "should prefer the domain component from the resolved SID" do
     expect(provider.members_to_s(['.\Administrators'])).to eq('BUILTIN\Administrators')
   end
+
+  it "should raise an error on a bad username" do
+    user_name = 'b_a_baracus'
```
ferventcoder: I'm not saying I have a user by that name, but you might insert some random characters so this doesn't fail in weird and unusual ways.

Iristyle:
Since we've recently uncovered issues with both groups and users that mirror each other, I took a quick look to see if the same sorts of changes need to be made to `lib/puppet/type/user.rb` / `lib/puppet/provider/user/windows_adsi.rb`.

Of note, https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/user/windows_adsi.rb#L53 has the same `if sid.account` check against what could be a potentially unresolvable SID based on arbitrary string -> group name resolution performed with

```ruby
sid = Puppet::Util::Windows::SID.name_to_sid_object(group_name)
```

We should be able to induce a similar failure with a manifest like:

```puppet
user { 'jimbob':
  ensure => present,
  groups => ['Users', 'Southerners'],
}
```

However, the above code path is never reached - because the `groups` member is run through `name_sid_hash` at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/user/windows_adsi.rb#L21

```ruby
def groups
    @groups ||= Puppet::Util::Windows::ADSI::Group.name_sid_hash(user.groups)
    @groups.keys
  end
```

And `name_sid_hash` at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/adsi.rb#L133-L144 will raise a similar error:

```ruby
  def name_sid_hash(names)
      return {} if names.nil? || names.empty?

      sids = names.map do |name|
        sid = Puppet::Util::Windows::SID.name_to_sid_object(name)
        raise Puppet::Error.new( "Could not resolve name: #{name}" ) if !sid
        [sid.to_s, sid]
      end

      Hash[ sids ]
    end
  end
```

So my question is whether we can implement something in `provider/group/windows_adsi.rb` that allows `name_sid_hash` to propagate the message like we're doing with `user`.  I spiked this and wasn't successful in getting it to work (mostly by changing `members` at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/group/windows_adsi.rb#L56-L58) from 

```ruby
def members
  group.members
end
```

To something like

```ruby
def members
  @members ||= Puppet::Util::Windows::ADSI::User.name_sid_hash(group.members)
  @memberss.keys
end
```

I would be in favor of getting an approach like this working if we can, because then it makes user / group share similar implementation approaches.

lib/puppet/type/group.rb
Iristyle: Nope, its necessary.

ferventcoder: Does it fail specifically by having `Puppet::Util::Windows::Error`? I'm guessing not as that file has been loaded and all is well.

Iristyle:
I do have an alternate implementation that I'll throw up in another PR shortly - though this approach is still probably the correct one.  (Implementing my suggestion above will be difficult without a number of changes that we can't really safely make right now.)

lib/puppet/type/group.rb
Iristyle: Strike my earlier comment.  `Puppet::Util::Windows::Error` loads OK on other platforms, and specs pass across the board.

ferventcoder:
Closed in favor of moving forward with #4024

PR #4011, opened 2015-06-08
(PUP-4631) Introduce Windows 2003 deprecation msg

Iristyle:
Note - permalink doesn't exist yet.

### DIFF
--- a/lib/puppet/feature/base.rb
+++ b/lib/puppet/feature/base.rb
@@ -77,6 +77,7 @@ Puppet.features.add(:manages_symlinks) do
           true
         rescue LoadError => err
           Puppet.debug("CreateSymbolicLink is not available")
+          Puppet.deprecation_warning("Windows 2003 and below versions are deprecated, and will be officially unsupported in the next major Puppet release.  Please plan accordingly and see http://links.puppetlabs.com/deprecate-windows-2003 for more information.")
           false
         end
       end


lib/puppet/feature/base.rb
```diff
@@ -77,6 +77,7 @@ def self.is_implemented
           true
         rescue LoadError => err
           Puppet.debug("CreateSymbolicLink is not available")
+          Puppet.deprecation_warning("Windows 2003 and below versions are deprecated, and will be officially unsupported in the next major Puppet release.  Please plan accordingly and see http://links.puppetlabs.com/deprecate-windows-2003 for more information.")
```
Iristyle: We could break this out into it's own check, but we already know that `CreateSymbolicLinkW` is appropriate feature detection for Windows 2003.

ferventcoder: And when will we create the url behind that link?

ferventcoder: I'd prefer it go out to it's own check.

ferventcoder: Just saw the note about the permalink.

stahnma:
I'm a little afraid this sets expectations that we'd do this for all platforms. We haven't ever before, and I don't know that we'd want to for each linux variant.

Iristyle:
@stahnma I generally agree with you, but I think in this case Windows 2003 is a unique situation.

lib/puppet/feature/base.rb
kylog: It's own check makes more sense to me. E.g. I know nano drops some APIs (maybe not this one), and who knows what might come down the pike next. An explicit version check seems safest.

Also wording wise, some questions:
1) replace "Windows 2003 and below versions are deprecated" with "Support for Windows 2003 and below versions are deprecated"?
2) should we call out 2003r2 as well?
3) it says "next major puppet version". On the foss side does that really mean puppet 5? Or can it be puppet 4.3?

ferventcoder: @kylog agreed. I should have called out better on why like you did.

joshcooper:
Talked with @Iristyle about this. My concern is that we're introducing a deprecation warning that users can't turn off (without turning off all deprecation warnings). Other thoughts include adding logic to the MSI installer. It could be a popup message box during attended install and/or writing to the windows event log.

lib/puppet/feature/base.rb
Iristyle: * I was working from the incorrect assumption that features were lazy-evaluated.  That made me think I was preventing the addition of extra code to call the feature resolution code, but it turns out that's not how it works.  So we can segregate out to a `windows_2003` or similar feature and introduce some redundant code if Puppet is the place to introduce the warning. (I also wanted to make sure I hit a code path that was used by `agent`, `apply`, `config`, etc)
* Text was mostly a placeholder at the moment until we could polish out the messaging.  I should have mentioned that with the PR.  Pointeds noted @kylog - thanks.

So that said, I'm not sure this is the route we want to go given deprecation message noise.  Might be better to skin this cat inside https://github.com/puppetlabs/puppet_for_the_win instead.

ferventcoder:
We still have the rdoc failures with Windows?

ferventcoder:
Let's close this for now.

Iristyle:
Yes, closed in favor of https://github.com/puppetlabs/puppet_for_the_win/pull/114 + docs + email blast, etc.

PR #4015, opened 2015-06-08
Restore check for non-existent path

kylog:


lib/puppet/node/environment.rb
```diff
@@ -318,13 +318,17 @@ def warn_about_mistaken_path(path, name)
   def modules_by_path
     modules_by_path = {}
     modulepath.each do |path|
-      Dir.chdir(path) do
-        module_names = Dir.entries(path).select do |name|
-          Puppet::Module.is_module_directory?(name, path)
-        end
-        modules_by_path[path] = module_names.sort.map do |name|
-          Puppet::Module.new(name, File.join(path, name), self)
+      if File.exists?(path)
```
joshcooper: `Puppet::FileSystem.exists?(path)`?

branan:
There's a race condition here, but that's also in the current PE codebase. Good to hit parity, and we can sort out cleaning this mess up another day.

kylog:
@joshcooper good call; fix pushed.

@branan yeah, I created https://tickets.puppetlabs.com/browse/PUP-4713. There's a lot not to like about that code, and we may reach the point where a PMT rewrite is prefereable to servicing its tech debt.

MikaelSmith:
AppVeyor failure is rdoc.

PR #4016, opened 2015-06-09
(maint) add PE platform configs

justinstoller:
Prior to this there were no acceptance test host config files for PE
only platforms. The distinction between PE and FOSS is largely going
away (save for where the failures are reported).

This commit adds the necessary host configs to test Puppet against all
the platforms we will distribute puppet-agent on in the Shallow Gravy
timeframe.


er0ck:
:+1:

PR #4018, opened 2015-06-09
(PUP-4709) Fix brackets in title problem

hlindberg:


lib/puppet/resource/catalog.rb
```diff
@@ -62,7 +62,7 @@ def add_class(*classes)
 
   def title_key_for_ref( ref )
     s = ref.index('[')
-    e = ref.index(']')
+    e = ref.rindex(']')
```
kylog: Doh!

PR #4019, opened 2015-06-09
(packaging) (re-4770) Update init script to ignore old/incorrect sysconfig values

shrug:
…onfig values

Currently, the EL and SLES init scripts will read variables from /etc/sysconfig/puppet
and override variables set in puppet.conf or the init sctipt. This can lead to
undesired behavior, like an old sysconfig file setting variables incorrectly, causing
startup failures, etc.

This updates the sysconfig files to be more conistent with the defaults file we
ship for Debian, and updates the EL and SUSE init scripts to only use the
PUPPET_EXTRA_OPTS variable if it exists, and to ignore other settings that
are better set in puppet.conf or that should not need to be overridden at
all (like pidfile for example).

### DIFF
--- a/ext/redhat/client.init
+++ b/ext/redhat/client.init
@@ -17,15 +17,12 @@ PATH=/opt/puppetlabs/puppet/bin:/usr/bin:/sbin:/bin:/usr/sbin
 export PATH
 
 [ -f /etc/sysconfig/puppet ] && . /etc/sysconfig/puppet
-lockfile=${LOCKFILE-/var/lock/subsys/puppet}
-pidfile=${PIDFILE-/var/run/puppetlabs/agent.pid}
-puppetd=${PUPPETD-/opt/puppetlabs/puppet/bin/puppet}
+lockfile=/var/lock/subsys/puppet
+pidfile=/var/run/puppetlabs/agent.pid
+puppetd=/opt/puppetlabs/puppet/bin/puppet
 RETVAL=0
 
 PUPPET_OPTS="agent "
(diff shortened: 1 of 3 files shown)


haus:
@shrug looks like you changed some perms here. was that intentional?

shrug:
Nope! Fixed.

melissa:
build with this change going at http://jenkins-staging.delivery.puppetlabs.net/view/Build%20Toolchain/job/pl-build-tools_dynamic/102/

melissa:
what about other projects, like mco? Is this going to be something that needs to happen there as well?

shrug:
@melissa probably, although not in this PR :). I'l prep one for MCO

melissa:
you'll need to add the following as well:

```diff
diff --git a/ext/suse/client.init b/ext/suse/client.init
index c697947..63743fc 100755
--- a/ext/suse/client.init
+++ b/ext/suse/client.init
@@ -35,7 +35,7 @@
 [ -f /etc/sysconfig/puppet ] && . /etc/sysconfig/puppet
 lockfile=/var/lock/subsys/puppet
 pidfile=/var/run/puppetlabs/agent.pid
-puppetd=$/opt/puppetlabs/puppet/bin/puppet
+puppetd=/opt/puppetlabs/puppet/bin/puppet
 RETVAL=0

 PUPPET_OPTS="agent"
```

melissa:
^^ was probably my fault :)

melissa:
:+1: totally on board with this

stahnma:
:+1:

PR #4020, opened 2015-06-09
(PUP-4698) Change fqdn_rand's return type to Integer

hlindberg:
Before this, fqdn_rand always returned a String. That is not the best
choice for a numeric value now that the evaluator is aware of different
data types as comparissons will typically be wrong (since '2' > '19'
when compared leicographically).

This fix adds a utility method that returns an integer version while the
old utility method producing a string is retained for those that use
Puppet as an API. The function now uses the integer version of this
utility method.

A test is added that asserts that an Integer is produced.

At some point it would be beneficial to reimplement the function using
the 4.x function API as it will simply raise a generic error if given
input of the wrong type. It will also magically change a floating point
input value to integer.

### DIFF
--- a/spec/unit/parser/functions/fqdn_rand_spec.rb
+++ b/spec/unit/parser/functions/fqdn_rand_spec.rb
@@ -6,11 +6,11 @@ describe "the fqdn_rand function" do
   include PuppetSpec::Scope
 
   it "returns an integer" do
-    expect(fqdn_rand(3)).to satisfy {|n| n.is_a?(Integer) }
+    expect(fqdn_rand(3)).to be_an(Integer)
   end
 
   it "provides a random number strictly less than the given max" do
-    expect(fqdn_rand(3)).to satisfy {|n| n.to_i < 3 }
+    expect(fqdn_rand(3)).to satisfy {|n| n < 3 }
   end
 
   it "provides the same 'random' value on subsequent calls for the same host" do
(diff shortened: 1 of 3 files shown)


hlindberg:
App Veyor failing with known issues regarding rdoc

spec/unit/parser/functions/fqdn_rand_spec.rb
```diff
@@ -4,7 +4,11 @@
...
 describe "the fqdn_rand function" do
   include PuppetSpec::Scope
-  
+
+  it "returns an integer" do
+    expect(fqdn_rand(3)).to satisfy {|n| n.is_a?(Integer) }
+  end
+
   it "provides a random number strictly less than the given max" do
     expect(fqdn_rand(3)).to satisfy {|n| n.to_i < 3 }
```
kylog: can the `to_i` go away now?

kylog:
Other than that one nit in the spec, I'm :+1:

kylog:
And appveyor failures are the usual rdoc ones. Known bogo-failure.

spec/unit/parser/functions/fqdn_rand_spec.rb
```diff
@@ -4,7 +4,11 @@
 
 describe "the fqdn_rand function" do
   include PuppetSpec::Scope
-  
+
+  it "returns an integer" do
+    expect(fqdn_rand(3)).to satisfy {|n| n.is_a?(Integer) }
```
thallgren: Simpler as `expect(fqdn_rand(3)).to be_a(Integer)`

hlindberg:
ok - two nits - will update

PR #4021, opened 2015-06-09
(PUP-3774) Use gpasswd when making group membership changes in a user resource

branan:
once using usermod. This had a couple of problems when a user came
from a non-local PAM source

1) The user could potentially not exist in /etc/passwd. This would
simply cause `usermod` to fail.

2) The user could be in a group that's not considered valid by the
`usermod` command (especially a group that contains a space in its
name).

existing groups when adding a user to a new group, and will be able to
modify local group mappings even if the user comes from a different
PAM source.

valid operation with gpasswd (no local group would have such a name)

### DIFF
--- a/lib/puppet/property/list.rb
+++ b/lib/puppet/property/list.rb
@@ -1,4 +1,5 @@
 require 'puppet/property'
+require 'set'
 
 module Puppet
   class Property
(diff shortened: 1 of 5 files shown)


joshcooper:
Typo in the commit f8324ad subject, should be PUP-3774

branan:
@joshcooper bah. Fixed

McdonaldSeanp:
Needs to be retargeted at either 4.10.x, 5.3.x, or master

PR #4023, opened 2015-06-10
(PUP-4656) acceptance: service should start from file update using notify

er0ck:
This change adds an acceptance test against the service resource,
ensuring it will start and re-start after being notified from a File in
the same manifest.


er0ck:
this is a duplicate of tests/ticket_7165_no_refresh_after_starting_service.rb.  ticket will be re-purposed for higher-priority backlog of tests

PR #4024, opened 2015-06-10
(PUP-4386) Windows group invalid user error msgs

Iristyle:

```puppet
group { 'guests':
  ensure  => 'present',
  gid     => 'S-1-5-32-546',
}
```

```puppet
group { 'guests':
  ensure => present,
}
```

```
```

```
Notice: Compiled catalog for vagrant-2008r2.localdomain in environment production in 0.58 seconds
Error: Could not resolve name:
Error: /Stage[main]/Main/Group[Guests]/members: change from VAGRANT-2008R2\Guest,VAGRANT-2008R2\Administrator to VAGRANT-2008R2\Guest,VAGRANT-2008R2\Administrator, failed: Could not resolve name:
```

### DIFF
--- a/lib/puppet/provider/group/windows_adsi.rb
+++ b/lib/puppet/provider/group/windows_adsi.rb
@@ -38,6 +38,11 @@ Puppet::Type.type(:group).provide :windows_adsi do
     return '' if users.nil? or !users.kind_of?(Array)
     users = users.map do |user_name|
       sid = Puppet::Util::Windows::SID.name_to_sid_object(user_name)
+      if !sid
+        resource.debug("#{user_name} (unresolvable to SID)")
+        next user_name
+      end
+
       if sid.account =~ /\\/
         account, _ = Puppet::Util::Windows::ADSI::User.parse_name(sid.account)
       else
@@ -47,6 +52,10 @@ Puppet::Type.type(:group).provide :windows_adsi do
       "#{sid.domain}\\#{account}"
     end
     return users.join(',')
+  end
+
+  def member_valid?(user_name)
+    ! Puppet::Util::Windows::SID.name_to_sid_object(user_name).nil?
   end
 
   def group
--- a/lib/puppet/type/group.rb
+++ b/lib/puppet/type/group.rb
@@ -99,12 +99,20 @@ module Puppet
       def is_to_s(currentvalue)
         if provider.respond_to?(:members_to_s)
           currentvalue = '' if currentvalue.nil?
-          return provider.members_to_s(currentvalue.split(','))
+          currentvalue = currentvalue.is_a?(Array) ? currentvalue : currentvalue.split(',')
+
+          return provider.members_to_s(currentvalue)
         end
 
         super(currentvalue)
       end
       alias :should_to_s :is_to_s
+
+      validate do |value|
+        if provider.respond_to?(:member_valid?)
+          return provider.member_valid?(value)
+        end
+      end
     end
 
     newparam(:auth_membership, :boolean => true, :parent => Puppet::Parameter::Boolean) do
--- a/spec/unit/provider/group/windows_adsi_spec.rb
+++ b/spec/unit/provider/group/windows_adsi_spec.rb
@@ -35,11 +35,13 @@ describe Puppet::Type.type(:group).provider(:windows_adsi), :if => Puppet.featur
     let(:user1) { stub(:account => 'user1', :domain => '.', :to_s => 'user1sid') }
     let(:user2) { stub(:account => 'user2', :domain => '.', :to_s => 'user2sid') }
     let(:user3) { stub(:account => 'user3', :domain => '.', :to_s => 'user3sid') }
+    let(:invalid_user) { SecureRandom.uuid }
 
     before :each do
       Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with('user1').returns(user1)
       Puppet::Util::Windows::SID.stubs(:name_to_sid_object).with('user2').returns(user2)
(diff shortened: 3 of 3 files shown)


Iristyle:
This is an alternative implementation of #4010

Iristyle:
So this approach has the benefit of not throwing / catching a special exception... but it introduces new validation logic in the form of `member_valid?`.  It's `members_to_s` now returns the input `user_name` when it can't be resolved (which is more or less what the user provider does in a roundabout way.

Iristyle:
@joshcooper would appreciate your feedback on the separate proposed approaches here for PUP-4386.  I think they both have pros / cons.

lib/puppet/type/group.rb
```diff
@@ -99,12 +99,20 @@ def insync?(current)
       def is_to_s(currentvalue)
         if provider.respond_to?(:members_to_s)
           currentvalue = '' if currentvalue.nil?
-          return provider.members_to_s(currentvalue.split(','))
+          currentvalue = currentvalue.is_a?(Array) ? currentvalue : currentvalue.split(',')
```
Iristyle: This line is required because `is_to_s` gets called in the the resource harness at https://github.com/puppetlabs/puppet/blob/stable/lib/puppet/transaction/resource_harness.rb#L147

lib/puppet/type/group.rb
```diff
@@ -99,12 +99,20 @@ def insync?(current)
...
+          return provider.members_to_s(currentvalue)
         end
 
         super(currentvalue)
       end
       alias :should_to_s :is_to_s
+
+      validate do |value|
+        if provider.respond_to?(:member_valid?)
+          return provider.member_valid?(value)
```
ferventcoder: How does this validate against users being created within the same manifest? I am looking for autorequires here.

ferventcoder: Answer, it doesn't:

```
Notice: Compiled catalog for windows.vm in environment production in 3.40 seconds
Error: Could not resolve name: tim
Error: /Stage[main]/Main/Group[TestUsers]/members: change from  to BUILTIN\Administrators,tim failed: Could not resolve name: tim
Notice: /Stage[main]/Main/User[tim]/ensure: created
```

ferventcoder: Note that the user is created at the end of that script.

ferventcoder: That's a separate issue.

Iristyle: Yeah, thanks for reminding me about that.

I was going to add a note that this approach might have a problem with a manifest that creates users at the same time.  We need to test this appropriately to make sure it's not a problem.

ferventcoder: I filed https://tickets.puppetlabs.com/browse/PUP-4737 to follow up on that one.

ferventcoder: I think the group provider already has this issue, because it will fail later.

ferventcoder: Yep, against stable this is what you get:

```sh
Error: Could not resolve name: tim
Error: /Stage[main]/Main/Group[TestUsers]: Could not evaluate: Puppet::Util::Log requires a message
Notice: /Stage[main]/Main/User[tim]/ensure: created
```

ferventcoder: In short, we are no worse off based on this.

ferventcoder: I will check that explicit requires works though.

ferventcoder: Works against this PR as well.

spec/unit/provider/group/windows_adsi_spec.rb
```diff
@@ -35,11 +35,13 @@
     let(:user1) { stub(:account => 'user1', :domain => '.', :to_s => 'user1sid') }
     let(:user2) { stub(:account => 'user2', :domain => '.', :to_s => 'user2sid') }
     let(:user3) { stub(:account => 'user3', :domain => '.', :to_s => 'user3sid') }
+    let(:invalid_user) { SecureRandom.uuid }
```
ferventcoder: Doing this in a real life scenario has been a bit trying.

ferventcoder: ACLs will continue to contain deleted accounts, but I can't seem to get it to show me a user that doesn't exist in a group on Windows. In short, I'm not sure how to get a real-life scenario where puppet resource group would spit out a user that is invalid.

ferventcoder:
@Iristyle How did you get the following in a real life scenario of running `puppet resource user group`?

```puppet
group { 'guests':
  ensure => present,
}
```

lib/puppet/provider/group/windows_adsi.rb
```diff
@@ -38,6 +38,11 @@ def members_to_s(users)
     return '' if users.nil? or !users.kind_of?(Array)
     users = users.map do |user_name|
       sid = Puppet::Util::Windows::SID.name_to_sid_object(user_name)
+      if !sid
+        resource.debug("#{user_name} (unresolvable to SID)")
+        next user_name
```
ferventcoder: Why the explicit next?

ferventcoder:
I'm +1 to moving forward on this.

lib/puppet/provider/group/windows_adsi.rb
Iristyle: It returns the value early inside the `map`

ferventcoder: returns a value, right on. :)

ferventcoder:
Added backport of this at https://github.com/puppetlabs/puppet/pull/4399

PR #4027, opened 2015-06-12
(PUP-4751) Optimize & Secure safe_posix_fork

fihuer:
Before this, on POSIX systems, each time we execute a command, it forks a new process and this one will close inherited file descriptors by looping between 3 and 256 although there is more than 256 possible file descriptors. It's not safe and it's really slow.

Now, we try to list open file descriptors by reading "/proc/self/fd" and close the ones we don't want. This will fallback to the original method if it can't find "/proc/self/fd".

### DIFF
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -292,7 +292,15 @@ module Util
       $stdout.reopen(stdout)
       $stderr.reopen(stderr)
 
-      3.upto(256){|fd| IO::new(fd).close rescue nil}
+      begin
+        Dir.foreach('/proc/self/fd') do |f|
+          if f != '.' && f != '..' && f.to_i >= 3
+            IO::new(f.to_i).close rescue nil
+          end
+        end
+      rescue Errno::ENOENT # /proc/self/fd not found
+        3.upto(256){|fd| IO::new(fd).close rescue nil}
+      end
 
       block.call if block
     end
--- a/spec/unit/util_spec.rb
+++ b/spec/unit/util_spec.rb
@@ -229,11 +229,30 @@ describe Puppet::Util do
       (0..256).each {|n| IO.stubs(:new) }
     end
 
-    it "should close all open file descriptors except stdin/stdout/stderr" do
+    it "should close all open file descriptors except stdin/stdout/stderr when /proc/self/fd exists" do
+      # This is ugly, but I can't really think of a better way to do it without
+      # letting it actually close fds, which seems risky
+      fds = [".", "..","0","1","2","3","5","100","1000"]
+      fds.each do |fd|
+        if fd == '.' || fd == '..'
+          next
+        elsif ['0', '1', '2'].include? fd
+          IO.expects(:new).with(fd.to_i).never
+        else
+          IO.expects(:new).with(fd.to_i).returns mock('io', :close)
+        end
+      end
+
+      Dir.stubs(:foreach).with('/proc/self/fd').multiple_yields(*fds)
+      Puppet::Util.safe_posix_fork
+    end
+
+    it "should close all open file descriptors except stdin/stdout/stderr when /proc/self/fd doesn't exists" do
       # This is ugly, but I can't really think of a better way to do it without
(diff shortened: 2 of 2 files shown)


lib/puppet/util.rb
```diff
@@ -292,7 +292,15 @@ def safe_posix_fork(stdin=$stdin, stdout=$stdout, stderr=$stderr, &block)
...
 
-      3.upto(256){|fd| IO::new(fd).close rescue nil}
+      begin
+        Dir.foreach('/proc/self/fd') do |f|
+          if f != '.' && f != '..' && f.to_i >= 3
+            IO::new(f.to_i).close rescue nil
+          end
+        end
+      rescue Errno::ENOENT # /proc/self/fd not found
+        3.upto(256){|fd| IO::new(fd).close rescue nil}
```
joshcooper: @fihuer could you add a test for this behavior?

spec/unit/util_spec.rb
```diff
@@ -233,7 +233,15 @@ def get_mode(file)
       # This is ugly, but I can't really think of a better way to do it without
       # letting it actually close fds, which seems risky
       (0..2).each {|n| IO.expects(:new).with(n).never}
-      (3..256).each {|n| IO.expects(:new).with(n).returns mock('io', :close) }
+
+      fds = Dir.entries("/proc/self/fd")
```
joshcooper: This will fail on systems that don't have `/proc`, definitely windows, not sure about solaris. You may need to confine the example group to `:if => Puppet.features.posix?`

MikaelSmith: You could also mock Dir access to '/proc/self/fd' to return a fixed set of files.

peterhuene: The spec should fail on OSX, at least.

peterhuene: Also not sure if `self` is implemented on all systems that have a procfs implementation.  I think examples mocking its existence and non-existence would be best.

fihuer:
I've added a test when /proc/self/fd doesn't exists

spec/unit/util_spec.rb
```diff
@@ -229,11 +229,29 @@ def get_mode(file)
       (0..256).each {|n| IO.stubs(:new) }
     end
 
-    it "should close all open file descriptors except stdin/stdout/stderr" do
+    it "should close all open file descriptors except stdin/stdout/stderr when /proc/self/fd exists", :if => Dir.exists?('/proc/self/fd') do
```
peterhuene: I think we'd rather have `Dir.foreach` stubed for `/proc/self/fd` to return a known list of descriptors, including '.', '..', 0, 1, and 2.  For those special ones, I think we should stub `IO.expects(:new).with(foo).never` so that if it gets call, it will fail the example.

fihuer:
@peterhuene Specs now stubs `Dir.foreach` and it was already checking for FileDescriptors 0, 1, 2

spec/unit/util_spec.rb
```diff
@@ -229,11 +229,29 @@ def get_mode(file)
...
     end
 
-    it "should close all open file descriptors except stdin/stdout/stderr" do
+    it "should close all open file descriptors except stdin/stdout/stderr when /proc/self/fd exists" do
+      # This is ugly, but I can't really think of a better way to do it without
+      # letting it actually close fds, which seems risky
+      [".","..",0,1,2].each {|n| IO.expects(:new).with(n).never}
+      fds = [".", "..","0","1","2","3","5","100","1000"]
+      Dir.stubs(:foreach).with('/proc/self/fd').multiple_yields(*fds)
+      fds.delete(".")
```
peterhuene: I think we can remove the `delete` calls in favor of the following in the each lambda:

Note: not tested code follows:

```ruby
fds.each do |fd|
  if fd in ['.', '..', '0', '1', '2']
    IO.expects(:new).with(n).never if 
  else
    IO.expects(:new).with(n).returns mock('io', :close)
  end
end
```

PR #4028, opened 2015-06-12
DO NOT MERGE: (PUP-4747) Change PopsBridge::Expression.to_s to return the source

hlindberg:
Before this, the PopsBridge::Expression would return a string from to_s
that was produed by the ModelTreeDumper. That format is S-expression
like and intended for AST tree debugging.

This commit changes the return to instead be the source code associated
with the expression.

There are naturally issues with this approach as 4.x accepts potentially
complex expressions as a default value expression. The receiving end of
the result must be prepared to deal with multiple lines being returned,
and text that possibly includes comments, conditional logic, etc.

The solution should be as ok as the 3.x implementation that would have
similar issues (although not supporting complex expressions).

### DIFF
--- a/lib/puppet/parser/ast/pops_bridge.rb
+++ b/lib/puppet/parser/ast/pops_bridge.rb
@@ -20,7 +20,8 @@ class Puppet::Parser::AST::PopsBridge
     end
 
     def to_s
-      Puppet::Pops::Model::ModelTreeDumper.new.dump(@value)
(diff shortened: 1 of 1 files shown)


kylog:
@hlindberg, doing a scrub of older PRs. It looks like this one wasn't merged but #4030 was (for the same ticket number). Should this stay open?

henrik-lindberg:
Closing this - we opted for the more explicit approach (asking for the source as opposed to relying on to_s to return the right thing).

PR #4029, opened 2015-06-12
(PUP-4749) testdir incorrectly used in calling all functions test

er0ck:
The previous version of this test incorrectly handled the testdir
variable, as it could have been set from multiple agents, when they
exist (puppet server).  This is a simple change to only have a single
agents loop, rather than having to store an array or hash of testdirs.

### DIFF
--- a/acceptance/tests/parser_functions/calling_all_functions.rb
+++ b/acceptance/tests/parser_functions/calling_all_functions.rb
@@ -1,10 +1,21 @@
 test_name 'Calling all functions.. test in progress!'
 
+# create single manifest calling all functions
+step 'Apply manifest containing all function calls'
+def manifest_call_each_function_from_array(functions)
+  manifest = ''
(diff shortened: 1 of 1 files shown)


camlow325:
:see_no_evil: This looks good to me and looks like it's passing Travis.  Not sure about the appveyor failure but doesn't seem like this should have caused any Ruby spec tests to fail since this is just an acceptance test change.  @kurtwall, @er0ck, or @hlindberg - any reason we shouldn't just merge this?

camlow325:
Although, should this PR be targeted at stable and then merged back up to master after it goes in?

kurtwall:
LGTM. Merging because the AppVeyor build failure is unrelated.

camlow325:
@kurtwall - will this fix get merged back to the stable branch at some point?

kurtwall:
@camlow325 I presume with the next merge-up to stable.

PR #4030, opened 2015-06-12
(PUP-4747) Format resource_type arguments as source code

pcarlisle:
In the Puppet 3 parser, default parameter values on a serialized
resource_type object would appear to be puppet source code. This makes the
Puppet 4 parser behave roughly the same way, by retrieving the original source
code representing an argument.


pcarlisle:
I'm not exactly sure how to write the tests for this, the existing resource/type_spec.rb has no examples of creating a type from actual source which would be necessary here

hlindberg:
Will create a ticket for adding tests - merging

PR #4032, opened 2015-06-13
(PUP-3263) Adds `architecture` parameter

petems:
Can specify arch wanted when installing a package.

This is extremely rough, needs some help getting it across the line :smile:

### DIFF
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -9,7 +9,7 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
   These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
   or an array where each element is either a string or a hash."
 
-  has_feature :install_options, :versionable, :virtual_packages
+  has_feature :install_options, :versionable, :virtual_packages, :architecture
 
   commands :yum => "yum", :rpm => "rpm"
 
@@ -149,6 +149,10 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
         self.debug "Downgrading package #{@resource[:name]} from version #{is[:ensure]} to #{should}"
         operation = :downgrade
       end
+    end
+
+    if @resource[:architecture]
+      wanted = "#{wanted}.#{@resource[:architecture]}"
     end
 
     # Yum on el-4 and el-5 returns exit status 0 when trying to install a package it doesn't recognize;
--- a/lib/puppet/type/package.rb
+++ b/lib/puppet/type/package.rb
@@ -384,6 +384,10 @@ module Puppet
       desc "A read-only parameter set by the package."
     end
 
+    newproperty(:architecture) do
+      desc "The architecture version of a package."
+    end
+
     newparam(:adminfile) do
       desc "A file containing package defaults for installing packages.
 
--- a/spec/unit/provider/package/yum_spec.rb
+++ b/spec/unit/provider/package/yum_spec.rb
@@ -250,6 +250,31 @@ describe provider_class do
     end
   end
 
+  describe 'when specifying package architecture' do
+    let(:name) { "mypackage" }
+    let(:resource) do
+      Puppet::Type.type(:package).new(
+        :name         => name,
+        :architecture => 'x86_64',
+        :provider     => 'yum'
+      )
+    end
+
+    it 'should be able to set version and detect architecture when installed' do
+      resource[:ensure] = :installed
+      provider.expects(:yum).with('-d', '0', '-e', '0', '-y', :install, "mypackage.x86_64")
+      provider.install
+    end
+
+    it 'should be able to set version and detect architecture when version given' do
+      version = '1.2'
+      resource[:ensure] = version
+      provider.stubs(:query).returns :ensure => version
+      provider.expects(:yum).with('-d', '0', '-e', '0', '-y', :install, "mypackage-1.2.x86_64")
+      provider.install
+    end
+  end
+
   describe 'when uninstalling' do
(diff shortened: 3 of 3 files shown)


lib/puppet/type/package.rb
```diff
@@ -66,6 +66,7 @@ module Puppet
       :methods => [:package_settings_insync?, :package_settings, :package_settings=]
     feature :virtual_packages, "The provider accepts virtual package names for install and uninstall."
 
+
```
joshcooper: extra newline

lib/puppet/provider/package/yum.rb
```diff
@@ -9,7 +9,7 @@
   These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
   or an array where each element is either a string or a hash."
 
-  has_feature :install_options, :versionable, :virtual_packages
+  has_feature :install_options, :versionable, :virtual_packages, :package_arch
```
joshcooper: Often provider features are "manages_something", perhaps "manages_architecture"? /cc @nfagerlund

lib/puppet/type/package.rb
```diff
@@ -384,6 +385,10 @@ def change_to_s(currentvalue, newvalue)
       desc "A read-only parameter set by the package."
     end
 
+    newparam(:package_arch) do
+      desc "Choose the architecture version of a package"
+    end
+
```
joshcooper: since this is the package type, can we shorten this to `arch` or `architecture`?

lib/puppet/type/package.rb
```diff
@@ -384,6 +385,10 @@ def change_to_s(currentvalue, newvalue)
       desc "A read-only parameter set by the package."
     end
 
+    newparam(:package_arch) do
```
joshcooper: Since this is a parameter, puppet won't compare the current arch against the desired arch. So if you had `noarch` installed, but wanted `x86_64` puppet would still think it's in-sync. Would that ever happen in practice? If so, we'd need to make this a property, provided we can always retrieve the current `arch` and compare against the desired value.

joshcooper:
@petems Thanks for submitting this, it would be useful for a number of package providers. rpm and yum already parse the currently installed arch, so it's part of the way there. Also it'd be nice to be able to install platform specific gems. I've rekicked the failing gems as I think we had some CI fixed after you submitted this.

lib/puppet/provider/package/yum.rb
```diff
@@ -151,6 +151,10 @@ def install
       end
     end
 
+    if @resource[:package_arch]
+      wanted = "#{wanted}.#{@resource[:package_arch]}"
+    end
+
```
joshcooper: This code path is during install, should we be able to uninstall an arch specific package too?

lib/puppet/provider/package/yum.rb
nfagerlund: None of the package ones currently follow that standard — see http://docs.puppetlabs.com/references/latest/type.html#package-provider-features for the list. Since the standard doesn't seem to be universal, I don't particularly care, and would default to using the local style for that resource type. (In this case, using the name of the parameter as the name of the feature.)

lib/puppet/type/package.rb
petems: I think architecture, to avoid confusion with the Arch OS. Will fix :+1:

lib/puppet/type/package.rb
petems: Hmm, it's a possibility, I'll change to a property. I'll fix that.

lib/puppet/provider/package/yum.rb
petems: Changed.

lib/puppet/type/package.rb
petems: Fixed.

lib/puppet/provider/package/yum.rb
petems: Hmm, possibly. The Yum provider uses the RPM uninstall code, might need some help plumbing that.

hkenney:
@joshcooper This might be ready for merge now that tests are green? Could you take a final look at it?

lib/puppet/provider/package/yum.rb
MikaelSmith: Did your update attempt to address uninstall? We also wonder about how doing install with a different package architecture would work if the package is already installed. The in_sync check will confirm the package is already installed, requiring you to explicitly uninstall it in one run, then install the other version in another.

joshcooper:
@petems the current implementation makes it possible to specify the architecture during installation, but what about uninstall, e.g. `ensure => absent`? Also, what happens if you have a `noarch` package currently installed, and you specify `architecture => 'x86_64`'. I would expect puppet to install the more specific architecture package, but I don't think the PR supports this.

petems:
@joshcooper Good point, I feel like the arch parameter might need to get bubbled up to the RPM type, and have a method for collecting the current arch value. This would also help with the usecase of uninstall that @MikaelSmith mentioned...

I might try to break this down into a smaller PR just to add the arch parameter as a new property in the RPM T/P, and get that merged first, as the other bits are going to be a bit more complicated.

HAIL9000:
@petems Do you want to continue working off this pull request? Or did you want to close it and open a smaller pull request as you mentioned.

kylog:
@petems have you had a chance to look into this one any more?

branan:
@petems If we don't hear back from you in another week or so we're gonna go ahead and close this for inactivity. Feel free to re-open if you have time to look at it again, or bother us through the normal engineering process to address the ticket.

petems:
@branan Hi, sorry was on holiday, but to be honest I think I'll need engineering help with this. I'll close for now, pending some help.

PR #4036, opened 2015-06-16
(PUP-4458) Refactor validation of parameter signatures

thallgren:
This commit unifies how parameters are validated so that instantiations
of nodes, classes, and definitions use a validatio logic that has more
in common with the validation of lambda invocations and function calls.

### DIFF
--- a/lib/puppet/parser/resource.rb
+++ b/lib/puppet/parser/resource.rb
@@ -73,18 +73,15 @@ class Puppet::Parser::Resource < Puppet::Resource
   # Retrieve the associated definition and evaluate it.
   def evaluate
     return if evaluated?
-
     Puppet::Util::Profiler.profile("Evaluated resource #{self}", [:compiler, :evaluate_resource, self]) do
       @evaluated = true
-      if klass = resource_type and ! builtin_type?
-        finish
-        evaluated_code = klass.evaluate_code(self)
-
-        return evaluated_code
-      elsif builtin?
+      if builtin?
         devfail "Cannot evaluate a builtin type (#{type})"
-      else
+      elsif resource_type.nil?
         self.fail "Cannot find definition #{type}"
+      else
+        finish
+        resource_type.evaluate_code(self)
       end
     end
   end
@@ -115,9 +112,9 @@ class Puppet::Parser::Resource < Puppet::Resource
     @finished
   end
 
-  def initialize(*args)
-    raise ArgumentError, "Resources require a hash as last argument" unless args.last.is_a? Hash
-    raise ArgumentError, "Resources require a scope" unless args.last[:scope]
+  def initialize(type, title, attributes)
+    raise ArgumentError, 'Resources require a hash as last argument' unless attributes.is_a? Hash
+    raise ArgumentError, 'Resources require a scope' unless attributes[:scope]
     super
 
     @source ||= scope.source
@@ -271,7 +268,8 @@ class Puppet::Parser::Resource < Puppet::Resource
   end
 
   def add_scope_tags
-    if scope_resource = scope.resource
+    scope_resource = scope.resource
+    unless scope_resource.nil? || scope_resource.equal?(self)
       merge_tags(scope_resource)
     end
   end
--- a/lib/puppet/parser/resource.rb
+++ b/lib/puppet/parser/resource.rb
@@ -80,7 +80,7 @@ class Puppet::Parser::Resource < Puppet::Resource
       elsif resource_type.nil?
         self.fail "Cannot find definition #{type}"
       else
-        finish
+        finish(false) # Call finish but do not validate
         resource_type.evaluate_code(self)
       end
     end
@@ -98,13 +98,18 @@ class Puppet::Parser::Resource < Puppet::Resource
   end
 
   # Do any finishing work on this object, called before evaluation or
-  # before storage/translation.
-  def finish
+  # before storage/translation. The method does nothing the second time
+  # it is called on the same resource.
+  #
+  # @param do_validate [Boolean] true if validation should be performed
+  #
+  # @api private
+  def finish(do_validate = true)
     return if finished?
     @finished = true
     add_defaults
     add_scope_tags
-    validate
+    validate if do_validate
   end
 
   # Has this resource already been finished?
@@ -312,11 +317,15 @@ class Puppet::Parser::Resource < Puppet::Resource
 
   # Make sure the resource's parameters are all valid for the type.
   def validate
-    @parameters.each do |name, param|
-      validate_parameter(name)
+    if builtin_type?
+      begin
+        @parameters.each { |name, value| validate_parameter(name) }
+      rescue => detail
+        self.fail Puppet::ParseError, detail.to_s + " on #{self}", detail
+      end
+    else
+      resource_type.validate_resource(self)
     end
-  rescue => detail
-    self.fail Puppet::ParseError, detail.to_s + " on #{self}", detail
   end
 
   def extract_parameters(params)
--- a/lib/puppet/pops/types/type_asserter.rb
+++ b/lib/puppet/pops/types/type_asserter.rb
@@ -1,6 +1,7 @@
 # Utility module for type assertion
 #
-module Puppet::Pops::Types::TypeAsserter
+module Puppet::Pops::Types
+module TypeAsserter
   # Asserts that a type_to_check is assignable to required_type and raises
   # a {Puppet::ParseError} if that's not the case
   #
@@ -9,6 +10,7 @@ module Puppet::Pops::Types::TypeAsserter
   # @param type_to_check [PAnyType] Type to check against the required type
   # @return The type_to_check argument
   #
+  # @api public
   def self.assert_assignable(subject, expected_type, type_to_check)
     check_assignability(Puppet::Pops::Types::TypeCalculator.singleton, subject, expected_type, type_to_check)
     type_to_check
@@ -23,6 +25,7 @@ module Puppet::Pops::Types::TypeAsserter
   # @param nil_ok [Boolean] Can be true to allow nil value. Optional and defaults to false
   # @return The value argument
   #
+  # @api public
   def self.assert_instance_of(subject, expected_type, value, nil_ok = false)
     unless value.nil? && nil_ok
       tc = Puppet::Pops::Types::TypeCalculator.singleton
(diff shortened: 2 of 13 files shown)


hlindberg:
As noted earlier - this modifies puppet/resource.rb which is agent side specific and it makes the agent require pops && rgen even if not doing an apply.

hlindberg:
needs a rebase now that PUP-4702 is merged. Also remove the DO NOT MERGE: on one of the commits.

hlindberg:
Needs a rebase and preferably done so there is no commit that changes the agent part of the source except any final changes. (This to enable bisecting) - maybe squash the two 4458 commits ?

thallgren:
Rebased and squashed.

lib/puppet/pops/types/type_asserter.rb
```diff
@@ -1,6 +1,7 @@
 # Utility module for type assertion
 #
-module Puppet::Pops::Types::TypeAsserter
+module Puppet::Pops::Types
+module TypeAsserter
```
hlindberg: why this odd construct and not just a straight definition of the module? Is it because the Puppet::Pops::Types module does not exist yet and you want it created as an empty module to then redefine?

hlindberg:
Good to mark the methods / modules with @api public and private to indicate where we reserve the right to change in minor releases (i.e. among public ruby methods)

lib/puppet/resource.rb
```diff
@@ -11,9 +11,12 @@ class Puppet::Resource
   include Puppet::Util::Tagging
 
   include Enumerable
-  attr_accessor :file, :line, :catalog, :exported, :virtual, :validate_parameters, :strict
+  attr_accessor :file, :line, :catalog, :exported, :virtual, :strict
   attr_reader :type, :title
 
+  # @deprecated not used
```
hlindberg: You cannot know that - it may be unused in the puppet code base but it may be used elsewhere.

lib/puppet/resource.rb
```diff
@@ -116,7 +119,6 @@ def to_yaml_properties
   # Set a given parameter.  Converts all passed names
   # to lower-case symbols.
   def []=(param, value)
-    validate_parameter(param) if validate_parameters
```
hlindberg: Cannot remove this as it changes API.

lib/puppet/resource.rb
```diff
@@ -230,10 +232,7 @@ def initialize(type, title = nil, attributes = {})
       end
 
       # Set things like strictness first.
-      attributes.each do |attr, value|
-        next if attr == :parameters
-        send(attr.to_s + "=", value)
-      end
+      attributes.each { |attr, value| send(:"#{attr}=", value) unless attr == :parameters }
```
hlindberg: irrelevant change - remove this change

lib/puppet/resource.rb
```diff
@@ -256,7 +255,7 @@ def initialize(type, title = nil, attributes = {})
       tag_if_valid(self.title)
     end
 
-    if strict? and ! resource_type
+    if resource_type.nil? && strict?
```
hlindberg: remove this change

lib/puppet/resource.rb
```diff
@@ -385,104 +384,6 @@ def name
...
-  # Consult external data bindings for class parameter values which must be
-  # namespaced in the backend.
-  #
-  # Example:
-  #
-  #   class foo($port=0){ ... }
-  #
-  # We make a request to the backend for the key 'foo::port' not 'foo'
-  #
-  def lookup_external_default_for(param, scope)
```
hlindberg: This changes API

hlindberg:
There are too many changes in agent code. All of Puppet::Resource::* should be considered agent code and all changes have to be API compliant. Since any change here is super sensitive remove all stylistic and minor optimization changes. Remove all small adjustments that are irrelevant even if the old code is "stupid" or "seems to do it wrong". We are not fixing those "bugs" now.

lib/puppet/resource/type.rb
```diff
@@ -229,17 +239,6 @@ def name_is_regex?
     @name.is_a?(Regexp)
   end
 
-  def assign_parameter_values(parameters, resource)
```
hlindberg: changes API

lib/puppet/resource/type.rb
```diff
@@ -374,4 +411,93 @@ def warn_if_metaparam(param, default)
...
+      members[key_type] = arg_type
+    end
+    Puppet::Pops::Types::TypeFactory.struct(members)
+  end
+  private :create_params_struct
+
+  # Consult external data bindings for class parameter values which must be
+  # namespaced in the backend.
+  #
+  # Example:
```
hlindberg: use yard @example tag

hlindberg:
Puppet::Type should not be changed - none of the changes seems required

hlindberg:
The separation of concerns are truly horrible between agent and server side but there is nothing we can do about that unless we hold off doing this until the next major release. Need to discuss if we should do something about this, and hold off on these changes.

lib/puppet/pops/types/type_asserter.rb
thallgren: It might seem odd at first but this construct allows everything in the TypeAsserter module to reference everything in the Puppet::Pops::Types module without qualification. It has nothing to do with module initialization.

lib/puppet/resource.rb
thallgren: Really? It changes 'and' to '&&' and removes a redundant space after '!'.

lib/puppet/resource.rb
thallgren: No it does not. The lookup_external_default_for is a :private method (look beneath the method)

thallgren:
When restoring the 'validate_complete' in puppet/lib/resource.rb, I see that it uses the Puppet::Pops::Types::TypeCalculator (It's been that way since way back in 3.x). I guess validation and setting_resource_parameters never happens in agent code? I.e. the resource is never evaluated there?

thallgren:
Pushed a new squashed commit where the changes to puppet/resource.rb has been kept to a minimum. One method is added to puppet/type.rb to avoid ugliness and also to prepare for a future enhancement where both name and value is checked (currently it just dispatches to a method that checks the name).

All spec tests for Puppet::Resource and Puppet::Type remains untouched now to further check that there's no API breakage.

lib/puppet/pops/types/type_asserter.rb
hlindberg: And it does not when written as one FQN ? (somewhat surprised)

lib/puppet/parser/resource.rb
```diff
@@ -74,15 +74,12 @@ def add_edge_to_stage
...
-      return evaluated_code
-    elsif builtin?
+    if builtin?
       devfail "Cannot evaluate a builtin type (#{type})"
-    else
+    elsif resource_type.nil?
       self.fail "Cannot find definition #{type}"
+    else
+      finish
     end
```
hlindberg: Change does not seem to be needed. Please remove.

lib/puppet/parser/resource.rb
```diff
@@ -112,9 +109,9 @@ def finished?
     @finished
   end
 
-  def initialize(*args)
-    raise ArgumentError, "Resources require a hash as last argument" unless args.last.is_a? Hash
-    raise ArgumentError, "Resources require a scope" unless args.last[:scope]
+  def initialize(type, title, attributes)
+    raise ArgumentError, 'Resources require a hash as last argument' unless attributes.is_a? Hash
+    raise ArgumentError, 'Resources require a scope' unless attributes[:scope]
```
hlindberg: Change does not seem to be needed. Please remove.

lib/puppet/parser/resource.rb
```diff
@@ -213,7 +210,8 @@ def add_defaults
   end
 
   def add_scope_tags
-    if scope_resource = scope.resource
+    scope_resource = scope.resource
+    unless scope_resource.nil? || scope_resource.equal?(self)
```
hlindberg: Change not needed. please remove

hlindberg:
There are different kinds of validation, the validate_complete happens master side (and in apply) when compiling, other validation take place agents side.

hlindberg:
Sorry, made comments on the wrong file - deleted them.

lib/puppet/parser/resource.rb
```diff
@@ -74,15 +74,12 @@ def add_edge_to_stage
   def evaluate
     return if evaluated?
     @evaluated = true
-    if klass = resource_type and ! builtin_type?
-      finish
-      evaluated_code = klass.evaluate_code(self)
```
hlindberg: I can't figure out what happened with evaluation after this change and how the evaluate method can set @evaluated to true when it does not actually evaluate

lib/puppet/resource.rb
```diff
@@ -11,9 +11,12 @@ class Puppet::Resource
   include Puppet::Util::Tagging
 
   include Enumerable
-  attr_accessor :file, :line, :catalog, :exported, :virtual, :validate_parameters, :strict
+  attr_accessor :file, :line, :catalog, :exported, :virtual, :strict
   attr_reader :type, :title
 
+  # @deprecated
```
hlindberg: @kylog @joshcooper - the construct for attr_accessor for :validate_parameters looks odd - can one of you help shed some light on what it is used for, or if it is dead code. I an imagine that it is used to conveniently get empty methods that implementations of a resource type can supply extra validation for, but I am unsure.

lib/puppet/resource.rb
hlindberg: doh, beneath the *next* method - *sigh*

lib/puppet/resource.rb
```diff
@@ -393,73 +396,8 @@ def missing_arguments
...
-      v
-    end
-    found ? value : nil
-  end
-  private :lookup_in_module
-
+  # @deprecated Not used by Puppet
+  # @api private
   def set_default_parameters(scope)
     return [] unless resource_type and resource_type.respond_to?(:arguments)
```
hlindberg: Suggest adding a deprecation warning call for this - anyone calling this externally should be warned at that it will go away at a major version boundary.

lib/puppet/pops/types/type_asserter.rb
thallgren: It's Ruby. Never ceases to surprise you :-) This piece of code can be used to verify:
```
module A
  class B
  end
end

module A
  module C
    class D
      def initialize
        B.new
      end
    end
  end
end

module A::C
  class E
    def initialize
      B.new
    end
  end
end

A::C::D.new # succeeds
A::C::E.new # fails with "uninitialized constant A::C::E::B (NameError)"
```

lib/puppet/parser/resource.rb
thallgren: The `evalutate_code` is called from `validate` which is called from `finish`. The flow is correct but I admit that it looks rather odd. I'll change this.

lib/puppet/resource.rb
thallgren: Kylo's previous comment on that attribute got lost during rebase. He wrote:

"FWIW I just grepped a recent drop (May 27th) of the forge and found no uses of `validate_parameters`.

I haven't grokked this entire PR, but if we're merely deprecating this attribute, but not removing it, that seems okay."

thallgren:
Divided this PR into two commits where the first one contains only code clean-ups.

thallgren:
A mistake was made when dividing the commits in my previous push. This is now fixed.

thallgren:
@hlindberg I pushed two new commits. The first one is a rebase of the 'Refactor validation of...' where the `Puppet::Type.validate_parameter_value` method was removed. I chose to make that a rebase since it didn't make sense to review code that was first added and then removed. The second commit addresses your concern regarding always converting to `Puppet::Parser::Resource::Param` objects.

PR #4037, opened 2015-06-16
(PUP-2315) - Type validation error messages

thallgren:
This commit introduces a number of improvements to error messages
issued when validating call signatures. Improvements include:

- Unification
  Instantiations of nodes, classes, resources, and user defined types
  share the same validation logic as calls to lambdas, functions, and
  methods.
- Issue message by priority
  Report only one issue and not all issues that might or might not be
  a consequence of the reported issue. Examples include:
  - Argument count mismatch
  - Required but missing block
  - Unexpected block
- Suppress repeated messages
- Shorten type descriptions when possible
  When type names differ, leave out the type details. Does not apply
  when String is validated using Pattern or Enum
- Shorten Variant output to text like "one of T1, T2, or T3"
- Appoint details using paths into complex types instead of reporting
  large blobs of very similar texts

This commit builds on work in PUP-4458 (not yet merged)


thallgren:
Rebased on top of updated PR for PUP-4458.

thallgren:
I have been running acceptance tests on several platforms from commit 407a616f1bca66a07e811a0621447bbfa2e77eaa which is PUP-5009 merged onto this PR. The tests consistently produces one error. That error is howerver also produced when running from master so I'm fairly certain that it's not due to anything in in this PR.
```
       Total Suite Time: 4534.23 seconds
      Average Test Time: 15.97 seconds
              Attempted: 284
                 Passed: 230
                 Failed: 0
                Errored: 1
                Skipped: 52
                Pending: 1
                  Total: 284
```

The test in error is that `puppet_manages_own_configuration_in_robust_manner` exits with 1 and produces no output. It seems related to something in PUP-5009 (same test and incorrect exit code is mentioned there) that still isn't fixed:

```
Test Case tests/config/puppet_manages_own_configuration_in_robust_manner.rb reported #<Beaker::Host::CommandFailure: Host 'u7frsu736pic03q.delivery.puppetlabs.net' exited with 1 while running:
 puppet master
Last 10 lines of output were:
>
```

PR #4038, opened 2015-06-17
(PUP-4760) Service type should not retrieve ensure status when not needed

fihuer:
Before this, service type will always retrieve the ensure status even if no
ensure parameters is given.

But, because no Service Type parameter is dependant on whether the service
is running or not, we can just skip the ensure status retrieval.

Now, a service type will only retrieve the ensure status when a ensure
parameter is given.

### DIFF
--- a/lib/puppet/type/service.rb
+++ b/lib/puppet/type/service.rb
@@ -230,5 +230,9 @@ module Puppet
         debug "Skipping restart; service is not running"
       end
     end
+
+    def self.needs_ensure_retrieved
+      false
+    end
   end
 end
--- a/lib/puppet/type.rb
+++ b/lib/puppet/type.rb
@@ -1030,6 +1030,12 @@ class Type
     insync
   end
 
+  # Says if the ensure property should be retrieved if the resource is ensurable
+  # Defaults to true. Some resource type classes can override it
+  def self.needs_ensure_retrieved
+    true
+  end
+
   # Retrieves the current value of all contained properties.
   # Parameters and meta-parameters are not included in the result.
   # @todo As oposed to all non contained properties? How is this different than any of the other
@@ -1044,7 +1050,7 @@ class Type
     # Provide the name, so we know we'll always refer to a real thing
     result[:name] = self[:name] unless self[:name] == title
 
-    if ensure_prop = property(:ensure) or (self.class.validattr?(:ensure) and ensure_prop = newattr(:ensure))
+    if ensure_prop = property(:ensure) or (self.class.needs_ensure_retrieved and self.class.validattr?(:ensure) and ensure_prop = newattr(:ensure))
       result[:ensure] = ensure_state = ensure_prop.retrieve
     else
       ensure_state = nil
--- a/spec/unit/type_spec.rb
+++ b/spec/unit/type_spec.rb
@@ -638,6 +638,20 @@ describe Puppet::Type, :unless => Puppet.features.microsoft_windows? do
       @resource.property(:ensure).stubs(:retrieve).returns :absent
     end
 
+    it "should always retrieve the ensure value by default" do
+      @ensurable_resource = Puppet::Type.type(:file).new(:name => "/not/existent", :mode => "0644")
+      Puppet::Type::File::Ensure.stubs(:ensure).returns :absent
+      Puppet::Type::File::Ensure.any_instance.expects(:retrieve).once
+      @ensurable_resource.retrieve_resource
+    end
+
+    it "should not retrieve the ensure value if specified" do
+      @ensurable_resource = Puppet::Type.type(:service).new(:name => "DummyService", :enable => true)
+      @ensurable_resource.properties.each { |prop| prop.stubs(:retrieve) }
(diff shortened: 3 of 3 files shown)


joshcooper:
I wouldn't expect the ensure property's retrieve method to be called if a desired value hasn't been specified. Will need to look into this more.

lib/puppet/type/service.rb
```diff
@@ -230,5 +230,31 @@ def refresh
...
+        next if property.name == :ensure
+        if ensure_state == :absent
+          result[property] = :absent
+        else
+          result[property] = property.retrieve
+        end
+      end
+
+      result
+    end
```
joshcooper: Note to reviewers, the is mostly copy and pasted from https://github.com/puppetlabs/puppet/blob/3.x/lib/puppet/type.rb#L1039-L1063. The difference is the superclass `Puppet::Type#retrieve` does:

```
if ensure_prop = property(:ensure) or (self.class.validattr?(:ensure) and ensure_prop = newattr(:ensure))
```

But this PR changes the service type only to do:

```
 if ensure_prop = property(:ensure)
```

The PR modifies puppet to behave like I thought it did, only manage the state of resources that are expressed in the manifest. For some reason, the `ensure` property is handled special, and the behavior was added in 3f6c9481.

When is it necessary to evaluate the second half? Isn't it always true for `ensurable` types? I'm wondering if we can make this change in `Puppet::Type` instead of special casing the service type.

fihuer:
For some resources (File for example), properties can be dependent on the ensure status. A file without `ensure` but with a `mode` will never create a file because it will retrieve the status (thanks to the second half of the if) and it'll be `:absent`, thus setting any other parameter to `:absent`. Without this, it will try to chmod a non-existent file.
But this doesn't apply on all types : you can still `enable` a service without `ensure` (think HA).

ffrank:
During review, we noticed that this has been [discussed](https://groups.google.com/forum/#!msg/puppet-dev/P8ReCHvmd2o/FnKZ0bPxk5sJ) quite a while ago.

At the end of that thread, I linked a proof-of-concept I did then. Perhaps you can salvage some ideas from there.

ffrank:
Also it would appear that @puppetcla is drunk.

er0ck:
<scratches head>  i don't *recall* contributing this change :-)

joshcooper:
Given @zaphod42's comment in https://groups.google.com/d/msg/puppet-dev/P8ReCHvmd2o/M3AqXGGlyQAJ:

> Ah, so there is other code out there that relies on the current behavior of retrieve. I had suspect as much, but it is good to know :)

we can't accept this PR in its current form. If there's a way to preserve compatibility, while still getting the speed improvement, then that would be great. But I'm not seeing how at the moment.

fihuer:
This thread suggested to change the root "type" class behavior, this PR doesn't. It only changes the bahavior of the service type. And it doesn't break anything, it only saves a couple of unused status retrieval (which can be costy).
If this is too much for you, we can still add a parameter like "lazy_status_retrieval", or whatever else, which sets `self.needs_ensure_retrieved` to true.

joshcooper:
One of the messages on the thread https://groups.google.com/d/msg/puppet-dev/P8ReCHvmd2o/tuhzAM86wbYJ, says:

>  if ensure is not present in the resources returned from Type.retrieve, then modules using collected resources will break.

So this PR would break anyone collecting service resources.

We could add opt-in behavior, though if we did that, we'd probably want that to be useful for all types, not just service.

So next steps, let's take this conversation to the puppet-dev thread, as I think many folks have thought about this before and can help weigh options.

fihuer:
I think this assumption is wrong. Yep, it was breaking tests and collected resources, but I don't think it's because `ensure` is not present. Instead, it was breaking because the patch was ignoring `ensure` without a `ensure.should` attribute. Thus, breaking collected resource because collected resource have a `ensure` but no `ensure.should` (They are collected, not managed). 
With the previous patch, `puppet resource service` was broken (it's using collected resource, right ?), but not with this patch.

We can still take this conversation to puppet-dev, so you can convince yourself.

joshcooper:
@fihuer I respect everyone's opinion on that thread, and think it would be unwise to proceed without getting their feedback.

kylog:
We want to spend some time testing this, so I pulled it into our upcoming sprint.

fihuer:
@kylog Any news ?

joshcooper:
@fihuer sorry, we haven't forgotten, but we haven't had time to review.

MikaelSmith:
This approach makes sense as a general solution to types that have no `absent` state, where it doesn't exist. `service` appears to be the only core type that has that property.

MikaelSmith:
As I noted in https://tickets.puppetlabs.com/browse/PUP-4760?focusedCommentId=256449&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-256449, this should target master.

kylog:
Okey doke, per the comments above, we're going to close this against 3.x and retarget it against master.

MikaelSmith:
Retargeted as https://github.com/puppetlabs/puppet/pull/4803.

PR #4039, opened 2015-06-17
(PUP-4761) Add --show-changes option to summarize changes

fihuer:
With this we can add the --changes option to puppet runs.
It will complement the --summarize output by adding changes's title and status (success, failure, ...).

This helps log parsing and human readability.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -246,6 +246,7 @@ class Puppet::Configurer
   private :run_internal
 
   def send_report(report)
+    puts report.display_changes if Puppet[:show_changes]
     puts report.summary if Puppet[:summarize]
     save_last_run_summary(report)
     Puppet::Transaction::Report.indirection.save(report, nil, :environment => Puppet::Node::Environment.remote(@environment)) if Puppet[:report]
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -240,6 +240,46 @@ class Puppet::Transaction::Report
     host
   end
 
+  # Provide a summary of the changes
+  def display_changes
+    changes = sumup()
+    return nil if changes.empty?
+    ret = ""
+    # display the different changes for each type
+    changes.keys.sort.each { |type|
+      ret += "%s:\n" % type
+      changes[type].keys.sort.each { |event|
+        # Count the number of occurences for each value in the changes[type][event]
+        count = changes[type][event].inject(Hash.new(0)) {|h,i| h[i] += 1; h }
+        ret += count.to_a.collect { |status, nb|
+          if nb > 1
+            "   %15s %s (%s)" % [status + ":", event, nb]
+          else
+            "   %15s %s" % [status + ":", event]
+          end
+        }.join("\n")
+        ret += "\n"
+      }
+    }
+    return ret
+  end
+
+   # Copy events in a structure that can be return with the report
+   def sumup()
+     changes= {}
+     resource_statuses.each do |name, status|
+       type = status.resource_type
+       next if ["Schedule","Filebucket"].include? type
+       title = status.title
+       changes[type] ||= {}
+       changes[type][title] ||= []
+       status.events.each { |event|
+         changes[type][title].push(event.status)
+       }
+     end
+     changes
+   end
+
   # Provide a human readable textual summary of this report.
   # @note This is intended for debugging purposes
   # @return [String] A string with a textual summary of this report.
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1600,6 +1600,11 @@ EOT
         :default  => false,
         :type     => :boolean,
         :desc     => "Whether to print a transaction summary.",
+    },
+    :show_changes => {
+        :default  => false,
+        :type     => :boolean,
+        :desc     => "Whether to print the transaction changes.",
     }
   )
 


joshcooper:
To discuss in PR triage tomorrow.

lib/puppet/transaction/report.rb
```diff
@@ -70,6 +70,10 @@ class Puppet::Transaction::Report
   #
   attr_reader :metrics
 
+  # A hash of number of changes per resource
+  # @return [Hash<{String => Hash<{String => Event Status}>}] A map of type+title to number of events
```
peterhuene: This comment doesn't seem to be correct, based on `sumup`.  This is actually a `Hash<String => Hash<String => Array<Status>>`, correct?

lib/puppet/transaction/report.rb
```diff
@@ -70,6 +70,10 @@ class Puppet::Transaction::Report
...
+  attr_accessor :changes
```
peterhuene: Do we actually need an attribute here?  Seems like there could be a single method called `display_changes` that "sumups" and then displays them.

lib/puppet/transaction/report.rb
```diff
@@ -240,6 +244,46 @@ def name
...
+  end
+
+   # Copy events in a structure that can be return with the report
+   def sumup()
+       changes= {}
+       resource_statuses.each do |name, status|
+           status.events.each do |e|
+               e.resource =~ /([^\[]*)\[([^\]]*)\]/
+               type = $1
+               title = $2
```
joshcooper: I'm fairly certain all events for a status, are associated with the same resource, and you can get the type and title of that resource from the `status` object itself, avoiding this regex.

lib/puppet/configurer.rb
```diff
@@ -246,6 +246,12 @@ def run_internal(options)
   private :run_internal
 
   def send_report(report)
+    if Puppet[:show_changes]
```
peterhuene: I'm not sure we need an accessor and two different actions, `sumup` and `displaychanges` (should be named `display_changes` too).  I think this would be better as `puts report.display_changes if Puppet[:display_changes]` and let `display_changes` on the report handle summing up and displaying the changes if non-empty.

lib/puppet/transaction/report.rb
```diff
@@ -240,6 +240,46 @@ def name
...
+    }
+    return ret
+  end
+
+   # Copy events in a structure that can be return with the report
+   def sumup()
+     changes= {}
+     resource_statuses.each do |name, status|
+       type = status.resource_type
+       next if ["Schedule","Filebucket"].include? type
```
joshcooper: why are we skipping these?

lib/puppet/transaction/report.rb
```diff
@@ -240,6 +240,46 @@ def name
     host
   end
 
+  # Provide a summary of the changes
+  def display_changes
```
joshcooper: Based on the name `display_changes`, I'd expect this method to display the changes, or rename the method to `get_changes`, and have the caller print them like you're doing.

lib/puppet/configurer.rb
```diff
@@ -246,6 +246,7 @@ def run_internal(options)
   private :run_internal
 
   def send_report(report)
+    puts report.display_changes if Puppet[:show_changes]
```
joshcooper: During PR triage we were talking about whether it makes sense to combine the changes and summary in the `report.summary` method. Puppet already has too many settings, so we are leery about adding new ones.

That said, it looks like the `display_changes` will print one line for each (type,event) tuple, plus the blank between each type, so it could easily add hundreds of lines of output. So I'm okay having it be a separate option.

fihuer:
Maybe I haven't described that well the goal of this. What we wanted to have is a parsable/readable output which shows the resource statuses of the run. We can already do that by parsing the regular log but it's machine-specific or it can differ from the previous run. Hostnames, IPs or just evaluation time can show up. 
With this (and without the `report.summary` which is, kind-of, unique for every run) we can just parse this output across many machines and just compare it pretty easily.

BTW, I have to skip `Schedule` and `Filebucket` because they have no events apparently. And, thus, no statuses. Without skiping them, it adds empty line like :
```
Filebucket:
Schedule:
Notify:
 Sucess: blah
```

joshcooper:
Makes sense about wanting a "stable" output to compare. But you mention parsing the output across multiple machines, and comparing them, which makes me think a report processor would be better suited to automate the parsing of the report, rather than scraping the `puppet agent` output.

fihuer:
You're right if Puppet runs automatically, reports goes to the master, properly parsed and a global "report" could be generated. But when you run one-off runs across many machines (5000+) it's not anymore a good idea. 3 reasons : 
- Sending a report takes some time (~10secs in our case)
- Sending 5000+ reports (~500Ko) at almost the same time is not a good idea for your masters
- Parsing asynchronously YAML dumps of reports (if you don't have a proper report processor) takes ages because YAML is slow.

Moreover, if you run Puppet with parallel SSHs ([ClusterShell](http://cea-hpc.github.io/clustershell/)) , with this kind of output it's pretty easy to find out if something went wrong without a single grep.

hkenney:
@joshcooper Could you take another look at this?

joshcooper:
@fihuer we were talking about this during PR triage, and thinking this could be best handled by writing a puppet application that parses `Puppet[:lastrunreport]` and does whatever you want to the results. Puppet applications can be packaged in modules on the master, and are automatically distributed to all agents via pluginsync. Then you can run `puppet agent --onetime && puppet <myapp> <action>`

This way it doesn't add load to the server, you can decide what resources you care about, and how you want the output to look, and whether to output human readable text or JSON, etc.

fihuer:
@joshcooper Piping into another script or creating a new Puppet application might fit our needs. I will close this one

Thanks for your time.

PR #4040, opened 2015-06-20
Add -H flag to svcs to strip column headers

drewfisher314:
This change adds the -H flag to /usr/bin/svcs to strip out the column headers.

### DIFF
--- a/lib/puppet/provider/service/smf.rb
+++ b/lib/puppet/provider/service/smf.rb
@@ -32,7 +32,7 @@ Puppet::Type.type(:service).provide :smf, :parent => :base do
   end
 
   def self.instances
-   svcs.split("\n").select{|l| l !~ /^legacy_run/ }.collect do |line|
+   svcs("-H").split("\n").select{|l| l !~ /^legacy_run/ }.collect do |line|
      state,stime,fmri = line.split(/\s+/)
      status =  case state
                when /online/; :running
--- a/spec/unit/provider/service/smf_spec.rb
+++ b/spec/unit/provider/service/smf_spec.rb
@@ -28,7 +28,7 @@ describe provider_class, :if => Puppet.features.posix? do
     end
 
     it "should get a list of services (excluding legacy)" do
-      provider_class.expects(:svcs).with().returns File.read(my_fixture('svcs.out'))
+      provider_class.expects(:svcs).with('-H').returns File.read(my_fixture('svcs.out'))
       instances = provider_class.instances.map { |p| {:name => p.get(:name), :ensure => p.get(:ensure)} }
       # we dont manage legacy
       expect(instances.size).to eq(2)


drewfisher314:
I'm still fairly inexperienced to ruby/rspec so I'm not sure why this failed.  This runs just fine on my local Solaris zone.  Any advice would be appreciated.

stschulte:
There is a test that expects that puppet will call `svcs` (without any arguments) to get a list of current services. This expectation becomes false now, because you run `svcs` with arguments.

You have to modify `spec/unit/provider/service/smf_spec.rb`:
```ruby
provider_class.expects(:svcs).with().returns File.read(my_fixture('svcs.out'))
```

This should now become
```ruby
provider_class.expects(:svcs).with('-H').returns File.read(my_fixture('svcs.out'))
```

Now you probably wonder what `File.read(my_fixture('svcs.out'))` means. This will basically return the content of the file `spec/fixtures/unit/provider/service/smf/svcs.out`, so make sure the content of the file matches the result of a real `svcs -H` command.

You should now be able to run the test yourself (you have to have rspec and stuff installed):
```bash
bundle exec rspec --format d -c spec/unit/provider/service/smf_spec.rb
```
if this comes back all green run the whole testsuite
```bash
bundle exec rake spec
```

drewfisher314:
outstanding.  Thank you so much for the pointer.  The suite is running now and I hope to submit it again shortly.

joshcooper:
Hi @drewfisher314, thank you for your contribution. It looks like you have 2 commits and a merge commit that need to be squashed. You'll want to do:

```
$ git checkout ticket/2509-add-H-flag
$ git rebase -i 44e9490
# in the pick list, change "pick" to "squash" in the bottom commit, and save the file
```

That should create a single commit whose parent is 44e9490.

Also while you're there, can you amend the commit to follow our [contributor guidelines](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md). In particular, the subject line should reference the ticket "(PUP-2509) Add -H flag ..." and describe the before and after behavior of the change in the body of the commit. Once those changes are in, you'll want to force push to your topic branch (assuming the remote name of your fork is `origin`):

```
$ git push -f origin ticket/2509-add-H-flag
```

About the change itself, is the `-H` flag supported on all Solaris versions that puppet supports? Currently I think that list includes 10 and 11.

Finally, the appveyor windows & rdoc issue can be ignored. If you rebase your PR onto upstream/master, you'll pick up the puppetlabs fix that resolves it:

```
$ git fetch upstream
$ git rebase upstream/master
$ git push -f origin HEAD
```

where `upstream` should the name of the puppetlabs fork (as opposed to your fork). And you'll want to force push to your topic branch.

drewfisher314:
@joshcooper - thanks for the git help.  I'm clearly not a pro at git :)

Solaris 10 and 11 use the same flag (-H) to omit column headers.

Here's the S10 manpage
http://docs.oracle.com/cd/E23823_01/html/816-5165/svcs-1.html#REFMAN1svcs-1

Here's the S11 manpage
http://docs.oracle.com/cd/E23824_01/html/821-1461/svcs-1.html#REFMAN1svcs-1

Both refer to it as:

-H
    Omits the column headers.

joshcooper:
Awesome, thanks @drewfisher314 One minor nit, could you describe the old behavior in the body of the commit message? I can't tell if puppet failed to parse the output and aborted the run, or if puppet was confused and thought there was a service whose name came from the column header, or something else? Thanks again!

drewfisher314:
Puppet succeeds with no issues, it just also grabs the header row as I explained in the bug.  I've also updated the commit message.

joshcooper:
Thanks @drewfisher314 for your contribution!

PR #4041, opened 2015-06-21
(PUP-4771) static_compiler should get content from request environment

parkerd:
We've been using this fix in production to fix an issue we encountered with directory environments and the static_compiler catalog terminus. I didn't add tests because the store_content method appears to just be mocked and can't say I'm familiar enough with rspec or puppet code to test it properly.

### DIFF
--- a/lib/puppet/indirector/catalog/static_compiler.rb
+++ b/lib/puppet/indirector/catalog/static_compiler.rb
@@ -65,7 +65,7 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
   # from the resource, inserts the metadata into the file resource, and uploads
   # the file contents of the source to the file bucket.
   #
-  # @param host [String] The host name of the node requesting this catalog
+  # @param request [Puppet::Indirector::Request] The request for the catalog
   # @param resource [Puppet::Resource] The resource to replace the metadata in
   # @param file [Puppet::Type::File] The file RAL associated with the resource
   def find_and_replace_metadata(request, resource, file)
@@ -85,7 +85,7 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
   # This performs the actual metadata rewrite for the given file resource and
   # uploads the content of the source file to the filebucket.
   #
-  # @param host [String] The host name of the node requesting this catalog
+  # @param request [Puppet::Indirector::Request] The request for the catalog
   # @param resource [Puppet::Resource] The resource to add the metadata to
   # @param metadata [Puppet::FileServing::Metadata] The metadata of the given fileserver based file
   def replace_metadata(request, resource, metadata)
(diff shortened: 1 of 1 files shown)


hlindberg:
Thanks for this contribution. There is a small thing to fix; the documentation comments needs to be adjusted to reflect the change from being given a host to instead getting a request.

parkerd:
Updated the docs, although I'm not really sure what type the request is. The methods in this code seem to be better documented than other places, but the original request object coming in wasn't documented. I took a guess of "Puppet::Indirector::Request", if that's wrong just let me know and I'll update.

joshcooper:
Looks good, will review in more detail and merge

lib/puppet/indirector/catalog/static_compiler.rb
```diff
@@ -205,7 +206,7 @@ def store_content(resource)
       Puppet.info "Content for '#{resource[:source]}' already exists"
     else
       Puppet.info "Storing content for source '#{resource[:source]}'"
-      content = Puppet::FileServing::Content.indirection.find(resource[:source])
+      content = Puppet::FileServing::Content.indirection.find(resource[:source], {:environment => request.environment})
```
joshcooper: What happens if the file content cannot be found? I assume it returns nil. Are there tests for this?

parkerd: The bug we ran into involved requesting a new file from an environment other than production, results in the puppet run failing because the requested file does not exist in production.

Per my original comment, I didn't add tests because the store_content method appears to just be mocked and can't say I'm familiar enough with rspec or puppet code to test it properly.

joshcooper:
Thanks @parkerd, we had a few more questions from PR Triage. If you know the answer to the first one, that'd be great. I'll take a look at the latter two:

* How do environments work with custom mount points (so when the master is serving files not in any module)?
* If `FileContent.indirection.find` returns `nil`, we probably want to raise at that point
* It would be good to test the happy and failure paths for FileContent.

parkerd:
I don't know the general answer to how environments work with fileserver.conf, but in this instance it appears to ignore the environment. FWIW I recreated the bug in vagrant and tested with both fileserver and module files. Fileserver works as expected. Here's the error when requesting a new module file from a custom environment:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: undefined method `content' for nil:NilClass
```

Agree it would be better to have a nil check on content before continuing, I can add that if desired.

lib/puppet/indirector/catalog/static_compiler.rb
joshcooper: @parkerd Thanks for the offer. I think it's actually ok, because the `store_content` method is only called if we were able to retrieve metadata for the file, so we had to at least stat'ed the file. There is a race condition where the file could be deleted after we retrieve it's metadata and before we attempt to read its content, but it's unlikely.

I'll take a look at adding a test to ensure the correct environment is threaded through to the FileMetadata and FileContent methods, and merge this.

joshcooper: I tried updating the static_compiler_spec, but failed... The tests don't actually work in large part, e.g.

```
    resource_catalog.resources.each do |resource|
          next unless resource.type == "File"
          expect(resource[:owner]).to eq(0)
          expect(resource[:group]).to eq(0)
          expect(resource[:mode]).to  eq(420)
        end
```

The above expectations are never reached because the `resource_catalog` doesn't contain any `File` resources as mentioned in https://github.com/puppetlabs/puppet/pull/4052#issuecomment-115178163.

hkenney:
@joshcooper Could you take another look at this?

joshcooper:
I merged this as is since it's correct, and I didn't want to hold it up waiting for the static compiler tests to be rewritten. Thanks @parkerd for your contribution, this should be released in the next feature release of puppet, probably 4.3.0.

PR #4042, opened 2015-06-22
(PUP-4634) Add some timings/metrics in report

fihuer:
With this, reports have more timings. So reported time matches much better the wall time. On top of this, Reports now shows some HTTP pool metrics like number of
created connections, closed connections, borrowed connections or cached connections.


joshcooper:
To discuss in PR triage tomorrow

lib/puppet/configurer.rb
```diff
@@ -157,12 +173,17 @@ def run_internal(options)
...
       unless options[:catalog]
         begin
-          if node = Puppet::Node.indirection.find(Puppet[:node_name_value],
-              :environment => Puppet::Node::Environment.remote(@environment),
-              :configured_environment => Puppet[:environment],
-              :ignore_cache => true,
-              :transaction_uuid => @transaction_uuid,
-              :fail_on_404 => true)
+          node = nil
+          nodeRetrTime = thinmark do
```
peterhuene: Nit picky comment: the preferred naming convention for local variables is to use `snake_case`.

lib/puppet/configurer.rb
```diff
@@ -65,8 +71,10 @@ def retrieve_catalog(query_options)
     end
 
     return nil unless result
-
-    convert_catalog(result, @duration)
+    converted = nil
+    convertTime = thinmark { converted = convert_catalog(result, @duration) }
```
peterhuene: Nit picky comment: the preferred naming convention for local variables is to use `snake_case`.

lib/puppet/configurer.rb
```diff
@@ -246,9 +267,15 @@ def run_internal(options)
   private :run_internal
 
   def send_report(report)
+    reportSendTime = thinmark {
```
peterhuene: Nit picky comment: the preferred naming convention for local variables is to use `snake_case`.

lib/puppet/transaction.rb
```diff
@@ -79,8 +79,12 @@ def perform_pre_run_checks
   # necessary events.
   def evaluate(&block)
     block ||= method(:eval_resource)
-    generator = AdditionalResourceGenerator.new(@catalog, relationship_graph, @prioritizer)
-    @catalog.vertices.each { |resource| generator.generate_additional_resources(resource) }
+    generator = nil
+    additionResourceGenerationTime = thinmark do
```
peterhuene: Nit picky comment: the preferred naming convention for local variables is to use `snake_case`.

lib/puppet/configurer.rb
```diff
@@ -44,6 +44,12 @@ def init_storage
     end
   end
 
+  def init_http_pool_monitoring(report, http_pool)
+    if http_pool.is_a? Puppet::Network::HTTP::Pool
+      http_pool.report = report
```
joshcooper: I don't like storing the report in the pool this way. Once the run finishes, the pool still has a reference to the now obsolete report.

joshcooper:
@fihuer thank you for your contribution. As @hlindberg was saying in https://tickets.puppetlabs.com/browse/PUP-4634?focusedCommentId=176023&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-176023, since this adds metrics to the report, which are persisted in puppetdb, we should discuss requirements for this feature on puppet-dev.

hlindberg:
@joshcooper This was IIRC taken to the dev-list, and discussed there for a bit (asking for input etc.). It looks like the result of that is what ended up in this PR (but I only gave it a quick glance).

lib/puppet/configurer.rb
```diff
@@ -92,7 +104,9 @@ def get_facts(options)
       # get a hash with both of these pieces of information.
       #
       # facts_for_uploading may set Puppet[:node_name_value] as a side effect
-      facts_hash = facts_for_uploading
+      facts_hast = nil
```
joshcooper: should be `facts_hash`, but it's already initialized above, so initializing to `nil` isn't necessary.

lib/puppet/configurer.rb
```diff
@@ -246,9 +267,15 @@ def run_internal(options)
...
+    reportSendTime = thinmark {
+      Puppet::Transaction::Report.indirection.save(report, nil, :environment => Puppet::Node::Environment.remote(@environment)) if Puppet[:report]
+    }
+    # Includes the indirection save time in the last_run_summary.yaml
+    # and in the summary
+    report.add_times(:reporting, reportSendTime)
+    report.finalize_report
     puts report.summary if Puppet[:summarize]
     save_last_run_summary(report)
-    Puppet::Transaction::Report.indirection.save(report, nil, :environment => Puppet::Node::Environment.remote(@environment)) if Puppet[:report]
```
joshcooper: This means the report is sent before it's finalized, which I think will omit important information from what we send.

lib/puppet/configurer.rb
joshcooper: It would probably be better to push the report onto the context stack, and then access the report using `Puppet.lookup(:report)` from within the http pool.

fihuer: I haven't done it because I thought it's not as lightweight as just assigning a attribute of the `Puppet::Network::HTTP::Pool` object. 
For the obsolete reference, we could just reassign `@report` to `nil` when needed

hkenney:
@joshcooper Would you mind taking another look at this?

fihuer:
@joshcooper Any news on this PR ?

HAIL9000:
@kylog Could you take a look at this one please?

lib/puppet/configurer.rb
```diff
@@ -246,9 +266,16 @@ def run_internal(options)
...
+    report_send_time = thinmark {
+      Puppet::Transaction::Report.indirection.save(report, nil, :environment => Puppet::Node::Environment.remote(@environment)) if Puppet[:report]
+    }
+    # Includes the indirection save time in the last_run_summary.yaml
+    # and in the summary
+    report.add_times(:reporting, report_send_time)
+    report.finalize_report
     puts report.summary if Puppet[:summarize]
     save_last_run_summary(report)
-    Puppet::Transaction::Report.indirection.save(report, nil, :environment => Puppet::Node::Environment.remote(@environment)) if Puppet[:report]
```
joshcooper: I'm a little concerned about these changes, because this area of the code is pretty fragile, and breakage can have big consequences, submitted reports are missing info needed for console, compliance, etc.

This change means the report summary data (`report.raw_summary`) that we submitted is different than the last run report saved on the agent. I'd prefer that the report be immutable once it's submitted.

Is it safe to call `finalize_report` multiple times? This change means `finalize_report` will be called twice before we send the report, and a third time prior to saving the last run summary.

Previously `save_last_run_summary` & `report.summary` came before `Report.indirection.save`, and now the order is reversed. Does the call to `Report.indirection.save` rely on side effects caused by the former?

Given these questions I have to ask is there another way to collect the `report_send_time` that doesn't involve modifying the report? I understand why it's desirable to have, I'm just concerned about what it might break.

fihuer: @joshcooper The only way I see is to log it instead of putting it in the report. But as @hlindberg said in the [ticket](https://tickets.puppetlabs.com/browse/PUP-4634), it adds bloat and overhead to the logs. 

Logging it is fine for me.

fihuer:
@kylog Any news on this one ?

peterhuene:
@joshcooper to follow-up.

joshcooper:
@fihuer I still have concerns described in https://github.com/puppetlabs/puppet/pull/4042#discussion_r36777819, and think it would be fine to just log instead. It's not going to add a large number of log events, right? Also it looks like your branch has conflicts and needs a rebase. Thanks!

whopper:
@fihuer thanks for the updates. We're going to do another pass at reviewing this to make sure everything is looking good!

lib/puppet/configurer.rb
```diff
@@ -236,6 +262,7 @@ def run_internal(options)
       return nil
     ensure
       execute_postrun_command or return nil
+      terminate_http_pool_monitoring(Puppet.lookup(:http_pool))
```
joshcooper: If the postrun command fails, e.g. path is incorrect, then `execute_postrun_command` will return false and we will never call `terminate_http_pool_monitoring`

lib/puppet/configurer.rb
```diff
@@ -53,7 +65,7 @@ def initialize(factory = Puppet::Configurer::DownloaderFactory.new)
   end
 
   # Get the remote catalog, yo.  Returns nil if no catalog can be found.
-  def retrieve_catalog(query_options)
+  def retrieve_catalog(query_options, options)
```
MikaelSmith: `retrieve_catalog` is potentially public, we should default this to an empty hash.

lib/puppet/configurer.rb
MikaelSmith: Can we make these methods private?

ahpook:
Sorry this PR got lost to time - thanks for working on this all but I don't see us carrying this forward.

PR #4043, opened 2015-06-22
(PUP-4776) Make 'puppet agent arg' error instead of silently running the catalog

peculater:
Currently, 'puppet agent' only analyses options beginning with '--', such as '--test' and
'--disable'.  Any other words after agent are ignored.  That is different from the way that
commands like 'puppet resource' and 'puppet apply' work, mainly because those require
an argument, while 'puppet agent' does not.  This has led to, for instance, one of our new
sysadmins typing 'puppet agent disable' instead of 'puppet agent --disable' and being
terribly surprised when about 5 minutes later the server died.

This patch raises an error if someone on the CLI tries to invoke 'puppet agent' with what
has to be an invalid argument.

### DIFF
--- a/lib/puppet/application/agent.rb
+++ b/lib/puppet/application/agent.rb
@@ -372,6 +372,8 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
   end
 
   def setup
+    raise ArgumentError, "The puppet agent command does not take parameters" unless command_line.args.empty?
+
     setup_test if options[:test]
 
     setup_logs
--- a/spec/unit/application/agent_spec.rb
+++ b/spec/unit/application/agent_spec.rb
@@ -197,6 +197,11 @@ describe Puppet::Application::Agent do
       Puppet.stubs(:settraps)
     end
 
+    it "should not run with extra arguments" do
+      @puppetd.command_line.stubs(:args).returns(%w{disable})
+      expect{@puppetd.setup}.to raise_error ArgumentError, /does not take parameters/
+    end
+
     describe "with --test" do
       it "should call setup_test" do
         @puppetd.options[:test] = true


joshcooper:
@peculater thank you for your contribution! The rdoc appveyor failures are unrelated and can be ignored. We'll take a look at this before next week's PR triage on June 30th.

lib/puppet/application/agent.rb
```diff
@@ -359,6 +359,12 @@ def main(daemon)
     daemon.start
   end
 
+  def parse_args(args)
```
peterhuene: I'm not a fan of having a `parse_args` method that doesn't really parse the arguments, but instead ensures they are empty.

I think this really could simply be a one liner at the start of `setup`:

```ruby
raise ArgumentError, "The puppet agent command does not take parameters" unless command_line.args.empty?
```

peterhuene:
@kylog is this considered a breaking change?  Before `puppet agent foo` (note: not `--foo` as options are handled differently) would merrily go on its way, but now we'll error on startup.  My only concern is people that run puppet agent from cron or something that have typos in the command line.

kylog:
IMHO this is not a breaking change in the semver sense, i.e. there isn't an established contract that we are changing here. This is just a good old-fashioned bug fix :)

peterhuene:
Hi @peculater.  I just wanted to check-in to see if you saw my feedback above.  Generally this looks good and we should be able to merge this with the recommended changes applied.  Thanks!

peterhuene:
Hi @peculater.  Thanks for pushing up the changes.  One last thing and I'll get this merged this afternoon: could you either amend the second commit message to be prefixed with `(PUP-4776)` like the first commit or squash the second commit into the first one?  I have a preference for squashing, but it either way works for us.  Thanks again for your contribution!

peculater:
OK, I think that squashed correctly.  :-)

peterhuene:
:+1: will merge after CI passes.

peterhuene:
@peculater Thanks again for the contribution!  This has been merged into the upcoming Puppet 4.3.0 release.

PR #4044, opened 2015-06-23
(PUP-4712) fix environment scenario acceptance test nuances

er0ck:
This change fixes two environment scenario tests so they test against
the correct test environments.  Previously it erroneously used the
standard $codedir environmentpath.
This change also removes the duplicate explicit environment.conf test
which was incorrect and duplicated by another test.


MikaelSmith:
You should update your master branch, it doesn't appear to have appveyor.yml.

MikaelSmith:
Which was the environment.conf test duplicating?

er0ck:
I have an appveyor.yml

er0ck:
the environment_scenario-with_explicit_environment_conf.rb test was duplicating, badly, this test: directory_environment_with_environment_conf.rb

MikaelSmith:
Looks like it was your test in the first place; decided it was ill-advised?

er0ck:
it does appear that way, but we really just split out a larger test of joshua partlow's. 
looks like neither one of us noticed the dupe...

MikaelSmith:
:+1:

PR #4045, opened 2015-06-23
(PUP-4777) Allow hiera 3 as a puppet dependency

joshcooper:
Previously, installing puppet from source or a gem would refuse to use
hiera 3, even though that is the combination that will be included in
puppet-agent 1.2.0.

This commit allows puppet to use hiera 3 as a dependency. Bundler uses
the Gemfile to resolve dependencies when running puppet from source. Our
gem packaging tasks use ext/project_data.yaml when building puppet as a
gem. This way if you gem install puppet, its runtime dependencies will
also be installed. The .gemspec is special, because it only used when
other code, e.g. module, has a dependency on puppet and is running
against a source checkout of puppet (as opposed to a puppet gem).

### DIFF
--- a/.gemspec
+++ b/.gemspec
@@ -43,13 +43,13 @@ Gem::Specification.new do |s|
 
     if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
       s.add_runtime_dependency(%q<facter>, [">= 1.7", "< 3"])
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Did you do this before the rdoc fix was merged? Otherwise, :+1:

PR #4046, opened 2015-06-23
(PUP-4537) Skip rdoc tests on windows

joshcooper:
Appveyor has transient failures running rdoc parser tests on windows
only. Puppet uses rdoc as part of its `puppet doc` command. The
puppet-strings project has mostly, but not yet completely, replaced the
`puppet doc` command. At a future date `puppet doc` will be deprecated
and removed. This commit skips the tests on windows.

### DIFF
--- a/spec/integration/util/rdoc/parser_spec.rb
+++ b/spec/integration/util/rdoc/parser_spec.rb
@@ -2,7 +2,7 @@
 require 'spec_helper'
 require 'puppet/util/rdoc'
 
-describe "RDoc::Parser" do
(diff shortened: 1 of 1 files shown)


ferventcoder:
Thank you. :)

joshcooper:
Ah I forgot to put the merge after unfreeze label on this... I think it's ok, as hopefully we won't need a puppet 4.2.1, but let's hold off on future merges until puppet-agent 1.2.0 goes out.

ferventcoder:
Will do.

PR #4047, opened 2015-06-23
(PDOC-16) Improve strings compatibility

iankronquist:
This PR improves the compatibility of generated modules with the new puppet strings documentation
parser.
The .yardopts file tells yard, a code parsing library used by puppet strings, to parse documentation comments as markdown.
Yard has also been added to the Gemfile.
This PR addresses [PDOC 16](https://tickets.puppetlabs.com/browse/PDOC-16).

### DIFF
--- a/lib/puppet/module_tool/skeleton/templates/generator/manifests/init.pp.erb
+++ b/lib/puppet/module_tool/skeleton/templates/generator/manifests/init.pp.erb
@@ -1,37 +1,44 @@
-# == Class: <%= metadata.name %>
+# Class: <%= metadata.name %>
+# ===========================
 #
 # Full description of class <%= metadata.name %> here.
 #
-# === Parameters
+# Parameters
+# ----------
 #
 # Document parameters here.
 #
-# [*sample_parameter*]
-#   Explanation of what this parameter affects and what it defaults to.
-#   e.g. "Specify one or more upstream ntp servers as an array."
+# * `sample parameter`
+# Explanation of what this parameter affects and what it defaults to.
+# e.g. "Specify one or more upstream ntp servers as an array."
 #
-# === Variables
+# Variables
+# ----------
 #
 # Here you should define a list of variables that this module would require.
 #
-# [*sample_variable*]
-#   Explanation of how this variable affects the function of this class and if
-#   it has a default. e.g. "The parameter enc_ntp_servers must be set by the
-#   External Node Classifier as a comma separated list of hostnames." (Note,
-#   global variables should be avoided in favor of class parameters as
-#   of Puppet 2.6.)
+# * `sample variable`
+#  Explanation of how this variable affects the function of this class and if
+#  it has a default. e.g. "The parameter enc_ntp_servers must be set by the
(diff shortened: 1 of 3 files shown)


hkenney:
Hm, I can't decide if we want the .yardopts file it the .gitignore or not.

If I recall, YARD will also create a .yard (or something similar) file that we definitely want to have in the .gitignore.

hkenney:
One more thing: the skeleton produces an init.pp which has a comment header. This comment header should conform to our suggested style for strings (meaning it should Markdown for formatting and YARD tags).

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/module_tool/skeleton/templates/generator/manifests/init.pp.erb

MikaelSmith:
Also, the last commit should follow the contributor style guide, or be squashed to another commit.

iankronquist:
Oops, sorry, I'll go fix that.

hkenney:
Let's ago ahead and squash these commits into one since the changes are so small

lib/puppet/module_tool/skeleton/templates/generator/manifests/init.pp.erb
```diff
@@ -1,37 +1,43 @@
...
+#  External Node Classifier as a comma separated list of hostnames." (Note,
+#  global variables should be avoided in favor of class parameters as
+#  of Puppet 2.6.)
 #
-# === Examples
+# Examples
+# --------
 #
 #  class { '<%= metadata.name %>':
 #    servers => [ 'pool.ntp.org', 'ntp.local.company.com' ],
```
hlindberg: The above would not be recognized as code by yard I think. Needs an indent of 4. May also want to use the @example annotation from yard.

iankronquist: Good catch, thank you! I did indeed need to use 4 spaces of indentation and `@example`. I fixed this amended the squashed commit.

lib/puppet/module_tool/skeleton/templates/generator/Gemfile
```diff
@@ -5,3 +5,4 @@ gem 'puppet', puppetversion
 gem 'puppetlabs_spec_helper', '>= 0.8.2'
 gem 'puppet-lint', '>= 1.0.0'
 gem 'facter', '>= 1.7.0'
+gem 'yard', '>= 0.8.0'
```
hkenney: So after some thinking I think that we should remove this. This is only going to help developers if they are running from within their module directory (which is not likely). I think we may just have to stick with them installing YARD themselves to use with their install of puppet.

iankronquist: fixed.

hkenney:
Okay last thing! Let's go ahead and close this PR and make a new one retargeted at stable. This is a small change so we don't need to wait for 4.3.

Also don't forget to make sure the PR against Gareth's tool is up to date (commits are squashed, change to the Gemfile is removed)

PR #4048, opened 2015-06-23
(PUP-4780) convert versioncmp() function to 4x API

jhoblitt:
The conversion is needed to work around improper string -> float conversion
that appears to be happening when calling 3x functions.

### DIFF
--- a/lib/puppet/functions/versioncmp.rb
+++ b/lib/puppet/functions/versioncmp.rb
@@ -1,0 +1,36 @@
+require 'puppet/util/package'
+
+# Compares two version numbers.
+#
+# Prototype:
+#
+#     \$result = versioncmp(a, b)
+#
+# Where a and b are arbitrary version strings.
+#
+# This function returns:
+#
+# * `1` if version a is greater than version b
+# * `0` if the versions are equal
+# * `-1` if version a is less than version b
+#
+# @example
+#
+#     if versioncmp('2.6-1', '2.4.5') > 0 {
+#         notice('2.6-1 is > than 2.4.5')
+#     }
+#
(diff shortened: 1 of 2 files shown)


joshcooper:
@hkenney @hlindberg thoughts on this? Could this affect compatiblity, e.g. if a 3.x function tried to call `versioncmp`, would it be still be accessible?

jhoblitt:
@joshcooper - @hlindberg commented on the jira issue this morning that this is a non-go until the next major version bump because the compatibility issue you mentioned.

hlindberg:
Since the 3.x version of the function is removed stdlib would then not be able to support 3.x (without future parser). The removal also affects any calls to Scope.function_versioncmp, since that is no longer functioning.

Instead, just adding the new version of the function and keeping the old makes it backwards compatible with 3x versions without future parser. It would cause calls to the versioncmp function invoked via Scope.function_versioncmp (or equivalent) to pick the 3.x version (i.e. from ERB templates, or other 3.x functions) unless they use the new `call_function` method, in which case they would also get the 4.x version.

joshcooper:
@jhoblitt @hlindberg so what are next steps for this PR? Can we simply leave the existing 3x `versioncmp` function and add the 4x one? Or is something more drastic required?

jhoblitt:
@joshcooper @hlindberg I've un-removed the the 3x function from the PR and rebased on the current master.

hlindberg:
Looks fine. The change will not affect 3.x users using current parser. All others get the new implementation and will get an error if not passing two strings to the function. I think this is good to merge.

hkenney:
I'll test this out and merge :)

PR #4049, opened 2015-06-23
(PUP-4758) Update Rspec `raise_error` calls for Rspec 3.3

whopper:
In Rspec 3.3, any calls to `raise_error` without arguments
are met with loud warnings during test runs. This commit updates
every such call to include the expected exception type, and
expected exception output in most cases.


joshcooper:
Since this only affects tests, I'm thinking we should target stable, but hold off to merge until after puppet 4.2.0 is released.

whopper:
@joshcooper sounds good - retargeting.

PR #4050, opened 2015-06-23
(PUP-4758) Update Rspec `raise_error` calls for Rspec 3.3

whopper:
In Rspec 3.3, any calls to `raise_error` without arguments
are met with loud warnings during test runs. This commit updates
every such call to include the expected exception type, and
expected exception output in most cases.

### DIFF
--- a/spec/integration/provider/package_spec.rb
+++ b/spec/integration/provider/package_spec.rb
@@ -12,7 +12,9 @@ describe "Package provider" do
         options = {:name => "nosuch#{provider.name}", :provider => provider.name}
 
         pkg = Puppet::Type.newpackage(options)
-        expect { pkg.provider.install }.to raise_error
+        expect { pkg.provider.install }.to raise_error { |error|
+          expect(error).not_to eq("")
+        }
       end
 
       it "should be able to get a list of existing packages" do
--- a/spec/unit/graph/sequential_prioritizer_spec.rb
+++ b/spec/unit/graph/sequential_prioritizer_spec.rb
@@ -27,6 +27,6 @@ describe Puppet::Graph::SequentialPrioritizer do
   it "fails to prioritize a key contained in an unknown container" do
     expect do
       priorities.generate_priority_contained_in("unknown", "child 1")
-    end.to raise_error
+    end.to raise_error(NoMethodError, /`down' for nil/)
   end
 end
--- a/spec/unit/network/http/handler_spec.rb
+++ b/spec/unit/network/http/handler_spec.rb
@@ -131,11 +131,6 @@ describe Puppet::Network::HTTP::Handler do
       handler.process(request, response)
     end
 
-    it "should raise an error if the request is formatted in an unknown format" do
-      handler.stubs(:content_type_header).returns "unknown format"
-      expect { handler.request_format(request) }.to raise_error
-    end
-
     it "should still find the correct format if content type contains charset information" do
       request = Puppet::Network::HTTP::Request.new({ 'content-type' => "text/plain; charset=UTF-8" },
                                                    {}, 'GET', '/', nil)
--- a/spec/unit/application/apply_spec.rb
+++ b/spec/unit/application/apply_spec.rb
@@ -359,7 +359,7 @@ describe Puppet::Application::Apply do
         it "should raise an error if we can't find the facts" do
           Puppet::Node::Facts.indirection.expects(:find).returns(nil)
 
-          expect { @apply.main }.to raise_error
+          expect { @apply.main }.to raise_error(RuntimeError, /Could not find facts/)
         end
       end
 
--- a/spec/unit/application/filebucket_spec.rb
+++ b/spec/unit/application/filebucket_spec.rb
@@ -173,7 +173,7 @@ describe Puppet::Application::Filebucket do
     describe "the command backup" do
       it "should fail if no arguments are specified" do
         @filebucket.stubs(:args).returns([])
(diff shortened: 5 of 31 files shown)


spec/integration/provider/package_spec.rb
```diff
@@ -12,7 +12,7 @@
         options = {:name => "nosuch#{provider.name}", :provider => provider.name}
 
         pkg = Puppet::Type.newpackage(options)
-        expect { pkg.provider.install }.to raise_error
+        expect { pkg.provider.install }.to raise_error(/source ?(parameter)?|gem|pip|lock file/)
```
whopper: This test is mildly infuriating :). This `raise_error` captures one of *many* different exceptions, so we can't expect just one. Thus, I've resorted to capturing small fragments of the exception messages in this regex.

MikaelSmith:
Replaces #4049

whopper:
As @joshcooper mentioned in the closed PR:

Since this only affects tests, I'm thinking we should target stable, but hold off to merge until after puppet 4.2.0 is released.

spec/unit/network/http/handler_spec.rb
```diff
@@ -133,7 +133,7 @@ def respond(text)
 
     it "should raise an error if the request is formatted in an unknown format" do
       handler.stubs(:content_type_header).returns "unknown format"
-      expect { handler.request_format(request) }.to raise_error
+      expect { handler.request_format(request) }.to raise_error(NameError, /undefined local variable or method/)
```
joshcooper: "undefined local variable" is surprising, though maybe it's trying to lookup a method named "to_unknown format"?

spec/integration/provider/package_spec.rb
joshcooper: I'm guessing this returns potentially different error messages depending on the default package provider for the system where the test is run? Is the 'lock file' error coming from puppet's lock file when it's applying a catalog or one of the package providers?

whopper: @joshcooper I'll have to check again tomorrow, but it returns different error messages *and* exception types in a few instances. I'll make a list of where each of these message fragments comes from.

spec/unit/network/http/handler_spec.rb
whopper: I'll track down exactly what's going on again with this one tomorrow and give an update.

whopper: @joshcooper the error being thrown is:

```
#<NameError: undefined local variable or method `request' for #<RSpec::ExampleGroups::PuppetNetworkHTTPHandler::WhenProcessingARequest:0x007fb36c40c000>>
```

spec/integration/provider/package_spec.rb
whopper: Exceptions & messages:

OSX:
```
#<Puppet::Error: Mac OS X PKG DMG's must specify a package source.>
#<Puppet::ExecutionFailure: Execution of '/Users/whopper/.rbenv/versions/1.9.3-p547/bin/gem install --no-rdoc --no-ri nosuchgem' returned 2: ERROR:  Could not find a valid gem 'nosuchgem' (>= 0) in any repository>
#<NoMethodError: Could not locate the pip command.>
#<NoMethodError: Could not locate the pip3 command.>
```

MikaelSmith:
:+1: @joshcooper any more feedback?

whopper:
Before merging, let me test out that regex Josh and I were discussing. I am concerned it might not pass on all platforms, since it depends on which provider is being tested... I see now why that test was originally left without specifying expected output :)

spec/unit/network/http/handler_spec.rb
joshcooper: ah `request` is not defined, it needs to be `handler.request_format(a_request)`

joshcooper: Actually something is very wrong with that test. Looks like the code & tests evolved, but this one didn't

spec/unit/graph/sequential_prioritizer_spec.rb
```diff
@@ -27,6 +27,6 @@
   it "fails to prioritize a key contained in an unknown container" do
     expect do
       priorities.generate_priority_contained_in("unknown", "child 1")
-    end.to raise_error
+    end.to raise_error(NoMethodError, /undefined method/)
```
joshcooper: This one is a little odd.. underlying code calls `nil.down` resulting in `undefined method 'down' for nil:NilClass`. Ideally the code would protect against that and raise an ArgumentError when the container doesn't have a priority. But I am reluctant to make that change now. How about we include the `'down' for nil` part of the message, as opposed to some other undefined method?

spec/unit/network/http/handler_spec.rb
whopper: @joshcooper oh duh, lost sight of the forest for the trees. Looking for how this test applies in the modern world..

spec/unit/graph/sequential_prioritizer_spec.rb
whopper: @joshcooper sounds good - done

spec/unit/network/http/handler_spec.rb
whopper: @joshcooper yeah, I'm not sure this is actually applicable any longer. I'm going to remove it unless you happen to see something I don't.

whopper:
Note: Appveyor is failing due to a new FFI release which segfaults on Windows.

PR #4051, opened 2015-06-23
(PUP-4775) Make Puppet::Node serialize env as a string in YAML

hlindberg:
Before this, a Puppet::Node would serialize to YAML in such a way
that all of its instance variables were serialized. This caused
everything in a compelte Environment to be serialized (everything
loaded and evaluated), which in turn can cause OOM, or OOStack errors.

The cause of the problem is that Puppet::Node did not correctly
include Puppet::Util::PsyckSupport which handles the Yaml serialization
API and turns them into calls: to_data_hash, and initialize_from_hash.

The older way of using 'to_yaml' (one such implementation is found in
Environment itself) never gets called when it is not at the root of the
serialization.

This commit adds the missing include and refactors the initialization
from a hash to use the correct API. It also keeps the older way of a
class method from_data_hash, since it is not known if there are any
usage of this in modules implementing indirectors/caches. (It now
delegates to initialize_from_hash to keep the implementation DRY).

### DIFF
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -9,6 +9,9 @@ class Puppet::Node
   # the node sources.
   extend Puppet::Indirector
 
+  # Asymmetric serialization/deserialization required in this class via to/from datahash
+  include Puppet::Util::PsychSupport
+
   # Use the node source as the indirection terminus.
   indirects :node, :terminus_setting => :node_terminus, :doc => "Where to find node information.
     A node is composed of its name, its facts, and its environment."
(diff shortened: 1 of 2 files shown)


thallgren:
I didn't notice that this PR shouldn't be merged so I removed the merge commit from stable (removed as reset on previous commit and then in push --force). Please resubmit this PR.

joshcooper:
@thallgren please do not force push to upstream, as that can prevent others from fast-forwarding. I generally do `git revert -m 1` to create a revert commit, preserving history.

thallgren:
I know @joshcooper. I panicked. And I thought that at that time of day, nobody would have had a chance to pull the commit (so no problem fast-forwarding if it was gone).

Perhaps we should ask GitHub support to add `git config receive.denyNonFastForwards true`?

joshcooper:
I was just wondering if there was a way to do that on a per-repo basis. Seems that that'd be a good thing for the main repo, but I'm not sure how to also allow force pushing to topic branches?

PR #4052, opened 2015-06-24
(PUP-4781) filemode retrieved by static_compiler should be stringified

tdevelioglu:
When a file resource is declared without an explicit mode, the mode
retrieved from the metadata by the static_compiler catalog terminus is
not stringified, resulting in file resources with integer file modes in
the resultant catalog and a depreciation warning such as:

Warning: Non-string values for the file mode property are deprecated. It
must be a string, either a symbolic mode like 'o+w,a+r' or an octal
representation like '0644' or '755'.
(at /usr/lib/ruby/vendor_ruby/puppet/type/file/mode.rb:69:in `block (2
levels) in <module:Puppet>')

### DIFF
--- a/lib/puppet/indirector/catalog/static_compiler.rb
+++ b/lib/puppet/indirector/catalog/static_compiler.rb
@@ -91,9 +91,10 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
   # @param resource [Puppet::Resource] The resource to add the metadata to
   # @param metadata [Puppet::FileServing::Metadata] The metadata of the given fileserver based file
   def replace_metadata(request, resource, metadata)
-    [:mode, :owner, :group].each do |param|
-      resource[param] ||= metadata.send(param)
+    [:owner, :group].each do |param|
+      resource[param] ||= metadata.send(param).to_s
     end
+    resource[:mode] ||= metadata.send(:mode).to_s(8)
 
     resource[:ensure] = metadata.ftype
     if metadata.ftype == "file"


joshcooper:
@tdevelioglu Thank you for your contribution! Makes sense, though I'm surprised it doesn't cause this expectation to fail: https://github.com/puppetlabs/puppet/blob/master/spec/unit/indirector/catalog/static_compiler_spec.rb#L75 Could you look into that and either add a new test or fix the existing test to capture the new behavior?

tdevelioglu:
The catalog in the unit test is empty, it contains no file resources.
I'm not sure how to inject any and I kind of doubt that would be sufficient, can you help ?

joshcooper:
It would be good to fix up the test to contain a file resource, both the case where it contains a stringified mode, and case where it doesn't and should fail. We can help update the unit test.

hkenney:
Hey @tampakrap do you need some help fixing up the tests?

tdevelioglu:
@hkenney I assume you meant me :) Yeah, if you could give me some hints I could take it from there maybe ?

hkenney:
Oof yeah sorry about that! Too many GitHub handles. @joshcooper did you have some examples in mind that might help with writing this test?

MikaelSmith:
If I set a breakpoint at https://github.com/puppetlabs/puppet/blob/master/spec/unit/indirector/catalog/static_compiler_spec.rb#L73, it's never hit. The only values I see for `resource.type` are `Class` and `Stage`. This definitely deserves more investigation.

As far as updating the test for this change, ensuring the mode test confirms that the type is a String and not an Int (which may be handled by the rspec matchers once we fix the test) should be sufficient.

@hkenney will follow up on why the test appears broken.

joshcooper:
@MikaelSmith @HAIL9000 I tried updating static compiler specs for a different PR and was equally surprised that the catalog never contains `File` resources: https://github.com/puppetlabs/puppet/pull/4041#discussion_r34169301.

lib/puppet/indirector/catalog/static_compiler.rb
```diff
@@ -90,7 +90,7 @@ def find_and_replace_metadata(host, resource, file)
   # @param metadata [Puppet::FileServing::Metadata] The metadata of the given fileserver based file
   def replace_metadata(host, resource, metadata)
     [:mode, :owner, :group].each do |param|
-      resource[param] ||= metadata.send(param)
+      resource[param] ||= metadata.send(param).to_s
```
joshcooper: Actually this isn't quite right. They all need to be stringified but mode needs to be in octal. Otherwise we write out mode as "420" instead of "644". I can reproduce by trying to serve a directory containing a file. The parent directory has it's mode set to "420" and puppet will fail to create the child within.

tdevelioglu:
@joshcooper 
I'm having trouble reproducing the behavior you mention, could you explain a little more ?

tdevelioglu:
nvm, I had a default file mode set that I didn't notice. I can reproduce it now.

tdevelioglu:
Changed to stringify mode as octal. 
I did an amend, but didn't think of CI keeping copies around, so Appveyor is failing, can it be cleared?

Also found another issue. 
It seems that child resources are always inheriting the parent's metadata.

joshcooper:
Looks good, merging

PR #4053, opened 2015-06-24
(PUP-4770) Revert "(maint) Fix debug and error output"

whopper:
This reverts commit 1764674bea168a9e6619ddcbc4e698eb236742f8.
Originally, this commit changed the Solaris Zone provider's debug
and error messages to use the global utilities, which means that
they aren't prefixed with the provider context. This was done to
circumvent PUP-3088, which is now fixed.

### DIFF
--- a/spec/unit/provider/zone/solaris_spec.rb
+++ b/spec/unit/provider/zone/solaris_spec.rb
@@ -149,6 +149,25 @@ net:
         })
       end
     end
+    describe "with an invalid or unrecognized config" do
+      it "should produce an error message with provider context when given an invalid config" do
+        erroneous_zone_info =<<-EOF
+          physical: net1'
+        EOF
+
+        provider.expects(:zonecfg).with(:info).returns(erroneous_zone_info)
+        provider.expects('err').with("Ignoring '          physical: net1''")
+        provider.getconfig
+      end
+
+
+      it "should produce a debugging message with provider context when given an unrecognized config" do
+        unrecognized_zone_info = "dummy"
+        provider.expects(:zonecfg).with(:info).returns(unrecognized_zone_info)
+        provider.expects('debug').with("Ignoring zone output 'dummy'")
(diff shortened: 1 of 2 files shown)


whopper:
I played around with the zone provider to verify this is working correctly.

Before the revert:

```
Debug: Ignoring zone output '   allowed-address not specified'
```

After the revert:
```
Debug: Zone[myzone4](provider=solaris): Ignoring zone output '  allowed-address not specified'
```

MikaelSmith:
Do we have any tests that cover things like this?

whopper:
@MikaelSmith we do not (which is why the original change went in without spec updates), but I could probably add a test to ensure we get the provider context in the debug output.

MikaelSmith:
That seems like something that would be useful, since we've changed it before.

spec/unit/provider/zone/solaris_spec.rb
```diff
@@ -149,6 +149,25 @@
         })
       end
     end
+    describe "with an invalid or unrecognized config" do
```
whopper: @MikaelSmith here's my first take at a spec-based testing solution. The wording is possibly somewhat off (invalid, erroneous, unrecognized), but it covers the `provider.err` and `provider.debug` cases.

whopper:
Huh, that appveyor failure looks... unrelated?

```
Failure/Error: expect(ast.code.safeevaluate(scope)).to eq('chocolate') 
    Puppet::PreformattedError:
       Evaluation Error: Error while evaluating a Function Call, Error from DataBinding 'hiera' while looking up 'wonka::produces': uncaught throw :undefined_variable at line 2:12
```

spec/unit/provider/zone/solaris_spec.rb
MikaelSmith: Sure. An integration test could also cover the actual debug message by looking at report.logs.

PR #4056, opened 2015-06-25
(PUP-4775) Make Puppet::Node serialize env as a string in YAML

hlindberg:
Before this, a Puppet::Node would serialize to YAML in such a way
that all of its instance variables were serialized. This caused
everything in a compelte Environment to be serialized (everything
loaded and evaluated), which in turn can cause OOM, or OOStack errors.

The cause of the problem is that Puppet::Node did not correctly
include Puppet::Util::PsyckSupport which handles the Yaml serialization
API and turns them into calls: to_data_hash, and initialize_from_hash.

The older way of using 'to_yaml' (one such implementation is found in
Environment itself) never gets called when it is not at the root of the
serialization.

This commit adds the missing include and refactors the initialization
from a hash to use the correct API. It also keeps the older way of a
class method from_data_hash, since it is not known if there are any
usage of this in modules implementing indirectors/caches. (It now
delegates to initialize_from_hash to keep the implementation DRY).

### DIFF
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -9,6 +9,9 @@ class Puppet::Node
   # the node sources.
   extend Puppet::Indirector
 
+  # Asymmetric serialization/deserialization required in this class via to/from datahash
+  include Puppet::Util::PsychSupport
+
   # Use the node source as the indirection terminus.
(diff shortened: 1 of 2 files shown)


hlindberg:
Resubmitted PR since the previous was reverted in a non revertable way.

PR #4057, opened 2015-06-25
(PUP-4787) Initialize @vars of Puppet:Node with expected defaults

hlindberg:
Before this, (introduced in PUP-4775 where the serialization
was changed to use Psych support for asymetric serialization), two of
Puppet::Node's instance varaibles (@classes, and @parameters) where
initialized to nil on deserialization. When again serializing a check
was made if these where empty? (which fails).

This was caught by a roundtripping acceptance test where a
yaml-serialized node was turned into json via serialization.

The fix was to initialize to [] and {} respectively in the
initialize_from_hash method.

This also adds a unit test for a roundtrip serialization where classes
and parameters are not included in the serialized data.

### DIFF
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -22,15 +22,13 @@ class Puppet::Node
   attr_reader :server_facts
 
   def initialize_from_hash(data)
-    @name = data['name']
-    @classes = data['classes'] || []
+    @name       = data['name']       || (raise ArgumentError, "No name provided in serialized data")
+    @classes    = data['classes']    || []
     @parameters = data['parameters'] || {}
     @environment_name = data['environment']
   end
 
   def self.from_data_hash(data)
-    raise ArgumentError, "No name provided in serialized data" unless name = data['name']
-
     node = new(name)
     node.initialize_from_hash(data)
     node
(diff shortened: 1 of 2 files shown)


lib/puppet/node.rb
```diff
@@ -23,8 +23,8 @@ class Puppet::Node
 
   def initialize_from_hash(data)
     @name = data['name']
-    @classes = data['classes']
-    @parameters = data['parameters']
+    @classes = data['classes'] || []
+    @parameters = data['parameters'] || {}
     @environment_name = data['environment']
```
joshcooper: will `name` and `environment` ever have the same problem?

hlindberg: environment_name can be nil, if so, then it looks up the environment lazily. In practice 'name' can not be nil since the normal initializer requires a name. You would have to have a broken serialization for that to happen. It seems prudent to check name, and fail the same way as the normal initializer. I will add that.

hlindberg: Actually, the name is already checked by self.from_data_hash - I will instead mark the initialize_from_hash as an @api private method

hlindberg: Hm, actually, what I actually will do is to move the check, since initialize_from_hash is the rock bottom, not self_from_data_hash.

MikaelSmith: Line 53 checks whether @environment_name is nil, so that seems safe.

PR #4059, opened 2015-06-30
Some of the user attributes in Aix could be UPPERCASE 

kalugen:
and they should maintain the original casing. 

We where experiencing a strange behavior with some users using Kerberos authentication, this implies setting the "SYSTEM" attribute to a value of KRB5Afiles.

In hiera this was the (redacted) definition of one such users:

```yaml
  xxxxx:
    ensure: present
    comment: XXXXXX
    gid: 'xxxxx'
    groups:
    - ssh_grp
    managehome: 'true'
    attributes:
    - SYSTEM=KRB5Afiles
    - registry=KRB5Afiles
    - auth_domain=DOMAIN.AD
    home: "%{::default_home}/xxxxx"
    shell: "%{::default_shell}"
    uid: 'xxxxx'
```
And this is the (redacted) output from "lsuser -c xxxxx":
```bash
#name:id:pgrp:groups:home:shell:gecos:login:su:rlogin:daemon:admin:sugroups:tpath:ttys:expires:auth1:auth2:umask:registry:SYSTEM:loginretries:pwdwarntime:account_locked:minage:maxage:maxexpired:minalpha:minother:mindiff:maxrepeats:minlen:histexpire:histsize:auth_domain:fsize:cpu:data:stack:core:rss:nofiles:stack_hard:time_last_login:tty_last_login:host_last_login:unsuccessful_login_count
xxxxx:xxxxx:adm_nix:adm_nix,ssh_grp:/home/xxxxx:/usr/bin/bash:XXXXX:true:true:true:true:false:ALL:nosak:ALL:0:SYSTEM:NONE:22:KRB5Afiles:KRB5Afiles:0:0:false:0:0:-1:0:0:0:8:0:0:0:DOMAIN.AD:-1:-1:-1:-1:-1:-1:-1:-1:1435654795:/dev/pts/4:10.X.X.X:0
```

Running puppet resulted in the users always been updated, at every run, because the "SYSTEM" (UPCASE) field from lsuser was converted to a ":system"  (downcase) property. This caused a mismatch between the expected attribute keys and what was parsed from lsuser output, causing the provider to think the user was missing one of the managed attributes at each run.

### DIFF
--- a/lib/puppet/provider/aixobject.rb
+++ b/lib/puppet/provider/aixobject.rb
@@ -196,7 +196,7 @@ class Puppet::Provider::AixObject < Puppet::Provider
     end
 
     attrs.each { |val|
-      key = key_list.shift.downcase.to_sym
+      key = key_list.shift.to_sym
       properties = self.load_attribute(key, val, mapping, properties)
     }
     properties.empty? ? nil : properties


ffrank:
From what I understand, this change will have an impact on all provider classes that inherit `AixObject`. It strikes me as rather dangerous to eliminate the `downcase` in such a broad stroke.

Should we make this configurable for each concrete provider instead?

kalugen:
It's a possibility and probably trivial to implement, still I don't see
many situations in which altering the case of strings obtained from the
command line for the purpose of using them as hash keys would be useful and
not confusing to the end user.

There are just 2 providers, "user/aix.rb" and "group/aix.rb", inheriting
from AixObject and while we didn't develop a battery of rspec test againsts
this change, we still tested the solution on our Aix systems (with dozens
of users/group) and couldn't find a problem.

kalugen:
Also, there are two methods in AixObject to parse command line output:

   - parse_colon_list()
   - parse_attr_list()

While the former had the downcase, the latter has not. Making the
"donwcase" an option only for parse_colon_list() would still make the
methods behave differently.

One of the first test we made was to change the command defined in
"user/aix.rb" from "lsuser -c" to just "lsuser", which returns the
command's output in a space-separated format instead of the colon-separated
one, thus triggering parse_attr_list instead of parse_colon_list: it worked
for the uppercase property (because it was missing the forced downcase) but
introduced other problems, since one of our users attributes (the "gecos"
field) contains spaces - we use it to note the user's first name and
surname.

MikaelSmith:
Thanks for the PR. Can you also take a look at the [contributing guidelines](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md).

joshcooper:
@kalugen this makes sense, but we're a little concerned this could break existing behavior. For example, does puppet ever look at the key=value, and do something like `resource.send(key)` assuming that `key` has been downcased to match the name of a resource parameter/property? We will take a look at this and see if our acceptance tests pass, though it will be a little while before we can get to it.

kylog:
@kalugen can you file a ticket at tickets.puppetlabs.com? Your comment at the top looks like valuable context to add to that ticket. Thanks!

(And update the commit message to reference that ticket as well.)

kalugen:
Done. Let me know if you require anything else. 

For quick access, here is a link to the ticket: https://tickets.puppetlabs.com/browse/PUP-5015

kylog:
Thanks @kalugen! As mentioned above, we'd like to run this through acceptance on an aix box, but that requires a bit of coordination on our end. So hang tight, and we'll try to get to that in our next sprint.

peterhuene:
@johnduarte would it be possible to review this now that we have some semblance of AIX builds working?

johnduarte:
@kalugen and @peterhuene we are still WIP on AIX testing pipelines, but we should have this addressed by the end of the week. Once that is in place, we can test this PR through acceptance.

joshcooper:
Still waiting on AIX testing pipelines

joshcooper:
Still waiting on AIX

MikaelSmith:
AIX is available, merging.

PR #4061, opened 2015-07-02
(PUP-4814) Use ENV variables in the path of scheduled tasks folder.

digglife:
This might be a rare case, but fixing it could improve the code actually.

If a Windows Server 2003 system is upgraded from Windows Server 2000 or NT, the default `%SYSTEMROOT%` or `%WINDIR%` is `C:\WINNT` rather than `C:\Windows`. This will cause failure of `scheduled_task` resource, because [`taskscheduler.rb`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/taskscheduler.rb#L812) hardcoded the path of scheduled tasks as `C:\Windows\Tasks` when testing if a job exists.

```
Error: /Stage[main]/test/Scheduled_task[foo]: Could not evaluate: No such file or directory @ dir_initialize - C:/Windows/Tasks
```

I don't think it's a good way :

- The windows directory doesn't have to be `C:\Windows`
- The system drive doesn't have to be `C:`

It's better to use `"#{ENV['windir']}/Tasks"` here.

### DIFF
--- a/lib/puppet/util/windows/taskscheduler.rb
+++ b/lib/puppet/util/windows/taskscheduler.rb
@@ -809,7 +809,7 @@ module Win32
     # Returns whether or not the scheduled task exists.
     def exists?(job_name)
       bool = false
-      Dir.foreach('C:/Windows/Tasks'){ |file|
+      Dir.foreach("#{ENV['SystemRoot']}/Tasks"){ |file|
         if File.basename(file, '.job') == job_name
           bool = true
           break
--- a/spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
+++ b/spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
@@ -600,6 +600,16 @@ describe Puppet::Type.type(:scheduled_task).provider(:win32_taskscheduler), :if 
 
       expect(resource.provider.exists?).to eq(true)
     end
+
+    it "uses SystemRoot in the path of Task folder" do
+      Win32::TaskScheduler.unstub(:new)
+      exists_test_task = Win32::TaskScheduler.new
+      Dir.stubs(:foreach).with('D:/WinNT/Tasks').returns([])
+
+      Puppet::Util.withenv('SystemRoot' => 'D:/WinNT') do
+        exists_test_task.exists?('Test Task')
+      end
+    end
   end
 
   describe '#clear_task' do


digglife:
@puppetcla CLA signed.

lib/puppet/util/windows/taskscheduler.rb
```diff
@@ -809,7 +809,7 @@ def max_run_time=(max_run_time)
     # Returns whether or not the scheduled task exists.
     def exists?(job_name)
       bool = false
-      Dir.foreach('C:/Windows/Tasks'){ |file|
+      Dir.foreach("#{ENV['windir']}/Tasks"){ |file|
```
ferventcoder: prefer SystemRoot instead of WinDir. SystemRoot will always resolve.

ferventcoder: Here's why http://superuser.com/questions/638321/what-is-difference-between-windir-and-systemroot.

digglife: Make good sense. Learned something new. :)

digglife:
@ferventcoder  Just searched the repo with `windir`:

https://github.com/puppetlabs/puppet/search?utf8=%E2%9C%93&q=windir

update is needed for them, maybe. :)

ferventcoder:
@digglife likely - one is in markdown and the others are in spec tests. Something to look at fixing up, but not super high priority to fix. :)

ferventcoder:
Caught this in our vendored taskscheduler. Nice. :)

joshcooper:
@digglife Thank you for your contribution! Could you look at adding a unit test for this?

digglife:
@joshcooper It seems to me that the `exists?` behaviour is already covered by [`win32_scheduler_spec.rb`](https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb#L588), not directly though. Could you please let me what kind of test I should write? I'm not familiar with RSpec, but I would like to have a try.

cc @ferventcoder

joshcooper:
@digglife sounds good. I'd recommend adding a test within the "exists?" example group that does something like (I don't know if this will work verbatim, but something like it should work):

```
it "uses the current SystemRoot" do
  expect(Dir.foreach).with('D:/WinNT/Tasks').returns([])

  Puppet::Util.withenv('SystemRoot' => 'D:/WinNT') do
    @mock_task.exists?('Test Task')
  end
end
```

The `withenv` method should ensure the change to `SystemRoot` is only visible within the block, and that the old value is restored when the `withenv` method returns.

digglife:
@joshcooper Thanks for your hints! I submitted the unit test code. Please have a look.
Since the test has to be performed inside `exists?`, I unstub the new method of Win32::TaskScheduler and create a real task object rather than using the mocked one. I tried to add further tests for verifying the return value of `exists?`, but the logic is inside the `Dir.foreach` block, which is impossible for me to inject.(stub the foreach doesn't work actually, don't know if there is any other hack).

MikaelSmith:
:+1:

hkenney:
@joshcooper @ferventcoder Could one of you give a final :+1: on this before merging?

ferventcoder:
:+1:

PR #4062, opened 2015-07-02
(PDOC-16) Improve compatibility of skeleton module with puppet strings

iankronquist:
* Add .yardopts file. This tells yard to use markdown, not rdoc because we
  anticipate that this is what most people will actually want.
* Add generated .yardoc to .gitignore. Yard generates this file and you don't
  care about it.
* Convert init.pp comments to use markdown.


iankronquist:
Goofed up and made this against master, not stable.

PR #4065, opened 2015-07-02
(PUP-4810) Do not cache 0 ttl entries and evict aggressively

hlindberg:
This adds a better cache eviction algorithm to the environment cache. See commit comments for details.

### DIFF
--- a/lib/puppet/environments.rb
+++ b/lib/puppet/environments.rb
@@ -304,7 +304,12 @@ module Puppet::Environments
       if result = @cache[name]
         return result.value
       elsif (result = @loader.get(name))
-        @cache[name] = entry(result)
+        cache_entry = entry(result)
+        unless cache_entry.is_a?(NotCachedEntry)
+          @cache_expiration_service.created(result)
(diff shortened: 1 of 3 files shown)


lib/puppet/environments.rb
```diff
@@ -300,15 +309,34 @@ def search_paths
...
+        @cache_expiration_service.created(result)
+        add_entry(name, cache_entry)
         result
       end
     end
 
+    # Adds a cache entry to the cache
+    def add_entry(name, cache_entry)
+      Puppet.debug {"Caching environment '#{name}' #{cache_entry.label}"}
+      @cache[name] = cache_entry
```
thallgren: If memory consumption is a concern, then perhaps it would be a good idea to also have a maximum size on the cache? If that size is reached here, then the entry with the earliest expiration time should be evicted even if that time isn't reached yet.

hlindberg: A max env cache size could be added if that becomes a problem (separate ticket).

thallgren:
This looks good to me.

PR #4068, opened 2015-07-06
(PUP-4752) Improve validation of parameter names

hlindberg:
Before this, a badly formed parameter name would go unnoticed. This
causes a problem when not running with --strict_parameters beause
references to a non existing parameter does not cause a failure.
A user may have defined a parameter with 'define foo($Ateam) {}' and
tries to reference it with '$ateam' or '$Ateam' (the former would not
find the value, and the latter would fail with illegal parameter name).

This commit adds validation of the parameter name. It may start with a
lower case letter or underscore, and then any number of word characters.
A new issue code is added, and a new pattern since all existing patterns
(and error messages) are for fully qualified names.

This also correct typos on existing issues where \w did not output correctly.

### DIFF
--- a/lib/puppet/pops/issues.rb
+++ b/lib/puppet/pops/issues.rb
@@ -243,11 +243,15 @@ module Puppet::Pops::Issues
   end
 
   ILLEGAL_NAME = hard_issue :ILLEGAL_NAME, :name do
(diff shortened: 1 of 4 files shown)


MikaelSmith:
`paramter` in commit title?

hlindberg:
commit title spelling corrected ;-)

PR #4074, opened 2015-07-09
(PUP-4819) Add class info service

hlindberg:
This adds a class info service to be used by node classifier. See ticket for details.

### DIFF
--- a/lib/puppet/info_service/class_information_service.rb
+++ b/lib/puppet/info_service/class_information_service.rb
@@ -1,0 +1,111 @@
+require 'puppet'
+require 'puppet/pops'
+require 'puppet/pops/evaluator/literal_evaluator'
+
+class Puppet::InfoService::ClassInformationService
+
+  def initialize
+    @file_to_result = {}
+    @parser = Puppet::Pops::Parser::EvaluatingParser.new()
+  end
+
+  def classes_per_environment(env_file_hash)
+    # In this version of puppet there is only one way to parse manifests, as feature switches per environment
+    # are added or removed, this logic needs to change to compute the result per environment with the correct
+    # feature flags in effect.
+
+    unless env_file_hash.is_a?(Hash)
+      raise ArgumentError, 'Given argument must be a Hash'
+    end
+
+    result = {}
+
+    # for each environment
+    #   for each file
+    #     if file already processed, use last result or error
+    #
+    env_file_hash.each do |env, files|
+      env_result = result[env] = {}
+      files.each do |f|
+        env_result[f] = result_of(f)
+      end
+    end
+    result
+  end
+
+  private
+
+  def type_parser
+    # Safe to cache this as it would otherwise constantly build up the visitor cache
+    @@type_parser ||= Puppet::Pops::Types::TypeParser.new
+  end
+
+  def literal_evaluator
+    @@literal_evaluator ||= Puppet::Pops::Evaluator::LiteralEvaluator.new
+  end
+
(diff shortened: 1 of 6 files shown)


lib/puppet/info_service/class_information_service.rb
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def literal_evaluator
+    @@literal_evaluator ||= Puppet::Pops::Evaluator::LiteralEvaluator.new
+  end
+
+  def result_of(f)
+    entry =  @file_to_result[f]
+    if entry.nil?
+      @file_to_result[f] = entry = parse_file(f)
```
camlow325: Does the parsing of a file not need to take into account the environment in which the file is being parsed?  I'd been thinking from the JIRA conversations leading up to this that the parsing did need to take the environment into account and, therefore, that any backing cache in the `ClassInformationService` would have to be environment-specific.

spec/unit/info_service_spec.rb
```diff
@@ -0,0 +1,236 @@
...
+              :params=>[
+                {:name=>"fum_a"}
+              ]},
+          ],
+         } # end test env
+        }
+      )
+    end
+
+    it "avoids parsing file more than once when environments have same feature flag set" do
```
camlow325: What does 'same feature flag set' mean?  Maybe it doesn't mean anything specific today but will in the future?

lib/puppet/info_service/class_information_service.rb
cprice404: See comments on line 13-15.  I believe that @hlindberg just wanted to make the API more forward compatible with some environment-specific settings that he is planning for the future?

lib/puppet/info_service/class_information_service.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+    nil
+  end
+
+  # Extracts the source for the expression
+  def extract_value_source(value_expr)
+    position = Puppet::Pops::Adapters::SourcePosAdapter.adapt(value_expr)
+    position.extract_tree_text
+  end
+end
```
cprice404: Would it be worth adding an explicit `close` method to this class?  Just to give you the ability to free up resources if there ends up being anything you want to do there in the future?  And for now you could explicitly null out the instance variables to maybe help accelerate GC?

cprice404:
@hlindberg this seems good to me upon initial review... will be easier to say for sure once we've started calling into it :)  But looks good to me so far!

lib/puppet/info_service/class_information_service.rb
hlindberg: In 4.3 all environments have the same "color" - there are no feature flags similar to parser=future. The API is designed to handle such feature flags being added in the future. This means that this implementation does not have to do more than once parse per unique file, but it will need to do some in the future.

spec/unit/info_service_spec.rb
hlindberg: yes, in the future there can be one or more features (we anticipate at least one in the future, but there may be more).

lib/puppet/info_service/class_information_service.rb
hlindberg: Th way it is designed now, it only holds on to things while processing one request, then everything is GCd away. There does not seem to be a reason for this service to hold on to any state; that would be at the consuming end - keeping track of changing files, what to parse, how to update an aggregate index etc. So - no, no need for a close.

PR #4078, opened 2015-07-09
(PUP-4847) Reset Facter after changing $LOAD_PATH

MikaelSmith:
Facters behavior in 2.x was to query the $LOAD_PATH when requesting
Facts it doesn't know about, but only if querying individual facts or
querying all facts when they haven't been cached yet. This can lead to
very strange behavior, where different ways of querying facts yield
different results.

Facter 3.0 made that behavior consistent, but requires Facter.reset
when the $LOAD_PATH changes. Add that for cases besides apply and agent,
which already do Facter.reset.

### DIFF
--- a/acceptance/tests/pluginsync/4847_pluginfacts_should_be_resolvable_from_applications.rb
+++ b/acceptance/tests/pluginsync/4847_pluginfacts_should_be_resolvable_from_applications.rb
@@ -1,0 +1,58 @@
+test_name "Pluginsync'ed custom facts should be resolvable during application runs"
+
+#
+# This test is intended to ensure that custom facts downloaded onto an agent via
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Planning to add an acceptance test because this tests Facter's behavior, and we don't have a way to use Facter 3 in specs yet.

peterhuene:
I'm not sure why GitHub is not showing the CI statuses, but I've verified that this passes Travis and AppVeyor.

peterhuene:
@MikaelSmith Was this intended for master or stable?

MikaelSmith:
Stable. Also I think the acceptance test still needs work, it'll return an error code because the provider is incomplete.

PR #4080, opened 2015-07-10
(docs) Improve and shorten descriptions of epp and inline_epp functions

nfagerlund:
Now that we have core docs pages about using and writing EPP templates, we can
shorten the function docs and reduce their scope.

Committing on behalf of Garrett Guillotte, who did the revisions.


nfagerlund:
@gguillotte Here's where those docs end up; note that they're still in two places right now, pending more work on Puppet Strings.

joshcooper:
@hlindberg can you review this docs change to epp?

PR #4081, opened 2015-07-10
(PUP-4847) Fix using master to create tmpdir name

MikaelSmith:
Test config caused failures on Windows where the tmpdir name is
invalid. Fix by creating the tmpdir name with the agent.

[skip ci]

### DIFF
--- a/acceptance/tests/pluginsync/4847_pluginfacts_should_be_resolvable_from_applications.rb
+++ b/acceptance/tests/pluginsync/4847_pluginfacts_should_be_resolvable_from_applications.rb
@@ -5,10 +5,9 @@ test_name "Pluginsync'ed custom facts should be resolvable during application ru
 # pluginsync are resolvable by puppet applications besides agent/apply.
 #
 
-step "Create a codedir with a test module with external fact"
-codedir = master.tmpdir('4847-codedir')
-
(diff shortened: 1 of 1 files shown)


joshcooper:
Seeing a failure on windows, we expect exit code 1, which we get, but for the wrong reason:

```
be rake ci:test:aio SHA=977d2e1ecacc713ffcbbc6ce5a77ba46e50da3db CONFIG=config/nodes/win2012r2-rubyx64.yaml TESTS=tests/pluginsync/4847_pluginfacts_should_be_resolvable_from_applications.rb
...
jgpez71wfe0nq34.delivery.puppetlabs.net (agent-2012r2-x86_64-rubyx64) 16:07:02$ cmd.exe /c puppet resource test4847 --libdir C:/cygwin64/tmp/4847-codedir.T0AyCb/lib --factpath C:/cygwin64/tmp/4847-codedir.T0AyCb/facts
fact foo=bar, snafu=zifnab
Error: Could not run: undefined method `collect' for nil:NilClass
```

MikaelSmith:
That's actually the expected reason; I guess I could confirm it. I could also flesh out the type/provider so it doesn't error.

MikaelSmith:
Updated to no longer expect an error.

PR #4082, opened 2015-07-13
(RE-5087) Fix init script behavior on sles/suse when run by puppet

shrug:
On suse/sles 10, when executed by puppet, the init script will not start the puppet
service because startproc checks and finds a puppet process running.

This updates the init script to force start the process unless there is
a running process matching the pid in the pidfile. It also updates the status
command to only return true if the pidfile matches a running puppet pid.

This is a direct port of a fix for SLES 10 from previous PE releases.

### DIFF
--- a/ext/suse/client.init
+++ b/ext/suse/client.init
@@ -36,7 +36,6 @@
 lockfile=/var/lock/subsys/puppet
 pidfile=/var/run/puppetlabs/agent.pid
 puppetd=/opt/puppetlabs/puppet/bin/puppet
-RETVAL=0
 
 PUPPET_OPTS="agent"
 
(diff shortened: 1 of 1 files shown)


haus:
:+1:

shrug:
Tested on sles 10, seems to do the right thing:

```
Debug: Executing: '/sbin/service puppet status'
Debug: Executing: '/sbin/service puppet start'
Notice: /Service[puppet]/ensure: ensure changed 'stopped' to 'running'
Info: /Service[puppet]: Unscheduling refresh on Service[puppet]
Debug: Finishing transaction 70845210
Debug: Storing state
Debug: Stored state in 2.58 seconds
Debug: Executing: '/sbin/service puppet status'
service { 'puppet':
  ensure => 'running',
}
```

PR #4083, opened 2015-07-13
(PUP-4654) Add acceptance test for custom facts in puppet facts face

whopper:
This commit adds a new acceptance test to ensure that custom and
external facts are resolved correctly and present in the output
of the puppet facts face.

### DIFF
--- a/acceptance/tests/face/4654_facts_face.rb
+++ b/acceptance/tests/face/4654_facts_face.rb
@@ -1,0 +1,63 @@
+test_name "Puppet facts face should resolve custom and external facts"
+
+#
+# This test is intended to ensure that custom and external facts present
+# on the agent are resolved and displayed by the puppet facts face.
+#
+custom_fact = <<CFACT
+Facter.add('custom_fact') do
+  setcode do
+    'foo'
+  end
+end
+CFACT
+
+unix_external_fact = <<EFACT
+#!/bin/sh
+echo 'external_fact=bar'
+EFACT
+
+win_external_fact = <<EFACT
+@echo off
+echo external_fact=bar
+EFACT
+
+agents.each do |agent|
+  if agent['platform'] =~ /windows/
+    external_fact = win_external_fact
+    ext = '.bat'
+  else
+    external_fact = unix_external_fact
+    ext = '.sh'
+  end
+
+  step "Create custom and external facts in their default directories on the agent"
+
+  teardown do
+    on agent, "rm -rf #{agent.puppet['plugindest']}/facter"
+    on agent, "rm -rf #{agent.puppet['pluginfactdest']}/external#{ext}"
+  end
+
+  on agent, puppet('apply'), :stdin => <<MANIFEST
+  file { "#{agent.puppet['plugindest']}/facter":
+    ensure => directory,
+  }
+
+  file { "#{agent.puppet['plugindest']}/facter/custom.rb":
+    ensure  => file,
+    content => "#{custom_fact}",
+  }
+
+  file { "#{agent.puppet['pluginfactdest']}/external#{ext}":
+    ensure  => file,
+    mode    => "0755",
+    content => "#{external_fact}",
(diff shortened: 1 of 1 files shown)


acceptance/tests/face/4654_facts_face.rb
```diff
@@ -0,0 +1,63 @@
...
+    ensure  => file,
+    mode    => "0755",
+    content => "#{external_fact}",
+  }
+MANIFEST
+
+  step "Agent #{agent}: custom_fact and external_fact should be present in the output of `puppet facts`"
+  on agent, puppet('facts',
+                   '--libdir', File.join(factdir, 'lib'),
+                   '--pluginfactdest', File.join(factdir, 'facts.d')) do
```
er0ck: would it be valuable to test this with facts declared in the default paths and asking puppet facts without specifying --libdir and --pluginfactdest?

acceptance/tests/face/4654_facts_face.rb
```diff
@@ -0,0 +1,63 @@
...
+    assert_match(/"custom_fact": "foo"/, stdout)
+    assert_match(/"external_fact": "bar"/, stdout)
```
er0ck: this might be a bit loose.  maybe we should use assert_equal here?
also, supply a message when it fails so we know which one, quickly.

acceptance/tests/face/4654_facts_face.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  teardown do
+    on(agent, "rm -rf #{factdir}")
+  end
```
er0ck: best practice is to declare the teardown at the beginning of the test so readers know what is being cleaned up.

acceptance/tests/face/4654_facts_face.rb
whopper: @er0ck `puppet facts` still can only return a giant blob of all facts, so I believe we're limited to grepping through it's output. Will add failure messages.

acceptance/tests/face/4654_facts_face.rb
whopper: Yeah, makes sense. I believe there may be a simple way to get nix/windows paths out of beaker as well.

acceptance/tests/face/4654_facts_face.rb
er0ck: ah, right.  ok

acceptance/tests/face/4654_facts_face.rb
er0ck: we should probably only do that if beaker is asking puppet/facter what its paths are... otherwise we probably should be defining them in options.rb, particularly if used in more than one test.

whopper:
@er0ck updated and simplified by using the default external / custom dirs. I used `agent.puppet['plugindest']` and `agent.puppet['pluginfactdest']` to grab those dirs.

acceptance/tests/face/4654_facts_face.rb
```diff
@@ -0,0 +1,57 @@
...
+    ext = '.sh'
+  end
+
+  step "Create custom and external facts in their default directories on the agent"
+  on agent, puppet('apply'), :stdin => <<MANIFEST
+  file { "#{agent.puppet['plugindest']}/facter":
+    ensure => directory,
+  }
+
+  file { "#{agent.puppet['plugindest']}/facter/custom.rb":
```
er0ck: particularly now that we are using default dirs, these should be cleaned out via a teardown step.

whopper: @er0ck Updated. Verified that we start out with no `facter` dir in `plugindest`, and that `pluginfactdest` exists before the test, hence blowing away only the external fact.

er0ck: excellent.  :+1:

PR #4084, opened 2015-07-13
(PUP-3328) tidy should follow symlinks

clehene:


hkenney:
@MikaelSmith is going to look into this

kylog:
First off, thanks for the contribution and kudos for providing an acceptance test!

Next, a few thoughts/questions?
* it makes me a bit uneasy for this to be the *default* - e.g. this might be a breaking change for some (unknowingly even)
* should this be an attribute on tidy?
* last, specs are failing - can you take a look?

@clehene thoughts on the above?

kylog:
@clehene have you had a chance to take a look at the questions/comments above?

branan:
I'm +1 with what kylo wrote. Having this follow symlinks by default makes me incredible nervous, and dropping that bomb onto users is a breaking change. I'd be very happy to see a `follow_symlinks` parameter on the tidy type.

clehene:
Unfortunately no, and I'm not sure when / if I will.

branan:
@clehene I'm going to go ahead and close this for now since it's been inactive for a while. If/when you have time to pick it up feel free to re-open.

We'll keep the ticket open, since the general concept of following symlinks is a good enhancement request.

clehene:
@branan Sorry I couldn't get back to this. I think closing for now is fine. 
Thanks for pinging.

PR #4085, opened 2015-07-13
(PUP-4854) Shorten default PMT TEMP location on Windows

Iristyle:
- For modules that install on Windows and use a long hierarchical
   directory structure, the default TEMP path where PMT extracts the
   modules tarball can be problematic. Windows has a default maximum
   path length of 260 characters that must be carefully navigated.

   By default, the extracted temp location looks like:

   C:\ProgramData\PuppetLabs\puppet\cache\puppet-module\cache\tmp-unpackerYYYYMMDD-XXXX-xxxxxxx

   The default install location of a puppet 4.0+ module is:

   C:\ProgramData\PuppetLabs\code\environments\production\modules

   By initially extracting to such a long path, 30 valuable characters
   are wasted (92 char current temp path vs 62 char install path).

   On Windows, the TEMP directory can be used as a location to extract
   to, and this is the default for Dir.mkdir.  Depending on the user,
   this path will typically be one of the following:

   C:\Users\Administrator\AppData\Local\Temp  (41 chars)
   C:\Windows\Temp                            (15 chars)

### DIFF
--- a/lib/puppet/module_tool/applications/unpacker.rb
+++ b/lib/puppet/module_tool/applications/unpacker.rb
@@ -93,7 +93,7 @@ module Puppet::ModuleTool
       # @api private
       # @return [String] path to temporary unpacking location
       def tmpdir
-        @dir ||= Dir.mktmpdir('tmp-unpacker', Puppet::Forge::Cache.base_path)
+        @dir ||= Dir.mktmpdir('tmp', Puppet::Forge::Cache.base_path)
       end
     end
   end
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -531,7 +531,7 @@ module Puppet
       :desc     => "The module repository",
     },
     :module_working_dir => {
-        :default  => '$vardir/puppet-module',
+        :default  => (Puppet.features.microsoft_windows? ? Dir.tmpdir() : '$vardir/puppet-module'),
         :desc     => "The directory into which module tool data is stored",
     },
     :module_skeleton_dir => {


Iristyle:
@andersonmills @joshcooper not sure how you feel about this approach.  The issue could probably be resolved in a way that doesn't special case the code for Windows... but it would be a bit more involved.

lib/puppet/module_tool/applications/unpacker.rb
```diff
@@ -93,7 +93,12 @@ def move_into(dir)
       # @api private
       # @return [String] path to temporary unpacking location
       def tmpdir
-        @dir ||= Dir.mktmpdir('tmp-unpacker', Puppet::Forge::Cache.base_path)
+        base_path = nil
+        if !Puppet::Util::Platform.windows?
+          base_path = Puppet::Forge::Cache.base_path
+        end
+
+        @dir ||= Dir.mktmpdir('tmp-unpacker', base_path)
```
cowofevil: I suggest putting a comment in that clarifies the behavior of the "mktmpdir" method when a nil destination path is supplied.

joshcooper:
Makes sense. Prior to this PR, we extract to one filesystem and (maybe) install to another, which means FileUtils.mv isn't guaranteed to be atomic. But this PR doesn't change that situation.

Also, there's the issue of permissions, and writing to per-user temp directories is generally safe on Windows, e.g. LocalSystem's temp directory should not be writable by non-administrators, but might be good to confirm.

Iristyle:
@joshcooper - I should note that without this fix, `puppet module install` cannot install the current DSC module in the module pipeline acceptance tests, which is why I targeted this at stable.

However, as I understand it, only blockers should currently be targeted at stable.  Given changes can be made to the DSC pipeline to work with a nightly build of puppet-agent, rather than the 1.2.1 "gold" version, this might need to be targeted at master instead.

/cc @justinstoller

joshcooper:
@Iristyle so we could workaround this for DSC (in the short term) doing:

```
C:\work\puppet [stable +25 ~0 -0 !]> bundle exec puppet module install puppetlabs-stdlib --module_working_dir C:/Windows/Temp --debug
...
Debug: Extracting: C:/Windows/Temp/cache/tmp-unpacker20150714-3580-1kd7zvu/puppe
tlabs-stdlib-4.6.0/lib/facter/util/
Debug: Extracting: C:/Windows/Temp/cache/tmp-unpacker20150714-3580-1kd7zvu/puppe
tlabs-stdlib-4.6.0/lib/facter/util/puppet_settings.rb
Notice: Installing -- do not interrupt ...
C:/ProgramData/PuppetLabs/code/modules
└── puppetlabs-stdlib (v4.6.0)
```

However, if you use backslashes for `module_working_dir` then the PMT gets confused due to the mismatched back and forward slashes:

```
C:\work\puppet [stable +25 ~0 -0 !]> bundle exec puppet module install puppetlabs-stdlib --module_working_dir C:\Windows\Temp --debug
...
Error: Could not install package
  Package attempted to install file into
  "C:/Windows/Temp/cache/tmp-unpacker20150714-3276-hqzekd/puppetlabs-stdlib-4.6.0" under "C:\\Windows\\Temp/cache/tmp-unpacker20150714-3276-hqzekd".
```

Thoughts?

ferventcoder:
I didn't realize we had a workaround, so :+1:

Iristyle:
Not sure how I missed `module_working_dir` when looking at `Puppet::Forge::Cache.base_path` - https://github.com/puppetlabs/puppet/blob/master/lib/puppet/forge/cache.rb#L49

It still seems sensible to adjust the default here, but I can understand the risk aversion.

Iristyle:
Yeah, on second thought -- my PR breaks the ability to use `module_working_dir` on Windows and is therefore unacceptable.  Re-thinking...

Iristyle:
@joshcooper I refreshed my solution so that we don't ignore a user passed `--module_working_dir`.  For the DSC case it was necessary to shorten `tmp-unpacker` (in this case to `tmp`), otherwise we exhaust the PATH.

Iristyle:
I should also note that I tried to remove the second `cache` directory, but when doing that, PMT tries to flush all of  `Puppet::Forge::Cache::base_path` - in this case, the entire user's temp directory.  Pretty undesirable.

Iristyle:
Awaiting additional review from @joshcooper and @andersonmills

joshcooper:
:+1:

Secondary issue, is this supposed to go into 4.2.1 and PE or are branches frozen? /cc @kylog

Iristyle:
Opened backport PR against 3.x at #4090

PR #4087, opened 2015-07-14
(PUP-4476) Add bare bones lookup application

hkenney:
Add the skeleton for the `puppet lookup` application. Right now all
the arguments are accepted and parsed, but running `puppet lookup`
will only result in printing "Hello World" and the hash of options.


hkenney:
@hlindberg Let me know if this needs any changes, I'm going to keep working on the rest of the app in the meantime and I can rebase as needed.

PR #4092, opened 2015-07-16
Load systemd services from list-units

raphink:
list-unit-files and list-units list complementary information

On Debian 8 for example, `list-units` lists services
managed through the compatibility layer with sysvinit

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -20,6 +20,10 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
     output.scan(/^(\S+)\s+(disabled|enabled|masked)\s*$/i).each do |m|
       i << new(:name => m[0])
     end
+    output = systemctl('list-units', '--type', 'service', '--full', '--all',  '--no-pager')
+    output.scan(/^(?:\*)?\s+(\S+\.service)\s+(loaded)\s+/i).each do |m|
+      i << new(:name => m[0])
+    end
     return i
   rescue Puppet::ExecutionFailure
     return []
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -63,7 +63,7 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
       # flapping when simply trying to disable a masked service.
       return :mask if (@resource[:enable] == :mask) && (svc_info[:LoadState] == 'masked')
       return :true if svc_info[:UnitFileState] == 'enabled'
-      if Facter.value(:osfamily) == 'debian'
+      if Facter.value(:osfamily) == 'Debian'
         ret = debian_enabled?(svc_info)
         return ret if ret
       end


daenney:
:+1:

raphink:
Actually, this does not solve the issue with Debian 8 (and the `puppetserver` service in particular):

* by default (without this patch), the default provider on Debian 8 is `systemd`, but it won't find `puppetserver` in the unit files, so it defaults to the `init` provider (since `debian` is not default on Debian 8), which is not `enableable`
* with this patch, the `systemd` provider thinks it can manage `puppetserver`, but it actually still can't enable it, since there is not unit file
* it appears there is not link in `/etc/rc*.d` either, and systemd only uses the LSB headers of the init scripts to decide whether they are started at boot.

From what I can see, there is no way to ensure sysvinit-compatible services are enabled on Debian 8, since that would mean editing the LSB header of the script.

I'd be glad if someone can prove me wrong :smile:

mcanevet:
A solution would be to allow service's debian provider to be the default for all debian version as it used to be. When a package provides a service file, the systemd provider would be used because debian provider's self.instances will not list the service and when no systemd service file is present, it will fall back to debian provider but not init provider.

kylog:
Hmm interesting. @mcanevet's idea seems like it could fly. 

Ping also @melissa who looked into this in #3932.

MikaelSmith:
Is there a ticket related to this?

peterhuene:
@raphink Thanks for the contribution!  Judging from these changes, there's at least one (two?) bug(s) here, so we'll need a ticket filed and the commit messages amended to reference that ticket.

lib/puppet/provider/service/systemd.rb
```diff
@@ -59,7 +63,7 @@ def enabled?
       # flapping when simply trying to disable a masked service.
       return :mask if (@resource[:enable] == :mask) && (svc_info[:LoadState] == 'masked')
       return :true if svc_info[:UnitFileState] == 'enabled'
-      if Facter.value(:osfamily) == 'debian'
+      if Facter.value(:osfamily) == 'Debian'
```
haus: Would it be cleaner to just downcase the Facter.value instead of trying to make the RHS look right?

kylog:
@melissa do you have comments on this one?

haus:
This looks sane to me overall. I had one comment about downcasing, that I think would be a good change.

We do probably want to update the tests here as well.

melissa:
@kylog looking over it looks fine. I think we should experiment with @mcanevet's solution to the loading issues and see if that fixes things. Maybe @mcanevet, you would be able to submit a pull request making that change? It sounds like it could be a reasonable solution.

I am wondering though, @raphink is making it sound like this change makes puppet think it can manage certain services using systemd when it actually can't. Is that right? If so, then that would make this a misleading change. Why make it look like you can manage a service with systemd when you actually can't?

raphink:
Agreed this should probably be split, with one commit to fix the fact issue, and another one to try and address the main point.

That said, after testing, I'm not actually 100% sure this is the best option.

j-vizcaino:
@raphink I applied your fix for the Debian fact (this one only) and now things seems to run smoothly. Before that, Puppet was enabling services using SysVinit every time.
I don't see the use case for your patch using `list-units` and fear that you're going to add entries twice to the list of instances (once with `list-unit-files`, once with `list-units`).
When tackling the problem around services on Jessie, I ran `puppet resource service` (with Puppet 4.2.1). The first thing I noticed is that, services listed here come from the systemd provider but also the init/debian provider! This leads to entries being listed twice: for example, 'ntp' and 'ntp.service'. Is this the way it's supposed to work?

mcanevet:
@melissa I just opened an issue on Jira (https://tickets.puppetlabs.com/browse/PUP-5016) and created a PR to fix according to my comment (#4146)

whopper:
From what I understand, `list-units` will include services that don't actually have unit files, but are managed by the systemd-sysv compatibility layer - `systemd-sysv-generator` automatically generates unit files that points at init scripts in `/etc/init.d`. (I believe this is what @raphink was referencing earlier).

For services that *do* have unit files, the duplicate resource entries for services are already a problem, as we're picking up both the `name.service` unit files and the old init scripts in `/etc/init.d`, both of which actually seem to manage the same thing.

From what I can tell, fixing the existing problems for systemd in debian could take several more involved steps. I'm not sure what implications `list-units` would have, but I don't believe it will resolve the current issues, nor will making the debian provider a second default, as we don't currently actually allow multiple default providers.

whopper:
@raphink we're going to be taking a closer look at a more comprehensive solution to systemd's brokenness on Debian soon, so closing this PR. Thanks for taking a look!

raphink:
@whopper Great. Thanks for looking into this!

PR #4094, opened 2015-07-17
(RE-5032) fully qualify the label in the osx puppet plist

heathseals:
This commit fully qualifies the label in the osx puppet plist using the
identifier plus service name.  Some formatting was also cleaned up.

### DIFF
--- a/ext/osx/puppet.plist
+++ b/ext/osx/puppet.plist
@@ -10,7 +10,7 @@
                 <string>en_US.UTF-8</string>
         </dict>
         <key>Label</key>
-        <string>puppet</string>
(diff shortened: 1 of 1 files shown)


joshcooper:
Thanks @heathseals I think we want to target stable for this fix, but not merge until after we tag and release puppet 4.2.1

kylog:
Agree on targeting stable. Also can you prepend "(maint)" to the commit summary line? That will be kindness for our release script, which only knows about PUP as a ticketed project and maint/doc/packaging as non-ticketed commit tags..

heathseals:
closing and will retarget to stable.

PR #4095, opened 2015-07-20
(maint) fully qualify the label in the osx puppet plist

heathseals:
This commit fully qualifies the label in the osx puppet plist using the
identifier plus service name.  Some formatting was also cleaned up.


MikaelSmith:
Merge after 4.2.1 is tagged.

PR #4097, opened 2015-07-21
PUP-4918 - Updated logrotate and init scripts

lzap:
This patch updates logrotate script and aligns it what is defined in
defaults.rb, specifically:

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L1036-L1046

and

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L1258-L1274

Because of logrotate and puppet during restart were both "fighting" for correct
file permissions, this was causing issues with SELinux. Also fixed paths which
has been incorrect for Fedora/RHEL systems (`/var/log/puppetlabs` vs
`/var/log/puppet`). It contains both now which seems to work just fine.

I am not adding logrotate scriplet for master tho as I don't know if it
supports signals. When running master under webrick one should not expect fully
working log rotation tho. Open for discussion.

As part of this patch, I am also updating Red Hat startup scripts:

* RHEL6 traditional UNIX-like "reload" command was actually sending HUP signal
 and therefore restarting daemon instead correct USR1
* RHEL7 systemd unit was extended with "reload" command using USR1 signal

### DIFF
--- a/ext/redhat/logrotate
+++ b/ext/redhat/logrotate
@@ -1,9 +1,21 @@
-/var/log/puppetlabs/puppet.log {
+# Generic logrotate config for Red Hat systems. Works with both official
+# Fedora and RHEL packages as well as PuppetLabs distributions.
+
+/var/log/puppetlabs/masterhttp.log /var/log/puppet/masterhttp.log {
+  compress
   missingok
   notifempty
-  create 0644 puppet puppet
+  nocreate
+}
+
+/var/log/puppetlabs/puppetd.log /var/log/puppet/puppetd.log {
+  compress
+  missingok
+  notifempty
+  nocreate
   sharedscripts
   postrotate
-    [ -e /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1 || true
+   ([ -x /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1) ||
+    ([ -x /usr/bin/systemctl ] && /usr/bin/systemctl kill -s USR2 puppet.service > /dev/null 2>&1) || true
   endscript
 }
--- a/ext/debian/puppet.logrotate
+++ b/ext/debian/puppet.logrotate
@@ -1,11 +1,20 @@
-/var/log/puppetlabs/puppet.log {
-  missingok
-  sharedscripts
-  create 0644 puppet puppet
+/var/log/puppetlabs/masterhttp.log /var/log/puppet/masterhttp.log {
   compress
   rotate 4
+  missingok
+  notifempty
+  nocreate
+}
 
+/var/log/puppetlabs/puppetd.log /var/log/puppet/puppetd.log {
+  compress
+  rotate 4
+  missingok
+  notifempty
+  nocreate
+  sharedscripts
   postrotate
-    [ -e /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1 || true
+   ([ -x /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1) ||
+    ([ -x /usr/bin/systemctl ] && /usr/bin/systemctl kill -s USR2 puppet.service > /dev/null 2>&1) || true
   endscript
 }
--- a/ext/redhat/client.init
+++ b/ext/redhat/client.init
@@ -54,8 +54,16 @@ stop() {
 }
 
 reload() {
-    echo -n $"Restarting puppet agent: "
+    echo -n $"Reloading puppet agent: "
     killproc $pidopts $puppetd -HUP
+    RETVAL=$?
+    echo
+    return $RETVAL
+}
+
+rotate() {
+    echo -n $"Reopening log files for puppet agent: "
+    killproc $pidopts $puppetd -USR2
     RETVAL=$?
     echo
     return $RETVAL
@@ -91,6 +99,9 @@ case "$1" in
     restart)
         restart
     ;;
+    rotate)
+        rotate
+    ;;
     reload|force-reload)
         reload
     ;;
@@ -109,7 +120,7 @@ case "$1" in
         genconfig
     ;;
     *)
-        echo $"Usage: $0 {start|stop|status|restart|reload|force-reload|condrestart|once|genconfig}"
+        echo $"Usage: $0 {start|stop|status|restart|reload|force-reload|condrestart|rotate|once|genconfig}"
         exit 1
 esac
 
--- a/ext/redhat/puppet.spec.erb
+++ b/ext/redhat/puppet.spec.erb
@@ -181,12 +181,15 @@ echo "D /var/run/%{name} 0755 %{name} %{name} -" > \
 # Create puppet modules directory for puppet module tool
 mkdir -p %{buildroot}%{_sysconfdir}/%{name}/modules
 
-
 # Install a NetworkManager dispatcher script to pickup changes to
 # # /etc/resolv.conf and such (https://bugzilla.redhat.com/532085).
 mkdir -p %{buildroot}%{_sysconfdir}/NetworkManager/dispatcher.d
 cp -pr ext/puppet-nm-dispatcher \
   %{buildroot}%{_sysconfdir}/NetworkManager/dispatcher.d/98-%{name}
+
+# Masterhttp.log is created on server start and should be distributed
+# with correct permissions
+touch %{buildroot}%{_localstatedir}/log/puppet/masterhttp.log
 
 %files
 %defattr(-, root, root, 0755)
@@ -280,6 +283,7 @@ cp -pr ext/puppet-nm-dispatcher \
 %{_sysconfdir}/puppet/environments/example_env/README.environment
 %{_mandir}/man8/puppet-ca.8.gz
 %{_mandir}/man8/puppet-master.8.gz
+%attr(660, puppet, puppet) %{_localstatedir}/log/puppet/masterhttp.log
 
 # Fixed uid/gid were assigned in bz 472073 (Fedora), 471918 (RHEL-5),
 # and 471919 (RHEL-4)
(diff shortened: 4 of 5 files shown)


ext/redhat/logrotate
```diff
@@ -1,9 +1,23 @@
...
+  compress
+  missingok
+  notifempty
+  # permissions/owner must be aligned with puppet/defaults.rb
+  create 0640 root root
   sharedscripts
   postrotate
-    [ -e /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1 || true
+   ([ -e /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1) ||
+    ([ -x /usr/bin/systemctl ] && /usr/bin/systemctl reload puppet > /dev/null 2>&1) || true
```
lzap: This now works both on systemd and non-systemd Red Hat systems...

eLobato:
Working well on `el7` :+1:

MikaelSmith:
@haus does this change look appropriate?

MikaelSmith:
Thank you for your contribution. There are a couple guidelines we require contributors to follow at https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md, specifically filing a JIRA ticket and updating the commit message to conform to the described format.

ext/redhat/logrotate
```diff
@@ -1,9 +1,23 @@
...
+# Generic logrotate config for Red Hat systems. Works with both official
+# Fedora and RHEL packages as well as PuppetLabs distributions.
+
+/var/log/puppetlabs/masterhttp.log /var/log/puppet/masterhttp.log {
+  compress
   missingok
   notifempty
-  create 0644 puppet puppet
+  # permissions/owner must be aligned with puppet/defaults.rb
+  create 0660 puppet puppet
```
ffrank: Do we *need* to specify owner and permissions? Does not logrotate pick those up from the existing files automatically if left to its own devices?

MikaelSmith:
`puppetdlog` isn't supposed to be needed on known operating systems... could this be a sign that the POSIX syslog service is unavailable to us on SELinux?

ext/redhat/logrotate
```diff
@@ -1,9 +1,23 @@
...
+}
+
+/var/log/puppetlabs/puppetd.log /var/log/puppet/puppetd.log {
```
ffrank: Do we need the fallback log in the default logroate configuration? Documentation suggests otherwise:

      :desc => "The fallback log file. This is only used when the `--logdest` option
        is not specified AND Puppet is running on an operating system where both
        the POSIX syslog service and the Windows Event Log are unavailable. (Currently,
        no supported operating systems match that description.)

ext/redhat/logrotate
lzap: Actually deleting `create` statement is sufficient. Logrotate will inherit from existing logs. More than that, if puppet is able to create missing log files automatically (even via SIGUSR1/2) then I can add `nocreate`.

lzap: Assuming that and amending the change, it looks like this will work with SELinux which will do proper transition as well.

ext/redhat/logrotate
lzap: Well I am under impression that logrotate configuration won't hurt. Imagine situation when syslog is down and puppet falls back, this will prevent from filling the volume with logs effectively.

lzap:
@MikaelSmith sure will do the ticketing excercise once we agree this is acceptable solution :-)

ffrank:
Cool thanks for following up and making changes.

Finally, `USR1` appears as an odd choice for `reload`, to me at least. It will not cause the agent to consider configuration changes at all.

In recent Debian, the `rsyslog` initscript has received a (non-LSB?) `rotate` action (corresponding to `USR2` in puppet). Would something along those lines be a viable route?

lzap:
Ok implemented new `rotate` action for sysvinit script and also changed the logrotate systemd statement to actually send USR2 signal via `systemctl kill -s USR2 puppet.service` instead of sending `reload` command. Although systemd systems usually use journald, some distros still have syslog enabled in paralel.

If there are no objections I am creating a ticket for this one.

lzap:
What is the status?

Also one question, we have a report in Fedora about incorrect behavior (https://bugzilla.redhat.com/show_bug.cgi?id=1246238) and I wonder why the systemd unit in puppet upstream does leverage `--no-daemonize` instead of default and generally preferred forking.

I could incorporate the change into this PR if you want. Something like:

```diff
diff --git a/ext/systemd/puppet.service b/ext/systemd/puppet.service
index 7050c9f..7dec513 100644
--- a/ext/systemd/puppet.service
+++ b/ext/systemd/puppet.service
@@ -1,15 +1,15 @@
 [Unit]
 Description=Puppet agent
 Wants=basic.target
-After=basic.target network.target
+After=basic.target network.target puppetmaster.service

 [Service]
 EnvironmentFile=-/etc/sysconfig/puppetagent
 EnvironmentFile=-/etc/sysconfig/puppet
+Type=forking
-ExecStart=/opt/puppetlabs/puppet/bin/puppet agent $PUPPET_EXTRA_OPTS --no-daemonize
+ExecStart=/opt/puppetlabs/puppet/bin/puppet agent $PUPPET_EXTRA_OPTS
 KillMode=process
+PIDFile=/run/puppet/agent.pid
```

lzap:
I have amended one additional change which is related to the incorrect log file permissions. The package should distribute the file which is being created on every agent/server restart:

```diff
diff --git a/ext/redhat/puppet.spec.erb b/ext/redhat/puppet.spec.erb
index b2a6185..7b8681a 100644
--- a/ext/redhat/puppet.spec.erb
+++ b/ext/redhat/puppet.spec.erb
@@ -181,13 +181,16 @@ echo "D /var/run/%{name} 0755 %{name} %{name} -" > \
 # Create puppet modules directory for puppet module tool
 mkdir -p %{buildroot}%{_sysconfdir}/%{name}/modules
 
-
 # Install a NetworkManager dispatcher script to pickup changes to
 # # /etc/resolv.conf and such (https://bugzilla.redhat.com/532085).
 mkdir -p %{buildroot}%{_sysconfdir}/NetworkManager/dispatcher.d
 cp -pr ext/puppet-nm-dispatcher \
   %{buildroot}%{_sysconfdir}/NetworkManager/dispatcher.d/98-%{name}
 
+# Masterhttp.log is created on server start and should be distributed
+# with correct permissions
+touch %{buildroot}%{_localstatedir}/log/puppet/masterhttp.log
+
 %files
 %defattr(-, root, root, 0755)
 %doc LICENSE README.md examples
@@ -280,6 +283,7 @@ cp -pr ext/puppet-nm-dispatcher \
 %{_sysconfdir}/puppet/environments/example_env/README.environment
 %{_mandir}/man8/puppet-ca.8.gz
 %{_mandir}/man8/puppet-master.8.gz
+%attr(660, puppet, puppet) %{_localstatedir}/log/puppet/masterhttp.log
 
 # Fixed uid/gid were assigned in bz 472073 (Fedora), 471918 (RHEL-5),
 # and 471919 (RHEL-4)
```

HAIL9000:
@haus Could you weigh in on this one please?

ext/redhat/logrotate
```diff
@@ -1,9 +1,21 @@
...
+  compress
+  missingok
+  notifempty
+  nocreate
   sharedscripts
   postrotate
-    [ -e /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1 || true
+   ([ -e /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1) ||
```
haus: nitpick. the test here should be `-x` as we are executing the init script directly.

haus:
Aside from my `-x` comment, this looks good. We should ticket making the same changes on the debian side so we don't have too much drift between the two.

lzap:
Ok made the same change for Debian, I guess systemd fallback is also relevant on debian systems now.

ffrank:
Since nobody has brought this up in writing, apparently: Are we sure that the `agent` service's `reload` action should just kick of a new round-trip to the master and catalog transaction?

Users will expect `reload` to actually pick up changes in configuration files. I disbelieve that this is covered after this change.

lzap:
I don't follow. Here is the actual code: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/daemon.rb#L109-Lundefined

```
:HUP => :restart, :USR1 => :reload, :USR2 => :reopen_logs
```

Before the patch, on RHEL6 "reload" sent HUP and on RHEL7 this was not implemented.

With the patch, on both platforms the "reload" command sends USR1 which is, I believe, correct. Maybe I am looking on the wrong piece of code, but I have tested this and it seems to work. Please elaborate.

ffrank:
At the bottom of the [man page](https://docs.puppetlabs.com/references/latest/man/agent.html) it states that USR1 causes the `agent` to

> Immediately retrieve and apply configurations from the puppet master.

There is no mention about actually reloading agent configuration. That what HUP is for.

> Restart the puppet agent daemon.

I don't believe that the agent has a reload action in the sense of "re-read process configuration without terminating the service (briefly)". So HUP is the best if awkward fit.

lzap:
Oh ok the `:reload` method name was a bit misleading. I followed the code and I can confirm what man page says. Good catch, changed it to `HUP` in both cases.

branan:
@haus it looks like Debian got added here. Could you give this one final check? Thanks!

@lzap Since this now also includes debian, could you please update the commit message so it's a bit clearer what's contained in the PR.

haus:
This looks reasonable to me.

lzap:
> @lzap Since this now also includes debian, could you please update the commit message so it's a bit clearer what's contained in the PR.

Sure! Done.

-- 
Later,
 Lukas #lzap Zapletal

PR #4098, opened 2015-07-22
(docs) Update description of cfacter setting for Facter 3.0

nfagerlund:


joshcooper:
Looks good, just waiting until puppet#stable is unfrozen.

I'm thinking we should (in a separate non-docs PR), deprecate the `cfacter` setting and remove the the current hook logic, and possibly change `Puppet.features.cfacter?` to always return true... Or do we still use ruby facter for some platforms (while we port native facter)? /cc @peterhuene @MikaelSmith

MikaelSmith:
There are no platforms where we ship Ruby facter and native facter. There are still environments where we use Ruby facter obviously (install from gems). The cfacter feature tries to `require 'cfacter'`, so if we switched it to always return true we'd have to erase all functionality in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/feature/cfacter.rb. Which sounds reasonable.

kylog:
Agreed on deprecating the `cfacter` setting, and removing in puppet 5.

I guess I'd just leave `Puppet.features.cfacter?` as-is on the assumption that anyone tinkering around enough to want to set the `cfacter` setting true for some bizarre reason, actually is trying to get the original "cfacter" packages, so let them go for it.

But that part is mostly a don't-care for me. Deprecating the setting, though, is definitely something we should do.

joshcooper:
Filed deprecation as PUP-4932 for 4.3.0, and removing the setting in PUP-4933

PR #4100, opened 2015-07-23
(PUP-4919) static compiler always sets ensure parameter to 'file' if a

tdevelioglu:
The static_compiler catalog terminus always sets ensure to 'file' on
file resources as long as the resource has a source parameter declared.
It does this even if the ensure parameter is explicitly declared.

This is inconsistent with the behavior of the normal compiler, where a
file's ensure parameter gets enforced regardless of the value of the
source parameter.

This commit skips file resources with ensure set to 'absent' in the static
compiler.

### DIFF
--- a/lib/puppet/indirector/catalog/static_compiler.rb
+++ b/lib/puppet/indirector/catalog/static_compiler.rb
@@ -41,6 +41,8 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
     raise "Did not get catalog back" unless catalog.is_a?(model)
 
     catalog.resources.find_all { |res| res.type == "File" }.each do |resource|
+      next if resource[:ensure] == 'absent'
+
       next unless source = resource[:source]
       next unless source =~ /^puppet:/
 


MikaelSmith:
This looks appropriate. @joshcooper or @hlindberg have any background on the static compiler? This seems like a pretty obvious difference in behavior.

tdevelioglu:
The alternative would be to not overwrite the ensure parameter with metadata.ftype further along, but I can't really say what  unintended other effect that could have as I haven't tested.

Skipping absent files has the added benefit of not fetching file metadata when it's not needed.

hlindberg:
The change looks sane, but I do not understand if this could have consequences for the recurse case. Need input from the community on this.

MikaelSmith:
Hi @tdevelioglu, we'd like to see some validation that this also works in the recurse case - i.e. removing a whole directory. Could you also look at adding spec tests to https://github.com/puppetlabs/puppet/blob/master/spec/unit/indirector/catalog/static_compiler_spec.rb?

HAIL9000:
@tdevelioglu Thanks for the contribution!

We decided not to let the issues with the static compiler specs hold this up, I will update PUP-4934 to reflect the fact that the tests (once fixed) will want a test case for this scenario.

PR #4101, opened 2015-07-24
(PUP-2289) Add test when hosts file is malformed

johnduarte:
This commit adds an acceptance test to validate that puppet does
not delete existing host file entries when the hosts file contains
malformed content.

### DIFF
--- a/acceptance/tests/resource/host/pup_2289_should_not_destroy_data_when_malformed.rb
+++ b/acceptance/tests/resource/host/pup_2289_should_not_destroy_data_when_malformed.rb
@@ -1,0 +1,25 @@
+test_name "should not delete data when existing content is malformed"
+agents.each do |agent|
+  file = agent.tmpfile('host-not-delete-data')
+
+  teardown do
+    on(agent, "rm -f #{file}", :acceptable_exit_codes => (0..255))
+  end
+
+  step "(setup) populate test file with host information"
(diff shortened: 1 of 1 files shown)


joshcooper:
I'd be in favor of targeting stable.

Will this test work on windows? I think `printf` should be available via cygwin bash, alternatively you could use puppet to manage the file content, e.g.

```
content=<<END
127.0.0.2 existing alias
==
END
on(host, puppet("resource file '#{path}' ensure=file content='#{content}'"))
```

johnduarte:
@joshcooper yes, this works on Windows and is modeled on the other `host` resource tests in the same directory.

~~~~
o8snpyrxjkgmukp.delivery.puppetlabs.net 15:56:09$ printf '127.0.0.2 existing alias
' > C:/cygwin64/tmp/host-not-delete-data.0rRHeV

o8snpyrxjkgmukp.delivery.puppetlabs.net executed in 0.09 seconds
~~~~

johnduarte:
@joshcooper would you prefer that this be changed to use puppet to manage the file? Should this PR be closed and re-targeted to stable?

joshcooper:
@johnduarte I think closing this and targeting stable would be good (moar test coverage). I'm good with using `printf` like you're doing

johnduarte:
Closing in order to target stable.

PR #4102, opened 2015-07-24
(MAINT) Fix duplicate key which ruby 2.2 complains about

jasperla:
Ruby 2.2 complains with:
```
/usr/local/lib/ruby/site_ruby/2.2/puppet/defaults.rb:488: warning: duplicated key at line 489 ignored: :queue_type
```

The `:queue_type` is obviously duplicated so it's safe to remove one of them.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -486,10 +486,6 @@ module Puppet
       :default    => "stomp",
       :desc       => "Which type of queue to use for asynchronous processing.",
     },
-    :queue_type => {
-      :default    => "stomp",
-      :desc       => "Which type of queue to use for asynchronous processing.",
-    },
     :queue_source => {
       :default    => "stomp://localhost:61613/",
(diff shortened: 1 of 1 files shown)


kylog:
This is a trivial change which is fine to merge IMHO. 

But @jasperla just as a heads-up: puppet#3.x isn't tested against ruby 2.2 intentionally: there were a number of additional changes needed so that puppet 4 could be tested against ruby 2.2. And, in general, making 3.x work with ruby 2.2 would be a larger problem than it was for 4.x because 3.x also supports ruby 1.8 and uses some older gems (active*), both of which may lead to further 2.2 entanglements.

jasperla:
Currently we only have Puppet 3.x on OpenBSD where Ruby 2.2 has recently become the default Ruby version. I am working Puppet 4.x but it's not finished yet.  Though I won't be spending any time on fixing additional Ruby 2.2 "issues".

PR #4105, opened 2015-07-27
(PUP-4757) added puppet-agent support for PE-only platforms

kevpl:
[skip ci]

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -156,7 +156,16 @@ module Puppet
           on host, "curl -o /etc/apt/sources.list.d/#{list_filename} #{list_url}"
           on host, "apt-get update"
         else
-          host.logger.notify("No repository installation step for #{platform} yet...")
+          if project == 'puppet-agent'
+            opts = {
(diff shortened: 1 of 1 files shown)


acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -156,7 +156,16 @@ def install_repos_on(host, project, sha, repo_configs_dir)
...
           on host, "apt-get update"
         else
-          host.logger.notify("No repository installation step for #{platform} yet...")
+          if project == 'puppet-agent'
+            opts = {
+              :puppet_collection => 'PC1',
+              :puppet_agent_sha => ENV['SHA'],
+              :puppet_agent_version => ENV['SUITE_VERSION'] || ENV['SHA']
+            }
+            install_puppet_agent_dev_repo_on(agent, opts)
```
MikaelSmith: This effectively adds support for SLES, OSX, and Cumulus? See https://github.com/puppetlabs/beaker/blob/d2860c7a978a2ab534984813ac975346e21c4dbb/lib/beaker/dsl/install_utils/foss_utils.rb#L1074-L1101. It'd be nice if the commit message mentioned what effect we expect.

PR #4110, opened 2015-07-29
(PUP-4941) Add warnings for app management keywords

hlindberg:
This adds warnings when application managemnt keywords 'application',
'produces' and 'consumes' are used as bare words.

In order to support this, it was not enough to add them as Reserved, as
that means that an error is raised. Instead, the ReservedWord AST model
object got a flag, to describe that a token is reserved "in the future".
The checking logic now checks this flag, and emitts a warning or an
error.

Further, in preparation for more work, the new tokns were added as
APPLICATION_R, PRODUCES_R, etc. to reflect that they are reserved. In
the grammar, these tokens leads to the construction of a ReservedWord
AST element. Later, when a feature switch is introduced, non _R variants
of the tokens are delivered instead of the _R. This means that the
grammar can use APPLICATION token where it has meaning, and
APPLICATION_R for the reserved case (i.e. the grammar is capable of
handling both cases).


spec/unit/pops/parser/lexer2_spec.rb
```diff
@@ -91,12 +91,25 @@ def epp_tokens_scanned_from(s)
     "true"     => :BOOLEAN,
     "in"       => :IN,
     "unless"   => :UNLESS,
+    "private"  => :PRIVATE,
+    "type"     => :TYPE,
+    "attr"     => :ATTR,
```
thallgren: I don't understand this addition in context of PUP-4941. How was 'private', 'type', and 'attr' handled earlier?

hlindberg: They were just missing in the list of keywords being tested. It is unrelated. I forgot to make a comment to that effect.

PR #4114, opened 2015-07-30
(PUP-4196) Remove "{}" from variables in systemd files

haus:
The "{}" prevent systemd from interpreting values correctly and causes
services to fail to start. {} indicates that the variable is a single
argument[1].

This is largely a cherry-pick of the fix for PUP-3228.

[1] - http://www.freedesktop.org/software/systemd/man/systemd.service.html


mckern:
giphy

SYSTEMD

mckern:
:+1:

PR #4115, opened 2015-07-30
(FACT-1123) Adding acceptance test for facter with invalid locale

shaigy:
This commit adds an acceptance test to make sure facter is not crashing
when locale is set to an invalid value


MikaelSmith:
This should be added to Facter, not Puppet.

PR #4116, opened 2015-07-30
(PUP-3694) puppet cert fingerprint invalidhost returns zero instead of non-zero [#puppethack]

rnelson0:
Enforce a non-zero return code for errors in the puppet cert print|fingerprint functions.


rnelson0:
I have issues with the rspec tests for puppet (as noted in #puppethack) so I plugged along without tests but I would be glad to craft or adjust tests as directed and rebase the commit.

rnelson0:
I figured out the checks through some travis-and-error! Ready for review.

PR #4118, opened 2015-07-31
(PUP-1189) declare report handlers with overwrite

binford2k:
Previously, if a report handler failed during initialization, it wasn't
cleaned up. So the second time a report was submitted, Puppet would try
to instantiate a new class, and then fail because it had already been
instantiated. This instructs the class loader to overwrite if needed.

### DIFF
--- a/spec/unit/reports_spec.rb
+++ b/spec/unit/reports_spec.rb
@@ -43,6 +43,11 @@ describe Puppet::Reports, " when registering report types" do
     Puppet::Reports.register_report(:testing) { }
   end
 
+  it "should allow a failed report to be redefined" do
+    expect { Puppet::Reports.register_report(:testing) { raise 'failed report' } }.to raise_error(RuntimeError)
+    Puppet::Reports.register_report(:testing) { }
+  end
+
   it "should extend the report type with the Puppet::Util::Docs module" do
     mod = stub 'module', :define_method => true
 
--- a/lib/puppet/reports.rb
+++ b/lib/puppet/reports.rb
@@ -54,7 +54,11 @@ class Puppet::Reports
   def self.register_report(name, options = {}, &block)
     name = name.intern
 
-    mod = genmodule(name, :extend => Puppet::Util::Docs, :hash => instance_hash(:report), :block => block)
+    mod = genmodule(name,
+                    :extend    => Puppet::Util::Docs,
+                    :hash      => instance_hash(:report),
+                    :overwrite => true,
+                    :block     => block)
 
     mod.useyaml = true if options[:useyaml]
 


HAIL9000:
@binford2k Took a look at this in the PR triage and agreed it looks good. We were wondering if you could add a quick spec test for this?

MikaelSmith:
I don't have a complete picture, but it seems like a test in spec/unit/reports_spec.rb that throws an error in the block passed to register_report (and calls multiple times) would be similar to the ticket.

binford2k:
ping @HAIL9000 done, thanks!

HAIL9000:
@MikaelSmith I'm :+1: on this with the unit test. What do you think?

MikaelSmith:
I don't see it causing an error if I remove the `:overwrite` option. So I probably don't completely understand the circumstances around PUP-1189.

HAIL9000:
Oh it looks like Travis is upset about the commit message. @binford2k could you go ahead and prepend "(PUP-1189)" to the front of the spec test commit message?

spec/unit/reports_spec.rb
```diff
@@ -43,6 +43,12 @@
     Puppet::Reports.register_report(:testing) { }
   end
 
+  it "should allow a failed report to be redefined" do
+    Puppet::Reports.expects(:genmodule).returns(Module.new).twice
```
MikaelSmith: You want to remove this expects, it blocks the error from occuring. Instead, the next line should expect an exception. If I do that I get
```
  1) Puppet::Reports  when registering report types should allow a failed report to be redefined
     Failure/Error: Puppet::Reports.register_report(:testing) { }
     Puppet::ConstantAlreadyDefined:
       Class Testing is already defined in Puppet::Reports
     # ./lib/puppet/util/classgen.rb:164:in `handleclassconst'
     # ./lib/puppet/util/classgen.rb:128:in `genthing'
     # ./lib/puppet/util/classgen.rb:60:in `genmodule'
     # ./lib/puppet/reports.rb:57:in `register_report'
     # ./spec/unit/reports_spec.rb:51:in `block (2 levels) in <top (required)>'
```
without the fix, and test passes with the fix.

spec/unit/reports_spec.rb
```diff
@@ -43,6 +43,11 @@
     Puppet::Reports.register_report(:testing) { }
   end
 
+  it "should allow a failed report to be redefined" do
+    expect { Puppet::Reports.register_report(:testing) { raise 'failed report' } }.to raise_error(RuntimeError)
+    Puppet::Reports.register_report(:testing) { }
+  end
+
```
kylog: This seems like a confusing spec test for two reasons:
1) it's not just *failed* reports that should be able to be reloaded, it's all reports
2) if a report *is* reloaded, we'd expect it to be reloaded the same way

To that end, I'm wondering if the spec test should be something like:
```
  it "should allow a successful report to be reloaded" do
    Puppet::Reports.register_report(:testing) { }
    Puppet::Reports.register_report(:testing) { }
  end

  it "should allow a failed report to be reloaded and show the correct exception both times" do
    expect { Puppet::Reports.register_report(:testing) { raise TypeError, 'failed report' } }.to raise_error(TypeError)
    expect { Puppet::Reports.register_report(:testing) { raise TypeError, 'failed report' } }.to raise_error(TypeError)
  end
```

kylog:
I put up #4172 with that spec test idea. Closing this one in favor of that. Please chime in there with comments (or re-open this if that makes sense).

PR #4121, opened 2015-07-31
(PUP-4605) Unmask before attemping to start

melissa:
Prior to this commit, if a service was masked, and a user tried to start
the service, it would throw an error because it was masked. Even if the
user enabled the service (which calls unmask), an error would be thrown
because 'start' would be run before 'enable'. This commit ensures that
the start command has the ability to unmask a service before trying to
start it.


ScottGarman:
:+1:  if tests pass

melissa:
@trevor-vaughan you did the original work to add in masking support. Would you mind taking a look at this?

shrug:
:+1:

PR #4123, opened 2015-08-03
(docs) Improve map function description.

gguillotte:
Clarify how the `map` function works and provide more context in the
examples.

### DIFF
--- a/lib/puppet/functions/map.rb
+++ b/lib/puppet/functions/map.rb
@@ -1,33 +1,68 @@
-# Applies a parameterized block to each element in a sequence of entries from the first
-# argument and returns an array with the result of each invocation of the parameterized block.
+# Applies a [lambda](http://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html)
+# to every value in a data structure and returns an array containing the results.
 #
-# This function takes two mandatory arguments: the first should be an Array, Hash, or of Enumerable type
-# (integer, Integer range, or String), and the second a parameterized block as produced by the puppet syntax:
+# This function takes two mandatory arguments, in this order:
 #
-#       $a.map |$x| { ... }
-#       map($a) |$x| { ... }
+# 1. An array or hash the function will iterate over.
+# 2. A lambda, which the function calls for each element in the first argument. It can
+# request one or two parameters.
 #
-# When the first argument `$a` is an Array or of enumerable type, the block is called with each entry in turn.
-# When the first argument is a hash the entry is an array with `[key, value]`.
+# @example Using the `map` function
 #
-# @example Using map with two arguments
+# `$transformed_data = $data.map |$parameter| { <PUPPET CODE BLOCK> }`
 #
-#      # Turns hash into array of values
-#      $a.map |$x|{ $x[1] }
+# or
 #
-#      # Turns hash into array of keys
-#      $a.map |$x| { $x[0] }
+# `$transformed_data = map($data) |$parameter| { <PUPPET CODE BLOCK> }`
 #
-# When using a block with 2 parameters, the element's index (starting from 0) for an array, and the key for a hash
-# is given to the block's first parameter, and the value is given to the block's second parameter.args.
+# When the first argument (`$data` in the above example) is an array, Puppet passes each
+# value in turn to the lambda.
 #
-# @example Using map with two arguments
+# @example Using the `map` function with an array and a one-parameter lambda
 #
-#      # Turns hash into array of values
-#      $a.map |$key,$val|{ $val }
+# ~~~ puppet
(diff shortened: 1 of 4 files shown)


MikaelSmith:
Travis is enforcing a commit message of `(doc)` instead of `(docs)`.

nfagerlund:
@MikaelSmith Why in the world would it do that? That's not the convention in this repo. (doc) has about 6 commits in the last year, and (docs) has more like 30-something.

MikaelSmith:
Misunderstanding by @kylog and I probably.

kylog:
@nfagerlund yep, misunderstanding. Would you prefer that we check for `(doc)` only or `(docs)` only or accept either? Ideally, it'd be nice to apply the same to (at least) puppet/facter/hiera.

nfagerlund:
@kylog @MikaelSmith I think I'm cool with allowing both. I doubt people will want to do historical searches for all docs commits (such that we'd want consistency for easier searching), right? As far as I understand it, the use case for that commit tag is so you can tell when looking at some arbitrary commit whether it's a docs-only change, not for any kind of data mining. 

If we do want to enforce consistency, I'd vote for (docs) only.

gguillotte:
Feedback should be incorporated.

lib/puppet/functions/map.rb
```diff
@@ -1,33 +1,68 @@
...
-# @example Using map with two arguments
+# @example Using the `map` function with an array and a one-parameter lambda
 #
-#      # Turns hash into array of values
-#      $a.map |$key,$val|{ $val }
+# ~~~ puppet
+# # For the array $data, return an array containing the values
+# $data = [1,2,3]
+# $transformed_data = $data.map |$items| { $items }
+# # $transformed_data contains [1,2,3]
```
nfagerlund: This example could be improved by actually transforming the data somehow. Like, `$data.map |$item| { $item * 10 }` (produces `[10, 20, 30]`).

lib/puppet/functions/match.rb
```diff
@@ -1,19 +1,36 @@
-# Returns the match result of matching a String or Array[String] with one of:
+# Matches a regular expression against a string and returns an array containing the match
+# and any matched capturing groups.
 #
-# * Regexp
-# * String - transformed to a Regexp
-# * Pattern type
-# * Regexp type
+# The first argument a string or array of strings. The second argument is either a regular
```
nfagerlund: Typo: should be "The first argument **is** a string or..."

nfagerlund:
Two final comments; after that, :+1: :shipit: 

@kylog, would you prefer docs commits with revisions like this squashed into a single commit?

kylog:
Yes, I'd recommend squashing these into a single commit, since it's most intelligibly presented to future readers of the commits as a single cohesive change. (If you wanted to separate the `map` changes from the `match` changes, you could squash to two commits along those lines. But I'd probably just squash to a single commit.)

Thanks for checking in, and these changes look really good!

gguillotte:
Please let me know if I didn't squash these commits correctly.

kylog:
@gguillotte no that's not squashed. Once you squash you'll have a single commit with a single commit message. Here's a how-to: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html.  It will probably seem like an odd thing at first but once you get the hang of it, it's very handy.

You might grab someone to look over your shoulder when you do it the first few times. I'd offer but I'm working remote. But most of the devs around you have probably done it a bunch so ask around in hipchat or meatspace :)

gguillotte:
Yep, that's the page I followed. I'll poke folks as I'm also remote today.

On Tue, Aug 11, 2015 at 8:17 AM, Kylo Ginsberg <notifications@github.com>
wrote:

> @gguillotte <https://github.com/gguillotte> no that's not squashed. Once
> you squash you'll have a single commit with a single commit message. Here's
> a how-to:
> http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html.
> It will probably seem like an odd thing at first but once you get the hang
> of it, it's very handy.
>
> You might grab someone to look over your shoulder when you do it the first
> few times. I'd offer but I'm working remote. But most of the devs around
> you have probably done it a bunch so ask around in hipchat or meatspace :)
>
> —
> <https://github.com/puppetlabs/puppet/pull/4123#issuecomment-129926306>.
>



-- 
*PuppetConf 2015 <http://2015.puppetconf.com/> is coming to Portland,
Oregon! Join us October 5-9.*
*Register now to take advantage of the Early Bird discount
<https://www.eventbrite.com/e/puppetconf-2015-october-5-9-tickets-13115894995?discount=EarlyBird>
*
*—**save $249!*

nfagerlund:
With this all squashed and ready, I'm :+1: on it.

PR #4124, opened 2015-08-04
(PUP-3045) kill timed out execs

Sharpie:
This patch sends a TERM signal to POSIX subprocesses spawned by Exec resources
when the `timeout` paramenter triggers an exception. Any process that does not
respond to TERM before Puppet exits will be orphaned and left for init to clean
up --- this is the default behavior for all timeout events prior to this patch.

### DIFF
--- a/lib/puppet/util/execution.rb
+++ b/lib/puppet/util/execution.rb
@@ -1,3 +1,4 @@
+require 'timeout'
 require 'puppet/file_system/uniquefile'
 
 module Puppet
@@ -179,8 +180,23 @@ module Puppet::Util::Execution
       if execution_stub = Puppet::Util::ExecutionStub.current_value
         return execution_stub.call(*exec_args)
       elsif Puppet.features.posix?
-        child_pid = execute_posix(*exec_args)
-        exit_status = Process.waitpid2(child_pid).last.exitstatus
+        child_pid = nil
+        begin
+          child_pid = execute_posix(*exec_args)
+          exit_status = Process.waitpid2(child_pid).last.exitstatus
+          child_pid = nil
+        rescue Timeout::Error => e
+          # NOTE: For Ruby 2.1+, an explicit Timeout::Error class has to be
+          # passed to Timeout.timeout in order for there to be something for
+          # this block to rescue.
+          unless child_pid.nil?
+            Process.kill(:TERM, child_pid)
+            # Spawn a thread to reap the process if it dies.
+            Thread.new { Process.waitpid(child_pid) }
+          end
+
+          raise e
+        end
       elsif Puppet.features.microsoft_windows?
         process_info = execute_windows(*exec_args)
         begin
--- a/lib/puppet/provider/exec.rb
+++ b/lib/puppet/provider/exec.rb
@@ -56,7 +56,12 @@ class Puppet::Provider::Exec < Puppet::Provider
           exec_user = resource[:user]
         end
 
-        Timeout::timeout(resource[:timeout]) do
+        # Ruby 2.1 and later interrupt execution in a way that bypasses error
+        # handling by default. Passing Timeout::Error causes an exception to be
+        # raised that can be rescued inside of the block by cleanup routines.
+        #
+        # This is backwards compatible all the way to Ruby 1.8.7.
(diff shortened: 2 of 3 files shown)


Sharpie:
Currently, this PR doesn't address process termination on Windows.

lib/puppet/util/execution.rb
```diff
@@ -179,8 +179,19 @@ def self.execute(command, options = NoOptionsSpecified)
       if execution_stub = Puppet::Util::ExecutionStub.current_value
         return execution_stub.call(*exec_args)
       elsif Puppet.features.posix?
-        child_pid = execute_posix(*exec_args)
-        exit_status = Process.waitpid2(child_pid).last.exitstatus
+        begin
+          child_pid = execute_posix(*exec_args)
+          exit_status = Process.waitpid2(child_pid).last.exitstatus
+          child_pid = nil
+        rescue Timeout::Error => e
```
Sharpie: An alternative approach would be to use an `ensure` block, but it seems a bit heavy-handed.

peterhuene: We should probably have a comment here letting anyone know that this block will only be executed if you call `timeout` a certain way on Ruby 2.1+.

peterhuene:
Should we have a spec limited to Ruby 2.1+ that causes a timeout and ensures `kill` is called?

peterhuene:
Other than above :+1:

Sharpie:
I will add a unit test that checks for Process.kill getting called when the `Timeout::Error` occurs.

lib/puppet/util/execution.rb
Sharpie: Done.

Sharpie:
Spec test updated to check for invocation of `Process.kill`. The existing spec was highly POSIX-specific, so there's really nothing for windows that checks the behavior of exec timeouts.

peterhuene:
@Sharpie Looks good to merge, although I'm a bit confused as to how the spec passes on Ruby < 2.1.  Shouldn't the rescue clause be executed only on 2.1+ (and thus `kill` gets called)?  Perhaps I'm misunderstanding how the mechanism works in Ruby 2.1.

Sharpie:
The spec passes on multiple Ruby versions because passing an error class to `Timeout.timeout` is supported on all versions and results in consistent behavior across all versions.

The intended outcome is that `Process.kill` gets invoked when the timeout expires regardless of which Ruby version is in use.

peterhuene:
Ah, I had it backwards then.  I thought 2.1 added the ability to catch the exception in the block, hence my confusion. :+1:

geoffnichols:
@Sharpie @peterhuene Will we also need to target this fix at master?

kylog:
@geoffnichols we merge up 3.x -> stable -> master, so it should follow the merge ups.

joshcooper:
This passes on OSX (ruby 2.1.5), but fails in CI on Linux (ruby 2.1.1) and Windows (all rubies):

https://jenkins.puppetlabs.com/view/Jobs%20By%20Status/view/Failed/job/Puppet-Specs-3x/581/
https://jenkins.puppetlabs.com/view/Jobs%20By%20Status/view/Failed/job/Puppet-Specs-Windows-3x/858/

Sharpie:
This particular spec shouldn't run on Windows , I thought the `:as_platform => :posix` bit might do that. Looking into the Ruby 2.1.1 failure.

Sharpie:
Created PR #4159 to restore Windows compatibility to the spec test.

I was unable to re-produce the Ruby 2.1.1 failure. Tried on OS X with 2.1.1 running under rbenv and on vmpooler Debian 7.8 with 2.1.1 running under rvm. In both cases, `spec/unit/type/exec_spec.rb` passed without failure.

peterhuene:
Because the timeouts are so low in the spec, perhaps it's timing out in some instances before it enters the `begin` for the `rescue`?  That would cause it to sometimes not kill.  Oh the joys of asynchronous exceptions.

Sharpie:
That's a good thought --- the timeout is only a hundredth of a second. If that expires before `ruby -e 'sleep 0.02'` fully finishes spawning a subprocess, then no kill will be sent.

I'll bump the timeout value over in the other PR to a tenth of a second.

PR #4125, opened 2015-08-04
(PUP-4936) Add files needed for smf puppet service

haus:
There were previously no smf files for puppet that weren't geared toward
opencsw. This commit adds the smf files from the PE service, and updates
them for current pathing.

### DIFF
--- a/ext/solaris/smf/puppet.xml
+++ b/ext/solaris/smf/puppet.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0"?>
+<!DOCTYPE service_bundle SYSTEM "/usr/share/lib/xml/dtd/service_bundle.dtd.1">
+<!-- Original puppet manifest: Luke Kanies - puppetlabs.com -->
+<service_bundle type="manifest" name="puppet">
+
+  <service name="network/puppet" type="service" version="1">
+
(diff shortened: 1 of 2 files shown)


ext/solaris/smf/puppet.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0"?>
```
stahnma: Ow my :eyes:

ext/solaris/smf/puppet.xml
```diff
@@ -0,0 +1,44 @@
...
+<!DOCTYPE service_bundle SYSTEM "/usr/share/lib/xml/dtd/service_bundle.dtd.1">
+<!-- Original puppet manifest: Luke Kanies - puppetlabs.com -->
```
stahnma: Is this the actual same one from Lak? I feel like other should get to put their names in here if he does.

stahnma:
:+1:

ext/solaris/smf/puppet.xml
haus: Hah. This is from PE, so it has almost certainly seen some edits.

geoffnichols:
:+1:

PR #4126, opened 2015-08-04
(maint) Allow (docs) tag in commit messages

MikaelSmith:
Historically we've used (docs) more than (doc). Allow both (docs) and
(doc) in commit messages.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -83,12 +83,12 @@ task(:commits) do
   %x{git log --no-merges --pretty=%s master..$HEAD}.each_line do |commit_summary|
     # This regex tests for the currently supported commit summary tokens: maint, doc, packaging, or pup-<number>.
     # The exception tries to explain it in more full.
-    if /^\((maint|doc|packaging|pup-\d+)\)/i.match(commit_summary).nil?
+    if /^\((maint|doc|docs|packaging|pup-\d+)\)/i.match(commit_summary).nil?
       raise "\n\n\n\tThis commit summary didn't match CONTRIBUTING.md guidelines:\n" \
         "\n\t\t#{commit_summary}\n" \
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Fix to allow #4123 to be legal.

MikaelSmith:
@kylog Do release scripts allow both? And based on @nfagerlund's feedback it sounds like the help message should recommend (docs).

Rakefile
```diff
@@ -83,12 +83,12 @@ task(:commits) do
...
     # This regex tests for the currently supported commit summary tokens: maint, doc, packaging, or pup-<number>.
     # The exception tries to explain it in more full.
-    if /^\((maint|doc|packaging|pup-\d+)\)/i.match(commit_summary).nil?
+    if /^\((maint|doc|docs|packaging|pup-\d+)\)/i.match(commit_summary).nil?
       raise "\n\n\n\tThis commit summary didn't match CONTRIBUTING.md guidelines:\n" \
         "\n\t\t#{commit_summary}\n" \
         "\tThe commit summary (i.e. the first line of the commit message) should start with one of:\n"  \
         "\t\t(pup-<digits>) # this is most common and should be a ticket at tickets.puppetlabs.com\n" \
-        "\t\t(doc)\n" \
+        "\t\t(docs)\n" \
```
ferventcoder: So this also changes the recommendation from `(doc)` to `(docs)`. Is that on purpose?

ferventcoder: And I just read your statement above about @nfagerlund 's recommendation. :)

kylog:
The ticketmatch script does allow both: https://github.com/puppetlabs/ticketmatch/blob/master/ticketmatch.rb#L70

PR #4127, opened 2015-08-05
(PUP-4997) Fix pip provider for EL6

EslamElHusseiny:
The issue is that the latest and only available python-pip package via EPEL for EL6 "python-pip-7.1.0-1.el6.noarch" doesn't have /usr/bin/pip-python anymore which is defined to be the binary for pip provider in puppet package resource check https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/pip.rb#L42-L48
So I've added a backward compatibility, so pip provider will work either the binary is pip or pip-python

### DIFF
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -29,11 +29,8 @@ Puppet::Type.type(:package).provide :pip,
   # that's managed by `pip` or an empty array if `pip` is not available.
   def self.instances
     packages = []
-    if cmd.class == Array
-      pip_cmd = ( which(cmd.first) or which(cmd.last) ) or return []
-    else
-      pip_cmd = which(cmd) or return []
-    end
+    pip_cmd = cmd.map { |c| which(c) }.find { |c| c != nil }
+    return [] unless pip_cmd
     execpipe "#{pip_cmd} freeze" do |process|
       process.collect do |line|
         next unless options = parse(line)
@@ -44,7 +41,7 @@ Puppet::Type.type(:package).provide :pip,
   end
 
   def self.cmd
-    ["pip","pip-python"]
+    ["pip", "pip-python"]
   end
 
   # Return structured information about a particular package or `nil` if
@@ -113,23 +110,11 @@ Puppet::Type.type(:package).provide :pip,
   def lazy_pip(*args)
     pip *args
   rescue NoMethodError => e
-    if self.class.cmd.class == Array
-      if pathname = which(self.class.cmd.first)
-        self.class.commands :pip => pathname
-        pip *args
-      elsif pathname = which(self.class.cmd.last)
-        self.class.commands :pip => pathname
-        pip *args
-      else
-        raise e, "Could not locate the #{self.class.cmd.join(' and ')} commands.", e.backtrace
-      end
+    if pathname = self.class.cmd.map { |c| which(c) }.find { |c| c != nil }
+      self.class.commands :pip => pathname
+      pip *args
     else
-      if pathname = which(self.class.cmd)
-        self.class.commands :pip => pathname
-        pip *args
-      else
-        raise e, "Could not locate the #{self.class.cmd} command.", e.backtrace
-      end
+      raise e, "Could not locate command #{self.class.cmd.join(' and ')}.", e.backtrace
     end
   end
 
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -29,7 +29,11 @@ Puppet::Type.type(:package).provide :pip,
   # that's managed by `pip` or an empty array if `pip` is not available.
   def self.instances
     packages = []
-    pip_cmd = which(cmd) or return []
+    if cmd.class == Array
+      pip_cmd = ( which(cmd.first) or which(cmd.last) ) or return []
+    else
+      pip_cmd = which(cmd) or return []
+    end
     execpipe "#{pip_cmd} freeze" do |process|
       process.collect do |line|
         next unless options = parse(line)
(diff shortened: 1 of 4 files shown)


lib/puppet/provider/package/pip.rb
```diff
@@ -40,11 +40,7 @@ def self.instances
   end
 
   def self.cmd
-    if Facter.value(:osfamily) == "RedHat" and Facter.value(:operatingsystemmajrelease).to_i < 7
-      "pip-python"
-    else
-      "pip"
-    end
+    "pip"
```
kylog: I am concerned that this will break older RHEL6 and RHEL5 versions. (I haven't actually looked at those, but I'm going based on the history of this code.)  See #3470 for some additional context.

Perhaps we need to either:
a) for RHEL6 make this conditional on minor version
or
b) try `pip` and fall back to `pip-python`

EslamElHusseiny: Well actually the issue is that the latest python-pip package in EPEL repo and the only available one doesn't have /usr/bin/pip-python symbolic link anymore.

However you are right, I've check some old python-pip packages
one of them has only /usr/bin/pip-python
and another one has both /usr/bin/pip and /usr/bin/pip-python
and the latest one has only /usr/bin/pip

Anyway I've addressed this issue and added a backward compatibility for pip-python binary aswell

kylog:
This seems like a reasonable approach to me (try `pip` first then `pip-python`), but I'd like another set of eyes on it. 

@nibalizer, I know you work with pip a fair bit on different platforms. Any thoughts on this?

lib/puppet/provider/package/pip.rb
```diff
@@ -29,7 +29,11 @@ def self.parse(line)
   # that's managed by `pip` or an empty array if `pip` is not available.
   def self.instances
     packages = []
-    pip_cmd = which(cmd) or return []
+    if cmd.class == Array
```
trlinkin: Why is this done? Am I missing something here? Is this in case a new provider inherits this one?

EslamElHusseiny: To cover both cases having either /usr/bin/pip or /usr/bin/pip-python as the binaries for pip provider.
Because currently the latest and the only available python-pip package in EPEL repos for EL6 provides the "/usr/bin/pip" binary, whereas the older package was providing "/usr/bin/pip-python" binary.
make sense ?

kylog: I think what @trlinkin may be asking is when will `cmd` ever *not* be an array? I.e. since `cmd` is defined as an array on line 47 below.

lib/puppet/provider/package/pip.rb
```diff
@@ -29,7 +29,11 @@ def self.parse(line)
...
+      pip_cmd = ( which(cmd.first) or which(cmd.last) ) or return []
```
peterhuene: This makes an assumption that there are always two elements in the array.  A cleaner implementation would be:

```
...
pip_cmd = cmd.find { |c| which(c) }
return [] unless pip_cmd
...
``

We prefer not to use the `and` and `or` keywords as they can introduce unintentional behavior when used in logical expressions where developers don't expect their precedence compared to `&&` and `||`.  Additionally, using them for control flow, such as returning early from a function, makes it a little harder to spot what's happening.

File: lib/puppet/provider/package/pip.rb [hunk 4]
```diff
@@ -29,7 +29,11 @@ def self.parse(line)
...
+    else
+      pip_cmd = which(cmd) or return []
```
peterhuene: Same comment as above with respect to `or`.  If `cmd` is always an array, there's no need for this else anyway.

lib/puppet/provider/package/pip.rb
```diff
@@ -113,11 +113,23 @@ def update
   def lazy_pip(*args)
     pip *args
   rescue NoMethodError => e
-    if pathname = which(self.class.cmd)
-      self.class.commands :pip => pathname
-      pip *args
+    if self.class.cmd.class == Array
+      if pathname = which(self.class.cmd.first)
```
peterhuene: Same assumption as above about the number of elements in the array.  If some distro comes up with yet another way to name "pip", someone might append the new name to the array and suddenly skip `pip-python` checks, for example.

joshcooper:
Looks like @peterhuene has some outstanding questions on this one

HAIL9000:
Hey @EslamElHusseiny, thanks for the contribution! Are you still interested in working on this? If you get a chance, please let us know either way so we know how to move forward.

Thanks!

EslamElHusseiny:
@HAIL9000 I'm sorry for being too late, I've addressed comments I got, still need to fix the corresponding unit tests provided that the implementation is accepted.

@peterhuene you are right, I've used your simpler implementation provided that both pip and pip3 are now arrays whatever how many elements in the array.

@kylog I've defined cmd in both pip and pip3 as an array for simpler, unified handling.

kylog:
@EslamElHusseiny this PR now includes a number of files under `puppetlabs/puppet/cache`, probably from your debug set up. Can you remove those? It also has a merge commit; can you squash that out at the same time?

Thanks!

EslamElHusseiny:
@kylog sorry for that, it's fixed now

HAIL9000:
Ping @kylog for further review.

kylog:
Thanks @EslamElHusseiny! This should be released in puppet 4.4.

chad-upton:
@kylog When is 4.4 roughly scheduled for release? Just curious as this bug is one that has been bugging me for awhile.

Thanks!

cristim:
Is this by any chance also going to be backported to the 3.x branch anytime soon?

cristim:
@chad-upton it turns out 4.4.0 was released in mid-March.

PR #4129, opened 2015-08-05
(docs) Fix misleading statement about backslashes in install_options …

nfagerlund:
…parameter

Backslashes can/should be used in several places on Windows.


nfagerlund:
When this is merged, please resolve DOC-1709.

PR #4130, opened 2015-08-05
(PUP-4824) Fixup acceptance group / user tests for OSX

whopper:
Previously, several user and group acceptance tests dependeded
on `getent` to create and query users and groups on the system.
This did not work in OSX, and caused several test failures.

We now have cross-platform beaker methods specifically meant for
user and group management which we can use to replace the bare
`getent` calls.

### DIFF
--- a/acceptance/tests/resource/user/should_create.rb
+++ b/acceptance/tests/resource/user/should_create.rb
@@ -1,7 +1,5 @@
 test_name "should create a user"
 
-confine :except, :platform => /osx/ # See PUP-4824
-
 name = "pl#{rand(999999).to_i}"
 
 agents.each do |agent|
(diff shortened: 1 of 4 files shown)


acceptance/tests/resource/user/should_create.rb
```diff
@@ -16,7 +14,7 @@
   agent.user_get(name)
 
   case agent['platform']
-  when /sles/, /solaris/, /windows/
+  when /sles/, /solaris/, /windows/, /osx/
```
whopper: While debugging this test I realized that something like: ```user { 'testuser': ensure=>present }``` creates the user ```testuser``` in OSX, but not the group. I'm not sure if this is known or expected behavior, but for now I added OSX to the list of platforms we don't expect to create the group.

PR #4135, opened 2015-08-06
(PUP-4822) Fixup acceptance in OSX

whopper:
The first of these commits re-enables PMT-related tests for OSX.

We now bundle the GeoTrust Global CA with our vendored openssl,
so we can re-enable tests which require a connection to the
forge on OSX. See PUP-4688 for additional context.

The second commit cherry-picks PUP-4824 from master, which further
fixes OSX acceptance by removing bare calls to `getent`.

### DIFF
--- a/acceptance/tests/modules/install/basic_install.rb
+++ b/acceptance/tests/modules/install/basic_install.rb
@@ -4,8 +4,10 @@ extend Puppet::Acceptance::ModuleUtils
 
 hosts.each do |host|
   case host['platform']
-  when /solaris/, /osx/
+  when /solaris/
     # see PUP-4822, PUP-3450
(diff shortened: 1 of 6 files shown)


whopper:
I verified that the PMT tests are working as of the latest puppet-agent build (f0a37c85e5d7) in OSX 10-10:

       Total Suite Time: 16.56 seconds
      Average Test Time: 8.28 seconds
              Attempted: 2
                 Passed: 2
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 2

Also verified that ```/opt/puppetlabs/puppet/ssl/cert.pem``` exists on the test host and looks as we expect it to.

PR #4136, opened 2015-08-06
(maint) Fix bug in stat function

jpinsonault:
This is coming from a CI failure for `puppet-server`, which uses `puppet`
4.2.1, see SERVER-832 for details.

The stat function references an `agent` variable, which will coincidentally
be set correctly to the same value as `host` if the beaker config only has
one host with an agent role. But if there is more than one agent, the
'agent' variable will be a list of hosts, and the stat function will throw an
exception.

The bug was introduced in 8d2a872

### DIFF
--- a/acceptance/lib/puppet/acceptance/temp_file_utils.rb
+++ b/acceptance/lib/puppet/acceptance/temp_file_utils.rb
@@ -140,7 +140,7 @@ module Puppet
       # value containing only the file mode, excluding the type, e.g
       # S_IFDIR 0040000
       def stat(host, path)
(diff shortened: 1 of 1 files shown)


kylog:
Thanks @jpinsonault !

joshcooper:
Thanks @jpinsonault This is why I dislike beaker automagically creating variables in scope for each test case, e.g. stdout, agent, default, etc. Things work when they shouldn't.

jpinsonault:
@joshcooper No problem, yeah and for a while I didn't even understand how it could have ever worked, because I couldn't track down where the variable came from

PR #4140, opened 2015-08-07
(doc) 'JIRA tracker' link fix in the README

ryayon:
Currently, the 'JIRA tracker' link points to http://tickets.puppetlabs.com/
which redirects itself to https://tickets.puppetlabs.com/ with a 301 redirection.
I just changed this to point directly to the correct https page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ See [LICENSE](LICENSE) file.
 Support
 -------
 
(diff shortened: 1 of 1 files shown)


kylog:
Good catch! Thanks @ryayon!

ryayon:
Nice to hear !

Thanks

PR #4141, opened 2015-08-07
(PUP-5009) Fixup puppet-agent paths and pmt tests run on git install

johnduarte:
This commit sets the ensure_puppet-agent_paths.rb test to skip
unless the beaker option 'type' is set to aio. This ensures that
the test does not erroneously fail in a non puppet-agent
environment.


johnduarte:
Closing to resubmit against stable.

PR #4142, opened 2015-08-07
(PUP-5010) Skip upstart test on vivid

johnduarte:
This commit sets the handle upstart test to skip the 'vivid'
version of ubuntu since it uses systemd rather than upstart to
manage services.

### DIFF
--- a/acceptance/tests/resource/service/ticket_14297_handle_upstart.rb
+++ b/acceptance/tests/resource/service/ticket_14297_handle_upstart.rb
@@ -2,6 +2,8 @@ test_name 'Upstart Testing'
 
 # only run these on ubuntu vms
 confine :to, :platform => 'ubuntu'
+# vivid and above use systemd rather than upstart
+confine :except, :platform => /ubuntu-[v-z]/
 
 # pick any ubuntu agent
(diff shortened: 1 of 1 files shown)


acceptance/tests/resource/service/ticket_14297_handle_upstart.rb
```diff
@@ -2,6 +2,8 @@
 
 # only run these on ubuntu vms
 confine :to, :platform => 'ubuntu'
+# vidid and above use systemd rather than upstart
```
kylog: Typo: `vidid` should be `vivid`.

acceptance/tests/resource/service/ticket_14297_handle_upstart.rb
```diff
@@ -2,6 +2,8 @@
...
+confine :except, :platform => 'ubuntu-vivid'
```
kylog: Should we make this a regex so that we don't need to bump this for each ensuing ubuntu release? As the comment says, this is vivid and up.

And ubuntu releases are alphabetical so this might be sth like `/ubuntu-[v-z]/`

PR #4143, opened 2015-08-07
(PUP-5009) Fixup puppet-agent paths and pmt tests run on git install

johnduarte:


acceptance/tests/resource/package/common_package_name_in_different_providers.rb
```diff
@@ -14,6 +14,17 @@
   end
 end
 
+step "Verify gem and ruby-devel are installed on agents" do
+  agents.each do |agent|
+    unless check_for_package agent, 'gem'
```
MikaelSmith: Shouldn't you also check_for_package rubygems?

kylog:
Also, perhaps a nit, but should the first two commits be `(maint)`? AFAICT they're about adding support for git testing, and are unrelated to PUP-5009 (fedora 22 support).

johnduarte:
@kylog the common_package_name_in_different_providers test is already confined to to  el, centos, fedora. See the reference below for detail. I have confirmed that the `rubygems` and `ruby-devel` packages are available on a variety of these platforms.

Your comment about `(maint)` is well taken. Should these be submitted as a separate PR?

Ref:
https://github.com/johnduarte/puppet/blob/pup_5009_test_fixup/acceptance/tests/resource/package/common_package_name_in_different_providers.rb#L3

kylog:
Ok, thanks for confirming. I'm fine mixing maint and ticket commits in a single PR - I just find it handy for the summaries to be indicative. So I'd just amend the commits.

Other than that: :+1:

johnduarte:
Commits amended. Thanks @kylog

joshcooper:
All of the PMT tests fail when executing `git` style acceptance, because beaker tells the test to look in the old module path location, but the code uses the new AIO paths. For example, the code writes a module to the new location, e.g. `/etc/puppetlabs/code/modules`, but the test looks in the old location and can't find the newly written module.

@Iristyle already fixed this in https://github.com/puppetlabs/puppet/pull/4088 and merged to master in https://github.com/puppetlabs/puppet/commit/71bc9d0e41578e03db34609e108244d4969a66f4. I think we should drop commit https://github.com/johnduarte/puppet/commit/0e70e9206f5db72efe5440c74cace7e286c133e4, and backport @Iristyle's fix to stable.

We do need to keep the change to puppet-agent_paths, as @MikaelSmith pointed out the test needs to be confined to only run in AIO: https://tickets.puppetlabs.com/browse/PUP-4050?focusedCommentId=195781&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-195781

Also @johnduarte is https://github.com/johnduarte/puppet/commit/07d2bdf87c2c73a6a4dad5e164574620d8422cce only necessary for fedora 22 because our VMs don't have `gem` installed? If so, could you update the commit message so it's clear why we're installing those packages? Also do you know if those package names are valid for all platforms that the test runs on el/centos/fedora, or should we only ensure those packages on fedora 22?

johnduarte:
@joshcooper the `rubygems` and `ruby-devel` packages are required by all the el/centos/fedora platforms scoped by the test. This is added for external users who want to test this as well as for our VMs. If memory serves correctly, this was surfaced by the packages not being included in our fedora 22 VM templates.

These package names are common through the el/centos/fedora ecosystem, but I have not done an exhaustive search across all variants.

joshcooper:
@johnduarte I backported @Iristyle commits for fixing module related acceptance failures with git and verified on redhat-7, fedora-22, win2012r2. Could  you rebase your PR on upstream/stable and drop commit https://github.com/johnduarte/puppet/commit/0e70e9206f5db72efe5440c74cace7e286c133e4?

Adding the packages makes sense, I was just wondering if the package names are correct on platforms other than fedora 22. For example, will this break EL4 if we were to run `git` acceptance?

acceptance/tests/resource/package/common_package_name_in_different_providers.rb
```diff
@@ -14,6 +14,17 @@
   end
 end
 
+step "Verify gem and ruby-devel are installed on agents" do
```
joshcooper: I'm thinking this step should be skipped in `aio` since its part of the package, e.g. "gem"

johnduarte:
@joshcooper yes, your point about el4 is well taken. These are the correct package names for el4. However, the beaker method `install_package` does not support el4 due to the lack of a repo manager, so this step will fail on el4.

https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix/pkg.rb#L67

acceptance/tests/resource/package/common_package_name_in_different_providers.rb
```diff
@@ -15,6 +15,21 @@
   end
 end
 
+step "Verify gem and ruby-devel on fedora-22 and above if not aio" do
+  if type != 'aio' then
```
joshcooper: I get a failure:

```
$ be rake ci:test:aio TESTS=tests/ensure_puppet-agent_paths.rb,tests/resource/package/common_package_name_in_different_providers.rb SHA=1.2.5 CONFIG=config/nodes/redhat-7-x86_64.yaml
...
undefined local variable or method `type' for #<Beaker::TestCase:0x007facd91d5000>
```

I see a different test does `if @options[:type] != 'aio'`

acceptance/tests/ensure_puppet-agent_paths.rb
```diff
@@ -120,10 +120,15 @@ def config_options(agent)
 
 
 public_binaries = {
-  :posix => ['puppet', 'facter', 'hiera', 'mco'],
-  :win   => ['puppet.bat', 'facter.bat', 'hiera.bat', 'mco.bat']
+  :posix => ['puppet', 'facter', 'hiera'],
+  :win   => ['puppet.bat', 'facter.bat', 'hiera.bat']
 }
 
+if type != 'git' then
```
joshcooper: same comment here

johnduarte:
Sorry, @joshcooper . Patched.

When tested on fedora-22 `tests/resource/package/common_package_name_in_different_providers.rb` fails (does not create `Package[guid]`).

whopper:
@johnduarte that failure is because DNF requires a cache reset before the RPM becomes available. I've got a PR up (https://github.com/puppetlabs/puppet/pull/4321) that fixes that test.

MikaelSmith:
:+1:

joshcooper:
Verified using

```
$ be rake ci:test:aio TESTS=tests/ensure_puppet-agent_paths.rb,tests/resource/package/common_package_name_in_different_providers.rb SHA=047225665a95cba35e0be06c4ba11927ae7e8f41 SUITE_VERSION=0472256 CONFIG=config/nodes/fedora-22-x86_64.yaml
```

There is one failure due to the DNF requiring a cache reset mentioned above and will be fixed in a later PR

PR #4144, opened 2015-08-08
(maint) Remove redundant travis jobs

kylog:
When .travis.yml was updated for the latest 2.2 z release,
the exclusions matrix wasn't updated, which meant the commits
and rubocop jobs were being run on multiple rubies (no point
in that, never the intent).

So this commit just fixes up the excludes to prune those out
as intended.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,13 +18,13 @@ env:
 
 matrix:
   exclude:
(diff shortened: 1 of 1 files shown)


ferventcoder:
LGTM

PR #4145, opened 2015-08-10
(PUP-5014) Base class for JSON indirection terminus calls

tdevelioglu:
log_exception() wrong.

In Puppet::Indirector::JSON, Puppet.log_exception() is called without
the required exception as the first argument value, instead the message
is passed.

This breaks further along in format_exception() as it attempts to get
the exception message attribute but finds a string:

### DIFF
--- a/lib/puppet/indirector/json.rb
+++ b/lib/puppet/indirector/json.rb
@@ -17,7 +17,7 @@ class Puppet::Indirector::JSON < Puppet::Indirector::Terminus
 
     Puppet::Util.replace_file(filename, 0660) {|f| f.print to_json(request.instance) }
   rescue TypeError => detail
-    Puppet.log_exception "Could not save #{self.name} #{request.key}: #{detail}"
+    Puppet.log_exception(detail, "Could not save #{self.name} #{request.key}: #{detail}")
   end
 
   def destroy(request)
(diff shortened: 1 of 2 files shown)


kylog:
@tdevelioglu good catch! 

We did a quick audit and it looks like https://github.com/puppetlabs/puppet/blob/fb15a86cc822929fa346ea4ae27da3a6ac0e68bb/lib/puppet/indirector/msgpack.rb#L26 suffers from the same problem. (IIRC the msgpack support used the json support as a model, so this omission may have been faithfully reproduced.)

Do you mind extending your patch to include a fix for that one too?

HAIL9000:
Thanks the controbution @tdevelioglu! In addition to Kylo's comment, would you mind fixing up your commit message a little bit? 

It would be a little clearer if you described what the fix is (passing the exception into log_exception) and why the fix was necessary (this broke formatting further down the line and caused puppet to raise an error).

tdevelioglu:
@HAIL9000 
Fix extended to the msgpack base class and updated the commit message.

kylog:
Thanks again, @tdevelioglu !

kylog:
@HAIL9000 in our triage notes, you were going to grep jira for tickets that might map to this. Did you get a chance to do that?

HAIL9000:
@kylog Not yet! I'll make some time for that this week

PR #4146, opened 2015-08-10
(PUP-5016) Set service's debian provider to be the default for all de…

mcanevet:
…bian version.

Replaces : #4092

### DIFF
--- a/lib/puppet/provider/service/debian.rb
+++ b/lib/puppet/provider/service/debian.rb
@@ -16,8 +16,7 @@ Puppet::Type.type(:service).provide :debian, :parent => :init do
   # is resolved.
   commands :invoke_rc => "/usr/sbin/invoke-rc.d"
 
-  defaultfor :operatingsystem => :cumuluslinux
-  defaultfor :operatingsystem => :debian, :operatingsystemmajrelease => ['5','6','7']
+  defaultfor :operatingsystem => [:debian, :cumuluslinux]
 
   # Remove the symlinks
   def disable
--- a/spec/unit/provider/service/debian_spec.rb
+++ b/spec/unit/provider/service/debian_spec.rb
@@ -35,10 +35,7 @@ describe provider_class do
   operatingsystem = [ 'Debian', 'CumulusLinux' ]
   operatingsystem.each do |os|
     it "should be the default provider on #{os}" do
-      Facter.expects(:value).with(:operatingsystem).at_least_once.returns(os)
-      if os == 'Debian'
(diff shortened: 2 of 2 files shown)


melissa:
@mcanevet awesome! Thanks for filling this! 

What happens when you test it? Does it do what you were suggesting in PR#4092? Does it effectively allow the user to manage both systemd and sysv services? What about duplicate resources appearing in the list?

peterhuene:
Hi @mcanevet.  Thanks for the contribution!  Have you had a chance to review @melissa's comment above?

whopper:
@mcanevet @melissa I've been playing around with this idea, and it doesn't seem to be working correctly. The idea is to fall back to the `debian` provider when `systemd` can't find a unit file to work with, but it seems that even doing this results in the `init` provider being used.

For reference, before commit 7fe6164 (when everything was working correctly), we'd see:

```
root@ik7f93ryckp5kbe:~/puppet# bundle exec puppet resource service apache

(debugger) :provider=>#<Puppet::Type::Service::ParameterProvider:0x000000022b6130 @resource=#<Puppet::Type::Service:0x000000022b9e98 ...>, @value=:debian>

Debug: Service apache found in both debian and init; skipping the init version

service { 'apache':
  ensure => 'running',
  enable => 'true',
}
```

With the breaking change (and without your change), we get:

```
root@ik7f93ryckp5kbe:~/puppet# bundle exec puppet resource service apache

(debugger) Service[apache2](provider=init)

Debug: Service apache2 found in both init and debian; skipping the debian version

service { 'apache2':
  ensure => 'running',
}
```


And finally, with your change, I'm seeing the same thing:

```
Service[apache2](provider=init)

Debug: Service apache2 found in both init and debian; skipping the debian version

service { 'apache2':
  ensure => 'running',
}
```

It's still not completely clear to me why the debian provider isn't being used, but we don't seem to be falling back to the second default provider.

whopper:
Update:

I've done some digging on this, and found the issue that's currently preventing it from working.
In Puppet::type.defaultprovider, when we come across two default providers we pick the one with the highest specificity score (from Puppet::Provider.specificity). In this case, systemd wins out over the debian provider, which means the debian provider is tossed back into the 'suitable' list, behind the init provider. Thus, init gets chosen before debian.
It seems we need a way to deal with this dual-default provider issue, especially for Debian where systemd may or may not actually be suitable for certain services.

whopper:
@mcanevet we're going to be taking a closer look at a more comprehensive solution to systemd's brokenness on Debian soon, so closing this PR. Thanks for taking a look!

PR #4149, opened 2015-08-11
(docs) Revise `each` function docs.

gguillotte:


lib/puppet/functions/each.rb
```diff
@@ -1,38 +1,76 @@
-#  Applies a parameterized block to each element in a sequence of selected entries from the first
-#  argument and returns the first argument.
+# Runs a [lambda](http://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html)
+# repeatedly using each value in a data structure, then returning the values unchanged.
```
nfagerlund: "then returns"

lib/puppet/functions/each.rb
```diff
@@ -1,38 +1,76 @@
...
-# When the first argument is an Array (or of enumerable type other than Hash), the parameterized block
-# should define one or two block parameters.
-# For each application of the block, the next element from the array is selected, and it is passed to
-# the block if the block has one parameter. If the block has two parameters, the first is the elements
-# index, and the second the value. The index starts from 0.
+# @example Using the `each` function
 #
-#       $a.each |$index, $value| { ... }
-#       each($a) |$index, $value| { ... }
+# `$each_data = $data.each |$parameter| { <PUPPET CODE BLOCK> }`
```
nfagerlund: Okay, here's an example of how the iteration functions are all slightly different. With `each`, you'll _never_ want to capture the return value, because you already have it (in the same `$data` variable you're iterating over). It works mostly the same way `map` does, except that you're generally calling it solely for the side effects of the code in the lambda. 

Short version is that you're almost always writing resource declarations in an `each` lambda, and the point of calling `each` is to add those resources to the catalog.

lib/puppet/functions/each.rb
```diff
@@ -1,38 +1,76 @@
...
+# @example Using the `each` function with an array and a one-parameter lambda
+#
+# ~~~ puppet
+# # For the array $data, run a lambda outputting each value multiplied by 10 using the
+# # notice function, then return the array
+# $data = [1, 2, 3]
+# $each_data = $data.each |$item| { notice ($item * 10) }
+# # The notice function runs three times, outputting 10, then 20, then 30
+# # $each_data contains [1, 2, 3]
+# ~~~
```
nfagerlund: For that reason, I'd recommend using a slightly more in-depth example. The one I used in the lambdas and iteration reference pages is a good, simple candidate: take an array of strings, and construct a bunch of symlink resources from it.

lib/puppet/functions/each.rb
```diff
@@ -1,38 +1,76 @@
...
+#
+# When the first argument is a hash, Puppet passes each key and value pair to the lambda
+# as an array in the form `[key, value]` and returns a hash containing the results.
+#
+# @example Using the `filter` function with a hash and a one-parameter lambda
```
nfagerlund: `each`, not `filter`!

lib/puppet/functions/each.rb
```diff
@@ -1,38 +1,76 @@
...
+#
+# ~~~ puppet
+# # For the hash $data, run a lambda using each item as a key-value array and outputting
+# # the keys using the notice function, then return the hash
+# $data = {"a" => 0, "b" => 1, "c" => 2}
+# $each_data = $data.each |$items| { notice ($items[0]) }
+# # The notice function runs three times, outputting a, then b, then c
+# # $each_data contains {a => 0, b => 1, c => 2}
+# ~~~
```
nfagerlund: Here's an interesting exercise for you: can you think of an example that uses a hash to construct resources? Maybe in the same family of tasks (using data to create symlinks)?

lib/puppet/functions/each.rb
```diff
@@ -1,38 +1,76 @@
...
+# @example Using the `each` function with a hash and a two-parameter lambda
+#
+# ~~~ puppet
+# # For the hash $data, run a lambda outputting each key and its corresponding value, then
+# # return the hash
+# $data = {"a" => 0, "b" => 1, "c" => 2}
+# $each_data = $data.each |$key, $value| { notice ($key, $value) }
+# # The notice function runs three times, outputting a 0, then b 1, then c 2
+# # $each_data contains {a => 0, b => 1, c => 2}
+# ~~~
```
nfagerlund: The 2-parameter examples should probably be the same task as the 1-parameter ones, since that gives the best parallax.

nfagerlund:
I'm :+1: on this!

nfagerlund:
Per our conversation last week about letting the docs team merge docs commits... I'm gonna merge this.

PR #4132, opened 2015-08-05
(SERVER-832) Fix bug in stat function

jpinsonault:
This is coming from a CI failure for `puppet-server`, which uses `puppet` 4.2.1, see [SERVER-832](https://tickets.puppetlabs.com/browse/SERVER-832) for details. 

It seems that the `stat` function references the `agent` variable when it should be using the `host` parameter that gets passed in. I'm actually a bit confused as to how it functions at all right now, since I can't reproduce the test failure locally.

PR #4150, opened 2015-08-12
(PUP-4824) Fix typo on resource acceptance tests

whopper:
No idea why this didn't fail in our pipeline or locally, but this should fix the acceptance failures on the puppet-server board.

image


kylog:
Sigh. IIUC it worked in puppet acceptance but failed in puppetserver acceptance because the configs we use on puppet acceptance have a single agent role so `agent` is a hash, whereas if there are mutliple agent roles (as done in puppetserver acceptance) it becomes an array because why not. And then beaker exposes `agent` as a global. Woot!

See #4132 and #4136 for more of the same fun.

PR #4152, opened 2015-08-12
(PUP-5032) Change configured_environment behavior

MikaelSmith:
Puppet's communication with the node and catalog terminus included a
configured_environment option in the last release. That option sometimes
includes redundant or irrelevant information when `Puppet[:environment]`
has the default setting. That redundant information precludes it from
being used to identify when an agent has explicitly requested a
particular environment.

Change the behavior of configured_environment to only be set when the
`environment` setting is explicitly set.

### DIFF
--- a/spec/unit/settings_spec.rb
+++ b/spec/unit/settings_spec.rb
@@ -301,9 +301,15 @@ describe Puppet::Settings do
       expect(@settings.set_by_config?(:myval)).to be_truthy
     end
 
-    it "should not identify configured settings from environment" do
+    it "should not identify configured settings from environment by default" do
       Puppet.lookup(:environments).expects(:get_conf).with(Puppet[:environment].to_sym).never
       expect(@settings.set_by_config?(:manifest)).to be_falsey
+    end
+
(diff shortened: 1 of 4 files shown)


reidmv:
:+1:

spec/unit/settings_spec.rb
```diff
@@ -287,6 +287,58 @@ def metadata(setting)
...
+      @settings.instance_variable_get(:@value_sets)[:memory].expects(:lookup).with(:myval).returns('foo')
+      expect(@settings.set_by_config?(:myval)).to be_truthy
+    end
+
+    it "should identify configured settings from CLI" do
+      @settings.instance_variable_get(:@value_sets)[:cli].expects(:lookup).with(:myval).returns('foo')
+      expect(@settings.set_by_config?(:myval)).to be_truthy
+    end
+
+    it "should not identify configured settings from environment" do
```
MikaelSmith: This one's interesting, but made sense for the methods use in setting agent_specified_environment. We wouldn't expect the environment.conf to specify an environment.

In general, it would make sense to extend set_by_config? to take an environment argument as an option.

PR #4153, opened 2015-08-12
(PUP-4952) Use vendored curl in concurrency acceptance test

whopper:
Previously, the concurrent catalog request test was failing in
SLES 10 due to an old version of cURL on the system. This commit
updates the test to use our vendored cURL when it is present, or
fall back to the system cURL if it is not.

### DIFF
--- a/acceptance/tests/concurrency/ticket_2659_concurrent_catalog_requests.rb
+++ b/acceptance/tests/concurrency/ticket_2659_concurrent_catalog_requests.rb
@@ -85,7 +85,7 @@ with_puppet_running_on(master, master_opts, testdir) do
           sleep $sleep_for
           url='https://#{master}:8140/puppet/v3/catalog/#{agent_cert}?environment=production'
           echo "Curling: $url"
-          curl --tlsv1 -v -# -H 'Accept: text/pson' --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} $url
+          env PATH='#{agent[:privatebindir]}:${PATH}' curl --tlsv1 -v -# -H 'Accept: text/pson' --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} $url
           echo "$PPID Completed"
         ) > "#{agent_tmpdir}/catalog-request-$i.out" 2>&1 &
         echo "Launched $!"


whopper:
Tested locally in SLES 10, SLES 11, RedHat 7, CentOS 5, Debian 8, Oracle 6 and Ubuntu 1004. The test is currently skipped in Windows and OSX.

acceptance/tests/concurrency/ticket_2659_concurrent_catalog_requests.rb
```diff
@@ -85,7 +85,7 @@
           sleep $sleep_for
           url='https://#{master}:8140/puppet/v3/catalog/#{agent_cert}?environment=production'
           echo "Curling: $url"
-          curl --tlsv1 -v -# -H 'Accept: text/pson' --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} $url
+          env PATH='#{master[:privatebindir]}:${PATH}' curl --tlsv1 -v -# -H 'Accept: text/pson' --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} $url
```
whopper: This should actually be `agent[:privatebindir]` ... updating.

acceptance/tests/concurrency/ticket_2659_concurrent_catalog_requests.rb
```diff
@@ -85,7 +85,7 @@
           sleep $sleep_for
           url='https://#{master}:8140/puppet/v3/catalog/#{agent_cert}?environment=production'
           echo "Curling: $url"
-          curl --tlsv1 -v -# -H 'Accept: text/pson' --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} $url
+          env PATH='#{agent[:privatebindir]}:${PATH}' curl --tlsv1 -v -# -H 'Accept: text/pson' --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} $url
```
er0ck: privatebindir is a beaker default that perhaps we should not depend upon.  is there another way we can obtain this path?

joshcooper: `privatebindir` and `puppetbindir` are one of the few hardcoded beaker defaults as the location of the binaries depends on the `type` of acceptance and the unix vs windows host, e.g. `git` installs things in /usr/bin (mostly), `aio` installs things in `/opt/puppetlabs/puppet/bin`, etc.

joshcooper:
I'd normally suggest using beaker's `host['pathseparator']` instead of hardcoding `':'`, but since the test is confined to skip on windows, would require significant work to get it working, and we don't support windows master platforms, I'd say don't worry about it.

acceptance/tests/concurrency/ticket_2659_concurrent_catalog_requests.rb
whopper: Hm, so @joshcooper do you think there's a safer way, or are we stuck on `privatebindir`since it is one of the few places to find the bindir? I noticed that other tests use `privatebindir` whenever we need to find our ruby, so maybe this is the best option we have?

joshcooper: I'd go with `privatebindir`

er0ck: oh right.  thanks josh.  i guess that should stay in beaker and move to a puppet  beaker library later...  only other solution i can think of is specifying the windows and posix paths in options/hosts files and conditional on platform in all_the_tests ick.

PR #4154, opened 2015-08-13
(PUP-5033) Add capability mappings to defined types

lutter:


lib/puppet/parser/ast/pops_bridge.rb
```diff
@@ -73,6 +73,9 @@ class Program < Puppet::Parser::AST::TopLevelConstruct
     def initialize(program_model, context = {})
       @program_model = program_model
       @context = context
+      # @todo lutter 2015-03-10: we maintain a list of definitions we
+      # instantiated so that we can attach CapabilityMappings to them
+      @types = {}
```
thallgren: Judging from the code in `Puppet::Parser::E4ParserAdapter` where the `Progam` instance is created, it seems it's created once per file which would make it impossible to have consumes/produces span over several files. A limitation that I assume is undesirable. That means the actual binding cannot be done during parsing.

Perhaps an alternative could be to have the `instantiate_CapabilityMapping` return a model object that represents the blueprint (with the name of the resource added). It could be named `CapabilityBlueprintDefinition` and inherit from `Puppet::Pops::Model::Definition`. The binding of the blueprints to their respective types could then be deferred to a new method `Puppet::Parser::Compiler::evaluate_capability_blueprints` (could be called just before `evaluate_main`). Here, the `known_resource_types` is available so it should be easy to find the referenced types.

lutter: Yeah, I agree that the way it currently is, you won't be able to have cross-file capability annotations.

I am not sure what the right approach is - `PopsBridge::Program#instantiate` must return an array of `Puppet::Resource::Type`, strictly speaking objects that have a `type` method, and for which `Puppet::Resource::TypeCollection` has an `add_TYPE` method.

I can see how `instantiate_FunctionDefinition` does some dark magic to store function definitions in loaders - I am not sure where `instantiate_CapabilityMapping` would store its blueprints/annotations.

I've filed [APPMGMT-834](https://tickets.puppetlabs.com/browse/APPMGMT-834) to track the cross-file annotations issue to unblock this review; the functionality here is limited but still works as long as annotations are in the same file as the defined type.

kylog:
Any idea why it's failing Windows CI? It wasn't obvious to me from a quick glance.

spec/unit/pops/validator/validator_spec.rb
```diff
@@ -228,6 +228,27 @@ def validate(model)
     end
   end
 
+  context "capability annotations" do
+    before(:all) do
```
thallgren: Needs corresponding after(:all) that turns off capability management.

spec/unit/pops/parser/parser_rspec_helper.rb
```diff
@@ -4,6 +4,17 @@
 
 module ParserRspecHelper
   include FactoryRspecHelper
+
+  def with_app_management(flag)
+    # Horrible things have to be done to test this as the Lexer sets up the
+    # KEYWORD table as a frozen constant, and this list depends on the
+    # Puppet[:app_management] setting. These 'before all' and 'after all'
+    # clauses forces Ruby to reload the Lexer
+    Puppet[:app_management] = true
```
thallgren: Change to `Puppet[:app_management] = flag` instead of always setting it to `true`.

spec/unit/pops/parser/parse_capabilities_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+require 'spec_helper'
+require 'puppet/pops'
+
+# relative to this spec file (./) does not work as this file is loaded by rspec
+require File.join(File.dirname(__FILE__), '/parser_rspec_helper')
+
+describe "egrammar parsing of capability mappings" do
+  include ParserRspecHelper
+
+  before(:all) do
```
thallgren: Needs corresponding after(:all) that turns off capability management.

spec/unit/pops/parser/parse_application_spec.rb
```diff
@@ -0,0 +1,38 @@
+#! /usr/bin/env ruby
+require 'spec_helper'
+require 'puppet/pops'
+
+# relative to this spec file (./) does not work as this file is loaded by rspec
+require File.join(File.dirname(__FILE__), '/parser_rspec_helper')
+
+describe "egrammar parsing of 'application'" do
+  before(:all) do
```
thallgren: Needs corresponding after(:all) that turns off capability management.

thallgren:
The reason AppVeyor failed the pops/validator tests is that they introduces an order dependent problem with the `with_app_management(true)` call in a `before(:all)` declaration. It needs a corresponding `after(:all)` that turns app management off again.

This problem is not specific to Windows. I can reproduce it both on windows and on Fedora. Can be tested by running:

bundle exec rspec spec/unit/pops/validator/validator_spec.rb # Works OK
bundle exec rspec spec/unit/pops # Fails!

thallgren:
This PR should contain a test that verifies that a new `Resource::Type` declared with consumes/produces end up with blueprints in their respective consumes/produces arrays.

I think would be good to also add a test to check that the "produces/consumes clause references nonexistent type XXX" error is triggered correctly.

spec/unit/pops/parser/parse_application_spec.rb
lutter: Added that

spec/unit/pops/parser/parse_capabilities_spec.rb
lutter: Added

spec/unit/pops/parser/parser_rspec_helper.rb
lutter: Changed

spec/unit/pops/validator/validator_spec.rb
lutter: Added

lutter:
Here's what I've changed

* addressed the smaller comments around the handling of `with_app_management`
* added a test `spec/unit/capability_spec.rb` to the last commit that does the various checks against PopsBridge suggested above
* rebased to the latest master

spec/unit/pops/parser/parser_rspec_helper.rb
```diff
@@ -4,6 +4,17 @@
...
+
+  def with_app_management(flag)
+    # Horrible things have to be done to test this as the Lexer sets up the
+    # KEYWORD table as a frozen constant, and this list depends on the
+    # Puppet[:app_management] setting. These 'before all' and 'after all'
+    # clauses forces Ruby to reload the Lexer
+    Puppet[:app_management] = flag
+    Puppet::Pops::Parser.send(:remove_const, :Lexer2)
+    load 'puppet/pops/parser/lexer2.rb'
+  end
```
thallgren: Please add the following to ensure that the parser contained in the E4ParserAdapter is cleared if it exists.

```
Puppet::Parser::E4ParserAdapter.class_variable_set(:@@evaluating_parser, nil) if Puppet::Parser.const_defined?(:E4ParserAdapter)`
```
Also, please change the code in the _pops/parser/lexer2_spec.rb_ under "when app_management is (turned) on" so that it uses the `with_app_management(flag)` method.

This will take care of problems when executing tests in different order.

spec/unit/pops/parser/parse_application_spec.rb
```diff
@@ -0,0 +1,42 @@
+#! /usr/bin/env ruby
+require 'spec_helper'
+require 'puppet/pops'
+
+# relative to this spec file (./) does not work as this file is loaded by rspec
+require File.join(File.dirname(__FILE__), '/parser_rspec_helper')
```
thallgren: I think _require_relative 'parser_rspec_helper'_ works OK. I tested it with both rake and rspec.

spec/unit/pops/parser/parser_rspec_helper.rb
lutter: Wow .. great find ! I've made those changes

spec/unit/pops/parser/parse_application_spec.rb
lutter: Ok .. serves me right for mindlessly cutting and pasting from other tests ;) Fixed.

lutter:
Changes in the latest patch series:
* Fixed with_app_management to reach into E4ParserAdapter
* Dropped the patch that changes `$produces` to `$makes` in `evaluating_parser_spec` as that's not needed as long as app_management is feature-flagged
* Rebased to latest HEAD on master

PR #4155, opened 2015-08-13
(PUP-5035) undefined method `keys' for nil:NilClass in static_compiler

tdevelioglu:
get_child_resources() can return nil if the source attribute of the file
resource points to a file instead of a directory.

This results in a raised NoMethodError with the static compiler set as
the catalog terminus and a file resource declared that has the recurse
parameter set to true.

This commit prevents the attempt to add child resources to the catalog when
there are none.

### DIFF
--- a/lib/puppet/indirector/catalog/static_compiler.rb
+++ b/lib/puppet/indirector/catalog/static_compiler.rb
@@ -118,6 +118,13 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
     file = resource.to_ral
 
     children = get_child_resources(request, catalog, resource, file)
+    # get_child_resources() returned early because source is not
+    # a directory, but we still need to replace the metadata of the
+    # resource, so we do it here before returning.
+    if children.nil?
+      find_and_replace_metadata(request, resource, resource.to_ral)
+      return
+    end
 
     remove_existing_resources(children, catalog)
 


kylog:
This looks reasonable, but it's a little surprising to me that a pretty basic case like this would be broken all along, so I'd like to dig into it a bit more to understand the history and what's going on.

Also, @joshcooper, I know you've been looking at static compiler some lately: any comments on this one?

joshcooper:
@kylog yep, this has been broken all along, but it's also an edge case where you've specified:

```
file { '/path/to/thing':
  ensure  => directory,
  recurse => true,
  source  => 'puppet:///actually/path/to/file',
}
```

Since [`recurse` is true](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/catalog/static_compiler.rb#L51), we call `add_children`, which calls `get_child_resources`, but then inside the method we try to collect metadata and realize we stat'ed a file and [return nil](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/catalog/static_compiler.rb#L147).

@tdevelioglu I don't this fix is sufficient as we skip the `replace_metadata` call and never inline the metadata for the `/path/to/thing`. I'm also wondering if this should instead be an error? How does puppet behave when not using the static compiler?

kylog:
@tdevelioglu thoughts on Josh's comments above?

joshcooper:
It looks like puppet's current behavior is to ignore `recurse=>true` for file resources:

```
$ be puppet apply -e 'file { "/tmp/file.txt": recurse => true, source => "puppet:///modules/foo/file.txt" }'
...
Notice: /Stage[main]/Main/File[/tmp/file.txt]/ensure: defined content as '{md5}28c9ed5f4174d1fdcc40b0167545c1c9'
Notice: Applied catalog in 0.10 seconds
$ cat /tmp/file.txt
hi from module
```

I was thinking we should only call `add_children` if we know we're creating a directory and recurse is true, eg

```
if file[:ensure] == :directory && file.recurse?
  add_children(request, catalog, resource, file) 
```

But because the resource doesn't specify `ensure` explicitly, we don't know what we're going to create until we call `file.perform_recursion(src)`. So it makes sense to check the return value of `get_child_resources` like you're doing, but there's still the problem that we never inline the file's metadata.

tdevelioglu:
I remember amending the commit to inline the file's metadata as it dawned on me a while back (but didn't comment anything). Are you still seeing the outdated commit or have I completely missed the nail here ?

@joshcooper is right,  ```get_child_resources``` tells us if it's a file when it returns ```nil```, though I think it should fetch metadata in ```find`` and branch off there.

joshcooper:
@tdevelioglu ah, I think I missed your amended commit, I'll talk a look tomorrow.

peterhuene:
Ping @joshcooper to review.

joshcooper:
ping @tdevelioglu

joshcooper:
Superceded by PR #4402

PR #4156, opened 2015-08-13
(PUP-5012) Add feature flag for authorization

nwolfe:
This commit introduces a new feature flag called 'bypass_authorization'
that toggles on/off auth.conf authorization.

This is part of the larger effort towards replacing auth.conf with an
external system. See SERVER-111 for more information.

### DIFF
--- a/acceptance/tests/authorization/external_authorization.rb
+++ b/acceptance/tests/authorization/external_authorization.rb
@@ -1,0 +1,50 @@
+test_name "Authorization can be disabled with bypass_authorization"
+
+authconf_path = on(master, puppet('config print rest_authconfig')).stdout.strip
+
+step "Copy default auth.conf to restore on teardown" do
+  original_authconf = on(master, "cat #{authconf_path}").stdout
+  teardown do
+    create_remote_file(master, authconf_path, original_authconf)
+  end
+end
+
+cert = on(master, puppet('config print hostcert')).stdout.strip
+key = on(master, puppet('config print hostprivkey')).stdout.strip
+cacert = on(master, puppet('config print localcacert')).stdout.strip
+curl = "curl --cert #{cert} --key #{key} --cacert #{cacert} https://#{master}:8140"
+
+step "Authorization enabled by default" do
+  with_puppet_running_on(master, {}) do
+    on(master, curl + "/puppet/v3/node/foo?environment=production") do
+      assert_match(/Forbidden request/, stdout, 'Expected request denied')
+    end
+  end
+end
+
+step "Authorization disabled with bypass_authorization" do
+
+  step "Install deny-all auth.conf" do
+    create_remote_file(master, authconf_path, <<AUTHCONF)
+path /
+auth any
+deny all
+AUTHCONF
+  end
+
+  step "Verify requests denied by auth.conf" do
+    with_puppet_running_on(master, {:master => {:bypass_authorization => false}}) do
+      on(master, curl + "/puppet/v3/node/#{master}?environment=production") do
+        assert_match(/Forbidden request/, stdout, 'Expected request denied')
+      end
+    end
+  end
+
+  step "Enable bypass_authorization and verify requests succeed" do
+    with_puppet_running_on(master, {:master => {:bypass_authorization => true}}) do
+      on(master, curl + "/puppet/v3/node/#{master}?environment=production") do
+        assert_no_match(/Forbidden request/, stdout, 'Expected request success')
+      end
+    end
+  end
+end
--- a/lib/puppet/network/authconfig.rb
+++ b/lib/puppet/network/authconfig.rb
@@ -78,6 +78,12 @@ module Puppet
     # raise an Puppet::Network::AuthorizedError if the request
     # is denied.
     def check_authorization(method, path, params)
+      if Puppet.settings[:bypass_authorization]
+        Puppet.debug("Bypassing authorization check for call " +
+          "#{method} on #{path} because bypass_authorization is true")
+        return
+      end
+
       if authorization_failure_exception = @rights.is_request_forbidden_and_why?(method, path, params)
         Puppet.warning("Denying access: #{authorization_failure_exception}")
         raise authorization_failure_exception
@@ -86,7 +92,13 @@ module Puppet
 
     def initialize(rights=nil)
       @rights = rights || Puppet::Network::Rights.new
-      insert_default_acl
+
+      if Puppet.settings[:bypass_authorization]
+        Puppet.info("Bypassing insertion of default ACL " +
+          "because bypass_authorization is true")
+      else
+        insert_default_acl
+      end
     end
   end
 end
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1085,6 +1085,14 @@ EOT
       rest indirections.  This can be used as a fine-grained
       authorization system for `puppet master`.",
     },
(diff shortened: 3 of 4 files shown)


nwolfe:
FYI @jeffmccune @camlow325 

@joshcooper are there any other parts of the system that we should be gating behind this feature flag?
We're trying to disable authorization in Ruby Puppet so that it can be done in Puppet Server via the tk-auth replacement.

Happy to add more unit testing if appropriate.

I'm going to work on an acceptance test for this using `curl`+`auth.conf`+`bypass_authorization` in the meantime.

lib/puppet/network/authconfig.rb
```diff
@@ -78,6 +78,12 @@ def mk_acl(acl)
     # raise an Puppet::Network::AuthorizedError if the request
     # is denied.
     def check_authorization(method, path, params)
+      if Puppet.settings[:bypass_authorization] == true
```
camlow325: Cosmetic - do we need to have `== true`?

lib/puppet/network/authconfig.rb
```diff
@@ -86,7 +92,13 @@ def check_authorization(method, path, params)
 
     def initialize(rights=nil)
       @rights = rights || Puppet::Network::Rights.new
-      insert_default_acl
+
+      if Puppet.settings[:bypass_authorization] == true
```
camlow325: Same question here as above.

camlow325:
Are you hoping to add the acceptance test onto this PR?

camlow325:
:see_no_evil:

lib/puppet/network/authconfig.rb
nwolfe: Happy to go either way, this is what @jeffmccune had in his prototype.

I generally prefer to leave off `== true` but I could see an argument for explicitness.

nwolfe:
> Are you hoping to add the acceptance test onto this PR?

If it's ready in time, sure, otherwise just a follow-up PR under the same ticket number.
It might be valuable in this case to get these changes in as soon as possible to catch unexpected things and start getting mileage.

acceptance/tests/authorization/external_authorization.rb
```diff
@@ -0,0 +1,50 @@
...
+    on(master, curl + "/puppet/v3/node/foo?environment=production") do
+      assert_match(/Forbidden request/, stdout, 'Expected request denied')
+    end
+  end
+end
+
+step "Authorization disabled with bypass_authorization" do
+
+  step "Install deny-all auth.conf" do
+    create_remote_file(master, authconf_path, <<AUTHCONF)
```
camlow325: ruby--

acceptance/tests/authorization/external_authorization.rb
```diff
@@ -0,0 +1,50 @@
...
+path /
+auth any
```
camlow325: If you're doing `auth any`, suppose you could get away with not bothering to use a client certificate or key for the curl request.  Probably doesn't matter either way.

lib/puppet/network/authconfig.rb
joshcooper: Can you just do:

```
if Puppet[:bypass_authorization]
```

since the setting is defined to be a `:boolean`, you'll either get back ruby `true` or `false` values.

joshcooper:
I didn't see this explicitly mentioned, but will this affect puppet's `fileserver.conf`? Based on https://docs.puppetlabs.com/puppet/latest/reference/config_file_fileserver.html:

> However, this feature [fileserver.conf] predates the auth.conf file used in this version of Puppet, and we recommend against using it. If possible, you should keep all authorization rules centralized in auth.conf. 

So it sounds like, the default fileserver.conf allows all, and we encourage users to express ACLs in auth.conf, and this PR is in alignment with that, and won't break/affect fileserver.conf?

lib/puppet/network/authconfig.rb
nwolfe: Happily

nwolfe:
> I didn't see this explicitly mentioned, but will this affect puppet's fileserver.conf?

@joshcooper No, they're orthogonal.  I did some manual testing with an `auth.conf`, `fileserver.conf`, and `--[no-]bypass_authorization` and they behave as expected.

Looking at the code for `fileserver.conf`, it doesn't share much code with `auth.conf` other than the `AuthStore` class, which is not where bypass_authorization is honored.

camlow325:
@joshcooper Do you see any other issues with this PR?  Are you good with it being merged?  I'd like to get this in soon since it will be blocking some other Puppet Server work fairly soon.

joshcooper:
@camlow325 thanks, that matches my understanding of auth.conf and fileserver.conf

Also in thinking about how we handle `Puppet::Network::HTTP::Connection`, I'm thinking it might be cleaner to have AuthConfig delegate to an authentication provider, whose class defaults to a ruby implementation, but can be overridden from the clojure side. So something like

```
@authorization_class = Puppet::Network::AuthConfig

def self.authorization_class=(klass)
  @authorization_class = klass
end  
def initialize(...)
    @rights = rights || Puppet::Network::Rights.new
    @provider = self.class.authorization_class.new
    @provider.initialize
end

def check_authorization(..)
  @provider.check(...)
  ...
end
```

The ruby implementation would do what it normally does, eg. calling `insert_default_acls` and performing the check. The clojure side c
=  it be better to have a default ruby implementation that calls `insert_default_acls` and you could inject a different class that does a noop? That way we're not introducing a new setting that doesn't make sense to anyone outside of PL?

If we stick with a setting, can we change the PR to use `if Puppet[:bypass_authorization]` instead of redundant `if Puppet[:bypass_authorization] == true`?

nwolfe:
> If we stick with a setting, can we change the PR to use if Puppet[:bypass_authorization] instead of redundant if Puppet[:bypass_authorization] == true?

Ah whoops, meant to push that change.  Force-push update coming.

nwolfe:
Updated to remove redundant `== true`

camlow325:
> Also in thinking about how we handle Puppet::Network::HTTP::Connection, I'm thinking it might be cleaner to have AuthConfig delegate to an authentication provider

You mean like how in `Puppet::Network::Pool`, Puppet allows the type of `http_client_class` to be configured and would create an instance of the Ruby-provided `Puppet::Network::HTTP::Connection` by default but, in Puppet Server, we use the `http_client_class` hook to provide a reference to a different implementation, `Puppet::Server::HttpClient`?

That's an interesting idea.  I like that it would give the Puppet Server-provided implementation of the provider a little more control for doing something other than a pure no-op in the future, if that were ever needed.

It would mean, I think, that we'd have to ensure that the "hook" were invoked during Ruby initialization at an early enough point where nothing had tried to create an `AuthConfig` instance yet.  I imagine we could figure that out, though.

@cprice404 - what do you think?

nwolfe:
> Also in thinking about how we handle Puppet::Network::HTTP::Connection, I'm thinking it might be cleaner to have AuthConfig delegate to an authentication provider

Makes sense.  I do like the idea of not having to introduce a new Puppet setting.  I'll work up a new PR.

cprice404:
I've only skimmed the comments here so my input should be taken with a grain of salt, but, since you asked for it... :)

The swappable implementation / no-new-setting approach sounds pretty appealing to me, so I'm +1 for that if it's not too difficult.

nwolfe:
Closing in favor of #4180

PR #4157, opened 2015-08-14
(PUP-5037) Fixed `service puppetmaster status` incorrectly returning "puppet dead but subsys locked" on RHEL and derivatives

hamidnazari:
Details here: https://tickets.puppetlabs.com/browse/PUP-5037

### DIFF
--- a/ext/redhat/server.init
+++ b/ext/redhat/server.init
@@ -89,7 +89,7 @@ genconfig() {
 }
 
 rh_status() {
-    status $pidopts $PUPPETMASTER
+    status $pidopts -l $lockfile $PUPPETMASTER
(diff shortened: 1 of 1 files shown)


kylog:
@haus this looks reasonable to me, but I don't have a ton of context here. Thoughts?

haus:
This looks reasonable to me, but the commit message is a little thin. @hamidnazari could you add a little bit from the bug to the commit message so that it is clear why this change is needed and what the issue is?

hamidnazari:
Back to you @haus.

haus:
@kylog this looks like a sane change to me.

PR #4158, opened 2015-08-14
(PUP-5036) Ensure that future reserved words can be used

thallgren:
This commit ensures that the future reserved words 'application'
'consumes', and 'produces' can be used without errors. A warning
will be issued if an attempt is made to use it as a class name.


haus:
```
matthaus@wyclef-2  /Users/matthaus/src/puppet [issue/pup-5036/application-keyword-breaks-class]> puppet apply -e 'class application {}'
Notice: Compiled catalog for wyclef-2.local in environment production in 0.01 seconds
matthaus@wyclef-2  /Users/matthaus/src/puppet [issue/pup-5036/application-keyword-breaks-class]> puppet apply -e 'class application {}' --parser=future
Warning: Use of future reserved word: 'application', must later be quoted if intended to be a String value  at line 1:1
Notice: Compiled catalog for wyclef-2.local in environment production in 0.37 seconds
Notice: Finished catalog run in 0.04 seconds
```

:+1:

peterhuene:
I think we need a different warning when used as a class / defined type name.  Strings are not acceptable for class / defined type names in the grammar, so fixing this as suggested would just lead to a confusing compiler error.

thallgren:
The message is indeed confusing. I would like us to agree on a wording that would cover all cases though. Perhaps just skip the part about the need to quote? IMO, that's implied in for all cases where a string is expected anyway, and it's not confined to use of reserved words.

kylog:
I agree. I.e. just `Use of future reserved word: '#{word}'`?

thallgren:
Updated PR with new wording.

henrik-lindberg:
+1

PR #4160, opened 2015-08-14
(PUP-5041) "environment non existent" acceptance tests failing on puppetserver

er0ck:
This change ensures we run the apply, module, config tests on the master
even if the master does not have an agent role.  We removed the agent
role from the master in the puppet tests to decrease testing time at low
risk.
This change also fixes the environment_scenario-bad test which never
would have passed against current puppet nor server due to matching on
output that was fixed.

[skip ci]

### DIFF
--- a/acceptance/tests/environment/environment_scenario-bad.rb
+++ b/acceptance/tests/environment/environment_scenario-bad.rb
@@ -16,32 +16,32 @@ puppet_conf_backup_dir = create_tmpdir_for_user(master, "puppet-conf-backup-dir"
 apply_manifest_on(master, environment_manifest(testdir), :catch_failures => true)
 
 step  'Test'
+env_path = '/doesnotexist'
 master_opts = {
   'main' => {
-    'environmentpath' => '/doesnotexist',
+    'environmentpath' => "#{env_path}",
   }
 }
 env = 'testing'
-path = '/doesnotexist'
 
 results = use_an_environment(env, 'bad environmentpath', master_opts, testdir, puppet_conf_backup_dir, :directory_environments => true)
 
 expectations = {
   :puppet_config => {
     :exit_code => 1,
-    :matches => [%r{Could not find a directory environment named '#{env}' anywhere in the path.*#{path}}],
+    :matches => [%r{Could not find a directory environment named '#{env}' anywhere in the path.*#{env_path}}],
   },
   :puppet_module_install => {
     :exit_code => 1,
-    :matches => [%r{Could not find a directory environment named '#{env}' anywhere in the path.*#{path}}],
+    :matches => [%r{Could not find a directory environment named '#{env}' anywhere in the path.*#{env_path}}],
   },
   :puppet_module_uninstall => {
     :exit_code => 1,
-    :matches => [%r{Could not find a directory environment named '#{env}' anywhere in the path.*#{path}}],
+    :matches => [%r{Could not find a directory environment named '#{env}' anywhere in the path.*#{env_path}}],
   },
   :puppet_apply => {
     :exit_code => 1,
-    :matches => [%r{Could not find a directory environment named '#{env}' anywhere in the path.*#{path}}],
+    :matches => [%r{Could not find a directory environment named '#{env}' anywhere in the path.*#{env_path}}],
   },
   :puppet_agent => {
     :exit_code => 1,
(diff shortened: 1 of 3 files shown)


acceptance/tests/environment/environment_scenario-bad.rb
```diff
@@ -16,32 +16,32 @@
 apply_manifest_on(master, environment_manifest(testdir), :catch_failures => true)
```
rlinehan: @er0ck This wasn't the test that was failing for us, it was `environment_scenario-non_existent.rb`.

er0ck: thanks.  i forgot.  these changes were required when making the helper run the apply,module,config tests on the master.

i'll add agent roles back on the masters locally and see if the changes to the helper fixes your issues.

camlow325: @er0ck I'm confused as to what is being fixed in this PR.

Is this PR intended to fix the failures we've been seeing with the `environment_scenario-non_existent.rb` test in Puppet Server, as described in PUP-5041?  Note that we saw this fail again as recently as yesterday - http://jenkins-enterprise.delivery.puppetlabs.net/view/puppet-server/view/all/job/enterprise_puppet-server_integration-system_no-conditional_full-master/99/LAYOUT=redhat6-64ma-64a,LDAP_TYPE=default,PLATFORM=default,label=beaker/testReport/junit/(root)/ruby_puppet_acceptance_tests_environment/environment_scenario_non_existent_rb/.

It had seemed that @rlinehan had pinpointed the root cause in her first comment in the PUP-5041:

> It looks like it's using "#{testdir}/environments" as the path (which is the environment path), despite the fact that the path variable is set to master.puppet('master')['codedir'].

That had seemed like a plausible explanation for the mismatched expectations for the error messages that we see when these tests are run from within the Puppet Server CI pipeline.  Is this something you are planning on including in this PR or is there a separate PR coming for that?

er0ck: i added agent roles to the master in our testsuite and with these changes the tests run and pass.  It fixes the issue where the non-master commands were not running if master != agent but also should fix the issue you are seeing.
it runs and passes against whatever server is installed in our aio acceptance suite.

camlow325: @er0ck Is the discrepancy that @rlinehan brought up with `#{testdir}/environments` vs. `master.puppet('master')['codedir']` in the `environment_scenario-non_existent.rb` test also a problem though?  I wouldn't be able to explain why the `environment_scenario-non_existent.rb` test would pass for you without fixing up this discrepancy if that were the case, but do you disagree that this is a problem in the test code that should be addressed?  If you agree that it is a problem that should be fixed, do you think it would make sense to do that as part of this PR?

er0ck: @camlow325 i guess i have not been able to reproduce your issue.  using the code as is, if i add agents to our masters the tests don't fail.

er0ck:
@rlinehan @camlow325 fixed.  sorry for the run-around

camlow325:
Maybe could get rid of the `path = master.puppet('master')['codedir']` line in `environment_scenario-non_existent.rb` now if nothing uses the `path` variable anymore?  Otherwise, looks reasonable to me.  @rlinehan - what do you think?

er0ck:
(doh)  removed that path =

rlinehan:
:+1: I think this looks right to me?

camlow325:
@er0ck Do you think somebody else in QA needs to review this, or should we just merge it?

er0ck:
typically platform devs review, but if you guys are cool with it, i think we should just go for it.

camlow325:
Ah, okay.  So maybe @joshcooper or @hlindberg should give a stamp of approval as well?

rlinehan:
Merging because I don't want to wait another day to get our tests green. We can always revert it if there are objections but given that the tests currently don't work, trying to make them work seems like an improvement.

joshcooper:
@rlinehan sorry missed this, makes sense to merge and it looks like it passed per-commit pipeline in https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Stable/view/Puppet/job/platform_aio-puppet_intn-sys_stable/186/

PR #4163, opened 2015-08-17
(PUP-5002) acceptance: add metadata.json data binding to lookup.rb test

segatspiretech:
(PUP-5002) acceptance: add metadata.json data binding to lookup.rb test

The existing test creates manifests including the bindings directory in the module's lib/puppet/ directory.  The additional test code creates new manifests which remove the bindings directories and install a metadata.json file for each of the two modules.  The new manifest is applied and the asserts are checked
again to verify the same data is returned as was before when the data bindings were made in the module's lib/puppet/bindings.

### DIFF
--- a/acceptance/tests/parser_functions/lookup.rb
+++ b/acceptance/tests/parser_functions/lookup.rb
@@ -44,7 +44,6 @@ hiera_hash_value                = "hiera_class_c"
 hiera_array_value0              = "hiera_array_a"
 hiera_array_value1              = "hiera_array_b"
 
-
 def mod_manifest_entry(module_name = nil, testdir, module_data_implied_key,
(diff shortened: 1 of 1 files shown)


acceptance/tests/parser_functions/lookup.rb
```diff
@@ -90,6 +89,36 @@ def data()
...
+  "dependencies": [
+  ],
+  "data_provider": "function"
+}
+        ',
+        mode => "0644",
+      }
+      file { '#{testdir}/environments/production/modules/#{module_name}/lib/puppet/bindings':
+        ensure => absent,
+	force  => true,
```
er0ck: small whitespace issue.  heredocs have lots of whitespace indentation issues, but we can still be consistent.

PR #4164, opened 2015-08-17
(PUP-4738) Add proper global override support for launchd on 10.10+

keeleysam:
On 10.10+, the overrides are in a different location and in a different format.  This adds support for them, so services can actually be enabled or disabled.


branan:
Since we're going to be merging up 3.x anyway, we're going to close this and just let #4167 follow the normal merge-up path. Please re-open if you think that's really wrong for some reason.

PR #4165, opened 2015-08-17
(maint) remove overloaded git functions from acceptance lib

er0ck:
This change removes build_giturl() and lookup_in_env() from the puppet
acceptance library.  These methods have been moved to beaker where they
can be properly tested and maintained.
This change also uses the new method names, which have been updated to
conform to newer beaker standards.

[skip ci]


rick:
:+1:

joshcooper:
Verified install_utils is in beaker 2.23.0, which is the minimum version puppet currently requires and that a basic git acceptance works.

PR #4167, opened 2015-08-17
(PUP-5044) fix for launchd on OS X 10.10

keeleysam:


keeleysam:
I'd appreciate some help with the tests here, they are a lot more complicated than the ones in the master branch.

MikaelSmith:
I plan to take a look over the next couple days.

MikaelSmith:
I'm not running into this on my laptop, running Darwin 14.5. `/var/db/launchd.db/com.apple.launchd/overrides.plist` doesn't exist, so I must be missing how to trigger it.

keeleysam:
@MikaelSmith correct, `/var/db/launchd.db/com.apple.launchd/overrides.plist` doesn't exist on OS X 10.10 and later as it has been replaced with `/var/db/com.apple.xpc.launchd/disabled.plist`

MikaelSmith:
I think `com.apple.launchd/overrides.plist` wasn't removed as part of an update from OS X 10.9 to OS X 10.10 (which my machine went through, but our VMs didn't) and that the old location is still supported. So the error

lib/puppet/provider/service/launchd.rb
```diff
@@ -181,10 +194,9 @@ def self.job_list
   # it is 10.6 or greater. This allows us to implement different plist
   # behavior for versions >= 10.6
   def has_macosx_plist_overrides?
-    @product_version ||= self.class.get_macosx_version_major
     # (#11593) Remove support for OS X 10.4 & earlier
     # leaving this as is because 10.5 still didn't have plist support
-    return true unless /^10\.[0-5]/.match(@product_version)
```
MikaelSmith: I love how this accidentally disabled using plists for OS X 10.10 and 10.11, since they match the regex.

MikaelSmith: I think this function would be just as clear if it said `return self.class.get_os_version >= 9`. The comments are no longer as clear however, os get_os_version is returning Darwin versions.

Also, I'm not sure the comparison is correct, as according to https://en.wikipedia.org/wiki/Darwin_(operating_system) OS X 10.5 was Darwin 9.

lib/puppet/provider/service/launchd.rb
```diff
@@ -66,12 +66,25 @@ def self.launchd_paths
     ]
   end
 
+  # Gets the current darwin version, example 10.6 returns 9 and 10.10 returns 14
```
MikaelSmith: Linking to https://en.wikipedia.org/wiki/Darwin_(operating_system) might be helpful here.

MikaelSmith:
https://gist.github.com/MikaelSmith/85851aa3c1a9070ec5a6 should cover the spec changes needed. I think the main thing you need is `at_least_once` for the `get_os_version` expectations.

lib/puppet/provider/service/launchd.rb
MikaelSmith: Alternatively you could patch up the regex with `/^10\.[0-5]$/`.

MikaelSmith:
I'd like to see new spec tests that cover OS X 10.10, i.e. os_version 14, with the new plist path.

lib/puppet/provider/service/launchd.rb
```diff
@@ -66,12 +66,27 @@ def self.launchd_paths
...
+  #
+  # @api private
+  def self.get_os_version
+    @os_version ||= Facter.value(:operatingsystemmajrelease).to_i
+  end
+
   # Defines the path to the overrides plist file where service enabling
   # behavior is defined in 10.6 and greater.
   #
+  # With the rewrite of launchd in 10.10+, this moves and slightly changes format.
```
MikaelSmith: Is there an documentation from Apple on this change?

MikaelSmith:
Merging this; I'll follow-up with new specs and an acceptance test.

PR #4169, opened 2015-08-17
(PUP-5045) Evaluate produced/consumed resources 

lutter:


thallgren:
Don't find anything wrong with this code on first review but please add tests.

lutter:
Rebased to master; no tests added yet.

lutter:
Added some tests to check the behavior of producing/consuming resources in `spec/unit/capability_spec.rb`

lib/puppet/parser/resource.rb
```diff
@@ -199,6 +199,51 @@ def raw_tagged?(tag_array)
     super || ((scope_resource = scope.resource) && !scope_resource.equal?(self) && scope_resource.raw_tagged?(tag_array))
   end
 
+  # Fill resource params from a capability
+  #
+  # This backs 'consumes => Sql[one]'
```
hlindberg: We should put `@api private` on methods we do not want people to be calling from their own logic.

lib/puppet/parser/resource.rb
```diff
@@ -199,6 +199,51 @@ def raw_tagged?(tag_array)
     super || ((scope_resource = scope.resource) && !scope_resource.equal?(self) && scope_resource.raw_tagged?(tag_array))
   end
 
+  # Fill resource params from a capability
```
hlindberg: picky: uses imperative form "Fill", "Fills" is preferred. Also, it looks like it fills parameters from all entities that it is consuming.

lib/puppet/parser/resource.rb
```diff
@@ -199,6 +199,51 @@ def raw_tagged?(tag_array)
...
+  #
+  # This backs 'consumes => Sql[one]'
+  def add_parameters_from_consume
+    return unless self[:consume]
```
hlindberg: this should probably be `return if self[:consumes].nil?` to avoid being screwed if code managed to get a boolean false in there.

lib/puppet/parser/resource.rb
```diff
@@ -199,6 +199,51 @@ def raw_tagged?(tag_array)
...
+
+    map = {}
+    [ self[:consume] ].flatten.map do |ref|
+      # Resolve references
+      catalog.resource(ref.type, ref.title) or
```
hlindberg: We do not want Ruby 'or' in the Puppet code base. Please Refactor.

lib/puppet/parser/resource.rb
```diff
@@ -199,6 +199,51 @@ def raw_tagged?(tag_array)
...
+        if expr = blueprint[:mappings][name]
+          # Explicit mapping
+          value = expr.safeevaluate(cns_scope)
+        else
+          value = cns[name]
+        end
+        if value
+          # @todo lutter 2015-07-01: this should be caught by the checker
+          # much earlier. We consume several capres, at least two of which
+          # want to map to the same parameter (APPMGMT-760)
```
hlindberg: reference to ticket that is not public.

lib/puppet/parser/resource.rb
```diff
@@ -199,6 +199,51 @@ def raw_tagged?(tag_array)
...
+
+      # evaluate mappings in that scope
+      resource_type.arguments.keys.each do |name|
+        if expr = blueprint[:mappings][name]
+          # Explicit mapping
+          value = expr.safeevaluate(cns_scope)
+        else
+          value = cns[name]
+        end
+        if value
```
hlindberg: what if value is a boolean? Is the intent to test against nil ?

lib/puppet/parser/resource.rb
```diff
@@ -199,6 +199,51 @@ def raw_tagged?(tag_array)
...
+          # @todo lutter 2015-07-01: this should be caught by the checker
+          # much earlier. We consume several capres, at least two of which
+          # want to map to the same parameter (APPMGMT-760)
+          raise "Several consumed mappings try to set #{name}" if map[name]
+          map[name] = value
+        end
+      end
+    end
+
+    map.each { |name, value| self[name] = value unless self[name] }
```
hlindberg: this will probably also have problems with boolean values, the unless part should probably explicitly test against non nil. `self[name] = value if self[name].nil?`

lib/puppet/resource.rb
```diff
@@ -290,6 +290,12 @@ def is_capability?
     resource_type and resource_type.is_capability?
   end
 
+  # Return the value of the 'export' metaparam as an Array
```
hlindberg: Use "Returns" instead of "Return". Is the API public? if not use `@api private`

lib/puppet/resource/capability_finder.rb
```diff
@@ -0,0 +1,61 @@
...
+# @todo lutter 2015-03-10: determine whether this should be based on
+# Puppet::Pops::Evaluator::Collectors, or at least use
+# Puppet::Util::Puppetdb::Http
+
+require 'net/http'
+require 'cgi'
+
+module Puppet::Resource::CapabilityFinder
+
+  # Look the capability resource with the given +type+ and +title+ up from
```
hlindberg: "Looks up the..."

lib/puppet/resource/capability_finder.rb
```diff
@@ -0,0 +1,61 @@
...
+# A helper module to look a capability up from PuppetDB
+#
+# @todo lutter 2015-03-10: determine whether this should be based on
+# Puppet::Pops::Evaluator::Collectors, or at least use
+# Puppet::Util::Puppetdb::Http
+
+require 'net/http'
+require 'cgi'
+
+module Puppet::Resource::CapabilityFinder
```
hlindberg: The module is probably `@api private`

lib/puppet/resource/capability_finder.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    Puppet.notice "Capability lookup #{cap}]: #{query}"
+    response = http.get("/pdb/query/v4/resources?query=#{CGI.escape(query)}",
+                        { "Accept" => 'application/json'})
+
+    json = response.body
+
+    # @todo lutter 2014-11-04: use just JSON
+    data = PSON.parse(json)
+    data.is_a?(Array) or raise Puppet::DevError,
```
hlindberg: We do not want Ruby `or` in the code base, please refactor.

lib/puppet/parser/resource.rb
lutter: Added comment

lib/puppet/parser/resource.rb
lutter: Changed

lib/puppet/resource/capability_finder.rb
```diff
@@ -0,0 +1,61 @@
...
+                    ["=", "tag", "producer:#{environment}"]].to_json
+
+    Puppet.notice "Capability lookup #{cap}]: #{query}"
+    response = http.get("/pdb/query/v4/resources?query=#{CGI.escape(query)}",
+                        { "Accept" => 'application/json'})
+
+    json = response.body
+
+    # @todo lutter 2014-11-04: use just JSON
+    data = PSON.parse(json)
```
hlindberg: this parsing is unprotected and will produce a bad error message if there is something wrong with the response body in json. Need to catch errors from that and fail the same way (probably) as if an array was not produced.

lib/puppet/parser/resource.rb
lutter: Changed, though if this is false, we are in trouble, anyway - it should be a resource reference or an array of resource references to instances of capability types ...

lib/puppet/parser/resource.rb
lutter: Changed to
```ruby
if  catalog.resource(ref.type, ref.title).nil?
  raise "..."
end
```

lib/puppet/parser/resource.rb
lutter: Changed to PUP-5080

lib/puppet/resource/capability_finder.rb
```diff
@@ -0,0 +1,61 @@
...
+    data.is_a?(Array) or raise Puppet::DevError,
+    "Unexpected response from PuppetDB when looking up #{cap}: " +
+      "expected an Array but got #{data.inspect}"
+
+    # @todo lutter 2014-11-13: this was in the original capabilities
+    # prototype; can we really get entries back that do not have
+    # parameters set ?
+    data = data.select { |hash| hash["parameters"] }
```
hlindberg: We prefer `map` and `reduce` in the code base over `select` and `inject`.
Also, since the array comes from a call over http, any type of problem can be anticipated. In any case, the line in question neds some kind of explanation wrt. what is expected in terms of the return value - this maps all entries in the array, and expects them to be hashes, getting the "parameters" for each, and then returning an array with all of those.

lib/puppet/resource/capability_finder.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    Puppet.notice "Capability lookup #{cap}: response #{data}"
+    data.size <= 1 or fail Puppet::ParseError,
```
hlindberg: Do not want Ruby `or` in the code base, please refactor.

lib/puppet/parser/resource.rb
lutter: Yes, `blueprint[:mappings][name]` is either `nil` or a `PopsBridge::Expression` (gets set up in `PopsBridge#instantiate_CapabilityMapping`)

lib/puppet/parser/resource.rb
lutter: Changed

lib/puppet/resource/capability_finder.rb
```diff
@@ -0,0 +1,61 @@
...
+    "Multiple resources found in PuppetDB when looking " +
+      "up #{cap}: #{data.inspect}"
+
+    unless data.empty?
+      hash = data.first
```
hlindberg: can we use a more descriptive variable name than 'hash' here?

lib/puppet/resource.rb
lutter: Changed to "Returns", marked as private

lib/puppet/resource/capability_finder.rb
```diff
@@ -0,0 +1,61 @@
...
+      resource = Puppet::Resource.new(hash["type"], hash["title"])
+      real_type = Puppet::Type.type(resource.type) or
```
hlindberg: Ruby `or` - refactor

lib/puppet/resource/capability_finder.rb
lutter: Absolutely. Changed

lib/puppet/resource/capability_finder.rb
lutter: Changed

lib/puppet/resource/catalog.rb
```diff
@@ -4,6 +4,8 @@
 require 'puppet/util/tagging'
 require 'puppet/graph'
 
+require_relative 'capability_finder'
```
hlindberg: why a relative require? There should be no need for that?

lib/puppet/resource/capability_finder.rb
lutter: Changed to a `unless data.is_a?(Array) ..`

lib/puppet/resource/capability_finder.rb
lutter: Changed this to using `JSON.parse` and catching `JSON::JSONError`

lib/puppet/resource/capability_finder.rb
lutter: Changed to `map`. Added a comment pointing to [PuppetDB docs](http://docs.puppetlabs.com/puppetdb/3.0/api/query/v4/resources.html#response-format) and cleaned up the mess of how this was handled

lib/puppet/resource/capability_finder.rb
lutter: Changed

lib/puppet/resource/capability_finder.rb
lutter: Changed it to `resource_hash`

lib/puppet/resource/capability_finder.rb
lutter: Changed

lib/puppet/resource/catalog.rb
lutter: I thought that was best practice after Ruby 1.9, as it prevents various loadp ath shenanigans; changed it to an absolute require

lutter:
Made the changes indicated in the comments and rebased to latest master

lib/puppet/parser/resource.rb
hlindberg: well, if there is an "explicit mapping" - it calls ´expr.safeevaluate´ - then the result of that is assign to ´value´. I assume that can return a boolean ´false´ value. If it does, the value will not be used to set `map[name] = value` on line 243. 
Suggest changing to ´unless value.nil?´

lib/puppet/resource/type.rb
```diff
@@ -105,6 +105,42 @@ def child_of?(klass)
...
   end
 
+  # Evaluate the resources produced by the given resource. These resources are
+  # evaluated in a separate but identical scope from the rest of the resource.
+  def evaluate_produces(resource, scope)
+    # Only defined types can produce capabilities
+    return unless definition?
+
+    resource.export.map do |ex|
+      blueprint = produces.find { |pr| pr[:capability] == ex.type } or
```
hlindberg: use of `or`

lib/puppet/resource/type.rb
```diff
@@ -105,6 +105,42 @@ def child_of?(klass)
...
+        raise Puppet::ParseError, "Resource type #{resource.type} does not produce a #{ex.type}"
```
hlindberg: picky a/an article ? possibly rephrase to avoid having to us article?

lib/puppet/resource/type.rb
```diff
@@ -105,6 +105,42 @@ def child_of?(klass)
...
+      # Tag the produced resource so we can later distinguish it from
+      # copies of the resource that wind up in the catalogs of nodes that
+      # use this resource. We tag the resource with producer:<environment>,
+      # meaning produced resources need to be unique within their
+      # environment
+      # @todo lutter 2014-11-13: we would really like to use a dedicated
+      # metadata field to indicate the producer of a resource, but that
+      # requires changes to PuppetDB and its API; so for now, we just use
+      # tagging
+      produced_resource.tag("producer:#{scope.catalog.environment}")
```
hlindberg: I am somewhat concerned about the use of ':' in the tag. I know we do split tags on '::'. In general the spec of allowed characters in tags is weak (understatement), and it there may be parts of the ecosystem that could have difficulties - but it may also just work fine...

lib/puppet/resource/type.rb
```diff
@@ -351,6 +390,13 @@ def set_argument_types(name_to_type_hash)
     end
   end
 
+  # Instances of this class are never capability resources, but they end up
```
hlindberg: Should document what the method returns - suggest "Returns boolean true if an instance of this type is a capability. This implementation always returns false. This "duck-typing" interface is shared among other classes and makes it easier to detect capabilities when they are intermixed with non capability instances." (my interpretation of the comment)

lib/puppet/type.rb
```diff
@@ -1638,6 +1638,14 @@ class { 'apt-updates':
     }
   end
 
+  newmetaparam(:export, :parent => RelationshipMetaparam, :attributes => {:direction => :out, :events => :NONE}) do
+    desc "Control which produced capabilities are actually instantiated"
```
hlindberg: this is a little skimpy - this text is what will appear as the on-line documentation. Most others have around 10-15 lines of text and include an example.

lib/puppet/type.rb
```diff
@@ -1638,6 +1638,14 @@ class { 'apt-updates':
...
+  end
+
+  newmetaparam(:consume, :parent => RelationshipMetaparam, :attributes => {:direction => :in, :events => :NONE}) do
```
hlindberg: Same here as for line 1642

lib/puppet/parser/resource.rb
lutter: Changed to `unless value.nil?`

lib/puppet/resource/type.rb
lutter: Changed to an `if blueprint.nil?` after the assignment

lib/puppet/resource/type.rb
lutter: Removed the indefinite article - it works without it, too

lib/puppet/resource/type.rb
lutter: What are known valid characters for tags, especially ones that can't be used in environment names (not that the latter is too important here)

lib/puppet/resource/type.rb
lutter: Changed the comment to your suggestion.

spec/unit/capability_spec.rb
```diff
@@ -128,18 +135,138 @@ class mod {
...
+  describe "exporting a capability" do
+    before(:each) do
+      make_cap_type
+    end
+
+    after :each do
+      Puppet::Type.rmtype(:cap)
+    end
+
+    it "does not add produced resources that are not exported" do
```
hlindberg: nitpic: double negation is hard to read

spec/unit/capability_spec.rb
```diff
@@ -128,18 +135,138 @@ class mod {
...
+    end
+
+    def mock_cap_finding
+      cap = Puppet::Resource.new("Cap", "two")
+      cap["host"] = "ahost"
+      Puppet::Resource::CapabilityFinder.expects(:find).returns(cap)
+      cap
+    end
+
+    it "does not fetch a consumed resource when consume metaparam not set" do
```
hlindberg: are there two sides to this: "fetches a consumed resource when consume metaparam set" and then the test above. if so, it would be good if they are grouped (maybe there is such a test, but it is hard to manually correlate). e.g. using a context "consume metaparam controls fetching such as", it "when not set, does not fetch a consumed resource", and it "when set, fetches a consumed resource". (such grouping helps QA review test coverage).

lib/puppet/resource/type.rb
hlindberg: tags typically fall under the rules of names - that is if they are to be able to be processed by puppet logic - but there is no real specification. I am willing to let the ':' stay - it obviously works. What to do here depends on the requirements what users should be able to do. An underscore '_' would be preferable and remove the concern altogether. Not sure if that works given other requirements?

hlindberg:
Finished with nitpic code review. Ready to be merged when they have been addressed.

hlindberg:
If possible rewrite the 3 last commits to use imperative form - e.g. "Add method 'foo' to Class" to make them comply with policy.

lib/puppet/type.rb
lutter: Added more extensive docs

lib/puppet/type.rb
lutter: Added more extensive docs

lutter:
Addressed comments and rebased to latest master

PR #4170, opened 2015-08-18
(PUP-4639) always add -w when starting launchd

keeleysam:
Always adds `-w` when starting a job, as the status is cached and you can end up getting `unload -w` and then just `load`


keeleysam:
#4171

PR #4171, opened 2015-08-18
(PUP-4639) always add -w when starting launchd

keeleysam:
This is the right branch, closing #4170 although this should go to master as well.

Always adds `-w` when starting a job, as the status is cached and you can end up getting `unload -w` and then just `load`.

Adding the `-w` is a noop when it isn't needed.

### DIFF
--- a/lib/puppet/provider/service/launchd.rb
+++ b/lib/puppet/provider/service/launchd.rb
@@ -246,8 +246,10 @@ Puppet::Type.type(:service).provide :launchd, :parent => :base do
     did_enable_job = false
     cmds = []
     cmds << :launchctl << :load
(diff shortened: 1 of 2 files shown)


MikaelSmith:
I'm :+1: on this, I've tried all 4 combinations of refresh `ensure => [stopped, running], enable => [false, true]`, ensuring they're idempotent and correctly transition. I'm going to look at adding an acceptance test after merging up to stable.

MikaelSmith:
@nigelkersten if you have time, can you double-check whether this makes sense? Looks like you were involved in prior work on it (http://projects.puppetlabs.com/issues/2581).

nigelkersten:
This sounds reasonable and looks like you've tested the right things. +1

PR #4172, opened 2015-08-18
(PUP-1189) declare report handlers with overwrite

kylog:
Previously, if a report handler failed during initialization, it wasn't
cleaned up. So the second time a report was submitted, Puppet would try
to instantiate a new class, and then fail because it had already been
instantiated. This instructs the class loader to overwrite if needed.

(PUP-1189) add spec test

(PUP-1189) Corrected spec test for accuracy

Thanks for the help, @MikaelSmith

### DIFF
--- a/lib/puppet/reports.rb
+++ b/lib/puppet/reports.rb
@@ -54,7 +54,11 @@ class Puppet::Reports
   def self.register_report(name, options = {}, &block)
     name = name.intern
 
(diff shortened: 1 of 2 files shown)


kylog:
This is #4118 with the spec test tweak I suggested there.

MikaelSmith:
:+1:

binford2k:
:+1:

PR #4173, opened 2015-08-18
(PUP-5051) install.rb improvements

haus:
This PR includes 2 commits. The first removes ftools usage from install.rb, as ftools has not been available in ruby since 1.8.7. The second adds a new option to allow for skipping prerequisite library checks, which is useful in situations where the ruby/platform being built on won't be the ruby/platform at runtime.


haus:
This does the right thing for me in my uses.

shrug:
:+1:

PR #4174, opened 2015-08-19
(PUP-1455) Use CFPropertyList to read launchd plists

keeleysam:
This uses the CFPropertyList gem which is shipped with Puppet to read launchd plists instead of shelling out to plutil to convert them and then reading them in.  This saves a lot of time, between 10 and 20 seconds per Puppet run.


ahpook:
@keeleysam you're :fire: :fire: :fire: ! thanks for digging in to this!

ahpook:
I think this should point at `master` though, not 3.x? Unless y'all talked this through at triage, in which case ignore me.

keeleysam:
It should of course go to `master`, but I think it qualifies to go against `3.x` as well.  It's really a longstanding issue and 3.x ships with cfpropertylist.

keeleysam:
Puppet 4's current AIO packages do not include cfpropertylist, so that will need to be added for it to work there.

glarizza:
I think this behavior (reading/writing plists) would be best served in a Util class somewhere because it's going to be necessary across much of the OS X functionality...so nesting the reading/writing inside the methods of this provider is probably not quite the best place (but it's a start).  

If someone from Eng is on-board with that (and provides the appropriate namespaces for where we should create those methods), I've no problem helping to mock them up and provide some tests.

glarizza:
Yeah, I think I'm gonna take this on this week.  @ahpook do you have any guidance where a Util-like class with helper methods for reading/writing plists using CFP should live?  Give me a namespace and I'll make it happen.  Then we can just mix in those methods with Sam's PR and then I can add some tests that SHOULDN'T need to change much if I've encapuslated things well.

ahpook:
@glarizza  Puppet::Util seems like a good option.

glarizza:
Cool, let me dig into Puppet::Util::Plist and see if I can hack this up today or tomorrow.  I'll then try to mix it into this PR and see what I can do about other plist-based providers

ccaviness:
Pleeeaaase fix this. Shelling out to plutil hundreds of times is just terribad.

glarizza:
Turns out the work I did on my module (puppet-cfpropertylist) can just be ported over.  I did this thing --> https://github.com/glarizza/puppet-1/commit/d92499dbefc1339621807dd14578b9ad12d1e440

Does that look sane?  Two things I could use @ccaviness and @keeleysam help testing:  I had problems with `/System/Library/LaunchDaemons/org.cups.cupsd.plist` and `/System/Library/LaunchDaemons/org.ntp.ntpd.plist` being improperly parsed for varying reasons.  You'll see the stupid exception handling necessary for them.  They SEEM to be fixed in 10.10, right?  I had to work around it in 10.9, so I'm guessing they were fixed in 10.10 and higher...can you verify this for me?  

Especially insidious is THIS behavior:  https://github.com/glarizza/puppet-1/commit/d92499dbefc1339621807dd14578b9ad12d1e440#diff-ab9edadac4bb9b01d006980021378335R46

I'd like to NOT have to do that.  At any rate, this code seems to "work for me".  I'd like to then bolt it on to the user resource since we handle plists there in older versions.  Whatcha'll think?

keeleysam:
@glarizza thanks for your work here!  I can confirm that the NTP and CUPS plists are valid XML in 10.10 and 10.11, I'll figure out later where that changed as I'm on an airplane now with no VMs. 

I agree that the hack around for improperly formatted plists shouldn't have to exist, but I ran into this wonderful one last week: https://gist.github.com/keeleysam/3fd478bec6809d0db6f2.  Apple's tools don't care about the bad formatting as they assume you're giving them plists, but of course libxml blows up on that too.

For adding support to users and groups where we must deal with nested binary plists, it should probably be able to deal with plists from strings and not just files.  This could help remediate the shell out to `/usr/bin/dscl -plist . read /Users/username ShadowHashData` for every user on the system when the same data is in /var/db/dslocal/nodes/Default/users/username.plist... in a plist inside a plist inside a key.  I've been playing with that over the weekend and it saves a lot of runtime, but this will definitely enable it to be a lot cleaner.

I'm going to do some additional testing and make sure it all works right, but it will need the changes #4164 and #4171 too in order to function properly on 10.10+.

keeleysam:
A little more information on malformed plists...

`/System/Library/LaunchDaemons/org.cups.cupsd.plist` has that comment still, but since 10.10 it has been properly formatted.
10.9.5: ``
10.10.5, 10.11: `<!-- These environment variables are only used when CUPS is compiled with debug-printfs support -->`

The in-memory patching for the header is cool, but I know of at least two cases where it is insufficient: the AnyConnect one where the header is over two lines with no quotes, and this CUPS one on 10.9.

As much as I hate shelling out, it's probably okay in limited cases especially since it will only be done a couple of times at most instead of hundreds of times as it goes over every launchd.  I think the rescue case should be to send the plist through `plutil -convert xml1 -o -` which will return a valid XML document to stdout.  It strips the invalid comments and always gives us a correct header.

keeleysam:
Also I did want to say that the code does work as-is for me on 10.10 and 10.11, though I don't have any malformed plists so they never have to get fixed.

glarizza:
Okay....I don't really want to use the $stderr redirection on something built-in to the product (doing it out of my module I didn't really care).  If it's fixed in 10.10, I'm good with it, BUT we need a way to handle the error condition in a way where anyone hitting it knows exactly what borked and can be directed somewhere to fix it.

@ahpook - can you give us some guidance as to what versions of OS X we should be targeting to support?  I think we always aim for N - 1, which would be 10.9 (Mavericks) and higher...which means we would need the fix still.  

@keeleysam - As for the cisco plist, I'm sure you've alerted Cisco, but yes, any additional error handling would be quite awesome.  Do you wanna pull in my commit into your branch and apply it to the PR?  You can just `git cherry-pick` it after adding my remote and fetching all the branches.  The next thing to target would be tests; I have no problem getting started on those.

As for the users plist - yeah, reading a plist based on a string is definitely necessary.  I DO have that code in one of my branches SOMEWHERE as I remember making it work (whether in the puppet-cfpropertylist module or one of the branches of my puppet repo fork).

I'll be in Portland next week with more dedicated time to work on this.  I think we can have it ready for merge by the end of next week (barring me being pulled onto something)

ahpook:
@glarizza Puppet only really supports 10.9 and 10.10 at this point, and @keeleysam is helping us get ready for 10.11.

glarizza:
Okay, to bring the work around, I think I'm going to look to strip out the $stderr redirection since it works for 10.10 and presumably 10.11.  If there's some 10.9 around, let's test it on that.  If it BREAKS on 10.9, then we need a better way around it...so I'll need some help triaging that.

I'm going to clean up that code and add some tests today.

glarizza:
@keeleysam I added some spec tests in my branch at https://github.com/glarizza/puppet-1/tree/cfplist  I tried to file a pull request against your fork but that didn't look too hot.  Do you wanna try to cherry-pick in the most recent commits:

2c33029b04e292a7c170c113e2367bf49d61d6f6
1f1da2fe8f216e1f90258639574e57fb1f87817e
c472f9c6faa6728bb81c6bfda85abb4f414b825d
a8ad5398a00171f06a1be64f6126a369b40185c8

ccaviness:
I'm trying to patch this in to test. It seems to work with `puppet resource service` just fine (and holy fast!), but actual puppet runs start complaining about `undefined method 'parse_xml'`, which I think is from directoryservices.rb still using facter/util/plist and then ruby getting confused and augh I hate ruby. Also git.

glarizza:
@ccaviness Good catch, lemme look at that when I get a free second...

glarizza:
For all involved, I can't make the PR Triage tomorrow, but if someone CAN and wants to discuss it, this PR needs rebased to pull in more recent commits the puppet core.  Once that happens and tests pass, it can be subject for review.

I do think that we need to first chase down the issues Clay is having.  I'm not free tomorrow, but will be in Portland the following week to hopefully help move this down the line

branan:
@peterhuene @whopper Re: garry's comment above, would be good to give this a once-over in next week's pre-iage. I've scanned it and it looks basically sane, but my OSX-fu is pretty minimal and I of course can't test it in my environment

lib/puppet/util/plist.rb
```diff
@@ -0,0 +1,100 @@
+require 'cfpropertylist'
```
MikaelSmith: Facter may not always have a dependency on CFPropertyList. The Gemfile should explicitly include it.

lib/puppet/provider/service/launchd.rb
```diff
@@ -1,4 +1,4 @@
-require 'facter/util/plist'
+require 'puppet/util/plist'
```
joshcooper: This adds a hard dependency on `cfpropertylist` (via `puppet/util/plist`), which will prevent puppet code from being loaded on any system that doesn't have the `cfpropertylist` gem, e.g. on the master during compilation. The typical way to create a soft dependency from a provider is to create a `cfpropertylist` feature in `lib/puppet/features`:

```
require 'puppet/util/feature'
Puppet.features.add(:cfpropertylist, :libs => ["cfpropertylist"])
```

And then confine the provider to that feature:

```
confine :feature => :cfpropertylist
```

joshcooper:
@ccaviness @glarizza thanks for working on this, the speed differences are amazing. I'm going to suggest we bring this into the Client team sprint starting tomorrow and we can dedicate some time to ironing out the PR.

glarizza:
That's awesome Josh, thanks for that :) My other module has a feature defined for it, I should have pulled that in with this PR.

joshcooper:
@glarizza @ccaviness Unfortunately we had to bump this one until next sprint

tasks/cfpropertylist.rake
```diff
@@ -0,0 +1,15 @@
+task 'cfpropertylist' do
+  if defined? Pkg::Config and Pkg::Config.project_root
+    cfp_version = "2.3.1"
```
MikaelSmith: 2.3.1 won't work well with Facter 2.x, it pins to the 2.2 line (`~> 2.2.6`).

MikaelSmith:
I'm working on a number of updates to this, I'll put up a new PR when they're ready.

MikaelSmith:
Closing in favor of https://github.com/puppetlabs/puppet/pull/4447 (3.x) and https://github.com/puppetlabs/puppet/pull/4464 (master).

PR #4176, opened 2015-08-19
(PUP-5032) Change configured_environment behavior

MikaelSmith:
Puppet's communication with the node and catalog terminus included a
configured_environment option in the last release. That option sometimes
includes redundant or irrelevant information when Puppet[:environment]
has the default setting. That redundant information precludes it from
being used to identify when an agent has explicitly requested a
particular environment.

Change the behavior of configured_environment to only be set when the
`environment` setting is explicitly set.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -177,6 +177,8 @@ class Puppet::Configurer
               @environment = node.environment.to_s
               report.environment = @environment
               query_options = nil
+            else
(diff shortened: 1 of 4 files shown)


joshcooper:
@MikaelSmith looks like an extra merge commit? Could you rebase against stable and force push to your topic branch?

MikaelSmith:
I guess that's how the 'Update branch' button works, I'll rebase.

PR #4177, opened 2015-08-19
(PUP-3318) Change environment change warning to notice

MikaelSmith:
When the puppet agent is instructed to change environments by
classification, it issues a warning. That change is expected, so a
warning is not.

Change the warning to a notice. Later environment changes caused by
reclassification based on new facts are still warnings, as changing
environments multiple times isn't generally desirable.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -171,7 +171,7 @@ class Puppet::Configurer
             end
 
             if node.environment.to_s != @environment
-              Puppet.warning "Local environment: \"#{@environment}\" doesn't match server specified node environment \"#{node.environment}\", switching agent to \"#{node.environment}\"."
+              Puppet.notice "Local environment: \"#{@environment}\" doesn't match server specified node environment \"#{node.environment}\", switching agent to \"#{node.environment}\"."
               @environment = node.environment.to_s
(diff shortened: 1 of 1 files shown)


joshcooper:
Should the catalog convergence case also be a notice too? An argument in favor is if you have a brand new node that has never plugin synced before, and a NC rule that sets environment based on the fact. The initial node request will return "production", but then the node will pluginsync (from production), get the fact, submit the catalog request, and the compiler will consult the NC generating a catalog for a different environment.

I could also be convinced to leave the it as a warning too, more that I wanted to ask the question and get your thoughts.

MikaelSmith:
That's what part of the commit was alluding to; that's definitely a valid use-case. Maybe everything should be info/notice unless environment was explicitly specified, and then it should always be a warning?

MikaelSmith:
If it doesn't stabilize it'll error, which is the problem we really care about.

branan:
:+1: Notice-level as long as it stabilizes seems right - this is the normal/expected behavior of the agent in the face of an NC, so we want to describe what's going on for debugging but not make a big deal about it.

keeleysam:
👍🏻 Good stuff, this will help out in gathering actual warnings from testing environments.

PR #4178, opened 2015-08-19
(docs) Revise reduce function docs.

gguillotte:


lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,94 @@
...
-#
-#       Integer[1,4].reduce |$memo, $x| { $memo + $x }
-#       #=> 10
+# Applies a [lambda](https://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html)
+# to every value in a data structure from the first argument, carrying over the returned
+# value of each iteration, and returns the result of the lambda's final iteration. This
+# lets you create a new value or data structure by combining values from the first
+# argument's data structure.
+#
+# This function takes two mandatory arguments, in this order:
```
nfagerlund: Needs a line break before the ordered list.

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,94 @@
...
+# 1. An array or hash the function will iterate over.
+# 2. A lambda, which the functions calls for each element in the first argument.
+#
+# The lambda takes two parameters, in this order:
```
nfagerlund: line break.

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,94 @@
...
+# 1. A memo value that is overwritten after each iteration with the iteration's result.
+# 2. A second value that is overwritten after each iteration with the next value in the
+# function's first argument.
```
nfagerlund: What do you think about making this a nested list under the lambda, to make it clear where these parameters are associated?

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,94 @@
...
-#       #=> 10
+# Applies a [lambda](https://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html)
+# to every value in a data structure from the first argument, carrying over the returned
+# value of each iteration, and returns the result of the lambda's final iteration. This
+# lets you create a new value or data structure by combining values from the first
+# argument's data structure.
+#
+# This function takes two mandatory arguments, in this order:
+# 1. An array or hash the function will iterate over.
+# 2. A lambda, which the functions calls for each element in the first argument.
```
nfagerlund: typo: should be function (non-plural)

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,94 @@
...
+# the lambda in order, then executes the lambda. If you pass a start memo, Puppet instead
+# passes it and the data structure's first value to the lambda's parameters before its
+# first iteration.
+#
+# The lambda returns each iteration's result to the memo parameter (`$memo` in the above
+# examples), then passes the data structure's next value to the lambda's second parameter
+# (`$value` above). Puppet repeats the process for each of the data structure's remaining
+# values.
+#
+# If the structure has one value, Puppet returns the value and does not process the
```
nfagerlund: does not execute

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,94 @@
...
+#
+# When the first argument (`$data` in the above example) is an array, Puppet passes each
+# of the data structure's values in turn to the lambda's parameters. When the first
+# argument is a hash, Puppet converts each of the hash's values to an array in the form
+# `[key, value]`.
+#
+# If you don't pass a start memo, Puppet passes the data structure's first two values to
+# the lambda in order, then executes the lambda. If you pass a start memo, Puppet instead
+# passes it and the data structure's first value to the lambda's parameters before its
+# first iteration.
```
nfagerlund: I had to read these two sentences a few times before I got it... but that might just be because the behavior is bit complicated.

It might be more clear to say: 

> If you don't pass a start memo, Puppet uses the data structure's first value as the initial memo and starts calling the lambda with the _second_ value. If you do pass a start memo, Puppet will start with the structure's first value. 

I dunno. Do you have any more thoughts about this paragraph?

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,94 @@
...
+#
+# The lambda returns each iteration's result to the memo parameter (`$memo` in the above
+# examples), then passes the data structure's next value to the lambda's second parameter
+# (`$value` above). Puppet repeats the process for each of the data structure's remaining
+# values.
```
nfagerlund: I prefer starting this paragraph with the final sentence instead:

> Puppet calls the lambda for each of the data structure's remaining values. For each call, it passes the result of the previous call as the first parameter (`$memo` in the above examples) and the next value from the data structure as the second parameter (`$value`).

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,94 @@
...
+#
+# If the structure has one value, Puppet returns the value and does not process the
+# lambda.
+#
+# @example Using the `reduce` function with a two-parameter lambda
```
nfagerlund: Since you can't use a one- or three-parameter lambda, I'd drop that part and distinguish the examples by just the presence of the start memo.

lib/puppet/functions/reduce.rb
gguillotte: You can pass an initial start memo value; if you do, the lambda uses the provided memo value and the structure's first value. "Puppet will start with the structure's first value" isn't complete.

> If you pass a start memo value, Puppet executes the lambda with the memo value and the data structure's first value. Otherwise, Puppet passes the structure's first two values to the lambda.

Is that clearer?

lib/puppet/functions/reduce.rb
gguillotte: Can do.

lib/puppet/functions/reduce.rb
nfagerlund: Yeah! Use that.

lib/puppet/functions/reduce.rb
gguillotte: Amended commit with changes.

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,98 @@
...
+# argument's data structure.
+#
+# This function takes two mandatory arguments, in this order:
+#
+# 1. An array or hash the function will iterate over.
+# 2. A lambda, which the function calls for each element in the first argument. It takes
+# two mandatory parameters:
+#   1. A memo value that is overwritten after each iteration with the iteration's result.
+#   2. A second value that is overwritten after each iteration with the next value in the
+# function's first argument.
```
nfagerlund: This list item nesting won't work with our MD processor — best to follow the guidelines from [our confluence thingy](https://confluence.puppetlabs.com/display/DOCS/Markdown+Tips+for+Puppet+Labs+Stuff#MarkdownTipsforPuppetLabsStuff-ComplicatedListItems). For this one, that means indenting every line in the nested items by four spaces.

lib/puppet/functions/reduce.rb
nfagerlund: (This is actually resolved now; ignore.)

nfagerlund:
One last formatting glitch that needs fixing, and then we're good to go.

lib/puppet/functions/reduce.rb
```diff
@@ -1,61 +1,98 @@
...
+# lets you create a new value or data structure by combining values from the first
+# argument's data structure.
+#
+# This function takes two mandatory arguments, in this order:
+#
+# 1. An array or hash the function will iterate over.
+# 2. A lambda, which the function calls for each element in the first argument. It takes
+# two mandatory parameters:
+#     1. A memo value that is overwritten after each iteration with the iteration's result.
+#     2. A second value that is overwritten after each iteration with the next value in the
```
gguillotte: Indents increased.

nfagerlund: The line below needs indenting too, to match the list item it's a part of.

gguillotte: Done.

nfagerlund:
Cool. I'm gonna merge as soon as the travis build finishes.

PR #4179, opened 2015-08-19
(docs) Revise with function docs.

gguillotte:


nfagerlund:
I'm :+1:  on this.

MikaelSmith:
In what step will the markdown syntax be processed?

nfagerlund:
@MikaelSmith When we build docs for the website for that particular Puppet version. We build the [function reference](http://docs.puppetlabs.com/references/latest/function.html) in a semi-automated way using the text in the function files as a source, and that's where the MD gets turned into HTML. 

If you call `puppet doc --reference function` on the command line, you can get a bunch of unprocessed Markdown in your terminal. We consider that a fine second-class interface, because MD syntax remains pretty readable in its unprocessed state.

PR #4180, opened 2015-08-19
(PUP-5012) Introduce authorization delegate

nwolfe:
This commit introduces an object seam in Puppet::Network::AuthConfig
that will allow Puppet Server to provide a no-op implementation and
effectively disable authorization so it can be handled externally.

---

The approach I took with this was to wholesale rename the existing `Puppet::Network::AuthConfig` class to `Puppet::Network::DefaultAuthProvider`.

Then, create a new AuthConfig class that simply forwards the `check_authorization` call to an instance of authprovider class, which it will instantiate during initialization. The class can be set, but defaults to `DefaultAuthProvider`.

Happy to change class/module names and structuring.

### DIFF
--- a/lib/puppet/network/authconfig.rb
+++ b/lib/puppet/network/authconfig.rb
@@ -3,7 +3,7 @@ require 'puppet/network/http'
 
 module Puppet
   class ConfigurationError < Puppet::Error; end
-  class Network::AuthConfig
+  class Network::DefaultAuthProvider
     attr_accessor :rights
 
     def self.master_url_prefix
@@ -87,6 +87,26 @@ module Puppet
     def initialize(rights=nil)
       @rights = rights || Puppet::Network::Rights.new
       insert_default_acl
+    end
+  end
+
+  class Network::AuthConfig
+    @@authprovider_class = nil
+
+    def self.authprovider_class=(klass)
+      @@authprovider_class = klass
+    end
+
+    def self.authprovider_class
+      @@authprovider_class || Puppet::Network::DefaultAuthProvider
+    end
+
+    def initialize(rights=nil)
+      @authprovider = self.class.authprovider_class.new(rights)
+    end
+
+    def check_authorization(method, path, params)
+      @authprovider.check_authorization(method, path, params)
     end
   end
 end
--- a/spec/unit/network/authconfig_spec.rb
+++ b/spec/unit/network/authconfig_spec.rb
@@ -3,80 +3,80 @@ require 'spec_helper'
 
 require 'puppet/network/authconfig'
 
-describe Puppet::Network::AuthConfig do
+describe Puppet::Network::DefaultAuthProvider do
   before :each do
     Puppet::FileSystem.stubs(:stat).returns stub('stat', :ctime => :now)
     Time.stubs(:now).returns Time.now
 
-    Puppet::Network::AuthConfig.any_instance.stubs(:exists?).returns(true)
-    # FIXME @authconfig = Puppet::Network::AuthConfig.new("dummy")
+    Puppet::Network::DefaultAuthProvider.any_instance.stubs(:exists?).returns(true)
+    # FIXME @authprovider = Puppet::Network::DefaultAuthProvider.new("dummy")
   end
 
   describe "when initializing" do
     it "inserts default ACLs after setting initial rights" do
-      Puppet::Network::AuthConfig.any_instance.expects(:insert_default_acl)
-      Puppet::Network::AuthConfig.new
+      Puppet::Network::DefaultAuthProvider.any_instance.expects(:insert_default_acl)
+      Puppet::Network::DefaultAuthProvider.new
     end
   end
 
   describe "when defining an acl with mk_acl" do
     before :each do
-      Puppet::Network::AuthConfig.any_instance.stubs(:insert_default_acl)
-      @authconfig = Puppet::Network::AuthConfig.new
+      Puppet::Network::DefaultAuthProvider.any_instance.stubs(:insert_default_acl)
+      @authprovider = Puppet::Network::DefaultAuthProvider.new
     end
 
     it "should create a new right for each default acl" do
-      @authconfig.mk_acl(:acl => '/')
-      expect(@authconfig.rights['/']).to be
+      @authprovider.mk_acl(:acl => '/')
+      expect(@authprovider.rights['/']).to be
     end
 
     it "allows everyone for each default right" do
-      @authconfig.mk_acl(:acl => '/')
-      expect(@authconfig.rights['/']).to be_globalallow
+      @authprovider.mk_acl(:acl => '/')
+      expect(@authprovider.rights['/']).to be_globalallow
     end
 
     it "accepts an argument to restrict the method" do
-      @authconfig.mk_acl(:acl => '/', :method => :find)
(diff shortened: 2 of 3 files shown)


nwolfe:
Ping @joshcooper @camlow325 for review.

spec/unit/network/authconfig_spec.rb
```diff
@@ -107,3 +107,29 @@
...
+  end
+
+  it "delegates authorization check to authprovider_class" do
+    Puppet::Network::AuthConfig.authprovider_class = TestAuthProvider
+    TestAuthProvider.any_instance.expects(:check_authorization).with(:save, '/path/to/resource', {})
+    described_class.new.check_authorization(:save, '/path/to/resource', {})
+  end
+
+  it "uses DefaultAuthProvider by default" do
+    Puppet::Network::AuthConfig.authprovider_class = nil
```
nwolfe: Ugh...I don't like that I have to explicitly "reset" the `authprovider_class` class variable between tests, but I'm not sure how to do make rspec handle this (if it can).

This is why I wrote `AuthConfig` the way I did with the `@@authprovider_class = nil` and then the getter that returns `@@authprovider_class || DefaultAuthProvider`.

Ideally we'd have `@@authprovider_class = DefaultAuthProvider` but then I don't know how to write this test that ensures this default case.

joshcooper: you can reset the class using:

```
before :all do
  @@authprovider_class = nil
end
```

We do that in a few other places where global state is maintained.

nwolfe: Ha, actually...while writing my comment above I realized I could probably reorder the tests so that I ensure the default class is used first before testing that I can set the class.

Although, one advantage of the way `AuthConfig` is now is that it's impossible to have a `nil` authprovider accidentally.

Happy to try to reorder these tests and them change/simplify AuthConfig to have `@@authprovider_class = DefaultAuthProvider`.

nwolfe: > you can reset the class using:
>
> >before :all do
> > @@authprovider_class = nil
> > end

I was referring to the fact that I had to set it to nil anywhere, not so much where/how.

nwolfe: > Happy to try to reorder these tests and them change/simplify AuthConfig to have @@authprovider_class = DefaultAuthProvider.

I tried this out and it works.  There's less code.  I just moved the default test up to the top, and got rid of the setter call.  The AuthConfig would then look like:
```ruby
  class Network::AuthConfig
    @@authprovider_class = Puppet::Network::DefaultAuthProvider

    def self.authprovider_class=(klass)
      @@authprovider_class = klass
    end

    def initialize(rights=nil)
      if rights.nil?
        @authprovider = @@authprovider_class.new
      else
        @authprovider = @@authprovider_class.new(rights)
      end
    end

    def check_authorization(method, path, params)
      @authprovider.check_authorization(method, path, params)
    end
  end
```

One difference with this implementation is that it's now possible to set the authprovider_class to nil and things will blow up.  Like I said, the implementation in this PR at least protects against accidental nils by moving the default value into the getter.

Happy to go either way, not sure which one is better?

nwolfe: > Happy to go either way, not sure which one is better?

Actually I like the current implementation better than the one above in my comment.  Going to keep `AuthConfig` the way it is, though I will reorder the tests to assert the default first as it reads better and avoids the odd `= nil`.

joshcooper: Wasn't sure if you were implying this, but just in case, I wouldn't rely on the order of tests in the rspec group, e.g. if you run with rspec --order:random it can reorder tests, so "instantiates authprovider_class with rights" could come after "delegates authorization check to authprovider_class".

nwolfe: Ah damnit, you're right, forgot about that.  Ok, I guess each test will need to set it properly then.  Update coming.

nwolfe: Turns out I never changed the order in the first place :smile: 

Ready for more review.

lib/puppet/network/authconfig.rb
```diff
@@ -89,4 +89,28 @@ def initialize(rights=nil)
...
+    def self.authprovider_class=(klass)
+      @@authprovider_class = klass
+    end
+
+    def self.authprovider_class
+      @@authprovider_class || Puppet::Network::DefaultAuthProvider
+    end
+
+    def initialize(rights=nil)
+      if rights.nil?
```
camlow325: Do we need the `if` check here or can we just do `@authprovider = self.class.authprovider_class.new(rights)`?

spec/unit/network/authconfig_spec.rb
camlow325: Should we have something like an...

~~~~ruby
after :all do
  @@authprovider_class = nil
end
~~~~

... to avoid the possibility of the provider being set to a "test" one causing other tests outside of this block that happen to use the `Puppet::Network::AuthConfig` to fail when the default one isn't used?

camlow325:
:+1: other than my minor comments...  @joshcooper - what's your take on this PR overall?

joshcooper:
@camlow325 I agree with your comments, once that's resolved I'm :+1:

spec/unit/network/authconfig_spec.rb
nwolfe: Yep, good idea, I'll add that.

lib/puppet/network/authconfig.rb
nwolfe: We need the if because `someclass.new(nil)` is not the same as `someclass.new()`.

This way it allows for the authprovider class to use a default value for the initialize parameter.
Without the if we'd just be passing in nil and not allowing the class to use their own default.

nwolfe:
@joshcooper @camlow325 comments addressed.

I just squashed the addition of `after :all` into the single commit.

lib/puppet/network/authconfig.rb
nwolfe: Discussed further with @camlow325 - let's just pass down `rights` regardless of nil or not and let the implementations of authprovider_class deal with nil accordingly.

nwolfe:
Updated to remove the `rights.nil?` branching (squashed the change in).

Ready for review / merge

spec/unit/network/authconfig_spec.rb
```diff
@@ -107,3 +107,33 @@
     end
   end
 end
+
+describe Puppet::Network::AuthConfig do
+  after :all do
```
joshcooper: I think you want `after :each` otherwise state from one example can leak into another example

nwolfe: I went with `:all` since each test sets it explicitly as part of the test setup.

Sure, I can change it though.

nwolfe:
@joshcooper Updated to `after :all`.

@camlow325 Ready for merge

camlow325:
:+1: @joshcooper - any last comments from you?  If not, I'm going to go and merge this.

joshcooper:
:+1: 

One thing I wanted to mention is that this change makes it possible to disable authorization from any code in the ruby process, e.g. a module function, since we don't distinguish between trusted code on the server and extension code we load into our process. But who am I kidding, this is ruby and the land of monkey patching, so I'm sure this was possible before.

PR #4182, opened 2015-08-20
(PUP-4476) Add basic `puppet lookup` command

HAIL9000:
This pull request adds basically functionality to the `puppet lookup` command.
Right now the command basically serves as a conduit to the Puppet::Pops::Lookup.lookup
function and allows you to pass in parameters to that function. It does not yet support
the --explain or the --facts options.

### DIFF
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -40,6 +40,9 @@ class Puppet::Application::Lookup < Puppet::Application
     options[:default_value] = arg
   end
 
+  # not yet supported
+  option('--trusted')
+
   # Options for facts/scope
   option('--node NODE_NAME') do |arg|
     options[:node] = arg
@@ -65,14 +68,11 @@ class Puppet::Application::Lookup < Puppet::Application
       raise "No node was given via the '--node' flag for the scope of the lookup."
     end
 
-    if (options[:sort_merge_arrays] || options[:merge_hash_arrays] || options[:prefix] || options[:upack_arrays]) && options[:merge] != 'deep'
+    if (options[:sort_merge_arrays] || options[:merge_hash_arrays] || options[:prefix] || options[:unpack_arrays]) && options[:merge] != 'deep'
       raise "The options #{DEEP_MERGE_OPTIONS} are only available with '--merge deep'\n#{RUNHELP}"
     end
 
-    node = Puppet::Node.indirection.find("#{options[:node]}")
-    compiler = Puppet::Parser::Compiler.new(node)
-    compiler.compile
-    scope = compiler.topscope
+    scope = generate_scope
 
     use_default_value = !options[:default_value].nil?
     merge_options = nil
@@ -96,8 +96,13 @@ class Puppet::Application::Lookup < Puppet::Application
       end
     end
 
-    value = Puppet::Pops::Lookup.lookup(scope, options[:keys], options[:type], options[:default_value], use_default_value, {}, {}, merge_options)
+    puts Puppet::Pops::Lookup.lookup(scope, options[:keys], options[:type], options[:default_value], use_default_value, {}, {}, merge_options)
+  end
 
-    puts value
+  def generate_scope
+    node = Puppet::Node.indirection.find("#{options[:node]}")
+    compiler = Puppet::Parser::Compiler.new(node)
+    compiler.compile
+    compiler.topscope
   end
 end
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -1,4 +1,7 @@
 require 'puppet/application'
+require 'puppet/pops/lookup'
+require 'puppet/node'
+require 'puppet/parser/compiler'
 
 class Puppet::Application::Lookup < Puppet::Application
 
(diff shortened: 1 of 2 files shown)


HAIL9000:
Hey @thallgren, if you get a chance could you take a quick look at this and see if you see anything obviously wrong? I'm planning to add some basic unit tests for this, but I wanted to put it up here in case you spotted anything that needs to be changed.

lib/puppet/application/lookup.rb
```diff
@@ -49,8 +57,44 @@ class Puppet::Application::Lookup < Puppet::Application
...
+          merge_options.merge({'unpack_arrays' => options[:unpack_arrays]})
+        end
+
+      else
+        merge_options = {'strategy' => options[:merge]}
+      end
+
+      value = Puppet::Pops::Lookup.lookup(scope, options[:keys], options[:type], options[:default_value], use_default_value, {}, {}, merge_options)
+    else
+      value = Puppet::Pops::Lookup.lookup(scope, options[:keys], options[:type], options[:default_value], use_default_value, {}, {}, nil)
```
thallgren: This long statement can be moved out of the containing if/else instead of being repeated twice. Just always pass merge_options but set it to nil before the if/else.

lib/puppet/application/lookup.rb
```diff
@@ -49,8 +57,44 @@ class Puppet::Application::Lookup < Puppet::Application
...
-  end
+    if options[:keys].empty?
+     raise "No keys were given to lookup."
+    end
+
+    if (options[:sort_merge_arrays] || options[:merge_hash_arrays] || options[:prefix] || options[:upack_arrays]) && options[:merge] != 'deep'
+      raise "The options #{DEEP_MERGE_OPTIONS} are only available with '--merge deep'\n#{RUNHELP}"
+    end
+
+    node = Puppet::Node.indirection.find("#{options[:node]}")
```
thallgren: What happens if the node doesn't exist?

HAIL9000: Even when nil is passed in here we still seem to end up with a compiler and scope that work, at least in my environment. I was running on a master and it successfully found the data I had in the modules in my production environment. However I don't really know WHY this works or how... so I think in the mean time I'll just change this to error out if you don't specify a node.

HAIL9000:
I'm not sure if I feel like all these tests are meaningful/necessary, especially since some of them are very likely to break as development on the lookup subcommand continues. I'm fine with removing some of them.

lib/puppet/application/lookup.rb
```diff
@@ -49,8 +60,49 @@ class Puppet::Application::Lookup < Puppet::Application
...
+      end
+    end
+
+    puts Puppet::Pops::Lookup.lookup(scope, options[:keys], options[:type], options[:default_value], use_default_value, {}, {}, merge_options)
   end
 
+  def generate_scope
+    node = Puppet::Node.indirection.find("#{options[:node]}")
+    compiler = Puppet::Parser::Compiler.new(node)
+    compiler.compile
```
henrik-lindberg: This will produce the entire catalog - need to find a faster approach to get the top scope populated with facts.

thallgren: But what about all global variables declared manifests? IIRC, they are included in the top scope. And how will it find "data in modules" unless the configuration is parsed and metadata is loaded from all modules together with all potential "function" (and soon "hiera") providers?

hlindberg: It will need to do a lot, but not sure it needs to actually evaluate all of the code, produce the catalog etc. While it is possible to create top scope variables when loading code other than the site.pp (or directory representing site.pp), and/or node classifier. I can imagine adding a flag --no-compile that could speed up the lookup by not loading any manifests (i.e. when your data-lookup only depends on the facts).

hlindberg: There are other problems with the approach of compiling the catalog to lookup. Say your compilation fails because you have some problems with your data configuration and you want to use lookup with -explain to understand what is going on. If a compile is required it will fail, and you will not be able to lookup. While being useful to do the compile it is also somewhat different from how it will actually work at runtime since it will first compile everything, and then perform the lookup. In practice the lookups while take place in the middle of the evaluation somewhere (where some of the global variables may not yet have been evaluated). Also, when using -explain, if compiling the entire catalog, there will be many many calls to lookup things - should all of them be explained? 

Also, regarding speed, it is not reasonable to perform a lengthy compilation (a real case could take many minutes) to just lookup a value. 

For the default case we do need the compiler, and that the infrastructure is initialized (the environment instance, injectors etc. and that calls have been made to set_node_parameters, and create_settings_scope. For other variables than facts and settings it would be better if the user had to provide them directly (i.e as overrides) on the command line. This way you can test "what will be looked up if these variables are set this way....", then you know your data configuration is correct.

thallgren: Don't see why all lookups must be explained during compile just because the lookup initiated from the command wants that. The order to explain will be per lookup (and passed down in the context of that lookup). It will not be a global thing.

Regarding compile or not compile, where's the middle ground? Should all configurations be parsed, and then all ruby code loaded for provider functions? I'm not sure the separation between configuration, parsing, and evaluation is good enough to make that an easy task. I haven't investigated in depth though.

hlindberg: I don't think that is very difficult - there are very concrete steps taken in the "compile" method. I imagine that the faster way would only do the "initial import" - that would only do the evaluation of the site.pp top scope, it would load/parse all included classes and register them, but would not completely evaluate everything (thus, it would not drill down into recursive includes). We can probably also prevent the initial-import from taking place, which means there is a top scope with only facts (node parameters), and settings (in the settings scope) available. With both these setups, the infrastructure (binder, loaders, compiler, top-scope, etc.) are all initialized and possible to use - i.e. it is possible to use puppet functions for modules in data etc.

The same effect is obtained by doing a compile with an empty site.pp; this is perhaps a more appealing option - simply change the "manifest" setting for the compile. Remember that in essence, the lookup command is a shortcut equivalence to writing a .pp file with a lookup call in it and then using apply.

hlindberg:
Will merge this and add a ticket to make compilation more efficient.

PR #4183, opened 2015-08-20
(PUP-5074) Support consume/produces annotations site-wide

thallgren:
This commit ensures that the resolution of declared capability
mappings is deferred until all files for a particular environment
has been parsed.

The commit introduces a new Resource::Type named 'capability_mapping'
that acts as intermediate object created by the parser and added to a
hash in the known_resource_types TypeCollection which is later made
available to the compiler. The compiler uses these objects when it's
time to evaluate to perform the actual mapping between capabilities
and resources. Once the mapping is complete, the 'capability_mapping'
hash is no longer needed and is cleared out to conserve memory.

### DIFF
--- a/lib/puppet/parser/ast/pops_bridge.rb
+++ b/lib/puppet/parser/ast/pops_bridge.rb
@@ -73,9 +73,6 @@ class Puppet::Parser::AST::PopsBridge
     def initialize(program_model, context = {})
       @program_model = program_model
       @context = context
-      # @todo lutter 2015-03-10: we maintain a list of definitions we
-      # instantiated so that we can attach CapabilityMappings to them
-      @types = {}
       @ast_transformer ||= Puppet::Pops::Model::AstTransformer.new(@context[:file])
       @@evaluator ||= Puppet::Pops::Parser::EvaluatingParser.new()
     end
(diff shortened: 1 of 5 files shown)


thallgren:
The travis build fails with:

    	(APPMGMT-834) Support consume/produces annotations site-wide
    The commit summary (i.e. the first line of the commit message) should start with one of:    
	(pup-<digits>) # this is most common and should be a ticket at tickets.puppetlabs.com
    	(docs)
    	(maint)
    	(packaging)
    This test for the commit summary is case-insensitive.

What gives?

kylog:
We're enforcing commit summaries referencing PUP tickets, which helps with our release process, and release note generation. Can this ticket be moved to PUP and the commit summary amended?

lib/puppet/parser/ast/pops_bridge.rb
```diff
@@ -184,32 +181,23 @@ def instantiate_HostClassDefinition(o, modname)
...
-      end
-      nil
+      # Use an intermediate 'capability_mapping' type to pass this info to the compiler where the
+      # actual mapping takes place.
+      Puppet::Resource::Type.new(:capability_mapping, "#{o.resource} #{o.kind} #{o.capability}", { :arguments => {
+        'resource_name' => o.resource,
+        'kind'          => o.kind,
+        'blueprint'     => {
+          :capability => o.capability,
+          :mappings   => o.mappings.inject({}) do |memo, mapping|
```
hlindberg: picky: prefer map/reduce family of methods over select/inject methods

PR #4184, opened 2015-08-20
(PUP-5055) parent attributes are set from metadata too early in static_compiler

tdevelioglu:
In the static compiler we are setting the mode, owner and group
attributes on the parent too early. This is causing child
resources to inherit the same values.

Child resources should only inherit attribute values if they are
explicitly set on the parent resource. In other cases overriding
the child group, mode and owner attributes should depend on the
value of source_permissions.

This commit attempts to fix this by postponing the call to set the
parent's attributes until after all children have inherited.
It also prevents mode, owner and group from being set if
source_permissions is not set to 'use'.

### DIFF
--- a/lib/puppet/indirector/catalog/static_compiler.rb
+++ b/lib/puppet/indirector/catalog/static_compiler.rb
@@ -116,8 +116,6 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
   # @param resource [Puppet::Resource]
   # @param file [Puppet::Type::File] The file RAL associated with the resource
   def add_children(request, catalog, resource, file)
-    file = resource.to_ral
-
     children = get_child_resources(request, catalog, resource, file)
 
     remove_existing_resources(children, catalog)
@@ -161,10 +159,11 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
       end
     end
 
+    parent_meta = nil
     total.each do |meta|
       # This is the top-level parent directory
       if meta.relative_path == "."
-        replace_metadata(request, resource, meta)
+        parent_meta = meta
         next
       end
       child = children[meta.relative_path] ||=
@@ -180,6 +179,7 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
       end
       replace_metadata(request, child, meta)
     end
+    replace_metadata(request, resource, parent_meta)
 
     children
   end


lib/puppet/indirector/catalog/static_compiler.rb
```diff
@@ -91,8 +91,11 @@ def find_and_replace_metadata(request, resource, file)
   # @param resource [Puppet::Resource] The resource to add the metadata to
   # @param metadata [Puppet::FileServing::Metadata] The metadata of the given fileserver based file
   def replace_metadata(request, resource, metadata)
-    [:mode, :owner, :group].each do |param|
-      resource[param] ||= metadata.send(param)
+    file = resource.to_ral
+    if file.parameter(:source_permissions).value == :use
```
joshcooper: Sadly the source_permissions logic is a bit of mess. The client sends the `source_permissions` as an HTTP query parameter to the fileserver, and the fileserver either stat's the files or provides default values based on the current process (unix) or sane defaults (windows). The fileserver copies the values into the catalog, which means the agent can't differentiate between "it was set explicitly in the manifest" vs "the fileserver injected values from files on disk" vs "the fileserver injected default values". Then when the agent applies the catalog, it sometimes ignores the metadata, e.g. we ignore owner/group when not root, we skip certain behavior on Windows, etc.

I think it makes sense in an ideal world to only inline metadata when `source_permissions` is `use` or `use_when_creating` (and when the parameter is not explicitly managed in the manifest). But it would mean a statically compiled catalog would have a different effect when applied than a "normal" catalog. That is, when applying a static catalog where `source_permissions => use`, we would not inline metadata for unmanaged parameters, so when a root agent applies the catalog, it would rely on the process umask. But when the agent applies a regular catalog, it would query the server for metadata, and copy the source permissions to the local file, even for permissions that were originally unmanaged.

I think the behavior you are proposing is more correct (with the caveat we should handle `use_when_creating` also), but I'm concerned about having different behavior in static vs non-static catalogs.

lib/puppet/indirector/catalog/static_compiler.rb
```diff
@@ -178,6 +182,7 @@ def get_child_resources(request, catalog, resource, file)
       end
       replace_metadata(request, children[meta.relative_path], meta)
     end
+    replace_metadata(request, resource, parent_meta)
```
joshcooper: Makes sense, can `parent_meta` ever be nil?

tdevelioglu: As far as I can tell File.perform_recursion() always includes the top level (.relative_path = '.'), so technically there should always a point in the loop where parent_meta is set.

lib/puppet/indirector/catalog/static_compiler.rb
tdevelioglu: The File type seems to behave like this (and it seems like we could save on some metadata retrievals here):

**use**
  Client requests metadata through indirection (Puppet::FileServing::Metadata.indirection.find) and copies the values to the resource.

**ignore**
    Client requests metadata through indirection (Puppet::FileServing::Metadata.indirection.find) (why?) but does nothing with it.

**use_when_creating**
    Client requests metadata from through indirection (Puppet::FileServing::Metadata.indirection.find) and copies the value to the resource if it doesn't exist.

The current behavior of the static compiler is already inconsistent because it sets mode, group, owner from metadata regardless of source_permissions.

This change makes the static compiler consistent with non-static, with the only caveat being that the catalog is no longer static when :use_when_creating is set, and the client will have to fetch metadata from the server (because we can't tell if the file exists on the client).

tdevelioglu: I didn't account for checksum, so that would explain why the metadata is requested in all cases.

tdevelioglu:
Looks like this causes another issue with the following case:
```
  concat { '/tmp/concatenated.txt':
      owner => 'root',
      group => 'root',
      mode  => '0644',
    }
    concat::fragment { 'concatenated_test_header':
      target => '/tmp/concatenated.txt',
      source => 'puppet:///modules/role/f5/opt/test.txt',
      order  => 01,
    }
```

2015-08-21 16:27:13 +0200 Puppet (err): Parameter alias failed on File[/var/lib/puppet/concat/_tmp_concatenated.txt/fragments/01_concatenated_test_header]: /var/lib/puppet/concat/_tmp_concatenated.txt/fragments/01_concatenated_test_header can not create alias concat_fragment_concatenated_test_header: object already exists at /etc/puppet/environments/production/modules/concat/manifests/fragment.pp:132
/usr/lib/ruby/vendor_ruby/puppet/util/errors.rb:106:in `fail'
/usr/lib/ruby/vendor_ruby/puppet/type.rb:1403:in `block (3 levels) in <class:Type>'
/usr/lib/ruby/vendor_ruby/puppet/type.rb:1400:in `each'
/usr/lib/ruby/vendor_ruby/puppet/type.rb:1400:in `block (2 levels) in <class:Type>'
/usr/lib/ruby/vendor_ruby/puppet/parameter.rb:432:in `munge'
/usr/lib/ruby/vendor_ruby/puppet/parameter.rb:502:in `value='
/usr/lib/ruby/vendor_ruby/puppet/type.rb:654:in `[]='
/usr/lib/ruby/vendor_ruby/puppet/type.rb:2272:in `block in set_parameters'
/usr/lib/ruby/vendor_ruby/puppet/type.rb:2266:in `each'
/usr/lib/ruby/vendor_ruby/puppet/type.rb:2266:in `set_parameters'
/usr/lib/ruby/vendor_ruby/puppet/type.rb:2222:in `initialize'
/usr/lib/ruby/vendor_ruby/puppet/type/file.rb:481:in `initialize'
/usr/lib/ruby/vendor_ruby/puppet/resource.rb:349:in `new'
/usr/lib/ruby/vendor_ruby/puppet/resource.rb:349:in `to_ral'
/home/tdevelioglu/git/forks/puppet/lib/puppet/indirector/catalog/static_compiler.rb:95:in `replace_metadata'
/home/tdevelioglu/git/forks/puppet/lib/puppet/indirector/catalog/static_compiler.rb:82:in `find_and_replace_metadata'
/home/tdevelioglu/git/forks/puppet/lib/puppet/indirector/catalog/static_compiler.rb:54:in `block in find'
/home/tdevelioglu/git/forks/puppet/lib/puppet/indirector/catalog/static_compiler.rb:43:in `each'
/home/tdevelioglu/git/forks/puppet/lib/puppet/indirector/catalog/static_compiler.rb:43:in `find'
/usr/lib/ruby/vendor_ruby/puppet/indirector/indirection.rb:201:in `find'
/usr/lib/ruby/vendor_ruby/puppet/network/http/api/v1.rb:105:in `do_find'
/usr/lib/ruby/vendor_ruby/puppet/network/http/api/v1.rb:50:in `block in call'
/usr/lib/ruby/vendor_ruby/puppet/context.rb:64:in `override'
/usr/lib/ruby/vendor_ruby/puppet.rb:244:in `override'
/usr/lib/ruby/vendor_ruby/puppet/network/http/api/v1.rb:49:in `call'
/usr/lib/ruby/vendor_ruby/puppet/network/http/route.rb:82:in `block in process'
/usr/lib/ruby/vendor_ruby/puppet/network/http/route.rb:81:in `each'
/usr/lib/ruby/vendor_ruby/puppet/network/http/route.rb:81:in `process'
/usr/lib/ruby/vendor_ruby/puppet/network/http/handler.rb:63:in `block in process'
/usr/lib/ruby/vendor_ruby/puppet/util/profiler/around_profiler.rb:58:in `profile'
/usr/lib/ruby/vendor_ruby/puppet/util/profiler.rb:51:in `profile'
/usr/lib/ruby/vendor_ruby/puppet/network/http/handler.rb:61:in `process'
/usr/lib/ruby/vendor_ruby/puppet/network/http/webrick/rest.rb:37:in `block in service'
/usr/lib/ruby/vendor_ruby/puppet/network/http/webrick/rest.rb:36:in `synchronize'
/usr/lib/ruby/vendor_ruby/puppet/network/http/webrick/rest.rb:36:in `service'
/usr/lib/ruby/2.1.0/webrick/httpserver.rb:138:in `service'
/usr/lib/ruby/2.1.0/webrick/httpserver.rb:94:in `run'
/usr/lib/ruby/vendor_ruby/puppet/network/http/webrick.rb:33:in `block (2 levels) in listen'
/usr/lib/ruby/2.1.0/webrick/server.rb:295:in `call'
/usr/lib/ruby/2.1.0/webrick/server.rb:295:in `block in start_thread'

branan:
We need to have further design/discussion around the `source_permission` logic. If you pull that out and narrow the scope of this to just fixing the way permissions are propagated to child resources it'd be :+1: for merge.

It also looks like this needs a rebase

tdevelioglu:
@branan 
done and done  :+1:

PR #4185, opened 2015-08-21
(PUP-5062) static compiler fails to remove existing resources from children

tdevelioglu:
The static compiler never removes child resources that already exist
leading to duplicate file resource declarations.

The problem is in remove_existing_resources()
The intersection between existing_names and children.keys never matches
anything because existing_names are stringified resource objects (e.g.
File['/tmp/statically_compiled/test.txt']) and children.keys are
relative path strings (e.g. 'test.txt.).

This leads to duplicate files resources and catalog subsequently failing
with an error:
`Error: Could not retrieve catalog from remote server: Error 400 on
SERVER: Duplicate declaration: File[/tmp/statically_compiled/test.txt]
is already declared in file
/etc/puppet/environments/production/modules/role/manifests/traveler.pp:27;
cannot redeclare`

This commit fixes removes_child_resources() by ensuring we get the
intersection of stringified resources from both catalog and children.
It also cleans up both remove_existing_resources and get_child_resources a bit.

### DIFF
--- a/lib/puppet/indirector/catalog/static_compiler.rb
+++ b/lib/puppet/indirector/catalog/static_compiler.rb
@@ -166,31 +166,36 @@ class Puppet::Resource::Catalog::StaticCompiler < Puppet::Resource::Catalog::Com
         replace_metadata(request, resource, meta)
         next
       end
-      children[meta.relative_path] ||= Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
+      child = children[meta.relative_path] ||=
+        Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
 
       # I think this is safe since it's a URL, not an actual file
-      children[meta.relative_path][:source] = source + "/" + meta.relative_path
+      child[:source] = source + "/" + meta.relative_path
       resource.each do |param, value|
         # These should never be passed to our children.
         unless [:parent, :ensure, :recurse, :recurselimit, :target, :alias, :source].include? param
-          children[meta.relative_path][param] = value
+          child[param] = value
         end
       end
-      replace_metadata(request, children[meta.relative_path], meta)
+      replace_metadata(request, child, meta)
     end
 
     children
   end
 
-  # Remove any file resources in the catalog that will be duplicated by the
-  # given file resources.
+  # Remove any recursed file resources already in the catalog
   #
   # @param children [Array<Puppet::Resource>]
   # @param catalog [Puppet::Resource::Catalog]
   def remove_existing_resources(children, catalog)
-    existing_names = catalog.resources.collect { |r| r.to_s }
-    both = (existing_names & children.keys).inject({}) { |hash, name| hash[name] = true; hash }
-    both.each { |name| children.delete(name) }
+    relative_paths = children.keys
+    relative_paths.each do |relative_path|
+      child = children[relative_path]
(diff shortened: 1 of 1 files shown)


tdevelioglu:
It properly removes pre-existing resources from the list of children, but I bumped into another issue with resource aliases now:
```
  concat { '/tmp/concatenated.txt':
      owner => 'root',
      group => 'root',
      mode  => '0644',
    }
    concat::fragment { 'concatenated_test_header':
      target => '/tmp/concatenated.txt',
      source => 'puppet:///modules/role/f5/opt/test.txt',
      order  => 01,
    }
```

Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Parameter alias failed on File[/var/lib/puppet/concat/_tmp_concatenated.txt/fragments/01_concatenated_test_header]: /var/lib/puppet/concat/_tmp_concatenated.txt/fragments/01_concatenated_test_header can not create alias concat_fragment_concatenated_test_header: object already exists at /etc/puppet/environments/production/modules/concat/manifests/fragment.pp:132

lib/puppet/indirector/catalog/static_compiler.rb
```diff
@@ -188,9 +189,16 @@ def get_child_resources(request, catalog, resource, file)
...
-    existing_names = catalog.resources.collect { |r| r.to_s }
-    both = (existing_names & children.keys).inject({}) { |hash, name| hash[name] = true; hash }
-    both.each { |name| children.delete(name) }
+    # build a list of stringified file resources from the children
+    child_names = children.collect { |k, r| r.to_s }
+    # build a list of stringified file resources from the catalog
+    existing_names = catalog.resources.inject([]) do |list, res|
+      list << res.to_s if res.type == 'File'
+      list
+    end
```
joshcooper: Rather than iterate over every resource in the catalog (which could be large) and built up an array of existing file resource names, we could iterate over the recursed children, and check to see if it's already in the catalog, something like:

```
    relative_paths = children.keys
    relative_paths.each do |relative_path|
      child = children[relative_path]
      if catalog.resource(child.ref)
        Puppet.debug("Resource #{child.ref} already managed, removing from recursed children")
        children.delete(relative_path)
      end
    end
```

Ideally we could just iterate over `children`, but I don't think it's safe to mutate the Array while it's being iterated over.

joshcooper:
@tdevelioglu thank you for all of these static compiler PRs, we really appreciate your efforts here!

lib/puppet/indirector/catalog/static_compiler.rb
tdevelioglu: modified after your suggestion, thanks!

joshcooper:
If you explicitly manage a subdirectory with `recurse=>false` should we also remove its children? For example,

```
file { '/tmp/foo':
  ensure => directory,
  recurse => true,
  source => 'puppet:///modules/mymodule/foo',
}
file { '/tmp/foo/bar':
  ensure => directory,
  recurse => false,
  source => 'puppet:///modules/mymodule/foo/bar',
}
```

I would think the resulting catalog would not contain the children of `bar`. But I'm also curious how the normal compiler behaves (once the agent does `eval_generate` on the parent `/tmp/foo`)?

branan:
We're going to go ahead and merge this and start a discussion internally around behavior of nested `recurse` definitions

PR #4191, opened 2015-08-25
(PUP-5056) launchd service provider tests

MikaelSmith:
Launchd had a few bugs in managing services on newer platforms that were
not exposed by testing. Add acceptance tests to ensure managing services
with the launchd provider are tested.

### DIFF
--- a/acceptance/tests/resource/service/launchd_provider.rb
+++ b/acceptance/tests/resource/service/launchd_provider.rb
@@ -1,0 +1,90 @@
+test_name 'Mac OS X launchd Provider Testing'
+
+confine :to, {:platform => /osx/}, agents
+
+sloth_daemon_script = <<SCRIPT
+#!/usr/bin/env sh
+while true; do sleep 1; done
+SCRIPT
+
+svc = 'com.puppetlabs.sloth'
+launchd_script_path = "/Library/LaunchDaemons/#{svc}.plist"
+
+def assert_service_status_on(host, service, status, expect_running)
+  ensure_status = "ensure => '#{status[:ensure]}'" if status[:ensure]
+  enable_status = "enable => '#{status[:enable]}'" if status[:enable]
+
+  on host, puppet_resource('service', service) do
+    assert_match(/'#{service}'.+#{ensure_status}.+#{enable_status}/m, stdout, "Service status does not match expectation #{status}")
+  end
+
+  on host, 'launchctl list' do
+    if expect_running
+      assert_match(/#{service}/, stdout, 'Service was not found in launchctl list')
+    else
+      assert_no_match(/#{service}/, stdout, 'Service was not expected in launchctl list')
+    end
+  end
+end
+
+def ensure_service_on(host, service, status, expect_running)
+  ensure_status = "ensure => '#{status[:ensure]}'," if status[:ensure]
+  enable_status = "enable => '#{status[:enable]}'," if status[:enable]
+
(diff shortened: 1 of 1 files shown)


acceptance/tests/resource/service/launchd_provider.rb
```diff
@@ -0,0 +1,84 @@
...
+  create_remote_file(agent, launchd_script_path, launchd_script)
+
+  teardown do
+    on agent, "rm #{sloth_daemon_path} #{launchd_script_path}"
+  end
+
+  step "Verify the service exists on #{agent}"
+  assert_service_status_on agent, svc, :ensure => 'stopped', :enable => 'true'
+
+  require 'byebug'; byebug
```
joshcooper: Looks like a leftover debug line

MikaelSmith:
Depends on #4192

MikaelSmith:
ping @er0ck

acceptance/tests/resource/service/launchd_provider.rb
```diff
@@ -0,0 +1,82 @@
...
+
+def ensure_service_on(host, service, status)
+  ensure_status = "ensure => '#{status[:ensure]}'," if status[:ensure]
+  enable_status = "enable => '#{status[:enable]}'," if status[:enable]
+
+  apply_manifest_on host, "service { '#{service}': #{ensure_status} #{enable_status} }" do
+    assert_match(/Service\[#{service}\]\/ensure: ensure changed '\w+' to '#{status[:ensure]}'/, stdout,
+                 'Service status change failed') if status[:ensure]
+    assert_match(/Service\[#{service}\]\/enable: enable changed '\w+' to '#{status[:enable]}'/, stdout,
+                 'Service enable change failed') if status[:enable]
```
er0ck: LGTM.  do you not want to ensure the launchd provider is used in the puppet manifests?

acceptance/tests/resource/service/launchd_provider.rb
```diff
@@ -0,0 +1,82 @@
+test_name 'Mac OS X launchd Provider Testing'
+ 
+confine :to, {:platform => /osx/}, agents
```
er0ck: combining platform and agents here might not do what we think.  i haven't tried it yet...

MikaelSmith: I've done this other places, it seems to filter the agents list, and if agents is empty it will skip the test.

acceptance/tests/resource/service/launchd_provider.rb
MikaelSmith: Sure, I could add that.

MikaelSmith:
This test is somewhat circular; I think I'll add checks of `launchctl list | grep #{svc}`. I don't know a way to check whether the service is enabled/disabled other than with Puppet (or checking the overrides/disabled plist file). That would definitely be faster than running `puppet service`.

MikaelSmith:
I think the spec tests will now sufficiently cover ensuring the correct plist file is written.

MikaelSmith:
This is ready, and I've verified against 10.10. Running against 10.9 now.

MikaelSmith:
@er0ck get a chance to look at this again?

er0ck:
nice.  passes for me.  
we might want to think about using your assert_status_on() and enable_service_on() for the other service  enable tests... :neutral_face:

er0ck:
FYI the enable step (the last one) is super slow in this test/OSX.  not sure why...

MikaelSmith:
Generally checking status is really slow, which another PR to use cfpropertylist should speed up.

PR #4192, opened 2015-08-25
(maint) Fix up merge from 3.x

MikaelSmith:
The launchd provider fixes for (PUP-5044) weren't merged up correctly.
Fix the merge. Also merge new specs.


lib/puppet/provider/service/launchd.rb
```diff
@@ -298,13 +317,21 @@ def enabled?
...
+      overrides[resource[:name]] = false
+    end
     Plist::Emit.save_plist(overrides, self.class.launchd_overrides)
   end
 
   def disable
-    job_path, job_plist = plist_from_label(resource[:name])
-    job_plist["Disabled"] = true
-    Plist::Emit.save_plist(job_plist, job_path)
+    overrides = self.class.read_plist(self.class.launchd_overrides)
```
MikaelSmith: @kylog this removal was your work in Puppet 4 cleanup, but it looks to me like it chose the wrong piece to remove. The `enable` method kept the OS X 10.6 code branch (`has_macosx_plist_overrides? == true`) while the `disable` method kept the older `has_macosx_plist_overrides? == false` branch.

MikaelSmith:
There are spec changes that were dropped that would be nice to add back in, but I'll have to resolve a bunch anyway for https://github.com/puppetlabs/puppet/pull/4190 so I'm tempted to do it on that merge-up.

lib/puppet/provider/service/launchd.rb
MikaelSmith: Or @joshcooper, if you happen to be less busy today.

MikaelSmith:
New specs (merged up from 3.x) added to this PR.

spec/unit/provider/service/launchd_spec.rb
```diff
@@ -147,21 +198,45 @@
     end
   end
 
-  describe "when enabling the service" do
+  [[10, "10.6"], [13, "10.9"]].each do |kernel, version|
```
branan: This just needs to check 10.9 in puppet 4, right?

branan: I guess it doesn't hurt, since 10.6 through 10.9 share a code path

MikaelSmith: https://github.com/puppetlabs/puppet/commit/efd2a45646eaedd3aa744ebcf7beb94739653db5 specifically removes support for <= 10.6, so I could change this to test 10.7. I wanted to test the range of what we seem to support.

lib/puppet/provider/service/launchd.rb
MikaelSmith: This matches @keeleysam's changes in https://github.com/puppetlabs/puppet/pull/4164 to master.

spec/unit/provider/service/launchd_spec.rb
MikaelSmith: Yeah, I lean towards not changing it.

joshcooper:
I don't understand why github is showing these other commits....

PR #4193, opened 2015-08-25
(docs) Hiera function revisions

gguillotte:
Revising ~~lookup~~, `hiera`, `hiera_array`, `hiera_hash`, and `hiera_include`.

Squashed; per @nfagerlund, there's more discussion needed about 
terms we use in `lookup`.

### DIFF
--- a/lib/puppet/functions/hiera.rb
+++ b/lib/puppet/functions/hiera.rb
@@ -1,7 +1,80 @@
 require 'hiera/puppet_function'
-
-# @see lib/puppet/parser/functions/hiera.rb for documentation
-# TODO: Move docs here when the format has been determined.
+# Performs a standard priority lookup of the hierarchy and returns the most specific value
+# for a given key. The returned value can be any type of data.
+#
+# The function takes up to three arguments, in this order:
+#
+# 1. A string key that Hiera searches for in the hierarchy. **Required**.
+# 2. An optional default value to return if Hiera doesn't find anything matching the key.
+#     * If this argument isn't provided and this function results in a lookup failure, Puppet
+#     fails with a compilation error.
+# 3. The optional name of an arbitrary
+# [hierarchy level](https://docs.puppetlabs.com/hiera/latest/hierarchy.html) to insert at the
+# top of the hierarchy. This lets you temporarily modify the hierarchy for a single lookup.
+#     * If Hiera doesn't find a matching key in the overriding hierarchy level, it continues
+#     searching the rest of the hierarchy.
+#
+# The `hiera` function does **not** find all matches throughout a hierarchy, instead
+# returining the first specific value starting at the top of the hierarchy. To search
+# throughout a hierarchy, use the `hiera_array` or `hiera_hash` functions.
+#
+# @example Using `hiera`
+#
+# ~~~ yaml
+# # Assuming hiera.yaml
+# # :hierarchy:
+# #   - web01.example.com
+# #   - common
+#
+# # Assuming web01.example.com.yaml:
+# # users:
+# #   - "Amy Barry"
+# #   - "Carrie Douglas"
+#
+# # Assuming common.yaml:
+# users:
+#   admins:
+#     - "Edith Franklin"
+#     - "Ginny Hamilton"
+#   regular:
+#     - "Iris Jackson"
+#     - "Kelly Lambert"
+# ~~~
+#
+# ~~~ puppet
+# # Assuming we are not web01.example.com:
+#
+# $users = hiera('users', undef)
+#
+# # $users contains {admins  => ["Edith Franklin", "Ginny Hamilton"],
+# #                  regular => ["Iris Jackson", "Kelly Lambert"]}
+# ~~~
+#
+# You can optionally generate the default value with a
+# [lambda](https://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html) that
+# takes one parameter.
+#
+# @example Using `hiera` with a lambda
+#
+# ~~~ puppet
+# # Assuming the same Hiera data as the previous example:
+#
+# $users = hiera('users') | $key | { "Key \'${key}\' not found" }
+#
+# # $users contains {admins  => ["Edith Franklin", "Ginny Hamilton"],
+# #                  regular => ["Iris Jackson", "Kelly Lambert"]}
+# # If hiera couldn't match its key, it would return the lambda result,
+# # "Key 'users' not found".
+# ~~~
+#
+# The returned value's data type depends on the types of the results. In the example 
+# above, Hiera matches the 'users' key and returns it as a hash.
+#
+# See
+# [the documentation](https://docs.puppetlabs.com/hiera/latest/puppet.html#hiera-lookup-functions)
+# for more information about Hiera lookup functions.
+#
+# @since 4.0.0
 #
 Puppet::Functions.create_function(:hiera, Hiera::PuppetFunction) do
   init_dispatch
--- a/lib/puppet/parser/functions/hiera.rb
+++ b/lib/puppet/parser/functions/hiera.rb
@@ -1,33 +1,89 @@
 require 'hiera_puppet'
 
 module Puppet::Parser::Functions
-  newfunction(:hiera, :type => :rvalue, :arity => -2, :doc => "Performs a
-  standard priority lookup and returns the most specific value for a given key.
-  The returned value can be data of any type (strings, arrays, or hashes). 
+  newfunction(
+    :hiera,
+    :type => :rvalue,
+    :arity => -2,
+    :doc => <<DOC
+Performs a standard priority lookup of the hierarchy and returns the most specific value
+for a given key. The returned value can be any type of data.
 
-  The function can be called in one of three ways:
-  1. Using 1 to 3 arguments where the arguments are:
-     'key'      [String] Required
-           The key to lookup.
-     'default`  [Any] Optional
-           A value to return when there's no match for `key`. Optional
-     `override` [Any] Optional
-           An argument in the third position, providing a data source
-           to consult for matching values, even if it would not ordinarily be
-           part of the matched hierarchy. If Hiera doesn't find a matching key
-           in the named override data source, it will continue to search through the
-           rest of the hierarchy.
+The function takes up to three arguments, in this order:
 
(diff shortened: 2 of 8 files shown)


kylog:
Ping @hlindberg and @nfagerlund.

gguillotte:
Amended commit to use https links to docs.

hlindberg:
ping @thallgren (can you take a look)

lib/puppet/parser/functions/hiera.rb
```diff
@@ -1,33 +1,88 @@
...
-  <http://docs.puppetlabs.com/hiera/1/puppet.html#hiera-lookup-functions>
-  ") do |*args|
+  newfunction(
+    :hiera,
+    :type => :rvalue,
+    :arity => -2,
+    :doc => <<-DOC
+Performs a standard priority lookup of the hierarchy and returns the most specific value
+for a given key. The returned value can be any type of data, depending on the values'
+types and the number of results.
```
nfagerlund: The number of results doesn't have an effect on a priority lookup; you only ever get the first one.

lib/puppet/parser/functions/hiera.rb
```diff
@@ -1,33 +1,88 @@
...
+
+The function takes up to three arguments, in this order:
+
+1. A string key that Hiera searches for in the hierarchy. **Required**.
+2. An optional default value to return if Hiera doesn't find anything matching the key.
+  * If this argument isn't provided and this function results in a lookup failure, Puppet
+fails with a compilation error.
```
nfagerlund: Same indentation problems as I pointed out in the other PR; full text of this bullet should be indented 4 sp.

lib/puppet/parser/functions/hiera.rb
```diff
@@ -1,33 +1,88 @@
...
+
+$users = hiera('users', undef)
+
+# $users contains {admins => [Edith Franklin, Ginny Hamilton],
+#                 regular => [Iris Jackson, Kelly Lambert]}
+~~~
+
+Optionally, you can use a
+[lambda](https://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html) instead of
+an argument to generate the default value.
```
nfagerlund: It looks like the lambda takes one parameter, which is the key you provided as the function's first argument? We should say so here.

lib/puppet/parser/functions/hiera.rb
```diff
@@ -1,33 +1,88 @@
...
+#   - "Carrie Douglas"
+
+# Assuming common.yaml:
+users:
+  admins:
+    - "Edith Franklin"
+    - "Ginny Hamilton"
+  regular:
+    - "Iris Jackson"
+    - "Kelly Lambert"
```
nfagerlund: Maybe use two code blocks — highlight the first one as yaml, and the second as puppet.

lib/puppet/parser/functions/hiera.rb
```diff
@@ -1,33 +1,88 @@
...
+
+$users = hiera('users', undef)
+
+# $users contains {admins => [Edith Franklin, Ginny Hamilton],
+#                 regular => [Iris Jackson, Kelly Lambert]}
```
nfagerlund: Should specify that we aren't web01.example.com.

lib/puppet/parser/functions/hiera.rb
```diff
@@ -1,33 +1,88 @@
...
+Optionally, you can use a
+[lambda](https://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html) instead of
+an argument to generate the default value.
+
+@example Using `hiera` with a lambda
+
+~~~ puppet
+# Assuming the same Hiera data as the previous example:
+
+$users = hiera('users') | $user | { "Key \'${user}\' not found" }
```
nfagerlund: Using `$user` as the name for the param is confusing; I suggest renaming it `$key` or something.

lib/puppet/parser/functions/hiera_hash.rb
```diff
@@ -1,36 +1,79 @@
...
-  <http://docs.puppetlabs.com/hiera/1/puppet.html#hiera-lookup-functions>
-  ") do |*args|
+  newfunction(
+    :hiera_hash,
+    :type => :rvalue,
+    :arity => -2,
+    :doc => <<-DOC
+Finds all matches of a key throughout the hierarchy and returns them in a merged hash.
+If any of the matched hashes share keys, the function returns the highest-level,
+highest-priority key/value pair. This is called a
```
nfagerlund: I'd revise this sentence as something like:

> If any of the matched hashes share keys, the final hash will use the value from the highest priority match.

lib/puppet/parser/functions/hiera_hash.rb
```diff
@@ -1,36 +1,79 @@
...
+[hash merge lookup](https://docs.puppetlabs.com/hiera/latest/lookup_types.html#hash-merge).
+
```
nfagerlund: Maybe add a short (1-sentence) note about the :merge_behavior setting (https://docs.puppetlabs.com/hiera/latest/configuring.html#mergebehavior) in a new paragraph here.

lib/puppet/parser/functions/hiera_include.rb
```diff
@@ -1,46 +1,82 @@
...
-  More thorough examples of `hiera_include` are available at:
-  <http://docs.puppetlabs.com/hiera/1/puppet.html#hiera-lookup-functions>
-  ") do |*args|
+  newfunction(
+    :hiera_include,
+    :arity => -2,
+    :doc => <<-DOC
+Assigns classes to a node using an
+[array merge lookup](https://docs.puppetlabs.com/hiera/latest/lookup_types.htmlarray-merge)
+that retrieves the value for a user-specified key from a Hiera data source.
```
nfagerlund: "that retrieves all values for a user-specified key from Hiera's data."

lib/puppet/parser/functions/hiera_include.rb
```diff
@@ -1,46 +1,82 @@
...
+~~~ puppet
+# Assuming hiera.yaml
+# :hierarchy:
+#   - web01.example.com
+#   - common
+
+# Assuming web01.example.com.yaml:
+# classes:
+#   - apache
+#   - apache::mod::php
```
nfagerlund: Since we're doing an array merge, we can illustrate it better by adding a common.yaml file and showing that you get a combined result.

lib/puppet/parser/functions/hiera_array.rb
```diff
@@ -1,34 +1,75 @@
...
+# users:
+#   - 'cdouglas = regular'
+#   - 'efranklin = regular'
+
+# Assuming web01.example.com.yaml:
+# users: 'abarry = admin'
+
+$allusers = hiera_array('users', undef)
+
+# $allusers contains [cdouglas = regular, efranklin = regular, abarry = admin].
```
nfagerlund: Should quote these strings.

lib/puppet/parser/functions/hiera_hash.rb
```diff
@@ -1,36 +1,79 @@
...
+# Assuming web01.example.com.yaml:
+# users:
+#   administrators:
+#     'aberry': 'Amy Berry'
+
+$allusers = hiera_hash('users', undef)
+
+# $allusers contains {regular => {cdouglas => Carrie Douglas},
+#                     administrators => {aberry => Amy Berry}}
```
nfagerlund: Quote strings

nfagerlund:
A few of those comments (indentation, variable name, the lambda's argument) apply to all the functions, but I only wrote them in the first one. 

Other than these, I'm good with these updates.

lib/puppet/parser/functions/hiera_hash.rb
gguillotte: Will do on these as well. It's worth noting that the output here is as Puppet displays it in notices when I tested these commands, so I'm assuming that output format is unique to notices?

nfagerlund:
@gguillotte Re: the de-quoting: Not unique to notices, but it's definitely a loss of fidelity. It comes from these rules being applied recursively: http://docs.puppetlabs.com/puppet/4.2/reference/lang_data_string.html#conversion-of-interpolated-values

tbh I don't like that behavior; it's simple and elegant from the implementation side, since it's one set of rules acting recursively, but out here in real life it's unexpected and hairy. :/

lib/puppet/parser/functions/hiera.rb
```diff
@@ -1,33 +1,91 @@
...
+
+# $users contains {admins  => ["Edith Franklin", "Ginny Hamilton"],
+#                  regular => ["Iris Jackson", "Kelly Lambert"]}
+# If hiera couldn't match its key, it would return the lambda result,
+# "Key 'users' not found".
+~~~
+
+The returned value's data type depends on the types of the results. For example, if `hiera`
+finds a single string, it returns it as a string. If it finds multiple strings, as in the
+example above, it returns an array of strings. If it finds one or more hashes, it returns
```
nfagerlund: Ack, I missed this one the first time: like the bit at the top, this should reflect that `hiera` always just returns that first value, and never combines multiple values into a new value.

gguillotte: Revised it.

lib/puppet/functions/hiera.rb
```diff
@@ -1,7 +1,79 @@
-require 'hiera/puppet_function'
```
nfagerlund: This breaks the code in real life! Gotta put this back.

lib/puppet/parser/functions/hiera.rb
```diff
@@ -1,33 +1,89 @@
...
-  3. Like #1 but with all arguments passed in an array.
-
-  More thorough examples of `hiera` are available at:  
-  <http://docs.puppetlabs.com/hiera/1/puppet.html#hiera-lookup-functions>
-  ") do |*args|
+  newfunction(
+    :hiera,
+    :type => :rvalue,
+    :arity => -2,
+    :doc => <<-DOC
```
nfagerlund: Heredocs don't need the "-" character unless you're indenting the closing tag, which you aren't.

gguillotte: D'oh on both cases; fixed, pushed.

nfagerlund:
Now that Travis is passing, I'm gonna merge.

PR #4194, opened 2015-08-25
(packaging) Add condrestart to suse init scripts

shrug:
In the puppet-agent RPM spec we call the 'condrestart' action from
the %postun on package upgrade. This is currently missing from the
sles/suse init scripts.

This adds 'condrestart' as a synonym for the try-restart action.


haus:
:+1:

PR #4196, opened 2015-08-26
(PUP-5088) Ensure only => can be used in capability mappings

thallgren:
This commit relaxes that constraint that a CapabilityMapping
only can contain an AttributeOperation so that it instead can
contain an AbstractAttributeOperation. This was done to prevent
a cryptic error message when an AttributesOperation (note the
plural) was assigned due to use of *=> operator. The relaxed
constraint lets this pass so that it instead can be trapped
in the validator where a correct check was already in place.

### DIFF
--- a/spec/unit/capability_spec.rb
+++ b/spec/unit/capability_spec.rb
@@ -127,6 +127,34 @@ describe "Capability types" do
     expect(cns[:mappings]).to be_instance_of(Hash)
     expect(cns[:mappings]['host']).to be_instance_of(Puppet::Parser::AST::PopsBridge::Expression)
   end
+  it "does not allow operator '+>' in a mapping" do
+    expect do
+    compile_to_catalog(<<-MANIFEST)
+      define test($hostname) {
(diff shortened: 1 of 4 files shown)


spec/unit/capability_spec.rb
```diff
@@ -63,6 +63,34 @@
...
+      compile_to_catalog(<<-MANIFEST)
+      define test($hostname) {
+        notify { "hostname ${hostname}":}
+      }
+
+      Test consumes Cap {
+        *=> $hostname
+      }
+      MANIFEST
+    end.to raise_error(Puppet::ParseError, /The operator '\* =>' in a Block Expression is not supported/)
```
hlindberg: This does not seem right. I think the expression may be derived from BlockExpression and that a label is missing for it?

hlindberg:
needs a rebase

thallgren:
Rebased and fixed syntax in test case.

hlindberg:
+1, just waiting for checks to clear before merging

PR #4197, opened 2015-08-27
(FOR REVIEW) (PUP-5013) move profiling block for resource eval

cprice404:
This PR is for review only, I haven't run tests or anything.  Just trying to solicit input on the change in general.

When surfacing Puppet metrics in Puppet Server, w/o this change, a huge percentage of the metrics related to resource evaluation are never registered / tracked, because there are multiple code paths that can lead to resource evaluation during compilation, and only one of them has a profiling block around it.

e.g. whether or not a resource gets metrics tracked may depend on whether it was introduced via NC vs site.pp, etc.

In local tests where I've been experimenting with improvements for surfacing metrics data from Puppet Server, this change is a huge improvement and gives us much more interesting and complete data to look at.

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -380,9 +380,7 @@ class Puppet::Parser::Compiler
     exceptwrap do
       Puppet::Util::Profiler.profile("Evaluated definitions", [:compiler, :evaluate_definitions]) do
         !unevaluated_resources.each do |resource|
-          Puppet::Util::Profiler.profile("Evaluated resource #{resource}", [:compiler, :evaluate_resource, resource]) do
-            resource.evaluate
-          end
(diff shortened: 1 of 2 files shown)


cprice404:
ping @kylog @hlindberg as requested, here's a PR in hopes that that makes it easier to visualize the change I'm proposing.

hlindberg:
This is all server side and I see no problem with this approach.

kylog:
Agreed this is all compiler, no agent paths.

LGTM. Maybe update the commit message to include the explanatory background in the gh comment above (and remove the WIP).

Also do you want this in a particular 3.8 release? Just curious b/c we're barely doing anything there, but this looks safe so we can certainly merge it there.

kylog:
Closing this in favor of #4202.

PR #4198, opened 2015-08-27
(PUP-5097) Adding hostname and domainname in trusted facts

fihuer:
Before this there is no way to get the hostname/domain fact from a trustworthy source.

With this `hostname` and `domainname` are now part of the trusted facts. Both of them are derived
from the validated certificate name.

It can be used in Hiera as follows :
```yaml
:hierarchy:
  - "%{::trusted.domainname}/%{::trusted.hostname}"
  - "%{::trusted.domainname}"
  - common
```

### DIFF
--- a/lib/puppet/context/trusted_information.rb
+++ b/lib/puppet/context/trusted_information.rb
@@ -17,10 +17,28 @@ class Puppet::Context::TrustedInformation
   # @return [Hash{Object => Object}]
   attr_reader :extensions
 
+  # The domain name derived from the validated certificate name
+  #
+  # @return [String]
+  attr_reader :domainname
+
+  # The hostname derived from the validated certificate name
+  #
+  # @return [String]
+  attr_reader :hostname
+
   def initialize(authenticated, certname, extensions)
     @authenticated = authenticated.freeze
     @certname = certname.freeze
     @extensions = extensions.freeze
+    if @certname
+      hostname, domainname = @certname.split('.', 2)
+    else
+      hostname = nil
+      domainname = nil
+    end
+    @hostname = hostname.freeze
+    @domainname = domainname.freeze
   end
 
   def self.remote(authenticated, node_name, certificate)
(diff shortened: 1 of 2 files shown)


kylog:
@hlindberg or @nfagerlund thoughts on this? Is there an existing pattern in hiera for accomplishing the end goal here without adding to `trusted`?

hlindberg:
essentially this boils down to `hostname, domainname = @certname.split('.', 2)` - which would be difficult to do inside a hiera hierarchy due to lack of transformation capabilities. I can see how this feature is helpful and I can't see it being harmful - extensions are under a separate key so no risk of clashing.

nfagerlund:
I have a small problem, and a big one. 

The small one is that `domainname` is wrong; there's an equivalent Facter fact called `domain`, and their names should match to avoid confusion. (We've been here before with `certname` vs `clientcert`; let's learn from that mistake.)

The big problem is that these names are fundamentally misleading in the first place — `hostname` isn't actually the hostname, and `domain` isn't actually the domain. They're both just transformations of the `certname`, which is specified with a setting and which has no confirmable connection to a machine's DNS name. It defaults to the fqdn, sure, but by calling these transformations "hostname" and "domain" in the $trusted hash, we're giving the illusion that we can _confirm_ the machine's hostname. (And we can't.)

I think the general problem we're trying to solve is: some (most) users encode multiple pieces of information into their machines' certnames with consistent naming conventions, and they'd like to access that information in their Hiera hierarchies. This hardcodes a convenience for the most common convention (certname = fqdn), but what about extracting things like AWS region, etc.? Do we want to support that?

If we decide to bail on the general problem but still provide this convenience, I still want to name the facts something else, so we don't give the illusion that we're doing something impossible. If we're just providing a transformation on the certname, it should LOOK like we're providing a transformation on the certname.

hlindberg:
Lets continue the discussion on the ticket. I agree with @nfagerlund

PR #4200, opened 2015-08-28
(PUP-4821) acceptance test for automatic parameter lookup

shaigy:
Modified existing test to include automatic parameter lookup of
parametrized class


shaigy:
@er0ck As we discussed I modified the existing lookup.rb to include automatic parameter lookup.

er0ck:
looks good!

PR #4202, opened 2015-08-31
(PUP-5013) move profiling block for resource eval

cprice404:
When surfacing Puppet metrics in Puppet Server, w/o this change, a huge percentage of the metrics related to resource evaluation are never registered / tracked, because there are multiple code paths that can lead to resource evaluation during compilation, and only one of them has a profiling block around it.

e.g. whether or not a resource gets metrics tracked may depend on whether it was introduced via NC vs site.pp, etc.

In local tests where I've been experimenting with improvements for surfacing metrics data from Puppet Server, this change is a huge improvement and gives us much more interesting and complete data to look at.


cprice404:
This replaces #4197 .

cprice404:
@kylog I closed the other one and opened this one with all of the git hygiene resolved... and I've run tests locally and it seems good.

PR #4205, opened 2015-09-01
(PUP-4697) Make systemd the default service provider for all suse's *except* 10 and 11

kylog:
Previously, the default service provider for suse was 'redhat' with
an override of 'systemd' for suse 12 and 13. However this doesn't work for the rolling
release of suse (tumbleweed) or the next (leap), which are both systemd
based.

This reverses the logic so the default service provider is 'systemd'
with overrides for suse 10 and suse 11.

I also updated some unrelated spec tests in provider/service/systemd_spec.rb
to use 'stubs' rather than 'expects' - the latter made the tests unnecessarily
brittle and was testing implementation rather than behavior.

### DIFF
--- a/lib/puppet/provider/service/redhat.rb
+++ b/lib/puppet/provider/service/redhat.rb
@@ -8,7 +8,8 @@ Puppet::Type.type(:service).provide :redhat, :parent => :init, :source => :init 
 
   commands :chkconfig => "/sbin/chkconfig", :service => "/sbin/service"
 
-  defaultfor :osfamily => [:redhat, :suse]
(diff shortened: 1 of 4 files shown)


rombert:
@kylog - fix looks good to me, thank you! I've tested it ( backported on 3.7.5 for now ) on openSUSE Tumbleweed and it works as expected.

joshcooper:
Makes sense, `redhat` will be picked as the default on suse 10/11, because it has a higher specificity than `systemd`, otherwise we default to `systemd`. This change would switch the default for any suse version < 10, but since we don't support that :+1:

PR #4211, opened 2015-09-02
(PUP-5133) Add Solaris 10/11 SPARC node defs

johnduarte:
This commit adds node definitions for solaris sparc 10 and 11
agents using the SPARC architecture.


joshcooper:
stable here too

PR #4214, opened 2015-09-03
(PUP-5192) Do not assume that an event is a hash in resource/status.rb

roidelapluie:


joshcooper:
@roidelapluie Thank you for your contribution! We were just talking about this issue the other day. Could you update the commit message to describe under what conditions `event` is not a hash? Also could you add a test for this new behavior?

roidelapluie:
I will try to do that. Did you start any investigation on your side already? Is this the right place to fix this bug?

joshcooper:
@roidelapluie I think the issue shows up because we annotate the class when serializing to yaml:

```
 - !ruby/object:Puppet::Transaction::Event
      audited: false
      property: ensure
      previous_value: :absent
      desired_value: :file
      historical_value:
      message: created
      name: :file_created
      status: success
      time: 2015-08-24 11:54:51.523259000 -07:00
```

And when we read the yaml back in, the parser automatically creates an appropriate object. But I don't understand why we don't have similar problems with other types of objects in the report, e.g.

```
metrics:
  resources: !ruby/object:Puppet::Util::Metric
...
logs:
- !ruby/object:Puppet::Util::Log
...
resource_statuses:
  File[/tmp/foobar]: !ruby/object:Puppet::Resource::Status
  ...
    events:
    - !ruby/object:Puppet::Transaction::Event
    ...
  Schedule[puppet]: !ruby/object:Puppet::Resource::Status
```

Which we assume are always hashes: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction/report.rb#L198-L215

roidelapluie:
@joshcooper added test

branan:
We're going to spend a bit of time poking around trying to understand why this already worked for metrics and statuses before we merge this, so that we can make sure this is handled consistently across the various report classes. We'll get back to you soon.

roidelapluie:
Given what I have put in the gist:
running ruby ../5a6e15fa9a4cd49d9fb8/test.rb
c15a894612263efce9649cb44dcd4c6f534711b0 is the first bad commit
commit c15a894612263efce9649cb44dcd4c6f534711b0
Author: Henrik Lindberg <henrik.lindberg@cloudsmith.com>
Date:   Wed Oct 22 22:48:33 2014 +0200

    (PUP-3272) Change Status yaml deserialization to support psych
    
    This changes deserialization of a Status instance to use
    a custom initialization (based on params hash). The implementation
    is in a PsychSupport module to make it more visible and easier to add to
    other classes needed the same support.
    
    Since this change is not compatible with 1.8.7, the report_spec test
    that depends on this functionality is restricted to not run on Ruby
    1.8.7.

:040000 040000 99a0c3b8bccf97021d443f3ca3b1d5b68d314ddc 97d201c7711db92e30e954c726a8a76a4c2feadc M      lib
:040000 040000 3ec3325cc3ec4a0df70373df9e83e3c3b8b45271 cb6c5aa0a1a3aa29a8e9fe261ab1ce115853a21f M      spec
bisect run success


cc @hlindberg

roidelapluie:
It should help.

hlindberg:
Have a suspicion that the class(es) in question does not define that want to handle serialization with a to/from data hash instead of letting Yaml directly instantiate. IIRC, they should include "psyck support". This because if yaml is not told specifically what to do, it will serialize/deserialize ruby objects (all instance variables).

joshcooper:
@hlindberg we would need to update at least `Puppet::Util::Metric`, `Puppet::Util::Log`, and `Puppet::Transaction::Event`. However, `Puppet::Resource::Status` and some of the above classes already restrict what they serialize to yaml via their [`to_yaml_properties` method](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/application/agent.rb#L403). I'm thinking we should revert the change to `Puppet::Resource::Status`, and file a ticket to change all of the places that use `to_yaml_properties` for yaml serialization as that is [deprecated](https://github.com/tenderlove/psych/blob/1606060069c13b2bcdc672a1a0e89202a9eb5166/lib/psych/visitors/yaml_tree.rb#L392). Instead have them `include Puppet::Util::PsychSupport` and explicitly define what they the serialize/deserialize through `to_data_hash`/`initialize_from_hash`.

Also it appears the tests that ensured reports could be round-tripped through yaml were restricted to [running on ruby < 1.8.7 (which doesn't seem right)](https://github.com/puppetlabs/puppet/commit/c15a894612263efce9649cb44dcd4c6f534711b0#diff-c4cd3277e11756b71db425aba1d71f78R412), and then removed. We should add them back.

hlindberg:
@joshcooper agree - using the PsychSupport and our own API hides the details and makes it possible to  fix problems in one place (if/when it is changed again in the Ruby implementation).
You are probably right about the tests for roundtrips. Although... it was probably done for "no yaml on the wire", but that does not mean it should not roundtrip ok serialization wise.

joshcooper:
@roidelapluie could you update your PR to revert the `PsychSupport` change in `Puppet::Resource::Status` and add back the yaml roundtrip test in `report_spec.rb`

roidelapluie:
@joshcooper I have kept the test I have written myself and added back the other one.

They both fail and my test shows the same error

  1) Puppet::Resource::Status When converting to hash should be convertible again

     Failure/Error: @new_status = Puppet::Resource::Status.from_data_hash(@status.to_data_hash)

     NoMethodError:

       undefined method `[]' for #<Puppet::Transaction::Event @name="nil" @message="nil">

     # ./lib/puppet/transaction/event.rb:36:in `initialize_from_hash'

     # ./lib/puppet/transaction/event.rb:24:in `from_data_hash'

     # ./lib/puppet/resource/status.rb:184:in `block in initialize_from_hash'

     # ./lib/puppet/resource/status.rb:183:in `map'

     # ./lib/puppet/resource/status.rb:183:in `initialize_from_hash'

     # ./lib/puppet/resource/status.rb:103:in `from_data_hash'

     # ./spec/unit/resource/status_spec.rb:152:in `block (3 levels) in <top (required)>'

     # util/rspec_runner:39:in `run'

     # util/rspec_runner:53:in `<main>'

Finished in 3.67 seconds (files took 2.75 seconds to load)

  1) Puppet::Transaction::Report can make a round trip through yaml

     Failure/Error: Puppet[:report_serialization_format] = "yaml"

     ArgumentError:

       Attempt to assign a value to unknown setting :report_serialization_format

     # ./lib/puppet/settings.rb:1357:in `set'

     # ./lib/puppet/settings.rb:132:in `[]='

     # ./lib/puppet.rb:86:in `[]='

     # ./spec/unit/transaction/report_spec.rb:408:in `block (2 levels) in <top (required)>'

     # util/rspec_runner:39:in `run'

     # util/rspec_runner:53:in `<main>'

Finished in 32.15 seconds (files took 3.59 seconds to load)

roidelapluie:
The report_serialization_format was removed in 9692cac7dc625f292312eab802cf8582c730c941

joshcooper:
@roidelapluie ah yeah, I forgot about the `report_serialization_format` setting. You should be able to test yaml roundtripping directly as:

```
report = generate_report

yaml_output = report.render(:yaml)
tripped = Puppet::Transaction::Report.convert_from(:yaml, yaml_output)

yaml_output.should =~ /^--- /
expect_equivalent_reports(tripped, report)
```

Also I would remove the new test you added, because I don't think there will be a case where `from_data_hash` will be called with a report that already contains `Puppet::Resource::Status` objects.

roidelapluie:
@joshcooper done. let's see the outcome of tests.

joshcooper:
@roidelapluie Look great, thanks for your help with this. I've pulled it into our next sprint, and we'll get it merged in the next few days. Thanks again!

spec/unit/transaction/report_spec.rb
```diff
@@ -404,6 +404,16 @@ def add_statuses(count, type = :file)
     expect(error_report.render).to validate_against('api/schemas/report.json')
   end
 
+  it "can make a round trip through yaml" do
+    report = generate_report
+
+    yaml_output = report.render(:yaml)
+    tripped = Puppet::Transaction::Report.convert_from(:yaml, yaml_output)
+
+    yaml_output.should =~ /^--- /
```
joshcooper: I forgot this uses the deprecated rspec2 syntax which emits a deprecation warning. Could you change that to use rspec3 syntax:

```
    expect(yaml_output).to match(/^--- /)
```

roidelapluie:
done

joshcooper:
Awesome thanks @roidelapluie! While reviewing, I found that the generated report in `report_spec` never had any events! I added a commit to your and opened a new PR with both commits, so I'm going to close this one. Once it passes travis & appveyor, I'll merge.

PR #4215, opened 2015-09-03
(PUP-5024) Add acceptance test for enabling systemd-masked services

whopper:
See PUP-4605 for additional context.

### DIFF
--- a/acceptance/lib/puppet/acceptance/service_utils.rb
+++ b/acceptance/lib/puppet/acceptance/service_utils.rb
@@ -1,0 +1,17 @@
+require 'puppet/acceptance/common_utils'
+
+module Puppet
+  module Acceptance
+    module ServiceUtils
+
+      # Return whether a host supports the systemd provider.
+      # @param host [String] hostname
+      # @return [Boolean] whether the systemd provider is supported.
+      def supports_systemd? (host)
+        ruby = Puppet::Acceptance::CommandUtils.ruby_command(host)
+        suitable = on(host, "#{ruby} -e \"require 'puppet'; puts Puppet::Type.type(:service).provider(:systemd).suitable?\"" ).stdout.chomp
+        suitable == "true" ? true : false
+      end
+    end
+  end
+end
--- a/acceptance/tests/resource/service/ticket_5024_systemd_enabling_masked_service.rb
+++ b/acceptance/tests/resource/service/ticket_5024_systemd_enabling_masked_service.rb
@@ -1,0 +1,83 @@
+require 'puppet/acceptance/service_utils'
+extend Puppet::Acceptance::ServiceUtils
+
+test_name 'Systemd masked services are unmasked before attempting to start'
+
+# This test in intended to ensure that a service which was previously marked
+# as masked and then set to enabled will first be unmasked.
+confine :to, {}, agents.select { |agent| supports_systemd?(agent) }
+package_name = {'el'     => 'httpd',
+                'centos' => 'httpd',
+                'fedora' => 'httpd',
+                'debian' => 'apache2',
+                'sles'   => 'apache2',
+                'ubuntu' => 'apache2',
+}
+
+agents.each do |agent|
+  platform = agent.platform.variant
+  majrelease = on(agent, facter('os.release.major')).stdout.chomp
+
+  if ((platform == 'debian' && majrelease == '8') || (platform == 'ubuntu' && majrelease == '15.04'))
+    skip_test 'legit failures on debian8 and ubuntu15; see: PUP-5149'
+  end
+
+  init_script_systemd    = "/usr/lib/systemd/system/#{package_name[platform]}.service"
+  symlink_systemd        = "/etc/systemd/system/multi-user.target.wants/#{package_name[platform]}.service"
+  masked_symlink_systemd = "/etc/systemd/system/#{package_name[platform]}.service"
+
+  manifest_uninstall_httpd = %Q{
+    package { '#{package_name[platform]}':
+      ensure => absent,
+    }
+  }
+  manifest_install_httpd = %Q{
+    package { '#{package_name[platform]}':
+      ensure => present,
+    }
+  }
+  manifest_httpd_masked = %Q{
+    service { '#{package_name[platform]}':
+      provider => systemd,
+      enable => mask,
+    }
+  }
+  manifest_httpd_enabled = %Q{
+    service { '#{package_name[platform]}':
+      provider => systemd,
+      enable => true,
+      ensure => running,
+    }
+  }
+
+  teardown do
+    if platform == 'sles'
+      on agent, 'zypper remove -y apache2 apache2-prefork apache2-worker libapr1 libapr-util1'
+    else
+      apply_manifest_on(agent, manifest_uninstall_httpd)
+    end
+  end
+
+  step "Installing httpd"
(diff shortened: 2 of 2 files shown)


whopper:
Confirmed that this test passes with @melissa's change in https://github.com/puppetlabs/puppet/pull/4121, and that it fails against builds before that change:

```
        Info: Applying configuration version '1441319195'
        Error: Could not start Service[httpd]: Execution of '/usr/bin/systemctl start httpd' returned 1: Failed to issue method call: Unit httpd.service is masked.
        Error: /Stage[main]/Main/Service[httpd]/ensure: change from stopped to running failed: Could not start Service[httpd]: Execution of '/usr/bin/systemctl start httpd' returned 1: Failed to issue method call: Unit httpd.service is masked.
```

melissa:
:+1: Thanks Whopper!

acceptance/tests/resource/service/ticket_5024_systemd_enabling_masked_service.rb
```diff
@@ -0,0 +1,47 @@
+test_name 'Systemd masked services are unmasked before attempting to start'
+
+# This test in intended to ensure that a service which was previously marked
+# as masked and then set to enabled will first be unmasked.
+# Confine to RedHat 7, which uses the systemd provider.
+confine :to, :platform => 'el-7'
+confine :to, {}, hosts.select { |host| host['template'] =~ /redhat/ }
```
whopper: I decided to confine against redhat 7, as it's one of a handul of platforms with systemd as its default service provider. Is checking the template name crazy? It seems to be the only element that has the string 'redhat.' We can't confine against 'el-7' since systemd is not CentOS's default.

er0ck: i believe ubuntu15 and sles12 are systemd as well.

acceptance/tests/resource/service/ticket_5024_systemd_enabling_masked_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    assert_match(/enable => 'false'/, stdout, "Expected httpd service to be masked")
+  end
+
+  step "Enabling the httpd service"
+  apply_manifest_on(agent, manifest_httpd_enabled, :catch_failures => true)
+  on(agent, puppet_resource('service', 'httpd')) do
+    assert_match(/ensure => 'running'/, stdout, "Expected httpd service to be running")
+    assert_match(/enable => 'true'/, stdout, "Expected httpd service to be enabled")
+  end
+end
```
er0ck: this should probably have a teardown step to stop and uninstall apache/httpd

acceptance/tests/resource/service/ticket_5024_systemd_enabling_masked_service.rb
```diff
@@ -0,0 +1,47 @@
...
+
+agents.each do |agent|
+  step "Installing httpd"
+  apply_manifest_on(agent, manifest_install_httpd, :catch_failures => true)
+
+  step "Masking the httpd service"
+  apply_manifest_on(agent, manifest_httpd_masked, :catch_failures => true)
+  on(agent, puppet_resource('service', 'httpd')) do
+    assert_match(/ensure => 'stopped'/, stdout, "Expected httpd service to be stopped")
+    assert_match(/enable => 'false'/, stdout, "Expected httpd service to be masked")
```
er0ck: i don't think this asserts that it is masked.  i think you have to make sure the link points to /dev/null or similar?

whopper: @er0ck huh, I actually have no idea how to check where an arbitrary file has a symlink - especially if it might be /dev/null. Would it be alright to just check that the usual symlink in /etc doesn't exist?

er0ck: i think the symlink not existing is not enabled but not masked.  i forget some of those details.  should probably read up on what masking entails.  i think it might actually set the init script itself to /dev/null, so trying to start that service fails.  or maybe it's per-runlevel.
to check where a link points use `readlink` if it's on the host, or `ls -l` and parse out the -> entry

whopper: @er0ck ooh duh, I was misunderstanding what was supposed to be happening. Yeah, masking will literally make a symlink from /dev/null in the systemd /etc directory.

whopper:
@er0ck updated to check the symlink paths. Also added in your SLES teardown step.

I ran this against various redhats, centoses, and SLES. Like service_enable_linux, this test will legitimately fail in Debian 8 and Ubuntu 15.04, so I skipped them like you did in that test.

If you think the `supports_systemd?` method is sane, we could add `is_sysV` to it.

acceptance/lib/puppet/acceptance/service_utils.rb
```diff
@@ -0,0 +1,21 @@
+module Puppet
+  module Acceptance
+    module ServiceUtils
+
+      # Return whether a host supports the Systemd provider.
+      # @param host [String] hostname
+      # @return [Boolean] whether the systemd provider is supported.
+      def supports_systemd? (host)
```
kylog: This is going to require maintenance. What if we did the same thing the puppet code does in the provider: 
https://github.com/puppetlabs/puppet/blob/2871aaff1a5a5e1dbbeb0e485801dc4467caab6e/lib/puppet/provider/service/systemd.rb#L8?

whopper:
@kylog that works! I think having a new module for that is overkill anyway.

kylog:
@whopper actually I wasn't even thinking about the question of making it a module - that still makes sense to me, especially if that module gets fleshed out with an `is_sysV` method as well. I was just thinking of how to make the tests accommodate new distro releases without code changes.

acceptance/tests/resource/service/ticket_5024_systemd_enabling_masked_service.rb
```diff
@@ -0,0 +1,85 @@
+require 'puppet/acceptance/temp_file_utils'
+extend Puppet::Acceptance::TempFileUtils
+
+test_name 'Systemd masked services are unmasked before attempting to start'
+
+# This test in intended to ensure that a service which was previously marked
+# as masked and then set to enabled will first be unmasked.
+confine :to, {}, hosts.select { |host| dir_exists?(host, File.join('/', 'run', 'systemd', 'system'))
```
joshcooper: Another option is to ask puppet:

```
[root@qy4v8rqy8yixm1o ~]# /opt/puppetlabs/puppet/bin/ruby -e "require 'puppet'; puts Puppet::Type.type(:service).provider(:systemd).suitable?"
true
```

er0ck: i like this either way.  i'm pretty sure service_enable_linux.rb `is_sysV` will be the invert of whatever this is (in posix, outside of upstart).

er0ck:
@whopper this looks good.  did you want to make any changes according to josh or kylo's comments?

whopper:
@er0ck yeah, I'm just about to push an update to bring back the module - one moment

whopper:
@er0ck updated and retested - should be ready to go if everything looks good to you.

PR #4218, opened 2015-09-04
(PUP-5079) Check that 'nodes' parameter for applications is sound

thallgren:
This commit adds validation of the 'nodes' parameter in application
instances such that it must be a Hash where each key is a Node and
each value is a resource or an array of resources.

The code also checks that no resource is mapped to more than one
node.

A change was made to remove the prior more relaxed scheme where the
node could be on either side of the association in the hash. Now
it must be the key. This is natural since the node must be unique
and since the resources that it is associated with can be an array.

### DIFF
--- a/lib/puppet/network/http/api/master/v3/environment.rb
+++ b/lib/puppet/network/http/api/master/v3/environment.rb
@@ -20,14 +20,13 @@ class Puppet::Network::HTTP::API::Master::V3::Environment
     applications.each do |app|
       app_components = {}
       # Turn the 'nodes' hash into a map component ref => node name
-      node_mapping = app['nodes'].map do |k,v|
-        k.type == "Node" ? [k, v] : [v, k]
-      end.reduce({}) do |map, (node, comp)|
-        unless map[comp.ref].nil?
-          raise Puppet::ParseError, "Application #{app} maps component #{comp} to mutiple nodes"
+      node_mapping = {}
+      app['nodes'].each do |node, comps|
+        comps = [comps] unless comp.is_a?(Array)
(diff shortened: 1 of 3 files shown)


lib/puppet/network/http/api/master/v3/environment.rb
```diff
@@ -20,14 +20,13 @@ def call(request, response)
...
       app_components = {}
       # Turn the 'nodes' hash into a map component ref => node name
-      node_mapping = app['nodes'].map do |k,v|
-        k.type == "Node" ? [k, v] : [v, k]
-      end.reduce({}) do |map, (node, comp)|
-        unless map[comp.ref].nil?
-          raise Puppet::ParseError, "Application #{app} maps component #{comp} to mutiple nodes"
+      node_mapping = {}
+      app['nodes'].each do |node, comps|
+        comps = [comps] unless comp.is_a?(Array)
```
hlindberg: should we use flatten and perhaps unique here as well ?

thallgren: Not sure why. We already check that a node is associated to one or several resources. We don't allow nested arrays in that check. Regarding unique, is it meaningful to hide duplicates in a flat array? As it stands now a duplicate will trigger the 'maps component ... to multiple nodes' just below this comment (since the same component is mapped to the same node twice).

hlindberg: ok, will merge - we can always do that later if people use the same pattern here as for other similar things - i.e. that they want a kind of set operation  - "this set + that set". No big deal.

PR #4222, opened 2015-09-04
(PUP-4695) New acceptance test for calling hiera functions from within puppet te…

segatspiretech:
…mplates (epp, erb).

### DIFF
--- a/acceptance/tests/parser_functions/hiera_in_templates.rb
+++ b/acceptance/tests/parser_functions/hiera_in_templates.rb
@@ -1,79 +1,44 @@
 test_name "Calling Hiera function from inside templates"
 
-# Global variables
+@module_name = "hieratest"
+@coderoot = master.tmpdir("#{@module_name}")
+@resultdir = "#{@coderoot}/results"
 
-$module_name = "hieratest"
-$coderoot = master.tmpdir("#{$module_name}")
-$resultdir = "#{$coderoot}/results"
+@msg_default = 'message from default.yaml'
+@msg_production = 'message from production.yaml'
+@msg1os = 'message1 from {osfamily}.yaml'
+@msg2os = 'message2 from {osfamily}.yaml'
+@msg_fqdn = 'messsage from {fqdn}.yaml'
 
-$msg_default = 'message from default.yaml'
-$msg_production = 'message from production.yaml'
-$msg1os = 'message1 from {osfamily}.yaml'
-$msg2os = 'message2 from {osfamily}.yaml'
-$msg_fqdn = 'messsage from {fqdn}.yaml'
+@k1 = 'key1'
+@k2 = 'key2'
+@k3 = 'key3'
 
-$k1 = 'key1'
-$k2 = 'key2'
-$k3 = 'key3'
+@hval3os = 'hash_value3 from production.yaml'
+@hval2os = 'hash_value2 from production.yaml'
+@hval1p = 'hash_value1 from {osfamily}.yaml'
+@hval2p = 'hash_value2 from {osfamily}.yaml'
 
-$val1p = 'value1 from production.yaml'
-$val2p = 'value2 from production.yaml'
-$val2os = 'value2 from {osfamily}.yaml'
-$val3os = 'value3 from {osfamily}.yaml'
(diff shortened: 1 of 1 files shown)


er0ck:
This looks pretty good minus some bookkeeping syntax issues.
@segatspiretech you might want to review the [CONTRIBUTING.md|https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md] instructions in this repo.  In particular, the PR summary and commit headline need the ticket number preface in parens.
Also, the style at puppetlabs is to including a longer commit message after the commit headline that describes why the test is necessary.  This allows easier maintenance down the road, when someone else has to determine what happened at this commit.

acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -0,0 +1,403 @@
...
+
+$mod_default_msg = 'This file created by mod_default.'
+$mod_osfamily_msg = 'This file created by mod_osfamily.'
+$mod_production_msg = 'This file created by mod_production.'
+$mod_fqdn_msg = 'This file created by mod_fqdn.'
+
+
+def create_environment
+
+  #    Creates a directory tree like this on the master, under /tmp/hieratest/
```
er0ck: we should probably remove this comment.  most people around here assume a tree like this and the rest can be inferred from the puppet code "as documentation".

acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -0,0 +1,403 @@
+test_name "Calling Hiera function from inside templates"
+
+# Global variables
```
er0ck: remove this

acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -0,0 +1,403 @@
...
+
+$module_name = "hieratest"
```
er0ck: these don't have to be globals, nor should they be.  variables inside this test's namespace are fine. (remove the '$')

acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -0,0 +1,403 @@
...
+    content => template('#{$module_name}/hieratest_results_erb.erb'),
+  }
+}
+"
+}
+
+file { "#{moduledir}/manifests/mod_default.pp":
+  content => "
+class #{$module_name}::mod_default {
+  notify{\\"module mod_default invoked.\\\\n\\":}
```
er0ck: you could use single quotes here rather than escaping the doubles.  also, in most cases the \n is implied in a notify's output.

acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -0,0 +1,403 @@
...
+  file {'/#{$resultdir}/mod_default':
+    ensure  => 'file',
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
+    content => \\\"#{$mod_default_msg}\\\\n\\\",
```
er0ck: same thing with single quotes here.  otherwise these are difficult to read.  you might be able to omit the newline here as well. :question:

acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -0,0 +1,403 @@
...
+      agent,
+      puppet('agent', "-t --server #{master}"),
+      :acceptable_exit_codes => [2]
+    )
+
+    step "####### Verifying hiera calls from erb template #######"
+    on(agent, "cat #{$coderoot}/results/hieratest_results_erb")
+
+    step "Verifying hiera() call #1."
+    assert_match(/#{$h_m_call}: #{$msg_production}/, stdout)
```
er0ck: this comparison to stdout uses the magical/implied global 'result' object which is deprecated but probably doesn't warn.
typically we assign a result from the output of the on() method and compare to the stdout member of that result object.  this way we know to what result we are comparing.  
also, asserts in general need a message so one can quickly ascertain what happened if they are not familiar with the test.

segatspiretech:
Replacing with new PR.

PR #4231, opened 2015-09-09
[PUP-5212] Fix for pip package provider doesn't honor HTTP proxy settings

earsdown:
Technique for setting http_proxy taken from pkgdmg provider.

### DIFF
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -3,6 +3,7 @@
 
 require 'puppet/provider/package'
 require 'xmlrpc/client'
+require 'puppet/util/http_proxy'
 
 Puppet::Type.type(:package).provide :pip,
   :parent => ::Puppet::Provider::Package do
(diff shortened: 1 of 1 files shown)


earsdown:
Btw, apologies for using square brackets instead of regular brackets in the commit message - will remember for next time.

lib/puppet/provider/package/pip.rb
```diff
@@ -60,7 +61,16 @@ def query
   # cache of PyPI's package list so this operation will always have to
   # ask the web service.
   def latest
-    client = XMLRPC::Client.new2("http://pypi.python.org/pypi")
+    http_proxy_host = Puppet::Util::HttpProxy.http_proxy_host
+    http_proxy_port = Puppet::Util::HttpProxy.http_proxy_port
+    if http_proxy_host and http_proxy_port
```
peterhuene: We require the use of logical operators (&&) instead of the and/or operators as because in some instances (not this case, though) the precedence can cause unintended consequences.

lib/puppet/provider/package/pip.rb
```diff
@@ -60,7 +61,16 @@ def query
...
+      proxy = "#{http_proxy_host}:#{http_proxy_port}"
+    elsif http_proxy_host and not http_proxy_port
```
peterhuene: Same comment as above.

peterhuene:
Hi @earsdown.  Thanks for the contribution!  Could you please add a spec that tests the provider when the proxy host/port are set to confirm the settings are now being respected by the provider?  Additionally, please amend the commit message to use parenthesis instead of brackets for specifying the ticket number.  Thanks!

peterhuene:
Closing this in favor of #4232.

PR #4232, opened 2015-09-09
[PUP-5212] Fix for pip package provider doesn't honor HTTP proxy settings

earsdown:


earsdown:
Btw, apologies for using square brackets instead of regular brackets in the commit message - will remember for next time.

lib/puppet/provider/package/pip.rb
```diff
@@ -56,7 +57,16 @@ def query
   # cache of PyPI's package list so this operation will always have to
   # ask the web service.
   def latest
-    client = XMLRPC::Client.new2("http://pypi.python.org/pypi")
+    http_proxy_host = Puppet::Util::HttpProxy.http_proxy_host
+    http_proxy_port = Puppet::Util::HttpProxy.http_proxy_port
+    if http_proxy_host and http_proxy_port
```
peterhuene: We require the use of logical operators (&&) instead of the and/or operators as because in some instances (not this case, though) the precedence can cause unintended consequences.

lib/puppet/provider/package/pip.rb
```diff
@@ -56,7 +57,16 @@ def query
...
+      proxy = "#{http_proxy_host}:#{http_proxy_port}"
+    elsif http_proxy_host and not http_proxy_port
```
peterhuene: Same comment as above.

peterhuene:
Hi @earsdown. Thanks for the contribution! Could you please add a spec that tests the provider when the proxy host/port are set to confirm the settings are now being respected by the provider? Additionally, please amend the commit message to use parenthesis instead of brackets for specifying the ticket number; master currently validates the commit message format and this will break when merged up. Thanks!

peterhuene:
Also, if you could amend the commit message to not include "back port to 3.x", that would be great.  If this is accepted, we'll merge it up from 3.x to master.

joshcooper:
Hi @earsdown, thank you for your contribution! When you get a chance could you review @peterhuene's comments?

joshcooper:
ping @earsdown

HAIL9000:
Hey @earsdown, thanks for the contribution! Just wanted to check in and see if you still had time to work on this. Please let us know either way if you get a chance.

Thanks!

MikaelSmith:
Closing this for lack of response and a signed CLA.

PR #4237, opened 2015-09-10
(PUP-4476) Call lookup with the correct parameters

HAIL9000:
Prior to this commit, the lookup application was calling the internal
lookup function in a way that was compatible with the old version
of the function (eight parameters) rather than the current version
of the function (six parameters). This meant that attempts to use
the lookup application would fail due to the function being called
improperly.

Fix this issue and also remove a test that had too much stubbing
and was not longer useful.

### DIFF
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -96,7 +96,7 @@ class Puppet::Application::Lookup < Puppet::Application
       end
     end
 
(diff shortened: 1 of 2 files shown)


HAIL9000:
I also think that Shaigy and I are seeing a but where the errors raised in run_command (e.g. when you don't include a node) are being raised but not appearing on the CLI. My guess is that puppet is putting them somewhere else, but I'm not sure what the correct way to raise an error is so that this doesn't happen.

PR #4238, opened 2015-09-10
(PUP-4442) Only rmdir if mkdir succeeded

joshcooper:
Previously, if you tried to manage a file with content and the parent
directory didn't exist, puppet would report `rmdir` failed:

    Error: Could not set 'file' on ensure: No such file or directory @
    dir_s_rmdir - .../file20150909-11354-3uzm9m.lock

The file type uses `Puppet::Util.replace_file` to atomically overwrite
file content, which creates a unique file in the same directory as
the file we're trying to manage, to ensure we're on the same filesystem.
But if the parent directory doesn't exist, `mkdir` raises ENOENT, and
then we called `rmdir`, which also raises ENOENT, masking the original
exception.

This commit changes uniquefile to only call `rmdir` if `mkdir` succeeds.

### DIFF
--- a/lib/puppet/file_system/uniquefile.rb
+++ b/lib/puppet/file_system/uniquefile.rb
@@ -173,9 +173,11 @@ class Puppet::FileSystem::Uniquefile < DelegateClass(File)
     def locking(tmpname)
       lock = tmpname + '.lock'
       mkdir(lock)
-      yield
-    ensure
-      rmdir(lock) if lock
+      begin
+        yield
+      ensure
+        rmdir(lock)
+      end
     end
 
     def mkdir(*args)
--- a/spec/unit/file_system/uniquefile_spec.rb
+++ b/spec/unit/file_system/uniquefile_spec.rb
@@ -46,6 +46,11 @@ describe Puppet::FileSystem::Uniquefile do
     expect(Puppet::FileSystem.exist?(filename)).to be_falsey
   end
 
+  it "raises a mkdir exception when an ancestor directory doesn't exist" do
+    expect do
+      Puppet::FileSystem::Uniquefile.new('foo', File.expand_path('/parent/dir/does/not/exist'))
+    end.to raise_error(Errno::ENOENT, /mkdir/)
+  end
 
   context "Ruby 1.9.3 Tempfile tests" do
     # the remaining tests in this file are ported directly from the ruby 1.9.3 source,


kylog:
It looks like the spec expectation that `mkdir` is in the exception message doesn't hold for older rubies. Maybe just expect `ENOENT`?

kylog:
Quick before-and-after. I'm definitely :+1: on this PR (once we sort the specs). But even with this fix, the error messages remain pretty noisy in this case.

Before:

```
puppet apply -e 'file { "/path/does/not/exist": ensure => file, content => "boo" }'
Notice: Compiled catalog for kylo.local in environment production in 0.44 seconds
Error: Could not set 'file' on ensure: No such file or directory @ dir_s_rmdir - /path/does/not/exist20150910-6577-17sasrz.lock at 1:
Error: Could not set 'file' on ensure: No such file or directory @ dir_s_rmdir - /path/does/not/exist20150910-6577-17sasrz.lock at 1:
Wrapped exception:
No such file or directory @ dir_s_rmdir - /path/does/not/exist20150910-6577-17sasrz.lock
Error: /Stage[main]/Main/File[/path/does/not/exist]/ensure: change from absent to file failed: Could not set 'file' on ensure: No such file or directory @ dir_s_rmdir - /path/does/not/exist20150910-6577-17sasrz.lock at 1:
```

After:

```
puppet apply -e 'file { "/path/does/not/exist": ensure => file, content => "boo" }'
Notice: Compiled catalog for kylo.local in environment production in 0.33 seconds
Error: Could not set 'file' on ensure: No such file or directory @ dir_s_mkdir - /path/does/not/exist20150910-6878-1rlx8rq.lock at 1:
Error: Could not set 'file' on ensure: No such file or directory @ dir_s_mkdir - /path/does/not/exist20150910-6878-1rlx8rq.lock at 1:
Wrapped exception:
No such file or directory @ dir_s_mkdir - /path/does/not/exist20150910-6878-1rlx8rq.lock
Error: /Stage[main]/Main/File[/path/does/not/exist]/ensure: change from absent to file failed: Could not set 'file' on ensure: No such file or directory @ dir_s_mkdir - /path/does/not/exist20150910-6878-1rlx8rq.lock at 1:
```

joshcooper:
Yeah, I realized after I submitted this that the error message isn't what @rnelson0 was asking for directly, more what the follow up comment was about. We could check if the parent exists prior to calling `mkdir`, and raise a clearer message if it doesn't. I'm also not sure why we emit two Error messages.

Also there are a few different code paths to consider. The resource can be a dir/link or empty file, file with content (as a string), content (as a checksum), or source, and we'd need to make sure those are all handled consistently.

joshcooper:
Needs more work, closing

PR #4242, opened 2015-09-11
(PUP-5202) Add help for puppet lookup CLI command

HAIL9000:
Prior to this commit running `puppet lookup --help` would do nothing.
Now a help messages is displayed.

### DIFF
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -72,6 +72,102 @@ class Puppet::Application::Lookup < Puppet::Application
     Puppet::Util::Log.newdestination(:console)
   end
 
+  def help
+    <<-'HELP'
+
+puppet-lookup(8) -- Data in modules lookup function
+========
+
+SYNOPSIS
+--------
+The lookup command is used for debugging and testing a given data
+configuration. For a given data key, lookup will return which value
+will be produced with an optional explanation of how the system arrived
+at that value.
+
+USAGE
+-----
+puppet lookup [--help] [--type <TYPESTRING>] [--merge unique|hash|deep]
+  [--knock_out_prefix <PREFIX-STRING>] [--sort_merged_arrays]
+  [--unpack_arrays <STRING-VALUE>] [--merge_hash_arrays] [--explain]
+  [--default <VALUE>] [--node <NODE-NAME>] [--facts <FILE>] <keys>
+
+DESCRIPTION
+-----------
+The lookup command is a CLI interface for the puppet lookup function.
+When given one or more keys, the lookup command will return the first
+value found. The only component that is required in addition to the
+keys is a node (passed in via --node) which specifies which scope to
+perform the lookup in. It will exit with 0 if a value was returned
+and 1 otherwise.
+
+The other options are as passed into the lookup function, and the effect
+they have on the lookup is described in more detail in the header
+for the lookup function:
+
+https://github.com/puppetlabs/puppet/blob/master/lib/puppet/functions/lookup.rb
+
+OPTIONS
+-------
+These options and their effects are decribed in more detail in
+the puppet lookup function linked to above.
+
+* --help:
+  Print this help message.
+
+* --type <TYPESTRING>:
+  Assert that the value has the specified type.
+
+* --merge unique|hash|deep:
+  Specify the merge strategy. 'hash' performs a simple hash-merge by
+  overwriting keys of lower lookup priority. 'unique' appends everything
+  to an array containing no nested arrays and where all duplicates have been
+  removed. 'deep' Performs a deep merge on values of Array and Hash type. There
+  are additional option flags that can be used with 'deep'.
+
+* --knock_out_prefix <PREFIX-STRING>
+  Can be used with the 'deep' merge strategy. Specify string value to signify
+  prefix which deletes elements from existing element.
+
+* --sort_merged_arrays
+  Can be used with the 'deep' merge strategy. When this flag is used all
+  merged arrays will be sorted.
+
+* --unpack_arrays <STRING-VALUE>
+  Can be used with the 'deep' merge strategy. Specify a string value used
+  as a deliminator to join all array values and then split them again.
+
+* --merge_hash_arrays
+  Can be used with the 'deep' merge strategy. When this flag is used arrays
+  and hashes will be merged.
+
+* --explain
+  Print an explanation for how the result was obtained.
+
+* --default <VALUE>
+  A value produced if no value was found in the lookup.
+
+* --node <NODE-NAME>
+  Specify node which defines the scope in which the lookup will be performed.
+
+* --facts <FILE>
+  Specify a .json, or .yaml file holding key => value mappings that will
+  populate a scope with data that is used when looking up.
+
+EXAMPLE
+-------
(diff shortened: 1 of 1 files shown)


lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+
+puppet-lookup(8) -- Data in modules lookup function
+========
+
+SYNOPSIS
+--------
+The lookup command is used for debugging and testing a given data
+configuration. For a given data key, lookup will return which value
+will be produced with an optional explanation of how the system arrived
+at that value.
```
thallgren: Both value and explanation will not be returned in the the implementation that is pending for PUP-4475. It's either or. Reason being that with plain text explain, the result is clearly visible within the explanation, and if 'explain_json' (new option) is used, the producing the value in addition to the json output would clobber the value. If we decide that that's the way to go, then this doc needs to be updated.

lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+
+DESCRIPTION
+-----------
+The lookup command is a CLI interface for the puppet lookup function.
+When given one or more keys, the lookup command will return the first
+value found. The only component that is required in addition to the
+keys is a node (passed in via --node) which specifies which scope to
+perform the lookup in. It will exit with 0 if a value was returned
+and 1 otherwise.
+
```
thallgren: This is true when no explanation is wanted. The objective might well be to produce an explanation of why a value is not found so in PUP-4475, when explain is active, the exit code is always 0 (unless something is majorly wrong).

lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+  Can be used with the 'deep' merge strategy. Specify a string value used
+  as a deliminator to join all array values and then split them again.
+
+* --merge_hash_arrays
+  Can be used with the 'deep' merge strategy. When this flag is used arrays
+  and hashes will be merged.
+
+* --explain
+  Print an explanation for how the result was obtained.
+
```
thallgren: It will also print an explanation of why a result was not obtained so the wording needs to change.

Also, I think we should add '--explain_json' which produces a json formatted explanation.

lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+will be produced with an optional explanation of how the system arrived
+at that value.
+
+USAGE
+-----
+puppet lookup [--help] [--type <TYPESTRING>] [--merge unique|hash|deep]
+  [--knock_out_prefix <PREFIX-STRING>] [--sort_merged_arrays]
+  [--unpack_arrays <STRING-VALUE>] [--merge_hash_arrays] [--explain]
+  [--default <VALUE>] [--node <NODE-NAME>] [--facts <FILE>] <keys>
+
```
thallgren: I changed all options to use dashes instead of underscores in the option name since that seems to be what other puppet commands do.

thallgren: Add `[--render-as <format>]`

lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+  A value produced if no value was found in the lookup.
+
+* --node <NODE-NAME>
+  Specify node which defines the scope in which the lookup will be performed.
+
+* --facts <FILE>
+  Specify a .json, or .yaml file holding key => value mappings that will
+  populate a scope with data that is used when looking up.
+
+EXAMPLE
```
thallgren: Add `--render-as <FORMAT>` where `<FORMAT>` is one of _s_, _json_, _yaml_, _binary_ or _msgpack_ (these are the standard ones used by all facet commands). _s_ means plain text. The default when producing a value is _yaml_ and _s_ when producing an explanation.

lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+value found. The only component that is required in addition to the
+keys is a node (passed in via --node) which specifies which scope to
+perform the lookup in. It will exit with 0 if a value was returned
+and 1 otherwise.
+
+The other options are as passed into the lookup function, and the effect
+they have on the lookup is described in more detail in the header
+for the lookup function:
+
+https://github.com/puppetlabs/puppet/blob/master/lib/puppet/functions/lookup.rb
```
hlindberg: This is problematic as a link since it refers to the latest version (i.e, head on master). Maybe it is best to point to the documentation site using a google link (like we do for deprecations). Text should probably say something like "look at the documentation for the loopup() function for the version of puppet you are using" - with a link to the documentation page (via an indirect google link so we can update it). Also log a ticket to review that link once we are close to release to ensure that it works). (The other alternative is to include the documentation here (copy/paste/edit) from the lookup function, but that means we have to remember to make updates in two places.)

lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+  and hashes will be merged.
+
+* --explain
+  Print an explanation for how the result was obtained.
+
+* --default <VALUE>
+  A value produced if no value was found in the lookup.
+
+* --node <NODE-NAME>
+  Specify node which defines the scope in which the lookup will be performed.
```
hlindberg: What happens if it is not given?

lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+
+* --facts <FILE>
```
hlindberg: What happens if not given? If given, do they override values obtained for the node (or vice versa)?

lib/puppet/application/lookup.rb
```diff
@@ -57,6 +57,102 @@ class Puppet::Application::Lookup < Puppet::Application
...
+  Specify a .json, or .yaml file holding key => value mappings that will
+  populate a scope with data that is used when looking up.
+
+EXAMPLE
+-------
+  $ puppet lookup --node agent.local key_name
+  $ puppet lookup --node agent.local --merge deep --knock_out_prefix foo key_name_one key_name_two
+  $ puppet lookup --node agent.local --default bar key_name
+
```
hlindberg: A short description what each example does would be great.

lib/puppet/application/lookup.rb
HAIL9000: I don't actually know the answer to this question unfortunately

HAIL9000:
Updated! I didn't actually know the answer to the question about what happens if the --facts flag is not used. I can amend my most recent commit or just add that description when I add the --facts flag.

hlindberg:
@HAIL9000  This PR needs a rebase

hlindberg:
Will merge once it is rebased. We can log specific tickets for small remaining things.

HAIL9000:
@hlindberg Sorry about that, I totally missed your comment. Rebased!

PR #4244, opened 2015-09-11
WIP - DO NOT MERGE (maint) Add acceptance pre-suite to rm puppet on sparc

johnduarte:
This commit adds a beaker pre-suite to ensure that puppet packages
and components are removed when the platform is on sparc and the
hypervisor is none.

This ensures that the sparc testing platform does not have any
remaining puppet bits when it is not being managed by a provisioning
hypervisor.


joshcooper:
Should we do the same thing we're doing for facter https://github.com/puppetlabs/facter/pull/1182?

johnduarte:
@joshcooper yes, I will propagate the changes in  puppetlabs/facter#1182 when we have the new method in Beaker. Thanks.

PR #4245, opened 2015-09-14
(PUP-4475) Add explain-ability to the lookup and data provider APIs

thallgren:
This commit adds report callbacks to the Lookup::Invocation object
which is passed down to all instances involved in the lookup. These
callbacks will either do nothing at all (no Explainer attached to
the Invocation) or tell the explainer to accept data for that
particular report.

A new 'with' method was also added. It is used when a lookup changes
its context. This happens when the lookup is branching due to things
like several paths, global/environment/module scopes, or when
evaluating interpolations in a string. The with method is used as
a wrapper (yields to a block where the lookup continues) and will
push and pop contextual objects in the Explainer and thereby build
a full DAG of all branches that the lookup invocation took part in.

The LookupInvocation that resided in Puppet::DataBinding was moved
to Puppet::Pops::Lookup and renamed to Invocation to make it adjacent
to the Explainer. This was motivated by their close affinity and that
neither of those classes are specific to DataBinding. The impact of
this will be very minor given that no release has gone out where the
LookupInvocation has been included.

The lookup CLI command was change so that it now exposes the explain
feature.

### DIFF
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -1,19 +1,19 @@
 require 'puppet/application'
-require 'puppet/pops/lookup'
+require 'puppet/pops'
 require 'puppet/node'
 require 'puppet/parser/compiler'
 
 class Puppet::Application::Lookup < Puppet::Application
 
-  RUNHELP = "Run 'puppet lookup --help' for more details".freeze
-  DEEP_MERGE_OPTIONS = "--knock_out_prefix, --sort_merged_arrays, --unpack_arrays, and --merge_hash_arrays".freeze
+  RUN_HELP = "Run 'puppet lookup --help' for more details".freeze
+  DEEP_MERGE_OPTIONS = '--knock-out-prefix, --sort-merged-arrays, --unpack-arrays, and --merge-hash-arrays'.freeze
 
   # Options for lookup
   option('--merge TYPE') do |arg|
     if %w{unique hash deep}.include?(arg)
       options[:merge] = arg
     else
-      raise "The --merge option only accepts 'unique', 'hash', or 'deep' as arguments.\n#{RUNHELP}"
+      raise "The --merge option only accepts 'unique', 'hash', or 'deep' as arguments.\n#{RUN_HELP}"
     end
   end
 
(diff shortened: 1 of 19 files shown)


lib/puppet/application/lookup.rb
```diff
@@ -33,9 +33,12 @@ class Puppet::Application::Lookup < Puppet::Application
 
   option('--merge_hash_arrays')
 
-  # not yet supported
   option('--explain')
 
+  option('--explain_json')
```
hlindberg: Other commands in puppet use a separate `--format=f` flag to set the format of the output. I think this command should do the same.

lib/puppet/resource.rb
```diff
@@ -480,7 +480,7 @@ def lookup_in_environment(name, scope)
   def lookup_in_module(name, scope)
     found = false
     value = catch(:no_such_key) do
-      v = Puppet::DataProviders.lookup_in_module(name, Puppet::DataBinding::LookupInvocation.new(scope), nil)
+      v = Puppet::DataProviders.lookup_in_module(name, Puppet::Pops::Lookup::Invocation.new(scope), nil)
       found = true
```
hlindberg: Puppet::Resource is also agent side, but i suppose the methods lookup_xxx never gets called there - raises the issue of pops initialization on agent side otherwise... (not sure what we concluded there).

hlindberg:
Looks good. I was thinking that later we want to break out the presentation logic from the ExplainNode to make it a pure data representation/tree, and have the presentation of it be separate. The logic is not very different from the model tree dumper, which also has indent/dedent support. OTOH, a refactoring of that logic into something generic is not within scope of this ticket. One reason for breaking out the actual presentation is to be able to also internationalize the result (and keep that concern separate).

lib/puppet/application/lookup.rb
thallgren: The only command that uses `--format` is AFAICS the 'doc' command and there it seems to have a slightly different meaning (renders pdf or text). A more commonly used pattern for what we want to achieve here is to use `--render-as <format>` and then map that to a renderer returned by `Puppet::Network::FormatHandler.format()`. The user then has the option to use _s_ (text), _json_, _yaml_, _msgpack_ and _binary_ per default.

lib/puppet/application/lookup.rb
```diff
@@ -96,13 +98,31 @@ def run_command
...
+
+    type = options.include?(:type) ? Puppet::Pops::Types::TypeParser.new.parse(options[:type]) : nil
+
+    generate_scope do |scope|
+      lookup_invocation = Puppet::Pops::Lookup::Invocation.new(scope, {}, {}, explain)
+      begin
+        result = Puppet::Pops::Lookup.lookup(keys, type, options[:default_value], use_default_value, merge_options, lookup_invocation)
+        puts renderer.render(result) unless explain
+      rescue Puppet::Error => e
+        raise unless e.message =~ /lookup\(\) did not find a value/
```
hlindberg: This does not feel like the most robust API (checking error message)

hlindberg:
Lacks a test that explain output can be obtained in YAML.

hlindberg:
tests are failing due to differences in yaml rendering between ruby versions - need to come up with a more robust way of comparing results

PR #4246, opened 2015-09-14
(PUP-2575) Sort OSX group membership before comparison

ccaviness:
OSX group membership is currently returned as simple Arrays. Comparing these
is strict, and the order of the resource must exactly match the order given
in the manifest, or a spurious change notice will be generated.

To fix this, add a members_insync? method on the directoryservice group
provider.

### DIFF
--- a/lib/puppet/provider/group/directoryservice.rb
+++ b/lib/puppet/provider/group/directoryservice.rb
@@ -9,4 +9,18 @@ Puppet::Type.type(:group).provide :directoryservice, :parent => Puppet::Provider
   confine :operatingsystem => :darwin
   defaultfor :operatingsystem => :darwin
   has_feature :manages_members
+
+  def members_insync?(current, should)
+    return false unless current
+    if current == :absent and should.empty?
+      return true
+    else
+      if current.respond_to?(:sort) and should.respond_to?(:sort)
+        return current.sort.uniq == should.sort.uniq
+      else
+        return current == should
+      end
+    end
+  end
+
 end
--- a/lib/puppet/provider/group/directoryservice.rb
+++ b/lib/puppet/provider/group/directoryservice.rb
@@ -12,14 +12,10 @@ Puppet::Type.type(:group).provide :directoryservice, :parent => Puppet::Provider
 
   def members_insync?(current, should)
     return false unless current
-    if current == :absent and should.empty?
-      return true
+    if current == :absent
+      return should.empty?
     else
-      if current.respond_to?(:sort) and should.respond_to?(:sort)
-        return current.sort.uniq == should.sort.uniq
-      else
-        return current == should
-      end
+      return current.sort.uniq == should.sort.uniq
     end
   end
 
--- a/spec/unit/provider/group/directoryservice_spec.rb
+++ b/spec/unit/provider/group/directoryservice_spec.rb
@@ -1,0 +1,29 @@
+require 'spec_helper'
+
+describe Puppet::Type.type(:group).provider(:directoryservice) do
+  let :resource do
+    Puppet::Type.type(:group).new(
+      :title => 'testgroup',
+      :provider => :directoryservice,
+    )
+  end
+
+  let(:provider) { resource.provider }
+
+  it 'should return true for same lists of unordered members' do
+    expect(provider.members_insync?(['user1', 'user2'], ['user2', 'user1'])).to be_truthy
+  end
+
+  it 'should return false when the group currently has no members' do
+    expect(provider.members_insync?([], ['user2', 'user1'])).to be_falsey
+  end
+
+  it 'should return true for the same lists of members irrespective of duplicates' do
+    expect(provider.members_insync?(['user1', 'user2', 'user2'], ['user1', 'user2'])).to be_truthy
+  end
+
+  it "should return true when current and should members are empty lists" do
+    expect(provider.members_insync?([], [])).to be_truthy
+  end
+
+end
--- a/spec/unit/provider/group/directoryservice_spec.rb
+++ b/spec/unit/provider/group/directoryservice_spec.rb
@@ -26,4 +26,8 @@ describe Puppet::Type.type(:group).provider(:directoryservice) do
     expect(provider.members_insync?([], [])).to be_truthy
   end
 
+  it "should return true when current is :absent and should members is empty list" do
+    expect(provider.members_insync?(:absent, [])).to be_truthy
+  end
+
 end


MikaelSmith:
That's intriguing. Order seems like it should have no significance. Would you mind looking into extending the spec tests in https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/nameservice/directoryservice_spec.rb to verify it won't update if `current.sorted == desired.sorted`?

lib/puppet/provider/group/directoryservice.rb
```diff
@@ -9,4 +9,10 @@
   confine :operatingsystem => :darwin
   defaultfor :operatingsystem => :darwin
   has_feature :manages_members
+
+  def members_insync?(current, should)
+  	return false unless current
```
peterhuene: Nitpick, but it looks like the surrounding tab width is 2 and not 4.  We generally don't care about tab width so long as it is consistently applied in a file (assuming it uses spaces and not tabs).

lib/puppet/provider/group/directoryservice.rb
```diff
@@ -9,4 +9,10 @@
...
+	return current.sort == should.sort
```
joshcooper: Can `should` contain duplicates? On Windows it can, so we do https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/group/windows_adsi.rb#L33

ccaviness:
I'm really not sure how to write a test for this? With the old code, there was never actually a *change* happening; no calls to `dscl` to modify member lists, just a notice:

    Notice: /Stage[main]/Main/Group[order]/members: members changed 'root,clay,daemon' to 'clay,root,daemon'

Not really understanding spec (or, honestly, puppet's internals), I tried this:
```
    it 'should not update if sorted current and should membership lists match' do
      current = ["foo", "bar", "baz"]
      desired = ["baz", "foo", "bar"]
      group = 'ordering'

      @resource.stubs(:[]).with(:name).returns(group)
      @provider.instance_variable_set(:@property_value_cache_hash,
                                      { :members => current })
      @provider.expects(:update).never
    end
```

But it passed with the old and new code, so that's clearly not testing the right thing.

Any pointers?

lib/puppet/provider/group/directoryservice.rb
ccaviness: Ah, thanks. Bad editor!

lib/puppet/provider/group/directoryservice.rb
ccaviness: I'm probably ok with saying `should` shouldn't ever have duplicates.

ccaviness:
Yeah, I can't sort out the spec tests for this; I have a suspicion that something vaguely like what I've attempted in `spec/unit/provider/group/directoryservice_spec.rb` is the right way to go, where I have a group, and verify that it's in sync. But I really have no idea how to actually do that, and instead have been cargo-cult copying bits of code from other tests without really understanding what's going on.

I'll need some help with this, sorry.

ccaviness:
Ok, I have some sort of test. It fails if the `members_insync?` method is missing (or doesn't sort). I blindly cobbled some stuff from the `windows_adsi` test; if there's a better method for this, or I did something silly/incorrect, let me know.

MikaelSmith:
Thanks. I meant to respond and help out, but been working on some other things. I'll try to take a look at it in the next few days.

lib/puppet/provider/group/directoryservice.rb
MikaelSmith: I concur, looks there should never be duplicate group names.

ccaviness:
When you have a chance to help with tests, just discovered a case that will need testing. If you have an existing group with no members, then add members to it, it will fail with an undefined method error:

```
fox:~ crc$ echo "group { 'testgroup': ensure => present}" | sudo puppet apply
Notice: Compiled catalog for 045627f2-3d66-46b1-95bc-1bc74154e1d3 in environment gmac_unstable in 0.60 seconds
Notice: /Stage[main]/Main/Group[testgroup]/ensure: created
Notice: Finished catalog run in 0.37 seconds
fox:~ crc$ echo "group { 'testgroup': ensure => present, members => ['crc'] }" | sudo puppet apply
Notice: Compiled catalog for 045627f2-3d66-46b1-95bc-1bc74154e1d3 in environment gmac_unstable in 0.55 seconds
Error: undefined method `sort' for :absent:Symbol
Error: /Stage[main]/Main/Group[testgroup]/members: change from absent to ["crc"] failed: undefined method `sort' for :absent:Symbol
Notice: Finished catalog run in 0.26 seconds
```

I added a second commit to work around it, using a check for `.respond_to?(:sort)`

whopper:
Hey @ccaviness! Most of our team is out this week, so we'll have to wait a little while to take a closer look at these tests. We'll come back to it shortly, however, and help you get them running.

ccaviness:
Friendly no-rush ping. This code is working well for us, by the way.

whopper:
@ccaviness I just moved the Jira ticket onto into our triage bucket, meaning we should be able to pull this into one of our upcoming sprints. Stay tuned, and thanks again for your patience! :)

whopper:
@ccaviness just kidding, I had some time to look at this today! I agree that you are on the right path with your test. I'd suggest adding a few more to cover several other cases (like we do in `windows_adsi_spec.rb`. For example,

```
 13   describe '#members_insync?' do
 14     it 'should return true for same lists of unordered members' do
 15       expect(provider.members_insync?(['user1', 'user2'], ['user2', 'user1'])).to be_truthy
 16     end
 17
 18     it 'should return false when the group currently has no members' do
 19       expect(provider.members_insync?([], ['user2', 'user1'])).to be_falsey
 20     end
 21
 22     it "should return true for same lists of members irrespective of duplicates" do
 23       expect(provider.members_insync?(['user1', 'user2', 'user2'], ['user2', 'user1', 'user1'])).to be_truthy
 24     end
 25
 26     it "should return true when current and should members are empty lists" do
 27       expect(provider.members_insync?([], [])).to be_truthy
 28     end
 29   end
```

About the duplicate group names in `should`, I believe I've found one issue with this change as it stands. While testing, I'm seeing `#members_insync?` returning false when `should` has duplicates, causing puppet to try and update group membership every time:

```
lyokd3cqy68n68o:~ root# puppet apply -e "group { 'testgroup': members => ['test1', 'test2', 'test3','test3'] }"
Notice: Compiled catalog for lyokd3cqy68n68o.delivery.puppetlabs.net in environment production in 0.37 seconds
Notice: CURRENT: ["test1", "test2", "test3"]
Notice: SHOULD: ["test1", "test2", "test3", "test3"]
Notice: IN_SYNC? false
```

Perhaps duplicates should be stripped? `should.uniq.sort`?

The last thing I've noticed (unrelated to your change) is that we don't seem to be able to to remove users from groups:

```
lyokd3cqy68n68o:~ root# puppet resource group derp
group { 'derp':
  ensure  => 'present',
  gid     => '22',
  members => ['test1', 'test2', 'test3'],
}

lyokd3cqy68n68o:~ root# puppet apply -e "group { 'derp': members => ['test1', 'test2'] }"
Notice: Compiled catalog for lyokd3cqy68n68o.delivery.puppetlabs.net in environment production in 0.36 seconds
Notice: /Stage[main]/Main/Group[derp]/members: members changed 'test1,test2,test3' to 'test1,test2'

lyokd3cqy68n68o:~ root# puppet resource group derp
group { 'derp':
  ensure  => 'present',
  gid     => '22',
  members => ['test1', 'test2', 'test3'],
}
```

Have you encountered this behavior before?

whopper:
In addition, you'll need to add the `(PUP-2575)` to the beginning of your second commit to make Travis pass.

joshcooper:
@ccaviness Thank you for your contribution! When you get a chance could you review @whopper's comments and update this PR?

ccaviness:
I'll update the tests and the PR today.

@whopper I can remove users from groups just fine:
```
fox:~ root# puppet resource group remove
group { 'remove':
  ensure  => 'present',
  gid     => '23',
  members => ['root', 'daemon', 'crc'],
}
fox:~ root# puppet apply
group { 'remove':
  ensure  => 'present',
  members => ['root', 'daemon'],
}
Notice: Compiled catalog for 045627f2-3d66-46b1-95bc-1bc74154e1d3 in environment gmac_unstable in 0.58 seconds
Notice: /Stage[main]/Main/Group[remove]/members: members changed 'root,daemon,crc' to 'root,daemon'
Notice: Finished catalog run in 0.30 seconds
fox:~ root# puppet resource group remove
group { 'remove':
  ensure  => 'present',
  gid     => '23',
  members => ['root', 'daemon'],
}
```

This is on OS X 10.11 and (patched with this PR) puppet 3.8.2.

whopper:
@ccaviness interesting, and thanks for the version info!

ccaviness:
Ok, updated with some `uniq`-ing and more tests. I've also squashed the commits back into one, formatted so Travis should be happy.

whopper:
@ccaviness we discussed this in the PR triage today (https://www.youtube.com/watch?v=pjAPagAXJXM) and determined that we probably want one or two more tests that cover cases where `current` is `:absent`, `nil`, or other possible values that could be passed in if the group in question is set to, say, `ensure => absent` initially.

ccaviness:
I've added some logging locally:
```
    Puppet.info "current: #{current}"
    Puppet.info "should: #{should}"
```

It seems `members_insync?` doesn't get called when removing a group (ensuring absent) or creating a group (from absent to present), only when modifying an existing group (present to present).

```
fox:puppet crc$ sudo puppet apply --verbose
group { 'testgroup':
	ensure => 'present',
	members => ['root', 'daemon'],
}
Info: Loading facts
Notice: Compiled catalog for 045627f2-3d66-46b1-95bc-1bc74154e1d3 in environment gmac_unstable in 0.58 seconds
Info: Applying configuration version '1444758900'
Notice: /Stage[main]/Main/Group[testgroup]/ensure: created
Notice: Finished catalog run in 0.41 seconds
fox:puppet crc$ sudo puppet apply --verbose
group { 'testgroup':
	ensure => 'absent',
}
Info: Loading facts
Notice: Compiled catalog for 045627f2-3d66-46b1-95bc-1bc74154e1d3 in environment gmac_unstable in 0.59 seconds
Info: Applying configuration version '1444758970'
Notice: /Stage[main]/Main/Group[testgroup]/ensure: removed
Notice: Finished catalog run in 0.33 seconds
fox:puppet crc$ sudo puppet apply --verbose
group { 'testgroup':
	ensure => 'present',
	members => ['root', 'daemon'],
}
Info: Loading facts
Notice: Compiled catalog for 045627f2-3d66-46b1-95bc-1bc74154e1d3 in environment gmac_unstable in 0.59 seconds
Info: Applying configuration version '1444758996'
Notice: /Stage[main]/Main/Group[testgroup]/ensure: created
Notice: Finished catalog run in 0.41 seconds
fox:puppet crc$ diff lib/puppet/provider/group/directoryservice.rb /usr/lib/ruby/site_ruby/puppet/provider/group/directoryservice.rb 
fox:puppet crc$ sudo puppet apply --verbose
group { 'testgroup':
	ensure => 'present',
	members => ['root', 'daemon'],
}
Info: Loading facts
Notice: Compiled catalog for 045627f2-3d66-46b1-95bc-1bc74154e1d3 in environment gmac_unstable in 0.59 seconds
Info: Applying configuration version '1444759048'
Info: current: ["root", "daemon"]
Info: should: ["root", "daemon"]
Notice: Finished catalog run in 0.28 seconds
```

ccaviness:
@whopper ... meaning I'm not sure adding tests for those cases is necessary.

whopper:
Thanks for the quick response, @ccaviness. @joshcooper do you have any other feedback? I will be doing another pass at a functional review later today.

whopper:
Okay, did another review pass. Verified that ordering no longer matters. Like Josh said in the triage, though, it is kind of weird that duplicate users in the `members` attributes aren't handled gracefully deeper in the provider.

Currently, we see:

```
t5b0u6w1lyu4r65:~ root# puppet apply -e "group { 'group_a': members => ['test1', 'test1'] }"
Notice: Compiled catalog for t5b0u6w1lyu4r65.delivery.puppetlabs.net in environment production in 0.36 seconds
Notice: /Stage[main]/Main/Group[group_a]/members: members changed 'test1,test2' to 'test1,test1'
Notice: Applied catalog in 0.11 seconds

t5b0u6w1lyu4r65:~ root# puppet resource group group_a
group { 'group_a':
  ensure  => 'present',
  gid     => '21',
  members => ['test1', 'test2'],
}
```

So although the right thing is being done when checking if the two arrays are in sync, the duplicated users in the `ensure` parameter propagates its way all the way to the resource update, resulting in a misleading notice as seen above. I'm not sure if that's something this PR should address, or just an existig, separate issue.

ccaviness:
@whopper I would say that's a separate issue, but mostly because I have no idea how to fix it...

lib/puppet/provider/group/directoryservice.rb
```diff
@@ -9,4 +9,14 @@
   confine :operatingsystem => :darwin
   defaultfor :operatingsystem => :darwin
   has_feature :manages_members
+
+  def members_insync?(current, should)
+    return false unless current
+    if current.respond_to?(:sort) and should.respond_to?(:sort)
```
joshcooper: The `should` parameter should always be an array. If the `should` value is nil, e.g. the desired value wasn't specified in the manifest, then puppet will short circuit evaluation earlier in [`safe_insync?`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/property.rb#L256), since we're not managing that property.

If the `should` value is not an array, then the [default `insync?` logic fail hard](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/property.rb#L291).

Ideally, the `current` parameter would also always be an array. However, it appears that if the group currently has no members, then the provider has a bug and sends in current as `:absent`, instead of an empty array.

joshcooper:
This introduces a regression when specifying an empty array of members:

```
group { 'foobar':
  ensure => present,
  members => []
}
```

Previously, puppet would create a group with no members, and would remain idempotent:

```
130 ~/work/puppet (master)  $ sudo bundle exec puppet apply manifest.pp
Notice: Compiled catalog for arcturus.corp.puppetlabs.net in environment production in 0.03 seconds
Notice: /Stage[main]/Main/Group[foobar]/ensure: created
Notice: Applied catalog in 0.21 seconds
0 ~/work/puppet (master)  $ sudo bundle exec puppet apply manifest.pp
Notice: Compiled catalog for arcturus.corp.puppetlabs.net in environment production in 0.03 seconds
Notice: Applied catalog in 0.12 seconds
```

With this change, puppet thinks a change is necessary each time, since `members_insync?` is called with current `:absent` and should as an empty array.

```
0 ~/work/puppet (pr/4246)  $ sudo bundle exec puppet apply manifest.pp
Notice: Compiled catalog for arcturus.corp.puppetlabs.net in environment production in 0.04 seconds
Notice: /Stage[main]/Main/Group[foobar]/members: defined 'members' as ''
0 ~/work/puppet (pr/4246)  $ sudo bundle exec puppet apply manifest.pp
Notice: Compiled catalog for arcturus.corp.puppetlabs.net in environment production in 0.04 seconds
Notice: /Stage[main]/Main/Group[foobar]/members: defined 'members' as ''
```

I'm not sure if we should update the `members_insync?` method to handle case, or fix the provider to return an empty array of members (since the group does exist).

lib/puppet/provider/group/directoryservice.rb
ccaviness: So you'd like me to remove the `respond_to?` check on `should` entirely?

Let me try to re-arrange the logic.

ccaviness:
Ho-kay. How about this?

ccaviness:
I even added a test for `current` being `:absent`.

joshcooper:
Thank you @ccaviness for your contribution! Everything works great. There is still one unrelated issue to not being able to remove members [PUP-5334](https://tickets.puppetlabs.com/browse/PUP-5334), but the fix for that should be simple now that you have the `member_insync?` logic sorted out. I'll leave some comments on that other ticket.

PR #4248, opened 2015-09-15
(PUP-5328) Fix crash when using qualified lookups in interpolation

thallgren:
This commit fixes a typo in the hiera data provider and adds a test
to verify that interpolation expressions using qualified names such
as `%{os.name}` works as expected.

A test was also added to ensure that the hiera data provider does
not suffer from the problem described in HI-469.


joshcooper:
This was actually a fix for [PUP-5238](https://tickets.puppetlabs.com/browse/PUP-5238)

PR #4250, opened 2015-09-15
(PUP-5258) Unquote PUPPET_EXTRA_OPTS in SUSE init script

geoffnichols:
Prior to this commit, the Puppet service fails to start
on SLES11 with the error "Could not prepare for execution:
The puppet agent command does not take parameters".

This appears to be caused by an empty '' at the end of
Puppet's SUSE init script startproc call.

Since PUPPET_EXTRA_OPS is set to an empty string in the
default /etc/sysconfig/puppet, we update the SUSE
init script to remove the double quotes from
"${PUPPET_EXTRA_OPTS}".

### DIFF
--- a/ext/suse/client.init
+++ b/ext/suse/client.init
@@ -73,7 +73,7 @@ case "$1" in
                 rc_exit
             fi
         fi
-        startproc -f -w -p "${pidfile}" "${puppetd}" "${PUPPET_OPTS}" "${PUPPET_EXTRA_OPTS}" && touch "${lockfile}"
+        startproc -f -w -p "${pidfile}" "${puppetd}" "${PUPPET_OPTS}" ${PUPPET_EXTRA_OPTS} && touch "${lockfile}"
         # Remember status and be verbose
         rc_status -v
         ;;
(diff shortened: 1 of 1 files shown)


mwbutcher:
tested this change manually on sles 10 :+1:

joshcooper:
If this needs to go into 2015.2.x, we should target stable. If it can wait until 2015.3.x, then master is ok.

geoffnichols:
@joshcooper I think this can wait until 2015.3.x. This appears to be related to a feature introduced in https://github.com/puppetlabs/puppet/commit/e9597443d34ad728e7de78283826137c99a90854, which looks like is currently only on master.

demophoon:
:+1: from me, Validated by manually modifying the service script in a sles 10 install.

geoffnichols:
I'll create a PUP ticket for this.

joshcooper:
@geoffnichols It looks like the issue is still present in f0af8c08fc60c37d5b5a87ad6510a93e4197855c which is in 4.2.x. Can you confirm? Also, won't this show up in PE 2015.2.x testing, similar to the EL4 issue? /cc @jpartlow @demophoon

demophoon:
@joshcooper Yes, we ended up seeing it when testing installs on PE 2015.2.x when the puppet agent service was unable to start at the end of the installer.

geoffnichols:
@joshcooper I think we're only seeing an issue if the puppet-agent build includes https://github.com/puppetlabs/puppet/commit/e9597443d34ad728e7de78283826137c99a90854.

geoffnichols:
As follow-up to conversation with @joshcooper and @demophoon, I'm going to close this and retarget at stable.

PR #4251, opened 2015-09-16
(PUP-5228) Fixup ssh_authorized_key tests

whopper:
Previously, the ssh_authorized_key tests would append a new
key to authorized_keys with `echo.` On CentOS 4. simply appending
did not create a newline, and so the key was appended to the end
of the last key, meaning Puppet couldn't find or remove it.

This commit updates the test to add a newline to the file before
appending the new key.

In addition, there was an issue with the modifcation test in which the
key was appended improperly, and we then mistook Puppet's newly generated
key for a modified version of the old key, when in reality
'ensure=present' was just adding a brand new key.

### DIFF
--- a/acceptance/tests/resource/ssh_authorized_key/modify.rb
+++ b/acceptance/tests/resource/ssh_authorized_key/modify.rb
@@ -16,7 +16,7 @@ agents.each do |agent|
   on(agent, "cp #{auth_keys} /tmp/auth_keys", :acceptable_exit_codes => [0,1])
 
   step "(setup) create an authorized key in the #{auth_keys} file"
-  on(agent, "echo 'ssh-rsa mykey #{name}' >> #{auth_keys}")
+  on(agent, "echo '' >> #{auth_keys} && echo 'ssh-rsa mykey #{name}' >> #{auth_keys}")
 
   #------- TESTS -------#
   step "update an authorized key entry with puppet (present)"
@@ -30,6 +30,7 @@ agents.each do |agent|
   step "verify entry updated in #{auth_keys}"
   on(agent, "cat #{auth_keys}")  do |res|
     fail_test "didn't find the updated key for #{name}" unless stdout.include? "mynewshinykey #{name}"
+    fail_test "Found old key mykey #{name}" if stdout.include? "mykey #{name}"
   end
 
 end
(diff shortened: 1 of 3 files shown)


acceptance/tests/resource/ssh_authorized_key/modify.rb
```diff
@@ -30,6 +30,7 @@
   step "verify entry updated in #{auth_keys}"
   on(agent, "cat #{auth_keys}")  do |res|
     fail_test "didn't find the updated key for #{name}" unless stdout.include? "mynewshinykey #{name}"
+    fail_test "Found old key mykey #{name}" if stdout.include? "mykey #{name}"
```
whopper: This is necessary because in a case where the original key didn't exist, Puppet will just make a new 'myshinynewkey' and fool this test into passing. We want to make sure the old 'mykey' doesn't exist anymore. This came up due to the key being weirdly appended to the end of the previous key.

MikaelSmith:
:+1:

acceptance/tests/resource/ssh_authorized_key/query.rb
```diff
@@ -18,7 +18,7 @@
   on(agent, "cp #{auth_keys} /tmp/auth_keys", :acceptable_exit_codes => [0,1])
 
   step "(setup) create an authorized key in the #{auth_keys} file"
-  on(agent, "echo 'ssh-rsa mykey #{name}' >> #{auth_keys}")
+  on(agent, "echo '' >> #{auth_keys} && echo 'ssh-rsa mykey #{name}' >> #{auth_keys}")
```
MikaelSmith: Did this test exhibit any problems? Looks like you're just being safe by ensuring it's a properly formatted authorized_keys file.

whopper: Actually, I think there's an additional issue with the error message.

```
fail_test "found the ssh_authorized_key for #{name}" unless stdout.include? "#{name}"
```

This should probably say "Couldn't find the ssh_authorized_key for #{name}". Let me check this again - I'm not sure if it really should have been passing in CentOS 4.

whopper: @MikaelSmith Oh I see - we're skipping this test, as it's blocked on a different issue... looking into if it'll pass or not.

PR #4253, opened 2015-09-16
(docs) Fix inconsistencies in lookup function docs (DOCUMENT-414)

thallgren:
In this commit:
* Clarify use of options_hash
* Fix description of merge enum
* Clarify how override hash is used.
* Add table with mappings from Hiera to lookup

### DIFF
--- a/lib/puppet/functions/lookup.rb
+++ b/lib/puppet/functions/lookup.rb
@@ -19,21 +19,24 @@
 # The block, if present, is mutually exclusive to the `default_value` and will be called with the `name` used in the
 # lookup when no value is found. The value produced by the block then becomes the value produced by the lookup.
 #
-# The meaning of the parameters or content of the options hash is:
+# The meaning of each parameter is:
 #
 # * `name` - The name or array of names to lookup (first found is returned)
 # * `value_type` - The type to assert. Defaults to 'Data' See 'Type Specification' below.
-# * `default_value` - The default value if there was no value found (must comply with the data type)
-# * `override` - a hash with map from names to values that are used instead of the underlying bindings. If the name
-#   is found here it wins. Defaults to an empty hash.
-# * `default_values_hash` - a hash with map from names to values that are used as a last resort to obtain a value.
-#   Defaults to an empty hash.
-# * `merge` - A string of type Enum[unique, hash, merge] or a hash with the key 'strategy' set to that string. See
+# * `merge` - A string of type Enum[first, unique, hash, deep] or a hash with the key 'strategy' set to that string. Default is `first`. See
 #   'Merge Strategies' below.
+# * `default_value` - The default value if there was no value found (must comply with the data type)
+# * `options_hash`- A hash that may contain any of the above parameters, keyed by their name. In addition, it
+#   can also contain the following entries:
+#   * `override` - a hash with map from names to values that are used instead of the underlying bindings. If the `name`
+#     given to the lookup is a key in this hash, it wins, and the associated value is returned. Defaults to an empty hash.
+#   * `default_values_hash` - a hash with map from names to values that are used as a last resort to obtain a value.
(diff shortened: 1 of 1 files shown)


joshcooper:
thoughts @jtappa @nfagerlund @gguillotte?

lib/puppet/functions/lookup.rb
```diff
@@ -65,9 +70,30 @@
...
+# maps to the merge strategy:
+#
+#     Hiera merge + Config[:merge_behavior]	=> Merge Strategy
+#     ----------------------------------------------------------
+#     priority (default)                       first (default)
+#     array                                    unique
+#     hash + native                            hash
+#     hash + deeper                            deep (with Hashes)
+#     N/A                                      deep (with Arrays)
+#     hash + deep                              N/A
```
gguillotte: This can use a markdown table format, such as

~~~
# Hiera merge + Config[:merge_behavior]	| Merge strategy
# --------------------------------------|-------------------
# priority (default)                    | first (default)
# array                                 | unique
# hash + native                         | hash
# hash + deeper                         | deep (with hashes)
# N/A                                   | deep (with arrays)
# hash + deep                           | N/A
~~~

lib/puppet/functions/lookup.rb
```diff
@@ -65,9 +70,30 @@
...
+#
+# There's no 100% correspondence to the Hiera `hash + deeper` since deep will no longer fail when it encounters an
+# `Array` instance.
+#
+# The Hiera `deep` is not implemented. In the Hiera documentation it is considered "largely useless and should be avoided".
```
gguillotte: The Hiera documentation needs to either phrase or document that better, and we certainly don't need to repeat that "largely useless and should be avoided" phrasing here. 

If something isn't implemented, why are we documenting it?

lib/puppet/functions/lookup.rb
```diff
@@ -65,9 +70,30 @@
...
+#     ----------------------------------------------------------
+#     priority (default)                       first (default)
+#     array                                    unique
+#     hash + native                            hash
+#     hash + deeper                            deep (with Hashes)
+#     N/A                                      deep (with Arrays)
+#     hash + deep                              N/A
+#
+# There's no 100% correspondence to the Hiera `hash + deeper` since deep will no longer fail when it encounters an
+# `Array` instance.
```
gguillotte: s/100% correspondence/equivalent/
s/to the Hiera `hash + deeper`/to Hiera's `hash + deeper` option/
s/will no longer fail/no longer fails/

gguillotte:
I think @nfagerlund is part of some bigger ongoing conversations around our documentation of Hiera data in modules, so I'd defer to him on some of the details here.

lib/puppet/functions/lookup.rb
thallgren: I think the reason is that is is implemented and documented i Hiera. This is an attempt to answer the question, why isn't it implemented here.

lib/puppet/functions/lookup.rb
thallgren: I made some attempts to find the preferred format for how to present tables in this kind of documentation. Are there some guidelines published somewhere? It would be very helpful.

nfagerlund: https://confluence.puppetlabs.com/display/DOCS/Markdown+Tips+for+Puppet+Labs+Stuff#MarkdownTipsforPuppetLabsStuff-Tables

nfagerlund:
I'm cool with merging this for now, and we'll be coming back and revising it again soon anyway. We have an additional revision of the lookup function tracked as DOC-2225, which is related to the broader Puppet lookup docs (tracked as DOC-2311).

nfagerlund:
This is superseded by #4455. It should probably be closed.

PR #4254, opened 2015-09-16
(PUP-5026) Ensure that parser tokens are reported with position

thallgren:
This commit fixes the error method in the ParserSupport class so that
it correctly reports the position of TokenValue instances such as
COMMA when such tokens are misplaced.


hlindberg:
A unit test would be nice.

PR #4256, opened 2015-09-16
(PUP-5258) Unquote PUPPET_EXTRA_OPTS in SUSE init script

geoffnichols:
Prior to this commit, the Puppet service fails to start
on SLES11 with the error "Could not prepare for execution:
The puppet agent command does not take parameters".

This appears to be caused by an empty '' at the end of
Puppet's SUSE init script startproc call.

Here we update the SUSE init script to remove the
double quotes from "${PUPPET_EXTRA_OPTS}".

### DIFF
--- a/ext/suse/client.init
+++ b/ext/suse/client.init
@@ -73,7 +73,7 @@ case "$1" in
                 rc_exit
             fi
         fi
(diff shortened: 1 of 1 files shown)


geoffnichols:
Replaces https://github.com/puppetlabs/puppet/pull/4250.

shrug:
:+1:

PR #4257, opened 2015-09-17
(PUP-5057) Add test for managing puppet and mcollective services

whopper:
This commit adds a new acceptance test to ensure that the Puppet
and Mcollective services can be started, stopped and refreshed
in a number of scenarios.

### DIFF
--- a/acceptance/lib/puppet/acceptance/service_utils.rb
+++ b/acceptance/lib/puppet/acceptance/service_utils.rb
@@ -14,6 +14,57 @@ module Puppet
         suitable = on(host, "#{ruby} -e \"require 'puppet'; puts Puppet::Type.type(:service).provider(:systemd).suitable?\"" ).stdout.chomp
         suitable == "true" ? true : false
       end
+
+      # Alter the state of a service using puppet apply.
+      # @param host [String] hostname.
+      # @param service [String] name of the service.
+      # @param property [String] name of the attribute to be changed.
+      # @param value [String] value which the property should be set.
+      # @return None
+      def ensure_service_on_host(host, service, property, value)
+        manifest = %Q{
+          service { '#{service}':
+            #{property} => '#{value}'
+          }
+        }
+        # the process of creating the service will also start it
+        # to avoid a flickering test from the race condition, this test will ensure
+        # that the exit code is either
+        #   2 => something changed, or
+        #   0 => no change needed
+        on host, puppet_apply(['--detailed-exitcodes', '--verbose']),
+          {:stdin => manifest, :acceptable_exit_codes => [0, 2]}
+        # ensure idempotency
+        on host, puppet_apply(['--detailed-exitcodes', '--verbose']),
+          {:stdin => manifest, :acceptable_exit_codes => [0]}
+      end
+
+      # Checks that the status of a service is as expected.
+      # @param host [String] hostname.
+      # @param service [String] name of the service.
+      # @param expected_status [String] expected service status.
+      # @return None
+      def assert_service_status_on_host(host, service, expected_status)
+        on(host, puppet_resource('service', service)) do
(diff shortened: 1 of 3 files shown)


whopper:
First pass: successfully ran on: osx-1011, centos-4, redhat-7, debian-8, sles-12 and sles-11.

Legitimate failures in Solaris 10 (https://tickets.puppetlabs.com/browse/PUP-5262) and Win2012R2. The Solaris service provider returns before finishing the restart of the service, so immediately checking its status results in 'ensure => stopped' when really it's just in the middle of restarting. Starting Mco on Windows currently fails - see @joshcooper's comment on https://tickets.puppetlabs.com/browse/PUP-5057.

jpartlow:
I like what is currently here; I wonder if it is worthwhile to also explicitly test `puppet resource service ensure=running/stopped` calls.

I realize that they are basically just applies under the hood, but these calls are used both in Beaker and in some of the PE scripts, and I would hate to trip over something just because the puppet resource variation is a little different...

whopper:
@jpartlow no prob! I'll add some more tests.

acceptance/tests/resource/service/puppet_mcollective_service_management.rb
```diff
@@ -0,0 +1,44 @@
+test_name "Puppet and Mcollective services should be manageable with Puppet"
+
+confine :except, :platform => 'windows'
+confine :except, :platform => 'solaris' # See PUP-5262
```
er0ck: if this is later to be fixed on solaris due to this ticket, perhaps you could use an `expect_failure()` somewhere so that it starts failing when the fix comes in and the fix author is reminded to re-enable this test?

acceptance/lib/puppet/acceptance/service_utils.rb
```diff
@@ -14,6 +14,57 @@ def supports_systemd? (host)
...
+        }
+        # the process of creating the service will also start it
+        # to avoid a flickering test from the race condition, this test will ensure
+        # that the exit code is either
+        #   2 => something changed, or
+        #   0 => no change needed
+        on host, puppet_apply(['--detailed-exitcodes', '--verbose']),
+          {:stdin => manifest, :acceptable_exit_codes => [0, 2]}
+        # ensure idempotency
+        on host, puppet_apply(['--detailed-exitcodes', '--verbose']),
```
er0ck: i think this part is really the value add.  idempotency.  
we 'use' resource service to ensure services for our test use in various cases.  we also have tests on enable and block.

i'm unclear on if puppet resource should be tested for idempotency, because, as josh said, they are just apply, underneath.

joshcooper: Unfortunately I've been bitten by cases where `puppet resource` behaves differently than `apply`, e.g. 1e8a2cd86bafb4c73d699cea67e9605e78ac0d0a, so I'm :+1: for testing both.

acceptance/tests/resource/service/puppet_mcollective_service_management.rb
joshcooper: I forgot about `expect_failure`, I need to use that more.

er0ck: we all should.  i've been meaning to socialize it more.   it's amazing.  Chris Cowell did it one day out of the blue and PRd it into beaker.

kylog: +1 on using that more and socializing that more.

whopper:
Alright, @er0ck @joshcooper @jpartlow updated a bunch of stuff:

- Added tests for `puppet resource` enabling and stopping
- Added an expect_failure for solaris, but on Windows the mco issue is causing the test to straight up error which isn't being caught by the expect failure and was driving me crazy, so I left it as a confine.

er0ck:
LGTM.  worked for me on ubuntu14

PR #4258, opened 2015-09-17
(PUP-1963) allow dependencies on/for generated resources

branan:
Previously, generated resources could not have edge metaparams or
participate in autorequire, since they were created AFTER those
transformations had occured on the catalog. This changes resource
generation to occur before those transformations.

Because of this, the containment edges for generated resources now end
up in the final relationship graph. This caused issues with purged
resources, since the purge check now sees that a whit depends on the
purged resource. This was resolved by excluding whits from the purge
safety check. Since whits can only be generated by containment edges,
and generated resources previously could not have containment edges,
this is not a breaking change.

In order to maintain manifest ordering, it is necessary to add the
generated resources into the catalog adjacent to the generating
resource. This is achieved with new helpers on the catalog class to
handle ordered insertions.

Existing spec tests handle verifying manifest ordering works correctly
for generated resources, and integration tests for things like
resource purging handle reasonably complex generation cases. New tests
have been added only for the new edges that are being generated after
this commit.

### DIFF
--- a/lib/puppet/resource/catalog.rb
+++ b/lib/puppet/resource/catalog.rb
@@ -73,6 +73,25 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
     return a
   end
 
+  def add_resource_before(other, *resources)
+    resources.each do |resource|
+      other_title_key = title_key_for_ref(other.ref)
+      idx = @resources.index(other_title_key)
+      raise ArgumentError, "Cannot add resource #{resource.ref} before #{other.ref} because #{other.ref} is not yet in the catalog" if idx.nil?
+      add_one_resource(resource, idx)
+    end
+  end
+
+  def add_resource_after(other, *resources)
+    resources.each do |resource|
+      other_title_key = title_key_for_ref(other.ref)
+      idx = @resources.index(other_title_key)
+      raise ArgumentError, "Cannot add resource #{resource.ref} after #{other.ref} because #{other.ref} is not yet in the catalog" if idx.nil?
+      add_one_resource(resource, idx+1)
+    end
+  end
+
+
   def add_resource(*resources)
     resources.each do |resource|
       add_one_resource(resource)
@@ -86,13 +105,13 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
     adjacent(resource, :direction => :in)[0]
   end
 
-  def add_one_resource(resource)
+  def add_one_resource(resource, idx=-1)
     title_key = title_key_for_ref(resource.ref)
     if @resource_table[title_key]
       fail_on_duplicate_type_and_title(resource, title_key)
     end
 
-    add_resource_to_table(resource, title_key)
+    add_resource_to_table(resource, title_key, idx)
     create_resource_aliases(resource)
 
     resource.catalog = self if resource.respond_to?(:catalog=)
@@ -100,9 +119,9 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
   end
   private :add_one_resource
 
-  def add_resource_to_table(resource, title_key)
+  def add_resource_to_table(resource, title_key, idx)
     @resource_table[title_key] = resource
-    @resources << title_key
+    @resources.insert(idx, title_key)
   end
   private :add_resource_to_table
 
--- a/lib/puppet/transaction/additional_resource_generator.rb
+++ b/lib/puppet/transaction/additional_resource_generator.rb
@@ -5,6 +5,8 @@
 #
 # @api private
 class Puppet::Transaction::AdditionalResourceGenerator
+  attr_writer :relationship_graph
+
   def initialize(catalog, relationship_graph, prioritizer)
     @catalog = catalog
     @relationship_graph = relationship_graph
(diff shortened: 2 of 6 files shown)


lib/puppet/transaction/additional_resource_generator.rb
```diff
@@ -11,6 +11,10 @@ def initialize(catalog, relationship_graph, prioritizer)
     @prioritizer = prioritizer
   end
 
+  def relationship_graph=(relationship_graph)
```
haus: Why not add attr_writer for :relationship_graph instead of making your own setter?

branan: because I did not know that exists

branan:
I have NO IDEA how this could have failed on Windows, but I just pushed up an ammended commit, so we'll find out if it's reproducible shortly

kylog:
wtf appveyor?

branan:
Awesome. I guess I know what I'm doing today. Yay windows.

branan:
@kylog of course that test passes for me in my windows env :/

I can try to push up a version of this that just runs the failing test with some extra debug printing, we might be able to see what's up that way. Sounds not fun :(

branan:
Actually, looks like it passes when run in isolation but fails when I run the whole suite. Bisecting now.

branan:
OK, I figured it out. My new test created a type with the same name as a type made in another test. When they were ordered appropriately, my test would fail, since the type wouldn't have the expected properties

branan:
This should be ready for review now

kylog:
> My new test created a type with the same name as a type made in another test. When they were ordered appropriately, my test would fail, since the type wouldn't have the expected properties

Well, that sounds like fun.

bmjen:
Tested this on concat 2.0.1 and refreshes were appropriately triggered! :+1:

branan:
@bmjen :metal: Thanks for checking it ou!

lib/puppet/resource/catalog.rb
```diff
@@ -86,13 +108,12 @@ def container_of(resource)
     adjacent(resource, :direction => :in)[0]
   end
 
-  def add_one_resource(resource)
-    title_key = title_key_for_ref(resource.ref)
+  def add_one_resource(resource, title_key, idx)
```
MikaelSmith: Why was title_key added to the call signature, instead of retrieved from the resource? I don't see it needed anywhere else, it seems like the old behavior would work for add_resource_before and add_resource_after.

lib/puppet/transaction/additional_resource_generator.rb
```diff
@@ -102,16 +103,69 @@ def add_resources(generated, resource)
     end
   end
 
-  def add_resource(res, parent_resource, priority)
+  def add_resource(res, parent_resource, priority=nil)
```
MikaelSmith: Is the priority argument still used anywhere? If it's preserved only for backward compatibility, add a comment.

lib/puppet/resource/catalog.rb
MikaelSmith: A default argument for idx when we're appending seems useful too.

lib/puppet/transaction/additional_resource_generator.rb
```diff
@@ -22,11 +24,10 @@ def generate_additional_resources(resource)
     generated = [generated] unless generated.is_a?(Array)
     generated.collect do |res|
       @catalog.resource(res.ref) || res
-    end.each do |res|
-      priority = @prioritizer.generate_priority_contained_in(resource, res)
-      add_resource(res, resource, priority)
+    end.reverse_each do |res|
```
MikaelSmith: Not sure I understand this change. I brief description of why we do this in reverse would be useful.

lib/puppet/resource/catalog.rb
branan: I ended up needing the title_key in my functions that call this, so I wanted to just pass it through instead of fetching it multiple times.

MikaelSmith: I don't see that happening. Which functions are you talking about?

lib/puppet/transaction/additional_resource_generator.rb
```diff
@@ -102,16 +103,69 @@ def add_resources(generated, resource)
...
+      target = parent.to_resource
+    else
+      source = parent
+      target = child.to_resource
+    end
+
+    # For each potential relationship metaparam, check if parent or
+    # child references the other. If there are none, we should add our
+    # edge.
+    edge_exists = Puppet::Type.relationship_params.map { |param|
```
MikaelSmith: You could use `any?` with a block to do the same thing. It'll stop when it finds the first true, rather than iterating over every element in `map`.

lib/puppet/transaction/resource_harness.rb
```diff
@@ -99,10 +99,15 @@ def perform_changes(resource, context)
   def allow_changes?(resource)
     if resource.purging? and resource.deleting? and deps = relationship_graph.dependents(resource) \
             and ! deps.empty? and deps.detect { |d| ! d.deleting? }
-      deplabel = deps.collect { |r| r.ref }.join(",")
-      plurality = deps.length > 1 ? "":"s"
-      resource.warning "#{deplabel} still depend#{plurality} on me -- not purging"
-      false
+      real_deps = deps.reject { |r| r.class.name == :whit }
```
MikaelSmith: Could've also done this in the if with something like `deps = relationship_graph.dependents(resource).reject { |r| r.class.name == :whit }`, and avoid the extra if block. That whole thing could be cleaned up though.

lib/puppet/transaction/additional_resource_generator.rb
branan: Before, when stuff was being added to the evaluation graph, the new resources were given sequential priorities just after the parent resource.

Now they're being added to the catalog in sequence, each one immediately after the parent resource. This causes them to be reversed. Keeping their order would involve making the "Add this resource to the catalog" step way more complicated.

Order doesn't matter for any built-in resources, but I didn't want to change behavior

lib/puppet/transaction/resource_harness.rb
branan: I *think* the idea with doing deps as part of the if logic was to short-circuit walking the graph if we aren't purging. I agree this whole thing could use a cleanup, but I kinda feel it's outside the scope of this PR

lib/puppet/transaction/additional_resource_generator.rb
branan: Oh, I didn't realize any could take a block (and am maybe also used to lazy clojure sequences :wink: )

I can change this

lib/puppet/resource/catalog.rb
branan: `add_resource_before` and `add_resource_after`, but it looks like they don't actually need the title key. This might be cruft from an older version of the code, TBH. I'll clean it up

lib/puppet/transaction/additional_resource_generator.rb
branan: yeah, this function is now shared by two different code paths - one that adds just to the catalog (for generate) and one that adds to the relationship graph (eval_generate)

branan:
@MikaelSmith updated

MikaelSmith:
:+1:

PR #4260, opened 2015-09-17
(PUP-5087) Add site top container

hlindberg:
This adds a site expression. It is treated analogous to
a node expression - but it contains the resources for all cross
node resources as apposed to the resources for a specific node.

This changes the environment compiler to prune the resulting catalog
from unwanted resources instead of attempting to stop them from being
added.

This implementation allows both the pre-site-expression style (in top
scope) and that application/cross node resources are in a
site-expression. This is expected to change.


lib/puppet/parser/environment_compiler.rb
```diff
@@ -73,6 +98,35 @@ def add_resource(scope, resource)
...
   end
 
+
+  def evaluate_ast_node()
+    # Do nothing, the environment catalog is not built for a particular node.
+  end
+
+  # Evaluates the site - the top container for the environment catalog
+  # The site contain behaves analogous to a node - for the environment catalog node expressions are ignored
+  # as the result is cross node. The site expression serves as a container for everything that is across
```
thallgren: Would be helpful with a comma here. I assume you mean that "node expressions" are ignored in the "environment catalog"? Other interpretations can be "catalog node expressions" in the "environment" or "expressions" in the "environment catalog node" :-)

lib/puppet/parser/environment_compiler.rb
```diff
@@ -82,4 +136,12 @@ def evaluate_applications
       end
     end
   end
+
+  # Overrides the regular compiler to be able to return the list of resources after a prune
+  # has taken place in the graph representation. Before a prune, the list is the same as in the regular
+  # compiler
+  #
+  def resources
+    @pruned_resources || super
```
thallgren: Perhaps not ideal to have a attribute named @resources and an accessor method with the same name that doesn't return that attribute. Are both @resource and @pruned_resources needed or could the prune_catalog replace the former?

thallgren:
I don't find any tests that verifies that the prune_catalog does what it's supposed to.

hlindberg:
The existing tests will fail if the pruning is not done correctly. There are no explicit unit tests for the pruning method (on purpose as tests are for the expected compilation result, not how it arrived there)

lib/puppet/parser/environment_compiler.rb
hlindberg: it is indeed an override of of the @resource attribute - the alternative would be to poke a new value into @resources (which is really horrible as the super class declares it). We could do this with a protected setter

spec/unit/appmgmt_spec.rb
```diff
@@ -49,42 +49,152 @@ def compile_to_env_catalog(string)
         cons { two: host => ahost, consume => Cap[cap] }
```
hlindberg: @lutter This Cons resource ends up not having $host set in the environment catalog. I can see it evaluating this part `"cons { two: consume => Cap[cap] }"` which parser says comes from here.

spec/unit/appmgmt_spec.rb
```diff
@@ -105,76 +215,113 @@ def compile_to_env_catalog(string)
       expect(cons).not_to be_nil
       expect(cons[:consume].ref).to eq("Cap[cap]")
     end
+
+    context "when using a site expression" do
+      it "includes components and capability resources" do
+        require 'debugger'; debugger
+        catalog = compile_to_env_catalog(MANIFEST_WITH_SITE).to_resource
```
hlindberg: @lutter This is the test that fails when the resource mentioned above is validated.

PR #4261, opened 2015-09-17
(maint) Set up logging for lookup application

HAIL9000:
Prior to this commit, logging was not being registered for the
lookup application which meant that errors would not appear on
standard out. Add a setup method which will register logging so
errors appear accordingly.


hlindberg:
There is a test failing because it now gets more text in the error message.

The commit message says something like errors did not appear on standard out - it is a bit misleading as they should in fact appear on standard error; suggest saying "appear on the console" to be correct but unspecific...

HAIL9000:
@hlindberg Updated!

PR #4263, opened 2015-09-18
(PUP-4890) Add code_id to the catalog

joshcooper:
Adds a code_id accessor to the catalog, updates the api docs, schema, and tests.

Updates the compiler terminus' find method to add the code_id from the request to the catalog it returns.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -48,6 +48,9 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
     node.trusted_data = Puppet.lookup(:trusted_information) { Puppet::Context::TrustedInformation.local(node) }.to_h
 
     if catalog = compile(node)
+      if code_id = request.options[:code_id]
+        catalog.code_id = code_id
+      end
       return catalog
     else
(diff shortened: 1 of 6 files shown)


joshcooper:
@kylog @cprice404 @lutter @senior I hear you want a code_id?

kylog:
:+1:

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -48,6 +48,9 @@ def find(request)
     node.trusted_data = Puppet.lookup(:trusted_information) { Puppet::Context::TrustedInformation.local(node) }.to_h
 
     if catalog = compile(node)
+      if code_id = request.options[:code_id]
```
lutter: Sorry for not seeing this earlier: it looks like with this setup, `code_id` is not available during compilation, though we need that so we can query PDB with it when we look up consumed resources.

joshcooper: That's a bummer. @lutter can you point me to code where you need access to it?

lutter: Left a comment on PUP-4890 explaining that

PR #4264, opened 2015-09-18
(PUP-5076) Fail non app components in site catalog

hlindberg:
This stands on PUP 5087 (contained in this PR since it is not yet merged).


lib/puppet/parser/environment_compiler.rb
```diff
@@ -40,20 +46,62 @@ def compile
...
+    the_main_class_resource = @catalog.resource('Class', '')
+    the_site_resource = @catalog.resource('Site', 'site')
+
+    # Get downstream vertexes returns a hash where the keys are the resources and values nesting level
+    rooted_in_main = @catalog.downstream_from_vertex(the_main_class_resource).keys
+    if the_site_resource
+      keep_from_site = @catalog.downstream_from_vertex(the_site_resource).keys
+      keep_from_site << the_site_resource
     else
-      unless (resource.is_application_component? || resource.is_capability?)
```
hlindberg: @lutter - input here?

lutter: Yes, that seems right to me; inside the `site` statement we want, after evaluation, applications and their components. I am not entirely clear whether containment is set up right, but assuming it is, we should have inside `site` something like `app_instance -> component -> other_stuff` - we want the app_instance in there and the component; the other_stuff is not an error but we don't care about it. I am not sure whether that happens, but during environment compilation it's sufficient to evaluate just to the point where we know the type/title of components and the type/title of capability resources referenced by `export`, `consume` or any of the standard relationship metaparams. IOW, we do not need to actually evaluate the bodies of components.

lib/puppet/parser/environment_compiler.rb
```diff
@@ -40,20 +46,62 @@ def compile
...
-        raise ArgumentError, "Only components are allowed inside applications, not #{resource.ref}"
+      # keep_from_site is populated with any App resources. If a site expression is used, those outside the
+      # site expression are dropped.
+      # This logic essentially keeps the "current behavior" when not using the site expression.
```
lutter: I actually think we should not try and keep the current behavior - put your application instances into `site` expressions and nowhere else.

PR #4265, opened 2015-09-18
(PUP-4932) Deprecate --cfacter option

whopper:
This commit deprecates the --cfacter setting, as it is no longer
needed with native Facter being shipped in AIO by default.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -21,7 +21,8 @@ module Puppet
         :default => false,
         :type    => :boolean,
         :desc    => 'Whether to enable a pre-Facter 3.0 release of native Facter (distributed as
-          the "cfacter" package). This is not necessary if Facter 3.0 or later is installed.',
+          the "cfacter" package). This is not necessary if Facter 3.0 or later is installed.
+          This setting is deprecated, as Facter 3 is now the default in puppet-agent.',
         :hook    => proc do |value|
           return unless value
           raise ArgumentError, 'facter has already evaluated facts.' if Facter.instance_variable_get(:@collection)
--- a/lib/puppet/feature/cfacter.rb
+++ b/lib/puppet/feature/cfacter.rb
@@ -2,6 +2,7 @@ require 'facter'
 require 'puppet/util/feature'
 
 Puppet.features.add :cfacter do
+  Puppet.deprecation_warning("The cfacter setting is deprecated. You can use Facter 3 and higher without this setting.")
   begin
     require 'cfacter'
 


whopper:
@joshcooper is this wording reasonable? I wasn't sure what all of the possible scenarios would be that someone could be using this option, so maybe mentioning 'puppet-agent' wouldn't always make sense?

Also, I wanted to add a spec test for this, but we don't have any for features, and I couldn't seem to get it to fire off in `spec/unit/default_spec.rb`. I've tested it in a puppet-agent build, though, and the deprecation goes off as expected.

joshcooper:
> I suppose someone could be running puppet not within puppet-agent (from source, maybe?)

yeah, or as a gem, e.g. vagrant+puppet. They'll get ruby facter (2.x), and `cfacter` doesn't make sense there either.

@nfagerlund thoughts on the deprecation warning wording?

For tests, you could add a test to the `agent` or `apply` application to make sure a deprecation is issued, like we do about webrick in `spec/unit/application/master_spec.rb`? Or not, I could go either way.

nfagerlund:
For the setting description, I'm fine with what @whopper changed it to. 

For the deprecation warning, what do you think about `The cfacter setting is deprecated. You can use Facter 3 and higher without this setting.`

whopper:
@nfagerlund Ah yeah, much better - Updated.
@joshcooper Working on the test, but having some annoying 'Facter has already evaluated facts' issues..


 I also removed my original `deprecated => completely` from the setting in `defaults.rb`, as that causes a second deprecation warning to fire off, and I'm not sure that we commonly use that option anyway.

joshcooper:
@whopper sorry I think I just internalized what you said earlier about other use cases. So if someone is running puppet from source/gem, which pulls in ruby facter, but they really do want to use native facter, e.g. it's in their PATH, LOAD_PATH, etc. With this PR, they would get a deprecation warning even though they really do want native facter and the cfacter switch is the only way to tell puppet to use native facter:

```
0 ~/work/puppet (master)  $ be puppet apply -e "notify { 'foo': }" --cfacter
Warning: The cfacter setting is deprecated. You can use Facter 3 and higher without this setting.
   (at /Users/josh/work/puppet/lib/puppet/feature/cfacter.rb:5:in `block in <top (required)>')
Error: Could not initialize global default settings: cfacter version 0.2.0 or later is not installed.
```

But perhaps this isn't even possible anymore? I don't know if native facter still provides the `cfacter` shim to load native facter. @peterhuene,  @kylog?

kylog:
Facter 3 does not provide the cfacter.rb shim and cfacter as a separate project is abandoned at this point, so this setting doesn't make sense.

We don't have great documentation on how to run puppet from source with facter 3, but I think that should be addressed outside of this ticket (and shouldn't involve the `--cfacter` setting). 

Btw, not the same concern but somewhat related: https://github.com/puppetlabs/puppet-agent/pull/289

kylog:
@joshcooper or @whopper: while your'e thinking cfacter-is-gone thoughts, I PR'd the file_paths spec also: https://github.com/puppetlabs/puppet-specifications/pull/49

PR #4266, opened 2015-09-18
Revert "(maint) Optimize adding tags"

HAIL9000:
This reverts commit 0412515e2f9d75f00687bc5c01915da5a51c58e3.

This change to tag filtering meant that a single tag like foo::bar
would be split on the :: seperator and expaned into:
['foo::bar', 'foo', 'bar']. This meant that an agent would sync more
resources than the used intended when tag filtering is in effect.

In order to fix this problem, revert the commit that initially made
this change to filtering behavior.

### DIFF
--- a/lib/puppet/util/tagging.rb
+++ b/lib/puppet/util/tagging.rb
@@ -1,26 +1,20 @@
 require 'puppet/util/tag_set'
 
 module Puppet::Util::Tagging
-  ValidTagRegex = /^\w[-\w:.]*$/
-
-  # Add a tag to the current tag set.
-  # When a tag set is used for a scope, these tags will be added to all of
-  # the objects contained in this scope when the objects are finished.
-  #
+  # Add a tag to our current list.  These tags will be added to all
+  # of the objects contained in this scope.
   def tag(*ary)
     @tags ||= new_tags
 
-    ary.flatten.each do |tag|
-      name = tag.to_s.downcase
-      if name =~ ValidTagRegex
-        @tags << name
-        name.split("::").each do |section|
-          @tags << section
-        end
-      else
-        fail(Puppet::ParseError, "Invalid tag '#{name}'")
-      end
+    qualified = []
+
+    ary.collect { |tag| tag.to_s.downcase }.each do |tag|
+      fail(Puppet::ParseError, "Invalid tag #{tag.inspect}") unless valid_tag?(tag)
+      qualified << tag if tag.include?("::")
+      @tags << tag unless @tags.include?(tag)
     end
+
+    handle_qualified_tags( qualified )
   end
 
   # Answers if this resource is tagged with at least one of the given tags.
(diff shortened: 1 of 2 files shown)


HAIL9000:
@kylog @hlindberg This seems to fix the problem, but I'm not sure of any other potential repercussions of reverting this change.

hlindberg:
Yeah, but is that what we want to do (revert?) - If I understand the issue correctly it is the search that is wrong because the change ended up splitting what the user searched for (which it did not use to do).

kylog:
If I'm following along correctly, what happened with 0412515e was that `Puppet::Util::Tagging#handle_qualified_tags` was factored out of existence, rendering the override supplied at https://github.com/puppetlabs/puppet/blob/3.x/lib/puppet/transaction.rb#L353-L358 as orphaned code (later removed in 8848171b). 

The behavior before that was that there was different behavior on the compile side and the application side, thanks to transaction.rb overriding what handle_qualified_tags does.

If that's correct, I think this would work in 3.x, but would not work if we merged up to stable (because stable no longer has `handle_qualified_tags` in transaction.rb. Put another way, we'd want to revert 8848171b when this was merged up to stable.

If that all makes sense to everybody, my question would be whether we are losing any performance advantage that 0412515e2f may have introduced. The comment says "overall savings" but doesn't go into details.

Is there something we could benchmark that would shed light on that question?

hlindberg:
the revert will have performance impact

hlindberg:
the fix is really simple - add a method that answers if a tag should be split, override it in transaction.rb and make the splitting conditional - that adds one call per tag which is reasonable

kylog:
> the revert will have performance impact

Gotcha, but can we quantify / mitigate?

kylog:
So I think you're saying that the overall savings came from avoiding building up this intermediate `qualified` array, and we could bypass that by just adding the method you're describing.

lib/puppet/util/tagging.rb
```diff
@@ -1,26 +1,20 @@
...
-  #
+  # Add a tag to our current list.  These tags will be added to all
+  # of the objects contained in this scope.
   def tag(*ary)
     @tags ||= new_tags
 
-    ary.flatten.each do |tag|
-      name = tag.to_s.downcase
-      if name =~ ValidTagRegex
-        @tags << name
```
hlindberg: This line adds the qualified name as is

lib/puppet/util/tagging.rb
```diff
@@ -1,26 +1,20 @@
...
-        name.split("::").each do |section|
-          @tags << section
-        end
```
hlindberg: These three lines should be conditional on the return of a new `split_qualified_tags?` method. This method returns `true` for this class. In the Transaction class, where it used to override `handle_qualified_tags`, override the `split_qualified_tags?` method and return `false`

hlindberg:
Performance was gained by not doing redundant work (searching with include, and then splitting) mostly by using a Set containing all tags (uniqueness guaranteed faster than checking with include).

HAIL9000:
Closing this in favor of a PR that makes the needed changes without reverting the entire commit :+1:

petems:
Adding a comment with [PUP-4495](https://tickets.puppetlabs.com/browse/PUP-4495) so it's easy to find in when searching PR's :+1:

PR #4268, opened 2015-09-21
(PUP-4695) Fix for hiera_in_templates acceptance test

segatspiretech:
As originally checked in, hiera_in_templates.rb caused failures in CI for
agents whose OSs were Windows or Ubuntu.  The OS family fact, used for
creating hiera data filese, was resolved for the master only.  If the
agent had an OS different from the master's (e.g. Ubuntu or Windows), it
would not find its corresponding hiera file.  The test now polls the agents
and creates a hiera data file for each agent OS family.

In addition, the owner and group attributes were removed, because 'root'
is not recongnized in Windows.

Original check in comment:

Until now we had no acceptance tests for verifying behavior of hiera
function calls (hiera(), hiera_array(), hiera_hash(), and hiera_include())
from within a template (erb or epp).

This test calls hiera functions from with in an erb template and from within
an epp template.  The master is configured with a manifest.  The client
applies the manifest using the puppet agent -t call.  The file created by
templates are examined to determine whether the hiera function calls returned
the expected values.

hiera_include() is called from the epp template.  The included sub-modules
each create a file that is then checked for correctness.


acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -222,10 +220,8 @@ class #{@module_name}::mod_production {
   content => "
 class #{@module_name}::mod_fqdn {
   notify{\\"module mod_fqdn invoked.\\\\n\\":}
-  file {'/#{@resultdir}/mod_fqdn':
+  file {'#{@resultdir}/mod_fqdn':
     ensure  => 'file',
-    owner   => 'root',
-    group   => 'root',
```
er0ck: trailing whitespace on line 234 should be removed

er0ck:
needs a rebase, i guess because this was removed from master after this feature branch.  
besides that it runs and passes for me on windows and ubuntu
:+1:

segatspiretech:
@er0ck I think the rebasing is complete now.  Please review.  Thanks.

er0ck:
@segatspiretech please squash down your commits.

segatspiretech:
Something's not working.  Too much effort to untangle. Starting over.

PR #4271, opened 2015-09-21
(maint) Remove F20 from build targets

melissa:
Fedora 20 is now EOL, so we should no longer be building packages for
it.


kylog:
Wheee, fun appveyor failure! But, unrelated to this PR.

PR #4273, opened 2015-09-22
DO NOT MERGE (PUP-5272) Constrain attribute validation to checksum

johnduarte:
This commit updates the `resource/file/source_attribute.rb`
test to validate that the checksum of the files under test have
not changed. This will allow other attribute changes
(e.g. mtime, ctime) to change and pass.

### DIFF
--- a/acceptance/tests/resource/file/source_attribute.rb
+++ b/acceptance/tests/resource/file/source_attribute.rb
@@ -164,7 +164,7 @@ with_puppet_running_on(master, master_opts, testdir) do
 
     step "second run should not update file"
     on(agent, puppet('agent', "--test --server #{master}")) do
-      assert_no_match(/content changed/, stdout, "Shouldn't have overwrote any files")
+      assert_no_match(/content changed {(#{Regexp.new(checksums.join('|'))})}/, stdout, "Shouldn't have overwrote any files")
     end
   end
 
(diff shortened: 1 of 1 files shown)


johnduarte:
This is a proof of concept and needs testing before being merged.

acceptance/tests/resource/file/source_attribute.rb
```diff
@@ -164,7 +164,7 @@ class source_test_module {
 
     step "second run should not update file"
     on(agent, puppet('agent', "--test --server #{master}")) do
-      assert_no_match(/content changed/, stdout, "Shouldn't have overwrote any files")
+      assert_no_match(/content changed {(#{Regexp.new(checksums.join('|'))})}/, stdout, "Shouldn't have overwrote any files")
```
joshcooper: `checksums` includes all of the different checksum types, but the agent may only update a single file resource, e.g. for `mtime`, so this would make the regexp not match, causing the test to pass when it should fail.

acceptance/tests/resource/file/source_attribute.rb
```diff
@@ -164,7 +164,7 @@ class source_test_module {
 
     step "second run should not update file"
```
justinstoller: I may not be grokking the problem set here, but to make this test resilient to timing issues before this step we should should get the time on the master when we created the files and on the agent after the first application, then ensure that the time on the agent is after the time when we created the files on the master before running this step.

joshcooper:
@johnduarte I'm going to close this as I'm going to put up a different PR to detect when the agent applied a file whose ctime/mtime are in the past relative to the server's time, and handle that case appropriately.

PR #4274, opened 2015-09-22
(docs) Fix ambiguous reference to file: URIs in source attribute

nfagerlund:
These docs seemed to imply that you could download remote files with file: URIs,
when they're really just another way to write local file paths.

When this is merged, please resolve https://tickets.puppetlabs.com/browse/DOCUMENT-413.


joshcooper:
Same comment about appveyor, merging

PR #4275, opened 2015-09-22
(PUP-5273) Update Ruby to include OpenSSL 1.0.2d

melissa:


Iristyle:
I think we want a ticket number in JIRA for this, since a 1.0.0s to 1.0.2d bump is potentially significant.

PR #4277, opened 2015-09-23

whopper:
query packages.


whopper:
/cc @haus what weirds me out about this a little is that apparently, the epoch value overrules any other part of the release string, so we'd treat `1:1.2.3-4` as greater than `0:3.4.5-7`and so on. This is evidenced by https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/yum.rb#L223-#L235. 

Just checking - does that seem right? We've never included epochs in our ensure field before, so I wanted to make sure everything is sane.

joshcooper:
I don't know much about this. @branan @haus @geoffnichols @shrug?

haus:
This seems like the right thing to do.

kylog:
Fedora 22 is on the 2015.2 async list at https://confluence.puppetlabs.com/display/PROD/PE+2015.2.1+Agent+Platforms, so let's target stable for this PR.

PR #4279, opened 2015-09-23
(maint) update cron type doc to more closely match style guide

fiddyspence:
Update inline documentation for the cron type to more closely match the style guide

### DIFF
--- a/lib/puppet/type/cron.rb
+++ b/lib/puppet/type/cron.rb
@@ -18,29 +18,29 @@ Puppet::Type.newtype(:cron) do
 
     Example:
 
-        cron { logrotate:
-          command => "/usr/sbin/logrotate",
-          user    => root,
-          hour    => 2,
-          minute  => 0
+        cron { 'logrotate':
+          command => '/usr/sbin/logrotate',
+          user    => 'root',
+          hour    => '2',
+          minute  => '0',
         }
 
     Note that all periodic attributes can be specified as an array of values:
 
-        cron { logrotate:
-          command => "/usr/sbin/logrotate",
-          user    => root,
-          hour    => [2, 4]
+        cron { 'logrotate':
+          command => '/usr/sbin/logrotate',
+          user    => 'root',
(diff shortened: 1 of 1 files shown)


joshcooper:
Thanks Chris! LGTM. @nfagerlund?

lib/puppet/type/cron.rb
```diff
@@ -18,29 +18,29 @@
...
-        cron { logrotate:
-          command => "/usr/sbin/logrotate",
-          user    => root,
-          hour    => 2,
-          minute  => 0
+        cron { 'logrotate':
+          command => '/usr/sbin/logrotate',
+          user    => 'root',
+          hour    => '2',
+          minute  => '0',
```
nfagerlund: What's the deal with quoting these numbers? They ain't octal, so it should be fine to use real integers, right?

nfagerlund: Well, if it's just for consistency with strings like `'*/10'`, I'm fine with that, but in that case the array of numbers below should be an array of quoted numbers too.

nfagerlund:
Other than that nitpick, lgtm too.

joshcooper:
@fiddyspence Thank you for your contribution! When you get a chance, can you review @nfagerlund's comments and update this PR?

joshcooper:
ping @fiddyspence

lib/puppet/type/cron.rb
MikaelSmith: The style guide doesn't say anything I can find about numbers. So have a preference for all quoted or unquoted?

nfagerlund: Since we're not writing any of these in octal or hex, I prefer all unquoted, with quotes only around the non-numeric strings like `'*/10'`.

MikaelSmith:
Closing in favor of https://github.com/puppetlabs/puppet/pull/4440 with @nfagerlund's suggestion.

PR #4285, opened 2015-09-24

whopper:
query packages.

### DIFF
--- a/lib/puppet/provider/package/rpm.rb
+++ b/lib/puppet/provider/package/rpm.rb
@@ -281,6 +281,7 @@ Puppet::Type.type(:package).provide :rpm, :source => :rpm, :parent => Puppet::Pr
       self::NEVRA_FIELDS.zip(match.captures) { |f, v| hash[f] = v }
       hash[:provider] = self.name
       hash[:ensure] = "#{hash[:version]}-#{hash[:release]}"
+      hash[:ensure].prepend("#{hash[:epoch]}:") if hash[:epoch] != '0'
     else
       Puppet.debug("Failed to match rpm line #{line}")
     end
--- a/spec/unit/provider/package/rpm_spec.rb
+++ b/spec/unit/provider/package/rpm_spec.rb
@@ -12,6 +12,7 @@ describe provider_class do
     chkconfig 0 1.3.30.2 2.el5 x86_64
     myresource 0 1.2.3.4 5.el4 noarch
     mysummaryless 0 1.2.3.4 5.el4 noarch
+    tomcat 1 1.2.3.4 5.fc22 noarch
     RPM_OUTPUT
   end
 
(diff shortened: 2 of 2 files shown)


whopper:
@kylog @joshcooper re-targeted from https://github.com/puppetlabs/puppet/pull/4277

branan:
@haus @melissa do either of you understand RPM epochs and want to tag in on this?

haus:
@branan as i mentioned in #4277, this seems like the right thing.

melissa:
:+1: this aligns with my expectation on how epochs should be handled

lib/puppet/provider/package/rpm.rb
```diff
@@ -281,6 +281,7 @@ def self.nevra_to_hash(line)
       self::NEVRA_FIELDS.zip(match.captures) { |f, v| hash[f] = v }
       hash[:provider] = self.name
       hash[:ensure] = "#{hash[:version]}-#{hash[:release]}"
+      hash[:ensure].prepend("#{hash[:epoch]}:") if hash[:epoch] != '0'
```
joshcooper: The rpm provider uses [`Puppet::Util::Package.versioncmp`](https://github.com/whopper/puppet/blob/PUP-5025/RPM-epoch/lib/puppet/provider/package/rpm.rb#L142) to compare ensure values, but that method doesn't know how to handle epochs in version strings, such as what we're returning here.

Do we want to update `versioncmp` to handle epochs?

Or implement an `rpmversioncmp` that knows how to handle epochs? Based on http://www.rpm.org/max-rpm-snapshot/s1-rpm-depend-manual-dependencies.html:

>  Epoch is the most significant component of a package's complete version identifier with regards to RPM's version comparison algorithm.

I assume we'd want to first compare epochs, defaulting to epoch 0 if one is not specified. If the epochs are different, then return +/-1. If the epochs are the same, then delegate to the original `versioncmp`?

Assuming epochs are specific to rpms, I think we'd want to do the latter. Thoughts?

whopper: @joshcooper hm, it looks like the only places the RPM provider uses `versioncmp` is to check the version of RPM itself though, right?

From what I recall, the package that's actually being installed gets it's ensure version compared in the yum provider,  at this method: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/yum.rb#L226

From my previous (maybe flawed) understanding, it's the child providers of `rpm` that deal with the epoch. `rug`, for example, says: `# XXX: We don't actually deal with epochs here.` in it's `install` method. `zypper` and `urpmi` also do this. `Yum` seems to be the only provider that does any extensive version comparisons at all, for some reason.

(Also, there actually already is an `rpmvercmp` method, also, which is pretty gnarly: https://github.com/whopper/puppet/blob/PUP-5025/RPM-epoch/lib/puppet/provider/package/rpm.rb#L182)

PR #4286, opened 2015-09-24
(PUP-4495) Fix tag filtering regression

HAIL9000:
Prior to this commit a change made to the way tags are split when
filtering was causing the agent to create more resources than
nessecary.

Update the way a tag is recoreded when filtering to fix this issue.

### DIFF
--- a/lib/puppet/transaction.rb
+++ b/lib/puppet/transaction.rb
@@ -350,11 +350,8 @@ class Puppet::Transaction
     resource.appliable_to_host? && resource.appliable_to_device?
   end
 
-  def handle_qualified_tags( qualified )
-    # The default behavior of Puppet::Util::Tagging is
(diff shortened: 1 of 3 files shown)


kylog:
:+1: 

Nit: can you amend the commit message to reference the commit that introduced the regression - https://github.com/puppetlabs/puppet/commit/0412515e2 - and also fix the spelling of 'necessary' and 'recorded'?  

And btw 'necessary' is one of my least favorite words to spell ... :)

hlindberg:
Looks good (for PUP-4495). Mentioning the SHA for the regression commit is also good. A test would be good. "Tags given in a transaction searches for the complete tag, not the individual parts"

kylog:
:+1: 

I'll merge once travis/appveyor go green.

PR #4287, opened 2015-09-24
(PUP-4695) New hiera_in_templates acceptance test

segatspiretech:
Until now we had no acceptance tests for verifying hiera function calls
(hiera(), hiera_array(), hiera_hash(), and hiera_include()) from within
a template (erb or epp).

This test calls hiera functions from with in an erb template and from within
an epp template.  The master is configured with a manifest.  The client
applies the manifest using the puppet agent -t call.  The file created by
templates are examined to determine whether the hiera function calls returned
the expected values.

hiera_include() is called from the epp template.  The included sub-modules
each create a file that is then checked for correctness.

9/24/15

### DIFF
--- a/acceptance/tests/parser_functions/hiera_in_templates.rb
+++ b/acceptance/tests/parser_functions/hiera_in_templates.rb
@@ -1,0 +1,443 @@
+test_name "Calling Hiera function from inside templates"
+
+@module_name = "hieratest"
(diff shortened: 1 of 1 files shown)


segatspiretech:
ping @er0ck.

er0ck:
LGTM!  :+1:

PR #4288, opened 2015-09-24
(PUP-4890) Thread code_id into compiler

joshcooper:
Commit e14f2bc modified the catalog's compiler terminus to copy the
code_id from the request to the catalog that it returned. However, this
isn't sufficient as the code_id needs to be set on the catalog and
available during compilation.

This commit modifies the catalog's compiler terminus to pass the code_id
from the request to the compiler (Puppet::Parser::Compiler), which sets
it on the catalog it creates and returns.

Note the compiler terminus calls `catalog.filter` to remove exported
resources, making a copy of the catalog in the process, and is why the
catalog's `convert` method needs to copy the code_id.


joshcooper:
@hlindberg @lutter @kylog @cprice404

kylog:
Travis bonked on one test.

lutter:
Just from reading through the changes, this gets a (thumbsup) from me.

joshcooper:
@kylog should be good now

PR #4289, opened 2015-09-25
(PUP-5272) Allow test to pass if ctime/mtime don't converge

joshcooper:
Previously, the test required all checksum types to converge in one run.
However, ctime and mtime are sensitive to clock skew, and can require
multiple runs to converge. For example, suppose the source file's mtime
on the master is 11:00, but the agent's clock is behind at 10:00. The
agent downloads and applies a catalog, creating the file at 10:02. The
next time the agent runs it will compare its local time of 10:02 against
11:00, and decide the file is out of sync, and update it again. This
process will repeat until the file's ctime on the agent is greater than
or equal the file's ctime on the master.

This commit modifies the install step to query each host for its local
time, and does a rough calculation to determine the difference between
all hosts. Note the timestamp that beaker outputs is on the coordinator
node and so is useless when trying to correlate timestamps.

This commit also modifies the file source attributes test to allow exit
codes 0 and 2 during the second run, ensure md5* and sha* checksums do
not change, and warn, but not fail, if ctime and/or mtime change.

### DIFF
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -75,4 +75,22 @@ agents.each do |agent|
   on agent, "#{ruby} --version"
 end
 
+# Get a rough estimate of clock skew among hosts
+times = []
+hosts.each do |host|
+  ruby = Puppet::Acceptance::CommandUtils.ruby_command(host)
+  on(host, "#{ruby} -e 'puts Time.now.strftime(\"%Y-%m-%d %T.%L %z\")'") do |result|
+    times << result.stdout.chomp
+  end
+end
+times.map! do |time|
+  (Time.strptime(time, "%Y-%m-%d %T.%L %z").to_f * 1000.0).to_i
+end
+diff = times.max - times.min
+if diff < 60000
+  logger.info "Host times vary #{diff} ms"
+else
+  logger.warn "Host times vary #{diff} ms, tests may fail"
+end
+
 configure_gem_mirror(hosts)
--- a/acceptance/tests/resource/file/source_attribute.rb
+++ b/acceptance/tests/resource/file/source_attribute.rb
@@ -163,8 +163,25 @@ with_puppet_running_on(master, master_opts, testdir) do
     end
 
     step "second run should not update file"
-    on(agent, puppet('agent', "--test --server #{master}")) do
-      assert_no_match(/content changed/, stdout, "Shouldn't have overwrote any files")
+    on(agent, puppet('agent', "--test --server #{master}"), :acceptable_exit_codes => [0,2]) do
+      assert_no_match(/content changed.*(md5|sha256)/, stdout, "Shouldn't have overwritten any files")
+
+      # When using ctime/mtime, the agent compares the values from its
+      # local file with the values on the master to determine if the
+      # file is insync or not. If during the first run, the agent
+      # creates the files, and the resulting ctime/mtime are still
+      # behind the times on the master, then the 2nd agent run will
+      # consider the file to not be insync, and will update it
+      # again. This process will repeat until the agent updates the
+      # file, and the resulting ctime/mtime are after the values on
+      # the master, at which point it will have converged.
+      if stdout =~ /content changed.*ctime/
+        Log.warn "Agent did not converge using ctime"
+      end
+
+      if stdout =~ /content changed.*mtime/
+        Log.warn "Agent did not converge using mtime"
+      end
     end
   end
 


joshcooper:
I was running this with code that sets the agent time back by 3 minutes, and interestingly the agent was 5 minutes behind the master (so the VM was already 2 minutes off), which confirms the NTP issues we've been seeing.

```
v9j44f9adr8zmyx.delivery.puppetlabs.net (master) 23:23:46$ env PATH="/opt/puppetlabs/puppet/bin:${PATH}" ruby -e 'puts Time.now.strftime("%Y-%m-%d %T.%L %z")'
2015-09-24 23:25:25.489 -0700

v9j44f9adr8zmyx.delivery.puppetlabs.net (master) executed in 0.09 seconds

hg0n1znrdaoeicn.delivery.puppetlabs.net (agent) 23:23:46$ env PATH="/opt/puppetlabs/puppet/bin:${PATH}" ruby -e 'puts Time.now.strftime("%Y-%m-%d %T.%L %z")'
2015-09-24 23:20:46.991 -0700

hg0n1znrdaoeicn.delivery.puppetlabs.net (agent) executed in 0.07 seconds
Warning: Host times vary 278497 ms
```

acceptance/setup/aio/pre-suite/010_Install.rb
```diff
@@ -3,6 +3,13 @@
 
 extend Puppet::Acceptance::InstallUtils
 
+# Force clock skew, note agent can't be more than 5 minutes
```
er0ck: did you want to leave this commented out section in, @joshcooper ?

acceptance/setup/aio/pre-suite/010_Install.rb
```diff
@@ -75,4 +82,24 @@
...
+times = []
+hosts.each do |host|
+  ruby = Puppet::Acceptance::CommandUtils.ruby_command(host)
+  on(host, "#{ruby} -e 'puts Time.now.strftime(\"%Y-%m-%d %T.%L %z\")'") do |result|
+    times << result.stdout.chomp
+  end
+end
+times.map! do |time|
+  (Time.strptime(time, "%Y-%m-%d %T.%L %z").to_f * 1000.0).to_i
+end
```
er0ck: double loop.  couldn't this Time.strptime be put in the loop above?

acceptance/tests/resource/file/source_attribute.rb
```diff
@@ -163,8 +163,25 @@ class source_test_module {
...
+      # consider the file to not be insync, and will update it
+      # again. This process will repeat until the agent updates the
+      # file, and the resulting ctime/mtime are after the values on
+      # the master, at which point it will have converged.
+      if stdout =~ /content changed.*ctime/
+        Log.warn "Agent did not converge using ctime"
+      end
+
+      if stdout =~ /content changed.*mtime/
+        Log.warn "Agent did not converge using mtime"
```
er0ck: nice.

acceptance/setup/aio/pre-suite/010_Install.rb
joshcooper: I could go either way. It's handy so I don't have to remember how to do that, but if it's distracting I can remove.

acceptance/setup/aio/pre-suite/010_Install.rb
```diff
@@ -75,4 +82,24 @@
...
+diff = times.max - times.min
+logger.warn "Host times vary #{diff} ms"
```
joshcooper: @er0ck Using `logger.warn` seems to work, but I'm not sure if that's beaker best practice

acceptance/tests/resource/file/source_attribute.rb
```diff
@@ -163,8 +163,25 @@ class source_test_module {
...
+      # local file with the values on the master to determine if the
+      # file is insync or not. If during the first run, the agent
+      # creates the files, and the resulting ctime/mtime are still
+      # behind the times on the master, then the 2nd agent run will
+      # consider the file to not be insync, and will update it
+      # again. This process will repeat until the agent updates the
+      # file, and the resulting ctime/mtime are after the values on
+      # the master, at which point it will have converged.
+      if stdout =~ /content changed.*ctime/
+        Log.warn "Agent did not converge using ctime"
```
joshcooper: @er0ck similar question about best practice?

er0ck:
works for me on redhat7.  with clock skew forced:
```
  * Install repositories on target machines...
Installing puppetserver
Installing puppet-agent
Warning: Host times vary 180202 ms
```
followed, of course by:
```
 * second run should not update file
Warning: Agent did not converge using ctime
Warning: Agent did not converge using mtime
[...]
 * second run should not update any files using apply with puppet:/// URI source
tests/resource/file/source_attribute.rb passed in 92.06 seconds
```

acceptance/setup/aio/pre-suite/010_Install.rb
er0ck: stick it in the library somewhere and comment out the single line here?

PR to beaker
host.time++  :angel: 


i'm fine with it either way.  maybe reference the ticket/test in any comments?

joshcooper:
@er0ck I removed the commented out lines, will submit a ticket to beaker

PR #4290, opened 2015-09-25
Fix latest version lookup for pip configured with custom repo

Yuav:
The current implementation has pypi.python.org hardcoded. With newer pip versions, it supports having a /etc/pip.conf file specifying a different location. Existing behavior results in `pip install` running on every single Puppet run, if used to install a package available in a custom repo, but not on pypi.python.org

The implementation is the least hackish way to have Pip return a list of versions

### DIFF
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -56,17 +56,21 @@ Puppet::Type.type(:package).provide :pip,
     return nil
   end
 
-  # Ask the PyPI API for the latest version number.  There is no local
-  # cache of PyPI's package list so this operation will always have to
-  # ask the web service.
+  # Return latest version from PyPI repo as seen by pip
   def latest
-    client = XMLRPC::Client.new2("http://pypi.python.org/pypi")
-    client.http_header_extra = {"Content-Type" => "text/xml"}
-    client.timeout = 10
-    result = client.call("package_releases", @resource[:name])
-    result.first
-  rescue Timeout::Error => detail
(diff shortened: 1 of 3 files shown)


kylog:
Hi @Yuav, thanks for the contribution!

We have some bookkeeping we like to do for functional changes: can you file a ticket describing the issue at tickets.puppetlabs.com? And then reference it in the commit message summaries (like "(PUP-1234) Fix ....").

Thanks!

lib/puppet/provider/package/pip.rb
```diff
@@ -56,17 +56,21 @@ def query
...
   def latest
-    client = XMLRPC::Client.new2("http://pypi.python.org/pypi")
-    client.http_header_extra = {"Content-Type" => "text/xml"}
-    client.timeout = 10
-    result = client.call("package_releases", @resource[:name])
-    result.first
-  rescue Timeout::Error => detail
-    raise Puppet::Error, "Timeout while contacting pypi.python.org: #{detail}", detail.backtrace
+    pip_cmd = which(self.class.cmd) or return nil
+    # This is the least hackish way to have pip look up versions. Output we're interested in is:
```
kylog: Well this approach work for older versions of `pip` as well?

Yuav:
Hmm, seems there's a bug with pip 1.0 not listing versions found. From 1.5 and above it works

lawre:
What is the status of this being merged in?  We're hitting this issue now and are having to work around it by specifying package versions instead of "latest".  Thanks.

Yuav:
This has been included into Puppet 4.5.0

PR #4291, opened 2015-09-25
(PUP-5282) Use absolute_path to detect local installation of gem on windows

fiddyspence:
Attempting to install a gem from a local source on windows fails because of the way
that the provider code detects using URI.  This patch preserves the current behaviour
but allows the a puppet util method to detect a fully qualified path to pass to the gem
command, which allows install

### DIFF
--- a/lib/puppet/provider/package/gem.rb
+++ b/lib/puppet/provider/package/gem.rb
@@ -111,8 +111,13 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
         # we don't support puppet:// URLs (yet)
         raise Puppet::Error.new("puppet:// URLs are not supported as gem sources")
       else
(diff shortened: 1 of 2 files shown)


branan:
So the idea here is to not require `file://` on windows paths, right?

branan:
This seems reasonable, as long as there's no chance of a single-letter URI scheme that the gem command would be aware of and respect. That seems incredibly unlikely (mostly due to windows compatibility requirements)

fiddyspence:
Yeah - I kind of figured that the change of an absolute path colliding was pretty slim (and besides, file:///<windows path> doesn't work on windows anyway because see the output in the ticket).  Just for funsies I checked the IANA URI scheme registration page (http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) and there are no single character ones.

PR #4299, opened 2015-09-28
(MAINT) Fix bug in hiera-in-templates test

cprice404:
This test was written assuming that there would always be
exactly one agent SUT.  It threw an error when run in a
configuration that had multiple agent SUTs (such as Puppet Server);
this commit changes the code so that the test won't *definitely*
fail, but a better fix would probably mandate that the whole
test be restructured to support the ability to have a unique
tempdir per agent node.

### DIFF
--- a/acceptance/tests/parser_functions/hiera_in_templates.rb
+++ b/acceptance/tests/parser_functions/hiera_in_templates.rb
@@ -2,7 +2,7 @@ test_name "Calling Hiera function from inside templates"
 
 @module_name = "hieratest"
 @coderoot = master.tmpdir("#{@module_name}")
-@resultdir = agent.tmpdir("#{@module_name}_results")
+@resultdir = agents.first.tmpdir("#{@module_name}_results")
 
 @msg_default = 'message from default.yaml'
 @msg_production = 'message from production.yaml'


kylog:
Sigh. We just did a purge of exactly this type of mistake. Beaker makes it too easy to do this.

joshcooper:
Yup, and I filed that as https://tickets.puppetlabs.com/browse/BKR-558

acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -2,7 +2,7 @@
 
 @module_name = "hieratest"
 @coderoot = master.tmpdir("#{@module_name}")
-@resultdir = agent.tmpdir("#{@module_name}_results")
+@resultdir = agents.first.tmpdir("#{@module_name}_results")
```
joshcooper: I don't think this will work as you might have an posix agent on the master, and a windows agent, and there's no guarantee about which agent first will return, and the resulting path likely won't work for both.

cprice404: Agree.  Feel free to close, just putting this up as an illustration of the issue.

kylog:
Ping @segatspiretech.

cprice404:
Closing in favor of #4306

PR #4302, opened 2015-09-29
(PUP-5290) Catch and log exceptions raised by saving to cache

dalen:
Failing to cache requests could lead to silent exits. For example if the
apply application failed to cache to PuppetDB catalog terminus. This
will log the exception and continue.


dalen:
Retargeted against stable: #4303

PR #4303, opened 2015-09-29
(PUP-5290) Log exceptions raised by saving to cache

dalen:
Failing to cache requests could lead to silent exits. For example if the
apply application failed to cache to PuppetDB catalog terminus. This
will log the exception and continue.

### DIFF
--- a/lib/puppet/indirector/indirection.rb
+++ b/lib/puppet/indirector/indirection.rb
@@ -196,7 +196,12 @@ class Puppet::Indirector::Indirection
         result.expiration ||= self.expiration if result.respond_to?(:expiration)
         if cache?
           Puppet.info "Caching #{self.name} for #{request.key}"
-          cache.save request(:save, key, result, options)
+          begin
+            cache.save request(:save, key, result, options)
+          rescue => detail
+            Puppet.log_exception(detail)
+            raise detail
+          end
         end
 
         filtered = result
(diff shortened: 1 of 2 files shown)


branan:
Globally allowing failures of caching doesn't work - puppet currently uses the "cache" mechanism to store real data that we expect to be persisted, and that users expect to be consistent. Not failing if we don't write that data would potentially allow users to lose a bunch of data without being alerted.

`.find` does take an options hash, though. In the specific locations where the cache is really being used as a cache, modifying the find calls to have some sort of `allow_cache_failure` option would be reasonable.

dalen:
Well, we could likely at least log the exception and then re-raise it instead.

kylog:
Yeah, log and re-raise would solve the "silent" part.

joshcooper:
Here's an example of how we write cert related files as a side-effect of caching: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/ssl/host.rb#L77-L81,

dalen:
Here is the code that leads to the silent exits: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/application/apply.rb#L255-L260

But changing that would likely lead to double error messages from compiler errors.

dalen:
I changed it to log and reraise now

kylog:
This seems like only goodness to me. This changes no functional behavior other than logging an exception.

ahpook:
:tada: :balloon: :balloon: :tada:

PR #4304, opened 2015-09-29
PUP-5291 fix "stream closed" problem with SSH transport

jorhett:
PUP-5291 fix "stream closed" problem with SSH transport closing the session with the IO handles still open

### DIFF
--- a/lib/puppet/util/network_device/transport/ssh.rb
+++ b/lib/puppet/util/network_device/transport/ssh.rb
@@ -73,6 +73,7 @@ class Puppet::Util::NetworkDevice::Transport::Ssh < Puppet::Util::NetworkDevice:
       @channel = nil
       @ssh.close if @ssh
     rescue IOError
+      Puppet.debug "device terminated ssh session impolitely"
     end
   end
 
--- a/lib/puppet/util/network_device/transport/ssh.rb
+++ b/lib/puppet/util/network_device/transport/ssh.rb
@@ -68,9 +68,12 @@ class Puppet::Util::NetworkDevice::Transport::Ssh < Puppet::Util::NetworkDevice:
   end
 
   def close
-    @channel.close if @channel
-    @channel = nil
-    @ssh.close if @ssh
+    begin
+      @channel.close if @channel
+      @channel = nil
+      @ssh.close if @ssh
+    rescue IOError
+    end
   end
 
   def expect(prompt)


joshcooper:
@jorhett Thanks for your contribution. A minor nit, travis failed on the commit summary, could you update it to read "(PUP-5291) ..." We do this to ensure our automation can reconcile puppet commits and jira tickets.

Do you know if the switch is closing the idle ssh connection, and then puppet gets an exception when it tries to close the already closed connection? This can likely happen if the agent takes a long time to evaluate facts, longer than the switch's idle timeout. For example, I think that's what's happening in https://ask.puppetlabs.com/question/1389/puppet-device-fails-halfway-through-fact-discovery-on-cisco-catalyst-2960/

```
Error: Could not retrieve local facts: connection closed by remote host
```

Or is puppet leaking socket connections, and we only notice when we try to close the transport while connections are open? The bug description sounds like it's this case, and if so, I'd like to fix that case in puppet instead of swallowing the exception.

jorhett:
I'll fix the commit.

Yeah I think those are different problems.

The exception happens at '''@ssl.close'''. It's breaking the IO handles we have at that point, which is causing the problem. It might be fixable by closing up our handles first but there wasn't an obvious and clear way to do this inside the transport.

Looking through the net-ssh author's code, he always wraps the SSH object '''.close()''' with the same begin/rescue I used in my patch.

lib/puppet/util/network_device/transport/ssh.rb
```diff
@@ -68,9 +68,12 @@ def connect(&block)
   end
 
   def close
-    @channel.close if @channel
-    @channel = nil
-    @ssh.close if @ssh
+    begin
+      @channel.close if @channel
+      @channel = nil
+      @ssh.close if @ssh
```
joshcooper: The network device transport code calls [Net::SSH.start](https://github.com/jorhett/puppet/blob/fixssh/lib/puppet/util/network_device/transport/ssh.rb#L34), and since we don't specify a block, the [`Net::SSH.start` method returns the connection](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh.rb#L235), where the caller is responsible for closing the connection (really it's a `Net::SSH::Connection::Session` object). I think you are seeing an exception because the channel is being closed twice. The first time is when we call it above on line 72. Then on line 74, we call [`@ssh.close`](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/session.rb#L116-L121) which closes channels associated with the session, and likely raises an exception because the underlying socket has already been closed.

We could probably drop the call to `@channel.close` and just call `@ssh.close`, and remove the begin/rescue block. 

Another approach would be to use the block form of `Net::SSH.start` and let it close the channel(s) and session. You would need to change `@channel` to not be an instance variable and pass it as an argument to the various methods, e.g. `def process_ssh(channel)`, but that would be a bit more intrusive.

HAIL9000:
Hey @jorhett, thanks for the contribution! We wanted to check in and see if you were still interested in fixing this up since it's been a little while.

jorhett:
sorry, yes. Unfortunately every time I open up this to work on it something hits me. But I am poking at it, mostly because I found another issue. Expect a consolidated patch tomorrow which will address the testing issues.

jorhett:
Confirmed: your suggestion won't work. First I tried removing the begin/end and commenting out the @channel.close

```
$ puppet device --verbose
Info: starting applying configuration to switch01.example.com at ssh://switch01.example.com/
Info: Retrieving pluginfacts
Info: Retrieving plugin
Error: Could not retrieve local facts: closed stream
Error: Failed to apply catalog: Could not retrieve local facts: closed stream
```

Then I tried commenting out @channel = nil as well.

```
$ puppet device --verbose
Info: starting applying configuration to switch01.example.com at ssh://switch01.example.com/
Info: Retrieving pluginfacts
Info: Retrieving plugin
Error: Could not retrieve local facts: closed stream
Error: Failed to apply catalog: Could not retrieve local facts: closed stream
```

The issue at hand is the rude termination of SSH using RST. If this occurs, the SSH object itself will toss an IOError for any request made of it, including close.

This is where the maintainer is using the same approach himself: https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/buffered_io.rb with comments about same at https://www.omniref.com/github/net-ssh/net-ssh/2.1.1/symbols/Net::SSH::BufferedIo#line=152

Likewise, your own ticket for Beaker includes significant discussion of the dynamics here: https://tickets.puppetlabs.com/si/jira.issueviews:issue-html/BKR-70/BKR-70.html

jorhett:
Also, to be clear about severity: SSH doesn't work with Puppet device on any Cisco device I could get my hands on due to this error. It's pretty much a non-starter with anything described on your website for puppet device.

lib/puppet/util/network_device/transport/ssh.rb
```diff
@@ -68,9 +68,12 @@ def connect(&block)
...
+    rescue IOError
```
branan: I don't think it's worth being a warning, but it might be useful to have the exception logged at info or debug level

jorhett:
I've added the info log message as requested. Feel free to reword as you wish.

jorhett:
Note that I think that this info level output is pretty obnoxious when used with verbose mode, due to the fact that Puppet connects and reconnects over and over and over again.

```
$ puppet device --verbose
Info: starting applying configuration to switch01.example.com at ssh://switch01.example.com/
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: device terminated ssh session impolitely
Info: Caching catalog for switch01.example.com
Info: Applying configuration version '1450252973'
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Notice: /Stage[main]/Device_hiera::Vlans/Vlan[200]/ensure: created
Info: device terminated ssh session impolitely
Notice: /Stage[main]/Device_hiera::Vlans/Vlan[400]/ensure: created
Info: device terminated ssh session impolitely
Info: Class[Device_hiera::Vlans]: Unscheduling all events on Class[Device_hiera::Vlans]
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
Info: device terminated ssh session impolitely
```

It goes on and on like this for a page or more.

branan:
@jorhett Does `--verbose` include debug level? If not maybe that's the right place for this info.

If it does, I guess it's better to eat these than to make verbose logging totally unusable

jorhett:
Nope, verbose doesn't. I've set it down to debug locally for the last two weeks and that works fine for me. So I'll revise this PR...

branan:
@jorhett Just wanted to remind you that we're still waiting for that updated commit to use 'debug' level

MikaelSmith:
@branan has been updated

PR #4305, opened 2015-09-29
Resource type zpool create options

tnn:
On a (zfs) zpool there are options which can only be applied at creation
time. In order for puppet to handle the lifecycle of zpools, those
create options should be passed to the zpool create cli.

In particular the option `ashift` is required on hard drive using Advanced Formatting (4k).

This just serves as a initial prototype of how this could look. Please advise if Puppetlabs would accept this feature and what would be needed for this to be included, beside unit and acceptance tests.

Kind regards
Troels

### DIFF
--- a/lib/puppet/provider/zpool/zpool.rb
+++ b/lib/puppet/provider/zpool/zpool.rb
@@ -95,8 +95,28 @@ Puppet::Type.type(:zpool).provide(:zpool) do
     end
   end
 
+  #builds up a set of zpool options
+  #
+  # Options on a zpool that applies on creation time
+  # and are otherwise ignored.
+  # create_options is expected to be a map of key-pairs each denoting an option.
+  #
+  # For example:
+  #
+  # zpool { 'tank':
+  #   create_options => {
+  #     'ashift' => '12'
+  #     'altroot' => '/reservoir'
+  #   }
+  # }
+  def build_create_opts
+    @resource[:create_options]
+      .collect { |opt| "-o ${opt[1]}=$opt[2]}" }
+      .reduce { |memo, o| "${memo} ${opt}" }
+  end
+
   def create
-    zpool(*([:create, @resource[:pool]] + build_vdevs + build_named("spare") + build_named("log")))
+    zpool(*([:create, @resource[:pool]] + build_create_opts + build_vdevs + build_named("spare") + build_named("log")))
   end
 
   def destroy


whopper:
Hey @tnn - thanks for the contribution! 

First thing we'll need on this is a ticket in Jira. Would you mind creating one at https://tickets.puppetlabs.com/browse/PUP? Once it exists, you'll want to amend your commit message to begin with `(PUP-XYZ)` with the ticket number so we can reference it during a release.

And yeah, like you said, some spec tests would be a good start to getting this into a reviewable / merge-able state. Thanks!

branan:
AFAICT, there's no place create_options gets set right now? Do you have a change to the type that didn't make it into this commit?

Other than that, as @whopper said we'd like a proper ticket for this, and as you noted new tests would be great. Unit tests are usually sufficient, unless you feel particularly motivated to try to write beaker acceptance tests.

tnn:
@branan, @whopper Thanks for the quick feedback! This was just merely a prototype, I wanted to know if there was interest for having this features before spending time on a glorified, polished solution. Mostly I'm interesting in a) Is this a 'good' feature to have? b) What should the API/DSL look like? c) What implementation details should I be aware of?

lib/puppet/provider/zpool/zpool.rb
```diff
@@ -95,8 +95,28 @@ def build_vdevs
...
+  # zpool { 'tank':
+  #   create_options => {
+  #     'ashift' => '12'
+  #     'altroot' => '/reservoir'
+  #   }
+  # }
+  def build_create_opts
+    @resource[:create_options]
+      .collect { |opt| "-o ${opt[1]}=$opt[2]}" }
+      .reduce { |memo, o| "${memo} ${opt}" }
```
peterhuene: Could this call to `reduce` be replaced with a simple `.join(' ')`?

joshcooper:
@tnn Thank you for your contribution!

We generally prefer create explicit parameter/properties in the type so that puppet can validate manifests at compilation time, rather than later when the agent is applying the catalog. That way you catch problems earlier, before the agent has started to make changes on some platforms.

I'm assuming the set of zpool properties is a finite and known set? If so, can we create explicit parameters/properties for each one? For example, you can ensure values are numeric, string, boolean, enum, etc.

Puppet does allow arbitrary properties to be passed in some cases, e.g. `install_options` for the `package` type. However, that's because the options are completely arbitrary and for some providers, e.g. windows, are specific to the package being installed. In other words, the set of valid options are determined by the author of the package, not the type of package msi/exe.

HAIL9000:
Hey @tnn, just wanted to check in. Did @joshcooper's feedback make sense to you and/or are you still interested in working on this? Thanks!

tnn:
@HAIL9000 Hey Hailee, sorry - I've pretty busy lately. I had to make workaround - basically wrapping an `exec {}` since we were blocked on it. I'm not sure when I'll have time to look at it, but not in the next ~3 weeks.

MikaelSmith:
Is `ashift` only supported for ZFS on Linux?

HAIL9000:
@tnn @joshcooper @whopper How do we feel about closing this PR and filing a PUP ticket (which references this work) to continue work on this in the future?

tnn:
@MikaelSmith `ashift` is on all platforms.

PR #4306, opened 2015-09-29
(PUP-4695) Fixed hiera_in_templates acceptance test

segatspiretech:
The previous version of hiera_in_templates failed when there was more
than one agent in the test because of the way the temporary directories
were created using the agent.tmpdir method.  This fixes that problem.

9/28/15

### DIFF
--- a/acceptance/tests/parser_functions/hiera_in_templates.rb
+++ b/acceptance/tests/parser_functions/hiera_in_templates.rb
@@ -2,7 +2,6 @@ test_name "Calling Hiera function from inside templates"
 
 @module_name = "hieratest"
 @coderoot = master.tmpdir("#{@module_name}")
-@resultdir = agent.tmpdir("#{@module_name}_results")
 
 @msg_default = 'message from default.yaml'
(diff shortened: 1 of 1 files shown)


cprice404:
:see_no_evil:

acceptance/tests/parser_functions/hiera_in_templates.rb
```diff
@@ -257,23 +266,43 @@ class #{@module_name}::mod_fqdn {
...
+    t = agent.tmpdir("#{@module_name}_results")
+    tmp_dirs += "  #{h}: '#{t}'\n"
+    host_to_result_dir[h] = t
   end
-    osfamilies
+  result = {
+    'tmp_dirs' => tmp_dirs,
+    'host_to_result_dir' => host_to_result_dir
+  }
+  result
```
er0ck: maybe this is where josh was saying to return two objects?
i'm okay with it either way.  this test is pretty complex as-is.

er0ck:
:+1:  let's get server going again...

PR #4307, opened 2015-09-29
(PUP-5292) Allow empty arrays as first parameter to regsubst

dalen:


ahpook:
:+1: looks good to me! @whopper can you do the honors?

hlindberg:
This is not ideal - it allows `[undef, undef, undef,...]` to be passed. The type system bug should be fixed instead.

joshcooper:
Reverted in PR #4335

PR #4308, opened 2015-09-30
(PUP-4055) Add DNF package provider

whopper:
This series of commits adds a new DNF provider to Puppet, which is necessary due to Fedora's move from Yum to DNF.

A few updates to the Yum provider were required to make way for the new provider, which is its child.

### DIFF
--- a/acceptance/tests/resource/package/common_package_name_in_different_providers.rb
+++ b/acceptance/tests/resource/package/common_package_name_in_different_providers.rb
@@ -17,8 +17,9 @@ end
 
 def verify_state(hosts, pkg, state, match)
   hosts.each do |agent|
+    cmd = rpm_provider(agent)
     # Note yum lists packages as <name>.<arch>
-    on agent, 'yum list installed' do
+    on agent, "#{cmd} list installed" do
       method(match).call(/^#{pkg}\./, stdout)
     end
 
--- a/acceptance/tests/resource/package/yum.rb
+++ b/acceptance/tests/resource/package/yum.rb
@@ -17,8 +17,9 @@ end
 
 def verify_state(hosts, pkg, state, match)
   hosts.each do |agent|
-    # Note yum lists packages as <name>.<arch>
-    on agent, 'yum list installed' do
+    cmd = rpm_provider(agent)
+    # Note yum and dnf list packages as <name>.<arch>
+    on agent, "#{cmd} list installed" do
       method(match).call(/^#{pkg}\./, stdout)
     end
   end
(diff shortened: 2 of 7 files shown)


whopper:
There is an acceptance test at `acceptance/tests/resource/package/yum.rb` that needs to be fixed for Fedora 22, but I currently have no clue how to properly add and enable a local RPM repository with DNF. I swear I can get it to work manually on some machines, but not others...

acceptance/tests/resource/package/common_package_name_in_different_providers.rb
```diff
@@ -41,6 +41,12 @@ def verify_absent(hosts, pkg)
   clean_rpm agent, rpm_options
   setup_rpm agent, rpm_options
   send_rpm agent, rpm_options
+
+  # DNF requires a cache reset to make local repositories accessible.
+  cmd = rpm_provider(agent)
+  if cmd == 'dnf'
+    on(agent, 'dnf clean metadata')
+  end
```
joshcooper: Should this logic live in RpmUtils, or is the logic specific to this test?

Also meta question, does this affect puppet creating a repo and installing a package from the repo in the same run? Or is this only because we're creating a local repo for the test?

acceptance/tests/resource/package/yum.rb
```diff
@@ -37,6 +38,12 @@ def verify_absent(hosts, pkg)
   clean_rpm agent, rpm_options
   setup_rpm agent, rpm_options
   send_rpm agent, rpm_options
+
+  # DNF requires a cache reset to make local repositories accessible.
+  cmd = rpm_provider(agent)
+  if cmd == 'dnf'
+    on(agent, 'dnf clean metadata')
+  end
```
joshcooper: same comment as above

acceptance/tests/resource/package/common_package_name_in_different_providers.rb
whopper: @joshcooper I believe this was an issue specifically because of the local repo being created for the test, though I'll manually see what DNF does when a regular repo is added.

Yeah, I think that logic could make sense in the `send_rpm` method of `RpmUtils`. I'll move it over.

whopper:
@joshcooper updated to move logic into `RpmUtil`.

Also, re. your meta question, as far as I can tell, adding a regular repo doesn't require a DNF cache reset. We just have to do it after dropping in our RPM, otherwise DNF won't know about it. (Tested by adding the google-chrome repo and verifying that chrome-stable was available without refreshing the cache).

joshcooper:
Since this adds a new provider, should this target master for 4.3.0? We wouldn't have fedora 22 support until puppet-agent 1.3.0, but that doesn't see that far off. Thoughts? /cc @kylog

whopper:
@joshcooper oh yeah, not sure. I put this against stable based on Kylo's comment on an other PR:

Kylo:
Fedora 22 is on the 2015.2 async list at https://confluence.puppetlabs.com/display/PROD/PE+2015.2.1+Agent+Platforms, so let's target stable for this PR.

kylog:
Yep, confirmed that Fedora 22 is now targeted at 4.3.0 so please retarget these to master. Thanks!!

whopper:
CLosing in favor of https://github.com/puppetlabs/puppet/pull/4321

PR #4312, opened 2015-10-01
(PUP-5274) Create settings related files with the correct mode

joshcooper:
Previously, puppet would create settings related files using the process
umask, instead of the mode specified in defaults.rb, or overridden in
puppet.conf, and it would take a second puppet run to converge.

This issue was introduced in 691056e for PUP-2349, merged in 87b0b5d,
and has been present since 3.7.0. There are two issues with the code:

First, the case statement was trying to compare if mode was a string,
nil, or fixnum using ruby's case equality operator, but it was
actually comparing if mode's class was a string or nil, which was
never true, always falling through to the else case.

Second, the assignment statement `mode = case mode` introduced a `mode`
local variable whose value was set to nil, and shadowed the `mode`
accessor. This combined with above issue meant the method always yielded
nil, and Puppet would eventually call `File.open` with a nil mode when
creating the settings related file.

There is a DirectorySetting class that extends FileSetting, but it
doesn't have this problem because it uses puppet's file type to create
the directory, correctly passing mode to the `Dir.mkdir` call.

This commit avoids the shadow problem, and tests if `mode` is a String.
It also adds unit and acceptance tests as none of the existing tests
failed when the code was changed.

### DIFF
--- a/acceptance/lib/puppet/acceptance/common_utils.rb
+++ b/acceptance/lib/puppet/acceptance/common_utils.rb
@@ -123,7 +123,22 @@ module Puppet
           end
         end
       end
+    end
 
+    module CommandUtils
+      def ruby_command(host)
+        "env PATH=\"#{host['privatebindir']}:${PATH}\" ruby"
+      end
+      module_function :ruby_command
+
+      def gem_command(host)
+        if host['platform'] =~ /windows/
+          "env PATH=\"#{host['privatebindir']}:${PATH}\" cmd /c gem"
+        else
+          "env PATH=\"#{host['privatebindir']}:${PATH}\" gem"
+        end
+      end
+      module_function :gem_command
     end
   end
 end
--- a/acceptance/tests/ssl/ticket_5274_private_key_modes.rb
+++ b/acceptance/tests/ssl/ticket_5274_private_key_modes.rb
@@ -1,0 +1,44 @@
+test_name "PUP-5274: CA and host private keys should not be world readable"
+require 'puppet/acceptance/common_utils'
+
+confine :except, :platform => 'windows'
+
+def get_setting(host, ssldir, command)
+  on(host, puppet("agent --ssldir #{ssldir} #{command}")).stdout.chomp
+end
+
+def get_mode(host, path)
+  ruby = Puppet::Acceptance::CommandUtils.ruby_command(host)
+  on(host, "#{ruby} -e 'puts (File.stat(\"#{path}\").mode & 07777).to_s(8)'").stdout.chomp
+end
+
+hosts.each do |host|
+  ssldir        = host.tmpdir('ssldir')
+  cakey         = get_setting(host, ssldir, "--configprint cakey")
+  privatekeydir = get_setting(host, ssldir, "--configprint privatekeydir")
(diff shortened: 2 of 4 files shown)


joshcooper:
@er0ck @johnduarte I'm using `stat` in the acceptance test which I think will work for all of the 3.x platforms we test on. It will fail in 4.x platforms, e.g. solaris. In 4.x, we could use the `RubyCommand` logic to retrieve modes using ruby, but it's not present in 3.x acceptance. Thoughts?

johnduarte:
@joshcooper unfortunately, I think the only think that is universally available is `ls -l`. We could add a helper to perform a poor mans `stat --format '%a'` based on `ls -l` using something like the following which works when tested on Solaris 11, AIX 7.1, and Debian 7.

```
#!/bin/bash
tmp_=/tmp/tmp_ls_out
tmp_stats_=/tmp/tmp_stat_out
if [ -f "$1" ]; then ls -l "$1">$tmp_
else echo "Parameter must be regular file";exit 1;fi
awk '{print $1}' $tmp_ | sed -e 's/\(...\)/\1 /g' | awk '{print $1"\n"$2"\n"$3}'>$tmp_stats_
calc=`sed -n "$i s/-/0_X/g;s/r/4_X/g;s/w/2_X/g;s/x/1_X/g;s/_X/+/g;s/+$//p" $tmp_stats_|bc`
recalc=`echo $calc`;echo ${recalc// /}
```

er0ck:
nice work @johnduarte.  IMO, this should be part of beaker's host class as a public method which we can overload based on platform.

kylog:
Holy crap, I effed that change up badly. Sigh.

joshcooper:
@johnduarte @er0ck I cherry-picked a commit from stable for executing ruby commands, and used that to get the mode. Note the commit adds `privatebindir` to the PATH, which doesn't exist in 3.x, but it doesn't harm anything either. The command will still work in 3.x because we use system ruby there, and by keeping the commit as-is, it will cleanly merge to stable/master.

er0ck:
yep.  acceptance stuff looks fine to me :+1:

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -123,7 +123,22 @@ def reset_agent_ssl(resign = true)
           end
         end
       end
+    end
+
+    module CommandUtils
+      def ruby_command(host)
+        "env PATH=\"#{host['privatebindir']}:${PATH}\" ruby"
+      end
+      module_function :ruby_command
```
er0ck: nice.  all the spurious extend module "mixing" has been getting to me a bit.

johnduarte:
:+1:  Nicely done, @joshcooper

whopper:
Waiting for appveyor shenanigans to be resolved, but in the meantime I did a llittle pre-FR on this based on the issues described by @joshcooper in the ticket. All keys are indeed created with the correct permissions on the first agent run now, and `puppet cert generate` also results in files with the correct perms. :+1: lgtm, and with John and Eric's :+1:'s on the tests I believe we're ready to merge.

PR #4314, opened 2015-10-01
(PUP-5016) Fixup systemd / sysvinit issues in Debian and Ubuntu

whopper:
This series of commits makes some changes to the systemd, debian and init providers to work around the strange behavior of the sysvinit-systemd compatibility layer in Debian. In order:

### 1) Disable Init provider on Debian and Ubuntu

Previously, Debian 8 and Ubuntu 15.05, which are systemd-managed
platforms, would erroneously fall back to the init provider
when Systemd was unable to manage a SysVInit service without
a unit file. The Debian provider should be used instead, as
it inherits from init and provides the ability to enable and
disable services.

### 2) Fix check for Debian osfamily in systemd provider

Previously, we checked for lowercased 'debian' when trying
to determine if a platform is Debian-ish in the systemd provider.
Facter returns capitalized 'Debian', so this commit downcases
Facter's return value to ensure we properly detect the Debians.

### 3) Update linux services tests for Debian systemd behavior

Previously, services_enable_linux.rb was failing in Debian 8
and Ubuntu 15.05 due to issues with Systemd and SysVInit. The
Debian 8 issues are corrected by the actual fix to the systemd
and init providers, but the test continued to fail in Ubuntu
due to an upstream systemd-sysv compatibility bug.

This commit updates the test to use a service which has
a systemd unit file, cron, rather than apache2 which did not. This
allows the test to run correctly on Ubuntu. Once this upstream bug
is fixed, we'll probably want to undo the Ubuntu changes for
consistency.

In addition, the systemd masking test is also updated for strange
or broken Debian and Ubuntu systemd / Init behavior. In this case,
Ubuntu suffers from the same bug as above. In Debian, Apache2 does
not create the expected systemd symlinks as normal systemd services
do, so we use cron.service instead.

See https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1447807
for the upstream bug report.

### 4) Fix issue with Debian provider and SysVInit-Systemd compatibility layer

With the compatibility layer between systemd and SysVInit in Debian,
services can be managed in multiple ways. However, if a service is
masked with systemd, the equivalent SysVInit script becomes unable
to determine the status of the service. If we are trying to manage
the SysVInit version, we need to use systemctl to determine the
status in Debian 8 and Ubuntu 15.04.

Note that this is an issue only if querying the SysVInit version
of a service. `puppet resource service ntp` will fail to get the
status, while `puppet resource service ntp.service` will not.

### Testing efforts

I've run the full acceptance suite against several platforms, including Debian 8, Ubuntu 14.04, Centos 7, and Debian 7. This is still kind of a gnarly issue, so I wouldn't mind some help making sure everything is doing the right stuff.

### Other notes

A lot of this is solely necessary because of how weird things get in the SysVInit-Systemd compatibility layer. In Debian, you can manage the same services with either init scripts or unit files, but lots of services don't actually have real unit files. Puppet is currently taking into account that these init scripts and the compatibility-layer-generated unit scripts are different things, so we see services like `ntp` and `ntp.service`, where `ntp` is managed with the debian provider and `ntp.service` is managed with the systemd provider. There is a lot of weird overlap, such as the change I made in the Debian provider where we actually call `systemctl` in certain circumstances, even though we aren't trying to manage the service with systemd. 

(That particular case is necessary due to the init / systemd 'versions' of services not exactly playing along. Masking a service with systemd results in the init script not being able to determine the service status).

### DIFF
--- a/lib/puppet/provider/service/debian.rb
+++ b/lib/puppet/provider/service/debian.rb
@@ -63,4 +63,25 @@ Puppet::Type.type(:service).provide :debian, :parent => :init do
     update_rc "-f", @resource[:name], "remove"
     update_rc @resource[:name], "defaults"
   end
+
+  def statuscmd
+    os = Facter.value(:operatingsystem).downcase
+
+    if os == 'debian'
+      majversion = Facter.value(:operatingsystemmajrelease).to_i
+    else
+      majversion = Facter.value(:operatingsystemmajrelease).split('.')[0].to_i
+    end
+
+
+    if ((os == 'debian' && majversion >= 8) || (os == 'ubuntu' && majversion >= 15))
+      # SysVInit scripts will always return '0' for status when the service is masked,
+      # even if the service is actually stopped. Use the SysVInit-Systemd compatibility
+      # layer to determine the actual status. This is only necessary when the SysVInit
+      # version of a service is queried. I.e, 'ntp' instead of 'ntp.service'.
+      (@resource[:hasstatus] == :true) && ["systemctl", "is-active", @resource[:name]]
+    else
+      super
+    end
+  end
 end
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -59,7 +59,7 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
       # flapping when simply trying to disable a masked service.
       return :mask if (@resource[:enable] == :mask) && (svc_info[:LoadState] == 'masked')
       return :true if svc_info[:UnitFileState] == 'enabled'
-      if Facter.value(:osfamily) == 'debian'
+      if Facter.value(:osfamily).downcase == 'debian'
         ret = debian_enabled?(svc_info)
         return ret if ret
       end
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -32,7 +32,14 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
   end
 
   def get_start_link_count
-    Dir.glob("/etc/rc*.d/S??#{@resource[:name]}").length
+    # Start links don't include '.service'. Just search for the service name.
+    if @resource[:name].match(/\.service/)
+      link_name = @resource[:name].split('.')[0]
+    else
+      link_name = @resource[:name]
+    end
+
+    Dir.glob("/etc/rc*.d/S??#{link_name}").length
   end
 
   def enabled?
@@ -72,6 +79,11 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
     return :false
   end
 
+  # This method is required for Debian systems due to the way the SysVInit-Systemd
+  # compatibility layer works. When we are trying to manage a service which does not
+  # have a Systemd unit file, we need to go through the old init script to determine
+  # whether it is enabled or not. See PUP-5016 for more details.
+  #
   def debian_enabled?(svc_info)
     # If UnitFileState == UnitFileState then we query the older way.
     if svc_info[:UnitFileState] == 'UnitFileState'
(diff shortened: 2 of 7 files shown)


whopper:
@kylog yeah, it was never being called. At some point today I had an exact understanding of under what circumstances that block is called, but my brain is melted after going through all of this stuff. Tomorrow I'll do a check again. I'm pretty sure we still need it for regular systemd services (with units). Will experiment.

lib/puppet/provider/service/systemd.rb
```diff
@@ -32,7 +32,14 @@ def disable
   end
 
   def get_start_link_count
-    Dir.glob("/etc/rc*.d/S??#{@resource[:name]}").length
```
whopper: @kylog alright, I figured out why we need this block. When we query for something like 'ntp.service', we hit the systemd provider. (This is that weird thing where systemd manages the service because of the `service` appearing in the name, whereas `ntp` is managed by the debian provider and it's initscript.)

However, because of the sysvinit-systemd compatibility layer, services like `ntp` don't have a proper unit file (i.e, discoverable with `systemctl list-unit-files`). So we need to do the old `invoke-rc.d` stuff here, just like in the Debian provider, to figure out the status of the service.

It's pretty strange, but any service can be queried with `systemctl` in Debian 8 to get the running/stopped status, but services with no unit files depend on the old init scripts to get their enabled/disabled status.

This particular change is needed because although the resource and service name is `name.service`, systemd never stores and start scripts as `SXXname.service`. They are always `SXXname`, regardless of there being a unit file or not.

All in all, the compatibility layer is kind of a gross mess for us at this point.

kylog: Yikes. Good sleuthing. That makes sense (in a Debian-sort of way :)). I wonder if dropping some of your note here in as a comment in the code will help future-me when I squint at this again in six months.

whopper: Yep, done.

whopper:
@kylog one other thing: because of that upstream ubuntu bug, puppet will throw and error whenever we try to enable sysvinit services in ubuntu (but the service still gets enabled):

```
root@kll6h0k1238ywd5:~# systemctl enable ntp
Synchronizing state for ntp.service with sysvinit using update-rc.d...
...
Failed to execute operation: No such file or directory


root@kll6h0k1238ywd5:~# puppet resource service ntp enable=true
Error: Could not enable ntp:
Error: /Service[ntp]/enable: change from false to true failed: Could not enable ntp:
service { 'ntp':
  ensure => 'running',
  enable => 'true',
}

```

Ubuntu may have fixed this at some point, but our vmpooler VMs still have the issue. Should we do anything special on our end to work around this, or make it a documentation thing, or something else?

kylog:
I'd say we just add a note to that effect to the release notes, but don't try to work around it. 

My reasoning is that: it's fixed in 15.10, 15.04 is EOL in Jan 2016 (i.e. pretty soon), and it's not clear if we could detect *this* errors-but-not-really-a-failure from real failures (which I wouldn't want to suppress). I'd rather not incur the risk and tech-debt for a fix with at most a couple months of lifetime.

kylog:
@whopper LGTM. Can I hit the green button or did you want to do more testing?

lib/puppet/provider/service/debian.rb
```diff
@@ -63,4 +63,19 @@ def enable
     update_rc "-f", @resource[:name], "remove"
     update_rc @resource[:name], "defaults"
   end
+
+  def statuscmd
+    os = Facter.value(:operatingsystem).downcase
+    majversion = Facter.value(:operatingsystemmajrelease)
+
+    if ((os == 'debian' && majversion == '8') || (os == 'ubuntu' && majversion == '15.04'))
```
whopper: @kylog re. merging, I was thinking I should make this apply to future debians and ubuntus as well, was going to update, though we could also do that when they are released.

whopper:
@kylog alright, updated. Should be ready to go now.

PR #4315, opened 2015-10-02
(maint) Skip puppet / mco service test for non-AIO runs

whopper:


whopper:
Is stable the right place for this?

kylog:
Makes sense.

PR #4316, opened 2015-10-02
(PUP-5082) Add relationship validator

thallgren:
Adds a relationship validator to the compiler. The validator currently asserts that only 'consume', 'require', and 'export' are used when the RHS is referencind a capability resource.


lib/puppet/parser/catalog_validator.rb
```diff
@@ -0,0 +1,33 @@
+# Abstract class for a catalog validator that can be registered with the compiler to run at
+# a certain stage.
+module Puppet::Parser
```
hlindberg: Why in "Parser" as opposed to say "Catalog" namespace ?

lib/puppet/parser/environment_compiler.rb
```diff
@@ -32,12 +37,14 @@ def compile
 
       Puppet::Util::Profiler.profile("Env Compile: Prune", [:compiler, :prune_catalog]) { prune_catalog }
 
-      Puppet::Util::Profiler.profile("Env Compile: Validate Catalog", [:compiler, :validate_site_catalog]) { validate_site_catalog }
+      Puppet::Util::Profiler.profile("Env Compile: Validate Catalog pre-finish", [:compiler, :validate_pre_finish]) { validate_catalog(Puppet::Parser::CatalogValidator::PRE_FINISH) }
```
hlindberg: suggest breaking line to make it more readable

lib/puppet/parser/catalog_validators/site_validator.rb
```diff
@@ -0,0 +1,22 @@
+module Puppet::Parser
+  module CatalogValidators
+    # Validator that asserts that only application components can appear inside a site.
+  class SiteValidator < CatalogValidator
+
+      def self.stage?(stage)
```
hlindberg: the term 'stage' is asily confused with 'puppet stage' - suggest a more descriptive name 'validation_stage', or 'compilation_stage' or some such.

hlindberg:
Looks good - only minor things to address.

lib/puppet/parser/catalog_validator.rb
thallgren: The Compiler is in Parser namespace (which admittedly is a bit odd). Validation is performed by the compiler. In my view, the catalog is the result of compilation and the validation is a part of that.

PR #4321, opened 2015-10-07
(PUP-4055) Add DNF package provider

whopper:
This series of commits adds a new DNF provider to Puppet, which is necessary due to Fedora's move from Yum to DNF.

A few updates to the Yum provider were required to make way for the new provider, which is its child.

### DIFF
--- a/acceptance/lib/puppet/acceptance/rpm_util.rb
+++ b/acceptance/lib/puppet/acceptance/rpm_util.rb
@@ -4,19 +4,30 @@ module Puppet
       # Utilities for creating a basic rpm package and using it in tests
       @@defaults = {:repo => '/tmp/rpmrepo', :pkg => 'mypkg', :publisher => 'tstpub.lan', :version => '1.0'}
 
+      def rpm_provider(agent)
+        has_dnf = on(agent, 'which dnf', :acceptable_exit_codes => [0,1]).exit_code
+        if has_dnf == 0
+          'dnf'
+        else
+          'yum'
+        end
+      end
+
       def setup(agent)
+        cmd = rpm_provider(agent)
         required_packages = ['createrepo', 'rpm-build']
         required_packages.each do |pkg|
-          unless ((on agent, "yum list installed #{pkg}", :acceptable_exit_codes => (0..255)).exit_code == 0) then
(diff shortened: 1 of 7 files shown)


whopper:
Retargeted version of https://github.com/puppetlabs/puppet/pull/4308

acceptance/lib/puppet/acceptance/rpm_util.rb
```diff
@@ -4,19 +4,31 @@ module RpmUtils
       # Utilities for creating a basic rpm package and using it in tests
       @@defaults = {:repo => '/tmp/rpmrepo', :pkg => 'mypkg', :publisher => 'tstpub.lan', :version => '1.0'}
 
+      def rpm_provider(agent)
+        os = on(agent, facter('operatingsystem')).stdout.chomp
+        maj_version = on(agent, facter('operatingsystemmajrelease')).stdout.chomp
+        if os == 'Fedora' && maj_version.to_i >= 22
```
branan: Can we instead just test if the DNF executable exists, so we don't have to update this when other RPM-based distros move over?

lib/puppet/provider/package/dnf.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  This provider supports the `install_options` attribute, which allows command-line flags to be passed to dnf.
+  These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
+  or an array where each element is either a string or a hash."
+
+  has_feature :install_options, :versionable, :virtual_packages
+
+  commands :cmd => "dnf", :rpm => "rpm"
+
+  if command('rpm')
```
branan: Could we please get a comment here explaining why this was borrowed from the YUM provider and how note that we should investigate removing it

branan:
Besides my nits above, this looks reasonable

whopper:
@branan added that comment and updated `rpm_util.rb` to look for the dnf binary!

branan:
Just waiting on appveyor (not that this should affect appveyor)

branan:
It doesn't look like Appveyor wants to pick up on this PR. Since this is just linux-related changes I'm going to just push the button

PR #4322, opened 2015-10-07

whopper:
query packages.

### DIFF
--- a/lib/puppet/provider/package/rpm.rb
+++ b/lib/puppet/provider/package/rpm.rb
@@ -281,6 +281,7 @@ Puppet::Type.type(:package).provide :rpm, :source => :rpm, :parent => Puppet::Pr
       self::NEVRA_FIELDS.zip(match.captures) { |f, v| hash[f] = v }
       hash[:provider] = self.name
       hash[:ensure] = "#{hash[:version]}-#{hash[:release]}"
+      hash[:ensure].prepend("#{hash[:epoch]}:") if hash[:epoch] != '0'
     else
       Puppet.debug("Failed to match rpm line #{line}")
     end
--- a/spec/unit/provider/package/rpm_spec.rb
+++ b/spec/unit/provider/package/rpm_spec.rb
@@ -12,6 +12,7 @@ describe provider_class do
     chkconfig 0 1.3.30.2 2.el5 x86_64
     myresource 0 1.2.3.4 5.el4 noarch
     mysummaryless 0 1.2.3.4 5.el4 noarch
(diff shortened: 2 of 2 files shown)


whopper:
Retargeted version of https://github.com/puppetlabs/puppet/pull/4285, /cc @joshcooper

whopper:
Comments on the other PR from Josh:

The rpm provider uses [`Puppet::Util::Package.versioncmp`](https://github.com/whopper/puppet/blob/PUP-5025/RPM-epoch/lib/puppet/provider/package/rpm.rb#L142) to compare ensure values, but that method doesn't know how to handle epochs in version strings, such as what we're returning here.

Do we want to update `versioncmp` to handle epochs?

Or implement an `rpmversioncmp` that knows how to handle epochs? Based on http://www.rpm.org/max-rpm-snapshot/s1-rpm-depend-manual-dependencies.html:

>  Epoch is the most significant component of a package's complete version identifier with regards to RPM's version comparison algorithm.

I assume we'd want to first compare epochs, defaulting to epoch 0 if one is not specified. If the epochs are different, then return +/-1. If the epochs are the same, then delegate to the original `versioncmp`?

Assuming epochs are specific to rpms, I think we'd want to do the latter. Thoughts?

whopper:
My reply to Josh's comment:

@joshcooper hm, it looks like the only places the RPM provider uses `versioncmp` is to check the version of RPM itself though, right?

From what I recall, the package that's actually being installed gets it's ensure version compared in the yum provider,  at this method: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/yum.rb#L226

From my previous (maybe flawed) understanding, it's the child providers of `rpm` that deal with the epoch. `rug`, for example, says: `# XXX: We don't actually deal with epochs here.` in it's `install` method. `zypper` and `urpmi` also do this. `Yum` seems to be the only provider that does any extensive version comparisons at all, for some reason.

(Also, there actually already is an `rpmvercmp` method, which is pretty gnarly: https://github.com/whopper/puppet/blob/PUP-5025/RPM-epoch/lib/puppet/provider/package/rpm.rb#L182)

Yum is the only thing that uses it, and it passes in pieces of the release string in multiple steps, so the epoch ends up getting sent through it for comparison.

peterhuene:
This makes sense to me.  @joshcooper any feedback left to give?

joshcooper:
Ah, thanks @whopper 

It looks like we call `versioncmp` when trying to figure out if rpm 4.1 is installed. But the current version is extracted from `rpm --version`. If that returned epoch info, then we'd have a problem, but it would be preexisting and unrelated to this issue.

PR #4325, opened 2015-10-07
(maint) Refactor reloading in suse init script

haus:
The force-reload and reload init script actions had the exact same bash
code, so this commit moves them into the same block in the case
statement and consolidates the comments.

### DIFF
--- a/ext/suse/client.init
+++ b/ext/suse/client.init
@@ -104,22 +104,9 @@ case "$1" in
         # Remember status and be quiet
         rc_status
         ;;
-    force-reload)
-        ## Signal the daemon to reload its config. Most daemons
-        ## do this on signal 1 (SIGHUP).
-        ## If it does not support it, restart.
-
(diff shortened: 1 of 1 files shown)


joshcooper:
The comment indicates that the behaviors for `reload` and `force-reload` are different:

> Like force-reload, but if daemon does not support signalling, do nothing 

Should the behaviors actually be different? In which case, we shouldn't be collapsing them? Or was the comment just plain wrong?

/cc @shrug @mckern @geoffnichols

haus:
@joshcooper I'm not sure where the comment was lifted from initially. I read a fair number of scripts and could not find any where there were differences.

haus:
Ah! @joshcooper http://refspecs.linuxbase.org/LSB_3.1.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html says force-reload should reload if the service supports it, which puppet does, or restart if not.

PR #4326, opened 2015-10-08
(PUP-5340) Accept catalog_cache_terminus for apply

kylog:
Previously, the 'puppet agent' command accepted catalog_cache_terminus,
but 'puppet apply' did not. This makes 'apply' match 'agent'.

### DIFF
--- a/lib/puppet/application/apply.rb
+++ b/lib/puppet/application/apply.rb
@@ -315,6 +315,10 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
 
     Puppet.settings.use :main, :agent, :ssl
 
+    if Puppet[:catalog_cache_terminus]
+      Puppet::Resource::Catalog.indirection.cache_class = Puppet[:catalog_cache_terminus]
+    end
+
     # we want the last report to be persisted locally
     Puppet::Transaction::Report.indirection.cache_class = :yaml
 
--- a/spec/unit/application/apply_spec.rb
+++ b/spec/unit/application/apply_spec.rb
@@ -470,4 +470,12 @@ describe Puppet::Application::Apply do
       @apply.send(:apply_catalog, catalog)
     end
   end
+
+  it "should honor the catalog_cache_terminus setting" do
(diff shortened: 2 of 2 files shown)


kylog:
@joshcooper this is the thing I mentioned to you at the contributor summit.

joshcooper:
LGTM. One surprising behavior is that both catalog_terminus and catalog_cache_terminus are settings of type `terminus`, however, their behaviors are different. Puppet automatically "wires up" the former, but not the latter.

The magic is that the TerminusSetting converts the [setting value to a symbol](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/settings/terminus_setting.rb#L7), e.g. `:json`, and when puppet tries to resolve the terminus for an indirected type, it will try to lookup a [registered terminus with that name]( https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/terminus.rb#L32-L33). That way you don't have to call `Puppet::Resource::Catalog.indirection.terminus=` directly. But puppet doesn't treat the `cache_class` the same way, and it probably should (a separate bug I'd say).

```
0 ~/work/puppet (master)  $ ls  ~/.puppetlabs/opt/puppet/cache/client_data/catalog/
ls: /Users/josh/.puppetlabs/opt/puppet/cache/client_data/catalog/: No such file or directory
1 ~/work/puppet (master)  $ be puppet apply -e "notify { 'foo': }" --catalog_cache_terminus json
Notice: Compiled catalog for arcturus.delivery.puppetlabs.net in environment production in 0.03 seconds
Notice: foo
Notice: /Stage[main]/Main/Notify[foo]/message: defined 'message' as 'foo'
0 ~/work/puppet (master)  $ jq .  ~/.puppetlabs/opt/puppet/cache/client_data/catalog/arcturus.delivery.puppetlabs.net.json  | head
{
  "tags": [
    "settings"
  ],
  "name": "arcturus.delivery.puppetlabs.net",
  "version": 1444800118,
  "code_id": null,
  "environment": "production",
  "resources": [
    {
...
```

PR #4327, opened 2015-10-08
(PUP-5089) Allow classes to be application components

thallgren:
This commit makes it possible for a class to be an application
component by adding the following behavior:

The parser will accept the syntax:

Class[literal] consumes Cap
Class[literal] produces Cap

The parser AST model is changed so that the LHS of a
CapabilityMapping is an Expression rather than just a String
and the validation was enhanced to cope with that.

The name of the LHS in the CapabilityMapping was changed form
'resource_name' to 'component' since that is more accurate
now.

The evaluator will do either a find_definition or a
find_hostclass depending on what type of expression that is
found in 'component'.

A host class is considered a valid producer during evaluation.


spec/unit/capability_spec.rb
```diff
@@ -359,5 +359,51 @@ def mock_cap_finding
         expect(graph.dependents(capability).map {|d| d.to_s }).to include('Consumer[x]')
       end
     end
+
+    describe 'producing/consuming resources to/from classes' do
```
hlindberg: "classes producing and consuming" ?

hlindberg:
needs tests that export and consume via meta parameters also works

thallgren:
Not sure what you mean. `export` and `consume` are the two meta parameters that are part of the capability mapping and they are tested here. Perhaps easier if you give an example.

hlindberg:
I meant using the resource style instantiation of a class, and using the meta parameters in that instantiation e.g.
```
class  { 'foo': consumes => Cap }
```

hlindberg:
Think we also need to test that a reference to a class works in applications and the environment compiler.

thallgren:
I have a limited understanding of how this is supposed to work but the way I've understood it, the `consumes` keyword should be used to define that a type (or class) `consumes` some capability  type. This keyword should not be used during _instantiation_ of that type (or class). During instantiation, we can set the meta attribute `consume` to appoint a reference to a _specific_ instantiation of the capability resource (i.e. with type and title). Is this wrong?

PR #4328, opened 2015-10-08
(PUP-5315) Fix acceptance tests for AIX

johnduarte:
This commit patches acceptance tests for AIX. These changes fall
into three major categories.

  * PMT - skip tests due to lack of SSL cert chain for AIX
  * SSL - remove `-k` flag from curl calls
  * user/group - AIX limits user and group names to 9 characters
    by default. Test user and group names have been shorted.

### DIFF
--- a/acceptance/lib/puppet/acceptance/temp_file_utils.rb
+++ b/acceptance/lib/puppet/acceptance/temp_file_utils.rb
@@ -140,17 +140,13 @@ module Puppet
       # value containing only the file mode, excluding the type, e.g
       # S_IFDIR 0040000
       def stat(host, path)
-        stat_command = case host['platform']
-                       when /osx/
-                         "stat -f '%Su:%Sg:%p'"
-                       else
-                         "stat --format '%U:%G:%a'"
-                       end
+        require File.join(File.dirname(__FILE__),'common_utils.rb')
+        ruby = Puppet::Acceptance::CommandUtils.ruby_command(host)
+        owner = on(host, "#{ruby} -e 'require \"etc\"; puts (Etc.getpwuid(File.stat(\"#{path}\").uid).name)'").stdout.chomp
+        group = on(host, "#{ruby} -e 'require \"etc\"; puts (Etc.getgrgid(File.stat(\"#{path}\").gid).name)'").stdout.chomp
+        mode  = on(host, "#{ruby} -e 'puts (File.stat(\"#{path}\").mode & 07777)'").stdout.chomp.to_i
 
-        permissions = on(host, "#{stat_command} #{path}").stdout.chomp
-        owner, group, mode = permissions.split(':')
-
-        [owner, group, mode.to_i(8) & 07777]
+        [owner, group, mode]
       end
 
       def initialize_temp_dirs()
--- a/acceptance/tests/environment/can_enumerate_environments.rb
+++ b/acceptance/tests/environment/can_enumerate_environments.rb
@@ -54,18 +54,16 @@ if master.is_pe?
 end
 
 with_puppet_running_on(master, master_opts) do
-  agents.each do |agent|
-    step "Ensure that an unauthenticated client cannot access the environments list" do
-      on agent, "curl --tlsv1 -ksv https://#{master}:#{master_port(agent)}/puppet/v3/environments", :acceptable_exit_codes => [0,7] do
-        assert_match(/< HTTP\/1\.\d 403/, stderr)
-      end
+  step "Ensure that an unauthenticated client cannot access the environments list" do
+    on master, "curl --tlsv1 -ksv https://#{master}:#{master_port(master)}/puppet/v3/environments", :acceptable_exit_codes => [0,7] do
+      assert_match(/< HTTP\/1\.\d 403/, stderr)
     end
+  end
 
-    step "Ensure that an authenticated client can retrieve the list of environments" do
-      curl_master_from(agent, '/puppet/v3/environments') do
-        data = JSON.parse(stdout)
-        assert_equal(["env1", "env2", "production"], data["environments"].keys.sort)
-      end
+  step "Ensure that an authenticated client can retrieve the list of environments" do
+    curl_master_from(master, '/puppet/v3/environments') do
+      data = JSON.parse(stdout)
+      assert_equal(["env1", "env2", "production"], data["environments"].keys.sort)
     end
   end
 end
--- a/acceptance/tests/modules/install/already_installed.rb
+++ b/acceptance/tests/modules/install/already_installed.rb
@@ -3,7 +3,7 @@ require 'puppet/acceptance/module_utils'
 extend Puppet::Acceptance::ModuleUtils
 
 hosts.each do |host|
-  skip_test "skip tests requiring forge certs on solaris and aix" if host['platform'] =~ /solaris/
+  skip_test "skip tests requiring forge certs on solaris and aix" if host['platform'] =~ /solaris|aix/
 end
 
 module_author = "pmtacceptance"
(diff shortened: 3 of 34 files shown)


johnduarte:
@joshcooper this is the AIX PR that I was referring to in our discussion yesterday.

cc: @branan

branan:
@johnduarte Does this also stub out the tests that you've ticketed separately?

acceptance/lib/puppet/acceptance/temp_file_utils.rb
```diff
@@ -140,17 +140,13 @@ def chmod(host, mode, path)
...
-                       when /osx/
-                         "stat -f '%Su:%Sg:%p'"
-                       else
-                         "stat --format '%U:%G:%a'"
-                       end
+        require File.join(File.dirname(__FILE__),'common_utils.rb')
+        ruby = Puppet::Acceptance::CommandUtils.ruby_command(host)
+        owner = on(host,"ls -l -d #{path} | awk '{print $3;}'").stdout.chomp
+        group = on(host,"ls -l -d #{path} | awk '{print $4;}'").stdout.chomp
+        mode = on(host, "#{ruby} -e 'puts (File.stat(\"#{path}\").mode & 07777)'").stdout.chomp.to_i
```
joshcooper: since we're calling File.stat to get mode, you could have it return owner and group as well

acceptance/tests/environment/can_enumerate_environments.rb
```diff
@@ -56,7 +56,7 @@ def curl_master_from(agent, path, headers = '', &block)
 with_puppet_running_on(master, master_opts) do
   agents.each do |agent|
     step "Ensure that an unauthenticated client cannot access the environments list" do
-      on agent, "curl --tlsv1 -ksv https://#{master}:#{master_port(agent)}/puppet/v3/environments", :acceptable_exit_codes => [0,7] do
+      on agent, "curl --tlsv1 -s -v https://#{master}:#{master_port(agent)}/puppet/v3/environments", :acceptable_exit_codes => [0,7] do
```
joshcooper: This will break non-AIX platforms, because we're not specifying the `--cacert` option and curl will by default reject the SSL connection (which is why we're specifying the `-k` option).

acceptance/tests/modules/install/already_installed.rb
```diff
@@ -3,7 +3,7 @@
 extend Puppet::Acceptance::ModuleUtils
 
 hosts.each do |host|
-  skip_test "skip tests requiring forge certs on solaris and aix" if host['platform'] =~ /solaris/
+  skip_test "skip tests requiring forge certs on solaris and aix" if host['platform'] =~ /solaris|aix/
```
joshcooper: Ideally this wouldn't be necessary as we should be including the ca-cert-bundle on all non-Windows platforms: https://github.com/puppetlabs/puppet-agent/blob/master/configs/projects/puppet-agent.rb#L90

acceptance/tests/security/cve-2013-1652_improper_query_params.rb
```diff
@@ -21,7 +21,7 @@
 
       cert_path = on(agent, puppet('agent', "--configprint hostcert")).stdout.chomp
       key_path = on(agent, puppet('agent', "--configprint hostprivkey")).stdout.chomp
-      curl_base = "curl --tlsv1 -g --cert \"#{cert_path}\" --key \"#{key_path}\" -k -H 'Accept: pson'"
```
joshcooper: same curl comment

acceptance/tests/security/cve-2013-4761_resource_type.rb
```diff
@@ -46,7 +46,7 @@
         payload = "https://#{master}:8140/puppet/v3/resource_type/#{type_name}?environment=production"
         cert_path = on(agent, puppet("agent", "--configprint hostcert")).stdout.chomp
         key_path = on(agent, puppet("agent", "--configprint hostprivkey")).stdout.chomp
-        curl_base = "curl --tlsv1 -g --cert \"#{cert_path}\" --key \"#{key_path}\" -k -H 'Accept: pson'"
+        curl_base = "curl --tlsv1 -g --cert \"#{cert_path}\" --key \"#{key_path}\" -H 'Accept: pson'"
```
joshcooper: same curl comment

acceptance/tests/resource/user/should_not_create_existing.rb
```diff
@@ -1,7 +1,7 @@
 test_name "tests that user resource will not add users that already exist."
 
-user  = "user#{rand(999999).to_i}"
-group = "group#{rand(999999).to_i}"
+user  = "u#{rand(999999).to_i}"
+group = "g#{rand(999999).to_i}"
```
joshcooper: gotta love AIX and their 8 character restriction

acceptance/tests/resource/user/should_modify_gid.rb
```diff
@@ -1,5 +1,6 @@
 test_name "verify that we can modify the gid"
 confine :except, :platform => 'windows'
+confine :except, :platform => 'aix' # file ticket
```
joshcooper: needs ticket

johnduarte:
@branan no, this does not address the issues that were called out separately.

acceptance/tests/environment/can_enumerate_environments.rb
johnduarte: @joshcooper So, does that leave us in a position where all curl tests need to be skipped on AIX because it does not support the `-k` option?

branan:
@johnduarte sweet, I just wanted to be sure that getting this merged wouldn't make debugging other tickets confusing

acceptance/tests/environment/can_enumerate_environments.rb
joshcooper: I think you could switch the curl options based on agent platform, so only include `-k` on platforms other than AIX

branan: I feel like I solved this problem... let me check some stuff before anyone does any work :)

branan: Hmm, nope. Looks like I only solved it in one place - `wait_for_host_in_dashboard` in beaker, and not in any generalized curl helper :(

There is a `curl_on` helper in Beaker that could be fixed to handle the -k case, though

joshcooper: Since the test is testing that only authenticated agents can enumerate environments, we could just run the curl command on the master, and eliminate the issues with AIX and OSX. That way we know we're on a sane system that supports `curl -k` with and without client certs.

joshcooper:
nit, the commit message says "9 characters by default", but I think the actual limit is 8, which is why `pe-puppet` was a problem.

johnduarte:
off my 1 error. :-) thanks.

acceptance/tests/environment/can_enumerate_environments.rb
johnduarte: @joshcooper I don't follow you. This test also ensures that unauthenticated agents cannot enumerate environments.

Also, since we are on the stable branch, we do not have the ca-cert bundle in place to run this test on AIX regardless of the curl command. Seems like we should skip on Solaris and AIX and address this in the master branch.

Thoughts?

johnduarte:
Is there a way to re-run appveyor? It looks like this run is not going to return results.

ferventcoder:
I'm guessing the commit message was updated to include acceptance tests? We normally skip ci for commits that are only about acceptance or documentation changes -  https://github.com/puppetlabs/puppet/blob/master/appveyor.yml#L3

johnduarte:
Appveyor skips PRs containing `acceptance test`. So, this is not waiting on additional checks.

@joshcooper and @branan does this look ready to go to you?

PR #4330, opened 2015-10-08
(PUP-5189) Add beaker node defs for Win10(x86,x64)

johnduarte:
[skip-ci]


whopper:
@johnduarte we only need the one PR against stable and we can merge it up to master

PR #4331, opened 2015-10-09
(maint) clarify node lookup failure

wkalt:
I was encountering this error and found it hard to understand. I think this
wording is a little closer to the intent.


kylog:
Travis lost connection to the ruby 2.2 vm, but that's unrelated to this :)

PR #4332, opened 2015-10-09
(PUP-5274) Test generated ca is not world readable

johnduarte:
This commit adds an acceptance test to validate that the private
ca certificate is not world readable when generated via
`puppet cert generate`


whopper:
@johnduarte does this test from Josh's 3.x PR cover this adequately? If so, we might already have this covered. https://github.com/puppetlabs/puppet/blob/3.x/acceptance/tests/ssl/ticket_5274_private_key_modes.rb#L22-L34

johnduarte:
Indeed @whopper my bad.

whopper:
@johnduarte haha I thought I remembered that test but thought I was going crazy when I couldn't find it in stable or master :P

PR #4333, opened 2015-10-12
(PUP-5294) Do not validate parameters of unevaluated consumers

thallgren:
When compiled with the environment compiler, consumers of capability
resources that has non-default parameters will fail with a validation
error on parameters that are expected to be consumed from the
capability. At validation time only the type and title of the
capability is known.

This commit relaxes the parameter validation when the environment
compiler is used. The compiler will detect resources that consumes
capabilities (using require or consume meta parameters) and mark such
resources as `unevaluated_consumer`. The validation of such resources
will not consider a missing but required parameter an error.

### DIFF
--- a/lib/puppet/resource/type.rb
+++ b/lib/puppet/resource/type.rb
@@ -412,7 +412,7 @@ class Puppet::Resource::Type
   private :assign_defaults
 
   def validate_resource_hash(resource, resource_hash)
-    Puppet::Pops::Types::TypeMismatchDescriber.validate_parameters(resource.to_s, parameter_struct, resource_hash)
+    Puppet::Pops::Types::TypeMismatchDescriber.validate_parameters(resource.to_s, parameter_struct, resource_hash, resource.is_unevaluated_consumer?)
   end
   private :validate_resource_hash
 
--- a/spec/unit/appmgmt_spec.rb
+++ b/spec/unit/appmgmt_spec.rb
@@ -47,6 +47,7 @@ describe "Application instantiation" do
       application app {
         prod { one: host => ahost, export => Cap[cap] }
         cons { two: host => ahost, consume => Cap[cap] }
+        cons { three: consume => Cap[cap] }
       }
 
       site {
@@ -358,7 +359,7 @@ EOS
       app = apps.first
       expect(app["nodes"]).not_to be_nil
       comps = catalog.direct_dependents_of(app).map(&:ref).sort
-      expect(comps).to eq(["Cons[two]", "Prod[one]"])
+      expect(comps).to eq(["Cons[three]", "Cons[two]", "Prod[one]"])
 
       prod = catalog.resource("Prod[one]")
       expect(prod).not_to be_nil
--- a/lib/puppet/parser/environment_compiler.rb
+++ b/lib/puppet/parser/environment_compiler.rb
@@ -110,6 +110,7 @@ class Puppet::Parser::EnvironmentCompiler < Puppet::Parser::Compiler
     unless resource.class?
       @catalog.add_edge(scope.resource, resource)
     end
+    resource.mark_unevaluated_consumer if is_capability_consumer?(resource)
     assert_app_in_site(scope, resource)
   end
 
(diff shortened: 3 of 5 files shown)


spec/unit/appmgmt_spec.rb
```diff
@@ -46,7 +46,7 @@ def compile_to_env_catalog(string)
 
       application app {
         prod { one: host => ahost, export => Cap[cap] }
-        cons { two: host => ahost, consume => Cap[cap] }
```
hlindberg: Think it is of value to test the removed case as well.

lib/puppet/resource/type.rb
```diff
@@ -413,6 +413,9 @@ def assign_defaults(resource, scope, resource_hash)
 
   def validate_resource_hash(resource, resource_hash)
     Puppet::Pops::Types::TypeMismatchDescriber.validate_parameters(resource.to_s, parameter_struct, resource_hash)
+  rescue Puppet::ParseError => e
```
hlindberg: This is quite an ugly fix. Do we really need to do it this way? Isn't it possible to detect the special case beforehand and validate differently (or not at all) ?

thallgren: It's possible to test half of it. There's no backtrace unless an error is rescued. Also testing something that is an exceptional case (environment compiler + consume meta parameter) everytime some resource parameters are validated is seems rather unnecessary.

spec/unit/appmgmt_spec.rb
thallgren: What would the removed test really test though, that isn't tested on the previous or next line?

hlindberg: One is a path without problem (the one removed), the other exposes a problem = different code paths. Line 48 is for export.

lib/puppet/resource/type.rb
hlindberg: I guess I do not understand the state a resource may be in when the problem occurs. If I understand the proposed fix, it catches any problem that is found by parameter validation if the resource has a "consumes" in it during an "environment compilation"). Doesn't that include bona fide errors? 
The fix is also brittle since it is sensitive to the name/impl of the environment compiler.

I would much rather see resources that have the special trait be marked as having a special state ("partially specified") and that such resources only may exist under some conditions and that it is an error to have resources in that state under other conditions (e.g. when finishing a node catalog)

thallgren: It doesn't catch "any problem". It will just catch the case when a parameter expects a value, the environment compiler is on the stack, and the 'consume' metaparameter is present. It will re-raise the error for all other cases.

But this IS indeed a dirty fix. The real fix IMO would be to do what you suggest in the JIRA and I've suggested earlier, to use real types in the type system to describe the capabilities. Not sure I see the point in elaborating for less dirtyness.

spec/unit/appmgmt_spec.rb
thallgren: What I meant was, if I add the {{host => ahost}} declaration in a test that in all other aspects is equal to the one on line 49, then what would that actually test? That a parameter can be assigned? We know it can and there are several other tests that verifies it. What we want to test here is that the lack of such assignment doesn't cause a problem.

hlindberg: You are testing two things with one test; a) that consume works and b) that missing parameters works - seems best to have two separate test cases.

lib/puppet/resource/type.rb
hlindberg: ok - as a dirty fix there needs to be a ticket to fix it the right way, and a comment in the code that explains the dirty fix and that mentions the ticket with the real fix.

spec/unit/appmgmt_spec.rb
thallgren: A missing parameter doesn't work. It will yield and error. So there's no such thing as testing that "a missing parameter works". This test verifies that consume works, period. It tests that consume results in that a parameter that otherwise would be missing isn't missing.

lib/puppet/resource/type.rb
thallgren: Agreed, although I think that the "real fix" is part of a bigger refactoring/rewrite that isn't fully scoped and understood yet. Do you want me to add a JIRA suggesting that we use the Puppet Type system for Capabilities?

hlindberg:
Much better! Merging.

PR #4334, opened 2015-10-12
(PUP-5351) AuthConfigLoader hook for Puppet Server

camlow325:
This commit provides an accessor for controlling the AuthConfigLoader
class which is used to load an authorization config.  This would be used
by Puppet Server to disable the default AuthConfigLoader when it is
configured to use its own trapperkeeper-based authorization instead.

### DIFF
--- a/spec/unit/network/authorization_spec.rb
+++ b/spec/unit/network/authorization_spec.rb
@@ -1,5 +1,7 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
+require 'puppet/network/http'
+require 'puppet/network/http/api/indirected_routes'
 require 'puppet/network/authorization'
 
 describe Puppet::Network::Authorization do
@@ -9,7 +11,7 @@ describe Puppet::Network::Authorization do
 
   subject { AuthTest.new }
 
-  describe "when creating an authconfig object" do
+  context "when creating an authconfig object" do
     before :each do
       # Other tests may have created an authconfig, so we have to undo that.
       @orig_auth_config = Puppet::Network::AuthConfigLoader.instance_variable_get(:@auth_config)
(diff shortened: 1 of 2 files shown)


spec/unit/network/authorization_spec.rb
```diff
@@ -1,5 +1,7 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
+require 'puppet/network/http'
```
camlow325: I only added these requires because without them, it wasn't possible to run the `authorization_spec` tests in isolation.  Without these, you'd get an `undefined Puppet::Network::Authorization` error when trying to run `bundle exec rake spec TEST=spec/unit/network/authorization_spec.rb`.

camlow325:
@joshcooper Would appreciate it if you could take a look at this one when you have a chance.  This is basically just an extension of the work that we'd done in PUP-5012 to enable Puppet Server with a way to bypass the default authorization logic in favor of its own Clojure/Trapperkeeper based one.  We found that with the prior work, the auth.conf file would still be loaded (even though it wasn't used later on).  The loading of the file, though, has the side effect of logging info about parsed rules - which could be confusing to a user who had thought they had turned off Ruby-based auth.conf under Puppet Server.

camlow325:
@nwolfe ping.

nwolfe:
Yep, looks correct. :+1:

camlow325:
@dankreek - here's my other PR I was hoping you could review when you have time.  Thanks!

dankreek:
This looks pretty cut and dried to me. :+1:

nwolfe:
:sheep: :it:

camlow325:
@joshcooper We went ahead and merged this since it seemed low risk and it is blocking some other work in Puppet Server.  We will watch CI to ensure that nothing was broken.  I'd definitely appreciate any additional feedback you might have on this when/if you get a chance to look at it later on.  Thanks!

PR #4336, opened 2015-10-12
(PUP-5271) ADSI User - Do not manage unspecified password

ferventcoder:
Currently when the password policy for "Password must meet complexity
requirements" is enabled, attempts to have puppet create Windows users
without passwords fail with result `"OLE error code:800708C5 in Active
Directory. The password does not meet the password policy requirements.
Check the minimum password length, password complexity and password
history requirements."`

This behavior is expected when the password is specified but doesn't meet
the password policy, but not expected when the user has not attempted to
manage the password, e.g. leaving it empty. Correct Puppet not to try to set
the password unless it is specified.

Note that when the password policy "Minimum password length" is greater than
0, the password must always be specified. This is due to how Windows
validates new user creation and it is not possible to leave password
unspecified once that password policy is set.

Also it's important to note that when a user is specified with
`managehome => true`, the password must always be specified if it is not
an already existing user with a password.

### DIFF
--- a/spec/unit/util/windows/adsi_spec.rb
+++ b/spec/unit/util/windows/adsi_spec.rb
@@ -203,6 +203,19 @@ describe Puppet::Util::Windows::ADSI, :if => Puppet.features.microsoft_windows? 
         user.password = 'pwd'
       end
 
+       it "should be able manage a user without a password" do
+        adsi_user.expects(:SetPassword).with('pwd').never
+        adsi_user.expects(:SetInfo).at_least_once
+
+        flagname = "UserFlags"
+        fADS_UF_DONT_EXPIRE_PASSWD = 0x10000
+
+        adsi_user.expects(:Get).with(flagname).returns(0)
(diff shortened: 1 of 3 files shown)


ferventcoder:
Oops, missed fixing up an issue.

spec/unit/util/windows/adsi_spec.rb
```diff
@@ -203,6 +203,19 @@
         user.password = 'pwd'
       end
 
+       it "should be able manage a user without a password" do
+        adsi_user.expects(:SetPassword).with('pwd').never
+        adsi_user.expects(:SetInfo).at_least_once.never
```
joshcooper: `at_least_once.never` doesn't look right, I'd expect it to be one or the other?

ferventcoder: Yes, let me fix up.

ferventcoder: fixed.

PR #4337, opened 2015-10-13
(PUP-5360) Feature/autorelation defined types

asasfu:
This would add support for defined types to be used in autorelation methods inside of custom types.  I have tested it but maybe an rspec test should be created for this?

### DIFF
--- a/lib/puppet/type.rb
+++ b/lib/puppet/type.rb
@@ -2117,8 +2117,17 @@ end
     auto_rel = "eachauto#{rel_type}".to_sym
 
     self.class.send(auto_rel) { |type, block|
-      # Ignore any types we can't find, although that would be a bit odd.
-      next unless Puppet::Type.type(type)
+      if not Puppet::Type.type(type)
+        # We know it is not a type at this point, let's see if it exists as a defined type in the catalog
+        # We split and join on :: because defined types will likely have colon separators
+        # All defined types are of type :component
+        found_defines = rel_catalog.resources.select do |x|
+          x.type == :component && x.title_class == type.to_s.split('::').map(&:capitalize).join('::')
+        end
+        next if found_defines.empty?
+        # If this passes then it means the catalog contains a defined type of that name and to allow going forward
+        # with the block to add relations
+      end
 
       # Retrieve the list of names from the block.
       next unless list = self.instance_eval(&block)
--- a/lib/puppet/type/component.rb
+++ b/lib/puppet/type/component.rb
@@ -49,6 +49,11 @@ Puppet::Type.newtype(:component) do
     reference.to_s
   end
 
+  # This allows us to grab the defined type 'type' for optional use in other custom type's autorelation methods
+  def title_class
+    reference.type
+  end
+
   # We want our title to just be the whole reference, rather than @title.
   def title
     ref
--- a/spec/unit/type/component_spec.rb
+++ b/spec/unit/type/component_spec.rb
@@ -21,6 +21,16 @@ describe component do
     expect(comp.title).to eq(comp.ref)
   end
 
+  it "should have a reference type equivalent to its calling defined type" do
+    comp = component.new(:name => "Foo[bar]")
+    expect(comp.title_class).to eq("Foo")
+  end
+
+  it "should have a reference type equivalent to its calling defined type with colons" do
+    comp = component.new(:name => "Foo::Sub[bar]")
+    expect(comp.title_class).to eq("Foo::Sub")
+  end
+
   it "should not fail when provided an invalid value" do
     comp = component.new(:name => "Foo[bar]")
     expect { comp[:yayness] = "ey" }.not_to raise_error
--- a/spec/unit/type_spec.rb
+++ b/spec/unit/type_spec.rb
@@ -449,6 +449,174 @@ describe Puppet::Type, :unless => Puppet.features.microsoft_windows? do
     end
   end
 
+  context "autorelations of defined types" do
+    before :each do
+      @myral = <<-MANIFEST
+        define module::dummy {
+        }
+        module::dummy { 'dummyname': }
+      MANIFEST
+    end
+
+    describe "when building autorelations" do
+      it "should be able to autorequire resources" do
+        type = Puppet::Type.newtype(:autorelation_two) do
+          newparam(:name) { isnamevar }
+          autorequire('module::dummy') { ['dummyname'] }
+        end
+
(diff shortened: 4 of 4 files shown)


lib/puppet/type.rb
```diff
@@ -2117,8 +2117,17 @@ def autorelation(rel_type, rel_catalog = nil)
...
 
     self.class.send(auto_rel) { |type, block|
-      # Ignore any types we can't find, although that would be a bit odd.
-      next unless Puppet::Type.type(type)
+      if not Puppet::Type.type(type)
+        # We know it is not a type at this point, let's see if it exists as a defined type in the catalog
+        # We split and join on :: because defined types will likely have colon separators
+        # All defined types are of type :component
+        found_defines = rel_catalog.resources.select do |x|
+          x.type == :component && x.title_class == type.to_s.split('::').map(&:capitalize).join('::')
```
joshcooper: @hlindberg does puppet already have code to detect when a resource is a defined type?

Also what are your thoughts about allowing auto* relationships between types and defined types?

asasfu: I still have to write a Jira feature request for this in full description but see the note on the commit for some explanation.  One spot is that mighq/ipset module uses a defined type instead of a custom type.  IPset is an extension to iptables and I use it in my firewalld module.  I prefer to be able to grab their ipset name that they're using in their direct rule and do an autorequire on it inside my firewalld custom type to ensure ordering and also have the option for enforcing a hard requires if chosen by utilizing fail.  As the commit note points out.  This was only possible using autorequire(:component) do ... end  before.  This PR would fix that so I could do autorequire(:ipset) (which is a defined type)

lib/puppet/type/component.rb
```diff
@@ -49,6 +49,11 @@ def ref
     reference.to_s
   end
 
+  # This allows us to grab the defined type 'type' for optional use in other custom type's autorelation methods
+  def title_class
```
hlindberg: The name of this method is odd, and the documentation does not help. It should explain what it returns first and then explain the typical use case (if not apparent from the description). Also (I have not looked), if there already is a way to get `reference` that is API, that should be used instead of adding more convenience methods.

asasfu: reference is a private method inside of this class I believe.  The name can easily be changed, I wasn't completely sure how to describe it in a way that would be useful.  Something like title_type or something.  Because technically it's grabbing the sudo-type created using component.rb which is what we'd need in order to make this work.  
I'm open to any suggestions and modifications.  @hunner had mentioned this kind of addition might be useful in core and I was doing the 'hack' as mentioned in PUP-5360 and we had decided it might be nice to make that easier to use for other people.

lib/puppet/type.rb
hlindberg: @joshcooper There are methods named `builtin_type?()` and `builtin?()` (an alias for the former) on resource. Those differentiates between user defined types and "native" for the purpose of not having to evaluate the native resources (they have no code body that the compiler should evaluate). Note that app management adds new kinds of resources (predicate methods: `is_capability?`, `is_application_component?`) - notably "capabilities" (a.k.a "services") are implemented as native types.

hlindberg: @asasfu a jira ticket with the description would be helpful

asasfu: PUP-5360 has been created and placed in the PR header now.

hlindberg: @joshcooper Have not thought deeply about a native instance having an autorequire on a user defined resource, but seems like there is nothing that would prevent this as long as the containment is handled, plus the possibility that the autorequire could create a circular dependency (since autorequires are not computed by the compiler this is detected at apply time).

MikaelSmith: Further discussion in PR Triage:
- Determining whether a resource is a defined_type should be a query to the resource. Even if we have to use a heuristic on the name, it seems like we should encapsulate that in the resource.
- It seems like somehow https://github.com/puppetlabs/puppet/blob/master/lib/puppet/resource/type.rb#L25 should be accessible and usable in this instance.

joshcooper:
Thank you @asasfu for your contribution. We had a few questions during the triage:

* Can the autorequire result in a circular dependency? If it does, do we skip adding the second edge like we do for builtin types in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/graph/relationship_graph.rb#L174?
* Prior to your PR, can the graph contain an edge between a builtin type and a defined type? I believe it can, and that this PR is just making it easy to do so via auto-relations? Assuming that's correct, then this PR seems relatively safe. If not correct, then do other call sites need to be updated to handle that there can be edges between builtin types and defined types, e.g. when traversing the graph?
* We recently modified how generated resources are handled https://github.com/puppetlabs/puppet/pull/4258, wondering if this PR will work correctly with generated resources?

lib/puppet/type.rb
joshcooper: @hlindberg it looks like we can get access to the `definition?` predicate method using:

```
(rdb:2) rel_catalog.resource('Ipset', 'foo').instance_variable_get(:@reference).resource_type.definition?
true
```

where ipset is the type of the defined type and foo is the title. However, we're having to use a private instance variable, and seems like there should be a better way. For example, Puppet::Type knows if it's an `application?`, but not if it's a `definition?`. Thoughts on how to handle that?

hlindberg: Is that @reference in type/component.rb ? (Not sure where to search for it). In general I don't mind if the instances used have methods that return what kind of puppet "kind" they are via a predicate method. It is very confusing the way it is in terms of which ruby instances represent what in puppet terms. IIRC, there is a predicate for 'is builtin', but since 'not is_builtin' is not the same as 'definition?' there seems to me something missing.

whopper:
Hey, @asasfu! Just checking in to see if you're still interested in working on this?

asasfu:
Hey @whopper , It's ok with me, I'm not sure how much more I can do for it as I'm not hugely versed on puppet internals.  The last note was that it was going to go through languages discussion I thought?  Something like that.  
I basically just produced the code that would match it up to regular types as far as functionality from what I could tell and it happened to work.  Performance and framework for it was a little outside of my scope of checks.

branan:
I think we're going to go ahead and close this for now. I have some concerns about iterating over the catalog in every autorequire [that's at least O(n^2)], and we already have way too many nasty quadratic cases in the transaction. A solution to this might be to pre-populate the list of available defined types somewhere from `transaction.rb` and pass that in to the autorequires. If you're interested in doing that work feel free to re-open.

We'll leave the ticket open to track the feature request.

PR #4342, opened 2015-10-14
(PUP-4740) Update HTTP APIs for source_permissions and configured_environment

MikaelSmith:
##### (PUP-4740) Update http file_metadata api with source_permissions

Previously the source_permissions parameter was added to file_metadata
without updating the API documentation. Update the API documentation
with examples of its use.

##### (PUP-4740) Add configured_environment to node and catalog APIs

The client includes which environment was configured on the client (if
not using the application default) as the configured_environment in node
and catalog requests. This may be used by a node classifier when
determining the environment the client should use, as added in PUP-4521.

Also adds that the transaction_uuid can be included in the node request.

### DIFF
--- a/api/docs/http_catalog.md
+++ b/api/docs/http_catalog.md
@@ -8,7 +8,7 @@ Find
 
 Retrieve a catalog.
 
-    POST /puppet/v3/catalog/:nodename?environment=:environment
+    POST /puppet/v3/catalog/:nodename
     GET /puppet/v3/catalog/:nodename?environment=:environment
 
 ### Supported HTTP Methods
@@ -31,21 +31,26 @@ The examples below use the POST method.
 
 ### Parameters
 
-Three parameters should be provided to the POST or GET:
+Four parameters should be provided to the POST or GET:
 
+- `environment`: the environment name
 - `facts_format`: must be `pson`
 - `facts`: serialized pson of the facts hash.  One odd note: due to a long-ago misunderstanding in the code, this is
 doubly-escaped (it should just be singly-escaped).  To keep backward compatibility, the extraneous
 escaping is still used/supported.
 - `transaction_uuid`: a transaction uuid identifying the entire transaction (shows up in the report as well)
 
+An optional parameter can be provided to the POST or GET to notify a node classifier that the client requested a specific
+environment, which might differ from what the client believes is its current environment.
+- `configured_environment`: the environment configured on the client
+
 ### Example Response
 
 #### Catalog found
 
-    POST /puppet/v3/catalog/elmo.mydomain.com?environment=env
+    POST /puppet/v3/catalog/elmo.mydomain.com
 
-    facts_format=pson&facts=%7B%22name%22%3A%22elmo.mydomain.com%22%2C%22values%22%3A%7B%22architecture%22%3A%22x86_64%22%7D&transaction_uuid=aff261a2-1a34-4647-8c20-ff662ec11c4c
+    environment=env&configured_environment=canary_env&facts_format=pson&facts=%7B%22name%22%3A%22elmo.mydomain.com%22%2C%22values%22%3A%7B%22architecture%22%3A%22x86_64%22%7D&transaction_uuid=aff261a2-1a34-4647-8c20-ff662ec11c4c
 
     HTTP 200 OK
     Content-Type: text/pson
--- a/api/docs/http_catalog.md
+++ b/api/docs/http_catalog.md
@@ -41,7 +41,8 @@ escaping is still used/supported.
 - `transaction_uuid`: a transaction uuid identifying the entire transaction (shows up in the report as well)
 
 An optional parameter can be provided to the POST or GET to notify a node classifier that the client requested a specific
-environment, which might differ from what the client believes is its current environment.
+environment, which might differ from what the client believes is its current environment:
+
 - `configured_environment`: the environment configured on the client
 
 ### Example Response
--- a/api/docs/http_api_index.md
+++ b/api/docs/http_api_index.md
@@ -7,6 +7,9 @@ and they've historically relied on a lot of shared code to work correctly.
 This is gradually changing, although we expect external use of these APIs to
 remain low for the foreseeable future.
 
+Puppet will often send garbage URL parameters, such as `fail_on_404` and
+`ignore_cache`. The server will ignore any parameters it isn't expecting.
+
 V1/V2 HTTP APIs (Removed)
 ---------------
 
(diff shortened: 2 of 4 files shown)


api/docs/http_catalog.md
```diff
@@ -39,11 +39,15 @@ doubly-escaped (it should just be singly-escaped).  To keep backward compatibili
...
+An optional parameter can be provided to the POST or GET to notify a node classifier that the client requested a specific
+environment, which might differ from what the client believes is its current environment.
+- `configured_environment`: the environment configured on the client
+
 ### Example Response
 
 #### Catalog found
 
-    POST /puppet/v3/catalog/elmo.mydomain.com?environment=env
+    POST /puppet/v3/catalog/elmo.mydomain.com?environment=env&configured_environment=env
```
joshcooper: The request line isn't quite right. Since it's a POST, the parameters appear in the body of the request, not on the request line:

```
POST /puppet/v3/catalog/arcturus.corp.puppetlabs.net HTTP/1.1
Accept: pson, msgpack, yaml, dot, binary
X-Puppet-Version: 4.2.3
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
User-Agent: Ruby
Host: arcturus:8140
Content-Length: 8743
Content-Type: application/x-www-form-urlencoded

environment=production&facts_format=pson&facts=...&transaction_uuid=c09942fd-2c91-4f5b-8602-8f7291911045&fail_on_404=true"
```

I also noticed the agent sends `fail_on_404=true`. I don't know that the server ever uses this parameter (or if it just happens to be a Puppet::Indirector::Request parameter so that it can raise if the server returns 404).

api/docs/http_catalog.md
```diff
@@ -39,11 +39,15 @@ doubly-escaped (it should just be singly-escaped).  To keep backward compatibili
...
 
     facts_format=pson&facts=%7B%22name%22%3A%22elmo.mydomain.com%22%2C%22values%22%3A%7B%22architecture%22%3A%22x86_64%22%7D&transaction_uuid=aff261a2-1a34-4647-8c20-ff662ec11c4c
```
MikaelSmith: Is this the line showing the body?

joshcooper: Ah yep. Just need to move the query parameters from the request line:

`?environment=env&configured_environment=env`

to the body.

api/docs/http_catalog.md
MikaelSmith: How should we update line 11?

api/docs/http_catalog.md
```diff
@@ -39,13 +39,17 @@ doubly-escaped (it should just be singly-escaped).  To keep backward compatibili
...
+
 ### Example Response
 
 #### Catalog found
 
-    POST /puppet/v3/catalog/elmo.mydomain.com?environment=env
+    POST /puppet/v3/catalog/elmo.mydomain.com
 
-    facts_format=pson&facts=%7B%22name%22%3A%22elmo.mydomain.com%22%2C%22values%22%3A%7B%22architecture%22%3A%22x86_64%22%7D&transaction_uuid=aff261a2-1a34-4647-8c20-ff662ec11c4c
+    environment=env&configured_environment=env&facts_format=pson&facts=%7B%22name%22%3A%22elmo.mydomain.com%22%2C%22values%22%3A%7B%22architecture%22%3A%22x86_64%22%7D&transaction_uuid=aff261a2-1a34-4647-8c20-ff662ec11c4c
```
joshcooper: Given the last change to configured_environment, will the agent ever send the same `env` for both parameters? Or is `configured_environment` only sent when it's different than `environment`?

MikaelSmith: They can be the same. configured_environment is only nil when Puppet uses its application default (i.e. the environment wasn't configured on the client).

MikaelSmith: Since configured_environment is optional, omitting it would certainly be valid.

joshcooper:
:+1: 

@nfagerlund since you filed the ticket, could you give this a review and merge?

nfagerlund:
This is still a bit incomplete. What I'm looking for is:

- The node endpoint needs a "parameters" section that lists the parameters, like what we see in the catalog page. For each parameter, it needs a description of what it's for, whether it's optional or not, and what its default value is if omitted. 
- The file metadata endpoint's `find` action says "parameters: none." It needs info about the parameters as well!
- The file metadata search action also needs an update to its parameter list.

MikaelSmith:
Made some updates. I've left out things that we accidentally send on the request but ignore in the terminus (no different than sending anything else in the request). Also not sure about duplicating some of the documentation for source_permissions from https://docs.puppetlabs.com/references/latest/type.html#file-attribute-source_permissions.

MikaelSmith:
The docs aren't consistent about parameter list formatting, they use both
* `name` -- details
and
- `name`: details

joshcooper:
:+1: to overall changes. I like `name`: details

api/docs/http_catalog.md
```diff
@@ -8,7 +8,7 @@ Find
 
 Retrieve a catalog.
 
-    POST /puppet/v3/catalog/:nodename?environment=:environment
+    POST /puppet/v3/catalog/:nodename
```
nfagerlund: Hey... isn't this illegal?

nfagerlund: Wait, sorry, I see you're treating environment as part of the POST data instead. If that's the case, though, I'd like to see environment added to the parameters section, to make sure it's not overlooked in a POST request. (Catalog is the only traditional endpoint we POST to, right?)

nfagerlund:
@MikaelSmith I think you might have been asking in hipchat about some parameters that puppet agent usually sends but which have zero effect on the resulting response? I'd be cool with omitting those, but in that case I'd like you to edit `http_api_index.md` — we should mention that puppet agent will often send garbage URL parameters, mention the ones you mentioned, and explicitly state that we expect the server to _ignore_ any parameters it wasn't expecting.

nfagerlund:
With the exception of those last two comments, and unless anyone has more questions, I'm :+1: to merge!

api/docs/http_catalog.md
MikaelSmith: Yeah, changed it because you can't provide query strings on a POST. I'll update the Parameters section.

MikaelSmith:
Updated.

joshcooper:
LGTM, merging.

@nfagerlund if you have any concerns, feel free to comment on the PR and we'll fix it up.

nfagerlund:
Sweet. Thank you both for your time on this!

PR #4343, opened 2015-10-14
(PUP-5382) Allow undefined nodes key on app declarations

jolshevski:
Per the app spec, providing node mappings on app declarations is
optional. Prior to this commit, an error would be raised if the nodes
key on an app declaration was undefined. This commit prevents `.value`
from being called on an undefined value during application evaluation.

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -330,6 +330,7 @@ class Puppet::Parser::Compiler
     @applications.each do |app|
       components = []
       mapping = app.parameters[:nodes] ? app.parameters[:nodes].value : {}
+      raise Puppet::Error, "Invalid node mapping in #{app.ref}: Mapping must be a hash" unless mapping.is_a?(Hash)
       all_mapped = Set.new
       mapping.each do |k,v|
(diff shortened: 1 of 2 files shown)


lib/puppet/parser/compiler.rb
```diff
@@ -328,8 +328,8 @@ def prune_node_catalog
   # @api private
   def evaluate_applications
     @applications.each do |app|
-      mapping = app.parameters[:nodes].value
       components = []
+      mapping = app.parameters[:nodes] ? app.parameters[:nodes].value : {}
```
hlindberg: While at it, think we need to protect against a user doing `nodes => true` or `nodes => 'blah'` by checking that it is indeed a hash or nil. The code above would be fooled by a truthy value and trying to lookup :nodes. (Not sure if an application guards against that).

hlindberg:
Thanks for the PR, looks good - I wonder if we need to be concerned about a non hash value or not.

jolshevski:
Henrik, apt observation. Does the approach in 1b2076a work for you?

hlindberg:
Yes, that adds sufficient protection. (There may be checks elsewhere, but since other aspects of what may be found inside a :node hash was being checked I think this added check is worth having).

PR #4346, opened 2015-10-15
(PUP-5310) Shorten username for AIX acceptance

johnduarte:
This commit shortens the username used in the
`ticket_6857_password-disclosure-when-changing-a-users-password`
acceptance test to adhere to the default 8 character limit in AIX.


branan:
This test doesn't have a cleanup step, and "tstuser" is kind of generic - I worry about accidental collisions in the future.

Could you either add a cleanup step or make it a bit more unique? Maybe "pwtstusr"?

johnduarte:
ping @branan I have changed the username to our more common pattern of random numerical suffix.

branan:
@johnduarte do we want to add a cleanup step while we're playing with this test, or just punt it for now?

johnduarte:
@branan yeah, cleanup should happen. I have added a teardown step. Thanks.

PR #4349, opened 2015-10-15
(PUP-5380) Cache future_parser flag in environment

thallgren:
Prior to this commit, the Puppet.future_parser? method would load
the environment configuration from disk each time the method was
called. This had a very negative performance impact since it happens
in a lot of places.

This commit ensures that the future_parser flag is cached in the
environment instance.

All caches are cleared when the global configuration is changed.
Individual environment caches are cleared when the environment times
out.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -82,6 +82,8 @@ module Puppet
   # setting access and stuff
   def self.[]=(param,value)
     @@settings[param] = value
+    # Ensure that all environment caches are cleared if we're changing the parser
+    lookup(:environments).clear_all if param == :parser
   end
 
   def self.clear
@@ -242,6 +244,8 @@ module Puppet
   # @api private
   def self.override(bindings, description = "", &block)
     @context.override(bindings, description, &block)
+  ensure
+    lookup(:root_environment).instance_variable_set(:@future_parser, nil)
   end
 
   # @api private
@@ -264,14 +268,7 @@ module Puppet
   #
   def self.future_parser?(in_environment = nil)
     env = in_environment || Puppet.lookup(:current_environment) { return Puppet[:parser] == 'future' }
-    env_conf = Puppet.lookup(:environments).get_conf(env.name)
-
-    if env_conf.nil?
-      # Case for non-directory environments
-      Puppet[:parser] == 'future'
-    else
-      env_conf.parser == 'future'
-    end
+    env.future_parser?
   end
 end
 
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -2052,6 +2052,10 @@ EOT
     },
     :parser => {
       :default => "current",
+      :hook    => proc do |value|
+        envs = Puppet.lookup(:environments) { nil }
+        envs.clear_all unless envs.nil?
+      end,
       :desc => <<-'EOT'
         Selects the parser to use for parsing puppet manifests (in puppet DSL
         language/'.pp' files). Available choices are `current` (the default)
--- a/lib/puppet/environments.rb
+++ b/lib/puppet/environments.rb
@@ -38,6 +38,11 @@ module Puppet::Environments
         raise EnvironmentNotFound, name
       end
     end
+
+    def clear_all
+      root = Puppet.lookup(:root_environment) { nil }
+      root.instance_variable_set(:@future_parser, nil) unless root.nil?
+    end
   end
 
   # @!macro [new] loader_search_paths
@@ -311,6 +316,9 @@ module Puppet::Environments
       nil
     end
 
+    def clear_all
+      @loaders.each {|loader| loader.clear_all}
+    end
   end
 
   class Cached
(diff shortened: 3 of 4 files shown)


hlindberg:
I think this is not the right way to fix this. The Puppet.future_parser? method is the way to determine what it actually is set to at the point of call. Instead, excessive calling to this method should be avoided. I also suspect that the changes in this PR can cause problems for non directory environments. (Need to read again).

hlindberg:
I believe we also removed a bunch of calls to Puppet.future_parser? in versions after 3.7.5 (for performance reasons - i.e. by caching the result of calling Puppet.future_parser? When known to be safe/ok to do so in an object with a known lifecycle).

hlindberg:
I also believe we earlier had the problem that you could not change the parser=future flag in environment.conf - it would not react to that change unless you take down the server (assuming your are running with infinite cache)

thallgren:
Is this flag expected to change after the environment has been loaded? If not, then what's wrong with this fix?

thallgren:
The variable is cached in the environment instance. If that instance is recreated, then the cache is refreshed.

hlindberg:
I am really nervous about making changes in this area. We had so many surprises around environments (esp. non directory variant) and the order in which things occur. I would much rather find the place from which it is excessively called.

thallgren:
The calls to this method are all over the place. A quick count gives me about 30 locations in the code.

Not sure what there is to be nervous about. We're already using the environment as the scope for environment data provider (via adapter). I find it hard to believe that switching compiler type back and forth would be more frequent than altering environment data.

thallgren:
I see other fairly serious problems if the parser changes during the lifetime of the environment. Such a change will not cause cache evictions which means that the re-parse using the new parser might never happen. If it happens, it's not in any way synchronized with the parser change. In that respect, this fix is actually an improvement since it keeps the cached parsed material in sync with the reported parser type.

hlindberg:
Because there are non directory based environments (data in modules does not work with 3x and non directory based environments). You are in the old code now. And this PR changes the order of things - you must have an environment instance. Yes there are calls "all over the place", but it must be one of the places that is hogging the performance for this user. It is much safer to cache this in the compiler, and let all calls to the global future_parser? method done from within compiler, catalog etc. go to the compiler (it is always for one compilation).

hlindberg:
I believe that there is one (or a small number) of the calls that are actually problematic and that they are especially problematic for one set of users (IIRC there is some logic where compilation/catalog/resource operations are different depending on the flag, and that logic does not know about state of things; maybe some users just hit that logic way more frequently. It is far from everyone that has performance problems.

cprice404:
I'm pretty sure the OSS user who helped us track this down would be amenable to applying a patch that had some debugging statements in it if you want to see what type of resource is causing the issue.  That information would probably be useful for PE / support / @npwalker as well.

thallgren:
It would be very easy to ensure that all cached statuses are cleared whenever the `Puppet[:parser]` flag changes. That would include the one in the root environment and hence adress Henrik's concern about directory less environments.

I would prefer that approach from examining stacks to try and figure out where the calls come from. I also think that this might differ between users environments. Another concern I have with that approach is that when we find such places, how do we resolve the fact that the calls are made? By caching the result? If so, aren't we just moving the problem elsewhere?

thallgren:
"you must have an environment instance" is not true b.t.w. This line:

    env = in_environment || Puppet.lookup(:current_environment) { return Puppet[:parser] == 'future' }

actually returns the value of Puppet[:parser] when no environment could be found. The return is from the method, not just from the block.

thallgren:
Pushed a new commit for this where all environment caches (including parsed code) are cleared when the global `parser` setting is changed (or removed).
In contrast to my previous attempt, this commit does not alter any tests.

npwalker:
@thallgren are you saying that with this change I could change the parser setting and it would take effect immediately?  Because I've always had to restart the puppet master in order for the future parser to take affect.

thallgren:
@npwalker,  yes. With this change you will be able to switch parser without a restart.

hlindberg:
So, what happens now if environment.conf is changed wrt. the parser setting that may be in it, or that is added or removed (without the default Puppet[:parser] changing) ?

hlindberg:
Also, environments may be completely unaffected by the Puppet[:parser] setting if they have this setting in their environment.conf (no need to evict them, but OTOH also no real harm in doing so).

thallgren:
If the environment.conf changes, then that change will be picked up the next time the environment is evicted from the cache which vouch for consistency between reported parser and the already parsed code that the cache contains. The current code might well report 'future' while containing code parsed with 'current' if the environment.conf is modified.

thallgren:
And yes, the choice to always evict the cache when the global setting changes is probably not optimal in all situations, but then again, how often will that happen? Since it's not harmful I didn't think it was worth spending time on.

hlindberg:
So, with a cache time of "infinite" set in environment.conf, it is not possible to make changes take effect without taking the server down (or fiddling with the environment_timeout setting that causes all environments to be evicted). I don't think we have logic that evicts an environment if the environment.conf is modified...

cprice404:
The Puppet Server administrative HTTP API can be used to flush the environments, even with the `infinite` setting.

hlindberg:
Tickets for a found regression in file watching, and a ticket for watching environment.conf will be logged.

PR #4350, opened 2015-10-15
(maint) replace skips with expect_failure in service_enable_linux test

er0ck:
The previous version of this test had some heavy-handed skip_tests which
would not remind us to fix the test when upstream bug is fixed. This
change uses expect_failure to remind us to fix the test. This came out
of some some related beaker work with this file as an example. free bird
in the hand, or something.

[skip ci]

### DIFF
--- a/acceptance/tests/resource/service/service_enable_linux.rb
+++ b/acceptance/tests/resource/service/service_enable_linux.rb
@@ -17,10 +17,6 @@ agents.each do |agent|
   platform = agent.platform.variant
   majrelease = on(agent, facter('operatingsystemmajrelease')).stdout.chomp.to_i
 
-  if ((platform == 'debian' && majrelease == 8) || (platform == 'ubuntu' && majrelease == 15))
-    skip_test 'legit failures on debian8 and ubuntu15; see: PUP-5149'
-  end
-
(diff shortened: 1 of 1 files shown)


acceptance/tests/resource/service/service_enable_linux.rb
```diff
@@ -92,24 +88,34 @@
...
-      end
-    else
-      rc_symlinks = on(agent, "ls #{symlink_systemd} #{init_script_systemd}", :accept_all_exit_codes => true).stdout
-      assert_match("#{symlink_systemd}",     rc_symlinks, "did not find #{symlink_systemd}")
-      assert_match("#{init_script_systemd}", rc_symlinks, "did not find #{init_script_systemd}")
+
+  result = apply_manifest_on(agent, manifest_httpd_enabled)
+  if platform == 'ubuntu' && majrelease == 15
+    expect_failure 'legit failures on ubuntu15; see: PUP-5149' do
+      assert_no_match(/Could not enable/, result.output, 'could not enable http service')
```
er0ck: oops.  forgot to put this into a proc.  :neutral_face:

joshcooper:
@er0ck this needs a rebase. Also, you may want to retarget stable for 4.4.1.

PR #4352, opened 2015-10-16
(PUP-5151) Don't allow remote file buckets to be listed

joshcooper:
Previously the filebucket list command was implemented using the `find`
method on the `file_bucket_file` endpoint passing along `list_all` as a
query parameter. However, the filebucket is accessible to all users, so
any remote user could list the contents of the filebucket and retrieve
file content backed up by other hosts. And since retrieve and list both
use `find` where the only difference is the `find_all` query parameter,
it wasn't possible to write an ACL that allowed access to the
former, but denied access to the latter.

This commit modifies the dipper client code to raise an error when
trying to list a remote filebucket. It also modifies the
Puppet::FileBucketFile::File#list server-side code to reject remote
requests, for example, if you curl the REST API directly.

Note the list method raises an AuthorizationError so that the REST
endpoint returns HTTP 403 to the caller.

### DIFF
--- a/lib/puppet/indirector/file_bucket_file/file.rb
+++ b/lib/puppet/indirector/file_bucket_file/file.rb
@@ -38,6 +38,10 @@ module Puppet::FileBucketFile
     end
 
     def list(request)
+      if request.remote?
+        raise Puppet::Error, "Listing remote file buckets is not allowed"
+      end
+
       fromdate = request.options[:fromdate] || "0:0:0 1-1-1970"
       todate = request.options[:todate] || Time.now.strftime("%F %T")
       begin
--- a/lib/puppet/application/filebucket.rb
+++ b/lib/puppet/application/filebucket.rb
@@ -47,11 +47,12 @@ restore:
   to its original location.
 
 diff:
-  Print a diff in unified format between two checksums in the filebucket or between a checksum and
-  its matching file.
+  Print a diff in unified format between two checksums in the filebucket
+  or between a checksum and its matching file.
 
 list:
-  List all files in the current filebucket.
+  List all files in the current local filebucket. Listing remote
+  filebuckets is not allowed.
 
 DESCRIPTION
 -----------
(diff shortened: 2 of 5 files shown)


lib/puppet/indirector/file_bucket_file/file.rb
```diff
@@ -38,6 +39,10 @@ def find(request)
     end
 
     def list(request)
+      if request.remote?
+        raise Puppet::Network::AuthorizationError, "Listing remote file buckets is not allowed"
+      end
+
```
joshcooper: Puppet has two different ways for doing authz for remote requests. When the HTTP layer receives a request, it will ask the indirection if it allows remote requests:

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/http/api/indirected_routes.rb#L39

And the indirection will delegate to its current terminus:

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/indirection.rb#L176-L178

The default terminus behavior is to allow remote access: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/terminus.rb#L144

But in some cases we don't, e.g. keys should never be accessed remotely: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/key/file.rb#L10-L12

However, this mechanism doesn't pass the request in, so the terminus can't decide based on request options.

The other mechanism is for the terminus to call `request.remote?` as is done in this PR. The compiler terminus does something similar for `use_node`: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/catalog/compiler.rb#L131-L133

That said the former approach seems preferable since it's guaranteed that the request came in through the HTTP layer. The latter approach is more indirect as it assumes the request is remote if it has a `node` or `ip`.

Ideally `IndirectedRoutes` would call `Indirection#allow_remote_requests?` passing in the request. However, I don't see how to do that in a backwards compatible way, since people may have written termini that expect no parameters.

One option might be to create new methods on indirection and terminus that take a request parameter, `allow_remote?(request)`, deprecate the old methods, and switch them over to call the new methods. That seems more than I want to do now.

So I think we need to go with the approach in this PR.

branan:
:+1: This is an almost-distressingly simple patch

MikaelSmith:
:+1:

PR #4353, opened 2015-10-16
(PUP-5262) Skip pup/mco service test in Solaris

whopper:
Currently, there is a race condition when refreshing the puppet
service in Solaris. In the time that the service is restarting,
puppet sees the service as being stopped, but moments later sees
it as running. Skip the pup/mco service test in Solaris until this
can be addressed.


johnduarte:
:+1:

whopper:
@joshcooper @johnduarte updated to skip Solaris all together.

PR #4354, opened 2015-10-16
(PUP-5059) Add the --node feature to lookup CLI tool 

HAIL9000:
Prior to this commit, the --node flag did not actually work as
intended, it would simply always default to the machine you were
running lookup on due to some settings not being configured properly.

Now you can specify a node with --node and the host name and lookup
will find and compile that node to obtain a scope. If you do not
specify a node then lookup will default to using the scope of the
machine you are running on (which is intended to be the master).

### DIFF
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -8,6 +8,8 @@ class Puppet::Application::Lookup < Puppet::Application
   RUN_HELP = "Run 'puppet lookup --help' for more details".freeze
   DEEP_MERGE_OPTIONS = '--knock-out-prefix, --sort-merged-arrays, --unpack-arrays, and --merge-hash-arrays'.freeze
 
+  run_mode :master
+
   # Options for lookup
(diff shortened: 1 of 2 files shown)


lib/puppet/application/lookup.rb
```diff
@@ -72,6 +87,57 @@ def setup
...
+  def setup
+    setup_logs
+
+    exit(Puppet.settings.print_configs ? 0 : 1) if Puppet.settings.print_configs?
+
+    Puppet.settings.use :main, :master, :ssl, :metrics
+
+    setup_terminuses
+
+    # TODO: Do we need this in preview? It sets up a write only cache
```
hlindberg: this mentions "preview"

lib/puppet/application/lookup.rb
```diff
@@ -72,6 +87,57 @@ def setup
...
+    setup_node_cache
+
+    setup_ssl
+    puts "setup happened! o:"
```
hlindberg: looks like debugging output

PR #4355, opened 2015-10-16
(PUP-4954) Don't initialize environment object in indirector requests unless it's used

branan:
Previously, any indirector request to a URI that had the appropriate
environment/indirector/key pathing would cause the request to
initialize an environment object. With directory environments, this
does not work - the agent probably doesn't have a dir for the
environment of the request, so initializing the environment object
fails.

This change makes initialization of the request-associated environment
object lazy, such that it only happens if the environment is actually
requested. It also makes a small change to the http indirected_routes
logic so that it does not force creation of the environment object
simply to query its name.

### DIFF
--- a/lib/puppet/indirector/request.rb
+++ b/lib/puppet/indirector/request.rb
@@ -29,16 +29,25 @@ class Puppet::Indirector::Request
   def environment
     # If environment has not been set directly, we should use the application's
     # current environment
-    @environment ||= Puppet.lookup(:current_environment)
+    @environment ||= if @environment_name
+                         Puppet.lookup(:environments).get!(@environment_name)
+                     else
+                         Puppet.lookup(:current_environment)
+                     end
   end
 
   def environment=(env)
-    @environment =
     if env.is_a?(Puppet::Node::Environment)
-      env
+      @environment = env
+      @environment_name = nil
     else
-      Puppet.lookup(:environments).get!(env)
+      @environment = nil
+      @environment_name = env
     end
+  end
+
+  def environment_name
+    @environment_name || self.environment.name
   end
 
   def escaped_key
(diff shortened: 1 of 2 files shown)


Sharpie:
:metal:

MikaelSmith:
Kicking off an acceptance test with https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Manual/view/Puppet%20Agent%20Ad-Hoc/job/platform_aio-suite_stage-intn_adhoc/45/

MikaelSmith:
Manually ran from the package and all acceptance passed. This looks ok to me. :+1:

joshcooper:
Based on comments in https://tickets.puppetlabs.com/browse/PUP-4954?focusedCommentId=226298&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-226298, I'm not understanding why this is only an issue for filebuckets? For example, the agent pluginsyncs in an environment, and it likely doesn't have a directory named for that environment.

branan:
@joshcooper We use puppet File resources for plugins. Those bypass the indirector in various terrible ways in order to stream directly to disk, so we don't end up hitting the path where we create an indirector request object.

At least, that's what I think is happening. I'm still trying to figure out which way is up in the pluginsync code path

escher jareth

joshcooper:
Pretty sure we do use the indirector, e.g. https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/content.rb#L215

For example, in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/source.rb#L180, we're passing the environment to use when executing file_metadata request.

branan:
PUP-4653 may actually completely change how/if this needs to be fixed. This commit was written against a checkout of stable and opened against master. Master contains a59239d5f9e954cb605bfee48d0280bbe6f6c1aa from @ffrank, which changes the way pluginsync interacts with environments in request generation. I'm gonna revist this with master and see what things look like.

For now, I'm going to close this PR and kick the ticket back to doing. Part of why it looks weird on master is that it *is* weird on master - it was written to solve a particularly problematic codepath that no longer eixsts.

joshcooper:
I worked with @ffrank on that PR, happy to talk later today about it.

PR #4356, opened 2015-10-16
(maint) Stub Facter operatingsystemmajrelease value in DNF tests

whopper:
The install tests will fail when running on an OS with majrelease
< 5, so stub that value.


branan:
This fixes the spec tests on my laptop. I'll let them run in appveyor as a sanity check, then :+1:

whopper:
@branan yay arch linux!

PR #4357, opened 2015-10-17
(PUP-5311) Update teardown for aix pkg acceptance

johnduarte:
This commit updates the teardown step in the `aix_package_provider`
test to ensure that the packages involved in the test are removed
when the test is completed.

### DIFF
--- a/acceptance/tests/aix/aix_package_provider.rb
+++ b/acceptance/tests/aix/aix_package_provider.rb
@@ -4,10 +4,6 @@ confine :to, :platform => /aix/
 
 dir = "/tmp/aix-packages-#{$$}"
 
-teardown do
-  on hosts, "rm -rf #{dir}"
-end
-
 def assert_package_version(package, expected_version)
   # The output of lslpp is a colon-delimited list like:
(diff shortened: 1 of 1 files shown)


acceptance/tests/aix/aix_package_provider.rb
```diff
@@ -22,6 +18,11 @@ def assert_package_version(package, expected_version)
 version1 = '1.7.10.4'
 version2 = '1.8.6.4'
 
+teardown do
+  on hosts, "rm -rf #{dir}"
+  on hosts, puppet('resource', 'package', "'#{package}' ensure=absent")
```
joshcooper: I'm surprised that we're quoting the package, e.g. `puppet resource package 'mypackage' ensure=absent`? Is that actually removing the package? Or is it trying to remove a package whose name contains quotes, and since there's no package with that name, doing nothing?

branan: This gets shoved through the shell on the SUT, so the quotes will get eaten at that level. Redundant here since the package name doesn't contain any special characters, but probably not the worst practice.

branan: And all of the arguments get mashed into one string via https://github.com/puppetlabs/beaker/blob/master/lib/beaker/command.rb#L120

PR #4358, opened 2015-10-17
(PUP-5395) auto merge lookup

rnelson0:
HI-118: Allow automatic parameter lookups to use the deep merge strategy with hiera_advanced_parameter_bindings set to true and lookupoptions::* keys present.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -362,6 +362,11 @@ module Puppet
       :desc    => "The hiera configuration file. Puppet only reads this file on startup, so you must restart the puppet master every time you edit it.",
       :type    => :file,
     },
+    :hiera_advanced_parameter_bindings => {
+      :type    => :boolean,
+      :default => false,
+      :desc    => "Whether to provide advanced bindings to hiera automatic parameter lookups.",
+    },
     :binder_config => {
       :default => nil,
       :desc    => "The binder configuration file. Puppet reads this file on each request to configure the bindings system.
--- a/lib/puppet/indirector/hiera.rb
+++ b/lib/puppet/indirector/hiera.rb
@@ -18,7 +18,7 @@ class Puppet::Indirector::Hiera < Puppet::Indirector::Terminus
   def find(request)
     not_found = Object.new
     options = request.options
-    Puppet.debug "doing a hiera indirector lookup of #{request.key} with options #{options.inspect}"
+    Puppet.debug "Performing a hiera indirector lookup of #{request.key} with options #{options.inspect}"
     value = hiera.lookup(request.key, not_found, Hiera::Scope.new(options[:variables]), nil, convert_merge(options[:merge]))
     throw :no_such_key if value.equal?(not_found)
     value
--- a/lib/puppet/indirector/hiera.rb
+++ b/lib/puppet/indirector/hiera.rb
@@ -18,6 +18,7 @@ class Puppet::Indirector::Hiera < Puppet::Indirector::Terminus
   def find(request)
     not_found = Object.new
     options = request.options
+    Puppet.debug "doing a hiera indirector lookup of #{request.key} with options #{options.inspect}"
     value = hiera.lookup(request.key, not_found, Hiera::Scope.new(options[:variables]), nil, convert_merge(options[:merge]))
     throw :no_such_key if value.equal?(not_found)
     value
--- a/lib/puppet/resource/type.rb
+++ b/lib/puppet/resource/type.rb
@@ -566,16 +566,15 @@ class Puppet::Resource::Type
 
     qname = "#{name}::#{param}"
 
-    # TODO: create this puppet option
-    #if Puppet[:hiera_advanced_parameter_bindings]
-    opt_name = "lookupoptions::#{qname}"
-    in_global = lambda { lookup_with_databinding(opt_name, scope) }
-    in_env = lambda { lookup_in_environment(opt_name, scope) }
-    in_module = lambda { lookup_in_module(opt_name, scope) }
-    merge_strategy = lookup_search(in_global, in_env, in_module)
-    merge_strategy = merge_strategy['merge'] if merge_strategy
-    Puppet.debug "merge strategy for #{qname} is '#{merge_strategy}'"
-    #end
+    if Puppet[:hiera_advanced_parameter_bindings]
+      opt_name = "lookupoptions::#{qname}"
+      in_global = lambda { lookup_with_databinding(opt_name, scope) }
+      in_env = lambda { lookup_in_environment(opt_name, scope) }
+      in_module = lambda { lookup_in_module(opt_name, scope) }
+      merge_strategy = lookup_search(in_global, in_env, in_module)
+      merge_strategy = merge_strategy['merge'] if merge_strategy
+      Puppet.debug "The merge strategy for #{qname} is '#{merge_strategy}'" if merge_strategy
+    end
 
     in_global = lambda { lookup_with_databinding(qname, scope, merge_strategy) }
     in_env = lambda { lookup_in_environment(qname, scope, merge_strategy) }
--- a/lib/puppet/resource/type.rb
+++ b/lib/puppet/resource/type.rb
@@ -563,10 +563,23 @@ class Puppet::Resource::Type
   #
   def lookup_external_default_for(param, scope)
     return unless type == :hostclass
+
     qname = "#{name}::#{param}"
-    in_global = lambda { lookup_with_databinding(qname, scope) }
-    in_env = lambda { lookup_in_environment(qname, scope) }
-    in_module = lambda { lookup_in_module(qname, scope) }
+
+    # TODO: create this puppet option
+    #if Puppet[:hiera_advanced_parameter_bindings]
+    opt_name = "lookupoptions::#{qname}"
+    in_global = lambda { lookup_with_databinding(opt_name, scope) }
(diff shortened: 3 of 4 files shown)


rnelson0:
I was unable to find relevant spec tests for the lookup functions modified, nor do I understand where I might add them properly. However, I have described my testing environment in [this gist](https://tickets.puppetlabs.com/browse/HI-118).

rnelson0:
Also the commit log tests fail because the patch is against puppet but the ticket is for hiera. Should I create a PUP- ticket as well?

joshcooper:
Thank you for your contribution @rnelson0! Could you move the HI-118 ticket to the PUP project and update the commit summary to reference the PUP ticket? If you run into trouble moving the ticket, let me know I can do that.

There are unit tests under `spec/unit/resource/type_spec.rb` though I am not familiar with them. Perhaps @hlindberg or @thallgren can provide advice about how to test?

rnelson0:
@joshcooper Moved to PUP-5395, commit logs updated to match.

There was discussion in the JIRA ticket and it looks like there is no effective testing in place already due to the somewhat circular dependency here.

joshcooper:
@rnelson0 thanks!

About tests, ideally you could verify that calling `lookup_with_databinding` with various arguments results in a corresponding call to `Puppet::DataBinding.indirection.find`. However the `lookup_with_XXX` methods are private, and getting them to be called with the appropriate set of arguments could be challenging due. I defer to @hlindberg on that.

hlindberg:
@rnelson0 The test `spec/integration/data_binding.rb` seems like an appropriate place to add testing of the new feature. In that set of tests there are helpers to configure hiera.yaml, and the data. Simply use those to test the "matrix" of setting on/off, with and without merge strategy, and the separately test that all merge strategies work can be used (only need to test that parameters are use, i.e. no need testing all combinations only that parameters are picked up and have effect). The main chore is probably in configuring the data for the various cases.

Does that sound like a workable strategy for testing this?

rnelson0:
@hlindberg I can start working on that. However, I'm seeing failures in those tests currently. Is this my environment or are these tests failing for others?

````
$ bundle exec rspec spec/integration/data_binding.rb
FF

Failures:

  1) Data binding looks up data from hiera
     Failure/Error: expect(resource[:calling_module]).to eq("testing")

       expected: "testing"
            got: ""

       (compared using ==)
     # ./spec/integration/data_binding.rb:39:in `block (2 levels) in <top (required)>'

  2) Data binding works with the puppet backend configured, although it can't use it for lookup
     Failure/Error: catalog = compile_to_catalog("include testing::binding")
     Puppet::PreformattedError:
       Evaluation Error: Error while evaluating a Function Call, Could not find data item variable in any Hiera data file and no default supplied at /tmp/rspecrun20151020-16421-171c7de/puppetdir20151020-16421-1pjnp0b/testing/manifests/binding.pp:2:37 on node foonode
````

**Comment by rnelson0** [comment]
I think the tests here have simply been neglected. My fixes at https://github.com/puppetlabs/puppet/commit/3e9d898e3b010ebe8f9e583d659b3761122edefc

I get the first test to work by changing the class param to `$calling_module = $module_name` and removing the hiera key for that param (I'm guessing $calling_module isn't processed properly here or through the mock). I see three tickets (open: https://tickets.puppetlabs.com/browse/HI-474, https://tickets.puppetlabs.com/browse/PUP-2236, and closed: https://tickets.puppetlabs.com/browse/PUP-83) that indicate some ongoing problems with $calling_module, so perhaps this isn't something to resolve right now and $module_name is sufficient.

The second test simply didn't match up at all, the class created in the manifest was testing::binding but the tests specified testing::data, etc. I think I got that one down just fine.

I'll add a new test before the triage tomorrow.

**Comment by hlindberg** [comment]
@rnelson0 Strange that you see errors for those tests - they do run as part of automatic testing. As you found there may be issues with them, but they should not be failing "out of the box".

**Comment by joshcooper** [comment]
There is a spec order dependency, so running `rspec spec/integration/data_binding.rb` fails in isolation. I will file a ticket on that.

**Comment by MikaelSmith** [comment]
I'm :+1: on this. Confirmed new test is validating the new behavior, and the changes look pretty clean.

File: spec/integration/data_binding.rb [hunk 1]
```diff
@@ -1,5 +1,6 @@
 require 'spec_helper'
 
+require 'puppet/node/facts'
```
MikaelSmith: No longer needed?

rnelson0:
Cleanup of rspec tests as requested are completed. Thanks!

lib/puppet/defaults.rb
```diff
@@ -362,6 +362,11 @@ def self.default_diffargs
       :desc    => "The hiera configuration file. Puppet only reads this file on startup, so you must restart the puppet master every time you edit it.",
       :type    => :file,
     },
+    :hiera_advanced_parameter_bindings => {
```
hlindberg: Suggest naming this `:data_binding_lookup_options`, since the mechanism is not a hiera specific. It works with other data binding types as well.

lib/puppet/indirector/hiera.rb
```diff
@@ -18,6 +18,7 @@ def initialize(*args)
   def find(request)
     not_found = Object.new
     options = request.options
+    Puppet.debug "Performing a hiera indirector lookup of #{request.key} with options #{options.inspect}"
```
hlindberg: Always write calls to Puppet.debug using a lambda - i.e.:
````
Puppet.debug {"Performing a hiera indirector lookup of #{request.key} with options #{options.inspect}"}
````
This is 60% faster when not running in debug mode even for a simple string, and even faster when (as in this case) there is string interpolation required to construct the debug message).

This is important here, since looking up class parameters is performance critical.

File: lib/puppet/indirector/hiera.rb [hunk 4]
```diff
@@ -18,6 +18,7 @@ def initialize(*args)
...
     value = hiera.lookup(request.key, not_found, Hiera::Scope.new(options[:variables]), nil, convert_merge(options[:merge]))
```
hlindberg: This call should not go to hiera! It should go to the Puppet::Pops::Lookup (implementation of the lookup class). That way it works with Data in Modules as well.

lib/puppet/resource/type.rb
```diff
@@ -563,10 +563,22 @@ def create_params_struct
   #
   def lookup_external_default_for(param, scope)
     return unless type == :hostclass
+
     qname = "#{name}::#{param}"
-    in_global = lambda { lookup_with_databinding(qname, scope) }
-    in_env = lambda { lookup_in_environment(qname, scope) }
-    in_module = lambda { lookup_in_module(qname, scope) }
+
+    if Puppet[:hiera_advanced_parameter_bindings]
```
hlindberg: needs to change as consequence of renaming the option

lib/puppet/resource/type.rb
```diff
@@ -563,10 +563,22 @@ def create_params_struct
...
+      opt_name = "lookupoptions::#{qname}"
+      in_global = lambda { lookup_with_databinding(opt_name, scope) }
+      in_env = lambda { lookup_in_environment(opt_name, scope) }
+      in_module = lambda { lookup_in_module(opt_name, scope) }
+      merge_strategy = lookup_search(in_global, in_env, in_module)
+      merge_strategy = merge_strategy['merge'] if merge_strategy
+      Puppet.debug "The merge strategy for #{qname} is '#{merge_strategy}'" if merge_strategy
```
hlindberg: Use a lambda to construct the debug message

lib/puppet/resource/type.rb
```diff
@@ -563,10 +563,22 @@ def create_params_struct
...
-    in_global = lambda { lookup_with_databinding(qname, scope) }
-    in_env = lambda { lookup_in_environment(qname, scope) }
-    in_module = lambda { lookup_in_module(qname, scope) }
+
+    if Puppet[:hiera_advanced_parameter_bindings]
+      opt_name = "lookupoptions::#{qname}"
+      in_global = lambda { lookup_with_databinding(opt_name, scope) }
+      in_env = lambda { lookup_in_environment(opt_name, scope) }
+      in_module = lambda { lookup_in_module(opt_name, scope) }
+      merge_strategy = lookup_search(in_global, in_env, in_module)
```
hlindberg: The above can be performed with a call to `Puppet::Pops::Lookup.lookup` where it is also possible to to a type assertion of the result. (See the puppet/functions/lookup.rb for type signatures for the various options).

hlindberg:
Most of the work done in resource/type.rb should be possible to delegate to `Puppet::Pops::Lookup.lookup` since it already does all of the searching and merging. That would clean up type.rb and reduce the duplication of logic.

hlindberg:
Hey @rnelson0, looks like we have bandwidth in the language team to take this on and finish it in time for puppet 4.3.0 - do you mind if we take this over and drive it to completion?

rnelson0:
@hlindberg Sounds perfect to me. Thanks!

rnelson0:
@hlindberg I also opened [docs PR554](https://github.com/puppetlabs/puppet-docs/pull/554) which will need to be kept in sync. I kind of like writing docs, so ping me if you want me to update that when the solution is set in stone.

hlindberg:
@rnelson0 cool, thanks, that helps. FYI @thallgren will continue the work on this PR.

hlindberg:
Closed in favor of #4378 (which includes the work in this PR).

PR #4359, opened 2015-10-18
(PUP-5391) Update use of execute method

kylog:


johnduarte:
:100: I have confirmed that this works on AIX 5.3

PR #4360, opened 2015-10-20
(PUP-5275) Evalutate capability mappings after evaluating site

thallgren:
This commit ensures that capability mappings are evaluated twice. Once
before evaluate_main since it eventually calls on evaluate_classes
which may require capability mappings, and then again after
evaluate_site since it may introduce new mappings.

I have tested this manually in my own local setup (based on input from
Nick Lewis in the JIRA) but there should be acceptance tests for this
too.


hlindberg:
appveyor seems is hung - merging anyway.

PR #4361, opened 2015-10-20
(PUP-5060) Allow users to override facts

HAIL9000:
Add support for the --facts flag in puppet lookup. This allows the
user to give a .json or .yaml file full of fact names and values
via the --facts flag. The facts the user specified will be merged
with the existing facts for the current node. This means that if
a fact is specified in the file it's value will be overridden, but
if it is not specified, the original fact value for the node will
stay the same.

### DIFF
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -223,7 +223,8 @@ the puppet lookup function linked to above.
 
 * --facts <FILE>
   Specify a .json, or .yaml file holding key => value mappings that will
-  populate a scope with data that is used when looking up.
+  override the facts for the current node. Any facts not specified by the
+  user will maintain their original value.
 
 * --render-as s|json|yaml|binary|msgpack
   Determines how the results will be rendered to the standard output where
@@ -333,10 +334,17 @@ Copyright (c) 2015 Puppet Labs, LLC Licensed under the Apache 2.0 License
 
     if options[:fact_file]
       original_facts = node.facts.values
-      if options[:fact_file].match(/.*\.json/)
+      if options[:fact_file].end_with?("json")
         given_facts = JSON.parse(File.read(options[:fact_file]))
       else
         given_facts = YAML.load(File.read(options[:fact_file]))
+
+        # If the data is not correctly formatted this will be a string so we should warn the user
+        # Otherwise both JSON and YAML will provide sane parse errors for the user if the data is
+        # incorrectly formatted
+        if !given_facts.instance_of?(Hash)
+          raise "Incorrect formatted YAML data in #{options[:fact_file]} given via the --facts flag"
+        end
       end
 
       node.facts.values = original_facts.merge(given_facts)
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -332,19 +332,18 @@ Copyright (c) 2015 Puppet Labs, LLC Licensed under the Apache 2.0 License
 
     node = Puppet::Node.indirection.find(node) unless node.is_a?(Puppet::Node) # to allow unit tests to pass a node instance
 
-    if options[:fact_file]
-      original_facts = node.facts.values
-      if options[:fact_file].end_with?("json")
-        given_facts = JSON.parse(File.read(options[:fact_file]))
-      else
-        given_facts = YAML.load(File.read(options[:fact_file]))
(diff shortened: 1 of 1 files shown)


lib/puppet/application/lookup.rb
```diff
@@ -331,6 +330,18 @@ def generate_scope
...
 
     node = Puppet::Node.indirection.find(node) unless node.is_a?(Puppet::Node) # to allow unit tests to pass a node instance
+
+    if options[:fact_file]
+      original_facts = node.facts.values
+      if options[:fact_file].match(/.*\.json/)
+        given_facts = JSON.parse(File.read(options[:fact_file]))
+      else
+        given_facts = YAML.load(File.read(options[:fact_file]))
+      end
```
hlindberg: Both of the above paths an fail on faulty data, missing file etc. When that happens does those errors surface and give intelligent feedback to the user, or should the errors be rescued and reported?

lib/puppet/application/lookup.rb
```diff
@@ -331,6 +330,18 @@ def generate_scope
     end
 
     node = Puppet::Node.indirection.find(node) unless node.is_a?(Puppet::Node) # to allow unit tests to pass a node instance
+
+    if options[:fact_file]
+      original_facts = node.facts.values
+      if options[:fact_file].match(/.*\.json/)
```
hlindberg: Better to use the `String.end_with?` method. (Faster, when not requiring a pattern based search, and does not require human interpretation of the regexp). (The regexp is not quite right BTW :-) since it will return a match for strings with ".json" anywhere inside). Just change to "end_with?" which is a string comparisson, which is wanted here anyway.

hlindberg:
The help text for `--facts` needs an update as well since we changed our mind from the initial spec for this feature.

lib/puppet/application/lookup.rb
```diff
@@ -331,6 +331,25 @@ def generate_scope
     end
 
     node = Puppet::Node.indirection.find(node) unless node.is_a?(Puppet::Node) # to allow unit tests to pass a node instance
+
+    if options[:fact_file]
```
thallgren: Nitpick but given that `options[:fact_file}` is accessed up to four times, wouldn't it be better to first extract it to a `fact_file` variable and avoid repeated hash lookups?

lib/puppet/application/lookup.rb
```diff
@@ -331,6 +331,25 @@ def generate_scope
...
+      original_facts = node.facts.values
+      if options[:fact_file].end_with?("json")
+        given_facts = JSON.parse(File.read(options[:fact_file]))
+      else
+        given_facts = YAML.load(File.read(options[:fact_file]))
+
+        # If the data is not correctly formatted this will be a string so we should warn the user
+        # Otherwise both JSON and YAML will provide sane parse errors for the user if the data is
+        # incorrectly formatted
+        if !given_facts.instance_of?(Hash)
```
thallgren: Use `unless` instead of `if !`

lib/puppet/application/lookup.rb
```diff
@@ -57,7 +57,6 @@ class Puppet::Application::Lookup < Puppet::Application
     options[:node] = arg
   end
 
-  # not yet supported
   option('--facts FACT_FILE') do |arg|
     if %w{.yaml .yml .json}.include?(arg.match(/\.[^.]*$/)[0])
```
thallgren: Perhaps prudent to warn the user that the argument is ignored (and why) when it doesn't match the given regexp.

lib/puppet/application/lookup.rb
thallgren: This check is applied when using YAML only. This check is needed for JSON as well (the parsed result cannot be a String when parsing JSON but it might be an Array). So move the check out of this `if`

lib/puppet/application/lookup.rb
HAIL9000: There's an else statement here (that GitHub is hiding) which will raise and error if the wrong type of file is passed in. Does that seem correct?

HAIL9000:
@hlindberg @thallgren Updated!

lib/puppet/application/lookup.rb
thallgren: Yes. Don't understand how I could miss that one. Sorry for the noice.

PR #4363, opened 2015-10-21
(PUP-5339) Update mount tests to work in AIX

whopper:
(PUP-5339) Add mount_utils acceptance helper library

This commit adds new helpers to assist with dealing with mountpoints
and fstab on testing hosts. It is used in the suite of mount resource tests.

(PUP-5339) Update mount tests to work in AIX

This commit updates each of the mount resource tests to work in
AIX. The main difference between AIX and regular Linuxes is that
we need to use `crfs` to properly create a filesystem. Slight
modifications have been made to deal with the small differences
in setup before running puppet.

### DIFF
--- a/acceptance/lib/puppet/acceptance/mount_utils.rb
+++ b/acceptance/lib/puppet/acceptance/mount_utils.rb
@@ -1,0 +1,81 @@
+module Puppet
+  module Acceptance
+    module MountUtils
+
+      # Return the absolute path to the filesystem table file.
+      # @param host [String] hostname
+      # @return [String] path to the filesystem table file.
+      def filesystem_file(host)
+        case host['platform']
+        when /aix/
+          '/etc/filesystems'
+        when /el|fedora|sles|debian|ubuntu/
+          '/etc/fstab'
+        else
+          # TODO: Add Solaris and OSX support, as per PUP-5201 and PUP-4823
+          fail_test("Unable to determine filesystem table file location for #{host['platform']}")
+        end
+      end
+
+      # Return a standard filesystem type to use when creating filesysytems.
+      # @param host [String] hostname
+      # @return [String] filesystem type.
+      def filesystem_type(host)
+        case host['platform']
+        when /aix/
+          maj_version = on(host, facter('operatingsystemmajrelease')).stdout.chomp.to_i
+          if maj_version >= 6100
+            'jfs2'
+          else
+            'jfs'
+          end
+        when /el|fedora|sles|debian|ubuntu/
+          'ext3'
+        else
+          # TODO: Add Solaris and OSX support, as per PUP-5201 and PUP-4823
+          fail_test("Unable to determine a standard filesystem table type for #{host['platform']}")
+        end
+      end
+
+      # Appends a new filesystem entry to the filesystem table.
+      # @param host [String] hostname.
+      # @param mount_name [String] the name of the mount point. We use /tmp/name as the
+      # new filesystem, and /name as the actual mount point.
+      def add_entry_to_filesystem_table(host, mount_name)
+        fs_file = filesystem_file(host)
+        fs_type = filesystem_type(host)
+
+        case host['platform']
+        when /aix/
+          # Note: /dev/hd8 is the default jfs logging device on AIX.
+          on(host, "echo '/#{mount_name}:\n  dev = /dev/#{mount_name}\n  vfs = #{fs_type}\n  log = /dev/hd8' >> #{fs_file}")
+        when /el|fedora|sles|debian|ubuntu/
+          on(host, "echo '/tmp/#{mount_name}  /#{mount_name}  #{fs_type}  loop  0  0' >> #{fs_file}")
+        else
+          # TODO: Add Solaris and OSX support, as per PUP-5201 and PUP-4823
+          fail_test("Adding entries to the filesystem table on #{host['platform']} is not currently supported.")
+        end
+      end
+
+      # Creates a new filesystem on the host.
+      # @param host [String] hostname
+      # @param mount_name [String] the name of the mount point.
+      def create_filesystem(host, mount_name)
+        fs_type = filesystem_type(host)
+
+        case host['platform']
+        when /aix/
+          volume_group = on(host, 'lsvg').stdout.split("\n")[0]
+          on(host, "mklv -y #{mount_name} #{volume_group} 1M")
+          on(host, "mkfs -V #{fs_type} -l #{mount_name} /dev/#{mount_name}")
+        when /el|fedora|sles|debian|ubuntu/
+          on(host, "dd if=/dev/zero of=/tmp/#{mount_name} count=10240", :acceptable_exit_codes => [0,1])
+          on(host, "yes | mkfs -t #{fs_type} -q /tmp/#{mount_name}", :acceptable_exit_codes => (0..254))
+        else
+          # TODO: Add Solaris and OSX support, as per PUP-5201 and PUP-4823
+          fail_test("Creating filesystems on #{host['platform']} is not currently supported.")
+        end
+      end
+    end
+  end
+end
--- a/acceptance/tests/resource/mount/defined.rb
+++ b/acceptance/tests/resource/mount/defined.rb
@@ -2,16 +2,20 @@ test_name "defined should create an entry in filesystem table"
 
 confine :except, :platform => ['windows']
 confine :except, :platform => /osx/ # See PUP-4823
-confine :except, :platform => /solaris/
-confine :except, :platform => /aix/
+confine :except, :platform => /solaris/ # See PUP-5201
 
-fstab = '/etc/fstab'
+require 'puppet/acceptance/mount_utils'
+extend Puppet::Acceptance::MountUtils
+
 name = "pl#{rand(999999).to_i}"
 
 agents.each do |agent|
+  fs_file = filesystem_file(agent)
+  fs_type = filesystem_type(agent)
+
   teardown do
     #(teardown) restore the fstab file
-    on(agent, "mv /tmp/fstab #{fstab}", :acceptable_exit_codes => (0..254))
+    on(agent, "mv /tmp/fs_file_backup #{fs_file}", :acceptable_exit_codes => (0..254))
     #(teardown) umount disk image
     on(agent, "umount /#{name}", :acceptable_exit_codes => (0..254))
     #(teardown) delete disk image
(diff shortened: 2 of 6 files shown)


acceptance/lib/puppet/acceptance/mount_utils.rb
```diff
@@ -0,0 +1,56 @@
...
+    module MountUtils
+
+      # Return the absolute path to the filesystem table file.
+      # @param host [String] hostname
+      # @return [String] path to the filesystem table file.
+      def filesystem_file(host)
+        if host['platform'] =~ /aix/
+          '/etc/filesystems'
+        else
+          '/etc/fstab'
```
johnduarte: since we are pulling this out into a utility helper, we should guard more stringently for other platforms. For example, it looks like solaris uses `/etc/vfstab`. The default might need to be an error case.

Serving suggestion:
        case host['platform']
        when /aix/
          '/etc/filesystems'
        when /solaris/
          '/etc/vfstab'
        when /el||fedora|sles|debian|ubuntu/
          '/etc/fstab'
        else
          ''
        end

acceptance/lib/puppet/acceptance/mount_utils.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      # Return a standard filesystem type to use when creating filesysytems.
+      # @param host [String] hostname
+      # @return [String] filesystem type.
+      def filesystem_type(host)
+        if host['platform'] =~ /aix/
+          'jfs2'
+        else
+          'ext3'
```
johnduarte: This should also be more guarded now that it is in a helper. See comment for line 12.

acceptance/tests/resource/mount/destroy.rb
```diff
@@ -21,31 +24,28 @@
...
   on(agent, "mkdir /#{name}", :acceptable_exit_codes => [0,1])
 
-  step "(setup) add entry to filesystem table"
-  on(agent, "echo '/tmp/#{name}  /#{name}  ext3  loop  0  0' >> #{fstab}")
+  step "(setup) create new filesystem to be mounted"
+  create_filesystem(agent, name)
+
+  if agent['platform'] !~ /aix/
+    step "(setup) add entry to filesystem table"
+    add_entry_to_filesystem_table(agent, name)
```
johnduarte: You have logic to handle aix in this helper. So, why you are excluding it here?

acceptance/tests/resource/mount/modify.rb
```diff
@@ -22,18 +25,19 @@
...
   on(agent, "mkdir /#{name}", :acceptable_exit_codes => [0,1])
 
-  step "(setup) add entry to filesystem table"
-  on(agent, "echo '/tmp/#{name}  /#{name}  ext3  loop  0  0' >> #{fstab}")
+  step "(setup) create new filesystem to be mounted"
+  create_filesystem(agent, name)
+
+  if agent['platform'] !~ /aix/
+    step "(setup) add entry to filesystem table"
+    add_entry_to_filesystem_table(agent, name)
```
johnduarte: Same comment as above.

acceptance/lib/puppet/acceptance/mount_utils.rb
```diff
@@ -0,0 +1,56 @@
...
+      # @param host [String] hostname.
+      # @param mount_name [String] the name of the mount point. We use /tmp/name as the
+      # new filesystem, and /name as the actual mount point.
+      def add_entry_to_filesystem_table(host, mount_name)
+        fs_file = filesystem_file(host)
+
+        if host['platform'] =~ /aix/
+          on(host, "echo '/#{mount_name}:\n  dev = /tmp/#{mount_name}\n  vfs = jfs2' >> #{fs_file}")
+        else
+          on(host, "echo '/tmp/#{mount_name}  /#{mount_name}  ext3  loop  0  0' >> #{fs_file}")
```
johnduarte: Should this also be more guarded now that it is in a helper? See comment for line 12.

acceptance/lib/puppet/acceptance/mount_utils.rb
```diff
@@ -0,0 +1,56 @@
...
+      # @param host [String] hostname
+      # @param mount_name [String] the name of the mount point.
+      # Note: On AIX, `crfs` creates the filesystem *and* adds it to /etc/filesystems.
+      def create_filesystem(host, mount_name)
+        if host['platform'] =~ /aix/
+          volume_group = on(host, 'lsvg').stdout.split("\n")[0]
+          on(host, "crfs -v jfs2 -a size=1M -g #{volume_group} -m /#{mount_name}")
+        else
+          on(host, "dd if=/dev/zero of=/tmp/#{mount_name} count=10240", :acceptable_exit_codes => [0,1])
+          on(host, "yes | mkfs -t ext3 -q /tmp/#{mount_name}", :acceptable_exit_codes => (0..254))
```
johnduarte: Should this also be more guarded now that it is in a helper? See comment for line 12.

acceptance/tests/resource/mount/destroy.rb
whopper: @johnduarte that's because in the tests where we actually create a filesystem manually, we run `crfs` on AIX, which actually adds the entry to `/etc/filesystems` itself. However, there are some tests (`mount/defined.rb` is one example) where we don't manually create filesystems, but need to append an entry to fstab, hence the need for that method for AIX. Perhaps it'd be more clear if I made it so `create_filesystem` ends up appending the entry for both AIX and linux, for consistency. As it stands, it's a little wonky looking.

acceptance/lib/puppet/acceptance/mount_utils.rb
whopper: Makes sense - I'll look into the Solaris case.

acceptance/tests/resource/mount/destroy.rb
whopper: ^ Actually, yes - not having that method append the new filesystem to fstab everywhere is dumb. I'll update that and get rid of this bit.

johnduarte: Got it. Thanks @whopper

whopper:
@johnduarte I made a bunch of updates, and I believe this should be ready to go. I've been able to verify this passes on AIX 5.3 and 6.1 (the -agent boxes, haven't been able to test them on the -acceptance boxes yet), as well as some of the regular platforms (redhat 7, debian 8, fedora 21...).

acceptance/tests/resource/mount/defined.rb
```diff
@@ -3,15 +3,19 @@
 confine :except, :platform => ['windows']
 confine :except, :platform => /osx/ # See PUP-4823
 confine :except, :platform => /solaris/
```
johnduarte: Should we add the PUP-5201 ticket reference here?

acceptance/tests/resource/mount/destroy.rb
```diff
@@ -3,49 +3,51 @@
 confine :except, :platform => ['windows']
 confine :except, :platform => /osx/ # See PUP-4823
 confine :except, :platform => /solaris/
```
johnduarte: Should we add the PUP-5201 ticket reference here?

acceptance/tests/resource/mount/modify.rb
```diff
@@ -3,37 +3,44 @@
 confine :except, :platform => ['windows']
 confine :except, :platform => /osx/ # See PUP-4823
 confine :except, :platform => /solaris/
```
johnduarte: Should we add the PUP-5201 ticket reference here?

acceptance/tests/resource/mount/destroy.rb
```diff
@@ -3,49 +3,51 @@
...
+
   teardown do
-    #(teardown) restore the fstab file
-    on(agent, "mv /tmp/fstab #{fstab}", :acceptable_exit_codes => (0..254))
     #(teardown) umount disk image
     on(agent, "umount /#{name}", :acceptable_exit_codes => (0..254))
     #(teardown) delete disk image
-    on(agent, "rm /tmp/#{name}", :acceptable_exit_codes => (0..254))
+    if agent['platform'] =~ /aix/
+      on(agent, "yes | rmlv #{name}")
```
johnduarte: Will this return a non-zero exit code if the logical volume was never created?

acceptance/tests/resource/mount/modify.rb
```diff
@@ -3,37 +3,44 @@
...
+
   teardown do
-    #(teardown) restore the fstab file
-    on(agent, "mv /tmp/fstab #{fstab}", :acceptable_exit_codes => (0..254))
     #(teardown) umount disk image
     on(agent, "umount /#{name}", :acceptable_exit_codes => (0..254))
     #(teardown) delete disk image
+    if agent['platform'] =~ /aix/
+      on(agent, "yes | rmfs /dev/#{name}")
+      on(agent, "yes | rmlv #{name}")
```
johnduarte: Do we need to guard against non-zero exit codes if the fs or lv were never created?

whopper: @johnduarte yeah, good point. I'll add an `:acceptable_exit_codes` bit like the umount step.

acceptance/tests/resource/mount/mounted.rb
```diff
@@ -3,45 +3,61 @@
...
+
   teardown do
-    #(teardown) restore the fstab file
-    on(agent, "mv /tmp/fstab #{fstab}", :acceptable_exit_codes => (0..254))
     #(teardown) umount disk image
     on(agent, "umount /#{name}", :acceptable_exit_codes => (0..254))
     #(teardown) delete disk image
-    on(agent, "rm /tmp/#{name}", :acceptable_exit_codes => (0..254))
+    if agent['platform'] =~ /aix/
+      on(agent, "yes | rmlv #{name}")
```
johnduarte: Will this return a non-zero exit code if the logical volume was never created?

acceptance/tests/resource/mount/query.rb
```diff
@@ -1,35 +1,32 @@
...
 
   step "(setup) add entry to filesystem table"
-  on(agent, "echo '/tmp/#{name}  /#{name}  0  0' >> #{fstab}")
+  add_entry_to_filesystem_table(agent, name)
 
   #------- TESTS -------#
   step "verify mount with puppet"
   on(agent, puppet_resource('mount', "/#{name}")) do |res|
-    fail_test "didn't find the mount #{name}" unless res.stdout.include? "#{name}"
+    fail_test "didn't find the mount #{name}" unless res.stdout.match(/'\/#{name}':\s+ensure\s+=>\s+'unmounted'/)
```
johnduarte: Don't we want to verify that the resource is `ensure => mounted` rather than `unmounted`?

whopper: @johnduarte that was actually one of the original issues with this test - we never mount the filesystem (or even create it). We're just adding it the the filesystem table and seeing if puppet knows about it. Since it's not mounted, puppet will return with `unmounted`. It's necessary to check for that, though, because you can ask puppet about any non-existent mount resource and it will return with,

```
mount { 'nonexist':
    ensure => absent
}
```
 - meaning that this test would always pass, even if the entry wasn't in the fstab.

johnduarte: Thanks @whopper :+1:

johnduarte:
:+1: from me. Do we need other eyes on this?

whopper:
@branan, maybe, if he's done much with filesystems on AIX?

joshcooper:
@whopper @johnduarte should we retarget for master so that we can finish reviewing and merging before the end of the sprint?

branan:
I'd be +1 to retargeting

johnduarte:
:+1:  for retarget to master. It is not a blocker for 1.2.7 agent

PR #4368, opened 2015-10-21
(PUP-5391) Fixup AIX nim provider acceptance test

whopper:
This commit adds a teardown step to the nim package provider
test which removes the packages it installs once the test is
complete. This prevents failures on subsequent runs on the same
host when the test attempts to install the same package. In addition,
we now check stderr instead of stdout for the expected downgrade
error.


stahnma:
:+1: from me.

kylog:
:+1: 

Nice sleuthing @whopper.

whopper:
Acually, one sec - gonna bring the teardown step in line with the other AIX package test.

johnduarte:
Thank you @whopper . Tested on AIX 5.3 :+1:

PR #4374, opened 2015-10-23
(maint) Update teardown for aix pkg acceptance

johnduarte:
This commit updates the teardown step for the
`aix_package_provider` acceptance test to not use puppet
to remove the package.

This test is intended to test the provider. So, even if there
is a bug in the provider, we want to make sure the package is
removed. This commit uses the native AIX package call to ensure
that the package is uninstalled.

[skip ci]


joshcooper:
If this is for 4.3.0, please retarget for stable

johnduarte:
Closing to retarget for stable

PR #4375, opened 2015-10-23
(DO NOT MERGE) (PUP-4516) Exit immediately when sent INT or TERM

joshcooper:
Starting in ruby 2, signal handlers are not allowed to acquire a
mutex[1], because it could deadlock[2]. For example, the main thread
acquires a mutex, is interrupted by a signal, signal handler runs in a
separate thread, and will deadlock if tries to acquire the already
locked mutex. In ruby 1.8/9, the behavior was allowed, but wasn't safe.
In ruby 2, it's a hard failure.

Commit 6238616 removed thread-safety code because it was incomplete and
puppet primarily runs single threaded -- exceptions being its signal
handler and webrick request processing. PUP-202 was marked as fixed
since we no longer tried to acquire a mutex from a trap context.
However, the signal handler would still call back into puppet code, e.g.
reload, which led to PUP-1635 "current thread not owner".

Commit 245de036f modified the signal handler to queue the method to
execute, e.g. reload, and have the Daemon's event_loop periodically
check if a new signal had been received, calling the method from the
main ruby thread.

However, the event loop checks for new signals every 5 seconds, so
puppet will not immediately exit when sent INT or TERM. Also, if puppet
is making a blocking system call, e.g. agent is blocked reading from a
socket, then puppet will not exit until the system call returns.

This commit changes puppet's signal handlers to exit immediately in
response to INT or TERM. We call stop instead of exit to gracefully
shutdown, e.g. delete our pid file. There is a possibility the stop
method could cause a mutex to be acquired, e.g. in ruby's webrick or
logger, but it doesn't seem to happen in practice.

Also we preserve the existing async handling for HUP, USR1, and USR2.

This commit affects the agent and webrick master as both support
daemonizing.

[1] https://github.com/ruby/ruby/blob/v2_1_7/thread.c#L4384-L4386
[2] https://bugs.ruby-lang.org/issues/6416

### DIFF
--- a/spec/unit/daemon_spec.rb
+++ b/spec/unit/daemon_spec.rb
@@ -45,15 +45,24 @@ describe Puppet::Daemon, :unless => Puppet.features.microsoft_windows? do
   end
 
   describe "when setting signal traps" do
-    signals = {:INT => :stop, :TERM => :stop }
-    signals.update({:HUP => :restart, :USR1 => :reload, :USR2 => :reopen_logs}) unless Puppet.features.microsoft_windows?
-    signals.each do |signal, method|
-      it "should log and call #{method} when it receives #{signal}" do
-        Signal.expects(:trap).with(signal).yields
-
-        Puppet.expects(:notice)
+    [:INT, :TERM].each do |signal|
+      it "logs a notice and exits when sent #{signal}" do
+        Signal.stubs(:trap).with(signal).yields
+        Puppet.expects(:notice).with('Exiting...')
+        daemon.expects(:stop)
 
         daemon.set_signal_traps
+      end
+    end
+
+    {:HUP => :restart, :USR1 => :reload, :USR2 => :reopen_logs}.each do |signal, method|
+      it "logs a notice and remembers to call #{method} when it receives #{signal}" do
+        Signal.stubs(:trap).with(signal).yields
+        Puppet.expects(:notice).with("Caught #{signal}; storing #{method}")
+
+        daemon.set_signal_traps
+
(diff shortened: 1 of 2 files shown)


joshcooper:
Getting this up for review for now. We'd like to get this fixed in 3.8.5 and 4.3.0 but are in the process of releasing and need to hold off on merging.

kylog:
Can we target 3.x for this fix?

spec/unit/daemon_spec.rb
```diff
@@ -45,15 +45,24 @@ def run_loop(jobs)
...
 
   describe "when setting signal traps" do
-    signals = {:INT => :stop, :TERM => :stop }
-    signals.update({:HUP => :restart, :USR1 => :reload, :USR2 => :reopen_logs}) unless Puppet.features.microsoft_windows?
-    signals.each do |signal, method|
-      it "should log and call #{method} when it receives #{signal}" do
-        Signal.expects(:trap).with(signal).yields
+    [:INT, :TERM].each do |signal|
+      it "logs a notice and exits when sent #{signal}" do
+        Signal.stubs(:trap).with(signal).yields
```
joshcooper: This is a slightly evil way to test the code. Normally `Signal.trap` just registers a signal handler to be invoked when it receives the signal at some time in the future. But this stub causes the handler to be invoked when `set_signal_traps` is called. This pattern existed previously (above on line 52), so I left it as is.

joshcooper:
Superceded by PR #4382 which targets 3.x

PR #4376, opened 2015-10-23
(PUP-5427) Don't set PATH in puppet's service scripts

geoffnichols:
Prior to this commit, some of puppet's service
scripts were setting PATH. This appears to be
unneccesary. Additionally, the specific PATH that
was being set could lead to unexpected use of the
puppet-agent's "private" utilities, such as gem.

This commit removes PATH where it was set in
puppet's service scripts.

### DIFF
--- a/ext/debian/puppet.init
+++ b/ext/debian/puppet.init
@@ -8,7 +8,6 @@
 # Default-Stop:      0 1 6
 ### END INIT INFO
 
-PATH=/opt/puppetlabs/puppet/bin:/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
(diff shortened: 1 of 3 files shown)


haus:
@geoffnichols is SLES/solaris unaffected here?

geoffnichols:
@haus I have only checked the service configs we're including in the AIO puppet-agent, but it looks like we're not setting PATH on Solaris, SLES, and platforms using systemd.

haus:
:+1:

melissa:
and these work without path? I'm assuming yes since you said it was unnecessary, but just double checking

geoffnichols:
@melissa Yes, these appear to work without setting the path.

PR #4377, opened 2015-10-27
(PUP-5309) Correct behavior of yum package provider regarding yum-sec…

dreamlibrarian:
…urity plugin output.

Without this patch, yum check-updates output containing yum-security
plugin output causes the provider to bail and return undefined method
`[]' for nil:NilClass for all yum package resources. Security output
trails the check-updates output, so we can safely halt after these lines
start appearing.


MikaelSmith:
Other plugins seem like they could cause similar issues. Would it be safe to run `yum check-updates --noplugins`? Are there other plugins that might have behavior that's required?

/cc @stahnma

MikaelSmith:
Another option might be to add a way to configure which plugins are disabled using `--disableplugin`.

dreamlibrarian:
noplugins would get in the way if one was using spacewalk/satellite or pulp. Could probably blacklist plugins that get in the way. That's likely to be fun with discovery. 

Could also just be a little more picky in parsing. This is a 'right now' solution, to be sure.

vinzent:
I used repoquery for better scriptable yum repo interaction. downside: its in the yum-utils packge.

```repoquery --plugins --all --pkgnarrow=updates```

can be used with the ```-queryformat="%{NAME}"``` param (includes newline in contrary to the rpm command)

stahnma:
I tend to like repoquery better as well, but the dep on yum-utils is not so great. 

I think this solution fixes the acute problem, but we may run into issues with other plugins.

PR #4378, opened 2015-10-27
(PUP-5395) Make lookup sensitive to 'lookup_options' metadata

thallgren:
This PR is an elaboration of [PR-4358](/puppetlabs/puppet/pull/4358). The following was added in addition to the commits included in that PR:

1. Refactoring to get rid of lots of duplicated code and some inconsistencies in how a "key not found" was handled.
2. Rename of unit test data_binding.rb to data_binding_spec.rb (it wasn't included in a normal rspec run because of this ,hence the problem when executed in isolation).
3. Use of 'lookup_options' hash instead of individual 'lookupoptions::*' keys.
4. Removal of global `hiera_advanced_parameter_bindings`
5. Adjusted and added tests.

I would recommend reviewing the full PR first and consult individual commits when/if there's a need for it since some code that was added was later removed or changed.


lib/puppet/indirector/hiera.rb
```diff
@@ -18,6 +18,7 @@ def initialize(*args)
   def find(request)
     not_found = Object.new
     options = request.options
+    Puppet.debug "Performing a hiera indirector lookup of #{request.key} with options #{options.inspect}"
```
hlindberg: Always call `Puppet.debug` with a no-args-lambda if the argument requires evaluation (here it requires interpolation). i.e. 
```
Puppet.debug { "Performing a hiera indirector lookup of #{request.key} with options #{options.inspect}" }
```

lib/puppet/data_providers.rb
```diff
@@ -4,33 +4,27 @@ def self.assert_loaded
...
     end
     @loaded = true
   end
 
+  # @deprecated use `lookup_adapter(lookup_invocation).lookup` instead
   def self.lookup_in_environment(name, lookup_invocation, merge)
-    assert_loaded()
-    adapter = DataAdapter.adapt(Puppet.lookup(:current_environment))
-    adapter.env_provider.lookup(name, lookup_invocation, merge)
+    Puppet.deprecation_warning('The method Puppet::DataProviders.lookup_in_environment is deprecated and will be removed in the next major release of Puppet.')
```
hlindberg: The deprecation needs a ticket targeting Puppet 5.0.0 so we do not forget to remove the deprecated methods.

lib/puppet/data_providers/lookup_adapter.rb
```diff
@@ -0,0 +1,199 @@
...
+    # Support running tests without an injector being configured == using a null implementation
+    return EnvironmentDataProvider.new() unless injector
+
+    # Get the environment's configuration since we need to know which data provider
+    # should be used (includes 'none' which gets a null implementation).
+    #
+    env_conf = @env.configuration
+
+    # Get the data provider and find the bound implementation
+    # TODO: PUP-1640, drop the nil check when legacy env support is dropped
```
hlindberg: legacy env support has been dropped in puppet 4.x - there is always an env.conf now.

hlindberg:
I logged a ticket for the future removal of deprecated methods.

hlindberg:
meh - already added (did not read the commit message before I added new ticket; marked the one I added as a duplicate).

PR #4379, opened 2015-10-27
(PUP-5339) Update mount tests to work in AIX

whopper:
(PUP-5339) Add mount_utils acceptance helper library

This commit adds new helpers to assist with dealing with mountpoints
and fstab on testing hosts. It is used in the suite of mount resource tests.

(PUP-5339) Update mount tests to work in AIX

This commit updates each of the mount resource tests to work in
AIX. The main difference between AIX and regular Linuxes is that
we need to use crfs to properly create a filesystem. Slight
modifications have been made to deal with the small differences
in setup before running puppet.

### DIFF
--- a/acceptance/lib/puppet/acceptance/mount_utils.rb
+++ b/acceptance/lib/puppet/acceptance/mount_utils.rb
@@ -1,0 +1,81 @@
+module Puppet
+  module Acceptance
+    module MountUtils
+
+      # Return the absolute path to the filesystem table file.
+      # @param host [String] hostname
+      # @return [String] path to the filesystem table file.
+      def filesystem_file(host)
+        case host['platform']
+        when /aix/
+          '/etc/filesystems'
+        when /el|fedora|sles|debian|ubuntu/
+          '/etc/fstab'
+        else
+          # TODO: Add Solaris and OSX support, as per PUP-5201 and PUP-4823
+          fail_test("Unable to determine filesystem table file location for #{host['platform']}")
+        end
+      end
+
+      # Return a standard filesystem type to use when creating filesysytems.
+      # @param host [String] hostname
+      # @return [String] filesystem type.
+      def filesystem_type(host)
+        case host['platform']
+        when /aix/
+          maj_version = on(host, facter('operatingsystemmajrelease')).stdout.chomp.to_i
+          if maj_version >= 6100
+            'jfs2'
+          else
+            'jfs'
+          end
+        when /el|fedora|sles|debian|ubuntu/
+          'ext3'
+        else
+          # TODO: Add Solaris and OSX support, as per PUP-5201 and PUP-4823
+          fail_test("Unable to determine a standard filesystem table type for #{host['platform']}")
+        end
+      end
+
+      # Appends a new filesystem entry to the filesystem table.
+      # @param host [String] hostname.
+      # @param mount_name [String] the name of the mount point. We use /tmp/name as the
+      # new filesystem, and /name as the actual mount point.
+      def add_entry_to_filesystem_table(host, mount_name)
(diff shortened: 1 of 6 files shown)


whopper:
@joshcooper @johnduarte @branan - retargeted version of https://github.com/puppetlabs/puppet/pull/4363

acceptance/tests/resource/mount/mounted.rb
```diff
@@ -2,46 +2,62 @@
...
+
   teardown do
-    #(teardown) restore the fstab file
-    on(agent, "mv /tmp/fstab #{fstab}", :acceptable_exit_codes => (0..254))
     #(teardown) umount disk image
     on(agent, "umount /#{name}", :acceptable_exit_codes => (0..254))
     #(teardown) delete disk image
-    on(agent, "rm /tmp/#{name}", :acceptable_exit_codes => (0..254))
+    if agent['platform'] =~ /aix/
+      on(agent, "yes | rmlv #{name}", :acceptable_exit_codes => (0..254))
```
branan: O_O

whopper: @branan see https://github.com/puppetlabs/puppet/pull/4379/files#diff-fbfe185267e39ea41040cf8aa5ef7414R69. We're creating a logical volume for the test, and this is removing it when we're done.

acceptance/tests/resource/mount/mounted.rb
```diff
@@ -2,46 +2,62 @@
...
   step "create a mount with puppet (mounted)"
-  args = ['ensure=mounted',
-          'fstype=ext3',
-          'options=loop',
-          "device='/tmp/#{name}'"
-         ]
+  if agent['platform'] =~ /aix/
+    args = ['ensure=mounted',
+            "fstype=#{fs_type}",
+            "options='log=/dev/hd8'",
```
branan: Can we determine this programmatically instead of hard coding it? I don't want this test to fail if we ever rebuild our AIX infra or something

whopper: @branan probably, though `/dev/hd8` is the default jfs/jfs2 logging device on AIX (see http://www-01.ibm.com/support/docview.wss?uid=isg3T1000398).

"NOTE: AIX creates the default log device for rootvg on /dev/hd8"

acceptance/tests/resource/mount/mounted.rb
branan: my fright was at the `yes |` part :P

This is reasonable, it's just exciting to see an OS that doesn't have a --force flag for everything

acceptance/tests/resource/mount/mounted.rb
branan: if /dev/hd8 is the default I care a lot less

acceptance/lib/puppet/acceptance/mount_utils.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      # Return a standard filesystem type to use when creating filesysytems.
+      # @param host [String] hostname
+      # @return [String] filesystem type.
+      def filesystem_type(host)
+        case host['platform']
+        when /aix/
+          maj_version = on(host, facter('operatingsystemmajrelease')).stdout.chomp.to_i
+          if maj_version >= 6100
```
branan: Oh, AIX

branan:
Nothing in this is jumping out as terrible to me

acceptance/tests/resource/mount/mounted.rb
whopper: @branan oh it totally does have a force option! Updated :)

PR #4372, opened 2015-10-22
(PUP-5448) Add cumulus install support

johnduarte:
This commit adds conditional logic to find cumulus repo file
in order to install on cumulus.

[skip ci]

PR #4380, opened 2015-10-27
(PUP-5445) Add Arista support

LuvCurves:


LuvCurves:
Requires this PR https://github.com/puppetlabs/puppet/pull/4372

acceptance/config/aio/options.rb
```diff
@@ -7,6 +7,8 @@
     'setup/common/pre-suite/000-delete-puppet-when-none.rb',
     'setup/aio/pre-suite/010_Install.rb',
     'setup/aio/pre-suite/015_PackageHostsPresets.rb',
+    'setup/aio/pre-suite/020_InstallCumulusModules.rb',
```
joshcooper: looks like this file wasn't `git add`ed

acceptance/setup/aio/pre-suite/021_InstallAristaModule.rb
```diff
@@ -0,0 +1,18 @@
+platforms = hosts.map{|val| val[:platform]}
+skip_test "No arista hosts present" unless platforms.any? { |val| /eos/ =~ val }
```
joshcooper: I think we prefer using `confine :to, :platform => /eos/`, though I defer to @johnduarte and @er0ck on that

joshcooper:
@LuvCurves Could you updated the commit message to refer to the PUP ticket for this work? Usually we generate a set of tickets when adding a new platform. /cc @geoffnichols @melissa

LuvCurves:
@joshcooper There is no Pup ticket for this work.  It is being tracked as a QA ticket.  This, like Cumulus will NOT be a platform that will be run in Puppet Agent CI, but a "NetDev" CI.

acceptance/setup/aio/pre-suite/021_InstallAristaModule.rb
LuvCurves: I need to do more than work with the master in this script, hence the lack of a confine.

acceptance/config/aio/options.rb
LuvCurves: Updated the options.rb

acceptance/setup/aio/pre-suite/021_InstallAristaModule.rb
johnduarte: @joshcooper the need is to execute work on masters and agents, but only if one or more of the agents is of type `eos`. This pattern is the best that @LuvCurves and I could come up with to identify those constraints.

kylog:
@LuvCurves our release script (and TravisCI) will balk at the QA ticket reference (it looks for PUP tickets or one of "maint" or "docs" or "packaging"). To that end, can you change the commit summary to start with "(maint)"? Also, can you squash the commits?

Thanks!

PR #4381, opened 2015-10-28
(PUP-5445) Add Arista support

LuvCurves:
This covers work done for QA-2096

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -8,11 +8,12 @@ module Puppet
     module InstallUtils
       PLATFORM_PATTERNS = {
         :redhat        => /fedora|el|centos/,
-        :debian        => /debian|ubuntu/,
+        :debian        => /debian|ubuntu|cumulus/,
         :debian_ruby18 => /debian|ubuntu-lucid|ubuntu-precise/,
         :solaris_10    => /solaris-10/,
         :solaris_11    => /solaris-11/,
         :windows       => /windows/,
+        :eos           => /^eos-/,
       }.freeze
 
       # Installs packages on the hosts.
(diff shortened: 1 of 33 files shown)


acceptance/lib/puppet/acceptance/install_utils.rb
```diff
@@ -8,11 +8,12 @@ module Acceptance
     module InstallUtils
       PLATFORM_PATTERNS = {
         :redhat        => /fedora|el|centos/,
-        :debian        => /debian|ubuntu/,
+        :debian        => /debian|ubuntu|cumulus/,
         :debian_ruby18 => /debian|ubuntu-lucid|ubuntu-precise/,
         :solaris_10    => /solaris-10/,
         :solaris_11    => /solaris-11/,
         :windows       => /windows/,
+        :eos           => /eos/,
```
kylog: Paranoid thought: make the regex match more restrictive, e.g. `/eos-4-i386/` or '/^eos-/' or something.

We're right now being bitten by the `el` match on line 10 inadvertently matching `elcapitan`, which has me thinking we should err a bit on the side of restricting matches. E.g., here, what if we start testing against `mirageos` in the future?

Thoughts? /cc @joshcooper @johnduarte

joshcooper: :+1: to more restrictive match, e.g. `beos` would match the regex

acceptance/tests/resource/cron/should_be_idempotent.rb
```diff
@@ -1,5 +1,6 @@
 test_name "Cron: check idempotency"
-confine :except, :platform => 'windows'
+confine :except, :platform => ['windows']
```
joshcooper: why the change to windows?

LuvCurves:
Recrafting

PR #4385, opened 2015-10-28
(PUP-5452) Correct parsing of crontab environment variable lines

dreamlibrarian:
Without this, crontab files containing environment variables that have spaces after the variable name and before the equal sign cannot be parsed properly.

### DIFF
--- a/lib/puppet/provider/cron/crontab.rb
+++ b/lib/puppet/provider/cron/crontab.rb
@@ -9,7 +9,7 @@ Puppet::Type.type(:cron).provide(:crontab, :parent => Puppet::Provider::ParsedFi
 
   text_line :blank, :match => %r{^\s*$}
 
-  text_line :environment, :match => %r{^\s*\w+=}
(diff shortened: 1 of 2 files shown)


HAIL9000:
@dreamlibrarian, thanks so much for the contribution! There are a few things we'd like to clean up on this before we can review it an get it merged:

1) This ticket has been moved to the PUP project in JIRA, so could you please update your commit message to reference that ticket instead (PUP-5452). Since the PA project includes many different components, we're worried about things getting confusing if we start filing tickets for the individual components under the PA project. 

2) Rather than merging upstream puppet into your topic branch, it would be better to rebase the topic branch so that we don't get the merge commit. The details of how to do this in git are a little hard to capture in a comment, so if you get stuck please feel free to ping us in #puppet-dev on IRC so we can give some more detailed instructions.

Thank you!

dreamlibrarian:
Corrected ticket association, fixed up extra merge commit. Holler if you need anything else!

HAIL9000:
Thanks for the contribution and for the quick response! :)

PR #4386, opened 2015-10-28
(PUP-4516) Add signal name in exit message

roidelapluie:
PUP-4516 and commit c9a7e6d506768cb1c00767f785339bec512d069e introduce
a message before exiting after receiving a TERM or INT signal.

This commit adds the signal name to this error message, clarifying the
fact that we actually exit because we receive a signal and specifying
the signal.

### DIFF
--- a/spec/unit/daemon_spec.rb
+++ b/spec/unit/daemon_spec.rb
@@ -50,7 +50,7 @@ describe Puppet::Daemon, :unless => Puppet.features.microsoft_windows? do
     [:INT, :TERM].each do |signal|
       it "logs a notice and exits when sent #{signal}" do
         Signal.stubs(:trap).with(signal).yields
-        Puppet.expects(:notice).with('Exiting...')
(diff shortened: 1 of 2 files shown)


spec/unit/daemon_spec.rb
```diff
@@ -50,7 +50,7 @@ def run_loop(jobs)
     [:INT, :TERM].each do |signal|
       it "logs a notice and exits when sent #{signal}" do
         Signal.stubs(:trap).with(signal).yields
-        Puppet.expects(:notice).with('Exiting...')
+        Puppet.expects(:notice).with("Caught #{signal};  exiting")
```
MikaelSmith: Typo here, you have an extra space causing the specs to fail.

roidelapluie: fixed

joshcooper:
Thanks @roidelapluie!

roidelapluie:
@joshcooper you did the hard work ;)

PR #4388, opened 2015-10-29
(PUP-5425) Lookup creates phantom node in PuppetDB

HAIL9000:
Prior to this commit, since lookup searched for a node called "default"
to obtain the node the user was running lookup from, it was creating
a ghost node in PuppetDB that did not have a hostname and was unusable.
This meant that lookup would fail in the next run because it would
attempt to obtain the hostname of the ghost node.

In order to prevent this, switch to looking up the current node
using it's certname rather than "default" so that no false node is
created.


hlindberg:
Hate to be picky, but commit message is not on the correct form ;-)

PR #4390, opened 2015-10-29
(PUP-5441) Sanitize node's trusted info

kylog:
@hlindberg take another look. Since we chatted:
* I added the ability to ignore or restore a context key, which allowed me to flesh out the spec tests. 
* I also added some explanatory comments because the priority scheme for setting `trusted_data` begged for some explanation.
* Last, note that I moved the `to_h` to the *end* of the block passed to `lookup`. In the location I had it in the previous version of this PR, that would raise if lookup failed because it would try to call that `to_h` before it even got to the block. Fun stuff.

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -413,7 +413,7 @@ class Puppet::Parser::Compiler
   def_delegator :@catalog, :resource, :findresource
 
   def initialize(node, options = {})
-    @node = node
+    @node = sanitize_node(node)
     # Array of resources representing all application instances we've found
     @applications = []
     # We use @current_app and @current_components to signal to the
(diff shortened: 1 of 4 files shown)


kylog:
Squashed. More spec tests needed.

lib/puppet/parser/compiler.rb
```diff
@@ -781,6 +781,59 @@ def initvars
...
+    # 1) node came with trusted_data so use that
+    # 2) else if there is :trusted_information in the puppet context
+    # 3) else if the node provided a 'trusted' parameter (parsed out above)
+    # 4) last, fallback to local node trusted information
+    #
+    # Note that trusted_data should be a hash, but (2) and (4) are not
+    # hashes, so we to_h at the end
+    node.trusted_data ||= Puppet.lookup(:trusted_information) do
+      trusted_param || Puppet::Context::TrustedInformation.local(node)
+    end.to_h
```
hlindberg: Think the `.to_h` should go last on the line above to turn the TrustedInformation object into a hash, the `trusted_param` should already be a hash (if not nil). Th problem with Ruby 1.9.3 is that Hash does not respond to `to_h` it seems (as reported in the failed travis run).

kylog:
Argh. Ruby 1.9.3. Let's remove support for ruby < 2.1 in puppet 5.

Meanwhile, I shuffled that around to check for it being a hash before `to_h`ing it.

kylog:
The travis fail looks like a transient. I'll rerun it.

PR #4391, opened 2015-10-29
(PUP-5432) Use node_name to validate report test

johnduarte:
This commit modifies the acceptance test `reports/submission.rb`
to use the `agent.node_name` rather than the beaker
`agent.hostname` value to validate the contents
of the submitted report on the master.

[skip ci]

### DIFF
--- a/acceptance/tests/reports/submission.rb
+++ b/acceptance/tests/reports/submission.rb
@@ -71,7 +71,7 @@ else
     agents.each do |agent|
       on(agent, puppet('agent', "-t --server #{master}"))
 
-      on master, "grep -q #{agent} #{testdir}/*/*"
+      on master, "grep -q #{agent.node_name} #{testdir}/*/*"
     end
   end
 


branan:
Gross.

branan:
Ideally the system hostname always matches beaker's concept of the hostname. Is there a reason we can't do that for the test system that this is failing on?

acceptance/tests/reports/submission.rb
```diff
@@ -71,7 +71,8 @@ def query_last_report_time_on(agent)
     agents.each do |agent|
       on(agent, puppet('agent', "-t --server #{master}"))
 
-      on master, "grep -q #{agent} #{testdir}/*/*"
+      hostname = fact_on(agent, 'hostname')
```
joshcooper: I thought beaker had a method for getting the host's idea of its name (perhaps using facter)?

johnduarte: @joshcooper you are correct. In this context we are looking to verify `agent.node_name` which uses `puppet agent --configprint node_name_value`. I will update accordingly.

branan:
:+1:

PR #4392, opened 2015-10-30
(PUP-4917) Windows Packages Use QuietDisplayName

ferventcoder:
When DisplayName is not available but QuietDisplayName is, use
QuietDisplayName instead. If neither are found, continue the
behavior of returning nothing. Some software in Windows 2012
will install to QuietDisplayName instead of DisplayName, which
possibly keeps it out of the Programs and Features list, but
the value should continue to show up for Puppet.


Iristyle:
Merged in https://github.com/puppetlabs/puppet/commit/5080ce50b93a089bcd078fa6777cf37be7e478a6

PR #4394, opened 2015-10-30
(PUP-5097) Rename domainname trusted fact to domain

joshcooper:
Renames the domainname trusted fact to domain, since the latter is what
we use for the Facter fact.


kylog:
:+1:

PR #4397, opened 2015-11-02
(PUP-5395) Fix problem with lookup_options not found in module

thallgren:
A check for qualified name in modules prevented lookup of the
'lookup_options' key in a module. This commit adds that check and
a test that verifies that lookup_options declared in environment
and module are merged and used.

### DIFF
--- a/lib/puppet/data_providers/hiera_support.rb
+++ b/lib/puppet/data_providers/hiera_support.rb
@@ -21,7 +21,7 @@ module Puppet::DataProviders::HieraSupport
     lookup_invocation.with(:data_provider, self) do
       merge_strategy = Puppet::Pops::MergeStrategy.strategy(merge)
       lookup_invocation.with(:merge, merge_strategy) do
-        merged_result = merge_strategy.merge_lookup(data_providers(data_key(key), lookup_invocation)) do |data_provider|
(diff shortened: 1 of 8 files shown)


hlindberg:
BTW: What happens if you call the lookup function with the key `lookup_options` ?

hlindberg:
I am fine with the rest - but curious what the intended result is of a lookup call, and wonder if there should be a test for that.

thallgren:
A lookup of 'lookup_options' will be handled as any other lookup which means that it will be found in global and environment scope but not in module scope (since it's not prefixed with a module name).

The lookup will do what it always does and start with a "meta" lookup of the key. It doesn't matter that the key is 'lookup_options'. If there is a merge instruction for the 'lookup_options' key, then it's applied during the actual data lookup. It will not affect the meta looup.

hlindberg:
That behavior when looking up "lookup_options" is a bit odd. It would have been less surprising if it actually looked up "lookup_options" in every module and produced a hash of all the options. Naturally this would only be done when looking up `lookup_options`. This behavior is useful for debugging purposes. Think this deserves its own ticket (where this problem can be discussed and a decision made).

thallgren:
For the record, "if it actually looked up "lookup_options" in every module and produced a hash of all the options" would IMO be very surprising given that it never happens in a real situation. No module will ever merge its "lookup_options" with another module.

PR #4398, opened 2015-11-02
(PUP-5465) Add missing support for Resource References in collection

trlinkin:
The support for resource references in collector search expressions is
missing. This is a documented aspect of resource collectors and
previously worked in Puppet 3.

This PR attempts to provide a solution to this issue. Full disclosure though, I'm fairly new to the way the new parser works and it's still mostly a mystery to me. I've spent some time pulling it apart and I think I've found the most reasonable solution. If there is some other materials I should look at it to understand more, that'd be awesome.

### DIFF
--- a/lib/puppet/pops/evaluator/collector_transformer.rb
+++ b/lib/puppet/pops/evaluator/collector_transformer.rb
@@ -125,6 +125,14 @@ protected
     end
   end
 
+ def query_AccessExpression(o, scope)
+    pops_object = @@evaluator.evaluate(o, scope)
+
+    # Convert to Puppet 3 style objects since that is how they are represented
+    # in the catalog.
+    @@evaluator.convert(pops_object, scope, nil)
+  end
+
   def query_VariableExpression(o, scope)
     @@evaluator.evaluate(o, scope)
   end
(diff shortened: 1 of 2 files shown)


HAIL9000:
@trlinkin, thanks for the contribution!

@hlindberg, when you get a chance could you take a look and see what you think about this fix?

hlindberg:
Looks good. +1 on merging this for 4.3 (just in time), but it is too late to backport to 3.8.4.

I know it is more difficult to add a test for exported queries, it would be useful though to have one of those as well (for the logic in the query_xxx methods). Maybe there is room to have a unit test that checks that the generated query has the expected content. (That is, if resource references are supported there).

trlinkin:
Should I try to add those tests as part of this PR or leave that work for another time? I did sort of have the same thought about that testing. I at least validated exported resources manually by hand fwiw. I'm not 100% sure how it'd be done in rspec as I'm still comprehending the whole model based thing we're using now.

hlindberg:
Not sure where it is best to test exported resources - ping @er0ck

er0ck:
IMO, this needs an acceptance test to go full cycle with puppetdb and collection.  We can add it later/soon.  
https://tickets.puppetlabs.com/browse/PUP-5489.

HAIL9000:
Merged into master. Thanks for your contribution @trlinkin!

joshcooper:
I'm thinking we should backport this to stable for the next 4.3.x release  As it is now, it won't be released until 4.4 in Jan/Feb.

cdenneen:
+1 for 4.3.x backport

kylog:
#4511 is the backport.

PR #4399, opened 2015-11-02
(PUP-4426) backport (PUP-4386) Windows group invalid user error msgs 

ferventcoder:

group { 'guests':
  ensure  => 'present',
  gid     => 'S-1-5-32-546',
}

group { 'guests':
  ensure => present,
}

Notice: Compiled catalog for vagrant-2008r2.localdomain in environment production in 0.58 seconds
Error: Could not resolve name:
Error: /Stage[main]/Main/Group[Guests]/members: change from VAGRANT-2008R2\Guest,VAGRANT-2008R2\Administrator to VAGRANT-2008R2\Guest,VAGRANT-2008R2\Administrator, failed: Could not resolve name:

### DIFF
--- a/lib/puppet/provider/group/windows_adsi.rb
+++ b/lib/puppet/provider/group/windows_adsi.rb
@@ -39,6 +39,11 @@ Puppet::Type.type(:group).provide :windows_adsi do
     return '' if users.nil? or !users.kind_of?(Array)
     users = users.map do |user_name|
       sid = Puppet::Util::Windows::SID.name_to_sid_object(user_name)
+      if !sid
+        resource.debug("#{user_name} (unresolvable to SID)")
+        next user_name
+      end
+
       if sid.account =~ /\\/
         account, _ = Puppet::Util::Windows::ADSI::User.parse_name(sid.account)
       else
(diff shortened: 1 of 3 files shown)


ferventcoder:
Same as https://github.com/puppetlabs/puppet/pull/4024 - just targeted at 3.x

lib/puppet/provider/group/windows_adsi.rb
```diff
@@ -39,6 +39,11 @@ def members_to_s(users)
     return '' if users.nil? or !users.kind_of?(Array)
     users = users.map do |user_name|
       sid = Puppet::Util::Windows::SID.name_to_sid_object(user_name)
+      if !sid
+        resource.debug("#{user_name} (unresolvable to SID)")
+        next user_name
```
joshcooper: If the desired members can't be resolved, then we should fail the run, right?

ferventcoder: Great question. Probably better as a followup. This went in originally on June 10 this year.

ferventcoder: Weird, I called out the same section, but for other reasons... https://github.com/puppetlabs/puppet/pull/4024#discussion_r32166350

joshcooper: Can we file a ticket on this?

ferventcoder: @Iristyle do you want to follow up on this?

PR #4400, opened 2015-11-03
(PUP-5428) Add a query string with the environment when using POST

olivierHa:
Hello, 

This PR is solving issue https://tickets.puppetlabs.com/browse/PUP-5428.

When the length of the uri is greater that 1024, the GET is converted in POST and all data (environment, facts) are stored in the body request.

Adding a query string when posting allows to loadbalance catalog compilation (for example) to specific puppetserver.

Regards

Olivier

### DIFF
--- a/lib/puppet/indirector/rest.rb
+++ b/lib/puppet/indirector/rest.rb
@@ -96,7 +96,8 @@ class Puppet::Indirector::REST < Puppet::Indirector::Terminus
       # WEBrick in Ruby 1.9.1 only supports up to 1024 character lines in an HTTP request
       # http://redmine.ruby-lang.org/issues/show/3991
       if "GET #{uri_with_query_string} HTTP/1.1\r\n".length > 1024
-        http_post(req, uri, body, headers)
+        uri_with_env = "#{uri}?environment=#{request.environment.name}"
+        http_post(req, uri_with_env, body, headers)
       else
         http_get(req, uri_with_query_string, headers)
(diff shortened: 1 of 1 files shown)


MikaelSmith:
This is an intriguing change. It's adding metadata that Puppet will ignore, for use by a load balancer. I assume load balancers can't rely on the POST body because it would be encrypted with HTTPS.

I don't see a problem with this change, but the API documentation should be updated. Since it appears the indirector will convert any GET to a POST, I'd add a section in https://github.com/puppetlabs/puppet/blob/master/api/docs/http_api_index.md about it.

Also, @cprice404 does this present any problems that you can think of?

olivierHa:
Hey Mikael, 

I didn't see it before, but yes, the documentation says :

`Every HTTP endpoint that dispatches to the indirector follows the form: /puppet/v3/:indirection/:key?environment=:environment`

In this particular case, when a GET with an uri greater that 1024 is converted in a POST call, this statement becomes false.

Olivier

MikaelSmith:
Ah, that should be sufficient for documentation then.

olivierHa:
Hello, any updates ?

HAIL9000:
Hey @olivierHa, with people out for the holidays we haven't been able to get the eyes we need on this change. We'll get back to you as soon as we can.

MikaelSmith:
Ok, got sufficient thumbs up.

PR #4401, opened 2015-11-03
(PUP-5278) Include global data binding in explain output

thallgren:
The global data binding was forgotten in the explain support. This
commit adds the needed output and alters all test to accomodate its
presence.


hlindberg:
Is there a spec test that shows the expected explained result when a key *is* found in global hiera? I only spotted what it looks like when key is not found in global hiera.

thallgren:
I'll add that.

PR #4402, opened 2015-11-03
(PUP-5035)  Only add child resources for directory sources

joshcooper:
Previously, the static compiler would raise a NoMethodError if a file resource had recurse => true, but the source referred to a file. This change ensures we only try to generate child resources if the source has children.


joshcooper:
I'm looking into the travis failure, I can't repro in isolation, may be order dependent spec failure.

PR #4403, opened 2015-11-03
(PUP-5387) Wait for services to finish state transitions in AIX

whopper:
Previously when stopping a service, the AIX SRC provider would hit
the service with the stop signal via `stopsrc` and return immediately
as the service asynchronously shut down. In some cases, this would
cause a race condition if the service took a moment to actually stop.
In such cases, Puppet would became unaware of the actual state of the
service.

This is due to the asychronous nature of service management in AIX,

This commit adds a 60 second timeout when stopping services to check
that the service is actually stopped before returning. If a service
has not stopped by the end, it fails, since that may affect future
resources depending on the service being stopped.

For consistency, this commit also adds the same timeout for starting
services to ensure they are actually running after we start them.

### DIFF
--- a/lib/puppet/provider/service/src.rb
+++ b/lib/puppet/provider/service/src.rb
@@ -1,3 +1,5 @@
+require 'timeout'
+
 # AIX System Resource controller (SRC)
 Puppet::Type.type(:service).provide :src, :parent => :base do
 
@@ -61,6 +63,31 @@ Puppet::Type.type(:service).provide :src, :parent => :base do
 
   def disable
     rmitab(@resource[:name])
+  end
+
+  # Wait for the service to transition into the specified state before returning.
+  # This is necessary due to the asynchronous nature of AIX services.
+  # desired_state should either be :running or :stopped.
+  def wait(desired_state)
+    Timeout.timeout(60) do
+      loop do
+        status = self.status
+        break if status == desired_state.to_sym
+        sleep(1)
+      end
+    end
+  rescue Timeout::Error
+    raise Puppet::Error.new("Timed out waiting for #{@resource[:name]} to transition states")
+  end
+
+  def start
+    super
+    self.wait(:running)
+  end
+
+  def stop
+    super
+    self.wait(:stopped)
   end
 
   def restart
--- a/spec/unit/provider/service/src_spec.rb
+++ b/spec/unit/provider/service/src_spec.rb
@@ -37,7 +37,7 @@ describe provider_class do
   end
 
   describe ".instances" do
-    it "should has a .instances method" do
+    it "should have a .instances method" do
       expect(provider_class).to respond_to :instances
     end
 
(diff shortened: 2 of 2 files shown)


whopper:
Note: this timeout works for managing any service (including the puppet service itself) via puppet resource, apply, or the daemon. One concern we had was that the daemon stopping itself would always time out here, since it'd be waiting for its own run to finish before stopping. However, this is not an issue with the signal handling fix (c9a7e6d506768cb1c00767f785339bec512d069e in the 3.x branch).

Hold off on merging this until that fix is merged up from 3.x.

lib/puppet/provider/service/src.rb
```diff
@@ -63,6 +65,23 @@ def disable
...
+    begin
+      Timeout.timeout(60) do
+        loop do
+          status = self.status
+          break if status == :stopped
+          sleep(1)
+        end
+      end
+    rescue Timeout::Error
+      raise Puppet::Error.new("Timed out waiting for #{@resource[:name]} to stop")
```
whopper: Should I be doing:

```
@resource.fail Puppet::Error, "Timed out waiting for #{@resource[:name]} to stop"
```

here?

branan: The transaction should mark the resource as failed if the provider throws exceptions. Are other providers explicitly failing their resources?

whopper: @branan ah, only in a few places. I imagine they're probably some old crufty bits still left from long ago.

lib/puppet/provider/service/src.rb
```diff
@@ -63,6 +65,23 @@ def disable
     rmitab(@resource[:name])
   end
 
+  def stop
+    ucommand(:stop)
```
branan: This should call `super`

branan:
Should we also do the timeout on `start`, just so we're complete?

whopper:
@branan updated.

branan:
PR name needs update

MikaelSmith:
:+1:

lib/puppet/provider/service/src.rb
```diff
@@ -63,6 +65,31 @@ def disable
...
   end
 
+  # Wait for the service to transition into the specified state before returning.
+  # This is necessary due to the asynchronous nature of AIX services.
+  # desired_state should either be :running or :stopped.
+  def wait(desired_state)
+    Timeout.timeout(60) do
+      loop do
+        status = self.status
+        break if status == desired_state.to_sym
```
MikaelSmith: You're already passing `desired_state` as a symbol. Is the result of `status` always guaranteed to be a symbol?

whopper: @MikaelSmith yep, `status` will either return :active or :inoperative, or nil in transition states. And yeah, I suppose the `to_sym` is unnecessary here, since we're only calling wait in two known places.

PR #4405, opened 2015-11-04
(PUP-5437) Add 'lookup_options' to the lookup explainer

thallgren:
The explainer will currently only report the actions that lookup
performs during traversal of global data binders and providers in
order to find the data for a given key. This commit adds the lookup
of the new 'lookup_options' to the report.

Since this might make the report quite verbose, this commit also
adds an extra '--explain-data' option to the lookup CLI. This option
is mutally exclusive to '--explain' and will yield a report where the
'lookup_options' traversal is excluded.

### DIFF
--- a/spec/unit/application/lookup_spec.rb
+++ b/spec/unit/application/lookup_spec.rb
@@ -99,34 +99,70 @@ describe Puppet::Application::Lookup do
     } }
 
     let(:expected_yaml_hash) { {
-      :type => :merge,
-      :merge => :first,
-      :event => :result,
-      :value => 'This is A',
       :branches => [
-      { :key => 'a',
-        :event => :not_found,
-        :type => :global,
-        :name => :hiera
-      },
       {
-        :type => :data_provider,
-        :name => 'Hiera Data Provider, version 4',
-        :configuration_path => "#{environmentpath}/production/hiera.yaml",
+        :type => :meta,
         :branches => [
         {
-          :type => :data_provider,
-          :name => 'common',
+          :type => :merge,
+          :merge => :hash,
           :branches => [
           {
-            :key => 'a',
-            :value => 'This is A',
-            :event => :found,
-            :type => :path,
-            :original_path => 'common',
-            :path => "#{environmentpath}/production/data/common.yaml",
+            :type => :global,
+            :key => 'lookup_options',
+            :name => :hiera,
+            :event => :not_found
+          },
+          {
+            :type => :data_provider,
+            :name => 'Hiera Data Provider, version 4',
+            :configuration_path => "#{environmentpath}/production/hiera.yaml",
+            :branches => [
+            {
+              :type => :data_provider,
+              :name => 'common',
+              :branches => [
+              {
+                :type => :path,
+                :key => 'lookup_options',
+                :original_path => 'common',
+                :path => "#{environmentpath}/production/data/common.yaml",
+                :event => :not_found
+              }]
+            }]
           }]
-       }]
+        }]
+      },
+      {
+        :type => :merge,
+        :merge => :first,
+        :event => :result,
+        :value => 'This is A',
+        :branches => [
+        { :key => 'a',
+          :event => :not_found,
+          :type => :global,
+          :name => :hiera
+        },
+        {
+          :type => :data_provider,
+          :name => 'Hiera Data Provider, version 4',
+          :configuration_path => "#{environmentpath}/production/hiera.yaml",
+          :branches => [
+          {
+            :type => :data_provider,
+            :name => 'common',
+            :branches => [
+            {
+              :key => 'a',
+              :value => 'This is A',
+              :event => :found,
+              :type => :path,
+              :original_path => 'common',
+              :path => "#{environmentpath}/production/data/common.yaml",
+              }]
+           }]
+         }]
       }]
     } }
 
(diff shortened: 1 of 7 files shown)


spec/unit/application/lookup_spec.rb
```diff
@@ -145,6 +181,34 @@
...
+  Data Binding "hiera"
+    No such key: "a"
+  Data Provider "Hiera Data Provider, version 4"
+    ConfigurationPath "#{environmentpath}/production/hiera.yaml"
+    Data Provider "common"
+      Path "#{environmentpath}/production/data/common.yaml"
+        Original path: common
+        Found key: "a" value: "This is A"
+  Merged result: "This is A"
+      EXPLANATION
```
hlindberg: Indentation of EXPLANATION looks wrong

spec/unit/application/lookup_spec.rb
```diff
@@ -145,6 +181,34 @@
...
+Merge strategy first
+  Data Binding "hiera"
+    No such key: "a"
+  Data Provider "Hiera Data Provider, version 4"
+    ConfigurationPath "#{environmentpath}/production/hiera.yaml"
+    Data Provider "common"
+      Path "#{environmentpath}/production/data/common.yaml"
+        Original path: common
+        Found key: "a" value: "This is A"
+  Merged result: "This is A"
```
hlindberg: Do we have to say "Merged result". Isn't "Result" clearer? (Otherwise it leaves the question "as opposed to what, the 'result' - are there different results?"

spec/unit/application/lookup_spec.rb
```diff
@@ -145,6 +181,34 @@
...
+Searching for 'lookup_options'
+  Merge strategy hash
+    Data Binding "hiera"
+      No such key: "lookup_options"
+    Data Provider "Hiera Data Provider, version 4"
+      ConfigurationPath "#{environmentpath}/production/hiera.yaml"
+      Data Provider "common"
+        Path "#{environmentpath}/production/data/common.yaml"
+          Original path: common
+          No such key: "lookup_options"
```
hlindberg: This just fizzles - there is no real explanation. I would expect something at first level under "Searching for 'lookup options'" to say something like "Result: no lookup_options found, using default merge strategy 'first'" (or something like that).

thallgren: The intention here is to provide the exact same format for the explanation regardless if it's the lookup_options or data that is being looked up. Advantage being that only one implementation is needed and only one format needs to be learned and documented. We can improve that format overall for sure, but I'm not too keen on having a separate format for the lookup_options.

spec/unit/data_providers/hiera_data_provider_spec.rb
```diff
@@ -190,6 +190,90 @@ def compile(environment, code = nil)
...
+  Merge strategy hash
+    Data Binding "hiera"
+      No such key: "lookup_options"
+    Data Provider "Hiera Data Provider, version 4"
+      ConfigurationPath "#{environmentpath}/hiera_misc/hiera.yaml"
+      Data Provider "common"
+        Path "#{environmentpath}/hiera_misc/data/common.yaml"
+          Original path: common
+          Found key: "lookup_options" value: {
+            "one::lopts_test::hash" => {
```
hlindberg: In practice, there could be several hundred entries in the `lookup_options` hash. Suggest we prune the hash to only show the value for the key being looked up (unless we are already doing so - hard to tell from test).

spec/unit/application/lookup_spec.rb
hlindberg: This is after all for the benefit of the user - it seems beneficial to end a "Searching" with a conclusion of the result in all cases. The tree shows how the result was obtained.

spec/unit/data_providers/hiera_data_provider_spec.rb
thallgren: If we do, then we loose the ability to view the complete set of lookup options for a specific module. We could have a specific option for that of course, but the way it stands now, the --explain will give a full explanation. Subsequent lookups may well use --explain-data which doesn't include lookup_options at all.

Another way of doing this would be to all of the following:

1. Not allow lookup of 'lookup_options' (it's not 'data', it's 'meta-data').
2. Provide a specific option that would let you view all 'lookup_options' for a given module.
3. Never display 'lookup_options' when using --explain.

Pros: Less clutter, more specific, less confusing.
Cons: You see 'Merge strategy xxx' when using --explain but have to consult the 'lookup_options' merge result from another comman to see where it actually came from.

hlindberg: 1. Agree; it is confusing
2. Agree - this should show the full content of lookup_options for that module (and only those), if no module is given, keys in the top name space would be shown.
3. Disagree, think the specifics for the looked up key should be included.

joshcooper:
If this is for PUP 4.3.0, please retarget for stable

hlindberg:
This is for 4.3.0

spec/unit/data_providers/hiera_data_provider_spec.rb
thallgren: Please define "specifics" and how it differs from what we have now. I assume that you want the explanation pruned from everything that doesn't relate to the key? If so, please consider the following:

A _lookup_options_ hash is found by a data provider but the key in question is not included. What does that mean to the explainer? I can see two alternatives:
1. Explain that the hash was found and display all keys in it so that the user can discover that the key in question is misspelled (this is what we currently do).
2. Treat this as a "not found" and prune it from the explanation.

I don't like the pruning for two reasons. a) the user will not discover misspelled keys, and b) we create a special format that will differ from other explanations of a hash merge.

My preference for #3 is to simply add an entry to the explanation stating that the merge options stem from a _lookup_options_ hash with no further info. The option in #2 can be used if the user wants specifics beyond that. Suggested option name for #2 is `--explain-options`. It takes no arguments and can be used in combination with --explain.

thallgren:
Closed to retarget to stable.

PR #4406, opened 2015-11-04
(PUP-5470) Update handle_upstart test to case out Ubuntus by version

camlow325:
This commit modifies the confine block on the handle_upstart test to
confine out the execution of the ticket_14297_handle_upstart test based
on the platform version numbers, e.g., 15.04 or greater, as opposed to
just the platform code name, e.g., Vivid or Werewolf.  This is needed
because the platform field is populated for code names on some CI
pipelines - core Puppet - but by version number on other pipelines -
e.g., Puppet Server.

### DIFF
--- a/acceptance/tests/resource/service/ticket_14297_handle_upstart.rb
+++ b/acceptance/tests/resource/service/ticket_14297_handle_upstart.rb
@@ -3,7 +3,7 @@ test_name 'Upstart Testing'
 # only run these on ubuntu vms
 confine :to, :platform => 'ubuntu'
 # vivid and above use systemd rather than upstart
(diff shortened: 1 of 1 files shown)


camlow325:
Ping @johnduarte for review.  This may not be the best approach for this, but it should work for Puppet Server's needs for now.  I'm open to another approach, but would like to get something in soon that will allow for this test to pass in the Puppet Server CI pipeline for Vivid.

camlow325:
Oh, and ping @nwolfe for second review.

johnduarte:
:+1:  lgtm @camlow325

nwolfe:
:+1:

PR #4409, opened 2015-11-05
(maint) Update teardown for aix pkg acceptance

johnduarte:
This commit updates the teardown step for the
`aix_package_provider` acceptance test to not use puppet
to remove the package.

This test is intended to test the provider. So, even if there
is a bug in the provider, we want to make sure the package is
removed. This commit uses the native AIX package call to ensure
that the package is uninstalled.

[skip ci]

### DIFF
--- a/acceptance/tests/aix/aix_package_provider.rb
+++ b/acceptance/tests/aix/aix_package_provider.rb
@@ -20,7 +20,7 @@ version2 = '1.8.6.4'
 
 teardown do
   on hosts, "rm -rf #{dir}"
-  on hosts, puppet('resource', 'package', "'#{package}' ensure=absent")
+  on hosts, "installp -u #{package}"
 end
 
 step "download packages to use for test"


acceptance/tests/aix/aix_package_provider.rb
```diff
@@ -20,7 +20,7 @@ def assert_package_version(package, expected_version)
 
 teardown do
   on hosts, "rm -rf #{dir}"
-  on hosts, puppet('resource', 'package', "'#{package}' ensure=absent")
+  on hosts, "installp -u #{package}"
```
joshcooper: I know nothing about AIX so take this with a grain of salt. It looks like puppet's nim provider can either execute [`rpm -e` or `installp -gu`](https://github.com/puppetlabs/puppet/blob/stable/lib/puppet/provider/package/nim.rb#L35-L48). Should we always be using `installp` for puppet-agent packages? Should we be using `-g` to remove dependencies also?

johnduarte: This change only affects the package installed by the test ({{sudo.rte}}) and not puppet-agent packages. I am not sure that we want this teardown to extend beyond the scope of the packages the test installs.

joshcooper: Ah right, makes sense.

PR #4411, opened 2015-11-05
(PUP-5422) Setup signal trap before waiting for certs

MikaelSmith:
When running the agent, waiting for certificates can sleep and wait for
a significant period of time. During that time a SIGTERM event can come
in (from Ctrl-C or service shutdown), causing an exception to be thrown
that prevents daemon cleanup.

Setup signal traps before waiting for certs so we ensure daemon cleanup
occurs. Update the master as well so all call paths only call
set_signal_traps once, and do so immediately after possibly daemonizing.

### DIFF
--- a/lib/puppet/application/master.rb
+++ b/lib/puppet/application/master.rb
@@ -295,6 +295,9 @@ Copyright (c) 2012 Puppet Labs, LLC Licensed under the Apache 2.0 License
     daemon.server = Puppet::Network::Server.new(Puppet[:bindaddress], Puppet[:masterport])
     daemon.daemonize if Puppet[:daemonize]
 
+    # Setup signal traps immediately after daemonization so we clean up the daemon
+    daemon.set_signal_traps
+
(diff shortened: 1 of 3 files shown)


lib/puppet/application/master.rb
```diff
@@ -295,6 +295,9 @@ def start_webrick_master
     daemon.server = Puppet::Network::Server.new(Puppet[:bindaddress], Puppet[:masterport])
     daemon.daemonize if Puppet[:daemonize]
 
+    # Setup signal traps immediately after daemonization so we clean up the daemon
+    daemon.set_signal_traps
```
joshcooper: should this happen before we daemonize like we do for the agent (when `onetime` is false)?

MikaelSmith: Seems like it, but I'm not sure how that would interact with the `fork`. Potentially we could still SIGINT/SIGTERM between daemonize and setting up signal traps.

PR #4412, opened 2015-11-05
(PUP-5262) Wait for services to finish state transitions in Solaris

whopper:
When stopping, starting, and restarting services, SMF tracks state
transitions through through two states: `state` for the
current state, and `nstate` for the next state. Because of this,
Puppet would occasionally encounter race conditions when attempting
to act on a service that was in the middle of a state transition.

This is due to the asychronous nature of service management in Solaris,

This commit adds a 60 second timeout when stopping or starting services
to ensure that the state transition has completed before continuing
operatings.


joshcooper:
If this is required for PUP 4.3.0, please retarget for stable.

PR #4414, opened 2015-11-05
(PUP-4610) Allow trailing comma in resource references

HAIL9000:
Prior to this commit, a trailing comma was not accepted in multi
resource references (e.g. $x = $y[1,2,] was not allowed). Update
the grammar so that this case is parsed properly.


joshcooper:
This will need to be backported to stable to make 4.3.0

PR #4416, opened 2015-11-05
(PUP-5262) Wait for services to finish state transitions in Solaris

whopper:
When stopping, starting, and restarting services, SMF tracks state
transitions through through two states: `state` for the
current state, and `nstate` for the next state. Because of this,
Puppet would occasionally encounter race conditions when attempting
to act on a service that was in the middle of a state transition.

This is due to the asychronous nature of service management in Solaris,

This commit adds a 60 second timeout when stopping or starting services
to ensure that the state transition has completed before continuing
operatings.

### DIFF
--- a/lib/puppet/provider/service/smf.rb
+++ b/lib/puppet/provider/service/smf.rb
@@ -1,3 +1,5 @@
+require 'timeout'
+
 # Solaris 10 SMF-style services.
 Puppet::Type.type(:service).provide :smf, :parent => :base do
   desc <<-EOT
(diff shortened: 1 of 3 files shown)


whopper:
Retargeted version of https://github.com/puppetlabs/puppet/pull/4412

peterhuene:
So despite `svcadm` having a dedicated option to imply a synchronous action, it still returns before the service enters the desired state?  I'm just ಠ_ಠ

lib/puppet/provider/service/smf.rb
```diff
@@ -74,6 +76,47 @@ def startcmd
     end
   end
 
+  # Wait for the service to transition into the specified state before returning.
+  # This is necessary due to the asynchronous nature of SMF services.
+  # desired_state should be online, offline, disabled, or uninitialized.
+  # See PUP-5474 for long-term solution to this issue.
+  def wait(desired_state)
```
peterhuene: Could this just be `*desired_state`?  Would mean we don't need to flatten or convert to an array and the one call that does need multiple states can just do `wait('offline', 'disabled', 'uninitialized')` which is a little cleaner looking imo.

whopper:
@peterhuene I actually just found https://blogs.oracle.com/SolarisSMF/entry/synchronous_actions which says synchronous actions for `svcadm` became available in Solaris 11.2. However, all of our VMs are 10.10 or 11.1, so we get the following:

Stopping:
```
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
disabled       -
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
offline        online
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
offline        online
...
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
online         -

Starting:
```
...
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
offline        online
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
online         -
```

Restarting:
```
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
online         -
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
offline        online
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
offline        online
root@wqgnn154xvgygoi:~# svcs -H -o state,nstate puppet
online         -
```

whopper:
@peterhuene, Branan is taking this PR over.

peterhuene:
Ah, didn't realize they shipped a service administration tool that didn't initially have a way of waiting on the action.  That's double ಠ_ಠ

branan:
@peterhuene Updated to use the splat syntax

peterhuene:
:+1: will merge after CI is green.

whopper:
Appveyor seems to be hanging again... :(

PR #4418, opened 2015-11-06
(PUP-5478) Fix tests failing in Fedora 22

whopper:
This commit updates two acceptance tests which were failing in Fedora
22: service_enable_linux.rb and
common_package_name_in_different_providers.rb.

The systemd provider is already the default for all versions of Fedora,
so there is no need to bump the provider.

For the common package name test, we need to use puppet-agent's private
gem binary, as rubygems is not installed on our Fedora 22 testing VMs.

### DIFF
--- a/acceptance/tests/resource/package/common_package_name_in_different_providers.rb
+++ b/acceptance/tests/resource/package/common_package_name_in_different_providers.rb
@@ -5,6 +5,8 @@ confine :except, :platform => /centos-4|el-4/ # PUP-5227
 
 require 'puppet/acceptance/rpm_util'
 extend Puppet::Acceptance::RpmUtils
+require 'puppet/acceptance/common_utils'
+extend Puppet::Acceptance::CommandUtils
 
 rpm_options = {:pkg => 'guid', :version => '1.0'}
 
@@ -30,6 +32,14 @@ step "Verify gem and ruby-devel on fedora-22 and above if not aio" do
   end
 end
 
+def gem_provider
+  if @options[:type] == 'aio'
+    'puppet_gem'
+  else
+    'gem'
+  end
+end
+
 def verify_state(hosts, pkg, state, match)
   hosts.each do |agent|
     cmd = rpm_provider(agent)
@@ -38,7 +48,7 @@ def verify_state(hosts, pkg, state, match)
       method(match).call(/^#{pkg}\./, stdout)
     end
 
-    on agent, 'gem list --local' do
+    on agent, "#{gem_command(agent, @options[:type])} list --local" do
       method(match).call(/^#{pkg} /, stdout)
     end
   end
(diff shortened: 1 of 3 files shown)


whopper:
Tested with `ci:test:aio` on F22, F21 and CentOS 7, as well as `ci:test:git` on F22 and CentOS 7.

acceptance/tests/resource/package/common_package_name_in_different_providers.rb
```diff
@@ -30,6 +32,17 @@
   end
 end
 
+# This is necessary so that the puppet-agent gem binary is in the path
+# during `puppet apply` runs. Otherwise, the gem provider is considered
+# unsuitable.
+def path
+  if @options[:type] == 'aio'
+    "#{agents[0]['privatebindir']}:${PATH}"
```
joshcooper: This won't work (reliably) if the master has the agent role, and you're testing a windows agent.

acceptance/tests/resource/package/common_package_name_in_different_providers.rb
```diff
@@ -38,7 +51,13 @@ def verify_state(hosts, pkg, state, match)
       method(match).call(/^#{pkg}\./, stdout)
     end
 
-    on agent, 'gem list --local' do
+    if @options[:type] == 'aio'
+      gem_cmd = gem_command(agent)
+    else
+      gem_cmd = on(agent, 'which gem').stdout.chomp
+    end
```
joshcooper: Can we push this logic into CommandUtils, similar to what we do for ruby?

whopper:
@joshcooper this test required a bit of work in order to correctly deal with paths for each agent. I refactored the manifest applications on `agents` to instead work in a loop on all agents like other tests. 

This test is confined to `(?:centos|el-|fedora)` agents, so I verified it passes with `ci:test:aio` on F22, F21, Centos 5, 6 and 7, and RedHat 6 and 7. I also verified `ci:test:git` with F22, F21, CentOS 7 and RedHat 7.

acceptance/tests/resource/package/common_package_name_in_different_providers.rb
```diff
@@ -30,26 +30,35 @@
...
+def verify_absent(host, pkg)
+  verify_state(host, pkg, '(?:purged|absent)', :assert_no_match)
+end
+
+def agent_path(host)
+  # This is necessary so that the puppet-agent gem binary is in the path
+  # during `puppet apply` runs. Otherwise, the gem provider is considered
+  # unsuitable.
+  if @options[:type] == 'aio'
+    "#{agent['privatebindir']}:${PATH}"
```
whopper: agh, missed this agent! Updating...

acceptance/tests/resource/package/common_package_name_in_different_providers.rb
```diff
@@ -30,26 +30,35 @@
...
+
+def agent_path(host)
+  # This is necessary so that the puppet-agent gem binary is in the path
+  # during `puppet apply` runs. Otherwise, the gem provider is considered
+  # unsuitable.
+  if @options[:type] == 'aio'
+    "#{host['privatebindir']}:${PATH}"
+  else
+    '${PATH}'
+  end
```
joshcooper: If we want to make sure we use puppet's gem command so we can install a gem in a manifest, we should just use the `puppet_gem` provider, e.g.:

```
package {'gem-guid':
  provider => puppet_gem,
  name => 'guid',
  ensure => installed,
}
```

That way we don't have to monkey with pathing.

whopper: @joshcooper Doh! Well that makes things easier...

whopper:
@joshcooper updated. I'm not sure if `gem_provider` warrants its own helper method. Verified `ci:test:gem` is still working with the change to `gem_command`.

joshcooper:
looks good, testing now

joshcooper:
not sure what's up with appveyor, but not going to wait

PR #4424, opened 2015-11-09
(PUP-5356) check_puppet.rb does not work on CentOS 7

darues:
(PUP-1234) The extension check_puppet.rb for NRPE does not work on CentOS 7

The check_puppet.rb extension to check Puppet is running with NRPE does not work on CentOS 7 (probably on all EL7), there is a problem with the used syntax in Ruby.


HAIL9000:
Hey @darues, thanks for your contribution! Before this can be reviewed and merged, the commit message needs to be in the correct formed (as described by our [contributing guidelines](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md)). If you have a chance could you please do a quick rebase and update your commit appropriately?

Thanks!

ahpook:
:+1: on this, it looks like no functional changes, just moving the syntax away from two Ruby idioms that went away in >=1.9.

HAIL9000:
Closing this one in favor of #4477 (which is the rebased version of these commits).

PR #4425, opened 2015-11-09
Flag that a type could not be found

nbarrientos:

again unnecessarily.


nbarrientos:
Will follow CONTRIBUTING.md :)

PR #4429, opened 2015-11-09
(maint) Skip puppet/mco service test on AIX

whopper:
This commit skips the puppet/mco service test on AIX
due to PA-67. Once Mcollective's `cmdargs` have been
updated to not daemonize in AIX, the test should be
unskipped.


joshcooper:
If the fix for PA-67 is just https://github.com/whopper/puppet-agent/commit/8675048f351511d39ace838d35430bbce42c58a9#diff-6deca9245796453a3f9796217a765684R1, how about fixing the issue in puppet-agent 1.3.0, and closing this PR?

whopper:
@joshcooper sounds good to me - @stahnma had marked it for PA 1.3.1, so wasn't sure if it was something RE had time to look at review in time.

PR #4432, opened 2015-11-09
(PUP-5480) Windows dirs should inherit SYSTEM

Iristyle:
- Previously, as part of b578ed4,
   when Puppet would manage a directory where SYSTEM was not explicitly
   set, it would re-add SYSTEM to the directory in an effort to ensure
   that Windows (and services like Puppet itself) could later access the
   directory:

   `NT AUTHORITY\SYSTEM:(F)`

   However, this approach was flawed, since files could be written to
   that directory (by Puppet or otherwise), and they would no longer be
   accessible to SYSTEM since inheritance was not set on the directory.
   Instead, the permissions should be set as:

   `NT AUTHORITY\SYSTEM:(F)(CI)(OI)`

   This allows the SYSTEM permission to propagate to any newly created
   files or directories within the parent managed directory.

 - As it turns out, Puppet itself is affected by this flaw on the first
   run that it writes an agents private key, public key and the ca cert
   when the agent is not run as SYSTEM (which is the case during
   acceptance tests).  On the next Puppet agent run, file modes are
   explicitly set via:
   https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L709-L759

   With debug logging on, change notifications are emitted when the
   settings catalog is applied, demonstrating the re-addition of the missing SYSTEM permissions like:

```
   [Debug: /File[C:/ProgramData/PuppetLabs/puppet/etc/ssl/private_keys/lmugyz1yds193au.delivery.puppetlabs.net.pem]/mode: mode changed '4000640' to '0640'
   [Debug: /File[C:/ProgramData/PuppetLabs/puppet/etc/ssl/public_keys/lmugyz1yds193au.delivery.puppetlabs.net.pem]/mode: mode changed '4000640' to '0644'
   [Debug: /File[C:/ProgramData/PuppetLabs/puppet/etc/ssl/certs/ca.pem]/mode: mode changed '4000640' to '0644'
```
   Also note, that the cache/client_data/catalog dir (client_datadir)
   exhibits the same behavior.

 - This commit ensures that SYSTEM is always added as inheritable to directories.

### DIFF
--- a/lib/puppet/util/windows/security.rb
+++ b/lib/puppet/util/windows/security.rb
@@ -372,7 +372,10 @@ module Puppet::Util::Windows::Security
     dacl.allow(well_known_nobody_sid, nobody_allow)
 
     # TODO: system should be first?
-    dacl.allow(well_known_system_sid, system_allow)
+    flags = !isdir ? 0 :
+      Puppet::Util::Windows::AccessControlEntry::CONTAINER_INHERIT_ACE |
+      Puppet::Util::Windows::AccessControlEntry::OBJECT_INHERIT_ACE
+    dacl.allow(well_known_system_sid, system_allow, flags)
 
     # add inherit-only aces for child dirs and files that are created within the dir
     inherit_only = Puppet::Util::Windows::AccessControlEntry::INHERIT_ONLY_ACE
--- a/spec/integration/util/windows/security_spec.rb
+++ b/spec/integration/util/windows/security_spec.rb
@@ -237,6 +237,24 @@ describe "Puppet::Util::Windows::Security", :if => Puppet.features.microsoft_win
             system_aces.each do |ace|
               ace.mask.should == klass::FILE_ALL_ACCESS && ! ace.inherited?
             end
+
+            if Puppet::FileSystem.directory?(path)
+              system_aces.each do |ace|
+                ace.object_inherit?.should be_true
+                ace.container_inherit?.should be_true
+              end
+
+              # it's critically important that this file be default created
+              # and that this file not have it's owner / group / mode set by winsec
+              nested_file = File.join(path, 'nested_file')
+              File.new(nested_file, 'w').close
+
+              system_aces = winsec.get_aces_for_path_by_sid(nested_file, sids[:system])
+              # even when SYSTEM is the owner (in CI), there should be an inherited SYSTEM
+              system_aces.any? do |ace|
+                ace.mask == klass::FILE_ALL_ACCESS && ace.inherited?
+              end.should be_true
+            end
           end
 
           describe "for modes that require deny aces" do


Iristyle:
Verified as `SYSTEM` with

```
C:\source\puppet>whoami
nt authority\system

C:\Windows\system32>ruby --version
ruby 2.0.0p481 (2014-05-08) [x64-mingw32]

C:\Windows\system32>cd \source\puppet

C:\source\puppet>bundle exec rspec .\spec\integration\util\windows\security_spec.rb
..................................................**........................................................................**.........................

Pending:
  Puppet::Util::Windows::Security file behaves like a securable object on a volume that supports ACLs for an administrator when the sid is NULL should retrieve an empty owner sid
    # Not yet implemented
    # ./spec/integration/util/windows/security_spec.rb:470
  Puppet::Util::Windows::Security file behaves like a securable object on a volume that supports ACLs for an administrator when the sid is NULL should retrieve an empty group sid
    # Not yet implemented
    # ./spec/integration/util/windows/security_spec.rb:471
  Puppet::Util::Windows::Security directory behaves like a securable object on a volume that supports ACLs for an administrator when the sid is NULL should retrieve an empty owner sid
    # Not yet implemented
    # ./spec/integration/util/windows/security_spec.rb:470
  Puppet::Util::Windows::Security directory behaves like a securable object on a volume that supports ACLs for an administrator when the sid is NULL should retrieve an empty group sid
    # Not yet implemented
    # ./spec/integration/util/windows/security_spec.rb:471

Finished in 3.16 seconds
151 examples, 0 failures, 4 pending
```

Iristyle:
@joshcooper / @ferventcoder - ready to :guitar:

joshcooper:
Running some tests locally. Not sure what's up with appveyor?

lib/puppet/util/windows/security.rb
```diff
@@ -372,7 +372,10 @@ def set_mode(mode, path, protected = true)
     dacl.allow(well_known_nobody_sid, nobody_allow)
 
     # TODO: system should be first?
-    dacl.allow(well_known_system_sid, system_allow)
+    flags = !isdir ? 0 :
+      Puppet::Util::Windows::AccessControlEntry::CONTAINER_INHERIT_ACE |
+      Puppet::Util::Windows::AccessControlEntry::OBJECT_INHERIT_ACE
+    dacl.allow(well_known_system_sid, system_allow, flags)
```
ferventcoder: So we are adding object/container inheritance for any directories created?

ferventcoder: Are there places where we don't want to allow this inheritance? Or do we believe it's fine to always allow `System` access. If so... :+1:.

Iristyle:
As noted in the ticket, a simple manifest to verify the behavior is

```puppet
file { 'c:/windows/temp/foo_dir':
  ensure => 'directory',
  group => 'None',
  mode => '600'
} ~>
 
exec { 'show_dir_perms':
  logoutput => true,
  command => 'c:\\windows\\system32\\cmd.exe /c icacls c:\\windows\\temp\\foo_dir'
} ~>
 
file { 'c:/windows/temp/foo_dir/bar.txt':
  ensure => 'present'
} ~>
 
exec { 'show_file_perms':
  logoutput => true,
  command => 'c:\\windows\\system32\\cmd.exe /c icacls c:\\windows\\temp\\foo_dir\\bar.txt'
}
```

In the example, set `foo_dir` group / owner / mode to anything that prevents `SYSTEM` from being explicitly set on the directory, prior to writing the `bar.txt` file.

lib/puppet/util/windows/security.rb
joshcooper: If we don't allow inheritance, and the child (file/dir) is not something puppet is managing permissions on, `Dir.mkdir('/path/to/client_data/catalog')`, then this ACE won't apply to the child, and System can't write the cached catalog. Similarly, we don't manage the mode on the agent cert itself, e.g. `$ssldir\certs\foo.pem`, and System can't read the cert.

Also System is not normally the owner or group (that's only the case when you're running as System, e.g. as a service), then the CreatorOwner and CreatorGroup ACEs don't apply.

joshcooper:
I ran specs on 2008r2, not going to wait for appveyor

MikaelSmith:
Not sure why it's not running, should be.

joshcooper:
The most recent appveyor webhook for this PR successfully POST'ed in https://github.com/puppetlabs/puppet/settings/hooks/4573918#details...

MikaelSmith:
:+1:

kylog:
I think it may be because the commit message includes the text "acceptance test". Maybe we should reconsider https://github.com/puppetlabs/puppet/blob/3.x/appveyor.yml#L2-L3?

ferventcoder:
@kylog perhaps?

kylog:
@ferventcoder see #4441.

PR #4433, opened 2015-11-09
(PUP-5123) Replace http:// links with https://

HAIL9000:
This replaces most http:// links with https:// The
replaced ones have all been verified to serve
identical content.  Those which aren't replaced are
contained within XML schema or are internal puppetlabs
links which are unable to be verified externally. Some
are also outdated and inaccessible and should be updated
seperately.


HAIL9000:
I opened a ticket to track the rest of the link updates which are mentioned on the previous pull request:

https://tickets.puppetlabs.com/browse/PUP-5485

PR #4434, opened 2015-11-10
(PUP-5483) Update diff regex in filebucket test for AIX

whopper:
In AIX, the output of diff varies from other platforms. This was
causing our filebucket diff test to fail due to a strict regex.
This commit updates the regexes to allow diff formats from AIX
5.3, 6.1 and 7.1.

[skip ci]


whopper:
Verified with acceptance runs against RedHat 7 and AIX 6.1. Verifying against AIX 5.3 now.

johnduarte:
:+1: validated on redhat, AIX 6.1 and 5.3

PR #4436, opened 2015-11-10
(PUP-3149) Add uninstall method to zypper provider

abednarik:
Added zypper purge to be able to remove packages using zypper instead of rpm, which in some cases could break systems due to lack of dependencies check.

### DIFF
--- a/lib/puppet/provider/package/zypper.rb
+++ b/lib/puppet/provider/package/zypper.rb
@@ -123,4 +123,23 @@ Puppet::Type.type(:package).provide :zypper, :parent => :rpm do
     # zypper install can be used for update, too
     self.install
   end
+
+  def uninstall
+    #extract version numbers and convert to integers
+    major, minor, patch = zypper_version.scan(/\d+/).map{ |x| x.to_i }
+
+    if major < 1
+      super
+    else
+      options = [:remove, '--no-confirm']
+      if major == 1 && minor < 6
+          options << '--force-resolution'
+      end
+
+      options << @resource[:name]
+    	
+      zypper *options
+    end
+
+  end
 end
--- a/spec/unit/provider/package/zypper_spec.rb
+++ b/spec/unit/provider/package/zypper_spec.rb
@@ -214,4 +214,18 @@ describe provider_class do
       @provider.install
     end
   end
+
+  describe 'when uninstalling' do
+    it 'should use remove to uninstall on zypper version 1.6 and above' do
+      @provider.stubs(:zypper_version).returns '1.6.308'
+      @provider.expects(:zypper).with(:remove, '--no-confirm', 'mypackage')
+      @provider.uninstall
+    end
+
+    it 'should use remove  --force-solution to uninstall on zypper versions between 1.0 and 1.6' do
+      @provider.stubs(:zypper_version).returns '1.0.2'
+      @provider.expects(:zypper).with(:remove, '--no-confirm', '--force-resolution', 'mypackage')
+      @provider.uninstall
+    end
+  end
 end


whopper:
Hey @abednarik, thanks for the contribution! Before we discuss this change, could you please create a new ticket at https://tickets.puppetlabs.com/browse/PUP describing the method and the need for it. Once it exists, please amend your commit message to include the ticket number in parentheses at the start I.e, '(PUP-XYZ) Add purge method to zypper provider'

abednarik:
Hi @whopper. Thanks for taking the time to review my PR. Hope now everything is just fine.
Cheers.

whopper:
@abednarik excellent, thanks! Most of us are out until Monday, and I'll be spending more time to review this then. 

In the meantime, I'm trying to figure out if zypper's `remove` semantically matches what we consider `purge` to mean. In yum and apt, we have `yum erase` and `apt-get remove --purge`, which are different than the lesser `yum uninstall` and `apt-get remove` in whether configuration files are left behind or not. 

The other option would be to make this the `uninstall` method for zypper instead of purge, if that more closely maps to what `zypper remove` is doing under the hood (currently, the zypper provider is inheriting both the `uninstall` and `purge` methods from the RPM provider.)

I haven't had time to dig into the differences much yet, but you may already know significantly more than I do about it - in which case, feel free to correct me if I'm wrong on anything :).

One last point - whether this ends up being the `purge` or `uninstall` method, we'll want to get some updated tests in `spec/unit/provider/package/zypper_spec.rb`. We'll want to add some coverage around uninstalling, similar to how our yum provider tests do it. Depending on which route this PR goes, that could look like:

```
  describe 'when uninstalling' do
    it 'should use remove to purge' do
      Puppet::Util::Execution.expects(:execute).with(['/usr/bin/zypper', '--non-interactive', :remove, name])
      provider.purge
    end
  end
```

... or ...

```
  describe 'when uninstalling' do
    it 'should use remove to uninstall' do
      Puppet::Util::Execution.expects(:execute).with(['/usr/bin/zypper', '--non-interactive', :remove, name])
      provider.uninstall
    end
  end
```

lib/puppet/provider/package/zypper.rb
```diff
@@ -123,4 +123,8 @@ def update
     # zypper install can be used for update, too
     self.install
   end
+
+  def purge
+      execute([command(:cmd), "--non-interactive", :remove, @resource[:name]])
```
whopper: One last nitpick: your indentation is four spaces here when it should be two :)

abednarik:
Many thanks!

Using just remove is similar to apt-get remove or yum uninstall. If we want to also remove any dependency no longer needed we need to append --clend-deps
I will fix the incorrect identation and a test. Finally i will need to squash all commits I think and probable rebase against the master.

whopper:
Makes sense - so it sounds like we might want to add a `uninstall` method which does what your current code does, and a `purge` method that does `zypper --non-interactive --clean-deps <package>`.

abednarik:
Ok, will try to do that.

Thanks.

abednarik:
This is what i having at the end

````puppet
  def uninstall
    execute([command(:zypper), "--non-interactive", :remove, @resource[:name]])
  end

  def purge
    execute([command(:zypper), "--non-interactive --clean-deps", :remove, @resource[:name]])
  end
```

I can't make tests to  work

With something like this

```puppet

  describe 'when uninstalling' do
    it 'should use remove to uninstall' do
      Puppet::Util::Execution.expects(:execute).with(['/usr/bin/zypper', '--non-interactive', :remove, 'mypackage'])
    end
  end

```

I get this

```puppet
 bundle exec rake spec TEST=spec/unit/provider/package/zypper_spec.rb
rspec spec/unit/provider/package/zypper_spec.rb
......................F

Failures:

  1) Puppet::Type::Package::ProviderZypper when uninstalling should use remove to uninstall
     Failure/Error: Puppet::Util::Execution.expects(:execute).with(['/usr/bin/zypper', '--non-interactive', :remove, 'mypackage'])
     Mocha::ExpectationError:
       not all expectations were satisfied
       unsatisfied expectations:
       - expected exactly once, not yet invoked: Puppet::Util::Execution.execute(['/usr/bin/zypper', '--non-interactive', :remove, 'mypackage'])
       satisfied expectations:
       - allowed any number of times, not yet invoked: #<Mock:resource>.command(:zypper)
       - allowed any number of times, not yet invoked: #<Mock:resource>.[](:ensure)
       - allowed any number of times, not yet invoked: #<Mock:resource>.[](:name)
       - allowed any number of times, not yet invoked: #<Mock:resource>.[](any_parameters)
     # ./spec/unit/provider/package/zypper_spec.rb:225:in `block (3 levels) in <top (required)>'

Finished in 0.09244 seconds (files took 1.13 seconds to load)
23 examples, 1 failure

Failed examples:

rspec ./spec/unit/provider/package/zypper_spec.rb:224 # Puppet::Type::Package::ProviderZypper when uninstalling should use remove to uninstall

rake aborted!
Command failed with status (1): [rspec spec/unit/provider/package/zypper_sp...]
/Users/abednarik/work/code/puppet/puppet/Rakefile:67:in `block in <top (required)>'
Tasks: TOP => spec
(See full trace by running task with --trace)
```

whopper:
Hey @abednarik, thanks for working on those tests! No worries - I'm off until Monday but I'd be glad to take a look and help you figure out what's going wrong then. I'll let you know when I've poked around!

abednarik:
Hey @whopper. Thanks :) It's holidays here, so i will be working to fix this.

whopper:
@abednarik alright, I've had some time to look at this! A few things:

First, I believe I was wrong about needing both a purge and uninstall option. The `--clean-deps` option is not available in older versions of zypper (such as the case in SLES 10), and I believe just having the `uninstall` method is enough to fix this particular issue anyway, if that makes sense to you.

Secondly, looking at the rest of the provider, I realized that the `uninstall` method should be consistent with how `install` operates. It should look something like this (which will make it easier to test):

```
    def uninstall
      options = ['--non-interactive', :remove, @resource[:name]]
      zypper *options
    end
```

As for the test, something like this should work if you update the method as I did above:

```
  describe 'when uninstalling' do
    it 'should use remove to uninstall' do
      @provider.expects(:zypper).with('--non-interactive', :remove, 'mypackage')
      @provider.uninstall
    end
  end
```

(make sure to call the method with `@provider.uninstall`, which your previous example was missing.)

One last question: do you know how long the `--non-interactive` option has been available in yum? Every version I've tested supports it (SLES 10 -12, OpenSuSE 11), so I *think* it's safe, but if there are versions that don't support it we might look into the `--no-confirm` option, as the install method uses.

abednarik:
HI @whopper. Many thanks for help me up with this. Looks like using --no-confirm sounds better, since is compatible with all zypper version and also is consistent with install method.
 Just updated PR, to make it clean, squashing previous commit and doing a rebase.

Hope things are better now.

I really appreciate your collaboration :)

lib/puppet/provider/package/zypper.rb
```diff
@@ -123,4 +123,9 @@ def update
     # zypper install can be used for update, too
     self.install
   end
+
+  def uninstall
+    options = ['--no-confirm', :remove, @resource[:name]]
```
whopper: @abednarik no problem! One last thing. With the options specified like this, Puppet calls zypper with:

```
Debug: Executing: '/usr/bin/zypper --no-confirm remove emacs'
Error: Execution of '/usr/bin/zypper --no-confirm remove emacs' returned 2: Unknown option '--no-confirm'
```
`--no-confirm` looks to be a subcommand of `remove`, so we need to be calling `zypper remove --no-confirm emacs`.
If you move the '--no-confirm' to be the second element of the options array, it'll work correctly:

```
options = [:remove, '--no-confirm', @resource[:name]]
```

You'll also need to change the order in the spec test.

abednarik:
Done.

whopper:
@abednarik excellent - thanks a bunch, this looks good! We'll be having our weekly PR triage meeting tomorrow morning, and I'll bring this up for review and hopefully a merge :).

abednarik:
Thanks for your patience and dedication on this.

whopper:
Oh there is one other very nit-picky thing :)! Could you change the commit title to say "Add uninstall method to zypper provider" rather than purge?

And no problem! Thanks for contributing to puppet core!

abednarik:
Sure thing. done :)

whopper:
Hey @abednarik! We went over this PR today in our PR triage, and everything looks good except one issue with how old versions of zypper (present on SLES 10 and OpenSUSE 11) do dependency management. Using zypper 1.0.2, if I try to uninstall something like emacs which has dependencies, I get:

```
dinxjb1qymsvwyl:~ # zypper remove --no-confirm emacs
Loading repository data...
Reading installed packages...
Resolving package dependencies...

Problem: emacs-x11-22.3-4.6.x86_64 requires emacs = 22.3-4.6, but this requirement cannot be provided
 Solution 1: deinstallation of emacs-x11-22.3-4.6.x86_64
 Solution 2: do not ask to delete all solvables providing emacs
 Solution 3: Ignore some dependencies of emacs-x11

Choose from above solutions by number or cancel [1/2/3/C]: c
```

`--no-confirm` just cancels out and returns 0, so Puppet will think that it's uninstalled the package when in reality it did not. I've run into this in two versions of Zypper: 1.0.2 and 0.6.201. Version 1.0.2 has the ability to find a solution automatically, while 0.6 does not.

```
dinxjb1qymsvwyl:~ # zypper --version
zypper 1.0.2
dinxjb1qymsvwyl:~ # zypper help remove
...
  Command options:
-R, --no-force-resolution       Do not force the solver to find solution, let it ask.
    --force-resolution          Force the solver to find a solution (even an agressive).
```

We discussed this and decided that it'd be best to check for zypper version and do the right thing based on what we find. If the version is >= 1.6, we'll use your command. If it's less than 1.6 but >= 1.0, we can  simply add the `--force-resolution` option to the `remove` sub-command (`zypper remove --force-resolution --no-confirm <package>` does the right thing). Finally, if the version is less than 1.0, we should call `super` to fall back to the RPM provider, which will error as it does now rather than "succeeding" but silently failing. I think this is probably the right thing to do, since very old versions of zypper don't have a way to automatically determine dependency solutions, so the provider shouldn't lie about success.

The zypper version can be found in the same way it's done in the `install` method. The following is a proof of concept:

```
def uninstall
  #extract version numbers and convert to integers
  major, minor, patch = zypper_version.scan(/\d+/).map{ |x| x.to_i }

  if major < 1
    super
  else
    options = [:remove, '--no-confirm']
    if major == 1 && minor < 6
        options << '--force-resolution'
    end

    options << @resource[:name]
  end

  zypper *options
end
```

The tests would look something like this:

```
  describe 'when uninstalling' do
    it 'should use remove to uninstall on zypper version 1.6 and above' do
      @provider.stubs(:zypper_version).returns '1.6.308'
      @provider.expects(:zypper).with(:remove, '--no-confirm', 'mypackage')
      @provider.uninstall
    end

    it 'should use remove  --force-solution to uninstall on zypper versions between 1.0 and 1.6' do
      @provider.stubs(:zypper_version).returns '1.0.2'
      @provider.expects(:zypper).with(:remove, '--no-confirm', '--force-resolution', 'mypackage')
      @provider.uninstall
    end
  end
```

Sorry about all the changes! Let me know if you could use any more help, and thanks again - we really appreciate you running through all of this!

abednarik:
Hey, don't worry. I will have a look, should be easy since you already made near everything :)

abednarik:
HI @whopper  let see if this commit is just fine and working.

whopper:
:+1:

lib/puppet/provider/package/zypper.rb
```diff
@@ -123,4 +123,22 @@ def update
...
+    else
+      options = [:remove, '--no-confirm']
+      if major == 1 && minor < 6
+          options << '--force-resolution'
+      end
+
+      options << @resource[:name]
+    end
+
+    zypper *options
```
whopper: @abednarik doh! One last thing (my fault!) This `zypper` call should go in the `else` clause after `@resource[:name]` is added to `options`, otherwise it could be called even if we call `super`.

After that change, we're ready to merge :)

abednarik:
You are absolutely right. It was my fault :)

@whopper  hope is fine now.

whopper:
Thanks @abednarik - merged and going through internal CI!

abednarik:
Great,  thanks you.

PR #4437, opened 2015-11-10
(PUP-5487) Add ability to supress warnings about storeconfig

jimmybergman:
This patch adds the ability to disable warnings about storeconfigs
not beeing enabled with:
--disable_warnings storeconfigs


hlindberg:
As per the discussion in the ticket - I think this should be done a different (and much better way) by only issuing warnings when the expressions in question are evaluated/used. See the ticket for how that is easily implemented.

The way this is changed in this ticket means that it would stop issuing warnings at evaluation time as well  as when validating. It would fall back to the lowest level warning (deep inside) that gives no information about where the problem is in the manifests.

So, I am closing this PR.

PR #4439, opened 2015-11-10
(maint) Update links.

gguillotte:
And where necessary, also update the surrounding content.

Re: PUP-5485


HAIL9000:
@gguillotte Thanks so much for doing this! Would you mind updating the commit to reference PUP-5485

HAIL9000:
Oh just kidding, it's referenced in the commit body. Thanks! :100:

PR #4442, opened 2015-11-10
(maint) Update lookup help to specify running on master

HAIL9000:
Prior to this commit, the --help message for lookup did not make it
clear that lookup needs to be run on a puppet master or a masterless
node. Update the docs to reflect this fact so the workflow is more
clear to users.


HAIL9000:
Whoops, accidentally targeted master. Closing this to open a new pull request.

PR #4444, opened 2015-11-12
(PUP-5390) Audit acceptance tests for old skips and confines

whopper:
These commits update several acceptance tests to either un-skip for certain platforms or add better references / notes about why they're skipped.

### DIFF
--- a/acceptance/tests/config/puppet_manages_own_configuration_in_robust_manner.rb
+++ b/acceptance/tests/config/puppet_manages_own_configuration_in_robust_manner.rb
@@ -5,8 +5,6 @@
 # expect that after correcting their actions, puppet will work correctly.
 test_name "Puppet manages its own configuration in a robust manner"
 
-confine :except, :platform => 'fedora-19'
-
 skip_test "JVM Puppet cannot change its user while running." if @options[:is_puppetserver]
 
 # when owner/group works on windows for settings, this confine should be removed.
--- a/acceptance/tests/modules/changes/module_with_modified_file.rb
+++ b/acceptance/tests/modules/changes/module_with_modified_file.rb
@@ -1,9 +1,5 @@
 test_name 'puppet module changes (on a module with a modified file)'
 
-hosts.each do |host|
-  skip_test "skip tests requiring forge certs on solaris and aix" if host['platform'] =~ /solaris/
-end
-
 step 'Setup'
 
 stub_forge_on(master)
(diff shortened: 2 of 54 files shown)


johnduarte:
lgtm :+1:  :100: 

thanks @whopper !!!

acceptance/tests/config/puppet_manages_own_configuration_in_robust_manner.rb
```diff
@@ -5,8 +5,6 @@
 # expect that after correcting their actions, puppet will work correctly.
 test_name "Puppet manages its own configuration in a robust manner"
 
-confine :except, :platform => 'fedora-19'
```
er0ck: why?  is there a ticket for this?
should we use skip_test with a comment, where possible?
should we use expect_failure when we expect one of the asserts to throw an exception?

acceptance/tests/modules/changes/module_with_modified_file.rb
```diff
@@ -1,9 +1,5 @@
 test_name 'puppet module changes (on a module with a modified file)'
 
-hosts.each do |host|
-  skip_test "skip tests requiring forge certs on solaris and aix" if host['platform'] =~ /solaris/
```
er0ck: should we remove the reference to aix in this output?

er0ck:
LGTM also.  please ignore my previous comments.  i'm colorblind in the am, i guess.  :tired_face:

whopper:
@er0ck hah, funny - that exact thing happened to John yesterday :)

er0ck:
maybe not.  are solaris builds daily only?

johnduarte:
@er0ck yes, solaris packages are only build in the daily run.

whopper:
@er0ck oh yeah, the invocation I used to test for all the Solarises and AIXes was:

```
SHA=8be23474284a9f8e2bae551d266f5730af6681cd SUITE_VERSION=1.2.7.405.g8be2347 CONFIG=config/nodes/solaris-11-x86_64.yaml OPTIONS='--preserve-hosts' TESTS=./tests/modules bundle exec rake ci:test:aio
```

er0ck:
@johnduarte this passes on solaris11.  should we run this against AIX before merge?

johnduarte:
@er0ck verified on aix 5.3

joshcooper:
Thinking we probably want to backport this to stable and set fixversion to PUP 4.3.1 so that we have confidence in upcoming puppet-agent 1.3.1 release. Thoughts?

johnduarte:
:+1: @joshcooper

PR #4445, opened 2015-11-13
Fix up LOAD_PATH before calling Kernel.load

njm506:
(PUP-4450) Sometimes, we're autoloading something that _require_s a library that lives in a directory which Puppet knows about, but the Ruby kernel doesn't.

I've compared spec results; this change makes no difference to them.

This change should be portable to Puppet 4; we've only tested it Puppet 3, though.

### DIFF
--- a/lib/puppet/util/autoload.rb
+++ b/lib/puppet/util/autoload.rb
@@ -63,7 +63,7 @@ class Puppet::Util::Autoload
       $clean_loadpath ||= $LOAD_PATH.dup
       $env_loadpath ||= {}
       $env_loadpath[env] ||= search_directories(env)
-      $current_loadpath ||= nil
+      $current_loadpath ||= $clean_loadpath
 
       loadpath = $env_loadpath[env]
       last_loadpath = $current_loadpath
--- a/lib/puppet/util/autoload.rb
+++ b/lib/puppet/util/autoload.rb
@@ -55,7 +55,23 @@ class Puppet::Util::Autoload
     # Load a single plugin by name.  We use 'load' here so we can reload a
     # given plugin.
     def load_file(name, env)
-      load_path = $LOAD_PATH
+      # $clean_loadpath     - remembers LOAD_PATH from where we started
+      # $env_loadpath[env]  - remembers the LOAD_PATH we want to apply for each environment
+      # $current_loadpath   - which LOAD_PATH we have set right now
+      # loadpath            - which LOAD_PATH we want to apply
+      # last_loadpath       - which LOAD_PATH was set when we started
+      $clean_loadpath ||= $LOAD_PATH.dup
+      $env_loadpath ||= {}
+      $env_loadpath[env] ||= search_directories(env)
+      $current_loadpath ||= nil
+
+      loadpath = $env_loadpath[env]
+      last_loadpath = $current_loadpath
+      if loadpath != $current_loadpath
+        $current_load_path = loadpath
+        $LOAD_PATH.replace loadpath
+      end
+
       file = get_file(name.to_s, env)
       return false unless file
       begin
@@ -70,11 +86,11 @@ class Puppet::Util::Autoload
         Puppet.log_exception(detail, message)
         raise Puppet::Error, message, detail.backtrace
       ensure
-        # Restore original LOAD_PATH
-        $LOAD_PATH.each do |p|
-          $LOAD_PATH.delete(p) unless load_path.include? p
+        # Restore previous LOAD_PATH if we changed it
+        if loadpath != last_loadpath
+          $LOAD_PATH.replace last_loadpath
+          $current_loadpath = last_loadpath
         end
-        $LOAD_PATH.uniq!
       end
     end
 
--- a/lib/puppet/util/autoload.rb
+++ b/lib/puppet/util/autoload.rb
@@ -76,7 +76,6 @@ class Puppet::Util::Autoload
       return false unless file
       begin
         mark_loaded(name, file)
-        $LOAD_PATH.push *(search_directories(env))
         Kernel.load file, @wrap
         return true
       rescue SystemExit,NoMemoryError
--- a/lib/puppet/util/autoload.rb
+++ b/lib/puppet/util/autoload.rb
@@ -55,10 +55,12 @@ class Puppet::Util::Autoload
     # Load a single plugin by name.  We use 'load' here so we can reload a
     # given plugin.
     def load_file(name, env)
+      load_path = $LOAD_PATH
       file = get_file(name.to_s, env)
       return false unless file
       begin
         mark_loaded(name, file)
+        $LOAD_PATH.push *(search_directories(env))
         Kernel.load file, @wrap
         return true
       rescue SystemExit,NoMemoryError
@@ -67,6 +69,12 @@ class Puppet::Util::Autoload
         message = "Could not autoload #{name}: #{detail}"
         Puppet.log_exception(detail, message)
         raise Puppet::Error, message, detail.backtrace
+      ensure
+        # Restore original LOAD_PATH
+        $LOAD_PATH.each do |p|
+          $LOAD_PATH.delete(p) unless load_path.include? p
+        end
+        $LOAD_PATH.uniq!
       end
     end
 
--- a/lib/puppet/util/autoload.rb
+++ b/lib/puppet/util/autoload.rb
@@ -62,7 +62,11 @@ class Puppet::Util::Autoload
       # last_loadpath       - which LOAD_PATH was set when we started
       $clean_loadpath ||= $LOAD_PATH.dup
       $env_loadpath ||= {}
-      $env_loadpath[env] ||= search_directories(env)
+      unless $env_loadpath[env]
+        $LOAD_PATH.replace $clean_loadpath
+        $current_loadpath = $clean_loadpath
+        $env_loadpath[env] = search_directories(env)
+      end
       $current_loadpath ||= $clean_loadpath
 
       loadpath = $env_loadpath[env]


lib/puppet/util/autoload.rb
```diff
@@ -55,10 +55,12 @@ def changed?(name)
     # Load a single plugin by name.  We use 'load' here so we can reload a
     # given plugin.
     def load_file(name, env)
+      load_path = $LOAD_PATH
       file = get_file(name.to_s, env)
       return false unless file
       begin
         mark_loaded(name, file)
+        $LOAD_PATH.push *(search_directories(env))
```
hlindberg: `search_directories` will include everything already on the `$LOAD_PATH` since its definition is `[gem_directories, module_directories(env), libdirs(), $LOAD_PATH].flatten`. I think the result is lots of extra entries on the path. This has performance implications as it will search through all of them.

lib/puppet/util/autoload.rb
```diff
@@ -67,6 +69,12 @@ def load_file(name, env)
         message = "Could not autoload #{name}: #{detail}"
         Puppet.log_exception(detail, message)
         raise Puppet::Error, message, detail.backtrace
+      ensure
+        # Restore original LOAD_PATH
+        $LOAD_PATH.each do |p|
+          $LOAD_PATH.delete(p) unless load_path.include? p
+        end
+        $LOAD_PATH.uniq!
```
hlindberg: The above restore will have performance impact. It should be possible to do this in a more light way fashion. The above has two linear array searches for each entry on the path + the extra uniq set operation at the end.

njm506: I agree this isn't an ideal way to do it.  I'm not entirely sure what's
already on LOAD_PATH (apart from itself) - certainly
module_directories(env) needs adding.

In terms of the restore, my preference would have been to set $LOAD_PATH
directly, but it's not a normal array (you can't directly set it).

Though it's not ideal, it doesn't seem to be causing any performance issues
on my system.  The first instance of a match short-circuits the rest of the
array.  It's the exceptional case where something doesn't exist that will
take the worst hit.

What might be a more suitable solution would be to prepare LOAD_PATH before
compilation (but only when the real client environment is known), and then
drop it back to its previous state after compilation is finished.

I'd gladly accept suggestions of how to achieve this (i.e. where this could
reasonably be hooked in) and update my PR accordingly.

On 13 November 2015 at 13:47, Henrik Lindberg <notifications@github.com>
wrote:

> In lib/puppet/util/autoload.rb
> <https://github.com/puppetlabs/puppet/pull/4445#discussion_r44783036>:
>
> > @@ -67,6 +69,12 @@ def load_file(name, env)
> >          message = "Could not autoload #{name}: #{detail}"
> >          Puppet.log_exception(detail, message)
> >          raise Puppet::Error, message, detail.backtrace
> > +      ensure
> > +        # Restore original LOAD_PATH
> > +        $LOAD_PATH.each do |p|
> > +          $LOAD_PATH.delete(p) unless load_path.include? p
> > +        end
> > +        $LOAD_PATH.uniq!
>
> The above restore will have performance impact. It should be possible to
> do this in a more light way fashion. The above has two linear array
> searches for each entry on the path + the extra uniq set operation at the
> end.
>
> —
> <https://github.com/puppetlabs/puppet/pull/4445/files#r44783036>.
>



-- 


Nick Moriarty
Linux Systems Administrator and Developer

Information Services
University of York
Heslington
York
YO10 5DD
+44 (0)1904 32 3484



e-mail disclaimer: http://www.york.ac.uk/docs/disclaimer/email.htm

njm506:
So, once a "good place" to do this is known, the following seem to be a nice way of setting $: -
clean_load_path = $LOAD_PATH.dup
$LOAD_PATH.replace(search_directories(env))
... (process the catalog) ...
$LOAD_PATH.replace(clean_load_path)

The above happily replaces the content of the array.  Note the 'dup' is necessary, otherwise old_load_path refers to the same array (and so the replace operation affects both variables).

hlindberg:
I am concerned about the performance since the change affects each single load. It may be that the change is not actually executed that often as attempted loads for non existing entries bail out after the search for a file using the expected set of paths (i.e. search_directories). Thus all "misses" are no slower than today.

hlindberg:
We probably have the same problem with loading 4.x functions.
Finding "a good place" is hard - it is related to the notion of "current environment" which unfortunately is still a quite fluid concept in the 3.x runtime. It is better in 4.x but still spread out over the logic. Additional difficulties arise since logic is shared between agent and master in certain parts.

We try to deal with "global" things in the Puppet context. It is configured to contain what was earlier in ruby global variables (i.e. Puppet.lookup(:key), Puppet.override, and related methods). A search for where `:current_environment` is used in source shows where the touch points are.

There are several ways we could approach a fix:
* Saving the "clean load path" just once and the looking it up should save cycles as the "clean path" does then not have to be duplicated for each and every load. This "save" could be done where `:current_environment` is set (although it may be in quite a few places).
* The load_path could perhaps also be set at the same places.
* Restoring the load_path becomes a bit more difficult as the current Puppet context logic does not have restore logic that can reset the global state of Ruby (i.e. there is logic that just drops pushed contexts)
* A middle ground could perhaps be to both save the "clean path", and compute "effective_path". Then the logic in the autoloader simple does two replace operations - first to the "effective", and then back to the "clean". That would save the cycles needed to compute it over and over again. (i.e. only call search_directories(env) once) and caching it in the puppet context).

Longer term, the messy fluid "current environment" concept should be cleaned up along with the loading logic - using the new loaders used by the 4.x function API. But that is a different story. Now we have a problem with a regression. I have seen work arounds that use relative require to load dependencies (not pretty).

hlindberg:
An additional concern when dealing with loadpath and global state is that there may be issues related to running in JRuby on Puppet Server as well - but I think it should be fine using the discussed approach - but it may require change to logic in puppet server as well (since it also IIRC has logic that sets up the puppet context).

njm506:
That should do it.

A more elegant solution could be the addition of a helper method to replace 'require' in module plugin code, which searches the appropriate locations rather than relying directly on LOAD_PATH.  That would require a lot of people to change a lot of code, though.

lib/puppet/util/autoload.rb
```diff
@@ -55,6 +55,27 @@ def changed?(name)
     # Load a single plugin by name.  We use 'load' here so we can reload a
     # given plugin.
     def load_file(name, env)
+      # $clean_loadpath     - remembers LOAD_PATH from where we started
+      # $env_loadpath[env]  - remembers the LOAD_PATH we want to apply for each environment
+      # $current_loadpath   - which LOAD_PATH we have set right now
+      # loadpath            - which LOAD_PATH we want to apply
+      # last_loadpath       - which LOAD_PATH was set when we started
+      $clean_loadpath ||= $LOAD_PATH.dup
+      $env_loadpath ||= {}
```
hlindberg: Sorry, but I don't think it is a good approach to add global variables for this (or for anything else for that matter). Also, since it is a global cache of env to loadpath and there is no cache eviction this will leak memory. As I suggested earlier this should go into the Puppet context (or possibly in the environment object).

hlindberg:
I don' think there is ever the case for loading anything into anything but the current environment, except possibly when there is no environment at all - in which case the original load path is sufficient (there is no module path, etc.)

Puppet source base historically suffers from unclear demarcations like "executing for environment x" (no knowing where it starts, nor when it ends), with lots of helpers and modules all attempting to be autonomous and reactive to "puppet being in any state". This has led to lots of subtle bugs and wasted performance. So, I don't think we should add more such stuff by trying to juggle the load path within the autoloader module.

njm506:
I agree, this is not a good way to fix the issue - globals are almost always the wrong answer - but it does (crudely) solve the problem.  I am not at all familiar with the Puppet code base, so I'm at a bit of loss as to how and where to do this properly.  I'd be happy to put some more effort into it, but I'll need some pointers as to which class(es) this belongs in, and which methods are likely to be involved.

What I do need is for the problem to be solved - the ticket for PUP-4450 has been open for 7 months now and is still sitting unassigned with "Needs Information".  I understand that Puppet is open-source and relies on some degree of community support, which is why I've expended some time to debug the problem and come up with a crude solution to demonstrate that it can be fixed.

HAIL9000:
Hey @njm506, thank you so much for the time you spent on this contribution. It's clear to us that a lot of people are invested in this fix, so we're going to pull PUP-4450 into one of our upcoming sprints so we will give this issue the time it needs to fix it.

I propose that we close this pull request, and pass this off to someone internal. Unless you would like to continue running with this. What do you think?

njm506:
That's great news!

I'd be more than happy for people who know the code base well to fix this
properly, rather than have me guessing.


On 30 November 2015 at 19:27, Hailee Kenney <notifications@github.com>
wrote:

> Hey @njm506 <https://github.com/njm506>, thank you so much for the time
> you spent on this contribution. It's clear to us that a lot of people are
> invested in this fix, so we're going to pull PUP-4450 into one of our
> upcoming sprints so we will give this issue the time it needs to fix it.
>
> I propose that we close this pull request, and pass this off to someone
> internal. Unless you would like to continue running with this. What do you
> think?
>
> —
> <https://github.com/puppetlabs/puppet/pull/4445#issuecomment-160732955>.
>



-- 


Nick Moriarty
Linux Systems Administrator and Developer

Information Services
University of York
Heslington
York
YO10 5DD
+44 (0)1904 32 3484



e-mail disclaimer: http://www.york.ac.uk/docs/disclaimer/email.htm

branan:
For posterity, the related ticket is https://tickets.puppetlabs.com/browse/PUP-4450

PR #4446, opened 2015-11-13
(PUP-5437) Add 'lookup_options' to the lookup explainer

thallgren:
The explainer will currently only report the actions that lookup
performs during traversal of global data binders and providers in
order to find the data for a given key. This commit adds an option
to include the lookup of the new 'lookup_options' hash to the report.

The '--explain-options' option takes no argument can be used alone or
in combination with the '--explain' option. When used alone, and
when no keys are passed to the lookup command, the result will be
the global lookup_options hash (merge of global data binding and
environment data provider). When keys are passed, the lookup_options
will be the full hash that includes a hash found in the data provider
of the module identified by the key.

### DIFF
--- a/spec/unit/data_providers/hiera_data_provider_spec.rb
+++ b/spec/unit/data_providers/hiera_data_provider_spec.rb
@@ -81,10 +81,8 @@ describe "when using a hiera data provider" do
     expect(resources[0]).to eq('A, B, C, MA, MB, MC')
   end
 
-  it "can lookup the 'lookup_options' hash as a regular value" do
-    resources = compile_and_get_notifications('hiera_misc', '$r = lookup(lookup_options, Hash[String,Hash[String,String]], hash) notify{"${r[one::lopts_test::hash][merge]}":}')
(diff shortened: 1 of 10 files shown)


spec/unit/data_providers/hiera_data_provider_spec.rb
```diff
@@ -189,7 +187,179 @@ def compile(environment, code = nil)
...
+          }
+    Merged result: {
+      "one::loptsm_test::hash" => {
+        "merge" => "deep"
+      },
+      "one::lopts_test::hash" => {
+        "merge" => "deep"
+      }
+    }
+Merge strategy deep
```
hlindberg: a "Searching for 'one::loptsm_test::hash'" here would be really helpful in understanding that it is now reporting for the actual lookup.

hlindberg:
one of the travis checks failed - I relaunched that job to see if it is a real problem or a temporary one.

hlindberg:
Much better now - only waiting for thumbs up on merging to stable.

PR #4447, opened 2015-11-13
(PUP-5505) Improve performance of reading service plists

MikaelSmith:
Reading text-based plist files with plutil appears to incur a lot of
overhead. Attempt to read all plists as text first, falling back to
plutil if that fails. Failing to read a binary plist appears to incur
very little overhead.

This appears to compare favorably to other approaches using
CFPropertyList, and is a much smaller change. Average times for `puppet
resource service` improved from 30 seconds to 5 seconds on a variety of
tested systems (OS X 10.9 with Ruby 1.9.3, OS X 10.10 with Ruby 2.0.0,
OS X 10.11 with Ruby 2.1.7).

### DIFF
--- a/lib/puppet/provider/service/launchd.rb
+++ b/lib/puppet/provider/service/launchd.rb
@@ -205,6 +205,12 @@ Puppet::Type.type(:service).provide :launchd, :parent => :base do
   # format.
   def self.read_plist(path)
     begin
+      return Plist::parse_xml(path)
+    rescue ArgumentError => detail
+      Puppet.debug("Error reading #{path}: #{detail}. Retrying with plutil.")
+    end
+
+    begin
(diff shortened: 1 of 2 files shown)


MikaelSmith:
I've tested this merged into a Puppet 4.2.3 install and running https://github.com/puppetlabs/puppet/blob/stable/acceptance/tests/resource/service/launchd_provider.rb. The test passes (and runs in 45 seconds instead of 170 seconds).

kylog:
:+1: 

Will merge after puppet-agent 1.3.2 ships. (I know this is targeted at 3.x so the merge itself wouldn't be a concern, but I have a mild preference to do the merge-ups at the same time.)

MikaelSmith:
Should be clear to do so now.

kylog:
Merged up to stable and master too.

PR #4448, opened 2015-11-13
(maint) Fixes re-running preserved hosts

MikaelSmith:
The layout of logs from beaker has changed. Update the
`test_against_preserved_hosts` Rake task to match the new layout.


MikaelSmith:
A change is needed, but not because of the reason I originally changed. The file is no longer called preserved_config.yaml.

PR #4449, opened 2015-11-15
(PE-12534) Fix yumrepo param specifying false fail

reidmv:
Previously, there was no munge logic for YUM_BOOLEAN parameters of the yumrepo type. Boolean false is a valid value for these parameters. Due to the way Puppet::Parameter::Value default validation methods work however, if a parameter's @should value is represented as Boolean false it will effectively always be considered insync - which is wrong, in the case of how yumrepo parameters work.

This commit adds common munging logic to each parameter that accepts YUM_BOOLEAN values, converting Boolean false to string false to ensure the default validation routines work correctly and property insync is correctly detected.

While working on this problem, another issue became apparent, which is that in Puppet 4 "absent" passed to YUM_BOOLEAN parameters comes back as a string. However, the provider only knows to remove the relevant line from the config file if absent is a symbol. Therefore, the munge method also converts string absent to symbol absent.

### DIFF
--- a/lib/puppet/type/yumrepo.rb
+++ b/lib/puppet/type/yumrepo.rb
@@ -21,6 +21,13 @@ Puppet::Type.newtype(:yumrepo) do
   YUM_BOOLEAN=/^(True|False|0|1|No|Yes)$/i
   YUM_BOOLEAN_DOC="Valid values are: False/0/No or True/1/Yes."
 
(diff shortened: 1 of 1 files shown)


reidmv:
Investigating the CI's request for a PUP ticket (not a PE) I found the discussion in PUP-3528, which suggests the desired fix for this is not to deal correctly with false, but to reject false (the correct specification being instead False). Closing the PR for now to adjust.

PR #4450, opened 2015-11-15
(PUP-5682) Fix yumrepo param specifying false fail

reidmv:
Previously, there was no munge logic for YUM_BOOLEAN parameters of the
yumrepo type. Boolean false parses as a valid value for these parameters
due to the way Puppet::Parameter::Value default validation methods work.
However, "false" or Boolean false is not actually valid. Only "False".

Furthermore, recieving a Boolean false should value will actually cause
the provider to incorrectly believe the property to be insync when it is
not.

This commit adds common munging logic to each parameter that accepts
YUM_BOOLEAN values, converting Boolean false to string false to ensure the
default validation routines work correctly and property insync is correctly
detected.

While working on this problem, another issue became apparent, which is that in
Puppet 4 "absent" passed to YUM_BOOLEAN parameters comes back as a string.
However, the provider only knows to remove the relevant line from the config
file if absent is a symbol. Therefore, the munge method also converts string
absent to symbol absent.


reidmv:
Although the tests are written to expect downcased true/false, discussion in PUP-3528 indicates the downcase version of those parameters is not actually expected to be valid by the type. Additionally, the YUM_BOOLEAN regex contains only True/False. Because the default regex validation is case insensitive it happens to be the case that downcased true/false have been accepted even though they are not actually in the regex list of valid values.

whopper:
Hey @reidmv, thanks for tackling this! A few things: Could you update the yumrepo tests to account for the new YUM_BOOLEAN strings? Also, we'd prefer to reference a PUP ticket here, and it looks like [PUP-5682](https://tickets.puppetlabs.com/browse/PUP-5682) would cover this change.

reidmv:
@whopper tests updated, commit message changed to reference PUP-5682.

tux-o-matic:
This might have introduced a problem when parsing INI style .repo file using a space around the equal sign.
For example:
```ini
enabled = 0
```
Since this update no longer gets parsed as a negative but and leads to:
```shell
/Stage[main]/Main/Yumrepo[rhel-6-server-extras-rpms]/enabled: enabled changed ' 0' to 'False'
```
Maybe a little trim when parsing the value might be enough. RHEL generates its repository files with spaces so failing to handle those with the Yumrepo resource is problematic.

PR #4451, opened 2015-11-16
(PUP-5487) Remove runtime warning about storeconfig not beeing set and instead only warn on evaluation time

jimmybergman:
(PUP-5487) Remove runtime warning about storeconfig not beeing set and instead only warn on evaluation time.

Also since the evaluation time warning included less information it is changed to also include information about where the collection happened.

### DIFF
--- a/lib/puppet/pops/evaluator/collectors/exported_collector.rb
+++ b/lib/puppet/pops/evaluator/collectors/exported_collector.rb
@@ -19,7 +19,7 @@ class Puppet::Pops::Evaluator::Collectors::ExportedCollector < Puppet::Pops::Eva
   # evaluate method
   def evaluate
     if Puppet[:storeconfigs] != true
-      Puppet.warning "Not collecting exported resources without storeconfigs"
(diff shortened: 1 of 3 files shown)


hlindberg:
Much better now! In order to also be able to produce a better error message (that includes file, line, pos) the check should be made in the evaluator (in `evaluator_impl.rb` - method `eval_CollectExpression`, it currently delegates the job to a transformer - a check should be made before calling the transformer.

You can see a similar construct for evaluation of exported resources where it makes this call:
```
      optionally_fail(Puppet::Pops::Issues::RT_NO_STORECONFIGS_EXPORT, o);
```
Have not looked if there is an issue code available for collection of exported - one may have to be added.

jimmybergman:
Thanks for your great feedback, it really helped figuring out how how this works. With the new commit it now works perfectly.

No warning with a program that contains collection of exported resources but doesn't evaluate these expressions without storeconfigs set.

When evaluating them the warning looks identical to the old parse time warning.

So pretty much I believe that it now behaves like everyone would want, the only difference from before is that warnings are hidden for non-evaluated expressions.

PR #4453, opened 2015-11-17
(PUP-5444) Create user and group accounts once

joshcooper:
Previously, we ensured the `symuser` and `symgroup` accounts existed for
every symbolic file mode combination, of which there are over 40. Due to
BKR-628, we were updating the `PrimaryGroupID` of the `symgroup` group
everytime on OSX, leading to an apparent race condition where we would
`chmod` a file to the old gid, and then try to resolve the gid to a
name, but since the gid no longer existed, the test would fail.

This commit modifies the acceptance test to use puppet to ensure the
user and group accounts (avoiding the beaker issue), moves the ensure
check to once per agent, and adds a teardown to ensure the accounts are
removed when the test is finished.

[skip ci]

### DIFF
--- a/acceptance/tests/resource/file/symbolic_modes.rb
+++ b/acceptance/tests/resource/file/symbolic_modes.rb
@@ -82,13 +82,8 @@ class ModifiesModeTest < ActionModeTest
 
     @start_mode = start_mode
 
-    user = 'symuser'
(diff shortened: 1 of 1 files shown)


joshcooper:
Waiting until after puppet 4.3.1

PR #4455, opened 2015-11-19
(docs) Rewrite description of the lookup function

nfagerlund:
When this is merged, please resolve DOC-2225.

### DIFF
--- a/lib/puppet/functions/lookup.rb
+++ b/lib/puppet/functions/lookup.rb
@@ -1,153 +1,129 @@
-# Looks up data defined using Data Binding, and Data Providers using different strategies. The lookup searches in
-# Data Bindings first (if configured; typically Hiera), then in the environments data provider (if any), and last in
-# the module's data provider (if any) of the module the call to lookup originates from. Thus, the global Data Binding
-# has higher priority than data provided in the environment, which has higher priority than data provided in a module,
+# Uses the Puppet lookup system to retrieve a value for a given key. By default,
+# this returns the first value found (and fails compilation if no values are
+# available), but you can configure it to merge multiple values into one, fail
+# gracefully, and more.
 #
-# The lookup function can be called in one of these ways:
+# When looking up a key, Puppet will search up to three tiers of data, in the
+# following order:
 #
-#     lookup(name)
-#     lookup(name, value_type)
-#     lookup(name, value_type, merge)
-#     lookup(name, value_type, merge, default_value)
-#     lookup(options_hash)
-#     lookup(name, options_hash)
+# 1. Hiera.
+# 2. The current environment's data provider.
+# 3. The indicated module's data provider, if the key is of the form
+#    `<MODULE NAME>::<SOMETHING>`.
 #
-# The function may optionally be called with a code block / lambda with the following signature:
+# #### Arguments
 #
-#     lookup(...) |$name| { ... }
+# You must provide the name of a key to look up, and can optionally provide other
+# arguments. You can combine these arguments in the following ways:
 #
-# The block, if present, is mutually exclusive to the `default_value` and will be called with the `name` used in the
-# lookup when no value is found. The value produced by the block then becomes the value produced by the lookup.
+# * `lookup( <NAME>, [<VALUE TYPE>], [<MERGE BEHAVIOR>], [<DEFAULT VALUE>] )`
+# * `lookup( [<NAME>], <OPTIONS HASH> )`
+# * `lookup( as above ) |$key| { # lambda returns a default value }`
 #
-# The meaning of the parameters or content of the options hash is:
+# Arguments in `[square brackets]` are optional.
 #
-# * `name` - The name or array of names to lookup (first found is returned)
-# * `value_type` - The type to assert. Defaults to 'Data' See 'Type Specification' below.
-# * `default_value` - The default value if there was no value found (must comply with the data type)
-# * `override` - a hash with map from names to values that are used instead of the underlying bindings. If the name
-#   is found here it wins. Defaults to an empty hash.
-# * `default_values_hash` - a hash with map from names to values that are used as a last resort to obtain a value.
-#   Defaults to an empty hash.
-# * `merge` - A string of type Enum[unique, hash, merge] or a hash with the key 'strategy' set to that string. See
-#   'Merge Strategies' below.
+# The arguments accepted by `lookup` are as follows:
 #
-# It is not permitted to pass the `name` as both a parameter and in the options hash.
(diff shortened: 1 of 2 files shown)


lib/puppet/functions/lookup.rb
```diff
@@ -1,153 +1,128 @@
...
+# available), but you can configure it to merge multiple values into one, fail
+# gracefully, and more.
+#
+# When looking up a key, Puppet will search up to three tiers of data, in the
+# following order:
+#
+# 1. Hiera.
+# 2. The current environment's data provider.
+# 3. The indicated module's data provider, if the key is of the form
+#    `<MODULE NAME>::something`.
```
gguillotte: Can we use a capped bracketed keyword (maybe <KEY>?) in place of "something" for consistency?

lib/puppet/functions/lookup.rb
```diff
@@ -1,153 +1,128 @@
...
+# Arguments in `[square brackets]` are optional.
+#
+# The arguments accepted by `lookup` are as follows:
+#
+# 1. `<NAME>` (string or array) --- The name of the key to look up.
+#     * This can also be an array of keys. If Puppet doesn't find anything for the
+#     first key, it will try again with the subsequent ones, only resorting to a
+#     default value if none of them succeed.
+# 2. `<VALUE TYPE>` (data type) --- A
+# [data type](http://docs.puppetlabs.com/puppet/latest/reference/lang_data_type.html)
```
gguillotte: s/http/https/

lib/puppet/functions/lookup.rb
```diff
@@ -1,153 +1,128 @@
...
+# found, Puppet will pass the requested key to the lambda and use its result as
+# the default value.
+#
+# #### Merge Behaviors
+#
+# Puppet lookup uses a hierarchy of data sources, and a given key might have
+# values in multiple sources. By default, Puppet returns the first value it finds,
+# but it can also continue searching and merge all the values together.
+#
+# **Note:** Data sources can use the special `lookup_options` metadata key to
```
gguillotte: Can we indent/blockquote notes here?

lib/puppet/functions/lookup.rb
nfagerlund: Yeah, works for me.

lib/puppet/functions/lookup.rb
nfagerlund: GOOD CATCH

lib/puppet/functions/lookup.rb
```diff
@@ -1,153 +1,128 @@
...
+# request a specific merge behavior for a key. The `lookup` function will use that
+# requested behavior unless you explicitly specify one.
+#
+# The valid merge behaviors are:
+#
+# * `'first'` --- Returns the first value found, with no merging. Puppet lookup's
+# default behavior.
+# * `'unique'` (AKA "array merge") --- Combines any number of arrays and scalar
```
gguillotte: "AKA" seems awkward here. Can we note that it's an array merge in the description?

lib/puppet/functions/lookup.rb
```diff
@@ -1,153 +1,128 @@
...
+#     values. If Puppet can't find a normal value, it will check this hash for the
+#     requested key before giving up. You can combine this with `default_value` or
+#     a lambda, which will be used if the key isn't present in this hash. Defaults
+#     to an empty hash.
+#     * `'override'` (hash) --- A hash of lookup keys and override values. Puppet
+#     will check for the requested key in the overrides hash _first;_ if found, it
+#     returns that value as the _final_ value, ignoring merge behavior. Defaults
+#     to an empty hash.
+#
+# Finally, `lookup` can take a lambda, which should accept a single parameter.
```
gguillotte: Why hedge this with "should"? In what circumstances the lambda not take a single parameter?

lib/puppet/functions/lookup.rb
nfagerlund: :+1:

lib/puppet/functions/lookup.rb
nfagerlund: Well... this was one of those arbitrary UI changes, where they changed the name of something. In Hiera, this behavior is called "array", and in Puppet lookup it's called "unique". So I could reword it to say "(called 'array merge' in classic Hiera)" or something.

lib/puppet/functions/lookup.rb
nfagerlund: ah, you're right. I always miss my chance to drop a MUST.

lib/puppet/functions/lookup.rb
```diff
@@ -1,153 +1,128 @@
...
+# values.
+# * `'deep'` --- Combines the keys and values of any number of hashes to return a
+# merged hash. If the same key exists in multiple source hashes, Puppet will
+# recursively merge hash or array values (with duplicate values removed from
+# arrays). For conflicting scalar values, the highest-priority value will win.
+# * `{'strategy' => 'first|unique|hash'}` --- Same as the string versions of these
+# merge behaviors.
+# * `{'strategy' => 'deep', <DEEP OPTION> => <VALUE>, ...}` --- Same as `'deep'`,
+# but can adjust the merge with additional options. The `strategy` key is
+# required, and you can set any of these optional keys:
```
gguillotte: What does "The `strategy` key is required" mean here? Does it mean you can't set options for the `deep` merge behavior unless you use the `strategy` hash? Or does it mean you always have to use the `strategy` hash? I'm guessing it's the former.

lib/puppet/functions/lookup.rb
```diff
@@ -1,153 +1,128 @@
...
+#     * `'knockout_prefix'` (string or undef) --- A string prefix to indicate a
+#     value should be _removed_ from the final value. Defaults to `undef`, which
```
gguillotte: removed from the final **result**, instead of the final **value**?

lib/puppet/functions/lookup.rb
```diff
@@ -1,153 +1,128 @@
...
+#
+# @example Look up a key and return the first value found
+#
+#     lookup('ntp::service_name')
+#
+# @example Do a unique merge lookup of class names, then add all of those classes to the catalog (like `hiera_include`)
+#
+#     lookup('classes', Array[String], 'unique').include
+#
+# @example Do a deep hash merge lookup of user data (where the data sources can remove items with a knockout prefix)
```
gguillotte: Maybe be more specific in the description?

> Do a deep hash merge lookup of user data, but knock out matches prefixed by '--'

gguillotte:
:+1: gonna push the :radio_button:

PR #4457, opened 2015-11-20
(PUP-5127) Fix hung process caused by unclosed multiline comments

thallgren:
The lexer did not detect a multiline comment that wasn't closed and
returned the same parse position over and over again causing the
parser to loop forever. This commit remedies that and adds a new
issue UNCLOSED_MLCOMMENT to the set of issues reported by the parser.


hlindberg:
Merge of this needs to wait until 4.3.1. is released.

hlindberg:
Reviewed - good to merge.

PR #4459, opened 2015-11-21
(PUP-5525) Fix pseudo variable assignment that happens too late

thallgren:
The pseudo variable for calling_module was assigned too late when
a resource was evaluated which resulted in that a hiera interpolation
used when finding default values didn't interpolate the value
correctly.
This commit ensures that the value is assigned before the external
values are resolved which means that hiera will find the right
values.

I suggest the addition of an acceptance test to verify this behavior.


hlindberg:
Reviewed: looks good to merge. Ping @kylog

PR #4460, opened 2015-11-23
(PUP-5513) Allow blank lines in ssh known hosts file

whopper:
Previously, blank lines in the default ssh_known_hosts file would
cause errors in the sshkey type due to a faulty regex. This commit
updates said regex to properly match blank lines with no whitespace
characters.


joshcooper:
:+1: from me, but I'll be out tomorrow and don't want to merge right before I go.

PR #4462, opened 2015-11-24
(PUP-5285) Update systemd provider to detect all unit types

whopper:
Prior to this commit, when searching for discoverable systemd
services we filtered our results down to only units of the 'service'
type. There are several other valid types, such as timer and socket
which can be managed as well. This commit updates our systemctl
invocation to allow any unit type to be returned in its output.

Note that managing systemd unit types other than the service type
requires the fully-qualified service name to be provided. For example,
to manage the unit 'foo.timer', Puppet must be provided with 'foo.timer'
rather than just 'foo'. Otherwise, the provider will default to managing
the service unit type.

### DIFF
--- a/acceptance/tests/resource/service/systemd_unit_types.rb
+++ b/acceptance/tests/resource/service/systemd_unit_types.rb
@@ -1,0 +1,23 @@
+test_name 'Systemd provider should recognize non-service unit types'
+
+confine :to, :platform => /fedora-22/
+
+# This test is intended to ensure that systemd unit types other than service are
+# also discoverable via puppet. Examples include timer and socket units.
+
+agents.each do |agent|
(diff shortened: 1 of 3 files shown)


whopper:
The test I added currently only runs on Fedora 22, as I'm not familiar with any packages in Debian which have timer units. I believe testing in F22 alone is probably sufficient, but if need be we can probably find some timers in Debian.

whopper:
@kylog btw, I just found https://tickets.puppetlabs.com/browse/PUP-1732 and realized that this goes back on that. Do you have any thoughts on which way is "correct"?

kylog:
Good question. I raised it on the Jira ticket which probably has more stakeholders watching.

PR #4464, opened 2015-11-24
(PUP-1455) Use CFPropertyList to read launchd plists

MikaelSmith:
This uses the CFPropertyList gem which is shipped with Puppet to read launchd
plists instead of shelling out to plutil to convert them and then reading
them in. This saves a lot of time, between 10 and 20 seconds per Puppet run.

This commit introduces a Puppet::Util::Plist namespace for handling plists
with CFPropertyList and modifies the existing launchd service provider to
use it. Activities in Puppet::Util::Plist that will use common libraries
have been wrapped in their own helper methods (which will allow us to mock
out those methods for testing purposes). This level of encapsulation will
allow us to swap out the underlying implementation should things change in
the future.


MikaelSmith:
I've run scenarios managing `user/group/service/appdmg/pkgdmg`. `macauthorization` stopped working with OS X 10.9, so I don't have a convenient way to check that it still works, but the specs pass.

MikaelSmith:
The PR from @keeleysam won't pass specs because there's a conflict between `Puppet::Util::Plist and Plist` in the other providers.

keeleysam:
👍🏻

MikaelSmith:
This is most effective with nokogiri installed.

kylog:
@MikaelSmith does this need a rebase now that #4447 is in?

MikaelSmith:
Oh, probably.

kylog:
Closing and re-opening to wake travis up.

PR #4466, opened 2015-11-25
(PUP-5548) Ensure systemctl is present before using it for service status

whopper:
Currently, the Debian service provider uses systemctl to determine
service status on Debian 8 and Ubuntu 15.04, which use systemd by
default. This is necessary due to the way masked services behave
when queried with the sysvint-systemd compatibility layer.

It is possible to remove systemd from Debian 8 and Ubuntu 15.04
and only use init. In such a case, the Debian provider will error
when trying to use systemctl. This commit updates the provider to
check that systemctl exists on the system before attempting to use
it to determine service status. If it is not present, the provider
falls back to the service init script.

### DIFF
--- a/lib/puppet/provider/service/debian.rb
+++ b/lib/puppet/provider/service/debian.rb
@@ -73,8 +73,8 @@ Puppet::Type.type(:service).provide :debian, :parent => :init do
       majversion = Facter.value(:operatingsystemmajrelease).split('.')[0].to_i
     end
 
-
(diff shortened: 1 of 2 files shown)


whopper:
My VPN access is out so I haven't been able to thoroughly test this yet. I want to A) revisit the masked service issue on Debian 8 / Ubuntu 15.04 to make sure using `systemctl` really is necessary (or if there's another way to do it using only sysvinit) and B) remove systemd from a box and try this out in a real scenario.

/cc @aqueos

PR #4467, opened 2015-11-25
(PUP-5060) Fix lookup --facts bug

HAIL9000:
Prior to this commit, the facts passed in via the --facts were not
being promoted to topscope because they were not being added to the
node parameters. This meant they would not be found in a lookup.

Now, facts in the given fact file are merged into the node parameters
so that when the node is compiled they will be promoted to topscope.

### DIFF
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -336,7 +336,7 @@ Copyright (c) 2015 Puppet Labs, LLC Licensed under the Apache 2.0 License
     fact_file = options[:fact_file]
 
     if fact_file
-      original_facts = node.facts.values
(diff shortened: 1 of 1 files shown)


hlindberg:
Did you by any chance test what happens if you try --facts with a fact like `trusted => { certname => 'x' }`? 
(This to ensure it does not do anything crazy).

hlindberg:
Can we test this somehow?

HAIL9000:
@hlindberg Sure, my vm is still up. I'll test that right now

hlindberg:
Merging, we need to specify how 'trusted' should work - the problem this PR addresses should be fixed anyhow.

PR #4469, opened 2015-11-25
(PUP-5588) Prevent redundant cert loading

reidmv:
We monkey patch OpenSSL::X509::Store#set_default_paths to load our own
additional certs. Sometimes one or more of our certs are already loaded,
however. Previously the code would try to load the cert and in doing so
raise, catch, and warn on the cert already being loaded. Under certain
conditions this superflous warning could print very verbosely, obscuring
the useful output.

This commit quiesces the warn, replacing it with a Puppet#debug. Such a
message at the warn level is not useful to the user and can cause
unnecessary concern or frustration regarding the product.

### DIFF
--- a/lib/puppet/util/monkey_patches.rb
+++ b/lib/puppet/util/monkey_patches.rb
@@ -114,15 +114,19 @@ if Puppet::Util::Platform.windows?
   require 'openssl'
 
   class OpenSSL::X509::Store
+    @puppet_certs_loaded = false
     alias __original_set_default_paths set_default_paths
     def set_default_paths
       # This can be removed once openssl integrates with windows
       # cert store, see https://rt.openssl.org/Ticket/Display.html?id=2158
-      Puppet::Util::Windows::RootCerts.instance.to_a.uniq.each do |x509|
-        begin
-          add_cert(x509)
-        rescue OpenSSL::X509::StoreError => e
-          warn "Failed to add #{x509.subject.to_s}"
+      unless @puppet_certs_loaded
+        @puppet_certs_loaded = true
+        Puppet::Util::Windows::RootCerts.instance.to_a.uniq.each do |x509|
+          begin
+            add_cert(x509)
+          rescue OpenSSL::X509::StoreError => e
+            warn "Failed to add #{x509.subject.to_s}"
+          end
         end
       end
 
--- a/spec/unit/util/monkey_patches_spec.rb
+++ b/spec/unit/util/monkey_patches_spec.rb
@@ -186,6 +186,14 @@ describe OpenSSL::X509::Store, :if => Puppet::Util::Platform.windows? do
     store.set_default_paths
   end
 
+  it "doesn't warn when calling set_default_paths multiple times" do
+    with_root_certs([cert])
+    store.expects(:warn).never
+
+    store.set_default_paths
+    store.set_default_paths
+  end
+
   it "ignores duplicate root certs" do
     with_root_certs([cert, cert])
 


kylog:
This seems like a reasonable change.

But ping @Iristyle or @ferventcoder for two questions:
1) Is the AppVeyor fail a familiar pattern? It isn't obviously specific to this PR.
2) There's a 2+ year old comment in the code being changed that "This can be removed once openssl integrates with windows cert store". Do we know if that ever happened? (The link provided now requires a login, so I found no easy answer.)

joshcooper:
About the backstory, the ruby devs tried submitting a patch to openssl so that `set_default_paths` would allow openssl to integrate with the Windows cert store: https://bugs.ruby-lang.org/issues/2586, but it was rejected by openssl devs, more commentary at http://www.mail-archive.com/openssl-dev@openssl.org/msg26956.html.

As for changing to `debug`, we already `uniq` the array of certs that we get from the cert store. So we should never try to call `add_cert` with a duplicate. See https://tickets.puppetlabs.com/browse/PUP-734

But perhaps that's not working as expected? It doesn't look like the `OpenSSL::X509::Certificate` class defines `hash` or `eql?`, so perhaps `uniq` isn't working as it should.

reidmv:
I ran into the issue when using a custom Puppet face which repeatedly calls out to other faces (e.g. `Puppet::Face[:module].install`) as well as re-using the non-face `apply` application. Is `set_default_paths` global? Is it possible the `OpenSSL::X509::Store#set_default_paths` is being called repeatedly?

ferventcoder:
> 1) Is the AppVeyor fail a familiar pattern? It isn't obviously specific to this PR.

~~~sh
1) OpenSSL::X509::Store debugs when adding a certificate that already exists
     Failure/Error: store.expects(:debug).with('Failed to add /DC=com/DC=microsoft/CN=Microsoft Root Certificate Authority')
 
     Mocha::ExpectationError:
       not all expectations were satisfied
       unsatisfied expectations:
       - expected exactly once, not yet invoked: #<OpenSSL::X509::Store:0x139de960>.debug('Failed to add /DC=com/DC=microsoft/CN=Microsoft Root Certificate Authority')
       satisfied expectations:
       - expected exactly once, invoked once: Puppet::Util::Windows::RootCerts.instance(any_parameters)
     # ./spec/unit/util/monkey_patches_spec.rb:201:in `block (2 levels) in <top (required)>'
~~~

Seems related?

ferventcoder:
For `2)`, I think @joshcooper has the best answer for that.

kylog:
@ferventcoder oops my bad, you're right, the AppVeyor failure does look related. (I may have been fooled by the slew of Pending test outputs. Really want to clean that up some day.)

ferventcoder:
Yeah @kylog clean it up or fix them ;)

joshcooper:
`set_default_paths` is an instance method on the X509 store associated with the Net::HTTP object. Puppet only calls this method in one place: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/forge/repository.rb#L137, and it only calls the method once. It's possible some other code is getting access to the Net::HTTP object, getting the store, and calling `set_default_paths` on that. It would be good to understand why that's so.

reidmv:
Ah - the error is me not knowing exactly how to rewrite the test. Pushed a change just now that might do it but I'm borrowing the CI to do my Windows tests for me - I don't already have a Window box set up in a way that makes it easy to do that.

reidmv:
@joshcooper the face I'm working on uses `open-uri`. I've worked on types/providers in the past that use Net::HTTP directly too (e.g. https://github.com/lwf/puppet-remote_file).

joshcooper:
@reidmv ah, yes that'd do it. I'm thinking this is a bug in our monkey_patch. It should only try to load certs once, even if `set_default_paths` is called multiple times.

Iristyle:
Right, so @reidmv could you try changing the definition at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/monkey_patches.rb#L116-L130 to something like:


```ruby
class OpenSSL::X509::Store
  @@puppet_certs_loaded = false
  alias __original_set_default_paths set_default_paths
  def set_default_paths
    # This can be removed once openssl integrates with windows
    # cert store, see https://rt.openssl.org/Ticket/Display.html?id=2158
    if !@@puppet_certs_loaded
      Puppet::Util::Windows::RootCerts.instance.to_a.uniq.each do |x509|
        begin
          add_cert(x509)
        rescue OpenSSL::X509::StoreError => e
          warn "Failed to add #{x509.subject.to_s}"
        end
      end
      @@puppet_certs_loaded = true
    end

    __original_set_default_paths
  end
end
```

joshcooper:
On Thursday, December 3, 2015, Ethan J. Brown <notifications@github.com>
wrote:

> Right, so @reidmv <https://github.com/reidmv> could you try changing the
> definition at
> https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/monkey_patches.rb#L116-L130
> to something like:
>
> class OpenSSL::X509::Store
>   @@puppet_certs_loaded = false
>
>
I think we just want an instance variable.

>
>   alias __original_set_default_paths set_default_paths
>   def set_default_paths
>     # This can be removed once openssl integrates with windows
>     # cert store, see https://rt.openssl.org/Ticket/Display.html?id=2158
>     if !@@puppet_certs_loaded
>       Puppet::Util::Windows::RootCerts.instance.to_a.uniq.each do |x509|
>         begin
>           add_cert(x509)
>         rescue OpenSSL::X509::StoreError => e
>           warn "Failed to add #{x509.subject.to_s}"
>         end
>       end
>       @@puppet_certs_loaded = true
>     end
>
>     __original_set_default_paths
>   endend
>
> —
> <https://github.com/puppetlabs/puppet/pull/4469#issuecomment-161798757>.
>


-- 
Josh Cooper
Developer, Puppet Labs

joshcooper:
Thanks @reidmv for your contribution! I had a couple of comments to wrap this up.

* Since this turns out to be a bug in puppet, could you file a PUP ticket and update the commit message summary?
* The spec "doesn't warn when calling set_default_paths multiple times" is confined (in an outer describe block) to only run on Windows, and it looks like it's failing in AppVeyor (where we run PRs on Windows). Could you take a look?
* Could you squash the two commits?

joshcooper:
ping @Iristyle

joshcooper:
ping @Iristyle @craig-gomes Can we get this pulled into a Windows sprint and get it merged? I think it's ready to go.

lib/puppet/util/monkey_patches.rb
```diff
@@ -114,16 +114,20 @@ def initialize(*args)
...
-          warn "Failed to add #{x509.subject.to_s}"
+      if !@puppet_certs_loaded
+        Puppet::Util::Windows::RootCerts.instance.to_a.uniq.each do |x509|
+          begin
+            add_cert(x509)
+          rescue OpenSSL::X509::StoreError => e
+            warn "Failed to add #{x509.subject.to_s}"
+          end
         end
+        @puppet_certs_loaded = true
```
ferventcoder: Should this come first to lessen the chances of a double load?

ferventcoder: It's going to set this value to true anyway.

lib/puppet/util/monkey_patches.rb
```diff
@@ -114,16 +114,20 @@ def initialize(*args)
...
     alias __original_set_default_paths set_default_paths
     def set_default_paths
       # This can be removed once openssl integrates with windows
       # cert store, see https://rt.openssl.org/Ticket/Display.html?id=2158
-      Puppet::Util::Windows::RootCerts.instance.to_a.uniq.each do |x509|
-        begin
-          add_cert(x509)
-        rescue OpenSSL::X509::StoreError => e
-          warn "Failed to add #{x509.subject.to_s}"
+      if !@puppet_certs_loaded
```
glennsarti: Prefer unless

```ruby
unless @puppet_certs_loaded
....
end
```

lib/puppet/util/monkey_patches.rb
glennsarti: @reidmv @ferventcoder Agreed.  As it's just a warning not raise, not worth warning continuously.

lib/puppet/util/monkey_patches.rb
reidmv: changed to `unless`

lib/puppet/util/monkey_patches.rb
reidmv: Moved `@puppet_certs_loaded = true` to come first as soon as the block is entered

Iristyle:
I think this should have been targeted at `stable` based on the version we intended to ship this with.  I'm going to cherry-pick 4373d41b3943cc7830277a37c3ba100d04037254 back to stable.

PR #4471, opened 2015-11-27
(PUP-5555) Fix yum package provider in case of broken update notice

ckaenzig:
This PR fixes a bug where the yum package provider would fail to
`ensure => latest` when `yum check-updates` lists broken update notices.


kylog:
@ckaenzig, this looks great: thanks!

PR #4472, opened 2015-11-28
(maint) Minor fixes to the pkgng package provider

xaque208:
This work updates the pkgng package provider to ensure that port names as resource titles are handled correctly, and that fewer calls to 'pkg version' are made.

### DIFF
--- a/lib/puppet/provider/package/pkgng.rb
+++ b/lib/puppet/provider/package/pkgng.rb
@@ -83,8 +83,16 @@ Puppet::Type.type(:package).provide :pkgng, :parent => Puppet::Provider::Package
     case resource[:ensure]
     when true, false, Symbol
       installname = resource[:name]
-    else
-      installname = resource[:name] + '-' + resource[:ensure]
+      else
+        # If resource[:name] is actually an origin (e.g. 'www/curl' instead of
+        # just 'curl'), drop the category prefix. pkgng doesn't support version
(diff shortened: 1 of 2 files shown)


HAIL9000:
@xaque208 Thanks for the contribution! It looks like some indentation got messed up with the case statement, would you mind fixing that up?

xaque208:
Fixed the indentation.

lib/puppet/provider/package/pkgng.rb
```diff
@@ -17,7 +17,7 @@ def self.get_query
   end
 
   def self.get_version_list
-    pkg(['version', '-voRL='])
+    @version_info_list ||= pkg(['version', '-voRL='])
```
branan: I know that in theory the agent is forked, but caching in providers still makes me kinda nervous.

It looks like this is only used in `self.instances`. I tink it would make more sense for that method to call `get_version_list` and then pass the list in to `get_latest_version`, instead of `get_latest_version` trying to query the version list each time it runs.

That way we only query the version list once a run still, but we don't risk stale data hanging around if the agent doesn't fork.

xaque208:
@branan I believe the last commit addresses your concern about caching in the provider.  Thank you for raising it.  I think this is a bit cleaner now.

branan:
:+1: from me at this point

PR #4473, opened 2015-12-01
(PUP-5549) Revert "(PUP-5025) RPM Epoch should be included in ensure field

whopper:
This reverts commit 1fdaeae9fa4f18c1c96da3d6fd84613dab6024f8, which
introduced a regression when using the yum provider under certain
circumstances.

This change was originally meant to fix a DNF issue when installing RPMs
which included the epoch field in their release strings. However, it
also brought about a new issue in the yum provider when trying to
manage packages which use the epoch field. Upon every run, Puppet throws
a notice to indicate that the package version has been updated to
include the epoch. Adding the epoch into a manifest results in an error
when installing a package which isn't already on the system.

This commit reverts the original change until further investigation and
more comprehensive changes can be made.


kylog:
Can you point this at stable?

PR #4475, opened 2015-12-01
(PUP-5552) Validate EPP parameters

thallgren:
The refactor of parameter validation that was made in PUP-4458 left
the EPP template parameters out. That was an oversight. This commit
ensures that EPP templates are validated using the same logic that
is used when validating resource parameters.


lib/puppet/pops/evaluator/closure.rb
```diff
@@ -238,6 +199,45 @@ def create_callable_type()
     Puppet::Pops::Types::TypeFactory.callable(*(types + range))
   end
 
+  def create_params_struct
```
hlindberg: This should probably be "@api private"...

PR #4470, opened 2015-11-27
(PUP-5554) Improve parser performance on JRuby

thallgren:
This PR contains two commits. One that optimizes the lexer and one that optimizes the parser
factory. The measurements that I've done indicates that when this PR is applied, parsing will spend:

55% less CPU time with JRuby (tested using JRuby 9.0.4)
13% less CPU time with MRI (tested on 1.9.3, 2.1.1, and 2.2.1).

The parser is 11% faster with MRI than with JRuby.

PR #4476, opened 2015-12-01
(PUP-5554) Improve parser performance

thallgren:
This PR contains two commits. One that optimizes the lexer and one that optimizes the parser
factory. The measurements that I've done indicates that when this PR is applied, parsing will spend:

55% less CPU time with JRuby (tested using JRuby 9.0.4)
13% less CPU time with MRI (tested on 1.9.3, 2.1.1, and 2.2.1).

The parser is 11% faster with MRI than with JRuby.

This PR is for the stable branch and supersedes #4470 which was for master.


hlindberg:
needs a rebase

PR #4478, opened 2015-12-02
(PUP-5548) Use `service` in debian provider to interact with all init systems

whopper:
In the Debian family, the `service` command provides an
abstraction around available init systems, allowing it to be used
to properly determine service state regardless of which init system
is active on the system.

This commit updates the Debian provider's `statuscmd` method to use
`service` at all times, allowing the provider to properly query
services regardless of OS version or init system in place.

### DIFF
--- a/lib/puppet/provider/service/debian.rb
+++ b/lib/puppet/provider/service/debian.rb
@@ -15,6 +15,7 @@ Puppet::Type.type(:service).provide :debian, :parent => :init do
   # http://projects.reductivelabs.com/issues/2538
   # is resolved.
   commands :invoke_rc => "/usr/sbin/invoke-rc.d"
+  commands :service => "/usr/sbin/service"
 
   defaultfor :operatingsystem => :cumuluslinux
   defaultfor :operatingsystem => :debian, :operatingsystemmajrelease => ['5','6','7']
@@ -65,23 +66,9 @@ Puppet::Type.type(:service).provide :debian, :parent => :init do
   end
 
   def statuscmd
-    os = Facter.value(:operatingsystem).downcase
-
-    if os == 'debian'
-      majversion = Facter.value(:operatingsystemmajrelease).to_i
-    else
-      majversion = Facter.value(:operatingsystemmajrelease).split('.')[0].to_i
-    end
-
-
-    if ((os == 'debian' && majversion >= 8) || (os == 'ubuntu' && majversion >= 15))
-      # SysVInit scripts will always return '0' for status when the service is masked,
-      # even if the service is actually stopped. Use the SysVInit-Systemd compatibility
-      # layer to determine the actual status. This is only necessary when the SysVInit
-      # version of a service is queried. I.e, 'ntp' instead of 'ntp.service'.
-      (@resource[:hasstatus] == :true) && ["systemctl", "is-active", @resource[:name]]
-    else
-      super
-    end
+      # /usr/sbin/service provides an abstraction layer which is able to query services
+      # independent of the init system used.
+      # See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775795
+    (@resource[:hasstatus] == :true) && [command(:service), @resource[:name], "status"]
   end
 end
--- a/spec/unit/provider/service/debian_spec.rb
+++ b/spec/unit/provider/service/debian_spec.rb
@@ -27,6 +27,7 @@ describe provider_class do
 
     @provider.stubs(:command).with(:update_rc).returns "update_rc"
     @provider.stubs(:command).with(:invoke_rc).returns "invoke_rc"
+    @provider.stubs(:command).with(:service).returns "service"
 
     @provider.stubs(:update_rc)
     @provider.stubs(:invoke_rc)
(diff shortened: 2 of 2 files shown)


whopper:
Retargeted version of https://github.com/puppetlabs/puppet/pull/4466

kylog:
:+1:

Will merge after puppet-agent 1.3.2 is out.

whopper:
@kylog actually, maybe don't quite merge this yet, as there may be more to the systemd saga yet. There's lots of interesting context to be found in this debian puppet bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775795.

Specifically this bit, which better illuminates how insane the situation is on Debian. Included also is the patch which Debian is currently putting on top of puppet, which is promising. It doesn't currently work due to an issue with certain services including an '@', but we may be able to modify it to work. See https://bugs.debian.org/cgi-bin/bugreport.cgi?msg=84;bug=775795;att=1;filename=0001-Fix-service-listing-and-enable-disable-in-Debian.patch.

Excerpt:

Unfortunately, switching the default provider to systemd wouldn't work 
either (see below). Besides that, we still have architectures where sysv-rc is
the default.

To put things in perspective, we currently have three types of services:

 A those shipping an initscript only
 B those shipping an initscript and a systemd service file
 C those shipping only a systemd service file and no initscript. This 
   type includes all systemd-specific services, and - inevitably - many 
   services written/managed by sysadmins.

Gaudenz's patch makes sure that start/stop works for all three types of 
services, however it does not change the enable/disable logic nor the 
service listing, which are still problematic:

 - `update-rc.d enable|disable' does not work for type C services, as 
   demonstrated by Faidon.
 - `update-rc.d -f remove && update-rc.d defaults' as currently used in 
   the `enable' does *not* enable type C *and* type B services at all.  
   `update-rc.d enable' would enable type B services but not type C.
 - querying service enable status using invoke-rc.d also doesn't work 
   for type C services.

If OTOH we were to change the default provider to systemd, we wouldn't 
be much better off:

 - `systemctl is-enabled' doesn't work for type A services, which are 
   still the majority:

   $ systemctl is-enabled ferm
   Failed to get unit file state for ferm.service: No such file or directory

 - `systemctl list-unit-files' as used by systemd's self.instances, does 
   not list type A services as well:

   $ systemctl list-unit-files --all --type service --no-pager | grep ferm
   $

So, we actually need a hybrid provider that will do the following:

 - Detect if we're running systemd as PID 1. This is trivial to do by 
   checking for the existence of /run/systemd/system.

 - Keep compatibility with sysv-rc (and I'd suggest to drop pre-2.88 
   support and spare an ugly call to dpkg).

 - If running under systemd:
   • Use `systemctl enable|disable' for *all* services. This works 
     correctly with our current systemd version for all three types of 
     services and invokes update-rc.d if appropriate.

   • Use `systemctl is-enabled` to query for enabled services only for 
     types B and C. Type A services can be detected because their 
     (auto-generated) units have the SourcePath property set to the 
     initscript path:

     $ systemctl show -pSourcePath ferm.service
     SourcePath=/etc/init.d/ferm

     For type A we need to keep the current invoke-rc.d implementation.

   • For self.instances, augment the type B and C services returned by
     `systemctl list-unit-files' with type A services from the init provider's
     self.instances.

whopper:
@kylog updated this with a new twist. It turns out `service` is actually an abstraction around init systems in the Debian family, so we can safely use it to get status. I ran all of our service acceptance tests against this is Ubuntu 14.04 / 15.04 and Debian 7 and 8.

There are also the two alternate init system scenarios that this fixes:

1) The case where systemd has been thoroughly purged and replaced with sysvinit, as is detailed here: http://without-systemd.org/wiki/index.php/How_to_remove_systemd_from_a_Debian_jessie/sid_installation

2) The case where systemd is replaced with sysv-rc as the default init system on the box, but the systemd package is still around. This happens when you follow Debian's guide in Jessie's release notes: https://www.debian.org/releases/jessie/amd64/release-notes/ch-information.en.html#systemd-upgrade-default-init-system.

`service` is also present in cumulus linux, though I'm going to poke around a bit more to make sure everything's fine there.

/cc @branan, if you have any time I'd appreciate some extra review of this. This issue is pretty thorny and I want to make sure to get it right this time :P

branan:
@whopper I'm looking over this. Everything in your comments makes me super sad. Please add it to the Release Engineering document on Debian in Confluence.

whopper:
Looks to be working fine in cumulus as well:

```
root@cumulus:~# puppet resource service mcollective --param provider
service { 'mcollective':
  ensure   => 'running',
  enable   => 'true',
  provider => 'debian',
}
root@cumulus:~# puppet resource service mcollective --param provider ensure=stopped enable=false
Notice: /Service[mcollective]/ensure: ensure changed 'running' to 'stopped'
service { 'mcollective':
  ensure   => 'stopped',
  enable   => 'false',
  provider => 'debian',
}
root@cumulus:~# puppet resource service mcollective --param provider
service { 'mcollective':
  ensure   => 'stopped',
  enable   => 'false',
  provider => 'debian',
}
root@cumulus:~# puppet resource service mcollective --param provider ensure=running enable=true
Notice: /Service[mcollective]/ensure: ensure changed 'stopped' to 'running'
service { 'mcollective':
  ensure   => 'running',
  enable   => 'true',
  provider => 'debian',
}
root@cumulus:~# puppet resource service mcollective --param provider
service { 'mcollective':
  ensure   => 'running',
  enable   => 'true',
  provider => 'debian',
}
```

whopper:
Also seems to work on an Ubuntu 15.04 using upstart over systemd (which can be accomplished by following http://www.serverlab.ca/tutorials/linux/administration-linux/removing-systemd-from-ubuntu-15-2/)

kylog:
My mind is reeling. But it does seem like:
a) if debian is providing this as an abstraction, we should be using it
and b) it apparently works in all these scenarios so ....

:+1:

MikaelSmith:
Should we try co check this back to Debian 5 or 6?

whopper:
@MikaelSmith sure, I'll check them out.

kylog:
We don't ship Debian 5 packages either.

stahnma:
@branan your comment is best.

PR #4479, opened 2015-12-02
(PUP-5520) Exclude Yocto init scripts from service init provider

shermdog:
Yocto Linux ships a series of init scripts that are not safe to execute.  This creates unintended consequences such as sending shutdown signals to daemons.
This commit excludes the known list of unsafe scripts so that Puppet does not try to execute them via the service init provider.

### DIFF
--- a/acceptance/tests/resource/file/symbolic_modes.rb
+++ b/acceptance/tests/resource/file/symbolic_modes.rb
@@ -82,13 +82,8 @@ class ModifiesModeTest < ActionModeTest
 
     @start_mode = start_mode
 
-    user = 'symuser'
-    group = 'symgroup'
-    agent.user_present(user)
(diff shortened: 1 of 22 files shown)


glennmatthews:
They do not - thanks for checking. :+1: from me.

shermdog:
For reference:  https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-core/initscripts/initscripts-1.0

joshcooper:
Is this targeting the right branch? Currently, stable is for 4.3.x and master will eventually be 4.4.x

shermdog:
@joshcooper if that's the case I need to re-issue the PR against stable.  We're targeting 4.3.2.

PR #4480, opened 2015-12-03
(PUP-5520) Exclude Yocto init scripts from service init provider

shermdog:
Yocto Linux ships a series of init scripts that are not safe to execute.  This creates unintended consequences such as sending shutdown signals to daemons.
This commit excludes the known list of unsafe scripts so that Puppet does not try to execute them via the service init provider.

https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/recipes-core/initscripts/initscripts-1.0

Replaces #4479

### DIFF
--- a/lib/puppet/provider/service/init.rb
+++ b/lib/puppet/provider/service/init.rb
@@ -49,6 +49,13 @@ Puppet::Type.type(:service).provide :init, :parent => :base do
     excludes += %w{wait-for-state portmap-wait}
     # these excludes were found with grep -r -L start /etc/init.d
     excludes += %w{rcS module-init-tools}
+    # Prevent puppet failing on unsafe scripts from Yocto Linux
+    if Facter.value(:osfamily) == "cisco-wrlinux"
+      excludes += %w{banner.sh bootmisc.sh checkroot.sh devpts.sh dmesg.sh
+                   hostname.sh mountall.sh mountnfs.sh populate-volatile.sh
+                   rmnologin.sh save-rtc.sh sendsigs sysfs.sh umountfs
+                   umountnfs.sh}
+    end
     # Prevent puppet failing to get status of the new service introduced
     # by the fix for this (bug https://bugs.launchpad.net/ubuntu/+source/lightdm/+bug/982889)
     # due to puppet's inability to deal with upstart services with instances.
--- a/spec/unit/provider/service/init_spec.rb
+++ b/spec/unit/provider/service/init_spec.rb
@@ -39,7 +39,7 @@ describe Puppet::Type.type(:service).provider(:init) do
     before :each do
       described_class.stubs(:defpath).returns('tmp')
 
-      @services = ['one', 'two', 'three', 'four']
+      @services = ['one', 'two', 'three', 'four', 'umountfs']
       Dir.stubs(:entries).with('tmp').returns @services
       FileTest.expects(:directory?).with('tmp').returns(true)
       FileTest.stubs(:executable?).returns(true)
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/service/init.rb
```diff
@@ -61,6 +61,11 @@ def self.excludes
     excludes += %w{cryptdisks-udev}
     excludes += %w{statd-mounting}
     excludes += %w{gssd-mounting}
+    # Prevent puppet failing on unsafe scripts from Yocto Linux
+    excludes += %w{banner.sh bootmisc.sh checkroot.sh devpts.sh dmesg.sh
+                 hostname.sh mountall.sh mountnfs.sh populate-volatile.sh
+                 rmnologin.sh save-rtc.sh sendsigs sysfs.sh umountfs
+                 umountnfs.sh}
```
joshcooper: Are these safe to exclude for all platforms that support the `init` service provider?

stahnma: FWIW, you don't see *.sh files in init normally, so most of these would be just WRL specific. However, for safety it does seem like we should confine this to a specific platform.

mckern: @stahnma ORLLY?

```
vagrant@vagrant-deb7:~$ ls /etc/init.d/*.sh
/etc/init.d/bootmisc.sh             /etc/init.d/hostname.sh            /etc/init.d/mountdevsubfs.sh       /etc/init.d/mtab.sh
/etc/init.d/checkfs.sh              /etc/init.d/hwclock.sh             /etc/init.d/mountkernfs.sh         /etc/init.d/umountnfs.sh
/etc/init.d/checkroot-bootclean.sh  /etc/init.d/mountall-bootclean.sh  /etc/init.d/mountnfs-bootclean.sh
/etc/init.d/checkroot.sh            /etc/init.d/mountall.sh            /etc/init.d/mountnfs.sh
```

shermdog: From what I've seen a fair amount of these do exist on Debian based systems, but these scripts are relegated to boot and shutdown time.

In particular, some of the shutdown scripts behave okay on Deb when you try and status them, but they are definitely shutdown only bits.

* sendsigs
 * Kill all remaining processes.
* umountfs
  * Turn off swap and unmount all local file systems.
* umountnfs.sh
  * Unmount all network filesystems except the root fs.

Examples of the boot ones
* bootmisc.sh
  * Miscellaneous things to be done during bootup.
* hostname.sh
  * Set hostname based on /etc/hostname
* mountall.sh
  * Mount all filesystems.
* mountnfs
  * Wait for network file systems to be mounted
* rmnologin
  * Remove /etc/nologin at boot

stahnma: Friggen Debian. YA RLY.

kylog:
FWIW Debian is confined off this provider already: https://github.com/puppetlabs/puppet/blob/4.3.1/lib/puppet/provider/service/init.rb#L19-L23

However, that there may still be other distros that do similar fun things.

+1 on making this conditional on platform. Is there any downside to doing so?

stahnma:
@shermdog can you add the confines?

shermdog:
Updated with confinement

mckern:
works for giant man

stahnma:
:+1:

joshcooper:
Sorry to be a downer. Can we add a test that at least one of these scripts are excluded on cisco-wrlinux, and not excluded on other platforms?

shermdog:
@joshcooper where do you prefer that test live?

joshcooper:
Probably here in [init_spec](https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/service/init_spec.rb#L48-L60)

ScottGarman:
:+1:

PR #4481, opened 2015-12-04
(PUP-5577) Don't require /run/systemd/system for systemd

ianw:
Revert this part of 7fe61647f23650fc4c93cd6b54c654a20ff7c9f9 because
it means systemd is not detected if running puppet within a chroot
environment where this is not yet mounted.

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -5,7 +5,14 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
 
   commands :systemctl => "systemctl"
 
-  confine :exists => "/run/systemd/system"
+  if Facter.value(:osfamily).downcase == 'debian'
+    # With multiple init systems on Debian, it is possible to have
+    # pieces of systemd around (e.g. systemctl) but not really be
+    # using systemd.  We do not do this on other platforms as it can
+    # cause issues when running in a chroot without /run mounted
+    # (PUP-5577)
+    confine :exists => "/run/systemd/system"
+  end
 
   defaultfor :osfamily => [:archlinux]
   defaultfor :osfamily => :redhat, :operatingsystemmajrelease => "7"


kylog:
Heads up @whopper, since you've been looking at some other systemd use cases. Not sure if this overlaps at all or not, but figured it's worth a ping.

whopper:
This isn't directly related to the recent systemd issues we've been having, as this potentially affects any platforms that run systemd.

Removing this bit would mean that the provider is only confined by the existence of the `systemctl` command. In most cases, this doesn't change puppet's behavior, since usually if systemd itself isn't functional, the command won't be on the system. However, there are consequences to this in the Debian family where all kinds of weird things can be done to rip out systemd to replace it with sysvinit.

For example, by following Debian's guide on using init over systemd in Jessie (https://www.debian.org/releases/jessie/amd64/release-notes/ch-information.en.html#systemd-upgrade-default-init-system), you'll end up with a system that uses sysvinit as its init system, but *still* has pieces of systemd left over even though it isn't functional. While the directory `/run/systemd/system` won't exist, the `systemctl` command will still be on the system. 

This means that the systemd provider will be "suitable" on such a system, even though systemd cannot be used to manage services. The result of this is puppet displaying a bunch of services with the systemd provider which cannot actually be managed:

```
root@debian-8-jessie:~# puppet resource service --param provider
...
service { 'umountfs':
  ensure   => 'running',
  enable   => 'false',
  provider => 'debian',
}
...
service { 'syslog.service':
  ensure   => 'stopped',
  enable   => 'false',
  provider => 'systemd',
}
service { 'systemd-networkd.service':
  ensure   => 'stopped',
  enable   => 'false',
  provider => 'systemd',
}
service { 'umountnfs.service':
  ensure   => 'stopped',
  enable   => 'false',
  provider => 'systemd',
}
service { 'umountnfs.sh':
  ensure   => 'running',
  enable   => 'false',
  provider => 'debian',
}
...
```

The biggest issue comes when trying to manage a service which has both systemd and sysvinit bits, such as cron:

```
service { 'cron':
  ensure   => 'running',
  enable   => 'true',
  provider => 'debian',
}
service { 'cron.service':
  ensure   => 'stopped',
  enable   => 'false',
  provider => 'systemd',
}
```

If you try to do `puppet resource service cron ensure=stopped', puppet will actually try to use the systemd provider to stop the service, which will fail. Thus, it becomes impossible to stop the cron service. With this confine still in place, puppet won't try to use the systemd provider at all. In addition, doing `puppet resource service cron provider=systemd` will result in `Error: /Service[cron]: Provider systemd is not functional on this host` as it should, since systemd really isn't functional. 

@melissa, do you remember if this was the specific reason for adding that confine in your original commit?

ianw:
So another option that might work for everyone is to wrap this confine so it is Debian only?

whopper:
@ianw ah, I can't think of a reason that wouldn't work. @branan what do you think of that idea?

MikaelSmith:
This looks like a more general case where the Puppet run sets up the init system, and then wants to use the provider but it's already been excluded. There may be a more general solution to this problem.

margamanterola:
This fix fixed the chroot functionality for non-debian systems, but let it be broken for debian systems.  In particular, during the installation of a Debian system, with a puppet version that includes the :confine, all systemd resources fail with an error, when with previous puppet versions they would exit without doing anything, because of the policy-rc.d file that states that the system isn't functional.

smu-ggl:
Note: The better check to make sure systemd is in use on Debian systems would be to check if /sbin/init links to /lib/systemd/systemd (if systemd is used, /sbin/init is such a symlink, even in chroots, at least on Debian)

PR #4482, opened 2015-12-04
(PUP-5578) use exist? instead of '' to determine if scope key exists.

crayfishx:
This PR is the suggested fix for PUP-5578


thallgren:
Looks good but a unit test is failing. You'll need to fix the last test in `spec/unit/hiera/scope_spec.rb`. Ideally, there should be two tests. One to verify that `nil` is a valid value and another to validate that a missing key is indeed missing.

hlindberg:
Please make the commit comment start with `(HI-484)`

crayfishx:
@hlindberg @thallgren - commit renamed and spec tests improved to cover this....

PR #4483, opened 2015-12-04
(PUP-5502) Improve error messages for faulty hiera.yaml

thallgren:
Errors originating from syntax errors in the hiera.yaml were nested
and the error messages were repeated, resulting in a long blob of
unhelpful text where the original stack trace was lost.

This commit ensures that no message is repeated and that the
stack backtrace is retained.


hlindberg:
Merged. Should have gone to stable though. Not sure if this cherry picks over...

PR #4485, opened 2015-12-04
(PUP-5378) acceptance: port all-functions test to 3.x

er0ck:
:exclamation:To merger: this should be merged up to stable and master as well.  There will be conflicts in the current version in stable and master, but the version in this PR should be retained, conflicts ignored.  

This change adds a test for all parser functions.
This test calls all the parser functions that ship with puppet. It
ensures sane output in most cases. The remainder of the coverage for
functions lies in the spec tests.

[skip ci]

### DIFF
--- a/acceptance/tests/parser_functions/calling_all_functions.rb
+++ b/acceptance/tests/parser_functions/calling_all_functions.rb
@@ -4,8 +4,8 @@ test_name 'Calling all functions.. test in progress!'
 step 'Apply manifest containing all function calls'
 def manifest_call_each_function_from_array(functions)
   manifest = ''
-  # use index to work around puppet's imutable variables
-  # use variables so we can contatenate strings
+  # use index to work around puppet's immutable variables
+  # use variables so we can concatenate strings
   functions.each_with_index do |function,index|
     if function[:rvalue]
       manifest << "$pre#{index} = \"sayeth #{function[:name].capitalize}: Scope(Class[main]): \" "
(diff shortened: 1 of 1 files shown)


er0ck:
i forgot that all of our 4x functions work in 3x with future parser turned on.

so this test can stay in one piece, provided we can get it to work on 3.x/stable/master.

this is subtly difficult.
the PR as-is works for the functions that are not commented out.

@hlindberg @thallgren:
The examples we have to get working are:
include,hiera_include, etc: i can't seem to get these to find classes when running in apply mode:
sayeth Hiera_include: \nWarning: Scope(Class[main]): Could not look up qualified variable '::trusted';
i've tried this with store_configs and trusted_server_facts both turned off. and these work fine in 4.x...
i'm specifying our temp manifestpath on the commandline to puppet apply.

and any function that uses a regex for "expected":
in ruby 1.9.3 i just plain can not get these to be concatenated to strings properly where a match is still found. converting to_s and back seems to add a bunch of stuff that makes the match never succeed.
we end up with things like this in the expected string:
(?-mix:\\d+) which i guess don't get converted back to regex properly in minitest's assert_match
(Multiline, case Insensitve, eXtended) options turned on which should match properly. but i have a suspicion that assert_match is doing some funky conversion under the covers. i guess i'll dig into it tomorrow.

if you have any ideas for searching including classes in apply mode in 3.x, that'd be great

acceptance/tests/parser_functions/calling_all_functions.rb
```diff
@@ -0,0 +1,214 @@
+test_name 'Calling all functions.. test in progress!'
+
+# create single manifest calling all functions
+step 'Apply manifest containing all function calls'
+def manifest_call_each_function_from_array(functions)
+  manifest = ''
+  # use index to work around puppet's imutable variables
```
kylog: typo: immutable

acceptance/tests/parser_functions/calling_all_functions.rb
```diff
@@ -0,0 +1,214 @@
...
+  # use variables so we can contatenate strings
```
kylog: typo: concatenate

er0ck:
good catch on the typos, @kylog.  fixed.

PR #4486, opened 2015-12-07
(PUP-5582) Enforce no return in ensure blocks

kylog:
A return in an ensure block is ignored in ruby, so a return
statement in an ensure block is meaningless. This commit
removes the one usage of that in puppet and enforces this
going forward with rubocop.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,9 +25,8 @@ Lint/UnreachableCode:
 Lint/UselessComparison:
   Enabled: true
 
-# MAYBE useful - no return inside ensure block.
 Lint/EnsureReturn:
-  Enabled: false
+  Enabled: true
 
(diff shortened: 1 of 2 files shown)


MikaelSmith:
:+1:

joshcooper:
The specs are failing in a way that seems to rely on the ensure block returning nil:

```
3) Puppet configuration client when executing a catalog run should apply the catalog, send the report, and return nil if the post-run command fails
     Failure/Error: expect(@agent.run).to be_nil
       expected: nil
            got: 0
     # ./spec/unit/configurer_spec.rb:341:in `block (3 levels) in <top (required)>'
     # util/rspec_runner:39:in `run'
     # util/rspec_runner:53:in `<main>'
```

kylog:
Hmmm, I'll take a look.

PR #4487, opened 2015-12-08
(PUP-5061) Alter resurrection of trusted retain 'authenticated'

thallgren:
The current resurrection of the 'trusted' hash changes the
authentication to `false`. This commit removes that behavior so that
the hash is left untouched.


kylog:
+1. From memory this is all I thought we needed to do now.

hlindberg:
Well, we still have to fix the various node termini or we cannot deliver on the promise to always restore what was restored as none of the termni except puppet db stores the trusted information. Need a different ticket for that now. (Lower prio though).

PR #4489, opened 2015-12-10
(PUP-5063) Ensure that 4x-functions are profiled

thallgren:
This commit ensures that calls to 4x functions are profiled. The
description of the profiling data is simply the method name since
I wanted to avoid unnecessary string formatting. The entry is
easy enough to understand anyway given that the metric_id is
:function.

### DIFF
--- a/lib/puppet/pops/functions/function.rb
+++ b/lib/puppet/pops/functions/function.rb
@@ -89,7 +89,11 @@ class Puppet::Pops::Functions::Function
     raise ArgumentError, "Function #{self.class.name}(): cannot call function '#{function_name}' - no loader specified" unless the_loader
 
     func = the_loader.load(:function, function_name)
-    return func.call(scope, *args, &block) if func
+    if func
+      Puppet::Util::Profiler.profile(function_name, [:functions, function_name]) do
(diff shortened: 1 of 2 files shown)


hlindberg:
Looks good, ready to go, but since there are no tests and no real specification of what exactly is supposed to be logged - I am waiting on confirmation from reporting user before merging.

hlindberg:
got thumbs up  merging

lib/puppet/pops/functions/function.rb
```diff
@@ -89,7 +89,11 @@ def internal_call_function(scope, function_name, args, &block)
     raise ArgumentError, "Function #{self.class.name}(): cannot call function '#{function_name}' - no loader specified" unless the_loader
 
     func = the_loader.load(:function, function_name)
-    return func.call(scope, *args, &block) if func
+    if func
+      Puppet::Util::Profiler.profile(function_name, [:functions, function_name]) do
```
cprice404: Is this the same way we generate the metric IDs for the old functions?  I ask because in Puppet Server we surface metrics about function calls in a few places such as our status endpoint.  As long as this naming convention matches what we were doing for the old functions then these should Just Work.

PR #4490, opened 2015-12-10
(PUP-5595) Do not retrieve generated resources until they are applied

branan:
PUP-1963 changed resource generation such that resources were added to
the *catalog*, not just the final relationship graph. This required
the Type instances returned by the `generate` method to be converted
to Resource instances.

Unfortunately, the `to_resource` method calls `retrieve`, presumably
for the sake of the RAL. This would cause resources to be retrieved at
generation time, rather than at application time.

Instead of using `to_resource`, the additional resource generator now
constructs a Resource instance using its own logic, including only the
parameters that were used when the Type instance was created.

### DIFF
--- a/lib/puppet/transaction/additional_resource_generator.rb
+++ b/lib/puppet/transaction/additional_resource_generator.rb
@@ -138,10 +138,10 @@ class Puppet::Transaction::AdditionalResourceGenerator
   def add_generated_directed_dependency(parent, child, label=nil)
     if parent.depthfirst?
       source = child
-      target = parent.to_resource
+      target = parent
     else
       source = parent
-      target = child.to_resource
+      target = child
     end
 
     # For each potential relationship metaparam, check if parent or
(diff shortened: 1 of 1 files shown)


lib/puppet/transaction/additional_resource_generator.rb
```diff
@@ -170,8 +170,11 @@ def add_generated_directed_dependency(parent, child, label=nil)
     }
 
     if not edge_exists
+      parameters = target.original_parameters
+      resource = ::Puppet::Resource.new(target.class, target.title, :parameters => parameters)
```
branan: I was having issues with namespacing here, and being totally explicit seemed the best way to move forward

reidmv:
I applied this patch locally and verified that it resolves PUP-5595.

lib/puppet/transaction/additional_resource_generator.rb
kylog: Style (I think) nit, but is there any reason not to forgo the local and do something like:

```
      resource = ::Puppet::Resource.new(target.class, target.title, 
                                        :parameters => target.original_parameters)
```

kylog:
Is there anything (not insane) we can do at spec test level to enforce this doesn't regress?

lib/puppet/transaction/additional_resource_generator.rb
branan: I had split it for debug prints, and never put it back together. I don't have strong opinions about the format here, I'm hapy to change it.

branan:
@kylog I can add a spec test that generate doesn't call retrieve, but that feels very "bug-fix-specific-unit-test" to me. I feel like these old low-level puppet classes are a mess, and any given unit test probably won't actually prevent real regressions (which will probably happen in another code path)

I should definitely add a comment as to why to_resource isn't being used, though

kylog:
@branan :+1: on adding an explanatory comment rather than a test-the-implementation-not-the-interface spec test.

branan:
@kylog updated with comment, and addressed your style nit

kylog:
:+1: 

Will let travis go green.

kylog:
Travis seems borked. Let's try Jenkins.

PR #4494, opened 2015-12-11
(PUP-5594) Fixes yum check-updates private method

petems:
* Previous code was using `#warn`
* Would cause error: `private method `warn' called for Puppet::Type::Package::ProviderYum:Class`
* Changed to correct method `warning`
* Updated spec


petems:
:+1:

MikaelSmith:
Merged up to stable for Puppet 4.3.2.

PR #4495, opened 2015-12-11
(PUP-4744) Fixes `reposdir` split by whitespace

petems:
* `reposdir` can be split by whitespace - http://yum.baseurl.org/gitweb?p=yum.git;a=blob;f=yum/config.py;h=97e5e3d3b063294a96227a67cce43a9655cd3aa6;hb=30c993c19b8b85f44e7cec436484862ab55a6961#l171
- Puppet previously only split by commas
- Now changed to a regex of [,|\s] which will catch whitespace and commas
- Added new spec for this new regex

Ticket https://tickets.puppetlabs.com/browse/PUP-4744

### DIFF
--- a/lib/puppet/provider/yumrepo/inifile.rb
+++ b/lib/puppet/provider/yumrepo/inifile.rb
@@ -74,9 +74,12 @@ Puppet::Type.type(:yumrepo).provide(:inifile) do
   def self.reposdir(conf='/etc/yum.conf', dirs=['/etc/yum.repos.d', '/etc/yum/repos.d'])
     reposdir = find_conf_value('reposdir', conf)
     # Use directories in reposdir if they are set instead of default
-    dirs = reposdir.split(",").map(&:strip) if reposdir
-    
-
+    if reposdir
+      # Follow the code from the yum/config.py
+      dirs = reposdir.gsub!("\n", ' ')
+      dirs = reposdir.gsub!(',', ' ')
+      dirs = reposdir.split
+    end
     dirs.select! { |dir| Puppet::FileSystem.exist?(dir) }
     if dirs.empty?
       Puppet.debug('No yum directories were found on the local filesystem')
--- a/spec/unit/provider/yumrepo/inifile_spec.rb
+++ b/spec/unit/provider/yumrepo/inifile_spec.rb
@@ -249,6 +249,22 @@ describe Puppet::Type.type(:yumrepo).provider(:inifile) do
       expect(described_class.reposdir('/etc/yum.conf')).to include("/etc/yum/extra.repos.d")
     end
 
+    it "includes the directory if the value is split by whitespace" do
+      Puppet::FileSystem.expects(:exist?).with("/etc/yum/extra.repos.d").returns(true)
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/yumrepo/inifile.rb
```diff
@@ -74,8 +74,7 @@ def self.prefetch(resources)
   def self.reposdir(conf='/etc/yum.conf', dirs=['/etc/yum.repos.d', '/etc/yum/repos.d'])
     reposdir = find_conf_value('reposdir', conf)
     # Use directories in reposdir if they are set instead of default
-    dirs = reposdir.split(",").map(&:strip) if reposdir
-    
+    dirs = reposdir.split(/[,|\s]/).map(&:strip) if reposdir
```
joshcooper: This will only match a single white space character, perhaps we should match multiple spaces (but not multiple commas)?

petems: Are multiple spaces valid? I couldn't work it out from looking at the yum code.

joshcooper: Good question, I think it does split and strip multiple whitespace characters:

```
>>> "foo \n  \r bar".split()
['foo', 'bar']
```

Also https://docs.python.org/3/library/stdtypes.html#str.split

petems: Hmm does `\s` match newlines? Should probably catch that too...

joshcooper: It looks like they replace `\n` and `,`, with ` `, and then split on that. I'm not sure if our `find_conf_value` will return a multiline string?

petems: Fixed to catch newlines and split on them, added spec for it

lib/puppet/provider/yumrepo/inifile.rb
```diff
@@ -74,8 +74,7 @@ def self.prefetch(resources)
   def self.reposdir(conf='/etc/yum.conf', dirs=['/etc/yum.repos.d', '/etc/yum/repos.d'])
     reposdir = find_conf_value('reposdir', conf)
     # Use directories in reposdir if they are set instead of default
-    dirs = reposdir.split(",").map(&:strip) if reposdir
-    
+    dirs = reposdir.split(/[,|\s|\n]/).map(&:strip) if reposdir
```
joshcooper: I'm not sure this is sufficient, maybe we can tighten this up?

```
irb(main):001:0> "foo    bar".split(/[,|\s|\n]/).map(&:strip)
=> ["foo", "", "", "", "bar"]
```

MikaelSmith:
:+1:

MikaelSmith:
This could go against stable.

joshcooper:
In reading the ticket, it sounds like several people are affected by this. :+1: to retargeting for stable (sorry @petems I should have noticed that sooner)

petems:
@joshcooper No problem, do you want me to change the target?

kylog:
@petems yeah. You'll need to close this PR and open another. (I wish wish wish that github allowed changing target branches on open PRs. Sigh.)

PR #4496, opened 2015-12-11
(PUP-5598) Ensure that keys starting with lookup_options fail lookup

thallgren:
The `lookup_options`special key should not be found using a normal
lookup and neither should any of its subkeys when a dotted key is
used. This commit ensures that this holds true and also adds a new
explanation of 'Invalid key' to be used when such lookup attempts are
made.


thallgren:
Closing this PR since this should go to 3.x

PR #4497, opened 2015-12-11
(PUP-5590) Detect duplicate parameter names

thallgren:
Adds validation logic to detect duplicate parameter names in host
class definitions, defines, and EPP templates.


thallgren:
Supersedes #4493.

PR #4499, opened 2015-12-14
(PUP-5553) Add Ubuntu 15.10 support to systemd

phervieux:
Add ubuntu 15.10 in the library defining system using systemd.


joshcooper:
Thank you for your contribution @phervieux! We had a few questions/comments:

* Can the confine be changed to `>=` so that we don't have to repeat this process when 16.04 comes out?
* Could you retarget this PR for stable? We'd like to add support for 15.10 as soon as possible. You'll need to close this PR and open a new one for stable.
* Can you drop the extra merge commit 5729b46? You should be able to just do `git reset --hard HEAD^`. If there are conflicts you'll want to rebase your PR on the latest `stable` branch.
* Can you add a unit test for this in https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/service/systemd_spec.rb#L107, similar to the one for 15.04?

joshcooper:
@phervieux I went ahead and pulled your commit into a new PR #4504, preserving your authorship on the original commit. Thank you for your contribution!

PR #4501, opened 2015-12-14
(PUP-5609) Protect shared state with filesystem lock

dalen:
by apply and agent).

### DIFF
--- a/lib/puppet/application/apply.rb
+++ b/lib/puppet/application/apply.rb
@@ -6,6 +6,9 @@ class Puppet::Application::Apply < Puppet::Application
   require 'puppet/util/splayer'
   include Puppet::Util::Splayer
 
+  require 'puppet/util/locker'
+  include Puppet::Util::Locker
+
   option("--debug","-d")
   option("--execute EXECUTE","-e") do |arg|
     options[:code] = arg
@@ -164,11 +167,15 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
   end
 
   def run_command
-    if options[:catalog]
-      apply
-    else
-      main
+    lock do
+      if options[:catalog]
+        apply
+      else
+        main
+      end
     end
+  rescue Puppet::LockError
+    Puppet.notice "Run of #{client_class} already in progress; skipping  (#{lockfile_path} exists)"
   end
 
   def apply
--- a/lib/puppet/agent.rb
+++ b/lib/puppet/agent.rb
@@ -4,8 +4,8 @@ require 'puppet/error'
 # A general class for triggering a run of another
 # class.
 class Puppet::Agent
-  require 'puppet/agent/locker'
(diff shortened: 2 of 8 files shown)


dalen:
Finally cleaned up this little patch I originally wrote ~2 years ago (and reported the bug) :)

MikaelSmith:
My concern would be that this is a breaking change. Was there a puppet-dev thread from 2 years ago talking about whether blocking running `agent` and `apply` at the same time would break anyone's workflow?

dalen:
If people actually have that as part of their workflow they also have corrupt state files as part of their workflow.

MikaelSmith:
It seems like a bad idea, just checking. If they want to make it happen, they'd need to use different vardirs.

dalen:
Yup. And if they use different vardirs they also get different lock files, so then it wouldn't be an issue even with this patch.

MikaelSmith:
:+1:

lib/puppet/application/apply.rb
```diff
@@ -164,11 +167,15 @@ def app_defaults
...
-      main
+    lock do
+      if options[:catalog]
+        apply
+      else
+        main
+      end
     end
+  rescue Puppet::LockError
+    Puppet.notice "Run of #{client_class} already in progress; skipping  (#{lockfile_path} exists)"
```
MikaelSmith: Oops, this isn't defined in `apply`. Caught by an acceptance run at https://jenkins.puppetlabs.com/job/platform_puppet_intn-van-sys_master/24/SLAVE_LABEL=beaker,TEST_TARGET=redhat7-64a/console.

Which means our acceptance runs have multiple puppet apply happening or are failing to cleanup the lock file. That's interesting.

MikaelSmith:
This potentially messes with using Puppet to stop the puppet service. We have acceptance tests that do that. @jpartlow may know of other cases where we run apply to manage services.

jpartlow:
@MikaelSmith various PE installer scripts make use of puppet resource to manage the puppet service.

MikaelSmith:
That doesn't use this code path, so should be fine.

MikaelSmith:
Going to revert this and do a little work. The `tests/resource/service/puppet_mcollective_service_managament` acceptance test is showing some interesting behavior: the file doesn't seem to dissappear immediately, but hitting the LockError still returns exit code 0 so the tests appear to pass.

lib/puppet/application/apply.rb
MikaelSmith: I'd recommend we add a spec test that expects lock to throw an error, and ensure we get a notice and return exit(1).

This should also be updated so that we `exit 1` (equivalent to https://github.com/puppetlabs/puppet/blob/master/lib/puppet/application/agent.rb#L360-L361).

MikaelSmith:
Opened a PR to revert. Need a few changes and to investigate why the acceptance test fails.

PR #4503, opened 2015-12-15
(PUP-5212) Pip package honors HTTP proxy

MikaelSmith:
The pip package provider previously did not work behind a http proxy. It
will now honor system proxy settings for querying latest packages from
`pypi.python.org`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+sudo: false
 bundler_args: --without development
 script: "bundle exec rake \"parallel:spec[2]\""
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: ruby
 sudo: false
+before_install:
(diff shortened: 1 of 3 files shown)


kylog:
Hmm, the travis fail on the first commit on ruby 1.9.3 (https://travis-ci.org/puppetlabs/puppet/builds/97052540) looks like it may be https://github.com/bundler/bundler/issues/3558, which (I think) @DavidS mentioned running into elsewhere.

I didn't run this down so don't know why it's intermittent, or really what's going on.

Fix might be to force a newer version of bundler with:

```
before_install:
  - gem install bundler -v 1.10
```

Mayyyyybe. Not sure.

MikaelSmith:
I'll give it a try.

DavidS:
Yes, I ran into this and workarounded it the same as here (adding a link to the bug report). It worked for me, but it is a HACK.

PR #4505, opened 2015-12-16
(PUP-5615) Deprecate unused options to PRC.apply

kylog:
@joshcooper or @MikaelSmith: followup on the deprecation topic that we uncovered when looking into tags. Not sure what legwork we might want to do to verify that there aren't external consumers of this API who might care about this deprecation.

### DIFF
--- a/lib/puppet/resource/catalog.rb
+++ b/lib/puppet/resource/catalog.rb
@@ -195,6 +195,14 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
   def apply(options = {})
     Puppet::Util::Storage.load if host_config?
 
+    # The only option still supported is report.
+    # Issue a deprecation warning for any others.
+    deprecated_options = options.keys - [:report]
+    if !deprecated_options.empty?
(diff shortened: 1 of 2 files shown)


lib/puppet/resource/catalog.rb
```diff
@@ -195,6 +195,14 @@ def alias(resource, key)
   def apply(options = {})
     Puppet::Util::Storage.load if host_config?
 
+    # The only option still supported is report.
+    # Issue a deprecation warning for any others.
+    deprecated_options = options.keys - [:report]
+    if !deprecated_options.empty?
+      Puppet.deprecation_warning("Puppet::Resource::Catalog.apply no longer " \
+                                 "takes the #{deprecated_options} option(s)")
```
joshcooper: It looks like several applications have `options` that get passed blindly to the configurer and then to this method, e.g. `:pluginsync`, `:network_device`, ...

kylog:
Hmm, I thought I had run down the relevant code paths, but apparently not :(

Anyway, this little deprecation is super low-priority so closing this PR.

PR #4507, opened 2015-12-16
(PUP-5616) fixes to improve error display

jorhett:
- adds error display for device commands, which used to fail silently
- removes duplicate command input from the error output

This also removes some tests which caused never-achieve-convergence
as it was dropping device properties that could be supplied as input

### DIFF
--- a/lib/puppet/util/network_device/cisco/device.rb
+++ b/lib/puppet/util/network_device/cisco/device.rb
@@ -47,7 +47,13 @@ class Puppet::Util::NetworkDevice::Cisco::Device < Puppet::Util::NetworkDevice::
   end
 
   def execute(cmd)
-    transport.command(cmd)
+    transport.command(cmd) do |out|
+      if out =~ /^%/mo or out =~ /^Command rejected:/mo
+        # strip off the command just sent
+        error = out.sub(cmd,'')
+        Puppet.err "Error while executing '#{cmd}', device returned: #{error}"
+      end
+    end
   end
 
   def login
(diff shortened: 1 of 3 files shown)


jorhett:
I just fixed the test failure, and added another test for VLAN description validity. The Cisco specs all say alphanumeric no spaces or special characters, and our regex requires matching `\w` so this error is more useful than the previous situation which would constantly fail to converge.

whopper:
Thanks @jorhett - we're having our PR triage tomorrow, and I've added this to the discussion list.

whopper:
Ping @LuvCurves, we're looking for a way to review this. Do you have any thoughts on this change?

LuvCurves:
Looks good to me.  We now have VMs in the VMpooler that can be used for verification.

MikaelSmith:
It would be nice to see more testing on this, but I'm not sure what it would look like. @LuvCurves if you think this is correct behavior go ahead and merge it.

joshcooper:
@binford2k This didn't make it in time for 4.4.0. You could retarget stable for 4.4.1 or we can merge to master for 4.5.0.

binford2k:
ping @jorhett

jorhett:
Not sure what you want from me, as I can't affect your release targets can I?

Do I need to merge against a different branch?

hlindberg:
@jorhett If you want this to go into 4.1.0 retarget the PR against stable, otherwise it will go into 4.5.0. I think @joshcooper  is saying it is up to you.

jorhett:
Forgive my ignorance, but I don't know what "target against stable" means in terms of an actionable process. It's been different at every firm I've worked with, and you guys are much process heavier than most.

As far as I know the rest API feature for this still doesn't work, so I must submit a new PR yes?

hlindberg:
"target against stable"
On your branch: git rebase stable
close this PR and open a new that "targets" stable.

We are doing release work a.t.m, so will take a day or two before being ready to merge to stable.

branan:
I think at this point this has missed the boat for 4.4.0. Leaving it targeted at master for 4.5.0 is the way to go, and we should be able to get this merged in as soon as we have bandwidth post-4.4.40 release.

hlindberg:
@jorhett which means that the way it is now targeting master branch is correct (it is for 4.5.0 now).

jorhett:
Sounds good. Although it has missed 3 releases now, so what can we do to get this included in the next?

hlindberg:
Hint: there is a big green "Merge pull request" button on this page. I guess someone should press that :-)
Joking aside - everyone is busy with release of 4.4.0 atm. And it is unclear from the  thread above if everyone involved was +1 on merging.

PR #4508, opened 2015-12-16
(PUP-5617) enable keepalive on the TCP socket opened with the master

mfournier:
This will allow the socket with the master to get closed when the
network connection breaks unexpectedly, and the agent to terminate with
an error, rather than waiting forever.

According to linux's tcp(7) manpage, such zombies connections get closed
by default after a bit more than 2 hours of idleness. Configuring this
is possible but not considered portable.

### DIFF
--- a/lib/puppet/network/http/pool.rb
+++ b/lib/puppet/network/http/pool.rb
@@ -81,8 +81,17 @@ class Puppet::Network::HTTP::Pool
 
       Puppet.debug("Starting connection for #{site}")
       http.start
+      setsockopts(http.instance_variable_get(:@socket))
       http
     end
+  end
+
+  # Set useful socket option(s) which lack from default settings in Net:HTTP
+  #
+  # @api private
+  def setsockopts(netio)
+    socket = netio.io
+    socket.setsockopt(Socket::SOL_SOCKET, Socket::SO_KEEPALIVE, true)
   end
 
   # Release a connection back into the pool.
--- a/spec/unit/network/http/connection_spec.rb
+++ b/spec/unit/network/http/connection_spec.rb
@@ -163,6 +163,7 @@ describe Puppet::Network::HTTP::Connection do
 
       Net::HTTP.any_instance.stubs(:start)
       Net::HTTP.any_instance.stubs(:request).returns(httpok)
+      Puppet::Network::HTTP::Pool.any_instance.stubs(:setsockopts)
 
       connection.get('request')
     end
--- a/spec/unit/network/http/pool_spec.rb
+++ b/spec/unit/network/http/pool_spec.rb
@@ -117,8 +117,13 @@ describe Puppet::Network::HTTP::Pool do
       # On windows, Socket.getsockopt() doesn't return exactly the same data
       # as an equivalent Socket::Option.new() statement, so we strip off the
       # unrelevant bits only on this platform.
+      #
       # To make sure we're not voiding the test case by doing this, we check
       # both with and without the keepalive bit set.
+      #
+      # This workaround can be removed once all the ruby versions we care about
+      # have the patch from https://bugs.ruby-lang.org/issues/11958 applied.
+      #
       keepalive   = Socket::Option.bool(:INET, :SOCKET, :KEEPALIVE, true).data
       nokeepalive = Socket::Option.bool(:INET, :SOCKET, :KEEPALIVE, false).data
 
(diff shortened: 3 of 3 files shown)


mfournier:
@joshcooper I've just updated the patch to make the tests pass again (thanks to @mcanevet for helping me with this !) and also added a test case for this new socket option.

I hope the way this is done is OK. If not, please let me know how to organise things correctly :) Thanks !

kylog:
@mfournier I see this is failing in Windows CI like so:

```
1) Puppet::Network::HTTP::Pool when yielding a connection sets keepalive bit on network socket
     Failure/Error: expect(s.getsockopt(Socket::SOL_SOCKET, Socket::SO_KEEPALIVE).bool).to eq(true)
 
     TypeError:
       size differ.  expected as sizeof(int)=4 but 1
     # ./spec/unit/network/http/pool_spec.rb:117:in `bool'
     # ./spec/unit/network/http/pool_spec.rb:117:in `block (3 levels) in <top (required)>'
```

The `TypeError` was a little surprising to me, so I dug into it a little bit. It looks like one gets that error just from `s.getsockopt(Socket::SOL_SOCKET, Socket::SO_KEEPALIVE).bool` (not from the `expect ... to eq(true)` which was my original assumption).

This makes me wonder if Socket::Option is really supported (or how correctly) by Ruby on Windows. This will take a little digging I think.

mfournier:
Thanks for taking a look @kylog !

I just changed the spec file to avoid casting to a bool. Let's see how this goes on windows...

kylog:
Still failing.

This is going to take more Windows chops than I have unfortunately. Ping @Iristyle or @glennsarti.

kylog:
I'm :+1: on the idea, but we do need to figure out whether this is supported on Windows or not and update accordingly.

joshcooper:
I think the travis failure is ruby-head (2.4.0dev) related, we just merged a travis config change to use 2.3.0 instead, let me close and reopen this PR to rerun it. Appveyor will still fail for now.

joshcooper:
Well that didn't work, might need to rebase to clear up the travis issue?

KaysLounge:
Stop 

Sent from Outlook Mobile




On Tue, Jan 5, 2016 at 11:13 PM -0800, "Josh Cooper" <notifications@github.com> wrote:












I think the travis failure is ruby-head (2.4.0dev) related, we just merged a travis config change to use 2.3.0 instead, let me close and reopen this PR to rerun it. Appveyor will still fail for now.

mfournier:
Ok, I think I've nailed it !

I reworked the last commit (see commit message + comments in the code for the details) and rebased my branch on current master.

NB: both issues we encountered with the tests on windows look pretty much like a bug in ruby itself, or at least an unfortunate behaviour discrepency on this platform. So this makes me wonder if ruby actually sets the keepalive flag on the socket. Testing equivalent code in irb on windows doesn't raise an error, but I don't know this platform well enough to check if the OS honours ruby's system call (basically, run an equivalent to linux's `netstat -nto`).

To recap:
 * the code works as expected on linux: I had this running on my fleet for a couple of weeks now
 * windows itself does support tcp keepalives, obviously
 * we have a workaround in the tests to take care of ruby's `Socket.getsockopt()` odd behaviour on windows
 * *setting* socket options (which is all is used in the application code) seems to work fine on windows. We only had problems when *getting* socket options, which is only done in the specs.

mfournier:
For the record, I just opened https://bugs.ruby-lang.org/issues/11958

kylog:
@mfournier great work and thanks for sorting that all out! 

It would be nice if we could verify that Windows is honoring ruby's `setsockopt` here, esp since there definitely seems to be a bug with ruby's `getsockopt` on Windows. I'm going to give a couple days for some of the more knowledgeable windows developers mentioned above to give ideas on how we could verify that.

But pending that I'm :+1:

mfournier:
@kylog, @joshcooper: so the ruby bug was confirmed (and fixed). I amended the comment in the specfile to mention it, so the workaround can be removed in the future.

whopper:
Thanks @mfournier - we think this is looking good, and just want to spend a bit of time poking it in Windows before merging. Stay tuned!

Iristyle:
Looks like Ruby acknowledged the bug and it landed in https://bugs.ruby-lang.org/projects/ruby-trunk/repository/revisions/53561/diff

kylog:
Thanks for all your work on this @mfournier! Merging, so this should be released in puppet 4.4.0.

PR #4407, opened 2015-11-04
(PUP-5146) Update platform field to use version numbers

joshcooper:
Previously, beaker tried to install puppet-agent on OSX 10.11 twice
because the 'el' regex matched the elcapitan platform. When the 'el'
regex was made to be more specific, beaker added PermitUserEnvironment
to the pre 10.11 location, because beaker expects platform to contain
version numbers, not codenames.

This commit changes all OSX host configs to version numbers as
codenames, and makes the 'el' regex more specific, as 'el' could occur
in a future OS name.

It also backports the osx-1011 node definition to stable, as we haven't
merged down yet.

For consistency, this commit also updates the platform identifier for
smoke testing gem installs.

PR #4510, opened 2015-12-16
(maint) Don't allow elcapitan to match redhat el regex

joshcooper:
We were using `el` in various redhat related regexs, but `elcapitan`
unfortunately matches. This commit changes the regex to be more
specific to avoid false positives.


joshcooper:
This is a follow up from https://github.com/puppetlabs/puppet/pull/4407#issuecomment-153855464

kylog:
Ah right! Thanks for circling back to this one.

PR #4511, opened 2015-12-17
(PUP-5465) Add missing support for Resource References in collection

kylog:
The support for resource references in collector search expressions is
missing. This is a documented aspect of resource collectors and
previously worked in Puppet 3.

This commit adds support for resource references in collection search
expressions.

As part of the behavior, the new code converts from the POPS object into
the Puppet 3 style objects as the other internal mechanisms being
interacted with still work in with those objects.

### DIFF
--- a/lib/puppet/pops/evaluator/collector_transformer.rb
+++ b/lib/puppet/pops/evaluator/collector_transformer.rb
@@ -125,6 +125,14 @@ protected
     end
   end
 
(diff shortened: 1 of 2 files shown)


kylog:
@joshcooper you suggested this backport over in https://github.com/puppetlabs/puppet/pull/4398#issuecomment-159442360. Makes sense to me.

PR #4512, opened 2015-12-17
(PUP-5618) Ensure module data provider isolation.

thallgren:
The module data provider was a singleton that was cached in the scope
of the environment lifecycle. Yet, it was configured on a per module
basis. The first configuration of the first module to be configured
was cached and subsequently used for all other modules.

This commit ensures that each module that provides data has a unique
module data provider instance.


hlindberg:
@thallgren I think you wanted this to go to stable.

hlindberg:
for completeness - this was cherry picked to stable

PR #4515, opened 2015-12-17
(DOCUMENT-448) Specify format for 'ensure' value when using a version.

rnelson0:
https://tickets.puppetlabs.com/browse/DOCUMENT-448

### DIFF
--- a/lib/puppet/type/package.rb
+++ b/lib/puppet/type/package.rb
@@ -74,6 +74,12 @@ module Puppet
         value. On packaging systems that manage configuration files separately
         from "normal" system files, you can uninstall config files by
         specifying `purged` as the ensure value. This defaults to `installed`.
+
+        Version numbers must match the full version to install, including
+        release if the provider uses a release moniker. Ranges or semver
+        patterns are not accepted except for the `gem` package provider. For
+        example, to install the bash package from the rpm
+        `bash-4.1.2-29.el6.x86_64.rpm`, use the string `'4.1.2-29.el6'`.
       EOT
 
       attr_accessor :latest


lib/puppet/type/package.rb
```diff
@@ -74,6 +74,11 @@ module Puppet
         value. On packaging systems that manage configuration files separately
         from "normal" system files, you can uninstall config files by
         specifying `purged` as the ensure value. This defaults to `installed`.
+
+        Version numbers must match the full version to install, including
+        release if the provider uses a release moniker. Ranges or semver
+        patterns are not accepted. E.g. to install the rpm
```
gguillotte: For clarity, I'd replace "E.g." with "For example,".

gguillotte:
Travis is failing because the commit starts with a DOCUMENT project ticket. Can you amend the commit to replace (DOCUMENT-448) with (docs), and note the ticket number in the commit description instead?

One other suggestion, then :+1: from me.

rnelson0:
@gguillotte Done and done! Just waiting on travis.

rnelson0:
Yay, tests pass this time!

lib/puppet/type/package.rb
```diff
@@ -74,6 +74,11 @@ module Puppet
         value. On packaging systems that manage configuration files separately
         from "normal" system files, you can uninstall config files by
         specifying `purged` as the ensure value. This defaults to `installed`.
+
+        Version numbers must match the full version to install, including
+        release if the provider uses a release moniker. Ranges or semver
+        patterns are not accepted. For example, to install the rpm
```
joshcooper: minor nit, the `gem` package provider does support ranges and semver constraints. This was added in https://tickets.puppetlabs.com/browse/PUP-4131. AFAIK none of the other providers support them.

gguillotte: Should we note provider-specific exceptions in the type reference? Or couch that statement with something along the lines of "... unless the package provider explicitly allows it"?

joshcooper: not sure... for comparison the `source` parameter does https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/package.rb#L360-L363

rnelson0:
I added a second commit to address the `gem` provider. If this is better, let me know and I'll rebase, otherwise I'll ditch it in favor of something better.

gguillotte:
@rnelson0 @joshcooper If `gem` is the only exception, then I think all that needs to change over the original commit is to mention that as briefly as possible: "Ranges or semver patterns are not accepted **except for the `gem` package provider**."

Otherwise I'm :+1: for rebasing and merging this..

rnelson0:
Okay, I think I got this now!

gguillotte:
I'm good with this. I can't seem to merge it, though, GH is saying the branch is out of date with `master`. :/ I'm having problems with a different PR that make me think there might be an issue on my end or with GH. CC @nfagerlund

gguillotte:
Got the merge button and merged it! Thanks @rnelson0!

PR #4516, opened 2015-12-17
(PUP-5625) Addressed some issues with the allow_dup_csrs acceptance test

camlow325:
This commit addresses a few issues seen when running the
`ticket_3360_allow_duplicate_csr_with_option_set.rb` test in a local
beaker run:

1) Use `refute_equal` instead of `assert_not_equals` for negative
   comparisons.

2) Changed the logic a bit around the iteration of certs on the master
   to ensure that only the csrs/certs for the agents under test (and not
   unrelated files that might be on the CA) will be used.

3) Validate that the master csr can be reissued - for the case that the
   master and agent are the same node.

4) Minor changes to the regexs used to pick the list of agent csrs/certs
   from the output of the `puppet cert` command.

### DIFF
--- a/acceptance/tests/ticket_3360_allow_duplicate_csr_with_option_set.rb
+++ b/acceptance/tests/ticket_3360_allow_duplicate_csr_with_option_set.rb
@@ -3,9 +3,15 @@ test_name "#3360: Allow duplicate CSR when allow_duplicate_certs is on"
 agent_hostnames = agents.map {|a| a.to_s}
 
 with_puppet_running_on(master, {'master' => {'allow_duplicate_certs' => true}}) do
+  agents_with_cert_name = {}
   agents.each do |agent|
-    step "Generate a certificate request for the agent"
+    step "Collect fqdn for the agent"
     fqdn = on(agent, facter("fqdn")).stdout.strip
+    agents_with_cert_name[fqdn] = agent
+  end
+
+  agents_with_cert_name.each do |fqdn, agent|
+    step "Generate a certificate request for the agent"
     on(agent, puppet("certificate generate #{fqdn} --ca-location remote --server #{master}"))
   end
 
@@ -15,14 +21,16 @@ with_puppet_running_on(master, {'master' => {'allow_duplicate_certs' => true}}) 
 
   old_certs = {}
   original_certs.stdout.each_line do |line|
-    if line =~ /^\+ (\S+) \((.+)\)$/
+    if line =~ /^\+ \"(\S+)\" \(?(.+)\)?$/ && agents_with_cert_name[$1]
       old_certs[$1] = $2
       puts "old cert: #{$1} #{$2}"
     end
   end
 
-  agents.each do |agent|
-    fqdn = on(agent, facter("fqdn")).stdout.strip
+  assert_equal(agents.count, old_certs.count,
+               "Expected original number of agent csrs on master to equal number of agents")
+
+  agents_with_cert_name.each do |fqdn, agent|
     step "Make another request with the same certname"
     on(agent, puppet("certificate generate #{fqdn} --ca-location remote --server #{master}"))
   end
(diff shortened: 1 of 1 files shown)


camlow325:
@er0ck, curious to get your feedback on this one.

acceptance/tests/ticket_3360_allow_duplicate_csr_with_option_set.rb
```diff
@@ -15,14 +21,16 @@
...
       puts "old cert: #{$1} #{$2}"
     end
   end
 
-  agents.each do |agent|
-    fqdn = on(agent, facter("fqdn")).stdout.strip
+  assert_equal(agents.count, old_certs.count,
+               "Expected original number of agent csrs on master to equal number of agents")
+
+  agents_with_cert_name.each do |fqdn, agent|
```
er0ck: good catch.

acceptance/tests/ticket_3360_allow_duplicate_csr_with_option_set.rb
```diff
@@ -33,18 +41,19 @@
...
+  assert_equal(agents.count, new_certs.count,
+               "Expected new number of agent csrs on master to equal number of agents")
+
   # using the agent name as the key may cause errors;
   # agent name from cfg file is likely to have short name
   # where certs might be signed with long names.
   old_certs.each_key { |key|
-    next if key.include? master # skip the masters cert, only care about agents
-    assert_not_equal(old_certs[key], new_certs[key], "Expected #{key} to have a changed key")
+    refute_equal(old_certs[key], new_certs[key], "Expected #{key} to have a changed key")
```
joshcooper: Why aren't we skipping the master's cert (or are you wanting to test the agent on the master)?

Why is `assert_not_equal` not good?

er0ck:
LGTM.

acceptance/tests/ticket_3360_allow_duplicate_csr_with_option_set.rb
camlow325: > Why aren't we skipping the master's cert (or are you wanting to test the agent on the master)?

Yeah, I was wanting to test the agent on the master.  Seemed useful given that we often just run with that matrix in CI?

> Why is assert_not_equal not good?

I made a reference about that in the JIRA ticket.  I found that when running the beaker tests against a local VM that I'd get an exception, leading the test to fail, because `assert_not_equal` could not be found:

````
Test Case ./ruby/puppet/acceptance/tests/ticket_3360_allow_duplicate_csr_with_option_set.rb reported: #<RuntimeError: PuppetAcceptance::DSL::Helpers.with_puppet_running_on failed (check backtrace for location) because: undefined method `assert_not_equal' for #<Beaker::TestCase:0x007f97945fd218>
./ruby/puppet/acceptance/tests/ticket_3360_allow_duplicate_csr_with_option_set.rb:48:in `block (3 levels) in run_test'
````

It didn't look to me like `assert_not_equal` was defined in the minitest library, which I think is where we get most of these assertion helpers from, right?  So I'm not entirely sure why this hasn't blown up in our CI pipeline.  I was speculating that maybe because some other gems alias `assert_not_equal` to `refute_equal` that that maybe one of those was implicitly executed before those tests run in standard CI, not sure.  I was thinking `refute_equal` would be better since it is actually defined in the minitest that we use?

If you have more insight into what this problem may be, though, happy to go with your recommendation instead.

er0ck:
assert_not_equal is defined in ruby's stdlib for unit tests... maybe it was being pulled in from there somehow.  
http://ruby-doc.org/stdlib-2.0.0/libdoc/test/unit/rdoc/Test/Unit/Assertions.html#method-i-assert_not_equal

joshcooper:
Or perhaps `oldcerts` was always empty due to the [regex](https://github.com/puppetlabs/puppet/pull/4516/files#diff-07074e1992a999b15c23a824456031ffR24) not containing quotes, so we never tried to call the method? Either way, :+1:

PR #4517, opened 2015-12-17
(PUP-4744) Fixes `reposdir` split by whitespace (Against stable)

petems:
* `reposdir` can be split by whitespace - http://yum.baseurl.org/gitweb?p=yum.git;a=blob;f=yum/config.py;h=97e5e3d3b063294a96227a67cce43a9655cd3aa6;hb=30c993c19b8b85f44e7cec436484862ab55a6961#l171
- Puppet previously only split by commas
- Now changed to a regex of [,|\s|\n] which will catch whitespace, commas and newlines (like in the original python code)
- Added new spec for this new regex

(New PR against stable as requested :+1:)

Continued from https://github.com/puppetlabs/puppet/pull/4495


petems:
gahhh, against master again...

PR #4525, opened 2015-12-20
(PUP-1293) Don't ignore status when using upstart provider

ffrank:
As requested by @baurmatt (I believe), here's a PR to backport the fix for PUP-1293 to `3.x`.


baurmatt:
Awesome, thanks! :)

PR #4529, opened 2015-12-28
(PUP-5628) Include file name in errors from puppet function loader

thallgren:
This commit ensures that the file name of a loaded puppet function is
propagated properly to the lexer so that errors are can be reported
including this information.


kylog:
Ah, I was just commenting on the unknown/nil thing when I saw you update the PR :)

LGTM. I'll let CI clear and then push the button.

PR #4532, opened 2015-12-29
(PUP-5589) Ensure that errors from terminus filter is propagated

thallgren:
Prior to this commit, the errors from the method
`Puppet::Indirector::Indirection.find` that occurred when calling
the `terminus.filter(result)` method was never logged. This caused
the puppet apply command to return 1 without any information since
the command assumes that the indirection has logged all
Puppet::Error exceptions.

This commit ensures that these particular errors are logged by
the `Indirector.find` method.

### DIFF
--- a/lib/puppet/indirector/indirection.rb
+++ b/lib/puppet/indirector/indirection.rb
@@ -207,10 +207,14 @@ class Puppet::Indirector::Indirection
         filtered = result
         if terminus.respond_to?(:filter)
           Puppet::Util::Profiler.profile("Filtered result for #{self.name} #{request.key}", [:indirector, :filter, self.name, request.key]) do
-            filtered = terminus.filter(result)
+            begin
+              filtered = terminus.filter(result)
+            rescue Puppet::Error => detail
+              Puppet.log_exception(detail)
+              raise detail
+            end
           end
         end
-
         filtered
       end
     end


joshcooper:
It'd be great to have a test for this

thallgren:
Here's why I didn't write a unit test.

Back in the days, almost all messages were printed twice. Then that improved with a fix so that the messages were printed only once. The fix introduced a regression so that some messages that were printed only once in the first place, now weren't printed at all. This fixes that regression.

Writing a test for this regression fix alone seems pointless. The proper test would be to write code that triggers every conceivable message and then asserts that it is printed exactly once when using the puppet apply command. That in turn, is a huge task.

joshcooper:
I hear what you're saying, and agree it's non-trivial to verify all possible exceptions are only logged once, but I wouldn't say adding a unit test is pointless. The contract for `Indirection#find` seems to be if an exception occurs within the method, then the method should log it before re-raising. For example, if we fail to save to the cache [we log and re-raise](https://github.com/thallgren/puppet/blob/a15b5b7909b1be36f8eacf1c93d9909b95b8b223/lib/puppet/indirector/indirection.rb#L199-L204).

thallgren:
I agree that a general test, with the objective to prevent similar regressions  in _puppet apply_, would be useful. It should test that some exceptions that may occur inside of `Indirection#find` are logged and re-raised. It would cover the re-throw you point to (not altered by this PR) and a similar clause below. We must also have tests that covers re-throws in `terminal#find` (in this case the terminus is the compiler, but there are many others) and the methods `request`, `prepare`, and `find_in_cache` since they all must honor the same contract. I think this should be covered in a separate JIRA.

PR #4534, opened 2015-12-29
(PUP-5640)(PUP-5591) Fix Win32 Registry Reads which cause errors

glennsarti:
- Added rescue clause around pointer read for out of bounds errors to trap
corrupted reg values
- Fixed REG_BINARY reads referring to a non existent local variable

### DIFF
--- a/lib/puppet/util/windows/registry.rb
+++ b/lib/puppet/util/windows/registry.rb
@@ -207,21 +207,27 @@ module Puppet::Util::Windows
         # buffer is raw bytes, *not* chars - less a NULL terminator
         string_length = (byte_length / FFI.type_size(:wchar)) - 1 if byte_length > 0
 
-        case type
-        when Win32::Registry::REG_SZ, Win32::Registry::REG_EXPAND_SZ
-          result = [ type, data_ptr.read_wide_string(string_length) ]
-        when Win32::Registry::REG_MULTI_SZ
-          result = [ type, data_ptr.read_wide_string(string_length).split(/\0/) ]
-        when Win32::Registry::REG_BINARY
-          result = [ type, data.read_bytes(0, byte_length) ]
-        when Win32::Registry::REG_DWORD
-          result = [ type, data_ptr.read_dword ]
-        when Win32::Registry::REG_DWORD_BIG_ENDIAN
-          result = [ type, data_ptr.order(:big).read_dword ]
-        when Win32::Registry::REG_QWORD
-          result = [ type, data_ptr.read_qword ]
-        else
-          raise TypeError, "Type #{type} is not supported."
+        begin
+          case type
+            when Win32::Registry::REG_SZ, Win32::Registry::REG_EXPAND_SZ
+              result = [ type, data_ptr.read_wide_string(string_length) ]
+            when Win32::Registry::REG_MULTI_SZ
+              result = [ type, data_ptr.read_wide_string(string_length).split(/\0/) ]
+            when Win32::Registry::REG_BINARY
+              result = [ type, data_ptr.read_bytes(byte_length) ]
+            when Win32::Registry::REG_DWORD
+              result = [ type, data_ptr.read_dword ]
+            when Win32::Registry::REG_DWORD_BIG_ENDIAN
+              result = [ type, data_ptr.order(:big).read_dword ]
+            when Win32::Registry::REG_QWORD
+              result = [ type, data_ptr.read_qword ]
+            else
+              raise TypeError, "Type #{type} is not supported."
+          end
+        rescue IndexError => ex
+          raise if (ex.message !~ /^Memory access .* is out of bounds$/i)
+          parent_key_name = key.parent ? "#{key.parent.keyname}\\" : ""
+          Puppet.warning "A value in the registry key #{parent_key_name}#{key.keyname} is corrupt or invalid"
         end
       end
 
--- a/spec/integration/util/windows/registry_spec.rb
+++ b/spec/integration/util/windows/registry_spec.rb
@@ -163,5 +163,96 @@ describe Puppet::Util::Windows::Registry, :if => Puppet::Util::Platform.windows?
         end
       end
     end
+
+    context "when reading binary values" do
+      let (:hklm) { Win32::Registry::HKEY_LOCAL_MACHINE }
+      let (:puppet_key) { "SOFTWARE\\Puppet Labs"}
+      let (:subkey_name) { "PuppetRegistryTest" }
+      let (:guid) { SecureRandom.uuid }
+
+      before :each do
+        hklm.create("#{puppet_key}\\#{subkey_name}", Win32::Registry::KEY_ALL_ACCESS) do |reg|
+          reg.write("#{guid}", Win32::Registry::REG_BINARY, "abinarystring")
+        end
+      end
+
+      after(:each) do
+        hklm.open(puppet_key, Win32::Registry::KEY_ALL_ACCESS) do |reg|
+          subject.delete_key(reg, subkey_name)
+        end
+      end
+
+      it "should return binary values" do
+
+        hklm.open("#{puppet_key}\\#{subkey_name}", Win32::Registry::KEY_READ) do |reg|
+          vals = subject.values(reg)
+
+          expect(vals).to have_key(guid)
+          written = vals[guid]
+          expect(written).to eq("abinarystring")
+        end
+
+      end
+    end
+
+    context "when reading corrupt values" do
+      let (:hklm) { Win32::Registry::HKEY_LOCAL_MACHINE }
+      let (:puppet_key) { "SOFTWARE\\Puppet Labs"}
+      let (:subkey_name) { "PuppetRegistryTest" }
+      let (:guid) { SecureRandom.uuid }
+
+      before :each do
+
+        module BadRegistryValues
+          extend FFI::Library
+
+          # https://msdn.microsoft.com/en-us/library/windows/desktop/ms724923(v=vs.85).aspx
+          # LONG WINAPI RegSetValueEx(
+          #   _In_             HKEY    hKey,
+          #   _In_opt_         LPCTSTR lpValueName,
+          #   _Reserved_       DWORD   Reserved,
+          #   _In_             DWORD   dwType,
+          #   _In_       const BYTE    *lpData,
+          #   _In_             DWORD   cbData
+          # );
+          ffi_lib :advapi32
+          attach_function :RegSetValueExW,
+            [:handle, :pointer, :dword, :dword, :pointer, :dword], :win32_long
+
+          def BadRegistryValues.write_bad_dword(reg,valuename)
+            # Normally DWORDs contain 4 bytes.  This bad data only has 2
+            bad_data = "\00\00".unpack('CC')
+            FFI::Pointer.from_string_to_wide_string(valuename) do |name_ptr|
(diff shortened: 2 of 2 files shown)


spec/unit/util/windows/registry_spec.rb
```diff
@@ -163,5 +163,77 @@ def expects_registry_value(array)
...
+
+    # TODO Not sure about this.  It seems a bit hacky....
+    context "when reading corrupt values" do
+      let (:hklm) { Win32::Registry::HKEY_LOCAL_MACHINE }
+      let (:puppet_key) { "SOFTWARE\\Puppet Labs"}
+      let (:subkey_name) { "PuppetRegistryTest" }
+      let (:guid) { SecureRandom.uuid }
+
+      before :each do
+        # Create a registry import file with a corrupt DWORD
```
glennsarti: I tried using REG.EXE and Ruby API calls but they all guarded against invalid values.  Eventually found I could craft a .REG file to make a bad value.  Need to test on 2008 though

ferventcoder:
Does this supersede my initial PR?

glennsarti:
That is the intent yes.  I wanted to get it working before I commented on your PR

lib/puppet/util/windows/registry.rb
```diff
@@ -207,21 +207,30 @@ def read(key, name_ptr, *rtype)
...
+            when Win32::Registry::REG_DWORD
+              result = [ type, data_ptr.read_dword ]
+            when Win32::Registry::REG_DWORD_BIG_ENDIAN
+              result = [ type, data_ptr.order(:big).read_dword ]
+            when Win32::Registry::REG_QWORD
+              result = [ type, data_ptr.read_qword ]
+            else
+              raise TypeError, "Type #{type} is not supported."
+          end
+        rescue IndexError => ex
```
ferventcoder: Is it always an IndexError?

lib/puppet/util/windows/registry.rb
```diff
@@ -207,21 +207,30 @@ def read(key, name_ptr, *rtype)
...
+          if (ex.message =~ /^Memory access .* is out of bounds$/i)
+            parent_key_name = key.parent ? "#{key.parent.keyname}\\" : ""
+            Puppet.warning "A value in the registry key #{parent_key_name}#{key.keyname} is corrupt or invalid"
+          else
+            raise ex
```
ferventcoder: Is this handled elsewhere or do we still want this to crash?

spec/unit/util/windows/registry_spec.rb
```diff
@@ -163,5 +163,77 @@ def expects_registry_value(array)
...
+        tempfilename = "#{ENV['temp']}\\BadDword-PuppetRegistryTest.Reg"
+        tempfile = open(tempfilename,'w')
+        tempfile.write("Windows Registry Editor Version 5.00\n")
+        tempfile.write("\n")
+        tempfile.write("[HKEY_LOCAL_MACHINE\\#{puppet_key}\\#{subkey_name}]\n")
+        tempfile.write("\"#{guid}\"=hex(4):00\n")
+        tempfile.close
+
+        # Execute regedit and import the corrupt DWORD
+        retval = `regedit /s "#{tempfilename}"`
```
ferventcoder: interesting way about it.

spec/unit/util/windows/registry_spec.rb
```diff
@@ -163,5 +163,77 @@ def expects_registry_value(array)
...
+
+        # Cleanup
+        retval = File.delete(tempfilename)
```
ferventcoder: What do you use `retval` for?

lib/puppet/util/windows/registry.rb
```diff
@@ -207,21 +207,30 @@ def read(key, name_ptr, *rtype)
...
-        else
-          raise TypeError, "Type #{type} is not supported."
+        begin
+          case type
+            when Win32::Registry::REG_SZ, Win32::Registry::REG_EXPAND_SZ
+              result = [ type, data_ptr.read_wide_string(string_length) ]
+            when Win32::Registry::REG_MULTI_SZ
+              result = [ type, data_ptr.read_wide_string(string_length).split(/\0/) ]
+            when Win32::Registry::REG_BINARY
+              result = [ type, data_ptr.read_bytes(byte_length) ]
```
ferventcoder: This is where the change for binary is.

spec/unit/util/windows/registry_spec.rb
glennsarti: Suppressing any STDOUT or other output

lib/puppet/util/windows/registry.rb
glennsarti: Yes.  That was from your original PR

lib/puppet/util/windows/registry.rb
glennsarti: Yes.  The underlying call is to FFI MemoryPointer get_bytes

http://www.rubydoc.info/github/ffi/ffi/FFI%2FAbstractMemory%3Aget_bytes
"(IndexError) — if length is too great"

lib/puppet/util/windows/registry.rb
glennsarti: My thinking was to only handle specific error conditions in this low level function.  Errors should be handled by the publicly accessible methods like it is .open and .root

https://github.com/glennsarti/puppet/blob/ticket/stable/PUP-5591-5640-registry/lib/puppet/util/windows/registry.rb#L30-L32

However that is currently not the case.  Do you think I should be far more defensive about error handling in the other public methods?  Perhaps at least catch the errors and rethrow as a Puppet::Util::Windows::Error class?

spec/unit/util/windows/registry_spec.rb
glennsarti: Not required though.  Removing.

lib/puppet/util/windows/registry.rb
ferventcoder: yup, just adding a note here :)

spec/unit/util/windows/registry_spec.rb
ferventcoder: I see.

lib/puppet/util/windows/registry.rb
ferventcoder: This is probably okay. Do we have anything to cleanup that we would want to do in an ensure?

glennsarti: I believe the MemoryPointers will automatically be cleaned up when it's freed, so I can't see anything else that needs to be cleaned up.

See autorelease in;
http://www.rubydoc.info/github/ffi/ffi/FFI/Pointer

glennsarti:
This PR supersedes https://github.com/puppetlabs/puppet/pull/4488

spec/integration/util/windows/registry_spec.rb
```diff
@@ -163,5 +163,75 @@ def expects_registry_value(array)
...
+
+    context "when reading corrupt values" do
+      let (:hklm) { Win32::Registry::HKEY_LOCAL_MACHINE }
+      let (:puppet_key) { "SOFTWARE\\Puppet Labs"}
+      let (:subkey_name) { "PuppetRegistryTest" }
+      let (:guid) { SecureRandom.uuid }
+
+      before :each do
+        # Create a registry import file with a corrupt DWORD ( < 4 bytes in length)
+        tempfilename = "#{ENV['temp']}\\BadDword-PuppetRegistryTest.Reg"
```
Iristyle: Puppet already has temp file helpers... they track the file, and later clean up afterwards.

https://github.com/puppetlabs/puppet/blob/master/spec/lib/puppet_spec/files.rb#L27-L37

I wonder if we can do this without files and calling out to `regedit.exe` though

spec/integration/util/windows/registry_spec.rb
```diff
@@ -163,5 +163,75 @@ def expects_registry_value(array)
...
+
+      after(:each) do
+        hklm.open(puppet_key, Win32::Registry::KEY_ALL_ACCESS) do |reg|
+          subject.delete_key(reg, subkey_name)
+        end
+      end
+
+      it "should return binary values" do
+
+        hklm.open("#{puppet_key}\\#{subkey_name}", Win32::Registry::KEY_ALL_ACCESS) do |reg|
```
Iristyle: Do you need `KEY_ALL_ACCESS` here just for reads?

lib/puppet/util/windows/registry.rb
Iristyle: FFI impl is https://github.com/ffi/ffi/blob/master/ext/ffi_c/AbstractMemory.c#L450-L472

Looks like `IndexError` is indeed the only thing that we can deal with...

lib/puppet/util/windows/registry.rb
Iristyle: As a minor matter of style, might want to rewrite to raise early like this:

```ruby
raise if (ex.message !~ /^Memory access .* is out of bounds$/i)
```

Note that Ruby supports re-raising to preserve original exception / stacktrace / etc, by supplying no arg to `raise` (your way is also correct).

For reference, I originally put this sample together to verify Ruby usage, as we were doing it wrong in a number of spots inside of Puppet code:

```ruby
def foo
  raise ArgumentError.new("bad arg yo")
end

def bar
  begin
    foo
  # good
  # rescue ArgumentError => e
    # raise
    # raise e, e.message, e.backtrace
    # raise e
  rescue
    # to preserve backtrace good
    # raise
    # raise $!
    # raise $!, "bad arg yo -- part 2", $!.backtrace
    # raise $!, $!.message, $!.backtrace

    #bad
    # raise $!, "Just smoked the backtrace on existing exception"
    # raise "Just smoked the backtrace and changed the exception type "
    # raise RuntimeError.new($!)
    # raise $!, $!.backtrace
  end
end

def doit
  begin
    bar
  rescue
    puts $!.class
    puts $!
    puts $!.backtrace
  end
end

doit
```

ferventcoder:
Some of the original commit message from #4488 should probably make it into the commit message:


Previously, the read method was using an older variable name when
reading data for binary values. It would result in errors such as the
following when reading a binary value from the registry.

~~~
Error: Could not run: undefined local variable or method `data' for
Puppet::Provider::Package::Windows::Package:Class
C:/Program Files/Puppet
Labs/Puppet/puppet/lib/puppet/util/windows/registry.rb:216:in `block in
read'
~~~

This was introduced in b46ede7 (Puppet
v4.0.0).

Without this fix, any registry enumerations that contain a binary
value will fail.

spec/integration/util/windows/registry_spec.rb
glennsarti: I will look at using the helper class instead of the current method.

I tried creating invalid values using API calls, WMI and REG.EXE and they all protected me from making a bad call.  REGEDIT.EXE was the only method that I could find that would allow me to inject a bad DWORD.  I do not believe regedit will accept STDIN so the only method left was a crafted temporary reg file to import.

spec/integration/util/windows/registry_spec.rb
glennsarti: Probably not. As it was just the spec tests I did not see the need to find the precise permissions required.  I will investigate other options.

lib/puppet/util/windows/registry.rb
glennsarti: Corrected.

glennsarti:
@Iristyle Bad data injection has been changed from REG.EXE to using raw Win32 API

PR #4535, opened 2015-12-29
(PUP-5612) Detect and report resource definitions with missing title

thallgren:
A resource declared without title becomes a BlockExpression with
two statments, a QualifiedName and a LiteralHash. This commit adds
validation logic that detects and reports such statements as a new
`RESOURCE_WITHOUT_TITLE` issue. It also ensures that when such a
declaration is found last in a node, define, or host class, it's
reported instead of the `IDEM_NOT_ALLOWED_LAST` since reporting both
would be confusing.

### DIFF
--- a/lib/puppet/pops/validation/checker4_0.rb
+++ b/lib/puppet/pops/validation/checker4_0.rb
@@ -213,11 +213,24 @@ class Puppet::Pops::Validation::Checker4_0
     rvalue(o.right_expr)
   end
 
+  def resource_without_title?(o)
+    if o.instance_of?(Puppet::Pops::Model::BlockExpression)
+      statements = o.statements
+      statements.length == 2 && statements[0].instance_of?(Puppet::Pops::Model::QualifiedName) && statements[1].instance_of?(Puppet::Pops::Model::LiteralHash)
+    else
+      false
+    end
+  end
+
   def check_BlockExpression(o)
-    o.statements[0..-2].each do |statement|
-      if idem(statement)
-        acceptor.accept(Issues::IDEM_EXPRESSION_NOT_LAST, statement)
-        break # only flag the first
+    if resource_without_title?(o)
+      acceptor.accept(Issues::RESOURCE_WITHOUT_TITLE, o, :name => o.statements[0].value)
+    else
+      o.statements[0..-2].each do |statement|
+        if idem(statement)
+          acceptor.accept(Issues::IDEM_EXPRESSION_NOT_LAST, statement)
+          break # only flag the first
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/puppet/pops/validation/checker4_0.rb
```diff
@@ -213,11 +213,24 @@ def check_BinaryExpression(o)
...
-        acceptor.accept(Issues::IDEM_EXPRESSION_NOT_LAST, statement)
-        break # only flag the first
+    if resource_without_title?(o)
+      acceptor.accept(Issues::RESOURCE_WITHOUT_TITLE, o, :name => o.statements[0].value)
+    else
+      o.statements[0..-2].each do |statement|
+        if idem(statement)
+          acceptor.accept(Issues::IDEM_EXPRESSION_NOT_LAST, statement)
+          break # only flag the first
+        end
```
hlindberg: the above changes looks strange - it assumes there is always a block with exactly a name and a hash - clearly the error can occur first, last, in the middle of a block - I am not sure that this impl covers all cases.

OTOH, it may be that the rewrite rules for function calls and resouces always produce exactly such a block as the one being found by `resource_without_title` - is that really the case?

spec/unit/pops/validator/validator_spec.rb
```diff
@@ -107,6 +107,15 @@ def validate(model)
...
       end
+
+      it "detects a resource declared without title in #{type}" do
+        source = <<-SOURCE
+          #{type} nope {
+            notify { message => 'Nope' }
+          }
+        SOURCE
+        expect(validate(parse(source))).to have_issue(Puppet::Pops::Issues::RESOURCE_WITHOUT_TITLE)
+      end
```
hlindberg: I want to see additional test cases with epressions before, before and after, and after the faulty expression.

lib/puppet/pops/validation/checker4_0.rb
thallgren: Not sure I understand. The RESOURCE_WITHOUT_TITLE can occur anywhere. This addition checks THE block that is the result of such a construct. That block is very likely to exist within another block but that's not tested here at all. I.e. the addition is not concerned with how such a block is nested within another block and it doesn't have to be.

thallgren:
Added the desired test cases.

lib/puppet/pops/validation/checker4_0.rb
hlindberg: I can answer my own question by looking at the additional tests; apparently there is always a block produced of size 2 (just what the predicate method checks) instead of a some kind of broken resource expression, or keeping the two expressions in questions in the same list of statements as what surrounds it. This fix is good enough for now as it is covered by tests - at some point the grammar may be refactored in this area and the tests will then catch that.

PR #4537, opened 2015-12-30
(PUP-5209) Ensure correct visibility between modules

thallgren:
This commit deals with function visibility between modules. It fixes
two major problems and adds a test matrix containing 27 distinct
test cases.

First issue was that a puppet function didn't use the private loader.
This prevented puppet functions in general from calling functions in
other modules.

Second issue was that calls made from outside of a function (i.e.
from .pp code not associated with a 'pops' loader) had access to
all functions in all modules. This happened because the caller
defaulted to using the private environment loader. This commit adds
logic that will attempt to find out what module a call originates from
and, if successful, use the private loader of that module to satisfy
the function lookup. A better fix for this would be to use the new
loader framework throughout but that's a much larger task (see
PUP-1833).

The test matrix tests calls from three different call locations:
- directly from manifests/init.pp
- from a puppet function declared under the modules functions directory
- from a puppet function declared in the modules manifests/init.pp
- from a ruby function

Each call location performs a call to the dependent module that targets
- a puppet function declared under the functions diretory
- a puppet function declared in manifests/init.pp
- a ruby function

Each call is tested using the following conditions:
- Metadata present with a valid dependency to the callee
- Metadata present that lacks a dependency to the callee (should fail)
- No metadata present (will succeed since all modules are visible)


hlindberg:
Ensure PUP 1833 ticket is linked to this ticket PR to find code that needs to be addressed  when dealing with PUP 1833

lib/puppet/pops/loader/puppet_function_instantiator.rb
```diff
@@ -43,7 +43,8 @@ def self.create(loader, typed_name, source_ref, pp_code_string)
     # Adapt the function definition with loader - this is used from logic contained in it body to find the
     # loader to use when making calls to the new function API. Such logic have a hard time finding the closure (where
     # the loader is known - hence this mechanism
-    Puppet::Pops::Adapters::LoaderAdapter.adapt(the_function_definition).loader = loader
+    private_loader = loader.private_loader
+    Puppet::Pops::Adapters::LoaderAdapter.adapt(the_function_definition).loader = private_loader
```
hlindberg: This is not correct. We are currently not using the private loader as that is for functions that are declared to be private - and that means that they are private to the module. Until we have that, all functions in a module are public. All modules that are allowed to load public things from the module should be able to access things loaded from its public loader, but not from its private. The problem is elsewhere - in that a search should start from the private counterpart for any loaded item.

thallgren: If that's the case, why is the private loader used in the `RubyFunctionInstantiator`? And why does this work while using the original loader doesn't?

hlindberg: that sounds like a problem as well... need to read up on how the dependent loaders are organized; module A depending on module B, should see module B's public loader, but not its private. That is how it (should be) designed to work.

thallgren: Ah, but this is about module A using module A's private loader, which in turn will see module B's public loader.

hlindberg:
I think I figured it out. Functions that are loaded are actually all loaded by the public module loader (it is the one instantiated with path support) - this means that the entry is made in the public loader. Then, the loaded entity should be associated with that loader's private loader as it searches first up the parent chain, then in itself, and then in its dependencies.

When we introduce the notion of a private function, the loader should store it in the private loader. At all times should the associated loader be the private one.

With that said, the change is ok. I do think there is an additional place where potentially the wrong loader is used - checkout pops bridge which is used when a manifest no loaded by the loaders is found.

lib/puppet/pops/loader/puppet_function_instantiator.rb
hlindberg: See general comment. This is ok. I was confused by other code (that is wrong in this respect).

thallgren:
I checked the `instantiate_FunctionDefinition` method in the pops bridge. It passes the loader to the `PuppetFunctionInstantiator` which in turn uses the private loader. That part looks correct to me. If it's not, how can I verify that with a test in the test matrix provided with this PR? I'd like that matrix to cover all cases.

hlindberg:
you can place a function definition in a module's init.pp manifest, and include the module's class (to force the init.pp to be loaded). Then, in that class call the function defined in the init.pp, and have it call functions in the same module, and visible and invisible modules.

hlindberg:
which test checks that "a function defined in modulea/manifests/init.pp2 can be called from a module that depends on modulea after having included the class modulea (to load the init.pp)?

thallgren:
That would be [Case 8](https://github.com/thallgren/puppet/blob/5784d9432408e0ae947d2589f7f212f965d40693/spec/fixtures/unit/pops/loaders/loaders/dependent_modules_with_metadata/modules/user/manifests/init.pp#L52-L57 in this init.pp)

PR #4538, opened 2015-12-30
(PUP-5583) Bump to beaker-hostgenerator 0.3.1

kylog:
The puppet-agent components need the osinfo version 1 of b-hg,
so that centos4 is correctly specified as a platform name, so
that puppet-agent can be installed on centos4.

This commit moves to b-hg 0.3.1, the first version implementing
that, and specifies the osinfo version when invoking b-hg.

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -13,7 +13,7 @@ def location_for(place, fake_version = nil)
 end
 
 gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 2.28')
-gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 0.3.1")
+gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 0.3")
 gem "rake", "~> 10.1"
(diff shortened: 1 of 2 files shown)


acceptance/Gemfile
```diff
@@ -13,7 +13,7 @@ def location_for(place, fake_version = nil)
 end
 
 gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 2.28')
-gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 0.2")
+gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 0.3.1")
```
waynr: Wouldn't this prevent you from getting new versions of the gem > `0.3`? ie, `0.4.0`, `0.5.1`...is that intentional?

kylog: Yeah I'll drop that back to 0.3 so we don't get left behind.

kylog:
@waynr if this looks good to you, merge away. Same for https://github.com/puppetlabs/hiera/pull/335 and https://github.com/puppetlabs/facter/pull/1247.

waynr:
:+1:

PR #4539, opened 2016-01-01
(PA-139) unify pa acceptance workflows

er0ck:
(PA-139) acceptance: allow multiple options to beaker using OPTIONS

(PA-139) acceptance: move beaker host generation to own task

    This change moves host file generation to its own task. This separates
    its concerns from the testing tasks, and allows the testing tasks to be
    dependent upon it.

 (PA-139) add tmp hostgenerator dir to rake clean list

(PA-139) move common options out of Rakefile

    Many people run with beaker standalone. Common and required options have
    been moved to the options files to support this workflow in this change.
    We also move common options to their own file and load them into the
    other options files that are loaded by beaker.

(PA-139) acceptance: fix running against preserved_hosts

(PA-139) fix releasing of preserved hosts

    Fix the releasing of hosts helper task by using the user's vmpooler
    token from their .fog file. We also remove some lingering
    preserved_config logic, which is no longer used due to improvements in
    beaker.

(PA-139) fix listing of preserved hosts

    This change fixes developer helper tasks made for managing acceptance
    VMs easier. We also remove methods that are superseded by beaker
    functionality.

(PA-139) acceptance: Rakefile: deprecate CONFIG in favor of HOSTS

(PA-139) acceptance: remove passenger and packages installs

[skip ci]


er0ck:
i only tested the test_against_preserved_hosts task with an empty simple test.  further testing reveals that beaker wipes out the HOSTS hash when using any options file that points at a preserved_hosts file.  i believe a work around is available that i'm working out for deployer.  digging into the morass of coupling that is the beaker options parser is probably not the best route for now...

johnduarte:
suggestion to use `BEAKER_HOSTCONFIG` rather than `HOSTS`  thoughts?

er0ck:
yeah, i'm sold on BEAKER_HOSTCONFIG use.  i have a couple other updates for this PR.  but maybe we should just get those minimal changes in, similar to your hiera PR and then i'll rebase this one as it has extensive unrelated changes.

johnduarte:
floating an update to `hiera` component to use `BEAKER_HOSTCONFIG`. This can be propagated to other `puppet-agent` components when approved.
https://github.com/puppetlabs/hiera/pull/338

acceptance/Rakefile
```diff
@@ -18,22 +19,9 @@ EPEL_MIRROR = ENV['BEAKER_EPEL_MIRROR']
...
-      :keys => ["id_rsa_acceptance", "#{ENV['HOME']}/.ssh/id_rsa-acceptance"],
-    },
-    :xml => true,
-    :timesync => false,
-    :repo_proxy => true,
-    :add_el_extras => true,
     :preserve_hosts => 'onfail',
-    :forge_host => 'forge-aio01-petest.puppetlabs.com',
-    :'master-start-curl-retries' => 30
   }
```
er0ck: review note: moved most of these to the config files so one can run more easily from beaker standalone without repeated code.

acceptance/Rakefile
```diff
@@ -324,73 +258,53 @@ EOS
...
 
-  desc "List acceptance runs from the past day which had hosts preserved."
-  task :list_preserved do
-    preserved = list_preserved_configurations
-    print_preserved(preserved)
+  desc "List preserved hosts from latest (default) or runs within given hours; e.g.: rake ci:list_hosts[24]"
+  task :list_hosts, :hours_ago do |t, args|
+    latest = 0
+    secs_ago = args[:hours_ago].to_f*60*60 || latest
+    print_hosts(preserved_hosts(secs_ago))
```
er0ck: review note: task with optional number of hours argument

MikaelSmith:
Is this starting from a baseline that's compatible with the changes in FACT-1281?

er0ck:
@MikaelSmith not yet.  but it can be.  we should probably wait until after the release for stuff like this PR

er0ck:
wrong branch name.  re-opening

PR #4541, opened 2016-01-01
Updated copyright to 2016

aqnouch:


whopper:
@aqnouch thanks for the contribution, but we'll actually be updating the copyright internally across all of our repos :)

PR #4542, opened 2016-01-04
(PUP-5644) Avoid that lookup command configures node cache and ssl

thallgren:
The setup of the puppet lookup application performed a full master
config/setup process. This is not necessary and this commit removes
setup of the write-only node-cache and ssl configuration
that creates CA-cert.

### DIFF
--- a/lib/puppet/application/lookup.rb
+++ b/lib/puppet/application/lookup.rb
@@ -94,31 +94,6 @@ class Puppet::Application::Lookup < Puppet::Application
     Puppet::FileBucket::File.indirection.terminus_class = :file
   end
 
-  def setup_ssl
-    # Configure all of the SSL stuff.
-    if Puppet::SSL::CertificateAuthority.ca?
-      Puppet::SSL::Host.ca_location = :local
-      Puppet.settings.use :ca
-      Puppet::SSL::CertificateAuthority.instance
-    else
-      Puppet::SSL::Host.ca_location = :none
-    end
-    # These lines are not on stable (seems like a copy was made from master)
-    #
-    # Puppet::SSL::Oids.register_puppet_oids
-    # Puppet::SSL::Oids.load_custom_oid_file(Puppet[:trusted_oid_mapping_file])
-  end
-
-  # Sets up a special node cache "write only yaml" that collects and stores node data in yaml
-  # but never finds or reads anything (this since a real cache causes stale data to be served
-  # in circumstances when the cache can not be cleared).
-  # @see puppet issue 16753
-  # @see Puppet::Node::WriteOnlyYaml
-  # @return [void]
-  def setup_node_cache
-    Puppet::Node.indirection.cache_class = Puppet[:node_cache_terminus]
-  end
-
   def setup
     setup_logs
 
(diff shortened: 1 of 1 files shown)


hlindberg:
I wonder if this has the potential to break EYAML Lookup should work if the global hiera has an EYAML provider. I think it is ok since EYAML logic should have nothing to do with cert creation.

lib/puppet/application/lookup.rb
```diff
@@ -127,11 +102,6 @@ def setup
     Puppet.settings.use :main, :master, :ssl, :metrics
 
     setup_terminuses
-
-    # TODO: Do we need this in lookup? It sets up a write only cache
```
hlindberg: This comment is wrong - it sets up the configured node cache.

lib/puppet/application/lookup.rb
```diff
@@ -94,31 +94,6 @@ def setup_terminuses
...
-  end
-
-  # Sets up a special node cache "write only yaml" that collects and stores node data in yaml
-  # but never finds or reads anything (this since a real cache causes stale data to be served
-  # in circumstances when the cache can not be cleared).
-  # @see puppet issue 16753
-  # @see Puppet::Node::WriteOnlyYaml
-  # @return [void]
-  def setup_node_cache
-    Puppet::Node.indirection.cache_class = Puppet[:node_cache_terminus]
```
hlindberg: Note that this sets up the configured node cache terminus (whatever it may be). Comment is referring to the default puppet configuration, but if the config has another node cache then, not setting it may cause problems. This logic was added is part of PUP 5059 to support `--node`.

I am unsure if there is a test case that covers the potential problem this "fixed"

thallgren: I'm pretty sure there isn't since removing it has absolutely no effect on neither unit, integration, nor acceptance test.

hlindberg:
I think we need to assert that we have full coverage of the use-cases that are supposed to work to ensure that it is ok to remove things like the node cache configuration. We should not need to load logic that is for signing certs, We do need to be able to handle a looked up node with trusted data, not sure if there is the need in any of that logic to have SSL things set up.

lib/puppet/application/lookup.rb
thallgren: It's consistent with the documentation of the method where it says _Sets up a special node cache "write only yaml"_.

thallgren:
I'd like to see a new JIRA where all use-cases that needs testing are specified with desired outcome. Given recent comments there seems to be a lot of them. Root, non-root, ENC or not, PuppetDB or not, Is EYAML affected?, master/agent, masterless, etc.

lib/puppet/application/lookup.rb
hlindberg: Yeah, but that is also wrong

hlindberg:
Given that we are unsure about tests and if the things this PR removes has any effect (or are just slowing things down), do we really want to merge this without first having those tests?

thallgren:
If it was up to me, we'd merge because this PR does fix annoyances that were reported and it doesn't affect any of the tests we currently have. If merged in combination with PUP-5461, then I think the risk that what we remove have any effect at all is a lot smaller since we no longer evaluate everything. A proper test suite will be very valuable anyway.

PR #4543, opened 2016-01-04
(PUP-5646) - pad seed fqdn_rand seed with max number

traylenator:
Currently with a default seed of the ::fqdn there is
seemingly a correlation between fqdn_rand(60) and
fqdn_rand(24) reguardless of the fqdn input.

The extra padding ensures the seed is different for the
two intentially independed fqdn calls.

### DIFF
--- a/lib/puppet/parser/functions/fqdn_rand.rb
+++ b/lib/puppet/parser/functions/fqdn_rand.rb
@@ -16,6 +16,6 @@ Puppet::Parser::Functions::newfunction(:fqdn_rand, :arity => -2, :type => :rvalu
   node. (For example, `fqdn_rand(30)`, `fqdn_rand(30, 'expensive job 1')`, and
   `fqdn_rand(30, 'expensive job 2')` will produce totally different numbers.)") do |args|
     max = args.shift.to_i
-    seed = Digest::MD5.hexdigest([self['::fqdn'],args].join(':')).hex
+    seed = Digest::MD5.hexdigest([self['::fqdn'],max,args].join(':')).hex
     Puppet::Util.deterministic_rand_int(seed,max)
 end


traylenator:
https://tickets.puppetlabs.com/browse/PUP-5646

kylog:
@traylenator thanks for the contribution! I'll have to stare at this a bit to grok what was happening here.

Meanwhile, a small piece of bookkeeping: can you amend the commit message to reference `(PUP-5646)` (rather than MCO-xxxx)? That will make our commit message checker happy :)

Thanks!

hlindberg:
As I understand this, since the salt will be different when max is also included in the input to generate the salt, this will mean that a different series will be produced and this will be reported as a change. Is that ok to introduce at a minor version?

hlindberg:
I think I understand what is happening - since fqdns are too similar, they get the same or seeds that are too close to each other. By including max, the seeds will be different. The same effect is achieved by simply giving additional arguments - i.e. calling `fqdn_rand(24, 24)`, or for that matter, any number of additional arguments to make the seeds longer and different. There is really no need o modify the function itself to get this effect. Those that want a wider range, simply add more arguments.

traylenator:
Thanks for looking through this.

I think it is worth changing. Patching every module on the forge that uses fqdn_rand to add a second parameter is quite time consuming.

Re the seed it is currently not just close but completely equal for hour and minute.

It will indeed introduce a change to the catalog but only from one random number to a better random number. Is that really a change?

traylenator:
Just to add with a time of fqdn_rand(24):fqdn_rand(60) although the jira ticket subject is "only 336 of 1440 slots are used" it's even worse since the vast majority of times used are in the "leading diagonal" so only 60 slots in a day.

hlindberg:
I also think it is worth changing. The question is at which version boundary. The problem with random numbers changing to better random numbers is naturally a good thing, and not an issue. However since the random series also will cause a catalog that is different, it will cause people to have to review unanticipated changes and for them to understand why those changes occurred. It may be difficult to understand that it is because of a better, more wide spread random number series and that it should be anticipated. Users are sensitive to the smallest things in .z releases. Now this is targeting master which would be Puppet 4.4.0 next. My question is if this change is tolerable there, or if it has to wait until 5.0.

kylog:
Btw, poking at the code a teensy bit, I think there's some dead code. See #4591.

traylenator:
5.0.0 is fine from my point of view, I have it deployed already.

hlindberg:
Thanks @traylenator - we decided this was ok to go into the next minor release, and this is now merged. Will appear in Puppet 4.3.3

joshcooper:
nit, this will appear in Puppet 4.4.0, since this is merged to master.

PR #4544, opened 2016-01-04
(maint) acceptance-fix with() function usage

er0ck:
In the calling_all_functions test we were passing a block to with() with
the wrong number of arguments. This passed in ruby 1.9 due to...
reasons. ruby 1.8 caught this when we ported the test to 3.x.
This change fixes the with() usage.

re: PUP-5378
[skip ci]

### DIFF
--- a/acceptance/tests/parser_functions/calling_all_functions.rb
+++ b/acceptance/tests/parser_functions/calling_all_functions.rb
@@ -100,7 +100,7 @@ agents.each do |agent|
     {:name => :reduce,           :args => '[4,5,6]',                           :lambda => '|$sum, $n| { $sum+$n }', :expected => '15', :rvalue => true},
     #         :reuse,:recycle
     {:name => :slice,            :args => '[1,2,3,4,5,6], 2',                  :lambda => nil, :expected => '[[1, 2], [3, 4], [5, 6]]', :rvalue => true},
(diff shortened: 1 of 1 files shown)


joshcooper:
Looks good. I did notice an unrelated issue:

```
ix57icb9ydnhadj.delivery.puppetlabs.net (agent) 10:55:27$ puppet apply --trusted_node_data  --color=false --modulepath /tmp/calling_all_functions.xbdkl8/environments/production/modules/ /tmp/apply_manifest.pp.v1AvjT
  Warning: Unrecognised escape sequence '\w' in file /tmp/apply_manifest.pp.v1AvjT at line 32
  Warning: Unrecognised escape sequence '\1' in file /tmp/apply_manifest.pp.v1AvjT at line 32
```

Which I think is coming from:

```
    {:name => :regsubst,         :args => '"Cersei","Cer(\\w)ei","Daenery\\1"',:lambda => nil, :expected => 'Daenerys', :rvalue => true},
```

I think we don't want to escape the backslash, since it's using outer single quotes, e.g. `\w`

joshcooper:
Actually it looks like it's not enough backslashes, and should be:

```
    {:name => :regsubst,         :args => '"Cersei","Cer(\\\w)ei","Daenery\\\1"',:lambda => nil, :expected => 'Daenerys', :rvalue => true},
```

PR #4545, opened 2016-01-04
(maint) Relax fqdn_rand regex

joshcooper:
Previously, we required `fqdn_rand(100000)` to return a 5 digit number,
but it can return any number greater than or equal to 0 and less than
100000.  So relax the regex to require between 1 and 5 digits.

Also the regsubst function was defining a regex in a single quoted
string which is written to a manifest as:

    regsubst("Cersei","Cer(\w)ei","Daenery\1")

But `\w` in a double quoted string is an invalid escape sequence,
causing puppet to output a warning, but somehow successfully apply
the regex:

    Warning: Unrecognised escape sequence '\w'

This commit escapes the backslash so that the manifest contains:

    regsubst("Cersei","Cer(\\w)ei","Daenery\\1")

This eliminates the warning, while continuing to verify that the
regsubst function returns `Daenerys`.


joshcooper:
@er0ck Here's the PR for that issue. It should also resolve the issues in today's 3.x acceptance run https://jenkins.puppetlabs.com/view/legacy_Puppet/view/3.x/job/legacy_puppet_packaging-callback_3.x/9/

er0ck:
(facepalm) 1..5 of course.

PR #4546, opened 2016-01-05
(PUP-5461) Prevent that lookup evaluates site manifest

thallgren:
Prior to this commit, the `lookup` application would compile a full
catalog which gave undesired side effects. This commit ensures that
this compilation does not take place unless the flag `--compile` is
given on the command line. Without this option, the lookup command will
instead compile the string `undef` to produce the scope needed
for the lookup.

### DIFF
--- a/spec/unit/application/lookup_spec.rb
+++ b/spec/unit/application/lookup_spec.rb
@@ -78,7 +78,7 @@ describe Puppet::Application::Lookup do
   end
 
 
-  context 'when asked to explain' do
+  context 'when given a valid configuration' do
     let (:lookup) { Puppet::Application[:lookup] }
 
     # There is a fully configured 'sample' environment in fixtures at this location
@@ -86,7 +86,7 @@ describe Puppet::Application::Lookup do
 
     let(:facts) { Puppet::Node::Facts.new("facts", {}) }
 
-    let(:node) { Puppet::Node.new("testnode", :facts => facts, :environment => environment) }
+    let(:node) { Puppet::Node.new("testnode", :facts => facts, :environment => 'production') }
 
     let(:expected_json_hash) { {
       'type' => 'merge',
@@ -162,8 +162,8 @@ describe Puppet::Application::Lookup do
       end
     end
 
-    it 'produces human readable text by default' do
-      lookup.options[:node] = Puppet::Node.new("testnode", :facts => facts, :environment => 'production')
+    it '--explain produces human readable text by default' do
+      lookup.options[:node] = node
       lookup.options[:explain] = true
       lookup.command_line.stubs(:args).returns(['a'])
       expect { lookup.run_command }.to output(<<-EXPLANATION).to_stdout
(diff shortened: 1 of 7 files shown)


spec/unit/application/lookup_spec.rb
```diff
@@ -78,7 +78,7 @@
   end
 
 
-  context 'when asked to explain' do
+  context 'when using a proper fixture' do
```
hlindberg: Uh - this is an odd name for a context, all labels should be meaningful as names of examples, now it is named after a technical implementation detail. The old name was better. Has it changed so it is not just about explain. If so perhaps say "when given a valid configuration"

thallgren:
And why on earth does Ruby 1.9.3 output an exclamation mark when it renders the output? But just for one single statement?

hlindberg:
Do we have tests for these cases? 
* "when compiling, does not cause exports to be stored" (we do not want a lookup to store exported data in the configured store e.g. PDB).
* "when compiling, gets variables from a configured ENC" (this to ensure that compilation is as close as possible to a real catalog compilation, when an ENC is used it can assign classes and variable values). I included this as a test, as ENC is part of the initial sequence of compilation and the lookup command is in charge of the initial setup/config.
* "when compiling, and a call is made to hira_include those included classes are instantiated and their variables are available" (I see no reason why this would cause problems, and requires a bit of work to add, so probably skip this test).

thallgren:
No, I don't think we have tests for any of those cases. Using PDB and/or a configured ENC sounds like acceptance testing rather than simple unit tests.

thallgren:
I do think we should consider a puppet lookup with --compile as something that might have side effects though. When that's requested, it does compile the full catalog for a reason and the compiler will interact with other services. Isn't that what we intend?

spec/unit/application/lookup_spec.rb
thallgren: Heh, chose the term 'fixture' because I found it more descriptive than 'configuration' since it implies _actual data from files on disk_.

hlindberg: As opposed to "files not on disk" ? Please write it from an example point of view. Is this testing files on disk as opposed to something else? Is it supposed to test " proper fixture" as opposed to "improper fixture"?

Te name is supposed to describe *what* is tested (features/functonaity), not *how* it is tested.

hlindberg:
It should interact with other services as a consumer, but not as a producer - exporting has effect on what other real compilations see as exported - it may produce complete nonsense esp. if testing with hand crafted facts.

spec/unit/application/lookup_spec.rb
thallgren: No, as opposed to just "a configuration" that might be provided as ruby hashes or mocks.

hlindberg:
Thinking more about using `class noop` - can we use some other "empty" statement to produce an empty catalog?

hlindberg:
One more test case:
* "when not-compiling, the ENC should not be called" (mainly for performance reasons)

thallgren:
Regarding `class noop`. Any compilable statement will do except for an empty string. Suggestions welcome.

hlindberg:
"1+1" :-)

thallgren:
New commit is rebased on stable, has better context description, and compiles "undef" instead of "class noop {}"

hlindberg:
much better, waiting for tests to go green before merging

PR #4547, opened 2016-01-05
(PUP-5593) update Cumulus node config

LuvCurves:


kylog:
If the desired config is all in the pooler now (which it looks like, hooray!), can we just dump this file altogether in favor of beaker-hostgenerator?

LuvCurves:
@kylog Yes, we should be able to dump the file, but I wanted to update it appropriately first.

kylog:
Why update the file if we're going to turn around and delete it right away? I'm okay if that's what we need to do, just trying to understand why?

FWIW, here's the output of b-hg. Should this work?
```
 be beaker-hostgenerator redhat7-64m-cumulus25-64a 2> /dev/null
---
HOSTS:
  redhat7-64-1:
    pe_dir:
    pe_ver:
    pe_upgrade_dir:
    pe_upgrade_ver:
    hypervisor: vmpooler
    platform: el-7-x86_64
    template: redhat-7-x86_64
    roles:
    - agent
    - master
  cumulus25-64-1:
    pe_dir:
    pe_ver:
    pe_upgrade_dir:
    pe_upgrade_ver:
    hypervisor: vmpooler
    platform: cumulus-2.5-x86_64
    template: cumulus-vx-25-x86_64
    roles:
    - agent
CONFIG:
  nfs_server: none
  consoleport: 443
  pooling_api: http://vmpooler.delivery.puppetlabs.net/
```

Does that look

joshcooper:
Talked with @LuvCurves in hipchat, going to delete all netdev host configs from puppet and other component repos. We can either close this, or repurpose for deleting the host configs.

PR #4548, opened 2016-01-05
(maint) acceptance: fix calling all functions-split regex

er0ck:
Previous versions of this test overrode the expectation for the split()
function when puppet_version was deemed to be 3.x. This failed in the
3.x job with future parser turned on.  The override has been removed as
it was confusing and the single regex now works for both 3.x jobs and
4.x jobs.

[skip ci]


joshcooper:
Wondering if we should merge this in 3.x, but add a separate PR to puppet#stable to ensure split returns `[9, 8, 7]` since IIUC it will only return 987 in 3.x without future parser?

er0ck:
good idea.  will PR against stable

PR #4550, opened 2016-01-05
(PUP-5584) Force ASCII-8BIT encoding for cached catalogs

branan:
Previously, puppet would attempt to treat JSON read from a cached
catalog as part of the system encoding. This could cause oddities with
how inlined file content is treated.

With this change, catalogs on disk will be treated as ASCII-8BIT. This
essentially tells Ruby to just treat it as a bunch of bytes,
preserving file content and avoiding any encoding issues.

### DIFF
--- a/lib/puppet/indirector/json.rb
+++ b/lib/puppet/indirector/json.rb
@@ -15,7 +15,7 @@ class Puppet::Indirector::JSON < Puppet::Indirector::Terminus
     filename = path(request.key)
     FileUtils.mkdir_p(File.dirname(filename))
 
-    Puppet::Util.replace_file(filename, 0660) {|f| f.print to_json(request.instance) }
+    Puppet::Util.replace_file(filename, 0660) {|f| f.print to_json(request.instance).force_encoding(Encoding::ASCII_8BIT) }
   rescue TypeError => detail
     Puppet.log_exception(detail, "Could not save #{self.name} #{request.key}: #{detail}")
   end
@@ -52,7 +52,7 @@ class Puppet::Indirector::JSON < Puppet::Indirector::Terminus
     json = nil
 
     begin
-      json = File.read(file)
+      json = Puppet::FileSystem.read(file).force_encoding(Encoding::ASCII_8BIT)
     rescue Errno::ENOENT
       return nil
     rescue => detail
--- a/spec/unit/indirector/json_spec.rb
+++ b/spec/unit/indirector/json_spec.rb
@@ -81,7 +81,7 @@ describe Puppet::Indirector::JSON do
           # I don't like this, but there isn't a credible alternative that
           # also works on Windows, so a stub it is. At least the expectation
           # will fail if the implementation changes. Sorry to the next dev.
-          File.expects(:read).with(file).raises(Errno::EPERM)
+          Puppet::FileSystem.expects(:read).with(file).raises(Errno::EPERM)
           expect { subject.find(request) }.
             to raise_error Puppet::Error, /Could not read JSON/
         end


branan:
I'm actually kind of curious if maybe on Windows we get UTF-16 by default... Let me hop into Pry on a windows VM before anybody merges this

Sharpie:
YMMV, but when I tried this on Windows I got IBM-437 instead of any UTF.

branan:
@Sharpie After a bit more testing, I'm seeing even cooler behavior than that - within one run I got "UTF-8" for one JSON blob, and "ASCII-8BIT" for another. No, I don't know how/why this is the case.

I'm not sure there's gonna be good consistency, and it's possible we're really better off using "ASCII-8BIT" in this and just treating cached JSON as a blob of "random data"

lib/puppet/indirector/json.rb
```diff
@@ -52,7 +52,7 @@ def load_json_from_file(file, key)
     json = nil
 
     begin
-      json = File.read(file)
+      json = File.read(file).force_encoding(Encoding::UTF_8)
```
ferventcoder: All FileSystem calls should be going through https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_system/file_impl.rb#L74

ferventcoder: I had started something with Windows where it attempts to read in a few formats and falls back to system encoding.

ferventcoder: Sorry, that link is https://github.com/puppetlabs/puppet/blob/stable/lib/puppet/file_system.rb#L125

ferventcoder: I can probably figure out where I did that and share it - it may have merit with fixing this issue.

ferventcoder: I take it back, it is there https://github.com/puppetlabs/puppet/blob/8bb8c227ff46b6018376a96531921ede65f8a942/lib/puppet/file_system/windows.rb#L96-L102 

However, we may want to consider moving this logic into read.

ferventcoder: Or rename the function :)

ferventcoder: I had started at one point to go into the remaining File calls and redirect them through `Puppet::FileSystem` - I wonder if that should be investigated agai.

joshcooper: Forcing the encoding is only safe if we're positive the contents were originally written as UTF-8, but I'm not sure that's guaranteed. The [`save`](https://github.com/branan/puppet/blob/bug/stable/pup-5584-cached-catalog-encoding/lib/puppet/indirector/json.rb#L18) method calls replace_file, which opens a uniquefile, but doesn't specify an explicit encoding. It then calls `IO#print to_json` which as far as I can tell will transcode the string from the internal default encoding to the external default encoding, but according to http://ruby-doc.org/core-2.1.4/Encoding.html#class-Encoding-label-Internal+encoding, that only happens if the internal default encoding is non-nil. On *nix, I get:

```
$ ruby -e 'p Encoding.default_internal'
nil
$ ruby -e 'p Encoding.default_external'
#<Encoding:UTF-8>
```

On Windows, internal is nil and default external depends on the codepage:

```
C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin> ruby -e 'p Encoding.default_internal'
nil
C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin> ruby -e 'p Encoding.default_external'
#<Encoding:IBM437>
```
In other words, I *think* the `save` method will write the string to the file preserving whatever encoding the string is currently in, which may not be UTF-8. For example, if `to_json` returns a UTF-16LE string (for some reason) whose byte representation is [0xDE, 0x00, 0xAD, 0x00, 0xBE, 0x00, 0xEF, 0x00], then the file would contain those bytes. If we then read the file back in and force encoding to UTF-8, the string would only contain 0xDE, as the next byte would be interpreted as the NULL terminator.

We might need to modify the `save` method to ensure we're always writing a UTF-8 string, e.g.

```ruby
f.print to_json(request.instance).encode('UTF-8')
```

and then force_encoding when we read it back in.

ferventcoder: So the read_preserve actually checks to see if the encoding is valid, would it be even with bad characters in this case? @joshcooper I think I understand that you are saying without saving it in the proper encoding it will come back with a valid encoding incorrectly.

joshcooper: I think we're saying the same thing, but here's an example. Ruby on Windows sets its default external encoding to match the current code page. So suppose we're in code page 1252 and we write out the [euro symbol](http://www.fileformat.info/info/unicode/char/20aC/index.htm). In CP1252, the euro character is a single byte 0x80. When ruby writes the character to disk, it doesn't change the byte representation, so the file on disk contains a single byte 0x80. When we read the file in, we read the contents back in as a CP1252 encoded string (since that's the default external encoding). The string at that point is still valid. If we then force the encoding to UTF-8, we essentially "corrupt" the string since the sequence 0x80 is not a valid UTF-8 encoded string. For example, given:

```ruby
# -*- coding: windows-1252 -*-

puts "encoding internal=#{Encoding.default_internal}, external=#{Encoding.default_external}"

euro = "\u20AC".encode(Encoding::CP1252)
puts "#{euro}=#{euro.unpack('H*')[0]}"

File.open('foo.json', 'w') do |f|
  f.print euro
end

str = File.read('foo.json')
puts "#{str}=#{str.unpack('H*')[0]}"

str.force_encoding('utf-8')
puts "#{str}=#{str.unpack('H*')[0]}"
```

Running this with codepage 1252:

```
C:\users\albert> ruby .\encoding.rb
encoding internal=, external=Windows-1252
€=80
€="80"
�="80"
```

Note how the last string after we forced the encoding is invalid.

The only time we can force encoding UTF-8 is if we're sure the data was written out in UTF-8. So that's why I was saying we should modify the `save` method to explicitly call `String#encode('utf-8')`. For example, if we change it to:

```
  f.print euro.encode('utf-8')
```

Then the `force_encoding` is required:

```
C:\users\albert> ruby .\encoding.rb
encoding internal=, external=Windows-1252
€=80
â‚¬=e282ac
€=e282ac
```

where e2, 82, ac is the correct UTF-8 encoding of the euro symbol.

ferventcoder: Ah, makes sense that we'd force the encoding on save. :+1:

joshcooper: nit, we'd want to encode the string on save:

```
  f.print euro.encode('utf-8')
```

Not force the encoding
```
  euro.force_encoding('utf-8')
  f.print euro
```

joshcooper: @branan do you want to make that change ^? Also I'm wondering if we'd have the same issue with the `yaml` terminus and need to apply the same logic?

branan: @joshcooper Yeah, I can make that change. I've been banging on that facter ticket this week so far.

The YAML terminus probably has the same problem, I'll look in to it as well

kylog: I suggest we deal with the yaml terminus separately since the pain point for this ticket is specific to json (and the yaml situation is arguably more complex to fix - may be a YAGNI).

branan:
@joshcooper this is updated to encode on write, and force encoding on read.

Unfortunately, the YAML code is shared across cached catalog and facts terminii, as well as used for extended CSR attributes. I don't think we want to force the encodings of either of those to UTF-8 (right now, anyway). Given that, and given the fact that we default to JSON and generally don't want to be using YAML for stuff, I'm inclined to say that we just document the limition re: encodings with the YAML cached catalog terminus instead of trying to disentangle our YAML stuff for this ticket.

@ferventcoder I don't think read_preserve_line_endings is what we want to use here - Puppet::Util.replace_file doesn't write in binary, which means line endings will be munged on the write to disk. If we then use preserve_line_endings, we'll end up reading the transformed file, which is not what we want. In addition, the encoding detection isn't necessary if we are enforcing encoding when we write. Is there any other reason to be using the `file_system` stuff in this code path?

ferventcoder:
>  Is there any other reason to be using the file_system stuff in this code path?

@branan All puppet use of file system calls should be going through the file_system stuff. Not specifically the preserve_line_endings bit, but one of the other calls. If that means a method needs to be added to it, then a method should be added there. It guarantees that all file access calls are handled cross platform appropriately.

lib/puppet/indirector/json.rb
ferventcoder: Going through the call I suggested here at the beginning is synonymous with File.read, so you can still call force_encoding on it.

https://github.com/puppetlabs/puppet/blob/stable/lib/puppet/file_system.rb#L125

ferventcoder:
> Puppet::Util.replace_file doesn't write in binary, which means line endings will be munged on the write to disk.

@branan It is read back in in binary because Windows... I also mentioned the name of the function is probably not technically the best earlier - it's actually reading a file with open binary read and try UTF-8 encoding, see if it is valid, fall back to the default_external encoding, see if that is valid, then fall back to just File.read.

@joshcooper followup from the original conversation - the read_preserve_line_endings only looks at a file with no idea what the encoding was and it doesn't attempt to force an encoding. I was originally asking if the read_preserve_line_endings, upon evaluating these badly saved files would come back with something since it does what I mentioned above.

~~~ruby
def read_preserve_line_endings(path)
    contents = path.read( :mode => 'rb', :encoding => Encoding::UTF_8)
    contents = path.read( :mode => 'rb', :encoding => Encoding::default_external) unless contents.valid_encoding?
    contents = path.read unless contents.valid_encoding?

    contents
  end
~~~

joshcooper:
I'm fine skipping 'yaml'.

Using `file_system.read` seems fine.

@ferventcoder `read_preserve_line_endings` is problematic, because it's just guessing what the encoding was when the file was written out. So it's possible for a non-UTF8 encoded file to appear as valid UTF8 when read back in.

branan:
OK, it's updated to use file_system.read here. I'm not gonna go in and make `replace_file` use file_system under the hood for this ticket - way out of scope. So writing is still going straight through to the ruby File stuff AFAICT from scanning that code

ferventcoder:
> OK, it's updated to use file_system.read here.

:+1: 

> I'm not gonna go in and make replace_file use file_system under the hood for this ticket - way out of scope.

That's fair.

ferventcoder:
> `read_preserve_line_endings` is problematic, because it's just guessing what the encoding was when the file was written out. So it's possible for a non-UTF8 encoded file to appear as valid UTF8 when read back in.

@joshcooper Yep, agreed. It's meant for when we don't know what the encoding is supposed to be. In this case we have a deterministic way of knowing so I agree with the approach. :+1:  :D

joshcooper:
I'm seeing a problem when trying to download a catalog, and maybe it's the ASCII-8BIT issue you were seeing earlier. I have a module that contains the chocolatey.config.erb on my master (webrick). When I download the catalog, with this PR, and change my locale on the agent, I get:

```
$ env LC_CTYPE=de_DE.ISO8859-1 bundle exec puppet agent -t --server arcturus --http_debug
...
Error: "\xEF" from ASCII-8BIT to UTF-8
Error: Could not retrieve catalog from remote server: "\xEF" from ASCII-8BIT to UTF-8
```

The problem occurs when trying to serialize the catalog to json: https://github.com/puppetlabs/puppet/blob/stable/lib/puppet/indirector/json.rb#L73-L75. Specifically `object.render('pson')` is returning an `ASCII-8BIT` encoded string, preserving the BOM character, so then the `encode('utf-8')` fails.

Interestingly, the upstream pure json library does the following when converting [UTF-8 strings to json](https://github.com/flori/json/blob/master/lib/json/pure/generator.rb#L46):

```ruby
string = string.dup
string.force_encoding(::Encoding::ASCII_8BIT)
string.gsub!(/["\\\x0-\x1f]/) { MAP[$&] }
string.force_encoding(::Encoding::UTF_8)
string
```

But our [vendored pure ruby](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/external/pson/pure/generator.rb#L42-L50) is missing the last `force_encoding` line. So our vendored pure ruby is always returning ASCII-8BIT strings, which seems wrong:

```
string = string.dup
string << '' # XXX workaround: avoid buffer sharing
string.force_encoding(Encoding::ASCII_8BIT)
string.gsub!(/["\\\x0-\x1f]/) { MAP[$MATCH] }
string
```

So a couple of things:

* Adding `string.force_encoding(::Encoding::UTF_8)` to our vendored generator seems to do the trick.
* I'm surprised our vendored PSON library has been "leaking" ASCII-8BIT strings. I'd assume that would cause problems elsewhere in puppet.
* I wonder what other fixes have been made to upstream pure ruby

branan:
@joshcooper I wonder if it actually makes sense for us to just cache the catalogs as ASCI-8BIt instead of UTF-8? It shouldn't affect the structure of the document at all, and in cases where we have weirdly-encoded stuff in file content it'll just pass it through.

joshcooper:
>  I wonder if it actually makes sense for us to just cache the catalogs as ASCI-8BIt 

I think that should work and is probably fine for now. I'll take a closer look.

> especially since PSON likes to be all ASCII-8BIT-y

I think it only uses `ASCII-8BIT` so that it can apply a regex to individual bytes (not chars), so I see it as an "internal implementation detail" that the caller shouldn't be aware of. Also the [`utf_to_json`](https://github.com/flori/json/blob/master/lib/json/pure/generator.rb#L42) method says: 

>  # Convert a UTF8 encoded Ruby string _string_ to a JSON string, encoded with
>  # UTF16 big endian characters as \u????, and return it.

So it accepts a UTF8 encoded string and represents UTF16 chars using \u, so the resulting string is valid UTF8:

```
irb(main):002:0> "hello \u20AC"
=> "hello €"
irb(main):003:0> "hello \u20AC".encoding
=> #<Encoding:UTF-8>
```

So I think the method should be returning valid UTF8, but I'm fine making that change later when we have more time to test.

joshcooper:
> I think that should work and is probably fine for now. I'll take a closer look.

Changing the save method to force encoding to ASCII-8BIT works for me (both when creating a new file, update existing file, and when running off cache). So yeah, let's update this PR.

branan:
@joshcooper this is updated to use ASCII-8BIT now.

We can't actually `.encode` to ASCII-8BIT on write - ruby rightly barfs when there's a unicode codepoint greater than 255 in the string. The force_encoding there isn't strictly necessary, since we're just dumping it to a file immediately, but I feel like it signals that we intend to treat cached catalogs as just a byte stream at all times, so I left it in.

This works right for me with with Charlie's dataset on Linux - cached catalog still parses as UTF-8 bytes when I examine it, and Puppet doesn't change anything when I force the cached catalog to be used.

Sharpie:
> We can't actually .encode to ASCII-8BIT on write

Is this a situation where we should use `File.open(..., 'wb')` where the `b` mode disables encoding and line ending conversions?

Quoth [the docs](http://ruby-doc.org/core-2.1.7/IO.html#method-c-new):

> "b"  Binary file mode
     Suppresses EOL <-> CRLF conversion on Windows. And
     sets external encoding to ASCII-8BIT unless explicitly
     specified.

branan:
@Sharpie as long as we write and read with the same use (or lack thereof) of 'b', it'll match and not matter. I think it's worth it to let line endings be swapped on write, to keep the files more human-readable on Windows with editors that don't understand Unix line-endings.

Sharpie:
Well, if an encoding conversion happens on write and then again on read, there could be issues as not all  encoding pairs are reversible. Using the 'b' mode is one way to ensure conversions don't happen at all.

branan:
@joshcooper Per our chat before lunch, I've double-checked the encoding of the catalog on the wire. It looks like even on Linux the catalog PSON is being detected as "ASCII-8BIT" from the HTTP request. Given that I think we want to use this as-is, instead of going back to UTF-8

```
[5] pry(#<Puppet::Resource::Catalog::Rest>)> uri
=> "/puppet/v3/catalog/localhost"
[6] pry(#<Puppet::Resource::Catalog::Rest>)> body.encoding
=> #<Encoding:ASCII-8BIT>
```

branan:
@Sharpie the only thing that gets munged in text mode is newlines, and that is totally symmetric and lossless. I think we actually *want* that munging as it makes it easier to view the JSON file with a windows text editor

ferventcoder:
Everything here is good. The only statement I disagree with is:

> I think it's worth it to let line endings be swapped on write, to keep the files more human-readable on Windows with editors that don't understand Unix line-endings.

There are two things to consider here

 * You are making this change on read, but not saving it back to the file. If the file is checksummed, it could have unintended consequences.
* This is not being done with any customer files. 

Since this is likely only used as an internal file for Puppet, swapping line endings at read time only may not cause any real harm. But if it was anything user facing like their files, then I would suggest not changing the files at all. We don't know all of the scenarios in which customer files can break based on making any change to the file. In most cases changing line endings causes undesired behavior - see https://tickets.puppetlabs.com/browse/PUP-927.

With the following two constraints in mind, it is almost never okay to swap line endings. Perhaps this is a case where it is.

ferventcoder:
> I think we actually want that munging as it makes it easier to view the JSON file with a windows text editor

There is only one text editor that I know of that has these issues (notepad.exe). Almost no other editor on Windows has this issue. One of the first things that Windows folks find is an editor replacement that doesn't lack even the simplest of features (not to mention just understanding simply `LF` is also a line ending).

Iristyle:
Quick additional note about this PR for anyone else who comes along and is confused about the name of the encoding used, `Encoding::ASCII_8BIT` is an alias for `Encoding::BINARY` in Ruby.

```
[2] pry(main)> Encoding::ASCII_8BIT == Encoding::BINARY
=> true
```

PR #4555, opened 2016-01-07
{WIP}(PUP-5538) User and group failures when using Unicode in Windows

glennsarti:
The win32-security gem that Puppet uses for SID conversion uses ANSI versions of APIs that can make it fail to convert Unicode group / user names to SIDs properly.


ferventcoder:
At least you have failing tests up :+1:

PR #4558, opened 2016-01-08
(maint) Fix issue with PUP-5445

LuvCurves:


acceptance/setup/aio/pre-suite/021_InstallAristaModule.rb
```diff
@@ -12,6 +12,8 @@
 
   step 'add puppet user to switch' do
     switchs.each do |switch|
+      on(switch, "useradd -U puppet")
+      on(switch, "useradd -U missinguser")
```
joshcooper: If we create the `missinguser` then we're subverting what the test is trying to do. I'm thinking it's better to just skip the test on arista.

acceptance/setup/aio/pre-suite/021_InstallAristaModule.rb
```diff
@@ -12,6 +12,8 @@
 
   step 'add puppet user to switch' do
     switchs.each do |switch|
+      on(switch, "useradd -U puppet")
```
joshcooper: If we create the `puppet` user, then perhaps it would be easier to create the `puppet` group and remove all of the other workarounds below? The downside there is we're not really running puppet acceptance in a valid environment, since puppet-agent doesn't create the `puppet` user/group. 

Another option, don't create the `puppet` user and skip the tests that try to run with an alternate `confdir`?

I have a mild preference for the latter, since I think it would be better to execute tests in a realistic scenario without the puppet user, than it is to create the puppet user and run all of the tests (possibly invalidating all tests).

That said, I'm fine if you want to merge as is. I'd really rather see Arista fix their bug, because who knows what else is expecting Etc.getpwnam to return nil for non-existent accounts.

acceptance/setup/aio/pre-suite/021_InstallAristaModule.rb
LuvCurves: I agree, I just looked closer at the test, 'missinguser' should go

acceptance/setup/aio/pre-suite/021_InstallAristaModule.rb
LuvCurves: @joshcooper & I chatted. Arista will have to inform customers to add puppet user and group.  I will remove the 'missinguser' and add confines to the other tests

LuvCurves: BTW, the line "useradd -U puppet" adds the puppet user and group (the -U flag adds group)

joshcooper: cool, thanks!

PR #4560, opened 2016-01-08
(PUP-5538) Rewrite SID conversions

Iristyle:
* Use Windows API directly rather than consuming win32-security gem

Still left to do:

- [x] break up commits in pieces of possible to make changes clearer
- [ ] remove `Win32::Security` constants
- [ ] add manifest based acceptance test (may need to create a user with raw bytes in a PowerShell call if the manifest itself won't properly handle UTF8)
- [x] verify that codepaths are necessary / simplify any codepaths where necessary
- [x] potentially rename methods that return hashes now instead of the `SID` object and deprecate old methods
- [ ] add some better docs around API calls noting what expected inputs / outputs are
- [x] track down answers to any TODOs

The constants can be removed later.  I think we can get away without an acceptance test, unless someone wants to write one.  Docs are probably OK for now as well.

### DIFF
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -59,14 +59,16 @@ module Puppet::Util::Windows::ADSI
     # This method should *only* be used to generate WinNT://<SID> style monikers
     # used for IAdsGroup::Add / IAdsGroup::Remove.  These URIs are not useable
     # to resolve an account with WIN32OLE.connect
+    # Valid input is a SID::Principal, S-X-X style SID string or any valid
+    # account name with or without domain prefix
     # @api private
     def sid_uri_safe(sid)
-      return sid_uri(sid) if sid.kind_of?(Win32::Security::SID)
+      return sid_uri(sid) if sid.kind_of?(Puppet::Util::Windows::SID::Principal)
 
       begin
-        sid = Win32::Security::SID.new(Win32::Security::SID.string_to_sid(sid))
+        sid = Puppet::Util::Windows::SID.name_to_sid_object(sid)
         sid_uri(sid)
-      rescue SystemCallError
+      rescue Puppet::Util::Windows::Error, Puppet::Error
         nil
       end
     end
@@ -75,8 +77,9 @@ module Puppet::Util::Windows::ADSI
     # used for IAdsGroup::Add / IAdsGroup::Remove.  These URIs are not useable
     # to resolve an account with WIN32OLE.connect
     def sid_uri(sid)
-      raise Puppet::Error.new( "Must use a valid SID object" ) if !sid.kind_of?(Win32::Security::SID)
-      "WinNT://#{sid.to_s}"
+      raise Puppet::Error.new( "Must use a valid SID::Principal" ) if !sid.kind_of?(Puppet::Util::Windows::SID::Principal)
+
+      "WinNT://#{sid.sid}"
     end
 
     def uri(resource_name, resource_type, host = '.')
@@ -140,7 +143,7 @@ module Puppet::Util::Windows::ADSI
       sids = names.map do |name|
         sid = Puppet::Util::Windows::SID.name_to_sid_object(name)
         raise Puppet::Error.new( "Could not resolve name: #{name}" ) if !sid
-        [sid.to_s, sid]
+        [sid.sid, sid]
       end
 
       Hash[ sids ]
@@ -251,20 +254,12 @@ module Puppet::Util::Windows::ADSI
 
 
     def add_group_sids(*sids)
-      group_names = []
-      sids.each do |sid|
-        group_names << Puppet::Util::Windows::SID.sid_to_name(sid)
-      end
-
+      group_names = sids.map { |s| s.domain_account }
       add_to_groups(*group_names)
     end
 
     def remove_group_sids(*sids)
-      group_names = []
-      sids.each do |sid|
-        group_names << Puppet::Util::Windows::SID.sid_to_name(sid)
-      end
-
+      group_names = sids.map { |s| s.domain_account }
       remove_from_groups(*group_names)
     end
 
@@ -277,7 +272,7 @@ module Puppet::Util::Windows::ADSI
 
       desired_groups = desired_groups.split(',').map(&:strip)
 
-      current_hash = Hash[ self.group_sids.map { |sid| [sid.to_s, sid] } ]
+      current_hash = Hash[ self.group_sids.map { |sid| [sid.sid, sid] } ]
       desired_hash = self.class.name_sid_hash(desired_groups)
 
       # First we add the user to all the groups it should be in but isn't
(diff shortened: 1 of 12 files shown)


ferventcoder:
Will also need to verify with ACL module (and others)

ferventcoder:
Need to verify this will work with ACL module and others.

Iristyle:
Requires #4564

lib/puppet/util/windows/sid.rb
```diff
@@ -24,14 +24,26 @@ def name_to_sid(name)
...
+      begin
+        string_to_sid_ptr(name) do |sid_ptr|
+          valid = ! sid_ptr.nil? && ! sid_ptr.null?
+          raw_sid_bytes = sid_ptr.read_array_of_uchar(get_length_sid(sid_ptr))
+        end
+      rescue
+      end
+
+      # TODO: using . for domain seems wrong if we accept DOMAIN\USER for name
+      raw_sid_bytes ? lookup_account_sid(nil, raw_sid_bytes) : lookup_account_name('.', name)
```
ferventcoder: This jumped out at me.

lib/puppet/util/windows/adsi.rb
```diff
@@ -491,14 +488,14 @@ def self.create(name)
...
-        return true if sid.account_type == 'group'
+        return true if sid[:type] == :SidTypeGroup
 
         # 'well known group' is special as it can be a group like Everyone OR a user like SYSTEM
         # so try to resolve it
         # https://msdn.microsoft.com/en-us/library/cc234477.aspx
-        well_known = sid.account_type == 'well known group'
-        return false if sid.account_type != 'alias' && !well_known
-        name_or_sid = "#{sid.domain}\\#{sid.account}"
+        well_known = sid[:type] == :SidTypeWellKnownGroup
```
ferventcoder: The name of this should be `account_type` instead of `type`. Correct me if I am wrong, but the call to `sid.domain` would still resolve to the hashed value `sid[:domain]` in ruby? 

If so we can ensure compat with `sid.account_type` if used elsewhere and thus why requested that it match the old name.

lib/puppet/util/windows/sid.rb
```diff
@@ -131,6 +157,110 @@ def valid_sid?(string_sid)
...
+                   raise Puppet::Util::Windows::Error.new('Failed to call LookupAccountNameW')
+                  end
+
+                  domain = domain_ptr.read_wide_string(domain_length_ptr.read_dword)
+
+                  return { :system => system_name,
+                     :account => account_name,
+                     :sid_bytes => sid_ptr.read_bytes(sid_length_ptr.read_dword),
+                     :sid => sid_ptr_to_string(sid_ptr),
+                     :domain_account => domain && !domain.empty? ? "#{domain}\\#{account_name}" : account_name,
```
ferventcoder: Shouldn't we also include `:domain` here with a value of `.`? Otherwise the two returns do not match.

lib/puppet/util/windows/sid.rb
Iristyle: Yeah, @glennsarti and I discussed this and decided we're going to file a new ticket for that particular problem, since we don't believe it has ever worked.

lib/puppet/util/windows/adsi.rb
Iristyle: No, Ruby hashes don't work like that (I think you might be thinking of PowerShell).

```
irb(main):001:0> foo = { :bar => 2, :baz => 'hello' }
=> {:bar=>2, :baz=>"hello"}
irb(main):002:0> foo[:bar]
=> 2
irb(main):003:0> foo.bar
NoMethodError: undefined method `bar' for {:bar=>2, :baz=>"hello"}:Hash
        from (irb):3
        from C:/tools/ruby215/bin/irb:11:in `<main>'
irb(main):004:0>
```

ferventcoder: PowerShell and Javascript. Too many languages :trolleybus:

ferventcoder:
Verified against the ACL module with commits as of this time.

~~~sh
c:\code\puppetlabs\modules\puppetlabs-acl> bundle install --path .bundle/win --without system_tests
~~~

Replace the `.bundle/win/ruby/1.9.1/gems/4.3.1-x86-mingw32` with the PR code and run

~~~sh
c:\code\puppetlabs\modules\puppetlabs-acl> bundle exec rspec spec
364/364 |=== 100 ====>| Time: 00:00:17

Finished in 17.39 seconds
364 examples, 0 failures
~~~

Then I replaced the installed puppet with the PR code, ran the manual code (below) and verified the output against what the file states each should output with icacls.

~~~sh
 c:\code\puppetlabs\modules\puppetlabs-acl> puppet apply tests\init.pp --debug
~~~

All good :+1:

Iristyle:
@puppetlabs/windows please try to break this.  It should be ready for merge now.

lib/puppet/util/windows/principal.rb
```diff
@@ -0,0 +1,177 @@
...
+              FFI::MemoryPointer.new(:dword, 1) do |domain_length_ptr|
+                FFI::MemoryPointer.new(:uint32, 1) do |name_use_enum_ptr|
+
+                sid_length_ptr.write_dword(MAXIMUM_SID_BYTE_LENGTH)
+                success = LookupAccountNameW(system_name_ptr, account_name_ptr, sid_ptr, sid_length_ptr,
+                  FFI::Pointer::NULL, domain_length_ptr, name_use_enum_ptr)
+                last_error = FFI.errno
+
+                if (success == FFI::WIN32_FALSE && last_error != ERROR_INSUFFICIENT_BUFFER)
+                  raise Puppet::Util::Windows::Error.new('Failed to call LookupAccountNameW', last_error)
```
glennsarti: If a valid SID that does not exist last_error is 1332 (No mapping between account....) whereas is you use a byte array that is not a valid SID e.g. [0] or [5] (anything but [1]), last_error is 87 which is invalid parameter.  Should the two scenarios be surfaced differently (Error code 1332 meant the API was called successfully but the SID supplied didn't exist)

spec/unit/util/windows/principal_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+      principal = Puppet::Util::Windows::SID::Principal.lookup_account_sid(administrator_bytes)
+      expect(principal.account).to eq('Administrators')
+      expect(principal.sid_bytes).to eq(administrator_bytes)
+      expect(principal.sid).to eq('S-1-5-32-544')
+      expect(principal.domain).to eq('BUILTIN')
+      expect(principal.domain_account).to eq('BUILTIN\\Administrators')
+      expect(principal.account_type).to eq(:SidTypeAlias)
+    end
+
+    it "should raise an error when trying to lookup a SID that doesn't exist" do
```
glennsarti: Should add a test for looking up an invalid SID e.g. [0] or [5] (just not [1])

lib/puppet/util/windows/principal.rb
Iristyle: I don't think this is necessary given the `Windows::Error.new` automatically grabs the error code and looks it up.

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/error.rb#L13

spec/unit/util/windows/principal_spec.rb
Iristyle: Thanks - added.

PR #4565, opened 2016-01-11
(PUP-1985) Enable parameter default to reference earlier parameter

thallgren:
A parameter in a resource or class declaration was not in scope for
subsequent parameters in the same parameter list. This commit changes
that by introducing a temporary parameter scope during parameter default
evaluation. The temporary scope also forms the hash that is used for the
subsequent validation of all the parameters which means that this commit
should have a zero impact from a performance perspective.


hlindberg:
So this makes it possible to use Ephemeral and Local scopes where the "scope" is passed to logic that uses the lookupvar(key, options) API. I assume it did not work without that.

I understand why it was done this way as the default could contain any expression and using anything else than scope would be out of the question.

This PR then touches puppet/resource/type - shared with the agent code, and that is always scary. We also want default in functions to work the same way, I don't think the solution in this PR covers that.

Since this touches resource/type there would need to be a lot more tests.

thallgren:
When you say "default in functions", I assume you mean functions written in the puppet language only, not functions written using dispatch and ruby?

"A lot more tests" sounds fun. What should they cover?

hlindberg:
You are right, functions in PP supports default values, in RB you do that a different way (with defaults in the implementation methods).

Here is one surprising case. There are other variations on this theme:
```
$a = surprise
function foo($a = 1, $b = $a) {
  notice "values are $a, and $b"
}
function bar() {
  $a = wtf
  foo()
}

bar()
```

thallgren:
Closing as we retarget to master. See [PR-4568](https://github.com/puppetlabs/puppet/pull/4568)

PR #4566, opened 2016-01-11
(maint) Acceptance Rake - give CONFIG priority

johnduarte:
This commit changes the logic in the beaker_test rake task for
acceptance giving BEAKER_HOSTS priority over TEST_TARGET.
It also changes the name of the ENV variable from CONFIG to
BEAKER_HOSTS

In the case that a static beaker host config file is needed for
running acceptance testing, the file path to this config should
be assigned to the ENV variable BEAKER_HOSTS. This value
will take precedence over any value set for TEST_TARGET.

TEST_TARGET is still used to supply a string for dynamically
generating beaker host configs.

skip:ci

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -117,17 +117,17 @@ EOS
   tests_opt = "--tests=#{tests}" if tests
 
   agent_target = ENV['TEST_TARGET']
-  if agent_target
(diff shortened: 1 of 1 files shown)


joshcooper:
:+1:

PR #4567, opened 2016-01-13
(maint) Acceptance - ensure test aix packages absent

johnduarte:
This commit adds a step to the aix nim_package_provider test to
ensure that all test packages are removed prior to the other steps
in the test.

[skip ci]

### DIFF
--- a/acceptance/tests/aix/nim_package_provider.rb
+++ b/acceptance/tests/aix/nim_package_provider.rb
@@ -61,6 +61,14 @@ package_types = {
     }
 }
 
+step "Setup: ensure test packages are not installed" do
+  pkgs = ['cdrecord', 'bos.atm.atmle']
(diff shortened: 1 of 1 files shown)


whopper:
Huh, I wonder why the teardown step isn't doing the trick?

```
teardown do
    test_apply('cdrecord', 'absent', '')
    test_apply('bos.atm.atmle', 'absent', '')
end
```

Theoretically, this would cause the test to fail when tearing down, since `test_apply` validates that the desired action occurred.

johnduarte:
The teardown step works, but if the environment is not clean prior to the first run of the test it will fail because the teardown only runs after the test. On subsequent runs, the test passes because the teardown has done its job.

whopper:
:+1:, passes on pe-aix-61-agent

PR #4568, opened 2016-01-13
(PUP-1985) Fix correct scope for parameter default evaluation

thallgren:
This commit ensures that a correct scope is used when evaluating
parameter default expressions. It impacts the parameter declarations
of functions, EPP templates, classes, and resources.

A new ephemeral scope `ParameterScope` was added to keep track of
which parameters that exists and in which order they are evaluated
and accessed. This scope is only in effect during the evaluation
of the parameter defaults. Since the parameters are evaluated left
to right, and each parameter is added to the scope, a default
expression of a parameter can use the parameters declared to the
left.

Each parameter in turn is evaluated using a private MatchScope to
ensure that the evaluation of one parameter cannot interfere with
another. No LocalScope is pushed though, since it's forbidden
to assign new variables in this context.

The commit also contain two new test suites (one for function and
epp-template parameters, and one for class and resource parameters).
They all test default parameter evaluation much in the same way but
for different targets.

### DIFF
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -174,6 +174,94 @@ class Puppet::Parser::Scope
 
   end
 
+  # @api private
+  class ParameterScope < Ephemeral
+    class Access
+      attr_accessor :value
+
+      def assigned?
+        instance_variable_defined?(:@value)
+      end
+    end
+
+    # A parameter default must be evaluated using a special scope. The scope that is given to this method must
+    # have a `ParameterScope` as its last ephemeral scope. This method will then push a `MatchScope` while the
+    # given `expression` is evaluated. The method will catch any throw of `:unevaluated_parameter` and produce
+    # an error saying that the evaluated parameter X tries to access the unevaluated parameter Y.
+    #
+    # @param name [String] the name of the currently evaluated parameter
+    # @param expression [Puppet::Parser::AST] the expression to evaluate
+    # @param scope [Puppet::Parser::Scope] a scope where a `ParameterScope` has been pushed
+    # @return [Object] the result of the evaluation
+    #
+    # @api private
+    def evaluate3x(name, expression, scope)
+      scope.with_guarded_scope do
+        bad = catch(:unevaluated_parameter) do
+          scope.new_match_scope(nil)
+          return as_read_only { expression.safeevaluate(scope) }
+        end
+        raise Puppet::Error, "default expression for $#{name} tries to illegally access not yet evaluated $#{bad}"
+      end
+    end
+
+    def evaluate(name, expression, scope, evaluator)
+      scope.with_guarded_scope do
+        bad = catch(:unevaluated_parameter) do
+          scope.new_match_scope(nil)
+          return as_read_only { evaluator.evaluate(expression, scope) }
+        end
+        raise Puppet::Error, "default expression for $#{name} tries to illegally access not yet evaluated $#{bad}"
+      end
+    end
+
+    def initialize(parent, param_names)
+      super(parent)
+      @params = {}
+      param_names.each { |name| @params[name] = Access.new }
+    end
+
+    def [](name)
+      access = @params[name]
+      return super if access.nil?
+      throw(:unevaluated_parameter, name) unless access.assigned?
+      access.value
+    end
+
+    def []=(name, value)
+      raise Puppet::Error, "Attempt to assign variable #{name} to read-only parameter scope" if @read_only
+      @params[name] ||= Access.new
+      @params[name].value = value
+    end
+
+    def bound?(name)
+      @params.include?(name)
+    end
+
+    def include?(name)
+      @params.include?(name) || super
+    end
+
+    def is_local_scope?
+      true
+    end
+
+    def as_read_only
+      read_only = @read_only
+      @read_only = true
+      begin
+        yield
+      ensure
+        @read_only = read_only
+      end
+    end
+
+    def to_hash
+      Hash[@params.select {|_, access| access.assigned? }.map { |name, access| [name, access.value] }]
+    end
+  end
+
+
   # Returns true if the variable of the given name has a non nil value.
   # TODO: This has vague semantics - does the variable exist or not?
   #       use ['name'] to get nil or value, and if nil check with exist?('name')
@@ -738,15 +826,33 @@ class Puppet::Parser::Scope
 
   alias_method :inspect, :to_s
 
-  # remove ephemeral scope up to level
-  # TODO: Who uses :all ? Remove ??
+  # Pop ephemeral scopes up to level and return them
   #
+  # @deprecated use #pop_epehemeral
+  # @api private
   def unset_ephemeral_var(level=:all)
+    Puppet.deprecation_warning('Method Parser::Scope#unset_ephemeral_var() is deprecated')
     if level == :all
       @ephemeral = [ MatchScope.new(@symtable, nil)]
     else
       @ephemeral.pop(@ephemeral.size - level)
     end
+  end
+
+  # Pop ephemeral scopes up to level and return them
+  #
+  # @param level [Fixnum] a positive integer
+  # @return [Array] the removed ephemeral scopes
+  # @api private
+  def pop_ephemerals(level)
+    @ephemeral.pop(@ephemeral.size - level)
+  end
+
+  # Push ephemeral scopes onto the ephemeral scope stack
+  # @param ephemeral_scopes [Array]
+  # @api private
+  def push_ephemerals(ephemeral_scopes)
+    ephemeral_scopes.each { |ephemeral_scope| @ephemeral.push(ephemeral_scope) } unless ephemeral_scopes.nil?
   end
 
   def ephemeral_level
@@ -759,6 +865,53 @@ class Puppet::Parser::Scope
       @ephemeral.push(LocalScope.new(@ephemeral.last))
     else
       @ephemeral.push(MatchScope.new(@ephemeral.last, nil))
+    end
+  end
+
+  # Execute given block in global scope with no ephemerals present
+  #
+  # @yieldparam [Scope] global_scope the global and ephemeral less scope
+  # @return [Object] the return of the block
+  #
+  # @api private
+  def with_global_scope(&block)
+    find_global_scope.without_ephemeral_scopes(&block)
+  end
+
+  # Execute given block with all ephemeral popped from the ephemeral stack
+  #
+  # @api private
+  def without_ephemeral_scopes
+    save_ephemeral = @ephemeral
+    begin
+      @ephemeral = [ @symtable ]
+      yield(self)
+    ensure
+      @ephemeral = save_ephemeral
+    end
+  end
+
+  # Nests a parameter scope
+  # @api private
+  def with_parameter_scope(param_names)
+    param_scope = ParameterScope.new(@ephemeral.last, param_names)
+    with_guarded_scope do
+      @ephemeral.push(param_scope)
+      yield(param_scope)
+    end
+  end
+
+  # Execute given block and ensure that ephemeral level is restored
+  #
+  # @return [Object] the return of the block
+  #
+  # @api private
+  def with_guarded_scope
+    elevel = ephemeral_level
+    begin
+      yield
+    ensure
+      pop_ephemerals(elevel)
     end
   end
 
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -229,7 +229,7 @@ class Puppet::Parser::Scope
     end
 
     def []=(name, value)
-      raise Puppet::Error, "Attempt to assign variable #{name} to read-only parameter scope" if @read_only
+      raise Puppet::Error, "Attempt to assign variable #{name} when evaluating parameters" if @read_only
       @params[name] ||= Access.new
       @params[name].value = value
     end
--- a/lib/puppet/pops/evaluator/closure.rb
+++ b/lib/puppet/pops/evaluator/closure.rb
@@ -25,22 +25,7 @@ class Puppet::Pops::Evaluator::Closure < Puppet::Pops::Evaluator::CallableSignat
   # Evaluates a closure in its enclosing scope after having matched given arguments with parameters (from left to right)
   # @api public
   def call(*args)
-    variable_bindings = combine_values_with_parameters(args)
-
-    tc = Puppet::Pops::Types::TypeCalculator.singleton
-    final_args = tc.infer_set(parameters.reduce([]) do |tmp_args, param|
-      if param.captures_rest
-        tmp_args.concat(variable_bindings[param.name])
-      else
-        tmp_args << variable_bindings[param.name]
-      end
-    end)
-
-    if type.callable?(final_args)
-      @evaluator.evaluate_block_with_bindings(@enclosing_scope, variable_bindings, @model.body)
-    else
-      raise ArgumentError, Puppet::Pops::Types::TypeMismatchDescriber.describe_signatures(closure_name, [self], final_args)
-    end
+    call_with_scope(@enclosing_scope, args)
   end
 
   # This method makes a Closure compatible with a Dispatch. This is used when the closure is wrapped in a Function
@@ -48,18 +33,27 @@ class Puppet::Pops::Evaluator::Closure < Puppet::Pops::Evaluator::CallableSignat
   # checks of the argument type/arity validity).
   # @api private
   def invoke(instance, calling_scope, args, &block)
-    call(*args, &block)
+    @enclosing_scope.with_global_scope do |global_scope|
+      call_with_scope(global_scope, args, &block)
+    end
   end
 
   # Call closure with argument assignment by name
   def call_by_name(args_hash, enforce_parameters)
     if enforce_parameters
-      args_hash = args_hash.dup
-      parameters.each do |p|
-        name = p.name
-        # only set result of default expr if it is defined (it is otherwise not possible to differentiate
-        # between explicit undef and no default expression
-        args_hash[name] = @evaluator.evaluate(p.value, @enclosing_scope) if args_hash[name].nil? && !p.value.nil?
+      # Push a temporary parameter scope used while resolving the parameter defaults
+      @enclosing_scope.with_parameter_scope(parameter_names) do |param_scope|
+        args_hash.each { |k, v| param_scope[k] = v unless v.nil? && parameter_names.include?(k) }
+        parameters.each do |p|
+          name = p.name
+          # only set result of default expr if it is defined (it is otherwise not possible to differentiate
+          # between explicit undef and no default expression
+          arg = args_hash[name]
+          if arg.nil? && !p.value.nil?
+            param_scope[name] = param_scope.evaluate(name, p.value, @enclosing_scope, @evaluator)
+          end
+        end
+        args_hash = param_scope.to_hash
       end
       Puppet::Pops::Types::TypeMismatchDescriber.validate_parameters(closure_name, params_struct, args_hash)
     end
@@ -125,52 +119,70 @@ class Puppet::Pops::Evaluator::Closure < Puppet::Pops::Evaluator::CallableSignat
 
   private
 
-  def combine_values_with_parameters(args)
-    variable_bindings = {}
+  def call_with_scope(scope, args)
+    variable_bindings = combine_values_with_parameters(scope, args)
 
-    parameters.each_with_index do |parameter, index|
-      param_captures     = parameter.captures_rest
-      default_expression = parameter.value
-
-      if index >= args.size
-        if default_expression
-          # not given, has default
-          value = @evaluator.evaluate(default_expression, @enclosing_scope)
-          if param_captures && !value.is_a?(Array)
-            # correct non array default value
-            value = [value]
-          end
-        else
-          # not given, does not have default
-          if param_captures
-            # default for captures rest is an empty array
-            value = []
-          else
-            @evaluator.fail(Puppet::Pops::Issues::MISSING_REQUIRED_PARAMETER, parameter, { :param_name => parameter.name })
-          end
-        end
+    tc = Puppet::Pops::Types::TypeCalculator.singleton
+    final_args = tc.infer_set(parameters.reduce([]) do |tmp_args, param|
+      if param.captures_rest
+        tmp_args.concat(variable_bindings[param.name])
       else
-        given_argument = args[index]
-        if param_captures
-          # get excess arguments
-          value = args[(parameter_count-1)..-1]
-          # If the input was a single nil, or undef, and there is a default, use the default
-          # This supports :undef in case it was used in a 3x data structure and it is passed as an arg
-          #
-          if value.size == 1 && (given_argument.nil? || given_argument == :undef) && default_expression
-            value = @evaluator.evaluate(default_expression, @enclosing_scope)
-            # and ensure it is an array
-            value = [value] unless value.is_a?(Array)
+        tmp_args << variable_bindings[param.name]
+      end
+    end)
+
+    if type.callable?(final_args)
+      @evaluator.evaluate_block_with_bindings(scope, variable_bindings, @model.body)
+    else
+      raise ArgumentError, Puppet::Pops::Types::TypeMismatchDescriber.describe_signatures(closure_name, [self], final_args)
+    end
+  end
+
+  def combine_values_with_parameters(scope, args)
+    scope.with_parameter_scope(parameter_names) do |param_scope|
+      parameters.each_with_index do |parameter, index|
+        param_captures     = parameter.captures_rest
+        default_expression = parameter.value
+
+        if index >= args.size
+          if default_expression
+            # not given, has default
+            value = param_scope.evaluate(parameter.name, default_expression, scope, @evaluator)
+
+            if param_captures && !value.is_a?(Array)
+              # correct non array default value
+              value = [value]
+            end
+          else
+            # not given, does not have default
+            if param_captures
+              # default for captures rest is an empty array
+              value = []
+            else
+              @evaluator.fail(Puppet::Pops::Issues::MISSING_REQUIRED_PARAMETER, parameter, { :param_name => parameter.name })
+            end
           end
         else
-          value = given_argument
+          given_argument = args[index]
+          if param_captures
+            # get excess arguments
+            value = args[(parameter_count-1)..-1]
+            # If the input was a single nil, or undef, and there is a default, use the default
+            # This supports :undef in case it was used in a 3x data structure and it is passed as an arg
+            #
+            if value.size == 1 && (given_argument.nil? || given_argument == :undef) && default_expression
+              value = param_scope.evaluate(parameter.name, default_expression, scope, @evaluator)
+              # and ensure it is an array
+              value = [value] unless value.is_a?(Array)
+            end
+          else
+            value = given_argument
+          end
         end
+        param_scope[parameter.name] = value
       end
-
-      variable_bindings[parameter.name] = value
+      param_scope.to_hash
     end
-
-    variable_bindings
   end
 
   def create_callable_type()
--- a/lib/puppet/resource/type.rb
+++ b/lib/puppet/resource/type.rb
@@ -174,12 +174,9 @@ class Puppet::Resource::Type
 
     if code
       if @match # Only bother setting up the ephemeral scope if there are match variables to add into it
-        begin
-          elevel = scope.ephemeral_level
+        scope.with_guarded_scope do
           scope.ephemeral_from(@match, file, line)
           code.safeevaluate(scope)
-        ensure
-          scope.unset_ephemeral_var(elevel)
         end
       else
         code.safeevaluate(scope)
@@ -346,6 +343,9 @@ class Puppet::Resource::Type
   end
 
   # Validate and set any arguments passed by the resource as variables in the scope.
+  #
+  # This method is known to only be used on the server/compile side.
+  #
   # @param resource [Puppet::Parser::Resource] the resource
   # @param scope [Puppet::Parser::Scope] the scope
   #
@@ -360,9 +360,6 @@ class Puppet::Resource::Type
     resource.add_parameters_from_consume
     inject_external_parameters(resource, scope)
 
-    resource_hash = {}
-    resource.each { |k, v| resource_hash[k.to_s] = v.value unless k == :name || k == :title }
-
     if @type == :hostclass
       scope[TITLE] = resource.title.to_s.downcase
       scope[NAME] =  resource.name.to_s.downcase
@@ -370,12 +367,21 @@ class Puppet::Resource::Type
       scope[TITLE] = resource.title
       scope[NAME] =  resource.name
     end
-
     scope.class_set(self.name,scope) if hostclass? || node?
 
-    assign_defaults(resource, scope, resource_hash)
-    validate_resource_hash(resource, resource_hash)
-    resource_hash.each { |param, value| exceptwrap { scope[param] = value }}
+    param_hash = scope.with_parameter_scope(arguments.keys) do |param_scope|
+      # Assign directly to the parameter scope to avoid scope parameter validation at this point. It
+      # will happen anyway when the values are assigned to the scope after the parameter scoped has
+      # been popped.
+      resource.each { |k, v| param_scope[k.to_s] = v.value unless k == :name || k == :title }
+      assign_defaults(resource, param_scope, scope)
+      param_scope.to_hash
+    end
+
+    validate_resource_hash(resource, param_hash)
+
+    # Assign parameter values to current scope
+    param_hash.each { |param, value| exceptwrap { scope[param] = value }}
   end
 
   # Lookup and inject parameters from external scope
@@ -395,19 +401,17 @@ class Puppet::Resource::Type
   end
   private :inject_external_parameters
 
-  def assign_defaults(resource, scope, resource_hash)
+  def assign_defaults(resource, param_scope, scope)
     return unless resource.is_a?(Puppet::Parser::Resource)
     parameters = resource.parameters
-    hashed_types = parameter_struct.hashed_elements
     arguments.each do |param_name, default|
       next if default.nil?
       name = param_name.to_sym
       param = parameters[name]
       next unless param.nil? || param.value.nil?
-
-      value = default.safeevaluate(scope)
+      value = exceptwrap { param_scope.evaluate3x(param_name, default, scope) }
       resource[name] = value
-      resource_hash[param_name] = value
+      param_scope[param_name] = value
     end
   end
   private :assign_defaults
--- a/spec/integration/parser/parameter_defaults_spec.rb
+++ b/spec/integration/parser/parameter_defaults_spec.rb
@@ -123,7 +123,7 @@ require 'puppet_spec/language'
         expect_fail(params, [], /default expression for \$b tries to illegally access not yet evaluated \$a/)
       end
 
-      it 'will use the referenced parameters given value' do
+      it "will use the referenced parameter's given value" do
         expect_log(params, [2], ['$a == 2', '$b == 2'])
       end
 
@@ -139,11 +139,11 @@ require 'puppet_spec/language'
       SOURCE
       }
 
-      it 'will use the referenced parameters default value when no value is given for the referenced parameter' do
+      it "will use the referenced parameter's default value when no value is given for the referenced parameter" do
         expect_log(params, [], ['$a == 10', '$b == 10'])
       end
 
-      it 'will use the referenced parameters given value' do
+      it "will use the referenced parameter's given value" do
         expect_log(params, [2], ['$a == 2', '$b == 2'])
       end
 
@@ -174,7 +174,7 @@ require 'puppet_spec/language'
     end
 
     context 'with regular expressions' do
-      it 'evaluates unset match scope parameters to undef' do
+      it "evaluates unset match scope parameter's to undef" do
         expect_log([<<-SOURCE, 2], [], ['$a == ', '$b == '])
         $a = $0,
         $b = $1
@@ -196,7 +196,7 @@ require 'puppet_spec/language'
         SOURCE
       end
 
-      it 'can have nexted match expressions' do
+      it 'can have nested match expressions' do
         expect_log([<<-SOURCE, 2], [], ['$a == [true, h, oo, h, i]', '$b == '] )
         $a = ['hi' =~ /(h)(.*)/, $1, if'foo' =~ /f(oo)/ { $1 }, $1, $2],
         $b = $0
@@ -266,7 +266,7 @@ require 'puppet_spec/language'
         SOURCE
       end
 
-      it 'will not make match scope available to function body' do
+      it "will not make match scope available to #{call_type} body" do
         expect_log([<<-SOURCE, call_type == 'Function' ? <<-BODY : <<-EPP_BODY], [], ['Yes'])
         $a = "hello" =~ /.*/
         SOURCE
--- a/spec/integration/parser/parameter_defaults_spec.rb
+++ b/spec/integration/parser/parameter_defaults_spec.rb
@@ -1,0 +1,336 @@
+require 'spec_helper'
+require 'puppet_spec/language'
+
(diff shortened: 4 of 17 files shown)


lib/puppet/parser/scope.rb
```diff
@@ -174,6 +173,84 @@ def add_entries_to(target = {})
 
   end
 
+  class ParameterScope < Ephemeral
```
hlindberg: Should be documented as @api private

lib/puppet/parser/scope.rb
```diff
@@ -174,6 +173,84 @@ def add_entries_to(target = {})
...
+    # @return [Object] the result of the evaluation
+    #
+    # @api private
+    def self.evaluate3x(name, expression, scope)
+      elevel = scope.ephemeral_level
+      bad = catch(:unevaluated_parameter) do
+        scope.new_match_scope(nil)
+        return expression.safeevaluate(scope)
+      end
+      raise Puppet::Error, "$#{name} tries to access unevaluated $#{bad}"
```
hlindberg: hm, it is not the variable that tries to access. How about "default expression for ${name} tries to illegally access not yet evaluated $#{bad}"

lib/puppet/parser/scope.rb
```diff
@@ -174,6 +173,84 @@ def add_entries_to(target = {})
...
+    end
+
+    def self.evaluate(name, expression, scope, evaluator)
+      elevel = scope.ephemeral_level
+      bad = catch(:unevaluated_parameter) do
+        scope.new_match_scope(nil)
+        return evaluator.evaluate(expression, scope)
+      end
+      raise Puppet::Error, "$#{name} tries to access unevaluated $#{bad}"
+    ensure
```
hlindberg: This sometimes does not work - a block around the body of the method is recommended. (We have been bitten in the past by ensures and catch in the implicit block).

lib/puppet/parser/scope.rb
```diff
@@ -174,6 +173,84 @@ def add_entries_to(target = {})
...
+
+    def [](name)
+      access = @params[name]
+      return super if access.nil?
+      throw(:unevaluated_parameter, name) unless access.assigned?
+      access.value
+    end
+
+    def []=(name, value)
+      @params[name] ||= Access.new
```
hlindberg: what if user writes something confused like some variation of this:
```
function foo($x = ($y = 2), $y = 5) { }
```
It does not introduce new variables, but assigns to parameters. I am asking because if this method is called and it is not a parameter I think it should error.

Remember that if default expression calls functions, the scope will potentially be accessed by 3d party ruby code (doing odd things they are not allowed to).

lib/puppet/parser/scope.rb
```diff
@@ -738,17 +815,27 @@ def to_s
...
       @ephemeral = [ MatchScope.new(@symtable, nil)]
+      result
     else
       @ephemeral.pop(@ephemeral.size - level)
     end
   end
 
+  # Push ephemeral scopes onto the ephemeral scope stack
+  # @param ephemeral_scopes [Array]
+  def restore_ephemerals(ephemeral_scopes)
```
hlindberg: @api private (unless you see a valid reason why other than PL would use this)

lib/puppet/parser/scope.rb
```diff
@@ -773,6 +860,13 @@ def new_match_scope(match_data)
     @ephemeral.push(MatchScope.new(@ephemeral.last, match_data))
   end
 
+  # Nests a parameter scope
+  def new_parameter_scope(param_names)
```
hlindberg: @api private

lib/puppet/parser/scope.rb
```diff
@@ -738,17 +815,27 @@ def to_s
 
   alias_method :inspect, :to_s
 
-  # remove ephemeral scope up to level
-  # TODO: Who uses :all ? Remove ??
```
hlindberg: There is not a single use case of unset_ephemeral_var(:all) and if someone called that it is a sure way of creating havoc. Suggest not changing the implementation to correctly return the popped ephemeral scopes. I would much rather see a deprecation warning.

lib/puppet/parser/scope.rb
```diff
@@ -738,17 +815,27 @@ def to_s
...
+  # Pop ephemeral scopes up to level and return them
   #
+  # @param level [Symbol,Fixnum] The symbol `:all` or a positive integer
+  # @return [Array] the removed ephemeral scopes
+  # @api private
   def unset_ephemeral_var(level=:all)
     if level == :all
+      result = @ephemeral
       @ephemeral = [ MatchScope.new(@symtable, nil)]
+      result
```
hlindberg: as per the comment above - don't change this - if you want to improve this, add a deprecation warning; there should be no calls to unset_ephemeral_var(:all) and if someone does that they should be told. The proposed change would just break them (or make them screw up as it changes the broken API. (And yes, this is horrible).

lib/puppet/pops/evaluator/closure.rb
```diff
@@ -25,9 +28,13 @@ def initialize(evaluator, model, scope)
   # Evaluates a closure in its enclosing scope after having matched given arguments with parameters (from left to right)
   # @api public
   def call(*args)
+    # We only want to see the global scope here. Other ephemeral scopes must be popped
```
hlindberg: "We" - who is we in this context. Please rephrase - e.g. "only global scope should be visible here"...

lib/puppet/pops/evaluator/closure.rb
```diff
@@ -53,18 +65,38 @@ def invoke(instance, calling_scope, args, &block)
 
   # Call closure with argument assignment by name
   def call_by_name(args_hash, enforce_parameters)
+    # We only want to see the global scope here. Other ephemeral scopes must be popped
```
hlindberg: same comment as above

lib/puppet/pops/evaluator/closure.rb
```diff
@@ -53,18 +65,38 @@ def invoke(instance, calling_scope, args, &block)
...
+        args_hash = param_scope.to_hash
+      ensure
+        @enclosing_scope.unset_ephemeral_var(elevel)
       end
-      Puppet::Pops::Types::TypeMismatchDescriber.validate_parameters(closure_name, params_struct, args_hash)
+      Types::TypeMismatchDescriber.validate_parameters(closure_name, params_struct, args_hash)
     end
 
     @evaluator.evaluate_block_with_bindings(@enclosing_scope, args_hash, @model.body)
+  ensure
```
hlindberg: ensure in a body of a method - should be in a begin

lib/puppet/pops/evaluator/closure.rb
```diff
@@ -167,10 +203,12 @@ def combine_values_with_parameters(args)
         end
       end
 
-      variable_bindings[parameter.name] = value
+      param_scope[parameter.name] = value
     end
 
-    variable_bindings
+    param_scope.to_hash
+  ensure
```
hlindberg: ensure in body without matching begin

hlindberg:
This PR should have been broken up into several commits. There is a lot going on here with refactoring of names (using relative names instead of absolute), cleanups, added validation etc. which makes it a lot slower to review.

lib/puppet/resource/type.rb
```diff
@@ -360,9 +360,6 @@ def set_resource_parameters(resource, scope)
     resource.add_parameters_from_consume
     inject_external_parameters(resource, scope)
 
-    resource_hash = {}
```
hlindberg: Since this is changes in resource/types.rb (that also is used client side), it requires extra caution. Now, this happens to be in a method that only makes sense on the compiling side. I think we should update the method documentation to state that (makes the next person working on this code less anxious about making changes). (Github did not allow me to add a note on the actual comment for this method, so I inserted the comment here.

spec/integration/parser/epp_parameters_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+end
+
+describe "EPP parameter default expressions" do
+  extend EppRunner
+
+  context "to the left of a parameter can use that parameter's value" do
+    run_expectations('<%| $a = 10, $b = $a |%><%= notice($a) %><%= notice($b) %>',
+      [ [{}, ['10', '10']],
+        [{'a' => 2}, ['2', '2']],
+        [{'a' => 2, 'b' => 5}, ['2', '5'] ]])
```
hlindberg: This would be greatly enhanced if there was a string in each expectation that explained that example. Here the first would be "$b is assigned default value of $a", the second "$b is assigned the given value of $a", etc. This helps both reading the examples (now it took a while to figure out what the example meant. The method run_expectations should then use the given string in the "it" - to make it easier to find which one of the expectations that failed (since the failure is reported by rspec against the line where the "it" is. (I had the pleasure of fixing > 100 such failures on an API change in another rspec test using something similar.

spec/integration/parser/epp_parameters_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  end
+
+  context 'will not permit assignments' do
+    run_expectations('<%| $a = $x = $0 |%><%= notice($a) %>',
+      [ [{}, [ArgumentError, /Syntax error at '='/]] ])
+
+    run_expectations('<%| $a = [$x = 3] |%><%= notice($a) %>',
+      [ [{}, [ArgumentError, /Assignment not allowed here/]] ])
+  end
+
```
hlindberg: need test of case ($a = ($b = 1), $b = 2)  - to ensure that assignment to $b is not allowed in default expr of $a, but allowed as a parameter assignment when evaluating $b's default expr.

Also need something like this:

´´´
$a = [1,2,3], $b = 0, $c = $a.map |$x| { $b = $x /* ok, $b is var in local scope */; $b * $a.reduce |$x, $y| {$x + $y}

$a => [1,2,3]
$b => 0
$c = [6, 12, 18]
´´´

spec/integration/parser/epp_parameters_spec.rb
hlindberg: It will be easier to review completeness of the examples when there are strings that explains each.

spec/integration/parser/function_parameters_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+        'example(2)' => ['$a == 2', '$b == 2'],
+        'example(2,5)' => ['$a == 2', '$b == 5'] })
+
+    run_expectations('function example($a, $b = $a) { notice("\$a == ${a}") notice("\$b == ${b}") }',
+      { 'example(10)' => ['$a == 10', '$b == 10'],
+        'example(10,20)' => ['$a == 10', '$b == 20'] })
+  end
+
+  context 'that references a variable to the right' do
+    run_expectations('function example($a = 10, $b = $c, $c = 20) { notice("\$a == ${a}") notice("\$b == ${b}") notice("\$c == ${c}") }',
```
hlindberg: need to break the long lines, use a heredoc to make it easier to read the puppet source ?

spec/integration/parser/function_parameters_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    run_expectations('function example($a = $0, $b = $1) { notice("\$a == ${a}") notice("\$b == ${b}") }',
+      { 'example()' => ['$a == ', '$b == '] })
+
+    run_expectations("function example($a = ['hello' =~ /(h)(.*)/, $1, $2], $b = $1) { notice(\"\\$a == ${a}\") notice(\"\\$b == ${b}\") }",
+      { 'example()' => ['$a == [true, h, ello]', '$b == '] })
+
+    run_expectations("function example($a = ['hello' =~ /(h)(.*)/, $1, $2], $b = ['hi' =~ /(h)(.*)/, $1, $2], $c = $1) { notice(\"\\$a == ${a}\") notice(\"\\$b == ${b}\")  notice(\"\\$c == ${c}\")}",
+      { 'example()' => ['$a == [true, h, ello]', '$b == [true, h, i]', '$c == '] })
+
+    run_expectations("function example($a = ['hi' =~ /(h)(.*)/, $1, if'foo' =~ /f(oo)/ { $1 }, $1, $2], $b = $0) { notice(\"\\$a == ${a}\") notice(\"\\$b == ${b}\")}",
```
hlindberg: heredoc would help here - this is hard to read, github seems to not even show the entire line on the example above
```
run_expectations(<<-SOURCE)
  function example( 
    $a = ['hi' =~ /(h)(.*)/, $1, if'foo' =~ /f(oo)/ { $1 }, $1, $2],
    $b = $0
  ) {
    notice("\$a == ${a}")
    notice("\$b == ${b}")
  }
SOURCE
´´´

BTW, if all of the examples have the same function body (notice each arg), that could be interpolated via a method
say 
```
<<-"SOURCE"
function example(.....) #{body_reporting(:a, :b, :c)}
SOURCE
```
as that would relieve the reader from having to figure out if one example is different from another and having to carefully see if there is a difference in escapes etc.

File: spec/unit/functions/epp_spec.rb [hunk 18]
```diff
@@ -27,9 +27,14 @@
...
     end
 
-    it "can use values from the enclosing scope for defaults" do
+    it "can use values from the global scope for defaults" do
       scope['phantom'] = 'of the opera'
-      expect(eval_template("<%- |$phantom = $phantom| -%><%= $phantom %>")).to eq("of the opera")
+      expect(eval_template("<%- |$phantom = $::phantom| -%><%= $phantom %>")).to eq("of the opera")
+    end
+
+    it "can use values from the enclosing scope for defaults" do
```
hlindberg: Well, it lies, it tests that it can look up using a relative variable using an enclosing scope that also happens to be the global scope. seems a nested scope that is enclosing should be used and that there is a different value in the global scope that it does not select.

spec/unit/pops/evaluator/evaluating_parser_spec.rb
```diff
@@ -861,7 +861,8 @@
         # NOTE: these meta-esque parameters are not recognized as such
         "notify { id: message=>explicit} Notify[id][title]"   => /does not have a parameter called 'title'/,
         "notify { id: message=>explicit} Notify[id]['type']"   => /does not have a parameter called 'type'/,
-        "notify { id: message=>explicit } Notify[id]{message=>override}" => /'message' is already set on Notify\[id\]/
+        "notify { id: message=>explicit } Notify[id]{message=>override}" => /'message' is already set on Notify\[id\]/,
+        "notify { id: message => 'once', message => 'twice' }" => /'message' has already been set in this resource body/
```
hlindberg: I think the reference to "body" here is not needed (I don't think users understand what a Resource Body is). If you feel it is important to make a distinction between the resource instance, and the PP language construct you could use the term "resource statement", or "resource declaration" (the last is the term used in the puppet documentation).

spec/unit/pops/validator/validator_spec.rb
```diff
@@ -229,6 +229,20 @@ def validate(model)
...
+      end
+    end
+
+    ['class', 'define', 'function'].each do |word|
+      it "should not permit assignments in #{word} parameter default nested expressions" do
+        expect(validate(parse("#{word} foo($a = [$x = 10]) {}"))).to have_issue(Puppet::Pops::Issues::ILLEGAL_ASSIGNMENT_CONTEXT)
+      end
+    end
+  end
+
```
hlindberg: what if it is an assignment to a parameter, and to itself. $a = $a, and $a = 10, $b = ($a = 10). Also, if parameter default expression contains a lambda, assignments in the lambdas local scope should be allowed (so, test is needed for that).

spec/unit/pops/validator/validator_spec.rb
```diff
@@ -313,20 +327,14 @@ def validate(model)
...
-      # We simulate a file by inserting a block between the program and the contained function
-      program = parse(source).current
-      function = program.body
-      program.body = Puppet::Pops::Model::BlockExpression.new
-      program.body.addStatements(function)
-      expect(validate(program)).not_to have_issue(Puppet::Pops::Issues::NOT_ABSOLUTE_TOP_LEVEL)
+    it 'does not produce an error when function is in a top level block' do
+      source = "function x() {}\nfunction y() {}"
+      expect(validate(parse(source))).not_to have_issue(Puppet::Pops::Issues::NOT_ABSOLUTE_TOP_LEVEL)
     end
```
hlindberg: since some examples when it does not error where added, for completeness sake maybe also add one that checks that functions and other expressions can be mixed:

$a = 10
function foo() { }

etc

spec/unit/resource/type_spec.rb
```diff
@@ -272,6 +289,128 @@ def variable_expression(name)
...
+    context 'referencing a variable to the right of the default expression' do
+      before :each do
+        @type.set_arguments({
+          :first => number_expression(10),
+          :second => variable_expression('third'),
+          :third => number_expression(20)
+        })
+      end
+
+      it 'is ok when no defaults are used' do
```
hlindberg: "is ok" is too vague

spec/unit/resource/type_spec.rb
```diff
@@ -272,6 +289,128 @@ def variable_expression(name)
...
+        @type.set_resource_parameters(@resource, @scope)
+
+        expect(@scope['first']).to eq(1)
+        expect(@scope['second']).to eq(2)
+        expect(@scope['third']).to eq(20)
+      end
+
+      it 'fails when the default expression is evaluated' do
+        @resource[:first] = 1
+        expect { @type.set_resource_parameters(@resource, @scope) }.to raise_error(Puppet::Error, '$second tries to access unevaluated $third')
```
hlindberg: note I wanted the error message changed earlier

spec/unit/resource/type_spec.rb
```diff
@@ -272,6 +289,128 @@ def variable_expression(name)
...
+          :first => matchref_expression('0'),
+          :second => matchref_expression('1'),
+        })
+        @type.set_resource_parameters(@resource, @scope)
+
+        expect(@scope).not_to include('first')
+        expect(@scope).not_to include('second')
+      end
+
+      it 'match scope exists for a parameter, but does not affect match scopes to the right' do
```
hlindberg: hard to read and understand this example title - suggesting "a match scope to the left of a parameter is not visible to it"

spec/unit/resource/type_spec.rb
```diff
@@ -272,6 +289,128 @@ def variable_expression(name)
...
+          :first => parse_expression("['hello' =~ /(h)(.*)/, $1, $2]"),
+          :second => matchref_expression('1'),
+        })
+        @type.set_resource_parameters(@resource, @scope)
+
+        expect(@scope['first']).to eq([true, 'h', 'ello'])
+        expect(@scope['second']).to be_nil
+      end
+
+      it 'match scope nests per parameter' do
```
hlindberg: scopes (plural)

lib/puppet/parser/scope.rb
thallgren: Exactly what is it that doesn't work? A specific construct? I don't find any reports on this. I'm interested because I don't think this is the only place in the code where this exists.

spec/unit/pops/evaluator/evaluating_parser_spec.rb
thallgren: I will change the Issue text. It wasn't added by this PR though.

lib/puppet/resource/type.rb
thallgren: Not sure what you are suggesting here. Should we update all method comments to state whether or not they are executed on the client side? If so, then someone else needs to do that (perhaps in scope for another JIRA). I don't know how to figure that out.

spec/integration/parser/epp_parameters_spec.rb
thallgren: I contemplated doing that but rejected it since the expression is very self explanatory in the context where it evaluates.

spec/unit/pops/validator/validator_spec.rb
thallgren: All of these works as expected. I'll add the tests.

spec/integration/parser/epp_parameters_spec.rb
thallgren: When adding these tests I found that the last test fails. Need to investigate that.

spec/unit/functions/epp_spec.rb
thallgren: If this should work the same way as a function invocation, then it shouldn't use the enclosing scope at all. It should always use the global scope. But given the nature of templates (inlined in code) I guess this is one place where a function parameter declaration and a template parameter declaration may differ?

thallgren:
Pushed a new commit where I think all issues have been addressed. I removed things related to refactoring/cleanup. Not really related to this PR anyway.

lib/puppet/resource/type.rb
hlindberg: I was suggesting adding opportunistically on each method as we learn that it is client side only (like in this case, this particular method).

lib/puppet/parser/scope.rb
```diff
@@ -738,17 +821,27 @@ def to_s
...
       @ephemeral = [ MatchScope.new(@symtable, nil)]
     else
       @ephemeral.pop(@ephemeral.size - level)
     end
   end
 
+  # Push ephemeral scopes onto the ephemeral scope stack
+  # @param ephemeral_scopes [Array]
+  # @api private
+  def restore_ephemerals(ephemeral_scopes)
```
hlindberg: The name of the method does not rhyme with the documentation and implementation - this does not look like some kind of restore operation, it just adds ephemerals.
How about calling this `push_ephemerals`, and create an alias `pop_ephemerals` for the weird `unset_ephemeral_var` (we keep that method intact for backwards compatibility reasons).

lib/puppet/pops/evaluator/closure.rb
```diff
@@ -48,18 +48,39 @@ def call(*args)
   # checks of the argument type/arity validity).
   # @api private
   def invoke(instance, calling_scope, args, &block)
-    call(*args, &block)
+    scopes_to_restore = @enclosing_scope.unset_ephemeral_var(1)
```
hlindberg: I think this makes this valid example fail:
```
$a = case "hello" {
  /(h)(.*)/ : {  
    [1,2,3].map |$x| { "$x-$2" }
  }
}
```
Result should be:
```
$a = ["1-ello", "2-ello", "e-ello"]
```

lib/puppet/pops/evaluator/closure.rb
```diff
@@ -48,18 +48,39 @@ def call(*args)
...
+    begin
+      # Only global scope should be visible to the called function. Other ephemeral scopes must be popped
+      # and later restored
+      call(*args, &block)
+    ensure
+      unless scopes_to_restore.nil?
+        @enclosing_scope.unset_ephemeral_var(1)
```
hlindberg: this "1" also looks suspect (potentially an issue - as described above).

hlindberg: The call_by_name does what I expect, it takes the level and restores it. Thats it.
Not sure what the restore stuff is doing. Remember that the match scopes are opaque - if you try to lookup say $9 and the closest matchscope has an empty match then the search stops there. Thus, it is just a matter of assigning a match to the current matchscope, or pushing one (empty or with a match), then, at the end pop back to the level that was remembered on entry. 

But maybe I am reading the code wrong and I am being confused of what invoke does.

lib/puppet/resource/type.rb
```diff
@@ -370,12 +367,27 @@ def set_resource_parameters(resource, scope)
...
-    assign_defaults(resource, scope, resource_hash)
-    validate_resource_hash(resource, resource_hash)
-    resource_hash.each { |param, value| exceptwrap { scope[param] = value }}
+    begin
+      elevel = scope.ephemeral_level
+      param_scope = scope.new_parameter_scope(arguments.keys)
+
+      # Assign directly to the parameter scope to avoid scope parameter validation at this point. It
+      # will happen anyway when the values are assigned to the scope after the parameter scoped has
+      # been popped.
```
hlindberg: which makes me want to add this test case:
```
function example($a = "hello" =~ /.*/) {
  notify { test: message "Y$0es" }
}
```
should have notify "test" with message "Yes"

spec/integration/parser/epp_parameters_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+      end
+    end
+  end
+end
+
+describe "EPP parameter default expressions" do
+  extend EppRunner
+
+  context "to the left of a parameter can use that parameter's value" do
+    run_expectations(<<-SOURCE.gsub(/^ {6}/, ''),
```
hlindberg: The gsub trick is very brittle as it depends on the indentation used. If the code is indented more, it does no longer do the right thing. I don't think whitespace matters in these cases, so the gsub can simply be removed.

thallgren: It was added for a reason. The epp evaluator does not accept leading spaces.

spec/integration/parser/epp_parameters_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+        end
+      end
+    end
+  end
+end
+
+describe "EPP parameter default expressions" do
+  extend EppRunner
+
+  context "to the left of a parameter can use that parameter's value" do
```
hlindberg: This is unreadable - (I commented on this in the previous version, but it was kept as is).

lib/puppet/resource/type.rb
thallgren: Two comments about that.
How did we learn that this is client side only?
Isn't there a fair chance that if we add this to some methods, we might create a false impression that it's added to all methods where it makes sense?

I'm not at all against making the distinction. I just think it in order to help it must be done throughout. Adding it to some methods might actually add to the confusion.

spec/integration/parser/epp_parameters_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+      <%|
+        $a = 10,
+        $b = $a
+      |%>
+      <%
+        notice($a)
+        notice($b)
+      %>
+      SOURCE
+      [ [{}, ['10', '10']],
```
hlindberg: I would like this to read:
```
[ [ "when no args are given, param lookup finds default value",  {}, ['10', '10']],
  ["when one arg is given, param lookup finds given value", {'a' => 2}, ['2', '2']],
  ["when two args are given, given value is used for both", {'a' => 2, 'b' => 5}, ['2', '5'] ]])
```

So that those strings can be included in the "it" part in the generic method

thallgren: And you don't think that's abundantly clearly stated in the next argument?

lib/puppet/parser/scope.rb
hlindberg: I am still unsure about this. Can a ParameterScope leak to 3d party ruby user code? If so, will it create a problem if they add variable? Say of a parameter is assigned to the value of a function call
```
function example($a = evil()) {
}
```
As the evil function can get calling scope, it would be able to add a parameter by using the []? operator (at least it seams so).

lib/puppet/resource/type.rb
hlindberg: You could say:
```
This method is known to only be used on the server/compile side.
```
I don't think that can be read to imply anything about any other method.

spec/integration/parser/epp_parameters_spec.rb
hlindberg: ah, yes, true. Calling a method that strips leading whitespace would be better (due to a possible later indentation change problem).

lib/puppet/parser/scope.rb
thallgren: The adding of an Access element here is intentional. If the variable doesn't represent a valid parameter then an error will be raised later on when the parameter Struct is used to validate the result of the ParameterScope#to_hash. The alternative would be to raise that error right here but that would violate the logic that compresses the error reporting.

spec/integration/parser/epp_parameters_spec.rb
hlindberg: no, and when it fails, the rspec output makes it more difficult to find what failed, and why (you do not get the line of the actual example - instead you get the line of the "it" in run expectations, then you have to try to find something like "{}", then read up on how the pieces fit together . I had to spend a fair amount of time to understand what these tests did on the first read, and I would not have requested this to be changed unless I thought it was worth the time changing it.

spec/integration/parser/epp_parameters_spec.rb
thallgren: Sure, the gsub can be moved. Then again, then it looses the context (the code is indeed indented with 6 spaces right here, the receiver of the string doesn't know that). Just stripping leading white-space will not cut it since there's might be more than one EPP expression.

spec/integration/parser/epp_parameters_spec.rb
thallgren: Here's a sample output from forced failure:
```
  1) EPP parameter default expressions to the left of a parameter can use that parameter's value '<%|
  $a,
  $b = $a
|%>
<%
  notice($a)
  notice($b)
%>
' called with {"a"=>10} should produce ["10", "10"]
```
In what way would the string "when one arg is given, param lookup finds given value" bring more clarity? I understand that the first version was hard to read, unformatted as it was. But now?

lib/puppet/pops/evaluator/closure.rb
thallgren: The above sample does not fail but I will add the tests.

lib/puppet/pops/evaluator/closure.rb
thallgren: I mentioned this in the scrum. The call_by_name (used when assigning EPP templates) has access to the enclosing scope (as you explained, this is wrong, only inline_epp should have that access). That's why the call_by_name just takes the level and restores it once all parameter defaults have been evaluated.

The invoke is different. It is what differentiates a function call from a lambda call. A function call must only see the global scope. It should not see any ephemeral scopes that are pushed onto it. So all such scopes are temporarily popped during the call and then later restored once the call returns.

lib/puppet/resource/type.rb
thallgren: Actually, it produces an error 'Illegal numeric variable name' but using `"Y${0}es"` instead of `"Y$0es"` works OK though. I'll add that test.

thallgren:
Pushed a new rebased commit where new tests mentioned here were added. I also refactored the Function and EPP parameter default tests and merged into a 'parameter_defaults_spec'. My hope is that their purpose are now clearer with more concise context info.

lib/puppet/parser/scope.rb
hlindberg: So, a) if someone did add a variable to the scope say `scope[foo] = 42`, and there is no foo parameter defined in the function  - would it then error later? and b) same case, but there is a parameter `foo` defined, but it is to the right of the function call to `evil()`. Will it error, and when?

thallgren: a) yes. And this is essential since it happens when such arguments are passed in a call using named parameters (as with resource or class definitions and epp templates).
b) It will not error. Assignments are not permitted in the PL.

I'm not sure what you are after here. If we pass the scope to a function written in Ruby and that function has malicious intent, then there's no end to the harm that this function can do. If it want's to mimic the assignment of a parameter to the right of where it's executing, then who's to stop that? The parameter scope must allow such assignments in order for the proper default evaluation to succeed.

hlindberg: Not even by getting calling_scope in a ruby function? (I am going on about this, because there are those that write ruby functions that adds variables in the calling scope - a very smelly design, but not illegal. Also, while we do not have this yet, I can imagine advanced core puppet functions (e.g. macros) that need this ability).

May not be worth the effort to prevent users from doing these advanced mistakes. It will not occur by accident for sure.

lib/puppet/parser/scope.rb
```diff
@@ -738,10 +815,12 @@ def to_s
...
   alias_method :inspect, :to_s
 
-  # remove ephemeral scope up to level
-  # TODO: Who uses :all ? Remove ??
+  # Pop ephemeral scopes up to level and return them
   #
+  # @deprecated use #pop_epehemeral
+  # @api private
   def unset_ephemeral_var(level=:all)
+    Puppet.deprecation_warning('Method Parser::Scope#unset_ephemeral_var() is deprecated')
```
hlindberg: It is only unset_ephemeral_var(:all) that is deprecated. This line needs to go inside the if below. And it should mention :alll in the error. (We do not want people to have to make changes to use pop_ephemerals, and later have to change again when we change how scope works). It is simply the :all case that we know will no longer be supported (and is causing potential harm if used).

spec/integration/parser/parameter_defaults_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+
+      it 'fails when no value is provided for required first parameter', :if => call_type == 'Function' do
+        expect_fail(params, [], /expects between 1 and 2 arguments, got none/)
+      end
+
+      it 'fails when no value is provided for required first parameter', :if => call_type == 'EPP' do
+        expect_fail(params, [], /default expression for \$b tries to illegally access not yet evaluated \$a/)
+      end
+
+      it 'will use the referenced parameters given value' do
```
hlindberg: parameter's

spec/integration/parser/parameter_defaults_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+
+      it 'can evaluate expressions in separate match scopes' do
+        expect_log([<<-SOURCE, 3], [], ['$a == [true, h, ell, o]', '$b == [true, h, i, ]', '$c == '])
+        $a = ['hello' =~ /(h)(.*)(o)/, $1, $2, $3],
+        $b = ['hi' =~ /(h)(.*)/, $1, $2, $3],
+        $c = $1
+        SOURCE
+      end
+
+      it 'can have nexted match expressions' do
```
hlindberg: typo: nexted

lib/puppet/parser/scope.rb
thallgren: A ruby function does get the full scope which means that, yes, it can change it. I can see two possible ways of preventing that from happening:

1. We do for Ruby function what we do for a Puppet function, just send the global scope.
2. Let the ParameterScope be read-only during the evaluation.

I guess the first alternative, although from a puristic standpoint very attractive, would create backward compatibility issues. The second on the other hand, doesn't prevent a malicious Ruby function from overriding the read-only status by doing things like `scope.instancevariable_get(:@params)['foo'] = bar` or replacing ephemeral scopes. It will however prevent a ruby function author from doing silly mistakes.

spec/integration/parser/parameter_defaults_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+          <% called_from_template($a, $b) %>
+          BODY
+          function called_from_template($ax, $bx) {
+            notice("\\$ax == $ax")
+            notice("\\$bx == $bx")
+          }
+          $tmp = 'foo' =~ /(f)(o)(o)/
+          CODE
+        end
+
```
hlindberg: Additional test is needed: since match variables are available in the parameter scope, a test is needed to ensure that they are not available in the epp body (this because inline_epp that specifies parameters hides the calling scopes variables from the body (but they are available in the parameter scope).

spec/integration/parser/parameter_defaults_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+        expect_log([<<-SOURCE, 1], [], ['$a == [1-ello, 2-ello, 3-ello]'])
+        $a = case "hello" {
+          /(h)(.*)/ : {
+            [1,2,3].map |$x| { "$x-$2" }
+          }
+        }
+        SOURCE
+      end
+
+      it 'will not make match scope available to function body' do
```
hlindberg: should probably say "epp body"

spec/unit/pops/validator/validator_spec.rb
```diff
@@ -229,6 +229,34 @@ def validate(model)
...
+
+      it "should not permit assignments to subsequently declared parameters in #{word} parameter default nested expressions" do
+        expect(validate(parse("#{word} foo($a = ($b = 3), $b = 5) {}"))).to have_issue(Puppet::Pops::Issues::ILLEGAL_ASSIGNMENT_CONTEXT)
+      end
+
+      it "should not permit assignments to previously declared parameters in #{word} parameter default nested expressions" do
+        expect(validate(parse("#{word} foo($a = 10, $b = ($a = 10)) {}"))).to have_issue(Puppet::Pops::Issues::ILLEGAL_ASSIGNMENT_CONTEXT)
+      end
+
+      it "should permit assignments permit assignments in #{word} parameter default inside nested lambda expressions" do
```
hlindberg: typo "permit assignments permit assignments"

lib/puppet/parser/scope.rb
thallgren: I pushed a new commit with a fix using approach #2.

thallgren:
Just realized that my last push made the comments about ruby function doing bad things to the parameter scope outdated. The last commit ensures that the ParameterScope is read-only during all evaluation.

lib/puppet/parser/scope.rb
hlindberg: I was thinking that the parameter scope would only allow assignments to the defined parameters. Thus, if bad ruby code sets it before it set the expected way (assigning the value of the parameter), it would fail with "already set", the same thing, if ruby code tries to set it after the parameter as expected has been given its value. This would require that the parameter scope gets to know all parameters by name when it is created.

Since this is ruby, a malicious user can do anything - we are only trying to prevent accidental bad things from happening.

lib/puppet/parser/scope.rb
thallgren: I decided to deprecate that method as a whole since you also wanted me to add an alias `pop_ephemerals`for it. All calls, except from the ones using :all (only from tests) have been replaced to use `pop_ephemerals` so I think it's safe to say that this method (@api private as it is) is indeed deprecated.

hlindberg:
Can you explain. (I added a comment to the thread on this topic above). What makes it "read only", the fact that it is not seen by "evil" code i.e the ´$a=evil()` case, or something else?

spec/integration/parser/parameter_defaults_spec.rb
thallgren: No, it should say "#{call_type} body" since it's applicable to both function and epp.

spec/integration/parser/parameter_defaults_spec.rb
thallgren: There is a test that covers that (the one that said 'function body' but covers both) at line 269.

thallgren:
A new boolean @read_only variable was introduced. Assignments are not permitted if it's been set.

The methods #evaluate and #evalutate3x (called for all parameter default evaluation) now do:
```
return as_read_only { evaluator.evaluate(expression, scope) }
```
and
```
return as_read_only { expression.safeevaluate(scope) }
```
The method as_read_only is defined on ParameterScope as:
```
    def as_read_only
      read_only = @read_only
      @read_only = true
      begin
        yield
      ensure
        @read_only = read_only
      end
    end
```

lib/puppet/parser/scope.rb
hlindberg: ok fair, it was @api private to begin with.

spec/integration/parser/parameter_defaults_spec.rb
hlindberg: right

lib/puppet/parser/scope.rb
```diff
@@ -174,6 +174,94 @@ def add_entries_to(target = {})
...
+
+    def [](name)
+      access = @params[name]
+      return super if access.nil?
+      throw(:unevaluated_parameter, name) unless access.assigned?
+      access.value
+    end
+
+    def []=(name, value)
+      raise Puppet::Error, "Attempt to assign variable #{name} to read-only parameter scope" if @read_only
```
hlindberg: "Attempt to assign variable ${name} when evaluating parameters" (users do not know what a "read-only parameter scope" is).

hlindberg:
All good now. Waiting on Travis to go green before merging.

thallgren:
Ruby 2.0.0 timed out on Travis. All other succeed. Not related to my last commit.

hlindberg:
Paint finally dried....

PR #4569, opened 2016-01-13
(PUP-1607) Acceptance - no master in symbolic_modes

johnduarte:
This commit confines hosts eligible to be included in the execution
of the `resource/file/symbolic_modes` test.

It removes the master from the hosts list.
It also moves the logic to exclude windows and solaris-10 from
the body of the test and places them as confine statements at
the beginning.

### DIFF
--- a/acceptance/tests/resource/file/symbolic_modes.rb
+++ b/acceptance/tests/resource/file/symbolic_modes.rb
@@ -1,5 +1,8 @@
 test_name "file resource: symbolic modes"
 confine :except, :platform => /^eos-/ # See ARISTA-37
+confine :except, :platform => /^solaris-10/
+confine :except, :platform => /^windows/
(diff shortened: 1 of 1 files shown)


acceptance/tests/resource/file/symbolic_modes.rb
```diff
@@ -1,5 +1,8 @@
 test_name "file resource: symbolic modes"
 confine :except, :platform => /^eos-/ # See ARISTA-37
+confine :except, :platform => /^solaris-10/
+confine :except, :platform => /^windows/
+confine :to, {}, hosts.select { |host| ! host[:roles].include?('master') }
```
joshcooper: Nicely done @johnduarte! I wonder if other tests run all hosts (because why not?) when really they should only run on agent hosts?

PR #4571, opened 2016-01-14
(RE-6205) updating ruby for CVE-2015-7551

McdonaldSeanp:
This will update ruby 2.0.0 and 1.9.3 for windows
per CVE at: https://tickets.puppetlabs.com/browse/RE-6141


underscorgan:
Pending appveyor and travis :100: 

These MSIs have been built and tested for both x86 and x64.

PR #4573, opened 2016-01-15
(PUP-5106) Add catalog_uuid identifier to catalog

whopper:
Currently, every agent run generates a unique transaction_uuid which
is attached to the catalog and report after the run has finished.
This allows the catalog to be linked to a specific report. However,
in the case of running off of cached catalogs, a single catalog can
have multiple reports.

When this happens, the catalog retains its original transaction_uuid
while the new reports contain the newly generated UUID, making it
impossible to link the cached catalog to any reports except for the
very first one.

This commit adds a new `catalog_uuid` identifier to the catalog,
which will allow a single catalog to be linked to multiple reports.

### DIFF
--- a/lib/puppet/resource/catalog.rb
+++ b/lib/puppet/resource/catalog.rb
@@ -3,6 +3,7 @@ require 'puppet/indirector'
 require 'puppet/transaction'
 require 'puppet/util/tagging'
 require 'puppet/graph'
+require 'securerandom'
 
 require 'puppet/resource/capability_finder'
 
@@ -31,6 +32,9 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
 
   # The id of the code input to the compiler.
   attr_accessor :code_id
+
+  # The UUID of the catalog
+  attr_accessor :catalog_uuid
 
   # How long this catalog took to retrieve.  Used for reporting stats.
   attr_accessor :retrieval_duration
@@ -272,6 +276,7 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
   def initialize(name = nil, environment = Puppet::Node::Environment::NONE, code_id = nil)
     super()
     @name = name
+    @catalog_uuid = SecureRandom.uuid
     @classes = []
     @resource_table = {}
     @resources = []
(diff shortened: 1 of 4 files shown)


api/schemas/report.json
```diff
@@ -93,6 +93,11 @@
             "type":        "string"
         },
 
+        "catalog_uuid": {
+            "description": "A UUID covering a specific catalog.",
+            "type":        "string"
+        },
+
```
joshcooper: I think this change snuck in.

lib/puppet/resource/catalog.rb
```diff
@@ -32,6 +33,9 @@ class DuplicateResourceError < Puppet::Error
   # The id of the code input to the compiler.
   attr_accessor :code_id
 
+  # The UUID of the catalog
+  attr_accessor :catalog_uuid
+
```
joshcooper: Can we add a spec test that newly created catalogs have a `catalog_uuid`?

lib/puppet/resource/catalog.rb
```diff
@@ -379,6 +384,10 @@ def self.from_data_hash(data)
       result.code_id = code_id
     end
 
+    if catalog_uuid = data['catalog_uuid']
+      result.catalog_uuid = catalog_uuid
+    end
+
```
joshcooper: Can we test that deserializing the catalog preserves the catalog_uuid?

lib/puppet/resource/catalog.rb
```diff
@@ -420,7 +429,8 @@ def to_data_hash
       'name'      => name,
       'version'   => version,
       'code_id'   => code_id,
-      'environment' => environment.to_s,
+      'catalog_uuid' => catalog_uuid,
+      'environment'  => environment.to_s,
```
joshcooper: Similar, that serializing the catalog writes a hash containing the catalog_uuid?

lib/puppet/resource/catalog.rb
```diff
@@ -536,6 +546,7 @@ def to_catalog(convert)
 
     result.version = self.version
     result.code_id = self.code_id
+    result.catalog_uuid = self.catalog_uuid
```
joshcooper: Can we add a spec test that the copy constructor behavior preserves the catalog_uuid?

joshcooper:
Code looks correct to me. I think there are some `code_id` unit tests that do similar things.

whopper:
@joshcooper hah yeah I kind of neglected those tests - added several, believe I've covered everything!

PR #4574, opened 2016-01-15
(PUP-5712) Fix so that TypeParser can handle negative numbers

thallgren:
Prior to this commit, the TypeParser would not recognize an
UnaryMinusExpression and hence raise an error when one was encountered.
This commit ensures that an unary minus is handled correctly so that
types like `Integer[default, -1]` can be used in a function dispatch.


hlindberg:
+1 on merging but waiting on green light to merge to stable (waiting for release work to take place).

PR #4575, opened 2016-01-15
(PUP-5702) Add code_id and catalog_uuid to the report

whopper:


whopper:
@joshcooper this is a (probably naive) first pass at threading this good stuff into the report. I went back and forth on whether I should alter the constructor of the `Report` class to take these as params or just add them to the report when applying the catalog (which is what I did).

Tests may or may not be a little light too - we don't seem to have a lot of `Report` tests that check serializing and deserializing.

lib/puppet/configurer.rb
```diff
@@ -229,6 +229,9 @@ def run_internal(options)
       end
 
       execute_prerun_command or return nil
+
+      options[:report].code_id = catalog.code_id
+      options[:report].catalog_uuid = catalog.catalog_uuid
```
whopper: Also I had some inner doubts about whether this would break anything or not when running with different agent / server version combos, as old catalogs wouldn't have these elements. I landed on thinking that it *wouldn't*, since I believe we are only preserving old agents talking to newer masters.

joshcooper: Good questions. We actually do want to preserve compatibility with new agents & old masters and vice-versa.

#### New agents & Older masters

Older masters won't send `catalog_uuid` so I would expect the agent to generate a new `catalog_uuid` when it deserializes the `/puppet/v3/catalog` response and include that in the report.

Masters before 4.3 and won't send `code_id`, and the agent will [ignore that it's missing](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/resource/catalog.rb#L383-L385), so the call to `catalog.code_id` should return `nil`.

FOSS 4.3.0 masters will send `code_id` with a `nil` value. And PE masters based on 4.3 will send the git sha. So I'd expect the agent to just transfer whatever is in the catalog to the report.

#### Old agents & New masters

Older agents won't send `code_id` or `catalog_uuid` in the report, and newer masters should ignore that they are missing, similar to how the agent ignores when `catalog_uuid` is missing.

There's a separate ticket to verify network compatibility as I'm making a lot of "it should do this" statements, and it'd be good to check that.

joshcooper:
Ran agent twice, once to get a new catalog and again using the cached catalog:

```
$ bundle exec puppet agent -t --server arcturus
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Caching catalog for arcturus.delivery.puppetlabs.net
Info: Applying configuration version '1453267493'
$ bundle exec puppet agent -t --server arcturus --use_cached_catalog
Info: Retrieving pluginfacts
Info: Retrieving plugin
Notice: Using cached catalog
Info: Applying configuration version '1453267493'
```

Verified the catalog contains catalog_uuid and code_id (note I had to edit master code to inject a `code_id=abc` like puppetserver does):

```
$ cat ~/.puppetlabs/opt/puppet/cache/client_data/catalog/arcturus.delivery.puppetlabs.net.json | jq .catalog_uuid
"cb09b406-29c7-450e-86a3-88feb65ef9e0"
$ cat ~/.puppetlabs/opt/puppet/cache/client_data/catalog/arcturus.delivery.puppetlabs.net.json | jq .code_id
"abc"
```

Both reports contain the same `catalog_uuid` and `code_id` as expected:

```
$ grep catalog_uuid  ~/test/master/catalog_uuid/reports/arcturus.delivery.puppetlabs.net/*.yaml
/Users/josh/test/master/catalog_uuid/reports/arcturus.delivery.puppetlabs.net/201601200524.yaml:catalog_uuid: cb09b406-29c7-450e-86a3-88feb65ef9e0
/Users/josh/test/master/catalog_uuid/reports/arcturus.delivery.puppetlabs.net/201601200525.yaml:catalog_uuid: cb09b406-29c7-450e-86a3-88feb65ef9e0
```

```
$ grep code_id  ~/test/master/catalog_uuid/reports/arcturus.delivery.puppetlabs.net/*.yaml
/Users/josh/test/master/catalog_uuid/reports/arcturus.delivery.puppetlabs.net/201601200524.yaml:code_id: abc
/Users/josh/test/master/catalog_uuid/reports/arcturus.delivery.puppetlabs.net/201601200525.yaml:code_id: abc
```

lib/puppet/configurer.rb
whopper: Excellent, thanks for the details, @joshcooper! Very good context.

PR #4576, opened 2016-01-18
(PUP-5637) Add ExecReload= on Puppet 3.x systemd unit

roidelapluie:


roidelapluie:
In Puppet 4.x this has been fixed with PUP-4918 (which contains a lot more than this change)

whopper:
@roidelapluie thanks for the contribution!

@melissa do you have any feelings / thoughts on this? We weren't totally sure what this unit file is used for anymore in Puppet 4.x, or whether vanagon has taken over generating this stuff. Does this seem sane for Puppet 3.x?

melissa:
@whopper this file gets used by vanagon. It is pulled in at https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/puppet.rb#L25

PR #4578, opened 2016-01-19
(PUP-5648) Add Iterable and Iterator to Puppet Type System

thallgren:
This PR adds the new types `Iterable` and `Iterator` to the Puppet
type system together with the runtime support needed to represent an
object that has been transformed into an Iterator by a call to
`Iterable.on`. This method transforms an instance of an object that
Puppet can use in iterative functions into an Iterator. The Iterator
in turn, provides basic support for the iterative functions.

The PR also changes the iterative functions `each`, `filter`,
`map`, `reduce`, and `slice` so that the type of their declared
_enumerable_ parameter is `Iterable` instead of `Any`

### DIFF
--- a/lib/puppet/pops/types/enumeration.rb
+++ b/lib/puppet/pops/types/enumeration.rb
@@ -1,33 +1,16 @@
 # The Enumeration class provides default Enumerable::Enumerator creation for Puppet Programming Language
 # runtime objects that supports the concept of enumeration.
 #
-class Puppet::Pops::Types::Enumeration
-  # Produces an Enumerable::Enumerator for Array, Hash, Integer, Integer Range, and String.
-  #
-  def self.enumerator(o)
-    @@singleton ||= new
-    @@singleton.enumerator(o)
-  end
+module Puppet::Pops::Types
+  class Enumeration
+    def self.enumerator(o)
+      Puppet.deprecation_warning('Enumeration.enumerator is deprecated. Use Iterable.on instead')
+      Iterable.on(o)
+    end
 
-  # Produces an Enumerator for Array, Hash, Integer, Integer Range, and String.
-  #
-  def enumerator(o)
-    case o
-    when String
-      x = o.chars
-      # Ruby 1.9.3 returns an Enumerator, and 2.0.0 an Array
-      x.is_a?(Array) ? x.each : x
-    when Integer
-      o.times
-    when Array
-      o.each
-    when Hash
-      o.each
-    when Puppet::Pops::Types::PIntegerType
-      # Not enumerable if representing an infinite range
-      o.enumerable? ? o.each : nil
-    else
-      nil
+    def enumerator(o)
+      Puppet.deprecation_warning('Enumeration.enumerator is deprecated. Use Iterable.on instead')
+      Iterable.on(o)
     end
   end
 end
--- a/lib/puppet/pops/types/iterable.rb
+++ b/lib/puppet/pops/types/iterable.rb
@@ -84,7 +84,7 @@ module Puppet::Pops::Types
         max = o.max
         if min.is_a?(Integer) && max.is_a?(Integer) && max >= min
           IntegerRangeIterator.new(PIntegerType.new(min, max))
-        elsif min.is_a?(String) && max.is_a?(String) && max > min
+        elsif min.is_a?(String) && max.is_a?(String) && max >= min
           # A generalized element type where only the size is inferred is used here since inferring the full
           # range might waste a lot of memory.
           if min.length < max.length
--- a/lib/puppet/pops/types/iterable.rb
+++ b/lib/puppet/pops/types/iterable.rb
@@ -1,0 +1,291 @@
+module Puppet::Pops::Types
+  # The runtime Iterable type for an Iterable
+  module Iterable
+    # Produces an `Iterable` for one of the following types with the following characterstics:
+    #
+    # `String`       - yields each character in the string
+    # `Array`        - yields each element in the array
+    # `Hash`         - yields each key/value pair as a two element array
+    # `Integer`      - when positive, yields each value from zero to the given number
+    # `PIntegerType` - yields each element from min to max (inclusive) provided min < max and neither is unbounded.
+    # `PEnumtype`    - yields each possible value of the enum.
+    # `Range`        - yields an iterator for all elements in the range provided that the range start and end
+    #                  are both integers or both strings and start is less than end using natural ordering.
+    # `Dir`          - yields each name in the directory
+    #
+    # An `ArgumentError` is raised for all other objects.
+    #
+    # @param o [Object] The object to produce an `Iterable` for
+    # @return [Iterable,nil] The produced `Iterable` or `nil` if it couldn't be produced
+    # @raise [ArgumentError] In case an `Iterable` cannot be produced
+    # @api public
+    def self.asserted_iterable(caller, obj)
+      iter = self.on(obj)
+      raise ArgumentError, "#{caller.class}(): wrong argument type (#{obj.class}; is not Iterable." if iter.nil?
+      iter
+    end
+
+    # Produces an `Iterable` for one of the following types with the following characterstics:
+    #
+    # `String`       - yields each character in the string
+    # `Array`        - yields each element in the array
+    # `Hash`         - yields each key/value pair as a two element array
+    # `Integer`      - when positive, yields each value from zero to the given number
+    # `PIntegerType` - yields each element from min to max (inclusive) provided min < max and neither is unbounded.
+    # `PEnumtype`    - yields each possible value of the enum.
+    # `Range`        - yields an iterator for all elements in the range provided that the range start and end
+    #                  are both integers or both strings and start is less than end using natural ordering.
+    # `Dir`          - yields each name in the directory
+    #
+    # The value `nil` is returned for all other objects.
+    #
+    # @param o [Object] The object to produce an `Iterable` for
+    # @return [Iterable,nil] The produced `Iterable` or `nil` if it couldn't be produced
+    #
+    # @api public
+    def self.on(o)
+      case o
+      when Iterable
+        o
+      when String
+        Iterator.new(PStringType.new(PIntegerType.new(1, 1)), o.each_char)
+      when Array
+        tc = TypeCalculator.singleton
+        Iterator.new(tc.unwrap_single_variant(PVariantType.new(o.map {|e| tc.infer_set(e) })), o.each)
+      when Hash
+        # Each element is a two element [key, value] tuple.
+        tc = TypeCalculator.singleton
+        Iterator.new(PTupleType.new([
+          tc.unwrap_single_variant(PVariantType.new(o.keys.map {|e| tc.infer_set(e) })),
+          tc.unwrap_single_variant(PVariantType.new(o.values.map {|e| tc.infer_set(e) }))], PIntegerType.new(2,2)), o.each_pair)
+      when Integer
+        o > 0 ? IntegerRangeIterator.new(PIntegerType.new(0, o - 1)) : nil
+      when PIntegerType
+        o.enumerable? ? IntegerRangeIterator.new(o) : nil
+      when PEnumType
+        Iterator.new(o, o.values)
+      when Range
+        min = o.min
+        max = o.max
+        if min.is_a?(Integer) && max.is_a?(Integer) && max > min
+          IntegerRangeIterator.new(PIntegerType.new(min, max))
+        elsif min.is_a?(String) && max.is_a?(String) && max > min
+          # A generalized element type where only the size is inferred is used here since inferring the full
+          # range might waste a lot of memory.
+          if min.length < max.length
+            shortest = min
+            longest = max
+          else
+            shortest = max
+            longest = min
+          end
+          Iterator.new(PStringType.new(PIntegerType.new(shortest.length, longest.length)), o.each)
+        else
+          # Unsupported range. It's either descending or nonsensical for other reasons (float, mixed types, etc.)
+          nil
+        end
+      when Dir
+        # Enumerable over file names
+        Iterator.new(PStringType::NON_EMPTY, o.each)
+      else
+        # Not supported. We cannot determine the element type
+        nil
+      end
+    end
+
+    # Answers the question if there is an end to the iteration. Puppet does not currently provide any unbounded
+    # iterables.
+    #
+    # @return [Boolean] `true` if the iteration is unbounded
+    def self.unbounded?(object)
+      case object
+      when Iterable
+        object.unbounded?
+      when String,Integer,Array,Hash,Enumerator,PIntegerType,PEnumType,Dir
+        false
+      else
+        TypeAsserter.assert_instance_of('', PIterableType::DEFAULT, object, false)
+        !object.respond_to?(:size)
+      end
+    end
+
+    def each(&block)
+      step(1, &block)
+    end
+
+    def element_type
+      PAnyType::DEFAULT
+    end
+
+    def reverse_each(&block)
+      # Default implementation cannot propagate reverse_each to a new enumerator so chained
+      # calls must put reverse_each last.
+      raise ArgumentError 'Step is not implemented'
+    end
+
+    def step(step, &block)
+      # Default implementation cannot propagate step to a new enumerator so chained
+      # calls must put stepping last.
+      raise ArgumentError 'Step is not implemented'
+    end
+
+    def to_a
+      raise Puppet::Error, 'Attempt to create an Array from an unbounded Iterable' if unbounded?
+      super
+    end
+
+    def unbounded?
+      true
+    end
+  end
+
+  # @api private
+  class Iterator
+    # Note! We do not include Enumerable module here since that would make this class respond
+    # in a bad way to all enumerable methods. We want to delegate all those calls directly to
+    # the contained @enumeration
+    include Iterable
+
+    def initialize(element_type, enumeration)
+      @element_type = element_type
+      @enumeration = enumeration
+    end
+
+    def element_type
+      @element_type
+    end
+
+    def size
+      @enumeration.size
+    end
+
+    def respond_to_missing?(name, include_private)
+      @enumeration.respond_to?(name, include_private)
+    end
+
+    def method_missing(name, *arguments, &block)
+      @enumeration.send(name, *arguments, &block)
+    end
+
+    def step(step, &block)
+      raise ArgumentError if step <= 0
+      r = self
+      r = r.step_iterator(step) if step > 1
+
+      if block_given?
+        begin
+        if block.arity == 1
+          loop { yield(r.next) }
+        else
+          loop { yield(*r.next) }
+        end
+        rescue StopIteration
+        end
+        self
+      else
+        r
+      end
+    end
+
+    def reverse_each(&block)
+      r = Iterator.new(@element_type, @enumeration.reverse_each)
+      block_given? ? r.each(&block) : r
+    end
+
+    def step_iterator(step)
+      StepIterator.new(@element_type, self, step)
+    end
+
+    def unbounded?
+      Iterable.unbounded?(@enumeration)
+    end
+  end
+
+  # @api private
+  class StepIterator < Iterator
+    include Enumerable
+
+    def initialize(element_type, enumeration, step_size)
+      super(element_type, enumeration)
+      raise ArgumentError if step_size <= 0
+      @step_size = step_size
+    end
+
+    def next
+      result = @enumeration.next
+      skip = @step_size - 1
+      if skip > 0
+        begin
+          skip.times { @enumeration.next }
+        rescue StopIteration
+        end
+      end
+      result
+    end
+
+    def reverse_each(&block)
+      r = Iterator.new(@element_type, to_a.reverse_each)
+      block_given? ? r.each(&block) : r
+    end
+
+    def size
+      super / @step_size
+    end
+  end
+
+  # @api private
+  class IntegerRangeIterator < Iterator
+    include Enumerable
+
+    def initialize(range, step = 1)
+      raise ArgumentError if step == 0
+      @range = range
+      @step_size = step
+      @current = (step < 0 ? range.to : range.from) - step
+    end
+
+    def element_type
+      @range
+    end
+
+    def next
+      value = @current + @step_size
+      if @step_size < 0
+        raise StopIteration if value < @range.from
+      else
+        raise StopIteration if value > @range.to
+      end
+      @current = value
+    end
+
+    def reverse_each(&block)
+      r = IntegerRangeIterator.new(@range, -@step_size)
+      block_given? ? r.each(&block) : r
+    end
+
+    def size
+      (@range.to - @range.from) / @step_size.abs
+    end
+
+    def step_iterator(step)
+      # The step iterator must use a range that has its logical end truncated at an even step boundary. This will
+      # fulfil two objectives:
+      # 1. The element_type method should not report excessive integers as possible numbers
+      # 2. A reversed iterator must start at the correct number
+      #
+      range = @range
+      step = @step_size * step
+      mod = (range.to - range.from) % step
+      if mod < 0
+        range = PIntegerType.new(range.from - mod, range.to)
+      elsif mod > 0
+        range = PIntegerType.new(range.from, range.to - mod)
+      end
+      IntegerRangeIterator.new(range, step)
+    end
+
+    def unbounded?
+      false
+    end
+  end
+end
--- a/lib/puppet/pops/types/iterable.rb
+++ b/lib/puppet/pops/types/iterable.rb
@@ -16,7 +16,7 @@ module Puppet::Pops::Types
     # An `ArgumentError` is raised for all other objects.
     #
     # @param o [Object] The object to produce an `Iterable` for
-    # @return [Iterable,nil] The produced `Iterable` or `nil` if it couldn't be produced
+    # @return [Iterable,nil] The produced `Iterable`
     # @raise [ArgumentError] In case an `Iterable` cannot be produced
     # @api public
     def self.asserted_iterable(caller, obj)
@@ -50,24 +50,39 @@ module Puppet::Pops::Types
       when String
         Iterator.new(PStringType.new(PIntegerType.new(1, 1)), o.each_char)
       when Array
-        tc = TypeCalculator.singleton
-        Iterator.new(tc.unwrap_single_variant(PVariantType.new(o.map {|e| tc.infer_set(e) })), o.each)
+        if o.empty?
+          Iterator.new(PUnitType::DEFAULT, o.each)
+        else
+          tc = TypeCalculator.singleton
+          Iterator.new(tc.unwrap_single_variant(PVariantType.new(o.map {|e| tc.infer_set(e) })), o.each)
+        end
       when Hash
         # Each element is a two element [key, value] tuple.
-        tc = TypeCalculator.singleton
-        Iterator.new(PTupleType.new([
-          tc.unwrap_single_variant(PVariantType.new(o.keys.map {|e| tc.infer_set(e) })),
-          tc.unwrap_single_variant(PVariantType.new(o.values.map {|e| tc.infer_set(e) }))], PIntegerType.new(2,2)), o.each_pair)
+        if o.empty?
+          Iterator.new(PHashType::DEFAULT_KEY_PAIR_TUPLE, o.each)
+        else
+          tc = TypeCalculator.singleton
+          Iterator.new(PTupleType.new([
+            tc.unwrap_single_variant(PVariantType.new(o.keys.map {|e| tc.infer_set(e) })),
+            tc.unwrap_single_variant(PVariantType.new(o.values.map {|e| tc.infer_set(e) }))], PHashType::KEY_PAIR_TUPLE_SIZE), o.each_pair)
+        end
       when Integer
-        o > 0 ? IntegerRangeIterator.new(PIntegerType.new(0, o - 1)) : nil
+        if o == 0
+          Iterator.new(PUnitType::DEFAULT, o.times)
+        elsif o > 0
+          IntegerRangeIterator.new(PIntegerType.new(0, o - 1))
+        else
+          nil
+        end
       when PIntegerType
-        o.enumerable? ? IntegerRangeIterator.new(o) : nil
+        # a finite range will always produce at least one element since it's inclusive
+        o.finite_range? ? IntegerRangeIterator.new(o) : nil
       when PEnumType
         Iterator.new(o, o.values)
       when Range
         min = o.min
         max = o.max
-        if min.is_a?(Integer) && max.is_a?(Integer) && max > min
+        if min.is_a?(Integer) && max.is_a?(Integer) && max >= min
           IntegerRangeIterator.new(PIntegerType.new(min, max))
         elsif min.is_a?(String) && max.is_a?(String) && max > min
           # A generalized element type where only the size is inferred is used here since inferring the full
@@ -84,9 +99,6 @@ module Puppet::Pops::Types
           # Unsupported range. It's either descending or nonsensical for other reasons (float, mixed types, etc.)
           nil
         end
-      when Dir
-        # Enumerable over file names
-        Iterator.new(PStringType::NON_EMPTY, o.each)
       else
         # Not supported. We cannot determine the element type
         nil
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -1,3 +1,13 @@
+require_relative 'iterable'
+require_relative 'enumeration'
+require_relative 'type_asserter'
+require_relative 'type_assertion_error'
+require_relative 'type_calculator'
+require_relative 'type_factory'
+require_relative 'type_parser'
+require_relative 'class_loader'
+require_relative 'type_mismatch_describer'
+
 require 'rgen/metamodel_builder'
 
 # The Types model is a model of Puppet Language types.
@@ -66,13 +76,7 @@ module Puppet::Pops
         false
       end
 
-      # Subclasses that are enumerable will override this method to return `true`
-      # @return [Boolean] false#
-      def enumerable?
-        is_a?(Enumerable)
-      end
-
-      # Generalizes value specific types. Types that are not value specific will return `self` otherwize
+      # Generalizes value specific types. Types that are not value specific will return `self` otherwise
       # the generalized type is returned.
       #
       # @return [PAnyType] The generalized type
@@ -88,6 +92,29 @@ module Puppet::Pops
       # @api private
       def kind_of_callable?(optional=true)
         false
+      end
+
+      # Returns `true` if an instance of this type is iterable, `false` otherwise
+      # The method #iterable_type must produce a `PIterableType` instance when this
+      # method returns `true`
+      #
+      # @return [Boolean] flag to indicate if instances of  this type is iterable.
+      def iterable?
+        false
+      end
+
+      # Returns the `PIterableType` that this type should be assignable to to, or `nil` if no such type exists.
+      # A type that returns a `PIterableType` must respond `true` to `#iterable?`.
+      #
+      # Any Collection[T] is assignable to an Iterable[T]
+      # A String is assignable to an Iterable[String] iterating over the strings characters
+      # An Integer is assignable to an Iterable[Integer] iterating over the 'times' enumerator
+      # A Type[T] is assignabel to an Iterable[Type[T]] if T is an Integer or Enum
+      #
+      # @return [PIterableType,nil] The iterable type that this type is assignable to or `nil`
+      # @api private
+      def iterable_type
+        nil
       end
 
       def hash
@@ -201,6 +228,29 @@ module Puppet::Pops
 
       def hash
         31 * @type.hash
+      end
+
+      def iterable?
+        case @type
+        when PEnumType
+          true
+        when PIntegerType
+          @type.enumerable?
+        else
+          false
+        end
+      end
+
+      def iterable_type
+        # The types PIntegerType and PEnumType are Iterable
+        case @type
+        when PEnumType
+          @type.each
+        when PIntegerType
+          @type.enumerable? ? @type.each : nil
+        else
+          nil
+        end
       end
 
       def ==(o)
@@ -368,8 +418,6 @@ module Puppet::Pops
     # @api public
     #
     class PEnumType < PScalarType
-      include Enumerable
-
       attr_reader :values
 
       def initialize(values)
@@ -378,12 +426,17 @@ module Puppet::Pops
 
       # Returns Enumerator if no block is given, otherwise, calls the given
       # block with each of the strings for this enum
-      def each
-        if block_given?
-          values.each { |x| yield x }
-        else
-          values.to_enum
-        end
+      def each(&block)
+        r = Iterable.on(self)
+        block_given? ? r.each(&block) : r
+      end
+
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        PIterableType.new(PStringType::DEFAULT)
       end
 
       def hash
@@ -483,9 +536,10 @@ module Puppet::Pops
     # @api public
     #
     class PIntegerType < PNumericType
-      # The integer type is enumerable when it defines a range
-      include Enumerable
-
+      # Answers the question, "is this instance of PIntegerType enumerable?" (as opposed to if instances described by
+      # this type is enumerable). Will respond `true` for any range that is bounded at both ends.
+      #
+      # @return [Boolean] `true` if the type is enumerable.
       def enumerable?
         @from != -Float::INFINITY && @to != Float::INFINITY
       end
@@ -496,6 +550,15 @@ module Puppet::Pops
 
       def instance?(o)
         o.is_a?(Integer) && o >= numeric_from && o <= numeric_to
+      end
+
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        # It's unknown if the iterable will be a range (min, max) or a "times" (0, max)
+        PIterableType.new(PIntegerType::DEFAULT)
       end
 
       # Returns Float.Infinity if one end of the range is unbound
@@ -512,12 +575,9 @@ module Puppet::Pops
 
       # Returns Enumerator if no block is given
       # Returns nil if size is infinity (does not yield)
-      def each
-        if block_given?
-          enumerable? ? @from.upto(@to) { |x| yield x } : nil
-        else
-          to_enum
-        end
+      def each(&block)
+        r = Iterable.on(self)
+        block_given? ? r.each(&block) : r
       end
 
       # Returns a range where both to and from are positive numbers. Negative
@@ -571,6 +631,14 @@ module Puppet::Pops
         @element_type.hash * 31 + @size_type.hash
       end
 
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        PIterableType.new(element_type)
+      end
+
       def ==(o)
         self.class == o.class && @element_type == o.element_type && @size_type == o.size_type
       end
@@ -603,6 +671,111 @@ module Puppet::Pops
       end
     end
 
+    class PIterableType < PAnyType
+      attr_reader :element_type
+
+      def initialize(type)
+        @element_type = type
+      end
+
+      def instance?(o)
+        if @element_type.nil? || @element_type.assignable?(PAnyType::DEFAULT)
+          # Any element_type will do
+          case o
+          when Iterable, String, Hash, Enumerable, Enumerator, Range, PEnumType
+            true
+          when Integer
+            o > 0
+          when PIntegerType
+            o.enumerable?
+          else
+            false
+          end
+        else
+          assignable?(TypeCalculator.infer(o))
+        end
+      end
+
+      def generalize
+        @element_type.nil? ? DEFAULT : PIterableType.new(@element_type.generalize)
+      end
+
+      def hash
+        67 * @element_type.hash
+      end
+
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        self
+      end
+
+      def ==(o)
+        self.class == o.class && @element_type == o.element_type
+      end
+
+      DEFAULT = PIterableType.new(nil)
+
+      protected
+
+      # @api private
+      def _assignable?(o)
+        if @element_type.nil? || @element_type.assignable?(PAnyType::DEFAULT)
+          # Don't request the iterable_type since it might be expensive to compute the
+          # type of its elements
+          o.iterable?
+        else
+          o = o.iterable_type
+          o.nil? || o.element_type.nil? ? false : @element_type.assignable?(o.element_type)
+        end
+      end
+    end
+
+    # @api public
+    #
+    class PIteratorType < PAnyType
+      attr_reader :element_type
+
+      def initialize(type)
+        @element_type = type
+      end
+
+      def instance?(o)
+        o.is_a?(Iterable) && (@element_type.nil? || @element_type.assignable?(o.element_type))
+      end
+
+      def generalize
+        @element_type.nil? ? DEFAULT : PIteratorType.new(@element_type.generalize)
+      end
+
+      def hash
+        71 * @element_type.hash
+      end
+
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        element_type.nil? ? PIteratbleType::DEFAULT : PIterableType.new(@element_type)
+      end
+
+      def ==(o)
+        self.class == o.class && @element_type == o.element_type
+      end
+
+      DEFAULT = PIteratorType.new(nil)
+
+      protected
+
+      # @api private
+      def _assignable?(o)
+        o.is_a?(PIteratorType) && (@element_type.nil? || @element_type.assignable?(o.element_type))
+      end
+    end
+
     # @api public
     #
     class PStringType < PScalarType
@@ -619,6 +792,14 @@ module Puppet::Pops
 
       def hash
         @size_type.hash * 31 + @values.hash
+      end
+
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        PIterableType.new(PStringType::DEFAULT)
       end
 
       def ==(o)
@@ -840,9 +1021,24 @@ module Puppet::Pops
       def hashed_elements
         @hashed ||= @elements.reduce({}) {|memo, e| memo[e.name] = e; memo }
       end
- 
+
       def hash
         @elements.hash
+      end
+
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        if self == DEFAULT
+          PIterableType.new(PTupleType.new([PAnyType::DEFAULT], PHashType::TUPLE_SIZE))
+        else
+          tc = TypeCalculator.singleton
+          key_type = tc.infer_and_reduce_type(@elements.map {|se| se.key_type })
+          value_type = tc.infer_and_reduce_type(@elements.map {|se| se.value_type })
+          PIterableType.new(PTupleType.new([key_type, value_type], PHashType::TUPLE_SIZE))
+        end
       end
 
       def ==(o)
@@ -981,6 +1177,14 @@ module Puppet::Pops
           return false unless (types[index] || types[-1]).instance?(element)
         end
         true
+      end
+
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        PIterableType.new(TypeCalculator.singleton.infer_and_reduce_type(types))
       end
 
       # Returns the number of elements accepted [min, max] in the tuple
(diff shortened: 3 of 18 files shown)


hlindberg:
This must be broken up into multiple commits:

- adding the iterable type with related tests
- one commit per existing function switching to support iterator
- one commit per added function

I can also see breaking this up into multiple tickets; the features do need separate documentation etc.

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,88 @@
+# Reverses the order of the elements of something that is iterable and optionally runs a
```
hlindberg: We must check if it is ok to just define the 4.x function, or if we have to do a 3.x version only to get the documentation generated - ping @nfagerlund

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,88 @@
...
+# $data.reverse_each |$item| {
+#  notify { $item:
+#    message => $item
+#  }
+# }
+# # Puppet creates one resource for each of the three items in $data in reverse order. Each resource is
+# # named after the item's value and uses the item's value in a parameter.
+# ~~~
+#
+# When no second argument is present, Puppet returns a new iterable so that a new function that takes
```
hlindberg: When no block is given...

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,88 @@
...
+# $data = [1,2,3]
+# $transformed_data = map(reverse_each($data)) |$item| { $item * 10 }
+# # $transformed_data contains [30,20,10]
+# ~~~
+#
+# @since 4.4.0
+#
+Puppet::Functions.create_function(:reverse_each) do
+  dispatch :reverse_each do
+    param 'Any', :iterable
```
hlindberg: Can we have tighter checking here instead of delegating to Types::Iterable? Specific types would also show up in documentation (which is better than seeing Any).

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  dispatch :reverse_each_block do
+    param 'Any', :iterable
```
hlindberg: Same here

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+# $data = Integer[0,20]
+# $transformed_data = map(step($data, 5)) |$item| { $item * 10 }
+# # $transformed_data contains [0,50,100,150,200]
+# ~~~
+#
+# @since 4.4.0
+#
+Puppet::Functions.create_function(:step) do
+  dispatch :step do
+    param 'Any', :iterable
```
hlindberg: same here; Any

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+# This function takes two to three arguments.
+#
+# 1. An iterable that the function will iterate over.
+# 2. An integer step factor. This must be a positive integer.
+# 3. An optional lambda, which the function calls for each element in the interval. It must
+# request one parameter.
+#
+# @example Using the `step` function
+#
+# `$data.step(<n>) |$parameter| { <PUPPET CODE BLOCK> }`
```
hlindberg: needs to be concrete since it is an example - no <n> or <PUPPET CODE BLOCK>.

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+#
+# or
+#
+# `step($data, <n>) |$parameter| { <PUPPET CODE BLOCK> }`
+#
+# or
+#
+# `$stepped_data = step($data, <n>)`
+#
+# When no third argument is present, Puppet returns an iterable that represents the nth element
```
hlindberg: When no block is given

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+# of its first argument. This allows methods on iterables to be chained.
```
hlindberg: change "methods" to "functions"

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+#
+# When a lamdba is given as the third argument, Puppet iterates the first argument and passes each
```
hlindberg: lambdas are not given as arguments in the perspective of a user - it is a given block. Rephrase.

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+# nth value in turn to the lambda, then returns the first argument unchanged.
```
hlindberg: use "block" instead of lambda

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+#
+# @example Using the `step` function with an array, a step factor, and a one-parameter lambda
```
hlindberg: here lambda is fine, since it is on the calling side, the mechanism with which a block is given to the function. (Pointing this out just to avoid the otherwise obvious question).

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+#
+# ~~~ puppet
+# # For the array $data, run a lambda that creates a resource in reverse order for each item in reverse.
```
hlindberg: "run a lambda" => "call a lambda". The description seems wrong - the lambda does not reverse (copy/paste error?), the lambda just notices. Here it is the step function that picks every 3:d element that the lambda notices.

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+# $data = [1,2,3,4,5,6,7,8]
+# $data.step(3) |$item| {
+#  notice($item)
+# }
+# # Puppet will output 3 notices with the texts '1', '4', '7'.
```
hlindberg: The typical documentation language for this would be "Puppet notices the values '1', '4' and '7'" (i.e. the concept of "noticing" is known to the reader).

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+# ~~~
+#
+# When no lambda is present, Puppet returns a new iterable so that a new function that takes
```
hlindberg: "When no lambda is given", or "When no block is present" (depending on point of view).

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+# an iterable as an argument can use it as input.
+#
+# @example Using the `step` function chained with a `map` function.
+#
+# # For the array $data, return an array containing each value multiplied by 10 in reverse order
```
hlindberg: probably better to say "each value in reverse order multiplied by 10" (as you may otherwise wonder how a "multiplication in reverse" is performed.

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,88 @@
...
+# $data = Integer[0,20]
+# $transformed_data = $data.step(5).map |$item| { $item * 10 }
+# # $transformed_data contains [0,50,100,150,200]
+# ~~~
+#
+# @example The same example using `step` function chained with a `map` in alternative syntax
+#
+# # For the array $data, return an array containing each value multiplied by 10 in reverse order
```
hlindberg: same remark as on the previous example.

lib/puppet/pops/types/enumeration.rb
```diff
@@ -1,33 +1,70 @@
...
+  # @api public
   def enumerator(o)
     case o
     when String
-      x = o.chars
-      # Ruby 1.9.3 returns an Enumerator, and 2.0.0 an Array
-      x.is_a?(Array) ? x.each : x
-    when Integer
-      o.times
+      EnumerationIterable.new(PStringType::DEFAULT, o.each_char)
```
hlindberg: Does this work on 1.9.3? (I suppose travis will tell)

spec/shared_behaviours/iterative_functions.rb
```diff
@@ -67,3 +59,25 @@
...
+  it 'raises an error when called without a block' do
+    expect do
+      compile_to_catalog(<<-MANIFEST)
+          [1].#{func}
+      MANIFEST
+    end.to raise_error(Puppet::Error, /expects a block/)
+  end
+end
+
+shared_examples_for 'all iterative functions that requires no block' do |func|
```
hlindberg: does this mean they optionally accept a block, or that they have no opinion what so ever about a block?

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'puppet'
+require 'spec_helper'
+require 'puppet_spec/compiler'
+
+require 'shared_behaviours/iterative_functions'
+
+describe 'the reverse_each method' do
```
hlindberg: "method" => "function"

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  include PuppetSpec::Compiler
+
+  it 'raises an error when used against an unsupported type' do
```
hlindberg: "raises an error when given an unsupported type"

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    expect do
+      compile_to_catalog(<<-MANIFEST)
+        3.14.reverse_each |$v| {  }
+      MANIFEST
+    end.to raise_error(Puppet::Error, /Cannot be coerced into an Iterable/)
+  end
+
+  it 'raises an error when called with any parameters besides a block' do
```
hlindberg: "parameters" => "arguments"

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+      MANIFEST
+    end.to raise_error(Puppet::Error, /expects 1 argument, got 2/)
+  end
+
+  it 'raises an error when called with something that is not a block' do
+    expect do
+      compile_to_catalog(<<-MANIFEST)
+        [1].reverse_each(1)
+      MANIFEST
+    end.to raise_error(Puppet::Error, /expects 1 argument, got 2/)
```
hlindberg: This test looks strage - it fails on getting two arguments (ike example above), not on being given something that is not a block) - you cannot really do that in puppet source as the only way to get a block there is to specify one.

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  it 'should produce "times" interval of integer in reverse' do
+    expect(eval_and_collect_notices('5.reverse_each |$x| { notice($x) }')).to eq(['4', '3', '2', '1', '0'])
+  end
+
+  it 'should produce range Integer[5,8] in reverse' do
+    expect(eval_and_collect_notices('Integer[5,8].reverse_each |$x| { notice($x) }')).to eq(['8', '7', '6', '5'])
+  end
+
+  it 'should produce negative nth element of range Integer[5,20] when chained after a step' do
```
hlindberg: decending ? (n is never negative)

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    expect(eval_and_collect_notices('Integer[5,20].step(4).reverse_each |$x| { notice($x) }')).to eq(['17', '13', '9', '5'])
+  end
+
+  it 'should produce negative nth element of times 5 when chained after a step' do
```
hlindberg: descending

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    expect(eval_and_collect_notices('5.step(2).reverse_each |$x| { notice($x) }')).to eq(['4', '2', '0'])
+  end
+
+  it 'should produce negative nth element of range Integer[5,20] when chained after a step' do
```
hlindberg: descending

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    expect(eval_and_collect_notices('[5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20].step(4).reverse_each |$x| { notice($x) }')).to eq(['17', '13', '9', '5'])
```
hlindberg: long line, please break it up

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  it_should_behave_like 'all iterative functions that requires no block', 'reverse_each'
```
hlindberg: suggested name change since "requires no block" is not very clear

spec/unit/functions/step_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'puppet'
+require 'spec_helper'
+require 'puppet_spec/compiler'
+
+require 'shared_behaviours/iterative_functions'
+
+describe 'the step method' do
+  include PuppetSpec::Compiler
+
+  it 'raises an error when used against an unsupported type' do
```
hlindberg: "raises an error when give an argument of unsupported type" (or even better), "of a type that cannot be iterated"

spec/unit/functions/step_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    expect do
+      compile_to_catalog(<<-MANIFEST)
+        3.14.step(1) |$v| {  }
+      MANIFEST
+    end.to raise_error(Puppet::Error, /Cannot be coerced into an Iterable/)
+  end
+
+  it 'raises an error when called with any parameters besides a block' do
```
hlindberg: "parameters" => "arguments"

spec/unit/functions/step_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      MANIFEST
+    end.to raise_error(Puppet::Error, /expects 2 arguments, got 3/)
+  end
+
+  it 'raises an error when called with something that is not a block' do
+    expect do
+      compile_to_catalog(<<-MANIFEST)
+        [1].step(1,2)
+      MANIFEST
+    end.to raise_error(Puppet::Error, /expects 2 arguments, got 3/)
```
hlindberg: as before this test is questionable - unless tested with a call that is constructed in ruby

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,121 @@
+require 'spec_helper'
+require 'puppet/pops'
+
+module Puppet::Pops::Types
+describe 'The enumeration support' do
+  it 'produces an iterable for Hash' do
+    expect(Iterable.on({:a=>1})).to respond_to(:step)
```
hlindberg: Can this be made more precise? Or is the entire Iterable API based on object having a step method?

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  it 'produces a reverse iterator on an integer' do
+    expect{ |b| Iterable.on(5).reverse_each(&b) }.to yield_successive_args(4,3,2,1,0)
+  end
+
+  it 'produces an iterable on a integer range' do
+    expect{ |b| Iterable.on(2..7).each(&b) }.to yield_successive_args(2,3,4,5,6,7)
+  end
+
+  it 'produces correct element type for an iterable on a integer range' do
```
hlindberg: replace "correct" with what it explicitly does

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  it 'produces correct element type for an iterable on a character range' do
+    expect(Iterable.on('a'..'f').element_type).to eq(PStringType::DEFAULT)
+  end
+
+  it 'produces an iterable on a bounded Integer type' do
+    expect{ |b| Iterable.on(PIntegerType.new(2,7)).each(&b) }.to yield_successive_args(2,3,4,5,6,7)
+  end
+
+  it 'produces correct element type for an iterable on bounded Integer type' do
```
hlindberg: "correct" is vague, be explicit

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    expect(Iterable.on(PIntegerType.new(2,7)).element_type).to eq(PIntegerType.new(2,7))
+  end
+
+  it 'produces an iterable on String' do
+    expect{ |b| Iterable.on('eat this').each(&b) }.to yield_successive_args('e', 'a', 't', ' ', 't', 'h', 'i', 's')
+  end
+
+  it 'produces correct element type for an iterable on Array' do
```
hlindberg: "correct" is vague

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    expect(Iterable.on('eat this').element_type).to eq(PStringType::DEFAULT)
+ end
+
+  it 'produces an iterable on Array' do
+    expect{ |b| Iterable.on([1,5,9]).each(&b) }.to yield_successive_args(1,5,9)
+  end
+
+  it 'produces correct element type for an iterable on Array' do
```
hlindberg: "correct"

spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -234,6 +235,14 @@ def type_from_class(c)
...
     end
 
+    it 'enumerator translates to PIterableType' do
+      expect(calculator.infer([1,2].each)).to be_a(Puppet::Pops::Types::PIterableType)
+    end
+
+    it 'iterable translates to PIterableType' do
+      expect(calculator.infer(Puppet::Pops::Types::Iterable.on(1))).to be_a(Puppet::Pops::Types::PIterableType)
+    end
+
```
hlindberg: As noted earlier, it seems better to have two types; `Iterable` - which includes all the kinds of things for which an Iterator can be created, and `Iterator`, the thing that iterates. That way, the functions can use  a signature of `Iterable` instead of `Any`, and we can document that `Iterable` matches String, Integer, Array, Hash, and Iterator (and later possibly Range which we do not yet have as a runtime type in the Puppet Language, nor in the type system).

lib/puppet/pops/types/enumeration.rb
hlindberg: Yes it works. Test is green on 1.9.3

lib/puppet/functions/reverse_each.rb
thallgren: I like the concept of being able to describe that something is transformable into something else. Could you create a specific PR for that? And meanwhile, perhaps we could keep Any here since it's rather complicated to describe with a Variant.
```
Variant[Integer,Type[Integer[<with range spec>],String[1],Array[1],Hash[1],Iterable,...]
```
Perhaps we'd like to make the Type[Tuple] and a and Type[Struct] Iterable? If we do, then that too (and any other changes we make in the future), will have an impact on the documentation of this method.

Also, for the sake of consistency, `Any` is used as the type for the enumerable in the functions `each`, `filter`, `reduce`, and `map` so a tighter checking here should imply a tighter checking in  them too.

lib/puppet/functions/step.rb
thallgren: Using "or" in examples seems to have been widely accepted. I followed the style that was used in `each`, `filter`,  `map`, and `reduce` to make the documentation of these functions consistent and easy to compare.

lib/puppet/functions/step.rb
thallgren: I agree that it should be called a block and that is't not an argument. In general though, I find a lot of the function documentations confusing because many of the mentions that the function takes <n> "arguments" and one of the arguments is often a "lambda" (see `each` for instance). Perhaps we should have a JIRA on how to be more consistent about this throughout in the documentation.

lib/puppet/functions/step.rb
thallgren: My preference would be to use the term block throughout. The distinction between the terms block and lambda is fairly subtle and using both doesn't make things clearer. Actually, I think there's a risk that it adds confusion.

lib/puppet/functions/step.rb
thallgren: Oops. Definitely a copy paste error.

lib/puppet/functions/step.rb
thallgren: Uhm, in your comment above you wanted "When no block is given" (which sort of proves my point on using both terms). I think I'll stick with that.

lib/puppet/functions/step.rb
thallgren: Yes, or even better, "each 5th value in reverse order multiplied by 10" :-)

spec/shared_behaviours/iterative_functions.rb
thallgren: In this context it means that they do not require a block. Whether or not they optionally accepts one isn't tested here.

thallgren:
PR updated so that:

1. It now only adds the `Iterable` type. No new functions are added.
2. It's divided into two commits. One that adds the type and one that removes the many times duplicated method `asserted_enumerable` in favor of one single method declared on the `Enumeration` class.

lib/puppet/pops/types/enumeration.rb
```diff
@@ -1,33 +1,72 @@
...
+  # Produces an `Enumerator` for `String`, `Integer`, `PIntegerType` or anything that respond to `each` by
+  # returning an `Enumerator`. The argument is returned in case it already is an `Enumerator`.
   #
+  # @param o [Object] The object to produce an `Enumerator` for
+  # @return [Enumerator,nil] The produced `Enumerator` or `nil` if it couldn't be produced
+  # @api public
   def self.enumerator(o)
     @@singleton ||= new
     @@singleton.enumerator(o)
   end
```
hlindberg: I wonder why I did it like this. Don't think we need a singleton as opposed to just using the class method `enumerator` directly. Can't see we want to use different implementations at runtime, so no need to pass one around etc.

lib/puppet/pops/types/enumeration.rb
```diff
@@ -1,33 +1,72 @@
...
-      o.times
+      EnumerationIterable.new(PStringType::DEFAULT, o.each_char)
     when Array
-      o.each
-    when Hash
-      o.each
-    when Puppet::Pops::Types::PIntegerType
-      # Not enumerable if representing an infinite range
-      o.enumerable? ? o.each : nil
+      EnumerationIterable.new(TypeCalculator.singleton.infer_and_reduce_type(o), o.each)
```
hlindberg: Is this really what is wanted? (infer_and_reduce). Consider the resulting type of `[1,2, 101, 102]`. I think this will be `Iterable[Integer[1,102]`. Which will then not match the more precisely specified `Interable[Variant[Integer[1,10]], Variant[Integer[100,110]]`. If the Iterable retained the full inference it would be of more value.

lib/puppet/pops/types/enumeration.rb
```diff
@@ -1,33 +1,72 @@
...
+    when Integer
+      o > 0 ? IntegerRangeIterator.new(PIntegerType.new(0, o - 1)) : nil
```
hlindberg: I think this means that an integer behaves as if you did `n.times`. Not sure that is great since if an integer contains something to iterate over, then it would be its bits. Seeing something like this:
```
3.each [ $x | { notice $x }
```
is confusing - is that an iteration of only [3], or [1,2,3], or [0,1,2,3], or [0,1,2]? (I see the impl is the range 0..(n-1).
Think a `times´ function is needed instead of directly getting an enumerator.

lib/puppet/pops/types/enumeration.rb
```diff
@@ -1,33 +1,72 @@
...
 
-  # Produces an Enumerator for Array, Hash, Integer, Integer Range, and String.
+  # Produces an `Enumerator` for `String`, `Integer`, `Range`, `PIntegerType` or anything that respond to `each` by
+  # returning an `Enumerator`. The argument is returned in case it already is an `Enumerator` or an `Iterable`.
   #
+  # @param o [Object] The object to produce an `Enumerator` for
+  # @return [Enumerator,nil] The produced `Enumerator` or `nil` if it couldn't be produced
+  # @api public
   def enumerator(o)
     case o
```
hlindberg: It would be nice to add iteration of all PTypes that contain types (but that is a separate ticket).

lib/puppet/pops/types/enumeration.rb
```diff
@@ -1,33 +1,72 @@
...
+      EnumerationIterable.new(PStringType::DEFAULT, o.values)
+    when Range
+      min = o.min
+      if min.is_a?(Integer)
+        max = o.max
+        max > min ? IntegerRangeIterator.new(PIntegerType.new(min, max)) : nil
+      else
+        # A generalized form of the inferred type for the 'min' is used here since inferring the full
+        # range might waste a lot of memory.
+        EnumerationIterable.new(TypeCalculator.singleton.infer(min).generalize, o.each)
```
hlindberg: if type is a string, you know that the length of the resulting strings will be between length of the shortest, and length of the longest. To support this in the puppet language we are going to need a PRange type as well, but nice to infer a slightly richer type for strings here. (It will be used since people do generate sequences of unique names).

For Float, the min, max are known. 

Ruby allows creation of ranges like this `([1]..[10])`, which is kind of useless except as a Tuple of two values since most default range operations do not work - it will fail on the `o.each` (that is probably good enough protection against bad stuff).

lib/puppet/pops/types/enumeration.rb
```diff
@@ -1,33 +1,72 @@
...
+      end
+    when Enumerator, Iterable
+      o
     else
-      nil
+      begin
+        result = o.each
```
hlindberg: responds_to? is better here, since you may otherwise hide an error inside of the each implementation if it calls an undefined method.

lib/puppet/pops/types/enumeration.rb
```diff
@@ -1,33 +1,72 @@
...
+        result.is_a?(Enumerator) ? result : nil
```
hlindberg: Hm, we have no type information here, requires you to accept `Iterable[Any]`. Do we really need this very open ended solution? Can we require a richer API? As an example, if we allow producer functions, and we have return types, we do know the produced type. Perhaps have either another method enumerator(o, T) that you need to use if what you created does not have type information. (That implies that the values that will be emitted needs to be type checked before yielded).

lib/puppet/pops/types/enumeration.rb
thallgren: This is how the TypeCalculator will infer the element type of an Array, and hence, also how the element type of the Iterator is calculated. It could be improved of course (the infer_and_reduce method), but it's not in scope for this PR IMO.

lib/puppet/pops/types/enumeration.rb
thallgren: I also reacted to this but decided not to change it since it's not related to this PR.

lib/puppet/pops/types/enumeration.rb
thallgren: The code used to read:
```
    when Integer
      o.times
```
I totally agree that a dedicated `times` function would much better. This PR doesn't change this behavior however. A separate JIRA perhaps?

lib/puppet/pops/types/enumeration.rb
thallgren: Let's discuss. I added this as a fallback. Thought it would be better than to just bail out. Perhaps it isn't.

lib/puppet/pops/types/enumeration.rb
hlindberg: ok, we can clean that up later.

lib/puppet/pops/types/enumeration.rb
hlindberg: But there is another way of doing the infer that retains the information of all the values. That is only possible when inferring from values. i.e. why not use `infer_set` and skip reduction to common single element type?

lib/puppet/pops/types/enumeration.rb
hlindberg: Ah, I misread since these lines looked like they were added, So I thought it was a new feature.
Since it is there already... :hear_no_evil: :see_no_evil: :speak_no_evil:

lib/puppet/pops/types/enumeration.rb
hlindberg: @thallgren easy to add better inference here for Range of String, Integer, Float.  I think it is easier to add support for those here now, that to do a separate ticket and all the admin around that.

I played with enumeration of Ruby Range, it it errors on [1..3.13], or ["a"..10] etc. The infer and reduce here is therefore not that valuable.

lib/puppet/pops/types/enumeration.rb
hlindberg: @thallgren ping

lib/puppet/pops/types/enumeration.rb
hlindberg: I think it is better to bail out since there is loss of fidelity in that last case anyway, and that can cause problems if there are other mistakes made elsewhere that makes it end up here.

lib/puppet/pops/types/enumeration.rb
thallgren: The motivation for the current implementation is consistency. An `Array[T]` should yield `Iterable[T]` in my opinion. Do you see that differently?

thallgren:
Pushed a new commit that I believe addresses all concerns, both regarding too relaxed element types and uncertainties on what types that are supported. That set is now very distinct.

lib/puppet/pops/types/iterable.rb
```diff
@@ -0,0 +1,291 @@
...
+    # `PEnumtype`    - yields each possible value of the enum.
+    # `Range`        - yields an iterator for all elements in the range provided that the range start and end
+    #                  are both integers or both strings and start is less than end using natural ordering.
+    # `Dir`          - yields each name in the directory
+    #
+    # An `ArgumentError` is raised for all other objects.
+    #
+    # @param o [Object] The object to produce an `Iterable` for
+    # @return [Iterable,nil] The produced `Iterable` or `nil` if it couldn't be produced
+    # @raise [ArgumentError] In case an `Iterable` cannot be produced
```
hlindberg: This contradicts the line above

lib/puppet/pops/types/iterable.rb
```diff
@@ -0,0 +1,291 @@
...
+          else
+            shortest = max
+            longest = min
+          end
+          Iterator.new(PStringType.new(PIntegerType.new(shortest.length, longest.length)), o.each)
+        else
+          # Unsupported range. It's either descending or nonsensical for other reasons (float, mixed types, etc.)
+          nil
+        end
+      when Dir
```
hlindberg: Are you using this for something? Otherwise I see no reason to add this.

lib/puppet/pops/types/types.rb
```diff
@@ -90,6 +94,29 @@ def kind_of_callable?(optional=true)
...
+      # Returns `true` if an instance of this type is iterable, `false` otherwise
+      # The method #iterable_type must produce a `PIterableType` instance when this
+      # method returns `true`
+      #
+      # @return [Boolean] flag to indicate if instances of  this type is iterable.
+      def iterable?
+        false
+      end
+
+      # Returns the `PIterableType` that this type should be assignable to to, or `nil` if no such type exists.
```
hlindberg: typo "to to"

lib/puppet/pops/types/types.rb
```diff
@@ -90,6 +94,29 @@ def kind_of_callable?(optional=true)
...
+      # A type that returns a `PIterableType` must respond `true` to `#iterable?`.
+      #
+      # Any Collection[T] is assignable to an Iterable[T]
+      # A String is assignable to an Iterable[String] iterating over the strings characters
+      # An Integer is assignable to an Iterable[Integer] iterating over the 'times' enumerator
+      # A Type[T] is assignabel to an Iterable[Type[T]] if T is an Integer or Enum
+      #
```
hlindberg: These are just examples, since the actual implementation that makes these true are elsewhere (in the respective types). Maybe point that out.

lib/puppet/pops/types/types.rb
```diff
@@ -203,6 +230,29 @@ def hash
         31 * @type.hash
       end
 
+      def iterable?
+        case @type
+        when PEnumType
+          true
+        when PIntegerType
+          @type.enumerable?
```
hlindberg: isn't that method deprecated?  - Should this be `@type.iterable?`

lib/puppet/pops/types/types.rb
```diff
@@ -203,6 +230,29 @@ def hash
...
+        end
+      end
+
+      def iterable_type
+        # The types PIntegerType and PEnumType are Iterable
+        case @type
+        when PEnumType
+          @type.each
+        when PIntegerType
+          @type.enumerable? ? @type.each : nil
```
hlindberg: same here, iterable? vs. enumerable?

lib/puppet/pops/types/types.rb
```diff
@@ -378,12 +426,17 @@ def initialize(values)
...
+        r = Iterable.on(self)
+        block_given? ? r.each(&block) : r
+      end
+
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        PIterableType.new(PStringType::DEFAULT)
```
hlindberg: Doesn't this lose precision? For an Enum, the strings are known. so it is possible to do either Iterable[String[min,max]], or why not just Iterable[<this enum>] ?

lib/puppet/pops/types/types.rb
thallgren: No. `iterable?` and `enumerable?` has different meaning (type versus instances of the type). I did add specific method documentation to explain what they do.

lib/puppet/pops/types/iterable.rb
thallgren: It was the only implementation in Ruby that we doesn't otherwise cover that seemed to make sense. I can remove it.

lib/puppet/pops/types/types.rb
thallgren: It should be changed to `Iterable[String[1,1]]` since an iterable on a string yields all it's characters.

lib/puppet/pops/types/types.rb
```diff
@@ -603,6 +671,111 @@ def _assignable?(o)
...
+      end
+
+      def instance?(o)
+        if @element_type.nil? || @element_type.assignable?(PAnyType::DEFAULT)
+          # Any element_type will do
+          case o
+          when Iterable, String, Hash, Enumerable, Enumerator, Range, PEnumType
+            true
+          when Integer
+            o > 0
```
hlindberg: Shouldn't this be `>= 0` ? 0.times is just as valid as iterating over an empty array

lib/puppet/pops/types/types.rb
```diff
@@ -603,6 +671,111 @@ def _assignable?(o)
...
+
+      def initialize(type)
+        @element_type = type
+      end
+
+      def instance?(o)
+        if @element_type.nil? || @element_type.assignable?(PAnyType::DEFAULT)
+          # Any element_type will do
+          case o
+          when Iterable, String, Hash, Enumerable, Enumerator, Range, PEnumType
```
hlindberg: why are Enumerable and Enumerator here? Shouldn't this be Array? We are not really supporting any Ruby type that can be enumerated since we do not know what they return. Although we have the same problem with Hash and Array we have chosen to use the Ruby objects as Puppet runtime objects, so we have to live with those.

lib/puppet/pops/types/types.rb
```diff
@@ -603,6 +671,111 @@ def _assignable?(o)
...
+            true
+          when Integer
+            o > 0
+          when PIntegerType
+            o.enumerable?
```
hlindberg: Why enumerable and not iterable? Is this asking about the Ruby quality enumerable as opposed to the Puppet qulity of being iterable? (Confusing)

lib/puppet/pops/types/types.rb
```diff
@@ -603,6 +671,111 @@ def _assignable?(o)
...
+      DEFAULT = PIterableType.new(nil)
+
+      protected
+
+      # @api private
+      def _assignable?(o)
+        if @element_type.nil? || @element_type.assignable?(PAnyType::DEFAULT)
+          # Don't request the iterable_type since it might be expensive to compute the
+          # type of its elements
+          o.iterable?
```
hlindberg: Hm, don't understand how it can accept any iterable

lib/puppet/pops/types/types.rb
```diff
@@ -603,6 +671,111 @@ def _assignable?(o)
...
+
+      DEFAULT = PIterableType.new(nil)
+
+      protected
+
+      # @api private
+      def _assignable?(o)
+        if @element_type.nil? || @element_type.assignable?(PAnyType::DEFAULT)
+          # Don't request the iterable_type since it might be expensive to compute the
+          # type of its elements
```
hlindberg: I got tricked here. Worth adding this to the comment ", since this Iterable accepts Any element, it is enough that o is iterable"

lib/puppet/pops/types/types.rb
thallgren: See comment above regarding enumerable? versus iterable?

lib/puppet/pops/types/types.rb
```diff
@@ -603,6 +671,111 @@ def _assignable?(o)
...
+    #
+    class PIteratorType < PAnyType
+      attr_reader :element_type
+
+      def initialize(type)
+        @element_type = type
+      end
+
+      def instance?(o)
+        o.is_a?(Iterable) && (@element_type.nil? || @element_type.assignable?(o.element_type))
```
hlindberg: How do you know that the `o` in `o.element_type` has such a member?

lib/puppet/pops/types/types.rb
```diff
@@ -603,6 +671,111 @@ def _assignable?(o)
...
+        self.class == o.class && @element_type == o.element_type
+      end
+
+      DEFAULT = PIteratorType.new(nil)
+
+      protected
+
+      # @api private
+      def _assignable?(o)
+        o.is_a?(PIteratorType) && (@element_type.nil? || @element_type.assignable?(o.element_type))
```
hlindberg: same thing here with `o.element_type`

lib/puppet/pops/types/types.rb
```diff
@@ -621,6 +794,14 @@ def hash
         @size_type.hash * 31 + @values.hash
       end
 
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        PIterableType.new(PStringType::DEFAULT)
```
hlindberg: I don't get this, why are all these not using the information in the type?

lib/puppet/pops/types/types.rb
thallgren: Because `o.is_a?(Iterable)` and `Iterable` defines the method `element_type`.

lib/puppet/pops/types/types.rb
thallgren: Because `o.is_a?(PIteratorType)` and `PIteratorType` defines the method `element_type`

lib/puppet/pops/types/types.rb
```diff
@@ -840,11 +1021,26 @@ def generalize
...
         @elements.hash
       end
 
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        if self == DEFAULT
+          PIterableType.new(PTupleType.new([PAnyType::DEFAULT], PHashType::TUPLE_SIZE))
```
hlindberg: Do not understand, does this create an Iterable[Iterable[Any, what?? ) Think a comment is needed here

lib/puppet/pops/types/types.rb
thallgren: Yes, that would be an improvement. Will fix.

lib/puppet/pops/types/types.rb
thallgren: Agree. This is inconsistent. Will change.

lib/puppet/pops/types/types.rb
hlindberg: Is this for the type `Struct[{}]` - If having one of those it does not really matter what the elementy type is, anything goes. I should use Tuple[Unit, Unit] I think (i.e. "whatever you want the type to be").

lib/puppet/pops/types/types.rb
```diff
@@ -840,11 +1021,26 @@ def generalize
...
+        else
+          tc = TypeCalculator.singleton
+          key_type = tc.infer_and_reduce_type(@elements.map {|se| se.key_type })
```
hlindberg: why reduce and generalize? doesn't that lose information

lib/puppet/pops/types/types.rb
```diff
@@ -840,11 +1021,26 @@ def generalize
...
+          value_type = tc.infer_and_reduce_type(@elements.map {|se| se.value_type })
```
hlindberg: same here

lib/puppet/pops/types/types.rb
hlindberg: Reading again, I think this is for any hash, and then it looks ok.

lib/puppet/pops/types/types.rb
```diff
@@ -983,6 +1179,14 @@ def instance?(o)
         true
       end
 
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        PIterableType.new(TypeCalculator.singleton.infer_and_reduce_type(types))
```
hlindberg: lossy, this is really Iterable[Variant [each element_type in tuple]]

lib/puppet/pops/types/types.rb
```diff
@@ -1263,6 +1467,18 @@ def instance?(o)
...
 
+      def iterable?
+        true
+      end
+
+      def iterable_type
+        if self == DEFAULT || self == EMPTY
+          PIterableType.new(PTupleType.new([PAnyType::DEFAULT], TUPLE_SIZE))
+        else
+          PIterableType.new(PTupleType.new([@key_type, element_type], TUPLE_SIZE))
```
hlindberg: is element_type a typo? or is it element_type() (a call), or should it be @element_type, etc...

lib/puppet/pops/types/types.rb
```diff
@@ -1272,6 +1488,7 @@ def is_the_empty_hash?
       end
 
       DEFAULT = PHashType.new(nil, nil)
+      TUPLE_SIZE = PIntegerType.new(2,2)
```
hlindberg: Don't get the naming here. What TUPLE_SIZE?

lib/puppet/pops/types/types.rb
```diff
@@ -1307,8 +1524,9 @@ class PVariantType < PAnyType
 
       attr_reader :types
 
+      # @param types [Array[PAnyType]] the variants
       def initialize(types)
-        @types = types.freeze
+        @types = types.uniq.freeze
```
hlindberg: Other optimizations here would be to reduce the set based on assignability. e.g:
`Variant[Integer, Float, Any]`, that is just Variant[Any], or infact, Any :-)

lib/puppet/pops/types/types.rb
```diff
@@ -1394,6 +1612,15 @@ def instance?(o)
         assignable?(TypeCalculator.infer(o))
       end
 
+      def iterable?
+        c = class_from_string(@runtime_type_name)
+        c.nil? ? false : c < Enumerable || c < Iterable
+      end
```
hlindberg: think this makes any Ruby Enumerable be Iterable. Don't think we want that. It is too open ended.

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,229 @@
+require 'spec_helper'
+require 'puppet/pops'
+
+module Puppet::Pops::Types
+describe 'The iterable support' do
+
+  [5, (3..10), %w(a b c), 'hello', PIntegerType.new(1, 4), PEnumType.new(%w(yes no)) ].each do |obj|
```
hlindberg: suggest formatting each input case on a separate row - makes it easier to spot the last line
`].each do |obj| ` and to identify each case

hlindberg: shouldn't there be a hash there?

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+    it "should consider instances of #{obj.class.name} to be Iterable" do
+      expect(PIterableType::DEFAULT.instance?(obj)).to eq(true)
+    end
+
+    it "should yield an Iterable instance when Iterable.on is called with a #{obj.class.name}" do
+      expect(Iterable.on(obj)).to be_a(Iterable)
+    end
+  end
+
+  { -1 => 'a negative Integer', 0 => 'the value 0', PIntegerType.new(nil, nil) => 'an unbounded Integer type' }.each_pair do |obj, desc|
```
hlindberg: one line per input case

hlindberg: This list needs a Ruby class that is Eumerable. To test that we do not accept just any Enumerable.

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+    expect(Iterable.on(PIntegerType.new(2,7)).element_type).to eq(PIntegerType.new(2,7))
+  end
+
+  it 'produces an iterable on String' do
+    expect{ |b| Iterable.on('eat this').each(&b) }.to yield_successive_args('e', 'a', 't', ' ', 't', 'h', 'i', 's')
+  end
+
+  it 'produces an iterable with element type String[1,1] for an iterable created on a String' do
+    expect(Iterable.on('eat this').element_type).to eq(PStringType.new(PIntegerType.new(1,1)))
+ end
```
hlindberg: so I was wrong earlier this example is correct; which means I misunderstood what the code was for...

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+
+  it 'produces an iterable on Array' do
+    expect{ |b| Iterable.on([1,5,9]).each(&b) }.to yield_successive_args(1,5,9)
+  end
+
+  it 'produces an iterable with element type inferred from the array elements for an iterable on Array' do
+    expect(Iterable.on([1,5,5,9,9,9]).element_type).to eq(PVariantType.new([PIntegerType.new(1,1), PIntegerType.new(5,5), PIntegerType.new(9,9)]))
+  end
+
+  it 'can chain reverse_each after step on Enumerator' do
```
hlindberg: Enumerator?

spec/unit/pops/types/iterable_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+  end
+
+  it 'can chain step after reverse_each on Integer range' do
+    expect{ |b| Iterable.on(PIntegerType.new(0, 5)).reverse_each.step(2, &b) }.to yield_successive_args(5,3,1)
+  end
+
+  it 'will produce the same result for each as for reverse_each.reverse_each' do
+    x1 = Iterable.on(5)
+    x2 = Iterable.on(5)
+    expect(x1.inject([]) { |a,i| a << i; a}).to eq(x2.reverse_each.reverse_each.inject([]) { |a,i| a << i; a})
```
hlindberg: always use `reduce` instead of `inject`

spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -1965,20 +1981,20 @@ class Foo
   context 'when asking for an enumerable ' do
```
hlindberg: asking for an iterable

spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -1965,20 +1981,20 @@ class Foo
...
     it 'should produce an enumerable for an Integer range that is not infinite' do
```
hlindberg: enumerable

hlindberg:
I am fine with the general approach now!
I think that I made several comments where I am complaining about type inference being lossy, that may be because I don't understand the exact purpose of methods (I later found examples that showed that I got it wrong in some cases.

I will try to write some example in .pp that should illustrate more advanced type inference.

lib/puppet/pops/types/types.rb
thallgren: Oversight. This should be changed to use the same improved logic that was added to Iterable.

lib/puppet/pops/types/types.rb
thallgren: The size of a TUPLE in a key pair. What other tuples are there :-). Seriously. I'll change it to KEY_PAIR_TUPLE_SIZE.

lib/puppet/pops/types/types.rb
thallgren: Assume you do infer an element type for [0,1,1,1,1,1,1,2,2,2,2,2,2]. Do you really want that to result in a Variant with 13 Integer types? Or is 3 enough?

hlindberg: ok, it is fine to have it here, but it should really be in the others since it seems it would waste cycles first creating types for all of the integers, and then they are made unique. In your example case it could even be optimized into Integer[0,2]. We should perhaps use a Set when creating the set of types for Variant (not sure how wan avoid creating the type to add though, so we still end up with garbage). --- i.e. leave it like you have it in this PR. May want to optimize some of the inferences later.

nfagerlund:
@hlindberg Hey, sorry for the delay. You asked about 4x functions vs. 4+3x for docs — yes, we still need a dummy 3x function for docs. This will have to continue until Puppet Strings can run cleanly against an unmodified checkout of the Puppet source; right now it results in a bit of a mess, so we're unable to rewrite our references tooling to rely on it.

hlindberg:
@nfagerlund are there tickets logged for what is missing? We want to fix those instead of doing the busy work of maintaining two versions (since it is not just a copy paste operation due to different formats)

hlindberg:
The documentation for each of the functions that support Iterable must be changed to reflect this.

hlindberg:
For documentation, I suggest we do not expand on the concept of being Iterable in each of the functions. Only what the rules are for that function. I suggest that we mention all the concrete things that can be iterated over, and say, something like "you can iterate over Array, Hash, ....., i.e. values that are of Iterable type". Later we can update with links to docs site that have the description of the "Iterable concept", and/or adjust the phrasing (Docs needs to be involved then). 

The functions must describe that "When the function is called without a block it will return a value of  Iterable[T] type that can be passed around as a first order data value, and later be iterated on by giving it as an argument to a function that accepts an Iterable[T]. Some returned Iterable[T] values will be the original LHS value, others will return an Iterable value that is of type Iterator[T]". Again, leave out the long explanation, what I put here in quotes should be enough (or something to that effect).

The documentation should also mention that the "snaking" of values is lazy as this is important for the order of evaluation; you may otherwise believe that a map followed by reduce will first call the map block n times, then the reduce block n times. I.e. "The block for each chained iteration function is called in succession for each value when it is consumed at the end of the chain".

The documentation has to show what happens when iterating over Hashes, What the rules are wrt tuples/hash.

I think there should be one example in each that shows iterable without block. Something simple like
$x = 3.each()
notice $x.map |$x| { $x * 10 }
# will notice [0, 10, 20]

Where the example uses the function in question on the first row

thallgren:
The functions cannot be called without a block though. This has not changed. Only new functions like `step` and `reverse_each` (in separate PR's) can do that and it's mentioned in their documentation.

In essence, nothing has changed in the behavior of the current iterative functions besides a more stringent check that the argument is an instance of the Iterable type rather than an instance of Any. They all return exactly the same result as before (none of them returns a transformed `Iterable`).

thallgren:
I would suggest that instead of repeating the long list of concrete `Iterable` objects in each function, we should say something short like "Array, Hash, or other Iterable object". It will be very cumbersome otherwise, to update each and every function when adding a new object that one can iterate over (and to remember that it must be done). It's also very confusing to mention that a String and an Integer are iterable without further explaining _how_ they are iterable. IMO, we must either leave them out and mention the Iterable as a concept explained elsewhere, or explain that concept it in every function.

hlindberg:
ok, clear on documentation - needs to be short, it snowballs otherwise.

nfagerlund:
@hlindberg https://tickets.puppetlabs.com/browse/PDOC-59 is the epic. I don't think I know about all the problems Strings runs into when you point it at core, but I captured several.

hlindberg:
Thanks @nfagerlund, not sure why my radar has not picked up those tickets. Will look into them. We need to fix those (or as you pointed out, fix the documentation comments/style of the code it is documenting if it does not follow the "standard way").

hlindberg:
Thanks @nfagerlund PDOC tickets were missing from our radar screen. We now see them and will try to do something about them asap.

PR #4579, opened 2016-01-19
(PUP-5538) Fix SID::Principal domain_account

Iristyle:
- Previously, the Puppet::Util::Windows::SID::Principal initializer
   didn't understand that an account may be passed in with a domain
   prefix that should be parsed off when building the domain_account
   value.

   When calling:

   Puppet::Util::Windows::SID::Principal.lookup_account_name('NT AUTHORITY\\SYSTEM')

   account was previously set to 'NT AUTHORITY\\SYSTEM' and
   domain_account to 'NT AUTHORITY\\NT AUTHORITY\\SYSTEM'

   This commit changes account to 'SYSTEM' and
   domain_account to 'NT AUTHORITY\\SYSTEM' in the above instance


glennsarti:
+1

ferventcoder:
:+1:

PR #4580, opened 2016-01-19
(PUP-5713) Let ClassInformationService use source of parameter defaults

thallgren:
A parameter default that is not a literal must be represented as
source in the info structure created by the ClassInformationService.
A bug in the SourcePosAdapter where all contained elements were
assumed to have a non-nil offset prevented this function.

This commit fixes the bug and adds a test to verify that the the
source is extracted correctly.

The commit also changes the behavior so that the 'default_source'
entry will always be present in the info structure when a default
expression is present, regardless of if it also contains a
'default_literal' entry.

### DIFF
--- a/lib/puppet/pops/adapters.rb
+++ b/lib/puppet/pops/adapters.rb
@@ -91,6 +91,7 @@ module Puppet::Pops::Adapters
       last = first + @adapted.length
       @adapted.eAllContents.each do |m|
         m_offset = m.offset
+        next if m_offset.nil?
         first = m_offset if m_offset < first
         m_last = m_offset + m.length
         last = m_last if m_last > last
--- a/lib/puppet/info_service/class_information_service.rb
+++ b/lib/puppet/info_service/class_information_service.rb
@@ -79,10 +79,7 @@ class Puppet::InfoService::ClassInformationService
     value_expr = p.value
     return structure if value_expr.nil?
     default_value = value_as_literal(value_expr)
-    if !default_value.nil?
-      structure[:default_literal] = default_value
-      return structure
-    end
+    structure[:default_literal] = default_value unless default_value.nil?
     structure[:default_source] = extract_value_source(value_expr)
     structure
   end
--- a/spec/unit/info_service_spec.rb
+++ b/spec/unit/info_service_spec.rb
@@ -19,6 +19,9 @@ describe "Puppet::InfoService" do
            class bar($bar_a, Integer $bar_b, String $bar_c = 'c default value') { }
            class bar2($bar2_a, Integer $bar2_b, String $bar2_c = 'c default value') { }
         CODE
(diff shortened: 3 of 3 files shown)


hlindberg:
We cannot merge this yet as 4.3.2 release work is pending.

lib/puppet/pops/adapters.rb
```diff
@@ -91,6 +91,7 @@ def extract_tree_text
       last = first + @adapted.length
       @adapted.eAllContents.each do |m|
         m_offset = m.offset
+        next if m_offset.nil?
```
hlindberg: Not a good sign that some expressions lack offset. If it is last in a sequence, part of the source will be missing. Think we need to figure out what causes it to be nil. and see if we can set the offset and length. OTOH, it may be only special nodes where the structure is known to continue and skipping will always be correct.

thallgren: The culprit in this case is the TextExpression representing $::os_family.

camlow325:
I can't really speak to the technical merits of this PR but I have observed that @thallgren's one line check for nil fix seemed to address all of the instances of the "undefined method `<' for nil:NilClass" errors that I'd previously been seeing with different flavors of expressions embedded in the default parameter values for classes.  I've been playing around with generating catalogs from our internal Ops Puppet code and where before, I would see several hundred instances of this error when compiling a catalog for a single environment, I now don't seem to be seeing any of these failures.

lib/puppet/pops/adapters.rb
thallgren: Unless this check is present, we should have a unit-test that verifies that all conceivable expressions gets the offset and length set. Do we have a test somewhere that tests all types of expressions for some other reason?

hlindberg: We do not have tests that all expressions gets offset and length set correctly. Needless to say that would be a substantial task as source snippets covering all possible paths through the grammar would need to be covered. You almost would need to generate them from the grammar itself to be certain. 
Coming back to this PR, there will be a problem unless there always is an empty string expression at the end of the interpolation. I can see either fixing the missing offset in this ticket, or creating a new ticket for that problem. (Should not be hard to fix, probably just one call to locate in the grammar).

thallgren: The expression here is "default with interpolated $::os_family" which means that the culprit TextExpression is as last as it can be. The text is still extracted correctly. I also tested with just "$::os_family". It still works. The reason is does is that offset and length of the container `ConcatenatedString` already defines the boundaries.

So unless a `TextExpression` can exist somewhere outside of a `ConcatenatingString` then fixing it doesn't give us anything besides being able to remove the check for nil. And that check cannot be removed anyway without investing in the aforementioned substantial task. So I suggest we leave things as they are.

PR #4582, opened 2016-01-20
WIP (PUP-5609) Protect shared state with filesystem lock

MikaelSmith:
by apply and agent). Exits with error code 1 if lock is encountered.

### DIFF
--- a/spec/unit/application/apply_spec.rb
+++ b/spec/unit/application/apply_spec.rb
@@ -9,6 +9,9 @@ require 'fileutils'
 describe Puppet::Application::Apply do
   before :each do
     @apply = Puppet::Application[:apply]
+    # So we don't actually try to hit the filesystem.
+    @apply.stubs(:lock).yields
+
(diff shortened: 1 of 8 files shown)


spec/unit/application/apply_spec.rb
```diff
@@ -171,6 +178,30 @@
...
+
+      it "should exit" do
+        expect(@apply.run_command).to raise_error(SystemExit)
+      end
+
+      it "should return an exit code of 1" do
+        begin
+          @apply.run_command
+        rescue SystemExit => e
+          expect(e.status).to eq(0)
```
MikaelSmith: This test doesn't appear to work, as it should fail.

MikaelSmith:
Still need to figure out why the acceptance test fails. /cc @dalen

dalen:
I fixed the exit status and tests for that in https://github.com/puppetlabs/puppet/pull/4584

PR #4584, opened 2016-01-20
(PUP-5609) Protect shared state with filesystem lock

dalen:
by apply and agent).

This time it exits with status 1 if lock is already taken and has some tests for that from @MikaelSmith (from https://github.com/puppetlabs/puppet/pull/4582).

### DIFF
--- a/spec/unit/application/apply_spec.rb
+++ b/spec/unit/application/apply_spec.rb
@@ -9,6 +9,9 @@ require 'fileutils'
 describe Puppet::Application::Apply do
   before :each do
     @apply = Puppet::Application[:apply]
+    # So we don't actually try to hit the filesystem.
+    @apply.stubs(:lock).yields
+
     Puppet::Util::Log.stubs(:newdestination)
     Puppet[:reports] = "none"
   end
@@ -19,6 +22,10 @@ describe Puppet::Application::Apply do
 
     Puppet::Node.indirection.reset_terminus_class
     Puppet::Node.indirection.cache_class = nil
+  end
+
+  it "should include the Locker module" do
+    expect(Puppet::Application::Apply.ancestors).to be_include(Puppet::Util::Locker)
   end
 
   [:debug,:loadclasses,:test,:verbose,:use_nodes,:detailed_exitcodes,:catalog, :write_catalog_summary].each do |option|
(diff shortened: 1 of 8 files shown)


MikaelSmith:
I'm still not sure why `tests/resource/service/puppet_mcollective_service_management.rb` fails. That needs some more investigation.

MikaelSmith:
The acceptance test appears to run into a lag between unlinking the lockfile and exiting puppet, and when the file is released. I have to introduce a slight lag (200-300ms) between `puppet apply` runs. I want to investigate whether `puppet agent` has the same problem. This is irritating, but may not be an issue in practice.

MikaelSmith:
My feeling is that it would be better to hold off merging until we also introduce a wait on the lockfile, and should default it to 0.5s.

MikaelSmith:
https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/agent/agent_disable_lockfile.rb covers acceptance testing the agent lockfile. It doesn't appear to test that you can do back-to-back runs.

MikaelSmith:
@parisiale did you run into anything like this running on pxp-agent? Maybe testing service stop/start?

spec/unit/application/apply_spec.rb
```diff
@@ -171,6 +178,30 @@
...
+    end
+
+    describe "when a filesystem lock is encountered" do
+      before :each do
+        @apply.expects(:lock).raises(Puppet::LockError, 'locked')
+        Puppet.expects(:notice).with(regexp_matches /Run of Puppet already in progress; skipping  (.+ exists)/)
+      end
+
+      it "should exit" do
+        expect { @apply.run_command }.to raise_error(SystemExit)
```
MikaelSmith: Found out from @joshcooper that there's a helper for this, https://github.com/puppetlabs/puppet/blob/master/spec/integration/application/apply_spec.rb#L38, so we can drop lines 192-201 down to one spec test.

peterhuene: Hi @dalen. Did you see @MikaelSmith's comment above?

peterhuene:
Blocked on sprint triage.

kylog:
@MikaelSmith what do you think next steps are on this one?

MikaelSmith:
Investigate if there's something broken in the lock implementation ([PUP-3914](https://tickets.puppetlabs.com/browse/PUP-3914) comes up as a possible issue). I also left next steps at https://github.com/puppetlabs/puppet/pull/4584#issuecomment-173342757.

hlindberg:
A thought, what happens when you debug using apply, and quit in the middle. Does that clean up and release the lock?

MikaelSmith:
It's supposed to be cleaned up on the next run, when Puppet detects the old process is no longer running.

spec/unit/application/apply_spec.rb
dalen: Better late than never. But I've fixed that now.

dalen:
Rebased this against master branch

MikaelSmith:
This appears to pass acceptance tests for me, now.

PR #4585, opened 2016-01-21
(PUP-5728) Always load manifests as UTF-8

Iristyle:
- In PUP-2564, documentation was changed to indicate that all manifest
   are expected to be in UTF-8 as part of Puppet 4.0.

   However, this is not actually enforced, and causes problems on Windows.

There are more details to be sorted out on this, and I haven't run tests fully yet - but this is a good starting point for looking at this problem.

### DIFF
--- a/lib/puppet/file_system.rb
+++ b/lib/puppet/file_system.rb
@@ -122,8 +122,8 @@ module Puppet::FileSystem
   #
   # @api public
   #
-  def self.read(path)
-    @impl.read(assert_path(path))
+  def self.read(path, opts = {})
+    @impl.read(assert_path(path), opts)
   end
 
   # Read a file keeping the original line endings intact. This
--- a/lib/puppet/file_system/memory_impl.rb
+++ b/lib/puppet/file_system/memory_impl.rb
@@ -39,7 +39,7 @@ class Puppet::FileSystem::MemoryImpl
     object.path
   end
 
-  def read(path)
+  def read(path, opts = {})
     handle = assert_path(path).handle
     handle.read
   end
--- a/lib/puppet/pops/parser/lexer2.rb
+++ b/lib/puppet/pops/parser/lexer2.rb
@@ -640,7 +640,7 @@ class Puppet::Pops::Parser::Lexer2
   #
   def lex_file(file)
     initvars
-    contents = Puppet::FileSystem.exist?(file) ? Puppet::FileSystem.read(file) : ''
+    contents = Puppet::FileSystem.exist?(file) ? Puppet::FileSystem.read(file, :encoding => 'utf-8') : ''
     @scanner = StringScanner.new(contents.freeze)
     @locator = Puppet::Pops::Parser::Locator.locator(contents, file)
   end
--- a/spec/unit/face/parser_spec.rb
+++ b/spec/unit/face/parser_spec.rb
@@ -98,6 +98,16 @@ describe Puppet::Face[:parser, :current] do
       expect(@logs[0].message).to eq("Syntax error at end of file")
       expect(@logs[0].level).to eq(:err)
     end
+
+    it "logs an error if the input begins with a UTF-8 BOM (Byte Order Mark)" do
+      utf8_bom_manifest = file_containing('utf8_bom.pp', "\uFEFFnotice hi")
+
+      output = parser.dump(utf8_bom_manifest)
+
+      expect(output).to eq("")
+      expect(@logs[1].message).to eq("Syntax error at '' at line 1:1")
+      expect(@logs[1].level).to eq(:err)
+    end
   end
 
   def from_an_interactive_terminal
--- a/spec/unit/file_system_spec.rb
+++ b/spec/unit/file_system_spec.rb
@@ -134,6 +134,32 @@ describe "Puppet::FileSystem" do
     end
   end
 
+  context "read should allow an encoding to be specified" do
+    # First line of Rune version of Rune poem at http://www.columbia.edu/~fdc/utf8/
+    # characters chosen since they will not parse on Windows with codepage 437 or 1252
+    # Section 3.2.1.3 of Ruby spec guarantees that \u strings are encoded as UTF-8
+    let (:rune_utf8) { "\u16A0\u16C7\u16BB" } # 'ᚠᛇᚻ'
+
+    it "and should read a UTF8 file properly" do
+      temp_file = file_containing('utf8.txt', rune_utf8)
+
+      contents = Puppet::FileSystem.read(temp_file, :encoding => 'utf-8')
+
+      expect(contents.encoding).to eq(Encoding::UTF_8)
+      expect(contents).to eq(rune_utf8)
+    end
+
+    it "does not strip the UTF8 BOM (Byte Order Mark) if present in a file" do
+      bom = "\uFEFF"
+
+      temp_file = file_containing('utf8bom.txt', "#{bom}#{rune_utf8}")
+      contents = Puppet::FileSystem.read(temp_file, :encoding => 'utf-8')
+
+      expect(contents.encoding).to eq(Encoding::UTF_8)
+      expect(contents).to eq("#{bom}#{rune_utf8}")
+    end
+  end
+
   describe "symlink",
     :if => ! Puppet.features.manages_symlinks? &&
     Puppet.features.microsoft_windows? do
--- a/spec/unit/pops/loaders/dependency_loader_spec.rb
+++ b/spec/unit/pops/loaders/dependency_loader_spec.rb
@@ -53,6 +53,59 @@ describe 'dependency loader' do
       expect(function.class.name).to eq('testmodule::foo')
       expect(function.is_a?(Puppet::Functions::Function)).to eq(true)
     end
+
+    describe "when parsing files from disk" do
+
+      # First line of Rune version of Rune poem at http://www.columbia.edu/~fdc/utf8/
+      # characters chosen since they will not parse on Windows with codepage 437 or 1252
+      # Section 3.2.1.3 of Ruby spec guarantees that \u strings are encoded as UTF-8
+      let (:node) { Puppet::Node.new('node') }
+      let (:rune_utf8) { "\u16A0\u16C7\u16BB" } # ᚠᛇᚻ
+      let (:code_utf8) do <<-CODE
+Puppet::Functions.create_function('testmodule::foo') {
+  def foo
+    return \"#{rune_utf8}\"
+  end
+}
+      CODE
+      end
+
+      context 'when loading files from disk' do
+        it 'should always read files as UTF-8' do
+          if Puppet.features.microsoft_windows? && Encoding.default_external == Encoding::UTF_8
+            raise 'This test must be run in a codepage other than 65001 to validate behavior'
+          end
+
+          module_dir = dir_containing('testmodule', {
+          'lib' => { 'puppet' => { 'functions' => { 'testmodule' => {
+            'foo.rb' => code_utf8
+          }}}}})
+
+          loader = loader_for('testmodule', module_dir)
+
+          function = loader.load_typed(typed_name(:function, 'testmodule::foo')).value
+          expect(function.call({})).to eq(rune_utf8)
+        end
+
+        it 'currently ignores the UTF-8 BOM (Byte Order Mark) when loading module files' do
+          bom = "\uFEFF"
+
+          if Puppet.features.microsoft_windows? && Encoding.default_external == Encoding::UTF_8
+            raise 'This test must be run in a codepage other than 65001 to validate behavior'
+          end
+
+          module_dir = dir_containing('testmodule', {
+          'lib' => { 'puppet' => { 'functions' => { 'testmodule' => {
+            'foo.rb' => "#{bom}#{code_utf8}"
+          }}}}})
+
+          loader = loader_for('testmodule', module_dir)
+
+          function = loader.load_typed(typed_name(:function, 'testmodule::foo')).value
+          expect(function.call({})).to eq(rune_utf8)
+        end
+      end
+    end
   end
 
   def loader_for(name, dir)
--- a/spec/unit/pops/parser/lexer2_spec.rb
+++ b/spec/unit/pops/parser/lexer2_spec.rb
@@ -702,4 +702,43 @@ describe 'Lexer2' do
       expect_issue('"asd', Puppet::Pops::Issues::UNCLOSED_QUOTE)
     end
   end
+
+end
+
+describe Puppet::Pops::Parser::Lexer2 do
+
+  include PuppetSpec::Files
+
+  # First line of Rune version of Rune poem at http://www.columbia.edu/~fdc/utf8/
+  # characters chosen since they will not parse on Windows with codepage 437 or 1252
+  # Section 3.2.1.3 of Ruby spec guarantees that \u strings are encoded as UTF-8
+  # ᚠᛇᚻ᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢᚱ᛫ᚠᛁᚱᚪ᛫ᚷᛖᚻᚹᛦᛚᚳᚢᛗ
+  let (:rune_utf8) { "\u16A0\u16C7\u16BB\u16EB\u16D2\u16E6\u16A6\u16EB\u16A0\u16B1\u16A9\u16A0\u16A2\u16B1\u16EB\u16A0\u16C1\u16B1\u16AA\u16EB\u16B7\u16D6\u16BB\u16B9\u16E6\u16DA\u16B3\u16A2\u16D7" }
+
+  context 'when lexing files from disk' do
+    it 'should always read files as UTF-8' do
+      if Puppet.features.microsoft_windows? && Encoding.default_external == Encoding::UTF_8
+        raise 'This test must be run in a codepage other than 65001 to validate behavior'
+      end
+
+      manifest_code = "notify { '#{rune_utf8}': }"
+      manifest = file_containing('manifest.pp', manifest_code)
+      lexed_file = described_class.new.lex_file(manifest)
+
+      expect(lexed_file.string.encoding).to eq(Encoding::UTF_8)
+      expect(lexed_file.string).to eq(manifest_code)
+    end
+
+    it 'currently preserves the UTF-8 BOM (Byte Order Mark) when lexing files' do
+      bom = "\uFEFF"
+
+      manifest_code = "#{bom}notify { '#{rune_utf8}': }"
+      manifest = file_containing('manifest.pp', manifest_code)
+      lexed_file = described_class.new.lex_file(manifest)
+
+      expect(lexed_file.string.encoding).to eq(Encoding::UTF_8)
+      expect(lexed_file.string).to eq(manifest_code)
+    end
+  end
+
 end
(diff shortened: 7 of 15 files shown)


lib/puppet/pops/parser/lexer2.rb
```diff
@@ -640,7 +640,7 @@ def file
   #
   def lex_file(file)
     initvars
-    contents = Puppet::FileSystem.exist?(file) ? Puppet::FileSystem.read(file) : ''
+    contents = Puppet::FileSystem.exist?(file) ? Puppet::FileSystem.read(file, :encoding => 'utf-8') : ''
```
joshcooper: This method tells ruby to interpret the bytes in the file as `UTF-8` regardless of the default external encoding, right? Essentially, read the bytes, don't do any transcoding, and tag the string as UTF-8. If so, :+1:

Iristyle: Indeed it does.

lib/puppet/file_system/memory_impl.rb
```diff
@@ -39,7 +39,7 @@ def path_string(object)
     object.path
   end
 
-  def read(path)
+  def read(path, opts = {})
     handle = assert_path(path).handle
     handle.read
```
Iristyle: Interestingly, [StringIO#read](http://ruby-doc.org/stdlib-1.9.3/libdoc/stringio/rdoc/StringIO.html#method-i-read) doesn't take a hash like the other `read` implementations, even though it's also built on `IO` - https://github.com/ruby/ruby/blob/trunk/ext/stringio/stringio.c#L1276-L1337

Iristyle:
All tests pass against ee042bab784fc601ac47967a4e06e5bf950f62a4 (the one failing check was for commit msg formatting given `(WIP)`)

hlindberg:
I know there is at least one additional place puppet manifests are read; in this method in `module_loaders.rb`. That method is used to read both ruby and puppet source. Unless we want to also load Ruby with UTF-8 rather than native we should check the extension of the effective_path argument, and then provide the encoding option. The method below is currently used for loading functions, but it will be used for loading more things in the future.
```
    def get_contents(effective_path)
      Puppet::FileSystem.read(effective_path)
    end
```
The other places were we feed a String to the lexer are probably only in tests, or internal methods that obtained the code from parsed manifests, or where we know for sure the strings only contain ascii.

Iristyle:
@hlindberg if you could spot check the spec changes in a764725, that would be great

spec/unit/file_system_spec.rb
```diff
@@ -134,6 +134,36 @@ def increment_counter_in_multiple_processes(file, num_procs, options)
     end
   end
 
+  context "read should allow an encoding to be specified" do
+    # First line of Rune version of Rune poem at http://www.columbia.edu/~fdc/utf8/
+    # characters chosen since they will not parse on Windows with codepage 437 or 1252
+    let (:rune_utf8) { "\u16A0\u16C7\u16BB" } # 'ᚠᛇᚻ'
```
joshcooper: This assumes that ruby will create a utf8 literal string, but I think that depends how ruby is executed (and potentially the magic encoding comment at the top of this file). It might be safer to just say `{  "\u16A0\u16C7\u16BB".encode('utf-8') }` so you always know what you have, and then remove the `before :each` below.

spec/unit/file_system_spec.rb
```diff
@@ -134,6 +134,36 @@ def increment_counter_in_multiple_processes(file, num_procs, options)
...
+
+    before :each do
+      # verify Ruby has treated string as UTF-8
+      expect(rune_utf8.encoding).to eq(Encoding::UTF_8)
```
joshcooper: general comment, I try to avoid putting expectations in `before :each`, and friends

spec/unit/pops/parser/lexer2_spec.rb
```diff
@@ -702,4 +702,47 @@ def expect_issue(code, issue)
...
+end
+
+describe Puppet::Pops::Parser::Lexer2 do
+
+  include PuppetSpec::Files
+
+  # First line of Rune version of Rune poem at http://www.columbia.edu/~fdc/utf8/
+  # characters chosen since they will not parse on Windows with codepage 437 or 1252
+  # ᚠᛇᚻ᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢᚱ᛫ᚠᛁᚱᚪ᛫ᚷᛖᚻᚹᛦᛚᚳᚢᛗ
+  let (:rune_utf8) { "\u16A0\u16C7\u16BB\u16EB\u16D2\u16E6\u16A6\u16EB\u16A0\u16B1\u16A9\u16A0\u16A2\u16B1\u16EB\u16A0\u16C1\u16B1\u16AA\u16EB\u16B7\u16D6\u16BB\u16B9\u16E6\u16DA\u16B3\u16A2\u16D7" }
```
joshcooper: same comment about making this assumption explicit via `.encode('utf-8')`

spec/unit/file_system_spec.rb
Iristyle: Yeah, this was the pre-flight check to ensure we were in the right encoding, but I'll switch over to using raw bytes and `pack('c*')` like I have in the past if you don't like this style ;0

spec/unit/pops/parser/lexer2_spec.rb
```diff
@@ -702,4 +702,47 @@ def expect_issue(code, issue)
...
+
+end
+
+describe Puppet::Pops::Parser::Lexer2 do
+
+  include PuppetSpec::Files
+
+  # First line of Rune version of Rune poem at http://www.columbia.edu/~fdc/utf8/
+  # characters chosen since they will not parse on Windows with codepage 437 or 1252
+  # ᚠᛇᚻ᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢᚱ᛫ᚠᛁᚱᚪ᛫ᚷᛖᚻᚹᛦᛚᚳᚢᛗ
```
hlindberg: wut? Ironically the encoding is wrong :-)

spec/unit/pops/parser/lexer2_spec.rb
```diff
@@ -702,4 +702,47 @@ def expect_issue(code, issue)
...
+
+      manifest_code = "notify {'#{rune_utf8}'] }"
+      manifest = file_containing('manifest.pp', manifest_code)
+      lexed_file = described_class.new.lex_file(manifest)
+
+      expect(lexed_file.string.encoding).to eq(Encoding::UTF_8)
+      expect(lexed_file.string).to eq(manifest_code)
+    end
+
+    it 'currently preserves the UTF-8 BOM when lexing files' do
```
hlindberg: Spell out BOM (Byte Order Mark) - I did not know what it was until I asked.

spec/unit/face/parser_spec.rb
```diff
@@ -98,6 +98,16 @@
       expect(@logs[0].message).to eq("Syntax error at end of file")
       expect(@logs[0].level).to eq(:err)
     end
+
+    it "logs an error if the input begins with a UTF-8 BOM" do
```
hlindberg: spell out "BOM (Byte Order Mark)"

spec/unit/face/parser_spec.rb
```diff
@@ -98,6 +98,16 @@
...
+      utf8_bom_manifest = file_containing('utf8_bom.pp', "\uFEFFnotice hi")
+
+      output = parser.dump(utf8_bom_manifest)
+
+      expect(output).to eq("")
+      expect(@logs[1].message).to eq("Syntax error at '' at line 1:1")
```
hlindberg: we should be able to do better than that - the lexer could raise a lexing error instead when it sees FEFF or FFFE as the first two bytes.

spec/unit/file_system_spec.rb
joshcooper: I was thinking that a ruby magic comment could cause ruby to use a different encoding for that string, but it doesn't appear to be the case:

```ruby
# Encoding: cp1252
puts "ᚠᛇᚻ".encoding
puts "\u16A0\u16C7\u16BB".encoding
```

Results in:

```
C:\Program Files\Puppet Labs\Puppet> chcp 1252
Active code page: 1252
C:\Program Files\Puppet Labs\Puppet> ruby C:\string.rb
Windows-1252
UTF-8
```

The thing I don't know is whether it guaranteed that the "\u..." string will internally be represented as UTF-8? For example, if I specify the default internal encoding, the "\u..." string is still `UTF-8`:

```
C:\Program Files\Puppet Labs\Puppet> ruby -E cp1252 C:\string.rb
Windows-1252
UTF-8
```

So long story short, I think what you have is fine.

Iristyle: :+1: 

And to add to this, the `\u` is the escape for Unicode codepoints 0000 through FFFF, and was introduced in 1.9.  Strings using `\u` are always encoded in UTF8 per https://books.google.com/books?id=jcUbTcr5XWwC&pg=PA50&lpg=PA50&dq=ruby+%5Cu&source=bl&ots=fJFkra8zcB&sig=uyxrADkv11eAYEBKjjq6E52FOA8&hl=en&sa=X&ved=0ahUKEwjc5eDhus3KAhUQ6GMKHav8B5QQ6AEITTAH#v=onepage&q=ruby%20%5Cu&f=false

More info at
https://leejava.wordpress.com/2009/03/11/unicode-escape-in-ruby/

joshcooper: Good to know it's part of a spec and not accidental behavior, thanks!

spec/unit/face/parser_spec.rb
Iristyle: Yeah, we can file another ticket for it once this PR is merged

spec/unit/pops/parser/lexer2_spec.rb
Iristyle: Oh GitHub...

spec/unit/pops/parser/lexer2_spec.rb
```diff
@@ -702,4 +702,43 @@ def expect_issue(code, issue)
...
+end
+
+describe Puppet::Pops::Parser::Lexer2 do
+
+  include PuppetSpec::Files
+
+  # First line of Rune version of Rune poem at http://www.columbia.edu/~fdc/utf8/
+  # characters chosen since they will not parse on Windows with codepage 437 or 1252
+  # Section 3.2.1.3 of Ruby spec guarantees that \u strings are encoded as UTF-8
+  # ᚠᛇᚻ᛫ᛒᛦᚦ᛫ᚠᚱᚩᚠᚢᚱ᛫ᚠᛁᚱᚪ᛫ᚷᛖᚻᚹᛦᛚᚳᚢᛗ
```
hlindberg: encoding of the above comment ?!?! I see garbage on the last line

Iristyle: This is probably browser failing at rendering... if your font doesn't have Rune characters, you'll be out of luck (but they're good in an editor).

spec/unit/pops/loaders/dependency_loader_spec.rb
```diff
@@ -53,6 +53,52 @@
...
+
+          module_dir = dir_containing('testmodule', {
+          'lib' => { 'puppet' => { 'functions' => { 'testmodule' => {
+            'foo.rb' => "#{bom}Puppet::Functions.create_function('testmodule::foo') { def foo; return \"#{rune_utf8}\"; end; }"
+          }}}}})
+
+          loader = loader_for('testmodule', module_dir)
+
+          function = loader.load_typed(typed_name(:function, 'testmodule::foo')).value
+          expect(function.call({})).to eq(rune_utf8)
```
Iristyle: I wasn't expecting that the file with BOM would work, but Rubys eval is responsible for parsing it, and actually ignores it:

```
[1] pry(main)> bom = "\uFEFF"
=> "\uFEFF"
[4] pry(main)> code = 'puts "hello"'
=> "puts \"hello\""
[5] pry(main)> eval(code)
hello
=> nil
[6] pry(main)> eval(bom + code)
hello
=> nil
[7] pry(main)> eval(bom)
=> nil
```

Iristyle:
@puppetlabs/windows this is ready for review / merge, pending AppVeyor and Travis

spec/unit/face/parser_spec.rb
Iristyle: [PUP-5819](https://tickets.puppetlabs.com/browse/PUP-5819) filed

spec/unit/pops/loaders/dependency_loader_spec.rb
```diff
@@ -53,6 +53,52 @@
...
+          bom = "\uFEFF"
+
+          if Puppet.features.microsoft_windows? && Encoding.default_external == Encoding::UTF_8
+            raise 'This test must be run in a codepage other than 65001 to validate behavior'
+          end
+
+          module_dir = dir_containing('testmodule', {
+          'lib' => { 'puppet' => { 'functions' => { 'testmodule' => {
+            'foo.rb' => "#{bom}Puppet::Functions.create_function('testmodule::foo') { def foo; return \"#{rune_utf8}\"; end; }"
+          }}}}})
```
hlindberg: Picky:
Long line and too much going on, on the same line. Is better to use separate lines and indentation.
However, not worth updating if that is the only thing. If there are other things that needs fixing, then also do this.

spec/unit/pops/parser/lexer2_spec.rb
hlindberg: This has to change - we cannot have comments that are unreadable unless you have a particular font installed.
Remove it.

lib/puppet/file_system.rb
```diff
@@ -122,8 +122,8 @@ def self.each_line(path, &block)
   #
   # @api public
   #
-  def self.read(path)
-    @impl.read(assert_path(path))
+  def self.read(path, opts = {})
+    @impl.read(assert_path(path), opts)
```
joshcooper: Should we warn if `:encoding` is not specified? If it's missing, the caller is likely confusing byte vs character streams. And if they really do want a byte stream, then they could specify as `:encoding => ascii_8bit`.

Or we could add a `read_characters` method that requires an `:encoding`?

It's times like these where I miss Java's InputStreams and Readers.

hlindberg:
The comment with characters that requires a special font to be installed must change. I added a picky comment about a long line that could be changed, but isn't required.
Other than that, this is fine to merge.

spec/unit/pops/loaders/dependency_loader_spec.rb
Iristyle: That was copy / pasted from an existing example (aside from the new code).  Will break out the function def though.

lib/puppet/file_system.rb
Iristyle: Agreed that Ruby's abstractions here are very weak compared to .NET as well (which has amazing encoding and locale support).

How would you like to surface a warning... via `Puppet.warning` or are you thinking of some other mechanism?

spec/unit/pops/parser/lexer2_spec.rb
Iristyle: The comments here are just a sanity check copy of the actual Rune characters specified with `\uXXXX` style code points so that if you need to see what they actually look like you can compare.  I don't think there's anything particularly wrong with including this, even if your current app isn't properly rendering them - https://en.wikipedia.org/wiki/Runic_(Unicode_block)#Fonts for which fonts carry the appropriate characters.  Usually when fonts miss characters, applications follow a font substitution protocol...

I specifically chose these rune characters as they cannot easily be converted to a local codepage on Windows. In my experience, I have no issues in editors (note that I use `Source Code Pro` as my editor font which doesn't include these characters) or nearly all browsers:

### Vim on Mac

screen shot 2016-02-05 at 11 37 02 am

### Sublime on Windows

screen shot 2016-02-05 at 11 39 02 am

### Chrome on Windows

screen shot 2016-02-05 at 11 43 11 am

### Safari on Mac

screen shot 2016-02-05 at 11 45 55 am

### FireFox on Mac

screen shot 2016-02-05 at 11 51 50 am


The only place where I'm actually seeing a problem is the Mac version of Chrome, and I don't think it's worth changing this for what appears to be a browser bug.

Ultimately however, the comment isn't strictly necessary, so if you're really hung up on it, it can be removed - but I'd prefer not to.

ferventcoder: I feel less strongly here. I actually prefer to know what the encoded values are and a comment like this helps.

ferventcoder:
@hlindberg why must it change? My comments above, I prefer to see what is being encoded as a comment. It's super helpful when attempting to make changes or alter the tests. I've ran into this before in other parts of Puppet code and the comments saved me all kinds of time versus not having them.

I guess with the link to the url, it's probably fine as long as that Url never goes away.

lib/puppet/file_system.rb
Iristyle: I should also note that I spiked a changeover of the remaining `File.read` calls in a separate branch last week at https://github.com/Iristyle/puppet/commits/ticket/stable/PUP-5728-convert-remaining-File-read-calls-to-Puppet-FileSystem .  I figured these calls weren't crucial to the ticket at hand, but should also be changed as part of another cleanup effort later - that might be the better spot to introduce the warning you're suggesting?

I think I'd also prefer not to add a new `File.read_characters` method - only because it's yet another special Puppet FileSystem behavior that implementers have to remember to use.

Iristyle:
FYI - I got a (nearly) clean run through an acceptance run via https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-puppet-ad-hoc/29/ - there were a couple of server 500s that caused signing failures, but aside from that I believe all the tests needed to pass passed. /cc @glennsarti

joshcooper:
The acceptance failures are due to a puppetserver bug where under jdk8 and apache commons, if a child process, e.g. /bin/true, doesn't drain from stdin, then it raises an exception. The acceptance tests have been updated in puppet. Also it looks like your PR ran against puppetserver from their master branch.

hlindberg:
@ferventcoder Because you can only read what is there if you have the font. Isn't that reason enough? it looks like garbage everywhere else. If the comment states why that it is there, and that you have to have the font to see it (yada, yada) then I may be persuaded :-) Or if there is a table that also shows the codes. (from my understanding they are a bit further down).

If you cannot see what is there, there is a risk that it gets changed inadvertently. Say if a contributor makes a PR and their encoding is wrong, they then mangle the "garbage". You cannot tell when reviewing because it just looks like "garbage" only different "garbage" than what it is supposed to be....

I think that is bad.

Alternatively, if this is important, write it to a separate file with a substantial warning/instruction header.
(And if being more paranoid that I am, also write a test that checks that the file is correct) so there is no chance to f-up without it being noticed).

hlindberg:
I suppose the comment is the rendering of this:

"\u16A0\u16C7\u16BB\u16EB\u16D2\u16E6\u16A6\u16EB\u16A0\u16B1\u16A9\u16A0\u16A2\u16B1\u16EB\u16A0\u16C1\u16B1\u16AA\u16EB\u16B7\u16D6\u16BB\u16B9\u16E6\u16DA\u16B3\u16A2\u16D7"

If you want to see that rendered it is trivial to do that in irb or in the ruby debugger. Which I tried, and it is still garbage, so not just a browser thing - I simply do not have a font that includes those.

So maybe you should add a PNG to the repo with a correct rendition.

glennsarti:
Will lodge a maint PR to address the rune characters in the comments

lib/puppet/file_system.rb
ferventcoder: I had something similar to this once. :) It was in line with PUP-927 that I started looking down that route.

PR #4586, opened 2016-01-21
(PUP-5696) Add content_uri parameter to file type

HAIL9000:
In order to support direct puppet, add a content_uri parameter to
the file resource which the master can use to uniquely identify file
content.

A file resource may never contain content_uri and content, but may
contain content_uri and source.


lib/puppet/type/file/content_uri.rb
```diff
@@ -0,0 +1,42 @@
+require 'net/http'
+require 'uri'
+
+module Puppet
+  Puppet::Type.type(:file).newparam(:content_uri) do
+
+    desc <<-'EOT'
+      A unique identifier for file content, which must be a URI which points
```
MikaelSmith: I'd use "be a URI that points" instead. "which ... which" rubs me the wrong way.

lib/puppet/type/file/content_uri.rb
```diff
@@ -0,0 +1,42 @@
...
+      to a file in a module. This parameter allows the content of a file resource
+      to remain consistent regardless of code updates on the master.
```
MikaelSmith: Also source file updates for things in the puppet file server?

lib/puppet/type/file/content_uri.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+
+      self.fail "Cannot use opaque URLs '#{source}'" unless uri.hierarchical?
+      unless uri.scheme == "puppet"
+        self.fail "Must use URLs of type puppet as content URI"
+      end
+    end
+
+    def get_from_puppet_source(source_or_content, &block)
+      request = Puppet::Indirector::Request.new(:code_content, :find, resource[:content_uri], nil, :environment => resource.catalog.environment_instance)
```
MikaelSmith: I forget how this works. This will error if used against a master without a code_content end-point?

It'd be nice to have some sort of spec test for it.

MikaelSmith:
Failed on Ruby 2.3.0, for some reason.

kylog:
That Travis fail was weird - no logs? - so I kicked it.

lib/puppet/type/file/content_uri.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    validate do |path|
+      begin
+        uri = URI.parse(URI.escape(path))
+      rescue
+        self.fail Puppet::Error, "Could not understand URI #{path}: #{detail}", detail
+      end
+
+      self.fail "Cannot use opaque URLs '#{source}'" unless uri.hierarchical?
+      unless uri.scheme == "puppet"
```
joshcooper: Could we do `if uri.scheme != "puppet"`?

lib/puppet/type/file/content_uri.rb
kylog: Should there be added text like:

```
NOTE: this parameter is generated by puppet internally, but should never be specified in manifests.
```

/cc @joshcooper

joshcooper: > Also source file updates for things in the puppet file server?

Yeah not sure how to describe it. The term `code updates on the master` is a little confusing because it almost sounds like the server's clojure or ruby code. It's more the content of files managed by the puppet fileserver, and our consistency guarantees only apply when the file content is within a per-environment directory (eg `$codedir/environments/<env>`) backed by version control (built-in PE or wired up in FOSS).

joshcooper: :+1: to note that the property should not be specified in manifests.

MikaelSmith: Does the fileserver handle things that aren't in modules? Seems like we could specifically refer to module updates.

joshcooper: For Burnside we're only planning on having it serve "files" directories from modules in an environment. We've talked about using the same `code_content` endpoint for other use cases, but there's some debate about whether to reuse the same endpoint or create a new one.

lib/puppet/type/file/content_uri.rb
joshcooper: The call below to `connection.request_get` will make the `GET` request and should raise a ruby HTTP exception. I assume we'd only make a `code_content` request if we have a `content_uri` parameter in the catalog, which we must have gotten from a new master. There is a chance you could have a cached catalog from a newer master, but point the agent to an older master. But at that point, the agent is going to need a new catalog.

:+1: to adding a spec test using `vcr`. I'd also be curious how the 404 exception propagates out, and at what point is it caught and logged. Presumably, it would cause the `resource` to fail with a meaningful error message. That said there is also a ticket to investigate network compatibility for this sort of thing, so I'd be ok deferring that work until later (and not blocking this PR).

kylog:
Oh cool. The 2.3.0 segfault reproduced at my desk.

lib/puppet/type/file/content_uri.rb
MikaelSmith: If this is a new end-point, we should probably add it to the docs folder as well.

kylog: Do you mean: https://github.com/puppetlabs/puppet/tree/master/api?

Yeah, that makes *some* sense. It's a little tricky since the endpoint isn't going to be implemented in ruby at all, but it is going to be *used* from ruby.

@KevinCorcoran is the puppet-server repo gonna carry docs for this endpoint?

MikaelSmith: Yup, that's what I was thinking of. We tend to add docs for the request and expected response, which doesn't necessarily mean the end-point needs to be implemented in Puppet.

KevinCorcoran: It certainly could.  My hope is that we document in one of those places, but not both.  Both places make some kind of sense.  The most similar docs currently in puppet-server are for the "admin API" - https://github.com/puppetlabs/puppet-server/tree/master/documentation/admin-api/v1.  Given that the entire implementation is in puppet-server, that seems like a better place for the documentation to leave.  Perhaps @nfagerlund would care to weigh-in.

kylog: Yeah let's definitely document it in one place, not more, not less :)

I'm happy to follow Nick F's guidance on this.

joshcooper: @HAIL9000 I think I misread the ticket you're working on. I was originally thinking of this as being a 1 pt, add the parameter to the file type, as you suggested in hipchat. We can add the method for making the network request in https://tickets.puppetlabs.com/browse/PUP-5114, along with adding vcr related specs and API docs. So how about if we drop the `get_from_puppet_source` method from this PR and pick it up with the later ticket?

HAIL9000: I'm :+1: on this. Do we feel like all the work that needs to be done is clearly reflected in PUP-5114?

KaysLounge: Please delete me out of this 

Sent from Outlook Mobile




On Thu, Jan 21, 2016 at 4:49 PM -0800, "Josh Cooper" <notifications@github.com> wrote:












In lib/puppet/type/file/content_uri.rb:
> +
> +    validate do |path|
> +      begin
> +        uri = URI.parse(URI.escape(path))
> +      rescue
> +        self.fail Puppet::Error, "Could not understand URI #{path}: #{detail}", detail
> +      end
> +
> +      self.fail "Cannot use opaque URLs '#{source}'" unless uri.hierarchical?
> +      unless uri.scheme == "puppet"
> +        self.fail "Must use URLs of type puppet as content URI"
> +      end
> +    end
> +
> +    def get_from_puppet_source(source_or_content, &block)
> +      request = Puppet::Indirector::Request.new(:code_content, :find, resource[:content_uri], nil, :environment => resource.catalog.environment_instance)



@HAIL9000 I think I misread the ticket you're working on. I was originally thinking of this as being a 1 pt, add the parameter to the file type, as you suggested in hipchat. We can add the method for making the network request in https://tickets.puppetlabs.com/browse/PUP-5114, along with adding vcr related specs and API docs. So how about if we drop the get_from_puppet_source method from this PR and pick it up with the later ticket?

joshcooper: Probably not :smile: I'll take a look and update it.

spec/unit/type/file/content_uri_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  include PuppetSpec::Files
+
+  describe "#validate" do
+
+    let(:path) { tmpfile('file_content_uri_validate') }
+    let(:resource) { Puppet::Type.type(:file).new(:path => path) }
+
+    it "should fail if the URI is not a valid URI" do
+      URI.expects(:parse).with('not_a_uri').raises RuntimeError
+      expect(lambda { resource[:content_uri] = 'not_a_uri' }).to raise_error(Puppet::Error)
```
joshcooper: The ruby segfault backtrace includes line 14:

```
c:0035 p:0021 s:0142 e:000140 LAMBDA /home/travis/build/puppetlabs/puppet/spec/unit/type/file/content_uri_spec.rb:14 [FINISH]
```
I'm thinking we should remove the `lambda` and just `expect(resource[:content_uri] = 'not_uri')` directly?

MikaelSmith: You're expecting an exception, I think it might need to be `expect { resource[:content_uri] = 'not_uri' }`. But avoiding the lambda might fix it.

lib/puppet/type/file/content_uri.rb
MikaelSmith: I'd drop this snippet in PUP-5114 to preserve it.

spec/unit/type/file/content_uri_spec.rb
kylog: I tried removing the lambda at my desk and it still segfaults.

HAIL9000:
@MikaelSmith @kylog @joshcooper Addressed some of the comments, let me know what you think of the description now. I still wasn't that sure what would be more clear.

Trying right now to reproduce the seg fault locally so I can see if tweaking those tests will help.

joshcooper:
:+1:

PR #4587, opened 2016-01-21
(PUP-5670) bump facter release tag

McdonaldSeanp:
msi windows builds for puppet need to grab facter from the
github repo, so we need to bump the release tag to the newest
facter before releasing puppet


stahnma:
:+1:

PR #4588, opened 2016-01-21
(PUP-5729) Add iterative reverse_each function

thallgren:
This PR adds the new function `reverse_each(iterable)`. The function takes an iterable argument and an optional block.

If a block is given, it will be called with all elements of the given argument in reverse order.

Without a block, the function will return a new Iterable that will yield each element of the given argument in reverse order.

### DIFF
--- a/lib/puppet/functions/reverse_each.rb
+++ b/lib/puppet/functions/reverse_each.rb
@@ -4,7 +4,7 @@
 #
 # This function takes one to two arguments.
 #
-# 1. An iterable that the function will iterate over.
+# 1. An `Iterable` that the function will iterate over.
 # 2. An optional lambda, which the function calls for each element in the first argument. It must
 # request one parameter.
 #
@@ -27,20 +27,20 @@
 # When no second argument is present, Puppet returns an `Iterable` that represents the reverse
 # order of its first argument. This allows methods on `Iterable` to be chained.
 #
-# When a lamdba is given as the second argument, Puppet iterates the first argument in reverse order
-# and passes each value in turn to the lambda, then returns the first argument unchanged.
+# When a lamdba is given as the second argument, Puppet iterates the first argument in reverse
+# order and passes each value in turn to the lambda, then returns the first argument unchanged.
 #
 # @example Using the `reverse_each` function with an array and a one-parameter lambda
 #
 # ~~~ puppet
-# # Puppet will print a notice for each of the three items in $data in reverse order.
+# # Puppet will log a notice for each of the three items in $data in reverse order.
 # $data = [1,2,3]
 # $data.reverse_each |$item| { notice($item) }
 #
 # ~~~
 #
-# When no second argument is present, Puppet returns a new `Iterable` which allows it to be directly chained into
-# another function that takes an `Iterable` as an argument.
+# When no second argument is present, Puppet returns a new `Iterable` which allows it to
+# be directly chained into another function that takes an `Iterable` as an argument.
 #
 # @example Using the `reverse_each` function chained with a `map` function.
 #
--- a/lib/puppet/functions/reverse_each.rb
+++ b/lib/puppet/functions/reverse_each.rb
@@ -24,8 +24,8 @@
 #
 # `$reverse_data = reverse_each($data)`
 #
-# When no second argument is present, Puppet returns an iterable that represents the reverse
-# order of its first argument. This allows methods on iterables to be chained.
+# When no second argument is present, Puppet returns an `Iterable` that represents the reverse
+# order of its first argument. This allows methods on `Iterable` to be chained.
 #
 # When a lamdba is given as the second argument, Puppet iterates the first argument in reverse order
 # and passes each value in turn to the lambda, then returns the first argument unchanged.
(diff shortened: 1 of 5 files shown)


lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,88 @@
...
+# and passes each value in turn to the lambda, then returns the first argument unchanged.
+#
+# @example Using the `reverse_each` function with an array and a one-parameter lambda
+#
+# ~~~ puppet
+# # For the array $data, run a lambda that creates a resource in reverse order for each item in reverse.
+# $data = ["routers", "servers", "workstations"]
+# $data.reverse_each |$item| {
+#  notify { $item:
+#    message => $item
```
hlindberg: people will find this strange, you do not need to specify the message when it is identical to the title. Suggest removing that. i.e. just `notify { $item: }`

hlindberg: Also, adding resources in a particular order is kind of meaningless since they are lazily evaluated. It best if the example shows something where order actually matters.

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,88 @@
...
+#  }
+# }
+# # Puppet creates one resource for each of the three items in $data in reverse order. Each resource is
+# # named after the item's value and uses the item's value in a parameter.
+# ~~~
+#
+# When no second argument is present, Puppet returns a new iterable so that a new function that takes
+# an iterable as an argument can use it as input.
```
hlindberg: ... which allows it to be directly chained into another function that takes an Iterable as an argument.

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,88 @@
...
+# $data.reverse_each |$item| {
+#  notify { $item:
+#    message => $item
+#  }
+# }
+# # Puppet creates one resource for each of the three items in $data in reverse order. Each resource is
+# # named after the item's value and uses the item's value in a parameter.
+# ~~~
+#
+# When no second argument is present, Puppet returns a new iterable so that a new function that takes
```
hlindberg: Iterable (with capital I)

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,88 @@
...
+# an iterable as an argument can use it as input.
+#
+# @example Using the `reverse_each` function chained with a `map` function.
+#
+# # For the array $data, return an array containing each value multiplied by 10 in reverse order
+# $data = [1,2,3]
+# $transformed_data = $data.reverse_each.map |$item| { $item * 10 }
+# # $transformed_data contains [30,20,10]
```
hlindberg: maybe better to say `$transformed_data is set to` (contains sounds like a contains? function; yes I think it was my suggestion...)

lib/puppet/functions/reverse_each.rb
hlindberg: Or just use a toy example with [1,2,3], and leave the rest to the reader.

spec/unit/functions/reverse_each_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    end.not_to raise_error
+  end
+
+  it 'should produce "times" interval of integer in reverse' do
+    expect(eval_and_collect_notices('5.reverse_each |$x| { notice($x) }')).to eq(['4', '3', '2', '1', '0'])
+  end
+
+  it 'should produce range Integer[5,8] in reverse' do
+    expect(eval_and_collect_notices('Integer[5,8].reverse_each |$x| { notice($x) }')).to eq(['8', '7', '6', '5'])
+  end
```
hlindberg: Add test for other iterables, or just the special ones? Enum[a,b,c], Array, Hash

hlindberg:
only minor things to address in documentation and tests - then ready to be merged

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,83 @@
+# Reverses the order of the elements of something that is iterable and optionally runs a
+# [lambda](http://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html) for each
+# element.
+#
+# This function takes one to two arguments.
+#
+# 1. An iterable that the function will iterate over.
```
hlindberg: I missed one lower case iterable here

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,83 @@
...
+# When no second argument is present, Puppet returns an `Iterable` that represents the reverse
+# order of its first argument. This allows methods on `Iterable` to be chained.
+#
+# When a lamdba is given as the second argument, Puppet iterates the first argument in reverse order
+# and passes each value in turn to the lambda, then returns the first argument unchanged.
+#
+# @example Using the `reverse_each` function with an array and a one-parameter lambda
+#
+# ~~~ puppet
+# # Puppet will print a notice for each of the three items in $data in reverse order.
```
hlindberg: pedantic, it does not actually print, notice is a logging function and when running apply it logs to the console; hence it is perceived as "printing". Say either "will notice", or "will log a notice".

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,83 @@
...
+# $data = [1,2,3]
+# $data.reverse_each |$item| { notice($item) }
+#
+# ~~~
+#
+# When no second argument is present, Puppet returns a new `Iterable` which allows it to be directly chained into
```
hlindberg: line too wide?

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,83 @@
...
+# another function that takes an `Iterable` as an argument.
+#
+# @example Using the `reverse_each` function chained with a `map` function.
+#
+# # For the array $data, return an array containing each value multiplied by 10 in reverse order
```
hlindberg: "produce" instead of "return"

lib/puppet/functions/reverse_each.rb
```diff
@@ -0,0 +1,83 @@
...
+# $data = [1,2,3]
+# $transformed_data = $data.reverse_each.map |$item| { $item * 10 }
+# # $transformed_data is set to [30,20,10]
+# ~~~
+#
+# @example The same example using `reverse_each` function chained with a `map` in alternative syntax
+#
+# # For the array $data, return an array containing each value multiplied by 10 in reverse order
```
hlindberg: same here

hlindberg:
Tests are good now. some picky docs things than can be improved to do.

lib/puppet/functions/reverse_each.rb
thallgren: I disagree with this. "return" is more precise than "produces". A method can produce an error and or output that has nothing to do with its returned value. The documentation here is referencing the returned value specifically.

PR #4589, opened 2016-01-21
(PUP-5730) Add iterative step function

thallgren:
This PR adds the new function `step(iterable, n)`. The function takes two arguments, an iterable and an integer, and an optional block.

If a block is given, it will be called with each nth element of the given argument.

Without a block, the function will return a new Iterable that will yield each nth element of the given argument.

### DIFF
--- a/lib/puppet/functions/step.rb
+++ b/lib/puppet/functions/step.rb
@@ -25,16 +25,16 @@
 #
 # `$stepped_data = step($data, <n>)`
 #
-# When no block is given, Puppet returns an iterable that represents every nth element
-# of its first argument. This allows functions on iterables to be chained.
+# When no block is given, Puppet returns an `Iterable` that yields the first element and every nth successor
+# element, from its first argument. This allows functions on iterables to be chained.
 #
-# When a block is given, Puppet iterates the first argument and passes each nth value in turn to
-# the block, then returns the first argument unchanged.
+# When a block is given, Puppet iterates the first argument and calls the block with the first element and then with
+# every nth successor element.
 #
 # @example Using the `step` function with an array, a step factor, and a one-parameter block
 #
 # ~~~ puppet
-# # For the array $data, call a block with each 3rd element of $data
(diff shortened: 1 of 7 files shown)


hlindberg:
only some doc and test things to address - then ready to be merged

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,89 @@
+# Provides stepping with given interval over elements in an iterable and optionally runs a
+# [lambda](http://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html) for each
+# element.
+#
+# This function takes two to three arguments.
+#
+# 1. An iterable that the function will iterate over.
```
hlindberg: lower case iterable

lib/puppet/functions/step.rb
```diff
@@ -0,0 +1,89 @@
...
+# # For the array $data, return an array, set to the first element and each 5th successor element, in reverse
+# # order multiplied by 10
+# $data = Integer[0,20]
+# $transformed_data = $data.step(5).map |$item| { $item * 10 }
+# # $transformed_data contains [0,50,100,150,200]
+# ~~~
+#
+# @example The same example using `step` function chained with a `map` in alternative syntax
+#
+# # For the array $data, return an array containing the first and each 5th successor, in reverse order, multiplied by 10
```
hlindberg: "return" => "produce"

I prefer "produce" unless specifically talking about a function or lambda that returns - here the sentence is general as it says "for the array $data, return".

hlindberg:
only picky doc comments remain before merge

lib/puppet/functions/step.rb
thallgren: I would argue that "return" is more precise. A function can "produce" other things than a return value (notices for instance). Also, the term "return" happens to coincide with the @return annotation and is a very well known concept. Why confuse things by using "produce"?

PR #4590, opened 2016-01-21
(PUP-5734) detect malformed metadata dependencies

vicinus:
If the value for the key dependencies in metadata.json of any module isn't an
array no catalog can be generated because the code expects an array and fails
with the following error message if no array is supplied:

undefined method `each' for ...

The patch checks if the value is an array and if not a appropiate error message
is generated.


kylog:
Ping @andersonmills or @adreyer

joshcooper:
@vicinus could you add a test that verifies an exception is raised in this case?

@andersonmills or @adreyer thoughts on this?

andersonmills:
I think this adds better error messaging, but I would also like to see a test. @vicinus

vicinus:
Sorry, for my late reply, work was a little crazy with an upcoming deadline, but I thought I could squeeze writing the test somewhere between, which didn't happen...

But now I have more time again and added a test. Let me know if there are further improvements.

spec/unit/module_spec.rb
```diff
@@ -347,6 +347,18 @@
...
+    it "should throw an error if invalid dependencies are specified" do
+      expect {
+        PuppetSpec::Modules.create(
+          'foobar',
+          @modpath,
+          :metadata => {
+            :dependencies => ""
+          }
+        )
+      }.to raise_error(Puppet::Module::MissingMetadata)
```
andersonmills: Since MissingMetadata is used elsewhere, could you check for the correct occurrence of the error with something like
```
      }.to raise_error(
        Puppet::Module::MissingMetadata, 
        /dependencies in the file metadata.json of the module foobar must be an array, not/)
```

vicinus:
Sorry, but I'm not sure that I understand correctly what you want. Should I shorten the first part of the error message like in your example? Should I change the used Error class, because it's already used elsewhere and don't match exactly (missing vs. invalid)? Or did I misunderstand you completely and you meant something entirely different?

andersonmills:
@vicinus I just wanted the test to check for the text of the message. I was suggesting a regex to check that you get the correct message from the Error that you raise with the code you added to module.rb.

The code change I suggest is only for the test. It was a line note on line 359 of spec/unit/module_spec.rb.

vicinus:
Sorry, it was my bad. I was so focused on the bugfix that I totally forgot about the test. I have included your suggestions and the checks are green, so hopefully this is the last cycle.

andersonmills:
@vicinus Before a merge, I try to always run acceptance tests. I'm doing so now, but the process had changed since I'd done it, so it took a bit to get that up and running. Will let you know once they've run.

andersonmills:
@vicinus It looks like there's some trouble in some branch merging in puppet right now. I'm happy to shepherd the merge from here, so I believe your work is done. I'll get this in when I can, and let you know if there are any actual problems with it.

andersonmills:
:+1: Yay!

PR #4592, opened 2016-01-21
(PUP-5726) Support non-english values in environment variables

glennsarti:
The ENV class in Ruby is not able to interpret non-english values and
can corrupt the environment.  Without these changes, the process
environment for Puppet will corrupt when there are environment variables
with non-english characters and the withenv utility function is used
e.g. the windows and powershell Exec providers.

- The withenv function was extended with an additional parameter of mode.
  The default will use the existing code path, but the windows mode will
  use the Wide String aware windows environment API calls

- The execute_windows function in the Execution module was modified to
  use the windows mode for withenv

- The get_environment_strings and set_environment_variable functions
  were added to the Windows Process module.  These functions use the
  Windows Wide String API calls

- The read_arbitrary_wide_string_up_to function was extended to use a
  specified terminating character sequence.  This was required as the
  memory block from GetEnvironmentStringsW is double null terminated
  whereas the function previously only used a single null terminating
  charater

### DIFF
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -25,21 +25,66 @@ module Util
 
   # Run some code with a specific environment.  Resets the environment back to
   # what it was at the end of the code.
-  def self.withenv(hash)
-    saved = ENV.to_hash
-    hash.each do |name, val|
-      ENV[name.to_s] = val
-    end
-
+  # Windows can store unicode chars in the environment, so need to use API calls
+  #   to get and set the environment instead of ENV due to encoding problems.
+  def self.withenv(hash, mode=:posix)
+    saved = get_environment(mode)
+    merge_environment(hash, mode)
     yield
   ensure
-    ENV.clear
-    saved.each do |name, val|
-      ENV[name] = val
+    if saved
+      clear_environment(mode)
+      merge_environment(saved, mode)
     end
   end
 
+  private
+  # @api private
+  def self.get_environment(mode=:posix)
+    case mode
+      when :posix
+        ENV.to_hash
+      when :windows
+        require 'puppet/util/windows/process'
+        Puppet::Util::Windows::Process.get_environment_strings
+      else
+        raise "Unable to retrieve the environment for mode #{mode}"
+    end
+  end
 
+  # @api private
+  def self.merge_environment(hash, mode=:posix)
+    case mode
+      when :posix
+        hash.each do |name, val|
+          ENV[name.to_s] = val
+        end
+      when :windows
+        require 'puppet/util/windows/process'
+        hash.each do |name, val|
+          name = name.to_s unless name.is_a? String
+          Puppet::Util::Windows::Process.set_environment_variable(name,val)
+        end
+      else
+        raise "Unable to merge the environment for mode #{mode}"
+    end
+  end
+
+  # @api private
+  def self.clear_environment(mode=:posix)
+    case mode
+      when :posix
+        ENV.clear
+      when :windows
+        Puppet::Util::Windows::Process.get_environment_strings.each do |key, val|
+          Puppet::Util::Windows::Process.set_environment_variable(key,nil)
+        end
+      else
+        raise "Unable to clear the environment for mode #{mode}"
+    end
+  end
+
+  public
   # Execute a given chunk of code with a new umask.
   def self.withumask(mask)
     cur = File.umask(mask)
--- a/lib/puppet/util/windows/process.rb
+++ b/lib/puppet/util/windows/process.rb
@@ -224,6 +224,50 @@ module Puppet::Util::Windows::Process
   end
   module_function :windows_major_version
 
+  ENVSTRINGS_TERMINATOR_WCHAR = [0,0]
+  # Returns a hash of the current environment variables encoded as UTF-8
+  # The memory block returned from GetEnvironmentStringsW is double-null terminated and the vars are paired as below;
+  # Var1=Value1\0
+  # Var2=Value2\0
+  # VarX=ValueX\0\0
+  # Note - Some env variable names start with '=' and are safe to ignore
+  # Note - The env_ptr MUST be freed using the FreeEnvironmentStringsW function
+  # Note - There is no technical limitation to the size of the environment block returned.
+  #   However a pracitcal limit of 64K is used as no single environment variable can exceed 32KB
+  def get_environment_strings
+    env_ptr = GetEnvironmentStringsW()
+
+    pairs = env_ptr.read_arbitrary_wide_string_up_to(65534, ENVSTRINGS_TERMINATOR_WCHAR)
+      .split(?\x00)
+      .reject { |env_str| env_str.nil? || env_str[0] == '=' }
+      .map { |env_pair| env_pair.split('=', -2) }
+    Hash[ pairs ]
+  ensure
+    if env_ptr && ! env_ptr.null?
+      if FreeEnvironmentStringsW(env_ptr) == FFI::WIN32_FALSE
+        Puppet.debug "FreeEnvironmentStringsW memory leak"
+      end
+    end
+  end
+  module_function :get_environment_strings
+
+  def set_environment_variable(name,val)
+    FFI::MemoryPointer.from_string_to_wide_string(name) do |name_ptr|
+      if (val.nil?)
+        if SetEnvironmentVariableW(name_ptr, FFI::MemoryPointer::NULL) == FFI::WIN32_FALSE
+          raise Puppet::Util::Windows::Error.new("Failed to remove environment variable: #{name}")
+        end
+      else
+        FFI::MemoryPointer.from_string_to_wide_string(val) do |val_ptr|
+          if SetEnvironmentVariableW(name_ptr, val_ptr) == FFI::WIN32_FALSE
+            raise Puppet::Util::Windows::Error.new("Failed to set environment variable: #{name}")
+          end
+        end
+      end
+    end
+  end
+  module_function :set_environment_variable
+
   # Returns whether or not the OS has the ability to set elevated
   # token information.
   #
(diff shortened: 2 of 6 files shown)


kylog:
Nit: can you rephrase the summary line in the commit message to start with an imperative verb? Thanks!

lib/puppet/util/windows/process.rb
```diff
@@ -224,6 +224,56 @@ def windows_major_version
   end
   module_function :windows_major_version
 
+  ENVSTRINGS_TERMINATOR_WCHAR = [0,0]
+  # Returns a hash of the current environment variables encoded as UTF-8
```
Iristyle: Looks like because we're calling `read_arbitrary_wide_string_up_to`, we're calling `read_wide_string`, which re-encodes the UTF-16LE string as UTF-8.  :+1:

lib/puppet/util/windows/process.rb
```diff
@@ -224,6 +224,56 @@ def windows_major_version
...
+          env_hash.store(pair[0],"")
+        else
+          env_hash.store(pair[0],pair[1])
+        end
+      end
+    end
+    env_hash
+  ensure
+    if (!env_ptr.null?)
+      if FreeEnvironmentStringsW(env_ptr) == FFI::WIN32_FALSE
```
Iristyle: I don't think we need to `raise` here, since there's nothing that can be done.  In similar spots, we just log a message about a potential leak like:

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/api_types.rb#L84-L86

lib/puppet/util.rb
```diff
@@ -25,17 +25,47 @@ module Util
...
   # Run some code with a specific environment.  Resets the environment back to
   # what it was at the end of the code.
-  def self.withenv(hash)
-    saved = ENV.to_hash
-    hash.each do |name, val|
-      ENV[name.to_s] = val
-    end
+  # Windows can store unicode chars in the environment, so need to use API calls
+  #   to get and set the environment instead of ENV due to encoding problems.
+  def self.withenv(hash, mode=:posix)
```
Iristyle: It might be a little cleaner (due to the nested `begin` / `ensure`) to leave the same basic control flow in place for this method, but to add a couple of different `private` helpers for posix vs windows, for saving and restoring the env vars.

lib/puppet/util/windows/process.rb
```diff
@@ -224,6 +224,56 @@ def windows_major_version
...
+        raise Puppet::Util::Windows::Error.new("FreeEnvironmentStringsW failed")
+      end
+    end
+  end
+  module_function :get_environment_strings
+
+  def set_environment_variable(name,val)
+    FFI::MemoryPointer.from_string_to_wide_string(name) do |name_ptr|
+      if (val.nil?)
```
Iristyle: Might want to consider collapsing the logic / branching here... or adding a separate `remove_environment_variable(name)` helper that gets called when `val` is `nil` or `''`

glennsarti: Empty strings ('') are valid environment variable values.  Nil is not, and will remove the variable from the environment

I did originally have a remove_env... method but I chose to keep the functionality of passing nil to remove it.  Firstly, that's how the API is called anyway and having it abstracted didn't seem to save me much.  Also the ruby ENV class (and indeed CMD.EXE and Powershell for that matter) implements removing environment variables in a similar way.

I also did consider modifying the Memory Pointer creator to return NULL pointers if a null string was passed but that seemed to touch too many code paths.

lib/puppet/util/windows/process.rb
glennsarti: Agreed.

lib/puppet/util.rb
glennsarti: I don't agree.  If I replace the single if statement with a single begin-ensure, I end up with 3 if blocks doing different things.  If I move the functionality out to private helpers, I may as well take it further and withenv becomes a single if statement, branching out to private helpers called withenv_posix and withenv_windows and move ALL of the logic into there.  This can be done, as long as the yield statements have the same affect.

Iristyle: Yeah, so what I'm thinking is that we have a few little helpers so that this code can be written as something like this:

```ruby
def self.withenv(hash, mode = :posix)
  saved = get_env(mode)
  set_env(hash, mode)
  yield
ensure
  set_env(saved, mode)
end
```

spec/unit/util_spec.rb
```diff
@@ -24,48 +24,115 @@ def get_mode(file)
...
     end
 
+    describe "windows mode", :if => Puppet.features.microsoft_windows? do
+      before :each do
+        @original_path = ENV["PATH"]
+        @new_env = {:PATH => "/some/bogus/path"}
+      end
+
+      it "should preserve UTF8 environment variables" do
+        env_var_name = SecureRandom.uuid.force_encoding(Encoding::UTF_8)
```
Iristyle: I don't think you need `force_encoding` here:

```
[2] pry(main)> f = SecureRandom.uuid
=> "1aef9610-d9fb-4c06-9627-72ef578efe95"
[3] pry(main)> f.encoding
=> #<Encoding:UTF-8>
```

`force_encoding` is generally something that you *don't* want to call explicitly - it tells Ruby to ignore the current encoding of the string that you have and to pretend that the bytes are representative of another encoding.  In this case, Ruby has generated a string, which by default is `UTF-8`.  (If you really wanted to *change* the encoding of the string to something else, you would call `.encode(Encoding::UTF_8)`)

We do need `force_encoding` a couple of lines later because we have bytes that we know are UTF-8 and that we ask Ruby to pack as a string... but by default, it doesn't know that:

```ruby
[5] pry(main)> s = [0xE2, 0x80, 0x93, 0xE2, 0x84, 0xA2].pack('c*')
=> "\xE2\x80\x93\xE2\x84\xA2"
[6] pry(main)> s.encoding
=> #<Encoding:ASCII-8BIT>

[7] pry(main)> s.force_encoding('utf-8')
=> "\u2013\u2122"
```

lib/puppet/util.rb
```diff
@@ -25,17 +25,47 @@ module Util
...
-    ENV.clear
-    saved.each do |name, val|
-      ENV[name] = val
+    if (mode == :windows)
+      begin
+        require 'puppet/util/windows/process'
+
+        saved = Puppet::Util::Windows::Process.get_environment_strings
+        hash.each do |name, val|
+          # TODO Is it safe to assume a name in this hash is not utf8? Need to use .to_s for symbols
```
Iristyle: I don't understand the comment about needing symbols. Ruby hashes can use plain old strings as keys.

lib/puppet/util/windows/process.rb
```diff
@@ -224,6 +224,58 @@ def windows_major_version
...
+  # Note - The env_ptr MUST be freed using the FreeEnvironmentStringsW function
+  # Note - There is no technical limitation to the size of the environment block returned.
+  #   However a pracitcal limit of 64K is used as no single enviornment variable can exceed 32KB
+  def get_environment_strings
+    env_ptr = GetEnvironmentStringsW()
+
+    env_hash = {}
+    env_ptr.read_arbitrary_wide_string_up_to(65534, ENVSTRINGS_TERMINATOR_WCHAR).split(?\x00).each do |env_pair|
+      pair = env_pair.split("=")
+      unless pair[0] == ""
```
Iristyle: Can shorten this up to:

```ruby
pairs = env_ptr.read_arbitrary_wide_string_up_to(65534, ENVSTRINGS_TERMINATOR_WCHAR)
  .split(?\x00)
  .reject { |env_str| env_str.nil? || env_str[0] == '=' }
  .map { |env_pair| env_pair.split('=', -2) }
Hash[ pairs ]
```

The `-2` in split ensures there's always 2 members in the array after the split, and if it has to fill in a second element, its an empty string.

And of course `Hash[ pairs]` is a little Ruby trick that can initialize a hash with an array of 2 element arrays.

spec/unit/util_spec.rb
glennsarti: Will fix this.

lib/puppet/util/windows/process.rb
glennsarti: Will add this shortened syntax

lib/puppet/util.rb
glennsarti: When passing a symbol, later on in the call stack (wide_string) it raises an error as the Symbol class doesn't have a '+' method.  I'll modify the wide_string function to convert the input as a string if it isn't one.

lib/puppet/util.rb
glennsarti: Agreed.  Will change.

lib/puppet/util.rb
```diff
@@ -25,20 +25,65 @@ module Util
...
+        require 'puppet/util/windows/process'
+        hash.each do |name, val|
+          Puppet::Util::Windows::Process.set_environment_variable(name,val)
+        end
+      else
+        raise "Unable to set the environment for mode #{mode}"
+    end
+  end
+
+  def self.restore_environment(saved, mode=:posix)
```
Iristyle: So I generally like the refactor - makes it much clearer now IMHO.  A couple of minor nits though:

* `set_environment` should probably be renamed to `extend_environment` or similar (yeah, I realized I suggested the original name - but really weren't not setting anything, we're in reality appending, and method name should probably reflect that)
* `restore_environment` is really really close to the current impl of `set_environment` ... but should probably be named `set_environment`, since the goal is that the given hash is becoming the new environment.  That said, I think we can probably reuse the prior method like:

```ruby
def self.set_environment(hash, mode = :posix)
  case mode
  when :posix
    ENV.clear
    extend_environment(hash, mode)
  when :windows
    current = Puppet::Util::Windows::Process.get_environment_strings
    (current.keys - hash.keys).each { |extra_key| hash[extra_key] = nil }
    # extraneous env vars are set to nil, as this indicates they should be unset for Windows
    extend_environment(hash, :windows)
  else
    raise "Unable to set environment for mode #{mode}"
  end
end
```

Iristyle: I'm thinking that `merge_environment!` might be the name to use for the add / append - this is a pretty common name / behavior in Ruby land - i.e. http://ruby-doc.org/core-2.2.0/Hash.html#method-i-merge

Will continue to think about `restore` vs alternatives

Iristyle: Also.. as an FYI, the code oesn't deal with case sensitivity properly.  In the default `ENV` for Windows, `ENV['foo'] == ENV['FoO']` (this is not true for POSIX which allows case sensitive variations on ENV var names).

So we'll need to add some code to deal with that.

lib/puppet/util/windows/string.rb
```diff
@@ -9,6 +9,10 @@ def wide_string(str)
     # causing buffer overruns.
     #
     # See http://svn.ruby-lang.org/cgi-bin/viewvc.cgi?revision=41920&view=revision
+    unless str.is_a? String
```
Iristyle: I figured out why you were dealing with the symbols - for some reason we use symbols as hash keys in the specs and other places (even though it doesn't make a lot of sense TBH).

That can be addressed elsewhere - let's leave this method alone for now.

lib/puppet/util.rb
glennsarti: Added code and a test to not be case sensitive for env. var. names

Iristyle:
merged to master in 7e9a8fcc92c256121f4cb92fd7e2fe528febde96

PR #4593, opened 2016-01-22
(PUP-917) Identify in the report why a cached catalog was use

whopper:
Currently, the agent run report does not provide any information
about why a cached catalog was used. This can cause issues when
a remote catalog request unexpectedly fails, causing the agent
to fall back to a cached catalog. In such a case, the report will
indicate that the run was successful without pointing out that a
cached catalog had to be used.

There are two ways a cached catalog may be used: by running the agent
with the `--use_cached_catalog` option, or by a catalog failing to
compile on the master, causing the agent to automatically fall back to
a cached catalog as long as the `usecacheonfailing` setting is set.

This commit adds the `cached_catalog_status` field to Puppet::Transaction::Report,
with possible values 'unused', 'use_cached_catalog' and
'use_cache_on_failure'.

### DIFF
--- a/api/docs/http_report.md
+++ b/api/docs/http_report.md
@@ -53,7 +53,7 @@ example is formatted for readability)
      "transaction_uuid"=>"df34516e-4050-402d-a166-05b03b940749",
      "code_id"=>null,
      "catalog_uuid"=>"827a74c8-cf98-44da-9ff7-18c5e4bee41e",
-     "report_format"=>4,
+     "report_format"=>5,
      "puppet_version"=>"3.3.0",
      "kind"=>"apply",
      "status"=>"unchanged",
@@ -109,7 +109,8 @@ example is formatted for readability)
          "skipped"=>false,
          "change_count"=>0,
          "out_of_sync_count"=>0,
-         "events"=>[]}}}
+         "events"=>[]}},
+      "cached_catalog_status"=> "unused"}
 
 Schema
 ------
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -103,10 +103,19 @@
             "type":        "string"
         },
 
+        "cached_catalog_status": {
+            "description": "Whether a cached catalog was used, and if so, why it was used. Enumerator with one of the values:\n\n* `unused`, if a cached catalog was not used.\n* `use_cached_catalog`, if a cached catalog was used because the use_cached_catalog option was set.\n* `use_cache_on_failure`, if a cached catalog was used because the usecacheonfailure setting was set and the agent failed to download a new catalog",
+            "enum": [
+              "unused",
+              "use_cached_catalog",
+              "use_cache_on_failure"
+            ]
+        },
+
         "report_format": {
             "description": "The report format version documented by this schema",
             "type":        "integer",
-            "enum":        ["4", 4]
+            "enum":        ["5", 5]
         },
 
         "puppet_version": {
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -47,6 +47,7 @@ class Puppet::Configurer
   def initialize(factory = Puppet::Configurer::DownloaderFactory.new)
     @running = false
     @splayed = false
+    @cached_catalog_status = 'unused'
     @environment = Puppet[:environment]
     @transaction_uuid = SecureRandom.uuid
     @handler = Puppet::Configurer::PluginHandler.new(factory)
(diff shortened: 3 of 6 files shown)


whopper:
Note: this is a work in progress until I get naming feedback from Nick F around what the 'keys' of the three states of cached catalog usage should be.

lib/puppet/configurer.rb
```diff
@@ -55,13 +56,20 @@ def initialize(factory = Puppet::Configurer::DownloaderFactory.new)
   # Get the remote catalog, yo.  Returns nil if no catalog can be found.
   def retrieve_catalog(query_options)
     query_options ||= {}
-    # First try it with no cache, then with the cache.
-    unless (Puppet[:use_cached_catalog] and result = retrieve_catalog_from_cache(query_options)) or result = retrieve_new_catalog(query_options)
-      if ! Puppet[:usecacheonfailure]
-        Puppet.warning "Not using cache on failed catalog"
-        return nil
+    if (Puppet[:use_cached_catalog] && result = retrieve_catalog_from_cache(query_options))
```
whopper: A bit of thinking aloud here, as I'm not 100% happy with what this ended up looking like.

There are a couple of cache-related situations that get resolved in this method. Once it has returned, we can't tell why we used the cache, so the logic of distinguishing between the following items must happen here:

1) `use_cached_catalog` is set *and* there is a catalog in the cache to be successfully retrieved.
  In this case, scenario 1 is true: We're using a cached catalog because it was explicitly requested.
2) `use_cached_catalog` is set *but* there is no cached catalog to be found.
  This is the section scenario, where we *asked* for a cached catalog but will instead have to get a new one.
3) We didn't ask for a cached catalog, and successfully downloaded a new one.
  In this case, no cached catalog is involved at all.
4) We didn't ask for a cached catalog, but remote compilation failed so we fell back using `usecacheonfailure`
  In this case, we're using the cache as a fallback.


Later on in the caller of this method, cases 1 and 4 will become indistinguishable. There is also the edge case of when we need a cached catalog (it was specifically requested or the new catalog failed to compile) but there is no cached catalog available. In this case we need to specify that no cached catalog was used.

Although the keys ('not used', 'used because...') will be condensed into one word values, I'm not sure I like hardcoding them in here like this. Should this logic live elsewhere, or be abstracted into a conditional below when editing the report?

lib/puppet/configurer.rb
```diff
@@ -55,13 +56,20 @@ def initialize(factory = Puppet::Configurer::DownloaderFactory.new)
...
+      @cached_catalog_status = 'used because use_cached_catalog was set'
+    else
+      result = retrieve_new_catalog(query_options)      
+
+      if !result 
+        if ! Puppet[:usecacheonfailure]
```
kylog: Style thing: I'd like to be consistent about whether we have a space after the `!` or not (and the above two lines differ). I'm sure the puppet codebase overall is *super* inconsistent, but at least within a PR or file there's a charm in being consistent.

Fwiw, rubocop prefers no-space after the `!`: https://github.com/bbatsov/rubocop/blob/master/config/enabled.yml#L659-L662

lib/puppet/configurer.rb
```diff
@@ -55,13 +56,20 @@ def initialize(factory = Puppet::Configurer::DownloaderFactory.new)
...
+          Puppet.warning "Not using cache on failed catalog"
+          return nil
+        end
+
+        @cached_catalog_status = 'used because new catalog failed to compile and usecacheonfailure was set'
+        result = retrieve_catalog_from_cache(query_options)
```
kylog: Not sure if we really specified this or not, but should we set `cached_catalog_status` here only if we successfully retrieved a catalog from the cache? I.e. saying "used" if we couldn't use it ... feels wrong.

whopper: Oh good point, I think this is another scenario I missed. Will update.

api/schemas/report.json
```diff
@@ -103,6 +103,15 @@
             "type":        "string"
         },
 
+        "cached_catalog_status": {
+            "description": "Whether a cached catalog was used, and if so, why it was used. Enumerator with one of the values:\n\n",
+            "enum": [
+              "not used",
+              "used because use_cached_catalog was set",
+              "used because new catalog failed to compile and usecacheonfailure was set"
```
whopper: These will soon change to one-word (hopefully) descriptors.

api/docs/http_report.md
```diff
@@ -109,7 +109,8 @@ example is formatted for readability)
          "skipped"=>false,
          "change_count"=>0,
          "out_of_sync_count"=>0,
-         "events"=>[]}}}
+         "events"=>[]}},
```
MikaelSmith: Seems like report_format in the example should bump to 5. Line 56.

whopper: @MikaelSmith oh yeah, missed that one. Updated.

MikaelSmith:
:+1:

PR #4594, opened 2016-01-22
(PUP-5120) Send checksum_type and static_catalog in catalog request

MikaelSmith:
Currently, when the agent applies a catalog, it sends a `file_metadata`
request for each file resource along with the `checksum_type` that the
master should use to compute the checksum. The `checksum_type` originates
in either the manifest (and preserved in the catalog), or if none was
specified, the agent's default `Puppet[:digest_algorithm]`.

When static catalogs are enabled, the file metadata inlining process
will need to choose a digest algorithm if none is specified in the
manifest. To ensure the master chooses a checksum suitable for the
agent, the agent needs to send its desired checksum in the catalog
request as a request parameter. Add a checksum_type query param to the
catalog request with an ordered list of algorithms the agent supports,
with the value of `Puppet[:digest_algorithm]` as the first item.
Providing a list is for future proofing in the case where the agent's
default is not supported by the master.

Also, request a static catalog by sending `static_catalog=true` query
parameter. The server will decide whether to build a static catalog.

### DIFF
--- a/api/docs/http_catalog.md
+++ b/api/docs/http_catalog.md
@@ -40,7 +40,8 @@ doubly-escaped (it should just be singly-escaped).  To keep backward compatibili
 escaping is still used/supported.
 - `transaction_uuid`: a transaction uuid identifying the entire transaction (shows up in the report as well)
 - `static_catalog`: a boolean requesting a static catalog if available; should always be `true`
-- `checksum_type`: a checksum type supported by the agent, for use in file resources of a static catalog.
+- `checksum_type`: a dot-separated list of checksum types supported by the agent, for use in file resources of a static
+catalog. The order signifies preference, highest first.
 
 An optional parameter can be provided to the POST or GET to notify a node classifier that the client requested a specific
 environment, which might differ from what the client believes is its current environment:
@@ -53,7 +54,7 @@ environment, which might differ from what the client believes is its current env
 
     POST /puppet/v3/catalog/elmo.mydomain.com
 
-    environment=env&configured_environment=canary_env&facts_format=pson&facts=%7B%22name%22%3A%22elmo.mydomain.com%22%2C%22values%22%3A%7B%22architecture%22%3A%22x86_64%22%7D&transaction_uuid=aff261a2-1a34-4647-8c20-ff662ec11c4c&static_catalog=true&checksum_type=md5
+    environment=env&configured_environment=canary_env&facts_format=pson&facts=%7B%22name%22%3A%22elmo.mydomain.com%22%2C%22values%22%3A%7B%22architecture%22%3A%22x86_64%22%7D&transaction_uuid=aff261a2-1a34-4647-8c20-ff662ec11c4c&static_catalog=true&checksum_type=md5.sha256
 
     HTTP 200 OK
     Content-Type: text/pson
--- a/api/docs/http_catalog.md
+++ b/api/docs/http_catalog.md
@@ -31,7 +31,7 @@ The examples below use the POST method.
 
 ### Parameters
 
-Four parameters should be provided to the POST or GET:
+Six parameters should be provided to the POST or GET:
 
 - `environment`: the environment name
 - `facts_format`: must be `pson`
@@ -39,6 +39,8 @@ Four parameters should be provided to the POST or GET:
 doubly-escaped (it should just be singly-escaped).  To keep backward compatibility, the extraneous
 escaping is still used/supported.
 - `transaction_uuid`: a transaction uuid identifying the entire transaction (shows up in the report as well)
+- `static_catalog`: a boolean requesting a static catalog if available; should always be `true`
+- `checksum_type`: a checksum type supported by the agent, for use in file resources of a static catalog.
 
 An optional parameter can be provided to the POST or GET to notify a node classifier that the client requested a specific
 environment, which might differ from what the client believes is its current environment:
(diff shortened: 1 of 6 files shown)


MikaelSmith:
Needs new spec tests.

MikaelSmith:
ping @joshcooper, going to add spec tests tomorrow

api/docs/http_catalog.md
```diff
@@ -39,6 +39,8 @@ Four parameters should be provided to the POST or GET:
 doubly-escaped (it should just be singly-escaped).  To keep backward compatibility, the extraneous
 escaping is still used/supported.
 - `transaction_uuid`: a transaction uuid identifying the entire transaction (shows up in the report as well)
+- `static_catalog`: a boolean requesting a static catalog if available; should always be `true`
+- `checksum_type`: a checksum type supported by the agent, for use in file resources of a static catalog. Can be repeated multiple times to list multiple supported types, in order of preference, in case the master doesn't support one or more checksum types.
```
joshcooper: Is it guaranteed that the order of HTTP query parameters will be preserved, e.g. by jetty/trapperkeeper when it parses the HTTP request and hands it off for processing?

kylog: From quick google-fu and a glance at https://tools.ietf.org/html/rfc2616, I couldn't find a statement guaranteeing query parameter order. Maybe it could be specified once as a something-delimited value?

MikaelSmith:
Nevermind, added spec tests.

api/docs/http_catalog.md
MikaelSmith: Hmm, so munge it manually into a string. We could pick just about any separator, the checksum types are all alphanumeric.

lib/puppet/util/checksums.rb
```diff
@@ -6,11 +6,14 @@
...
+  def valid_checksum_types
+    [:sha256, :sha256lite, :md5, :md5lite, :sha1, :sha1lite, :mtime, :ctime]
+  end
+
   # It's not a good idea to use some of these in some contexts: for example, I
   # wouldn't try bucketing a file using the :none checksum type.
   def known_checksum_types
-    [:sha256, :sha256lite, :md5, :md5lite, :sha1, :sha1lite,
-      :mtime, :ctime, :none]
+    self.preferred_checksum_types.push :none
```
joshcooper: I've always wondered what `:none` is for?

MikaelSmith: I think it exists instead of nil. I'm not really sure, except maybe it's required by the indirector so we know it's not just using the default. I'm not sure how you'd ever have :none without an error though.

lib/puppet/configurer.rb
```diff
@@ -49,6 +51,14 @@ def initialize(factory = Puppet::Configurer::DownloaderFactory.new)
     @splayed = false
     @environment = Puppet[:environment]
     @transaction_uuid = SecureRandom.uuid
+    @static_catalog = true
+
+    # Get checksum types known to the agent, preferring the default specified
+    # in the digest_algorithm setting.
+    @checksum_type = valid_checksum_types
```
MikaelSmith: The only valid values for digest_algorithm are md5 and sha256. Does it make sense to send more than that?

joshcooper: :+1: to excluding things other than `md5` and `sha256`.

One problem with sending multiple checksums is we don't have a way for a FIPS-enabled agent to say never use `md5`. For example, with this PR, you could set `Puppet[:digest_algorithm] = 'sha256'` on the agent, but the agent would still send `sha256,md5`, which would be misleading. It seems like we either need to send a single checksum type, or allow the list of checksum types to be configurable.

MikaelSmith: I guess at that point I'd lean towards a single checksum type. If the agent default isn't supported by the master, they can just set `digest_algorithm` on their agents.

MikaelSmith:
Updated to send a single checksum type.

joshcooper:
:+1:

I want to check that an existing 4.3.x master doesn't reject the `catalog` request due to query params it doesn't know about.

joshcooper:
It occurred to me on the way home that we may need the agent to send all supported checksums after all... For example, in the very near future, I want to add support for the `git` checksum type, as described in https://tickets.puppetlabs.com/browse/PUP-5714. This way the master doesn't compute any checksums at all. It just looks up the object id from git, and inlines that in the catalog.

But what happens if the agent only sends one checksum type `git`? If the master doesn't support it, then what does it fall back to? The master has to guess, and we're essentially back to where we are now, not knowing if the agent supports `md5` or not.

If the agent sent a list of supported digest algorithms, e.g. `git,sha256`, then the master could use `git` if it is capable of using it, otherwise fallback to `sha256`, but in no circumstance would the master select `md5` since the agent said it doesn't support it.

So perhaps a new puppet setting `supported_digest_algorithms`?

joshcooper:
I verified that the latest puppetserver nightly build from stable (2.2.2.stable.SNAPSHOT.2016.01.24T0326) ignores the `static_compiler` and `checksum_type` parameters as expected.

MikaelSmith:
Discussion continued at https://tickets.puppetlabs.com/browse/PUP-5120?focusedCommentId=256357&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-256357. Seems feasible, so I'll work up an extra commit to add `supported_digest_algorithms`.

MikaelSmith:
Added a `supported_digest_algorithms` setting.

lib/puppet/defaults.rb
```diff
@@ -786,6 +786,20 @@ def self.default_diffargs
...
                       Valid values are md5, sha256. Default is md5.',
+    },
+    :supported_checksum_types => {
+      :default => ['md5', 'sha256'],
+      :type    => :array,
+      :desc    => 'Checksum types supported by this agent for use in file resources of a static
+        catalog. Valid types are md5, md5lite, sha256, sha256lite, sha1, sha1lite, mtime, ctime.',
+      :hook    => proc do |value|
+        values = munge(value)
+        valid   = ['md5', 'md5lite', 'sha256', 'sha256lite', 'sha1', 'sha1lite', 'mtime', 'ctime']
```
joshcooper: Can the list of valid values be shared with https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/checksum.rb#L12?

MikaelSmith: I'd like to, wasn't sure how to structure it.

joshcooper: We have the same issue with [log levels](https://github.com/MikaelSmith/puppet/blob/PUP-5120/lib/puppet/defaults.rb#L87) so I'm ok not trying to solve that here.

MikaelSmith:
Added new specs and changed to `supported_checksum_types`.

PR #4596, opened 2016-01-22
(PUP-5708) Deprecate pluginsync setting

whopper:
Soon, the behavior of pluginsyncing will be based on whether or
not a cached catalog is being used. This is to prevent scenarios
where the set of plugins used by the agent is different than the
set that was used when the catalog was originally compiled.

This commit deprecates the setting. Future work will handle the
switchover to the new logic based on cached catalog usage.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1570,7 +1570,9 @@ EOT
     :pluginsync => {
       :default    => true,
       :type       => :boolean,
-      :desc       => "Whether plugins should be synced with the central server.",
+      :desc       => "Whether plugins should be synced with the central server. This setting is
+        deprecated.",
+     :deprecated => :completely,
     },
 
     :pluginsignore => {


lib/puppet/defaults.rb
```diff
@@ -1570,7 +1570,11 @@ def self.default_diffargs
     :pluginsync => {
       :default    => true,
       :type       => :boolean,
-      :desc       => "Whether plugins should be synced with the central server.",
+      :desc       => "Whether plugins should be synced with the central server. This setting is deprecated,
+        as the decision to sync plugins with the central server is now based on whether a cached catalog is
+        being applied or not. If a cached catalog is being used, plugins will not be synced to ensure that
+        the same set of plugins are used that existed when the catalog was originally compiled.",
```
MikaelSmith: I'd modify this to say `If a cached catalog is being used, this setting will be ignored and plugins will not be synced ...`.

Also, is there a standard width for these? Your description seems to be wider than most of the others.

whopper: @MikaelSmith I like it! Updated.

lib/puppet/defaults.rb
```diff
@@ -1570,7 +1570,10 @@ def self.default_diffargs
     :pluginsync => {
       :default    => true,
       :type       => :boolean,
-      :desc       => "Whether plugins should be synced with the central server.",
+      :desc       => "Whether plugins should be synced with the central server. This setting is
```
MikaelSmith: @jtappa this language make sense to you? Related to https://tickets.puppetlabs.com/browse/PUP-5497.

whopper: There's good specific context in https://tickets.puppetlabs.com/browse/PUP-5703,

jtappa: @MikaelSmith @whopper legit. So, if I remember correctly, pluginsync used to be enabled by default, this changes that?

joshcooper: Sorry, for not making PUP-5708 more clear (the desired behavior changed since that ticket was written).

For 4.x, we will preserve the current pluginsync behavior if the user specifies an explicit value. So if they say `pluginsync=false` then we won't pluginsync, even if we're using a cached catalog.

But if the user doesn't then the default value will be determined based on the `use_cached_catalog` setting.

So for this PR we just want to issue a deprecation warning if the user explicitly sets the property. I would remove the part about "this setting will be ignored and pluginsync will not be synced."

In 5.0, we will remove the `pluginsync` setting, and the agent will always pluginsync when getting a new catalog, or not pluginsync when applying a cached catalog.

lib/puppet/defaults.rb
```diff
@@ -1570,7 +1570,10 @@ def self.default_diffargs
...
+        deprecated. If a cached catalog is being used, this setting will be ignored and
+        plugins will not be synced.",
+     :deprecated => :completely,
```
joshcooper: I think this makes sense, whether set on the command line or puppet.conf.

lib/puppet/defaults.rb
jtappa: Thanks, @joshcooper this makes more sense. I think it's still important to notate the behavior if a cached catalog is being used? I think this language is okay as is. :+1:

MikaelSmith: @joshcooper Ok, the part I still don't quite follow: will pluginsync be performed if using a cached catalog and `pluginsync=true` is set explicitly?

whopper:
@joshcooper @jtappa @MikaelSmith cool, makes sense. I removed that last bit.

MikaelSmith:
:+1:

joshcooper:
:+1:

PR #4597, opened 2016-01-23
(PUP-5741) Fix confusing errors when comparing values

thallgren:
Prior to this commit, the method MessageData#label had two functions.
If called with an argument, it presented the String form for that
argument. If called without, it returned a `ModelLabelProvider`. This
made it impossible to produce a message for `nil`.

This commit splits the method in two. One `provider`method that takes
no arguments and just returns the `ModuleLabelProvider` and one that
always returns the String representation of its argument. All calls
are adjusted accordingly.

### DIFF
--- a/lib/puppet/pops/issues.rb
+++ b/lib/puppet/pops/issues.rb
@@ -73,19 +73,30 @@ module Puppet::Pops::Issues
       instance_eval &block
     end
 
-    # Returns the label provider given as a key in the hash passed to #format.
-    # If given an argument, calls #label on the label provider (caller would otherwise have to
-    # call label.label(it)
+    # Obtains the label provider given as a key `:label` in the hash passed to #format. The label provider is
+    # return if no arguments are given. If given an argument, returns the result of calling #label on the label
+    # provider.
     #
-    def label(it = nil)
-      raise "Label provider key :label must be set to produce the text of the message!" unless @data[:label]
-      it.nil? ? @data[:label] : @data[:label].label(it)
+    # @param args [Object] one object to obtain a label for or zero arguments to obtain the label provider
+    # @return [Puppet::Pops::LabelProvider,String] the label provider or label depending on if an argument is given or not
+    # @raise [Puppet::Error] if no label provider is found
+    def label(*args)
+      args.empty? ? label_provider : label_provider.label(args[0])
+    end
+
+    # Returns the label provider given as key `:label` in the hash passed to #format.
+    # @return [Puppet::Pops::LabelProvider] the label provider
+    # @raise [Puppet::Error] if no label provider is found
+    def label_provider
+      label_provider = @data[:label]
+      raise Puppet::Error, 'Label provider key :label must be set to produce the text of the message!' unless label_provider
+      label_provider
     end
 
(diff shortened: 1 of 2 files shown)


lib/puppet/pops/issues.rb
```diff
@@ -130,7 +135,7 @@ def self.hard_issue(issue_code, *args, &block)
   # @todo configuration
   #
   NAME_WITH_HYPHEN = issue :NAME_WITH_HYPHEN, :name do
-    "#{label.a_an_uc(semantic)} may not have a name containing a hyphen. The name '#{name}' is not legal"
+    "#{provider.a_an_uc(semantic)} may not have a name containing a hyphen. The name '#{name}' is not legal"
```
hlindberg: We could just make `a_an`, `a_an_uc`, `the`, and `the_uc` available directly. Here that would be:
```
"#{a_an_uc(semantic)} may not have a name containing a hyphen. The name '#{name}' is not legal"
```

thallgren: We could, but that's not within scope for this PR IMO.

hlindberg: The "provider.a_an_uc" does not read nicely, hence my suggestion to get rid of it completely. I think that is acceptable to do within the scope of this ticket. It is a very small additional change.

hlindberg:
Confusing - thomas' view says it is merged, my view says it is closed.... Dear github what is it?
I pulled the 3.x branch, and I could not see the commits in this PR there.

lib/puppet/pops/issues.rb
```diff
@@ -58,7 +58,8 @@ def format(hash ={})
   # in the issue's message producing block.
   # @api private
   #
-  class MessageData
+  class MessageData < Puppet::Pops::LabelProvider
```
hlindberg: Yeah, so that is a quick way of getting the methods. But it is not right :-)
This class is not allowed to have those methods itself as it is language agnostic. It is the label provider's job to do that. We happen to have an english provider that has a base implementation for all different kinds of english label providers.

The support should be implemented as delegation to the label provider that is referencing.

hlindberg: i.e. this class must have the mehods, and they must then call the same method on the provider.

hlindberg: The I18N implementation is only embryonic so hard to guess this from just looking at the code.

hlindberg: You can do a method missing, and delegate to the provider if it is set, and it it responds to the missing method.

thallgren: Regarding delegation. Delegating will not make a future I18N transition easier because the methods themselves are wrong. `a_an`, `a_an_uc` for instance, is hardly applicable in, say Swedish. We would need others though, to differentiate 'en', 'ett' etc. Hence, all issues and the label provider must be rewritten.

The only way to make it simpler is to retain my previous attempt for this PR and go via `provider.a_an` in the actual text of the message since that's the place where things must change.

hlindberg: After discussion we decided to change the label method instead so it takes *args. That fixes the actual problem and leaves the rest alone.

PR #4598, opened 2016-01-25
(PUP-917) Update naming of cached catalog status in report

whopper:
This commit renames each of the three states of the
'cached_catalog_status' key of the report in order to
better convey their meaning.

The new values are 'not_used', 'explicitly_requested',
and 'on_failure'.

### DIFF
--- a/api/docs/http_report.md
+++ b/api/docs/http_report.md
@@ -110,7 +110,7 @@ example is formatted for readability)
          "change_count"=>0,
          "out_of_sync_count"=>0,
          "events"=>[]}},
-      "cached_catalog_status"=> "unused"}
+      "cached_catalog_status"=> "not_used"}
 
 Schema
 ------
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -104,11 +104,11 @@
         },
 
         "cached_catalog_status": {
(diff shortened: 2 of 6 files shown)


whopper:
I'm mostly putting this up to see if anyone has any opinions about the way this is named. I realized that I named the first cached catalog state 'use_cached_catalog', which matches the name of the setting it refers to. Any thoughts on whether this value should be the same?

whopper:
Ping @nfagerlund

jtappa:
+1 to states `none`, `explicit`, `failure`. This makes sense to me, and seems less redundant/wordy/confusing than including the same names of the settings as states.

whopper:
@jtappa also helped me come up with `unused`, `explicitly_requested` and `on_failure` just now. That may be better than the above suggestion?

nfagerlund:
I think I prefer `not_used` to `unused`, but I feel pretty good about `explicitly_requested` and `on_failure`.

whopper:
@nfagerlund word! These feel way better to me as well - updating now..

MikaelSmith:
:+1:

joshcooper:
:+1:

PR #4599, opened 2016-01-25
(PUP-5718) Update puppet.bat to current standard

melissa:
With the switch over to building puppet-agent for windows with Vanagon,
we need to effectively install puppet with install.rb. This script
installs puppet.bat, which is a wrapper that lets windows handle ruby
scripts. This commit updates puppet.bat to use our environment.bat file if
we install it, and to be more in line with the puppet.bat file installed
with the puppet-agent package as built by puppet_for_the_win

This takes advantage of `environment.bat` in puppet-agent installed with https://github.com/puppetlabs/puppet-agent/pull/512

### DIFF
--- a/ext/windows/run_puppet_interactive.bat
+++ b/ext/windows/run_puppet_interactive.bat
@@ -1,0 +1,9 @@
+@echo Running puppet on demand ...
+@echo off
+SETLOCAL
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET "PATH=%~dp0;%PATH%"
+)
+elevate.exe "%~dp0puppet_interactive.bat"
--- a/install.rb
+++ b/install.rb
@@ -406,7 +406,7 @@ def install_binfile(from, op_file, target)
 
   File.open(from) do |ip|
     File.open(tmp_file.path, "w") do |op|
-      op.puts "#!#{ruby}"
+      op.puts "#!#{ruby}" unless $operatingsystem == "windows"
       contents = ip.readlines
       contents.shift if contents[0] =~ /^#!/
       op.write contents.join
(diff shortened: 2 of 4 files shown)


melissa:
Hmm, this might not be the best approach. We might want do go the route of mco (https://github.com/puppetlabs/marionette-collective/pull/365) given that we'll also need to install `puppet_interactive.bat` and `puppet_shell.bat`

melissa:
Still testing

ext/windows/run_puppet_interactive.bat
```diff
@@ -0,0 +1,7 @@
+@echo Running puppet on demand ...
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET PATH=%~dp0;%PATH%
+)
+elevate.exe "%~dp0puppet_interactive.bat"
```
melissa: depends on https://github.com/puppetlabs/puppet-agent/pull/515

shrug:
:+1:

melissa:
cc/ @MikaelSmith @kylog @joshcooper could one of you review this?

MikaelSmith:
It'd be nice to be consistent about whether we're keeping files on disk that install.rb copies, or generating them from install.rb.

install.rb
```diff
@@ -416,30 +416,35 @@ def install_binfile(from, op_file, target)
...
-    end
+      if File.exists?("#{from}.cmd")
+        FileUtils.install("#{from}.cmd", File.join(target, "#{op_file}.cmd"), :mode => 0755, :preserve => true, :verbose => true)
+        installed_wrapper = true
+      end
 
-    if not installed_wrapper
-      tmp_file2 = Tempfile.new('puppet-wrapper')
-      cwv = <<-EOS
+      if not installed_wrapper
```
MikaelSmith: This is so weird, why don't we just have a `puppet.bat` file as well?

MikaelSmith:
The contents of the batch files look fine, going to test them all out tomorrow.

install.rb
melissa: No idea... I'm trying to do the least amount of damage to anyone who might be using this script to install puppet from source...

MikaelSmith: Ok.

MikaelSmith:
Looks good in stand-alone install and a puppet-agent build. :+1:

PR #4600, opened 2016-01-26
(PUP-5745) Use PuppetDB 4 API for querying where applicable

nicklewis:
Previously, the CapabilityFinder was using the
Puppetdb::Util::Http.action(...) method to query PuppetDB. In PuppetDB 4
however, the action() method has changed to require a new optional
argument, which breaks compatibility. At the same time, a top-level
query_puppetdb() method has been introduced which is intended to be the
public API going forward. We now check for and use the query_puppetdb()
method where it exists, falling back to the old internal action() method
when necessary.

### DIFF
--- a/lib/puppet/resource/capability_finder.rb
+++ b/lib/puppet/resource/capability_finder.rb
@@ -35,15 +35,10 @@ module Puppet::Resource::CapabilityFinder
           ['select_catalogs',
             ['=', 'code_id', code_id]]]]
     end
-    query = query_terms.to_json
 
-    Puppet.notice "Capability lookup #{cap}]: #{query}"
+    Puppet.notice "Capability lookup #{cap}]: #{query_terms}"
 
-    http = Puppet::Util.const_get('Puppetdb').const_get('Http')
-    response = http.action("/pdb/query/v4/resources?query=#{CGI.escape(query)}") do |conn, uri|
-      conn.get(uri, { 'Accept' => 'application/json'})
-    end
-    json = response.body
+    data = query_puppetdb(query_terms)
 
     # The format of the response body is documented at
     #   http://docs.puppetlabs.com/puppetdb/3.0/api/query/v4/resources.html#response-format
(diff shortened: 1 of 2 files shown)


nicklewis:
Pending potential further changes in puppetlabs/puppetdb:1824...

nicklewis:
This has been updated per puppetlabs/puppetdb#1824 and should be ready to go.

lib/puppet/resource/capability_finder.rb
```diff
@@ -35,15 +35,10 @@ def self.find(environment, code_id, cap)
           ['select_catalogs',
             ['=', 'code_id', code_id]]]]
     end
-    query = query_terms.to_json
 
-    Puppet.notice "Capability lookup #{cap}]: #{query}"
+    Puppet.notice "Capability lookup #{cap}]: #{query_terms}"
```
briancain: I'm assuming this no longer needs to be json anymore?

nicklewis: It looks basically the same either way (except for spaces), so it seemed like there was no need to JSONify it just for consistency with the old behavior. Could always change it if need be.

Before:
```
["and",["=","type","Sql"],["=","title","foo"],["=","tag","producer:production"],["in","certname",["extract","certname",["select_catalogs",["=","code_id","123456"]]]]]
```

After:
```
["and", ["=", "type", "Sql"], ["=", "title", "foo"], ["=", "tag", "producer:production"], ["in", "certname", ["extract", "certname", ["select_catalogs", ["=", "code_id", "123456"]]]]]
```

briancain: Cool cool!

briancain:
Looks good to me!

nicklewis:
@kylog Ping for visibility as this is blocking orchestrator integration tests. Would be extremely helpful if someone could take a look.

kylog:
:+1:

PR #4603, opened 2016-01-26
(PUP-5538) Improve SID::Principal.lookup_account_sid validation

glennsarti:
Previously the lookup_account_sid function would allow any
type of argument when calling lookup_account_sid, however this would
cause inconsistent error messages and possible memory errors when using
zero byte arrays or nils.  This commit adds validation checking to
sid_bytes before allocating memory pointers and adds tests for these
scenarios.

### DIFF
--- a/lib/puppet/util/windows/principal.rb
+++ b/lib/puppet/util/windows/principal.rb
@@ -87,6 +87,10 @@ module Puppet::Util::Windows::SID
 
     def self.lookup_account_sid(system_name = nil, sid_bytes)
       system_name_ptr = FFI::Pointer::NULL
+      if (sid_bytes.nil? || (!sid_bytes.is_a? Array) || (sid_bytes.length == 0))
+        raise Puppet::Util::Windows::Error.new('Byte array for lookup_account_sid must not be nil and must be at least 1 byte long')
+      end
+
       begin
         if system_name
           system_name_wide = Puppet::Util::Windows::String.wide_string(system_name)
--- a/lib/puppet/util/windows/principal.rb
+++ b/lib/puppet/util/windows/principal.rb
@@ -12,12 +12,18 @@ module Puppet::Util::Windows::SID
       @sid_bytes = sid_bytes
       @sid = sid
       @domain = domain
-      # when domain is available, combine it with parsed account
-      # otherwise use the account value directly
-      @domain_account = domain && !domain.empty? ?
-        "#{domain}\\#{@account}" : account
-
       @account_type = account_type
+      # When domain is available and it is a Domain principal, use domain only
+      #   otherwise if domain is available then combine it with parsed account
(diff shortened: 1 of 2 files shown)


lib/puppet/util/windows/principal.rb
```diff
@@ -87,6 +87,10 @@ def self.lookup_account_name(system_name = nil, account_name)
 
     def self.lookup_account_sid(system_name = nil, sid_bytes)
```
Iristyle: Should `system_name` get some validation also?

glennsarti: All of the bad cases will be trapped by the first call which converts it to a wide string. Even an empty string will still be two bytes long due to the encoding.  The same issue of zero length pointers will not occur.

Iristyle: :+1:

Iristyle:
I think we should tack on a quick test for the `:SidTypeDomain`resolution as well in another commit - i.e. call `lookup_account_name` using an empty string for `account_name` / and lookup_account_sid with the byte equivalent of `S-1-5-32

Iristyle:
Minor - might want to reword commit subject to something like `(PUP-5538) Improve lookup_account_sid validation` since we're handling an edge case rather than fixing something that's completely broken.

glennsarti:
Added BUILTIN Domain test cases

spec/integration/util/windows/principal_spec.rb
```diff
@@ -111,5 +142,12 @@
         principal.lookup_account_sid([1, 1, 0, 0, 0, 0, 0, 1, 1, 0, 0, 0])
       }.to raise_error(Puppet::Util::Windows::Error, /Failed to call LookupAccountSidW:  No mapping between account names and security IDs was done/)
     end
+    
+    it "should return a domain principal for BUILTIN SID S-1-5-32" do
+      principal = Puppet::Util::Windows::SID::Principal.lookup_account_sid([1, 1, 0, 0, 0, 0, 0, 5, 32, 0, 0, 0])
+      expect(principal.account_type).to eq(:SidTypeDomain)
+      expect(principal.sid).to eq('S-1-5-32')
+      expect(principal.domain).to eq('BUILTIN')
+    end
```
Iristyle: Ah yeah, but what about looking up `BUILTIN` by name?  we should see that `account_domain` is hosed as `BUILTIN\BUILTIN` - which is the part I was interested in fixing now (one extra commit is fine to address that)

glennsarti: @Iristyle Extra commit added for the account_domain generation and tests

Iristyle:
Follow-up commit at https://github.com/puppetlabs/puppet/commit/d376814db9c8110517272c778041a3ce7004332a

PR #4604, opened 2016-01-26
(PUP-5695) Add checksum_value parameter to file type

MikaelSmith:
Adds a checksum_value parameter to the file type. Validates that it's
format matches the specified checksum parameter.


MikaelSmith:
Updated.

PR #4605, opened 2016-01-27
(PUP-5708) Only emit deprecation warning for pluginsync if explicitly set

whopper:
This commit updates the pluginsync deprecation warning to only present
itself if the setting is actually set via the commandline or by config.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1572,7 +1572,9 @@ EOT
       :type       => :boolean,
       :desc       => "Whether plugins should be synced with the central server. This setting is
         deprecated.",
-     :deprecated => :completely,
+      :hook => proc { |value|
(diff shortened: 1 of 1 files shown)


joshcooper:
:+1: just waiting on ci checks:

```
$ bundle exec puppet agent -t
...
$ bundle exec puppet agent -t --pluginsync
Warning: Setting 'pluginsync' is deprecated.
   (at /Users/josh/work/puppet/lib/puppet/defaults.rb:1576:in `block in <module:Puppet>')
...
$ bundle exec puppet agent -t --no-pluginsync
Warning: Setting 'pluginsync' is deprecated.
   (at /Users/josh/work/puppet/lib/puppet/defaults.rb:1576:in `block in <module:Puppet>')
...
```

PR #4606, opened 2016-01-27
(PUP-3077) Handle cached catalog inconsistencies

joshcooper:
Previously, it was possible to run an agent in noop mode in a canary environment, but during the next normal agent run, it could fall back to the previously cached catalog from the noop run, causing changes to be deployed into production there were never meant to be.
 
The short version for this PR is: Don't save cached catalogs when running in `noop` mode and only fallback to a cached catalog whose environment matches the agent's configured environment, either through the ENC or agent-specified environment.

### DIFF
--- a/lib/puppet/application/agent.rb
+++ b/lib/puppet/application/agent.rb
@@ -404,7 +404,9 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
     # we want the last report to be persisted locally
     Puppet::Transaction::Report.indirection.cache_class = :yaml
 
-    if Puppet[:catalog_cache_terminus]
+    if Puppet[:noop]
+      Puppet::Resource::Catalog.indirection.cache_class = nil
+    elsif Puppet[:catalog_cache_terminus]
       Puppet::Resource::Catalog.indirection.cache_class = Puppet[:catalog_cache_terminus]
     end
 
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -72,6 +72,8 @@ class Puppet::Configurer
     query_options ||= {}
     if (Puppet[:use_cached_catalog] && result = retrieve_catalog_from_cache(query_options))
       @cached_catalog_status = 'explicitly_requested'
+
+      Puppet.info "Using cached catalog from environment '#{result.environment}'"
     else
       result = retrieve_new_catalog(query_options)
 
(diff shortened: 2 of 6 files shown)


lib/puppet/application/agent.rb
```diff
@@ -404,7 +404,7 @@ def setup
     # we want the last report to be persisted locally
     Puppet::Transaction::Report.indirection.cache_class = :yaml
 
-    if Puppet[:catalog_cache_terminus]
+    if Puppet[:catalog_cache_terminus] && !Puppet[:noop]
```
joshcooper: It might be better to explicitly set `cache_class` to nil if `Puppet[:noop]`.

lib/puppet/configurer.rb
```diff
@@ -71,13 +73,18 @@ def retrieve_catalog(query_options)
 
         if result
           @cached_catalog_status = 'on_failure'
+
+          if result.environment == @environment
+            Puppet.info "Using cached catalog from environment '#{result.environment}'"
+          else
+            Puppet.err "Not using cached catalog because its environment '#{result.environment}' does not match '#{@environment}'"
+            result = nil
```
whopper: I think we decided that the report shouldn't indicate the cached catalog was used ("on_failure") if we don't actually use the cache, so I think we need to only set it above if `result.environment == @environment`

joshcooper: good catch, will update

joshcooper: Some discussion about whether we should always fail the run here. If the agent is configured in an environment `production`, and the ENC normally assigns the node to a different one `dev`, then the next time the agent runs, if the node request fails, then the agent's current environment will never match the cached catalog's environment, and we'd never apply the cached catalog.

joshcooper:
I'll resolve conflicts

joshcooper:
```
$ bundle exec puppet agent -t --environment dev
Info: Using configured environment 'dev'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Caching catalog for <host>
Info: Applying configuration version '1454529937'
$ cat  ~/.puppetlabs/opt/puppet/cache/client_data/catalog/<host> | jq .environment
"dev"
$ bundle exec puppet agent -t --environment production
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Caching catalog for <host>
Info: Applying configuration version '1454529949'
$ cat  ~/.puppetlabs/opt/puppet/cache/client_data/catalog/<host> | jq .environment
"production"
$ bundle exec puppet agent -t --environment dev --noop
Info: Using configured environment 'dev'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Applying configuration version '1454529962'
$ cat  ~/.puppetlabs/opt/puppet/cache/client_data/catalog/<host> | jq .environment
"production"
```

whopper:
:+1: verified that `noop` runs don't result in a new cached catalog and that if the catalog fails to compile on a noop run, we *never* fall back to using a cached catalog.

Also verified that `cached_catalog_status` is set correctly in each of the three cases (not used, explicitly requested, and on failure). Checked that the status is not set to `on_failure` if the cached catalog is not used.

PR #4608, opened 2016-01-28
(maint) Use draining /bin/true for autosign tests

nwolfe:
This is a temporary adjustment to the CA autosign acceptance tests to
use a version of /bin/true that will drain the stdin stream. This is
necessary to get the test passing against Puppet Server while we work on
a bug fix for SERVER-1116.

### DIFF
--- a/acceptance/tests/ssl/autosign_command.rb
+++ b/acceptance/tests/ssl/autosign_command.rb
@@ -32,9 +32,25 @@ test_name "autosign command and csr attributes behavior (#7243,#7244)" do
   fqdn = master.execute('facter fqdn')
 
   step "Step 1: ensure autosign command can approve CSRs" do
+
+    # Our script needs to consume stdin (SERVER-1116)
+    # We should revert back to /bin/true once resolved
+    autosign_true_script = <<-EOF
+!#/bin/bash
+while read line
+do
+    : # noop command
+done
+exit 0
+EOF
+
+    autosign_true_script_path = "#{testdirs[master]}/mytrue"
+    create_remote_file(master, autosign_true_script_path, autosign_true_script)
+    on(master, "chmod 777 #{autosign_true_script_path}")
+
     master_opts = {
       'master' => {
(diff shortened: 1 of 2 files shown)


acceptance/tests/ssl/certificate_extensions.rb
```diff
@@ -28,7 +28,7 @@
       'environmentpath' => environments_dir,
     },
     'master' => {
-      'autosign' => '/bin/true',
+      'autosign' => true,
```
nwolfe: This test doesn't care how the autosign happens so we don't need to use the executable form here

acceptance/tests/ssl/autosign_command.rb
```diff
@@ -32,9 +32,22 @@ def assert_key_generated(name)
   fqdn = master.execute('facter fqdn')
 
   step "Step 1: ensure autosign command can approve CSRs" do
+
+    # Our script needs to consume stdin (SERVER-1094)
+    # We should revert back to /bin/true once resolved
+    autosign_true_script = <<-EOF
+!#/bin/bash
+less
```
joshcooper: Will this block if `PAGER` is defined? Is there another way to drain stdin without involving terminals and possibly hanging, something like:

```bash
#!/bin/bash
while read line
do
    : # noop command
done
exit 0
```

nwolfe: Yeah I was suspicious about this.  I like the explicit readlines too, I'll change to this.

nwolfe:
@joshcooper Updated. Running tests locally.

joshcooper:
Could you change the summary to `(maint)` and reference the server ticket in the body of the commit message?

nwolfe:
@joshcooper Done. Tests passed locally.

joshcooper:
Our ci check expects lowercase on `(maint)` to assist with automated ticket reconciliation.

nwolfe:
@joshcooper Ah - updated.

PR #4609, opened 2016-01-28
(PUP-5117) Add inline_metadata method during catalog compile

MikaelSmith:
Adds an inline_metadata method called during catalog compilation, but
after parsing the catalog. This method will be used for inlining
metadata as part of building a static catalog.

Inlines checksum_value for plain files.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -2,12 +2,14 @@ require 'puppet/node'
 require 'puppet/resource/catalog'
 require 'puppet/indirector/code'
 require 'puppet/util/profiler'
+require 'puppet/util/checksums'
 require 'yaml'
 
 class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
   desc "Compiles catalogs on demand using Puppet's compiler."
 
   include Puppet::Util
+  include Puppet::Util::Checksums
 
   attr_accessor :code
 
@@ -47,7 +49,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
     node = node_from_request(request)
     node.trusted_data = Puppet.lookup(:trusted_information) { Puppet::Context::TrustedInformation.local(node) }.to_h
 
-    if catalog = compile(node, request.options[:code_id])
+    if catalog = compile(node, request.options)
       return catalog
     else
       # This shouldn't actually happen; we should either return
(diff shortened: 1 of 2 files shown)


lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -98,6 +119,16 @@ def compile(node, code_id)
       end
     end
 
+    if node.environment.static_catalogs? && static_catalog && config.is_a?(model)
+      str = "Inlined resources into static catalog for for #{node.name}"
```
joshcooper: double "for for"

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -81,8 +81,29 @@ def add_node_data(node)
...
+
+      if file.recurse?
+        # TODO: find and replace metadata for children recursively
+      else
+        # TODO: find and replace metadata for the file
+      end
+    end
+
+    catalog
+  end
```
joshcooper: Would this logic be easier to test if it was moved to a different class, e.g. StaticCatalog{Filter,Inliner,Transformer,...} or some such, and then have the compiler terminus delegate to it, e.g. `catalog = StaticCatalogFilter.new(self, node).filter`?

Related, the `compiler` terminus already has some filtering logic to remove virtual resources, but the filter method gets called after `find` returns the catalog. I'm wondering if we should filter in a consistent way, either all the filtering happens in the `filter` method or in the `compile` method? 

Then we could have a base filter class that removes virtual resources, and have a static catalog filter class that extends it and also removes the resources above? Or we could chain filters?

MikaelSmith: Separating it out as a new class would make testing easier (`compile` is private so testing is indirect). Related, https://github.com/joshcooper/puppet/commit/81d4f41f2f19a8a371f9f5c13cd67b1005b24c3a has already tread this ground a bit.

I'll investigate whether we can inline resources in the `Compiler#filter`, or if that's too late.

MikaelSmith: Update: filtering happens after `storeconfigs` (and caching?), so we probably want to inline metadata before that.

MikaelSmith: I went with keeping it a private method of the Compiler, but simplified the tests by calling that method directly.

spec/unit/indirector/catalog/compiler_spec.rb
```diff
@@ -300,4 +337,114 @@ def a_request_that_contains(facts)
...
+              r.expects(:to_ral).returns(ral)
+
+              metadata = stub 'metadata'
+              metadata.stubs(:checksum).returns("{#{checksum}}#{sha}")
+              metadata.stubs(:ftype).returns("file")
+
+              source = stub 'source'
+              source.stubs(:metadata).returns(metadata)
+
+              ral.expects(:parameter).with(:source).returns(source)
```
MikaelSmith: Stubbing this was tricky because it reaches deep into the File type. The best way to clean it up would probably be to add a `metadata` method to the File type.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -81,8 +83,54 @@ def add_node_data(node)
...
+    # use the first entry in it that is also in known_checksum_types.
+    # If no match is found, give an error.
+    if agent_checksum_type
+      agent_checksum_types = agent_checksum_type.split('.').map {|type| type.to_sym}
+      checksum_type = agent_checksum_types.drop_while do |type|
+        not known_checksum_types.include? type
+      end.first
+    end
+
+    raise Puppet::Error, "Unable to find a common checksum type between agent '#{agent_checksum_types}' and master '#{known_checksum_types}'." unless checksum_type
```
joshcooper: If an older agent connects, it won't send a checksum_type param. Should master default to "md5" as that's the current default?

MikaelSmith: It won't send the static_catalog option either, and so won't enter this code. I do want to force the agent to declare its supported checksums.

joshcooper: If someone curls the endpoint and sets `static_catalog` but doesn't set `checksum_type` then we will enter this code, but fail on the "Unable to find a common checksum type..." Thinking it would better to fail the request sooner if only `static_catalog` is set, but `checksum_type` is not.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -98,6 +146,16 @@ def compile(node, code_id)
       end
     end
 
+    if node.environment.static_catalogs? && options[:static_catalog] && config.is_a?(model)
+      str = "Inlined resource metadata into static catalog for #{node.name}"
+      str += " in environment #{node.environment}" if node.environment
+      benchmark(:notice, str) do
+        Puppet::Util::Profiler.profile(str, [:compiler, :static_inline, node.environment, node.name]) do
+          config = inline_metadata(config, node.environment, options[:checksum_type])
```
joshcooper: `inline_metadata` only takes two arguments, looks like `node.environment` is extra.

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: If there's no shared checksum type, should we fail the run also, or do a non-static catalog?

joshcooper: Since `static_catalogs` are a new opt-in feature, I'd prefer to make the run fail, as that situation sounds like a misconfiguration issue, as opposed to a combination we should try to support. And in the future, clients will always request static catalogs, so the less legacy behavior the better.

MikaelSmith: :+1: that was my thinking as well

MikaelSmith:
Updated.

PR #4611, opened 2016-01-30
(PUP-5703) Handle cached catalog and pluginsync inconsistencies

whopper:
Prior to this commit, when attempting to use a cached catalog
the client would always make a node request and pluginsync,
even if there was a cached catalog available. This could cause
inconsistencies where the set of plugins used by the agent is
different than the set that was used when the catalog was
originally compiled.

This commit updates the client to first try to retrieve a cached
catalog if requested, skipping the node request and pluginsync.

In addition, if the agent's current environment differs from that
of the cached catalog, we now immediately switch to the cached
catalog's environment rather than going through the environment
convergence loop.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -22,6 +22,18 @@ class Puppet::Configurer
     "Puppet configuration client"
   end
 
+  def self.should_pluginsync?
+    if Puppet.settings.set_by_cli?(:pluginsync) || Puppet.settings.set_by_config?(:pluginsync)
+      Puppet[:pluginsync]
+    else
+      if Puppet[:use_cached_catalog]
+        false
+      else
+        true
+      end
+    end
+  end
+
   def execute_postrun_command
     execute_from_setting(:postrun_command)
   end
@@ -127,6 +139,15 @@ class Puppet::Configurer
     catalog
   end
 
+  def prepare_and_retrieve_catalog_from_cache
+    result = retrieve_catalog_from_cache({:transaction_uuid => @transaction_uuid, :static_catalog => @static_catalog})
+    if result
+      Puppet.info "Using cached catalog from environment '#{result.environment}'"
+      return convert_catalog(result, @duration)
+    end
+    nil
+  end
+
   # Retrieve (optionally) and apply a catalog. If a catalog is passed in
   # the options, then apply that one, otherwise retrieve it.
   def apply_catalog(catalog, options)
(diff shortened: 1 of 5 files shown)


whopper:
@joshcooper @MikaelSmith  this is sort of a proof of concept based around our discussions. As it stands, this is kind of weird and hard to test. It'd be much easier to use `retrieve_catalog`, but the behavior there is to fall back to retrieving a new catalog if the cached catalog retrieval fails. I don't think we want that to happen before doing a node request / pluginsync if we're going to need to download a new catalog due to a cache failure.

lib/puppet/configurer.rb
```diff
@@ -22,6 +22,18 @@ def self.to_s
...
 
+  def self.should_pluginsync?
+    if !(Puppet.settings.set_by_cli?(:pluginsync) || Puppet.settings.set_by_config?(:pluginsync))
+      if Puppet[:use_cached_catalog]
+        false
+      else
+        true
+      end
+    else
+      Puppet[:pluginsync]
```
joshcooper: nit, could you flip this so it's:

```ruby
if Puppet.settings.set_by_cli?(:pluginsync) || Puppet.settings.set_by_config?(:pluginsync))
  Puppet[:pluginsync]
else
  if Puppet[:use_cached_catalog]
    ...
  end
end
```

whopper:
@joshcooper what do you think about some kind of `retrieve_catalog_from_cache_only` method to try and make this easier to test? I'm trying to think of a way to get around not being able to use `prepare_and_retrieve_catalog` due to that chain of methods falling back to downloading a new catalog if cached catalog retrieval fails.

lib/puppet/configurer.rb
```diff
@@ -168,6 +189,23 @@ def run_internal(options)
...
+        options[:catalog] = catalog
+        @cached_catalog_status = 'explicitly_requested'
+        @environment = catalog.environment
+        report.environment = catalog.environment
+      else
+        # Don't try to retrieve a catalog from the cache again after we've already
+        # failed to do so the first time.
+        Puppet[:use_cached_catalog] = false
+        Puppet[:usecacheonfailure] = false
+        options[:pluginsync] = Puppet::Configurer.should_pluginsync?
```
joshcooper: This makes sense, no need to try the cache again, unless some code below has a side-effect of storing a cached catalog. But I don't think that's the case.

spec/unit/configurer_spec.rb
```diff
@@ -765,4 +796,32 @@ def method_missing(*args)
       @agent.convert_catalog(@oldcatalog, 10)
     end
   end
+
+  describe "when determining whether to pluginsync" do
+    it "should default to Puppet[:pluginsync] when explicitly set by the commandline" do
+      Puppet.settings[:pluginsync] = false
```
joshcooper: I'm not sure why, but sometimes we do `Puppet[:name]` and sometimes `Puppet.settings[:name]` in unit tests?

spec/unit/configurer_spec.rb
```diff
@@ -765,4 +796,32 @@ def method_missing(*args)
...
+      Puppet.settings.expects(:set_by_cli?).returns(true)
+
+      expect(described_class.should_pluginsync?).to eq(false)
```
joshcooper: Nit, with predicate methods you can do `expect(described_class).to be_should_pluginsync` which can provide a more useful error message if the test fails (instead of "true doesn't equal false"). I'm also fine if you want to leave it as is.

whopper:
@joshcooper updated with the fancy predicate matchers! Still no idea why we do both `Puppet[:blah]` and `Puppet.settings[:blah]`

spec/unit/configurer_spec.rb
```diff
@@ -599,6 +600,22 @@ def method_missing(*args)
         expect(@agent.retrieve_catalog({})).to eq(@catalog)
       end
 
+      it "should not make a node request or pluginsync when a cached catalog is successfully retrieved" do
+        Puppet::Node.indirection.expects(:find).never
+        @agent.expects(:download_plugins).never
+
+        @agent.expects(:prepare_and_retrieve_catalog_from_cache).returns(@catalog)
```
joshcooper: Could you set the expectation on `Puppet::Resource::Catalog.indirection.find` with `ignore_terminus => true` as opposed to the private method of the class we're testing?

whopper: @joshcooper ah yeah, done.

PR #4614, opened 2016-02-02
(PUP-5724) Backport PUP-5693 Make function_loading benchmark operational

hlindberg:
Before this the function_loading benchmark failed with a out of stack
error. This was caused by a combination of the deep recursion set up by
the benchmark, and that the rutime needs additional stack levels per
call.

The benchmark logic also had flaws that had gone undetected in older
puppet versions. Now with stricter module visibility rules the benchmark
generated modules did not have the correct deendencies in meta data.

This changes the recursion to not go as deep and fixes the module
dependency problems. The number of runs/modules/funcions were also
lowered to reduce the running time of the benchmark as he additional
iterations does not provide more information.

One problem remains - each iteration takes about 25% longer than the
previous. There could be a natural explanatoin for this (GC), but
manually performing gc beteeen runs did not change the outcome. Newer
riibies performed better, but still with degradation of the performance.
It looks like there is either a memory leak (either real, or caused by
the benchmark's construction).

This also corrects a typo in the base_loader that would cause an
error-reporting case to fail with a method not found instead of the
actual error.

### DIFF
--- a/benchmarks/function_loading/benchmarker.rb
+++ b/benchmarks/function_loading/benchmarker.rb
@@ -20,7 +20,13 @@ class Benchmarker
 
   def run(args=nil)
     envs = Puppet.lookup(:environments)
+
     envs.clear('benchmarking')
+#    # Uncomment to get some basic memory statistics
+#    ObjectSpace.garbage_collect
+#    result = {}
+#    ObjectSpace.count_objects(result)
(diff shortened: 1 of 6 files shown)


hlindberg:
This code was reviewed when merged to master. It was cleanly cherry picked over to stable and the benchmark runs fine on stable as well. No need to review again. Will just wait for Travis to go green and then merge this.

kylog:
:+1:

PR #4616, opened 2016-02-02
(PUP-5802) Make the Yum package provider correctly parse epochs

mveytsman:
Without this patch applied, the Yum package provider assumes all version
epochs are "0", i.e. a package with version "1:1.2.3-4.el5" will be
interpreted as having epoch "0" not "1". This causes version comparisons
made by puppet to be incorrect when comparing packages of different
epoch's.

This bug is caused by the line:

```
e = String(Bignum(e)) rescue '0'
```
Bignum is not a function, so the line always throws an exception and the
rescue is always invoked.

This patch uses Integer instead of Bignum, rescues the specific
exception needed (this is so that an epoch like "foo" is coerced into
"0"), and provides a test for non-zero epochs

### DIFF
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -211,9 +211,14 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
       e = s[0,ei]
       s = s[ei+1,s.length]
     else
-      e = nil
+      e = '0'
     end
-    e = String(Bignum(e)) rescue '0'
+    begin
+      e = String(Integer(e))
+    rescue ArgumentError
+      # If there are non-digits in the epoch field, default to 0
+      e = '0'
+    end
     ri = s.index('-')
     if ri
       v = s[0,ri]
--- a/spec/unit/provider/package/yum_spec.rb
+++ b/spec/unit/provider/package/yum_spec.rb
@@ -51,6 +51,13 @@ describe provider_class do
       expect(v[:release]).to eq('4.el5')
     end
 
+    it 'should parse evr with non-zero epoch' do
+      v = provider.yum_parse_evr('1:1.2.3-4.el5')
+      expect(v[:epoch]).to eq('1')
+      expect(v[:version]).to eq('1.2.3')
+      expect(v[:release]).to eq('4.el5')
+    end
+
     it 'should parse version only' do
       v = provider.yum_parse_evr('1.2.3')
       expect(v[:epoch]).to eq('0')


whopper:
@mveytsman interesting! Thanks for the contribution!

We've run into lots of trouble around Yum / DNF epoch parsing (fairly recently, even. See https://tickets.puppetlabs.com/browse/PUP-5549). We'll be wanting to have a careful look at this to make sure all is well in both Yum and DNF. From a glance, this seems sane, and we plan to take time to do more review on our end soon.

MikaelSmith:
So I tried this on CentOS 7 with `puppet apply -e "package {'vim-minimal': ensure => '2:7.4.160'}"` and got
```
Error: Could not update: Execution of '/usr/bin/yum -d 0 -e 0 -y install vim-minimal-2:7.4.160' returned 1: Error: Nothing to do
Error: /Stage[main]/Main/Package[vim-minimal]/ensure: change from purged to 2:7.4.160 failed: Could not update: Execution of '/usr/bin/yum -d 0 -e 0 -y install vim-minimal-2:7.4.160' returned 1: Error: Nothing to do
Notice: Applied catalog in 0.38 seconds
```

That doesn't seem right. Do you get successful behavior?

whopper:
@MikaelSmith yeah, I thought that might be the case. Yum requires you to specify the `arch` field when you are providing an epoch, like so:

```
epoch:name-ver-rel.arch
```

When we specify the version '2:7.4.160' Puppet doesn't understand this constraint, as it doesn't have a field for arch. We then end up calling: `yum install vim-minimal-2:7.4.160` rather than `yum install vim-minimal-2:7.4.160.x86_64.` https://tickets.puppetlabs.com/browse/PUP-1364 describes this problem, which has been around forever. The yum provider has never had the ability to understand version differences when epochs are involved for this reason.

MikaelSmith:
I'm fine merging this, but I'd like to understand the use-case it fixes @mveytsman

mveytsman:
@MikaelSmith 

Before this patch puppet's version comparison would ignore epochs. For instance, if a user had package with a version "1.2" installed, and asked to install "3:1.2", we would not recognize these as different versions.

`yum_parse_evr` is called several times in `install` (https://github.com/puppetlabs/puppet/blob/b1d25510a0be3d4e1c08299830773c77b0d09a57/lib/puppet/provider/package/yum.rb#L133-L179), and I belive this bug will cause incorrect behavior such as what @whopper mentioned --- https://tickets.puppetlabs.com/browse/PUP-5549)

whopper:
@mveytsman would you mind pasting a manifest that this fixes? I'm trying to work up a reproducible test case but am having no luck thus far.

mveytsman:
@whopper  I don't have one on hand -- I came across the issue reading the code after reading a blog post that references it http://blog.jasonantman.com/2014/07/how-yum-and-rpm-compare-versions/ and I found the bug through code review.

The unit test I include, https://github.com/mveytsman/puppet/blob/fix/master/yumepochparsing/spec/unit/provider/package/yum_spec.rb#L54-L59, will fail without my patch, which is how i concluded that puppet always sets the epoch to 0.

PR #4619, opened 2016-02-03
(PUP-5699) Inline symlink metadata in static catalog

whopper:
This commit updates the static compilation inlining logic to
handle symlinks which have their `links` parameter set to manage.
The `source` parameter of such a resource references one of two
items: a link or a regular file.

When `source` references a link, the destination of that link is
inlined directly into the `target` parameter of the new resource.

If `source` references a regular file, the existing logic holds
and file content is copied into the new resource.

When `links` is set to `follow`, we fall back to the regular
content inlining behavior, which sets the `checksum_value` of
the new resource.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -85,6 +85,14 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
 
   # Rewrite a given file resource with the metadata from a fileserver based file
   def replace_metadata(resource, metadata)
+    if resource[:links] == "manage" && metadata.ftype == "link"
+      resource.delete(:source)
+      resource[:ensure] = "link"
+      resource[:target] = metadata.destination
+    elsif resource[:links] == "follow"
+      resource[:ensure] = metadata.ftype
+    end
+
     if metadata.ftype == "file"
       resource[:checksum_value] = sumdata(metadata.checksum)
     end
--- a/spec/unit/indirector/catalog/compiler_spec.rb
+++ b/spec/unit/indirector/catalog/compiler_spec.rb
@@ -432,6 +432,58 @@ describe Puppet::Resource::Catalog::Compiler do
       end
     end
 
+    describe "when inlining symlinks" do
+      it "sets ensure and target for links which are managed" do
+        catalog = build_catalog(node, 1, ['puppet:///modules/mymodule/config_file_link.txt'], {:ensure => 'link', :links => 'manage'})
+
+        catalog.resources.select {|r| r.type == 'File'}.each do |r|
+          ral = r.to_ral
+          r.expects(:to_ral).returns(ral)
+
+          metadata = stub 'metadata'
+          metadata.stubs(:ftype).returns('link')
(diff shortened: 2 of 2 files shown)


whopper:
There are four cases to look out for here, detailed below. My master-side setup is as follows:

/etc/puppetlabs/code/modules/foo/files
    real_file
    link_file -> real_file

1) `links => manage` and `source` points to an actual symlink:
```
  file { '/tmp/puppet_managed_file':
    ensure => link,
    source => "/etc/puppetlabs/code/modules/foo/files/link_file",
    links => manage,
  }
```

In this case, the goal is to have `/tmp/puppet_managed_file` point to the same destination that `link_file` does. So, we remove the new resource's `source` and set its `target` to `metadata.destination`, effectively bypassing `source` as the way to figure out what the destination is.

Note that this behavior is a little weird, since if the destination doesn't exist on the agent, you'll end up with a broken symlink. This is how Puppet currently is as well. It makes sense if you know what your filesystem looks like client-side while writing the manifests and you're just making assumptions about what will be there.

2)
```
  file { '/tmp/puppet_managed_file':
    ensure => link,
    source => "/etc/puppetlabs/code/modules/foo/files/real_file",
    links => manage,
  }
```

In this case, `source` is actually just a file. `/tmp/puppet_managed_file` should just be a regular file with the copied contents of `real_file`.

3)
```
  file { '/tmp/puppet_managed_file':
    ensure => link,
    source => "/etc/puppetlabs/code/modules/foo/files/link_file",
    links => follow,
  }
```

Here, `/tmp/puppet_managed_file` should be a regular file with the copied contents of `link_file`

4)

```
  file { '/tmp/puppet_managed_file':
    ensure => link,
    source => "/etc/puppetlabs/code/modules/foo/files/real_file",
    links => manage,
  }
```

Here, `/tmp/puppet_managed_file` should be a regular file with the copied contents of `real_file`

whopper:
I'll test this out some more tomorrow.

whopper:
@joshcooper updated tests.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -85,6 +85,14 @@ def add_node_data(node)
 
   # Rewrite a given file resource with the metadata from a fileserver based file
   def replace_metadata(resource, metadata)
+    if resource[:links] == "manage" && metadata.ftype == "link"
+      resource.delete(:source)
```
joshcooper: It seems inconsistent to delete `source` here, but not `links => follow` below? Should we delete `source` when `ensure => directory`? Originally we were planning on always preserving `source`, but perhaps it's not needed in some cases? How about if we preserve `source` for now, merge this PR, and talk about when to preserve `source` at today's sync up?

joshcooper: Ah, my bad, my previous comment was on the commit, but @whopper explained why we *have* to delete `source` in this case: https://github.com/whopper/puppet/commit/6fc1f7da26c9bb7a9634fb7cec69ecd1f8013f0f#commitcomment-15871702. We'll talk about the "should we delete source in all cases" later today.

PR #4621, opened 2016-02-03
(PUP-5604) (WIP) Make systemctl better at bubbling up errors to puppet logs

petems:
```bash
Error: Could not start Service[corosync]: Execution of '/usr/bin/systemctl start corosync' returned 1: 
Job for corosync.service failed. See 'systemctl status corosync.service' and 'journalctl -xn' for details.
```

Figure out a way of giving visability in logs when systemctl commands fail. Actually run the journatclt command to get the error, rather than requiring manual intervention.

I'm not sure the best way of doing this, and its been a long time since I've tried doing anything for puppet-core. I'll need some dev guidance on this if possible! :smile:

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -148,16 +148,35 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
   end
 
   def restartcmd
-    [command(:systemctl), "restart", @resource[:name]]
+    begin
+      [command(:systemctl), "restart", @resource[:name]]
+    rescue Puppet::ExecutionFailure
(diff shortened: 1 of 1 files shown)


DavidS:
Note here about the discussion we had online: spelunking through the code, it seems that this is the right thing to do, but at the wrong place. tracing calls through the inheritance, the start/stop/restart methods of the baseclass (service.rb) need to be overridden instead.

branan:
Yeah. As DavidS said, the service providers rely largely on their base class(es) to do the work. The rescues you added here will never fire, since those `*cmd` methods just return the command name to be executed. No execution actually happens there.

Since this has been sitting since February I'm gonna go ahead and close it, but if you have time to poke this again let us know

PR #4622, opened 2016-02-03
(PUP-5804) update acceptance test to support cisco

LuvCurves:


LuvCurves:
This requires the next version of beaker (2.33.0+) and beaker-hostgenerator (0.3.2+)

kylog:
This is making me wonder if we should look at a scheme for moving which-test-to-run-for-a-platform outside the repo itself. Thoughts @LuvCurves or @johnduarte or @joshcooper?

johnduarte:
@kylog I actually find it valuable to see the confines directly in the tests. This provides direct feedback on where the test is either not applicable or where there are known issues.

just my 2 cents.

LuvCurves:
@kylog I like the idea of having a single location to make changes like this.  I'm certain I'll have to do this exercise again when cisco-7 is ready and then again when Huawei is ready.

kylog:
I realize my comment above was an implementation detail without a motivating concern :)

FWIW then, my concern was that this approach means issuing releases of components *just* so that we can skip tests in those components. Avoiding that was why we went through the b-hg exercise in December so this feels like a regression in that regard.

LuvCurves:
Beaker 2.34.0 is available

sbeaulie:
Is this ready to go?

LuvCurves:
@sbeaulie Nope.  There are a few tests I'm currently skipping that I need to get a root cause on their failure.

LuvCurves:
Closing. Will refactor against latest stable branch

PR #4627, opened 2016-02-03
(PUP-5472) Acceptance test for puppet lookup command

seangriff:
This test calls puppet lookup to find values for symbols resolved: (1) global hiera, (2) per-environment data providers (hiera, ruby-function, and puppet-function), (3) per-module data providers (hiera, ruby-function, and puppet-function).

Environments are selected by (1) default, (2) command line option (--environment <env>), (3) by setting in puppet.conf and (4) by a simple external node classifier (enc.rb) specified in puppet.conf.

Tests are executed for symbols defined at each location, symbols defined in more than one location (to ensure definition is from proper location in hierarchy) and symbols that are not resolved at any provider.

[skip ci]


kylog:
@seangriff can you update the commit message to reflect the ticket number?

PR #4628, opened 2016-02-03
(PUP-5813) Prevent memory leaks due to environments that doesn't get garbage collected

thallgren:
This PR contains two commits. The first that deals with a memory leak cause by a class-level instance variable that holds on to functions (puppet functions created in Ruby) that in turn have a dynamic scope that holds on to an environment. The second deals with the global $env_module_directories, a hash that uses environments as keys. Both commits uses the same approach for solving the problem. A dedicated adapter object that is adapted to the environment for which they hold data.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -238,11 +238,6 @@ class Puppet::Configurer
       execute_postrun_command or return nil
     end
   ensure
-    # Between Puppet runs we need to forget the cached values.  This lets us
-    # pick up on new functions installed by gems or new modules being added
-    # without the daemon being restarted.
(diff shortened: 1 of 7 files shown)


hlindberg:
Ruby 2.2.4 failed in travis with an error about a "kill" that was expected but ever occurred. I restarted that job as it is unrelated to the changes.

hlindberg:
I would have liked to see some kind of test that a memory leak was plugged. But we really do not have a framework to do such tests. Meanwhile, the confirmation from the customer that tried this as a patch is sufficient as evidence that this does indeed cure leaks.

And to everyone reading this; DO NOT USE GLOBAL VARIABLES!

PR #4632, opened 2016-02-04
(PUP-1780) Make puppet always warn on undefined variables

hlindberg:
This makes Puppet always warn on undefined variables. Earlier puppet only warned about non existing qualified variables with more than one segment.

A new logging method `warn_once` has been added to be able to better handle non-deprecation warnings that should be issued only once.

This also adds the ability to disable 'undefined_variables' in the setting `disable_warnings` (the same way as 'deprecations' can be disabled).

Warnings are only issues when --strict_variables == false (naturally).

The changes in runtime3support are made in preparation when --strict_variables=true is the default. The intent was to use this right away, but that meant that API breaking changes were required. I did leave that in this PR though (but it can be taken out if so desired).


hlindberg:
Hm, there are problems with Mocha and test order :-(  it does not take mocks down

thallgren:
Looks good, just some minor things and nitpicks, then ready to merge.

thallgren:
Can't get this one to build clean. Suggest rebase to get the updated .travis.yml.

hlindberg:
Yeah, that was probably the Ruby 2.1.8 problem - rebased.

PR #4633, opened 2016-02-04
(PUP-5701) Use checksum_value instead of content if specified

MikaelSmith:
If the checksum_value property is specified, use it instead of content
for verifying whether a file is insync.

Changes checksum_value to a property, and moves implementation of
writing data from source to the source parameter, rather than delegating
through a content property (which is no longer assumed to exist).

Separates writing source from content so it can be used by
checksum_value. In some cases content still triggers the write.

### DIFF
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -377,7 +377,7 @@ Puppet::Type.newtype(:file) do
 
     self.fail "You cannot specify more than one of #{CREATORS.collect { |p| p.to_s}.join(", ")}" if creator_count > 1
 
-    self.fail "You cannot specify a remote recursion without a source" if !self[:source] and self[:recurse] == :remote
+    self.fail "You cannot specify a remote recursion without a source" if !self[:source] && self[:recurse] == :remote
 
     self.fail "You cannot specify source when using checksum 'none'" if self[:checksum] == :none && !self[:source].nil?
 
@@ -385,7 +385,7 @@ Puppet::Type.newtype(:file) do
       self.fail "You cannot specify content when using checksum '#{checksum_type}'" if self[:checksum] == checksum_type && !self[:content].nil?
     end
 
-    self.warning "Possible error: recurselimit is set but not recurse, no recursion will happen" if !self[:recurse] and self[:recurselimit]
+    self.warning "Possible error: recurselimit is set but not recurse, no recursion will happen" if !self[:recurse] && self[:recurselimit]
 
     if @parameters[:content] && @parameters[:content].actual_content
       # Now that we know the checksum, update content (in case it was created before checksum was known).
@@ -410,7 +410,7 @@ Puppet::Type.newtype(:file) do
 
   # Determine the user to write files as.
   def asuser
-    if self.should(:owner) and ! self.should(:owner).is_a?(Symbol)
+    if self.should(:owner) && ! self.should(:owner).is_a?(Symbol)
       writeable = Puppet::Util::SUIDManager.asuser(self.should(:owner)) {
         FileTest.writable?(::File.dirname(self[:path]))
       }
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -124,11 +124,6 @@ Puppet::Type.newtype(:file) do
     end
   end
 
-  newparam(:checksum_value) do
-    desc "The checksum of the source contents. Only md5 and sha256 are supported when
-      specifying this parameter."
-  end
-
   newparam(:recurse) do
     desc "Whether to recursively manage the _contents_ of a directory. This attribute
       is only used when `ensure => directory` is set. The allowed values are:
@@ -730,7 +725,9 @@ Puppet::Type.newtype(:file) do
   end
 
   def retrieve
-    if source = parameter(:source)
+    # `checksum_value` implies explicit management of all metadata, so skip metadata
+    # retrieval. Otherwise, if source is set, retrieve metadata for source.
+    if (source = parameter(:source)) && property(:checksum_value).nil?
       source.copy_source_values
     end
     super
@@ -811,7 +808,7 @@ Puppet::Type.newtype(:file) do
 
   # Write out the file.  Requires the property name for logging.
   # Write will be done by the content or source property, with checksum
-  # computation handled by content
+  # computation handled by content or checksum_value
   def write(property = nil)
     remove_existing(:file)
 
@@ -930,9 +927,9 @@ Puppet::Type.newtype(:file) do
   # simply opening the file with 'w' as done in write is enough to truncate
   # or write an empty length file.
   def write_contents(file)
-    # Ideally only source or content would exist. However, if source is specified
-    # it will be used to generate content; content will trigger writing the file,
-    # but we should still use source to do the final write, so put it first.
+    # Ideally only source or content would exist. However, if source is specified without
+    # checksum_value, source will be used to generate content; content will trigger
+    # writing the file, but we should still use source to do the final write, so put it first.
     if source = parameter(:source)
       source.write(file)
     elsif content = property(:content)
@@ -958,6 +955,7 @@ Puppet::Type.newtype(:file) do
       thing.sync unless thing.safe_insync?(currentvalue)
     end
   end
+
 end
 
 # We put all of the properties in separate files, because there are so many
@@ -966,6 +964,7 @@ end
 require 'puppet/type/file/checksum'
 require 'puppet/type/file/content'     # can create the file
 require 'puppet/type/file/source'      # can create the file
+require 'puppet/type/file/checksum_value' # can create the file, in place of content
 require 'puppet/type/file/content_uri'
 require 'puppet/type/file/target'      # creates a different type of file
 require 'puppet/type/file/ensure'      # can create the file
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -810,8 +810,9 @@ Puppet::Type.newtype(:file) do
   end
 
   # Write out the file.  Requires the property name for logging.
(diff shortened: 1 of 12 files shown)


MikaelSmith:
This is a big refactor, so throwing it up early. I still have lots of new testing I want to write, and going to run acceptance tests.

lib/puppet/type/file.rb
```diff
@@ -719,7 +716,9 @@ def remove_existing(should)
   end
 
   def retrieve
-    if source = parameter(:source)
+    # `checksum_value` implies explicit management of all metadata, so skip metadata
+    # retrieval. Otherwise, if source is set, retrieve metadata for source.
+    if (source = parameter(:source)) && property(:checksum_value).nil?
```
MikaelSmith: This may cause problems for people providing checksum_value on their own. However we haven't really resolved how permissions and ownership are updated, so it's not clear to me what the alternative would look like.

lib/puppet/type/file.rb
```diff
@@ -396,6 +391,8 @@ def self.title_patterns
       self.fail "Checksum value '#{self[:checksum_value]}' is not a valid checksum type #{self[:checksum]}"
     end
 
+    self.fail "You cannot specify checksum_value without a source or content" if !self[:source] and !self[:content] and self[:checksum_value]
```
haus: This section (10-20 lines preceding) has an awesome mix of `and` and `&&`.

MikaelSmith: I guess I could clean that up while I'm at it.

MikaelSmith:
Acceptance run on RedHat 7 passed, excepting augeas tests that fail because I'm running from git (and pre-suite doesn't install augeas).

MikaelSmith:
Separated refactoring from new code.

MikaelSmith:
Added unit tests. Still want to add a few integration cases to confirm updating files when checksum_value is specified.

MikaelSmith:
Ah, Jenkins+Ruby hiccups.

MikaelSmith:
Updated. I believe this work's complete.

joshcooper:
more appveyor hiccups, rekicking

MikaelSmith:
Rebased to go back to Ruby 2.1.7

lib/puppet/type/file.rb
```diff
@@ -821,13 +819,26 @@ def write(property)
...
+    tempfile = Tempfile.new("puppet-file")
+    tempfile.open
+
+    write_contents(tempfile)
+
+    tempfile.close
+
+    yield tempfile.path
+
+    tempfile.delete
```
joshcooper: Might be good in maint commit to ensure this tempfile gets deleted. (I know it was that way before, but seems like it could leak temp file contents)?

MikaelSmith: As in
```
  def write_temporarily
    ...
    yield tempfile.path
  ensure
    tempfile.delete if tempfile
  end
```
?

joshcooper: yeah or if you nest the begin-ensure you don't need to check for tempfile:

```
def write_temporarily
   tempfile = ...
   begin
     ..
   ensure
     tempfile.delete
   end
end
```

MikaelSmith:
Pushed up the `ensure` change.

lib/puppet/type/file.rb
```diff
@@ -904,8 +915,15 @@ def fail_if_checksum_is_wrong(path, content_checksum)
...
-    (content = property(:content)) && content.write(file)
+  def write_contents(file)
+    # Ideally only source or content would exist. However, if source is specified without
+    # checksum_value, source will be used to generate content; content will trigger
+    # writing the file, but we should still use source to do the final write, so put it first.
+    if source = parameter(:source)
+      source.write(file)
+    elsif content = property(:content)
+      content.write(file)
+    end
```
joshcooper: Since this method is called by either `content` or `checksum_value`, can the caller specify which property to use so that we don't have to guess?

MikaelSmith: Yes, we could move this check into content, and then have each caller pass the property to write and write_temporarily. I think the only reason I preferred here was I didn't want content to know about source. I'll check whether there was another complication tomorrow.

lib/puppet/type/file/checksum_value.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    desc "The checksum of the source contents. Only md5 and sha256 are supported when
+      specifying this parameter."
+
+    def insync?(is)
+      # If checksum_value is specified, only manage source.
+      # Content manages its own syncing... FOR NOW, duh duh duh
+      if resource.parameter(:source).nil?
+        return true
+      end
```
joshcooper: I was confused by the comment "only manage source". If `checksum_value` is specified, but `source` is nil (so `checksum_value` must have been set explicitly with `content`) then we let `content` handle the `insync?`. If `source` is not nil, then `content` can't be set, and we handle the `insync?` logic below.

lib/puppet/type/file/checksum_value.rb
```diff
@@ -0,0 +1,74 @@
...
+      # insync? only returns false if it expects to manage the file content,
+      # so instruct the resource to write its contents.
+      return_event = @resource.stat ? :file_changed : :file_created
+      @resource.write
+      return_event
+    end
+
+    def retrieve
+      # Intentionally mirrors content#retrieve. Eventually content will be a
+      # parameter, and content syncing will be managed by checksum_value.
```
joshcooper: That's a little surprising to me. I've been thinking (some day) `content` would not have any `actual_content` or magic behavior around munging the value to the checksum. Instead the munged `content` value would just be the desired value specified in the manifest, and we would overwrite the `insync?` and `property_matches?` logic to compare checksums (but without relying on the existence of the `checksum_value` property. That way `content` and `source`/`checksum_value` are completely independent, and file writing is triggered either by calling `Content#sync` or `ChecksumValue#sync`.

But if I understand correctly, you're thinking of creating a single `checksum_value` property whose value is obtained either from `content` or `source`. So all of the `insync?` logic would be in one place, but the "get the desired checksum value" could be delegated to `content` and `source`?

That seems appealing though it's unintuitive to think of `content` as a parameter. But I suppose that's what `actual_content` was doing, acting as a parameter within the property.

lib/puppet/type/file/content.rb
```diff
@@ -87,6 +82,10 @@ def content
     # Also, fix #872: when content is used, and replace is true, the file
     # should be insync when it exists
     def insync?(is)
+      if resource[:source] && resource[:checksum_value]
+        self.fail Puppet::Error, "Content should not exist if source and checksum_value are specified"
+      end
```
joshcooper: The file type checks that you don't specify multiple ['should' values for content and source](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file.rb#L366-L374). Is this check redundant?

lib/puppet/type/file.rb
joshcooper: Would it ever make sense for someone to specify `source` and `checksum_value` (so it would have to be a file, and not a recursive directory), but rely on the `source` to provide permissions? I'm thinking not, given puppet's default behavior of ignoring source permissions. If you specify `checksum_value`, I think it makes sense to also specify permissions that you care about. Maybe add a note to `checksum_value` description?

lib/puppet/type/file/checksum_value.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+    end
+
+    def sync
+      # insync? only returns false if it expects to manage the file content,
+      # so instruct the resource to write its contents.
+      return_event = @resource.stat ? :file_changed : :file_created
+      @resource.write
+      return_event
+    end
```
joshcooper: The `sync` and `retrieve` methods are the same for `content` and `checksum_value`, and the `insync?` and `property_matches?` methods are very similar. Refactor to avoid duplicate logic? Or are you are thinking of removing the `content` logic once that flips to becoming a parameter?

lib/puppet/type/file/content.rb
MikaelSmith: Yeah, it was more in the nature of an assert. I could remove it.

MikaelSmith:
@joshcooper I think this is fixed up now. I'm running the acceptance tests/resource/file suite to check.

lib/puppet/type/file/checksum.rb
```diff
@@ -33,8 +33,6 @@ def sum_stream(&block)
     "{#{type}}#{checksum}"
   end
 
-  private
-
   # Return the appropriate digest algorithm with fallbacks in case puppet defaults have not
   # been initialized.
   def digest_algorithm
```
MikaelSmith: This was awkward, but seems necessary. I think the date checking was actually still incorrect; if you set to mtime via the digest_algorithm setting, the non-equals checking would always fail. Of course no one could ever set `digest_algorithm = mtime`, so it'd never cause a problem.

lib/puppet/type/file/contents_sync.rb
```diff
@@ -0,0 +1,82 @@
+require 'puppet/util/checksums'
+require 'date'
+require 'tempfile'
+
+module Puppet
+  module ContentsSync
```
MikaelSmith: This name isn't my favorite.

lib/puppet/type/file.rb
MikaelSmith: Done.

lib/puppet/type/file/checksum.rb
MikaelSmith: I was wrong about this, it has defaultto the setting. I was just testing it incorrectly. I'll restore `private`.

MikaelSmith:
Updated with test cleanup and fixes.

joshcooper:
LGTM, waiting on CI

PR #4634, opened 2016-02-04
(PUP-5742) Add ability to create aliases for Puppet Types

thallgren:
This PR:
1. Adds the type alias, type definition, and ability to use type as a function to the PP parser
2. A clean up of module namespaces
3. Evaluation of type aliases


hlindberg:
The namespace thing... I think everyone involved in Puppet needs to know this

hlindberg:
Suggest breaking out that first commit as a separate thing (maint), so we can merge that right away. You can do that merge on your own.

hlindberg:
I went through this once at a surface level. As noted in all of the comments you need to break this up into several commits. Get the fundamental stuff out of the way first. Then add feature by feature. I expect tests to run green between each commit. That way, when looking at a commit, the extent of the changes is known by looking at the tests it changed.

We may also want to break up the type.rb into separate files - it is both time consuming and difficult to review changes there since changes are peepholes  and you don't know where you are; add to that that there are multiple reasons for changes.... Maybe it looks better if changes are feature by feature.

The review to this point took me 4 hours, and it was not in depth.

thallgren:
I will break this into several feature commits but doing so before someone else has set eyes on it has its disadvantages. What if the reviewer has a completely different opinion on how things should be implemented that would render major parts of the work useless?

Given that I do not know what the final functional implementation will look like until I actually reach the point where it passes all the tests, and that I make changes all over the place until I reach that point, there will always be a large amount of changes. During the implementation period it would slow me down significantly if I had to do endless changes and rebases of commits. I do of course commit often, but one commit will not seldom completely rewrite what a prior commit added. Reviewing them would hence be much worse than reviewing the final outcome.

So here's a proposal for future work:
1. For a large implementation like this, I do what I've done here. I implement and deliver in major parts (here it was parser versus evaluator).
2. You do a high level review with the purpose to determine if the general approach is acceptable.
3. Assuming that we agree on approach, I split things up into several feature commits.
4. You do a detailed review of each feature.

hlindberg:
Nicely done Sir !
I appreciate the break down in to separate commits - made it a delight to review.

I basically suggested a couple of additional tests. The rest are just small points to make things a bit more readable when you parachute in...

hlindberg:
Looks great now, and thanks for addressing the review comments with new commits. That made it very easy to review the changes.

I added a ticket PUP 5843 for the suggested optimization to reduce the number of times we have to give up and use a TypeReference instead of a Resource type.

PR #4635, opened 2016-02-04
(PUP-5819) Error on byte order mark in lexer

hlindberg:
See the commit e16a204 for details about this PR.

### DIFF
--- a/lib/puppet/pops/issues.rb
+++ b/lib/puppet/pops/issues.rb
@@ -699,5 +699,9 @@ module Issues
   HEREDOC_MULTIPLE_AT_ESCAPES = hard_issue :HEREDOC_MULTIPLE_AT_ESCAPES, :escapes do
     "An escape char for @() may only appear once. Got '#{escapes.join(', ')}'"
   end
+
+  ILLEGAL_BOM = hard_issue :ILLEGAL_BOM, :format_name, :bytes do
+    "Illegal #{format_name} Byte Order mark at beginning of input: #{bytes} - remove these from the puppet source"
+  end
 end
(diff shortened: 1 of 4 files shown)


hlindberg:
Rebased on to fixes in PUP-1780 - test order issue and mocha....

hlindberg:
Uh oh, tests failing because Ruby is missing.... I don't think that is my fault :-) respinning.

thallgren:
There's a lot of errors like:
```
rspec './spec/unit/pops/parser/lexer2_spec.rb[1:180:2]' # Lexer2 when dealing with non UTF-8 and Byte Order Marks (BOMs) errors on the byte order mark for UTF_16_1 '[FE FF]'
```
A rebase is probably good too, to get the new .travis.yml.

hlindberg:
I just figured out what is going wrong - the platform is using UTF-16 and it reacts badly to the test data. The Ruby 1.9.3 works fine for some reason - perhaps running on something where UTF-8 is default (or some other Encoding that does not barf on that input.

There is a PR up that will force file reading to use UTF-8, the problem will go away then. 

Maybe I should move the check earlier to avoid the problem; the reported issues are raised when the illegal String is given to the StringScanner.

hlindberg:
Rebased and approach changed to asserting the string before creating the scanner. Also with test that checks if a UTF-16 string with the content in question can be given to the lexer with expected error rather than an encoding error (which travis flagged earlier).

thallgren:
Looks as this one needs to be rebased on master. It's not parented by the PUP-1780 commits that went in.

PR #4636, opened 2016-02-05
[WIP] (PUP-5119) Only statically inline metadata for files in the environment

branan:


branan:
This is "done" and functional, but I wanted to get some eyes on it before I try to sort out the new unit tests.

lib/puppet/file_serving/metadata.rb
```diff
@@ -121,23 +138,30 @@ def initialize(path,data={})
...
+        'type'   => checksum_type,
+        'value'  => checksum
+      },
+      'type'         => ftype,
+      'destination'  => destination,
+    }
+
+    # To maintain wire compatibility, we don't add the inlinable
+    # attribute to the serialized format unless it was explicitly set.
+    if !inlinable.nil?
```
branan: I've got these little checks in a few places around file metadata, trying to ensure that in cases where agent and server versions don't match, we don't change the wire format of these structs. I don't really know how strict Puppet's validation is or if this is a worthwhile endeavor

lib/puppet/file_serving/metadata.rb
```diff
@@ -88,6 +90,21 @@ def collect_stat(path)
   def collect(source_permissions = nil)
     real_path = full_path
 
+    if static_catalog
+      if environment && environment.static_catalogs?
+        env_path = File.join(Puppet[:environmentpath], environment.name.to_s) # environment.name is a symbol sometimes
+        puts "Checking if #{real_path} is in #{env_path}"
+        if real_path.start_with? env_path
```
joshcooper: Could the inlinable logic be moved out of Metadata and into the compiler itself (since the compiler already knows the environment, and you can call `Metadata#full_path` to get the resolved path)? That way we don't have to lookup the `:environmentpath` each time we collect metadata or modify metadata serialization?

branan: We could, but then we can't short-circuit here to skip the stat call when we don't need it

branan: Also, for purposes of PUP-5814, we'll need to know whether links are manage or follow, which we don't from the compiler layer

lib/puppet/file_serving/terminus_helper.rb
```diff
@@ -12,6 +12,9 @@ def path2instance(request, path, options = {})
     result.checksum_type = request.options[:checksum_type] if request.options[:checksum_type]
     result.source_permissions = request.options[:source_permissions] if request.options[:source_permissions]
 
+    result.environment = request.environment
```
branan: Gating this environment setter on the `static_catalog` option would avoid a bunch of the changes to tests below, but I kind of am happier with the terminus_helper not knowing the details of the models (specifically, that environment is only used when statically compiling)

lib/puppet/file_serving/metadata.rb
joshcooper: This code gets called on both the agent and server (through their respective `:file_*` termini), which is a bit surprising. Also it forces us to thread the environment and static catalog settings through to metadata and create a metadata object that we throw away if it's not inlineable. Also, soon we'll need to inline ["search" metadata](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file.rb#L677-L688) for directories, and more work will be required to thread that to the metadata class, because the terminus helper's [plural path2instances method doesn't forward any options](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_serving/terminus_helper.rb#L27-L29).

One possibility is to have the `:file_server` terminus do the check after it's [asked the mount to resolve the path](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/file_server.rb#L29). That way the inline check is only called on the server, and we don't even bother creating a metadata object (in the case we can't inline). However, the `:file_server` implementation is shared for both `metadata` and `content` models, which is terrible and is why `content` implements [`collect`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_serving/content.rb#L24-L27). We could do something hacky in `:file_server` like:

```ruby
def find(..)
  ...
  return nil unless path = mount.find(relative_path, request)
  return nil unless inlineable?(request, path)
  path2instance(request, path)
end

def inlineable?(request, path)
   if model == Puppet::FileServing::Metadata
    # check if path starts with request.environment dir
  else
    true
  end
end
```

That's pretty aweful too. Ideally `content` and `metadata` would have different `file_server` implementations. But those indirected types delegate to a single [`selector`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_serving/metadata.rb#L13). So we'd need to provide different selectors, e.g. `indirects :file_metadata, :terminus_class => :metadata_selector`, and then have `metadata_selector` delegate to `metadata_file_server` for puppet URIs. That seems like it would work, but seems like a lot of trouble.

A third option is not worry about the extra file stat, and just have the compiler do:

```
  metadata = file.parameter(:source).metadata
  if metadata && metadata.path.start_with?(env_path)
    replace_metadata(resource, metadata)
  end
```

About PUP-5814, the compiler has access to `resource[:links]` and `resource[:target]` (in the `manage` case).

The downside is we'd be performing an extra stat for any file resource whose source parameter was a puppet URI and resolved to outside the per-env code dir. I'm guessing that's a small number of files as compared to the number of files sourced from per-env modules?

joshcooper: Another option might be to add an `inlineable?` class method to `Puppet::FileServing::Content` and `Puppet::FileServing::Metadata`, and then have the file server just call:

```
return nil unless model.inlineable?(request, path)
```

where `Content.inlineable?` always returns true, and `Metadata.inlineable?` has the logic you've implemented, but just as a class method?

branan: I actually explicitly wanted to return a metadata object even when it's not inlinable - this lets us differentiate between the "the file is not inlinable" case and the "the file does not exist" case, so we can throw a compilation error on the latter. I'd be fine still returning the full metadata with an "is this inlinable?" attribute or method, though.

I still kind of prefer to do the actual inlinable determination in the `collect` call, since in the symlink follow case that involves additional stats, and I want to make it clear where "expensive" work happens

Re a couple of your other concerns:
* It looks like `path2instances` just calls out to the singular version, passing along the same request object. Options should make it just fine?
* The agent calling the same code is awful, but all the new behavior is gated on the `static_catalog` option being set, which it only will be for requests coming from the compiler

joshcooper: > so we can throw a compilation error on the latter

I would be reluctant to throw a compilation error, because we don't do that now. For example, a file resource can have multiple sources, and it's ok if some of the them don't exist on the master.

lib/puppet/file_serving/metadata.rb
joshcooper: Since the inlining process occurs on the master, and the agent is unaware that that's happening, I wouldn't think serialization would need to change.

PR #4643, opened 2016-02-07
DO NOT MERGE (maint) JUST TESTING TRAVIS RUBY BUILDS

kylog:


Iristyle:
It appears at first blush, that Ruby 2.1.8 is the problem - I think we should replace 2.1.8 with 2.1.7 for now.

Iristyle:
Addressed in #4644

PR #4646, opened 2016-02-07
(PUP-5835) Make node params String if they are Symbol

hlindberg:
Before this, if a node parameter had a Ruby Symbol value it would leak
to Scope, and into variables. This was especially true for the
$environment variable after at enviroment name is made into a Symbol.

This commit fixes the problem both by ensuring that when a node sets a
parameter (in itself) based on the environment's name it transformed the
symbolic name to a Strig, and in the general case, the compiler when
setting parameters for a node, will also transform a Symbol value to
Strig. The later to ensure the same kind of mistake does not leak out to
Scope.

This commit also includes unit tests that asserts that environmet name
as a symbol dos not leak, nor any other node parameter.

### DIFF
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -75,7 +75,7 @@ class Puppet::Node
 
     # Keep environment_name attribute and parameter in sync if they have been set
     unless @environment.nil?
-      @parameters[ENVIRONMENT] = @environment.name if @parameters.include?(ENVIRONMENT)
+      @parameters[ENVIRONMENT] = @environment.name.to_s if @parameters.include?(ENVIRONMENT)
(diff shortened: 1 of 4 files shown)


kylog:
Note that this change causes this test to fail:

https://github.com/puppetlabs/puppet/blob/a81f968650748442a2ccf6c9acd656c78b94c4e6/spec/unit/node_spec.rb#L90-L96

We could tweak that test of course, but it does highlight the magic around `to_s`ing node parameters. Thoughts?

kylog:
Also, some nits:
* can you rebase to pick up the travis.yml change?
* can you correct the spelling of "String" and "environment" in the commit message?

hlindberg:
meh, I missed that other failure

hlindberg:
Ping @kylog  it is green now.

PR #4648, opened 2016-02-08
(PUP-5820) Improve test coverage for managed symlinks

whopper:
This commit adds two test cases to the existing managed
symlink acceptance test. The first ensures that flat files
created with 'links => manage' are symlinks, and the second
ensures that symlinks inside of directories are copied
correctly when using 'links => manage' and 'recurse => true'.

### DIFF
--- a/acceptance/tests/resource/file/should_create_symlink.rb
+++ b/acceptance/tests/resource/file/should_create_symlink.rb
@@ -1,24 +1,16 @@
 test_name "should create symlink"
 
-message = 'hello world'
-agents.each do |agent|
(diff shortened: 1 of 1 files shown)


joshcooper:
I removed an extra comma, squashed it into your commit, and merged in 36dbfb5.

joshcooper:
hmm that's annoying, jira shows the PR as declined... guess I won't do that next time.

joshcooper:
Also I tested on ubuntu15.10, windows2012r2, and 2008r2-6432 and didn't have any issues.

PR #4649, opened 2016-02-09
(PUP-5698) In-line directory metadata

HAIL9000:
Update the catalog compiler so it inlines directory metadata for
static catalogs. This means that when there is a file resource
that is a directory and has recurse set to true, we will add resources
for all it's children to the catalog.

In order to support this, make a few changes to the file type as
well so that we can easily reuse functionality for static catalogs.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -135,7 +135,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
       file = resource.to_ral
 
       if file.recurse?
-        file.recurse_remote_metadata do |meta|
+        file.recurse_remote_metadata.each do |meta|
           # Don't create a new resource for the parent directory
           next if meta.relative_path == "."
 
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -84,7 +84,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
   end
 
   # Rewrite a given file resource with the metadata from a fileserver based file
-  def replace_metadata(resource, metadata)
+  def replace_metadata(resource, metadata, recurse = false)
     if resource[:links] == "manage" && metadata.ftype == "link"
       resource.delete(:source)
       resource[:ensure] = "link"
@@ -95,6 +95,13 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
 
     if metadata.ftype == "file"
       resource[:checksum_value] = sumdata(metadata.checksum)
+    end
+
+    if recurse
+      resource[:source] = metadata.source
+      if metadata.ftype == "file"
+        resource[:checksum] = metadata.checksum_type
+      end
     end
   end
 
@@ -128,7 +135,27 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
       file = resource.to_ral
 
       if file.recurse?
-        # TODO: find and replace metadata for children recursively
+        file.recurse_remote_metadata do |meta|
+          # Don't create a new resource for the parent directory
+          next if meta.relative_path == "."
+
+          # TODO: Conditionally copy owner, group, and mode when we can check if permissions are set
+          child_resource = Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
+          replace_metadata(child_resource, meta, true)
+
+          # Copy parameters from original parent directory
+          file.original_parameters.each do |param, value|
+            # These should never be passed to our children
+            unless [:parent, :ensure, :recurse, :recurselimit, :target, :alias, :source].include? param
+              child_resource[param] = value.to_s
+            end
+          end
+          #TODO: Add edge between `child_resource` and its nearest ancestor and
+          # add an edge between `child_resource` and all other resources that depend
+          # on `file`
+          catalog.add_resource(child_resource)
+          catalog.add_edge(file, child_resource)
+          end
       else
         metadata = file.parameter(:source).metadata
         raise "Could not get metadata for #{resource[:source]}" unless metadata
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -641,7 +641,7 @@ Puppet::Type.newtype(:file) do
 
   # Recurse against our remote file.
   def recurse_remote(children)
-    recurse_remote_metadata do |meta|
+    recurse_remote_metadata.each do |meta|
       if meta.relative_path == "."
         parameter(:source).metadata = meta
         next
@@ -654,12 +654,12 @@ Puppet::Type.newtype(:file) do
     children
   end
 
-  def recurse_remote_metadata(&block)
+  def recurse_remote_metadata
     sourceselect = self[:sourceselect]
 
     total = self[:source].collect do |source|
       next unless result = perform_recursion(source)
-      return if top = result.find { |r| r.relative_path == "." } and top.ftype != "directory"
+      return [] if top = result.find { |r| r.relative_path == "." } and top.ftype != "directory"
       result.each do |data|
         if data.relative_path == '.'
           data.source = source
@@ -682,9 +682,7 @@ Puppet::Type.newtype(:file) do
       end
     end
 
-    total.each do |meta|
-      yield meta
-    end
+    total
   end
 
   def perform_recursion(path)
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -641,6 +641,20 @@ Puppet::Type.newtype(:file) do
 
   # Recurse against our remote file.
   def recurse_remote(children)
+    recurse_remote_metadata do |meta|
+      if meta.relative_path == "."
+        parameter(:source).metadata = meta
+        next
+      end
+      children[meta.relative_path] ||= newchild(meta.relative_path)
+      children[meta.relative_path][:source] = meta.source
+      children[meta.relative_path].parameter(:source).metadata = meta
+    end
+
+    children
+  end
+
+  def recurse_remote_metadata(&block)
     sourceselect = self[:sourceselect]
 
     total = self[:source].collect do |source|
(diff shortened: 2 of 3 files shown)


lib/puppet/type/file.rb
```diff
@@ -656,25 +677,18 @@ def recurse_remote(children)
...
-      children[meta.relative_path].parameter(:source).metadata = meta
+      yield meta
     end
-
-    children
   end
 
   def perform_recursion(path)
+    Puppet.warning("Search meatadatas for #{self[:path]}")
     Puppet::FileServing::Metadata.indirection.search(
```
HAIL9000: Whoops this is for debugging, I'll take this out in my next commit

lib/puppet/type/file.rb
```diff
@@ -656,25 +677,18 @@ def recurse_remote(children)
...
-      children[meta.relative_path][:source] = meta.source
-      children[meta.relative_path].parameter(:source).metadata = meta
+      yield meta
     end
-
-    children
   end
 
   def perform_recursion(path)
+    Puppet.warning("Search meatadatas for #{self[:path]}")
```
HAIL9000: Whoops this is for debugging, I'll take this out in my next commit

lib/puppet/type/file.rb
```diff
@@ -641,12 +641,33 @@ def recurse_local
...
       next unless result = perform_recursion(source)
       return if top = result.find { |r| r.relative_path == "." } and top.ftype != "directory"
-      result.each { |data| data.source = "#{source}/#{data.relative_path}" }
+      result.each do |data|
+        if data.relative_path == '.'
+          data.source = source
+        else
+          # REMIND: appending file paths to URL may not be safe, e.g. foo+bar
+          data.source = "#{source}/#{data.relative_path}"
+        end
```
joshcooper: Could you split out the refactor of `recurse_remote` so it calls `recurse_remote_metadata` into one commit and preserve the old behavior of:

```
result.each { |data| data.source = "#{source}/#{data.relative_path}" }
```

And then add a second commit for the compiler to call the new `recurse_remote_metadata` method and makes this change (so `data.source = source` when relative_path is `.`?) That way the refactor doesn't make any behavior changes?

branan:
After chatting with @HAIL9000 , I think we need to do something very much like `File.remove_less_specific_files`. Possibly we can just call that method on the generating resource, as long as this code is structured apropriately

lib/puppet/type/file.rb
```diff
@@ -656,25 +677,18 @@ def recurse_remote(children)
       found = []
       total.reject! do |data|
         result = found.include?(data.relative_path)
-        found << data.relative_path unless found.include?(data.relative_path)
+        found << data.relative_path unless result
```
HAIL9000: @joshcooper When I split this into two commits, can this change happen in the first commit? Or does it have to wait until the second one. Not sure if it's a behavior change or not

joshcooper: I'd say the overall behavior is unchanged, but I also try not to make optimizations while refactoring, so I'd lean towards keeping it out of the first commit.

HAIL9000:
@joshcooper Updated! :+1:

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -127,7 +127,33 @@ def inline_metadata(catalog, checksum_type)
...
+        file.recurse_remote_metadata do |meta|
+          # Don't create a new resource for the parent directory
+          next if meta.relative_path == "."
+
+          child_resource = Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
+          child_resource[:ensure] = meta.ftype
+          # TODO: Make the coying of owner, group, and mode conditional on using permissions
+          child_resource[:owner]  = meta.owner.to_s
+          child_resource[:group]  = meta.group.to_s
+          child_resource[:mode]   = meta.mode.to_s(8).rjust(4, '0')
```
joshcooper: Can we take the owner/group/mode lines out and implement that part in PUP-5697 along with the conditional logic? As it is now the static compiler will always cause the source permissions to be copied to the agent, which is the opposite of the non-static catalog behavior (when `source_permissions => ignore`):

```
Notice: /Stage[main]/Ntp/File[/tmp/dir/dir1]/group: group changed 'wheel' to 'staff'
Notice: /Stage[main]/Ntp/File[/tmp/dir/file]/group: group changed 'wheel' to 'staff'
Notice: /Stage[main]/Ntp/File[/tmp/dir/dir1/dir2]/group: group changed 'wheel' to 'staff'
Notice: /Stage[main]/Ntp/File[/tmp/dir/dir1/file1]/group: group changed 'wheel' to 'staff'
```

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -127,7 +127,33 @@ def inline_metadata(catalog, checksum_type)
...
+
+          # Copy parameters from original parent directory
+          file.original_parameters.each do |param, value|
+            # These should never be passed to our children
+            unless [:parent, :ensure, :recurse, :recurselimit, :target, :alias, :source].include? param
+              child_resource[param] = value.to_s
+            end
+          end
+          catalog.add_resource(child_resource)
+          catalog.add_edge(file, child_resource)
```
joshcooper: Short version: We need to add an edge between the `child_resource` and its nearest ancestor (not necessarily the `file` we started recursing in). We also need to add an edge (I think) between the `child_resource` and all other resources that depend on `file`. I think the relationship logic is sufficient complex that we should ticket it separately.

Long version: Currently when an agent sources a recursive directory during [`eval_generate`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file.rb#L472), it makes a `file_metadatas` request (plural) and for each `metadata` object, constructs a `Puppet::Type::File` instance. It returns the array of generated resources to the [`additional_resource_generator`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction/additional_resource_generator.rb#L54-L58). The generator does 3 things:

1. Adds the generated resources to the catalog
1. Ensures all of the generated resources are contained in the recursive directory resource
1. Adds relationships from the each generated resource to its nearest ancestor.

I'm realizing now that the static compiler has more bugs. It does 1, doesn't handle 2, and incorrectly does 3.

About 2, the static compilation process needs to make sure all of the generated resources come before whatever comes after the file resource we started recursing in, so we don't regress in https://github.com/puppetlabs/puppet/commit/479cd1c087099c481009a68e57b6e0f617b790f3. For example, if you have:

```puppet
file { '/tmp/dir':
  ensure => directory,
  recurse => true,
  source => 'puppet:///modules/mymodule/dir'
} ->
notify { 'after tmp dir': }
```

Then we need to make sure all of the children of `/tmp/dir` are evaluated before the `notify`. The agent currently handles that logic in [`contain_generated_resources_in`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction/additional_resource_generator.rb#L71-L86).

However, we can't create containment edges in the compiler. But I think we can do what the `contain` function does by [adding relationship edges from each generated resource to the `notify`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/parser/functions/contain.rb#L34-L36)

The third issue is that we need to add an edge from the nearest ancestor to the generated resource. So create an edge like:

```
      "source": "File[/tmp/dir/dir1]",
      "target": "File[/tmp/dir/dir1/dir2]"
```

Instead of:

```
      "source": "File[/tmp/dir]",
      "target": "File[/tmp/dir/dir1/dir2]"
```

Otherwise there's no guarantee that `/tmp/dir/dir1` will be created before `/tmp/dir/dir1/dir2`. The agent has similar logic for [eval generated resources](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction/additional_resource_generator.rb#L97-L106). We could also do nothing and rely on autorequiring to create the edges for us.

HAIL9000: Does it seem appropriate that I add a TODO comment for now for this work? Or is there anything that needs to happen in this pull request.

joshcooper: I think it's fine to leave this part as is. We can revisit the "create an edge between the file and its nearest ancestor" and "creating edges from the generated resource to resources that depend on the original directory"

joshcooper: I added https://tickets.puppetlabs.com/browse/PUP-5873.

joshcooper:
@branan I'm thinking we can handle the `remove_less_specific_files` issue as part of PUP-5118, as there are (at least) two ways you can get duplicates. Either recursive directories overlap, e.g.

```puppet
file { '/tmp/foo':
  ensure  => directory,
  recurse => true,
  source  => 'puppet:///modules/mymodule/foo'
}
file { '/tmp/foo/bar':
  ensure  => directory,
  recurse => true,
  source  => 'puppet:///modules/mymodule/foo/bar'
}
```

Or a resource is generated during recursion and is also explicitly managed:

```puppet
file { '/tmp/foo':
  ensure  => directory,
  recurse => true,
  source  => 'puppet:///modules/mymodule/foo'
}
file { '/tmp/foo/bar':
  ensure  => file,
  content => '...'
}
```

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -128,7 +128,33 @@ def inline_metadata(catalog, checksum_type)
       file = resource.to_ral
 
       if file.recurse?
-        # TODO: find and replace metadata for children recursively
+        file.recurse_remote_metadata.each do |meta|
+          # Don't create a new resource for the parent directory
+          next if meta.relative_path == "."
+
+          # TODO: Conditionally copy owner, group, and mode when we can check if permissions are set
```
MikaelSmith: Lines 137-143 seem like they could be combined with replace_metadata pretty reasonably. That seems like it'd make ensuring the correct metadata is updated easier.

spec/unit/indirector/catalog/compiler_spec.rb
```diff
@@ -542,5 +542,31 @@ def build_catalog(node, num_resources, sources = nil, parameters = {:ensure => '
...
+        child_metadata.stubs(:ftype).returns('file')
+        child_metadata.stubs(:source).returns('puppet:///modules/mymodule/directory/myfile.txt')
+        child_metadata.stubs(:checksum_type).returns(checksum_type)
+        child_metadata.stubs(:checksum).returns('{md5}b1946ac92492d2347c6235b4d2611184')
+
+        ral.expects(:recurse_remote_metadata).returns([parent_metadata, child_metadata])
+      end
+
+      expect(@compiler.send(:inline_metadata, catalog, checksum_type)).to eq(catalog)
+      catalog.resources.select { |r| r.type == 'File' }.size.should equal(2)
```
joshcooper: Sorry just noticed this, could you change this to rspec3 syntax `expect(catalogs.resources(...).size).to eq(2)`? There's some good docs in https://github.com/rspec/rspec-expectations/blob/master/Should.md about that.

HAIL9000:
@joshcooper @MikaelSmith Updated! Let me know if you have more feedback

PR #4650, opened 2016-02-09
(PUP-5119) Only inline metadata if the source is in the environment

branan:
This adds a check around replacing the resource's metadata to verify
the file is versionable (exists within the environment path tied to
the catalog's code id). Doing this at the compiler layer greatly
simplifies the code compared to trying to forward all the needed
parameters to the file server layer. This does mean we occasionally
stat a file we won't inline, but that's worth it (until we see
real-world numbers saying it's expensive) to avoid making this super
complicated.


branan:
Ugh, looks like I was on an outdated master and need to update a few more tests

branan:
It would be cool to try to de-duplicate some of the source/metadata stubbing happening in these tests - it's getting fairly complex. Not sure if I want to pull that in to this PR, though

MikaelSmith:
This seems reasonable. And de-duplicating would be nice, but not urgent yet.

PR #4652, opened 2016-02-10
(PUP-5569) acceptance: ensure Resource Collectors can use resource references in search expressions

er0ck:
This change adds a test that uses a yaml terminus for storeconfigs to
test exported resources.  Due to a deficiency in the yaml storeconfigs
terminus we create here, we can not use with_puppet_running_on.

This test solely collects all resources of a given type.  Because filtering is implemented in the puppetdb terminus, it does no good to test it here.  Even testing collecting various types would do no good since this matching is also done in the terminus.  

[skip ci]

### DIFF
--- a/acceptance/lib/puppet/acceptance/environment_utils.rb
+++ b/acceptance/lib/puppet/acceptance/environment_utils.rb
@@ -1,4 +1,5 @@
 require 'puppet/acceptance/module_utils'
+require 'rbtree'
 
 module Puppet
   module Acceptance
@@ -357,6 +358,37 @@ module Puppet
         end
         assert failures.empty?, "Failed Review:\n\n#{failures.join("\n")}\n"
       end
+
+      # generate a random string of 6 letters and numbers.  NOT secure
+      def random_string
+        [*('a'..'z'),*('0'..'9')].shuffle[0,8].join
+      end
+
+      # if the first test to call this has changed the environmentpath, this will cause trouble
+      #   maybe not the best idea to memoize this?
+      def environmentpath
+        @@memoized_environmentpath ||= master.puppet['environmentpath']
+      end
+
+      # create a tmpdir to hold a temporary environment bound by puppet environment naming restrictions
+      # symbolically link environment into environmentpath
+      def mk_tmp_environment(environment)
+        # add the tmp_environment to a sortedset so we ensure no collisions
+        @@tmp_environment_set ||= SortedSet.new
+        deadman = 100; loop_num = 0
+        while @@tmp_environment_set.include?(tmp_environment = environment.downcase + '_' + random_string) do
+          break if (loop_num = loop_num + 1) > deadman
+        end
+        @@tmp_environment_set << tmp_environment
+        tmpdir = File.join('','tmp',tmp_environment)
+
+        on master, "mkdir -p #{tmpdir}/manifests #{tmpdir}/modules"
+        # WARNING: this won't work with filesync (symlinked environments are not supported)
+        on master, "ln -sf #{tmpdir} #{File.join(environmentpath,tmp_environment)}"
+        return tmp_environment
+      end
+      module_function :mk_tmp_environment, :environmentpath
+
     end
   end
 end
--- a/acceptance/tests/language/exported_resources.rb
+++ b/acceptance/tests/language/exported_resources.rb
@@ -1,0 +1,170 @@
+test_name "C94788: exported resources using a yaml terminus for storeconfigs" do
+require 'puppet/acceptance/environment_utils'
+extend Puppet::Acceptance::EnvironmentUtils
+
+  app_type = File.basename(__FILE__, '.*')
+  tmp_environment   = mk_tmp_environment(app_type)
+  exported_username = 'er0ck'
+
+  teardown do
+    step 'stop puppet server' do
+      on(master, "service #{master['puppetservice']} stop")
+    end
+    step 'remove cached agent pson catalogs from the master' do
+      on(master, "rm -f #{File.join(master.puppet['yamldir'],'catalog','*')}",
+         :accept_all_exit_codes => true)
+    end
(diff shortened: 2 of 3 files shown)


er0ck:
@branan FYI.  thanks again for the yaml terminus!

acceptance/lib/puppet/acceptance/environment_utils.rb
```diff
@@ -357,6 +357,31 @@ def assert_review(review)
...
+      # if the first test to call this has changed the environmentpath, this will cause trouble
+      #   maybe not the best idea to memoize this?
+      def environmentpath
+        @@memoized_environmentpath ||= master.puppet['environmentpath']
+      end
+
+      # create a tmpdir to hold a temporary environment bound by puppet environment naming restrictions
+      # symbolically link environment into environmentpath
+      def mk_tmp_environment(environment)
+        tmp_environment = environment.downcase + '_' + random_string
```
hlindberg: if you are unlucky you can get the same random name mor than once I think - would that be a problem

acceptance/tests/language/exported_resources.rb
```diff
@@ -0,0 +1,173 @@
...
+        raise ArgumentError.new("You can only save objects that respond to :name") unless request.instance.respond_to?(:name)
+        file = path(request.key)
+        basedir = File.dirname(file)
+        # This is quite likely a bad idea, since we are not managing ownership or modes.
+        Dir.mkdir(basedir) unless Puppet::FileSystem.exist?(basedir)
+        begin
+          #Puppet::Util::Yaml.dump(request.instance, file)
+          # yeah, this is the yaml terminus, but we cannot dump anonymous modules
+          #   and we have some in the catalog (request object)
+          File.open(file, "w") { |f| f.write request.instance.to_pson }
```
hlindberg: maybe rename the terminus to something that clearly states it is an acceptance test mock ?

acceptance/tests/language/exported_resources.rb
```diff
@@ -0,0 +1,173 @@
...
+        puts "storeconfigs filter: \#{request.options[:filter].inspect}"
+        results = Dir.glob(catalog_dir).collect { |file|
+           # yes, i know this is the yaml terminus, but we had to serialize out the catalogs in pson
+           #   see the catalog yaml_storeconfigs temrinus for why (if any)
+          catalog = Puppet::Resource::Catalog.convert_from(:pson, File.read(file))
+          if catalog.name == request.options[:host]
+            next
+          end
+          catalog.resources.select { |resource|
+            puts "resource: \#{resource.inspect}"
```
hlindberg: do you always want this puts, or was  this for debugging?

acceptance/tests/language/exported_resources.rb
```diff
@@ -0,0 +1,173 @@
...
+  ',
+}
+MANIFEST
+    apply_manifest_on(master, manifest, :catch_failures => true)
+  end
+
+  # must specify environment in puppet.conf for it to pickup the terminus code in an environment module
+  #   but we have to bounce the server to pickup the storeconfigs... config anyway
+  # we can't use with_puppet_running_on here because it uses puppet resource to bounce the server
+  #   puppet resource tries to use yaml_storeconfig's path() which doesn't exist and fails back to yaml which indicates an attempted directory traversal and fails.
```
hlindberg: picky: long line

acceptance/tests/language/exported_resources.rb
```diff
@@ -0,0 +1,173 @@
...
+  #   we could implemnt path() properly, but i'm just going to start the server the old fashioned way
```
hlindberg: picky comment typo "implemnt"

acceptance/lib/puppet/acceptance/environment_utils.rb
er0ck: yeah.  i've been mildly worried about this.  we use this method in a few places.  
we could use SecureRandom.hex with more places.  i don't think SecureRandom.uuid will be ok in many places we need a random name, and they take up too much room :-)

hlindberg:
Done with review, my comments are about picky small things, or are questions about intention and a possible problem with the random names. Fine to merge otherwise.

acceptance/tests/language/exported_resources.rb
er0ck: it was for debugging, but it can't hurt?   i could make it more verbose about whom is logging (and/or use our logger).

hlindberg: you are a typical user of this; reading the output from acceptance tests. Do what you think is best.

acceptance/lib/puppet/acceptance/environment_utils.rb
hlindberg: A simple cure is to add them to a Set. If there are duplicates that will get rid of those.

acceptance/tests/language/exported_resources.rb
er0ck: well, i could rename it pson_storeconfigs, and use that as its setting. but i'm not sure how to indicate that it's SaveOnly as the 4 termini for catalog, facts, node, resource indirection are all used based on that setting.  and the resource terminus, for instance, is certainly not SaveOnly.  
the test creates this terminus then removes it from the environment/modulepath, so it's relatively safe.  i suppose i could raise on instances of attempts to read via the catalog terminus?  but i'm not really sure how to do that offhand.
i could look into it, but it probably won't be done for this sprint, give my other work.

er0ck:
updated to address @hlindberg 's concerns. 
push it

hlindberg:
This changes nothing that unit/integration tests will capture. Merging. @er0ck if you want to change to a regular Set you can always do that later.

er0ck:
i figure, if it has to lookup the environment name in every test, it may as well be fast.  so i used a sortedset with rbtree.

PR #4654, opened 2016-02-11
(PUP-5876) Enable static catalogs by default

MikaelSmith:
They will only be generated if code_id is provided when compiling the
catalog.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -144,13 +144,13 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
 
   # Compile the actual catalog.
   def compile(node, options)
-    if node.environment.static_catalogs? && options[:static_catalog]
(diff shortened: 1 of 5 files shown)


lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -144,13 +144,13 @@ def inline_metadata(catalog, checksum_type)
 
   # Compile the actual catalog.
   def compile(node, options)
-    if node.environment.static_catalogs? && options[:static_catalog]
+    if node.environment.static_catalogs? && options[:static_catalog] && options[:code_id]
```
joshcooper: Other parts of the code check for `node.environment` is not nil, so this access may not be safe. The nil check was added a long time ago, and may not be necessary (outside of rspec tests). But seems safer to do the check now.

MikaelSmith: Updated.

joshcooper:
:+1: just waiting on appveyor

PR #4657, opened 2016-02-11
(PUP-5837) Change reverse_each and step functions to return undef when given a block

thallgren:
This PR changes the `step` and `reverse_each` functions so that so that they return the undef value instead of their first argument. This change was made to avoid any confusion on if the returned value is reversed, stepped, or if it somehow contains values returned from the given block.


hlindberg:
Travis reports one test on ruby 1.9.3 as failing. It is a timeout test so I am re-spinning to se if it persists.

PR #4658, opened 2016-02-11
(PUP-5878) acceptance: ensure agent can reconnect after master failure

er0ck:
[skip ci]

### DIFF
--- a/acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
+++ b/acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
@@ -1,0 +1,57 @@
+require 'securerandom'
+test_name 'C97323: agents can reconnect after master crash/restart' do
+
+  username = 'er0ck'
+  teardown do
+    step 'stop puppet agent' do
+      on(agents, puppet_resource("service puppet ensure=stopped"))
+    end
+    step 'remove test user' do
+      on(agent, puppet_resource("user #{username} ensure=absent"))
+    end
+  end
+
+  environmentpath = master.puppet['environmentpath']
+  environment     = master.puppet['environment']
+  step 'create a file resource manifest for checking the puppet runs' do
+    manifest = "user { '#{username}': ensure=>present }"
+    apply_manifest_on(master, manifest, :catch_failures => true)
+    apply_manifest_on(master, <<MANIFEST, :catch_failures => true)
+     file {'#{environmentpath}/#{environment}/manifests/site.pp':
+       content => 'user { "#{username}": ensure=>present }'
+     }
+MANIFEST
+  end
+
+  step 'start puppet master' do
+    with_puppet_running_on(master, {}) do
+      agents.each do |agent|
+        step 'start puppet agent daemon at rapid runinterval' do
+          on(agent, puppet_resource("service puppet ensure=running"))
+        end
+        step "check users list for successful agent run" do
+          on(agent, puppet_resource("user #{username}")).stdout do |result|
+            assert_match(/#{username}/, result, 'did not find evidence of successful puppet run')
+          end
+        end
+        step 'kill master, clear user resource' do
+          on master, "service #{master['puppetservice']} stop"
+          on(agent, puppet_resource("user #{username} ensure=absent"))
+          sleep(10)
+        end
+        step 'check for lack of successful agent run' do
+          on(agent, puppet_resource("user #{username}")).stdout do |result|
+            refute_match(/#{username}/, result, 'found evidence of successful puppet run')
+          end
+        end
+        step 'restart master, assert agent run' do
+          on master, "service #{master['puppetservice']} start"
+          on(agent, puppet_resource("user #{username}")).stdout do |result|
+            assert_match(/#{username}/, result, 'did not find evidence of subsequent successful puppet run')
+          end
+        end
+      end
+    end
+  end
+
+end


acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
```diff
@@ -0,0 +1,50 @@
+test_name 'C97323: agents can reconnect after master crash/restart' do
+
+  logfiles = {}
+  agents.each do |agent|
+    logfiles[agent.hostname] = agent.tmpfile(File.basename(__FILE__, '.*'))
+    on(agent, puppet_resource("user puppet ensure=present"))
```
joshcooper: If you specify `--user agent['user']` then I think (but am not 100% sure), you won't have to create/remove the puppet user. The agent will chown the log file that the agent is running as.

er0ck: good tip.  fixed.

On Thu, Feb 11, 2016 at 10:23 AM, Josh Cooper <notifications@github.com>
wrote:

> In acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
> <https://github.com/puppetlabs/puppet/pull/4658#discussion_r52643520>:
>
> > @@ -0,0 +1,50 @@
> > +test_name 'C97323: agents can reconnect after master crash/restart' do
> > +
> > +  logfiles = {}
> > +  agents.each do |agent|
> > +    logfiles[agent.hostname] = agent.tmpfile(File.basename(__FILE__, '.*'))
> > +    on(agent, puppet_resource("user puppet ensure=present"))
>
> If you specify --user agent['user'] then I think (but am not 100% sure),
> you won't have to create/remove the puppet user. The agent will chown the
> log file that the agent is running as.
>
> —
> <https://github.com/puppetlabs/puppet/pull/4658/files#r52643520>.
>



-- 
eric thompson        QA
(hc) erict     (t) @r0ckdata

acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  step 'start puppet master' do
+    with_puppet_running_on(master, {}) do
+      agents.each do |agent|
+        step 'start puppet agent at rapid runinterval' do
+          on(agent, puppet("agent --server #{master.hostname} --runinterval 10 " +
+                           "--logdest #{logfiles[agent.hostname]} " +
+                           "--user #{agent['user']}"))
```
joshcooper: I don't think it will work for windows agents, because `puppet agent ...` won't daemonize, but will run in the foreground every runinterval seconds. So I'd expect the beaker to block indefinitely while the agent runs.

To run async, you could set the runinterval in settings (`puppet config set ..`), and then use "puppet resource service puppet ensure=running". There also used to be a beaker helper for running an agent async.

I don't know if specifying logdest as a setting will work though, as I think it may require registry modifications to get the service control manager to pass --logdest at agent startup. /cc @Iristyle @glennsarti

er0ck:
boooo 
i really should have tried this in windows before throwing it over the fence.

acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
Iristyle: I think you can instead manually call to SC and pass the params (at least that's what our docs say ;0):

`sc start --server #{master.hostname} --runinterval 10 --logdest #{logfiles[agent.hostname]} --user #{agent['user']}`

Just keep in mind `#{logfiles[agent.hostname]}` should be a Windows style path.

er0ck: > should be a Windows style path.
that's a problem for agent.tmpfile ;)

MikaelSmith:
Still working on it?

er0ck:
@johnduarte, josh mentioned this daemonized behavior might create some issues on AIX.  thoughts?

acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
```diff
@@ -0,0 +1,57 @@
+require 'securerandom'
+test_name 'C97323: agents can reconnect after master crash/restart' do
+
+  username = 'er0ck'
+  teardown do
+    step 'stop puppet agent' do
+      on(agents, puppet_resource("service puppet ensure=stopped"))
+    end
+    step 'remove test user' do
+      on(agent, puppet_resource("user #{username} ensure=absent"))
```
johnduarte: Probably want to allow exit 1 also, in case the user was not successfully created in the body of the test.

johnduarte:
Attempting to run this fails on the master for me:

```
fu5saeucpfn58p4.delivery.puppetlabs.net (fu5saeucpfn58p4.delivery.puppetlabs.net) executed in 1.62 seconds
    Exited: 74

    *************************
    * Dumping master log    *
    *************************

    fu5saeucpfn58p4.delivery.puppetlabs.net (fu5saeucpfn58p4.delivery.puppetlabs.net) 16:15:25$ tail -n 100 /var/log/messages
...
     Feb 17 16:14:15 localhost puppet-master[2134]: Could not change user to puppet. User does not exist and is required to continue.
      Feb 17 16:15:24 localhost puppet-master[2309]: Could not change user to puppet. User does not exist and is required to continue.
```

er0ck:
@johnduarte who is asking for user puppet?  sounds like beaker might be using the wrong user on aix?

acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
```diff
@@ -0,0 +1,57 @@
...
+     }
+MANIFEST
+  end
+
+  step 'start puppet master' do
+    with_puppet_running_on(master, {}) do
+      agents.each do |agent|
+        step 'start puppet agent daemon at rapid runinterval' do
+          on(agent, puppet_resource("service puppet ensure=running"))
+        end
```
joshcooper: Will this use the default 30 min runinterval?

acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
```diff
@@ -0,0 +1,57 @@
...
+        step "check users list for successful agent run" do
+          on(agent, puppet_resource("user #{username}")).stdout do |result|
+            assert_match(/#{username}/, result, 'did not find evidence of successful puppet run')
```
joshcooper: The agent run above happens asynchronously so we can't guarantee it will have completed by the time this check happens.

acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
```diff
@@ -0,0 +1,57 @@
...
+          end
+        end
+        step 'kill master, clear user resource' do
+          on master, "service #{master['puppetservice']} stop"
+          on(agent, puppet_resource("user #{username} ensure=absent"))
```
joshcooper: To test the behavior the agent needs to connect to the master process (completed TCP and SSL handshakes) and send an HTTP request to the server. If the server dies without replying with a TCP RST, then the agent will block indefinitely (before the fix). If you try to use beaker to stop the puppet master while the agent is running, it will most likely send a RST to the agent, invalidating the test. In that case, you'll probably get an EOF error described in PUP-3238.

It might be possible to `kill -9` the server process, but you'd have to do that sometime after the agents sends the HTTP request and before the server replies. I'm not sure how to do that in a beaker test?

MikaelSmith: You'd probably have to get some kind of proxy to simulate a failure. https://github.com/Shopify/toxiproxy looks interesting.

acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
er0ck: yeah, there are a bunch of changes that seem to have gotten lost in some of my rebasing....

acceptance/tests/agent/agent_can_reconnect_after_master_failure.rb
er0ck: the proxy would still require a race to set it up after the agent connects.

this test is much more complicated than at first glance.  i could setup a long compilation, and poll on the master log 
until the agent connects, and kill -9 the server at that time.

er0ck:
closing, as this is not the right approach.  see ticket.

PR #4660, opened 2016-02-11
(maint) Use catalog and resource matchers

joshcooper:
Refactor the inline catalog unit tests to use `manifest_to_catalog` helper and resource matchers. I ran into an issue where generated resources didn't have an ensure value and calling `to_ral` doesn't work when you're on *nix and it's a Windows file path. The ensure value part is fixed in this PR, and I added a pending test for the second part. I recommend using `git log -p --reverse --patience 0dfeecd..` to review the changes.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -165,8 +165,6 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
         end
       end
     end
-
-    catalog
   end
 
   # Compile the actual catalog.
@@ -197,7 +195,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
       str += " in environment #{node.environment}" if node.environment
       benchmark(:notice, str) do
         Puppet::Util::Profiler.profile(str, [:compiler, :static_inline, node.environment, node.name]) do
-          config = inline_metadata(config, checksum_type)
+          inline_metadata(config, checksum_type)
         end
       end
     end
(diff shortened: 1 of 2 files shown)


kylog:
Does the "Set ensure value for generated resources" part merit a PUP ticket? Not sure what the consequences are tbh.

joshcooper:
It was working when I tested against @HAIL9000 PR earlier, but I'm guessing the refactor today regressed on that. I'll add the ticket to the commit summary.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -141,6 +141,7 @@ def inline_metadata(catalog, checksum_type)
 
           # TODO: Conditionally copy owner, group, and mode when we can check if permissions are set
           child_resource = Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
+          child_resource[:ensure] = meta.ftype
```
MikaelSmith: This was in the first version, I didn't get to reviewing the refactor.

MikaelSmith: Is the to_ral call at line 135 safe?

MikaelSmith:
Seems ok :+1:

lib/puppet/indirector/catalog/compiler.rb
joshcooper: we'll need to remove the `to_ral` call later on, and call `Puppet::FileServing::Metadata.indirection.{find,search}` directly, instead of creating a `Puppet::Type::File` instance.

MikaelSmith: Ok, is there another ticket for that?

joshcooper: Not yet, going to file it later. But the pending test shows the issue.

PR #4662, opened 2016-02-12
DO NOT MERGE (PUP-5873) Contain generated resources

joshcooper:
Adding this PR to get feedback about the approach. Note this is based on PR https://github.com/puppetlabs/puppet/pull/4660, which needs to be merged first.

Previously, we copied relationship parameters as strings from the parent
directory to each generated resource. This ensured somewhat accidentally
that generated resources were contained, meaning all of the generated
resources were evaluated before any resource downstream from the parent
directory.

However, if the parent had multiple relationships, we were calling
Array#to_s, causing the catalog to contain:

    "before": "[Notify[hi]{:name=>\"hi\"}, Notify[there]{:name=>\"there\"}]"

This caused the agent run to fail, since there's no resource with that
name.

This commit ensures we copy the relationship object to the generated
resource. It can either be a String or Array depending on the number of
relationship edges.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -135,6 +135,22 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
       file = resource.to_ral
 
       if file.recurse?
+        # memoize all resources that require/subscribe the parent file
+        direct_dependents = { :require => [], :subscribe => []}
+        catalog.resources.find_all do |other|
+          if other.ref != resource.ref
+            [:require, :subscribe].each do |param|
+              # other[param] can be nil, String or Array
+              target_refs = [other[param]].flatten.compact
+              target_refs.each do |target|
+                if target.ref == resource.ref
+                  direct_dependents[param] << other
+                end
+              end
+            end
+          end
+        end
+
         # memoize parent and children resources so we don't have to
         # call `catalog.resource` each time, which generates a
         # Puppet::Resource object.
@@ -184,6 +200,13 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
           parent = resource_table[File.dirname(child.title)]
           parent[:before] ||= []
           parent[:before] << child
+
+          # ensure child is evaluated before dependents that require the parent
+          {:before => :require, :notify => :subscribe}.each_pair do |child_param, dependent_param|
+            if !direct_dependents[dependent_param].empty?
+              child[child_param].concat(direct_dependents[dependent_param])
+            end
+          end
         end
       else
         metadata = file.parameter(:source).metadata
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -156,7 +156,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
         # Puppet::Resource object.
         children = []
         resource_table = {}
-        resource_table[resource.title] = resource
+        resource_table[resource[:path] || resource.title] = resource
 
         file.recurse_remote_metadata.each do |meta|
           # Don't create a new resource for the parent directory
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -208,6 +208,8 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
             end
           end
         end
+        # We're done with the parent, don't recurse again on the agent.
+        resource[:recurse] = false
       else
         metadata = file.parameter(:source).metadata
         raise "Could not get metadata for #{resource[:source]}" unless metadata
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -135,28 +135,56 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
       file = resource.to_ral
 
       if file.recurse?
+        # memoize parent and children resources so we don't have to
+        # call `catalog.resource` each time, which generates a
+        # Puppet::Resource object.
+        children = []
+        resource_table = {}
+        resource_table[resource.title] = resource
+
         file.recurse_remote_metadata.each do |meta|
           # Don't create a new resource for the parent directory
           next if meta.relative_path == "."
 
           # TODO: Conditionally copy owner, group, and mode when we can check if permissions are set
-          child_resource = Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
+          title = File.join(file[:path], meta.relative_path)
+          child_resource = Puppet::Resource.new(:file, title)
           child_resource[:ensure] = meta.ftype
           replace_metadata(child_resource, meta, true)
 
           # Copy parameters from original parent directory
           file.original_parameters.each do |param, value|
             # These should never be passed to our children
+            # REMIND: refactor this with Puppet::Type::File#recurse and skip source_permissions
             unless [:parent, :ensure, :recurse, :recurselimit, :target, :alias, :source].include? param
-              child_resource[param] = value.to_s
+              if [:before, :notify].include? param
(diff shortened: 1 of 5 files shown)


joshcooper:
Given a manifest like:

```puppet
class ntp {
  file { '/tmp/foo':
    ensure => directory,
  }
   
  file { '/tmp/foo/dir':
    ensure  => directory,
    recurse => remote,
    source  => 'puppet:///modules/ntp/dir',
    before  => [Notify['hi'], Notify['there']]
  }

  notify { 'hi': }

  notify { 'there': }
}
```

Compiling the catalogs and running the agent with `--graph` generate the following graphs. The Resource and Relationship graphs are generated (on the agent) after parsing the resource catalog (using the resources and edges in the catalog):

resources

relationship

The agent builds the expanded relationship graph and is what it uses when applying resources. Since the parent directory comes before both `Notify` resources, all of the generated resources come before them as well:

expanded

MikaelSmith:
This doesn't cover auto-require relationships. But I'm not sure auto-requires on generated resources works anyway. /cc @branan

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -141,13 +141,20 @@ def inline_metadata(catalog, checksum_type)
...
           file.original_parameters.each do |param, value|
             # These should never be passed to our children
             unless [:parent, :ensure, :recurse, :recurselimit, :target, :alias, :source].include? param
-              child_resource[param] = value.to_s
+              # This logic belongs in the file type/individual parameters,
+              # e.g. param.copy(child_resource)
+              if [:before, :require, :subscribe, :notify].include? param
+                child_resource[param] = value
+              else
+                child_resource[param] = value.to_s
```
MikaelSmith: Why did we need `to_s` in the first place?

joshcooper:
>  I'm not sure auto-requires on generated resources works anyway

Good question. Auto* relationships work for generated resources in static catalogs, because the resources are in the catalog before we [build_autorelationship_dependencies](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/graph/relationship_graph.rb#L25). But they don't work with `eval_generate` or `generate` resources, since those resources are added to the catalog later.

lib/puppet/indirector/catalog/compiler.rb
joshcooper: Not sure, the [static_compiler](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/catalog/static_compiler.rb#L184) doesn't do it, and the file type passes the value to [newchild](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file.rb#L536-L540), which ends up calling [`resource[param] = value`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type.rb#L1213-L1215). So I'd say we can drop it, but I think @HAIL9000 ran into a problem when not calling `to_s`.

HAIL9000: Actually... I was using `to_s` because I was originally trying to use `file.parameters` rather than `file.original_parameters`. However, it may be the case that because `original_parameters` is just a plain old hash that it will return a string for the value and we may not need to call `to_s`. I haven't actually tested it either way though.

joshcooper:
@branan commented in meatspace that this doesn't handle the case where a non-file resources has a relationship on the recursive parent:

```puppet
file { '/tmp/foo/dir':
  ensure  => directory,
  recurse => remote,
  source  => 'puppet:///modules/ntp/dir',
}

notify { 'hi':
  require => File['/tmp/foo/dir']
}
```

Need to create edge(s) going the other way from each generated resource, e.g. `File['/tmp/foo/dir/file'] -> Notify['hi']` for full containment.

spec/unit/resource/catalog_spec.rb
```diff
@@ -339,6 +340,32 @@
...
+      expect {
+        @catalog.add_resource_before(@one, @two)
+      }.to raise_error(ArgumentError, "Cannot add resource Notify[two] before Notify[one] because Notify[one] is not yet in the catalog")
+    end
+
+    it "adds resources after an existing resource in reverse order" do
+      @catalog.add_resource(@one)
+      @catalog.add_resource_after(@one, @two, @three)
+
+      expect(@catalog.resources).to eq([@one, @three, @two])
```
joshcooper: aside from this PR, no one calls `add_resource_after` with multiple args, so we may want to switch the behavior of that method so it preserves the order of the args you pass to it, e.g. returns `[@one, @two, @three]`

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -135,28 +135,93 @@ def inline_metadata(catalog, checksum_type)
...
+            end
+          end
+        end
+
+        # memoize parent and children resources so we don't have to
+        # call `catalog.resource` each time, which generates a
+        # Puppet::Resource object.
+        children = []
+        resource_table = {}
+        resource_table[resource.title] = resource
```
joshcooper: This doesn't handle the case where resource[:path] is specified

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -135,28 +135,93 @@ def inline_metadata(catalog, checksum_type)
...
+        end
+
+        # Have to add children in reverse order
+        children.reverse.each do |child|
+          catalog.add_resource_after(resource, child)
+
+          # ensure child's immediate ancestor is evaluated before the child
+          parent = resource_table[File.dirname(child.title)]
+          parent[:before] ||= []
+          parent[:before] << child
```
joshcooper: We probably want to alway create `notify` relationships from the recursive parent to each generated child like we do during eval_generate: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction/additional_resource_generator.rb#L80

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -135,28 +135,81 @@ def inline_metadata(catalog, checksum_type)
...
+
+          # ensure child is evaluated before dependents that require the parent
+          {:before => :require, :notify => :subscribe}.each_pair do |child_param, dependent_param|
+            if !direct_dependents[dependent_param].empty?
+              child[child_param].concat(direct_dependents[dependent_param])
```
joshcooper: The recursive parent directory can't have a pre-existing relationship with any of the direct dependents, because that would imply there's a cycle. Since metaparams are dup'ed from the recursive parent to each generated children, it follows that none of the generated children can have pre-existing relationships with any of the direct dependents. So adding the relationship here should be safe, assuming we generated each child at inlining time.

However, if a child is explicitly managed (and takes precedence over the generated one), it could have a preexisting relationship to the direct dependent. So unconditionally `concat`ing could create a cycle. Probably need to only concat dependents that the child doesn't have some sort of relationship with the the `child_param` relationship we're trying to add.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -135,28 +135,81 @@ def inline_metadata(catalog, checksum_type)
...
 
       if file.recurse?
+        # memoize all resources that require/subscribe the parent file
+        direct_dependents = { :require => [], :subscribe => []}
+        catalog.resources.find_all do |other|
+          if other.ref != resource.ref
+            [:require, :subscribe].each do |param|
+              # other[param] can be nil, String or Array
+              target_refs = [other[param]].flatten.compact
+              target_refs.each do |target|
```
joshcooper: N-1 resources x 2 metaparams x M target refs where M varies, but is probably 0-2 for most resources.

joshcooper: Actually, this loop happens inside the the `catalog.resources.find_all`, so `O(n^2)`, not good. We should be able to precompute all of the "other" resources that have require/subscribe relations on recursive file resources, so we only scan the catalog once.

spec/unit/resource/catalog_spec.rb
MikaelSmith: Seems like a useful guarantee.

joshcooper:
We are going to handle this a different way. See PUP-5114

PR #4664, opened 2016-02-12
(PUP-5892) Alias types are presented with the wrong name in type mismatches

thallgren:
This PR fixes three things.

1. The PTypeAlias is renamed to PTypeAliasType
2. The PTypeReference is renamed to PTypeReferenceType (didn't deserve an issue of its own)
3. The type mismatch describer reports type aliases by alias name instead of type name


hlindberg:
It lacks a test that the mismatch describer is doing the right thing for the example given in the ticket.
If you do update the last commit with such a test, also fix the typo "henc" in the commit message.

The rest is fine.

Thank you very much for the excellent split in multiple commits. It made it very easy to review this.

thallgren:
I replaced the last commit with one that contains the requested test and a fix for the typo in the commit message.

PR #4665, opened 2016-02-12
(PUP-5879) Use FileSystem.read where applicable to read JSON, settings and other files as UTF-8

Iristyle:
Adds additional UTF-8 support to Puppet when reading files from disk, including:

* Settings like `puppet.conf`
* PMT ignore files - `.gitignore` and `.pmtignore`
* Hiera YAML and JSON files
* Modules `metadata.json`
* Route auth files
* MsgPack

### DIFF
--- a/lib/puppet/module.rb
+++ b/lib/puppet/module.rb
@@ -48,7 +48,7 @@ class Puppet::Module
 
   def self.is_module_namespaced_name?(name)
     # it must match the full module name according to forge validator
-    return true if name =~ /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/   
+    return true if name =~ /^[a-zA-Z0-9]+[-][a-z][a-z0-9_]*$/
     return false
   end
 
@@ -90,7 +90,7 @@ class Puppet::Module
     return false unless Puppet::FileSystem.exist?(metadata_file)
 
     begin
-      metadata =  JSON.parse(File.read(metadata_file))
+      metadata =  JSON.parse(File.read(metadata_file, :encoding => 'utf-8'))
     rescue JSON::JSONError => e
       Puppet.debug("#{name} has an invalid and unparsable metadata.json file.  The parse error: #{e.message}")
       return false
@@ -145,7 +145,7 @@ class Puppet::Module
   end
 
   def load_metadata
-    @metadata = data = JSON.parse(File.read(metadata_file))
+    @metadata = data = JSON.parse(File.read(metadata_file, :encoding => 'utf-8'))
     @forge_name = data['name'].gsub('-', '/') if data['name']
 
     [:source, :author, :version, :license, :dependencies].each do |attr|
--- a/spec/unit/module_spec.rb
+++ b/spec/unit/module_spec.rb
@@ -609,21 +609,21 @@ describe Puppet::Module do
 
   it "should have metadata if it has a metadata file and its data is not empty" do
     Puppet::FileSystem.expects(:exist?).with(@module.metadata_file).returns true
-    File.stubs(:read).with(@module.metadata_file).returns "{\"foo\" : \"bar\"}"
+    File.stubs(:read).with(@module.metadata_file, {:encoding => 'utf-8'}).returns "{\"foo\" : \"bar\"}"
 
     expect(@module).to be_has_metadata
   end
 
   it "should have metadata if it has a metadata file and its data is not empty" do
     Puppet::FileSystem.expects(:exist?).with(@module.metadata_file).returns true
-    File.stubs(:read).with(@module.metadata_file).returns "{\"foo\" : \"bar\"}"
+    File.stubs(:read).with(@module.metadata_file, {:encoding => 'utf-8'}).returns "{\"foo\" : \"bar\"}"
 
     expect(@module).to be_has_metadata
   end
 
   it "should not have metadata if has a metadata file and its data is empty" do
     Puppet::FileSystem.expects(:exist?).with(@module.metadata_file).returns true
-    File.stubs(:read).with(@module.metadata_file).returns "This is some invalid json.\n"
+    File.stubs(:read).with(@module.metadata_file, {:encoding => 'utf-8'}).returns "This is some invalid json.\n"
     expect(@module).not_to be_has_metadata
   end
 
@@ -646,7 +646,7 @@ describe Puppet::Module do
 
   it "should tolerate failure to parse" do
     Puppet::FileSystem.expects(:exist?).with(@module.metadata_file).returns true
-    File.stubs(:read).with(@module.metadata_file).returns(my_fixture('trailing-comma.json'))
+    File.stubs(:read).with(@module.metadata_file, {:encoding => 'utf-8'}).returns(my_fixture('trailing-comma.json'))
 
     expect(@module.has_metadata?).to be_falsey
   end
@@ -656,7 +656,7 @@ describe Puppet::Module do
 
     mod = Puppet::Module.new("foo", "/path", mock("env"))
     mod.stubs(:metadata_file).returns "/my/file"
-    File.stubs(:read).with("/my/file").returns text
+    File.stubs(:read).with("/my/file", {:encoding => 'utf-8'}).returns text
     mod
   end
 
@@ -681,12 +681,36 @@ describe Puppet::Module do
       it "should fail if #{attr} is not present in the metadata file" do
         @data.delete(attr.to_sym)
         @text = @data.to_pson
-        File.stubs(:read).with("/my/file").returns @text
+        File.stubs(:read).with("/my/file", {:encoding => 'utf-8'}).returns @text
         expect { @module.load_metadata }.to raise_error(
           Puppet::Module::MissingMetadata,
           "No #{attr} module metadata provided for foo"
         )
       end
+    end
+  end
+
+  describe "when loading the metadata file from disk" do
+    it "should properly parse utf-8 contents" do
+      rune_utf8 = "\u16A0\u16C7\u16BB" # ᚠᛇᚻ
+      metadata_json = tmpfile('metadata.json')
+      File.open(metadata_json, 'w') do |file|
+        file.puts <<-EOF
+  {
+    "license" : "GPL2",
+    "author" : "#{rune_utf8}",
+    "version" : "1.0",
+    "source" : "http://foo/",
+    "dependencies" : []
+  }
+        EOF
+      end
+
+      mod = Puppet::Module.new('foo', '/path', mock('env'))
+      mod.stubs(:metadata_file).returns metadata_json
+
+      mod.load_metadata
+      expect(mod.author).to eq(rune_utf8)
     end
   end
 
--- a/spec/unit/pops/loaders/loaders_spec.rb
+++ b/spec/unit/pops/loaders/loaders_spec.rb
@@ -107,7 +107,7 @@ describe 'loaders' do
     end
 
     it 'loader allows loading a function more than once' do
-      File.stubs(:read).with(user_metadata_path).returns ''
+      File.stubs(:read).with(user_metadata_path, {:encoding => 'utf-8'}).returns ''
 
       env = environment_for(File.join(dependent_modules_with_metadata, 'modules'))
       loaders = Puppet::Pops::Loaders.new(env)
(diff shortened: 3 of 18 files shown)


spec/unit/pops/loaders/loaders_spec.rb
```diff
@@ -107,7 +107,7 @@
     end
 
     it 'loader allows loading a function more than once' do
-      File.stubs(:read).with(user_metadata_path).returns ''
+      Puppet::FileSystem.stubs(:read).with(user_metadata_path, {:encoding => 'utf-8'}).returns ''
```
Iristyle: @hlindberg this PR has triggered a number of test failures in `loaders_spec.rb` that I haven't had time to fully chase down yet.. wondering if you quickly have an idea on how to address the test failures.  If you don't know OTOH, no big deal - I'll eventually track them down, but thought it might be worth checking before I get to that point.

hlindberg: @Iristyle You are changing from `File` to `FileSystem` - does it work if you change that?

spec/unit/pops/loaders/loaders_spec.rb
```diff
@@ -234,15 +234,15 @@ def compile_and_get_notifications(code)
...
     it "an autoloaded type can reference an autoloaded type alias from another module when dependency is present in metadata.json" do
-      File.stubs(:read).with(user_metadata_path).returns user_metadata.merge('dependencies' => [ { 'name' => 'test-usee'} ]).to_pson
+      Puppet::FileSystem.stubs(:read).with(user_metadata_path, {:encoding => 'utf-8'}).returns user_metadata.merge('dependencies' => [ { 'name' => 'test-usee'} ]).to_pson
       expect(eval_and_collect_notices(<<-CODE, node)).to eq(['ok'])
         assert_type(User::WithUseeZero, [0])
         notice(ok)
       CODE
     end
 
     it "an autoloaded type can reference an autoloaded type alias from another module when other module has it declared in init.pp" do
```
glennsarti: This seems to be failing because the stubbing is firing for other files e.g. `<puppet dir> spec/fixtures/unit/pops/loaders/loaders/dependent_modules_with_metadata/environment.conf`

This is because we used to stub File.read and call PuppetFileSystem in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/settings.rb#L1194 .  But now that you're stubbing Puppet::Filesystem.read, it's not working properly.

ferventcoder: I'm trying to understand what unit this entire class refers to. It appears it is more of a whitebox integration type test.

ferventcoder: @hlindberg From what I'm seeing this doesn't really appear to testing a unit within Puppet code, more of an integration of several pieces of Puppet working together. Correct me if I'm wrong here. 

So mocking isn't really working for something that tests multiple parts of Puppet and not a singular unit. Some/all of these specs in this class should likely be moved over to integration and the mocking bits removed.

ferventcoder: The mocking here is a bit fragile, as we're trying to bring more of Puppet under using the FileSystem abstraction we are finding it doesn't work out well with this spec due to mocking not really working for something that tests multiple parts of Puppet and not a singular unit. Some/all of these specs in this class should likely be moved over to integration and the mocking bits removed.

hlindberg: It is unit testing that it has the ability to reference things elsewhere that is should have visibility into. An integration test would require laying out the involved modules, the type aliases in separate files etc. and use no mocking. The reason it is using a function call, and a type is to make the test compact and readable - it would be quite long to do exactly the same in ruby. The same style is used in many places throughout our code base - i.e. using a compilation of puppet code to trigger the exact sequence of calls required to unit test the thing being tested.

ferventcoder: It's likely a religious battle on what constitutes a unit. Let's side step that for a moment. 

> An integration test would require laying out the involved modules, the type aliases in separate files etc. and use no mocking.

Terminology is hard, "integration" is overloaded.  It depends on your definition of "integration testing" - I specifically used the wording "white box integration" (apologies though, it was only in the statement prior to your mention). White box integration means you can still reach into the system and put certain pieces together, use test doubles for other areas of code, and test how interactions work within those integrated parts of the system under test, which is what I'm seeing here in these tests. 

What I'm hearing by your use of "integration" seems more black box integration, acceptance testing and/or end to end testing?

`spec/unit` versus `spec/integration` in this case.

> The same style is used in many places throughout our code base

Agreed, but those other places in the code base may also fall under white box integration and thus are only relevant to the current discussion in that they may also could be subject to a similar discussion eventually.

ferventcoder: My statement about "The mocking here is a bit fragile, as we're trying to bring more of Puppet under using the FileSystem abstraction" is more the fault of mocha's inability to handle the use case of wanting to mock only certain calls to a method and pass through the rest.

Any ideas about how we can better move everything to the FileSystem abstraction without running into these issues?

hlindberg: Ping @thallgren see above for a test that uses mocking in a way that is hard to work around. Any ideas how to rewrite it to make it easier to use the FileSystem abstraction?

thallgren: Sorry @hlindberg, the only thing I can think of is to exercise a smaller unit of code to get rid of other side effects. Perhaps test the Puppet.settings, rather than doing a full compile. Not sure if that's meaningful though.

Iristyle:
Peculiar unrelated registry spec error on Windows:

```
Failures:
 
  1) Puppet::Util::Windows::Registry#values when reading values should return REG_MULTI_SZ values
     Failure/Error: expect(written).to eq(pair[:value])
 
       expected: ["string1", "string2"]
            got: ["string1", "string2", "\u6F00"]
 
       (compared using ==)
     # ./spec/integration/util/windows/registry_spec.rb:222:in `block (6 levels) in <top (required)>'
     # ./spec/integration/util/windows/registry_spec.rb:213:in `block (5 levels) in <top (required)>'
 
Finished in 21 minutes 2 seconds (files took 16.53 seconds to load)
20166 examples, 1 failure, 62 pending
 
Failed examples:
 
rspec './spec/integration/util/windows/registry_spec.rb[1:3:5:3]' # Puppet::Util::Windows::Registry#values when reading values should return REG_MULTI_SZ values
```

glennsarti:
@Iristyle We picked that up in PUP-5993

spec/unit/pops/loaders/loaders_spec.rb
Iristyle: The simplest thing to do at this point was to *not* change the 2 lines of code in `lib/puppet/module.rb` to use `Puppet::FileSystem.read` instead of `File.read`.  Instead, I updated the `File.read` calls to supply the encoding, and mostly left the mocking alone.

Probably something that should be fixed at a later date, but not totally pressing considering we want to get these changes merged in today.

Iristyle:
#4898 is the next PR in this series

lib/puppet/module.rb
```diff
@@ -90,7 +90,7 @@ def has_metadata?
     return false unless Puppet::FileSystem.exist?(metadata_file)
 
     begin
-      metadata =  JSON.parse(File.read(metadata_file))
+      metadata =  JSON.parse(File.read(metadata_file, :encoding => 'utf-8'))
```
joshcooper: This makes me nervous. The metadata file is originally constructed by whomever published the module, eg on windows. So that file could be encoded using the windows active code page. The user could also have published the module awhile ago, e.g. using puppet 3.7, using PMT that didn't enforce utf-8 encoding when writing the metadata.

The user install it on a *nix master using puppet 4.5. The master then tries to load the module and parse the metadata, and it raises an exception because the metadata contents are not utf-8 encoded.

Do we specify somewhere that module metadata must be utf-8 encoded? Should we fall back to the old behavior if a specific encoding-related exception is raised?

Iristyle: Fair point and thanks for bringing it up.  A few counterpoints / ideas:

* JSON is specified as Unicode in [RFC 4627](https://www.ietf.org/rfc/rfc4627.txt), with the default being UTF-8.  The Encoding section of the spec basically states "look at the BOM" to be certain, but only declares UTF-8, UTF-16LE, UTF-16BE, UTF32-LE and UTF32-BE as valid encodings.
* That said, it doesn't mean users creating files with a text editor are necessarily conforming to the spec.
* From a pragmatic standpoint, I think the likelihood that there are characters in metadata using other encodings, that can't parse as UTF-8 is unlikely. I'm not sure how likely it might be to corrupt the payload - for instance:
  * with codepages that might trigger a multibyte character sequence in UTF-8, thereby throwing off the rest of the file parsing
  * with codepages using completely invalid UTF-8 byte sequences (there are some tix in JIRA where this happened trying to parse control sequences from linux command line tool output)
  * codepages that have characters like `"`, `{` or other crucial JSON constructs at different positions than ASCII or UTF-8
  * that is, I believe it's more common to see `Invalid Byte Sequence` problems when going to `US-ASCII` or from `UTF-16LE` / `UTF-8` to a local codepage
* I think the biggest concern IMHO is if the file was written on Windows as `UCS-2` and ends up being read as `UTF-8`, which typically introduces characters spaced between NULLs due to the extra bytes - which should show up as a parsing error when trying to treat it as JSON.
* I don't think this is actually a behavioral change for the master (which *should* be using UTF-8 by default anyhow), but it is a change for agents using PMT I believe.  In other words, this makes the `.read` explicit - but I don't believe Ruby would have parsed `UCS-2` correctly previously anyhow.  We can certainly validate that to make 100% sure.

So I think your point about being more defensive might be a good one, since we have no control over module payloads... though I wonder if the best spot to nip problems in the bud is during a `module build` when the `.tar.gz` is produced ... or upon upload to the forge (I think that `module validate` already does this, but not 100% certain).

So I think we can do 2 things to test this a bit more:
* Create a module with a `UCS-2` metadata.json and install it on masters / agents prior to this change / with this change.. to verify.
* Systematically install every module from the forge given the local forge backup ... and see what happens.

spec/unit/module_spec.rb
```diff
@@ -690,6 +690,30 @@ def a_module_with_metadata(data)
     end
   end
 
+  describe "when loading the metadata file from disk" do
+    it "should properly parse utf-8 contents" do
```
Iristyle: @joshcooper I created some real files on disk in `UTF-16LE` and `UTF-16LE` with BOM (which should approximate `UCS-2`) and added a couple more similar tests locally. Since Ruby does no detection when opening files to read them, the results were what I expected (against both the previous and updated code) - unparseable JSON.  So no change in behavior there:

```
1) Puppet::Module when loading the metadata file from disk should properly parse utf-16le contents
     Failure/Error: @metadata = data = JSON.parse(File.read(metadata_file))

     JSON::ParserError:
       757: unexpected token at '{'
     # ./.bundle/gems/ruby/2.1.0/gems/json-1.8.3/lib/json/common.rb:155:in `parse'
     # ./.bundle/gems/ruby/2.1.0/gems/json-1.8.3/lib/json/common.rb:155:in `parse'
     # ./lib/puppet/module.rb:151:in `load_metadata'
     # ./spec/unit/module_spec.rb:698:in `block (3 levels) in <top (required)>'

  2) Puppet::Module when loading the metadata file from disk should properly parse utf-16le with BOM contents
     Failure/Error: @metadata = data = JSON.parse(File.read(metadata_file))

     JSON::ParserError:
       757: unexpected token at '??{'
     # ./.bundle/gems/ruby/2.1.0/gems/json-1.8.3/lib/json/common.rb:155:in `parse'
     # ./.bundle/gems/ruby/2.1.0/gems/json-1.8.3/lib/json/common.rb:155:in `parse'
     # ./lib/puppet/module.rb:151:in `load_metadata'
     # ./spec/unit/module_spec.rb:706:in `block (3 levels) in <top (required)>'
```

I will try to construct some other sample files to verify old / new behavior.

spec/unit/pops/loaders/loaders_spec.rb
ferventcoder: > the only thing I can think of is to exercise a smaller unit of code to get rid of other side effects.

@thallgren that's what I was thinking, and what led me to state that these tests are more of white box integration type tests.

PR #4666, opened 2016-02-12
(PUP-5122) Add acceptance test for direct puppet running from cache

whopper:
This commit adds the skeleton of an acceptance test for the scenario
where the server tells the agent to run from its cached catalog and
remediates local drift to file content.

Once further work has been done on actually retrieving file content
using content_uri (PUP-5114), this test will be expanded.

### DIFF
--- a/acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
+++ b/acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
@@ -1,0 +1,114 @@
+test_name "PUP-5122: Puppet remediates local drift using code_id and content_uri" do
+  basedir = master.tmpdir(File.basename(__FILE__, '.*'))
+  module_dir = "#{basedir}/environments/production/modules"
+  modulepath = "#{module_dir}"
+  agent_test_file_path = agent.tmpfile('foo_file')
+
+  master_opts = {
+   'master' => {
+      'static_catalogs' => true
+    },
+   'main' => {
+      'environmentpath' => "#{basedir}/environments"
+    }
+  }
+
+  step "Add versioned-code parameters to puppetserver.conf and ensure the server is running" do
+    puppetserver_config = "#{master['puppetserver-confdir']}/puppetserver.conf"
+    versioned_code_settings = {"code-id-command" => "#{basedir}/code_id.sh", "code-content-command" => "#{basedir}/code_content.sh"}
+
+    modify_tk_config(master, puppetserver_config, versioned_code_settings)
+    on(master, puppet_resource('service', 'puppetserver', 'ensure=running'))
+  end
+
+  step "Create a module and a file with content representing the first code_id version, and Setup code-id-command and code-content-command scripts" do
+    code_id_command = <<EOF
+    #! /bin/sh
+
+    echo 'code_version_1'
+EOF
+
+    code_content_command = <<EOF
+    #! /bin/sh
+
+    if [ \\\$2 == 'code_version_1' ] ; then
+      echo 'code_version_1'
+    else
+      echo 'newer_code_version'
+    fi
+EOF
+
+    apply_manifest_on(master, <<MANIFEST, :catch_failures => true)
+    File {
+      ensure => directory,
+      mode => "0750",
+      owner => #{master.puppet['user']},
+      group => #{master.puppet['group']},
+    }
+
+    file {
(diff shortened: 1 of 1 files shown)


whopper:
Update: passed on Windows.

Strange note: originally I was running `puppet agent -t --static_catalogs...` (which is unnecessary). This didn't affect the run in Linux, but in Windows resulted in `Error: Could not parse application options: invalid option: --static_catalogs`. I used the same SHA and SUITE_VERSION on both platforms. I don't think this matters, but I thought I'd point it out!

er0ck:
created a ticket for the non-exposed --static-catalogs option:
https://tickets.puppetlabs.com/browse/PUP-5896

acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
```diff
@@ -0,0 +1,112 @@
+test_name "PUP-5122: Puppet remediates local drift using code_id and content_uri"
```
er0ck: FYI @ all
test_names and steps should all have blocks (do end most of the time), moving forward.  this helps beaker with indentation and beaker.next with reporting and other user features.

acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
```diff
@@ -0,0 +1,112 @@
...
+
+step "Add versioned-code parameters to puppetserver.conf and ensure the server is running"
+basedir = master.tmpdir("master_basedir")
+puppetserver_config = "#{master['puppetserver-confdir']}/puppetserver.conf"
+versioned_code_settings = {"code-id-command" => "#{basedir}/code_id.sh", "code-content-command" => "#{basedir}/code_content.sh"}
+
+modify_tk_config(master, puppetserver_config, versioned_code_settings)
```
er0ck: coooooool

acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
whopper: @er0ck ah, good to know! Updating now.

acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
```diff
@@ -0,0 +1,112 @@
+test_name "PUP-5122: Puppet remediates local drift using code_id and content_uri"
+
+step "Add versioned-code parameters to puppetserver.conf and ensure the server is running"
+basedir = master.tmpdir("master_basedir")
```
er0ck: to make these more known, i sometimes send `File.basename(__FILE__, '.*')` as the tmpdir name.
each run is unique, but it's known what test they are for, and they can just be left hanging around for debug

whopper:
@er0ck updated steps and test_name with blocks

acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
```diff
@@ -0,0 +1,114 @@
...
+      ensure => file,
+      path => "#{module_dir}/foo/files/foo.txt",
+      content => "code_version_1",
+      mode => "0640",
+    }
+
+    file { '#{basedir}/code_id.sh':
+      ensure => file,
+      content => "#{code_id_command}",
+      mode => "0700",
```
joshcooper: This assumes the user that puppet is running as (via beaker) is the same as what `puppetserver` is running as. I'm not sure if that'll be an issue for PE and pe-puppet? May want to relax to 0755.

acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    file { '#{basedir}/code_content.sh':
+      ensure => file,
+      content => "#{code_content_command}",
+      mode => "0700",
```
joshcooper: same comment

whopper:
@joshcooper ah, makes sense. Updated with relaxes perms.

joshcooper:
:+1:

joshcooper:
I made a note in https://tickets.puppetlabs.com/browse/PUP-5114 that we'll want to update this test once the agent is using content_uri to retrieve versioned file content.

PR #4667, opened 2016-02-13
(PUP-5897) Add systemd as a default provider on Ubuntu 16.04.

nightfly19:
SSIA


kylog:
Thanks @nightfly19! This will be released in puppet 4.4.

A side note though: I don't know if we'll have 16.04 *packages* in that same timeframe, as that's a bit more work.

PR #4669, opened 2016-02-13
(PUP-5697) Inline file permissions to the static catalog

MikaelSmith:
Inlines `owner`, `group`, and `mode` permissions when
`source_permissions` is `use`. Not inlined when it's `ignore`, and
raises an error if `use_when_creating` is requested.

For Windows, modified how we assert that `source_permissions => use`
will not be used on Windows so it's still checked for the static
catalog. Check is done client-side, as the master usually doesn't care
what OS the client is.

Also added a maint change to always set :ensure and :checksum to prompt discussion.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -84,24 +84,31 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
   end
 
   # Rewrite a given file resource with the metadata from a fileserver based file
-  def replace_metadata(resource, metadata, recurse = false)
+  def replace_metadata(resource, metadata)
+    # TODO: when we stop using to_ral, we'll have to add a case for ensure => <path>
+    # implying a link.
     if resource[:links] == "manage" && metadata.ftype == "link"
       resource.delete(:source)
-      resource[:ensure] = "link"
       resource[:target] = metadata.destination
-    elsif resource[:links] == "follow"
-      resource[:ensure] = metadata.ftype
     end
+
+    # Most of the time, ensure should match the returned metadata. Exceptions are when
+    # ensure is `present` (it could be anything), or a path (implies creating a link).
+    # `to_ral` caused the path case to munge to managing links, so we warn if there's
+    # a mismatch that's not `present`.
+    if resource[:ensure] != :present && resource[:ensure] != metadata.ftype
+      Puppet.warning "#{resource} expected to create a #{resource[:ensure]}, but its source "+
+        "is a #{metadata.ftype}. The source metadata will override the 'ensure' property."
+    end
+
+    resource[:ensure] = metadata.ftype
 
     if metadata.ftype == "file"
+      # Since we use the checksum from metadata, it makes sense to pick up the checksum type
+      # from the same source. This should never differ, but seems like the most correct
+      # expression of intent.
       resource[:checksum_value] = sumdata(metadata.checksum)
-    end
-
-    if recurse
-      resource[:source] = metadata.source
-      if metadata.ftype == "file"
-        resource[:checksum] = metadata.checksum_type
-      end
+      resource[:checksum] = metadata.checksum_type
     end
   end
 
@@ -141,8 +148,8 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
 
           # TODO: Conditionally copy owner, group, and mode when we can check if permissions are set
           child_resource = Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
-          child_resource[:ensure] = meta.ftype
-          replace_metadata(child_resource, meta, true)
+          child_resource[:source] = meta.source
+          replace_metadata(child_resource, meta)
 
           # Copy parameters from original parent directory
           file.original_parameters.each do |param, value|
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -96,7 +96,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
     # ensure is `present` (it could be anything), or a path (implies creating a link).
     # `to_ral` caused the path case to munge to managing links, so we warn if there's
     # a mismatch that's not `present`.
-    if resource[:ensure] != :present && resource[:ensure] != metadata.ftype
+    if resource[:ensure] != "present" && resource[:ensure] != metadata.ftype
       Puppet.warning "#{resource} expected to create a #{resource[:ensure]}, but its source "+
         "is a #{metadata.ftype}. The source metadata will override the 'ensure' property."
     end
@@ -109,6 +109,19 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
       # expression of intent.
       resource[:checksum_value] = sumdata(metadata.checksum)
       resource[:checksum] = metadata.checksum_type
+    end
+
+    case resource[:source_permissions]
+    when "use"
+      resource[:owner] ||= metadata.owner
+      resource[:group] ||= metadata.group
+      if resource[:mode].nil?
+        mode = metadata.mode
+        mode = mode.to_s(8) if mode.is_a?(Numeric)
+        resource[:mode] = mode
+      end
+    when "use_when_creating"
+      raise Puppet::Error, "source_permissions => use_when_creating cannot be set when creating static catalogs"
     end
   end
 
@@ -146,7 +159,6 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
           # Don't create a new resource for the parent directory
           next if meta.relative_path == "."
 
-          # TODO: Conditionally copy owner, group, and mode when we can check if permissions are set
           child_resource = Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
           child_resource[:source] = meta.source
           replace_metadata(child_resource, meta)
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -727,6 +727,24 @@ Puppet::Type.newtype(:file) do
   end
 
   def retrieve
+    # This check is done in retrieve to ensure it happens before we try to use
+    # metadata in `copy_source_values`, but so it only fails the resource and not
+    # catalog validation (because that would be a breaking change from Puppet 4).
+    if Puppet.features.microsoft_windows? && parameter(:source) &&
+      [:use, :use_when_creating].include?(self[:source_permissions])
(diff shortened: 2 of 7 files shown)


lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -84,24 +84,30 @@ def add_node_data(node)
...
+  def replace_metadata(resource, metadata)
     if resource[:links] == "manage" && metadata.ftype == "link"
       resource.delete(:source)
-      resource[:ensure] = "link"
       resource[:target] = metadata.destination
-    elsif resource[:links] == "follow"
-      resource[:ensure] = metadata.ftype
     end
 
+    resource[:ensure] = metadata.ftype
```
MikaelSmith: The 3 values for ftype are `link`, `file`, and `directory`. Previously we updated if `links => manage` or `links => follow`; the third option is `links => ignore`.

The cases where this would now update, but hadn't before:
- request a link but get a file or directory
- request a file but get a directory
- request a directory but get a file

The `#copy_from_source` method in the `source` parameter would do those things. If we don't do them in the static catalog, I'm not sure we get correct behavior, since we skip `#copy_from_source` if `checksum_value` is set. I need to investigate.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -84,24 +84,30 @@ def add_node_data(node)
...
+
     if metadata.ftype == "file"
       resource[:checksum_value] = sumdata(metadata.checksum)
+      resource[:checksum] = metadata.checksum_type
```
MikaelSmith: This was another piece pulled from `#copy_source_values`.

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: We definitely don't get the correct behavior; if the resource specifies a directory but the source is a file, the static catalog's behavior without this PR would cause a directory to be created.

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: This could also be a devfail if they ever don't match.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -84,24 +84,30 @@ def add_node_data(node)
...
+    when "use"
+      resource[:owner] = metadata.owner if (resource[:owner].nil? || resource[:owner] == :absent)
+      resource[:group] = metadata.group if (resource[:group].nil? || resource[:group] == :absent)
+      if resource[:mode].nil? || resource[:mode] == :absent
+        mode = metadata.mode
+        mode = mode.to_s(8) if mode.is_a?(Numeric)
+        resource[:mode] = mode
       end
+    when "use_when_creating"
+      raise Puppet::Error, "source_permissions => use_when_creating cannot be set when creating static catalogs"
```
MikaelSmith: Oh, apparently I already did it.

branan: Sweet. Ship it

lib/puppet/type/file/source.rb
```diff
@@ -250,6 +235,10 @@ def found?
     def copy_source_value(metadata_method)
       param_name = (metadata_method == :checksum) ? :content : metadata_method
       if resource[param_name].nil? or resource[param_name] == :absent
+        if Puppet.features.microsoft_windows? && [:owner, :group, :mode].include?(metadata_method)
+          devfail "Should not have tried to use source owner/mode/group on Windows"
+        end
+
```
joshcooper: I think we previously would allow copying of mode from a local source, e.g. `source => 'file:///C:/tmp/foo'`, but it looks like a hard failure now?

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -84,24 +84,44 @@ def add_node_data(node)
...
 
-    if recurse
-      resource[:source] = metadata.source
-      if metadata.ftype == "file"
-        resource[:checksum] = metadata.checksum_type
+    case resource[:source_permissions]
+    when "use"
+      resource[:owner] = metadata.owner if (resource[:owner].nil? || resource[:owner] == :absent)
+      resource[:group] = metadata.group if (resource[:group].nil? || resource[:group] == :absent)
+      if resource[:mode].nil? || resource[:mode] == :absent
```
joshcooper: Why set owner/group/mode if `resource[:x] == :absent`?

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -84,24 +84,44 @@ def add_node_data(node)
...
     end
 
-    if recurse
-      resource[:source] = metadata.source
-      if metadata.ftype == "file"
-        resource[:checksum] = metadata.checksum_type
+    case resource[:source_permissions]
+    when "use"
+      resource[:owner] = metadata.owner if (resource[:owner].nil? || resource[:owner] == :absent)
+      resource[:group] = metadata.group if (resource[:group].nil? || resource[:group] == :absent)
```
joshcooper: `metadata.owner/group` will always be a String on Windows, but I think it returns uid/gid on POSIX. I did a spike awhile ago in: https://github.com/joshcooper/puppet/commit/81d4f41f2f19a8a371f9f5c13cd67b1005b24c3a#diff-05b3e3b37c22ea04514d6ea8e1affeadR20 and https://github.com/joshcooper/puppet/commit/81d4f41f2f19a8a371f9f5c13cd67b1005b24c3a#diff-8776b2626abd15d25f6507b7ed2ff23bR65. IIRC, I think the server would happily serialize file_metadata response containing numerics, and the agent would deserialize them to strings... But since we're inlining these, I think they need to be quoted.

lib/puppet/type/file.rb
```diff
@@ -393,6 +393,22 @@ def self.title_patterns
 
     self.warning "Checksum value is ignored unless content or source are specified" if self[:checksum_value] && !self[:content] && !self[:source]
 
+    if Puppet.features.microsoft_windows? && parameter(:source) &&
+      [:use, :use_when_creating].include?(self[:source_permissions])
+      err_msg = "Copying owner/mode/group from the source file on Windows" <<
+      " is not supported; use source_permissions => ignore."
+
+      if self[:owner] == nil || self[:owner] == :absent || self[:group] == nil ||
+        self[:group] == :absent || self[:mode] == nil || self[:mode] == :absent
```
joshcooper: Is `nil` returned if the property is not managed and `absent` returned if the resource doesn't exist?

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: Ok, quoting seems reasonable.

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: I took that from https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/source.rb#L252, but since we're operating on a Resource it would never be `:absent`. I'll remove that check.

lib/puppet/type/file.rb
MikaelSmith: The only comment I saw was https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/mode.rb#L165-L166. This may not be necessary, but `source_permissions => use` will cause `:absent` to be replaced by the source. Maybe it makes sense to keep handling of `:absent` (skip updating it on Windows) in `source.rb`.

lib/puppet/type/file/source.rb
MikaelSmith: I'm not sure that was allowed. I think the failure above would've precluded it. I'll try it out.

lib/puppet/type/file.rb
joshcooper: Ah the audit comment is referring to the [`audit` metaparam](https://docs.puppetlabs.com/puppet/latest/reference/metaparameter.html#audit). The behavior may be triggered by setting the `audit` metaparam and using `puppet inspect`...

MikaelSmith: `puppet inspect` relies on the cached catalog, so I'm not sure how you would ever get a cached catalog that has `source_permissions => use` and mode/owner/group marked absent. I think we can skip checking `:absent`, and rely on the devfail.

joshcooper: :+1:

MikaelSmith: However, moving the check here changes the behavior: instead of `source_permissions => use` failing the resource, it now invalidates the catalog.

MikaelSmith: Created https://tickets.puppetlabs.com/browse/PUP-5919 to review the failure behavior later. For now I've implemented it to preserve the prior behavior.

Also, I can't find a way that owner/group/mode get set to absent, so I removed those checks.

lib/puppet/type/file/source.rb
MikaelSmith: Added https://tickets.puppetlabs.com/browse/PUP-5918 to fix this later.

MikaelSmith:
Should be ready, I think I've fixed the specs.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -84,24 +84,44 @@ def add_node_data(node)
...
-    elsif resource[:links] == "follow"
-      resource[:ensure] = metadata.ftype
     end
 
+    # Most of the time, ensure should match the returned metadata. Exceptions are when
+    # ensure is `present` (it could be anything), or a path (implies creating a link).
+    # `to_ral` caused the path case to munge to managing links, so we warn if there's
+    # a mismatch that's not `present`.
+    if resource[:ensure] != :present && resource[:ensure] != metadata.ftype
+      Puppet.warning "#{resource} expected to create a #{resource[:ensure]}, but its source "+
```
MikaelSmith: @jtappa This PR adds some new warnings, care to look at them?

MikaelSmith:
Updated. Also added confirmation of absence of warnings, and found a mistake comparing string to symbol.

jtappa:
:+1: for warning text

joshcooper:
Looks good, always setting `ensure` to the `metadata.ftype` (when not using `present`) and warning if it doesn't match makes sense. Also setting the checksum type based on metadata seems more correct. Just have the quoting comments and then ready to merge.

MikaelSmith:
Updated to quote everything.

joshcooper:
I'm wrong about the requirement to quote `owner`/`group`. We require `mode` to be quoted in manifests to disambiguate `mode => 755`, as the language will interpret the mode as decimal (due to absence of leading 0), and that can cause surprising behavior, e.g. PUP-2156. And we removed support for non-quoted modes in PUP-2350.

PR #4671, opened 2016-02-15
(PUP-5761) Log warning instead of raising error when bad keys are detected in module data

thallgren:
This PR changes the behavior when validating the data of a module data provider so that:
- It is no longer considered an error if a key is unqualified. Instead a warning is logged.
- The validation works the same way for all types of module data providers (a design flaw prevented the Hiera module data provider from validating correctly).

Details can be found in the commit comments.

### DIFF
--- a/lib/puppet/plugins/data_providers/data_provider.rb
+++ b/lib/puppet/plugins/data_providers/data_provider.rb
@@ -114,17 +114,26 @@ module Puppet::Plugins::DataProviders
     end
     protected :data_key
 
-    # Assert that all keys in the given _data_ are prefixed with the given _module_name_.
+    # Assert that all keys in the given _data_ are prefixed with the given _module_name_. Remove entries
+    # that does not follow the convention and log a warning.
     #
     # @param data [Hash] The data hash
     # @param module_name [String] The name of the module where the data was found
-    # @return [Hash] The data_hash unaltered
+    # @return [Hash] The possibly pruned hash
+    # @api public
     def validate_data(data, module_name)
       module_prefix = "#{module_name}::"
-      data.each_key do |k|
-        unless k.is_a?(String) && (k == LOOKUP_OPTIONS || k.start_with?(module_prefix))
-          raise Puppet::DataBinding::LookupError, "Module data for module '#{module_name}' must use keys qualified with the name of the module"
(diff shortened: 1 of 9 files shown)


lib/puppet/plugins/data_providers/data_provider.rb
```diff
@@ -112,22 +110,29 @@ def data_key(key, lookup_invocation)
       throw :no_such_key if qual_index.nil?
       key[0..qual_index-1]
     end
-    protected :data_key
 
-    # Assert that all keys in the given _data_ are prefixed with the given _module_name_.
+    # Assert that all keys in the given _data_ are prefixed with the given _module_name_. Remove entries
```
hlindberg: Should really be "Asserts" - but not worth fixing only this

spec/unit/functions/lookup_spec.rb
```diff
@@ -378,11 +395,24 @@ def compile_and_get_notifications(code)
   end
 
   context 'when accessing bad data' do
-    it 'will raise an exception when key in the function provided module data is not prefixed' do
-      Puppet[:code] = 'include bad_data'
-      expect do
-        compiler.compile()
-      end.to raise_error(Puppet::ParseError, /data for module 'bad_data' must use keys qualified with the name of the module/)
+    it 'will log a warning when key in the function provided module data is not prefixed' do
```
hlindberg: "a warning will be logged...."

spec/unit/functions/lookup_spec.rb
```diff
@@ -378,11 +395,24 @@ def compile_and_get_notifications(code)
...
+      Puppet[:code] = "include bad_data\nlookup('bad_data::c')"
+      logs = []
+      Puppet::Util::Log.with_destination(Puppet::Test::LogCollector.new(logs)) do
+        compiler.compile
+      end
+      warnings = logs.select {|log| log.level == :warning }.map {|log| log.message }
+      expect(warnings).to include("Module data for module 'bad_data' must use keys qualified with the name of the module")
+    end
+
+    it 'will log a warning when key in the hiera provided module data is not prefixed' do
```
hlindberg: - " -

hlindberg:
made some very picky comments - I am fine merging anyway, let me know if you want to do those small fixed @thallgren  ?

PR #4672, opened 2016-02-15
(PUP-5899) Fix bug when handling empty yaml for hiera data provider

thallgren:
Before this commit, an empty yaml file would cause an internal error.

An empty yaml file is valid. When it's parsed it produces a `nil`. This
commit ensures that the yaml-based data provider for Hiera treats that
the same way as an empty hash.

For completeness, this commit also adds a test of what happens when
an empty JSON file is encountered. That should (and does) raise a
lookup exception indicating a parse error.


spec/unit/functions/lookup_spec.rb
```diff
@@ -416,6 +416,19 @@ def compile_and_get_notifications(code)
...
 
+  context 'when accessing empty files' do
+    # An empty YAML file is OK and should be treated as a file that contains no keys
+    it "will fail normally with a 'did not find a value' error when a yaml file is empty" do
+      Puppet[:code] = "include empty_yaml\nlookup('empty_yaml::a')"
+      expect { compiler.compile }.to raise_error(Puppet::ParseError, /did not find a value for the name 'empty_yaml::a'/)
+    end
+
+    # An empty JSON file is not OK. Should yield a parse error
+    it "will fail with a LookupError indicating a parser failure when a json file is empty" do
```
hlindberg: Why is that an error when an empty yaml is not ? Just curious....

thallgren: Because empty does not represent anything in JSON, not even null. Good answer with links here if you want more details: http://stackoverflow.com/questions/18419428/what-is-the-minimum-valid-json

hlindberg:
Thomas is right, and the legit case of needing a JSON that contains nothing can be achieved, we do not want to treat invalid JSON as valid.

PR #4673, opened 2016-02-15
(PUP-5898) Make :undef be inferred as Undef type (aka. bring back the undead)

hlindberg:
Before this, if :undef was found in data coming from 3.x API into 4.x it
was inferred as Runtime[ruby, 'Symbol'] causing type mismatches to
occur.

This is now changed so that it is inferrd to be an Undef.

Unfortunately we had to revert the change that made :undef be treated as
an alien Ruby Symbol. It was intended to catch :undef leaking, but all
we achived was to catch ourselves in the act - and it is difficult to do
anything about this as some resurces do need 3.x call semantics and
others don't. It is not known a priori which ones.


hlindberg:
I did not look into fixing it the way we would want it. Just changing transform to not transforming to 3.x for resource parameters causes 50-100 failures in tests. There seems to be a lot of different things to fix. Hence this much smaller but much smellier fix.

thallgren:
This PR should be rebased on master.

hlindberg:
Closing this in favor of rebased to stable branch

PR #4674, opened 2016-02-16
(PUP-4929) Fix acceptance test to specify a gem source

MikaelSmith:
If the environment variable GEM_SOURCE is specified on the master,
add it to the gem package resources as a source to remove external
dependencies. Without this we get transient failures when rubygems.org
has issues.


cprice404:
:see_no_evil:

PR #4676, opened 2016-02-16
(PUP-5118) Manage duplicate file resources

HAIL9000:
Prior to this commit, the compiler could not handle duplicate file
resources when recursively in-lining metadata. This meant that a case
where a file is explicitly managed and also covered by its parent
directory which has recurse turned on would raise an error.

Now this case is handled by using remove_less_specific_files to
prune out duplicate resources. We will always keep the explicitly
managed version of the resource over the one generated by recursion.


joshcooper:
Looks good to me. Are you still working on the unit tests for the second commit?

Also side note, it looks like the existing test for `remove_less_specific_files` is duplicated, https://github.com/puppetlabs/puppet/blob/master/spec/unit/type/file_spec.rb#L521-L546. Could you remove the extra one?

PR #4677, opened 2016-02-17
(PUP-5815) Add strict_environment_mode option

whopper:
This commit adds a new Puppet option: `strict_environment_mode`.
This option ensures that the agent-specified environment is considered
authoritative, and fails the run if the environment of the retrieved
catalog does not match.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -226,7 +226,7 @@ class Puppet::Configurer
       configured_environment = Puppet[:environment] if Puppet.settings.set_by_config?(:environment)
 
       # We only need to find out the environment to run in if we don't already have a catalog
-      unless options[:catalog]
+      unless (options[:catalog] || Puppet[:strict_environment_mode])
         begin
           if node = Puppet::Node.indirection.find(Puppet[:node_name_value],
               :environment => Puppet::Node::Environment.remote(@environment),
@@ -274,6 +274,11 @@ class Puppet::Configurer
       query_options[:configured_environment] = configured_environment
 
       unless catalog = prepare_and_retrieve_catalog(options, query_options)
+        return nil
+      end
+
+      if Puppet[:strict_environment_mode] && catalog.environment != @environment
+        Puppet.err "Not using catalog because its environment '#{catalog.environment}' does not match agent specified environment '#{@environment}' and strict_environment_mode is set"
         return nil
       end
 
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -207,8 +207,13 @@ class Puppet::Configurer
       if catalog = prepare_and_retrieve_catalog_from_cache
         options[:catalog] = catalog
         @cached_catalog_status = 'explicitly_requested'
-        @environment = catalog.environment
-        report.environment = catalog.environment
+
+        if @environment != catalog.environment
+          Puppet.notice "Local environment: '#{@environment}' doesn't match the environment of the cached catalog '#{catalog.environment}', switching agent to '#{catalog.environment}'."
+          @environment = catalog.environment
+        end
+
+        report.environment = @environment
       else
         # Don't try to retrieve a catalog from the cache again after we've already
         # failed to do so the first time.
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -153,6 +153,12 @@ module Puppet
       :type       => :boolean,
       :desc       => "Whether to compile a static catalog."
     },
+    :strict_environment_mode => {
+      :default    => false,
+      :type       => :boolean,
+      :desc       => "Whether the agent specified environment should be considered authoritative,
+        causing the run to fail if the retrieved catalog does not match it.",
+    },
     :autoflush => {
       :default => true,
       :type       => :boolean,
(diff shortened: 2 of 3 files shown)


whopper:
/cc @joshcooper 

A few questions:

First, are we alright with this output?
```
[root@f6sj4055u18pwdy puppet]# bundle exec puppet agent -t --server=o1ihzvsvenmrgeg.delivery.puppetlabs.net --strict_environment_mode --environment=production --use_cached_catalog
Info: Using cached catalog from environment 'test'
Error: Not using catalog because its environment 'test' does not match agent specified environment 'production' and strict_environment_mode is set
```

I.e, when we grab the cached catalog, we always say "Info: Using cached catalog from environment 'blah'`, but in this case that environment doesn't match, so we then fail. Does this seem reasonable, or should that first `Info` not happen at all?

lib/puppet/configurer.rb
```diff
@@ -207,7 +207,7 @@ def run_internal(options)
       if catalog = prepare_and_retrieve_catalog_from_cache
         options[:catalog] = catalog
         @cached_catalog_status = 'explicitly_requested'
-        @environment = catalog.environment
+        @environment = catalog.environment unless Puppet[:strict_environment_mode]
```
whopper: Two things here:

1) What should the `environment` field of the report say when `strict_environment_mode` is set? With the way the code is not, If we're using a cached catalog, the report will include the environment of the catalog (along with the fact that the run failed and the error message). Does that seem sane, or should `environment` be empty or the agent specified environment?

2) Unrelated to `strict_environment_mode`, I realized that we probably should log some info saying that we're changing environments when using cached catalogs. We do it in the convergence loop with "Local environment doesn't match blah, restarting agent run with blah.." I'm going to add a second commit to add a notice here.

joshcooper: I'm wondering if `strict_environment_mode` should apply when using a cached catalog? I'm thinking the least surprising thing would be to always use the environment in the cached catalog.

Otherwise, I think we need to take into account whether the user specified an environment on the command line/settings, and really did intend to only use the cached catalog from a particular environment. That would be a lot easier if the agent didn't default to `production` so we could differentiate between `nil` (has no agent environment) vs `agent-specified environment`.

Thoughts @MikaelSmith?

Adding a notice message when switching environments sounds good.

MikaelSmith: Having `strict_environment_mode` use the environment from the cached catalog seems reasonable. The run would always fail if `strict_environment_mode`, `environment`, and caching are all set and the environment and cached catalog don't match.

We should make that part of the documentation for it.

whopper: Sounds reasonable - updating.

whopper:
@MikaelSmith @joshcooper updated and I'll do some more testing tomorrow. One last behavior question: what should happen under the following conditions?

1) The agent has a cached catalog with environment 'production' from a previous run
2) The NC on the server is changed and the agent will now be in environment 'test'
2.5) We run `puppet agent --onetime --strict_environment_mode --environment=test`
3) There happens to be a syntax error in one of the manifests in the 'test' environment, thus causing the agent to try to fall back to its cached catalog (while still in `strict_environment_mode`)
4) The agent sees that the cached catalog is of the environment `production`, rather than `test`.

The way this code stands, the agent will fail this run with `Error: Not using catalog because its environment 'production' does not match agent specified environment 'test' and strict_environment_mode is set`.

Right now, I think this makes sense. If you're specifying `strict_environment_mode` and expect to be retrieving a new catalog, it seems surprising to *fallback* to a cached catalog of a different environment.

I think if you are (for some reason) explicitly specifying `--use_cached-catalog` along with the new mode, then you probably expect to use your cached catalog, and it's fine to switch environments. Falling back, however, seems surprising. Thoughts?

MikaelSmith:
Pinging @nicklewis for this one. I can see any of 3 options working
- `strict_environment_mode` disables cached catalogs all together
- we fall back to cached catalog and accept it's environment
- we fail in the manner you outline

I think all are reasonable and comprehensible.

nicklewis:
If we consider `strict_environment_mode` to be a general mode of operation paired with the `environment` setting, then it seems like option three (fail) makes the most intuitive sense.

Note however that the orchestrator always specifies `--no-usecacheonfailure`, so the behavior is fairly inconsequential for it either way.

joshcooper:
:+1: to `strict_environment_mode` being a general mode of operation, independent of AO.

kylog:
This was all making sense to me up until:

> I think if you are (for some reason) explicitly specifying --use_cached-catalog along with the new mode, then you probably expect to use your cached catalog, and it's fine to switch environments. Falling back, however, seems surprising. Thoughts?

Wrt that first sentence, specifying `--strict_environment_mode` and then switching environments is surprising to me.

whopper:
@kylog that's the case where you're specifying `--use_cached_catalog` and `--strict_environment_mode` at the same time. Which option should win?

Earlier we decided that if you explicitly set `--use_cached_catalog`, then we'd consider the cached catalog's environment authoritative and immediately switch to it. With that being the case, it doesn't make much sense to specify `--use_cached_catalog` with `--strict_environment_mode` and `--environment=some_env` (we even thought about making them mutually exclusive options).

All of that being said, we went the route of saying if you explicitly request a cached catalog, then `strict_environment_mode` and a specified env are effectively noop, and you're going to use that cached catalog no matter what, regardless of its environment.

Does that make sense? That's just the way I've been thinking about it, but that could be totally flawed :)

PR #4679, opened 2016-02-17
(maint) Properly stub configurer methods to prevent errors in tests

whopper:
When running Puppet::Configurer::run_internal in spec tests,
attempting to send the report and write the class file error out
due to arbitrary files not existing on the filesystem. Stub
the methods responsible for these calls out so the error we are
trying to actually test can more cleanly be tested.

### DIFF
--- a/spec/unit/configurer_spec.rb
+++ b/spec/unit/configurer_spec.rb
@@ -677,6 +677,11 @@ describe Puppet::Configurer do
 
     describe "and strict environment mode is set" do
       before do
+        @catalog.stubs(:to_ral).returns(@catalog)
+        @catalog.stubs(:write_class_file)
+        @catalog.stubs(:write_resource_file)
+        @agent.stubs(:send_report)
+        @agent.stubs(:save_last_run_summary)
         Puppet.settings[:strict_environment_mode] = true
       end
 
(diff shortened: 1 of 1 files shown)


MikaelSmith:
:+1:

spec/unit/configurer_spec.rb
```diff
@@ -690,7 +690,8 @@ def expects_neither_new_or_cached_catalog
         @agent.instance_variable_set(:@environment, 'second_env')
         expects_new_catalog_only(@catalog)
 
-        Puppet.expects(:err).with(regexp_matches(/Could not/)).times(3)
+        @agent.stubs(:send_report)
+        @agent.stubs(:convert_catalog).returns(@catalog)
```
joshcooper: Since this is a private method in configurer, can we not stub it, and instead stub the `to_ral` part?

whopper: @joshcooper ah, makes sense. `to_ral` didn't do it, though. It seems like the only way to avoid the class and resource file errors is to stub out `@catalog.write_class_file` and `@catalog.write_resource_file`

joshcooper: sounds good.

PR #4681, opened 2016-02-17
(PUP-5735) Remove requirement of Win32-Security gem from Puppet

glennsarti:
The functionality of the Win32-Security gem has been merged into Puppet.
This commit moves the last of enumerations and updates the test suites to
use the newer namespace of Puppet::Util::Windows:SID.  This commit removes
Puppet's requirement of the win32-security gem, however the gem itself
still exists in the ruby gem library.  It will be removed from the library
in later commits.

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -35,6 +35,7 @@ gem_platform_dependencies:
       win32-dir: '~> 0.4.9'
       win32-eventlog: '~> 0.6.2'
       win32-process: '~> 0.7.4'
+      # Use of win32-security is deprecated
       win32-security: '~> 0.2.5'
       win32-service: '~> 0.8.6'
       minitar: '~> 0.5.4'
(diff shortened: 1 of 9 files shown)


ext/project_data.yaml
```diff
@@ -35,7 +35,6 @@ gem_platform_dependencies:
       win32-dir: '~> 0.4.9'
       win32-eventlog: '~> 0.6.2'
       win32-process: '~> 0.7.4'
-      win32-security: '~> 0.2.5'
```
Iristyle: Unfortunately we can't remove this here just yet... `project_data.yaml` is used for a couple of things:

* Development workflows (including running specs in CI), when performing a `bundle install` pick up these gem dependencies.  For the most part, this is unlikely to have many negative consequences (but it's possible there's some scenario that could feel some fallout)
* During the release process when we cut gems.  This is the bit we have to be more careful about since excluding the gem here is roughly equivalent to cutting it out of the `puppet-win32-ruby` repo for the sake of the MSI packaging.  For this reason, we should wait for a deprecation cycle / next semver bump before we completely yank this (and it should be synced up with the removal from the vendored gems).

Iristyle:
Aside from the `project_data.yaml` change, this looks good.

ext/project_data.yaml
glennsarti: Fixed.  Added a comment that use of the gem is deprecated

PR #4683, opened 2016-02-17
(PUP-5815) Handle cached catalogs with strict environment mode setting

whopper:
This commit ensures that if a cached catalog is explicitly requested
and strict environment mode is also requested, we do *not* switch the
local environment to match the cached catalog, as we do in all other
cases. This will cause the run to fail if the cached catalog's
environment does not match the agent specified environment.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -208,7 +208,7 @@ class Puppet::Configurer
         options[:catalog] = catalog
         @cached_catalog_status = 'explicitly_requested'
 
-        if @environment != catalog.environment
+        if @environment != catalog.environment && !Puppet[:strict_environment_mode]
           Puppet.notice "Local environment: '#{@environment}' doesn't match the environment of the cached catalog '#{catalog.environment}', switching agent to '#{catalog.environment}'."
           @environment = catalog.environment
         end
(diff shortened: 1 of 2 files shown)


MikaelSmith:
The use case where this would be irritating is if someone wants to normally `use_cached_catalog`, and also default to `strict_environment_mode` when they explicitly run without cached catalogs. Then it might make sense to set both options, and expect `strict_environment_mode` to have no effect unless they disable `use_cached_catalog`.

However, I have no problem with making people be explicit, and failing with combinations that don't make sense.

joshcooper:
:+1: to being explicit.

AO does run with `--use_cached_catalog` by default, but is planning on running jobs using `--no-use_cached_catalog --no-usecacheonfailure --strict_environment_mode`, so I think we're okay there. /cc @lutter

lutter:
We tell users to configure their agents to make periodic runs (out of cron or from the daemon) with {{--use_cached_catalog}} In that case, {{--strict_environment_mode}} should not be used.

When the orchestrator kicks off an agent run, we run with {{--no-use_cached_catalog}} and will use {{strict_environment_mode}} in the future.

IOW, this commit will not affect AO, but the behavior in this PR makes perfect sense to me.

PR #4684, opened 2016-02-18
(WIP) (PUP-5114) Retrieve static file content via the code_id parameter

branan:
Things this still needs:

* acceptance run (building a puppet-agent with the changes now)
* full commit message
* add some unit tests on the `File` side of things

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -84,9 +84,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
   end
 
   # Rewrite a given file resource with the metadata from a fileserver based file
-  def replace_metadata(resource, metadata)
-    # TODO: when we stop using to_ral, we'll have to add a case for ensure => <path>
-    # implying a link.
+  def replace_metadata(resource, metadata, environment_path)
     if resource[:links] == "manage" && metadata.ftype == "link"
       resource.delete(:source)
       resource[:target] = metadata.destination
@@ -123,6 +121,18 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
     when "use_when_creating"
       raise Puppet::Error, "source_permissions => use_when_creating cannot be set when creating static catalogs"
     end
+
+    # The static file content server doesn't know how to expand
+    # mountpoints, so we need to do that ourselves from the actual
+    # system math of the source file. This does that, while preserving
+    # any user-specified server or port.
+    source_path = Pathname.new(metadata.full_path)
+    environment_path = Pathname.new(environment_path)
+    path = source_path.relative_path_from(environment_path).to_s
+    source_as_uri = URI.parse(CGI.escape(metadata.source))
+    server = source_as_uri.host
+    port = ":#{source_as_uri.port}" if source_as_uri.port
+    resource[:content_uri] = "puppet://#{server}#{port}/#{path}"
   end
 
   # Determine which checksum to use; if agent_checksum_type is not nil,
@@ -163,7 +173,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
 
           child_resource = Puppet::Resource.new(:file, File.join(file[:path], meta.relative_path))
           child_resource[:source] = meta.source
-          replace_metadata(child_resource, meta)
+          replace_metadata(child_resource, meta, environment_path)
 
           # Copy parameters from original parent directory
           file.original_parameters.each do |param, value|
@@ -190,7 +200,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
         raise "Could not get metadata for #{resource[:source]}" unless metadata
         if metadata.full_path.start_with? environment_path
           # If the file is in the environment directory, we can safely inline
-          replace_metadata(resource, metadata)
+          replace_metadata(resource, metadata, environment_path)
         end
       end
     end
--- a/lib/puppet/type/file/source.rb
+++ b/lib/puppet/type/file/source.rb
@@ -265,7 +265,13 @@ module Puppet
     end
 
     def get_from_puppet_source(source_uri, &block)
-      request = Puppet::Indirector::Request.new(:file_content, :find, source_uri, nil, :environment => resource.catalog.environment_instance)
+      options = { :environment => resource.catalog.environment_instance }
+      if content_uri = resource[:content_uri]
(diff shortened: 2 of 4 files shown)


spec/unit/indirector/catalog/compiler_spec.rb
```diff
@@ -506,6 +506,7 @@ def expects_no_source_metadata
             .with_parameter(:checksum, checksum_type)
             .with_parameter(:checksum_value, sha)
             .with_parameter(:source, 'puppet:///modules/mymodule/config_file.txt')
+            .with_parameter(:content_uri, 'puppet:///modules/mymodule/config_file.txt')
```
joshcooper: content_uri should be of the form `'`puppet:///modules/mymodule/files/config_file.txt', perhaps an issue with how the metadata is stubbed in the test?

branan: Nope, that's me mis-reading the ticket. I didn't realize content_uri was expected to be the expanded environment-relative path of the file (which makes sense now that I've realized it - static_file_content doesn't want to know where to find modules and munge paths)

I can adjust the inlining logic to do the right thing there - I guess looking at `metadata.full_path`, and grabbing the environment-relative portion of it as `content_uri` is the correct implementation?

branan:
Also I thought I pushed up a fix for the typo, but apparently I only committed locally

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -103,6 +103,18 @@ def replace_metadata(resource, metadata, recurse = false)
...
     end
+
+    # The static file content server doesn't know how to expand
+    # mountpoints, so we need to do that ourselves from the actual
+    # system math of the source file. This does that, while preserving
+    # any user-specified server or port.
+    source_path = Pathname.new(metadata.full_path)
+    environment_path = Pathname.new(environment_path)
+    path = source_path.relative_path_from(environment_path).to_s
+    source_as_uri = URI.parse(metadata.source)
```
joshcooper: I think we need to:

```ruby
      uri = URI.parse(URI.escape(metadata.source))
```

as I believe `metadata.source` is not URL encoded, and is why we do https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/source.rb#L73. For example if you have in your manifest:

```puppet
file { '/tmp/foo bar':
  ensure => file,
  source => 'puppet:///modules/mymodule/foo bar'
}
```

Then `metadata.source` will contain a space, and trying to parse it will fail:

```
(byebug) metadata.source
"puppet:///modules/ntp/foo bar"
(byebug) URI.parse(metadata.source)
*** URI::InvalidURIError Exception: bad URI(is not URI?): puppet:///modules/ntp/foo bar
```

Secondary issue, `URI.escape` is deprecated, and we should probably be using `CGI.escape` instead. http://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape That's probably worthy of a separate non-direct puppet related ticket, as we do URI.escape in several places.

lib/puppet/type/file/source.rb
```diff
@@ -276,7 +276,11 @@ def chunk_file_from_disk
     end
 
     def get_from_puppet_source(source_uri, &block)
-      request = Puppet::Indirector::Request.new(:file_content, :find, source_uri, nil, :environment => resource.catalog.environment_instance)
+      if content_uri = resource[:content_uri]
+        request = Puppet::Indirector::Request.new(:static_file_content, :find, content_uri, nil, :code_id => resource.catalog.code_id)
```
MikaelSmith: No environment needed? I thought the static_file_content end-point expected environment.

branan: Just to close the loop on what happened in Hipchat: Ticket lied to me about it being needed. It is apparently required, and I've got an updated version for this

lib/puppet/indirector/catalog/compiler.rb
branan: fixed in upcoming push

whopper:
This looks to be working from my poking around with this branch. To get it to work, you need to:

1) Disable legacy_auth in puppetserver.conf
2) Add the following to auth.conf:
```
{
    match-request: {
        path: "/puppet/v3/static_file_content"
        type: path
        method: get
    }
    allow: "*"
    sort-order: 500
    name: "puppetlabs static file content"
},
```

You should then be able to observe the following behavior:

1) run an agent and get file content version 1 for some arbitrary file.
2) change that file on the server
3) remove that file on the agent
4) run the agent with --use_cached_catalog. The request will send the original code_id back to the server in the content request, and the file should be restored on the agent with the *original* content (not the newly changed server-side content)

:+1:

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -84,9 +84,7 @@ def add_node_data(node)
   end
 
   # Rewrite a given file resource with the metadata from a fileserver based file
-  def replace_metadata(resource, metadata)
-    # TODO: when we stop using to_ral, we'll have to add a case for ensure => <path>
```
MikaelSmith: Why drop this comment? It still needs to be done.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -123,6 +121,18 @@ def replace_metadata(resource, metadata)
...
     end
+
+    # The static file content server doesn't know how to expand
+    # mountpoints, so we need to do that ourselves from the actual
+    # system math of the source file. This does that, while preserving
+    # any user-specified server or port.
+    source_path = Pathname.new(metadata.full_path)
+    environment_path = Pathname.new(environment_path)
+    path = source_path.relative_path_from(environment_path).to_s
+    source_as_uri = URI.parse(CGI.escape(metadata.source))
```
MikaelSmith: Should this be in a try? Need to check why that was the case in source.rb.

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: Should never happen when source is specified.

lib/puppet/indirector/catalog/compiler.rb
branan: We should probably wrap it in a try and then print a warning about being unable to inline the specific resource, rather than let the exception fail the whole compile

branan:
@MikaelSmith Is this wholly subsumed by #4689 ?

MikaelSmith:
We think so.

MikaelSmith:
Superceded by #4689

PR #4685, opened 2016-02-18
(PUP-5923) Fix Admin Token detection on legacy Windows OS

glennsarti:
Without this commit, the detection of administrative rights on legacy
Windows operating systems (2003/XP) is incorrect.  The helper function
read_win32_bool returns a boolean type however it is compared against
an integer which causes the detection to always succeed.  This commit
uses the helper function return value directly.

### DIFF
--- a/lib/puppet/util/windows/user.rb
+++ b/lib/puppet/util/windows/user.rb
@@ -43,7 +43,7 @@ module Puppet::Util::Windows::User
         end
 
         # Is administrators SID enabled in calling thread's access token?
-        is_admin = ismember_pointer.read_win32_bool != FFI::WIN32_FALSE
+        is_admin = ismember_pointer.read_win32_bool
       end
     end
 


ferventcoder:
No spec updates? ;)

lib/puppet/util/windows/user.rb
```diff
@@ -43,7 +43,7 @@ def check_token_membership
         end
 
         # Is administrators SID enabled in calling thread's access token?
-        is_admin = ismember_pointer.read_win32_bool != FFI::WIN32_FALSE
+        is_admin = ismember_pointer.read_win32_bool
```
Iristyle: Good catch - not sure what I was thinking in the original code.

glennsarti:
To truly test this requires an acceptance test running under a useraccount that does not have the administrator SID in it's token.  Our test suite is not able to do this as it either runs as admin or local system.  It may be possible to mock the API call and do some testing there but it seemed too fragile to do so. I can attempt to add the unit test if you think it's worth it.

ferventcoder:
I think it is fine. This area of code may be removed at some point anyway.

PR #4686, opened 2016-02-18
(PUP-5922) Add feature flag to disable inlining recursive directories

MikaelSmith:
Handling recursive directories is fraught with issues, and unsure we can
resolve them all by release. Add a feature flag to disable inlining
metadata for recursive directories.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -153,6 +153,13 @@ module Puppet
       :type       => :boolean,
       :desc       => "Whether to compile a static catalog."
     },
+    :static_catalogs_recursion => {
+      :default    => false,
+      :type       => :boolean,
+      :internal   => true,
+      :desc       => "Whether to inline metadata for recursive directory resources
+        for static catalogs."
+    },
     :strict_environment_mode => {
       :default    => false,
       :type       => :boolean,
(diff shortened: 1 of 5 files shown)


MikaelSmith:
Is https://docs.puppetlabs.com/puppet/latest/reference/configuration.html auto-generated from defaults.rb?

lib/puppet/defaults.rb
```diff
@@ -153,6 +153,13 @@ def self.default_diffargs
       :type       => :boolean,
       :desc       => "Whether to compile a static catalog."
     },
+    :static_catalogs_recursion => {
+      :default    => false,
+      :type       => :boolean,
+      :internal   => true,
```
MikaelSmith: Added `:internal`, but we don't currently do anything or document what it means. I'm not sure where that should happen.

joshcooper:
It may be autogenerated using `puppet doc --reference configuration` or maybe `puppet-strings` @nfagerlund? If it's the former, it should be possible to omit internal settings in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/reference/configuration.rb#L3-L5

nfagerlund:
@joshcooper @MikaelSmith Yup, that's where configuration.html comes from. 

Having a way to hide settings (or even disable setting them in puppet.conf) would be excellent. Defaults.rb has long been a dumping ground for global constants that shouldn't necessarily be user-configurable, and that would be a step towards draining the stinkier parts of the swamp. (There's still a little over 200 settings available.)

joshcooper:
We probably won't need this after all.

joshcooper:
We are going to handle this a different way. See PUP-5114

PR #4687, opened 2016-02-19
(PUP-5925) Treat undef as a valid value from a binding

thallgren:
Before this commit, a binding for a host class definition parameter
that was declared as `undef` was treated as "not found" and hence,
would yield an error if the parameter didn't receive a value some
other way.

This commit changes this so that an `undef` value found in a binding
is considered a valid parameter default. An `undef` value will however
never replace a value produced by a parameter default expression.

### DIFF
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -438,19 +438,22 @@ class Puppet::Resource
     end
 
     missing_arguments.collect do |param, default|
-      # Using 'send' since this private method moved from here to type. The caller method is deprecated
-      # and doesn't really motivate making the callee public
-      external_value = resource_type.send(:lookup_external_default_for, param, scope)
-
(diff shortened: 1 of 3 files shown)


hlindberg:
The spec/unit/resource_spec.rb has too many changes - it is very hard to assess the change's effect on existing tests. Can you break apart the refactoring of tests from the actual changes. Ideally I would like to see a commit with almost no changes to the existing test code except cases that show the new behavior. That way it is easy to review that we did not cause a regression in expected behavior.

thallgren:
It looks much worse than it actually is. I needed divide the context in two, one with a default given for a parameter and one without. Hard to divide that up in a separate commit since one of the contexts then is empty.

hlindberg:
Thanks, that was much clearer - did not see it was just a whitespace issue.

PR #4689, opened 2016-02-19
(PUP-5114) Implement retrieving files via content_uri

MikaelSmith:
Inline metadata into the catalog. The file type uses that metadata when
available in place of making remote metadata requests.

This adds two new fields to the catalog: `metadata` and `metadatas`.
Metadata is a hash of File resource titles to the metadata for that
resource (for non-recursive resources). Metadatas is a hash of File
resource titles to a hash mapping recursive file sources to lists of
metadata found via recursion.

Obviates adding `content_uri` to the file type; it is now part of
`FileServing::Metadata`.

This approach was chosen instead of previous efforts because we found
much of the file type's logic was being duplicated in the master,
breaking type separation of concerns. Some knowledge of how files can be
queried is still needed to retrieve metadata, but much less than
previous work.

Major credits to @branan, @whopper, @joshcooper, and @HAIL9000 for
contributions to this PR and understanding the file type.


lib/puppet/file_serving/metadata.rb
```diff
@@ -13,6 +13,7 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
   indirects :file_metadata, :terminus_class => :selector
 
   attr_reader :path, :owner, :group, :mode, :checksum_type, :checksum, :ftype, :destination, :source_permissions
+  attr_accessor :content_uri
```
MikaelSmith: We should validate content_uri= to be a puppet:/// URI.

lib/puppet/file_serving/metadata.rb
```diff
@@ -136,7 +139,8 @@ def to_data_hash
         },
         'type'         => ftype,
         'destination'  => destination,
-      }
+        'source'       => source,
```
MikaelSmith: Skip inlining source when nil.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -83,48 +83,6 @@ def add_node_data(node)
...
-      resource.delete(:source)
-      resource[:target] = metadata.destination
-    end
-
-    # Most of the time, ensure should match the returned metadata. Exceptions are when
-    # ensure is `present` (it could be anything), or a path (implies creating a link).
-    # `to_ral` caused the path case to munge to managing links, so we warn if there's
-    # a mismatch that's not `present`.
-    if resource[:ensure] != "present" && resource[:ensure] != metadata.ftype
-      Puppet.warning "#{resource} expected to create a #{resource[:ensure]}, but its source "+
```
MikaelSmith: We could still add these warnings to the File type.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -138,59 +96,105 @@ def common_checksum_type(agent_checksum_type)
     checksum_type
   end
 
+  def get_content_uri(metadata, source, environment_path)
+    # The static file content server doesn't know how to expand mountpoints, so
+    # we need to do that ourselves from the actual system math of the source file.
```
kylog: s/math/path/ :)

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -138,59 +96,105 @@ def common_checksum_type(agent_checksum_type)
...
+  end
+
   # Inline file metadata for static catalogs
   # Initially restricted to files sourced from codedir via puppet:/// uri.
   def inline_metadata(catalog, checksum_type)
-    environment_path = File.join(Puppet[:environmentpath], catalog.environment, "")
+    environment_path = Pathname.new File.join(Puppet[:environmentpath], catalog.environment, "")
     list_of_resources = catalog.resources.find_all { |res| res.type == "File" }
 
+    # TODO: get property/parameter defaults if entries are nil in the resource
```
MikaelSmith: Ticket this for minor cleanup.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -138,59 +96,105 @@ def common_checksum_type(agent_checksum_type)
...
+              end
+            end
 
-          # Copy parameters from original parent directory
-          file.original_parameters.each do |param, value|
-            # These should never be passed to our children
-            unless [:parent, :ensure, :recurse, :recurselimit, :target, :alias, :source].include? param
-              child_resource[param] = value.to_s
+            source_to_metadatas[source] = list_of_data
+            # Optimize for returning less data if sourceselect is first
```
MikaelSmith: We may not want to do this, depending on what metadata means when it's added to puppetdb. Since that won't be the next release, we can remove it later if needed.

lib/puppet/resource/catalog.rb
```diff
@@ -277,6 +283,8 @@ def initialize(name = nil, environment = Puppet::Node::Environment::NONE, code_i
     super()
     @name = name
     @catalog_uuid = SecureRandom.uuid
+    @metadata = {}
```
MikaelSmith: These names are terrible. Also need to update API doc.

MikaelSmith:
TODO:
- fix specs
- remove content_uri from the File type

lib/puppet/resource/catalog.rb
```diff
@@ -433,7 +464,9 @@ def to_data_hash
       'environment'  => environment.to_s,
       'resources' => @resources.collect { |v| @resource_table[v].to_data_hash },
       'edges'     => edges.   collect { |e| e.to_data_hash },
-      'classes'   => classes
+      'classes'   => classes,
+      'metadata'  => metadata_hash,
```
MikaelSmith: We should not serialize if these are empty, to avoid sending extra fields to agents that don't support static catalogs.

MikaelSmith:
I removed specs that no longer made sense. There may still be a little cleanup to do there, and a new class of specs verifying how we call the FileServing::Metadata terminus would be useful.

MikaelSmith:
Open issue with this PR: if the master chooses a checksum besides the agent's digest_algorithm, the metadata from the server will never match the retrieve result for the file type.

Example setup in puppet.conf
```
[agent]
supported_checksum_types = sha256,md5
```

MikaelSmith:
Need to check whether any other params used in the metadata lookup depend on Puppet settings. Also, setting metadata for directories is strange, but need to do something to ensure files use the correct checksum.

MikaelSmith:
@ffrank I'd appreciate your thoughts on letting metadata be authoritative over the resource for `insync?` checks and the HTTP implementation.

api/schemas/catalog.json
```diff
@@ -94,6 +94,30 @@
             "items": {
                 "type": "string"
             }
+        },
+        "metadata": {
+            "description": "A mapping of non-recursive File resources to their required metadata",
+            "type": "object",
+            "patternProperties": {
+                "^.+$": {
```
MikaelSmith: I couldn't get `"$ref"` to work, but I really want this to refer to file_metadata.json.

branan:
@MikaelSmith I wonder if it actually makes sense to send along the negotiated checksum type as a separate field in the catalog, rather than per-resource. Sending it per-resource means that anything outside the environment path will use the Agent's default checksum type, which is potentially unsupported on the master. Now that we have that negotiation in place, it would be cool to take advantage of it across all file resources, not just the ones that we inlined metadata for.

MikaelSmith:
@branan we had that discussion during standup, and resolved to make the FileMetadata authoritative over the checksum type. I've updated the PR with that change. Then we're not sending anything new per-resource, we're using the information already available in the metadata. Then we get the other behavior for free (although currently I don't think the master will fallback to a different checksum if the one sent to the FileServing::Metadata terminus isn't supported).

MikaelSmith:
Pushed up new specs. I believe this is now ready to merge.

api/schemas/catalog.json
MikaelSmith: Ok, just added the definition to this file.

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: https://tickets.puppetlabs.com/browse/PUP-5959

MikaelSmith:
I didn't update the catalog or file_metadata docs because all the new return values are optional. @joshcooper @jtappa any thoughts on whether we should explicitly document the static catalog versions ?

joshcooper:
How about if we file a separate ticket to document that? I think there might be a way to validate the `file_metadata` value objects. I'm less confident we can validate that the keys refer to specific resources in the catalog.

lib/puppet/resource/catalog.rb
```diff
@@ -36,6 +36,14 @@ class DuplicateResourceError < Puppet::Error
   # The UUID of the catalog
   attr_accessor :catalog_uuid
 
+  # Inlined file metadata for non-recursive find
+  # A hash of title => metadata
+  attr_accessor :metadata
+
+  # Inlined file metadata for recursive search
+  # A hahs of title => { source => [metadata, ...] }
```
joshcooper: `hahs` typo

MikaelSmith:
Filed https://tickets.puppetlabs.com/browse/PUP-5960 for further documentation.

api/schemas/catalog.json
```diff
@@ -94,8 +94,88 @@
...
+            "description": "A mapping of recursive File resources to a hash of sources and arrays of associated metadata for that resource",
+            "type": "object",
+            "patternProperties": {
+                "^.+$": {
+                    "type": "object",
+                    "patternProperties": {
+                        "^.+$": {
+                            "type": "array",
+                            "items": {
+                                "$ref": "#/definitions/file_metadata"
```
joshcooper: I think if we wrap the current `file_metadata.json`schema in a `definitions`, something like:

```json
{
  "definitions": {
    "file_metadata": {
    "title":       "File Metadata",
    "description": "Metadata about a file, directory, or symbolic link",
    "type":        "object",
    ...
    }
  }
}
```

Then I think you can reference the schema from a different file like:

```json
   { "$ref": "file_metadata.json#/definitions/file_metadata" }
```

MikaelSmith: I'll try again, but that wasn't working for me earlier.

joshcooper: no worries, sorry didn't mean that as a blocker for this PR. We can handle that as part of the docs ticket

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -138,59 +96,105 @@ def common_checksum_type(agent_checksum_type)
...
+        options = {
+          :environment        => catalog.environment_instance,
+          :links              => resource[:links] ? resource[:links].to_sym : :manage,
+          :checksum_type      => resource[:checksum] ? resource[:checksum].to_sym : checksum_type.to_sym,
+          :source_permissions => resource[:source_permissions] ? resource[:source_permissions].to_sym : :ignore
+        }
+
+        metadata = nil
+        sources.each do |source|
+          if data = Puppet::FileServing::Metadata.indirection.find(source, options)
```
ffrank: Is this the master looking up the metadata? This is new here, right? In the HTTP case, will the master do all the necessary requests?

If so, I guess it looks good to me!

MikaelSmith: It is. The master only does lookup for metadata on sources from an environment in codedir. Everything else is left to happen agent-side.

The change for HTTP is that now
```
file {'/tmp/http_file':
  source => 'http://www.example.com/a_file',
}
```
will use MD5 locally to compare against `content-md5`, or mtime to compare against `last-modified`, based on the response from the server (i.e. without having to explicitly select it in the resource).

joshcooper: Also, we've added a `checksum_value` parameter to the file type, so you can specify the desired checksum type and value, and puppet will only attempt to retrieve the file if the checksum doesn't match, eliminating the need to retrieve http metadata:

```puppet
file { '/tmp/foo':
  ensure   => file,
  checksum => md5,
  checksum_value => '4e9531911f7a8caef3f20198457ba89b',
  source   => 'https://raw.githubusercontent.com/puppetlabs/puppet/master/Gemfile',
}
```

MikaelSmith: Probably want to point to a fixed SHA in that case. It'll fail to download a new file if master/Gemfile changes.

PR #4691, opened 2016-02-20
(PUP-5937) Detect duplicate keys in hashes

thallgren:
Before this commit, duplicate keys in hashes would go undetected
This commit changes this so that a duplicate key results in a warning:

"The key 'the_key' is declared more than once"

### DIFF
--- a/lib/puppet/pops/validation/validator_factory_4_0.rb
+++ b/lib/puppet/pops/validation/validator_factory_4_0.rb
@@ -27,6 +27,7 @@ class ValidatorFactory_4_0 < Factory
 
     p[Issues::FUTURE_RESERVED_WORD]          = :deprecation
 
+    p[Issues::DUPLICATE_KEY]                 = :warning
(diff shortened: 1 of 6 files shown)


hlindberg:
Marked as blocked pending decision how to deal with improved validations / warnings / errors. This because we have several such tickets...

lib/puppet/pops/validation/validator_factory_4_0.rb
```diff
@@ -27,6 +27,7 @@ def severity_producer
 
     p[Issues::FUTURE_RESERVED_WORD]          = :deprecation
 
+    p[Issues::DUPLICATE_KEY]                 = :warning
```
hlindberg: This should be rebased on PUP-5889 and then the level should be set to ```Puppet[:string]``` (the value is a compatible symbol). This allows the user to control strictness and if it is an error, warning or ignore.

hlindberg:
Replaced by #4705

PR #4692, opened 2016-02-20
(PUP-5844) Add Object type to the Puppet Type system

thallgren:
This commit adds a new `Object` type to the Puppet Type system. The
type is very similar to `Struct` in that it defines a hash of named
types. There are some notable differences though:

1. The `Object` type adds inheritance.
2. An `Object` is only assignable to itself or to a parent type of itself.
3. All keys of the `Object` must be non empty strings.
4. An `Object` may redefine a parent member, but only if the type is equal or assignable to the type of that member. I.e. it may redefine it to a more special type.

### DIFF
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -2080,13 +2080,16 @@ class PObjectType < PAnyType
   #
   # @api public
   def initialize(parent, members)
-    raise Puppet::Error, "An Object can only inherit from a Type"  unless parent.nil? || parent.is_a?(PAnyType)
+    raise Puppet::Error, 'An Object can only inherit from a Type' unless parent.nil? || parent.is_a?(PAnyType)
     @parent = parent
 
     if members.nil?
       members = PStructType::DEFAULT if members.nil?
     else
-      raise Puppet::Error, "The members of an Object must be represented by a Struct" unless members.is_a?(PStructType)
+      unless members.is_a?(PStructType)
+        raise Puppet::Error, 'The members (e.g. attributes or functions) of an Object must be represented by a Struct'
+      end
+
       # Assert that each key pair is valid
       rp = resolved_parent
       parent_hash = rp.is_a?(PObjectType) ? rp.members(true).hashed_elements : {}
@@ -2095,9 +2098,9 @@ class PObjectType < PAnyType
         parent_member = parent_hash[name]
         unless parent_member.nil? || parent_member.value_type.assignable?(e.value_type)
           raise Puppet::Error,
-            "Member '#{name}' redefines member of inherited object to type that does not match"
+            "The member (e.g. attribute or function) '#{name}' redefines inherited member to a type that does not match"
         end
-        TypeAsserter.assert_assignable('Member key', MEMBER_NAME_TYPE, e.key_type)
+        TypeAsserter.assert_assignable('Member (e.g. attribute or function) key', MEMBER_NAME_TYPE, e.key_type)
       end
     end
     @members = members
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -114,6 +114,14 @@ class PAnyType < TypedModelObject
     false
   end
 
+  # Called from the `PTypeAliasType` when it detects self recursion. The default is to do nothing
+  # but some self recursive constructs are illegal such as when a `PObjectType` somehow inherits itself
+  # @param originator [PTypeAliasType] the starting point for the check
+  # @raise Puppet::Error if an illegal self recursion is detected
+  # @api private
+  def check_self_recursion(originator)
+  end
+
   # Generalizes value specific types. Types that are not value specific will return `self` otherwise
   # the generalized type is returned.
   #
@@ -1968,6 +1976,10 @@ class PTypeAliasType < PAnyType
     guarded_recursion(guard, false) { |g| resolved_type.callable_args?(callable, g) }
   end
 
+  def check_self_recursion(originator)
+    resolved_type.check_self_recursion(originator) unless originator.equal?(self)
+  end
+
   def kind_of_callable?(optional=true, guard = nil)
     guarded_recursion(guard, false) { |g| resolved_type.kind_of_callable?(optional, g) }
   end
@@ -2009,6 +2021,7 @@ class PTypeAliasType < PAnyType
         guard = RecursionGuard.new
         accept(NoopTypeAcceptor::INSTANCE, guard)
         @self_recursion = guard.recursive_this?(self)
+        @resolved_type.check_self_recursion(self) if @self_recursion
       rescue
         @resolved_type = nil
         raise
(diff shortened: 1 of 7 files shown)


hlindberg:
Looks fine - mostly terminology and language changes. Test contexts/titles needs work and some seem to be in the wrong place (testing TypeParser things in the ObjectType spec); while it is fine to use the TypeParser to produce instances of ObjectType to perform tests on them, the tests that asserts that the TypeParser is doing its job should be in the TypeParser tests.

thallgren:
Please note that the TypeParser is just used to trigger the loader. It doesn't really parse the Object type (it's parsed by the pp_parser) so I think the remarks on "this is a TypeParser test" is wrong. One could argue that everything is more like integration tests since more than one component is involved but that's the case with an abundance of tests in puppet. These tests all test code in the PObjectType class. That' why they are in the same unit test file.

thallgren:
Regarding "features" versus "members". I'm aware that this is a type and not an oo-class. But regardless of that, I consider the term member better suited than feature. I know that the feature is the term used in ECore but I never liked it and I'm not comfortable using it.

The way I see it, inheritance is a feature of an Object. Being iterable is another feature, and having members is yet another. Features are very generic and not seldom something that can be controlled (turned on or off, like features of an XML parser, a JSON serializer, or, for that matter, the Puppet runtime). A member is a more specific. The fact that it's also a term in oo terminology should hardly prevent it (or be a disadvantage) here.

hlindberg:
Java calls them type variables (well sort of), C++ calls them type traits (also not quite the same thing)and those terms are not part of the OO hierarchies in those languages. In Ruby they are though (i.e. attributes and methods; I have not seen a common term).  I have strong distaste for the term "member", a member of a class is an instance of that class to me (akin to being a member of a club). A member is something that is "part of a group" an hence not the best term. I think the OO community did not get that quite right. Puppet users are not computer language scientists, and they are typically not familiar with OO. In an abstract sense "being a member of a group" means "being in a set", and you can naturally view the definition of a type as "a set of rules" - so it is not wrong.

You are right that attributes and methods/operations are not the only features, the term includes all there is to know about something; and our Object Type describes the features of instances of that type (in various ways). The term "aspect" is also in the same family of terms, but where the categorization is based on a "point of view" / bias - thus you can talk about the inheritance aspect, the logging aspect, serialization aspect. To me "feature" is more similar to "trait" in that they are both synonyms to "characteristic". The term 'property' is also a synonym, but I rule that one out because it is already overloaded with meaning in puppet.

My preferred choice of 'feature' is because it is unused in puppet terminology and I like it better than the other terms I feel are also acceptable: 'attribute', 'feature', 'trait', 'characteristic'. That other smart people use the term with precisely the same meaning is a bonus.

thallgren:
Java names variables for generic types "type variables" because they are just that, variables. The common name for methods and fields in Java is member (where incidentally, Member is the super class of Constructor, Field, and Method). The term member is commonly used in the C++ community too.

I'm convinced that the number of people working with Java or C++ that would confuse a member of a class with an instance of a class are far less than the number of people that would get confused if the term features where used to describe class members.

We're talking about feature switches in Puppet. Admittedly a very different context, but the term it's not completely unused. Feature is also a term that is nice to have available for other "feature" like constructs like the ones I mentioned before to control behavior by enabling or setting features. Members on the other hand, is something that I don't remember that I ever encountered in Puppet.

thallgren:
Noting that the term 'member' is [commonly used in C](https://en.wikipedia.org/wiki/Struct_(C_programming_language)) as well.

thallgren:
Updated PR with rebased commits (needed to fix the commit messages anyway), better titles for tests, and more tests.

hlindberg:
I am sure it is understandable from context either way. Our audience is neither C, C++, nor Java programmers, and they typically have no or little knowledge about OO. What the concept is called only matters when there are user facing error messages since those reach users even if they were not the ones authoring the logic that caused the error. Those that have to really understand an Object Type must naturally understand the concepts.

I just think it is ugly and unnatural to refer to these things as members when describing things abstractly. I would much rather use the term "feature" to mean the characteristics of different kinds that is being modeled.

I suggested in the comment that the issue with the error message could be avoided if instead using a concrete term; i.e "attribute", "function", etc.(a thing/concept that users see in the language),  instead of the abstract "member" (or "feature").

thallgren:
I can see two ways to distinguish an attr from a function:

1. Everything that is of type Callable is considered a function and everything else is an attribute.
2. We explicitly declare what's what in the definition.

The latter has the disadvantage in that it rules out a normal Struct declaration as the vehicle for keeping the members with all that that entails in terms of writing string representation logic, assignable logic, etc. It has the advantage of being able to tell the difference between an attribute of type Callable and a function. A new `Function` type could fix this but it's not ideal either since one would think of a Function as something named which wouldn't be applicable here.

hlindberg:
In that case it may be better to refer to both "methods" and "attributes" as being "attributes". Although I much prefer to say "features", maybe even with an explanation in the error say "features (e.g. attributes, or object functions)". (I am less concerned in using the term "member" in the implementation; we know what it means even if it is ugly ;-)).

thallgren:
Given that a large part of the world (includes Java, C++, C, Javascript, Python, even Smalltalk :-) ) uses "members" as a common term for attributes, methods, constructors, and struct entries, why can't we do that too? I'm fine changing the message to "members (e.g. attributes, or object functions)"

hlindberg:
Please use the term I suggested with the understanding that it is just an english word "features" that for the non technical is a synonym to "characterstics", which "member" is not. I have already stated this reason.

thallgren:
I've read your argumentation Henrik. I just don't agree. The term "features" sounds all wrong to me. Can we ask someone to judge this?

hlindberg:
We are not using it as a technical term here - you can keep all the references to "member" in the logic - it is just plain english for people that are not terribly technical when it comes to programming languages..

hlindberg:
tests are failing also

thallgren:
I don't understand your argumentation regarding "technical". In my world, the named thingies contained by an object are not features, technical or non-technical. Features describe behavior, not content. The fact that most languages have adopted the term "members" for this should speak for itself. Even UML uses the term "members".

hlindberg:
http://www.thesaurus.com/browse/feature vs. http://www.thesaurus.com/browse/member - which English non technical word is best suited to be the word for the abstract characteristics of an Object Type ? (where "non technical" means "not requiring a formal definition in a programming language specification").

You are giving the word "feature" a technical meaning - it means the same thing as "attribute", or "charactersitic" in english (i.e. attribute, property, etc.).

I can buy the error message "Where the characteristics of the Object Type must be described by a Struct" if you prefer that over "feature". (it means the same thing).

thallgren:
Here's what I think.

- Only users who design new object types will be subjected to the error messages.
- Users that create new object types are not ignorant to the programming terminology.
- A thesaurus gives little or no knowledge of how a word is perceived in a specific domain.
- Users that create new object types do that, because they want something oo-like.
- It's very likely that such users got their inspiration from languages like Javascript, Python, or C++.
- Using a domain specific terminology with well recognized and adopted terms is a key to success.

Here's what I don't think.

- An inexperienced member that encounters a message saying "The members (e.g. attributes and functions) of an Object type must be described by a Struct" will be very confused.
- An inexperienced user that sees the message "The characteristics of an Object type must be described by a Struct" will be much wiser.

hlindberg:
I can change the error message for you if you like.

joshcooper:
> My preferred choice of 'feature' is because it is unused in puppet terminology

Btw, puppet does have a concept of features, as they are used to determine feature suitability: https://docs.puppetlabs.com/guides/provider_development.html#suitability. Since we don't support polymorphic types, puppet allows behavior of the type to be customized based on the selected provider at runtime on the agent (aka leaky abstraction layer). For example, the service type's `ensure` value can only be set to `manual` when using the windows service provider. The user type's `managehome` parameter can only be set when using a provider that supports `manages_homedir`, etc.

hlindberg:
@joshcooper thanks for pointing that out, I forgot about those kinds of features only thinking about the more internal things like "root feature". The concept of "feature" as described for fact suitability fits very well with how I am proposing we use the term - for facter it is "does this system (where facter is running) have these characteristics (features)", and the term feature is used as it is about abstract features of the os/platform of various kinds. It would have been unnatural to call those "members".

Now, if this rules out using the term "feature" elsewhere then we are down to "characteristics", or "trait" I think as other terms are overloaded (like "attribute" and "property").

joshcooper:
It may be that the existing usage of "feature" is always prefaced with "provider features" and perhaps there wouldn't be conflict with the features you are talking about? Thoughts @nfagerlund?

thallgren:
We can agree completely that the term "features" as used for suitability is correct. That's more or less exactly the meaning that I was looking for above. It describes behavioral traits. I have no problem understanding the meaning of the term feature (or for that matter, characteristics) in that context.

I still don't understand what's wrong with what I'm proposing. It's by far the most commonly used term in our domain for what we're trying to name here and it's not used in Puppet terminology.

hlindberg:
We as software engineers understand the abstract "member of a set of things" as in "member of a struct", "member of a definition". "member of a class" etc. (the definition of an attribute is a definition among a set of definitions in a container that is also a definition that is of 'class'-type). I am not at all sure that the majority of our user basse understand this - it does not matter how many programming languages that use the abstract notion of "member of a set" as most users no nothing about any of those languages. If they know programming, they are more likely to know Ruby than any of the other, and there you do not talk about "members of sets" for these things.

We are already accused of having to much comp sci in puppet, using as everyday words as possible in anything that users see is important. Albeit, "feature of an ObjectType" may be just as much techno babble as "Member of an Object Type" when this shows up in a PE console. If that is the case then we are down to taste; where my preference is has already been made clear.

I think we can led Nick Fagerlund decide. (And to make it easier for him here is a summary;  ping @nfagerlund  - this is a debate which term to use as the common name for attributes and functions/methods (and later additional things like invariant rules, and some kind of polymorphism for provider-specific attributes in a resource type). We need this term in error messages but also in documentation.

One proposal is "member" because this is what it is called in several programming languages (not Ruby though). The other is "feature" because it is a synonym to "characteristic". Term to be used like "The names of members of an ObjectType must be unique", "Names of features of an Object types must be unique" etc. "What are the features of this type" and so on.

thallgren:
Thanks Henrik, for that clarification. My only objection to what you wrote is that "the majority of our user base" will not be involved in writing Object types. Writing such types is fairly advanced and as users move into that domain, they will be less negative to "comp sci" lingo. That "power language" is there for a reason and it would be wrong not to use it in advanced domains in fear of scaring of less advanced users.

hlindberg:
Thomas, do not forget that errors are in the face of everyone - not necessarily in the face of only the author. You do not know why something breaks at some point. Also documentation is read by more people even if they are not in the target audience for constructing something - they are reading other peoples code. The more complicated words/concepts we use in general, the more complicated it is to understand. (This drives me nuts sometimes as it is hard to explain things in layman's terms).

thallgren:
As I read up on UML terminology (trying to understand their motivation for member), I came across the following sentence:

"Sometimes the relationship between two classes is very weak. They are not implemented with member variables at all. Rather they might be implemented as member function arguments."

While perhaps not directly applicable here, it made me realize one thing. The term "member" has a containment implication. The term "feature" is more prominent. Replacing the term "member function" with "feature function", weakens the parent association and makes the sentence harder to read. "feature function" sounds like a function that belongs to the feature, rather than a being a thing in the parent.

Another motivation (the same, probably, since it's about containment) seems to be that they wanted to unify the term used when describing "members of a package" with "members of a class".

hlindberg:
As you correctly point out, "member" is a term used in a programming language to define one kind of "containment" (it is part of a set of similar things). We are describing an interface with the object type - it describes the characteristics of an object -  what its traits / features are. It does not say how they are implemented. I think it is already established why "member" is used at all as a term - it is math, "member of a set" and it is the most abstract term you can use here. The term "member" is synonym to "belongs to", but does not define what it is, or what the thing it belongs to is - it only describes "belongingness". The word "feature" OTOH *is* synonym with "characteristic" which is exactly what these things are. Thus when you read "member function" it is as non descriptive as "belonging function" (which make sense in languages where there is only one kind of applicable "belonging"). Now, I don't think we are ever going to use a term like "feature function", it is not a function of a feature - one is the more abstract term for the other (and other things like it)  - i.e. as intelligent as saying "fruit apple" or "vehicle car".

thallgren:
It's just that here, we describe attributes and functions that belong to an Object. Not "characteristics", or "features" (I agree those are synonyms).

hlindberg:
This is blocked until master frees up as this is targeting puppet 4.5.0

lib/puppet/pops/types/types.rb
```diff
@@ -2058,5 +2071,137 @@ def guarded_recursion(guard, dflt)
...
+
+      # Assert that each key pair is valid
+      rp = resolved_parent
+      parent_hash = rp.is_a?(PObjectType) ? rp.members(true).hashed_elements : {}
+      members.each do |e|
+        name = e.name
+        parent_member = parent_hash[name]
+        unless parent_member.nil? || parent_member.value_type.assignable?(e.value_type)
+          raise Puppet::Error,
+            "The member (e.g. attribute or function) '#{name}' redefines inherited member to a type that does not match"
```
hlindberg: And same here.

lib/puppet/pops/types/types.rb
```diff
@@ -2058,5 +2071,137 @@ def guarded_recursion(guard, dflt)
...
+        end
+        TypeAsserter.assert_assignable('Member (e.g. attribute or function) key', MEMBER_NAME_TYPE, e.key_type)
```
hlindberg: and here

spec/unit/pops/types/object_type_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+        expect{ |b| tp.members.each {|m| m.value_type.to_s.tap(&b) }}.to yield_successive_args('Integer[0, default]', 'Boolean')
+        expect{ |b| tp.members(true).each {|m| m.name.tap(&b) }}.to yield_successive_args('a', 'b', 'd')
+        expect{ |b| tp.members(true).each {|m| m.value_type.to_s.tap(&b) }}.to(
+          yield_successive_args('Integer[0, default]', 'Callable', 'Boolean'))
+      end
+
+      it 'can not redefine inherited member to a unassignable type' do
+        loader.expects(:load).with(:type, 'myderivedobject').returns(
+          type_object_t('MyDerivedObject', 'MyObject', '{ a=> String, d => Boolean }'))
+        expect { parser.parse('MyDerivedObject', scope) }.to raise_error(Puppet::Error, /redefines inherited member/)
```
hlindberg: needs update

hlindberg:
The term "members" need to change also before merging.

thallgren:
I thought we agreed on having a second opinion on that?

hlindberg:
We are not done with the design of this.

PR #4694, opened 2016-02-22
(PUP-4748) Allow purging of resources that have dependencies

branan:
Due to how resource generation was implemented, the purge check for
resource dependencies was actually broken until PUP-1963, which
accidentally caused that check to work.

Since that check has been broken for a long time, and users generally
have an expectation of being able to attach dependencies to a purge,
the check should be removed.

### DIFF
--- a/lib/puppet/transaction/resource_harness.rb
+++ b/lib/puppet/transaction/resource_harness.rb
@@ -69,8 +69,6 @@ class Puppet::Transaction::ResourceHarness
   def perform_changes(resource, context)
     cache(resource, :checked, Time.now)
 
-    return [] if ! allow_changes?(resource)
-
     # Record the current state in state.yml.
     context.audited_params.each do |param|
       cache(resource, param, context.current_values[param])
@@ -95,29 +93,6 @@ class Puppet::Transaction::ResourceHarness
 
     capture_audit_events(resource, context)
   end
-
-  def allow_changes?(resource)
-    if resource.purging? and resource.deleting? and deps = relationship_graph.dependents(resource) \
-            and ! deps.empty? and deps.detect { |d| ! d.deleting? }
-      real_deps = deps.reject { |r| r.class.name == :whit }
-      if real_deps.empty?
-        true
-      else
-        deplabel = real_deps.collect { |r| r.ref }.join(",")
-        plurality = real_deps.length > 1 ? "":"s"
-        resource.warning "#{deplabel} still depend#{plurality} on me -- not purging"
-        false
-      end
-    else
-      true
-    end
-  end
-
-  # allow_changes? is only made public to enable the existing spec tests to run
-  # under rspec 3 (apparently rspec 2 didn't enforce access controls?). Please do not
-  # treat this as part of a public API.
-  # Possible future improvement: rewrite to not require access to private methods.
-  public :allow_changes?
 
   def sync_if_needed(param, context)
     historical_value = context.historical_values[param.name]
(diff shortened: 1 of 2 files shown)


MikaelSmith:
My only question is whether this leaves anything describing dependencies that needs to be cleaned up. I need to read through the tickets to get some context.

MikaelSmith:
I don't know how to specify a dependency on a generated resource - besides containment - and if I could I don't see a reason it makes sense to do so. So this seems reasonable to ignore.

joshcooper:
Commit https://github.com/puppetlabs/puppet/commit/ab604527648e91b84529c3fb83761b5e9b269b77#diff-129f1c7cf7c3e8df7a5586dde253d845R48 added the check to make sure we don't delete a resource with an "implicit" relationship. That was later changed to checking if the resource was going to be purged in https://github.com/puppetlabs/puppet/commit/4a6d705bc086540e7d569c88f0130ad599bed5b1#diff-129f1c7cf7c3e8df7a5586dde253d845L43, and it updated the `resources` type to flag the generated resource as `purging`, so that the check actually worked.

And at that point in time, generated resources were added to the catalog before the relationship graph was constructed: https://github.com/puppetlabs/puppet/blob/4a6d705bc086540e7d569c88f0130ad599bed5b1/lib/puppet/transaction.rb#L478-L481. So that raises two questions:

1. Why wasn't this same issue triggered then?
2. At some point resource generation was moved to happen after relationship_graph creation (and we moved it back in PUP-1963). Was the first change an oversight (and we restored previous behavior in 1963)? Or was the first change done intentionally and 1963 regressed (in a different way than the case we fixed)?

joshcooper:
Also https://tickets.puppetlabs.com/browse/PUP-1963?focusedCommentId=266923&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-266923

MikaelSmith:
@branan and I had a brief discussion about this: it seems abusive to use `generate` for purging resources, vs `eval_generate`. And `eval_generate` would not partake in adding dependency edges.

However, the implementation of tidy is unique, and looks like it might need a rewrite to work effectively post-PUP-1963.

joshcooper:
The issue with tidy is because previously we added generated resources to the catalog in the order the `generate` method returned. For tidy, it uses depth first order, e.g. `/a/b/c`, `/a/b`, `/a`. Each time we added a resource (prior to PUP-1963), we validated its metaparameter relationships. So by the time we validated `/a`, all of its children had already been added to the catalog.

Tidy fails now (with and without this PR), because we call `reverse_each`, which causes us to validate metaparameters in the reverse order, e.g. validation of `/a` fails because `/a/b/` hasn't been added yet.

That is a bug with how PUP-1963 was implemented.

branan:
That makes sense, and should be pretty easily fixable (by making the catalog add_before and add_after methods a llittle smarter, so we don't need the reverse first). I'll ticket that and get it sorted today.

Regarding `generate` vs `eval_generate`, though, I do think it's worth examining which one we're using across all resources. I've been working on a doc that goes over that stuff, but obviously hadn't made it to Tidy yet before you found this bug.

MikaelSmith:
Based on my understanding, this seems ok. :+1:

joshcooper:
The check (we're removing) seemed to conflate "does this resource have downstream dependencies" vs "does the resource's downstream dependencies need to be deleted first". The first use case is completely valid, e.g. purge resources and notify the services to restart. The second use case seems like user error. If puppet tries to delete a parent directory before the children, then the resources in the manifest aren't ordered correctly. Unfortunately, puppet's file `autorequire` logic doesn't switch the direction of the relationship if you ensure=>absent, leading to circular dependencies: https://projects.puppetlabs.com/issues/14518#note-2.

MikaelSmith:
Now https://tickets.puppetlabs.com/browse/PUP-2451.

PR #4695, opened 2016-02-22
(PUP-5867) Add acceptance test for cached catalog status in report

whopper:
Update: beaker problems figured out! This passes in both Centos and Windows now.


whopper:
Note: since reports are named using timestamps, I went the route of removing all reports before each test and assuming there will only be one in the report directory when we need to read.

whopper:
Targeting 4.4.1, on hold until 4.4.0 is out.

joshcooper:
Needs to be retargeted to stable

PR #4696, opened 2016-02-22
(PUP-5569) acceptance: use Set for mk_tmp_environment instead of SortedSet

er0ck:
This change uses Set instead of SortedSet which requires external gems
and could be actually slower in many cases than searching a Set which is
a valueless hash in ruby.

[skip ci]


er0ck:
small change reviewed by @hlindberg.  self-merging

PR #4700, opened 2016-02-23
(maint) acceptance-skip exported resource test on arista

er0ck:
[skip ci]


whopper:
:+1: will merge after travis is happy or CI actually gets skipped :)

kylog:
Unfortunately that's just what those status checks look like if you specify `skip ci`. It'd be nicer if travis/appveyor reported back green (or lime-green or something).

PR #4702, opened 2016-02-23
(PUP-5121) Add profiling for static catalog and file metadata

haus:
This PR includes commits which add some profiling to help compare static catalog compiles to non-static catalog compiles as well as maintain counts for how often file metadata is and is not being inlined, and when not, for what reasons.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -123,6 +123,12 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
     end
   end
 
+  # Helper method to log file resources that could not be inlined because they
+  # fall outside of an environment.
+  def log_file_outside_environment
+    Puppet::Util::Profiler.profile("Not inlining file outside environment", [:compiler, :static_compile_inlining, :skipped_file_metadata, :file_outside_environment]) { true }
+  end
+
   # Inline file metadata for static catalogs
   # Initially restricted to files sourced from codedir via puppet:/// uri.
   def inline_metadata(catalog, checksum_type)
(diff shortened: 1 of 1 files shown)


lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -213,26 +242,28 @@ def compile(node, options)
     config = nil
 
     benchmark(:notice, str) do
-      Puppet::Util::Profiler.profile(str, [:compiler, :compile, node.environment, node.name]) do
+      compile_type = checksum_type ? :static_compile : :compile
+      Puppet::Util::Profiler.profile(str, [:compiler, compile_type, node.environment, node.name]) do
```
HAIL9000: I think I'm missing something, but what is str here?

MikaelSmith:
Seems reasonable.

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: It's constructed a couple lines up.

HAIL9000: Ah yeah, my bad GitHub was hiding it

joshcooper: Since the inlining metrics aren't grouped under `:compiler, :static_compile` anymore, is there value in `:compiler, :static_compile` as distinct from `:compiler, :compile`? It seems like it could get weird if there are existing metrics that are grouped under `:compiler, :compile`.

haus: defined above, outside the benchmark.

haus: @joshcooper the value would be in distinguishing static compiles from non-static compiles

nathanielksmith:
+1, but in this part of the code I don't think my +1s are worth much

MikaelSmith:
@joshcooper any more feedback on this?

PR #4703, opened 2016-02-23
(maint) Remove expectation of failure in direct puppet cache test

whopper:
The first of these commits adds a new static catalog helper in lib/acceptance, abstracted away from the original direct puppet from cache test.

The second commit corrects an error with the `versioned-code` puppetserver.conf settings and removes the `expect_failure` from the test.

Tested in Centos and windows, both pass.

/cc @MikaelSmith @joshcooper

### DIFF
--- a/acceptance/lib/puppet/acceptance/static_catalog_utils.rb
+++ b/acceptance/lib/puppet/acceptance/static_catalog_utils.rb
@@ -1,0 +1,46 @@
+module Puppet
+  module Acceptance
+    module StaticCatalogUtils
+
+      # Adds code-id-command and code-content-command scripts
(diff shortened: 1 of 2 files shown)


MikaelSmith:
:+1:

whopper:
ack, messed up a rebase. One sec

MikaelSmith:
Yeah, where'd the augeas commit come from?

whopper:
@MikaelSmith augeas thing was a PR merged yesterday, accidentally did an interactive rebase of the last 3 commits, somehow that ended up in here (even though that commit is already in master). Should be good now.

PR #4705, opened 2016-02-24
(PUP-5937) Detect duplicate keys in hashes (Merge after PUP-5889).

hlindberg:
This PR is rebased on top of the PR for PUP-5889 which adds the --strict flag.

Merge PUP-5889 before merging this.

Replaces #4691


hlindberg:
I am +1 on the work in PUP-5937 from Thomas.

PR #4706, opened 2016-02-24
(PUP-5963) Restore functionality of recursive tidy resources

branan:
Recursive tidy resources depended on the resources returned by tidy's
`generate` to be handled by the additional_resource_generator in a
specific order. This restores that order. I actually think that this
is "correct" for all depth-first resources, and any depth-first
generation has been wrong since PUP-1963. This also adds a test for
Tidy, since we apparently didn't have any.

### DIFF
--- a/spec/integration/type/tidy_spec.rb
+++ b/spec/integration/type/tidy_spec.rb
@@ -11,6 +11,23 @@ describe Puppet::Type.type(:tidy) do
     Puppet::Util::Storage.stubs(:store)
   end
 
+  it "should be able to recursively remove directories" do
+    dir = tmpfile("tidy_testing")
+    FileUtils.mkdir_p(File.join(dir, "foo", "bar"))
+
+    tidy = Puppet::Type.type(:tidy).new :path => dir, :recurse => true, :rmdirs => true
+
+    catalog = Puppet::Resource::Catalog.new
+    catalog.add_resource(tidy)
+    # avoid crude failures because of nil resources that result
+    # from implicit containment and lacking containers
+    catalog.stubs(:container_of).returns tidy
+
+    catalog.apply
+
+    expect(Puppet::FileSystem.directory?(dir)).to be_falsey
+  end
+
   # Testing #355.
   it "should be able to remove dead links", :if => Puppet.features.manages_symlinks? do
     dir = tmpfile("tidy_link_testing")
--- a/spec/integration/type/tidy_spec.rb
+++ b/spec/integration/type/tidy_spec.rb
@@ -1,11 +1,13 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
 
+require 'puppet_spec/compiler'
 require 'puppet_spec/files'
 require 'puppet/file_bucket/dipper'
 
 describe Puppet::Type.type(:tidy) do
   include PuppetSpec::Files
+  include PuppetSpec::Compiler
 
   before do
     Puppet::Util::Storage.stubs(:store)
@@ -15,15 +17,12 @@ describe Puppet::Type.type(:tidy) do
     dir = tmpfile("tidy_testing")
     FileUtils.mkdir_p(File.join(dir, "foo", "bar"))
 
-    tidy = Puppet::Type.type(:tidy).new :path => dir, :recurse => true, :rmdirs => true
-
-    catalog = Puppet::Resource::Catalog.new
-    catalog.add_resource(tidy)
-    # avoid crude failures because of nil resources that result
-    # from implicit containment and lacking containers
-    catalog.stubs(:container_of).returns tidy
-
-    catalog.apply
+    apply_compiled_manifest(<<-MANIFEST)
+      tidy { '#{dir}':
+        recurse => true,
+        rmdirs  => true,
+      }
+    MANIFEST
 
     expect(Puppet::FileSystem.directory?(dir)).to be_falsey
   end
@@ -36,15 +35,11 @@ describe Puppet::Type.type(:tidy) do
     Dir.mkdir(dir)
     Puppet::FileSystem.symlink(target, link)
 
-    tidy = Puppet::Type.type(:tidy).new :path => dir, :recurse => true
-
-    catalog = Puppet::Resource::Catalog.new
-    catalog.add_resource(tidy)
-    # avoid crude failures because of nil resources that result
-    # from implicit containment and lacking containers
-    catalog.stubs(:container_of).returns tidy
-
-    catalog.apply
+    apply_compiled_manifest(<<-MANIFEST)
+      tidy { '#{dir}':
+        recurse => true,
+      }
+    MANIFEST
 
     expect(Puppet::FileSystem.symlink?(link)).to be_falsey
   end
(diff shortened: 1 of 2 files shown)


branan:
@MikaelSmith @joshcooper

spec/integration/type/tidy_spec.rb
```diff
@@ -11,6 +11,28 @@
     Puppet::Util::Storage.stubs(:store)
   end
 
+  it "should be able to recursively remove directories" do
+    dir = tmpfile("tidy_testing")
+    foo = File.join(dir, "foo")
```
MikaelSmith: This could be
```
dir = tmpfile("tidy_testing")
FileUtils.mkdir_p(File.join(dir, "foo", "bar"))
```

spec/integration/type/tidy_spec.rb
```diff
@@ -11,6 +11,28 @@
...
+    Dir.mkdir(foo)
+    Dir.mkdir(bar)
+
+    tidy = Puppet::Type.type(:tidy).new :path => dir, :recurse => true, :rmdirs => true
+
+    catalog = Puppet::Resource::Catalog.new
+    catalog.add_resource(tidy)
+    # avoid crude failures because of nil resources that result
+    # from implicit containment and lacking containers
+    catalog.stubs(:container_of).returns tidy
```
MikaelSmith: That's strange, but appears to be copied from the other spec test.

spec/integration/type/tidy_spec.rb
```diff
@@ -11,6 +11,28 @@
...
+
+    catalog.apply
+
+    expect(Puppet::FileSystem.directory?(dir)).to be_falsey
```
MikaelSmith: Without the change above, did this fail?

branan:
```
[branan@pacifica puppet]$ bundle exec rspec spec/integration/type/tidy_spec.rb 
F.

Failures:

  1) Puppet::Type::Tidy should be able to recursively remove directories
     Failure/Error: raise error

     Puppet::ResourceError:
       Could not find dependency File[/tmp/rspecrun20160223-30534-1dvovhi/tidy_testing20160223-30534-1454pg7/foo] for File[/tmp/rspecrun20160223-30534-1dvovhi/tidy_testing20160223-30534-1454pg7]
     # ./lib/puppet/util/errors.rb:106:in `fail'
     # ./lib/puppet/type.rb:1491:in `block in validate_relationship'
     # ./lib/puppet/type.rb:1488:in `each'
     # ./lib/puppet/type.rb:1488:in `validate_relationship'
     # ./lib/puppet/type.rb:2468:in `block in finish'
     # ./lib/puppet/type.rb:2466:in `collect'
     # ./lib/puppet/type.rb:2466:in `finish'
     # ./lib/puppet/type/file.rb:463:in `finish'
     # ./lib/puppet/transaction/additional_resource_generator.rb:135:in `add_resource'
     # ./lib/puppet/transaction/additional_resource_generator.rb:41:in `block in generate_additional_resources'
     # ./lib/puppet/transaction/additional_resource_generator.rb:40:in `each'
     # ./lib/puppet/transaction/additional_resource_generator.rb:40:in `generate_additional_resources'
     # ./lib/puppet/transaction.rb:84:in `block in evaluate'
     # ./lib/puppet/transaction.rb:84:in `each'
     # ./lib/puppet/transaction.rb:84:in `evaluate'
     # ./lib/puppet/resource/catalog.rb:214:in `block in apply'
     # ./lib/puppet/util/log.rb:157:in `with_destination'
     # ./lib/puppet/transaction/report.rb:118:in `as_logging_destination'
     # ./lib/puppet/resource/catalog.rb:213:in `apply'
     # ./spec/integration/type/tidy_spec.rb:31:in `block (2 levels) in <top (required)>'

Finished in 0.83699 seconds (files took 0.72687 seconds to load)
2 examples, 1 failure
```

spec/integration/type/tidy_spec.rb
branan: Yeah, I don't know what in here is "weird legacy stuff" and just cargo culted it. I could double-check that this line is necessary

spec/integration/type/tidy_spec.rb
branan: oh, it's in FileUtils >_<. I tried `Dir.mkdir_p` and it didn't exist so I gave up

spec/integration/type/tidy_spec.rb
branan: Just double-checked, it is necessary

spec/integration/type/tidy_spec.rb
branan: Fixed

spec/integration/type/tidy_spec.rb
branan: So it fails on a line above, and I guess I could change it to an "expect the apply to not raise", but this is more what the test ought to look like for testing the actual functionality (vs testing the bug that was introduced)

MikaelSmith: The test works for me as-is.

branan: It definitely fails for me if I check out the previous version of the code. Are you sure you got the updated test file alongside the old version of the `additional_resource_generator`?

This gist shows getting the old version of the code, failing test, grabbing code from this PR, and then test passing: https://gist.github.com/branan/77a266161f92f82ff43b

spec/integration/type/tidy_spec.rb
joshcooper: You might be able to use `apply_compiled_manifest` like we do in https://github.com/puppetlabs/puppet/blob/master/spec/unit/transaction_spec.rb#L640-L645 instead. That way you don't have to stub implementation details.

spec/integration/type/tidy_spec.rb
joshcooper: nit, `Puppet::FileSystem.directory?(dir)` could return false (and pass the test) if it's a file, but we're really expecting the directory to not exist. That said I can't imagine the tidy type would ever do that...

branan: I thought of that. I can't imagining that ever happening, and this mirrors the test below that expects `symlink?` to be false to remove a symlink.

MikaelSmith: Sorry, writing failure. I meant what you have here seems fine.

MikaelSmith:
@branan do you plan to rewrite with `apply_compiled_manifest`? Otherwise I'm :+1:

spec/integration/type/tidy_spec.rb
branan: Probably. If I do that I'd rather refactor the existing test in this file to follow the same pattern. I can do a maint commit that moves the `apply_compiled_manifest` on helper to a shareable location and rewrites these tests.

Do you want that as part of this PR, or should I open it separately once this is merged?

branan: Probably. If I do that I'd rather refactor the existing test in this file to follow the same pattern. I can do a maint commit that moves the `apply_compiled_manifest` on helper to a shareable location and rewrites these tests.

Do you want that as part of this PR, or should I open it separately once this is merged?

joshcooper: I think it would be good to be part of this PR, because we're having to stub an implementation detail of the code we're testing. I think it's as easy as requiring the helper and adding `include PuppetSpec::Compiler` to the describe block. But I'm also fine if you want to handle that later as part of a maint.

branan: Maint commit pushed up to this PR

MikaelSmith:
:+1:

PR #4708, opened 2016-02-24
(PUP-4742) Switch from vendored Semantic to use the SemanticPuppet gem.

pdrakeweb:
Without this patch applied, puppet includes vendored code in the Semantic
namespace.  This namespace collides with the Semantic gem.  This patch
fixes the problem by removing the vendored Semantic code and using the
SemanticPuppet gem.


kylog:
Ping @andersonmills.

This also implies corresponding work in https://github.com/puppetlabs/puppet-agent/tree/master/configs/components to add the semantic_puppet gem there, so that the puppet-agent package remains batteries-included wrt this functionality.

Also this may need to remove https://github.com/puppetlabs/puppet/blob/fdc85d3cc05ee2c79140027237143517162a3321/lib/puppet/vendor/load_semantic.rb and modify https://github.com/puppetlabs/puppet/blob/fdc85d3cc05ee2c79140027237143517162a3321/lib/puppet/vendor/require_vendored.rb#L4.

joshcooper:
Ping @andersonmills.

Also @melissa @underscorgan I think this would have puppet-agent implications.

melissa:
Can someone open a PA ticket to get semantic_puppet into puppet-agent? I'm not sure how soon RE will be able to get to it though.

hlindberg:
This blocks work for Language team as there are bugs in the semver logic that needs to be fixed.

kylog:
LGTM. @thallgren, I know you were doing work with dependencies on the semantic_puppet gem, can you review?

thallgren:
The SemanicPuppet gem changes the name of the global module `Semantic` to `SemanticPuppet` and this PR changes all occurrences, which is fine. The problem is that this PR was created before we added the `SemVer` and `SemVerRange` types which introduces some new occurrences of `Semantic`. This PR must be updated to cover them as well before it can be merged.

.gemspec
```diff
@@ -40,4 +40,5 @@ Gem::Specification.new do |s|
   s.specification_version = 3
   s.add_runtime_dependency(%q<facter>, [">= 1.7", "< 4"])
   s.add_runtime_dependency(%q<hiera>, [">= 2.0", "< 4"])
+  s.add_runtime_dependency(%q<semantic_puppet>, "0.1.1")
```
kylog: This should be 0.1.2 now. /cc @thallgren

thallgren:
Superseded [PR-4742](/puppetlabs/puppet/pull/4742).

PR #4710, opened 2016-02-24
(PUP-5889) add a strict flag

hlindberg:
This adds a --strict flag that can be set to warning, error or ignore. In this PR this new setting is made to play nicely with the --strict_variables such that, if --strict_variables is true then it always errors (this is what users expect). If it is not set, the behavior is controlled by the --strict setting.

The default for --strict is warn. This is also what is expected for --strict_variables (by default). Users can opt out by setting --strict=off.

It is expected that the --strict flag will be used in different ways for new kinds of validations. In this PR one tricky thing is to maintain the API for those that call directly into scope (not being prepared to catch the symbol :undefined_variable. This is now done by always throwing :undefined_variable, and catching the resulting exception if there were no catcher. Earlier the trowing was conditional which led to other problems.

Now the API is - if a caller wants to handle the error/warning aspect of an undefined variable it should catch the thrown symbol. If it does not the --strict_variables/--strict flags control the behavior (like before).

### DIFF
--- a/lib/puppet/pops/evaluator/runtime3_support.rb
+++ b/lib/puppet/pops/evaluator/runtime3_support.rb
@@ -502,14 +502,12 @@ module Runtime3Support
       #
       if Puppet[:strict_variables]
         p[Issues::UNKNOWN_VARIABLE] = :error
-      elsif Puppet[:strict] == :ignore
+      elsif Puppet[:strict] == :off
         p[Issues::UNKNOWN_VARIABLE] = :ignore
       else
         Puppet[:strict_variables] 
         p[Issues::UNKNOWN_VARIABLE] = Puppet[:strict]
       end
-
-#      Puppet[:strict_variables] ? :error : :warning
 
       # Store config issues, ignore or warning
       p[Issues::RT_NO_STORECONFIGS_EXPORT]    = Puppet[:storeconfigs] ? :ignore : :warning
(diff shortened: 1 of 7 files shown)


hlindberg:
Blocked while waiting for master to free up - this is for Puppet 4.5.0

hlindberg:
Rebased to latest master

spec/unit/defaults_spec.rb
```diff
@@ -72,6 +72,24 @@
 
   end
 
+  describe 'strict' do
+    it 'should accept the vaid value ignore' do
```
hlindberg: should say 'off'

lib/puppet/pops/evaluator/runtime3_support.rb
```diff
@@ -90,7 +90,7 @@ def get_variable_value(name, o, scope)
     unless name =~ Puppet::Pops::Patterns::NUMERIC_VAR_NAME
       optionally_fail(Puppet::Pops::Issues::UNKNOWN_VARIABLE, o, {:name => name})
     end
-    nil # in case unknown variable is configured as a warning
+    nil # in case unknown variable is configured as a warning or ignore
```
hlindberg: should say 'off'

hlindberg:
Found a couple of typos that I fixed and amended last commit.

PR #4713, opened 2016-02-24
(PUP-5872) Add acceptance test for catalog and report correlation

whopper:
This commit adds a new acceptance test which ensures that reports from cached catalog runs can be linked to a specific catalog using catalog_uuid.


whopper:
As it stands, this test only has two steps:

1) Upon retrieving a brand new catalog on the very first run, the newly cached catalog and newly filed report should have the same `catalog_uuid`.
2) Running a second time with `--use_cached_catalog` should result in a second report with the same `catalog_uuid` and the cached catalog.

Should we add any more test cases? Maybe additional runs with new catalogs mixed with cached catalogs?

I'm also wondering if it'd be meaningful to verify that `transaction_uuid` is different but `catalog_uuid` is the same. I think that's covered by the fact that I'm deleting the report after the first test, though.

joshcooper:
Targeting 4.4.1, on hold until 4.4.0 is out.

joshcooper:
Needs to be retargeted to stable

PR #4717, opened 2016-02-25
(PUP-5233) Restore `puppet inspect`

MikaelSmith:
Fix `puppet inspect` so it looks in the correct location for the cached
catalog.

### DIFF
--- a/spec/unit/application/inspect_spec.rb
+++ b/spec/unit/application/inspect_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 
 require 'puppet/application/inspect'
 require 'puppet/resource/catalog'
-require 'puppet/indirector/catalog/yaml'
+require 'puppet/indirector/catalog/json'
 require 'puppet/indirector/report/rest'
 require 'puppet/indirector/file_bucket_file/rest'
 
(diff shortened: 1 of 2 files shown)


whopper:
@MikaelSmith ah, you beat me to this! :+1: looks just like the changes I came up with :). As for the additional comments in the ticket that `inspect` doesn't display changes to local resources: I poked around in Puppet 3.x and observed the same behavior. As far as I can tell, `inspect` has only ever provided audit information, not whether a resource is `out_of_sync` or not.

So, the conclusion I came to was that this change is all that is needed to restore `inspect` functionality, and anything beyond that is an additional feature request. I could also be totally wrong about that, though :)

spec/unit/application/inspect_spec.rb
```diff
@@ -31,6 +31,12 @@
       Puppet[:report] = false
       expect { @inspect.setup }.to raise_error(/report=true/)
     end
+
+    it "should respect the catalog_cache_terminus if set" do
```
whopper: Do you think it's worth it to check that the default is `:json`? Then we'd know it was broken again if we changed our default cached catalog storage format.

MikaelSmith: Yeah, checking that the default is json makes sense.

whopper: Hah, sure

PR #4718, opened 2016-02-25
(PUP-5970) Add catalog format

joshcooper:
Adds a `catalog_format` field to the catalog identifying the schema
version. That way future readers of the catalog can know what data
to expect. Also updates the api doc and schema to reflect the current
catalog_format version 1.

### DIFF
--- a/lib/puppet/resource/catalog.rb
+++ b/lib/puppet/resource/catalog.rb
@@ -36,6 +36,11 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
   # The UUID of the catalog
   attr_accessor :catalog_uuid
 
+  # @return [Integer] catalog format version number. This value is constant
+  #  for a given version of Puppet; it is incremented when a new release of
+  #  Puppet changes the API for the various objects that make up the catalog.
+  attr_accessor :catalog_format
+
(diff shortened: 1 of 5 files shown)


hlindberg:
+1

joshcooper:
Btw, this really is meant to target stable and land in 4.4.0.

MikaelSmith:
This version is for the internal catalog format, rather than the wire format used by PuppetDB in https://docs.puppetlabs.com/puppetdb/latest/api/wire_format/catalog_format_v6.html. Someday we might normalize these two.

lib/puppet/resource/catalog.rb
```diff
@@ -398,6 +404,10 @@ def self.from_data_hash(data)
       result.catalog_uuid = catalog_uuid
     end
 
+    if catalog_format = data['catalog_format']
```
MikaelSmith: If catalog_format isn't present, should we default to 0? For a newer agent talking to an older master.

joshcooper: makes sense, let me fix it.

joshcooper: Updated

MikaelSmith:
:+1:

PR #4719, opened 2016-02-25
(PUP-5979) regsubst: hash as replacement parameter

vicinus:
Allow a hash as replacement parameter to the function regsubst again as
it was possible with puppet 3.x.

### DIFF
--- a/lib/puppet/functions/regsubst.rb
+++ b/lib/puppet/functions/regsubst.rb
@@ -16,9 +16,10 @@
 # @param regexp [String|Regexp|Type[Regexp]]
 #      The regular expression matching the target string.  If you want it anchored at the start
 #      and or end of the string, you must do that with ^ and $ yourself.
-# @param replacement [String]
+# @param replacement [String|Hash[String, String]]
 #      Replacement string. Can contain backreferences to what was matched using \\0 (whole match),
 #      \\1 (first set of parentheses), and so on.
+#      If the second argument is a Hash, and the matched text is one of its keys, the corresponding value is the replacement string.
 # @param flags [String]
 #      Optional. String of single letter flags for how the regexp is interpreted (E, I, and M cannot be used
 #      if pattern is a precompiled regexp):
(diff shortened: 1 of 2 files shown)


lib/puppet/functions/regsubst.rb
```diff
@@ -37,18 +38,18 @@
...
 Puppet::Functions.create_function(:regsubst) do
   dispatch :regsubst_string do
-    param          'Variant[Array[String],String]',  :target
-    param          'String',                         :pattern
-    param          'String',                         :replacement
-    optional_param 'Optional[Pattern[/^[GEIM]*$/]]', :flags
-    optional_param "Enum['N','E','S','U']",          :encoding
+    param          'Variant[Array[String],String]',       :target
+    param          'String',                              :pattern
+    param          'Variant[String,Hash[String,String]]', :replacement
```
hlindberg: Is an empty string ever ok as key?

lib/puppet/functions/regsubst.rb
```diff
@@ -37,18 +38,18 @@
...
   end
 
   dispatch :regsubst_regexp do
-    param          'Variant[Array[String],String]',  :target
-    param          'Variant[Regexp,Type[Regexp]]',   :pattern
-    param          'String',                         :replacement
-    optional_param 'Pattern[/^G?$/]',                :flags
+    param          'Variant[Array[String],String]',       :target
+    param          'Variant[Regexp,Type[Regexp]]',        :pattern
+    param          'Variant[String,Hash[String,String]]', :replacement
```
hlindberg: same here

hlindberg:
We prefer commit messages to be written in imperative form on their first line e.g.
"(PUP-5979) Make regsubst function accept a hash to define replacement".

hlindberg:
I originally asked if it was valid with empty string as key, but I think it is. I have used that to make a match that matches the empty strings between characters and replaced those with a delimiter - I therefore deleted those questions.

This is all ready to be merged but I would like the commit message to be changed if it is not too much of a hassle for you. Let me know in that case and I can fix that up.

vicinus:
I have changed the commit message as requested. Let me know if I can do anything further.

hlindberg:
Thank you for the contribution.

PR #4720, opened 2016-02-26
(PUP-5341) Emit TagSets as arrays in report YAML

branan:
Resource::Status and Util::Log objects were not including PsychSupport
to allow them to serialize their tags member as an array.

### DIFF
--- a/lib/puppet/resource/status.rb
+++ b/lib/puppet/resource/status.rb
@@ -1,5 +1,6 @@
 require 'time'
 require 'puppet/network/format_support'
+require 'puppet/util/psych_support'
 
 module Puppet
   class Resource
@@ -11,6 +12,7 @@ module Puppet
     #
     # @api private
     class Status
+      include Puppet::Util::PsychSupport
       include Puppet::Util::Tagging
       include Puppet::Network::FormatSupport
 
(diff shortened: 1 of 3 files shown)


branan:
I'm not sure what the best way to test this is - do we have an actual schema for reports that I could try to make stricter or something like that?

kylog:
I don't think we have a schema. But that would be great :)

What we have in facter, where there's a schema and then an acceptance test that generates output to validate against that schema, might be the basis for such a test. Otoh, it might be riggable as a spec/integration test, which would be nice for faster feedback loop.

hlindberg:
There are other tests in puppet that do stuff with schemas - I think we used it to validate that the schemas we wrote were valid.

hlindberg:
+1 to the fix, including PsychSupport was the right thing to do

branan:
This is updated with a test now (woohoo). Maybe could be unit instead of integration? It's definitely hitting a pretty wide swath of the codebase so doesn't feel "unit-y" to me, but also it doesn't interact with the underlying system. I'm not sure how the tests are suposed to be split up in Puppet core.

We need to wait for the 4.4.0 tag before we merge this, since it's targeted for 4.4.1

MikaelSmith:
Needs a rebase.

branan:
@MikaelSmith rebase done, tests are running

PR #4721, opened 2016-02-26
(PUP-5959) Derive static catalog metadata lookup defaults from File type

whopper:
Previously, while inlining file metadata for the static catalog
our default fallbacks for the :links and :source_permissions parameters
were hardcoded to match the file type.

This commit removes those hardcoded values in favor of querying a
dummy file resource for its actual defaults. An instantiated
resource is necessary due to Puppet::Type parameter defaults not being
resolvable until a specific provider is known. Since :links and
:source_permissions are provider-independent, it is safe to instantiate
the resource on the master.

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -140,8 +140,9 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
     environment_path = Pathname.new File.join(Puppet[:environmentpath], catalog.environment, "")
     list_of_resources = catalog.resources.find_all { |res| res.type == "File" }
 
-    # TODO: get property/parameter defaults if entries are nil in the resource
-    # For now they're hard-coded to match the File type.
+    defaults_resource = Puppet::Type.type(:file).new(:name => '/defaults')
+    links_default = defaults_resource[:links]
+    source_permissions_default = defaults_resource[:source_permissions]
 
     file_metadata = {}
     list_of_resources.each do |resource|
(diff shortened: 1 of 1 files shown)


whopper:
This feels really terribly gross but I'm not sure how else we'd be able to get defaults out of those parameters :(. Do we think it's worth it?

MikaelSmith:
Given that it's only two settings, I'm tempted to not do it.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -140,8 +140,9 @@ def inline_metadata(catalog, checksum_type)
     environment_path = Pathname.new File.join(Puppet[:environmentpath], catalog.environment, "")
     list_of_resources = catalog.resources.find_all { |res| res.type == "File" }
 
-    # TODO: get property/parameter defaults if entries are nil in the resource
-    # For now they're hard-coded to match the File type.
+    defaults_resource = Puppet::Type.type(:file).new(:name => '/defaults')
```
MikaelSmith: This path fails on Windows specs.

whopper: Ah damn, right. I think that means we've come full circle with needing the provider for the agent :).

MikaelSmith: Not really, just need to check whether the master is running on Windows.

whopper: Oh wait yeah this is master only (although the master can't run on Windows)

MikaelSmith: It might someday, and the spec tests exercise this code.

PR #4722, opened 2016-02-26
(PUP-5843) Add simple optimization of resource type loading

hlindberg:
Before this, the resolution of a type alias could not be made when a
resource type name was used in short form. Instead this would always be
a TypeReference to a a type via name to be resolved later.

Now, the StaticLoader defines all the resource types that are in puppet
such that a reference to e.g. File results in the type Resource['File'].
The removes the need to later resolve the reference.

This commit contains a simple static mapping where each type is
hard-coded. It is expected that we later replace this by loading a
complete type model of the puppet-runtime resource types. At that time
we define both a resource reference and the actual type with all of its
parameters.


hlindberg:
There is very little chance the list needs to be updated. Longer term we will scan available types and construct more comprehensive information directly - i.e. more than just the type reference. Worst case would be the removal of a resource type; which would be an unmissable big event.

Will adress the other issues.

PR #4723, opened 2016-02-26
(PUP-5964) Improve how type mismatches are asserted and reported

thallgren:
This PR contains several commits targeted to improve reporting of type mismatches and to make them more consistent. The details are in the commit messages.

This PR is for stable and supersedes [PR-4709](/puppetlabs/puppet/pull/4709)

### DIFF
--- a/lib/puppet/pops/types/type_mismatch_describer.rb
+++ b/lib/puppet/pops/types/type_mismatch_describer.rb
@@ -20,6 +20,12 @@ module Types
     alias :eql? :==
   end
 
+  class SubjectPathElement < TypePathElement
+    def to_s
+      key
+    end
+  end
+
   class MemberPathElement < TypePathElement
     def to_s
       "struct member #{key}"
@@ -392,6 +398,34 @@ module Types
       else
         raise Puppet::ParseError.new("#{subject}:\n #{errors.join("\n ")}")
       end
+    end
+
+    # Describe a confirmed mismatch using present tense
+    #
+    # @param name [String] name of mismatch
+    # @param expected [PAnyType] expected type
+    # @param actual [PAnyType] actual type
+    #
+    def describe_mismatch_present(name, expected, actual)
+      errors = describe(expected, actual, [SubjectPathElement.new(name)])
+      case errors.size
+      when 0
+        ''
+      when 1
+        errors[0].to_s.strip
+      else
+        errors.join("\n ")
+      end
+    end
+
+    # Describe a confirmed mismatch using past tense
+    #
+    # @param name [String] name of mismatch
+    # @param expected [PAnyType] expected type
+    # @param actual [PAnyType] actual type
+    #
+    def describe_mismatch(name, expected, actual)
+      describe_mismatch_present(name, expected, actual).gsub(/expects/, 'expected')
     end
 
     # @param subject [String] string to be prepended to the exception message
--- a/lib/puppet/pops/types/type_mismatch_describer.rb
+++ b/lib/puppet/pops/types/type_mismatch_describer.rb
@@ -72,7 +72,42 @@ module Types
     end
   end
 
+  # Module to handle present/past tense.
+  #
+  # All method names prefixed with "it_" to avoid conflict with Mocha expectations. Adding a method
+  # named 'expects' just doesn't work.
+  #
+  module TenseVariants
+    def it_expects(tense)
+      case tense
+      when :present
+        'expects'
+      else
+        'expected'
+      end
+    end
+
+    def it_does_not_expect(tense)
+      case tense
+      when :present
+        'does not expect'
+      else
+        'did not expect'
+      end
+    end
+
+    def it_has_no(tense)
+      case tense
+      when :present
+        'has no'
+      else
+        'did not have a'
+      end
+    end
+  end
+
   class Mismatch
+    include TenseVariants
     attr_reader :path
 
     def initialize(path)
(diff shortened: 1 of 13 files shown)


lib/puppet/pops/types/type_mismatch_describer.rb
```diff
@@ -66,7 +72,42 @@ def to_s
     end
   end
 
+  # Module to handle present/past tense.
+  #
```
hlindberg: Why is this needed? Don't understand.

spec/unit/pops/types/type_mismatch_describer_spec.rb
```diff
@@ -16,4 +59,62 @@
...
+      expect(subject.describe_signatures('function', [dispatch], args_tuple, :present)).to eq("'function' does not expect a block")
+    end
+  end
+
+  context 'when using past tense' do
+    let(:parser) { TypeParser.new }
+    let(:subject) { TypeMismatchDescriber.singleton }
+    it 'reports a missing parameter as "did not have a parameter"' do
+      t = parser.parse('Struct[{a=>String}]')
+      expect { subject.validate_parameters('v', t, {'a'=>'a','b'=>'b'}, false, :past) }.to raise_error(Puppet::Error, "v: did not have a parameter named 'b'")
```
hlindberg: past tense here looks odd...

hlindberg:
I don't get all the stuff with pasta and present tense - why are they not all in present tense?

thallgren:
Type assertion has always used past tense. Consider:

"The default value returned from the block has wrong type: Expected Integer, got String"

"The default value returned from the block has wrong type: Expects Integer, got String"

the latter doesn't look too good. This didn't use to be a problem because the TypeAsserter always used past tense. Now it uses the TypeMismatchDescriber. Incidentally, it is also used when validating parameter assignments. So we see things like:

"Function 'a' has no parameter named 'x'"

this could of course be changed to always used past tense:

"Function 'a' did not have a parameter named 'x'"

but as you noticed above, that too looks odd. There is also a common ground:

"Function 'a' parameter 'x' expects an Integer value, got String"

hlindberg:
It must say: "Function 'a' has no parameter named 'x'", since if it says "Function 'a' did not have a parameter named 'x'", it is undefined how far back in the past it did not have it, did it loose it, does it have it now, but not a while ago?
Past tense on the expectancy makes sense as it refers to the just failed assertion. 
Thus, this works: "The default value returned from the block has wrong type: expected an Integer value, got String".

So why not always "expected" ?

thallgren:
Parameter validation always uses present tense. The test is just there to verify that the mechanism works for past tense as well. I can remove that test if you don't want it there. The mechanism is generic and past tense is only used for failed assertions.

hlindberg:
I am not referring to the test. The handling of "tense" seems like something that should not be needed.
The condition is naturally in present tense since it was and still is the same way (e.g. "has no parameter"). The check just took place and it then expected. I don't understand what it is I am not getting here. Can you give an example of an error that would look bad with "expected" ?

thallgren:
Yes I can. But it requires an explanation.

Things are often nested. When multiple signatures are used, an none of them matches, each mismatch is reported as the signature 'expects' something. That "something" can be complex in itself and use the same tense as the top level or things will look odd.

You concluded above that "Function 'a' did not have a parameter named 'x'". This is one of the "somethings" I refer to in the previous paragraph. So past tense is out for a "something". Consequently, we cannot use "expected" here. All parameter mismatch reporting must be in present tense (as it always has been).

Type assertion is different. I gave examples of why it has to be past tense above.

hlindberg:
Can you show one that would look odd for the multiple signatures case? (it needs to state what it exects since it still expects the same signatures after the error is reported). I am not sure it really looks odd to say expected after that.

thallgren:
Does it matter that the tense is correct throughout?

hlindberg:
I am trying to understand what you consider correct tense and why all of this is needed.

thallgren:
Before I made the TypeAsserter take advantage of the TypeMismatchDescriber, the type asserter always used past tense and the TypeMismatchDescriber always used present tense. When I made the TypeAsserter use the TypeMismatchDescriber, some constructs became bad.

I'm sure there are several ways this can be resolved. I chose a way that made it possible to control whether or not the TypeMismatchDescriber should use past or present tens. This works in all cases. Unless there's a compelling reason not to do it this way, I would prefer if we let that solution stand rather than wasting more time coming up with other solutions.

hlindberg:
Seems a bad thing that they use different tense, and then fixing it with more code. Are people helped by messages being in different tense? I wondered why it was this way, your answer is basically that they are different because they are different and with a parameter they can continue to be different.  I got that, you would not have added the parameter otherwise.

I asked for examples to be able to asses if it is best to do what you did, or if messages of one kind should be changed to the other (or if that is worse), or if there is middle ground.  The example showing a change to past tense in the first part and present in the second did indeed look strange - also because the condition that is mentioned in the first part still exist. Maybe the messages need to make a difference because they have to make distinction between conditions that still exist, and things that were produced (and do now not exist) - in such sentences there can be a difference of tense.

Back to my question: is it really needed (or is the parameter there only for the reason to bridge between two different styles of messages)? I have not said one word about that I think we must change this, I just asked why. 

I am happy to help edit messages if that makes it a better solution for users. I don't think it is a waste of time to reduce complexity and making things clear.

thallgren:
OK. Let's leave this PR unmerged for now then. We have more important things to do. It would have been nice to get this in now (I obviously need some of it) but it's not that important.

hlindberg:
Stable is blocked at the moment yes. Has nothing to do with asking about the "tense" and messages.

hlindberg:
This needs a rebase. And we can then talk about the handling of "tense" in person.

hlindberg:
This was manually rebased (well git did the work), and merged to stable at f1a03019d906e7cee8a97c0a57cf25121603c867 and then on to master. Closing this PR.

PR #4725, opened 2016-02-26
(PUP-5862) Static catalog acceptance test: agent requests file content only when drift is detected.

seangriff:
This PR adds a few test steps to an existing acceptance test, puppet/acceptance//tests/direct_puppet/cached_catalog_remediate_local_drive.rb.  The new steps ensure that when the file has not drifted: (1) the client does not make any requests to the server (for file data or file metadata) and (2) the cached catalog application results in report to the server.

It works by doing the following:
- Remove reports from server's report directory, verifies "ls -l <report_directory> | wc -l" returns "0".
- Start a puppet run with options given to use the cached catalog and to specify a non-existent server (no_such_host) and to specify the actual server as the report server destination.  If the agent issues a request to the server (except for report delivery), the run will fail (and the test will fail) because the server identified doesn't exist.
- Verify a report was delivered to the server by checking that "ls -l <report_directory> | wc -l" returns "1" on the server.

[skip ci]

### DIFF
--- a/acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
+++ b/acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
@@ -44,16 +44,20 @@ test_name "PUP-5122: Puppet remediates local drift using code_id and content_uri
       '#{module_dir}/foo/files':;
     }
 
-    file { "site.pp":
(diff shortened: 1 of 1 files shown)


seangriff:
Blocked until stable is open for 4.4.1 additions

PR #4728, opened 2016-02-27
(PUP-5983) Enable interpolation of nested data in hiera data provider

thallgren:
This commit improves the Hiera data provider so that it performs
interpolation of data found in arrays and hashes. The interpolation of
such data is recursive. Prior to this commit, only interpolations found
in strings at the topmost level were interpolated.

### DIFF
--- a/spec/unit/data_providers/hiera_interpolation_spec.rb
+++ b/spec/unit/data_providers/hiera_interpolation_spec.rb
@@ -13,7 +13,7 @@ describe "Puppet::DataProviders::HieraInterpolate" do
   context 'when interpolating nested data' do
     let(:nested_hash) { { 'a' => { 'aa' => "%{alias('aaa')}" } } }
 
-    it 'should should produce an empty string for the interpolation' do
+    it 'produces a nested hash with arrays from nested aliases with hashes and arrays' do
       Puppet::Pops::Lookup.expects(:lookup).with('aaa', nil, '', true, nil, lookup_invocation).returns({ 'b' => { 'bb' => "%{alias('bbb')}" } })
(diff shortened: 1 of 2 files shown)


spec/unit/data_providers/hiera_interpolation_spec.rb
```diff
@@ -6,14 +6,23 @@
...
-    }
-    let(:interpolator) { Class.new { include Puppet::DataProviders::HieraInterpolate }.new }
+  let(:interpolator) { Class.new { include Puppet::DataProviders::HieraInterpolate }.new }
+  let(:scope) { {} }
+  let(:lookup_invocation) { Puppet::Pops::Lookup::Invocation.new(scope, {}, {}, nil) }
+
+  context 'when interpolating nested data' do
+    let(:nested_hash) { { 'a' => { 'aa' => "%{alias('aaa')}" } } }
+
+    it 'should should produce an empty string for the interpolation' do
```
hlindberg: "should should"

hlindberg:
Sure this is exactly what hiera does both behavior- and datatype- wise? (I would not be surprised if hiera did something odd).

thallgren:
Yes I'm sure. Hiera produces the exact same result given the same values. I have a test for that in the Hiera code-base. Do you want me to submit that as a (maint) commit?

hlindberg:
If you have an additional test for hiera that would be great (there are way too few tests there). Just push it as a (maint) as you suggested, no need to review. 

However, stable branches are in lockdown until release, so both this merge and the push of the (maint) to hiera has to wait a bit.

hlindberg:
Good to merge when block is lifted.

hlindberg:
Ping @thallgren. This needs a rebase and you can then merge it to stable.

PR #4729, opened 2016-02-28
(PUP-2695) Deprecate reopening and merging classes other than "top scope"

hlindberg:
Merge PUP-5889 before this PR as this depends on the --strict setting being present.

This PR errors or warns under the control of --strict if a user redefines/reopens a class. This is still allowed for the class named '' (a.k.a 'main'). as that is the compilers way of merging all top scope logic.


thallgren:
Needs a rebase.

hlindberg:
Should not be needed as the commits are the same - but will do one anyway.

PR #4731, opened 2016-02-28
(PUP-5862) Only request file content when remediating drift 

seangriff:
…t only when drift is detected.

This PR adds a few test steps to an existing acceptance test,
puppet/acceptance//tests/direct_puppet/cached_catalog_remediate_local_drive.rb.
The new steps ensure that when the file has not drifted:
(1) the client does not make any requests to the server (for
file data or file metadata) and (2) the cached catalog application
results in report to the server.

It works by doing the following:
- Remove reports from server's report directory,
  verifies "ls -l <report_directory> | wc -l" returns "0".
- Start a puppet run with options given to use the cached
  catalog and to specify a non-existent server (no_such_host)
  and to specify the actual server as the report server
  destination.  If the agent issues a request to the server
  (except for report delivery), the run will fail (and the
  test will fail) because the server identified doesn't exist.
- Verify a report was delivered to the server by checking that
  "ls -l <report_directory> | wc -l" returns "1" on the server.

[skip ci]

### DIFF
--- a/acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
+++ b/acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
@@ -26,22 +26,22 @@ test_name "PUP-5122: Puppet remediates local drift using code_id and content_uri
 
   step "Create a module and a file with content representing the first code_id version" do
     apply_manifest_on(master, <<MANIFEST, :catch_failures => true)
-    File {
-      ensure => directory,
-      mode => "0750",
-      owner => #{master.puppet['user']},
-      group => #{master.puppet['group']},
-    }
+File {
+  ensure => directory,
+  mode => "0750",
+  owner => #{master.puppet['user']},
+  group => #{master.puppet['group']},
+}
 
-    file {
-      '#{basedir}':;
-      '#{basedir}/environments':;
-      '#{basedir}/environments/production':;
-      '#{basedir}/environments/production/manifests':;
-      '#{module_dir}':;
-      '#{module_dir}/foo':;
-      '#{module_dir}/foo/files':;
-    }
+file {
+  '#{basedir}':;
+  '#{basedir}/environments':;
+  '#{basedir}/environments/production':;
+  '#{basedir}/environments/production/manifests':;
+  '#{module_dir}':;
+  '#{module_dir}/foo':;
+  '#{module_dir}/foo/files':;
+}
 MANIFEST
   end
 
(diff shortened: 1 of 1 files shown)


seangriff:
Blocked waiting for go-ahead to check into stable for 4.4.1.

joshcooper:
Looks good @seangriff, just had a couple of nits:

* Could you change the commit summary to use the imperative form, e.g. (PUP-5862) Only request file content when remediating drift
* It looks like there is some extra trailing whitespace. From your topic branch, you should be able to run `git show --check`. You can also setup vim to show you trailing whitespace. Amend your commit and force push to your topic branch
* Typo in https://github.com/puppetlabs/puppet/pull/4731/files#diff-faff23644bfa490b5302fd35c267250cR81
* The test deletes all of the reports, runs the agent once, and requires that only a single report is created. That assumes there are no other agents running in the background (as a service), which I think is ok assumption because beaker should install puppet-agent, but leave the puppet service stopped.

acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
```diff
@@ -51,41 +51,69 @@
...
+        "--report_server", master.hostname,
+        "--onetime",
+        "--no-daemonize",
+        "--detailed-exitcodes",
+        "--verbose"
+      )).stderr
+      assert_equal(r, "", "Fail: Did agent try to contact server?")
+
+      step "Verify report was delivered to server"
+      r = on(master, "ls /opt/puppetlabs/server/data/puppetserver/reports/#{agent.hostname} | wc -l").stdout.chomp
```
joshcooper: The path on disk will be based on the agent's node name. By default, that will be the agent's fqdn (from the perspective of the agent), which may be different from the name that the beaker uses to ssh to the agent (which is `agent.hostname`). When using the vmpooler the agent's `fqdn` and `agent.hostname` are the same, but there are cases where they may not be.

In this test, I think we want to use `agent.node_name` which will ask the node what it thinks its name is, and should always match what ends up on disk. /cc @johnduarte

joshcooper:
This fails on windows, because puppet-agent#stable is currently pinned to [puppet_for_the_win#4.3.2](https://github.com/puppetlabs/puppet-agent/blob/stable/configs/components/windows_puppet.json#L3). The older version sets `pluginsync = true` to puppet.conf. As a result, the agent attempts to pluginsync from the non-existent master, and the test correctly fails the test:

```
 * Verify puppet run without drift does not make file request from server

  atsayjd6uuj5fvs.delivery.puppetlabs.net (windows2012r2-64-1) 15:40:55$ cmd.exe /c puppet agent --use_cached_catalog --server no_such_host --report_server qq8gx739m1hkdqc.delivery.puppetlabs.net --onetime --no-daemonize --detailed-exitcodes --verbose
    Warning: Setting 'pluginsync' is deprecated.
       (at C:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/defaults.rb:1607:in `block in <module:Puppet>')
    Info: Using cached catalog from environment 'production'
    Info: Retrieving pluginfacts
    Error: /File[C:/ProgramData/PuppetLabs/puppet/cache/facts.d]: Failed to generate additional resources using 'eval_generate': getaddrinfo: No such host is known.
    Info: Retrieving plugin
    Error: /File[C:/ProgramData/PuppetLabs/puppet/cache/facts.d]: Could not evaluate: Could not retrieve file metadata for puppet:///pluginfacts: getaddrinfo: No such host is known.
    Error: /File[C:/ProgramData/PuppetLabs/puppet/cache/lib]: Failed to generate additional resources using 'eval_generate': getaddrinfo: No such host is known.
    Error: /File[C:/ProgramData/PuppetLabs/puppet/cache/lib]: Could not evaluate: Could not retrieve file metadata for puppet:///plugins: getaddrinfo: No such host is known.
    Info: Applying configuration version '1457134850'

  atsayjd6uuj5fvs.delivery.puppetlabs.net (windows2012r2-64-1) executed in 20.41 seconds
  Warning:
  Fail: Did agent try to contact server?.
  --- expected
  +++ actual
  @@ -1,7 +1 @@
  -"\e[1;33mWarning: Setting 'pluginsync' is deprecated.
  -   (at C:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/defaults.rb:1607:in `block in <module:Puppet>')\e[0m
  -\e[1;31mError: /File[C:/ProgramData/PuppetLabs/puppet/cache/facts.d]: Failed to generate additional resources using 'eval_generate': getaddrinfo: No such host is known. \e[0m
  -\e[1;31mError: /File[C:/ProgramData/PuppetLabs/puppet/cache/facts.d]: Could not evaluate: Could not retrieve file metadata for puppet:///pluginfacts: getaddrinfo: No such host is known. \e[0m
  -\e[1;31mError: /File[C:/ProgramData/PuppetLabs/puppet/cache/lib]: Failed to generate additional resources using 'eval_generate': getaddrinfo: No such host is known. \e[0m
  -\e[1;31mError: /File[C:/ProgramData/PuppetLabs/puppet/cache/lib]: Could not evaluate: Could not retrieve file metadata for puppet:///plugins: getaddrinfo: No such host is known. \e[0m
  -"
  +""
```

We need to update puppet-agent#stable to use puppet_for_the_win#4.4.0, before this can be merged.

joshcooper:
I'll retry this now that puppet-agent PR was merged.

joshcooper:
Passed using the following on redhat7 and windows (had to remember to remove puppet/acceptance/tmp which had an older MSI):

```
$ bundle exec rake ci:test:aio SHA=7282f093e3c77b3382938a1c2e22d12b59b7973e SUITE_VERSION=1.3.5.353.g7282f09 TESTS=tests/direct_puppet/cached_catalog_remediate_local_drift.rb
```

joshcooper:
Also note this PR uses `agent.node_name` so we shouldn't have issues like https://tickets.puppetlabs.com/browse/PUP-6017 when this test is run on Solaris

PR #4732, opened 2016-02-29
(PUP-5903) Add local type alias to 4.x function api

hlindberg:
This adds the ability to add type aliases that are local to a function. See commit for example.

At this point, this is up for a preliminary review only and there are yet no tests of the new behavior.

1. The `TypeParser` is modified so it can operate on a `Scope` or a `Loader` (there is no scope available that would find the loader in question, so this was required).
2. The `RubyFunctionInstantiator` provides a `Binding` where loader is bound when doing the eval of the loaded function .rb file
3. The `create_function` method on `Function` picks up the loader bound in the `Binder` and delegates to a new function `create_loaded_function` (name not very good). This method is useful when unit testing, but should otherwise not be used directly by users.
4. ~~A user that wants to use type alias must use create_loaded_function instead of create_function, and must supply the loader from the Binding (i.e. just give it as an argument, see commit message).~~
5. A new builder was added to handle the `local_types` clause in the  function definition. It creates a new nested loader that the Function class remembers.
6. The Function class new operator is overloaded to set the loader from local_types instead of the user given one in case a more specialized loader should be used.

This design was favoured as it only includes one piece of magic - relaying the "loader" via divine powers. This controlling behavior from a far is not ideal, but it is used in a controlled situation.

Is it fine that if there is a local_types clause there will always be a nested loader for local types even if user did not add any local types.

### DIFF
--- a/lib/puppet/functions.rb
+++ b/lib/puppet/functions.rb
@@ -153,6 +153,28 @@ module Puppet::Functions
   #
   # @api public
   def self.create_function(func_name, function_base = Function, &block)
+    # Ruby < 2.1.0 does not have method on Binding, can only do eval
+    # and it will fail unless protected with an if defined? if the local
+    # variable does not exist in the block's binder.
+    #
+    loader = block.binding.eval('loader_injected_arg if defined?(loader_injected_arg)')
+    create_loaded_function(func_name, loader, function_base, &block)
+  end
+
+  # Creates a function in, or in a local loader under the given loader.
+  # This method should only be used when manually creating functions 
+  # for the sake of testing. Functions that are autoloaded should
+  # always use the `create_function` method and the autoloader will supply
+  # the correct loader.
+  #
+  # @param func_name [String, Symbol] a simple or qualified function name
+  # @param loader [Puppet::Pops::Loaders::Loader] the loader loading the function
+  # @param block [Proc] the block that defines the methods and dispatch of the
+  #   Function to create
+  # @return [Class<Function>] the newly created Function class
+  #
+  # @api public
+  def self.create_loaded_function(func_name, loader, function_base = Function, &block)
     if function_base.ancestors.none? { |s| s == Puppet::Pops::Functions::Function }
       raise ArgumentError, "Functions must be based on Puppet::Pops::Functions::Function. Got #{function_base}"
     end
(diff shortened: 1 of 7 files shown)


thallgren:
The one thing that I react to instantly is that a user that wants to benefit from this must use a different create function and provide a loader. Why is that? Isn't it possible to create such a loader on the fly in case there's a need for it? I don't mind having another function that enables some other user-defined loader to be used but I think that's a corner case that very few will explore.

thallgren:
Perhaps one solution could be to defer all evaluation of parameter types until the function is instantiated.

hlindberg:
I agree, the need to pass the "loader" on like that is not satisfactory. The problem is that the binding makes it available to the given block, but when I need to pick it up (before evaluating anything since it may be the call to 'local_types' I had trouble finding it anywhere. I did a number of trials of using block.binding.get_local_variable. Maybe the problem is that just did not find the right spot - i.e. in "create_function" but before it evaluates that block.

hlindberg:
meh, failures on Ruby 1.9.3 and 2.0.0. Have to look into if what I want is possible there.

hlindberg:
Found a way around the problem that should work on Ruby 1.9.3 and later

thallgren:
I think this looks good now. Wish I had it when I wrote the dispatchers for the lookup function. I made some nitpick comment. A rebase is needed.

hlindberg:
Some follow up in other tickets will be needed in the type mismatch describer to make the messages understandable - the user cannot see the local types anywhere (at least not easily).

hlindberg:
The local_types support must also have a way to use the defined types in the logic - i.e. find the loader. (it is already but not in a nice way as you have to use `self.class.loader.load(:type, ''typename') to find one.

hlindberg:
Part of this was cherry picked as it solves other problems targeted for stable. Some tests were added there. @thallgren and I agree this + a couple of additional tests makes this complete.

nedi00:
The problem is that the binding makes it available to the given block, but when I need to pick it up (before evaluating anything since it may be the call to 'local_types' I had trouble finding it anywhere. I did a number of trials of using block.binding.get_local_variable. Maybe the problem is that just did not find the right spot - i.e. in "create_function" but before it evaluates that block.??

thallgren:
This PR is superseded by  [PR-4768](/puppetlabs/puppet/pull/4768)

thallgren:
@nedi00 , sounds like you're using the create_function in a scenario that we haven't anticipated (i.e. not using the auto loader). The unit tests in the new PR may be of help but it's nothing that I'd recommend for a production system as all the things related to accessing the loader in a dispatch is api-private.

hlindberg:
@nedi00 hit me up on IRC helindbe if you tried using this before it was merged and you had issues. If this was intended as advice, I think we figured out how it works - the solution looks good now.

PR #4734, opened 2016-02-29
DO NOT MERGE! Serialization prototype

thallgren:
This is work in progress. PR is for review purposes only.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ gem "puppet", :path => File.dirname(__FILE__), :require => false
 gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 2.0', '< 4'])
 gem "hiera", *location_for(ENV['HIERA_LOCATION'] || ['>= 2.0', '< 4'])
 gem "rake", "10.1.1", :require => false
+gem "msgpack", "~> 0.7.6", :require => false
 
 group(:development, :test) do
   gem "rspec", "~> 3.1", :require => false
@@ -65,7 +66,6 @@ group(:extra) do
   gem "net-ssh", '~> 2.1', :require => false
   gem "puppetlabs_spec_helper", :require => false
   gem "tzinfo", :require => false
-  gem "msgpack", :require => false
 end
 
 require 'yaml'
--- a/lib/puppet/pops/parser/locator.rb
+++ b/lib/puppet/pops/parser/locator.rb
@@ -8,7 +8,7 @@ class Locator
   #
   def self.locator(string, file, index = nil, char_offsets = false)
     if char_offsets
-      LocatorForChars.new(string, file, index);
+      LocatorForChars.new(string, file, index)
     else
       Locator19.new(string, file, index)
     end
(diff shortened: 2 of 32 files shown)


lib/puppet/pops/parser/locator.rb
```diff
@@ -262,6 +266,28 @@ def extract_text(offset, length)
...
+      attributes => {
+        string => String[1],
+        file => String,
+        line_index => {
+          type => Optional[Array[Integer]],
+          value => undef
+        }
+      }
+    }
+      OBJECT
```
hlindberg: don't get this `OBJECT` can you elaborate?

Gemfile
```diff
@@ -27,6 +27,7 @@ gem "puppet", :path => File.dirname(__FILE__), :require => false
 gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 2.0', '< 4'])
 gem "hiera", *location_for(ENV['HIERA_LOCATION'] || ['>= 2.0', '< 4'])
 gem "rake", "10.1.1", :require => false
+gem "msgpack", "~> 0.7.6", :require => false
```
joshcooper: We'll need to resolve the issues we had with semantic_puppet before adding a hard dependency on msgpack, e.g. puppet-agent, puppetserver, Gemfile, ext/project_data.yaml, etc.

lib/puppet/pops/parser/locator.rb
```diff
@@ -262,6 +266,28 @@ def extract_text(offset, length)
...
+      @type = Pcore.add_object('Pcore::AST::Locator', locator_def, loader).resolve(Types::TypeParser.singleton, loader)
+      ir.register_implementation('Pcore::AST::Locator', self, loader)
+    end
```
hlindberg: Why is AST in the {{Pcore}} name space as opposed to say {{Puppet}} ? While it is expressed in Pcore, it isn't Pcore itself is it?

lib/puppet/pops/types/p_runtime_type.rb
```diff
@@ -3,8 +3,22 @@ module Types
 
 # @api public
 class PRuntimeType < PAnyType
+
   TYPE_NAME_OR_PATTERN = PVariantType.new([PStringType::NON_EMPTY, PTupleType.new([PRegexpType::DEFAULT, PStringType::NON_EMPTY])])
 
+  def self.register_ptype(loader, ir)
+    create_ptype(loader, ir, 'ScalarType',
```
hlindberg: ? is a runtime Scalar? (Don't get this)

hlindberg:
In general this looks really good. Some things are hard to follow due to the nature of how things compose into a framework to support serialization.
At this point I just had a couple of questions regarding namespacing and some things I did not understand.

lib/puppet/pops/parser/locator.rb
thallgren: no, but I can remove it ;-)

PR #4735, opened 2016-02-29
(PUP-5959) Add test to ensure static catalog file params match defaults

whopper:
When inlining file metadata, we sometimes hardcode the default values
for the `links` and `source_permissions` parameters to match
the real file type defaults.

This commit adds a new test case to ensure that the values we've
hardcoded actually match those defaults. Should they ever change,
this test will fail, thus notifying us that our hardcoded values
need to change.

This is being done with a test rather than retrieving the defaults
during compilation due to the fact that determining said defaults
requires instantiated file resources, which depend on knowing which
file provider to use. Creating a new dummy file resource on the
master is clunky and inelegant.

### DIFF
--- a/spec/unit/indirector/catalog/compiler_spec.rb
+++ b/spec/unit/indirector/catalog/compiler_spec.rb
@@ -493,6 +493,34 @@ describe Puppet::Resource::Catalog::Compiler do
       expect(catalog.recursive_metadata).to be_empty
     end
 
+    it "uses file parameters which match the true file type defaults" do
+      catalog = compile_to_catalog(<<-MANIFEST, node)
+        file { '#{path}':
+          ensure => file,
+          source => '#{source}'
+        }
+      MANIFEST
+
+      if Puppet::Util::Platform.windows?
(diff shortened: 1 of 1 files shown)


whopper:
Letting appveyor test the Windows side of this.

spec/unit/indirector/catalog/compiler_spec.rb
```diff
@@ -493,6 +493,34 @@ def stubs_directory_metadata(relative_path)
...
+        default_file = Puppet::Type.type(:file).new(:name => 'C:\defaults')
+      else
+        default_file = Puppet::Type.type(:file).new(:name => '/defaults')
+      end
+
+      metadata = stubs_file_metadata(checksum_type, checksum_value, 'modules/mymodule/files/config_file.txt')
+
+      options = {
+        :environment => catalog.environment_instance,
+        :links => default_file[:links],
```
whopper: This could also be done in a few existing test cases, though that may muddy what they're testing and why they fail.

joshcooper:
Looks like a rebase is needed?

whopper:
@joshcooper rebased.

PR #4737, opened 2016-03-01
(PUP-5992) Support custom status command for systemd provider

liaoishere:
The service provider on systemd systems currently only use systemctl
command to check to service status. The patch aims to support customize
status command when using service type.

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -114,15 +114,6 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
     end
   end
 
-  def status
-    begin
-      systemctl("is-active", @resource[:name])
-    rescue Puppet::ExecutionFailure
-      return :stopped
-    end
-    return :running
-  end
-
   def enable
     self.unmask
     output = systemctl("enable", @resource[:name])
@@ -158,6 +149,10 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
 
   def stopcmd
     [command(:systemctl), "stop", @resource[:name]]
+  end
+
+  def statuscmd
+    [command(:systemctl), "is-active", @resource[:name]]
   end
 end
 
--- a/spec/unit/provider/service/systemd_spec.rb
+++ b/spec/unit/provider/service/systemd_spec.rb
@@ -253,16 +253,26 @@ describe Puppet::Type.type(:service).provider(:systemd) do
   # Note: systemd provider does not care about hasstatus or a custom status
   # command. I just assume that it does not make sense for systemd.
   describe "#status" do
-    it "should return running if active" do
+    it "should return running if if the command returns 0" do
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/service/systemd.rb
```diff
@@ -114,15 +114,6 @@ def debian_enabled?(svc_info)
     end
   end
 
-  def status
-    begin
-      systemctl("is-active", @resource[:name])
-    rescue Puppet::ExecutionFailure
```
peterhuene: This is now relying on `texecute` from base service provider in `service.rb` to catch `Puppet::ExecutionFailure`, which calls `fail`.  I suspect this is done this way on purpose to report `:stopped` when `is-active` returns a non-zero code, perhaps?

peterhuene:
Hi @liaoishere.  Thanks for the contribution!  Some specs are failing where mocks are being done to `status` that would now need to be replaced with stubs for this new `statuscmd`. 

Also see my question above regarding the new behavior when the command fails.

Additionally, this needs a rebase to remove the merge commit.

Thanks!

liaoishere:
Hi @peterhuene. Thinks for your comment. `status` is definded in base service provider in `base.rb` and will call `$CHILD_STATUS.exitstatus` to report `:stopped` when command returns a non-zero code. See code in:

> https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/service/base.rb#L55-L72

 I don't know why the test fails, do you have any advice?

peterhuene:
Hi @liaoishere:

Ah, I missed where `false` was being passed into `ucommand`, which will result in no possibility that the rescue here will be executed: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/service/service.rb#L27

So behaviorally this change seems fine; ignore my previous comment.

With respect to the spec failures, the problem is that the two failing specs stub a call to `systemctl` that is no longer made:

```ruby
provider.expects(:systemctl).with('is-active', 'sshd.service').returns 'active'
```

This should probably now be:
```ruby
provider.expects(:execute).with(['/bin/systemctl','is-active','sshd.service'], :failonfail => false, :override_locale => false, :squelch => false, :combine => true).returns 'active'
```

And similar for the other failing spec.

liaoishere:
Hi, @peterhuene 

Thanks for your help and guidance :). I've updated the unit test code. Help me review if you have time.

peterhuene:
Thanks, @liaoishere.  We've accepted the change and it has been merged into master.  It should show up in a future release of Puppet.  Thanks again for the contribution!

peterhuene:
Hi @liaoishere.  Sorry one last thing.  Normally we have a bot that comments on our PRs for users to sign our CLA (Contributor License Agreement).  Unfortunately it looks like that bot is not running properly at the moment.  Would you mind signing the agreement here: https://cla.puppetlabs.com.  Thanks!

PR #4738, opened 2016-03-01
(maint) Use /bin/bash instead of /bin/sh in direct puppet test

whopper:
On some platforms, `/bin/sh` is actually a link to `/bin/dash`
which doesn't handle conditional statements in the same way as
bash. This commit updates the static catalog utils code id and
code content scripts to use `bash` instead of `sh`.

[skip-ci]


whopper:
Running this through CI with an ubuntu master.

joshcooper:
:+1:

PR #4739, opened 2016-03-01
(maint) Add default TEST_TARGET when none specified

MikaelSmith:
Provides default test targets to run on both Windows and Linux agents
when TEST_TARGET isn't specified. Provides a sane set of defaults for
verifying that tests will work.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -16,6 +16,10 @@ CLEAN.include('*.tar', REPO_CONFIGS_DIR, 'merged_options.rb')
 # then assume we can depend on them here
 EPEL_MIRROR = ENV['BEAKER_EPEL_MIRROR']
 
+# Default test target if none specified
+DEFAULT_MASTER_TEST_TARGET = 'redhat7-64m'
+DEFAULT_TEST_TARGETS = "#{DEFAULT_MASTER_TEST_TARGET}a-windows2012r2-64a"
+
 module HarnessOptions
   defaults = {
     :type => 'git',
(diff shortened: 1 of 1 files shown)


MikaelSmith:
ping @joshcooper @er0ck 

Seems like it could be useful to have a sane default.

acceptance/Rakefile
```diff
@@ -16,6 +16,9 @@ CLEAN.include('*.tar', REPO_CONFIGS_DIR, 'merged_options.rb')
 # then assume we can depend on them here
 EPEL_MIRROR = ENV['BEAKER_EPEL_MIRROR']
 
+# Default test target if none specified
+DEFAULT_TEST_TARGET = 'centos7-64a-windows2012r2-64a'
```
joshcooper: We were talking about this in the context of puppetserver acceptance, where they run an agent on the master. What do you think about defaulting to that configuration (as opposed to one master and two different agents)? That way we have a workflow for making sure we don't break them with each new acceptance test.

It does make the logic a little more complicated, since you'd have to add the `agent` role to the `MASTER_TEST_TARGET`... thoughts?

MikaelSmith: That seems ok too. I'll take a crack at it.

MikaelSmith:
Updated.

whopper:
:+1: I ran this locally and it worked! Tests ran successfully against an agent on the  Redhat master and a Windows agent. Do we want to hold off on merging until 4.4.0 is out?

PR #4740, opened 2016-03-01
(PUP-5994) Fix module tar file extraction with long pathnames

glennsarti:
The MiniTar extraction was failing intermittently for modules with long
filenames.  This was due to using entry.name intead of entry.full_name.
The full_name property is derived from the the filename prefix plus the
filename property.  By using the name property some files and directories
were silently dropped from the extraction.

This commit changes the validation to use full_name and adds a test to
extract an example tarball to enusre these files are indeed extracted.

Ustar Format:
https://en.wikipedia.org/wiki/Tar_(computing)#UStar_format
Field Offset 345 - Filename prefix

full_name function:
https://github.com/atoulme/minitar/blob/master/lib/archive/tar/minitar.rb#L547-L553


joshcooper:
Nice debugging @glennsarti :+1:

spec/integration/module_tool/tar/mini_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'spec_helper'
+require 'puppet/module_tool'
+require 'puppet_spec/files'
+
+describe Puppet::ModuleTool::Tar::Mini, :if => (Puppet.features.minitar? and Puppet.features.zlib?) do
```
Iristyle: Prefer to use `&&` instead of `and` here as the precedence rules are different.  `and` is intended to be a control flow statement (I presume this was copied from the existing test at https://github.com/puppetlabs/puppet/blob/master/spec/unit/module_tool/tar/mini_spec.rb#L4 but we should probably not copy that part)

http://stackoverflow.com/a/1426835

Iristyle:
Could we also mention in the commit message which operating systems are generally impacted here.  The usage of these code paths are guided by the `minitar` and `zlib` features, which only impacts certain OSes.

johnduarte:
No issues with AIX or Solaris since the forge SSL cert chain is not supported on these platforms for puppet 3.x.

joshcooper:
@johnduarte do we support installing from a local tar.gz on those platforms and puppet version?

johnduarte:
@joshcooper yes we do. Thanks for the reminder. This works as expected with this change.

```
root@d3673jyq1kmor73:~# puppet module install --ignore-dependencies puppetlabs-motd.tar.gz
Notice: Preparing to install into /etc/puppet/modules ...
Notice: Installing -- do not interrupt ...
/etc/puppet/modules
└── puppetlabs-motd (v1.4.0)
root@d3673jyq1kmor73:~# facter os
{"name"=>"Solaris", "family"=>"Solaris", "release"=>{"major"=>"11", "minor"=>"1", "full"=>"11.1"}}
```

glennsarti:
@johnduarte Can you please try with the puppetlabs-dsc module as this was the module that tripped this error.  The motd module doesn't seem to have long pathnames.

johnduarte:
@glennsarti results look good on Solaris 11 with the puppetlabs-dsc module

```
root@hc2o7u21q5e880r:~# puppet module install --ignore-dependencies puppetlabs-dsc-1.0.0.tar.gz
Notice: Preparing to install into /etc/puppet/modules ...
Notice: Created target directory /etc/puppet/modules
Notice: Installing -- do not interrupt ...
/etc/puppet/modules
└── puppetlabs-dsc (v1.0.0)
root@hc2o7u21q5e880r:~# find /etc/puppet/modules/dsc/ -name *.psm1 | wc -l
     245

root@hc2o7u21q5e880r:~# facter os
{"name"=>"Solaris", "family"=>"Solaris", "release"=>{"major"=>"11", "minor"=>"1", "full"=>"11.1"}}
```

spec/integration/module_tool/tar/mini_spec.rb
glennsarti: Fixed

glennsarti:
Commit message modified

PR #4741, opened 2016-03-02
(maint) Correct error in the Puppet lookup command test.

seangriff:
Before this fix, this test fails when there is more than
one agent.  The beaker variable, agent is treated was treated as
a scalar.  When there is more than one agent, the beaker variable
is an array.  The fix copies the agent to a scalar variable, checking
if it is an array, in which case the scalar variable takes the value
of one element of the array.

### DIFF
--- a/acceptance/tests/parser_functions/puppet_lookup_cmd.rb
+++ b/acceptance/tests/parser_functions/puppet_lookup_cmd.rb
@@ -6,12 +6,11 @@ test_name "Puppet Lookup Command"
 
 ### @testroot = "/etc/puppetlabs"
 @testroot = master.tmpdir("#{@module_name}")
-
 @coderoot = "#{@testroot}/code"
 @confdir = "#{@testroot}/puppet"
-
-@mastername = on(master, facter('fqdn')).stdout.chomp
-@agentname = on(agent, facter('fqdn')).stdout.chomp
+@agent = [agent].flatten.find { |host| host != master }
+@mastername = master.hostname
+@agentname = @agent.hostname
 
 @master_opts = {
   'main' => {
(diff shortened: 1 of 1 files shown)


acceptance/tests/parser_functions/puppet_lookup_cmd.rb
```diff
@@ -10,8 +10,9 @@
 @coderoot = "#{@testroot}/code"
 @confdir = "#{@testroot}/puppet"
 
-@mastername = on(master, facter('fqdn')).stdout.chomp
-@agentname = on(agent, facter('fqdn')).stdout.chomp
+@agent = agent.is_a?(Array) ? (agent.count > 1 ? agent[1] : agent[0]) : agent  # agent[0] might be on master machine.
```
joshcooper: mentioned in meatspace, better to do something like:

```ruby
@agent = [agent].flatten.find { |host| host != master }
```
rather than relying on the ordering of which nodes have the master role.

acceptance/tests/parser_functions/puppet_lookup_cmd.rb
```diff
@@ -6,12 +6,11 @@
 
 ### @testroot = "/etc/puppetlabs"
 @testroot = master.tmpdir("#{@module_name}")
-
 @coderoot = "#{@testroot}/code"
 @confdir = "#{@testroot}/puppet"
-
-@mastername = on(master, facter('fqdn')).stdout.chomp
-@agentname = on(agent, facter('fqdn')).stdout.chomp
+@agent = [agent].flatten.find { |host| host != master }
```
camlow325: Does this mean that the test would only run on at most one non-master agent?  For the case where we would be running tests with a configuration in which more than one non-master agent is present in the configuration, would we consider it a problem that only one of those agents would be exercised?  Also, what would happen if this test were run in a single agent configuration where that one agent is running on the master?  Would the test blow up in that case?

AFAIK, no teams are currently running with either of the configurations that I mentioned above in their standard CI pipelines but it would seem good to future-proof this for cases where this might be done at some point.

camlow325:
Seems like this PR might need to be re-based on the latest stable branch before it could be merged?

acceptance/tests/parser_functions/puppet_lookup_cmd.rb
MikaelSmith: Agreed, seems strange this test assumes a single agent. Most of them are written to apply iteratively over agents.

joshcooper:
Superceded by #4756

PR #4742, opened 2016-03-02
(PUP-5844) Add Object type to the Puppet Type system

thallgren:
This commit adds a new Object type to the Puppet Type system. An
Object may have attributes, functions, equality, checks, and
annotations declared. It may also inherit from a parent type.

The Object type can be seen as the interface of an implementation, in
that it describes the contract that the implementation must fulfill.
The type will however not contain the actual implementation logic.

Things not included in this commit:
1. Logic around 'checks' (awaits spec)
2. Logic around 'annotations' (awaits spec)
3. The new-operator. Will be implemented together with 'checks' logic


hlindberg:
The approach is good. Construct is clear and if you get the other types there should be no problem reading and understanding the object type.

The ambition on including keywords went a bit to far, and we can not translate a literal default to a string value. That makes it impossible to use with the meaning 'default' in matches etc. It has to stay a symbol.

thallgren:
Regarding the keywords, all those keywords are allowed as attribute names. If we limit them further then we must somehow motivate why they are limited in one place but not in another. I though the approach I took was the most consistent one.

Also, 'default' is not treated as a string in general. It's only when it's used as a key in a hash. I don't see what use-cases there are for doing that other than when you indeed want a literal string.

hlindberg:
You can write expressions that result in a value in a key in a hash, therefore we cannot treat keywords that are part of value producing expressions in a special way. Changing what default means (even if "just" in hash key means an API change), and then this sails off to Puppet 5. Having default evaluate to string in some cases but not others is just asking for trouble IMHO.

Literal default has special meaning in arrays when used with match operator,  in case and in selector - it is basically the same as an "ignore". It cannot be a string there.

So yes, each choice I made was rational.

The situation is different when it comes to attributes - you cannot use expressions there.

Remember when I wanted us to use proper syntax instead as that would allow us any meaning associated with keywords - it was for exactly these reasons.

Making the definition keywords into bare words when in certain positions is harmless, does not break API, and those keywords have meaningful names (elsif and friends not so much) - it is bad enough already that you can do $if, and have attributes named 'if' that you cannot reference with a bare word, cannot differentiate from an expression in short form interpolation (without extra gymnastics) etc. etc.

Lets not add additional corners to paint ourselves into than we absolutely have to.

Oh, and BTW now that I think about this more, giving the type keyword special powers in a hash key position means you cannot use type(x) as the key. That introduces another corner special case. I do not like those. 

I am much more in favour of an operator that removes special meaning from everything (we discussed this earlier - the macro - for the price of one operator you get the tokens instead of the AST, or that there is other proper syntax/grammar. Another alternative is to make S-expressions possible, they are very suitable for these kinds of things (but there are other issues with those).

It starts out nicely:
```
ObjectType[&(
  attributes ( 
    (name String (default "hello"))
    (birthdate Timestamp)
    ( ... )
}]
```
but gets trickier with functions and complex things, lots of parentheses, and it is different.

We could also (perhaps) make hashes used in brackets after QualifiedReferences special. Perhaps also only when inside of a type T = T expression (and potentially other places). i.e. where the types need to be literal anyway. In normal code, you want expressions to work when creating a type you are using on the fly for matching.

We can make Object a keyword - we are reserving it anyway - but then, why are the rules here different than elsewhere.

We can add an unary operator that means literal hash - no expressions allowed. Then it is your choice if you want to quote keywords or use the alternative form. We can do the same for array. Array and hash then become like S-expressions. e.g. &{ } and &[], or ^{} ^[ ] (inside of those all { }, and [ ] are equally literal (no expressions), and you can even allow any token (*, + - etc) and you don't need commas between elements (optional , like in clojure). When evaluated they become hashes and arrays, all barewords and qualified references become instances of a PWordType (we can add additional types as runtime objects without things blowing up), strings, numbers and regexps are instances of the same types as now. This means it is easy to check the correctness of such an S-expression with a type.

Some examples...
```
# like this (almost like "now")
ObjectType[^{ attributes => { name => { type => String, default => "n.n" }, ...}
# More like an S expression (but not like existing types)
ObjectType[^{ attributes =>  [ 
  [name [type String] [default 10]],
  [birthdate Timestamp],
  ...
  ]}]

# variations that would also work (needs extra parsing)
  [name, type String, default 10],
```

After having written all that, there is a simpler solution - the lexer can emit different tokens for 'type'. If 'type' is followed by /\s*[([A-Z] / it is a keyword like now else it is a bare word. Then we can use it and the other definition keywords like you started in this PR.

thallgren:
We will need that change in the lexer regardless I think, because this is a syntax error regardless if type is allowed as a hash key:
```
puppet apply -e 'notice { type(3) => x }'
```
In general, I'm against introducing new syntax scopes using keywords and operators where the normal syntax no longer applies. It makes the language less clean and harder to learn. I like it better if we can resolve this without new inventions in that direction.

hlindberg:
If the problem is reported as a Syntax error (from parser), it is unlikely caused by the transform_resource_wo_title method (IIRC it does nothing, transforms, or errors - I don't remember if it fakes a Syntax error (that is normally just the generated parser that does that). Separate issue though - did you file one?

hlindberg:
Looking and thinking about 'type' again. If this
```
{ type => ... }
```
treats type special, without interfering with its use anywhere else, then it is fine. I guess that having the special case in the same Racc rule makes the single TYPE followed by '=>' win over the alternatives. (That is sort of the same as what I suggested regarding fixing this in the lexer).

Naturally, it cannot have special meaning here:
```
$x[type]
```
You would have to quote type there. That is fine.

thallgren:
The PR is updated now to address what has been discussed. I changed the commits though (some had incorrect comments anyway) and there will be more tests coming.

hlindberg:
great, np with rewriting the commits - did not really do a very detailed review

thallgren:
JIRA about the syntax error problem: https://tickets.puppetlabs.com/browse/PUP-6002

hlindberg:
I forgot one case, you must be allowed to do this:
```
$a = {type(3) => "a 3, the integer 3"}
```

hlindberg:
Note: cdb6eb6 requires a spec change (keywords allowed as keys/entries in arrays)

spec/unit/pops/types/enumeration_spec.rb
```diff
@@ -1,50 +1,51 @@
 require 'spec_helper'
 require 'puppet/pops'
 
+module Puppet::Pops::Types
```
hlindberg: This makes me nervous - I don't trust rspec/mocha well enough to be confident it does not leave stuff behind in the module that later may affect other tests.
I am probably being paranoid. Ping @kylog have we used this construct elsewhere?
@thallgren do we really have to do this?

thallgren: Yes, I REALLY want to do this. And it works as it supposed to. Why shouldn't it?

hlindberg:
Rubucop complains about something about optional parameters. Looks like it is completely wrong. Ping @thallgren

thallgren:
You need to refresh your page. The rubocop problem is already fixed and pushed.

spec/unit/pops/types/enumeration_spec.rb
hlindberg: Because all of the stuff rspec /mocha is doing - it now does that inside of the module under test; while that can be debated if it is right or wrong philosophically, I am nervous there is stuff left behind in the module with the possible negative effect on other tests of things in the very same module.

thallgren: Also, please note. I've been using this pattern for some time now. We currently have this pattern in at least six files on stable.

thallgren: If rspec/mocha is leaving things behind between tests, then it's probably no worse to leave it in a module than leave in the global scope. I've never seen anything like that happen though.

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+    # @return [PAnyType] the type of this member
+    # @api public
+    attr_reader :type
+
+    # @return [Hash{PType => Hash}] the annotations or `nil`
+    # @api public
+    attr_reader :annotations
+
+    # @param name [String] The name of the member
+    # @param [Hash{String=>Object}] i12n_hash
```
hlindberg: the name of the param is the wrong place (should be before the type), and does not have a description

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+    # the inherited member accepts to be overridden by this member.
+    #
+    # @param parent_members [Hash{String=>PAnnotatedMember}] the hash of inherited members
+    # @return [PAnnotatedMember] this instance
+    # @raises [Puppet::ParseError] if the assertion fails
+    # @api private
+    def assert_override(parent_members)
+      parent_member = parent_members[@name]
+      if parent_member.nil?
+        raise Puppet::ParseError, "expected #{label} to override an inherited #{feature_type}" if @override
```
hlindberg: half of the message seems to be missing - "no such found" or words to that effect

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+
+    # Checks if the given _member_ can override this member.
+    #
+    # @param member [PAnnotatedMember] the overriding member
+    # @return [PAnnotatedMember] its argument
+    # @raises [Puppet::ParseError] if the assertion fails
+    # @api private
+    def assert_can_be_overridden(member)
+      raise Puppet::ParseError, "#{member.label} overrides inherited #{feature_type}" unless self.class == member.class
+      raise Puppet::ParseError, "attempt to override final #{label}" if @final
```
hlindberg: would be great to be told where that attribute is made final

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+    end
+
+    # Checks if the given _member_ can override this member.
+    #
+    # @param member [PAnnotatedMember] the overriding member
+    # @return [PAnnotatedMember] its argument
+    # @raises [Puppet::ParseError] if the assertion fails
+    # @api private
+    def assert_can_be_overridden(member)
+      raise Puppet::ParseError, "#{member.label} overrides inherited #{feature_type}" unless self.class == member.class
```
hlindberg: the reason why that i wrong is missing in the message

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+      raise Puppet::ParseError, "attempt to override final #{label}" if @final
+      raise Puppet::ParseError, "#{label} attempts to override without having override defined" unless member.override?
```
hlindberg: would be great to be told where the attribute is defined that it is attempting to override

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+      raise Puppet::ParseError, "attempt to override #{label} with a type that does not match" unless @type.assignable?(member.type)
```
hlindberg: Would want to know where the objecting attribute is defined

lib/puppet/pops/types/p_object_type.rb
thallgren: The #{label} contains the name of the feature.

lib/puppet/pops/types/p_object_type.rb
thallgren: No it isn't actually. The output from this will be:
"attribute 'foo' overrides inherited function"
  or
"function 'foo' overrides inherited attribute"

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+    end
+
+    # @api public
+    def hash
+      @name.hash ^ @type.hash
+    end
+
+    # @api public
+    def eql?(o)
+      self.class == o.class && @name == o.name && @type == o.type && @override == o.override? && @final == o.final? && @annotations == o.annotations
```
hlindberg: I wonder if annotations should be included - my spontaneous reaction is that they should not. From the perspective of the object they do not really exist... (although they do technically as they need to serialized with the object, and adapters need to find their respective data).

lib/puppet/pops/types/p_object_type.rb
thallgren: Again `#{label}` contains the name of the feature.

lib/puppet/pops/types/p_object_type.rb
thallgren: The `#{label}` contains the name of the feature.

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+    # @return [String,nil] The attribute kind as defined by #TYPE_ATTRIBUTE_KIND, or `nil` to
+    #   indicate that
+    attr_reader :kind
+
+    # @return [Object] the default value or the symbol :default to denote that there is no default value
+    # @api public
+    attr_reader :value
+
+    # @param name [String] The name of the attribute
+    # @param [Hash{String=>Object}] i12n_hash
```
hlindberg: isn't type and parameter name in the wrong order here?

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+    end
+
+    # Returns the member as a hash suitable as an argument for constructor. Name is excluded
+    # @return [Hash{String=>Object}] the hash
+    # @api private
+    def i12n_hash
+      hash = super
+      unless @kind.nil?
+        hash[KEY_KIND] = @kind
+        hash.delete(KEY_FINAL) if @kind == ATTRIBUTE_KIND_CONSTANT # final is implied
```
hlindberg: a final => false would go undetected it seems?

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+      end
+      hash[KEY_VALUE] = @value unless @value == :default
+      hash
+    end
+
+    # @return [Boolean] `true` if a value has been defined for this attribute. The value may be `nil`
+    def value?
+      value != :default
```
hlindberg: Hm, this means it is  impossible to have a default value that is the literal default. That needs to be changed.

lib/puppet/pops/types/p_object_type.rb
thallgren: final => false is never added. See super implementation.

lib/puppet/pops/types/p_object_type.rb
thallgren: But this was one of the things we agreed on? That default wasn't a valid value. Instead the correct default for the type must be added (shorter anyway).

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+    end
+
+    @equality_include_type = i12n_hash[KEY_EQUALITY_INCLUDE_TYPE]
+    @equality_include_type = true if @equality_include_type.nil?
+
+    equality = i12n_hash[KEY_EQUALITY]
+    equality = [equality] if equality.is_a?(String)
+    if equality.is_a?(Array)
+      unless equality.empty?
+        raise Puppet::ParseError, 'equality_include_type = false cannot be combined with attributes' unless @equality_include_type
```
hlindberg: "non empty equality specification" instead of "attributes" (you can certainly have attributes, not just in the equality specification).

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+        equality.each do |attr_name|
+
+          attr = parent_members[attr_name]
+          if attr.nil?
+            attr = @attributes[attr_name] || @functions[attr_name]
+          elsif attr.is_a?(PAttribute)
+            # Assert that attribute is not already include by parent equality
+            parent_eq_attrs ||= parent_object_type.equality_attributes
+            if parent_eq_attrs.include?(attr_name)
+              raise Puppet::ParseError, "equality is referencing attribute '#{attr_name}' which is already included by parent equality"
```
hlindberg: in which parent?

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+  end
+
+  def eql?(o)
+    self.class == o.class &&
+      @parent == o.parent &&
+      @attributes == o.attributes &&
+      @functions == o.functions &&
+      @equality == o.equality &&
+      @checks == o.checks &&
+      @annotations == o.annotations
```
hlindberg: should annotations be included?

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+    @equality_include_type
+  end
+
+  # Returns the functions of this `Object` type. If _include_parent_ is `true`, then all
+  # inherited functions will be included in the returned `Hash`.
+  #
+  # @param include_parent [Boolean] `true` if inherited functions should be included
+  # @return [Hash{String=>PFunction}] a hash with the functions
+  # @api public
+  def functions(include_parent = false)
```
hlindberg: should there be something akin to attributes() and functions() that returns all features?

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -0,0 +1,539 @@
...
+      collector.merge!(@attributes.reject { |_, attr| attr.kind == ATTRIBUTE_KIND_CONSTANT })
+    else
+      collector.merge!(Hash[@equality.map { |attr_name| [attr_name, @attributes[attr_name]] }])
+    end
+    nil
+  end
+
+  # @api private
+  def label
+    'object'
```
hlindberg: freeze? (use constant perhaps)

lib/puppet/pops/types/p_object_type.rb
thallgren: There is a #members method that returns both attributes and functions (line 417). Mixing that with other features (equality, annotations, checks) is IMO pointless since they don't live in the same namespace.

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+  def parse_object(name, body_string)
+    expect_object(name, body_string)
+    parser.parse(name, loader)
+  end
+
+  context 'when dealing with attributes' do
+    it 'raises an error when the attribute type is not a type' do
+      obj = <<-OBJECT
+        attributes => {
+          a => 23
```
hlindberg: although, that could mean that it is a constant and type is inferred.

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+        }
+      OBJECT
+      expect { parse_object('MyObject', obj) }.to raise_error(TypeAssertionError,
+        /attribute 'a' had wrong type, expected a Type value, got Integer/)
```
hlindberg: the past tense throws me every time  when it is reporting this error 'a' still has the wrong type unless it was magically repaired

spec/unit/pops/types/p_object_type_spec.rb
thallgren: For constants it could work, but if we do that, then I'm not so the kind 'constant' should be implied because inferring could work for other kinds as well (the inferred type would need to be generalized).

Then again, yet another special rule to remember. So I'm ambivalent.

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+        attributes => {
+          a => { type => Integer[0,10], override => true }
+        }
+      OBJECT
+      expect_object('MyObject', parent)
+      tp = parse_object('MyDerivedObject', obj)
+      expect(tp['a'].type).to eql(PIntegerType.new(0,10))
+    end
+
+    it 'raises an error when the an attribute overrides a function' do
```
hlindberg: "the an" typo

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+        attributes => {
+          a => { type => String, override => true }
+        }
+      OBJECT
+      expect_object('MyObject', parent)
+      expect { parse_object('MyDerivedObject', obj) }.to raise_error(Puppet::ParseError,
+        "attempt to override attribute 'a' with a type that does not match")
+    end
+
+    it 'raises an error when the an attribute overrides a final attribute' do
```
hlindberg: "the an" typo

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+        attributes => {
+          a => { type => Integer, override => true }
+        }
+      OBJECT
+      expect_object('MyObject', parent)
+      expect { parse_object('MyDerivedObject', obj) }.to raise_error(Puppet::ParseError,
+        "attempt to override final attribute 'a'")
+    end
+
+    it 'raises an error when the an overriding attribute is not declared with override => true' do
```
hlindberg: "the an"

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+      OBJECT
+      obj = <<-OBJECT
+        parent => MyObject,
+        attributes => {
+          a => Integer
+        }
+      OBJECT
+      expect_object('MyObject', parent)
+      expect { parse_object('MyDerivedObject', obj) }.to raise_error(Puppet::ParseError,
+        "attribute 'a' attempts to override without having override defined")
```
hlindberg: actually, override must be true - message should say that, override => false is also a definition

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+    end
+
+    it 'raises an error when the an attribute declared with override => true does not override' do
```
hlindberg: "the an"

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+          b => Integer
+        },
+        equality => [a,b]
+      OBJECT
+      tp = parse_object('MyObject', obj)
+      expect(tp.equality).to eq(['a','b'])
+      expect(tp.equality_attributes.keys).to eq(['a','b'])
+    end
+
+    it 'a single attribute can be declared a name' do
```
hlindberg: typo, does not read right

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+          d => Integer
+        }
+      OBJECT
+      expect_object('MyObject', parent)
+      tp = parse_object('MyDerivedObject', obj)
+      expect(tp.equality).to be_nil
+      expect(tp.equality_attributes.keys).to eq(['a','c','d'])
+    end
+
+    it 'will allow that equality contains parent attributes when parent equality does not' do
```
hlindberg: "will allow" (when will it? doesn't it do it now?)  use "allows", or "parent defined attributes can be included in equality if not already included by a parent"

lib/puppet/pops/types/p_object_type.rb
thallgren: Sure, it could be frozen. But in general I try to freeze things where it matters (for performance or because immutability must be protected). This label is @api private and only used for errors.

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+        equality_include_type => false
+      OBJECT
+      expect { parse_object('MyObject', obj) }.to raise_error(Puppet::ParseError,
+        'equality_include_type = false cannot be combined with attributes')
+    end
+  end
+
+  it 'raises an error when initialization hash contains invalid keys' do
+    expect_object('MyObject', <<-OBJECT)
+      attribrutes => {
```
hlindberg: attribrute - a less refined form of attributes :-) love the pun.

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+      t = parse_object('MyDerivedObject', derived)
+      members = t.members.values
+      expect{ |b| members.each {|m| m.name.tap(&b) }}.to yield_successive_args('c', 'd')
+      expect{ |b| members.each {|m| m.type.simple_name.tap(&b) }}.to yield_successive_args('String', 'Boolean')
+      members = t.members(true).values
+      expect{ |b| members.each {|m| m.name.tap(&b) }}.to yield_successive_args('a', 'b', 'c', 'd')
+      expect{ |b| members.each {|m| m.type.simple_name.tap(&b) }}.to(yield_successive_args('Integer', 'Callable', 'String', 'Boolean'))
+    end
+
+    it 'will be assignable to its inherited type' do
```
hlindberg: rephrase "will be"

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+      p = expect_object('MyObject', parent)
+      t = parse_object('MyDerivedObject', derived)
+      expect(p).to be_assignable(t)
+    end
+
+    it 'will be assignable not consider inherited type to be assignable' do
```
hlindberg: "will be"

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+    context 'that in turn inherits another Object type' do
+      let(:derived2) { <<-OBJECT }
+        parent => MyDerivedObject,
+        attributes => {
+          e => String,
+          f => Boolean
+        }
+      OBJECT
+
+      it 'will be assignable to all inherited types' do
```
hlindberg: "will be"

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+        p = expect_object('MyObject', parent)
+        d1 = expect_object('MyDerivedObject', derived)
+        d2 = parse_object('MyDerivedObject2', derived2)
+        expect(p).to be_assignable(d2)
+        expect(d1).to be_assignable(d2)
+      end
+
+      it 'will not consider any of the inherited types to be assignable' do
```
hlindberg: "will not" => "does not"

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -0,0 +1,599 @@
...
+        functions => {
+          a => Callable[MyObject]
+        }
+      OBJECT
+      expect(t['a'].type).to eql(PCallableType.new(PTupleType.new([t])))
+    end
+  end
+
+  context 'when using the initialization hash' do
+    it 'produced hash use compact form attributes' do
```
hlindberg: rephrase - does not read right

lib/puppet/pops/types/p_object_type.rb
hlindberg: and the "name of the feature" includes the type where the feature is defined?

lib/puppet/pops/types/p_object_type.rb
hlindberg: yeah but where is the 'foo' defined that it clashes with?

lib/puppet/pops/types/p_object_type.rb
thallgren: The type? You mean the anonymous Object? It doesn't have a name.

lib/puppet/pops/types/p_object_type.rb
hlindberg: but... in order to inherit, you must specify the name of the parent type. Hence there must be a binding to a type name. So it can say attempt to override "foo" with a type that does not match "MyGreatGrandfather#foo" (or similar) given that this parent is Father, its parent is GrandFather etc...

lib/puppet/pops/types/p_object_type.rb
hlindberg: so, if I specify an attribute of constant type and explicitly say final => false - should I not get an error?

lib/puppet/pops/types/p_object_type.rb
hlindberg: no, not at all - we agreed that a literal default had no special meaning - whereas earlier it meant a default value of the type specified for the attribute. It is valid to have an attribute that accepts literal default and you may want it to default to that. You cannot use :default to signal "no value" here.

lib/puppet/pops/types/p_object_type.rb
hlindberg: ok. fine.

lib/puppet/pops/types/p_object_type.rb
thallgren: The problem is that all attributes and functions live in an Object. The Object is in itself anonymous. It may be aliased, and the parent may be such an alias, but there's no guarantee that it is. For some errors where you've had the same objection, it's not even certain that there is a parent.

We could change this of course, and not use a type alias as the mechanism to name an otherwise anonymous declaration. That way, all Objects would be guaranteed to have a name.

lib/puppet/pops/types/p_object_type.rb
hlindberg: ok, nevermind then

lib/puppet/pops/types/p_object_type.rb
hlindberg: Same here - where is the thing it is trying to override. I want to know if it was unexpected that a parent defined it

hlindberg:
(I deleted some resolved things, and added a thumbs up to those that require no action)
In general this looks fine.

I would like to see some puppet language level tests:

* operators == and != 
* operators =~ and !~ 
* that an instance of an object type matches an object type in a case option
* that attribute values in an object can be accessed, also for inherited attribtues (i.e. o['defined_here'] and o['defined_in_parent'])
* that an instance of object can be used as a hash key
* not sure if there was a complete object type creation and use in puppet language code (or if they were all partly done with ruby code + some puppet).

IIRC there were no tests added to general type testing (i.e. all the tests that deal with < > between types) -

lib/puppet/pops/types/p_object_type.rb
thallgren: There are two questions that might need an answer.

1. In which parent was the attribute defined?
2. In which parent was the attribute added to the equality definition?

Just looking where the actual attribute will not be enough since it might be added by a sub-type to the type where the attribute is declared. In order to figure this out, it will be necessary to traverse the parent chain and check which of the parents it is that actually adds the attribute. Is the added info worth that kind of complexity?

thallgren:
Rebased and pushed two new commits. One that improves how Object type is formatted, and one that addresses most of the issues that were pointed out in review. Some of the requested have to wait until the operator `new` has been implemented. There is currently no way of testing "instance of".

We need to settle the discussion on how to report where a feature is defined. My preference would be to keep things as they are since it's trivial to find the problem given the current messages together with a location in code.

lib/puppet/pops/types/p_object_type.rb
hlindberg: It is referring to its parent via name isn't it ? all the way up to Object.

lib/puppet/pops/types/p_object_type.rb
hlindberg: yes, especially for novices that sees that error - they need to figure out who to talk to - getting references to the conflicting points is very helpful.
In this case - I think it is enough to point to the type that already included the attribute - as the conflict is between that point, and the subtype where it is again added.

hlindberg:
I think it is of value to improve the error reporting ; to know both references when something is in conflict; type-name#attribute-name and file/line if not difficult to achieve. While an author using Object type surely will understand what is wrong from messages that do not have these references, other users that are just consumers have no clue regarding what may be wrong. They need to find both places - maybe there are more than two modules involved, maybe one of them is outdated. With the extra bit of information there is a greater chance they can solve issues, or quicker get in contact with someone who can fix the problem.

lib/puppet/pops/types/type_formatter.rb
```diff
@@ -295,7 +325,7 @@ def string_FalseClass(t)   ; 'false'       ; end
 
   # @api private
   def string_Hash(t)
-    append_hash('', Hash[t.map {|k,v| [string(k), string(v)]}])
+    append_hash('',Hash[t.map {|k,v| [string(k), string(v)]}])
```
hlindberg: huh - why remove that space?

thallgren: That is a very good question :-)

hlindberg:
https://tickets.puppetlabs.com/browse/PUP-6137 logged to cover naming and improvement of error messages (that need names).

hlindberg:
Ready for merge when travis goes green.

PR #4744, opened 2016-03-02
(maint) Skip updating bundler in AppVeyor

MikaelSmith:
AppVeyor instances now come with bundler pre-installed (see
https://www.appveyor.com/docs/installed-software). Remove the step
installing bundler.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,6 @@ init:
 install:
   - SET PATH=C:\Ruby21-x64\bin;%PATH%
   - SET LOG_SPEC_ORDER=true
-  - gem install bundler --quiet --no-ri --no-rdoc
   - bundle install --jobs 4 --retry 2 --without development extra
   - type Gemfile.lock
   - ruby -v


kylog:
:+1: pending a successful appveyor run.

appveyor.yml
```diff
@@ -5,7 +5,6 @@ init:
 install:
   - SET PATH=C:\Ruby21-x64\bin;%PATH%
   - SET LOG_SPEC_ORDER=true
-  - gem install bundler --quiet --no-ri --no-rdoc
```
ferventcoder: Yikes. Is this something that needs to be done other places?

MikaelSmith: Not sure what you're referring to.

ferventcoder: Just read the commit. A lot of times this was for the older versions of Ruby - to ensure that it was installed and updated.

ferventcoder: Puppet is probably fine to drop this line though. We may still need it in modules.

ferventcoder: maybe not.

MikaelSmith: If you're using AppVeyor's pre-installed Ruby versions (1.9.3-2.24) they all have Bundler 1.11.2 pre-installed.

ferventcoder: Sweet. :+1:

PR #4745, opened 2016-03-03
(PUP-5995) Fix source text extraction when preceded by multibyte character

hlindberg:
This fixes a regression (on some platforms, and real error on others) regarding source text extraction when extracted source is preceded by multi byte characters.

Before this, the encoding was the platform encoding, and if it was not UTF-8, some operations would operate on a per byte level. When this was changed to UTF-8 the extract_text and extract_tree_text method ended up using a char based slice where it should have used a byte based slice.

This resulted in the wrong text being extracted.

The fundamental problem was that the two methods in question thought that they could handle the slicing on their own. Now they delegate to corresponing methods in the Locator where the knowledge resides regarding the unit of "offset" (byte or char).

This also adds unit tests in the lexer and parser to assert that the positions returned by the lexer are correct. 
PR contains two maint fixes (see commits for details).


thallgren:
Looks OK but needs a rebase.

hlindberg:
rebased

PR #4747, opened 2016-03-03
(PUP-5099) Add code_id to environment catalog from request

nicklewis:
This follows the model of agent catalog compilation, passing code_id
from the HTTP request to the environment compiler, which adds it to the
catalog. We then read code_id from the environment catalog and include
it in our response.

This allows environment catalogs and agent catalogs to be correlated
with one another and against a specific version of code.

### DIFF
--- a/lib/puppet/network/http/api/master/v3/environment.rb
+++ b/lib/puppet/network/http/api/master/v3/environment.rb
@@ -5,14 +5,18 @@ class Puppet::Network::HTTP::API::Master::V3::Environment
   def call(request, response)
     env_name = request.routing_path.split('/').last
     env = Puppet.lookup(:environments).get(env_name)
+    code_id = request.params[:code_id]
 
     if env.nil?
       raise Puppet::Network::HTTP::Error::HTTPNotFoundError.new("#{env_name} is not a known environment", Puppet::Network::HTTP::Issues::RESOURCE_NOT_FOUND)
     end
 
-    catalog = Puppet::Parser::EnvironmentCompiler.compile(env).to_resource
+    catalog = Puppet::Parser::EnvironmentCompiler.compile(env, code_id).to_resource
 
-    env_graph = {:environment => env.name, :applications => {}}
+    # This reads code_id off the catalog rather than using the one from the
+    # request. There shouldn't really be a case where the two differ, but if
+    # they do, the one from the catalog itself is authoritative.
+    env_graph = {:environment => env.name, :applications => {}, :code_id => catalog.code_id}
     applications = catalog.resources.select do |res|
       type = res.resource_type
       type.is_a?(Puppet::Resource::Type) && type.application?
--- a/spec/unit/appmgmt_spec.rb
+++ b/spec/unit/appmgmt_spec.rb
@@ -11,10 +11,10 @@ describe "Application instantiation" do
   # no really elegant way to do this
   include ParserRspecHelper
 
-  def compile_to_env_catalog(string)
+  def compile_to_env_catalog(string, code_id=nil)
     Puppet[:code] = string
     env = Puppet::Node::Environment.create("test", ["/dev/null"])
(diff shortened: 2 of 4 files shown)


nicklewis:
This needs a puppet-server change to actually pass code-id through to the request handler, but this change can go in on its own.

briancain:
:+1: Looks good to me.

nicklewis:
@thallgren @hlindberg Any chance one of you could take a glance at this?

hlindberg:
@nicklewis reading...

lib/puppet/network/http/api/master/v3/environment.rb
```diff
@@ -5,14 +5,15 @@ class Puppet::Network::HTTP::API::Master::V3::Environment
...
 
     if env.nil?
       raise Puppet::Network::HTTP::Error::HTTPNotFoundError.new("#{env_name} is not a known environment", Puppet::Network::HTTP::Issues::RESOURCE_NOT_FOUND)
     end
 
-    catalog = Puppet::Parser::EnvironmentCompiler.compile(env).to_resource
+    catalog = Puppet::Parser::EnvironmentCompiler.compile(env, code_id).to_resource
 
-    env_graph = {:environment => env.name, :applications => {}}
+    env_graph = {:environment => env.name, :applications => {}, :code_id => catalog.code_id}
```
hlindberg: A comment above here perhaps that it gets `catalog.code_id` which may not be the same as the `code_id`. (to prevent someone messing this up by thinking `code_id` should be used...

spec/unit/appmgmt_spec.rb
```diff
@@ -574,6 +574,11 @@ class { cons: consume => Cap[cap]}
         }.to raise_error(/Only application components can appear inside a site - Notify\[fail me\] is not allowed at line 20/)
       end
     end
+
+    it "includes code_id if specified" do
+      catalog = compile_to_env_catalog(MANIFEST_WITH_SITE, "12345")
+      expect(catalog.code_id).to eq("12345")
+    end
   end
```
hlindberg: suggest adding  tests that check that the behaviour when a new code_id is returned in the catalog is the expected, and when code_id is nil.

nicklewis:
@hlindberg Updated with a comment explaining the `code_id` vs. `catalog.code_id` as well as improved tests around unspecified/differing code ids.

spec/unit/network/http/api/master/v3/environment_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  around :each do |example|
+    environment = Puppet::Node::Environment.create(:production, [], '/manifests')
+    loader = Puppet::Environments::Static.new(environment)
+    Puppet.override(:environments => loader) do
+      example.run
+    end
+  end
+
+  it "returns the environment catalog" do
+    request = Puppet::Network::HTTP::Request.from_hash(:headers => { 'accept' => 'application/json' }, :routing_path => "environment/production")
```
hlindberg: line a bit on the wide side... (picky, picky)

spec/unit/network/http/api/master/v3/environment_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    catalog = JSON.parse(response.body)
+    expect(catalog['environment']).to eq('production')
+    expect(catalog['applications']).to eq({})
+  end
+
+  it "returns 404 if the environment doesn't exist" do
+    request = Puppet::Network::HTTP::Request.from_hash(:routing_path => "environment/development")
+
+    expect { subject.call(request, response) }.to raise_error(Puppet::Network::HTTP::Error::HTTPNotFoundError, /development is not a known environment/)
```
hlindberg: picky: here as well

hlindberg:
Good to merge - only found a couple of wide lines (not that bad feel free to ignore).
Added tests are all good.

PR #4748, opened 2016-03-03
Link to Puppet Server-specific API endpoints.

gguillotte:
Puppet Server 2.3 adds its own `/puppet/v3/` endpoints for static file content and per-environment classes. Link to the Puppet Server documentation being written for these new endpoints from the Puppet API index, and add a supporting link for static catalogs to the `http_catalog` documentation.

Supports SERVER-1108 and SERVER-1114.


gguillotte:
... against the wrong branch.

PR #4750, opened 2016-03-03
(maint) Restore 2.3.0 testing in travis

kylog:
Now that https://bugs.ruby-lang.org/issues/11928 is fixed in the 2.3 ruby
branch (but not yet released), we can test against the 2.3.0-dev rvm version
in travis.

When 2.3.1 is released, we should bump to 2.3.1 as we don't really need to
track 2.3.0-dev.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ script:
 notifications:
   email: false
 rvm:
+  - 2.3.0-dev
   - 2.2.4
   - 2.1.7
   - 2.0.0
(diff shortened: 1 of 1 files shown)


kylog:
@MikaelSmith @HAIL9000 when we yanked 2.3.0 coverage a few weeks back, I think we talked about restoring it like this (until 2.3.1 is released).

MikaelSmith:
:+1:

PR #4752, opened 2016-03-03
(PUP-6000) Add only unique hooks during initialization

mateusz-gozdek-sociomantic:
This should avoid slowing down if we initialize settings multiple times, what, for example, `foreman-proxy` does.

### DIFF
--- a/lib/puppet/settings.rb
+++ b/lib/puppet/settings.rb
@@ -562,7 +562,7 @@ class Puppet::Settings
           # will have associated hooks that it ends up being less work this
           # way overall.
           if setting.call_hook_on_initialize?
-            @hooks_to_call_on_application_initialization << setting
+            @hooks_to_call_on_application_initialization |= [ setting ]
           else
             setting.handle(ChainedValues.new(
               preferred_run_mode,
@@ -908,7 +908,7 @@ class Puppet::Settings
         if tryconfig.call_hook_on_define?
           call << tryconfig
         elsif tryconfig.call_hook_on_initialize?
-          @hooks_to_call_on_application_initialization << tryconfig
+          @hooks_to_call_on_application_initialization |= [ tryconfig ]
         end
       end
 


hlindberg:
settting_spec.rb has expectations that are no longer met (that is why tests are failing)

mateusz-gozdek-sociomantic:
Unfortunately, I don't have experience with ruby unit tests. How about clearing `hooks_to_call_on_application_initialization` during `Puppet.settings.initialize_global_settings` and `Puppet.settings.initialize_app_defaults` calls? Can I push another commit to see if it's working?

branan:
I think this needs to use `|=` instead of `|`. At least one of the failing tests is related to a hook just not being called.

Once that's fixed there may still be other failures, but that should narrow the list somewhat

mateusz-gozdek-sociomantic:
Updated as @branan suggested.

branan:
Thanks @mateusz-gozdek-sociomantic. The tests look good now.

One more change before we can merge this: Would you mind updating your commit with a bit of a longer description of why you need the change? See the example commit message in our contributing documentation https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md#making-changes.

branan:
@mateusz-gozdek-sociomantic Whoops, there's actually one other thing. We need you to sign our CLA before we can accept your PR. https://cla.puppetlabs.com

Normally there's a bot that's supposed to tell you this when you open a PR, but it's apparently taking a nap

mateusz-gozdek-sociomantic:
Updated with better commit message. About CLA, I am waiting for my lead to decide which version should I sign. I will inform you as soon as I sign it.

mateusz-gozdek-sociomantic:
@branan Why my patch can't be used with `Trivial Patch Exemption Policy` ?

kylog:
@mateusz-gozdek-sociomantic the trivial exemption is only for changes that are both small number of lines *and* make no functional change. This does make a functional change, so is not exempt. Thanks for working through the CLA process and let us know if you have any more questions.

Thanks!

leandro-lucarella-sociomantic:
> @mateusz-gozdek-sociomantic the trivial exemption is only for changes that are both small number of lines and make no functional change. This does make a functional change, so is not exempt. Thanks for working through the CLA process and let us know if you have any more questions.

Just a quick question. From the [Trivial Patch Exemption Policy](https://docs.puppetlabs.com/community/trivial_patch_exemption.html):

> Puppet Labs Definition of Trivial: A trivial patch is a pull request that does not contain creative work. As a rule of thumb, changes are trivial patches if they:
>
> * are fewer than 10 lines, and
> * introduce no new functionality.

I don't think this PR is introducing any new functionality, so shouldn't that document be updated if this PR doesn't really falls into the Trivial Patch Exemption Policy? It seems a bit misleading.

hlindberg:
I agree with leandro-lucarella-sociomantic here. The reason we have the CLA is for IP rights purposes. If this PR is considered to have IP then no bug fix however small can be waived (they all alter function from unwanted to wanted), that only leaves typos in comments, docs, etc. Doubt that the intention of our policy is to be that strict. I read "introduce no new functionality" as non tech-speak for "introduce no new feature". Just my 2c. OTOH, signing the CLA does not hurt.

leandro-lucarella-sociomantic:
> OTOH, signing the CLA does not hurt.

I know, but we have to convince lawyers first, which might take time. I understand the CLA, if we have to end up signing it is fine (I think it will happen eventually), but I think the policy should be clear. (and in this particular case, I really think is overkill, is a 8 characters patch, not even words, just symbols, I even wonder if this could be under any copyright laws :) )

peterhuene:
My two cents (obviously IANAL): bug fixes are indeed functionality change (before the software did one thing and now it does something different) and triviality shouldn't be measured by how few characters changed because we all know that "it's only a one line change" is what one declares immediately before witnessing total protonic reversal.

Is this particular fix novel or causing patent trolls to salivate? Of course not, but the CLA is a necessity to protect the project and Puppet Labs from litigation.  I don't want to see us go down the road of viewing simple bug fixes as meeting the "trivial exception bar" on a case-by-case basis because, in all honesty, that's not something we are qualified to do as software developers.

I believe the exception's text should be amended to definitely state what is and is not exempt as this is not the first time the vague language has caused confusion on both sides.

kylog:
Also IANAL, but I agree the CLA text is insufficient guidance. In the past, we have generally treated the CLA as applying to 'functional changes' (which I think applies unambiguously to both bug fixes and new features) whereas 'new functionality' otoh is clearly ambiguous.

I also think that the line count language is a little misleading and could be tightened up. It's clearly leading to some perception that number of characters is a reasonable metric, though clearly a 1- or 2-character change could be hugely consequential.

I'll see if I can poke someone who *is* a lawyer to help on this one. Thanks for the great input @leandro-lucarella-sociomantic!

hlindberg:
If our policy is that strict, it should state that more clearly. Other projects require CLA for every contribution and have far more rigorous rules regarding IP.

kylog:
Hey all, I did check with our in-house lawyer and she's researching CLA and Trivial Patch language, so hang tight on this question. I'll update here as soon as we have updates to share. :)

kylog:
@mateusz-gozdek-sociomantic that comment from the clabot above only reflects the bot being turned on, not the answer from legal I was looking for. I'm continuing to try to get the language clarified (not just for this PR, btw - we need to tighten the language in general).

So hang tight, and apologies again for the delays.

PR #4753, opened 2016-03-03
(maint) Fix package_spec on Fedora 23

MikaelSmith:
The package_spec test was incorrect when run on Fedora 23, because the
default provider changed. Update the test to report `dnf` on Fedora 23.

### DIFF
--- a/spec/integration/type/package_spec.rb
+++ b/spec/integration/type/package_spec.rb
@@ -28,7 +28,11 @@ describe Puppet::Type.type(:package), "when choosing a default package provider"
         :yum
       end
     when 'Fedora'
-      :yum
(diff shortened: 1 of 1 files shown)


whopper:
Aaah this is the second time this has happened. I'll test this on F23.

spec/integration/type/package_spec.rb
```diff
@@ -28,7 +28,11 @@ def provider_name(os)
         :yum
       end
     when 'Fedora'
-      :yum
+      if Puppet::Util::Package.versioncmp(Facter.value(:operatingsystemmajrelease), '23') >= 0
```
whopper: This is true for both F22 and F23, so this test probably also fails on F22.

whopper:
:+1: passes on F21, F22 and F23

PR #4754, opened 2016-03-03
(PUP-5960) Document static catalog response

MikaelSmith:
Documents the response when requesting a static catalog from a Puppet
Server with static catalog support configured.


MikaelSmith:
ping @joshcooper

PR #4755, opened 2016-03-03
(PUP-2744) Always use RedHat service provider instead of init

whopper:
This commit ensures that for init-based services on RedHat
platforms, the RedHat service provider is used to properly
handle them rather than its more general parent the init provider.

### DIFF
--- a/lib/puppet/provider/service/init.rb
+++ b/lib/puppet/provider/service/init.rb
@@ -17,9 +17,11 @@ Puppet::Type.type(:service).provide :init, :parent => :base do
   end
 
   # Debian and Ubuntu should use the Debian provider.
+  # RedHat systems should use the RedHat provider.
   confine :true => begin
-     os = Facter.value(:operatingsystem).downcase
-     !(os == 'debian' || os == 'ubuntu')
+      os = Facter.value(:operatingsystem).downcase
+      family = Facter.value(:osfamily).downcase
+      !(os == 'debian' || os == 'ubuntu' || family == 'redhat')
   end
 
   # We can't confine this here, because the init path can be overridden.


whopper:
Note: this is a work in progress around what I believe is at least part of the solution for PUP-2744, which describes an issue where on newer `redhat` platforms, legacy init services are being managed by the init provider rather than the `redhat` provider.

IIRC, the actual problem under the hood is that Puppet sees both the `systemd` and `redhat` providers as system-wide defaults. `Systemd` is chosen first as the most specific, but for these legacy services Puppet discovers that it can't use this provider. We then have to fall back to another suitable provider, of which there are several in a list.

Although the `redhat` provider is also a 'default', because it wasn't chosen first it's simply added to the `suitable` list along with the `init` provider. Ordering is not guaranteed in this list, and `init` happens to be element `[0]`, and so is chosen.

Clearly, the real problem here is multiple-default ordering of providers. We've had discussions before about this, and the consensus has always been that allowing multiple defaults would be insane and defeat the purpose of a 'default'. That still leaves us in a bind, since it doesn't allow us to specify a priority ordering of the remaining suitable providers.

An easy way out of this in this particular scenario is to just disable the `init` provider for `redhat` systems, as the `redhat` provider is tailored for the peculiarities of init-based services in the redhats. This is exactly what we did for the Debians.

trevor-vaughan:
So, the issues with systemd vs initd are VERY real and not to be taken lightly. In particular, in some versions of RH7, auditd *must* be managed via init.d otherwise it fails. In theory, any discrepancy that is identified should be pushed upstream to Red Hat, Inc. but that doesn't really help in the short term.

Outside of that, there *could* be a list of known services with issues that are hard coded into the Red Hat type to switch the provider at the type level until something better can be done. This should be relatively easy even if it's an awful practice.

I would definitely vote for making systemd the default but I would like to see specific services that we know don't work dropped back to initd (there shouldn't be many).

In terms of your specific issue, you may be hitting something in systemd where the service isn't allowed to be enabled due to masking or something like that. I haven't investigated this to any level.

pyther:
@trevor-vaughan Thanks! This is an interesting problem.

Are you aware that `systemctl reload auditd` will issue a SIGHUP to the auditd which will caused auditd to reconfigure. Are there situations that you are aware of where reconfiguring auditd is not sufficient?

Perhaps a `exec { 'auditd-restart':  command => 'service restart auditd' }` would be appropriate in such cases?

trevor-vaughan:
@pyther I am indeed aware of both of these solutions. However, this does not address the issue at hand which is that there is one core, and may be other, services which natively do not support systemd as expected.

Now, I'm certainly aware that we can simply say "do X for auditd" and go on our way but that's not exactly a seamless solution for the omnipotent service resource.

And, yes, I can change the `restart` command in the service resource to say `systemctl reload auditd`. But again, in theory, the service resource should "just work" since it has a provider that is handling Red Hat explicitly and differentiating between systemd and non-systemd systems.

I guess my argument is; if we know about an edge case, why not handle it?

pyther:
@trevor-vaughan I don't necessarily disagree with you. I think there are two situations at play here.

The first situation - auditd is a systemd service (/usr/lib/systemd/system/auditd.service). It does not have a init script, but it does have some initscript helpers/actions (/usr/libexec/initscripts/legacy-actions/auditd) which are used by the `service` command.

The most ideal situation here is to
- manage enable/disable state with systemd
- manage ensure state (running/stopped) with the `service` command
- manage the restart/reload action using the `service` command

The second situation - there are services files that have an init script, but do not have systemd unit file. In rhel, systemd will correctly detect the init scripts and issue the appropriate start/stop/reload/enable/disable commands. Note, in debian systemctl enable/disable does work with init scripts. Logic has been added to fallback to their init tools. https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/service/systemd.rb

Example:
```
$ systemctl enable netconsole
netconsole.service is not a native service, redirecting to /sbin/chkconfig.
Executing /sbin/chkconfig netconsole on
$ systemctl disable netconsole
netconsole.service is not a native service, redirecting to /sbin/chkconfig.
Executing /sbin/chkconfig netconsole off
$ systemctl start netconsole
Job for netconsole.service failed because the control process exited with error code. See "systemctl status netconsole.service" and "journalctl -xe" for details.
$ systemctl status netconsole
● netconsole.service - SYSV: Initializes network console logging
   Loaded: loaded (/etc/rc.d/init.d/netconsole)
   Active: failed (Result: exit-code) since Thu 2016-03-10 16:15:48 EST; 45s ago
     Docs: man:systemd-sysv-generator(8)
  Process: 79110 ExecStart=/etc/rc.d/init.d/netconsole start (code=exited, status=6)
```
Note: netconsole failed to start due system configuration.

Based on this, I don't think anything special needs to be done for legacy init scripts on rhel. 

Are there other packages like auditd that ship and use a systemd unit file, but rely on legacy init functionality? I suspect not, but I don't know. If this can be implemented elegantly that would be great, but I would hate to see over complicated code to handle this one situation - for what is essentially an operating system bug.

I'll see if I can parse some RHEL/Centos repo metadata and figure out which packages install files in /usr/libexec/initscripts/legacy-actions/. This should help better answer my question.

pyther:
I did a quick search with yum and it looks like iptables and auditd are the only packages that provide files in /usr/libexec/initscripts. Iptables only has hooks for 'save' and 'panic'.
```
[root@gyurgyik-desktop ~]# yum whatprovides /usr/libexec/initscripts/legacy-actions/*/* | grep 'x86_64 : ' | sort | uniq
audit-2.4.1-5.el7.x86_64 : User space tools for 2.6 kernel auditing
iptables-services-1.4.21-16.el7.x86_64 : iptables and ip6tables services for
```

lib/puppet/provider/service/init.rb
```diff
@@ -18,8 +18,9 @@ def self.defpath
 
   # Debian and Ubuntu should use the Debian provider.
```
rnelson0: Probably should add a comment about RedHat's provider as well.

alexharv074:
Any update here?

MikaelSmith:
Based on testing for PUP-6140, I'm :+1: on this.

PR #4757, opened 2016-03-04
(maint) Update test to ensure systemd is the default in F22 and F23

whopper:


whopper:
There isn't really any need to get this in before 4.4.0, The systemd provider has `defaultfor :operatingsystem => :fedora` which covers all releases, so this update is really just for consistency.

PR #4758, opened 2016-03-06
(PUP-5902) Output type alias mismatches with aliased type

thallgren:
This PR modifies the type mismatch describer so that it will output type aliases together with the type that they are aliasing. The change was split into several commits.

The details are in the commit comments.


hlindberg:
The refactored name can be done in one commit on top of the others. I like the rest.

PR #4759, opened 2016-03-07
(PUP-6017) Use beaker node_name in static cat test

johnduarte:
This commit updates the acceptance
`direct_puppet/static_catalog_env_control` test to use the beaker
`node_name` value when referencing value for puppet catalog
validation.

Prior to this change, the test would fail if the `hostname` was
not the same as the `node_name`.

### DIFF
--- a/acceptance/tests/direct_puppet/static_catalog_env_control.rb
+++ b/acceptance/tests/direct_puppet/static_catalog_env_control.rb
@@ -14,7 +14,7 @@ require 'json'
 @agent_manifests = {}
 @catalog_files = {}
 agents.each do |agent|
-  hn = agent.hostname
+  hn = agent.node_name
   resdir = agent.tmpdir('results')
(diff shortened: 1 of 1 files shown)


joshcooper:
Do we run acceptance with Solaris master? If so we may need to update some other acceptance tests...

johnduarte:
This has been manually validated on the following platforms
  * debian-8-amd64
  * solaris-11-x86
  * windows server 2012r2 64-bit
  * solaris-10-sparc

johnduarte:
@joshcooper no, we do not acceptance tests with Solaris master.

joshcooper:
Passed for me in:

```
$ bundle exec rake ci:test:aio SHA=7282f093e3c77b3382938a1c2e22d12b59b7973e SUITE_VERSION=1.3.5.353.g7282f09 TESTS=tests/direct_puppet/static_catalog_env_control.rb
```

PR #4760, opened 2016-03-08
(PUP-6020) Use eql? and proper method override for == for Puppet Types

thallgren:
The Puppet type system used a `==` method in the topmost `PAnyType` and
then other types added overrides for that method. The topmost class
also declared `eql?` as an alias for `==`. This didn't work very well
because an alias is hard wired to the method in the class where it is
declared. It does not recognize that the method is overridden.

This commit changes all `==` methods in the `PAnyType` hierarchy to
`eql?` methods and adds a proper delegation method in the topmost
superclass.

### DIFF
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -171,11 +171,13 @@ class PAnyType < TypedModelObject
     true
   end
 
-  def ==(o)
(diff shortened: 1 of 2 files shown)


hlindberg:
This can be merged as soon as appveyor is green.

PR #4761, opened 2016-03-08
(PUP-5901) Make type mismatch describer use normalized types.

thallgren:
This PR contains new functionality that enables puppet types to be normalized and then what's needed to make the TypeMismatchDescriber take advantage of that.

The details are in the commit messages.


hlindberg:
Looks good. Writing tests with example titles really help - thanks for that. 
I also like the output that this change achieves - so much better than before.

PR #4763, opened 2016-03-08
(PUP-5877) acceptance: puppet class inheritance

er0ck:
This change adds a test to ensure we are inheriting classes properly and
from the correct namespace. We've regressed a couple times here.

no need to merge this just yet. Wait until acceptance is unfrozen after upcoming release(s)

[skip ci]

### DIFF
--- a/acceptance/tests/language/class_inheritance.rb
+++ b/acceptance/tests/language/class_inheritance.rb
@@ -1,0 +1,18 @@
+test_name 'C14943: Class inheritance works correctly' do
+
+  agents.each do |agent|
+    test_manifest = <<MANIFEST
+      class bar { notice("This is class bar") }
+      class foo::bar { notice("This is class foo::bar") }
+      class foo inherits bar { notice("This is class foo") }
+
+      include foo
+MANIFEST
+
+    results = apply_manifest_on(agent, test_manifest, :accept_any_exit_code => true)
+    assert_match(/Scope\(Class\[Bar\]\): This is class bar/, results.stdout, 'did not find class bar')
+    assert_match(/Scope\(Class\[Foo\]\): This is class foo/, results.stdout, 'did not find class foo')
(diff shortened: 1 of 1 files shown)


acceptance/tests/language/class_inheritance.rb
```diff
@@ -0,0 +1,19 @@
...
+      class foo::bar { notice("This is class $name") }
+      class foo inherits bar { notice("This is class $name") }
+
+      include foo
+MANIFEST
+
+    apply_manifest_on agent, test_manifest do
+      assert_match(/Scope\(Class\[Bar\]\): This is class bar/, stdout, 'did not find class bar')
+      assert_match(/Scope\(Class\[Foo\]\): This is class foo/, stdout, 'did not find class foo')
+      refute_match(/This is class foo::bar/, stdout, 'should not have found class foo::bar')
```
joshcooper: Will these matchers still be evaluated if applying the manifest fails and returns non-zero exit code?

acceptance/tests/language/class_inheritance.rb
```diff
@@ -0,0 +1,19 @@
+test_name 'C14943: Class inheritance works correctly' do
+
+  agents.each do |agent|
+    test_manifest = <<MANIFEST
+      class bar { notice("This is class $name") }
+      class foo::bar { notice("This is class $name") }
+      class foo inherits bar { notice("This is class $name") }
+
```
hlindberg: You are relying on $name being correctly set - here I think it would be better with hardcoded names since the purpose is not to test that $name is set to the classname. I am being pedantic, feel free to ignore.

er0ck:
updated based upon review comments.  thanks!

er0ck:
@hlindberg @thallgren @peterhuene this should be ready to go, when you have a chance

peterhuene:
:+1:

PR #4765, opened 2016-03-09
(maint) Return PuppetSpec::File paths as UTF-8

Iristyle:
- Re-encode string returned by PuppetSpec::Files.tmpfile and
   PuppetSpec::Files.tmpdir so that they can be concatenated with other
   UTF-8 strings used in tests.

   While Ruby internally will always return these strings intially in
   the current local codepage on Windows (which may be fixed in later
   Ruby versions), this primarily impacts testing scenarios where it is
   useful to create temporary files that have UTF-8 names.

### DIFF
--- a/spec/lib/puppet_spec/files.rb
+++ b/spec/lib/puppet_spec/files.rb
@@ -28,7 +28,7 @@ module PuppetSpec::Files
   def self.tmpfile(name, dir = nil)
     # Generate a temporary file, just for the name...
     source = dir ? Tempfile.new(name, dir) : Tempfile.new(name)
-    path = source.path
+    path = source.path.encode(Encoding::UTF_8)
     source.close!
 
     record_tmp(File.expand_path(path))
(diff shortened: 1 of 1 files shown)


Iristyle:
I haven't fully vetted this yet and it will probably cause some failures, but kicking off Travis / AppVeyor to see what they might be cross-platform.

/cc @puppetlabs/windows

Iristyle:
Looks like there aren't any failures, so I think this is safe to merge.

glennsarti:
As this is only affect spec tests, this is low impact change.  Merging.

Iristyle:
Probably should have noted in the commit message, but the reason for this falls back on Ruby:

* {Dir.mktmpdir} calls back to https://github.com/ruby/ruby/blob/v2_1_8/lib/tmpdir.rb#L25 - `ENV['TMP']`, etc will return strings in the local codepage due to buggy Ruby behavior - addressed in https://bugs.ruby-lang.org/issues/8822 / https://bugs.ruby-lang.org/issues/9715
* {{Tempfile.new}} folllows a similar code path, calling in at https://github.com/ruby/ruby/blob/v2_1_8/lib/tempfile.rb#L136 to {{Tmpname.create}} which effectively does the same thing at https://github.com/ruby/ruby/blob/v2_1_8/lib/tmpdir.rb#L125

PR #4766, opened 2016-03-09
(PUP-6024) Restrict sources of inlined file metadata

joshcooper:
Previously, we would inline file metadata for any file in the
per-environment directory, however, puppetserver would only serve file
content for files within the per-environment modules/<module>/files directory.
As a result, we were creating a catalog whose static file content could
not be retrieved in some situations, notably when using a control repo
with a source like:

    puppet:///modules/profile/puppetmaster/update-classes.sh

which uses a custom modulepath in environment.conf to serve the
file from:

    site/profile/files/puppetmaster/update-classes.sh

Note the module 'name' in this case is 'profile'.

This commit ensure we only inline file metadata when using the 'modules'
mount and the resolved path is in a per-environment directory with a
relative path of the form: <base>/<module>/files/

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -4,6 +4,7 @@ require 'puppet/indirector/code'
 require 'puppet/util/profiler'
 require 'puppet/util/checksums'
 require 'yaml'
+require 'uri'
 
 class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
   desc "Compiles catalogs on demand using Puppet's compiler."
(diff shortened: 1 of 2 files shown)


lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -123,6 +124,17 @@ def inlineable?(resource, sources)
     end
   end
 
+  # Return true if metadata is inlineable, meaning the request's source is
+  # for the 'modules' mount and the resolved path is of the form:
+  #   $codedir/environments/$environment/*/*/files/**
+  def inlineable_metadata?(metadata, source, environment_path)
+    source_as_uri = URI(URI.escape(source))
+    location = Puppet::Module::FILETYPES['files']
+
```
MikaelSmith: I don't see why the !! Is necessary. Otherwise looks good.

joshcooper: It's not strictly necessary, but since it's a predicate method, it's ruby convention to double negate to ensure a boolean value is returned. A truthy or falsey value would also work, but it leaks internal state.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -123,6 +124,17 @@ def inlineable?(resource, sources)
...
+    !!(source_as_uri.path =~ /^\/modules\// &&
+       metadata.full_path =~ /#{environment_path}[^\/]+\/[^\/]+\/#{location}\/.+/)
```
joshcooper: I think this is roughly the same as: https://github.com/puppetlabs/puppet-server/pull/952/files#diff-bf54139d4adcaf5196b063fb102a545fR355. One possible difference is I'm requiring a non-empty string after "files/".

lib/puppet/indirector/catalog/compiler.rb
MikaelSmith: Figured the `&&` would ensure a Boolean.

joshcooper: Unfortunately, ruby is full of surprises:

```ruby
irb(main):001:0> Time.new && Time.new
=> 2016-03-08 23:51:56 -0800
```

KevinCorcoran: ruby

PR #4768, opened 2016-03-09
(PUP-5903) Add local type aliases to 4.x function API

thallgren:
This PR adds a new feature that makes it possible to define local type aliases in a 4.x Ruby function. These type aliases are only available inside the function.

Aliases are defined inside a block given to `local_types` by calling `type` with a puppet source string on the form `AliasT = T`.
```
Puppet::Functions.create_function(:example) do
  local_types do
    type 'MyType = Array[Integer]'
    type ...
  end
  dispatch :example do
    param 'MyType', :x
  end
end
```

This PR supersedes [PR-4732](/puppetlabs/puppet/pull/4732)

### DIFF
--- a/lib/puppet/functions.rb
+++ b/lib/puppet/functions.rb
@@ -290,6 +290,46 @@ module Puppet::Functions
         dispatch(meth_name, &block)
       end
     end
+
+    # Allows types local to the function to be defined to ease the use of complex types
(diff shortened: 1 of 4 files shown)


hlindberg:
+1 on the merge btw - we worked on this together

PR #4769, opened 2016-03-09
(PUP-6031) Preserve source host and port during inlining

joshcooper:
If the puppet master's host and port are specified in the `source`
parameter, then we were not preserving the host and port when inlining
content_uri in the file metadata. As a result, the agent would either
use a host:port returned from DNS SRV records (if enabled) or the
Puppet[:server] and Puppet[:masterport] settings, which could be
different from what the manifest originally specified.

The code tried to do this, but `CGI.escape` should only be used on query
parameters, not the entire URI. And we were missing a test to verify the
behavior.

Also modifies another call site to use URI.parse(uri) instead of
URI(uri). The former should be called when parsing a String. The latter
should be called when you have either a String or a URI.


MikaelSmith:
:+1:

PR #3141, opened 2014-10-01
PUP-1253 Add systemd service masking support

peiriannydd:
This updated the systemd service type to add support for masking. If a
service is masked, it is deemed to also be disabled. If a service is
masked and changed to enabled, it will first be unmasked since the
standard 'systemctl enable' command does not properly unmask the command
first.

PR #4770, opened 2016-03-09
(PUP-5825) use systemctl is-enabled to determine service enable state

pyther:
Submitting this pull request to start a discussion. I have not tested this on Debian and have only spot checked this on RHEL 7 (`puppet apply -e "service { 'autofs': ensure => true }'"`).

`systemctl is-enabled` will print the following statuses to stdout:
- enabled / enabled-runtime
- linked / linked-runtime
- masked / masked-runtime
- static
- indirect
see the systemctl(1) man page for description of each

No output is generated if systemctl can't find the unit file.

The networking service is a good example of a service that has an init script but does not have a unit file. I have tested `systemctl is-enabled networking` on Debian 8.0, 8.3, Ubuntu 15.04, and Ubuntu 15.10. The output is the same on all systems.

```
$ systemctl is-enabled networking 2>/dev/null
$ systemctl is-enabled networking 1>/dev/null
Failed to get unit file state for networking.service: No such file or directory
$ systemctl show networking --property LoadState --property UnitFileState
LoadState=loaded
UnitFileState=
$
```

Note, is-enabled was being used but was removed when masked support was merged in. Commit d374de80db52acf843ae5e18f193fc6475cdfa8d.

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -50,37 +50,21 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
   end
 
   def enabled?
-    begin
-      systemctl_info = systemctl(
-         'show',
-         @resource[:name],
-         '--property', 'LoadState',
-         '--property', 'UnitFileState',
-         '--no-pager'
-      )
+    cmd = [command(:systemctl), 'is-enabled', @resource[:name]]
+    output = execute(cmd, :failonfail => false).strip
 
-      svc_info = Hash.new
-      systemctl_info.split.each do |svc|
-        entry_pair = svc.split('=')
-        svc_info[entry_pair.first.to_sym] = entry_pair.last
-      end
-
-      # The masked state is equivalent to the disabled state in terms of
-      # comparison so we only care to check if it is masked if we want to keep
-      # it masked.
-      #
-      # We only return :mask if we're trying to mask the service. This prevents
-      # flapping when simply trying to disable a masked service.
-      return :mask if (@resource[:enable] == :mask) && (svc_info[:LoadState] == 'masked')
-      return :true if svc_info[:UnitFileState] == 'enabled'
-      if Facter.value(:osfamily).downcase == 'debian'
-        ret = debian_enabled?(svc_info)
-        return ret if ret
-      end
-    rescue Puppet::ExecutionFailure
-      # The execution of the systemd command can fail for quite a few reasons.
-      # In all of these cases, the failure of the query indicates that the
-      # service is disabled and therefore we simply return :false.
+    # The masked state is equivalent to the disabled state in terms of
+    # comparison so we only care to check if it is masked if we want to keep
+    # it masked.
+    #
+    # We only return :mask if we're trying to mask the service. This prevents
+    # flapping when simply trying to disable a masked service.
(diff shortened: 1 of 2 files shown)


kylog:
In case anyone else goes looking, the commit above landed in #3141. Much conversation in that PR, not sure if that's a help here or not. Just compiling prior art :)

lib/puppet/provider/service/systemd.rb
```diff
@@ -50,37 +50,21 @@ def get_start_link_count
...
+    # The masked state is equivalent to the disabled state in terms of
+    # comparison so we only care to check if it is masked if we want to keep
+    # it masked.
+    #
+    # We only return :mask if we're trying to mask the service. This prevents
+    # flapping when simply trying to disable a masked service.
+    return :mask if (@resource[:enable] == :mask) && (output == 'masked')
+    return :true if output == 'enabled'
+    return :false if ['disabled', 'linked', 'static', 'indirect', 'masked'].include? output
+    if (output.empty?) && (Facter.value(:osfamily).downcase == 'debian')
```
peterhuene: Previously the debian fallback was always executed, but now we check to see if the output was empty.  Could there be a case where the `is-enabled` command returns output we don't understand (i.e. not masked, enabled, etc.) on Debian the fallback doesn't get executed?

peterhuene:
@pyther Thanks for the contribution!  Overall the change makes sense, but we have a few comments and the need to test this out on a few platforms before merging.

Ping @whopper.  Would it be possible to verify this change on Debian and a few other platforms?  Haus says you're the systemd-guy now :)

lib/puppet/provider/service/systemd.rb
pyther: That situation is certainly possible. If `is-enabled From my point of view, the debian fallback should only be executed when

pyther:
@peterhuene Is there anything else that I can do to help move this along?

whopper:
@peterhuene @pyther I'll have a look at this a bit later today, and do some extra testing in Debian!

whopper:
As far as I can tell, everything looks good. I did some manual testing in various Debian 8 scenarios (and Ubuntu 15.04), including the default systemd/init setup *and* a system with systemd disabled. In both cases, services worked in the same way as they did prior to this change. That is to say, they work as expected.

`networking` and `apache2` are both examples of services without unit files, and both fall back to the `debian_enabled?` method, as expected.

The same is also true for RedHat 7.

So, since this looks to work in non-chroot environments, 👍  from me. (I don't currently have a chroot to test this in.)

lib/puppet/provider/service/systemd.rb
MikaelSmith: This doesn't handle `-runtime` states (enabled-runtime, linked-runtime, masked-runtime), but the previous implementation didn't either. If `systemctl show` returned `UnitFileState=enabled-runtime`, that would be treated as disabled.

pyther: @MikaelSmith `systemctl is-enabled' should return `enabled-runtime`, `linked-runtime`, or `masked-runtime` per the [https://www.freedesktop.org/software/systemd/man/systemctl.html#is-enabled%20NAME...](systemctl man page).

This was something I noticed when developing this patch. I didn't worry about it because as you noted the previous implementation didn't. I forgot to mention this when I submitted the PR.

PR #4772, opened 2016-03-09
(PUP-5233) Restore `puppet inspect`

whopper:
Fix `puppet inspect` so it looks in the correct location for the cached
catalog.


whopper:
Added a second new tests from Michael's PR https://github.com/puppetlabs/puppet/pull/4717, which this now supersedes.

MikaelSmith:
:+1:

branan:
:+1: from me as wel. Looks ready to go once we have a safe stable

whopper:
Actually, I just did a little digging around inspect and found that it really is horribly broken beyond just not being able to find the cached catalog.

It isn't actually able to recognize changes to resources on the agent, and so `inspect` reports will not identify changes, completely defeating the purpose of the audit.

PR #4773, opened 2016-03-09
(PUP-6032) Remove ignorecache from documentation strings

nfagerlund:
This commit only removes references to ignorecache and implications that it does
anything; it does not remove the setting itself.


kylog:
#4774 now includes this. Woot!

PR #4774, opened 2016-03-10
(PUP-6032) Deprecate the ancient ignorecache setting

kylog:


kylog:
New and improved! Now includes the goodness that is #4773.

nfagerlund:
:+1: yaaay

PR #4775, opened 2016-03-10
(PUP-6036) Escape path when validating content_uri

joshcooper:
The content_uri metadata is composed from the source scheme, host, and
port, and the path of the file on disk, relative to the per-environment
directory. Since the path can contain characters that must be
percent-encoded, e.g. 'foo%bar', we need to escape the path prior to
trying to parse the URI.

### DIFF
--- a/lib/puppet/file_serving/metadata.rb
+++ b/lib/puppet/file_serving/metadata.rb
@@ -31,7 +31,7 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
 
   def content_uri=(path)
     begin
-      uri = URI.parse(path)
+      uri = URI.parse(URI.escape(path))
     rescue URI::InvalidURIError => detail
       raise(ArgumentError, "Could not understand URI #{path}: #{detail}")
     end
--- a/spec/unit/file_serving/metadata_spec.rb
+++ b/spec/unit/file_serving/metadata_spec.rb
@@ -98,7 +98,13 @@ describe Puppet::FileServing::Metadata do
 
     describe "when assigning a content_uri" do
       it "should fail if uri is invalid" do
-        expect { metadata.content_uri = 'some foo' }.to raise_error ArgumentError, /Could not understand URI some foo: /
+        expect { metadata.content_uri = '://' }.to raise_error ArgumentError, /Could not understand URI :\/\//
+      end
+
+      it "should accept characters that require percent-encoding" do
+        uri = 'puppet:///modules/foo/files/ %:?#[]@!$&\'()*+,;='
+        metadata.content_uri = uri
+        expect(metadata.content_uri).to eq(uri)
       end
 
       it "should fail if uri is opaque" do


MikaelSmith:
Not everything in your test string works. Using
```
file { '/tmp/ %:?#[]@!$&\'()*+,;=':
  ensure => file,
  source => 'puppet:///modules/moda/ %:?#[]@!$&\'()*+,;=',
}
```
I get
```
<- "GET /puppet/v3/file_metadata/modules/moda/%20%25:?environment=production&links=manage&checksum_type=md5&source_permissions=ignore HTTP/1.1\r\nAccept: pson, yaml, binary\r\nX-Puppet-Version: 4.4.0\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Ruby\r\nHost: Michaels-MacBook-Pro.local:8140\r\n\r\n"
-> "HTTP/1.1 404 Not Found\r\n"
-> "Date: Thu, 10 Mar 2016 04:04:09 GMT\r\n"
-> "Content-Type: text/plain; charset=ISO-8859-1\r\n"
-> "X-Puppet-Version: 4.4.0\r\n"
-> "Content-Length: 56\r\n"
-> "Server: Jetty(9.2.10.v20150310)\r\n"
-> "\r\n"
reading 56 bytes...
-> "Not Found: Could not find file_metadata modules/moda/ %:"
read 56 bytes
Conn keep-alive
Error: /Stage[main]/Main/File[/tmp/ %:?#[]@!$&'()*+,;=]: Could not evaluate: Could not retrieve information from environment production source(s) puppet:///modules/moda/ %:?#[]@!$&'()*+,;=
```

That's without using a static catalog. That's not a reason to avoid this fix though.

MikaelSmith:
Had to modify my `code_content.sh` to use `  cat "/Users/michaelsmith/puppetlabs/puppet-server/target/master-code-dir/environments/$1/${*:3}"`. With that it appears to retrieve content from the server correctly.

joshcooper:
The first problem is `?`. It's a [reserved character](https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters), but because `URI.escape` doesn't percent-encode it, `URI.parse` thinks `?` is the delimiter between the path and query params as can be seen from: `GET /puppet/v3/file_metadata/modules/moda/%20%25:?environment...`.

The next problem is `#`, which is also a reserved character, since it's an anchor.

If I remove those, then `[` and `]` cause problems:

```
irb(main):005:0> URI.parse("puppet:///modules/ntp/[").path
URI::InvalidURIError: bad URI(is not URI?): puppet:///modules/ntp/[
...
irb(main):006:0> URI.parse("puppet:///modules/ntp/]").path
URI::InvalidURIError: bad URI(is not URI?): puppet:///modules/ntp/]
...
```

And if I remove those, then I can serve files through webrick:

```
Notice: /Stage[main]/Ntp/File[/tmp/ %:@!$&'()*+,;=]/ensure: defined content ...
```

One thing I don't understand is why other reserved characters like `(`, `&`, don't cause problems when calling `URI.parse`?

I'm thinking next steps are to file a bug to ensure those special characters are escaped (when using static catalogs or not). We should also sync with the server team to see which characters jetty assumes are percent-encoded.

PR #4777, opened 2016-03-10
(DOCUMENT-515) Clarify systemd unit type.

binford2k:


kylog:
Replacing with #4791, which is the same commit with the summary line amended. I'll merge that once travis is happy over there.

PR #4778, opened 2016-03-10

kylog:


joshcooper:
Will we have to redo this as files are merged in from stable? If so, we may want to target stable?

kylog:
Excellent point. I've gone one step further and targeted 3.x in #4793. If we merge that up to stable I'll take another pass at stable after for any tests there, and create another PR if needed. LIkewise master after that.

PR #4688, opened 2016-02-19
(PUP-4545): Too many restarts on interface changes

danzilio:
This commit removes `ext/puppet-nm-dispatcher` and cleans up the RPM
spec accordingly.

PR #4780, opened 2016-03-11
(PUP-4545): Too many restarts on interface changes

danzilio:
This commit removes `ext/puppet-nm-dispatcher` and cleans up the RPM spec accordingly. Rebase of #4688


kylog:
Thanks @danzilio! This slipped off my radar.

In an oob discussion, @shermdog pointed out that this would mean a regression on https://projects.puppetlabs.com/issues/2776.

It seems like in our more, modern, container-y world PUP-4545 is more of an active pain point than the original redmine issue, so this *might* be an okay tradeoff. Otoh, perhaps we should consider one of the other suggestions in the ticket, e.g. either a regex blacklist or monitoring resolv.conf.

The regex blacklist seems like it could work but might be fidgety / high-maint (or maybe not, I don't know enough in this space). The monitoring resolv.conf sounds more like the ideal but also might be more work than anyone has time for atm.

Thoughts?

danzilio:
I think monitoring `resolv.conf` is the least duct tape-y solution. I can think of at least a dozen ways to do this, but I'm not sure any of them are ideal.

What about a configuration option in `puppet.conf`?

danzilio:
For comparison, Chef seems to have kicked the can on this in chef/chef#2894

kylog:
In net, this seems like the lesser of two evils: the current behavior of 3.x fixes a narrow-ish race condition at the expense of many puppet restarts, which is effecting a number of users.

Also, worth noting that puppet 4 packaging does *not* include this script, so this would bring 3.8.x in sync with puppet 4.

Merging, and we'll add a release note with instructions on how to address the resolv.conf bug if someone runs into that.

PR #4782, opened 2016-03-11
(PUP-5967) Remove reference to nonexistent links value in file type

whopper:
Previously, the description of the `links` parameter in the file type
included misleading references to a nonexistent `ignore` value. This
commit removes those references.


MikaelSmith:
Let's target this at stable too.

PR #4783, opened 2016-03-11
(maint) Unwrap arrayification of config_version from execute call

haus:
Previously, the config_version was wrapped in an array for the
Puppet::Util::Execution.execute call because an older instance of the
method (when it was Puppet::Util.execute), would raise an exception if
the argument to it was not an array. The current implementation accepts
both string and array, so the wrapping of it in an array is no longer
required.

### DIFF
--- a/lib/puppet/resource/type_collection.rb
+++ b/lib/puppet/resource/type_collection.rb
@@ -187,7 +187,7 @@ class Puppet::Resource::TypeCollection
       if environment.config_version.nil? || environment.config_version == ""
         @version = Time.now.to_i
       else
-        @version = Puppet::Util::Execution.execute([environment.config_version]).strip
+        @version = Puppet::Util::Execution.execute(environment.config_version).strip
       end
     end
 
--- a/spec/unit/resource/type_collection_spec.rb
+++ b/spec/unit/resource/type_collection_spec.rb
@@ -320,7 +320,7 @@ describe Puppet::Resource::TypeCollection do
       let(:environment) { Puppet::Node::Environment.create(:testing, [], '', '/my/foo') }
 
       it "should use the output of the environment's config_version setting if one is provided" do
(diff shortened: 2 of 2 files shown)


haus:
Relates to https://github.com/puppetlabs/puppet-server/pull/962. I'll probably try adding an acceptance test for a more complex config_version command as well.

lib/puppet/resource/type_collection.rb
```diff
@@ -187,7 +187,7 @@ def version
       if environment.config_version.nil? || environment.config_version == ""
         @version = Time.now.to_i
       else
-        @version = Puppet::Util::Execution.execute([environment.config_version]).strip
+        @version = Puppet::Util::Execution.execute(environment.config_version).strip
```
joshcooper: Will this break puppet not running under puppetserver? Will this cause the `config_version` to be executed by the shell instead of `execve`?

haus: I'm not sure what `execve` is.

According to the yard doc for Puppet::Util::Execution.execute, the first argument should only be an array if the first element of the array is the executable, and should be a string otherwise.

joshcooper: Sorry, I mean if you pass an `Array` to ruby's `Kernel.exec` it will execute the process using `execv` directly, bypassing the shell. If you pass a string, it will shell out: https://github.com/ruby/ruby/blob/ruby_2_1/process.c#L3127-L3135.

If we don't arrayify and you're using webrick/passenger, I think we'll end up shelling out when previously we didn't (but it would be good to confirm that).

haus: Today I learned new things about ruby and process execution.

@joshcooper that's good to know. If that's why this and other calls to execute are all still arrayified, that would be good to note in comments and/or the yard doc for the execute method.

PR #4784, opened 2016-03-11
(maint) replace skips with expect_failure in service enable linux

er0ck:
The previous version of this test had some heavy-handed skip_tests which
would not remind us to fix the test when upstream bug is fixed. This
change uses expect_failure to remind us to fix the test. This came out
of some some related beaker work with this file as an example. free bird
in the hand, or something.

closes: https://github.com/puppetlabs/puppet/pull/4350
[skip ci]

### DIFF
--- a/acceptance/tests/resource/service/service_enable_linux.rb
+++ b/acceptance/tests/resource/service/service_enable_linux.rb
@@ -89,44 +89,60 @@ agents.each do |agent|
   is_sysV = ((platform == 'centos' || platform == 'el') && majrelease < 7) ||
               platform == 'debian' || platform == 'ubuntu' ||
              (platform == 'sles'                        && majrelease < 12)
+
   apply_manifest_on(agent, manifest_service_disabled, :catch_failures => true)
-  apply_manifest_on(agent, manifest_service_enabled, :catch_failures => true) do
-    if is_sysV
-      # debian platforms using sysV put rc runlevels directly in /etc/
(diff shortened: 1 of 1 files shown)


acceptance/tests/resource/service/service_enable_linux.rb
```diff
@@ -119,10 +124,17 @@
     if is_sysV
       rc_symlinks = on(agent, "find /etc/ -name *#{package_name[platform]}", :accept_all_exit_codes => true).stdout
       # sles removes rc.d symlinks
-      if platform != 'sles'
+      if platform != 'sles' && (platform != 'debian' && majrelease != 8)
         (start_runlevels + kill_runlevels).each do |runlevel|
```
kylog: So is this entire step a skip on sles? If so, it would be clearer if the `platform != sles` conditional were up around the entire step block.

acceptance/tests/resource/service/service_enable_linux.rb
```diff
@@ -119,10 +124,17 @@
...
           assert_match("#{runlevel}.d/#{kill_symlink}", rc_symlinks, "did not find #{kill_symlink} in runlevel #{runlevel}")
         end
+      elsif platform == 'debian' && majrelease == 8
```
kylog: If this comparison/block is ordered to come first, then we don't need two separate "is it deb8?" comparisons.

joshcooper:
This has a conflict and needs rebasing to master.

er0ck:
no longer applicable

PR #4787, opened 2016-03-11
(maint) Update catalog format api document for version

haus:
Previously, version was specified to be an integer. However, version is often
overridden by using a custom config_version script, which will return a
string. This commit updates the schema to include string for the
version type.

### DIFF
--- a/api/schemas/catalog.json
+++ b/api/schemas/catalog.json
@@ -16,7 +16,7 @@
             "type": "string"
         },
         "version": {
-            "type": "integer"
+            "type": ["string", "integer"]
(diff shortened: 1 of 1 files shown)


hlindberg:
:+1: 
Feel free to push the "Merge pull request" button when travis goes green (there may be a test to check schema).

joshcooper:
:+1: I think we're also holding off until after the releases.

hlindberg:
travis for ruby 2.0.0 failed (looked like a timing issue / process kill). 

Restarted the job.

PR #4788, opened 2016-03-13
PUP-6048: always create a basic default environment

vicinus:
At the moment puppet only creates a basic default environment, if the
environment is named: `production`. I don't see a reason why production needs a
special handling in this case. Also changing this behavior will resolve the
problem, that 'puppet facts --environment test' throws an error if the
environment doesn't exist in the file system.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -184,12 +184,10 @@ module Puppet
       # in case the configured environment (used for the default sometimes)
       # doesn't exist
       default_environment = Puppet[:environment].to_sym
-      if default_environment == :production
-        loaders << Puppet::Environments::StaticPrivate.new(
-          Puppet::Node::Environment.create(default_environment,
-                                           basemodulepath,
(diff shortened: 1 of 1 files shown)


kylog:
Good conversation in the ticket. Let's resolve what we want to do functionally in the ticket before tackling the PR.

That said, process nit: the PR commit message is what's causing the travis failure. You'll want to update the commit message to have parens around the ticket number like so: `(PUP-6048) always create a default environment`.

logicminds:
@kylog it would be great to have some automated process that checks for this kind of metadata in the PR `(PUP-6048)` and comments in the PR with the corrective action.

kylog:
@logicminds completely agreed. Right now, we just have the travis check but that's a bit buried. A comment in the PR would definitely be nicer. I have a hack-day project that does just this - I should dust that off.

logicminds:
@kylog See https://github.com/servo/highfive

PR #4789, opened 2016-03-13
(PUP-6050) Improve puppet lookup cli

ripienaar:
This fixes the documentation and handling of knock out prefix and unpack
arrays options


hlindberg:
Thanks @ripienaar, we will look at this asap. May be too late for 4.4.0, but should go into 4.4.1.

ripienaar:
no worries, thought this would be clearer than the description of the problem :) discard/rewrite at will

hlindberg:
@ripienaar thomas also merged this into stable; so will be released in 4.4.1 or 4.5.0 whatever comes first. Thanks for the contribution.

PR #4790, opened 2016-03-15
(PUP-6059) Perform fsync on temp file before doing checksum verification

mpkut:
Checksum verification of new files can incorrectly fail on some filesystem
types such as GlusterFS, which causes Puppet runs to fail. This patch calls
file.fsync before performing the checksum verification, which increases
the chance that all file descriptors opened on the file will return
consistent data and allow the checksum verification to succeed.

### DIFF
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -856,6 +856,18 @@ Puppet::Type.newtype(:file) do
         devfail 'a property should have been provided if write_temporary_file? returned true' if property.nil?
         content_checksum = property.write(file)
         file.flush
+        begin
+          file.fsync
+        rescue NotImplementedError
+          # fsync may not be implemented by Ruby on all platforms, but
+          # there is absolutely no recovery path if we detect that.  So, we just
+          # ignore the return code.
+          #
+          # However, don't be fooled: that is accepting that we are running in
+          # an unsafe fashion.  If you are porting to a new platform don't stub
+          # that out.
+        end
+
         fail_if_checksum_is_wrong(file.path, content_checksum) if validate_checksum?
         if self[:validate_cmd]
           output = Puppet::Util::Execution.execute(self[:validate_cmd].gsub(self[:validate_replacement], file.path), :failonfail => true, :combine => true)


peterhuene:
Hi @mpkut Thanks for the contribution.  If you haven't already, would you mind [signing the CLA](https://cla.puppetlabs.com)?

This looks reasonable; when the block returns the file will be flushed and fsync'd again in the utility function, but that should have a negligible impact on perf.

mpkut:
Hi Peter,

Thank you for your review!

It does seem redundant to call fsync twice on the same file, so this is
definitely a pragmatic - rather than elegant - suggested fix.

My understanding is that our Puppet team within UT Austin ITS recently
conveyed a signed copy to your legal folks, authorizing members of the
"UTAustinPuppet" organization to make submissions per the CLA.
@cbowlesut is handling the setup as we get things going.  If that
paperwork is not yet on your records, please let us know and we'll work
with the legal folks to sort it out.

Regards,

MPK

jeffweiss:
@peterhuene @MikaelSmith I can confirm that Legal has received a CLA for members of the UTAustinPuppet org. Given that this has occurred outside the normal CLA tool, I'll start talking to Legal about what needs to be done to get record of it into the CLA tool so we do not constantly badger @mpkut  and compatriots.

@mpkut We do have the option to "Sign for My Company" in the CLA tool. I may have you or @cbowlesut do this, which will create the appropriate OAuth'd records in CLA. I can then attach the PDF from 1 Feb 2016 signed by Julienne Vanderziel that we have on file.

Thanks for your patience. 😃

MikaelSmith:
This seems reasonable. Which release were you looking to target it at? It could go against either stable (for the next 4.x bug fix release) or 3.x (for the next 3.8 bug fix release, and merged up for the next 4.x bug fix).

mpkut:
Thank you for the review! Given that we are currently using Puppet Enterprise (pe-puppet-3.6.2.14-1.pe.el6.noarch), I suppose the 3.x release would be the better fit for our particular needs, unless PE releases have now moved to the 4.x code.

MikaelSmith:
PE releases are on 4.x code (as of PE 2015.2). However, we still issue bug fix release for PE 3.8.

mpkut:
Fair enough - 4.x seems the better choice. Thanks!

minorOffense:
I would vote for a bug fix for 3.8. We're currently using Glusterfs and this would solve some provisioning issues for us. We plan to move to 4.x but we haven't had the chance to finish the migration.

MikaelSmith:
Since my last comment we've stopped doing 3.x releases.

minorOffense:
Well then. I guess we'll need to patch or upgrade.

PR #4793, opened 2016-03-17

kylog:


MikaelSmith:
Changing run_specs_all_ruby_versions.sh seems odd. Everything else :+1:

PR #4795, opened 2016-03-17
(PUP-6067) Add features to Pip3 provider

MikaelSmith:
Features are not inherited from the parent provider. Without this the
Pip3 provider is pretty limited. Specify features provided by Pip as
also provided by Pip3. Also add spec tests for Pip and Pip3 verifying
their features.


branan:
:+1: this is simple enough

PR #4796, opened 2016-03-17
(PUP-5968) Add deprecation warning for Symbol monkey patch

branan:
We re-introduced a monkey patch that had previously come from a
vendored dependency in order to maintain compatibility during the 4.x
series. This adds a warning that the patch is deprecated.

### DIFF
--- a/lib/puppet/util/monkey_patches.rb
+++ b/lib/puppet/util/monkey_patches.rb
@@ -20,6 +20,14 @@ end
 
 class Symbol
   def <=> (other)
+    if (other.class != Symbol)
+      case Puppet[:strict]
+      when :warning
+        Puppet.warn_once('deprecation', 'symbol_comparison', 'Comparing Symbols to non-Symbol values is deprecated')
+      when :error
+        raise ArgumentError.new("Comparing Symbols to non-Symbol values is no longer allowed")
+      end
+    end
     self.to_s <=> other.to_s
   end
 
--- a/spec/unit/util/monkey_patches_spec.rb
+++ b/spec/unit/util/monkey_patches_spec.rb
@@ -5,9 +5,73 @@ require 'puppet/util/monkey_patches'
 
 
 describe Symbol do
+  after :all do
+    $unique_warnings.delete('symbol_comparison')
+  end
+
   it "should return self from #intern" do
     symbol = :foo
     expect(symbol).to equal symbol.intern
+  end
+
+  describe "when :strict is off" do
+    before :each do
+      Puppet.settings[:strict] = :off
+    end
+
+    after :all do
+      Puppet.settings[:strict] = Puppet.settings.setting(:strict).default
+    end
+
+    it "should not warn if compared against another symbol" do
+      Puppet.expects(:warn_once).never
+      expect(:foo <=> :bar).to equal(1)
+    end
+
+    it "should not warn if compared against a non-symbol value" do
+      Puppet.expects(:warn_once).never
+      expect(:foo <=> "foo").to equal(0)
+    end
+  end
+
+  describe "when :strict is warning" do
+    before :each do
+      Puppet.settings[:strict] = :warning
+    end
+
+    after :all do
+      Puppet.settings[:strict] = Puppet.settings.setting(:strict).default
+    end
+
+    it "should not warn if compared against another symbol" do
+      Puppet.expects(:warn_once).never
+      expect(:foo <=> :bar).to equal(1)
+    end
+
+    it "should warn if compared against a non-symbol value" do
+      Puppet.expects(:warn_once).once
+      expect(:foo <=> "foo").to equal(0)
+    end
+  end
+
+  describe "when :strict is error" do
+    before :each do
+      Puppet.settings[:strict] = :error
+    end
+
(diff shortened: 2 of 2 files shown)


lib/puppet/util/monkey_patches.rb
```diff
@@ -20,6 +20,9 @@ def self.caller(skip=nil)
 
 class Symbol
   def <=> (other)
+    if (other.class != Symbol)
```
kylog: I think there was some concern that this comparison itself could be expensive if there were a lot of them. I don't think this concern warrants a standing benchmark but maybe a one-off test to see if there could be performance implications.

branan: Yeah. I wanted to benchmark how bad it was to add a check for `Puppet[:strict]` here, and I was gonna check the base "no warnings at all" case, too

branan:
Here's my benchmarking results. Here's the [code that generated these numbers](https://gist.github.com/branan/4ab1ee5af3a9493baa0b)

```
simple symbol/symbol comparison    0.250000   0.000000   0.250000 (  0.252422)
simple symbol/string comparison    0.250000   0.000000   0.250000 (  0.250750)
warning symbol/symbol comparison   0.290000   0.000000   0.290000 (  0.287773)
warning symbol/string comparison  16.380000   0.010000  16.390000 ( 16.390936)
strict symbol/symbol comparison    0.300000   0.000000   0.300000 (  0.297906)
strict symbol/string comparison   18.760000   0.000000  18.760000 ( 18.758131)
```

joshcooper:
I wonder whether the expensive part is resolving ` Puppet[:strict]`, calling `Puppet.deprecation_warning`, or both? Could you try looking up `Puppet[:strict]` once (outside the loop) and storing that in a variable?

branan:
@joshcooper It's only looking up `Puppet[:strict]` in the third case. It's calling `Puppet.deprecation_warning` that's super slow

hlindberg:
There is also a warn_once, which I wonder if it is any faster than the general deprecation.

branan:
The numbers above have been updated for the warn_once case. It's much faster than the `deprecation_warning` case, which makes sense to me. `deprecation_warning` has logic that allows it to be issued multiple times if the offending code is called from multiple locations - so basically it's "warn the user for each method they have that calls this deprecated code". `warn_once` does exactly as its name implies - warns once.

The former seems to me to be the behavior we want here, since this is ruby code that users could be accidentally relying on from any number of places in their custom plugins

hlindberg:
Note that warn_once relies on a key wrt uniqueness. The deprecation_warning figures out where a call comes from (and this may be why it is expensive). The deprecation warning is also usually completely wrong about the caller since the actual call often/typically comes from a higher level. It would need to look at the entire call stack in order to figure out if the call is unique or not - IIRC it does not do that (and that would be even more expensive). What I think I am saying is - use the fastest method since you need to debug what is going on anyway to actually know if/what the problem is.

branan:
If the machinery for `deprecation_warning` trying to figure out where the offending code was called from isn't particularly reliable, then I think it's worth it to just use `warn_once` and be done

hlindberg:
IIRC it only looks at immediate caller - which is probably some kind of hash, sort, or other higher order method - not the actual "business logic" a user can do anything about. This is what it does:
```
    if Puppet[:trace]
      caller()[2..-1]
    else
      [caller()[2]]
    end
```

There is also `puppet_deprecation_warning` where the file and line can be fed in. To say anything useful, it would have to search backwards on the stack and use divine powers to figure out which one to select.

So yeah - warn_once.

kylog:
Is that logic in `deprecation_warning` (to report the immediate caller) useful? As I reflect on it, I usually see that as noise. Should we toss it?

hlindberg:
It works for top level API calls. The take away is - you have to think about the thing you are deprecating and which form of reporting is best. We could have an expanded take on deprecation warnings/warnings and errors in general. There are lots of strange things going on in this domain.

branan:
This is now updated to use `warn_once` and to check the value of `:strict` to determine how it should alert the user that a deprecated invocation has occured.

lib/puppet/util/monkey_patches.rb
```diff
@@ -20,6 +20,13 @@ def self.caller(skip=nil)
...
+      if Puppet[:strict] == :warning
+        Puppet.warn_once('deprecation', 'symbol_comparison', 'Comparing Symbols to non-Symbol values is deprecated')
+      elsif Puppet[:strict] == :error
+        raise ArgumentError.new("Comparing Symbols to non-Symbol values is no longer allowed")
```
hlindberg: You are retrieving the setting twice if it is set to off. Suggest
```
case Puppet[:strict]
when :warning 
...
when :error
end
```

branan: You're right. I've been spoiled by optimizing compilers (and much less flexible case statements) off in C++ land ;)

I'll clean that up

MikaelSmith:
Updated to a case statement. Any more feedback @kylog @hlindberg?

hlindberg:
:+1:  to merge

PR #4797, opened 2016-03-18
(PUP-6071) Wrong selinux type for files in symlinked directories

vinzent:
matchpathcon does only act on the passed string and not resolve the real path.

problematic if puppet manages /etc/init.d/script on RHEL which symlinks
/etc/init.d to /etc/rc.d/init.d. matchpathcon will return the wrong label.

This mimics restorecon behaviour.

### DIFF
--- a/lib/puppet/util/selinux.rb
+++ b/lib/puppet/util/selinux.rb
@@ -47,8 +47,25 @@ module Puppet::Util::SELinux
     rescue Errno::EACCES, Errno::ENOENT
       mode = 0
     end
-
-    retval = Selinux.matchpathcon(file, mode)
+    # matchpathcon does only act on the passed string and not resolve the real path
+    # problematic if puppet manages /etc/init.d/script on RHEL which symlinks
+    # /etc/init.d to /etc/rc.d/init.d. matchpathcon will return the wrong label.
+    #
+    # This mimics restorecon behaviour.
+    #
+    # Not all files requested exist. Fall back to the requested file string.
+    begin
+      if !File.symlink?(file) && File.directory?(file)
+        file_realpath = File.realpath(file)
+      else
+        dirname_realpath = File.realpath(File.dirname(file))
+        file_basename = File.basename(file)
+        file_realpath = dirname_realpath + '/' + file_basename
+      end
+    rescue
+      file_realpath = file
+    end
+    retval = Selinux.matchpathcon(file_realpath, mode)
     if retval == -1
       return nil
     end


vinzent:
i'll try to fix the tests after a general feedback concerning the patch.

lib/puppet/util/selinux.rb
```diff
@@ -47,8 +47,25 @@ def get_selinux_default_context(file)
...
+    # matchpathcon does only act on the passed string and not resolve the real path
+    # problematic if puppet manages /etc/init.d/script on RHEL which symlinks
+    # /etc/init.d to /etc/rc.d/init.d. matchpathcon will return the wrong label.
+    #
+    # This mimics restorecon behaviour.
+    #
+    # Not all files requested exist. Fall back to the requested file string.
+    begin
+      if !File.symlink?(file) && File.directory?(file)
+        file_realpath = File.realpath(file)
```
peterhuene: How does the `File.realpath(file)` here differ from `File.realpath(File.dirname(file)) + '/' + File.basename(file)` performed in the else?  I would assume `File.realpath` resolves a file underneath a symlinked directory correctly, but perhaps that's not the case?

lib/puppet/util/selinux.rb
```diff
@@ -47,8 +47,25 @@ def get_selinux_default_context(file)
...
+      else
+        dirname_realpath = File.realpath(File.dirname(file))
+        file_basename = File.basename(file)
+        file_realpath = dirname_realpath + '/' + file_basename
+      end
+    rescue
```
peterhuene: Is there an exception that can be caught here rather than just a general rescue?  This may hide legitimate bugs and should be avoided; always rescue the most specific exception one is able to, thus avoiding unintentional exception swallowing.

lib/puppet/util/selinux.rb
```diff
@@ -47,8 +47,25 @@ def get_selinux_default_context(file)
...
+    # This mimics restorecon behaviour.
+    #
+    # Not all files requested exist. Fall back to the requested file string.
+    begin
+      if !File.symlink?(file) && File.directory?(file)
+        file_realpath = File.realpath(file)
+      else
+        dirname_realpath = File.realpath(File.dirname(file))
+        file_basename = File.basename(file)
+        file_realpath = dirname_realpath + '/' + file_basename
```
peterhuene: This should be a `File.join` call instead of manually appending a '/' (at least one spec failure is because of passing a file in the root directory, e.g. `/foo` results in `//foo`).

peterhuene:
Hi @vinzent:

Thanks for the contribution!  I've left a few comments on the PR of things that should be addressed.  Please let me know if you have any questions.  Cheers!

lib/puppet/util/selinux.rb
vinzent: you're probably right. This seems to be a leftover from before i've splitted the path in dirname and basename in the else clause. i'll remove it and re-test.

lib/puppet/util/selinux.rb
vinzent: i'll check what concrete exceptions are thrown.

vinzent:
@peterhuene in between i'm not sure if this should be fixed in Puppet. 

tried to reproduce the problem on RHEL7  - but i could not. The matchpathcon utility from libselinux-utils  (v. 2.2.2) gives correct results for `/etc/init.d/xyz` . RHEL6 has version 2.0.94. 

It seems to me the root problem is a bug in the selinux code and already fixed.

kylog:
I do see a couple items in the selinux changelog that look related, e.g.:
* https://github.com/SELinuxProject/selinux/blob/master/libselinux/ChangeLog#L183
* https://github.com/SELinuxProject/selinux/blob/master/libselinux/ChangeLog#L263

Unfortunately they don't reference an issue tracker so hard to get details :(

kylog:
Okay @vinzent that makes sense. Thanks for running that down in the selinux code. I'll go ahead and close this, but feel free to re-open if you learn more and think that this should be fixed in puppet.

PR #4800, opened 2016-03-18
(PUP-6073) Handle launchd plists with line continuations

ccaviness:
Even though line continuations in text plist files is valid XML,
the plist parser in facter 2.x can't handle it. In this case, rescue
NoMethodError and let `plutil -convert xml1' normalize it.


peterhuene:
Hi @ccaviness:

Thanks for the contribution!  It looks like this PR needs a rebase to be able to be merged.

MikaelSmith:
This may be appropriate for 3.8. However it needs to be retargeted at the 3.x branch to fix that.

This problem has already been fixed in Puppet 4.4.

PR #4803, opened 2016-03-22
(PUP-4760) Service type should not retrieve ensure status when unneeded

MikaelSmith:
Before this, service type will always retrieve the ensure status even if no
ensure parameters is given.

But, because no Service Type parameter is dependant on whether the service
is running or not, we can just skip the ensure status retrieval.

Now, a service type will only retrieve the ensure status when a ensure
parameter is given.

### DIFF
--- a/lib/puppet/type.rb
+++ b/lib/puppet/type.rb
@@ -1044,6 +1044,12 @@ class Type
     insync
   end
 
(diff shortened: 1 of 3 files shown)


whopper:
👍

PR #4805, opened 2016-03-22

kylog:


kylog:
After merging up #4793, this is the stable branch followup. /cc @MikaelSmith

kylog:
Thanks. Merged up to master, and nothing in lib/spec there had executable bits. So this little journey is at an end ....

PR #4806, opened 2016-03-23
(PUP-6011) Fix line endings in mini_spec.rb

glennsarti:
The line endings in spec/integration/module_tool/tar/mini_spec.rb
were accidentally merged as CRLF instead of LF.  This commit changes
the endings back to LF.

Ticket to track work PUP-6011


spec/integration/module_tool/tar/mini_spec.rb
```diff
@@ -1,28 +1,28 @@
...
+
+    it "unpacks a tar with a short path length" do
+      extractdir = PuppetSpec::Files.tmpdir('minitar')
+
+      minitar.unpack(sourcetar,extractdir,'module')
+      expect(File).to exist(File.expand_path("#{extractdir}/#{shortfilepath}"))
+    end
+
+    it "unpacks a tar with a long path length" do
+      extractdir = PuppetSpec::Files.tmpdir('minitar')
```
kylog: FYI the houndci integration is something we're experimenting with. I think we'll want to turn this particular cop off because we'll hit it a lot.

/cc @HAIL9000 since we were talking about over-zealous cops yesterday :)

spec/integration/module_tool/tar/mini_spec.rb
```diff
@@ -1,28 +1,28 @@
...
+require 'puppet/module_tool'
+require 'puppet_spec/files'
+
+describe Puppet::ModuleTool::Tar::Mini, :if => (Puppet.features.minitar? && Puppet.features.zlib?) do
+  let(:minitar)    { described_class.new }
+
+  describe "Extracts tars with long and short pathnames" do
+    let (:sourcetar) { File.expand_path('../../../../fixtures/module.tar.gz', __FILE__) }
+
+    let (:longfilepath)  { "puppetlabs-dsc-1.0.0/lib/puppet_x/dsc_resources/xWebAdministration/DSCResources/MSFT_xWebAppPoolDefaults/MSFT_xWebAppPoolDefaults.schema.mof" }
```
glennsarti: Well of course it's too long.  I'm checking for long file names!

glennsarti:
This has a ticket number (PUP-) instead of maint so that the work would not be forgotten.

PR #4807, opened 2016-03-23
(PUP-6085) Prevent that subkeys can do lookups in string values

thallgren:
Before this commit, a subkey of a dotted key could be matched by a
string that contained a matching text. Hiera would require that the
value responded to :[] and that it was not an Array (arrays should use
numeric index). A string fulfills that requirement.

This commit adds the constraint that the value cannot be a String when
subjected to a lookup.

### DIFF
--- a/lib/puppet/data_providers/hiera_interpolate.rb
+++ b/lib/puppet/data_providers/hiera_interpolate.rb
@@ -100,10 +100,16 @@ module Puppet::DataProviders::HieraInterpolate
       throw :no_such_key if value.nil?
       if segment =~ /^[0-9]+$/
         segment = segment.to_i
-        raise Puppet::DataBinding::LookupError, "Data provider type mismatch: Got #{value.class.name} when Array was expected to enable lookup using key '#{segment}'" unless value.instance_of?(Array)
+        unless value.instance_of?(Array)
+          raise Puppet::DataBinding::LookupError,
+                "Data provider type mismatch: Got #{value.class.name} when Array was expected to enable lookup using key '#{segment}'"
+        end
         throw :no_such_key unless segment < value.size
       else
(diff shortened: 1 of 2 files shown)


spec/unit/data_providers/hiera_interpolation_spec.rb
```diff
@@ -217,6 +218,11 @@ def expect_lookup(*keys)
       expect_lookup('a.d')
       expect(interpolator.interpolate("a dot f: %{lookup(\"'a.d'\")}", lookup_invocation, true)).to eq('a dot f: ')
     end
+
+    it 'should not find a subkey that is matched within a string' do
+      expect_lookup('key')
```
thallgren: I prefer single quoted strings because they cause less overhead for the Ruby parser/evaluator.

lib/puppet/data_providers/hiera_interpolate.rb
```diff
@@ -100,10 +100,16 @@ def qualified_lookup(segments, value)
...
         segment = segment.to_i
-        raise Puppet::DataBinding::LookupError, "Data provider type mismatch: Got #{value.class.name} when Array was expected to enable lookup using key '#{segment}'" unless value.instance_of?(Array)
+        unless value.instance_of?(Array)
+          raise Puppet::DataBinding::LookupError,
+                "Data provider type mismatch: Got #{value.class.name} when Array was expected to enable lookup using key '#{segment}'"
+        end
         throw :no_such_key unless segment < value.size
       else
-        raise Puppet::DataBinding::LookupError, "Data provider type mismatch: Got #{value.class.name} when a non Array object that responds to '[]' was expected to enable lookup using key '#{segment}'" unless value.respond_to?(:'[]') && !value.instance_of?(Array)
+        unless value.respond_to?(:'[]') && !(value.instance_of?(Array) || value.instance_of?(String))
```
thallgren: I don't think 80 characters is a good limit. Perhaps in commit comments, but not in code. I tend to use a 120 character limit.

hlindberg:
:+1:  merge after unfreeze

PR #4808, opened 2016-03-23
(maint) Add a first pass at HoundCI configuration

kylog:


kylog:
Ping @thallgren and @glennsarti who got hit with some of the first houndbot comments.

Also ping @HAIL9000 @MikaelSmith @peterhuene whom I suggested this to yesterday during PR triage.

MikaelSmith:
I have no familiarity with Rubocop. So sure.

hlindberg:
The line length 120 is problematic - we need to allow longer lines for some error messages when they occur in indented logic. I avoid to avoid us writing slow code just to avoid being slammed on this. Naturally you want shorter lines if possible, but if that means using lots of temporary variables when none are needed just to shorten interpolations, or use funky linebreaks in interpolation to just please the cop...

And the enforcement of " " over ' ' is just wrong for performance reasons.

HAIL9000:
The enabled options all look fine to me 
thumbs-up-computer-kid-gif_348542

PR #4810, opened 2016-03-24
(maint) Update Hound to use puppet's current rubocop config

kylog:


kylog:
@hlindberg this updates rubocop version to the version 0.36 that hound seems to be using.

kylog:
Btw, since I'm taking off for a week: I've disabled Hound for now. I think that the next steps here will be:

- [ ] 1)  merge this PR
- [ ] 2) enable Hound
- [ ] 3) verify that it's not noisy with some real-world PRs - fix if it is still noisy
- [ ] 4) put up a followon PR enabling just 1 more cop, with a combination of sanitizing and rubocop disable comments, to get a feel for how that works

PR #4811, opened 2016-03-24
(PUP-1364) Detect arch argument in yum installs

petems:
* Checks for arch in package name
* Moves to end of install arguments

Originally there was a plan to add an arch parameter (See https://tickets.puppetlabs.com/browse/PUP-3263), but this doesn't help the use case where two packages have to be installed eg. foo-lib.i686 and foo-lib.x86_64 both have to be installed. Having two packages with the same name would be a conflict.

### DIFF
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -13,6 +13,30 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
 
   commands :cmd => "yum", :rpm => "rpm"
 
+  YUM_ARCH_LIST = [
+    'i386',
+    'i686',
+    'ppc',
+    'ppc64',
+    'armv3l',
+    'armv4b',
+    'armv4l',
+    'armv4tl',
+    'armv5tel',
+    'armv5tejl',
+    'armv6l',
+    'armv7l',
+    'm68kmint',
+    's390',
+    's390x',
+    'ia64',
+    'x86_64',
+    'sh3',
+    'sh4',
+  ]
+
+  ARCH_REGEX = Regexp.new(YUM_ARCH_LIST.join('|\.'))
+
   if command('rpm')
     confine :true => begin
       rpm('--version')
@@ -149,6 +173,10 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
     else
       # Add the package version
       wanted += "-#{should}"
+      if wanted.scan(ARCH_REGEX)
+        self.debug "Detected Arch argument in package! - Moving arch to end of version string"
+        wanted.gsub!(/(.+)(#{ARCH_REGEX})(.+)/,'\1\3\2')
+      end
       is = self.query
       if is && yum_compareEVR(yum_parse_evr(should), yum_parse_evr(is[:ensure])) < 0
         self.debug "Downgrading package #{@resource[:name]} from version #{is[:ensure]} to #{should}"
--- a/spec/unit/provider/package/yum_spec.rb
+++ b/spec/unit/provider/package/yum_spec.rb
@@ -249,6 +249,32 @@ describe provider_class do
       Puppet::Util::Execution.expects(:execute).with(['/usr/bin/yum', '-d', '0', '-e', '0', '-y', :install, name])
       provider.install
     end
+
+    context 'with arch arguments in package name' do
+      let(:package_name) { 'mypackage' }
+      let(:arch_string) { 'x86_64' }
+      let(:name) { "#{package_name}.#{arch_string}" }
+
+      it 'should be able to set version with a name with arch string' do
+        version = '1.2'
+        resource[:ensure] = version
+        Puppet::Util::Execution.expects(:execute).with(['/usr/bin/yum', '-d', '0', '-e', '0', '-y', :install, "#{package_name}-#{version}.#{arch_string}"])
+        provider.stubs(:query).returns :ensure => version
+        provider.install
+      end
+
+      it 'should be able to set installed with a name with arch string' do
+        resource[:ensure] = :installed
+        Puppet::Util::Execution.expects(:execute).with(['/usr/bin/yum', '-d', '0', '-e', '0', '-y', :install, "#{package_name}.#{arch_string}"])
+        provider.install
+      end
+
+      it 'should be able to set installed with a name with arch string' do
+        resource[:ensure] = :installed
+        Puppet::Util::Execution.expects(:execute).with(['/usr/bin/yum', '-d', '0', '-e', '0', '-y', :install, "#{package_name}.#{arch_string}"])
+        provider.install
+      end
+    end
   end
 
   describe 'when uninstalling' do


kylog:
@petems I turned on HoundCI as an experiment, but it's clearly too verbose. Ignore all those bot comments. Cleanup coming.

petems:
@kylog No worries, some of them are pretty valid and I can fix! :smile:

peterhuene:
Ignore the double quoted string one though (rageface).

petems:
Not sure if this is the best way of doing things, but this ticket's fairly old now (2014, imported from RedMine) and could probably do with fixing sooner rather than later? 

I made a module as a workaround (https://github.com/petems/petems-yum_arch) that basically follows this behaviour as a dedicated provider.

lib/puppet/provider/package/yum.rb
```diff
@@ -13,6 +13,30 @@
...
+    'm68kmint',
+    's390',
+    's390x',
+    'ia64',
+    'x86_64',
+    'sh3',
+    'sh4',
+  ]
+
+  ARCH_REGEX = Regexp.new(YUM_ARCH_LIST.join('|\.'))
```
kylog: Seems like this should prepend `\.` to the regex so that the first item is also prefaced with a `.`.

kylog: Also is arch always last in the package name? If so, we could tighten the regex some accordingly.

kylog:
Also, @petems I think we want to tackle this at the same time that we tackle https://tickets.puppetlabs.com/browse/PUP-6097, https://tickets.puppetlabs.com/browse/PUP-5802, and https://tickets.puppetlabs.com/browse/PUP-5025. We've tentatively scheduled those two sprints out from now, so we'll circle back to this PR around then. It might be good to have some input on use cases for some of this stuff so we might ping you for ideas if you're up for that.

joshcooper:
One concern is whether `arch` should ever be part of the insync comparison? If so then `compare_EVR` needs to take arch into account. Based on:

> That should really also be included in the package identity somehow, as it's perfectly valid to tell yum to install "packagename-1.2.3.x86_64" and "packagename-1.2.3.i686" simultaneously

I think it does need to be part of the insync check. For example, if you only have i686 installed, and you tell puppet to install both, it should recognize that i686 is already installed but x86_64 is not.

petems:
@kylog Cool 👍

lib/puppet/provider/package/yum.rb
petems: Technically know, but if it's not you're doing it wrong according to the docs:

> While RPM will run just as well if a package file has been renamed, when the packages are created during RPM's build process, they follow a specific naming convention. The convention is:
name-version-release.architecture.rpm 
- http://www.rpm.org/max-rpm/ch-rpm-file-format.html

Do we want to cover the case of incorrectly named RPMs?

petems:
@joshcooper Are you talking about the usecase of something like:

```puppet
    package{ 'firefox.x86_64':
      ensure   => 'latest',
      provider => yum_arch,
    }

    package{ 'firefox.i686':
      ensure   => 'latest',
      provider => yum_arch,
    }
```

With `firefox.x86_64` installed and `firefox.i686` not? Because that seems to work in my workaround module...

joshcooper:
@petems oh interesting. That works because the arch is part of the package `name`, not the `ensure` version. So I think that's the big question, should `arch` always be part of the name? Or part of the `ensure` version? Would you ever have/want to do something like:

```puppet
 package{ 'firefox':
   ensure   => '46.0.x86_64',
   provider => yum,
}

package{ 'firefox':
  ensure   => '46.0.i686',
  provider => yum,
}
```

petems:
@joshcooper I don't think so. Packages with a different arch are considered different in yum and and if you want to specify a specifc arch version when doing a yum install you do `yum install packagename.arch` so it makes sense to keep the arch as the name variable rather than the version:

```
[root@centos-testing-2gb~]# yum install firefox.i686
Loaded plugins: fastestmirror
Setting up Install Process
Loading mirror speeds from cached hostfile
 * base: mirror.sov.uk.goscomb.net
 * epel: mirrors.coreix.net
 * extras: centos.serverspace.co.uk
 * updates: mirrors.coreix.net
Resolving Dependencies
--> Running transaction check
---> Package firefox.i686 0:45.1.0-1.el6.centos will be installed
--> Processing Dependency: nss >= 3.21.0 for package: firefox-45.1.0-1.el6.centos.i686
--> Processing Dependency: nspr >= 4.11.0 for package: firefox-45.1.0-1.el6.centos.i686
--> Processing Dependency: system-bookmarks for package: firefox-45.1.0-1.el6.centos.i686
--> Processing Dependency: redhat-indexhtml for package: firefox-45.1.0-1.el6.centos.i686
--> Processing Dependency: mozilla-filesystem for package: firefox-45.1.0-1.el6.centos.i686
--> Processing Dependency: libz.so.1 for package: firefox-45.1.0-1.el6.centos.i686
```

This means you can specify the version and the arch also:

```puppet
package{ 'firefox.x86_64':
   ensure   => '38.0.1-1.el6.centos',
   provider => yum_arch,
}
```

BTW my workaround module for it is here, it's basically the code from this PR split off into it's own namespace: https://github.com/petems/petems-yum_arch

spec/unit/provider/package/yum_spec.rb
```diff
@@ -249,6 +249,32 @@
...
+        provider.install
+      end
+
+      it 'should be able to set installed with a name with arch string' do
+        resource[:ensure] = :installed
+        Puppet::Util::Execution.expects(:execute).with(['/usr/bin/yum', '-d', '0', '-e', '0', '-y', :install, "#{package_name}.#{arch_string}"])
+        provider.install
+      end
+
+      it 'should be able to set installed with a name with arch string' do
```
whopper: @petems this test looks to be a duplicate of the one above it

whopper:
@petems cool, thanks! We're working on fixing the yum behavior around epochs when the version is specified in `ensure` so we're going to want to look at this in conjunction with our other work to ensure everything is in order. We'll review this soon, hopefully within the next few days if all goes well on our end. See https://github.com/whopper/puppet/commit/d76506dd3c941c847c8b4ac0a42a8aa49acbb262.

whopper:
@petems @joshcooper r.e `arch` in `ensure` string, there is one case where arch must be provided in the version string, and that's when the package has an epoch.

For example, if package `foo` is version 1:1.2.3, the yum commands that would work to install it would be: `yum install foo-1.2.3` and `yum install foo-1:1.2.3.x86_64`. If the `epoch` is given in that string, then so too must be the `arch`:

Yum install formats:
```
name.arch
name-ver
name-ver-rel
name-ver-rel.arch
name-epoch:ver-rel.arch
epoch:name-ver-rel.arch
```

We'll need to look into the implications of this for the work we're doing around epochs.

whopper:
@petems we've absorbed your work into https://github.com/puppetlabs/puppet/pull/4926, which rolls the arch detection into a few other fixes. Thanks for the contribution!

petems:
Rad 👍

PR #4812, opened 2016-03-24
(PUP-6028) Treat unresolved type references as errors

thallgren:
This PR changes the type parser so that it relies fully on the loader that is passed to it. If the loader cannot find a type that is well known, then it will always return a TypeReference type. The PR also contains what is needed to give the correct feedback when such unresolved references are encountered (they should never be present if the code is correct).

The details are in the commit comments.

### DIFF
--- a/lib/puppet/metatype/manager.rb
+++ b/lib/puppet/metatype/manager.rb
@@ -116,6 +116,14 @@ module Manager
     klass.providerloader.loadall Puppet.lookup(:current_environment)
     klass.providify unless klass.providers.empty?
 
+    if klass.is_capability?
+      # A capability type is a resource (sort of) so it must be registered with the loader
+      loader = Puppet::Pops::Loaders.loaders.public_environment_loader
+      string_name = name.to_s
+      typed_name = Puppet::Pops::Loader::Loader::TypedName.new(:type, string_name.downcase)
+      loader.set_entry(typed_name, Puppet::Pops::Types::PResourceType.new(string_name.capitalize))
+    end
+
     klass
   end
 
--- a/lib/puppet/metatype/manager.rb
+++ b/lib/puppet/metatype/manager.rb
@@ -116,13 +116,8 @@ module Manager
     klass.providerloader.loadall Puppet.lookup(:current_environment)
     klass.providify unless klass.providers.empty?
 
-    if klass.is_capability?
-      # A capability type is a resource (sort of) so it must be registered with the loader
-      loader = Puppet::Pops::Loaders.loaders.public_environment_loader
-      string_name = name.to_s
-      typed_name = Puppet::Pops::Loader::Loader::TypedName.new(:type, string_name.downcase)
-      loader.set_entry(typed_name, Puppet::Pops::Types::PResourceType.new(string_name.capitalize))
-    end
+    loc = block_given? ? block.source_location : nil
+    Puppet::Pops::Loaders.register_runtime3_type(name, loc.nil? ? nil : URI("#{loc[0]}?line=#{loc[1]}"))
 
     klass
   end
--- a/lib/puppet/parser/ast/pops_bridge.rb
+++ b/lib/puppet/parser/ast/pops_bridge.rb
@@ -189,7 +189,9 @@ class Puppet::Parser::AST::PopsBridge
     end
 
     def instantiate_ResourceTypeDefinition(o, modname)
-      Puppet::Resource::Type.new(:definition, o.name, @context.merge(args_from_definition(o, modname)))
+      instance = Puppet::Resource::Type.new(:definition, o.name, @context.merge(args_from_definition(o, modname)))
+      Puppet::Pops::Loaders.register_runtime3_type(instance.name, Puppet::Pops::Adapters::SourcePosAdapter.adapt(o).to_uri)
+      instance
     end
 
     def instantiate_CapabilityMapping(o, modname)
--- a/lib/puppet/pops/loader/static_loader.rb
+++ b/lib/puppet/pops/loader/static_loader.rb
@@ -32,13 +32,12 @@ class StaticLoader < Loader
   def to_s()
     "(StaticLoader)"
   end
+
   private
 
   def load_constant(typed_name)
     @loaded[typed_name]
   end
-
-  private
 
   # Creates a function for each of the specified log levels
   #
--- a/lib/puppet/pops/loader/static_loader.rb
+++ b/lib/puppet/pops/loader/static_loader.rb
@@ -136,11 +136,17 @@ class StaticLoader < Loader
       Zfs
       Zone
       Zpool
-    }.each do |name |
-      typed_name = TypedName.new(:type, name.downcase)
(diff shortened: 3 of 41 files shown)


lib/puppet/pops/loader/static_loader.rb
```diff
@@ -93,6 +93,7 @@ def create_resource_type_references()
       Exec
       File
       Filebucket
+      Generator
```
hlindberg: Out of curiosity - where is that type defined?

spec/lib/puppet_spec/compiler.rb
```diff
@@ -7,8 +7,8 @@ def compile_to_catalog(string, node = Puppet::Node.new('foonode'))
     Puppet::Parser::Compiler.compile(node).filter { |r| r.virtual? }
   end
 
-  def compile_to_ral(manifest)
-    catalog = compile_to_catalog(manifest)
+  def compile_to_ral(manifest, node = Puppet::Node.new('foonode'))
```
hlindberg: "foonode" does not seem to be a very good name here, was it selected because of existing tests using that name everywhere?

hlindberg:
The change "Let PResourceType retain capital letters in the type name" means we should update documentation that users should use a correctly cased type reference - preferably in quotes - i.e. `Resource['File']` rather than `Resource['file']`.

lib/puppet/resource.rb
```diff
@@ -516,6 +524,36 @@ def prune_parameters(options = {})
...
+    elsif argtitle                                             then [ argtype,            argtitle      ]
+    elsif argtype.is_a?(Puppet::Type)                          then [ argtype.class.name, argtype.title ]
+    else raise ArgumentError, "No title provided and #{argtype.inspect} is not a valid resource reference"
+    end
+  end
+  private_class_method :extract_type_and_title
+
+  def self.munge_type_name(value)
+    return :main if value == :main
+    return TYPE_CLASS if value == "" or value.nil? or value.to_s.downcase == "component"
```
hlindberg: Use `||` instead of `or`

lib/puppet/resource.rb
```diff
@@ -516,6 +524,36 @@ def prune_parameters(options = {})
...
+    if type == TYPE_CLASS
+      title = title == '' ? :main : munge_type_name(title)
+    end
+    [type, title]
+  end
+
+
+  def self.extract_type_and_title(argtype, argtitle)
+    if    (argtype.nil? || argtype == :component || argtype == :whit) &&
+      argtitle =~ /^([^\[\]]+)\[(.+)\]$/m                 then [ $1,                 $2            ]
```
hlindberg: picky, logic slightly mis-aligned

hlindberg:
The commit "Make TypeMismatchDescriber discover unresolved types" has typos in the commit message (picky) "and" instead of "an"

lib/puppet/metatype/manager.rb
```diff
@@ -118,10 +118,9 @@ def newtype(name, options = {}, &block)
 
     if klass.is_capability?
       # A capability type is a resource (sort of) so it must be registered with the loader
-      loader = Puppet::Pops::Loaders.loaders.public_environment_loader
-      string_name = name.to_s
-      typed_name = Puppet::Pops::Loader::Loader::TypedName.new(:type, string_name.downcase)
-      loader.set_entry(typed_name, Puppet::Pops::Types::PResourceType.new(string_name.capitalize))
+      loader = Puppet::Pops::Loaders.loaders.resource_type_loader
```
hlindberg: I don't get this. Why is it asking for a "resource_type_loader" specifically here instead of an environment or module related loader? Seems like a mis-match in concepts

lib/puppet/parser/ast/pops_bridge.rb
```diff
@@ -189,7 +189,11 @@ def instantiate_HostClassDefinition(o, modname)
     end
 
     def instantiate_ResourceTypeDefinition(o, modname)
-      Puppet::Resource::Type.new(:definition, o.name, @context.merge(args_from_definition(o, modname)))
+      instance = Puppet::Resource::Type.new(:definition, o.name, @context.merge(args_from_definition(o, modname)))
+      loader = Puppet::Pops::Loaders.loaders.resource_type_loader
```
hlindberg: same here

spec/lib/puppet_spec/compiler.rb
thallgren: Yes. This is the name of the default node for compile_to_catalog (see lines above the diff).

lib/puppet/metatype/manager.rb
thallgren: The resource_type_loader _is_ environment related. Can't use something module related since all types are environment global.

lib/puppet/pops/loaders.rb
```diff
@@ -92,6 +92,10 @@ def puppet_system_loader
     @puppet_system_loader
   end
 
+  def resource_type_loader
+    @resource_type_loader
+  end
+
```
hlindberg: This looks wrong conceptually as the loaders are built in a hierarchy, all things loaded fits into that hierarchy and "resource types in the environment" is not a separate hierarchical layer.
I understand that since it is loading using a 3.x loader this loader is also a module loader, and it is thus quite different from both the environment and module loaders. I think I would be happier if it was named "3xLegacyLoader" or something like that - that shows the intent in a much better way. (I think that is all that is needed to address my concern/problem both here, and where I questioned why the resource_loader was directly requested from pops bridge.

lib/puppet/pops/types/type_parser.rb
```diff
@@ -207,20 +207,23 @@ def interpret_QualifiedReference(name_ast, context)
...
-          loader = context
-        else
-          loader = Puppet::Pops::Adapters::LoaderAdapter.loader_for_model_object(name_ast, context)
-        end
-        unless loader.nil?
-          type = loader.load(:type, name)
-          type = type.resolve(self, loader) unless type.nil?
-        end
-        type || TypeFactory.resource(name)
+      loader = loader_from_context(name_ast, context)
```
hlindberg: does it handle the nil case ?

thallgren:
Pushed a new sequence of commits, rebased on current master where the following has been fixed:

- The 3x resource loader is now named `Runtime3TypeLoader` (a name cannot start with a digit, and Runtime3XXX has been used elsewhere).
- The commit concerning the `Generator` type was removed. It was a mistake on my part an it shouldn't have been there in the first place.
- All ruby types are now registered with the Runtime3TypeLoader. Not just the ones related to capabilities.
- `&&` and `||` is used instead of keywords `and` and `or`
- Misaligned expression was adjusted
- Typo in commit comment fixed

lib/puppet/pops/loader/static_loader.rb
thallgren: It isn't. My mistake. This commit has now been removed from the PR.

hlindberg:
Ok to merge, but I think we need to represent special resource types like Capability and others as separate types to be 100% correct. We can do that work later.

PR #4816, opened 2016-03-25
(maint) Remove comment referencing trac

haus:
The comment at the start of lib/puppet/reference/providers.rb was added in https://github.com/puppetlabs/puppet/commit/7835d2927ef360216612f1a782e88a6606a79d6b in 2007.
As we no longer use trac, this commit removes that comment.


kylog:
But ... I loved that comment so much. Sniff.

kylog:
Also: since the PR branch was on upstream, I went ahead and deleted that branch after merge.

PR #4818, opened 2016-03-28
(PUP-5482) Flag that a type could not be found

haus:

again unnecessarily.

### DIFF
--- a/benchmarks/missing_type_caching/benchmarker.rb
+++ b/benchmarks/missing_type_caching/benchmarker.rb
@@ -1,0 +1,63 @@
+require 'erb'
+require 'ostruct'
+require 'fileutils'
+require 'json'
+
+class Benchmarker
+  include FileUtils
+
+  def initialize(target, size)
+    @target = target
+    @size = size
+  end
+
+  def setup
+    require 'puppet'
+    config = File.join(@target, 'puppet.conf')
+    Puppet.initialize_settings(['--config', config])
+    Puppet[:always_retry_plugins] = false
+  end
+
+  def run(args=nil)
+    env = Puppet.lookup(:environments).get('benchmarking')
+    node = Puppet::Node.new("testing", :environment => env)
+    Puppet::Resource::Catalog.indirection.find("testing", :use_node => node)
+  end
+
+  def generate
+    environment = File.join(@target, 'environments', 'benchmarking')
+    templates = File.join('benchmarks', 'missing_type_caching')
+    test_module_dir = File.join(environment, 'modules', 'testmodule',
+                                'manifests')
+
+    mkdir_p(File.join(environment, 'modules'))
+    @size.times.each do |i|
+      mkdir_p(File.join(environment, 'modules', "mymodule_#{i}", 'lib',
+                        'puppet', 'type'))
+      mkdir_p(File.join(environment, 'modules', "mymodule_#{i}", 'manifests'))
+    end
+    mkdir_p(File.join(environment, 'manifests'))
+
+    mkdir_p(test_module_dir)
+
+
+    render(File.join(templates, 'site.pp.erb'),
+           File.join(environment, 'manifests', 'site.pp'),
+           :size => @size)
+
+    render(File.join(templates, 'module', 'testmodule.pp.erb'),
+           File.join(test_module_dir, 'init.pp'),
+           :name => "foo")
+
+    render(File.join(templates, 'puppet.conf.erb'),
+           File.join(@target, 'puppet.conf'),
+           :location => @target)
+  end
+
+  def render(erb_file, output_file, bindings)
+    site = ERB.new(File.read(erb_file))
+    File.open(output_file, 'w') do |fh|
+      fh.write(site.result(OpenStruct.new(bindings).instance_eval { binding }))
+    end
+  end
+end
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -316,10 +316,10 @@ module Puppet
       :default  => false,
       :hook     => proc { |value|
         Puppet.deprecation_warning "Setting 'always_cache_features' is
-deprecated and has been replaced by 'always_cache_misses'."
+deprecated and has been replaced by 'always_retry_plugins'."
       },
       :desc     => <<-'EOT'
-        This setting is deprecated and has been replaced by always_cache_misses.
+        This setting is deprecated and has been replaced by always_retry_plugins.
 
         Affects how we cache attempts to load Puppet 'features'.  If false, then
         calls to `Puppet.features.<feature>?` will always attempt to load the
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -330,6 +330,27 @@ module Puppet
         improvement for features that are checked frequently.
       EOT
     },
+    :always_cache_misses => {
+        :type     => :boolean,
+        :default  => false,
+        :desc     => <<-'EOT'
+        Affects how we cache attempts to load Puppet 'types' and 'features'.  If
+        false, then calls to `Puppet.type.<type>?` `Puppet.feature.<feature>?`
+        will always attempt to load the type or feature (which can be an
+        expensive operation) unless it has already been loaded successfully.
+        This makes it possible for a single agent run to, e.g., install a
+        package that provides the underlying capabilities for a type or feature,
+        and then later load that type or feature during the same run (even if
+        the type or feature had been tested earlier and had not been available).
+
+        If this setting is set to true, then types and features will only be
+        checked once, and if they are not available, the negative result is
+        cached and returned for all subsequent attempts to load the type or
+        feature.  This behavior is almost always appropriate for the server,
+        and can result in a significant performance improvement for types and
+        features that are checked frequently.
+      EOT
+    },
     :diff_args => {
         :default  => lambda { default_diffargs },
         :desc     => "Which arguments to pass to the diff command when printing differences between
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -340,7 +340,7 @@ deprecated and has been replaced by 'always_cache_misses'."
         :type     => :boolean,
         :default  => false,
         :desc     => <<-'EOT'
-        Affects how we cache attempts to load Puppet 'types' and 'features'.  If
+        Affects how we cache attempts to load Puppet resource types and features.  If
         false, then calls to `Puppet.type.<type>?` `Puppet.feature.<feature>?`
         will always attempt to load the type or feature (which can be an
         expensive operation) unless it has already been loaded successfully.
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -319,6 +319,8 @@ module Puppet
 deprecated and has been replaced by 'always_cache_misses'."
       },
       :desc     => <<-'EOT'
+        This setting is deprecated and has been replaced by always_cache_misses.
+
         Affects how we cache attempts to load Puppet 'features'.  If false, then
         calls to `Puppet.features.<feature>?` will always attempt to load the
         feature (which can be an expensive operation) unless it has already been
@@ -332,8 +334,6 @@ deprecated and has been replaced by 'always_cache_misses'."
         for all subsequent attempts to load the feature.  This behavior is almost
         always appropriate for the server, and can result in a significant performance
         improvement for features that are checked frequently.
-
-        This setting is deprecated and has been replaced by always_cache_misses.
       EOT
     },
     :always_cache_misses => {
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -314,6 +314,10 @@ module Puppet
     :always_cache_features => {
       :type     => :boolean,
       :default  => false,
+      :hook     => proc { |value|
+        Puppet.deprecation_warning "Setting 'always_cache_features' is
+deprecated and has been replaced by 'always_cache_misses'."
+      },
       :desc     => <<-'EOT'
         Affects how we cache attempts to load Puppet 'features'.  If false, then
         calls to `Puppet.features.<feature>?` will always attempt to load the
@@ -328,6 +332,8 @@ module Puppet
         for all subsequent attempts to load the feature.  This behavior is almost
         always appropriate for the server, and can result in a significant performance
         improvement for features that are checked frequently.
+
+        This setting is deprecated and has been replaced by always_cache_misses.
       EOT
     },
     :always_cache_misses => {
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -336,12 +336,12 @@ deprecated and has been replaced by 'always_cache_misses'."
         improvement for features that are checked frequently.
       EOT
     },
-    :always_cache_misses => {
+    :always_retry_plugins => {
         :type     => :boolean,
-        :default  => false,
+        :default  => true,
         :desc     => <<-'EOT'
         Affects how we cache attempts to load Puppet resource types and features.  If
-        false, then calls to `Puppet.type.<type>?` `Puppet.feature.<feature>?`
+        true, then calls to `Puppet.type.<type>?` `Puppet.feature.<feature>?`
         will always attempt to load the type or feature (which can be an
         expensive operation) unless it has already been loaded successfully.
         This makes it possible for a single agent run to, e.g., install a
@@ -349,7 +349,7 @@ deprecated and has been replaced by 'always_cache_misses'."
         and then later load that type or feature during the same run (even if
         the type or feature had been tested earlier and had not been available).
 
-        If this setting is set to true, then types and features will only be
+        If this setting is set to false, then types and features will only be
         checked once, and if they are not available, the negative result is
         cached and returned for all subsequent attempts to load the type or
         feature.  This behavior is almost always appropriate for the server,
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -266,6 +266,8 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
         rescue Puppet::Error => detail
           Puppet.err(detail.to_s) if networked?
           raise
+        ensure
+          Puppet::Type.clear_misses if Puppet[:always_cache_misses]
         end
 
         if checksum_type && config.is_a?(model)
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -267,7 +267,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
           Puppet.err(detail.to_s) if networked?
           raise
         ensure
-          Puppet::Type.clear_misses if Puppet[:always_cache_misses]
+          Puppet::Type.clear_misses if Puppet[:always_retry_plugins]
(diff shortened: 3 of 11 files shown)


lib/puppet/defaults.rb
```diff
@@ -328,6 +332,29 @@ def self.default_diffargs
         for all subsequent attempts to load the feature.  This behavior is almost
         always appropriate for the server, and can result in a significant performance
         improvement for features that are checked frequently.
+
+        This setting is deprecated and has been replaced by always_cache_misses.
+      EOT
+    },
+    :always_cache_misses => {
```
haus: Ping @nfagerlund . Wanted to see if you had any ideas for a better name for this setting.

lib/puppet/defaults.rb
```diff
@@ -328,6 +332,29 @@ def self.default_diffargs
         for all subsequent attempts to load the feature.  This behavior is almost
         always appropriate for the server, and can result in a significant performance
         improvement for features that are checked frequently.
+
+        This setting is deprecated and has been replaced by always_cache_misses.
```
nfagerlund: Please move this line to the top of the setting description, since it's the most important sentence in it.

haus: good call

lib/puppet/defaults.rb
```diff
@@ -328,6 +332,29 @@ def self.default_diffargs
...
+      EOT
+    },
+    :always_cache_misses => {
+        :type     => :boolean,
+        :default  => false,
+        :desc     => <<-'EOT'
+        Affects how we cache attempts to load Puppet 'types' and 'features'.  If
```
nfagerlund: `'types' and 'features'` —> `resource types and features` (no single-quotes)

lib/puppet/defaults.rb
```diff
@@ -330,6 +336,27 @@ def self.default_diffargs
         improvement for features that are checked frequently.
       EOT
     },
+    :always_cache_misses => {
```
nfagerlund: I suggest reversing the behavior of the setting, and naming it to focus on the repeated checks, rather than the cache. As programmers, it's natural for you to see the cache as the most important part, but for a user, it's going to be more useful to think about it in terms of repeated work. 

Best name I can think of on the fly is `:always_retry_plugins`. (I think "plugins" is the best umbrella name for types and provider features... anyone got a better one?)

haus: That sounds like a good place to start.

(And I totally agree about cache vs work, I just couldn't think of a concise setting name that encapsulated it.)

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -267,7 +267,7 @@ def compile(node, options)
           Puppet.err(detail.to_s) if networked?
           raise
         ensure
-          Puppet::Type.clear_misses if Puppet[:always_cache_misses]
+          Puppet::Type.clear_misses if Puppet[:always_retry_plugins]
```
nfagerlund: Does this need a "not"?

haus: no, it needs a not or unless. i missed that in the reversal. this is here so that if a user accidentally uses a type in a manifest before it is available, that the type lookup failure is only cached for the duration of the compile. that maintains backward compatibility with current behavior. otherwise if the user used a type before it was available, they would need to flush the jruby pool or restart the server.

haus:
@joshcooper @hlindberg @thallgren this PR is the new #4427

lib/puppet/defaults.rb
```diff
@@ -314,7 +314,13 @@ def self.default_diffargs
     :always_cache_features => {
       :type     => :boolean,
       :default  => false,
+      :hook     => proc { |value|
+        Puppet.deprecation_warning "Setting 'always_cache_features' is
+deprecated and has been replaced by 'always_cache_misses'."
```
nfagerlund: Gotta update these references to the old name, too!

haus: @nfagerlund thanks for keeping me honest!

nfagerlund: haha no prob! And thank you for all the work on this.

hlindberg:
I have asked @thallgren to take a closer look at what is going on that makes the "cache misses" be of such great value during compilation. We just merged #4812 which removes one peculiarity where puppet creates a resource for the purpose of finding one. I would like us to understand the cause of the performance problem in more depth and fix it in such a way that it always applies during compilation (without an opt-in flag). What we just merged in #4812 may help us  do that.

haus:
@hlindberg sounds great. thanks!

thallgren:
This PR looks good to me. I don't think the changes in #4812 has any impact so I would prefer if this one was merged in.

A benchmark that shows the difference in performance would be very helpful. Both to measure the impact of this PR and future enhancements.

haus:
@nbarrientos I've been trying to construct a benchmark to show the impact of the change, and I've been having a hard time replicating the bad behavior. Could you share some sample manifests that trigger this behavior? How many modules do you have in your environment?

nbarrientos:
@haus sorry, it's been a long time since we reported the issue and amended it locally and I fail to recall all the details. My apologies in advance if I'm not of much help.

As you're probably very aware of, #4427  contains some information that we collected as we were testing the patch and rolling it out across our Puppet pools.

The base catalog that a Puppet-managed node gets over here has 122 resources of type Concat::Fragment. This type is a Puppet defined type and not a Ruby type, so every time the parser was finding it, it had to scan the entire module patch (answering your question, we currently have around 500 modules) generating a huge amount of stat()s just to end up finding out that the type was not a Ruby type (this might have changed in the latest version of the concat module, don't think ours is up to date). All of this was generating ~60k stat() calls unnecessarily (a big fraction of the total, read below).

The way we measured the impact of the patch was by stracing the puppet-server process (and children, of course) and doing statistics on the number of stat()s. We saw that with the patch applied the master needed ~62k stat()s per compilation, and ~125k without it (the ones saved were all the ones trying to find the never-existent lib/puppet/type/concat.rb across all the modules). Fortunately for us, we saw the same ratio improvement filer-side (our Puppet code is served via NFS to the masters) in a global fashion when we deployed the patch everywhere. If you're testing this locally the performance impact should be lower as stat()ing on a single local disk is cheaper than having a bunch of masters hitting a shared NFS volume.

Perhaps new code was introduced between Nov 2015 and today's codebase and the issue was somehow fixed in a different way, I don't know. We're currently running 3.8.4 on top of puppet-server with the patch I initially proposed applied.

Hope this helps you benchmarking the patch proposed on this MR and hopefully confirming that it is still relevant.

N

haus:
@nbarrientos thanks! that helped a bunch. I was able to get a good benchmark going.

haus:
I just pushed out a commit that adds a benchmark for this.

Results of this benchmark before the change (or with the default value of the setting):
```
SIZE=500 bundle exec rake benchmark:missing_type_caching:run
mkdir -p /var/folders/zm/nk5zs3kj1zjdwf3mjcdjs19m0000gq/T/missing_type_caching20160404-6783-1t6mgs8
Run 1        2.260000   1.180000   3.450000 (  3.459899)
Run 2        2.290000   1.180000   3.470000 (  3.473193)
Run 3        2.220000   1.230000   3.450000 (  3.461159)
Run 4        2.230000   1.270000   3.500000 (  3.491843)
Run 5        2.180000   1.120000   3.300000 (  3.308424)
Run 6        2.400000   1.280000   3.680000 (  3.679814)
Run 7        2.360000   1.270000   3.630000 (  3.620446)
Run 8        2.310000   1.200000   3.510000 (  3.519161)
Run 9        2.330000   1.300000   3.630000 (  3.647568)
Run 10       2.340000   1.310000   3.650000 (  3.661020)
> total:    22.920000  12.340000  35.270000 ( 35.322526)
> avg:       2.292000   1.234000   3.527000 (  3.532253)
```

Results with the setting disabled (caching applied):
```
SIZE=500 bundle exec rake benchmark:missing_type_caching:run
mkdir -p /var/folders/zm/nk5zs3kj1zjdwf3mjcdjs19m0000gq/T/missing_type_caching20160404-6886-1oqbd81
Run 1        0.420000   0.050000   0.490000 (  0.479969)
Run 2        0.360000   0.010000   0.370000 (  0.368980)
Run 3        0.290000   0.010000   0.300000 (  0.299707)
Run 4        0.300000   0.010000   0.310000 (  0.307468)
Run 5        0.280000   0.010000   0.290000 (  0.297273)
Run 6        0.320000   0.010000   0.330000 (  0.319802)
Run 7        0.350000   0.010000   0.360000 (  0.359947)
Run 8        0.300000   0.010000   0.310000 (  0.307349)
Run 9        0.260000   0.010000   0.270000 (  0.280855)
Run 10       0.300000   0.010000   0.310000 (  0.320165)
> total:     3.180000   0.140000   3.340000 (  3.341514)
> avg:       0.318000   0.014000   0.334000 (  0.334151)
```

benchmarks/missing_type_caching/benchmarker.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def generate
+    environment = File.join(@target, 'environments', 'benchmarking')
+    templates = File.join('benchmarks', 'missing_type_caching')
+    test_module_dir = File.join(environment, 'modules', 'testmodule',
+                                'manifests')
+
+    mkdir_p(File.join(environment, 'modules'))
+    @size.times.each do |i|
```
haus: These extra directories aren't strictly necessary, but they make the performance gains more apparent. Without these lines, the following results are achieved.

Default setting value (true, no miss caching):
```
 SIZE=500 bundle exec rake benchmark:missing_type_caching:run
mkdir -p /var/folders/zm/nk5zs3kj1zjdwf3mjcdjs19m0000gq/T/missing_type_caching20160404-7358-1ghwvae
Run 1        0.600000   0.100000   0.720000 (  0.714323)
Run 2        0.570000   0.090000   0.660000 (  0.665010)
Run 3        0.620000   0.090000   0.710000 (  0.706283)
Run 4        0.520000   0.090000   0.610000 (  0.609977)
Run 5        0.570000   0.090000   0.660000 (  0.678567)
Run 6        0.550000   0.090000   0.640000 (  0.633102)
Run 7        0.560000   0.090000   0.650000 (  0.649624)
Run 8        0.590000   0.080000   0.670000 (  0.667839)
Run 9        0.560000   0.090000   0.650000 (  0.651638)
Run 10       0.640000   0.080000   0.720000 (  0.722062)
> total:     5.780000   0.890000   6.690000 (  6.698426)
> avg:       0.578000   0.089000   0.669000 (  0.669843)
```

always_retry_plugins set to false (caching enabled):
```
SIZE=500 bundle exec rake benchmark:missing_type_caching:run
mkdir -p /var/folders/zm/nk5zs3kj1zjdwf3mjcdjs19m0000gq/T/missing_type_caching20160404-7451-17f2ypv
Run 1        0.340000   0.020000   0.360000 (  0.372802)
Run 2        0.340000   0.000000   0.340000 (  0.347129)
Run 3        0.290000   0.010000   0.300000 (  0.290051)
Run 4        0.250000   0.000000   0.250000 (  0.263602)
Run 5        0.280000   0.010000   0.290000 (  0.276459)
Run 6        0.300000   0.000000   0.300000 (  0.304726)
Run 7        0.330000   0.000000   0.330000 (  0.334759)
Run 8        0.300000   0.000000   0.300000 (  0.303248)
Run 9        0.260000   0.010000   0.270000 (  0.266815)
Run 10       0.250000   0.010000   0.260000 (  0.253278)
> total:     2.940000   0.060000   3.000000 (  3.012867)
> avg:       0.294000   0.006000   0.300000 (  0.301287)
```

haus:
@hlindberg @thallgren at a guess, i think the value here comes from the search involved in finding a defined type that isn't in a file called `#{name}.pp` but instead in `init.pp`. This is just a guess, but my benchmarks didn't show a difference until I pulled in and was using the concat module, where the concat defined type is in init.pp.

hlindberg:
The search for something like a::b::c::d starts in a/b/c/d.pp, then a/b/c.pp, etc until a.pp, then in init.pp - so if you want to make it really horrible use nested names.

hlindberg:
One run seems to suggest 23 vs 3, and the other 6 vs 3 - What is the difference? Which one resembles real world best?

haus:
The difference is in one case there would be 500 empty module directories (or 500 additional modules installed) and the other is just the single module/type used repeatedly.

The 6 vs 3 is a more isolated case. A single module isn't very real world, but I don't know if 500 modules is either.

thallgren:
@haus, depending on when you ran these tests, can you please try re-running with the absolute latest on master? It would be very interesting to see if my latest fix in PUP-6121 makes a difference. It was merged 3 hours ago though, so perhaps you already did that.

haus:
@thallgren Here are the benchmarks once I rebased off master.

```
 SIZE=500 bundle exec rake benchmark:missing_type_caching:run
mkdir -p /var/folders/zm/nk5zs3kj1zjdwf3mjcdjs19m0000gq/T/missing_type_caching20160404-9730-145ffd9
Run 1        0.450000   0.060000   0.530000 (  0.534420)
Run 2        0.370000   0.010000   0.380000 (  0.384577)
Run 3        0.320000   0.020000   0.340000 (  0.336909)
Run 4        0.350000   0.020000   0.370000 (  0.365459)
Run 5        0.330000   0.010000   0.340000 (  0.349135)
Run 6        0.400000   0.020000   0.420000 (  0.411353)
Run 7        0.350000   0.010000   0.360000 (  0.355262)
Run 8        0.320000   0.010000   0.330000 (  0.338412)
Run 9        0.350000   0.030000   0.380000 (  0.385575)
Run 10       0.350000   0.030000   0.380000 (  0.385661)
> total:     3.590000   0.220000   3.830000 (  3.846763)
> avg:       0.359000   0.022000   0.383000 (  0.384676)
```

```
SIZE=500 bundle exec rake benchmark:missing_type_caching:run
mkdir -p /var/folders/zm/nk5zs3kj1zjdwf3mjcdjs19m0000gq/T/missing_type_caching20160404-9800-1n30u8x
Run 1        2.570000   1.490000   4.070000 (  4.070733)
Run 2        2.600000   1.460000   4.060000 (  4.069924)
Run 3        2.430000   1.440000   3.870000 (  3.865010)
Run 4        2.440000   1.410000   3.850000 (  3.860736)
Run 5        2.490000   1.460000   3.950000 (  3.957650)
Run 6        2.610000   1.570000   4.180000 (  4.210169)
Run 7        2.520000   1.490000   4.010000 (  4.035456)
Run 8        2.680000   1.700000   4.380000 (  4.392967)
Run 9        2.580000   1.570000   4.150000 (  4.160348)
Run 10       2.400000   1.410000   3.810000 (  3.826011)
> total:    25.320000  15.000000  40.330000 ( 40.449004)
> avg:       2.532000   1.500000   4.033000 (  4.044900)
```

thallgren:
I'm confused. Is the current master more than 10 times slower? That's horrible!

haus:
@thallgren i think i reversed the order in one vs the other. i'll fix that up.

Current master matches up with my previous results pretty closely.

hlindberg:
@haus thanks - so roughly from 2x to 7x speedup (extreme case) - that is pretty good.

lib/puppet/metatype/manager.rb
```diff
@@ -146,16 +156,19 @@ def type(name)
     # We are overwhelmingly symbols here, which usually match, so it is worth
     # having this special-case to return quickly.  Like, 25K symbols vs. 300
     # strings in this method. --daniel 2012-07-17
-    return @types[name] if @types[name]
+    return @types[name] if @types.include? name
```
joshcooper: This will do a linear scan which could be bad if there are many types. Might be better to store a sentinel value in the case of a miss, e.g. `@types[name] = CacheMiss`, and then check for that value here, e.g. `return @types[name] if @types[name] != CacheMiss`

haus: Trying to set a value here of anything but nil seems to do some bad things that I don't understand yet (it makes a notify resource invalid). So far I've tried: a class called CacheMiss, a string "CacheMiss", a symbol :CacheMiss.

lib/puppet/reference/providers.rb
```diff
@@ -2,7 +2,7 @@
   types = []
   Puppet::Type.loadall
   Puppet::Type.eachtype do |klass|
-    next unless klass.providers.length > 0
+    next unless klass and klass.providers.length > 0
```
hlindberg: use `&& ` instead of `and`

lib/puppet/metatype/manager.rb
hlindberg: Isn't `@types` here a Hash? If so, `include?` is a hash lookup, which is hard to beat with a faster construct. One can experiment if one of the methods like `key?(name)` is faster (it is probably the very same thing though).

adrienthebo: Looking at the implementation of `hash.c` provides this:

```c
    rb_define_singleton_method(envtbl,"include?", env_has_key, 1);
    rb_define_singleton_method(envtbl,"member?", env_has_key, 1);
    rb_define_singleton_method(envtbl,"has_key?", env_has_key, 1);
    rb_define_singleton_method(envtbl,"key?", env_has_key, 1);
```

`#include?`, `#member?`, `#has_key?`, and `#key?` are all synonymous.

joshcooper: Sorry, my bad, I internalized that as `@types.keys.include?`... so nevermind

haus:
@hlindberg updated to use `&&`.

haus:
I think all of the comments so far have been addressed.

nbarrientos:
Glad you managed to reproduce the gain :)

kylog:
@hlindberg any reservations about merging this at this point? I know there is some work in flight from you and @thallgren and want to make sure this still makes sense in light of that work.

hlindberg:
@thallgren was already satisfied with the approach as there is no overlap between what is done here and other work we have done recently wrt resource types and loading.

thallgren:
Yes, please merge this.

kylog:
Merging. This should be released in the next puppet Y release, so puppet 4.5. Props to @nbarrientos for the original PR, and to @haus for the followon work. Thanks all!!

peterhuene:
Merge commit (sadpanda)

kylog:
Ohhh yeah. I wonder if we could automate a check for that.

joshcooper:
Looks good overall, thanks everyone!

I had one follow-up comment. If an agent evaluates a feature during its run, it will [clear the rubygems cache of file paths](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/feature.rb#L81-L85), so that it can discover and load new gems that may have been installed earlier during the run. See https://tickets.puppetlabs.com/browse/PUP-1879. Each time it does this, rubygems will rescan the gem libdirs to rebuild its cache.

However, this code path is also executed on the master, which causes it to clear the cache and rebuild it each time we evaluate a feature, really for no good reason. It might be interesting to see if we get an additional perf boost on the master if we never clear the rubygems cache.

That said, the boost may be minimal because the number of features is relatively small and fixed, e.g. types and report processors loaded in the master. Also when `always_retry_plugins=false` we cache feature misses, so we only incur the penalty of rescanning gem libdirs once per feature, not once per feature evaluation. That said, it might improve JRuby initialization time?

haus:
@joshcooper would it make sense to guard the clear call with a check against always_retry_plugins/always_cache_features?

joshcooper:
@haus yeah that's what I was thinking (at least for the new setting you added), but I'm not sure if/what the perf benefit would be?

PR #4819, opened 2016-03-28
(PUP-6095) Service provider shouldn't die if an smf service is in state "degraded"

shawnferry:
(PUP-6095) Service provider shouldn't die if an smf service is in state "degraded"

Without this patch attempts to view or manage smf services fail when any service
is in a degraded state. After patching, services in a degraded state are treated in an
identical manner to those in a maintenance state.

### DIFF
--- a/lib/puppet/provider/service/smf.rb
+++ b/lib/puppet/provider/service/smf.rb
@@ -34,11 +34,12 @@ Puppet::Type.type(:service).provide :smf, :parent => :base do
   end
 
   def self.instances
-   svcs("-H").split("\n").select{|l| l !~ /^legacy_run/ }.collect do |line|
-     state,stime,fmri = line.split(/\s+/)
+   svcs("-H", "-o", "state,fmri" ).split("\n").select{|l| l !~ /^legacy_run/ }.collect do |line|
+     state,fmri = line.split(/\s+/)
      status =  case state
                when /online/; :running
                when /maintenance/; :maintenance
+               when /degraded/; :degraded
                else :stopped
                end
      new({:name => fmri, :ensure => status})
@@ -69,7 +70,7 @@ Puppet::Type.type(:service).provide :smf, :parent => :base do
   def startcmd
     self.setupservice
     case self.status
-    when :maintenance
+    when :maintenance, :degraded
       [command(:adm), :clear, @resource[:name]]
     else
       [command(:adm), :enable, "-s", @resource[:name]]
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/service/smf.rb
```diff
@@ -34,11 +34,12 @@ def setupservice
   end
 
   def self.instances
-   svcs("-H").split("\n").select{|l| l !~ /^legacy_run/ }.collect do |line|
-     state,stime,fmri = line.split(/\s+/)
```
shawnferry: Moving to an explicit list of fields allows us to drop the unused stime variable

lib/puppet/provider/service/smf.rb
```diff
@@ -71,6 +72,8 @@ def startcmd
     case self.status
     when :maintenance
```
haus: I think maintenance and degraded could be combined into the same when, as they have the same action.

```
when :maintenance, :degraded
  [command(:adm), :clear, @resource[:name]]
```

shawnferry: Agreed, I had been thinking about treating it differently.

haus:
@shawnferry Based on the travis failure, I think you'll need to update the test at spec/unit/provider/service/smf_spec.rb on line 32 to match the svcs call you've updated.

haus:
This change looks reasonable to me, I'm :+1: once tests are passing.

@shawnferry if you need a hand updating the tests, I'd be happy to help you.

shawnferry:
@haus I can't run the tests locally due to a known bug prior to puppet 4 but I think this will have taken care of it. If not I would appreciate the assistance. Thanks

haus:
@shawnferry yea, i think if you squash them all into the first commit it will pass that check.

adding a little more detail to that commit about why the change is being made would be good though.

shawnferry:
@haus into the overall commit message? I'm not sure what else to say other than...it doesn't work because degraded is un-handled.

haus:
@shawnferry yep. exactly what you added. looks good to me.

haus:
:+1: from me.

joshcooper:
@shawnferry thank you for your contribution! Can you sign our Contributor License Agreement at https://cla.puppetlabs.com?

shawnferry:
@joshcooper It's being weird for me. I should be covered by a corp cla from Oracle which I already signed last year and I just tried it again as corp or self.

PR #4821, opened 2016-03-30
(PUP-5965) Add 'new' call and function

hlindberg:
This adds the function 'new' which operates on a type. The same function can be called by directly calling a type e.g. `T()`.
Each type implements a `new_function` method that returns the 4.x function that acts as the initializer/instantiator function.
This is done to later enable obtaining its Puppet Signature and to allow that each type can specify a specific set of parameters.

The conversion to String is by far the most comprehensive, and it is implemented by delegating to a StringConverter utility - which is largely
modeled on Ruby's Kernel.format method, but catering to many specializations only available in the Puppet Language.

Simple conversions/instantiations are supposed to be simple. More comprehensive conversions should be seen as an interface to puppet's core to be used as
the foundation of more easy to use functions written in the puppet language. (As an example, it is now possible to implement sprintf in the puppet language,
and all conversions are fully specified by the new function.

### DIFF
--- a/lib/puppet/pops/types/string_converter.rb
+++ b/lib/puppet/pops/types/string_converter.rb
@@ -654,7 +654,7 @@ class StringConverter
       Kernel.format(f.orig_fmt, val.to_f)
 
     when :c
-      char = Kernel.format('%c', val)
+      char = [val].pack("U")
       char = f.alt? ? "\"#{char}\"" : char
       char = Kernel.format(f.orig_fmt.gsub('c','s'), char)
 
--- a/lib/puppet/pops/types/string_converter.rb
+++ b/lib/puppet/pops/types/string_converter.rb
@@ -248,7 +248,6 @@ class StringConverter
   #
   def initialize
     @@string_visitor   ||= Visitor.new(self, "string", 3, 3)
-    @@type_calculator = TypeCalculator.singleton()
   end
 
   DEFAULT_INDENTATION = Indentation.new(0, true, false).freeze
@@ -455,7 +454,7 @@ class StringConverter
 
     if string_formats.is_a?(String)
       # add the format given for the exact type
-      t = @@type_calculator.infer_set(value)
+      t = TypeCalculator.infer_set(value)
       string_formats = { t => string_formats }
     end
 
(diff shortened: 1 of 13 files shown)


lib/puppet/pops/types/types.rb
```diff
@@ -784,6 +916,59 @@ def merge(o)
     end
   end
 
+  # Returns a new function that produces Integer or Float value
```
thallgren: Don't think this one will produce an Integer value

thallgren:
The 'predefined_loader.rb' file is missing.

lib/puppet/pops/types/string_converter.rb
```diff
@@ -0,0 +1,923 @@
...
+  # @api public
+  def self.singleton
+    @tconv_instance ||= new
+  end
+
+  # @api private
+  #
+  def initialize
+    @@string_visitor   ||= Visitor.new(self, "string", 3, 3)
+    @@type_calculator = TypeCalculator.singleton()
```
thallgren: This is not needed. All occurrences of {{@@type_calculator.infer_set}} can be replaced with {{TypeCalculator.infer_set}} (it's implemented on the class).

lib/puppet/pops/types/string_converter.rb
```diff
@@ -0,0 +1,923 @@
...
+  end
+
+  # @api private
+  def string_PStructType(val_type, val, format_map, indentation)
+    string_PHashType(val_type, val, format_map, indentation)
+  end
+
+  # @api private
+  def string_PType(val_type, val, format_map, _)
+    f = get_format(val_type, format_map)
```
thallgren: This should be merged with the TypeFormatter. It's prepared for that and should utilize the array/hash formatting used here. Can be done as a separate JIRA.

thallgren:
This looks great. Just some minor comments (and the file that is missing).

lib/puppet/pops/types/string_converter.rb
hlindberg: Sure - I see this as a starting point. Please file a ticket for what you had in mind.

hlindberg:
Pushed fixes for problem with `to_h` and a UTF-8 space char that caused Ruby 1.9.3 to fail running `new_spec.rb`

thallgren:
The build still fails on Ruby 1.9.3. Something related to unicode. Also, picky, but perhaps you could condense the commit message of 3bed524? It seems to be the result of a big squash with lots of concatenated comments.

hlindberg:
Hm, tests ran fine on 1.9.3 locally after I fixed the other unicode problem.... looking.

PR #4824, opened 2016-03-30
(PUP-6099) add additional autorequires for file/mount interactions

mschuchard:
With this improvement, file resources will execute before their mount point is mounted and file resources will execute after their parent mount is mounted.

I need some help writing the spec tests because I am unable to fully parse how the Puppet spec API is working.  Please assist.

Also I probably could use some code criticism.  The coding style between the mount and file types and their spec tests were rather different and I tried to stick to the style within each.

### DIFF
--- a/lib/puppet/type/file.rb
+++ b/lib/puppet/type/file.rb
@@ -324,20 +324,22 @@ Puppet::Type.newtype(:file) do
     defaultto '%'
   end
 
-  # Autorequire the nearest ancestor directory found in the catalog.
-  autorequire(:file) do
-    req = []
-    path = Pathname.new(self[:path])
-    if !path.root?
-      # Start at our parent, to avoid autorequiring ourself
-      parents = path.parent.enum_for(:ascend)
-      if found = parents.find { |p| catalog.resource(:file, p.to_s) }
-        req << found.to_s
+  # Autorequire the nearest ancestor directory or mount point found in the catalog.
+  [:file => :path, :mount => :name].each do |type, property|
+    autorequire(type) do
+      req = []
+      path = Pathname.new(self[property])
+      if !path.root?
+        # Start at our parent, to avoid autorequiring ourself
+        parents = path.parent.enum_for(:ascend)
+        if found = parents.find { |p| catalog.resource(type, p.to_s) }
+          req << found.to_s
+        end
       end
+      # if the resource is a link, make sure the target is created first
+      req << self[:target] if self[:target] && type == :file
+      req
     end
-    # if the resource is a link, make sure the target is created first
-    req << self[:target] if self[:target]
-    req
   end
 
   # Autorequire the owner and group of the file.
--- a/lib/puppet/type/mount.rb
+++ b/lib/puppet/type/mount.rb
@@ -293,5 +293,8 @@ module Puppet
       dependencies[0..-2]
     end
 
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(@parameters[:name].value) }
+
   end
 end
--- a/spec/unit/type/file_spec.rb
+++ b/spec/unit/type/file_spec.rb
@@ -1278,6 +1278,15 @@ describe Puppet::Type.type(:file) do
         expect(reqs[0].target).to eq(file)
       end
 
+      it "should autorequire its parent mount point" do
+        mount = Puppet::Type.type(:mount).stubs(:defaultprovider).new(:name => File.dirname(path))
+        catalog.add_resource file
+        catalog.add_resource mount
+        reqs = file.autorequire
+        expect(reqs[0].source).to eq(mount)
+        expect(reqs[0].target).to eq(file)
+      end
+
       it "should autorequire its nearest ancestor directory" do
         dir = described_class.new(:path => File.dirname(path))
         grandparent = described_class.new(:path => File.dirname(File.dirname(path)))
@@ -1287,6 +1296,18 @@ describe Puppet::Type.type(:file) do
         reqs = file.autorequire
         expect(reqs.length).to eq(1)
         expect(reqs[0].source).to eq(dir)
+        expect(reqs[0].target).to eq(file)
+      end
+
+      it "should autorequire its nearest ancestor mount point" do
+        mount = Puppet::Type.type(:mount).stubs(:defaultprovider).new(:name => File.dirname(path))
+        grandparent = Puppet::Type.type(:mount).stubs(:defaultprovider).new(:name => File.dirname(File.dirname(path)))
+        catalog.add_resource file
+        catalog.add_resource mount
+        catalog.add_resource grandparent
+        reqs = file.autorequire
+        expect(reqs.length).to eq(1)
+        expect(reqs[0].source).to eq(mount)
         expect(reqs[0].target).to eq(file)
       end
 
--- a/spec/unit/type/mount_spec.rb
+++ b/spec/unit/type/mount_spec.rb
@@ -566,35 +566,55 @@ describe Puppet::Type.type(:mount), :unless => Puppet.features.microsoft_windows
     let(:root_mount) { create_resource("/") }
     let(:var_mount)  { create_resource("/var") }
     let(:log_mount)  { create_resource("/var/log") }
+    let(:var_file) do
+      Puppet::Type.type(:file).stubs(:defaultprovider).new(
+        :path => '/var',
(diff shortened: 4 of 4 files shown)


mschuchard:
I did a good bit more work on the spec tests.  I noticed there are no spec tests on the user and group autorequires for the file resource, so I understand I am probably foraging into unknown territory here by trying to implement cross-resource autorequire spec tests here.  If I succeed, then you guys could potentially use this pattern for the other unimplemented spec tests.

Also this PR would prompt an update to the file and mount resource type documentation.

Still would appreciate some code criticism and test assistance though.

lib/puppet/type/file.rb
```diff
@@ -324,20 +324,22 @@ def self.title_patterns
...
+      path = Pathname.new(self[property])
+      if !path.root?
+        # Start at our parent, to avoid autorequiring ourself
+        parents = path.parent.enum_for(:ascend)
+        if found = parents.find { |p| catalog.resource(type, p.to_s) }
+          req << found.to_s
+        end
       end
+      # if the resource is a link, make sure the target is created first
+      req << self[:target] if self[:target] && type == :file
```
MikaelSmith: Any explicit reason you don't require the mount happens before creating a link to the mount point?

lib/puppet/type/mount.rb
```diff
@@ -293,5 +293,8 @@ def value(name)
       dependencies[0..-2]
     end
 
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(@parameters[:name].value) }
```
MikaelSmith: We now have a autobefore method. The file type is pretty core, where-as the mount type isn't as central and may later be moved to a separate module (that would generally be part of the standard distribution, but could be removed). How about moving the dependency logic for mount <-> file to an autobefore in the mount type, so the File doesn't have to know about mounts?

MikaelSmith:
Thanks for this contribution! I can't think of any reasons at the moment why this would cause problems (/cc @branan). I'll take a look into helping with specs a bit later.

mschuchard:
Thank so much for the feedback.  I will take a look at these comments and update accordingly (I have been resetting and force pushing to keep everything in one commit).

I notice a lot of random unrelated spec tests have been failing for this changeset.  I am wondering if I have done something really bad or if I had an unlucky fork.

lib/puppet/type/mount.rb
branan: I'm +1 to moving the logic to mount, In addition to what @MikaelSmith said, I think it's worth doing in mount just from a code-cleanliness standpoint. There are actually two sides to this:

* Mount should depend on the file resource that creates the mountpoint directory
* Mount should happen before any files on that mountpoint

Keeping all that in one place seems cleaner to me than having it split between file and mount.

mschuchard:
Github is currently ignoring all of my pushes to the Puppet fork, so I reforked it and Github is still ignoring my pushes for some reason.

While I run down Github's technical difficulties, this is the diff of the newest changes:

```
diff --git a/lib/puppet/type/mount.rb b/lib/puppet/type/mount.rb
index 9e798a1..3fa7427 100644
--- a/lib/puppet/type/mount.rb
+++ b/lib/puppet/type/mount.rb
@@ -14,7 +14,11 @@ module Puppet
 
       **Autorequires:** If Puppet is managing any parents of a mount resource ---
       that is, other mount points higher up in the filesystem --- the child
-      mount will autorequire them."
+      mount will autorequire them. If Puppet is managing the file path of a
+      mount point, the mount resource will autorequire it.
+
+      **Autobefores:**  If Puppet is managing any child file paths of a mount
+      point, the mount resource will autobefore them."
 
     feature :refreshable, "The provider can remount the filesystem.",
       :methods => [:remount]
@@ -293,5 +297,16 @@ module Puppet
       dependencies[0..-2]
     end
 
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(@parameters[:name].value) }
+
+    # Autobefore the mount point's child file paths
+    autobefore(:file) do
+      dependencies = []
+      Pathname.new(@parameters[:path].value).descend do |child|
+        dependencies.push child.to_s
+      end
+      dependencies[0..-2]
+    end
   end
 end
diff --git a/spec/unit/type/mount_spec.rb b/spec/unit/type/mount_spec.rb
index 4cd94a4..0976e13 100644
--- a/spec/unit/type/mount_spec.rb
+++ b/spec/unit/type/mount_spec.rb
@@ -545,15 +545,22 @@ describe Puppet::Type.type(:mount), :unless => Puppet.features.microsoft_windows
     end
   end
 
-  describe "establishing autorequires" do
+  describe "establishing autorequires and autobefores" do
 
-    def create_resource(path)
+    def create_mount_resource(path)
       described_class.new(
         :name => path,
         :provider => providerclass.new(path)
       )
     end
 
+    def create_file_resource(path)
+      Puppet::Type.type(:file).stubs(:defaultprovider).new(
+        :path => path,
+        :provider => Puppet::Type.type(:file).stubs(:defaultprovider).new(:path => path).provider
+      )
+    end
+
     def create_catalog(*resources)
       catalog = Puppet::Resource::Catalog.new
       resources.each do |resource|
@@ -563,38 +570,71 @@ describe Puppet::Type.type(:mount), :unless => Puppet.features.microsoft_windows
       catalog
     end
 
-    let(:root_mount) { create_resource("/") }
-    let(:var_mount)  { create_resource("/var") }
-    let(:log_mount)  { create_resource("/var/log") }
+    let(:root_mount) { create_mount_resource("/") }
+    let(:var_mount)  { create_mount_resource("/var") }
+    let(:log_mount)  { create_mount_resource("/var/log") }
+    let(:var_file) { create_file_resource('/var') }
+    let(:log_file) { create_file_resource('/var/log') }
+    let(:puppet_file) { create_file_resource('/var/log/puppet') }
 
     before do
-      create_catalog(root_mount, var_mount, log_mount)
+      create_catalog(root_mount, var_mount, log_mount, var_file, log_file)
     end
 
     it "adds no autorequires for the root mount" do
       expect(root_mount.autorequire).to be_empty
     end
 
-    it "adds the parent autorequire for a mount with one parent" do
+    it "adds the parent autorequire and the file autorequire for a mount with one parent" do
       parent_relationship = var_mount.autorequire[0]
+      file_relationship = var_mount.autorequire[1]
 
-      expect(var_mount.autorequire).to have_exactly(1).item
+      expect(var_mount.autorequire).to have_exactly(2).item
 
       expect(parent_relationship.source).to eq root_mount
       expect(parent_relationship.target).to eq var_mount
+
+      expect(file_relationship.source).to eq var_file
+      expect(file_relationship.target).to eq var_mount
     end
 
-    it "adds both parent autorequires for a mount with two parents" do
+    it "adds both parent autorequires and the file autorequire for a mount with two parents" do
       grandparent_relationship = log_mount.autorequire[0]
       parent_relationship = log_mount.autorequire[1]
+      file_relationship = log_mount.autorequire[2]
 
-      expect(log_mount.autorequire).to have_exactly(2).items
+      expect(log_mount.autorequire).to have_exactly(3).items
 
       expect(grandparent_relationship.source).to eq root_mount
       expect(grandparent_relationship.target).to eq log_mount
 
       expect(parent_relationship.source).to eq var_mount
       expect(parent_relationship.target).to eq log_mount
+
+      expect(file_relationship.source).to eq log_file
+      expect(file_relationship.target).to eq log_mount
+    end
+
+    it "adds the child autobefore for a mount with one file child" do
+      child_relationship = log_mount.autobefore[0]
+
+      expect(log_mount.autobefore).to have_exactly(1).item
+
+      expect(child_relationship.source).to eq log_mount
+      expect(child_relationship.target).to eq puppet_file
+    end
+
+    it "adds both child autbefores for a mount with two file childs" do
+      child_relationship = var_mount.autobefore[0]
+      grandchild_relationship = var_mount.autobefore[1]
+
+      expect(var_mount.autobefore).to have_exactly(2).items
+
+      expect(child_relationship.source).to eq var_mount
+      expect(child_relationship.target).to eq log_file
+
+      expect(grandchild_relationship.source).to eq var_mount
+      expect(grandchild_relationship.target).to eq puppet_file
     end
   end
 end
```

I am unsure how to get Puppet to traverse unknown child paths of a mount point, unless it is working from the file resources themselves (which would be back to the autorequire). I am sure the mimicked code I put in will not actually work correctly.

Changed up the spec test for everything being in mount, but still expecting the file resources to not be generated correctly.

mschuchard:
Github broke my fork (I think they were having some technical/networking issues in the last hour) so I have to open a new PR.

PR #4826, opened 2016-03-30
(MAINT) Refactor load_custom_oid_file

fpringvaldsen:
Refactor the load_custom_oid_file method by breaking out the
file parsing into a separate method, parse_custom_oid_file.

I've opened this PR, as part of the work for [SERVER-1150](https://tickets.puppetlabs.com/browse/SERVER-1150). For that, we need the array of custom oid definitions, but the original method did not return that.

PR #4827, opened 2016-03-30
(MAINT) Refactor load_custom_oid_file

fpringvaldsen:
Refactor the load_custom_oid_file method by breaking out the
file parsing into a separate method, parse_custom_oid_file.

This is necessary so this logic can be shared by Puppet Server.

### DIFF
--- a/lib/puppet/ssl/oids.rb
+++ b/lib/puppet/ssl/oids.rb
@@ -69,7 +69,7 @@ module Puppet::SSL::Oids
     end
   end
 
-  # Parse and load custom OID mapping file that enables custom OIDs to be resolved
+  # Parse custom OID mapping file that enables custom OIDs to be resolved
   # into user-friendly names.
   #
   # @param custom_oid_file [String] File to obtain custom OIDs mapping from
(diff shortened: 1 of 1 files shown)


fpringvaldsen:
Replaces https://github.com/puppetlabs/puppet/pull/4826

ping @hlindberg

lib/puppet/ssl/oids.rb
```diff
@@ -84,7 +84,7 @@ def self.register_puppet_oids()
   #  '1.3.6.1.4.1.34380.1.2.1.2':
   #    shortname: 'myothershortname'
   #    longname: 'Other Long name'
-  def self.load_custom_oid_file(custom_oid_file, map_key='oid_mapping')
+  def self.parse_custom_oid_file(custom_oid_file, map_key='oid_mapping')
```
hlindberg: Well, this breaks API, the method is neither Ruby private nor API @private. The original method name must be retained for semver reasons.
Puppet can use the new method name, and the old can delegate to the new after having issued a deprecation warning.

In new code always add @api private to classes, or all methods to prevent not being allowed to make changes. It is allowed to make methods @api private without breaking semver.

fpringvaldsen: I think I'm a little confused by your comment. I essentially broke this into two methods. This one is called `parse_custom_oid_file`, but the `load_custom_oid_file` method below retains the same name and functionality as the previously existing `load_custom_oid_file`. Also, this whole class is marked `@private`.

hlindberg: Doh... (I am the one being confused) - please disregard everything I said :-)

joshcooper:
:+1: 

My only comment is about the target branch. If this needs to go into the next 4.4.z release, then stable is correct. If this is for 4.5, then we'd want master.

fpringvaldsen:
@joshcooper We want this in the next 4.4.z release, so I believe it's targeted at the correct branch

PR #4828, opened 2016-03-31
(PA-272) - Add support for artifactory HTTP headers for checksum validation.

heliocentric:
…s includes sha256, sha1, and md5, which take priority over content-md5 if they exist.

### DIFF
--- a/lib/puppet/util/monkey_patches.rb
+++ b/lib/puppet/util/monkey_patches.rb
@@ -122,15 +122,19 @@ if Puppet::Util::Platform.windows?
   require 'openssl'
 
   class OpenSSL::X509::Store
+    @puppet_certs_loaded = false
     alias __original_set_default_paths set_default_paths
     def set_default_paths
       # This can be removed once openssl integrates with windows
       # cert store, see https://rt.openssl.org/Ticket/Display.html?id=2158
-      Puppet::Util::Windows::RootCerts.instance.to_a.uniq.each do |x509|
-        begin
-          add_cert(x509)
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Thanks for the contribution! A few steps are required before we're able to merge a PR. Please review https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md for details, and ensure specs are passing.

lib/puppet/file_serving/http_metadata.rb
```diff
@@ -20,7 +20,21 @@ def initialize(http_response, path = '/dev/null')
       checksum = checksum.unpack("m0").first.unpack("H*").first
       @checksums[:md5] = "{md5}#{checksum}"
     end
+    # Add support for artifactory headers.
+    # This adds support for all three algorithms, sha256, sha1, and md5.
+    if checksum = http_response['X-Checksum-Md5']
```
MikaelSmith: These are non-standard, and could potentially be used for a different purpose on other websites. As such, I'm not comfortable hard-coding these headers as fallbacks.

However, this could be done as settings that can be configured in `puppet.conf`. See https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L825-L839 for an example of a setting that takes an array. My suggestion would be to introduce 3 settings - one each for custom md5, sha1, and sha256 headers - that can be an array of possible headers. Then `#initialize` would iterate through them to look for a valid checksum.

heliocentric:
I am going to pull back this pull request and make it more palatable.

PR #4832, opened 2016-04-04
(PUP-6120) Pip package provider compatibility with CLI

Yuav:
Reduce dependencies to only require pip installed.

This results in:

Removed complexity:

    Poll from PyPI API directly
    Special case proxies

Added functionality:

    pip can now properly manage pip package from pip repository


spec/unit/provider/package/pip3_spec.rb
```diff
@@ -2,282 +2,15 @@
...
-    @provider = provider_class.new(@resource)
-    @client = stub_everything('client')
-    @client.stubs(:call).with('package_releases', 'real_package').returns(["1.3", "1.2.5", "1.2.4"])
-    @client.stubs(:call).with('package_releases', 'fake_package').returns([])
-    XMLRPC::Client.stubs(:new2).returns(@client)
+  it "should inherit most things from pip provider" do
+    expect(provider_class < Puppet::Type.type(:package).provider(:pip))
   end
 
-  describe 'provider features' do
```
MikaelSmith: We should maintain these, as they're not inherited from the parent provider.

kylog:
@Yuav thanks for the contribution! Do you mind signing the CLA at https://cla.puppetlabs.com/?

kylog:
This generally looks reasonable. It makes me wonder why this code was ever going straight to the XML-RPC api.

Yuav:
CLA signed

Yuav:
I would assume the original approach is due to the fact that Pip doesn't have any official way to look up versions from PyPI. My approach is a hack, but it works

spec/unit/provider/package/pip3_spec.rb
Yuav: What's not inherited? The pip3 implementation is really small, levering almost everything from the pip provider. Why have duplicated tests?

MikaelSmith: I meant specifically features. Features don't appear to be inherited, which was fixed in https://github.com/puppetlabs/puppet/commit/45c2656d2dfe72eb640198ebc474b4f5d9332113.

Yuav:
Tests for features re-added, as per comments from MikaelSmith

Yuav:
Hmm, that Travis failure doesn't seem to be related

lib/puppet/provider/package/pip.rb
```diff
@@ -131,4 +146,33 @@ def lazy_pip(*args)
   def install_options
     join_options(@resource[:install_options])
   end
+
+  def latest_with_new_pip
+    # Less resource intensive approach for pip version 1.5.4 and above
+    execpipe ["#{self.class.pip_cmd}", "install", "#{@resource[:name]}==versionplease"] do |process|
```
MikaelSmith: Any reason you're using install, instead of something like
```
$ pip search sslyze
SSLyze (0.13.3)  - Fast and full-featured SSL scanner.
  INSTALLED: 0.13.5
  LATEST:    0.13.3
nassl (0.13.3)   - Experimental OpenSSL wrapper for Python 2.7 and SSLyze.
  INSTALLED: 0.13.5
  LATEST:    0.13.3
```

MikaelSmith: Ok, looking at how you're using it, seems like it's probably easier to parse install.

MikaelSmith:
You were right. I'm double-checking functionality, but this looks good.

PR #4833, opened 2016-04-04
(PUP-6121) Fix auto-loading of ruby types

thallgren:
This commit fixes a regression introduced in PUP-6028 where a Resource
Not Found error is triggered on ruby Puppet::Type declarations that
resides in a module.


hlindberg:
ok to merge when travis goes green

PR #4835, opened 2016-04-05
PUP-6072 Make --lastrunreport be a path-style argument

richardc:
file-type argument to being a path-type argument.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1517,8 +1517,7 @@ EOT
     },
     :lastrunreport =>  {
       :default  => "$statedir/last_run_report.yaml",
-      :type     => :file,
-      :mode     => "0640",
+      :type     => :path,
       :desc     => "Where puppet agent stores the last run report in yaml format."
     },
     :graph => {
--- a/lib/puppet/indirector/yaml.rb
+++ b/lib/puppet/indirector/yaml.rb
@@ -19,17 +19,20 @@ class Puppet::Indirector::Yaml < Puppet::Indirector::Terminus
   def save(request)
     raise ArgumentError.new("You can only save objects that respond to :name") unless request.instance.respond_to?(:name)
 
-    file = path(request.key)
+    files = path(request.key)
+    files = [files].flatten.uniq
 
-    basedir = File.dirname(file)
+    files.each do |file|
+      basedir = File.dirname(file)
 
-    # This is quite likely a bad idea, since we're not managing ownership or modes.
-    Dir.mkdir(basedir) unless Puppet::FileSystem.exist?(basedir)
+      # This is quite likely a bad idea, since we're not managing ownership or modes.
+      Dir.mkdir(basedir) unless Puppet::FileSystem.exist?(basedir)
 
-    begin
-      Puppet::Util::Yaml.dump(request.instance, file)
-    rescue TypeError => detail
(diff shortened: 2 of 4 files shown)


lib/puppet/indirector/yaml.rb
```diff
@@ -19,17 +19,20 @@ def find(request)
   def save(request)
     raise ArgumentError.new("You can only save objects that respond to :name") unless request.instance.respond_to?(:name)
 
-    file = path(request.key)
+    files = path(request.key)
```
MikaelSmith: I'd think `#find` also needs to be updated.

lib/puppet/defaults.rb
```diff
@@ -1517,8 +1517,7 @@ def self.default_diffargs
     },
     :lastrunreport =>  {
       :default  => "$statedir/last_run_report.yaml",
-      :type     => :file,
-      :mode     => "0640",
+      :type     => :path,
```
MikaelSmith: It's not clear how multiple paths should be used. Does this mean it saves to multiple locations? When reading the lastrunreport, which location should it use?

richardc: It saves to multiple locations.

From my testing, a lastrunreport is never read by puppet - that's why #find was not updated.

lib/puppet/indirector/yaml.rb
richardc: My testing shows that those operations are never called on a Indirector::Report::Yaml object, that is to say the lastrunreport is only ever involved in `#save` calls.

MikaelSmith: I'm not a fan of leaving them all broken. @joshcooper any thoughts on what we should do if it's dead code?

richardc: It's not dead code for other indirected types, but it is uncalled when the subclass of Puppet::Indirector::Yaml is Puppet::Indirector::Report::Yaml.

In order to extend for completeness, we need to figure out the semantics that should be used for each operation:

* save - case introduced here, saves multiple copies of the indirected item, as path returned an array
* find - return the contents of the first element of path? returns the contents of all the elements of path combined in some manner? what happens if one or more of the elements is absent?
* destroy - unlinks all the elements of path? what if one or more is already absent?
* search - searches in order and stops once 'found'?  loads all to combine?

MikaelSmith: If it's dead code, my thought was we should remove it and throw exceptions. However, since the default setting is a single path, it's relatively safe to leave it. I'd probably accept comments on each of the other methods mentioning they don't currently support multiple paths, but that the code is essentially unused; if anyone ever uses it they'll need to extend in the ways you mention.

joshcooper: I'm a little reluctant to make this change. It breaks the indirector model where find, etc implement CRUD operations for a single object. Also `lib/puppet/indirector/yaml.rb` is the base implementation for all yaml-based indirected models, not just reports...

Is there a reason we can't just copy the report to the other place after saving it?

Alternatively maybe a `composite` based terminus that extends terminus and contains N termini?

richardc: I reworked so it so it's more contained to the Puppet::Indirector::Report::Yaml class, and have staged that for further review as #4914.

PR #4837, opened 2016-04-05
(PUP-6125) log to console if verbose/debug enabled, and logdest is unset

pyther:
If the verbose or debug options are used, the console is added as a logging destination. However, the `Puppet::Util::Log.setup_default` method is also called. This results in output being sent to both the console and syslog on linux or eventlog on windows. Additionally, output would still go to the console even the logdest option was set.

The documentation states

> * --logdest:
  Where to send log messages. Choose between 'syslog' (the POSIX syslog
  service), 'eventlog' (the Windows Event Log), 'console', or the path to a log
  file. If debugging or verbosity is enabled, this defaults to 'console'.
  Otherwise, it defaults to 'syslog' on POSIX systems and 'eventlog' on Windows.

With this PR the following happens:
  * logs are sent to sources defined by logdest
  * logs are sent to the console when verbose or debug options are used
  * logs are sent to the default source when logdest is unspecified and verbose/debug option are not used

### DIFF
--- a/lib/puppet/application.rb
+++ b/lib/puppet/application.rb
@@ -357,13 +357,16 @@ class Application
   end
 
   def setup_logs
-    if options[:debug] || options[:verbose]
-      Puppet::Util::Log.newdestination(:console)
+    unless options[:setdest]
+      if options[:debug] || options[:verbose]
+        Puppet::Util::Log.newdestination(:console)
+      else
+        Puppet::Util::Log.setup_default
+      end
     end
 
     set_log_level
 
-    Puppet::Util::Log.setup_default unless options[:setdest]
   end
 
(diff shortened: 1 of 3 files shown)


pyther:
@puppetcla I have signed the CLA.

MikaelSmith:
It'd be nice to have the rationale in the commit message. Otherwise the change looks good, but I'm soliciting more feedback from people who might rely on the existing behavior before changing it.

pyther:
@MikaelSmith I updated the commit message

branan:
Per discussion in the ticket and in our PR triage meeting, @pyther has volunteered to try a different approach. We're closing this PR  for now.

PR #4838, opened 2016-04-05
(PUP-6099) add additional autorequires for file/mount interactions

mschuchard:
With this improvement, file resources will execute before their mount point is mounted and file resources will execute after their parent mount is mounted.

Continues #4824 

After moving all the changes to the mount type, I am unsure how to get Puppet to traverse unknown child paths of a mount point, unless it is working from the file resources themselves and ascending to find mount points. I am sure the mimicked code I put in will not actually work correctly.

Changed up the spec test for everything being in mount, but still expecting the file resources to not be generated correctly.

### DIFF
--- a/lib/puppet/type/mount.rb
+++ b/lib/puppet/type/mount.rb
@@ -14,7 +14,11 @@ module Puppet
 
       **Autorequires:** If Puppet is managing any parents of a mount resource ---
       that is, other mount points higher up in the filesystem --- the child
-      mount will autorequire them."
+      mount will autorequire them. If Puppet is managing the file path of a
+      mount point, the mount resource will autorequire it.
+
+      **Autobefores:**  If Puppet is managing any child file paths of a mount
+      point, the mount resource will autobefore them."
 
     feature :refreshable, "The provider can remount the filesystem.",
       :methods => [:remount]
@@ -293,5 +297,18 @@ module Puppet
       dependencies[0..-2]
     end
 
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(self[:name]) }
+
+    # Autobefore the mount point's child file paths
+    autobefore(:file) do
+      dependencies = []
+      file_resources = catalog.resources.select { |resource| resource.type == :file }
+      children_file_resources = file_resources.select { |resource| File.expand_path(resource[:path]) =~ %r(#{self[:name]}/.) }
+      children_file_resources.each do |child|
+        dependencies.push Pathname.new(child[:path])
+      end
+      dependencies
+    end
   end
 end
--- a/spec/unit/type/mount_spec.rb
+++ b/spec/unit/type/mount_spec.rb
@@ -545,12 +545,20 @@ describe Puppet::Type.type(:mount), :unless => Puppet.features.microsoft_windows
     end
   end
 
-  describe "establishing autorequires" do
+  describe "establishing autorequires and autobefores" do
 
-    def create_resource(path)
+    def create_mount_resource(path)
       described_class.new(
         :name => path,
         :provider => providerclass.new(path)
+      )
+    end
+
+    def create_file_resource(path)
+      file_class = Puppet::Type.type(:file)
+      file_class.new(
+        :path => path,
+        :provider => file_class.new(:path => path).provider
       )
     end
 
@@ -563,38 +571,71 @@ describe Puppet::Type.type(:mount), :unless => Puppet.features.microsoft_windows
       catalog
     end
 
-    let(:root_mount) { create_resource("/") }
-    let(:var_mount)  { create_resource("/var") }
-    let(:log_mount)  { create_resource("/var/log") }
+    let(:root_mount) { create_mount_resource("/") }
+    let(:var_mount)  { create_mount_resource("/var") }
+    let(:log_mount)  { create_mount_resource("/var/log") }
+    let(:var_file) { create_file_resource('/var') }
+    let(:log_file) { create_file_resource('/var/log') }
+    let(:puppet_file) { create_file_resource('/var/log/puppet') }
 
     before do
-      create_catalog(root_mount, var_mount, log_mount)
+      create_catalog(root_mount, var_mount, log_mount, var_file, log_file, puppet_file)
     end
 
     it "adds no autorequires for the root mount" do
       expect(root_mount.autorequire).to be_empty
     end
 
-    it "adds the parent autorequire for a mount with one parent" do
+    it "adds the parent autorequire and the file autorequire for a mount with one parent" do
       parent_relationship = var_mount.autorequire[0]
+      file_relationship = var_mount.autorequire[1]
 
-      expect(var_mount.autorequire).to have_exactly(1).item
+      expect(var_mount.autorequire).to have_exactly(2).items
 
       expect(parent_relationship.source).to eq root_mount
       expect(parent_relationship.target).to eq var_mount
+
+      expect(file_relationship.source).to eq var_file
+      expect(file_relationship.target).to eq var_mount
     end
 
-    it "adds both parent autorequires for a mount with two parents" do
+    it "adds both parent autorequires and the file autorequire for a mount with two parents" do
       grandparent_relationship = log_mount.autorequire[0]
       parent_relationship = log_mount.autorequire[1]
+      file_relationship = log_mount.autorequire[2]
 
-      expect(log_mount.autorequire).to have_exactly(2).items
+      expect(log_mount.autorequire).to have_exactly(3).items
 
       expect(grandparent_relationship.source).to eq root_mount
       expect(grandparent_relationship.target).to eq log_mount
 
       expect(parent_relationship.source).to eq var_mount
       expect(parent_relationship.target).to eq log_mount
+
+      expect(file_relationship.source).to eq log_file
+      expect(file_relationship.target).to eq log_mount
+    end
+
+    it "adds the child autobefore for a mount with one file child" do
+      child_relationship = log_mount.autobefore[0]
+
+      expect(log_mount.autobefore).to have_exactly(1).item
+
+      expect(child_relationship.source).to eq log_mount
+      expect(child_relationship.target).to eq puppet_file
+    end
+
+    it "adds both child autobefores for a mount with two file childs" do
+      child_relationship = var_mount.autobefore[0]
+      grandchild_relationship = var_mount.autobefore[1]
+
+      expect(var_mount.autobefore).to have_exactly(2).items
+
+      expect(child_relationship.source).to eq var_mount
+      expect(child_relationship.target).to eq log_file
+
+      expect(grandchild_relationship.source).to eq var_mount
+      expect(grandchild_relationship.target).to eq puppet_file
     end
   end
 end


lib/puppet/type/mount.rb
```diff
@@ -293,5 +297,16 @@ def value(name)
       dependencies[0..-2]
     end
 
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(@parameters[:name].value) }
+
+    # Autobefore the mount point's child file paths
+    autobefore(:file) do
+      dependencies = []
+      Pathname.new(@parameters[:path].value).descend do |child|
```
MikaelSmith: The spec tests are pointing out that `path` is nil (not set). Sounds like this should be using `name` instead.

mschuchard:
I was ignoring the test issues because they are arising from high-level and not low-level problems. From a general design standpoint, like I mentioned above, I have no idea how a Mount resource is supposed to know what potential File resource child paths exist. The similar implementations of these autorequires involve ascending into known parents and not theoretical children, but I followed the suggestion of moving everything into the package resource, so now I am stuck from a high-level design/algorithm perspective.

If someone can point out how a Mount resource would be able to gather information about actual File resource child directories, then the implementation should go quickly afterwards. As it is, this is pretty much dead in the water.

MikaelSmith:
It likely needs inverse behavior to File then; walk all File resources in the Catalog, and check if this mount is an ancestor of the File.

mschuchard:
Would that not be very computationally inefficient?

It could be mitigated if, instead, I immediately filtered out all file resources with paths that did not match the mount resource name and then iterated over those remaining for ancestry checking. That also may be what you were suggesting anyway.

Do you know a code snippet to grab all the file resources in the catalog? I can go hunting myself if this is unfamiliar territory.

MikaelSmith:
Hmm, it might be. You're switching from: for each file resource, traverse its ancestors and look for a matching mount resource with `parents.find {|p| catalog.resource(:file, p.to_s)}`; to: for each mount point, search for all file resources that are children of the mount point, which will end up iterating over all file resources for each mount point. So from `O(<num files> * <average path depth>)` (assuming constant-time lookup of a specific file resource) to `O(<num mounts> * <num files>)`.

That's a hard trade-off to compare; number of mount points could vary wildly.

As far as implementation, the [User type](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/user.rb#L395) has an example of using `catalog.resources`.

/cc @joshcooper @branan for other ideas

mschuchard:
Thanks for the snippet. I don't think I understand its use within user completely, because it seems to me you could just match the gid or the gid-->group without iterating. I can certainly make use of the `catalog.resources.find` though.

Regarding the efficiency tradeoff, I feel like in general there will be substantially fewer file resources than mount resources. That being said, I was comparing the efficiency of the sorting algorithms and not the file versus mount aspect. I will see what I can come up with to save time.

mschuchard:
Put some morework into this. Starting to get ugly and complicated due to all of the constraints, but at least there seems to be some kind of path forward now.

MikaelSmith:
There are certainly holes in testing; we continually try to fill them in and ensure new code doesn't introduce more holes. Some hope is definitely involved.

mschuchard:
So does this look like it should work now? What should I do about the tests?

mschuchard:
New push today. Looks like good news and bad news.

Good news seems to be that I figured out a way to do cross-resource autorequire/autobefore testing and that the autorequire is working. Bad news is that I did not do it perfectly as several other random tests seem to be failing. Also it seems that my autobefore code is not working based on the relevant test results.

Could this get passed around to other team members to suggest how to patch up the hole(s) in the way I created the file resource for testing within mount and to start creating tests for all the other cross resource autorequire tests in Puppet since they are not being tested?

lib/puppet/type/mount.rb
```diff
@@ -293,5 +297,18 @@ def value(name)
       dependencies[0..-2]
     end
 
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(@parameters[:name].value) }
+
+    # Autobefore the mount point's child file paths
+    autobefore(:file) do
+      dependencies = []
+      file_resources = catalog.resources.reject { |resource| resource.type != Puppet::Type.type(:file) }
```
joshcooper: `resource.type` will return the symbolic name for the type (specified when the type is defined in Puppet::Type.newtype(:something)). For example, `:file`, `:stage`, etc.

`Puppet::Type.type(:file)` will return the `Puppet::Type::File` class that puppet's DSL generates.

So I think you'll want to say `{ |resource| resource.type != :file }` or alternatively, `collect` resources types that equal `:file`.

lib/puppet/type/mount.rb
```diff
@@ -293,5 +297,18 @@ def value(name)
...
+      children_file_resources = file_resources.select { |resource| resource.parameters[:path].value =~ %r(#{@parameters[:name].value}/.) }
```
joshcooper: I think you can simplify `resource.parameters[:path].value` to `resource[:path]`. Calling `resource.parameters[:path]` will return the instance of `Puppet::Type::File::ParameterPath` (created via the `newparam(:path)` DSL method). 

Similarly, I think you can simplify `@parameters[:name].value` to `self[:path]`. The autorequire block is evaluated in the context of the specific `mount` resource, so `self` here is an instance of `Puppet::Type::Mount`.

lib/puppet/type/mount.rb
joshcooper: Also there is a convenient method for looking up a specific resource: `catalog.resource(:file, '/path/to/file')` but it doesn't help the case where you want all resources of a given type.

mschuchard: Good call. I modified that lambda from the one in `puppet/lib/puppet/face/catalog/select.rb` but I misinterpreted the variable type of 'type.'

lib/puppet/type/mount.rb
mschuchard: You lost me a little here. I think that is probably because I am a polyglot and tend to just look at languages sometimes as doing C++ in that language (hence why I was googling 'ruby copy constructor' yesterday).

Basically my interpretation of what you are saying is I should be accessing the instance of `Puppet::Type::File/Mount` and then the hash value of its parameter from there, instead of an instance of `Puppet::Type::File/Mount::Parameter` and accessing its instance variable from there.

As for `@parameters[:name].value` versus `self[:name]`, I was just doing a quick paste from the autorequire block for nested mounts since I know that was accurately accessing the mount name parameter. I will try changing the autorequire block I added and see what happens since it seems that part was passing my new spec tests.

mschuchard:
New push today. With the new error being reported back from Travis and what is in the end of `puppet/lib/puppet/face/catalog/select.rb`, I am wondering if `catalog.resources.each { |resource| resource.type }` is a string and not a symbol.

lib/puppet/type/mount.rb
```diff
@@ -293,5 +297,18 @@ def value(name)
       dependencies[0..-2]
     end
 
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(self[:name]) }
+
+    # Autobefore the mount point's child file paths
+    autobefore(:file) do
+      dependencies = []
+      file_resources = catalog.resources.collect { |resource| resource.type == :file }
```
MikaelSmith: I believe you want `select` instead of `collect`. Collect is creating an array of booleans (result of `resource.type == :file`).

mschuchard: You are absolutely correct. I tend to use `map` instead of `collect`, so I forgot what it does and blindly followed Josh's suggestion. I should have paid closer attention.

mschuchard:
Apologies for no new push in a while. I have been busy with other projects, including my own and those of others. I promise something new by end of day Thursday at the latest.

MikaelSmith:
I definitely understand, thanks for being so responsive.

lib/puppet/type/mount.rb
```diff
@@ -293,5 +297,18 @@ def value(name)
...
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(self[:name]) }
+
+    # Autobefore the mount point's child file paths
+    autobefore(:file) do
+      dependencies = []
+      file_resources = catalog.resources.select { |resource| resource.type == :file }
+      children_file_resources = file_resources.select { |resource| resource[:path].to_s =~ %r(#{self[:name]}/.) }
+      children_file_resources.each do |child|
+        dependencies.push Pathname.new(child[:path].to_s)
```
mschuchard: Do I need these type conversions to string or are the `resource[:parameter]` already strings?

mschuchard:
Holy crap it is working. I mean... yeah. This was a rough PR due to the new autobefore feature and the new cross-resource autodepedency spec test with my own insistence that I figure out how to do that to patch up the hole in Puppet testing.

Couple of notes:
1. Can you get the word out that it seems I figured out cross-resource autodependency testing so that it can get added to all the places in the spec tests where it is needed (i.e. user resource autorequire group resource that is a group it belongs to).

2. Why are ruby 1.9.3 and 2.0.0 part of the TravisCI if HEAD is ruby 2.1.x?

lib/puppet/type/mount.rb
joshcooper: It depends on the parameter. File `:path` should return a string, but `:links` should return a symbol, e.g. `:follow`. So in this case, I think `to_s` is not needed.

lib/puppet/type/mount.rb
joshcooper: Doh, sorry about that! `collect` and `map` are actually the same method: https://github.com/ruby/ruby/blob/f98c9a0aa63c5db16005709eff807dcad525bf57/array.c#L6098-L6101, and `select` is the different one.

lib/puppet/type/mount.rb
```diff
@@ -293,5 +297,18 @@ def value(name)
...
     end
 
+    # Autorequire the mount point's file resource
+    autorequire(:file) { Pathname.new(self[:name]) }
+
+    # Autobefore the mount point's child file paths
+    autobefore(:file) do
+      dependencies = []
+      file_resources = catalog.resources.select { |resource| resource.type == :file }
+      children_file_resources = file_resources.select { |resource| resource[:path].to_s =~ %r(#{self[:name]}/.) }
```
joshcooper: This may not match if the file `:path` isn't canonicalized, e.g. `/foo/bar/../baz`. You may be able to do `File.expand_path(resource[:path]) =~ %r(#{self[:name]}/.)`

mschuchard: Good point. Will resolve in next push.

lib/puppet/type/mount.rb
mschuchard: Thanks, will resolve in the next push.

mschuchard:
New push to resolve Josh's suggestions.

kylog:
LGTM. Last call for review!

Btw @mschuchard re your questions:
> Can you get the word out that it seems I figured out cross-resource autodependency testing so that it can get added to all the places in the spec tests where it is needed (i.e. user resource autorequire group resource that is a group it belongs to).

Do you mind filing a ticket sketching out what would be done here? 

> Why are ruby 1.9.3 and 2.0.0 part of the TravisCI if HEAD is ruby 2.1.x?

Although the puppet-agent packages include ruby 2.1, people also run puppet as a gem with a range of ruby versions.

We treat dropping a ruby version as a breaking change, so we won't drop support for older ruby versions until puppet 5. (Which, I know, seems a bit nuts since 1.9.3 is EOL, but we do have a variety of stakeholders with different operating environments.)

mschuchard:
Thanks for the clarifications.

I will fill out a ticket sketching out the additional unit test gaps that could be satisfied now.

I did not consider people running Puppet as a gem within a different ruby ABI instead of the one packaged with Puppet in its distribution packages, so that definitely makes sense. We just use the ruby packaged with since we are PE customers.

kylog:
Thanks for all the hard work on this one @mschuchard! This should be released in puppet 4.5.0.

PR #4840, opened 2016-04-06
(PUP-6124) Update Acceptance tests for huaweios

LuvCurves:


acceptance/tests/concurrency/ticket_2659_concurrent_catalog_requests.rb
```diff
@@ -87,7 +87,7 @@
           sleep $sleep_for
           url='https://#{master}:8140/puppet/v3/catalog/#{agent_cert}?environment=production'
           echo "Curling: $url"
-          env PATH='#{agent[:privatebindir]}:${PATH}' curl --tlsv1 -v -# -H 'Accept: text/pson' --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} $url
+          env PATH="#{agent[:privatebindir]}:$PATH" curl --tlsv1 -v -# -H 'Accept: text/pson' --cert #{cert_path} --key #{key_path} --cacert #{cacert_path} $url
```
joshcooper: Since we're in a HEREDOC, variable interpolation should already be occurring. Is there a reason this change is needed for cisco? I'm concerned it may break other platforms, as quoting can be tricky with beaker, net-ssh and the remote side.

joshcooper: So I think using double quotes is the correct thing to do.

I think what happened is we used single quotes resulting in `env PATH='/opt/puppetlabs/puppet/bin/:$PATH' curl ...` which corrupts the PATH when trying to execute curl. However, it has worked because we build and distribute curl in puppet-agent, and we want to use ours, so we're sure it supports tlsv1 and isn't some archiac system openssl.

The part that's strange is why does the current test fail on huaweios? Are we not building/including curl on that platform? Does it rely on other executables which are not in puppet's privatebindir?

LuvCurves: @joshcooper No curl in /opt/puppetlabs/puppet/bin on huaweios.  Also, simply having  /opt/puppetlabs/puppet/bin in the PATH first should be adequate to make sure OUR curl is used (if it's being built).

joshcooper: > No curl in /opt/puppetlabs/puppet/bin on huaweios.

Is that a puppet-agent bug? or are we intentionally excluding it on huaweios?

LuvCurves: It was intentional

joshcooper:
@LuvCurves looks good. Could you update the commit summary about why the change to double quotes is necessary?

LuvCurves:
@joshcooper Done.

joshcooper:
@LuvCurves sorry I misspoke. Could you update the body of the commit message? I'm not going to remember the details of why we made this change in the future.

LuvCurves:
@joshcooper Done.

LuvCurves:
@puppetlabs-jenkins retest this

PR #4841, opened 2016-04-06
(PUP-6128) Retain case in QualifiedReference

thallgren:
Before this commit, all qualified references were downcased immediately
in the AST model. Attempts where then made to restore the case when
reporting errors by capitalizing all segments in the name. So names
like `MyGreateResource` became `Mygreatresource`.

This commit changes the QualifiedReference so that it instead stores
the original name and will deliver the downcased name when asked to
do so. The logic is also changed in several places to avoid repeated
downcasing and capitalizing.

### DIFF
--- a/lib/puppet/pops/types/type_formatter.rb
+++ b/lib/puppet/pops/types/type_formatter.rb
@@ -280,7 +280,7 @@ class TypeFormatter
     if t.parameters.empty?
       t.name
     else
-      append_array(t.name.clone, t.parameters.map {|p| string(p) })
+      append_array(t.name, t.parameters.map {|p| string(p) })
     end
   end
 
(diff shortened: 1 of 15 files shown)


lib/puppet/pops/types/type_formatter.rb
```diff
@@ -290,18 +290,26 @@ def to_s
...
+  # that existing camel casing is lost.
+  #
   # @param qualified_name [String] the name to capitalize
   # @return [String] the capitalized name
   #
   def capitalize_segments(qualified_name)
-    segments = qualified_name.split(NAME_SEGMENT_SEPARATOR)
-    if segments.size == 1
-      qualified_name.capitalize
+    if !qualified_name.is_a?(String) || qualified_name =~ STARTS_WITH_ASCII_CAPITAL
```
hlindberg: What happens here if the string is something like "Abc::def" ? looks like it would then return "Abc::def" (which is wrong).

thallgren: That will get caught later when the string is compared to the `Patterns::CLASSREF_EXT` pattern, which is correct IMO. If someone is writing partially capitalized names, then they are doing something wrong.

thallgren: Tested. The parser doesn't let that pass. It raises an 'Illegal class reference' error.

thallgren: I should add that the reason the check is there now is that the most common case after the change, is that the name is OK already. Splitting it, just to secure that all segments are OK is rather expensive and this happens very often. And, with the parser trapping such names anyway, the current approach should be quite safe.

hlindberg: It is an API public method, it breaks the API as it will potentially not correctly capitalize a string if it starts with an initial upper case letter. Sorry.

thallgren: Hmm, ok. so let's set it @api private then. I don't think the TypeFormatter is in has been included in any release just yet.

thallgren: Or, rather it has been released, but in the release the method is actually private. So no problem making it @api private.

thallgren:
Pushed a rebased commit where TypeFormatter#capitalize_segments is marked `@api private`

PR #4842, opened 2016-04-06
(maint) Bump to rubocop 0.39

kylog:


kylog:
@joshcooper since you were kind enough to merge my other rubocop PR, could you take a look at this one? It turns out I misread some docs so the previous PR brought us up to 0.36, but they are now on 0.39, so this update us to latest.

joshcooper:
Looks good though sometime in the future we should enable the  Style/IfUnlessModifierOfIfUnless check because it catches some crazy code like:

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/ssl/host.rb#L201

```ruby
return nil unless Certificate.indirection.find("ca", :fail_on_404 => true) unless ca?
```

PR #4843, opened 2016-04-06
(PUP-6132) Emit formatted XML when using Puppet::Util::Plist

ccaviness:
Puppet::Util::Plist's dump_plist and write_plist_file methods both emit poorly-formatted XML plists by default. Thankfully, the cfpropertylist module this depends on has a :formatted option for both .to_str and .save. Setting this to true results in properly-formatted XML.

Also fix a call of Puppet.error to Puppet.err


kylog:
Looks reasonable to me.

MikaelSmith:
Thanks for the contribution!

PR #4844, opened 2016-04-07
(PUP-6131) Make NotUndef.new be transparent

hlindberg:
Before this, and attempt to create a new `NotUndef` resulted in
an error unless its type was narrowed.
NotUndef[Integer](42) worked but not
NotUndef(42)

This commit adds a new function to the Unit type. This is used by
NotUndef if it does not have a narrowed type.

### DIFF
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -450,7 +450,12 @@ class PNotUndefType < PTypeWithContainedType
   end
 
   def new_function(loader)
-    type.new_function(loader)
+    # If only NotUndef, then use Unit's null converter
+    if type.nil?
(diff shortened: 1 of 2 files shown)


lib/puppet/pops/types/types.rb
```diff
@@ -493,6 +498,19 @@ def instance?(o, guard = nil)
     true
   end
 
+  # A "null" implementation - that simply returns the given argument
+  def self.new_function(_, loader)
+    @new_function ||= Puppet::Functions.create_loaded_function(:new_unit, loader) do
+      dispatch :from_args do
```
thallgren: Seems a bit overkill to use a dispatcher here.

PR #4845, opened 2016-04-07
(PUP-6105) Avoid comparing symbol to string in terminus

hlindberg:
Before this, there was an attempt to create a symbol from an empty
string (or worse case from nil). With the removal of monkey patching
symbol to compare equal to string and introducing warnings, it is
important to clean up odd usage of string symbol comparisons.

This is now changed to deal with the error condition of nil/empty string
before converting to symbol and using the value.

This also makes a small refactoring to the variable name as the
implementation in general uses a very confusing style where some
attributes have readers but no writers this resulting in confusion over
if the use of the attribute is a local variable or a function call to
the attribute reader.

### DIFF
--- a/lib/puppet/indirector/terminus.rb
+++ b/lib/puppet/indirector/terminus.rb
@@ -66,18 +66,18 @@ class Puppet::Indirector::Terminus
       # Yay, class/instance confusion.
       subclass.terminus_type = self.name
 
-      # Our subclass is specifically associated with an indirection.
+      # This subclass is specifically associated with an indirection.
       raise("Invalid name #{longname}") unless names.length > 0
-      indirection_name = names.pop.sub(/^[A-Z]/) { |i| i.downcase }.gsub(/[A-Z]/) { |i| "_#{i.downcase}" }.intern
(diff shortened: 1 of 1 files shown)


lib/puppet/indirector/terminus.rb
```diff
@@ -66,18 +66,18 @@ def inherited(subclass)
...
       subclass.terminus_type = self.name
 
-      # Our subclass is specifically associated with an indirection.
+      # This subclass is specifically associated with an indirection.
       raise("Invalid name #{longname}") unless names.length > 0
-      indirection_name = names.pop.sub(/^[A-Z]/) { |i| i.downcase }.gsub(/[A-Z]/) { |i| "_#{i.downcase}" }.intern
+      processed_name = names.pop.sub(/^[A-Z]/) { |i| i.downcase }.gsub(/[A-Z]/) { |i| "_#{i.downcase}" }
 
-      if indirection_name == "" or indirection_name.nil?
+      if processed_name == "" || processed_name.nil?
```
thallgren: Neither `String#sub` nor `String#gsub` will never return `nil` so test for nil can be removed. Suggest `processed_name.empty?`

PR #4846, opened 2016-04-07
(PUP-6089) Do not show unit type for empty collections, and make creation of empty Array/Hash illegal

hlindberg:
This modifies string representation of empty Array and Hash to `Array[0, 0]´ and `Hash[0, 0]´ respectively.
This also makes it illegal to attempt to type an empty collection  - these are now illegal:
* `Array[Integer, 0, 0]`
* `Hash[Integer, Integer, 0, 0]`

Internally it is still allowed to create such types if the key/value types are `Unit` (since that is how the empty states are encoded).

### DIFF
--- a/lib/puppet/pops/types/type_formatter.rb
+++ b/lib/puppet/pops/types/type_formatter.rb
@@ -213,7 +213,7 @@ class TypeFormatter
   # @api private
   def string_PArrayType(t)
     if t.has_empty_range?
-      'Array[0, 0]'
+      append_array('Array', ['0', '0'])
     else
       append_array('Array', t == PArrayType::DATA ? EMPTY_ARRAY : [string(t.element_type)] + range_array_part(t.size_type))
     end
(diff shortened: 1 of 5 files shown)


lib/puppet/pops/types/type_formatter.rb
```diff
@@ -206,14 +206,26 @@ def string_PRuntimeType(t)
...
 
+  def is_empty_range?(from, to)
+    from == 0 && to == 0
+  end
+
   # @api private
   def string_PArrayType(t)
-    append_array('Array', t == PArrayType::DATA ? EMPTY_ARRAY : [string(t.element_type)] + range_array_part(t.size_type))
+    if t.has_empty_range?
+      'Array[0, 0]'
```
thallgren: Would prefer append_array('Array', ['0', '0']) to keep all array formatting in one place.

lib/puppet/pops/types/type_formatter.rb
```diff
@@ -206,14 +206,26 @@ def string_PRuntimeType(t)
...
+    else
+      append_array('Array', t == PArrayType::DATA ? EMPTY_ARRAY : [string(t.element_type)] + range_array_part(t.size_type))
+    end
   end
 
   # @api private
   def string_PHashType(t)
-    append_array('Hash', t == PHashType::DATA ? EMPTY_ARRAY : [string(t.key_type), string(t.element_type)] + range_array_part(t.size_type))
+    if t.has_empty_range?
+      'Hash[0, 0]'
```
thallgren: Would prefer append_array('Hash', ['0', '0']) to keep all array formatting in one place.

PR #4847, opened 2016-04-07
(PUP-6139) Log HTTP errors at debug level

haus:
Previously when an indirection raised an HTTPError the error was logged
at the info level. This causes a lot of noise for resources such as a
file resource with sourceselect, where several sources may 404 while
trying to find a successful source. This commit moves the message to the
debug level to avoid that problem.

### DIFF
--- a/lib/puppet/network/http/api/indirected_routes.rb
+++ b/lib/puppet/network/http/api/indirected_routes.rb
@@ -113,7 +113,7 @@ class Puppet::Network::HTTP::API::IndirectedRoutes
 
   def do_http_control_exception(response, exception)
     msg = exception.message
-    Puppet.info(msg)
+    Puppet.debug(msg)
     response.respond_with(exception.status, "text/plain", msg)
   end
 


KevinCorcoran:
I 🙈  here but would love to get @kylog or @joshcooper's opinion.  There's some [backstory in JIRA](https://tickets.puppetlabs.com/browse/PUP-6139) if you're interested.

kylog:
Hmm, I dunno. I'm concerned that this will suppress *useful* info messages for other users with real-world use cases where knowing about `HTTPError`s might be of genuine debugging value.

Thoughts?

KevinCorcoran:
It seems like more of an `info` than a `debug` to me as well.

@kylog Do you have any ideas why the reporter of https://tickets.puppetlabs.com/browse/PUP-6139 was not seeing this log message when running under passenger.  I can only suspect some funky passenger logging configuration, but that's just a WAG.

haus:
I'm inclined to close this. There doesn't seem to be a great way immediately to consolidate the disparate loglevels between puppet and puppet-server, and this info might be useful. We can open a docs ticket to note that puppet-server will be slightly chattier out of the box than puppet on passenger is.

kylog:
@KevinCorcoran he was logging at notice level rather than info: https://tickets.puppetlabs.com/browse/PUP-6139?focusedCommentId=142900&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-142900

haus:
@kylog notice is the default puppet loglevel

PR #4848, opened 2016-04-07
(PUP-6113) Don't fetch password_min_age on Solaris if it doen't exist

branan:
This check was lost in a refactor back in 2014. Both the `password` and `password_max_age` getters retained their checks for absence.


MikaelSmith:
Should these have specs to ensure they don't disappear again?

branan:
That's kind of why I broke out the "we have a shadow entry" check into its own line in password as well, even though it wasn't needed to fix this bug. I don't want to have test cases for every null check that we fix, and I think the new code structure makes it much clearer that null is a potentially expected value and is being dealt with as part of the code.

PR #4849, opened 2016-04-07
(PUP-6142) Only warn once for missing module dependencies

cprice404:
This commit changes the warning for missing module dependencies
to be less verbose.  It also adds support for the `:strict`
setting.  If `:strict` is set to `:warning`, then we only log
the missing module dependency once (per ruby instance).  If `:strict`
is set to `:error`, then we raise a LoaderError.  If `:strict`
is set to `:off`, then we don't log anything.

### DIFF
--- a/lib/puppet/pops/loaders.rb
+++ b/lib/puppet/pops/loaders.rb
@@ -306,9 +306,18 @@ class Loaders
 
     def create_loader_with_only_dependencies_visible(from_module_data)
       if from_module_data.unmet_dependencies?
-        Puppet.warning("ModuleLoader: module '#{from_module_data.name}' has unresolved dependencies"+
-          " - it will only see those that are resolved."+
-          " Use 'puppet module list --tree' to see information about modules")
+        if Puppet[:strict] != :off
+          msg = "ModuleLoader: module '#{from_module_data.name}' has unresolved dependencies"+
+              " - it will only see those that are resolved."+
+              " Use 'puppet module list --tree' to see information about modules"
+          if Puppet[:strict] == :error
+            raise LoaderError.new(msg)
+          elsif Puppet[:strict] == :warning
+            Puppet.warn_once(:unresolved_module_dependencies,
+                             "unresolved_dependencies_for_module_#{from_module_data.name}",
+                             msg)
+          end
+        end
       end
       dependency_loaders = from_module_data.dependency_names.collect { |name| @index[name].public_loader }
       Loader::DependencyLoader.new(from_module_data.public_loader, from_module_data.name, dependency_loaders)
--- a/lib/puppet/pops/loaders.rb
+++ b/lib/puppet/pops/loaders.rb
@@ -310,9 +310,10 @@ class Loaders
           msg = "ModuleLoader: module '#{from_module_data.name}' has unresolved dependencies"+
               " - it will only see those that are resolved."+
               " Use 'puppet module list --tree' to see information about modules"
-          if Puppet[:strict] == :error
-            raise LoaderError.new(msg)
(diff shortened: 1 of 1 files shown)


cprice404:
I've tested this by hand, and it doesn't break any existing spec tests, but I wasn't sure how to add a spec test for this, or whether it was worthwhile to do so.  Setting up some modules to be in a state where they'd exercise this code seems like it might be tricky for a spec test; if it seems worthwhile and if there is any prior art I can look at about how to mock out some modules, I'm happy to try to add a spec test.

lib/puppet/pops/loaders.rb
```diff
@@ -306,9 +306,18 @@ def create_loader_with_all_modules_visible(from_module_data)
 
     def create_loader_with_only_dependencies_visible(from_module_data)
       if from_module_data.unmet_dependencies?
-        Puppet.warning("ModuleLoader: module '#{from_module_data.name}' has unresolved dependencies"+
-          " - it will only see those that are resolved."+
-          " Use 'puppet module list --tree' to see information about modules")
+        if Puppet[:strict] != :off
+          msg = "ModuleLoader: module '#{from_module_data.name}' has unresolved dependencies"+
+              " - it will only see those that are resolved."+
+              " Use 'puppet module list --tree' to see information about modules"
```
hlindberg: You don't need the '+' operator in Ruby, it will concatenate strings for you in the parsing phase. With + it will instead copy/concatenate at runtime.

cprice404: OK - those were already there, but I can remove them if you like.

lib/puppet/pops/loaders.rb
```diff
@@ -306,9 +306,18 @@ def create_loader_with_all_modules_visible(from_module_data)
...
+          if Puppet[:strict] == :error
+            raise LoaderError.new(msg)
+          elsif Puppet[:strict] == :warning
+            Puppet.warn_once(:unresolved_module_dependencies,
+                             "unresolved_dependencies_for_module_#{from_module_data.name}",
+                             msg)
+          end
+        end
```
hlindberg: I would do something like
```
if Puppet[:strict] != :off 
  msg = " "
        " "
        " "
  case Puppet[:strict]
  when :error
  when :warning
  end
```

hlindberg:
Ping @er0ck - want an acceptance test for this?

lib/puppet/pops/loaders.rb
hlindberg: nah...

hlindberg:
I am fine with the code as it stands. Not sure how often that logic is exercised - i.e. if there are lots of cases where it is exercising the Puppet[:strict] == :warning path it is worth changing the + to parser based concat (due to intermediate extra objects - not sure how smart the Ruby parser is).

lib/puppet/pops/loaders.rb
cprice404: I just tried this without the plus signs and it does not include the additional lines in the error message.

cprice404:
I added a commit to port it over to use a `case` statement.  However, as I mentioned above, I tried removing the plus symbols and the result was that only the first string of the three showed up in the error messages.

hlindberg:
@cprice404 sorry, a \ is needed at the end of lines to make it concat

cprice404:
@hlindberg ah, I see.  Pushed another commit that gets rid of the concatenation operations.

er0ck:
i don't see any unit tests on this, but it seems relatively medium risk (annoying to the user).  we can do it later.  i'll add a case for it to our tracker.

PR #4851, opened 2016-04-07
(MAINT) Disable 'add_el_extras' for Beaker

geoffnichols:
Prior to this commit, we install and enable
an EPEL repo config during puppet acceptance
testing. However, we don't currently appear
to install any packages from EPEL during
testing.

To reduce potential occurrence of transient
failures in CI (due to issues with repo metadata,
etc), this commit sets 'add_el_extras' to
'false' so that Beaker doesn't set up the EPEL
repos on an SUT.


geoffnichols:
Closing this to retarget against stable.

PR #4852, opened 2016-04-07
(PUP-6140) Acceptance test for init.d services when systemd is default

MikaelSmith:
An acceptance test for managing init.d services on systems where the
systemd provider is the default. This doesn't work without PUP-2744,
which ensures the fallback on RedHat systems is the RedHat provider
- which works - instead of the init provider - which doesn't on RedHat.

It also helps verify that managing SysV init scripts continues to work
after a fix for PUP-5296/PUP-5825, where enabled status was reported
incorrectly.

[skip ci]

### DIFF
--- a/acceptance/tests/resource/service/init_on_systemd.rb
+++ b/acceptance/tests/resource/service/init_on_systemd.rb
@@ -1,0 +1,179 @@
+test_name 'SysV on default Systemd Service Provider Validation' do
+
+  confine :to, :platform => /el-|centos|fedora/ do |h|
+    on h, 'which systemctl', :acceptable_exit_codes => [0, 1]
+    stdout =~ /systemctl/
+  end
+
+  require 'puppet/acceptance/service_utils'
+  extend Puppet::Acceptance::ServiceUtils
+
+  svc = 'puppetize'
+  initd_location = "/etc/init.d/#{svc}"
+  pidfile = "/var/run/#{svc}.pid"
+
+  # Some scripts don't have status command.
+  def initd_file(svc, pidfile, initd_location, status)
+    initd = <<INITD
+#!/bin/bash
+# #{svc} daemon
+# chkconfig: 2345 20 80
+# description: #{svc} daemon
+
+DESC="#{svc} daemon"
+PIDFILE=#{pidfile}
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This should be extended to the level of coverage the launchd provider has. @whopper is working on some cleanup to support that.

whopper:
👍

MikaelSmith:
We should add Debian testing to this as well.

MikaelSmith:
@er0ck this is the start of the test. Looking at expanding it to all EL and Debian platforms, and sharing implementation with the launchd acceptance test.

There's a little overlap with service_enable_linux, but that test is covering the default providers. This ensures that, where the system default doesn't apply, the fallback provider works correctly.

er0ck:
definitely some benefit in sharing code, obviously, but lots of benefit for keeping the actual test isolated so we can quickly tell what breaks when the test fails (subsequent asserts in a test get skipped).  trade-offs here constantly in cognitive load of understanding the test coverage when looking in the acceptance dir, and creating lots of isolated, independent, slow tests.

acceptance/tests/resource/service/init_on_systemd.rb
```diff
@@ -0,0 +1,188 @@
+test_name 'SysV on default Systemd Service Provider Validation'
```
er0ck: nit: test_name should have a block/do-end

acceptance/tests/resource/service/init_on_systemd.rb
```diff
@@ -0,0 +1,188 @@
...
+    assert_match(/Service\[puppetize\]\/ensure: ensure changed 'running' to 'stopped'/, stdout,
+                 'Service ensure change failed')
+  end
+  on agent, puppet_resource('service', 'puppetize') do
+    assert_match(/ensure => 'stopped'/, stdout, 'Service ensure reporting failed')
+    assert_match(/enable => 'true'/, stdout, 'Service enable reporting failed')
+  end
+  # TODO: verify service is not running, apply is idempotent
+
+  step 'Create initd script with status command'
```
er0ck: nit: step should also have a block/do-end

PR #4854, opened 2016-04-07
(PUP-6129) Add new function support to type alias

hlindberg:
Before this, it was not possible to create a new instance when using a
type alias. Now, calls to new delegate to the aliased type.


thallgren:
AppVeyor apparently failed but the Details link is gone (or just not visible to me).

hlindberg:
WUT...

hlindberg:
I rebased to trigger new checks

PR #4856, opened 2016-04-08
(maint) add provider name to 'provider must have feature' error message

fessyfoo:
knowing which provider was selected can be helpful in figuring out
why you're getting the 'Provider must have features...' message.


kylog:
Hi @fessyfoo, thanks for the contribution! This seems reasonable. 

Minor nit: can you amend the commit message to prepend "(maint) " to the summary line? That should turn travis green, and will later appease our release script (which looks for commit messages following a certain pattern). Thanks!

PR #4857, opened 2016-04-08
(PUP-6137) Change Object Type from anonymous to named

thallgren:
This PR adds one commit to change the Object type from being anonymous to being named, and one commit that leverages this to make feature related errors more descriptive.

### DIFF
--- a/lib/puppet/pops/loader/type_definition_instantiator.rb
+++ b/lib/puppet/pops/loader/type_definition_instantiator.rb
@@ -42,14 +42,36 @@ class TypeDefinitionInstantiator
     # the loader is known - hence this mechanism
     private_loader = loader.private_loader
     Adapters::LoaderAdapter.adapt(type_definition).loader = private_loader
-
-    Types::PTypeAliasType.new(name, type_definition.type_expr)
+    create_type(type_definition, loader)
   end
 
   def self.create_from_model(type_definition, loader)
-    name = type_definition.name
-    [Loader::TypedName.new(:type, name.downcase), Types::PTypeAliasType.new(name, type_definition.type_expr)]
+    typed_name = Loader::TypedName.new(:type, type_definition.name.downcase)
+    type = create_type(type_definition, loader)
+    loader.set_entry(
+      typed_name,
+      type,
+      Adapters::SourcePosAdapter.adapt(type_definition).to_uri)
+    type
   end
+
+  def self.create_type(type_definition, loader)
+    type_expr = type_definition.type_expr
+    name = type_definition.name
+    if object_definition?(type_expr)
+      # No need for an alias. The Object type itself will receive the name instead
+      i12n_hash_expr = type_expr.keys.empty? ? nil : type_expr.keys[0]
+      Types::PObjectType.new(name, i12n_hash_expr)
+    else
+      Types::PTypeAliasType.new(name, type_expr)
+    end
+  end
+  private_class_method :create_type
+
+  def self.object_definition?(te)
+    te.is_a?(Model::AccessExpression) && (left = te.left_expr).is_a?(Model::QualifiedReference) && left.cased_value == 'Object'
+  end
+  private_class_method :object_definition?
 end
 end
 end
--- a/spec/unit/pops/types/p_object_type_spec.rb
+++ b/spec/unit/pops/types/p_object_type_spec.rb
@@ -28,7 +28,7 @@ describe 'The Object Type' do
         }
       OBJECT
       expect { parse_object('MyObject', obj) }.to raise_error(TypeAssertionError,
-        /attribute 'a' had wrong type, expected a Type value, got Integer/)
+        /attribute MyObject\[a\] had wrong type, expected a Type value, got Integer/)
     end
 
     it 'raises an error if the type is missing' do
@@ -48,7 +48,7 @@ describe 'The Object Type' do
         }
       OBJECT
       expect { parse_object('MyObject', obj) }.to raise_error(TypeAssertionError,
-        /attribute 'a' value had wrong type, expected an Integer value, got String/)
+        /attribute MyObject\[a\] value had wrong type, expected an Integer value, got String/)
     end
 
     it 'raises an error if the kind is invalid' do
@@ -98,7 +98,7 @@ describe 'The Object Type' do
           a => Integer
         }
       OBJECT
-      expect { tp['a'].value }.to raise_error(Puppet::Error, /attribute 'a' has no value/)
+      expect { tp['a'].value }.to raise_error(Puppet::Error, 'attribute MyObject[a] has no value')
     end
 
     context 'that are constants' do
@@ -125,7 +125,7 @@ describe 'The Object Type' do
           }
         OBJECT
         expect { parse_object('MyObject', obj) }.to raise_error(Puppet::ParseError,
-          /attribute 'a' of kind 'constant' requires a value/)
+          "attribute MyObject[a] of kind 'constant' requires a value")
       end
 
       it 'raises an error when final => false' do
@@ -139,7 +139,7 @@ describe 'The Object Type' do
           }
         OBJECT
         expect { parse_object('MyObject', obj) }.to raise_error(Puppet::ParseError,
-          /attribute 'a' of kind 'constant' cannot be combined with final => false/)
+          "attribute MyObject[a] of kind 'constant' cannot be combined with final => false")
       end
     end
   end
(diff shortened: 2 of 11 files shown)


lib/puppet/pops/loader/type_definition_instantiator.rb
```diff
@@ -42,14 +42,61 @@ def self.create(loader, typed_name, source_ref, pp_code_string)
...
+
+  def self.create_type(type_definition, loader)
+    type_expr = type_definition.type_expr
     name = type_definition.name
-    [Loader::TypedName.new(:type, name.downcase), Types::PTypeAliasType.new(name, type_definition.type_expr)]
+    if object_definition?(type_expr)
+      i12n_hash_expr = type_expr.keys[0]
+      defined_name = object_name(i12n_hash_expr)
+      if defined_name.nil? || defined_name == name
+        # No need for an alias. The Object type itself will be receive the name instead
```
hlindberg: picky: typo in comment

lib/puppet/pops/loader/type_definition_instantiator.rb
```diff
@@ -42,14 +42,61 @@ def self.create(loader, typed_name, source_ref, pp_code_string)
...
+        Types::PObjectType.new(name, i12n_hash_expr)
+      else
+        # Object type has a name that is different from the alias, it must be bound under that name as well
+        type = Types::PObjectType.new(defined_name, i12n_hash_expr)
```
hlindberg: Does this (dual names) imply that $a = Object[{name => foo}] will result in type 'foo' being bound to that Object? (I don't think it should).

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -203,9 +199,9 @@ def parse_object(name, body_string)
           a => { type => Callable, override => true }
         }
       OBJECT
-      expect_object('MyObject', parent)
+      parse_object('MyObject', parent)
       expect { parse_object('MyDerivedObject', obj) }.to raise_error(Puppet::ParseError,
-        "function 'a' overrides inherited attribute")
+        'function MyDerivedObject[a] attempts to override attribute MyObject[a]')
```
hlindberg: great

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -462,7 +458,7 @@ def parse_object(name, body_string)
         equality => [a,b]
       OBJECT
       expect { parse_object('MyObject', obj) }.to raise_error(Puppet::ParseError,
-        "equality is referencing function 'b'")
+        "MyObject equality is referencing function MyObject[b]")
```
hlindberg: think it needs to say why that is wrong "function reference not allowed in equality" or somesuch.

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -448,7 +444,7 @@ def parse_object(name, body_string)
         equality => [a,b]
       OBJECT
       expect { parse_object('MyObject', obj) }.to raise_error(Puppet::ParseError,
-        "equality is referencing constant attribute 'b'")
+        "MyObject equality is referencing constant attribute MyObject[b]")
```
hlindberg: think it needs to say why that is wrong "reference to constant not allowed in equality" or somesuch.

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -490,21 +486,21 @@ def parse_object(name, body_string)
...
 
   it 'raises an error when initialization hash contains invalid keys' do
-    expect_object('MyObject', <<-OBJECT)
+    obj = <<-OBJECT
       attribrutes => {
         a => Integer
       }
     OBJECT
-    expect { parser.parse('MyObject', loader) }.to raise_error(TypeAssertionError, /object initializer had wrong type, extraneous key 'attribrutes'/)
+    expect { parse_object('MyObject', obj) }.to raise_error(TypeAssertionError, /object initializer had wrong type, extraneous key 'attribrutes'/)
```
hlindberg: Many users are not native english speakers, can we simplify "extraneous"? "unrecognized key", "invalid key" or something like that?

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -490,21 +486,21 @@ def parse_object(name, body_string)
...
 
   it 'raises an error when attribute contains invalid keys' do
-    expect_object('MyObject', <<-OBJECT)
+    obj = <<-OBJECT
       attributes => {
         a => { type => Integer, knid => constant }
       }
     OBJECT
-    expect { parser.parse('MyObject', loader) }.to raise_error(TypeAssertionError, /initializer for attribute 'a' had wrong type, extraneous key 'knid'/)
+    expect { parse_object('MyObject', obj) }.to raise_error(TypeAssertionError, /initializer for attribute MyObject\[a\] had wrong type, extraneous key 'knid'/)
```
hlindberg: "extraneous"

spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -687,6 +675,35 @@ def parse_object(name, body_string)
       expect(eval_and_collect_notices(code)).to eql(['true'])
     end
 
+    it 'a named object is added to the loader irrespective of alias declaration' do
+      code = <<-CODE
+      $x = Object[{ name => 'MyFirstObject' }]
+      notice($x == MyFirstObject)
+      CODE
```
hlindberg: I think this is wrong - just because we have an instance of ObjectType does not mean that it is loaded/imported. I would expect the reference on line 681 to fail with undefined "MyFirstObject". The only way to get a name is to bind an object to a type with `type = ...` or when the object type is in a type set. All other instances may have a name, or be anonymous. They are useful when doing type algebra.

The rationale for this is that it must be possible to juggle ObjectType instances without having them pollute the name space.

thallgren: If that is wrong, then it is also wrong that this bind the object to two names.
```
type Foo = Object[{name => 'Bar'}]
```
I thought we agreed that it should?

spec/unit/pops/types/p_object_type_spec.rb
thallgren: Ok, we can improve the TypeMismatchDescriber to use different wording. That's not related to this PR though.

spec/unit/pops/types/p_object_type_spec.rb
hlindberg: I thought about this for a while now.
I think it must be a deliberate act to bind an object type to a type name (or for a type set, if you want to import all of them). I think that is a more important principle than the other things we talked about.
That also means, that it is strange if binding to one name also binds to another (you may not want the object bound to that name).

I think that using a name => in an object time is really only useful when using the type in matches - you want to match on type name. And you would probably never bind those (you cannot actually because you cannot say `type Foo = $x`, which only leaves the example you showed above. There you would simply leave the `Bar` out - why name it something other than what you are binding to? If you want both you can always do:
```
type Foo = Object[{...}]
type Bar = Foo
```

spec/unit/pops/types/p_object_type_spec.rb
hlindberg: Fine if it is a (maint) commit in this PR

spec/unit/pops/types/p_object_type_spec.rb
thallgren: I'm fine with changing so that no binding is made by the name used in the hash, but we need to sort out what should happen with both names are given:
```
type Foo = Object[{name => 'Bar'}]
```
Should this be allowed at all? And if so, which name should be used in error messages? When do names conflict? It will get mildly confusing to write:
```
type Foo = Object[{
  name => 'Bar',
  attributes => {
    'id' => { type => Integer, final => true }
  }
}]
```
and then, in some other file:
```
type Bar = Object[{ parent => Foo
  attributes => {
    'id' => { type => Integer[0], override => true }
  }
}]
```
and have the parser tell you `Bar[id]` attempts to override a final attribute in `Bar`. With the code as it stands right now, this can never happen since the loader would raise an 'attempt to redefine Bar' exception.

thallgren:
Pushed a rebased commit where:
1. A type is made known to the loader only by using an alias construct
2. An alias construct overrides a name declared in the initialization hash

I also pushed a (maint) commit chaning 'extraneous' to 'unrecognized' and an extra commit for more elaborated 'equality' errors.

PR #4858, opened 2016-04-08
(PUP-6064) Fix problem with given undef and optional parameter in EPP

thallgren:
A given argument with an undef value didn't make it all the way to
the parameter validation. An error was therefore issued that the
the parameter was missing. This commit ensures that undef values will
be propagated all the way but not override a given default.


hlindberg:
ok to merge when travis/app-veyor goes green

PR #4861, opened 2016-04-09

jearls:
is a hash of key=>value pairs; keys must be non-empty strings, and
values must either be `undef` or non-empty strings.  `undef` values are
translated to ruby `nil` values before being passed to the provider.

The `yum` package provider is updated to provide the
`:settable_environment` feature.

`environment` parameter.

### DIFF
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -9,7 +9,7 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
   These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
   or an array where each element is either a string or a hash."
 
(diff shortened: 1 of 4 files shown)


jearls:
This is a much-worked-over first attempt at providing an `environment` parameter and its associated `settable_environment` feature for the `package` type.

I'm not sure if the implementation is consistent with Puppetlabs style.  In particular, I wanted the use of the `environment` parameter to fail if the provider does not support it, but the default action (when using the `:required_feature` option to `newparam`) is to silently _(except for a debugging message)_ ignore the `environment` parameter if the provider does not support the `settable_environment` feature.  Instead of using `:required_feature`, I therefore implemented my own failure in the `validation` block if the provider does not satisfy the `settable_environment` feature requirement.

jearls:
Bad push, will create new PR after it's fixed

PR #4862, opened 2016-04-09

jearls:
is a hash of key=>value pairs; keys must be non-empty strings, and
values must either be `undef` or non-empty strings.  `undef` values are
translated to ruby `nil` values before being passed to the provider.

The `yum` package provider is updated to provide the
`:settable_environment` feature.

`environment` parameter.

### DIFF
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -25,14 +25,6 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
 
   defaultfor :osfamily => :redhat
 
-  def execute(*args)
-    if @resource and @resource[:environment]
-      Puppet::Util::Execution.execute(*args, :custom_environment => @resource[:environment])
-    else
-      Puppet::Util::Execution.execute(*args)
-    end
-  end
-
   def self.prefetch(packages)
     raise Puppet::Error, "The yum provider can only be used as root" if Process.euid != 0
     super
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -9,7 +9,7 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
   These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
   or an array where each element is either a string or a hash."
 
-  has_feature :install_options, :versionable, :virtual_packages
+  has_feature :install_options, :versionable, :virtual_packages, :settable_environment
 
   commands :cmd => "yum", :rpm => "rpm"
 
@@ -24,6 +24,14 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
   end
 
   defaultfor :osfamily => :redhat
+
+  def execute(*args)
+    if @resource and @resource[:environment]
+      Puppet::Util::Execution.execute(*args, :custom_environment => @resource[:environment])
+    else
+      Puppet::Util::Execution.execute(*args)
+    end
+  end
 
   def self.prefetch(packages)
     raise Puppet::Error, "The yum provider can only be used as root" if Process.euid != 0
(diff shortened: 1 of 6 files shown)


jearls:
This is a much-worked-over first attempt at providing an `environment` parameter and its associated `settable_environment` feature for the `package` type.

I'm not sure if the implementation is consistent with Puppetlabs style.  In particular, I wanted the use of the `environment` parameter to fail if the provider does not support it, but the default action (when using the `:required_feature` option to `newparam`) is to silently _(except for a debugging message)_ ignore the `environment` parameter if the provider does not support the `settable_environment` feature.  Instead of using `:required_feature`, I therefore implemented my own failure in the `validation` block if the provider does not satisfy the `settable_environment` feature requirement.

MikaelSmith:
This looks pretty usable. The original ticket (PUP-1526) specifically called out Debian; did you have plans to expand this work to more providers?

MikaelSmith:
It might be useful to implement an abstraction for executing within the specified environment, maybe a method in the type. That might simplify some of the changes to each provider.

lib/puppet/provider/package/yum.rb
```diff
@@ -25,6 +25,14 @@
 
   defaultfor :osfamily => :redhat
 
+  def execute(*args)
```
MikaelSmith: Yeah, maybe this would be useful in the type so we can use it in multiple providers. Or in a common base-class for the providers.

jearls:
Unfortunately, I have no experience at all with anything besides RHEL, and so do not feel qualified to extend this solution to other providers.

lib/puppet/provider/package/yum.rb
ferventcoder: 👍

jearls: I'm looking into this.  To provide this capability in a more general sense is going to require some more wide-spread modifications.  Specifically, I'll need to make modifications to the base provider class to allow providers to modify how the custom commands are executed.  I'm not sure enough about my ruby knowledge to be positive I'm doing this correctly, so I may ask for help after my next commit. :)

MikaelSmith: That could be more elegant, but I was only suggesting that a helper like `#execute` be moved to a shared location so all package providers can make use of it. Maybe rename it to something like `#execute_in_environment`.

jearls:
I will continue working on this to get it to not fail.  I am not able to run any of the validations on my own vm, so I'm forced to use travis for it, but my future changes will be in a sub-branch until I can get it to go clean, at which point I'll merge back into this PR.

ahpook:
Nice work @jearls - just catching up on ticket traffic and I'm excited to see this is rolling along. Please feel free to join a community PR triage if you can make it - we keep track at voxpupuli/community-triage

jearls:
@ahpook sure, when do they happen and how do I join?

jearls:
@ahpook I just checked the puppet-dev google group and see that the triage meetings happen tuesday mornings.  Unfortunately, I can't make it during business hours, so I'll have to decline your offer.

jearls:
Etiquette question:  Is rebasing an open PR frowned upon?  I've got this code working now, but would like to rebase it all back into a single commit, rather than a series of fix-this/fix-that commits.

jearls:
never mind.  i'm actually going to close this PR and resubmit, since I'm going to rework the `package` environment parameter to match the same parameter in `exec`.

kylog:
@jearls thanks for digging into this! 

To answer the etiquette question, rebasing and force-pushing PRs is not frowned upon on this repo; people do it all the time. (The only downside is that comments on the pre-force-push commits are hidden on github, but it's easy to unhide them.) We often do a mix of tacking on fix-this/fix-that commits along with force-pushes to restructure the work into more readable/digestible/coherent commits, all depending on the circumstances. No etiquette worries one way or the other.

And re the triage session, if we end up wanting more back-and-forth on the issue, we are also happy to drop into IRC (or now, slack) to chat through ideas.

Hope that all helps, and thanks again for tackling this long-standing ticket!

PR #4864, opened 2016-04-11
(PUP-6145) Prevent self recursion in TypeAlias assignable? method

thallgren:
A `PTypeAliasType` with a resolved type that is a `PVariantType` where
one of the variants points directly the alias, would cause an endless
recursion since the recursion check resided in the `#_assignable?`
method. That method was never called due to special handling of type
alias and assignable in `PAnyType`. This commit fixes this by letting
the `PTypeAliasType` override the `#assignable?` method.

The commit also deals with a problem that occurred when calculating
whether or not a type was self recursive. If the computation involved
yet another alias that was not yet resolved, then it would yield a false
negative. Only the other type would be considered recursive (first type
was resolved when it resolves). This commit ensures that contained
aliases recompute their self-recursive state when a self recursion has
been detected.

When the above bug was fixed, it revealed a flaw in how `#instance?`
was computed for cases where self recursion was detected. Fixing that
required the addition of the API private method `#really_instance?`which
detects if a positive outcome stems from traversing a real type or just
a sequence of type aliases and variants.

NOTE!
When merging to master, since some functionality was moved, the
following call must be added at the very end of the method
`PTypeAliasType#when_self_recursion`.
```
@resolved_type.check_self_recursion(self) if @self_recursion
```

### DIFF
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -64,7 +64,7 @@ class PAnyType < TypedModelObject
   # @api public
   def assignable?(o, guard = nil)
     case o
-      when Class
+    when Class
       # Safe to call _assignable directly since a Class never is a Unit or Variant
       _assignable?(TypeCalculator.singleton.type(o), guard)
     when PUnitType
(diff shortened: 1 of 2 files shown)


hlindberg:
There were a couple of merge conflict when merging up to master - thomas to fix conflict and push to both stable and master directly (rather than having stable not be cleanly mergeable for a while).

PR #4868, opened 2016-04-11
(PUP-6133) Make struct functional as hash key

thallgren:
It's currently not possible to use a Struct type as a Hash key. This
is because the `PStructType#eql?` method is using array == which in
turn relies on the `#==` method of its elements. The `PStructElement`
class implements `#eql?` but not `#==`. This commit changes this so
that both methods can be used.


hlindberg:
Good to merge

PR #4869, opened 2016-04-11
(PUP-5928) Add pattern matching to Runtime type

thallgren:
This commit extends the Runtime type so that it can describe a pattern
match between a Puppet type name and a Runtime type name.

### DIFF
--- a/lib/puppet/pops/types/p_runtime_type.rb
+++ b/lib/puppet/pops/types/p_runtime_type.rb
@@ -47,10 +47,12 @@ class PRuntimeType < PAnyType
     iterable?(guard) ? PIterableType.new(self) : nil
   end
 
+  # @api private
   def runtime_type_name
     @name_or_pattern.is_a?(String) ? @name_or_pattern : nil
   end
 
+  # @api private
   def from_puppet_name(puppet_name)
     if @name_or_pattern.is_a?(Array)
       substituted = puppet_name.sub(*@name_or_pattern)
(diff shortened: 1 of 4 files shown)


lib/puppet/pops/types/p_runtime_type.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+    end
+    false
+  end
+
+  def iterable_type(guard = nil)
+    iterable?(guard) ? PIterableType.new(self) : nil
+  end
+
+  def runtime_type_name
```
hlindberg: do we want this @api public at this point?

lib/puppet/pops/types/p_runtime_type.rb
```diff
@@ -0,0 +1,90 @@
...
+    @name_or_pattern.is_a?(String) ? @name_or_pattern : nil
+  end
+
+  def from_puppet_name(puppet_name)
```
hlindberg: same here @api

hlindberg:
Looks fine - only small concern that two new methods on PRuntimeType should probably be marked as @api private to make it easier to modify them if needed as we continue with the implementation of the rest of pcore/serializatin

lib/puppet/pops/types/p_runtime_type.rb
thallgren: It was public previously (attr_reader). Not sure what the rules are for marking it @api private

hlindberg:
Suggest adding the two `©api private` to indicate that we don't want people to use those methods. Nothing really happens to those that already do use them, so it is harmless to add `@api private`.

PR #4870, opened 2016-04-11
(PUP-6136) new-feature/pup-6136/newparam-downgrade

mschuchard:
With this new package parameter, package downgrades can be disabled.

Tackled a couple of the easier ones. Was hoping for some preliminary feedback on whether this is the route I should be taking stylistically. I am planning on adding this for all the providers where it would not involve adding huge chunks of code.

### DIFF
--- a/lib/puppet/provider/package/rpm.rb
+++ b/lib/puppet/provider/package/rpm.rb
@@ -120,6 +120,8 @@ Puppet::Type.type(:package).provide :rpm, :source => :rpm, :parent => Puppet::Pr
 
     # RPM gets upset if you try to install an already installed package
     return if @resource.should(:ensure) == version || (@resource.should(:ensure) == :latest && version == latest)
+    # Do now downgrade the package if :dowgrade = false
+    return if @resource.should(:ensure) <= version && !@resource.downgrade?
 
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/package/rpm.rb
```diff
@@ -120,6 +120,8 @@ def install
 
     # RPM gets upset if you try to install an already installed package
     return if @resource.should(:ensure) == version || (@resource.should(:ensure) == :latest && version == latest)
+    # Do now downgrade the package if :dowgrade = false
```
jyaworski: Should be:

# Do not downgrade the package if :downgrade == false

mschuchard: Yeah I noticed that too right after the PR. I fixed it post `git reset HEAD~` and the typo will be gone in the next force push.

mschuchard:
Apologies for not focusing on this in a long time. I have been focusing on PUP-6099 when I have time to spare for Puppet PR and am still uncertain if I want to finish this given the conversation with Kylo.

mschuchard:
Closing this PR. Consult PUP-6136 in JIRA for discussion.

PR #4875, opened 2016-04-12
(PUP-3827) Return errors matching schema from indirected routes

justinstoller:
Prior to this the indirected routes would intercept any issues that
arose in processing the request and format them into a plain text
response with the proper error code.

All other routes pass issues up to the generic route handler which will
format them according to a documented schema.

This moves the indirected routes handler from never raising to always
raising on error, similar to other route handlers. This allows the
indirected routes to share the error handling code with other routes
and provides consistently formatted errors throughout the rest api.

Running the acceptance suite locally (via `SHA=$(git rev-parse HEAD) TEST_TARGET=ubuntu1404-64a bundle exec rake ci:test:git` I get:
```
Errored Tests Cases:
            Test Case tests/direct_puppet/static_catalog_env_control.rb reported: #<Beaker::Host::CommandFailure: Host 'y383sdp13cn6x5a.delivery.puppetlabs.net' exited with 1 runn
ing:
           mv /etc/puppetserver/conf.d/puppetserver.conf.bak /etc/puppetserver/conf.d/puppetserver.conf
          Last 10 lines of output were:
                mv: cannot stat ‘/etc/puppetserver/conf.d/puppetserver.conf.bak’: No such file or directory>
            Test Case tests/language/exported_resources.rb reported: #<Beaker::Host::CommandFailure: Host 'y383sdp13cn6x5a.delivery.puppetlabs.net' exited with 1 running:
           ln -sf /tmp/exported_resources_9p4y75k3 /etc/puppetlabs/code/environments/exported_resources_9p4y75k3
          Last 10 lines of output were:
                ln: failed to create symbolic link ‘/etc/puppetlabs/code/environments/exported_resources_9p4y75k3’: No such file or directory>
```

Both failures above are during the setup phases of the tests and have to do with files, I think, would normally be managed by packages. I reran the suite against the commit in master I branched from and it also failed with the same two failures. So I think those are an issue with either the current code base, or running from git.

### DIFF
--- a/lib/puppet/network/http/api/indirected_routes.rb
+++ b/lib/puppet/network/http/api/indirected_routes.rb
@@ -52,7 +52,6 @@ class Puppet::Network::HTTP::API::IndirectedRoutes
   def uri2indirection(http_method, uri, params)
     # the first field is always nil because of the leading slash,
     indirection_type, version, indirection_name, key = uri.split("/", 5)[1..-1]
-
     url_prefix = "/#{indirection_type}/#{version}"
     environment = params.delete(:environment)
 
--- a/lib/puppet/network/http/api/indirected_routes.rb
+++ b/lib/puppet/network/http/api/indirected_routes.rb
@@ -47,15 +47,12 @@ class Puppet::Network::HTTP::API::IndirectedRoutes
     Puppet.override(:trusted_information => trusted) do
       send("do_#{method}", indirection, key, params, request, response)
     end
-  rescue Puppet::Network::HTTP::Error::HTTPError => e
-    return do_http_control_exception(response, e)
-  rescue StandardError => e
-    return do_exception(response, e)
   end
 
   def uri2indirection(http_method, uri, params)
     # the first field is always nil because of the leading slash,
     indirection_type, version, indirection_name, key = uri.split("/", 5)[1..-1]
+
     url_prefix = "/#{indirection_type}/#{version}"
     environment = params.delete(:environment)
 
@@ -110,24 +107,6 @@ class Puppet::Network::HTTP::API::IndirectedRoutes
   end
 
   private
-
-  def do_http_control_exception(response, exception)
-    msg = exception.message
-    Puppet.info(msg)
-    response.respond_with(exception.status, "text/plain", msg)
-  end
-
-  def do_exception(response, exception, status=400)
-    if exception.is_a?(Puppet::Network::AuthorizationError)
-      # make sure we return the correct status code
-      # for authorization issues
-      status = 403 if status == 400
-    end
-
-    Puppet.log_exception(exception)
-
-    response.respond_with(status, "text/plain", exception.to_s)
-  end
 
   # Execute our find.
   def do_find(indirection, key, params, request, response)
--- a/lib/puppet/network/http/api/indirected_routes.rb
+++ b/lib/puppet/network/http/api/indirected_routes.rb
@@ -88,7 +88,11 @@ class Puppet::Network::HTTP::API::IndirectedRoutes
       params[:environment] = configured_environment
     end
 
-    check_authorization(method, "#{url_prefix}/#{indirection_name}/#{key}", params)
+    begin
+      check_authorization(method, "#{url_prefix}/#{indirection_name}/#{key}", params)
+    rescue Puppet::Network::AuthorizationError => e
+      raise Puppet::Network::HTTP::Error::HTTPNotAuthorizedError.new(e.message)
+    end
 
     if configured_environment.nil?
       raise ArgumentError, "Could not find environment '#{environment}'"
--- a/spec/integration/network/http/api/indirected_routes_spec.rb
+++ b/spec/integration/network/http/api/indirected_routes_spec.rb
@@ -54,14 +54,14 @@ describe Puppet::Network::HTTP::API::IndirectedRoutes do
       end
     end
 
-    describe "an error from IndirectedRoutes" do
-      require 'rack/mock'
+    describe "an error from IndirectedRoutes", :if => Puppet.features.rack? do
+      require 'rack/mock' if Puppet.features.rack?
 
       let(:handler) { Puppet::Network::HTTP::RackREST.new }
-      let(:response) { Rack::Response.new }
 
       describe "returns json" do
         it "when a standard error" do
+          response = Rack::Response.new
           request = Rack::Request.new(
             Rack::MockRequest.env_for("/puppet/v3/invalid-indirector"))
 
@@ -73,6 +73,7 @@ describe Puppet::Network::HTTP::API::IndirectedRoutes do
           expect(resp["issue_kind"]).to be_a_kind_of(String)
         end
         it "when a server error" do
+          response = Rack::Response.new
           request = Rack::Request.new(
             Rack::MockRequest.env_for("/puppet/v3/unknown_indirector"))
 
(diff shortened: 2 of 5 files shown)


haus:
@justinstoller looks like adding the `rack/test` require may have broken windows: 

```
C:/projects/puppet/spec/integration/network/http/api/indirected_routes_spec.rb:58:in `require': cannot load such file -- rack/mock (LoadError)
	from C:/projects/puppet/spec/integration/network/http/api/indirected_routes_spec.rb:58:in `block (3 levels) in <top (required)>'
```

justinstoller:
Looking into it `rack/mock` is part of `rack` proper, but `rack` itself is part of the `extra` [gem group](https://github.com/puppetlabs/puppet/blob/4.4.1/Gemfile#L64) and isn't [installed during windows testing](https://github.com/puppetlabs/puppet/blob/4.4.1/appveyor.yml#L8). It seems like there maybe a `Puppet.feature.rack?` method that I could give to RSpec's `describe` to limit it running. I'll look into it further...

justinstoller:
I've gated the tests in question from running if rack is not available. I've similarly gated requiring `rack/mock` and moved the require to the top of the file, per the prior art I could find on testing with rack in the puppet specs.

haus:
:+1: from me.

spec/integration/network/http/api/indirected_routes_spec.rb
```diff
@@ -2,6 +2,8 @@
 require 'spec_helper'
 require 'puppet/network/http'
 require 'puppet/network/http/api/indirected_routes'
+require 'rack/mock' if Puppet.features.rack?
```
KevinCorcoran: The `if` in this line is surprising to my 👀  although I claim no real familiarity here.  Would it be more conventional to push the `require 'rack/mock'` down into the test case which actually needs it?

spec/integration/network/http/api/indirected_routes_spec.rb
```diff
@@ -52,5 +54,36 @@
...
+          response = Rack::Response.new
+          request = Rack::Request.new(
+            Rack::MockRequest.env_for("/puppet/v3/unknown_indirector"))
+
+          handler.process(request, response)
+
+          expect(response.header["Content-Type"]).to match(/json/)
+          resp = JSON.parse(response.body.first)
+          expect(resp["message"]).to match(/Could not find indirection/)
+          expect(resp["issue_kind"]).to be_a_kind_of(String)
```
KevinCorcoran: 👍 on this test, exactly what I *expected* (pun intended) to find, based on the ticket description.

KevinCorcoran:
🙈  Just had the one comment about the `require`

spec/integration/network/http/api/indirected_routes_spec.rb
haus: not really. it's more idiomatic to keep them all together.

although it might make sense to move the conditional require to the end of the list.

justinstoller: Yeah, I had initially put the require down in the describe in https://github.com/puppetlabs/puppet/pull/4875/commits/363d36e232a5be40646d1412bbc958281677a581 but that failed because it either needed to be conditionally required or the require needed to move into the `it` block. I moved it up there based on reviewing things like [the rack specs](https://github.com/puppetlabs/puppet/blob/f50edf20fd963a357f98924efa6000e5ac238e76/spec/unit/network/http/rack_spec.rb#L3). Though I have no strong preference to where it goes.

justinstoller:
@kylog @joshcooper @whopper @branan 

Folks from the Server team generally want the work in this PR, but would love a review from Client incase we missed anything concerning.

spec/integration/network/http/api/indirected_routes_spec.rb
KevinCorcoran: As discussed in meatspace, I think is fine and not really worth changing at this point.

lib/puppet/network/http/api/indirected_routes.rb
```diff
@@ -111,24 +107,6 @@ def uri2indirection(http_method, uri, params)
...
-    Puppet.info(msg)
-    response.respond_with(exception.status, "text/plain", msg)
-  end
-
-  def do_exception(response, exception, status=400)
-    if exception.is_a?(Puppet::Network::AuthorizationError)
-      # make sure we return the correct status code
-      # for authorization issues
-      status = 403 if status == 400
-    end
```
joshcooper: This logic still needs to be handled somewhere, otherwise we'll return status 400 for authorization issues?

justinstoller: Ahh, good catch. In my head when reading it I substituted `Puppet::Network::HTTP::Error::HTTPNotAuthorizedError` (which will behave as expected with the error handling at the higher level).

Implementation-wise, since we have an [v3Auth class](https://github.com/puppetlabs/puppet/blob/98d22be27447bcb34c33aa4a44610ab8768390f5/lib/puppet/network/http/api/master/v3/authorization.rb) that's calling the auth provider and wrapping any `AuthorizationError`s in the appropriate HTTPError for our environment(s) [routes](https://github.com/puppetlabs/puppet/blob/98d22be27447bcb34c33aa4a44610ab8768390f5/lib/puppet/network/http/api/master/v3.rb#L13-L16). But, IIUC, we need to do the URI munging in `uri2indirection` before we can call our authorization for indirected routes. I would, at this point in my thinking, wrap [the call to `check_authorization`](https://github.com/puppetlabs/puppet/blob/98d22be27447bcb34c33aa4a44610ab8768390f5/lib/puppet/network/http/api/indirected_routes.rb#L95) in a similar catch and wrap block to what the v3Auth class is doing. That goes with the generic HTTPHandler only really caring about proper HTTPErrors and that we're already doing a lot of check and raise work in `uri2indirection` (but I don't know if the fact we're already doing a bunch of catch-and-raise work necessarily makes it a good thing to add more...).

Does that approach sound sane, or do you think that handling general Auth errors is something the Handler should know about, or is there better route that I haven't investigated (eg. I could see wanting to refactor `uri2indirection` into a set of smaller methods but I don't know if it's something that would be valuable within the scope of this work).

justinstoller: Updated in 2ebdb75537f3101faab141ed6c0b307c43605ff0

justinstoller: Lemme know if you think this approach is sufficient. I've updated the tests to explicitly test for an auth failure and that such a failure will raise the appropriate `HTTPError` though I'm happy to go through a deeper refactor, or a different approach if you think its warranted.

branan:
When I try to curl a catalog with a mismatched cert I'm getting an appropriate 403 here, so I think the change you made in 2ebdb75 is working as expected. I think I'm (thumbsup) on this.

```
branan$ curl -I --cert $SSLDIR/certs/localhost.pem --key $SSLDIR/private_keys/localhost.pem --cacert $SSLDIR/certs/ca.pem -H 'Accept: pson' https://localhost:8140/puppet/v3/catalog/pacifica\?environment\=production
HTTP/1.1 403 Forbidden 
X-Puppet-Version: 4.4.2
Content-Type: application/json
Server: WEBrick/1.3.1 (Ruby/2.2.4/2015-12-16) OpenSSL/1.0.2g
Date: Tue, 03 May 2016 21:32:06 GMT
Content-Length: 164
Connection: Keep-Alive
```

justinstoller:
nbsu4jn



The regular expression relaxing doesn't fix some of the tests, which are still failing deeper within test helper methods. It shouldn't be too hard to fix these on Monday. But doing further surgery on a Friday evening is not a practical path forward. Reverting in PR #4948 until we can work on this Monday.

PR #4877, opened 2016-04-13
(PUP-6168) Fix problem with Runtime3Loader double updates

thallgren:
The Runtime3Loader calls on the known_resource_types of the current
environment to find a type. That call might ultimately result in that
the loader is being updated with the queried entry. When the call
returns, the loader must therefore check if it has the entry before
it tries to store it. This commit adds that check.


hlindberg:
Is it possible to unit test this?

hlindberg:
I am going to merge this as we have failing acceptance tests. Would be nice to have this covered by a unit test though.

PR #4878, opened 2016-04-13
(PUP-6167) Fix problem default Tuple having max size of zero

thallgren:
A default tuple has no size, but when compared it creates a range based
on the size of it's contained types. The default tuple has an empty list
of types which resulted in the range 0,0.

This commit changes this so that the computed range is 0 - infinity when
there are no types.

### DIFF
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -1448,7 +1448,7 @@ class PTupleType < PAnyType
   def size_range
     if @size_type.nil?
       types_size = @types.size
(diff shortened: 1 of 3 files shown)


hlindberg:
It is all rather strange - a tuple without specification of what the elements must be... isn't that the same as `Array`? except that it is `Array[Any, 0, default]` instead of `Array[Data, 0, default]`. I am fine with this though even if using this wide Tuple is kind of meaningless.

It makes more sense though than Tuple[0, 0] as the default, sp merging

PR #4879, opened 2016-04-13
(PUP-5588) Update monkey patch to detect duplicate certificates correctly

glennsarti:
The .uniq method used to remove duplicates, was not actually removing duplicates
due to the way the .hash function works on the Certificate object. Even for
certificates with the same content, the .hash return value was different.  This
commit uses the .to_der method as the basis for comparing certificate objects.
DER was used as opposed to other methods as it generates smaller strings
(compared to PEM) and is exposed easily in the Certificate object.

Paired with Ethan Brown (@iristyle)


joshcooper:
👍

Iristyle:
Related to #4469

PR #4880, opened 2016-04-13
Revert "(PUP-5737) Backport test fix from a pe-puppet fork"

kylog:
This reverts commit e623253792810533ba7df72c204b3fcf76851d98.

Note that the backported test fix actually breaks puppet testing because of
PUP-4755. The only impetus to backport was to minimize branch drift, but that
is unlikely to happen given that there is no active develoment on 3.x.

[skip ci]


hlindberg:
+1 when tests go green

kylog:
Green.

PR #4881, opened 2016-04-13

whopper:

list and run as well.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -273,6 +273,32 @@ def config
   ENV['BEAKER_HOSTS']
 end
 
+
+def get_test_sample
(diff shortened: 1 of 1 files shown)


whopper:
Tested and timed against Debian 8 and Windows 2012 R2 with the following results:

Debian: `7.31s user 0.76s system 0% cpu 22:08.98 total`
Windows: `46.66s user 1.24s system 3% cpu 22:36.47 total`

whopper:
Also, any opinions on the naming of this task? `quick`, 'aio_per_commit`, `subset` ?

kylog:
* I'd target stable rather than master so we get this exposed to both branches.
* Re naming, `quick` works for me.
* Is there a matching PR to ci-job-configs?

PR #4883, opened 2016-04-14
(PUP-5929) Add mapping between runtime classes and puppet types

thallgren:
This PR adds the ability to create mappings between runtime types (here, only ruby since the runtime is ruby) and types in the Puppet Type system. Mappings are declared using an alias like expression where the left hand side is a Runtime type (either a single type or array with regex + replacement) and the right hand side is either a Puppet type or an array with regex + replacement.

A mapping is registered in string form in the new class `ImplementationRegistry`. A loader (the effective loader at the place where the mapping occurs) is also part of the mapping. This means that the singleton registry will be able to produce a Type from a Runtime (or vice versa) from any place in the code. This is necessary since no loader is present in the majority of cases. The `TypeCalculator` in particular, have no access to a loader instance and it must be able to produce types from a mapping when type inference is performed.


lib/puppet/pops/types/class_loader.rb
```diff
@@ -103,7 +101,7 @@ def self.provide_from_name_path(name, name_path)
   end
 
   def self.find_class(name_path)
-    name_path.reduce(Object) do |ns, name|
+    name_path.inject(Object) do |ns, name|
```
hlindberg: Always prefer `reduce` over `inject` (we use map/reduce as the paradigm instead of collect/inject).

lib/puppet/parser/ast/pops_bridge.rb
```diff
@@ -273,11 +273,15 @@ def instantiate_TypeAlias(type_alias, modname)
...
     # This is for 4x evaluator/loader
     #
-    def instantiate_TypeMapping(type_alias, modname)
-      # TODO: Register the type mapping
+    def instantiate_TypeMapping(type_mapping, modname)
+      loader = Puppet::Pops::Loaders.find_loader(modname)
+      tf = Puppet::Pops::Types::TypeParser.new
+      lhs = tf.interpret(type_mapping.type_expr, loader)
+      rhs = tf.interpret_any(type_mapping.mapping_expr, loader)
+      Puppet::Pops::Types::ImplementationRegistry.singleton.register_type_mapping(lhs, rhs, loader)
```
hlindberg: This raises the question what the lifecycle of that singleton is.

lib/puppet/pops/types/implementation_registry.rb
```diff
@@ -0,0 +1,143 @@
...
+      # Register all known type implementations. The Type itself does not follow the standard naming convention
+      register_implementation('Type', 'Puppet::Pops::Types::PType', static_loader)
+
+      # The rest of the types all follow a standard naming convention
+      register_implementation_regexp([/^([A-Z]\w*)$/, 'Puppet::Pops::Types::P\1Type'], [/^Puppet::Pops::Types::P(\w+)Type$/, '\1'], static_loader)
+    end
+
+    # Register a bidirectional type mapping.
+    #
+    # @override register_type_mapping(runtime_type, puppet_type)
```
hlindberg: You probably meant "@overload" (IIRC, "@override" is proposed but not implemented yet in yard + this is not an override).

lib/puppet/pops/types/implementation_registry.rb
```diff
@@ -0,0 +1,143 @@
...
+    #   @param runtime_type [PRuntimeType] type that represents the runtime module or class to map to a puppet type
+    #   @param puppet_type [PAnyType] type that will be mapped to the runtime module or class
+    #   @param loader [Loader::Loader] the loader to use when resolving names
+    # @override register_type_mapping(runtime_type, pattern_replacement)
```
hlindberg: overload

lib/puppet/parser/ast/pops_bridge.rb
thallgren: Agree. This needs to change.

PR #4885, opened 2016-04-14

whopper:

list and run as well.

[skip ci]

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -273,6 +273,32 @@ def config
   ENV['BEAKER_HOSTS']
 end
 
+
+def get_test_sample
+  # This set represents a reasonable sample of puppet acceptance tests,
+  # covering a wide range of features and code susceptible to regressions.
+  tests = [ 'tests/direct_puppet/cached_catalog_remediate_local_drift.rb',
+            'tests/resource/file/content_attribute.rb',
+            'tests/environment/environment_scenario-default.rb',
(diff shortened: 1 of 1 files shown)


whopper:
@kylog retargeted version of https://github.com/puppetlabs/puppet/pull/4881.

There's no ci-job-configs PR as of yet. Working on that now.

acceptance/Rakefile
```diff
@@ -273,6 +273,32 @@ def config
...
+            'tests/parser_functions/calling_all_functions.rb',
+            'tests/ticket_4622_filebucket_diff_test.rb',
+            'tests/pluginsync/4420_pluginfacts_should_be_resolvable_on_agent.rb',
+            'tests/ssl/puppet_cert_generate_and_autosign.rb',
+            'tests/resource/service/puppet_mcollective_service_management.rb'
+          ]
+
+  # Add any tests modified within the last two weeks to the list
+  modified = `git log --name-only --pretty="format:" --since 2.weeks ./tests`
+  tests += modified.split("\n").reject { |s| s.empty? }.collect { |s| s.sub('acceptance/', '')}
```
MikaelSmith: Probably worth checking if this behaves well on Windows. Although we likely won't be running the Rake task from there.

MikaelSmith: Looks like it'll be fine, msysgit uses forward-slashes.

MikaelSmith:
👍

PR #4886, opened 2016-04-14
(PUP-6176) Fix error when a resource reference is appointing undef

thallgren:
PUP-6028 fixed a bug in how the resource type and title was extracted
where the case when both where `nil` was undetected. This revealed
another bug in the `RelationshipValidator` where it attempts to create
a resource from a parameter with the value `nil`.

This commit ensures that no attempt is made to create a resource from
`nil` during parameter validation.


hlindberg:
good to merge when travis goes green

haus:
:+1: looks good to me. thanks @thallgren !

hlindberg:
There was a timeout when running the Ruby 2.0.0 tests. Respinning that job in travis.

hlindberg:
Green...

PR #4887, opened 2016-04-14
(PUP-5353) Don't try to disable static systemd services

whopper:
Static systemd services are defined as services that cannot
be enabled or disabled and are usually dependencies or tasks
of other services  Previously, Puppet would erroneously attempt
to disable these services, which is a no-op on the system. Upon
doing so, Puppet was fooled into believing that the service was
being successfully disabled in every run.

This commit updates the systemd provider to not attempt to disable
static services, and log a debugging message when a manifest
attempts to.

In addition, a new 'cached_enabled?' method is added which prevents
the provider from unnecessarily shelling out to systemctl multiple
times.

Thanks to Michael Smith for the research and part of the implementation
for this fix!

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -36,12 +36,21 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
     return []
   end
 
-  def disable
-    output = systemctl(:disable, @resource[:name])
-  rescue Puppet::ExecutionFailure
-    raise Puppet::Error, "Could not disable #{self.name}: #{output}", $!.backtrace
+  # This helper ensures that the enable state cache is always reset
+  # after a systemctl enable operation. A particular service state is not guaranteed
+  # after such an operation, so the cache must be emptied to prevent inconsistencies
+  # in the provider's believed state of the service and the actual state.
+  # @param action [String,Symbol] One of 'enable', 'disable', 'mask' or 'unmask'
+  def systemctl_change_enable(action)
+    output = systemctl(action, @resource[:name])
+  rescue
+    raise Puppet::Error, "Could not #{action} #{self.name}: #{output}", $!.backtrace
   ensure
     @cached_enabled = nil
+  end
+
+  def disable
+    systemctl_change_enable(:disable)
   end
 
   def get_start_link_count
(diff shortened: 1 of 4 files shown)


MikaelSmith:
The less intrusive alternative would be to make it behave like the Redhat provider, and just treat it as enabled. However, these changes are still isolated to a single type, so I'm not too worried about it.

lib/puppet/provider/service/systemd.rb
```diff
@@ -40,6 +40,14 @@ def disable
     output = systemctl(:disable, @resource[:name])
   rescue Puppet::ExecutionFailure
     raise Puppet::Error, "Could not disable #{self.name}: #{output}", $!.backtrace
+  ensure
+    @cached_enabled = nil
+  end
+
+  def disable_static
+    Puppet.debug("Unable to enable or disable static service #{@resource[:name]}")
+    @cached_enabled = 'static'
```
MikaelSmith: This seems like it should be unnecessary, any reason you're setting it?

whopper: Ah, actually just forgot to remove that.

lib/puppet/type/service.rb
```diff
@@ -73,6 +73,18 @@ def retrieve
         provider.enabled?
       end
 
+      # This only makes sense on systemd systems. Static services cannot be enabled
+      # or disabled.
+      def insync?(current)
+        if provider.respond_to?(:cached_enabled?) && provider.cached_enabled? == 'static'
+          if provider.respond_to?(:disable_static)
+           return provider.disable_static
```
whopper: Would it be better to return true here after the `disable_static` call, or continue returning true in `disable_static` itself?

whopper:
@MikaelSmith cleaned this up a bit and I think it's ready to go! Running through service acceptance again.

(I realized there was no need for a separate `disable_static` method in the systemd provider, as we can just log our debug method in the service type and declare the service to be insync before hitting the provider)

lib/puppet/type/service.rb
```diff
@@ -73,6 +73,17 @@ def retrieve
         provider.enabled?
       end
 
+      # This only makes sense on systemd systems. Static services cannot be enabled
+      # or disabled manually.
+      def insync?(current)
+        if provider.respond_to?(:cached_enabled?) && provider.cached_enabled? == 'static'
```
MikaelSmith: This could be abstracted to something, like `unenforcable?`. However, I don't see a strong need for it yet.

MikaelSmith:
👍

whopper:
`tests/resource/service` tests passed in Debian 8, RedHat 7 and RedHat 6.

lib/puppet/provider/service/systemd.rb
```diff
@@ -40,6 +40,8 @@ def disable
     output = systemctl(:disable, @resource[:name])
   rescue Puppet::ExecutionFailure
     raise Puppet::Error, "Could not disable #{self.name}: #{output}", $!.backtrace
+  ensure
+    @cached_enabled = nil
```
MikaelSmith: It might be simpler to pull this out into a helper for calls to systemctl. I wrote it, but it makes me a little nervous.

lib/puppet/provider/service/systemd.rb
```diff
@@ -36,10 +36,21 @@ def self.instances
     return []
   end
 
+  # This helper ensures that the enable state cache is always reset
+  # after a systemctl enable operation. A particular service state is not guaranteed
+  # after such an operation, so the cache must be emptied to prevent inconsistencies
+  # in the provider's believed state of the service and the actual state.
+  # @param action [String,Symbol] One of 'enable', 'disable', 'mask' or 'unmask'
+  def systemctl_change_enable(action)
```
whopper: @MikaelSmith abstracted `enable` operations into a helper - is this what you had in mind? The service tests all still pass.

MikaelSmith: Yeah, this actually looks much better.

PR #4889, opened 2016-04-15
(PUP 6091) Align Puppet Lookup, key and interpolation handling, with Hiera 3.0

thallgren:
This PR provides the commits necessary to align the Puppet lookup command with Hiera 3.0 with respect to how dotted keys and interpolations are handled. Details can be found in the commit comments.


lib/puppet/data_providers/hiera_interpolate.rb
```diff
@@ -1,161 +1,12 @@
...
-      segments.map! do |segment|
-        segment.strip!
-        segment.start_with?('"') || segment.start_with?("'") ? segment[1..-2] : segment
-      end
-    else
-      raise yield('Syntax error')
-    end
+    Puppet::deprecation_warning(
+      'Puppet::DataProviders::HieraInterpolate#qualified_lookup is deprecated, use Puppet::Pops::Lookup::SubLookup#sub_lookup')
+    sub_lookup('<unknown key>', Puppet::Pops::Lookup::Invocation.new(nil), segments, value)
```
hlindberg: Deprecation warnings are not liked by users. It needs to be conditional on the Puppet[:strict] setting (:off, :warning, :error), and when set to :warning, only do it once using the `warn_once` method. 
When warning, a user needs to know what to change - it is probably the implementation of a hiera backend that causes this, and thus ideally, the logged entry should point there (rather than just logging the key being looked up (maybe the deprecation_warning handles that ?)

hlindberg: i.e. I worry that some users could fill their logs with this deprecation...

thallgren: Please note that this deprecation will only affect fairly special early adopters that a) adopted 4.x data providers, b) decided to write their own Hiera-like provider, and c), decided to reuse the functionality exposed with the #qualified_lookup function. I wouldn't be too surprised if the actual number of users that has done that is zero. Personally, I'm convinced that if any user has done that, they will appreciate the warning, especially given the context of the change because their specialized provider will henceforth not be able to a) report the full key of a subkey that misses, or b)  participate in the explanation support.

thallgren:
I added use of `Puppet[:strict]`. Message is the same though, since I don't see how it could point to a third party backend without also providing a stack trace. A user can get that by running with strict set to error and in combination with --trace.

PR #4891, opened 2016-04-15
(PUP-6025) Update GetFileAttributesW calls in Windows

glennsarti:
The calls to GetFileAttributesW were added do detect symlinks/reparse points
in Windows files systems.  However the way these calls were implemented has
caused a large amount of Disk IO and costly error raising within ruby which is
just ignored.  This commit updates the following;

- Updates the exist? method to use the GetFileAttributesW to both detect if
  an item exists and whether it is a symlink.  This reduces Disk IO.  Also fixed
  unknown bug whereby the exist? method did not walk symlink chains greater than
  a depth of two (sym -> sym -> file)

- Updates the get_attributes helper method to optional not raise errors and pass
  back the raw result from GetFileAttributesW.  This is useful in cases like
  file existance checks as error messages are not required.

- Updates the symlink? method to not catch/raise errors if the target path
  does not exist and instead just return false

### DIFF
--- a/lib/puppet/file_system/windows.rb
+++ b/lib/puppet/file_system/windows.rb
@@ -4,21 +4,7 @@ require 'puppet/util/windows'
 class Puppet::FileSystem::Windows < Puppet::FileSystem::Posix
 
   def exist?(path)
-    if ! Puppet.features.manages_symlinks?
-      return ::File.exist?(path)
-    end
-
-    path = path.to_str if path.respond_to?(:to_str) # support WatchedFile
-    path = path.to_s # support String and Pathname
-
-    begin
-      if Puppet::Util::Windows::File.symlink?(path)
-        path = Puppet::Util::Windows::File.readlink(path)
-      end
-      ! Puppet::Util::Windows::File.stat(path).nil?
-    rescue # generally INVALID_HANDLE_VALUE which means 'file not found'
-      false
-    end
+    return Puppet::Util::Windows::File.exist?(path)
   end
 
   def symlink(path, dest, options = {})
--- a/lib/puppet/util/windows/file.rb
+++ b/lib/puppet/util/windows/file.rb
@@ -89,12 +89,42 @@ module Puppet::Util::Windows::File
   end
   module_function :symlink
 
+
+  def exist?(path)
+    path = path.to_str if path.respond_to?(:to_str) # support WatchedFile
+    path = path.to_s # support String and Pathname
+
+    seen_paths = []
+    loop do
+      # return false if this path has been seen before.  This is protection against circular symlinks
+      return false if seen_paths.include?(path.downcase)
+
+      result = get_attributes(path,false)
+
+      # return false for path not found
+      return false if result == INVALID_FILE_ATTRIBUTES
+
+      # return true if path exists and it's not a symlink
+      # Other file attributes are ignored. https://msdn.microsoft.com/en-us/library/windows/desktop/gg258117(v=vs.85).aspx
+      return true if (result & FILE_ATTRIBUTE_REPARSE_POINT) != FILE_ATTRIBUTE_REPARSE_POINT
+
+      # walk the symlink and try again...
+      seen_paths << path.downcase
+      path = readlink(path)
+    end
+  end
+  module_function :exist?
+
+
   INVALID_FILE_ATTRIBUTES = 0xFFFFFFFF #define INVALID_FILE_ATTRIBUTES (DWORD (-1))
 
-  def get_attributes(file_name)
+  def get_attributes(file_name, raise_on_invalid = true)
     result = GetFileAttributesW(wide_string(file_name.to_s))
-    return result unless result == INVALID_FILE_ATTRIBUTES
-    raise Puppet::Util::Windows::Error.new("GetFileAttributes(#{file_name})")
+    if raise_on_invalid && result == INVALID_FILE_ATTRIBUTES
+      raise Puppet::Util::Windows::Error.new("GetFileAttributes(#{file_name})")
+    end
+
+    result
   end
   module_function :get_attributes
 
@@ -179,13 +209,10 @@ module Puppet::Util::Windows::File
 
   FILE_ATTRIBUTE_REPARSE_POINT = 0x400
   def symlink?(file_name)
-    begin
-      attributes = get_attributes(file_name)
-      (attributes & FILE_ATTRIBUTE_REPARSE_POINT) == FILE_ATTRIBUTE_REPARSE_POINT
-    rescue
-      # raised INVALID_FILE_ATTRIBUTES is equivalent to file not found
-      false
-    end
+    attributes = get_attributes(file_name, false)
+
+    return false if (attributes == INVALID_FILE_ATTRIBUTES)
+    (attributes & FILE_ATTRIBUTE_REPARSE_POINT) == FILE_ATTRIBUTE_REPARSE_POINT
   end
   module_function :symlink?
 
--- a/spec/unit/util/windows/file_spec.rb
+++ b/spec/unit/util/windows/file_spec.rb
@@ -1,0 +1,34 @@
+#! /usr/bin/env ruby
+require 'spec_helper'
+require 'puppet/util/windows'
+
+describe "Puppet::Util::Windows::File", :if => Puppet::Util::Platform.windows? do
+  include PuppetSpec::Files
+
+  let(:nonexist_file) { 'C:\somefile\that\wont\ever\exist' }
+  let (:invalid_file_attributes) { 0xFFFFFFFF } #define INVALID_FILE_ATTRIBUTES (DWORD (-1))
+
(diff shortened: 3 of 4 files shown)


lib/puppet/util/windows/file.rb
```diff
@@ -89,10 +89,31 @@ def symlink(target, symlink)
   end
   module_function :symlink
 
+
+  def file_exist(file_name)
+    path = file_name.to_s
+    attempts = 10
```
Iristyle: `10` seems like a pretty arbitrary depth limit to me....

glennsarti: Yes.  To avoid circular references a simple max depth was set.  Do you think it should be higher?

lib/puppet/util/windows/file.rb
```diff
@@ -89,10 +89,31 @@ def symlink(target, symlink)
...
+
+
   INVALID_FILE_ATTRIBUTES = 0xFFFFFFFF #define INVALID_FILE_ATTRIBUTES (DWORD (-1))
 
-  def get_attributes(file_name)
+  def get_attributes(file_name, raise_on_error = true)
     result = GetFileAttributesW(wide_string(file_name.to_s))
+    return result unless raise_on_error
     return result unless result == INVALID_FILE_ATTRIBUTES
     raise Puppet::Util::Windows::Error.new("GetFileAttributes(#{file_name})")
```
Iristyle: I think the multiple returns / conditions makes things a little less clean - maybe better to refactor like this:

```ruby
def get_attributes(file_name, raise_on_invalid = true)
  result = GetFileAttributesW(wide_string(file_name.to_s))
  if raise_on_invalid && result == INVALID_FILE_ATTRIBUTES
    raise Puppet::Util::Windows::Error.new("GetFileAttributes(#{file_name})")
  end

  result
end
```

I'm still a little torn about whether we should file a ticket to make the default `false` for the raise in a future release. We can easily change the two other callsites at `add_attributes` (only used in 1 place in `spec/integration/util/windows/security_spec.rb`) and `remove_attributes` (only used in 1 place in `puppet/util/windows/security.rb`) to throw if necessary.  I would be surprised if we had any downstream consumers in module code, etc.

lib/puppet/util/windows/file.rb
```diff
@@ -179,12 +200,11 @@ def self.device_io_control(handle, io_control_code, in_buffer = nil, out_buffer
 
   FILE_ATTRIBUTE_REPARSE_POINT = 0x400
   def symlink?(file_name)
-    begin
-      attributes = get_attributes(file_name)
-      (attributes & FILE_ATTRIBUTE_REPARSE_POINT) == FILE_ATTRIBUTE_REPARSE_POINT
-    rescue
-      # raised INVALID_FILE_ATTRIBUTES is equivalent to file not found
+    attributes = get_attributes(file_name, false)
```
Iristyle: It's too bad that `INVALID_FILE_ATTRIBUTES` is all 1s.... but you can still remove the extra `else` branch if you want:

```ruby
def symlink?(file_name)
  attributes = get_attributes(file_name, false)

  return false if (attributes == INVALID_FILE_ATTRIBUTES)
  (attributes & FILE_ATTRIBUTE_REPARSE_POINT) == FILE_ATTRIBUTE_REPARSE_POINT
end
```

lib/puppet/file_system/windows.rb
```diff
@@ -4,21 +4,7 @@
...
-
-    begin
-      if Puppet::Util::Windows::File.symlink?(path)
-        path = Puppet::Util::Windows::File.readlink(path)
-      end
-      ! Puppet::Util::Windows::File.stat(path).nil?
-    rescue # generally INVALID_HANDLE_VALUE which means 'file not found'
-      false
-    end
+    return Puppet::Util::Windows::File.file_exist(path)
```
Iristyle: 👍

lib/puppet/util/windows/file.rb
```diff
@@ -89,10 +89,31 @@ def symlink(target, symlink)
   end
   module_function :symlink
 
+
+  def file_exist(file_name)
+    path = file_name.to_s
```
Iristyle: I think you've lost the ability to support `WatchedFile`

Also, please keep the note about why we `to_s` here.  We might want to backfill tests for that if nothing broke during testing.

glennsarti: Why do you say `WatchedFile` would be lost?  It's using the `file_system/windows.rb` implementation of stat.

lib/puppet/util/windows/file.rb
Iristyle: You could also maintain a list of seen links and if you revisit a seen one, `raise`.

If you want to use a bounded loop, the canonical Ruby way is with `0.upto(10)`

glennsarti: Fair point.  I'll keep a list of seen links.

lib/puppet/util/windows/file.rb
```diff
@@ -89,10 +89,31 @@ def symlink(target, symlink)
...
+    attempts = 10
+    loop do
+      result = get_attributes(path,false)
+      # return false for file not found
+      return false if result == INVALID_FILE_ATTRIBUTES
+      # return true if we found a file/dir and it's not a symlink
```
Iristyle: I think your logic is right, though someone reading it might think it a little peculiar.  Might want to leave a note about all the other file attributes that could be set:  https://msdn.microsoft.com/en-us/library/windows/desktop/gg258117(v=vs.85).aspx

lib/puppet/util/windows/file.rb
```diff
@@ -89,10 +89,31 @@ def symlink(target, symlink)
...
+      return true if (result & FILE_ATTRIBUTE_REPARSE_POINT) != FILE_ATTRIBUTE_REPARSE_POINT
+      # walk the symlink and try again...
+      path = readlink(path)
```
Iristyle: So I just verified that walking the links is how we're supposed to behave, and how we document this functioning.  Based on the code changes, it seems like a change in behavior... and I was going to suggest the following tests at line 246 of `file_system_spec.rb`:

```ruby
it "should return true for exist? when resolving a symlink chain pointing to a file" do
      # point symlink -> file
      symlink = tmpfile("somefile_link")
      Puppet::FileSystem.symlink(file, symlink)

      # point symlink2 -> symlink
      symlink2 = tmpfile("somefile_link2")
      Puppet::FileSystem.symlink(symlink, symlink2)

      expect(Puppet::FileSystem.exist?(symlink2)).to be_truthy
    end

    it "should return false for exist? when resolving a symlink chain that dangles" do
      # point symlink -> file
      symlink = tmpfile("somefile_link")
      Puppet::FileSystem.symlink(file, symlink)

      # point symlink2 -> symlink
      symlink2 = tmpfile("somefile_link2")
      Puppet::FileSystem.symlink(symlink, symlink2)

      # yank file, and make symlink dangle
      ::File.delete(file)

      # symlink2 is now indirectly dangled
      expect(Puppet::FileSystem.exist?(symlink2)).to be_falsey
    end
```
What's weird is that these tests pass before and after your changes, but I haven't debugged in to understand why.

lib/puppet/util/windows/file.rb
Iristyle: Because the call to `get_attributes` on line 97 expects a string.  To maintain compatibility with Ruby itself, `exist?` can be passed a `WatchedFile` instance, which needs `to_str` called against it to get the actual path string.

in other words, this line from the refactor never made it into this method:

```
path = path.to_str if path.respond_to?(:to_str) # support WatchedFile
```

Iristyle: Note that you also lost the check for `manages_symlinks?` / the fallback to built-in Ruby behavior - which is probably OK in this case, as that simply checks for the existence of `CreateSymbolicLinkW` (not that you actually have the token privilege to use it).

I'm quite certain the end result will be the same, because `GetFileAttributes` should behave OK everywhere, it's just that `FILE_ATTRIBUTE_REPARSE_POINT` will never be returned where there's no symlink support.

👍

glennsarti: Correct.  There was no need for that check anymore.

glennsarti:
Updated all code as per comments

spec/unit/util/windows/file_spec.rb
```diff
@@ -0,0 +1,34 @@
+#! /usr/bin/env ruby
+require 'spec_helper'
+require 'puppet/util/windows/file'
```
Iristyle: I think you want to move this file to `spec/integration/util/windows/file_spec.rb` and change this include to:

```ruby
require 'puppet/util/windows'
```

That should make sure it doesn't fail on non-Windows.

glennsarti: Yes I realized my mistake on the require fairly quickly.  I disagree that it should be in integration though.

PR #4892, opened 2016-04-15
(PUP-6096) Acceptance test for long file paths with PMT

zreichert:


acceptance/tests/modules/install/long_path_name.rb
```diff
@@ -0,0 +1,36 @@
+test_name "puppet module install (long file path)"
+require 'puppet/acceptance/module_utils'
+extend Puppet::Acceptance::ModuleUtils
+
+confine :except, :platform => /centos-4|el-4/ # PUP-5226
+confine :except, :platform => /aix/ # PUP-5501
+
+module_author = 'ztr'
+module_name   = 'longpath'
```
Iristyle: Note that your module needs to be of a specific tar format to expose the problem you're trying to expose.  I believe v7 format archives had the problem, but POSIX formats do not.

zreichert: @Iristyle I can ensure that the module tar is in the v7 format.

acceptance/tests/modules/install/long_path_name.rb
```diff
@@ -0,0 +1,34 @@
+test_name "puppet module install (long file path)"
```
er0ck: nit: test_name should have do/end now

acceptance/tests/modules/install/long_path_name.rb
```diff
@@ -0,0 +1,34 @@
...
+require 'puppet/acceptance/module_utils'
+extend Puppet::Acceptance::ModuleUtils
+
+confine :except, :platform => /centos-4|el-4/ # PUP-5226
+confine :except, :platform => /aix/ # PUP-5501
+
```
er0ck: could these use expect_failure?

zreichert: @er0ck this was pulled from existing PMT tests

acceptance/tests/modules/install/long_path_name.rb
```diff
@@ -0,0 +1,34 @@
...
+    @module_path = /(Notice: Preparing to install into )(.*)( \.\.\.)/.match(result.stdout)[2]
+    assert_module_installed_ui(stdout, module_author, module_name)
+  end
+
+  path = "#{@module_path}/#{module_name}"
+  path = on(agent, "cygpath -u #{path}").stdout.chomp if /win/ =~ agent.platform
+
+  step 'Count the files extracted from the tar'
+  on(agent, "find #{path} -not -path '*/\\.*' -type f | wc -l") do |result|
+    assert_match(/2602/, result.stdout, 'The correct number of files was not observed')
```
er0ck: is this the best way to ensure files with long paths are installed?
a hard-coded number feels wrong in many ways.  maybe we could grab the tarball and introspect it and count the files and compare?

acceptance/tests/modules/install/long_path_name.rb
```diff
@@ -0,0 +1,34 @@
...
+  step "install module '#{module_author}-#{module_name}'"
+  on(agent, puppet("module install #{module_author}-#{module_name} --version #{module_version}")) do |result|
+    @module_path = /(Notice: Preparing to install into )(.*)( \.\.\.)/.match(result.stdout)[2]
+    assert_module_installed_ui(stdout, module_author, module_name)
+  end
+
+  path = "#{@module_path}/#{module_name}"
+  path = on(agent, "cygpath -u #{path}").stdout.chomp if /win/ =~ agent.platform
+
+  step 'Count the files extracted from the tar'
```
er0ck: i don't see a tarball extraction above.  this seems midlly confusing

acceptance/tests/modules/install/long_path_name.rb
zreichert: @er0ck This artifact is a know bad artifact that specifically caused a failure with PMT on Windows.  This module is on the Acceptance forge

acceptance/tests/modules/install/long_path_name.rb
er0ck: oh right.  i'm pretty sure the aix confine in the other tests is mine. 
if they all do this, because... pmt, could probably use an abstraction across tests.  oh well.  another ticket

acceptance/tests/modules/install/long_path_name.rb
er0ck: sure.  i'm just saying we can make the test a bit more robust.  
ensure this module still uses long paths, whatever that means.
and introspect the tarball so when the number of files changes, this test doesn't need updating.

Iristyle: @er0ck I agree... except that would require a little rejiggering.

The tarball would need to be pulled down to the Beaker coordinator, cracked open / examined (which would have to be done with `minitar` - which is partly what this test is testing).

Alternatively, you could include a small tarball in the PR as a fixture to reduce the dependency on upstream forge (which is probably a good thing IMHO).  Then the file would already be on the Beaker coordinator for examination.  Then the file could be scp'd around to the agents.

PMT install from forge and PMT install from local tarball should have the same behavior for the sake of this test I believe.

er0ck: the goal in my comments above is to eliminate the '2602' magic number, and ensure the test is actually testing long paths.  i'm probably missing something, but i don't see how some 'small tarball' as a fixture addresses either of those.
the tarball of the actual module only has to be introspected once, on the runner, or the first agent, maybe.  whichever is easier.

Iristyle: I'm not disagreeing about the magic number removal - I was talking about what would need to be done to facilitate that.

* You probably want to have access to the tarball to get the file list / count on the coordinator side
* Windows agents probably shouldn't introspect the tarball:
  * `tar` command in in Cygwin, but not elsewhere... I think you would have to pipe the extraction through `| wc`, which is a little weird
  * If using a Ruby script on Windows that would require using minitar to examine contents (not recommended given that's where the extract bug presented)
  * It's a bit weird to ask agent for count of files in tarball, to then plug that back into the coordinator side, to then ask the agent to verify

Therefore, 2 options:

* Pull the `.tar.gz` from fake forge / Internet to coordinator, and examine
* Have the `.tar.gz` in this repo, which reduces network dependency (the comment about file size was to not bloat this git repo with large binaries)

IMHO, there's no good reason to make an extra pull over the network to grab the module, as copying it via SCP from Beaker coordinator to agents and then calling `puppet module install tarball.tar.gz` achieves the same result.

acceptance/tests/modules/install/long_path_name.rb
```diff
@@ -0,0 +1,35 @@
+test_name "puppet module install (long file path)" do
+  require 'puppet/acceptance/module_utils'
+  extend Puppet::Acceptance::ModuleUtils
+
+  confine :except, :platform => /centos-4|el-4/ # PUP-5226
+  confine :except, :platform => /aix/ # PUP-5501
+
+  module_author = 'puppetlabs'
+  module_name   = 'dsc'
+  module_version = '1.0.0'
```
zreichert: @er0ck I have pinned the dsc module used in test to '1.0.0'.

er0ck:
ahhhhhhhh. i hadn't noticed it's pinned to a version.  that solves most of my issues. 👍

PR #4893, opened 2016-04-18
(PUP-6179) Fix the `puppet ca` face

MikaelSmith:
Calling setup on a face is required to initialize command-line
arguments. Without this the face silently fails to execute. Fix so when
we override setup in the CA face, we still call out to `super`.


MikaelSmith:
This should probably have an acceptance test, I'm going to create a separate ticket to address that.

PR #4895, opened 2016-04-19
(PUP-6025) Fix Windows FileSystem exist? symlink?

Iristyle:
In c64fa0749f673cc75c09b986f3e50a80c7237314 code was introduced to add
symlink support on Windows. The GetFileAttributes API is used to detect
if a given file is a Windows reparse point, which means for Puppets
purpose that the path is a symlink.  However, due to the FFI
implementation of GetFileAttributes, exceptions are thrown under fairly
typical circumstances, when checking for file existence. Puppet performs
a fair amount of Disk IO as it loads (partially exploring module
directories for source to load) -- and this can generate many thrown /
silently eaten exceptions. In a basic Puppet run on Windows, nearly 1000
exceptions are thrown, creating a performance impact. It's uncertain how
much this might affect typical user code, but it's not negligible.

This commit updates the following;

- Updates the exist? method to use the GetFileAttributesW to both detect
  if an item exists and whether it is a symlink.  This reduces Disk IO.

- Also fixed unknown bug whereby the exist? method did not walk symlink
  chains greater than a depth of two (sym -> sym -> file), making the
  Windows implementation previously inconsistent with non-Windows
  platforms.

- Updates the get_attributes helper method to optionally not raise
  errors and pass back the raw result from GetFileAttributesW.  This
  is useful in cases like file existence checks as error messages are
  not required.

- Updates the symlink? method to not catch/raise errors if the target
  path does not exist and instead just return false

- Additional tests for previously unspecified behavior for exists?

### DIFF
--- a/lib/puppet/file_system/windows.rb
+++ b/lib/puppet/file_system/windows.rb
@@ -4,21 +4,7 @@ require 'puppet/util/windows'
 class Puppet::FileSystem::Windows < Puppet::FileSystem::Posix
 
   def exist?(path)
-    if ! Puppet.features.manages_symlinks?
-      return ::File.exist?(path)
-    end
-
-    path = path.to_str if path.respond_to?(:to_str) # support WatchedFile
-    path = path.to_s # support String and Pathname
-
-    begin
-      if Puppet::Util::Windows::File.symlink?(path)
-        path = Puppet::Util::Windows::File.readlink(path)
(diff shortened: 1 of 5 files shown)


Iristyle:
Supercedes #4891 with some minor adjustments...

ferventcoder:
👍

PR #4896, opened 2016-04-19
(PUP-5727) acceptance: fix utf8 cached_catalog test for windows10 32bit

er0ck:
This change fixes the previous version of this test for windows10 32bit.
Tempfiles go in a different place (as addressed from cygwin) in 32bit
windows.

[skip ci]

### DIFF
--- a/acceptance/tests/direct_puppet/supports_utf8.rb
+++ b/acceptance/tests/direct_puppet/supports_utf8.rb
@@ -5,14 +5,19 @@ extend Puppet::Acceptance::EnvironmentUtils
   app_type = File.basename(__FILE__, '.*')
   tmp_environment   = mk_tmp_environment(app_type)
 
-  filenames         = {'windows' => "C:/cygwin64/tmp/#{tmp_environment}_file"}
-  filenames.default = "/tmp/#{tmp_environment}_file"
+  tmp_file = {}
+  agents.each do |agent|
+    tmp_file[agent.hostname] = agent.tmpfile(tmp_environment)
+  end
+
   teardown do
     step 'remove the tmp environment symlink' do
       on master, "rm -rf #{File.join(environmentpath, tmp_environment)}"
     end
     step 'clean out produced resources' do
-      on(agents, "rm #{filenames['windows']}; rm #{filenames.default}", :accept_all_exit_codes => true)
+      agents.each do |agent|
+        on(agent, "rm #{tmp_file[agent.hostname]}", :accept_all_exit_codes => true)
+      end
     end
   end
 
@@ -22,11 +27,7 @@ extend Puppet::Acceptance::EnvironmentUtils
 file { '#{environmentpath}/#{tmp_environment}/manifests/site.pp':
   ensure => file,
   content => '
-\$test_path = \$::osfamily ? {
-  "Windows" => "#{filenames['windows']}",
-  default   => "#{filenames.default}",
-}
-
+\$test_path = \$::fqdn ? #{tmp_file}
 file { \$test_path:
   content => @(UTF8)
     #{file_contents}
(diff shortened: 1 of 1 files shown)


acceptance/tests/direct_puppet/supports_utf8.rb
```diff
@@ -5,14 +5,20 @@
   app_type = File.basename(__FILE__, '.*')
   tmp_environment   = mk_tmp_environment(app_type)
 
-  filenames         = {'windows' => "C:/cygwin64/tmp/#{tmp_environment}_file"}
-  filenames.default = "/tmp/#{tmp_environment}_file"
+  tmp_file = {}
+  agents.each do |agent|
```
Iristyle: You can't just ask for the first Windows agent and the first non-Windows agent to avoid the loop?

acceptance/tests/direct_puppet/supports_utf8.rb
```diff
@@ -5,14 +5,20 @@
...
+    tmp_file[osfamily] = agent.tmpfile(tmp_environment) unless tmp_file[osfamily]
+  end
   teardown do
     step 'remove the tmp environment symlink' do
       on master, "rm -rf #{File.join(environmentpath, tmp_environment)}"
     end
     step 'clean out produced resources' do
-      on(agents, "rm #{filenames['windows']}; rm #{filenames.default}", :accept_all_exit_codes => true)
+      agents.each_with_index do |agent, index|
+        on(agents, "rm #{tmp_file['windows']}; rm #{tmp_file['other']}", :accept_all_exit_codes => true)
```
Iristyle: I see the appeal of doing this in a one-liner, but it does seem mildly dangerous given the paths generated are platform specific.

er0ck: you mean if there happens to be another file in `tmpdir` using the same name?  a name containing the name of this test, with a random string tacked on to it and another random string tacked on to that?

acceptance/tests/direct_puppet/supports_utf8.rb
er0ck: i suppose i could.  but i'm pretty sure any way i do that will be a loop or a collection.  same same?
eg: `first_windows = agents.collect{ |a| a.platform =~ /windows/ }.first`

acceptance/tests/direct_puppet/supports_utf8.rb
seangriff: This version fails for the same reason as before if you have a mix of win-386 and win-x86_64 agents.  The path in tmp_file['windows'] will either be "C:/cygwin/tmp/<file>" or "C:/cygwin64/tmp/<file>".  Neither will work for all windows agents.

er0ck: yes, that's true.  do we have an occasion where we'd have that?

seangriff: Keeping track of the temporary files by agent is not so bad.  Setup:

  tmp_file = {}
  agents.each do |agent|
    tmp_file[agent.hostname] = agent.tmpfile(tmp_environment)
  end

  teardown do
    step 'remove the tmp environment symlink' do
      on master, "rm -rf #{File.join(environmentpath, tmp_environment)}"
    end
    step 'clean out produced resources' do
      agents.each do |agent|
        on(agent, "rm #{tmp_file[agent.hostname]}", :accept_all_exit_codes => true)
      end
    end
  end


Here is a trick that makes the handling of the manifest easy too.  The conditional expression for the content of the temporary file can be expressed by interpolating the tmp_file hash, indexed by agent.hostname.

    manifest = <<MANIFEST
file { '#{environmentpath}/#{tmp_environment}/manifests/site.pp':
  ensure => file,
  content => '
\$test_path = \$::fqdn ? #{tmp_file}
file { \$test_path:
  content => @(UTF8)
    #{file_contents}
    | UTF8
}
  ',
}
MANIFEST


Then later, instead of indexing with ostype (windows/other), just use the agent.hostname and fact fqdn:

  step 'run agent(s)' do
    with_puppet_running_on(master, {}) do
      agents.each do |agent|
        fqdn = agent.hostname

        ...

        on(agent, "cat #{tmp_file[fqdn]}").stdout do |result|
          assert_equal(file_contents, result, 'file contents did not match accepted')


See https://gist.github.com/seangriff/de0ca02ac448bd3e740dafebde6f3c49 for complete file.

PR #4898, opened 2016-04-20
(PUP-6188) Add UTF-8 support to FileSystem.open / exclusive_open

Iristyle:
WIP

A couple of other things to scope out:
- Other FileSystem calls that don't specify encoding
- in `filetype.rb` for instance, `Tempfile` and `IO.open` / `IO.popen`

``` ruby
file = Tempfile.new("puppet_suntab")
file.print "text"

IO.popen("#{cmdbase()} -", "w") { |p|
```


Iristyle:
Closing in favor of #5354

PR #4904, opened 2016-04-22
(PUP-2419) Deprecate never used req_bits setting

joshcooper:
The req_bits setting was added in bb4b5a54, but has only been used in
the test_helper.rb invoked by spec_helper.rb.

This commit deprecates the setting completely and removes it from the
test_helper.rb.


lib/puppet/test/test_helper.rb
```diff
@@ -238,7 +238,6 @@ def self.initialize_settings_before_each()
       #
       # I would make these even shorter, but OpenSSL doesn't support anything
       # below 512 bits.  Sad, really, because a 0 bit key would be just fine.
```
branan: This comment would be truly horrifying if it wasn't in a test

PR #4905, opened 2016-04-22
(PUP-4617) Display human readable cert extensions

joshcooper:
Previously, `puppet cert print` would always show the OID for
custom X509v3 cert extensions:

    X509v3 extensions:
        1.3.6.1.4.1.34380.1.2.1.1:
            ..somedata

With this commit, if a `trusted_oid_mapping_file` is specified mapping
OIDs to their short and long human readable names, then the long name
will be displayed, similarly to what is displayed for puppet OIDs like
"Puppet Node UUID" and other well-known OIDs:

    $ puppet cert print <hostname> --trusted_oid_mapping_file oids.yaml
    ...
    X509v3 extensions:
        Netscape Comment:
            Puppet Ruby/OpenSSL Internal Certificate
        X509v3 Subject Key Identifier:
            47:BC:D5:14:33:F2:ED:85:B9:52:FD:A2:EA:E4:CC:00:7F:7F:19:7E
        Puppet Node UUID:
            ED803750-E3C7-44F5-BB08-41A04433FE2E
        My Long Name:
            ..somedata

If the trusted_oid_mapping_file doesn't exist (the default), then there
is no change in behavior.

### DIFF
--- a/lib/puppet/application/cert.rb
+++ b/lib/puppet/application/cert.rb
@@ -246,6 +246,7 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
     exit(Puppet.settings.print_configs ? 0 : 1) if Puppet.settings.print_configs?
 
     Puppet::SSL::Oids.register_puppet_oids
+    Puppet::SSL::Oids.load_custom_oid_file(Puppet[:trusted_oid_mapping_file])
 
     Puppet::Util::Log.newdestination :console
 


joshcooper:
Thoughts on how to test this? It causes so low level changes at the ruby/openssl layer that are probably best handled in acceptance?

Also /cc @nfagerlund @gguillotte as this has docs impact https://docs.puppet.com/puppet/latest/reference/config_file_oid_map.html

And /cc @mrzarquon @adrienthebo

gguillotte:
CC @jtappa ^^

adrienthebo:
If I recall correctly registering additional OIDs permanently changes the state of OpenSSL under the hood and it's probably a fair amount of work to try to delete registered OIDs - that is, if it's possible at all. Acceptance tests seem like a reasonable approach.

Implementation seems sound and unsurprising, +1 from me.

kylog:
Ping @er0ck re acceptance tests.

er0ck:
thanks @kylog 
the ticket is marked QA-risk as "medium"
meaning we'll validate and automate this "later", possibly _much_ later.
we're tracking it as a test-case here:
https://testrail.ops.puppetlabs.net//index.php?/cases/view/97925

is there any reason i might be missing that this should be automated soon (qa-risk as high)?
e.g.: either its user severity or probability of occurrance are high?

HAIL9000:
I'm going to go ahead and merge this for the sake of getting it through the current Client sprint, and we can definitely continue docs/testing discussion on the ticket itself.

PR #4906, opened 2016-04-22
Additional maint fixes related to Puppet Type system internals

thallgren:
This PR contains some additional related to internal logic in the Puppet Type system.


hlindberg:
ok to merge once travis goes green

PR #4907, opened 2016-04-22
(PUP-6228) Unpin run order in netconsole rc.d links

johnduarte:
This commit removes the pinned run order of the expected symlinks
created for the `netconsole` service. Prior to this commit,
Scientific Linux 6 instances were failing this test when the
generated run order did not equal '50'.


johnduarte:
This has been tested against Scientific 6.

whopper:
Appveyor and travis seem to have been stuck for quite a while, but this is an acceptance-only change so 👍

PR #4912, opened 2016-04-25
(PUP-6224) Add a Ruby generator that emits code based on Object types

thallgren:
This PR adds a Ruby generator capable of generating a default implementation of an `Object` type. Instances of the object can be created either from a hash (validated by the `#i12n_type` of the `Object`) or by a constructor using anonymous parameters.

The generator is capable of generating both anonymous ruby classes and a static implementation intended to be written to files and then included in a code base.

### DIFF
--- a/lib/puppet/pops/types/puppet_object.rb
+++ b/lib/puppet/pops/types/puppet_object.rb
@@ -1,0 +1,21 @@
+module Puppet::Pops
+module Types
+
+# Marker module for implementations that are mapped to Object types
+# @api public
+module PuppetObject
+  # Returns all classes that includes this module
+  def self.descendants
+    ObjectSpace.each_object(Class).select { |klass| klass < self }
+  end
+
+  # Returns the Puppet Type for this instance. The implementing class must
+  # add the {#_ptype} as a class method.
+  #
+  # @return [PObjectType] the type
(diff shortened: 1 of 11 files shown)


lib/puppet/pops/types/puppet_object.rb
```diff
@@ -0,0 +1,21 @@
+module Puppet::Pops
+module Types
+
+# Marker module for implementations that are mapped to Object types
+# @api public
+module PuppetObject
+  # Returns all classes that includes this module
+  def self.descendants
+    ObjectSpace.each_object(Class).select { |klass| klass < self }
```
hlindberg: Where/how is this used? Looks like it could potentially be very slow (iterates over everything in memory)

hlindberg:
Have some small concerns about the performance of ruby_generator.rb - but that can wait until we have real workloads - for now it is more important that the code is easy to read and maintain (which it is).

hlindberg:
Good to merge once tests go green.

hlindberg:
One question - the I12n_type looks very much like what the new function is doing - where that type is described by dispatch; that is probably overkill here (needs to be fast) - but it feels like it is the same concept. Suggest dealing with that when adding support for the new function to Object type.

PR #4913, opened 2016-04-25
(PUP-6232) Allow module_tool to keep supporting ruby 1.8.7

ghoneycutt:
Rake 11 dropped support for ruby 1.8.7, so if we are using that, we need
to pin rake to v10.


kylog:
👍 

Makes sense. Ping @andersonmills.

andersonmills:
:+1:

This is good by me.

andersonmills:
looks like things are good to merge with our tests.

PR #4914, opened 2016-04-26
(PUP-6072) Make --lastrunreport be a path-style argument

richardc:
file-type argument to being a path-type argument.

commit fb252414284a7e31d1290fddcbec155599bc296f

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1517,8 +1517,7 @@ EOT
     },
     :lastrunreport =>  {
       :default  => "$statedir/last_run_report.yaml",
-      :type     => :file,
-      :mode     => "0640",
+      :type     => :path,
       :desc     => "Where puppet agent stores the last run report in yaml format."
     },
     :graph => {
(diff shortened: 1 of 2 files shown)


richardc:
Alternate implementation for PR #4835

kylog:
The travis failures are unrelated to this PR and have just cropped up recently.

lib/puppet/defaults.rb
```diff
@@ -1517,8 +1517,7 @@ def self.default_diffargs
     },
     :lastrunreport =>  {
       :default  => "$statedir/last_run_report.yaml",
-      :type     => :file,
-      :mode     => "0640",
```
MikaelSmith: I'm suspicious of losing this mode enforcement. Also, I'd like to update the description to be clear about how the path is used.

richardc: Well in order to not lose it, it needs a new ::Puppet::Settings::FilePathSetting (or a better name) to be created, to combine the behaviours of type :file and type :path.

MikaelSmith:
Continued in https://github.com/puppetlabs/puppet/pull/4930.

PR #4915, opened 2016-04-26
(PUP-6255) Add new function to object type

thallgren:
This PR adds the ability to create new instances of an Object type from the Puppet Language. It also adds the ability to call members (attributes and functions) on the object since it's difficult to test instantiations of an object if it is possible to examine the result.

### DIFF
--- a/lib/puppet/pops/types/p_object_type.rb
+++ b/lib/puppet/pops/types/p_object_type.rb
@@ -40,6 +40,7 @@ class PObjectType < PAnyType
     KEY_VALUE => PAnyType::DEFAULT
   })
   TYPE_ATTRIBUTES = TypeFactory.hash_kv(TYPE_MEMBER_NAME, TypeFactory.not_undef)
+  TYPE_ATTRIBUTE_CALLABLE = TypeFactory.callable(0,0)
 
   TYPE_FUNCTION_TYPE = PType.new(PCallableType::DEFAULT)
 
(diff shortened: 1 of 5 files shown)


hlindberg:
Restarted travis in case the errors were transient.

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -333,6 +335,76 @@ def initialize(name_or_i12n_hash, i12n_hash_expression = nil)
...
+          self.class.impl_class.from_asserted_hash(hash)
+        end
+      end
+
+      dispatcher.add_dispatch(create_type, :create, param_names, nil, EMPTY_ARRAY, EMPTY_ARRAY, false)
+      def create(*args)
+        self.class.impl_class.new(*args)
+      end
+    end
+  end
```
hlindberg: There is quite a lot going on in this method - I think more comments would help a reader (I understand a bit more because I have to context of this PR). For instance what is the @impl_class, what is asserted_hash method.

hlindberg:
Regarding invocation of methods on an object - I wonder if it would be best to differentiate between function call style i.e. `func(arg1, arg2)` and method call style `arg1.method(arg2)` such that function call style *never* invokes a method. That also makes it possible to call a function, using function style with a first argument being an object type from within a method. Related to this, if there is a desire to call a super version of a method - that could be performed by using the function style of call (given that the function is found can be referenced/named and called). Not sure how to do that the best way - the name of a function would be something like `MyType.func` which is currently an illegal combination (you would need to write `MyType['func'](args)`). However, using explicitly named super types is not a good thing, and something like a reference to Super or super would be needed.

I think I would like to take out the functionality to call object functions until we have specified how it should work. We do however have the immediate problem of how to handle calling a super method for initialization.

thallgren:
Without the ability to call object functions, we also have the problem that the created objects are useless and cannot be tested. So in that case, I'd prefer if we defer this PR completely. No use in creating new instances unless they can be used in some context, even if it's only testing at this point.

hlindberg:
True - they could have some kind of default "to_s" that can be used for testing; or that you can read their attribute values without also being able to call the methods.

hlindberg:
I am happy merging this, and then changing it, but I think we should talk about this first.

thallgren:
Branch name was wrong on this PR so I'm closing it. It's superseded by [PR-4923](/puppetlabs/puppet/pull/4923).

PR #4917, opened 2016-04-26
DO NOT MERGE (maint) Not really a change

kylog:
Just here to test travis ci

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ group(:development, :test) do
 
   gem 'rdoc', "~> 4.1", :platforms => [:ruby]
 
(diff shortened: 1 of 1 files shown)


MikaelSmith:
looks like we're now picking up webmock 2.0.0 instead of 1.24.5.

MikaelSmith:
Oops, didn't pin webmock at all: https://github.com/puppetlabs/puppet/blob/master/Gemfile#L53

kylog:
\o/

Thanks @MikaelSmith ! I'll actually point the Gemfile change at stable.

PR #4922, opened 2016-04-27
smf provider updates

shawnferry:


shawnferry:
@hlindberg  do you have any pointers for that failing test? I'm not having that problem when I'm testing myself. 

If found that I needed to loadfacts to get it working locally. I don't understand why it's failing on travis.

```
ruby 2.1.7p400 (2015-08-18 revision 51632) [x86_64-darwin15.0]

bundle exec rspec -e Puppet::Type::Service::ProviderSmf spec/unit/provider/service
zsh: correct 'rspec' to 'spec' [nyae]? n
Run options:
  include {:full_description=>/Puppet::Type::Service::ProviderSmf/}
  exclude {:broken=>true, :benchmark=>true}
..............................

Finished in 1.4 seconds (files took 1.69 seconds to load)
30 examples, 0 failures

bundle check
Resolving dependencies...
The Gemfile's dependencies are satisfied
```

spec/unit/provider/service/smf_spec.rb
```diff
@@ -174,14 +176,35 @@
...
+        Facter.stubs(:fact).with(:kernelrelease).returns Facter.add(:kernelrelease) { setcode { '5.12' } }
+        @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "-s", "/system/myservice"], true)
+        @provider.expects(:wait).with('online')
+        @provider.restart
+      end
     end
 
     it "should error if timeout occurs while restarting the service" do
-      @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "/system/myservice"], true)
+      Facter.loadfacts # With facts cleared timeout testing fails; load them again
```
MikaelSmith: Were these tests not previously run? I'm not sure why Facter.loadfacts is now needed.

MikaelSmith: Ah, because you're resetting facts earlier. You probably still need to do something to resolve facts before setting the expectation on Timeout.

HAIL9000:
Ping @whopper @branan on this spec failure since y'all added the timeout

spec/unit/provider/service/smf_spec.rb
```diff
@@ -174,17 +174,42 @@
...
-      @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "/system/myservice"], true)
+      @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "-s", "/system/myservice"], true)
       Timeout.expects(:timeout).with(60).raises(Timeout::Error)
       expect { @provider.restart }.to raise_error Puppet::Error, ('Timed out waiting for /system/myservice to transition states')
     end
 
+    context 'with :kernelrelease == 5.10' do
+      it "should call 'svcadm restart /system/myservice'" do
+        Facter.instance_variable_get(:@collection).instance_variable_get(:@facts).delete_if{|k,v| k==:kernelrelease}
+        Facter.stubs(:fact).with(:kernelrelease).returns Facter.add(:kernelrelease) { has_weight 500; setcode { '5.10' } }
```
shawnferry: I've reduced the scope of the Facter change. I thought it should only be affecting the smaller set of tests but the timeout test still fails.

spec/unit/provider/service/smf_spec.rb
```diff
@@ -174,17 +174,42 @@
...
   describe "when restarting" do
-    it "should call 'svcadm restart /system/myservice'" do
-      @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "/system/myservice"], true)
-      @provider.expects(:wait).with('online')
-      @provider.restart
-    end
 
     it "should error if timeout occurs while restarting the service" do
-      @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "/system/myservice"], true)
+      @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "-s", "/system/myservice"], true)
```
whopper: @shawnferry the spec failure is happening here because `Facter.value(:kernelrelease)` isn't being stubbed in the test, so Puppet is falling back to the restart `else` clause where we don't call with the `-s` flag:

```
  def restartcmd
    if Puppet::Util::Package.versioncmp(Facter.value(:kernelrelease), '5.11') >= 0
      [command(:adm), :restart, "-s", @resource[:name]]
    else
      # Solaris 10 does not have synchronous restart
      [command(:adm), :restart, @resource[:name]]
    end
  end
```

To fix it, you can just add the following to line 179:

```
Facter.stubs(:value).with(:kernelrelease).returns '5.11'
```

Additionally, you can remove the `Facter.instance_variable` and `Facter.add` bits and use the same method as above to set the `kernelrelease` fact.

spec/unit/provider/service/smf_spec.rb
whopper: These two lines can just be:

```
Facter.stubs(:value).with(:kernelrelease).returns '5.10'
```

spec/unit/provider/service/smf_spec.rb
```diff
@@ -174,17 +174,42 @@
...
+        Facter.stubs(:fact).with(:kernelrelease).returns Facter.add(:kernelrelease) { has_weight 500; setcode { '5.10' } }
+        @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "/system/myservice"], true)
+        @provider.expects(:wait).with('online')
+        @provider.restart
+      end
+    end
+
+    context 'with :kernelrelease == 5.11' do
+      it "should call 'svcadm restart -s /system/myservice'" do
+        Facter.instance_variable_get(:@collection).instance_variable_get(:@facts).delete_if{|k,v| k==:kernelrelease}
```
whopper: Same comment as above.

spec/unit/provider/service/smf_spec.rb
```diff
@@ -174,17 +174,42 @@
...
+        Facter.stubs(:fact).with(:kernelrelease).returns Facter.add(:kernelrelease) { has_weight 500; setcode { '5.11' } }
+        @provider.expects(:texecute).with(:restart, ["/usr/sbin/svcadm", :restart, "-s", "/system/myservice"], true)
+        @provider.expects(:wait).with('online')
+        @provider.restart
+      end
+    end
+
+    context 'with :kernelrelease > 5.11' do
+      it "should call 'svcadm restart -s /system/myservice'" do
+        Facter.instance_variable_get(:@collection).instance_variable_get(:@facts).delete_if{|k,v| k==:kernelrelease}
```
whopper: Same comment as above - last one :)

shawnferry:
@whopper thanks, I would swear I tried that but apparently not.

branan:
@shawnferry I think this looks good at this point! Would you mind just squashing down your test fixes into the original commit, then we should be good to go

shawnferry:
@branan did you mean to close this without merging? I squashed the commit but you can't that here.

branan:
@shawnferry That was a mis-click ("Close" is right next to "Comment"). Unfortunately github won't let me re-open a PR once the branch has been force pushed to. If you open a new PR I'll go ahead and merge it, though.
ou open a new PR I'll go ahead and merge it, though.

Sorry about this :(

PR #4923, opened 2016-04-27
(PUP-6225) Add new function to object type

thallgren:
This PR adds the ability to create new instances of an Object type from the Puppet Language. It also adds the ability to call members (attributes and functions) on the object since it's difficult to test instantiations of an object if it is possible to examine the result.

### DIFF
--- a/lib/puppet/pops/evaluator/runtime3_support.rb
+++ b/lib/puppet/pops/evaluator/runtime3_support.rb
@@ -505,7 +505,7 @@ module Runtime3Support
       elsif Puppet[:strict] == :off
         p[Issues::UNKNOWN_VARIABLE] = :ignore
       else
-        Puppet[:strict_variables] 
+        Puppet[:strict_variables]
         p[Issues::UNKNOWN_VARIABLE] = Puppet[:strict]
       end
 
@@ -520,7 +520,7 @@ module Runtime3Support
     def accept(diagnostic)
       super
       IssueReporter.assert_and_report(self, {
-        :message => "Evaluation Error:", 
+        :message => "Evaluation Error:",
         :emit_warnings => true,  # log warnings
         :exception_class => Puppet::PreformattedError
       })
--- a/lib/puppet/pops/types/p_object_type.rb
+++ b/lib/puppet/pops/types/p_object_type.rb
@@ -1,3 +1,5 @@
+require_relative 'ruby_generator'
+
 module Puppet::Pops
 module Types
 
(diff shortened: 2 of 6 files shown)


thallgren:
This PR supersedes [PR-4915](/puppetlabs/puppet/pull/4915) due to an issue number confusion in the branch name. Same commits except for some added comments in the #new_function method in response to comment from @hlindberg.

lib/puppet/pops/evaluator/runtime3_support.rb
```diff
@@ -266,6 +266,14 @@ def external_call_function(name, args, scope, &block)
   end
 
   def call_function(name, args, o, scope, &block)
+    if args.size >= 1
+      receiver_type = Types::TypeCalculator.infer(args[0])
+      if receiver_type.is_a?(Types::PObjectType)
+        member = receiver_type[name]
+        return member.invoke(scope, args, &block) unless member.nil?
+      end
+    end
```
hlindberg: As per the specification - this functionality should be in the evaluator where it can differentiate between a method call and a function call.

lib/puppet/pops/types/p_object_type.rb
```diff
@@ -333,6 +372,86 @@ def initialize(name_or_i12n_hash, i12n_hash_expression = nil)
     end
   end
 
+  # @api private
+  def new_function(loader)
+    @new_function ||= create_new_function(loader)
+  end
+
+  # @api private
+  def create_new_function(loader)
```
hlindberg: The documentation of this needs to go into the documentation of the `new` function as that is the only documentation that will be visible to users. At some later point documentation needs to be extracted from object types.

hlindberg:
What should happen when instantiating an object type that requires implementation (i.e. it has derived attributes and declares functions) ? By default I think that should raise an error. It is however of value to be able to instantiate such objects when you are aware of this  - the question then is how to signal this as the new call passes on all arguments to the object. The other alternative is to ignore this completely and simply error when trying to use the features that require an implementation (i.e. calling a_person.age() when age is a derived attribute). Third option - default implementation adds methods that accepts the arguments as specified and always returns undef after possibly logging a warning.

I think I prefer it to just error when requiring impl and not having one.

hlindberg:
The act of binding implementation to interface (object type) requires a proper specifcation. And the error suggested (previous comment) will ensure that only the sensical case of object not requiring an impl, and not having an impl binding gets a workable dynamic impl. Then we can work on the next phase without breaking people's code).

hlindberg:
So, I would like to see tests that provoke those errors.

thallgren:
PR rebased and the member call logic moved.

thallgren:
Regarding when to error out if an implementation is required. Consider the case when you want to do a model to model transformation. Requiring an implementation of derived attributes or functions is then not necessary unless the transformation requires access to them (which often isn't the case). Hence the current design where the Ruby generator emits code that will yield an "no method is implemented for ...". My preference would be to keep it that way so that models can be instantiated dynamically even when no implementation exists.

lib/puppet/pops/types/p_object_type.rb
thallgren: I think this is problematic. The only documentation that can be added at this point would be documentation describing how a `new` function interacts with the meta-type `Object`. At present, the documentation is all about type conversions. Adding meta-type documentation at the same level will be confusing. It needs to be split up somehow.

thallgren:
Regarding spec for binding implementation to interface. IMO, that should be part of the Object type spec.

lib/puppet/pops/types/p_object_type.rb
hlindberg: Yes, I know (I just said so). Something needs to be said about this in the `new` function until we have a better way.

hlindberg:
Regarding errors when there is no impl: While there should be a way to get a default (dynamic) implementation, having that as the default is problematic. When you are doing transformations and similar things you know you are not going to be using implementations and you can take action to state this is ok. In general users just loading stuff unknowingly will end up with errors like "no method implemented" errors which is going to mean nothing to them. It is far better to error at the earliest possible point.

The difference in when an error (and what the error is) is less of a concern than the possible API changes this implies.

hlindberg:
I am merging this as the basis for further work. There are several things missing that we are just coming to consensus on.
There will be tickets...

PR #4925, opened 2016-04-28
(PUP-5284) Make test skip reasons explicit

johnduarte:
Update acceptance tests to make the reasons for their status as 'skipped' explicit.

### DIFF
--- a/acceptance/tests/resource/service/ticket_4123_should_list_all_running_redhat.rb
+++ b/acceptance/tests/resource/service/ticket_4123_should_list_all_running_redhat.rb
@@ -1,12 +1,10 @@
 test_name "#4123: should list all running services on Redhat/CentOS"
+confine :to, :platform => /(el|centos|oracle|redhat|scientific)-5/
+
 step "Validate services running agreement ralsh vs. OS service count"
 # This will remotely exec:
 # ticket_4123_should_list_all_running_redhat.sh
 
 hosts.each do |host|
-  if host['platform'] =~ /el-5/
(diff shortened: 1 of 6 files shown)


acceptance/tests/resource/service/ticket_4123_should_list_all_running_redhat.rb
```diff
@@ -1,12 +1,10 @@
 test_name "#4123: should list all running services on Redhat/CentOS"
+confine :to, :platform => /el-5/
```
whopper: @johnduarte this was an issue before this change, but `el-5` only covers RedHat 5, despite the test title specifying RedHat *and* CentOS. This test is being skipped in CentOS and probably always has.

acceptance/tests/resource/service/ticket_4124_should_list_all_disabled.rb
```diff
@@ -1,12 +1,10 @@
 test_name "#4124: should list all disabled services on Redhat/CentOS"
+confine :to, :platform => /el-5/
```
whopper: Same comment as above

johnduarte: Pragmatically, 'el-5' covers all of the 'el' derivatives, since the `:platform` defines the OS apis for Beaker not necessarily the OS which is normally expressed with the `:template` that is used to deploy the VM. However, all are part of the supported whitelist for Beaker supported platforms. So, I agree that it make sense to be explicit here.

Thanks @whopper !

PR #4926, opened 2016-04-28
(PUP-5025) Fix epoch and architecture handling for RPM systems

whopper:
This commit updates logic in the RPM and Yum providers to
better handle the presence or absence of user specified epochs
and architecture fields. We now assert that absent epoch and
architecture values from the user are always insync with those
values as found by querying the system. This allows Puppet to
make the assumption that an unprovided epoch value can match
anything available on the system, not just the former default
of '0'.

In addition, this change also enables the detection of an
architecture in the title of a package and moves it to the
end of the version string for installation. Thanks to
Peter Souter (petems) for the work on this!

Note that there are still limitations based on the underlying
package manager when installing packages. DNF (a child of the
yum provider), for example requires the epoch to be specified
when a package includes one due to a bug. In yum, if an epoch is
specified then the architecture must also be specified.

This commit also refactors the Yum provider slightly to move
some of the more primitive version comparison methods up into
the parent RPM provider.

The second commit adds new acceptance test cases to the existing Yum
test.

### DIFF
--- a/acceptance/lib/puppet/acceptance/rpm_util.rb
+++ b/acceptance/lib/puppet/acceptance/rpm_util.rb
@@ -70,6 +70,7 @@ Summary: A very simple toy bin rpm package
 Name: #{o[:pkg]}
 Version: #{o[:version]}
 Release: 1
+Epoch: #{o[:epoch] || 0}
 License: GPL+
 Group: Development/Tools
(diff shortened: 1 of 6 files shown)


whopper:
This still needs spec test updates, which @branan is working on currently.

PR #4927, opened 2016-04-29
(PUP-6221) Add Semantic versioning to the Puppet Type System

thallgren:
This PR adds the two types `SemVer` and `SemVerRange` to the Puppet
Type System. The `SemVer` is a parameterized type and it's arguments can
be either strings or `SemVerRange` instances.

A `SemVer` type is constrained by it's contained `SemVerRanges`. If it has
no ranges, then the type represents all `SemVers`. If it has ranges, then
the type represents all types that matches any of the ranges.

The ranges of a `SemVer` type will be normalized so that any range that
intersects with another range automatically forms a new range. This is
needed to get the type arithmetic working correctly in cases like:
```
$x = SemVer['>=1.0.0 <2.0.0', '>=1.5.0 <3.0.0']
$y = SemVer['>=1.2.0 <2.8.0']
$x > $y == true
```


hlindberg:
Waiting on reviewing until travis passes this for Ruby 2.3.0

hlindberg:
Thomas, did you avoid tests that triggers the faulty behavior in the semver implementation in puppet?

thallgren:
I added a maint commit that reverses the failing comparison to circumvent the problem with SemVer#== on Ruby 2.3.0. It uses:
```
:undef == x
```
instead of
```
x == :undef
```
While it certainly doesn't resolve the actual problem, it does unblock this PR.

lib/puppet/pops/types/p_sem_ver_range_type.rb
```diff
@@ -0,0 +1,155 @@
...
+      # an array of ranges that may be disparate. The {{Semantic::VersionRange}} inherits
+      # from the standard ruby range. It must be possible to describe that range in terms
+      # of min, max, and exclude max.
+      #
+      # The Puppet Version type is parameterized and accepts multiple ranges so creating such
+      # constraints is still possible. It will just require several parameters rather than one
+      # parameter containing the '||' operator.
+      #
+      dispatch :from_string do
+        param "Pattern[/#{range_expr}/]", :str
```
hlindberg: This Pattern type should be a local type alias rather than refer to the `range_expr` via interpolation.  OTOH, because of the complex regexp, doesn't this result in a horrible barf of regexp if the given string does not match? Would it be better to accept any string and perform a manual assertion in the called method with a different kinder error message? (Not a valid semver range). Ideally a check would point at exactly where the semver range is wrong and why (but that is overkill).

lib/puppet/pops/types/p_sem_ver_range_type.rb
```diff
@@ -0,0 +1,155 @@
...
+      #
+      # The logical or || operator is not implemented since it effectively builds
+      # an array of ranges that may be disparate. The {{Semantic::VersionRange}} inherits
+      # from the standard ruby range. It must be possible to describe that range in terms
+      # of min, max, and exclude max.
+      #
+      # The Puppet Version type is parameterized and accepts multiple ranges so creating such
+      # constraints is still possible. It will just require several parameters rather than one
+      # parameter containing the '||' operator.
+      #
```
hlindberg: Since the interface to this `new` is in the new function  - the documentation needs to go there until we have a better way of documenting `new` per type.

lib/puppet/pops/types/p_sem_ver_type.rb
```diff
@@ -0,0 +1,121 @@
...
+  end
+
+  # @api private
+  def self.version_pattern
+    nr = '(?<nr>0|[1-9][0-9]*)'
+    "#{nr}\\.\\g<nr>\\.\\g<nr>(?:-#{parts_pattern})?(?:\\+\\g<parts>)?"
+  end
+
+  # @api private
+  def self.new_function(_, loader)
```
hlindberg: The documentation for this needs to be in the `new` function as that is the user facing function atm. We have not other way (yet) to generate documentation for the operations available on types.

lib/puppet/pops/types/p_sem_ver_type.rb
```diff
@@ -0,0 +1,121 @@
...
+    version_expr = "\\A#{version_pattern}\\Z"
+    parts_expr = "\\A#{parts_pattern}\\Z"
+    @@new_function ||= Puppet::Functions.create_loaded_function(:new_Version, loader) do
+      local_types do
+        type 'Unsigned = Integer[0,default]'
+        type "Qualifier = Pattern[/#{parts_expr}/]"
+      end
+
+      dispatch :from_string do
+        param "Pattern[/#{version_expr}/]", :str
```
hlindberg: why isn't this a local type alias ?

lib/puppet/pops/types/p_sem_ver_type.rb
```diff
@@ -0,0 +1,121 @@
...
+
+  # @api private
+  def self.new_function(_, loader)
+    version_expr = "\\A#{version_pattern}\\Z"
+    parts_expr = "\\A#{parts_pattern}\\Z"
+    @@new_function ||= Puppet::Functions.create_loaded_function(:new_Version, loader) do
+      local_types do
+        type 'Unsigned = Integer[0,default]'
+        type "Qualifier = Pattern[/#{parts_expr}/]"
+      end
```
hlindberg: Given that that these type names are shown in errors they should be more descriptive.

spec/unit/pops/types/p_sem_ver_type_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+        [ '~1.2',          '1.2.3' ] => true,
+        [ '~1',            '1.2.3' ] => true,
+        [ '1.2.x',         '1.2.3' ] => true,
+        [ '1.x',           '1.2.3' ] => true,
+
+        [ '1.2.3',         '1.2.4' ] => false,
+        [ '>=1.2.3',       '1.2.4' ] => true,
+        [ '<=1.2.3',       '1.2.4' ] => false,
+        [ '1.2.3 - 1.2.4', '1.2.4' ] => true,
+#        [ '~1.2.3',        '1.2.4' ] => true, Awaits fix for PUP-6242
```
hlindberg: Ideally this should be a separate pending test (but we have so many that just scroll by that people have stopped caring)

lib/puppet/pops/types/p_sem_ver_range_type.rb
thallgren: This begs the question, should it be possible to annotate complex types with a formatted error message that could override what's normally emitted by the TypeMismatchDescriber?

Why would you prefer a local type alias here? The type is only used in one single place.

lib/puppet/pops/types/p_sem_ver_type.rb
thallgren: Why would it be?

lib/puppet/pops/types/p_sem_ver_range_type.rb
thallgren: Not applicable to this specific case, but in general. Relaxing the type here and perform manual assertion later will not work in cases where the type matching controls the selection of the dispatcher unless you do that too manually. This also speaks for being able to annotate types with descriptive text.

lib/puppet/pops/types/p_sem_ver_range_type.rb
thallgren: Before we move this into the `new` function I think the documentation already present there must be reorganized and have a structure that will allow easy overview of what types that can be created and how. At present it's a just a long list of "conversions" with no index, and just patching more text onto that doesn't strike me as a solution that will be helpful to the user. Reorging and preparing that doc to be extended is beyond the scope of this PR. I therefor suggest that we leave the doc for the SemVer and SemVerRange where it is right now and instead create a separate ticket on how to improve the documentation of the `new` function to make it more extendable.

Side note:
I think the term "conversion" is fine when there's one argument, or perhaps one argument with an additional argument that says how the first argument should be interpreted (like Integer('0xFF', 16)) but it doesn't fit the bill for expressions like these:
```
$v1 = Version.new(1, 2, 3)
$v2 = Version.new(3, 2, 0, 'beta')
$range = VersionRange.new($v1, $v2, false)
```

hlindberg: Well, you are converting from something that is not a Version to something that is ;-) 
Jokes aside - agree about separate ticket for improving the documentation of `new` - although stating something simple as placeholder there about Version and Version range does not harm, lest we forget / run out of time.

PR #4928, opened 2016-04-30

jearls:
is either a string of the form "NAME=VALUE" or an array of such strings.

The Puppet::Provider::Package is modified to include wrapped versions
of the `execute`, `execpipe`, and `has_command` methods which include
the custom environment if the calling resource defines it.  This
includes a required modification to the `CommandDefiner` class in
Puppet::Provider to allow the executor and resolver to be overridden.

All package providers are modified to use the Puppet::Provider::Package
versions of `execute`, `execpipe`, and `has_command`.

The `yum` package provider is updated to actually support the
`environment` parameter and `:settable_environment` feature.

`environment` parameter.

### DIFF
--- a/lib/puppet/provider.rb
+++ b/lib/puppet/provider.rb
@@ -239,6 +239,8 @@ class Puppet::Provider
       @optional = false
       @confiner = confiner
       @custom_environment = {}
+      @command_resolver = Puppet::Util
+      @command_executor = Puppet::Util::Execution
     end
 
     def is_optional
@@ -249,12 +251,20 @@ class Puppet::Provider
       @custom_environment = @custom_environment.merge(env)
     end
 
+    def resolver(res)
+      @command_resolver = res
+    end
+
+    def executor(exec)
+      @command_executor = exec
+    end
+
     def command
       if not @optional
         @confiner.confine :exists => @path, :for_binary => true
       end
 
-      Puppet::Provider::Command.new(@name, @path, Puppet::Util, Puppet::Util::Execution, { :failonfail => true, :combine => true, :custom_environment => @custom_environment })
+      Puppet::Provider::Command.new(@name, @path, @command_resolver, @command_executor, { :failonfail => true, :combine => true, :custom_environment => @custom_environment })
     end
   end
 
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -9,7 +9,7 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
   These options should be specified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}),
   or an array where each element is either a string or a hash."
 
-  has_feature :install_options, :versionable, :virtual_packages
+  has_feature :install_options, :versionable, :virtual_packages, :settable_environment
 
   commands :cmd => "yum", :rpm => "rpm"
 
@@ -42,14 +42,15 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
   # @param enablerepo [Array<String>] A list of repositories to enable for this query
   # @param disablerepo [Array<String>] A list of repositories to disable for this query
   # @param disableexcludes [Array<String>] A list of repository excludes to disable for this query
+  # @param environment [Hash<String,String>] A hash of environment variables to use
   # @return [Hash<Symbol, String>]
-  def self.latest_package_version(package, enablerepo, disablerepo, disableexcludes)
+  def self.latest_package_version(package, enablerepo, disablerepo, disableexcludes, environment=nil)
 
     key = [enablerepo, disablerepo, disableexcludes]
 
     @latest_versions ||= {}
     if @latest_versions[key].nil?
-      @latest_versions[key] = check_updates(enablerepo, disablerepo, disableexcludes)
+      @latest_versions[key] = check_updates(enablerepo, disablerepo, disableexcludes, environment)
     end
 
     if @latest_versions[key][package]
@@ -64,15 +65,20 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
   # @param enablerepo [Array<String>] A list of repositories to enable for this query
   # @param disablerepo [Array<String>] A list of repositories to disable for this query
   # @param disableexcludes [Array<String>] A list of repository excludes to disable for this query
+  # @param environment [Hash<String,String>] An optional hash of environment variables to apply
   # @return [Hash<String, Array<Hash<String, String>>>] All packages that were
   #   found with a list of found versions for each package.
-  def self.check_updates(enablerepo, disablerepo, disableexcludes)
+  def self.check_updates(enablerepo, disablerepo, disableexcludes, environment=nil)
     args = [command(:cmd), 'check-update']
     args.concat(enablerepo.map { |repo| ["--enablerepo=#{repo}"] }.flatten)
     args.concat(disablerepo.map { |repo| ["--disablerepo=#{repo}"] }.flatten)
     args.concat(disableexcludes.map { |repo| ["--disableexcludes=#{repo}"] }.flatten)
 
-    output = Puppet::Util::Execution.execute(args, :failonfail => false, :combine => false)
+    if environment.nil?
+      output = Puppet::Util::Execution.execute(args, :failonfail => false, :combine => false)
+    else
+      output = Puppet::Util::Execution.execute(args, :failonfail => false, :combine => false, :custom_environment => environment)
+    end
 
     updates = {}
     if output.exitstatus == 100
@@ -213,7 +219,7 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
 
   # What's the latest package version available?
   def latest
-    upd = self.class.latest_package_version(@resource[:name], enablerepo, disablerepo, disableexcludes)
+    upd = self.class.latest_package_version(@resource[:name], enablerepo, disablerepo, disableexcludes, @resource[:environment])
     unless upd.nil?
       # FIXME: there could be more than one update for a package
       # because of multiarch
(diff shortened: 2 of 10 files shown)


jearls:
Note - this PR still only provides the `:settable_environment` feature on the `yum` provider, because I don't have the facilities or knowledge to properly test this out with the other package providers.  But theoretically all that should be required is to include the `:settable_environment` feature in a provider, and ensure it doesn't make any direct calls to `Puppet::Util::Execution` .

jearls:
this replaces pr #4862

HAIL9000:
Hey @jearls, thank you for the contribution! We're going to try and pull this into our next sprint so we can get some eyes on it and hopefully get it merged.

lib/puppet/type/package.rb
```diff
@@ -296,6 +298,61 @@ def self.title_patterns
...
+    #
+    # I am skipping the :required_features in the declaration, since
+    # that causes the parameter to be [mostly] silently ignored.  I will
+    # enforce the existence of the :settable_environment feature during
+    # validation, instead.
+    #   newparam(:environment, :required_features=>:settable_environment) do
+    newparam(:environment) do
+      desc "Any additional environment variables you want to set for operations
+        against this package.  The value should be a single string of the form
+        `NAME=value` or an array of such strings.
```
branan: This makes me sad, but I agree that consistency with `exec` is good here :(

lib/puppet/type/package.rb
```diff
@@ -296,6 +298,61 @@ def self.title_patterns
...
+
+        An error will be raised if this parameter is used with a provider that
+        does not support it."
+
+      validate do |values|
+        # enforce provider feature here
+        unless provider.satisfies?([:settable_environment])
+          raise ArgumentError, "environment is not supported by the #{provider.class.name} provider"
+        end
+        values = [values] unless values.is_a? Array
```
branan: I tend to prefer the `values = [values].flatten` format, so that we're not tied to objects being specific ruby types in the future.

lib/puppet/type/package.rb
```diff
@@ -296,6 +298,61 @@ def self.title_patterns
...
+        environment = {}
+
+        if envlist
+          envlist = [envlist] unless envlist.is_a? Array
+          envlist.each do |setting|
+            if setting =~ /^(\w+)=((.|\n)+)$/
+              env_name = $1
+              value = $2
+              if environment.include?(env_name) || environment.include?(env_name.to_sym)
+                warning "Overriding environment setting '#{env_name}' with '#{value}'"
```
branan: I'm not sure this deserves warning level.

It should also probably be clearer about the fact that it is a duplicate key from the same parameter, and isn't related to anything system-level

lib/puppet/type/package.rb
```diff
@@ -296,6 +298,61 @@ def self.title_patterns
...
+      munge do |envlist|
+        environment = {}
+
+        if envlist
+          envlist = [envlist] unless envlist.is_a? Array
+          envlist.each do |setting|
+            if setting =~ /^(\w+)=((.|\n)+)$/
+              env_name = $1
+              value = $2
+              if environment.include?(env_name) || environment.include?(env_name.to_sym)
```
branan: `environment` shouldn't have any symbol keys, based on my reading of the code here. Why this check?

lib/puppet/type/package.rb
```diff
@@ -296,6 +298,61 @@ def self.title_patterns
...
+        end
+      end
+
+      # transform environment string or array into a hash
+      # code stolen from `Puppet::Provider::Exec`
+      munge do |envlist|
+        environment = {}
+
+        if envlist
+          envlist = [envlist] unless envlist.is_a? Array
```
branan: Same as above regarding array-ification

lib/puppet/provider.rb
```diff
@@ -242,6 +242,8 @@ def initialize(name, path, confiner)
       @optional = false
       @confiner = confiner
       @custom_environment = {}
+      @command_resolver = Puppet::Util
+      @command_executor = Puppet::Util::Execution
```
branan: I can't seem to find where you use this change in the code pushed up here - is this leftover from an early iteration, or am I missing something?

lib/puppet/provider/package/yum.rb
```diff
@@ -64,15 +65,20 @@ def self.latest_package_version(package, enablerepo, disablerepo, disableexclude
...
   #   found with a list of found versions for each package.
-  def self.check_updates(enablerepo, disablerepo, disableexcludes)
+  def self.check_updates(enablerepo, disablerepo, disableexcludes, environment=nil)
     args = [command(:cmd), 'check-update']
     args.concat(enablerepo.map { |repo| ["--enablerepo=#{repo}"] }.flatten)
     args.concat(disablerepo.map { |repo| ["--disablerepo=#{repo}"] }.flatten)
     args.concat(disableexcludes.map { |repo| ["--disableexcludes=#{repo}"] }.flatten)
 
-    output = Puppet::Util::Execution.execute(args, :failonfail => false, :combine => false)
+    if environment.nil?
```
branan: Does it make more sense to pass around an empty hash as your default value instead of `nil`, so that this check isn't needed?

lib/puppet/provider/package/yum.rb
```diff
@@ -64,15 +65,20 @@ def self.latest_package_version(package, enablerepo, disablerepo, disableexclude
...
+      output = Puppet::Util::Execution.execute(args, :failonfail => false, :combine => false)
+    else
+      output = Puppet::Util::Execution.execute(args, :failonfail => false, :combine => false, :custom_environment => environment)
```
branan: This is still using `Puppet::Util::Execution.execute` instead of the helper you added to the base package provider. Is there a reason for that?

branan:
I don't really like moving all the providers to use a common helper if you're not going to enable the `:settable_environment` feature on them - it seems like it's just complexity and code churn without much gain in those cases

lib/puppet/type/package.rb
jearls: As I mentioned in the comments, I lifted that code straight out of the `exec` provider.   It appears that I removed one line (which was setting `environment[:PATH]`), and removing that line obviates the need for the symbol check.

lib/puppet/type/package.rb
jearls: Once again, this code came from `exec`.  I can remove the check entirely, or I can change the warning to some other notification type (if you tell me what notification function to use, since i don't know the puppet internals) and reword it to indicate that it's overriding a "previously-set" environment setting.

lib/puppet/provider.rb
jearls: The `@command_resolver` and `@command_executor` are passed to the `Puppet::Provider::Command` instantiation on line 278.

lib/puppet/provider/package/yum.rb
jearls: I wasn't sure if passing `custom_environment = {}` to `Puppet::Util::Execution.execute` was the same as _not_ passing `custom_environment` at all.

jearls: More to the point, I'm still not sure if it will always in the future be the same - for example, if the default value for `custom_environment` gets changed for some reason.

lib/puppet/provider/package/yum.rb
jearls: Because the helper in the base package provider is an instance method, and `check_updates` is a class method.

jearls:
The reason I didn't add the custom environment to any other provider is because I have no expertise in any other packaging system and no experience with non-RedHat Linux OSes.  I originally had just added this to only to the `yum` package type, and was requested to make it more generally available since the original request _was_ for a different packaging system.  However, because I don't know how those other packaging systems work and have no real way to test them, I chose to create the framework and allow someone who knows what they're doing to implement the feature in the other packaging systems.

lib/puppet/type/package.rb
jearls: Yes, I had originally made this accept a hash.  I guess I could try to let it accept either, but I thought that would be more confusing than just keeping it consistent.

branan:
@jearls That makes sense. This is a useful feature for us to get in, and the work you've done building a framework for the other providers looks like it's in the right direction to me, so I'm not going to push for its removal.

It looks like this has picked up some conflicts with the base branch in the interim. If you could get it rebased that'd be awesome.  You'll also probably have to squash your second commit in to your first, or at the very least add the ticket number to it, in order to make our commit message checker happy.

darkfoxprime:
Sorry it's taken so long.  I completely missed the update from July.  I just rebased and force-pushed, but I don't know if I've got the spec file correct, since I haven't seen/used the "shared examples" thing before.

prometheanfire:
once this merges I'll update the portage package provider, this is useful for overriding FEATURES and the like if needed.

lib/puppet/provider.rb
```diff
@@ -239,6 +239,8 @@ def initialize(name, path, confiner)
       @optional = false
       @confiner = confiner
       @custom_environment = {}
+      @command_resolver = Puppet::Util
```
joshcooper: The `Puppet::Util` module is [mixed into the provider](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider.rb#L38), so I don't think this is necessary.

lib/puppet/provider.rb
```diff
@@ -249,12 +251,20 @@ def environment(env)
...
+      @command_executor = exec
+    end
+
     def command
       if not @optional
         @confiner.confine :exists => @path, :for_binary => true
       end
 
-      Puppet::Provider::Command.new(@name, @path, Puppet::Util, Puppet::Util::Execution, { :failonfail => true, :combine => true, :custom_environment => @custom_environment })
+      Puppet::Provider::Command.new(@name, @path, @command_resolver, @command_executor, { :failonfail => true, :combine => true, :custom_environment => @custom_environment })
```
joshcooper: Similarly, I think you can just pass `self` here, e.g. `Command.new(@name, @path, self, ...)`

MosesMendoza:
This PR hasn't been updated in some time, and I'm afraid the team won't have time to review it in the near future. Let's track future conversation in the ticket, PUP-1526, and we can re-open this PR if/when there's bandwidth to support it. Closing for now.

rledousa:
I commented into PUP-1526.  We can definitely use this feature to work around a bug in pip3.6

PR #4930, opened 2016-05-02
(PUP-6072) Make --lastrunreport be a multi-file-style argument

MikaelSmith:
file-type argument to being a multi-file-type argument.

commit fb252414284a7e31d1290fddcbec155599bc296f


MikaelSmith:
Still working on tests.

MikaelSmith:
Need to fix up the test, but this works. Myonly concern is that it'll break any modules that use `lastrunreport` directly.

PR #4938, opened 2016-05-03
(PUP-6259) Fix no_proxy support for File resource

patcable:
File resources now have the ability to pull from an http/https source. However, the `no_proxy` variable doesn't seem to be respected. I traced this to behavior related to the `Net::HTTP::Proxy` object that seems to have been deprecated. Replacing it with `Net::HTTP.new` seems to work

### DIFF
--- a/lib/puppet/util/http_proxy.rb
+++ b/lib/puppet/util/http_proxy.rb
@@ -4,12 +4,10 @@ require 'openssl'
 module Puppet::Util::HttpProxy
   def self.proxy(uri)
     if self.no_proxy?(uri)
-      proxy_class = Net::HTTP::Proxy(nil)
+      return Net::HTTP.new(uri.host, uri.port, nil)
     else
-      proxy_class = Net::HTTP::Proxy(self.http_proxy_host, self.http_proxy_port, self.http_proxy_user, self.http_proxy_password)
+      return Net::HTTP.new(uri.host, uri.port, self.http_proxy_host, self.http_proxy_port, self.http_proxy_user, self.http_proxy_password)
     end
-
-    return proxy_class.new(uri.host, uri.port)
   end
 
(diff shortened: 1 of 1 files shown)


patcable:
Well, this was more complicated than I had anticipated 😄  -- the tests fail in `repository_spec.rb`.

It seems like some module is parsing the port as `[port]the_path`, which is weird. As far as I can tell, the URI that's being sent is http://fake.comthepath instead of http://fake.com/the_path. I tried changing the paths around, but then VCR complained that it didn't know how to replay the request.

I'm sure someone with more in-depth rspec/codebase knowledge can clue me in on what's going on here. It's also possible that there may need to be two behaviors: the old method for rubies <2.0 and this for rubies >2.0

branan:
We still (unfortunately :( ) support Ruby 1.9 in Puppet 4, so if the new code doesn't work there we can't take it (unless there's a compatibility path).

As far as the VCR stuff, that was all done by @ffrank. Felix, if you have the time, do you mind helping out here?

ffrank:
Sure, sounds fun :)

Before we dig into this - @patcable could you update the ticket with information? The file resource should be using the same HTTP client code as the PMT (i.e. `puppet module install`). Is that afflicted as well?

patcable:
@ffrank sure thing. I copied that comment over there. 

Yes, it would seem that puppet module install breaks with my fix (as evidenced by the tests). So there's something odd about how no_proxy is handled for that code vs the File resource.

MosesMendoza:
This PR has not been updated in some time and is now being closed due to inactivity. Please re-open if this is in error - apologies in advance for any mistake.

thiagofigueiro:
Is ruby 1.9 still supported? If not, perhaps this could be merged now?

PR #4941, opened 2016-05-05
(PUP-6046) Add functions 'dig', 'then', and 'lest'

hlindberg:
This adds three functions 'dig', 'then', and 'lest', which together with the existing `assert_type` function makes complex data access easier to perform without complex conditional logic and temporary variables.

The *dig* function performs a sequence of access operations on a complex collection similar to the 'dot-notation' used in hiera and facter.
For example:
```
$some_data.dig(somekey, 1, otherkey)
```
If, at any point, the intermediate result is `undef` the operation stops and returns `undef`. It is an error to attempt to access something in a value that is neither Array nor Hash.

The *then* function, accepts a single argument, and if it is not `undef` if calls a given block with the value. If given `undef` the block is not called and `undef` is returned.

The *lest* function is the reverse of *then*. If given `undef` it calls the given block (without arguments), and returns what the block produced. If given something that is not `undef` it returns that value.

```
$some_structure.process()
  .then |$x| { $x.process_some_more() }
  .then |$x| { $x.process_further() }
  .lest || { fail("processing of structure resulted in undef value") }
```

Or if a default value is wanted - in the middle of the chain if `undef` was the temporary result:
```
$some_structure.process()
  .then |$x| { $x. process_some_more() }
  .lest || { 42 }
  .then |$x| { $x.process_further() }
```

Naturally, for simple cases, a regular if/then/else expression or case expression is clearer - but for any chain of lookups, transformations, etc. where intermediate results are obtained and those can be undef the functional composition chain is a valuable tool as it results in more readable logic.

### DIFF
--- a/lib/puppet/functions/dig.rb
+++ b/lib/puppet/functions/dig.rb
@@ -3,7 +3,7 @@
 # using a sequence of keys / indexes to access a value from which
 # the next key/index is accessed recursively.
 #
-# The first encountered `undef` value stops the "dig" and `undef` is returned.
+# The first encountered `undef` value or key stops the "dig" and `undef` is returned.
 #
 # An error is raised if an attempt is made to "dig" into
 # something other than an `undef` (which immediately returns `undef`), an Array or a Hash. 
@@ -28,7 +28,7 @@ Puppet::Functions.create_function(:dig) do
   def dig(data, *args)
     walked_path = []
     args.reduce(data) do | d, k |
-      return nil if d.nil?
+      return nil if d.nil? || k.nil?
       if !(d.is_a?(Array) || d.is_a?(Hash))
         raise ArgumentError, "The given data does not contain a Collection at #{walked_path}, got '#{d.class}'"
       end
--- a/lib/puppet/functions/dig.rb
+++ b/lib/puppet/functions/dig.rb
@@ -1,8 +1,12 @@
-# Returns a value for a sequence of given keys/indexes
-# used to "dig into" a complex data structure. The first
-# encountered undef value stops the "dig" and undef is returned.
+# Returns a value for a sequence of given keys/indexes into a structure.
+# This function is used to "dig into" a complex data structure by
+# using a sequence of keys / indexes to access a value from which
+# the next key/index is accessed recursively.
+#
+# The first encountered `undef` value stops the "dig" and `undef` is returned.
+#
 # An error is raised if an attempt is made to "dig" into
-# something other than an Array or a Hash. 
+# something other than an `undef` (which immediately returns `undef`), an Array or a Hash. 
 #
 # @example Using `dig`
 #
--- a/lib/puppet/functions/dig.rb
+++ b/lib/puppet/functions/dig.rb
@@ -1,0 +1,35 @@
+# Returns a value for a sequence of given keys/indexes
+# used to "dig into" a complex data structure. The first
+# encountered undef value stops the "dig" and undef is returned.
+# An error is raised if an attempt is made to "dig" into
+# something other than an Array or a Hash. 
+#
+# @example Using `dig`
+#
+# ~~~ puppet
+# $data = {a => { b => [{x => 10, y => 20}, {x => 100, y => 200}]}}
+# notice $data.dig(a, b, 1, x)
+# ~~~
+#
+# Would notice the value 100.
+#
+# @since 4.5.0
+#
+Puppet::Functions.create_function(:dig) do
+  dispatch :dig do
+    param 'Collection', :data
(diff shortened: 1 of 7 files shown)


lib/puppet/functions/lest.rb
```diff
@@ -0,0 +1,53 @@
+# Call a [lambda](https://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html)
```
thallgren: Use puppet.com.

lib/puppet/functions/then.rb
```diff
@@ -0,0 +1,74 @@
+# Call a [lambda](https://docs.puppetlabs.com/puppet/latest/reference/lang_lambdas.html)
```
thallgren: Use puppet.com

spec/unit/functions/dig_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  include PuppetSpec::Compiler
+  include Matchers::Resource
+
+  it 'returns a value from an array index via integer index' do
+    expect(compile_to_catalog("notify { [testing].dig(0): }")).to have_resource('Notify[testing]')
+  end
+
+  it 'returns undef if given an undef key' do
+  expect(compile_to_catalog(<<-SOURCE)).to have_resource('Notify[testing]')
+    notify { "test${[testing].dig(undef)}ing": }
```
thallgren: Nitpick, but you don't know if the return is really an undef or something else that represents itself as an empty string (applies to several tests).

lib/puppet/functions/dig.rb
```diff
@@ -0,0 +1,39 @@
+# Returns a value for a sequence of given keys/indexes into a structure.
+# This function is used to "dig into" a complex data structure by
+# using a sequence of keys / indexes to access a value from which
+# the next key/index is accessed recursively.
+#
+# The first encountered `undef` value or key stops the "dig" and `undef` is returned.
+#
+# An error is raised if an attempt is made to "dig" into
+# something other than an `undef` (which immediately returns `undef`), an Array or a Hash.
```
thallgren: What's your though on accessing object attributes? I can see how it might introduce ambiguities.

lib/puppet/functions/lest.rb
```diff
@@ -0,0 +1,53 @@
...
+#
+Puppet::Functions.create_function(:lest) do
+  dispatch :lest do
+    param 'Any', :arg
+    block_param 'Callable[0,0]', :block
+  end
+
+  def lest(arg)
+    if arg.nil?
+      yield()
```
thallgren: No need for parenthesis unless there are arguments.

spec/unit/functions/then_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    end.to raise_error(/'then' block expects 1 argument, got 2/m)
+  end
+
+  it 'errors when lambda wants too few args' do
+    expect do
+      compile_to_catalog('then(1) || { }')
+    end.to raise_error(/'then' block expects 1 argument, got none/m)
+  end
+
+end
```
thallgren: Although perhaps not strictly necessary since each function is tested, it would be nice to see the examples from the PR-comment as unit tests where everything is combined.

lib/puppet/functions/lest.rb
hlindberg: I may be wrong, but I though yield without args yielded the same args as the method got - but I may have confused that with a call to super. So () here to be explicit that no args are passed.

spec/unit/functions/then_spec.rb
hlindberg: That would be some kind of integration test or acceptance test as it involves user facing functionality / support for a use case. The same goes for many other functions that are used together.

lib/puppet/functions/dig.rb
hlindberg: Support for that is ok to add. If it is an Object, then if the key is an attribute - proceed if not return undef. If attribute exists and is not a Collection, or Object (or undef) error. Would work fine I think. (Needs a separate ticket "Add dig support to Objects").

lib/puppet/functions/lest.rb
hlindberg: I can remove them, no problem, I just wanted to be explicit. So, should I change it?

spec/unit/functions/then_spec.rb
thallgren: This PR provides three functions intended to be used as a concept and hence provided in one single PR. The PR-comment even shows the intended use in action. I think that's enough to motivate a unit test, but as I wrote, not strictly necessary.

PR #4942, opened 2016-05-05
(PUP-6269) Backport PUP-3341 to 3.x (fix, apply.rb only)

robmbrooks:
apply.rb changes from this commit:

https://github.com/puppetlabs/puppet/commit/04634314408fa227220a8c6d0429d419c1754a4d

### DIFF
--- a/lib/puppet/application/apply.rb
+++ b/lib/puppet/application/apply.rb
@@ -190,17 +190,25 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
       facts.name = Puppet[:node_name_value]
     end
 
-    configured_environment = Puppet.lookup(:current_environment)
+    # Find our Node
+    unless node = Puppet::Node.indirection.find(Puppet[:node_name_value])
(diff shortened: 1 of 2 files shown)


robmbrooks:
This is exempt from CLA as it is Felix Frank's code, already commited in the 4.x branch.

hlindberg:
Two things:
Can you update the commit message so it reads:
```
(PUP-6269) Backport PUP-3341 to 3.x
```
(I fixed the title of this PR already)
We use the tags in the commit message to correlate commits with tickets when we do a release - hence being picky.

Not sure how to deal with the CLA; clearly Felix Frank authored the fix, but it does not show in the commit. We may need to patch that up unless you sign the CLA (which is a simple process).

robmbrooks:
I have backported the tests too now, so I have...

Finished in 4 minutes 13 seconds
21477 examples, 0 failures, 126 pending

Would you like a single commit with all the changes? Should I just make a new pull request?

robmbrooks:
This looks like a mess, I will make a pull request with a single correctly named commit.

PR #4943, opened 2016-05-05
(PUP-6159) Convert CFPropertyList binary strings to CFPropertyList::Blob's

whopper:
Previously, Puppet would fail to set passwords in OSX via the
directoryservice provider due to CFPropertyList encoding problems
when converting entropy and salt strings into binary.

This was due to the fact that CFPropertyList was treating these
binary fields as CFStrings, which would then fail to
convert to binary due to invalid UTF-8 byte sequences.

In order to prevent this, the 'salt' and 'entropy' fields must
be converted into CFPropertyList::Blob's, which are interpreted
as CFPropertyList::CFData objects when parsed by CFPropertyList.
This is the correct data type for binary data values. (Thanks to 
@branan  for the help tracking this down!)

In addition, when setting shadow hash data, the binary_plist
string must also be converted into a Blob for the same reasons.

The second commit adds a simple acceptance test to verify the
ability to set passwords on the system.

### DIFF
--- a/acceptance/tests/resource/user/should_allow_password_salt_iterations_osx.rb
+++ b/acceptance/tests/resource/user/should_allow_password_salt_iterations_osx.rb
@@ -1,0 +1,26 @@
+test_name "should allow password, salt, and iteration attributes in OSX"
+
+confine :to, :platform => /osx/
+
+agents.each do |agent|
+  teardown do
+    puppet_resource("user", 'testuser', "ensure=absent")
+  end
+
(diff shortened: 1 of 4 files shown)


branan:
We could probably write an acceptance test that does what we did on your laptop - create a user, then try to `su` to it. I don't think there's an easy unit test

whopper:
@branan 👍  I like it - I'll whip one up.

PR #4944, opened 2016-05-05
(PUP-6261) Increase timeout for scheduled tasks deletion in test

glennsarti:
It is possible that due to contention in a VM that it will take longer than
5 seconds for a scheduled task to appear as deleted.  This commit increases the
timeout to a large value to give the operating system time to delete the task


johnduarte:
:+1:  LGTM
Thanks @glennsarti

PR #4945, opened 2016-05-06
(PUP-6269) Backport PUP-3341 to 3.x

robmbrooks:
[PUP-6269](https://tickets.puppetlabs.com/browse/PUP-6269) backporting fix from [PUP-3341](https://tickets.puppetlabs.com/browse/PUP-3341) to 3.x branch

### DIFF
--- a/lib/puppet/application/apply.rb
+++ b/lib/puppet/application/apply.rb
@@ -190,17 +190,25 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
       facts.name = Puppet[:node_name_value]
     end
 
-    configured_environment = Puppet.lookup(:current_environment)
+    # Find our Node
+    unless node = Puppet::Node.indirection.find(Puppet[:node_name_value])
+      raise "Could not find node #{Puppet[:node_name_value]}"
+    end
+
+    configured_environment = node.environment || Puppet.lookup(:current_environment)
+
     apply_environment = manifest ?
       configured_environment.override_with(:manifest => manifest) :
       configured_environment
 
-    Puppet.override({:current_environment => apply_environment}, "For puppet apply") do
-      # Find our Node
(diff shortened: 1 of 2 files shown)


hlindberg:
There were test failures for Ruby 1.8.7 - respun the spec test run in case it was temporary. It may be a difference in rspec version used on 3.x and 4.x that causes this.
It helps to get it running locally on a ruby 1.8.7 to see if what is going on, and to debug the problem. Maybe original author of tests can spot the problems immediately...

robmbrooks:
I kicked up an Ubuntu 12.04 as this ships with Ruby 1.8.7 and it looks like I have the same failures, not immediately obvious to me why though.

hlindberg:
The first thing to look at is the difference in rspec version (much newer on 4.x vs 3.x). Ping @er0ck - do we need to include backported tests?

robmbrooks:
@hlindberg does the 4.x branch get tested against Ruby 1.8.7?

hlindberg:
We dropped support for Ruby 1.8.7 in 4.x. So... No 4.x testing does not run against Ruby 1.8.7. Puppet 4.x simply does not work on Ruby 1.8.7.
We also updated several gems in 4.x (rspec and mocha specifically).

robmbrooks:
@hlindberg bizarrely the inline scripts appear to have caused the failures with Ruby 1.8.7.

Fixed!

branan:
@robmbrooks This looks good to me, and I think we're nearly ready to merge it in.

One more quick change: would you mind squashing your test fixes into the backport commit to keep the history a bit cleaner? After that I can hit the big green button

robmbrooks:
@branan I think this is correct.

I rebased, squashed the two commits into one and force pushed.

robmbrooks:
@branan the requested commit squash has been done, I think "waiting on contributor" can be removed?

HAIL9000:
Merged, thank you for the contribution @robmbrooks!

PR #4947, opened 2016-05-06
[DNM] (PUP-6270) Update Acceptance tests for PUP-3827

justinstoller:
PUP-3827 updated the error messages from the indirector to return valid
JSON.

There are acceptance tests that parse error messages - assuming plain text -
to validate that improper behavior cannot be accomplished. This
relaxes the validation to allow for JSON formatted error messages.

PR #4948, opened 2016-05-07
Revert "(PUP-3827) Return errors matching schema from indirected routes"

shrug:
Reverts puppetlabs/puppet#4875

Tests failing - PUP-6270


geoffnichols:
👍

justinstoller:
👍 

The original PR has definitely caused a regression in the test suite and we've failed to get it fixed by EOD Friday. Reverting for clean builds and we will update #4947 on Monday with appropriate fixes.

PR #4949, opened 2016-05-09
(PUP-6277) Add conflicts to types

raphink:
This PR adds a new method for types, allowing to declare conflicts between resources. It implements such conflicts between the `augeas` and `file` resources (other resource types might benefit, too). Example usage:

``` puppet
augeas { "test":
  incl    => '/tmp/toto',
  lens    => 'Xml.lns',
  changes => [],
}

file { "/tmp/toto": }
```

``` shell
$ puppet apply conflict.pp
Notice: Compiled catalog for example.com in environment production in 0.10 seconds
Error: Failed to apply catalog: Resource augeas[test] conflicts with file[/tmp/toto]
```

Ideally, this would fail during catalog compilation, but I need a catalog in order to check the conflicts.

### DIFF
--- a/lib/puppet/type/augeas.rb
+++ b/lib/puppet/type/augeas.rb
@@ -219,6 +219,22 @@ Puppet::Type.newtype(:augeas) do
     end
 
     cons.flatten.uniq
+
+    files = []
+
+    catalog.resources.find do |r| 
+      next unless r.is_a?(Puppet::Type.type(:file))
+      next unless cons.include?(r.title)
+      next unless r.parameter(:content) || r.parameter(:source)
+      
+      if rep = r.parameter(:replace)
+        next if rep.value == false
+      end
+
+      files << r
+    end
+
+    files
   end
 
 end
--- a/lib/puppet/type/augeas.rb
+++ b/lib/puppet/type/augeas.rb
@@ -208,4 +208,17 @@ Puppet::Type.newtype(:augeas) do
     end
   end
 
+  conflict(:file) do
+    cons = []
+    if incl = @parameters[:incl]
+      cons << incl.value
+    end
+
+    if context = @parameters[:context]
+      cons << context.value.gsub(%r{^/files}, '')
+    end
+
+    cons.flatten.uniq
+  end
+
 end
--- a/lib/puppet/type/augeas.rb
+++ b/lib/puppet/type/augeas.rb
@@ -224,7 +224,13 @@ Puppet::Type.newtype(:augeas) do
 
     catalog.resources.find do |r| 
       next unless r.is_a?(Puppet::Type.type(:file))
-      next unless cons.include?(r.title)
+
+      if p = r.parameter(:path)
+        next unless cons.include?(p.value)
+      else
+        next unless cons.include?(r.title)
+      end
+
       next unless r.parameter(:content) || r.parameter(:source)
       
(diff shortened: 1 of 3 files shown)


domcleal:
A file resource will only conflict with an augeas resource if they're managing the same attribute, i.e. file with either content or source.  It'd be quite legitimate (and possibly common) to have augeas resources managing bits of the content, plus a file resource to manage the mode, owner etc.  I think the conflict might need to be more specific.

raphink:
@domcleal Refined the block with a few more conditions:

- file should have a `path` or `title` conflicting with augeas resource's `incl` or `context`
- file should declare a `content` or `source`
- file should not be `replace => false`

crayfishx:
Carrying on from the IRC discussion....

This still only affects things managed by the file resource type - there are plenty of ways for puppet to modify a file that isn't going to contain a `File[]` resource in the catalog... yumrepos for example, or types and providers that issue commands to applications to manage their files, such as firewalld-cmd or hostnamectl.....

Having a check for files is certainly useful, but it's not going to protect against any conflict

raphink:
@crayfishx I agree. The reason for making this generic is that it wouldn't only be used in the `augeas` resource type. Any type could declare conflicts. Also, `augeas` could conflict with `yumrepo` if necessary.

hlindberg:
Without having looked deep into this - Isn't there a potential performance problem here in large catalogs?

raphink:
Yes there could be, similar to autorequires for the user type currently...

legooolas:
Could this even (performance problems aside...) check against the file content to see if the augeas proposed change is identical to the destination file content, and not to conflict in that case?
(or does a non-static catalog not contain the file content?)

raphink:
Ideally, I'd like conflicts to fail before the catalogue is retrieved, which means the file is not accessible.

branan:
I have a couple of thoughts on this:

1) This won't catch any resources that are `eval_generate`d
2) This is an interesting feature, but I'm not sure it's going to be great to say "we added this new conflicts thing and it works for augeas/file" when we can't commit internally to expanding it to other core resources and supported modules. Puppet has plenty of half-implemented features as it is. If we had commitments from the community to start pushing more of that work forward  that might soften my feelings.

raphink:
@branan Augeas is just a demo of what that could be. If we can agree on what it should look like and how it should work, I could provide more implementations for native types.

lib/puppet/type/augeas.rb
```diff
@@ -208,4 +208,39 @@ def sync
...
+
+    if context = @parameters[:context]
+      cons << context.value.gsub(%r{^/files}, '')
+    end
+
+    cons.flatten.uniq
+
+    files = []
+
+    catalog.resources.each do |r|
```
MikaelSmith: Could you use `catalog.resource(:file).each do |r|` to narrow the list to just file types? That might allow us to later optimize creating that list, separate from using it here.

MikaelSmith:
I'd re-iterate @branan's point earlier as well. An RFC seems useful to gather input on what types might benefit from this, as well as what pitfalls we might encounter where conflicts wouldn't be appropriate (or you might want to explicitly override them).

MosesMendoza:
This PR has not been updated in some time and is now being closed due to inactivity. Please re-open it and add a comment explaining if this is in error. My apologies in advance for any mistake on this.

PR #4950, opened 2016-05-09
PUP-4403 - k5login selinux/owner

jcpunk:
This is an attempt to get PUP-4403 resolved.

Step 1, get the selinux bits (that are really really biting me) fixed.
Step 2, get the owner bits (that are biting me) fixed.

### DIFF
--- a/lib/puppet/type/k5login.rb
+++ b/lib/puppet/type/k5login.rb
@@ -1,5 +1,6 @@
 # Plug-in type for handling k5login files
 require 'puppet/util'
+require 'puppet/util/selinux'
 
 Puppet::Type.newtype(:k5login) do
   @doc = "Manage the `.k5login` file for a user.  Specify the full path to
@@ -31,9 +32,59 @@ Puppet::Type.newtype(:k5login) do
     defaultto { "644" }
   end
 
+  # To manage the selinux user of the file
+  newproperty(:seluser) do
+    desc "What the SELinux user component of the context of the file should be.
+      Any valid SELinux user component is accepted.  For example `user_u`.
+      If not specified it defaults to the value returned by matchpathcon for
+      the file, if any exists.  Only valid on systems with SELinux support
+      enabled."
+
+    defaultto { "user_u" }
+  end
+
+  # To manage the selinux role of the file
+  newproperty(:selrole) do
+    desc "What the SELinux role component of the context of the file should be.
+      Any valid SELinux role component is accepted.  For example `role_r`.
+      If not specified it defaults to the value returned by matchpathcon for
+      the file, if any exists.  Only valid on systems with SELinux support
+      enabled."
+
+    defaultto { "object_r" }
+  end
+
+  # To manage the selinux type of the file
+  newproperty(:seltype) do
+    desc "What the SELinux type component of the context of the file should be.
+      Any valid SELinux type component is accepted.  For example `tmp_t`.
+      If not specified it defaults to the value returned by matchpathcon for
+      the file, if any exists.  Only valid on systems with SELinux support
+      enabled."
+
+    # to my knowledge, `krb5_home_t` is the only valid type for .k5login
+    defaultto { "krb5_home_t" }
+  end
+
+  # To manage the selinux range of the file
+  newproperty(:selrange) do
+    desc "What the SELinux range component of the context of the file should be.
+      Any valid SELinux range component is accepted.  For example `s0` or
+      `SystemHigh`.  If not specified it defaults to the value returned by
+      matchpathcon for the file, if any exists.  Only valid on systems with
+      SELinux support enabled and that have support for MCS (Multi-Category
+      Security)."
+
+    defaultto { "s0" }
+  end
+
+  # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # 
+
   provide(:k5login) do
     desc "The k5login provider is the only provider for the k5login
       type."
+
+    include Puppet::Util::SELinux
 
     # Does this file exist?
     def exists?
@@ -76,6 +127,56 @@ Puppet::Type.newtype(:k5login) do
     # Set the file mode, converting from a string to an integer.
     def mode=(value)
       File.chmod(Integer("0#{value}"), @resource[:name])
+    end
+
+    # Set the file seluser
+    def seluser
+      if selinux_support?
+        context = get_selinux_current_context(@resource[:name])
+        return parse_selinux_context(:seluser, context)
+      end
+    end
+
+    def seluser=(value)
+      set_selinux_context(@resource[:name], value, :seluser)
+    end
+
+    # Set the file selrole
+    def selrole
+      if selinux_support?
+        context = get_selinux_current_context(@resource[:name])
+        return parse_selinux_context(:selrole, context)
+      end
+    end
+
+    # Set the file seltype
+    def selrole=(value)
+      set_selinux_context(@resource[:name], value, :selrole)
+    end
+
+    # Set the file seltype
+    def seltype
+      if selinux_support?
+        context = get_selinux_current_context(@resource[:name])
+        return parse_selinux_context(:seltype, context)
+      end
+    end
+
+    # Set the file selrange
+    def seltype=(value)
+      set_selinux_context(@resource[:name], value, :seltype)
+    end
+
+    # Set the file selrange
+    def selrange
+      if selinux_support?
+        context = get_selinux_current_context(@resource[:name])
+        return parse_selinux_context(:selrange, context)
+      end
+    end
+
+    def selrange=(value)
+      set_selinux_context(@resource[:name], value, :selrange)
     end
 
     private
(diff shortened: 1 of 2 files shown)


branan:
It looks like you added the new properties to the type, but haven't added any logic to the `provides` section to enforce them on the system. Is there something happening behind the scenes in this type that I'm missing?

jcpunk:
Alas there is nothing sneaky happening in the background.  I can't seem to figure out how the File provider is doing the work.  I'd planned to copy it from there, but I fear my ability to fake ruby isn't up to snuff.

Any chance you can point me in the right direction?  lib/puppet/type/file/selcontext.rb is up to something, but I can't figure out how to decode it to clone the behavior......

branan:
Hi @jcpunk 

Sorry I didn't get back to you sooner. This completely fell off my radar.

I would not use the file type as an example of how to do anything. It's very "special", and does lots of custom hacky things (including the "retrieve_default_context" method that you're trying to use here.

We do have documentation for [custom types](https://docs.puppet.com/guides/custom_types.html) and [their providers](https://docs.puppet.com/guides/provider_development.html). I'd start there, and implement the selinux properties yourself, instead of trying to use what's in the file type as a template.

joshcooper:
This failed Travis and Appveyor.

spec/unit/type/k5login_spec.rb
```diff
@@ -69,6 +77,18 @@ def resource(attrs = {})
           end
         end
 
+        context "with selinux" do
```
Magisus: I started trying to debug the unit tests failures on this, which at very least are failing due to some issues with the mocking. I was able to get the test mostly working by restructuring it like so:
```
        context "with selinux" do
          Puppet::Util::SELinux.stubs(:selinux_support?).returns true
          Puppet::Util::SELinux.stubs(:get_selinux_current_context).returns "user_u:role_r:type_t:s0"
          subject { resource(:content => "daniel@EXAMPLE.COM\n",).retrieve }
          it "should return correct values based on SELinux state" do
            expect(subject[:seluser]).to eq("user_u")
            expect(subject[:selrole]).to eq("role_r")
            expect(subject[:seltype]).to eq("type_t")
            expect(subject[:selrange]).to eq("s0")
          end
        end
```

The issue now is this apparently doesn't correctly stub `selinux_support?`, it's still false. Not sure why.

lib/puppet/type/k5login.rb
```diff
@@ -78,6 +128,54 @@ def mode=(value)
       File.chmod(Integer("0#{value}"), @resource[:name])
     end
 
+    # Set the file seluser
+    def seluser
+      context = get_selinux_current_context(@resource[:name])
```
Magisus: You are missing a call to `selinux_support?` here.

lib/puppet/type/k5login.rb
```diff
@@ -78,6 +128,54 @@ def mode=(value)
...
+      return context[:seluser]
```
Magisus: `context` here is a string, not a hash. You need to run this through `parse_selinux_context`:
```
return parse_selinux_context(:seluser, context)
```

lib/puppet/type/k5login.rb
```diff
@@ -78,6 +128,54 @@ def mode=(value)
...
+
+    def seluser=(value)
+      set_selinux_context(@resource[:name], value, :seluser)
+    end
+
+    # Set the file selrole
+    def selrole
+      if selinux_support?
+        context = get_selinux_current_context(@resource[:name])
+        return context[:selrole]
```
Magisus: Same here, need `parse_selinux_context`.

lib/puppet/type/k5login.rb
```diff
@@ -78,6 +128,54 @@ def mode=(value)
...
+    # Set the file seltype
+    def selrole=(value)
+      set_selinux_context(@resource[:name], value, :selrole)
+    end
+
+    # Set the file seltype
+    def seltype
+      if selinux_support?
+        context = get_selinux_current_context(@resource[:name])
+        return context[:seltype]
```
Magisus: Same here, need `parse_selinux_context`.

lib/puppet/type/k5login.rb
```diff
@@ -78,6 +128,54 @@ def mode=(value)
...
+    # Set the file selrange
+    def seltype=(value)
+      set_selinux_context(@resource[:name], value, :seltype)
+    end
+
+    # Set the file selrange
+    def selrange
+      if selinux_support?
+        context = get_selinux_current_context(@resource[:name])
+        return context[:selrange]
```
Magisus: Same here, need `parse_selinux_context`.

lib/puppet/type/k5login.rb
```diff
@@ -31,10 +32,59 @@
...
 
+  # To manage the selinux user of the file
+  newproperty(:seluser) do
+    desc "What the SELinux user component of the context of the file should be.
+      Any valid SELinux user component is accepted.  For example `user_u`.
+      If not specified it defaults to the value returned by matchpathcon for
+      the file, if any exists.  Only valid on systems with SELinux support
+      enabled."
+
+    defaultto { 'user_u' }
```
melissa: I'm not sure if I'm missing context from previous iterations, but I'm curious about the default here. I know on the file type, it looks like it's getting the default context of the file in question.

 https://github.com/puppetlabs/puppet/blob/51636c065e0625b32af2cd98f2060eb7428704f5/lib/puppet/type/file/selcontext.rb#L103

Is there any reason you're not doing something similar? Or would that be part of the extra complication that @branan suggested staying away from?

spec/unit/type/k5login_spec.rb
melissa: @jhelwig and I figured out what's going on here. It has to do with how Puppet dynamically creates and names classes, and took a bit of digging to figure out what the class name should to be.

This will get the tests green.
```diff
diff --git a/spec/unit/type/k5login_spec.rb b/spec/unit/type/k5login_spec.rb
index a3aeae5..291c6a8 100644
--- a/spec/unit/type/k5login_spec.rb
+++ b/spec/unit/type/k5login_spec.rb
@@ -78,10 +78,12 @@ describe Puppet::Type.type(:k5login), :unless => Puppet.features.microsoft_windo
         end

         context "with selinux" do
-          Puppet::Util::SELinux.stubs(:selinux_support?).returns true
-          Puppet::Util::SELinux.stubs(:get_selinux_current_context).returns "user_u:role_r:type_t:s0"
           subject { resource(:content => "daniel@EXAMPLE.COM\n",).retrieve }
           it "should return correct values based on SELinux state" do
+            Puppet::Type::K5login::ProviderK5login.any_instance.stubs(:selinux_support?).returns true
+            Puppet::Type::K5login::ProviderK5login.any_instance.stubs(:get_selinux_current_context).returns "user_u:role_r:type_t:s0"
             expect(subject[:seluser]).to eq("user_u")
             expect(subject[:selrole]).to eq("role_r")
             expect(subject[:seltype]).to eq("type_t")
```

lib/puppet/type/k5login.rb
jcpunk: That makes sense

spec/unit/type/k5login_spec.rb
jcpunk: I'll get that change in place and pushed up shortly.

spec/unit/type/k5login_spec.rb
```diff
@@ -69,6 +77,19 @@ def resource(attrs = {})
...
+          subject { resource(:content => "daniel@EXAMPLE.COM\n",).retrieve }
+          it "should return correct values based on SELinux state" do
+            Puppet::Type::K5login::ProviderK5login.any_instance.stubs(:selinux_support?).returns true
++            Puppet::Type::K5login::ProviderK5login.any_instance.stubs(:get_selinux_current_context).returns "user_u:role_r:type_t:s0"
```
jhelwig: Looks like a stray `+` from the diff made it in here at the beginning of the line.

jcpunk:
Fixed

lib/puppet/type/k5login.rb
```diff
@@ -31,10 +32,60 @@
...
 
+  # To manage the selinux user of the file
+  newproperty(:seluser) do
+    desc "What the SELinux user component of the context of the file should be.
+      Any valid SELinux user component is accepted.  For example `user_u`.
+      If not specified it defaults to the value returned by matchpathcon for
+      the file, if any exists.  Only valid on systems with SELinux support
+      enabled."
+
+    defaultto { self.retrieve_default_context(:seluser) }
```
melissa: Oops, it looks like this is defined in

https://github.com/puppetlabs/puppet/blob/51636c065e0625b32af2cd98f2060eb7428704f5/lib/puppet/type/file/selcontext.rb#L40

I was assuming this was defined in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/selinux.rb instead.

Let's revert this back to the defaults you originally had so we can get this merged in. I've opened [PUP-8233](https://tickets.puppetlabs.com/browse/PUP-8233) to move `retrieve_default_context` to `util/selinux.rb` so we can use these defaults in both places in the future.

jcpunk: Reverted and pushed new version

jcpunk:
Last batch of fixes seems to pass CI

PR #4951, opened 2016-05-10
(PUP-6229) Fix unclear error when dynamic expressions used in aliases

thallgren:
A Puppet Type Alias expression may not contain dynamic expressions. The
error issued when it did was not very informative since the actual
source of the dynamic expression was left out. This commit ensures that
it is included in the error message.


hlindberg:
Good to merge once travis/appveyor goes green.

PR #4952, opened 2016-05-10
(PUP-6151) Output Iterator value as Iterator-Value

thallgren:
Use dash between the type and the text 'value' and capitalize
'value' to make the output of an iterator type more readable.


hlindberg:
Good to merge when the light goes green...

PR #4953, opened 2016-05-10
(PUP-4856) Defer evaluation of 'default' node

thallgren:
This commit ensures that the evaluation of the default node is deferred
until all classes are known. This to ensure that classes included by the
ENC can override instantiations made by the 'default' node.

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -556,7 +556,15 @@ class Puppet::Parser::Compiler
     # of resources.
     resource = astnode.ensure_in_catalog(topscope)
 
-    resource.evaluate
+    # Evaluate the AST node unless it is the 'default' node. The 'default' node will
+    # then be evaluated together with other ENC added classes when #eval_definitions
+    # is called, giving those classes a chance to override definitions that the
+    # default node is using.
+    #
+    # If the 'default' node is evaluated right here, then any definition that it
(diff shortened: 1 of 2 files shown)


lib/puppet/parser/compiler.rb
```diff
@@ -556,7 +556,15 @@ def evaluate_ast_node
...
-    resource.evaluate
+    # Evaluate the AST node unless it is the 'default' node. The 'default' node will
+    # then be evaluated together with other ENC added classes when #eval_definitions
+    # is called, giving those classes a chance to override definitions that the
+    # default node is using.
+    #
+    # If the 'default' node is evaluated right here, then any definition that it
+    # instantiates, will be evaluated during the #eval_definitions phase and
+    # before evaluation of classes included by ENC.
+    resource.evaluate unless astnode.name == 'default'
```
hlindberg: This is scary:
* I think this means that the problem still exists if the node is matched by a specific node expression. (i.e. if the test in this pr has a node expr that matches 'the_node).
* The observed behavior means that a Resource Override is evaluation order dependent (a similar evaluation order problem can be constructed without involving the ENC). 
* What is the expected behavior when using both ENC and node expressions to classify?

(Think this discussion is best had on the ticket).

hlindberg:
Marked as BLOCKED due cause of the problem. See ticket.

ahpook:
Gonna close this PR since I closed the related ticket as "won't fix".

PR #4954, opened 2016-05-10
(PUP-6286) Remove cve-2013-1654_sslv2_downgrade_agent.rb test

kylog:
https://puppet.com/security/cve/cve-2013-1654/ describes a vulnerability
where an agent could connect to a rogue master and be downgraded to
sslv2. The puppet-agent packages that are used for acceptance tests
bundle an openssl that is built without sslv2 support, so this can't
happen, and this test has been skipped (at runtime) for some time now.

So just remove the test.

[skip ci]


shrug:
👍

johnduarte:
:+1:

PR #4959, opened 2016-05-11
(PUP-6253) Add SemVer and SemVerRange new documentation to `new`

hlindberg:


lib/puppet/functions/new.rb
```diff
@@ -396,6 +396,82 @@
 # Conversion to a `Struct` works exactly as conversion to a `Hash`, only that the constructed hash is
 # asserted against the given struct type.
 #
+# Creating a SemVer
+# ---
+# A SemVer object represents a single [Semantic Versions](http://semver.org/).
```
ahpook: should be "Semantic Version", not plural

lib/puppet/functions/new.rb
```diff
@@ -396,6 +396,82 @@
...
+#         )
+# function SemVer.new(SemVerHash $hash_args)
+#
+# ~~~
+#
+# Examples of SemVer and SemVerRange usage:
+#
+# ~~~ puppet
+#
+# # Note that as a type, SemVer can describe disjunct ranges which versions can be matched
```
ahpook: strike _Note that_, it's already in a callout box therefore is already "notable"

lib/puppet/functions/new.rb
```diff
@@ -396,6 +396,82 @@
...
+# notice(SemVer('3.4.5') =~ $t) # true
+#
+# ~~~
+#
+# Creating a SemVerRange
+# ---
+# A SemVerRange object represents a range of SemVer. It can be created from
+# a String, from two SemVer instances, where either end can be given as
+# a literal `default` to indicate infinity. The string format of a SemVerRange is specified by
+# the [SemVer Range Grammar](https://github.com/npm/node-semver#range-grammar).
```
ahpook: Suggest linking to https://github.com/npm/node-semver#ranges instead, as that anchor describes the usage of the Range, not the BNF.

ahpook:
👍

PR #4960, opened 2016-05-12
(maint) Fix bug for Semantic::Version in TypeCalculator::infer_set

thallgren:
The TypeCalculator::infer_set does not use the normal dynamic dispatch
algorithm. Instead, to speed things up, it uses a case/when. When adding
the SemVer type, an infer_set_Version method was added, but it was never
called because it was not added to that case/when. This commit fixes
this oversight and adds the tests needed to verify that infer_set works
for Semantic::Version.


hlindberg:
Fine to merge when tests go green.

PR #4962, opened 2016-05-12
(maint) Remove an ill-advised spec test

kylog:
This spec test should really be an acceptance test (see PUP-6003
for that task), because it runs against the actual package providers
found on the system executing the spec tests.

This becomes a problem in some environments, notably puppet CI's
spec runners: they are now puppetized, and so the test finds puppet-agent's
gem provider and attempts to run that *with* rvm adjusted paths for
the ruby version being spec tested. This makes no sense and ends up
mixing two ruby version's paths.

Since we plan to replace the test with an actual acceptance test
and since it fails in puppet CI for no reason related to the code being
tested, remove it.


shrug:
👍

PR #4965, opened 2016-05-16
(PUP-6301) Make AIX service test setup idempotent

johnduarte:
This commit alters the setup step for ensuring that the test service is
stopped. It removes the validation that the service changed state. This allows
the step to pass if the service was already stopped prior to running the
setup step that stops it.

Prior to this commit, if the system did not automatically start the service
upon creation the test would fail during the setup step intended to ensure
that it was stopped.

### DIFF
--- a/acceptance/tests/resource/service/AIX_service_provider.rb
+++ b/acceptance/tests/resource/service/AIX_service_provider.rb
@@ -46,6 +46,12 @@ def lssrc_assert_status(host, service, expected_status)
   end
 end
 
+teardown do
+  agents.each do |agent|
+    on(agent, "rmssys -s sloth_daemon", :allowable_exit_codes => [0,1])
+  end
+end
+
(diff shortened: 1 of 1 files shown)


whopper:
Verified that this test fails without the fix, and passes with it. Also manually verified that `mkssys -s testservice -p /tmp/derp -u 0 -S -n 15 -f 9` creates the `testservice` subsystem but does not start it:

```
[0] [AIX] root@pe-aix-71-agent:~ # lssrc -s testservice
Subsystem         Group            PID          Status
 testservice                                           inoperative
```

This failure is actually my fault! :D In commit https://github.com/puppetlabs/puppet/commit/d8a8ef269ff5ff5bac7933c100b7d0da5f1b073e I added the bit about `mksys` starting the service, which is wrong at least in AIX 7.1.

PR #4966, opened 2016-05-17
(PUP-6233) Serivce SMF restart should use -s

shawnferry:
The restartcmd in smf.rb needs to add the -s flag to make
sure that service restarts are synchronous. Without this
patch smf will return prior to the completion of the restart.

(PUP-6238) Serivce SMF should set has_feature :refreshable

SMF supports the concept of restart instead of calling start/stop
on each service. This patch informs the higher level service
provider of that capability.


shawnferry:
This is identical to the end state of [4922](https://github.com/puppetlabs/puppet/pull/4922) which was closed in error.

whopper:
Ran this through our service acceptance test and it passed in both Solaris 10 and 11. 👍

PR #4968, opened 2016-05-17
(PUP-6318) Add --no-batch-files option to install.rb

McdonaldSeanp:
puppet-agent carries the batch files for all executables with the new move to
vanagon for windows projects, we will need an option to omit batch file
creation during installation of puppet so there are not more than one set of
batch files. So this commit adds a --no-batch-files option for cmake that omits
those files.


ScottGarman:
:+1:  assuming this has been tested.

PR #4972, opened 2016-05-19
(PUP-6220) Add TypeSet type to the Puppet Type system

thallgren:
This PR adds the TypeSet type to the Puppet Type system parser,
evaluator, and string formatter.

A TypeSet can define new types using a Hash construct where the key
is the name of the object and the value is a Type. The Type is typically
an Object[] declaration but can also be something else, in which case
the key => value pair is treated as a type alias.

A TypeSet may also contain references to other TypeSets. Just references
are resolved during the normal #resolve phase for the type (called from
the TypeParser when the type is loaded). The references are resolved
using a loader.

Types in the TypeSet itself are resolved using a special loader that
is aware of the TypeSet so that all types can be cross-referenced. This
also means that types within the TypeSet can access types from the
referenced TypeSets.


lib/puppet/pops/loader/typed_name.rb
```diff
@@ -24,7 +26,7 @@ def initialize(type, name)
     @name_parts = parts
 
     # Use a frozen compound key for the hash and comparison
-    @compound_name = "#{type}/#{name}".freeze
+    @compound_name = "#{name_authority}/#{type}/#{name}".freeze
```
hlindberg: Does this have performance implications - all (in this impl) compound names will then start with the same string. Maybe we should consider reversing it?
Alternatively internalize it (make it a symbol) since comparison is then on identity.

lib/puppet/pops/types/annotatable.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def i12n_hash
+    result = {}
+    result[KEY_ANNOTATIONS] = @annotations unless @annotations.nil?
+    result
+  end
+end
+end
+end
```
hlindberg: missing end of line

lib/puppet/pops/types/annotatable.rb
```diff
@@ -0,0 +1,30 @@
+module Puppet::Pops
+module Types
+
+KEY_ANNOTATIONS = 'annotations'.freeze
+
+module Annotatable
```
hlindberg: # @api private
plus some comment 1-2 lines explaining what this is would be nice. "Behaviour common to all Pcore annotateable classes" or somesuch

lib/puppet/pops/types/p_meta_type.rb
```diff
@@ -2,8 +2,18 @@
...
+
+KEY_NAME = 'name'.freeze
+KEY_TYPE = 'type'.freeze
+KEY_VALUE = 'value'.freeze
+
 class PMetaType < PAnyType
-  KEY_NAME = 'name'.freeze
+  include Annotatable
+
+  def accept(visitor, guard)
```
hlindberg: doc comment would be nice for this method

lib/puppet/pops/types/type_factory.rb
```diff
@@ -181,13 +181,7 @@ def self.any
...
   #
   def self.regexp(pattern = nil)
-    if pattern
-      t = PRegexpType.new(pattern.is_a?(Regexp) ? pattern.inspect[1..-2] : pattern)
-      t.regexp unless pattern.nil? # compile pattern to catch errors
-      t
-    else
-      PRegexpType::DEFAULT
-    end
+    pattern ?  PRegexpType.new(pattern) : PRegexpType::DEFAULT
```
hlindberg: Instead of doing this, can the URI pattern set the option inside of the delimiters  - that is possible instead of using a trailing 'x'.
That way there is no chance of trailing options leak into puppet land (at least not via the type).

hlindberg:
The commit c989e4a0b498603b1ca4ce4363ca027e8b6c1c21 has a typo in its commit message

lib/puppet/pops/pcore.rb
```diff
@@ -2,29 +2,50 @@
...
+  def self.register_implementations(impls, name_authority = RUNTIME_NAME_AUTHORITY)
+    Loaders.loaders.register_implementations(impls, name_authority = RUNTIME_NAME_AUTHORITY)
   end
 
-  def self.register_aliases(aliases)
+  def self.register_aliases(aliases, name_authority = RUNTIME_NAME_AUTHORITY)
     loader = Loaders.loaders.private_environment_loader
-    aliases.each { |name, type_string| add_type(Types::PTypeAliasType.new(name, Types::TypeFactory.type_reference(type_string), nil), loader) }
+    aliases.each { |name, type_string| add_type(
+      Types::PTypeAliasType.new(name, Types::TypeFactory.type_reference(type_string), nil), loader, name_authority) }
```
hlindberg: Boo from the Rubo Cop:
```
== lib/puppet/pops/pcore.rb ==
C: 48:117: Style/BlockEndNewline: Expression at 48, 117 should be on its own line.
```
Since the `{` is not on the same line, the `}` should be on its own line.

lib/puppet/pops/pcore.rb
```diff
@@ -2,14 +2,32 @@
...
+  PCORE_URI = 'http://puppet.com/2016.1/pcore'
+  PCORE_VERSION = Semantic::Version.new(1,0,0)
+  PARSABLE_PCORE_VERSIONS = Semantic::VersionRange.parse('1.x')
+
   RUNTIME_NAME_AUTHORITY = 'http://puppet.com/2016.1/runtime'
 
   def self.init(loader, ir)
-    add_alias('Puppet::Pcore::QualifiedReference', TYPE_QUALIFIED_REFERENCE, loader)
+    add_alias('Pcore::URI_RX', TYPE_URI_RX, loader)
+    add_alias('Pcore::URI', TYPE_URI, loader)
```
hlindberg: Should the URI type really be just a pattern as opposed to a real runtime object?

lib/puppet/pops/types/type_factory.rb
thallgren: The string form of the `PRegexpType` is also changed in this commit to do exactly what you suggest. The above code does not "leak options into puppet land". It makes sure that such options doesn't corrupt the regexp.

lib/puppet/pops/pcore.rb
thallgren: It should, but adding the `URI` type deserves a PR of its own. Once added, the initialization hashes in `TypeSet` and `TypeSetReference` must accept both a `String` and an `URI`  (just like they do now with `SemVer` and `SemVerRange`).

lib/puppet/pops/types/type_factory.rb
hlindberg: Can you elaborate - I don't quite understand. Do you mean that it accepts `//opt` as input in string form, but when presenting that adds option wrappers?

thallgren: The above change ensures that a `Regexp` instance can pass through without intervention. This is OK because the `PRegexpType` constructor will accept either a `String` or a `Regexp`. The actual string representation is not dictated by the instance itself, but by which method that is chosen to represent it (#inspect, #source, #to_s). The `Regexp` instance itself keeps the `source` and `options` separate.

Prior to this change, the `Regexp` was converted to a `String` using #inspect which always produces the `/pat/x` form. The subsequent substring expression then converted the string to "pat/". Although still a valid expression, it no longer represents the original expression. I.e. the regexp is now corrupted.

By passing the `Regexp` directly to the `PRegexpType` constructor, the responsibility of producing the string representation is moved. The `PRegexType` infers the string by either using the `#source` method (which doesn't include the options) or by using the `#to_s` method which will include the options using the `/(?x:pat)/` syntax. The reason for not always using the `#to_s` method is that it will _always_ include the options, even when no options have been set (as `-xmi`).

lib/puppet/pops/pcore.rb
hlindberg: Shouldn't it take `Initializer[URI]` which would be a variant of the signatures to URI new? Agree it is a separate ticket, and so is Initializer functionality.

spec/unit/pops/types/p_type_set_type_spec.rb
```diff
@@ -0,0 +1,424 @@
...
+              ts = <<-OBJECT
+                pcore_version => '1.0.0',
+                version => '1.0.0',
+                references => {
+                  A => { name => 'Vehicle::Cars', version_range => '1.x' },
+                  V => { name => 'Vehicle::Cars', version_range => '1.x' },
+                }
+              OBJECT
+              expect { parse_type_set('MySet', ts) }.to raise_error(ArgumentError,
+                /references TypeSet 'http:\/\/puppet\.com\/2016\.1\/runtime\/Vehicle::Cars'/)
```
hlindberg: this error message seems a bit off - how would the user know it means the same set more than once?

spec/unit/pops/types/p_type_set_type_spec.rb
```diff
@@ -0,0 +1,424 @@
...
+              end
+
+              it 'has a name that is not a QRef' do
+                ts = <<-OBJECT
+                  pcore_version => '1.0.0',
+                  version => '1.0.0',
+                  references => { Ref => { name => 'cars', version_range => '1.x' } }
+                OBJECT
+                expect { parse_type_set('MySet', ts) }.to raise_error(TypeAssertionError,
+                  /entry 'references' entry 'Ref' entry 'name' expected a match for Pattern\[\/\\A\(\(::\)\{0,1\}\[A-Z\]\[\\w\]\*\)\+\\z\/\], got 'cars'/)
```
hlindberg: Do we allow leading `::` in names ? That would allow breaking out of the type set namespace and reference anything anywhere without having a reference to it.
Am I reading this wrong?

spec/unit/pops/types/p_type_set_type_spec.rb
```diff
@@ -0,0 +1,424 @@
...
+                  name => 'Transports',
+                  version_range => '1.x'
+                }
+              }
+          OBJECT
+        end
+      end
+    end
+  end
+end
```
hlindberg: And from that day they lived hap...

hlindberg: Missing end of line.

hlindberg:
I have lots of questions, but code looks good. It is one big step along the way.

hlindberg:
Questions resolved. There are only the small changes mentioned that needs to be addressed then this can be merged.

thallgren:
I rebased this PR on master and added a commit to address the small changes.

hlindberg:
There were two travis runs that failed - seemed to be timing / time related. Restarted jobs.

hlindberg:
> When green is all there is to be
> It could make you wonder why, but why wonder?
> Why Wonder, I am green and it'll do fine, it's beautiful!
> And I think it's what I want to be.
> _kermit the frog_

PR #4973, opened 2016-05-19
(PUP-6324) Allow DNF to update packages

petems:
* `dnf` inherits from `yum`
* `yum` allows the updating of packages with `install`
* `dnf` does not:
```
/bin/dnf -y install cockpit
Last metadata expiration check performed 0:18:07 ago on Thu May 19 13:40:06 2016.
Package cockpit-0.67-2.fc22.x86_64 is already installed, skipping.
Dependencies resolved.
Nothing to do.
Complete!
```
* ~~So with the `dnf` provider, we need to use the `upgrade` command if the user is trying to install anything newer than the current install~~

EDIT: After some reading I've moved the login into the yum provider, same behaviour.

Also DRY'd up the specs as I was getting bored of copy+pasting the stuff from the Yum provider specs 😄


lib/puppet/provider/package/dnf.rb
```diff
@@ -30,6 +30,53 @@
 
   defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => ['22', '23']
 
+  def install
```
petems: So yum also allows the `upgrade` command for updating packages, so it might be easier to add this logic back up to the yum provider. Smaller footprint of changes, DRY-s up the code. Just need to see if `yum upgrade` is backwards compatible so it doesnt cause a regression for older releases...

lib/puppet/provider/package/yum.rb
```diff
@@ -159,6 +172,11 @@ def install
         self.debug "Downgrading package #{@resource[:name]} from version #{is[:ensure]} to #{should}"
         operation = :downgrade
       end
+
+      if is && rpm_compareEVR(rpm_parse_evr(should), rpm_parse_evr(is[:ensure])) > 0
```
petems: Meh, this could probably be drier or refactored into a method...

lib/puppet/provider/package/yum.rb
```diff
@@ -159,6 +172,11 @@ def install
...
+        self.debug "Updrading package #{@resource[:name]} from version #{is[:ensure]} to #{should}"
```
adrienthebo: Typo: "Updrading" -> "Upgrading"

adrienthebo: And a note on this debug message - does this log information that isn't already being logged by the resource harness/transaction?

petems: Not sure, I just followed the pattern from the downgrade block, but that's fairly old (9ccd29f3aacdab03f2ea9a693b5bca928439683b, 2010 commit!) so can remove if needed?

HAIL9000:
@adrienthebo Are you 👍 on this now? If so I'll go ahead and merge it!

lib/puppet/provider/package/yum.rb
adrienthebo: This code is slightly confusing - it's effectively an `elsif` condition paired with line 171. I would rather have these statements joined rather than two effectively connected but separate conditionals.

petems: Like this? 

```ruby
if is
        if rpm_compareEVR(rpm_parse_evr(should), rpm_parse_evr(is[:ensure])) < 0
          self.debug "Downgrading package #{@resource[:name]} from version #{is[:ensure]} to #{should}"
          operation = :downgrade
        end

        if rpm_compareEVR(rpm_parse_evr(should), rpm_parse_evr(is[:ensure])) > 0
          self.debug "Upgrading package #{@resource[:name]} from version #{is[:ensure]} to #{should}"
          operation = update_command
        end
      end
```

adrienthebo: Not quite, though the outer `if` statement does clean things up. I was thinking of this though:

```
if is
        if rpm_compareEVR(rpm_parse_evr(should), rpm_parse_evr(is[:ensure])) < 0
          self.debug "Downgrading package #{@resource[:name]} from version #{is[:ensure]} to #{should}"
          operation = :downgrade
        elsif rpm_compareEVR(rpm_parse_evr(should), rpm_parse_evr(is[:ensure])) > 0
          self.debug "Upgrading package #{@resource[:name]} from version #{is[:ensure]} to #{should}"
          operation = update_command
        end
      end
```

Those two different conditionals can't both be true, but a casual reading of the code with the independent conditionals doesn't indicate this.

petems: Cool, fixed to that. I also changed `is` because that variable name seems too vague

adrienthebo:
ship it

HAIL9000:
Thanks @petems for the contribution and @adrienthebo for the extra set of eyes!!

PR #4974, opened 2016-05-19
(packaging) Bad conflicts for vim and emacs can uninstall puppet

underscorgan:
Since the vim and emacs extension packages conflicted with puppet <=
current version, if that package was upgraded it would remove puppet and
anything dependant on puppet. This is not correct behavior. Add puppet
>= current version as a dependency of those packages instead.

### DIFF
--- a/ext/debian/control
+++ b/ext/debian/control
@@ -123,17 +123,15 @@ Description: Centralised configuration management - master setup to run under mo
 
 Package: vim-puppet
 Architecture: all
-Depends: ${misc:Depends}
+Depends: ${misc:Depends}, puppet (= ${source:Version})
 Recommends: vim-addon-manager
-Conflicts: puppet (<< ${source:Version})
 Description: syntax highlighting for puppet manifests in vim
  The vim-puppet package provides filetype detection and syntax highlighting for
  puppet manifests (files ending with ".pp").
 
 Package: puppet-el
 Architecture: all
-Depends: ${misc:Depends}, emacsen-common
-Conflicts: puppet (<< ${source:Version})
+Depends: ${misc:Depends}, emacsen-common, puppet (= ${source:Version})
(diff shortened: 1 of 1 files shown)


ext/debian/control
```diff
@@ -123,17 +123,15 @@ Description: Centralised configuration management - master setup to run under mo
 
 Package: vim-puppet
 Architecture: all
-Depends: ${misc:Depends}
+Depends: ${misc:Depends}, puppet (>= ${source:Version})
```
stahnma: Since vim is built as a subpackage of puppet in this paradigm, why not make the requirement = rather than >= ?

underscorgan: ahh, good call. Will update.

underscorgan:
Updated, build running at http://jenkins-release.delivery.puppetlabs.net/job/puppet-packaging-2016-05-19-12-31-04-7c326eeece4f773ffc0e1ed2a5ee04acadb4c229

underscorgan:
```
root@yoj1he976ryhbrd:~# dpkg -i vim-puppet_3.8.7.3-1puppetlabs1_all.deb
(Reading database ... 94764 files and directories currently installed.)
Preparing to unpack vim-puppet_3.8.7.3-1puppetlabs1_all.deb ...
Unpacking vim-puppet (3.8.7.3-1puppetlabs1) over (3.8.7.3-1puppetlabs1) ...
dpkg: dependency problems prevent configuration of vim-puppet:
 vim-puppet depends on puppet (= 3.8.7.3-1puppetlabs1); however:
  Version of puppet on system is 3.8.5-1puppetlabs1.
```
This should prevent vim-puppet and puppet-el from being installed with a version other than the version of puppet that's installed.

mckern:
This looks not unreasonable 👍

McdonaldSeanp:
👍

PR #4975, opened 2016-05-19
(PUP-6327) better invalid path error

binford2k:
This clarifies that the path is invalid. Prior to this, it wasn't super
clear why the PMT install failed.


whopper:
@binford2k there are a few Windows-specific tests failing in `spec/unit/module_tool/tar/mini_spec.rb` because they haven't been updated with your new error message.

joshcooper:
ping @binford2k

binford2k:
Wow, those checks take forever. All done now, ping @joshcooper

branan:
:+1: Thanks for taking the time to make Puppet better!

PR #4976, opened 2016-05-20
(PUP-6321) Handle autorequire with undef values

thallgren:
Prior to this commit, an autorequire that produced `undef` values would
produce an error:
```
No title provided and :xxx is not a valid resource reference
```
This commit ensures that `undef` entries provided by an autorequire are
skipped over.


hlindberg:
Can this be unit tested (avoid a future regression), or does it require acceptance testing?

thallgren:
I added some unit tests.

PR #4977, opened 2016-05-20
(PUP-4742) Switch to use semantic_puppet gem

thallgren:
This PR removes the vendored (and embedded in the Puppet code-base) 'semantic' gem in favor of adding a dependency to the 'semantic_puppet' gem.

The PR supersedes [PR-4708](/puppetlabs/puppet/pull/4708).


hlindberg:
Travis failed Ruby 2.3.1 run - respinning but may be a legit error on 2.3.1 - the tests failed (IIRC) with nil vs. not nil expectancy.

kylog:
Oh interesting. We havent't seen much in the way of 2.3 breakage. I'm curious on the root cause.

hlindberg:
Three spec failures in module tool specs. Copied the output to the ticket as well for posterity and visibility.

thallgren:
The problem is very similar to the one I fixed in [PR-12](/puppetlabs/semantic_puppet/pull/12), albeit this time the culprit is the `SemanticPuppet::Dependency::ModuleRelease`. I created [PUP-6341](https://tickets.puppetlabs.com/browse/PUP-6341) as an issue that blocks this one.

PR #4979, opened 2016-05-20
(PUP-6318) Update install.rb to actually honor InstallOptions.batch_files

McdonaldSeanp:
Install.rb was just updated to add an install option to omit batch file
generation during installation. But this change was incomplete. This commit
should suffice to actually omit all batch files during install


jcoconnor:
👍

geoffnichols:
@McdonaldSeanp I think we may want to update the commit message to refer to the original PUP ticket.

PR #4981, opened 2016-05-21
(PUP-6336) Fix problem with diff Jruby/MRI in String#==

hlindberg:
Puppet has a monkey patch for `Symbol#<=>(String)` that enables
comparison. This is deprecated and slated for removal in puppet 5.0.0.

On JRuby, `String#==(other)` is implemented as `other<=>(self) == 0` which
causes the monkey patch in puppet to trigger.

In 1547756ad7caa5cb1f0add8161e51ac07f61bd67 we reversed a test from `x
== :undef` to `:undef == x`. And this caused the problem reported in
PUP-6336 as this logic now (via the MRI incompatible implementation in
JRuby) caused the monkey patch in puppet to kick in.

This commit changes the test `:undef == x` to `:undef.equal?(x)` which
avoids both the JRuby incompatibility, and the problem the change
solved in the first place (ending up in the same insane monkey patch).

### DIFF
--- a/lib/puppet/pops/evaluator/runtime3_support.rb
+++ b/lib/puppet/pops/evaluator/runtime3_support.rb
@@ -82,7 +82,7 @@ module Runtime3Support
       # Must convert :undef back to nil - this can happen when an undefined variable is used in a
       # parameter's default value expression - there nil must be :undef to work with the rest of 3x.
       # Now that the value comes back to 4x it is changed to nil.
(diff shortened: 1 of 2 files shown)


haus:
:+1:

hlindberg:
With @haus +1 and @kylog 's in hipchat - I am merging this so we get an acceptance run as early as possible.

PR #4982, opened 2016-05-21
(PUP-6125) verbose/debug no longer changes logdest

pyther:
* `--test` implies changing the default logdest to console (which would still be overridden by `--logdest`).
* `--verbose` and `--debug` should not imply writing to console. This is a change in existing behavior that allows people using verbose/debug in daemon runs to continue doing so. If they wish to still write to console, they'll have to add `--logdest` console.


pyther:
Been trying to add a unit test to make sure the console is the default when --test is passed, but I'm not having any luck. Any help would be appreciated.
```
      it "should set console as the log destination if --test was passed" do
        @puppetd.options[:setdest] = false
        @puppetd.options[:test] = true

        Puppet::Util::Log.expects(:newdestination).with(:console).once

        @puppetd.setup_test
        @puppetd.setup_logs
      end
```

pyther:
@MikaelSmith I stole some of your wording from the puppet ticket for the commit messages, I hope you don't mind.

lib/puppet/application.rb
```diff
@@ -357,10 +357,6 @@ def setup
   end
 
   def setup_logs
-    if options[:debug] || options[:verbose]
-      Puppet::Util::Log.newdestination(:console)
-    end
-
```
hlindberg: This modifies the API for puppet application, for all modules that derive from it and have expectations on how it behaves.
Not all such applications are in the puppet code base.

hlindberg:
I think this breaks API. If the change can be done so that abstract base classes (e.g. "Application") is unchanged it is fine to do whatever in the agent and master applications. The commands that are interactive should still use console IMO (e.g. puppet master compile). 

Could this be done in a way that preserves backwards compatibility. I.e. those that do not want to log to console when using --verbose and --debug can turn that off? (Then it is not a surprise of anyone, or a possible break to someone's workflow).

whopper:
Closed in favor of https://github.com/puppetlabs/puppet/pull/4987

PR #4983, opened 2016-05-23
prefer already loaded core functions to searching for 4x api functions

keymone:
i understand this is not perfect, as there are reasons to override some 3x functions with 4x implementations, but traversing every gem and module in `LOADPATH` in search for `include.rb` is just not acceptable.

is there a class/object somewhere that can tell me a list of defined core functions to prefer that instead of all defined 3x functions?

i'm happy to implement any suggestions. this patch is about 20% performance boost and i wouldn't like to give it up.

### DIFF
--- a/lib/puppet/pops/evaluator/runtime3_support.rb
+++ b/lib/puppet/pops/evaluator/runtime3_support.rb
@@ -134,7 +134,7 @@ module Puppet::Pops::Evaluator::Runtime3Support
   end
 
   def set_scope_nesting_level(scope, level)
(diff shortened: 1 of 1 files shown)


kylog:
Ping @hlindberg

hlindberg:
4.x functions must always win over 3.x functions or things will go very bad. Changing the load order is not an option.

hlindberg:
The proposed fix would break the specification and we just cannot do it this way. In 4.x the loaders cache misses. I am therefore going to close this PR.

PR #4985, opened 2016-05-23
Cache environment config same way as environment

keymone:
right now environment config is never cached. this means calls to things like `Puppet.future_parser?` are always reloading environment configuration which slows down everything.

lets cache environment config same way as environment is cached.

i got ~30-40% speedup on large catalog, reduced number of stat calls from 375k to 137k.

### DIFF
--- a/lib/puppet/environments.rb
+++ b/lib/puppet/environments.rb
@@ -313,6 +313,7 @@ module Puppet::Environments
       @loader = loader
       @cache_expiration_service = Puppet::Environments::Cached.cache_expiration_service
       @cache = {}
+      @cache_conf = {}
 
       # Holds expiration times in sorted order - next to expire is first
       @expirations = SortedSet.new
@@ -365,6 +366,7 @@ module Puppet::Environments
     # (The intention is that this could be used from a MANUAL cache eviction command (TBD)
     def clear(name)
       @cache.delete(name)
+      @cache_conf.delete(name)
     end
 
     # Clears all cached environments.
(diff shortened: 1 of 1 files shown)


branan:
Hi @keymone,

This looks fine to me at first glance. The tests do need to be updated before we can merge. I also want to talk to someone from the compiler team who can verify my understanding of environment expiration is correct before I push the button

branan:
Your commit message also does not follow our requirements. Please read [contributing.md](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md). Specifically, it needs the ticket number at the start and a more detailed body explaining the change.

hlindberg:
if the config is not dynamically loaded (i.e. cached) and timeout is set to a very long time then you have to take down the server to make it react to a change.
This may be a moot point since the environment is still cached.

hlindberg:
Environment config is also used by the client/agent, it may have different expectations (that I am not aware of).

branan:
That was my understanding. I don't know if there's a real use case for having the config be replaceable if the rest of the environment is on a long timeout. My gut says "no", though.

branan:
@keymone Hi, we haven't heard back from you. Do you think you'll be able to take the time to fix up the tests and commit message? We'll leave this open a little while longer but will probably close it in a couple weeks if we haven't heard from you.

joshcooper:
This seems still applicable, but needs to be fixed up (single spec failures) and targeted to a later branch.

MosesMendoza:
While this PR seems like a valuable addition, I'm afraid the team doesn't have the capacity to review/incorporate it and add the required tests for validation. I'm going to close for now, but we can reopen if/when we can pick it up again.

PR #4986, opened 2016-05-23
(PUP-6294) Indicate noop run mode in puppet report

kbrezina:
PE needs an exact info about noop puppet run. The proposed solution seems to be the easiest way to implement it. However, it can have consequences I'm not aware of. Please, provide feedback.

This problem is reported as https://tickets.puppetlabs.com/browse/PUP-6294.

Intention of this PR is to open discussion about this problem and the proper implementation - marking as Blocked.

### DIFF
--- a/api/docs/http_report.md
+++ b/api/docs/http_report.md
@@ -1,7 +1,7 @@
 Report
 ======
 This document describes the Puppet master's report endpoint and the schema for
-Report Format 4 in technical term. Also see the
+Report Format 6 in technical term. Also see the
 [documentation](https://docs.puppetlabs.com/puppet/latest/reference/format_report.html).
 
 The `report` endpoint allows clients to send reports to the master via `http`
@@ -58,6 +58,7 @@ example is formatted for readability)
      "puppet_version"=>"3.3.0",
      "kind"=>"apply",
      "status"=>"unchanged",
+     "noop"=>false,
      "environment"=>"test_environment",
      "logs"=>
       [{"level"=>"warning",
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -108,6 +108,11 @@ class Puppet::Transaction::Report
   #
   attr_reader :report_format
 
+  # Whether the puppet run was started in noop mode
+  # @return [Boolean]
+  #
+  attr_reader :noop
+
   def self.from_data_hash(data)
     obj = self.allocate
     obj.initialize_from_hash(data)
@@ -183,7 +188,7 @@ class Puppet::Transaction::Report
     @host = Puppet[:node_name_value]
     @time = Time.now
     @kind = kind
-    @report_format = 5
+    @report_format = 6
     @puppet_version = Puppet.version
     @configuration_version = configuration_version
     @transaction_uuid = transaction_uuid
@@ -192,6 +197,7 @@ class Puppet::Transaction::Report
     @cached_catalog_status = nil
     @environment = environment
     @status = 'failed' # assume failed until the report is finalized
+    @noop = Puppet[:noop]
   end
 
   # @api private
@@ -202,6 +208,7 @@ class Puppet::Transaction::Report
     @transaction_uuid = data['transaction_uuid']
     @environment = data['environment']
     @status = data['status']
+    @noop = data['noop']
     @host = data['host']
     @time = data['time']
 
@@ -255,6 +262,7 @@ class Puppet::Transaction::Report
       'puppet_version' => @puppet_version,
       'kind' => @kind,
       'status' => @status,
+      'noop' => @noop,
       'environment' => @environment,
 
       'logs' => @logs,
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -15,6 +15,7 @@
         "puppet_version",
         "kind",
         "status",
+        "noop",
         "environment"
     ],
 
@@ -115,7 +116,7 @@
         "report_format": {
             "description": "The report format version documented by this schema",
             "type":        "integer",
-            "enum":        ["5", 5]
+            "enum":        ["6", 6]
         },
 
         "puppet_version": {
@@ -138,6 +139,11 @@
                 "changed",
                 "unchanged"
             ]
+        },
+
+        "noop": {
+            "description": "Whether the puppet run was started in noop mode",
+            "type":        "boolean"
         },
 
         "environment": {
--- a/spec/unit/transaction/report_spec.rb
+++ b/spec/unit/transaction/report_spec.rb
@@ -70,6 +70,24 @@ describe Puppet::Transaction::Report do
     expect(report.cached_catalog_status).to eq("explicitly_requested")
   end
 
+  it "should set noop to true if Puppet[:noop] is true" do
+    Puppet[:noop] = true
+    report = Puppet::Transaction::Report.new("apply")
+    expect(report.noop).to be_truthy
+  end
+
+  it "should set noop to false if Puppet[:noop] is false" do
+    Puppet[:noop] = false
+    report = Puppet::Transaction::Report.new("apply")
+    expect(report.noop).to be_falsey
+  end
+
+  it "should set noop to false if Puppet[:noop] is unset" do
+    Puppet[:noop] = nil
+    report = Puppet::Transaction::Report.new("apply")
+    expect(report.noop).to be_falsey
+  end
+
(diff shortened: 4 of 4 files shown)


kylog:
Thanks @kbrezina!

This looks like the general direction I was thinking of. Two comments:

1) re how to represent the new bit of data in the report: rather than a "noop_run" boolean, I was thinking of a new field providing an array of enums (or strings) for select options that were in effect for the catalog application. This would allow the new field to be extended with other options in the future (e.g. "use_cached_catalog" etc).

2) I think we'll want to bump the report format number from 4 to 5 as part of this change.

Thoughts?

kbrezina:
@kylog Thanks for your feedback!

1) How would you name the new field? What would it contain in the first version? Or, how can I recognize items that the array should contain?
2) At the moment, the current report format is 5 (see https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction/report.rb#L186). So, it needs to be bumped to 6.

api/docs/http_report.md
```diff
@@ -58,6 +58,7 @@ example is formatted for readability)
      "puppet_version"=>"3.3.0",
      "kind"=>"apply",
      "status"=>"unchanged",
+     "noop_run"=>false,
```
kbarber: Does it need to be 'noop_run'? Why not just make it noop?

kbrezina: +1

kylog:
> 1) How would you name the new field? What would it contain in the first version? Or, how can I recognize items that the array should contain?

I was thinking something like this as an addition to the schema:

```
        "settings": {
            "description": "Zero or more settings",
            "type":        "array",
            "items": {
                "enum": [
                    "noop"
                ]
            }
        },

```

That would make it a bit future-proof in that we could add more settings to it in the future (albeit only boolean ones, hmmm, need to think more on that).

> 2) At the moment, the current report format is 5 (see https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction/report.rb#L186). So, it needs to be bumped to 6.

Doh! You're right, and we never documented that. I filed https://tickets.puppetlabs.com/browse/DOCUMENT-556 to fix that.

So yes, this would be format 6.

kbrezina:
@kylog Wouldn't it be better to use "hash" instead of "array" for the "settings"? Then, it would contain `{"noop" => true}`. And, it could contain even non boolean values.

kylog:
@kbrezina yes +1 on hash rather than array.

kbrezina:
@kylog @kbarber Please, take a look at the updated version which is using `settings` attribute.

kylog:
👍 here, but I'd like to hear from @kbarber too, and I'd like some resolution on the sidebar email thread about this before merging.

lib/puppet/transaction/report.rb
```diff
@@ -175,6 +181,11 @@ def finalize_report
   end
 
   # @api private
+  def get_settings
+    {"noop" => (Puppet[:noop] || false)}
```
kbarber: Shouldn't puppet defaults handle the defaulting already to false? Seems like we're going to be rewriting the default settings system inherit in puppet if we go down this road of providing defaults here also.

lib/puppet/transaction/report.rb
```diff
@@ -183,7 +194,7 @@ def initialize(kind, configuration_version=nil, environment=nil, transaction_uui
     @host = Puppet[:node_name_value]
     @time = Time.now
     @kind = kind
-    @report_format = 5
+    @report_format = 6
```
kbarber: Are we really bumping the version for every new field? Is that our versioning policy here @kylog? Seems like we've broken this a few times in the past if thats true.

lib/puppet/transaction/report.rb
kylog: Ah yeah, good point. Seems like this could just be:
```
{"noop" => Puppet[:noop]}
```

lib/puppet/transaction/report.rb
kbarber: Where is the policy recorded for this thing incidentally? We have our own document outlining our versioning policy on these items, but I haven't seen such a thing for Puppet related changes.

kylog: We talked about this some with @nfagerlund last time around, and the policy (I think) we converged on was that we would bump it for any new (or removed) field, but not more often than once per release. (E.g. if we add another change before 4.6 we wouldn't need to bump the report format number *again*.)

I think the reasoning here was that the report format gives a consumer one place to look at to know what fields to expect (and bumping it has very little (any?) cost).

Nick, does that match your memory?

kylog: > Where is the policy recorded for this thing incidentally? 

It isn't. This would be a great time to agree on where to record this policy. And to record it. :)

lib/puppet/transaction/report.rb
```diff
@@ -108,6 +108,12 @@ class Puppet::Transaction::Report
   #
   attr_reader :report_format
 
+  # A hash of selected puppet run settings:
+  #   :noop [Boolean] - whether the puppet run was started in noop mode
+  # @return [Hash<{String => Object}>]
+  #
+  attr_reader :settings
```
kbarber: The concept of having a hash in Puppet is an interesting one, I like it in a way. Having said that, be aware that hash traversal of this nature isn't easy in PDB yet, so we may be left just mapping this to the top level for the moment in our terminus: https://tickets.puppetlabs.com/browse/PDB-2633

lib/puppet/transaction/report.rb
nfagerlund: Yep, that matches my memory!

nfagerlund: @kbarber and yeah, we've broken it a lot in the past. It resulted in chaos, so I made a small tantrum about how let's not do that anymore. :)

kbrezina:
@kylog @kbarber How do we progress with this PR? Is the report version the very last thing that needs to be sorted out? Who will make the final decision?

kbarber:
@kbrezina we debated this issue somewhat in that email thread.

The last points I made on that thread:

* We should have only a top-level boolean noop field, not settings as discussed. This is because it aligns closer to the PuppetDB api, and the settings stuff for the moment is YAGNI. We also can't search hashes like this easily enough in PDB yet.
* For the new noop status, we add a new field also. (but not a part of this patch)

At the end of the day final decisions on PR merges are always up to the team that owns it (so its Kylo). I think Kylo and I have a consensus anyway.

kbrezina:
@kylog @kbarber Reimplemented as discussed above.
I'll create a follow up ticket for the new noop_status field.

kbrezina:
I removed Blocked label since the PR seems to be mergeable now.

kbarber:
+1

kbarber:
@kylog ?

kbrezina:
@kylog This stuff should go to PE Davis. Is master the right target branch?

kylog:
@kbrezina:
* master is the right branch; i.e. today, this would land in puppet 4.6, and thence PE Davis.
* @kbarber: I think the pdb variant of the report already has a top-level `noop` boolean. Does calling this new field `noop` here in the puppet report make things complicated in any way? Asking mostly out of ignorance.
* this LGTM but I'd like to hold off on merging until we get some agreement on how we want to handle the status fields (being discussed in a g-doc atm) lest we end up wanting to revisit this as a result of what we sort there

kbarber:
@kylog there is no conflict within our terminus, we'll have to implement a fallback to the old behaviour, if the new behaviour doesn't exist (for legacy support), but thats bau for this kind of work: https://github.com/puppetlabs/puppetdb/blob/master/puppet/lib/puppet/reports/puppetdb.rb#L41-L58

kbrezina:
@kylog @kbarber https://github.com/puppetlabs/puppet/pull/5007 is a proposal for noop_status implementation.

kylog:
@kbrezina apologies again for my slow turnaround time on the PRs and the g-doc. I think I'm caught up now.

Can you split these into 2 branches/PRs, one just for the PUP-6294 change, and one (can branch off the first) for the PUP-6369 change? Also I didn't look at why travis-ci reported failure. If it still does after the branch/PR split, let's figure out what's going on there.

kbrezina:
@kylog Oops, it was not my intention to push both commits into this branch - fixed. I'll rebase https://github.com/puppetlabs/puppet/pull/5007 once this PR gets merged.

kbrezina:
@kylog travis-ci looks better now

PR #4987, opened 2016-05-23
(PUP-6125) send verbose/debug messages to console when logdest is unset

pyther:
When the verbose or debug flags are used log output is sent to the
default logger (syslog/eventlog) and console regardless if logdest is
set.

Without this PR
`puppet agent --verbose` prints to syslog and console
`puppet agent --verbose --logdest=syslog` **prints to syslog and console**
`puppet agent --verbose --logdest=console` prints only to console

With this PR
`puppet agent --verbose` prints to syslog and console
`puppet agent --verbose --logdest=syslog` **only prints to syslog**
`puppet agent --verbose --logdest=console` prints only to console

### DIFF
--- a/lib/puppet/application.rb
+++ b/lib/puppet/application.rb
@@ -357,8 +357,10 @@ class Application
   end
 
   def setup_logs
-    if options[:debug] || options[:verbose]
-      Puppet::Util::Log.newdestination(:console)
+    unless options[:setdest]
+      if options[:debug] || options[:verbose]
+        Puppet::Util::Log.newdestination(:console)
+      end
     end
 
     set_log_level
--- a/spec/integration/agent/logging_spec.rb
+++ b/spec/integration/agent/logging_spec.rb
@@ -86,21 +86,22 @@ describe 'agent logging' do
       end
 
     else
+      if no_log_dest_set_in(argv)
+        it "when evoked with #{argv}, logs to #{expected[:loggers].inspect} at level #{expected[:level]}" do
+          # This logger is created by the Puppet::Settings object which creates and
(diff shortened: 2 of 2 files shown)


hlindberg:
The described behavior looks acceptable to me. Even if it somewhat changes the behavior - I think it can be argued that `--logdest syslog` also logging to console is a bug  (since it does not do the opposite for `--logdest console`).

hlindberg:
ping @kylog and @cprice404

MikaelSmith:
Yeah, this seems like a good compromise.

whopper:
👍 I just verified the change manually and it makes sense. I'm looking into the failing tests.

whopper:
@pyther the tests are failing because the tests (`spec/integration/agent/logging_spec.rb`, line 90) explicitly expect `Puppet::Util::Log.newdestination(logclass)` to always be called. This can be fixed by wrapping the `it` block with a conditional: `if no_log_dest_set_in(argv)`, like so:

```
 90       if no_log_dest_set_in(argv)                                                                 
 91         it "when evoked with #{argv}, logs to #{expected[:loggers].inspect} at level #{expected[:level]}" 
...
104
105           expect(Puppet::Util::Log.level).to eq(expected[:level])
106         end
107       end
```
`argv` will be an array that may or may not contain `--logdest`, and the `no_log_dest_set_in` method returns true if `argv` doesn't contain a `logdest`.
```
ARGV: ["--onetime", "--verbose", "--logdest=console"]
ARGV: ["--onetime", "--verbose"]
```

whopper:
Thanks @pyther!

PR #4988, opened 2016-05-24
(PUP-6345) In config reference, override default value of hiera_config

nfagerlund:
This setting's default value is now complicated, and the config reference gives up
and reports the literal location for the current local user. So like with the special
directories, we should override the default with a description.


HAIL9000:
@hlindberg @nfagerlund Any objection to me merging this?

nfagerlund:
Nope, fire away!

PR #4989, opened 2016-05-24
(PUP-6349) Don't quote mailalias recipients with +

vholer:
mailalias aliases povider double-quotes recipient address
if it contains non-alphanumeric and few selected characters (- . @).
This list can be extended with plus ("+") character as it's very
common in e-mail addresses and it's not necessary to quote such e-mail
address. Current behavior doesn't break anything (it's not bug),
on the other hand if Puppet works on same file with other tools,
Puppet keeps changing these recipients, although not necessary.

This simple patch adds plus character into list of accepted
characters for NOT quoting the recipient.


branan:
I like simple fixes. :+1:

PR #4990, opened 2016-05-25
(PUP-3713) Fix zfs type volume with volsize param

Miminoux:
Zfs of type "volume" (not filesystem) must be created with the -V option and the volsize param is only used in this case.
So puppet should not specify -o volsize=N but -V N . 

It has already been asked the 11/26/2014 and updated twice in 2015 : https://tickets.puppetlabs.com/browse/PUP-3713

Thx


hlindberg:
There are spec failures in the ZFS specs. When fixing that, please also update the commit message to read:
`(PUP-3713) fix zfs type volume with volsize param` as that makes our tooling do the right thing when correlating tickets to commits.

MosesMendoza:
Thanks for your contribution @Miminoux. I've subsumed these changes into https://github.com/puppetlabs/puppet/pull/6293, which also updates the impacted test.

Superseded by #6293

PR #4991, opened 2016-05-25
(doc) Recommend Puppet Extension add module name

ferventcoder:
Even with using the namespace ('user') as part of the PuppetX naming
convention, there still could be conflicts with other existing
modules in the same namespace. This could cause two files from
different modules that share the same name to overwrite each other
when they are synced to the same folder. It may not be immediately
obvious this is the reason when this occurs, and the results can
be disastrous. Due to these reasons, we should recommended that
Puppet Extensions submodules further define folder structure to
include the module name as well.

For example, a Puppet Extension for 'puppetlabs-powershell' would live at `lib/puppet_x/puppetlabs/powershell/<extension.rb>`.

### DIFF
--- a/lib/puppet_x.rb
+++ b/lib/puppet_x.rb
@@ -1,6 +1,12 @@
 # The Puppet Extensions Module.
 #
-# Submodules of this module should be named after the publisher (e.g. 'user' part of a Puppet Module name).
+# Submodules of this module should be named after the publisher (e.g.
+# 'user' part of a Puppet Module name). You should also add the module
+# name to avoid further conflicts in the same namespace. So the
+# structure should be
+# `lib/puppet_x/<namespace>/<module_name>/<extension.rb>`.
+# A Puppet Extension for 'puppetlabs-powershell' would live at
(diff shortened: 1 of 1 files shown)


ferventcoder:
cc @kylog @joshcooper @Iristyle

ferventcoder:
cc @puppetlabs/windows

joshcooper:
👍

Also cc @jtappa @nfagerlund. I'm not sure if we have module docs best practices around namespacing helper code?

ferventcoder:
@joshcooper we don't yet - I filed an issue to get that created. cc @jbondpdx

nfagerlund:
s/"It is further recommended as a best practice to additionally add the module name"/"You should also add the module name"/

nfagerlund:
Other than that, 👍

hlindberg:
This has been sitting around since June... @ferventcoder do you want to make the change suggested by nfagerlund? Even if not, the updated comment i better than it was earlier.

ferventcoder:
Yeah, I can make the adjustment.

ferventcoder:
@hlindberg @nfagerlund how does that look?

PR #4992, opened 2016-05-25
(PUP-3964) fix previous test commit teardown.

er0ck:
This test was previously not cleaning out its production site.pp, which
broke subsequent tests.  This test really should be using a temp
environmentpath, to be even safer, but this is a bit slower as it
requires a server restart.  This change also only cleans out the files
on the master, saving a bit of time.


shrug:
👍

seangriff:
This should fix it.

PR #4993, opened 2016-05-26
(PUP-6354) Make Enum values uniq

hlindberg:
Before this something like `Enum[a,b,a,b]` would keep all values. Thus, when iterating the values ´a, a, b, b´ would be produced. 
Now the values are made unique.


hlindberg:
Meh, done against the wrong branch...

PR #4996, opened 2016-05-26
(PUP-6159) acceptance: OSX Passwords not setting

er0ck:
This change adds a test for user resource password and salt setting,
which regressed in puppet 4x on OSX.  This test also ensures idempotency
of these properties.

stable frozen until PA 1.5.1 release.  merge after that, pls

[skip ci]

### DIFF
--- a/acceptance/tests/resource/user/should_set_passwd_and_salt.rb
+++ b/acceptance/tests/resource/user/should_set_passwd_and_salt.rb
@@ -1,0 +1,34 @@
+test_name 'C97905: user resource shall set password, salt and iterations' do
+  confine :except, :platform => /^eos-/ # See ARISTA-37
+  confine :except, :platform => /^cisco_/ # See PUP-5828
+
+  app_type = File.basename(__FILE__, '.*')
+  name = "#{app_type}_#{[*('a'..'z'),*('0'..'9')].shuffle[0,8].join}"
+
+  teardown do
+    step "delete the user, and group, if any" do
+      agents.each do |agent|
+        on(agent, puppet("resource user #{name} ensure=absent"), :accept_any_exit_code => true)
+        on(agent, puppet("resource group #{name} ensure=absent"), :accept_any_exit_code => true)
+      end
+    end
+  end
+
+  manifest = <<MANIFEST
+    user { '#{name}':
+      ensure     => 'present',
+      iterations => '46948',
+      password   => '9690da8dd8f90f6e3fed4f267c86110e29d75e8448efdacdee9bd5cc20f81a563c9e6c6c328694fac80910ba99508cc373525ac592b87fbec0ac1a1e26a51f01873c25f2450aa78e09c8498df0f11fa930c3f655e7aeed6bc61e8475ca84297b3a2273d31974ddd232e872d9b66be82d0246d094d60155c93c6b7a27ba1aa390',
+      salt       => '62133b77a7aeecf506ffe99d064b3f8c068344de0a619a573a871f2fd6fe9eaf',
+    }
+MANIFEST
+
+  step 'apply manifest creating user with password and salt' do
+    apply_manifest_on(agents, manifest)
+  end
+
+  step 'apply manifest again, to ensure idempotent' do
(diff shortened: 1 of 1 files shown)


whopper:
@er0ck I wonder if we should combine this with `tests/resource/user/should_allow_password_salt_iterations_osx.rb` (in master). That one is OSX specific but only because I used `dcsl` to verify the password on the system. That test would also probably  benefit from another `puppet apply` to ensure idempotency, like you're doing here.

er0ck:
hah.  for sure!  i didn't see that test, as i only searched through stable.  will PR against that one.  thanks!

joshcooper:
Well it's a few releases later... Can you retarget master?

er0ck:
yeah, i think this and the other test need combining, though.

er0ck:
the other test is actually better.  
we should have a test with a larger number of iterations.  but i can't get osx to retain the iteration value.  either we are doing it wrong, or they are.  
it's possible that we have to set the hashed password, salt and iterations in the plist all at the same time.  but this difficult as they evidently do not calculate the hash the same way we would in ruby/openssl.

i set the iterations using the directoryservice provider methods:
```
def set_shadow_hash_data(users_plist, binary_plist)
  binary_plist = Puppet::Util::Plist.string_to_blob(binary_plist)
  if users_plist.has_key?('ShadowHashData')
    users_plist['ShadowHashData'][0] = binary_plist
  else
    users_plist['ShadowHashData'] = [binary_plist]
  end
  write_users_plist_to_disk(users_plist)
end
def flush_dscl_cache
  system('/usr/bin/dscacheutil -flushcache')
end
def write_users_plist_to_disk(users_plist)
  Puppet::Util::Plist.write_plist_file(users_plist, "#{users_plist_dir}/#{@username}.plist", :binary)
end

password   = 'helloworld'
#salt       = '377e8b60e5fdfe509cad188d5b1b9e40e78b418f8c3f0127620ea69d4c32789c'
#iterations = 40000
salt       = 'd8265b2f081f6e37de39af91fc3c7c9d7fc048bff8325eb4ccb8d50829e67f72'
iterations = 46948
digest     = OpenSSL::Digest::SHA512.new
#hash       = OpenSSL::PKCS5.pbkdf2_hmac(password, salt, iterations, 128, digest).unpack('H*').first
#hash       = '837cf6a7f0ea8527405f0bdb6c68742dc028f450dbd3bb1e4e656691b9d540fb8be0c3a4ae315866740629fc7c5fd2afa0676ebdef756f19c4a2cf3cfbe99ca4a9ff87403d7eac4cf9c6872565cf7723c2595d976eedca2b712b0dca4dc53af62d7905509237e02cf9b5929fb7196d5c7ee55f58fe0b0f429e212ad4067573dd'

system("dscl . -create /Users/#{@username}")
flush_dscl_cache
users_plist = get_users_plist(@username)
shadow_hash_data = get_shadow_hash_data(users_plist)
set_salted_pbkdf2(users_plist, shadow_hash_data, 'iterations', iterations)
flush_dscl_cache
qxumiu97pm0lc1o:~ root#  dscl . -read /Users/i_am_some_user_1234987 iterations
No such key: iterations
qxumiu97pm0lc1o:~ root# dscl . -passwd /Users/i_am_some_user_1234987 helloworld
qxumiu97pm0lc1o:~ root# puppet resource user i_am_some_user_1234987
user { 'i_am_some_user_1234987':
  ensure     => 'present',
  iterations => 35971,
  password   => '7e031eede9edae424100a5546e2c5875460dccd3353b07fdc146ee05784c1c8fc6d7ce3a5c8927687a7eaec5e94d41ff881e63832a2be1b07ead223e0f2fd77ee768ea7e53f203eb7577edbf51d7d27d2000d899dd0bcd12302925061e2bf1bfdb57a034c0ddacdc92c76da8592563451722528bf1551cc6253bf58ed2836d0f',
  salt       => '55f787c0177eb0908640f126b5a90928945be62e292f3482d79ed62b1b70160e',
}
```
and as shown, the iterations is the osx 10.11 default.

PR #4998, opened 2016-05-27
Patch from PUP-6365

buzzdeee:
trusted facts not available when running "puppet master" webrick server

This makes the trusted facts available to $trusted.


branan:
This appears to be a no-op - `custom_exts.map` will return an Array of Hashes, just like what you're trying to do with the separate extensions array. I'm really not sure how this fixes the problem as reported.

buzzdeee:
@branan indeed, I did retest, and I have no clue, why I initially made that patch. I reverted my local patch, and it's still working the same way as without the patch.

so please disregard this PR, and sorry for the noise.

PR #4999, opened 2016-05-27
(PUP-6372) Remove unused Puppet::Transaction.failed? method

kylog:
The one use of Puppet::Transaction#failed? was removed in
https://github.com/puppetlabs/puppet/commit/84be695ddee004d7e358d9256f7929e1a2c400a3,
so we can now remove the method itself.

Note that the entire class is marked as api private, so no one should
be using the method.

### DIFF
--- a/lib/puppet/transaction.rb
+++ b/lib/puppet/transaction.rb
@@ -230,10 +230,6 @@ class Puppet::Transaction
     end
   end
 
-  def failed?(resource)
-    s = resource_status(resource) and s.failed?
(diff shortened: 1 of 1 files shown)


MosesMendoza:
thanks @Iristyle. The direction we ultimately took with PUP-6803 (the issue I was tackling) ended up not marking individual resources as failed (rather, the whole run) and so strictly speaking I think this PR is unrelated.

That being said, this seems reasonable.

PR #5001, opened 2016-05-31
(WIP)(PUP-6374) Minitar verify module tarballs

Iristyle:
- Previously it was possible, depending on platform, to create a
  .tar.gz that won't properly extract with Minitar on puppet versions
  prior to 3.8.7 or 4.4.0.  This has particularly caused problems
  with the DSC module on Windows.
  
  While the bug was fixed in #4740 /
  f888fb09fbef5f7c74ef799b3877ec6f7843a409, these older versions may
  still be unable to install modules as a result on platforms that
  rely on minitar - namely Windows, AIX and Solaris
- Provide some ability to catch these modules as they're packed on
  platforms with a native tar.  This should primarly be OSX, where
  bsdtar is the norm.  The use of gnu tar typically creates archives
  compatible with the old version of Puppet that does not contain
  the minitar fix.

### DIFF
--- a/lib/puppet/module_tool/applications/builder.rb
+++ b/lib/puppet/module_tool/applications/builder.rb
@@ -21,6 +21,7 @@ module Puppet::ModuleTool
         Puppet.notice "Building #{@path} for release"
         pack
         relative = Pathname.new(archive_file).relative_path_from(Pathname.new(File.expand_path(Dir.pwd)))
+        minitar_verify
 
         # Return the Pathname object representing the path to the release
         # archive just created. This return value is used by the module_tool
@@ -38,6 +39,22 @@ module Puppet::ModuleTool
 
       def archive_file
         File.join(@pkg_path, "#{metadata.release_name}.tar.gz")
+      end
+
+      def minitar_verify
+        begin
+          # assume valid if minitar inaccessible (i.e. immediate rescue loading gem)
+          require 'archive/tar/minitar'
+        rescue
+          return true
+        end
+
+        # minitar present, so assume invalid
+        files = Puppet::ModuleTool::Tar::Mini.new.entries(archive_file)
+        valid = files.all? { |f| f.full_name == f.name }
+        if !valid
+          raise Puppet::ModuleTool::Errors::ModuleToolError, "Archive contains invalid entries - may fail minitar extraction on Windows."
+        end
       end
 
       def pack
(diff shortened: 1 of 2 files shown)


lib/puppet/module_tool/applications/builder.rb
```diff
@@ -40,6 +41,22 @@ def archive_file
...
+          require 'archive/tar/minitar'
+        rescue
+          return true
+        end
+
+        # minitar present, so assume invalid
+        files = Puppet::ModuleTool::Tar::Mini.new.entries(archive_file)
+        valid = files.all? { |f| f.full_name == f.name }
+        if !valid
+          raise Puppet::ModuleTool::Errors::ModuleToolError, "Archive contains invalid entries - may fail minitar extraction on Windows."
```
Iristyle: I really want this to be a warning locally, but an error in CI.  Perhaps this needs to be controlled by a global setting.

lib/puppet/module_tool/applications/builder.rb
```diff
@@ -40,6 +41,22 @@ def archive_file
         File.join(@pkg_path, "#{metadata.release_name}.tar.gz")
       end
 
+      def minitar_verify
+        begin
+          # assume valid if minitar inaccessible (i.e. immediate rescue loading gem)
+          require 'archive/tar/minitar'
```
Iristyle: The general idea here is that modules would opt into this behavior by adding `minitar` to their `Gemfile`.

This wouldn't take effect under normal circumstances.

lib/puppet/module_tool/applications/builder.rb
```diff
@@ -40,6 +41,22 @@ def archive_file
...
+        rescue
+          return true
+        end
+
+        # minitar present, so assume invalid
+        files = Puppet::ModuleTool::Tar::Mini.new.entries(archive_file)
+        valid = files.all? { |f| f.full_name == f.name }
```
Iristyle: Not even positive this comparison is accurate - on OSX, when run with DSC module, the entries contain a bunch of oddly encoded strings.  Not sure if minitar just doesn't work right on OSX?

MosesMendoza:
This PR has been open for some time and unfortunately it seems like we're not going to be able to pick it up any time soon. I'm going to close it for now, but we can re-open if/when PUP-6374 is prioritized for next steps.

PR #5002, opened 2016-06-02
(PUP-6380) Fix HTTP file source for GET-only URIs

mcasper:
The Problem:
When using an HTTP URI as the source of a `file` type, URIs that only
work for GET requests fail.

The Use Case:
Pre-signed Amazon S3 URIs include the HTTP method in the signature,
which is usually GET for downloading files. This makes the URI return a
403 when responding to a HEAD request, but a 200 when responding to a
GET. Github uses Amazon S3 to store project 'releases', snapshots of a
repository at a certain release, usually prepackaged for different
operating systems and architectures.

The Code:
The `file` type fires off a HEAD request to the file's source when
trying to get metadata for the file,
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/source.rb#L188.
Through the indirectors, this eventually makes its way to
Puppet::Util::HttpProxy, when a method of `:head`.

The Changes:
There are two main changes here, the first being that we now pass both
the `uri.path` and `uri.query` to HTTP requests. URI query parameters
can contain necessary authorization tokens or signatures, and many URIs
will fail without them. Once support for Ruby 1.9 is dropped, the entire
URI object can be passed into HTTP requests, rather than needed to
reconstruct the URI.

The second change is when a HEAD request fails, attempt a GET request
to see if that works. This should only happen in the file metadata case,
as this is the only case I could find that specifically uses the
HttpProxy class to make a HEAD request.

If there's a better way to do this/implement this or it's already being
taken care of, please let me know, as I'm still pretty new to this repo :)

Issue ticket: https://tickets.puppetlabs.com/browse/PUP-6380

### DIFF
--- a/lib/puppet/util/http_proxy.rb
+++ b/lib/puppet/util/http_proxy.rb
@@ -157,7 +157,7 @@ module Puppet::Util::HttpProxy
   end
 
   # Retrieve a document through HTTP(s), following redirects if necessary.
-  # 
+  #
   # Based on the the client implementation in the HTTP pool.
   #
   # @see Puppet::Network::HTTP::Connection#request_with_redirects
@@ -172,7 +172,10 @@ module Puppet::Util::HttpProxy
 
     0.upto(redirect_limit) do |redirection|
       proxy = get_http_object(current_uri)
-      response = proxy.send(:head, current_uri.path)
+      # This is a workaround to support backwards compatibility for Ruby 1.9.3,
+      # once that support is dropped the entire URI object can be used.
+      request_path = "#{current_uri.path}?#{current_uri.query}"
+      response = proxy.send(:head, request_path)
 
       if [301, 302, 307].include?(response.code.to_i)
         # handle the redirection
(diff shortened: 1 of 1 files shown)


hlindberg:
@mcasper Thanks for the contribution. You need to look at the errors from travis and address those.

mcasper:
@hlindberg Corrections have been made, all tests are passing

branan:
Does this mean that if a `head` request fails, we'll do a `get` for the metadata, and then another `get` for the content if necessary? What are we doing with the content in the metadata request case? If it's actually coming down the wire, it would be nice to stash it away somewhere instead of just discarding it and fetching it again later. If we're closing the connection "as early as possible" to avoid downloading the whole file that's also probably fine.

mcasper:
Yes, right now it does mean that when a `head` request fails, it'll attempt a `get` request for the metadata, and then a separate `get` request for the content.

The metadata request is coming from [here](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/file/source.rb#L171), and currently does not cause any caching of the response to occur.

If we wanted to, we could wrap the `Net::HTTP::Proxy` used in `lib/puppet/util/http_proxy.rb` in an `HTTP::Cache` of sorts, caching responses by their URIs

branan:
I really really consider this a bug in S3 (not allowing HEAD on resources that allow GET is kinda crazy), but given that behavior, this is a very reasonable thing to want Puppet to do correctly.

We're going to poke this internally a bit to see if there are any truly terrible performance implications of doing two GET requests. We will probably merge this either way, but if there are performance issues we'll want to make sure we get that ticketed and are aware of it first (otherwise we'll forget ;) ). For now we're just leaving this un-merged so it stays on our radar/TODO list

Iristyle:
We need to delve into this PR a little bit more, as I agree with @branan here.  Issuing a fallback `GET` after a `HEAD` fails is a little odd IMHO.

jtappa:
superseded by #7051

PR #5003, opened 2016-06-02
(PUP-6381) Use systemd default service provider for Cumulus 3

ccaum:
Prior to this pull request, Cumulus Linux defaults to the debian provider which uses init as the parent.  Since Cumulus Linux 3.0 uses systemd, service resources are broken for CL 3 unless the user specifies the systemd provider.  This pull request changes the default provider for CL 3 to be systemd

### DIFF
--- a/lib/puppet/provider/service/debian.rb
+++ b/lib/puppet/provider/service/debian.rb
@@ -17,7 +17,7 @@ Puppet::Type.type(:service).provide :debian, :parent => :init do
   commands :invoke_rc => "/usr/sbin/invoke-rc.d"
   commands :service => "/usr/sbin/service"
 
-  defaultfor :operatingsystem => :cumuluslinux
+  defaultfor :operatingsystem => :cumuluslinux, :operatingsystemmajrelease => ['1','2']
   defaultfor :operatingsystem => :debian, :operatingsystemmajrelease => ['5','6','7']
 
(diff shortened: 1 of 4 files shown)


ccaum:
Is there a way to say that systemd should be used for CL 3 and greater?

ccaum:
I can't seem to get the tests to pass and can't get them to run locally on my workstation.  Any pointers appreciated.

branan:
@ScottGarman Yo, is this something your team should be tackling instead of client?

ScottGarman:
Yes, I'll take a look at the tests today.

ScottGarman:
@ccaum in your debian_spec.rb you just need to change your two :operatingsystem Facter expects to:

Facter.expects(:value).with(:operatingsystem).at_least_once.returns()

for CumulusLinux and Debian, respectively.

ccaum:
@ScottGarman Is that in place of the stubs?  If I replace the current expect, how do I test that it's the default provider?

ccaum:
Hey @geoffnichols, @ScottGarman said you would be the one to review and merge this. Can you review and merge if good?

geoffnichols:
This looks sane to me. 👍 

@shermdog, when you have a chance, can you please review?

shermdog:
:+1: Testing locally with VX image.

PR #5004, opened 2016-06-03
(PUP-6385) Make create_resources filter out undef relationships

hlindberg:
This adds logic that filters out all parameters that are undef from the
instruction built by create_resources.

This is done to prevent validation of relationships to flag
relationships to undef as an error.
This is done for all parameters since an undef parameter means that the
default value should be used. (The language evaluator does the same
thing).


hlindberg:
travis - timeout problem - job restarted

PR #5005, opened 2016-06-04
Immutable typed name 3.x

keymone:
TypedName can be made immutable value-type and instances of it cached. Significantly improves performance.

### DIFF
--- a/lib/puppet/pops/loader/loader.rb
+++ b/lib/puppet/pops/loader/loader.rb
@@ -171,10 +171,8 @@ class Puppet::Pops::Loader::Loader
     # of it cached.
     def self.new(type, name)
       @cache ||= {}
-      @cache.fetch(:"#{type}/#{name.gsub /^::/, ''}") do
-        puts "new TypedName registered: #{type}/#{name}, count: #{@cache.size}"
-        super
-      end
+      key = :"#{type}/#{name.to_s.gsub /^::/, ''}"
+      @cache.fetch(key) { @cache[key] = super }
     end
 
     # Force identity equality since this is now immutable
--- a/lib/puppet/pops/loader/loader.rb
+++ b/lib/puppet/pops/loader/loader.rb
@@ -166,15 +166,23 @@ class Puppet::Pops::Loader::Loader
       @hash
     end
 
-    def ==(o)
-      o.class == self.class && type == o.type && name == o.name
+    # For all intents and purposes TypedName can be immutable,
+    # equality operations aliased to identity equality and instances
+    # of it cached.
+    def self.new(type, name)
+      @cache ||= {}
+      @cache.fetch(:"#{type}/#{name.gsub /^::/, ''}") do
+        puts "new TypedName registered: #{type}/#{name}, count: #{@cache.size}"
+        super
+      end
     end
 
-    alias eql? ==
+    # Force identity equality since this is now immutable
+    alias == equal?
+    alias eql? equal?
 
     def to_s
       "#{type}/#{name}"
     end
   end
 end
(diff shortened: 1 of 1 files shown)


keymone:
i think this is applicable to 4.x versions too

hlindberg:
Thanks for the contribution - performance is always nice (and caching is always hard) ;-)

The approach taken here means that there will be memory leakage as every type name ever loaded in any environment will be kept in memory in a class member variable. A real cache implementation would need to have cache eviction, or the cache must have a lifecycle that is the same as the environment. (There is a way to achieve that by using an adapter pattern that associates the cache with the environment). Both such implementations will however add overhead that detracts from the performance gained by caching in the first place (it may still be of value).

Also a couple of house-keeping things:
* A Jira ticket is needed where the problem is described, and the performance improvements are shown
* Commit messages must comply with the form "(PUP-<ticketnbr>) <Imperative formed sentence>"

The speed up may be applicable to 4.x as well. The TypedName has undergone some changes in 4.x.

Ping @thallgren.

As this PR stands it cannot be accepted, and I am marking it as blocked while the issues are being resolved.

hlindberg:
On second thought - the cache should follow the Loaders life cycle (loaders lifecycle is already managed).

hlindberg:
Code is refactored on master and the TypedName is broken out into a separate ruby file.
It may be worth considering just using a Symbol created from its parts - that way Ruby would deal with the memory management of the instances at least from Ruby 2.2 when Symbol GC was introduced (again ping @thallgren). That approach cannot be taken with Puppet 3.x though as it runs on Ruby 1.8.7. For 4.x we are still on Ruby 2.1 in AIO and plan to upgrade. We do however support Ruby 1.9.3 still. This means the logic must be different depending on Ruby runtime version if an approach using Symbol is chosen.

keymone:
@hlindberg thanks for feedback.

yes, objects will be retained in memory but i don't think it's such a big deal. in our fairly large repo (~2000 puppet types/classes and ~200 puppet functions) i only see about 100 cache entries during compilation run. granted i've tested it against puppet 3.8 and i noticed that only TypedName(:function, XXX) are registered - that is probably not the case for puppet 4. if we're generous this will be 10K objects in long-running processes consuming ~1000 bytes each - no where close to system-impacting memory leak.

i don't think there is need to tie this to environment-eviction and loaders life-cycle, because these objects don't carry any semantics that are different between environments. i could implement something like LRU-eviction and cap number of memoized entries - would that work?

hlindberg:
Easiest is to store the cache in the loaders instance via an adapter. The 4.x loaders gets used more and more and eventually will hold every name of every loaded thing (define, class, function, data type) in every environment unless the cache gets evicted. In 3.x/future parser the 4.x loaders are only used for functions.
There are sites where there are several hundred environments, and where the environments are quite different. For such sites memory consumption is a big issue (perhaps even more so than performance).

I am also curious what a symbol based implementation would give us for Ruby 2.2.0 as we get the cache/GC for free (requires more refactoring though as TypedName is expected). Using symbols and not doing GC on them may be ok even if there is "leakage" as the overhead would be less than with the TypedName objects.

In 4.x the TypedName holds one more element (a name authority) which would also benefit from memoization.

hlindberg:
This PR has now gone stale. It is still interesting to proceed with optimization of the TypedName. We are however close to the EOL of 3.x and the work should really be retargeted to master (4.x) where an approach using symbols could be explored. Also note the earlier remark that TypedName has undergone some change from 3.x.
If continuing on this work it seems best to open a new PR, and I am therefore closing this.

PR #5006, opened 2016-06-05
(maint) Remove 'use's of :graphing settings section

kylog:


joshcooper:
Although there are no default settings defined under the `graphing` section, if a user specified settings in puppet.conf under the `[graphing]` setting (by accident), I was concerned this change might cause the setting to no longer be applied, e.g.

```ini
[graphing]
lastrunfile = /tmp/foo.yaml
```

But it looks like puppet will report an error if puppet.conf contains a section it doesn't know about, so this change should be a noop:

```
$ bundle exec puppet apply -e 'notify { "boo": }'
Error: Could not parse /Users/josh/.puppetlabs/etc/puppet/puppet.conf: Illegal section 'graphing' in config file /Users/josh/.puppetlabs/etc/puppet/puppet.conf at line 1. The only valid puppet.conf sections are: [main, master, agent, user]. Please use the directory environments feature to specify environments. (See https://docs.puppet.com/puppet/latest/reference/environments.html)
```

PR #5007, opened 2016-06-06
(PUP-6369) Add noop_pending flag to puppet report

kbrezina:
`noop_pending` flag is set to `true` when noop resource events are available

### DIFF
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -16,6 +16,7 @@
         "kind",
         "status",
         "noop",
+        "noop_pending",
         "environment"
     ],
 
@@ -143,6 +144,11 @@
 
         "noop": {
             "description": "Whether the puppet run was started in noop mode",
+            "type":        "boolean"
+        },
+
+        "noop_pending": {
+            "description": "Whether there are changes that we decided not to apply because of noop",
             "type":        "boolean"
         },
 
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -113,6 +113,11 @@ class Puppet::Transaction::Report
   #
   attr_reader :noop
 
+  # Whether there are changes that we decided not to apply because of noop
+  # @return [Boolean]
+  #
+  attr_reader :noop_pending
+
   def self.from_data_hash(data)
     obj = self.allocate
     obj.initialize_from_hash(data)
@@ -163,6 +168,11 @@ class Puppet::Transaction::Report
   end
 
   # @api private
+  def has_noop_events?(resource)
+    resource.events.any? { |event| event.status == 'noop' }
+  end
+
+  # @api private
   def prune_internal_data
     resource_statuses.delete_if {|name,res| res.resource_type == 'Whit'}
   end
@@ -177,6 +187,7 @@ class Puppet::Transaction::Report
     add_metric(:changes, {"total" => change_metric})
     add_metric(:events, calculate_event_metrics)
     @status = compute_status(resource_metrics, change_metric)
+    @noop_pending = @resource_statuses.any? { |name,res| has_noop_events?(res) }
   end
 
   # @api private
@@ -198,6 +209,7 @@ class Puppet::Transaction::Report
     @environment = environment
     @status = 'failed' # assume failed until the report is finalized
     @noop = Puppet[:noop]
+    @noop_pending = false
   end
 
   # @api private
@@ -209,6 +221,7 @@ class Puppet::Transaction::Report
     @environment = data['environment']
     @status = data['status']
     @noop = data['noop']
+    @noop_pending = data['noop_pending']
     @host = data['host']
     @time = data['time']
 
@@ -263,6 +276,7 @@ class Puppet::Transaction::Report
       'kind' => @kind,
       'status' => @status,
       'noop' => @noop,
+      'noop_pending' => @noop_pending,
       'environment' => @environment,
 
       'logs' => @logs,
--- a/api/docs/http_report.md
+++ b/api/docs/http_report.md
@@ -59,6 +59,7 @@ example is formatted for readability)
      "kind"=>"apply",
      "status"=>"unchanged",
      "noop"=>false,
+     "noop_pending"=>false,
      "environment"=>"test_environment",
      "logs"=>
       [{"level"=>"warning",
--- a/spec/unit/transaction/report_spec.rb
+++ b/spec/unit/transaction/report_spec.rb
@@ -236,7 +236,6 @@ describe Puppet::Transaction::Report do
       end
     end
 
-
     [:time, :resources, :changes, :events].each do |type|
       it "should add #{type} metrics" do
         @report.finalize_report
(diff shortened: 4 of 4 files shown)


kbrezina:
ping @kylog

lib/puppet/transaction/report.rb
```diff
@@ -158,6 +168,21 @@ def compute_status(resource_metrics, change_metric)
   end
 
   # @api private
+  def has_noop_change_events?(resource)
+    resource.events.any? { |event| (event.status == 'noop' && event.name == :content_changed) }
```
kbarber: Events that include changes don't always have a `name` field with value `:content_changed`, in fact, that field is a composition of `property + "_changed"` (a little useless I know). So actually, you're only catching `content` field change events specifically with this mechanism only, I presume you were testing with `file { "foo": content => "blah" }`.

kbarber: Point in case, this is a mode change on file (in yaml):

```
    events:
    - !ruby/object:Puppet::Transaction::Event
      audited: false
      property: mode
      previous_value: '0644'
      desired_value: '777'
      historical_value: '0644'
      message: mode changed '0644' to '0777'
      name: :mode_changed
      status: success
      time: 2016-06-07 11:05:46.888644000 +01:00
```

Notice it is `:mode_changed`, not `:content_changed`.

kbrezina: There are probably more change options:

```
    events:
    - !ruby/object:Puppet::Transaction::Event
      audited: false
      property: content
      previous_value: "{md5}4e4cdf645e477bdcee0810dccc9c538d"
      desired_value: "{md5}81e4d91298307865cae873f56768b2f4"
      historical_value:
      message: current_value {md5}4e4cdf645e477bdcee0810dccc9c538d, should be {md5}81e4d91298307865cae873f56768b2f4
        (noop)
      name: :content_changed
      status: noop
      time: 2016-06-07 05:37:14.165072743 -07:00
```

Maybe, it should compare `previous_value` and `desired_value` instead.

kbarber: Afaik, all events are "changes" or "proposed changes" @kbrezina, (the exception being audit events). The presence of an event suggests there was a change. You shouldn't need to compare previous/desired.

kbrezina:
@kbarber @kylog I've added spec tests and improved the algorithm based on comments from @kbarber. Could you take a look?

kbrezina:
@kylog @kbarber Anything to improve here?

kbarber:
+1, looks good, nothing immediate came to mind.

kylog:
LGTM. Ping @branan also.

branan:
To be clear on what I'm reading in this code: When an agent run is triggered with `--noop` AND there are resources with `noop=false` set AND those resources change but all "normal" resources are in-sync THEN the normal status will be 'changed' and the noop status will be 'unchanged'

I think if this is the expected behavior in that case, then it makes sense to also set `noop_status`  to the status of any explicitly `noop=true` resources in an enforcement run. Right now the behavior is different between the different values of the `--noop` flag

kbrezina:
The status of any explicitly `noop=true` resources in an enforcement run is not important for us. I'm not sure if anyone else would need this information.
When running in noop mode, the `noop_status` is an estimate of the `status` in enforcement mode.

kbrezina:
I can remove the `if @noop` from the `compute_noop_status` but then the `noop_state` would be set to `failed` or `unchanged` in enforcement mode even if there are no `noop` resources.

branan:
I'm just arguing against special-casing things that don't need to be special-cased. Since we already have the separate `noop` flag (we don't rely on `noop_status` being nil to determine run mode or anything like that), we may as well always set the status.

I think the semantic of "unchanged" in that case is fine - it just means that there are no pending changes. The same would happen in a `--noop` run where every resource was marked as `noop=false` - again, there are no pending changes. That field can't be trusted to provide information about the existence of noop resources that are in-sync, regardless of whether they come from individual settings or the global flag.

"failed" raises an interesting question - Now that I think about it more, I almost think that `noop_status` should always reflect the existence (or not) of pending changes, and rely on the global status being "failed" to indicate a failed run. Mirroring that `failed` flag doesn't give the downstream report processor anymore information, and actually HIDES information about pending resources in an overall failed run.


All that being said, and now that I've had a bit more time to mull this over: would this be better as a simple boolean named `pending` or something along those lines. That would give us:

* The `status` field, indicating whether the run failed (possibly making changes), changed the system successfully, or left the system unchanged
* The `noop` field, indicating whether `--noop` was passed to the agent
* The `pending` field, indicating that one or more resources set to noop (either by the flag or in the manifest) want to make changes.

I believe from these three fields the console can still synthesize whatever view onto an idea of a "noop" run it wants to, and they much more closely map to how Puppet actually works. @kylog and @kbarber thoughts on this?

kbrezina:
@branan Thanks for your feedback. I agree that this should provide all we need for PE Console. 

I'm just thinking if the two states (in-sync and out-of-sync) are sufficient for the `pending` field. AFAIK, we cannot have client side failures in noop runs now. However, shouldn't we be ready for that? For example a dependency cycle could cause this kind of failure, right? Currently, the dependency cycles are not properly detected - see https://tickets.puppetlabs.com/browse/PE-5396.

kbrezina:
ping @scottyw

kbarber:
I spoke with @kbrezina on @branan's last comment, and I think his point is valid. 'failed' masks things, and is redundant, and pending/notpending has enough resolution combined with noop to give us the details that 'a noop run failed with pending'. I think anything pending is interesting on its own as well.

kbrezina:
Based on comments from @branan and @kbarber I reimplemented this PR.

@kbarber @branan @kylog Could you provide feedback? I'd like to finalize this feature.

api/schemas/report.json
```diff
@@ -146,6 +147,11 @@
             "type":        "boolean"
         },
 
+        "noop_pending": {
+            "description": "Whether noop resource events exist",
```
branan: I'd hope that elsewhere in the report schema was specify what a "noop event" is, but maybe it's worthwhile to wordsmith this a little bit to indicate that it means "there are changes that we decided not to apply because of noop". @jtappa what are your thoughts here?

branan:
Beyond the docs question, I'm :+1: on this at this point. Thanks for making the changes, @kbrezina

api/schemas/report.json
jtappa: I agree, this description could be a little more helpful to the user. Perhaps, "Whether or not changes will be applied due to existing noop events." ?

kbrezina: @jtappa The Branan's description is more exact. Should I replace the description with "Whether there are changes that we decided not to apply because of noop"?

kbrezina:
I've changed the description in the report schema.

@branan @kbarber @kylog Is it mergeable now?

kbarber:
+1

branan:
:+1: I'm not sure I'm entirely happy with the phrasing in that doc line yet, but if I come up with something better I'll just open a `(doc)` PR for it. I think this is ready.

Thanks again for being patient with my last-minute redesign suggestion, and much thanks to @kbarber for helping out on the review

PR #5009, opened 2016-06-07
Use systemd by default on Ubuntu 16.10

nacc:


kylog:
Thanks @nacc! 

One minor nit: can you update the commit message on the refactor commit to prepend "(maint)" to the summary line, e.g. so it looks like:
`(maint) Refactor ubuntu version checking`. (We have a release script that keys off certain formats in the commit summary lines, and a matching travis check for that same format.)

nacc:
@kylog Just updated the commit message. Will update the Jira issue as well! Thanks for the quick feedback!

nacc:
@kylog apologies for asking in a closed issue, but from a maintenance perspective, would it make sense to adjust the checks so that systemd is assumed to be the default on >= 15.04? e.g. in lib/puppet/provider/service/systemd.rb:

defaultfor :operatingsystem => :ubuntu, (:operatingsystemmajrelease <=> "15.04") == 1

Otherwise, we're going to have to make similar changes for each new release of Ubuntu (and it will be broken when we do). 

I'm not sure how to modify/update the other functionality, though, in spec/unit/provider/service/systemd_spec.rb, correspondingly:

[ '15.04', '15.10', '16.04', '16.10' ].each do |ver|

(I don't know Ruby very well, I apologize)

nacc:
@kylog alternatively, could we flip the default so that Ubuntu is assumed to have systemd and just put the (now finite) set of releases that did have upstart as the default in a list?

kylog:
@nacc, yep that makes total sense, and takes a little work in `defaultfor`. I think that's what is being tackled in #5069.

PR #5010, opened 2016-06-07
(PUP-6150) Don't include IP address in auth error response

rlinehan:
Previously, an unauthenticated client could make a request to a route
protected by auth.conf and get back an error message including the IP address
and rule description. This is more information than is probably good to
provide over http, especially since it could potentially include internal IPs
of proxy servers, depending on how those are configured. This commit makes it
so that the response includes minimal information.

### DIFF
--- a/lib/puppet/network/rights.rb
+++ b/lib/puppet/network/rights.rb
@@ -53,17 +53,7 @@ class Rights
 
     # if we end up here, then that means we either didn't match or failed, in any
     # case will return an error to the outside world
-    host_description = args[:node] ? "#{args[:node]}(#{args[:ip]})" : args[:ip]
-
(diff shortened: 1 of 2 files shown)


rlinehan:
The ticket mentions that we could log the IP address and rule info but not return it in the response. However, that seemed like it would require a bigger refactor than is worth the effort.

cprice404:
+1

rlinehan:
Hmmm... I used the completely wrong jira ticket on this somehow. Going to push up a new commit with that changed and change the title.

camlow325:
Functionally tested this PR locally with an unauthorized client request.  Omits all of the sensitive info - including the ip address and reference to the corresponding rule - as advertised.

So, +1 from me and am merging this....

PR #5011, opened 2016-06-08
(PUP-6394) Relax code id requirement when finding capabilities

nicklewis:
Previously when looking up capability resources, the compiler would
lookup resources resources strictly by code id. This effectively meant
that a node couldn't run against a certain code version until all of its
dependencies had run against that version.

For iterative development on changes that don't actually touch those
dependencies, this is an extremely onerous requirement. Further, it's
not sufficient for a node to only run after its dependencies; if the
capabilities it requires are changing, it needs to be run *quickly*
after its dependencies, which is a property that isn't satisfied by the
ordering requirement alone.

In order to provide more flexibility around running agents, particularly
when capabilities aren't being affected by a given change, we now use
any available version of a consumed capability resource, rather than
requiring it come from the current code id.

This creates the possibility that a capability may be moved from one
node to another node, causing two copies to exist in the database for a
time. In that case we repeat the lookup, this time using code id to
disambiguate. That disambiguation prevents the catalog from referring to
the wrong version of a capability resource, though it does reintroduce
the requirement that the producer has run against the current code id.

### DIFF
--- a/lib/puppet/resource/capability_finder.rb
+++ b/lib/puppet/resource/capability_finder.rb
@@ -12,7 +12,11 @@ require 'json'
 # @api private
 module Puppet::Resource::CapabilityFinder
 
-  # Looks the capability resource from PuppetDB.
+  # Looks up a capability resource from PuppetDB. Capability resources are
+  # required to be unique per environment and code id. If multiple copies of a
+  # capability resource are found, the one matching the current code id is
+  # used.
+  #
   # @param environment [String] environment name
   # @param code_id [String,nil] code_id of the catalog
   # @param cap [Puppet::Type] the capability resource type instance
@@ -22,6 +26,29 @@ module Puppet::Resource::CapabilityFinder
       raise Puppet::DevError, 'PuppetDB is not available'
     end
 
+    resources = search(environment, nil, cap)
+
+    if resources.size > 1 && code_id
+      Puppet.debug "Found multiple resources when looking up capability #{cap}, filtering by code id #{code_id}"
+      resources = search(environment, code_id, cap)
+    end
+
+    if resources.size > 1
+      raise Puppet::DevError,
+        "Unexpected response from PuppetDB when looking up #{cap}:\n" \
+        "expected exactly one resource but got #{resources.size};\n" \
+        "returned data is:\n#{resources.inspect}"
+    end
+
+    if resource_hash = resources.first
+      instantiate_resource(resource_hash)
+    else
+      Puppet.debug "Could not find capability resource #{cap} in PuppetDB"
+      nil
+    end
+  end
+
+  def self.search(environment, code_id, cap)
     query_terms = [
       'and',
       ['=', 'type', cap.type.capitalize],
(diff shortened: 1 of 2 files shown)


lib/puppet/resource/capability_finder.rb
```diff
@@ -22,6 +26,29 @@ def self.find(environment, code_id, cap)
...
+        "Unexpected response from PuppetDB when looking up #{cap}:\n" +
+        "expected exactly one resource but got #{resources.size};\n" +
+        "returned data is:\n#{resources.inspect}"
+    end
+
+    if resource_hash = resources.first
+      instantiate_resource(resource_hash)
+    else
+      Puppet.debug "Could not find capability resource #{cap} in PuppetDB"
+      nil
```
briancain: Is this nil really required here? Maybe pry is lying to me, but it seems like if you don't explicitly return nil, you will still get nil back when find is called

```
[1] pry(main)> def mymethod(test)
[1] pry(main)*   if test
[1] pry(main)*     puts "yay"
[1] pry(main)*   else
[1] pry(main)*     puts "no"
[1] pry(main)*     nil
[1] pry(main)*   end
[1] pry(main)* end
=> nil
[2] pry(main)> mymethod(false)
no
=> nil
[3] pry(main)> mymethod(true)
yay
=> nil
```

```
[5] pry(main)> def mymethod(test)
[5] pry(main)*   if test
[5] pry(main)*     puts "yay"
[5] pry(main)*   else
[5] pry(main)*     puts "no"
[5] pry(main)*   end
[5] pry(main)* end
=> nil
[6] pry(main)> mymethod(true)
yay
=> nil
[7] pry(main)> mymethod(false)
no
=> nil
```

nicklewis: Without the `nil`, it will return whatever the return value of `Puppet.debug` is. I think that that's only `nil` if you're not running with debug level logging. If you have debug on, it seems like it will return a Log object. In any case, relying on the return value of `Puppet.debug` seems bad.

briancain: Ah yeah, that seems gross.

lib/puppet/resource/capability_finder.rb
```diff
@@ -22,6 +26,29 @@ def self.find(environment, code_id, cap)
...
+    end
+
+    if resources.size > 1
+      raise Puppet::DevError,
+        "Unexpected response from PuppetDB when looking up #{cap}:\n" +
+        "expected exactly one resource but got #{resources.size};\n" +
+        "returned data is:\n#{resources.inspect}"
+    end
+
+    if resource_hash = resources.first
```
briancain: Where does resource_hash come from...?

nicklewis: It's being set on this line by the `=`.

briancain: Ah....I thought the single = was a bug/typo here.

lib/puppet/resource/capability_finder.rb
```diff
@@ -22,6 +26,29 @@ def self.find(environment, code_id, cap)
...
+    if resources.size > 1 && code_id
+      Puppet.debug "Found multiple resources when looking up capability #{cap}, filtering by code id #{code_id}"
+      resources = search(environment, code_id, cap)
+    end
+
+    if resources.size > 1
+      raise Puppet::DevError,
+        "Unexpected response from PuppetDB when looking up #{cap}:\n" +
+        "expected exactly one resource but got #{resources.size};\n" +
+        "returned data is:\n#{resources.inspect}"
```
hlindberg: Picky, it is better to use a `\` instead of a `+` as the `\` is handled by the ruby parser (creating a single string) vs. the `+` which creates individual strings that are then concatenated,

lib/puppet/resource/capability_finder.rb
```diff
@@ -95,8 +81,40 @@ def self.query_puppetdb(query)
...
+    # The format of the response body is documented at
+    #   http://docs.puppetlabs.com/puppetdb/3.0/api/query/v4/resources.html#response-format
+    unless result.is_a?(Array)
+      raise Puppet::DevError,
+      "Unexpected response from PuppetDB when looking up #{cap}: " +
+        "expected an Array but got #{result.inspect}"
+    end
+
+    result
   rescue JSON::JSONError => e
```
hlindberg: top level rescue sometimes cause problems (have seen it in testing/mocking) - better to always use a begin/end

nicklewis:
@hlindberg I squashed in a change to switch `\` for `+` and to make the `begin`/`rescue` explicit.

hlindberg:
@nicklewis do you have this on your board for review, or would you like me to merge?

nicklewis:
@hlindberg You can go ahead and merge if you're ready.

PR #5012, opened 2016-06-08
(PUP-3768) Scheduled_task keeps reapplying

DLuCJ:
Initially, puppet would re-apply a scheduled task even though
there was no change.  The reason being, only the first 256
characters of the arguments parameter (or rather, any
parameter that accepts a string as input) were being read in
for comparison.  Therefore, the first 256 characters of the
string were being compared to the full string, resulting in
changed status.

This commit increases the comparison limit for the command
and arguments parameters to 260 and 4096, respectively.
The 260 character limit is due to Windows API MAX_PATH
limit for pathnames.  The arguments parameter doesn't appear
to have an upper bound, so 4096 was chosen as a reasonably
large limit.  An integration test was included and the
should_modify acceptance test was changed, both to exercise
scheduled_task with larger string params.

### DIFF
--- a/acceptance/tests/resource/scheduled_task/should_modify.rb
+++ b/acceptance/tests/resource/scheduled_task/should_modify.rb
@@ -3,6 +3,8 @@ test_name "should modify a scheduled task"
 name = "pl#{rand(999999).to_i}"
 confine :to, :platform => 'windows'
 
+verylongstring = 'a' * 1024
+
 agents.each do |agent|
   # Have to use /v1 parameter for Vista and later, older versions
   # don't accept the parameter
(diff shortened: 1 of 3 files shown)


acceptance/tests/resource/scheduled_task/should_modify.rb
```diff
@@ -3,6 +3,13 @@
 name = "pl#{rand(999999).to_i}"
 confine :to, :platform => 'windows'
 
+verylongstring = <<-STR
```
Iristyle: Ruby has the ability to create a long string like:

```ruby
'a' * 1024
```

spec/integration/util/windows/taskscheduler_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      STR
+
+      subject = Puppet::Util::Windows::TaskScheduler.new(resource[:name], dummy_time_trigger)
+      subject.application_name=(resource[:command])
+      subject.parameters=(verylongstring)
+
+      expect(subject.parameters).to eq(verylongstring)
+    end
+  end
+end
```
Iristyle: Missing the newline here.

spec/integration/util/windows/taskscheduler_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+        'start_minute'            => now.min,
+        'trigger_type'            => Win32::TaskScheduler::ONCE,
+    }
+  end
+  
+end
+
+
+
+describe "Puppet::Util::Windows::TaskScheduler", :if => Puppet.features.microsoft_windows? do
```
Iristyle: I think we want these tests in https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb

glennsarti: Seems to make more sense in integration because we're testing the entire class, not specific methods, and we are actually calling real Windows APIs.

acceptance/tests/resource/scheduled_task/should_modify.rb
DLuCJ: Done

spec/integration/util/windows/taskscheduler_spec.rb
DLuCJ: Done

Iristyle:
Closed in favor of #5035

PR #5013, opened 2016-06-10
Make $trusted facts work with apache/unicorn or nginx/unicorn

buzzdeee:
via X-SSL-Client-Cert variable

See PUP-6366

Also see #PR4997 that is for/against Puppet 3.x branch.

### DIFF
--- a/lib/puppet/network/http/rack/rest.rb
+++ b/lib/puppet/network/http/rack/rest.rb
@@ -103,7 +103,21 @@ class Puppet::Network::HTTP::RackREST
     # NOTE: The SSL_CLIENT_CERT environment variable will be the empty string
     # when Puppet agent nodes have not yet obtained a signed certificate.
     if cert.nil? || cert.empty?
-      nil
+      # When running with unicorn, the SSL_CLIENT_CERT variable is not available
+      # in the environment, therefore we have to pass a header: 'X-SSL-Client-Cert'
+      cert = request.env['HTTP_X_SSL_CLIENT_CERT']
+      if cert.nil? || cert.empty?
+        nil
+      else
+        # in contrast to the environment variable, the client cert is passed in
+        # as single string, therefore restore the certificate to a valid pem
+        # encoded certificate
+        cert.gsub!(/ /, "\n")
+        cert.gsub!(/BEGIN\nCERT/, "BEGIN CERT")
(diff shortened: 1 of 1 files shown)


lib/puppet/network/http/rack/rest.rb
```diff
@@ -103,7 +103,21 @@ def client_cert(request)
     # NOTE: The SSL_CLIENT_CERT environment variable will be the empty string
     # when Puppet agent nodes have not yet obtained a signed certificate.
     if cert.nil? || cert.empty?
-      nil
+      # When running with unicorn, the SSL_CLIENT_CERT variable is not available
+      # in the environment, therefore we have to pass a header: 'X-SSL-Client-Cert'
+      cert = request.env['HTTP_X_SSL_CLIENT_CERT']
```
adrienthebo: If this value is sent as a header, is it possible to explicitly pull this value with `request.params['HTTP_X_SSL_CLIENT_CERT']` instead of using the environment? This will make the code more clear on its intent.

adrienthebo: On second review this is a header and not a param, and Rack might not directly expose headers in a separate object and only provides them in the request environment. If it's possible to extract this value as a header I would prefer that, but if this the best way to do this then I'm fine with it.

adrienthebo:
It looks like some of the linting for this pull request failed; the Travis CI test at (https://travis-ci.org/puppetlabs/puppet/jobs/136823269) indicates that the commit message needs to be reformatted for consistency.

lib/puppet/network/http/rack/rest.rb
buzzdeee: I tried with request.params, but the extensions were an empty hash, reverting back to request.env, and the hash was filled again.

branan:
@buzzdeee We're still waiting on the commit message format to be fixed up before we can merge this. It would also be cool if you could figure out how to write tests for it (although not a deal-breaker if you don't, since we consider this code deprecated)

buzzdeee:
damnit, @branan ,thanks for the ping. I somehow missed that earlier comment, would have fixed that much earlier.

HAIL9000:
Closing this one in favor of #5169 which has the reformatted commit message

PR #5014, opened 2016-06-10
PUP-4904 Correctly format output from puppet describe

seangriff:
Prior to this commit, if a type has very long strings in its @doc value,
then the output becomes garbled.  This commit fixes a regular expression
so that long lines are properly are properly split into managable lengths
for output fomatting.


hlindberg:
A unit test should be added that shows that the change works as intended.

PR #5015, opened 2016-06-13
(DOC) create 3x stubs for all new 4x functions

hlindberg:
This moves the documentation of all newly added 4.x functions from the (real) implementation file to a 3.x stub to allow the documentation to be processed by the documentation pipeline.

Minor edits have been done to comply with the style of other functions, as well as fixing a few typos.
A small text stub is left in the implementation file.

The intention is to move the documentation back again and to remove the 3.x stubs when the documentation pipeline can process 4.x function documentation output by Puppet Strings.


hlindberg:
Ping @jtappa

hlindberg:
Ping @kylog (4.x functions in puppet 4.4 and 4.5. for which no docs were produced).

hlindberg:
meh, some multibyte UTF-8 chars snuck in there... fixing.

lib/puppet/parser/functions/lest.rb
```diff
@@ -0,0 +1,49 @@
+Puppet::Parser::Functions::newfunction(
+  :lest,
+  :type => :rvalue,
+  :arity => -2,
+  :doc => <<-DOC
+Call a [lambda](https://docs.puppet.com/puppet/latest/reference/lang_lambdas.html)
+(which should accept no arguments) if the argument given to the function is `undef`.
+Returns the result of calling the lambda if the argument is `undef`, otherwwise the
```
jtappa: "otherwise" typo

lib/puppet/parser/functions/new.rb
```diff
@@ -0,0 +1,530 @@
+Puppet::Parser::Functions::newfunction(
+  :new,
+  :type => :rvalue,
+  :arity => -1,
+  :doc => <<-DOC
+Creates a new instance/object of a given data type.
+
+This functions makes it possible to create new instances of
```
jtappa: functions  => function

lib/puppet/parser/functions/new.rb
```diff
@@ -0,0 +1,530 @@
...
+  String $value,
+  Radix $radix = 10
+)
+
+function Integer.new(
+  Variant[Numeric, Boolean] $value
+)
+```
+
+* When converting from `String` the default radix is 10
```
jtappa: Each line in this bulleted list should end in `.`, including the ones currently ending with commas.

lib/puppet/parser/functions/new.rb
```diff
@@ -0,0 +1,530 @@
...
+* If radix is not specified an attempt is made to detect the radix from the start of the string:
+  * `0b` or `0B` is taken as radix 2,
+  * `0x` or `0X` is taken as radix 16,
+  * `0` as radix 8,
+  * all other are decimal
```
jtappa: all other => All others

lib/puppet/parser/functions/new.rb
```diff
@@ -0,0 +1,530 @@
...
+```
+
+Conversion to String
+--------------------
+
+Conversion to `String` is the most comprehensive conversion as there are many
+use cases where a string representation is wanted. The defaults for the many options
+have been chosen with care to be the most basic "value in textual form" representation.
+The more advanced forms of formatting are intended to enable writing special purposes formatting
+functions in the puppet language.
```
jtappa: puppet => Puppet

lib/puppet/parser/functions/new.rb
```diff
@@ -0,0 +1,530 @@
...
+    string_formats => { Integer => '%#x' }
+  }
+}
+$str = String([1,2,3], $formats) # produces '(0x1, 0x2, 0x3)'
+```
+
+The given formats are merged with the default formats, and matching of values to convert against format is based on
+the specificity of the mapped type; for example, different formats can be used for short and long arrays.
+
+### Integer to String
```
jtappa: In lieu of making a lot of small line comments, from 258-386, the format descriptions need some consistency. Each description should start with capital letter, and end with a period. This includes the shorter descriptions, so `same as d` should be `Same as d.`

lib/puppet/parser/functions/step.rb
```diff
@@ -0,0 +1,84 @@
...
+$data = Integer[0,20]
+$transformed_data = $data.step(5).map |$item| { $item * 10 }
+$transformed_data contains [0,50,100,150,200]
+```
+
+**Example:** The same example using `step` function chained with a `map` in alternative syntax
+
+```puppet
+# For the array $data, return an array, set to the first and each 5th
+# successor,in reverse order, multiplied by 10
```
jtappa: successor,in => successor, in

lib/puppet/parser/functions/then.rb
```diff
@@ -0,0 +1,73 @@
...
+
+```puppet
+$data = {a => { b => [{x => 10, y => 20}, {ex => 100, why => 200}]}}
+notice $data.dig(a, b, 1, x).then |$x| { $x * 2 }
+```
+
+Which would notice `undef` since the last lookup of 'x' results in `undef` which
+is returned (without calling the lambda given to the `then` function).
+
+As a result there was no need for conditional logic or a temporary (non local)
```
jtappa: was => is

hlindberg: This is changed in the commit below afaict.

jtappa:
:+1:

PR #5019, opened 2016-06-15
(PUP-6370) Honor systemd exitcodes as truth

ody:
This commit removes the tracking of specific systemd states and
  replaces it with the exit code of the "is-enabled" subcommand.

  This needs to be done so that the systemd provider always follows the
  operating system's view of what enabled or disabled is, according to
  its documenation[1] and shipped version of the init system.

  [1] https://www.freedesktop.org/software/systemd/man/systemctl.html#is-enabled%20NAME...

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -72,6 +72,7 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
 
   def enabled?
     output = cached_enabled?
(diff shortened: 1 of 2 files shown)


whopper:
👍 I just gave this a run through to make sure we wouldn't see any regressions since we've had so many problems with this provider recently - looks good. All puppet service tests also pass with this change in Debian 8, so this is ready to go! Thanks @ody!

PR #5022, opened 2016-06-15
(PUP-6410) puppet/util/plist should handle malformed plist data

ccaviness:
CFPropertyList::List.new() can raise "NoMethodError: undefined method `elements' for nil:NilClass" if the data to parse isn't a valid (e.g., zero-length or not a plist).
* Catch NoMethodError in parse_plist
* explicitly return convert_cfpropertylist_to_native_types()
* add a 'return nil' at the end of read_plist_file so callers won't end up with the return value of Puppet.warning(...) if things fall through.

### DIFF
--- a/lib/puppet/util/plist.rb
+++ b/lib/puppet/util/plist.rb
@@ -33,7 +33,7 @@ module Puppet::Util::Plist
       # NOTE: We used IO.read originally to be Ruby 1.8.x compatible.
       if read_file_with_offset(file_path, binary_plist_magic_number.length) == binary_plist_magic_number
         plist_obj = new_cfpropertylist(:file => file_path)
-        convert_cfpropertylist_to_native_types(plist_obj)
+        return convert_cfpropertylist_to_native_types(plist_obj)
       else
         plist_data = open_file_with_args(file_path, "r:UTF-8")
         plist = parse_plist(plist_data, file_path)
@@ -48,6 +48,7 @@ module Puppet::Util::Plist
           Puppet.warning("Cannot read file #{file_path}; Puppet is skipping it.\n" + "Details: #{detail}")
         end
       end
+      return nil
     end
 
     # Read plist text using the CFPropertyList gem.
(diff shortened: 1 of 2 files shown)


lib/puppet/util/plist.rb
```diff
@@ -60,7 +61,7 @@ def parse_plist(plist_data, file_path = '')
 
       begin
         plist_obj = new_cfpropertylist(:data => plist_data)
-      rescue CFFormatError => e
+      rescue CFFormatError, NoMethodError => e
```
branan: Could you add a comment to the effect of "The CFPropertyList library will raise NoMethodError for invalid data"? It's not entirely clear what's happening here, since we're just wrapping an external library at this point

spec/unit/util/plist_spec.rb
```diff
@@ -54,6 +54,12 @@ module Puppet::Util::Plist; end
     </dict>
     </plist>'
   end
+  let(:non_plist_data) do
+    "Take my love, take my land
+     Take me where I cannot stand
+     I don't care, I'm still free
+     You can't take the sky from me."
+   end
```
branan: This is definitely the best fixture data I've ever seen

lib/puppet/util/plist.rb
branan: ACTUALLY now that I think about it more, does it make more sense for us to rescue `NoMethodError` in a separate block so we can format something to the extent of "This is probably invalid plist data" in the log message?

ccaviness: To be honest, this is probably(?) a bug in `cfpropertylist` and it should catch the `NoMethodError` and throw `CFFormatError`. (I haven't dug into the code, though.) So in that sense both `CFFormatError` and `NoMethodError` are catching the same thing.

I'm happy to do it either way, though.

ccaviness: I've added a comment here for now; let me know if you'd prefer I break it out.

branan: Given your explanation I think just a comment is fine. Thanks!

branan:
:+1: Thanks @ccaviness!

PR #5024, opened 2016-06-15
(PUP-6413) Use https when talking to pypi

nibalizer:
We really shouldn't use this code any more. But we are using it.
Pypi changed behaviour today because of vulnerabilities in urrlib, which
is used by pip.

http://blog.blindspotsecurity.com/2016/06/advisory-http-header-injection-in.html

Pypi changed to denying http requests entirely instead of redirecting to
https. (I guess? I don't really understand)

Without this patch:

root@derpderp:~# cat foo.pp
    package { 'diskimage-builder':
      ensure   => latest,
      provider => pip,
   }

root@derpderp:~# puppet apply foo.pp
Warning: Setting templatedir is deprecated. See http://links.puppetlabs.com/env-settings-deprecations
   (at /usr/lib/ruby/vendor_ruby/puppet/settings.rb:1139:in `issue_deprecation_warning')
Notice: Compiled catalog for derpderp in environment production in 0.14 seconds
Error: Could not get latest version: HTTP-Error: 403 Must access using HTTPS instead of HTTP
Error: /Stage[main]/Main/Package[diskimage-builder]/ensure: change from 1.17.0 to latest failed: Could not get latest version: HTTP-Error: 403 Must access using HTTPS instead of HTTP
Notice: Finished catalog run in 0.67 seconds

So this breaks any puppet 3.x using the default pip provider.

With this patch, it just works.

### DIFF
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -66,7 +66,7 @@ Puppet::Type.type(:package).provide :pip,
       proxy = http_proxy_host
     end
 
-    client = XMLRPC::Client.new2("http://pypi.python.org/pypi", proxy)
+    client = XMLRPC::Client.new2("https://pypi.python.org/pypi", proxy)
     client.http_header_extra = {"Content-Type" => "text/xml"}
     client.timeout = 10
     result = client.call("package_releases", @resource[:name])
--- a/spec/unit/provider/package/pip_spec.rb
+++ b/spec/unit/provider/package/pip_spec.rb
@@ -120,7 +120,7 @@ describe provider_class do
     context "connecting directly" do
 
       before :each do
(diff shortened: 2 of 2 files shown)


whopper:
@nibalizer woot, thanks! Seems simple enough 👍, I think we should be able to review this soon

nibalizer:
As a workaround, we can use the 'openstack_pip' provider here http://git.openstack.org/cgit/openstack-infra/puppet-pip/tree/

whopper:
@nibalizer ah, there are a few instances of `http://pypi` in `spec/unit/provider/package/pip_spec.rb`, lines 123 and 148. If you make those `https` the tests should pass.

joshuaspence:
+1 on this pull request.

Poil:
+1 We need this ASAP on 3.x and 4.x

nibalizer:
@whopper updated.

joshuaspence:
Can we expect a new 3.x release soon?

PR #5026, opened 2016-06-16
(PUP-6415) static_compiler: fix symlinks in recursed directories

bootc:
As discussed in [PUP-6415](https://tickets.puppetlabs.com/browse/PUP-6415). My grasp of Ruby is unfortunately not good enough to be able to write a test that covers this - I have tried!


branan:
Merged, thank you! Based on Josh's comments in the ticket, if you think this also ought to land in 3.x feel free to cherry-pick the commit over to that branch and open a PR - we'll get it merged as well

PR #5027, opened 2016-06-17
(doc) Update link to pypa in comments

carthik:
The comment in the code pointed to a URL for pypa docs that was out of
date. This commit fixes the URLs.


whopper:
👍

PR #5029, opened 2016-06-17
(PUP-2316) Allow single element attribute array for AIX user.

Magisus:
This is fixed in Puppet 4 due to a rewrite of the parser.

This commit updates the handling of the attributes input to make sure
they are converted to an array. The parser previously converts single
element arrays into strings, causing this bug.


Gemfile
```diff
@@ -13,6 +13,7 @@ end
 # C Ruby (MRI) or Rubinius, but NOT Windows
 platforms :ruby do
   gem 'pry', :group => :development
+  gem 'pry-nav', :group => :development
```
whopper: @Magisus ah we forgot to take this out

PR #5032, opened 2016-06-18
(PUP-6417) Do not raise error when line parsing failed

liaoishere:
Raise a error when parsing will cause the prefetch method fail of some
providers, which will lead to flush the file content that is not manged
by Puppet.

When using  resource types like ssh_authorized_keys or host, Puppet will use Puppet::Util::FileParsing module for file line parsing. Specifically, Puppet will call parse for file parsing method during prefetch stage, the parse method is like:

``` ruby
  # Split a bunch of text into lines and then parse them individually.
  def parse(text)
    count = 1
    lines(text).collect do |line|
      count += 1
      if val = parse_line(line)
        val
      else
        error = Puppet::ResourceError.new("Could not parse line #{line.inspect}")
        error.line = count
        raise error
      end
    end
  end
```

As the code above shows, Puppet will raise a Puppet::ResourceError if some line in the file can't be parsed correctly(May cause by wrong format). This will lead to failure of prefetch method of a resource type. However, Puppet will still write content to the file cause the flush method in the provider of these resource types. For example, the flush method in ssh_authorized_keys provider will flush the ~/.ssh/authorized_keys file using content in @property_hash. But the flush will delete the original content in that file cause @property_hash is not set by prefetch method. See flush method below:

``` ruby
  def flush
    # Make sure we've got a target and name set.

    # If the target isn't set, then this is our first modification, so
    # mark it for flushing.
    unless @property_hash[:target]
      @property_hash[:target] = @resource.should(:target) || self.class.default_target
      self.class.modified(@property_hash[:target])
    end
    @resource.class.key_attributes.each do |attr|
      @property_hash[attr] ||= @resource[attr]
    end

    self.class.flush(@property_hash)
  end
```

As above shows ,@property_hash is set by using resource attribute when it is empty. This will lead to lose original content in the file after flush the file.

How to reproduce this bug:

Say the ssh authorized_keys file is like:

``` bash
$ cat ~/.ssh/authorized_keys
ssh-rsa ORIGINAL_KEY COMMENT
A LINE THAT CAN'T PARSE BY PUPPET
```

Then apply a puppet manifest like:

``` puppet
ssh_authorized_key { 'test':
  user => 'root',
  type => 'ssh-rsa',
  key  => 'A_NEW_KEY',
}
```

the authorized_keys becomes like:

``` bash
$ cat ~/.ssh/authorized_keys
ssh-rsa A_NEW_KEY test
```

The original content in that file lost.

### DIFF
--- a/lib/puppet/util/fileparsing.rb
+++ b/lib/puppet/util/fileparsing.rb
@@ -222,16 +222,17 @@ module Puppet::Util::FileParsing
   # Split a bunch of text into lines and then parse them individually.
   def parse(text)
     count = 1
-    lines(text).collect do |line|
+    res = lines(text).collect do |line|
       count += 1
       if val = parse_line(line)
         val
       else
-        error = Puppet::ResourceError.new("Could not parse line #{line.inspect}")
-        error.line = count
-        raise error
+        Puppet.warning("Could not parse line #{line.inspect}")
+        nil
       end
     end
+    # Delete lines of wrong format
+    res.compact
   end
 
   # Handle parsing a single line.


lib/puppet/util/fileparsing.rb
```diff
@@ -222,16 +222,17 @@ def lines(text)
...
       else
-        error = Puppet::ResourceError.new("Could not parse line #{line.inspect}")
-        error.line = count
-        raise error
+        Puppet.warning("Could not parse line #{line.inspect}")
+        nil
       end
     end
+    # Delete lines of wrong format
+    res.reject! { |i| i.nil? }
```
MikaelSmith: This could be `res.compact`.

branan:
I wonder if this isn't better handled in `transaction.rb` somewhere - in my mind the bug is that we keep trying to apply resources even after prefetching them failed. While making prefetch for parsed files more robust is a good thing, it doesn't solve the root misbehavior of puppet's application process.

branan:
It also looks like this change had pretty broad fallout in the tests for providers which use the file parsing code (especially mount).

branan:
So I guess two questions for you:

1) How do you feel about fixing this issue here vs in a more centralized location
2) Do you need help debugging and working through the test failures?

liaoishere:
@branan thanks for your advice. I'm not familiar with much code in Puppet. So I'm not sure if there could be some better solutions. And I have a workaround solution for handling unrecognized lines during file parsing.

For example, add a regex that matching anything in the last in the  provider. So that any line don't match the regular regex will match the last one, which will not cause the prefetch stage fail. And calling the flush will not delete any original lines in the file.


Say, in ssh_authorized_keys provider, using the text_line method to create a new FileRecord object that match anything else:

```ruby
Puppet::Type.type(:ssh_authorized_key).provide(
  :parsed,
  :parent         => Puppet::Provider::ParsedFile,
  :filetype       => :flat,
  :default_target => ''
) do
  desc "Parse and generate authorized_keys files for SSH."

  text_line :comment, :match => /^\s*#/
  text_line :blank, :match => /^\s*$/

  record_line :parsed,
    :fields   => %w{options type key name},
    :optional => %w{options},
    :rts => /^\s+/,
    :match    => Puppet::Type.type(:ssh_authorized_key).keyline_regex,
    :post_parse => proc { |h|
      h[:name] = "" if h[:name] == :absent
      h[:options] ||= [:absent]
      h[:options] = Puppet::Type::Ssh_authorized_key::ProviderParsed.parse_options(h[:options]) if h[:options].is_a? String
    },
    :pre_gen => proc { |h|
      # if this name was generated, don't write it back to disk
      h[:name] = "" if h[:unnamed]
      h[:options] = [] if h[:options].include?(:absent)
      h[:options] = h[:options].join(',')
    }

  record_line :key_v1,
    :fields   => %w{options bits exponent modulus name},
    :optional => %w{options},
    :rts      => /^\s+/,
    :match    => /^(?:(.+) )?(\d+) (\d+) (\d+)(?: (.+))?$/

  text_line :else, :match => /^.*$/                       # This will match any line in the last
```

The record type is ordered in new_line_type method, and put into variable @record_order, when parsing lines, the parse_line method will iterate @record_order variable. So any lines unregular will match the last record type which will match everything:

```ruby
  # Handle parsing a single line.
  def parse_line(line)
    raise Puppet::DevError, "No record types defined; cannot parse lines" unless records?

    @record_order.each do |record|
      # These are basically either text or record lines.
      method = "handle_#{record.type}_line"
      if respond_to?(method)
        if result = send(method, line, record)
          record.send(:post_parse, result) if record.respond_to?(:post_parse)
          return result
        end
      else
        raise Puppet::DevError,
          "Somehow got invalid line type #{record.type}"
      end
    end

    nil
  end
```

Same as ssh_authorized_keys, we can add this special regex in hosts provider or others that use fileparsing.

How do you think this kind solution?

BTW, I'm quite confusing with the test failures in  this patch. Thanks for your help, could you help me select a solution for this bug and debugging with the test failures?

PS: My english is poor. Thanks for understanding my words.

kylog:
@liaoishere: first, sorry for the long delay in getting back to you on this PR, and its companion #5031. As a heads-up, I just merged #5314 (which extended your contribution in #5031 with a spec test). I think that addresses PUP-6417, and that fix should be released with puppet 4.8.0.

Meanwhile, it might be best to file a separate issue at tickets.puppetlabs.com for what this PR is trying to fix. It's really not specific to PUP-6417 (even if it was exposed by that issue), and as @branan suggests, it may be exposing a more fundamental issue in the puppet transaction logic. Are you interested in submitting that issue?

Many thanks for your work both in #5031 and here!

MosesMendoza:
This PR hasn't been updated in some time, and unfortunately it seems like we're not going to be able to pick it up any time soon. I'm going to close it for now, but we can re-open if/when a new ticket is filed related to PUP-6417 and the team has sufficient time to incorporate it into the codebase.

PR #5033, opened 2016-06-20
(PUP-5604) Journald on failure in Systemd

petems:
The ideas from this are working as an extracted T/P: https://github.com/petems/petems-systemd_logs

Error before:
> Error: Could not start Service[foo]: Execution of '/usr/bin/systemctl start foo' returned 6: Failed to start foo.service: Unit foo.service failed to load: Invalid argument. See system logs and 'systemctl status foo.service' for details.

Error after:
> Error: Systemd start for foo failed:journalctl log for foo: -- Logs begin at Mon 2016-06-20 12:07:17 UTC, end at Mon 2016-06-20 12:08:06 UTC. --
Jun 20 12:08:06 centos-7-x64 systemd[1]: foo.service lacks both ExecStart= and ExecStop= setting. Refusing.
Jun 20 12:08:06 centos-7-x64 systemd[1]: foo.service lacks both ExecStart= and ExecStop= setting. Refusing.

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -145,5 +145,37 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
   def statuscmd
     [command(:systemctl), "is-active", @resource[:name]]
   end
+
+  def restart
+    begin
+      super
+    rescue Puppet::Error => e
+      raise Puppet::Error.new(prepare_error_message(@resource[:name], 'restart', e))
+    end
+  end
+
+  def start
+    begin
+      super
+    rescue Puppet::Error => e
+      raise Puppet::Error.new(prepare_error_message(@resource[:name], 'start', e))
+    end
+  end
+
+  def stop
+    begin
+      super
+    rescue Puppet::Error => e
+      raise Puppet::Error.new(prepare_error_message(@resource[:name], 'stop', e))
+    end
+  end
+
+  def prepare_error_message(name, action, exception)
+    error_return = "Systemd #{action} for #{name} failed!\n"
+    journalctl_command = "journalctl -n 50 --since '5 minutes ago' -u #{name} --no-pager"
+    Puppet.debug("Runing journalctl command to get logs for systemd #{action} failure: #{journalctl_command}")
+    journalctl_output = execute(journalctl_command)
+    error_return << "journalctl log for #{name}:\n#{journalctl_output}"
+  end
 end
 
--- a/spec/unit/provider/service/systemd_spec.rb
+++ b/spec/unit/provider/service/systemd_spec.rb
@@ -140,8 +140,23 @@ describe Puppet::Type.type(:service).provider(:systemd) do
     it "should start the service with systemctl start otherwise" do
       provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service'))
       provider.expects(:systemctl).with(:unmask, 'sshd.service')
-      provider.expects(:execute).with(['/bin/systemctl','start','sshd.service'], :failonfail => true, :override_locale => false, :squelch => false, :combine => true)
+      provider.expects(:execute).with(['/bin/systemctl','start','sshd.service'], {:failonfail => true, :override_locale => false, :squelch => false, :combine => true})
       provider.start
+    end
+
+    it "should show journald logs on failure" do
+      provider = described_class.new(Puppet::Type.type(:service).new(:name => 'sshd.service'))
+      provider.expects(:systemctl).with(:unmask, 'sshd.service')
+      provider.expects(:execute).with(['/bin/systemctl','start','sshd.service'],{:failonfail => true, :override_locale => false, :squelch => false, :combine => true})
+        .raises(Puppet::ExecutionFailure, "Failed to start sshd.service: Unit sshd.service failed to load: Invalid argument. See system logs and 'systemctl status sshd.service' for details.")
+      journalctl_logs = <<-EOS
+-- Logs begin at Tue 2016-06-14 11:59:21 UTC, end at Tue 2016-06-14 21:45:02 UTC. --
+Jun 14 21:41:34 foo.example.com systemd[1]: Stopping sshd Service...
+Jun 14 21:41:35 foo.example.com systemd[1]: Starting sshd Service...
+Jun 14 21:43:23 foo.example.com systemd[1]: sshd.service lacks both ExecStart= and ExecStop= setting. Refusing.
+      EOS
+      provider.expects(:execute).with("journalctl -n 50 --since '5 minutes ago' -u sshd.service --no-pager").returns(journalctl_logs)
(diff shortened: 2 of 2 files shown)


branan:
Besides the tests exploding this looks like a decent change to me.

Is it worth adding this to any other systemctl commands? (stop is the big one missing from your patch)

lib/puppet/provider/service/systemd.rb
```diff
@@ -145,5 +157,12 @@ def stopcmd
   def statuscmd
     [command(:systemctl), "is-active", @resource[:name]]
   end
+
+  def prepare_error_message(name, action)
+    error_return = "Systemd #{action} for #{name} failed:"
+    journalctl_output = execute("journalctl -n 50 --since '5 minutes ago' -u #{name} --no-pager")
```
supercow: What happens if this command fails?

petems: I'm assuming it would raise a normal Puppet execution error? I could add a spec in for that

petems:
@branan Yeah, good point, I guess stopping could also fail if the service file changed the command or something.

adrienthebo:
I'm not hugely worried about failures on stopping; if you fail to stop a service then something very wonky is going on. We can cover that as well but it's not worth fretting about a huge deal.

branan:
I don't have strong feelings about stop, tbh. Are there other systemctl interactions we're missing, though? Do enable/disable point you at journalctl if they fail, or do they print useful output? Any other commands we're missing?

petems:
I might be wrong but IIRC enable/disable just simlinks the .service files into a boot directory, so I don't think there's much failure logic there.

lib/puppet/provider/service/systemd.rb
```diff
@@ -129,13 +129,25 @@ def unmask
...
 
-  def restartcmd
-    [command(:systemctl), "restart", @resource[:name]]
+  def restart
+    cmds = []
+    cmds << :systemctl << 'restart' << @resource[:name]
+    begin
+      execute(cmds)
+    rescue Puppet::ExecutionFailure
+      raise Puppet::Error.new(prepare_error_message(@resource[:name], 'restart'), $!)
```
adrienthebo: Prefer `$ERROR_STATUS` over `$!` with the [English](http://ruby-doc.org/stdlib-2.0.0/libdoc/English/rdoc/English.html) module; I've forgotten as much Perl as I can and can't recall what the magic variables mean anymore :)

lib/puppet/provider/service/systemd.rb
adrienthebo: If this command fails you most likely have much bigger problems, and there's a good chance that Puppet will crash and burn entirely shortly. But if this does happen and throws an exception, you'll most likely end up here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/transaction/resource_harness.rb#L26

lib/puppet/provider/service/systemd.rb
adrienthebo: Alternately, instead of using a magic variable, do this:

```ruby
rescue Puppet::ExecutionFailure => e
  raise Puppet::Error.new(prepare_error_message(@resource[:name], 'restart'), e)
```

lib/puppet/provider/service/systemd.rb
```diff
@@ -145,5 +157,12 @@ def stopcmd
   def statuscmd
     [command(:systemctl), "is-active", @resource[:name]]
   end
+
+  def prepare_error_message(name, action)
+    error_return = "Systemd #{action} for #{name} failed:"
+    journalctl_output = execute("journalctl -n 50 --since '5 minutes ago' -u #{name} --no-pager")
+    error_return << "journalctl log for #{name}: #{journalctl_output}"
+    raise Puppet::Error, error_return, $!.backtrace
```
adrienthebo: This use of `$!` is using an exception that was thrown in a different context, which is a bit confusing. Instead of using a global variable you can catch the exception and store it into a variable as noted above, then explicitly pass it into this function to get the backtrace.

petems: Good plan, fixed

lib/puppet/provider/service/systemd.rb
petems: Fixed! 👍

petems:
Thanks to some guidance from @adrienthebo, tests are green too 😄

adrienthebo:
LGTM

daenney:
I think this is good. The one, really nitpick-y comment I have:

> Error: Systemd start for foo failed:journalctl log for foo:

I'd like a space between `failed:` and `journalctl`.

lib/puppet/provider/service/systemd.rb
```diff
@@ -129,21 +129,52 @@ def unmask
...
-  end
-
   def startcmd
     self.unmask
     [command(:systemctl), "start", @resource[:name]]
   end
 
+  def restartcmd
+    [command(:systemctl), "restart", @resource[:name]]
+  end
```
igalic: this seems unnecessary.

lib/puppet/provider/service/systemd.rb
```diff
@@ -129,21 +129,52 @@ def unmask
...
     [command(:systemctl), "is-active", @resource[:name]]
   end
+
+  def restart
+    begin
+      super
+    rescue Puppet::Error => e
+      raise Puppet::Error.new(prepare_error_message(@resource[:name], 'restart', e))
+    end
+  end
```
igalic: could we just alias the other methods?

lib/puppet/provider/service/systemd.rb
petems: Hah, yeah that was just some whitespace weirdness, will revert

lib/puppet/provider/service/systemd.rb
```diff
@@ -145,5 +145,36 @@ def stopcmd
...
+    rescue Puppet::Error => e
+      raise Puppet::Error.new(prepare_error_message(@resource[:name], 'stop', e))
+    end
+  end
+
+  def prepare_error_message(name, action, exception)
+    error_return = "Systemd #{action} for #{name} failed:"
+    journalctl_output = execute("journalctl -n 50 --since '5 minutes ago' -u #{name} --no-pager")
+    error_return << "journalctl log for #{name}: #{journalctl_output}"
+    raise Puppet::Error, error_return, exception.backtrace
```
adrienthebo: The calling methods are calling this method to generate an error message for an exception, but this code itself raises an exception - which makes the `raise` calls in the start/stop/restart methods dead code.

petems: Good point! 👍 Fixed

petems:
@daenney Yep, makes sense!

New output looks like the following:

> Error: Systemd start for foo failed! 
journalctl log for foo:
-- Logs begin at Mon 2016-06-20 12:07:17 UTC, end at Mon 2016-06-20 12:08:06 UTC. --
Jun 20 12:08:06 centos-7-x64 systemd[1]: foo.service lacks both ExecStart= and ExecStop= setting. Refusing.
Jun 20 12:08:06 centos-7-x64 systemd[1]: foo.service lacks both ExecStart= and ExecStop= setting. Refusing.

daenney:
@petems 💯

petems:
@branan @kylog Any chance of a merge? 😄

petems:
Yay 💃

PR #5034, opened 2016-06-20
(PUP-6376) Add `server_list` option to settings

Magisus:
This adds a server_list option to Puppet's settings. Because having both
server_list and server be set could result in ambiguous behavior, a
warning is emitted when the user sets both. The actual logic handling
the conflict is still to be determined and implemented.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1395,7 +1395,26 @@ EOT
     },
     :server => {
       :default => "puppet",
-      :desc => "The puppet master server to which the puppet agent should connect."
+      :desc => "The puppet master server to which the puppet agent should connect.",
+      :call_hook => :on_initialize_and_write,
+      :hook => proc { |value|
+          if Puppet.settings.set_by_config?(:server_list)
+            Puppet.deprecation_warning('Attempted to set both server and server_list. Server setting will not be used.')
+          end        
+          value
+      }
+    },
+    :server_list => {
+      :default => [],
+      :type => :server_list,
+      :desc => "The list of puppet master servers to which the puppet agent should connect,
+        in the order that they will be tried.",
+      :call_hook => :on_initialize_and_write,
+      :hook => proc { |value| 
+            if Puppet.settings.set_by_config?(:server)  
+              Puppet.deprecation_warning('Attempted to set both server and server_list. Server setting will not be used.')
+            end
+        }
     },
     :use_srv_records => {
       :default    => false,
--- a/lib/puppet/settings.rb
+++ b/lib/puppet/settings.rb
@@ -29,6 +29,7 @@ class Puppet::Settings
   require 'puppet/settings/config_file'
   require 'puppet/settings/value_translator'
   require 'puppet/settings/environment_conf'
+  require 'puppet/settings/server_list_setting'
 
   # local reference for convenience
   PuppetOptionParser = Puppet::Util::CommandLine::PuppetOptionParser
@@ -667,6 +668,7 @@ class Puppet::Settings
       :symbolic_enum   => SymbolicEnumSetting,
       :priority   => PrioritySetting,
       :autosign   => AutosignSetting,
+      :server_list => ServerListSetting
   }
 
   # Create a new setting.  The value is passed in because it's used to determine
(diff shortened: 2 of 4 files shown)


lib/puppet/defaults.rb
```diff
@@ -1395,7 +1395,24 @@ def self.default_diffargs
...
+          end        
+      }
+    },
+    :server_list => {
+      :default => ["puppet"],
+      :type => :array,
+      :desc => "The list of puppet master servers to which the puppet agent should connect,
+        in the order that they will be tried.",
+      :hook => proc { |value| 
+            if Puppet.settings.set_by_config?(:server)
```
branan: does `set_by_config` trigger if a setting is set on the command line?

Also, despite that fact that only one the two hooks will detect the conflict (depending on the order of the settings in the config) we should probably treat them as being set "simultaneously" as far as error messaging goes, and use a common "We detected both settings but we're gonna use `server_list` because it's new" message

Magisus: Yes, it does detect command line setting, as well as setting in code and by config file. I believe there is a test to that effect, in addition to the tests for set_by_config itself.

branan: :+1: sweet.

Good to know our accurate naming conventions stand within this part of the codebase ;)

lib/puppet/defaults.rb
```diff
@@ -1395,7 +1395,24 @@ def self.default_diffargs
     },
     :server => {
       :default => "puppet",
-      :desc => "The puppet master server to which the puppet agent should connect."
+      :desc => "The puppet master server to which the puppet agent should connect.",
+      :hook => proc { |value|
+          if Puppet.settings.set_by_config?(:server_list)
```
joshcooper: deprecating `server` is going to effect many code paths... Can we either not deprecate it and accept both (with one taking priority) or just change `server` to be multivalued, so that existing code doesn't need to change and will continue to do the right thing?

Magisus: The current plan sounds like to not deprecate and have server_list take priority. But we will be discussing this more this afternoon.

kylog: What was the outcome of the deprecation discussion? Fwiw, my least favorite option is that we live with two different settings forever and ever. Allowing `server` to be multivalued would be one approach, but failing that, I'd favor deprecating `server` over having both settings forever.  If concern over deprecation is high, we could hold off on deprecating until we know we're on the last 4.y (i.e. once we decide when to do a puppet 5).

Thoughts?

joshcooper: `--server` is used in many places outside of the puppet repo. There are a bunch of github results:

[bash](https://github.com/search?l=bash&q=%22puppet+agent+--server%22&type=Code&utf8=%E2%9C%93)
[ruby](https://github.com/search?l=ruby&q=%22puppet+agent+--server%22&type=Code&utf8=%E2%9C%93)

One one hand, some of those results are forks of puppet. But on the other hand, it's only looking for cases where --server follows agent, so not `puppet agent -t --server`.

Also it doesn't include all of the puppet, PE and modules acceptance tests that would need to be updated to remove the deprecation. There's also MCO, though we just removed the --server option by default.

There's also customer integration/workflows to consider, e.g. running puppet from cron, etc and documentation:

http://shapeshed.com/connecting-clients-to-a-puppet-master/
https://www.unixmen.com/puppet-agent-install-and-configure-to-all-linuxunix/
http://www.juniper.net/techpubs/en_US/junos-puppet2.0/topics/task/installation/automation-junos-puppet-agent-installing.html

branan:
I think given the commonality of the server setting, both in config files and on the command line, that it behooves us to be slow in our deprecation

My thought was to only warn when both settings are set for now, add a real deprecation warning in 5.0, and remove in 6.0. This is probably a several-year deprecation cycle, but I don't think that's crazy for a setting that's been in puppet since more or less the beginning

branan:
@Magisus It looks like you lost the `call_hook` flags in one of your rebases

kylog:
Oh crud. Forgot to respond.

I like the slow deprecation cycle idea. But did we discuss/reject the idea of not adding a new setting at all and allowing `server` to be multivalued?

branan:
@kylog my understanding is that the new setting is intended to be managed by HA tooling, and we don't want to have a potential puppet resource conflict with users' manifests (we don't currently manage the `server` setting).

Magisus:
@branan :call_hook flags re-added :)

whopper:
This commit is now wrapped into Branan's PR (https://github.com/puppetlabs/puppet/pull/5055/commits). Closing this.

PR #5036, opened 2016-06-20
(PUP-1498) Windows Taskscheduler changes fail to apply

Iristyle:


Iristyle:
Still need to run acceptance test mod through CI locally

Iristyle:
Most of this ended up getting merged in accidentally out of order via #5038 

Introducing a new fixup PR to adjust the problem with the test at #5042

PR #5037, opened 2016-06-20
(PUP-5831) Add binary type

hlindberg:
This adds a Binary type to the type system along with language operations. This PR also contains a function that reads a file in binary mode and returns the content as a Binary. PR also has exploratory work for setting the content of a File to be Binary, and to make this work for a puppet apply.

What remains to make the Binary type useful is the ability to include it in a catalog (i.e. basically serialization).

### DIFF
--- a/lib/puppet/functions/binary_file.rb
+++ b/lib/puppet/functions/binary_file.rb
@@ -1,0 +1,36 @@
+# Loads a binary file from a module or file system and returns its contents as a Binary.
+#
+# The argument to this function should be a `<MODULE NAME>/<FILE>`
+# reference, which will load `<FILE>` from a module's `files`
+# directory. (For example, the reference `mysql/mysqltuner.pl` will load the
+# file `<MODULES DIRECTORY>/mysql/files/mysqltuner.pl`.)
+#
+# This function also accepts an absolute file path that allows reading
+# binary file content from anywhere on disk.
+#
+# An error is raised if the given file does not exists.
+#
+# To search for the existence of files, use the `find_file()` function.
+#
+# @since 4.6.0
+#
+Puppet::Functions.create_function(:binary_file, Puppet::Functions::InternalFunction) do
+  dispatch :binary_file do
+    scope_param
+    param 'String', :path
+  end
+
+  def binary_file(scope, unresolved_path)
+    path = nil
+    found = Puppet::Parser::Files.find_file(unresolved_path, scope.compiler.environment)
+    if found && Puppet::FileSystem.exist?(found)
+      path = found
+    end
+
+    if path
+      Puppet::Pops::Types::PBinaryType::Binary.from_binary_string(Puppet::FileSystem.binread(path))
+    else
+      raise Puppet::ParseError, "binary_file(): The given file '#{path}' does not exist"
+    end
+  end
+end
--- a/lib/puppet/functions/binary_file.rb
+++ b/lib/puppet/functions/binary_file.rb
@@ -1,18 +1,7 @@
 # Loads a binary file from a module or file system and returns its contents as a Binary.
+# (Documented in 3.x stub)
 #
-# The argument to this function should be a `<MODULE NAME>/<FILE>`
-# reference, which will load `<FILE>` from a module's `files`
-# directory. (For example, the reference `mysql/mysqltuner.pl` will load the
-# file `<MODULES DIRECTORY>/mysql/files/mysqltuner.pl`.)
-#
-# This function also accepts an absolute file path that allows reading
-# binary file content from anywhere on disk.
-#
-# An error is raised if the given file does not exists.
-#
-# To search for the existence of files, use the `find_file()` function.
-#
-# @since 4.6.0
+# @since 4.8.0
 #
 Puppet::Functions.create_function(:binary_file, Puppet::Functions::InternalFunction) do
   dispatch :binary_file do
--- a/lib/puppet/functions/find_file.rb
+++ b/lib/puppet/functions/find_file.rb
@@ -1,19 +1,7 @@
 # Finds an existing file from a module and returns its path.
+# (Documented in 3.x stub)
 #
-# The argument to this function should be a String as a `<MODULE NAME>/<FILE>`
-# reference, which will search for `<FILE>` relative to a module's `files`
-# directory. (For example, the reference `mysql/mysqltuner.pl` will search for the
-# file `<MODULES DIRECTORY>/mysql/files/mysqltuner.pl`.)
-#
-# This function can also accept:
-#
-# * An absolute String path, which will check for the existence of a file from anywhere on disk.
-# * Multiple String arguments, which will return the path of the **first** file
-#   found, skipping non existing files.
-# * An array of string paths, which will return the path of the **first** file
-#   found from the given paths in the array, skipping non existing files.
-#
-# @since 4.6.0
+# @since 4.8.0
 #
 Puppet::Functions.create_function(:find_file, Puppet::Functions::InternalFunction) do
   dispatch :find_file do
--- a/lib/puppet/functions/find_file.rb
+++ b/lib/puppet/functions/find_file.rb
@@ -1,0 +1,43 @@
+# Finds an existing file from a module and returns its path.
+#
+# The argument to this function should be a String as a `<MODULE NAME>/<FILE>`
+# reference, which will search for `<FILE>` relative to a module's `files`
+# directory. (For example, the reference `mysql/mysqltuner.pl` will search for the
+# file `<MODULES DIRECTORY>/mysql/files/mysqltuner.pl`.)
+#
+# This function can also accept:
+#
+# * An absolute String path, which will check for the existence of a file from anywhere on disk.
+# * Multiple String arguments, which will return the path of the **first** file
+#   found, skipping non existing files.
+# * An array of string paths, which will return the path of the **first** file
+#   found from the given paths in the array, skipping non existing files.
+#
+# @since 4.6.0
+#
+Puppet::Functions.create_function(:find_file, Puppet::Functions::InternalFunction) do
+  dispatch :find_file do
+    scope_param
+    repeated_param 'String', :paths
+  end
+
+  dispatch :find_file_array do
+    scope_param
+    repeated_param 'Array[String]', :paths_array
+  end
+
+  def find_file_array(scope, array)
+    find_file(scope, *array)
+  end
+
+  def find_file(scope, *args)
+    path = nil
+    args.each do |file|
+      found = Puppet::Parser::Files.find_file(file, scope.compiler.environment)
+      if found && Puppet::FileSystem.exist?(found)
+        return found
+      end
+    end
+    nil
+  end
+end
--- a/lib/puppet/functions/new.rb
+++ b/lib/puppet/functions/new.rb
@@ -26,6 +26,6 @@ Puppet::Functions.create_function(:new, Puppet::Functions::InternalFunction) do
   end
 
   def assert_type(type, value)
-    Puppet::Pops::Types::TypeAsserter.assert_instance_of("Converted value from #{type.to_s}.new()", type, value)
+    Puppet::Pops::Types::TypeAsserter.assert_instance_of(['Converted value from %s.new()', type], type, value)
   end
 end
--- a/lib/puppet/functions/new.rb
+++ b/lib/puppet/functions/new.rb
@@ -26,6 +26,6 @@ Puppet::Functions.create_function(:new, Puppet::Functions::InternalFunction) do
   end
 
   def assert_type(type, value)
-    Puppet::Pops::Types::TypeAsserter.assert_instance_of('new():', type, value)
+    Puppet::Pops::Types::TypeAsserter.assert_instance_of("Converted value from #{type.to_s}.new()", type, value)
   end
 end
--- a/lib/puppet/parser/functions/new.rb
+++ b/lib/puppet/parser/functions/new.rb
@@ -709,6 +709,8 @@ When given a single value as argument:
 * An empty `Hash` becomes an empty array.
 * An `Array` is simply returned.
 * An `Iterable[T]` is turned into an array of `T` instances.
+* A `Binary` is converted to an `Array[Integer[0,255]]` of byte values
+
 
 When given a second Boolean argument:
 
--- a/lib/puppet/parser/functions/new.rb
+++ b/lib/puppet/parser/functions/new.rb
@@ -630,7 +630,7 @@ Defaults to `s` at top level and `p` inside array or hash.
 
 | Format    | Default formats
 | ------    | ---------------
-| s         | binary as unquoted UTF-8 characters (errors if byte sequence is invalid UTF-8)
+| s         | binary as unquoted UTF-8 characters (errors if byte sequence is invalid UTF-8). Alternate form escapes non ascii bytes.
 | p         | 'Binary("<base64strict>")'
 | b         | '<base64>' - base64 string with newlines inserted
 | B         | '<base64strict>' - base64 strict string (without newlines inserted)
@@ -638,8 +638,10 @@ Defaults to `s` at top level and `p` inside array or hash.
 | t         | 'Binary' - outputs the name of the type only
 | T         | 'BINARY' - output the name of the type in all caps only
 
-The alternate form flag `#` will quote the binary or base64 text output
-The width and precision values are applied to the text part only in `%p` format.
+* The alternate form flag `#` will quote the binary or base64 text output.
+* The format `%#s` allows invalid UTF-8 characters and outputs all non ascii bytes
+  as hex escaped characters on the form `\\xHH` where `H` is a hex digit.
+* The width and precision values are applied to the text part only in `%p` format.
 
 ### Array & Tuple to String
 
@@ -864,7 +866,7 @@ The formats have the following meaning:
 | b | The data is in base64 encoding, padding as required by base64 strict is added by default
 | u | The data is in URL safe base64 encoding
 | B | The data is in base64 strict encoding
-| s | The data is a puppet string used verbatim as a binary byte sequence
+| s | The data is a puppet string seen as the verbatim byte sequence representing the string in its given encoding.
 
 The default format is `%b`.
 
--- a/lib/puppet/parser/functions/new.rb
+++ b/lib/puppet/parser/functions/new.rb
@@ -828,6 +828,7 @@ function SemVerRange.new(
 For examples of `SemVerRange` use see "Creating a SemVer"
 
 * Since 4.5.0
+* Binary type since 4.8.0
 
 DOC
 ) do |args|
--- a/lib/puppet/parser/functions/new.rb
+++ b/lib/puppet/parser/functions/new.rb
@@ -626,6 +626,21 @@ Defaults to `s` at top level and `p` inside array or hash.
 | s         | Same as d.
 | p         | Same as d.
 
+### Binary value to String
+
+| Format    | Default formats
+| ------    | ---------------
+| s         | binary as unquoted characters
+| p         | 'Binary("<base64strict>")'
+| b         | '<base64>' - base64 string with newlines inserted
+| B         | '<base64strict>' - base64 strict string (without newlines inserted)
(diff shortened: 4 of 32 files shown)


lib/puppet/functions/binary_file.rb
```diff
@@ -0,0 +1,25 @@
...
+Puppet::Functions.create_function(:binary_file, Puppet::Functions::InternalFunction) do
+  dispatch :binary_file do
+    scope_param
+    param 'String', :path
+  end
+
+  def binary_file(scope, unresolved_path)
+    path = nil
+    found = Puppet::Parser::Files.find_file(unresolved_path, scope.compiler.environment)
+    if found && Puppet::FileSystem.exist?(found)
```
thallgren: Unless I read the code incorrectly, a file returned from `Puppet::Parser::Files#find_file` is guaranteed to exist (see module.rb:137), so this extra `#exist?` is unnecessary and might be expensive.

lib/puppet/functions/find_file.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def find_file_array(scope, array)
+    find_file(scope, *array)
+  end
+
+  def find_file(scope, *args)
+    path = nil
+    args.each do |file|
+      found = Puppet::Parser::Files.find_file(file, scope.compiler.environment)
+      if found && Puppet::FileSystem.exist?(found)
```
thallgren: Same here, the extra `#exists?()` is unnecessary.

lib/puppet/functions/binary_file.rb
```diff
@@ -0,0 +1,25 @@
...
+      path = found
+    end
+
+    if path
+      Puppet::Pops::Types::PBinaryType::Binary.from_binary_string(Puppet::FileSystem.binread(path))
```
thallgren: If we're keen on keeping the number of calls to `#stat` to a minimum, then we should have a `#find_file` variant that doesn't check for existence at all and then rely on the `Errno::ENOENT` raised here when the file is missing. At present, this call will result in a third `#stat` (or similar).

lib/puppet/functions/new.rb
```diff
@@ -26,6 +26,6 @@ def new_function_for_type(t, scope)
   end
 
   def assert_type(type, value)
-    Puppet::Pops::Types::TypeAsserter.assert_instance_of('new():', type, value)
+    Puppet::Pops::Types::TypeAsserter.assert_instance_of("Converted value from #{type.to_s}.new()", type, value)
```
thallgren: Better to use:
```
Puppet::Pops::Types::TypeAsserter.assert_instance_of(['Converted value from %s.new()', type], type, value)
```
to avoid unnecessary formatting when value is an instance of type.

lib/puppet/pops/evaluator/access_operator.rb
```diff
@@ -30,6 +30,10 @@ def access_Object(o, scope, keys)
     fail(Issues::OPERATOR_NOT_APPLICABLE, @semantic.left_expr, :operator=>'[]', :left_value => o)
   end
 
+  def access_Binary(o, scope, keys)
+    Puppet::Pops::Types::PBinaryType::Binary.from_binary_string(access_String(o.binary_buffer, scope, keys))
```
thallgren: How can I pick an individual byte from a Binary and read it as an Integer?

lib/puppet/pops/types/p_binary_type.rb
```diff
@@ -0,0 +1,194 @@
...
+      Base64.encode64(@binary_buffer)
+    end
+
+    # Returns the binary content as a url safe base64 string (where + and / are replaced by - and _)
+    #
+    def urlsafe_to_s
+      Base64.urlsafe_encode64(@binary_buffer)
+    end
+
+    def eql?(o)
```
thallgren: Should probably have a #hash method as well.

lib/puppet/pops/types/p_binary_type.rb
```diff
@@ -0,0 +1,194 @@
...
+    o.is_a?(Binary)
+  end
+
+  def eql?(o)
+    self.class == o.class
+  end
+
+  def hash?
+    super
+  end
```
thallgren: This method is redundant.

lib/puppet/pops/types/p_binary_type.rb
```diff
@@ -0,0 +1,194 @@
...
+      @binary_buffer.length
+    end
+  end
+
+  def self.register_ptype(loader, ir)
+    create_ptype(loader, ir, 'AnyType')
+  end
+
+  def initialize()
+  end
```
thallgren: This method is redundant.

lib/puppet/pops/types/string_converter.rb
```diff
@@ -385,6 +386,21 @@ def initialize
...
   # | p         | same as d
   #
+  # ### Binary (value)
+  #
+  # | Format    | Default formats
+  # | ------    | ---------------
+  # | s         | binary as unquoted characters
+  # | p         | 'Binary("<base64strict>")'
+  # | b         | '<base64>' - base64 string with newlines inserted
+  # | B         | '<base64strict>' - base64 strict string (without newlines inserted)
```
thallgren: It seems a bit inconsistent that '%s' maps to unquoted with newlines while '%B' maps to the `Binary#to_s()` method. Intuitively, I'd assume that '%s' would be the same as `#to_s()`.

lib/puppet/pops/types/p_binary_type.rb
```diff
@@ -0,0 +1,194 @@
...
+
+  DEFAULT = PBinaryType.new
+
+  protected
+
+  def _assignable?(o, guard)
+    o.class == self.class
+  end
+
+end
```
thallgren: The ticket mentions encoding and how that affects assignability. Was that abandoned or deferred?

lib/puppet/functions/binary_file.rb
hlindberg: While I don't think this matters performance wise, there is no need to check twice. Will remove the extra existence check.

lib/puppet/functions/binary_file.rb
hlindberg: Sure. Given though that, there is an issue currently with ENOENT vs. permission failure on JRuby and that the expected frequency of calls to `binary_file` is both low, and minor in relation to actually reading the file, the optimization of this would be best performed later - and then looking at file reading and existence checks in general.

lib/puppet/pops/evaluator/access_operator.rb
hlindberg: Create an Array from the binary. Slice it before or after the conversion.

lib/puppet/pops/types/string_converter.rb
hlindberg: We could use a different letter if that helps. The intent is that when you use %s you want the binary as text. Normally binary data would not be text and you would spew out binary gobbledygook. So the distinction needs to exist. The `to_s` variant is typically used blindly in all sorts of "lets print it out" so, it can not be the "binary as text" option.

lib/puppet/pops/types/p_binary_type.rb
hlindberg: Abandoned. Bad idea. A Binary is simply a binary buffer and any encoding is about how it is presented in some medium.

lib/puppet/functions/binary_file.rb
thallgren: What makes you think that it doesn't matter? I think that an `#exist?` call is likely to do a `stat()`, and even if the result is cached by the OS, it's still a switch between user space and system space.

hlindberg:
I tried with removing what appears to be redundant existence checks, but that did not end well. It then errors with bad error message. I reverted that (local) change favouring getting a better error message.

hlindberg:
@thallgren Do you think it is worth changing the %s format to something else? I have not made that change.

lib/puppet/functions/binary_file.rb
thallgren: I see now that if an absolute path is used, the `#exists?` call is still needed since `#find_file` simply assumes that the absolute path exists. Guess that's another thing we could take a look at if/when we do a bigger overhaul to lessen the number of file accesses.

thallgren:
The current interpretation of '%s' is  "binary as unquoted characters". This is broken since there are several character sets to choose from. What encoding was/is assumed? ISO-8859-1, extended ASCII, UTF-8, or something else? Representing a `Binary` as a `String` should never be possible without knowledge of encoding. In other words, the fact that Ruby handles this badly should not be allowed to leak into Puppet.

The String converter already defines the default encoding as strict base64 (it uses '%B' if nothing else is specified). That is a good choice. A default encoding when using '%s' must also be specified. The best candidate is probably 'UTF-8' and the documentation for '%s' should be changed to "binary UTF-8 encoded string".

What if a `Binary` that is known to represent an ISO-8859-1 encoded String? Should it be possible to convert that using the `StringConverter`? Could encoding be passed as a second argument somehow? Then again, that goes both ways so perhaps we need a separate ticket for encode/decode ?

thallgren:
Please note that tests are failing for Ruby <= 2.0.0.

hlindberg:
Will look at broken tests.
Agree, that %s should state unquoted UTF-8 encoded string (it should be an error if it is not valid UTF-8). It should also be more careful when creating the string.

hlindberg:
I have addressed all comments, and I found a couple of problems myself:
* the 3.x stubs for the two new functions binary_file and find_file were missing in the PR (now added)
* the %s format now enforces UTF-8 encoding (the '%#s' format does not however as it outputs escaped bytes). It errors if strings is invalid UTF-8.
* the older rubies issue was that the base64 methods have bugs that simply truncates unpadded base64 encoded text - manual padding was added to fix this

hlindberg:
When merging this also merge the corresponding specification PR

lib/puppet/parser/functions/new.rb
```diff
@@ -626,6 +626,21 @@
...
+| ------    | ---------------
+| s         | binary as unquoted UTF-8 characters (errors if byte sequence is invalid UTF-8)
+| p         | 'Binary("<base64strict>")'
+| b         | '<base64>' - base64 string with newlines inserted
+| B         | '<base64strict>' - base64 strict string (without newlines inserted)
+| u         | '<base64urlsafe>' - base64 urlsafe string
+| t         | 'Binary' - outputs the name of the type only
+| T         | 'BINARY' - output the name of the type in all caps only
+
+The alternate form flag `#` will quote the binary or base64 text output
```
thallgren: A '%#s' doesn't just quote the string. It also uses escaped bytes rather than as UTF-8. I think that deserves a mention.

spec/unit/pops/types/p_binary_type_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+
+        it 'the Binary type is equal to itself only' do
+          code = <<-CODE
+            $t = Binary
+            notice(Binary =~ Type[ Binary ])
+            notice(Binary == Binary)
+            notice(Binary < Binary)
+            notice(Binary > Binary)
+          CODE
+          expect(eval_and_collect_notices(code)).to eql(['true', 'true', 'false', 'false'])
```
thallgren: The test doesn't assert what it claims to assert.

spec/unit/pops/types/p_binary_type_spec.rb
```diff
@@ -0,0 +1,218 @@
...
+    it "has a length method in ruby returning the length measured in bytes" do
+      # \u{1f452} is "woman's hat emoji - 4 bytes in UTF-8"
+      a_binary = PBinaryType::Binary.new("\u{1f452}")
+      expect(a_binary.length).to be(4)
+    end
+  end
+
+end
+end
+end
```
thallgren: A new type should also be added to the `all_types` array defined in `spec/shared_contexts/types_setup.rb`. Doing so will test it together with other types in several different ways. It will also test that the type is serializable though, so perhaps such an addition must wait until Binary serialization has been added? Need to be remembered (this also applies to the new `Sensitive` type. It hasn't been added there either).

lib/puppet/parser/functions/new.rb
hlindberg: ok, will fix

lib/puppet/parser/functions/new.rb
```diff
@@ -810,7 +827,59 @@
...
+```
+
+The formats have the following meaning:
+
+| format | explanation |
+| ----   | ----        |
+| b | The data is in base64 encoding, padding as required by base64 strict is added by default
+| u | The data is in URL safe base64 encoding
+| B | The data is in base64 strict encoding
+| s | The data is a puppet string used verbatim as a binary byte sequence
```
hlindberg: well, unfortunately it is not specified in Puppet that strings in the language are UTF-8. The only thing we know is that strings created from text in a .pp file is UTF-8 since every .pp file is UTF-8. Strings returned from functions are however not required to have any particular encoding since they can represent binary file content or some string. It should really not be that way. I can change the wording of the explanation and say something like "used as a binary byte sequence with encoding ignored".

spec/unit/pops/types/p_binary_type_spec.rb
hlindberg: Add a ticket?

spec/unit/pops/types/p_binary_type_spec.rb
hlindberg: This should be replaced by adding the Binary type in the types_setup.rb - not meaningful to have another place to check assignability against all other types.
The test here (that you commented on) is quite meaningless. I am removing it.

lib/puppet/parser/functions/new.rb
thallgren: I don't think that wording is correct either. The string does have an encoding, even if that encoding is unknown. It can't be ignored. Here, it's apparently up to Ruby to turn the string into a sequence of bytes. How does it do that?

Let's assume we would implement this in Java. A String in Java uses 16-bit characters. Would that mean that you'd get twice as many bytes using this method?

spec/unit/pops/types/p_binary_type_spec.rb
thallgren: It's been added. See [PUP-6706](https://tickets.puppetlabs.com/browse/PUP-6706).

lib/puppet/parser/functions/new.rb
```diff
@@ -864,7 +866,7 @@
 | b | The data is in base64 encoding, padding as required by base64 strict is added by default
 | u | The data is in URL safe base64 encoding
 | B | The data is in base64 strict encoding
-| s | The data is a puppet string used verbatim as a binary byte sequence
+| s | The data is a puppet string seen as a binary sequence of bytes with ignored encoding
```
thallgren: I don't think that wording is correct either. The string, when represented as bytes, does have an encoding, even if that encoding is unknown. It can't be ignored. Here, it's apparently up to Ruby to turn the string into a sequence of bytes. How does it do that?

Let's assume we would implement this in Java where a string uses 16-bit characters. Would that mean that you'd get twice as many bytes using this method?

hlindberg: Eh, it works the other way around. Ruby encodes when an operation takes place on the string. It is always a binary byte sequence internally. The ruby method that returns the bytes simply returns them. If you implemented this in Java, the Binary byte buffer created from a UTF-16 string would contain two bytes per character (or more since UTF-16 also has multi-chars IIRC). Do you have a suggestion on how to phrase this? Technically it takes a copy and enforces an '8 bit ascii' encoding (which is ruby's "no encoding, any byte sequence allowed" - i.e. "ignored encoding").

hlindberg: It needs to be phrased so that it is understandable by a puppet user without having to understand things like forced encoding - "uses the bytes of the string, whatever they are" to use a layman's term. Will change to "seen as the verbatim byte sequence representing the string in its given encoding" in anticipation you find that more correct.

thallgren: What I don't understand, is why we must expose the internal workings of one specific platform (Ruby) to the end user. That whole thing seems broken to me. Perhaps I don't understand the use-case, but the way I see it, there's just no way you can create a Binary from a string unless a decoding is given. The "use the bytes of the string, whatever they are" is a misnomer, since some string implementations aren't even based on bytes. The conversion opens up a can of worms as it exposes the internal runtime implementation in a bad way (to my knowledge, 8-bit-ascii doesn't exist outside of Ruby).

Can you describe how this is supposed to be used?

hlindberg: The use case is to enable use of Binary instead of just a String (as it is currently used), but where the encoding of the content is not known (it is just binary/raw data - it may be encoded in some text encoding, but that is not known). Ruby does not have a byte buffer, so currently functions return ascii 8 bit strings, or UTF-8 strings. I understand your concern regarding platform independence - would it help if it is specified as "8 bit ascii" (which I believe translates to ISO-8859-1 in Java as it allows any byte values in the string even if it implies that they represent particular glyphs). The alternative is to add the ability to specify the wanted encoding. Those are however platform specific. In any case, a port to some other language would need to revise what functions return - if it should return a Binary or a String and it would need to know when it is doing the reading. If you in Java were to do the stupid thing of creating a string that is really a binary, you could read it as ISO-8859-1.

Once the Binary type is introduced in Puppet we can deprecate / enforce (in some major version) that all puppet strings must be UTF-8, and that Binary is used for everything else.

hlindberg: The specification can state that for Ruby, the string will retain its encoding and used byte by byte. Other (future) platforms must create text with the correct encoding such that when mapped to a String, and then to a Binary using Binary.new(str, '%s') that it retains the read original byte sequence verbatim. It can also state that functions should not do this, and instead directly return Binary.

thallgren: "8-bit-ascii" does not help, since its just as platform dependent as saying nothing.

When converting a Binary to a String using '%s', the encoding will be "UTF-8". The most consistent approach would be to then say that 's' here, also means "UTF-8". If not, then converting to and through using '%s' and 's' will give somewhat surprising results.

My preference would therefore be to say that the default encoding is "UTF-8" but that other encodings can be specified. We must then either not allow "8-bit-ascii" since it doesn't exist outside the Ruby community or allow it but raise a red flag stating that it is 100% platform specific and should be avoided at length. My preference would be the former and instead provide a set of encodings known to be standardized.

hlindberg: There must be a way to get "the encoding that is used by the String", which in Ruby means "whatever" - it is not in our power to change that right here. We also do not have any other platforms to consider atm. When we do, that implementation must be done such that reading a binary byte sequence into a String (a totally stupid thing to do that should be forbidden as Binary should be used, but if you went ahead and did that anyway) it must either use the same style as Ruby and embed the encoding in the string, or be read so that Binary(str, '%s') produces the same byte sequence as was originally read (details how it does that is of no importance but could be reading ISO-8859-1, and then converting with the same encoding).

hlindberg: In puppet language the distinction between how something is read is done by calling `file >> String` (assumed but not enforced UTF-8), or `binary_file >> Binary` (raw, bytes). We have however no control whatsoever over functions in the wild, and there is no specification other than the existing behavior where puppet is using the encoding embedded in the Ruby string. Note that the `file` read does not enforce UTF-8 - as it performs a binary read and any later requirement that the format must be UTF-8 makes things blow up later.

thallgren: Do you agree that '%s' should use the same encoding when converting to/from a Binary? If the answer to that is yes, then "%s" has to mean "UTF-8". Some other format must then be added to mean "whatever" or, perhaps "ISO-8859-1". As you mentioned before, it's a valid encoding that will accept the full 8-bit range of bytes. I wouldn't be surprised if it's a no-op in Ruby (i.e. bytes == string characters).

hlindberg:
THe %s for Binary.new is now changed to require valid UTF-8, or string transcodable to valid UTF-8. A new format %r is added to do what %s did earlier. That solution is symmetrical (that is `String.new(bin, "%s")` and `Binary.new(str, %s)` both operate on UTF-8 strings). The `%r` format is specified as a backwards compatibility temporary solution that eventually will be deprecated.

PR #5040, opened 2016-06-21
(PUP-6318) Add --no-batch-files option to install.rb

McdonaldSeanp:
Cherry-pick https://github.com/puppetlabs/puppet/pull/4968 and https://github.com/puppetlabs/puppet/pull/4979 to stable

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -475,7 +475,7 @@ FileUtils.cd File.dirname(__FILE__) do
   #build_ri(ri) if InstallOptions.ri
   do_configs(configs, InstallOptions.config_dir) if InstallOptions.configs
   do_bins(bins, InstallOptions.bin_dir)
-  do_bins(windows_bins, InstallOptions.bin_dir, 'ext/windows/') if $operatingsystem == "windows"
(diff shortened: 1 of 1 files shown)


geoffnichols:
👍

joshcooper:
Ping @McdonaldSeanp @geoffnichols does this need to be merged to stable

McdonaldSeanp:
@joshcooper This is a judgement call. During a previous mergeup from master to stable of P-A we grabbed some of the windows/vanagon work, which caused some breakages. The breakages happened because I added some --no-batch-files options to the puppet/facter/hiera/mco configs in P-A which were carried over during the mergeup. But the component changes to actually create those options weren't picked up yet. 

So this is one of those component changes to create the option. The judgement call comes in because the builds are no longer failing, during the build process there will be a warning about this option not existing, but nothing will fail.

Because we are not shipping the windows/vanagon MSI for 1.5.3 this can be left alone and should not affect any other platform.

joshcooper:
So sounds like we can close the PR?

McdonaldSeanp:
I believe so yes.

PR #5041, opened 2016-06-21
(PUP-6398) Executing parallel specs on Windows is not reliable

DLuCJ:
Previously, executing rake parallel:spec on Windows would fail,
even when passing on other platforms.  The thought was that the
tests may improperly share state or the like on Windows.

This PR makes changes in three parts.  

One commit resolves puppet/util/plist namespace issues in a few 
spec tests by removing the require dependency on a 
Puppet.features.cfpropertylist? conditional in the corresponding 
lib/provider file.  The conditional is moved into plist.rb.  This brings 
extra code into scope, but the tests in question no longer rely on a 
separate test to bring in puppet/util/plist.  Alternatively, we could declare 
Puppet::Util::Plist namespace in each test. 

The second commit makes the following changes:

* Fixes colliding directory read/writes in installer_spec and
    upgrader_spec by stubbing Puppet.settings[:module_working_dir]
    to a new directory before each test in installer_spec and in
    upgrader_spec.

    unpacker_spec was changed to follow a similar pattern due to
    discussion about Puppet settings maybe becoming immutable after
    first load in the future.

* Guards the "when adding users and groups to the catalog" describe
    block in settings_spec on Windows.  Previously, settings_spec
    would fail on two iterations of the path_to_uri() call in
    puppet/util.rb due to the Windows platform check being
    stubbed to false. So the statement:

    path = '/' + path

    would not be executed and URI::Generic.build(params) call
    would fail to parse the path with an "Expected absolute path"
    error.

    However, there were some tests in the block that passed
    vacuously on Windows, so Windows-specific versions of those
    tests were moved into the "on Microsoft Windows" describe
    block, which already had a test related to adding users and
    groups to catalogs.

The third makes the following changes:

On Windows, $CHILD_STATUS is nil while the tests in
spec/unit/provider/service/*.rb. and pkg_spec are running.
This results in a "Can't modify frozen object"
runtime error.  However, the tests are able to pass
when run serially with the tests in capability_spec, for
example. The reason being, some tests in capability_spec
make a compile_to_catalog() call which, as a side
effect, generates a pid.  $CHILD_STATUS is able to
latch on to this pid, resolving the issue.

The bug shows up when the tests are run in isolation or in
parallel (on occasion, when the tests are grouped in
such a way and executed in such an order that
$CHILD_STATUS is not properly set by another test before
these tests run).

* Replaces $CHILD_STATUS.exitstatus calls with stubbed / mocked
    Puppet::Util::Execution.exitstatus calls in execution_spec
    and pacman_spec

* In the other specs, it generates a pid for $CHILD_STATUS by
    spinning up a process.

    The previous strategy does not work for these files because
    they call into $CHILD_STATUS in lib/provider files, not just
    in the specs. And, Puppet::Util::Execution.exitstatus is a
    private class method, so that it  is not possible to outright
    replace these $CHILD_STATUS.exitstatus calls.  This may be
    resolved with some refactoring in the future.

    Also, require 'English' does not set $CHILD_STATUS on
    Windows as it does on other platforms.  After requiring,
    $CHILD_STATUS (and $?) is still nil.

### DIFF
--- a/lib/puppet/provider/package/pkgdmg.rb
+++ b/lib/puppet/provider/package/pkgdmg.rb
@@ -9,7 +9,7 @@
 # in /var/db/.puppet_pkgdmg_installed_<name>
 
 require 'puppet/provider/package'
-require 'puppet/util/plist' if Puppet.features.cfpropertylist?
+require 'puppet/util/plist'
 require 'puppet/util/http_proxy'
 
 Puppet::Type.type(:package).provide :pkgdmg, :parent => Puppet::Provider::Package do
--- a/spec/unit/module_tool/applications/installer_spec.rb
+++ b/spec/unit/module_tool/applications/installer_spec.rb
@@ -4,6 +4,8 @@ require 'puppet_spec/module_tool/shared_functions'
 require 'puppet_spec/module_tool/stub_source'
 require 'semver'
 
+require 'tmpdir'
+
 describe Puppet::ModuleTool::Applications::Installer do
   include PuppetSpec::ModuleTool::SharedFunctions
   include PuppetSpec::Files
@@ -30,6 +32,13 @@ describe Puppet::ModuleTool::Applications::Installer do
     SemanticPuppet::Dependency.clear_sources
     installer = Puppet::ModuleTool::Applications::Installer.any_instance
     installer.stubs(:module_repository).returns(remote_source)
+  end
+
+  if Puppet.features.microsoft_windows?
+    before :each do
+      Puppet.settings.stubs(:[])
+      Puppet.settings.stubs(:[]).with(:module_working_dir).returns(Dir.mktmpdir('installertmp'))
+    end
   end
 
   def installer(modname, target_dir, options)
@@ -358,6 +367,6 @@ describe Puppet::ModuleTool::Applications::Installer do
         end
       end
     end
-
   end
+
 end
--- a/spec/unit/module_tool/applications/upgrader_spec.rb
+++ b/spec/unit/module_tool/applications/upgrader_spec.rb
@@ -3,6 +3,7 @@ require 'puppet/module_tool/applications'
 require 'puppet_spec/module_tool/shared_functions'
 require 'puppet_spec/module_tool/stub_source'
 require 'semver'
+require 'tmpdir'
 
 describe Puppet::ModuleTool::Applications::Upgrader do
   include PuppetSpec::ModuleTool::SharedFunctions
@@ -29,6 +30,13 @@ describe Puppet::ModuleTool::Applications::Upgrader do
     SemanticPuppet::Dependency.clear_sources
     installer = Puppet::ModuleTool::Applications::Upgrader.any_instance
     installer.stubs(:module_repository).returns(remote_source)
+  end
+
+  if Puppet.features.microsoft_windows?
+    before :each do
+      Puppet.settings.stubs(:[])
+      Puppet.settings.stubs(:[]).with(:module_working_dir).returns(Dir.mktmpdir('upgradertmp'))
+    end
   end
 
   def upgrader(name, options = {})
--- a/spec/unit/provider/package/pacman_spec.rb
+++ b/spec/unit/provider/package/pacman_spec.rb
@@ -438,13 +438,13 @@ EOF
 
     it 'should raise an error on non-zero pacman exit without a filter' do
       executor.expects(:open).with('| /usr/bin/pacman -Sgg 2>&1').returns 'error!'
-      $CHILD_STATUS.stubs(:exitstatus).returns 1
+      Puppet::Util::Execution.expects(:exitstatus).returns 1
       expect { described_class.get_installed_groups(installed_packages) }.to raise_error(Puppet::ExecutionFailure, 'error!')
     end
 
     it 'should return empty groups on non-zero pacman exit with a filter' do
       executor.expects(:open).with('| /usr/bin/pacman -Sgg git 2>&1').returns ''
-      $CHILD_STATUS.stubs(:exitstatus).returns 1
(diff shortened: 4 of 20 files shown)


lib/puppet/provider/package/pkgdmg.rb
```diff
@@ -9,7 +9,7 @@
 # in /var/db/.puppet_pkgdmg_installed_<name>
 
 require 'puppet/provider/package'
-require 'puppet/util/plist' if Puppet.features.cfpropertylist?
+require 'puppet/util/plist'
```
Iristyle: I would probably isolate the `plist` bits into its own commit

spec/unit/provider/package/pacman_spec.rb
```diff
@@ -10,6 +10,12 @@
   let(:resource) { Puppet::Type.type(:package).new(:name => 'package', :provider => 'pacman') }
   let(:provider) { described_class.new(resource) }
 
+  if Puppet.features.microsoft_windows?
```
Iristyle: Is this the best thing for us to do on Windows?  What about running these tests in parallel makes this necessary?

Probably the fastest way to create / exit a process on Windows is `cmd.exe /c "exit 0"` - but I'm not quite understanding why this is necessary for `$CHILD_STATUS` ... is there no other way to stub that constant?

spec/unit/provider/service/launchd_spec.rb
```diff
@@ -3,6 +3,10 @@
 
 require 'spec_helper'
 
+if Puppet.features.microsoft_windows?
```
Iristyle: I thought this bit wasn't necessary anymore given the change to `lib/puppet/util/plist` ?

spec/unit/provider/service/launchd_spec.rb
```diff
@@ -65,7 +75,7 @@
   end
 
   [[10, '10.6'], [13, '10.9']].each do |kernel, version|
-    describe "when checking whether the service is enabled on OS X #{version}" do
+    describe "when checking whether the service is enabled on OS X #{version}", :unless => Puppet.features.microsoft_windows? do
```
Iristyle: Since the behavior here is mostly mocked / stubbed... is it best to guard, or best to fix any mocking within Windows?

spec/unit/provider/package/pacman_spec.rb
Iristyle: Looking at this in a little more detail, it looks like maybe `require 'English'` is sufficient to bring `$CHILD_STATUS` into scope on Windows.

I wonder if we can then add it to the top-level include that all specs drag in.... I believe that is `spec/spec_helper.rb`

spec/unit/provider/service/launchd_spec.rb
Iristyle: Looks like the top of `launchd.rb` has this:

```
require 'puppet/util/plist' if Puppet.features.cfpropertylist?
```

That can likely be edited as you've done elsewhere....

spec/unit/settings_spec.rb
```diff
@@ -1377,8 +1377,6 @@ def assert_accessing_setting_is_deprecated(settings, setting)
     describe "when adding users and groups to the catalog" do
       before do
         Puppet.features.stubs(:root?).returns true
-        Puppet.features.stubs(:microsoft_windows?).returns false
```
Iristyle: We'll have to look at this one more closely... this one might be a can of worms.

spec/unit/provider/package/pacman_spec.rb
Iristyle: So it looks like any place that uses `$CHILD_STATUS` like that... can probably be replaced with:

```ruby
Puppet::Util::Execution.expects(:exitstatus).returns 1
```

lib/puppet/provider/package/pkgdmg.rb
DLuCJ: done

DLuCJ:
Difference between Appveyor pass/fail is a modified commit message.
 
Also, error looks to be a separate issue from my changes, particularly because Appveyor history is showing later builds to be failing with the same.

spec/unit/module_tool/applications/installer_spec.rb
```diff
@@ -32,6 +34,12 @@
     installer.stubs(:module_repository).returns(remote_source)
   end
 
+  if Puppet.features.microsoft_windows?
+    before :each do
+      Puppet.settings[:module_working_dir] = Dir.mktmpdir('installertmp')
```
Iristyle: We have talked recently about making settings immutable after first load, so this change would add tech debt that later needs to be cleaned up... is there another better way that we could do this? (Stubbing the value with a test framework would be better if that's possible - but if there are other better solutions that would be preferred)

spec/unit/module_tool/applications/upgrader_spec.rb
```diff
@@ -31,6 +32,12 @@
     installer.stubs(:module_repository).returns(remote_source)
   end
 
+  if Puppet.features.microsoft_windows?
+    before :each do
+      Puppet.settings[:module_working_dir] = Dir.mktmpdir('upgradertmp')
```
Iristyle: Likewise here for modifying Puppet settings at runtime.

Iristyle:
Could we also pull all the `$CHILDSTATUS` related bits into a separate commit in the PR so that they're isolated logically?

spec/unit/provider/service/launchd_spec.rb
DLuCJ: My mistake here
Mocking on Windows looks to be fine. Error was also caused by plist namespace issues.

Iristyle:
Closing in favor of #5114

PR #5043, opened 2016-06-22
(PUP-6423) scheduled_task command case insensitive

Iristyle:
- Previously, a scheduled_task command was treated as case sensitive,
   even though Windows paths / commands are case insensitive.

   Adjust the task to be more Windows friendly.


lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -112,6 +112,11 @@ def trigger
     @triggers
   end
 
+  def command_insync?(current, should)
+    should_value = File.expand_path(should[0]).gsub(/\//, '\\')
```
Iristyle: Why is `should` an array?

Iristyle:
Needs some additional tests

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
joshcooper: Puppet::Property is inherently multivalued but often it's used to manage a single value. So you'll see lots of things like:

```ruby
should.all? { |value| value == .. }
```

Iristyle: Ah yeah, thanks

lib/puppet/provider/scheduled_task/win32_taskscheduler.rb
```diff
@@ -401,6 +405,10 @@ def validate_trigger(value)
 
   private
 
+  def canonical_command(command)
+    File.expand_path(command).gsub(/\//, '\\')
```
joshcooper: I've always thought `expand_path` was a good way to canonicalize paths in ruby, but it recently came to light that it expands `~` based on the user's `HOME`, which isn't good. Can Windows paths contains `~`? If so, we might want to use `Pathname` to canonicalize?

joshcooper: According to https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx, windows paths can contain `~`, which leads to problems like:

```ruby
irb(main):025:0> File.expand_path("C:\\Users\\Administrator").casecmp(File.expand_path("~")) == 0
=> true
```

I'd recommend we instead use `File.identical?`:

```ruby
irb(main):020:0> File.identical?("C:\\Users\\Administrator", "c:/users/administrator")
=> true
irb(main):021:0> File.identical?("C:\\Users\\Administrator", "~")
=> false
```

joshcooper:
I'm ok continuing to use `File.expand_path` for now, and filing a separate ticket to evaluate usages of `File.identical?` instead. Also, should we target 4.10.x for this fix?

glennsarti:
@joshcooper Did you file a ticket for this? if so, what was the number?

joshcooper:
@glennsarti I don't believe I filed one

glennsarti:
This PR still looks like it should be merged.  Code looks ok. @joshcooper You ok with me to move this along?

lib/puppet/type/scheduled_task.rb
```diff
@@ -39,6 +39,10 @@
       # we are slash-insensitive. See #13009
       File.expand_path(value).gsub(/\//, '\\')
     end
+
+    def insync?(current)
+      provider.command_insync?(current, @should)
```
joshcooper: What if `current` or `@should` are `:absent` or `nil`?

In the not-nil/absent case, is `@should` an array containing a single element or is it a String?

Iristyle:
Closing this as any work for this will land in https://github.com/puppetlabs/puppetlabs-scheduled_task

PR #5044, opened 2016-06-22
(PUP-6094) puppet --help should show help

joshcooper:
Previously, calling `puppet --help` would generate an error, because the
absence of a subcommand meant we created a NilSubcommand with an arg
count greater than 0:

    Error: Could not parse application options: invalid option: --help

This commit alters the `find_subcommand` logic so that if a subcommand
is not specified, but --help or -h are, then we return an
ApplicationSubcommand for the `help` application. The result is the same
as if the user had typed `puppet help`, and will show global help for
all installed applications.

Since we return an ApplicationSubcommand and the subcommand is not agent
or master, we update the $LOAD_PATH allowing puppet to return help for
applications in modules. So the behavior of --help vs puppet help is the
same.

### DIFF
--- a/lib/puppet/util/command_line.rb
+++ b/lib/puppet/util/command_line.rb
@@ -81,7 +81,11 @@ module Puppet
 
       def find_subcommand
         if subcommand_name.nil?
-          NilSubcommand.new(self)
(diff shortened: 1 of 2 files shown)


MikaelSmith:
👍

PR #5047, opened 2016-06-22
(PUP-3182) Call Indirection#save with the correct arguments

joshcooper:
Previously, `puppet catalog download` failed with an obscure error:

    Instance name "<hostname>" does not match requested key {}

The problem occurred because Puppet::Indirector::Face called
Puppet::Indirector::Indirection#save(key, options) with the wrong set of
arguments. The save method expects to be called with:

    def save(instance, key = nil, options={})

Note how the instance to save should be the first argument, key is the
second, and options is optional. But Face was passing key as the instance,
and options as the key, leading to "does not match requested key".

The code originally worked in 2.7.x, but passing options never worked
correctly for any face application. During a security event, we added
validation to the Indirection methods, ensuring the instance being
passed in matched the key that the client was authorized for[1]. So for
example, you can't be authorized to send a report, but send something
else in the body of the request. This change was made in
f877cf5d63ea4b6d3bc110af6212e5187f900ee9, and released in 3.1.1, but
broke saving face-related models.

This commit modifies Puppet::Indirector::Face to call the save method
with the correct arguments. It also updates `puppet catalog download`
to output where the catalog was downloaded to:

    Notice: Saved catalog for <host> to /opt/puppetlabs/puppet/cache/client_yaml/catalog/<host>.yaml

[1] http://projects.puppetlabs.com/issues/19392

### DIFF
--- a/lib/puppet/face/catalog.rb
+++ b/lib/puppet/face/catalog.rb
@@ -123,7 +123,7 @@ Puppet::Indirector::Face.define(:catalog, '0.0.1') do
 
       Puppet::Resource::Catalog.indirection.terminus_class = :yaml
       Puppet::Face[:catalog, "0.0.1"].save(catalog)
-      Puppet.notice "Saved catalog for #{Puppet[:certname]} to yaml"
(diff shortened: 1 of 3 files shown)


joshcooper:
https://access.redhat.com/solutions/1373553

PR #5049, opened 2016-06-22
(WIP)(PUP-5359) scheduled_task can't use 'last'

Iristyle:
- TODO: add some docs / notices about usage of 'last'

Note the following manifest works with 28, but uncomment the line that
uses 'last' in the schedule and watch it explode.

```puppet
 scheduled_task { 'Test-Task':
   ensure  => 'present',
   command => "C:\windows\system32\notepad.exe",
   enabled => true,
   trigger => {
     schedule => monthly,
     on => [1,3,17,28],
     start_time => '08:00',
     months => [1, 3, 5]
   }
 }
```

The returned HRESULT for the SetTrigger call is -2147024809.:  The parameter is incorrect


Iristyle:
I think this is going to end up turning into documentation changes, until we can get task scheduler v2 support in.

glennsarti:
This has been commited as part of https://github.com/puppetlabs/puppet/pull/6194

PR #5051, opened 2016-06-22
(PUP-6428) Preserve HOME env variable in gem provider

abottchen:
Prior to this patch, the HOME environment variable was stripped
before the gem command was executed by the provider.  Since the
gem command uses the HOME environment variable to locate the
appropriate gemrc file to load, this was causing it to look at
/.gemrc.  Since Puppet runs as root, it is far more reasonable for
a customer to expect to find the gemrc file in /root/.gemrc.  This
patch maintains the previously existing HOME environment variable,
which forces gem to look in /root/.gemrc during puppet runs.

### DIFF
--- a/lib/puppet/provider/package/gem.rb
+++ b/lib/puppet/provider/package/gem.rb
@@ -33,7 +33,7 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
     end
 
     begin
-      list = execute(gem_list_command).lines.
+      list = execute(gem_list_command, {:custom_environment => {"HOME"=>ENV["HOME"]}}).lines.
         map {|set| gemsplit(set) }.
         reject {|x| x.nil? }
     rescue Puppet::ExecutionFailure => detail
@@ -125,7 +125,7 @@ Puppet::Type.type(:package).provide :gem, :parent => Puppet::Provider::Package d
 
     command += install_options if resource[:install_options]
 
-    output = execute(command)
+    output = execute(command, {:custom_environment => {"HOME"=>ENV["HOME"]}})
     # Apparently some stupid gem versions don't exit non-0 on failure
     self.fail "Could not install: #{output.chomp}" if output.include?("ERROR")
   end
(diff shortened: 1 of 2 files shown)


branan:
`HOME` can have pretty broad effects on commands that might run (including potentially the setup scripts for Gems that are being installed). Is there a more gem-specific variable that can be set to make us do the right thing here?

I'm also not sure looking in root's homedir is ever actually correct for things running as services, but if Rubygems doesn't have a system location that it can look  in, ~root is as good as it gets :disappointed:

branan:
You might also be interested in https://github.com/puppetlabs/puppet/pull/4928

abottchen:
The gem command will also look for system-wide settings in /etc/gemrc.  In my testing, this worked just as well as placing settings in /.gemrc.

That said, even if a customer uses a system-wide /etc/gemrc file to avoid this issue, I've noticed is that gem will create a .gem directory to store state data on gem repositories.  It automatically creates this directory in HOME, so by default we are currently placing this directory in "/.gem".  With this patch, it is placed in "/root/.gem", which seems more correct for a process running as root.

As far as other variables, from looking at /usr/lib/ruby/site_ruby/1.8/rubygems/config_file.rb, it appears the gemrc path is either /etc/gemrc, ~/.gemrc, or a path specified on the gem command line in the --config-file parameter.  I do not see any calls out to environment variables for customization aside from a call to USERPROFILE, which I'm assuming is for Windows.

All that said, I think this patch may be exposing an issue with the custom_environment parameter for command().  Using that setting does address the issue of getting the path corrected in the command and allows it to function, but it seems to interfere with the return code and makes Puppet think the package resource is successful even when it is not.  This PR should not be merged until this additional issue is resolved.

abottchen:
I updated the PR to include failonfail and combine.  According to https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/execution.rb#L136-L140, they default to false when parameters are passed to execute(), which was causing the runs to report success even on failure with this patch.  This updated version avoids that issue by setting them as well.

flokli:
@branan ping

branan:
I think the path forward here is to close #5560, rebase this to use the open ticket (since we closed PUP-6428 as a dup of the other), and land this one - this looks much more complete than the other PR.

I still have concerns about passing HOME around, but there's little we can do about it given how gems work, unfortunately :(

flokli:
@branan okay. @abottchen, can you squash and rebase?

branan:
Closing this in favor of #6099

PR #5052, opened 2016-06-22
(PUP-6430) Add Fedora 24 support for DNF

petems:
* Also add spec for default provider-ness

### DIFF
--- a/lib/puppet/provider/package/dnf.rb
+++ b/lib/puppet/provider/package/dnf.rb
@@ -28,7 +28,7 @@ Puppet::Type.type(:package).provide :dnf, :parent => :yum do
       end
   end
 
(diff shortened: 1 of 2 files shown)


whopper:
@petems thanks a bunch for doing this! You've been killing it lately! :D 

This is definitely good, and now the question is whether we should retarget it at stable for Puppet 4.5.3 (/cc @kylog). We even have F24 boxes in the vmpooler so I can hopefully test that F24 DNF plays nicely with the DNF provider.

kylog:
I'm fine repointing this at stable for 4.5.3. Thanks @petems!

whopper:
Gave this a quick manual test in F24, looks good! @petems I'll merge after you retarget this against stable! Thanks again!

petems:
Re-opened against stable in https://github.com/puppetlabs/puppet/pull/5060 👍

PR #5054, opened 2016-06-23
(maint) Stub operatingsystem fact in SMF provider test

whopper:


spec/unit/provider/service/smf_spec.rb
```diff
@@ -20,6 +20,7 @@
     FileTest.stubs(:executable?).with('/usr/sbin/svcadm').returns true
     FileTest.stubs(:file?).with('/usr/bin/svcs').returns true
     FileTest.stubs(:executable?).with('/usr/bin/svcs').returns true
+    Facter.stubs(:value).with(:operatingsystem).returns('solaris')
```
MikaelSmith: If you're going to do this, I think you should also stub osfamily. The goal should be to get smf to be the default service provider in these tests, via https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/service/smf.rb#L17.

whopper: @MikaelSmith updated

MikaelSmith:
:+1:

PR #5055, opened 2016-06-23
(PUP-6376) Add failover logic to the agent

branan:
This adds master failover logic to the agent using the new `server_list` setting. The failover happens early in the configurer, with gating to ensure it doesn't happen when the legacy `server` setting is being used, or if the configurer is actually an `apply` run.

The use of the selected master is handled using Puppet's `lookup` system. This allows us to scope the bound server info to all logic that happens within a configuration run, without needing to refactor the entire indirector.

The changes on the indirector side are fairly simple - check if server/port have been bound with `lookup` and if not fall back to the `server` and `masterport` settings.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -181,16 +181,6 @@ class Puppet::Configurer
   # which accepts :tags and :ignoreschedules.
   def run(options = {})
     pool = Puppet::Network::HTTP::Pool.new(Puppet[:http_keepalive_timeout])
-    begin
-      Puppet.override(:http_pool => pool) do
-        run_internal(options)
-      end
-    ensure
-      pool.close
-    end
-  end
-
-  def run_internal(options)
     # We create the report pre-populated with default settings for
     # environment and transaction_uuid very early, this is to ensure
     # they are sent regardless of any catalog compilation failures or
@@ -200,6 +190,41 @@ class Puppet::Configurer
     init_storage
 
     Puppet::Util::Log.newdestination(report)
+
+    begin
+      Puppet.override(:http_pool => pool) do
+
+        # Skip failover logic if the server_list setting is empty
+        if Puppet.settings[:server_list].nil? || Puppet.settings[:server_list].empty?
+          do_failover = false;
+        else
+          do_failover = true
+        end
+        # When we are passed a catalog, that means we're in apply
+        # mode. We shouldn't try to do any failover in that case.
+        if options[:catalog].nil? && do_failover
+          found = find_functional_server()
+          server = found[:server]
+          if server.nil?
+            Puppet.warning "Could not select a functional puppet master"
+            server = [nil, nil]
+          end
+          Puppet.override(:server => server[0], :serverport => server[1]) do
+            Puppet.debug "Selected master: #{server[0]}:#{server[1]}"
+
+            run_internal(options.merge(:node => found[:node]))
+          end
+        else
+          run_internal(options)
+        end
+      end
+    ensure
+      pool.close
+    end
+  end
+
+  def run_internal(options)
+    report = options[:report]
 
     # If a cached catalog is explicitly requested, attempt to retrieve it. Skip the node request,
     # don't pluginsync and switch to the catalog's environment if we successfully retrieve it.
@@ -233,7 +258,7 @@ class Puppet::Configurer
       # We only need to find out the environment to run in if we don't already have a catalog
       unless (options[:catalog] || Puppet[:strict_environment_mode])
         begin
-          if node = Puppet::Node.indirection.find(Puppet[:node_name_value],
+          if node = options[:node] || Puppet::Node.indirection.find(Puppet[:node_name_value],
               :environment => Puppet::Node::Environment.remote(@environment),
               :configured_environment => configured_environment,
               :ignore_cache => true,
@@ -327,6 +352,35 @@ class Puppet::Configurer
     Puppet.pop_context
   end
   private :run_internal
+
+  def find_functional_server()
+    configured_environment = Puppet[:environment] if Puppet.settings.set_by_config?(:environment)
+
+    node = nil
+    selected_server = Puppet.settings[:server_list].find do |server|
+      # Puppet.override doesn't return the result of its block, so we
+      # need to handle this manually
+      found = false
+      server[1] ||= Puppet[:masterport]
+      Puppet.override(:server => server[0], :serverport => server[1]) do
+        begin
+          node = Puppet::Node.indirection.find(Puppet[:node_name_value],
+              :environment => Puppet::Node::Environment.remote(@environment),
+              :configured_environment => configured_environment,
+              :ignore_cache => true,
+              :transaction_uuid => @transaction_uuid,
+              :fail_on_404 => false)
+          found = true
+        rescue Exception => e
+          # Nothing to see here
+        end
+      end
+      found
+    end
+    { :node => node,
+      :server => selected_server }
+  end
+  private :find_functional_server
 
   def send_report(report)
     puts report.summary if Puppet[:summarize]
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1395,7 +1395,25 @@ EOT
     },
     :server => {
       :default => "puppet",
-      :desc => "The puppet master server to which the puppet agent should connect."
+      :desc => "The puppet master server to which the puppet agent should connect.",
+      :call_hook => :on_initialize_and_write,
+      :hook => proc { |value|
+        if Puppet.settings.set_by_config?(:server) && Puppet.settings.set_by_config?(:server_list)
+          Puppet.deprecation_warning('Attempted to set both server and server_list. Server setting will not be used.', :SERVER_DUPLICATION)
+        end
+      }
+    },
+    :server_list => {
+      :default => [],
+      :type => :server_list,
+      :desc => "The list of puppet master servers to which the puppet agent should connect,
+        in the order that they will be tried.",
+      :call_hook => :on_initialize_and_write,
+      :hook => proc { |value|
+        if Puppet.settings.set_by_config?(:server) && Puppet.settings.set_by_config?(:server_list)
+          Puppet.deprecation_warning('Attempted to set both server and server_list. Server setting will not be used.', :SERVER_DUPLICATION)
+        end
+      }
     },
     :use_srv_records => {
       :default    => false,
--- a/lib/puppet/indirector/request.rb
+++ b/lib/puppet/indirector/request.rb
@@ -178,7 +178,7 @@ class Puppet::Indirector::Request
     return(uri ? uri : "/#{indirection_name}/#{key}")
   end
 
-  def do_request(srv_service=:puppet, default_server=Puppet.settings[:server], default_port=Puppet.settings[:masterport], &block)
+  def do_request(srv_service=:puppet, default_server=nil, default_port=nil, &block)
     # We were given a specific server to use, so just use that one.
     # This happens if someone does something like specifying a file
     # source using a puppet:// URI with a specific server.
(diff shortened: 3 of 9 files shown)


branan:
Important notes:

* This is still missing spec tests for the new functionality, although in theory it now cleanly passes the existing specs
* We include #5034, since that setting is needed for this logic
* A pass at reporting was made in the workshop session Monday, but I pulled that out for now since it's got additional tests that need to be cleaned up.

lib/puppet/configurer.rb
```diff
@@ -224,12 +249,12 @@ def run_internal(options)
     end
 
     begin
+      configured_environment = Puppet[:environment] if Puppet.settings.set_by_config?(:environment)
+
```
branan: This got moved because I was originally doing failover logic in this area, but moved it up because it was easier to wrap the bindings around all of run_internal than indent the entire body of this function by another block. I'll move this back to its old location on the next update of this PR

lib/puppet/configurer.rb
```diff
@@ -181,25 +181,50 @@ def apply_catalog(catalog, options)
...
+        # Skip failover logic if the server_list setting is empty
+        if Puppet.settings[:server_list].nil? || Puppet.settings[:server_list].empty?
+          do_failover = false;
+        else
+          do_failover = true
+        end
+        # When we are passed a catalog, that means we're in apply
+        # mode. We shouldn't try to do any failover in that case.
+        if options[:catalog].nil? && do_failover
+          server = find_functional_server();
```
kylog: You're writing Ruby now ;)

branan: Can we just make semicolons part of the Puppet Ruby style guide? That seems easier than me changing my habits ;)

lib/puppet/configurer.rb
```diff
@@ -181,25 +181,50 @@ def apply_catalog(catalog, options)
   # which accepts :tags and :ignoreschedules.
   def run(options = {})
     pool = Puppet::Network::HTTP::Pool.new(Puppet[:http_keepalive_timeout])
+    # We create the report pre-populated with default settings for
+    # environment and transaction_uuid very early, this is to ensure
+    # they are sent regardless of any catalog compilation failures or
+    # exceptions.
+    options[:report] ||= Puppet::Transaction::Report.new("apply", nil, @environment, @transaction_uuid)
+    report = options[:report]
```
branan: Report initialization is moved up here so that logging around master failover ends up in the report (and also for PUP-6378 adding the selected master to the report)

branan:
That appveyor failure looks unrelated to my changes - something in `registry_spec.rb`

whopper:
My completely unrelated PR is having the same appveyor issue, so definitely unrelated.

whopper:
Nothing jumps out at me as being terribly broken here, but I think it'd be good to get some expert eyes on this. Perhaps @joshcooper or @MikaelSmith if you have some time to spare?

lib/puppet/configurer.rb
```diff
@@ -181,25 +181,50 @@ def apply_catalog(catalog, options)
...
     end
   end
 
   def run_internal(options)
-    # We create the report pre-populated with default settings for
-    # environment and transaction_uuid very early, this is to ensure
-    # they are sent regardless of any catalog compilation failures or
-    # exceptions.
-    options[:report] ||= Puppet::Transaction::Report.new("apply", nil, @environment, @transaction_uuid)
     report = options[:report]
```
MikaelSmith: This is used. I see we're not consistent about using `report` vs `options[:report]` later on.

lib/puppet/configurer.rb
```diff
@@ -328,6 +353,31 @@ def run_internal(options)
...
+  def find_functional_server()
+    configured_environment = Puppet[:environment] if Puppet.settings.set_by_config?(:environment)
+
+    Puppet.settings[:server_list].find do |server|
+      # Puppet.override doesn't return the result of its block, so we
+      # need to handle this manually
+      found = false
+      Puppet.override(:server => server[0], :serverport => server[1]) do
+        begin
+          Puppet::Node.indirection.find(Puppet[:node_name_value],
```
MikaelSmith: This appears to introduce an additional node request. Is there any way we can return that and avoid calling it later?

branan: Yeah, I can probably pass that around.

branan:
@MikaelSmith updated to pass that info around. The `find_functional_server` method is getting a little clunky looking. I'm not sure if that's just the nature of the beast or if I'm missing some useful ruby-fu here. Any suggestions?

lib/puppet/configurer.rb
```diff
@@ -233,7 +259,7 @@ def run_internal(options)
       # We only need to find out the environment to run in if we don't already have a catalog
       unless (options[:catalog] || Puppet[:strict_environment_mode])
         begin
-          if node = Puppet::Node.indirection.find(Puppet[:node_name_value],
+          if node = options[:node] || Puppet::Node.indirection.find(Puppet[:node_name_value],
```
MikaelSmith: Will this do assignment correctly? Not sure about `=` vs `||` precedence.

lib/puppet/configurer.rb
```diff
@@ -181,25 +181,51 @@ def apply_catalog(catalog, options)
...
+        # mode. We shouldn't try to do any failover in that case.
+        if options[:catalog].nil? && do_failover
+          found = find_functional_server()
+          server = found[:server]
+          if server.nil?
+            Puppet.warning "Could not select a functional puppet master"
+            server = [nil, nil]
+          end
+          server[1] ||= Puppet.settings[:masterport]
+          Puppet.override(:server => server[0], :serverport => server[1]) do
```
MikaelSmith: What's the reason we shouldn't just update Puppet.settings?

lib/puppet/settings/server_list_setting.rb
```diff
@@ -0,0 +1,20 @@
...
+  def type
+    :server_list
+  end
+
+  def munge(value)
+    servers = super 
+    servers.map! { |server| 
+      case server
+      when String
+        server.split(':')
```
MikaelSmith: Split on `:`? I thought it was comma-separated.

MikaelSmith:
I think this looks fine, just some clarifying questions.

lib/puppet/settings/server_list_setting.rb
branan: the `super` call handles comma-sepration via the `array` class - this is splitting each `server:port` pair into their array, so a string of the form `"foo:123,bar:456"` becomes `[["foo",123],["bar",456]]`

lib/puppet/configurer.rb
branan: If we do that now it's one more thing we have to change when we move away from manipulating settings at runtime. Taking advantage of puppet's existing dynamic binding functionality seemed better than doing something in a way we already know we don't like

lib/puppet/configurer.rb
branan:
```
2.2.4 :006 > foo = nil
 => nil 
2.2.4 :007 > bar = "baz"
 => "baz" 
2.2.4 :008 > qux = foo || bar
 => "baz" 
2.2.4 :009 > qux
 => "baz"
```

branan:
Rubocop doesn't like me having the result of `find` and the lambda's argument be the same name (I guess it's *strictly* correct about shadowing, but it's still kinda wrong here). I'm gonna push up a quick change to clean that up

MikaelSmith:
👍

lib/puppet/configurer.rb
```diff
@@ -328,6 +353,35 @@ def run_internal(options)
...
 
+  def find_functional_server()
+    configured_environment = Puppet[:environment] if Puppet.settings.set_by_config?(:environment)
+
+    node = nil
+    selected_server = Puppet.settings[:server_list].find do |server|
+      # Puppet.override doesn't return the result of its block, so we
+      # need to handle this manually
+      found = false
+      server[1] ||= Puppet[:masterport]
```
branan: I just moved this "missing port" logic from the `run` method, since we need a port in order to contact the master when checking if it exists. This fixes a bug @Magisus found when working on the reporting aspect of this.

MikaelSmith:
Failed CI.

branan:
It looks like our understanding of when settings hooks get called and when the set_by_config? predicate is true is not quite correct - this will require a bit more debugging/digging :(

branan:
So there were two issues

1) Hooks got called more often than we expected, which required us to add an additional check in each hook that the setting being called was set explicitly, not just the opposing setting.
2) The way settings were implemented, a hook wouldn't be able to tell that its associated setting was explicitly set until after the hook returned.

This now adds a bit of an exception-handling dance to `Puppet::Settings::Values` to ensure that a setting can know it was explicitly set, while still preserving the previous behavior that if a hook threw the setting would be un-changted. That code has been squashed into @Magisus' original setting commit

leandramang:
its ok pupet hug

leandramang:
thank you so much!!! <3 puppet linuexdies

leandramang:
ok puppet

leandramang:
questions? puppet!

lib/puppet/indirector/request.rb
```diff
@@ -198,8 +198,19 @@ def do_request(srv_service=:puppet, default_server=Puppet.settings[:server], def
 
     # ... Fall back onto the default server.
     Puppet.debug "No more servers left, falling back to #{default_server}:#{default_port}" if Puppet.settings[:use_srv_records]
```
MikaelSmith: I think this message should be changed to report the correct server:port.

branan: oh, yup

lib/puppet/defaults.rb
```diff
@@ -1395,7 +1395,25 @@ def self.default_diffargs
     },
     :server => {
       :default => "puppet",
-      :desc => "The puppet master server to which the puppet agent should connect."
+      :desc => "The puppet master server to which the puppet agent should connect.",
+      :call_hook => :on_initialize_and_write,
+      :hook => proc { |value|
+        if Puppet.settings.set_by_config?(:server) && Puppet.settings.set_by_config?(:server_list)
+          Puppet.deprecation_warning('Attempted to set both server and server_list. Server setting will not be used.', :SERVER_DUPLICATION)
```
MikaelSmith: Ah, this differs subtly from the route we went in pxp-agent. Because we made it command-line only, we could say that it's an error if both are in the config file, but if you use the command-line it overrides either. Here you don't differentiate between things set on the command-line or in a config file.

MikaelSmith:
There's still a concern about doing an unnecessary node request when using cached catalogs, but I'm comfortable addressing that separately.

lib/puppet/indirector/rest.rb
```diff
@@ -34,12 +34,44 @@ def self.srv_service
...
+      end
+    end
   end
 
   def self.port
-    Puppet.settings[port_setting || :masterport].to_i
+    setting = port_setting()
+    srv_setting = server_setting()
+    if (setting && setting != :masterport && Puppet.settings.set_by_config?(setting))
+        || (srv_setting && srv_setting != :server && Puppet.settings.set_by_config?(srv_setting))
```
branan: Latest push adds this change to the default port selection logic. This now says "If this indirection has a custom server set, but no custom port set, we should use the related port setting EVEN IF it's unset" We were encountering an issue in our testing where we'd set `--server_list=garbage:123 --report_server=localhost` , and we were failing to submit reports to the working report server since the previous logic here said "`report_port` is unset, so I'll use what came from failover"

branan:
This fixes my dumb syntax error and adds a test for the new expected server/port setting interaction

lib/puppet/indirector/rest.rb
```diff
@@ -193,6 +230,10 @@ def save(request)
   # to request.do_request from here, thus if we change what we pass or how we
   # get it, we only need to change it here.
   def do_request(request)
+    if self.class.port == 0
+      require 'pry'
```
MikaelSmith: Probably don't want this.

lib/puppet/property.rb
```diff
@@ -424,6 +424,7 @@ def set(value)
     name = self.class.value_name(value)
     if method = self.class.value_option(name, :method) and self.respond_to?(method)
       begin
+        puts "sending #{method.inspect}"
```
MikaelSmith: Or this.

branan: I very carefully `git add -p`'d the bits I wanted, then used my "AMEND IT ALL INTO ONE COMMIT AND FORCE PUSH" script 


Because I am smart

PR #5057, opened 2016-06-24
(maint) Don't run Facter so much

MikaelSmith:
Facter was run on every agent multiple times. Also, if we had two
Windows boxes in the acceptance run we could get the incorrection
version string. Fix so we only run facter on the box we're testing.


glennsarti:
Closed in preference to https://github.com/puppetlabs/puppet/pull/5059

PR #5058, opened 2016-06-24
(PUP-1498) Fix scheduled_task acceptance test on ruby 1.9.3

glennsarti:
Ruby 1.9.3 has an ecoding bug in REXML, where it disregards the text encoding on a
string in preference for the XML header encoding value.  This causes an issue because
the XML file is encoded as UTF8.  In this commit we modify the XML Encoding header
returned by schtasks.exe to be UTF-8 not UTF-16 if the ruby version matches 1.9x

Additionally fixed a typo in acceptance test as it was calling facter on all agents
instead of the agent under test

The encoding gets very strange because beaker is running a command over SSH so it is hard to determine exactly what the encoding should be.  In this case the string always comes back as UTF8

### DIFF
--- a/acceptance/tests/resource/scheduled_task/should_modify.rb
+++ b/acceptance/tests/resource/scheduled_task/should_modify.rb
@@ -11,7 +11,7 @@ agents.each do |agent|
   # Have to use /v1 parameter for Vista and later, older versions
   # don't accept the parameter
   version = '/v1'
-  on agents, facter('kernelmajversion') do
+  on agent, facter('kernelmajversion') do
     version = '' if stdout.chomp.to_f < 6.0
   end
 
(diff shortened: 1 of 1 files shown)


Iristyle:
@glennsarti I don't believe we use `SERVER` ticket numbers in puppet commits (especially for one that's about fixing CI).  I think we should reference the original ticket number that cause this problem in the first place.

glennsarti:
Also includes the patch from https://github.com/puppetlabs/puppet/pull/5057

Iristyle:
This should be targeted at stable and merged up to master... closing in favor of a new PR.

glennsarti:
New PR against stable at https://github.com/puppetlabs/puppet/pull/5059

PR #5060, opened 2016-06-26
(PUP-6430) Add Fedora 24 support for DNF

petems:
* Also add spec for default provider-ness

Closes #5052 and re-targets against stable

### DIFF
--- a/lib/puppet/provider/package/dnf.rb
+++ b/lib/puppet/provider/package/dnf.rb
@@ -28,7 +28,7 @@ Puppet::Type.type(:package).provide :dnf, :parent => :yum do
       end
   end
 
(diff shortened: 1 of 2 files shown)


geoffnichols:
👍

shermdog:
I'd like to propose we change the provider to default to DNF for >= 22. If and when there is a new package management tool we can limit the scope then :)

petems:
@shermdog Agreed for Fedora 25, but that'll take time and review before someone extends the `defaultfor` method to accept operators, but I'll open a ticket for that for the future 👍

petems:
Just opened https://tickets.puppetlabs.com/browse/PUP-6452 for that 👍

PR #5061, opened 2016-06-27
(PUP-5909) Add Serializer and Deserializer and a JSON packer/unpacker

thallgren:
This PR adds what's needed to perform serialization/deserialization of primitives, arrays, hashes, and complex objects that are described using the Puppet Object type. The mechanism is designed to work on differnent protocols using an underlying abstract reader/writer. A reader/writer pair for JSON is included in this PR.

### DIFF
--- a/lib/puppet/pops/serialization/serializer.rb
+++ b/lib/puppet/pops/serialization/serializer.rb
@@ -1,0 +1,91 @@
+require_relative 'extension'
+
+module Puppet::Pops
+module Serialization
+  # The serializer is capable of writing, arrays, maps, and complex objects using an underlying protocol writer. It takes care of
+  # tabulating and disassembling complex objects.
+  # @api public
+  class Serializer
+    # @param [AbstractWriter] writer the writer that is used for writing primitive values
+    # @api public
+    def initialize(writer)
+      @written = {}
+      @writer = writer
+    end
+
+    # Tell the underlying writer to finish
+    # @api public
+    def finish
+      @writer.finish
+    end
+
+    # Write an object
+    # @param [Object] value the object to write
+    # @api public
+    def write(value)
+      case value
+      when Integer, Float, String, true, false, nil, Time
+        @writer.write(value)
+      when :default
+        @writer.write(Extension::Default::INSTANCE)
+      else
+        index = @written[value.object_id]
+        if index.nil?
+          write_tabulated_first_time(value)
+        else
+          @writer.write(Extension::Tabulation.new(index)) unless index.nil?
+        end
+      end
+    end
+
+    # Write the start of an array.
+    # @param [Integer] size the size of the array
+    # @api private
+    def start_array(size)
+      @writer.write(Extension::ArrayStart.new(size))
+    end
+
+    # Write the start of a map (hash).
+    # @param [Integer] size the number of entries in the map
+    # @api private
+    def start_map(size)
+      @writer.write(Extension::MapStart.new(size))
+    end
+
+    # Write the start of a complex object
+    # @param [String] type_ref the name of the type
+    # @param [Integer] attr_count the number of attributes in the object
+    # @api private
+    def start_object(type_ref, attr_count)
+      @writer.write(Extension::ObjectStart.new(type_ref, attr_count))
+    end
+
+    # First time write of a tabulated object. This means that the object is written and then remembered. Subsequent writes
+    # of the same object will yield a write of a tabulation index instead.
+    # @param [Object] value the value to write
+    # @api private
+    def write_tabulated_first_time(value)
+      @written[value.object_id] = @written.size
+      case value
+      when Symbol, Regexp, Semantic::Version, Semantic::VersionRange
+        @writer.write(value)
+      when Array
+        start_array(value.size)
+        value.each { |elem| write(elem) }
+      when Hash
+        start_map(value.size)
+        value.each_pair { |key, val| write(key); write(val) }
+      when Types::PTypeReferenceType
+        @writer.write(value)
+      when Types::PuppetObject
+        value._ptype.write(value, self)
+      else
+        impl_class = value.class
+        type = Loaders.implementation_registry.type_for_module(impl_class)
+        raise SerializationError, "No Puppet Type found for #{impl_class.name}" unless type.is_a?(Types::PObjectType)
+        type.write(value, self)
+      end
+    end
+  end
+end
+end
--- a/lib/puppet/pops/types/p_sem_ver_range_type.rb
+++ b/lib/puppet/pops/types/p_sem_ver_range_type.rb
@@ -5,6 +5,10 @@ module Types
 #
 # @api public
 class PSemVerRangeType < PScalarType
+  def self.register_ptype(loader, ir)
+    create_ptype(loader, ir, 'ScalarType')
+  end
+
   # Check if a version is included in a version range. The version can be a string or
   # a `Semantic::SemVer`
   #
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -11,6 +11,7 @@ require_relative 'type_factory'
 require_relative 'type_parser'
 require_relative 'class_loader'
 require_relative 'type_mismatch_describer'
+require_relative 'puppet_object'
 
 module Puppet::Pops
 module Types
@@ -32,14 +33,40 @@ EMPTY_HASH = {}.freeze
 #
 # TODO: See PUP-2978 for possible performance optimization
 class TypedModelObject < Object
+  include PuppetObject
   include Visitable
   include Adaptable
+
+  def self._ptype
+    @type
+  end
+
+  def self.create_ptype(loader, ir, parent_name, attributes_hash = EMPTY_HASH)
+    @type = Pcore::create_object_type(loader, ir, self, "Pcore::#{simple_name}Type", "Pcore::#{parent_name}", attributes_hash)
+  end
+
+  def self.register_ptypes(loader, ir)
+    types = []
+    Types.constants.each do |c|
+      cls = Types.const_get(c)
+      next unless cls.is_a?(Class) && cls < self
+      type = cls.register_ptype(loader, ir)
+      types << type unless type.nil?
+    end
+    tp = TypeParser.singleton
+    types.each { |type| type.resolve(tp, loader) }
+  end
 end
 
 # Base type for all types
 # @api public
 #
 class PAnyType < TypedModelObject
+
+  def self.register_ptype(loader, ir)
+    @type = Pcore::create_object_type(loader, ir, self, 'Pcore::AnyType', 'Any', EMPTY_HASH)
+  end
+
   # Accept a visitor that will be sent the message `visit`, once with `self` as the
   # argument. The visitor will then visit all types that this type contains.
   #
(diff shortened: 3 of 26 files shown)


lib/puppet/pops/types/p_sem_ver_range_type.rb
```diff
@@ -5,6 +5,10 @@ module Types
 #
 # @api public
 class PSemVerRangeType < PScalarType
+  def self.register_ptype(loader, ir)
+    create_ptype(loader, ir, 'ScalarType')
```
hlindberg: What is 'ScalarType'? Why not a frozen constant?

lib/puppet/pops/types/types.rb
```diff
@@ -2799,6 +2800,12 @@ def really_instance?(o, guard = nil)
     resolved_type.really_instance?(o, guard)
   end
 
+  # @return `nil` to prevent serialization of the type_expr used when first initializing this instance
+  # @api private
+  def type_expr
+    nil
```
hlindberg: Why is this needed for serialiation - later types are backed by Pcore and I assume pcore is directing the serialization.
Mostly curious.

hlindberg: BTW, there is a typo in the commit message "type_exp"

lib/puppet/pops/types/types.rb
```diff
@@ -364,6 +402,16 @@ def eql?(o)
...
 class PType < PTypeWithContainedType
+
+  def self.register_ptype(loader, ir)
+    create_ptype(loader, ir, 'AnyType',
+       'type' => {
+         KEY_TYPE => POptionalType.new(PType::DEFAULT),
+         KEY_VALUE => nil
+       }
+    )
+  end
```
hlindberg: isn't this type abstract? Why have this here -seems repeated in derived classes (I looked at PNotUndefType).

lib/puppet/pops/types/types.rb
```diff
@@ -403,7 +451,7 @@ def eql?(o)
     self.class == o.class && @type == o.type
   end
 
-  def simple_name
+  def self.simple_name
     # since this the class is inconsistently named PType and not PTypeType
```
hlindberg: Maybe we should change that?

lib/puppet/pops/types/p_sem_ver_range_type.rb
thallgren: Using a constant would not be ideal since the string becomes a type reference that is later resolved to the parent type. The parent type in turn, has a name, which is frozen. A frozen constant here would just result in two frozen strings with the same content (the name of the is constructed in the #create_ptype method and later frozen in the initializer, not reachable using a constant).

lib/puppet/pops/types/types.rb
thallgren: The Object reader/writer assumes that all attributes that are passed to the initializer also can be accessed using an attribute reader. This particular attribute is very special since it is volatile (the expression is only used in the creation of the alias and then meaningless after the expression has been resolved).

lib/puppet/pops/types/types.rb
thallgren: No, this type is not abstract. It's a meta-type. No meta-types are abstract.
Not sure what you mean with "repeated". This is a class method that registers the meta-type. The method must be present on all classes that implement such a meta-type.

lib/puppet/pops/types/p_sem_ver_range_type.rb
hlindberg: ok not important - it was just that the same string is created multiple times = more garbage

lib/puppet/pops/types/types.rb
hlindberg: I got that - should we support volatile (non serialized) attributes in Pcore perhaps? Would that solve this case?

lib/puppet/pops/types/types.rb
hlindberg: It is derived from PTypeWithContainedType - cannot all derived classes from PTypeWithContainedType use one implementation?
There were several occurrences of self.register_ptype with what looked like the very same call in several classes derived from PTypeWithContainedType.

thallgren: There's no Pcore::TypeWithContainedType so I didn't want to add a registration method to that class. An alternative approach could be to do that and then add an exception for it in the loop that registers all meta-types. I chose the current approach because it felt cleaner.

lib/puppet/pops/types/types.rb
thallgren: Sure. While we're at it, we should also change `PHostClassType` to `PClassType` since type is named `Class`. Separate JIRA?

lib/puppet/pops/types/types.rb
thallgren: It would, at least partly. I've created a generic Object reader/writer that makes the assumption that the arguments passed to the initializer are accessible as readers. In this case the initializer takes three arguments, the last one (the resolved expression) is optional, but only if the second argument (the unresolved expression) is given instead. Conversely, the second argument is of course not needed when the third argument is given. It must be passed in as `nil` though, since unnamed parameters are used.

lib/puppet/pops/serialization/serializer.rb
```diff
@@ -0,0 +1,91 @@
...
+    def finish
+      @writer.finish
+    end
+
+    # Write an object
+    # @param [Object] value the object to write
+    # @api public
+    def write(value)
+      case value
+      when Integer, Float, String, true, false, nil, Time
```
hlindberg: Integer includes Big integers. Think we need to be more specific in Pcore. Also heads up that Ruby MRI plans to change the Integer impl to not have separate types for Big numbers.
Puppet Language specifies Integer to be unsigned 64 bit integer. Not quite sure how we should deal with that here though.

spec/unit/pops/serialization/packer_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+    end
+
+    it 'Integer' do
+      val = 32
+      write(val)
+      val2 = read
+      expect(val2).to be_a(Integer)
+      expect(val2).to eql(val)
+    end
+
```
hlindberg: test for error on out of range integers for 64 bit unsigned

spec/unit/pops/serialization/packer_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+    it 'Regexp' do
+      val = /match me/
+      write(val)
+      val2 = read
+      expect(val2).to be_a(Regexp)
+      expect(val2).to eql(val)
+    end
+
+    it 'Time created by TimeFactory' do
+      # It does succeed on rare occasions, so we need to repeat
```
hlindberg: huh?

spec/unit/pops/serialization/packer_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      val = TimeFactory.now
+      write(val)
+      val2 = read
+      expect(val2).to be_a(Time)
+      expect(val2).to eql(val)
+    end
+
+    it 'Version' do
+      # It does succeed on rare occasions, so we need to repeat
```
hlindberg: huh huh ?

spec/unit/pops/serialization/packer_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      val = Semantic::Version.parse('1.2.3-alpha2')
+      write(val)
+      val2 = read
+      expect(val2).to be_a(Semantic::Version)
+      expect(val2).to eql(val)
+    end
+
+    it 'VersionRange' do
+      # It does succeed on rare occasions, so we need to repeat
```
hlindberg: - " -

spec/unit/pops/serialization/packer_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+      val = Semantic::VersionRange.parse('>=1.2.3-alpha2 <1.2.4')
+      write(val)
+      val2 = read
+      expect(val2).to be_a(Semantic::VersionRange)
+      expect(val2).to eql(val)
+    end
+
+    it 'can write and read Time created by TimeFactory' do
+      failed = false
+      # It does succeed on rare occasions, so we need to repeat
```
hlindberg: You must mean "it fails on rare occasions" - and that is probably not true either since the intent here is to test that the implementation is robust and avoids brittle time methods that IF those were used then there may be a failure when iterating. (Otherwise, if we have an implementation that may fail - then that is not very good). Think you need to elaborate on the comment.

spec/unit/pops/serialization/serialization_spec.rb
```diff
@@ -0,0 +1,228 @@
...
+    it 'Regexp' do
+      val = /match me/
+      write(val)
+      val2 = read
+      expect(val2).to be_a(Regexp)
+      expect(val2).to eql(val)
+    end
+
+    it 'Time created by TimeFactory' do
+      # It does succeed on rare occasions, so we need to repeat
```
hlindberg: Not adding comments about these comments below...

lib/puppet/pops/types/types.rb
hlindberg: Should we consider bypassing `initialize` and instead poke values into instance attributes directly for special cases?

lib/puppet/pops/types/types.rb
hlindberg: sure (separate ticket(s)).

lib/puppet/pops/types/types.rb
thallgren: I would prefer not to as it seems like an unclean design (less clean than what we have here). I don't think it's needed.

lib/puppet/pops/serialization/serializer.rb
thallgren: That sounds like we need a new type (a BigInteger) and changes in how type inference is made. I think that deserves a separate JIRA.

spec/unit/pops/serialization/packer_spec.rb
thallgren: As mentioned in earlier comment, I think we need a new type with type inference and all. New test cases must be added to serialization once we add that. Until we do, use of big integers in puppet cannot be deterministic.

thallgren:
Pushed a commit that is rebased on master. It also contains a revised version of the packer_spec.rb where irrelevant copy pasted comments are removed and TimeFactory tests were rewritten.

lib/puppet/pops/serialization/serializer.rb
hlindberg: There are many problems with Integer. Both unsigned 64 bits, and bigints (>64 bit) in most serialization formats. Also, in puppet, these larger values do not work throughout the infrastructure. Hence, we cannot allow any Ruby Integer. On 32 bit platforms (IIRC) FixNum is not full 64 bit, and MRI switches to BigInt when overflowing 32 bits. Therefore it is not just a matter of changing the Integer to a FixNum (as it will be too restrictive on some platforms).  I think Pcore should have signed Integer 64 bit as max for integers and that it is enforced by serialization. We can have another ticket to add support for BigInt in Pcore (a separate data type).

spec/unit/pops/serialization/packer_spec.rb
hlindberg: See earlier comment on cap to 64 bit values I think this should be enforced or there is risk for silent bad behavior.

hlindberg:
AppVeyor was failing on what seems to be unrelated (non existing methods called by facter). Restarted the AppVeyor build job on this PR.

PR #5062, opened 2016-06-27
(PUP-6445) Pin Win32 gems to explicit versions

glennsarti:
The Win32-Service gem was updated to 0.8.9 however the new gem pulled in a
dependant gem called ffi-win32-extensions which has a name collision on the
FFI:Pointer.read_wide_string method.  This causes Puppet to fail in mutiple
locations.  This commit pins the Win32-Service gem to a prior version without
the dependency.

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -34,22 +34,24 @@ gem_platform_dependencies:
     gem_runtime_dependencies:
       # Pinning versions that require native extensions
       ffi: '~> 1.9.6'
-      win32-dir: '~> 0.4.9'
-      win32-eventlog: '~> 0.6.2'
-      win32-process: '~> 0.7.4'
+      # win32-xxxx gems are pinned due to PUP-6445
+      win32-dir: '= 0.4.9'
+      win32-eventlog: '= 0.6.5'
+      win32-process: '= 0.7.5'
       # Use of win32-security is deprecated
-      win32-security: '~> 0.2.5'
-      win32-service: '~> 0.8.6'
+      win32-security: '= 0.2.5'
+      win32-service: '= 0.8.7'
       minitar: '~> 0.5.4'
   x64-mingw32:
     gem_runtime_dependencies:
       ffi: '~> 1.9.6'
-      win32-dir: '~> 0.4.9'
-      win32-eventlog: '~> 0.6.2'
-      win32-process: '~> 0.7.4'
+      # win32-xxxx gems are pinned due to PUP-6445
+      win32-dir: '= 0.4.9'
+      win32-eventlog: '= 0.6.5'
+      win32-process: '= 0.7.5'
       # Use of win32-security is deprecated
-      win32-security: '~> 0.2.5'
-      win32-service: '~> 0.8.6'
+      win32-security: '= 0.2.5'
+      win32-service: '= 0.8.7'
       minitar: '~> 0.5.4'
 bundle_platforms:
   universal-darwin: ruby


ext/project_data.yaml
```diff
@@ -39,7 +39,8 @@ gem_platform_dependencies:
       win32-process: '~> 0.7.4'
       # Use of win32-security is deprecated
       win32-security: '~> 0.2.5'
-      win32-service: '~> 0.8.6'
+      # win32-service gem is pinned due to PUP-6445
+      win32-service: '= 0.8.7'
```
Iristyle: I think this is OK... but I'm not sure about Ruby compat here - have you verified the differences between `0.8.6` and `0.8.7` to make sure there are no implications there?

https://github.com/djberg96/win32-service/compare/win32-service-0.8.6...win32-service-0.8.7

glennsarti:
Last known good win32-service gem is 0.8.7 at;
https://ci.appveyor.com/project/puppetlabs/puppet/build/4.1.0.2801

First known CI failure was using 0.8.9 at;
https://ci.appveyor.com/project/puppetlabs/puppet/build/4.1.0.2824

There was a 0.8.8 (24th June) released however it was quickly outdated by 0.8.9 (26th June)

joshcooper:
puppet-agent bundles 0.8.6, maybe we should update that to 0.8.7 (sometime later, not to block this PR)? https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/rubygem-win32-service.rb

Iristyle:
I'm wondering if we should proactively pin some of the other win32* gems now before any other breaking changes sneak in?

glennsarti:
@Iristyle So mirror puppet-agent pinning perhaps?

Iristyle:
@glennsarti they are vendored at https://github.com/puppetlabs/puppet-win32-ruby/tree/2.1.9.0-x64/ruby/lib/ruby/gems/2.1.0/gems in the current (older style Puppet For The Win) MSI building process.

They are expressed a little differently in the puppet-agent repo as individual components - https://github.com/puppetlabs/puppet-agent/tree/master/configs/components

I think we're out of sync in a few spots, but I think that for the win32-* gems, that we should just pin last known good in the `project_data.yaml`

glennsarti:
Fair enough.  I'll update the commit

ext/project_data.yaml
```diff
@@ -34,22 +34,24 @@ gem_platform_dependencies:
     gem_runtime_dependencies:
       # Pinning versions that require native extensions
       ffi: '~> 1.9.6'
-      win32-dir: '~> 0.4.9'
-      win32-eventlog: '~> 0.6.2'
-      win32-process: '~> 0.7.4'
+      # win32-xxxx gems are pinned due to PUP-6445
+      win32-dir: '= 0.4.9'
+      win32-eventlog: '= 0.6.5'
```
Iristyle: Our vendored versions at https://github.com/puppetlabs/puppet-win32-ruby/tree/2.1.x-x64/ruby/lib/ruby/gems/2.1.0/gems are currently:

````yaml
win32-eventlog: '= 0.6.2`
win32-process: '= 0.7.4'
win32-service: '= 0.8.6'
```

We've already covered `win32-service` - what about the differences with the other 2?  Are those versions currently being used in AppVeyor?  If so, that's good enough for me.

glennsarti: Yes.  I took those versions from last known good CI run

Iristyle: You want to do a separate PR to https://github.com/puppetlabs/puppet-agent/tree/master/configs/components to bump those in sync?

glennsarti: Yes and I can cross reference as well.

Iristyle: After reviewing changes:

* https://github.com/djberg96/win32-eventlog/compare/win32-eventlog-0.6.2...win32-eventlog-0.6.5 - 0.6.5 is fine
* https://github.com/djberg96/win32-process/compare/win32-process-0.7.4...win32-process-0.7.5 - 0.7.5 is fine
* https://github.com/djberg96/win32-service/compare/win32-service-0.8.6...win32-service-0.8.7 - 0.8.7 is fine

PR #5063, opened 2016-06-27
(PUP-6233) Only use smf restart -s flag in Solaris 11.2 and above

whopper:
The synchronous `-s` flag for `smf restart` only exists in Solaris
11.2 and above. This commit ensures that we don't try to use it in
older Solaris versions.

See https://blogs.oracle.com/SolarisSMF/entry/synchronous_actions


whopper:
This appveyor failure is puzzling - I'm not sure what happened there.

Magisus:
Looks good to me.

PR #5064, opened 2016-06-28
(PUP-6450) Fix Windows group and user authoritative membership

Iristyle:
Fixes issues with:

* The implementation of `:auth_membership => true` for the Windows ADSI Group provider
* The implementation of `:membership => inclusive` for the Windows ADSI User provider


glennsarti:
@Iristyle This also fixes;

The implementation of `:membership => minimal` for the Windows ADSI User provider

PR #5069, opened 2016-06-28
(PUP-6452) Allow defaultfor to accept regex

petems:
- Allows easier matching for defaults. For example, Fedora 22+ uses DNF. It's easier to just give a regex of 22-29 rather than add a new one each time

Code is rough right now, but working according to the specs! 😄

### DIFF
--- a/lib/puppet/provider.rb
+++ b/lib/puppet/provider.rb
@@ -291,16 +291,8 @@ class Puppet::Provider
   end
 
   def self.fact_match(fact, values)
-    values = [values] unless values.is_a? Array
-    values.map! { |v| v.to_s.downcase.intern }
-
-    if fval = Facter.value(fact).to_s and fval != ""
-      fval = fval.to_s.downcase.intern
-
-      values.include?(fval)
-    else
-      false
-    end
+    fact_val = Facter.value(fact).to_s.downcase
+    fact_val != "" and [*values].any? { |v| v === fact_val }
   end
 
   def self.feature_match(value)
--- a/spec/unit/provider_spec.rb
+++ b/spec/unit/provider_spec.rb
@@ -279,6 +279,45 @@ describe Puppet::Provider do
       expect(subject.name).to eq(type.defaultprovider.name)
     end
 
+    describe "regex matches" do
+      it "should match a singular regex" do
+        Facter.expects(:value).with(:osfamily).at_least_once.returns "solaris"
+
+        one = type.provide(:one) do
+          defaultfor :osfamily => /solaris/
+        end
+
+        expect(one).to be_default
+      end
+
+      it "should not match a non-matching regex " do
+        Facter.expects(:value).with(:osfamily).at_least_once.returns "redhat"
+
+        one = type.provide(:one) do
+          defaultfor :osfamily => /solaris/
+        end
+
+        expect(one).to_not be_default
+      end
+
+      it "should allow a mix of regex and string" do
+
+        Facter.expects(:value).with(:operatingsystem).at_least_once.returns "fedora"
+        Facter.expects(:value).with(:operatingsystemmajrelease).at_least_once.returns "24"
+
+        one = type.provide(:one) do
+          defaultfor :operatingsystem => "fedora", :operatingsystemmajrelease => /^2[2-9]$/
+        end
+
+        two = type.provide(:two) do
+          defaultfor :operatingsystem => /fedora/, :operatingsystemmajrelease => '24'
+        end
+
+        expect(one).to be_default
+        expect(two).to be_default
+      end
+    end
+
     describe "when there are multiple defaultfor's of equal specificity" do
       before :each do
         subject.defaultfor :operatingsystem => :os1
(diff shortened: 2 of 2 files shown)


branan:
I was just thinking earlier today that this would be super useful (especially for things like yum/dnf and systemd). Thanks for taking the time to open it! I'm not gonna do a serious review at 22:30, but I'll get back to you on this as soon as I have some spare cycles.

nacc:
Would it be possible to make this more flexible than just regex? AIUI, even for Fedora, it's really F22+, not just F2{2-9}. That is, when F30 comes out, the regex is no longer correct and there would be a regression, right? Similarly for Ubuntu, we want to say 15.04 was the turning point for systemd being the default, and just be able to go off that value as to what the default init-script provider is.

kylog:
Another idea: have `defaultfor` follow the pattern of `confine` and allow a code block. Then for the dnf case the code could be something like:

```
defaultfor :true begin
  Facter.value(:operatingsystem) == :fedora && Facter.value(:operatingsystemmajrelease).to_i >= 22
end
```

I haven't looked to see if this would fly or not, just tossing out ideas :)

branan:
@kylog that's probably the way to go moving forward. In my mind that makes regex not super useful - but I'm also not sure anyone on Client will have the bandwidth to make that change. I don't think we should block small improvements on us having a better idea but not actually doing it

@petems I'm happy to merge this as-is, unless you feel like you want to take on the defaultfor accepting a block change, in which case I'd consider that to supersede this PR

petems:
@branan Right now I wouldn't have the bandwidth without some pairing on how to do that (never really got block argument stuff in Ruby). How do we feel about going for Regex in the mean time?

branan:
@petems I'd much rather have regex than nothing, and I don't think implementing  regex actively *prevents* us from doing more complex block stuff in the future. How do you feel about just getting this in @kylog ?

kylog:
Yeah, I'm +1.

petems:
Ok, I'm cool with getting this merged and opening a ticket for more complex block arguments in the future? 👍

petems:
@kylog @branan Can I get this unblocked and merged? 👍

lib/puppet/provider.rb
```diff
@@ -292,13 +292,25 @@ def self.default_match
...
+      end
+    end
 
     if fval = Facter.value(fact).to_s and fval != ""
       fval = fval.to_s.downcase.intern
-
-      values.include?(fval)
+      if values.any? {|v| v.is_a? Regexp }
+        regex_match = Regexp.union(values)
+        fval =~ regex_match
```
petems: So, I ended up asking a question over on Code Review on Stack Exchange, because I felt this logic was overly complex, any I got a suggestion that would fix the above issue _and_ make he code simpler:

```
fact_val = Facter.value(fact).to_s.downcase
fact_val != "" and [*values].any? { |v| v === fact_val }
```

So this works, and all the unit tests pass, except the ones about multiple `defaultfor`... have you got any ideas why that would be?

petems:
```
Failures:

  1) Puppet::Provider default providers should find the default provider
     Failure/Error: expect(subject.name).to eq(type.defaultprovider.name)

       expected: :nondefault
            got: :default

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -:nondefault
       +:default

     # ./spec/unit/provider_spec.rb:279:in `block (3 levels) in <top (required)>'

  2) Puppet::Provider default providers should consider any true value enough to be default
     Failure/Error: expect(subject.name).to eq(type.defaultprovider.name)

       expected: :nondefault
            got: :default

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -:nondefault
       +:default

     # ./spec/unit/provider_spec.rb:381:in `block (3 levels) in <top (required)>'

  3) Puppet::Provider default providers when there are multiple defaultfor's of equal specificity should be default for the first defaultfor
     Failure/Error: expect(provider).to be_default
       expected `Puppet::Type::Test::ProviderDefault.default?` to return true, got false
     # ./spec/unit/provider_spec.rb:332:in `block (4 levels) in <top (required)>'

  4) Puppet::Provider default providers when there are multiple defaultfor's of equal specificity should be default for the last defaultfor
     Failure/Error: expect(provider).to be_default
       expected `Puppet::Type::Test::ProviderDefault.default?` to return true, got false
     # ./spec/unit/provider_spec.rb:339:in `block (4 levels) in <top (required)>'

  5) Puppet::Provider default providers when there are multiple defaultfor's with different specificity should be default for a more specific, but matching, defaultfor
     Failure/Error: expect(provider).to be_default
       expected `Puppet::Type::Test::ProviderDefault.default?` to return true, got false
     # ./spec/unit/provider_spec.rb:357:in `block (4 levels) in <top (required)>'

  6) Puppet::Provider default providers when there are multiple defaultfor's with different specificity should be default for a more specific, but matching, defaultfor with regex
     Failure/Error: expect(provider).to be_default
       expected `Puppet::Type::Test::ProviderDefault.default?` to return true, got false
     # ./spec/unit/provider_spec.rb:365:in `block (4 levels) in <top (required)>'

  7) Puppet::Provider default providers when there are multiple defaultfor's with different specificity should be default for a less specific, but matching, defaultfor
     Failure/Error: expect(provider).to be_default
       expected `Puppet::Type::Test::ProviderDefault.default?` to return true, got false
     # ./spec/unit/provider_spec.rb:372:in `block (4 levels) in <top (required)>'

Finished in 2.86 seconds (files took 1.4 seconds to load)
233 examples, 7 failures
```

MosesMendoza:
@petems resurrecting this PR - do you want to push up your latest code that caused the test failures you referenced in your last comment? We can help debug it.

petems:
@MosesMendoza Done!

Magisus:
The spec failures are equality failures: the first group of failures are due to `values` containing "Darwin" and being compared against "darwin". In the second group, `values` contains a Symbol, `:os1`, but the Facter value is a string. So it seems like this logic might actually get more complex to correctly handle all these different cases...

Magisus:
@petems will you have time to play with this more? Probably going to need a different solution than `===`.

MosesMendoza:
@petems i'm going to push a commit up to your branch that addresses these failures if i have access, otherwise I'll close this and open a new with your changes and mine

MosesMendoza:
ok I don't have push access so i'll open a new branch/pr

MosesMendoza:
superseded by https://github.com/puppetlabs/puppet/pull/6292

PR #5070, opened 2016-06-28
(PUP-6323) Support tdnf package provider

Magisus:
The tdnf package provider is used on PhotonOS as a smaller replacement for DNF. This commit adds support for using that package provider when running on Photon OS.

### DIFF
--- a/lib/puppet/provider/package/tdnf.rb
+++ b/lib/puppet/provider/package/tdnf.rb
@@ -1,0 +1,28 @@
+Puppet::Type.type(:package).provide :tdnf, :parent => :dnf do
+  desc "Support via `tdnf`.
+
+  This provider supports the `install_options` attribute, which allows command-line flags to be passed to tdnf.
+  These options should be spcified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}), or an
+  array where each element is either a string or a hash."
+
+  has_feature :install_options, :versionable, :virtual_packages
+
+  commands :cmd => "tdnf", :rpm => "rpm"
+
+  # Note: this confine was borrowed from the Yum provider. The
+  # original purpose (from way back in 2007) was to make sure we
+  # never try to use RPM on a machine without it. We think this
+  # has probably become obsolete with the way `commands` work, so
+  # we should investigate removing it at some point.
+  if command('rpm')
+    confine :true => begin
+      rpm('--version')
+      rescue Puppet::ExecutionFailure
+        false
+      else
+        true
+      end
+  end
+
+  defaultfor :operatingsystem => "PhotonOS" 
+end
--- a/spec/unit/provider/package/tdnf_spec.rb
+++ b/spec/unit/provider/package/tdnf_spec.rb
@@ -1,0 +1,18 @@
+require 'spec_helper'
+
+# Note that much of the functionality of the tdnf provider is already tested with yum provider tests,
+# as yum is the parent provider, via dnf
+
+provider_class = Puppet::Type.type(:package).provider(:tdnf)
+
+context 'default' do
+  it 'should be the default provider on PhotonOS' do
+    Facter.stubs(:value).with(:osfamily).returns(:redhat)
+    Facter.stubs(:value).with(:operatingsystem).returns("PhotonOS")
+    expect(provider_class).to be_default
+  end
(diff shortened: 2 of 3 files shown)


lib/puppet/provider/package/tdnf.rb
```diff
@@ -0,0 +1,36 @@
...
+  This provider supports the `install_options` attribute, which allows command-line flags to be passed to tdnf.
+  These options should be spcified as a string (e.g. '--flag'), a hash (e.g. {'--flag' => 'value'}), or an
+  array where each element is either a string or a hash."
+
+  has_feature :install_options, :versionable, :virtual_packages
+
+  commands :cmd => "tdnf", :rpm => "rpm"
+
+# Note: this confine was borrowed from the Yum provider. The
+# original purpose (from way back in 2007) was to make sure we
```
whopper: @Magisus looks like you lost the remainder of this comment :)

whopper:
Other than my one comment, this looks good. I'm still going back and forth on whether this should really be a separate provider, or if we should just have a conditional that says `if PhotonOS, command = `/usr/bin/tdnf`. I suppose having a separate provider will allow us to easily patch for differences between tdnf and dnf if they pop up, so maybe this is the correct way to proceed.

lib/puppet/provider/package/tdnf.rb
```diff
@@ -0,0 +1,39 @@
...
+      rescue Puppet::ExecutionFailure
+        false
+      else
+        true
+      end
+  end
+
+  defaultfor :operatingsystem => "PhotonOS" 
+
+  def self.update_command
```
whopper: Ah, I don't believe need this since we inherit from DNF, which already overrides `update_command`.

lib/puppet/provider/package/tdnf.rb
```diff
@@ -0,0 +1,39 @@
...
+    # In TDNF, update is deprecated for upgrade
+    'upgrade'
+  end
+
+  # The value to pass to DNF as its error output level.
+  # TDNF differs from Yum slightly with regards to error outputting.
+  def self.error_level
```
whopper: Same as above

whopper:
No clue why that exec test failed. I ran specs against this branch locally / in Ubuntu and had no problems. A force push / travis rerun may clear it.

spec/unit/provider/package/tdnf_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'spec_helper'
+
+# Note that much of the functionality of the tdnf provider is already tested with yum provider tests,
+# as yum is the parent provider, via dnf
+
+provider_class = Puppet::Type.type(:package).provider(:tdnf)
+
+describe provider_class do
+  it_behaves_like 'RHEL package provider', provider_class, 'tdnf'
+end
```
whopper: @Magisus I lied, there is one test we should add here :)

Let's have a simple test case that ensures TDNF is the default provider for PhotonOS, similar to how the DNF test does it:

```
  [ 22, 23, 24 ].each do |ver|
    it "should be the default provider on fedora#{ver}" do
      Facter.stubs(:value).with(:osfamily).returns(:redhat)
      Facter.stubs(:value).with(:operatingsystem).returns(:fedora)
      Facter.stubs(:value).with(:operatingsystemmajrelease).returns("#{ver}")
      expect(provider_class).to be_default
    end
  end
```

... though more closely resembling this check in the upstart provider:

```
  it "should be the default provider on Ubuntu" do
    Facter.expects(:value).with(:operatingsystem).returns("Ubuntu")
    Facter.expects(:value).with(:operatingsystemmajrelease).returns("12.04")
    expect(described_class.default?).to be_truthy
  end
```

Magisus:
@whopper Is that appveyor failure something I should worry about? All tests are passing locally...

whopper:
@Magisus I've never seen that before, looks unrelated. I don't know how to re-run appveyor other than closing and reopening, so I did that! Let's see if it passes this time...

branan:
-1073741819 is 0xC0000005, which is "Access Violation". It'll probably clear up on a re-run, but if it doesn't we'll want to play with it in a local Windows VM to try debugging it

whopper:
It passed this time! :)

PR #5072, opened 2016-06-29
(PUP-6378) Add failover information to the report

Magisus:
This commit adds information about which master was successfully
contacted to the report under the master field. This field will be blank
if no master was used (e.g. for `puppet apply` runs) and otherwise take
the form "servername:port".

### DIFF
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -114,6 +114,11 @@
             ]
         },
 
+        "master_used": {
+            "description": "The name of the master that was used to compile the catalog. If failover occurred, this will hold the first master successfully contacted. If this run had no master (e.g. a `puppet apply` run), this field will be blank",
+            "type": "string"
+        },
+
         "report_format": {
             "description": "The report format version documented by this schema",
             "type":        "integer",
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -210,7 +210,10 @@ class Puppet::Configurer
             server = [nil, nil]
           end
           Puppet.override(:server => server[0], :serverport => server[1]) do
-            Puppet.debug "Selected master: #{server[0]}:#{server[1]}"
+            if !server.first.nil?
+              Puppet.debug "Selected master: #{server[0]}:#{server[1]}"
+              report.master_used = "#{server[0]}:#{server[1]}"
+            end
 
             run_internal(options.merge(:node => found[:node]))
           end
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -59,6 +59,10 @@ class Puppet::Transaction::Report
   # or 'on_failure'
   attr_accessor :cached_catalog_status
 
+  # Contains the name and port of the master that was successfully contacted
+  # @return [String] a string of the format 'servername:port'
+  attr_accessor :master_used
+
   # The host name for which the report is generated
   # @return [String] the host name
   attr_accessor :host
@@ -206,6 +210,7 @@ class Puppet::Transaction::Report
     @code_id = nil
     @catalog_uuid = nil
     @cached_catalog_status = nil
+    @master_used = nil
     @environment = environment
     @status = 'failed' # assume failed until the report is finalized
     @noop = Puppet[:noop]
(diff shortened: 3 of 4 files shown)


Magisus:
Upon further investigation, it seems even non-required attributes always appear in the report, even if they are not set. Current examples tend to just leave the field blank, so that is what I will do here.

api/schemas/report.json
```diff
@@ -114,6 +114,11 @@
             ]
         },
 
+        "master": {
+            "description": "The name of the master that was used to compile the catalog. If failover occurred, this will hold the first master successfully contacted. If this run had no master (e.g. a `puppet apply` run), this field will be blank",
```
whopper: Will this also be populated with a master string when failover is not used? I.e, the traditional single-master workflow?

Magisus: No wait, it should always populate I think. Maybe the description is misleading. I have to investigate if it actually *does*, but it seems to me like it *should*, unless doing a run with no master.

whopper: Agreed, given the name "master", I'd expect it to always have the name of the master regardless of whether failover happened. (The exception being when masterless (apply))

Magisus: I'm not actually sure what happens if no master can be contacted because of an error, because when I tried it no report was generated. There's a chance it would just be ":" because of the string interpolation, but I'm not sure how to test it.

If we just have the server setting and not a list, how do we get the port?

whopper: Ah right, that probably didn't work because you didn't specify a separate (functional) report server. Normally, if no masters could be contacted, that would mean that the default report server also couldn't be contacted (it's $server), and hence no report.

The report is actually always first generated agent-side, so we're guaranteed to have one during the run. If you set `report_server` to a working server (that isn't in your master list), you should be able to get it. It will probably have a lot of empty fields.

In the case where the user set `server` and not `server_list`, we'd have to check the `masterport` setting for the port. It will have its default value if it wasn't set.

lib/puppet/configurer.rb
```diff
@@ -181,25 +181,51 @@ def apply_catalog(catalog, options)
...
+        if options[:catalog].nil? && do_failover
+          found = find_functional_server()
+          server = found[:server]
+          if server.nil?
+            Puppet.warning "Could not select a functional puppet master"
+            server = [nil, nil]
+          end
+          Puppet.override(:server => server[0], :serverport => server[1]) do
+            Puppet.debug "Selected master: #{server[0]}:#{server[1]}"
+            report.master = "#{server[0]}:#{server[1]}"
```
branan: This only happens in the failover case. The non-failover path is on line 219. If we want to add the master info to that path we'd need to add a line there to set `report.master` from the old settings

api/schemas/report.json
Magisus: So when using the `server` setting and specifying a bogus master and `report_server` to localhost, I get a report with the bogus master in it. If this is not the functionality we want, we'll have to move where I'm doing the reporting. 

When using `server_list`, no report is generated. Could this be a bug with the way we're handling server_list? I know we had questions about how it would affect other servers the user had specified.

whopper: If I understand correctly:

Question 1: when only specifying a single server and it fails, should the report `master` key report the failing server name. This question seems applicable to `server_list` as well. What if we have three potential servers and each fails? In that case, no entry seems best. Perhaps then we should do the same for a single server?

Question 2: Why are we not receiving the report on the `report_server` when using `server_list`? This will require some debugging around report submission. We want to make sure that the report server is correct. If you specify `report_server` AND `server_list`, does Puppet keep the correct value in `report_server` at the point of submission in the Configurer?

api/schemas/report.json
```diff
@@ -114,6 +114,11 @@
             ]
         },
 
+        "master_used: {
```
whopper: You forgot a `"`! :)

branan:
This should be cherry-pick onto master now that my thing is merged - it looks like they're still on different versions of my commits

lib/puppet/configurer.rb
```diff
@@ -377,6 +380,7 @@ def find_functional_server()
       end
       found
     end
+
```
branan: random newline here

PR #5073, opened 2016-06-29
(PUP-6390) Unpin run order for service symlinks

johnduarte:
This commit removes the pinned run order of the expected symlinks
created for the services in the `service_enable_linux` acceptance
test.

Prior to this commit, SUSE Enterprise Linux instances on s390x were
failing this test when the generated run order did not equal the
statically defined number.

[skip-ci]


johnduarte:
Manually tested against the following

- sles-11-i386
- sles-11-s390x
- sles-12-x86_64
- sles-12-s390x
- debian-7-i386
- debian-8-i386

er0ck:
much better.  i like it! 👍

PR #5075, opened 2016-06-30
(PUP-4809) Fix yumrepo boolean value handling and spec test

ahenroid:
Boolean value parsing is now case-sensitive and values must be specified
in all lowercase. Valid values are true|false|yes|no|1|0

### DIFF
--- a/lib/puppet/type/yumrepo.rb
+++ b/lib/puppet/type/yumrepo.rb
@@ -18,8 +18,8 @@ Puppet::Type.newtype(:yumrepo) do
   # Doc string for properties that can be made 'absent'
   ABSENT_DOC="Set this to `absent` to remove it from the file completely."
   # False can be false/0/no and True can be true/1/yes in yum.
-  YUM_BOOLEAN=/^(True|False|0|1|No|Yes)$/i
-  YUM_BOOLEAN_DOC="Valid values are: False/0/No or True/1/Yes."
+  YUM_BOOLEAN=/^(true|false|0|1|no|yes)$/
+  YUM_BOOLEAN_DOC="Valid values are: false/0/no or true/1/yes."
 
   # Common munge logic for YUM_BOOLEAN values. Munges for two requirements:
   # 1) Because of how regex validation works in Puppet::Parameter::Value,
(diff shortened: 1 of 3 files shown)


whopper:
Woot, good work @ahenroid! For some reason travis / appveyor didn't run on this PR - looking into that. I'll also do a little pre-functional review and try this out with a yumrepo resource or two.

whopper:
@ahenroid it looks like puppet interprets any bare uppercased word as a puppet type, and will fail for upper case `Yes` and `No`:

```
yumrepo { 'mariadb' :
  ensure   => present,
  enabled  => Yes,
  baseurl  => 'http://yum.mariadb.org/10.0/centos7-amd64',
  gpgcheck => No,
  gpgkey   => 'https://yum.mariadb.org/RPM-GPG-KEY-MariaDB',
}


Error: Evaluation Error: Resource type not found: Yes at /root/puppet/test.pp:3:15 on node localhost.localdomain
```

I think we should just disallow anything uppercased, including `Yes` and `No`. From the Yum config reference:

"Parse a string containing a boolean value. 1, yes, and true will evaluate to True; and 0, no, and false will evaluate to False. Case is ignored."

lib/puppet/type/yumrepo.rb
```diff
@@ -17,18 +17,24 @@
...
   # Common munge logic for YUM_BOOLEAN values. Munges for two requirements:
   # 1) Because of how regex validation works in Puppet::Parameter::Value,
-  # Boolean false and lowercase false will not be considered invalid. However,
-  # if the user specified false (or true), they meant False (or True).
+  # Boolean false and lowercase false will not be considered invalid.
   # 2) In order for parameter removal to work correctly, when absent is passed
   # as a string it needs to be munged back to a symbol.
   munge_yum_bool = Proc.new do |val|
-    val.to_s == 'absent' ? :absent : val.to_s.capitalize
+    case val.to_s
```
whopper: Looking at this further (and playing with yum repos), I realized that we don't actually need to alter this munging logic at all. The only change that needs to be made is to remove the capital `True`, `False` `No` and `Yes` from `YUM_BOOLEAN_DOC`.

I believe `YUM_BOOLEAN` is fine with the capital words as we're upcase-ing the input. This is really just an error message cleanup, as it turns out.

whopper:
Looks to work correctly, though as of commit bbd4d58e in master, puppet is no longer emitting the correct error message.. I'm going to file a ticket for that.

whopper:
@ahenroid two more things: first, your commit message isn't accurate anymore and needs to be updated, and secondly I just realized there's a second set of spec tests for yumrepo, `spec/integration/provider/yumrepo_spec.rb`. There is a single failing test in that file which needs to be updated.

PR #5078, opened 2016-07-01
(PUP-6463) Improved error message when accessing unassigned parameter

thallgren:
A parameter default expression that made an attempt to access a
parameter to the left in the parameter list when that parameter in turn
was unassigned, resulted in an error message "$y tries to illegally
access not yet evaluated $x". That message should only be used when
accessing parameters to the right. An access to an unassigned parameter
to the left should result in the same message as when a required
parameter is unassigned. This commit fixes that.


thallgren:
Supersedes [PR-5077](/puppetlabs/puppet/pull/5077)

PR #5080, opened 2016-07-01
(PUP-6384) Acceptance: agent failover interacts correctly with server_list setting

Magisus:
These tests verify that warnings are issued when the user attempts to
specify both server and server_list, as well as that the server_list
setting takes priority when both are specified.

### DIFF
--- a/acceptance/tests/server_list_setting.rb
+++ b/acceptance/tests/server_list_setting.rb
@@ -1,0 +1,52 @@
+test_name "Priority of server_list setting over server setting" do
+  master_port = 8140
+
+  step "Conflict warnings for server settings"
+  with_puppet_running_on(master, {}) do
+    agents.each do |agent|
+      step "Should emit a warning when trying to set both server and server_list" do
+        step "when server is set first" do
+          on(agent, puppet("agent", "-t", "--server #{master}", "--server_list #{master}:#{master_port},another:123"), 
+            :acceptable_exit_codes => [0, 2]) do |result|
+              assert_match(/Attempted to set both server and server_list/, result.stderr, "a warning should have been issued because both server setttings were used")
+            end
+        end
+
+        step "when server_list is set first" do
+          on(agent, puppet("agent", "-t", "--server_list #{master}:#{master_port},another:123", "--server #{master}"), 
+            :acceptable_exit_codes => [0, 2]) do |result|
+              assert_match(/Attempted to set both server and server_list/, result.stderr, "a warning should have been issued because both server setttings were used")
+            end
+        end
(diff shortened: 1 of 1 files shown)


whopper:
@Magisus looking at this now. If you take the report commit out of this PR it could be merged now rather than waiting for your other PR to land (unless there's a reason that commit needs to be here for this test)

acceptance/tests/server_list_setting.rb
```diff
@@ -0,0 +1,58 @@
+test_name "Priority of server_list setting over server setting" do
+  master_port = 8140
+
+  master_opts = {
```
whopper: I think this may be unnecessary, as masterport is 8140 by default. You can do the following instead:

```
with_puppet_running_on(master, {}) do
...
end
```

acceptance/tests/server_list_setting.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  step "Conflict warnings for server settings"
+  with_puppet_running_on(master, master_opts) do
+    agents.each do |agent|
+      step "Should emit a warning when trying to set both server and server_list" do
+        step "when server is set first" do
+          on(agent, puppet("agent", "-t", "--server #{master}", "--server_list #{master}:#{master_port},another:123"), 
+            :acceptable_exit_codes => [0, 2]) do |result|
+              assert_match(/Attempted to set both server and server_list/, result.stderr, "a warning should have been issued because both server setttings were used")
+             end
```
whopper: Indentation here is funky compared to line 25 :D

acceptance/tests/server_list_setting.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+
+        step "when server_list is set first" do
+          on(agent, puppet("agent", "-t", "--server_list #{master}:#{master_port},another:123", "--server #{master}"), 
+            :acceptable_exit_codes => [0, 2]) do |result|
+              assert_match(/Attempted to set both server and server_list/, result.stderr, "a warning should have been issued because both server setttings were used")
+             end
```
whopper: Same indentation business

acceptance/tests/server_list_setting.rb
Magisus: What exactly are you referring to? Everything looks lined up to me...

whopper: The `end` has one too many spaces before it to line up with the `:`

whopper:
This looks good, and it passes! Yay!

whopper:
👍

PR #5081, opened 2016-07-01
(PUP-4660) Stop checking Puppet.features.external_facts?

ahenroid:


branan:
This still leaves the feature itself in place. I'd like to trash that, but probably it's public API that we need to leave until a major. @kylog what are your thoughts on that?

If we want to keep the feature around we should file a ticket targeted at 5.0 to remove it

joshcooper:
Good point, adding a ticket to remove completely makes sense to me. FWIW, puppet's feature code relies on method missing (IIRC), so if we do remove the feature, then it would erroneously return false:

```
irb(main):002:0> Puppet.features.foobar?
=> false
```

kylog:
To play it safe, +1 on a ticket to remove at the next major. Since this seems unlikely to be widely adopted (if at all) as an API, I'd suggest we deprecate in docs only rather than with a runtime warning.

Also, should we:
* always return true here: https://github.com/puppetlabs/puppet/blob/ed209118fceb6433b51c88c10883cb7415b36adb/lib/puppet/feature/external_facts.rb#L4
* remove this stub: https://github.com/puppetlabs/puppet/blob/ed209118fceb6433b51c88c10883cb7415b36adb/spec/unit/face/plugin_spec.rb#L15-L17

??

branan:
Yup, I agree those two changes should be part of this PR. I'll file the ticket to remove the feature in 5.0

PR #5082, opened 2016-07-01
(PUP-1134) Add correct path for init scripts on AIX

Magisus:
This commit edits the init service provider to use the correct path for
init scripts on AIX.

### DIFF
--- a/lib/puppet/provider/service/init.rb
+++ b/lib/puppet/provider/service/init.rb
@@ -11,6 +11,8 @@ Puppet::Type.type(:service).provide :init, :parent => :base do
       "/sbin/init.d"
     when "Archlinux"
       "/etc/rc.d"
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/service/init.rb
```diff
@@ -11,6 +11,8 @@ def self.defpath
       "/sbin/init.d"
     when "Archlinux"
       "/etc/rc.d"
+    when "AIX"
+      "etc/rc.d/init.d"
```
whopper: This is missing the leading `/`, which causes it to fail to find the path to `init.d` in AIX 5.3. I'm confused as to why it worked in 7.1 - maybe your local edit there had the `/`?

whopper:
👍

PR #5083, opened 2016-07-01
(PUP-6468) Add acceptance tests from PE

johnduarte:
This commit adds acceptance tests that were previously tested
as part of Puppet Enterprise. These tests are universally applicable
to the puppet product and are therefore being moved here.


johnduarte:
pending manual validation

LuvCurves:
+1

PR #5085, opened 2016-07-05
(PUP-6473) Pin FFI to 1.9.10

glennsarti:
An update to FFI (1.9.11) has caused downstream gems to error due to a bug
introduced in 1.9.11.  FFI Github issue 506 (https://github.com/ffi/ffi/issues/506)
shows this is affecting other users of FFI as well.  This commit pins FFI to the
latest known good working version of 1.9.10.  This change will be reverted in the
future once issue 506 is resolved.


jpogran:
Verified desired version of ffi installed by running bundle install.

PR #5086, opened 2016-07-05
(PUP-6384) Use empty test puppet.conf for failover test

whopper:
In addition, clean up indentation and beaker steps

[skip ci]


Magisus:
This looks good, I ran these on Windows and they all pass now 👍

PR #5087, opened 2016-07-05
(PUP-6440) Support sensitive data tagging

adrienthebo:
This commit adds a `sensitive` attribute to properties to indicate that
the `should` and `is` values of a given property are sensitive and
should be specially handled. This commit also adds a `redacted` field to
events to indicate that the data contained in this event was sensitive
and has been redacted to prevent information leakage.

### DIFF
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -594,6 +594,11 @@
                     "type":        "string"
                 },
 
+                "redacted": {
+                  "description": "Whether this event has been redacted",
+                  "type": "boolean"
+                },
+
                 "status": {
                     "description": "One of the following strings:\n\n* `success` - property was out of sync, and was successfully changed to be in sync.\n* `failure`- property was out of sync, and couldn't be changed to be in sync due to an error.\n* `noop` - property was out of sync, and wasn't changed due to noop mode.\n* `audit` - property was in sync, and was being audited.\n\ndepending on the type of the event. Always `audit` in reports of `inspect` kind.\n",
                     "enum": [
--- a/lib/puppet/property.rb
+++ b/lib/puppet/property.rb
@@ -55,6 +55,10 @@ class Puppet::Property < Puppet::Parameter
   #
   attr_writer :noop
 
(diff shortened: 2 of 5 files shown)


adrienthebo:
The schema failures are known; this change might mean changing the report format. I'll investigate this more shortly.

api/schemas/report.json
```diff
@@ -117,7 +117,7 @@
         "report_format": {
             "description": "The report format version documented by this schema",
             "type":        "integer",
-            "enum":        ["6", 6]
+            "enum":        ["7", 7]
```
branan: No need to bump this - the bump from 5 to 6 is already part of the 4.6.x series

branan:
This looks like a reasonable way to get the redaction info from property to report

lib/puppet/property.rb
```diff
@@ -55,6 +55,10 @@ class Puppet::Property < Puppet::Parameter
   #
   attr_writer :noop
 
+  # @!attribute [rw] sensitive
+  #   @return [true, false] If this property has been tagged as sensitive.
+  attr_accessor :sensitive
```
hlindberg: Isn't it of value to also have sensitive status on parameters?

hlindberg:
@adrienthebo, this looks reasonable

lib/puppet/property.rb
branan: I don't think so? We never sync parameters, so they shouldn't show up in reports or logs (unless the provider itself logs information taken from a parameter)

adrienthebo: Branan's comments are why I didn't push this to the `Puppet::Parameter` class; eventually when we pass instances of the `Sensitive` type to the RAL it may be useful - but at that point we won't need an attribute on the Parameter class and will handle the value on a per-parameter value basis.

api/schemas/report.json
adrienthebo: Fixed.

adrienthebo:
This is ready for merge on my end.

PR #5088, opened 2016-07-05
(PUP-6467) Fix Parallel Spec grouping for RSpec 3.5.0

DLuCJ:
Rspec_grouper currently tries to access an array of example groups
through a private function children() in RSpec::Core::ExampleGroups
in order to calculate example counts per spec file.  With the RSpec
3.5.0 release, the example groups are no longer accessible through
children().  So, rspec_grouper cannot find example groups and is
reporting that no examples can be found.

This commit uses RSpec.world.example_groups instead to access the
array.


DLuCJ:
will close once #5098 goes through

PR #5091, opened 2016-07-06
(PUP-6347) Use Puppet syntax in programmatic string representation

thallgren:
Before this commit, the '%p' format and some alternative '#' formats
for string would output a programmatic string representation that
was using Ruby escapes. This commit changes this so that the
programmatic string representation conforms to Puppet syntax.

The following rules apply when the programmatic string representation
is created:
- The following codepoints will be escaped on output:
  * 0x0a => \n
  * 0x0d => \r
  * 0x09 => \t
  * 0x22 => \"
  * 0x24 => \$
  * 0x5c => \\
- All other characters with a codepoint < 0x20 or > 0x7f are represented
  using \u{X} syntax (where X is a hexadecimal representation of the codepoint).
- Strings that are safe to quote using single quotes will be single
  quoted.

### DIFF
--- a/lib/puppet/pops/types/string_converter.rb
+++ b/lib/puppet/pops/types/string_converter.rb
@@ -557,7 +557,7 @@ class StringConverter
 
   def string_PDefaultType(val_type, val, format_map, _)
     f = get_format(val_type, format_map)
-    format_literal(f, case f.format
+    apply_string_flags(f, case f.format
     when :d, :s, :p
       f.alt? ? '"default"' : 'default'
     when :D
@@ -570,7 +570,7 @@ class StringConverter
   # @api private
   def string_PUndefType(val_type, val, format_map, _)
     f = get_format(val_type, format_map)
-    format_literal(f, case f.format
+    apply_string_flags(f, case f.format
     when :n
       f.alt? ? 'null' : 'nil'
     when :u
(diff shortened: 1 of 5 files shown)


hlindberg:
Ping @peterhuene is the C++ implementation also using single quotes if it is safe to do so?
A thought why not always use double quotes as that makes the output predictable without knowledge of the string content (what Ruby does).

thallgren:
On the other hand, if the motivation for a "programmatic" format is that it can be used to produce code that is readable by Puppet, then why not make that code as efficient as possible? I know that's not what Ruby does but why let a bad example set the standard?

hlindberg:
Nevermind - lets use single quotes when possible, not worth haggling about. I do however not accept as a fact that Ruby's way is "a bad example" - that is a matter of opinion.

If we are focusing on optimizing reading the string back in again (a rare use case I would imagine) instead of human comprehension of what is in the string, 
then I want to understand the rules for optimizing to single quote. Is a string like `"\$foo"` optimized to `'$foo'` (since in SQ form, the $ has not meaning)? (Reading PR again).

hlindberg:
ok, so the rule as implemented only treats simple ascii strings as single quotable - that is actually good as all complicated strings then are treated the same. That scheme will not lead to the kind of problems I was thinking about.

PR #5092, opened 2016-07-06
(PUP-6475) Use first server_list entry if failover hasn't occured

branan:
Without this change, if failover has not occured, Puppet would ignore
the server_list setting and instead use the legacy server setting.

Now, the agent will use the first entry of server_list if it's
available, and only fall back to the legacy setting if the new setting
is not in use. This will catch any requests made outside of the
"configurer" which is what handles the fallback logic


adrienthebo:
It's dirty, but it's about as good as we're going to get. :+1:

PR #5093, opened 2016-07-07
(PUP-5908) Add MsgPack serialization

thallgren:
This commit adds a serialization reader/writer pair that is based on
MsgPack.

### DIFF
--- a/lib/puppet/pops/serialization/abstract_writer.rb
+++ b/lib/puppet/pops/serialization/abstract_writer.rb
@@ -117,6 +117,11 @@ class AbstractWriter
   end
 
   # @api private
+  def type_registered?(payload_class)
+    @packer.type_registered?(payload_class)
+  end
+
+  # @api private
   def register_types
     # 0x00 - 0x0F are reserved for low-level serialization / tabulation extensions
 
--- a/lib/puppet/pops/serialization/abstract_writer.rb
+++ b/lib/puppet/pops/serialization/abstract_writer.rb
@@ -45,6 +45,10 @@ class AbstractWriter
     @packer.flush
   end
 
+  def supports_binary?
+    false
+  end
+
   # Write a value on the underlying stream
   # @api public
   def write(value)
@@ -163,7 +167,7 @@ class AbstractWriter
     end
 
     register_type(Extension::TYPE_REFERENCE, Types::PTypeReferenceType) do |o|
-      build_payload { |ep| write_tpl_qname(ep, o.type_string) }
+      build_payload { |ep| ep.write(o.type_string) }
     end
 
     register_type(Extension::SYMBOL, Symbol) do |o|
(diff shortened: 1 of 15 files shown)


thallgren:
This changes the Gem requirements (msgpack becomes required). The ramifications of that needs to be ironed out before this PR is merged.

hlindberg:
second commit message is not in imperative form

lib/puppet/pops/serialization/abstract_writer.rb
```diff
@@ -178,12 +191,16 @@ def register_types
...
 
-    register_type(Extension::SENSITIVE, Types::PSensitiveType::Sensitive) do |o|
-      build_payload { |ep| ep.write(o.unwrap) }
-    end
-
-    register_type(Extension::BINARY, Types::PBinaryType::Binary) do |o|
-      build_payload { |ep| ep.write(o.to_s) }
+    if supports_binary?
+      register_type(Extension::BINARY, Types::PBinaryType::Binary) do |o|
+        # The Ruby MessagePack implementation have special treatment for "ASCII-8BIT" strings. They
```
hlindberg: "implementation has"

lib/puppet/pops/serialization/extension.rb
```diff
@@ -26,8 +27,8 @@ module Extension
   TIMESPAN = 0x34
   VERSION = 0x35
   VERSION_RANGE = 0x36
-  SENSITIVE = 0x37
-  BINARY = 0x38
+  BINARY = 0x37
+  BASE64 = 0x38
```
hlindberg: Why do we have a BASE64 serialization type? Isn't that a choice per serializer how to treat a Binary ?

thallgren: It is, and a binary capable serializer (e.g. msgpack) will use BINARY whereas a non-binary serializer (e.g. json) will use BASE64 when writing to a stream. But consider what happens if one version of a serializer isn't capable whereas of writing binary content but another version is. The latter would then need to differentiate between the two when reading. Using two separate extension points (both for the Binary) makes this possible.

melissa:
@thallgren @hlindberg do we still need this now that rich data is a thing? Can this PR be closed, or should we keep it to be reworked?

hlindberg:
IMO, this should be reworked - Pcore over MsgPack is an improvement to both "human readable JSON" and "human readable JSON in MsgPack format" - (both actually being rich data).
The improvements come from Pcore serialization being able to tabulate data and thereby reducing size on the wire as well as memory needed at deserialized end.

melissa:
In that case, should we close this PR? Or would it be best to leave it open?

thallgren:
I would like this PR to be left open.

melissa:
Great! Thank you for the update @hlindberg and @thallgren !!

joshcooper:
> The improvements come from Pcore serialization being able to tabulate data and thereby reducing size on the wire

By default the puppetserver catalog endpoint applys `gzip` compression (true for most/all of the endpoints), where compression is handled in native code. Have you benchmarked "human readable JSON over msgpack and then gzip compressed" compared to "pcore tabulation over msgpack and then gzip compressed", where pcore tabulation occurs in ruby, and I'd assume is much slower?

> as well as memory needed at deserialized end.

Do you have any numbers for this?

Aside, several conflicts need to be resolved.

joshcooper:
@thallgren I noticed this hasn't seen any activity in a long time, do you think you'll be able to work on it in the near future?

joshcooper:
Since this hasn't seen any activity in over a year, I'm going to go ahead and close this pull request for the time being. Please re-open this pull request once the next actions are addressed (conflicts and gem dependencies issue are resolved).

PR #5094, opened 2016-07-07
(PUP-4742) Switch to use semantic_puppet gem

thallgren:
This PR removes the vendored (and embedded in the Puppet code-base) 'semantic' gem in favor of adding a dependency to the 'semantic_puppet' gem.

The PR supersedes [PR-4977](/puppetlabs/puppet/pull/4977) which was merged and then reverted.


hlindberg:
Ping @thallgren - for what I understand this should be good to go as the updated gem is in puppet agent and all builds now contain it.

hlindberg:
Needs a rebase though.

hlindberg:
Reviewed and merged in the past (got reverted in wait for gem to be packaged into all builds).
Here we go again - merging...

PR #5099, opened 2016-07-08
(maint) Bump travis to test 2.1.9 same as p-a ships

kylog:


kylog:
Ping @branan or @whopper or @HAIL9000. This came up because @rnelson0 and I discovered that we weren't running travis against the same version of ruby that we bundle in puppet-agent (see the puppet-lint PR above that references this one). I don't expect the difference to turn up any spec failures, but might as well default to the ruby version we choose to bundle.

branan:
Looks like Travis survived this change. :+1:

PR #5101, opened 2016-07-08
(PUP-6042) Add `no-use_cached_catalog` setting when running `puppet agent --test`

Magisus:
Using the `--test` option when running `puppet agent` should override
use_cached_catalog behavior, because forcing an agent run should build
and apply a new catalog. This commit adds the `no-use_cached_catalog`
parameter to the special settings for the `--test` flag.


whopper:
As discussed in person, there's a test in `spec/unit/application/agent_spec.rb` which should be updated to match this.

PR #5102, opened 2016-07-08
(PUP-6442) Redact resource harness change and audit events

adrienthebo:


adrienthebo:
This pull request demonstrates that redacting sensitive properties plays pretty poorly with auditing parameters - when audited values are redacted users can only see that there was a value that was redacted and nothing more.

kylog:
@adrienthebo I was wondering how this stuff would interact with audit. Given that we have audit marked for deprecation (https://tickets.puppetlabs.com/browse/PUP-893), I'm *tempted* to say we release note that weakness. Otoh, functionality *like* audit (if not the audit implementation) may still be of interest, so that might be a dangerous temptation. Chat with Beth?

adrienthebo:
I'm pretty sure that auditing itself still works for indicating change - it'll just be much less useful for seeing what exactly changed. This sort of thing shouldn't be too bad by itself because we're not leaking information (outside of `state.yaml` which isn't a regression), but combining auditing and redaction will reduce the value of auditing.

On the other hand, who is going to use both?

kylog:
> who is going to use both?

Exactly. I suspect no one would.

adrienthebo:
@kylog did you mean to close this PR?

kylog:
I'm an idiot. Reopening.

adrienthebo:
<3

lib/puppet/property.rb
```diff
@@ -205,6 +206,22 @@ def change_to_s(current_value, newvalue)
...
+  #   introduce issues with unexpected property values.
+  #
+  # @see String#format
+  # @param fmt [String] The format string to interpolate.
+  # @param args [Array<String>] One or more strings to conditionally redact and interpolate into the format string.
+  #
+  # @return [String]
+  def format(fmt, *args)
+    fmt % args.map { |arg| @sensitive ? "[redacted]" : arg.to_s }
+  end
```
branan: I don't really like how much of sorting out "is" and "should" is left to the resource_harness in this... but that's really part of the existing design of the code, and I think this is as good as it's going to get without some major restructuring of the responsibilities of the resource_harness vs properties.

adrienthebo: In general these APIs are pretty horrid. For example, as noted in the doc comment `#is_to_s` returns the current value unmodified, so it's not necessarily a string. To make this more fun `Puppet::Parameter.format_value_for_display` specifically handles non-string values, and `Property#change_to_s` calls that method with the output of `#is_to_s`, so we have functions indicating they return strings passing data to values that specifically handle values other than strings.

It's like Camelot - a silly place that perhaps we shouldn't visit.

branan:
I'm :+1: on this. Kinda gross, but we expected that to happen

lib/puppet/property.rb
```diff
@@ -205,6 +205,22 @@ def change_to_s(current_value, newvalue)
     end
   end
 
+  # Format the given string and conditionally redact the provided interpolation variables, depending on if
```
hlindberg: "Formats the given string and conditionally redacts..."

hlindberg:
:+1: not too familiar with this part of town, but looks ok to me (we are carefully adding stuff, not reimplementing).

lib/puppet/property.rb
adrienthebo: Good catch, fixed.

PR #5103, opened 2016-07-09
(PUP-6483) Prevent `puppet resource user` logins

Iristyle:
- `puppet resource user` will make calls to LogonUserW for each local
   user that exists when enumerating / populating the user resources.

   Unfortunately, it does so with a nil password (given the values are
   not derived from a manifest).  The Windows API, even when a user
   actually has an empty password, will return 1327 for account
   restrictions when trying to call LogonUser.

   This prevents additional calls to LogonUser with the wrong password
   (which could cause accounts to be locked out), and it also adds
   some warnings to Puppet to inform users that when users have blank
   / empty passwords, that Puppet will always set them, and will
   always generate spurious change notices as a result.

### DIFF
--- a/lib/puppet/type/user.rb
+++ b/lib/puppet/type/user.rb
@@ -200,6 +200,10 @@ module Puppet
           well as the password.
         * Windows passwords can only be managed in cleartext, as there is no Windows API
           for setting the password hash.
+        * Windows passwords shouldn't be specified as empty (zero-length) strings, even
+          if you've configured Windows to allow blank passwords. If you do specify a
+          blank password, Puppet will report a change on every run because there is no
+          way to verify that an existing password is blank.
 
         [stdlib]: https://github.com/puppetlabs/puppetlabs-stdlib/
 
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -191,6 +191,8 @@ module Puppet::Util::Windows::ADSI
     end
 
     def self.logon(name, password)
+      # this underlying check will throw when password is nil or empty
+      # as the Windows LogonUserW API does not support that
       Puppet::Util::Windows::User.password_is?(name, password)
     end
 
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -238,6 +238,9 @@ module Puppet::Util::Windows::ADSI
 
     def password=(password)
       if !password.nil?
+        if password == ''
+          Puppet.warning("Blank (zero length) passwords are not recommended. Blank passwords might be allowed on your system based on policy, but Puppet cannot verify a blank password with the operating system. Because of this a password change event will be executed by Puppet on every run.")
+        end
         native_user.SetPassword(password)
         commit
       end
--- a/lib/puppet/util/windows/user.rb
+++ b/lib/puppet/util/windows/user.rb
@@ -52,9 +52,20 @@ module Puppet::Util::Windows::User
   module_function :check_token_membership
 
   def password_is?(name, password)
-    logon_user(name, password) { |token| }
-  rescue Puppet::Util::Windows::Error
-    false
+    # Windows API LogonUserW always disallows this with error 1327:
+    # Logon failure: user account restriction. Possible reasons are blank passwords
+    # not allowed, logon hour restrictions, or a policy restriction has been enforced.
+    if password.nil? || password == ''
+      raise Puppet::Util::Windows::Error.new(
+        "Windows does not have the ability to verify NULL or empty passwords"
+      )
+    end
+
+    begin
+      logon_user(name, password) { |token| }
+    rescue Puppet::Util::Windows::Error
+      return false
+    end
   end
   module_function :password_is?
 
--- a/lib/puppet/provider/user/windows_adsi.rb
+++ b/lib/puppet/provider/user/windows_adsi.rb
@@ -117,6 +117,9 @@ Puppet::Type.type(:user).provide :windows_adsi do
   end
 
   def password
+    # avoid a LogonUserW style password check when the resource is not yet
+    # populated with a password (as is the case with `puppet resource user`)
+    return nil if @resource[:password].nil? || @resource[:password] == ''
     user.password_is?( @resource[:password] ) ? @resource[:password] : nil
   end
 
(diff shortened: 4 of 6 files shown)


Iristyle:
@nfagerlund we discovered while working on this ticket that Windows can never validate blank passwords through the `LogonUser` API.  What this effectively means is that if you try to set a blank password for a `user` resource in a manifest - Puppet will think that it's changing the password on every run and will generate a spurious change notice on every run.

We're going to add some simple docs specifying a warning about that behavior in the `user` type (and we'll ping you again when that's ready).  Is there anywhere else that we should add documentation?

nfagerlund:
@Iristyle Yes! This page could use a note as well: https://docs.puppet.com/puppet/4.5/reference/resources_user_group_windows.html

It lives in the puppet-docs repo, at `source/puppet/<VERSION>/reference/resources_user_group_windows.markdown`.

Iristyle:
I just updated docs - these changes will land in https://docs.puppet.com/puppet/latest/reference/types/user.html

ping @nfagerlund for review

lib/puppet/type/user.rb
```diff
@@ -200,6 +200,12 @@ def sync
           well as the password.
         * Windows passwords can only be managed in cleartext, as there is no Windows API
           for setting the password hash.
+        * Windows password cannot be set to nil, and should not be set to an empty string.
+          While this may technically be allowed, with a non-default local system level
+          password complexity configuration, there is no way to verify that an existing
+          password is blank. Therefore each Puppet run will emit a change notice for any
+          Windows user configured with a blank password, even if the users existing
+          password was already blank.
```
ferventcoder: That was a fun learning opportunity!

lib/puppet/util/windows/user.rb
```diff
@@ -52,9 +52,20 @@ def check_token_membership
...
-  rescue Puppet::Util::Windows::Error
-    false
+    # Windows API LogonUserW always disallows this with error 1327:
+    # Logon failure: user account restriction. Possible reasons are blank passwords
+    # not allowed, logon hour restrictions, or a policy restriction has been enforced.
+    if password.nil? || password == ''
+      raise Puppet::Util::Windows::Error.new(
+        "Windows does not have the ability to verify NULL or empty passwords"
+      )
+    end
```
ferventcoder: The raise here has no repercussions with pulling existing users who may have an empty password, correct?

Iristyle: Correct, and thanks for asking.

One weird thing that I did notice though is that the warning is not showing:

```
bundle exec puppet apply -e "user { 'ramon': managehome => false, password => '' }"
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.

Notice: Compiled catalog for vagrant-2008r2.corp.puppetlabs.net in environment production in 0.06 seconds
Notice: /Stage[main]/Main/User[ramon]/ensure: created
Notice: Applied catalog in 0.22 seconds


C:\source\puppet [ticket/stable/PUP-6843-puppet-resource-user-calls-login-with-null-password-on-Windows +63 ~0 -0 !]> net user

User accounts for \\VAGRANT-2008R2

-------------------------------------------------------------------------------
Administrator            bla                      bob
cyg_server               foo_bar                  Guest
ramon                    sshd                     vagrant
The command completed successfully.
```

Iristyle: Oh right, now I remember why.  New users call into the ADSI create directly, and the provider code path for password isn't touched.

But when you modify an existing user:

```
bundle exec puppet apply -e "user { 'ramon': managehome => false, password => '' }"
Notice: Compiled catalog for vagrant-2008r2.corp.puppetlabs.net in environment production in 0.06 seconds
Warning: Blank (zero length) passwords are not recommended. Blank passwords may be allowed on your system based on policy, but Puppet cannot verify a blank password with the operating system. Because of this a password change event will be executed by Puppet on every run.
Notice: /Stage[main]/Main/User[ramon]/password: changed password
Notice: Applied catalog in 0.22 seconds
```

I think we need to move the location of the warning.

MosesMendoza:
👍

lib/puppet/type/user.rb
nfagerlund: What does "set to nil" mean in the context of a Puppet manifest? `password => undef,`?

nfagerlund: If the latter, I suggest changing to the following: 

```
* Windows passwords can't be set to `undef`.
* Windows passwords shouldn't be specified as empty (zero-length) strings, even
  if you've configured Windows to allow blank passwords. If you do specify a
  blank password, Puppet will report a change on every run because there is no
  way to verify that an existing password is blank. 
```

ferventcoder: @Iristyle do you want to clean this up?

lib/puppet/util/windows/adsi.rb
```diff
@@ -236,6 +238,9 @@ def add_flag(flag_name, value)
 
     def password=(password)
       if !password.nil?
+        if password == ''
+          Puppet.warning("Blank (zero length) passwords are not recommended. Blank passwords may be allowed on your system based on policy, but Puppet cannot verify a blank password with the operating system. Because of this a password change event will be executed by Puppet on every run.")
```
nfagerlund: Style guide: "may" is an ambiguous word. In this case, we should replace it with "might."

lib/puppet/type/user.rb
Iristyle: @nfagerlund thanks for pointing that out.  For manifest users the terminology makes no sense (and thats not what `undef` does).

So, l removed that bit and just focused on the blank passwords.

lib/puppet/util/windows/adsi.rb
Iristyle: Fixed

lib/puppet/util/windows/user.rb
ferventcoder: Move now or later?

Iristyle: Its already moved and the tests were changed as well...

Discovered that there's no way to get a nil in there either (unless you call the code directly from a custom type and provider), so removed the special `nil` message.

ferventcoder:
I'm 👍 , just waiting on final AV/TrCI jobs.

PR #5104, opened 2016-07-09
(PUP-6563) Wrapping version in quotes on windows

chrismatteson:
This allows consistent behavior for both gem provider on both linux and windows.


branan:
The commit mesage needs to be of the form "(TICKET-666)" for the Puppet project.

Additionally, we have a policy for our open-source repos of only referencing tickets that the open-source community can reference. Please move or duplicate this PE ticket to the PUP project and use that ticket number in your PR.

Otherwise, this looks good! Thanks for taking the time to contribute

PR #5105, opened 2016-07-09
(PUP-6486) Add the local lib directory to $LOAD_PATH before loading p…

pdrakeweb:
…uppet version.

Without this patch, previously installed mismatched versions of the puppet gem may
cause a version file from another puppet gem to be improperly loaded.

By adding the local lib directory to $LOAD_PATH before loading the puppet version,
we ensure that the proper version file is loaded.


branan:
:+1: This is much more straightforward than whatever we were trying to do with the gemspec before.

PR #5107, opened 2016-07-11
(WIP)(PUP-6490) Zero sensitive Win32 API strings

Iristyle:
- Needs some tests

/cc @joshcooper

### DIFF
--- a/lib/puppet/util/windows/api_types.rb
+++ b/lib/puppet/util/windows/api_types.rb
@@ -223,8 +223,7 @@ module Puppet::Util::Windows::APITypes
         end
       end
 
-      # TODO: I don't think this is actually defined!
-      def ==(other) Windows.memcmp(other, self, size) == 0 end
+      def ==(other) FFI::WIN32.memcmp(other, self, size) == 0 end
     end
 
     # https://msdn.microsoft.com/en-us/library/windows/desktop/ms724950(v=vs.85).aspx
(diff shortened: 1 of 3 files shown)


Iristyle:
This is a quick spike more than anything...

lib/puppet/util/windows/api_types.rb
```diff
@@ -33,6 +33,22 @@ def self.from_string_to_wide_string(str, &block)
       nil
     end
 
+    def self.from_string_to_secure_wide_string(str, &block)
```
Iristyle: I wonder if this is ultimately a little pointless given copies of the same strings floating around in the Ruby managed heap?

joshcooper:
Yeah, I think that's going to be tough.

joshcooper:
Has conflicts and needs to retarget a different branch. I'm also not familiar enough with ruby memory to know if this has the desired effect.

McdonaldSeanp:
needs to be retargeted at 4.10.x, 5.3.x, or master

jsane:
2 cents.. I don't quite follow the ruby syntax but it appears the intent is to zeroize the memory for the wide version of the strings (as generated by from_string_to_secure_wide_strings) when being freed. And with invocations thru FFI ensures that actually happens - that looks reasonable. What am not entirely sure of, as you noted earlier, is whether the original copy of the string still stays around. Perhaps we can try to stomp out their contents as well given that strings in ruby are mutable?

joshcooper:
I'm going to close this for now.

PR #5108, opened 2016-07-12
(PUP-6295) New corrective_change field

kbarber:
The corrective_change field in events marks an event that was presumed to be a
manual out-of-band change from Puppet, that Puppet then corrected. In the
context of a noop or non-applying change, the same is true except that the
change was not corrected but is intended to be corrected once a real run is
applied.

This patch introduces the work necessary to calculate this change, introducing a
transaction store that stores the last known best applied value so we can
calculate that a change was indeed unexpected.

For each report, we also mark the resource that had a corrective_change plus
mark an overall report of this also so this information can be stored in
backends like PuppetDB for quick searching.

A metric has also been included, to count all resources that had
corrective_change marked true as part of the implicit metric gathering for flags
on resources.

### DIFF
--- a/lib/puppet/resource/status.rb
+++ b/lib/puppet/resource/status.rb
@@ -30,7 +30,7 @@ module Puppet
       attr_accessor :evaluation_time
 
       # Boolean status types set while evaluating `@real_resource`.
-      STATES = [:skipped, :failed, :failed_to_restart, :restarted, :changed, :out_of_sync, :scheduled]
+      STATES = [:skipped, :failed, :failed_to_restart, :restarted, :changed, :out_of_sync, :scheduled, :corrective_change]
       attr_accessor *STATES
 
       # @!attribute [r] source_description
@@ -87,6 +87,14 @@ module Puppet
       #   @return [Array<Puppet::Resource>] A cache of all
       #   dependencies of this resource that failed to apply.
       attr_accessor :failed_dependencies
+
+      # @!attribute [rw] corrective_change
+      #   @return [Boolean] true if the resource contained a corrective change.
+      attr_accessor :corrective_change
+
+      # @!attribute [r] real_resource
+      #   @return [Puppet::Type] returns the real resource that this status relates to
+      attr_reader :real_resource
 
       def dependency_failed?
         failed_dependencies && !failed_dependencies.empty?
@@ -153,6 +161,7 @@ module Puppet
         @out_of_sync = false
         @skipped = false
         @failed = false
+        @corrective_change = false
 
         @file = resource.file
         @line = resource.line
@@ -181,7 +190,7 @@ module Puppet
         @changed = data['changed']
         @skipped = data['skipped']
         @failed = data['failed']
-
+        @corrective_change = data['corrective_change']
         @events = data['events'].map do |event|
           # in YAML (for reports) we serialize this as an object, but
           # in PSON it becomes a hash. Depending on where we came from
@@ -212,6 +221,7 @@ module Puppet
           'change_count' => @change_count,
           'out_of_sync_count' => @out_of_sync_count,
           'events' => @events,
+          'corrective_change' => @corrective_change,
         }
       end
 
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -1,5 +1,6 @@
 require 'puppet'
 require 'puppet/indirector'
+require 'puppet/transaction/persistence'
 
 # This class is used to report what happens on a client.
 # There are two types of data in a report; _Logs_ and _Metrics_.
@@ -122,6 +123,10 @@ class Puppet::Transaction::Report
   #
   attr_reader :noop
 
+  # @return [Boolean] true if the report contains any events and resources that had corrective
+  #    changes.
+  attr_reader :corrective_change
+
   def self.from_data_hash(data)
     obj = self.allocate
     obj.initialize_from_hash(data)
@@ -184,6 +189,7 @@ class Puppet::Transaction::Report
   # @api private
   def finalize_report
     prune_internal_data
+    process_resource_inferences
 
     resource_metrics = add_metric(:resources, calculate_resource_metrics)
     add_metric(:time, calculate_time_metrics)
@@ -215,6 +221,7 @@ class Puppet::Transaction::Report
     @status = 'failed' # assume failed until the report is finalized
     @noop = Puppet[:noop]
     @noop_pending = false
+    @corrective_change = false
   end
 
   # @api private
@@ -229,6 +236,7 @@ class Puppet::Transaction::Report
     @noop_pending = data['noop_pending']
     @host = data['host']
     @time = data['time']
+    @corrective_change = data['corrective_change']
 
     if master_used = data['master_used']
       @master_used = master_used
@@ -288,10 +296,10 @@ class Puppet::Transaction::Report
       'noop_pending' => @noop_pending,
       'environment' => @environment,
       'master_used' => @master_used,
-
       'logs' => @logs,
       'metrics' => @metrics,
       'resource_statuses' => @resource_statuses,
+      'corrective_change' => @corrective_change,
     }
   end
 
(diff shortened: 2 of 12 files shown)


lib/puppet/transaction/report.rb
```diff
@@ -386,6 +394,89 @@ def self.default_format
...
+    new_resources = {}
+
+    resource_statuses.each do |name, status|
+      # Initialize data structure for resource
+      new_resources[name] ||= {}
+      params = new_resources[name]["parameters"] ||= {}
+
+      # Populate new hash with events mapped to params
+      param_to_event = {}
+      status.events.each do |ev|
```
adrienthebo: My implementation of event redaction does so as events are generated, so I think that my changes will strip out the information you're using before you can look at it. We'll have to sync up our two approaches.

lib/puppet/transaction/report.rb
```diff
@@ -386,6 +394,89 @@ def self.default_format
 
   private
 
+  # Using the last applied values from our persistence store, and the current
+  # report data, iterate across all events finding any corrective_change that
+  # might exist and marking the event, resource and report appropriately.
+  def calculate_corrective_change
+    persistence = Puppet::Transaction::Persistence.new
```
adrienthebo: This function handles loading persisted data, calculating corrective change, and then saving the persisted data again - I would rather we have the actual data processing logic in a separate function and another function that handles the persistence loading/saving and invoking the data processing logic.

lib/puppet/transaction/report.rb
kbarber: I believe I've solved this by skipping sensitive events in the last patch.

lib/puppet/transaction/report.rb
kbarber: Take another look, I've refactored somewhat.

kbarber:
@adrienthebo @branan @kylog ... this one is ready for review again btw. Thanks.

lib/puppet/transaction/report.rb
```diff
@@ -386,6 +394,101 @@ def self.default_format
...
+      # For non events, we just want to store the parameters agent value.
+      property.value
+    end
+  end
+
+  # @param [Puppet::Transaction::Event] event event to use for processing
+  # @param [Puppet::Property] property correlating property
+  # @param [Object] system_value system_value from last transaction
+  # @return [bool] system_value to persist
+  def is_corrective_change(property, event, system_value)
```
adrienthebo: This seems to be calculating if a given event is due to corrective change; if that's the case why is this implemented on reports? This is of specific interest because calculating this here requires that `Event#corrective_change` is an `attr_accessor`; I don't love the fact that what seems like it's an intrinsic property of an event is writable outside of the object.

lib/puppet/transaction/report.rb
```diff
@@ -386,6 +394,101 @@ def self.default_format
...
+    persistence.save
+  end
+
+  # Given an event and its property, calculate the system_value to persist
+  # for future calculations.
+  # @param [Puppet::Transaction::Event] event event to use for processing
+  # @param [Puppet::Property] property correlating property
+  # @param [Object] system_value system_value from last transaction
+  # @return [Object] system_value to be used for next transaction
+  def new_system_value(property, event, system_value)
```
adrienthebo: Same comment as `#is_corrective_change`, if this doesn't use any information that's intrinsically part of the report then I think it should be defined on `Event`.

lib/puppet/resource/status.rb
```diff
@@ -88,6 +88,14 @@ class Status
       #   dependencies of this resource that failed to apply.
       attr_accessor :failed_dependencies
 
+      # @!attribute [rw] corrective_change
+      #   @return [Boolean] true if the resource contained a corrective change.
+      attr_accessor :corrective_change
```
adrienthebo: Per the other comments, I don't think that this should be writable since it seems like the event itself should be calculating this.

lib/puppet/transaction/report.rb
```diff
@@ -386,6 +394,101 @@ def self.default_format
 
   private
 
+  # Using the last applied values from our persistence store, and the current
+  # report data, iterate across all events finding any corrective_change that
+  # might exist and marking the event, resource and report appropriately.
+  def process_resource_inferences
+    persistence = Puppet::Transaction::Persistence.new
```
adrienthebo: I still don't like that we're tying together the loading/saving of persistence and doing the corrective change calculations in the same function; I would rather have them decoupled to separate the responsibilities and make it easier to test things.

lib/puppet/transaction/report.rb
```diff
@@ -386,6 +394,101 @@ def self.default_format
...
+
+      # Populate new hash with events mapped to params
+      param_to_event = {}
+      status.events.each do |ev|
+        param_to_event[ev.property] = ev
+      end
+
+      status.real_resource.parameters.each do |pname,param|
+        # Skip if if the data is sensitive, or if its just a property
+        next if !param.is_a?(Puppet::Property) || param.sensitive
```
adrienthebo: Is there a reason we're not using `#select` to pick the non-sensitive properties and then iterate over those instead of iterating over all and selectively skipping the loops? `next` is kind of a funny thing and I would rather we avoid using it.

lib/puppet/transaction/report.rb
kbarber: So last weekend and earlier this week I attempted to do the calculation for the new system_value in ResourceHarness and the corrective_change calculation in Event. I must admit, it became a complex fairly quickly.

Let me elaborate ... (it will help me think, bear with me :-)

* I need to provide a path back to the Persistence object that we create, this means that I need to have paths from Event & ResourceHarness back to the Transaction to obtain the object reference. There are other links that are missing that are required for this also. Not impossible, but requires work.
* The Persistence mechanism needs to be changed, so that old values and new values for resources are accessible separately. The calculation for the new system_value will need to store the value, while in another place we still need to retrieve the old value for corrective_change calculation. Even with correct ordering (which can be tricky) whereby we replace old with new in the same pass, this leaves us with dangling/orphaned data at the end, as we aren't doing a full flush of old data, and we're back to the state.yaml problem. Without something like this, we're also back to forcing the storing/calculation in one inherit place. I don't think this is impossible, its just not a pattern I'm comfortable with, but perhaps there is another approach, or I can learn to like it.
* Parts of the inner loop (ResourceHarness for example) work in ways to solve the problem they were originally assigned. Trying to follow the same paths as I have on the outer loop (that is, the output) isn't as clear and I found myself with a harder state to debug. This is largely because the code never has had a demand to do what I'm doing.

Mostly I'm afraid this might mean some structural changes or scary changes. At least thats where my patch ended up, I wasn't impressed with my attempt anyway, and more so I didn't get it working. To be straight, this implementation was easier.

Having said that, I feel you are right, but not for the reasons you have listed per se, I think more because if the core of Puppet isn't aligned for this use-case then it possible should be. I think this would aid future ideas around this kind of calculation (and my future interest in detecting "unexpected changes" without an event as opposed to corrective_change and also - detecting idempotency issues with resources automatically).

So I'm going to give it another go, and see what mess I end up with. You're a bastard btw. Just sayin' :-).

lib/puppet/transaction/report.rb
kbarber: Well, in this particular case it is actually required for all non-sensitive properties, not just ones that generate an event. Event might not be the place, is my point, probably ResourceHarness or Property.

lib/puppet/transaction/report.rb
kbarber: Matter of taste I'd say. I don't really care either way.

Why is next funny? I get more amused by redo.

scottyw:
I just merged this by accident. Sorry everyone. Reverted here: https://github.com/puppetlabs/puppet/pull/5138

I'm going to follow up in an email.

lib/puppet/transaction/report.rb
kbarber: New PR has these recommended changes: https://github.com/puppetlabs/puppet/pull/5139

kbarber:
New PR is #5139

PR #5109, opened 2016-07-12
(W.I.P) Explore resource type in pcore

hlindberg:
This PR contains the basic work to support Resource Type Implementations in Pcore and instead of loading the implementation via `Puppet::Type`, this implementation will load a mytype.pp file under <root>/resource_types/mytype.pp if such exists.

The work consists mostly of changes that ensure that the compiler's code paths to loading resource type implementations and definitions follow one and only one path (at the end).

This work is not fully functional as the puppet resource type implementation instantiator cannot yet load Pcore.

What remains:

* Make the PuppetResourceTypeImplInstantiator returns an actual instance of Puppet::Pops::Resource::Resource::ResourceTypeImpl
* Add support so that Puppet::Type is called as part of 4.x loading instead of from the new Puppet::Pops::Evaluator::Runtime3ResourceSupport as this has the potential of short circuiting out the searches for pcore based types when the 3x ruby implementation should be used. Obviously such loading should be tried last after having searched all modules for a pcore implementation. The idea is that a module only containing pcore could be placed earlier on the module path to get pcore loading instead of ruby loading. However, once a Puppet::Type has been loaded, this should (as quickly as possible) result in a find. This may be as easy to implement as just a call to Puppet::Type, after a search for a resource type impl results in nil, the result of Puppet::Type is then added to the loaded structure (as a cache instead of a cache 'not found'.
* Remove the Mocking implementation that does things as a mock of "Notify" (useful as a hack while developing).
* Write tests
** An important test is that a puppet compilation never loads a resource type impl in ruby when a corresponding pcore exists.
** The second important test is that puppet apply gets an actual Puppet::Type based implementation when applying the catalog.
** It is suggested that a simple "notify like" resource is used (ruby impl, and with a pcore description).


hlindberg:
I am dealing with the merge conflict atm.

peterhuene:
WRT the load path, I'd prefer `types`, which we already have, over a `resource_types` directory.  I'm thinking ahead to when we properly support something like this:

```Puppet
# Resource type Foo
type foo {
  attr String bar;
  attr Numeric baz;
  ...
}
```

I don't think the `types` directory should just be for aliases.  It should be anything we consider a user type: alias or resource type.

Given how users should not be able to declare an alias or a resource type that conflicts with existing aliases, existing resource types, or built-in types, I don't foresee an issue using the `types` directory for pcore definitions.

Thoughts?

lib/puppet/parser/scope.rb
```diff
@@ -18,7 +18,6 @@ class Puppet::Parser::Scope
   extend Forwardable
   include Puppet::Util::MethodHelper
 
-  include Puppet::Resource::TypeCollectionHelper
```
thallgren: The `Puppet::Resource::TypeCollectionHelper` is also included in `Puppet::Parser::Resource` and `Puppet::Pops::Parser`. Suggest removing those includes too, and then also remove the actual module.

lib/puppet/pops/puppet_stack.rb
```diff
@@ -0,0 +1,43 @@
...
+      Kernel.eval("obj.send(message, *args, &block)", Kernel.binding(), file, line)
+    else
+      Kernel.eval("obj.send(message, *args)", Kernel.binding(), file, line)
+    end
+  end
+
+  def self.stacktrace
+    result = caller().reduce([]) do |memo, loc|
+      if loc =~ /^(.*\.pp)?:([0-9]+):in `stack'/
+        memo << [$1.nil? ? 'unknown' : $1, $2.to_i]
```
thallgren: Why is `nil` replaced with the string 'unknown' here but with an empty string on line 24?

spec/unit/pops/puppet_stack_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    expect(Puppet::Pops::PuppetStack.stack("test.pp", 1, StackTraceTest.new, :gets_block, ['got_it']) {|x| x }).to eql('got_it')
+  end
+
+  it 'allows passing multiple variables and a block' do
+    expect(
+      Puppet::Pops::PuppetStack.stack("test.pp", 1, StackTraceTest.new, :gets_args_and_block, ['got_it', 'again']) {|x, y| [x,y].join(' ')}
+      ).to eql('got_it again')
+  end
+
+end
```
thallgren: missing EOL

lib/puppet/pops/puppet_stack.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    if block_given?
+      Kernel.eval("obj.send(message, *args, &block)", Kernel.binding(), file, line)
+    else
+      Kernel.eval("obj.send(message, *args)", Kernel.binding(), file, line)
+    end
+  end
+
+  def self.stacktrace
+    result = caller().reduce([]) do |memo, loc|
```
thallgren: Remove parenthesis around empty argument list.

lib/puppet/pops/puppet_stack.rb
```diff
@@ -0,0 +1,43 @@
...
+module PuppetStack
+  # Sends a message to an obj such that it appears to come from
+  # file, line when calling stacktrace.
+  #
+  def self.stack(file, line, obj, message, args, &block)
+    file = '' if file.nil?
+    line = 0 if line.nil?
+
+    if block_given?
+      Kernel.eval("obj.send(message, *args, &block)", Kernel.binding(), file, line)
```
thallgren: Remove parenthesis around empty argument list.

lib/puppet/pops/puppet_stack.rb
```diff
@@ -0,0 +1,43 @@
...
+    else
+      Kernel.eval("obj.send(message, *args)", Kernel.binding(), file, line)
```
thallgren: Remove parenthesis around empty argument list.

lib/puppet/pops/puppet_stack.rb
thallgren: Suggest that `"obj.send(message, *args, &block)"` is a single quoted frozen constant.

lib/puppet/pops/puppet_stack.rb
hlindberg: `nil` cannot be used as the name of a file IIRC, and I did not want to use the name `unknown` as a "special name" when setting it, now it is only turned into the string 'unknown' when presenting it.

hlindberg:
@peterhuene I did consider using the `type` directory, and I guess we could - What is required is that the loaded resource type implementation makes two entries in the loader - one for the type reference, and one for the implementation. The implementation should never be returned to the Puppet Language - for example the resource type `Foo` should resolve to an instance of `PResourceType` with `"foo"` as the type name. In the evaluator we do need to get the actual implementation (as a Ruby impl, or as a Pcore impl) so we can assert the correctness of attributes). This means we need to ask for a different thing from the loaders.
Thus - when asking for a `:type` or a `:resource_type` we should return one or the other, but in both cases load both things.

That may be a better implementation.

hlindberg:
To continue on the same topic (loading from `/types`). There are some additional difficulties with name space as modules are not allowed to introduce names in the global namespace thus `/types/foo` in `mymodule` maps to `Mymodule::Foo`. This does not work for resource type implementations. If we make it so that modules can define types under `/types` that are global, then all other definitions have to be under a `mymodule` directory. This also opens up for the same kind of anarchy we already have wrt names.

This problem together with the difference between a resource reference and the implementation led me to use a separate directory.
The loading is also subject to the user process discussion  - if artefacts are checked into modules or if it is a pure runtime concern at the env level.

thallgren:
I'm concerned that using the name `/types` might make the confusion regarding what that name actually means worse. Puppet already use `lib/type` for resource types defined in Ruby. Now we have a `types` directory where we can declare _puppet_ types and aliases, but apparently not _resource_ types. As a novice I would assume that a `/type` and a `/types` directory would be somehow related. Instead the proposed `/resource_types` becomes the equivalent of `lib/type`.

If only _puppet_ types can reside in `/types`, then in order to make the distinction clear and explicit, perhaps we should revisit the naming of `/types` and instead use `/puppet_types` and `/resource_types`  (or use some alternate naming scheme altogether)?

hlindberg:
Perhaps use `.../lib/puppet/type/typename.pp` - then the pcore representation is next to the ruby impl (although a bit odd).

PR #5110, opened 2016-07-12
(PUP-6491) fix parser_validate acceptance test for puppetserver

er0ck:
puppet acceptance tests run from the root of puppetserver in a submodule
which is at `puppetserver/ruby/puppet`.  In order to use the fixutres in
this test (or any tests) we should use a path relative to the test file
itself, not where beaker thinks itself to be.

[skip ci]


rlinehan:
👍

kurtwall:
skip ci?

kurtwall:
🙈

PR #5112, opened 2016-07-12
(maint) change noop_pending in transaction/report.rb to attr_accessor

wkalt:
This will make it easier to spec test from PuppetDB.

### DIFF
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -71,6 +71,11 @@ class Puppet::Transaction::Report
   # @return [String] the environment name
   attr_accessor :environment
 
+  # Whether there are changes that we decided not to apply because of noop
+  # @return [Boolean]
+  #
+  attr_accessor :noop_pending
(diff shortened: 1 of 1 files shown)


branan:
`noop_pending` is intended to be used/accessed the same way as `status` - it's computed in `finalize` based on the individual resources status information. It shouldn't be any harder to test than `status` which is also an `attr_reader`. I'm not sure why it needs to be directly settable when `status` doesn't.

wkalt:
@branan having it settable directly lets us to test our logic around handling versions of puppet where the parameter is and is not defined. We don't have this issue with status since status is present in all versions we support.

wkalt:
@branan bump

branan:
This PR definitely breaks encapsulation, but the "correct" API of adding events to the report and finalizing it is a lot of complexity to PuppetDB's tests, for very little safety gain in any real-world use of the report API. I'm going to go ahead and merge this.

PR #5113, opened 2016-07-13
(PUP-6499) Ensure Windows ADSI strings UTF-8

Iristyle:
- When retrieving the names of users / groups and their attributes
   through WMI, Ruby will unfortunately convert any returned strings
   into the local codepage (Default.external_encoding) instead of
   converting the values from UTF-16LE to UTF-8 for internal
   consumption.

   This then causes problems with Puppets logging when attempting to
   combine strings that are in UTF-8 and the local encoding.  This was
   observed on French Windows where the default encoding is CP850.

   Without these changes `puppet resource user` will fail to run with
   the error:

Error: Could not run: incompatible character encodings: UTF-8 and CP850

### DIFF
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -334,6 +334,10 @@ module Puppet::Util::Windows::ADSI
       user_name
     end
 
+    def self.current_user_sid
+      Puppet::Util::Windows::SID.name_to_sid_object(current_user_name)
+    end
+
     def self.exists?(name_or_sid)
       well_known = false
       if (sid = Puppet::Util::Windows::SID.name_to_sid_object(name_or_sid))
--- a/lib/puppet/util/windows/adsi.rb
+++ b/lib/puppet/util/windows/adsi.rb
@@ -194,7 +194,10 @@ module Puppet::Util::Windows::ADSI
     end
 
     def [](attribute)
-      native_user.Get(attribute)
+      value = native_user.Get(attribute)
+      # Rubys WIN32OLE errantly converts UTF-16 values to Encoding.default_external
+      return value.encode(Encoding::UTF_8) if value.is_a?(String)
+      value
     end
 
     def []=(attribute, value)
@@ -244,7 +247,8 @@ module Puppet::Util::Windows::ADSI
       # https://msdn.microsoft.com/en-us/library/aa746342.aspx
       # WIN32OLE objects aren't enumerable, so no map
       groups = []
-      native_user.Groups.each {|g| groups << g.Name} rescue nil
+      # Rubys WIN32OLE errantly converts UTF-16 values to Encoding.default_external
+      native_user.Groups.each {|g| groups << g.Name.encode(Encoding::UTF_8)} rescue nil
       groups
     end
 
@@ -363,7 +367,8 @@ module Puppet::Util::Windows::ADSI
 
       users = []
       wql.each do |u|
-        users << new(u.name)
+        # Rubys WIN32OLE errantly converts UTF-16 values to Encoding.default_external
+        users << new(u.name.encode(Encoding::UTF_8))
       end
 
       users.each(&block)
@@ -455,7 +460,8 @@ module Puppet::Util::Windows::ADSI
     def members
       # WIN32OLE objects aren't enumerable, so no map
       members = []
-      native_group.Members.each {|m| members << m.Name}
+      # Rubys WIN32OLE errantly converts UTF-16 values to Encoding.default_external
+      native_group.Members.each {|m| members << m.Name.encode(Encoding::UTF_8)}
       members
     end
 
@@ -524,7 +530,8 @@ module Puppet::Util::Windows::ADSI
 
       groups = []
       wql.each do |g|
-        groups << new(g.name)
+        # Rubys WIN32OLE errantly converts UTF-16 values to Encoding.default_external
+        groups << new(g.name.encode(Encoding::UTF_8))
       end
 
       groups.each(&block)
--- a/spec/integration/util/windows/adsi_spec.rb
+++ b/spec/integration/util/windows/adsi_spec.rb
@@ -1,0 +1,87 @@
+#!/usr/bin/env ruby
+require 'spec_helper'
+require 'puppet/util/windows'
+
+describe Puppet::Util::Windows::ADSI::User,
+  :if => Puppet.features.microsoft_windows? do
+
+  describe '.each' do
+    it 'should return a list of users with UTF-8 names' do
+      begin
+        original_codepage = Encoding.default_external
+        Encoding.default_external = Encoding::CP850 # Western Europe
+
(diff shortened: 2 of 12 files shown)


lib/puppet/parameter.rb
```diff
@@ -543,7 +543,7 @@ def to_s
   def self.format_value_for_display(value)
     if value.is_a? Array
       formatted_values = value.collect {|v| format_value_for_display(v)}.join(', ')
-      "[#{formatted_values}]"
+      "[#{formatted_values}]".encode(Encoding.default_external)
```
Iristyle: Changing parameter formatting here seems wrong... need to better understand the strings being passed through here during `puppet resource user` as some are `UTF-8`, while others are `CP850` and this code crashes with encoding issues.

This is probably the wrong spot to address this, and there's likely an issue with how strings propagate from the ADSI user provider.

But this hammer works at the moment - more investigation needed!

lib/puppet/util/log/destinations.rb
```diff
@@ -164,6 +164,7 @@ def handle(msg)
     }
 
     str = msg.respond_to?(:multiline) ? msg.multiline : msg.to_s
+    str = str.encode(Encoding.default_external)
```
Iristyle: This is caused by debug output generated at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/user/windows_adsi.rb#L61 - it's trying to smash together strings of different encodings.

It's possible that Ruby's COM support is handing back strings that have been re-encoded from `UTF-16LE` to the local codepage, and we're later trying to smash them together unsuccessfully with `UTF-8`.

Note that any time our internal COM helpers read memory for strings, we treat it as `UTF-16LE` (i.e. for task scheduler - https://github.com/puppetlabs/puppet/blob/6b318af97f4254fd4b4f04583b9b677f8f396176/lib/puppet/util/windows/api_types.rb#L56-L81), but Ruby has a bad habit of wanting to convert API results into local codepage all the time.

lib/puppet/util/windows/adsi.rb
```diff
@@ -234,7 +235,7 @@ def groups
       # https://msdn.microsoft.com/en-us/library/aa746342.aspx
       # WIN32OLE objects aren't enumerable, so no map
       groups = []
-      native_user.Groups.each {|g| groups << g.Name} rescue nil
+      native_user.Groups.each {|g| groups << g.Name.encode(Encoding::UTF_8)} rescue nil
```
ferventcoder: I take it we don't have access to Name here?

lib/puppet/util/windows/adsi.rb
```diff
@@ -353,7 +354,7 @@ def self.each(&block)
 
       users = []
       wql.each do |u|
-        users << new(u.name)
+        users << new(u.name.encode(Encoding::UTF_8))
```
ferventcoder: Should "name" be capitalized to signify the same as `group.Name`? Perhaps we can provide an override for name and return it as a string. Seems like that may be quite a bit of work though.

Iristyle: Not sure what you mean.

lib/puppet/util/windows/adsi.rb
Iristyle: Not sure what you mean.

spec/integration/util/windows/principal_spec.rb
```diff
@@ -103,31 +116,45 @@
...
-
-      # Windows API LookupAccountSid behaves differently if current user is SYSTEM
-      if Puppet::Util::Windows::ADSI::User.current_user_name != 'SYSTEM'
-        account_type = :SidTypeWellKnownGroup
-      else
-        account_type = :SidTypeUser
-      end
-
-      expect(principal.account_type).to eq(account_type)
+    it "should create an instance from a user SID" do
```
Iristyle: This test is net new since we didn't have any checks on `SidTypeUser`.

Unfortunately to do something that should work everywhere, this requires some "SID math" on the byte array.  Thoughts?

lib/puppet/util/windows/adsi.rb
```diff
@@ -320,6 +322,12 @@ def self.current_user_name
       user_name
     end
 
+    def self.current_user_sid
+      # TODO: another method of doing this is calling GetTokenInformation
```
Iristyle: oh shoot.. looks like I need to yank this `TODO`

MosesMendoza:
Spec tests pass on German Windows in combination with https://github.com/puppetlabs/puppet/pull/5124:

https://gist.github.com/MosesMendoza/81f382837609bcdaa05b271e308a4d23

Iristyle:
I pulled commit https://github.com/puppetlabs/puppet/pull/5113/commits/e28c4498313e2c9b9e9b81dec573164bc1235bba off and dropped it into its own maint PR given the issues are related, but independent - see #5131 for the new PR

Iristyle:
#5124 should be merged prior to this

Iristyle:
@puppetlabs/windows repushed this one - and should be ready for merge, pending validation on German Windows.

MosesMendoza:
I'm testing this

MosesMendoza:
confirmed spec tests pass, and `puppet resource user` and `puppet resource group` both work.

original failure repro (puppet/stable):
```
PS C:\Users\moses\development\puppet> be puppet resource user
Error: Could not run: incompatible character encodings: CP850 and UTF-8
```
success:
https://gist.github.com/MosesMendoza/1f1d30ebdf264cc9d037b28a8dd5c783

MosesMendoza:
👍  from me

spec/integration/util/windows/principal_spec.rb
glennsarti: Given that we found that the -1001 user was a new user called Administrator, and the -500 is still valid, should we change this -500 or just use Guest (-501) ?

spec/integration/util/windows/adsi_spec.rb
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env ruby
+require 'spec_helper'
+require 'puppet/util/windows'
+
+def fail_if_utf8_codepage
+  if Puppet.features.microsoft_windows? && Encoding.default_external == Encoding::UTF_8
+    raise 'This test must be run in a codepage other than 65001 to validate behavior'
```
glennsarti: Not sure if this should be a `raise` e.g. Later testing on Nano server uses 65001 as the default codepage.

Iristyle: Yeah, we have a few other tests that do the same thing - and perhaps should be rewritten.

The problem is that we can't validate the "fixed" behavior unless we're in another codepage.

glennsarti: Can we optionally mark them as pending?

Iristyle:
#5419 contains some improvements to `WIN32OLE` behavior with respect to handling UTF-8

PR #5115, opened 2016-07-13
(PUP-6427) Prevent AppVeyor password length fails

Iristyle:
- When temp users are created for the sake of tests, they are not
   assigned a password, and this causes the ADSI types to create a blank
   password for the given user.  At certain times within AppVeyor,
   machines are being used to run tests that have a minimum password
   policy in place.  This causes test failures for not satisfying the
   systems password complexity policy.

   Generate random passwords in these instances, ensuring that the
   string contains 3 of 5 character classes - upper case, numeric and
   additional non-alphanumeric character.

Paired-with: Glenn Sarti <glenn.sarti@puppet.com>
             Daniel Lu   <daniel.lu@puppet.com>

### DIFF
--- a/spec/integration/util/windows/security_spec.rb
+++ b/spec/integration/util/windows/security_spec.rb
@@ -476,6 +476,7 @@ describe "Puppet::Util::Windows::Security", :if => Puppet.features.microsoft_win
           it "should retrieve the user sid" do
             sid = nil
             user = Puppet::Util::Windows::ADSI::User.create("puppet#{rand(10000)}")
+            user.password = 'PUPPET_RULeZ_123!'
             user.commit
             begin
               sid = Puppet::Util::Windows::ADSI::User.new(user.name).sid.sid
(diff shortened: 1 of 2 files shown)


spec/integration/util/windows/security_spec.rb
```diff
@@ -476,6 +476,7 @@ def grant_everyone_full_access(path)
           it "should retrieve the user sid" do
             sid = nil
             user = Puppet::Util::Windows::ADSI::User.create("puppet#{rand(10000)}")
+            user.password = 'PUP123!'
```
ferventcoder: Maybe you should go to 8 characters? ;)

Iristyle: (facepalm)

Originally I was using random GUIDs, but then modified to make sure that we always have the 3 of 5 categories covered and just removed the GUID... totally spaced on length!

glennsarti:
👍  for merge

ferventcoder:
👍

PR #5117, opened 2016-07-13
(PUP-6335) Add note explaining duplication to oids.rb

haus:
The information captured in PUPPET_OIDS in oids.rb is duplicated in the
puppetserver codebase. Creating a shared oids file isn't easily
achieveable today, so instead this commit adds a comment to the oids.rb
file explaining that any changes to the oids need to be replicated in
the puppetserver codebase.


haus:
See also: https://github.com/puppetlabs/puppetserver/pull/1140

PR #3043, opened 2014-09-03
(PUP-3116) Optimize the read of the system mounts in selinux

trevor-vaughan:
Util::SELinux currently reads /proc/mounts for each file in the system.

This can cause a massive lag/burden on the system especially if the
system uses a great number of mount entries (such as OpenShift).

However, this is not a great patch since it creates a global variable
insted of using whatever global caching mechanism that Puppet may have
internally.

Submitting for review but hoping for improvement.

PR #5119, opened 2016-07-13
(PUP-6502) Use some memoization to stop calling dscl pathologically

ccaviness:
We get all the data we need from the initial `dscl -readall /Users`; there's no need to specifically go through each user to ask for `ShadowHashData` later on.

Define post_resource_eval in lib/puppet/provider/nameservice/directoryservice.rb to reset @all_present as a poor man's cache invalidation.

If a user record contains both SALTED-SHA512-PBKDF2 and SALTED-SHA512 data, use SALTED-SHA512-PBKDF2, not the other way around.

### DIFF
--- a/lib/puppet/provider/user/directoryservice.rb
+++ b/lib/puppet/provider/user/directoryservice.rb
@@ -121,7 +121,7 @@ Puppet::Type.type(:user).provide :directoryservice do
     end
     attribute_hash[:ensure]         = :present
     attribute_hash[:provider]       = :directoryservice
-    attribute_hash[:shadowhashdata] = get_attribute_from_dscl('Users', attribute_hash[:name], 'ShadowHashData')
+    attribute_hash[:shadowhashdata] = input_hash['dsAttrTypeNative:ShadowHashData']
 
     ##############
     # Get Groups #
@@ -145,12 +145,12 @@ Puppet::Type.type(:user).provide :directoryservice do
       attribute_hash[:password] = '*'
     else
       embedded_binary_plist = get_embedded_binary_plist(attribute_hash[:shadowhashdata])
-      if embedded_binary_plist['SALTED-SHA512']
-        attribute_hash[:password] = get_salted_sha512(embedded_binary_plist)
-      else
+      if embedded_binary_plist['SALTED-SHA512-PBKDF2']
         attribute_hash[:password]   = get_salted_sha512_pbkdf2('entropy', embedded_binary_plist)
         attribute_hash[:salt]       = get_salted_sha512_pbkdf2('salt', embedded_binary_plist)
         attribute_hash[:iterations] = get_salted_sha512_pbkdf2('iterations', embedded_binary_plist)
+      elsif embedded_binary_plist['SALTED-SHA512']
+        attribute_hash[:password] = get_salted_sha512(embedded_binary_plist)
       end
     end
 
@@ -178,7 +178,7 @@ Puppet::Type.type(:user).provide :directoryservice do
   # plist library doesn't read binary plists, so we need to
   # extract the binary plist, convert it to XML, and return it.
   def self.get_embedded_binary_plist(shadow_hash_data)
-    embedded_binary_plist = Array(shadow_hash_data['dsAttrTypeNative:ShadowHashData'][0].delete(' ')).pack('H*')
+    embedded_binary_plist = Array(shadow_hash_data[0].delete(' ')).pack('H*')
     convert_binary_to_hash(embedded_binary_plist)
   end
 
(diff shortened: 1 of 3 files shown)


ccaviness:
N.B.: We run puppet in standalone (masterless) mode, using `puppet apply`, not `puppet agent`; I'm not sure if/how this will affect things in agent mode.

kylog:
@ccaviness thanks for the PR! This looks like a nice performance improvement.

On a quick look, I think the `apply` vs `agent` distinction matters a lot here because in the latter case we want some way to invalidate the cache (drop the memoization). In that regard, this PR is similar to #3043, which was predicated on somewhere to store state for the duration of a catalog application: see https://tickets.puppetlabs.com/browse/PUP-4002, which sadly no one has gotten to tackling.

ccaviness:
Heh - well, I barely understand what I changed in this PR, much less the all puppet internals being discussed in #3043. Does this mean this PR's going to hang around in limbo, too?

I'm not using Puppet Inc. provided packages, so I can pretty easily just patch this in for my own use, but any other macOS users out there would miss out.

lib/puppet/provider/user/directoryservice.rb
```diff
@@ -145,12 +145,12 @@ def self.generate_attribute_hash(input_hash)
       attribute_hash[:password] = '*'
     else
       embedded_binary_plist = get_embedded_binary_plist(attribute_hash[:shadowhashdata])
-      if embedded_binary_plist['SALTED-SHA512']
-        attribute_hash[:password] = get_salted_sha512(embedded_binary_plist)
-      else
+      if embedded_binary_plist['SALTED-SHA512-PBKDF2']
```
adrienthebo: It seems like this diff here just reorders the conditionals, why was this done?

adrienthebo:
We don't have a generalized mechanism for provider cache management, but we have http://www.rubydoc.info/gems/puppet/Puppet/Provider#post_resource_eval-class_method that will be run for each suitable provider at the end of each transaction. For cache invalidation all we need is something like this in lib/puppet/provider/nameservice/directoryservice.rb:

```ruby
def self.post_resource_eval
  @all_present = nil
end
```

That could work until we get around to implementing a proper cache.

lib/puppet/provider/user/directoryservice.rb
ccaviness: On my 10.11.x machine, my user record's ShadowHashData contains both SALTED-SHA512-PBKDF2 and SALTED-SHA512 hash data; SALTED-SHA512-PBKDF2 is more secure, and preferred, so if it exists, return it, then fall through to SALTED-SHA512.

Actually, to expand a bit, the ShadowHashData had these: ["CRAM-MD5", "SALTED-SHA512-PBKDF2", "NT", "SRP-RFC5054-4096-SHA512-PBKDF2", "RECOVERABLE"]`

I expect puppet should eventually start supporting  "SRP-RFC5054-4096-SHA512-PBKDF2"

ccaviness:
@adrienthebo Ok, I've added the `post_resource_eval` call, and I've also updated the pull description and spec tests to reflect the reasoning for the re-ordering of the conditionals.

lib/puppet/provider/user/directoryservice.rb
adrienthebo: Okay, that makes sense. In general I prefer things like this to be broken out into different pull requests so we can incrementally get separate changes in instead of having to okay all changes in a single PR. That being said I'm not too worried that, just an FYI here.

adrienthebo:
This generally looks good to me. I'm still getting up to speed on things again so I'll need to run this PR by someone else, but we should be able to do this fairly soon.

branan:
It looks like removing some of the stubs from the unit tests is causing these tests to fail on Windows (where we don't install the CFPropertyList gem). If that gem is ruby-only we may be able to change our Gemfile so that it installs that gem for development on all platforms.

It's weird that we currently grab that gem on Linux, but I imagine that was done for exactly this reason.

branan:
@ccaviness Trying to sort out the cfpropertylist gem is something we might have the bandwidth to explore, but we'd of course appreciate it if you have the time to poke at it a little bit

ccaviness:
@branan Yes, CFPropertyList is pure ruby.

What are you asking me to sort out?

branan:
@ccaviness "Why doesn't cfpropertylist get installed on Windows, but it does on Linux". I more-or-less just figured it out now while trying to give you a useful write up of where to get started, though. I am going to go ahead and open a PR myself to fix that, and once it's in we should be able to get appveyor to pass in this PR.

ccaviness:
Ah, gotcha - glad you had time.

branan:
@ccaviness My PR was been merged, so if you rebase this then *hopefully* the appveyor tests will pass.

ccaviness:
@branan "so if you rebase this"

I can never manage to do this right the first time.

ccaviness:
... Ok, I think I may have done it? Let's see what appveyor says.

ccaviness:
Looks like it's all green.

ccaviness:
Is there anything else I need to take care of?

adrienthebo:
@ccaviness this is still on our radar, but we've been pretty swamped - we'll try to get back to this ASAP! @branan, LGTM, wanna merge this? I can run the merge and shepard it but my brain is not running at 100% right now.

branan:
@adrienthebo I'm just waiting for the release to run through and then I'm gonna go on a wild and crazy PR-merging rampage :)

branan:
:boom: Done! Thanks @ccaviness

PR #5120, opened 2016-07-14
(maint) Fixes for SSL Extension work

justinstoller:
This fixes three important issues with the recent SSL Extension work that's landed in Puppet.

### DIFF
--- a/lib/puppet/ssl/certificate_authority.rb
+++ b/lib/puppet/ssl/certificate_authority.rb
@@ -295,8 +295,6 @@ class Puppet::SSL::CertificateAuthority
       cert_type = :server
       issuer = host.certificate.content
 
-      # This allows for various bootstrapping scenarios
-      options[:allow_dns_alt_names] = true if hostname == Puppet[:certname].downcase
       # Make sure that the CSR conforms to our internal signing policies.
       # This will raise if the CSR doesn't conform, but just in case...
       check_internal_signing_policies(hostname, csr, options) or
@@ -330,6 +328,8 @@ class Puppet::SSL::CertificateAuthority
   def check_internal_signing_policies(hostname, csr, options = {})
     options[:allow_authorization_extensions] ||= false
     options[:allow_dns_alt_names] ||= false
+    # This allows for masters to bootstrap themselves in certain scenarios
+    options[:allow_dns_alt_names] = true if hostname == Puppet[:certname].downcase
 
     # Reject unknown request extensions.
     unknown_req = csr.request_extensions.reject do |x|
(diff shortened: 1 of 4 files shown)


justinstoller:
Ping @nathanielksmith, @haus for review. Also @rlinehan, 1a9a2f7 resolves our current CI issues.

lib/puppet/ssl/certificate_authority.rb
```diff
@@ -288,14 +288,15 @@ def sign(hostname, options={})
...
     else
-      options[:allow_dns_alt_names] = true if hostname == Puppet[:certname].downcase
       unless csr = Puppet::SSL::CertificateRequest.indirection.find(hostname)
         raise ArgumentError, "Could not find certificate request for #{hostname}"
       end
 
       cert_type = :server
       issuer = host.certificate.content
 
+      # This allows for various bootstrapping scenarios
```
nathanielksmith: i trust you but um i cannot understand how this change affects a thing?

haus: Can the `Puppet::SSL::CertificateRequest.indirection.find(hostname)` call change the options hash? Otherwise I'm in the same boat as @nathanielksmith and am not sure how this helps.

justinstoller: So this change moves what was being set at line 291 down to where it's actually consumed and comments why we do this overriding of default behavior in the first place. This comment and line of code are then copied to the [certificate authority interface's sign method](https://github.com/puppetlabs/puppet/pull/5120/files/1a9a2f7217f5b3d3dfdfbaa6dc4a369ba3f17aab#diff-ebefe776564cbe300da0fb3a405e0713R283) where we now also call `check_internal_signing_policies`.

That copy pasta train wreck is the "fix" because we now call `check_internal_signing_policies` before we call `sign` when using the `puppet cert` command (so we can get potential issues to the user faster). The bug was that the conditional overriding of default behavior didn't follow the method invocation into the certificate authority interface (we saw this with our duplicate cert acceptance test).

Thinking about this more, I think it might be a better idea to move the hostname == certname check into the `check_internal_signing_policies` so that the callers don't have to be concerned (ie this check seems to be part of our internal signing policies). What do you guys think?

justinstoller: I went ahead and did that refactor in b744d0d, lemme know what you think of it.

haus:
These changes look sane to me. I think it would probably be worth writing a small spec test to cover this failure so that if it happens again fixing it will have a shorter feedback cycle.

justinstoller:
@haus that sounds like a great idea. I'll do that right away. For anyone interested I have a functional test environment here: `risx7cqu5ecuglt.delivery.puppetlabs.net`

nathanielksmith:
thanks for explanation justin; :+1: once that spec test is done (agreed that that is a great idea)

haus:
@justinstoller nice. I'll merge once travis goes green.

PR #5121, opened 2016-07-15
(PUP-6508) Make tidy resources non-isomorphic

ccaviness:
Since tidy sets path as namevar, puppet tries to create aliases for it. This means multiple tidy resources with the same path (but different matches, e.g.) will fail.

### DIFF
--- a/lib/puppet/type/tidy.rb
+++ b/lib/puppet/type/tidy.rb
@@ -16,6 +16,9 @@ Puppet::Type.newtype(:tidy) do
     actual deletion.
     "
 
+  # Tidy names are not isomorphic with the objects.
+  @isomorphic = false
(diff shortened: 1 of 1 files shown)


branan:
@ccaviness This looks like the correct code change for what you want to do. I want to have a quick chat with @ahpook about whether he has any thoughts/plans for the tidy resource before I push the button.

If he thinks this should go in, we'll get it merged.

ahpook:
LGTM 👍 Love to see `tidy` getting some attention.

ccaviness:
friendlyping

adrienthebo:
In exchange for this merge I am stealing your excellent gif. Thanks for the contribution!

PR #5122, opened 2016-07-15
(PUP-6042) acceptance: test flag overrides use_cached_catalog

johnduarte:
This commit adds an acceptance test to verify that using the
`--test` flag with `puppet agent` overrides the `--use_cached_catalog`
setting. This combination should not result in the use of a cached
catalog.


seangriff:
Looks good.  Verified against windows 10 ent x86_64 and ubuntu 10.04 x86_64 clients.

PR #5125, opened 2016-07-19
(PUP-6457) Always attempt remount for out of sync options

HAIL9000:
Prior to this commit, it was possible for puppet to thing that a
mounted device was in sync with it's resource definition when it was
not. This was because if a resource had already been mounted, and a
remount was attempted with bad options, the remount would fail even
though the entry in /etc/fstab would be updated. Puppet was only
checking the resource options against etc/fstab which meant that
after this point, it would believe the mounted object to be in sync
with the puppet resource describing it, even though it was not.

Now, puppet will check the resource's options against both /etc/fstab
and the output of the mount command (on linux operating systems). This
means that it can now detect if the correct object is not mounted and
will attempt a remount.

### DIFF
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -272,7 +272,7 @@ Puppet::Type.type(:mount).provide(
     end
     mount_output.each do |line|
       if match = regex.match(line) and name = match.captures.first
-        live_options_platforms = ["Linux"]
+        live_options_platforms = ["Linux", "Darwin"]
         if live_options_platforms.include? Facter.value(:kernel)
           options = line[/\(.*\)/].tr('()', '')
           instances << {:name => name, :mounted => :yes, :live_options => options}
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -209,14 +209,20 @@ Puppet::Type.type(:mount).provide(
 
     # Update fstab entries that are mounted
     providers.each do |prov|
-      if mounts.delete({:name => prov.get(:name), :mounted => :yes}) then
-        prov.set(:ensure => :mounted)
+      mounts.delete_if do |mount|
+        if mount[:name] == prov.get(:name) && mount[:mounted] == :yes
+          prov.set(:ensure => :mounted)
+          prov.set(:live_options => mount[:live_options])
+          true
+        else
+          false
+        end
       end
     end
 
     # Add mounts that are not in fstab but mounted
     mounts.each do |mount|
-      providers << new(:ensure => :ghost, :name => mount[:name])
+      providers << new(:ensure => :ghost, :name => mount[:name], :live_options => mount[:live_options])
     end
     providers
   end
@@ -232,6 +238,7 @@ Puppet::Type.type(:mount).provide(
     #   to fstab we need to know if the device was mounted before)
     mountinstances.each do |hash|
       if mount = resources[hash[:name]]
+        mount.provider.set(:live_options => hash[:live_options])
         case mount.provider.get(:ensure)
         when :absent  # Mount not in fstab
           mount.provider.set(:ensure => :ghost)
@@ -254,6 +261,7 @@ Puppet::Type.type(:mount).provide(
       else
         / on (\S*)/
     end
+
     instances = []
     mount_output = mountcmd.split("\n")
     if mount_output.length >= 2 and mount_output[1] =~ /^[- \t]*$/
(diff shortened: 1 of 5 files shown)


lib/puppet/provider/mount/parsed.rb
```diff
@@ -23,7 +23,7 @@
   when "Solaris"
     @fields = [:device, :blockdevice, :name, :fstype, :pass, :atboot, :options]
   else
-    @fields = [:device, :name, :fstype, :options, :dump, :pass]
+    @fields = [:device, :name, :fstype, :options, :dump, :pass, :live_options]
```
branan: My guess is that the reason our output is broken is because `parsedfile` is actually trying to append the new `live_options` field. I think you want to omit this from the fields variable. We may need to find a way to trick `parsedfile` into accepting the extra data and passing it into the `property_hash` once we remove this, though

adrienthebo: The implementation of [`Puppet::Provider#set`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider.rb#L543-L547) simply assigns values to `@property_hash`, so we can just call `#set` without having to explicitly add the `live_options` param/value/thing.

HAIL9000: Thank you!!

lib/puppet/type/mount.rb
```diff
@@ -177,6 +177,29 @@ def syncothers
         appear in fstab. For many platforms this is a comma delimited string.
         Consult the fstab(5) man page for system-specific details."
 
+          def insync?(is)
+            if @resource[:ensure] == :mounted
+              fstab_options = provider.property_hash[:options] || ''
```
adrienthebo: `property_hash.fetch(:options, '')` is nice when you want to pick out a value from a hash and provide a default value if the key isn't set.

lib/puppet/provider/mount/parsed.rb
```diff
@@ -264,7 +272,12 @@ def self.mountinstances
     end
     mount_output.each do |line|
       if match = regex.match(line) and name = match.captures.first
-        instances << {:name => name, :mounted => :yes} # Only :name is important here
+        if Facter.value(:kernel) == 'Linux'
```
adrienthebo: Are we planning on only supporting Linux for remounting on mismatched options, or Solaris and others?

HAIL9000: Bringing in @branan for feedback on this one. I wasn't quite sure what conclusion we came to, if we should include that work in this ticket or file a new one. It seems like Darwin and SunOS at least would be pretty trivial to add support for but I'm not sure about anything else

spec/unit/type/mount_spec.rb
```diff
@@ -530,6 +530,11 @@ def run_in_catalog(*resources)
       run_in_catalog(resource)
     end
 
+    it "should detect out of sync options" do
+      #provider.property_hash[:live_options] = "foo";
```
HAIL9000: I'm not sure what to do about this test. It technically works (e.g. it fails without the fix and passes with it) but that is only because if there are no live options retrieved from the output of the mount command, they default to an empty string. I would feel better if we could actually stub that value to something, but whenever I try to add live_options to the property hash of the provider the test freaks out because there's an unrecognized parameter.

HAIL9000:
Hm, some of the failures I saw before are back. They might be related to the switch to using "fetch". Looking into it

spec/unit/type/mount_spec.rb
HAIL9000: This actually has caused me to realize an issue with this solution- if we're running on a non-linux platform no options will be extracted and then in the insync method they will become an empty string meaning that the options will always appear out of sync. Trying to fix this now

adrienthebo:
@HAIL9000 what's the status of this PR? Are the Travis Ci failures known?

HAIL9000:
@adrienthebo Unfortunately the spec test I wrote was only passing because of a bug in my solution and since I fixed it I haven't been able to fix the spec test. What I really need to be able to do is stub `live_options` but I can't figure out a way to do that without Puppet raising an error. So I was just waiting for someone to review this PR and see if they had suggestions (you can see my comments on the spec test above for more details)

branan:
I'm :+1: on this at this point. It doesn't solve all the weird edge cases, but it'll handle the case in the ticket.

We're going to have to put some thought into how to phrase the release notes for this, though

branan:
@HAIL9000 should this be targeted at stable to land in 4.7.1?

HAIL9000:
@branan That seems like a good idea? The customer has been waiting on it for a while. I'm like a little stressed because I FEEL like it SHOULDN'T break anything else but I can't be certain

PR #5127, opened 2016-07-19
(PUP-6434) Define Sensitive data type

adrienthebo:
This commit defines the Sensitive data type and instances of that type.
The intent of this new data type is to wrap plaintext information to
prevent accidental leakage of marked information and ultimately enable
the automatic transformation of plaintext information into an encrypted or
obfuscated representation.


adrienthebo:
@hlindberg I've parameterized the type itself but don't know specifically how to take advantage of that. Are there API docs on how the Puppet type system works at a higher level, with the type factory, type parser, type calculator, and so forth?

hlindberg:
@adrienthebo no docs on how to create new parameterized types, but possible to mimic some type that already exists. It is basically the places you mentioned.
Using the Sensitive[T] would for instance be to have say `define foo(Sensitive[Numeric] $val) { }`, and the users can create instance of say Sensitive(42), or Sensitive(3.14), which would be a Sensitive[Integer], or Sensitive[Float], both assignable to Sensitive[Numeric]. I.e. that Sensitive[T] is assignable from a Sensitive[T2], if T is assignable from T2.

But. maybe that was not your question...

lib/puppet/pops/types/p_sensitive_type.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    def inspect
+      "#<Sensitive>"
+    end
+  end
+
+  attr_reader :element_type
+
+  def initialize(element_type = Object)
```
hlindberg: The @element_type should be a puppet type system type, so it should be PAnyType by default (instead of Object which is a Ruby class).

lib/puppet/pops/types/p_sensitive_type.rb
```diff
@@ -0,0 +1,69 @@
...
+    @element_type = element_type
+  end
+
+  def instance?(o, guard = nil)
+    o.is_a?(Sensitive)
```
hlindberg: Add:
`&& @element_type.instance?(o.value, guard)`
since, for example, a `Sensitive("foo")` is not an instance of `Sensitive[Integer]`

lib/puppet/pops/types/p_sensitive_type.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def eql?(o)
+    self.class == o.class
```
hlindberg: for two instances of Sensitive to be equal, their @element_type attributes must also be equal - that is; Sensitive[Integer] == Sensitive[Integer], and Sensitive[Numeric] != Sensitive[Integer]

lib/puppet/pops/types/p_sensitive_type.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def generalize
+    DEFAULT
+  end
+
+  def self.new_function(_, loader)
+    @new_function ||= Puppet::Functions.create_loaded_function(:new_Sensitive, loader) do
+      dispatch :new do
```
hlindberg: I have a feeling that the name 'new' for a ruby method is asking for trouble. Suggest using `from_any` to follow the pattern used in many other type's new_function

lib/puppet/pops/types/p_sensitive_type.rb
```diff
@@ -0,0 +1,69 @@
...
+        param 'Any', :o
+      end
+
+      def new(o)
```
hlindberg: rename to `from_any`

lib/puppet/pops/types/p_sensitive_type.rb
hlindberg: Probably also want to unwrap if given a Sensitive value. That is, if users does something like this:
```
$a = Sensitive(42)
$b = Sensitive($a)
```
Is the type of $b `Sensitive[Sensitive[Integer]]`? That seems nonsensical. Thus, a dispatcher for `Sensitive[Any]` would simply return the given value since it is already a `Sensitive` instance.

lib/puppet/pops/types/p_sensitive_type.rb
```diff
@@ -0,0 +1,69 @@
...
+        Sensitive.new(o)
+      end
+    end
+  end
+
+  private
+
+  def _assignable?(o, guard)
+    self.class == o.class
```
hlindberg: Add:
`&& @element_type.assignable?(o, guard)`
since a `Sensitive[Integer] cannot be assigned from a Sensitive[String]`

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -644,6 +644,11 @@ def infer_Symbol(o)
   end
 
   # @api private
+  def infer_Sensitive(o)
+    PSensitiveType::DEFAULT
```
hlindberg: This would make all sensitive instances be of the type `Sensitive[Any]`. It should instead infer the type of o and pass that on as a parameter. That is:
```
PSensitiveType.new(infer(o.value))
```

lib/puppet/pops/types/type_factory.rb
```diff
@@ -40,6 +40,13 @@ def self.float
     PFloatType::DEFAULT
   end
 
+  # Produces the Sensitive type
+  # @api public
+  #
+  def self.sensitive
+    PSensitiveType::DEFAULT
+  end
```
hlindberg: You want to be able to create parameterized types, so something like:
```
def self.sensitive(t = PAnyType)
  PSensitiveType.new(t)
end
```
Many of the other types use a `nil` element_type (or equivalent) to mean `Any` as that makes it easier to share the singleton `DEFAULT`, but it is not required to do so (it requires a bit more logic in a couple of other places if @element_type.nil? should mean the same as @element_type = PAnyType).

lib/puppet/pops/types/type_formatter.rb
```diff
@@ -130,6 +130,9 @@ def string_PDataType(_)    ; @bld << 'Data'    ; end
   def string_PNumericType(_) ; @bld << 'Numeric' ; end
 
   # @api private
+  def string_SensitiveType(_) ; @bld << 'Sensitive' ; end
+
```
hlindberg: You need something a bit more elaborate, to make it format the parameter T (that is `element_type`) as well. (There is a typo as well as the type is PSensitiveType, not SensitiveType).
```
  def string_PSensitiveType(t)
    element_type = t.element_type
    append_array('Sensitive', element_type.nil? || element_type.class == PAnyType) do
        append_string(element_type)
    end
  end
```
This formats `Sensitive[Any]` as `Sensitive`, and a `Sensitive[Integer]` as `Sensitive[Integer]`, which would be expected (since the `Any` in `Sensitive[Any]` is redundant).

lib/puppet/pops/types/type_parser.rb
```diff
@@ -140,6 +140,7 @@ def self.type_map
         'iterable'     => TypeFactory.iterable,
         'iterator'     => TypeFactory.iterator,
         'string'       => TypeFactory.string,
+        'sensitive'    => TypeFactory.sensitive,
```
hlindberg: In addition to this, since we want to be able to create parameterized Sensitive type instances, for example `Sensitive[Numeric]`, the method `interpret_AccessExpression` must also be modified with an entry for `'sensitive'`

something like this:
```
    when 'sensitive'
      if parameters.size != 1
        raise_invalid_parameters_error('Sensitive', 1, parameters.size)
      end
      param = parameters[0]
      assert_type(ast, param)
      TypeFactory.sensitive(param)
```

spec/unit/pops/types/p_sensitive_type_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+describe 'Sensitive Type' do
+  include PuppetSpec::Compiler
+
+  context 'as a type' do
+    it 'can be created with the type factory' do
+      t = TypeFactory.sensitive
+      expect(t).to be_a(PSensitiveType)
+      expect(t).to eql(PSensitiveType::DEFAULT)
+    end
+
```
hlindberg: + can be created with a parameter, for example `TypeFactory.sensitive(PIntegerType::DEFAULT)`

spec/unit/pops/types/p_sensitive_type_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+          $t = Sensitive
+          notice(Sensitive =~ Type[ Sensitive ])
+          notice(Sensitive == Sensitive)
+          notice(Sensitive < Sensitive)
+          notice(Sensitive > Sensitive)
+        CODE
+        expect(eval_and_collect_notices(code)).to eq(['true', 'true', 'false', 'false'])
+      end
+    end
+  end
```
hlindberg: Here we would expect that `Sensitive[Numeric] > Sensitive[Integer]`

spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -208,6 +208,12 @@ def types
       end
     end
 
+    context 'sensitive' do
+      it 'translates to PSensitiveType' do
+        expect(calculator.infer(PSensitiveType::Sensitive.new("hunter2")).class).to eq(PSensitiveType)
+      end
+    end
```
hlindberg: We would expect that the type of `Sensitive("hunter")` is `Sensitive[String[6,6]]`

hlindberg:
@adrienthebo you are well on the way - I made detailed comments about what you need to do to make Senstive be a parameterized type. Hope my comments make sense.

lib/puppet/pops/types/p_sensitive_type.rb
adrienthebo: I originally tried that, but ran into some issues with ruby require order and backed off. I can swap this out easily enough.

lib/puppet/pops/types/p_sensitive_type.rb
adrienthebo: To confirm, o is an instance of a subclass of PAnyType?

adrienthebo: I was trying to figure out how to parameterization here but I originally thought that o was an actual data instance instead of a type instance.

lib/puppet/pops/types/p_sensitive_type.rb
adrienthebo: Oh that's perfect, good catch!

lib/puppet/pops/types/type_parser.rb
adrienthebo: I implemented it like so, making the parameterized type optional:

```ruby
    when 'sensitive'
      if parameters.size == 0
        TypeFactory.sensitive
      elsif parameters.size == 1
        param = parameters[0]
        assert_type(ast, param) unless param.is_a?(String)
        TypeFactory.sensitive(param)
      else
        raise_invalid_parameters_error('Sensitive', '0 to 1', parameters.size)
      end
```

What do you think of this approach?

spec/unit/pops/types/p_sensitive_type_spec.rb
adrienthebo: Done.

spec/unit/pops/types/type_calculator_spec.rb
adrienthebo: We actually need to expect the opposite to prevent information leakage - exposing the fact that this is a 6 character string leaks a huge amount of information. If this was a password this information reduces the brute force attack space of this password enormously. I've pushed another commit to enforce generalization of the contained type; I think this needs more thinking.

spec/unit/pops/types/p_sensitive_type_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+      code =<<-CODE
+        $o = Sensitive("hunter2")
+        $x = Sensitive($o)
+        notice(assert_type(Sensitive[String], $x))
+      CODE
+      expect(eval_and_collect_notices(code)).to eq(['Sensitive [value redacted]'])
+    end
+  end
+
+  it "enforces wrapped type constraints" do
```
adrienthebo: @hlindberg this test actually fails even before I added the contained type generalization, which is incredibly weird. The verification that the actual parameter matches `Sensitive[Array[String[4, 8]]]` is performed by the `#_assignable?` method, right?

lib/puppet/pops/types/type_parser.rb
hlindberg: That is fine. You probably do not want to accept a string instead of a type. For example the type `Sensitive['herp derp']` does not make sense. 
(You may have gleaned at the `Optional` type, which allows a string as a special thing when `Optional` is used to describe entries in a `Struct`. If you want something like that to work, more things are needed elsewhere as that is truly special.

So, drop the `unless param.is_a?(String)` bit.

spec/unit/pops/types/type_calculator_spec.rb
hlindberg: This is indeed a tricky problem. Generalizing the type removes all details, but it also makes it impossible to use more precise typing when say, Sensitive[String[6]] or Sensitive[Enum[red,blue, green]]. Would it be enough to present the generalized type T when the type is stringified? It could perhaps output the type as Stringified[Redacted] or somesuch.

Another alternative would be to support a Sensitive subtype that the inference uses. This type does not reveal the detailed type, but the type system would have access to the detailed information. Done that way, user entered types like `Sensitive[Enum[blue,red,green]]` would print the details, but an inferred type would not.

Indeed a hard problem.

spec/unit/pops/types/p_sensitive_type_spec.rb
hlindberg: It (assert_type function) does this: `Puppet::Pops::Types::TypeCalculator.instance?(type,value)`. There is a default impl of that which ends up in `#_assignable?` - so yes, but depends on the implementation of `instance?` on that type (there could be special rules for `nil` element type, and the like, or the question can be answered without inference and an assignability check.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -644,6 +644,11 @@ def infer_Symbol(o)
   end
 
   # @api private
+  def infer_Sensitive(o)
+    PSensitiveType.new(infer(o.unwrap))
```
hlindberg: this gives generalized inference, in general `infer_set` should be used to get the richest information.

lib/puppet/pops/types/type_parser.rb
adrienthebo: I dropped that after reviewing the behavior more closely.

spec/unit/pops/types/type_calculator_spec.rb
adrienthebo: I'm also considering treating this as a limitation - if you want to use sensitive information, your ability to deeply validate data goes away. This is less of an issue with the Sensitive type since we're still handling sensitive information, but when we start ingesting Encrypted information we absolutely cannot expose information about the encrypted information. Exposing the type information alone is somewhat risky; including any information in the associated type information is actively dangerous.

spec/unit/pops/types/p_sensitive_type_spec.rb
hlindberg: We may need Redacted[T], and RedactedType[T] sooner that expected to handle things like this ?

spec/unit/pops/types/type_calculator_spec.rb
adrienthebo: And I'm concerned about trying to limit when we do the type generalization, because we run the risk of missing some code that does `puts o.type.type` and leaks that it's a `String[10, 10]` instead of using some blessed pathway. If we retain sensitive information then we risk leaking it later so wherever possible I think we should destroy any sensitive information as soon as possible.

lib/puppet/pops/types/type_calculator.rb
adrienthebo: What would be a good way to test this code path with `infer_set`?

spec/unit/pops/types/p_sensitive_type_spec.rb
adrienthebo: I don't think that Redacted[T] will play nice here; we're effectively redacting integers and the work that it'll take to allow Redacted[Integer] to be used where Integers once were seems daunting. I had envisioned that Redacted[T] would be used in reports and catalogs and wouldn't be used too much deep within the internals of Puppet. Moreover I meant to use those types to cleanly fail when someone tried to unwrap redacted information; in this case we don't want to blow up if someone tries to read this information - I just want that information elided.

adrienthebo:
@hlindberg I am done making additional changes and am ready for a final review on this.

hlindberg:
Ok, will review. I think the decision to generalize is a decent tradeoff for not leaking detailed type information. We can revise that later when we have more time to come up with something that allows detailed typechecking during compilation.

lib/puppet/pops/types/p_sensitive_type.rb
```diff
@@ -0,0 +1,68 @@
...
+      dispatch :from_any do
+        param 'Any', :value
+      end
+
+      def from_any(value)
+        Sensitive.new(value)
+      end
+
+      def from_sensitive(value)
+        Sensitive.new(value.unwrap)
```
hlindberg: Do you want a copy? The value is immutable so could return the very same instance.

lib/puppet/pops/types/type_calculator.rb
```diff
@@ -644,6 +644,11 @@ def infer_Symbol(o)
   end
 
   # @api private
+  def infer_Sensitive(o)
+    PSensitiveType.new(infer_set(o.unwrap))
```
hlindberg: I know I told you to do this :-). With the generalize in the constructor, the rich information produced by `infer_set` will be erased. We could do a small optimization here, and use `infer` (as it generalizes).
OTOH, we then have the decision that the type generalizes in two places, and we cannot trust all types given to PSensitiveType.new to be generalized so it must be done there anyway.
Suggest keeping it as it is now. Just wanted to make you aware of this.

lib/puppet/pops/types/type_parser.rb
```diff
@@ -432,6 +433,17 @@ def interpret_AccessExpression(ast, context)
       end
       TypeFactory.string(size_type)
 
+    when 'sensitive'
+      if parameters.size == 0
+        TypeFactory.sensitive
+      elsif parameters.size == 1
+        param = parameters[0]
+        assert_type(ast, param) unless param.is_a?(String)
+        TypeFactory.sensitive(param)
```
hlindberg: hm, the `unless param.is_a?(String)` is still there...

lib/puppet/pops/types/p_sensitive_type.rb
adrienthebo: Good call, will do.

kylog:
confetti1

PR #5131, opened 2016-07-21
(maint) Ensure WindowsInstaller emits UTF-8

Iristyle:
- Previously the Windows package provider would use Rubys COM support
  to retrieve values.  Unfortunately, Ruby returns strings in the
  local codepage instead of UTF-8 which can cause downstream breakages
  in Puppet when string manipulation / concatenation is performed.

### DIFF
--- a/lib/puppet/provider/package/windows/msi_package.rb
+++ b/lib/puppet/provider/package/windows/msi_package.rb
@@ -23,7 +23,7 @@ class Puppet::Provider::Package::Windows
           MsiPackage.new(get_display_name(values),
                          values['DisplayVersion'],
                          name, # productcode
-                         inst.ProductInfo(name, 'PackageCode'))
+                         inst.ProductInfo(name, 'PackageCode').encode(Encoding::UTF_8))
         end
       end
     end


Iristyle:
@puppetlabs/windows this one should be an easy merge - could add tests if anyone cares.

lib/puppet/provider/package/windows/msi_package.rb
```diff
@@ -23,7 +23,7 @@ def self.from_registry(name, values)
           MsiPackage.new(get_display_name(values),
                          values['DisplayVersion'],
                          name, # productcode
-                         inst.ProductInfo(name, 'PackageCode'))
+                         inst.ProductInfo(name, 'PackageCode').encode(Encoding::UTF_8))
```
ferventcoder: ramifications here? This looks like it may need a ticket?

ferventcoder:
Ha! Even @Iristyle rewrites his own commits! 
image

lib/puppet/provider/package/windows/msi_package.rb
ferventcoder: It looks like a tiny change, but I'd almost prefer tests.

Iristyle: The fortunate thing here is that this a GUID, so shouldn't ever be outside of the range of `[0-9A-F\-]`

That said, Ruby is still going to pass this out as a current codepage string (based on `Encoding.default_external`), which can make Ruby concatenation (and other string ops) explode depending on where they occur.

The string smashing was the root cause of https://github.com/puppetlabs/puppet/pull/5113

ferventcoder:
Still would love to see a JIRA ticket for this.

Iristyle:
Once #5419 is merged, this should no longer be an issue as the default string encoding for any interactions with `WIN32OLE` will be set to UTF-8

PR #5133, opened 2016-07-21
(PUP-6531) improve service provider detection on Ubuntu

nacc:
Rather than relying on version strings to determine which init-system is
in use, use the fact that the presence of /run/systemd/system (on
Debian-based systems generally, but I only can confirm for Ubuntu)
implies systemd is suitable.

Additionally, do not attempt to to use upstart in that case. However, if
systemd is not in use and upstart is the init-system (as detected by the
presence of the appropriate binaries as currently), treat upstart as
suitable.

Not-signed-off-by: Nishanth Aravamudan nish.aravamudan@canonical.com

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -9,13 +9,19 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
 
   commands :systemctl => "systemctl"
 
+  Facter.add(:running_on_systemd) do
+    setcode do
+      Dir.exists? "/run/systemd/system"
+    end
+  end
+
   if Facter.value(:osfamily).downcase == 'debian'
     # With multiple init systems on Debian, it is possible to have
     # pieces of systemd around (e.g. systemctl) but not really be
     # using systemd.  We do not do this on other platforms as it can
     # cause issues when running in a chroot without /run mounted
     # (PUP-5577)
-    confine :exists => "/run/systemd/system"
+    confine :true => Facter.value(:running_on_systemd)
   end
 
   defaultfor :osfamily => [:archlinux]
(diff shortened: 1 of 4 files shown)


nacc:
To be clear, I'm expecting this to fail to build (as it did). I don't know puppet or ruby well enough to express that systemd should be suitable if /run/systemd/system is detected and upstart should be suitable if not. Both will be further constrained by the expected executables.

nacc:
I just tweaked this a bit, to basically not let upstart only rely on the executable detection for suitability. I think it's incorrect, actually, and upstart should only be used if systemd is not detected as being running.

Also, I've found `puppet describe service --providers`, but I've not yet found a way to tell me the list of puppet providers that are suitable? I can run with --debug and get a list of those not suitable, but I can't yet determine if puppet is using systemctl because of the debian provider or the systemd provider.

nacc:
Ok, this version works now for CI purposes and I think we'll pull this into Ubuntu, at least, as it means we don't need to worry that upstream's puppet has the right version checks in it.

I think it also fixes a long-standing bug (pseudo-bug?) with puppet where just having upstart (package) installed was enough for upstart to be considered a suitable init-system (which leads to issues like https://bugs.launchpad.net/ubuntu/+source/puppet/+bug/1570472).

nacc:
@kylog would you be able to take a look at this? Just want  some verification this is sane, so we can at least apply it to Ubuntu's puppet for the time being.

fnoop:
Could we have some movement on this?  This is a really basic, fundamental bug in puppet that effectively makes it unusable on recent versions of Ubuntu.

jhelwig:
@nacc This all looks pretty reasonable to me. There's a minor conflict in the systemd provider that prevents merging though.

nacc:
I am out of the country right now, but I will do my best to refresh my PR
soon. If someone wants to rebase it themselves in the meanwhile, feel free.

Thanks,
Nish

>

kris-bosland:
Hi @nacc, with the 5.5.x branch being cut, can you check again on this conflict?  Thanks!

branan:
Hi,

We haven't recieved responses to our last couple comments - Please let us know if you want to keep moving forward with this, otherwise I'm going to close it in a couple more weeks

melissa:
Since there has been no movement on this PR in the last few weeks, I'm going to close this. This doesn't mean we don't value this work. If you decide to pick this up again, please either reopen this pull request or create a new one. Thank you!

PR #5134, opened 2016-07-22
(PUP-5949) Load Resource Types from Pcore based Resource Type description

thallgren:
Work in progress on environment isolation and loaders.


lib/puppet/pops/loader/runtime3_type_loader.rb
```diff
@@ -21,28 +22,51 @@ def to_s()
...
+  # instances, then check for a Puppet::Type and lastly check for a defined type.
+  #
+  def find_impl(typed_name)
     name = typed_name.name
-    value = @environment.known_resource_types.find_definition(name)
-    if value.nil?
+    te = Types::TypeParser.type_map.include?(name) ? nil : @resource_3x_loader.find(typed_name)
+    if te.nil? || te.value.nil?
       # Look for Puppet::Type
       value = Puppet::Type.type(name)
```
hlindberg: this step can be skipped if the name is qualified since no Puppet::Type ever finds a qualified name (there may be many thousands such calls)

lib/puppet/pops/loader/runtime3_type_loader.rb
```diff
@@ -21,28 +22,51 @@ def to_s()
...
       if value.nil?
-        # Cache the fact that it wasn't found
-        set_entry(typed_name, nil)
-        return nil
+        # Look for a defined type
```
hlindberg: helps to say "user defined type" here

hlindberg:
Will merge as this now contains the required functionality. More tests needs to be added for the "bigger picture".

PR #5137, opened 2016-07-23
(PUP-5938) Fix localized_domain test mocks

Iristyle:
- A new method localized_domains was introduced in 05d1ca2 to
   Puppet::Util::Windows::ADSI::Shared to handle properly generating
   URIs when the names BUILTIN and NT AUTHORITY are localized on
   non-English versions of Windows.

   At that time, the calls should have been mocked in:

   * provider/group/windows_adsi_spec.rb
   * provider/user/windows_adsi_spec.rb

   The calls actually touch the system, and depending on test
   ordering and how calls to name_to_sid_object are mocked, the code
   execution path could produce failures.

   Fix this oversight by simply mocking the return values for
   localized_domains as they are not needed for these tests.

Paired-with: Daniel Lu <daniel.lu@puppet.com>


Iristyle:
@glennsarti this should be good to merge once the tests pass

Iristyle:
Closed in favor of #5171 which targets `stable` instead of `master`

PR #5139, opened 2016-07-25
(PUP-6295) Add corrective_change field to report and report events

kbarber:
The corrective_change field in events marks an event that was presumed to be a
manual out-of-band change from Puppet, that Puppet then corrected. In the
context of a noop or non-applying change, the same is true except that the
change was not corrected but is intended to be corrected once a real run is
applied.

This patch introduces the work necessary to calculate this change, introducing a
transaction store that stores the last known best applied value so we can
calculate that a change was indeed unexpected.

For each report, we also mark the resource that had a corrective_change plus
mark an overall report of this also so this information can be stored in
backends like PuppetDB for quick searching.

### DIFF
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -160,6 +160,11 @@
         "environment": {
             "description": "The name of the environment that was used for the puppet run (e.g. \"production\").",
             "type":        "string"
+        },
+
+        "corrective_change": {
+            "description": "True if a change or noop event in this report was caused by an unexpected change to the system between Puppet runs.",
+            "type":        "boolean"
         }
     },
 
@@ -304,7 +309,8 @@
                                     "skipped",
                                     "failed_to_restart",
                                     "restarted",
-                                    "scheduled"
+                                    "scheduled",
+                                    "corrective_change"
                                 ]
                             },
                             {
@@ -317,7 +323,8 @@
                                     "Skipped",
                                     "Failed to restart",
                                     "Restarted",
-                                    "Scheduled"
+                                    "Scheduled",
+                                    "Corrective change"
                                 ]
                             },
                             {
@@ -526,6 +533,11 @@
                     "description": "An array of strings; each element represents a container (type or class) that, together, make up the path of the resource in the catalog.",
                     "type":        "array",
                     "items":       { "type": "string" }
+                },
+
+                "corrective_change": {
+                    "description": "True if a change or noop event on this resource was caused by an unexpected change to the system between Puppet runs.",
+                    "type":        "boolean"
                 }
             },
 
@@ -537,7 +549,8 @@
                 "tags",
                 "events",
                 "skipped",
-                "containment_path"
+                "containment_path",
+                "corrective_change"
             ],
             "additionalProperties":   false
         },
@@ -616,6 +629,11 @@
                 "time": {
                     "description": "The time at which the property was evaluated. In ISO 8601 format with 9 characters second-fragment",
                     "type":        "string"
+                },
+
+                "corrective_change": {
+                    "description": "True if this event was caused by an unexpected change to the system between Puppet runs.",
+                    "type":        "boolean"
                 }
             },
             "required": [
@@ -624,7 +642,8 @@
                 "message",
                 "name",
                 "status",
-                "time"
+                "time",
+                "corrective_change"
             ],
 
             "additionalProperties": false
--- a/lib/puppet/property.rb
+++ b/lib/puppet/property.rb
@@ -102,6 +102,22 @@ class Puppet::Property < Puppet::Parameter
       raise ArgumentError, "Supported values for Property#array_matching are 'first' and 'all'" unless [:first, :all].include?(value)
       @array_matching = value
     end
+
+    # Used to mark a type property as having or lacking idempotency (on purpose
+    # generally). This is used to avoid marking the property as a
+    # corrective_change when there is known idempotency issues with the property
+    # rendering a corrective_change flag as useless.
+    # @return [Boolean] true if the property is marked as idempotent
+    def idempotent
+      @idempotent.nil? ? @idempotent = true : @idempotent
+    end
+
+    # Attribute setter for the idempotent attribute.
+    # @param [bool] value boolean indicating if the property is idempotent.
+    # @see idempotent
+    def idempotent=(value)
+      @idempotent = value
+    end
   end
 
   # Looks up a value's name among valid values, to enable option lookup with result as a key.
@@ -346,6 +362,29 @@ class Puppet::Property < Puppet::Parameter
     end
   end
 
+  # This method tests if two values are insync? outside of the properties current
+  # should value. This works around the requirement for corrective_change analysis
+  # that requires two older values to be compared with the properties potentially
+  # custom insync? code.
+  #
+  # @param [Object] should the value it should be
+  # @param [Object] is the value it is
+  # @return [Boolean] whether or not the values are in sync or not
+  # @api private
+  def insync_values?(should, is)
+    # Here be dragons. We're duplicating the resource property itself so we can pass
+    # a new should value without side-effect. This kind of mechanism is undesirable,
+    # but it avoids the risk of causing side-effects on live properties somewhere
+    # else in code. The duplicate should be short-lived and is only required for this
+    # comparison so it can be GC'd directly afterwards. Unfortunately there isn't
+    # an api compatible way of avoiding this, as both should and insync? behaviours
+    # are part of the API. Future API work should factor this kind of arbitrary comparison
+    # into the API to remove this complexity. -ken
+    test_property = self.dup
+    test_property.should = should
+    test_property.insync?(is)
+  end
+
   # Checks if the given current and desired values are equal.
   # This default implementation performs this check in a backwards compatible way where
   # the equality of the two values is checked, and then the equality of current with desired
@@ -388,6 +427,12 @@ class Puppet::Property < Puppet::Parameter
   # @return [Boolean] whether the {array_matching} mode is set to `:all` or not
   def match_all?
     self.class.array_matching == :all
+  end
+
+  # @return [Boolean] whether the property is marked as idempotent for the purposes
+  #   of calculating corrective change.
+  def idempotent?
+    self.class.idempotent
   end
 
   # @return [Symbol] the name of the property as stated when the property was created.
--- a/lib/puppet/transaction/event.rb
+++ b/lib/puppet/transaction/event.rb
@@ -11,8 +11,8 @@ class Puppet::Transaction::Event
   include Puppet::Util::Logging
   include Puppet::Network::FormatSupport
 
-  ATTRIBUTES = [:name, :resource, :property, :previous_value, :desired_value, :historical_value, :status, :message, :file, :line, :source_description, :audited, :invalidate_refreshes, :redacted]
-  YAML_ATTRIBUTES = %w{@audited @property @previous_value @desired_value @historical_value @message @name @status @time @redacted}.map(&:to_sym)
+  ATTRIBUTES = [:name, :resource, :property, :previous_value, :desired_value, :historical_value, :status, :message, :file, :line, :source_description, :audited, :invalidate_refreshes, :redacted, :corrective_change]
+  YAML_ATTRIBUTES = %w{@audited @property @previous_value @desired_value @historical_value @message @name @status @time @redacted @corrective_change}.map(&:to_sym)
   attr_accessor *ATTRIBUTES
   attr_accessor :time
   attr_reader :default_log_level
@@ -28,6 +28,7 @@ class Puppet::Transaction::Event
   def initialize(options = {})
     @audited = false
     @redacted = false
+    @corrective_change = false
 
     set_options(options)
     @time = Time.now
@@ -45,6 +46,7 @@ class Puppet::Transaction::Event
     @time = data['time']
     @time = Time.parse(@time) if @time.is_a? String
     @redacted = data.fetch('redacted', false)
+    @corrective_change = data['corrective_change']
   end
 
   def to_data_hash
@@ -58,11 +60,13 @@ class Puppet::Transaction::Event
       'name' => @name,
       'status' => @status,
       'time' => @time.iso8601(9),
-      'redacted' => @redacted
+      'redacted' => @redacted,
+      'corrective_change' => @corrective_change,
     }
   end
 
   def property=(prop)
+    @property_instance = prop
     @property = prop.to_s
   end
 
@@ -92,6 +96,23 @@ class Puppet::Transaction::Event
 
   def to_yaml_properties
     YAML_ATTRIBUTES & super
+  end
+
+  # Calculate and set the corrective_change parameter, based on the old_system_value of the property.
+  # @param [Object] old_system_value system_value from last transaction
+  # @return [bool] true if this is a corrective_change
+  def calculate_corrective_change(old_system_value)
+    # Only idempotent properties, and cases where we have an old system_value
+    # are corrective_changes.
+    if @property_instance.idempotent? &&
+       !@property_instance.sensitive &&
+       !old_system_value.nil?
+
+      # If the values aren't insync, we have confirmed a corrective_change
+      @corrective_change = !@property_instance.insync_values?(old_system_value, previous_value)
+    else
+      @corrective_change = false
+    end
   end
 
   private
--- a/lib/puppet/transaction/persistence.rb
+++ b/lib/puppet/transaction/persistence.rb
@@ -1,0 +1,81 @@
+require 'yaml'
+require 'puppet/util/yaml'
+
+# A persistence store implementation for storing information between
+# transaction runs for the purposes of information inference (such
+# as calculating corrective_change).
+# @api private
+class Puppet::Transaction::Persistence
+  def initialize
+    @old_data = {}
+    @new_data = {"resources" => {}}
+  end
+
+  # Obtain the full raw data from the persistence store.
+  # @return [Hash] hash of data stored in persistence store
+  def data
+    @old_data
+  end
+
+  # Retrieve the system value using the resource and parameter name
+  # @param [String] resource_name name of resource
+  # @param [String] param_name name of the parameter
+  # @return [Object,nil] the system_value
+  def get_system_value(resource_name, param_name)
+    if !@old_data["resources"].nil? &&
+       !@old_data["resources"][resource_name].nil? &&
+       !@old_data["resources"][resource_name]["parameters"].nil? &&
+       !@old_data["resources"][resource_name]["parameters"][param_name].nil?
+      @old_data["resources"][resource_name]["parameters"][param_name]["system_value"]
+    else
+      nil
+    end
+  end
+
+  def set_system_value(resource_name, param_name, value)
+    @new_data["resources"] ||= {}
+    @new_data["resources"][resource_name] ||= {}
+    @new_data["resources"][resource_name]["parameters"] ||= {}
+    @new_data["resources"][resource_name]["parameters"][param_name] ||= {}
+    @new_data["resources"][resource_name]["parameters"][param_name]["system_value"] = value
+  end
+
+  # Load data from the persistence store on disk.
+  def load
+    filename = Puppet[:transactionstorefile]
+    unless Puppet::FileSystem.exist?(filename)
+      return
+    end
+    unless File.file?(filename)
+      Puppet.warning("Persistence file #{filename} is not a file, ignoring")
+      return
+    end
+
+    result = nil
+    Puppet::Util.benchmark(:debug, "Loaded transaction store file") do
+      begin
+        result = Puppet::Util::Yaml.load_file(filename)
+      rescue Puppet::Util::Yaml::YamlLoadError => detail
+        Puppet.err "Transaction store file #{filename} is corrupt (#{detail}); replacing"
+
+        begin
+          File.rename(filename, filename + ".bad")
+        rescue
+          raise Puppet::Error, "Could not rename corrupt transaction store file #{filename}; remove manually", detail.backtrace
+        end
+      end
+    end
+
+    unless result.is_a?(Hash)
+      Puppet.err "Transaction state file #{filename} is valid YAML but not returning a hash. Check the file for corruption, or remove it before continuing."
+      return
+    end
+
+    @old_data = result
+  end
+
+  # Save data from internal class to persistence store on disk.
+  def save
+    Puppet::Util::Yaml.dump(@new_data, Puppet[:transactionstorefile])
+  end
+end
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -122,6 +122,11 @@ class Puppet::Transaction::Report
   #
   attr_reader :noop
 
+  # @!attribute [r] corrective_change
+  #   @return [Boolean] true if the report contains any events and resources that had
+  #      corrective changes.
+  attr_reader :corrective_change
+
   def self.from_data_hash(data)
     obj = self.allocate
     obj.initialize_from_hash(data)
@@ -184,6 +189,7 @@ class Puppet::Transaction::Report
   # @api private
   def finalize_report
     prune_internal_data
+    calculate_report_corrective_change
 
     resource_metrics = add_metric(:resources, calculate_resource_metrics)
     add_metric(:time, calculate_time_metrics)
@@ -215,6 +221,7 @@ class Puppet::Transaction::Report
     @status = 'failed' # assume failed until the report is finalized
     @noop = Puppet[:noop]
     @noop_pending = false
+    @corrective_change = false
   end
 
   # @api private
@@ -229,6 +236,7 @@ class Puppet::Transaction::Report
     @noop_pending = data['noop_pending']
     @host = data['host']
     @time = data['time']
+    @corrective_change = data['corrective_change']
 
     if master_used = data['master_used']
       @master_used = master_used
@@ -288,10 +296,10 @@ class Puppet::Transaction::Report
       'noop_pending' => @noop_pending,
       'environment' => @environment,
       'master_used' => @master_used,
-
       'logs' => @logs,
       'metrics' => @metrics,
       'resource_statuses' => @resource_statuses,
+      'corrective_change' => @corrective_change,
     }
   end
 
@@ -385,6 +393,13 @@ class Puppet::Transaction::Report
   end
 
   private
+
+  # Mark the report as corrective, if there are any resource_status marked corrective.
+  def calculate_report_corrective_change
+    @corrective_change = resource_statuses.any? do |name, status|
+      status.corrective_change
+    end
+  end
 
   def calculate_change_metric
     resource_statuses.map { |name, status| status.change_count || 0 }.inject(0) { |a,b| a+b }
--- a/lib/puppet/transaction/resource_harness.rb
+++ b/lib/puppet/transaction/resource_harness.rb
@@ -10,6 +10,7 @@ class Puppet::Transaction::ResourceHarness
 
   def initialize(transaction)
     @transaction = transaction
+    @persistence = transaction.persistence
   end
 
   def evaluate(resource)
@@ -92,6 +93,25 @@ class Puppet::Transaction::ResourceHarness
     end
 
     capture_audit_events(resource, context)
+    persist_system_values(resource, context)
+  end
+
+  # We persist the last known values for the properties of a resource after resource
+  # application.
+  # @param [Puppet::Type] resource resource whose values we are to persist.
+  # @param [ResourceApplicationContent] context the application context to operate on.
+  def persist_system_values(resource, context)
+    param_to_event = {}
+    context.status.events.each do |ev|
+      param_to_event[ev.property] = ev
+    end
+
+    context.system_value_params.each do |pname, param|
+      @persistence.set_system_value(resource.name, pname.to_s,
+                                    new_system_value(param,
+                                                     param_to_event[pname.to_s],
+                                                     @persistence.get_system_value(resource.name, pname.to_s)))
+    end
   end
 
   def sync_if_needed(param, context)
@@ -138,6 +158,7 @@ class Puppet::Transaction::ResourceHarness
       raise
     ensure
       if event
+        event.calculate_corrective_change(@persistence.get_system_value(context.resource.name, param.name.to_s))
         context.record(event)
         event.send_log
         context.synced_params << param.name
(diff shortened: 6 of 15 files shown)


kbarber:
Replaces #5108

lib/puppet/transaction/event.rb
```diff
@@ -94,6 +98,28 @@ def to_yaml_properties
     YAML_ATTRIBUTES & super
   end
 
+  # Calculate and set the corrective_change parameter, based on the old_system_value of the property.
+  # @param [Object] old_system_value system_value from last transaction
+  # @return [bool] true if this is a corrective_change
+  def calculate_corrective_change(old_system_value)
+    # Only idempotent properties, and cases where we have an old system_value
+    # are corrective_changes.
+    @corrective_change = if @property_instance.idempotent? &&
```
adrienthebo: This combines an `if` expression that has side effects and uses the return value, which is confusing. I would prefer something like this:

```ruby

@corrective_change = @property_instance.idempotent? && !@property_instance.sensitive && !old_system_value.nil?

if @corrective change
  # ...
else
  false
end
```

lib/puppet/transaction/persistence.rb
```diff
@@ -0,0 +1,81 @@
...
+    unless Puppet::FileSystem.exist?(filename)
+      return
+    end
+    unless File.file?(filename)
+      Puppet.warning("Persistence file #{filename} is not a file, ignoring")
+      return
+    end
+
+    result = nil
+    Puppet::Util.benchmark(:debug, "Loaded transaction") do
```
adrienthebo: `"Loaded transaction"` - what transaction?

lib/puppet/transaction/persistence.rb
```diff
@@ -0,0 +1,81 @@
...
+        begin
+          File.rename(filename, filename + ".bad")
+        rescue
+          raise Puppet::Error, "Could not rename corrupt #{filename}; remove manually", detail.backtrace
+        end
+      end
+    end
+
+    unless result.is_a?(Hash)
+      Puppet.err "State got corrupted"
```
adrienthebo: I've seen this error message pop up when using Puppet and I've found it really confusing; can we indicate that the persistence file itself was corrupted?

lib/puppet/transaction/report.rb
```diff
@@ -386,6 +395,16 @@ def self.default_format
 
   private
 
+  # Mark the report as corrective, if there are any resource_status marked corrective.
+  def calculate_report_corrective_change
```
adrienthebo: :+1:

lib/puppet/transaction/resource_harness.rb
```diff
@@ -92,6 +93,21 @@ def perform_changes(resource, context)
     end
 
     capture_audit_events(resource, context)
+    persist_system_values(resource, context)
+  end
+
+  def persist_system_values(resource, context)
```
adrienthebo: A system_value is actual system state after resource application?

lib/puppet/transaction/event.rb
kbarber: Okay, but @corrective_change doesn't equal the first part of the `if` clause, its equal to the insync? result further down. I can rework this though, to not use the return value of the `if` easily enough.

lib/puppet/transaction/persistence.rb
kbarber: Can do.

lib/puppet/transaction/persistence.rb
kbarber: Should be loaded transaction state file.

lib/puppet/transaction/resource_harness.rb
kbarber: Yes, thats basically it.

kbarber: I forgot a docstring here, I'll add it.

kbarber:
@adrienthebo amended based on your comments.

lib/puppet/transaction/report.rb
kbarber: Cool bananas.

lib/puppet/transaction/report.rb
```diff
@@ -1,5 +1,6 @@
 require 'puppet'
 require 'puppet/indirector'
+require 'puppet/transaction/persistence'
```
adrienthebo: Small note: I don't think we actually use the `P::T::Persistence` in this file.

lib/puppet/resource/catalog.rb
```diff
@@ -214,6 +219,7 @@ def alias(resource, key)
   # @api public
   def apply(options = {})
     Puppet::Util::Storage.load if host_config?
+    persistence.load if host_config?
```
adrienthebo: I see why we're loading the persistence file here, but I don't like it. We're adding in the concept of persistence to the catalog which doesn't need to know about it; it just happens that `Puppet::Resource::Catalog#apply` happens to manage the lifetime of the transaction (and knows that we're a host catalog) and the `Storage` class was loaded and saved here, but really neither of these are related to the catalog. I would rather have this moved to `Puppet::Transaction#evaluate` to manage the persistence object lifetime; at that point `Catalog#host_config?` should still be accessible and we won't have to have the persistence file managed so far way from where it's used.

lib/puppet/transaction/event.rb
```diff
@@ -94,6 +98,30 @@ def to_yaml_properties
...
+  def calculate_corrective_change(old_system_value)
+    # Only idempotent properties, and cases where we have an old system_value
+    # are corrective_changes.
+    if @property_instance.idempotent? &&
+       !@property_instance.sensitive &&
+       !old_system_value.nil?
+
+      # Here we duplicate the property to avoid side-effects, and modify its should value
+      # so we can use insync?
+      test_property = @property_instance.dup
```
adrienthebo: This scares the crap out of me. The RAL isn't intended to duplicate properties like this, and while this is possible it's working against how things are built. My biggest concern is that I can't predict how this will behave but I do suspect that this might not be safe.

/cc @branan

lib/puppet/transaction/report.rb
kbarber: Cool fixed.

lib/puppet/resource/catalog.rb
kbarber: No worries, fixed.

lib/puppet/transaction/event.rb
kbarber: So its a squirrely one. I've incidentally now pushed this logic into the property so we can see it on its own. Let me point at that code to discuss this further.

lib/puppet/property.rb
```diff
@@ -346,6 +362,21 @@ def insync?(is)
...
+  # should value. This works around the requirement for corrective_change analysis
+  # that requires two older values to be compared with the properties potentially
+  # custom insync? code.
+  #
+  # @param [Object] orig_should the original should value
+  # @param [Object] orig_is the original is value
+  # @return [Boolean] whether or not the values are in sync or not
+  # @api private
+  def insync_values?(orig_should, orig_is)
+    test_property = self.dup
```
kbarber: @adrienthebo @branan so here is the code in question. I've moved this into `property.rb` from `event.rb`, because its weird and weirdness like this probably belongs closer to the problem.

So ... following on from the previous PR comment ... this methodology provides results, but as you say you are concerned with the side-effects of a dup.

Let me begin by outlining our limitations. Today the insync? comparison is my best method for proving things are 'equal' or 'insync' after the fact. Without it, ensure properties won't match for example (say `present => absent` on a file fails because its really `file => absent`).

However the API of `#insync?` dictates you must set the `@should` instance variable (in this case via a `#should=` method) to do so. Unfortunately, `#insync?` is an abstract method often overwritten by custom types, and so an overridden method expects this as its API to work with. To point, we're limited to what we can change because it will break custom extensions.

As far as other ideas and implementations to achieve similar results ... I've incidentally tried another methodology, whereby I backed up the old value, do an insync? and restore the value, this at the time produced odd results that I didn't understand (as in, it was marking things out of sync when they were not). Unless you have a better idea, I can resurrect that methodology ^^ for investigation.

lib/puppet/type/notify.rb
```diff
@@ -6,7 +6,7 @@ module Puppet
   Type.newtype(:notify) do
     @doc = "Sends an arbitrary message to the agent run-time log."
 
-    newproperty(:message) do
+    newproperty(:message, :idempotent => false) do
```
Iristyle: Is it better to modify the underlying property like you have you done here, or would it be less disruptive to provide a filter / whitelist somewhere later?

lib/puppet/property.rb
branan: I'm also not super thrilled with this, but I agree with your reasoning, Ken :(

If we were dup'ing a property in a more complex situation I'd be a lot more scared, but given that this is tiny function and the lifetime of the dup'd property is easy to understand, I think it's OK-ish. Just don't let it escape into the wild where somebody might call #sync on it, or we're gonna have a situation

lib/puppet/type/notify.rb
kbarber: This has been debated both ways, @adrienthebo had liked this methodology I think. Allowing someone to mark something via the type interface does have the benefit of re-usability, and in some ways its neater because of that, but we don't know who needs it yet TBH, so this could be YAGNI. One could argue things in the wild might have known idempotency issues, but a flag to hide that might not be what is desired I guess.

There has been some debating of also filtering exec, but I'm unsure if the same implementation applies in this case, since we wouldn't filter because its inherently non-idempotent but for other boring reasons I can get into if you're curious.

lib/puppet/property.rb
kbarber: Lol. Its a pita. I can always right a dpittman-style essay on why its bad in the comments :-).

I'll try to do the other methodology, see how far I get, at least try to understand why it doesn't work. Having it contained like you state feels better than doing it from-a-distance at least.

adrienthebo: I would be happy for a dpittman essay, so it's clear to anyone that sees this that we don't want this behavior to replicate.

kbarber: if its contained, and the API is good enough, we can fix the implementation presumably.

+1 on big fat comment, if we're up for a type refactor in the future - people will want reasons.

Also, unit tests. I was waiting for stability of structure.

lib/puppet/type/notify.rb
kbarber: @Iristyle have you talked me out of this implementation? This is public API, and more knobs without reason suck. @adrienthebo, would a whitelist style implementation suck? Can we do something like this, and make it private I wonder? Being able to mark idempotency issues for other custom types really just fell out of this effort, its not essential to success (accidental feature?), at least I'm not entirely sure it is. We can always add it in, when its called for?

adrienthebo: If we do a whitelist then that's probably going to have to live in core; with an idempotent property that can be done on a per-custom type basis which is nice. Also, this implementation works (it works, right?) and we're under pressure to ship, so I would rather just take this as-is and roll with it. If this is YAGNI then it's very small YAGNI with limited scope and minimal impact.

lib/puppet/transaction/resource_harness.rb
```diff
@@ -92,6 +93,25 @@ def perform_changes(resource, context)
...
 
     capture_audit_events(resource, context)
+    persist_system_values(resource, context)
+  end
+
+  # We persist the last known values for the properties of a resource after resource
+  # application.
+  # @param [Puppet::Type] resource resource whose values we are to persist.
+  # @param [ResourceApplicationContent] context the application context to operate on.
+  def persist_system_values(resource, context)
```
Iristyle: I assume this will persist the actual file contents in a situation like this?

```puppet
file { '/tmp/foo': content => 'really_long_string', }
```

Iristyle: Also - with respect to files... if the upstream source is a `puppet://` style URI -- and the backing contents of that file change, is that classified as a `corrective_change` or is it expected?

lib/puppet/transaction/event.rb
```diff
@@ -58,11 +60,13 @@ def to_data_hash
...
       'status' => @status,
       'time' => @time.iso8601(9),
-      'redacted' => @redacted
+      'redacted' => @redacted,
+      'corrective_change' => @corrective_change,
     }
   end
 
   def property=(prop)
+    @property_instance = prop
```
Iristyle: There might have been a reason that the property object itself was never cached previously.  Not sure if @joshcooper might know why.

lib/puppet/transaction/report.rb
Iristyle: Can be shortened to `resource_statuses.any?`

lib/puppet/transaction/persistence.rb
```diff
@@ -0,0 +1,81 @@
+require 'yaml'
+require 'puppet/util/yaml'
+
+# A persistence store implementation for storing information between
+# transaction runs for the purposes of information inference (such
+# as calculating corrective_change).
+# @api private
+class Puppet::Transaction::Persistence
```
Iristyle: The last run report can't be extended to meet your needs I presume (I'm not entirely familiar with the difference b/w this new YAML file and what we already have)

lib/puppet/type/notify.rb
Iristyle: You raise a good point @adrienthebo - without a mechanism for extending the whitelist, type and provider builders can't opt into the "non-idempotent" property behavior. Hmm...

lib/puppet/transaction/resource_harness.rb
kbarber: No, it stores the "shortened" version, depending on the provider property this can be different, for file its the md5.

> Also - with respect to files... if the upstream source is a puppet:// style URI -- and the backing contents of that file change, is that classified as a corrective_change or is it expected?

So this all translates to content property actions. In this case its much like setting the content field, so it won't be a corrective_change, its intended.

lib/puppet/transaction/persistence.rb
kbarber: We went down that road also, kylo and eric0 had concerns about extending the report format this way, when we had no intention of using it directly (ie. PuppetDB wasn't going to consume these system values, nor was it ready for that). Since its a rather public API, we decided it best to avoid that kind of thinking for later.

lib/puppet/transaction/report.rb
kbarber: @Iristyle good thinking, done.

lib/puppet/type/notify.rb
Iristyle: Thinking about this a little more... what happens when modules you use trigger errant `corrective_change` events?  And the module is not being updated by the maintainer, etc.

lib/puppet/transaction/resource_harness.rb
Iristyle: Ah, thanks for clarifying.

lib/puppet/transaction/persistence.rb
Iristyle: Ah, so you've already been chatting with those guys on this - great.  That reasoning all makes sense.

Sorry I'm a bit late to the party then - carry on!

lib/puppet/type/notify.rb
kbarber: So bugs that cause idempotency issues will show as corrective_change's today, this is a known limitation. Also some DSL code can do this also, like services with incorrect status handling and no status => overrides.

The future, would hopefully be able to bring detecting differences between bugs and true corrective_changes, but we were trying to avoid boiling the ocean for now.

If you're interested I have some points on next steps, I've put these forward as phase 2 project ideas basically to the project team.

kbarber: So yeah, I've already asked about putting this point in some sort of errata/known issues for the release notes fwiw, most probably as a leading ticket as well people can vote/watch. Mainly, so its less surprising/documented somewhere.

lib/puppet/transaction/persistence.rb
kbarber: Its cool, its nice to know I wasn't the only one to consider this path :-). But yeah, devil in the details.

lib/puppet/type/notify.rb
Iristyle: 👍 sounds like you have a staged plan of attack that's reasonable.

I'll see if I can come up with situations where the addition of `idempotent` to a property could break module authors - at the very least it might be worth a quick grep through the modules for collisions (seems unlikely to dig anything up).

joshcooper:
I'm joining this conversation in the middle, apologies if I'm asking questions that have already been answered elsewhere.

1. Are we're reinventing `audit`functionality, keeping track of old values so that we can compare against current values and infer if it was changed by something other than puppet in-between? Guess I'm a bit surprised because we've been talking about deprecating audit for a long time: https://tickets.puppetlabs.com/browse/PUP-893.
1. Does it make sense to mark the entire resource as idempotent or not, e.g. `notify`, `exec`, instead of individual properties?
1. Would it make sense to create an `idempotent` metaparam like `audit` instead of modifying the type definitions? That way users (not module authors) can opt-in/out easily? Or a combination of both (define default behavior in the type and allow opt-in/out in manifest)?
1. `Puppet::Transaction::Persistence` behaves just like `Puppet::Util::Storage`, plus adds getter/setter methods for the data you care about. Why not use `Storage` to handle the persistence and have your class provide methods for manipulating the "subtree" of yaml data you care about?
1. Since the report schema is changing, we should bump the report version/format
1. I didn't follow the comments in https://github.com/puppetlabs/puppet/pull/5139/files#diff-93ff56b1885dd711e694649d0cad0b77R374. Why do we need to create a temp property? Does the `dup` correctly handle structured values, I'm thinking of the Windows ACL module specifically.

adrienthebo:
I can answer some of these questions.

1. We're partially reimplementing audit, but we're doing so because audit caches the historical state before Puppet makes changes; we need the historical state cached after Puppet makes changes. (Right, @kbarber?)
2. I don't think so; exec can be idempotent when configured correctly. Conversely, service resources have had a history of becoming non-idempotent (specifically the `enable` property when using systemd with a legacy init script), but the resource itself should be idempotent.
3. Maybe? Allowing resources to be easily marked as non-idempotent feels like we're enabling/supporting misbehaving types on an ad-hoc basis, whereas with notify the intended behavior is non-idempotent.
4. `P::T::Persistence` is based on `P::Util::Storage`, but the main difference is that `Storage` doesn't discard old information from resources that were removed and will bloat forever, while `Persistence` throws away information about resources that were removed.
5. We've already bumped the report format version for the 4.6.0 release for a couple of other features, so we don't need to bump it again.
6. We're using `Puppet::Property#insync?` to detect drift from a recorded historical value. Because `#insync` relies on `@should` we can't just call that method with a new `@should` value; we have to replace the value entirely. This issue is the reason why I wanted to bring more people into this conversation; I can't say for certain that this will be safe. It might be, but it also might break all sorts of things depending on what people have done with properties. My understanding is that `#dup` just copies all of the instance variables to a new object, and will call `#initialize_copy` if it's defined, but that should be the scope of the operation.

kbarber:
1. See Adrien's response. Audit doesn't store the right thing, but it was so close. I looked into the audit file, and almost was convinced, but it has issues and others talked me out of using it also: a) it had a hard API to change file-wise (values assigned directly to keys, I couldn't just add my new value) b) its semantics are more around making small transactional changes to a file c) it has no cleanup, just yesterday I had a guy on IRC complaining, and my advice was to stop using it basically d) everyone in the team I spoke to had issues with it, so it felt like beating a dead horse to improve it. So basically, yes, its close but not quite to audit, however if you wanted to use this instead of audit its possible, we architecturally have left that an open door for future conversations. In a way though, philosophically, this entire functionality kind of does replace some of audit's/inspects goals with a slightly more modern flair perhaps, so if/when that conversation happens about replacing audit, we might want to take this into account. Great question btw.
2. You make an interesting point re: property v resource. Let me ponder it.
3. Nice idea. Its a maybe/yes for me too. We can also add that feature too later right?
4. See Adrien's response.
5. Ditto.
6. I added unit tests to property_spec.rb to test this method against the same stuff insync? is tested with. There are some arrays in there, but we can go deeper structure if you're concerned no problemo. The reason I dup, is because I can't get setting a should to a value working without a dup, it seemed like a reasonable alternative but everyone dislikes it. I've been trying to figure out why just forcing should to something, do the insync? then restoring should doesn't work, but it makes no sense. Perhaps if this is a critical thing (and please, say so) I could reproduce it and seek help from someone. At the very least, we moved this weirdness into its own place to contain it.

@joshcooper cheers :-).

lib/puppet/type/notify.rb
Iristyle: FYI - I just threw `:idempotent => false` on some properties in existing types on the existing codebase, and ran the spec suite (just to see if anything would break).  I hoping this serves as a proxy for newer modules opting to use `:idempotent` running on an older Puppet where that concept doesn't exist.

````
diff --git a/lib/puppet/type/exec.rb b/lib/puppet/type/exec.rb
index f2526ba..11673e5 100644
--- a/lib/puppet/type/exec.rb
+++ b/lib/puppet/type/exec.rb
@@ -64,7 +64,7 @@ module Puppet
       @checks.keys
     end

-    newproperty(:returns, :array_matching => :all, :event => :executed_command) do |property|
+    newproperty(:returns, :array_matching => :all, :event => :executed_command, :idempotent => false) do |property|
       include Puppet::Util::Execution
       munge do |value|
         value.to_s
diff --git a/lib/puppet/type/file/content.rb b/lib/puppet/type/file/content.rb
index 614d9d7..9c98a58 100644
--- a/lib/puppet/type/file/content.rb
+++ b/lib/puppet/type/file/content.rb
@@ -6,7 +6,7 @@ require 'puppet/util/checksums'
 require 'puppet/type/file/data_sync.rb'

 module Puppet
-  Puppet::Type.type(:file).newproperty(:content) do
+  Puppet::Type.type(:file).newproperty(:content, :idempotent => false) do
     include Puppet::Util::Checksums
     include Puppet::DataSync

diff --git a/lib/puppet/type/package.rb b/lib/puppet/type/package.rb
index cab3c67..2b3e159 100644
--- a/lib/puppet/type/package.rb
+++ b/lib/puppet/type/package.rb
@@ -296,7 +296,7 @@ module Puppet
       [ [ /(.*)/m, [ [:name] ] ] ]
     end

-    newproperty(:package_settings, :required_features=>:package_settings) do
+    newproperty(:package_settings, :required_features=>:package_settings, :idempotent => false) do
       desc "Settings that can change the contents or configuration of a package.

         The formatting and effects of package_settings are provider-specific; any
diff --git a/lib/puppet/type/scheduled_task.rb b/lib/puppet/type/scheduled_task.rb
index 97215a2..1d6eb96 100644
--- a/lib/puppet/type/scheduled_task.rb
+++ b/lib/puppet/type/scheduled_task.rb
@@ -54,7 +54,7 @@ Puppet::Type.newtype(:scheduled_task) do
       should be specified as a space-separated string."
   end

-  newproperty(:user) do
+  newproperty(:user, :idempotent => false) do
     desc "The user to run the scheduled task as.  Please note that not
       all security configurations will allow running a scheduled task
       as 'SYSTEM', and saving the scheduled task under these
diff --git a/lib/puppet/type/service.rb b/lib/puppet/type/service.rb
index 1f8d266..8a5508c 100644
--- a/lib/puppet/type/service.rb
+++ b/lib/puppet/type/service.rb
@@ -122,7 +122,7 @@ module Puppet
       end
     end

-    newproperty(:flags, :required_features => :flaggable) do
+    newproperty(:flags, :required_features => :flaggable, :idempotent => false) do
       desc "Specify a string of flags to pass to the startup script."
     end

diff --git a/lib/puppet/type/user.rb b/lib/puppet/type/user.rb
index 3c1156e..78c9f36 100644
--- a/lib/puppet/type/user.rb
+++ b/lib/puppet/type/user.rb
@@ -98,7 +98,7 @@ module Puppet
       end
     end

-    newproperty(:home) do
+    newproperty(:home, :idempotent => false) do
       desc "The home directory of the user.  The directory must be created
         separately and is not currently checked for existence."
     end
```

All the specs passed on Windows:

```
Finished in 11 minutes 58 seconds (files took 15.14 seconds to load)
21219 examples, 0 failures, 62 pending
```

kbarber:
2. (after pondering some more) it should be the per property at the very least. In pondering this, I could see puppetlabs-firewall (which is chock full of properties) having idempotency issues that are specific to say 1 property, but the others are fine (happened a lot due to its shoddy parser, sigh). However, I could see convenience added to an entire resource as sugar at any point in the future, if it became worth it.

lib/puppet/transaction/event.rb
kbarber: Be interesting to know, I was careful to have this avoid the YAML serialization at least. Its very easy to accidentally drop crap into the report for example without realizing it :-).

lib/puppet/transaction/resource_harness.rb
kbarber: I was worried about space too. Our original plan was to do this all in PuppetDB, but we only get the catalog which contains the full string, so it would have been hard to get the md5 instead etc. Keeping it in the agent, we benefit from having these short-forms in hand, and its native comparison capability also, which makes the solution accurate (well as accurate as insync?).

lib/puppet/type/notify.rb
kbarber: Cool.

adrienthebo:
My understanding is that we're under a time crunch for this feature and that it's blocking a lot of other work, so I want to push this PR forward. While there are some outstanding concerns about the use of `Puppet::Property#dup`, I don't know of any specific case where this is guaranteed to break. @joshcooper @Iristyle if either of you know of a case where this could cause issues, we need to know very soon.

I see that [puppetlabs-acl](https://github.com/puppetlabs/puppetlabs-acl/blob/master/lib/puppet/type/acl/ace.rb#L24-L44) does have some nontrivial `#initialize` behavior, but since `#dup` should only copy internal variable state and shouldn't initialize the object I think we're okay. For the other properties I've seen with some custom initialization logic it's nothing too dangerous, so all in all I think we can get away with this.

While it would be nice to have a feature flag to allow this feature to be turned off if needed I don't think we have the time to build that in, so it's kind of an all or nothing proposition.

In general, unless anybody has a firm reason to reject this PR or halt it for a specific reason, I think we should merge this. We should definitely revisit this after the release and address some of the outstanding concerns, but for the sake of getting this work in so that other work can complete I'm :+1: for merge. I'm currently at a conference so I don't feel comfortable merging this myself, but @branan if you could do this or get someone else to merge it and shepard it for CI, I'll buy you lunch when I get back.

/cc @kbarber @scottyw @ploubser

api/schemas/report.json
```diff
@@ -160,6 +160,11 @@
         "environment": {
             "description": "The name of the environment that was used for the puppet run (e.g. \"production\").",
             "type":        "string"
+        },
+
+        "corrective_change": {
+            "description": "True if this report suffered an unexpected change, and Puppet corrected it, or was intending to correct it if the resource or entire run was noop.",
```
branan: How about "True if a change or noop event was caused by an unexpected change to the system between Puppet runs"?

lib/puppet/property.rb
```diff
@@ -102,6 +102,22 @@ def array_matching=(value)
       raise ArgumentError, "Supported values for Property#array_matching are 'first' and 'all'" unless [:first, :all].include?(value)
       @array_matching = value
     end
+
+    # Used to mark a type property as having or lacking idempotency (on purpose
+    # generally). This is used to avoid marking the property as a
+    # corrective_change when there is known idempotency issues with the property
+    # rendering a corrective_change flag as useless.
+    # @return [Boolean] true if the property is marked as non-idempotent
```
branan: "Returns true if the property is non-idempotent" sounds backwards. Is that just a docs issue?

lib/puppet/property.rb
```diff
@@ -346,6 +362,29 @@ def insync?(is)
     end
   end
 
+  # This method tests if two values are insync? outside of the properties current
+  # should value. This works around the requirement for corrective_change analysis
+  # that requires two older values to be compared with the properties potentially
+  # custom insync? code.
+  #
+  # @param [Object] orig_should the original should value
```
branan: The phrasing of "original" here definitely shows its heritage from corrective_change. These should probably just be "should" and "is", and the documentation made more generic. This can be a post-merge cleanup, IMO

lib/puppet/type/notify.rb
kylog: Fwiw, I'm 👍 on the use of `:idempotent` here as opposed to the whitelist approach. And thanks for that experiment @Iristyle - it seems like a reasonable proxy.

api/schemas/report.json
kbarber: Fixed.

lib/puppet/property.rb
kbarber: Fixed.

lib/puppet/property.rb
kbarber: Fixed.

lib/puppet/transaction/event.rb
kylog: I spelunked back to 2009 on this (https://github.com/puppetlabs/puppet/commit/977595bd if you're curious) and I didn't find a reason why the property object itself wasn't cached. My *guess* is that it was following a pattern of "normalize things to strings" and there just hadn't been a reason before. But dunno.

That said, I'm not coming up with a reason adding this would cause problems.

lib/puppet/transaction/event.rb
```diff
@@ -94,6 +98,25 @@ def to_yaml_properties
...
+       !@property_instance.sensitive &&
+       !old_system_value.nil?
+
+      # If the values aren't insync, we have confirmed a corrective_change
+      @corrective_change = !@property_instance.insync_values?(old_system_value, previous_value)
+    else
+      @corrective_change = false
+    end
+
+    return @corrective_change
```
kylog: I think the `return` can be dropped. Actually I think the line can be dropped.

kbarber: fixed.

joshcooper:
What happens when the agent changes environment? Do we still report corrective changes between the previous run (that happened in a different environment)? Does that make sense?

Related, do we persist previous system values during a noop run? If so that seems bad, because it will lead to false alerts. I don't think it blocks this PR, but seems like something we should fix.

kylog:
@kbarber total nit, but can you amend the commit message to be in the imperative, e.g. "Add corrective_change field to report and report events" or some such. Buried in https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md is the request to make summaries imperative.

kbarber:
@joshcooper if puppet changes environment, its fine. We compare the last set values to the value we get from a `#retrieve`, we don't care what puppet is changing things to, so I'm not sure how an environment change could affect things against this patch.

During a noop run, we persist the old stored system value, and throw a correct_change. Thus we still get a corrective_change for each noop run, until its finally corrected and the new set value is set in the system_value.

kbarber:
@kylog done.

kylog:
I'm 👍 on this PR as-is. There are several good follow-ups in the extensive commenting above; it'd be nice to pull actionable items out into the appropriate places (probably tickets or docs).

branan:
I agree with Kylo. Let's get actionable stuff someplace where we can track it, and this is fine to merge at this point

kbarber:
I'm sorry, I can't tell what's important to people or not any more. What do people want me to document or raise tickets on?

kylog:
Oh good, travis sprung back (it seemed to have lost track of the PR for a bit). Once travis and appveyor flip green, I'll merge.

joshcooper:
> During a noop run, we persist the old stored system value, and throw a correct_change. Thus we still get a corrective_change for each noop run:

Sorry, what I mean is, it seems incorrect to call [`persistence.save`](https://github.com/puppetlabs/puppet/pull/5139/files#diff-129f1c7cf7c3e8df7a5586dde253d845R148) during a noop run as that will alter the stored system values. So when the agent next runs (not in noop), it will be confused about what the previous system value was, and won't correctly flag the corrective_change. I'm fine to merge, we can discuss separately.

kylog:
@kbarber for starters, I'm inclined to send you a bottle of your beverage of choice :) I know this PR was a gauntlet. Thanks for all your work on this.

And re action items, after lunch I'll take a pass through the PR to cull out possible follow-up items and drop another comment in the PR.

kbarber:
@joshcooper so the short answer here is, it just works? Meaning, it doesn't get confused ... but perhaps I'm not explaining the implementation well. It doesn't care about the overall noop run state true, however it behaves differently when it sees that a event was a non-success however: https://github.com/puppetlabs/puppet/pull/5139/files#diff-466a7aa2dfd3724f006c1984f9ce87acR270 ... I agree there is probably some efficiency to be found here potentially, but with the uncertainty of perhaps some resources becoming noop => false even when the entire runs is noop (I can't recall how, but I've heard this is possible) I'm inclined to think I took the safer approach for now. Oh, and this also works for failed, skipped etc.

kylog:
Okay, @kbarber, I re-read the comment thread for follow-ups and I mostly concluded that this change needs release notes. In terms of mechanics, I'd think you could just use the Release Notes Summary field in https://tickets.puppetlabs.com/browse/PUP-6295, though maybe there ends up being enough content for a pre-docs PR.

In terms of content: offhand, in addition to explaining the general idea of the `corrective_change` fields, I'd think the release notes would cover use of `:idempotent` and interaction with the new Sensitive type. Also, in https://github.com/puppetlabs/puppet/pull/5139#discussion_r72809223, Ken, you mentioned release notes errata (and a leading ticket) re conditions where this *doesn't* work.

/cc @jtappa as a heads-up that a somewhat substantive release note will be inbound for puppet 4.6.

Reviewers of this PR: if I missed some follow-ups, please chime in!

PR #5140, opened 2016-07-25
(PUP-6545) Fix tests for Ruby 2.3 on Windows in Appveyor

glennsarti:
Previously only ruby 2.1.8 was used in the the Appveyor tests.  This PR adds
Ruby 2.3.0 to the test matrix.  The complete and up to date list
of installed software can be found at;
https://www.appveyor.com/docs/installed-software

Ruby bug 8822 has been partially resolved in Ruby 2.3.0.  Therefore the test in
`spec/unit/util_spec.rb:120` is now failing.  This PR guards this expected
bad behavior test, to Ruby versions prior to 2.3.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,16 +1,22 @@
 version: 4.1.0.{build}
 clone_depth: 10
-init:
-  - SET
 install:
-  - SET PATH=C:\Ruby21-x64\bin;%PATH%
+build: off
+
+environment:
+  matrix:
+  - RUBY_VER: 23-x64
+  - RUBY_VER: 21-x64
+
+test_script:
+  - SET
+  - SET PATH=C:\Ruby%RUBY_VER%\bin;%PATH%
   - SET LOG_SPEC_ORDER=true
   - bundle install --jobs 4 --retry 2 --without development extra
   - type Gemfile.lock
   - ruby -v
-build: off
-test_script:
   - bundle exec rspec spec
+
 on_failure:
   - appveyor PushArtifact .\spec_order.txt
 notifications:
--- a/spec/unit/util_spec.rb
+++ b/spec/unit/util_spec.rb
@@ -117,7 +117,7 @@ describe Puppet::Util do
       end
     end
 
-    it "works around Ruby bug 8822 (which fails to preserve UTF-8 properly when accessing ENV)" do
+    it "works around Ruby bug 8822 (which fails to preserve UTF-8 properly when accessing ENV) (Ruby <= 2.1) ", :if => (RUBY_VERSION =~ /^(1\.|2\.0\.|2\.1\.)/) && (Puppet.features.microsoft_windows?) do
       env_var_name = SecureRandom.uuid
       utf_8_bytes = [225, 154, 160] # rune ᚠ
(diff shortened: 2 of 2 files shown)


appveyor.yml
```diff
@@ -1,16 +1,23 @@
...
-init:
-  - SET
 install:
-  - SET PATH=C:\Ruby21-x64\bin;%PATH%
+build: off
+
+environment:
+  matrix:
+  - RUBY_VER: 23-x64
+  - RUBY_VER: 22-x64
```
Iristyle: We don't need Ruby 2.2

spec/unit/util_spec.rb
```diff
@@ -150,6 +150,78 @@ def get_mode(file)
...
+        # when these tests no longer pass, Ruby has fixed its bugs and workarounds can be removed
+        # interestingly we would expect some of these tests to fail when codepage is 65001
+        # but instead the env values are in Encoding::ASCII_8BIT!
+
+        # the ENV keys are stored in a UTF8 format, therefore accessing the ENV var by UTF8 name should
+        # succeed.  However accessing the ENV var by a keyname that has been localized to the external
+        # code may not succeeded.  As we're using the rune characters in the default IBM437 codepage this
+        # will not succeed as the run characters appear mangled
+        codepage_key = utf_8_str.dup.force_encoding(Encoding.default_external)
+        expect(ENV.key?(codepage_key)).to eq(false)
```
Iristyle: If this is the only expectation difference, why not just adjust the assertion instead of duplicating the entire test?

glennsarti: I felt it made more sense to chunk the test into "this test behaves like this for ruby FOO" etc.  I can easily have one test and then use Ifs in there to same effect.  I "looks" simpler with the duplication

Iristyle:
There's something weird in 898b13e

Iristyle:
You can pull the last commit off as it's been moved into #5151.

I can't tell from the commits if 2.3 is already passing, but if not.. maybe we want to allow the fail until all the tests are sorted?
https://www.appveyor.com/docs/build-configuration#failing-strategy

glennsarti:
Commit ` 898b13e` has been dropped

kylog:
Hmm, master now has the json_pure pin, and I rekicked travis here, which I thought would pick up that change (it did for an unrelated PR targeting stable). But it seems not to have? Maybe rebase this one and see if it flips green?

Iristyle:
Double checked the Travis logs and it is grabbing `refs/pull/5140/merge` for the 1.9.3 branch, which *should* be this branch merged with stable HEAD.

That indeed includes the line `gem 'json_pure', '~> 1.8', :require => false` at https://github.com/puppetlabs/puppet/blob/stable/Gemfile#L32 ... yet the log shows bundler installing `2.0.2` ... so indeed peculiar.

glennsarti:
Can we get this merged and fix the globbing issue in PUP-6557

Iristyle:
Edited / merged to stable in https://github.com/puppetlabs/puppet/commit/603aee2652554c8befb84fdafc5873d8fde7d220

PR #5141, opened 2016-07-25
(PUP-6435) Convert sensitive params to resource metadata

adrienthebo:
Our current JSON based serialization format is limited in the data types
it can handle; this falls over for rich data types such as the Sensitive
type. This commit adds a sensitive field to resources to pass parameter
level sensitivity in a serializable manner.

### DIFF
--- a/lib/puppet/parser/resource.rb
+++ b/lib/puppet/parser/resource.rb
@@ -106,6 +106,7 @@ class Puppet::Parser::Resource < Puppet::Resource
     @finished = true
     add_defaults
     add_scope_tags
+    replace_sensitive_data
     validate if do_validate
   end
 
@@ -301,6 +302,15 @@ class Puppet::Parser::Resource < Puppet::Resource
     scope_resource = scope.resource
     unless scope_resource.nil? || scope_resource.equal?(self)
       merge_tags(scope_resource)
+    end
+  end
+
+  def replace_sensitive_data
+    parameters.each_pair do |name, param|
+      if param.value.is_a?(Puppet::Pops::Types::PSensitiveType::Sensitive)
+        @sensitive_parameters << name
+        param.value = param.value.unwrap
+      end
     end
   end
 
--- a/api/schemas/catalog.json
+++ b/api/schemas/catalog.json
@@ -51,6 +51,13 @@
                     "exported": {
                         "type": "boolean"
                     },
+                    "sensitive_parameters": {
+                        "description": "An optional list of resource parameters to be treated as sensitive.",
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
+                    },
                     "tags": {
                         "description": "Tags: regex is from https://docs.puppetlabs.com/puppet/latest/reference/lang_reserved.html",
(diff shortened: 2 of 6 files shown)


adrienthebo:
I've also considered adding a more general "metadata" key, in case that we need to pass more information in a sideband manner in case we need to support more values before the PCore based serialization is ready.

/cc @hlindberg @branan

lib/puppet/parser/resource.rb
```diff
@@ -304,6 +305,17 @@ def add_scope_tags
     end
   end
 
+  def replace_sensitive_data
+    if builtin_type?
+      parameters.each_pair do |name, param|
```
adrienthebo: Looping over every parameter for every resource might be somewhat costly; instead of doing this we could change `#set_parameter` to unwrap sensitive values on parameter assignment. Thoughts?

hlindberg: There is a better way I think. (Will make separate comment about that). Problem with unwrapping on a `set` is that puppet logic can access the unwrapped value once it has been set.

hlindberg:
I think the implementation should be in Puppet::Resource since serialization of such an instance should again turn a Sensitive instance into the same kind of data hash.  That would make to/from data hash symmetrical. Unsure if that would cause a problem anywhere.

That would also help performance wise as it would take one iteration over values in the `to_data_hash` method (an iteration that it is already doing).

In general - reading the old existing serialization logic makes me want to poke out my eyes with a fork...
* A temporary array is created for each resource: `[:type, :title, :tags] + ATTRIBUTES`, it uses a reduce with lambda and uses `send` to get attribute values
  ** Since it knows the exact 6 attributes it wants, it could call those methods directly in sequence (no loop, no temporary data). 
* The resource is transformed to a hash
* This hash is then rebaked into another hash
* For each parameter it calls the #namevar method (see code below) - it clearly only needs to do that once since the concept of namevar must be static
```
    if builtin_type? && !(t = resource_type).nil? && t.key_attributes.length == 1
      t.key_attributes.first
    else
      :name
    end
```
* I could not continue reading as it was hard to see the code through my tears...

The old logic is quirky, and must be much slower than what it can be.

adrienthebo:
The problem with implementing the conversion in the serialization method is that `puppet apply` doesn't serialize the catalog, which means the Sensitive values will be passed through as-is. In theory we could implement the transformation in the serialization method as well, but I think that we've got to do this transformation after the catalog is compiled.

branan:
In addition to what @adrienthebo said about the `apply` codepath, we also will eventually be ripping this out. Having a simple post-processing step is very clean to remove when we have the machinery in place to transfer datatype information from the compiler to the agent

branan:
@adrienthebo Do you mind rebasing this so it'll pick up the fixed json_pure gem requirement? Otherwise I'm happy with it at this point

@hlindberg I understand your desire to see this in serialization, but I'm definitely in agreement with Adrien here. Did you have any other thoughts on this before I merge it?

adrienthebo:
Rebased.

PR #5143, opened 2016-07-26
sudo support added for installing rpm via provider

ripun:
Hi Team,
we are using puppet provider "rpm" to install rpm packages on redhat based OS like RHEL/Fedora/Centos
**however we don't have  "root" access to install rpms , but sudo access we do  have** 

**Problem Statement:**
How to install rpm using puppet rpm provider and sudo unix user 

**Solution:**
so in order to overcome this usecase i have modified code to support "sudo" use case as well  without breaking current compatibility
that is code is 100% BWC

**sample example:**
_package{'libtool':
            provider => 'rpm',
            ensure => 'present',
            source => '/home/user/libtool-2.0-1.el7.x86_64.rpm',
        }_

Modified new code  will on its own decide if user is root or non root 

1. if user is root it will continue to work as it was earlier
2. if user is non-root , then command execution will be done via sudo logic 
3. if user has sudo permission then rpm will be installed else it will complain sudo privilege is required to install rpm

**P.S:** 
"Defaults   requiretty" needs to be commented in /etc/sudoers on RHEL based systems 
Redhat has acknowledged that  they  see no pros with require tty  and its a bug , please see below details
[https://bugzilla.redhat.com/show_bug.cgi?id=1350922](url)
[https://bugzilla.redhat.com/show_bug.cgi?id=1020147](url)

Defaults   requiretty  issue could be overcome by simple WA doing some ssh , but Puppet does not work with ssh but with HTTPS so its required to fix the issue in /etc/sudoers if any system does not have line "Defaults   requiretty " commented 

Please review changes and this is very useful feature and let me know if support needs to be added for other providers as well
**I will be very glad to hear your suggestions and improvements for adding sudo support to providers instead of using "root" user** 
Looking forward to this change 

**Thanks,
Ripunjay Godhani**

### DIFF
--- a/lib/puppet/provider/package/rpm.rb
+++ b/lib/puppet/provider/package/rpm.rb
@@ -48,7 +48,8 @@ Puppet::Type.type(:package).provide :rpm, :source => :rpm, :parent => Puppet::Pr
   ARCH_REGEX = Regexp.new(ARCH_LIST.join('|\.'))
 
   commands :rpm => "rpm"
-
+  commands :sudo => "sudo"
+  
   if command('rpm')
     confine :true => begin
       rpm('--version')
@@ -149,7 +150,15 @@ Puppet::Type.type(:package).provide :rpm, :source => :rpm, :parent => Puppet::Pr
     flag = ["-i"]
     flag = ["-U", "--oldpackage"] if version && (version != :absent && version != :purged)
     flag += install_options if resource[:install_options]
-    rpm flag, source
+    user = `whoami`
+    if Puppet.features.root?
+    	Puppet.debug "Exceution as root user #{user} ,#{flag}, #{source}"
+    	rpm flag, source
+    else
+    	Puppet.debug "Execution as non root sudo user  #{user} ,#{flag}, #{source}"
+    	cmd = [command(:rpm), flag,  source]
+    	sudo *cmd
+    end
   end
 
   def uninstall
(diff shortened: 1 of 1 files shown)


ripun:
Hi Team,
Also we can add a new provider called as sudo rpm if we think we dont want to merge this sudo  logic in existing rpm.rb provider

**Proposed: new provider sudorpm**
_package{'libtool':
            provider => 'sudorpm',
            ensure => 'present',
            source => '/home/user/libtool-2.0-1.el7.x86_64.rpm',
        }_
**Pros of separate provider sudorpm:**
**1.** it will support what is supposed to support that is providing support for installing rpm via sudo
**2**. existing rpm provider code will be intact and not touched

**Cons of separate provider sudorpm :**
**1.** we are not reusing the code , which is not good
**2.** BWC is maintained so it should not be a separate provider but this support should be present in provider rpm.rb

**Please suggest and advise** 

Thanks,
**Ripunjay Godhani**

kylog:
@ripun: apologies for the delay in getting back to you on this PR. I don't think puppet would want to carry and maintain a patch like this for two general reasons:
# it's specific to one provider. E.g. what if you want to manage files or services that require sudo access next? That would require sprinkling similar logic across all/many providers
# it seems to be implementing a change in puppet to workaround a site-specific permissions issue. I don't know if there's enough general utility to that scenario to warrant even a generic addition of sudo support

My best recommendation for your situation is to ask around (e.g. in slack or irc) to see how other folks have tackled your issue. Also note that if you want to write your own `sudorpm` provider, you could probably inherit from the `rpm` provider and get most of the code reuse you have here.

Last, a bit of hygiene: please submit an issue at https://tickets.puppetlabs.com/browse/PUP for any pull request to puppet and then reference that ticket in the commit messages like so `(PUP-1234) Fix a thing ...`. That helps with our bookkeeping.

Thanks @ripun!

kylog:
@ripun per my comments above, I'm going to close this PR. Please do feel free to raise an issue at tickets.puppetlabs.com or on slack/irc/etc to get more feedback on ideas on how to proceed.

Thanks for digging into this!

PR #5144, opened 2016-07-26
(maint) install cfpropertylist gem on all platforms

branan:
The cfpropertylist gem is needed to run unit tests against code that
parses plist files. We've gotten around this with lots of mocking and
stubbing in the past, but it's coming up to bite us in
https://github.com/puppetlabs/puppet/pull/5119

The core issue is that Windows is not the "ruby" platform - it's
"mingw". This adds a new special "all" platform to project_data.yaml,
where we can add gems which shouldn't be platform-constrained.

Probably what we really need is more complex logic around whether
platform-specific gems are needed only at runtime or also for tests,
and around which platform-specific gems are installable on a broader
array of platforms. This unblocks the above PR, though.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -79,9 +79,15 @@ data['gem_platform_dependencies'].each_pair do |gem_platform, info|
   next if gem_platform == 'x64-mingw32' && !x64_platform
   if bundle_deps = info['gem_runtime_dependencies']
     bundle_platform = bundle_platforms[gem_platform] or raise "Missing bundle_platform"
-    platform(bundle_platform.intern) do
+    if bundle_platform == "all"
       bundle_deps.each_pair do |name, version|
         gem(name, version, :require => false)
+      end
+    else
+      platform(bundle_platform.intern) do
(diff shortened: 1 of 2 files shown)


branan:
I've looked at puppetlabs/packaging a bit, and AFAICT the `bundle_platforms` stuff is only used locally in this repo - the packaging repo accepts the key as valid and ignores it. I assume the packaging repo cares only about the "real" platform names in `gem_platform_dependencies`

mckern:
> I assume the packaging repo cares only about the "real" platform names in gem_platform_dependencies

More or less. This shouldn't have any effect with regards to the packaging repo.

branan:
This appears to be failing due to the `json_pure` Ruby 1.9 incompatibility. Otherwise appveyor confirms that we're now installing cfpropertylist on Windows

adrienthebo:
:+1:, happy to merge once the json_pure issue is resolved.

kylog:
I kicked travis, to see if it goes green with the json_pure version specified.

PR #5145, opened 2016-07-26
(PUP-6549) fix test systemd_enabling_masked_service for Fedora 23

shermdog:
Upstream httpd package requires an updated libnghttp2 package but is not defined as such.
https://bodhi.fedoraproject.org/updates/httpd-2.4.23-4.fc23
https://bugzilla.redhat.com/show_bug.cgi?id=1358875


geoffnichols:
👍

kylog:
On puppet, we merge up stable -> master, so you don't need to (in fact, shouldn't) PR both branches.

PR #5147, opened 2016-07-26
(PUP-6527) Ignore hiera_include calls from environment compiler

nicklewis:
A common pattern for hiera_include is to put it at topscope, which
causes environment compilation to fail as the behavior of hiera_include
(or hiera in general) isn't well-defined for the environment compiler.

In order to allow users to immediately start using application
management and direct change without needing to modify how they do
classification, we now simply ignore and warn if hiera_include is called
by the environment compiler.

### DIFF
--- a/spec/unit/parser/environment_compiler_spec.rb
+++ b/spec/unit/parser/environment_compiler_spec.rb
@@ -480,7 +480,6 @@ EOS
     end
   end
 
-
   describe "in the environment catalog" do
     it "does not fail if there is no site expression" do
       expect {
@@ -488,6 +487,18 @@ EOS
         notify { 'ignore me':}
       EOC
       }.to_not raise_error()
+    end
+
+    it "ignores usage of hiera_include() at topscope for classification" do
+      Puppet.expects(:debug).with(regexp_matches /Ignoring hiera_include/)
+
+      expect {
+        catalog = compile_to_env_catalog(<<-EOC).to_resource
+        hiera_include('classes')
+        site { }
+        EOC
+      }.to_not raise_error()
+
     end
 
     it "includes components and capability resources" do
--- a/lib/puppet/parser/environment_compiler.rb
+++ b/lib/puppet/parser/environment_compiler.rb
@@ -15,6 +15,16 @@ class Puppet::Parser::EnvironmentCompiler < Puppet::Parser::Compiler
     end
   end
 
+  def initialize(node, options = {})
+    super
+    add_function_overrides
+  end
+
+  def add_function_overrides
+    hiera_include = proc { Puppet.debug "Ignoring hiera_include() during environment catalog compilation" }
+    loaders.puppet_system_loader.add_entry(:function, 'hiera_include', hiera_include, nil)
+  end
+
   def add_catalog_validators
     super
     add_catalog_validator(CatalogValidator::SiteValidator)


nicklewis:
@hlindberg I opened this PR for further discussion of the specific solution.

hlindberg:
I am not keen at all on this solution. I would prefer if the environment compiler added a "hiera_include" that does nothing to the loaders. That dummy, since added before anything is evaluated will then be used instead of the hiera_include that would be otherwise loaded from disk.
Then all of the functionality is determined by the environment compiler, the env compiler does not have to be loaded etc.

Also, stopping hiera_include from doing something does not prevent any other similar logic from having the same bad effect. I know people have changed from using hiera_include (as it stopped working for some during 3.x) and instead use hiera_array('classes').each |$c| { include $c } or similar constructs with future parser.

nicklewis:
@hlindberg I took a stab at overriding hiera_include with a dummy version in the environment compiler, but I'm not sure I did it correctly. Please let me know what you think.

hlindberg:
Will look at how to do that tomorrow.  Basically you have to create the function the same way as 4.x functions are created and then add that to the loader. Adding nil would mean that `hiera_include` was not found. Not sure what it does if adding a proc as I think it expects to get a Function back. It may work though since we do equate Function instances with Proc instances interface wise (for lambdas).

hlindberg:
If test goes green I guess the override with a proc worked :-)

spec/unit/functions/hiera_spec.rb
```diff
@@ -121,5 +122,12 @@
       hiera_include.expects(:call_function_with_scope).with(scope,'include', %w[key foo])
       hiera_include.call(scope, 'foo') { |k| ['key', k] }
     end
+
+    it 'should do nothing when called during environment compilation' do
```
hlindberg: This is not a unit test of the hiera_include function - should be moved to the spec for environment_compiler

spec/unit/functions/hiera_spec.rb
```diff
@@ -121,5 +122,12 @@
...
+      scope = create_test_scope_for_environment('production')
+      hiera_include = scope.compiler.loaders.puppet_system_loader.load(:function, 'hiera_include')
```
hlindberg: This does not test the full code path from a call in a manifest. Suggest using `scope.call_function('hiera_include', ['foo']) instead.
Also, since this assumes that 'foo' is a key in hiera data that would return something other than nil, maybe check that the expected debug message is printed to ensure it hits the overridden function (the proc) instead of just returning nil for "not found"

nicklewis:
@hlindberg I updated the tests to compile a manifest, which tests exactly the case I care about.

spec/unit/parser/environment_compiler_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'puppet_spec/compiler'
 require 'puppet/parser/environment_compiler'
 
-describe "Application instantiation" do
+describe Puppet::Parser::EnvironmentCompiler do
```
hlindberg: Minor nit...

With this change here, the describe labels makes less sense - for example it used to say "Application instantiation in the environment catalog ..."
Also, the use of the constant Puppet::Parser::EnvironmentCompiler instantiates it - is it used anywhere?

hlindberg:
Looks good now. Only had a minor nit about the resulting readability of spec tests lables

nicklewis:
@hlindberg Updated to change the test name back.

PR #5149, opened 2016-07-26
(PUP-6439) Retain sensitive context from resources to RAL properties

adrienthebo:
Depends on https://github.com/puppetlabs/puppet/pull/5141.

### DIFF
--- a/lib/puppet/transaction/resource_harness.rb
+++ b/lib/puppet/transaction/resource_harness.rb
@@ -103,7 +103,7 @@ class Puppet::Transaction::ResourceHarness
       if param.should && !param.safe_insync?(current_value)
         event = create_change_event(param, current_value, historical_value)
         if do_audit
-          event = audit_event(event, param)
+          event = audit_event(event, param, context)
         end
 
         brief_audit_message = audit_message(param, do_audit, historical_value, current_value)
@@ -174,10 +174,20 @@ class Puppet::Transaction::ResourceHarness
   end
   private :are_audited_values_equal
 
-  def audit_event(event, property)
+  # Populate an existing event with audit information.
+  #
+  # @param event [Puppet::Transaction::Event] The event to be populated.
+  # @param property [Puppet::Property] The property being audited.
+  # @param context [ResourceApplicationContext]
+  #
+  # @return [Puppet::Transaction::Event] The given event, populated with the audit information.
+  def audit_event(event, property, context)
     event.audited = true
     event.status = "audit"
-    if !are_audited_values_equal(event.historical_value, event.previous_value)
+
+    # The event we've been provided might have been redacted so we need to use the state stored within
+    # the resource application context to see if an event was actually generated.
+    if !are_audited_values_equal(context.historical_values[property.name], context.current_values[property.name])
       event.message = property.format("audit change: previously recorded value %s has been changed to %s",
                  property.is_to_s(event.historical_value),
                  property.is_to_s(event.previous_value))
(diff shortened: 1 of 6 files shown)


lib/puppet/type/file.rb
```diff
@@ -887,6 +887,31 @@ def write(property = nil)
 
   private
 
+  def set_sensitive_properties(sensitive_parameters)
```
adrienthebo: This is some of the most evil code I've written this year, and I don't think there's a better way to deal with this.

lib/puppet/type/file.rb
```diff
@@ -887,6 +887,31 @@ def write(property = nil)
...
+      sensitive_parameters << :content
+      # The `source` parameter will generate the `content` property when the resource state is retrieved
+      # but that's long after we've set the sensitive context. Force the early creation of the `content`
+      # attribute so we can mark it as sensitive.
+      newattr(:content)
+    end
+
+    # If we have content that's marked as sensitive but the file doesn't exist then the ensure
+    # property will be responsible for syncing content, so we have to mark ensure as sensitive as well.
+    if sensitive_parameters.include?(:content)
```
adrienthebo: This is only necessary if the file doesn't exist and might unnecessarily redact information for files that exist. That being said I think this'll only come up in bizarre edge cases so we might be able to wing it.

lib/puppet/transaction/resource_harness.rb
```diff
@@ -174,10 +174,20 @@ def are_audited_values_equal(a, b)
...
+  #
+  # @return [Puppet::Transaction::Event] The given event, populated with the audit information.
+  def audit_event(event, property, context)
     event.audited = true
     event.status = "audit"
-    if !are_audited_values_equal(event.historical_value, event.previous_value)
+
+    # The event we've been provided might have been redacted so we need to use the state stored within
+    # the resource application context to see if an event was actually generated.
+    if !are_audited_values_equal(context.historical_values[property.name], context.current_values[property.name])
```
branan: As far as I could tell when we looked at this, this check is redundant in all the codepaths that we actually call `audit_event` from - does it make sense to just remove it instead of passing the context in?

lib/puppet/type/file.rb
```diff
@@ -887,6 +887,37 @@ def write(property = nil)
...
+      sensitive_parameters.delete(:source)
+      # The `source` parameter will generate the `content` property when the resource state is retrieved
+      # but that's long after we've set the sensitive context. Force the early creation of the `content`
+      # attribute so we can mark it as sensitive.
+      newattr(:content).sensitive = true
+    end
+
+    # If we have content that's marked as sensitive but the file doesn't exist then the ensure
+    # property will be responsible for syncing content, so we have to mark ensure as sensitive as well.
+    if sensitive_parameters.include?(:content)
```
branan: This doesn't look like it properly carries transitively from sensitive source all the way to sensitive ensure

lib/puppet/transaction/resource_harness.rb
adrienthebo: While this is true, I would rather not tackle this now; if I'm back in the resource harness I'll clean this up.

lib/puppet/type/file.rb
adrienthebo: Correct, I broke this in a refactor. Fixed.

lib/puppet/transaction/resource_harness.rb
branan: :+1:

PR #5150, opened 2016-07-27
(PUP-6557) fix Dir.glob shenanigans

Iristyle:
Ruby 2.3 changed the behavior of `Dir.glob` as a result of https://bugs.ruby-lang.org/issues/10819 / https://github.com/ruby/ruby/commit/45df1c24d269f93a2bc1e7a6fe0ffcecc1193051

This requires test / code changes on Windows.

### DIFF
--- a/lib/puppet/pops/loader/module_loaders.rb
+++ b/lib/puppet/pops/loader/module_loaders.rb
@@ -215,7 +215,9 @@ module ModuleLoaders
 
     def existing_path(effective_path)
       # Optimized, checks index instead of visiting file system
-      @path_index.include?(effective_path) ? effective_path : nil
+      # TODO: has this killed the optimization? The problem is we may be comparing 8.3 paths to expanded paths found by Dir.glob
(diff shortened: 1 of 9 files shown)


lib/puppet/pops/loader/module_loaders.rb
```diff
@@ -215,7 +215,9 @@ def initialize(parent_loader, loaders, module_name, path, loader_name, loadables
 
     def existing_path(effective_path)
       # Optimized, checks index instead of visiting file system
-      @path_index.include?(effective_path) ? effective_path : nil
+      # TODO: has this killed the optimization? The problem is we may be comparing 8.3 paths to expanded paths found by Dir.glob
+      # and there is no canonical form
+      @path_index.find { |f| File.identical?(f, effective_path) }
```
hlindberg: this does indeed kill the optimization

Iristyle:
Alternate (and probably better / simpler) proposal in #5156 ... think I'll close this one off

PR #5151, opened 2016-07-27
{WIP}(PUP-6551) Pin json_pure to 2.0.1 if on Ruby 1.x

Iristyle:
- Hiera has an unbound dependency on json_pure. json_pure 2.0.2+
   requires Ruby >= 2.0.

   Pin json_pure to maximum version of 2.0.1 on Ruby 1.x.

   Without this commit, bundler will pick json_pure 2.0.2 and then
   fail to bundle install on Ruby 1.x.

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -19,6 +19,7 @@ gem_required_rubygems_version: '> 1.3.1'
 gem_runtime_dependencies:
   facter: ['> 2.0', '< 4']
   hiera: ['>= 2.0', '< 4']
+  # TODO: but what about this?
   json_pure:
 gem_rdoc_options:
   - --title
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,10 @@ gem "puppet", :path => File.dirname(__FILE__), :require => false
 gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 2.0', '< 4'])
 gem "hiera", *location_for(ENV['HIERA_LOCATION'] || ['>= 2.0', '< 4'])
 gem "rake", "10.1.1", :require => false
+# Hiera has an unbound dependency on json_pure
+# json_pure 2.0.2+ requires Ruby >= 2.0
+# Ensure json_pure maximum version 2.0.1 on Ruby 1.x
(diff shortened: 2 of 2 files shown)


Iristyle:
@glennsarti I took goofy commit https://github.com/puppetlabs/puppet/commit/898b13eec2b85a343553b2c7e0e1749addcb9e67 from #5140 and moved it here as a separate PR.

The only problem here is that this doesn't address the shipped gem (see the `TODO` in `project_data.yaml`) - and I'm not sure if we want to address that.

@kylog thoughts?

ext/project_data.yaml
```diff
@@ -19,6 +19,7 @@ gem_required_rubygems_version: '> 1.3.1'
 gem_runtime_dependencies:
   facter: ['> 2.0', '< 4']
   hiera: ['>= 2.0', '< 4']
+  # TODO: but what about this?
```
Iristyle: Do we want to set a range on ship? (is there anything useful in 2.0+ of json_pure that we even care about?)

Iristyle:
FYI - here's the json_pure CHANGELOG - https://github.com/flori/json/blob/master/CHANGES.md

camlow325:
@Iristyle, @kylog - Should we put this PR up against the 3.x branch and merge it up to stable and on to master from there?  Seems like the same problem would happen for CI off of the 3.x branch.  It is where Puppet Server's CI references the Puppet 3.x branch for the Ruby spec tests it runs.

kylog:
I'd rather not pin to different gem versions based on different ruby versions unless our hand is absolutely forced. That feels like it's inviting future discoveries and we're better just constraining ourselves to a gem version that works for all supported ruby versions.

And yes, let's target the json_pure solution to 3.x.

And unless we need something in 2.0.0/2.0.1 I'd lean toward just making it a classic pessimistic style pin to `~> 1.latest_Y`.

Thoughts on all that?

glennsarti:
:+1:  on Kylo's comments.  Makes sense.

Iristyle:
`< 2` / `~> 1` should be functionally equivalent, unless you want to go with `~> 1.8` which would be a bit more constrained to the `1.8` range.

Iristyle:
Closing in favor of #5152 against 3.x and merging up from there.

PR #5152, opened 2016-07-27
(PUP-6551) Pin json_pure to ~> 1.8

Iristyle:
- Hiera has an unbound dependency on json_pure. json_pure 2.0.2+
   explicitly requires Ruby >= 2.0, though support was dropped for
   Ruby 1.9 in json_pure 2.0.0 (though mistakently the Gemfile was
   not updated).

   Pin json_pure to a 1.8 series release everywhere.

   Without this commit, bundler will pick json_pure 2.0.2 and then
   fail to bundle install on Ruby 1.x.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,9 @@ end
 gem "facter", *location_for(ENV['FACTER_LOCATION'] || ['> 1.6', '< 3'])
 gem "hiera", *location_for(ENV['HIERA_LOCATION'] || '~> 1.0')
 gem "rake", "10.1.1", :require => false
+# Hiera has an unbound dependency on json_pure
+# json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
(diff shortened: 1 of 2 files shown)


kylog:
Thanks @Iristyle; this is what I was thinking.

But, do you know what's up with the appveyor fail?

ferventcoder:
Looks unrelated based on not creating a user?

ferventcoder:
I rekicked the appveyor build - 

Lots of these messages:
~~~sh
An error occurred in an after hook
  Puppet::Util::Windows::Error: Failed to convert string SID: S-1-5-:  The security ID str
  occurred at C:/projects/puppet/lib/puppet/util/windows/sid.rb:106:in `block (2 levels) in string_to_sid_ptr'
F
~~~

Iristyle:
This requires #5153 to be merged first, to gem pin win32-* gems that are causing us problems (covered in PUP-6445).

PR #5153, opened 2016-07-27
(PUP-6445) Pin win32 gems to explicit versions

Iristyle:
The Win32-Service gem was updated to 0.8.9 however the new gem pulled in
a dependent gem called ffi-win32-extensions which has a name collision
on the FFI:Pointer.read_wide_string method.  This causes Puppet to fail
in mutiple locations.  This commit pins the Win32-Service gem to a
prior version without the dependency.

Additionally, all other win32-xxx gems are pinned to explicit versions
so that Puppet may not experience a similar issue in the future.

This was performed in d4d6734e5f858b4e3f4db24c356ea7cb9db2d991 against
stable, but should have also targeted 3.x as well.

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -29,21 +29,25 @@ gem_platform_dependencies:
     gem_runtime_dependencies:
       # Pinning versions that require native extensions
       ffi: '~> 1.9.5'
-      win32-dir: '~> 0.4.9'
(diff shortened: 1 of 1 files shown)


ferventcoder:
Good idea

glennsarti:
Confirmed in local dev environment, expected gems are installed.

PR #5154, opened 2016-07-27
3.x merge to stable

kylog:
This is a merge up from 3.x. Note that several of the 3.x changes amounted to noops on stable, specifically:
* the pip provider is no longer XMLRPC based, but shells out to the pip cli, see https://github.com/puppetlabs/puppet/pull/4832
* the windows gem pinning already happened in stable, see https://github.com/puppetlabs/puppet/pull/5062


branan:
LGTM. Just waiting on Travis to do its thing

PR #5156, opened 2016-07-28
(PUP-6557) Adjust module loading for Dir.glob

Iristyle:
- In Ruby 2.3, Dir.glob changed behavior on Windows to address a bug
   filed in https://bugs.ruby-lang.org/issues/10819

   The changes landed in Ruby commit
   ruby/ruby@45df1c2

 - Unfortunately, this introduced a behavioral change in how 8.3
   paths were evaluated in the Dir.glob call, expanding the paths to
   their full path equivalent.

   Prior to Ruby 2.3:

   Dir.glob('C:/Users/Admini~1')
   => ['C:/Users/Admini~1/']

   Ruby 2.3:

   Dir.glob('C:/Users/Admini~1')
   => ['C:/Users/Administrator/']

 - This throws off any path comparison logic in module loading code,
   because Puppet::Pops::Loader::ModuleLoaders implementations will
   cache off filenames, later comparing the initially given path to
   the results from Dir.glob - which may now vary on Windows. Short
   8.3 paths no longer match Dir.glob returns, confusing the module
   loaders and their notion of what has already been loaded.

   In particular, this can result in custom functions failing to be
   loaded from modules, but may have other potential failure modes.

 - The basic solution, since module loaders are only created by the
   environments they live in, is to ensure that when an Environment is
   created, that it always uses expanded paths to begin with.  This
   prevents any mismatch problem due to Dir.glob expanding paths, by
   passing Dir.glob already expanded paths.

   After analysis, it was determined that outside of tests, an
   Environment is either created by Environments, or is created more
   directly via the factory method Environment.create.  Therefore only
   these call sites were updated to minimize any performance impact of
   the module loading code.

   Paths are compared by strings to avoid touching disk / stat'ing
   files, which can be quite costly.  The benchmark:function_loading
   micro-benchmark was run to ensure no performance degradation was
   introduced as a result

 - Introduce a new Puppet::FileSystem.expand_path method that behaves
   just like the Ruby File.expand_path method (and in fact it's just a
   wrapper for non-Windows).  For Windows, expand_path will call the
   Windows API GetLongPathName when the path contains a ~. Otherwise
   it will adhere to the same behavior as File.expand_path, even when
   given the 2nd optional parameter `dir_string`, used as a base for
   a relative path.

 - In some cases, tests were failing due to a tmpdir / tmpfile
   generated path (which is 8.3) being compared to the path loaded
   inside a module loader.  Rather than fix all of these individual
   calls, it made more sense to adjust the behavior of the tmpdir /
   tmpfile helpers to always pass out fully expanded paths, rather
   than 8.3 style.  This minimizes code churn.

Paired-with: Daniel Lu <daniel.lu@puppet.com>
Paired-with: Moses Mendoza <moses@puppet.com>


Iristyle:
@puppetlabs/windows + @McdonaldSeanp - Still needs a test to ensure we test this behavior under AppVeyor (and will break should something change).... but this isn't totally unreasonable (if a little clunky)

@hlindberg / @thallgren - another option for solving this in the short term.

lib/puppet/pops/loader/module_loaders.rb
```diff
@@ -215,7 +215,8 @@ def initialize(parent_loader, loaders, module_name, path, loader_name, loadables
 
     def existing_path(effective_path)
       # Optimized, checks index instead of visiting file system
-      @path_index.include?(effective_path) ? effective_path : nil
+      # TODO: should we check glob'd, then literal version?
```
Iristyle: one open question on this one

lib/puppet/module.rb
```diff
@@ -180,7 +180,7 @@ def match_manifests(rest)
 
     # (#4220) Always ensure init.pp in case class is defined there.
     init_manifest = manifest("init.pp")
-    if !init_manifest.nil? && !searched_manifests.include?(init_manifest)
+    if !init_manifest.nil? && !searched_manifests.include?(Dir.glob(init_manifest).first)
```
hlindberg: So this would search for the file on disk again to get the path in the same form.

lib/puppet/pops/loader/module_loaders.rb
```diff
@@ -215,7 +215,8 @@ def initialize(parent_loader, loaders, module_name, path, loader_name, loadables
...
+      @path_index.include?(Dir.glob(effective_path).first) ? effective_path : nil
```
hlindberg: This searches again for every search.

hlindberg:
The change now does a Dir.glob for every path search in every loader. The result is a very large number of expensive stats in an environment with a typical amount of modules.
Can the Dir.glob result be sanitized on Windows only? As I understand it, we don't need the  8.3 per se, only that the output matches the input. Such a solution would have different performance impact - slightly slower, but only once per module per environment per loaded kind as opposed to once per "not yet loaded thing" x "number of modules". Stats are expensive, and they are especially expensive when running in a VM and/or with networked disks.

Alternatively, can the path that we use to search for be transformed to the format Dir.glob returns on Windows without touching the disk?

I suspect that having to do a Dir.glob for every possible searched path is worse than the non optimized solution (using a stat per path).

The use case here does not require a canonical "same file", it is enough to answer "same path", and somehow that seems like it should be possible to do without touching the disk.

lib/puppet/module.rb
Iristyle: One thought here is that we could pay the tax only on Windows, by using a method on `Puppet::FileSystem` here to expand the path out.

lib/puppet/pops/loader/module_loaders.rb
Iristyle: Worth measuring the actual impact.

Iristyle:
@hlindberg if we can get a pathological case here, I can perform some worst case measurements.  Running the full test suite on Windows, showed no degradation - but I don't think the suite stresses the right things.

Note I've isolated the expansion code to Windows only now, so there won't be any hit on non-Windows.

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,14 @@
 
 class Puppet::FileSystem::Windows < Puppet::FileSystem::Posix
 
+  def path_string(path)
+    # attempt to expand
+    # Dir.glob(path.to_s).first || path.to_s
```
Iristyle: We have 2 options here... use `Dir.glob` on Windows, which does a bunch of work... or just expand the file the simpler way with the Windows API.

hlindberg:
I think we are miscommunicating. Since we know what the path we would like to see I would be surprised if it was not possible to map it to the form that is returned by Windows Dir.glob - at least if the LongForm for the module root and environment roots are known (I can imagine that part to differ and include things like drive letters and the like).

Regarding benchmarks. Have to do some spelunking to see if there some that are applicable. At the moment we are not loading more than a couple of things but loaders are getting used for more and more things - with environment isolation also for resource types - and later also for classes and user defined types. The design of the new loaders were made (with plenty of micro benchmarking, to reduce the amount of disk stats and reads).

I would not imagine that the test suite would show regressions since it starts from scratch over and over again and have a very small search space.

Iristyle:
@hlindberg if you could have another look....

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,16 @@
 
 class Puppet::FileSystem::Windows < Puppet::FileSystem::Posix
 
+  def path_string(path)
```
Iristyle: I'm not sure that using `path_string` here is the best... I think it might be better to add a new method to file_system like path_string... or perhaps just use `expand_path` where necessary.  Going to try an `expand_path` variant.

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,16 @@
...
+    return path.to_s if !glob_expands_paths
```
hlindberg: Why the `to_s` here ? Isn't the input always a String ?

lib/puppet/file_system/windows.rb
```diff
@@ -100,4 +110,18 @@ def raise_if_symlinks_unsupported
     end
   end
 
+  def glob_expands_paths
```
hlindberg: Should this be `glob_expands_path?` since it seems to be a boolean query method ?

hlindberg: Is this method private to the Windows file system? if not, it should have some documentation.

spec/unit/pops/loaders/loader_paths_spec.rb
```diff
@@ -38,6 +38,7 @@ def foo; end;
     expect(effective_paths.size).to eq(1)
     expect(module_loader.path_index.size).to eq(1)
     path_index = module_loader.path_index
-    expect(path_index).to include(File.join(module_dir, 'lib', 'puppet', 'functions', 'foo4x.rb'))
+    expected = File.join(module_dir, 'lib', 'puppet', 'functions', 'foo4x.rb')
+    expect(path_index).to include(Dir.glob(expected).first)
```
hlindberg: The use of Dir.glob exposes an implementation detail. Shouldn't this go via `FileSystem::path_string(expected)` instead? (Or did I misunderstand).

lib/puppet/file_system/windows.rb
Iristyle: Yes, will add the sigil.  `glob_expands_paths` is already private, yes

spec/unit/pops/loaders/loader_paths_spec.rb
Iristyle: Yeah, I'm still debating on whether this makes complete sense as implemented.

The way I've implemented things now, the behavior varies based on platform / Ruby version - which I'm honestly not an enormous fan of.  I'd rather do something to make the behavior consistent across at least the platform.

Iristyle: I did another spike last night on calling `File.expand_path` in a few key locations, so that paths were always of the desired format (which I think would make things simpler).  I used `dtrace` to make sure `expand_path` doesn't make any extra `fstat64` calls on OSX.

However, `expand_path` on Windows didn't have the desired behavior, as it only expanded the last segment of the path with a `~`:

```
[7] pry(#<Puppet::Module>)> File.expand_path('C:/Users/ADMINI~1/')
=> "C:/Users/Administrator"
[8] pry(#<Puppet::Module>)> File.expand_path('C:/Users/ADMINI~1/AppData')
=> "C:/Users/ADMINI~1/AppData"
```

One thought was building up a new `FileSystem` method called `expand_path_recursive` or similar that made sure we got the desired behavior.

Iristyle:
For reference, the old code is at https://github.com/puppetlabs/puppet/pull/5156/commits/7a0e03d18707ff73d4836f2f5bdbd9d7a235faa7 until Github does a compact / removal (and I have it tucked away in a backup branch).

I took a slightly different approach, and tried to whittle down the overall impact here, and have repushed as 43223d5

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,20 @@
...
+  def expanded_path_string(path)
+    string_path = path.to_s
+    # if no tildes, nothing to expand, return original string
+    # WARN: this will return a / on Windows if passed a /
+    return string_path if string_path !~ /~/
+
+    expanded = ''
+    Pathname.new(string_path).descend do |f|
+      expanded = File.expand_path(File.join(expanded, File.basename(f)))
+    end
```
Iristyle: If we can reduce the number of method calls here, that would be good.

spec/unit/pops/loaders/loader_paths_spec.rb
Iristyle: This approach has replaced the one here as of https://github.com/puppetlabs/puppet/commit/3a9af880bd4da01a246b7acf10b1942569d78da3

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,20 @@
...
 
+  def expanded_path_string(path)
+    string_path = path.to_s
+    # if no tildes, nothing to expand, return original string
+    # WARN: this will return a / on Windows if passed a /
+    return string_path if string_path !~ /~/
+
+    expanded = ''
+    Pathname.new(string_path).descend do |f|
+      expanded = File.expand_path(File.join(expanded, File.basename(f)))
```
MosesMendoza: Will the parameter `path` ever be a relative path, e.g., `~/../../` ?

lib/puppet/file_system.rb
```diff
@@ -348,6 +349,21 @@ def self.path_string(path)
...
+  # be expanded to C:\Users\Administrator\AppData.  On POSIX filesystems, the
+  # value ~ will be expanded to something like /Users/Foo
+  #
+  # This method exists primarlily to resolve a Ruby deficiency where
+  # File.expand_path doesn't handle ~ in each segment on a Windows path
+  #
+  # @param path [Object] a path handle produced by {#pathname}
+  # @return [String] a string representation of the path
+  #
+  def self.expanded_path_string(path)
```
hlindberg: is `expand_path_string` a better name? (a verb rather than a noun which suggest something like `to_expanded_path_string`.

lib/puppet/module.rb
```diff
@@ -128,6 +128,8 @@ def has_metadata?
       end
 
       return nil unless Puppet::FileSystem.exist?(full_path)
+      # TODO: line left in while testing to see if this is necessary
+      Puppet.warning("Unexpanded path #{full_path}") if full_path.upcase != Puppet::FileSystem.expanded_path_string(full_path).upcase
```
hlindberg: This code path is not used (IIRC) from the 4.x loaders, and as comments said this is used primarily for file serving, so I don't have anything to say here. Hope that piece of code is not needed, and if needed users should have done the "expand" piece earlier if that is required.

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,20 @@
 
 class Puppet::FileSystem::Windows < Puppet::FileSystem::Posix
 
+  def expanded_path_string(path)
+    string_path = path.to_s
+    # if no tildes, nothing to expand, return original string
+    # WARN: this will return a / on Windows if passed a /
+    return string_path if string_path !~ /~/
```
hlindberg: this is slow in comparison to using `String#index` since it is a one character search. Using a regexp is overkill that creates result object(s), sets global variables and more...

lib/puppet/file_system.rb
hlindberg: `expanded_path` is quite confusing as there is a `File#expand_path` that is not the same thing.

lib/puppet/file_system/windows.rb
Iristyle: I don't believe it will (there are other similar assumptions that the code already makes around these sorts of questions)

lib/puppet/file_system.rb
Iristyle: Agreed on the naming - it was a placeholder until we came up with something that made more sense after looking at the implementation.

On the other hand, it is basically the same thing as `expand_path`, but with fixes.... so, maybe `expand_path` is the right name?

lib/puppet/pops/loader/module_loaders.rb
```diff
@@ -214,8 +214,12 @@ def initialize(parent_loader, loaders, module_name, path, loader_name, loadables
     end
 
     def existing_path(effective_path)
+      expanded_path = Puppet::FileSystem.expanded_path_string(effective_path)
```
hlindberg: Here I am thinking, that it would be better if `effective_path` was already an "implementation specific path" since that is the purpose of this entire mechanism (comparing a tentative path against an existing path obtained from Dir.glob). Now the logic would first compose a non file system specific path out of "a root for a module or env" + a number of directory "segments" (for example 'functions', or 'type'). The construction of those segments into the "effective path" could perhaps do the "impl specific path" right away using fewer intermediate objects and scans over the characters in the same (or very similar) strings.

lib/puppet/module.rb
Iristyle: That note was more for me while running through tests.  In prior iterations of this PR I returned `expanded_path_string`.  But when I moved to just setting the path to expanded in the initializer, I presumed I could remove all the other calls.  This is just a sanity check in that vein.

lib/puppet/pops/loader/module_loaders.rb
Iristyle: The only caller of `expanded_path` now in the code is at https://github.com/Iristyle/puppet/blob/206dfb86b5179f83489c79c1014be2859e24d9e0/lib/puppet/pops/loader/module_loaders.rb#L133

I took a look at the code paths, and saw that I could influence the cached values much earlier in the lifecycle by modifying the `AbstractPathBasedModuleLoader` so that the initializer calls `expand` - means that the various SmartPath instances would have the right style path fed in in the first place, making later checks unnecessary.

Then these modifications can be removed, which I've done in my latest PR update (in addition to fixing broken tests)

Iristyle:
So I was reasonably happy with the code... until I re-ran the performance benchmarks.  The prior iteration was pretty on-par with existing perf.

This variant has slowed `benchmark:function_loading` by a factor of 2x-3x, so will have to rethink.

lib/puppet/file_system/windows.rb
```diff
@@ -9,7 +9,8 @@ def expanded_path_string(path)
     # WARN: this will return a / on Windows if passed a /
     return string_path if !string_path.index('~')
 
-    expand_path_cached(path)
+    Puppet::Util::Windows::File.get_long_pathname(string_path)
+    # expand_path_cached(path)
```
hlindberg: remove commented out code

lib/puppet/file_system.rb
hlindberg: It is not really the same as expand_path. We don't want to expand '..' and the like here as that is extra cost (must touch filesystem).
Best if name is distinct.

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,16 @@
 
 class Puppet::FileSystem::Windows < Puppet::FileSystem::Posix
 
+  def expanded_path_string(path)
+    string_path = path.to_s
+    # if no tildes, nothing to expand, return original string
+    # WARN: this will return a / on Windows if passed a /
+    return string_path if !string_path.index('~')
```
hlindberg: `unless` instead of `if !`

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,16 @@
...
+
+    Puppet::Util::Windows::File.get_long_pathname(string_path)
+    # expand_path_cached(path)
```
hlindberg: commented out code (I commented about it on the commit too I think...)

lib/puppet/file_system/windows.rb
```diff
@@ -89,6 +99,25 @@ def read_preserve_line_endings(path)
 
   private
 
+  def expand_path_cached(path)
```
hlindberg: afaict, this method is not used (call is commented out).

lib/puppet/file_system/windows.rb
```diff
@@ -89,6 +99,25 @@ def read_preserve_line_endings(path)
...
+    @memoized_expanded_paths ||= {}
```
hlindberg: what is the lifecycle of this cache? Risk of leaking between environments? What happens when files are removed or changed form (file to dir and vice versa)?

lib/puppet/file_system.rb
hlindberg: Also on n*x we don't want to expand ~ either. Thus for n*x there is nothing to do but to return the original. (Hence: this is not the same as expand_path)

lib/puppet/file_system/file_impl.rb
```diff
@@ -22,6 +22,10 @@ def path_string(path)
     path.to_s
   end
 
+  def expanded_path_string(path)
+    ::File.expand_path(path_string(path))
```
hlindberg: We don't what this expansion for 'nix.

lib/puppet/file_system/windows.rb
```diff
@@ -89,6 +99,25 @@ def read_preserve_line_endings(path)
...
+    split_index = path.index('/', path.index('~'))
+    to_expand = path[0..(split_index -1)]
+    if !@memoized_expanded_paths.key?(to_expand)
+      expanded = ''
+
+      # Ruby limitation: must call expand_path on each subdirectory
```
hlindberg: wat?

Iristyle: This is mostly the reason this PR exists.  Read the docs for the original `expanded_path_string`

lib/puppet/file_system/windows.rb
Iristyle: I thought we had a ban on `unless` in the Puppet codebase ;0

Iristyle:
@hlindberg thanks for the comments - not sure if you caught that the last few iterations have been failed attempts to improve perf.... I think I finally hit paydirt though if we make a few adjustments to `environment` and leave `module` and `module_loader` alone.  Then the correct path trickles down everywhere.  I benchmarked that implementation and perf is back on par with what it was previously.  <sigh of relief>

I wanted to make sure it passed specs though, so I ran it locally before I dozed off - and the outlook is pretty good.

There is just one spot I need to figure out... I'll point it out after the next push.

lib/puppet/file_system/windows.rb
Iristyle: This only tracks dirs with '~' in them, and maintains a string to string map like:

* `c:\users\admini~1` -> `c:\users\Administrator`

I don't believe any of your concerns should apply.

I will probably end up removing this bit though because I think it adds only a negligible perf gain, but additional code / complexity.

lib/puppet/file_system/windows.rb
Iristyle: Yeah, based on WIP perf stuff.

lib/puppet/file_system.rb
Iristyle: I've thought about this a bit more and I disagree, because I think we ultimately want a corrected `expand_path` for the sake of Windows (just as we have fixed `stat` and `lstat` methods for Windows in our `FileSystem` abstraction).  I think the best API here is to follow the semantics of `expand_path` rather than some weird / confusing in-between thing with only one purpose.

Then we can just choose at the call site if we want to expand or not - which pushes complexity out a little bit for the sake of keeping a more reasonably `FileSystem` API.  So it may be that we just don't call it on 'nix for Environment / Module loading... but my latest example might not carry as much of a penalty given the path is now adjusted at the highest Environment level.

I'll have to run a few more perf tests to prove that out, but things are looking good on Windows again.

lib/puppet/pops/loaders.rb
```diff
@@ -195,7 +196,10 @@ def create_environment_loader(environment)
       loader = Loader::SimpleEnvironmentLoader.new(@runtime3_type_loader, loader_name)
     else
       # View the environment as a module to allow loading from it - this module is always called 'environment'
-      loader = Loader::ModuleLoaders.module_loader_from(@runtime3_type_loader, self, 'environment', env_conf.path_to_env)
+      # TODO: temporary hack
+      path_to_env = env_conf.path_to_env
```
Iristyle: @hlindberg this is the part that I didn't quite get to figuring out ... the Environment has fully expanded paths at this point, *but* the call to `Puppet.lookup(:environments).get_conf(environment.name).path_to_env` returns an 8.3.

I want to be able to remove the expansion call here.

lib/puppet/node/environment.rb
```diff
@@ -82,8 +82,12 @@ def self.remote(name)
...
   def initialize(name, modulepath, manifest, config_version)
     @name = name
-    @modulepath = modulepath
-    @manifest = manifest
+    @modulepath = modulepath.map { |p| Puppet::FileSystem.expanded_path_string(p) }
+    if manifest && !manifest.is_a?(Symbol) && Puppet::FileSystem.exist?(manifest)
+      @manifest = Puppet::FileSystem.expanded_path_string(manifest)
+    else
+      @manifest = manifest
+    end
```
Iristyle: @hlindberg this environment initialize code is the now the key to getting the right format path in from the start.

spec/lib/puppet_spec/files.rb
```diff
@@ -28,7 +28,7 @@ def tmpfile(name, dir = nil) PuppetSpec::Files.tmpfile(name, dir) end
   def self.tmpfile(name, dir = nil)
     # Generate a temporary file, just for the name...
     source = dir ? Tempfile.new(name, dir) : Tempfile.new(name)
-    path = source.path.encode(Encoding::UTF_8)
+    path = Puppet::FileSystem.expanded_path_string(source.path.encode(Encoding::UTF_8))
```
Iristyle: Need to make sure this doesn't significantly impact test runtimes.  On Windows in a VM, no perceptible difference - but need to look at AppVeyor, Travis, local OSX, etc.

lib/puppet/node/environment.rb
Iristyle: Maybe this can even be moved up to `environments.rb`

Iristyle: One thing I need to understand better @hlindberg is the usage patterns here... is it necessary to create an environment to get instances of `module` and `module_loaders`?

If so, it probably just makes sense to do this work here and have it fan out as I mentioned.  But if those other instances are used stand-alone, outside of an environment... then that approach won't always work.

hlindberg: An environment is essential to initialize the loaders, their lifecycle is tied to the environment's.

lib/puppet/node/environment.rb
```diff
@@ -82,8 +82,12 @@ def self.remote(name)
   # @param name [Symbol] The environment name
   def initialize(name, modulepath, manifest, config_version)
     @name = name
-    @modulepath = modulepath
-    @manifest = manifest
+    @modulepath = modulepath.map { |p| Puppet::FileSystem.expanded_path_string(p) }
+    if manifest && !manifest.is_a?(Symbol) && Puppet::FileSystem.exist?(manifest)
```
hlindberg: What is up with the is_a?(Symbol) case? Is that for the NoManifest case ? If so I think it is best if it is explicit.

lib/puppet/node/environment.rb
hlindberg: It makes sense to bake the path that is the environment root in the environment. There are probably many use cases that use that as the root for path construction.
What I am wary of though is that differences bubble up. I can imagine environment having a separate method for "file_system_impl_specific_path", and that it is cached in the environment.
That takes the guesswork out of code in general. Did it get the specific one, should it process it again? Hard to know. Thus, we make every use case of file paths responsible for using the right kind of path.
In the loaders the "smart paths", and when creating a tentative path to test with. Since it has explicit knowledge about the paths, there should be no need to expand the parts relative to the root.

hlindberg:
@Iristyle I think that the file system specific path could be expanded in the env_conf and cached there. The env conf and the env has the same lifecycle.
Then, when loaders are initialized, they should be initialized with the "file system specific path" from the env conf. It is the env conf that is authoritative where the environment's root is.

lib/puppet/node/environment.rb
Iristyle: I can debug and let you know for sure the values.. but while debugging calls that entered here, I saw several different symbols come through as the value of `manifest`, in addition to actual path strings.  I wasn't certain what they meant - and also wasn't sure if we got both `Pathname` and `String` instances here - so that's why I didn't call `exist?` on only string values.

Will report back on this one.

Iristyle:
Ok @hlindberg this appears to do the trick - the expansion has been moved all the way up to `environments.rb`... and as long as the `tmpdir` / `tmpfile` helpers are updated to use the same style paths, all tests pass, and there is basically no performance impact here for environment / module loading.

I'm going to experiment a bit more with removing the changes to `FileSystem` to simplify the approach even further, but before I finish the clean-up and some more analysis, any final thoughts on this one? (I'm also examining the call sequence to make sure that the environment / loader classes are not consumed [aside from tests] directly anywhere else in the code - which is now my main concern)

Iristyle:
I ran some numbers for the entire suite before / after (mostly since I was curious about any impact to `tmpdir` / `tmpfile`) and I found no perceptible difference on my local VM:


### stable

#### 2.1.8

Finished in 12 minutes 12 seconds (files took 14.81 seconds to load)
21223 examples, 0 failures, 62 pending

#### 2.3.1

Finished in 12 minutes 44 seconds (files took 15.55 seconds to load)
21138 examples, 12 failures, 62 pending

### PR

#### 2.1.8

Finished in 12 minutes 17 seconds (files took 15.19 seconds to load)
21223 examples, 0 failures, 62 pending

#### 2.3.1

Finished in 12 minutes 46 seconds (files took 15.8 seconds to load)
21138 examples, 0 failures, 62 pending

Iristyle:
Added experimental commit dc7a9e4 which removes any `FileSystem` changes and just relies on `Dir.glob`

lib/puppet/file_system.rb
```diff
@@ -1,4 +1,5 @@
 module Puppet::FileSystem
+  require 'puppet/util'
```
thallgren: why is this extra requirement needed?

lib/puppet/module.rb
```diff
@@ -60,6 +60,10 @@ def self.is_module_namespaced_name?(name)
 
   def initialize(name, path, environment)
     @name = name
+    # TODO: testing to see if necessary to wrap path with expand_path
+    not_expanded = (path.index('~'))
```
thallgren: Remove redundant parenthesis. Ideally, since `not_expanded` is used as a boolean, coerce it. I.e.
```
not_expanded = !!path.index('~')
```

lib/puppet/module.rb
```diff
@@ -128,6 +132,10 @@ def has_metadata?
       end
 
       return nil unless Puppet::FileSystem.exist?(full_path)
+      # TODO: testing to see if necessary to wrap full_path with expand_path
+      not_expanded = (full_path.index('~'))
```
thallgren: Use
```
not_expanded = !!full_path.index('~')
```

lib/puppet/node/environment.rb
```diff
@@ -82,8 +82,21 @@ def self.remote(name)
   # @param name [Symbol] The environment name
   def initialize(name, modulepath, manifest, config_version)
     @name = name
+    # TODO: testing to see if necessary to wrap each modulepath with expand_path
+    modulepath.each do |p|
+      not_expanded = (p.index('~'))
```
thallgren:
```
not_expanded = !!p.index('~')
```

lib/puppet/node/environment.rb
```diff
@@ -82,8 +82,21 @@ def self.remote(name)
...
+      stack = not_expanded ? caller.find_all { |call| call =~ /^C\:\/source\/puppet\/(lib|spec)\// }.join("\n") : ''
+      puts "\nEnvironment.new - modulepath #{p} is not expanded - call stack:\n#{stack}\n" if not_expanded
+    end
     @modulepath = modulepath
+    # TODO: testing to see if necessary to wrap manifest with expand_path
+    not_expanded = (manifest != :no_manifest && manifest.index('~'))
```
thallgren: Mix of boolean and numeric/nil. Better written as:
```
not_expanded = (manifest != :no_manifest && !!manifest.index('~'))
```

lib/puppet/pops/loader/module_loaders.rb
```diff
@@ -74,6 +74,10 @@ def initialize(parent_loader, loaders, module_name, path, loader_name, loadables
       super parent_loader, loader_name
 
       @module_name = module_name
+      # TODO: testing to see if necessary to wrap path with expand_path
+      not_expanded = (path.index('~'))
```
thallgren: Use !!

lib/puppet/pops/loader/module_loaders.rb
```diff
@@ -227,6 +235,10 @@ def to_s()
     end
 
     def add_to_index(smart_path)
+      # TODO:  testing to see if this is necessary to wrap smart_path.generic_path with Puppet::FileSystem.expand_path
+      not_expanded = (smart_path.generic_path.index('~'))
+      stack = not_expanded ? caller.find_all { |call| call =~ /^C\:\/source\/puppet\/(lib|spec)\// }.join("\n") : ''
+      puts("\nAbstractPathBasedModuleLoader#add_to_index - Unexpanded path #{smart_path.generic_path} - call stack:\n#{stack}\n") if not_expanded
```
thallgren: The above three lines seems to occur at least 9 times in this PR and the only difference seems to the text and path and some difference in how `not_expanded` is computed. Perhaps consider turning that into a utility method (differences in how not_expanded is computed could be provided as a default implementation and an optional block).

Iristyle: That code is in there simply for analysis, and will not be in the final PR update - its totally throwaway ;0

lib/puppet/file_system.rb
Iristyle: See 	f7c3b21

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,27 @@
...
+    begin
+      string_path = Puppet::Util::Windows::File.get_long_pathname(string_path)
+      # TODO: ignore file not found errors and pass out original string
+      # to preserve the regular expand_path behavior
+      # no need to do existence check up front as GetLongPathName implies that check is performed
+      # and it should be the exception that files aren't actually present
+    rescue Puppet::Util::Windows::Error => e
+      raise if (e.code != Puppet::Util::Windows::File::ERROR_FILE_NOT_FOUND &&
+        e.code != Puppet::Util::Windows::File::ERROR_PATH_NOT_FOUND)
+      puts "#{e.code} error on with GetLongPathName and #{path}\n"
```
hlindberg: This error message reads a bit strange "xxx error on with..."

hlindberg:
This is looking good now. The expansion (with possible touching of the file system on Windows) is now done as few times as possible and direct comparison of tentative paths to load are just hash lookups.

lib/puppet/file_system/windows.rb
Iristyle: Whoops - that `puts` statement was only used during debugging - I have to remove that and the above `TODO`

spec/unit/file_system_spec.rb
```diff
@@ -649,5 +649,49 @@ def increment_counter_in_multiple_processes(file, num_procs, options)
...
+        expect(Puppet::FileSystem.exist?(nonexist_path)).to be_falsey
+        # this will change c:\\progra~1 to c:/progra~1 (existing Ruby behavior), but not expand any ~
+        ruby_expanded = File.expand_path(nonexist_path)
+        expect(ruby_expanded).to match(/~/)
+        expect(Puppet::FileSystem.expand_path(nonexist_path)).to eq(ruby_expanded)
+      end
+
+      it "should expand a shortened path completely, unliked Ruby File.expand_path" do
+        tmp_long_dir = tmpdir('super-long-thing-that-Windows-shortens')
+        short_path = Puppet::Util::Windows::File.get_short_pathname(tmp_long_dir)
```
Iristyle: I implemented Windows `GetShortPathName` API solely for this test.  @puppetlabs/windows any thoughts about that?

Iristyle:
Thanks @hlindberg - it took me a bit of time to understand precisely where instances of things we care about are created.  I used 10k+ lines of debug output to narrow down the code paths and reduced that to common paths identified here - https://gist.github.com/Iristyle/9389f62d60a402f861df8a98012535c9

What I found was that an `Environment` was either created through `Environments`, or in some cases, created more directly through the `Environment` factory function `Environment.create` (some of the settings and CLI code does it that way).  As long as the correct path was in the `Environment` to start, then all of the module code loaders behaved correctly; module loaders are never created directly outside of tests.  Do you ever think module loading will be a public API?  I'd like to add a test *somewhere* so that if anything changes with respect to that we have some level of insurance against behavioral breakage, if possible.

hlindberg:
Seems to have legit test failures

Iristyle:
Hmm.. Travis and AppVeyor have pulled a disappearing act.

Iristyle:
Final call @hlindberg before I squash d275362 into a99ed7c ... ?

lib/puppet/util/windows/file.rb
```diff
@@ -257,6 +257,45 @@ def readlink(link_name)
...
+          raise Puppet::Util::Windows::Error.new("Failed to call GetShortPathName")
+        end
+
+        converted = converted_ptr.read_wide_string(buffer_size - 1)
+      end
+    end
+
+    converted
+  end
+  module_function :get_short_pathname
```
ferventcoder: Was this just added for parity sake?

ferventcoder: I mean it has the test coverage and I see that it is used in specs, but I was curious if it had any prod code use?

glennsarti: It's needed in one of the spec tests.

But it may come in handy for customers if they want the short pathname.

glennsarti: In short, no prod code use. Yet.

ferventcoder: Yeah, I'm glad that is does also have its own tests, otherwise I would have voted for the removal until needed.

ferventcoder:
This looks good to me.

lib/puppet/util/windows/file.rb
Iristyle: There are multiple places where `get_short_pathname` is essential to the tests.

hlindberg:
@Iristyle I am not sure I follow the squashing. Happy to review something that is coherent :-). The code in this PR looks ok now - if that helps.

lib/puppet/util/windows/file.rb
ferventcoder: Right, and I mentioned that [here](https://github.com/puppetlabs/puppet/pull/5156#discussion_r75538345). I was just curious if it had any use in the non spec code. Mostly just an observation, not asking for it's removal.

PR #5159, opened 2016-07-29
(PUP-6459) Update win32-service to 0.8.8

Iristyle:
- Version 0.8.7 of the win32-service gem had a bug that prevented it
   from properly retrieving all services on Windows 10 with the
   Win32::Service.services call.

   In particular, the services pvhdparser and vhdparser don't have
   delayed start info, which would previously cause errors.

   This is fixed in the 0.8.8 version of the gem


geoffnichols:
👍

Iristyle:
Actually - don't merge this yet - line endings are bungled.

MikaelSmith:
Is stable the branch targeting Puppet 4.6.0 now?

PR #5160, opened 2016-07-29
(PUP-6459) Acceptance test for 'puppet resource service'

james-stocks:
Add a basic acceptance test to ensure that 'puppet resource service' does not error.

[skip ci]

### DIFF
--- a/acceptance/tests/resource/service/should_query_all.rb
+++ b/acceptance/tests/resource/service/should_query_all.rb
@@ -1,0 +1,9 @@
+test_name "should query all services"
+
+agents.each do |agent|
+  step "query with puppet"
+  on(agent, puppet_resource('service'), :accept_all_exit_codes => true) do
+    assert_equal(exit_code, 0, "'puppet resource service' should have an exit code of 0")
+    assert(/^service/ =~ stdout, "'puppet resource service' should present service details")
+  end
+end


acceptance/tests/resource/service/should_query_all.rb
```diff
@@ -0,0 +1,11 @@
+test_name "should query all services"
+
+agents.each do |agent|
+  next if agent == master
```
Iristyle: Was there a reason to skip the masters agent here?

acceptance/tests/resource/service/should_query_all.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  step "query with puppet"
+  on(agent, puppet_resource('service'), :accept_all_exit_codes => true) do
+    assert_equal(exit_code, 0, "'puppet resource service' should have an exit code of 0")
+    assert_match(stdout, /^service/, "'puppet resource service' should present service details")
```
MosesMendoza: If puppet issues some other output prior to the service listing, ie. a deprecation of some kind, will this still match?

acceptance/tests/resource/service/should_query_all.rb
```diff
@@ -0,0 +1,9 @@
+test_name "should query all services"
+
+agents.each do |agent|
+  step "query with puppet"
+  on(agent, puppet_resource('service'), :accept_all_exit_codes => true) do
+    assert_equal(exit_code, 0, "'puppet resource service' should have an exit code of 0")
+    assert(/^service/ =~ stdout, "'puppet resource service' should present service details")
```
james-stocks: @MosesMendoza I tested, this will match any line in the output. So it's asserting that we have at least one line starting with "service"

MosesMendoza: 👍

Iristyle:
👍 thanks @james-stocks !

PR #5161, opened 2016-07-31
(PUP-5948) Add code generator for resource types

hlindberg:
This adds the "generate" face based application to puppet (written by @peterhuene). This PR also contains the integration work (changes to generator and the implementation for which the generator generates code), as well as a number of unit and integration tests.


lib/puppet/generate/models/type/property.rb
```diff
@@ -0,0 +1,70 @@
...
+                strings << Util::to_puppet_string(a.to_s)
+              end
+            end
+
+            # If no string or regexes, default to Any type
+            return 'Any' if strings.empty? && regexes.empty?
+
+            # Calculate a variant of supported values
+            enum = "Enum[#{strings.join(', ')}]" unless strings.empty?
+            pattern = "Pattern[#{regexes.join(', ')}]" unless regexes.empty?
```
thallgren: This is nit picking, but the `enum` and `pattern` variables are uninitialized here when their corresponding collections are empty. While uninitialized and `nil` is the same, I still think it's prudent to explicitly initialize to nil, i.e.
```
enum = strings.empty? ? nil : "Enum[#{strings.join(', ')}]"
```
The reason I think this is important is readability (reader don't need to scan the code above to verify that `enum` doesn't retain some other value in case strings is empty).

lib/puppet/generate/templates/type/pcore.erb
```diff
@@ -0,0 +1,40 @@
...
+Puppet::PCore::Resource.new({
+  name => <%= name %>,
+  properties => [
+<%- properties.each_with_index do |property, index| -%>
+  <%- unless property.doc.empty? -%>
+    <%- property.doc.each_line do |line| -%>
+    # <%= line -%>
+    <%- end %>
+  <%- end -%>
+    Param(<%= property.type %>, <%= property.name %><% if property.is_namevar? %>, true<% end %>)<% if index + 1 < properties.size %>,
```
thallgren: `Param(` must be `Puppet::Resource::Param.new(`

lib/puppet/generate/templates/type/pcore.erb
```diff
@@ -0,0 +1,40 @@
...
+<%- end %>
+  ],
+  parameters => [
+<%- parameters.each_with_index do |parameter, index| -%>
+  <%- unless parameter.doc.empty? -%>
+    <%- parameter.doc.each_line do |line| -%>
+    # <%= line -%>
+    <%- end %>
+  <%- end -%>
+    Param(<%= parameter.type %>, <%= parameter.name %><% if parameter.is_namevar? %>, true<% end %>)<% if index + 1 < parameters.size %>,
```
thallgren: `Param(` must be `Puppet::Resource::Param.new(`

lib/puppet/generate/templates/type/pcore.erb
```diff
@@ -0,0 +1,40 @@
...
+
+    <%- end -%>
+<%- end %>
+  ],
+  title_patterns => {
+<%- title_patterns.each_with_index do |mapping, index| -%>
+    <%= mapping[0] %> => [<%= mapping[1].join(', ')%>]<%= "," if index + 1 < title_patterns.size %>
+<%- end -%>
+  }
```
thallgren: The keys in this initializer hash can be left out when their corresponding hash/array is empty.

lib/puppet/generate/templates/type/pcore.erb
```diff
@@ -0,0 +1,40 @@
...
+})
```
thallgren: The keys in this initializer hash can be left out when their corresponding hash/array is empty.
Also, I'm missing logic for the `isomorphic` key.

lib/puppet/generate/type.rb
```diff
@@ -0,0 +1,200 @@
...
+          Puppet::FileSystem::exist?(output_path) && (Puppet::FileSystem::stat(@path) <=> Puppet::FileSystem::stat(output_path)) <= 0
+        end
+
+        # Gets the path to the output file.
+        # @return [String] Returns the path to the output file.
+        def output_path
+          @output_path ||=
+            case @format
+            when :pcore
+              File.join(@base, 'pcore', 'types', "#{File.basename(@path, '.rb')}.pp")
```
thallgren: Should be `'.resourcetypes'` instead of `'pcore/types'`.

lib/puppet/generate/util/helpers.rb
```diff
@@ -0,0 +1,18 @@
...
+    module Util
+      # Converts a Ruby string to a Puppet string literal.
+      # @param string [String] The string to convert to a Puppet string literal.
+      # @return [String] Returns the Puppet string literal (may not be quoted if the input is a bare word).
+      def self.to_puppet_string(string)
+        return string if !Puppet::Pops::Parser::Lexer2::KEYWORDS.keys.include?(string) &&
+                         string =~ Puppet::Pops::Parser::Lexer2::PATTERN_BARE_WORD
+        quoted = string.inspect
+        return "'#{string}'" if !string.include?('\'') && "\"#{string}\"" == quoted
+        quotede
```
thallgren: typo `quotede`.

lib/puppet/generate/templates/type/pcore.erb
```diff
@@ -0,0 +1,41 @@
...
+Puppet::Resource::ResourceType3.new(
+  <%= name %>,
+  [
+<%- properties.each_with_index do |property, index| -%>
+  <%- unless property.doc.empty? -%>
+    <%- property.doc.each_line do |line| -%>
+    # <%= line -%>
+    <%- end %>
+  <%- end -%>
+    Puppet::Resource::Param(<%= property.type %>, <%= property.name %><% if property.is_namevar? %>, true<% end %>)<% if index + 1 < properties.size %>,
```
thallgren: Should be `Puppet::Resource::Param.new(`

lib/puppet/generate/templates/type/pcore.erb
```diff
@@ -0,0 +1,41 @@
...
+<%- end %>
+  ],
+  [
+<%- parameters.each_with_index do |parameter, index| -%>
+  <%- unless parameter.doc.empty? -%>
+    <%- parameter.doc.each_line do |line| -%>
+    # <%= line -%>
+    <%- end %>
+  <%- end -%>
+    Puppet::Resource::Param(<%= parameter.type %>, <%= parameter.name %><% if parameter.is_namevar? %>, true<% end %>)<% if index + 1 < parameters.size %>,
```
thallgren: Should be `Puppet::Resource::Param.new(`

lib/puppet/generate/util/helpers.rb
```diff
@@ -0,0 +1,18 @@
+require 'puppet/pops/parser/lexer2'
+
+module Puppet
+  module Generate
+    module Util
+      # Converts a Ruby string to a Puppet string literal.
+      # @param string [String] The string to convert to a Puppet string literal.
+      # @return [String] Returns the Puppet string literal (may not be quoted if the input is a bare word).
+      def self.to_puppet_string(string)
```
thallgren: Why not use the `StringConverter` and '%p' ?

lib/puppet/generate/models/type/property.rb
```diff
@@ -0,0 +1,72 @@
...
+            end
+
+            # If no string or regexes, default to Any type
+            return 'Any' if strings.empty? && regexes.empty?
+
+            # Calculate a variant of supported values
+            # Note that boolean strings are mapped to Variant[Boolean, Enum['true', 'false']]
+            # because of tech debt...
+            enum = "Enum[#{strings.join(', ')}]" unless strings.empty?
+            pattern = "Pattern[#{regexes.join(', ')}]" unless regexes.empty?
```
thallgren: This is nit picking, but the `enum` and `pattern` variables are uninitialized here when their corresponding collections are empty. While uninitialized and `nil` is the same, I still think it's prudent to explicitly initialize to nil, i.e.
```
enum = strings.empty? ? nil : "Enum[#{strings.join(', ')}]"
```
The reason I think this is important is readability (reader don't need to scan the code above to verify that `enum` doesn't retain some other value in case strings is empty).

lib/puppet/generate/templates/type/pcore.erb
hlindberg: why? you can call the type that way to call new.

lib/puppet/generate/util/helpers.rb
hlindberg: It did not work ok when peter wrote this. Can be changed.

lib/puppet/generate/models/type/property.rb
hlindberg: sure will change that

lib/puppet/generate/util/helpers.rb
hlindberg: Changing to use StringConverter would make this problem go away too. Will make that change.

PR #5162, opened 2016-08-01
(PUP-6568) handle non-plist binary blobs in launchd service provider

ccaviness:
If the launchd service provider encounters a file that's not plain text or a binary plist, it will error.

Catch the ArgumentError when trying to pattern match against bad_xml_doctype, and return nil.


HAIL9000:
👍 from me and @Magisus

branan:
LGTM too. Let's do it!

PR #5163, opened 2016-08-01
(PUP-6570) Remove English MCO assert

glennsarti:
This commit removes the English STDOUT assertion. Now that i18n
work is being integrated into Puppet products, English assertions
are no longer universally valid.

This test was failing when run against non-English versions of
Windows due to the localized translation of the service start/stop
messages. The test will now rely on the exit code to determine
success.

This commit duplicates commit `405e75a` from `pe_acceptance_tests`
repository.

### DIFF
--- a/acceptance/tests/windows/QA-563_windows_exit_mcollective.rb
+++ b/acceptance/tests/windows/QA-563_windows_exit_mcollective.rb
@@ -52,9 +52,7 @@ end
 step 'Shutdown MCollective service on Windows agent'
 #Shutdown MCollective Service on Windows agent and make sure it successfully exits
 agents.each do |agent|
-  on agent, 'net stop mcollective' do |result|
-    assert_match(/The Marionette Collective Server service was stopped successfully/, result.stdout, "Failed to stop MCollective service")
-  end
+  on agent, 'net stop mcollective'
 end
 
 sleep 5
@@ -62,7 +60,5 @@ sleep 5
 step 'Start MCollective service on Windows agent'
 #Bring the MCllective backup
 agents.each do |agent|
-  on agent, 'net start mcollective' do |result|
-    assert_match(/The Marionette Collective Server service was started successfully/, result.stdout, "Failed to start MCollective service")
-  end
+  on agent, 'net start mcollective'
 end


acceptance/tests/windows/QA-563_windows_exit_mcollective.rb
```diff
@@ -53,7 +53,7 @@
 #Shutdown MCollective Service on Windows agent and make sure it successfully exits
 agents.each do |agent|
   on agent, 'net stop mcollective' do |result|
-    assert_match(/The Marionette Collective Server service was stopped successfully/, result.stdout, "Failed to stop MCollective service")
+    on agent, 'net stop mcollective'
```
kylog: This leaves two invocations of `on agent, 'net stop mcollective'`, the second inside the block for the first. Did you intend to replace the entire 3-line block with a single invocation (a la the change below)?

glennsarti: Yes you are right.  Thanks @kylog I'll fix that up.

acceptance/tests/windows/QA-563_windows_exit_mcollective.rb
```diff
@@ -52,17 +52,13 @@
 step 'Shutdown MCollective service on Windows agent'
 #Shutdown MCollective Service on Windows agent and make sure it successfully exits
 agents.each do |agent|
-  on agent, 'net stop mcollective' do |result|
-    assert_match(/The Marionette Collective Server service was stopped successfully/, result.stdout, "Failed to stop MCollective service")
-  end
+  on agent, 'net stop mcollective'
```
Iristyle: Shouldn't we at least be checking an exit status?

glennsarti: The default is to only accept exit code 0

Iristyle: @glennsarti has pointed out that Beaker defaults to only accepting exit code of `0` per https://github.com/puppetlabs/beaker/blob/dbc5be2d107f58373dbf4a9942a925a1a8313e7b/docs/concepts/shared_options_for_executing_beaker_commands.md#acceptable_exit_codes

I think it might still be a bit more self-documenting to explicitly specify them in the test, so that it's clear an actual assertion happens.

PR #5164, opened 2016-08-02
(PUP-6558) Fix problem when loading UDRTs in non-directory environment

thallgren:
The PE-installer uses a non-directory environment during the install
of some modules. The Runtime3xLoader didn't handle this very well since
it then tried to cache a loaded user defined resource type in a
NullLoader instance which resulted in a `nil` return value. As a,
consequence, the loaded resource type was lost.

This commit fixes the loader so that it handles the case of not having
an environment path. It then skips the attempt to load a pcore type
definition and will always cache the value in the Runtime3Loader (as it
did before the pcore resource type definitions were introduced).


hlindberg:
Shame on the installer !! If it is compiling a catalog a directory based environment should really be required, there are no other types of environments supported for compilation. Amazed that it kept working with non directory environments being deprecated since 4.0.0.

PR #5165, opened 2016-08-02
(PUP-6572) Codify the `site:modules:$basemodulepath` pattern as default

binford2k:
For a very long time, CS has been recommending and educating the
`modulepath = site:modules:$basemodulepath` pattern via
`environment.conf`. It's now codified as a CS standard:

https://github.com/puppetlabs/cs-standards/blob/master/control_repo_contents.md#directory-structure-and-contents

It's always felt a bit weird to be explaining to people why the best
practice wasn't the Puppet default, so how about we make it the default?
This should not impact existing users unless they're currently using
`$environment/site` for other purposes, and I'd wager that percentage is
rather low.

### DIFF
--- a/lib/puppet/settings/environment_conf.rb
+++ b/lib/puppet/settings/environment_conf.rb
@@ -99,7 +99,7 @@ class Puppet::Settings::EnvironmentConf
   end
 
   def modulepath
-    default_modulepath = [File.join(@path_to_env, "modules")] + @global_module_path
+    default_modulepath = [File.join(@path_to_env, "site")] + [File.join(@path_to_env, "modules")] + @global_module_path
     get_setting(:modulepath, default_modulepath) do |modulepath|
       path = modulepath.kind_of?(String) ?
         modulepath.split(File::PATH_SEPARATOR) :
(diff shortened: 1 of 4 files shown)


npwalker:
I'd probably be safer to go with modules:site... as is the case in 

https://github.com/puppetlabs/control-repo/blob/production/environment.conf#L1

adrienthebo:
I'm not convinced that this is a backwards compatible. People may for whatever reason have site directories that they didn't intend to include in the modulepath, and this change will suddenly include that directory. That being said this case is should be an extreme rarity and will generally be helpful for the majority of our users. So... +0?

joshcooper:
I share @adrienthebo's concern that this may not be backwards compatible. If it's not, then we will need to wait until Puppet 6. @binford2k/@npwalker can you provide context?

npwalker:
Yea I think it's technically not backward compatible.  If it weren't a hassle I'd say we should set it up on new installs and not touch environment.conf on upgrades.  

Maybe when we make this change we can switch to laying down an environment.conf.example instead of the actual config file?

MosesMendoza:
@binford2k we also have a ticket to add a flag behind which we gate breaking / puppet 6 targeted features, PUP-7677. Do you think this would go behind that flag cleanly?

MosesMendoza:
This PR hasn't been updated in some time, and it appears we're unlikely to be able to pick it up again in the near future. I'm going to close this for now, but if/when the team picks up PUP-6572 we can re-open this.

PR #5166, opened 2016-08-03
(PUP-6497) Improve error message with conflicting capability resources

nicklewis:
When a capability resource is moved from one node to another, either due
to a code change or because an application component is classified to a
different node, the database can contain multiple resource instances
matching the capability ref. This happens when the new node has run
puppet but the old node hasn't. In those cases, we try to decide which
resource to use by repeating our search but limiting it to only
resources with the same code_id as the catalog we're currently
compiling.

When the code_id search doesn't find _any_ matching resources, we were
previously returning `nil` and the Catalog#resource method was raising
an error indicating the capability couldn't be found and may not have
been produced yet. This is misleading, since the problem is actually
that there are _too many_ copies of the resources, not that it's
missing. We now instead raise an error indicating that we found a
conflict in our initial search.

### DIFF
--- a/lib/puppet/parser/resource.rb
+++ b/lib/puppet/parser/resource.rb
@@ -270,7 +270,7 @@ class Puppet::Parser::Resource < Puppet::Resource
       t = Puppet::Pops::Evaluator::Runtime3ResourceSupport.find_resource_type(scope, t) unless t == 'class' || t == 'node'
       cap = catalog.resource(t, ref.title)
       if cap.nil?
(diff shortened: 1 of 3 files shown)


nicklewis:
Rebased and conflict free!

joshcooper:
Sorry @nicklewis, but needs a rebase again. I'll merge this time, promise!

nicklewis:
@joshcooper Rebased again!

PR #5167, opened 2016-08-03
(PUP-6549) add libnghttpd2 for fedora23 acceptance

johnduarte:
Upstream httpd package requires an updated libnghttp2 package but is not
defined as such.
https://bodhi.fedoraproject.org/updates/httpd-2.4.23-4.fc23
https://bugzilla.redhat.com/show_bug.cgi?id=1358875


shermdog:
:+1:

branan:
Should we maybe do this for all Fedora (or redhat osfamily?) platforms? If we know that Redhat has incorrect dependencies on these it could come up and bite us on basically any other OS version when they change the versions of these libraries.

johnduarte:
Using `--best` is DNF specific for fedora23 and does not play nicely with other RedHat variants.

PR #5168, opened 2016-08-03
(PUP-6530) Add spec test that create_resources write file/line info

hlindberg:
This adds a spec test that asserts that create_resources writes
file/line information to all resources created by the call.
The file/line information is the file/line of the call expression.

The implementation of the behavior was merged in PUP-5948 as it adds the
logic required to compute the location, and a refactoring made all code
that creates resources follow one and the same logical path. The issue
PUP-6530 was fixed as a side effect (but no tests were then added for
this problem).


hlindberg:
The content of this PR is a test. Merging it myself.

PR #5170, opened 2016-08-03
(PUP-6366) Support $trusted facts with apache/unicorn or nginx/unicorn

HAIL9000:
Add support for $trusted facts work with apache/unicorn
or nginx/unicorn via X-SSL-Client-Cert variable.

Opened in favor of #4997 (due to the reformatted commit message)


joshcooper:
Since 3.x is EOL, we should target 4.0.x and add a test for this.

McdonaldSeanp:
Needs to be retargeted at 4.10.x, 5.3.x or master

MosesMendoza:
This PR hasn't been updated in some time, and it seems unlikely to get picked up in the near future. I'm going to close this for now, and we can re-open if/when PUP-6366 is prioritized for the 5.x series. We can track future conversation in that ticket.

PR #5173, opened 2016-08-04
(maint) Add test to assert that apply doesn't load pcore resource type

thallgren:
This commit contains a test that verifies that an apply on an already
existing catalog doesn't trigger a load of a pcore resource type.
The commit also adds an eql? method to the ResourceTypeImpl that
corresponds to its pcore description.


hlindberg:
Good to merge - not sure if we have master branch lookdown or not (marked Blocked). I am +1 on merging if there is no Block, or when block lifts.

PR #5176, opened 2016-08-04
(WIP)(PUP-6448) Indirector URIs should remain UTF-8

Iristyle:
- By not preserving the original Path encoding of UTF-8 used for
  incoming requests to the indirector, future string comparisons (using
  Regex for instance) are invalid and will cause exceptions.
  
  This problem is trivially reproducible with a manifest like:
  
  file { "/tmp/test":
    ensure => file,
    source => 'puppet://modules/utf_8/静的',
  }
  
  The key of the request previously was "utf_8/\xE9\x9D\x99\xE7\x9A\x84"
  instead of the appropriate "utf_8/静的"

### DIFF
--- a/lib/puppet/indirector/request.rb
+++ b/lib/puppet/indirector/request.rb
@@ -243,6 +243,8 @@ class Puppet::Indirector::Request
   def set_uri_key(key)
     @uri = key
     begin
+      # calling URI.escape for UTF-8 characters will % escape them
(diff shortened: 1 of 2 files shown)


Iristyle:
Needs tests, just a POC for now.

ferventcoder:
Wow.

MosesMendoza:
superseded by https://github.com/puppetlabs/puppet/pull/5462

PR #5177, opened 2016-08-05
(PUP-6585) Deprecate Puppet::Util.absolute_path?

Iristyle:
In an effort to validate some UTF-8, I noticed that we have this method kicking around that Ruby already implements.

I also noticed that based on it's current implementation, it's being abused and allowing a lot of bad calls against it, since it doesn't verify the incoming object is a String / Pathname.  This is an effort to clean up all the errant calls as well, which are failing / succeeding for the wrong reasons when they pass in log objects, nil, floats, etc.


lib/puppet/file_serving/base.rb
```diff
@@ -80,6 +80,9 @@ def to_data_hash
   end
 
   def self.absolute?(path)
-    Puppet::Util.absolute_path?(path, :posix) or (Puppet.features.microsoft_windows? and Puppet::Util.absolute_path?(path, :windows))
+    # TODO: this is the one spot that will be tricky
```
Iristyle: Need to find an answer for this, so that we don't need to perform multiple style `absolute_path?` checks that are OS agnostic.

lib/puppet/file_system.rb
```diff
@@ -212,6 +212,12 @@ def self.mkpath(path)
     @impl.mkpath(assert_path(path))
   end
 
+  def self.absolute_path?(path)
+    return false if path.nil?
```
Iristyle: Should really find all the current callers passing `nil` - their logic is likely flawed.

lib/puppet/util/log/destinations.rb
```diff
@@ -47,7 +47,11 @@ def handle(msg)
   require 'fileutils'
 
   def self.match?(obj)
-    Puppet::Util.absolute_path?(obj)
+    # TODO: what on earth is this supposed to be doing
```
Iristyle: Need to spend some time understanding this code path... it's passing in `LogCollector` instances.

spec/unit/util_spec.rb
```diff
@@ -259,33 +259,34 @@ def withenv_utf8(&block)
...
-        end
-      end
-
-      %w[/ . ./foo \foo /foo /foo/../bar //foo C:foo/bar foo//bar/baz].each do |path|
-        it "should return false for #{path}" do
-          expect(Puppet::Util).not_to be_absolute_path(path, :windows)
-        end
-      end
-    end
+    # TODO: we wouldn't need these if we could find a way to fake out the OS
```
Iristyle: The hope is that we don't need this anymore - but that relies on us being able to rework `file_serving/base`

Iristyle:
My quick feedback validation is through:

```
bundle exec rspec ./spec/integration/indirector/direct_file_server_spec.rb ./spec/integration/indirector/file_content/file_server_spec.rb ./spec/integration/indirector/file_metadata/file_server_spec.rb ./spec/integration/network/http/api/indirected_routes_spec.rb ./spec/integration/parser/scope_spec.rb ./spec/integration/type/file_spec.rb ./spec/unit/application/inspect_spec.rb ./spec/unit/configurer/downloader_spec.rb ./spec/unit/file_serving/base_spec.rb ./spec/unit/provider/exec/posix_spec.rb ./spec/unit/puppet_spec.rb ./spec/unit/resource/catalog_spec.rb ./spec/unit/settings/autosign_setting_spec.rb ./spec/unit/ssl/certificate_authority_spec.rb ./spec/unit/transaction/resource_harness_spec.rb ./spec/unit/type/exec_spec.rb ./spec/unit/type/file/checksum_value_spec.rb ./spec/unit/type/file/source_spec.rb ./spec/unit/type/file_spec.rb ./spec/unit/util/logging_spec.rb ./spec/unit/util_spec.rb
```

joshcooper:
Needs rebase and has CI failures

McdonaldSeanp:
needs to be retargeted at either 4.10.x, 5.3.x or master

lib/puppet/util/log/destinations.rb
joshcooper: It's trying to determine if this type of log destination should be used based on the specified log destination object. That way you can specify `--logdest console` vs `--logdest /path/to/log`, and the file based log destination will only match the latter. It's hacky but works given the logging API we have to work with.

PR #5178, opened 2016-08-05
(PUP-6182) Improve validation of component mappings in applications

nicklewis:
This pull request includes two improvements to how application node/component
mapping specifications are validated:
- Require `nodes` param for application declarations

Previously, the `nodes` parameter of applications was marked as optional in the
type system, but would cause a failure when processing the catalog and
attempting to iterate the nodes hash. Since this was always implicitly required
and there's no good reason for it not to always be specified, we now explicitly
require it so as to cause a clear, early failure.
- Fail when non-existent components have nodes assigned

Previously, non-existent components in the node mapping of an application would
simply be ignored. Because this frequently happens when the user incorrectly
types the component name, it can be especially confusing as the only complaint
will be that no node was mapped to the _correct_ component. We now explicitly
fail if there are non-existent components, and we do so before checking whether
there are unmapped components.

### DIFF
--- a/lib/puppet/network/http/api/master/v3/environment.rb
+++ b/lib/puppet/network/http/api/master/v3/environment.rb
@@ -13,40 +13,59 @@ class Puppet::Network::HTTP::API::Master::V3::Environment
 
     catalog = Puppet::Parser::EnvironmentCompiler.compile(env, code_id).to_resource
 
-    # This reads code_id off the catalog rather than using the one from the
-    # request. There shouldn't really be a case where the two differ, but if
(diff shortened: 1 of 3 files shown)


hlindberg:
Looks like the appveyor failure has nothing to do with the PR. Needs a respin.

nicklewis:
Rebased and the tests are passing now.

joshcooper:
@nicklewis sorry but there's a conflict, can you rebase?

nicklewis:
@joshcooper Rebased!

PR #5179, opened 2016-08-05
(PA-461) remove pre-suite setup of CA certs on Windows

MosesMendoza:
Prior to this commit, the Puppet presuite would install the certs needed to
connect to forge over PMT. This was needed as the Windows MSI did not include
certs. That changed in RE-6111, in which the cert bundle was included in the
MSI, but this pre-suite item remained. This commit removes it, because it is no
longer needed, and in PA-461 was found to be masking a misconfiguration in the
MSI which resulted in a regression in functionality.


MosesMendoza:
Closing and retargeting at stable

PR #5180, opened 2016-08-05
(PA-461) remove pre-suite setup of CA certs on Windows

MosesMendoza:
Prior to this commit, the Puppet presuite would install the certs needed to
connect to forge over PMT. This was needed as the Windows MSI did not include
certs. That changed in RE-6111, in which the cert bundle was included in the
MSI, but this pre-suite item remained. This commit removes it, because it is no
longer needed, and in PA-461 was found to be masking a misconfiguration in the
MSI which resulted in a regression in functionality.


MosesMendoza:
Turns out, this will likely fail as-is unless it's being run by packages.

Iristyle:
Hmmm... on second thought, not sure we can do this.

While we can rely on the CA bundle in tests that use MSI packages, I think we're going to have problems running `ci:test:git`, which needs them.

I think we can do this in a way that includes this pre-suite step *only* for `ci:test:git`

Iristyle:
Potentially we can just move `070` out of https://github.com/puppetlabs/puppet/tree/master/acceptance/setup/common/pre-suite and into https://github.com/puppetlabs/puppet/tree/master/acceptance/setup/git/pre-suite

PR #5182, opened 2016-08-06
W.I.P (PUP-6582) Loader bound functions must have nil closure scope

hlindberg:
@thallgren I got this working, but there is one spec test that is giving me problems. One test (changed) where giving me problems and there the solution was to just push a contect with global_scope set (at the right point). I could not quite figure out what was wrong with the still failing tests. Can you take a look at that?

### DIFF
--- a/lib/puppet/pops/evaluator/closure.rb
+++ b/lib/puppet/pops/evaluator/closure.rb
@@ -27,7 +27,7 @@ class Closure < CallableSignature
   # Evaluates a closure in its enclosing scope after having matched given arguments with parameters (from left to right)
   # @api public
   def call(*args)
-    call_with_scope(@enclosing_scope, args)
+    call_with_scope(enclosing_scope, args)
   end
 
   # This method makes a Closure compatible with a Dispatch. This is used when the closure is wrapped in a Function
@@ -35,7 +35,7 @@ class Closure < CallableSignature
   # checks of the argument type/arity validity).
   # @api private
   def invoke(instance, calling_scope, args, &block)
-    @enclosing_scope.with_global_scope do |global_scope|
+    enclosing_scope.with_global_scope do |global_scope|
       call_with_scope(global_scope, args, &block)
     end
   end
@@ -44,7 +44,7 @@ class Closure < CallableSignature
   def call_by_name(args_hash, enforce_parameters)
     if enforce_parameters
       # Push a temporary parameter scope used while resolving the parameter defaults
-      @enclosing_scope.with_parameter_scope(closure_name, parameter_names) do |param_scope|
+      enclosing_scope.with_parameter_scope(closure_name, parameter_names) do |param_scope|
         # Assign all non-nil values, even those that represent non-existent paramaters.
         args_hash.each { |k, v| param_scope[k] = v unless v.nil? }
         parameters.each do |p|
@@ -56,7 +56,7 @@ class Closure < CallableSignature
               # No default. Assign nil if the args_hash included it
               param_scope[name] = nil if args_hash.include?(name)
             else
-              param_scope[name] = param_scope.evaluate(name, p.value, @enclosing_scope, @evaluator)
+              param_scope[name] = param_scope.evaluate(name, p.value, enclosing_scope, @evaluator)
             end
           end
         end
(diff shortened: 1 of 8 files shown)


thallgren:
A puppet apply on this branch that executes the same code as the test also fails so there's nothing wrong with the test.
```
bundle exec bin/puppet apply -e 'type UnprivilegedPort = Integer[1024,65537]; function check_port(UnprivilegedPort $port) {}; check_port(34)'
Error: Evaluation Error: Resource type not found: UnprivilegedPort  at line 1:66 on node home.tada.se
```
I think this code actually relies on the fact that the function is bound to the global scope. How else should it find the type?

lib/puppet/pops/migration/migration_checker.rb
```diff
@@ -6,6 +6,10 @@ class Puppet::Pops::Migration::MigrationChecker
   def initialize()
   end
 
+  def self.singleton
+    @@null_checker ||= self.new
+  end
+
```
thallgren: Is this related to PUP-6582 ?

lib/puppet/pops/loader/puppet_function_instantiator.rb
```diff
@@ -64,18 +65,19 @@ def self.create(loader, typed_name, source_ref, pp_code_string)
...
 
-  def self.create_function_class(function_definition, closure_scope)
+  def self.create_function_class(function_definition)
     # Create a 4x function wrapper around a named closure
     Puppet::Functions.create_function(function_definition.name, Puppet::Functions::PuppetFunction) do
+      # This is highly problematic - it binds both an Evaluator and closure_scope
       init_dispatch(Evaluator::Closure::Named.new(
         function_definition.name,
-        Evaluator::EvaluatorImpl.new(), function_definition, closure_scope))
+        Evaluator::EvaluatorImpl.new(), function_definition, nil))
```
thallgren: @hlindberg, this line is the culprit that makes the test (and a puppet apply) fail. Changing this line to
```
        Evaluator::EvaluatorImpl.new(), function_definition, Puppet.lookup(:global_scope) { {} }))
```
makes it work (but does of course not fix the problem). Could it be that the global scope, at the time of the call, doesn't see the type definitions defined in the same file as the function definition? Perhaps the test is all wrong (the ducks are not in a row) since it contains a type definition, a function definition, and a call in one single file?

lib/puppet/pops/migration/migration_checker.rb
hlindberg: it should have been in the maint commit. Will move it.

lib/puppet/pops/loader/puppet_function_instantiator.rb
hlindberg: Yeah, while it fixes the failing test, it also brings back the problem this PR is trying to fix; i.e. Function and Closure instances that are bound by the loader cannot bind to their runtime closure scope since it will be different with each compilation.

spec/integration/parser/parameter_defaults_spec.rb
```diff
@@ -26,12 +26,16 @@
...
     def collect_notices(code)
       logs = []
       Puppet[:code] = code
       Puppet::Util::Log.with_destination(Puppet::Test::LogCollector.new(logs)) do
-        compiler.compile
-        yield
+        Puppet.override(:global_scope => topscope) do
+          compiler.compile
+          yield
+        end
```
thallgren: I think  a more accurate fix here would be:
```
compiler.compile do |catalog|
  yield
  catalog
end
```
The yield will then end up within the override performed by the compiler and no extra override is necessary.

lib/puppet/pops/evaluator/closure.rb
```diff
@@ -122,6 +122,18 @@ def initialize(name, evaluator, model, scope)
...
+    #
+    def enclosing_scope
+      # Named closures are typically used for puppet functions and they cannot be defined
+      # in an enclosing scope as they are cashed and reused. They need to bind to the
+      # global scope at time of use rather at time of definition.
+      # Unnamed closures are always a runtime construct, they are never bound by a loader
+      # and are thus garbage collected at end of a compilation.
+      #
+      super || Puppet.lookup(:global_scope) { {} }
+    end
```
thallgren: There is a lingering reference to the attribute @enclosing_scope at line 248. That's whats causing the problems in the unit tests.

thallgren:
Superseded by [PR-5183](/puppetlabs/puppet/pull/5183)

lib/puppet/pops/evaluator/closure.rb
hlindberg: fixed

PR #5183, opened 2016-08-08
(PUP-6582) Remove binding of loaders to AST model elements

thallgren:
Before this commit, 4x loaders were bound to AST model elements. That
was less than ideal since an AST model element has the same life cycle
as the environment. A new compile on the same environment would hence
find the loader created and attached to the adapter during the first
compile.

This commit retains the adapter but instead of attaching the actual
loader instance, it attaches the name of the loader. The loader is
then retrieved using `scope.compiler.loaders[loader_name]`

### DIFF
--- a/lib/puppet/pops/loader/loader.rb
+++ b/lib/puppet/pops/loader/loader.rb
@@ -29,7 +29,7 @@ class Loader
 
   # @param [String] name the name of the loader. Must be unique among all loaders maintained by a {Loader} instance
   def initialize(loader_name)
-    @loader_name = loader_name
+    @loader_name = loader_name.freeze
   end
 
   # Produces the value associated with the given name if already loaded, or available for loading
--- a/lib/puppet/pops/loader/loader.rb
+++ b/lib/puppet/pops/loader/loader.rb
@@ -22,9 +22,15 @@
 module Puppet::Pops
 module Loader
 class Loader
+  attr_reader :loader_name
 
   # Describes the kinds of things that loaders can load
   LOADABLE_KINDS = [:func_4x, :func_4xpp, :type_pp, :resource_type_pp].freeze
+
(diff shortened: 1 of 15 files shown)


lib/puppet/pops/loader/loader.rb
```diff
@@ -22,10 +22,16 @@
...
 module Loader
 class Loader
+  attr_reader :loader_name
 
   # Describes the kinds of things that loaders can load
   LOADABLE_KINDS = [:func_4x, :func_4xpp, :type_pp, :resource_type_pp].freeze
 
+  # @param [String] name the name of the loader. Must be unique among all loaders maintained by a {Loader} instance
+  def initialize(loader_name)
+    @loader_name = loader_name
```
hlindberg: Name should probably be frozen since it is a String (bad things will otherwise happen if the name is modified externally).

lib/puppet/pops/loaders.rb
```diff
@@ -102,6 +106,16 @@ def self.loaders
     loaders
   end
 
+  def [](loader_name)
+    loader = @loaders_by_name[loader_name]
+    if loader.nil?
+      # Unable to find the module private loader. Try resolving the module
+      loader = private_loader_for_module(loader_name[0..-9]) if loader_name.end_with?(' private')
```
hlindberg: This seems a bit brittle as it depends on an undocumented (and not enforced) naming convention.

hlindberg: resolved by documentation

hlindberg:
good to merge when branches open up (may want this to go to stable at that point though).

thallgren:
Superseded by [PR-5192](https://github.com/puppetlabs/puppet/pull/5192) which is rebased on stable

PR #5184, opened 2016-08-08
(PUP-6087) Error when a capability mapping is redefined.

adreyer:
Before this commit redefining a capability mapping added both mappings to
the capability list. When producing and consuming resources only the
first mapping would be used. Now trying to add the second mapping
results in a ParseError.

In addition we previously evaluated mappings twice. Once before
evaluating main and a second time after evaluating applications. The
capability mappings were not properly cleared after the first
evaluation. As a consequence capability mappings were added twice. It
seems better if we only only evaluate them once so that they are
available for producing resources at a consistent place in compilation.

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -169,8 +169,6 @@ class Puppet::Parser::Compiler
 
       activate_binder
 
-      Puppet::Util::Profiler.profile("Compile: Evaluated capability mappings", [:compiler, :evaluate_capability_mappings]) { evaluate_capability_mappings }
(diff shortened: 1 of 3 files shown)


nicklewis:
:+1:

nicklewis:
It wasn't clear what the process was to get it merged or whether it was safe to go in because of current releases...

kylog:
However, somewhere along the line between when this PR went up and today, it now doesn't pass specs. So reverted in #5289. Sigh.

So @adreyer or @nicklewis, when you get a chance take another look.

PR #5185, opened 2016-08-09
(PUP-5623) Add return type to Callable type (functions and lambdas)

thallgren:
This PR adds a `return_type` attribute to the `Callable` Puppet type
and the necessary syntax additions to support it, both in the Puppet
language parser and in the ruby function dispatchers. Tests are also
added to validate that the return type is used for validating the
returned value from lambdas, puppet functions, and ruby functions.


lib/puppet/functions.rb
```diff
@@ -466,6 +465,15 @@ def optional_block_param(*type_and_name)
       @block_type = Puppet::Pops::Types::TypeFactory.optional(@block_type)
     end
 
+    # Defines the return type. Defaults to 'Any'
+    # @param [String] type a reference to a Puppet Data Type
+    #
+    # @api public
+    def return_type(type)
```
hlindberg: This is a setter so method should be `return_type=`

spec/unit/pops/evaluator/evaluating_parser_spec.rb
```diff
@@ -1040,6 +1040,12 @@ def assert_no_undef(x)
       expect{parser.evaluate_string(scope, "assert_no_undef({1 => undef})")}.to_not raise_error()
     end
 
+    it 'a lambda return value is checked using the return type' do
+      expect(parser.evaluate_string(scope, "[1,2].map |Integer $x| >> Integer { $x }")).to eql([1,2])
+      expect { parser.evaluate_string(scope, "[1,2].map |Integer $x| >> String { $x }") }.to raise_error(
+        /value returned from lambda had wrong type, expected a String value, got Integer/)
+    end
+
```
hlindberg: example that when return is not specified makes the return type default to Any is missing (there should be an explicit test even if all other examples do not have return type specified).

hlindberg: an example that shows that 3.x functions are allowed to return Any is also warranted (to make behavior explicit).

spec/unit/pops/parser/parse_functions_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      expect(dump(parse('function foo() { 1 }'))).to eq("(function foo (block\n  1\n))")
+    end
+  end
+
+  context 'with return type' do
+    it 'function foo() >> Integer { 1 }' do
+      expect(dump(parse('function foo() >> Integer { 1 }'))).to eq("(function foo (return_type integer) (block\n  1\n))")
+    end
+  end
+end
```
hlindberg: Add the corresponding tests for lambda ?

spec/unit/pops/types/type_parser_spec.rb
```diff
@@ -272,6 +272,18 @@ module Types
...
+  end
+
+  it 'parses a parameterized callable type with min/max and return type' do
+    expect(parser.parse("Callable[[String, Integer, 1, default],Float]")).to be_the_type(types.callable([String, Integer, 1, :default], Float))
+  end
+
+  it 'parses a parameterized callable type with block and return type' do
+    expect(parser.parse("Callable[[String, Callable[Boolean]],Float]")).to be_the_type(types.callable([String, types.callable(true)], Float))
+  end
+
```
hlindberg: is `'Callable[[], Integer]'` acceptable?

hlindberg:
How does type inference work when there are two callables with different return type?

hlindberg:
Can different dispatchers define different return type in a 4.x ruby function? (I think so). In that case, how is the return_type of the function as such described? (maybe we do aggregate the overloading anywhere).

thallgren:
Not sure what you mean by "type inference". A callable can be inferred from a Proc but the Proc is ruby and hence untyped (hence, the return_type is Any). If you mean choice of dispatcher, then that choice is solely based on parameter values (the return value doesn't exist at the time of choosing).

Each dispatcher defines its return_type. Is that what you mean by "function as such"?

spec/unit/pops/evaluator/evaluating_parser_spec.rb
thallgren: Any suggestion on how to write a test that shows that there is no restriction on the returned value?

hlindberg: As a unit test: That it checks against Any when not specified perhaps. As an integration (behavioral) test: that it can return a couple of different types (that are disjunct,  say a Hash and an Integer - depending on value of a given parameter).

hlindberg:
I mean if you have an array of Callable, what is the resulting Array type when inferring something like `[Callable[[], Integer], Callable[[],String]` ?

hlindberg:
Regarding function with multiple dispatches - it basically has multiple return types (depending on picked dispatcher) - do we ever describe such a function with one Callable?

thallgren:
We don't describe functions with callables AFAIK. A dispatcher can produce a callable.

spec/unit/pops/evaluator/evaluating_parser_spec.rb
thallgren: It doesn't check against Any when not specified. It doesn't check at all. It _could_ check against Any of course, and that would have been just as correct but whether or not it does is an implementation detail and I see no value in writing a unit test with an expectation for a certain type of check (or lack thereof).

I'm not comfortable writing a test using some arbitrary types either. There are already a lot of tests that calls methods returning all kinds of objects (all function calls that take in the test framework). If some unexpected test using some unexpected type would take place when a function returned, then hundreds of tests would break. So what would be the point in adding yet another test?

thallgren:
The common type of two Callable types is computed using assignability. The most generic one is chosen. If neither can be assigned to the other, then the default callable is chosen. There is no attempt to compute common parameter types (or return_type).

spec/unit/pops/evaluator/evaluating_parser_spec.rb
hlindberg: The tests serves as examples and specifies what the requirements are. `it "allows any type when return type is not specified"...` It also helps someone reviewing the functionality and trying to figure out if the acceptance criteria are met or not. (Other than that I agree with you - nothing much would work if the requirement is not met).

PR #5186, opened 2016-08-09
(PUP-6596) Update CFPropertyList gem to 2.3.5 and add plain format to puppet/util/plist

keeleysam:
https://tickets.puppetlabs.com/browse/PUP-6596

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -29,7 +29,7 @@ gem_rdoc_options:
 gem_platform_dependencies:
   universal-darwin:
     gem_runtime_dependencies:
-      CFPropertyList: '~> 2.2.6'
+      CFPropertyList: '~> 2.3.5'
   x86-mingw32:
     gem_runtime_dependencies:
       # Pinning versions that require native extensions
--- a/tasks/cfpropertylist.rake
+++ b/tasks/cfpropertylist.rake
@@ -1,6 +1,6 @@
 task 'cfpropertylist' do
   if defined? Pkg::Config and Pkg::Config.project_root
-    cfp_version = "2.2.7"
+    cfp_version = "2.3.5"
     libdir = File.join(Pkg::Config.project_root, "lib")
     source = "https://github.com/ckruse/CFPropertyList/archive/cfpropertylist-#{cfp_version}.tar.gz"
     target_dir = Pkg::Util::File.mktemp
--- a/lib/puppet/util/plist.rb
+++ b/lib/puppet/util/plist.rb
@@ -114,6 +114,8 @@ module Puppet::Util::Plist
         plist_format = CFPropertyList::List::FORMAT_XML
       elsif format.to_sym == :binary
         plist_format = CFPropertyList::List::FORMAT_BINARY
+      elsif format.to_sym == :plain
+        plist_format = CFPropertyList::List::FORMAT_PLAIN
       else
         raise FormatError.new "Unknown plist format #{format}"
       end


keeleysam:
The tests are failing because project_data.yaml declares a dependence on Facter 2, which also included this old version.  I'm not sure what the best way forward here is - shouldn't tests be running with Facter 3 only?

branan:
Running ruby unit tests against facter 3 is a pain, since it requires compilation of the facter library. This means we still use Facter 2 in Travis and Appveyor.

Beyond that technical issue, though, we do still consider Puppet 4 to be compatible with Facter 2. That probably means that we would need to do a lockstep update/release of both Facter 2 and Puppet in order to update cfpropertylist - that seems unlikely to happen just for the sake of this ticket

branan:
If you can come up with a compatible way forward to get the gems updated I would really love to hear it, though - blocking the gem update and a useful feature because of weird old compatibility issues bums me out

keeleysam:
@branan We've discovered a big issue with how strings are handled by the current version of the CFPropertyList gem - the fix has been released in 2.3.3. I've also opened a PR on Facter 2 so this can be coordinated.  https://github.com/puppetlabs/facter/pull/1420

er0ck:
@keeleysam FYI: we're working on this on our end.  we haven't forgotten about it

ccaviness:
Any update?

MosesMendoza:
👍

branan:
I'd like to try to get a Facter 3 gem put together, so we can drop support for Facter 2 in Puppet 5. This would allow us to roll this update in without needing a lockstep release of Facter.

Whether that facter gem is going to happen is still in the air. It depends on my workload on other issues during the Puppet 5 developent process.

ext/project_data.yaml
```diff
@@ -29,7 +29,7 @@ gem_rdoc_options:
 gem_platform_dependencies:
   universal-darwin:
     gem_runtime_dependencies:
-      CFPropertyList: '~> 2.2.6'
+      CFPropertyList: '~> 2.3.3'
```
Iristyle: Lets change this to 2.3.5 please

tasks/cfpropertylist.rake
```diff
@@ -1,6 +1,6 @@
 task 'cfpropertylist' do
   if defined? Pkg::Config and Pkg::Config.project_root
-    cfp_version = "2.2.7"
+    cfp_version = "2.3.3"
```
Iristyle: Similarly 2.3.5 here

Magisus:
Thanks for updating this. It's probably fine to squash these all into one commit with an updated commit message that references the 2.3.5 bump and the ticket number, to appease our Travis commit message check.

ext/project_data.yaml
```diff
@@ -29,7 +29,7 @@ gem_rdoc_options:
 gem_platform_dependencies:
   universal-darwin:
     gem_runtime_dependencies:
-      CFPropertyList: '~> 2.2.6'
+      CFPropertyList: '~> 2.3.5'
```
Iristyle: Travis is failing because of the gem conflict.  For now (and until we've cut the next Facter 2.5.0 gem), we're actually going to have to make Puppet less restrictive by changing this to:

```
      CFPropertyList: '~> 2.2'
```

As has been done in https://github.com/puppetlabs/facter/pull/1569

Once Facter 2.5.0 is released to the wild, and before we ship a Puppet gem, we can make Puppet more restrictive once again like we've done here (though it may be beneficial to just let Puppet grab latest in 2 series if available).

Iristyle:
Superseded by #5904 - thanks for your contribution @keeleysam !

PR #5190, opened 2016-08-11
(PUP-6588) Fixed --module-repository to use full remote URL

jainishshah17:
Instead of just host and port now user can use full path (including context path) for remote repository.

### DIFF
--- a/lib/puppet/forge/repository.rb
+++ b/lib/puppet/forge/repository.rb
@@ -45,7 +45,7 @@ class Puppet::Forge
     # Return a Net::HTTPResponse read for this +path+.
     def make_http_request(path, io = nil)
       Puppet.debug "HTTP GET #{@host}#{path}"
-      request = get_request_object(path)
+      request = get_request_object(@uri.path.chomp('/')+path)
       return read_response(request, io)
     end
 


jainishshah17:
I have only 1 line change in the code which just fixed functionality.

jainishshah17:
@branan: I will work on adding unit test for this. Do you have any other recommendation for this PR?

jainishshah17:
@branan : There is no test which uses `--module_repository` parameter. I don't know how to write unit test in this case. I have made changes in my code which takes care of leading/trailing slashes.
Please take a look

MikaelSmith:
The repository_spec tests are where this would be covered. I'm taking a look at it; functionally I wonder if this implies a new setting like `http_proxy_path`; currently `http_proxy_host`, `http_proxy_port`, `http_proxy_user`, and `http_proxy_password` can be used to configure proxies. They're overridden by setting the HTTP_PROXY environment variable.

MikaelSmith:
Ah, got mixed up looking at this solution and using http_proxy to work-around it. Nevermind modifying how proxy works.

jainishshah17:
@MikaelSmith Thank you

MikaelSmith:
Closing in favor of https://github.com/puppetlabs/puppet/pull/5600.

PR #5193, opened 2016-08-11
(WIP)(PUP-6184) Implement Windows SHGetKnownFolderPath

Iristyle:
- Replace win32-dir helpers with `SHGetKnownFolderPath` API that uses FFI.
  
  This code is being added now so that internal usage of Dir
  constants may be removed in the future.


Iristyle:
This will ultimately cause `win32-dir` gem to be removed... but is only partially done as of now.  

Was originally part of the effort in #5150 - pull it out to its own PR to finish up.

MikaelSmith:
What's still WIP on this?

MikaelSmith:
Ping. Should I put together an updated PR with my suggestions?

Iristyle:
Yeah, need to actually use `get_known_folder_path` somewhere.  I think we'll try to push this over the line after puppet-agent 1.8 ships... at least as a starting point towards any rework around `win32-dir` removal.

joshcooper:
ping @Iristyle, given de-prioritization of win32 gem removal, should we close this?

melissa:
Hey @Iristyle and @puppetlabs/platform-os, could we get some eyes on this? Should this work be picked back up, or should we close this?

McdonaldSeanp:
With the recent work platform OS has done to aid in the removal of win32 gems I think this is still something we should get in, if we have time/priority.

melissa:
Would it be worth it to leave this open or would it makes sense to close it and either re-open or create a new PR when we get around to doing this work?

joshcooper:
Given the conflicts and lack of movement on the PR and ticket, I'm going to close this. But feel free to reopen if you have time to revisit. I'll also keep the ticket open so we'll preserve the link to this PR in the future.

PR #5195, opened 2016-08-12
(PUP-6619) add tests for installdir and version

MosesMendoza:
(PA-466) This PR adds a new test to ensure the presence of the version file as laid down by the AIO in accordance with the puppet agent path specification after a recent set of changes to packaging of the AIO impacted the VERSION file creation and a regression was discovered in associated logic. This is added as a new test (as opposed to an additional step in 'ensure_puppet-agent_paths') to take advantage of the benefits of `skip_test` since this test only applies to
packaged artifacts and not git/gem which don't create a VERSION file. 

This PR also adds a test to ensure the fact env_windows_installdir is present and set to the correct directory. This fact is set via the environment.bat file on Windows installations and recently was impacted by a regression in that file, so a test is warranted. This test is added as its own file as opposed to the 4654_facts_face.rb file because that test seemed in spirit to be testing the functionatlity of `puppet facts` rather than any specific fact, while this test is for an explicit fact. This test is confined to Windows and test runs running from AIO packages (PE/FOSS) as those
environments are expected to have sourced a functional environment.bat.

### DIFF
--- a/acceptance/tests/ensure_version_file.rb
+++ b/acceptance/tests/ensure_version_file.rb
@@ -1,0 +1,30 @@
+require 'puppet/acceptance/temp_file_utils'
+extend Puppet::Acceptance::TempFileUtils
+
+# ensure a version file is created according to the puppet-agent path specification:
+# https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md
+
+test_name 'PA-466: Ensure version file is created on agent' do
+
+  skip_test 'requires version file which is created by AIO' if [:gem, :git].include?(@options[:type])
+
+  step "test for existence of version file" do
+    agents.each do |agent|
+      platform = agent[:platform]
+      ruby_arch = agent[:ruby_arch] || 'x86' # ruby_arch defaults to x86 if nil
+
+      if platform =~ /windows/
+        version_file = platform =~ /-64$/ && ruby_arch == 'x86' ?
+          "C:/Program Files (x86)/Puppet Labs/Puppet/VERSION" :
+          "C:/Program Files/Puppet Labs/Puppet/VERSION"
+      else
+        version_file = "/opt/puppetlabs/puppet/VERSION"
+      end
+
+      if !file_exists?(agent, version_file)
+        fail_test("Failed to find version file #{version_file} on agent #{agent}")
+      end
+    end
+  end
+end
+
--- a/acceptance/tests/windows/env_windows_installdir_fact.rb
+++ b/acceptance/tests/windows/env_windows_installdir_fact.rb
@@ -1,0 +1,22 @@
+# (PA-466) The env_windows_installdir fact is set as an environment variable
+# fact via environment.bat on Windows systems. Test to ensure it is both
+# present and accurate.
+test_name 'PA-466: Ensure env_windows_installdir fact is present and correct' do
+
+  confine :to, :platform => 'windows'
+
+  agents.each do |agent|
+    step "test for presence/accurance of fact on #{agent}" do
+      platform = agent[:platform]
+      ruby_arch = agent[:ruby_arch] || 'x86' # ruby_arch defaults to x86 if nil
+
+      install_dir = platform =~ /-64$/ && ruby_arch == 'x86' ?
+        "C:\\\\Program Files (x86)\\\\Puppet Labs\\\\Puppet" :
+        "C:\\\\Program Files\\\\Puppet Labs\\\\Puppet"
+
+      on agent, puppet('facts') do
+        assert_match(/"env_windows_installdir": "#{install_dir}"/, stdout, "env_windows_installdir fact did not match expected output")
+      end
+    end
+  end
+end


MosesMendoza:
appveyor test failed on ruby 2.3 with something seemingly unrelated?:
```
bundle exec rspec spec
Failed to load feature test for rack: cannot load such file -- rack
C:/projects/puppet/spec/unit/pops/evaluator/comparison_ops_spec.rb:265: warning: key "a" is duplicated and overwritten on line 265
.......................................................................................................................................................................................................................................................................................................................................................................................................................**......................................................................................................*..................................................................................................................................................................................*.............................................................................................................................................................................................................................................................................*.................................................................*.....................................................................................................................*................................................................................................................................**........................................................................**......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*...
.
.....................................................................................................................................................[0;36mDebug: Failed to load library 'syslog' for feature 'syslog'
.....................................................................................*.........................................................................................................................................................................................................................................................................................................................................*.......................*..*.........*................................................................................................................................................................................................................................................................................................................................................................................................................................................................*......*..*.*...*.......*....*....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................**...................*....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*.................................................................................................................................................................................................................................................................................................................................................*............................******............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*.................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*..........................................................................................................................................................................................................*.*.*.....................*................................................................................................................................*..................................................................**......................................................**..........................................................................................................................................................................................................................................................................................**.......*.......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*.........*......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................********..................................................................................................................**..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*...........................................................................................................**...............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................*.............................................................................................................................................................................................................*...............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................Command exited with code -1073741819
Error pushing artifact: File not found: C:\projects\puppet\spec_order.txt
Command exited with code 2
```
Investigating.

MosesMendoza:
These tests have been exercised and passed on the correct platforms in an ad-hoc run of the puppet-agent pipeline, here: https://jenkins.puppetlabs.com/job/platform_puppet-agent_intn-van-sys_suite-manual-puppet-ad-hoc/68/

Gist of related output, here:
https://gist.github.com/MosesMendoza/99c066b6513aecd9e0a34ce2caa1d433

Iristyle:
I think we want to flag these with `PA-466`?

acceptance/tests/windows/env_windows_installdir_fact.rb
```diff
@@ -0,0 +1,15 @@
...
+# fact via environment.bat on Windows systems. Test to ensure it is both
+# present and accurate.
+test_name 'PA-466: Ensure env_windows_installdir fact is present and correct' do
+
+  confine :to, :platform => 'windows'
+
+  agents.each do |agent|
+    step "test for presence/accurance of fact on #{agent}" do
+      on agent, puppet('facts') do
+        assert_match(/"env_windows_installdir": "C:\\\\Program Files\\\\Puppet Labs\\\\Puppet"/, stdout, "env_windows_installdir fact did not match expected output")
```
Iristyle: Don't forget the Ruby architecture which changes the location

https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/ensure_puppet-agent_paths.rb#L137-L152

acceptance/tests/ensure_version_file.rb
```diff
@@ -0,0 +1,23 @@
...
+# https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md
+
+test_name 'PA-466: Ensure version file is created on agent' do
+
+  skip_test 'requires version file which is created by AIO' if [:gem, :git].include?(@options[:type])
+
+  step "test for existence of version file" do
+    agents.each do |agent|
+      version_file = agent[:platform] =~ /windows/ ?
+        'C:/Program Files/Puppet Labs/Puppet/VERSION' :
```
Iristyle: Don't forget the Ruby architecture which changes the location - https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/ensure_puppet-agent_paths.rb#L137-L152

Iristyle:
Aside from the arch part, I think this looks pretty good...

acceptance/tests/ensure_version_file.rb
MosesMendoza: Ah yes. Thanks for pointing that out. It looks like that is saying if the ruby arch (aka msi arch) and windows arch don't match we install to C:/Program Files (x86). I assume that's for the case one is installing a 32 bit agent MSI on a 64 bit version of Windows?

Iristyle: Yes, we need to look at `Program Files(x86)` when installing a 32-bit agent on 64-bit Windows.  Since Cygwin is running as 64-bit on a 64-bit OS, it can see both `Program Files` and `Program Files (x86)`.

MosesMendoza:
@Iristyle re: tagging - do you mean tagging in the beaker DSL sense https://github.com/puppetlabs/beaker/blob/master/docs/concepts/test_tagging.md ? A quick scan of other tests shows tagging used for risk level?

MosesMendoza:
Update commit subjects to include `PA-472` which now tracks explicitly the addition of tests (as opposed to PA-466 which tracked the regression fix). These aren't in the commit subject line because of the check that is done to ensure commit subjects only start with `(maint)`, `(packaging)`, or `(pup-)`

Iristyle:
To follow up on this one... looks like we've only let a `PA-` ticket sneak into commit history once or twice - but usually things that are packaging related are marked `packaging`.

However, in this case... I think we take the PA ticket for the tests, move it to `PUP`, and use that here?  Does that seem reasonable?

MosesMendoza:
Filed PUP-6619 and linked to PA-466, updated commit messages/subject lines

acceptance/tests/ensure_version_file.rb
```diff
@@ -0,0 +1,30 @@
+require 'puppet/acceptance/temp_file_utils'
+extend Puppet::Acceptance::TempFileUtils
+
+# ensure a version file is created according to the puppet-agent path specification:
+# https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md
+
+test_name 'PA-466: Ensure version file is created on agent' do
```
ferventcoder: Huge 👍

acceptance/tests/windows/env_windows_installdir_fact.rb
```diff
@@ -0,0 +1,22 @@
...
+  confine :to, :platform => 'windows'
+
+  agents.each do |agent|
+    step "test for presence/accurance of fact on #{agent}" do
+      platform = agent[:platform]
+      ruby_arch = agent[:ruby_arch] || 'x86' # ruby_arch defaults to x86 if nil
+
+      install_dir = platform =~ /-64$/ && ruby_arch == 'x86' ?
+        "C:\\\\Program Files (x86)\\\\Puppet Labs\\\\Puppet" :
+        "C:\\\\Program Files\\\\Puppet Labs\\\\Puppet"
```
ferventcoder: This is fine - I don't see any of our testing attempting to override the default installation folder.

acceptance/tests/windows/env_windows_installdir_fact.rb
```diff
@@ -0,0 +1,22 @@
+# (PA-466) The env_windows_installdir fact is set as an environment variable
+# fact via environment.bat on Windows systems. Test to ensure it is both
+# present and accurate.
+test_name 'PA-466: Ensure env_windows_installdir fact is present and correct' do
+
+  confine :to, :platform => 'windows'
+
+  agents.each do |agent|
+    step "test for presence/accurance of fact on #{agent}" do
```
Iristyle: Oops - should have been an `occurence` ;0

PR #5197, opened 2016-08-13
(PUP-6397) mount remove autorequire and fix autobefore

mschuchard:
Removes the mount autorequire on mountpoint file resources to avoid circular dependencies. Future new feature will resolve the File[mountpoint] --> Mount[dir] --> File[permissions] issue that people currently hack around with an exec.

Tightens the regexp on the mount autobefore on nested file resources to ensure only correct file resources are captured. Adds a test to ensure false autobefores are not being captured.


branan:
\o/ Thanks for the effort on this. Looks right to me. This will go out in the next puppet-agent 1.6.z release!

PR #5200, opened 2016-08-15
(PUP-6613) Fix given undef and filtering in relationships

hlindberg:
`:undef` is the gift that keeps on giving...

The fix we added that filters out nil/undef values in create_resources proved to add too much filtering as it blocked the ability to give a parameter an explicit undef when it did not have a default value expression. OTOH, keeping the undef values meant that validation of relationships failed later. 
This PR fixes both problems and adds tests for use cases that were not earlier covered by tests.

### DIFF
--- a/lib/puppet/parser/compiler/catalog_validator/relationship_validator.rb
+++ b/lib/puppet/parser/compiler/catalog_validator/relationship_validator.rb
@@ -30,7 +30,7 @@ class Puppet::Parser::Compiler
         # all other relationships requires the referenced resource to exist when mode is strict
         refs = param.value.is_a?(Array) ? param.value.flatten : [param.value]
         refs.each do |r|
-          next if r.nil?
+          next if r.nil? || r == :undef
           unless catalog.resource(r.to_s)
             msg = "Could not find resource '#{r.to_s}' in parameter '#{param.name.to_s}'"
             if Puppet[:strict] == :error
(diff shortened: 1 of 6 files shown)


lib/puppet/parser/resource.rb
```diff
@@ -187,7 +187,9 @@ def to_hash
       unless value.nil?
         case param.name
         when :before, :subscribe, :notify, :require
-          value = value.flatten if value.is_a?(Array)
+          if value.is_a?(Array)
```
thallgren: What happens if `value == :undef` here?

lib/puppet/parser/compiler/catalog_validator/relationship_validator.rb
```diff
@@ -30,7 +30,7 @@ def validate_relationship(param)
         # all other relationships requires the referenced resource to exist when mode is strict
         refs = param.value.is_a?(Array) ? param.value.flatten : [param.value]
         refs.each do |r|
-          next if r.nil?
+          next if r.nil? || r == :undef
```
thallgren: Where does this `:undef` symbol stem from? I though we had gotten rid of that.

hlindberg: The 3.x (wonky, bat-shit-crazy) API is different depending on where an `undef` is - sometimes it is nil, sometimes it is `:undef`. You can observe the difference by using `a => undef` vs. `a => [undef]`. (This also answers the question above, why value is not :undef when it is not in an array).

hlindberg: IIRC trying to remove the last use of :undef caused downstream problems

hlindberg:
Thomas gave a +1 on this. merging

PR #5201, opened 2016-08-15
(PUP-6586) Check auth success codes for Windows logon

MosesMendoza:
The password getter in the Windows user provider issues a logon attempt to verify the password and determine if the value needs synchronized. Prior to this commit, any failure to logon would result in a false return from Puppet::Util::Windows::User.password_is? which would trigger a password reset. As (somewhat) documented in
https://msdn.microsoft.com/en-us/library/windows/desktop/ms681385(v=vs.85).aspx, there are several error codes that indicate logon failure for a different reason than incorrect username/password. After some investigation, it was determined that these codes could be used to indicate a successful authentication even if logon failed. This commit modifies the method to explicitly check for these error codes and return 'true' if encountered, to prevent unnecessary password resets.

### DIFF
--- a/acceptance/tests/resource/user/should_not_modify_disabled.rb
+++ b/acceptance/tests/resource/user/should_not_modify_disabled.rb
@@ -1,0 +1,37 @@
+test_name 'PUP-6586 Ensure puppet does not continually reset password for disabled user' do
+
+  confine :to, :platform => 'windows'
+
+  name = "pl#{rand(99999).to_i}"
+
+  teardown do
+    agents.each do |agent|
+      on(agent, puppet_resource('user', "#{name}", 'ensure=absent'))
+    end
+  end
+
+  manifest = <<-MANIFEST
+user {'#{name}':
+  ensure    => present,
+  password  => 'P@ssword!',
+}
+MANIFEST
+
+  agents.each do |agent|
+    step "create user #{name} with puppet" do
+      apply_manifest_on(agent, manifest, :catch_failures => true)
+    end
+
+    step "disable user #{name}" do
+      on(agent, "net user #{name} /ACTIVE:NO", :acceptable_exit_codes => 0)
+    end
+
+    step "test that password is not reset by puppet" do
+      # :catch_changes will fail the test if there were changes during
+      # run execution
+      apply_manifest_on(agent, manifest, :catch_changes => true)
+    end
+  end
+end
+
+
--- a/lib/puppet/util/windows/user.rb
+++ b/lib/puppet/util/windows/user.rb
@@ -60,15 +60,6 @@ module Puppet::Util::Windows::User
   module_function :check_token_membership
 
   def password_is?(name, password)
-    # Windows API LogonUserW always disallows this with error 1327:
-    # Logon failure: user account restriction. Possible reasons are blank passwords
-    # not allowed, logon hour restrictions, or a policy restriction has been enforced.
-    if password.nil? || password == ''
-      raise Puppet::Util::Windows::Error.new(
-        "Windows does not have the ability to verify NULL or empty passwords"
-      )
-    end
-
     begin
       logon_user(name, password) { |token| }
     rescue Puppet::Util::Windows::Error => detail
@@ -92,7 +83,7 @@ module Puppet::Util::Windows::User
     token = nil
     begin
       FFI::MemoryPointer.new(:handle, 1) do |token_pointer|
-        if LogonUserW(wide_string(name), wide_string('.'), wide_string(password),
+        if LogonUserW(wide_string(name), wide_string('.'), password.nil? ? FFI::Pointer::NULL : wide_string(password),
             fLOGON32_LOGON_NETWORK, fLOGON32_PROVIDER_DEFAULT, token_pointer) == FFI::WIN32_FALSE
           raise Puppet::Util::Windows::Error.new("Failed to logon user #{name.inspect}")
         end
--- a/lib/puppet/util/windows/user.rb
+++ b/lib/puppet/util/windows/user.rb
@@ -22,6 +22,14 @@ module Puppet::Util::Windows::User
   # https://msdn.microsoft.com/en-us/library/windows/desktop/ee207397(v=vs.85).aspx
   SECURITY_MAX_SID_SIZE = 68
 
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/ms681385(v=vs.85).aspx
+  # These error codes indicate successful authentication but failure to
+  # logon for a separate reason
+  ERROR_ACCOUNT_RESTRICTION = 1327
+  ERROR_INVALID_LOGON_HOURS = 1328
+  ERROR_INVALID_WORKSTATION = 1329
+  ERROR_ACCOUNT_DISABLED    = 1331
+
   def check_token_membership
     is_admin = false
     FFI::MemoryPointer.new(:byte, SECURITY_MAX_SID_SIZE) do |sid_pointer|
@@ -63,8 +71,16 @@ module Puppet::Util::Windows::User
 
     begin
       logon_user(name, password) { |token| }
-    rescue Puppet::Util::Windows::Error
-      return false
+    rescue Puppet::Util::Windows::Error => detail
+
+      authenticated_error_codes = Set[
+        ERROR_ACCOUNT_RESTRICTION,
+        ERROR_INVALID_LOGON_HOURS,
+        ERROR_INVALID_WORKSTATION,
+        ERROR_ACCOUNT_DISABLED,
+      ]
+
+      return authenticated_error_codes.include?(detail.code)
     end
   end
   module_function :password_is?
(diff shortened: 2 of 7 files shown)


spec/integration/util/windows/user_spec.rb
```diff
@@ -117,6 +117,28 @@ def expect_logon_failure_error(&block)
       it "should return false given a nil username and an incorrect password" do
         expect(Puppet::Util::Windows::User.password_is?(nil, bad_password)).to be_falsey
       end
+
+      context "with a correct password" do
+        it "should return true even if account restrictions are in place " do
+          Puppet::Util::Windows::User.stubs(:logon_user).raises(Puppet::Util::Windows::Error.new('', 1327))
```
MosesMendoza: I'm not sure if these are the most robust tests...strictly speaking, they test the contract, but if those codes change meaning (however unlikely) we may not catch it?

lib/puppet/util/windows/user.rb
```diff
@@ -22,6 +22,14 @@ def admin?
   # https://msdn.microsoft.com/en-us/library/windows/desktop/ee207397(v=vs.85).aspx
   SECURITY_MAX_SID_SIZE = 68
 
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/ms681385(v=vs.85).aspx
+  # These error codes indicate successful authentication but failure to
+  # logon for a separate reason
+  ERROR_ACCOUNT_RESTRICTION = 1327
+  ERROR_INVALID_LOGON_HOURS = 1328
+  ERROR_INVALID_WORKSTATION = 1329
+  ERROR_ACCOUNT_DISABLED    = 1331
```
MosesMendoza: There was another code I considered, `ERROR_PASSWORD_EXPIRED = 1330` but since the password setter also sets the password to not expire I figured there wasn't much value in it? I suppose it could avoid a password set on the very first run.

ferventcoder: It's fine. Once the password gets reset, it won't expire under the resource.

lib/puppet/util/windows/user.rb
```diff
@@ -63,8 +71,16 @@ def password_is?(name, password)
...
+    rescue Puppet::Util::Windows::Error => detail
+
+      authenticated_error_codes = Set[
+        ERROR_ACCOUNT_RESTRICTION,
+        ERROR_INVALID_LOGON_HOURS,
+        ERROR_INVALID_WORKSTATION,
+        ERROR_ACCOUNT_DISABLED,
+      ]
+
+      return authenticated_error_codes.include?(detail.code)
```
ferventcoder: Is it possible to pop on invalid logon hours/workstation when using an improper password?

lib/puppet/util/windows/user.rb
MosesMendoza: 👍

lib/puppet/util/windows/user.rb
MosesMendoza: In my testing all of these came back with `1326` when triggered but with an invalid password.

spec/integration/util/windows/user_spec.rb
Iristyle: Under the circumstances I think these are probably OK

We might want to add one acceptance test around a disabled account (but not the other restrictions)

lib/puppet/util/windows/user.rb
Iristyle: I was hoping to find a definitive MSDN page with all the possible return values for `LogonUser` rather than just the exhaustive error code list.  But it looks like that page doesn't exist.  Hmmph.

spec/integration/util/windows/user_spec.rb
Iristyle: Do we want to add any negative tests here for the other error codes?

```
ERROR_PASSWORD_MUST_CHANGE = 1907
ERROR_LOGON_FAILURE = 1326
ERROR_NO_LOGON_SERVERS = 1311
ERROR_ACCOUNT_LOCKED_OUT = 1909
ERROR_ACCOUNT_EXPIRED = 1793
ERROR_PASSWORD_EXPIRED = 1330
```

`1909` in particular is an interesting one... because from what I've read, that one can be raised even when the creds are valid.

ferventcoder: Hmm, I'm not completely sure about 1907 or 1330, because both may be valid, but the user/password combination should sync instead of returning true in those cases. So letting them be false to me seems like the right path forward. Or what we are saying for negative tests, that's cool as well. 👍

MosesMendoza: A couple checks reveals...
- 1909 is indeed raised if account is locked out even when creds are valid
- in both the case that account is locked or account is expired, puppet will proceed to set the password anyway after failed logon.

Both of these (and probably others) are cases we could improve on...
I think to avoid further scope creep on this ticket perhaps its best to limit functionality to what's here, and target further improvements for https://tickets.puppetlabs.com/browse/PUP-6569 ?

Iristyle: Agreed for targeting those changes at the other ticket.

Iristyle:
@MosesMendoza I think we also want to alter the docs that we changed before, to remove the note about Puppet being unable to validate this - that should be in `type/user.rb` or similar.  I would double check prior commit on this thread to validate what we previously touched.

MosesMendoza:
@Iristyle makes sense, thanks for the reminder. I had intended to do that as part of this and forgot - will review/update.

MosesMendoza:
@Iristyle pushed new commit  f45da40d which revises previous iterations warning/documenting/commenting on inability to manage blank passwords. Given recent determination that we actually can, it seemed justified to reverse course there?

lib/puppet/util/windows/user.rb
```diff
@@ -52,19 +60,18 @@ def check_token_membership
   module_function :check_token_membership
 
   def password_is?(name, password)
-    # Windows API LogonUserW always disallows this with error 1327:
-    # Logon failure: user account restriction. Possible reasons are blank passwords
-    # not allowed, logon hour restrictions, or a policy restriction has been enforced.
-    if password.nil? || password == ''
-      raise Puppet::Util::Windows::Error.new(
```
MosesMendoza: Per commit msg in f45da40d I found we can successfully supply both `nil` and `''` (zero-length) passwords -  strictly speaking, from an external interface perspective, I think all that's needed is to remove the second operand to `||` (check for `''`) but I also didn't still see a reason to `raise` on `nil`. That being said, `LogonUserW` doesn't appear to distinguish between a password of value `NULL` and of value `''`, so perhaps that would be a reason to `raise` on `nil` - prevent ambiguity? Not sure..

MosesMendoza:
I still want to add an acceptance test to this w/a disabled user.

ferventcoder:
@MosesMendoza Travis CI really hit up a nice check on the commit message - can you fix that up?

MosesMendoza:
d'oh! I was going to squash that into the first commit anyway. Updated

MosesMendoza:
output from a failure of the new acceptance test against a puppet version without these changes: https://gist.github.com/MosesMendoza/2c65261bf175ea0825f87be1a946eb56

MosesMendoza:
output from a successful run of the new test against this code: https://gist.github.com/MosesMendoza/aba16aae0d712f3bf8d7391fb7cd9878 (from https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-puppet-ad-hoc/75/SLAVE_LABEL=beaker,TEST_TARGET=windows2012r2-64a/consoleFull)

lib/puppet/util/windows/user.rb
Iristyle: After looking at this a bit more closely, a `nil` value may be passed successfully for a user with a blank password... however, for it to work properly, it also requires changing how the pointer is passed to FFI:

```ruby
password.nil? ? FFI::Pointer::NULL : wide_string(password)
```

Iristyle:
Also, in the commit message I don't think you want to refer to commit https://github.com/MosesMendoza/puppet/commit/fa2b3ea67feca21de27ea14e325e037f7a2af260 (which is in your branch)... instead refer to https://github.com/puppetlabs/puppet/commit/70426ea1effce8aaa545db911c3f340c09c25f56

MosesMendoza:
@Iristyle yes - thanks

spec/integration/util/windows/user_spec.rb
MosesMendoza: Additional note left in PUP-6569

acceptance/tests/resource/user/should_not_modify_disabled.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    step "disable user #{name}" do
+      on(agent, "net user #{name} /ACTIVE:NO", :acceptable_exit_codes => 0)
+    end
+
+    step "test that password is not reset by puppet" do
+      # :catch_changes will fail the test if there were changes during
+      # run execution
+      apply_manifest_on(agent, manifest, :catch_changes => true)
```
Iristyle: 👍 - seems better

PR #5020, opened 2016-06-15
(PUP-6141) Move from forgeapi.puppetlabs.com to forgeapi.puppet.com

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

PR #5202, opened 2016-08-15
Revert "(PUP-6141) Move from forgeapi.puppetlabs.com to forgeapi.puppet.com"

johnduarte:
Reverts puppetlabs/puppet#5020

The certificate on our test forge instance is not properly honoring the 'puppet.com' domain and is causing the PMT test to fail. Our OPS team is investigating currently. PR #5020 should be re-merged as soon as the certificate issue has been addressed. Our apologies for the delay in getting this rolled out.


er0ck:
👍

PR #5203, opened 2016-08-16
(PUP-6621) Log corrective_change exceptions at info instead of error level

branan:
Previously, when a custom type was incompatible with corrective
change, it would log any exception caught at the error level. This
causes Puppet to fail the transaction.

Instead, we should log at a lower level, so that users can look into
making their types compatible without causing catalog applications to
fails in the meantime.

### DIFF
--- a/lib/puppet/property.rb
+++ b/lib/puppet/property.rb
@@ -387,7 +387,7 @@ class Puppet::Property < Puppet::Parameter
     rescue => detail
       # Certain operations may fail, but we don't want to fail the transaction if we can
       # avoid it
-      Puppet.log_exception(detail, "Unknown failure comparing values #{should} and #{is} using insync? on type: #{self.resource.ref} property: #{self.name}")
+      Puppet.log_exception(detail, "Unknown failure comparing values #{should} and #{is} using insync? on type: #{self.resource.ref} property: #{self.name}", :default, { :level => :info })
 
       # Return nil, ie. unknown
       nil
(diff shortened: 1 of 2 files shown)


kbarber:
Warning instead? I don't mind dropping it from an error though.

branan:
@kbarber The problem is that it's not necessarily in the user's control - 3rd party modules are a pretty likely source of custom types. Spitting out a bunch of warnings at users when it's not their fault isn't super cool. I'd rather log at a lower level, and be able to give them instructions "If corrective change isn't being reported for certain resources, you can find out what error your type is having by looking at the logs"

branan:
Specifically, a ton of stuff in our Cisco integrations isn't playing nice (some horrible munge/validate things that I'm going to try to fix). Having every user of that end up with lots of warnings in their logs and reports sounds nasty to me :(

kbarber:
Well, info will show up in their logs and reports anyway. So you're concern isn't the volume of data, its the level and I guess the CLI log color that is chosen. I'm okay with whatever tbh. To debug, it will require --trace at the end of the day, so as long as that works the same I think we can deal with whatever choices we make.

branan:
@kbarber Info level avoids upsetting the sorts of users that have their cron jobs set to email them any time a job prints "warning". It happens, and they tend to be grumpy when we introduce new warnings that they can't silence :)

branan:
I'd rather it still be printed and in logs for visibility on manual expansion, just trying to be at a level that avoids automated tooling waking people up at night

kbarber:
As an aside I'm +1 on allowing log_exception to accept different levels, I want to use this in app_modeling (so outside of puppet core), since sometimes our tests throw exceptions in third party code that is impossible to debug if we just grab the exception message.

PR #5204, opened 2016-08-16
(PUP-6623) Define an unwrap function

adrienthebo:
This commit adds an unwrap function for sensitive objects; this serves
to expose the plaintext value in case the wrapped value needs to be
operated on/interpolated/etc. While this does remove the safety measure
that the Sensitive object provides it also allows more general use of
the data type.

### DIFF
--- a/lib/puppet/functions/unwrap.rb
+++ b/lib/puppet/functions/unwrap.rb
@@ -1,0 +1,40 @@
+# Unwraps a Sensitive value and returns the wrapped object.
+#
+# ~~~puppet
+# $plaintext = 'hunter2'
+# $pw = Sensitive.new($plaintext)
+# notice("Wrapped object is $pw") #=> Prints "Wrapped object is Sensitive [value redacted]"
+# $unwrapped = $pw.unwrap
+# notice("Unwrapped object is $unwrapped") #=> Prints "Unwrapped object is hunter2"
+# ~~~
+#
+# You can optionally pass a block to unwrap in order to limit the scope where the
+# unwrapped value is visible.
+#
+# ~~~puppet
+# $pw = Sensitive.new('hunter2')
+# notice("Wrapped object is $pw") #=> Prints "Wrapped object is Sensitive [value redacted]"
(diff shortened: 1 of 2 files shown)


lib/puppet/functions/unwrap.rb
```diff
@@ -0,0 +1,39 @@
...
+Puppet::Functions.create_function(:unwrap) do
+  dispatch :unwrap do
+    param 'Sensitive', :arg
+    optional_block_param
+  end
+
+  def unwrap(arg)
+    unwrapped = arg.unwrap
+    if block_given?
+      yield(unwrapped)
```
hlindberg: Seems better if it returns the result of the block instead of the unwrapped value (on line 37). This both to reduce the risk of leaking the value, and to make it more convenient to compute a derived value that is returned.

hlindberg:
Looks good apart from not returning the result from the optionally given block.

lib/puppet/functions/unwrap.rb
adrienthebo: I considered that, but I think I like the consistency of just returning the same value regardless of if a block is passed. However, scanf returns the result of the block and not the passed value as well. Is this a common convention for functions accepting an optional block param?

branan: FWIW, returning the result of the block seems the more intuitive behavior to me. Being able to pass in a block that keeps the unwrapped value "contained" so I can do a simple computation on it and then return the result of that to the outside world, without having to bring out the unwrapped value, feels both safer from the perspective of sensitive data handling, and also cleaner from a functional-ish perspective

adrienthebo: Changed.

adrienthebo:
Returning the result of the block also allows users to re-wrap sensitive information which is pretty neat; I've changed the function accordingly.

branan:
I'm :+1: on this once it rolls through Travis/Appveyor

adrienthebo:
@hlindberg thanks!

PR #5205, opened 2016-08-17
(PUP-6622) Fix two 4.6 performance regressions

thallgren:
This PR contains two performance improvements to remedy regressions that were introduced in 4.6.0. It makes the "function_loading" performance benchmark run more than twice as fast (but not quite as fast as in 4.5.x). The main reason that performance is not on par is that in 4.6.0 we introduce a Puppet call stack. It is expected to slow function calls slightly.

### DIFF
--- a/function_loading.samples
+++ b/function_loading.samples
@@ -1,0 +1,11 @@
+timestamp,elapsed,responsecode,success,name
+1471457716000,7100,200,true,function_loading
+1471457723000,6603,200,true,function_loading
+1471457730000,6420,200,true,function_loading
+1471457736000,6727,200,true,function_loading
+1471457743000,6624,200,true,function_loading
+1471457750000,6834,200,true,function_loading
+1471457757000,6787,200,true,function_loading
+1471457763000,6812,200,true,function_loading
+1471457770000,6431,200,true,function_loading
+1471457777000,7370,200,true,function_loading
--- a/lib/puppet/pops/evaluator/runtime3_support.rb
+++ b/lib/puppet/pops/evaluator/runtime3_support.rb
@@ -267,7 +267,7 @@ module Runtime3Support
 
   def call_function(name, args, o, scope, &block)
     file, line = extract_file_line(o)
-    loader = Adapters::LoaderAdapter.loader_for_model_object(o, scope)
+    loader = Adapters::LoaderAdapter.loader_for_model_object(o, scope, file)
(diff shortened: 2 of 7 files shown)


function_loading.samples
```diff
@@ -0,0 +1,11 @@
...
+1471448363000,15208,200,true,function_loading
+1471448378000,14873,200,true,function_loading
+1471448393000,15732,200,true,function_loading
+1471448409000,15531,200,true,function_loading
+1471448425000,15913,200,true,function_loading
+1471448440000,16297,200,true,function_loading
+1471448457000,16862,200,true,function_loading
+1471448474000,17126,200,true,function_loading
+1471448491000,17219,200,true,function_loading
+1471448508000,17062,200,true,function_loading
```
hlindberg: this file should not be checked in

lib/puppet/pops/evaluator/runtime3_support.rb
```diff
@@ -267,11 +267,11 @@ def external_call_function(name, args, scope, &block)
 
   def call_function(name, args, o, scope, &block)
     file, line = extract_file_line(o)
-    loader = Adapters::LoaderAdapter.loader_for_model_object(o, scope)
+    loader = Adapters::LoaderAdapter.loader_for_model_object(o, scope, file)
     if loader && func = loader.load(:function, name)
       Puppet::Util::Profiler.profile(name, [:functions, name]) do
         # Add stack frame when calling
-        return Puppet::Pops::PuppetStack.stack(file, line, func, :call, [scope, *args], &block)
+        return Kernel.eval('func.call(scope, *args, &block)', Kernel.binding, file || '', line)
```
hlindberg: hm, this is not ideal as it is much harder to understand when taken out of the context of Puppet::Stack. Understand why this moved here. Suggest a comment stating `# See Puppet::Pops::PuppetStack` as the relationship is otherwise lost to the reader.

lib/puppet/pops/parser/lexer2.rb
```diff
@@ -500,6 +500,7 @@ def initialize()
         else
           @scanner.pos += 1
         end
+        ctx[:current_line_start] = @scanner.pos
```
hlindberg: Looks like this will be wrong after a newline_jump (line 498) as that position is not at the beginning of the line IIRC.

hlindberg:
good to merge when tests are green

PR #5206, opened 2016-08-17
(PUP-6619) Correct acceptance test for env_windows_installdir fact

MosesMendoza:
Prior to this commit, the env_windows_installdir_fact.rb test would fail because after all escaping was done, beaker would match a string literal path with two back slashes against a string literal path with four back slashes. The output of Beaker's assert match will have escaped backslashes '\' in facts output, so the output we match against looks like `C:\\\\Program Files...`.

This commit modifies the test to parse the fact output as JSON and perform a string comparison expecting the known correct value. No escaping necessary, so our test is more readable as well.

### DIFF
--- a/acceptance/tests/windows/env_windows_installdir_fact.rb
+++ b/acceptance/tests/windows/env_windows_installdir_fact.rb
@@ -5,17 +5,21 @@ test_name 'PA-466: Ensure env_windows_installdir fact is present and correct' do
 
   confine :to, :platform => 'windows'
 
+  require 'json'
+
   agents.each do |agent|
-    step "test for presence/accurance of fact on #{agent}" do
+    step "test for presence/accuracy of fact on #{agent}" do
       platform = agent[:platform]
       ruby_arch = agent[:ruby_arch] || 'x86' # ruby_arch defaults to x86 if nil
 
       install_dir = platform =~ /-64$/ && ruby_arch == 'x86' ?
-        "C:\\\\Program Files (x86)\\\\Puppet Labs\\\\Puppet" :
-        "C:\\\\Program Files\\\\Puppet Labs\\\\Puppet"
+        "C:\\Program Files (x86)\\Puppet Labs\\Puppet" :
+        "C:\\Program Files\\Puppet Labs\\Puppet"
 
-      on agent, puppet('facts') do
-        assert_match(/"env_windows_installdir": "#{install_dir}"/, stdout, "env_windows_installdir fact did not match expected output")
+      on agent, puppet('facts', '--render-as json') do |result|
+        facts = JSON.parse(result.stdout)
+        actual_value = facts["values"]["env_windows_installdir"]
+        assert_equal(install_dir, actual_value, "env_windows_installdir fact did not match expected output")
       end
     end
   end


MosesMendoza:
example of failed test prior to this PR: https://gist.github.com/MosesMendoza/fb227df6b279f94bff6b4cc7fae4e6f1

MosesMendoza:
tested with `be rake ci:test:aio TEST_TARGET='windows2012r2-64a' SHA='1.6.0' TESTS='tests/windows/env_windows_installdir_fact.rb' SERVER_VERSION='2.5.0'`, resulting in pass: https://gist.github.com/MosesMendoza/9e15c3a98c468fff4e77e820840286b2

acceptance/tests/windows/env_windows_installdir_fact.rb
```diff
@@ -6,13 +6,20 @@
...
+      # The output of Beaker's assert match will have escaped backslashes '\'
+      # in facts output, so the output we match against looks like
+      # "C:\\\\Program Files...". However, we can't just supply four slashes
+      # in our string argument to assert_match, because they're escaped
+      # resulting in asserting a match for "C:\\Program Files', which fails.
+      # So we double escape '\\' (eight), resulting in asserting a match for
+      # "C:\\\\Program Files...".
       install_dir = platform =~ /-64$/ && ruby_arch == 'x86' ?
-        "C:\\\\Program Files (x86)\\\\Puppet Labs\\\\Puppet" :
-        "C:\\\\Program Files\\\\Puppet Labs\\\\Puppet"
```
Iristyle: So the value coming back from `puppet facts` is `console` format (which could be explicitly specified with `--render-as console` - but it's also just a pretty printed JSON.

Maybe it's better to just do a `JSON.parse` on the string that comes back, and lookup `env_windows_installdir` that way?

acceptance/tests/windows/env_windows_installdir_fact.rb
```diff
@@ -6,13 +6,20 @@
...
+        "C:\\\\\\\\Program Files (x86)\\\\\\\\Puppet Labs\\\\\\\\Puppet" :
+        "C:\\\\\\\\Program Files\\\\\\\\Puppet Labs\\\\\\\\Puppet"
```
ferventcoder: wow... perhaps just test the value contains program files?

Iristyle: I think part of the escaping is related to it being JSON.  If we let the JSON parser do it's thing, then this should make it a little cleaner.

acceptance/tests/windows/env_windows_installdir_fact.rb
MosesMendoza: 👍  will update

acceptance/tests/windows/env_windows_installdir_fact.rb
ferventcoder: Much better.

acceptance/tests/windows/env_windows_installdir_fact.rb
```diff
@@ -5,17 +5,21 @@
...
       install_dir = platform =~ /-64$/ && ruby_arch == 'x86' ?
-        "C:\\\\Program Files (x86)\\\\Puppet Labs\\\\Puppet" :
-        "C:\\\\Program Files\\\\Puppet Labs\\\\Puppet"
+        "C:\\Program Files (x86)\\Puppet Labs\\Puppet" :
+        "C:\\Program Files\\Puppet Labs\\Puppet"
 
-      on agent, puppet('facts') do
-        assert_match(/"env_windows_installdir": "#{install_dir}"/, stdout, "env_windows_installdir fact did not match expected output")
+      on agent, puppet('facts', '--render-as json') do
+        facts = JSON.parse(stdout)
```
Iristyle: 👍

MosesMendoza:
This was tested to pass using the same beaker invocation as the last test.

acceptance/tests/windows/env_windows_installdir_fact.rb
Iristyle: Oh wait.. actually... where is `stdout` coming from?

Shouldn't there be a block there like this:

```
on agent, puppet('facts', '--render-as json') do |stdout|
```

Iristyle: Actually, I think I've more frequently seen `result.stdout`, when given:

```ruby
on agent, puppet('facts', '--render-as json') do |result|
```

I guess `stdout` is magically available even without anything passed to the block per - https://github.com/puppetlabs/beaker/blob/ea263617bcb3e4d4a27c65c48bd55c56c8064612/lib/beaker/dsl/helpers/host_helpers.rb#L85-L94

But it also looks like `stdout` is deprecated - https://github.com/puppetlabs/beaker/blob/ea263617bcb3e4d4a27c65c48bd55c56c8064612/lib/beaker/dsl/helpers/host_helpers.rb#L128-L136

MosesMendoza: updated

acceptance/tests/windows/env_windows_installdir_fact.rb
```diff
@@ -5,17 +5,21 @@
...
-        "C:\\\\Program Files\\\\Puppet Labs\\\\Puppet"
+        "C:\\Program Files (x86)\\Puppet Labs\\Puppet" :
+        "C:\\Program Files\\Puppet Labs\\Puppet"
 
-      on agent, puppet('facts') do
-        assert_match(/"env_windows_installdir": "#{install_dir}"/, stdout, "env_windows_installdir fact did not match expected output")
+      on agent, puppet('facts', '--render-as json') do |result|
+        facts = JSON.parse(result.stdout)
+        actual_value = facts["values"]["env_windows_installdir"]
+        assert_equal(install_dir, actual_value, "env_windows_installdir fact did not match expected output")
```
ferventcoder: Looks good to me. @Iristyle ?

Iristyle: Yar

MosesMendoza:
confirmed test passes with new explicit block parameter: https://gist.github.com/MosesMendoza/87d1d02858b7709931bf2433e79e43d2

PR #5208, opened 2016-08-17
(PUP-6628) Access @should directly when calculating corrective change

branan:
Currently, the corrective change logic uses the normal `should` and
`should=` accessors when messing with property values. This can cause
failures in property `validate` and `should=` if `unmunge` does not
put the data into a state that `munge` or `validate` is happy with.

By using a small amount of metaprogramming magic, we can access the
guts of the property value directly, bypassing those layers of logic
and creating a cleaner "transplant" of the value.

### DIFF
--- a/lib/puppet/property.rb
+++ b/lib/puppet/property.rb
@@ -379,10 +379,10 @@ class Puppet::Property < Puppet::Parameter
     # this kind of arbitrary comparisons into the API to remove this complexity. -ken
 
     # Backup old should, set it to the new value, then call insync? on the property.
-    old_should = self.should
+    old_should = @should
 
     begin
-      self.should = should
+      @should = should
       insync?(is)
     rescue => detail
       # Certain operations may fail, but we don't want to fail the transaction if we can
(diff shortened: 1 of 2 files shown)


kbarber:
Obviously it goes without stating that the instance_variable tapping feels horrid, but the change itself in concept is understandable. This will generate some cc events the first time the user is upgraded I think, because the old values pre-upgrade won't always be the same.

branan:
Users with types that have bad or missing `unmunge` functions might actually need to blow away the persistence layer entirely, unfortunately. We have no way of knowing if the values in there (that get propagated every run if there is no change event) are good or bad :(

kbarber:
Hmm, good point. We should probably make sure this advice gets transferred into the release notes on this one.

branan:
Yeah. Fortunately it should only affect users for whom things failed in p-a 1.6.0, so that's easy enough to write a note on

kbarber:
@branan will this make davis in time?

branan:
@kbarber that's the plan.

PR #5209, opened 2016-08-17
(PUP-6627) Redact file diffs when content is sensitive

adrienthebo:


branan:
:+1:

spec/unit/type/file/content_spec.rb
```diff
@@ -184,25 +184,30 @@
           expect(content.respond_to?("diff")).to eq(false)
         end
 
-        [true, false].product([true, false]).each do |cfg, param|
```
branan: It looks like you lost some negative tests on the overall Puppet[:show_diff] / Resource[show_diff] matrix here?

branan: This is resolved with new tests below

PR #5211, opened 2016-08-17
(PUP-6604) acceptance: compilation should get new variable values each time

er0ck:
[skip ci]

### DIFF
--- a/acceptance/tests/environment/variables_refreshed_each_compilation.rb
+++ b/acceptance/tests/environment/variables_refreshed_each_compilation.rb
@@ -1,0 +1,68 @@
+test_name 'C98115 compilation should get new values in variables on each compilation' do
+require 'puppet/acceptance/environment_utils'
+extend Puppet::Acceptance::EnvironmentUtils
+
+  app_type               = File.basename(__FILE__, '.*')
+  tmp_environment        = mk_tmp_environment_with_teardown(master, app_type)
+  fq_tmp_environmentpath = "#{environmentpath}/#{tmp_environment}"
+
+  create_remote_file(master, "#{fq_tmp_environmentpath}/environment.conf", <<-CONF)
+    environment_timeout = unlimited
+  CONF
+  # the module function loading logic is different from inside a single manifest
+  #   we exercise both here
+  on(master, "mkdir -p #{fq_tmp_environmentpath}/modules/uptime/{manifests,functions}")
+  create_remote_file(master, "#{fq_tmp_environmentpath}/modules/uptime/manifests/init.pp", <<-FILE)
+    class uptime {
+      notify { 'current uptime':
+        message => uptime::my_system_uptime()
+      }
+    }
+  FILE
+  create_remote_file(master, "#{fq_tmp_environmentpath}/modules/uptime/functions/my_system_uptime.pp", <<-FILE)
+    function uptime::my_system_uptime() {
+      $facts['system_uptime']
+    }
+  FILE
+  create_sitepp(master, tmp_environment, <<-SITE)
+    function bar() {
+      $facts['system_uptime']['seconds']
+    }
+    class foo::bar {
+      notify { "local_${bar()}_local": }
+    }
+    include foo::bar
+    include uptime
+  SITE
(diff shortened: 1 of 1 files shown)


acceptance/tests/environment/variables_refreshed_each_compilation.rb
```diff
@@ -0,0 +1,65 @@
...
+  on(master, "chmod -R 0777 #{fq_tmp_environmentpath}/")
+
+  step "run agent in #{tmp_environment}, ensure it increments the uptime with each run" do
+    with_puppet_running_on(master,{}) do
+      uptime = nil
+      module_uptime = nil
+      agents.each do |agent|
+        on(agent, puppet("agent -t --server #{master.hostname} --environment #{tmp_environment}"),
+           :accept_all_exit_codes => true) do |result|
+          #assert_match(/Notice: \d+/, result.stdout, 'first uptime was not as expected')
```
johnduarte: Can we remove the unused assertions?

johnduarte:
Using `puppet-agent` and `puppet-server` packages containing `puppet` and SHA 4e7bbb0, and uncommenting the in-manifest function and assertions; this test passes on Debian 8 and Windows 2012r2.

acceptance/tests/environment/variables_refreshed_each_compilation.rb
```diff
@@ -0,0 +1,66 @@
...
+  on(master, "chmod -R 0777 #{fq_tmp_environmentpath}/")
+
+  step "run agent in #{tmp_environment}, ensure it increments the uptime with each run" do
+    with_puppet_running_on(master,{}) do
+      uptime = nil
+      module_uptime = nil
+      local_uptime_pattern = 'local_(\d+)_local'
+      agents.each do |agent|
+        on(agent, puppet("agent -t --server #{master.hostname} --environment #{tmp_environment}"),
+           :accept_all_exit_codes => true) do |result|
```
johnduarte: Nit. Accepting all exit codes does not comply with the Beaker style guide to accept the minimum expected exit codes:
https://github.com/puppetlabs/beaker/blob/master/docs/concepts/style_guide.md#acceptable-exit-codes

acceptance/tests/environment/variables_refreshed_each_compilation.rb
```diff
@@ -0,0 +1,66 @@
...
+          assert_match(/Notice: #{local_uptime_pattern}/, result.stdout, 'first uptime was not as expected')
+          assert_match(/"seconds"=>\d+,/, result.stdout, 'first module uptime was not as expected')
+          uptime = result.stdout.match(/Notice: #{local_uptime_pattern}/)[0]
+          module_uptime = result.stdout.match(/"seconds"=>(\d+),/)[0]
+        end
+        sleep 1
+        on(agent, puppet("agent -t --server #{master.hostname} --environment #{tmp_environment}"),
+           :accept_all_exit_codes => true) do |result|
```
johnduarte: Nit. Accepting all exit codes does not comply with the Beaker style guide to accept the minimum expected exit codes:
https://github.com/puppetlabs/beaker/blob/master/docs/concepts/style_guide.md#acceptable-exit-codes

er0ck: yes, except this line:
`In the last case, when any exit code is allowed, one must follow-up with a valid assertion test.`
which i believe i've fulfilled.  
actually i really really think we should amend the style guide to not allow errors when failures can be achieved.  such is the case with `#on`.  actually i'd love beaker 3 to fail on unexpected error codes instead of erroring, but i haven't thought through all the repercussions of this.

johnduarte: +1 on having beaker fail based on unexpected error codes.

The asserts below do not check for desired exit codes. So, the agent run could have errors in it and pass so long as the uptime regexes are satisfied.

er0ck: i suppose that's true, however unlikely.  i also don't know if we should care in the context of this test.
but it's easy and safe to add, so i probably should.

PR #5212, opened 2016-08-17
DO NOT MERGE! (maint) Add unit test for multiple loads of non-autoloaded function

thallgren:
Adds a test that loads a function twice using the same environment
but different compiles.


hlindberg:
The relevant parts of this PR have already been merged.

PR #5213, opened 2016-08-17
(PUP-6642) Provide a way to set transaction_uuid for an agent run

MikaelSmith:
Without this, a report can be passed to an agent run, but the
transaction_uuid used in that report can't match the transaction_uuid
used for node requests and sending the report. Allow the
transaction_uuid to be passed as an argument to the agent run.

### DIFF
--- a/lib/puppet/agent.rb
+++ b/lib/puppet/agent.rb
@@ -39,7 +39,7 @@ class Puppet::Agent
     block_run = Puppet::Application.controlled_run do
       splay client_options.fetch :splay, Puppet[:splay]
       result = run_in_fork(should_fork) do
-        with_client do |client|
+        with_client(client_options[:transaction_uuid]) do |client|
           begin
             client_args = client_options.merge(:pluginsync => Puppet::Configurer.should_pluginsync?)
             lock { client.run(client_args) }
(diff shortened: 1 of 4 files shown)


MikaelSmith:
@joshcooper @branan thoughts on this? It supports [PCP-549](https://tickets.puppetlabs.com/browse/PCP-549), where we want to invoke an Agent run and pass in the report so we can review it after the run has completed.

MikaelSmith:
This enables https://github.com/puppetlabs/pxp-agent/pull/472.

MikaelSmith:
An alternative to this might be to change Puppet::Agent's merge on client_options to 'merge!', which would allow passing in an empty hash and getting the report after the run.

adrienthebo:
Seems sensible to me, but how would you expect this would be used? (API wise that is)

MikaelSmith:
Which part, using `merge!` or passing in transaction_uuid?

With `merge!`, I'd create an empty hash of client_options and pass it to `run`, and expect it to be filled when `:report` is set on the hash. So after `run` returned, I'd use `client_options[:report]` to access the report.

For transaction_uuid, I linked an example already.

MikaelSmith:
Retargeted against master, as Puppet 4.8.0 is probably when this would get released.

MikaelSmith:
ping

PR #5214, opened 2016-08-18
(PUP-6629) Tighten up corruption handling for transactionstorefile

kbarber:
This adds better testing for the corruption case, and stops the system from
hard failing if it has to do a rename.

### DIFF
--- a/lib/puppet/transaction/persistence.rb
+++ b/lib/puppet/transaction/persistence.rb
@@ -47,7 +47,7 @@ class Puppet::Transaction::Persistence
       return
     end
     unless File.file?(filename)
-      Puppet.warning("Persistence file #{filename} is not a file, ignoring")
+      Puppet.warning("Transaction store file #{filename} is not a file, ignoring")
       return
     end
 
@@ -56,18 +56,21 @@ class Puppet::Transaction::Persistence
       begin
         result = Puppet::Util::Yaml.load_file(filename)
       rescue Puppet::Util::Yaml::YamlLoadError => detail
-        Puppet.err "Transaction store file #{filename} is corrupt (#{detail}); replacing"
+        Puppet.log_exception(detail, "Transaction store file #{filename} is corrupt (#{detail}); replacing", { :level => :warning })
 
         begin
           File.rename(filename, filename + ".bad")
-        rescue
+        rescue => detail
+          Puppet.log_exception(detail, "Unable to rename corrupt transaction store file: #{detail}")
           raise Puppet::Error, "Could not rename corrupt transaction store file #{filename}; remove manually", detail.backtrace
         end
+
+        result = {}
       end
     end
 
     unless result.is_a?(Hash)
-      Puppet.err "Transaction state file #{filename} is valid YAML but not returning a hash. Check the file for corruption, or remove it before continuing."
+      Puppet.err "Transaction store file #{filename} is valid YAML but not returning a hash. Check the file for corruption, or remove it before continuing."
       return
     end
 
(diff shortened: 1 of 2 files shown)


lib/puppet/transaction/persistence.rb
```diff
@@ -56,18 +56,21 @@ def load
       begin
         result = Puppet::Util::Yaml.load_file(filename)
       rescue Puppet::Util::Yaml::YamlLoadError => detail
-        Puppet.err "Transaction store file #{filename} is corrupt (#{detail}); replacing"
+        Puppet.log_exception(detail, "Transaction store file #{filename} is corrupt (#{detail}); replacing")
```
branan: Should this be using the new "log exceptions as a warning" machinery that I added?

lib/puppet/transaction/persistence.rb
```diff
@@ -56,18 +56,21 @@ def load
...
           raise Puppet::Error, "Could not rename corrupt transaction store file #{filename}; remove manually", detail.backtrace
         end
+
+        result = {}
       end
     end
 
     unless result.is_a?(Hash)
-      Puppet.err "Transaction state file #{filename} is valid YAML but not returning a hash. Check the file for corruption, or remove it before continuing."
+      Puppet.err "Transaction store file #{filename} is valid YAML but not returning a hash. Check the file for corruption, or remove it before continuing."
```
branan: I assume this should stay an error, because if we hit this we've gone well past the new line 68 above and something's really gone horribly wrong.

lib/puppet/util/yaml.rb
```diff
@@ -9,9 +9,19 @@ module Puppet::Util::Yaml
...
-    yaml = YAML.load_file(filename)
-    yaml || default_value
+  def self.load_file(filename, default_value = false, strip_tags = false)
+    if(strip_tags) then
+      data = YAML::parse_file(filename)
+      data.root.each do |o|
+        o.tag = nil if o.respond_to?(:tag=)
+      end
+      require 'pp'
+      pp data.to_ruby
```
branan: debug print

lib/puppet/util/yaml.rb
```diff
@@ -9,9 +9,19 @@ module Puppet::Util::Yaml
 
   class YamlLoadError < Puppet::Error; end
 
-  def self.load_file(filename, default_value = false)
-    yaml = YAML.load_file(filename)
-    yaml || default_value
+  def self.load_file(filename, default_value = false, strip_tags = false)
```
branan: Are there unit tests for this function that should be expanded?

kbarber: totally

branan:
Wow, a behavior difference in Ruby 1.9. This is just the worst.

I should have some time today to get poke this, but damn :(

lib/puppet/transaction/persistence.rb
kbarber: Yes, and in an oh-shit moment, I found a bug in property.rb with how many args you're passing. I'll fix this case here, and fix the other bug too.

kbarber:
@branan I'm going to break this PR in half because I'm worried about the YAML solution. This PR covers the transactionstore corruption for now only.

lib/puppet/transaction/persistence.rb
kbarber: https://github.com/puppetlabs/puppet/pull/5219

kbarber:
@branan this one at least passes on its own.

PR #5215, opened 2016-08-18
(PUP-6582) loader bound functions cannot bind scope instance

hlindberg:
This refactors Closure into two specific subclasses: Dynamic (for lambdas), and Named (static, bound by loader). This is done to ensure that all Named function closures refer to top-scope in a way that is resolved on demand rather than being bound in a member variable. The dynamic closures must bind to a scope. Such closures may never be bound in something that survives a compilation.

### DIFF
--- a/lib/puppet/pops/evaluator/closure.rb
+++ b/lib/puppet/pops/evaluator/closure.rb
@@ -27,7 +27,7 @@ class Closure < CallableSignature
   # Evaluates a closure in its enclosing scope after having matched given arguments with parameters (from left to right)
   # @api public
   def call(*args)
-    call_with_scope(@enclosing_scope, args)
+    call_with_scope(enclosing_scope, args)
   end
 
   # This method makes a Closure compatible with a Dispatch. This is used when the closure is wrapped in a Function
@@ -35,7 +35,7 @@ class Closure < CallableSignature
   # checks of the argument type/arity validity).
   # @api private
   def invoke(instance, calling_scope, args, &block)
-    @enclosing_scope.with_global_scope do |global_scope|
+    enclosing_scope.with_global_scope do |global_scope|
       call_with_scope(global_scope, args, &block)
     end
   end
@@ -44,7 +44,7 @@ class Closure < CallableSignature
   def call_by_name(args_hash, enforce_parameters)
     if enforce_parameters
       # Push a temporary parameter scope used while resolving the parameter defaults
-      @enclosing_scope.with_parameter_scope(closure_name, parameter_names) do |param_scope|
+      enclosing_scope.with_parameter_scope(closure_name, parameter_names) do |param_scope|
         # Assign all non-nil values, even those that represent non-existent paramaters.
         args_hash.each { |k, v| param_scope[k] = v unless v.nil? }
         parameters.each do |p|
@@ -56,7 +56,7 @@ class Closure < CallableSignature
               # No default. Assign nil if the args_hash included it
               param_scope[name] = nil if args_hash.include?(name)
             else
-              param_scope[name] = param_scope.evaluate(name, p.value, @enclosing_scope, @evaluator)
+              param_scope[name] = param_scope.evaluate(name, p.value, enclosing_scope, @evaluator)
             end
           end
         end
@@ -65,7 +65,7 @@ class Closure < CallableSignature
       Types::TypeMismatchDescriber.validate_parameters(closure_name, params_struct, args_hash)
     end
 
-    @evaluator.evaluate_block_with_bindings(@enclosing_scope, args_hash, @model.body)
+    @evaluator.evaluate_block_with_bindings(enclosing_scope, args_hash, @model.body)
   end
 
   def parameters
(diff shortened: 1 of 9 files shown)


lib/puppet/pops/evaluator/closure.rb
```diff
@@ -56,7 +55,7 @@ def call_by_name(args_hash, enforce_parameters)
               # No default. Assign nil if the args_hash included it
               param_scope[name] = nil if args_hash.include?(name)
             else
-              param_scope[name] = param_scope.evaluate(name, p.value, @enclosing_scope, @evaluator)
+              param_scope[name] = param_scope.evaluate(name, p.value, enclosing_scope, @evaluator)
```
thallgren: Given that #enclosing_scope now might be a full lookup, it should only take place once in #call_by_name and then be accessed from a local variable.

lib/puppet/pops/evaluator/evaluator_impl.rb
```diff
@@ -42,19 +42,26 @@ class EvaluatorImpl
...
+
+  # @api private
+  def static_initialize
     @@eval_visitor     ||= Visitor.new(self, "eval", 1, 1)
     @@lvalue_visitor   ||= Visitor.new(self, "lvalue", 1, 1)
     @@assign_visitor   ||= Visitor.new(self, "assign", 3, 3)
     @@string_visitor   ||= Visitor.new(self, "string", 1, 1)
 
-    @@type_calculator  ||= Types::TypeCalculator.new()
+    @@type_calculator  ||= Types::TypeCalculator.singleton()
```
thallgren: Nitpick, loose the extra parenthesis

lib/puppet/pops/evaluator/evaluator_impl.rb
```diff
@@ -42,19 +42,26 @@ class EvaluatorImpl
...
 
-    @@compare_operator     ||= CompareOperator.new()
+    @@compare_operator      ||= CompareOperator.new()
```
thallgren: Nitpick, loose the extra parenthesis

lib/puppet/pops/evaluator/evaluator_impl.rb
```diff
@@ -42,19 +42,26 @@ class EvaluatorImpl
   Issues = Issues
 
   def initialize
+    @@initialized ||= static_initialize
+
+    # Use null migration checker unless given in context
+    @migration_checker = (Puppet.lookup(:migration_checker) { Migration::MigrationChecker.singleton() })
```
thallgren: Two pairs of unnecessary parenthesis here. One around the expression and one after singleton.

lib/puppet/pops/loader/puppet_function_instantiator.rb
```diff
@@ -64,18 +65,19 @@ def self.create(loader, typed_name, source_ref, pp_code_string)
   #   typed name, and an instantiated function with global scope closure associated with the given loader
   #
   def self.create_from_model(function_definition, loader)
-    closure_scope = Puppet.lookup(:global_scope) { {} }
-    created = create_function_class(function_definition, closure_scope)
+    closure_scope = nil; # Puppet.lookup(:global_scope) { {} }
```
thallgren: Remove this line (pass nil to the constructor instead)

lib/puppet/pops/migration/migration_checker.rb
```diff
@@ -6,6 +6,10 @@ class Puppet::Pops::Migration::MigrationChecker
   def initialize()
   end
 
+  def self.singleton
+    @@null_checker ||= self.new
```
thallgren: This is a class variable on a class method. Isn't a class instance variable more appropriate?

lib/puppet/pops/loader/puppet_function_instantiator.rb
```diff
@@ -64,18 +65,19 @@ def self.create(loader, typed_name, source_ref, pp_code_string)
...
+    created = create_function_class(function_definition)
     typed_name = TypedName.new(:function, function_definition.name)
     [typed_name, created.new(closure_scope, loader)]
   end
 
-  def self.create_function_class(function_definition, closure_scope)
+  def self.create_function_class(function_definition)
     # Create a 4x function wrapper around a named closure
     Puppet::Functions.create_function(function_definition.name, Puppet::Functions::PuppetFunction) do
+      # This is highly problematic - it binds both an Evaluator and closure_scope
```
thallgren: This comment seems obsolete.

lib/puppet/pops/migration/migration_checker.rb
hlindberg: as I understand it, @@name makes it shared among subclasses (even if done at class level) as opposed to @name in a class - I have been bit by that in the past. Here I don't think it matters.

lib/puppet/pops/loader/puppet_function_instantiator.rb
hlindberg: Half of it is obsolete - it still binds an evaluator - which it does not need to do - the evaluator holds on to a migration checker (it either creates a new null instance of one, or looks one up). If instead using a shared evaluator we would gain 2 object instances per function, and some performance. Changing the comment to a TODO wrt to the evaluator.new

PR #5216, opened 2016-08-18
(PUP-6631) ensure packages are not autoremoved

binford2k:
Someone with more Debian chops than me should review this heavily. I
don't know whether apt-mark is even installed by default on all Debian
flavors, or whether there will be unwanted side effects.

### DIFF
--- a/lib/puppet/provider/package/apt.rb
+++ b/lib/puppet/provider/package/apt.rb
@@ -12,6 +12,7 @@ Puppet::Type.type(:package).provide :apt, :parent => :dpkg, :source => :dpkg do
 
   commands :aptget => "/usr/bin/apt-get"
   commands :aptcache => "/usr/bin/apt-cache"
+  commands :aptmark => "/usr/bin/apt-mark"
   commands :preseed => "/usr/bin/debconf-set-selections"
 
   defaultfor :osfamily => :debian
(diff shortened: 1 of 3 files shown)


binford2k:
(and for the record, I don't see why it doesn't get marked as manual by default, since we're just calling `apt-get install`)

adrienthebo:
On debian 8:

```
root@c09f6df09c30:/# dpkg -S /usr/bin/apt-mark
apt: /usr/bin/apt-mark
```

branan:
I don't think this will actually fix the issue on the ticket. My understanding is that the issue is as follows:

* Install package `foo`, which depends on `bar`
* Apply a puppet manifest that wants `bar` installed. This is a no-op
* Uninstall `foo`. `bar` will be uninstalled.

The change that's needed, then, is that when Puppet has a package resource on debian it should consider it out-of-sync if that package isn't marked as `manual`, so that the marking can happen. Adding marking to all operations won't fix that, and shouldn't be necessary - any *explicit* call to install a package with apt-get will have already marked that package as `manual`

adrienthebo:
It is possible to make package marking work as desired, though it'll require more work. In order for a package to be marked that's not installed it needs to be treated as out of sync, which requires a new `mark` property. Let me know if you want to explore this implementation and I can give more details, but as it stands this pull request can't be merged since it doesn't address the root issue.

MosesMendoza:
@binford2k if you're interested in proceeding further with this PR, please ping @adrienthebo to continue with the needed changes. In the interim I'm going to go ahead and close this as it's not merge-able as is and its been a couple months, and we can reopen once changes are complete.

PR #5217, opened 2016-08-18
(PUP-6502) Reinstate dscl read for OSX user exists?

johnduarte:
This commit adds back the dscl read implementation for the `exists?`
method of the user directoryservice provider. Currently, the
`@property_hash` is not being refreshed upon a `create` event. Therefore,
the previous query to `@property_hash.any?` returned `false even
though the user had been created.

### DIFF
--- a/lib/puppet/provider/user/directoryservice.rb
+++ b/lib/puppet/provider/user/directoryservice.rb
@@ -237,7 +237,13 @@ Puppet::Type.type(:user).provide :directoryservice do
 ##                   ##
 
   def exists?
-    return @property_hash.any?
+    begin
+      dscl '.', 'read', "/Users/#{@resource.name}"
+    rescue Puppet::ExecutionFailure => e
+      Puppet.debug("User was not found, dscl returned: #{e.inspect}")
+      return false
+    end
+    true
   end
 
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/user/directoryservice.rb
```diff
@@ -151,6 +151,10 @@ def self.generate_attribute_hash(input_hash)
         attribute_hash[:iterations] = get_salted_sha512_pbkdf2('iterations', embedded_binary_plist)
       elsif embedded_binary_plist['SALTED-SHA512']
         attribute_hash[:password] = get_salted_sha512(embedded_binary_plist)
+      else
```
adrienthebo: This is a bit funny because we have a catchall conditional but we're specifically calling the `get_salted_sha512_pbkdf2` function, which seems specific for SALTED-SHA512-PBKDF2. Why does this work, and what conditionals are we catching or what algorithm are we using?

Magisus:
Verified that this fixes the [test failure](https://jenkins.puppetlabs.com/view/puppet-agent/view/master/view/puppet-agent/job/platform_puppet-agent_intn-van-sys_suite-daily-puppet-master/189/). Going to merge, barring any complaints, to unblock CI.

PR #5219, opened 2016-08-19
(maint) Correct usage of log_exception in property.rb

kbarber:
We were using a four argument log_exception instead of a 3. This fixes the problem
and puts in a test for the condition.


kbarber:
@branan this was the bug I found for log_exception.

branan:
Whoops, I guess I should have double-checked our negative tests :/. Good catch.

PR #5220, opened 2016-08-19
(PUP-6629) Be more careful about parsing YAML for transactionstore

kbarber:
This patch strips Ruby tags from the transactionstore before parsing, so we avoid
some deserialization issues.


ferventcoder:
Looks like it will do the needful

PR #5222, opened 2016-08-19
(PUP-2018) Only submit one CSR with `certificate generate`

adrienthebo:
Puppet uses a fair amount of magic to request a CSR when using HTTPS for
the first time. While this normally works well enough when running a
Puppet agent this falls over badly with `puppet certificate generate`.
If the command is run on a node that hasn't submitted a CSR for itself
it will submit a CSR for itself before it submits a CSR for the node
requested on the CLI. If `puppet certificate generate` is used to
submit a CSR for the current node this fails badly as Puppet will try to
submit two CSRs for the same node, causing the command to fail.

The culprit behind this is `Puppet::SSL::Host.localhost`, which when
called will generate a CSR for the given node if it doesn't exist. This
function is used when setting up SSL validation and identity checking,
so by trying to validate a connection a CSR will be submitted. This code
path is brittle but since so much other code relies on this magic
initialization we can't really change this.

In order to sidestep this issue, this commit uses the Puppet context to
store the SSL::Host used for SSL validation and conditionally
substitutes another host object in when no signed certs exist.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -197,7 +197,8 @@ module Puppet
       :http_pool => proc {
         require 'puppet/network/http'
         Puppet::Network::HTTP::NoCachePool.new
-      }
+      },
+      :ssl_host => proc { Puppet::SSL::Host.localhost },
     }
   end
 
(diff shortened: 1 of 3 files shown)


adrienthebo:
This PR could really stand having an acceptance test added before we merge.

PR #5224, opened 2016-08-21
(PUP-6637) New attribute :manage_mountpoint for Mount type

mschuchard:
With this new feature, Puppet will be able to manage mountpoint directories prior to mounting or after unmounting without causing a multiple declaration error.

(WIP)
Questions:
1. Should the created mountpoint directory have 400 permissions, or is root:root ownership good enough?
2. How does the mount type behave if the parent directory or mountpoint do not exist? Is that being verified somewhere or do I need to implement some code for it?
3. What should be done if the mountpoint directory has contents after unmounting?
4. The provider spec tests have a name of `'/'` for all of the test resources. This doesn't really work for any tests I would write for this new attribute. What direction should I go in for that?

### DIFF
--- a/lib/puppet/provider/mount.rb
+++ b/lib/puppet/provider/mount.rb
@@ -5,6 +5,9 @@ require 'puppet'
 module Puppet::Provider::Mount
   # This only works when the mount point is synced to the fstab.
   def mount
+    # Create the mountpoint directory if it does not exist and manage_mountpoint is true
+    Dir.mkdir(resource[:name]) if resource[:manage_mountpoint] && !File.exist?(resource[:name])
+
     args = []
 
     # In general we do not have to pass mountoptions because we always
@@ -43,6 +46,9 @@ module Puppet::Provider::Mount
   # This only works when the mount point is synced to the fstab.
   def unmount
     umount(resource[:name])
+
+    # Remove the mountpoint directory if it exists and manage_mountpoint is true
+    Dir.rmdir(resource[:name]) if resource[:manage_mountpoint] && File.exist?(resource[:name])
 
     # Update property hash for future queries (e.g. refresh is called)
     case get(:ensure)
--- a/lib/puppet/type/mount.rb
+++ b/lib/puppet/type/mount.rb
@@ -275,6 +275,23 @@ module Puppet
       end
     end
 
+    newparam(:manage_mountpoint) do
+      desc "Whether to create the directory mountpoint prior to mounting the
+        directory and remove the directory mountpoint after unmounting the
+        directory. Does nothing during a remount.
+
+        This is useful if you have a `file` resource establishing
+        permissions of a mounted directory and also need the mountpoint
+        directory created prior to mounting, or if you want to cleanly change
+        the `file` `ensure` attribute after unmounting (both avoiding a
+        multiple declaration).
+
+        Valid values are `true` or `false`. Default is `false`."
+
+      newvalues(:true, :false)
+      defaultto false
+    end
+
     def refresh
       # Only remount if we're supposed to be mounted.
       provider.remount if self.should(:fstype) != "swap" and provider.mounted?
--- a/spec/unit/type/mount_spec.rb
+++ b/spec/unit/type/mount_spec.rb
@@ -50,7 +50,7 @@ describe Puppet::Type.type(:mount), :unless => Puppet.features.microsoft_windows
   end
 
   describe "when validating attributes" do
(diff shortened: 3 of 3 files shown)


branan:
Thanks for doing the work on this so quickly!

I do not have answers to your questions yet (or even some of my own). I'm going to talk to @ahpook or another product person about this design "soon" (realistically probably that means "in the next few weeks") and get back to you on this.

mschuchard:
Thanks for responding. I realized I needed to add a couple words on 2 to clarify that I meant the mount type in Puppet and not the mount command on *nix.

Since there are several ways to go about doing this, I want to make sure I am doing it the optimal way. With some guidance from you guys I should be able to achieve that. I had also looked into invoking the file type from the mount provider to create or remove the directory, but I was concerned about potential unforeseen side effects there.

mschuchard:
New push: two word differences in the `desc` which I think improves clarity and attempting to fix the two `type` spec errors.

ffrank:
Is it a conscious decision to forgo the additional resource generator (i.e., add a `generate` hook to the `mount` type) in favor of handling the checking and directory creation "manually"?

The `nagios` types already do something similar in order to seamlessly manage config file permissions.

mschuchard:
So this reminds me: the 'default false' test is failing because it is returning `nil`. Unless I did that test incorrectly, that invalidates the 'no default' tests in the spec (which I mimicked) because they all consider returning `nil` to be a success, but that might actually be happening because the default value is being returned incorrectly.

mschuchard:
Soft ping on the original PR and the concern over the mount rspec tests.

lib/puppet/provider/mount.rb
```diff
@@ -5,6 +5,9 @@
 module Puppet::Provider::Mount
   # This only works when the mount point is synced to the fstab.
   def mount
+    # Create the mountpoint directory if it does not exist and manage_mountpoint is true
+    Dir.mkdir(resource[:name]) if resource[:manage_mountpoint] && !File.exist?(resource[:name])
```
melissa: The directory permissions here probably don't matter, but a default of `0500` is probably just fine. I don't think it's something we want to make configurable though.

lib/puppet/provider/mount.rb
```diff
@@ -44,6 +47,9 @@ def remount
   def unmount
     umount(resource[:name])
 
+    # Remove the mountpoint directory if it exists and manage_mountpoint is true
+    Dir.rmdir(resource[:name]) if resource[:manage_mountpoint] && File.exist?(resource[:name])
```
melissa: I don't think we should even be attempting to remove the directory on unmount. If the user really don't want this to exist anymore, they should explicitly set that in a file resource. Otherwise this is inconsistent with established behavior. As far as I'm aware, puppet doesn't clean anything up unless explicitely told to.

I would also suggest changing the param name to something like `create_mountpoint` to more accurately reflect this change.

lib/puppet/type/mount.rb
```diff
@@ -275,6 +275,23 @@ def munge(value)
       end
     end
 
+    newparam(:manage_mountpoint) do
```
melissa: You can add `:boolean => true` here, which will get us the use of `manage_mountpoint?`, courtesy of https://github.com/puppetlabs/puppet/blob/327b03a9e1ecdc7b925720a4eec59795f4456a56/lib/puppet/type.rb#L283-L293

melissa:
The issue with the test to make sure it's defaulting to false is actually a bug in the code. I tracked it to https://github.com/puppetlabs/puppet/blob/327b03a9e1ecdc7b925720a4eec59795f4456a56/lib/puppet/type.rb#L834
Since `parameter.default` in this case is `false`, that's short circuiting that check and removing `manage_mountpoint` from the list of parameters. I have a feeling fixing this is going to cause at least a few unexpected changes. I've opened https://tickets.puppetlabs.com/browse/PUP-8442 so we can devote some time to making sure we fix this properly.

ahpook:
i'm not at all opposed to this PR but note that the issues around management of the mounted filesystem vs the underlying mountpoints have some prior art that @jhelwig and i worked on wayyyy back in the day, so there might be something to look at here:  https://github.com/puppetlabs/puppetlabs-mount_providers

melissa:
@ahpook makes a good point about `mount_providers`. That's actually a better place to target this PR. We want to slowly and gently start to push users to that module rather than this one. I'm going to close this PR, and retarget the associated ticket to indicate that's where this work should happen

PR #5225, opened 2016-08-22
(doc) (DOCUMENT-579) Fix EPP function example path

roman-mueller:
Currently the example at: https://docs.puppet.com/puppet/latest/reference/function.html#epp
uses an incorrect path in the example.

This PR fixes that.


branan:
Thanks for this! Documentation bugs are easy to overlook

PR #5226, opened 2016-08-22
(PUP-6592) acceptance: Sensitive data

er0ck:
* add a test for Sensitive data type and its functions
* test employs a first pass at table driven testing
  * could be used for other tests for puppet language and resource types
  * the code generation portion could use a bit of work prior to expansion

[skip ci]

### DIFF
--- a/acceptance/tests/language/sensitive_data_type.rb
+++ b/acceptance/tests/language/sensitive_data_type.rb
@@ -1,0 +1,100 @@
+test_name 'C98120, C98077: Sensitive Data is redacted on CLI, logs, reports' do
+require 'puppet/acceptance/puppet_type_test_tools.rb'
+extend Puppet::Acceptance::PuppetTypeTestTools
+
+  app_type        = File.basename(__FILE__, '.*')
+  tmp_environment = mk_tmp_environment_with_teardown(master, app_type)
+  fq_tmp_environmentpath  = "#{environmentpath}/#{tmp_environment}"
+
+  tmp_filename_win = tmp_filename_else = ''
+  agents.each do |agent|
+    # ugh... this won't work with more than two agents of two types
+    tmp_filename_win  = "C:\\cygwin64\\tmp\\#{tmp_environment}.txt"
+    tmp_filename_else = "/tmp/#{tmp_environment}.txt"
+    if agent.platform =~ /windows/
+      tmp_filename = tmp_filename_win
+    else
+      tmp_filename = tmp_filename_else
+    end
+    on agent, "echo 'old content' > /tmp/#{tmp_environment}.txt"
+  end
+
+  # first attempts at a reasonable table driven test.  needs API work
+  # FIXME:
+  #   expand this to other resource types, make parameters arbitrary, make assertions arbitrary
+  # FIXME: add context messaging to each instance
+  notify_redacted = 'Sensitive \[value redacted\]'
+  file_redacted   = 'changed \[redacted\] to \[redacted\]'
+  test_resources = [
+    {:type => 'notify', :parameters => {:namevar => "1:${Sensitive.new('sekrit1')}"},
+       :assertions => [{:refute_match => 'sekrit1'}, {:assert_match => "1:#{notify_redacted}"}]},
+    {:type => 'notify', :parameters => {:namevar => "2:${Sensitive.new($meh2)}"}, :pre_code => '$meh2="sekrit2"',
+       :assertions => [{:refute_match => 'sekrit2'}, {:assert_match => "2:#{notify_redacted}"}]},
+    {:type => 'notify', :parameters => {:namevar => "3:meh", :message => '"3:${Sensitive.new(\'sekrit3\')}"'},
+       :assertions => [{:refute_match => 'sekrit3'}, {:assert_match => "3:#{notify_redacted}"}]},
+    {:type => 'notify', :parameters => {:namevar => "4:meh", :message => "Sensitive.new($meh4)"}, :pre_code => '$meh4="sekrit4"',
+       :assertions => {:expect_failure => {:refute_match => 'sekrit4', :message => 'you can always spill redacted data, if you want to'}}},
+    {:type => 'notify', :parameters => {:namevar => "5:meh", :message => "$meh5"}, :pre_code => '$meh5=Sensitive.new("sekrit5")',
+       :assertions => {:expect_failure => {:refute_match => 'sekrit5', :message => 'you can always spill redacted data, if you want to'}}},
+    {:type => 'notify', :parameters => {:namevar => "6:meh", :message => '"6:${meh6}"'}, :pre_code => '$meh6=Sensitive.new("sekrit6")',
+       :assertions => [{:refute_match => 'sekrit6'}, {:assert_match => "6:#{notify_redacted}"}]},
+    {:type => 'notify', :parameters => {:namevar => "7:${Sensitive('sekrit7')}"},
+       :assertions => [{:refute_match => 'sekrit7'}, {:assert_match => "7:#{notify_redacted}"}]},
+    # unwrap(), these should be en-clair
+    {:type => 'notify', :parameters => {:namevar => "8:${unwrap(Sensitive.new('sekrit8'))}"},
+       :assertions => {:assert_match => "8:sekrit8"}},
+    {:type => 'notify', :parameters => {:namevar => "9:meh", :message => '"9:${unwrap(Sensitive.new(\'sekrit9\'))}"'},
+       :assertions => {:assert_match => "9:sekrit9"}},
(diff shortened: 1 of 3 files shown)


er0ck:
@johnduarte i'd love some review on the funky table driven (ish) test i've got going here.  i'd like to know how obvious it is to you, how easy it'd be for you to add a test snippet, etc

johnduarte:
brilliant

@er0ck this is effin amazing! We should bake this table meta-block construct in Beaker itself.

This makes adding a snippet trivially easy.

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  # FIXME:
+  #   expand this to other resource types
+  # first attempts at a reasonable table driven test.  needs API work
+  # FIXME: add context messaging to each instance
+  redacted = 'Sensitive \[value redacted\]'
+  test_resources = [
+    {:type => 'notify', :parameters => {:namevar => "1:${Sensitive.new('sekrit1')}"}, :refute => {:match => 'sekrit1'}, :assert => {:match => "1:#{redacted}"}},
+    {:type => 'notify', :parameters => {:namevar => "2:${Sensitive.new($meh2)}"}, :pre_code => '$meh2="sekrit2"', :refute => {:match => 'sekrit2'}, :assert => {:match => "2:#{redacted}"}},
+    {:type => 'notify', :parameters => {:namevar => "3:meh", :message => '"3:${Sensitive.new(\'sekrit3\')}"'}, :refute => {:match => 'sekrit3'}, :assert => {:match => "3:#{redacted}"}},
```
johnduarte: Refactoring suggestion: The `:message` param could be placed inside another hash in order to allow type parameters to be isolated from the other parameters. This would prevent coupling the type data and the type in the logic below. The extrapolation of the type parameters would then just be an iterator.

E.G.
```
test_resources = {
    {:type => 'notify',
        :parameters => {
            :type_params => {:namevar => "3:meh", :message => '"3:${Sensitive.new(\'sekrit3\')}"'}},
            :refute => {:match => 'sekrit3'}, :assert => {:match => "3:#{redacted}"}}
    }

...
      sitepp_contents << resource[:type] + '{"' + resource[:parameters][:namevar] + '":'
      resource[:parameters][:type_params].each do |key, value|
          sitepp_contents << "#{key} =>  '#{value}',\n"
      end
...
```

er0ck: agree completely.  i have a file resource i added recently.  the params should be auto-populated into puppet resource code.

er0ck:
heh.  thanks.  epic gif.  
yeah, i really need a beaker-puppet library to put this in.  its state of missing has been keeping me from making changes upstream in a few ways. (facepalm)

er0ck:
@johnduarte can you take a look?

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -0,0 +1,86 @@
+test_name 'C98120, C98077: Sensitive Data is redacted on CLI, logs, reports' do
+require 'puppet/acceptance/puppet_type_test_tools.rb'
+extend Puppet::Acceptance::PuppetTypeTestTools
+
+  app_type        = File.basename(__FILE__, '.*')
+  tmp_environment = mk_tmp_environment_with_teardown(master, app_type)
+  fq_tmp_environmentpath  = "#{environmentpath}/#{tmp_environment}"
+
+  tmp_filename = "/tmp/sensitive_content_#{random_string}.txt"
```
johnduarte: Unfortunately, this path is not valid on Windows. Either the `generate_` commands need to be aware of the context of the target agent, or the resource hashes themselves need to include conditionals for `osfamily`.

er0ck:
@johnduarte ugh... thanks, dunno what i was thinking.  fixed this up.

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -0,0 +1,99 @@
...
+           :accept_all_exit_codes => true) do |result|
+          assert(result.exit_code==2,'puppet agent run failed')
+          run_assertions(assertion_code, result)
+        end
+
+        unless agent.platform =~ /windows/
+          step "assert no redacted data in syslog" do
+            #sigh.  gee, thanks for the help, beaker!
+            result = dump_puppet_log(agent)
+            raise if result.length < 3
```
johnduarte: This helper method is defined as private and is not designed to return a value on all platforms. This currently causes the test to fail on solaris where return value is nil. See [beaker source](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers/puppet_helpers.rb#L274-L287).

whopper: @er0ck looks like John's comment got buried. Just leaving this note with the hope that it'll get unburied.

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -0,0 +1,99 @@
...
+        on(agent, puppet("agent -t --debug --trace --show_diff --server #{master.hostname} --environment #{tmp_environment}"),
+           :accept_all_exit_codes => true) do |result|
+          assert(result.exit_code==2,'puppet agent run failed')
+          run_assertions(assertion_code, result)
+        end
+
+        if agent['platform'] =~ /fedora|centos|el|redhat|scientific|ubuntu|debian|cumulus/
+          step "assert no redacted data in syslog" do
+            #sigh.  gee, thanks for the help, beaker!
+            result = dump_puppet_log(agent)
```
whopper: Argh, GitHub! Here's John's comment about `dump_puppet_log`, just for everyone's visibility:

This helper method is defined as private and is not designed to return a value on all platforms. This currently causes the test to fail on solaris where return value is nil. See [beaker source](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers/puppet_helpers.rb#L274-L287).

er0ck: the above changed that.  not the best fix, and i suppose we can write our own helper for grabbing the puppet log on various platforms.  but for now this works for now, based on copy, paste of the platforms supported in beaker.

i'd prefer to leave it like this.  but i can re-write if people think it's important in this sprint

johnduarte: @er0ck , sorry I am missing something. I am not sure what you are referencing when you say that "the above changed that". For Solaris, the beaker helper does not set a value for `syslog`. So, nothing is returned and the section for `logger` is not called.

whopper:
Oh, my mistake! The above conditional ` if agent['platform'] =~ /fedora|centos|el|redhat|scientific|ubuntu|debian|cumulus/
` covers everything and prevents the Solaris failures.

acceptance/tests/language/sensitive_data_type.rb
whopper: Oh, my mistake! The above conditional ` if agent['platform'] =~ /fedora|centos|el|redhat|scientific|ubuntu|debian|cumulus/
` covers everything and prevents the Solaris failures.

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -0,0 +1,99 @@
...
+            run_assertions(assertion_code, result[string_io_index].string)
+          end
+        end
+
+        # don't do this before the agent log scanning, above. it will skew the results
+        step "assert no redacted data in vardir" do
+          on(agent, "grep -R sekrit #{agent.puppet['vardir']}", :accept_all_exit_codes => true) do |result|
+            refute_match(/sekrit(1|2|3|6|7)/, result.stdout, 'found redacted data we should not have')
+            #TODO: if/when this is fixed, we should just be able to eval(assertion_code_ in this result block also!
+            expect_failure 'file resource contents will end up in the cached catalog en-clair' do
```
johnduarte: This passes unexpectedly on solaris10-x64 using `1.6.2.133.g82d98be`, but not on Windows.

```
Warning: tests/language/sensitive_data_type.rb errored in 53.77 seconds
Test Case tests/language/sensitive_data_type.rb reported: #<RuntimeError: PuppetAcceptance::DSL::Helpers.with_puppet_running_on failed (check backtrace for location) because: An assertion was expected to fail, but passed. This is probably because a product bug was fixed, and "expect_failure()" needs to be removed from this test. Additional info: 'file resource contents will end up in the cached catalog en-clai
```

er0ck: do you have a link to the output?  i'm not clear how this could happen

hlindberg:
@whopper @johnduarte - are you happy with this PR now ?

whopper:
👍 passes on centos 7, solaris 11 and win 2012r2

PR #5230, opened 2016-08-24
(PUP-6338) Make contain, include and require include return class references

hlindberg:
This makes the contain, include and require functions return an Array[Type[Class]] with references to the classes the function got as arguments.
This allows chaining further operations on that set without having to use additional local variables.

This work triggered a refactor of the functions from the 3.x API to 4.x, for tests to move (and be sanitized).  A redundant test was removed.


spec/unit/functions/contain_spec.rb
```diff
@@ -295,5 +295,4 @@ class container {
 
   it_should_behave_like 'all functions transforming relative to absolute names', :contain
   it_should_behave_like 'an inclusion function, regardless of the type of class reference,', :contain
-
```
thallgren: The commit comment claims that a test is added in this commit. Where is it?

lib/puppet/functions/include.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def include(scope, *classes)
+    classes = scope.transform_and_assert_classnames(classes.flatten)
+    result = classes.map {|name| Puppet::Pops::Types::TypeFactory.host_class(name) }
+    scope.compiler.evaluate_classes(classes, scope, false)
+
+    # Result is an Array[Class, 1, n] which allows chaining other operations
+    result
+  end
+end
```
thallgren: Missing newline

lib/puppet/functions/require.rb
```diff
@@ -0,0 +1,40 @@
...
+    # This is the same as calling the include function (but faster) since it again
+    # would otherwise need to perform the optional absolute name transformation
+    # (for no reason since they are already made absolute here).
+    #
+    scope.compiler.evaluate_classes(classes, scope, false)
+    krt = scope.environment.known_resource_types
+
+    classes.each do |klass|
+      # lookup the class in the scopes
+      if classobj = krt.find_hostclass(klass)
```
thallgren: Please avoid assignments in places where a boolean expression is expected. Very easy to read as `if classobj == krt.find_hostclass(klass)`.

spec/unit/functions/require_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+
+describe 'The "require" function' do
+  include PuppetSpec::Compiler
+  include ContainmentMatchers
+  include Matchers::Resource
+
+  before(:each) do
+    compiler  = Puppet::Parser::Compiler.new(Puppet::Node.new("foo"))
+    @scope = Puppet::Parser::Scope.new(compiler)
```
thallgren: Do you really need a separate scope here? I've had problems with this construct in the past. Solved by using:
```
@scope = compiler.topscope
```

PR #5232, opened 2016-08-25
(PUP-6650) Ensure that environment compiler overrides are removed

thallgren:
Before this commit, the EnvironmentCompiler would add function overrides
when it was instantiated (once for each compile). This approach was
fatal in a long lived environment (where a compile happens more than
once) since a loader doesn't accept a redefine.

This commit ensures that overridden functions are added at the beginning
of a compile and then subsequently removed at the end of that compile.
Overridden functions are preserved and restored in the process.


hlindberg:
Fails on Windows with Ruby 2.3.0 on unrelated issue (missing spec_order.txt).

PR #5233, opened 2016-08-25
(PUP-6629) Respect the passed log level when log_exception is passed a custom message

branan:
The first time we tried to fix PUP-6629, we did not fix all the code
paths in log_exception - specifically, when a custom message was
passed (which insync_values? does), we still logged at error level.

This now handles that case, and also adds a test for insync_values? to
ensure it does not log at error level if insync? fails.

### DIFF
--- a/lib/puppet/util/logging.rb
+++ b/lib/puppet/util/logging.rb
@@ -67,7 +67,7 @@ module Logging
           :node => exception.node
         }.merge(log_metadata))
     else
-      err(format_exception(exception, message, trace))
+      send_log(options[:level] || :err, format_exception(exception, message, trace))
     end
   end
 
--- a/spec/unit/agent_spec.rb
+++ b/spec/unit/agent_spec.rb
@@ -116,7 +116,7 @@ describe Puppet::Agent do
 
     it "should not fail if a client class instance cannot be created" do
       AgentTestClient.expects(:new).raises "eh"
-      Puppet.expects(:err)
+      Puppet::Util::Log.expects(:create).with(has_entry(:level => :err))
       @agent.run
     end
 
@@ -124,7 +124,7 @@ describe Puppet::Agent do
       client = AgentTestClient.new
       AgentTestClient.expects(:new).returns client
       client.expects(:run).raises "eh"
-      Puppet.expects(:err)
+      Puppet::Util::Log.expects(:create).with(has_entry(:level => :err))
       @agent.run
     end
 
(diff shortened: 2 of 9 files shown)


lib/puppet/util/logging.rb
```diff
@@ -67,7 +67,7 @@ def log_exception(exception, message = :default, options = {})
           :node => exception.node
         }.merge(log_metadata))
     else
-      err(format_exception(exception, message, trace))
+      send_log(options[:level] || :err, format_exception(exception, message, trace))
```
Iristyle: To take a step back on this one... why is that the transaction code needs to call `log_exception`?  And is logging exceptions at a user configurable level really something we want to allow?

I get that there are a few "helpers" in here pulling some extra data out of the exception - but that's almost entirely related to functionality that the parser needs (added in https://github.com/puppetlabs/puppet/commit/45e3f41a38f7f4af819ed46f08c6fb8606436ed7).

Is there a reason to not instead just call `Puppet.info` at https://github.com/puppetlabs/puppet/blob/stable/lib/puppet/property.rb#L370 like this instead?

```
Puppet.info("Unknown failure comparing values #{should} and #{is} using insync? on type: #{self.resource.ref} property: #{self.name}")
```

Then you don't have to touch this shared logging method...

branan: Ideally nothing will ever raise an exception during this, so logging at `info` level instead of letting it bubble up is an acknowledgement that we're doing new/weird things, and that not all code will be compatible.

By logging the full exception and allowing `--trace` to work correctly, we provide the information that module authors need to make their code play nice with this functionality, if so desired.

branan:
Ugh, looks like there are some places that expected `log_exception` to call `err` directly :( Once again the levels at which we choose to mock make me sad.

Those are easy enough to clear up, though.

branan:
I'm not really happy with the way I changed the mocks here - I had issues trying to add expects for `Puppet::Util::Logging::log_exception`, and sorta gave up in the interests of getting it working. I do want to spend more time trying to make those less dumb if I can.

branan:
running the ACL tests against this PR passes! I still want to figure out if I can make the test changes less terrible, but this should get us to where we want to be.

Iristyle:
I have pushed up #5241 as an alternative implementation, that requires fewer code changes (and no spec changes).  I'd be more comfortable with just changing the offending caller than the behavior of the logging code, considering this is targeted at a bugfix z release that we're shooting to turn around quickly.

If we want to change logging behavior, I suggest we slate that for a future release.

PR #5234, opened 2016-08-26
(PUP-6653) Avoid ruby type loading and type redefine clashes

thallgren:
This commit ensures that Puppet::Resource instances created during
compile will not make attempts to load a Puppet::Type.

### DIFF
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -226,12 +226,12 @@ class Puppet::Resource
 
       src.to_hash.each do |p, v|
         if v.is_a?(Puppet::Resource)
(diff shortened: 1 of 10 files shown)


lib/puppet/resource.rb
```diff
@@ -274,12 +274,13 @@ def initialize(type, title = nil, attributes = {})
 
       @type, @title = self.class.type_and_title(type, title)
 
-      if params = attributes[:parameters]
+      rt = resource_type
```
hlindberg: Use `rt` on line 290 as well?

hlindberg:
Needs tests as well as the holes now found were not found by existing tests.

PR #5235, opened 2016-08-29
(PUP-5871) Add Timespan and Timestamp Puppet Types and implementations

thallgren:
This commit adds two new types. A Timespan that represents a moment
in time and a Timespan that represents a duration. Both types uses
nano-second precision and can participate in arithmetic operations.

The types have `new` functions that allow creation from a string a hash.

A strftime function is also included to provide a flexible way of
representing the instances in string form. The name was chosen since
it's well known in languages such as Ruby, C++, and Python. It's also
less likely to collide with other global functions declared in modules
than, say 'format'.

An approach of extending the String constructor and give it special
logic for Timespan and Timestamp was rejected since it would introduce
ambiguities with different formatting characters. `String(x, '%p')`
(expected to produce a programmatic representation) would print 'AM' or
'PM' in case `x` is a Timestamp.

The Timespan and Timestamp implementation share an abstract class
TimeData which in turn inherits from Numeric. This approach was selected
since it required a minimum of adjustments to make the classes work
with arithmetic and comparison operators in PL.

### DIFF
--- a/lib/puppet/functions/strftime.rb
+++ b/lib/puppet/functions/strftime.rb
@@ -1,0 +1,14 @@
+# (Documentation in 3.x stub)
+#
+# @since 4.7.0
+#
+Puppet::Functions.create_function(:strftime) do
+  dispatch :format do
+    param 'Variant[Timestamp,Timespan]', :time_object
+    param 'String', :format
+  end
+
+  def format(time_object, format)
+    time_object.format(format)
+  end
+end
--- a/lib/puppet/parser/functions/strftime.rb
+++ b/lib/puppet/parser/functions/strftime.rb
@@ -1,0 +1,173 @@
+Puppet::Parser::Functions::newfunction(
+  :strftime,
+  :type => :rvalue,
+  :arity => 2,
+  :doc => <<DOC
+Formats timestamp or timespan according to the directives in the given format string. The directives begins with a percent (%) character.
+Any text not listed as a directive will be passed through to the output string.
+
+The directive consists of a percent (%) character, zero or more flags, optional minimum field width and
+a conversion specifier as follows:
+```
+%[Flags][Width]Conversion
+```
+
+### Flags that controls padding
+
+| Flag  | Meaning
+| ----  | ---------------
+| -     | Don't pad numerical output
+| _     | Use spaces for padding
+| 0     | Use zeros for padding
+
+### `Timestamp` specific flags
+
+| Flag  | Meaning
+| ----  | ---------------
+| #     | Change case
+| ^     | Use uppercase
+| :     | Use colons for %z
+
+### Format directives applicable to `Timestamp` (names and padding can be altered using flags):
+
+**Date (Year, Month, Day):**
+
+| Format | Meaning |
+| ------ | ------- |
+| Y | Year with century, zero-padded to at least 4 digits |
+| C | year / 100 (rounded down such as 20 in 2009) |
+| y | year % 100 (00..99) |
+| m | Month of the year, zero-padded (01..12) |
+| B | The full month name ("January") |
+| b | The abbreviated month name ("Jan") |
+| h | Equivalent to %b |
+| d | Day of the month, zero-padded (01..31) |
+| e | Day of the month, blank-padded ( 1..31) |
+| j | Day of the year (001..366) |
+
+**Time (Hour, Minute, Second, Subsecond):**
+
+| Format | Meaning |
+| ------ | ------- |
+| H | Hour of the day, 24-hour clock, zero-padded (00..23) |
+| k | Hour of the day, 24-hour clock, blank-padded ( 0..23) |
+| I | Hour of the day, 12-hour clock, zero-padded (01..12) |
+| l | Hour of the day, 12-hour clock, blank-padded ( 1..12) |
+| P | Meridian indicator, lowercase ("am" or "pm") |
+| p | Meridian indicator, uppercase ("AM" or "PM") |
+| M | Minute of the hour (00..59) |
+| S | Second of the minute (00..60) |
+| L | Millisecond of the second (000..999). Digits under millisecond are truncated to not produce 1000 |
+| N | Fractional seconds digits, default is 9 digits (nanosecond). Digits under a specified width are truncated to avoid carry up |
+
+**Time (Hour, Minute, Second, Subsecond):**
+
+| Format | Meaning |
+| ------ | ------- |
+| z   | Time zone as hour and minute offset from UTC (e.g. +0900) |
+| :z  | hour and minute offset from UTC with a colon (e.g. +09:00) |
+| ::z | hour, minute and second offset from UTC (e.g. +09:00:00) |
+| Z   | Abbreviated time zone name or similar information.  (OS dependent) |
+
+**Weekday:**
+
+| Format | Meaning |
+| ------ | ------- |
+| A | The full weekday name ("Sunday") |
+| a | The abbreviated name ("Sun") |
+| u | Day of the week (Monday is 1, 1..7) |
+| w | Day of the week (Sunday is 0, 0..6) |
+
+**ISO 8601 week-based year and week number:**
+
+The first week of YYYY starts with a Monday and includes YYYY-01-04.
+The days in the year before the first week are in the last week of
+the previous year.
+
+| Format | Meaning |
+| ------ | ------- |
+| G | The week-based year |
+| g | The last 2 digits of the week-based year (00..99) |
+| V | Week number of the week-based year (01..53) |
+
+**Week number:**
+
+The first week of YYYY that starts with a Sunday or Monday (according to %U
+or %W). The days in the year before the first week are in week 0.
+
+| Format | Meaning |
+| ------ | ------- |
+| U | Week number of the year. The week starts with Sunday. (00..53) |
+| W | Week number of the year. The week starts with Monday. (00..53) |
+
+**Seconds since the Epoch:**
+
+| Format | Meaning |
+| s | Number of seconds since 1970-01-01 00:00:00 UTC. |
+
+**Literal string:**
+
+| Format | Meaning |
+| ------ | ------- |
+| n | Newline character (\n) |
+| t | Tab character (\t) |
+| % | Literal "%" character |
+
+**Combination:**
+
+| Format | Meaning |
+| ------ | ------- |
+| c | date and time (%a %b %e %T %Y) |
+| D | Date (%m/%d/%y) |
+| F | The ISO 8601 date format (%Y-%m-%d) |
+| v | VMS date (%e-%^b-%4Y) |
+| x | Same as %D |
+| X | Same as %T |
+| r | 12-hour time (%I:%M:%S %p) |
+| R | 24-hour time (%H:%M) |
+| T | 24-hour time (%H:%M:%S) |
+
+**Example**: Using `strftime` with a `Timestamp`:
+
+~~~ puppet
+$duration = Timestamp('2016-08-24T12:13:14')
+
+# Notice the duration using a format that notices <hours>:<minutes>:<seconds>
+notice($duration.strftime('%F')) # outputs '2016-08-24'
+
+# Notice the duration using a format that notices <minutes>:<seconds>
+notice($duration.strftime('%c')) # outputs 'Wed Aug 24 12:13:14 2016'
+~~~
+
+### Format directives applicable to `Timespan`:
+
+| Format | Meaning |
+| ------ | ------- |
+| D | Number of Days |
+| H | Hour of the day, 24-hour clock |
+| M | Minute of the hour (00..59) |
+| S | Second of the minute (00..59) |
+| L | Millisecond of the second (000..999). Digits under millisecond are truncated to not produce 1000. |
+| N | Fractional seconds digits, default is 9 digits (nanosecond). Digits under a specified length are truncated to avoid carry up |
+
+The format directive that represents the highest magnitude in the format will be allowed to overflow.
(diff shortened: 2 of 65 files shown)


lib/puppet/functions/strftime.rb
```diff
@@ -0,0 +1,14 @@
+# (Documentation in 3.x stub)
+#
+# @since 4.7.0
+#
+Puppet::Functions.create_function(:strftime) do
+  dispatch :format do
+    param 'Variant[Timestamp,Timespan]', :time_object
+    param 'String', :format
```
hlindberg: no default format?

lib/puppet/parser/functions/strftime.rb
```diff
@@ -0,0 +1,160 @@
...
+| Flag  | Meaning
+| ----  | ---------------
+| #     | Change case
+| ^     | Use uppercase
+| :     | Use colons for %z
+
+### Format directives applicable to `Timestamp` (names and padding can be altered using flags):
+
+**Date (Year, Month, Day):**
+| Format | Meaning |
```
hlindberg: add blank line above

lib/puppet/parser/functions/strftime.rb
```diff
@@ -0,0 +1,160 @@
...
+| m | Month of the year, zero-padded (01..12) |
+| B | The full month name ("January") |
+| b | The abbreviated month name ("Jan") |
+| h | Equivalent to %b |
+| d | Day of the month, zero-padded (01..31) |
+| e | Day of the month, blank-padded ( 1..31) |
+| j | Day of the year (001..366) |
+
+**Time (Hour, Minute, Second, Subsecond):**
+| Format | Meaning |
```
hlindberg: Add blank line above (between heading and table)

lib/puppet/parser/functions/strftime.rb
```diff
@@ -0,0 +1,160 @@
...
+| l | Hour of the day, 12-hour clock, blank-padded ( 1..12) |
+| P | Meridian indicator, lowercase ("am" or "pm") |
+| p | Meridian indicator, uppercase ("AM" or "PM") |
+| M | Minute of the hour (00..59) |
+| S | Second of the minute (00..60) |
+| L | Millisecond of the second (000..999). Digits under millisecond are truncated to not produce 1000 |
+| N | Fractional seconds digits, default is 9 digits (nanosecond). Digits under a specified width are truncated to avoid carry up |
+
+**Time (Hour, Minute, Second, Subsecond):**
+| Format | Meaning |
```
hlindberg: blank line

lib/puppet/parser/functions/strftime.rb
```diff
@@ -0,0 +1,160 @@
...
+| ------ | ------- |
+| z   | Time zone as hour and minute offset from UTC (e.g. +0900) |
+| :z  | hour and minute offset from UTC with a colon (e.g. +09:00) |
+| ::z | hour, minute and second offset from UTC (e.g. +09:00:00) |
+| Z   | Abbreviated time zone name or similar information.  (OS dependent) |
+
+**Weekday:**
+| Format | Meaning |
```
hlindberg: blank line

lib/puppet/parser/functions/strftime.rb
```diff
@@ -0,0 +1,160 @@
...
+| ------ | ------- |
+| A | The full weekday name ("Sunday") |
+| a | The abbreviated name ("Sun") |
+| u | Day of the week (Monday is 1, 1..7) |
+| w | Day of the week (Sunday is 0, 0..6) |
+
+**ISO 8601 week-based year and week number:**
+The first week of YYYY starts with a Monday and includes YYYY-01-04.
```
hlindberg: Add blank line between heading and text

lib/puppet/parser/functions/strftime.rb
```diff
@@ -0,0 +1,160 @@
...
+The days in the year before the first week are in the last week of
+the previous year.
+| Format | Meaning |
```
hlindberg: blank line before table

lib/puppet/parser/functions/strftime.rb
```diff
@@ -0,0 +1,160 @@
...
+| ------ | ------- |
+| G | The week-based year |
+| g | The last 2 digits of the week-based year (00..99) |
+| V | Week number of the week-based year (01..53) |
+
+**Week number:**
+The first week of YYYY that starts with a Sunday or Monday (according to %U
```
hlindberg: blank line

lib/puppet/parser/functions/strftime.rb
```diff
@@ -0,0 +1,160 @@
...
+notice($duration.strftime('%H:%M:%S')) # outputs '03:20:30'
+
+# Notice the duration using a format that outputs <minutes>:<seconds>
+notice($duration.strftime('%M:%S')) # outputs '200:30'
+~~~
+
+- Since 4.7.0
+DOC
+) do |args|
+  function_fail(["strftime() is only available when parser/evaluator future is in effect"])
```
hlindberg: We used this error messages in other 3.x functions but the message is odd given that there is only the future parser. The only way it could have been called this way is if it was loaded the 3.x way and then called via Functions, or via Scope#function_strftime.  Suggest we change this (and in all other functions using the same wording) to say "xxx() can only be called using the 4.x function API", and if we want to be helpful add "See Scope#call_function".

lib/puppet/pops/time/timespan.rb
```diff
@@ -0,0 +1,686 @@
...
+  KEY_DAYS = 'days'.freeze
+  KEY_HOURS = 'hours'.freeze
+  KEY_MINUTES = 'minutes'.freeze
+  KEY_SECONDS = 'seconds'.freeze
+  KEY_MILLI_SECONDS = 'milli_seconds'.freeze
+  KEY_MICRO_SECONDS = 'micro_seconds'.freeze
+  KEY_NANO_SECONDS = 'nano_seconds'.freeze
+
+  # TimeData is a Numeric that stores its value internally as nano-seconds but will be considered to be seconds and fractions of
+  # seconds when used in arithmetic or comparison with other Numeric types, .
```
hlindberg: ends with `, .` - is something missing?

lib/puppet/functions/strftime.rb
thallgren: A `String(x)' will give you the default. Didn't see any point in adding it here. Optional arguments have the trade-off of hiding forgotten ones.

lib/puppet/pops/time/timespan.rb
```diff
@@ -0,0 +1,686 @@
...
+      when self.class
+        @nsecs <=> o.nsecs
+      when Integer
+        to_int <=> o
+      when Float
+        to_f <=> o
+      else
+        nil
+      end
+    end
```
hlindberg: This method suggests that time is comparable to values. It is also possible to perform arithmetic. That suggests that it should be possible to compare time vs numeric in puppet as well.

lib/puppet/parser/functions/strftime.rb
thallgren: What markdown editor are you using? I'm validating everything by pasting the text here https://jbt.github.io/markdown-editor/. It shows up just fine without a blank line.

lib/puppet/pops/time/timespan.rb
```diff
@@ -0,0 +1,686 @@
...
+        Timespan.new(@nsecs - o.nsecs)
+      when Integer, Float
+        # Subtract seconds
+        Timespan.new(@nsecs - (o * NSECS_PER_SEC).to_i)
+      else
+        raise ArgumentError, "#{a_an_uc(o)} cannot be subtracted from a Timespan"
+      end
+    end
+
+    def -@
```
hlindberg: ?? what is this for

spec/unit/pops/evaluator/arithmetic_ops_spec.rb
```diff
@@ -73,5 +74,139 @@
...
+      it 'Timespan * Timestamp is an error' do
+        code = 'notice(Timespan({days => 3}) * Timestamp())'
+        expect { eval_and_collect_notices(code) }.to raise_error(Puppet::Error, /A Timestamp cannot be multiplied by a Timestamp/)
+      end
+
+      it 'Timespan / Timestamp is an error' do
+        code = 'notice(Timespan({days => 3}) / Timestamp())'
+        expect { eval_and_collect_notices(code) }.to raise_error(Puppet::Error, /A Timespan cannot be divided by a Timestamp/)
+      end
+    end
```
hlindberg: Tests for 2nd numeric operand missing - Timespan [+ - * /] Numeric

lib/puppet/pops/time/timespan.rb
thallgren: Unary minus

lib/puppet/pops/time/timespan.rb
thallgren: Yes. A Timespan or Timestamp can be compared to an number that represents seconds (or seconds since epoch) in Puppet.

spec/unit/pops/time/timespan_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    it 'using %N as the biggest quantity' do
+      expect(Timespan.parse('112211123456789', '%N')).to eql(complex)
+    end
+
+    it 'where biggest quantity is not frist' do
+      expect(Timespan.parse('11:1630', '%S:%M')).to eql(simple)
+    end
+  end
```
hlindberg: A common way to express a timespan in puppet is to use something like 3d12h14m48s, where each higher value format is optional. Can you use such a format or are all parts required if you specify something like ´%Dd%Hh...`?

spec/unit/pops/types/p_timespan_type_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+            notice($o1 == $o2)
+            notice($o1 != $o2)
+        CODE
+        expect(eval_and_collect_notices(code)).to eq(%w(false false true true false true))
+      end
+    end
+  end
+end
+end
+end
```
hlindberg: comparisons against numeric missing

spec/unit/pops/types/p_timestamp_type_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+            notice($o1 == $o2)
+            notice($o1 != $o2)
+        CODE
+        expect(eval_and_collect_notices(code)).to eq(%w(false false true true false true))
+      end
+    end
+  end
+end
+end
+end
```
hlindberg: comparison against numeric missing

spec/unit/pops/time/timespan_spec.rb
thallgren: All parts are required but it's possible to use an array of formats. First one that succeeds wins.

lib/puppet/pops/evaluator/compare_operator.rb
```diff
@@ -92,13 +92,23 @@ def cmp_Symbol(a, b)
     end
   end
 
+  def cmp_Timespan(a, b)
+    raise ArgumentError.new('Timespans not comparable to non Timespans') unless b.is_a?(Time::Timespan)
```
hlindberg: Elsewhere it is stated that timespans can be compared against Numeric

lib/puppet/pops/evaluator/compare_operator.rb
```diff
@@ -92,13 +92,23 @@ def cmp_Symbol(a, b)
...
+    a <=> b
+  end
+
+  def cmp_Timestamp(a, b)
+    raise ArgumentError.new('Timestamps not comparable to non Timestamps') unless b.is_a?(Time::Timestamp)
```
hlindberg: Elsewhere it is stated that timestamps can be compared against Numeric

spec/unit/pops/evaluator/arithmetic_ops_spec.rb
thallgren: I will remove these tests. Arithmetic is fully covered by the "when used in Puppet expressions" section in the tests for the respective types.

thallgren: No, they are not missing. See line 111 - 124.

thallgren:
PR updated. I chose to squash some of the changes since I rewrote parts of the code dealing with the common type. A new (maint) commit takes care of unifying the fail message when 3x functions are overridden but called anyway. All above comments should be addressed.

spec/unit/pops/time/timespan_spec.rb
hlindberg: will see if there is a test that shows that...

lib/puppet/parser/functions/fail_is4x.rb
```diff
@@ -0,0 +1,3 @@
+Puppet::Parser::Functions::newfunction(:fail_is4x, :arity => 1, :doc => 'Fail with a parse error indicating that 4x function API should be used.') do |args|
+    raise Puppet::ParseError, "#{args[0]}() can only be called using the 4.x function API. See Scope#call_function"
+end
```
hlindberg: Not a very good idea (sorry) - this function ends up as something users can call, it shows up in documentation and what not. Remove this.
If you feel that it is bad to copy/paste the error message, create an error class instead that takes the function name as argument.

spec/unit/pops/time/timespan_spec.rb
thallgren: Yes, you missed it. Check p_timespan_type_spec.rb, line 63 and p_timestamp_type_spec.rb line 64.

thallgren:
I replaced the maint commit with a new solution, implemented according to your proposal.

PR #5236, opened 2016-08-29
(PUP-6103) Make epp render use facts from localhost

hlindberg:
This makes epp render use facts from the local node by default, or from a given node with `--node`.
Facts given on the command line (via a reference to a file .json  or .yaml) is merged on top of the node's facts.

### DIFF
--- a/lib/puppet/face/epp.rb
+++ b/lib/puppet/face/epp.rb
@@ -62,6 +62,8 @@ Puppet::Face.define(:epp, '0.0.1') do
     EOT
     when_invoked do |*args|
       options = args.pop
+      # pass a dummy node, as facts are not needed for validation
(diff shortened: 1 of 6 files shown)


lib/puppet/face/epp.rb
```diff
@@ -291,8 +303,8 @@
       summary "A .pp or .yaml file that is processed to produce a hash of values for the template."
     end
 
-    option("--facts <yaml_file>") do
-      summary "A .yaml file containing a hash of facts made available in $facts"
+    option("--facts <yaml_file>[<json_file>]") do
```
thallgren: The option argument description looks a bit off. Shouldn't it be `"--facts [<yaml_file>|<json_file>]"` ? Or does the current syntax have specific meaning for the options parser?

hlindberg: will change that to say `facts_file` - not sure what happened there :-)

PR #5237, opened 2016-08-29
(PUP-6659) Remove Stacktraces in API responses

justinstoller:
Previously, Internal Server Errors would return an API response that
included a stacktrace.

This is not considered a best practice and on non-standard setups
filesystem locations disclosed in a stacktrace could be considered
sensitive information.

This commit moves to logging the stacktrace at the error level while
returning only the exception message via the HTTP API.

### DIFF
--- a/lib/puppet/network/http/handler.rb
+++ b/lib/puppet/network/http/handler.rb
@@ -68,7 +68,8 @@ module Puppet::Network::HTTP::Handler
     new_response.respond_with(e.status, "application/json", e.to_json)
   rescue StandardError => e
     http_e = Puppet::Network::HTTP::Error::HTTPServerError.new(e)
-    Puppet.err(http_e.message)
+    log_msg = [http_e.message, *http_e.backtrace].join("\n")
+    Puppet.err(log_msg)
     new_response.respond_with(http_e.status, "application/json", http_e.to_json)
   ensure
     if profiler
--- a/spec/unit/network/http/handler_spec.rb
+++ b/spec/unit/network/http/handler_spec.rb
@@ -81,6 +81,7 @@ describe Puppet::Network::HTTP::Handler do
       handler = PuppetSpec::Handler.new(
         Puppet::Network::HTTP::Route.path(/.*/).get(lambda { |_, _| raise error}))
 
+      # Stacktraces should be included in logs
       Puppet.expects(:err).with("Server Error: the sky is falling!\na.rb\nb.rb")
 
       req = a_request("GET", "/vtest/foo")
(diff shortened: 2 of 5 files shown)


haus:
:+1: seems lovely

spec/unit/network/http/handler_spec.rb
```diff
@@ -88,8 +93,7 @@ def respond(text)
       expect(res[:content_type_header]).to eq("application/json")
       expect(res_body["issue_kind"]).to eq(Puppet::Network::HTTP::Issues::RUNTIME_ERROR.to_s)
       expect(res_body["message"]).to eq("Server Error: the sky is falling!")
-      expect(res_body["stacktrace"].is_a?(Array) && !res_body["stacktrace"].empty?).to be_truthy
-      expect(res_body["stacktrace"][0]).to match("spec/unit/network/http/handler_spec.rb")
+      expect(res_body["stacktrace"]).to be_nil
```
KevinCorcoran: This line would probably look weird without the context of this ticket, probably worth adding a comment / link to ticket.

lib/puppet/network/http/handler.rb
```diff
@@ -68,7 +68,8 @@ def process(request, response)
     new_response.respond_with(e.status, "application/json", e.to_json)
   rescue StandardError => e
     http_e = Puppet::Network::HTTP::Error::HTTPServerError.new(e)
-    Puppet.err(http_e.message)
+    log_msg = [http_e.message, *http_e.backtrace].join("\n")
+    Puppet.err(log_msg)
```
KevinCorcoran: Is it possible to simply log `e`?  i.e. `Puppet.err(e)`.

Also, not related to the change in this PR, but should that `rescue Exception` instead?

KevinCorcoran:
🙈  looks fine except for the few comments I left

lib/puppet/network/http/handler.rb
justinstoller: `rescue Exception` isn't encouraged by the Ruby maintainers. Kinda like catching `Throwable` in Java.

justinstoller: I'll double check, but I'm pretty sure that `Puppet.err(e)` would only log the `message` field.

justinstoller: In the common case we will just call `to_s` on whatever is passed into `Puppet.err`, which for exceptions will return the `message` field (though what exactly we do depends on our log destination).

KevinCorcoran:
👍

haus:
ron-swanson-is-excited

PR #5240, opened 2016-08-29
(PUP-6604) Extend variables_refreshed sleep on Solaris

johnduarte:
This commit extends the `sleep` statement in the
`variables_refreshed_each_compilation` test to 61 seconds when
the platform is Solaris.

As per FACT-1497, the `system_uptime.seconds` fact updates
less frequently on Solaris. Prior to this change this test failed
on Solaris due to this issue with facter.

This is a work around and should be reverted when the underlying
facter issue is resolved.


er0ck:
LGTM

PR #5243, opened 2016-08-30
(PUP-6513) Log lookup explainer info at debug level

thallgren:
This PR ensures that a lookup explainer is present when
`Puppet[:debug]` is `true` and that the information that it collects
is logged. For cases when `Puppet[:debug]` is `true` and an explainer is
present already (as when using the lookup command with --explain or
--explain-options), then the information from that explainer is mirrored
to the debug logger.

### DIFF
--- a/lib/puppet/pops/lookup.rb
+++ b/lib/puppet/pops/lookup.rb
@@ -54,10 +54,22 @@ module Lookup
       elsif has_default
         answer = assert_type('Default value', value_type, default_value)
       else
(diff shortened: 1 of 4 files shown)


hlindberg:
What happens if the CLI `lookup` is used with `--explain` and `--debug` at the same time?

thallgren:
Then the info from the explainer is mirrored to the debug logger.

hlindberg:
Want tests that shows: "Does not log explanation when log level is < 'debug'" as a guard against future regressions.

hlindberg:
Good to merge when travis goes green.

PR #5245, opened 2016-08-30
(PUP-6275) Log profiler output at INFO level

Sharpie:
Profiler output requires `puppet agent` or `puppet apply` to be run with
the `--profile` flag or for `profiler=true` to be set in puppet.conf.
Prior to this patch, profile output was logged at DEBUG level which meant
that the global log level for Puppet Server had to be raised, which required
hand-editing an XML file along with a  service restart before and after
profiling activities.

This patch raises Puppet Server profile output to the default INFO level since
an additional option is already required to enable the profiler. This change
allows on-demand profiling via `puppet agent -t --profile` without the need to
manually re-start the Puppet Server and edit configuration.


cprice404:
+1

PR #5246, opened 2016-08-30
(maint) Fix problem with arithmetic involving TimeData on right side

thallgren:
This commit fixes problems that occurs when a Timestamp or Timespan
is on the right hand side of an arithmetic operation and the left
hand side is an Integer or Float. The result of an addition,
subtraction, or multiplication was of the wrong type and some
nonsensical operations that really should be prohibited, were allowed.

A side effect of this fix is an improvement of the error message
when using '%' with a Float on the right hand side. The expression

1 % 1.0

would, before this commit, result in the message:

Operator '%' is not applicable to an Integer

now, the message will instead be:

Operator '%' is not applicable to an Integer when right side is a Float

### DIFF
--- a/lib/puppet/pops/time/timestamp.rb
+++ b/lib/puppet/pops/time/timestamp.rb
@@ -33,9 +33,12 @@ class Timestamp < TimeData
     end
   end
 
-  undef_method :-@, :+@, :%,:modulo,:divmod, :div, :fdiv, :abs, :abs2, :magnitude # does not make sense on a Timestamp
+  undef_method :-@, :+@, :div, :fdiv, :abs, :abs2, :magnitude # does not make sense on a Timestamp
   if method_defined?(:negative?)
     undef_method :negative?, :positive?
+  end
+  if method_defined?(:%)
+    undef_method :%, :modulo, :divmod
   end
 
   def +(o)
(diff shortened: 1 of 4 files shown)


lib/puppet/pops/time/timestamp.rb
```diff
@@ -33,10 +33,13 @@ def self.parse(str, format = DEFAULT_FORMATS)
     end
   end
 
-  undef_method :-@, :+@, :%,:modulo,:divmod, :div, :fdiv, :abs, :abs2, :magnitude # does not make sense on a Timestamp
+  undef_method :-@, :+@, :div, :fdiv, :abs, :abs2, :magnitude # does not make sense on a Timestamp
   if method_defined?(:negative?)
     undef_method :negative?, :positive?
   end
+  if method_defined?(:%)
+    undef_method :%, :modulo, :divmod
```
pcarlisle: whether `modulo` and `divmod` are defined doesn't seem tied to `%`. My jruby Numeric has both of them but not `%`.

thallgren: Ok. I thought, better safe than sorry. It doesn't matter, really. We never
expose those methods.

On Tue, Aug 30, 2016 at 10:13 PM, Patrick Carlisle <notifications@github.com

PR #5247, opened 2016-08-30
(PUP-6241) Acceptance unreadable environment

johnduarte:
This commit adds an acceptance test to ensure that executing a
`puppet agent` run against an unreadable environment does not
succeed without an error.

### DIFF
--- a/acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
+++ b/acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
@@ -1,5 +1,5 @@
 test_name "C97899 - Agent run should fail if environment is unreadable" do
-  skip_test 'requires a master for managing the environment' if hosts_with_role(hosts, 'master').length == 0
+  skip_test 'requires a puppetserver master for managing the environment' if hosts_with_role(hosts, 'master').length == 0 or not @options[:is_puppetserver]
 
   testdir = ''
   env_path = ''
--- a/acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
+++ b/acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
@@ -1,0 +1,58 @@
+test_name "C97899 - Agent run should fail if environment is unreadable" do
+  skip_test 'requires a master for managing the environment' if hosts_with_role(hosts, 'master').length == 0
+
+  testdir = ''
+  env_path = ''
+  test_env = ''
+
+  step 'setup environments' do
+    testdir = create_tmpdir_for_user master, 'c97899_unreadable_envdir'
+    env_path = "#{testdir}/environments"
+    test_env = "#{env_path}/testing"
+
+    apply_manifest_on(master, <<-MANIFEST, :catch_failures => true)
+      File {
+        ensure => directory,
+        mode => "0770",
+        owner => #{master.puppet['user']},
+        group => #{master.puppet['group']},
+      }
+      file {
+        '#{env_path}':;
+        '#{env_path}/production':;
+        '#{test_env}':;
+        '#{test_env}/manifests':;
+        '#{test_env}/modules':;
+      }
+      file { '#{test_env}/manifests/site.pp':
+        ensure => file,
+        mode => "0640",
+        content => 'node default { notify { "Hello agent": } }',
+      }
+    MANIFEST
+  end
+
+  step 'change permissions of envdir to 644' do
+    on(master, "chmod 644 #{test_env}")
+  end
+
+  step 'verify environment fails with puppet agent run' do
+    master_opts = {
+      'main' => {
+        'environmentpath' => env_path,
+      }
+    }
+    with_puppet_running_on master, master_opts, testdir do
+      agents.each do |agent|
+        on(agent, puppet("agent --test --server #{master} --trace --environment testing"), :accept_all_exit_codes => true) do |result|
+          refute_equal(2, result.exit_code, 'agent run should not apply changes')
+          expect_failure('expected to fail until PUP-6241 is resolved') do
+            refute_equal(0, result.exit_code, 'agent run should not succeed')
+            refute_empty(result.stderr, 'an appropriate error is expected')
+          end
+        end
+      end
+    end
+  end
+
+end
--- a/acceptance/tests/environment/negative/agent_run_should_fail_if_site_pp_unreadable.rb
+++ b/acceptance/tests/environment/negative/agent_run_should_fail_if_site_pp_unreadable.rb
@@ -1,4 +1,4 @@
-test_name "C97899 - Agent run should fail if environment is unreadable" do
+test_name "C98160 - Agent run should fail if an environment's site.pp is unreadable" do
   skip_test 'requires a master for managing the environment' if hosts_with_role(hosts, 'master').length == 0
 
   testdir = ''
@@ -26,14 +26,10 @@ test_name "C97899 - Agent run should fail if environment is unreadable" do
       }
       file { '#{test_env}/manifests/site.pp':
         ensure => file,
-        mode => "0640",
+        mode => "0000",  # <- NOTE: this is unreadable
         content => 'node default { notify { "Hello agent": } }',
       }
     MANIFEST
-  end
-
-  step 'change permissions of envdir to 644' do
-    on(master, "chmod 644 #{test_env}")
   end
 
   step 'verify environment fails with puppet agent run' do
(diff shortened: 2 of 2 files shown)


acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
```diff
@@ -0,0 +1,57 @@
...
+      file { '#{test_env}/manifests/site.pp':
+        ensure => file,
+        mode => "0640",
+        content => 'node default { notify { "Hello agent": } }',
+      }
+    MANIFEST
+  end
+
+  step 'change permissions of envdir to 644' do
+    on(master, "chmod 644 #{test_env}")
```
james-stocks: There's two possible cases here: the environment folder has overly restrictive permissions; and the environment folder manifest has correct permissions but site.pp has overly restrictive permissions.
This code seems to restrict both permissions but will only really be testing the folder permissions. Perhaps this test file (or a separate test file) should test the case of the folder having correct permissions and site.pp having restrictive permissions?

(I imagine we could also have cases where both the folder and site.pp have correct permissions but then other manifest files/folders are restricted - I think that should be a separate ticket though)

acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  step 'verify environment fails with puppet agent run' do
+    master_opts = {
+      'main' => {
+        'environmentpath' => env_path,
+      }
+    }
+    with_puppet_running_on master, master_opts, testdir do
```
thallgren: It seems like this starts a 'puppet master' and not a puppet server. The problem is only reproduced using a puppet server.

johnduarte: If run with the `ci:test:aio` rake task, the `with_puppet_running_on` uses the deployed puppetserver rather than a passenger based master. We could throttle the test to only run in the context of a `type == aio` run. If we want this test to be added to the regression suite, I would recommend that we run it without such a restriction.

acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
johnduarte: @james-stocks , I am unable to reproduce the case where the `site.pp` file is too restrictive. I have added another test case to this PR for that scenario. It currently, fails the `expect_failure` condition when I run it against a recent master puppet-agent (SHA=e15c8b567fee6846f06aa8f0e74a07dd8fe90441).

acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
joshcooper: The `master` in `with_puppet_running_on master` is actually the host with the `master` role defined in the beaker host configuration. It's confusing because beaker also supports methods like `on(host, puppet("master ..."))` which really does mean run the webrick master... 

Also, `ci:test:aio` always uses `puppetserver`, but `ci:test:git` uses `webrick`, and we no longer support or test passenger packages. Since the failure only occurs under `puppetserver`, I think that means the overall test will fail when run in `ci:test:git`, e.g. due to the `expect_failure` below? If that's true, then I think we should confine to `aio` only.

MikaelSmith:
@johnduarte is this fixed up now?

johnduarte:
@MikaelSmith this needs to be tested with `ci:test:git` prior to merge.

johnduarte:
@MikaelSmith these tests have been amended to skip when there is no master role defined in the given hosts config in order to be compliant with `ci:test:git`. This PR is now ready for merge. Tests passing against current master SHA.
```
SHA=3db0f62a0ca40768965f9b8f75f0d70560b0047e SUITE_VERSION=1.9.3.193.g3db0f62 TESTS=tests/environment/negative be rake ci:test:aio
...
       Total Suite Time: 57.68 seconds
      Average Test Time: 28.84 seconds
              Attempted: 2
                 Passed: 2
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 2
```

acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
```diff
@@ -0,0 +1,58 @@
...
+      agents.each do |agent|
+        on(agent, puppet("agent --test --server #{master} --trace --environment testing"), :accept_all_exit_codes => true) do |result|
```
er0ck: is `--trace` required here?  it might slow down this test.  same with test below.  we might be able to optimize these tests even further by combining them, and not using puppet to setup the environment.

acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
```diff
@@ -0,0 +1,58 @@
+test_name "C97899 - Agent run should fail if environment is unreadable" do
+  skip_test 'requires a master for managing the environment' if hosts_with_role(hosts, 'master').length == 0
+
+  testdir = ''
+  env_path = ''
+  test_env = ''
```
joshcooper: nit, it's not necessary to set these to `''` since we set them on lines 9-11 below, and there isn't conditional logic that could cause them to not be set. Not blocking this PR, just general comment.

acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
joshcooper: @johnduarte can you confirm if the test still passes with `ci:test:git`?

acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
er0ck: i think `step` has its own scope, that's probably why they need instancing here.  then its value is used across two different steps

johnduarte: That is correct. The step blocks have there own scope.

acceptance/tests/environment/negative/agent_run_should_fail_if_env_unreadable.rb
johnduarte: @joshcooper it looks like one of the pre-suite tasks for `ci:test:git` is currently broken on master. So, I am not able to successfully test that task.
```
Errored Tests Cases:
  Test Case setup/git/pre-suite/000_EnvSetup.rb reported: #<Beaker::Host::CommandFailure: Host 'ychcq6ueup73370.delivery.puppetlabs.net' exited with 1 running:
 cd /opt/puppet-git-repos/puppet-win32-ruby; cp -r ruby/* /
Last 10 lines of output were:
        bash: line 0: cd: /opt/puppet-git-repos/puppet-win32-ruby: No such file or directory
        cp: cannot stat 'ruby/*': No such file or directory>
    Test line: setup/git/pre-suite/000_EnvSetup.rb:119:in `block (2 levels) in run_test'
```
Filed [PUP-7376](https://tickets.puppetlabs.com/browse/PUP-7376)

johnduarte: fyi: [PUP-7376](https://tickets.puppetlabs.com/browse/PUP-7376) is a dupe of [PUP-7009](https://tickets.puppetlabs.com/browse/PUP-7009)

er0ck:
i think i'm 👍, as-is on this.  there are optimizations that are available.  but given many on-going initiatives to make our testing more efficient, with as-of-yet unknown outcomes, i think we should just get this in to ensure coverage.

joshcooper:
I'm reviewing/going to merge the `ci:test:git` change, let finish that, then I'll review/merge this.

johnduarte:
@joshcooper PUP-7009 has been resolved.

I can confirm that this still passes `ci:test:aio`
```
SHA=466d924fc94fc790277b8370eff359d4cc62357f TESTS=tests/environment/negative be rake ci:test:aio
...
      Test Suite: tests @ 2017-05-26 15:51:32 -0700

       Total Suite Time: 54.98 seconds
      Average Test Time: 27.49 seconds
              Attempted: 2
                 Passed: 2
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 2
```
... and with `ci:test:git`
```
SHA=7bee5bc TEST_TARGET=windows2012r2-64a TESTS=tests/environment/negative bundle exec rake ci:test:git
...
      Test Suite: tests @ 2017-05-26 16:42:14 -0700

       Total Suite Time: 21.61 seconds
      Average Test Time: 10.81 seconds
              Attempted: 2
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 1
                Pending: 0
                  Total: 2
```

I believe this can be safely merged.

johnduarte:
Updating with rebase on current master.

Confirming that this passes using `ci:test:aio`
```
SHA=5.0.0 TESTS=tests/environment/negative be rake ci:test:aio
...
      Test Suite: tests @ 2017-07-14 03:15:19 -0700

       Total Suite Time: 73.86 seconds
      Average Test Time: 36.93 seconds
              Attempted: 2
                 Passed: 2
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 2
```

and with `ci:test:git`
```
SHA=5.0.0 TESTS=tests/environment/negative be rake ci:test:git
...
      Test Suite: tests @ 2017-07-14 04:30:07 -0700

       Total Suite Time: 32.70 seconds
      Average Test Time: 16.35 seconds
              Attempted: 2
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 1
                Pending: 0
                  Total: 2
```

er0ck:
merging.  let's get this thing in there and we'll fixup some tech-debt upon acceptance audit.

PR #5248, opened 2016-08-30
(PUP-6662) Make Sensitive data work in user defined resources

hlindberg:
Before this, a Sensitive type value could only be used directly
in a native resource.

In order to fix this, there are some hoops to jump through.
* a puppet apply does not serialize the resources in the catalog which
means that the processing of sensitive data must take place in the
resources themselves.
* The early unwrapping and marking parameters as sensitive then occur
too early (finish) for classes and user defined types. For native types
there is no evaluation, and finish is only called post compile, but for
user defined resources / classes this occurs when the resource is
evaluated.

### DIFF
--- a/lib/puppet/parser/resource.rb
+++ b/lib/puppet/parser/resource.rb
@@ -77,7 +77,7 @@ class Puppet::Parser::Resource < Puppet::Resource
       elsif resource_type.nil?
         self.fail "Cannot find definition #{type}"
       else
-        finish(false) # Call finish but do not validate
+        finish_evaluation() # do not finish completely (as that destroys Sensitive data)
         resource_type.evaluate_code(self)
       end
     end
@@ -94,8 +94,18 @@ class Puppet::Parser::Resource < Puppet::Resource
     end
   end
 
-  # Do any finishing work on this object, called before evaluation or
-  # before storage/translation. The method does nothing the second time
+  # Finish the evaluation by assigning defaults and scope tags
+  # @api private
+  #
+  def finish_evaluation
+    return if @evaluation_finished
+    add_defaults
+    add_scope_tags
+    @evaluation_finished = true
+  end
+
+  # Do any finishing work on this object, called before
+  # storage/translation. The method does nothing the second time
   # it is called on the same resource.
   #
   # @param do_validate [Boolean] true if validation should be performed
@@ -104,8 +114,7 @@ class Puppet::Parser::Resource < Puppet::Resource
   def finish(do_validate = true)
     return if finished?
     @finished = true
-    add_defaults
-    add_scope_tags
+    finish_evaluation
     replace_sensitive_data
     validate if do_validate
   end
--- a/lib/puppet/resource/type.rb
+++ b/lib/puppet/resource/type.rb
@@ -427,7 +427,15 @@ class Puppet::Resource::Type
   # Validate that all parameters given to the resource are correct
   # @param resource [Puppet::Resource] the resource to validate
   def validate_resource(resource)
-    validate_resource_hash(resource, Hash[resource.parameters.map { |name, value| [name.to_s, value.value] }])
+    # Since Sensitive values have special encoding (in a separate parameter) an unwrapped sensitive value must be
+    # recreated as a Sensitive in order to perform correct type checking.
+    sensitives = Set.new(resource.sensitive_parameters)
+    validate_resource_hash(resource,
+      Hash[resource.parameters.map do |name, value|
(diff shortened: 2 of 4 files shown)


lib/puppet/resource/type.rb
```diff
@@ -427,7 +427,15 @@ def validate_resource_hash(resource, resource_hash)
   # Validate that all parameters given to the resource are correct
   # @param resource [Puppet::Resource] the resource to validate
   def validate_resource(resource)
-    validate_resource_hash(resource, Hash[resource.parameters.map { |name, value| [name.to_s, value.value] }])
+    # Since Sensitive values have special encoding (in a separate parameter) an unwrapped sensitive value must be
+    # recreated as a Sensitive in order to perform correct type checking.
+    sensitives = Set.new(resource.sensitive_parameters)
+    validate_resource_hash(resource,
+      Hash[resource.parameters.map do |name, value|
+        value_to_validate = sensitives.include?(name) ? Puppet::Pops::Types::PSensitiveType::Sensitive.new(value.value) : value.value
```
hlindberg: @branan This is the insane part. The Sensitive values have at this point been unwrapped and marked in `sensitive_parameters`. The type signature for the resource type may however have typed a parameter with `Sensitive`. To make it work an instance of `Sensitive` must be used when checking the type signature - here done as a Hash of values checked against a Struct data type.

lib/puppet/parser/resource.rb
```diff
@@ -94,8 +94,18 @@ def exported=(value)
...
-  # Do any finishing work on this object, called before evaluation or
-  # before storage/translation. The method does nothing the second time
+  # Finish the evaluation by assigning defaults and scope tags
+  # @api private
+  #
+  def finish_evaluation
+    return if @evaluation_finished
+    add_defaults
+    add_scope_tags
+    @evaluation_finished = true
```
branan: `finish` sets its guard flag early - presumably to make sure it gets set even in the case of an exception. Should the same thing be done here?

branan: Although I think the most likely step to raise an exception is `validate`, which is not part of the `finish_evaluation` scope

lib/puppet/resource/type.rb
branan: Is there any chance of this creating a `Sensitive` type agent-side when using the client/server model? It looks like not to me (`validate_resource` is only called from `parser/resource.rb`)

lib/puppet/type/component.rb
```diff
@@ -71,6 +71,10 @@ def to_s
     reference.to_s
   end
 
+  # Overrides the default implementation to do nothing
```
branan: Would be nice to have a bit more description about what's special about component. Rough draft:

```
# This type is used to represent a Class or Define before the catalog is converted from the containment graph to the relationship graph. As such, it contains data (its class/define parameters), but does not have actual parameters or properties at the Type level. We can simply ignore anything flagged as sensitive here, since the contained resources will handle that sensitivity themselves. There is no risk of this information leaking into reports, since no Component instances survive the graph transmutation.
```
```

lib/puppet/parser/resource.rb
hlindberg: I could change that, but it would then be lying while add_defaults and add_scope_tags are being evaluated - if one of those fails, it would fail the entire compilation.

branan: > if one of those fails, it would fail the entire compilation.

That's good enough for me. I'm fine with this as-is

lib/puppet/type/component.rb
hlindberg: ok, accepted - will add that

lib/puppet/resource/type.rb
hlindberg: This would be harmless on the agent side (if it is some day called there as it creates a temporary hash to use when asserting the type). Other uses of Sensitive type (or no use) is unaffected by this.

nicklewis:
@hlindberg This appears to cause a regression of [PUP-5294](https://tickets.puppetlabs.com/browse/PUP-5294). I guess this has caused parameters to be validated for components in the environment catalog when they weren't previously.

In this case, if a components consumes a capability and uses it to set some parameter without providing a default value, it will fail. That should actually be okay, because it will be set during the agent run by the consumed resource.

nicklewis:
@hlindberg ...never mind on that, I may have misdiagnosed the issue. I'm seeing a strange failure and this looked potentially related, but it doesn't appear to be the cause as I dig deeper.

hlindberg:
@nicklewis ok - glad it was not this thing. I started looking into it and was just about to post something when your second message appeared.

PR #5249, opened 2016-08-31
(maint) Upgrade to new signing key

melissa:
This commit updates the key that is used to sign packages. We are
in the process of introducing a new signing key to verify our packages.
As a part of this effort, we need to actually switch over to this new
key for our individual projects.


melissa:
Currently blocked on getting some releases out the door, and then we'll go and bulk merge everything. I'm not sure which branches are currently in development, so I did everything that could potentially have a release in the future. If for some reason a release is cut from a branch that does not have these changes, we will see failures.

ScottGarman:
:+1:

PR #5252, opened 2016-09-01
(PUP-6670) Update tag validation regex to handle newlines

Sharpie:
This patch updates the tag validation regex so that it is anchored by `\A` and
`\Z` instead of `^` and `$`. The issue with using `$` is that it will terminate
validation when a newline is encountered, which allows non-valid tags to pass
through, which causes catalogs to be rejected by PuppetDB. The updated pattern
matches the anchors PuppetDB uses to validate tags.

### DIFF
--- a/lib/puppet/util/tagging.rb
+++ b/lib/puppet/util/tagging.rb
@@ -1,7 +1,7 @@
 require 'puppet/util/tag_set'
 
 module Puppet::Util::Tagging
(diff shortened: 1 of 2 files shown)


hlindberg:
@Sharpie Can I bother you with adding a test with an example of expected behavior?

Sharpie:
Sure! I can probably find a spot to add a test or two. Probably won't get to it until next week.

Sharpie:
Unit test added.

PR #5254, opened 2016-09-02
(PUP-4643) Add evaluation control functions break(), next(), and return()

hlindberg:
This adds the functions break(), next() and return() which control the order of evaluation:

* break() breaks the innermost iteration as if end of input was reached
* next() returns a value from a lambda, function, class body or define body
* return() returns a value from a function, class body or define body

When next() or return() is used to break() a class or define body, the value is ignored.

### DIFF
--- a/lib/puppet/functions/break.rb
+++ b/lib/puppet/functions/break.rb
@@ -1,0 +1,22 @@
+# Make iteration break as if there were no more values to process
+#
+# @since 4.7.0
+#
+Puppet::Functions.create_function(:break) do
+  dispatch :break_impl do
+  end
+
+  def break_impl()
+    stacktrace = Puppet::Pops::PuppetStack.stacktrace()
+    if stacktrace.size > 0
+      file, line = stacktrace[0]
+    else
+      file = nil
+      line = nil
+    end
+    # PuppetStopIteration contains file and line and is a StopIteration exception
+    # so it can break a Ruby Kernel#loop or enumeration
+    #
+    raise Puppet::Pops::Evaluator::PuppetStopIteration.new(file, line)
+  end
+end
--- a/lib/puppet/functions/next.rb
+++ b/lib/puppet/functions/next.rb
@@ -1,0 +1,23 @@
+# Make iteration continue with the next value optionally given a value for this iteration.
+# If a value is not given it defaults to `undef`
+#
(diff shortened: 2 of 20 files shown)


lib/puppet/functions/next.rb
```diff
@@ -0,0 +1,23 @@
...
+    stacktrace = Puppet::Pops::PuppetStack.stacktrace()
+    if stacktrace.size > 0
+      file, line = stacktrace[0]
+    else
+      file = nil
+      line = nil
+    end
+
+    exc = Puppet::Pops::Evaluator::Next.new(value, file, line)
+    raise exc
```
thallgren: Why divide this into two lines with an `exc` variable?

lib/puppet/functions/break.rb
```diff
@@ -0,0 +1,22 @@
+# Make iteration break as if there were no more values to process
+#
+# @since 4.7.0
+#
+Puppet::Functions.create_function(:break) do
+  dispatch :break_impl do
+  end
+
+  def break_impl()
+    stacktrace = Puppet::Pops::PuppetStack.stacktrace()
```
thallgren: unnecessary parenthesis

lib/puppet/functions/next.rb
```diff
@@ -0,0 +1,23 @@
...
+#
+# @since 4.7.0
+#
+Puppet::Functions.create_function(:next) do
+  dispatch :next_impl do
+    optional_param 'Any', :value
+  end
+
+  def next_impl(value = nil)
+    stacktrace = Puppet::Pops::PuppetStack.stacktrace()
```
thallgren: unnecessary parenthesis

lib/puppet/functions/return.rb
```diff
@@ -0,0 +1,22 @@
...
+#
+# @since 4.7.0
+#
+Puppet::Functions.create_function(:return, Puppet::Functions::InternalFunction) do
+  dispatch :return_impl do
+    optional_param 'Any', :value
+  end
+
+  def return_impl(value = nil)
+    stacktrace = Puppet::Pops::PuppetStack.stacktrace()
```
thallgren: unnecessary parenthesis

thallgren:
Found only nitpicks. Merged anyway.

PR #5258, opened 2016-09-06
(PUP-6681) Add documentation for Timespan.new and Timestamp.new

thallgren:
Adds documentation that explains how to create an instance of
Timespan och Timestamp using the Puppet `new()` function.

### DIFF
--- a/lib/puppet/parser/functions/new.rb
+++ b/lib/puppet/parser/functions/new.rb
@@ -153,6 +153,278 @@ $a_number = Numeric("010")   # results in 8
 $a_number = Numeric("3.14")  # results in 3.14 (a float)
 ```
 
+Conversion to Timespan
+-------------------
+
+A new `Timespan` can be created from `Integer`, `Float`, `String`, and `Hash` values. Several variants of the constructor are provided.
+
+#### Timespan from seconds
+
+```puppet
+function Timespan.new(
+  Variant[Float, Integer] $value
+)
+```
+
+#### Timespan from days, hours, mintues, seconds, and fractions of a second
+
+The arguments can be passed separately in which case the first four, days, hours, minutes, and seconds are mandatory and the rest are optional:
+
+```puppet
+function Timespan.new(
+  Integer $days, Integer $hours, Integer $minutes, Integer $seconds,
+  Integer $milli_seconds = 0, Integer $micro_seconds = 0, Integer $nano_seconds = 0
+)
+```
+
+or, all arguments can be passed as a `Hash`, in which case all entries are optional:
+
+```puppet
+function Timespan.new(
+  Struct[{
+    Optional[negative] => Boolean,
+    Optional[days] => Integer,
+    Optional[hours] => Integer,
+    Optional[minutes] => Integer,
+    Optional[seconds] => Integer,
+    Optional[milli_seconds] => Integer,
+    Optional[micro_seconds] => Integer,
+    Optional[nano_seconds] => Integer
+  }] $hash
+)
+```
+
+#### Timespan from String and patterns consisting of format directives
+
+The first argument is parsed using the format directives optinally passed as a string or array of strings. If the second argument is omitted, an array of default format directives will be used:
+
+```puppet
+function Timespan.new(
+  String $string, Variant[String[2],Array[String[2]], 1] $format = <default format>)
+)
+```
+
+the arguments may also be passed as a `Hash`:
+
+```puppet
+function Timespan.new(
+  Struct[{
+    string => String[1],
+    Optional[format] => Variant[String[2],Array[String[2]], 1]
+  }] $hash
+)
+```
+
+The directive consists of a percent (%) character, zero or more flags, optional minimum field width and
+a conversion specifier as follows:
+```
+%[Flags][Width]Conversion
+```
+
+##### Flags:
+
+| Flag  | Meaning
+| ----  | ---------------
+| -     | Don't pad numerical output
+| _     | Use spaces for padding
+| 0     | Use zeros for padding
+
+##### Format directives:
+
+| Format | Meaning |
+| ------ | ------- |
+| D | Number of Days |
+| H | Hour of the day, 24-hour clock |
+| M | Minute of the hour (00..59) |
+| S | Second of the minute (00..59) |
+| L | Millisecond of the second (000..999) |
+| N | Fractional seconds digits |
+
+The format directive that represents the highest magnitude in the format will be allowed to
+overflow. I.e. if no "%D" is used but a "%H" is present, then the hours may be more than 23.
+
+The default array contains the following patterns:
+
+```
+['%D-%H:%M:%S', '%D-%H:%M', '%H:%M:%S', '%H:%M']
+```
+
+Examples - Converting to Timespan
+
+```puppet
+$duration = Timespan(13.5)       # 13 seconds and 500 milliseconds
+$duration = Timespan({days=>4})  # 4 days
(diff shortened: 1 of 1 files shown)


lib/puppet/parser/functions/new.rb
```diff
@@ -153,6 +153,278 @@
...
+Conversion to Timespan
+-------------------
+
+A new `Timespan` can be created from `Integer`, `Float`, `String`, and `Hash` values. Several variants of the constructor are provided.
+
+#### Timespan from seconds
+
+```puppet
+function Timespan.new(
+  Variant[Float, Integer] $value
```
hlindberg: Since Float is supported, does the decimal part denote fractions of a second? (Documentation should say).

lib/puppet/parser/functions/new.rb
```diff
@@ -153,6 +153,278 @@
...
+```
+
+#### Timespan from days, hours, mintues, seconds, and fractions of a second
+
+The arguments can be passed separately in which case the first four, days, hours, minutes, and seconds are mandatory and the rest are optional:
+
+```puppet
+function Timespan.new(
+  Integer $days, Integer $hours, Integer $minutes, Integer $seconds,
+  Integer $milli_seconds = 0, Integer $micro_seconds = 0, Integer $nano_seconds = 0
```
hlindberg: With the fractions divided on milli, micro, nano it is possible to give a value at a higher resolution that spills over into the next bucket - if you do that, do they carry over or is it an error to overflow?

lib/puppet/parser/functions/new.rb
```diff
@@ -153,6 +153,278 @@
...
+Conversion to Timestamp
+-------------------
+
+A new `Timestamp` can be created from `Integer`, `Float`, `String`, and `Hash` values. Several variants of the constructor are provided.
+
+#### Timestamp from seconds since epoch (1970-01-01 00:00:00 UTC)
+
+```puppet
+function Timestamp.new(
+  Variant[Float, Integer] $value
```
hlindberg: What does decimals on Float represent (fractions of seconds I assume, but docs should state that).

lib/puppet/parser/functions/new.rb
```diff
@@ -153,6 +153,278 @@
...
+)
+```
+
+#### Timestamp from String and patterns consisting of format directives
+
+The first argument is parsed using the format directives optinally passed as a string or array of strings. If the second argument is omitted, an array of default format directives will be used:
```
hlindberg: Where does it state that the default directives are tried in order, and the first that matches the value is used?

lib/puppet/parser/functions/new.rb
```diff
@@ -153,6 +153,278 @@
...
+    Optional[milli_seconds] => Integer,
+    Optional[micro_seconds] => Integer,
+    Optional[nano_seconds] => Integer
+  }] $hash
+)
+```
+
+#### Timespan from String and patterns consisting of format directives
+
+The first argument is parsed using the format directives optinally passed as a string or array of strings. If the second argument is omitted, an array of default format directives will be used:
```
hlindberg: Where does it say that the defaults are tried in order and the first matching format is used?

lib/puppet/parser/functions/new.rb
```diff
@@ -153,6 +153,278 @@
...
+```
+
+##### Flags:
+
+| Flag  | Meaning
+| ----  | ---------------
+| -     | Don't pad numerical output
+| _     | Use spaces for padding
+| 0     | Use zeros for padding
+| #     | Change case
```
hlindberg: # is "alternative format", the description "change case" is confusing, change what from what? Does it mean that #P (am or pm) becomes AM PM and that #p (AM or PM) becomes am pm. ?

lib/puppet/parser/functions/new.rb
```diff
@@ -153,6 +153,278 @@
...
+
+**Time (Hour, Minute, Second, Subsecond):**
+
+| Format | Meaning |
+| ------ | ------- |
+| H | Hour of the day, 24-hour clock, zero-padded (00..23) |
+| k | Hour of the day, 24-hour clock, blank-padded ( 0..23) |
+| I | Hour of the day, 12-hour clock, zero-padded (01..12) |
+| l | Hour of the day, 12-hour clock, blank-padded ( 1..12) |
+| P | Meridian indicator, lowercase ("am" or "pm") |
```
hlindberg: Why is upper case P used for lower case 'am', 'pm', and vice versa? Seems confusing, I would expected it to be the other way around.

lib/puppet/parser/functions/new.rb
```diff
@@ -153,6 +153,278 @@
...
+| N | Fractional seconds digits, default is 9 digits (nanosecond). Digits under a specified width are truncated to avoid carry up |
+
+**Time (Hour, Minute, Second, Subsecond):**
+
+| Format | Meaning |
+| ------ | ------- |
+| z   | Time zone as hour and minute offset from UTC (e.g. +0900) |
+| :z  | hour and minute offset from UTC with a colon (e.g. +09:00) |
+| ::z | hour, minute and second offset from UTC (e.g. +09:00:00) |
+| Z   | Abbreviated time zone name or similar information.  (OS dependent) |
```
hlindberg: Do we really need the Z as it is problematic - if anything can go there it may not be understood if the value is later parsed on a different platform. Can we specify this in a strict way?

lib/puppet/parser/functions/new.rb
thallgren: Yes. '#' changes case. Not sure why that is confusing? I could add that "upper case becomes lower case and vice versa", but isn't that overstating it a bit?

Also, explaining differences between these flags and the ones used for String will probably just confuse things (this has nothing to do with "alternate format", which is a String format flag).

lib/puppet/parser/functions/new.rb
thallgren: Me too, but this is what strftime/strptime does in Ruby. In Python and C++ the lower case `%p` results in upper cased AM/PM and the upper case `%P` is missing.

Rectifying this in Puppet would call for a strftime parser/formatter implementation. Not a small task to write.

lib/puppet/parser/functions/new.rb
thallgren: Again, that would call for a parser/formatter implementation. At present, we just use the Ruby variant.

thallgren: That said, we could of course just remove this from the documentation/spec and treat the fact that %Z will work anyway as an unfortunate side-effect.

lib/puppet/parser/functions/new.rb
hlindberg: ok - so for the sake of consistency with other implementations... I am fine with this.

lib/puppet/parser/functions/new.rb
hlindberg: It is ok - coming to think about it, the feature may be needed when formatting (or parsing) for the purpose of interfacing with something on a specific system that uses that specific (and platform dependent) format. Leave it as is.

lib/puppet/parser/functions/new.rb
hlindberg: You could think it means changing the case of the format, say %B to %b as you have not seen that formats output things like "January". Also, you may think that "january" becomes "JANUARY", or that "January" becomes "jANUARY". I think the flag changes initial capitalization not case. Maybe say "Change initial capitalization" ?

PR #5259, opened 2016-09-06
(PUP-4643) Add acceptance for return,break,next

johnduarte:
This commit adds acceptance tests for the new `return`, `break`,
and `next` iteration control functions introduced as per PUP-4643.

### DIFF
--- a/acceptance/tests/language/return.rb
+++ b/acceptance/tests/language/return.rb
@@ -1,0 +1,72 @@
+test_name 'C98162 - Validate `return` immediately returns from a block of code' do
+  agents.each do |agent|
+
+    step 'apply class with return' do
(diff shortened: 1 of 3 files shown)


acceptance/tests/language/return.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    step 'apply function with return' do
+      manifest =<<EOS
+function example($x) {
+  $secret_word = "foo"
+  if $x == $secret_word {
+    return($secret_word)
+    notify{ 'You should not see this message': }
+  }
+  return('')
```
er0ck: you might want to put some code after the `return()` and assert we never get there

PR #5262, opened 2016-09-08
(PUP-6686) Improve how Timespan parses and stringifies fractions of second (%L and %N)

thallgren:
This PR aims to improve how a `Timespan` parses fractions. Instead of just treating the parsed value as milliseconds or nanoseconds, it will consistently be treated as fractions of a second. This is more intuitive and also aligned with how the Ruby implementation of `strptime()` does this (internally used as parser for `Timestamp`).

The PR also fixes the default format so that it contains fractions to avoid loss of precision when converting to and from string. Brought together, those changes made the string output more compact (default format strips trailing zeroes) and hence, more suitable to use a the default representation than the hash.

### DIFF
--- a/lib/puppet/pops/time/timespan.rb
+++ b/lib/puppet/pops/time/timespan.rb
@@ -343,7 +343,7 @@ module Time
           @literal.concat(codepoint)
         end
 
-        def multiplier
+        def nanoseconds
           0
         end
       end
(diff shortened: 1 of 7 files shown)


lib/puppet/pops/time/timespan.rb
```diff
@@ -671,7 +711,7 @@ def internal_parse(str)
     end
 
     class Format
-      DEFAULTS = ['%D-%H:%M:%S', '%D-%H:%M', '%H:%M:%S', '%H:%M'].map { |str| FormatParser.singleton.parse_format(str) }
+      DEFAULTS = ['%D-%H:%M:%S.%-N', '%H:%M:%S.%-N', '%M:%S.%-N', '%S.%-N', '%D-%H:%M:%S', '%H:%M:%S', '%D-%H:%M', '%S'].map { |str| FormatParser.singleton.parse_format(str) }
```
hlindberg: This needs to be in the spec as well. IIRC it now has the earlier version of default formats.

thallgren: No, it doesn't. The spec is also updated.

PR #5264, opened 2016-09-08
(PUP-6474) Incorporate gettext_setup gem into pupppet

ahenroid:


ahenroid:
This support was previously reverted. Puppet-agent now includes the gettext-setup gem and
dependencies. Some minor changes were required to this PR, so submitting a new commit
rather than simply reverting the revert.

locales/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # This is used in comments in the .pot and .po files to indicate what
+  # project the files belong to and should bea little more desctiptive than
+  # <project_name>
+  package_name: Puppet
+  # The locale that the default messages in the .pot file are in
+  default_locale: en
+  # The email used for sending bug reports.
+  bugs_address: docs@puppet.com
+  # The holder of the copyright.
+  copyright_holder: Puppet, LLC.
```
kylog: Elsewhere, Puppet has identified as "Puppet, Inc.", e.g. https://github.com/puppetlabs/puppet/blob/4.6.2/LICENSE#L3

IANAL so just double-checking whether this divergence was intentional.

locales/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # called <project_name>.pot?
+  project_name: 'puppet'
+  # This is used in comments in the .pot and .po files to indicate what
+  # project the files belong to and should bea little more desctiptive than
+  # <project_name>
+  package_name: Puppet
+  # The locale that the default messages in the .pot file are in
+  default_locale: en
+  # The email used for sending bug reports.
+  bugs_address: docs@puppet.com
```
kylog: Do we actually want people to email bug report to this address?

locales/config.yaml
```diff
@@ -0,0 +1,21 @@
+---
+# This is the project-specific configuration file for setting up
+# fast_gettext for your project.
+gettext:
+  # This is used for the name of the .pot and .po files; they will be
+  # called <project_name>.pot?
+  project_name: 'puppet'
+  # This is used in comments in the .pot and .po files to indicate what
+  # project the files belong to and should bea little more desctiptive than
```
kylog: s/bea/be a/

locales/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # <project_name>
+  package_name: Puppet
```
kylog: No idea what the conventions are here, but this doesn't actually seem more descriptive than the project_name :)

kylog:
Three questions:
* Where did we land on cross-project convention for including the generated files (e.g. .pot) in the repo?
* Does puppetserver need an update to add these gems as git submodules (for development workflows)?
* Do we need the work in https://tickets.puppetlabs.com/browse/EZ-85 to complete before landing this in puppet?

locales/config.yaml
ahenroid: Yes, this probably should be Inc. LLC came from an earlier commenter, but it doesn't sound correct. I'll track down the correct name.

locales/config.yaml
ahenroid: This came from the internal i18n template, but probably not the right address.

ahenroid:
We don't have a firm resolution on any of these questions, integration with the translation service is still a WIP, but I'll follow up with project owners and/or file some tickets around these questions.

ahenroid:
Pushed minor updates to address review comments and resolve CI failures. However,
this PR should not be merged until we bottom-out on answers for the questions above.

Gemfile
```diff
@@ -30,6 +30,12 @@ gem "rake", "10.1.1", :require => false
 # Hiera has an unbound dependency on json_pure
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
+# i18n support (gettext-setup and dependencies)
+gem 'text', '~> 1.3.1', :require => false
+gem 'locale', '~> 2.1.2-2', :require => false
+gem 'gettext', '~> 3.2.2', :require => false
+gem 'fast_gettext', '~> 1.1.0', :require => false
+gem 'gettext-setup', '~> 0.6', :require => false
```
joshcooper: Do we need to list all of the dependencies? I'm used to only listing the gems that puppet directly requires & calls, but not the transitive dependencies.

joshcooper: We also need to list the required dependencies in [`project_data.yaml`](https://github.com/puppetlabs/puppet/blob/master/ext/project_data.yaml#L19-L22) which is used when building puppet as a gem, and in [`.gemspec`](https://github.com/puppetlabs/puppet/blob/master/.gemspec#L37-L38) which is used when another project (typically a module) tests against a source checkout of puppet (like we do in module CI).

branan:
This is superseded by #5329

PR #5265, opened 2016-09-08
(PUP-6682) Clean $settings values from implementation runtime types

hlindberg:
Before this, symbols (and most likely other data values) leaked into the
puppet language and caused problems (a Symbol is not equal to a String
value).

This commit cleans all values and turns them into Integer, Float,
String, Array or Hash (containing clean values recursively).

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -868,10 +868,10 @@ class Puppet::Parser::Compiler
     settings_hash = {}
     Puppet.settings.each do |name, setting|
       next if name == :name
-      name = name.to_s
+      s_name = name.to_s
       # Construct a hash (in anticipation it will be set in top scope under a name like $settings)
-      settings_hash[name] = transform_setting(env[name])
-      scope[name.to_s] = settings_hash[name]
+      settings_hash[s_name] = transform_setting(env[name])
+      scope[s_name] = settings_hash[s_name]
     end
   end
 
@@ -884,7 +884,9 @@ class Puppet::Parser::Compiler
     when Array
       val.map {|entry| transform_setting(entry) }
     when Hash
-      Hash[val.map {|k,v| [transform_setting(k), transform_setting(v)] }]
+      result = {}
+      val.each {|k,v| result[transform_setting(k)] = transform_setting(v) }
+      result
     else
       # not ideal, but required as there are settings values that are special
       val.to_s
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -865,11 +865,32 @@ class Puppet::Parser::Compiler
     scope = @topscope.class_scope(settings_type)
 
     env = environment
+    settings_hash = {}
     Puppet.settings.each do |name, setting|
       next if name == :name
-      scope[name.to_s] = env[name]
+      name = name.to_s
+      # Construct a hash (in anticipation it will be set in top scope under a name like $settings)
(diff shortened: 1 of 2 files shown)


lib/puppet/parser/compiler.rb
```diff
@@ -865,11 +865,32 @@ def create_settings_scope
     scope = @topscope.class_scope(settings_type)
 
     env = environment
+    settings_hash = {}
     Puppet.settings.each do |name, setting|
       next if name == :name
-      scope[name.to_s] = env[name]
+      name = name.to_s
+      # Construct a hash (in anticipation it will be set in top scope under a name like $settings)
+      settings_hash[name] = transform_setting(env[name])
```
thallgren: Converting `name` to a `String`  before passing it to `env[name]` is inefficient. The first thing that happens in `settings#value()` is that it is converted back to a `Symbol`. I suggest keeping the behavior of passing it as a symbol.

lib/puppet/parser/compiler.rb
```diff
@@ -865,11 +865,32 @@ def create_settings_scope
...
+  def transform_setting(val)
+    case val
+    when Integer, Float, String
+      val
+    when Symbol
+      val == :undef ? nil : val.to_s
+    when Array
+      val.map {|entry| transform_setting(entry) }
+    when Hash
+      Hash[val.map {|k,v| [transform_setting(k), transform_setting(v)] }]
```
thallgren: Not sure how performance critical this is but I've done some tests and concluded that this is a slow way of mapping a `Hash` (see [PR-2 on ruby-performance](/zaphod42/ruby-performance/pull/2)). It's faster to:
```
result = {}
val.each { |k,v| result[transform_setting(k)] = transform_setting(v) }
result
```

lib/puppet/parser/compiler.rb
hlindberg: ok, it is needed as a string key in the settings_hash and in scope, will lookup with the symbol in env

lib/puppet/parser/compiler.rb
hlindberg: it is done once per compilation - so not time critical, still worth making it faster and produce less garbage.

spec/integration/parser/compiler_spec.rb
```diff
@@ -153,6 +153,17 @@ class c inherits b { notify { hi: } }
     end
   end
 
+  context 'when working with $settings name space' do
+    include PuppetSpec::Compiler
+    it 'makes $settings::strict available as string' do
+      node = Puppet::Node.new("testing")
+      catalog = compile_to_catalog(<<-MANIFEST, node)
+          notify { 'test': message => $settings::strict == 'warning' }
```
er0ck: does this need more spec tests/acceptance?
it looks to me (ignorantly) like this is only testing the String case in the switch statement in the PR.  i.e.: this spec change doesn't test the Symbol, Hash, Array cases (the fore-most being what the ticket is about)

hlindberg: It tests the symbol case since the strict setting is a symbol `:off`, `:warning`, `:error`.
IIRC, there is another PR dealing with additional types - we missed handling booleans correctly. That PR adds more tests.

er0ck: i saw the boolean addition.  seems fully tested if the strict setting is a Symbol.  i'm not overly worried about arrays and hashes, do we even have settings of those types?

PR #5267, opened 2016-09-08
(PUP-6641) Fix Service integration test for Windows 10

glennsarti:
Previously the integration test for the Service provider for Windows used the
lmhosts service as a test fixture.  However in Windows 10 the start type for
this service changed to Manual (Triggered) which now fails the integration test.

This commit adds `:manual` as a valid return value of `enabled?`, which now
matches the provider for service in Windows.

Note that this platform is not tested for integration tests but can easily be tested locally via VMPooler

### DIFF
--- a/spec/integration/provider/service/windows_spec.rb
+++ b/spec/integration/provider/service/windows_spec.rb
@@ -26,11 +26,12 @@ describe Puppet::Type.type(:service).provider(:windows), '(integration)',
 
   context 'should return valid values when querying a service that does exist' do
     let(:service) do
+      # This service should be ubiquitous across all supported Windows platforms
       Puppet::Type.type(:service).new(:name => 'lmhosts')
     end
 
-    it "with a valid boolean when asked if enabled" do
-      expect([:true, :false]).to include(service.provider.enabled?)
+    it "with a valid enabled? value when asked if enabled" do
+      expect([:true, :false, :manual]).to include(service.provider.enabled?)
     end
 
     it "with a valid status when asked about status" do


McdonaldSeanp:
:+1:

MosesMendoza:
@glennsarti I think I must be missing it - I don't see in the test(s) where we actually require the service's startup type to be automatic?

glennsarti:
The actual test is below;
https://github.com/glennsarti/puppet/blob/bcee048502c530c64d43c7e5cd953204286aa058/spec/integration/provider/service/windows_spec.rb#L34

Manual services come through as `manual` not `true` or `false`

MosesMendoza:
Ah I see. It looks like the Windows service provider's `enabled?` method is actually intended to return three possible values, `true`, `false`, and `manual` so perhaps instead that test would be better updated to expect one of those possible values?
 (https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/service/windows.rb#L52)

glennsarti:
I must admit, I don't really see the points of those tests.  They do not make me fell confident that the code is behaving correctly.

MosesMendoza:
Yeah I can I see that. Perhaps the idea was just a double check that the contract/signature of the method isn't accidentally changed unintentionally, but it seems like possibly a better test would be to set the service status before hand or even just stub the underlying service status query (aka to return `Win32::Service.get_start_type(Win32::Service::SERVICE_DEMAND_START)`) and then ensure that the `enabled?` method actually returns the right thing..

glennsarti:
After investigating the spec tests test whether specific values from FFI map to the correct provider values.  This integration test is merely a smoke test against a real Windows operating system i.e. Does calling the actual Windows OS generate a valid value (not necessarily a correct value).

MosesMendoza:
👍

PR #5268, opened 2016-09-09
(PUP-6693) Add --strict control over assignment to $server_facts

hlindberg:
Before this assignments to $server_facts would go unnoticed unless
--trusted_server_facts was turned on.
Now a warning/error (or ignore) is under the control of the --strict
option.

### DIFF
--- a/lib/puppet/pops/evaluator/runtime3_support.rb
+++ b/lib/puppet/pops/evaluator/runtime3_support.rb
@@ -59,10 +59,18 @@ module Runtime3Support
     # The error is not specific enough to allow catching it - need to check the actual message text.
     # TODO: Improve the messy implementation in Scope.
     #
+    if name == "server_facts"
+      if Puppet[:trusted_server_facts] || Puppet[:strict] == :error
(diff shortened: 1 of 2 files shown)


lib/puppet/pops/evaluator/runtime3_support.rb
```diff
@@ -59,11 +59,19 @@ def set_variable(name, value, o, scope)
...
+        file, line = extract_file_line(o)
+        msg = "Assignment to $server_facts is deprecated"
+        Puppet.warn_once(:deprecation, msg, msg, file, line)
+      end
+    end
+
     if scope.bound?(name)
       if Puppet::Parser::Scope::RESERVED_VARIABLE_NAMES.include?(name)
         fail(Issues::ILLEGAL_RESERVED_ASSIGNMENT, o, {:name => name} )
-      elsif name == "server_facts" && Puppet[:trusted_server_facts]
```
thallgren: Is the intention here to actually allow assignment to "server_facts" here when Puppet[:strict] is turned off or warning?

PR #5273, opened 2016-09-12
(PUP-6291) Add 'abs' support to #new function of Numeric types

thallgren:
This commit adds an additional 'abs' argument to the #new functions
of the Numeric, Integer, and Float types.

### DIFF
--- a/spec/unit/functions/new_spec.rb
+++ b/spec/unit/functions/new_spec.rb
@@ -90,6 +90,34 @@ describe 'the new function' do
       )).to have_resource('Notify[Integer, 1]')
     end
 
+    it "produces an absolute value when third argument is 'true'" do
+      expect(eval_and_collect_notices(<<-MANIFEST
+        notice(Integer.new(-42, 10, true))
+      MANIFEST
+      )).to eql(['42'])
+    end
+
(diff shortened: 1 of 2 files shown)


spec/unit/functions/new_spec.rb
```diff
@@ -362,6 +390,34 @@
...
+      MANIFEST
+      )).to eql(['42.3'])
+    end
+
+    it "does not produce an absolute value from hash {from => val, abs => false}" do
+      expect(eval_and_collect_notices(<<-MANIFEST
+        notice(Numeric.new({from => -42.3, abs => false}))
+      MANIFEST
+      )).to eql(['-42.3'])
+    end
```
hlindberg: The added examples (4 of them) use a Float value, but are in the context of testing Integer.

hlindberg:
Good to merge when it is made clear that the Numeric examples are not for testing in the context of "Integer values"

spec/unit/functions/new_spec.rb
thallgren: Yep, no idea how they ended up under Integer.

hlindberg: Doh...

PR #5275, opened 2016-09-13
(PUP-6636) Only match /environment/:env route as the whole path

McdonaldSeanp:
Previously, the /environment/:env route wasn't properly anchored to the
start of the path, so routes with other prefixes would also match.
Because auth is performed separately based on the path, this allowed
requests like /status/environment/production to bypass auth checks and
retrieve environment catalogs. This also meant that other valid requests
ending in /environment/:foo, such as for files, would be handled
incorrectly.

We now properly anchor the /environment/:env regex to the beginning of
the path, so only explicit environment requests will match.

This is a cherry-pick


branan:
I'm going to let the tests finish just in case, but this looks like a clean cherry-pick to me

Magisus:
Looks good to me 👍

PR #5276, opened 2016-09-13
(PUP-6783) Allow notify resources on both devices and normal hosts

DavidS:
This avoids unexpected failures from skipping notify resources, when logging
on devices.

### DIFF
--- a/lib/puppet/type/notify.rb
+++ b/lib/puppet/type/notify.rb
@@ -6,6 +6,8 @@ module Puppet
   Type.newtype(:notify) do
     @doc = "Sends an arbitrary message to the agent run-time log."
 
+    apply_to_all
+
     newproperty(:message, :idempotent => false) do
       desc "The message to be sent to the log."
       def sync


branan:
I doubt anyone is around who remembers the design decisions that led to notify not being a device resource. I don't see any reason not to do this, from my understanding of it

Iristyle:
Maybe @ahpook knows?

DavidS:
@branan how certain can you even be that there was a design decision involved at all?

Iristyle:
For historical reference it looks like the `apply_to_all` bit was added at https://github.com/puppetlabs/puppet/commit/07b677c5f6af8def03c5c30393fd83bc3986239a and only initially applied to `schedule`, while `apply_to_device` was marked for `interface` and `vlan`.

I think this change makes sense based on what you're seeing, but it would be nice to have someone comment who initially worked on device support.

@stahnma was the one who reviewed that PR back in 2011 ;0

er0ck:
@DavidS can you open a ticket for this, please.

DavidS:
Created PUP-6783 to track this.

Iristyle:
@DavidS if you can update the commit message, I think we can get this merged in.

DavidS:
Done!

PR #5277, opened 2016-09-13
(PUP-6701) Make settings data type cleanup handle boolean values

hlindberg:
A regression was introduced in PUP-6682 as the transformation of
settings values to safe values (no symbols) omitted handling boolean
values. They ended up being transformed to string, with the unfortunate
effect that a boolean `false` turned into the string `"true"` which is
"truthy".

This commit adds the correct handling of FalseClass and TrueType so that
they are not transformed to String values.

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -877,7 +877,7 @@ class Puppet::Parser::Compiler
 
   def transform_setting(val)
     case val
-    when Integer, Float, String
+    when Integer, Float, String, TrueClass, FalseClass
       val
     when Symbol
       val == :undef ? nil : val.to_s
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -877,7 +877,7 @@ class Puppet::Parser::Compiler
 
   def transform_setting(val)
     case val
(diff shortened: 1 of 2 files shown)


jpartlow:
Are there any other settings types?

lib/puppet/parser/compiler.rb
```diff
@@ -877,7 +877,7 @@ def create_settings_scope
 
   def transform_setting(val)
     case val
-    when Integer, Float, String
+    when Integer, Float, String, TrueClass, FalseClass
```
branan: should NilClass also be covered here?

hlindberg:
it should now support all the types of values that the puppet language can understand. There are other custom settings types I am sure - but they did not work earlier as rich data types - in fact, some of them did not work at all (that is what PUP-6682 fixed).

lib/puppet/parser/compiler.rb
hlindberg: It could - can a setting be nil?

hlindberg: Fixing, nil should not be turned into a string

jpartlow:
Ah ok, I see the final to_s now.  :+1:

PR #5278, opened 2016-09-14
(PUP-6291) Add documentation of `abs` parameter to #new() function.

thallgren:
Adds documentation of the optional `abs` parameter to Numeric#new(),
Integer#new(), and Float#new().


hlindberg:
WAT, Ruby 1.9.3 failed in travis. Cannot see how that is caused by this change.

hlindberg:
On closer look - it seems you managed to get some non ASCII in the file

PR #5281, opened 2016-09-15
(PUP-5834) Add support for resource ext_parameters to catalog

thallgren:
This PR adds a new `ext_parameters` entry to the JSON object that represents a `Resource`. The `ext_parameters`, like `parameters` is an object keyed by parameter names. The values are however not plain JSON types but instead an array representing an extended type created by the Puppet Serialization API.

Special hooks into the serializer enables several values to be generated, one at a time, while using the same tabulation context for all resources in the catalog.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1926,4 +1926,24 @@ EOT
           generate manifest documentation.",
     }
   )
+
+  define_settings(
+    :main,
+    :rich_data => {
+      :default  => false,
+      :type     => :boolean,
+      :hook    => proc do |value|
+        envs = Puppet.lookup(:environments) { nil }
+        envs.clear_all unless envs.nil?
+      end,
+      :desc     => <<-'EOT'
+        Enables storage of extended data in the catalog by introducing the new Resource
+        property `ext_parameters`. When enabled, values of type Regexp, SemVer, SemVerRange,
+        Timespan, and Timestamp, will no longer be converted to strings. Instead, they will
+        be serialized using Pcore serialization which allowes them to retain their type. It
+        will also be possible to use parameters that store a Binary.
+      EOT
+    }
+  )
+
 end
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1937,11 +1937,10 @@ EOT
         envs.clear_all unless envs.nil?
       end,
       :desc     => <<-'EOT'
-        Enables storage of extended data in the catalog by introducing the new Resource
-        property `ext_parameters`. When enabled, values of type Regexp, SemVer, SemVerRange,
-        Timespan, and Timestamp, will no longer be converted to strings. Instead, they will
-        be serialized using Pcore serialization which allowes them to retain their type. It
-        will also be possible to use parameters that store a Binary.
+        Enables having extended data in the catalog by adding the key `ext_parameters` to serialized
+        resources. When enabled, resource containing values of the data types `Binary`, `Regexp`,
+        `SemVer`, `SemVerRange`, `Timespan` and `Timestamp`, as well as instances of types derived
+        from `Object` retain their data type and are serialized using Pcore in `ext_parameters`.
       EOT
     }
   )
--- a/lib/puppet/pops/evaluator/runtime3_converter.rb
+++ b/lib/puppet/pops/evaluator/runtime3_converter.rb
@@ -86,32 +86,6 @@ class Runtime3Converter
     raise Puppet::Error, 'Use of an Iterator is not supported here'
   end
 
-  def convert_Regexp(o, scope, undef_value)
-    # Puppet 3x cannot handle parameter values that are reqular expressions. Turn into regexp string in
-    # source form
-    o.inspect
-  end
-
-  def convert_SemVer(o, scope, undef_value)
-    # Puppet 3x cannot handle SemVers. Use the string form
-    o.to_s
-  end
-
-  def convert_SemVerRange(o, scope, undef_value)
-    # Puppet 3x cannot handle SemVerRanges. Use the string form
-    o.to_s
-  end
-
-  def convert_Timespan(o, scope, undef_value)
-    # Puppet 3x cannot handle Timespans. Use the string form
-    o.to_s
-  end
-
-  def convert_Timestamp(o, scope, undef_value)
-    # Puppet 3x cannot handle Timestamps. Use the string form
-    o.to_s
-  end
-
   def convert_Symbol(o, scope, undef_value)
     o == :undef && !@inner ? undef_value : o
   end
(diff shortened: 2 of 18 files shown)


lib/puppet/pops/evaluator/runtime3_converter.rb
```diff
@@ -161,13 +135,47 @@ def catalog_type_to_split_type_title(catalog_type)
...
+  def convert_Timespan(o, scope, undef_value)
+    # Puppet 3x cannot handle Timespans. Use the string form unless allow_extended
+    o.to_s
+  end
+
+  def convert_Timestamp(o, scope, undef_value)
+    # Puppet 3x cannot handle Timestamps. Use the string form unless allow_extended
+    o.to_s
+  end
+
```
thallgren: Absolutely, once they are added to serialization.

spec/unit/pops/evaluator/runtime3_converter_spec.rb
```diff
@@ -5,21 +5,107 @@
...
+      expect(c).to be_a(String)
+      expect(c).to eql('0-00:00:00.1234')
+    end
+
+    it 'converts a Timestamp instance to string' do
+      c = converter.convert(Puppet::Pops::Time::Timestamp.parse('2016-09-15T12:24:47.193 UTC'), {}, nil)
+      expect(c).to be_a(String)
+      expect(c).to eql('2016-09-15T12:24:47.193 UTC')
+    end
+  end
```
thallgren: They are not serializable yet.

lib/puppet/pops/evaluator/runtime3_converter.rb
```diff
@@ -51,98 +52,75 @@ def map_args(args, scope, undef_value)
...
 
   def convert_Iterator(o, scope, undef_value)
     raise Puppet::Error, 'Use of an Iterator is not supported here'
   end
-  alias :convert2_Iterator :convert_Iterator
 
   def convert_Regexp(o, scope, undef_value)
     # Puppet 3x cannot handle parameter values that are reqular expressions. Turn into regexp string in
-    # source form
+    # source form unless allow_extended
```
thallgren: Part of another solution attempt. Should have been removed (and is removed now).

lib/puppet/defaults.rb
```diff
@@ -1926,4 +1926,24 @@ def self.default_diffargs
...
+      :hook    => proc do |value|
+        envs = Puppet.lookup(:environments) { nil }
+        envs.clear_all unless envs.nil?
+      end,
+      :desc     => <<-'EOT'
+        Enables storage of extended data in the catalog by introducing the new Resource
+        property `ext_parameters`. When enabled, values of type Regexp, SemVer, SemVerRange,
+        Timespan, and Timestamp, will no longer be converted to strings. Instead, they will
+        be serialized using Pcore serialization which allowes them to retain their type. It
+        will also be possible to use parameters that store a Binary.
```
hlindberg: "When enabled, resource attributes having values of the data types Binary, Regexp, SemVer, SemVerRange, Timespan and Timestamp, as well as instances of types derived from Object retain their data type and are serialized using Pcore in `ext_parameters`.

hlindberg: Sensitive is not mentioned here - I suppose because it is transformed before being stored as a value.

lib/puppet/resource.rb
```diff
@@ -71,23 +80,54 @@ def to_data_hash
...
-
+    ext_params = {}
+    params = {}
+    self.to_hash.each_pair do |param, value|
       # Don't duplicate the title as the namevar
-      next hash if param == namevar and value == title
-
-      hash[param] = Puppet::Resource.value_to_pson_data(value)
-      hash
+      unless param == namevar and value == title
```
hlindberg: Use `&&` instead of `and`

lib/puppet/resource.rb
```diff
@@ -71,23 +80,54 @@ def to_data_hash
...
+        value = Puppet::Resource.value_to_pson_data(value)
```
hlindberg: Should this be done first? Now this is done also for "non json type"

spec/unit/pops/evaluator/runtime3_converter_spec.rb
hlindberg: They are now - that logic is now merged.

hlindberg:
Think this needs unit tests that take a catalog containing ext_parameters written to disk and then applied to show that rich data types survive catalog serialization and can be used agent side. This in turn may require a new resource type as none of the existing types make use of rich data types.

thallgren:
I agree that a test must be written that involves a master and an agent where ext_parameters are applied. I think though, that we really must have an acceptance test that does that. Just writing unit test that writes the json to disk and then read it again will only test disk I/O.  If you agree that an acceptance test is needed, then duplicating more or less the same steps in a unit test is just pointless.

lib/puppet/defaults.rb
thallgren: Resource _attributes_? Should't it be Resource _parameters_? Seems odd that _attributes_ are serialized in _ext_parameters_.

hlindberg: A resource has parameters and properties - together they are the attributes of the resource (the expressions that set them have names like AttributeOperation). When serialized they end up under "parameters" irrespective of what kind of attribute they are (although some are properties).
In the sentence I wrote I am also deliberately using 'attributes' as a non technical term to distinguish it from the technical Parameter and Property kinds of attributes. I could even imagine saying "resources containing values of the data types..."

hlindberg:
The unit test is far from meaningless and much more than just deserialization as it tests an apply cycle without first having done a compilation - i.e. it tests that a "puppet apply of catalog with rich data" works (and does not regress). The two codepaths (apply and apply catalog from disk) are very different as when you are using apply from source the compilation results will be used without serialization thus using resources that hold the instances that the evaluator created and assigned. The unit test should jog that it is possible to find types on the agent side and instantiate them. Maybe it is not possible to do this without too much mocking - but it would be best if there was a unit test for this.

For the actual interaction between an agent and a master, other features (also) come into play. Not sure if we are missing features here - maybe the request to the master needs to contain if the agent accepts rich data or not (rather than it being a configuration option master side per environment). We will leave it as implemented for starters and then discuss this more widely. As this is experimental at this point, it does not matter that much that the configuration has to be changed on both master and agent side (at least I assumed the agent also needed to turn on rich data since the setting is in the MAIN section).

hlindberg:
Ping @adrienthebo - this is the work that enables things like Binary and Encrypted values to be transported to an agent in the catalog. (Experimental behind a feature flag to begin with).

thallgren:
The only advantage of having coverage in unit tests of code-paths that will be covered in acceptance tests, is that they trigger earlier in the process. Do we really need that for this particular case? I find that questionable. Add the concern for "too much mocking" (which I agree with), and the motivation for spending time on writing such a unit test becomes rather hard to find. Do you really want to make this a priority @hlindberg?

lib/puppet/resource.rb
thallgren: A test for `is_json_type(value)` on a value that is a `Resource` would return `false`, so yes, this has to be done first to avoid that a `Resource` is serialized as an `ext_parameter`.

hlindberg:
I do think it is of value to have non acceptance tests that cover the functionality. As the code stands now, only "half" of the solution is covered together with the lower level features used by the agent side. I think it is of value to show that - if a suitable plugin sync has been performed to an agent, then, the configuration of "puppet apply" will work (essentially that the configuration of environment/loaders/modules agent side indeed works). An acceptance test is needed to test that a real plugin sync via http access works and that it brings over the artifacts in such a way that the agent can use them - but an acceptance test is not required to test the rest. The integration test I am thinking of would contain a custom type, and a custom provider and should assert that this works if the module containing that is placed where pluginsync places it. It could do this by writing a canned catalog to ensure that a compilation is not performed where resources created by the compilation ends up being reused. My suspicion is that it simply does not work agent side as the code stands (for example an Object type bound to an implementation). I believe a canned catalog apply is relatively easy to achieve (there are already similar tests in spec/application/apply_spec.rb).

I tried finding unit/integration tests for plugin sync (what exactly is synced and how) as those would also benefit from not requiring an acceptance test for every detail. Have not found them yet at least.

hlindberg:
@thallgren this is failing the commits test because of a FACTER commit that I guess must be on master (in error). Not sure how to proceed.

thallgren:
No, not FACTER, it fails on some commit comment constraint. Looks like a bug in the constraint to me, because I can't see anything wrong with the commit comment.

PR #5280, opened 2016-09-15
(PUP-6703) Tweak failover_master test for robustness

MosesMendoza:
On 9/14 we had a failed cell (spurious failure) in the matrix run for test
acceptance/tests/reports/failover_master.rb due to a timeout sending the report
to the master for the test modified in this commit, 'master_field should not
appear when no master could be contacted'.

Prior to this commit, this test used two nonexistent servers in a server list
and specified a timeout of 3 seconds. In the beaker run noted, this led to a
test execution time of approximately 25 seconds. This commit updates the
timeout to 5 seconds, but reduces the check to only one nonexistent master as a
way to diminish the impact on overall test time of increasing the timeout.  The
fact that two masters are specified isn't actually intrinsic to the test, so
this should effectively be a no-op in terms of test effectiveness while
decreasing the likelihood of encountering a timeout during report submission.

PR #5282, opened 2016-09-15
(maint) Adjust rake commit check for stable

MosesMendoza:
On PR https://github.com/puppetlabs/puppet/pull/5280 against puppet, the travis
CI commit check is failing because a prior commit unrelated to that PR be753a4
did not adhere to the commit guidelines and the underlying check code lists
commits unreachable from master to HEAD of the PR. This commit adjust this to
be stable, and this PR is intended for the stable branch only. On merge up to
master, the commit should be taken but master's version of this change
accepted (for example by doing git merge --no-commit, checking out master's version, and proceeding)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -77,10 +77,15 @@ end
 
 desc "verify that commit messages match CONTRIBUTING.md requirements"
 task(:commits) do
-  # This git command looks at the summary from every commit from this branch not in master.
-  # Ideally this would compare against the branch that a PR is submitted against, but I don't
-  # know how to get that information. Absent that, comparing with master should work in most cases.
-  %x{git log --no-merges --pretty=%s master..$HEAD}.each_line do |commit_summary|
+  # This rake task looks at the summary from every commit from this branch not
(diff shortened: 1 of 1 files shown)


branan:
Instead of hard-coding this on any branches, can we just query it from the environment? (either git or Travis) Having it different on every branch is a great opportunity for us to break it during a merge-up or merge-down, and it is an extra thing to keep track of if we have to make additional branches

branan:
It looks like there is a `TRAVIS_BRANCH` that shows the target of the PR

MosesMendoza:
+1 on dynamically finding the branch. I just figured I'd make sure to do the merge-up, so we'd only have to handle it once, but I agree that's more brittle

MosesMendoza:
@branan updated with dynamic check, but using `TRAVIS_COMMIT_RANGE` which actually works quite well for this purpose.

Magisus:
Good catch on this! 👍 I find it funny that the commit that prompted us to look into this would have broken this check regardless...

PR #5283, opened 2016-09-16
DO NOT MERGE (PUP-6708) Cache last run report even if report_server unreachable

MosesMendoza:
Prior to this commit, when the agent could not reach its report server after an
agent run, it would not save a copy of its report locally either. This is
problematic because if an agent's report server is unreachable, there is
effectively no reporting on the agent's activities during this period of
unreachability.

This behavior existed because last run report was saved via the indirector
cache - specifically, Puppet::Indirector::Indirection#save issued the terminus
save request first, then the cache save request. If a failure to send the
report to the report_server occured, the cache save was never reached.

However, unlike the last run report, the last run summary is explicitly saved
via a configurer method, save_last_run_summary, which attempts to save the last
run summary file prior to calling Puppet::Indirector::Indirection#save.

This commit moves the saving of the last run report file to a
save_last_run_report method to configurer.rb which saves the report before
sending it, following the pattern set for save_last_run_summary.

In addition to always attempting to save the report, this has an additional
benefit of explicitly managing the mode of the last_run_summary.yaml file, as
opposed to the report handler caching yaml which did not manage file mode at
all.

### DIFF
--- a/lib/puppet/application/agent.rb
+++ b/lib/puppet/application/agent.rb
@@ -401,8 +401,6 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
     Puppet.settings.use :main, :agent, :ssl
 
     Puppet::Transaction::Report.indirection.terminus_class = :rest
-    # we want the last report to be persisted locally
-    Puppet::Transaction::Report.indirection.cache_class = :yaml
 
     if Puppet[:noop]
       Puppet::Resource::Catalog.indirection.cache_class = nil
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -388,6 +388,7 @@ class Puppet::Configurer
   def send_report(report)
     puts report.summary if Puppet[:summarize]
     save_last_run_summary(report)
+    save_last_run_report(report)
     Puppet::Transaction::Report.indirection.save(report, nil, :environment => Puppet::Node::Environment.remote(@environment)) if Puppet[:report]
   rescue => detail
     Puppet.log_exception(detail, "Could not send report: #{detail}")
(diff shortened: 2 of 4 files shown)


lib/puppet/configurer.rb
```diff
@@ -402,6 +403,15 @@ def save_last_run_summary(report)
     Puppet.log_exception(detail, "Could not save last run local report: #{detail}")
   end
 
+  def save_last_run_report(report)
+    mode = Puppet.settings.setting(:lastrunreport).mode
+    Puppet::Util.replace_file(Puppet[:lastrunreport], mode) do |fh|
+      fh.print YAML.dump(report)
+    end
+  rescue => detail
+    Puppet.log_execption(detail, "Could not save last run local report: #{detail}")
```
joshcooper: typo

MosesMendoza: Thanks @joshcooper, updated. That also pointed out this case wasn't sufficiently tested - test added.

Magisus:
This seems like a great addition. Also the test coverage looks good 👍

lib/puppet/application/agent.rb
```diff
@@ -401,8 +401,6 @@ def setup
     Puppet.settings.use :main, :agent, :ssl
 
     Puppet::Transaction::Report.indirection.terminus_class = :rest
-    # we want the last report to be persisted locally
-    Puppet::Transaction::Report.indirection.cache_class = :yaml
```
joshcooper: By unsetting this, we'll use the default `cache_class`. If that's `nil` then everything is ok, but if it's `yaml` or `json` we'll end up caching the report a second time. I pretty sure that's not the case, but wanted to ask.

lib/puppet/configurer.rb
MosesMendoza: thanks for the tip @joshcooper, will validate

MosesMendoza:
@joshcooper good catch there. In this PR in the specific case where we're running as `Puppet::Application::Agent`, `cache_class` is `nil` and thus we don't call `cache.save` (only writing the report once).

However, when running via a different mechanism, ie `Puppet::Application::Apply`, `cache_class` is still set to `yaml` and later we do indeed write the report twice, once via the new `Puppet::Configurer#save_last_run_report` and then again via `Puppet::Indirector::Indirection#save`. In order for this PR to work as intended it would need to remove the `Puppet::Transaction::Report.indirection.cache_class` everywhere it's set. I'm not sure if this feels quite right and I'm going to explore if there are simpler/cleaner ways to accomplish this.

MosesMendoza:
An alternative to the implementation in this PR would be to `ensure` we call `cache.save` after `terminus.save` in `Puppet::Indirector::Indirection#save` - exploring implications

MosesMendoza:
`ensure` ing `cache.save` won't work - some indirectors expect failure to cause us to _avoid_ a cache save. Something slightly different will be needed. I'm closing this until we can get it into a sprint properly and focus on it.

PR #5284, opened 2016-09-18
(maint) Remove unused: requre 'xmlrpc/client'

kylog:
In 152299cc, as part of PUP-6120, use of the xmlrpc library
was removed from the code, but the require itself was not
removed. This commit simply removes the vestigial require.

Note that this change is required to support ruby 2.4,
which no longer includes the xmlrpc library.


kylog:
Hey maintainers, anyone wanna peek at this PR? I think it's pretty harmless.

PR #5285, opened 2016-09-19
(PUP-6758) Fix pkgng provider condition

xaque208:
In cases where pkg execution fails within instances, the result returned
to prefetch is nil causing prefetch to fail poorly calling .find on a
nil object.  Here we modify the return value of instances in the event
of an exception to be an array instead of nil.  This leaves a cleaner
prefetch behavior intact, and allows the run to continue, even in the
absence of a successful package call.

### DIFF
--- a/lib/puppet/provider/package/pkgng.rb
+++ b/lib/puppet/provider/package/pkgng.rb
@@ -56,7 +56,7 @@ Puppet::Type.type(:package).provide :pkgng, :parent => Puppet::Provider::Package
 
       return packages
     rescue Puppet::ExecutionFailure
-      nil
+      return []
     end
   end
 
--- a/spec/unit/provider/package/pkgng_spec.rb
+++ b/spec/unit/provider/package/pkgng_spec.rb
@@ -65,6 +65,11 @@ describe provider_class do
       expect(nmap.properties[:version]).to eq(nmap.properties[:latest])
     end
 
+    it "should return an empty array when pkg calls raise an exception" do
+      provider_class.stubs(:get_query).raises(Puppet::ExecutionFailure, 'An error occurred.')
+      expect(provider_class.instances).to eq([])
+    end
+
     describe "version" do
       it "should retrieve the correct version of the current package" do
         zsh = provider_class.instances.find {|i| i.properties[:origin] == 'shells/zsh' }
@@ -108,6 +113,13 @@ describe provider_class do
         arg.should include('FreeBSD')
       end
       resource.provider.install
+    end
+  end
+
+  context "#prefetch" do
+    it "should fail gracefully when " do
+      provider_class.stubs(:instances).returns([])
+      expect{ provider_class.prefetch({}) }.to_not raise_error
     end
   end
 


xaque208:
@hlindberg Thank you.  Both the PR and the commit now use the imperative mode, referencing the ticket number.

branan:
That delightful bug in our commit checker hit this PR. The commit format is fine, and Travis has otherwise passed

lib/puppet/provider/package/pkgng.rb
```diff
@@ -61,7 +61,7 @@ def self.instances
   end
 
   def self.prefetch(resources)
-    packages = instances
+    packages = instances || []
```
Iristyle: I'm not sure that hiding such a failure is the correct behavior.

I can't find other package providers that work like this.

For instance, the `yum` provider will bubble up errors like this: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/yum.rb#L28-L31

It relies on the behavior of the `rpm` provider, which also will throw errors if the list of packages cannot be retrieved with a call to `instances` - see https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/rpm.rb#L78-L95

You've noted in the ticket:


> The fix here is to return an empty array of package instances in the event of a non-zero exit call to pkg(8) during instances enumeration of the existing packages.  This might happen for a variety of reasons, one of which would be an empty list of repositories.


But that's not quite what you've done here.  I think this PR should change the implementation of `instances`, so that when there are no sources it returns an empty array.  If there are other valid reasons for no packages to be listed, we should handle that scenario appropriately.  But if there are complete failures to  execute the command to list packages, then that should be a failure condition with the resource.

xaque208: Thanks for the review @Iristyle.  Ah, you're correct, I've not modified the instance to return an empty package list.  I can reword as well, or give more backstory in the commit/pr/ticket.  This didn't used to be a problem until a relatively recent version of pkg(8) began exiting non-zero for some cases.  This only currently comes up when there are no repositories, or at least that is the use case I'm targeting.  The solution as it stands now (wording aside) should mean that calls to `pkg` exiting non-zero causing a failure in the `instances` call ends up with an empty package list rather than bad provider objects.  I think I leaned that way out of habit for other types/providers I work on where the results of instances aren't as clean as they should be.

Do you think modifying line 58 of the provider here return an empty array rather than nil is the right move then?  It seems like that would be the other simple option for this change.

xaque208: @Iristyle I believe I've adjusted the code to be in line with your desires.  It feels cleaner to handle the exception in the instances.  I've left the old test in place and added a new one as well.  I've also updated the commit message and the PR message.  The ticket wording should be accurate as is.  Let me know how this looks now.

xaque208:
I believe I've addressed the feedback, and the travis failure on commit message is not accurate as Branan mentions.  Is there anything else on this PR that needs attention?

branan:
@Iristyle are you happy with the adjustments made here? You're still marked as a "changes requested" review

xaque208:
Is there anything else I can help with to move this along?

Iristyle:
Sorry I lost track of this one @xaque208 - I still think there are some improvements that can be made to the `prefetch` / `instances` behavior, but with these new changes you're pretty inline with what's already shipping.  Given this solves your problem, I say 👍  to at least get this in.

It would be nice at some point to be a little more careful about which errors are swallowed, in the event there's a catastrophic problem that prevents Puppet from managing packages / attempting to install already installed packages / etc.

xaque208:
Excellent, thank you.  I appreciate the review and commit.  Will this likely be in 4.7.1?

adrienthebo:
Given that we just shipped 4.8.0 I don't expect that we'll do a 4.7.1. In addition this was merged into master so pulling this into 4.8.1 will require backporting it, so as-is it'll most likely land in 4.9.0.

xaque208:
Cool, thank you @adrienthebo!

Iristyle:
The LTS release is taking high impact bug fixes, so it's possible this gets out in the next puppet-agent 1.7.z packages / Puppet 4.7 if you think it's high enough priority.

xaque208:
If you can get it in a release, that would be great.  I don't tend to track the latest puppet on the FreeBSD port, but this alone warrants a bump.  If it has to wait, well, it has to wait.  Call it punishment for the bug in the first place. :)

PR #5288, opened 2016-09-21
(PUP-6706) Add Pcore serialization for Binary data type

thallgren:
This commit adds everything needed to let instances of the Binary
data type participate in Pcore serialization. The binary will be
serialized using strict base64 encoding. Although not optimal (it
has a 33% overhead), it's widely available, relatively cheap in terms
of CPU-cycles, and well known.

Other algorithms such as base85 (7% less space overhead) is more
expensive to compute and not as common. basE91 has even less space
overhead but contains the double quote. It doesn't seem worth while to
pursue either of them. The pros with base64 outweighs the cons.

Compression was considered but rejected since: a) large binaries might
have be compressed already, b) it's very CPU demanding at both ends, and
c) it's not uncommon that compression is part of the actual transfer
protocol.

### DIFF
--- a/lib/puppet/pops/serialization/abstract_reader.rb
+++ b/lib/puppet/pops/serialization/abstract_reader.rb
@@ -152,6 +152,10 @@ class AbstractReader
     register_type(Extension::VERSION_RANGE) do |data|
       read_payload(data) { |ep| Semantic::VersionRange.parse(ep.read) }
     end
+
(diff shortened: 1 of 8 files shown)


thallgren:
Marking this as blocked since it's based on top of [PUP-6705/PR-5286](https://github.com/puppetlabs/puppet/pull/5286) to avoid merge conflicts.

PR #5290, opened 2016-09-22
(PUP-6723) collector_transformer: negation query for array type properties is broken

isaiahfrantz:
(PUP-6723) collector_transformer: negation query for array type properties is broken

Only in the == case is the query property getting tested to see if it is an array which then uses the compare_operator.include? method. The != case only uses compare_operator.equals which makes it so you cant do negated queries on array type properties like tags :(

With this test code:
> $ cat t.pp
@notify {'This is a test 1':
	tag => 'one'
}
@notify {'This is a test 2':
	tag => 'two'
}
@notify {'This is a test 3':
	tag => ['one', 'two']
}
@notify {'This is a test 4':
	tag => 'three'
}

Adding these at the end would all work as expected
> Notify <| tag = 'one' |>
Notify <| tag = 'two' |>
Notify <| tag = 'one' or tag = 'two' |>
Notify <| tag = 'one' and tag = 'two' |>

The negated versions would not
> Notify <| tag != 'one' |>
Notify <| tag != 'two' |>
Notify <| !tag = 'one' or !tag = 'two' |>
Notify <| !tag = 'one' and !tag = 'two' |>

The first should only return 2 and 4 but also returns 3:
> $ puppet apply t.pp 
Notice: Compiled catalog for defiant.cequintecid.com in environment production in 0.08 seconds
Notice: This is a test 2
Notice: /Stage[main]/Main/Notify[This is a test 2]/message: defined 'message' as 'This is a test 2'
Notice: This is a test 3
Notice: /Stage[main]/Main/Notify[This is a test 3]/message: defined 'message' as 'This is a test 3'
Notice: This is a test 4
Notice: /Stage[main]/Main/Notify[This is a test 4]/message: defined 'message' as 'This is a test 4'

The second should only return 1 and 4 but also returns 3:
> $ puppet apply t.pp 
Notice: Compiled catalog for defiant.cequintecid.com in environment production in 0.07 seconds
Notice: This is a test 1
Notice: /Stage[main]/Main/Notify[This is a test 1]/message: defined 'message' as 'This is a test 1'
Notice: This is a test 3
Notice: /Stage[main]/Main/Notify[This is a test 3]/message: defined 'message' as 'This is a test 3'
Notice: This is a test 4
Notice: /Stage[main]/Main/Notify[This is a test 4]/message: defined 'message' as 'This is a test 4'

The "or" case should only return 1,2, and 4. It should not return 3 but does:
> $ puppet apply t.pp 
Notice: Compiled catalog for defiant.cequintecid.com in environment production in 0.08 seconds
Notice: This is a test 1
Notice: /Stage[main]/Main/Notify[This is a test 1]/message: defined 'message' as 'This is a test 1'
Notice: This is a test 2
Notice: /Stage[main]/Main/Notify[This is a test 2]/message: defined 'message' as 'This is a test 2'
Notice: This is a test 3
Notice: /Stage[main]/Main/Notify[This is a test 3]/message: defined 'message' as 'This is a test 3'
Notice: This is a test 4
Notice: /Stage[main]/Main/Notify[This is a test 4]/message: defined 'message' as 'This is a test 4'

The "and" case should only return 4. It should not return 3 but it does:
> $ puppet apply t.pp 
Notice: Compiled catalog for defiant.cequintecid.com in environment production in 0.07 seconds
Notice: This is a test 3
Notice: /Stage[main]/Main/Notify[This is a test 3]/message: defined 'message' as 'This is a test 3'
Notice: This is a test 4
Notice: /Stage[main]/Main/Notify[This is a test 4]/message: defined 'message' as 'This is a test 4'

### DIFF
--- a/lib/puppet/pops/evaluator/collector_transformer.rb
+++ b/lib/puppet/pops/evaluator/collector_transformer.rb
@@ -123,7 +123,11 @@ protected
       end
     when :'!='
       proc do |resource|
-        !@@compare_operator.equals(resource[left_code], right_code)
+        if (tmp = resource[left_code]).is_a?(Array)
+          !@@compare_operator.include?(tmp, right_code, scope)
+        else
+          !@@compare_operator.equals(tmp, right_code)
+        end
       end
     end
   end
--- a/spec/integration/parser/collection_spec.rb
+++ b/spec/integration/parser/collection_spec.rb
@@ -122,16 +122,6 @@ describe 'collectors' do
       MANIFEST
     end
 
-    it "allows criteria to be combined with 'or'" do
-      expect_the_message_to_be(["the message", "other message"], <<-MANIFEST)
-        @notify { "testing": message => "the message" }
-        @notify { "other": message => "other message" }
-        @notify { "yet another": message => "different message" }
-
-        Notify <| title == "testing" or message == "other message" |>
-      MANIFEST
-    end
-
     it "allows criteria to be grouped with parens" do
       expect_the_message_to_be(["the message", "different message"], <<-MANIFEST)
         @notify { "testing":     message => "different message", withpath => true }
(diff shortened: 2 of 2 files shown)


hlindberg:
Thanks for the contribution and discussion on the ticket. As noted there, I will close this in favor for instead adding the ability to use the `in` operator in queries.

PR #5292, opened 2016-09-23
(PUP-6724) Make strftime compatible with function in stdlib

thallgren:
This commit adds a new disptacher to the strftime function which is
backward compatible with the stdlib function with the same name.

The stdlib function also accepted a second `timezone` argument which
the new function did not provideand that was an oversight. It must be
provided to enable formatting of timezones other than UTC. The same
thing is true when parsing timestamps. A timezone must be provided to
parse a timestamp string that doesn't include a timezone unless a
default of UTC is acceptable in all cases (which it of course, isn't).
This commit therefore provides the timezone functionality needed to both
parse and format a timestamp.

### DIFF
--- a/lib/puppet/parser/functions/new.rb
+++ b/lib/puppet/parser/functions/new.rb
@@ -302,11 +302,22 @@ The first argument is parsed using the format optionally passed as a string or a
 will be made to parse the string using the first entry and then with each entry in succession until parsing succeeds. If the second
 argument is omitted, an array of default formats will be used.
 
+A third optional timezone argument can be provided. The first argument will then be parsed as if it represents a local time in that
+timezone. The timezone can be any timezone that is recognized when using the '%z' or '%Z' formats, or the word 'current', in which
+case the current timezone of the evaluating process will be used. The timezone argument is case insensitive.
+
+The default timezone, when no argument is provided, or when using the keyword `default`, is 'UTC'.
+
+It is illegal to provide a timezone argument other than `default` in combination with a format that contains '%z' or '%Z' since that
+would introduce an ambiguity as to which timezone to use. The one extracted from the string, or the one provided as an argument.
+
 An exception is raised when no format was able to parse the given string.
 
 ```puppet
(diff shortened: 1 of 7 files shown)


lib/puppet/parser/functions/new.rb
```diff
@@ -439,6 +451,11 @@
 
 The default array contains the following patterns:
 
+When a timezone argument (other than `default`) is exclicitly provided:
```
hlindberg: typo 'exclicitly'

lib/puppet/parser/functions/new.rb
```diff
@@ -439,6 +451,11 @@
...
+```
+['%FT%T.L', '%FT%T', '%F']
+```
+otherwise:
```
hlindberg: blank line above and below pls.

lib/puppet/parser/functions/strftime.rb
```diff
@@ -135,8 +135,14 @@
...
-# Notice the timestamp using a format that notices weekday, month, day, time, and year
-notice($$timestamp.strftime('%c')) # outputs 'Wed Aug 24 12:13:14 2016'
+# Notice the timestamp using a format that notices weekday, month, day, time (as UTC), and year
+notice($timestamp.strftime('%c')) # outputs 'Wed Aug 24 12:13:14 2016'
+
+# Notice the timestamp using a specific timezone
+notice($timestamp.strftime('%F %T %z', 'PST')) # outputs '2016-08-24 04:13:14 -0800'
+
+# Notice the timestamp using timezone that is current for the evaluating process
+notice($timestamp.strftime('%F %T', 'CURRENT')) # outputs the timestamp using the timezone for the current process
```
hlindberg: This has 'CURRENT' in upper case. Elsewhere the same was stated in lower case 'current' - which is it (specific case, or case independent)?

thallgren: It is what the documentation states, case independent ;-)

PR #5294, opened 2016-09-23
(PUP-6670) Revert Revert Update tag validation regex to handle newlines

McdonaldSeanp:
This reverts commit 862b519b75a77ec904063ba8d0ee3c1cb29f62cf which reverted
06c710a2b227499d7d4710d5884a0f9d218e6467 since we did not want to ship that in
the 1.7.0 release, now that 1.7 has been released the original commit is good
to go back


branan:
That looks like the diff in the original commit. :+1: from me

PR #5296, opened 2016-09-23
(PUP-6738) Update version to 4.7.1 post release of 4.7.0

McdonaldSeanp:
After the agent 1.7 release, the version for the next puppet needs to be
bumped


branan:
:+1:

PR #5299, opened 2016-09-26
(PUP-6742) Add YARD tags for log functions

whopper:
Because the 8 puppet log functions are dynamically created,
YARD (and thus puppet-strings) cannot document them like normal
puppet functions. To remedy this, this commit inserts several
`@!parse` directives to give YARD hints as to what it needs
to generate documentation for.

### DIFF
--- a/lib/puppet/pops/loader/static_loader.rb
+++ b/lib/puppet/pops/loader/static_loader.rb
@@ -99,6 +99,81 @@ class StaticLoader < Loader
     @loaded[typed_name]
   end
 
+  # The following comments are used to document the metaprogrammed logging functions created
+  # in create_logging_functions below.
+
+  # @!parse
+  #   # Log a message on the server at level debug.
+  #   Puppet::Functions.create_function(:debug) do
+  #     # @param [String] message
+  #     # @return [Undef]
+  #     def debug(message)
+  #       Puppet::Util::Log.create({:level => :debug, :source => scope, :message => message})
+  #     end
+  #   end
+  # @!parse
+  #   # Log a message on the server at level info.
+  #   Puppet::Functions.create_function(:info) do
+  #     # @param [String] message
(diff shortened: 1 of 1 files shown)


whopper:
Running strings against this results in the following docs:
screen shot 2016-09-26 at 4 13 58 pm


The current (hand-written) docs look like this:
screen shot 2016-09-26 at 4 14 07 pm

Since this is using the 4.x API, `type => statement` no longer applies.

scotje:
Well this ended up being pretty clean, 👍  from me.

thallgren:
Shouldn't those functions be documented with a parameter (the message) ?

whopper:
@thallgren makes sense. I also added the call to `Puppet::Util::Log.create` so the `source` section of the docs would be accurate. This is definitely a bit clunkier, but it results in the following doc:

screen shot 2016-09-27 at 1 41 29 pm

Note that I don't think there's anyway we can tell what `scope` will be there.

thallgren:
The documentation looks a lot better now, but unless I'm guessing wrong, the only reason the `#create_logging_functions` was added in the first place, was to avoid declaring very similar functions again and again, and also automatically get a new function in case a new level was added. If that instead results in the code being duplicated as documentation, then it no longer fills its purpose. It's rather the opposite really. The documentation comes out alright, but maintaining this kind of code, and remembering why decisions like these were taken, becomes a nightmare.

I suggest dropping the `#create_logging_functions` altogether, and instead turn each log-level into a regular function in a file under `lib/puppet/functions`.

thallgren:
After proposing the above, it struck me that there might be some bootstrapping issue preventing it, so I had to try it myself :-). It worked OK though, so I added [PR-5302](/puppetlabs/puppet/pull/5302) to the ticket in hope that the documentation effort for those functions can be based on that instead.

PR #5301, opened 2016-09-28
(PUP-6749) race condition smf and manifest-import

shawnferry:
Package installs which update multiple manifests may fail later
resources because the affected services do not exist until manifest
import has completed.

Solaris 11.2 or newer pkg install and update operations will wait
for actuators to complete after this change.

23107546 race condition with smf provider and manifest-import

### DIFF
--- a/lib/puppet/provider/package/pkg.rb
+++ b/lib/puppet/provider/package/pkg.rb
@@ -203,10 +203,14 @@ Puppet::Type.type(:package).provide :pkg, :parent => Puppet::Provider::Package d
     else
       command = 'update'
     end
+    args = ['--accept']
+    if Puppet::Util::Package.versioncmp(Facter.value(:operatingsystemrelease), '11.2') >= 0
+      args.push('--sync-actuators-timeout', '900')
+    end
     unless should.is_a? Symbol
       name += "@#{should}"
     end
-    r = exec_cmd(command(:pkg), command, '--accept', name)
+    r = exec_cmd(command(:pkg), command, *args, name)
     return r if nofail
     raise Puppet::Error, "Unable to update #{r[:out]}" if r[:exit] != 0
   end
--- a/spec/unit/provider/package/pkg_spec.rb
+++ b/spec/unit/provider/package/pkg_spec.rb
@@ -201,105 +201,115 @@ describe Puppet::Type.type(:package).provider(:pkg) do
     end
 
     context ":install" do
-      it "should accept all licenses" do
-        provider.expects(:query).with().returns({:ensure => :absent})
-        Puppet::Util::Execution.expects(:execute).with(['/bin/pkg', 'install', '--accept', 'dummy'], {:failonfail => false, :combine => true}).returns ''
-        Puppet::Util::Execution.expects(:execute).with(['/bin/pkg', 'unfreeze', 'dummy'], {:failonfail => false, :combine => true}).returns ''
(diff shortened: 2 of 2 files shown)


shawnferry:
If you can ignore whitespace this is much less of a change.

MosesMendoza:
@shawnferry thank you for the contribution!  

What are your thoughts on pros/cons of using the timeout version of sync actuators, `--sync-actuators-timeout`,  ie from the docs:
```
--sync-actuators
When you specify this option, the pkg command will not return until all SMF actuators
have finished in the zone in which pkg was invoked (the global zone or a non-global zone).
--sync-actuators-timeout timeout
When you specify this option, the pkg command will not return until all SMF actuators
have finished or the timeout period is reached, whichever is shorter. If the actuators do not
finish within the given timeout in seconds, the pkg command continues operation and exits
with return code 8.
```

Perhaps if we specify some sane timeout (I don't know what that would be) we could prevent puppet hanging in the event of a smf actuator that doesn't return? Not sure what the general thinking is around timeouts in puppet (cc @hlindberg)

drewfisher314:
The problem this bug is fixing is when 'pkg install <something>' delivers an SMF manifest to /lib/svc/manifest/*, it restarts the manifest-import service asynchronously.  For packages with a small number of SMF manifests, manifest-import will usually finish before 'pkg install <something>' does and no harm will come.  The issue is for when pkg install delivers a very large number of SMF manifests (such as 'pkg install openstack' which has 80 SMF manifests) are delivered in a single run.  What happens is something like ...

package { 'cloud/openstack':
  ensure => installed,
}

service { 'application/openstack/keystone/keystone:default':
  ensure => running,
  require => Package['cloud/openstack'],
}

... *may* fail because the keystone service has not yet been registered with SMF. 

I don't know of a good "default" timeout for --sync-actuators-timeout <timeout> that accounts for slower systems (VMs, zones, etc.).  I've personally seen manifest-import take 2-3 minutes on a smaller zone, but I could easily see it taking longer if the block device is over iSCSI and a slow 100mb connection or something (or under heavy I/O load, etc.)

shawnferry:
@hlindberg @MosesMendoza I had thought about setting a timeout but I didn't have any real feeling for what it should be, given that the problem we are addressing is specifically caused by returning too soon I opted against it. 

Is there a very strong opinion on having one? It could also be something seemingly ludicrous like 30 minutes.

MosesMendoza:
Thanks @drewfisher314, that makes sense to me.

@shawnferry yeah my thought would be something intentionally pretty high, ie just to catch the extreme case that something has hung and isn't coming back. I was going to suggest something like 10 minutes based on @drewfisher314's comment but I wish we had more data around what a reasonable long case is. I think we may want less than 30 minutes because the default run interval for Puppet is 30 minutes and if a run is in progress it won't start another one causing puppet to start skipping runs maybe?

shawnferry:
@MosesMendoza okay, I'm calling it 15 minutes.

MosesMendoza:
👍  on this pr

MosesMendoza:
ah good catch @drewfisher314, I hadn't realized these were mutually exclusive. when I pulled this down to test it appeared to work fine but I suppose it could have been silently ignoring sync-actuators-timeout

shawnferry:
Updated to have only the timeout...unfortunately I need to force the push so the recommended changes view is lost.

MosesMendoza:
thanks for the contribution @shawnferry!

PR #5302, opened 2016-09-28
(PUP-6742) Replace auto-generation of logging functions

thallgren:
This commit replaces the #create_logging_functions with proper Puppet
functions so that they can be documented separately.


whopper:
@thallgren huh! We were under the impression that this wouldn't work, but 👍  if it does. So these functions don't need to be created within the static loader? Will they be loaded in the same way as they were before? /cc @hlindberg

whopper:
Thanks @thallgren! I opened https://github.com/puppetlabs/puppet/pull/5304 which is this commit and another on top to add @param and @return tags (as strings will warn if they aren't present).

PR #5303, opened 2016-09-28
(PUP-6754) Allow uppercase CLI short flags

holguinj:
This commit changes the regex pattern that validates short CLI options so that flags like `-F` and `-V` will be allowed, rather than only `-f` and `-v`.

### DIFF
--- a/spec/unit/interface/option_spec.rb
+++ b/spec/unit/interface/option_spec.rb
@@ -15,6 +15,11 @@ describe Puppet::Interface::Option do
       end
     end
 
+    it "should allow an uppercase short form" do
+      option = Puppet::Interface::Option.new(face, "-F json", "--format json")
+      expect(option.name).to eq(:format)
+    end
+
     [:foo, 12, nil, {}, []].each do |input|
       it "should fail sensible when given #{input.inspect}" do
         expect {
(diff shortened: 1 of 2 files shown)


ajroetker:
Looks like the only issue has to do with travis tests that don't allow for the ticket in the commit to be from other projects than PUP https://travis-ci.org/puppetlabs/puppet/jobs/163459888

holguinj:
Moved the ticket to PUP and updated.

spec/unit/interface/option_spec.rb
```diff
@@ -15,6 +15,11 @@
       end
     end
 
+    it "should allow an uppercase short form" do
+      option = Puppet::Interface::Option.new(face, "-F json", "--format json")
```
haus: Is there an easy way to validate that passing the face -F actually results in `:format` being set?

holguinj: not that I could figure out. FWIW, I did try it out with an existing puppet face (after these changes) and it worked.

haus: yay

MosesMendoza:
Is the motivation for this change that the face in question has already purposed the lowercase version of a desired flag? If so perhaps it's also worth considering using a different short flag than the first character of the long flag? Such a thing isn't without precedence - for example from the module face:
```
  --target-dir DIR | -i DIR      - The directory into which modules are
                                   installed.
```

holguinj:
This isn't worth the trouble at this point so I closed the ticket won't fix.

PR #5304, opened 2016-09-28
(PUP-6742) Separate logging functions

whopper:
These commits replace the #create_logging_functions with proper Puppet
functions so that they can be documented separately.


whopper:
screen shot 2016-09-28 at 9 51 30 am

PR #5306, opened 2016-09-28
(PUP-6756) Puppet::FileBucket::Dipper specs fail silently

shawnferry:
dipper_spec assumes gnu diff AND executes diff via ``s
notably this fails on Solaris but may also fail on other platforms.

On Solaris 11+ dipper_spec now invokes 'gdiff' and on all platforms
uses Puppet::Util::Execution.execute which causes the test to fail
if diff cannot be executed.

### DIFF
--- a/spec/unit/file_bucket/dipper_spec.rb
+++ b/spec/unit/file_bucket/dipper_spec.rb
@@ -91,15 +91,24 @@ describe Puppet::FileBucket::Dipper, :uses_checksums => true do
         end
 
         it "should properly diff files on the filebucket" do
-          file1 = make_tmp_file("OriginalContent")
-          file2 = make_tmp_file("ModifiedContent")
+          file1 = make_tmp_file("OriginalContent\n")
+          file2 = make_tmp_file("ModifiedContent\n")
           @dipper = Puppet::FileBucket::Dipper.new(:Path => tmpdir("bucket"))
           checksum1 = @dipper.backup(file1)
           checksum2 = @dipper.backup(file2)
 
           # Diff without the context
-          diff12 = `diff -uN #{file1} #{file2} | sed '1,2d'`
-          diff21 = `diff -uN #{file2} #{file1} | sed '1,2d'`
+          # Lines we need to see match 'Content' instead of trimming diff output filter out
+          # surrounding noise...or hard code the check values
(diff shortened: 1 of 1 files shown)


whopper:
Thanks for the contribution @shawnferry! Would you mind squashing your commits together just to keep the git history clean for this fix?

shawnferry:
@whopper all set

joshcooper:
Thanks for your contribution @shawnferry. I think your changes improve the current test and I'm happy to merge as is. I did have a couple of ideas for improvements, either in this ticket or a different one if you're up for it:

1. We try to avoid executing commands in unit tests, so I'd recommend stubbing out the execute calls in the dipper unit test, e.g. `Puppet::Util::Execution.stubs(:execute).with(regexp_matches(/#{file1} #{file2}/).returns("some expected diff output")`.
1. It sounds like puppet's default `diff` and `diff_args` settings should be different for Solaris 11 and up? For example, on AIX we use different [`diff_args`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L3-L8). And on Windows, we change the default diff command to empty string since there isn't a [suitable command installed by default](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L367).
1. If we change the defaults for Solaris, it would be a good idea to add an integration test for `Puppet::Util::Diff` to ensure it when it executes the diff command and args, it returns the expected output on different platforms.

shawnferry:
@joshcooper I'd happily replace that execution with the expected output. It was my first inclination but I didn't understand why the test was explicitly using `diff` and munging the output so I just cleaned up the behavior. 

The test in question has never been valid for Solaris it was just failing silently (dumping noise to stderr) because the invocation error wasn't being caught. 

Using the stock/default diff seems to be fine it was just the inline execution in the test assumed gnu diff options and I didn't spend enough time digging to see that it's just executing diff under the covers.

shawnferry:
@joshcooper looking further into this the vast majority of the tests are doing a mess of integration style testing. Just stubbing out the return here isn't going to be enough and I can't currently commit to refactoring it all out into integration tests at the moment. 

I created https://tickets.puppetlabs.com/browse/PUP-6769 for the remaining work.

ScottGarman:
:+1:

PR #5307, opened 2016-09-29
(PUP-6753) Ensure that log functions return undef

thallgren:
This commit changes the helper method for all puppet log functions so
that it returns `nil`. A return_type of 'Undef' is also added to the
dispatcher for the log functions

### DIFF
--- a/lib/puppet/functions/alert.rb
+++ b/lib/puppet/functions/alert.rb
@@ -5,6 +5,7 @@ Puppet::Functions.create_function(:alert, Puppet::Functions::InternalFunction) d
   dispatch :alert do
     scope_param
     repeated_param 'Any', :values
+    return_type 'Undef'
   end
 
   def alert(scope, *values)
--- a/lib/puppet/functions/crit.rb
+++ b/lib/puppet/functions/crit.rb
@@ -5,6 +5,7 @@ Puppet::Functions.create_function(:crit, Puppet::Functions::InternalFunction) do
   dispatch :crit do
     scope_param
     repeated_param 'Any', :values
+    return_type 'Undef'
   end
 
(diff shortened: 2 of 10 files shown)


whopper:
Actually, before I merge: is this a breaking change? It seems unlikely that many would be using the return value of these functions, but is it possible that someone is?

thallgren:
The returned object, when auto converted to a String, yields it's message. This statement compiles without problems:
```
notify { 'hello': message => notice('world') }
```
and the catalog contains the property 'message' => 'world' (the value is converted to a string). It's unlikely, but who knows?

hlindberg:
Could possibly affect Ruby code, if someone calls these functions from templates and they have figured out that they get a string. 
Not sure what to do, an alternative is to do a `to_s` and return String - then it is at least compliant in terms of what is meaningful to do from Puppet.
Change to return String?

thallgren:
That's an option. Another would be to simply leave things as they are and introduce this change in 5.0.

hlindberg:
While this is in theory a breaking change since puppet during 4.x have been leaking the Log object into Puppet, this would not have been the case earlier because the log functions where all of the :statement kind and from the 3.x Puppet language you could not get a value from them. Since we never told anyone or specified that they return a log the risk is minimal. While technically you could have called these functions to log from Ruby, there are easier ways to do so (there is no code in Puppet that expects the return to be a Log). Changing to Undef return type is the sane thing to do, and I consider this to be a bug.

If we take the super conservative approach; we could specify that the function returns a Log, and add that to the type system. But that is a lot of work for a non valuable feature.
I am leaning towards merging as is (that is, that `undef` is returned).

PR #5308, opened 2016-09-29
(PUP-6757) Make Binary use strict base64 encoding by default

thallgren:
This commit changes the default encoding used by the Binary constructor
from '%b' (relaxed base64) to '%B' (strict base64).

### DIFF
--- a/lib/puppet/parser/functions/new.rb
+++ b/lib/puppet/parser/functions/new.rb
@@ -896,13 +896,13 @@ The formats have the following meaning:
 
 | format | explanation |
 | ----   | ----        |
-| b | The data is in base64 encoding, padding as required by base64 strict is added by default
-| u | The data is in URL safe base64 encoding
 | B | The data is in base64 strict encoding
+| u | The data is in URL safe base64 encoding
+| b | The data is in base64 encoding, padding as required by base64 strict, is added by default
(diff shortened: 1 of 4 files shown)


joshcooper:
Requiring strict base64 encoding makes sense to me, but why would we allow relaxed? That seems like a recipe for disaster, especially if the data is coming from an external system, so it may not be obvious that the data is invalid.

thallgren:
@joshcooper, this ticket is about enforcing strict behavior. Allowing relaxed will then become a very conscious decision in that the format '%b' must be specified. Perhaps '%b' isn't particularly useful, but I think that whether or not it should be included is a separate issue.

joshcooper:
@thallgren yep makes sense. I just wanted to ask the question because '%b' doesn't seem useful to me. More concerning, is it would allow garbage data to get into the system, and would produce unexpected results as bits are silently dropped. It also means the user must make a choice '%B' vs '%b' and potentially get it wrong without understanding the consequences. But I agree the issue of removing '%b' shouldn't block this PR.

hlindberg:
As I understand it, the "relaxed" mode is actually defined by RFC 2045 (MIME Base64 encoding) and that format is likely in use in some places. Have no idea if RFC 2045 is of value to support.
The strict format is RFC 4648.

PR #5309, opened 2016-09-30
(PUP-6663) quote query format strings in rpm provider 

exodusftw:
When running large volumes of rpm queries as the root user to determine package state, locks are placed on the rpm database that can become stale and prevent package installation/removal through either the rpm or yum providers.

The fix to prevent stale locks from occurring in situations of large query volumes appears to be properly quoting the query format string (represented as NEVRA_FORMAT) in the rpm provider.


whopper:
@exodusftw sorry for the super long delay in response. If you rebase this on the head of master, travis should go green. I think this change is the right thing to do, but I'll run it through our acceptance tests to make sure. Coincidentally, I noticed that the nevra hash wasn't quoted while looking into another issue the other day, and I think that actually breaks older versions of RPM. This may solve both problems.

exodusftw:
@whopper thanks for the follow-up - I rebased and it looks to be good now.

whopper:
Thanks @exodusftw, running tests now.

exodusftw:
thanks @whopper

pkranenburg:
1. Can anybody please explain *why* putting quotes around the queryformat string would mitigate the stale lock problem.

2. The addition of the quotes changes the output format of the 'rpm query' command, which is not reflected in the NEVRA_REGEXP string. This will cause package names to be recorded with an extra single quote (') prepended. This in turn will prevent 'rpm erase' from working, as evidenced in the failure of "package { '<amything>': ensure => absent }"...

exodusftw:
@pkranenburg - The issue seems to stem from rpm queries initiated by the root user where a lock is placed on the various __db.00X files in /var/lib/rpm.

For Example, if you strace a query from a root/non-root user, you will see:

```bash
// Non-root user strace of an "rpm -qa"
# grep db.00 /tmp/rpm.non-root.strace 
#

// root user strace of an "rpm -qa"
# grep db.00 /tmp/rpm.root.strace 
access("/var/lib/rpm/__db.001", F_OK)   = 0
open("/var/lib/rpm/__db.001", O_RDWR)   = 3
open("/var/lib/rpm/__db.001", O_RDWR)   = 3
open("/var/lib/rpm/__db.002", O_RDWR)   = 3
open("/var/lib/rpm/__db.003", O_RDWR)   = 3
open("/var/lib/rpm/__db.004", O_RDWR)   = 3
access("/var/lib/rpm/__db.001", F_OK)   = 0
```

When there are large volumes of unquoted RPM queries as would be the case during a standard Puppet run where a sizable quantity of package resources are being managed, each query places a lock on the RPMDB that can create strange race conditions that appear to lead/contribute to errors such as:

```bash
# rpm -qa kernel
rpmdb: Thread/process ######/############## failed: Thread died in Berkeley DB library
error: db3 error(-30974) from dbenv->failchk: DB_RUNRECOVERY: Fatal error, run database recovery
error: cannot open Packages index using db3 -  (-30974)
error: cannot open Packages database in /var/lib/rpm
rpmdb: Thread/process ######/############## failed: Thread died in Berkeley DB library
error: db3 error(-30974) from dbenv->failchk: DB_RUNRECOVERY: Fatal error, run database recovery
error: cannot open Packages database in /var/lib/rpm

package db4 is not installed
rpmdb: Lock table is out of available lock entries
error: cannot open Packages index using db3 - Cannot allocate memory (12)
error: cannot open Packages database in /var/lib/rpm
rpmdb: Lock table is out of available lock entries
error: cannot open Packages database in /var/lib/rpm
```

The quote was added to ensure the query format adheres to the recommendation [here](http://rpm.org/user_doc/query_format.html) - the implication being that the non-terminated query format string can cause the lock to never properly release.

That said, this is obviously not the desired outcome with the prepended `'`:

It appears @heliocentric (:beers:) has a pending PR to `.gsub` this from the NEVRA_REGEX string at: https://github.com/puppetlabs/puppet/pull/5639

PR #5311, opened 2016-10-03
(PUP-6766) Don't depend on log function return value in tests

whopper:
This commit removes the assumption that the 'notice' log function
will return anything in the calling_all_functions tests, as those
functions now return Undef (as they always should have).

[skip ci]


kylog:
LGTM, let's give it a whirl ...

PR #5312, opened 2016-10-04
(PUP-6752) Fix diff in catalog lookup caused by app_management flag

thallgren:
Before this lookup, the code-paths for how a lookup was performed
differed slightly depending on the setting of the `app_management` flag.
While the setting should extend the lookup using the capability finder,
a malformed boolean expression made it have an additional and undesired
side effect.

This commit ensures that the code path is exactly the same until the
point where the decision is made to search for the capability.


hlindberg:
App Veyor errors seems legit

PR #5314, opened 2016-10-05
(PUP-6417) improve regex of line matching in authorized_keys

petems:
* Lines in ssh authorized_keys file may have mulitple space as filed
separator. We need to use \s+ in REGEX instead of plain space.
* Adds spec to detect multiple spaces in ssh line

#5031 rebased with a spec as requested

@kylog can I get a merge or do we need more unit tests? 💃


kylog:
Looks legit @petems. Thanks for picking this one up!

PR #5315, opened 2016-10-05
(PUP-6774) Add test for ext_parameters containing hashes and arrays rich data

thallgren:
This PR adds a test that verifies that arrays and hashes containing rich_data can be serialized in a catalog. It also contains a maint commit to fix a problem with unbalanced brackets in the serializer that was found when the test was written.


hlindberg:
Ideally, the (maint) commit should have had its own ticket

PR #5316, opened 2016-10-05
(PUP-6784) http_proxy specs fail if proxy is defined in test env

shawnferry:
If http_proxy is defined in the test environment specs fail due to
precedence of http_proxy over HTTP_PROXY.

With this change http_proxy is explicitly set to nil before tests as
needed.

### DIFF
--- a/spec/unit/forge/repository_spec.rb
+++ b/spec/unit/forge/repository_spec.rb
@@ -6,6 +6,11 @@ require 'puppet/forge/cache'
 require 'puppet/forge/errors'
 
 describe Puppet::Forge::Repository do
+  before(:all) do
+    # any local http proxy will break these tests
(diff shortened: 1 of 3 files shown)


shawnferry:
Are there any recommendations for how to handle this on windows vs *nix other than changing the spec per-type? The behavior for proxy detection appears to be inconsistent.

What I wanted was to be able to use stub_const from RSpec::Mocks to just mask ENV but a way  to do that while using moca is evading me.

MosesMendoza:
👍

shawnferry:
@MosesMendoza *ping* is there anything else this needs?
Also I just fixed a comment typo

MosesMendoza:
thanks for the ping @shawnferry, this fell off the radar

PR #5318, opened 2016-10-05
(PUP-6798) Yum provider family ensuring latest

petems:
* Previously, when ensuring latest the logic used to implement PUP-6324
didn’t take ensuring latest into account, just specifying a new version:

```
Debug: /Stage[main]/Main/Package[cockpit]/ensure: cockpit "0.55-1.fc22" is installed, latest is "0:0.67-2.fc22"
Debug: Package[cockpit](provider=dnf_update): Ensuring => latest
Debug: Executing: '/usr/bin/dnf -d 0 -e 1 -y install cockpit'
Notice: /Stage[main]/Main/Package[cockpit]/ensure: ensure changed '0.55-1.fc22' to '0:0.67-2.fc22'
Debug: /Stage[main]/Main/Package[cockpit]: The container Class[Main] will propagate my refresh event
```
* This meant the system reported that the package was updated, but it ran install rather than upgrade, which does nothing and exits zero
* Added new logic for when latest is used, add new spec, all working now!

### DIFF
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -156,6 +156,16 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
     operation = :install
 
     case should
+    when :latest
+      current_package = self.query
+      if current_package && !current_package[:ensure].to_s.empty?
+        operation = update_command
+        self.debug "Ensuring latest, so using #{operation}"
+      else
+        self.debug "Ensuring latest, but package is absent, so using #{:install}"
+        operation = :install
+      end
+      should = nil
     when true, false, Symbol
       # pass
       should = nil
--- a/spec/shared_examples/rhel_package_provider.rb
+++ b/spec/shared_examples/rhel_package_provider.rb
@@ -125,6 +125,22 @@ shared_examples "RHEL package provider" do |provider_class, provider_name|
         provider.stubs(:query).returns(:ensure => current_version).then.returns(:ensure => version)
         provider.install
       end
+      it 'should not run upgrade command if absent and ensure latest' do
+        current_version = ''
+        version = '1.2'
+        resource[:ensure] = :latest
+        Puppet::Util::Execution.expects(:execute).with(["/usr/bin/#{provider_name}", '-d', '0', '-e', error_level, '-y', :install, name])
+        provider.stubs(:query).returns(:ensure => current_version).then.returns(:ensure => version)
+        provider.install
+      end
+      it 'should run upgrade command if present and ensure latest' do
+        current_version = '1.0'
+        version = '1.2'
+        resource[:ensure] = :latest
+        Puppet::Util::Execution.expects(:execute).with(["/usr/bin/#{provider_name}", '-d', '0', '-e', error_level, '-y', upgrade_command, name])
+        provider.stubs(:query).returns(:ensure => current_version).then.returns(:ensure => version)
+        provider.install
+      end
       it 'should accept install options' do
         resource[:ensure] = :installed
         resource[:install_options] = ['-t', {'-x' => 'expackage'}]


lib/puppet/provider/package/yum.rb
```diff
@@ -156,6 +156,11 @@ def install
     operation = :install
 
     case should
+    when :latest
+      # use update_command when using latest for updating packages
+      operation = update_command
```
MikaelSmith: What if the package isn't installed yet and you're using latest?

MikaelSmith: Also, isn't the code at line 175/180 supposed to handle that? Why isn't it working? Oh, because latest doesn't compare correctly.

This should at least check that `self.query` returns a value. This seems like the logic should be moved down to the else block, and be checked as part of version comparison.

spec/shared_examples/rhel_package_provider.rb
```diff
@@ -125,6 +125,14 @@
         provider.stubs(:query).returns(:ensure => current_version).then.returns(:ensure => version)
         provider.install
       end
+      it 'should run upgrade command if present and ensure latest' do
+        current_version = '1.0'
+        version = '1.2'
+        resource[:ensure] = :latest
+        Puppet::Util::Execution.expects(:execute).with(["/usr/bin/#{provider_name}", '-d', '0', '-e', error_level, '-y', upgrade_command, name])
```
MikaelSmith: It'd be great to also have a test for when the package is absent, to ensure `install` is still used.

lib/puppet/provider/package/yum.rb
```diff
@@ -156,6 +156,11 @@ def install
     operation = :install
 
     case should
+    when :latest
```
whopper: @petems (Re contributor summit) I figured out where this is going wrong! :D 

We had `when :latest && current_package`, but since this is a case statement, that will always evaluate to false. If we turn this into an if statement instead, this will work with your new tests. See also comment on test about what needs to change there.

spec/shared_examples/rhel_package_provider.rb
whopper: Re contributor summit, we discovered that when upgrading a package with `:latest`, puppet will just call `upgrade <package>` *without* the version

lib/puppet/provider/package/yum.rb
```diff
@@ -155,7 +155,18 @@ def install
     self.debug "Ensuring => #{should}"
     operation = :install
 
+    current_package = self.query
```
MikaelSmith: I'm not a fan of doing this when it's not necessary (i.e. `when true, false, Symbol`).

petems: This was already in from 3180b9d9b2c844dade1d361326600f7001ec66dd, whats a better way?

lib/puppet/provider/package/yum.rb
```diff
@@ -155,7 +155,18 @@ def install
...
+
     case should
+    when :latest
+      if current_package[:ensure]
+        self.debug "Ensuring latest, so using #{operation}"
```
petems: Fixed 👍

lib/puppet/provider/package/yum.rb
MikaelSmith: Duplicate it in the `when :latest` block :|

MosesMendoza:
Thanks @petems! Did some manual validation on fedora 24. Prior to this change:
```
puppet apply -e " package { 'htop' : ensure => latest } " --debug --trace
...
Debug: /Stage[main]/Main/Package[htop]/ensure: htop "1.0.3-4.fc22" is installed, latest is "0:1.0.3-6.fc23"
Debug: Package[htop](provider=dnf): Ensuring => latest
Debug: Executing: '/usr/bin/dnf -d 0 -e 1 -y install htop'
Debug: /Stage[main]/Main/Package[htop]/ensure: htop "1.0.3-4.fc22" is installed, latest is "0:1.0.3-6.fc23"
Notice: /Stage[main]/Main/Package[htop]/ensure: ensure changed '1.0.3-4.fc22' to '0:1.0.3-6.fc23'
...
[root@ezhsuf4bmw40s50 ~]# dnf list installed htop
Last metadata expiration check: 0:08:31 ago on Thu Oct 20 16:56:18 2016.
Installed Packages
htop.x86_64                                                                             1.0.3-4.fc22                                                                              @localmirror-everything-22
```
😞 

with this change:
```
[root@ezhsuf4bmw40s50 ~]# puppet apply -e " package { 'htop' : ensure => latest } " --debug --trace
...
Debug: /Stage[main]/Main/Package[htop]/ensure: htop "1.0.3-6.fc23" is installed, latest is "0:2.0.1-1.fc24"
Debug: Package[htop](provider=dnf): Ensuring => latest
Debug: Executing: '/usr/bin/rpm -q htop --nosignature --nodigest --qf %{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n'
Debug: Package[htop](provider=dnf): Ensuring latest, so using upgrade
Debug: Executing: '/usr/bin/dnf -d 0 -e 1 -y upgrade htop'
Notice: /Stage[main]/Main/Package[htop]/ensure: ensure changed '1.0.3-6.fc23' to '0:2.0.1-1.fc24'
...
[root@ezhsuf4bmw40s50 ~]# dnf list installed htop
Last metadata expiration check: 0:08:47 ago on Thu Oct 20 16:56:18 2016.
Installed Packages
htop.x86_64                                                                                2.0.2-1.fc24                                                                                 @localmirror-updates
```
😄 

When ensuring a specific version to upgrade to (even one that's not "latest"):
```
root@ezhsuf4bmw40s50 ~]# puppet apply -e " package { 'htop' : ensure => '1.0.3-6.fc23.x86_64' } " --debug --trace
...
ug: Package[htop](provider=dnf): Ensuring => 1.0.3-6.fc23.x86_64
Debug: Package[htop](provider=dnf): Detected Arch argument in package! - Moving arch to end of version string
Debug: Executing: '/usr/bin/rpm -q htop --nosignature --nodigest --qf %{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n'
Debug: Package[htop](provider=dnf): Upgrading package htop from version 1.0.3-4.fc22 to 1.0.3-6.fc23.x86_64
Debug: Executing: '/usr/bin/dnf -d 0 -e 1 -y upgrade htop-1.0.3-6.fc23.x86_64'
Debug: Executing: '/usr/bin/rpm -q htop --nosignature --nodigest --qf %{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\n'
Debug: Executing: '/usr/bin/dnf check-update'
Debug: /Stage[main]/Main/Package[htop]/ensure: htop "1.0.3-4.fc22" is installed, latest is "0:2.0.1-1.fc24"
Notice: /Stage[main]/Main/Package[htop]/ensure: ensure changed '1.0.3-4.fc22' to '1.0.3-6.fc23.x86_64'
...
[root@ezhsuf4bmw40s50 ~]# dnf list installed htop
Last metadata expiration check: 0:11:57 ago on Thu Oct 20 16:56:18 2016.
Installed Packages
htop.x86_64                                                                             1.0.3-6.fc23                                                                              @localmirror-everything-23
```
👍

PR #5319, opened 2016-10-06
(PUP-6799) Allow data with unknown type in the catalog

thallgren:
Before this commit, data that wasn't stricly compatible with JSON would
generate an error when rich_data was set to false. That was a regression.
This commit ensures that such data will be passed through and instead
print a warning.

### DIFF
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -88,6 +88,9 @@ class Puppet::Resource
         value = Puppet::Resource.value_to_pson_data(value)
         if is_json_type?(value)
           params[param] = value
+        elsif json_serializer.nil?
+          Puppet.warning("Resource '#{to_s}' contains a #{value.class.name} value. It will be converted to the String '#{value}'")
+          params[param] = value
         else
           ext_params[param] = value
         end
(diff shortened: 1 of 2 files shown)


whopper:
👍 I just verified that this fixes the issue we had with fresh PE installs. The first agent run no longer fails

whopper:
The appveyor failure is odd. It looks like all of the tests passed but there were some bundler / environment issues.

```
Failed to load feature test for rack: cannot load such file -- rack
 exited with code -1073741819
File not found: C:\projects\puppet\spec_order.txt
Command exited with code 2
```

lib/puppet/resource.rb
```diff
@@ -88,6 +88,9 @@ def to_data_hash(json_serializer = nil)
         value = Puppet::Resource.value_to_pson_data(value)
         if is_json_type?(value)
           params[param] = value
+        elsif json_serializer.nil?
+          Puppet.warning("Resource '#{to_s}' contains a #{value.class.name} value. It will be converted to the String '#{value}'")
```
hlindberg: I am not sure that is true for all serialization formats - think PSON may allow symbols

hlindberg: but maybe the changes we have made makes it so that the warning is true

thallgren: I tried. Symbols are accepted, but they are converted to strings. They do not retain the type. This is what you get with PSON:
```
2.1.5 :001 > require 'puppet'
 => true 
2.1.5 :002 > PSON.parse({ :x => 'hello' }.to_pson)
 => {"x"=>"hello"}
2.1.5 :003 > PSON.parse({ :x => :undef }.to_pson)
 => {"x"=>"undef"} 
```

hlindberg: Thanks

PR #5320, opened 2016-10-06
(pup-6592) acceptance sensitive data fix

er0ck:
*  the previous commit did not take into account windows32 pathing
 differences when abstracted into cygwin.  This change should allow the
 test to run in both 64/32 bit windowses.

this will need a merge-up to master


whopper:
👍  @branan look good to you?

PR #5321, opened 2016-10-07
(PUP-6768) Add Pcore Object type Puppet::LookupContext

thallgren:
This PR adds a new Pcore Object type named `Puppet::LookupContext`
with attributes and methods.

### DIFF
--- a/lib/puppet/pops/lookup/context.rb
+++ b/lib/puppet/pops/lookup/context.rb
@@ -24,6 +24,7 @@ module Puppet::Pops
             'cache_all' => tf.callable([tf.hash_kv(tf.scalar, tf.any)], tf.undef),
             'cached_value' => tf.callable([tf.scalar], tf.any),
             'cached_entries' => tf.variant(
+              tf.callable([0, 0, tf.callable(1,1)], tf.undef),
               tf.callable([0, 0, tf.callable(2,2)], tf.undef),
               tf.callable([0, 0], tf.iterable(tf.tuple([tf.scalar, tf.any])))
             )
@@ -55,12 +56,16 @@ module Puppet::Pops
         @cache[key]
       end
 
-      def cached_entries
+      def cached_entries(&block)
         @cache
         if block_given?
           enumerator = @cache.each_pair
           @cache.size.times do
-            yield(*enumerator.next)
+            if block.arity == 2
+              yield(*enumerator.next)
+            else
+              yield(enumerator.next)
+            end
           end
           nil
         else
--- a/lib/puppet/pops/lookup/context.rb
+++ b/lib/puppet/pops/lookup/context.rb
@@ -1,0 +1,81 @@
+module Puppet::Pops
+  module Lookup
+    class Context
+      include Types::PuppetObject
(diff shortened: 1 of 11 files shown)


lib/puppet/pops/lookup/context.rb
```diff
@@ -0,0 +1,81 @@
...
+            }
+          },
+          {
+            'not_found' => tf.callable([0, 0], tf.undef),
+            'explain' => tf.callable([0, 0, tf.callable(0,0)], tf.undef),
+            'cache' => tf.callable([tf.scalar, tf.any], tf.undef),
+            'cache_all' => tf.callable([tf.hash_kv(tf.scalar, tf.any)], tf.undef),
+            'cached_value' => tf.callable([tf.scalar], tf.any),
+            'cached_entries' => tf.variant(
+              tf.callable([0, 0, tf.callable(2,2)], tf.undef),
```
hlindberg: If it is to mimic each, then also support 1 arg in the lambda to get each key/value tuple ?

spec/lib/puppet_spec/unindent.rb
```diff
@@ -0,0 +1,5 @@
+class String
+  def unindent
```
hlindberg: isn't there a risk in monkey patching String this way when running tests?

spec/unit/pops/lookup/context_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+    it 'can use the cache to merge hashes and individual entries' do
+      code = <<-PUPPET.unindent
+        $ctx = Puppet::LookupContext.new('e', 'm')
+        $ctx.cache_all({ 'v1' => 'first', 'v2' => 'second' })
+        $ctx.cache('v3', 'third')
+        $ctx.cached_entries |$key, $value| { notice($key); notice($value) }
+      PUPPET
+      expect(eval_and_collect_notices(code)).to eql(%w(v1 first v2 second v3 third))
+    end
+
```
hlindberg: test missing: can replace a value for a cache key `cache('x', 10); notice cached_value('x'); cache('x', 20); notice cached_value('x');`

spec/unit/pops/lookup/context_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        $ctx.not_found
+      PUPPET
+      expect { eval_and_collect_notices(code) }.to throw_symbol(:no_such_key)
+    end
+
+    context 'when used in an Invocation' do
+      let(:invocation) { Invocation.new({}) }
+      let(:invocation_with_explain) { Invocation.new({}, {}, {}, true) }
+
+      it 'will no call explain unless explanations are active' do
```
hlindberg: typo "will no call"

spec/lib/puppet_spec/unindent.rb
thallgren: If someone else also monkey patches String with the same method and does something differently, then a lot of tests will fail of course, but I'm not sure I'd call that a risk.

PR #5322, opened 2016-10-10
PUP-6792,6793,6795-6797 Solaris Updates

shawnferry:
(PUP-6792) SMF should recursively enable services

SMF supports enabling a service and all services which that service
requires. With this patch recursive starts are the default behavior.

(PUP-6793) pkg provider should be default for Solaris 12

Puppet needs to use the pkg provider to manage packages on Solaris 12

(PUP-6795) Solaris does not support gzipped man pages

Solaris does not support gzipped man pages. Skip gzipping when
installing on Solaris.

(PUP-6796) Support setting/checking shell in user_role_add

Puppet should allow a users shell to be set in a Solaris role

(PUP-6797) service should check for smf service first

(PUP-6860) puppet on solaris 11 should support exists? for services

Attempts to manage non-existent services resulted in uninformative
errors. With this patch non-existent services return better error
messages.

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -108,9 +108,13 @@ def do_man(man, strip = 'man/')
     FileUtils.makedirs(om, {:mode => 0755, :verbose => true})
     FileUtils.chmod(0755, om)
     FileUtils.install(mf, omf, {:mode => 0644, :preserve => true, :verbose => true})
-    gzip = %x{which gzip}
-    gzip.chomp!
-    %x{#{gzip} -f #{omf}}
+    # Solaris does not support gzipped man pages. When called with
+    # --no-check-prereqs/without facter the default gzip behavior still applies
+    unless $operatingsystem == "Solaris"
+      gzip = %x{which gzip}
+      gzip.chomp!
+      %x{#{gzip} -f #{omf}}
+    end
   end
 end
 
--- a/lib/puppet/provider/package/pkg.rb
+++ b/lib/puppet/provider/package/pkg.rb
@@ -23,7 +23,7 @@ Puppet::Type.type(:package).provide :pkg, :parent => Puppet::Provider::Package d
 
   confine :osfamily => :solaris
 
-  defaultfor :osfamily => :solaris, :kernelrelease => '5.11'
+  defaultfor :osfamily => :solaris, :kernelrelease => ['5.11', '5.12']
 
   def self.instances
     pkg(:list, '-Hv').split("\n").map{|l| new(parse_line(l))}
--- a/lib/puppet/provider/service/smf.rb
+++ b/lib/puppet/provider/service/smf.rb
@@ -129,14 +129,22 @@ Puppet::Type.type(:service).provide :smf, :parent => :base do
       return
     end
 
+    # Check to see if the service exists
+    cmd = Array[command(:svccfg), "select", @resource[:name]]
+    output = Puppet::Util::Execution.execute(
+        cmd, :combine => true, :failonfail => false)
+    if $CHILD_STATUS.exitstatus != 0
+      info output
+    end
+
     begin
       # get the current state and the next state, and if the next
       # state is set (i.e. not "-") use it for state comparison
       states = service_states
       state = states[1] == "-" ? states[0] : states[1]
     rescue Puppet::ExecutionFailure
-      info "Could not get status on service #{self.name}"
-      return :stopped
+      debug "Could not get status on service #{self.name} #{$!}"
+      return :absent
     end
 
     case state
--- a/lib/puppet/provider/service/smf.rb
+++ b/lib/puppet/provider/service/smf.rb
@@ -80,7 +80,7 @@ Puppet::Type.type(:service).provide :smf, :parent => :base do
     when :maintenance, :degraded
       [command(:adm), :clear, @resource[:name]]
     else
-      [command(:adm), :enable, "-s", @resource[:name]]
+      [command(:adm), :enable, "-rs", @resource[:name]]
     end
   end
 
--- a/lib/puppet/provider/service/smf.rb
+++ b/lib/puppet/provider/service/smf.rb
@@ -129,14 +129,6 @@ Puppet::Type.type(:service).provide :smf, :parent => :base do
       return
     end
 
-    # Check to see if the service exists
-    cmd = Array[command(:svccfg), "select", @resource[:name]]
-    output = Puppet::Util::Execution.execute(
-        cmd, :combine => true, :failonfail => false)
-    if $CHILD_STATUS.exitstatus != 0
-      info output
-    end
-
     begin
       # get the current state and the next state, and if the next
       # state is set (i.e. not "-") use it for state comparison
--- a/lib/puppet/provider/user/user_role_add.rb
+++ b/lib/puppet/provider/user/user_role_add.rb
@@ -12,6 +12,7 @@ Puppet::Type.type(:user).provide :user_role_add, :parent => :useradd, :source =>
   options :home, :flag => "-d", :method => :dir
   options :comment, :method => :gecos
   options :groups, :flag => "-G"
+  options :shell, :flag => "-s"
   options :roles, :flag => "-R"
   options :auths, :flag => "-A"
   options :profiles, :flag => "-P"
@@ -26,7 +27,21 @@ Puppet::Type.type(:user).provide :user_role_add, :parent => :useradd, :source =>
     value !~ /\s/
   end
 
+  def shell=(value)
+    check_valid_shell
+    set("shell", value)
+  end
+
   has_features :manages_homedir, :allows_duplicates, :manages_solaris_rbac, :manages_passwords, :manages_password_age, :manages_shell
+
+  def check_valid_shell
+    unless File.exists?(@resource.should(:shell))
+      raise(Puppet::Error, "Shell #{@resource.should(:shell)} must exist")
+    end
+    unless File.executable?(@resource.should(:shell).to_s)
+      raise(Puppet::Error, "Shell #{@resource.should(:shell)} must be executable")
+    end
+  end
 
   #must override this to hand the keyvalue pairs
   def add_properties
--- a/spec/unit/provider/package/pkg_spec.rb
+++ b/spec/unit/provider/package/pkg_spec.rb
@@ -25,6 +25,28 @@ describe Puppet::Type.type(:package).provider(:pkg) do
 
   it_should_respond_to :install, :uninstall, :update, :query, :latest
 
+  context 'default' do
+    [ 10 ].each do |ver|
+      it "should not be the default provider on Solaris #{ver}" do
+        Facter.stubs(:value).with(:osfamily).returns(:Solaris)
+        Facter.stubs(:value).with(:kernelrelease).returns("5.#{ver}")
+        Facter.stubs(:value).with(:operatingsystem).returns(:Solaris)
+        Facter.stubs(:value).with(:operatingsystemmajrelease).returns("#{ver}")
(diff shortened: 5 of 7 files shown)


MosesMendoza:
@shawnferry thanks yet again for the contribution on Solaris! Any chance you'd be willing to split these into atomic commits? This is for a couple reasons - one so that reviewers can more easily review your changes and associate test changes w/code changes and second in case we end up needing to back out some piece of functionality we can cleanly revert that piece instead of having to revert all of the changes.

shawnferry:
@MosesMendoza done, sorry for the delay I missed the comment and the label change

shawnferry:
@MosesMendoza  any idea why this is failing? I can't find the reported string anywhere in git log output.

```
Checking commits...
(PUP-6797) service should check for smf service first
...passed
(PUP-6796) Support setting/checking shell in user_role_add
...passed
(PUP-6792) SMF should recursively enable services
...passed
(PUP-6795) Solaris does not support gzipped man pages
...passed
(PUP-6793) pkg provider should be default for Solaris 12
...passed
rake aborted!
		Delete file commited to wrong repo
```

MosesMendoza:
Sorry about that failure @shawnferry - all of your commits are correctly formatted. Our commit format check has trouble actually checking _only_ the commits in your PR and gets caught up on prior commits unrelated to your PR. I dislike it and filed PUP-6849 to see about getting it removed until we can fix it.

Thanks again for the contribution! I'll try to get this reviewed...

lib/puppet/provider/package/pkg.rb
```diff
@@ -23,7 +23,7 @@
 
   confine :osfamily => :solaris
 
-  defaultfor :osfamily => :solaris, :kernelrelease => '5.11'
+  defaultfor :osfamily => :solaris, :kernelrelease => ['5.11', '5.12']
```
MosesMendoza: I think it would be good to get spec coverage for this change, ie it's the default for solaris 11 and 12, and not the default for solaris 10. It doesn't look like we have something checking this for 11 yet, but a similar example is here: https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/package/dnf_spec.rb#L8-L26

install.rb
```diff
@@ -108,9 +108,12 @@ def do_man(man, strip = 'man/')
     FileUtils.makedirs(om, {:mode => 0755, :verbose => true})
     FileUtils.chmod(0755, om)
     FileUtils.install(mf, omf, {:mode => 0644, :preserve => true, :verbose => true})
-    gzip = %x{which gzip}
-    gzip.chomp!
-    %x{#{gzip} -f #{omf}}
+    # Solaris does not support gzipped man pages
+    unless Facter["os"].value["family"] == "Solaris"
```
MosesMendoza: I think `install.rb` can actually be run without requiring that `facter` be installed by passing `--no-check-prereqs`. If that's done and facter isn't installed I think this breaks.

On line 242 we already do a check for OS after confirming we aren't in 'no-prereqs' mode and assign the result to `$operatingsystem`, and then later use it to decide if we should call `do_man`. Maybe `do_man` could be extended to include an optional param of whether to compress man pages (defaulting to true), and then when calling do_man on 480 trigger that option in the negative if we're on solaris?

lib/puppet/provider/service/smf.rb
```diff
@@ -80,7 +80,7 @@ def startcmd
     when :maintenance, :degraded
       [command(:adm), :clear, @resource[:name]]
     else
-      [command(:adm), :enable, "-s", @resource[:name]]
+      [command(:adm), :enable, "-rs", @resource[:name]]
```
MosesMendoza: In some cases it seems like recursively enabling services is just the default - ie with systemd, if a service expresses dependencies those dependencies are also run on start. Do you know if there's any specific reason this isn't just the default with solaris? ie are you aware of any reason why this change would cause unexpected issues/breakage for anyone in solaris land?

lib/puppet/provider/package/pkg.rb
shawnferry: done

lib/puppet/provider/service/smf.rb
shawnferry: We committed the original behavior and it's unlikely that we will ever change it. We leave it up to the user to choose the behavior they want when manually executing. However, we think it should be the default behavior for puppet.

install.rb
shawnferry: Would you think to add a --[no]compress-man or similar option?

I can very easily modify my invocation of install.rb to use that option but it would seem like changing my use of Facter to $operatingsystem would retain the default behavior if --no-check-prereqs is passed.

MosesMendoza: Hi @shawnferry - just changing the use of `Facter[...` to $operatingsystem should work. Feels a little less intuitive to me since $operatingsystem hasn't been defined yet but it's not worth the additional effort to substantially revise the options I suppose.

shawnferry: done, I'll squash the commit and force an update when we are done. Losing the history makes it all harder.

shawnferry:
@MosesMendoza when you are happy with everything I'll squash any new commits

MosesMendoza:
thanks @shawnferry - further review to do on this one just trying to make time for it :)

lib/puppet/provider/service/smf.rb
```diff
@@ -129,6 +129,14 @@ def status
       return
     end
 
+    # Check to see if the service exists
+    cmd = Array[command(:svccfg), "select", @resource[:name]]
+    output = Puppet::Util::Execution.execute(
+        cmd, :combine => true, :failonfail => false)
+    if $CHILD_STATUS.exitstatus != 0
+      info output
+    end
```
MosesMendoza: @shawnferry while reviewing this part I realized it appears as if this provider doesn't distinguish between a non-existent service from a stopped/disabled and absent services and can't ensure => absent. I.e., `foo` does not actually exist:
```
-bash-4.1# puppet resource service foo
service { 'foo':
  ensure => 'stopped',
  enable => 'false',
}
```

I filed [PUP-6860](https://tickets.puppetlabs.com/browse/PUP-6860) to address this.

Separate from that, can you pull this existence check logic out into its own method in the provider? I'd like to fix the aforementioned bug and anticipate being able to use it if you could.

Second, I think the way we try to do this now is to reference the command as a method, and use Puppet's logging methods rather than the logging functions (though I admit I'm not 100% sure on that last one). ie:
```
begin
  svccfg(:select, @resource[:name])
rescue Puppet::ExecutionFailure => detail
  Puppet.debug("service was not found, svccfg returned: #{detail.inspect}")
  return false
end
true
```

shawnferry: Puppet.debug == debug :: Puppet.info == info; I can fully specify the call if that is the style.

I don't think I can reference the method without the rescue which seems like extra work. as opposed to passing fail_on_fail => false

lib/puppet/provider/service/smf.rb
```diff
@@ -123,6 +123,14 @@ def service_states
...
   end
 
+  def exists?
+    # non-zero exit status is a non-existent service
+    Puppet::Util::Execution.execute(
+      [command(:svccfg), "select", @resource[:name]],
+      :failonfail => false, :combine => true
+    )[1] == 0
+  end
+
```
shawnferry: I think this resolves PUP-6860; it could be extracted into
```Ruby
svccfg ...
rescue
```
but I'm not sure it is better.

lib/puppet/provider/service/smf.rb
```diff
@@ -135,8 +143,8 @@ def status
       states = service_states
       state = states[1] == "-" ? states[0] : states[1]
     rescue Puppet::ExecutionFailure
-      info "Could not get status on service #{self.name}"
-      return :stopped
+      debug "Could not get status on service #{self.name} #{$!}"
+      return :absent
```
shawnferry: Strictly speaking I'm not sure that we should ever reach this code anymore but I'm not sure. Is there a way to get here without discovering the service doesn't exist via exists? 

If it is reached it also adds the complexity that :absent isn't a valid service state. 
Such that if status is :absent but the requested state is :stopped the end result is a change on every execution that fails to take an absent service to a stopped state.

However, I think that is all addressed via the addition of exists? which then seems to "do the right thing(TM)"

MosesMendoza: @shawnferry - my apologies. I led you astray with asking you to define the `exists?` method. I was mistaken in that it turns out the `service` type isn't actually `ensurable` - which means the `exists?` method (as far as I can tell) won't ever be called/used (unless we use it ourselves as a helper for some other task, and we effectively already have that via a failure of the `service_states` method). The only other example of a service provider defining `exists?` is `launchd` and a quick check seems to indicate that isn't used either. I think by simply changing line 147 to `return :absent` we have what we need to solve the problem of puppet printing `ensure => stopped` for non-existent services, and the `exists?` method can be removed.

This means we actually do still reach this rescue block whenever a service doesn't exist. However, I don't think that's a problem. Puppet won't issue a change in that case, but rather a failure to change. Which is what we want.. I.e., with non-existent service `foo`, with your change to `status` method:

```
-bash-4.1# puppet resource service foo
service { 'foo':
  ensure => 'absent',
}
-bash-4.1# puppet resource service foo ensure=stopped
Error: Could not stop Service[foo]: Execution of '/usr/sbin/svcadm disable -s foo' returned 1: svcadm: Pattern 'foo' doesn't match any instances
Error: /Service[foo]/ensure: change from absent to stopped failed: Could not stop Service[foo]: Execution of '/usr/sbin/svcadm disable -s foo' returned 1: svcadm: Pattern 'foo' doesn't match any instances
service { 'foo':
  ensure => 'absent',
}
-bash-4.1# puppet apply -e " service { 'foo' : ensure => stopped } " --detailed-exitcodes
Notice: Compiled catalog for solaris11.localdomain in environment production in 0.33 seconds
Error: Could not stop Service[foo]: Execution of '/usr/sbin/svcadm disable -s foo' returned 1: svcadm: Pattern 'foo' doesn't match any instances
Error: /Stage[main]/Main/Service[foo]/ensure: change from absent to stopped failed: Could not stop Service[foo]: Execution of '/usr/sbin/svcadm disable -s foo' returned 1: svcadm: Pattern 'foo' doesn't match any instances
Notice: Applied catalog in 0.08 seconds
-bash-4.1# echo $?
4
```

shawnferry: @MosesMendoza dropped exists. If that's good I'll squash the commits unless you now use the new-ish "squash commits" option on merge.

lib/puppet/provider/user/user_role_add.rb
```diff
@@ -26,8 +27,22 @@
     value !~ /\s/
   end
 
+  def shell=(value)
```
shawnferry: Yeah, that is a somewhat questionable fallback and it does look like it should "just work"™ in the case of defining the flag.

The answer is apparently; it originally came from a customer and our notes say it was duplicated but don't give an example.  We've been carrying it since Puppet 3.x

MosesMendoza: After sitting on it a bit i like the explicitness of this change. +1 to leaving as-is.

@shawnferry if you want to squash commits to per-ticket I think this is good to go

shawnferry: @MosesMendoza squashed

MosesMendoza:
👍

MosesMendoza:
I am running this through an ad-hoc acceptance run and will merge on success

MosesMendoza:
passed an acceptance run against tests impacted by this change

PR #5329, opened 2016-10-13
(PUP-6474) Add gettext-setup gem and externalize some early strings

branan:


kylog:
Typo in the PR title? Says "6464" but means "6474".

kylog:
Grumble. Not sure why the @#$)(*@#$ commit checker got confused here. We can probably just ignore that.

This LGTM. IIUC it's only the 1d0874f commit that we'd want to revert once the server change goes in - do I have that right?

branan:
@kylog did we ever fix the "wrong base branch" thing? 

And yeah, that's the only commit to revert. I tried to keep it contained

kylog:
@branan yes in e1efdebf. I think you may not have that in your branch. As an experiment, can you rebase this branch from current master and push?

branan:
@kylog done. Let's see what happens...

branan:
wait, darn. I forgot to fetch master. Mine was way out of date. Let's try this one more time

branan:
@kylog There we go, that did the trick

Rakefile
```diff
@@ -103,3 +103,8 @@ task(:commits) do
     puts "...passed"
   end
 end
+
+# Add i18n tasks (gettext:pot, gettext:po*)
```
Magisus: Do I don't know that much about rake: when does this run? (i.e. when does the pot file get generated?)

branan: This calls out to the gettext gem in order to add a couple of tasks. They have to be invoked manually to generate the pot file

Magisus:
This looks good. If it works, I say 👍

branan:
Looks like we still need more packaging changes to make gettext happy. I'm working on it.

branan:
The message it's failing on isn't in this PR. I'm going to try rebasing onto latest mater to see if that helps

Magisus:
@branan I verified that from this PR I can use the gem to generate a POT file on OSX and Ubuntu. I was hoping to do Windows too, but I've been having trouble just getting `bundle install` to complete cleanly on my VM, for reasons unrelated to this. Do you have any recommended ways to verify this on Windows?

MosesMendoza:
reviewing

lib/puppet.rb
```diff
@@ -19,6 +19,20 @@
 require 'puppet/external/pson/version'
 require 'puppet/external/pson/pure'
 
+## It would be really nice to use Puppet.features here, but we have to
+## do gettext setup before all of that machinery is loaded. Instead,
+## we will use a simple boolean to handle gettext availability in this
+## file, and drop in a "stub" translation function here.
+begin
```
MosesMendoza: Is there a longer term plan for ensuring gettext functions are available so that we can remove this?

lib/puppet.rb
```diff
@@ -19,6 +19,20 @@
...
+  require 'gettext-setup'
+  GETTEXT_AVAILABLE=true
+rescue LoadError
+  def _(msg)
```
MosesMendoza: `_()` is only one of the functions defined by gettext-setup. Are we intending to duplicate the pluralization function `n_()` too? I don't have something better off the top of my head butI worry about maintainability.

install.rb
```diff
@@ -294,6 +308,12 @@ def prepare_installation
     bindir = RbConfig::CONFIG['bindir']
   end
 
+  if not InstallOptions.localedir.nil?
+    localedir = InstallOptions.localedir
+  else
+    localedir = "/opt/puppetlabs/puppet/share/puppet/locale"
```
MosesMendoza: It looks like some of the other directories/files installed by `install.rb` include a Windows path. Does that apply here and if so are we saving the windows use-case for a later PR?

lib/puppet.rb
```diff
@@ -36,6 +50,29 @@ module Puppet
...
+      elsif File.exist?(posix_system_locale_path)
+        locale_path = posix_system_locale_path
+      else
+        # We couldn't load our locale data. Possibly we're loaded as a rubygem or something?
+        locale_path = nil
+      end
+
+      if locale_path
+        GettextSetup.initialize(locale_path)
+        FastGettext.locale = GettextSetup.negotiate_locale(ENV["LANG"])
```
MosesMendoza: I think `ENV["LANG"]` might be spotty at best on Windows.. In https://github.com/puppetlabs/puppet-agent/commit/29e90b9375c23ee4bd854ad720188bea13f65754 we added the `locale` gem - can use that to determine the locale? Might be more robust..

glennsarti:
Also the gettext-setup gem has no automated testing, only trivial spec test coverage and no OS specific acceptance tests.  Are we sure we want to depend on this gem?

lib/puppet.rb
Magisus: Longer term we may want the gettext-setup gem to handle getting the locale. For now, this seems like a good suggestion.

glennsarti: LANG is not a standard locale environment variable on Windows and is not set as part of Puppet Agent.  This effectively removes Windows as a supported platform for localisation.  I feel that if we want to merge this PR, I would like to see a ticket logged for this and referenced in this PR.

Magisus: @glennsarti do you think the gem referenced by Moses would allow this to work on Windows? [The docs](https://github.com/ruby-gettext/locale) seem to indicate Windows support.

install.rb
Magisus: We should add a Windows path. However, we don't seem to have an analog yet to the "share" folder used in the Unix path. Does anyone have a suggestion for where it might make sense to put these files, given Windows file structure? @Iristyle or @glennsarti ?

lib/puppet.rb
glennsarti: The gem does appear to work on Windows.  It's using the older API call GetThreadLocaleID whereas the more modern call should be GetUserDefaultLocaleName. https://msdn.microsoft.com/en-us/library/windows/desktop/dd318136(v=vs.85).aspx.  To be honest though, we could probably implement the call ourselves instead of pulling in yet-another-gem.

install.rb
glennsarti: I assume folder will contain the per-locale localised strings?  I imagine this would be the installation path of Puppet `\locales`  e.g. `C:\Program Files\Puppet Labs\Puppet\puppet\locales`

branan:
@glennsarti regarding the `gettext-setup` gem: It's owned by us, and is intended to be a lightweight way to share gettext initialization code across ruby projects. We can definitely evaluate pulling it out and handling that stuff internally, but probably we just need to take ownership of it now that it's past "prototype" stage and make it less bad.

install.rb
glennsarti: Ahh ok... TLDR  `ProgramFiles` for binaries, `ProgramData` for config information.  This should probably be in ProgramData so you can copy-pasta to configs above and use;

`localedir = File.join(Dir::COMMON_APPDATA, "PuppetLabs", "puppet", "share", "puppet", "locale")`

@Iristyle Sound reasonable?

branan: Are translations really "configuration"? I would not expect them to be user-modifiable or user-replaceable. We ship them with a given version of the project and that's it. In my mind that data belongs in "Program Files"

glennsarti: True....forgot about that.  Yup ProgFiles would be more appropriate then.

Magisus: So there are plenty of examples for how we get the `ProgramData` folder (Dir::COMMON_APPDATA). Is there a similar way to get `Program Files`?

lib/puppet.rb
MosesMendoza: @glennsarti we've already pulled in (and shipped in 1.8.0) the `locale` gem (mentioned above - https://github.com/puppetlabs/puppet-agent/commit/29e90b9375c23ee4bd854ad720188bea13f65754). Are you suggesting we remove the gem?

Maybe we could still use it and contribute improvements (ie using the newer API) upstream rather than maintaining our own logic? 

Then we also don't necessarily need to expose another API for locale within puppet and can avoid augmenting `Puppet::Util`..

glennsarti: It's in puppet-agent but not puppet.

MosesMendoza: Ah sure, that's true. In that case I think we should do as we did for `gettext-setup` in this PR and express gem dependencies on `locale`. That way our gem and source users and our package users don't have to reconcile differences in the stack and we can just use it cleanly here.

glennsarti: So my thinking was, that the actual code we need is only 8 lines or so (A single API call) for Windows.  So we could just write that code internally instead of adding yet another gem dependency for something trivial.  That's why I though it would be better to refactor that into a method call (perhaps Puppet::Util is a bad idea, but could still be a separate method) so we can then test it appropriately.

MosesMendoza: I definitely agree it's a valid concern that we avoid trivially expressing new dependencies - it's just this kind of logic seems to me to be outside of Puppet's core competency and carrying it ourselves will have its own long-term maintenance burden. Historically I feel like when we've tried to do similar things we've eventually regretted it but I'm failing to come up with examples. Maybe other @puppetlabs/puppet-maintainers have varying opinions?

glennsarti: Agree.  I'll defer to puppet maintainers for direction on this one.

locales/puppet.pot
```diff
@@ -0,0 +1,124 @@
+# SOME DESCRIPTIVE TITLE.
```
McdonaldSeanp: This should probably say something else.

MosesMendoza:
This PR is superseded by #910

PR #5330, opened 2016-10-14
(PUP-6803) report failed run on failure to generate resources

MosesMendoza:
Prior to this commit, if a resource type with a defined :generate or :eval_generate failed to generate resources (i.e., raising an exception in this process), an exception was logged by the resource and resource generation was skipped for that resource, but further catalog evaluation proceeded normally.

Because computing run status only took into account failed resource metrics in determining whether to fail a run, and the resource on which :generate or :eval_generate was called is not marked as failed, reports resulting from runs in which resources failed to generate still claimed success, despite this failure.

This commit adds a boolean field to Puppet::Transaction::AdditionalResourceGenerator to track whether any resources have raised an exception during resource generation. This value is queried by the calling transaction instance, and is used to populate a new field of the same name in Puppet::Transaction::Report.Puppet::Transaction::Report#compute_status takes this value into account and will now return `failed` if true. This commit also intentionally avoids changing the report format by excluding the `resources_failed_to_generate` from the yaml report.

An alternative initial approach kept a list of all resources that had failures during resource generation and surfaced this in the report, but the decision was made to avoid this in favor of a simple boolean to minimize resource use and complexity, and avoid having to change the report format. Because the resource logs an exception noting the failure, this failure information is still surfaced in the report with a `source` populated by the resource. Thus a user can connect their failed run status back to the exception logged in the report without needing additional information.

In the same spirit of avoiding backwards-incompatible changes (perhaps unfortunately) this PR does not change the resulting puppet run exit code if resources fail to generate - all things being equal this scenario will still return 0.


MosesMendoza:
example of a report with a resource that failed to generate resources before this PR:
https://gist.github.com/MosesMendoza/20aa16f7afa5e18399bc4e28974c43fb

same but after this PR:
https://gist.github.com/MosesMendoza/08ec7e1568a5622871e31875d53feb59

MosesMendoza:
appveyor spec failure is legit test failure on windows
```
1) Puppet::Transaction::AdditionalResourceGenerator when applying eval_generate sets resources_failed_to_generate to true if resource#eval_generate raises an exception
     Failure/Error: raise error
     Puppet::ResourceError:
       Parameter path failed on File[/foo]: File paths must be fully qualified, not '/foo' at line 1
     # ./lib/puppet/util/errors.rb:106:in `fail'
     # ./lib/puppet/type/file.rb:56:in `block (3 levels) in <top (required)>'
```

need to update the new test

MosesMendoza:
@Iristyle @joshcooper any chance one of you has some cycles to review?

lib/puppet/transaction/report.rb
```diff
@@ -127,6 +127,11 @@ class Puppet::Transaction::Report
   #      corrective changes.
   attr_reader :corrective_change
 
+  # @return [Boolean] true if one or more resources attempted to generate
+  #   resources and failed
+  #
+  attr_accessor :resources_failed_to_generate
```
MosesMendoza: I don't include this in the initializer(s) but I'm wondering if I should, specifically `initialize_from_hash` ?

MosesMendoza:
Ping @hlindberg for review as well

lib/puppet/transaction/report.rb
```diff
@@ -167,7 +172,7 @@ def add_resource_status(status)
 
   # @api private
   def compute_status(resource_metrics, change_metric)
-    if (resource_metrics["failed"] || 0) > 0
+    if (resources_failed_to_generate) || (resource_metrics["failed"] || 0) > 0
```
hlindberg: picky: parentheses not needed around resources_failed_to_generate

PR #5331, opened 2016-10-14
(PUP-6819) Use packaged cert when attempting to contact rubygems

Magisus:
Since the rubygems certificate update, Windows boxes provisioned by
beaker-hostgenerator have been unable to connect to rubygems due to a
bad cert. This adds a valid cert to the environment when calling out to
rubygems.

### DIFF
--- a/acceptance/lib/puppet/acceptance/common_utils.rb
+++ b/acceptance/lib/puppet/acceptance/common_utils.rb
@@ -134,7 +134,11 @@ module Puppet
       def gem_command(host, type='aio')
         if type == 'aio'
           if host['platform'] =~ /windows/
-            "env PATH=\"#{host['privatebindir']}:${PATH}\" cmd /c gem"
+            if host['platform'] =~ /-64$/ && host['ruby_arch'] != 'x64'
+              "env SSL_CERT_FILE=\"C:/Program Files (x86)/Puppet Labs/Puppet/puppet/ssl/cert.pem\" PATH=\"#{host['privatebindir']}:${PATH}\" cmd /c gem"
+            else
+              "env SSL_CERT_FILE=\"C:/Program Files/Puppet Labs/Puppet/puppet/ssl/cert.pem\" PATH=\"#{host['privatebindir']}:${PATH}\" cmd /c gem"
+            end
           else
             "env PATH=\"#{host['privatebindir']}:${PATH}\" gem"
           end


acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -134,7 +134,7 @@ def ruby_command(host)
       def gem_command(host, type='aio')
         if type == 'aio'
           if host['platform'] =~ /windows/
-            "env PATH=\"#{host['privatebindir']}:${PATH}\" cmd /c gem"
+            "env SSL_CERT_FILE=\"C:/Program Files/Puppet Labs/Puppet/puppet/ssl/cert.pem\" PATH=\"#{host['privatebindir']}:${PATH}\" cmd /c gem"
```
MosesMendoza: I think this path might vary based on architecture?

Magisus: We were wondering about that. Do you know what might be the more robust way to get it?

Magisus: Yeah it would be nice to have that somewhere central, but I think it's reasonable to add the arch test just here for now.

Magisus: @MosesMendoza do you have an example of platform you'd expect this to fail on? I want to verify both the problem and the fix.

MosesMendoza: I think it would fail on the beaker-hostgenerator platform `windows2012r2-6432` (signifying 32 bit puppet MSI installed on 64-bit host). This page https://docs.puppet.com/pe/latest/install_windows.html#what-gets-installed notes that distinction anyway

Magisus: Updated, verified that it broke and now works on that platform.

MosesMendoza:
Per conversation in PA-620 I think there's a possibility we haven't solved the whole issue with this as users trying to install gems on previous PE versions as part of support module install instructions (ie https://forge.puppet.com/puppetlabs/azure) are no longer able to since the cert switch? I'm just commenting that here to suggest we hold on merging this PR until the conversation in PA-620 is resolved - not trying to bifurcate the conversation

Magisus:
Closing this in favor of the alternative solution proposed in [PA-620](https://tickets.puppetlabs.com/browse/PA-620)

PR #5338, opened 2016-10-19
(PUP-1476) Include puppet and ruby versions in User-Agent

joshcooper:
Previously, puppet did not specify an HTTP User-Agent string, so ruby
would send the default "Ruby". This made it difficult to debug and
interpret server-side logs when different versions of puppet agents
connect.

This commit adds a puppet setting `http_user_agent` which includes the
puppet and ruby versions. The default value is of the form:

```
Puppet/4.7.0 Ruby/2.1.8-p440 (x86_64-darwin15.0)
```

The string is exposed as a setting for cases where puppet is being used
as a library, and the enclosing application may need to override that.

Puppet's HTTP connection code will send the User-Agent header with the
above value for all requests. The value is taken from the Puppet Module
Tool code. However, the strings are not exactly the same, because the
PMT prefixes the string with additional information:

```
PMT/1.1.1 (v3; Net::HTTP) Puppet/4.7.0 Ruby/2.1.8-p440 (x86_64-darwin15.0)
```

### DIFF
--- a/lib/puppet/network/http/connection.rb
+++ b/lib/puppet/network/http/connection.rb
@@ -197,6 +197,8 @@ module Puppet::Network::HTTP
     end
 
     def apply_options_to(request, options)
+      request["User-Agent"] = Puppet[:http_user_agent]
+
       if options[:basic_auth]
         request.basic_auth(options[:basic_auth][:user], options[:basic_auth][:password])
       end
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -517,6 +517,10 @@ deprecated and has been replaced by 'always_retry_plugins'."
       read after the elapsed interval then the connection will be closed. The default value is unlimited.
       #{AS_DURATION}",
     },
+    :http_user_agent => {
+      :default => "Puppet/#{Puppet.version} Ruby/#{RUBY_VERSION}-p#{RUBY_PATCHLEVEL} (#{RUBY_PLATFORM})",
+      :desc    => "The HTTP User-Agent string to send when making network requests."
+    },
     :filetimeout => {
       :default    => "15s",
       :type       => :duration,
--- a/lib/puppet/forge/repository.rb
+++ b/lib/puppet/forge/repository.rb
@@ -147,8 +147,7 @@ class Puppet::Forge
     def user_agent
       @user_agent ||= [
         @agent,
-        "Puppet/#{Puppet.version}",
-        "Ruby/#{RUBY_VERSION}-p#{RUBY_PATCHLEVEL} (#{RUBY_PLATFORM})",
+        Puppet[:http_user_agent]
       ].join(' ').freeze
     end
   end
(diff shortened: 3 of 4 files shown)


joshcooper:
@MosesMendoza good question. This is one of those things where we all agree the new behavior makes the `User-Agent` string usable for debugging and analysis, but maybe the change will break someone's workflow. I can't say for certain it won't break anyone, but to me it comes down to:

1. The current user-agent string is "Ruby", and is indistinguishable from any other ruby-based project which uses `Net::HTTP` and doesn't override the value.
1. The new behavior still sends "Ruby" in the string, so it would only break if someone was looking for an exact match as opposed to regexp.
1. The user-agent string is exposed as a property, so it is possible to set it to "Ruby" if someone really needed to.

MosesMendoza:
@joshcooper seems reasonable, that makes sense to me.

MosesMendoza:
👍

Iristyle:
Looks like the mode somehow changed from 644 to 755 on `connection_spec.rb`

lib/puppet/network/http/connection.rb
```diff
@@ -197,6 +197,8 @@ def request_with_redirects(request, options)
     end
 
     def apply_options_to(request, options)
```
joshcooper: Good question. For puppet (agent, webrick, passenger), `Puppet::Network::HttpPool.http_instance(url.host, url.port, use_ssl)` will return an instance of the default http client class `Puppet::Network::HTTP::Connection` and that's the only case where we would try to set the `User-Agent` above (in the call to Connection#apply_options_to).

Puppetserver on the other hand overrides the [http client class](https://github.com/puppetlabs/puppetserver/blob/fa34cedb5ba38f86ab156b40c007fe25fae600b1/src/ruby/puppetserver-lib/puppet/server/config.rb#L29), but in that case, it implements its own [post](https://github.com/puppetlabs/puppetserver/blob/master/src/ruby/puppetserver-lib/puppet/server/http_client.rb#L53) (and `get`) logic. That said, I don't know if puppetserver sets `User-Agent` for connections it makes, e.g. to puppetdb, and whether we want puppetserver to read the puppet settings for that purpose? /cc @cprice404 @camlow325

cprice404: We do not explicitly set a User-Agent in Puppet Server.  our client gets created here:

https://github.com/puppetlabs/puppetserver/blob/fa34cedb5ba38f86ab156b40c007fe25fae600b1/src/ruby/puppetserver-lib/puppet/server/http_client.rb#L162

And that ends up (through a couple of layers) instantiating an instance of the Apache async-http-client.  I don't believe that the Apache client provides a default value for that header so my guess is that we're not sending one at all.

If necessary we could certainly file a separate ticket for Puppet Server to have it add the header.

joshcooper:
I think this is good to go. I'd lean on not filing a puppetserver ticket until there is a need. In the case of the agent, it would be helpful to include the agent version in the User-Agent string so that the puppetserver access/error logs have this information for troubleshooting, since it's common to have different agent versions connecting to the same puppetserver.

PR #5341, opened 2016-10-19
(PUP-6827) fix test to meet password complexity reqs

MosesMendoza:
Prior to this commit, the test '#6857: redact password hashes when applying in
noop mode' would fail on a host that had 'Password must meet complexity
requirements' set to true, because puppet was applying user management
manifests containing passwords that did not meet said requirements. This commit
updates the test to meet the complexity requirements (documented here:
https://technet.microsoft.com/en-us/library/hh994562(v=ws.11).aspx), allowing tests to pass.

Assumedly prior hosts in the acceptance testing systems were either
pre-configured with this requirement disabled or it was disabled at test time
for other tests (the latter of which seems unlikely).

### DIFF
--- a/acceptance/tests/ticket_6857_password-disclosure-when-changing-a-users-password.rb
+++ b/acceptance/tests/ticket_6857_password-disclosure-when-changing-a-users-password.rb
@@ -28,14 +28,14 @@ end
 adduser_manifest = <<MANIFEST
 user { '#{username}':
   ensure   => 'present',
-  password => 'apassword',
+  password => 'Apassw0rd!',
 }
 MANIFEST
 
 changepass_manifest = <<MANIFEST
 user { '#{username}':
   ensure   => 'present',
-  password => 'newpassword',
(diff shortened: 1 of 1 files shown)


MosesMendoza:
This previously failed on the internal windows 2016 64-bit test host. 

Passing run: https://gist.github.com/MosesMendoza/2a7f178661ad0bf8c24f49bad254165f

Magisus:
For consistency's sake, do you think we ought to look into whether or not the other hosts had different settings, and update this new one to match (or vice versa, now that the test is fixed)?

MosesMendoza:
@Magisus seems reasonable. You mean update check the templates in the vm pool for inconsistencies? @jcoconnor thoughts?

MosesMendoza:
@Magisus I think that probably shouldn't hold up this PR though?

Magisus:
No, it doesn't need to hold up the PR. I'm 👍 on this change.

PR #5342, opened 2016-10-20
(PUP-6789) Allow capabilties to be consumed across environments.

adreyer:
Previously the capability finder would only search inside the
environment of the current catalog for capability resources to consume.
Now the capability finder will first search for a matching capability
across all environments and only limit it's search to a single
environment if a conflict is found. This allows users to test new code
on part of an application by moving nodes into a different environment
as long as they use globally unique capability titles.

### DIFF
--- a/lib/puppet/resource/capability_finder.rb
+++ b/lib/puppet/resource/capability_finder.rb
@@ -26,7 +26,12 @@ module Puppet::Resource::CapabilityFinder
       raise Puppet::DevError, 'PuppetDB is not available'
     end
 
-    resources = search(environment, nil, cap)
+    resources = search(nil, nil, cap)
+
+    if resources.size > 1
+      Puppet.debug "Found multiple resources when looking up capability #{cap}, filtering by environment #{environment}"
+      resources = resources.select { |r| r['tags'].any? { |t| t == "producer:#{environment}" } }
+    end
 
     if resources.size > 1 && code_id
       Puppet.debug "Found multiple resources when looking up capability #{cap}, filtering by code id #{code_id}"
@@ -54,8 +59,13 @@ module Puppet::Resource::CapabilityFinder
       'and',
       ['=', 'type', cap.type.capitalize],
       ['=', 'title', cap.title.to_s],
-      ['=', 'tag', "producer:#{environment}"]
     ]
+
+    if environment.nil?
+      query_terms << ['~', 'tag', "^producer:"]
+    else
+      query_terms << ['=', 'tag', "producer:#{environment}"]
+    end
 
     unless code_id.nil?
       query_terms << ['in', 'certname',
--- a/spec/unit/resource/capability_finder_spec.rb
+++ b/spec/unit/resource/capability_finder_spec.rb
@@ -71,29 +71,50 @@ describe Puppet::Resource::CapabilityFinder do
       let(:capability) { Puppet::Resource.new('Cap', 'cap') }
       let(:code_id) { 'b59e5df0578ef411f773ee6c33d8073c50e7b8fe' }
 
-      it 'should search for the resource without including code_id' do
+      it 'should search for the resource without including code_id or environment' do
         resources = [{"type"=>"Cap", "title"=>"cap", "parameters"=>{"host"=>"ahost"}}]
-        Puppet::Resource::CapabilityFinder.stubs(:search).with('production', nil, capability).returns resources
+        Puppet::Resource::CapabilityFinder.stubs(:search).with(nil, nil, capability).returns resources
 
         result = Puppet::Resource::CapabilityFinder.find('production', code_id, Puppet::Resource.new('Cap', 'cap'))
         expect(result['host']).to eq('ahost')
       end
 
       it 'should return nil if no resource is found' do
-        Puppet::Resource::CapabilityFinder.stubs(:search).with('production', nil, capability).returns []
+        Puppet::Resource::CapabilityFinder.stubs(:search).with(nil, nil, capability).returns []
 
         result = Puppet::Resource::CapabilityFinder.find('production', code_id, capability)
         expect(result).to be_nil
       end
 
-      describe 'when multiple results are returned' do
+      describe 'when multiple results are returned for different environments' do
         let(:resources) do
-          [{"type"=>"Cap", "title"=>"cap", "parameters"=>{"host"=>"ahost"}},
-           {"type"=>"Cap", "title"=>"cap", "parameters"=>{"host"=>"bhost"}}]
+          [{"type"=>"Cap", "title"=>"cap", "parameters"=>{"host"=>"ahost"}, "tags"=>["producer:production"]},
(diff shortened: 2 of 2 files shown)


thallgren:
I have two questions:
1. Wouldn't it be safer to first search the current catalog, and then widen the search only if necessary? With the suggested approach, it seems that another environment can case a breakage (a conflict) even if the capability is provided by the current environment.
2. Given the nature of the use-case (topic-branches), shouldn't this behavior be controlled by a setting? Or is it save to assume that the cross-environment search behavior is always desired?

MosesMendoza:
Do we have a sense of how often we expect a capability resource to exist across more than one environment and/or code id? Is it very often? I'm not entirely familiar with the structure of the data returned by the query to PDB, but I ask because if the answer is very often, would it be possible to actually perform the filtering of resources by `environment` and `code id` locally on the data structure returned from `Puppet::Resource::CapabilityFinder.search(nil, nil, cap)`, without having to re-query PDB? It seems like making up to three network calls per resource (in the worst case) may have less desirable performance/reliability characteristics than if we can make just one. Again, not sure if such an optimization would be worthwhile because I don't know how often we run into this situation.

adreyer:
@thallgren 
1. I think we only get here if the capability resource isn't in the current catalog https://github.com/puppetlabs/puppet/blob/master/lib/puppet/resource/catalog.rb#L372
2. I can't think of a good use case for having duplicate resources across environments so I would prefer simpler behavior of always searching across environments.

@MosesMendoza It shouldn't happen often the main case is when a component moves from one node to another but not yet run on the first node. It can then be found in the catalogs of both nodes. The new situation here is if a user doesn't make capabilities globally unique. We haven't been able to come up with a good use case for doing that. I do think that filtering should move into the ruby code and considered leaving a 'TODO' about that given that but decided not to do it now since I'm so short on time.

thallgren:
@adreyer
This is perhaps far fetched. But what if you have similar environments, one for staging, and one for production? Something happens to production that make the provided capability unavailable. It will then instead find it in the staging environment which is probably highly undesirable.

adreyer:
I don't think it's far fetched that someone might want that but I think that should be discouraged.

Ideologically I think Puppet environments are not the proper tool for separating production and staging nodes. Orchestration workflows are already heavily based on that assumption. We have an opportunity now before applications are widely adopted to get people into the habit of making capability resources unique across environments.  If some users do want to use environments that way they should add ${environment} to the capability title.

lib/puppet/resource/capability_finder.rb
```diff
@@ -26,7 +26,12 @@ def self.find(environment, code_id, cap)
       raise Puppet::DevError, 'PuppetDB is not available'
     end
 
-    resources = search(environment, nil, cap)
+    resources = search(nil, nil, cap)
+
+    if resources.size > 1
+      Puppet.debug "Found multiple resources when looking up capability #{cap}, filtering by environment #{environment}"
+      resources = search(environment, nil, cap)
```
nicklewis: Because the environment is indicated by a tag, we should be able to filter the resources we just retrieved, rather than making another query. We make a second query for the code id case just because the code id isn't returned as part of the resource.

adreyer:
@nicklewis fixed

thallgren:
I'm not very familiar with orchestration workflows, but I do have some concerns:
1. We do promote environments as a way to keep things separate. This solution breaks that axiom. I understand that the need to break it exists to support the topic-branch use case but I think it should be a very conscious act.
2. Discouraging from bad workflows is a good thing, but doing so by introducing strange side-effects for the discouraged workflows is not the way to go.
3. Wouldn't it be better to have a setting that pinpoints the environment to use, and thereby enable parallel setups? Seems like this solution would make that impossible.

adreyer:
1. This is a very conscious decision we've been discussing in depth since last spring.
2. I don't think we should break the undesired workflow out of spite but we shouldn't make the desired workflow more complex to make something that doesn't make sense easier. I've yet to see someone set up applications with the exact same declarations in each environment but different nodes. If someone wanted to it would still be easy to make the definition environment aware.
3. I'm not sure I understand fully. Can't that be accomplished with the titles.

thallgren:
#1. I was referring to a conscious decision made by the user, not by Puppet.
#2. As I wrote before, I'm not very familiar with the workflows that are used here, so perhaps my concerns are indeed unfounded.
#3. I'm sure it can be accomplished with titles. That's not my point. I'm thinking in terms of separation of concern. What if someone responsible for one setup decides to use the same title that is already present in another? They shouldn't need to know what titles that are used.

adreyer:
1. We expect users to use feature branches they shouldn't have to make a conscious decision for that to work with application management.
3a. In practice to prevent conflicts between application instances you already have to use string interpolation when creating capabilities something like "foo-${name}"*. To limit consumption to an environment you need "$name" to be unique across environments or name the capability "foo-${name}-${environment}".**
3b. Environments are a mess in puppet because we attempt to use them to solve too many problems. Encouraging users to use them for managing independent applications makes that problem worse. I do think we need to support independent applications but by using the already overloaded environment construct for that we make both environments and application management less usable.


*https://github.com/puppetlabs/puppetlabs-wordpress_app/blob/master/manifests/init.pp#L39
** Due to the bug around $environment not being available during catalog compilation users will currently have to set it themselves in site.pp. The fact that this bug still exists is strong evidence that users aren't doing anything with environment specific applications.

hlindberg:
I see no real harm in doing it this way as there is no loading of code from multiple environments going on. At worst the information loaded from another environment was produced by an incompatible version.
The same can be done by anyone in a manifest running a query against PDB. Users would certainly need to understand what they are doing and when using application orchestration they cannot really use the same master/PDB for different stages (production, test, etc). Users would also need to understand the limitations of what they can do in a feature branch wrt. incompatibilities on the main branch.

thallgren:
I know that environments have many issues. That's partly why I'm concerned. We've just made several improvements to improve environment isolation. I'd like to thing that environments can be made useful by increasing the level of strictness around how they are used and make their definition more explicit. That's why I find the concept of a random search across all environments that happens to be active in the PuppetDB at the time of a search somewhat scary and that's also why I'm asking if this could be controlling this with a setting. An environment can be though of as a scope. What is the scope of an application? Is it multiple environments? If so, shouldn't that set of environments be defined somewhere? Is it always OK so say that that scope always is "everything"?

adreyer:
@hlindberg Agreed we need to document the cases where inconsistency in application compilation are likely to cause problems.
1. Different code with incompatible types.
2. Different code results in different components between compilations.
3. Dynamic application generation is unstable between compilations.

@thallgren We're still trying to figure out what the scope of application graphs should be, how data and code should feed into application compilation, how users should provide inputs to that, and how jobs should be tied to those graphs(these questions are under active research by melinda). What is clear is that applications cannot not be tied one to one with environments.  We can come back and add scoping later after we know what type of scoping makes sense.

PR #5345, opened 2016-10-24
(PUP-6840) Add revert policy

johnduarte:
This commit outlines the revert policy for code changes that result
in test pipeline failures. The purpose of this is to establish a
clear protocol for when a revert should take place and how it should
be communicated to the original contributor.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,6 +104,30 @@ a ticket number.
 * After feedback has been given we expect responses within two weeks. After two
   weeks we may close the pull request if it isn't showing any activity.
 
+## Revert Policy
+By running tests in advance and by engaging with peer review for prospective
+changes, your contributions have a high probability of becoming long lived
+parts of the the project. After being merged, the code will run through a
+series of testing pipelines on a large number of operating system
+environments. These pipelines can reveal incompatibilities that are difficult
+to detect in advance.
+
+If the code change results in a test failure, we will make our best effort to
+correct the error. If a fix cannot be determined and committed within 24 hours
+of its discovery, the commit(s) responsible _may_ be reverted, at the
+discretion of the committer and Puppet maintainers. This action would be taken
+to help maintain passing states in our testing pipelines.
+
+The original contributor will be notified of the revert in the Jira ticket
+associated with the change. A reference to the test(s) and operating system(s)
+that failed as a result of the code change will also be added to the Jira
+ticket. This test(s) should be used to check future submissions of the code to
+ensure the issue has been resolved.
+
+### Summary
+* Changes resulting in test pipeline failures will be reverted if they cannot
+  be resolved within one business day.
+
 # Additional Resources
 
 * [Puppet community guidelines](https://docs.puppet.com/community/community_guidelines.html)


CONTRIBUTING.md
```diff
@@ -104,6 +104,30 @@ a ticket number.
...
 
+## Revert Policy
+By running tests in advance and by engaging with peer review for prospective
+changes, your contributions have a high probability of becoming long lived
+parts of the the project. After being merged, the code will run through a
+series of testing pipelines on a large number of operating system
+environments. These pipelines can reveal incompatibilities that are difficult
+to detect in advance.
+
+If the code change results in a test failure, we will make our best effort to
```
stahnma: What time do they get merged in? Wouldn't it make more sense to say after X amount of time, rather than at the end of the business day for a select portion of the population?

johnduarte: The regression discovery is the trigger for this situation. It is loosely coupled to the time of the merge depending on the status of the test pipelines and how quickly they get run after the merge. So, from the contributers standpoint, talking about a time duration based on the time of merge is more difficult to honor consistently.

Would it me more amenable to discuss an amount of time in terms of time from regression discovery?

Example
```
If the code change results in a test failure, we will make our best effort to
correct the error. If a fix cannot be determined and committed within 24 hours of its discovery,
the commit(s) responsible will be reverted.
```

MosesMendoza: @johnduarte I like this amended change - it makes more sense to me (thinking globally) to consider the merge => test failure => revert as a static time window rather than adjusting based on relation to a specific Portland time.

MosesMendoza:
I'm wary of language with a very specific, almost contractual, nature. If I'm not mistaken, the goal of this PR is to make contributors aware that changes that fail tests and cannot be resolved quickly will be reverted, but the language IMO has a second side-effect of implying a strict obligation to fulfill this policy. I don't have a ton of confidence that in _all_ cases a committer/maintainer will be in a position to revert a code change that causes failure within 24 hours of said failure, or that doing so will _always_ be the best course of action (ex: untangling a revert is estimated to take 48 hours and introduce great risk at the integration level, while working directly towards a fix may take 36).

If the goal of this PR is to make contributors aware of the nature of our internal testing and notify them that test failures will almost certainly result in code reversion if not fixable in 24 hours, then perhaps a slightly different wording would be akin to rfc-style "MAY" rather than rfc-style "MUST". I.e., something like:
{code}
If the code change results in a test failure, we will make our best effort to
correct the error. If a fix cannot be determined and committed within 24 hours of its discovery, the commit(s) responsible _may_ be reverted, at the discretion of the committer and Puppet maintainers.
{code}

I would rather create a policy that we can apply and fulfill consistently across the board than one to which we know will have exceptions.

johnduarte:
@MosesMendoza I agree with our assessment of the intent of the policy and approve of your phrasing to prepare the contributer for an action rather than establishing a firm contract.
I will amend the PR in accordance with your suggestion.

whopper:
The travis error is related to what appears to be an accidental commit from October, https://github.com/puppetlabs/puppet/commit/6e7949803cf4dde355f3fe140fdeee29ebbdba44. (I assume this is why we're turning off the commit check?)

I'm not sure why the appveyor run is failing, but it's crashing with `Command exited with code -1073741819`(FFFFFFFFC0000005). 

In any case, it looks like @johnduarte addressed all of the suggestions here. @MosesMendoza are you 👍 on this?

joshcooper:
Let's see if we can get a clean CI run, closing and reopening.

joshcooper:
The travis failure is unrelated to this PR.

PR #5346, opened 2016-10-24
(PUP-6780) Add new lookup framework and reroute calls to it

thallgren:
This PR adds the new lookup framwork as a set of classes residing in the
Puppet::Pops::Lookup module. The old lookup framework remains in place
under Puppet::DataProviders and Puppet::Plugins::Dataproviders to ensure
100% backward compatibility for the duration of the 4.x versions since many
of the classes had methods marked @api public.

Both the environment_data_provider setting in the environment.conf and the
data_provider setting in the metadata.json are still honored. If set to 'hiera'
or 'function', they will be rerouted to the new framework. Other custom
providers will continue to work but use the old framework.

The backends 'yaml' and 'json', when found in a hiera.yaml, will now use the
new `data_hash` functions yaml_data and json_data. Other custom backends
will continue to use the old framework.

Legacy data functions, taking zero arguments and returning a hash-map, are
now treated as a special variant of a `data_hash` function and will continue
to function as such although they are now called from the new framework.

All access to hiera.conf, custom backends, custom providers, or legacy
functions will render a deprecation warning unless Puppet[:strict] is set to
'off'.


MosesMendoza:
Hi @thallgren - I'm temporarily adding `don't merge` label until we can get pup 4.8.0 / PA 1.8.0 sorted tomorrow PDT - just to prevent landing this in that release

MosesMendoza:
"unfrozen" :)

lib/puppet/data_providers.rb
```diff
@@ -25,6 +27,9 @@ def self.lookup_in_module(name, lookup_invocation, merge)
   end
 
   def self.lookup_adapter(lookup_invocation)
+    unless Puppet[:strict] == :off
+      Puppet.deprecation_warning('The method Puppet::DataProviders.lookup_adapter is deprecated and will be removed in the next major release of Puppet.')
```
hlindberg: Does this use the "warn once" logic? Otherwise there is the risk of getting a lot of deprecation warnings in the log.

thallgren: Not likely, since this whole file is now deprecated. The only chance of someone calling here is basically if they'd written something other than data providers (another lookup framework). Puppet no longer calls this method.

hlindberg: ok - fine.

lib/puppet/application/lookup.rb
```diff
@@ -297,11 +297,13 @@ def main
...
         result = Puppet::Pops::Lookup.lookup(keys, type, options[:default_value], use_default_value, merge_options, lookup_invocation)
         puts renderer.render(result) unless explain
-      rescue Puppet::DataBinding::LookupError
+      rescue Puppet::DataBinding::LookupError => e
+        lookup_invocation.report_text { e.message }
         exit(1) unless explain
       end
       puts format == :s ? lookup_invocation.explainer.to_s : renderer.render(lookup_invocation.explainer.to_hash) if explain
     end
+    exit(0)
```
hlindberg: In the strictest sense this is an API break. The intent is that the lookup command line can be used in scripts and that it will error if not returning a value. If that feature should be carried over it can be supported with an extra flag controlling what it should do on conditions like 'not found'.

lib/puppet/pops/lookup/context.rb
```diff
@@ -1,86 +1,136 @@
...
+
+  # Resolve interpolation expressions in the given value
+  # @param [Object] value
+  # @return [Object] the value with all interpolation expressions resolved
+  def interpolate(value)
+    @function_context.interpolate(value, @lookup_invocation, true)
+  end
+
+  def not_found
+    throw :no_such_key
```
hlindberg: should this also explain the not found, or is that the user's responsibility, or is that handled elsewhere?

lib/puppet/data_providers.rb
```diff
@@ -1,4 +1,6 @@
+# TODO: API 5.0, remove this module
```
hlindberg: All the deprecations of the data providers (v4) are in the ticket PUP-6514. This commit should have its tag changed from PUP-6780 to PUP-6514.

lib/puppet/data_providers/hiera_config.rb
```diff
@@ -99,25 +103,35 @@ def create_configured_data_providers(lookup_invocation, parent_data_provider)
...
         datadir = @config_root + (he['datadir'] || default_datadir)
-        data_providers[name] = create_data_provider(parent_data_provider, provider_factory, name,
-          provider_factory.resolve_paths(datadir, original_paths, paths, lookup_invocation))
+        provider_name = he['backend']
+        data_providers[name] = case provider_name
+        when 'json', 'yaml'
+          Puppet::Pops::Lookup::DataHashFunctionProvider.new(name, parent_data_provider, "#{provider_name}_data", {},
+            resolve_paths(datadir, original_paths, lookup_invocation, ".#{provider_name}"))
+        else
+          # TODO: Remove support for injected provider factories
```
thallgren: No, not really. The comment concerns the specific part of it that is being used here.

lib/puppet/functions/json_data.rb
```diff
@@ -0,0 +1,18 @@
+# @since 4.8.0
+#
+Puppet::Functions.create_function(:json_data) do
+  dispatch :json_data do
+    param 'Struct[{path=>String[1]}]', :options
+    param 'Puppet::LookupContext', :context
+  end
```
thallgren: This particular function and it's yaml counterpart, both take exactly one option, the path. All other options (including a misspelled paht) are incorrect. So why allow "free form" options here?

lib/puppet/pops/lookup/context.rb
thallgren: It is handled elsewhere. An explanation here would be redundant.

thallgren:
PR updated with spelling corrections. You ask about the ability to specify options in the lookup.yaml. It's there. Look at lines 104 and 190-193 in lookup_config.rb.

lib/puppet/functions/json_data.rb
hlindberg: The yaml and json readers does not need options, but other functions that provide the entire data hash may need additional optional arguments.

hlindberg:
sorry if some comments appear to rehash the same thing, there was a pending review that I managed to not finish. 
A Rebase is also needed.

lib/puppet/pops/lookup/configured_data_provider.rb
```diff
@@ -0,0 +1,77 @@
...
+require_relative 'data_provider'
+require 'puppet/data_providers/hiera_config'
+
+module Puppet::Pops
+module Lookup
+# @api private
+class ConfiguredDataProvider
+  include DataProvider
+
+  # @param [LookupConfig,nil] config the configuration
```
hlindberg: name and type in the wrong order

lib/puppet/pops/lookup/configured_data_provider.rb
```diff
@@ -0,0 +1,77 @@
+require_relative 'lookup_config'
+require_relative 'data_provider'
+require 'puppet/data_providers/hiera_config'
+
+module Puppet::Pops
+module Lookup
+# @api private
+class ConfiguredDataProvider
```
hlindberg: Is this essentially "AbstractV4DataProvider" (new base class for the ones being deprecated), or the new baseclass for both the old and new (V5) function providers? I find the name confusing - "ConfiguredDataProvider" as opposed to what "NonConfigured" ? If this is a temporary thing (for v4) then nevermind.

lib/puppet/pops/lookup/context.rb
```diff
@@ -1,86 +1,136 @@
+require_relative 'interpolation'
+
 module Puppet::Pops
-  module Lookup
-    class Context
-      include Types::PuppetObject
+module Lookup
+# A FunctionContext is created for each unique hierarchy entry and adapted to the Compiler (and hence shares
+# the compilers life-cycle).
```
hlindberg: "compiler's"

lib/puppet/pops/lookup/data_hash_function_provider.rb
```diff
@@ -0,0 +1,100 @@
...
+    ctx = function_context(lookup_invocation, location)
+    ctx.data_hash ||= parent_data_provider.validate_data_hash(self, call_data_hash_function(ctx, lookup_invocation, location))
+  end
+
+  def call_data_hash_function(ctx, lookup_invocation, location)
+    ctx.function.call(lookup_invocation.scope, options(location), Context.new(ctx, lookup_invocation))
+  end
+end
+
+class LegacyDataHashFunctionProvider < DataHashFunctionProvider
```
hlindberg: I think this is for the v4 experimental data() function. Maybe comment that this should be removed (other code had such comments).

lib/puppet/pops/lookup/data_provider.rb
```diff
@@ -0,0 +1,84 @@
+module Puppet::Pops
+module Lookup
+# @api private
+module DataProvider
+  # The Pcore type for all keys and subkeys in a data hash.
+  TYPE_DATA_KEY = Types::PVariantType.new([
+    Types::PStringType::NON_EMPTY,
+    Types::PBooleanType::DEFAULT,
```
hlindberg: This is a bit surprising - is this a legacy from hiera (I think it means you can bind values to the boolean keys true and false)?

lib/puppet/pops/lookup/data_provider.rb
```diff
@@ -0,0 +1,84 @@
...
+    Types::PNumericType::DEFAULT
+  ])
+
+  # The Pcore type for all values and sub-values in a data hash.
+  TYPE_DATA_VALUE = Types::PVariantType.new([
+    Types::PScalarType::DEFAULT,
+    Types::PUndefType::DEFAULT,
+    Types::PHashType::DEFAULT,
+    Types::PArrayType::DEFAULT
+  ])
```
hlindberg: Why are values limited to those types? I can imagine inventing a data format that is based on rich data types.

lib/puppet/pops/lookup/data_provider.rb
```diff
@@ -0,0 +1,84 @@
...
+  # Asserts that _data_hash_ is a valid hash.
+  #
+  # @param data_provider [DataProvider] The data provider that produced the hash
+  # @param data_hash [Hash{String=>Object}] The data hash
+  # @return [Hash{String=>Object}] The data hash
+  def validate_data_hash(data_provider, data_hash)
+    Types::TypeAsserter.assert_instance_of(nil, Types::PHashType::DEFAULT, data_hash) { "Value returned from #{data_provider.name}" }
+    data_hash.each_pair { |k, v| validate_data_entry(data_provider, k, v) }
+    data_hash
+  end
```
hlindberg: We should probably allow anything except Runtime[Ruby]

thallgren: I deliberately limited this to Data to make deep_merge and data dig work without surprises. Deep merge requires that values comparable and that the all keys are scalar and all values are scalar, arrays, or hashes. Same thing applies when digging using sub-keys.
Another aspect to consider is the explanation support. A limited set of data is easy to represent. Adding rich data makes it harder, both to figure out a human readable representation and a representation that can be rendered with JSON (or other).

lib/puppet/pops/lookup/configured_data_provider.rb
thallgren: This is a data provider that corresponds to a hierarchy entry in the configuration. It's not related to V4.

lib/puppet/pops/lookup/data_provider.rb
hlindberg: While that is true, we are going to deliver Sensitive and Encrypted data (and more) via lookup. It must be possible to have all Pcore represented types as values (except the open ended Runtime type). After some thinking: The design must take that into account.

lib/puppet/pops/lookup/data_provider.rb
hlindberg: As noted elsewhere, any Pcore representable type except RuntimeType should be allowed as a value.

lib/puppet/pops/lookup/configured_data_provider.rb
hlindberg: Naming is hard. Have no better suggestion atm.

spec/unit/functions/lookup_spec.rb
```diff
@@ -1,733 +1,1253 @@
...
-  # - Return the name of all Notify resources in that catalog
-  #
-  # @param fmt [String] The puppet interpolated string used when creating the notify title
-  # @param *args [String] splat of args that will be concatenated to form the puppet args sent to lookup
-  # @return [Array<String>] List of names of Notify resources in the resulting catalog
-  #
-  def assemble_and_compile(fmt, *lookup_args, &block)
-    assemble_and_compile_with_block(fmt, "'no_block_present'", *lookup_args, &block)
-  end
+  context "using static 'Hiera version 4' fixture" do
```
hlindberg: Not a very good label - test implementation detail rather than part of an example title.

spec/unit/functions/lookup_spec.rb
```diff
@@ -1,733 +1,1253 @@
...
+          rescue Puppet::Error
+          end
+          expect(lookup_invocation.explainer.explain).to eq(<<EOS)
 Invalid key "lookup_options"
 EOS
+        end
       end
-    end
 
-    it 'will explain that "lookup_options" is
```
hlindberg: Not a very good example label

hlindberg: As per discussion elsewhere; should allow any Pcore type except Runtime

hlindberg: The tests are written from a test implementation standpoint rather than feature/example standpoint. That makes these tests hard to read and map to the complete set of features / desired behavior.

* "Data in the environment is found"
   * "When there is a lookup.conf"
     * "In which case a "hiera.yaml" is ignored
  * "When there is a hiera.yaml (but no lookup.conf)"
* "Old constructs are deprecated such that"
  * "Use of a hiera.yaml v4 issues deprecation"
  * "Use of function data provider issues deprecation with helpful message what to change to"
  * "Use of ... data provider..."
* "Data in a module is found"
  * "..."

etc

hlindberg: Looks like this, and the two following contexts (lookup_key, and data_dig) would benefit from using a common set of tests "behaves like lookup in general" wrt. env overrides, etc. That would make it easier to see commonality and differences. (Maybe also break up this long spec into several)

hlindberg: Are there additional tests that assert the separation of caches?

hlindberg: "will merge not merge" ??

lib/puppet/pops/lookup/lookup_adapter.rb
```diff
@@ -79,7 +84,7 @@ def lookup_global(name, lookup_invocation, merge_strategy)
     lookup_invocation.with(:global, terminus) do
       catch(:no_such_key) do
         return lookup_invocation.report_found(name, Puppet::DataBinding.indirection.find(name,
-            { :environment => environment, :variables => lookup_invocation.scope, :merge => merge_strategy }))
+                                                                                         { :environment => environment, :variables => lookup_invocation.scope, :merge => merge_strategy }))
```
hlindberg: formatting looks off

lib/puppet/pops/lookup/lookup_config.rb
```diff
@@ -0,0 +1,211 @@
...
+  def validate_config(config)
+    Types::TypeAsserter.assert_instance_of(["The Lookup Configuration at '%s'", @config_path], self.class.config_type, config)
+    config[KEY_HIERARCHY].each do |he|
+      name = he[KEY_NAME]
+      case FUNCTION_KEYS.count { |key| he.include?(key) }
+      when 1
+        # OK
+      when 0
+        raise Puppet::DataBinding::LookupError,
+          "#{@config_path}: One of #{combine_strings(FUNCTION_KEYS - [KEY_LEGACY_DATA_HASH])} must defined in hierarchy '#{name}'"
```
thallgren: The error message is prefixed with the full path of the file which makes it easy to determine if it concerns an environment or a module. The file in question then has an explicit `hierarchy:` key consisting of named entries. Referring to such entries as "in hierarchy 'common'" should be very clear IMHO. Clearer than "lookup configuration entry 'common'".

lib/puppet/pops/lookup/data_provider.rb
thallgren: I suggest we create a separate "Rich data in lookup" ticket for that. We need to discuss how to handle the data. Do we merge sensitive data? Do we navigate into objects?

thallgren:
Pushed commit with requested changes. It's just one commit but I also did a needed rebase onto current master.

spec/unit/functions/lookup_spec.rb
hlindberg: it makes it close to impossible to check that all features are covered by tests. Agree that we do not have to improve the v4 tests.

thallgren: An IDE that supports code-coverage and gives you colored margins in code editors and percentage summaries on folders is not half bad when checking things like that. Just sayin' ;-)

PR #5350, opened 2016-10-27
(PUP-6847) Omit whitespace from provider keyword in network_device

abottchen:
Prior to this update, any trailing whitespace left in a device
configuration file will cause `puppet device` to fail with
a cryptic message because it is using the string to construct
the path for an require statement.  Since require statements
should never have spaces in them, it makes more sense to have the
regex only accept the non-whitespace characters it finds in the
confiugration file.

### DIFF
--- a/lib/puppet/util/network_device/config.rb
+++ b/lib/puppet/util/network_device/config.rb
@@ -59,7 +59,7 @@ class Puppet::Util::NetworkDevice::Config
             device.options = { :debug => false }
             Puppet.debug "found device: #{device.name} at #{device.line}"
             devices[name] = device
-          when /^\s*(type|url|debug)(\s+(.+))*$/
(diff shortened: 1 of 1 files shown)


branan:
@abottchen Thanks! Two things:

1) Puppet tickets needs to start with PUP. You should be able to just move the PA ticket into the PUP project and edit your commit message
2) I'm personally not a fan of how many really complicated regular expressions we have in Puppet. I think this would be clearer if you leave the regex as-is and `strip()` it after the fact. Others might disagree with me, so this is really up to you if you want to change it ¯\_(ツ)_/¯

tphoney:
Thanks for the PR looks good to me. Re-architecting everything would be great, but this is a necessary fix and a great catch.  👍

PR #5354, opened 2016-10-28
(PUP-6188) Add utf-8 encoding to various calls throughout puppet

MosesMendoza:
This PR builds on / supersedes #4898 from @Iristyle 

[PUP-6188](tickets.puppetlabs.com/browse/PUP-6188) tracks an effort to ensure Puppet's various file system interactions are specifying utf-8 encoding where possible/applicable, possibly in `Puppet::FileSystem.open` and `Puppet::FileSystem.exclusive_open`, but notes the problematic situation resulting from the reordering of method arguments from `Puppet::FileSystem::FileImpl#open` to [`::File.open`](https://ruby-doc.org/core-2.3.1/File.html#method-c-open):

> - Rubys `perm` in our call is renamed as `mode`, and is moved from position 2 to position 1
> - Rubys `mode` in our call is renamed as `options`, and is moved from position 1 to position 2
> - We don't support the passing in of Rubys `opt` which is where one would set ``:encoding => 'utf-8'``

This is further complicated by the fact that ruby allows specifying mode, encoding, and other options a number of different ways via these parameters. For example, encoding and mode, which can be specified
- via `opt` (ruby optional param 3) in an options hash, ie `:mode => 'r', :encoding => 'BOM|UTF-8'` 
- via a `mode` string (ruby param 2), containing one of:
  - an IO Open Mode: `"r"`, `"r+"`, `"w"`
  - an IO Open Mode _with encoding_: `"r:IBM437"`, `"w:BOM|UTF-8"`
  - a list of `|` separated File Constants: `File::CREATE|File::APPEND`

This means it's not simply a matter of extending the definition of `Puppet::FileSystem::FileImpl#open` and `Puppet::FileSystem::FileImpl#exclusive_open` to support an options hash (and thus specifying `:encoding => 'UTF-8'` because such options may collide with those already specified by a user via the `mode` string (ruby param 2). We could probably handle this somewhat gracefully (a rudimentary first pass at this is in https://github.com/MosesMendoza/puppet/commit/ccb65927b22544239adbc49a7b054cc91016bdde) but it seems like less risk of backwards-incompatibility and less complexity overall to simply update the callers of `Puppet::FileSystem` inside of puppet to leverage the ability to pass encoding in via the `mode` string.

This PR does so in several places, explicitly specifying UTF-8 as part of the mode string and updating associated tests. When updating the `puppet config` face, it was discovered that the `:set` action has no unit testing associated with it (at least not that I can find) - all testing is at the lower level of `Puppet::Settings::Inifile`, so this PR also includes a commit that adds initial basic spec coverage of `puppet config set`. Additionally, where encountered, `File.read` was replaced with `Puppet::FileSystem.read` - it seemed as if some of these places existed before `Puppet::FileSystem` was implemented.

After discussion with @Iristyle it was determined that much of the test coverage here likely deserves substantial improvement. Specifically, creating new tests that actually create files containing unicode-only characters, passing them through the code under test, and validating the fidelity of the data actually remains intact. I'm not sure if this PR should be gated on the addition of such tests, but I don't disagree that they're justified.

### DIFF
--- a/spec/unit/face/config_spec.rb
+++ b/spec/unit/face/config_spec.rb
@@ -64,7 +64,7 @@ syslogfacility = file
       subject.set('foo', 'bar')
     end
 
-    it "sets the supplied config/value" do
+    it "sets the supplied config/value in the default section (main)" do
       Puppet::FileSystem.stubs(:open).with(path, anything, anything).yields(StringIO.new)
       config = Puppet::Settings::IniFile.new([Puppet::Settings::IniFile::DefaultSection.new])
       manipulator = Puppet::Settings::IniFile::Manipulator.new(config)
--- a/spec/unit/face/config_spec.rb
+++ b/spec/unit/face/config_spec.rb
@@ -44,6 +44,48 @@ syslogfacility = file
     subject.print('all')
   end
 
+  context "when setting config values" do
+    let(:config_file) { '/foo/puppet.conf' }
+    let(:path) { Pathname.new(config_file) }
+    before(:each) do
+      Puppet[:config] = '/foo/puppet.conf'
+      Puppet::FileSystem.stubs(:pathname).with(config_file).returns(path)
+      Puppet::FileSystem.stubs(:touch)
+    end
+
+    it "writes to the correct puppet config file" do
+      Puppet::FileSystem.expects(:open).with(path, anything, anything)
+      subject.set('foo', 'bar')
+    end
+
+    it "creates a config file if one does not exist" do
+      Puppet::FileSystem.stubs(:open).with(path, anything, anything).yields(StringIO.new)
+      Puppet::FileSystem.expects(:touch).with(path)
+      subject.set('foo', 'bar')
+    end
+
+    it "sets the supplied config/value" do
+      Puppet::FileSystem.stubs(:open).with(path, anything, anything).yields(StringIO.new)
+      config = Puppet::Settings::IniFile.new([Puppet::Settings::IniFile::DefaultSection.new])
+      manipulator = Puppet::Settings::IniFile::Manipulator.new(config)
+      Puppet::Settings::IniFile::Manipulator.stubs(:new).returns(manipulator)
+
+      manipulator.expects(:set).with("main", "foo", "bar")
+      subject.set('foo', 'bar')
+    end
+
+    it "sets the value in the supplied section" do
+      Puppet::FileSystem.stubs(:open).with(path, anything, anything).yields(StringIO.new)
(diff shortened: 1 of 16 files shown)


spec/unit/face/config_spec.rb
```diff
@@ -44,6 +44,53 @@ module PuppetFaceSpecs
...
+      Puppet::FileSystem.stubs(:pathname).with(config_file).returns(path)
+      Puppet::FileSystem.stubs(:touch)
+    end
+
+    it "writes to the correct puppet config file" do
+      Puppet::FileSystem.expects(:open).with(path, anything, anything)
+      subject.set('foo', 'bar')
+    end
+
+    it "creates a config file if one does not exist" do
```
Iristyle: This test is a little misleading (especially with the expectations written in this order).

That a file gets created is a side effect of the `touch` call - but the problem here with the test is that there's no assertion that the file didn't exist prior to calling `subject.set`

Iristyle: Rather than stubbing, I think it would be better to just use the disk to prove file doesn't exist, `set` gets called, file exists.

spec/unit/face/config_spec.rb
```diff
@@ -44,6 +44,53 @@ module PuppetFaceSpecs
...
+      Puppet::FileSystem.stubs(:open).with(path, anything, anything).yields(StringIO.new)
+      Puppet::FileSystem.expects(:touch).with(path)
+      subject.set('foo', 'bar')
+    end
+
+    it "sets the supplied config/value" do
```
Iristyle: "in the default main section"

spec/unit/face/config_spec.rb
MosesMendoza: 👍 . updating

Iristyle:
e6969f5 commit message talks about the `BOM` bit, but doesn't specify it in the code

MosesMendoza:
Thanks - updated message to remove reference to 'BOM'

Iristyle:
Some additional comments based on what I've looked at locally:

* The INI file UTF-8 work is being merged in #5361 prior to this code
* 2636a60 (SSL changes) should really be to `ASCII`, not `UTF-8`.  While `UTF-8` will work, it includes a broader range of characters than strictly necessary for the SSL files in question.
* 6a00f40 is a set of changes that I don't think we should make.  System files should generally be written in the systems encoding.  I don't think that there are any guarantees around `crontab` files on AIX or Solaris being `UTF-8`, so I suspect that we should leave the Ruby default behavior of `Encoding.default_external` here, which should match the systems expectations.

PR #5355, opened 2016-10-28
(pup-5000) acceptance test agent specified environment fact

er0ck:
```
(PUP-5000) acceptance: test agent specified environment fact
* test that the agent specified environment fact does the right thing
  * enc's do not receive the list of facts, so we don't have to test it
  * in there, making this test a bit more simple than first expected.
* this might be covered in unit tests
  * but there is NOT a test that ensures that we get the right
   environment from the options on the CLI and puppet.conf

(PUP-5000) acceptance: allow create_sitepp to create in production env

* previous instance of create_sitepp assumed a path to a tmp_environment
 the test in pup-5000 requires a sitepp in production.  this change
 tests if the name of the environment sent in is "production" and if
 so, creates a sitepp in stock environmentpath.  This needs to be
 torn-down manually.
```

### DIFF
--- a/acceptance/lib/puppet/acceptance/environment_utils.rb
+++ b/acceptance/lib/puppet/acceptance/environment_utils.rb
@@ -399,7 +399,11 @@ module Puppet
 
       # create sitepp in a tmp_environment as created by mk_tmp_environment_with_teardown
       def create_sitepp(host, tmp_environment, file_content)
-        file_path = File.join('','tmp',tmp_environment,'manifests','site.pp')
+        if tmp_environment == 'production'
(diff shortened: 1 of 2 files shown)


er0ck:
so we might not need this test in acceptance.  no reason to check grabbing facts from puppetdb, here.  and specs have good tests for this fact:
https://github.com/puppetlabs/puppet/blob/master/spec/integration/indirector/facts/facter_spec.rb#L82

whopper:
@er0ck 👍 on leaving this at the spec level, if you're still cool with that.

er0ck:
yep.  killing it 🔥

PR #5356, opened 2016-11-02
(PA-652) Bump Puppet version to 4.8.1

Magisus:
This updates the Puppet version to 4.8.1 following the release of 4.8.0
as part of puppet-agent 1.8.0.

### DIFF
--- a/lib/puppet/version.rb
+++ b/lib/puppet/version.rb
@@ -7,7 +7,7 @@
 
 
 module Puppet
-  PUPPETVERSION = '4.8.0'
+  PUPPETVERSION = '4.8.1'
 
   ##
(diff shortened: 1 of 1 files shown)


Magisus:
Travis commit checker is failing because it requires a PUP ticket. Do we want to allow to fail, or switch to maint? Might it be worthwhile to add PA tickets to the whitelist?

joshcooper:
For this sort of thing I usually use `(packaging)` as the prefix.

MosesMendoza:
+1 to `(packaging)`

Magisus:
Updated

MosesMendoza:
👍

PR #5359, opened 2016-11-03
(PUP-6534) Use parallel specs in AppVeyor

glennsarti:
Previously the AppVeyor spec tests were executed sequentially.  Now that all of
the unit tests are functioning correctly when run in parallel mode, this commit
changes the execution command for AppVeyor to use parallel spec with two
threads.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,7 +15,7 @@ test_script:
   - bundle install --jobs 4 --retry 2 --without development extra
   - type Gemfile.lock
   - ruby -v
-  - bundle exec rspec spec
+  - bundle exec rake parallel:spec[2]
 
 on_failure:
   - appveyor PushArtifact .\spec_order.txt


Magisus:
Since Appveyor does indeed seem to pass with the change, 👍 from me.

joshcooper:
That is awesome, nice work 🎆

Magisus:
Does this mean we can update our CI jobs to run these in parallel too?

Iristyle:
Yes, huge thanks to Wintern @DLuCJ who worked on making sure that state was untangled and that Windows could run specs in parallel this summer.

I thought we still had a little more work to do to finish it off, but guess not... this is great!

/cc @craig-gomes @puppetlabs/windows

peterhuene:
🎉 🎉 🎉

ferventcoder:
💯

PR #5361, opened 2016-11-04
(PUP-6876) Fix INI parsing for UTF-8 keys / values / sections and `config set`

Iristyle:
- Previously, INI files were using the `\w` regex character class to
   match the "key" in a INI file line like:

```ini
   key = value
```

   Unfortunately, `\w` is limited to ASCII characters `[a-zA-Z0-9_]` in
   Ruby and will fail to match properly on any Unicode word
   characters, such as:

```ini
   Aۿᚠ𠜎 = value
```

 - The \w regex character class was also used to match the section in a INI file line like:

```ini
   [Aۿᚠ𠜎]
```

   Unfortunately, due to the same limitation, an INI section like the above would not parse properly.

 - Change the `\w` character class to `[[:word]]` which Ruby describes as a:

> character in one of the following Unicode general categories Letter, Mark, Number, Connector_Punctuation

 - Supply tests demonstrating previously failing behavior

 - In addition to fixing any code paths using the INI file parser,
   this also fixes the CLI `config set` and `config print` commands
   when specifying existing keys.  Without the fix, duplicate sections
   may be written to an existing config file - which can then cause
   problems when retrieving values.

### DIFF
--- a/spec/unit/settings/ini_file_spec.rb
+++ b/spec/unit/settings/ini_file_spec.rb
@@ -4,6 +4,13 @@ require 'stringio'
 require 'puppet/settings/ini_file'
 
 describe Puppet::Settings::IniFile do
+  # different UTF-8 widths
+  # 1-byte A
+  # 2-byte ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
+  # 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
+  # 4-byte 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
+  let (:mixed_utf8) { "A\u06FF\u16A0\u{2070E}" } # Aۿᚠ𠜎
+
   it "preserves the file when no changes are made" do
     original_config = <<-CONFIG
     # comment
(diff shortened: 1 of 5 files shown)


Iristyle:
Should be merged prior to #5354

spec/unit/settings/ini_file_spec.rb
```diff
@@ -4,6 +4,13 @@
 require 'puppet/settings/ini_file'
 
 describe Puppet::Settings::IniFile do
+  # different UTF-8 widths
+  # 1-byte A
+  # 2-byte ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
+  # 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
+  # 4-byte 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
+  let (:mixed_utf8) { "A\u06FF\u16A0\u{2070E}" } # Aۿᚠ𠜎
```
MosesMendoza: I haven't seen `{`..`}` before in the unicode representation. `\u2070E` and `\u{2070E}` are indeed different characters - what do the brackets mean?

Iristyle: From https://ruby-doc.org/core-2.3.0/doc/syntax/literals_rdoc.html#label-Strings

```
\u{nnnn ...}   Unicode character(s), where each nnnn is 1-6 hexadecimal digits ([0-9a-fA-F])
```

Detailed a bit more at https://leejava.wordpress.com/2009/03/11/unicode-escape-in-ruby/

> \u is followed by exactly four hexadecimal digits (letters can be upper- or lowercase), which represent a Unicode codepoint between 0000 and FFFF.

... 


> A second form of the \u escape is followed by an open curly brace, one to six hexadecimal digits, and a close curly brace. The digits between the braces can represent any Unicode codepoint between 0 and 10FFFF, and leading zeros can be dropped in this form

PR #5362, opened 2016-11-04
(maint) Fix TravisCI runs on Ruby 1.9.3

Iristyle:
- The webmock gem has always depended on the addressable gem >= 2.3.6
   and therefore it has been installed transitively.  As of 2.5.0, the
   addressable gem released on Nov 4, 2016 has added a new dependency
   on the public_suffix gem, which requires Ruby 2.

   Since jobs still run against Ruby 1.9.3, pin addressable to a prior
   version that doesn't need the public_suffix gem.

 - Note that even when using `bundle install --without development`
   this problem occurs, even though the webmock gem should be
   excluded from installation. This is a bundler behavior.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,6 +53,9 @@ group(:development, :test) do
 
   gem 'rdoc', "~> 4.1", :platforms => [:ruby]
 
+  # webmock requires addressable as as of 2.5.0 addressable started
+  # requiring the public_suffix gem which requires Ruby 2
+  gem 'addressable', '< 2.5.0'
   gem 'webmock', '~> 1.24'
   gem 'vcr', '~> 2.9'
 end


McdonaldSeanp:
👍

branan:
@Iristyle is there a ticket filed to remove this pin when we remove the Ruby 1.9.3 support? (Presumably a 5.0.0 ticket, but if there's a better way to categorize "ruby 1.9.3 removal" I'm also OK with it there)

Magisus:
Since this now passes our commit hooks which were failing before the pin, 👍

Iristyle:
https://tickets.puppetlabs.com/browse/PUP-6894 added to track pin removal

PR #5363, opened 2016-11-07
(PUP-6897) Fix for agent mishandling in pcore-gen and cor-change tests

camlow325:
Previously, the pcore_generate_env_isolation.rb and
corrective_change_via_puppet.rb tests were failing when run in a CI
configuration with more than one agent in it.  In each case, the test
changed some state after an agent was run on the first node which was
unintentionally picked up as the agent was run on the second node.  This
commit changes both tests to perform all related actions on each of the
nodes before any state changes are made.

### DIFF
--- a/acceptance/tests/language/pcore_generate_env_isolation.rb
+++ b/acceptance/tests/language/pcore_generate_env_isolation.rb
@@ -52,10 +52,14 @@ test_name 'C98345: ensure puppet generate assures env. isolation' do
                        'did not produce environment isolation issue as expected')
         end
       end
-      step 'generate pcore files' do
-        on(master, puppet("generate types --environment #{tmp_environment}"))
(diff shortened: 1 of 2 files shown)


camlow325:
@er0ck and @james-stocks - would appreciate a review from one or both of you on this.  These tests are failing in the puppetserver pipeline and may block the Puppet Server 2.7.0 planned for tomorrow if not fixed today.

@cprice404, @jpinsonault - ping for a dev review from one of y'all.

er0ck:
lgtm 👍 

good catch on the pcore generate step.  dunno why that ended up in the agents loop

camlow325:
@er0ck - since it sounds like you're good with this, would you be okay with it being merged so that we can unblock puppetserver CI failures?

er0ck:
merged up to stable at: 075679f1302e6c011d51fd75f91606cd79609881

PR #5364, opened 2016-11-07
(PUP-6898) Fix sensitive_data_type test validation of syslog output

camlow325:
Previously, the sensitive_data_type test, when attempting to validate
the presence / absence of expected content in the syslog, performed
assertions against the full Beaker logger instead.  This would cause
assertions looking for the absence of sensitive content to fail for the
second agent run performed because the assertions themselves would
appear in the logger output.

This commit changes the syslog assertion logic to inspect just the
content of the syslog - without bringing in the rest of the current
Beaker logger.  The commit also only runs through the assertions which
confirm that sensitive data does not appear in the syslog and not the
assertions which confirm that other debug messages from the agent run
appear in the syslog.  This is done because the agent is run with a
"--debug" flag, which should cause agent log messages to go to the
console and not to the syslog anyway.

### DIFF
--- a/acceptance/tests/language/sensitive_data_type.rb
+++ b/acceptance/tests/language/sensitive_data_type.rb
@@ -64,6 +64,15 @@ extend Puppet::Acceptance::PuppetTypeTestTools
   sitepp_content = generate_manifest(test_resources)
   assertion_code = generate_assertions(test_resources)
 
+  refutation_resources = test_resources.collect do |assertion_group|
+    refutation_group = assertion_group.clone
+    refutation_group[:assertions] = assertion_group[:assertions].select do |assertion|
+      assertion.has_key?(:refute_match)
+    end
+    refutation_group
+  end
+  refutation_code = generate_assertions(refutation_resources)
+
   create_sitepp(master, tmp_environment, sitepp_content)
 
   step "run agent in #{tmp_environment}, run all assertions" do
@@ -75,13 +84,20 @@ extend Puppet::Acceptance::PuppetTypeTestTools
           run_assertions(assertion_code, result)
         end
 
-        if agent['platform'] =~ /fedora|centos|el|redhat|scientific|ubuntu|debian|cumulus/
-          step "assert no redacted data in syslog" do
-            #sigh.  gee, thanks for the help, beaker!
-            result = dump_puppet_log(agent)
-            raise if result.length < 3
-            string_io_index = 2
-            run_assertions(assertion_code, result[string_io_index].string)
+        step "assert no redacted data in syslog" do
+          #sigh.  gee, thanks for the help, beaker!
+          syslogfile = case agent['platform']
+                         when /fedora|centos|el|redhat|scientific/ then
+                           '/var/log/messages'
+                         when /ubuntu|debian|cumulus/ then
+                           '/var/log/syslog'
+                         else
+                           nil
+                       end
+          if syslogfile
+            result = agent.exec(Command.new("tail -n 100 #{syslogfile}"),
+                                :acceptable_exit_codes => [0, 1]).stdout.chomp
+            run_assertions(refutation_code, result)
           end
         end
 
--- a/acceptance/tests/language/sensitive_data_type.rb
+++ b/acceptance/tests/language/sensitive_data_type.rb
@@ -64,6 +64,27 @@ extend Puppet::Acceptance::PuppetTypeTestTools
   sitepp_content = generate_manifest(test_resources)
   assertion_code = generate_assertions(test_resources)
 
+  # Make a copy of the full set of 'test_resources' but filtered down to include
+  # only the assertions of type ':refute_match'.  So for example, where the
+  # 'test_resources' array might have an entry like this...
+  #
+  #  {:type => 'notify', ...
+  #   :assertions => [{:refute_match => 'sekrit1'},
+  #                   {:assert_match => "1:#{notify_redacted}"}]}
+  #
+  # ... the ':assert_match' entry would be filtered out in the new
+  # 'refutation_resources' array, producing:
+  #
+  #  {:type => 'notify', ...
+  #   :assertions => [{:refute_match => 'sekrit1'}]}
+  #
+  # This is done so that when validating the syslog output, we can refute the
+  # existence of any of the sensitive info in the syslog without having to
+  # assert that redacted info is in the syslog.  The redacted info appears in
+  # the console output from the Puppet agent run - by virtue of including a
+  # '--debug' flag on the agent command line - whereas the redacted info is not
+  # expected to be piped into the syslog.
+
   refutation_resources = test_resources.collect do |assertion_group|
     refutation_group = assertion_group.clone
     refutation_group[:assertions] = assertion_group[:assertions].select do |assertion|
@@ -87,16 +108,16 @@ extend Puppet::Acceptance::PuppetTypeTestTools
         step "assert no redacted data in syslog" do
           #sigh.  gee, thanks for the help, beaker!
           syslogfile = case agent['platform']
-                         when /fedora|centos|el|redhat|scientific/ then
+                         when /fedora|centos|el|redhat|scientific/
                            '/var/log/messages'
-                         when /ubuntu|debian|cumulus/ then
+                         when /ubuntu|debian|cumulus/
                            '/var/log/syslog'
                          else
+                           logger.warn "Could not determine syslog for #{agent['platform']}, so skipping syslog validation"
                            nil
                        end
           if syslogfile
-            result = agent.exec(Command.new("tail -n 100 #{syslogfile}"),
-                                :acceptable_exit_codes => [0, 1]).stdout.chomp
+            result = on(agent, "tail -n 100 #{syslogfile}").stdout.chomp
             run_assertions(refutation_code, result)
           end
         end


acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -75,13 +82,20 @@
           run_assertions(assertion_code, result)
         end
 
-        if agent['platform'] =~ /fedora|centos|el|redhat|scientific|ubuntu|debian|cumulus/
-          step "assert no redacted data in syslog" do
-            #sigh.  gee, thanks for the help, beaker!
-            result = dump_puppet_log(agent)
```
camlow325: The return value from `dump_puppet_log` is the logger, which can have other data from the Beaker test run in it, and not strictly just the syslog contents.

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -75,13 +82,20 @@
...
+                           '/var/log/messages'
+                         when /ubuntu|debian|cumulus/ then
+                           '/var/log/syslog'
+                         else
+                           nil
+                       end
+          if syslogfile
+            result = agent.exec(Command.new("tail -n 100 #{syslogfile}"),
+                                :acceptable_exit_codes => [0, 1]).stdout.chomp
+            run_assertions(refutation_code, result)
```
camlow325: Maybe it would also make sense to rework this test to see that if `puppet agent` is run with the `--logdest` set to `syslog` instead of defaulting to `console` that both the refutations of sensitive content pass and that expected debug output that is not sensitive is present in the syslog?  Seems like at least ensuring that the syslog doesn't have sensitive content in it may be enough, though, since both the assertions / refutations are run against the puppet agent console output as well.

camlow325:
@er0ck - here's one more test failure that we ran into with puppetserver CI today that we'd like to get fixed up as soon as possible so that we can unblock our Puppet Server 2.7.0 release (planned for tomorrow).  Could you take a look at this as soon as you can?

@cprice404, @jpinsonault - ping for a second 'dev' review as well.

jpinsonault:
As far as I understand the code and the goal, I think this looks good to me, but I'm pretty unqualified. I just had one style comment

acceptance/tests/language/sensitive_data_type.rb
er0ck: yeah, been meaning to move away from that travesty of a method.  although i guess i can't complain because its _comment_ says it is private.  :hurtrealbad: that we're duplicating the syslog location info all over the place, but this can be fixed in beaker later

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -75,13 +82,20 @@
...
+                         when /fedora|centos|el|redhat|scientific/ then
+                           '/var/log/messages'
+                         when /ubuntu|debian|cumulus/ then
+                           '/var/log/syslog'
+                         else
+                           nil
+                       end
+          if syslogfile
+            result = agent.exec(Command.new("tail -n 100 #{syslogfile}"),
+                                :acceptable_exit_codes => [0, 1]).stdout.chomp
```
er0ck: why doesn't this just use `on(agent, "tail -n 100 #{syslogfile}").stdout.chomp`
when does this need an exitcode of 1?  empty syslog? syslog < 100 lines? syslogfile doesn't exist?

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -64,6 +64,13 @@
   sitepp_content = generate_manifest(test_resources)
   assertion_code = generate_assertions(test_resources)
 
+  refutation_resources = test_resources.collect { |assertion_group|
+    refutation_group = assertion_group.clone
+    refutation_group[:assertions] = assertion_group[:assertions].select { |assertion|
+      assertion.has_key?(:refute_match) }
+    refutation_group }
```
camlow325: Ah, good call, thanks.  I just modified the commit in place, now f02cf08, replacing the prior curly braces with do/end blocks.

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -64,6 +64,15 @@
   sitepp_content = generate_manifest(test_resources)
   assertion_code = generate_assertions(test_resources)
 
+  refutation_resources = test_resources.collect do |assertion_group|
```
er0ck: actually i'm not really sure why we need to isolate the `refute_match` entries.

thus was the "elegance" in the previous solution.  you could just run the assertions in each state (debug output, syslog, etc) and they would always and should always be the same.  i don't see why that wouldn't be the case here

acceptance/tests/language/sensitive_data_type.rb
```diff
@@ -75,13 +84,20 @@
           run_assertions(assertion_code, result)
         end
 
-        if agent['platform'] =~ /fedora|centos|el|redhat|scientific|ubuntu|debian|cumulus/
```
er0ck: i think there are issues on some network device platforms, so this line might have to be retained.  but i'm not sure.  @LuvCurves do you recall?

acceptance/tests/language/sensitive_data_type.rb
er0ck: agreed with the last part.  i don't think we need to test `--logdest` here.

acceptance/tests/language/sensitive_data_type.rb
camlow325: TBH, I just copied that over from the dump_puppet_log method in beaker.  From experimenting with `tail`, the only time I could get it to return a 1 was when the file did not exist.  If you'd rather just go with the `on...` approach and not allow 1 as an acceptable exit code because the syslog file *should* always exist, I could do that instead.  WDYT?

acceptance/tests/language/sensitive_data_type.rb
er0ck: oh right.  it was because those are the only platforms supported by `dump_puppet_log`.

and indeed the only platforms supported by our replication here.  please retain this line

acceptance/tests/language/sensitive_data_type.rb
er0ck: it probably needs an `else logger.warn "we're not testing logs on #{platform}"` or something, but i wouldn't hold up this PR on that

acceptance/tests/language/sensitive_data_type.rb
er0ck: it probably needs an else logger.warn "we're not testing logs on #{platform}" or something, but i wouldn't hold up this PR on that

camlow325: If the `agent['platform']` isn't captured in the `case` statement below, the assertions will be skipped over completely.  I just took this line out because it seemed redundant to match the platforms to the agent both on this line and in the case statement below.  Does that make sense?

camlow325: Do you want me to have the platforms listed in the `if agent ['platform']...` line and in the case statement then?  Or do you think it would be enough just to put a `logger.warn "we're not testing logs...` message in the else block for the case statement?

acceptance/tests/language/sensitive_data_type.rb
camlow325: If the test is only looking at the `syslog` file content itself - and not the rest of the beaker log - the `assert_match` calls mostly (maybe all?) fail.  For example, the line with 'Sensitive [value redacted]' line will be written to the console debug output for the 'puppet agent' command but is not written to the 'syslog'.  I believe this is because 'puppet agent --debug' by default outputs to the console unless '--logdest syslog' is added to the command line.  If '--logdest syslog' were added to the command line, we'd get the extra lines like 'Sensitive [value redacted]' in the syslog but would not get them on the console, meaning that the console-related assertions in this test would fail.

I don't know of a way to have the Puppet agent "tee" the output both to the syslog and console.  Maybe that would be best for the purpose of this test if possible to do.  WDYT?

acceptance/tests/language/sensitive_data_type.rb
camlow325: Yeah, definitely not great to be duplicating syslog paths here.  Would be much better to have that abstracted across OSes by beaker.

er0ck: [BKR-924](https://tickets.puppetlabs.com/browse/BKR-924)

acceptance/tests/language/sensitive_data_type.rb
camlow325: > thus was the "elegance" in the previous solution. you could just run the assertions in each state (debug output, syslog, etc) and they would always and should always be the same

I think the only reason the syslog ones passed before is that the agent run console debug output was captured in the Beaker logger and the syslog assertions were looking at the Beaker logger - and not strictly the content of the `syslog` file.

acceptance/tests/language/sensitive_data_type.rb
er0ck: yes.  please use `#on()`
Command.new is used internally to beaker and is kinda... wordy for tests where we're supposed to use on helpers for various {{reasons}}

acceptance/tests/language/sensitive_data_type.rb
er0ck: strange.  it worked for me when using the same syslog files.  i'm not sure what you mean by "rest of the beaker log".
but i can see where it wouldn't work and am surprised it works in puppet pipeline. 

this is fine as is

acceptance/tests/language/sensitive_data_type.rb
er0ck: woops. you are right.  sorry, i missed that it'll skip the assertions because of the switch. 

this is fine also

er0ck:
i think this is fine after some commenting and use of `#on` rather than `Command.new`

acceptance/tests/language/sensitive_data_type.rb
camlow325: In b21fc59, I added a warning when the `platform` isn't matched.

acceptance/tests/language/sensitive_data_type.rb
camlow325: I switched this to using `on` in b21fc59.

acceptance/tests/language/sensitive_data_type.rb
camlow325: I added some comments around this in b21fc59.  Hope those are clear enough.

er0ck:
@camlow325 your branch has `master` in it, but this is targeted at LTS-1.7
is that intended?

camlow325:
@er0ck Ah, woops, no, that wasn't intended - good catch!  I based it on the HEAD of LTS-1.7 but named the branch wrong.  I can fix that up and reopen the PR.  Suppose I could squash the commits along with that if that makes sense.  Not sure if you all are still doing that in the Puppet repo.

er0ck:
merged to stable: 02cd1b4a3855a569881b5af91a34825629a86d2b
and master: 20e27225259133c069fb31993f9d0341453ca999

PR #5366, opened 2016-11-08
(PUP-6474) Incorporate gettext_setup gem in to Puppet

Magisus:
This is a continuation of the work @branan started in https://github.com/puppetlabs/puppet/pull/5329. The most recent commit by me contains initial work responding to feedback on that PR.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,6 +31,12 @@ gem "rake", "10.1.1", :require => false
 # Hiera has an unbound dependency on json_pure
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
+# i18n support (gettext-setup and dependencies)
+gem 'text', '~> 1.3.1', :require => false
+gem 'locale', '~> 2.1.2-2', :require => false
+gem 'gettext', '~> 3.2.2', :require => false
+gem 'fast_gettext', '~> 1.1.0', :require => false
+gem 'gettext-setup', '~> 0.6', :require => false
 
 group(:development, :test) do
   gem "rspec", "~> 3.1", :require => false
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ gem "rake", "10.1.1", :require => false
 gem 'json_pure', '~> 1.8', :require => false
 # i18n support (gettext-setup and dependencies)
 gem 'gettext-setup', '~> 0.6', :require => false
+gem 'locale', '~> 2.1', :require => false
 
 group(:development, :test) do
   gem "rspec", "~> 3.1", :require => false
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,7 @@ gem "rake", "10.1.1", :require => false
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
 # i18n support (gettext-setup and dependencies)
-gem 'gettext-setup', '~> 0.6', :require => false
+gem 'gettext-setup', '~> 0.8', :require => false
 gem 'locale', '~> 2.1', :require => false
 
 group(:development, :test) do
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,7 @@ gem "rake", "10.1.1", :require => false
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
 # i18n support (gettext-setup and dependencies)
-gem 'gettext-setup', '~> 0.8', :require => false
+gem 'gettext-setup', '>= 0.10', '< 1.0', :require => false
 gem 'locale', '~> 2.1', :require => false
 
 group(:development, :test) do
--- a/Gemfile
+++ b/Gemfile
@@ -32,10 +32,6 @@ gem "rake", "10.1.1", :require => false
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
 # i18n support (gettext-setup and dependencies)
-gem 'text', '~> 1.3.1', :require => false
-gem 'locale', '~> 2.1.2-2', :require => false
-gem 'gettext', '~> 3.2.2', :require => false
-gem 'fast_gettext', '~> 1.1.0', :require => false
 gem 'gettext-setup', '~> 0.6', :require => false
 
 group(:development, :test) do
--- a/install.rb
+++ b/install.rb
@@ -316,9 +316,9 @@ def prepare_installation
     localedir = InstallOptions.localedir
   else
     if $operatingsystem == "windows"
-      localedir = File.join(Dir::COMMON_APPDATA, "PuppetLabs", "puppet", "share", "puppet", "locale")
+      localedir = File.join(Dir::PROGRAM_FILES, "Puppet Labs", "Puppet", "puppet", "share", "locale")
     else
-      localedir = "/opt/puppetlabs/puppet/share/puppet/locale"
+      localedir = "/opt/puppetlabs/puppet/share/locale"
     end
   end
 
--- a/install.rb
+++ b/install.rb
@@ -205,7 +205,7 @@ def prepare_installation
     opts.on('--bindir[=OPTIONAL]', 'Installation directory for binaries', 'overrides RbConfig::CONFIG["bindir"]') do |bindir|
       InstallOptions.bindir = bindir
     end
-    opts.on('--localedir[=OPTIONAL]', 'Installation directory for locale information', 'Default /opt/puppetlabs/puppet/share/puppet/locale') do |localedir|
+    opts.on('--localedir[=OPTIONAL]', 'Installation directory for locale information', 'Default /opt/puppetlabs/puppet/share/locale') do |localedir|
       InstallOptions.localedir = localedir
     end
     opts.on('--ruby[=OPTIONAL]', 'Ruby interpreter to use with installation', 'overrides ruby used to call install.rb') do |ruby|
--- a/install.rb
+++ b/install.rb
@@ -315,7 +315,11 @@ def prepare_installation
   if not InstallOptions.localedir.nil?
     localedir = InstallOptions.localedir
   else
-    localedir = "/opt/puppetlabs/puppet/share/puppet/locale"
+    if $operatingsystem == "windows"
+      localedir = File.join(Dir::COMMON_APPDATA, "PuppetLabs", "puppet", "share", "puppet", "locale")
+    else
+      localedir = "/opt/puppetlabs/puppet/share/puppet/locale"
+    end
   end
 
   if not InstallOptions.sitelibdir.nil?
--- a/install.rb
+++ b/install.rb
@@ -118,6 +118,17 @@ def do_man(man, strip = 'man/')
   end
 end
 
+def do_locales(locale, strip = 'locales/')
+  locale.each do |lf|
+    next if File.directory? lf
+    olf = File.join(InstallOptions.locale_dir, lf.sub(/^#{strip}/, ''))
+    op = File.dirname(olf)
+    FileUtils.makedirs(op, {:mode => 0755, :verbose => true})
+    FileUtils.chmod(0755, op)
+    FileUtils.install(lf, olf, {:mode => 0644, :preserve => true, :verbose => true})
+  end
+end
+
 # Verify that all of the prereqs are installed
 def check_prereqs
   PREREQS.each { |pre|
@@ -193,6 +204,9 @@ def prepare_installation
     end
     opts.on('--bindir[=OPTIONAL]', 'Installation directory for binaries', 'overrides RbConfig::CONFIG["bindir"]') do |bindir|
       InstallOptions.bindir = bindir
+    end
+    opts.on('--localedir[=OPTIONAL]', 'Installation directory for locale information', 'Default /opt/puppetlabs/puppet/share/puppet/locale') do |localedir|
+      InstallOptions.localedir = localedir
     end
     opts.on('--ruby[=OPTIONAL]', 'Ruby interpreter to use with installation', 'overrides ruby used to call install.rb') do |ruby|
       InstallOptions.ruby = ruby
@@ -298,6 +312,12 @@ def prepare_installation
     bindir = RbConfig::CONFIG['bindir']
   end
 
+  if not InstallOptions.localedir.nil?
+    localedir = InstallOptions.localedir
+  else
+    localedir = "/opt/puppetlabs/puppet/share/puppet/locale"
+  end
+
   if not InstallOptions.sitelibdir.nil?
     sitelibdir = InstallOptions.sitelibdir
   else
@@ -335,6 +355,7 @@ def prepare_installation
   rundir = join(destdir, rundir)
   logdir = join(destdir, logdir)
   bindir = join(destdir, bindir)
+  localedir = join(destdir, localedir)
   mandir = join(destdir, mandir)
   sitelibdir = join(destdir, sitelibdir)
 
@@ -346,6 +367,7 @@ def prepare_installation
   FileUtils.makedirs(vardir)
   FileUtils.makedirs(rundir)
   FileUtils.makedirs(logdir)
+  FileUtils.makedirs(localedir)
 
   InstallOptions.site_dir = sitelibdir
   InstallOptions.codedir = codedir
@@ -356,6 +378,7 @@ def prepare_installation
   InstallOptions.var_dir = vardir
   InstallOptions.run_dir = rundir
   InstallOptions.log_dir = logdir
+  InstallOptions.locale_dir = localedir
 end
 
 ##
@@ -468,6 +491,7 @@ FileUtils.cd File.dirname(__FILE__) do
   ri    = glob(%w{bin/*.rb lib/**/*.rb}).reject { |e| e=~ /\.(bat|cmd)$/ }
   man   = glob(%w{man/man[0-9]/*})
   libs  = glob(%w{lib/**/*})
+  locales = glob(%w{locales/**/*})
 
   prepare_installation
 
@@ -481,5 +505,6 @@ FileUtils.cd File.dirname(__FILE__) do
   do_bins(bins, InstallOptions.bin_dir)
   do_bins(windows_bins, InstallOptions.bin_dir, 'ext/windows/') if $operatingsystem == "windows" && InstallOptions.batch_files
   do_libs(libs)
+  do_locales(locales)
   do_man(man) unless $operatingsystem == "windows"
 end
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -18,6 +18,7 @@ require 'puppet/util/run_mode'
 require 'puppet/external/pson/common'
 require 'puppet/external/pson/version'
 require 'puppet/external/pson/pure'
+require 'gettext-setup'
 
 #------------------------------------------------------------
 # the top-level module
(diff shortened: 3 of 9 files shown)


lib/puppet.rb
```diff
@@ -19,19 +19,7 @@
...
-begin
-  require 'gettext-setup'
-  GETTEXT_AVAILABLE=true
-rescue LoadError
-  def _(msg)
-    msg
-  end
-  GETTEXT_AVAILABLE=false
-end
+require 'gettext-setup'
```
glennsarti: You'll need to add a require for the puppet::util::locale class

lib/puppet.rb
```diff
@@ -19,6 +19,8 @@
 require 'puppet/external/pson/version'
 require 'puppet/external/pson/pure'
```
MosesMendoza: (nitpick) unnecessary extra whitespace

install.rb
```diff
@@ -294,6 +308,16 @@ def prepare_installation
     bindir = RbConfig::CONFIG['bindir']
   end
 
+  if not InstallOptions.localedir.nil?
+    localedir = InstallOptions.localedir
+  else
+    if $operatingsystem == "windows"
+      localedir = File.join(Dir::COMMON_APPDATA, "PuppetLabs", "puppet", "share", "puppet", "locale")
```
Magisus: If we want to put this data in `Program Files` instead of `ProgramData`, what is the best way to tell if we should be using `(x86)` or not?

Gemfile
```diff
@@ -30,6 +30,9 @@ gem "rake", "10.1.1", :require => false
 # Hiera has an unbound dependency on json_pure
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
+# i18n support (gettext-setup and dependencies)
+gem 'gettext-setup', '~> 0.6', :require => false
+gem 'locale', :require => false
```
glennsarti: IMO we should at least restrict to current X version i.e. `gem 'locale', '~> 2.1', :require => false`

install.rb
glennsarti: This one is tricky.  Strictly speaking we should not be putting directories into `Program Files` post MSI installation.  MSI considers these to be "unmanaged" (because it didn't create them) and will not remove them on MSI uninstall. @Iristyle thoughts?

locales/puppet.pot
```diff
@@ -0,0 +1,124 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) 2016 Puppet, Inc.
+# This file is distributed under the same license as the Puppet automation framework package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2016.
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: Puppet automation framework 4.6.2-137-g151642b\n"
```
kurtwall: Does this change?

Magisus: Since that looks like the result of a `git describe` I'm assuming it does. I'll regenerate this file and see if it updates 4.8.0-something.

Side note: we may end up not even checking this file into the repo. Since the important part needed by Transifex is the PO associated with each target language, and our goal is generate those files as part of a CI job, tracking this file would probably just cause unnecessary churn.

Magisus: Yes indeed, it does change. Appears to be based on `git describe`.

locales/puppet.pot
```diff
@@ -0,0 +1,128 @@
...
+#. Now we find the line with our summary, extract it, and return it.  This
+#. depends on the implementation coincidence of how our pages are
+#. formatted.  If we can't match the pattern we expect we return the empty
+#. string to ensure we don't blow up in the summary. --daniel 2011-04-11
+msgid "! Subcommand unavailable due to error. Check error logs."
+msgstr ""
+
+#. @api private
+msgid "Any"
+msgstr ""
```
Magisus: Not really sure where all of these came from...

lib/puppet.rb
```diff
@@ -36,6 +38,27 @@ module Puppet
...
+    elsif File.exist?(posix_system_locale_path)
+      locale_path = posix_system_locale_path
+    else
+      # We couldn't load our locale data. Possibly we're loaded as a rubygem or something?
+      locale_path = nil
+    end
+
+    if locale_path
+      GettextSetup.initialize(locale_path)
+      FastGettext.locale = GettextSetup.negotiate_locale(Locale.current.language)
```
glennsarti: Evaluates `en_AU` and `en_US` as `en` on Windows.  I assume this is expected.

Magisus: This is expected, and it does the same thing on all platforms. If we want the more specific info, there are other ways to interact with the API that are a bit more complicated but preserve the full name.

install.rb
Iristyle: So I think this should probably go under `Program Files` as static data, rather than living under user data directories, because I would expect the MSI to have all the translations baked in at install-time.

`install.rb` is supposed to roughly mimic how an MSI is installed, so this doesn't quite add up for me... but I might be misunderstanding how this all works?

I also would expect to see a PR to https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md that describes where translations end up.

If you want to add a directory under what is something like `Program Files`, I think you want to mod under https://github.com/Magisus/puppet/blob/848717c34d6accd1f4c2fd3c3c04950db27fc4fe/install.rb#L486 

We should probably chat about this to make sure we're on the same page - I'm coming into this pretty late, and not sure I grok the process yet.

Gemfile
```diff
@@ -30,6 +30,9 @@ gem "rake", "10.1.1", :require => false
 # Hiera has an unbound dependency on json_pure
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
+# i18n support (gettext-setup and dependencies)
+gem 'gettext-setup', :git => 'git@github.com:Magisus/gettext-setup-gem.git', :branch => 'mo-refactor'
```
glennsarti: Adding this for not approved so that this PR does not merged in by accident

install.rb
```diff
@@ -477,5 +505,7 @@ def install_binfile(from, op_file, target)
   do_bins(bins, InstallOptions.bin_dir)
   do_bins(windows_bins, InstallOptions.bin_dir, 'ext/windows/') if $operatingsystem == "windows" && InstallOptions.batch_files
   do_libs(libs)
+  # TODO: (PUP-6935) once we have determined the correct behavior on Windows, we might want to not exclude it
```
glennsarti: Adding this for not approved so that this PR does not merged in by accident

lib/puppet.rb
```diff
@@ -36,6 +38,28 @@ module Puppet
   require 'puppet/environments'
 
   class << self
+    # TODO: Handle the rubygems case?
```
glennsarti: Adding this for not approved so that this PR does not merged in by accident

locales/en/puppet.po
```diff
@@ -0,0 +1,128 @@
...
+msgid ""
+msgstr ""
+"Project-Id-Version: Puppet automation framework 4.8.0-51-gefaf423\n"
+"\n"
+"Report-Msgid-Bugs-To: https://tickets.puppetlabs.com\n"
+"POT-Creation-Date: 2016-11-14 15:35-0800\n"
+"PO-Revision-Date: 2016-11-14 15:35-0800\n"
+"Last-Translator: Automatically generated\n"
+"Language-Team: none\n"
+"Language: jp\n"
```
glennsarti: Unless I'm reading this wrong...Is this a Japanese language file, not EN?

Magisus: It is, but I didn't want to bother changing the locale on the system, so I just renamed the folder. This will be deleted before this PR gets merged.

glennsarti: Ahhhh ok.  Thanks

lib/puppet.rb
Magisus: Do we have enough information to handle this as part of this PR? @branan? What is the correct way to handle the Rubygems case?

Magisus:
Note, this is blocked on a release of the gettext-setup gem that contains https://github.com/puppetlabs/gettext-setup-gem/pull/14.

lib/puppet.rb
branan: We need to figure out how to package and ship the locale files as part of a gem. I have no idea how we'd go about doing that, TBH. I think we need to end up copying the locale files into lib/ during the gem packaging step - this is possibly something we can expand puppetlabs/packaging to handle (since it builds gems).

I don't think it's worth blocking this PR on figuring all this out, but we should file a ticket for it (probably blocking the 4.9 release) and update the TODO with the ticket numner

Magisus:
I have verified that with an update to the gettext-setup gem and [this puppet-agent PR](https://github.com/puppetlabs/puppet-agent/pull/910), the changes made in this set of commits allows us to generate a POT file, then from it PO files, then use `msgfmt` to compile those to MO files. If one then builds a puppet-agent with the change in that PR, it will install the files to the directories specified [here](https://github.com/puppetlabs/puppet-specifications/pull/91) and load the translated string from the MO files in those directories at run-time. The MO file appears to be cross-platform: I created it on my machine, and successfully loaded it on both Windows and Linux (Redhat 7).

Magisus:
A new version of the gettext-setup gem has been released, and the version has been bumped in puppetserver, so I believe all blockers for this PR have been cleared.

MosesMendoza:
reviewing

joshcooper:
Having recently dealt with the `semantic_puppet` unvendoring and adding a gem dependency, do we need to update https://github.com/puppetlabs/puppetserver/blob/60d95674376fc2240a527e2e8376d11bd30ff325/resources/ext/build-scripts/gem-list.txt? Or will puppetserver need to do that when they bump their puppet dependency? Are we going to get into situations where puppet-agent is updated on the server and expects a newer version of `gettext-setup` but it's an older puppetserver version that has the older gem version?

Magisus:
@joshcooper I already updated that [here](https://github.com/puppetlabs/puppetserver/pull/1297/files). PR got merged this morning.

MosesMendoza:
@Magisus you wouldn't happen to have run an ad-hoc acceptance run with these changes?

lib/puppet.rb
```diff
@@ -36,6 +38,27 @@ module Puppet
   require 'puppet/environments'
 
   class << self
+    # TODO: (PUP-6958) Handle the rubygems case
+    local_locale_path = File.absolute_path('../locales', File.dirname(__FILE__))
```
MosesMendoza: I'm inclined to suggest we should replace `File.absolute_path` with `Puppet::FileSystem.expand_path` here but I'm curious - was there a reason `File.absolute_path` was selected?

lib/puppet.rb
```diff
@@ -36,6 +38,27 @@ module Puppet
...
+    posix_system_locale_path = File.absolute_path('../../../share/locale', File.dirname(__FILE__))
+    win32_system_locale_path = File.absolute_path('../../../../../puppet/share/locale', File.dirname(__FILE__))
```
MosesMendoza: Not a merge blocker or anything but it would be nice (IMO) to have comments here with examples of these expanded out in the default case

lib/puppet.rb
Magisus: Not sure, @branan ?

lib/puppet.rb
Magisus: Agreed, I'll add them.

Magisus:
@MosesMendoza I don't think I've ever let it run all the way through acceptance, no. I've built it and manually tested it four or five times though. I'll start an acceptance run now.

locales/puppet.pot
```diff
@@ -0,0 +1,128 @@
+# SOME DESCRIPTIVE TITLE.
```
MosesMendoza: It would be nice to figure out if we can set this

locales/puppet.pot
```diff
@@ -0,0 +1,128 @@
...
+# Copyright (C) 2016 Puppet, Inc.
+# This file is distributed under the same license as the Puppet automation framework package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2016.
```
MosesMendoza: and update this to info@puppet.com

locales/puppet.pot
Magisus: I think these might be coming from the underlying system tools. gettext-setup passes a number of arguments to `rxgettext` [here](https://github.com/puppetlabs/gettext-setup-gem/blob/master/lib/tasks/gettext.rake#L36-L41), which seems to cover most of the [available options](https://gist.github.com/Magisus/2a26519e72a4c84df9be6a5e9aaa51d5). There are also [these docs](https://www.gnu.org/software/gettext/manual/html_node/Header-Entry.html#Header-Entry) which makes it sound like it has to be done manually...

MosesMendoza: Indeed. That's a bummer

locales/puppet.pot
hlindberg: Yes. And these should not be translated.

MosesMendoza: Thanks @hlindberg. Off the top of my head, it seems like we could solve this problem (at least) a couple ways..

One would be to augment `gettext-setup-gem` to support excluding files from translation. Currently it only accepts a `source_files` config, the value to which is passed to `Dir.glob` to get all the files to translate. We could add a new config option like `exclude_files` which we would subtract from the `Dir.glob`. Thoughts on that possibility @smcclellan?

Another option would be to modify these method definitions so they don't trip the gettext parser. `_` is a special ruby name that allows us to declare, but not make use of, a method parameter but it's also significant for gettext. As of ruby 2.0 I believe you can also use `_` as a prefix to unused variable names. So we could redefine 

```ruby
  # @api private
  def string_PAnyType(_)    ; @bld << 'Any'     ; end
```

as
```ruby
  # @api private
  def string_PAnyType(_v)    ; @bld << 'Any'     ; end
```

Doing so removes these entries from the .pot file.

hlindberg: It seems odd that a `_x` would not bind a new variable. Isn't the problem because of a bug in the gettext parser? There is no way the syntax here could be construed as being something to do with strings and translation.

MosesMendoza: @hlindberg oops I thought both `_` and `_x` bind new variables within their local scope, ie in this case the method `string_PAnyType`. The significance of using the `_` in either case is that ruby does not complain about an unused variable. Maybe I'm mistaken there?

I agree it seems likely the underlying issue here is the (ruby-)gettext parser itself. I'll file an issue with our finding in that project, but I think either of the workarounds proposed above seem reasonable too, in lieu of timing.

MosesMendoza: Filed https://github.com/ruby-gettext/gettext/issues/49

Magisus:
Green run of puppet acceptance tests using this branch: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-puppet-ad-hoc/110/

locales/puppet.pot
hlindberg: The difference when using `_` and a variable name is that a) Ruby accepts an argument without complaining and b) there is no variable assignment since `_` does not have a value. So a) must have a var or code fails b) faster & less memory used and clearly signals variable is not in use.

MosesMendoza: naming your method parameter `_` doesn't make the parameter optional, though when calling that method, correct?

Thus, 
```ruby
def foo(_)
# _ can go unused in this context, but is still an accessible variable bound to whatever was supplied to 
end

foo # ArgumentError: wrong number of arguments (given 0, expected 1)
foo(:bar) # success
```

I don't disagree that `_` signals variable not in use

Magisus: @MosesMendoza any update to gettext-setup would require another release of the gem. Do you think it's reasonable to extend the range of gettext-setup versions to include both 0.8 (current, used at runtime to pull in translations) and 0.9 (which would contain this update, used when performing string extraction in the separate CI job)? If not, we will most likely not be able to ship this functionality in Puppet 4.9 (PE 2017.1), since it would require an update of puppetserver, which is not expected to do another release till after 4.9 ships.

MosesMendoza: As long as the new gettext-setup version just contains support for excluding files I think it should be fine to pull it into puppet and not puppet server (in this case). This is because the changed functionality would only be used during POT file generation (ie from CI) and not by puppet server.

lib/puppet/face/help.rb
```diff
@@ -66,7 +66,7 @@
       facename, actionname = args
       if legacy_applications.include? facename then
         if actionname then
-          raise ArgumentError, "Legacy subcommands don't take actions"
+          raise ArgumentError, _("Legacy subcommands don't take actions")
```
kurtwall: IIRC, contractions ("don't") pose challenges for i18n insofar as non-native speakers often lack the language context to properly expand the contraction. Maybe use "do not" in this case?

Magisus: That may be true, but I think we want to avoid changing the content of existing strings as much as possible. Too hard to tell who might be relying on what (including us in our tests). In this case it probably doesn't matter, but I think I'll still leave it as-is till I hear otherwise from the people in touch with the translators.

install.rb
```diff
@@ -194,6 +205,9 @@ def prepare_installation
     opts.on('--bindir[=OPTIONAL]', 'Installation directory for binaries', 'overrides RbConfig::CONFIG["bindir"]') do |bindir|
       InstallOptions.bindir = bindir
     end
+    opts.on('--localedir[=OPTIONAL]', 'Installation directory for locale information', 'Default /opt/puppetlabs/puppet/share/puppet/locale') do |localedir|
```
geoffnichols: Comparing to `localedir` on line 321, is there an extra 'puppet' in the `Default` path here?

lib/puppet.rb
```diff
@@ -18,6 +18,8 @@
 require 'puppet/external/pson/common'
 require 'puppet/external/pson/version'
 require 'puppet/external/pson/pure'
+require 'gettext-setup'
+require 'locale'
```
MosesMendoza: I think when running from gems we get `locale` installed as a transitive dependency  - `gettext-setup` => `gettext` => `locale`. Since we require this here, thoughts on expressing a dependency on `locale` in `ext/project_data.yaml` too?

Gemfile
```diff
@@ -31,6 +31,9 @@ gem "rake", "10.1.1", :require => false
 # Hiera has an unbound dependency on json_pure
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
+# i18n support (gettext-setup and dependencies)
+gem 'gettext-setup', '>= 0.8', '< 1.0', :require => false
```
MosesMendoza: Since `0.10` contains support for the `exclude_files` configuration option we added, I think we should update dependencies to reflect `0.10` minimum.

lib/puppet.rb
MosesMendoza: They're used by our [packaging](https://github.com/puppetlabs/packaging/blob/master/tasks/gem.rake#L65-L73) when generating artifacts. In the case of gems, when we package up a gem of puppet, the packaging reads this yaml file to determine what dependencies that gem should express. Our Gemfile also reads this file in to gather platform specific dependencies: https://github.com/puppetlabs/puppet/blob/master/Gemfile#L78-L99.

It's a bit overly complicated/duplicitous I think, but you know.. tech debt.

Gemfile
Magisus: Will that cause issues with the fact that the server only has 0.8?

MosesMendoza: I don't think so.. The only issue I can see if we do some sort of version equality detection across the gem paths of the two projects but I don't think we do anything like that- it's just assumed we have compatible versions in both where needed. There isn't anything that changed between 0.8 and 0.10 that would be used in the context of Puppet server having loaded puppet's code, so dependency mismatch aside it should load/operate fine.

Magisus: Sounds good.

MosesMendoza:
👍

PR #5367, opened 2016-11-08
(PUP-6856) Catch :no_such_key during interpolation sub_lookup

thallgren:
An interpolation that uses dotted keys that in turn results in a
`:no_such_key` being thrown (because the key is missing), would not catch the
throw and hence, yield strange results. This commit ensures that the thrown
symbol `:no_such_key` is caught so that the call instead results in an empty
string.

### DIFF
--- a/spec/unit/pops/lookup/interpolation_spec.rb
+++ b/spec/unit/pops/lookup/interpolation_spec.rb
@@ -161,6 +161,10 @@ describe 'Puppet::Pops::Lookup::Interpolation' do
       expect(interpolator.interpolate('a dot e: %{a.d}', lookup_invocation, true)).to eq('a dot e: (scope) a dot d is a hash entry')
     end
 
+    it 'should report a key missing and replace with empty string when a dotted key is used to navigate into a structure and then not found' do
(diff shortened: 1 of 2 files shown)


spec/unit/pops/lookup/interpolation_spec.rb
```diff
@@ -161,6 +161,10 @@ def expect_lookup(*keys)
       expect(interpolator.interpolate('a dot e: %{a.d}', lookup_invocation, true)).to eq('a dot e: (scope) a dot d is a hash entry')
     end
 
+    it 'should report a key missing and replace with empty string when a dotted key is used to navigate into a structure and then not found' do
+      expect(interpolator.interpolate('a dot n: %{a.n}', lookup_invocation, true)).to eq('a dot n: ')
+    end
```
thallgren: It tests if the resulting string contains nothing (the empty string) at the place where the interpolation took place. The title could perhaps be more explicit. The sub-lookup "reports the key missing" and the interpolation reacts to that by inserting the empty string. Then again, it's all under the category "when using dotted keys".

hlindberg: alright... kind of hard to parse and understand, but I get it now that I read it again. (merging).

PR #5368, opened 2016-11-08
(PUP-6834) Ensure that input to MergeStrategy deep does not mutate

thallgren:
Before this commit, the primary source for a deep merge was dup'ed to avoid
that the deep merge would mutate it (it mutates the primary source). A dup
is however not enough when the source is a nested hash. The dup will only
perform a shallow copy. This commit ensures that the source is deep copied
instead.

### DIFF
--- a/lib/puppet/pops/merge_strategy.rb
+++ b/lib/puppet/pops/merge_strategy.rb
@@ -309,8 +309,20 @@ module Puppet::Pops
     def checked_merge(e1, e2)
       dm_options = { :preserve_unmergeables => false }
       options.each_pair { |k,v| dm_options[k.to_sym] = v unless k == 'strategy' }
-      # e2 (the destination) is dup'ed to avoid that the passed in object mutates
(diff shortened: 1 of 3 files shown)


hlindberg:
failed tests

lib/puppet/pops/merge_strategy.rb
```diff
@@ -309,8 +309,20 @@ def self.key
...
+    def deep_clone(value)
+      if value.is_a?(Hash)
+        result = value.clone
+        value.each{ |k, v| result[k] = deep_clone(v) }
+        result
+      elsif value.is_a?(Array)
+        value.map{ |v| deep_clone(v) }
+      else
+        value
+      end
```
thallgren: Marshah/Unmarshal is more efficient when the structure consists of arbitrary objects and introspection is needed to access the attributes. It's slower when the structure is known to be Data.

The key can something other than a String. It's however never anything that isn't scalar so there's no need to deep copy the key.

PR #5369, opened 2016-11-08
(maint) Fix some minor pcore type related issues

thallgren:
This maint commit fixes some minor issues that was found while browsing
through the code.


hlindberg:
The changes are fine, but I think this deserves a ticket "improve error when type alias cannot be resolved", then we also get a release note.

thallgren:
Only problem is, even if it's a code glitch, I haven't found a good way to provoke it (I don't say it can't happen but it would be some obscure combinations of several intertwined self-recursive aliases), so I'm fairly certain that no one has seen it (or will see it). I don't think a ticket/release note would give us anything. Don't know what to put there actually.

PR #5370, opened 2016-11-08
(PUP-6188) Puppet::FileSystem and explicit Encoding changes

Iristyle:
Supercedes #5354 

Channeling @MosesMendoza from prior PR:

This PR builds on / supersedes #4898 from @Iristyle 

[PUP-6188](tickets.puppetlabs.com/browse/PUP-6188) tracks an effort to ensure Puppet's various file system interactions are specifying utf-8 encoding where possible/applicable, possibly in `Puppet::FileSystem.open` and `Puppet::FileSystem.exclusive_open`, but notes the problematic situation resulting from the reordering of method arguments from `Puppet::FileSystem::FileImpl#open` to [`::File.open`](https://ruby-doc.org/core-2.3.1/File.html#method-c-open):

> - Rubys `perm` in our call is renamed as `mode`, and is moved from position 2 to position 1
> - Rubys `mode` in our call is renamed as `options`, and is moved from position 1 to position 2
> - We don't support the passing in of Rubys `opt` which is where one would set ``:encoding => 'utf-8'``

This is further complicated by the fact that ruby allows specifying mode, encoding, and other options a number of different ways via these parameters. For example, encoding and mode, which can be specified
- via `opt` (ruby optional param 3) in an options hash, ie `:mode => 'r', :encoding => 'BOM|UTF-8'` 
- via a `mode` string (ruby param 2), containing one of:
  - an IO Open Mode: `"r"`, `"r+"`, `"w"`
  - an IO Open Mode _with encoding_: `"r:IBM437"`, `"w:BOM|UTF-8"`
  - a list of `|` separated File Constants: `File::CREATE|File::APPEND`

This means it's not simply a matter of extending the definition of `Puppet::FileSystem::FileImpl#open` and `Puppet::FileSystem::FileImpl#exclusive_open` to support an options hash (and thus specifying `:encoding => 'UTF-8'` because such options may collide with those already specified by a user via the `mode` string (ruby param 2). We could probably handle this somewhat gracefully (a rudimentary first pass at this is in https://github.com/MosesMendoza/puppet/commit/ccb65927b22544239adbc49a7b054cc91016bdde) but it seems like less risk of backwards-incompatibility and less complexity overall to simply update the callers of `Puppet::FileSystem` inside of puppet to leverage the ability to pass encoding in via the `mode` string.

This PR does so in several places, explicitly specifying UTF-8 as part of the mode string and updating associated tests. When updating the `puppet config` face, it was discovered that the `:set` action has no unit testing associated with it (at least not that I can find) - all testing is at the lower level of `Puppet::Settings::Inifile`, so this PR also includes a commit that adds initial basic spec coverage of `puppet config set`. Additionally, where encountered, `File.read` was replaced with `Puppet::FileSystem.read` - it seemed as if some of these places existed before `Puppet::FileSystem` was implemented.

After discussion with @Iristyle it was determined that much of the test coverage here likely deserves substantial improvement. Specifically, creating new tests that actually create files containing unicode-only characters, passing them through the code under test, and validating the fidelity of the data actually remains intact. I'm not sure if this PR should be gated on the addition of such tests, but I don't disagree that they're justified.

### DIFF
--- a/lib/puppet/file_system.rb
+++ b/lib/puppet/file_system.rb
@@ -32,6 +32,15 @@ module Puppet::FileSystem
 
   # Opens the given path with given mode, and options and optionally yields it to the given block.
   #
+  # @param path [String, Pathname] the path to the file to operate on
+  # @param mode [Integer] The mode to apply to the file if it is created
+  # @param options [String] Extra file operation mode information to use
+  #   This is the standard mechanism Ruby uses in the IO class, and therefore
+  #   encoding may be specified explicitly as fmode : encoding or fmode : "BOM|UTF-*"
+  #   for example, a:ASCII or w+:UTF-8
+  # @yield The file handle, in the mode given by options, else read-write mode
+  # @return [Void]
+  #
   # @api public
   #
   def self.open(path, mode, options, &block)
@@ -98,10 +107,13 @@ module Puppet::FileSystem
   #
   # @param path [Pathname] the path to the file to operate on
   # @param mode [Integer] The mode to apply to the file if it is created
-  # @param options [Integer] Extra file operation mode information to use
-  #   (defaults to read-only mode)
+  # @param options [String] Extra file operation mode information to use
+  #   (defaults to read-only mode 'r')
+  #   This is the standard mechanism Ruby uses in the IO class, and therefore
+  #   encoding may be specified explicitly as fmode : encoding or fmode : "BOM|UTF-*"
+  #   for example, a:ASCII or w+:UTF-8
   # @param timeout [Integer] Number of seconds to wait for the lock (defaults to 300)
-  # @yield The file handle, in read-write mode
+  # @yield The file handle, in the mode given by options, else read-write mode
   # @return [Void]
   # @raise [Timeout::Error] If the timeout is exceeded while waiting to acquire the lock
   #
--- a/lib/puppet/indirector/file_bucket_file/file.rb
+++ b/lib/puppet/indirector/file_bucket_file/file.rb
@@ -115,7 +115,11 @@ module Puppet::FileBucketFile
     # @param files_original_path [String]
     #
     def matches(paths_file, files_original_path)
-      Puppet::FileSystem.open(paths_file, 0640, 'a+') do |f|
+      # Puppet will have already written the paths_file in the systems encoding
+      # given its possible that request.options[:bucket_path] or Puppet[:bucketdir]
+      # contained characters in an encoding that are not represented the
+      # same way when the bytes are decoded as UTF-8, continue using system encoding
+      Puppet::FileSystem.open(paths_file, 0640, 'a+:external') do |f|
         path_match(f, files_original_path)
       end
     end
@@ -138,7 +142,11 @@ module Puppet::FileBucketFile
           Puppet::FileSystem.dir_mkpath(paths_file)
         end
 
-        Puppet::FileSystem.exclusive_open(paths_file, 0640, 'a+') do |f|
+        # Puppet will have already written the paths_file in the systems encoding
+        # given its possible that request.options[:bucket_path] or Puppet[:bucketdir]
+        # contained characters in an encoding that are not represented the
+        # same way when the bytes are decoded as UTF-8, continue using system encoding
+        Puppet::FileSystem.exclusive_open(paths_file, 0640, 'a+:external') do |f|
           if Puppet::FileSystem.exist?(contents_file)
             verify_identical_file!(contents_file, bucket_file)
             Puppet::FileSystem.touch(contents_file)
--- a/lib/puppet/settings/file_setting.rb
+++ b/lib/puppet/settings/file_setting.rb
@@ -185,6 +185,11 @@ class Puppet::Settings::FileSetting < Puppet::Settings::StringSetting
   end
 
   # @api private
+  # @param option [String] Extra file operation mode information to use
+  #   (defaults to read-only mode 'r')
+  #   This is the standard mechanism Ruby uses in the IO class, and therefore
+  #   encoding may be explicitly like fmode : encoding or fmode : "BOM|UTF-*"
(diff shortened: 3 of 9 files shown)


spec/integration/ssl/autosign_spec.rb
```diff
@@ -12,10 +14,16 @@
         '1.3.6.1.4.1.34380.2.2' => # system IPs in hex
           [ 0xC0A80001, # 192.168.0.1
             0xC0A80101 ], # 192.168.1.1
+        # different UTF-8 widths
+        # 1-byte A
+        # 2-byte ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
+        # 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
+        # 4-byte 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
+        '1.2.840.113549.1.9.7' => utf8_value,
```
MosesMendoza: I think by including `1.2.840.113549.1.9.7`(Challenge Password) in `custom_attributes` we've invalidated the test `does not copy custom attributes to certificate` (or perhaps shown it wasn't valid to begin with) as that test seems to just be checking the cert has no attributes under `1.3.6.1.4.1.34380.2`.

Iristyle: Ah yeah, looked at that one, but missed what I *should* have changed there.

Should also add another line asserting that `1.2.840.113549.1.9.7` is not present in the certificate as well.

Iristyle:
I've moved all the bits outside of the original scope / spirit of PUP-6188 to a new ticket PUP-6925 and opened up a PR at #5384 to continue that effort

lib/puppet/file_system.rb
```diff
@@ -32,6 +32,15 @@ def self.overlay(*files, &block)
 
   # Opens the given path with given mode, and options and optionally yields it to the given block.
   #
+  # @param path [String, Pathname] the path to the file to operate on
+  # @param mode [Integer] The mode to apply to the file if it is created
+  # @param options [String] Extra file operation mode information to use
+  #   This is the standard mechanism Ruby uses in the IO class, and therefore
+  #   encoding may be explicitly like fmode : encoding or fmode : "BOM|UTF-*"
```
MosesMendoza: Was this intended to read something like `...may be _supplied_ explicitly like...` ?

lib/puppet/indirector/file_bucket_file/file.rb
```diff
@@ -115,7 +115,11 @@ def validate_key(request)
     # @param files_original_path [String]
     #
     def matches(paths_file, files_original_path)
-      Puppet::FileSystem.open(paths_file, 0640, 'a+') do |f|
+      # Puppet will have already written the paths_file in the systems encoding
+      # given its possible that request.options[:bucket_path] or Puppet[:bucketdir]
+      # contained characters in an encoding that are not represented the
+      # same way when the bytes are decoded as UTF-8, continue using system encoding
+      Puppet::FileSystem.open(paths_file, 0640, "a+:#{Encoding.default_external.name}") do |f|
```
MosesMendoza: Nice.

Do you think there's any chance there's an encoding for which the way it's specified as part of an IO Open Mode String differs from the string returned by `Encoding.<encoding>.name` ?

MosesMendoza: Also I can't imagine `Encoding.default_external` changing again in ruby any time soon but if we were ever to adopt such a ruby we may want to remember to warn users about possible side effects in areas like this (file bucket).

lib/puppet/settings/file_setting.rb
```diff
@@ -185,13 +185,24 @@ def validate(value)
...
+  #   encoding may be explicitly like fmode : encoding or fmode : "BOM|UTF-*"
+  #   for example, a:ASCII or w+:UTF-8
   def exclusive_open(option = 'r', &block)
     controlled_access do |mode|
       Puppet::FileSystem.exclusive_open(file(), mode, option, &block)
     end
   end
 
   # @api private
+  # TODO: need to find callers of this guy
```
MosesMendoza: Was this `TODO` intentionally left in?

spec/unit/file_system_spec.rb
```diff
@@ -17,6 +24,70 @@ def with_file_content(content)
     file.close
   end
 
+  context "#open" do
```
MosesMendoza: I'm surprised there are no tests around this already.. thanks for adding these

spec/unit/file_system_spec.rb
```diff
@@ -17,6 +24,70 @@ def with_file_content(content)
...
+    end
+
+    it "cannot accept a mode string" do
+      file = tmpfile('file_to_update')
+      expect {
+        Puppet::FileSystem.open(file, "444", 'a') { |fh| fh.write('') }
+      }.to raise_error(TypeError)
+    end
+
+    it "opens, creates ands allows updating of a new file, using by default, the external system encoding" do
```
MosesMendoza: 👍

lib/puppet/file_system.rb
Iristyle: Thanks

lib/puppet/settings/file_setting.rb
Iristyle: No, thanks!

spec/unit/file_system_spec.rb
Iristyle: 👍

Iristyle:
All ready now I think @MosesMendoza

MosesMendoza:
👍

MosesMendoza:
checking if CI is green before merging

PR #5371, opened 2016-11-10
(PUP-6911) Do uptime comparisons as integers in var-refresh test

camlow325:
Previously, the `variables_refreshed_each_compilation` acceptance test
could fail if the "earlier" uptime value that the test was comparing for
an agent run was lexicographically greater than the "later" uptime value.
For example, an earlier uptime of "9999" would be considered greater
than a later uptime of "10002" and, therefore, fail.

This commit changes the test to convert the uptimes to Integers before
comparing them so that the comparisons are done numerically rather than
as strings.


camlow325:
Ping @er0ck for review since I think you may have written this test originally.  This one just came up as a transient fail for us in the puppetserver pipeline overnight but isn't really a blocker for us since it goes away on a re-run.

Ping @cprice404 and @jpinsonault for a second dev review whenever.

jpinsonault:
👍  looks good to me

er0ck:
merged to stable at: 0d1c8a3 and on to master at: aeaf740

PR #5374, opened 2016-11-11
(docs) Minor HTTP API doc fixes

nfagerlund:
* Update puppetlabs.com hostnames.
* Add the pluginfacts mount point, which works the same as the plugins one.

### DIFF
--- a/api/docs/http_file_content.md
+++ b/api/docs/http_file_content.md
@@ -10,9 +10,12 @@ Get a file.
 
     GET /puppet/v3/file_content/:mount_point/:name
 
-`:mount_point` is one of mounts configured in the `fileserver.conf`.
-See [the puppet file server guide](https://docs.puppetlabs.com/guides/file_serving.html)
-for more information about how mount points work.
+The endpoint path includes a `:mount_point` which can be one of the following types:
+
+* Custom file serving mounts as specified in fileserver.conf --- see [the docs on configuring mount points](https://docs.puppet.com/puppet/latest/reference/file_serving.html).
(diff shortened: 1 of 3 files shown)


api/docs/http_file_content.md
```diff
@@ -10,9 +10,12 @@ Get a file.
...
 
-`:mount_point` is one of mounts configured in the `fileserver.conf`.
-See [the puppet file server guide](https://docs.puppetlabs.com/guides/file_serving.html)
-for more information about how mount points work.
+The endpoint path includes a `:mount_point` which can be one of the following types:
+
+* Custom file serving mounts as specified in fileserver.conf --- see [the docs on configuring mount points](https://docs.puppet.com/puppet/latest/reference/file_serving.html).
+* `modules/<MODULE>` --- a semi-magical mount point which allows access to the `files` subdirectory of `<MODULE>` --- see [the docs on file serving](https://docs.puppet.com/puppet/latest/reference/file_serving.html).
+* `plugins` --- a highly magical mount point which merges many directories together. Used for plugin sync; not intended for general consumption. Per-module sub-paths can not be specified.
+* `pluginfacts` --- a highly magical mount point which merges many directories together. Used for plugin sync; not intended for general consumption. Per-module sub-paths can not be specified.
```
joshcooper: But since it's not intended for general consumption, I think it's ok to not do this too.

nfagerlund: I debated it and didn't come to a firm conclusion! I went that way in the docs about mount points, tho. 

I'll just add it, in the interest of spreading potentially relevant knowledge as wide as possible.

PR #5376, opened 2016-11-11
Merge stable to master

Magisus:
This brings the fix for the sensitive_data_type test from stable to master.


Magisus:
The test refactor is broken on Solaris 10. Might as well fix before this merge up.

PR #5378, opened 2016-11-14
(PUP-6915) Specify log destination for sensitive data type test

Magisus:
This is a cherry-pick from stable to LTS, as well as an additional fix to allow the test to work on Solaris.

By default, puppet logs to different syslog directories depending on the
platform it is running on. When Fedora recently changed how to access
system logs, this test broke, since it was explicitly looking for
specific default log locations on various platforms. This commit refactors the test
to specify a temp file to receive the logs, thereby avoiding brittle
platform-specific code.


McdonaldSeanp:
👍

PR #5379, opened 2016-11-15
(PUP-6861) Revert PUP-6457 bad mount option handling

branan:
Revert "Merge pull request #5125 from HAIL9000/issue/master/PUP-6457_fail_on_bad_mount_options"

This reverts commit bcd1654afc5c2b4aa1397f7e0c1f13e02e269883, reversing
changes made to f4b86e08fa87159a84bc819458fdbe7017bcc088.

The original change relied on the idea that `/etc/mtab` would contain
an exact copy of the options specified in `/etc/fstab`. Unfortunately,
this is not correct, and causes this change to be overly-aggressive in
thinking that remounting failed.

Since there does not currently appear to be a way to check that a
mounted filesystem is in-sync with `fstab`, we unfortunately cannot
implement PUP-6457 as-described.

### DIFF
--- a/lib/puppet/provider/mount/parsed.rb
+++ b/lib/puppet/provider/mount/parsed.rb
@@ -209,20 +209,14 @@ Puppet::Type.type(:mount).provide(
 
     # Update fstab entries that are mounted
     providers.each do |prov|
-      mounts.delete_if do |mount|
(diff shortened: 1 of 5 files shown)


Magisus:
👍

PR #5380, opened 2016-11-15
(MAINT) Skip puppet_lookup_cmd.rb test when no non-master agents present

camlow325:
Previously, if the host configuration for a `puppet_lookup_cmd.rb` test
run were to include a single node which has both the 'master' and 'agent'
role, the 'host' would be set to nil, causing the test to fail as it hit
the first line of code which tried to make calls on the 'host' object.

This commit causes the `puppet_lookup_cmd.rb` test to be skipped if no
agent nodes that do not have the master role are present.

### DIFF
--- a/acceptance/tests/parser_functions/puppet_lookup_cmd.rb
+++ b/acceptance/tests/parser_functions/puppet_lookup_cmd.rb
@@ -2,6 +2,9 @@ test_name "Puppet Lookup Command"
 # doc:
 # https://docs.puppetlabs.com/puppet/latest/reference/lookup_quick_module.html
 
+host = agents.find { |host| host != master }
+skip_test "No agents present that are not running on a master" if host.nil?
+
 @module_name = "puppet_lookup_command_test"
 
 ### @testroot = "/etc/puppetlabs"
@@ -11,7 +14,6 @@ test_name "Puppet Lookup Command"
 @confdir = "#{@testroot}/puppet"
 
 @mastername = on(master, facter('fqdn')).stdout.chomp
-host = agents.find { |host| host != master }
 @agentname = on(host, facter('fqdn')).stdout.chomp
 
 @master_opts = {


camlow325:
@seangriff Since I think you wrote this test originally, would appreciate it if you could take a look at this change.  Thanks!

MosesMendoza:
I'm also curious - would this ever be run in a context where more than one host had the `agent` role and not the `master` role? As written I think the test seems to assume exactly one `agent` host - maybe it should also skip if there's more than one?

camlow325:
@MosesMendoza Seems theoretically possible, yeah, although I don't think any of our current OSS pipelines do this.  I can change the PR to skip for the case that there's not exactly one node that has the `agent` role but not the `master` role.

camlow325:
@MosesMendoza It seems like Travis doesn't like the header on my commit message.  Should I change it from (MAINT) to (maint) to keep it happy?

MosesMendoza:
@camlow325 actually that check is failing on a commit unrelated to your PR - now that @Iristyle's https://github.com/puppetlabs/puppet/pull/5365 is in we can see the range it's checking is off. Perhaps related, it looks like from your branch name this is intended for the LTS-1.7 branch. Should we retarget there?

camlow325:
Ah, whoops, yeah, I did intend to target this at LTS-1.7.  Will retarget, thanks.

camlow325:
@MosesMendoza I changed the base branch to LTS-1.7.  I think this test might actually be viable to run in a matrix configuration where more than one non-master agent node is present.  The test passed for me when run in a configuration where I had one node with master and agent roles and two nodes with just the agent role.  I think the test would just validate the behavior on one of the non-master agent nodes rather than all of them - which seems fine to me.  Maybe it would be good to get @seangriff's opinion as to whether or not there's more we want to do with this one?

acceptance/tests/parser_functions/puppet_lookup_cmd.rb
```diff
@@ -2,6 +2,9 @@
 # doc:
 # https://docs.puppetlabs.com/puppet/latest/reference/lookup_quick_module.html
 
+host = agents.find { |host| host != master }
+skip_test "No agents present that are not running on a master" if host.nil?
+
```
joshcooper: I thought the `lookup` command was only applicable when running on the same host as the master, so I'm not sure why it's important for the master to have an agent running on it?

camlow325: @joshcooper Yeah, thanks for pointing that out.  I obviously didn't really look at what this test is doing when I made this change.  I agree it doesn't really matter for the purpose of this test if there's an agent not running on the master - or even that there's an agent running anywhere for that matter.  I reworked my code change to just allow the two node names that this test is exercising for lookups be different from each other - using two hard-coded values which have nothing to do with the beaker host configuration.  This should hopefully allow this test to be run successfully if the configuration has just a single host using the 'master' and an 'agent' roles or any of the more common split node configurations that we typically run in CI.  I put up a replacement PR with this change as #5403 and am just going to close this one out.

PR #5381, opened 2016-11-16
(PUP-6918) Propagate error with CSR mismatch

Iristyle:
- When the agents public key does not match what's in the CSR
   generate a message for the user.

### DIFF
--- a/lib/puppet/ssl/host.rb
+++ b/lib/puppet/ssl/host.rb
@@ -227,7 +227,29 @@ ERROR_STRING
   # Generate all necessary parts of our ssl host.
   def generate
     generate_key unless key
-    generate_certificate_request unless certificate_request
+    # ask indirector to find any existing requests and download them
+    existing_request = certificate_request
+
+    # if CSR downloaded from master, but the local keypair was just generated and
+    # does not match the public key in the CSR, fail hard
+    if !existing_request.nil? &&
+      (key.content.public_key.to_s != existing_request.content.public_key.to_s)
+
+      raise Puppet::Error, <<ERROR_STRING
+The CSR retrieved from the master does not match the agent's public key.
+CSR fingerprint: #{existing_request.fingerprint}
+CSR public key: #{existing_request.content.public_key.to_text}
+Agent public key: #{key.content.public_key.to_text}
+To fix this, remove the CSR from both the master and the agent and then start a puppet run, which will automatically regenerate a CSR.
+On the master:
+  puppet cert clean #{Puppet[:certname]}
+On the agent:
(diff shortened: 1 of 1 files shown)


Iristyle:
@joshcooper / @adrienthebo have a look.

lib/puppet/ssl/host.rb
```diff
@@ -227,7 +227,29 @@ def validate_certificate_with_key
...
+      (key.content.public_key.to_s != existing_request.content.public_key.to_s)
+
+      raise Puppet::Error, <<ERROR_STRING
+The CSR retrieved from the master does not match the agent's public key.
+CSR fingerprint: #{existing_request.fingerprint}
+CSR public key: #{existing_request.content.public_key.to_text}
+Agent public key: #{key.content.public_key.to_text}
+To fix this, remove the CSR from both the master and the agent and then start a puppet run, which will automatically regenerate a CSR.
+On the master:
+  puppet cert clean #{Puppet[:certname]}
```
joshcooper: Due to PUP-1916, I don't think this will remove the CSR. I'm working on a patch in my free time.

Iristyle: Ah, right - you had mentioned that.

Iristyle:
The other thing that *might* be worth diving into on this one - the mismatch could be occurring because an interactive user without admin rights has generated a new set of SSL files *after* the service based master has run and submitted a CSR...  so it might be worth changing the message a bit taking that into account.

lib/puppet/ssl/host.rb
```diff
@@ -227,7 +227,29 @@ def validate_certificate_with_key
   # Generate all necessary parts of our ssl host.
   def generate
     generate_key unless key
-    generate_certificate_request unless certificate_request
+    # ask indirector to find any existing requests and download them
+    existing_request = certificate_request
+
+    # if CSR downloaded from master, but the local keypair was just generated and
```
adrienthebo: I don't believe that agents ever fetch the CSR from the master, have you seen otherwise?

lib/puppet/ssl/host.rb
```diff
@@ -227,7 +227,29 @@ def validate_certificate_with_key
...
+    # does not match the public key in the CSR, fail hard
+    if !existing_request.nil? &&
+      (key.content.public_key.to_s != existing_request.content.public_key.to_s)
+
+      raise Puppet::Error, <<ERROR_STRING
+The CSR retrieved from the master does not match the agent's public key.
+CSR fingerprint: #{existing_request.fingerprint}
+CSR public key: #{existing_request.content.public_key.to_text}
```
adrienthebo: The text of the public key is a respectable chunk of text and might not help in debugging it. Can we print the fingerprints of the public key for the host and the fingerprint of the public key in the CSR?

PR #5382, opened 2016-11-16
[WIP] (MODULES-4037) Used garethr/puppet-module-skeleton

HAIL9000:
Prior to this commit, 'puppet module generate' was using an
outdated module skeleton. Update it to use the skeleton used by
garethr's tool instead since it's more up to date and is already
more widely used by module developers. Plus this will create more
consistency for module authors since we'll no longer have two
different conflicting tools.


HAIL9000:
Travis is failing because the commit message references a MODULES ticket. I think this was just an oversight in the way the check was written, but if someone feels it's necessary we could migrate MODULES-4037 to a PUP ticket and reference that instead.

lib/puppet/module_tool/skeleton/templates/generator/.travis.yml
```diff
@@ -0,0 +1,39 @@
...
+bundler_args: --without development system_tests
+before_install: rm Gemfile.lock || true
+rvm:
+  - 1.8.7
+  - 1.9.3
+  - 2.0.0
+  - 2.1.0
+script: bundle exec rake test
+env:
+  - PUPPET_GEM_VERSION="~> 2.7.0"
```
bmjen: #sadness

HAIL9000: Oh gosh :x Clearly this was pretty out of date

HAIL9000:
Please don't merge this PR yet. It was filed prematurely and we're in the middle of more discussion on whether or not this is the right approach

HAIL9000:
We've concluded this isn't the write approach, planning to rescope and ticket this work. As such, closing this pull request.

PR #5383, opened 2016-11-16
(PUP-6924) Improve serialization of parameterized collection types

thallgren:
In this commit:
- The `element_type` attribute was removed from the `Collection` type and
  instead added to the `Array` type and as `value_type` to the `Hash` type.
- Serialization of registered `Object` types now uses the parameter names
  and order of the #initialize method to determine in which order the
  contained attributes are serialized.

### DIFF
--- a/lib/puppet/pops/types/p_object_type.rb
+++ b/lib/puppet/pops/types/p_object_type.rb
@@ -358,14 +358,14 @@ class PObjectType < PMetaType
   #   @param i12n_hash [Hash{String=>Object}] The hash describing the Object features
   #
   # @api private
-  def initialize(name_or_i12n_hash, i12n_hash_expression = nil)
+  def initialize(i12n_hash, i12n_hash_expression = nil)
     @attributes = EMPTY_HASH
     @functions = EMPTY_HASH
 
-    if name_or_i12n_hash.is_a?(Hash)
-      initialize_from_hash(name_or_i12n_hash)
+    if i12n_hash.is_a?(Hash)
+      initialize_from_hash(i12n_hash)
     else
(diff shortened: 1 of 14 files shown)


lib/puppet/pops/types/p_object_type.rb
```diff
@@ -491,6 +491,35 @@ def parameter_info(attr_readers = false)
...
+          if index.nil?
+            raise Serialization::SerializationError,
+              "Initializer for class #{impl_class.name} parameter '#{ip}' does not match any of the the attributes of type #{name}"
+          end
+          new_param_types << param_types[index]
+        end
+        param_names = init_param_names
+        param_types = new_param_types
+      end
+    end
```
thallgren: They are related to ordering according to parameters in the initialize function and very much related.

lib/puppet/pops/types/p_timespan_type.rb
```diff
@@ -1,10 +1,10 @@
...
 module Types
   class PAbstractTimeDataType < PAbstractRangeType
-    # @param [AbstractTime] min lower bound for this type. Nil or :default means unbounded
-    # @param [AbstractTime] max upper bound for this type. Nil or :default means unbounded
-    def initialize(min = nil, max = nil)
-      super(convert_arg(min, true), convert_arg(max, false))
+    # @param from [AbstractTime] lower bound for this type. Nil or :default means unbounded
+    # @param to [AbstractTime] upper bound for this type. Nil or :default means unbounded
+    def initialize(from = nil, to = nil)
+      super(convert_arg(from, true), convert_arg(to, false))
```
thallgren: Again, parameter names that must conform to names in the type.

PR #5385, opened 2016-11-17
(PUP-6921) Improve separation of concern between String and Enum types

thallgren:
Before this commit, the `PStringType` had two attributes. A `size_type` and
a `values` array. An inferred String would hence result in a String type
with two complex objects and all checkes would then take both objects into
account.

This commit simplifies the `PStringType` so that it only has one attribute.
The attribute can be either a size_type or a string value.

Since this means that a String can hold max one string value, the common
type of two different strings is now changed to instead be an Enum. I.e.
the inferred element type of the array ['a', 'b']. The logic for computing
assignability and commonality between Enum and String has been improved.

### DIFF
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -703,18 +703,11 @@ class PEnumType < PScalarType
 
   def generalize
     # General form of an Enum is a String
-    PStringType::DEFAULT
-  end
-
-  def normalize(guard = nil)
-    # General form of an Enum is a String
-    case @values.size
-    when 0
+    if @values.empty?
       PStringType::DEFAULT
-    when 1
-      PStringType.new(@values[0])
     else
-      self
+      range = @values.map(&:size).minmax
+      PStringType.new(PIntegerType.new(range.min, range.max))
     end
   end
 
@@ -748,7 +741,7 @@ class PEnumType < PScalarType
     end
     case o
       when PStringType
-        # if the set of strings are all found in the set of enums
+        # if the contained string is found in the set of enums
         v = o.value
         !v.nil? && svalues.any? { |e| e == v }
       when PEnumType
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -701,6 +701,23 @@ class PEnumType < PScalarType
     block_given? ? r.each(&block) : r
   end
 
+  def generalize
+    # General form of an Enum is a String
+    PStringType::DEFAULT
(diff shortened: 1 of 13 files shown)


hlindberg:
what happens when generalizing an Enum? Does Enum[a, bb, ccc] generalize to String[1,3] ?

lib/puppet/pops/types/types.rb
```diff
@@ -701,6 +701,23 @@ def each(&block)
...
+    # General form of an Enum is a String
+    case @values.size
+    when 0
+      PStringType::DEFAULT
+    when 1
+      PStringType.new(@values[0])
+    else
+      self
+    end
+  end
```
hlindberg: Hm, so if we infer String for "abc", it is `String[3,3]` (but remembering that it is "abc"). It seems like one normalization is that it is *only* `String[3,3]` - i.e not compatible with `Enum[abc]`. Now it looks like it will be normalized to the same thing as `Enum[abc]`, only encoded in the String type instead of in an enum.

This logic looks wrong to me.

lib/puppet/pops/types/types.rb
```diff
@@ -732,7 +749,8 @@ def _assignable?(o, guard)
     case o
       when PStringType
         # if the set of strings are all found in the set of enums
```
hlindberg: This comment is now not true

lib/puppet/pops/types/types.rb
thallgren: There is no normalization on a String type. It simply returns itself.

There didn't use to be any normalization on the Enum type either. I thought it made sense to normalize a one-element Enum to a String since there's not much point in having it as a Enum unless it has multiple choices. I can remove it for sure.

thallgren:
An Enum[a, bb, ccc] is generalized into String, not String[1,3]. This is analog to a String inferred from "ccc". It will also be generalized into a String (without parameters). For String, this hasn't changed. For Enum, it has. There a call to generalize used to be a no-op.

The reason for the change is that it works better with the TypeMismatchDescriber. It uses generalize when it knows that the types differ in such a way that there is no need to show type details. So rather than showing "expected Integer, got Enum[a, b, c]" it now shows, "expected Integer, got String". Good to keep in mind here is that generalization almost only happens on inferred types and there's no runtime correspondence to an Enum. The Enum pops up when inferring collection types (from strings).

An alternative would be to keep the old behavior and instead move this to the TypeMismatchDescriber. There, it's more certain that the "actual" is an inferred value.

PR #5387, opened 2016-11-17
(PUP-6759) Consider indirect systemd units to be disabled

whopper:
This commit updates the systemd provider to recognize services
with the 'indirect' enabled state. These units are disabled,
and thus when querying whether they are enabled we now return
false.

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -29,7 +29,7 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
   def self.instances
     i = []
     output = systemctl('list-unit-files', '--type', 'service', '--full', '--all',  '--no-pager')
-    output.scan(/^(\S+)\s+(disabled|enabled|masked)\s*$/i).each do |m|
+    output.scan(/^(\S+)\s+(disabled|enabled|masked|indirect)\s*$/i).each do |m|
       i << new(:name => m[0])
     end
     return i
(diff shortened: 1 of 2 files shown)


whopper:
See https://www.freedesktop.org/software/systemd/man/systemctl.html for the description of the 'indirect' state:

"The unit file itself is not enabled, but it has a non-empty Also= setting in the "[Install]" unit file section, listing other unit files that might be enabled."

Ran all of the service acceptance tests in RedHat 7 to make sure this didn't cause any regressions. All passed:

```
       Total Suite Time: 225.00 seconds
      Average Test Time: 14.06 seconds
              Attempted: 16
                 Passed: 7
                 Failed: 0
                Errored: 0
                Skipped: 9
                Pending: 0
                  Total: 16
```

adrienthebo:
```
[adrien@sky puppet]$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
[adrien@sky puppet]$ systemctl list-unit-files --type service --all --full --no-pager|grep talk
talk.service                                indirect
[adrien@sky puppet]$ bundle exec puppet resource service talk
service { 'talk':
  ensure => 'stopped',
  enable => 'true',
}
[adrien@sky puppet]$ git checkout whopper-PUP-6759/master/systemd-indirect
Switched to branch 'whopper-PUP-6759/master/systemd-indirect'
Your branch is up-to-date with 'whopper/PUP-6759/master/systemd-indirect'.
[adrien@sky puppet]$ bundle exec puppet resource service talk
service { 'talk':
  ensure => 'stopped',
  enable => 'false',
}
```

:+1:

PR #5388, opened 2016-11-18
(PUP-6926) Enable lookup of rich data values

thallgren:
This PR changes the Pcore type used to validate data found by a lookup
so that it may contain any type except Runtime. The key and value types are
registered as aliases `Puppet::LookupKey` and `Puppet::LookupValue`. The
latter alias is self-recursive in that keys and values found in hashes and
arrays must conform to their respective type, even if nested.

### DIFF
--- a/lib/puppet/pops/lookup/data_provider.rb
+++ b/lib/puppet/pops/lookup/data_provider.rb
@@ -16,7 +16,7 @@ module DataProvider
     if @key_type.nil?
       (@key_type, @value_type) = Pcore::register_aliases(
         # The Pcore type for all keys and subkeys in a data hash.
-        'Puppet::LookupKey' => 'Variant[String,Boolean,Numeric]',
+        'Puppet::LookupKey' => 'Variant[String,Numeric]',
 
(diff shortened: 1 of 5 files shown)


lib/puppet/pops/lookup/data_provider.rb
```diff
@@ -2,20 +2,38 @@ module Puppet::Pops
...
-    Types::PScalarType::DEFAULT,
-    Types::PUndefType::DEFAULT,
-    Types::PHashType::DEFAULT,
-    Types::PArrayType::DEFAULT
-  ])
+  def self.ensure_types_initialized
+    if @key_type.nil?
+      (@key_type, @value_type) = Pcore::register_aliases(
+        # The Pcore type for all keys and subkeys in a data hash.
+        'Puppet::LookupKey' => 'Variant[String,Boolean,Numeric]',
```
hlindberg: Why Boolean? What is the purpose of that?

thallgren: I'll remove it.

PR #5391, opened 2016-11-19
(PUP-6933) Change acceptance test to accept new explanations

thallgren:
The Puppet lookup explanations changed wording slightly as a result of the
changes introduced in PUP-6780. This commit adjusts the acceptance tests
accordingly.


hlindberg:
+1 thanks @thallgren for fixing this and merging

PR #5392, opened 2016-11-21
(PUP-6844) Ignore Mercurial artifacts when downloading plugins

whopper:
Just as we ignore SVN, CVS, and Git artifacts while pluginsyncing,
we should also ignore those of Mercurial. This commit adds files
with the `.hg` extension to the list of extensions in the
`pluginsignore` option.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1719,7 +1719,7 @@ EOT
     },
 
     :pluginsignore => {
-        :default  => ".svn CVS .git",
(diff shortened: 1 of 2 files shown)


branan:
Looks like the commit checker is still pulling in random stuff. This passed Travis just fine

whopper:
@branan ugh, yeah, this commit slipped in in October and broke the check: https://github.com/puppetlabs/puppet/commit/6e7949803cf4dde355f3fe140fdeee29ebbdba44

PR #5393, opened 2016-11-23
Kick should support splay and environment

dmitryilyin:
When the Puppet agent is kicked, it should
receive an option which will allow a user
to disable splay and initiate the Puppet
run instantly.

Add support for the environment options
for the puppet kick to set which environment the
agent should use.

Yes, I know that Puppet kick is a deprecated mechanics
and is removed in the later version.

### DIFF
--- a/lib/puppet/application/kick.rb
+++ b/lib/puppet/application/kick.rb
@@ -10,6 +10,7 @@ class Puppet::Application::Kick < Puppet::Application
   option("--ping","-P")
   option("--test")
   option("--ignoreschedules")
+  option("--splay","-s")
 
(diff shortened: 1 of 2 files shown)


joshcooper:
Hi @dmitryilyin, thank you for your contribution! It looks like there are valid travis CI failures. That said, [PE 3.8 is going EOL in 29 days](https://puppet.com/misc/puppet-enterprise-lifecycle), and once that happens, we won't be releasing new versions of open source Puppet 3.8.x. I'm going to close this PR, but know that we do appreciate your contribution, and we welcome any PRs you submit against puppet 4.x.

PR #5394, opened 2016-11-24
(PUP-6741) Correctly trigger tagged resources dependent on dynamically generated resources

MosesMendoza:
When generating resources with
Puppet::Transaction::AdditionalResourceGenerator#eval_generate, we create a completed_ whit resource to "contain" the generated resource such that when its dependencies are satisifed, it effectively signals that the generating resource evaluation has completed. However, prior to this commit, the applicable tags of the parent resource were not propagated to the whit resource (they are propogated to all child resources later in Puppet::Transaction::AdditionalResourceGenerator#add_resource). Because the whit does not have any of the tags of the parent resource, if puppet is filtering resources by tags, resources in a subscribing relationship with the parent generating resources would not be refreshed, even when the generating resource issued a refresh event.

An example (as noted in PUP-6741) is an exec that subscribes to a recursive file resource, in a class tagged 'foo': puppet run with --tags foo will not evaluate the exec even if the file resource generates a refresh event.

This commit augments the tags of the whit with the applicable tags of the parent (generating) resource. Applicable here means all tags except the tag which references the name of the parent's resource type. Since all resources are automatically tagged with the name of their resource type, it seems unituitve to tag the whit with the type name of the parent resource, as it already has a "whit" tag.

Performance benchmarks were run before/after this change, as it's possible Puppet::Util::Tagging#tag could incur a small unnecessary performance penalty associated with tag validation (unnecessary because we're operating with two sets of tags that have already been validated). Puppet::Util::Tagging does not have a mechanism (probably on purpose) for augmenting an instance's tags in place without performing tag validation but a performance regression may have justified such a mechanism. However, no substantial changes were reported by the benchmarks.

### DIFF
--- a/lib/puppet/transaction/additional_resource_generator.rb
+++ b/lib/puppet/transaction/additional_resource_generator.rb
@@ -79,6 +79,13 @@ class Puppet::Transaction::AdditionalResourceGenerator
 
   def contain_generated_resources_in(resource, made)
     sentinel = Puppet::Type.type(:whit).new(:name => "completed_#{resource.title}", :catalog => resource.catalog)
+    # Tag the completed whit with all of the tags of the generating resource
+    # except the type name to allow event propogation to succeed beyond the whit
+    # "boundary" when filtering resources with tags. We include auto-generated
+    # tags such as the type name to support implicit filtering as well as
+    # explicit. Note that resource#tags returns a duplicate of the resource's
+    # tags.
+    sentinel.tag(*resource.tags)
     priority = @prioritizer.generate_priority_contained_in(resource, sentinel)
     @relationship_graph.add_vertex(sentinel, priority)
 
--- a/spec/integration/transaction_spec.rb
+++ b/spec/integration/transaction_spec.rb
@@ -266,6 +266,80 @@ describe Puppet::Transaction do
     expect(Puppet::FileSystem.exist?(newfile)).to be_truthy
   end
 
+  # Ensure when resources have been generated with eval_generate that event
+  # propagation still works when filtering with tags
+  context "when filtering with tags" do
+    context "when resources are dependent on dynamically generated resources" do
+      it "should trigger (only) appropriately tagged dependent resources" do
+        source = dir_containing('sourcedir', {'foo' => 'bar'})
+        target = tmpdir('targetdir')
+        file1 = tmpfile("file1")
+        file2 = tmpfile("file2")
+
+        file = Puppet::Type.type(:file).new(
+          :path    => target,
+          :source  => source,
+          :ensure  => :present,
+          :recurse => true,
+          :tag     => "foo_tag",
+        )
+
+        exec1 = Puppet::Type.type(:exec).new(
+          :path        => ENV["PATH"],
+          :command     => touch(file1),
+          :refreshonly => true,
+          :subscribe   => file,
+          :tag         => "foo_tag",
+        )
+
+        exec2 = Puppet::Type.type(:exec).new(
+          :path        => ENV["PATH"],
+          :command     => touch(file2),
+          :refreshonly => true,
+          :subscribe   => file,
+        )
+
+        Puppet[:tags] = "foo_tag"
+        catalog = mk_catalog(file, exec1, exec2)
+        catalog.apply
+        expect(Puppet::FileSystem.exist?(file1)).to be_truthy
+        expect(Puppet::FileSystem.exist?(file2)).to be_falsey
+      end
+
+      it "should trigger implicitly tagged dependent resources, ie via type name" do
+        file1 = tmpfile("file1")
+        file2 = tmpfile("file2")
+
+        exec1 = Puppet::Type.type(:exec).new(
+          :name        => "exec1",
+          :path        => ENV["PATH"],
+          :command     => touch(file1),
+        )
+
+        exec1.stubs(:eval_generate).returns(
+          [ (Puppet::Type.type(:notify).new :name => "eval1_notify")]
+        )
+
+        exec2 = Puppet::Type.type(:exec).new(
+          :name        => "exec2",
+          :path        => ENV["PATH"],
+          :command     => touch(file2),
+          :refreshonly => true,
+          :subscribe   => exec1,
+        )
+        exec2.stubs(:eval_generate).returns(
+          [ (Puppet::Type.type(:notify).new :name => "eval2_notify")]
+        )
+
+        Puppet[:tags] = "exec"
+        catalog = mk_catalog(exec1, exec2)
+        catalog.apply
+        expect(Puppet::FileSystem.exist?(file1)).to be_truthy
+        expect(Puppet::FileSystem.exist?(file2)).to be_truthy
+      end
+    end
+  end
+
   describe "skipping resources" do
     let(:fname) { tmpfile("exec") }
 
--- a/spec/unit/transaction/additional_resource_generator_spec.rb
+++ b/spec/unit/transaction/additional_resource_generator_spec.rb
@@ -182,6 +182,17 @@ describe Puppet::Transaction::AdditionalResourceGenerator do
         'Generator[thing]', 'Whit[completed_thing]')
     end
 
+    it "should tag the sentinel with the tags of the resource" do
+      graph = relationships_after_eval_generating(<<-MANIFEST, 'Generator[thing]')
+        generator { thing:
+          code => 'notify { hello: }',
+          tag  => 'foo',
+        }
+      MANIFEST
+      whit = find_vertex(graph, :whit, "completed_thing")
+      expect(whit.tags).to be_superset(['thing', 'foo', 'generator'].to_set)
+    end
+
     it "should contain the generated resources in the same container as the generator" do
       catalog = compile_to_ral(<<-MANIFEST)
         class container {


MosesMendoza:
This should also address PUP-5777

MosesMendoza:
ping @joshcooper @Iristyle for review

lib/puppet/transaction/additional_resource_generator.rb
```diff
@@ -79,6 +79,12 @@ def replace_duplicates_with_catalog_resources(generated)
 
   def contain_generated_resources_in(resource, made)
     sentinel = Puppet::Type.type(:whit).new(:name => "completed_#{resource.title}", :catalog => resource.catalog)
+    # Tag the completed whit with all of the tags of the generating resource
+    # except the type name to allow event propogation to succeed beyond the whit
+    # "boundary" when filtering resources with tags. All resources are tagged
+    # with their type name by default, so we don't include this tag to avoid any
+    # confusion from the whit containing tags for two resource types.
+    sentinel.tag(*resource.tags.delete(resource.type.to_s))
```
Iristyle: Interesting - I'm going to have to dig into this code to understand it.  But kind of amazing that you could address the problem with a single line of code.

spec/integration/transaction_spec.rb
```diff
@@ -266,6 +266,48 @@ def touch(path)
     expect(Puppet::FileSystem.exist?(newfile)).to be_truthy
   end
 
+  # Ensure when resources have been generated with eval_generate that event
+  # propogation still works when filtering with tags
```
Iristyle: Minor `propagation`

lib/puppet/transaction/additional_resource_generator.rb
Iristyle: Wondering if we should be duplicating the `resource.tags` prior to the `delete`, like `resource.tags.dup.delete(resource.type.to_s))`?

`delete` will mutate the `tags` associated with the resource - have to look more at the code to see if thats OK or not

MosesMendoza: I was worried about that too but the `tags` method actually returns a dup of the resources's tags instance variable so I think this is ok?

Iristyle: Yeah, right you are - that was unexpected!

```
[11] pry(#<Puppet::Transaction::AdditionalResourceGenerator>)> resource.tags.delete('file')
=> #<Puppet::Util::TagSet: {"foo_tag"}>
[12] pry(#<Puppet::Transaction::AdditionalResourceGenerator>)> resource.tags
=> #<Puppet::Util::TagSet: {"file", "foo_tag"}>
```

Iristyle: Another way to get the list of tags given for the resource, without pulling in its type name is to use `resource.parameters[:tag].value`

But I'm not sure if there's a case where there are other tags that may have been calculated / added... that are a part of `resource.tags`, but not `resource.parameters[:tag].value`

Iristyle: So changing to `resource.parameters[:tag].value` makes the new tests in `additional_resource_generator_spec.rb` fail... so that probably answers my question.

MosesMendoza:
A couple relevant change sets / tickets /discussions that may help provide context:

* [#3382 - bug that led to creation of whits](https://projects.puppetlabs.com/issues/3382)
* and its resulting [change set](https://github.com/puppetlabs/puppet/commit/59a23d69c20a8ab1f7dc9ff1d109305f9027641c)

* [puppet-dev discussion that led to whits used to replace containers in the graph](https://groups.google.com/forum/#!topic/puppet-users/Wv7H6BIP2Jk)
* and [#5200 - the associated bug](https://projects.puppetlabs.com/issues/5200)
* and its resulting [change set](https://github.com/puppetlabs/puppet/commit/2a6c6cb8fabf82d2f2127c90db670c1a856427c5)  - this is when we introduced the behavior my PR is addressing

lib/puppet/transaction/additional_resource_generator.rb
Iristyle: Upon further research, it looks like perhaps we shouldn't remove any of the tags from the sentinel and should just copy them all over.  No other tagging code filters out like that *and* containment docs describe a bit more about how it should work (in general):

https://docs.puppet.com/puppet/latest/reference/lang_containment.html

spec/unit/transaction/additional_resource_generator_spec.rb
```diff
@@ -182,6 +182,31 @@ def find_vertex(graph, type, title)
...
+    it "should not tag the sentinel with the type of the resource" do
+      graph = relationships_after_eval_generating(<<-MANIFEST, 'Generator[thing]')
+        generator { thing:
+          code => 'notify { hello: }',
+        }
+      MANIFEST
+      whit = find_vertex(graph, :whit, "completed_thing")
+      expect(whit).not_to be_tagged("generator")
+    end
+
```
MosesMendoza: @Iristyle got something rigged up that should be what we need here. Per conversation, removed `tags.delete` of generating resource type name, and added a new test to the integration section that should cover the use case of triggering dependent resources from dynamically `eval_generated` resources using implicit tagging. Since as far as I can tell none of the existing puppet types support both dynamically generating resources with `eval_generate` and also something like a `refreshonly` parameter out of the box, we temporarily augment the exec type with `eval_generate` to replicate what we would need for the use case.

Word to the wise - don't stub `eval_generate` with a single return value (even one that's dynamic) - no matter how many times its called, it will always return the first incantation, leading to an infinite explosion in graph edges during transaction evaluation as new edges "mysteriously" multiply to the exact same generated resource instance.

I.e., an initial iteration of this did:
```ruby
Puppet::Type::Exec.any_instance.stubs(:eval_generate).returns (
  [ (Puppet::Type.type(:notify).new :name => "notify#{rand(9999)}") ]
)
```

Stubbing instead with a list of comma-separated return values will return the first value on first incantation of the stubbed method, second value with the second incantation, and so on.

spec/integration/transaction_spec.rb
```diff
@@ -266,6 +266,78 @@ def touch(path)
...
+
+        file1 = tmpfile("file1")
+        file2 = tmpfile("file2")
+
+        exec1 = Puppet::Type.type(:exec).new(
+          :name        => "exec1",
+          :path        => ENV["PATH"],
+          :command     => touch(file1),
+        )
+
```
Iristyle: So can we do something like `exec1.stubs(:eval_generate)` on this instance (and also on `exec2`)?

MosesMendoza: seems reasonable - will give it a shot

MosesMendoza: works like a charm

Iristyle: 👍

spec/integration/transaction_spec.rb
```diff
@@ -266,6 +266,80 @@ def touch(path)
...
+        )
+        exec2.stubs(:eval_generate).returns(
+          [ (Puppet::Type.type(:notify).new :name => "eval2_notify")]
+        )
+
+        Puppet[:tags] = "exec"
+        catalog = mk_catalog(exec1, exec2)
+        catalog.apply
+        expect(Puppet::FileSystem.exist?(file1)).to be_truthy
+        expect(Puppet::FileSystem.exist?(file2)).to be_truthy
```
Iristyle: I have verified that changing to the old code of `sentinel.tag(*resource.tags.delete(resource.type.to_s))` prevents the second `exec` from running and creating the file.  With the change to `sentinel.tag(*resource.tags)` it works properly.

Awesome.

Even though this example is a little contrived, I think it packs a lot of value for an interesting use case.

spec/unit/transaction/additional_resource_generator_spec.rb
```diff
@@ -182,6 +182,22 @@ def find_vertex(graph, type, title)
...
+          tag  => 'foo',
+        }
+      MANIFEST
+      whit = find_vertex(graph, :whit, "completed_thing")
+      # this matcher relies on Puppet::Util::Tagging#tagged? which accepts a
+      # list of tags and returns true if _at least one_ tag is present. So we
+      # explicitly check both.
+      expect(whit).to be_tagged("thing")
+      expect(whit).to be_tagged("foo")
+      expect(whit).to be_tagged("generator")
```
Iristyle: I think these assertions can be reduced to:

```ruby
expect(whit.tags).to be_superset(['thing', 'foo', 'generator'].to_set)
```

Unfortunately no inline `set` notation, so must use `Set.new()` or `[].to_set`

spec/integration/transaction_spec.rb
```diff
@@ -266,6 +266,80 @@ def touch(path)
...
+        exec1.stubs(:eval_generate).returns(
+          [ (Puppet::Type.type(:notify).new :name => "eval1_notify")]
+        )
+
+        exec2 = Puppet::Type.type(:exec).new(
+          :name        => "exec2",
+          :path        => ENV["PATH"],
+          :command     => touch(file2),
+          :refreshonly => true,
+          :subscribe   => Puppet::Resource.new(:exec, "exec1")
```
Iristyle: The subscribe can be simplified to:

```
:subscribe => exec1
```

spec/integration/transaction_spec.rb
```diff
@@ -266,6 +266,80 @@ def touch(path)
...
+          :ensure  => :present,
+          :recurse => true,
+          :tag     => "foo_tag",
+        )
+
+        exec1 = Puppet::Type.type(:exec).new(
+          :path        => ENV["PATH"],
+          :command     => touch(file1),
+          :refreshonly => true,
+          :subscribe   => Puppet::Resource.new(:file, target),
```
Iristyle: The `:subscribe` here can be simplified to just `file` since we're already storing the instance in a variable:

```
:subscribe => file
```

spec/integration/transaction_spec.rb
```diff
@@ -266,6 +266,80 @@ def touch(path)
...
+          :tag         => "foo_tag",
+        )
+
+        exec2 = Puppet::Type.type(:exec).new(
+          :path        => ENV["PATH"],
+          :command     => touch(file2),
+          :refreshonly => true,
+          :subscribe   => Puppet::Resource.new(:file, target),
```
Iristyle: The `:subscribe` here can be simplified to just `file` since we're already storing the instance in a variable:

```
:subscribe => file
```

Iristyle:
Minor changes, a squash.. and I think we're ready to :shipit: !

MosesMendoza:
squashed and updated commit message re:deleted type name tag

MosesMendoza:
test fixups added

PR #5395, opened 2016-11-26
(PUP-6576) Deprecate data_binding_terminus

hlindberg:
This adds a deprecation warning if a data binding terminus other than 'hiera' or 'none' are assigned.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -419,7 +419,21 @@ deprecated and has been replaced by 'always_retry_plugins'."
     :data_binding_terminus => {
       :type    => :terminus,
       :default => "hiera",
-      :desc    => "Where to retrieve information about data.",
+      :desc    =>
+        "This setting has been deprecated in favor of using lookup. It will be removed in major release of puppet.
+
+         This setting controls which data binding terminus to use for global automatic data binding (across all environments).
+         Typical (deprecated) usage is to use the default value 'hiera'. After having configured lookup per environment, this
+         setting should be changed to an empty string (or nil) as that will turn off global data binding.",
+      :call_hook => :on_initialize_and_write,
+      :hook => proc do |value|
+        if Puppet[:strict] != :off
+          s_val = value.to_s # because sometimes the value is a symbol
+          unless s_val == 'hiera' || s_val == 'none' || value == '' || value.nil?
+            Puppet.deprecation_warning "Setting 'data_binding_terminus' is deprecated. Convert custom terminus to hiera 5 API."
+          end
+        end
+      end
     },
     :hiera_config => {
       :default => lambda do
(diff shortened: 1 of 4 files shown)


lib/puppet/defaults.rb
```diff
@@ -419,7 +419,20 @@ def self.default_diffargs
     :data_binding_terminus => {
       :type    => :terminus,
       :default => "hiera",
-      :desc    => "Where to retrieve information about data.",
+      :desc    =>
+        "This setting has been deprecated. Use of any value other than 'hiera' should instead be configured
+         in a version 5 hiera.yaml. Until this setting is removed, it controls which data binding terminus
```
MosesMendoza: @hlindberg do you know where the differences between hiera.yaml versions are documented? I looked on our docs site but did not find them.

lib/puppet/defaults.rb
```diff
@@ -419,7 +419,20 @@ def self.default_diffargs
...
+         to use for global automatic data binding (across all environments). By default this value is 'hiera'.
+         A value of 'none' turns off the global binding.",
+      :call_hook => :on_initialize_and_write,
+      :hook => proc do |value|
+        if Puppet[:strict] != :off
+          s_val = value.to_s # because sometimes the value is a symbol
+          unless s_val == 'hiera' || s_val == 'none' || value == '' || value.nil?
+            Puppet.deprecation_warning "Setting 'data_binding_terminus' is deprecated. Convert custom terminus to hiera 5 API."
```
MosesMendoza: A minor wording suggestion:
change
```
Convert custom terminus to hiera 5 API
```
to
```
Convert custom terminus to hiera.yaml API version 5
```

or something similar. Basically when I first read this I was confused and thought the suggestion was to upgrade to Hiera 5 - but Hiera 3 is current. Perhaps specifying that we're talking about version 5 of the API is more intuitive?

hlindberg: Actually, a custom terminus is not converted to hiera.yaml version 5 - the terminus has to be rewritten to be callable as a puppet function (implemented in Ruby or Puppet).
So - it is far more than just changing the configuration. We expect that there are close to 0 users that have written custom data binding termini, but we don't know for sure - hence this deprecation.

We have made a decision to using "hiera 5" as the name for "lookup". The hiera project codebase will remain at version 3 (versions 4 and 5 are inside the puppet code base).
We will continue to include hiera version 3 with puppet agent for backwards compatibility for backends until Puppet 6.0 or so when it will be removed. Thus giving hiera backend authors > 1 year to migrate to the hiera 5 API (a.k.a "lookup").

With all that I am saying that I think it is correct the way it is in the PR.

thallgren:
@MosesMendoza this looks OK to me. Please merge if you approve the change of wording.

lib/puppet/defaults.rb
MosesMendoza: 👍  - thanks for explaining

PR #5396, opened 2016-11-29
(maint) Fix two minor glitches in the TypeFormatter

thallgren:
This commit fixes two minor problems in the TypeFormatter.
1. A String, when output in debug mode, didn't output a contained string
   value correctly.
2. A qualified type name resolved using a TypeSet reference would raise
   an exception due to a bogus method call in TypeSet#name_for method.

### DIFF
--- a/lib/puppet/pops/types/p_type_set_type.rb
+++ b/lib/puppet/pops/types/p_type_set_type.rb
@@ -216,27 +216,18 @@ class PTypeSetType < PMetaType
   # @return [String] the name by which the type is referenced within this type set
   #
   # @api private
-  def name_for(t)
+  def name_for(t, default_name)
     key = @types.key(t)
     if key.nil?
       qname = t.name
       if @references.empty?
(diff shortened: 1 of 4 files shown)


MosesMendoza:
Should this PR have tests associated with it?

thallgren:
Of course the PR should include tests. And now it has. Incidentally, one of the tests revealed a bug, which is now also fixed.

hlindberg:
Merging. (Would have preferred if this had a ticket rather than being a (maint) fix).

thallgren:
@henrik, regarding making this a ticket or not. A chose not to, because no user has ever been exposed to this. Using debug_print in the TypeFormatter is not possible form Puppet. I don't think it's possible to provoke the other problem either the way things stand now. A ticket would create just create extra work for QA, for no apparent reason.

henrik:
Wrong Henrik. Unsubscribing.

PR #5399, opened 2016-11-30
Revert merge of PR#5094 - switch to semantic_puppet

MosesMendoza:
Revert "Merge pull request #5094 from thallgren/issue/pup-4742/switch-to-semantic_puppet-gem"

This reverts commit b60a570a2eb86a9a09af01043c1ca3f1b955e3fd, reversing
changes made to 8a9e7d49d7d00da400cfdc6d06cd54c1ca29a8aa.

This changeset failed CI, possibly due to missing puppet server
requirements/configuration

### DIFF
--- a/.gemspec
+++ b/.gemspec
@@ -36,5 +36,4 @@ Gem::Specification.new do |s|
   s.specification_version = 3
   s.add_runtime_dependency(%q<facter>, [">= 2.0.1", "< 4"])
   s.add_runtime_dependency(%q<hiera>, [">= 2.0", "< 4"])
-  s.add_runtime_dependency(%q<semantic_puppet>, ['>= 0.1.3', '< 2'])
 end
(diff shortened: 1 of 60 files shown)


joshcooper:
We ran into a few problems that need to be sorted out before we can re-revert:

1. puppet acceptance tests are pinned to puppetserver 2.5.0, which doesn't have the necessary gem. See https://tickets.puppetlabs.com/browse/PUP-6956
1. Add the `semantic_puppet` gem dependency to https://github.com/puppetlabs/puppet/blob/master/ext/project_data.yaml#L19, which is used when building puppet as a gem (prior to shipping to rubygems.org). Puppet's `Gemfile` is only used when doing local development using bundler. `.gemspec`  is used when other code, e.g. stdlib puppet module, has a dependency on puppet and is using a git checkout of puppet.
1. Additional work may be required so that `bundle exec rake ci:test:gem` works?

joshcooper:
The first two issues should be addressed in https://github.com/puppetlabs/ci-job-configs/pull/2098 and https://github.com/puppetlabs/puppet/pull/5400.

Since `bundle exec rake ci:test:gem` isn't run automatically in CI, I think we can revisit that issue once we merge PR#5400 and the dust settles.

PR #5401, opened 2016-12-01
(PUP-6961) Add risk tag to acceptance tests

johnduarte:
Add a beaker tag to all of the existing acceptance
tests to reflect their assessed risk level. These tags will be used
to assist with tiering the execution of the test suite in order
to provide faster test suite execution.

### DIFF
--- a/acceptance/README.md
+++ b/acceptance/README.md
@@ -145,6 +145,34 @@ There also may be scenarios where you want to specify the host(s) to release. E.
     HOST_NAMES=lvwwr9tdplg351u bundle exec rake ci:release_hosts
     HOST_NAMES=lvwwr9tdplg351u,ylrqjh5l6xvym4t bundle exec rake ci:release_hosts
 
+### Running tagged tests
+
+Tests runs can be segmented by tags defined in the tests themselves. These
+tags have been standardized into the following.
+
+* `risk:high`
+* `risk:medium`
+* `risk:low`
+
+For CI runs at Puppet, these are used to adjust the cadence of test execution.
+They get run in the following pattern in order to provide timely feedback.
+
+* Per commit: risk:high + untagged tests
+* Nightly:    risk:high + risk:medium + untagged tests
+* Weekly:     risk:high + risk:medium + risk:low + untagged tests
+
+Individual tests are tagged in accordance with the frequency at which the test
+should be run to balance its value and its cost. Tests without tags are
+are not filtered by the Beaker tag facility an will therefore always be run.
+
+These tags can be utilized by modifying the `local_options.rb` file with the
+`tag` option, or using the `OPTIONS` environment variable when running the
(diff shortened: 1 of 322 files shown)


johnduarte:
@MosesMendoza I am interested in your thoughts regarding tests that can be tagged as low risk.

Iristyle:
Wow - do we have a guide / doc somewhere that states how the levels were selected?

johnduarte:
@Iristyle creating a doc is a great idea. We do not currently have one. Basic heuristics for this PR were:

1. Default to medium
1. Define existing per commit tests as high
1. Apply low to zfs and zone tests that pertain to a smaller subset of our supported matrix.

MosesMendoza:
@johnduarte thoughts on adding that information to a README? i.e., maybe `acceptance/README.md` (does not exist yet)

johnduarte:
@MosesMendoza I have added information about the tags to the README.md. Let me know if more detail should be incorporated.

acceptance/README.md
```diff
@@ -145,6 +145,33 @@ There also may be scenarios where you want to specify the host(s) to release. E.
...
     HOST_NAMES=lvwwr9tdplg351u,ylrqjh5l6xvym4t bundle exec rake ci:release_hosts
 
+### Running tagged tests
+
+Tests runs can be segmented by tags defined in the tests themselves. These
+tags have been standardized into the following.
+
+* risk:high
+* risk:medium
+* risk:low
```
johnduarte: Great questing @glennsarti . I have amended the README to clarify that untagged tests will be run weekly.

glennsarti: Hrmmm.... I would argue that untagged tests are high risk as;
1 - Having no tag (probably) shows no thought has gone into how risky they are, therefore we should run them as often as possible to figure out how risky they really are
2  - It's far too easy to forget to tag a test and then only find out  many days later that it failed

johnduarte: @glennsarti you have brought up a requirement that @MosesMendoza called to my attention as well. Your concern about new tests will be addressed by adding any tests that have changes to them dynamically to the per commit runs.

I would argue against defaulting untagged tests to the per commit runs. The purpose of the tiering workflow is to make sure the cost of execution for frequent CI runs remains as low as possible. I agree that there is risk in having untagged tests, but most new tests should not jump into the per commit runs.

Automatically running changed (which would include new) tests in the per commit run should mitigate this risk. I am also interested in your thoughts about adding a commit hook to ensure that committed tests contain a tag.

johnduarte: As it turns out, the way that Beaker implements tag filtering a test without any tags is not filtered out when using the '--tag' option facility. Therefore, untagged tests will always be run when using the '--tag' option.
@glennsarti , this behavior addresses your concern about frequently running untagged tests.
I have updated the README to reflect this.

PR #5402, opened 2016-12-01
(PUP-6963) Set default provider to systemd for CoreOS systems

garethr:


whopper:
Thanks @garethr, this looks good! 👍

PR #5403, opened 2016-12-01
(MAINT) Fix puppet_lookup_cmd test for single master/agent host config

camlow325:
Previously, the `puppet_lookup_cmd.rb` acceptance test used the FQDN of
the master and the FDQN of an agent not running on the master in order
to have two unique hosts to use for node lookups when running the puppet
lookup command.  When run under a host configuration with only a single
node that has both the 'master' and 'agent' role, this test would fail
because a non-nil name could not be found for an agent not on the
master.

This commit reworks the test to use two arbitrary node names, node1 and
node2, which are exercised for puppet lookups.  This change allows the
test to be runnable in any case where there is a master in the host
configuration, even if the only agent in the host configuration is the
same as the master.


joshcooper:
@er0ck can you review/merge this?

PR #5404, opened 2016-12-02
(PUP-6967) Reload puppetserver using HUP

joshcooper:
Bump beaker to 3.5.x, which added support for reloading puppetserver
using HUP.

Beaker will fail to reload the service if it isn't already running and
it will fall back to bouncing the service using puppet's resource
service ensure=stopped and ensure=running commands. To ensure reloading
always works, we now start puppetserver in the 045_EnsureMasterStarted
pre-suite step. We don't test against passenger or webrick, so there's
no point in making the step conditional on the server type.

Prior to commit 59cce2311, Beaker's `with_puppet_running_on` method would
start puppetserver twice. Once in the beginning before the test, and
once after. Each start takes ~30 seconds per test, and that caused an
overall 1+ hour increase in test times as compared to when we tested
against passenger packages.

In commit 59cce2311, we set `restart_when_done` to true to avoid the
second start, which saved ~30 minutes. However, we have to undo that
change in order to use beaker's reload functionality.

That said, we don't regress on test times, because reloading avoids both
puppetserver starts, and puts us back on par with test times prior to
the move to puppetserver.

This commit also moves the logic for setting `use-service` from an
explicit pre-suite step to the options hash, where all of the other
options are specified. Beaker converts the symbol to a string when
processing the options, so the existing behavior is preserved.

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -12,7 +12,7 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.1.0')
+gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.5.0')
 gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 0.3")
 gem "beaker-abs", *location_for(ENV['BEAKER_ABS_VERSION'] || "~> 0.2")
 gem "rake", "~> 10.1"
--- a/acceptance/config/aio/options.rb
+++ b/acceptance/config/aio/options.rb
@@ -1,17 +1,16 @@
 {
   :type => 'aio',
   :is_puppetserver => true,
+  :'use-service' => true,
   :puppetservice => 'puppetserver',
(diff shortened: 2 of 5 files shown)


joshcooper:
```
$ bundle exec bundle exec beaker-hostgenerator redhat7-64m-ubuntu1404-64a --disable-default-role > config.yml
$ bundle exec rake ci:test:aio SHA=1.8.1 SERVER_VERSION=2.7.1 BEAKER_HOSTS=config.yml
...
       Total Suite Time: 4777.34 seconds
      Average Test Time: 14.70 seconds
              Attempted: 325
                 Passed: 269
                 Failed: 0
                Errored: 0
                Skipped: 55
                Pending: 1
                  Total: 325

```

The most recent puppet-agent#stable run against all acceptance tests on the same platform combination took [`5978.59 seconds`](https://jenkins.puppetlabs.com/view/puppet-agent/view/stable/view/puppet-agent/job/platform_puppet-agent_intn-van-sys_suite-daily-puppet-stable/242/SLAVE_LABEL=beaker,TEST_TARGET=ubuntu1404-64a/), so this PR should eliminate about 20 minutes.

er0ck:
huge 👍 , i was just about to PR this.
any reason i shouldn't merge and merge-up?

MosesMendoza:
👍

PR #5405, opened 2016-12-02
(PUP-6850) Exclude commits unrelated to a PR

joshcooper:
TRAVIS_COMMIT_RANGE uses triple dots, which includes commits that are in
the base branch that are not in the PR. For example, if a commit is
pushed to master with an incorrect summary, then every PR created before
that commit will include the unrelated commit, causing the PR to fail
the summary check.

As suggested in https://github.com/travis-ci/travis-ci/issues/4596#issuecomment-139811122,
replace triple dots with double dots.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -84,7 +84,7 @@ task(:commits) do
   # falls back to `master..HEAD` as a next best bet as `master` is unlikely to
   # ever be absent.
   commit_range = ENV['TRAVIS_COMMIT_RANGE'].nil? ? 'master..HEAD' : ENV['TRAVIS_COMMIT_RANGE']
-  puts "Checking commits..."
+  puts "Checking commits #{commit_range}"
   %x{git log --no-merges --pretty=%s #{commit_range}}.each_line do |commit_summary|
     # This regex tests for the currently supported commit summary tokens: maint, doc, packaging, or pup-<number>.
     # The exception tries to explain it in more full.
--- a/Rakefile
+++ b/Rakefile
@@ -83,7 +83,7 @@ task(:commits) do
   # populated with the range of commits the PR contains. If not available, this
   # falls back to `master..HEAD` as a next best bet as `master` is unlikely to
   # ever be absent.
-  commit_range = ENV['TRAVIS_COMMIT_RANGE'].nil? ? 'master..HEAD' : ENV['TRAVIS_COMMIT_RANGE']
+  commit_range = ENV['TRAVIS_COMMIT_RANGE'].nil? ? 'master..HEAD' : ENV['TRAVIS_COMMIT_RANGE'].sub(/\.\.\./, '..')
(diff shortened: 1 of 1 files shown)


joshcooper:
@puppetlabs/puppet-maintainers for review

Rakefile
```diff
@@ -83,7 +83,7 @@ task(:commits) do
   # populated with the range of commits the PR contains. If not available, this
   # falls back to `master..HEAD` as a next best bet as `master` is unlikely to
   # ever be absent.
-  commit_range = ENV['TRAVIS_COMMIT_RANGE'].nil? ? 'master..HEAD' : ENV['TRAVIS_COMMIT_RANGE']
+  commit_range = ENV['TRAVIS_COMMIT_RANGE'].nil? ? 'master..HEAD' : ENV['TRAVIS_COMMIT_RANGE'].sub(/\.\.\./, '..')
   puts "Checking commits..."
```
Iristyle: Looks like this branch doesn't have the range output in the `puts` that was added to `stable` branch in #5365.

The output from TravisCI looks good at https://travis-ci.org/puppetlabs/puppet/jobs/180850536

```
Checking commits...
(PUP-6850) Exclude commits unrelated to a PR
...passed
```

Should we add it to this branch as part of this PR and deal with the merge conflict on the merge up?

joshcooper:
Good point @Iristyle. I pulled that change in:

```
$ env TRAVIS_COMMIT_RANGE=upstream/LTS-1.7...HEAD bundle exec rake commits
Checking commits upstream/LTS-1.7..HEAD
(PUP-6850) Exclude commits unrelated to a PR
...passed
(maint) Fix rake task that checks commits
...passed
```

joshcooper:
@MikaelSmith there's a travis-ci issue about the triple dots https://github.com/travis-ci/travis-ci/issues/4596#issuecomment-139811122. The behavior doesn't match the documentation, not sure why.

It's only been a problem now, because a commit was pushed directly to puppet with a broken subject, and for example, PR https://github.com/puppetlabs/puppet/pull/5415, was based off of stable, but the merge was targeted against master. So the triple-dot causes the commit check to run against all of the commits in the PR *and master* since their common ancestor 89b724dad18cb2cfd62be462f3b4ccb98d21d258.

Iristyle:
TravisCI now says `Checking commits 49cf1fdfc4f9f9128c6bfc48f8286fb66ec1914d..c6a75fc1419fc2ff4a4a85233ea2018bc3dab8b0`

👍

Iristyle:
Merged up to stable in https://github.com/puppetlabs/puppet/commit/ffabfc3d3019b1a4653c3015182a8996c8a86cde
Merged up to master in https://github.com/puppetlabs/puppet/commit/9dda0a424b5067cabdfd70f72b286fa7f29f604a

PR #5407, opened 2016-12-05
(PUP-6942) Let hiera.yaml be the name of the hiera version 5 configuration

thallgren:
This commit removes the previously added _lookup.yaml_ configuration file
in favor of always using the _hiera.yaml_. Instead of looking at different
files, all differences are instead triggered by the version number in the
_hiera.yaml_ file.


lib/puppet/pops/lookup/hiera_config.rb
```diff
@@ -71,25 +57,193 @@ def self.legacy_function_config(function_name)
...
+  def validate_config(config)
+    self.class.not_implemented(self, 'validate_config')
+  end
+
+  def version
+    self.class.not_implemented(self, 'version')
+  end
+
+  def name
+    "hiera version #{version}"
```
hlindberg: This should probably say 'hiera.yaml version' as opposed to "hiera version" as that may be confused with the hiera gem's version.

lib/puppet/pops/lookup/hiera_config.rb
```diff
@@ -71,25 +57,193 @@ def self.legacy_function_config(function_name)
...
+      raise Puppet::DataBinding::LookupError, "#{@config_path}: Name '#{name}' defined more than once" if data_providers.include?(name)
+      original_paths = he[KEY_PATHS] || [he[KEY_PATH] || name]
+      datadir = @config_root + (he[KEY_DATADIR] || default_datadir)
+      provider_name = he[KEY_BACKEND]
+      data_providers[name] = case provider_name
+      when 'json', 'yaml'
+        create_data_provider(name, parent_data_provider, KEY_DATA_HASH, "#{provider_name}_data", {},
+          resolve_paths(datadir, original_paths, lookup_invocation, ".#{provider_name}"))
+      else
+        # TODO: Remove support for injected provider factories
```
hlindberg: Is this for removal now, or are we supporting injected stuff still in puppet 4.9.0 ?

lib/puppet/pops/lookup/hiera_config.rb
```diff
@@ -240,9 +367,13 @@ def validate_defaults(defaults)
       # OK
     else
       raise Puppet::DataBinding::LookupError,
-            "#{@config_path}: Only one of #{combine_strings(FUNCTION_KEYS)} can defined in defaults"
+        "#{@config_path}: Only one of #{combine_strings(FUNCTION_KEYS)} can defined in defaults"
```
hlindberg: can *be* defined (word be is missing)

lib/puppet/pops/lookup/hiera_config.rb
thallgren: Or perhaps "hiera configuration version"? The fact that the configuration is in "yaml" format seems less relevant. What if we want to support other formats? Or get the config from a remote source (not sure if that will ever happen, but just to prove my point).

lib/puppet/pops/lookup/hiera_config.rb
thallgren: I'll remove that TODO. Not relevant until we decide to actually go ahead and break backward compatibility.

PR #5408, opened 2016-12-05
(PUP-6966) Allow file fifo and socket ensure absent

tkishel:
Prior to this commit:

- The `remove_existing` method didn't check the `self[:backup]` attribute before calling `can_backup` and `backup_existing`.
- The `remove_existing` method reported that it could not remove a file instead of reporting that it could not backup a file.
- The `can_backup` method returned `true` for all file types.

This caused File (and Tidy, which uses File) to fail to remove fifo and socket files and report a misleading error.

This commit allows File (and Tidy) to remove fifo and socket files.


joshcooper:
Thank you @tkishel. It looks like there are valid test failures, can you have a look.

joshcooper:
Accidentally merged to master, I reverted in c2df18b50d4f80600caec89d8644fa7419182d03. @tkishel to open a new PR

PR #5413, opened 2016-12-07
(PUP-6797) update smf tests for PUP-6796

MosesMendoza:
Prior to this PR, the puppet tests would check for puppet output indicating a previously nonexistent smf service by checking that it had transitioned from 'stopped' to 'running'. However, the fact that puppet would output `ensure => stopped` for non-existent services was in error, addressed as PUP-6797 - it should have been `ensure => absent` as it is with many other providers. This commit updates the various acceptance tests to instead look for the now-correct transition, which is the definition of ensure as running.

Second, prior to this PR, SMF testing was unnecessarily broken out into multiple files each with redundant setup and teardown steps, and in many cases actually testing the exact same thing. This PR merges these tests into the smf_basic_tests.rb file, with no impact to test value but cutting run time from ~30seconds to ~10seconds. Not a huge win, but this feels simpler/cleaner to me.


MosesMendoza:
ping @kurtwall @LuvCurves

acceptance/tests/resource/service/smf_basic_tests.rb
```diff
@@ -11,30 +11,35 @@
...
-
+  
   step "SMF: verify with svcs that the service is online"
   on agent, "svcs -l application/tstapp" do
     assert_match( /state\s+online/, result.stdout, "err: #{agent}")
   end
-  step "SMF: stop the service"
+  
+  step "SMF: ensure it is idempotent - ie not created again"
+  apply_manifest_on(agent, 'service {tstapp : ensure=>running, manifest=>"%s"}' % manifest, :catch_changes => true)
```
joshcooper: I think we lost the assertion that it doesn't change:

```
assert_no_match( /changed/, result.stdout, "err: #{agent}")
```

MosesMendoza:
successful run: https://gist.github.com/MosesMendoza/0b6da348cc20fa328ec0671a2b9deff9

acceptance/tests/resource/service/smf_basic_tests.rb
MosesMendoza: yeah - actually I did that on purpose, with switching to `catch_changes => true`, which I think accomplishes the same thing without the fragility of the grep. IMO ideally I think we'd update all of them to work off change detection/error codes but the best way to go about that with the others was less clear cut to me and I want to get this thing green again..

Magisus:
I definitely like the simplification into one file 👍

acceptance/tests/resource/service/smf_basic_tests.rb
joshcooper: Ah, I didn't know about `catch_changes`! 👍

kurtwall:
Thanks @MosesMendoza. We'll keep this in mind over the next few test cycles if there are new, interesting failures. :)

LuvCurves:
@MosesMendoza 💯  nice work. Should be easier to maintain.  Just need it in stable.

PR #5415, opened 2016-12-07
(maint) Use Windows line endings for batch file in integration test

glennsarti:
Previously the execution_spec.rb test was using LF line endings during the UTF8
test which worked on US English systems.  However when running on a Japanese
version of Windows, this test was failing incorrectly.  This commit forces the
line endings to Windows style CRLF so that the batch file will execute correctly.

### DIFF
--- a/spec/integration/util/execution_spec.rb
+++ b/spec/integration/util/execution_spec.rb
@@ -27,7 +27,7 @@ describe Puppet::Util::Execution do
                         131, 170, 227, 131, 134].pack('c*').force_encoding(Encoding::UTF_8)
     end
     let(:temputf8filename) do
-      script_containing(utf8text, :windows => "@ECHO OFF\nECHO #{utf8text}\nEXIT 100")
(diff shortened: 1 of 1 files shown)


joshcooper:
The travis failure is due to the typo'ed commit already in stable (note TRAVIS_COMMIT_RANGE comtains a triple dot). That will be fixed in https://github.com/puppetlabs/puppet/pull/5405.

joshcooper:
Interesting, I thought "\n" was a valid line ending for batch files on Windows (although it won't look right when using some editors). If Japanese Windows refuses to run batch files without "\r\n" then there could be many more places in puppet (and ecosystem) where things don't run correctly. I'm thinking about all of the puppet module files/templates, including PE.

joshcooper:
If you rebase against stable the travis commit check error should go away.

glennsarti:
The error popped up because I had UTF8 characters in the batch file.  It appears cmd.exe was barfing because it couldn't interpret where the end of the line was.

glennsarti:
Strange. The branch was already against stable (just pressed create PR too fast).  I'll rebase and see what happens.

glennsarti:
Rebased

PR #5416, opened 2016-12-08
(PUP-6986) Use binary encoding when searching for pids

Sharpie:
This patch updates the base service provider such that the binary
"ASCII-8BIT" string encoding is used when a process table scan is run by
service resources that have `hasstatus => false` set. This scan runs
regex matches against the output of `ps -ef` which could contain
arbitrary byte sequences depending on which services are running on the
system and how they are configured.

### DIFF
--- a/lib/puppet/provider/service/base.rb
+++ b/lib/puppet/provider/service/base.rb
@@ -32,15 +32,28 @@ Puppet::Type.type(:service).provide :base, :parent => :service do
     @resource.fail "Either stop/status commands or a pattern must be specified" unless @resource[:pattern]
     regex = Regexp.new(@resource[:pattern])
     ps = getps
+
     self.debug "Executing '#{ps}'"
-    IO.popen(ps) { |table|
-      table.each_line { |line|
-        if regex.match(line)
-          self.debug "Process matched: #{line}"
-          ary = line.sub(/^\s+/, '').split(/\s+/)
-          return ary[1]
-        end
-      }
+    table = Puppet::Util::Execution.execute(ps)
+
+    # The output of the PS command can be a mashup of several different
+    # encodings depending on which processes are running and what
+    # arbitrary data has been used to set their name in the process table.
+    #
+    # First, try a polite conversion to in order to match the UTF-8 encoding
+    # of our regular expression.
+    table = Puppet::Util::CharacterEncoding.convert_to_utf_8(table)
+    # If that fails, force to UTF-8 and then scrub as most uses are scanning
+    # for ACII-compatible program names.
+    table.force_encoding(Encoding::UTF_8) unless table.encoding == Encoding::UTF_8
+    table = Puppet::Util::CharacterEncoding.scrub(table) unless table.valid_encoding?
+
+    table.each_line { |line|
+      if regex.match(line)
+        self.debug "Process matched: #{line}"
+        ary = line.sub(/^[[:space:]]+/u, '').split(/[[:space:]]+/u)
+        return ary[1]
+      end
     }
 
     nil
--- a/spec/unit/provider/service/base_spec.rb
+++ b/spec/unit/provider/service/base_spec.rb
@@ -91,4 +91,24 @@ describe "base service provider" do
       expect {subject.stop }.to raise_error(Puppet::Error, 'Could not stop Service[test]: failed to stop')
     end
   end
+
+  context "when hasstatus is false" do
+    subject do
+      type.new(
+         :name  => "status test",
+         :provider => :base,
+         :hasstatus => false,
+         :pattern => "majestik m\u00f8\u00f8se",
+      ).provider
+    end
+
+    it "retrieves a PID from the process table" do
+      Facter.stubs(:value).with(:operatingsystem).returns("CentOS")
+      ps_output = File.binread(my_fixture("ps_ef.mixed_encoding")).force_encoding(Encoding::UTF_8)
+
+      executor.expects(:execute).with("ps -ef").returns(ps_output)
+
+      expect(subject.status).to eq(:running)
+    end
+  end
 end
--- a/spec/fixtures/unit/provider/service/base/ps_ef.mixed_encoding
+++ b/spec/fixtures/unit/provider/service/base/ps_ef.mixed_encoding
@@ -1,0 +1,3 @@
+UID        PID  PPID  C STIME TTY          TIME CMD
+root      4113 26549  0 11:07 pts/0    00:00:00 latin-1 args m�ni interesting furry animals
+root      4056 26549  0 11:06 pts/0    00:00:00 utf-8 args including the majestik møøse


lib/puppet/provider/service/base.rb
```diff
@@ -30,17 +30,23 @@ def getps
...
+    # Processes can add arbitrary sequences of bytes to the cmdline retrieved by
+    # 'ps -f', so drop encodings and match as binary.
+    regex = @resource[:pattern].dup
+    regex = Regexp.new(regex.force_encoding("ASCII-8BIT"))
+    table.force_encoding("ASCII-8BIT")
+
+    table.each_line { |line|
+      if regex.match(line)
+        self.debug "Process matched: #{line}"
+        ary = line.sub(/^\s+/, '').split(/\s+/)
```
Iristyle: I'm not sure what the `\s` character class means in a `BINARY` encoding - I'm a little surprised that it even works to be honest, but I'm assuming it's looking at things byte by byte?.  A whitespace ASCII character in Ruby is a `[ \t\r\n\f]` - i.e. any of the following bytes:

* 0x20 (space)
* 0x09 (tab \t)
* 0x0D (carriage return \r)
* 0x0A (line feed \n)
* 0x0C (form feed \f)

I'm going to see if I can craft some Unicode data, that when treated as a `BINARY` (i.e. `ASCII-8BIT`) string, will not split properly with the above code.  When there are multibyte Unicode characters that have any of the above bytes, then I presume this fails - but I will validate for certain.

Further, I'm not sure what we should ever be using `ASCII-8BIT` on strings that we later intend to manipulate in any way, for this reason.  Reading and writing files directly as `BINARY` is usually the right way to handle things I think, but when parsing values.. but when concatenating / regex searching / etc... I'm not sure that it's correct.

Sharpie: For `ASCII-8BIT` regex matches outside the ASCII range are done on a byte-by-byte basis. The test case in this patchset exercises this by matching the pattern `"majestic m\xd8\xd8se"` against a Latin-1 encoded file.

For the whitespace regexes, it looks like we can add a modifier to set them to binary mode: `/\s/n`:

http://ruby-doc.org/core-2.1.9/Regexp.html#class-Regexp-label-Encoding

lib/puppet/provider/service/base.rb
```diff
@@ -30,17 +30,23 @@ def getps
...
-          ary = line.sub(/^\s+/, '').split(/\s+/)
-          return ary[1]
-        end
-      }
+    table = Puppet::Util::Execution.execute(ps)
+
+    # Processes can add arbitrary sequences of bytes to the cmdline retrieved by
+    # 'ps -f', so drop encodings and match as binary.
+    regex = @resource[:pattern].dup
+    regex = Regexp.new(regex.force_encoding("ASCII-8BIT"))
```
Iristyle: Ok, found a way to make this misbehave:


```
# create a UTF-8 string
[44] pry(main)> string = "hello\u0904world!"
=> "helloऄworld!"

# force encoding to BINARY in a new copy
[45] pry(main)> string_bin = string.dup.force_encoding(Encoding::BINARY)
=> "hello\xE0\xA4\x84world!"

# we'll split on the Unicode bracket pattern [[:word:]] (Unicode version of \w)
[53] pry(main)> pattern = "[[:word:]]"
=> "[[:word:]]"
[54] pry(main)> pattern.encoding
=> #<Encoding:UTF-8>

# split our UTF-8 string with a UTF-8 regex
[57] pry(main)> string.split(Regexp.new(pattern))
=> ["", "", "", "", "", "", "", "", "", "", "", "!"]

# split our BINARY string with a BINARY regex
[58] pry(main)> string_bin.split(Regexp.new(pattern.dup.force_encoding(Encoding::BINARY)))
=> ["", "", "", "", "", "\xE0\xA4\x84", "", "", "", "", "!"]
```

Iristyle:
One other general comment I'll leave here is this.  We are mostly trying to use `UTF-8` as the default string encoding internally (so for all Puppet controlled files like manifests, templates, config files, over-the-wire formats, etc), except under a few circumstances:

* When we're capturing output from shell executables, our current thinking is that it should be using the `Encoding.default_external` (or maybe capturing as `BINARY`, encoded to `Encoding.default_external` by default internally, and optionally user configurable through `file` / `exec` / etc resources) - some of this effort is captured in [PUP-6432](https://tickets.puppetlabs.com/browse/PUP-6432)
* `Encoding.default_external` when reading / writing files that are part of the OS - this is really a best guess, because file encoding is probably application specific.  The hope is that this is mostly `UTF-8` data, and that users are running their machine with a `UTF-8` `LANG`, but I don't think we can ever be 100% 
* `BINARY` when reading / writing files, like when we copy to the file bucket.  When there's no manipulation, and we want to treat files byte for byte.

The interesting thing of course is that anything that gets generated as output - particularly logs / reports, has to end up as `UTF-8` in the end... so there's that.

Sharpie:
> One other general comment I'll leave here is this. We are mostly trying to use UTF-8 as the default string encoding internally (so for all Puppet controlled files like manifests, templates, config files, over-the-wire formats, etc), except under a few circumstances:

The problem here is that the encoding decisions of the `ps` output are made externally by different programs. One could decide to put Latin-1 characters into it's title and another could add raw binary, so it could end up being a mish-mash of different data.

I suppose the alternative would be to forcibly encode the output to UTF-8 and strip unknown bytes or replace them with the `\ufffd` "replacement character". I suppose this wouldn't result in a reduction in functionality as the current restrictions on manifests don't make it possible to specify a non-UTF-8 pattern.

Would that be a more acceptable approach?

Iristyle:
At least theoretically speaking, anything expressed in another encoding (other than `BINARY`) can be converted to UTF-8 since UTF-8 is a superset of all characters. In practical terms, I haven't exhaustively tested this - so not sure how robust Ruby's `iconv` support is - I'm sure there are probably times where it fails?  Sometimes there are also choices to make in terms of what UTF-8 codepoint corresponds to the originating character.

But I think this is generally the approach to take - so instead of `force_encoding` to modify just the metadata, actually perform a conversion by calling `.encode(Encoding::UTF_8)` against a string that is presumed to be in `Encoding.default_external`.  Then you can use a UTF-8 regex against that string.

Iristyle:
BTW - Ruby should use `UndefinedConversionError` when it can't convert to UTF-8 and I know I've seen this before - but the only tickets I see are when going from `ASCII-8BIT` / `BINARY` -> `UTF_8` which is understandable given the raw bytes are probably characters in another encoding, and without knowing what that encoding is, there's no sensible way to convert to UTF-8 if the byte sequences are invalid UTF-8...

See https://tickets.puppetlabs.com/browse/PUP-1179?jql=text%20~%20%22UndefinedConversionError%22 for a few of those

Going from UTF-8 back to something else makes a lot more sense for `UndefinedConversionError`

Sharpie:
PR updated to sanitize `ps` output by replacing invalid byte sequences with U+FFFD characters so that the result is in a valid UTF-8 encoding instead of dropping encodings and using binary matching.

Iristyle:
FYI @Sharpie I took a look at where `ps -ef` is called from, and it looks like it may already be excluded from OSes without `procfs` that you were were concerned about.  I may take a quick look into how hard it might be to use Ruby to read `proc` and if that can work around the encoding issues we're running into here - i.e. https://fredrb.github.io/2016/10/01/Understanding-proc/ (though at least in the pass I know that Ruby reading `/proc` was unreliable as there's an old Facter ticket about it on Redmine)

Iristyle:
So I think my `/proc/#{pid}/cmdline` theory works.

* I grabbed Ubuntu 16.10 out of vmpooler
* Created a file on disk called `Aۿᚠ𠜎` - this is a UTF-8 string we're using in tests that has a 1-byte, 2-byte, 3-byte and 4-byte UTF-8 character in it
* Got Ruby devkit installed with `apt-get install -y ruby-full`
* Installed pry + friends `gem install pry pry-byebug pry-stack_explorer`
* Launched another bash instance, and jammed the filename into the invocation via `--rcfile`

```
root     14550  1573  0 15:25 pts/0    00:00:00 bash --rcfile /root/Aۿᚠ𠜎 -i
root     14554 14550  0 15:25 pts/0    00:00:00 ps -ef
``` 

* Fire up pry, noting that Ruby currently has `Encoding.default_external` equivalent to UTF-8
* Make sure we can read the `cmdline` as it stands:

```
[5] pry(main)> cmdline = File.read('/proc/14550/cmdline')
=> "bash\u0000--rcfile\u0000/root/Aۿᚠ𠜎\u0000-i\u0000"
```

* Now change Encoding to something that will mangle the data, in this case `ISO_8859_1`

```
[8] pry(main)> Encoding.default_external = Encoding::ISO_8859_1
=> #<Encoding:ISO-8859-1>
```

* Inspect `cmdline` variable in pry, and notice that it's now displaying Unicode codepoints instead of inline characters:

```
[13] pry(main)> cmdline
=> "bash\u0000--rcfile\u0000/root/A\u06FF\u16A0\u{2070E}\u0000-i\u0000"
```

* Try to re-read `/proc/14550/cmdline` with the ISO-8859-1 encoding - as expected it's hosed:

```
[15] pry(main)> cmdline2 = File.read('/proc/14550/cmdline')
=> "bash\x00--rcfile\x00/root/AÛ¿á\x9A ð \x9C\x8E\x00-i\x00"
```

* There are 2 ways to get the right UTF-8 values at this point - directly read with an `:encoding` specified like:

```
[16] pry(main)> cmdline3 = File.read('/proc/14550/cmdline', :encoding => Encoding::UTF_8)
=> "bash\u0000--rcfile\u0000/root/A\u06FF\u16A0\u{2070E}\u0000-i\u0000"
```

* Or just tell Ruby that the bytes aren't really in `ISO-8859-1`, but rather `UTF-8`:

```
[17] pry(main)> cmdline2.force_encoding(Encoding::UTF_8)
=> "bash\u0000--rcfile\u0000/root/A\u06FF\u16A0\u{2070E}\u0000-i\u0000"
```

NOTE: This is not really an exhaustive test, because I did not test what happens when the bytes of a file name are written with a different encoding.  EXT4 doesn't specify an encoding for names - it just uses sequences of bytes, so really they should be expected to be in `Encoding.default_external`.  

I agree with your general sentiment that there may be times where command line params are passed to processes or files are written, where we can't be 100% certain of the encoding (and we should take extra steps to handle conversion issues where it makes sense / log those).  I still think we should work everything toward `UTF-8` - files managed with Puppet are specified in manifests with `UTF-8` names for instance, so that's how we should always be reading / writing against the file system.

MosesMendoza:
Since our manifest data (pattern) is coming in as UTF-8, I think we should attempt to convert the process data obtained by the system (`is` value) to UTF-8 and then perform the comparison.

https://github.com/puppetlabs/puppet/pull/5633 (PUP-7021) established a new `Puppet::Util::CharacterEncoding` module with a new standard UTF-8 conversion heuristic for puppet `::convert_to_utf_8`, which I think we can/should use here. https://github.com/puppetlabs/puppet/pull/5634 and https://github.com/puppetlabs/puppet/pull/5635 are also open to adopt use of this method.

MosesMendoza:
@Sharpie if you're not free ATM to continue developing on this PR let me know and I can pull it in and continue adding to it. It's in our current sprint.

Sharpie:
@MosesMendoza The problem here is that the system call is returning command lines from multiple processes. Most are probably using arguments a UTF-8 compatible encoding. Some might have raw bitstrings or Latin-1 in their arguments. All of this will be co-mingled into a single output string. In the situation where this bug was observed, one or more processes appeared to be updating their command lines with log messages that contained raw binary.

Because of this, we can't expect the output to be cleanly convertible to a single encoding like UTF-8. Asking users to handle processes with non-UTF8 arguments doesn't sound feasible either as those arguments may be required or the process might be owned by a different team which doesn't want to change it.

I think there are three approaches we can take in this situation:

  - Do a lossy encoding of the output to UTF-8 where replacement characters are substituted for incompatible bytes sequences. The problem that failed the last round of tests is that Ruby 1.9 and 2.0 consider `.encode!` to be a no-op if the string is already has its encoding set to "UTF-8". This option would preserve the ability to match process names in mixed output using UTF-8 regexes, which is probably the majority usecase. On SLES 11, the `ps` command its self this sort of normalization using replacement characters when `LC_CTYPE` was set to a UTF-8 locale.

  - If we can count on `ps` performing normalization on all POSIX platforms when `LC_CTYPE` is set, we could the environment to a UTF-8 locale when calling `ps` to get the process table.

  - Strip the encoding of the `ps` output down to raw binary and perform matches using binary regexes. This would also preserve the ability to match using UTF-8 strings, and any string once Puppet supports a Binary type, but the behavior of escape characters in a binary regex may be surprising.

Sharpie:
PR updated address Ruby 1.9 and 2.0 failures and to use the `String#scrub` method introduced in Ruby 2.1 when possible. Not sure what is up with the Appveyor build as the errors seem unrelated.

MosesMendoza:
> The problem here is that the system call is returning command lines from multiple processes. Most are probably using arguments with a UTF-8 compatible encoding. Some might have raw bitstrings or Latin-1 in their arguments. All of this will be co-mingled into a single output string. In the situation where this bug was observed, one or more processes appeared to be updating their command lines with log messages that contained raw binary.

Agreed a single string composed of bytes representing different encodings won't convert cleanly. What I was thinking was where we iterate over each line in `table`, we could convert that line as UTF-8 before matching against the regex. In that way we handle this case, I think. I.e.,

```ruby
table.each_line { |line|
  line = Puppet::Util::CharacterEncoding.convert_to_utf_8(line)
  if regex.match(line)
  ...
```

Side benefit would be failure in conversion there would provide a helpful message pointing to the exact process with characters that weren't convertible.

Sharpie:
> What I was thinking was where we iterate over each line in table, we could convert that line as UTF-8 before matching against the regex.

That still leaves the problem that the process causing the failure might not be the one that the user is interested in managing. So, they'll be blocked from managing a service with UTF-8 clean args until an encoding issue with another unrelated service is sorted out --- which may not be feasible.

MosesMendoza:
I think it's going to be a fairly rare case where `Puppet::Util::CharacterEncoding#transcode_to_utf_8` will actually raise an exception/fail (could you try it against the original output that led to this bug?)

But if that's a major concern, what about an augmented solution would handle that case a warning that the given process is unmanageable by puppet? ie:
```ruby
table.each_line { |line|
  begin
    line = Puppet::Util::CharacterEncoding.convert_to_utf_8(line)
  rescue Puppet::Error
    Puppet.debug("Could not manage process '#{line}' - cannot convert process name to UTF-8")
    next
  end
  if regex.match(line)
  ...
```
I think it's ok to print that in debug, as we already do for matched processes:
```
Debug: Service[foo](provider=base): Process matched: root      36570  35500 66 10:43 pts/0    00:00:02 /usr/bin/ruby2.3 /usr/local/bin/puppet apply -e  service { 'foo' : ensure => 'running', pattern => /my foo/, provider => 'base' }  --debug --trace
```

Sharpie:
Using `Puppet::Util::CharacterEncoding#transcode_to_utf_8` fails the test case included with this PR as a process with a non-UTF8 command line precedes the process with the command line that the regex is looking for.

> what about an augmented solution would handle that case a warning that the given process is unmanageable by puppet?

The problem here is that if we start skipping entries, then a regex looking for `/opt/bin/foo` would skip an entry containing `/opt/bin/foo <non-utf8 content>`, conclude the service isn't running and attempt to start a new copy --- which breaks idempotency and would lead to resource exhaustion if multiple copies of a service are spawned.

MosesMendoza:
> The problem here is that if we start skipping entries, then a regex looking for /opt/bin/foo would skip an entry containing /opt/bin/foo <non-utf8 content>, conclude the service isn't running and attempt to start a new copy --- which breaks idempotency and would lead to resource exhaustion if multiple copies of a service are spawned.

Ah of course. That makes sense.

joshcooper:
We resolved some issues with `Puppet::Util::CharacterEncoding` in the lead up to Puppet 5. @Sharpie can you rebase on 4.10.x and see if that resolves the appveyor CI issues?

Sharpie:
PR updated to use `Puppet::Util::CharacterEncoding` and fix issues with Windows tests. It doesn't look like the `base` service provider even supports Windows, but that's no excuse to break the test suite :)

Sharpie:
@joshcooper @MosesMendoza Any other changes needed for this one?

spec/unit/provider/service/base_spec.rb
```diff
@@ -91,4 +91,24 @@ def execute_command(command, options)
...
+         :name  => "status test",
+         :provider => :base,
+         :hasstatus => false,
+         :pattern => "majestik m\u00f8\u00f8se",
+      ).provider
+    end
+
+    it "retrieves a PID from the process table" do
+      Facter.stubs(:value).with(:operatingsystem).returns("CentOS")
+      ps_output = File.binread(my_fixture("ps_ef.mixed_encoding")).force_encoding(Encoding::UTF_8)
```
Iristyle: The flip side of this is what happens when the output contains valid UTF-8, but the string has been marked as something else (based on a users `LANG` + friends setup which has informed `Encoding.default_external`)?

lib/puppet/provider/service/base.rb
```diff
@@ -32,15 +32,28 @@ def getpid
...
+    # encodings depending on which processes are running and what
+    # arbitrary data has been used to set their name in the process table.
+    #
+    # First, try a polite conversion to in order to match the UTF-8 encoding
+    # of our regular expression.
+    table = Puppet::Util::CharacterEncoding.convert_to_utf_8(table)
+    # If that fails, force to UTF-8 and then scrub as most uses are scanning
+    # for ACII-compatible program names.
+    table.force_encoding(Encoding::UTF_8) unless table.encoding == Encoding::UTF_8
+    table = Puppet::Util::CharacterEncoding.scrub(table) unless table.valid_encoding?
```
Iristyle: Have you considered doing this per line vs doing the whole block of text at once? Trying to think if it makes any difference or not - i.e. could you / should you avoid the `scrub` by encoding the regular expression to the same encoding as the source string. (`scrub` is probably the way to go since we really want UTF-8 everywhere internally)

Sharpie: I don't think that will work. The problem that this patch is trying to solve is that the output of ps can be in a **mixed encoding** as each process running on the system may have decided to write arbitrary bytes into their individual /proc/cmdline entries. In the case of the ticket that sparked this pull request it looked like raw binary data was being dumped into the command line --- not even something that had a meaningful text encoding.

Given that Puppet manifests can only express UTF-8 regexes at the moment, scrubbing the entire output block to UTF-8 after trying a conversion from Encoding.default_external seems like the best way to go.

spec/unit/provider/service/base_spec.rb
Sharpie: The output will end up getting mangled. The text fixture was generated on a machine that had a locale set to UTF_8, so this is re-producing that setup in cases when the tests are run under a different locale.

Given the output of `ps` can be in a mixed encoding, my real goal with this PR is to keep regexes containing ASCII characters working instead of throwing errors --- as I suspect that's the majority use case for folks that are using this provider.

lib/puppet/provider/service/base.rb
```diff
@@ -32,15 +32,28 @@ def getpid
...
+
+    table.each_line { |line|
+      if regex.match(line)
+        self.debug "Process matched: #{line}"
+        ary = line.sub(/^\s+/u, '').split(/\s+/u)
```
Iristyle: While you're here, what do you think about changing the RegEx to be Unicode compatible - i.e. `\s` -> `[[:space:]]`?

Sharpie: Seems reasonable. Added.

Magisus:
This PR has been updated according to the suggestions of @MosesMendoza and @Iristyle, do you guys have any additional feedback, or can we get this merged?

adrienthebo:
@Sharpie should this still land in 4.10.x or can we land it in master?

Sharpie:
@adrienthebo Since it is a bugfix, it should land in 4.10.x and merge up --- unless there's a compelling reason to do it the other way.

adrienthebo:
@MosesMendoza it looks like you're still requesting changes on this PR, is that request still active?

MosesMendoza:
@adrienthebo Nope this looks good to me now. sorry for forgetting that.

MosesMendoza:
thanks @Sharpie

Iristyle:
Thanks for getting this one in @MosesMendoza ... sorry it took so long to get back on the radar and merged in @Sharpie.  Appreciate the original PR!

PR #5419, opened 2016-12-10
(PUP-6777) (PUP-7031) address utf-8 related comment bugs in user providers

MosesMendoza:
The nameservice provider leverages the ruby Etc::getpwnam method to return
information regarding users (subclasses include useradd and pw). In Ruby <
2.1.0, this method always returns values as strings in binary encoding. In Ruby
2.1.0 and later, this method returns strings in either the current locale
encoding (Encoding::default_external) OR binary encoding if the string returned
does not encode correctly in the current locale encoding (how they determine
this is opaque, at best, and not exposed publicly).

Prior to this commit, the :comment property of the user type would set the
external encoding of the supplied value for :comment to binary if running ruby
< 2.1.0. This effectively handles the case of incompatible encodings with ruby
< 2.1.0. However, for ruby >= 2.1.0, this will always fail in two scenarios:

- if the value of :comment contains character(s) that are not represented in
  the current locale, because ruby Etc will supply this 'is' value in binary and
  we will compare it to our 'should' value in whatever the current locale is,
  failing.

- if the current locale results in a system (is) value of :comment in a ruby
  Encoding incompatible with the formatting of the desired 'should' value from
  Puppet.

This commit instead overrides the Puppet::Property#insync? method for
Puppet::Type::User::Comment when using the nameservice provider, using
`force_encoding` to set the External encoding of the 'should' value to that of
the system 'is' value. This commit also overrides Puppet::Property#change_to_s
for Puppet::Type::User::Comment for scenarios where we are about to concatenate
strings with incompatible character encodings. In this case as well, set the
encoding of the new value to the encoding of the value returned by the system
for printing a message to the user. In both cases, we avoid manipulating the
external encoding of the underlying 'should' object, for fear of unexpected
side-effects in Puppet. One such side-effect was found in
Puppet::Util::Yaml::Dump, which will fail to write the run report if the
should value cannot actually be represented by it's assigned default external
encoding (i.e., if `valid_encoding?` would return false).

This PR also addresses PUP-7031.
Reported in PUP-7031, it was discovered that we were not appropriately
round-tripping UTF-8 values for the Description user field, using `IADs::Get`
and `IADs::Put`. Bytes for values returned by `Get` would be questions marks
('?', 63).

We are retrieving the values via the Ruby WIN32OLE COM interface. The values
are stored as BSTRs, a string data type used by COM encoded as UTF-16LE.

The Ruby WIN32OLE interface is initialized with a codepage - initially
determined by Encoding.default_internal, falling back to
Encoding.default_external. On US English Windows systems the default codepage
(and Encoding.default_external) is normally IBM 437, meaning WIN32OLE is
normally initialized with a codepage of IBM 437. Thus, in this default state,
WIN32OLE will return strings in IBM 437 encoding. The conversion from UTF-16LE
to IBM 437 is a lossy one if the string contains unicode-only characters -
hence the behavior we are seeing of values replaced by question marks.

This PR modifies takes the additional step of immediately setting the
WIN32OLE codepage to UTF-8 after we require it - specifically,
WIN32OLE::CP_UTF8 (codepage 65001). Thus, values returned are encoded by Ruby
as UTF-8.

This commit improves on f17c853 (PUP-6499) which encoded strings as UTF-8 after
they were returned, which was after the conversion from UTF16-LE to IBM437 had
already taken place.

### DIFF
--- a/lib/puppet/feature/base.rb
+++ b/lib/puppet/feature/base.rb
@@ -17,8 +17,15 @@ end
 Puppet.features.add(:microsoft_windows) do
   begin
     # ruby
-    require 'Win32API'          # case matters in this require!
-    require 'win32ole'
+    require 'Win32API' # case matters in this require!
+
+    # Note: Setting codepage here globally ensures all strings returned via
+    # WIN32OLE (Ruby's late-bound COM support) are encoded in Encoding::UTF_8
+    #
+    # Also, this does not modify the value of WIN32OLE.locale - which defaults
+    # to 2048 (at least on US English Windows) and is not listed in the MS
+    # locales table, here: https://msdn.microsoft.com/en-us/library/ms912047(v=winembedded.10).aspx
+    require 'win32ole' ; WIN32OLE.codepage = WIN32OLE::CP_UTF8
     # gems
     require 'win32/process'
     require 'win32/dir'
--- a/lib/puppet/provider/nameservice.rb
+++ b/lib/puppet/provider/nameservice.rb
@@ -288,5 +288,20 @@ class Puppet::Provider::NameService < Puppet::Provider
       raise Puppet::Error, "Could not set #{param} on #{@resource.class.name}[#{@resource.name}]: #{detail}", detail.backtrace
     end
   end
+
+  # From overriding Puppet::Property#insync? Ruby Etc::getpwnam < 2.1.0 always
+  # returns a struct with binary encoded string values, and >= 2.1.0 will return
+  # binary encoded strings for values incompatible with current locale charset,
+  # or Encoding.default_external if compatible. Compare a "should" value with
+  # encoding of "current" value, to avoid unnecessary property syncs and
+  # comparison of strings with different encodings. (PUP-6777)
+  #
+  # return basic string comparison after re-encoding (same as
+  # Puppet::Property#property_matches)
+  def comments_insync?(current, should)
+    # we're only doing comparison here so don't mutate the string
+    desired = should[0].to_s.dup
+    current == desired.force_encoding(current.encoding)
+  end
 end
 
--- a/lib/puppet/type/user.rb
+++ b/lib/puppet/type/user.rb
@@ -171,10 +171,23 @@ module Puppet
 
     newproperty(:comment) do
       desc "A description of the user.  Generally the user's full name."
-      if RUBY_VERSION < "2.1.0"
-        munge do |v|
-          v.respond_to?(:force_encoding) ? v.force_encoding(Encoding::ASCII_8BIT) : v
+      def insync?(is)
+        # nameservice provider requires special attention to encoding
+        # Overrides Puppet::Property#insync?
+        if !@should.empty? && provider.respond_to?(:comments_insync?)
+          return provider.comments_insync?(is, @should)
         end
+        super(is)
+      end
+
+      # In the case that our comments have incompatible encodings, set external
+      # encoding to support concatenation for display.
+      # overrides Puppet::Property#change_to_s
+      def change_to_s(currentvalue, newvalue)
+        if newvalue.respond_to?(:force_encoding) && !Encoding.compatible?(currentvalue, newvalue)
+          return super(currentvalue, newvalue.dup.force_encoding(currentvalue.encoding))
+        end
+        super(currentvalue, newvalue)
       end
     end
 
--- a/spec/unit/provider/nameservice_spec.rb
+++ b/spec/unit/provider/nameservice_spec.rb
@@ -301,4 +301,46 @@ describe Puppet::Provider::NameService do
     end
   end
 
+  describe "comments_insync?" do
+    # comments_insync? overrides Puppet::Property#insync? and will act on an
+    # array containing a should value (the expected value of Puppet::Property
+    # @should)
+    context "given strings with compatible encodings" do
+      it "should return false if the is-value and should-value are not equal" do
+        is_value = "foo"
+        should_value = ["bar"]
+        expect(provider.comments_insync?(is_value, should_value)).to be_falsey
+      end
+
+      it "should return true if the is-value and should-value are equal" do
+        is_value = "foo"
+        should_value = ["foo"]
+        expect(provider.comments_insync?(is_value, should_value)).to be_truthy
+      end
+    end
+
+    context "given strings with incompatible encodings" do
+      let(:snowman_iso) { "\u2603".force_encoding(Encoding::ISO_8859_1) }
+      let(:snowman_utf8) { "\u2603".force_encoding(Encoding::UTF_8) }
+      let(:snowman_binary) { "\u2603".force_encoding(Encoding::ASCII_8BIT) }
+      let(:arabic_heh_utf8) { "\u06FF".force_encoding(Encoding::UTF_8) }
+
+      it "should be able to compare unequal strings and return false" do
+        expect(Encoding.compatible?(snowman_iso, arabic_heh_utf8)).to be_falsey
+        expect(provider.comments_insync?(snowman_iso, [arabic_heh_utf8])).to be_falsey
+      end
+
+      it "should be able to compare equal strings and return true" do
+        expect(Encoding.compatible?(snowman_binary, snowman_utf8)).to be_falsey
+        expect(provider.comments_insync?(snowman_binary, [snowman_utf8])).to be_truthy
+      end
+
+      it "should not manipulate the actual encoding of either string" do
+        expect(Encoding.compatible?(snowman_binary, snowman_utf8)).to be_falsey
+        provider.comments_insync?(snowman_binary, [snowman_utf8])
+        expect(snowman_binary.encoding).to eq(Encoding::ASCII_8BIT)
+        expect(snowman_utf8.encoding).to eq(Encoding::UTF_8)
+      end
+    end
+  end
 end
--- a/spec/unit/type/user_spec.rb
+++ b/spec/unit/type/user_spec.rb
@@ -351,13 +351,39 @@ describe Puppet::Type.type(:user) do
       @user = described_class.new(:name => 'foo', :comment => @value)
     end
 
-    it "should be converted to ASCII_8BIT for ruby 1.9 / 2.0", :if => RUBY_VERSION < "2.1.0" && String.method_defined?(:encode) do
-      expect(@user[:comment].encoding).to eq(Encoding::ASCII_8BIT)
-      expect(@user[:comment]).to eq(@value.force_encoding(Encoding::ASCII_8BIT))
-    end
+    describe "#insync" do
+      it "should delegate to the provider's #comments_insync? method if defined" do
+        # useradd subclasses nameservice and thus inherits #comments_insync?
+        user = described_class.new(:name => 'foo', :comment => @value, :provider => :useradd)
+        comment_property = user.properties.find {|p| p.name == :comment}
+        user.provider.expects(:comments_insync?)
+        comment_property.insync?('bar')
+      end
 
-    it "must not be converted for ruby >= 2.1", :if => RUBY_VERSION >= "2.1.0" do
-      expect(@user[:comment].encoding).to eq(Encoding::UTF_8)
+      describe "#change_to_s" do
+        let(:is) { "\u2603" }
+        let(:should) { "\u06FF" }
+        let(:comment_property) { @user.properties.find { |p| p.name == :comment } }
+        context "given is and should strings with incompatible encoding" do
+          it "should return a formatted string" do
+            is.force_encoding(Encoding::ASCII_8BIT)
+            should.force_encoding(Encoding::UTF_8)
+            expect(Encoding.compatible?(is, should)).to be_falsey
+            # append Regexp with 'n' to set encoding to ASCII_8BIT
+            expect(comment_property.change_to_s(is,should)).to match(/changed '\xE2\x98\x83' to '\xDB\xBF'/n)
+          end
+        end
+
+        context "given is and should strings with compatible encoding" do
+          it "should return a formatted string" do
+            is.force_encoding(Encoding::UTF_8)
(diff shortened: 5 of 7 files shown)


spec/unit/type/user_spec.rb
```diff
@@ -351,13 +351,20 @@ def self.instances; []; end
...
     end
 
-    it "should be converted to ASCII_8BIT for ruby 1.9 / 2.0", :if => RUBY_VERSION < "2.1.0" && String.method_defined?(:encode) do
-      expect(@user[:comment].encoding).to eq(Encoding::ASCII_8BIT)
-      expect(@user[:comment]).to eq(@value.force_encoding(Encoding::ASCII_8BIT))
-    end
-
-    it "must not be converted for ruby >= 2.1", :if => RUBY_VERSION >= "2.1.0" do
-      expect(@user[:comment].encoding).to eq(Encoding::UTF_8)
+    it "should define the #insync?(is,should) method" do
```
MosesMendoza: TBH I'm kind of skeptical of these unit tests and think I'll be adding acceptance tests to really validate the behavior here.

lib/puppet/provider/nameservice.rb
```diff
@@ -288,5 +288,25 @@ def set(param, value)
...
+  # * "should" parameter will always be a string with one possible (strictly
+  # speaking) exception, which is the boolean value true. In current semantics,
+  # this would convert to a string. For strict backwards-compatibility, convert
+  # it to a string if this is the case.
+  #
+  # return basic string comparison after encoding (same as
+  # Puppet::Property#property_matches)
+  def comments_insync?(current, should)
+    should = should.to_s if should == true
+    should.force_encoding(Encoding::ASCII_8BIT) if current.encoding == Encoding::ASCII_8BIT
```
MosesMendoza: I'm confident given the constraints around this provider that we really only need to handle the case where we're dealing with a binary-encoded 'is', but another possibility here would be to do `should.force_encoding(current.encoding)` - feels riskier though but might be more..thorough.

spec/unit/type/user_spec.rb
```diff
@@ -351,13 +351,20 @@ def self.instances; []; end
...
+      comment_property = @user.properties.find {|p| p.name == :comment}
+      # class#instance_methods(false) returns methods only defined on the instance and no ancestors
+      expect(comment_property.class.instance_methods(false).include?(:insync?)).to be_truthy
+    end
+
+    describe "#insync" do
+      it "should delegate to the provider's #comments_insync? method if defined" do
```
MosesMendoza: should probably also have a case where `#comments_insync?` is not called but instead `super`, but it seems odd to just check that `super` is called.

MosesMendoza:
need to address spec issues

spec/unit/provider/nameservice_spec.rb
```diff
@@ -301,4 +301,42 @@
...
+    end
+
+    it "should return false if the is-value and should-value are not equal" do
+      is_value = "foo"
+      should_value = "bar"
+      expect(provider.comments_insync?(is_value, should_value)).to be_falsey
+    end
+
+    it "should return true if only the is-value is in binary encoding but the strings are otherwise identical" do
+      is_value = String.new("foo", :encoding => Encoding::ASCII_8BIT)
```
MosesMendoza: Apparently this form of the string constructor didn't exist until ruby 2.3.0 - updating

lib/puppet/provider/nameservice.rb
McdonaldSeanp: IMO I would tend to agree with the way that this is implemented, personally I think I agree that `force_encoding(current.encoding)` is riskier, and I would avoid the risk.

spec/unit/provider/nameservice_spec.rb
```diff
@@ -301,4 +301,42 @@
...
+    end
+
+    it "should do basic string comparison of the is-value with the should-value" do
+      is_value = "foo"
+      should_value = "bar"
+      is_value.expects(:'==').with("bar")
+      provider.comments_insync?(is_value, should_value)
+    end
+
+    it "should return false if the is-value and should-value are not equal" do
```
McdonaldSeanp: Should there be a test for returning false for different strings with different encodings?

spec/unit/type/user_spec.rb
McdonaldSeanp: I would add this case in, I don't think there is anything wrong with checking for super based on specific parameters

McdonaldSeanp:
@MosesMendoza This looks good. From the testing perspective I had a couple questions, but the implementation seems sound.

lib/puppet/provider/nameservice.rb
Iristyle: So your question is an interesting one @MosesMendoza.  When encodings mismatch, I think we have 2 potential ways we can fail:

* `current == should` explodes on the mismatched encodings
* On the attempt to match the encodings, a conversion error is raised at the `force_encoding` (let's say this is some weird case where Ruby doesn't give us a binary string - which you seem to indicate isn't possible)

Which propagates a more useful error message?

lib/puppet/provider/nameservice.rb
```diff
@@ -288,5 +288,25 @@ def set(param, value)
...
+  #
+  # * "should" parameter will always be a string with one possible (strictly
+  # speaking) exception, which is the boolean value true. In current semantics,
+  # this would convert to a string. For strict backwards-compatibility, convert
+  # it to a string if this is the case.
+  #
+  # return basic string comparison after encoding (same as
+  # Puppet::Property#property_matches)
+  def comments_insync?(current, should)
+    should = should.to_s if should == true
```
Iristyle: I don't quite understand when `should` is `true` - what are the circumstances?

Should we always just call `.to_s` on all incoming values in the event that there's some other type?  `to_s` on a string is roughly a no-op anyway.

MosesMendoza: yeah - the to_s is almost always going to be unnecessary (unless we go with `@should` as you suggest below). The only time this will be the case is if the user has provided the literal value `true` as the value to the `comment` property, ie
```ruby
user { 'foo' : comment => true }
```
before this change, that `true` would be cast to a string, so I wanted to ensure that stays the same way.

lib/puppet/type/user.rb
```diff
@@ -171,10 +171,15 @@ def sync
...
     newproperty(:comment) do
       desc "A description of the user.  Generally the user's full name."
-      if RUBY_VERSION < "2.1.0"
-        munge do |v|
-          v.respond_to?(:force_encoding) ? v.force_encoding(Encoding::ASCII_8BIT) : v
+      def insync?(current)
+        # nameservice provider requires special attention to encoding
+        # Use Puppet::Property#should instead of @should to avoid extra
+        # array handling logic
+        if !@should.empty? and provider.respond_to?(:comments_insync?)
```
MosesMendoza: I agree most implementations just pass `@should` value. I was certainly on the fence about that - not sure which way was better. Happy to update to handle accepting `@should`. In that case, I believe the value we're going to get passed in will almost always be an array containing a single string representing the comment value.

MosesMendoza: Yeah, probably best not to rely on implementation of `Puppet::Property#should`

lib/puppet/provider/nameservice.rb
MosesMendoza: Complicating this, I came across another case when running this through a new acceptance test, which is not covered by this code as-is:

When using `puppet apply` our `is` value, when not binary, will be in `Encoding.default_external`. In the case of our Ubuntu 16.04 acceptance hosts, that is `ISO-8859-1`. Perhaps this lends support to trying to set the encoding of `should` to `is` for comparison. In terms of error messaging, we could possibly rescue `EncodingError` with a more helpful error message.

lib/puppet/type/user.rb
MosesMendoza: Great question about whether we want the `munged` vs raw value (or even also `unmunged`). Turns out we call various munging and unmunging methods throughout the lifecycle of the property, but at the time of comparing is/should values for synced-ness, they are:
- is-value: a value that has been unmunged by the `unmunge` method defined, if any, on the corresponding _option_ in the provider ([example](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/user/useradd.rb#L35-L42)). This is true at least for providers inheriting from nameservice.
- should-value: a value that has been munged by the `munge` method defined, if any, on the property in the type definition ([example, just imagine unmunge instead of munge](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/user.rb#L119-L128))

So what we want for our should when overriding is not the unmunged value returned by `Puppet::Property#should` but rather the munged value in the `@should` array. That being said, none of the comment options define munge or unmunge and the type definition only defines munge (and won't anymore, after this PR), but they might some day down the road so it's good to get this right. Thanks for the pointing it out, will update.

MosesMendoza:
passing run of new acceptance test on osx 10.12:
https://gist.github.com/MosesMendoza/6e7c05acaeadfdfd874e98082406478b

passing run of new acceptance on ubuntu 16.04:
https://gist.github.com/MosesMendoza/642ff50dd2c6dea5b70cfb542d26117b

MosesMendoza:
OS X user provider can't handle UTF-8 characters when locale is POSIX (need to ticket). I wasn't getting much traction getting Windows to appropriately handle UTF-8 characters in description either (though the windows user provider isn't strictly impacted by the changes here, it may need to be ticketed separately too)

MosesMendoza:
https://tickets.puppetlabs.com/browse/PUP-7031 filed for issue on Windows

lib/puppet/parameter.rb
```diff
@@ -539,6 +539,7 @@ def to_s
   #
   # @return [String]
   def format(fmt, *args)
+    # TODO handle incompatible encodings
```
MosesMendoza: This TODO is ticketed as https://tickets.puppetlabs.com/browse/PUP-7021

MosesMendoza:
Passing run of new acceptance test on Windows 2012 R2 with fix for (PUP-7031): https://gist.github.com/MosesMendoza/ea8df6fea3207fc8e98d427c17271771

MosesMendoza:
There are several existing tests for Windows, https://github.com/puppetlabs/puppet/blob/master/spec/integration/util/windows/adsi_spec.rb#L20-L99, which cover that the strings we're retrieving via `win32ole` are returned in UTF-8 encoding in spite of a different `Encoding.default_external`. I think these tests passed prior to PUP-7031 because they weren't testing that the actual string values set/retrieved across the interface were the same - just that they were encoded correctly. It seems to me that this has now been covered by the new acceptance test in this PR (which failed prior to these changes on Windows and passes now).

Iristyle:
@MosesMendoza this looks good - would you just mind mentioning in efd997c that you have improved upon changes made in https://github.com/puppetlabs/puppet/commit/f17c853601c34f0cfd05e8d8e70157a111a516d2 as part of https://tickets.puppetlabs.com/browse/PUP-6499

Iristyle:
In efd997c commit message:

```
It turns out, WIN32OLE offers a class-level setting for Codepage,
initially determined by Encoding.default_internal, falling back to
Encoding.default_internal. On Windows systems in the default codepage, this
means the WIN32OLE interface was initialized with IBM 437 codepage.
```

Should be updated a bit:

```
It turns out, WIN32OLE offers a class-level setting for Codepage,
initially determined by Encoding.default_internal, falling back to
Encoding.default_external. On US English Windows systems the default external
codepage is typically IBM 437.  COM BSTRs, like those read through the Ruby
WIN32OLE interface, are encoded as UTF-16LE. Conversions from UTF-16LE
to other codepages like 437, can potentially result in corruption / loss of data.
```

Based on the docs at https://github.com/ruby/ruby/blob/trunk/ext/win32ole/win32ole.c#L2251-L2273:

```
/*
 *  call-seq:
 *     WIN32OLE.codepage = CP
 *
 *  Sets current codepage.
 *  The WIN32OLE.codepage is initialized according to
 *  Encoding.default_internal.
 *  If Encoding.default_internal is nil then WIN32OLE.codepage
 *  is initialized according to Encoding.default_external.
 *
 *     WIN32OLE.codepage = WIN32OLE::CP_UTF8
 *     WIN32OLE.codepage = 65001
 */
```

lib/puppet/feature/base.rb
```diff
@@ -18,7 +18,7 @@
   begin
     # ruby
     require 'Win32API'          # case matters in this require!
-    require 'win32ole'
+    require 'win32ole' ; WIN32OLE.codepage = WIN32OLE::CP_UTF8
```
Iristyle: Might be good to add a quick note here about this globally ensures we get strings with `Encoding::UTF_8` back across the board when using Ruby's late bound COM support through `WIN32OLE` (Don't need to mention it there, but our early-bound COM support goes through our FFI layer and we already handle string encodings properly there.)

Were you able to figure out how `WIN32OLE.locale` relates?  Might also be worth mentioning something here about that... and how the default is 2048 - which is not listed in the MS table at https://msdn.microsoft.com/en-us/library/ms912047(v=winembedded.10).aspx

lib/puppet/parameter.rb
MosesMendoza: 👍

lib/puppet/feature/base.rb
MosesMendoza: Note added. Haven't found the significance of `WIN32OLE.locale` in this context - also noted.

MosesMendoza:
@Iristyle I think I've addressed all your feedback - Commits modified to reflect more accurate problem descriptions/context, and several additional code comments have been modified/improved. Also as noted above the acceptance test was modified to only run against applicable hosts (i.e., exclude redundant tests on Windows). I think it's more intuitive to follow now.

Iristyle:
Great job @MosesMendoza - let's merge it up and run it through CI!

PR #5420, opened 2016-12-10
(PUP-6989) Use bundler to install dependencies for git acceptance

joshcooper:
Previously, `git` style acceptance cloned each repo specified in `config/git/options.rb`. If the repo contained `install.rb` at the root, then acceptance would call that file to install its lib directory and binstubs into system ruby.

However, this approach doesn't work for arbitrary dependencies, and `git` acceptance was broken when we added a hard dependency on `semantic_puppet`.

This commit installs bundler into system ruby, generates a Gemfile for the `puppet` gem at the ref specified by `ENV['SHA']`. It uses bundler to install a standalone bundle into system ruby and creates binstubs, e.g. `/usr/bin/puppet`. This eliminates the duplicate dependency information, and ensures `git` acceptance works as new dependencies are added.

A few tests need to be skipped for `git` because they either rely on AIO packaging, e.g. `/opt/puppetlabs/puppet/VERSION`, rely on `static_file_content` REST endpoints only found in puppetserver, or rely on puppetserver's service scripts.

This commit works around an issue where `puppet master --compile` writes the node.yaml as root, which prevents later catalog compilations for that node to fail, see [PUP-6995](https://tickets.puppetlabs.com/browse/PUP-6995). And it skips `no_exception_in_reduce_with_bignum.rb`, which needs additional work, see [PUP-6994](https://tickets.puppetlabs.com/browse/PUP-6994).

This has only been tested using centos-7-x86_64 agents and all tests pass or are skipped. It doesn't work on Windows, in particular, as it appears our VMs no longer contain msysgit (instead relying on cygwin git, so the call to `bundle install` fails).

### DIFF
--- a/acceptance/config/git/options.rb
+++ b/acceptance/config/git/options.rb
@@ -1,7 +1,5 @@
 {
   :install => [
-    'facter#2.x',
-    'hiera#1.3.4',
     'puppet',
   ],
   :pre_suite => [
--- a/acceptance/setup/git/pre-suite/000_EnvSetup.rb
+++ b/acceptance/setup/git/pre-suite/000_EnvSetup.rb
@@ -133,7 +133,9 @@ hosts.each do |host|
   case host['platform']
   when /solaris/
     step "#{host} Install json from rubygems"
-    on host, 'gem install json_pure --no-ri --no-rdoc'
+    on host, 'gem install json_pure --no-ri --no-rdoc --version 1.8.3' # json_pure 2.0 requires ruby 2
+    on host, 'gem install bundler --no-ri --no-rdoc'
+    on host, "ln -sf /opt/csw/bin/bundle #{host['puppetbindir']}/bundle"
   when /windows/
     on host, 'cmd /c gem install bundler --no-ri --no-rdoc'
   else
--- a/acceptance/setup/git/pre-suite/000_EnvSetup.rb
+++ b/acceptance/setup/git/pre-suite/000_EnvSetup.rb
@@ -133,6 +133,10 @@ hosts.each do |host|
   case host['platform']
   when /solaris/
     step "#{host} Install json from rubygems"
-    on host, 'gem install json_pure'
+    on host, 'gem install json_pure --no-ri --no-rdoc'
+  when /windows/
+    on host, 'cmd /c gem install bundler --no-ri --no-rdoc'
+  else
+    on host, 'gem install bundler --no-ri --no-rdoc'
   end
 end
--- a/acceptance/setup/git/pre-suite/010_TestSetup.rb
+++ b/acceptance/setup/git/pre-suite/010_TestSetup.rb
@@ -55,9 +55,10 @@ END
 
   step "Hosts: create environments directory like AIO does" do
     hosts.each do |host|
-      on host, "mkdir -p /etc/puppetlabs/code/environments/production/manifests"
-      on host, "mkdir -p /etc/puppetlabs/code/environments/production/modules"
-      on host, "chmod -R 755 /etc/puppetlabs/code"
+      codedir = host.puppet['codedir']
+      on host, "mkdir -p #{codedir}/environments/production/manifests"
+      on host, "mkdir -p #{codedir}/environments/production/modules"
+      on host, "chmod -R 755 #{codedir}"
     end
   end
 end
--- a/acceptance/setup/git/pre-suite/010_TestSetup.rb
+++ b/acceptance/setup/git/pre-suite/010_TestSetup.rb
@@ -29,13 +29,11 @@ extend Beaker::DSL::InstallUtils
         on host, "ln -s #{source_dir} #{checkout_dir}"
         on host, "cd #{checkout_dir} && if [ -f install.rb ]; then ruby ./install.rb ; else true; fi"
       else
-        install_from_git_on host, SourcePath, repository
-      end
-
-      if index == 1
-        versions[repository[:name]] = find_git_repo_versions(host,
-                                                             SourcePath,
-                                                             repository)
+        create_remote_file(host, '/tmp/Gemfile', <<END)
+source ENV['GEM_SOURCE'] || 'https://rubygems.org'
+gem '#{repository[:name]}', :git => '#{repository[:path]}', :ref => '#{ENV['SHA']}'
+END
+        on host, "cd /tmp && bundle install --system --binstubs /usr/bin --standalone"
       end
     end
   end
--- a/acceptance/setup/git/pre-suite/010_TestSetup.rb
+++ b/acceptance/setup/git/pre-suite/010_TestSetup.rb
@@ -29,11 +29,21 @@ extend Beaker::DSL::InstallUtils
         on host, "ln -s #{source_dir} #{checkout_dir}"
         on host, "cd #{checkout_dir} && if [ -f install.rb ]; then ruby ./install.rb ; else true; fi"
       else
+        puppet_dir = host.tmpdir('puppet')
+        on(host, "chmod 755 #{puppet_dir}")
+
         create_remote_file(host, "#{puppet_dir}/Gemfile", <<END)
 source '#{ENV["GEM_SOURCE"] || "https://rubygems.org"}'
 gem '#{repository[:name]}', :git => '#{repository[:path]}', :ref => '#{ENV['SHA']}'
 END
-        on host, "cd /tmp && bundle install --system --binstubs /usr/bin --standalone"
+        case host['platform']
+        when /windows/
+          raise ArgumentError, "Windows is not yet supported"
+        when /solaris/
+          on host, "cd #{puppet_dir} && bundle install --system --binstubs #{host['puppetbindir']} --standalone  --shebang #{host['puppetbindir']}/ruby"
+        else
+          on host, "cd #{puppet_dir} && bundle install --system --binstubs #{host['puppetbindir']} --standalone"
+        end
       end
     end
   end
--- a/acceptance/setup/git/pre-suite/010_TestSetup.rb
+++ b/acceptance/setup/git/pre-suite/010_TestSetup.rb
@@ -54,4 +54,12 @@ extend Beaker::DSL::InstallUtils
       end
     end
   end
+
+  step "Hosts: create environments directory like AIO does" do
+    hosts.each do |host|
+      on host, "mkdir -p /etc/puppetlabs/code/environments/production/manifests"
+      on host, "mkdir -p /etc/puppetlabs/code/environments/production/modules"
+      on host, "chmod -R 755 /etc/puppetlabs/code"
+    end
+  end
 end
--- a/acceptance/setup/git/pre-suite/010_TestSetup.rb
+++ b/acceptance/setup/git/pre-suite/010_TestSetup.rb
@@ -29,8 +29,8 @@ extend Beaker::DSL::InstallUtils
         on host, "ln -s #{source_dir} #{checkout_dir}"
         on host, "cd #{checkout_dir} && if [ -f install.rb ]; then ruby ./install.rb ; else true; fi"
       else
-        create_remote_file(host, '/tmp/Gemfile', <<END)
-source ENV['GEM_SOURCE'] || 'https://rubygems.org'
+        create_remote_file(host, "#{puppet_dir}/Gemfile", <<END)
+source '#{ENV["GEM_SOURCE"] || "https://rubygems.org"}'
 gem '#{repository[:name]}', :git => '#{repository[:path]}', :ref => '#{ENV['SHA']}'
 END
         on host, "cd /tmp && bundle install --system --binstubs /usr/bin --standalone"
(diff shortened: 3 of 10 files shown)


acceptance/setup/git/pre-suite/000_EnvSetup.rb
```diff
@@ -133,6 +133,10 @@
   case host['platform']
   when /solaris/
     step "#{host} Install json from rubygems"
-    on host, 'gem install json_pure'
+    on host, 'gem install json_pure --no-ri --no-rdoc'
```
joshcooper: missing bundler here, need to try on solaris too

acceptance/setup/git/pre-suite/010_TestSetup.rb
```diff
@@ -29,13 +29,11 @@
...
-
-      if index == 1
-        versions[repository[:name]] = find_git_repo_versions(host,
-                                                             SourcePath,
-                                                             repository)
+        create_remote_file(host, '/tmp/Gemfile', <<END)
+source ENV['GEM_SOURCE'] || 'https://rubygems.org'
+gem '#{repository[:name]}', :git => '#{repository[:path]}', :ref => '#{ENV['SHA']}'
+END
+        on host, "cd /tmp && bundle install --system --binstubs /usr/bin --standalone"
```
joshcooper: maybe use `host['puppetbindir']` instead of hardcoding `/usr/bin`

acceptance/config/git/options.rb
```diff
@@ -1,7 +1,5 @@
 {
   :install => [
-    'facter#2.x',
-    'hiera#1.3.4',
     'puppet',
```
joshcooper: maybe expand out `puppet` to `https://github.com/puppetlabs/puppet` so it's less magical

acceptance/setup/git/pre-suite/010_TestSetup.rb
```diff
@@ -54,4 +52,12 @@
       end
     end
   end
+
+  step "Hosts: create environments directory like AIO does" do
+    hosts.each do |host|
+      on host, "mkdir -p /etc/puppetlabs/code/environments/production/manifests"
```
MosesMendoza: Are these paths what we want on Windows? Or is this setup only for *nix?

acceptance/tests/ensure_version_file.rb
```diff
@@ -6,7 +6,7 @@
 
 test_name 'PA-466: Ensure version file is created on agent' do
 
-  skip_test 'requires version file which is created by AIO' if [:gem, :git].include?(@options[:type])
+  skip_test 'requires version file which is created by AIO' if @options[:type] != 'aio'
```
MosesMendoza: for future reference this is one i'd like to move to the puppet-agent test suite once that gets off the ground (https://github.com/puppetlabs/puppet-agent/pull/909)

acceptance/setup/git/pre-suite/000_EnvSetup.rb
joshcooper: updated, had to pin json_pure to ~> 1.8.3

acceptance/setup/git/pre-suite/010_TestSetup.rb
joshcooper: Used `host.tmpdir` but had to account for permissions when the puppet master drops privileges.

joshcooper:
Ran this against the following agent platforms:

```
2016-12-12 21:09:24	[+]	vmpooler	centos-7-x86_64	psqn9tv406vcykv.delivery.puppetlabs.net (centos7-64-1)
2016-12-12 21:09:24	[+]	vmpooler	ubuntu-14.04-amd64	eb3qdo7mnwoe1c6.delivery.puppetlabs.net (ubuntu1404-64-1)
2016-12-12 21:09:24	[+]	vmpooler	solaris-11-i386	mpqowkjjc10uki3.delivery.puppetlabs.net (solaris11-64-1)
```

Centos and ubuntu passed all tests, solaris failed on three:

```
                          Test Case tests/resource/tidy/resources_should_be_non_isomorphic.rb reported: #<Minitest::Assertion: Unexpected error was detected.
                        </Error:/> expected to not match
                        <"\e[1;31mError: Transaction store file /opt/puppetlabs/puppet/cache/state/transactionstore.yaml is corrupt (undefined method `root' for #<Psych::Nodes::Mapping:0x9031048>); replacing\nWrapped exception:\nundefined method `root' for #<Psych::Nodes::Mapping:0x9031048>\e[0m\n">.>
                            Test line: tests/resource/tidy/resources_should_be_non_isomorphic.rb:31:in `block (5 levels) in run_test'
                          Test Case tests/modules/install/basic_install.rb reported: #<Beaker::Host::CommandFailure: Host 'mpqowkjjc10uki3.delivery.puppetlabs.net' exited with 1 running:
                         puppet module install pmtacceptance-nginx
                        Last 10 lines of output were:
                        	[mNotice: Preparing to install into /etc/puppetlabs/code/environments/production/modules ...[0m
                        	[mNotice: Downloading from https://forgeapi.puppet.com ...[0m
                        	[1;31mError: Could not connect via HTTPS to https://forgeapi.puppet.com
                        	  Unable to verify the SSL certificate
                        	    The certificate may not be signed by a valid CA
                        	    The CA bundle included with OpenSSL may not be valid or up to date[0m>
                            Test line: tests/modules/install/basic_install.rb:22:in `block (2 levels) in run_test'
                          Test Case tests/modules/install/with_version.rb reported: #<Beaker::Host::CommandFailure: Host 'mpqowkjjc10uki3.delivery.puppetlabs.net' exited with 1 running:
                         puppet module install --version '< 1.7.0' pmtacceptance-java
                        Last 10 lines of output were:
                        	[mNotice: Preparing to install into /etc/puppetlabs/code/environments/production/modules ...[0m
                        	[mNotice: Downloading from https://forgeapi.puppet.com ...[0m
                        	[1;31mError: Could not connect via HTTPS to https://forgeapi.puppet.com
                        	  Unable to verify the SSL certificate
                        	    The certificate may not be signed by a valid CA
                        	    The CA bundle included with OpenSSL may not be valid or up to date[0m>
                            Test line: tests/modules/install/with_version.rb:22:in `block (2 levels) in run_test'
...

       Total Suite Time: 8510.30 seconds
      Average Test Time: 26.43 seconds
              Attempted: 322
                 Passed: 286
                 Failed: 1
                Errored: 2
                Skipped: 31
                Pending: 2
                  Total: 322
```

The first error is because we're using ruby19 on solaris with git. The other two are ssl issues because we're not using puppet-agent with our ca bundle.

acceptance/config/git/options.rb
joshcooper: decided to leave it as is.

joshcooper:
@puppetlabs/puppet-maintainers ready for review and merge

acceptance/tests/ensure_version_file.rb
er0ck: i'd really like to change the type name here to 'package' or something similar.  but it's fine to now do that in this PR.

er0ck:
```
✓ erict@erict ~/projects/puppet/acceptance
 (joshcooper-ticket/master/PUP-6989-git-acceptance %)$ be rake ci:test:git TESTS=empty.rb
rake aborted!
Requires commit SHA to be put under test as environment variable: SHA='<sha>'.
Also must set BEAKER_HOSTS=config/nodes/foo.yaml or include it in an options.rb for Beaker,
or specify TEST_TARGET in a form beaker-hostgenerator accepts, e.g. ubuntu1504-64a.
You may override the default master test target by specifying MASTER_TEST_TARGET.
You may set TESTS=path/to/test,and/more/tests.
You may set additional Beaker OPTIONS='--more --options'
If testing from git checkouts, you may optionally set the github fork to checkout from using PUPPET_FORK='some-other-puppet-fork' (you may change the HIERA_FORK and FACTER_FORK as well if you wish).
You may also optionally set the git server to checkout repos from using GIT_SERVER='some.git.mirror'.
Or you may set PUPPET_GIT_SERVER='my.host.with.git.daemon', specifically, if you have set up a `git daemon` to pull local commits from.  (You will need to allow the git daemon to serve the repo (see `git help daemon` and the docs/acceptance_tests.md for more details)).
If there is a Beaker options hash in a ./local_options.rb, it will be included.  Commandline options set through the above environment variables will override settings in this file.
/home/erict/projects/puppet/acceptance/Rakefile:314:in `block (2 levels) in <top (required)>'
```

should this just figure out the `SHA` or rather use `HEAD` of master, unless specified?

joshcooper:
We could default it to the current `HEAD`, at least that way there is a chance of the tests and code being installed remotely being in-sync. I'll update that for `git` style acceptance.

er0ck:
i think it uses HEAD.  so maybe SHA could default to `master` in the git case?

er0ck:
now i have that age-old gem install issue on windows:
```
s3yaimwmgmjxo8p.delivery.puppetlabs.net (windows2012r2-64-1) 21:35:28$ env PATH="/usr/bin:${PATH}" cmd /c gem source --add https://rubygems.org
  Error fetching https://rubygems.org:
        SSL_connect returned=1 errno=0 state=error: certificate verify failed (https://api.rubygems.org/specs.4.8.gz)

s3yaimwmgmjxo8p.delivery.puppetlabs.net (windows2012r2-64-1) executed in 2.25 seconds
Exited: 1
Beaker::Host::CommandFailure: Host 's3yaimwmgmjxo8p.delivery.puppetlabs.net' exited with 1 running:
 env PATH="/usr/bin:${PATH}" cmd /c gem source --add https://rubygems.org
Last 10 lines of output were:
        Error fetching https://rubygems.org:
                SSL_connect returned=1 errno=0 state=error: certificate verify failed (https://api.rubygems.org/specs.4.8.gz)
/home/erict/projects/puppet/acceptance/.bundle/gems/beaker-3.5.0/lib/beaker/host.rb:366
/home/erict/projects/puppet/acceptance/.bundle/gems/beaker-3.5.0/lib/beaker/dsl/helpers/host_helpers.rb:83
/home/erict/projects/puppet/acceptance/.bundle/gems/beaker-3.5.0/lib/beaker/shared/host_manager.rb:127
/home/erict/projects/puppet/acceptance/.bundle/gems/beaker-3.5.0/lib/beaker/dsl/patterns.rb:37
```

joshcooper:
On second thought defaulting `SHA` is magically and could fail for mysterious reasons. For example, you're working on a PR in a topic branch, you make commits, fix tests, etc. Then you go to run acceptance, but don't specify `SHA`. If we default to the current `HEAD`, then acceptance will clone from upstream puppet and it won't be able to checkout that `SHA`. If we default to `master` and you're branched off of `stable`, then there's a good chance the older tests won't work when run against the new code in the git checkout.

joshcooper:
About the Windows issue, you need to set `GEM_SOURCE=http://rubygems.delivery.puppetlabs.net`.  That said `ci:test:git` doesn't work on Windows right now due to the VM using cygwin git and bundler binstubs not working on Windows.

er0ck:
ah yes.  i see that exception...
maybe this in 000_EnvSetup instead? 
```
129 hosts.each do |host|
130   raise ArgumentError, "Windows is not yet supported" if host['platform'] =~ /windows/
131   # Only configure gem mirror after Ruby has been installed, but before any gems are installed.
132   configure_gem_mirror(host)
133 end
```

er0ck:
okay, i see Josh's concerns on defaulting SHA to `master`
we need it to fail fast when people are running against a Fork that might not be puppetlabs/ for new tests against new features/fixes, and they try to just run `rake ci:test:git`

er0ck:
ugh... sles11:
```
    Test Case setup/git/pre-suite/010_TestSetup.rb reported: #<Beaker::Host::CommandFailure: Host 'kp741mocucg1wli.delivery.puppetlabs.net' exited with 5 running:
   cd /tmp/puppet.R177jb && bundle install --system --binstubs /usr/bin --standalone
  Last 10 lines of output were:
        Don't run Bundler as root. Bundler can ask for sudo if it is needed, and
        installing your bundle as root will break this application for all non-root
        users on this machine.
        Fetching https://github.com/puppetlabs/puppet.git
        Fetching gem metadata from https://rubygems.org/..........
        Fetching version metadata from https://rubygems.org/.
        Resolving dependencies...
        fast_gettext-1.3.0 requires ruby version >= 2.1.0, which is incompatible with
        the current version, ruby 2.0.0p353>
      Test line: setup/git/pre-suite/010_TestSetup.rb:45:in `block (4 levels) in run_test'
```
hopefully we can just pin gettext to earlier

joshcooper:
Supporting sles would be a change to the main Gemfile (or `.gemspec`) to pin back to the earlier version. As it is now, I don't think you can do a `gem install puppet` on sles.

er0ck:
ok.  your platforms are fine for development.  although ensuring that tests get run against windows is generally pretty important. but given the problems there, i'm fine with this as-is

joshcooper:
🚢

MosesMendoza:
I'm +1 on requiring that SHA be specified - I've seen that inexplicable magic get hard to follow quickly and it makes that setting harder to discover for the (many) use-cases where we don't want HEAD.

er0ck:
i'm cool with requiring SHA, in this case it's super easy as it can be just `master` or `stable` or `er0ck/sweet_as_branch`

it's hellish for packages which require an actual github ref full SHA *AND* some version string

PR #5422, opened 2016-12-12
(PUP-6996) Add correct permissions to auto-generated production environment

abottchen:
When Puppet is run, it will create the production environment if it
is not already present.  If that run is done at the command line,
the resulting directory will be owned by root:root.  This is not
consistent with the ownership required by Code Manager and file
sync.

This patch corrects the permissions to be those set in the Puppet
configuration.  This prevents any possible permissions issues with
other Puppet components.

### DIFF
--- a/lib/puppet/settings.rb
+++ b/lib/puppet/settings.rb
@@ -1156,14 +1156,23 @@ Generated on #{Time.now}.
     if configured_environment == "production" && envdir && Puppet::FileSystem.exist?(envdir)
       configured_environment_path = File.join(envdir, configured_environment)
       if !Puppet::FileSystem.symlink?(configured_environment_path)
-        catalog.add_resource(
-          Puppet::Resource.new(:file,
-                               configured_environment_path,
-                               :parameters => { :ensure => 'directory',
-                                                :owner => self[:user],
-                                                :group => self[:group],
-                                                :mode => "0750" })
-        )
+        eid = Process::UID.eid
+        if eid == 0 or eid == Etc.getpwnam(self[:user]).uid
+          catalog.add_resource(
+            Puppet::Resource.new(:file,
+                                 configured_environment_path,
+                                 :parameters => { :ensure => 'directory',
+                                                  :owner => self[:user],
+                                                  :group => self[:group],
+                                                  :mode => "0750" })
+          )
+        else
+          catalog.add_resource(
+            Puppet::Resource.new(:file,
+                                 configured_environment_path,
+                                 :parameters => { :ensure => 'directory' })
+          )
+        end
       end
     end
   end
--- a/lib/puppet/settings.rb
+++ b/lib/puppet/settings.rb
@@ -1156,8 +1156,7 @@ Generated on #{Time.now}.
     if configured_environment == "production" && envdir && Puppet::FileSystem.exist?(envdir)
       configured_environment_path = File.join(envdir, configured_environment)
       if !Puppet::FileSystem.symlink?(configured_environment_path)
-        eid = Process::UID.eid
-        if eid == 0 or eid == Etc.getpwnam(self[:user]).uid
+        if @service_user_available and @service_group_available
           catalog.add_resource(
             Puppet::Resource.new(:file,
                                  configured_environment_path,
--- a/lib/puppet/settings.rb
+++ b/lib/puppet/settings.rb
@@ -1159,7 +1159,10 @@ Generated on #{Time.now}.
         catalog.add_resource(
           Puppet::Resource.new(:file,
                                configured_environment_path,
-                               :parameters => { :ensure => 'directory' })
+                               :parameters => { :ensure => 'directory',
+                                                :owner => self[:user],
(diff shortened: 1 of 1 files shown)


adrienthebo:
We're getting test failures because this change doesn't account for Puppet being run as a non-root, non-puppet user. If we're non-root and Puppet is running as a different user from the `user` setting then we'll need to disable this permission setting.

abottchen:
@adrienthebo Thanks, I did not understand why it was failing.  I added checks as you outlined.

abottchen:
@adrienthebo It looks as though Puppet is configured to run as the user "puppet" even though that user does not exist on the system.  Am I reading it incorrectly?

adrienthebo:
@abottchen that's my guess; Puppet looks to be using the default value of `puppet` for user and group but is running as the `travis` user in travis-ci.

lib/puppet/settings.rb
```diff
@@ -1156,11 +1156,23 @@ def add_environment_resources(catalog, sections)
...
-        )
+        eid = Process::UID.eid
+        if eid == 0 or eid == Etc.getpwnam(self[:user]).uid
+          catalog.add_resource(
+            Puppet::Resource.new(:file,
+                                 configured_environment_path,
+                                 :parameters => { :ensure => 'directory',
+                                                  :owner => self[:user],
+                                                  :group => self[:group],
+                                                  :mode => "0750" })
```
joshcooper: I think you want to set `owner` and `group` to `service`. If puppet is running as `root` and the service user/group account exists, then puppet will set the owner/group accordingly.

If we're not running as root, then puppet can't chown, so no point in trying.

If we are root, but the `puppet` user/group doesn't exist, then we fall back to either root or a platform-dependent account.

Note the `puppet` user/group accounts will typically only exist when the agent is running on a host that is also a puppetmaster, or the agent upgraded from 3.x.

lib/puppet/settings.rb
```diff
@@ -1156,11 +1156,23 @@ def add_environment_resources(catalog, sections)
     if configured_environment == "production" && envdir && Puppet::FileSystem.exist?(envdir)
       configured_environment_path = File.join(envdir, configured_environment)
       if !Puppet::FileSystem.symlink?(configured_environment_path)
-        catalog.add_resource(
-          Puppet::Resource.new(:file,
-                               configured_environment_path,
-                               :parameters => { :ensure => 'directory' })
-        )
+        eid = Process::UID.eid
+        if eid == 0 or eid == Etc.getpwnam(self[:user]).uid
```
joshcooper: and you should be able to drop this part

abottchen:
I don't see a variable named "service", but I very well may be overlooking something.  I just added a check to see if the user and group are available first to avoid the problems in CI.

lib/puppet/settings.rb
```diff
@@ -1156,11 +1156,22 @@ def add_environment_resources(catalog, sections)
     if configured_environment == "production" && envdir && Puppet::FileSystem.exist?(envdir)
       configured_environment_path = File.join(envdir, configured_environment)
       if !Puppet::FileSystem.symlink?(configured_environment_path)
-        catalog.add_resource(
-          Puppet::Resource.new(:file,
-                               configured_environment_path,
-                               :parameters => { :ensure => 'directory' })
-        )
+        if @service_user_available and @service_group_available
```
joshcooper: You'll want set the owner/group to the string `"service"` like we do here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L79-L81. It's a special value meaning set the owner/group to the service account if it exists, otherwise set it to `root`. The service account is the value of `Puppet[:user]` which is typically `puppet`. This way the caller doesn't have to duplicate that logic, so you can remove this line here and the else block below.

joshcooper:
Ping @abottchen, let me know if my comment makes sense.

abottchen:
@joshcooper When I set the user and group to "service":

```
      if !Puppet::FileSystem.symlink?(configured_environment_path)
        catalog.add_resource(
          Puppet::Resource.new(:file,
                               configured_environment_path,
                               :parameters => { :ensure => 'directory',
                                                :owner => "service",
                                                :group => "service" })
        )
      end
```

Puppet errors out:

```
Error: Could not set 'directory' on ensure: Could not find user service
Error: Could not set 'directory' on ensure: Could not find user service
Wrapped exception:
Could not find user service
```

The lib/puppet/service.rb code looks to be taking "service" literally and not interpolating it to another user name.

Magisus:
@joshcooper thoughts on the failures Adam mentioned? Is something missing, or might the bahavior have changed?

joshcooper:
So I think the issue is that we only map "service" to the `Puppet[:user]` for file settings, not for resources in general. We'd either need to create the prod environment directory a setting, which I don't think want?, or extract the code that handles the "service" accounts into a more generally reusable module.

ericldelaney:
Closing this PR, for a new one 
https://github.com/puppetlabs/puppet/pull/6751

PR #5423, opened 2016-12-12
(PUP-6966) Allow file fifo and socket ensure absent

tkishel:
Prior to this commit:

- The `remove_existing` method didn't check the `self[:backup]` attribute before calling `can_backup` and `backup_existing`.
- The `remove_existing` method reported that it could not remove a file instead of reporting that it could not backup a file.
- The `can_backup` method returned `true` for all file types.

This caused File (and Tidy, which uses File) to fail to remove fifo and socket files and report a misleading error.

This commit allows File (and Tidy) to remove fifo and socket files.


Magisus:
This looks reasonable; however it could have some unforeseen implications, particularly for Tidy, so we're going to want to make sure the change is documented.

Magisus:
After letting this run through our validation process, we realized that it contains one change that has more far-reaching consequences: we did not previously fail to attempt to remove if we could not backup. While we agree failing is probably the correct behavior, we want to be more deliberate about that change. It is also likely that in most scenarios, attempts to backup will fail in more specific ways themselves, or (as in the directory without `force` case) a different error will prevent the file from being removed. In light of this, I have put up https://github.com/puppetlabs/puppet/pull/6103 changing this failure to a warning for the time being.

PR #5426, opened 2016-12-13
(PUP-1209) Allow setting empty environment vars

petems:
* Adds branch to logic to allow setting empty environment variable
* For example, allows setting of `HTTP_PROXY` to empty for an exec

### DIFF
--- a/lib/puppet/provider/exec.rb
+++ b/lib/puppet/provider/exec.rb
@@ -41,6 +41,13 @@ class Puppet::Provider::Exec < Puppet::Provider
                 warning "Overriding environment setting '#{env_name}' with '#{value}'"
               end
               environment[env_name] = value
+            elsif setting =~ /^(\w+)=$/
+              env_name = $1
+              warning "Environment Setting #{env_name} set to `#{env_name}=`, if this is unexpected check the variable has been assigned in the string"
+              if environment.include?(env_name) || environment.include?(env_name.to_sym)
+                warning "Overriding environment setting '#{env_name}' with '#{value}'"
+              end
+              environment[env_name] = value
             else
               warning "Cannot understand environment setting #{setting.inspect}"
             end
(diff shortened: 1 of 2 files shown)


spec/unit/provider/exec/posix_spec.rb
```diff
@@ -106,6 +106,17 @@ def make_exe
...
+      provider.resource[:environment] = ['WHATEVER=']
+      command = make_exe
+
+      Puppet::Util::Execution.expects(:execute).with(command, instance_of(Hash)).returns(Puppet::Util::Execution::ProcessOutput.new('', 0))
+
+      provider.run(command)
+
+      expect(@logs.map {|l| "#{l.level}: #{l.message}" }).to eq(["warning: Environment Setting WHATEVER set to `WHATEVER=`, if this is unexpected check the variable has been assigned in the string"])
+    end
+
```
petems: This needs better tests, but I'm not sure about the tests around environment settings, they're a bit complicated, might need a bit of assistance 👍

lib/puppet/provider/exec.rb
```diff
@@ -41,6 +41,13 @@ def run(command, check = false)
                 warning "Overriding environment setting '#{env_name}' with '#{value}'"
               end
               environment[env_name] = value
+            elsif setting =~ /^(\w+)=$/
+              env_name = $1
+              warning "Environment Setting #{env_name} set to `#{env_name}=`, if this is unexpected check the variable has been assigned in the string"
```
MikaelSmith: Interesting that it always warns when setting env variables. I'd like to see the messages cleaned up a bit so we always produce one warning that describes whether we're unsetting an environment variable or having no effect. Especially since the 2nd warning right now will always have an empty-string value.

MosesMendoza:
This PR hasn't been updated in some time, and it seems unlikely to get picked up in the near future. I'm going to close this for now, and we can re-open if/when PUP-1209 is prioritized for the 5.x series. We can track future conversation in that ticket.

PR #5398, opened 2016-11-29
(PUP-6188) consider umask for default perms FS test

MosesMendoza:
Prior to this commit, a test in file_system_spec.rb would assert an expected
default permission for a newly created file by Puppet::FileSystem, but failed
to take into account that this would vary based on umask (which varies in linux
by root vs non-root user). This commit updates the test to instead calculate
the expected permission based on the current umask.

transaction spec

PR #5431, opened 2016-12-14
(PUP-6188) Fix SYSTEM perms for file_system_spec

Iristyle:
- When running specs as SYSTEM, the temp directory used is typically
   C:\Windows\Temp, and when creating new files the perms should be
   typically written as:

   SYSTEM:(F)
   Administrators:(F)

   When Puppet calculates a POSIX mode, the owner is SYSTEM and the
   group is Administrators, yielding a mode of 770

   In some circumstances there may be another ACE on the file (as can
   be the case when IIS_IUSRS has permissions to C:\Windows\Temp) and
   Puppets calculated mode will include the 2 prefix which is a
   sentinel value used to indicate extra aces, resulting in a mode of
   2000770.

 - Previously the tests assumed that tests were being run as another
   user that was writing to their own isolated temp directory. Such
   files are normally written as:

   MACHINE\USER:(I)(F)
   SYSTEM:(I)(F)
   Administrators:(I)(F)

   When Puppet calculates mode in this case, given the user has full
   control, but the Users group is not explicitly granted access,
   Puppet calculates a mode of 700.  With the extra aces for SYSTEM
   and Administrators group present, it includes the 2 prefix again,
   resulting in a mode of 2000700.

 - The test fixes here account for the differences in Puppets mode
   calculation

### DIFF
--- a/spec/unit/file_system_spec.rb
+++ b/spec/unit/file_system_spec.rb
@@ -24,6 +24,12 @@ describe "Puppet::FileSystem" do
     file.close
   end
 
+  SYSTEM_SID_BYTES = [1, 1, 0, 0, 0, 0, 0, 5, 18, 0, 0, 0]
+
+  def is_current_user_system?
+    SYSTEM_SID_BYTES == Puppet::Util::Windows::ADSI::User.current_user_sid.sid_bytes
+  end
+
   context "#open" do
     it "uses the same default mode as File.open, when specifying a nil mode (umask used on non-Windows)" do
       file = tmpfile('file_to_update')
@@ -32,11 +38,16 @@ describe "Puppet::FileSystem" do
       Puppet::FileSystem.open(file, nil, 'a') { |fh| fh.write('') }
 
       expected_perms = Puppet::Util::Platform.windows? ?
-        # default Windows mode based on temp file storage
-        '2000700' :
+        # default Windows mode based on temp file storage for SYSTEM user or regular user
+        # for Jenkins or other services running as SYSTEM writing to c:\windows\temp
+        # the permissions will typically be SYSTEM(F) / Administrators(F) which is 770
+        # but sometimes there are extra users like IIS_IUSRS granted rights which adds the "extra ace" 2
+        # for local Administrators writing to their own temp folders under c:\users\USER
+        # they will have (F) for themselves, and Users will not have a permission, hence 700
+        (is_current_user_system? ? ['770', '2000770'] : '2000700') :
         # default mode is applied. 100 == 'regular file'
         '100' + (666 - File.umask.to_s(8).to_i).to_s
-      expect(Puppet::FileSystem.stat(file).mode.to_s(8)).to eq(expected_perms)
+      expect([expected_perms].flatten).to include(Puppet::FileSystem.stat(file).mode.to_s(8))
 
       default_file = tmpfile('file_to_update2')
       expect(Puppet::FileSystem.exist?(default_file)).to be_falsey
@@ -54,8 +65,10 @@ describe "Puppet::FileSystem" do
 
       # Behavior may change in the future on Windows, to *actually* change perms
       # but for now, setting a mode doesn't touch them
-      expected_perms = Puppet::Util::Platform.windows? ? '2000700' : '100444'
-      expect(Puppet::FileSystem.stat(file).mode.to_s(8)).to eq(expected_perms)
+      expected_perms = Puppet::Util::Platform.windows? ?
+        (is_current_user_system? ? ['770', '2000770'] : '2000700') :
+        '100444'
+      expect([expected_perms].flatten).to include(Puppet::FileSystem.stat(file).mode.to_s(8))
 
       # On Windows, 0444 happens to round-trip properly (most octals *do not*)
       # Setting a mode controls file attribute setting (like archive, read-only, etc)


Iristyle:
This is a follow-up to #5398 that was just recently identified as failing in Jenkins CI.next (due to some anomalies in how the jobs were running / the fact that they weren't running all of the tests).

MosesMendoza:
👍

spec/unit/file_system_spec.rb
```diff
@@ -32,11 +38,16 @@ def with_file_content(content)
...
 
       expected_perms = Puppet::Util::Platform.windows? ?
-        # default Windows mode based on temp file storage
-        '2000700' :
+        # default Windows mode based on temp file storage for SYSTEM user or regular user
+        # for Jenkins or other services running as SYSTEM writing to c:\windows\temp
+        # the permissions will typically be SYSTEM(F) / Administrators(F) which is 770
+        # but sometimes there are extra users like IIS_IUSRS granted rights which adds the "extra ace" 2
+        # for local Administrators writing to their own temp folders under c:\users\USER
+        # they will have (F) for themselves, and Users will not have a permission, hence 700
```
Iristyle: So the `2` already covers the extra aces, which should be true all of the time.  To recap how we calculate modes (code lives at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/security.rb#L191-L256):

### On my machine

I should have used another user than `Administrator` as this may cause more confusion, but here are the results I'm seeing.

#### `C:\Windows\Temp`

```
C:\Users\Administrator\AppData\Local\Temp>icacls c:\Windows\Temp
c:\Windows\Temp NT AUTHORITY\SYSTEM:(OI)(CI)(S,RD)
                BUILTIN\IIS_IUSRS:(OI)(CI)(S,RD)
                BUILTIN\Users:(CI)(S,WD,AD,X)
                BUILTIN\Administrators:(F)
                BUILTIN\Administrators:(OI)(CI)(IO)(F)
                NT AUTHORITY\SYSTEM:(F)
                NT AUTHORITY\SYSTEM:(OI)(CI)(IO)(F)
                CREATOR OWNER:(OI)(CI)(IO)(F)

Successfully processed 1 files; Failed processing 0 files
```

* Puppet calculates both owner and group as `Administrators` (`S-1-5-32-544`) (Using `Puppet::Util::Windows::Security.get_group` and `get_owner`). 
* Running as `SYSTEM` or as `Administrator`, the mode gets calculated as `2000770`

#### For a the local `Administrator` temp directory:

```
C:\Users\Administrator\AppData\Local\Temp>icacls c:\Users\Administrator\AppData\Local\Temp
c:\Users\Administrator\AppData\Local\Temp Everyone:(RX)
                                          NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
                                          BUILTIN\Administrators:(I)(OI)(CI)(F)
                                          VAGRANT-2008R2\Administrator:(I)(OI)(CI)(F)

Successfully processed 1 files; Failed processing 0 files
```

* Puppet calculates the owner as `Administrators` and the group as `None`
* The `SYSTEM` and `Administrator` accounts both see the mode as `2000755`

### If I create some files within those directories, the outcome changes a little.

#### Assuming a file `c:\windows\temp\test.txt` created by `SYSTEM`:

I can create the file from cmd or Ruby and `icacls` shows the same thing:

```
C:\Users\Administrator\AppData\Local\Temp>icacls c:\Windows\Temp\test.txt
c:\Windows\Temp\test.txt NT AUTHORITY\SYSTEM:(I)(S,RD)
                         BUILTIN\IIS_IUSRS:(I)(S,RD)
                         BUILTIN\Administrators:(I)(F)
                         NT AUTHORITY\SYSTEM:(I)(F)

Successfully processed 1 files; Failed processing 0 files
```

But, if I create the file from Ruby with `File.open` I get:

* Puppet sees the owner as `Administrators` / group as `SYSTEM` (`S-1-5-18`)
* The `SYSTEM` and `Administrator` accounts both sees mode as `2000770`

If I create the file from inside `cmd` with `echo 'hi' > test.txt`, the results are different!

* Puppet sees the owner as `Administrators` / group as `None`
* The `SYSTEM` and `Administrator` accounts both see mode as `2000700`

#### Assuming a file `C:\Users\Administrator\AppData\Local\Temp` created by `Administrator`:

```
C:\Users\Administrator\AppData\Local\Temp>icacls test.txt
test.txt NT AUTHORITY\SYSTEM:(I)(F)
         BUILTIN\Administrators:(I)(F)
         VAGRANT-2008R2\Administrator:(I)(F)

Successfully processed 1 files; Failed processing 0 files
```

It doesn't matter if the file is created through cmd or Ruby - the behavior is the same.

* Puppet sees the owner as `Administrators` / group as `None`
* The `SYSTEM` and `Administrator` account both see the mode as `2000700`

### As for your question

This is what I *think* you're asking:

* Should we also include `700` (no extra ACEs)? Nope, shouldn't ever happen - it looks like the `2` comes in from the `Administrator` SID as it's not considered the `group` or `owner` (this seems weird to me to be honest)
* Should we also include `770`(the `None` group has full permissions)? Nope, `None` is not explicitly set at all so should always be `0` (this is also pretty weird to be honest)
* Should we also include `2000770`? Nope, for the same reason a above.

This makes me think that we should have a big table somewhere that shows real-world examples of modes on Windows because it's pretty confusing to be honest.

/cc @joshcooper

MosesMendoza: Thanks for the write-up @Iristyle. Agreed it's a bit confusing which seems like it could open us up to accidentally setting incorrect attributes somewhere.

PR #5432, opened 2016-12-14
(MAINT) Remove the execute bits from test files

kurtwall:
Mostly a tidiness change but I couldn't see a compelling reason for Beaker test files to have their execute bits set.


joshcooper:
@kurtwall could you fix the typo in the commit summary (in git not the PR itself)?

kurtwall:
@joshcooper Think I did this right.

PR #5437, opened 2016-12-14
(PUP-7015) Add multilanguage support for AppVeyor

glennsarti:
Previously Appveyor PR CI tests only tested against the default US English
language.  This commit modifies the AppVeyor tests to optionally set the system
locale prior to running the test suite.  This is configured using the LOCALE
environment variable.

The ruby test matrix, which was previously set via environment variables, has
been moved to the Platform setting to create 2 axis test matrix (Platform vs
Env. Variables).  The Platform setting is presented as an additional environment
variable in Appveyor jobs which is used by the `SET PATH....` statement to set
the correct ruby path and version.

This PR also fixes two tests which were failing on a Japanese locale on Windows.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,15 +2,47 @@ version: 4.1.0.{build}
 clone_depth: 10
 install:
 build: off
+matrix:
+  fast_finish: true
 
+# Ruby versions under test
+platform:
+  - Ruby23-x64
+  - Ruby21-x64
+
+# Note that locales are mainly code page changes and are not the FULL set of localization
+#  changes that happen when you install French/Germam/Japanese etc. Windows
 environment:
   matrix:
-  - RUBY_VER: 23-x64
-  - RUBY_VER: 21-x64
+  - LOCALE: en-US
+  - LOCALE: ja-JP
+
+before_test:
+  # Semicolons are required below.  This renders as a single line powershell command in AppVeyor
+  - ps: >-
+      [string]$testLocale = [Environment]::GetEnvironmentVariable("LOCALE","Process");
+      $currentLocale = (Get-WinSystemLocale).Name;
+      if ($testLocale -ne '') {
+        if ($testLocale -ne $currentLocale) {
+          Write-Output "Setting Locale to $testLocale ...";
+          Set-WinSystemLocale ($testLocale);
+          Write-Output "Restarting worker in 5 seconds...";
+          Start-Sleep -Seconds 5
+          Restart-Computer -Force -Confirm:$false;
+        } else {
+          Write-Output "Locale is already set to $currentLocale";
+        };
+      } else {
+        Write-Output "LOCALE environment variable not set";
+      };
 
 test_script:
+  # The sleep is needed to allow the Appveyor tracing to start correctly after a reboot (locale change)
+  - ps: Start-Sleep -Seconds 5
   - SET
-  - SET PATH=C:\Ruby%RUBY_VER%\bin;%PATH%
+  - chcp
+  - ps: Get-WinSystemLocale
+  - SET PATH=C:\%PLATFORM%\bin;%PATH%
   - SET LOG_SPEC_ORDER=true
   - bundle install --jobs 4 --retry 2 --without development extra
   - type Gemfile.lock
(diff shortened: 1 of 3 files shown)


Iristyle:
So this appears to double our runtime.

I'm OK with that for now, but wondering what we can do to improve it in the future?

appveyor.yml
```diff
@@ -2,15 +2,51 @@ version: 4.1.0.{build}
...
+          Write-Output "Locale is already set to $currentLocale";
+        };
+      } else {
+        Write-Output "LOCALE environment variable not set";
+      };
 
 test_script:
+  # The REMs are needed to help AppVeyor NOT to lose console messages after a reboot (locale change)
+  # Typically the first 2 or 3 are dropped.  Unknown as to why, it just is.
+  - REM Starting in 4
```
Iristyle: Have you asked AppVeyor team about this?

glennsarti:
We either;
- Need more appveyor workers (probably won't happen)
- Test tiering to reduce the number of tests
- Reduce matrix size to 2

spec/integration/test/test_helper_spec.rb
```diff
@@ -12,7 +12,11 @@
 
     Puppet::Util::Windows::Process.set_environment_variable(@varname, @rune_utf8)
   }
-
+  after(:all) {
+    @varname = 'test-helper-foo'
```
Iristyle: I think the intent is to reuse `@varname` everywhere - i.e. this should be unnecessary, no?

spec/integration/test/test_helper_spec.rb
```diff
@@ -12,7 +12,11 @@
 
     Puppet::Util::Windows::Process.set_environment_variable(@varname, @rune_utf8)
   }
-
+  after(:all) {
+    @varname = 'test-helper-foo'
+    Puppet::Util::Windows::Process.set_environment_variable(@varname, nil)
```
Iristyle: Add a comment here about why we're doing this?

glennsarti:
e.g. perhaps only run spec/integration for non US language?

spec/integration/test/test_helper_spec.rb
glennsarti: Yes this in unnecessary.  Will review.

spec/unit/util_spec.rb
```diff
@@ -182,14 +182,51 @@ def withenv_utf8(&block)
...
+    it "works around Ruby bug 8822 (which fails to preserve UTF-8 properly when accessing ENV) (Ruby >= 2.3.x, external encoding is not 65001/UTF8) ",
+      :if => ((match = RUBY_VERSION.match(/^2\.(\d+)\./)) && match.captures[0].to_i >= 3 && Encoding.default_external.name != "UTF-8" && Puppet.features.microsoft_windows?) do
 
       withenv_utf8 do |utf_8_key, utf_8_value, codepage_key|
         # Ruby 2.3 fixes access by the original UTF-8 key, and behaves differently than 2.1
         # keying by local codepage will work only when the UTF-8 can be converted to local codepage
         # the key selected for this test contains characters unavailable to a local codepage, hence doesn't work
-        expect(ENV.key?(codepage_key)).to eq(false)
+
+        if Encoding::default_external == Encoding.find("CP932")
```
Iristyle: I'm not sure this is how we want to fix this test - I think it might be better to just choose different test data.

Looking into it - not sure I entirely understand just yet.

glennsarti: I tried the mixed_utf8 from https://github.com/puppetlabs/puppet/blob/8a355b73875aa58acd8a95717e80c31fb0df0795/spec/integration/faces/config_spec.rb#L13 and long utf8 string https://github.com/puppetlabs/puppet/blob/ebd96213cab43bb2a8071b7ac0206c3ed0be8e58/spec/integration/test/test_helper_spec.rb#L11 .  Both exhibited the same behaviour.

glennsarti: and korean version of lorum ipsum (https://medium.com/korean-design-posts/%EB%B9%88%EA%B3%B5%EA%B0%84%EC%9D%84-%EC%B1%84%EC%9B%8C%EC%A3%BC%EB%8A%94-%EC%9C%A0%EC%9A%A9%ED%95%9C-%ED%88%B4-882b0363bb5a#.sw73xxjh6)

Iristyle:
Closed in favor of updates in #5452

PR #5439, opened 2016-12-15
(maint) Ensure that Puppet::LookupContext.cache returns the cached value

thallgren:


hlindberg:
LookupContext is API so needs a test. Also think this deserves a ticket.

PR #5443, opened 2016-12-16
(PUP-6982) Allow keys to be specified using patterns in lookup_options

thallgren:
Before this commit, in order for a key to have lookup options specified,
the options had to be keyed with that same exact key. Now, such options can
also be defined for keys that matches a specific pattern. This is achieved
by treating all keys that start with a '^' as regular expression patterns.
A key that matches such a pattern is considered to have the options below
that key.

The commit also adds a check that key or pattern under `lookup_options`
in a module only can identify keys that belong to that module. Keys must
 start with "<module_name>::" and patterns must start with "^" followed
 by "<module_name>::".

### DIFF
--- a/spec/unit/functions/lookup_spec.rb
+++ b/spec/unit/functions/lookup_spec.rb
@@ -250,6 +250,73 @@ describe "The lookup function" do
       end
     end
 
+    context 'with lookup_options configured using patterns' do
+      let(:environment_files) do
+        {
+          env_name => {
+            'hiera.yaml' => <<-YAML.unindent,
+              ---
+              version: 5
+              hierarchy:
+                - name: X
+                  paths:
+                  - first.yaml
+                  - second.yaml
+              YAML
+            'data' => {
+              'first.yaml' => <<-YAML.unindent,
+                a:
+                  aa:
+                    aaa: a.aa.aaa
+                b:
+                  ba:
+                    baa: b.ba.baa
+                  bb:
+                    bba: b.bb.bba
+                c:
+                  ca:
+                    caa: c.ca.caa
+                lookup_options:
+                  b:
+                    merge: hash
+                  '^[^b]':
+                     merge: deep
+                  '^c':
+                     merge: first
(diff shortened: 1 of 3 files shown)


spec/unit/functions/lookup_spec.rb
```diff
@@ -250,6 +250,73 @@ def explain(key, options = {})
...
+        expect(lookup('a')).to eql({'aa' => { 'aaa' => 'a.aa.aaa', 'aab' => 'a.aa.aab' }})
+      end
+
+      it 'gives a direct key match higher priority than a matching pattern' do
+        expect(lookup('b')).to eql({'ba' => { 'baa' => 'b.ba.baa' }, 'bb' => { 'bba'=>'b.bb.bba' }})
+      end
+
+      it 'uses the first matching pattern' do
+        expect(lookup('c')).to eql({'ca' => { 'caa' => 'c.ca.caa', 'cab' => 'c.ca.cab' }})
+      end
```
hlindberg: This only has one matching pattern. There needs to be a test that tests that the first matching pattern of a set of matching patterns is selected. For example ^c.* pattern comes before ^c pattern (both matching 'ca').

spec/unit/functions/lookup_spec.rb
```diff
@@ -250,6 +250,73 @@ def explain(key, options = {})
...
+    end
+
```
hlindberg: Test that pattern based lookup_entries merge across the layers based on the pattern. For example ^c in one layer merges with ^c in another, but does not merge with ^c.* (matches the same key, but is a different pattern).

spec/unit/functions/lookup_spec.rb
thallgren: Not sure I understand. 'c' matches both ^[^b] and '^c' so there are at least two matching patterns.

hlindberg: ok - I missed that

PR #5444, opened 2016-12-16
(PUP-7029) dnf provider does not support Fedora 25

blackknight36:
The dnf provider must be updated to add each new Fedora release.
This is caused by the defaultfor method in puppet which only accepts
hashes as arguments.  See https://tickets.puppetlabs.com/browse/PUP-6452
for more details.

### DIFF
--- a/lib/puppet/provider/package/dnf.rb
+++ b/lib/puppet/provider/package/dnf.rb
@@ -28,7 +28,7 @@ Puppet::Type.type(:package).provide :dnf, :parent => :yum do
       end
   end
 
-  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => ['22', '23', '24']
+  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => ['22', '23', '24', '25']
 
   def self.update_command
     # In DNF, update is deprecated for upgrade


MikaelSmith:
This is a simple change, looks good. Could you update the commit message to reflect our https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md guidelines?

blackknight36:
Sure thing.  I'll create a ticket and update the PR.

blackknight36:
Actually, I was thinking it may be better to address this issue using a range of releases to avoid having to update the provider file for every Fedora releases.  Ideally a simple >= statement would be used but that is not supported at this time.

For example, I've updated our internal puppet master as follows.

-  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => ['22', '23', '24']
+  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => (22..30).to_a

This buys some time until Fedora 31 is released but is still not a permanent solution.

MikaelSmith:
Thanks! I agree being able to rely on some info about version numbers would be nice for guarding these things. And it would be better to default to allowing more and fixing it if that turns out to be wrong (rather than always having to fix it). If you want to put up a PR allowing more versions, go for it.

PR #5445, opened 2016-12-16
(maint) Fix windows default directories in install.rb

Magisus:
Prior to this commit, the win32/dir gem was not being required properly
in all places that install.rb attempts to use its constants. This commit
moves the require to a higher-level location, which allows all the supplied
defaults to resolve correctly.

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -260,15 +260,19 @@ def prepare_installation
     $operatingsystem = Facter.value :operatingsystem
   end
 
-  if not InstallOptions.configdir.nil?
-    configdir = InstallOptions.configdir
-  elsif $operatingsystem == "windows"
+  if $operatingsystem == "windows"
     begin
+      # populates constants used to specify default Windows directories
       require 'win32/dir'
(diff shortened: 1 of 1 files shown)


install.rb
```diff
@@ -260,15 +260,12 @@ def prepare_installation
     $operatingsystem = Facter.value :operatingsystem
   end
 
+  # populates constants used to specify default Windows directories
+  require 'win32/dir' if $operatingsystem == "windows"
```
Iristyle: Do we care about the old message?  Or is the new Ruby failure to load `win32/dir` message enough to point us in the right direction?

Magisus: Oh good point, should probably put that back in.

Magisus: Is it reasonable to assume that if we don't have win32/dir, we also don't have the others the message mentions? Or should I make it more specific?

Iristyle: Yeah, it probably indicates a failure of some kind to setup an environment / package it.

PR #5446, opened 2016-12-19
(PUP-7030) Fix bug causing lexer hick-up on two adjacent unicode chars

thallgren:
The lexer was only capable of recognizing the first of several adjacent
unicode characters. This commit fixes that problem.


Iristyle:
TIL: `hickup` is the archaic spelling of `hiccup`

PR #5448, opened 2016-12-21
(PUP-6967) Reload puppetserver using HUP

melissa:
Bump beaker to 3.5.x, which added support for reloading puppetserver
using HUP.

Beaker will fail to reload the service if it isn't already running and
it will fall back to bouncing the service using puppet's resource
service ensure=stopped and ensure=running commands. To ensure reloading
always works, we now start puppetserver in the 045_EnsureMasterStarted
pre-suite step. We don't test against passenger or webrick, so there's
no point in making the step conditional on the server type.

Prior to commit 59cce2311, Beaker's `with_puppet_running_on` method would
start puppetserver twice. Once in the beginning before the test, and
once after. Each start takes ~30 seconds per test, and that caused an
overall 1+ hour increase in test times as compared to when we tested
against passenger packages.

In commit 59cce2311, we set `restart_when_done` to true to avoid the
second start, which saved ~30 minutes. However, we have to undo that
change in order to use beaker's reload functionality.

That said, we don't regress on test times, because reloading avoids both
puppetserver starts, and puts us back on par with test times prior to
the move to puppetserver.

This commit also moves the logic for setting `use-service` from an
explicit pre-suite step to the options hash, where all of the other
options are specified. Beaker converts the symbol to a string when
processing the options, so the existing behavior is preserved.

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -12,7 +12,7 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.1.0')
+gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.5.0')
 gem "beaker-hostgenerator", *location_for(ENV['BEAKER_HOSTGENERATOR_VERSION'] || "~> 0.3")
 gem "beaker-abs", *location_for(ENV['BEAKER_ABS_VERSION'] || "~> 0.2")
 gem "rake", "~> 10.1"
--- a/acceptance/config/aio/options.rb
+++ b/acceptance/config/aio/options.rb
@@ -1,17 +1,16 @@
 {
   :type => 'aio',
   :is_puppetserver => true,
+  :'use-service' => true,
   :puppetservice => 'puppetserver',
   :'puppetserver-confdir' => '/etc/puppetlabs/puppetserver/conf.d',
-  :restart_when_done => false,
   :pre_suite => [
     'setup/common/pre-suite/000-delete-puppet-when-none.rb',
     'setup/aio/pre-suite/010_Install.rb',
-    'setup/aio/pre-suite/015_PackageHostsPresets.rb',
     'setup/aio/pre-suite/020_InstallCumulusModules.rb',
     'setup/aio/pre-suite/021_InstallAristaModule.rb',
     'setup/common/pre-suite/025_StopFirewall.rb',
     'setup/common/pre-suite/040_ValidateSignCert.rb',
-    'setup/aio/pre-suite/045_EnsureMasterStartedOnPassenger.rb',
+    'setup/aio/pre-suite/045_EnsureMasterStarted.rb',
   ],
 }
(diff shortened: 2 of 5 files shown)


melissa:
According to PE-17417, the puppetserver backport this had been blocked on has been merged into 2016.4.x. @joshcooper do you know if there's anything else blocking this work?

MikaelSmith:
😹

melissa:
ping @joshcooper @camlow325 @er0ck if you could take a look at this, that would be wonderful

er0ck:
LGTM.  i'll probably run this locally to see if there are issues

er0ck:
this ran, but failed a few tests and errored a few against master:
```
                                                                                                                     │       Total Suite Time: 3837.41 seconds
                                                                                                                     │      Average Test Time: 12.14 seconds
                                                                                                                     │              Attempted: 316
                                                                                                                     │                 Passed: 249
                                                                                                                     │                 Failed: 10
                                                                                                                     │                Errored: 1
                                                                                                                     │                Skipped: 55
                                                                                                                     │                Pending: 1
                                                                                                                     │                  Total: 316
```

i haven't looked into the fails yet, but master passed PA tests yesterday

melissa:
Running this against LTS-1.7, I get no errors

```
      You can reproduce this run with:
      /home/melissa/puppet/acceptance/.bundle/gems/bin/beaker --options-file merged_options.rb --hosts=tmp/redhat7-64m-centos7-64a-069f2ec8-78ff-41ca-9e6c-910f5a09d86e.yaml

      Important ENV variables that may have affected your run:
          BUNDLE_PATH           .bundle/gems
          BUNDLE_BIN            .bundle/bin
          GEM_HOME              /home/melissa/puppet/acceptance/.bundle/gems
          GEM_PATH
          RUBYLIB               /home/melissa/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib:/home/melissa/.rbenv/rbenv.d/exec/gem-rehash
          PATH          /home/melissa/puppet/acceptance/.bundle/gems/bin:/home/melissa/.rbenv/versions/2.3.0/bin:/home/melissa/.rbenv/libexec:/home/melissa/.rbenv/plugins/ruby-build/bin:/home/melissa/.rbenv/shims:/home/melissa/.rbenv/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/melissa/bin
          home          /home/melissa
Beaker completed successfully, thanks.
```

I tested with a puppet-agent package built with these changes to puppet, which can be found at builds.delivery.puppetlabs.net/puppet-agent/8b2e8ed79855ed0d909ac4b99e8b1a2c9c333e7c

I

er0ck:
excellent.  i'm running it against master with a windows agent now.  
i'm almost fine merging this, but am concerned about it breaking master upon merge-up.
i'll wait and see what it does...

melissa:
This code is already in master and stable, it was a cherry pick back to the LTS branch

er0ck:
oh.  i see that now 😊

melissa:
woo!

PR #5449, opened 2016-12-22
(PUP-869) deprecate file face and add easier face deprecations

MosesMendoza:
PUP-869 tracks the deprecation of the `file` face - one of many upcoming deprecations of existing faces in the puppet codebase. After initial review, issuing a simple deprecation warning in a face proved more complex than one might hope - especially in testing that face is actually deprecated when a face is actually used.  I.e., we could just put a deprecation warning in the initial face definition block, but this block is only evaluated once, at load time - a deprecation there might be fine from a ux perspective but is executed via an `instance_eval` and difficult to test. This commit adds a simple `deprecated` attribute to the Puppet::Interface class, and ensures that if a face is deprecated, we issue an appropriate deprecation warning when an action is sent to that face.

The (perhaps most appreciable) benefit of this is that it allows us to deprecate a face with a simple call to `face.deprecate` in that face's definition block.

This PR then leverages the prior commit's support for face deprecation to deprecate the file face. The file face is broken in puppet 4.x and its actions are inconsistent. It will be removed in a future release.

### DIFF
--- a/lib/puppet/face/file.rb
+++ b/lib/puppet/face/file.rb
@@ -44,4 +44,7 @@ Puppet::Indirector::Face.define(:file, '0.0.1') do
   deactivate_action(:destroy)
 
   set_indirection_name :file_bucket_file
+
+  # The file face is deprecated
+  deprecate
 end
--- a/lib/puppet/application/face_base.rb
+++ b/lib/puppet/application/face_base.rb
@@ -245,6 +245,10 @@ class Puppet::Application::FaceBase < Puppet::Application
       end
     end
 
+    if @face.deprecated?
+      Puppet.deprecation_warning("'puppet #{@face.name}' is deprecated and will be removed in a future release")
+    end
+
     result = @face.send(@action.name, *arguments)
     puts render(result, arguments) unless result.nil?
     status = true
--- a/lib/puppet/face/help/face.erb
+++ b/lib/puppet/face/help/face.erb
@@ -1,4 +1,7 @@
 <%# encoding: UTF-8%>
+<% if face.deprecated? -%>
+<%= "Warning: 'puppet #{face.name}' is deprecated and will be removed in a future release." %>
+<% end %>
 <% if face.synopsis -%>
 USAGE: <%= face.synopsis %>
 
--- a/lib/puppet/interface.rb
+++ b/lib/puppet/interface.rb
@@ -180,6 +180,15 @@ class Puppet::Interface
     "Puppet::Face[#{name.inspect}, #{version.inspect}]"
   end
 
+  # @return [void]
+  def deprecate
+    @deprecated = true
+  end
+
+  # @return [Boolean]
+  def deprecated?
+    @deprecated
+  end
   ########################################################################
   # Action decoration, whee!  You are not expected to care about this code,
   # which exists to support face building and construction.  I marked these
--- a/spec/unit/application/face_base_spec.rb
+++ b/spec/unit/application/face_base_spec.rb
@@ -267,6 +267,22 @@ describe Puppet::Application::FaceBase do
       app.expects(:render).with(app.arguments.length + 1, ["myname", "myarg"])
       expect { app.main }.to exit_with(0)
     end
+
+    it "should issue a deprecation warning if the face is deprecated" do
+      # since app is shared across examples, stub to avoid affecting shared context
+      app.face.stubs(:deprecated?).returns(true)
+      app.face.expects(:foo).with(*app.arguments)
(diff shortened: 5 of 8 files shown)


MosesMendoza:
Admittedly might be way off base here with 7fefcea - may be tip-toeing up to over-engineering but I have a cold today.

lib/puppet/face/file.rb
```diff
@@ -44,4 +44,7 @@
   deactivate_action(:destroy)
 
   set_indirection_name :file_bucket_file
+
+  # The file face is deprecated
+  self.deprecate
```
joshcooper: nit, can we drop `self.`?

MosesMendoza: of course. updated

MosesMendoza:
Good question on acceptance tests - will do a local run and link results

MosesMendoza:
passed a full `rake ci:test:aio` run against ubuntu 16.04. https://gist.github.com/MosesMendoza/b918088437ba6392c3846e82c0956e13

MosesMendoza:
not sure why it shows that appveyor result as failed - following it to the job, it succeeded:
```
Finished in 8 minutes 8 seconds
21933 examples, 0 failures, 67 pending
```

joshcooper:
Looks like appveyor failed due to `Build execution time has reached the maximum allowed time for your plan (60 minutes).` but it doesn't look like the job actually took that long...?

MosesMendoza:
I re-ran the appveyor job and it passed as expected. Maybe just a problem with resources on appveyor's side.

MosesMendoza:
@joshcooper is this good to merge?

glennsarti:
Should the help text of a face also indicate if the face is deprecated?

e.g.

Shows a deprecation
```
C:\source\puppet-pr5449 [master ↑]> be puppet file info
Warning: 'puppet file' is deprecated and will be removed in a future release
   (at C:/Source/puppet-pr5449/lib/puppet/application/face_base.rb:249:in `main')
Run mode 'user': selector
```

Does not show a deprecation
```
C:\source\puppet-pr5449 [master ↑]> be puppet help file

USAGE: puppet file <action> [--terminus TERMINUS] [--extra HASH]

This subcommand interacts with objects stored in a local or remote
filebucket. File objects are accessed by their MD5 sum; see the
examples for the relevant syntax.

OPTIONS:
  --render-as FORMAT             - The rendering format to use.
  --verbose                      - Whether to log verbosely.
  --debug                        - Whether to log debug information.
  --extra HASH                   - Extra arguments to pass to the indirection
                                   request
  --terminus TERMINUS            - The indirector terminus to use.

ACTIONS:
  download    Download a file into the local filebucket.
  find        Retrieve a file from the filebucket.
  info        Print the default terminus class for this face.
  save        API only: create or overwrite an object.
  store       Store a file in the local filebucket.

TERMINI: file, rest, selector

See 'puppet man file' or 'man puppet-file' for full help.
```

branan:
I agree with Glenn that it would be nice to show the deprecation in the help output as well.

I also like the implementation of deprecation living in face_base - we will almost certainly be deprecating other faces in the future. I think this is a nice clean abstraction of that, and not over-engineered at all.

MosesMendoza:
@glennsarti @branan 👍  - deprecation added to help output as well.

glennsarti:
Looks good 

```
C:\Source\puppet-pr5449 [master ↑]> bundle exec puppet help file

Warning: 'puppet file' is deprecated and will be removed in a future release.

USAGE: puppet file <action> [--terminus TERMINUS] [--extra HASH]

This subcommand interacts with objects stored in a local or remote
filebucket. File objects are accessed by their MD5 sum; see the
examples for the relevant syntax.

OPTIONS:
  --render-as FORMAT             - The rendering format to use.
  --verbose                      - Whether to log verbosely.
  --debug                        - Whether to log debug information.
  --extra HASH                   - Extra arguments to pass to the indirection
                                   request
  --terminus TERMINUS            - The indirector terminus to use.

ACTIONS:
  download    Download a file into the local filebucket.
  find        Retrieve a file from the filebucket.
  info        Print the default terminus class for this face.
  save        API only: create or overwrite an object.
  store       Store a file in the local filebucket.

TERMINI: file, rest, selector

See 'puppet man file' or 'man puppet-file' for full help.
C:\Source\puppet-pr5449 [master ↑]> bundle exec puppet help facts


USAGE: puppet facts <action> [--terminus TERMINUS] [--extra HASH]

This subcommand manages facts, which are collections of normalized system
information used by Puppet. It can read facts directly from the local system
(with the default `facter` terminus).
...
```

PR #5450, opened 2016-12-22
(PUP-6981) Ensure that legacy Hiera logger is initialized

thallgren:
Before this commit, the class Hiera wasn't loaded properly and its logger
was not initialized. Consequently, any custom backend that attempted to use
the logger would run into an error similar to: "Unable to instantiate
backend '<backend name>': undefined method 'debug' for nil:NilClass".

This commit adds the proper require for the class and ensures that the
logger is initialized.

### DIFF
--- a/lib/puppet/pops/lookup/hiera_config.rb
+++ b/lib/puppet/pops/lookup/hiera_config.rb
@@ -241,7 +241,7 @@ class HieraConfigV3 < HieraConfig
         if @config.include?(KEY_LOGGER)
           Hiera.logger = @config[KEY_LOGGER].to_s
         else
-          Hiera.logger = 'console'
+          Hiera.logger = 'puppet'
         end
         Hiera::Config.instance_variable_set(:@config, @loaded_config)
 
@@ -261,7 +261,6 @@ class HieraConfigV3 < HieraConfig
     config[KEY_VERSION] ||= 3
     config[KEY_BACKENDS] ||= 'yaml'
     config[KEY_HIERARCHY] ||= %w(nodes/%{::trusted.certname} common)
-    config[KEY_LOGGER] ||= 'console'
     config[KEY_MERGE_BEHAVIOR] ||= 'native'
     config[KEY_DEEP_MERGE_OPTIONS] ||= {}
 
--- a/lib/puppet/pops/lookup/hiera_config.rb
+++ b/lib/puppet/pops/lookup/hiera_config.rb
@@ -236,8 +236,13 @@ class HieraConfigV3 < HieraConfig
       when 'json', 'yaml'
         create_data_provider(backend, parent_data_provider, KEY_V3_DATA_HASH, "#{backend}_data", { KEY_DATADIR => datadir }, paths)
       else
-        # Custom backend. Hiera v3 must be installed and it must be made aware of the loaded config
-        require 'hiera/config'
+        # Custom backend. Hiera v3 must be installed, it's logger configured, and it must be made aware of the loaded config
+        require 'hiera'
+        if @config.include?(KEY_LOGGER)
+          Hiera.logger = @config[KEY_LOGGER].to_s
+        else
+          Hiera.logger = 'console'
+        end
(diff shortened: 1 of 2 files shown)


lib/puppet/pops/lookup/hiera_config.rb
```diff
@@ -236,8 +236,13 @@ def create_configured_data_providers(lookup_invocation, parent_data_provider)
...
         create_data_provider(backend, parent_data_provider, KEY_V3_DATA_HASH, "#{backend}_data", { KEY_DATADIR => datadir }, paths)
       else
-        # Custom backend. Hiera v3 must be installed and it must be made aware of the loaded config
-        require 'hiera/config'
+        # Custom backend. Hiera v3 must be installed, it's logger configured, and it must be made aware of the loaded config
+        require 'hiera'
+        if @config.include?(KEY_LOGGER)
+          Hiera.logger = @config[KEY_LOGGER].to_s
+        else
+          Hiera.logger = 'console'
```
jpartlow: @thallgren I've tested this out on another vm, and puppet infrastructure configure succeeds with this patch. The default to the console logger seems like a change from HieraPuppet, where you would end up with the puppet logger (so we're getting a lot of debug output in the run). https://github.com/thallgren/puppet/blob/afad625267b5d8bff4f36b28d65625707df5ac98/lib/hiera_puppet.rb#L65

We can add a :logger: 'puppet' to the pe-installer-shim's hiera.yaml easily enough, but given that HieraPuppet is no longer in use (if I'm understanding these changes), it seems like this constitutes a behavior change for anyone using hiera?

jpartlow:
@thallgren another question I have is around one of the warnings that I'm seeing now:

2016-12-22 08:39:04,572 - [Warning]: Undefined variable 'trusted.certname'; 
   (file & line not available)

Is this coming from https://github.com/puppetlabs/pe-installer-shim/blob/2017.1.x/conf.d/hiera.yaml#L9

And if so, does that mean that hiera/lookup will no longer use fact interpolation to determine hierarchy?

lib/puppet/pops/lookup/hiera_config.rb
thallgren: I'll change the default logger to 'puppet'.

thallgren:
AFAIK, facts are available as top scope variables. There's no special interpolation strategy for them. So the question to ask here is, why aren't the facts available in the top scope?

thallgren:
Another thing worth noting is that if the facts are added to the scope at a later stage, that will be picked up during the next lookup. There would be a warning to that fact too (that variables used for interpolation in the config has changed).

jpartlow:
@thallgren I've verified on rwaqyic018h2t6z.delivery.puppetlabs.net that the nodes hierarchy is being used and parameters are being pulled in based on nodes/%{trusted.certname}.conf settings.

I think the warning is bubbling up from:

https://github.com/thallgren/puppet/blob/afad625267b5d8bff4f36b28d65625707df5ac98/lib/puppet/pops/lookup/hiera_config.rb#L161-L163

thallgren:
New commit is pushed. And yes, that's probably where the warning originates, reason being that the first times that the configuration is read, the variable isn't configured yet. It then comes into existence, and everything works as expected. The same thing probably happened before although with the old Hiera, there's no code checking if the config is stable. Instead, everything is completely reconfigured from scratch on every lookup.

jpartlow:
@thallgren the patch to default to the puppet logger isn't doing what expect, I think because it's already been set to console in the validate_config function:

https://github.com/thallgren/puppet/blob/afad625267b5d8bff4f36b28d65625707df5ac98/lib/puppet/pops/lookup/hiera_config.rb#L264

I don't know if you can change the fallback there, or if that has other repercussions.

thallgren:
@jpartlow, that's very surprising given that the line you point to was removed in the patch. Did you really test the patch?

jpartlow:
@thallgren my mistake; I only applied part of the patch.  Seems to be working fine. :+1: from me.

PR #5451, opened 2016-12-22
Pup 7046 update dnf provider

blackknight36:
The dnf provider will only work with a pre-defined list of Fedora releases.  To avoid having to update this provider every time there is a new Fedora release this patch updates the defaultfor argument to a range which will work until Fedora 31 is released.

### DIFF
--- a/lib/puppet/provider/package/dnf.rb
+++ b/lib/puppet/provider/package/dnf.rb
@@ -28,7 +28,7 @@ Puppet::Type.type(:package).provide :dnf, :parent => :yum do
       end
   end
 
-  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => ['22', '23', '24', '25']
+  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => (22..30).to_a
 
   def self.update_command
     # In DNF, update is deprecated for upgrade
--- a/lib/puppet/provider/package/dnf.rb
+++ b/lib/puppet/provider/package/dnf.rb
@@ -28,7 +28,7 @@ Puppet::Type.type(:package).provide :dnf, :parent => :yum do
       end
   end
 
-  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => (22..30).to_a
+  defaultfor :operatingsystem => :fedora, :feature => :supports_dnf
 
   def self.update_command
     # In DNF, update is deprecated for upgrade
--- a/lib/puppet/feature/supports_dnf.rb
+++ b/lib/puppet/feature/supports_dnf.rb
@@ -1,0 +1,6 @@
+require 'puppet/util/feature'
+
+# Enable dnf feature on platforms that support it
+Puppet.features.add(:supports_dnf) do
+  File.file?('/usr/bin/dnf')
+end
--- a/spec/unit/provider/package/dnf_spec.rb
+++ b/spec/unit/provider/package/dnf_spec.rb
@@ -6,20 +6,16 @@ require 'spec_helper'
 provider_class = Puppet::Type.type(:package).provider(:dnf)
 
 context 'default' do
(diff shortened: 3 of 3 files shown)


lib/puppet/provider/package/dnf.rb
```diff
@@ -28,7 +28,7 @@
       end
   end
 
-  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => ['22', '23', '24', '25']
+  defaultfor :operatingsystem => :fedora, :operatingsystemmajrelease => (22..30).to_a
```
Iristyle: Clearly this is the simplest thing to do, but I'm wondering if it might be better here to use a feature (which allows us to execute an arbitrary check) rather than to hardcode version numbers.

I haven't tested, but it would look something like this:

```ruby
defaultfor :operatingsystem => :fedora, :feature => supports_dnf
```


There's an example of this in the posix exec provider at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/exec/posix.rb#L6 which relies on the `:posix` feature definition at https://github.com/puppetlabs/puppet/blob/937505096316f2389d5ba0c7f13b3da4eef36e0f/lib/puppet/feature/base.rb#L11

The `:feature` support was fixed as part of PUP-1766 / https://github.com/puppetlabs/puppet/commit/784b5e6090ea9090c630055fcd940e8a8f2d4f05 and lives at https://github.com/puppetlabs/puppet/blob/bf3d94cc50a941d5e4d413568dbf611e95848eb9/lib/puppet/provider.rb#L285

blackknight36:
@Iristyle I agree it would be better to avoid using version numbers since they often change and this issue will just come back later.  I'll have to do some testing to see if the :feature option will work.

blackknight36:
@Iristyle Looks like :feature doesn't work in puppet 4.2.1, I'm getting an error when I run the agent as follows.

`Error: Could not autoload puppet/provider/package/dnf: undefined local variable or method `supports_dnf' for Puppet::Type::Package::ProviderDnf:Class
Did you mean?  supports_parameter?                                                                                                                                   
Error: Could not retrieve local facts: Could not autoload puppet/provider/package/dnf: undefined local variable or method `supports_dnf' for Puppet::Type::Package::ProviderDnf:Class                                                                                                                                                     
Did you mean?  supports_parameter?                                                                                                                                   
Error: Failed to apply catalog: Could not retrieve local facts: Could not autoload puppet/provider/package/dnf: undefined local variable or method `supports_dnf' for Puppet::Type::Package::ProviderDnf:Class                                                                                                                            
Did you mean?  supports_parameter?`

blackknight36:
Apparently the string has to be quoted.  I changed the defaultfor line to look like this and puppet is working now.

`defaultfor :operatingsystem => :fedora, :feature => 'supports_dnf'`

blackknight36:
I've updated the file as suggested.  dnf should now be used as the default provider on any platform that supports it.

whopper:
@blackknight36 looks like you still need to add the actual `supports_dnf` feature, similar to how it's done here: https://github.com/puppetlabs/puppet/blob/937505096316f2389d5ba0c7f13b3da4eef36e0f/lib/puppet/feature/base.rb#L11-L14

IIRC, in tests we used `which dnf` to determine if dnf was the suitable provider, but it's been a while and you may need a more robust check (or not, not sure) :)

joshcooper:
See also https://github.com/puppetlabs/puppet/pull/5069

joshcooper:
@blackknight36 did yo mean to close this?

blackknight36:
@joshcooper Yeah, I just can't figure out how to make this work.  Adding a feature really only moves the problem to another class, there would still be a need to determine which Fedora release is being used, we can't just assume dnf should be the default just because it is installed.  What happens when Fedora switches from dnf to something else?  etc. etc.

PR #5452, opened 2016-12-22
(PUP-7015) Enable / Fix Ruby specs for Japanese Windows

Iristyle:
Supercedes #5437 

See that PR for discussion about increase in matrix time, etc.  This turns on Japanese Windows spec testing on AppVeyor *and* fixes resulting test failures.

This just has some modifications to the final commit from prior PR to reduce test complication.


glennsarti:
👍 for merge

PR #5453, opened 2016-12-27
(PUP-6777) don't run utf-8 user tests on arista

MosesMendoza:
Arista EOS has known issues with the user provider (ARISTA-42, PUP-7049)
causing it to fail the tests associated with PUP-6777. Effectively,  the
`getpwnam` function on Arista will return data for a non-existent user. Ruby
`Etc` module relies on this to return info on users, and Puppet relies on Ruby
`Etc` module in the useradd provider. This commit confines this test to not run
on EOS hosts.


ScottGarman:
:+1:

PR #5454, opened 2016-12-30
(PUP-7053) shorten username in utf-8 test for aix

MosesMendoza:
By default AIX usernames are 8 chars. To ensure we pass in the default
configuration, shorten the username length in test utf8_user_comment to 8
characters.


MosesMendoza:
@kurtwall this is good to merge

PR #5455, opened 2017-01-03
(PUP-6992) Fix inconsistent autoload of environment puppet functions

thallgren:
This commit changes the autoload behavior for puppet functions and types
that reside under the 'functions' or 'types' directory in an environment
so that the naming becomes consistent with the ruby functions that
reside under 'lib/puppet/functions'.

In essence, the namespace 'environment' is moved into a subdirectory with
the same name that functions and types in the global namespace can reside
directly under their respective 'functions' and 'types' directory.


hlindberg:
Specifications needs an update. Would like to see that change first. Also, tests AppVeyor are failing atm.

hlindberg:
Ping @puppetlabs/puppet-maintainers note the unrelated error from AppVeyor - is that a known issue?

PR #5456, opened 2017-01-03
(PUP-7058) Add Ruby API to #new function

thallgren:
This commit adds the method `create` to the `PAnyType` to make it easy to
create instances of a type from Ruby. It also adds the method `from_hash`
to the PObjectType which enables instance creation using an initialization
hash.


hlindberg:
A rebase is needed

hlindberg:
ok to merge when rebased and tests go green again

PR #5460, opened 2017-01-04
(PUP-6134) Explicitly locate gemcmd windows

tkishel:
Problem: Puppet is unable to install ruby gems into anything other than Puppet's ruby on Windows. Cause: on Windows, the puppet ruby bin directory is prepended to the PATH (in environment.bat) and puppet uses the 'which' method in util.rb to find the first gem command in the PATH.

This fix updates gem.rb to use a custom 'which_windows_gemcmd' method to find the first non-puppet gem command in the PATH. This fix also updates puppet_gem.rb to remove its dependency on the puppet ruby bin directory being prepended to the PATH.

Alternate solutions include:

1. Implement the (optional) 'path' attribute of exec resources in package providers:

```
package { 'sinatra' : 
  provider => 'gem',
  path => 'c:\Ruby23\bin',
  ensure => installed,
}
```

2. Not prepending the puppet ruby bin directory to the PATH on Windows.

(The former would appear to be simpler, but I could not find a starting point.)


tkishel:
I made an error when updating an earlier different pull request.
Before I repeat that error, I'll document this update (@gem.rb:24)

```
    puppet_ruby_dir = Puppet::Util.get_env('RUBY_DIR') ? Puppet::Util.get_env('RUBY_DIR') : 'C:\Program Files\Puppet Labs\Puppet\sys\ruby'
    puppet_ruby_bin_dir = File.join(puppet_ruby_dir,'bin')
```

lib/puppet/provider/package/gem.rb
```diff
@@ -15,7 +15,52 @@
 
   has_feature :versionable, :install_options, :uninstall_options
 
-  commands :gemcmd => "gem"
+  def self.which_windows_gemcmd(bin)
```
Iristyle: If we're going to build our own `which` I would suggest we drop it in our custom file system support.  A more complete cross-platform implementation is at https://gist.github.com/steakknife/88b6c3837a5e90a08296 FWIW

Iristyle:
Other commentary here is that I'm not sure I like the existing mostly magic behavior that this PR is trying to mimic in the Linux provider.  I think this PR being necessary to establish "parity" with Linux highlights a design flaw in the original provider.

When there are multiple Rubies on a system, our current implementation (on the Linux side) is pretty non-deterministic based on which Ruby was installed when (i.e. which one found it's way into `PATH` first or which one symlink'd into `/usr/bin` last).  That implicit behavior could lead to inconsistent behavior across a set of nodes based on when Puppet was run / how manifests were modified over time.  In other words, I'm concerned that on machine A the gem gets installed into Ruby A, but on machine B, the gem gets installed into Ruby B instead of Ruby A.  

As mentioned, one solution is to manage a symlink with Puppet that always runs before using the `gem` package provider - to ensure that the `gem` being called is the expected one. On Windows symlinks don't make sense, so supplying a `PATH` to the package provider is a better solution I think.  More generically an [environment](https://docs.puppet.com/puppet/4.8/types/exec.html#exec-attribute-environment) attribute like the one on `exec` would make the provider more flexible - but have to make sure that jives OK with our current `package` type design (given existing `install_options`)

In the end, I think I'd rather see a new attribute than a custom `which` implementation for Windows, which still suffers from the afore-mentioned ordering problem.  Also, consider what might happen If you had 3 or 4 Rubies on the machine that you wanted to manage -- without a way to override the `PATH` / `environment`, it would be pretty painful to get gems installed across all of them (well, you'd have to use a bunch of `exec` calls with different environments and you'd have to implement the idempotency yourself)

Iristyle:
I should also mention that we can't mod the existing Puppet `.bat` files at this time.

lib/puppet/provider/package/gem.rb
tkishel: I'm not thrilled with what ends up being lightly modified duplicate of which() from util.rb either. 

I thought about extending the original which() to accept a second parameter to specifically exclude 'puppet' paths (or a parameter to be used as a pattern to exclude matching paths) from the each-do-dir loop. 

But I decided to isolate the which_windows_gemcmd() in gem.rb to minimize the change/testing footprint. I reviewed the other executables in the other puppet bin directories, and didn't find any other potential providers (although I anticipate Execs accidentally finding our curl and openssl executables) that would benefit from the new method. I assume that as use of Puppet on Windows increases, this will change.

I'm open to a new, more-generally-usable method with an exclude parameter based upon file_impl.rb.

tkishel:
Yeah, I didn't think we could easily/safely modify environment.bat to change the PATH.

tkishel:
Thanks for taking the time to review this and providing feedback. I agree with what you are saying about Linux and paths.

I'll take a look at passing an optional environment (and/or path) attribute from gem->package->provider down to where gem is executed as a command. I looked for how exec implements its environment and path attributes, but didn't make much progress. If you have any suggestions, please let me know. If it hit a wall and optional attributes are the best solution, I may have to defer to others with more familiarity.

lib/puppet/provider/package.rb
```diff
@@ -53,4 +53,18 @@ def join_options(options)
...
     end.flatten
   end
+
+  # Puppet on Windows prepends its paths to PATH.
+  # Provides a which method that does not traverse those paths.
+  # @api public
+  # @param command [String] the name of the executable to find.
+  # @return [String] the absolute path to the found executable.
+  def self.which_without_puppet_paths(command)
+    raise "This method is specific to Windows" unless Puppet::Util::Platform.windows?
```
ferventcoder: This is a bit scary to see - I would prefer that the check for Windows would simply return the command or a which output on non-Windows and do the following with Windows.

lib/puppet/provider/package/gem.rb
```diff
@@ -15,7 +15,14 @@
...
 
-  commands :gemcmd => "gem"
+  if Puppet.features.microsoft_windows?
+    # Puppet on Windows prepends its paths to PATH.
+    # One of those paths includes Puppet's vendored ruby and its gem command.
+    # Use a which method that does not traverse those paths.
+    commands :gemcmd => which_without_puppet_paths("gem")
+  else
+    commands :gemcmd => "gem"
+  end
```
ferventcoder: Just make this entire call use the `commands :gemcmd => which_without_puppet_path("gem")` and handle the non-Windows versus Windows as suggested above.

ferventcoder:
Also, perhaps some spec tests with this would be beneficial.

tkishel:
I pushed a minimal footprint fix to provide functionality on Windows until a comprehensive solution (as proposed above) is developed.

tkishel:
Rebase error :(

PR #5461, opened 2017-01-04
(PUP-7059) Improve Lookup Ruby API to handle multiple instances

thallgren:
This commit adds the ability to create a separate instances of configured
lookup layers. A new instance is created by subclassing `LookupAdapter`
and then pass that class to the `Invocation` constructor. By doing so, a
new adapter will be created and adapted to the compiler. All configured
providers are then properties of the new adapter.

The `LookupAdapter` can be configured to work in a "global layer only"
mode. The global configuration (normally determined by the Puppet setting
"hiera_config" can be overridden.


hlindberg:
ok to merge once tests go green

PR #5462, opened 2017-01-04
(PUP-6448) Indirector URIs should remain UTF-8

MosesMendoza:
This PR supersedes #5176 

From @Iristyle / #5176

> By not preserving the original Path encoding of UTF-8 used for
incoming requests to the indirector, future string comparisons (using
Regex for instance) are invalid and will cause exceptions.

> This problem is trivially reproducible with a manifest like:

> file { "/tmp/test":
> ensure => file,
> source => 'puppet://modules/utf_8/静的',
> }

> The key of the request previously was "utf_8/\xE9\x9D\x99\xE7\x9A\x84"
> instead of the appropriate "utf_8/静的"

This PR includes the original commit from #5176 as well as an acceptance test and fix for the issue.

### DIFF
--- a/acceptance/tests/resource/file/ticket_6448_file_with_utf8_source.rb
+++ b/acceptance/tests/resource/file/ticket_6448_file_with_utf8_source.rb
@@ -1,0 +1,82 @@
+require 'puppet/acceptance/environment_utils'
+extend Puppet::Acceptance::EnvironmentUtils
+
+test_name 'Ensure a file resource can have a UTF-8 source attribute, content, and path when served via a module' do
+  skip_test 'requires a master for serving module content' if master.nil?
+
+  tmp_environment = mk_tmp_environment_with_teardown(master, File.basename(__FILE__, '.*'))
+  agent_tmp_dirs = {}
+
+  agents.each do |agent|
+    agent_tmp_dirs[agent.hostname] = agent.tmpdir(tmp_environment)
+  end
+
+  teardown do
+    step 'remove all test files on agents' do
+      agents.each { |agent| on(agent, "rm -r #{agent_tmp_dirs[agent.hostname]}", :accept_all_exit_codes => true) }
+    end
+    # note - master teardown is registered by #mk_tmp_environment_with_teardown
+  end
+
+  step 'create unicode source file served via module on master' do
+    # 静 \u9759 0xE9 0x9D 0x99 http://www.fileformat.info/info/unicode/char/9759/index.htm
+    # 的 \u7684 0xE7 0x9A 0x84 http://www.fileformat.info/info/unicode/char/7684/index.htm
+    # ☃ \2603 0xE2 0x98 0x83 http://www.fileformat.info/info/unicode/char/2603/index.htm
+    setup_module_on_master = <<-MASTER_MANIFEST
+      File {
+        ensure => directory,
+        mode => "0755",
+      }
+
+      file {
+        '#{environmentpath}/#{tmp_environment}/modules/utf8_file_module':;
+        '#{environmentpath}/#{tmp_environment}/modules/utf8_file_module/files':;
+      }
+
+      file { '#{environmentpath}/#{tmp_environment}/modules/utf8_file_module/files/\u9759\u7684':
+        ensure => file,
+        content => "\u2603"
+      }
+    MASTER_MANIFEST
+    apply_manifest_on(master, setup_module_on_master, :expect_changes => true)
+  end
+
+  step 'create a site.pp on master containing a unicode file resource' do
+    site_pp_contents = <<-SITE_PP
+      \$test_path = \$::fqdn ? #{agent_tmp_dirs}
+      file { "\$test_path/\uff72\uff67\u30d5\u30eb":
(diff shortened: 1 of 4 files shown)


MosesMendoza:
new acceptance test run on windows 2012r2:
https://gist.github.com/MosesMendoza/6ec45a440971f217077354e532fabba2

new acceptance test run on ubuntu 1604:
https://gist.github.com/MosesMendoza/a226063a3c4230f7f6887f6ee4b8a235

MosesMendoza:
not sure what that travis error is. likely this needs rebased though

acceptance/tests/resource/file/ticket_6448_file_with_utf8_source.rb
```diff
@@ -0,0 +1,82 @@
...
+        '#{environmentpath}/#{tmp_environment}/modules/utf8_file_module':;
+        '#{environmentpath}/#{tmp_environment}/modules/utf8_file_module/files':;
+      }
+      
+      file { '#{environmentpath}/#{tmp_environment}/modules/utf8_file_module/files/\u9759\u7684':
+        ensure => file,
+        content => "\u2603"
+      }
+    MASTER_MANIFEST
+    apply_manifest_on(master, setup_module_on_master, :expect_changes => true)
```
Iristyle: Is there not already a helper somewhere to do this / should we just provide some fixture files in the repo that get copied over?

Because Puppet is being used to create the files, it's (remotely) possible that Puppet misbehaves in the same way on both sides of the wire, and gets a false positive success.

There used to be a PMT test at `acceptance/tests/modules/install/long_path_name.rb` from https://github.com/puppetlabs/puppet/pull/4892/files that installed a real module.


I can go either way on this one - testing via manifest on the master could also help detect problems with writing utf-8 files / paths.

MosesMendoza: Good question. I looked in other acceptance tests and modeled this approach after the direct puppet tests, ex here: https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/direct_puppet/static_catalog_env_control.rb#L19-L154, which lay down the files in this manner rather than using fixtures. There's a tradeoff - I think there would be less complexity in this test if we could just copy the files over from fixtures, for sure. But I also think there's a downside, in that a "later-us" trying to understand the test potentially has to go searching around many places in the code to understand what it's doing, and when we set up shared contexts we have less flexibility to change them later as they may be used by other tests. There's probably an in-between that would suffice for this test, but I'm actually inclined to leave as-is, if not just so that others looking to write utf-8 file resource acceptance tests can see this as example content all in one place.

acceptance/tests/resource/file/ticket_6448_file_with_utf8_source.rb
```diff
@@ -0,0 +1,82 @@
...
+  tmp_environment = mk_tmp_environment_with_teardown(master, File.basename(__FILE__, '.*'))
+  agent_tmp_dirs = {}
+
+  agents.each do |agent|
+    agent_tmp_dirs[agent.hostname] = agent.tmpdir(tmp_environment)
+  end
+
+  teardown do
+    step 'remove all test files on agents' do
+      agents.each { |agent| on(agent, "rm #{agent_tmp_dirs[agent.hostname]}", :accept_all_exit_codes => true) }
```
MosesMendoza: I think this might need to be `rm -r` - checking

MosesMendoza: Yeah, it does. I didn't catch it in the run output linked above probably because it didn't fail the test.
```
* remove all test files on agents

  vki1nsgskq6wndd.delivery.puppetlabs.net (ubuntu1604-64-1) 12:01:44$ rm /tmp/ticket_6448_file_with_utf8_source_5gd3rak8.jL7rlB
    rm:     cannot remove '/tmp/ticket_6448_file_with_utf8_source_5gd3rak8.jL7rlB'    : Is a directory
```

MosesMendoza: Updated test. New run output showing deletion:
https://gist.github.com/MosesMendoza/6e524a5727c77c53844db0091d29216c

```
* remove all test files on agents

  qyx9n9wu09r2gn9.delivery.puppetlabs.net (ubuntu1604-64-1) 11:51:37$ rm -r /tmp/ticket_6448_file_with_utf8_source_z3vsg9lk.vrqzML

  qyx9n9wu09r2gn9.delivery.puppetlabs.net (ubuntu1604-64-1) executed in 0.01 seconds
```

acceptance/tests/resource/file/ticket_6448_file_with_utf8_source.rb
Iristyle: Sounds good to me - as-is it is.

Iristyle:
I see that you repushed the branch - you were able to get the test cleanup issues straightened out @MosesMendoza ?

MosesMendoza:
@Iristyle yes, sorry - I believe I've addressed all known issues

PR #5463, opened 2017-01-04
(PUP-6476) Add Hocon data_set function for Hiera version 5

thallgren:
This commit adds a 'hocon' feature to Puppet which gets enabled when the
hocon library is present. It also adds a data_set function named
`hocon_data` capable of producing a Hash from a file in hocon format. The
function can only execute if the 'hocon' feature is enabled.

The Hiera version 3 and version 4 configs are changed so that when the
'hocon' feature is enabled, the data provider will use the added function
rather than attempting to use a custom backend or a presumed data binding.


hlindberg:
A rebase is needed

PR #5464, opened 2017-01-04
(PUP-7064) Isolate Windows registry specs per run

Iristyle:
- Current registry specs share global state in the same location within
   the registry.  This can cause specs run in parallel on the same
   machine to intermittently fail.

   Move the subkey in use to a distinct location for each set of tests
   so that tests do not share global state.


MikaelSmith:
Failed an unrelated test. Rerunning.

PR #5466, opened 2017-01-05
(PUP-7067) Make hiera work with empty config file

thallgren:
This commit ensures that an empty hiera.yaml is recognized as a default
Hiera version 3 config. This is needed in order to be backward compatible
with Hiera 3.

### DIFF
--- a/lib/puppet/pops/lookup/hiera_config.rb
+++ b/lib/puppet/pops/lookup/hiera_config.rb
@@ -108,7 +108,10 @@ class HieraConfig
 
         # For backward compatibility, we must treat an empty file, or a yaml that doesn't
         # produce a Hash as Hiera version 3 default.
-        loaded_config = HieraConfigV3::DEFAULT_CONFIG_HASH unless loaded_config.is_a?(Hash)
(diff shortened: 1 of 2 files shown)


hlindberg:
ok to merge once tests go green

lib/puppet/pops/lookup/hiera_config.rb
```diff
@@ -105,6 +105,10 @@ def self.create(config_path)
       config_root = config_path.parent
       if config_path.exist?
         loaded_config = YAML.load_file(config_path)
+
+        # For backward compatibility, we must treat an empty file, or a yaml that doesn't
+        # produce a Hash as Hiera version 3 default.
+        loaded_config = HieraConfigV3::DEFAULT_CONFIG_HASH unless loaded_config.is_a?(Hash)
```
DavidS: Do we need to notify the user that we're ignoring the requested config file?

thallgren: Sounds like a good idea. I'll add a warning.

PR #5467, opened 2017-01-05
(PUP-7066) return and deprecate stack trace in http server error response

MosesMendoza:
In 7a765b7 and 5ae2d155 we unintentionally introduced a backwards-incompatible change, removing the "stacktrace" array from the HTTP Error object schema. We also added the stack trace information to the Puppet error log. This partially reverts commits 7a765b7 and 5ae2d155, returning the stack trace as formerly constituted, except that rather than return the actual stack trace content, we return a deprecation warning. This will allow users warning in case any systems consuming this data need to change to remain compatible. However, the reverts are partial in that we retain the previous addition of the stack trace to the puppet error log to avoid surprising anyone who had already begun to rely on this behavior and because with the stack trace string content removed, retaining the content for the logs will allow users to still find this information if they need it. Eventual removal of the property is ticketed as PUP-7070.

### DIFF
--- a/api/schemas/error.json
+++ b/api/schemas/error.json
@@ -13,7 +13,7 @@
             "type": "string"
         },
         "stacktrace": {
-            "description": "For 5xx responses only, a ruby stacktrace to where an error occurred.",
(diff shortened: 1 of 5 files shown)


MosesMendoza:
retargeting at LTS-1.7

PR #5468, opened 2017-01-05
(PUP-7066) return and deprecate stack trace in http server error response

MosesMendoza:
In 7a765b7 and 5ae2d155 we unintentionally introduced a backwards-incompatible change, removing the "stacktrace" array from the HTTP Error object schema. We also added the stack trace information to the Puppet error log. These commits were originally released in Puppet 4.8.0. There is a request to backport the intent of these fixes, that is to remove the stack trace information from the HTTP response, to the LTS-1.7 branch. This PR first cherry-picks back the original commits introduced in the 4.8.0 series, 7a765b7 and 5ae2d155. Since we have determined some of this change was backwards-incompatible, this PR includes partial reverts of these commits - returning the stack trace as formerly constituted, except that rather than return the actual stack trace content, we return a deprecation warning. This gives notice to users in case there are systems expecting this error format that need to change. The reverts are _partial_ in that we retain the previous addition of the stack trace to the puppet error log to avoid surprising anyone who had already begun to rely on this behavior in the 4.8 series and because with the stack trace string content removed, retaining the content for the logs will allow users to still find this information if they need it. Eventual removal of the property is ticketed as PUP-7070.

### DIFF
--- a/spec/unit/network/http/handler_spec.rb
+++ b/spec/unit/network/http/handler_spec.rb
@@ -94,10 +94,6 @@ describe Puppet::Network::HTTP::Handler do
       expect(res[:content_type_header]).to eq("application/json")
       expect(res_body["issue_kind"]).to eq(Puppet::Network::HTTP::Issues::RUNTIME_ERROR.to_s)
       expect(res_body["message"]).to eq("Server Error: the sky is falling!")
-
-      # Stactraces may contain sensitive information, returning them to API
-      # consumers is not a best practice. See
-      # https://tickets.puppetlabs.com/browse/PUP-6659
       expect(res_body["stacktrace"]).to be_nil
       expect(res[:status]).to eq(500)
     end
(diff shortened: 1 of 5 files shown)


spec/unit/network/http/handler_spec.rb
```diff
@@ -89,7 +95,7 @@ def respond(text)
       expect(res_body["issue_kind"]).to eq(Puppet::Network::HTTP::Issues::RUNTIME_ERROR.to_s)
       expect(res_body["message"]).to eq("Server Error: the sky is falling!")
       expect(res_body["stacktrace"].is_a?(Array) && !res_body["stacktrace"].empty?).to be_truthy
-      expect(res_body["stacktrace"][0]).to match("spec/unit/network/http/handler_spec.rb")
+      expect(res_body["stacktrace"][0]).to match(/The 'stacktrace' property is deprecated/)
```
MikaelSmith: Should we have positive confirmation somewhere that the stack trace isn't included?

MosesMendoza: 👍  - added

Iristyle:
For future ref @MosesMendoza I think it's preferential to revert an entire commit and then recreate desired changes in a new commit, as it can make it easier to work with gits tooling.  It can also be tricky to find changes hidden inside of revert commits if they are edited.

Sorry I didn't catch / mention this sooner.

MosesMendoza:
Makes sense @Iristyle. I was just trying to make the changes as streamlined as possible, but agree that this can obscure later reference. Thanks for the note.

PR #5470, opened 2017-01-08
(PUP-6240) Deprecate use of Puppet::SemVer

thallgren:
This commit ensures that a deprecation warning is logged when an instance
of Puppet::SemVer is created unless `Puppet[:strict] == :off`

### DIFF
--- a/spec/unit/semver_spec.rb
+++ b/spec/unit/semver_spec.rb
@@ -212,6 +212,17 @@ describe SemVer do
       expect(version.tiny).to    eq(5)
       expect(version.special).to eq('-beta6')
     end
+
+    it 'should not log a deprecation warning when strict == off' do
+      Puppet[:strict] = :off
+      SemVer.new('1.2.3')
+      expect(@logs).to be_empty
+    end
+
+    it 'should log a deprecation warning unless strict == off' do
+      SemVer.new('1.2.3')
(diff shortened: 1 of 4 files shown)


spec/unit/semver_spec.rb
```diff
@@ -212,6 +212,17 @@
...
       expect(version.special).to eq('-beta6')
     end
+
+    it 'should not log a deprecation warning when strict == off' do
+      Puppet[:strict] = :off
+      SemVer.new('1.2.3')
+      expect(@logs).to be_empty
+    end
+
+    it 'should log a deprecation warning unless strict == off' do
```
hlindberg: This actually tests that it does not log by default (since strict ==:warning by default). It does not check the strict == :warning and strict == :error cases.

thallgren: I don't think it needs to, as long as default isn't off (which it isn't). That's also what the title suggests. A warning is printed for both warning and error.

hlindberg: but it isn't testing that - if it should warn both in warning and error, then it should test both, now it is only testing :off and the default (warning) - that leaves the :error branch untested.

thallgren: True, on the other hand, there's no code introduced that is specific warning and error.
Neither the spec, the code, nor the test is concerned about warning or error. It strictly tests if strict is off, then issue no warning. If it isn't then warn. I fail to see what the problem is here.

PR #5474, opened 2017-01-09
(PUP-873) deprecate status face

MosesMendoza:
Deprecate the `puppet status` face.

Context from PUP-873/@zaphod42, suggesting this face is unusable:

  "The status face could possibly be useful, but the usability of it is so
  atrocious we are better removing it and revisiting it later."

Additional detail on the poor UX and broken-ness of this face can be found in
PUP-873.


MosesMendoza:

glennsarti:
Confirmed that the faces listed match up with the discussion in the attached tickets and spec tests validate the behaviour. As this is a deprecation message only, this is a low risk change.

PR #5475, opened 2017-01-09
(PUP-871) deprecate key face

MosesMendoza:
The key face is (self-described) not intended to be used - per PUP-871 it
should be removed.

From key.rb:
"it should not be necessary to use this subcommand directly"


MosesMendoza:

glennsarti:
Confirmed that the faces listed match up with the discussion in the attached tickets and spec tests validate the behaviour. As this is a deprecation message only, this is a low risk change. LGTM.

PR #5476, opened 2017-01-10
(PUP-7079) Retry scheduled task deletion in specs

Iristyle:
- When run concurrently, there may be file sharing issues when using
   the task scheduler APIs that prevent a task deletion from
   succeeding with an error like:

   Failed to call
   with HRESULT: -2147024864.:  The process cannot access the file
   because it is being used by another process.

 - Introduce a task deletion helper that retries up to 3 times with
   a 1 second delay in between attempts, in an effort to clean up
   tasks under such circumstances

### DIFF
--- a/spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
+++ b/spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
@@ -1954,6 +1954,20 @@ describe Puppet::Type.type(:scheduled_task).provider(:win32_taskscheduler), :if 
       end
     end
 
+    def delete_task_with_retry(task, name, attempts = 3)
+      failed = false
+
+      attempts.times do
+        begin
+          task.delete(name) if Win32::TaskScheduler.new.exists?(name)
+        rescue Puppet::Util::Windows::Error
+          failed = true
+        end
+
+        if failed then sleep 1 else break end
+      end
+    end
+
     describe '#exists?' do
       it 'works with Unicode task names' do
         task_name = name + "\u16A0\u16C7\u16BB" # ᚠᛇᚻ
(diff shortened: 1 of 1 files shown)


Iristyle:
This is not awesome, but I don't see another great alternative here.

Note that we remoted into the machine that intermittently failed, and it *did* have a leftover scheduled task with a GUID based name... so aside from not writing tests that actually create tasks, a defensive delete is probably the best approach.

MosesMendoza:
If we try and fail three times this will continue as success - previously it would fail. Is that desired? Not sure what the implications are of potentially leaving the tasks in place (other than modifying the test environment state)

Iristyle:
@MosesMendoza the inability to delete a task is not indicative of an actual test failure... but it could mean that a task has been orphaned on a spec box.

The question is - is it important enough to fail a run, as a flag that someone should manually clean up the clutter later, in the event that 3 attempts fail?  A solo orphaned task on a machine shouldn't be a problem (as its uniquely named) - but if there are hundreds or thousands of orphans, I'm not sure if there is an impact.

FWIW, we've only seen this problem once... and when I remoted into the machine in question, it only had a single orphaned task in task scheduler.

MosesMendoza:
So seems like probably not worth failing the run. In that case, could we rescue the specific exception that happens when the task delete fails? Or does it vary?

spec/unit/provider/scheduled_task/win32_taskscheduler_spec.rb
```diff
@@ -1954,6 +1954,20 @@ def time_component
       end
     end
 
+    def delete_task_with_retry(name, attempts = 3)
+      failed = false
+
+      attempts.times do
+        begin
+          task.delete(name) if Win32::TaskScheduler.new.exists?(name)
+        rescue Puppet::Util::Windows::Error
```
Iristyle: Good suggestion on specifying an error type @MosesMendoza - updated ^

MosesMendoza:
From the appveyor failure it looks like we might need to pass the task object into `delete_task_with_retry`

Iristyle:
> From the appveyor failure it looks like we might need to pass the task object into `delete_task_with_retry`

Well, that's embarrassing... Wasn't it green before?

PR updated

MosesMendoza:
Yeah I thought it was green too

MosesMendoza:
👍  will merge when specs go green again

MosesMendoza:
Since specs failed due to a reason this will ostensibly fix, going to go ahead and merge

PR #5477, opened 2017-01-10
(PUP-6930) Make DUPLICATE_DEFAULT an error

hlindberg:
Before this, a DUPLICATE_KEY was under the control of the `strict`
setting. This commit changes that to be a hard error.

### DIFF
--- a/spec/unit/pops/validator/validator_spec.rb
+++ b/spec/unit/pops/validator/validator_spec.rb
@@ -88,23 +88,19 @@ describe "validating 4x" do
     end
   end
 
-  [:off, :warning, :error].each do |level|
-    context "with --strict set to #{level}" do
-      before(:each) { Puppet[:strict] = level }
+  context 'irrespective of --strict' do
+    it 'produces an error for duplicate default in a case expression' do
+      acceptor = validate(parse('case 1 { default: {1} default : {2} }'))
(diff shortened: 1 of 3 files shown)


spec/unit/pops/validator/validator_spec.rb
```diff
@@ -128,19 +86,25 @@ def validate(model)
...
     end
+  end
 
-    it 'does not produce an error for duplicate default in a case expression' do
-      acceptor = validate(parse('case 1 { default: {1} default : {2} }'))
-      expect(acceptor.warning_count).to eql(0)
-      expect(acceptor.error_count).to eql(0)
-      expect(acceptor).to_not have_issue(Puppet::Pops::Issues::DUPLICATE_DEFAULT)
-    end
+  [:off, :warning, :error].each do |level|
```
thallgren: Why test this with different strict levels if the error isn't controlled by strict?

hlindberg: :-)

hlindberg: Because it shows that the change has effect - ie. this is a transition from being controlled by strict to not being controlled by strict.
If I remove that check there would be no assertion that this PR makes the required change.
Long term it is wasteful to keep that.

PR #5481, opened 2017-01-10
(PUP-869) Add face deprecation message to help summaries

glennsarti:
Previously in commit 7fefceae40d5ac a deprecation message was added to the face
and the help text for a face.  However when running `puppet help` it was not
obvious which faces were deprecated or not.

This commit adds the text `(Deprecated)` to the face summary when running
`puppet help`.

### DIFF
--- a/lib/puppet/face/help.rb
+++ b/lib/puppet/face/help.rb
@@ -146,7 +146,9 @@ MSG
       if (is_face_app?(appname))
         begin
           face = Puppet::Face[appname, :current]
-          result << [appname, face.summary]
+          # Add deprecation message to summary if the face is deprecated
+          summary = face.deprecated? ? face.summary + _(" (Deprecated)") : face.summary
+          result << [appname, summary]
         rescue StandardError, LoadError
           result << [ "! #{appname}", _("! Subcommand unavailable due to error. Check error logs.") ]
         end
--- a/spec/unit/face/help_spec.rb
+++ b/spec/unit/face/help_spec.rb
@@ -148,6 +148,16 @@ describe Puppet::Face[:help, '0.0.1'] do
     end
   end
 
+  context "#all_application_summaries" do
+    it "appends a deprecation warning for deprecated faces" do
+      # Stub the module face as deprecated
+      Puppet::Face[:module, :current].expects(:deprecated?).returns(true)
+      result = Puppet::Face[:help, :current].all_application_summaries.each do |appname,summary|
+        expect(summary).to match(/Deprecated/) if appname == 'module'
+      end
+    end
+  end
+
   context "#legacy_applications" do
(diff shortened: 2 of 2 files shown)


glennsarti:
Example output

```
PS> puppet help

Usage: puppet <subcommand> [options] <action> [options]

Available subcommands:

  agent             The puppet agent daemon
  apply             Apply Puppet manifests locally
  ca                Local Puppet Certificate Authority management. (Deprecated)
  catalog           Compile, save, view, and convert catalogs.
  cert              Manage certificates and requests
  certificate       Provide access to the CA for certificate management.
  certificate_request  Manage certificate requests. (Deprecated)
  certificate_revocation_list  Manage the list of revoked certificates. (Deprecated)
  config            Interact with Puppet's settings.
  describe          Display help about resource types
  device            Manage remote network devices
  doc               Generate Puppet references
  epp               Interact directly with the EPP template parser/renderer.
  facts             Retrieve and store facts.
...
```

branan:
Woah, did we manage to dig up a new spec order bug? That looks totally unrelated to this PR

glennsarti:
Err that looks odd.....

glennsarti:
@branan   Rekicked - All green!

lib/puppet/face/help.rb
```diff
@@ -146,7 +146,9 @@ def all_application_summaries()
       if (is_face_app?(appname))
         begin
           face = Puppet::Face[appname, :current]
-          result << [appname, face.summary]
+          # Add deprecation message to summary if the face is deprecated
+          summary = face.respond_to?(:deprecated?) && face.deprecated? ? face.summary + _(" (Deprecated)") : face.summary
```
joshcooper: Can we drop the `respond_to?` check necessary since https://github.com/puppetlabs/puppet/commit/7fefceae40d5ace8df2bd85aefa8bdee3246d9e1#diff-715723a3711cdfcf7cf7e524596200adR189?

branan: I think we ought to be able to drop it - any external faces should still be built on top of the `face_base` from the current puppet version.

Maybe if an updated version of the `help` face got pluginsync-ed to on old agent things would break?

glennsarti: I was concerned about the comment "both legacy and Face applications" so I was being a little cautious about the evaluation.

Can definitely drop it though.

MosesMendoza: should be fine to remove `respond_to?` - this file requires `puppet/face` which requires and then defines itself as `puppet/interface`, which publicly defines `deprecated?`.

glennsarti:
Added a unit test for this.  I'm not completely happy about how I did this as it will fail if the word Deprecated is localised outside of English.

PR #5483, opened 2017-01-10
(PUP-6997) acceptance: can put pp functions in $environment/functions

er0ck:
put this test in with the other test because this is where we first found the issue.

this change also removes the spurious environment, and creates a hiera5 3 level hierarchy

[skip ci]

### DIFF
--- a/acceptance/tests/lookup/lookup_rich_values.rb
+++ b/acceptance/tests/lookup/lookup_rich_values.rb
@@ -1,16 +1,14 @@
-test_name 'C99044 lookup should allow rich data as values' do
+test_name 'C99044: lookup should allow rich data as values' do
   require 'puppet/acceptance/environment_utils.rb'
   extend Puppet::Acceptance::EnvironmentUtils
 
   app_type        = File.basename(__FILE__, '.*')
   tmp_environment = mk_tmp_environment_with_teardown(master, app_type)
   fq_tmp_environmentpath  = "#{environmentpath}/#{tmp_environment}"
-  tmp_environment2 = mk_tmp_environment_with_teardown(master, app_type)
-  fq_tmp_environmentpath2  = "#{environmentpath}/#{tmp_environment2}"
 
   sensitive_value_rb = 'foot, no mouth'
   sensitive_value_pp = 'toe, no step'
-
+  sensitive_value_pp2 = 'toe, no module'
 
   step "create ruby lookup function in #{tmp_environment}" do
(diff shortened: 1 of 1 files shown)


er0ck:
this probably has to be re-targeted as well

acceptance/tests/lookup/lookup_rich_values.rb
```diff
@@ -31,44 +33,54 @@ def rich_data_test(options, context)
...
-      notify { "${unwrap(lookup('environment_key'))}": }
-    SITE
+  end
+
+  step "create puppet language lookup function in #{tmp_environment} module" do
+    on(master, "mkdir -p #{fq_tmp_environmentpath}/modules/some_mod/functions")
+    create_remote_file(master, "#{fq_tmp_environmentpath}/modules/some_mod/functions/rich_data_test2.pp", <<-FUNC)
+function some_mod::rich_data_test2($options, $context) {
+  {
+    "environment_key2" => Sensitive('#{sensitive_value_pp}'),
```
thallgren: This isn't supposed to work. All keys in a module must be prefixed with the module name. The reason it does work is that it is part of an environment hierarchy.
@hlindberg, what's your opinion on this. Should we allow the hierarchy in a hiera.yaml that resides in an environment to reference functions provided by a module?

acceptance/tests/lookup/lookup_rich_values.rb
```diff
@@ -20,6 +18,10 @@
 hierarchy:
   - name: Test
     data_hash: rich_data_test
+  - name: Test2
+    data_hash: some_mod::rich_data_test2
```
thallgren: This entry is somewhat unorthodox. Not sure it should be allowed since it references a function provided by a module.

acceptance/tests/lookup/lookup_rich_values.rb
hlindberg: yes, that must be allowed. That is how a module can make several different data sets available. The author of the environment data can then choose which one to use. That is an important feature.

acceptance/tests/lookup/lookup_rich_values.rb
hlindberg: See comment below.

hlindberg:
needs a rebase

er0ck:
rebased

hlindberg:
ping @thallgren ok to merge now? (ok for me)

PR #5484, opened 2017-01-10
(PUP-3534) Deprecate caching certificate authority verify methods

MosesMendoza:
This PR deprecates methods in `Puppet::SSL::CertificateAuthority` that were used by the PE License library (but no longer are). `#x509_store` (and its callers) in particular is problematic because it can cache the x509_store object once and never refresh it, even if it is no longer valid.

Commit outline:
> This commit deprecates Puppet::SSL::CertificateAuthority#list_certificates, as
> it is no longer used by anything. The PE License check code that originally
> used this been migrated to querying PuppetDB.

> To avoid superflous warnings emanating from use of
> Puppet::SSL::CertificateAuthority#list, that method is updated to directly do
> an indirector search instead of indirectly through #list_certificates. This is
> what it originally did, before #list_certificates was added, and is
> backwards-compatible.

> Puppet::SSL::CertificateAuthority#x509_store is a caching layer for the x509
> object that was used by the PE License code. This commit tags it as deprecated,
> as the PE License code no longer uses it, and it was questionably advisable
> back then to return a potentially stale x509 store anyway.
>
> We also update #verify to call #create_x509_store instead of #x509_store,
> skipping the cache layer and always creating a new object - the behavior before
> x509_store was introduced. This is semantically API backwards-compatible, and
> now we can remove #x509_store at a major version, as its only callers are also
> deprecated.
> 
> Note #x509_store is API private, so we don't actually do a
> Puppet.deprecation_warning. The comment tag should be enough to warn off anyone
> trying to use it before it's removed.

> Puppet::SSL::CertificateAuthority#certificate_is_alive? was introduced to
> leverage the #x509_store cached x509 object for the PE License code. It is no
> longer used, and #verify (or the Puppet Server API) should be used instead.

### DIFF
--- a/lib/puppet/ssl/certificate_authority.rb
+++ b/lib/puppet/ssl/certificate_authority.rb
@@ -420,6 +420,11 @@ class Puppet::SSL::CertificateAuthority
   #   of the X509 Store
   #
   # @return [OpenSSL::X509::Store]
+  #
+  # @deprecated Strictly speaking, #x509_store is marked API private, so we
+  #   don't need to publicly deprecate it. But it marked as deprecated here to
+  #   avoid the exceedingly small chance that someone comes in and uses it from
+  #   within this class before it is removed.
   def x509_store(options = {})
     if (options[:cache])
       return @x509store unless @x509store.nil?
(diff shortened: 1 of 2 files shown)


MosesMendoza:
ping @joshcooper for review as you raised this ticket originally

Iristyle:
Everything here looks good to me.  It would be good to note what version of PE (and corresponding PUP version) transitioned away from these code paths for historical reference.

MosesMendoza:
@Iristyle good call - updated commit messages to note that the license code using these methods was removed in the PE 3.1 series.

PR #5487, opened 2017-01-11
DO NOT MERGE (PUP-7008) acceptance: hiera v5 can use hiera v3 config and datafiles

er0ck:


er0ck:
i probably have to re-target this at stable

PR #5488, opened 2017-01-11
(PUP-7087) Improve lookup performance

thallgren:
This commit contains a series of performance tweeks that were motivated
by findings when using a profiler. The optimizations makes the test
`benchmark:hiera_global_lookup` execute ~20% faster.

1. The special `Hiera::Scope` was improved to use frozen, predefined
   arrays when checking if a key was one of the caller keys.
2. The `Puppet::Parser::Scope` was improved to use a predefined empty
   `Hash` rather than creating a new dynamic hash on every call, and
   to only check using `#include?` when a found value was `nil`
3. The FunctionProvider has a hash of contexts instead of a DataBinding
   with a similar hash. This is OK since the provider itself is bound by
   an adapter.
4. A check for stable interpolations of the configuration was improved.
   Instead of using `#each_pair.all?` on a `Hash` it simply uses `#all?`.
5. The Invocation, which is created once for each lookup call, now caches
   it's adapter instead of calling on `#adapt` to retreive it each time
   it's used.
6. More efficient check for terminus. In particular an expensive `#to_s`
   on a `Symbol` is now avoided.
7. The logic that splits keys on dots will now first use a very simple
   regexp to determine if quotes or dots are present and return a one
   element array directly if that's not the case. This saves lots of
   evaluations of the more complex regexp that is otherwise used.
8. The interpolation logic was improved so that `#exist?` is called only
   when the found value was `nil`.

Aside from the optimizations, this commit also fixes a minor flaw in the
Hiera::Scope where `#exist?` called on the parent scope `#include?`
insted of its `#exist?` method. The fix is included here since it was
needed in order do one of the optimizations.

### DIFF
--- a/lib/hiera/scope.rb
+++ b/lib/hiera/scope.rb
@@ -1,9 +1,12 @@
 class Hiera
   class Scope
-    CALLING_CLASS = "calling_class"
-    CALLING_CLASS_PATH = "calling_class_path"
-    CALLING_MODULE = "calling_module"
-    MODULE_NAME = "module_name"
+    CALLING_CLASS = 'calling_class'.freeze
+    CALLING_CLASS_PATH = 'calling_class_path'.freeze
+    CALLING_MODULE = 'calling_module'.freeze
+    MODULE_NAME = 'module_name'.freeze
+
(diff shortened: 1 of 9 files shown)


hlindberg:
good to merge when tests go green.

PR #5499, opened 2017-01-12
(maint) bump LTS version to 4.7.2

MosesMendoza:
This commit updates the version in anticipation of an upcoming 4.7.2 release.

### DIFF
--- a/lib/puppet/version.rb
+++ b/lib/puppet/version.rb
@@ -7,7 +7,7 @@
 
 
 module Puppet
-  PUPPETVERSION = '4.7.1'
+  PUPPETVERSION = '4.7.2'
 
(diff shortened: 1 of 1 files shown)


McdonaldSeanp:
👍

Magisus:
According to https://confluence.puppetlabs.com/display/PROD/puppet-agent+Versions+and+Dependencies the version is supposed to be 4.7.1... is that incorrect?

McdonaldSeanp:
True, the version has already been updated.

MosesMendoza:
Great catch. I knew we didn't ship 4.8.2, but hadn't realized 4.7.1 was in the same boat. Thanks

PR #5500, opened 2017-01-12
(PUP-6958) Fix gem packaging to properly include i18n files

Magisus:
This PR contains two commits, one of which ensures that the `locales` directory is properly packaged with the gem, and a second which removes the build-time dependency on the `gettext-setup` gem.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -104,7 +104,3 @@ task(:commits) do
   end
 end
 
-# Add i18n tasks (gettext:pot, gettext:po*)
-spec = Gem::Specification.find_by_name 'gettext-setup'
-load "#{spec.gem_dir}/lib/tasks/gettext.rake"
-GettextSetup.initialize(File.absolute_path('locales', File.dirname(__FILE__)))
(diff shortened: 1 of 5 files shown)


underscorgan:
👍 looks reasonable from an RE standpoint

Magisus:
Verified that it builds successfully (http://jenkins-release.delivery.puppetlabs.net/job/puppet-packaging-2017-01-12-11-43-41-8c89f5a2ee5c5025848790444492195277e2dc40/1/) and that the resulting gem contains the `locales` dir and runs correctly.

MosesMendoza:
reviewing

MosesMendoza:
Can we remove these two comments now? https://github.com/Magisus/puppet/blob/8c89f5a2ee5c5025848790444492195277e2dc40/lib/puppet.rb#L47-L56

Magisus:
Yes, it can just use the local path. Updated.

Iristyle:
👍

PR #5502, opened 2017-01-13
(PUP-6494) Support sensitive commands in Exec resource type.

peterhuene:
This commit adds support for sensitive commands in the `Exec` resource type:
`command` parameters that are specified as `Sensitive.new(...)` are now
properly redacted when the command fails.  This supports using data from lookup
and hiera.

Example:
```puppet
exec { 'redacted':
  command => Sensitive.new('/usr/bin/false SECRET_TEXT')
}
```

Output:
```
Error: [command redacted] returned 1 instead of one of [0]
Error: /Stage[main]/Main/Exec[redacted]/returns: change from notrun to 0 failed:
[command redacted] returned 1 instead of one of [0]
```

### DIFF
--- a/lib/puppet/provider/exec.rb
+++ b/lib/puppet/provider/exec.rb
@@ -8,6 +8,7 @@ class Puppet::Provider::Exec < Puppet::Provider
     output = nil
     status = nil
     dir = nil
+    sensitive = resource.parameters[:command].sensitive
 
     checkexe(command)
 
@@ -23,7 +24,7 @@ class Puppet::Provider::Exec < Puppet::Provider
 
     dir ||= Dir.pwd
 
-    debug "Executing#{check ? " check": ""} '#{command}'"
+    debug "Executing#{check ? " check": ""} '#{sensitive ? '[redacted]' : command}'"
     begin
       # Do our chdir
       Dir.chdir(dir) do
@@ -59,7 +60,8 @@ class Puppet::Provider::Exec < Puppet::Provider
           output = Puppet::Util::Execution.execute(command, :failonfail => false, :combine => true,
                                   :uid => resource[:user], :gid => resource[:group],
                                   :override_locale => false,
-                                  :custom_environment => environment)
+                                  :custom_environment => environment,
+                                  :sensitive => sensitive)
         end
         # The shell returns 127 if the command is missing.
         if output.exitstatus == 127
--- a/lib/puppet/type.rb
+++ b/lib/puppet/type.rb
@@ -1218,6 +1218,10 @@ class Type
     resource = Puppet::Resource.new(self, title)
     resource.catalog = hash.delete(:catalog)
 
+    if sensitive = hash.delete(:sensitive_parameters)
+      resource.sensitive_parameters = sensitive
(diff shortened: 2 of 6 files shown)


peterhuene:
Note: this currently still prints out the unredacted command in the debug output from the provider; it'll require a change to the provider's interaction from the type to fix.

MosesMendoza:
@peterhuene thanks for putting this together - was briefly looking at a similar scenario in the user type last week. I saw you removed (wip) from this - is it ready for review?

peterhuene:
Yes, I think so.  Currently the base provider for exec will debug print the command in the clear, and so will `Puppet::Util::Execution` under the covers.  I decided not to fix those places (would require a few API changes to make work) considering they are debug-logging only.

MosesMendoza:
@adrienthebo would you mind holding off on merging for a bit? I had started reviewing too, hope to get my head around then will click the button

adrienthebo:
@MosesMendoza happy to hold off on merging; I was going to defer anyways due to CI migrations and 4.9 being released at some point.

MosesMendoza:
@peterhuene I think this is good to go, but I wish we could also redact debug output.. I think it would be a better/less surprising user experience, considering with `--debug` we'll not only print the command but also write it into reports. I did a quick spike running sensitive through the provider, https://github.com/MosesMendoza/puppet/commit/da1a5825b9637f6a2b31e00cc9cca3e2d80de2bc - it doesn't seem like it would be too much work to get it all the way through (still needs the same windows).. Thoughts? I could split out a separate ticket for debug in the provider but it doesn't seem like too stark of an api change

peterhuene:
@MosesMendoza your changes look good.  I'll incorporate them into this PR.

adrienthebo:
I'm not a super huge fan of pushing the concept of sensitive data into the execution code; it might be necessary but in the long run sensitive parameters/properties/providers will be replaced with sensitive _aware_ parameters/properties/providers. The changes that @MosesMendoza suggested aren't too bad and should be easy to remove when the time comes but I want us to keep in mind the fact that we'll be replumbing this code at some point.

That being said it's probably better to do the right thing in the interim and go with this rather than wait for a future that might be farther out than I'd like. $0.02.

MosesMendoza:
@adrienthebo agreed on all counts, and a more robust solution certainly at least generalizes this redaction logic outside of the exec provider and util/execution - I definitely didn't spend of a ton of time working that up. That said, perhaps with the possible exception of adding the `:redacted` key to the execution options hash I think it's fairly innocuous and can be easily ripped out/cleaned up later. I'd be fine working up something more in-depth in the next few weeks with you too but I'm not sure what our timeline is on PUP-6494 either.. so perhaps as-is is fine. I could go either way. I did file https://tickets.puppetlabs.com/browse/PUP-7138 though so at least this need is captured for later.

peterhuene:
@adrienthebo @MosesMendoza I've updated the PR to redact the debug output and exception message.

peterhuene:
@adrienthebo @MosesMendoza any chance to review the updated PR?  Thanks!

PR #5504, opened 2017-01-13
(PUP-7100) Add Hiera version 5 lookup_key function for eyaml

thallgren:
This PR adds a Hiera version 5 style lookup_key function that relies on the functionality provided by the [hiera_eyaml](/TomPoulton/hiera-eyaml) gem. The function will refuse to execute unless this gem is installed (controlled by the added Puppet `hiera_eyaml` feature).
The lookup_key function will also be used if an eyaml backend is defined in a Hiera version 3 configuration.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,6 +63,7 @@ group(:development, :test) do
   gem 'webmock', '~> 1.24'
   gem 'vcr', '~> 2.9'
   gem "hocon", :require => false
+  gem "hiera-eyaml", :require => false
 end
 
 group(:development) do
--- a/lib/puppet/functions/eyaml_lookup_key.rb
+++ b/lib/puppet/functions/eyaml_lookup_key.rb
@@ -1,0 +1,73 @@
+# @since 5.0.0
+#
+Puppet::Functions.create_function(:eyaml_lookup_key) do
+  unless Puppet.features.hiera_eyaml?
+    raise Puppet::DataBinding::LookupError, 'Lookup using eyaml lookup_key function is only supported when the hiera_eyaml library is present'
+  end
+
+  require 'hiera/backend/eyaml/encryptor'
+  require 'hiera/backend/eyaml/utils'
+  require 'hiera/backend/eyaml/options'
+  require 'hiera/backend/eyaml/parser/parser'
+
+  dispatch :eyaml_lookup_key do
+    param 'String[1]', :key
+    param 'Hash[String[1],Any]', :options
+    param 'Puppet::LookupContext', :context
+  end
+
+  def eyaml_lookup_key(key, options, context)
+    return context.cached_value(key) if context.cache_has_key(key)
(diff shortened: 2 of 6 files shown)


hlindberg:
Waiting with merge until decision is made if this should go to Puppet 4.9.0 or 5.0.0

lib/puppet/functions/eyaml_lookup_key.rb
```diff
@@ -0,0 +1,73 @@
+# @since 4.9.0
```
hlindberg: This is not true if this ends up going into Puppet 5.0.0. Need to wait for decision.

Gemfile
```diff
@@ -63,6 +63,7 @@ group(:development, :test) do
   gem 'webmock', '~> 1.24'
   gem 'vcr', '~> 2.9'
   gem "hocon", :require => false
+  gem "hiera-eyaml", :require => false
```
hlindberg: We probably require a min version here... ?

lib/puppet/functions/eyaml_lookup_key.rb
```diff
@@ -0,0 +1,73 @@
...
+#
+Puppet::Functions.create_function(:eyaml_lookup_key) do
+  unless Puppet.features.hiera_eyaml?
+    raise Puppet::DataBinding::LookupError, 'Lookup using eyaml lookup_key function is not supported without hiera_eyaml? library'
```
hlindberg: "not supported without" is a double negation. There is also a '?' in the error message. Suggest rephrasing to something like "is only supported when the hiera_eyaml library is present"

hlindberg:
Do we need to make tests conditional on the feature being present? Ot should they simply fail if not installed using the gem spec?

Gemfile
thallgren: Or not. This is just for testing. It would be very nice if the tests fail when a new version comes out that isn't compatible.

er0ck:
why is this only for `lookup_key`?  is eyaml setup in such a way that it can't return a hash, like yaml can?

Gemfile
hlindberg: ok

hlindberg:
if it was data_hash it would need to decrypt all keys at once which would not be nice

er0ck:
not be nice, performance-wise?  or because it opens them to prying eyes, somehow?

hlindberg:
performance wise - you pay for the entire cost of decrypting all of the values on first lookup of anything. Except that it delivers values by key, there is no difference in the result. The only thing to remember is that the entry for eyaml is different. There will be an error if the wrong kind of function is used in the hiera.yaml.

hlindberg:
failure on app veyor for ruby 2.1 with jp locale... (hm)

er0ck:
cool, you can change target branches now

PR #5509, opened 2017-01-17
(PUP-5756) remove dependency on win32/eventlog

MosesMendoza:
Remove dependency on eventlog by including our own ffi definitions of the
required functions for puppet logging. Most simple implementation - write-only
support for event log.

API compatible with previous implementation.

Uses W (unicode) versions of API functions. The spec for these functions can be
found here:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa363654(v=vs.85).aspx

Replace existing log destination of win32/eventlog with
Puppet::Util::Windows::EventLog. This means we don't need to check for the
eventlog library any more as part of the eventlog feature. We keep the eventlog
feature for backwards compatibility for now, but it can be removed.

Previous implementation with Win32::EventLog actually (unnecessarily) created
two handles to the event log - one opened with source 'Application' which we
never wrote to, and a second opened with source 'Puppet', which we used. Now we
just create one handle, on initialization, with event source 'Puppet' (hence no
longer needing to supply this as a param in #report_event)

### DIFF
--- a/ext/windows/service/daemon.rb
+++ b/ext/windows/service/daemon.rb
@@ -4,13 +4,14 @@ require 'fileutils'
 require 'win32/daemon'
 require 'win32/dir'
 require 'win32/process'
-require 'win32/eventlog'
+
+# This file defines utilities for logging to eventlog. While it lives inside
+# Puppet, it is completely independent and loads no other parts of Puppet, so we
+# can safely require *just* it.
+require 'puppet/util/windows/eventlog'
 
 class WindowsDaemon < Win32::Daemon
   CREATE_NEW_CONSOLE          = 0x00000010
-  EVENTLOG_ERROR_TYPE         = 0x0001
-  EVENTLOG_WARNING_TYPE       = 0x0002
-  EVENTLOG_INFORMATION_TYPE   = 0x0004
 
   @run_thread = nil
   @LOG_TO_FILE = false
@@ -127,23 +128,16 @@ class WindowsDaemon < Win32::Daemon
         File.open(LOG_FILE, 'a:UTF-8') { |f| f.puts("#{Time.now} Puppet (#{level}): #{msg}") }
       end
 
-      case level
-        when :debug, :info, :notice
-          report_windows_event(EVENTLOG_INFORMATION_TYPE,0x01,msg.to_s)
-        when :err, :alert, :emerg, :crit
-          report_windows_event(EVENTLOG_ERROR_TYPE,0x03,msg.to_s)
-        else
-          report_windows_event(EVENTLOG_WARNING_TYPE,0x02,msg.to_s)
-      end
+      native_type, native_id = Puppet::Util::Windows::EventLog.to_native(level)
+      report_windows_event(native_type, native_id, msg.to_s)
     end
   end
 
   def report_windows_event(type,id,message)
     begin
       eventlog = nil
-      eventlog = Win32::EventLog.open("Application")
+      eventlog = Puppet::Util::Windows::EventLog.open("Puppet")
       eventlog.report_event(
-        :source      => "Puppet",
         :event_type  => type,   # EVENTLOG_ERROR_TYPE, etc
         :event_id    => id,     # 0x01 or 0x02, 0x03 etc.
         :data        => message # "the message"
--- a/lib/puppet/util/log/destinations.rb
+++ b/lib/puppet/util/log/destinations.rb
@@ -212,34 +212,28 @@ Puppet::Util::Log.newdesttype :array do
 end
 
 Puppet::Util::Log.newdesttype :eventlog do
+  # Leaving these in for backwards compatibility - duplicates the same in
+  # Puppet::Util::Windows::EventLog
   Puppet::Util::Log::DestEventlog::EVENTLOG_ERROR_TYPE       = 0x0001
   Puppet::Util::Log::DestEventlog::EVENTLOG_WARNING_TYPE     = 0x0002
   Puppet::Util::Log::DestEventlog::EVENTLOG_INFORMATION_TYPE = 0x0004
 
   def self.suitable?(obj)
-    Puppet.features.eventlog?
+    Puppet.features.microsoft_windows?
   end
 
   def initialize
-    @eventlog = Win32::EventLog.open("Application")
+    @eventlog = Puppet::Util::Windows::EventLog.open("Puppet")
   end
 
   def to_native(level)
-    case level
-    when :debug,:info,:notice
-      [self.class::EVENTLOG_INFORMATION_TYPE, 0x01]
-    when :warning
-      [self.class::EVENTLOG_WARNING_TYPE, 0x02]
-    when :err,:alert,:emerg,:crit
-      [self.class::EVENTLOG_ERROR_TYPE, 0x03]
-    end
+    Puppet::Util::Windows::EventLog.to_native(level)
   end
 
   def handle(msg)
     native_type, native_id = to_native(msg.level)
 
     @eventlog.report_event(
-      :source      => "Puppet",
       :event_type  => native_type,
       :event_id    => native_id,
       :data        => (msg.source and msg.source != 'Puppet' ? "#{msg.source}: " : '') + msg.to_s
--- a/lib/puppet/util/windows/eventlog.rb
+++ b/lib/puppet/util/windows/eventlog.rb
@@ -1,0 +1,192 @@
+require 'ffi'
+
+# Puppet::Util::Windows::EventLog needs to be requirable without having loaded
+# any other parts of Puppet so it can be leveraged independently by the code
+# that runs Puppet as a service on Windows.
+#
+# For this reason we:
+# - Define Puppet::Util::Windows
+# - Replicate logic that exists elsewhere in puppet/util/windows
+# - Raise generic RuntimeError instead of Puppet::Util::Windows::Error if its not defined
+module Puppet; module Util; module Windows ; end ; end ; end
+
+class Puppet::Util::Windows::EventLog
+  extend FFI::Library
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa363679(v=vs.85).aspx
+  EVENTLOG_ERROR_TYPE       = 0x0001
+  EVENTLOG_WARNING_TYPE     = 0x0002
+  EVENTLOG_INFORMATION_TYPE = 0x0004
+
+  # These are duplicate definitions from Puppet::Util::Windows::ApiTypes,
+  # established here so this class can be standalone from Puppet, and public so
+  # we can reference them in tests.
+  NULL_HANDLE = 0
+  WIN32_FALSE = 0
+
+  # Register an event log handle for the application
+  # @param source_name [String] the name of the event source to retrieve a handle for
+  # @return [void]
+  # @api public
+  def initialize(source_name = 'Puppet')
+    @eventlog_handle = RegisterEventSourceW(FFI::Pointer::NULL, wide_string(source_name))
+    if @eventlog_handle == NULL_HANDLE
+      raise EventLogError.new("RegisterEventSourceW failed to open Windows eventlog", FFI.errno)
+    end
+  end
+
+  # Close this instance's event log handle
+  # @return [void]
+  # @api public
+  def close
+    DeregisterEventSource(@eventlog_handle)
+  ensure
+    @eventlog_handle = nil
+  end
+
+  # Report an event to this instance's event log handle. Accepts a string to
+  #   report (:data => <string>) and event type (:event_type => FixNum) and id
+  # (:event_id => FixNum) as returned by #to_native. The additional arguments to
+  # ReportEventW seen in this method aren't exposed - though ReportEventW
+  # technically can accept multiple strings as well as raw binary data to log,
+  # we accept a single string from Puppet::Util::Log
+  #
+  # @param args [Hash{Symbol=>Object}] options to the associated log event
+  # @return [void]
+  # @api public
+  def report_event(args = {})
+    raise ArgumentError, "data must be a string, not #{args[:data].class}" unless args[:data].is_a?(String)
+    from_string_to_wide_string(args[:data]) do |message_ptr|
+      FFI::MemoryPointer.new(:pointer) do |message_array_ptr|
+        message_array_ptr.write_pointer(message_ptr)
+        user_sid = FFI::Pointer::NULL
+        raw_data = FFI::Pointer::NULL
+        raw_data_size = 0
+        num_strings = 1
+        eventlog_category = 0
+        report_result = ReportEventW(@eventlog_handle, args[:event_type],
+          eventlog_category, args[:event_id], user_sid,
+          num_strings, raw_data_size, message_array_ptr, raw_data)
+
+        if report_result == WIN32_FALSE
+          raise EventLogError.new("ReportEventW failed to report event to Windows eventlog", FFI.errno)
+        end
+      end
+    end
+  end
+
+  class << self
+    # Feels more natural to do Puppet::Util::Window::EventLog.open("MyApplication")
+    alias :open :new
+
+    # Query event identifier info for a given log level
+    # @param level [Symbol] an event log level
+    # @return [Array] Win API Event ID, Puppet Event ID
+    # @api public
+    def to_native(level)
+      case level
+      when :debug,:info,:notice
+        [EVENTLOG_INFORMATION_TYPE, 0x01]
+      when :warning
+        [EVENTLOG_WARNING_TYPE, 0x02]
+      when :err,:alert,:emerg,:crit
+        [EVENTLOG_ERROR_TYPE, 0x03]
+      else
+        raise ArgumentError, "Invalid log level #{level}"
+      end
+    end
+  end
+
+  private
+  # For the purposes of allowing this class to be standalone, the following are
+  # duplicate definitions from elsewhere in Puppet:
+
+  # If we're loaded via Puppet we should keep the previous behavior of raising
+  # Puppet::Util::Windows::Error on errors. If we aren't, at least concatenate
+  # the error code to the exception message to pass this information on to the
+  # user
+  if defined?(Puppet::Util::Windows::Error)
+    EventLogError = Puppet::Util::Windows::Error
+  else
+    class EventLogError < RuntimeError
+      def initialize(msg, code)
+        super(msg + " (Win32 error: #{code})")
+      end
+    end
+  end
+
+  # Private duplicate of Puppet::Util::Windows::String::wide_string
+  # Not for use outside of EventLog! - use Puppet::Util::Windows instead
+  # @api private
+  def wide_string(str)
+    # if given a nil string, assume caller wants to pass a nil pointer to win32
+    return nil if str.nil?
+    # ruby (< 2.1) does not respect multibyte terminators, so it is possible
+    # for a string to contain a single trailing null byte, followed by garbage
+    # causing buffer overruns.
+    #
+    # See http://svn.ruby-lang.org/cgi-bin/viewvc.cgi?revision=41920&view=revision
+    newstr = str + "\0".encode(str.encoding)
+    newstr.encode!('UTF-16LE')
+  end
+
+  # Private duplicate of Puppet::Util::Windows::ApiTypes::from_string_to_wide_string
+  # Not for use outside of EventLog! - Use Puppet::Util::Windows instead
+  # @api private
+  def from_string_to_wide_string(str, &block)
+    str = wide_string(str)
+    FFI::MemoryPointer.new(:uchar, str.bytesize) do |ptr|
+      # uchar here is synonymous with byte
+      ptr.put_array_of_uchar(0, str.bytes.to_a)
+
+      yield ptr
+    end
+
+    # ptr has already had free called, so nothing to return
+    nil
+  end
+
+  ffi_convention :stdcall
+
+  # The following are typedefs in Puppet::Util::Winodws::ApiTypes, but here we
+  # use their original FFI counterparts:
+  # :uintptr_t for :handle
+  # :int32 for :win32_bool
+  # :uint16 for :word
+  # :uint32 for :dword
+  # :pointer for :lpvoid
+  # :uchar for :byte
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa363678(v=vs.85).aspx
+  # HANDLE RegisterEventSource(
+  # _In_ LPCTSTR lpUNCServerName,
+  # _In_ LPCTSTR lpSourceName
+  # );
+  ffi_lib :advapi32
+  attach_function :RegisterEventSourceW, [:buffer_in, :buffer_in], :uintptr_t
+  private :RegisterEventSourceW
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa363642(v=vs.85).aspx
+  # BOOL DeregisterEventSource(
+  # _Inout_ HANDLE hEventLog
+  # );
+  ffi_lib :advapi32
+  attach_function :DeregisterEventSource, [:uintptr_t], :int32
+  private :DeregisterEventSource
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa363679(v=vs.85).aspx
+  # BOOL ReportEvent(
+  #   _In_ HANDLE  hEventLog,
+  #   _In_ WORD    wType,
+  #   _In_ WORD    wCategory,
+  #   _In_ DWORD   dwEventID,
+  #   _In_ PSID    lpUserSid,
+  #   _In_ WORD    wNumStrings,
+  #   _In_ DWORD   dwDataSize,
+  #   _In_ LPCTSTR *lpStrings,
+  #   _In_ LPVOID  lpRawData
+  # );
+  ffi_lib :advapi32
+  attach_function :ReportEventW, [:uintptr_t, :uint16, :uint16, :uint32, :pointer, :uint16, :uint32, :pointer, :pointer], :int32
+  private :ReportEventW
+end
--- a/spec/unit/util/windows/eventlog_spec.rb
+++ b/spec/unit/util/windows/eventlog_spec.rb
@@ -1,0 +1,123 @@
+#!/usr/bin/env ruby
+
+require 'spec_helper'
+
+require 'puppet/util/windows'
+
+describe Puppet::Util::Windows::EventLog, :if => Puppet.features.microsoft_windows? do
+
+  before(:each) { @event_log = Puppet::Util::Windows::EventLog.new }
+  after(:each) { @event_log.close }
+
+  describe "class constants" do
+    it "should define NULL_HANDLE as 0" do
+      expect(Puppet::Util::Windows::EventLog::NULL_HANDLE).to eq(0)
+    end
+
+    it "should define WIN32_FALSE as 0" do
+      expect(Puppet::Util::Windows::EventLog::WIN32_FALSE).to eq(0)
+    end
+  end
+
+  describe "self.open" do
+    it "sets a handle to the event log" do
+      default_name = Puppet::Util::Windows::String.wide_string('Puppet')
+      # return nil explicitly just to reinforce that we're not leaking eventlog handle
+      Puppet::Util::Windows::EventLog.any_instance.expects(:RegisterEventSourceW).with(nil, default_name).returns(nil)
+      Puppet::Util::Windows::EventLog.new
+    end
+
+    context "when it fails to open the event log" do
+      before do
+        # RegisterEventSourceW will return NULL on failure
+        # Stubbing prevents leaking eventlog handle
+        Puppet::Util::Windows::EventLog.any_instance.stubs(:RegisterEventSourceW).returns(Puppet::Util::Windows::EventLog::NULL_HANDLE)
+      end
+
+      it "raises an exception warning that the event log failed to open" do
+        expect { Puppet::Util::Windows::EventLog.open('foo') }.to raise_error(Puppet::Util::Windows::EventLog::EventLogError, /failed to open Windows eventlog/)
+      end
+
+      it "passes the exit code to the exception constructor" do
+        fake_error = Puppet::Util::Windows::EventLog::EventLogError.new('foo', 87)
+        FFI.stubs(:errno).returns(87)
+        # All we're testing here is that the constructor actually receives the exit code from FFI.errno (87)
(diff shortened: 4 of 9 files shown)


MosesMendoza:
TODO: Need to figure out how to handle https://github.com/puppetlabs/puppet/blob/master/ext/windows/service/daemon.rb - the daemon code also uses `win32/eventlog`. It effectively duplicated the logic in destinations.rb, calling out to the various methods for log setup and reporting. It would be nice to avoid duplicating what's been added as `puppet/util/windows/eventlog` here in that file too but I'm not sure the best way to do that. Hence this PR keeps the gem dep on win32-eventlog until we get daemon.rb sorted out, however that doesn't necessarily need to block this one.

Iristyle:
Yeah @MosesMendoza if you need code in the service daemon *and* Puppet you'll need to basically build out something standalone that takes no dependencies on other Puppet code (like the FFI aliases in `api_types.rb`).  Based on how few API calls we have I think the scope is limited enough that that should be fairly doable (even if it leads to a small bit of duplication).

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,114 @@
+require 'puppet/util/windows'
```
Iristyle: The simplest way of making this consumable inside the service daemon is to start by removing the above `require 'puppet/util/windows'` and see what breaks.

Things I know of off the top of my head:

* You'll need an in-file version of `Puppet::Util::Windows::String.wide_string` - which is pretty easy
* `FFI::Pointer::NULL_HANDLE` comes from `api_types.rb`
* `Puppet::Util::Windows::Error` will have to be replaced OR we might be able to make that one usable within the service daemon as well as it's quite small and self-contained
* `FFI::WIN32_FALSE` comes from `api_types.rb`
* `FFI::MemoryPointer.from_string_to_wide_string` comes from `api_types.rb`
* `attach_function_private` comes from `api_types.rb`
* a bunch of the aliases in the API prototypes come from `api_types.rb`


It's also possible that we can make both `api_types.rb`and `windows/string.rb` consumable without the rest of Puppet (`string.rb` currently requires `puppet/util/windows` but I'm not sure why)

Food for thought anyhow...

MosesMendoza: @Iristyle yeah I think basically just redefining some of the FFI constants/methods, errors, and duplicate `wide_string` method would make it independent of puppet code. But then are you suggesting the same source file would somehow be leveraged by both daemon.rb and puppet, or just to copy the necessary code into daemon.rb? It seems to me it would be unusual to have daemon.rb use/know where eventlog.rb lives in puppet so I imagine the latter?

Iristyle: Might have to dig into to it to understand what it's capable of easily loading without a lot of hackery.  Obviously the current daemon is loading gems (i.e. `win32-eventlog` / `win32-service`) using `require` so it has access to the same Ruby env that Puppet has.

I think that means is could theoretically `require 'puppet/util/windows/api_types.rb'` and share the same files, but not sure if that's confusing to maintain.  I think we don't do a `require 'puppet'` because it drags along a bunch of startup initialization / settings parsing that clearly isn't needed for a simple service runner.  But it's possible we could load just a few files we need (as long as they're isolated, and we can make sure through tests that the service doesn't load more code than necessary).

MosesMendoza: I think I'm going to move toward trying to make this file stand alone / completely independent of puppet, and then try requiring it directly from daemon.rb. Not ideal - I dislike having external things requiring puppet's internals (bit us recently with the whole vendor/semantic thing) but since this daemon.rb does live in puppet itself and is exercised indirectly by our acceptance tests I expect any regression would be caught that way.

MosesMendoza:
@Iristyle took a stab at removing all use of puppet internals in this, and confirmed that this can post an event both from within puppet and externally, ie.
```ruby
C:\Users\moses\development\puppet(PUP-5756/master/remove_dependency_on_eventlog)> pry
[1] pry(main)> require 'puppet/util/windows/eventlog'
=> true
[2] pry(main)> log = Puppet::Util::Windows::EventLog.new("puppet")
=> #<Puppet::Util::Windows::EventLog:0x000000061bf3e0 @eventlog_handle=117178376>
[3] pry(main)> log.report_event({:data => 'foo', :event_type => Puppet::Util::Windows::EventLog::EVENTLOG_WARNING_TYPE, :event_id => 0x02})
=> nil
```

Definitely quite a bit of duplication, but I think with this we can actually remove win32-eventlog everywhere.

MosesMendoza:
I think this can be merged independently of removing win32-eventlog from daemon.rb and removing the gem dependency, so its good for review now. I will continue to work on those and will file separate PRs if this gets merged first.

MosesMendoza:
This should have an acceptance test run before merge

MosesMendoza:
ping @joshcooper for review

ext/windows/service/daemon.rb
```diff
@@ -127,23 +128,16 @@ def log(msg, level)
...
+      native_type, native_id = Puppet::Util::Windows::EventLog.to_native(level)
+      report_windows_event(native_type, native_id, msg.to_s)
     end
   end
 
   def report_windows_event(type,id,message)
     begin
       eventlog = nil
-      eventlog = Win32::EventLog.open("Application")
+      eventlog = Puppet::Util::Windows::EventLog.open("Puppet")
```
joshcooper: Previously we opened the `Application` event log and each log event had a `source` of `Puppet`. Is that still the case, or are we trying to open the `Puppet` event log?

lib/puppet/util/log/destinations.rb
```diff
@@ -212,34 +212,28 @@ def handle(msg)
...
   Puppet::Util::Log::DestEventlog::EVENTLOG_INFORMATION_TYPE = 0x0004
 
   def self.suitable?(obj)
-    Puppet.features.eventlog?
+    Puppet.features.microsoft_windows?
   end
 
   def initialize
-    @eventlog = Win32::EventLog.open("Application")
+    @eventlog = Puppet::Util::Windows::EventLog.open("Puppet")
```
joshcooper: Similar comment about `Application` vs `Puppet`

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+        user_sid, raw_data = nil
+        raw_data_size = 0
+        num_strings = 1
+        eventlog_category = 0
+        report_result = ReportEventW(@eventlog_handle, args[:event_type],
+          eventlog_category, args[:event_id], user_sid,
+          num_strings, raw_data_size, message, raw_data)
+
+        if report_result == WIN32_FALSE
+          raise EventlogError.new("failed to report event to Windows eventlog")
```
joshcooper: Will raising an exception cause us to try to write an event, and raise an exception, etc?

ext/windows/service/daemon.rb
MosesMendoza: Yeah that one mystified me for a bit as I traced through win32/eventlog and the MSDN function reference, which is murky at best on this topic. Turns out, we've been a bit suboptimal I think. What `Win32::EventLog` does when we call `Win32::EventLog.open("Application")` is open a handle via [`OpenEventLog`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363672(v=vs.85).aspx) with an event source parameter of "Application" https://github.com/djberg96/win32-eventlog/blob/ffi2/lib/win32/eventlog.rb#L94-L128. According to the MSDN docs, this "Opens a handle to the specified event log." and in the notes:
```
lpSourceName [in]
  The name of the log.
    If you specify a custom log and it cannot be found, the event logging service opens the Application log; however, there will be no associated message or category string file.
```

Seems pretty reasonable so far. However, later, when we actually write the log entry via `Win32::EventLog#report_event`, what that method effectively does is open a _new_ event log handle with the newly supplied source parameter value (in our case, "Puppet"). It does so via [`RegisterEventSource`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363678(v=vs.85).aspx) (https://github.com/djberg96/win32-eventlog/blob/ffi2/lib/win32/eventlog.rb#L668). The previous handle that was opened by `Win32::EventLog::open` is never used by `Win32::EventLog#report_event`. Wha? Well according to the docs, `RegisterEventSource` "Retrieves a registered handle to the specified event log." In order to be a registered handle, the event source needs to have been previously added to the registry (which I believe we actually do with our MSI). Side note, this:

```
Remarks

If the source name cannot be found, the event logging service uses the Application log. Although events will be reported , the events will not include descriptions because there are no message and category message files for looking up descriptions related to the event identifiers.
```
That fallback is what was previously happening when we ran via gem or source (and is still the case after my PR).

The example C++ code here: https://msdn.microsoft.com/en-us/library/windows/desktop/aa363680(v=vs.85).aspx only calls `RegisterEventSource` to get the handle, and then report events through it - we don't need call `OpenEventSource`. In my other searches online this seemed to be consistent.

In summary, our use of the Win32 code was obtaining an extra initial event log handle (the "Application" handle) and then never doing anything with it. So my PR just skips that step entirely and gets the initial handle via `RegisterEventSource` with the source name, "Puppet", and reports through that handle.

lib/puppet/util/windows/eventlog.rb
MosesMendoza: I don't believe so. Instantiating and raising a Puppet::Util::Windows::Error by itself doesn't trigger an event log write. A quick test confirmed no infinite loop.

```diff
diff --git a/lib/puppet/util/windows/eventlog.rb b/lib/puppet/util/windows/eventlog.rb
index ead6740..29e52af 100644
--- a/lib/puppet/util/windows/eventlog.rb
+++ b/lib/puppet/util/windows/eventlog.rb
@@ -59,9 +59,9 @@ class Puppet::Util::Windows::EventLog
         raw_data_size = 0
         num_strings = 1
         eventlog_category = 0
-        report_result = ReportEventW(@eventlog_handle, args[:event_type],
-          eventlog_category, args[:event_id], user_sid,
-          num_strings, raw_data_size, message, raw_data)
+        report_result = WIN32_FALSE # ReportEventW(@eventlog_handle, args[:event_type],
+          # eventlog_category, args[:event_id], user_sid,
+          # num_strings, raw_data_size, message, raw_data)

         if report_result == WIN32_FALSE
           raise EventlogError.new("failed to report event to Windows eventlog")
```

```ruby
[1] pry(main)> development\puppet(PUP-5756/master/remove_dependency_on_eventlog)> be pry
[2] pry(main)> require 'puppet'
=> true
[3] pry(main)> Puppet::Util::Log.setup_default
=> #<Puppet::Util::Log::DestEventlog:0x00000006748190 @eventlog=#<Puppet::Util::Windows::EventLog:0x00000006748118 @eventlog_handle=151388168>>
[4] pry(main)> Puppet.warning("foo")
Puppet::Util::Windows::Error: failed to report event to Windows eventlog:  The operation completed successfully.
from C:/Users/moses/development/puppet/lib/puppet/util/windows/eventlog.rb:67:in `block (2 levels) in report_event'
[5] pry(main)>
```

Looking up the call list through Puppet::Util::Log and Puppet::Util::Logging, it doesn't appear a failure to log an event generates a new log entry regarding that failure. A failure to set up the logging destination does (via `Puppet.log_exception`, [here](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/log.rb#L145-L150), which will be triggered if we fail to open the eventlog handle. With no destination, that message will get queued. However that's in a rescue that also ensures the console destination is added via `newdestination` after we `log_exception` which will flush the queue and log that to the console. So I think we're ok.

MosesMendoza:
rebased against latest puppet/master

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+  # duplicate definitions from elsewhere in Puppet:
+  NULL_HANDLE = 0
+  WIN32_FALSE = 0
+
+  # If we're loaded via Puppet we should keep the previous behavior of raising
+  # Puppet::Util::Windows::Error on errors. For daemon.rb we don't have
+  # Puppet::Util::Windows::Error, but daemon.rb just silently blanket rescues
+  # all Exceptions raised during a log attempt, so we're fine just raising a
+  # generic RuntimeError.
+  EventlogError = defined?(Puppet::Util::Windows::Error) ? Puppet::Util::Windows::Error : RuntimeError
```
Iristyle: `Puppet::Util::Windows::Error` has the nice side effect of calling `FormatMessage` to get the translated string value automatically based on the result of the last Windows API call made through FFI (exposed as `FFI.errno`, but really an automatic call to `GetLastError` API).  

It looks like the original `report_event` at least captured `FFI.errno`, which is useful for debugging things that have gone wrong - https://github.com/djberg96/win32-eventlog/blob/win32-eventlog-0.6.5/lib/win32/eventlog.rb#L669

I think we're OK without `FormatMessage` support, but if you wanted to re-implement inside this standalone helper it would look like this - https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/error.rb (I'm fine without it for now given we're already duplicating other code in here)

MosesMendoza: Just to make sure I understand, the concern about not capturing `FFI.errno` / `FormatMessage` would only apply to situations where `Puppet::Util::Windows::Error` is not defined/available (since we'll use it if it is)? If that's the case, I think I'm fine with leaving it a generic runtime error when `Puppet::Util::Windows::Error` isn't available. The reason being, the only place (at this time) outside of Puppet where we use this code is in `ext/windows/services/daemon.rb`, for running the puppet service itself, and `daemon.rb` just entirely ignores any failure to log something: https://github.com/puppetlabs/puppet/blob/master/ext/windows/service/daemon.rb#L151-L153. If we build on the use cases in the future I'd be +1 to extending the error handling.

MosesMendoza:
Just a reminder I need to kick off an ad-hoc acceptance run w/these changes to ensure acceptance doesn't break before merge

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+  #
+  # @param args [Hash{Symbol=>Object}] options to the associated log event
+  # @return [void]
+  # @api public
+  def report_event(args = {})
+    raise ArgumentError, "data must be a string, not #{args[:data].class}" unless args[:data].is_a?(String)
+    from_string_to_wide_string(args[:data]) do |message_ptr|
+      FFI::MemoryPointer.new(:pointer, 2) do |message|
+        message[0].write_pointer(message_ptr)
+        user_sid, raw_data = nil
```
Iristyle: Might be a stylistic thing, but I typically use `FFI::Pointer::NULL` for passing null to the Win32 APIs

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+  # );
+  ffi_lib :advapi32
+  attach_function :RegisterEventSourceW, [:buffer_inout, :buffer_inout], :uintptr_t
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa363642(v=vs.85).aspx
+  # BOOL DeregisterEventSource(
+  # _Inout_ HANDLE hEventLog
+  # );
+  ffi_lib :advapi32
+  attach_function :DeregisterEventSource, [:uintptr_t], :int32
```
Iristyle: Function prototype 👍

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+  # :pointer for :lpvoid
+  # :uchar for :byte
+
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa363678(v=vs.85).aspx
+  # HANDLE RegisterEventSource(
+  # _In_ LPCTSTR lpUNCServerName,
+  # _In_ LPCTSTR lpSourceName
+  # );
+  ffi_lib :advapi32
+  attach_function :RegisterEventSourceW, [:buffer_inout, :buffer_inout], :uintptr_t
```
Iristyle: Because these are `const` - i.e. LP**C**TSTR, I believe you should be using `:buffer_in` rather than `:buffer_inout` - https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/api_types.rb#L137

spec/unit/util/windows/eventlog_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      default_name = Puppet::Util::Windows::String.wide_string('Puppet')
+      # return nil explicitly just to reinforce that we're not leaking eventlog handle
+      Puppet::Util::Windows::EventLog.any_instance.expects(:RegisterEventSourceW).with(nil, default_name).returns(nil)
+      Puppet::Util::Windows::EventLog.new
+    end
+
+    it "raises an exception if the event log handle is not opened" do
+      # RegisterEventSourceW will return NULL on failure
+      # Stubbing prevents leaking eventlog handle
+      Puppet::Util::Windows::EventLog.any_instance.stubs(:RegisterEventSourceW).returns(FFI::Pointer::NULL_HANDLE)
```
Iristyle: This is really minor nit, but `FFI::Pointer::NULL_HANDLE` ... or our embedded `NULL_HANDLE`?

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+        raise ArgumentError, "Invalid log level #{level}"
+      end
+    end
+  end
+
+  private
+  # For the purposes of allowing this class to be standalone, the following are
+  # duplicate definitions from elsewhere in Puppet:
+  NULL_HANDLE = 0
+  WIN32_FALSE = 0
```
Iristyle: Make constants `public` so that they can be used in tests of this class... or leave as private detail?

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+  #   _In_ WORD    wCategory,
+  #   _In_ DWORD   dwEventID,
+  #   _In_ PSID    lpUserSid,
+  #   _In_ WORD    wNumStrings,
+  #   _In_ DWORD   dwDataSize,
+  #   _In_ LPCTSTR *lpStrings,
+  #   _In_ LPVOID  lpRawData
+  # );
+  ffi_lib :advapi32
+  attach_function :ReportEventW, [:uintptr_t, :uint16, :uint16, :uint32, :pointer, :uint16, :uint32, :buffer_inout, :pointer], :int32
```
Iristyle: I think for the `LPCTSTR *` we actually want a `:pointer` rather than a `:buffer_inout`

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+    end
+  end
+
+  # Close this instance's event log handle
+  # @return [void]
+  # @api public
+  def close
+    result_of_close = DeregisterEventSource(@eventlog_handle)
+    if result_of_close == WIN32_FALSE
+      raise EventlogError.new("failed to close Windows eventlog")
```
Iristyle: I don't think we want to raise here ... just log the error and move on.

Iristyle: I don't see anywhere in the code that we actually call close?  So I have a few thoughts on how you might handle this:

* Register some sort of `at_exit` handler like we do in TaskScheduler - https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/taskscheduler.rb#L194-L200 - noting the problems with `at_exit`.  For one, you can't guarantee ordering of `at_exit` handlers IIRC, so it's plausible that something could still try to use the event log after the handle has been closed. If the code was made more defensive, we could prevent that from being a problem *BUT* we might mysteriously lose event log messages, which could be tricky to diagnose.
* Don't store the handle in `@eventlog_handle` at all - instead grab a new one before each logging attempt, and  then close it immediately after use.  There will be some additional overhead, but given our limited usage of `report_event`, this doesn't seem so terrible to me.
* Don't worry about ever calling `close` - when the process closes it will eventually close down any open kernel objects (including handles), and I don't think there will be any real-world negative consequences.

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+  # technically can accept multiple strings as well as raw binary data to log,
+  # we accept a single string from Puppet::Util::Log
+  #
+  # @param args [Hash{Symbol=>Object}] options to the associated log event
+  # @return [void]
+  # @api public
+  def report_event(args = {})
+    raise ArgumentError, "data must be a string, not #{args[:data].class}" unless args[:data].is_a?(String)
+    from_string_to_wide_string(args[:data]) do |message_ptr|
+      FFI::MemoryPointer.new(:pointer, 2) do |message|
```
Iristyle: I think you're allocating one too many pointers here given there's only one string... should be this I believe:

```ruby
FFI::MemoryPointer.new(:pointer) do |message_array_ptr|
```

Note API def (which as noted elsewhere should be `:pointer` and not `:buffer_inout` in the prototype):

```
_In_ LPCTSTR *lpStrings,

lpStrings [in]
A pointer to a buffer containing an array of null-terminated strings that are merged into the message before Event Viewer displays the string to the user. This parameter must be a valid pointer (or NULL), even if wNumStrings is zero. Each string is limited to 31,839 characters.
Prior to Windows Vista:  Each string is limited to 32K characters.
```

lib/puppet/util/windows/eventlog.rb
```diff
@@ -0,0 +1,180 @@
...
+        message[0].write_pointer(message_ptr)
```
Iristyle: Based on above comment, I think you can remove the array indexing here and just do a:

```ruby
message_array_ptr.write_pointer(message_ptr)
```

But I haven't validated this for certain...

Iristyle:
@MosesMendoza I'll scope out the tests in more detail once you've made the FFI changes.

lib/puppet/util/windows/eventlog.rb
Iristyle: I see what you're saying.  But a couple of counterpoints:

* Is that behavior guaranteed in the future? (i.e. that there's only one place that will catch this `RuntimeError` and that it will just eat it)
* Should we at least be writing to `$STDERR` or something on failure as a last ditch effort there?  Otherwise we completely fail to provide a user with any useful feedback about the failure to log to the event log (plausibly a corrupt event log or perms issues could cause something like this)...

lib/puppet/util/windows/eventlog.rb
MosesMendoza: 👍 . I think they're equivalent in FFI land but happy to update for consistency

lib/puppet/util/windows/eventlog.rb
MosesMendoza: 👍

lib/puppet/util/windows/eventlog.rb
MosesMendoza: I believe we call `close` here from the `:eventlog` destination type's `close` method [defined here](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/log/destinations.rb#L249-L254). Each destination type optionally defines `close`. This is in turn called by [`Puppet::Util::Log::close`](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/log.rb#L52-L60), used by various methods in that class.

I agree we probably always want to set the handle to nil - given what you say about minimal real-world consequences I'm inclined to remove the `raise` altogether and instead do something like
```ruby
begin
  DeregisterEventSource(@eventlog_handle)
ensure
  @eventlog_handle = nil
end
```
thoughts?

spec/unit/util/windows/eventlog_spec.rb
MosesMendoza: Ah nice catch. More likely the latter.

lib/puppet/util/windows/eventlog.rb
MosesMendoza: I'm so wary of letting new public API out into the universe...but in this case it seems relatively harmless to establish these publicly.

lib/puppet/util/windows/eventlog.rb
MosesMendoza: Ah you're right. Thanks for the catch. Not sure why I was doubling the size there. Maybe leftover from previous experimentation or misinterpretation of the function reference.

lib/puppet/util/windows/eventlog.rb
MosesMendoza: I think you're right, since `write_pointer(ptr)` is effectively `put_pointer(0, ptr)`, ie write to your address with 0 offset, and index 0 is just the pointer's starting address.

lib/puppet/util/windows/eventlog.rb
MosesMendoza: Agreed. Updated to leverage `SystemCallError` when `Puppet::Util::Windows::Error` is unavailable. `SystemCallError` is what `win32-eventlog` uses and it accepts an exit code parameter in its constructor so we can leverage its error message mappings. Also updated exception message to include exit code.

MosesMendoza:
@Iristyle responded to all feedback - ready for review. Will squash before merge

lib/puppet/util/windows/eventlog.rb
```diff
@@ -158,13 +158,15 @@ def from_string_to_wide_string(str, &block)
   # );
   ffi_lib :advapi32
   attach_function :RegisterEventSourceW, [:buffer_in, :buffer_in], :uintptr_t
+  private :RegisterEventSourceW
```
Iristyle: 👍 

Equivalent to our `attach_function_private` helper at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/api_types.rb#L12-L18 (which you don't have access to)

lib/puppet/util/windows/eventlog.rb
```diff
@@ -31,7 +31,7 @@ class Puppet::Util::Windows::EventLog
   def initialize(source_name = 'Puppet')
     @eventlog_handle = RegisterEventSourceW(FFI::Pointer::NULL, wide_string(source_name))
     if @eventlog_handle == NULL_HANDLE
-      raise EventlogError.new("failed to open Windows eventlog")
+      raise EventLogError.new("RegisterEventSourceW failed to open Windows eventlog (exit code #{FFI.errno})", FFI.errno)
```
Iristyle: This seemed like a great solution - but then I tried something out, and I don't think it will quite work out.  Here's why:

`SystemCallError` attempts to look up the given error code and translate it to something in the Ruby library, which won't make sense given Win32 error codes don't map to Ruby errors.  Assume the Win32 error was `5`, which is `Access Denied` in Windows.

This is what Ruby does if you pass the value `5` through to the `SystemCallError` initializer:

```
[1] pry(main)> raise SystemCallError.new('hi', 5)
Errno::EIO: Input/output error - hi
from (pry):1:in `__pry__'
```

Iristyle: What if we were to just move the logic for the error generation to a method (instead of aliasing the type) - like

```ruby
def new_eventlog_error(msg, code = FFI.errno)
  defined?(Puppet::Util::Windows::Error) ?
    Puppet::Util::Windows::Error.new(msg) :
    SystemCallError.new(msg + " (Win32 error: #{code})")
end
```

And then the callers can just do this:

```ruby
raise new_eventlog_error('something bad happened')
```

MosesMendoza: Ha. Nice. So the mappings are just wrong on Windows..
```
However, operating systems typically report errors using plain integers. Module Errno is created dynamically to map these operating system errors to Ruby classes, with each error number generating its own subclass of SystemCallError
```
...from https://ruby-doc.org/core-2.2.0/Errno.html

Oh well. In that case there are some options, but they're not as clean. A couple include...write format error support. A really terrible option is since RuntimeException (nor any of the other exception class I glanced at) take an error code in the constructor, define EventLogError as an exception class that ignores the second parameter i.e., 
class EventLogError < RuntimeError
  def initialize(msg, code)
    super(msg)
  end
end

I'm leaning towards format error support but we're running out of time on this PR for 4.10

MosesMendoza: oh yeah I like your idea - similarly effective but cleaner

MosesMendoza: Updated

MosesMendoza:
passed ad-hoc pipeline, https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent_intn-van-sys_suite-ad-hoc-vmpooler-puppet-ad-hoc/7/ (and preceding build)

MosesMendoza:
confirmed that the service logging via ext/windows/service/daemon.rb still works correctly with 64-bit MSI on 64-bit windows. Testing 32bit on 32bit Windows and 32bit on 64bit Windows

MosesMendoza:
"rebased" against stable for targeting 4.10.0

Iristyle:
I think we're ready to squash and ship...

MosesMendoza:
Squashed

ext/windows/service/daemon.rb
Iristyle: Looping back on this one, strictly speaking, the change to `Puppet` from `Application` was incorrect.

`RegisterEventSource` looks under the `EventLog` registry key, and will use the name given *ELSE* it will fall back to `Application` which is known to always exist.  So what we've changed here is that Puppet will use a `Puppet` event log should it exist, but gracefully falls back to `Application` otherwise.

In practical terms, this doesn't cause any problems as the old behavior is preserved. However, it also has the bonus feature of accidentally enabling logging to a different log than `Application`, which is a long-standing feature request for end users.  So this ends up being a dark feature, which solves https://tickets.puppetlabs.com/browse/PUP-5068 and is easily turned on by end users without any additional changes to Puppet or packaging.  So this turns out to be an incredibly happy accident!

I validated against agent package 5.1.0 the following:

* created a new 2012R2 host
* ran the PowerShell code `New-EventLog -Source Puppet -LogName Puppet -MessageResource C:\Program Files\Puppet Labs\Puppet\bin\puppetres.dll` to create a new log named `Puppet` with `Puppet` as a valid source (this creates the underlying event log file at `C:\Windows\System32\winevt\Logs\Puppet.evtx`)
* installed the agent - which registers the source `Puppet` under `Application`
* opened up Event Viewer to take a look at the new `Puppet` log, where all of the messages are now going to instead of `Application`

What makes this particular solution great is that the call to `New-EventLog` can easily be put into a PowerShell based `exec` resource. While initial Puppet messages will still be in the `Application` log, anything after the resource has been applied will go into the `Puppet` log instead:

screen shot 2017-08-24 at 11 49 35 am

MosesMendoza: 🎉 🎉 🎉

PR #5510, opened 2017-01-18
(PUP-4947) Make app management keywords standard

hlindberg:
This changes the optional use of application management to always be "on".
* The setting 'app_management' is deprecated (rather than just removed) in order to not break people's settings when upgrading
* The keywords application, produces, consumes and site are now always keywords
* All behavior that was different depending on the app_management setting now behaves as if it was true

When these changes were made, a test in settings_spec that attempted to pretend it was running on windows failed because of the need to load more classes in the Utils::Windows name space (which is undefined because not loaded earlier - in turn because not running on windows). This test was made conditional on actually running on windows. If this was wong, or there is a better way of doing this, please holler.


hlindberg:
Ping @Iristyle regarding review of last commit modifying 'settings_spec.rb'

spec/unit/settings_spec.rb
```diff
@@ -1308,7 +1308,7 @@ def assert_accessing_setting_is_deprecated(settings, setting)
       @settings.to_catalog
     end
 
-    describe "on Microsoft Windows" do
+    describe "on Microsoft Windows", :if => Puppet.features.microsoft_windows? do
```
Iristyle: The original commit adding this test was https://github.com/hlindberg/puppet/commit/47058abc0c5647d59b0dd21181e67dbfdd908292 from @joshcooper 

If you're going to add the `:if => Puppet.features.microsoft_windows?` line here, then (which I think is OK given we have more robust testing now), then you can remove the stub line below:

```ruby
Puppet.features.stubs(:microsoft_windows?).returns true
```

hlindberg: for sure :-)

joshcooper:
@hlindberg can you file a follow-up ticket for the `puppet_agent` module to remove the deprecated `app_management ` setting during an upgrade?

hlindberg:
@joshcooper filed https://github.com/puppetlabs/puppetlabs-puppet_agent/issues/206 - I hope that was the right place.

PR #5511, opened 2017-01-18
(PUP-5954) Make class merge an error

hlindberg:
Before this, if a class was defined more than once their bodies would be merged. The output of a warning or error was under the control of the --strict flag.
Now, this is always an error except for the top scope class named '' (a.k.a 'main').

### DIFF
--- a/lib/puppet/resource/type_collection.rb
+++ b/lib/puppet/resource/type_collection.rb
@@ -73,18 +73,13 @@ class Puppet::Resource::TypeCollection
     instance
   end
 
+  # @api private
   def handle_hostclass_merge(instance)
+    # Only main class (named '') can be merged (for purpose of merging top-scopes).
+    return instance unless instance.name == ''
     if instance.type == :hostclass && (other = @hostclasses[instance.name]) && other.type == :hostclass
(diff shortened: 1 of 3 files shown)


lib/puppet/resource/type_collection.rb
```diff
@@ -49,11 +49,6 @@ def inspect
     }.inspect
   end
 
-  def <<(thing)
```
thallgren: I think this should remain. We haven't had any deprecation warnings on it and it isn't @api private.

hlindberg: Will put it pack and mark the entire TypeCollection as being @api private in 5.0.0. Most of the methods in that class should never ever be called by anyone outside of the puppet code base. For this particular method we can add deprecations but not for the others unless we start looking at where calls are coming from.... (and we won't). Eventually I want to see TypeCollection gone from the compiler all together.

hlindberg:
I dropped the commit that removed the << method. Instead I marked the entire class as @api private. I know that is not 100% correct - there are probably tools out there that need to use its methods. But I thought it was at least best to put up the "do not walk on the grass" sign.

PR #5400, opened 2016-12-01
(PUP-4742) Switch to use semantic_puppet gem - Take 3

joshcooper:
Revert the revert. Remove the vendored semantic gem, modify puppet to depend on the `external semantic_puppet` gem, and add the `semantic_puppet` runtime dependency when puppet is built as a gem.

This is blocked on https://github.com/puppetlabs/ci-job-configs/pull/2098.

PR #5008, opened 2016-06-07
Revert "(PUP-4742) Switch to use semantic_puppet gem"

kylog:
Reverts puppetlabs/puppet#4977

PR #5512, opened 2017-01-19
(PUP-7114) Re-vendor semantic_puppet gem

Iristyle:
- As part of PUP-4742 the dependency on semantic_puppet was moved to
   a gem.  The sequence of events leading up to its conclusion is
   rather thorny, but ultimately code landed in
   c111fb76b948d547fda7afa901802ff2e8b5a623

 - Because puppet-server was not also shipping the semantic_puppet gem
   this necessitated a strict conflict in the puppet-agent package
   on puppet-server < 2.7.2 and pe-puppet-server < 2017.0.1.3

   This has had quite a bit of fallout, including unintended upgrades
   of the puppet-server package, which is extermely undesirable in a
   production environment.

 - Until Puppet 5 is released, which will allow for more latitude of
   breaking changes and dependency requirements, it has been decided
   that semantic_puppet should be temporarily re-vendored into the
   Puppet repository.  Then there is no need to match the gem in both
   puppet-server and puppet-agent.  Instead, the code from this
   repository is loaded.

 - References to the semantic_puppet gem code have been removed from
   the Gemfile, project_data.yaml and the Puppet .gemspec.

   The code from semantic_puppet 0.1.4 has been directly copied into
   lib/semantic_puppet

### DIFF
--- a/lib/semantic_puppet.rb
+++ b/lib/semantic_puppet.rb
@@ -1,0 +1,9 @@
+require 'gettext-setup'
+
+module SemanticPuppet
+  GettextSetup.initialize(File.absolute_path('../locales', File.dirname(__FILE__)))
+
+  autoload :Version, 'semantic_puppet/version'
+  autoload :VersionRange, 'semantic_puppet/version_range'
+  autoload :Dependency, 'semantic_puppet/dependency'
+end
--- a/.gemspec
+++ b/.gemspec
@@ -36,6 +36,7 @@ Gem::Specification.new do |s|
   s.specification_version = 3
   s.add_runtime_dependency(%q<facter>, [">= 2.0.1", "< 4"])
   s.add_runtime_dependency(%q<hiera>, [">= 2.0", "< 4"])
-  s.add_runtime_dependency(%q<semantic_puppet>, ['>= 0.1.3', '< 2'])
+  # PUP-7115 - return to a gem dependency in Puppet 5
+  # s.add_runtime_dependency(%q<semantic_puppet>, ['>= 0.1.3', '< 2'])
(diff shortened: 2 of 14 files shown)


MosesMendoza:
It seems like we can't put this in `lib/puppet/vendor` with the other vendored deps because we need it loaded earlier? Unfortunate - that seems like it might have been more intuitive there.

lib/semantic_puppet.rb
```diff
@@ -0,0 +1,9 @@
+require 'gettext-setup'
+
+module SemanticPuppet
+  GettextSetup.initialize(File.absolute_path('../locales', File.dirname(__FILE__)))
```
MosesMendoza: Is the plan still to avoid a hard dependency on `gettext-setup`? If so is that change intended for a separate PR?

Iristyle: Yeah, I was thinking we'd do a roundup on all the gettext related work in a separate PR, but in this PR would also work if anyone feels strongly.

Iristyle:
@MosesMendoza we could place in `lib/puppet/vendor` but then would have to change any existing `require` statements.  It can be done, but felt simpler to do a drop-in replacement, since we know it will be short-lived.

MosesMendoza:
Sounds good. Care to add a short-lived exclude .rubocop.yml for semantic_puppet? We already exclude other things in vendor

lib/semantic_puppet.rb
MosesMendoza: 👍

Magisus:
👍

Iristyle:
For historical purposes, there's a long history here of 6 PRs back to February 2016:

* #4708 
* #4977 
* #5008 
* #5094 
* #5399 
* #5400

PR #5513, opened 2017-01-19
(PUP-7073) Normalize selinux category names

branan:
SELinux category names have two forms: an internal 'numeric' form, and
a human-readable name. Several SELinux commands rely on a service
called `mcstransd` to convert between them. When that service is not
functioning correctly, things will break.

Instead of using `mcstransd`, Puppet now queries the list of
translations itself and uses them to normalize all internal operations
on the numeric form.

### DIFF
--- a/lib/puppet/type/file/selcontext.rb
+++ b/lib/puppet/type/file/selcontext.rb
@@ -29,7 +29,12 @@ module Puppet
     def retrieve
       return :absent unless @resource.stat
       context = self.get_selinux_current_context(@resource[:path])
-      parse_selinux_context(name, context)
+      is = parse_selinux_context(name, context)
+      if name == :selrange and selinux_support?
+        self.selinux_category_to_label(is)
+      else
+        is
+      end
     end
 
     def retrieve_default_context(property)
@@ -55,6 +60,18 @@ module Puppet
         true
       else
         super
+      end
+    end
+
+    def unsafe_munge(should)
+      if not selinux_support?
+        return should
+      end
+
+      if name == :selrange
+        self.selinux_category_to_label(should)
+      else
+        should
       end
     end
 
--- a/lib/puppet/util/selinux.rb
+++ b/lib/puppet/util/selinux.rb
@@ -137,6 +137,42 @@ module Puppet::Util::SELinux
     nil
   end
 
+  ##
+  # selinux_category_to_label is an internal method that converts all
+  # selinux categories to their internal representation, avoiding
+  # potential issues when mcstransd is not functional.
+  #
+  # It is not marked private because it is needed by File's
+  # selcontext.rb, but it is not intended for use outside of Puppet's
+  # code.
+  #
+  # @param category [String] An selinux category, such as "s0" or "SystemLow"
+  #
+  # @return [String] the numeric category name, such as "s0"
+  def selinux_category_to_label(category)
+    # We don't cache this, but there's already a ton of duplicate work
+    # in the selinux handling code.
+
+    path = Selinux.selinux_translations_path
+    begin
+      File.open(path).each do |line|
+        line.strip!
+        next if line.empty?
+        next if line[0] == "#" # skip comments
+        line.gsub!(/[[:space:]]+/m, '')
+        mapping = line.split("=", 2)
+        if category == mapping[1]
+          return mapping[0]
+        end
+      end
+    rescue SystemCallError => ex
+      log_exception(ex)
+      raise Puppet::Error, _("Could not open SELinux category translation file %{path}.") % { context: context }
+    end
+
+    category
+  end
+
   ########################################################################
   # Internal helper methods from here on in, kids.  Don't fiddle.
   private
--- a/lib/puppet/util/selinux.rb
+++ b/lib/puppet/util/selinux.rb
@@ -62,16 +62,22 @@ module Puppet::Util::SELinux
     if context.nil? or context == "unlabeled"
       return nil
     end
-    unless context =~ /^([^\s:]+):([^\s:]+):([^\s:]+)(?::([\sa-zA-Z0-9:,._-]+))?$/
+    components = /^([^\s:]+):([^\s:]+):([^\s:]+)(?::([\sa-zA-Z0-9:,._-]+))?$/.match(context)
+    unless components
       raise Puppet::Error, _("Invalid context to parse: %{context}") % { context: context }
     end
-    ret = {
-      :seluser => $1,
-      :selrole => $2,
-      :seltype => $3,
-      :selrange => $4,
-    }
-    ret[component]
+    case component
+    when :seluser
+      components[1]
+    when :selrole
+      components[2]
+    when :seltype
+      components[3]
+    when :selrange
+      components[4]
+    else
+      raise Puppet::Error, _("Invalid SELinux parameter type")
+    end
   end
 
   # This updates the actual SELinux label on the file.  You can update
--- a/spec/unit/type/file/selinux_spec.rb
+++ b/spec/unit/type/file/selinux_spec.rb
@@ -10,6 +10,8 @@ require 'spec_helper'
       @path = make_absolute("/my/file")
       @resource = Puppet::Type.type(:file).new :path => @path
       @sel = property.new :resource => @resource
+      @sel.stubs(:normalize_selinux_category).with("s0").returns("s0")
+      @sel.stubs(:normalize_selinux_category).with(nil).returns(nil)
     end
 
     it "retrieve on #{param} should return :absent if the file isn't statable" do


lib/puppet/util/selinux.rb
```diff
@@ -137,6 +142,33 @@ def set_selinux_default_context(file)
...
+  # code.
+  #
+  # @return [String] the human-readable category name
+  def normalize_selinux_category(category)
+    # We don't cache this, but there's already a ton of duplicate work
+    # in the selinux handling code.
+
+    File.open(Selinux.selinux_translations_path).each do |line|
+      stripped = line.strip!
+      next if stripped[0] == "#" # skip comments
```
MosesMendoza: is the comment guaranteed to be the first character on the line? Or is it legal to include whitespace before the comment character?

lib/puppet/util/selinux.rb
```diff
@@ -137,6 +142,33 @@ def set_selinux_default_context(file)
...
+      line.gsub!(/\s+/m, '')
```
MosesMendoza: Can we use `[[:space:]]` instead of `\s` ? Helps guard against encoding issues.

> POSIX bracket expressions are also similar to character classes. They provide a portable alternative to the above, with the added benefit that they encompass non-ASCII characters. For instance, /\d/ matches only the ASCII decimal digits (0-9); whereas /[[:digit:]]/ matches any character in the Unicode Nd category.
https://ruby-doc.org/core-2.2.0/Regexp.html. PUP-6875 is related.

lib/puppet/util/selinux.rb
```diff
@@ -137,6 +142,33 @@ def set_selinux_default_context(file)
...
+  ##
+  # normalize_selinux_category is an internal, method that converts
+  # all selinux categories to the human-readable representation,
+  # avoiding potential issues when mcstransd is not functional.
+  #
+  # It is not marked private because it is needed by File's
+  # selcontext.rb, but it is not intended for use outside of Puppet's
+  # code.
+  #
+  # @return [String] the human-readable category name
```
MosesMendoza: Would you mind adding a @param entry too with an example of a category string?

lib/puppet/util/selinux.rb
branan: I'm not particularly concerned about funky unicode spaces ending up in selinux configurations, but it's good to be consistent. I can fix this up

lib/puppet/util/selinux.rb
branan: :+1:

lib/puppet/util/selinux.rb
```diff
@@ -65,13 +65,18 @@ def parse_selinux_context(component, context)
...
-    ret = {
-      :seluser => $1,
-      :selrole => $2,
-      :seltype => $3,
-      :selrange => $4,
-    }
-    ret[component]
+    case component
+    when :seluser
+      $1
```
MosesMendoza: wow this syntax really surprised me - I don't think I've ever seen using these global $<int> variables for regex matches before in ruby.

lib/puppet/type/file/selcontext.rb
```diff
@@ -58,6 +63,18 @@ def insync?(value)
       end
     end
 
+    def unsafe_munge(should)
+      if not selinux_support?
+        should
```
MosesMendoza: It may not be very ruby-like but given how many implicit returns are in these methods, maybe it would be more readable to do explicit `return should` etc.?

lib/puppet/util/selinux.rb
```diff
@@ -137,6 +142,33 @@ def set_selinux_default_context(file)
...
+  # selcontext.rb, but it is not intended for use outside of Puppet's
+  # code.
+  #
+  # @return [String] the human-readable category name
+  def normalize_selinux_category(category)
+    # We don't cache this, but there's already a ton of duplicate work
+    # in the selinux handling code.
+
+    File.open(Selinux.selinux_translations_path).each do |line|
+      stripped = line.strip!
```
MosesMendoza: `strip!` will return nil if the end of the file doesn't have a new line, so stripped will be nil and this will then fail when trying to index on the next line

MosesMendoza:
Just to confirm - for example, given a system with a MCS numeric = label mapping such as `s0=SystemLow`, the point here is that if an `is` value or `should` value starts as `SystemLow`, it will be set to `s0` ?

lib/puppet/util/selinux.rb
```diff
@@ -137,6 +142,33 @@ def set_selinux_default_context(file)
...
+  # It is not marked private because it is needed by File's
+  # selcontext.rb, but it is not intended for use outside of Puppet's
+  # code.
+  #
+  # @return [String] the human-readable category name
+  def normalize_selinux_category(category)
+    # We don't cache this, but there's already a ton of duplicate work
+    # in the selinux handling code.
+
+    File.open(Selinux.selinux_translations_path).each do |line|
```
MosesMendoza: Are you worried about the performance implications of a system with many file resources with managed `:selrange` properties? I'm not sure if caching is a better idea.. The `chcat` source code doesn't appear to anyway: https://github.com/SELinuxProject/selinux/blob/7935dee8f62931159e3bdb7bc028c2fdb7ecd7da/python/chcat/chcat#L371-L380

lib/puppet/type/file/selcontext.rb
branan: yeah, that's probably reasonable

lib/puppet/util/selinux.rb
branan: My main goal with this quick refactor was to get rid of the extra hash allocation/lookup while I was in here, but I no longer actually need to change this method in this commit. I'll pull this out and push up a separate maint commit (still part of this PR, probably) that will clean up this function a little more than just the hash->case change

lib/puppet/util/selinux.rb
branan: I still get nervous about trying to do caching in Puppet - I'm inclined to wait and see if anybody complains.

lib/puppet/util/selinux.rb
branan: that seems way less useful than the risk of somebody like me accidentally using a nil -_-

But of course with the `!` version of strip I don't need to use the return value at all. Not really sure what I was thinking when I wrote it this way...

branan:
@MosesMendoza That's the deal. It also passes through unrecognized values unchanged, so if the user puts in `s0` it'll "just work", and if they put in `potato` it'll end up out of sync and then fail to make the change, since that's an invalid category.

lib/puppet/util/selinux.rb
MosesMendoza: :thu

MosesMendoza:
@branan sounds good. Given the esoteric nature of the functionality we're working with here, any chance we could come up with a more intuitive name for the `normalize_selinux_category` method? I was scratching my head trying to figure out what was going on.. I wonder if something like `category_label_to_value(category)` might be more expressive. Thoughts?

lib/puppet/util/selinux.rb
vinzent: What happens if the file doesn't exist?

there would exist a trans_context / untrans_context in  (https://github.com/SELinuxProject/selinux/blob/master/mcstrans/src/mcstrans.h) - but its not available with libselinux-ruby.

branan: @vinzent How likely is that file to not exist when selinux is being used? My poking around (admittedly just on a couple of distros) show it seems to come along with the `selinux-policy` packages, so should basically always exist.

If it doesn't exist, we have two options: bail with an exception (the current case), or silently go forward without translations. Going forward silently would essentially be the same behavior as what this patch is trying to solve. I'm hesitant to leave behavior that can cause downstream resources to be refreshed accidentally, and would much rather just bail the catalog.

branan:
@MosesMendoza I came up with `selinux_category_to_label`, which is the best I've got from reading the docs on what these various words are supposed to mean ¯\\\_(ツ)_/¯. I'm not sure that's really entirely correct, but everything is terrible.

lib/puppet/util/selinux.rb
vinzent: My experience with "a file exists always" is mostly that its not true. although I couldn't imagine that it ever wouldn't exist before there was the one important case where it didn't exist. ;-)

MosesMendoza: might be worth falling back to the old behavior if the file doesn't exist

branan:
@MosesMendoza This is updated with slightly nicer error handling if the `setrans.conf` file is missing. 

I suspect I should rebase atop latest master and add gettext markers as needed around my new strings?

Magisus:
This was originally targeted at stable, are we switching to master?

And yes please on the string marking!

branan:
@Magisus Yeah, this is being re-targeted at master. The ticket was updated but the PR target apparently wasn't. I'll get it cleaned up.

branan:
@Magisus Rebased and added translation. Can you make sure I did that right?

lib/puppet/util/selinux.rb
```diff
@@ -62,16 +62,22 @@ def parse_selinux_context(component, context)
...
+    when :seluser
+      components[1]
+    when :selrole
+      components[2]
+    when :seltype
+      components[3]
+    when :selrange
+      components[4]
+    else
+      raise Puppet::Error, "Invalid SELinux parameter type"
```
Magisus: This error should probably be externalized too.

branan: @Magisus fixed

PR #5514, opened 2017-01-19
(PUP-7098) This PR adds testing of utf-8 characters in the catalog.

seangriff:
Verifies proper behavior when catalog is compiled and then applied
from cache.

### DIFF
--- a/acceptance/tests/utf8/utf8-in-catalog.rb
+++ b/acceptance/tests/utf8/utf8-in-catalog.rb
@@ -1,0 +1,100 @@
+test_name 'utf-8 characters in cached catalog' do
+  utf8chars = "€‰ㄘ万竹ÜÖ"
+  file_content = "This is the file content. file #{utf8chars}"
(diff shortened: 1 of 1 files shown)


acceptance/tests/utf8/utf8-in-catalog.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    assert_match(
+      /#{file_content}/,
+      result.stdout,
+      "cached catalog does not contain expected file content"
+    )
+
+    step "apply cached catalog"
+    result = on(agent, puppet("resource file #{agent_file} ensure=absent"))
+    result = on(agent, puppet("catalog apply --terminus json"))
```
er0ck: i don't think these result assignments are doing anything.

Iristyle:
Could you change the commit msg subject to follow the guidelines at https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md#making-changes ?

Thanks!

PR #5515, opened 2017-01-19
(PUP-7117) New acceptance test: utf-8 characters in file resource

seangriff:
This PR adds a new accecptance test to verify correct behavior of
file resources when they include utf-8 characters.


acceptance/tests/utf8/utf8-in-file-resource.rb
```diff
@@ -0,0 +1,48 @@
+test_name 'utf-8 characters in resource title and param values'
+
+confine :except, :platform => [ 'windows', 'ubuntu-16']   # PUP-6983
+
+utf8chars = "€‰ㄘ万竹ÜÖ"
+
+agents.each do |agent|
+  agent_file = agent.tmpfile("file" + utf8chars)
```
er0ck: nice.

PR #5517, opened 2017-01-19
(PUP-7119) New Acceptance test: utf-8 characters in 'puppet describe'

seangriff:
This PR adds a new acceptance test to verify correct behavior when
utf-8 chararcters are used the module description that is displayed
with the 'puppet describe' command.


acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,56 @@
+test_name 'utf-8 characters in module doc string, puppet describe'
```
er0ck: the `test_name` and `step` calls need do/end blocks in all these PRs

PR #5518, opened 2017-01-19
(PUP-7116) Only use gettext functionality if gettext-setup gem is available

Magisus:
We cannot guarantee that the gettext-setup gem will be present or of the
most recent version when running puppet from within puppetserver
(e.g. if the server version is < 2.7.2, the correct version of the gem
 will not be present). This commit handles the absent case by stubbing
out the `_` method used for loading translations, and the out-of-date
case by using the old API for `initialize`, which removes support for MO
translation files. If gettext-setup < 0.8 is loaded, only PO translation
files will be used.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -20,8 +20,20 @@ require 'puppet/util/run_mode'
 require 'puppet/external/pson/common'
 require 'puppet/external/pson/version'
 require 'puppet/external/pson/pure'
-require 'gettext-setup'
-require 'locale'
+
+# When running within puppetserver, the gettext-setup gem might not be available, so
+# we need to skip initializing i18n functionality and stub out methods normally
+# supplied by gettext-setup. Can be removed in Puppet 5. See PUP-7116.
+begin
+  require 'gettext-setup'
+  require 'locale'
+  Puppet::GETTEXT_AVAILABLE = true
+rescue LoadError
+  def _(msg)
+    msg
+  end
+  Puppet::GETTEXT_AVAILABLE = false
+end
 
 #------------------------------------------------------------
 # the top-level module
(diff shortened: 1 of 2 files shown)


Iristyle:
Can we get #5512 merged, and then update this PR to also make the necessary modifications to the vendored `semantic_puppet` code?

lib/puppet.rb
```diff
@@ -20,8 +20,20 @@
...
+
+# When running within pupeptserver, the gettext-setup gem is not available, so
+# we need to skip initializing i18n functionality and stub out methods normally 
+# supplied by gettext-setup.
+begin
+  require 'gettext-setup'
+  require 'locale'
+  GETTEXT_AVAILABLE = true
+rescue LoadError
+  def _(msg)
```
MosesMendoza: Is `_` the only method we're using from gettext?

lib/puppet.rb
```diff
@@ -20,8 +20,20 @@
 require 'puppet/external/pson/common'
 require 'puppet/external/pson/version'
 require 'puppet/external/pson/pure'
-require 'gettext-setup'
-require 'locale'
+
+# When running within pupeptserver, the gettext-setup gem is not available, so
```
MosesMendoza: s/pupeptserver/puppetserver

MosesMendoza: also I this comment might be more accurate as "the gettext-setup gem might not be available"

lib/puppet.rb
```diff
@@ -20,8 +20,20 @@
...
+# we need to skip initializing i18n functionality and stub out methods normally 
+# supplied by gettext-setup.
+begin
+  require 'gettext-setup'
+  require 'locale'
+  GETTEXT_AVAILABLE = true
```
MosesMendoza: Since this is in the global ruby namespace to avoid the possibility of collision we could put this inside Puppet, i.e., `Puppet::GETTEXT_AVAILABLE`. Since we've already required 'puppet/version' the Puppet module is defined.

lib/puppet.rb
Magisus: Yes, besides the ones used in this file.

Magisus:
Yes @Iristyle that seems good.

lib/semantic_puppet.rb
```diff
@@ -1,7 +1,18 @@
-require 'gettext-setup'
+begin
+  require 'gettext-setup'
+  GETTEXT_AVAILABLE = true
```
Magisus: The Puppet module has not been loaded here, what should we do about namespacing?

MosesMendoza: Hmm interesting. Perhaps avoid the constant altogether, instead just checking if gettext is defined? i.e., `if defined?(GettextSetup)` ? Also not foolproof

Magisus: I'll take it

Iristyle:
Against this PR, I removed references to `gettext-setup` like this:

```diff
diff --git a/.gemspec b/.gemspec
index 3215154..e59f903 100644
--- a/.gemspec
+++ b/.gemspec
@@ -38,5 +38,5 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency(%q<hiera>, [">= 2.0", "< 4"])
   # PUP-7115 - return to a gem dependency in Puppet 5
   # s.add_runtime_dependency(%q<semantic_puppet>, ['>= 0.1.3', '< 2'])
-  s.add_runtime_dependency(%q<gettext-setup>, [">= 0.10", "< 1"])
+  # s.add_runtime_dependency(%q<gettext-setup>, [">= 0.10", "< 1"])
 end
diff --git a/Gemfile b/Gemfile
index 7547e78..ff36df3 100644
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,7 @@ gem "rake", "10.1.1", :require => false
 # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
 gem 'json_pure', '~> 1.8', :require => false
 # i18n support (gettext-setup and dependencies)
-gem 'gettext-setup', '>= 0.10', '< 1.0', :require => false
+# gem 'gettext-setup', '>= 0.10', '< 1.0', :require => false
 gem 'locale', '~> 2.1', :require => false

 group(:development, :test) do
diff --git a/ext/project_data.yaml b/ext/project_data.yaml
index 8f8c765..58f4e1d 100644
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -22,7 +22,7 @@ gem_runtime_dependencies:
   json_pure: '~> 1.8'
   # PUP-7115 - return to a gem dependency in Puppet 5
   # semantic_puppet: ['>= 0.1.3', '< 2']
-  gettext-setup: ['>= 0.10', '< 1']
+  # gettext-setup: ['>= 0.10', '< 1']
   locale: '~> 2.1'
 gem_rdoc_options:
   - --title
```

Confirmed no copies of `gettext`

```
bundle exec gem list | grep gettext
```

Ran specs

```
bundle exec rake parallel:spec
Processing 22 spec group(s) with 8 worker(s)
............................................................................................................................................................

<snip />

Finished in 2 minutes 10.6 seconds
22454 examples, 0 failures, 46 pending
```

Ran `puppet help` without error:

```
bundle exec puppet help
Warning: Support for ruby version 1.9.3 is deprecated and will be removed in a future release. See https://docs.puppet.com/puppet/latest/system_requirements.html#ruby for a list of supported ruby versions.
   (at /Users/Iristyle/source/puppet-vendor-semantic-puppet/lib/puppet.rb:167:in `<module:Puppet>')

Usage: puppet <subcommand> [options] <action> [options]

Available subcommands:

  agent             The puppet agent daemon
  apply             Apply Puppet manifests locally
  ca                Local Puppet Certificate Authority management. (Deprecated)
  catalog           Compile, save, view, and convert catalogs.
  cert              Manage certificates and requests
  certificate       Provide access to the CA for certificate management.
  certificate_request  Manage certificate requests. (Deprecated)
  certificate_revocation_list  Manage the list of revoked certificates. (Deprecated)
  config            Interact with Puppet's settings.
  describe          Display help about resource types
  device            Manage remote network devices
  doc               Generate Puppet references
  epp               Interact directly with the EPP template parser/renderer.
  facts             Retrieve and store facts.
  file              Retrieve and store files in a filebucket (Deprecated)
  filebucket        Store and retrieve files in a filebucket
  generate          Generates Puppet code from Ruby definitions.
  help              Display Puppet help.
  inspect           Send an inspection report
  key               Create, save, and remove certificate keys. (Deprecated)
  lookup            Data in modules lookup function
  man               Display Puppet manual pages.
  master            The puppet master daemon
  module            Creates, installs and searches for modules on the Puppet Forge.
  node              View and manage node definitions.
  parser            Interact directly with the parser.
  plugin            Interact with the Puppet plugin system.
  report            Create, display, and submit reports.
  resource          The resource abstraction layer shell
  resource_type     View classes, defined resource types, and nodes from all manifests.
  status            View puppet server status. (Deprecated)

See 'puppet help <subcommand> <action>' for help on a specific subcommand action.
See 'puppet help <subcommand>' for help on a specific subcommand.
Puppet v4.8.2
```

👍

PR #5519, opened 2017-01-19
(maint) Ensure graceful lookup fail without hocon

Iristyle:
- Add a test demonstrating the behavior of the lookup function when
   Hocon is specified as a backend, but the hocon library is not
   present.

   This demonstrates what will happen if a newer agent is running
   on puppet-server < 2.7.0, which does not include the hocon gem


MosesMendoza:
👍

Iristyle:
Forgot that `cause` is not supported on Ruby 1.9.3 / Ruby 2.0.0.  I've rewritten a simpler variation of the test.

PR #5520, opened 2017-01-20
(PUP-7021) Use UTF-8 encoding for strings returned by Etc in Puppet::Provider::Nameservice

MosesMendoza:
This PR sets external encoding to UTF8 for the strings we get back from the ruby Etc module. The Etc module can return strings in system encoding, which makes its way into puppet and wreaks havoc as we try to concatenate and compare them with strings generated in puppet itself which are utf8. Rather than try to find all the different places where we concatenate, compare, or otherwise manipulate strings from this provider, just set external encoding to UTF8 at their inception.

From the commit messages:
> Use UTF-8 encoding in Puppet::Provider::NameService#listbyname
>
> The #listbyname method is used in the nameservice provider to generate the
> values for self.instances. If a name returned by the Etc module is a utf-8
> value but the system encoding is not, self.instances, and thus things like
> `puppet resource user` fail with encoding errors.
>
> This commit addresses this by setting the external encoding (not modifying byte
> values) of the returned string to UTF8. This is one of the entry points for
> these strings into the puppet system, so we prevent a cascade of downstream
> failures by setting this to utf-8 here.

> (PUP-7021) Use UTF-8 encoding in Puppet::Provider::NameService#groups
> 
> The #groups method is used by the nameservice provider (not surprisingly) in
> querying group membership. We use the Etc module to parse the group file, which
> can return strings in system encoding. Ruby considers two identical stringis
> (in bytes) with different external encodings to be unequal. Since the names in
> the groups array returned by Etc can be in system encoding, and our user is in
> utf-8 encoding, our group membership check will fail.
> 
> This commit addresses this by setting the external encoding (not modifying byte
> values) of the strings returned by Etc to UTF8. This is one of the entry points
> for these strings into the puppet system, so we prevent a cascade of downstream
> failures by setting this to utf-8 here.

> (PUP-7021) Use UTF-8 encoding in Puppet::Provider::NameService#info2hash
> 
> the #info2hash method in the nameservice provider leverages Etc module to
> gather information about a user or group. However, the strings returned by the
> module can be in system encoding, rather than UTF-8. Later, when puppet tries
> to use the information for comparison or display purposes with Puppet's UTF-8
> strings, we fail in many, many ways.
> 
> This commit addresses this by setting the external encoding (not modifying byte
> values) of the strings returned by Etc to UTF8. This is one of the entry points
> for these strings into the puppet system, so we prevent a cascade of downstream
> failures by setting this to utf-8 here.

### DIFF
--- a/lib/puppet/provider/nameservice.rb
+++ b/lib/puppet/provider/nameservice.rb
@@ -55,7 +55,7 @@ class Puppet::Provider::NameService < Puppet::Provider
       Etc.send("set#{section()}ent")
       begin
         while ent = Etc.send("get#{section()}ent")
-          names << ent.name
+          names << ent.name.force_encoding(Encoding::UTF_8)
           yield ent.name if block_given?
         end
       ensure
--- a/lib/puppet/provider/nameservice.rb
+++ b/lib/puppet/provider/nameservice.rb
@@ -249,9 +249,9 @@ class Puppet::Provider::NameService < Puppet::Provider
     # Now iterate across all of the groups, adding each one our
     # user is a member of
     while group = Etc.getgrent
-      members = group.mem
+      members = group.mem.map { |member| member.force_encoding(Encoding::UTF_8) }
 
-      groups << group.name if members.include? user
+      groups << group.name.force_encoding(Encoding::UTF_8) if members.include? user
     end
 
     # We have to close the file, so each listing is a separate
--- a/lib/puppet/provider/nameservice.rb
+++ b/lib/puppet/provider/nameservice.rb
@@ -266,7 +266,11 @@ class Puppet::Provider::NameService < Puppet::Provider
     hash = {}
     self.class.resource_type.validproperties.each do |param|
       method = posixmethod(param)
-      hash[param] = info.send(posixmethod(param)) if info.respond_to? method
+      if info.respond_to?(method)
+        value = info.send(posixmethod(param))
+        value.force_encoding(Encoding::UTF_8) if value.respond_to?(:force_encoding)
+        hash[param] = value
+      end
     end
 
     hash
--- a/spec/unit/provider/nameservice_spec.rb
+++ b/spec/unit/provider/nameservice_spec.rb
@@ -60,6 +60,14 @@ describe Puppet::Provider::NameService do
     resource
   end
 
+  let(:snowman_iso)     { "\u2603".force_encoding(Encoding::ISO_8859_1) }
+  let(:snowman_utf8)    { "\u2603".force_encoding(Encoding::UTF_8) }
+  let(:snowman_binary)  { "\u2603".force_encoding(Encoding::ASCII_8BIT) }
+  let(:arabic_heh_iso)  { "\u06FF".force_encoding(Encoding::ISO_8859_1) }
+  let(:arabic_heh_utf8) { "\u06FF".force_encoding(Encoding::UTF_8) }
+  let(:runic_fehu_iso)  { "\u16A0".force_encoding(Encoding::ISO_8859_1) }
+  let(:runic_fehu_utf8)  { "\u16A0".force_encoding(Encoding::UTF_8) }
+
   describe "#options" do
     it "should add options for a valid property" do
       described_class.options :foo, :key1 => 'val1', :key2 => 'val2'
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/nameservice.rb
```diff
@@ -55,7 +55,7 @@ def listbyname
       Etc.send("set#{section()}ent")
       begin
         while ent = Etc.send("get#{section()}ent")
-          names << ent.name
+          names << ent.name.force_encoding(Encoding::UTF_8)
```
branan: I think we want `encode` here, rather than `force_encoding`, maybe falling back to `force_encoding` if for some reason the local locale (heh) has characters that can't be represented in Unicode. `force_encoding` just reinterprets the bytes as the new encoding, which turns them in to garbage - they may not even be (and probably aren't, if we're coming from a different encoding) valid UTF-8 data.

`encode` instead will do the proper conversion, taking each character from the external encoding to utf-8.

branan:
A couple of other thoughts related to encodings and user data:

1) There are actually two incompatible cases we can solve for:
  * The data we're reading is in the correct system encoding. In that case, we want to `encode` it to utf-8 on read, then `encode` it back to the system encoding on write.
  * The data we're reading is already unicode, despite what the system encoding says. In this case `force_encoding` is the way to go.

The first of those is the one I'd expect us to want to implement, but this PR appears to address the second one.

2) In the first case, we don't only have to worry about this when printing to console. We also need to make sure that the data we shove into the `Etc` module is in the correct encoding. That means that in addition to doing a `.encode` on reading from the `Etc` module, we'd also want a `.encode(Encoding.default_external())` on writing information to the `Etc` module. We'd also want to audit all providers for correct use of the system encoding when putting strings into OS data structures.

MosesMendoza:
@branan seems reasonable. I was really wary of manipulating the actual byte representations, and thought I had conceived of a way we would be working with raw bytes in our is/should comparison (where all we really need for ruby to treat them correctly is equivalent external encodings) but chatting with @Iristyle too it seems like this might be a can of worms to go down and the safest bet is actually to encode. Thinking..

Iristyle:
I generally agree with the sentiment that calling `.force_encoding(Encoding::UTF_8)` is problematic @branan. If we're going to do that strictly for the sake of comparison, it's probably best to use `Encoding::BINARY` instead, which is a clearer signal IMHO about what we're trying to accomplish.

That said, we're still stuck in between a rock and a hard place.  

#### Consider an instance where strings do come back from the system in encoding X

Given our manifests are always `UTF-8` and we have no way of expressing bytes in the language per https://docs.puppet.com/puppet/latest/lang_data_string.html (something I don't think we want in the language either if I'm being honest) - this creates a horrible user experience.  A user will have to be responsible for representing the `UTF-8` encoded version of the values returned by their system in its native encoding X in a manifest for comparison.

While everything is supposed to be able to convert from encoding X to `UTF-8`, that isn't always the case in Ruby.  Some encodings only have partial support, and I believe it's also the case that not all strings cleanly roundtrip.  That is, value A in encoding X may translate to value B in `UTF-8` (which a user would put in a manifest), but converting value B from `UTF-8` may instead translate to value C in encoding X.  Now what?

#### Alternate solution

For that reason, I'm inclined to want to address this in a different way:

* Documentation caveats about "system encoding" unpredictability on systems not running `UTF-8`.
* Best attempt made by Puppet to convert to `UTF-8`.  I think a small wrapper around `Etc` might be our best bet.  Rather than try to doctor `Etc` results at individual callsites, create a wrapper for the things that Puppet cares about.  In that wrapper, call `.encode` on values that are not coming back as `UTF-8`.  There are 80 instances where the string `Etc` appears, and not all of them are actually code, so I don't think this is a ton of effort.
    * We could theoretically attempt to fix all the `Etc` problems in one spot
    * It gives us a clean point at which to emit a warning about conversion failures (or the fact users should stick to `utf-8` for correct behavior)
    * Should something still go wrong during a conversion (I'd be inclined to let `.encode` crash instead of providing substitution characters), it points us to a very clear failure point in a call stack during a crash.

MosesMendoza:
+1 @branan @Iristyle. Will rework this a bit in that spirit. Given that we'd be modifying strings and potentially newly (or at least differently) failing in some circumstances, I think this would be a puppet 5 change instead of a 4.9.0 change. Closing for now.

PR #5522, opened 2017-01-20
(PUP-7122) Add wrapper for gettext:update_pot rake task

Magisus:
This adds a wrapper rake task for a new update_pot task that has been
added to the gettext-setup gem. This task will cache the old POT file,
generate a new one, diff them, and only replace the old one if actual
string changes have occurred. This can be used to reduce churn from
changing dates and SHAs.

The second commit adds the `vendor` directory to the ignore list for gettext-setup. Vendored code should handle its own internationalization.

### DIFF
--- a/locales/config.yaml
+++ b/locales/config.yaml
@@ -26,4 +26,4 @@ gettext:
     - 'lib/puppet/pops/types/type_formatter.rb'
     # The semantic_puppet gem is temporarily vendored (PUP-7114), and it
     # handles its own translation.
-    - 'lib/semantic_puppet/**/*.rb'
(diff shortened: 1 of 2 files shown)


Magisus:
This is blocked on https://github.com/puppetlabs/gettext-setup-gem/pull/19 and a subsequent release of that gem.

Magisus:
The gettext gem version 0.13 with the new rake task has been released, so this PR is now unblocked.

McdonaldSeanp:
👍

Magisus:
Rebased this onto 5.0, since that's where all the string externalization is getting merged to.

ScottGarman:
:+1:

PR #5524, opened 2017-01-20
(docs) Minor heading fix for yumrepo reference

jtappa:
These headings should be h3 because all the h2 headings are for the resource types themselves. So currently it's causing it to display incorrectly in the docs navigation.


MikaelSmith:
I see it https://docs.puppet.com/puppet/latest/type.html#fetching-instances; ends up in the ToC which is very messy.

PR #5526, opened 2017-01-20
(PUP-7124) Pass options declared in Hiera v5 hierarchy to hiera3_backend

thallgren:
This commit fixes an oversight in the PR that was merged for PUP-7089
causing the options declared in a hierarchy entry not to be passed on to
the declared backend.

The commit also fixes a bug that made it impossible to declare more than
one hiera3_backend in the hierarchy.


hlindberg:
Fails on Ruby 1.9.3 which is still supported for 4.x

PR #5529, opened 2017-01-23
(PUP-7123) Allow use of beaker 3.x in acceptance tests

ScottGarman:
in particular).

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -12,7 +12,7 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.5.0')
+gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.5')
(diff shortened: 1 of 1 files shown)


ScottGarman:
Marking this do not merge until some unexpected issues with beaker are sorted out.

ScottGarman:
Ok, the beaker-facter gem has been updated to resolve the issues we were seeing. This PR is now safe to merge.

MosesMendoza:
@ScottGarman looking at the latest on BKR-1044, it seems like we still might be waiting on something. Is that your understanding too?

ScottGarman:
@MosesMendoza thanks for catching that. This PR is blocked until https://github.com/puppetlabs/facter/pull/1495 is addressed.

Iristyle:
Beaker has cut an updated gem which should fix this - so I think we're ready to merge now.

See also: https://github.com/puppetlabs/facter/pull/1495#issuecomment-275470658

ScottGarman:
Closing this in favor of https://github.com/puppetlabs/puppet/pull/5606 which is targeted at stable.

PR #5533, opened 2017-01-24
(PUP-7114) Vendor semantic_puppet locales

Iristyle:
- Add locales content from the semantic_puppet gem that should have
   been added in 65d1d94
 - Previously in the gem, the content was laid out like:

   lib/semantic_puppet.rb
   locales/config.yaml

   Therefore the appropriate relative path to find config.yaml was:

   ../locales/config.yaml

 - However, under the puppet lib directory, the locales directory must
   be isolated since it may otherwise conflict with other gems inside
   lib, so the content is instead laid out like:

   lib/semantic_puppet.rb
   lib/semantic_puppet/locales

   Therefore the appropriate relative path to find config.yaml becomes:

   semantic_puppet/locales/config.yaml

### DIFF
--- a/lib/semantic_puppet.rb
+++ b/lib/semantic_puppet.rb
@@ -8,7 +8,7 @@ end
 
 module SemanticPuppet
   if defined?(GettextSetup)
-    GettextSetup.initialize(File.absolute_path('../locales', File.dirname(__FILE__)))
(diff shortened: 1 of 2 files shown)


ScottGarman:
:+1:

PR #5534, opened 2017-01-24
(PUP-7128) Error when reserved options are used in Hiera config

thallgren:
This commit checks that options specified in a Hiera version 5 hierarchy
entry does not use the reserved keys 'path' or 'uri'.


hlindberg:
good to merge when light goes green!

PR #5535, opened 2017-01-24
(PUP-7098) New acceptance test: utf-8 characters in the catalog

seangriff:
This PR adds new acceptance test to verify correct behavior when
catalog containing utf-8 character5s is compiled and then applied
from cache.

### DIFF
--- a/acceptance/tests/utf8/utf8-in-catalog.rb
+++ b/acceptance/tests/utf8/utf8-in-catalog.rb
@@ -1,0 +1,100 @@
+test_name 'utf-8 characters in cached catalog' do
+  utf8chars = "€‰ㄘ万竹ÜÖ"
+  file_content = "This is the file content. file #{utf8chars}"
+  tmpdir = master.tmpdir("code")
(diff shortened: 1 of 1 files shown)


acceptance/tests/utf8/utf8-in-catalog.rb
```diff
@@ -0,0 +1,100 @@
...
+  on(master, "rm -rf #{tmpdir}")
+  env_dir = "#{tmpdir}/environments"
+  agents.each do |agent|
+    agent_file = agent.tmpfile("file" + utf8chars) 
+
+    step "Apply manifest" do
+      on(agent, "rm -rf #{agent_file}")
+    
+      master_manifest =
+<<PP
```
er0ck: this is hard to read and makes the indenting less valuable.  the heredoc should be up next to the ` = `

er0ck: meh.  this is no big deal.  👍 on the rest

PR #5536, opened 2017-01-24
(PUP-7117) New acceptance test: utf-8 characters in file resource

seangriff:
This PR adds a new accecptance test to verify correct behavior of
file resources when they include utf-8 characters.

### DIFF
--- a/acceptance/tests/utf8/utf8-in-file-resource.rb
+++ b/acceptance/tests/utf8/utf8-in-file-resource.rb
@@ -1,0 +1,55 @@
+test_name 'utf-8 characters in resource title and param values' do
+  confine :except, :platform => [ 'windows', 'ubuntu-16']   # PUP-6983
+  utf8chars = "€‰ㄘ万竹ÜÖ"
+  agents.each do |agent|
+    agent_file = agent.tmpfile("file" + utf8chars) 
+    # remove this file, so puppet can create it and not merely correct
+    # its drift.
+    on(agent, "rm -rf #{agent_file}")
+
+    manifest =
+<<PP
+
+file { "#{agent_file}" :
+  ensure => file,
+  mode => "0644",
+  content => "This is the file content. file #{utf8chars} 
+",
+}
+
+PP
+
+    step "Apply manifest" do
+      apply_manifest_on(
+        agent,
+        manifest,
+        {:acceptable_exit_codes => [0, 2], :catch_failures => true, }
+      )
+      result = on(agent, "cat #{agent_file}")
+      assert_equal(result.exit_code, 0)
+      assert_match(
+        /#{utf8chars}/,
+        result.stdout,
+        "result stdout did not contain"
+      )
+    end
+
+    step "Drift correction" do
+      on(agent, "> #{agent_file}")
+      result = on(agent, "cat #{agent_file}")
+      assert_equal(result.stdout, "", "expected empty file")
+      apply_manifest_on(
+        agent,
+        manifest,
+        {:acceptable_exit_codes => [0, 2], :catch_failures => true, }
+      )
+      result = on(agent, "cat #{agent_file}")
+      assert_match(
+        /#{utf8chars}/,
+        result.stdout,
+        "result stdout did not contain"
+      )
+    end
+  end
+end
+


acceptance/tests/utf8/utf8-in-file-resource.rb
```diff
@@ -0,0 +1,55 @@
...
+      assert_equal(result.exit_code, 0)
+      assert_match(
+        /#{utf8chars}/,
+        result.stdout,
+        "result stdout did not contain"
+      )
+    end
+
+    step "Drift correction" do
+      on(agent, "> #{agent_file}")
```
er0ck: never new you could redirect nothing to a file this way.  it's kinda mysterious what is going on.  you might want to use `touch` here to make it more readable

seangriff: Touch won't work.  The point is to change the content (by emptying it). Touch will leave the content unchanged.

er0ck: sorry, thought it was a new file.

acceptance/tests/utf8/utf8-in-file-resource.rb
```diff
@@ -0,0 +1,55 @@
+test_name 'utf-8 characters in resource title and param values' do
+  confine :except, :platform => [ 'windows', 'ubuntu-16']   # PUP-6983
+  utf8chars = "€‰ㄘ万竹ÜÖ"
```
MosesMendoza: to prevent various possible errors in display / parse, I think its best to use the "\u"-style notation here for characters, i.e., "\u2603" (snowman)

acceptance/tests/utf8/utf8-in-file-resource.rb
```diff
@@ -0,0 +1,55 @@
...
+    # its drift.
+    on(agent, "rm -rf #{agent_file}")
+
+    manifest =
+<<PP
+
+file { "#{agent_file}" :
+  ensure => file,
+  mode => "0644",
+  content => "This is the file content. file #{utf8chars}
```
MosesMendoza: In the context of the comment above about using "\u" notation, to ensure the interpolation happens we can also use heredoc with \u escape sequences enabled, i.e., 

https://docs.puppet.com/puppet/latest/lang_data_string.html#enabling-escape-sequences

If that, this would be something like:
```ruby
$file_content = @("doc"/$u)
This is the utf8 file content: #{utf8chars}
| doc

...
mode => "0644",
content => $file_content, 
}
```

acceptance/tests/utf8/utf8-in-file-resource.rb
Iristyle: Yes, please do that. It is also helpful to know what the Unicode code points are so that they compared to bytes / other encodings / etc.

acceptance/tests/utf8/utf8-in-file-resource.rb
```diff
@@ -0,0 +1,55 @@
...
+",
+}
+
+PP
+
+    step "Apply manifest" do
+      apply_manifest_on(
+        agent,
+        manifest,
+        {:acceptable_exit_codes => [0, 2], :catch_failures => true, }
```
MosesMendoza: some of our vmpooler acceptance hosts don't have their locale set to utf-8 by default, which (has and) will cause myriad issues with these utf-8 chars. To ensure the behavior we're checking for, until we've fixed it so that puppet on *nix handles utf-8 encoding even when its environment is not utf-8, we need to specify the encoding via environment here. An example is `acceptance/tests/resource/user/utf8_user_comments.rb`, added here: https://github.com/puppetlabs/puppet/pull/5419/files

acceptance/tests/utf8/utf8-in-file-resource.rb
MosesMendoza: Just for posterity, this fails on cisco nx, because to escape the cisco management interface and get to bash, commands are prefixed with `ip netns exec management`.

The redirection here is interpreted as redirecting the output of a command `ip netns exec management` to the file
(but there isn't a command to run to redirect the output of). The output looked like:
```
01:29:25       * Drift correction
01:29:25         
01:29:25         chtifw7vi39jwhs.delivery.puppetlabs.net (cisconx-64-1) 09:29:24$ source /etc/profile; sudo sh -c " ip netns exec management > /tmp/file€‰ㄘ万竹ÜÖ.cA1zbl "
01:29:25           No cmd specified
01:29:25         
01:29:25         chtifw7vi39jwhs.delivery.puppetlabs.net (cisconx-64-1) executed in 0.08 seconds
01:29:25         Exited: 1
```

PR #5538, opened 2017-01-24
(PUP-7089) Fix problem with hiera5_backend and missing paths

thallgren:
A hiera5_backend declared in such a way that no valid paths got configured
would cause a `undefined method `empty?' for nil:NilClass` problem when
the backend was configure. This commit ensures that this cannot happen.


hlindberg:
good to merge when tests go green (appveyor slow)

PR #5544, opened 2017-01-25
(maint) Add Ruby 2.4 to Travis / AppVeyor

Iristyle:


Iristyle:
Scoping this out to see if anything fails in Ruby 2.4 that doesn't fail in Ruby 2.3.  Determining if it would be better to use Ruby 2.3 or Ruby 2.4 for upcoming Puppet 5.

Iristyle:
Oh so that's fun @geoffnichols / @MosesMendoza - 228 failures when run on Ruby 2.4 - https://travis-ci.org/puppetlabs/puppet/jobs/195041501

thallgren:
@Iristyle, I'm looking into the language related failures.

thallgren:
Fixnum and Bignum is unified into Integer in Ruby 2.4. This causes some problems. Captured here: https://tickets.puppetlabs.com/browse/PUP-7130. There seems to be 29 failures that are unrelated to this though that has to do with SSL.

hlindberg:
we will address the Fixnum related issues found in the next sprint

Iristyle:
@adrienthebo would you want to take a look at the SSL related failures?

.travis.yml
```diff
@@ -2,12 +2,14 @@ language: ruby
 sudo: false
 before_install:
   - gem install bundler -v 1.10
+  - gem update --system --no-doc
```
Iristyle: Note that is unfortunately due to the `rainbow` gem and some incompatibilities with rubygems, requiring rubygems `2.6.9` or higher.

That issue is tracked here - https://github.com/sickill/rainbow/issues/44

Might be worth seeing what is pulling in rainbow and if it can be removed in any way, as an alternate option.

adrienthebo:
FWIW our rdoc dependency is pulling in the json gem, which doesn't build against ruby 2.4.0. We'll need to chase this detail down at some point.

Iristyle:
@adrienthebo as part of Puppet 5, we've talked about removing the reference to `json` altogether given it's been shipped inside the Ruby runtime for a very long time now.  The problem will be tracking down all the other gem dependencies which refer to it as well.  Don't know about the rest of our dependency tree, but it looks like `rdoc 4.3.0` was the first to stop explicitly mentioning `json` - see https://rubygems.org/gems/rdoc/versions/4.3.0

Iristyle:
On the OpenSSL related issues, it's possible that adding support for OpenSSL 1.0.1 changed Ruby interfaces - not sure yet - but see https://bugs.ruby-lang.org/issues/12324 / https://github.com/ruby/ruby/compare/trunk...rhenium:feature/openssl-110-v1

adrienthebo:
See https://github.com/puppetlabs/puppet/pull/5555 for my current fixes for Ruby 2.4.0 + Puppet::SSL

Iristyle:
https://github.com/puppetlabs/puppet/pull/5556 has the changes related to `BigNum` / `FixNum` consolidating into `Integer`

thallgren:
@Iristyle, actually #5556 is not about BigNum/FixNum. It's about capping integer values.

Iristyle:
Ah thanks @thallgren - I thought they were one and the same - I'll pay better attention next time!

MikaelSmith:
Is this expected to pass now?

thallgren:
Yes, it should pass now.

Iristyle:
Well, except that AppVeyor doesn't have Ruby 2.4 yet (and the commit msg should have been updated prior to merge).

AppVeyor is waiting on RubyInstaller to release a 2.4 installation before they can move the ball forward on 2.4 support.  But RubyInstaller is also looking for community help to produce the builds.

PR #5384, opened 2016-11-16
(PUP-7134) Use Puppet::FileSystem for file access / explicitly specify encodings for SSL files

Iristyle:
~~Builds on #5370 which must be merged first~~

PR #5545, opened 2017-01-25
(PUP-6925) Use Puppet::FileSystem for file access / explicitly specify encodings

Iristyle:
For encodings that are currently specified implicitly (through Rubys `Encoding.default_external` for instance), make sure they are specified explicitly instead.

Should be merged prior to #5384

### DIFF
--- a/lib/puppet/file_system/windows.rb
+++ b/lib/puppet/file_system/windows.rb
@@ -3,6 +3,14 @@ require 'puppet/util/windows'
 
 class Puppet::FileSystem::Windows < Puppet::FileSystem::Posix
 
+  def open(path, mode, options, &block)
+    # PUP-6959 mode is explicitly ignored until it can be implemented
+    # Ruby on Windows uses mode for setting file attributes like read-only and
+    # archived, not for setting permissions like POSIX
+    raise TypeError.new('mode must be specified as an Integer') if mode && !mode.is_a?(Numeric)
+    ::File.open(path, options, nil, &block)
+  end
+
   def expand_path(path, dir_string = nil)
     # ensure `nil` values behave like underlying File.expand_path
     string_path = ::File.expand_path(path.nil? ? nil : path_string(path), dir_string)
--- a/lib/puppet/provider/user/directoryservice.rb
+++ b/lib/puppet/provider/user/directoryservice.rb
@@ -622,6 +622,7 @@ Puppet::Type.type(:user).provide :directoryservice do
 
   # This is a simple wrapper method for writing values to a file.
   def write_to_file(filename, value)
+    Puppet.deprecation_warning("Puppet::Type.type(:user).provider(:directoryservice).write_to_file is deprecated and will be removed in Puppet 5.")
     begin
       File.open(filename, 'w') { |f| f.write(value)}
     rescue Errno::EACCES => detail
--- a/lib/puppet/provider/zone/solaris.rb
+++ b/lib/puppet/provider/zone/solaris.rb
@@ -263,7 +263,10 @@ Puppet::Type.type(:zone).provide(:solaris) do
 
       unless Puppet::FileSystem.exist?(sysidcfg)
         begin
-          File.open(sysidcfg, "w", 0600) do |f|
+          # For compatibility reasons use System encoding for this OS file
+          # the manifest string is UTF-8 so this could result in conversion errors
+          # which should propagate to users
+          Puppet::FileSystem.open(sysidcfg, 0600, "w:#{Encoding.default_external.name}") do |f|
             f.puts cfg
           end
         rescue => detail
--- a/lib/puppet/resource/catalog.rb
+++ b/lib/puppet/resource/catalog.rb
@@ -538,7 +538,10 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
 
   # Store the classes in the classfile.
   def write_class_file
-    ::File.open(Puppet[:classfile], "w") do |f|
+    # classfile paths may contain UTF-8
+    # https://docs.puppet.com/puppet/latest/reference/configuration.html#classfile
+    classfile = Puppet.settings.setting(:classfile)
+    Puppet::FileSystem.open(classfile.value, classfile.mode.to_i(8), "w:UTF-8") do |f|
       f.puts classes.join("\n")
     end
   rescue => detail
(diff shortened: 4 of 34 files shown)


lib/puppet/provider/user/directoryservice.rb
```diff
@@ -619,13 +619,4 @@ def set_salted_pbkdf2(users_plist, shadow_hash_data, field, value)
   def write_users_plist_to_disk(users_plist)
     Puppet::Util::Plist.write_plist_file(users_plist, "#{users_plist_dir}/#{@resource.name}.plist", :binary)
   end
-
-  # This is a simple wrapper method for writing values to a file.
-  def write_to_file(filename, value)
```
MosesMendoza: should this removal be a deprecation instead? doesn't appear to be api-private

lib/puppet/provider/zone/solaris.rb
```diff
@@ -263,7 +263,8 @@ def start
 
       unless Puppet::FileSystem.exist?(sysidcfg)
         begin
-          File.open(sysidcfg, "w", 0600) do |f|
+          # For compatibility reasons use System encoding for this OS file
+          Puppet::FileSystem.open(sysidcfg, "w:#{Encoding.default_external.name}", 0600) do |f|
```
MosesMendoza: It looks like the content of `sysidcfg` file is coming in as a zone resource property value, which will be UTF-8. Any risk to this conversion, you think?

MosesMendoza: It looks like the content of `sysidcfg` file is coming in as a zone resource property value, which will be UTF-8. Any risk to this conversion, you think?

lib/puppet/resource/catalog.rb
```diff
@@ -538,7 +538,10 @@ def filter(&block)
 
   # Store the classes in the classfile.
   def write_class_file
-    ::File.open(Puppet[:classfile], "w") do |f|
+    # classfile paths may contain UTF-8
+    # https://docs.puppet.com/puppet/latest/reference/configuration.html#classfile
+    classfile = Puppet.settings.setting(:classfile)
+    Puppet::FileSystem.open(classfile.value, classfile.mode.to_i(8), "w:UTF-8") do |f|
```
MosesMendoza: There was some recent question about our mode handling on Windows - will this do the right thing there?

spec/integration/util/execution_spec.rb
```diff
@@ -17,6 +17,14 @@
     end
   end
 
+  describe "#execute (non-Windows)", :if => !Puppet.features.microsoft_windows? do
```
MosesMendoza: wow.

lib/puppet/provider/user/directoryservice.rb
Iristyle: You could make a case for that, and usually I would.  However, in this case, I think we're OK.

* `write_to_file` was introduced by Gary / Jeff back in 2012 https://github.com/puppetlabs/puppet/commit/960ab6b03e32ed770ecdab4bf2bc269d532b8001 and was called by `write_sha1_hash` which in turn was called by the `password=` setter
* In https://github.com/Iristyle/puppet/commit/efd2a45646eaedd3aa744ebcf7beb94739653db5 / [PUP-3553](https://tickets.puppetlabs.com/browse/PUP-3553) the `password=` setter was completely rewritten when we dropped support for OSX < 10.6.  It appears that this `write_to_file` method was accidentally left in at that point, as it no longer had any callers - see full file at https://github.com/Iristyle/puppet/blob/efd2a45646eaedd3aa744ebcf7beb94739653db5/lib/puppet/provider/user/directoryservice.rb
* You could argue that this is still public API - however, this a very specific provider implementation on OSX, so while not impossible, I would be very surprised if anyone was using it.  Specs used to use it prior to PUP-3553, but its usage there was removed as well.

If you feel really strongly that we should deprecate this method to be safe, we can do that too.

I should probably update commit message with the above info regardless of whether we decide on removing or deprecating / removing in 5.

lib/puppet/provider/zone/solaris.rb
Iristyle: Interesting - did you look at a Solaris machine... or did you find docs specifying it somewhere?

I was basically trying to mimic what we were already doing - but if we have docs that say this is always UTF-8, happy to switch.

I should also change `Encoding.default_external` to `Encoding.find('filesystem')`, whose existence I only recently learned of if we're not doing `Encoding::UTF_8`

lib/puppet/provider/user/directoryservice.rb
MosesMendoza: That context is helpful. I think I'd still be +1 to deprecating even in lieu of it. It'll only be around for another minor version, and I can sleep better the night of the 4.9.0 release 😄

lib/puppet/resource/catalog.rb
Iristyle: That's a great question.

This implementation will do exactly what `::File.open` does - see https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_system/file_impl.rb#L30-L32

That said, it's likely not going to be what we want on Windows, as this will set Windows file attributes (read-only, archive, etc bits) rather than security permissions.  I will change this to only apply mode on non-Windows, with a note referencing the ticket that fixes our `Puppet::FileSystem` implementation to use the mode value for permissions (I think we have a ticket somewhere for that)

lib/puppet/provider/zone/solaris.rb
Iristyle: On second thought `Encoding::find('filesystem')` is used for strings in file names / paths **not** file contents per https://docs.ruby-lang.org/en/2.1.0/Encoding.html#class-Encoding-label-Filesystem+encoding

So this either needs to change to `utf-8` or remain as `Encoding.default_external.name` (since we're passing in a string)

MosesMendoza: Re: `sysidcfg` I was just going off the type reference, https://docs.puppet.com/puppet/latest/type.html#zone-attribute-sysidcfg. I was thinking since this will be coming in as a manifest value that it would be utf-8. Is that right?

I didn't know about `Encoding.find('filesystem')`. Cool.

lib/puppet/util/rdoc/generators/puppet_generator.rb
```diff
@@ -225,7 +225,7 @@ def gen_top_index(collection, title, template, filename)
         'style_url'  => style_url('', @options.css),
       }
 
-      File.open(filename, "w") do |f|
+      Puppet::FileSystem.open(filename, "w:UTF-8") do |f|
```
MosesMendoza: I think we need to pass `nil` explicitly as the second argument for these two.

lib/puppet/file_system/windows.rb
```diff
@@ -3,6 +3,13 @@
 
 class Puppet::FileSystem::Windows < Puppet::FileSystem::Posix
 
+  def open(path, mode, options, &block)
+    # PUP-6959 mode is explicitly ignored until it can be implemented
+    # Ruby on Windows uses mode for setting file attributes like read-only and
+    # archived, not for setting permissions like POSIX
+    ::File.open(path, options, nil, &block)
```
MosesMendoza: I think it would be good to add a test asserting this change.

lib/puppet/resource/catalog.rb
MosesMendoza: Note for posterity, decision was to move the application of ignoring mode down a level into `Puppet::FileSystem::Windows`

Iristyle:
🎉

PR #5548, opened 2017-01-25
(maint) Exclude vendored code from Puppet's gettext

Magisus:
The vendored semantic_puppet gem handles its own internationalization.
This commit adds a line to Puppet's gettext-setup config file which
causes it to ignore strings within semantic_puppet.

### DIFF
--- a/locales/config.yaml
+++ b/locales/config.yaml
@@ -24,3 +24,6 @@ gettext:
   # content, relative to the project root directory
   exclude_files:
     - 'lib/puppet/pops/types/type_formatter.rb'
+    # The semantic_puppet gem is temporarily vendored (PUP-7114), and it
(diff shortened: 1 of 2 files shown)


locales/config.yaml
```diff
@@ -24,3 +24,6 @@ gettext:
   # content, relative to the project root directory
   exclude_files:
     - 'lib/puppet/pops/types/type_formatter.rb'
+    # The semantic_puppet gem is temporarily vendored (PUP-7114), and it
+    # handles its own translation.
+    - 'lib/semantic_puppet/*.rb'
```
Iristyle: Can you use globs here?  We have both `lib/semantic_puppet.rb` and `lib/semantic_puppet/**/*` I believe.

PR #5550, opened 2017-01-26
(PUP-7118) New Acceptance test: utf-8 characters in function args

seangriff:
This PR adds a new acceptance test to verify correct behavior
when utf-8 characters are in arguments to puppet functions.


er0ck:
@seangriff this needs do/end closures on the test_name and steps

seangriff:
File updated, look now please.
thanks.

acceptance/tests/utf8/utf8-in-function-args.rb
```diff
@@ -0,0 +1,298 @@
...
+        )
+      end
+    
+      step 'assert_type' do
+        on(
+          agent,
+          puppet(
+            "apply",
+            "-e",
+            "'notice(assert_type(String, \"#{utf8chars}\"))'"
```
er0ck: nice.

PR #5552, opened 2017-01-27
(PUP-7114) Move semantic_puppet to lib/puppet/vendor

thallgren:
The vendored semantic_puppet ended up directly under /lib in the puppet
repository. The decision to put it there seems to be based on the
misunderstanding to otherwise be forced to change `require` statements.

Since that's not needed, this commit moves the semantic_puppet source
to lib/puppet/vendor to avoid future confusion.


MosesMendoza:
This was originally merged in as https://github.com/puppetlabs/puppet/pull/5512. 

We chatted about it briefly and somehow came to the conclusion that there was some code that was loaded before we do `require puppet/vendor' that required semantic_puppet. This PR passes specs cleanly which clearly implies that wasn't the case. I'm nervous though - is there some chance that we do need `semantic_puppet` before it is loaded via `puppet/vendor` ?

thallgren:
@MosesMendoza expressed some concerns regarding requiring `semantic_puppet` even before the vendor packages were loaded so I investigated that and concluded that there's no risk of that. Reasons being:

- In Puppet 4.8.1, Puppet didn't use SemanticPuppet at all.
- The vendor sources are loaded in puppet.rb.
- At present, only the module tool and code under puppet/pops uses it. Both are initialized after the vendor sources are loaded.

Iristyle:
Odd - @MosesMendoza suggested this initially, but I ran into problems with the gem in the location that you've moved it to.  I'll have to revisit.

hlindberg:
@Iristyle would not have worked well unless the feature hook that adds the vendored lib to the path was also present. Could it have been that?
Waiting with merging until you give a thumbs up.

Iristyle:
@hlindberg I definitely did not add the `lib/puppet/vendor/load_semantic_puppet.rb` as Thomas has - so it could have been my ignorance.

hlindberg:
thanks @Iristyle, it would not have worked well/at all without that hook. So, with that, and the fact that @thallgren did look into where the gem is used - I am merging this.

PR #5553, opened 2017-01-27
(PUP-7102) Log a warning when environment path has permission problems

thallgren:
Prior to this PR, a permission problem in one of the directories in
an `environmentpath` would go unnoticed. The directory would simply be
deemed invalid and excluded from the path. This would later lead to
an `EnvironmentNotFound` exception being raised with no clue as to why
the path wasn't included.

This PR ensures that the directory with permission problems is logged
as a warning (once). The warning can be silenced using a `disable_warnings`
setting that includes 'permission_denied'.

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -93,7 +93,8 @@ module Puppet
 
         * `deprecations` --- disables deprecation warnings.
         * `undefined_variables` --- disables warnings about non existing variables.
-        * `undefined_resources` --- disables warnings about non existing resources.",
+        * `undefined_resources` --- disables warnings about non existing resources.
+        * `permission_deined` --- disables warnings about directories with access problems.",
       :hook      => proc do |value|
         values = munge(value)
         valid   = %w[deprecations undefined_variables undefined_resources]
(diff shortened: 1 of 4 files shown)


hlindberg:
Didn't we say on the ticket that this should be an error? I wonder if there is a legit usecase where some users should skip over unreachable environments only to find one later in the envpath that they can access. Sounds like a sure way to screw things up. Most likely users don't have such a setup, so it always ends with "not found". Then, when on an agent doing 'puppet agent -t', you don't get to see the warning, so user still has no clue to the "not found" without going digging in the logs.

hlindberg:
appveyor also has a problem with this PR it seems.

thallgren:
The ticket description and acceptance criteria only mentions that messages indicating that parts of the environmentpath is inaccessible should make it into the log. Looking deeper I discovered your comment about always make this an error. If that's concluded, then it would be great if the acceptance criteria was also updated.

Interesting to see that a file name containing asterisks gives an EINVAL error on Windows. I'll add a check for that.

thallgren:
Succeeded by [PR-5650[(/puppetlabs/puppet/pull/5650).

PR #5555, opened 2017-01-27
(maint) SSL fixups for Ruby 2.4.0 

adrienthebo:
Further validation on these changes are needed; they work but I don't fully understand what changed to necessitate these fixes.

### DIFF
--- a/spec/unit/ssl/certificate_request_spec.rb
+++ b/spec/unit/ssl/certificate_request_spec.rb
@@ -92,14 +92,13 @@ describe Puppet::SSL::CertificateRequest do
 
     it "should set the subject to [CN, name]" do
       request.generate(key)
-      # OpenSSL::X509::Name only implements equality as `eql?`
-      expect(request.content.subject).to eql OpenSSL::X509::Name.new([['CN', key.name]])
+      expect(request.content.subject).to eq OpenSSL::X509::Name.new([['CN', key.name]])
     end
 
     it "should set the CN to the :ca_name setting when the CSR is for a CA" do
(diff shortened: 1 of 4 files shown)


spec/unit/ssl/certificate_request_spec.rb
```diff
@@ -92,14 +92,13 @@
 
     it "should set the subject to [CN, name]" do
       request.generate(key)
-      # OpenSSL::X509::Name only implements equality as `eql?`
-      expect(request.content.subject).to eql OpenSSL::X509::Name.new([['CN', key.name]])
+      expect(request.content.subject).to eq OpenSSL::X509::Name.new([['CN', key.name]])
```
Iristyle: So that remark was left in https://github.com/puppetlabs/puppet/commit/97c8b0c5eb95cee1bb08571512b258179d61da9f#diff-dc34e6762244dae4d2f0c66adfbe4cd8R93

stdlib docs mention this about `eql` - https://ruby-doc.org/stdlib-2.0.0/libdoc/openssl/rdoc/OpenSSL/X509/Name.html#method-i-eql-3F

> Returns true if name and other refer to the same hash key.

Clearly the tests pass and you mention the reason in the commit msg - just making sure you had reviewed the previous remarks.

Iristyle:
Ok @adrienthebo - we can probably fixup the commit message now to reference the Ruby bug https://bugs.ruby-lang.org/issues/13170 - PR against Ruby 2.4.0 OpenSSL at https://github.com/ruby/openssl/pull/100

PR #5556, opened 2017-01-27
(PUP-3170) Cap integer size

hlindberg:
This modifies the runtime3_converter to raise errors when converting integer values out of 64 bit range, or BigDecimal values out of range for Float.

### DIFF
--- a/lib/puppet/pops.rb
+++ b/lib/puppet/pops.rb
@@ -10,7 +10,7 @@ module Puppet
   #
   # @api public
   module Pops
-     EMPTY_HASH = {}.freeze
+    EMPTY_HASH = {}.freeze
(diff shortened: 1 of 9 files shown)


hlindberg:
Do you mean where the .pp source text becomes an Integer value? (That can be done) or, do you mean when the evaluator performs arithmetic (that means overflow in temporary values in chains of arithmetic - which, I guess (sigh) that we really should do since in puppet a Maxint + 1 - 1 should really fail.
What I added in the PR does produce file/line information when it errors (for the call/resource attribute set at least).

thallgren:
I mean both I guess. An illegal constant should be invalid as well as an expression that produces an illegal value.

PR #5558, opened 2017-01-30
(pe-19366) Gather resource (package) inventory

ploubser:
Here we add the ability to gather package inventory information and add
it to the report.

This functionality defaults to being disabled, and can be enabled by
setting the `pe_enable_resource_inventory` configuration option to `true`.
When enabled Puppet::Indirector.search will be used to gather a list of
all package resources and add it to a new "resource_inventory" field in
the report.

While resource gathering is currently limited to pacakges, the intent is
to add more resources in the future.

### DIFF
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -165,6 +165,18 @@
         "corrective_change": {
             "description": "True if a change or noop event in this report was caused by an unexpected change to the system between Puppet runs.",
             "type":        "boolean"
+        },
+
+        "resource_inventory": {
+            "description": "Object with one property per resource-type collected for inventory.",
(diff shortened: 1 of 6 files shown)


ploubser:
Ping @scottyw

MosesMendoza:
to keep the threads together, there is some historical commentary on an earlier version of this change set, here: https://github.com/ploubser/puppet/commit/e3b0e5fbf8b8cb8cd32a19dd9cebb9e5f253fc65

ploubser:
I'm closing this PR for now while we explore other options (as discussed in the email thread).

Thanks everyone for the feedback.

PR #5559, opened 2017-01-30
(PUP-7042) Ignore translations in unit tests

Magisus:
This PR should be merged before the other string externalization PRs, to avoid spec failures.

The first commit:
...stubs out the `_` translate method in unit tests, where it interferes
with some of our mocking. This is a temporary solution, which precludes
testing any translations in unit tests. Ideally we will be able to find
a way to redo our mocks in such a way that they are not triggered by the
inner workings of FastGettext.

The second commit:
... edits some strings to avoid extracting timestamps into the
    POT file, which cause unneccessary churn, as well as fixes a mistake in
    a previously marked string.


MosesMendoza:
@Magisus the pr description mentions two commits but there's just one in the PR. I think it makes sense to separate the externalization from stubbing `_`, would you mind splitting them out?

Magisus:
Updated.

Magisus:
Closing and opening new to fix Travis.

PR #5560, opened 2017-01-30
(PUP-5645) run gem install with HOME enviromnent

flokli:
At least on Archlinux, the gem install command fails otherwise with

> wrong argument type nil (expected Regexp)


flokli:
@geoffnichols, @branan: ping?

flokli:
https://github.com/puppetlabs/puppet/pull/5051 should also fix this.

Magisus:
We are going to close this, in favor of #5051.

PR #5561, opened 2017-01-30
(PUP-7042) Mark strings in puppet.rb and lib/puppet/type/zpool.rb

Magisus:
This commit marks user-facing strings for translation in a couple of
files, chosen more or less at random as a way to test the workflow.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -1,7 +1,7 @@
 require 'puppet/version'
 
 if RUBY_VERSION < "1.9.3"
-  raise LoadError, "Puppet #{Puppet.version} requires ruby 1.9.3 or greater."
+  raise LoadError, _("Puppet #{Puppet.version} requires ruby 1.9.3 or greater.")
 end
 
 Puppet::OLDEST_RECOMMENDED_RUBY_VERSION = '2.1.0'
@@ -164,7 +164,7 @@ module Puppet
   # Now that settings are loaded we have the code loaded to be able to issue
   # deprecation warnings. Warn if we're on a deprecated ruby version.
   if RUBY_VERSION < Puppet::OLDEST_RECOMMENDED_RUBY_VERSION
-    Puppet.deprecation_warning("Support for ruby version #{RUBY_VERSION} is deprecated and will be removed in a future release. See https://docs.puppet.com/puppet/latest/system_requirements.html#ruby for a list of supported ruby versions.")
+    Puppet.deprecation_warning(_("Support for ruby version #{RUBY_VERSION} is deprecated and will be removed in a future release. See https://docs.puppet.com/puppet/latest/system_requirements.html#ruby for a list of supported ruby versions."))
   end
 
   # Initialize puppet's settings. This is intended only for use by external tools that are not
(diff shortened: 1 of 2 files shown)


lib/puppet.rb
```diff
@@ -1,7 +1,7 @@
 require 'puppet/version'
 
 if RUBY_VERSION < "1.9.3"
-  raise LoadError, "Puppet #{Puppet.version} requires ruby 1.9.3 or greater."
+  raise LoadError, _("Puppet #{Puppet.version} requires ruby #{RUBY_VERSION} or greater.")
```
MosesMendoza: I don't think we want `RUBY_VERSION` here - 1.9.3 was the right one.

lib/puppet.rb
```diff
@@ -211,7 +211,7 @@ def self.initialize_facts
   # code was deprecated in 2008, but this is still in heavy use.  I suppose
   # this can count as a soft deprecation for the next dev. --daniel 2011-04-12
   def self.newtype(name, options = {}, &block)
-    Puppet.deprecation_warning("Creating #{name} via Puppet.newtype is deprecated and will be removed in a future release. Use Puppet::Type.newtype instead.")
+    Puppet.deprecation_warning(_("Creating #{name} via Puppet.newtype is deprecated and will be removed in a future release. Use Puppet::Type.newtype instead."))
```
MosesMendoza: Still getting up to speed on externalization so have patience - but is this how we're supposed to externalize strings with interpolated variables? Will translators know to ignore the part inside `#{...}` (or should they, even)?

Magisus: I am very much hoping they can be asked to ignore those parts. If not, we have an loootttt more work ahead of us. The interpolation functions correctly if the `#{...}` is left alone: Ruby performs the interpolation after it loads the translated string.

MosesMendoza: Yeah it seems to work, I just want to make sure our translators can handle that and won't try to translate the variable name inside `#{...}` I'm shooting an email over to @LarissaLane.

Just in case anyone else is following along, the resulting pot file entry looks like:
```
msgid "\"Creating #{name} via Puppet.newtype is deprecated and will be removed in a future release. Use Puppet::Type.newtype instead.\""
msgstr ""
```

Magisus: Alright, thanks!

LarissaLane: Something that's become clear over the past few weeks as I've been working with translators on classifier service strings is that every time we have a variable we need to add a comment to translators to explain what the variable is a placeholder for, unless it is completely obvious. For example "Message: {1}" is obvious. @Magisus We may be able to get help with this as part of the string review. cc @kayayarai 

Having variables in strings is fine @MosesMendoza but I need to know the patterns we use so I can document them in [these translator instructions](https://docs.google.com/document/d/1NPa1uZYD8gqlMx7guuZlji6TheMfrQYRCjtJQxYg_10/edit?usp=sharing)

lib/puppet/type/zpool.rb
```diff
@@ -85,7 +85,7 @@ def insync?(is)
 
     validate do
       has_should = [:disk, :mirror, :raidz].select { |prop| self.should(prop) }
-      self.fail "You cannot specify #{has_should.join(" and ")} on this type (only one)" if has_should.length > 1
+      self.fail _("You cannot specify #{has_should.join(" and ")} on this type (only one)") if has_should.length > 1
```
MosesMendoza: This is kind of dicey - strictly speaking it adheres to the rules, but it results in this string to be translated:
```
msgid "\"You cannot specify #{has_should.join(\" and \")} on this type (only one)\""
```
As long as the translator's not confused by the double quotes inside the string it's ok.

Magisus:

Magisus:

PR #5563, opened 2017-01-30
(PUP-7042) Mark strings in transaction

Magisus:
This commit marks user-facing error and info level strings in
`lib/puppet/transaction.rb` and `lib/puppet/transaction/*` for
translation.

### DIFF
--- a/lib/puppet/transaction.rb
+++ b/lib/puppet/transaction.rb
@@ -79,7 +79,7 @@ class Puppet::Transaction
       prerun_errors.each do |res, detail|
         res.log_exception(detail)
       end
-      raise Puppet::Error, "Some pre-run checks failed"
+      raise Puppet::Error, _("Some pre-run checks failed")
     end
   end
 
@@ -95,7 +95,7 @@ class Puppet::Transaction
 
     persistence.load if catalog.host_config?
 
-    Puppet.info "Applying configuration version '#{catalog.version}'" if catalog.version
+    Puppet.info _("Applying configuration version '#{catalog.version}'") if catalog.version
 
     continue_while = lambda { !stop_processing? }
 
@@ -118,7 +118,7 @@ class Puppet::Transaction
       # is one, it must have been selected by the user.
       return if missing_tags?(resource)
       if resource.provider
-        resource.err "Provider #{resource.provider.class.name} is not functional on this host"
+        resource.err _("Provider #{resource.provider.class.name} is not functional on this host")
       else
         providerless_types << resource.type
       end
@@ -134,14 +134,14 @@ class Puppet::Transaction
     teardown = lambda do
       # Just once per type. No need to punish the user.
       providerless_types.uniq.each do |type|
-        Puppet.err "Could not find a suitable provider for #{type}"
+        Puppet.err _("Could not find a suitable provider for #{type}")
       end
 
       post_evalable_providers.each do |provider|
         begin
           provider.post_resource_eval
         rescue => detail
-          Puppet.log_exception(detail, "post_resource_eval failed for provider #{provider}")
+          Puppet.log_exception(detail, _("post_resource_eval failed for provider #{provider}"))
         end
       end
 
@@ -157,11 +157,11 @@ class Puppet::Transaction
                                 :canceled_resource_handler => canceled_resource_handler,
                                 :teardown => teardown) do |resource|
       if resource.is_a?(Puppet::Type::Component)
-        Puppet.warning "Somehow left a component in the relationship graph"
+        Puppet.warning _("Somehow left a component in the relationship graph")
       else
-        resource.info "Starting to evaluate the resource" if Puppet[:evaltrace] and @catalog.host_config?
+        resource.info _("Starting to evaluate the resource") if Puppet[:evaltrace] and @catalog.host_config?
         seconds = thinmark { block.call(resource) }
-        resource.info "Evaluated in %0.2f seconds" % seconds if Puppet[:evaltrace] and @catalog.host_config?
+        resource.info _("Evaluated in %0.2f seconds") % seconds if Puppet[:evaltrace] and @catalog.host_config?
       end
     end
 
@@ -231,7 +231,7 @@ class Puppet::Transaction
     add_resource_status(status)
     event_manager.queue_events(ancestor || resource, status.events) unless status.failed?
   rescue => detail
-    resource.err "Could not evaluate: #{detail}"
+    resource.err _("Could not evaluate: #{detail}")
   end
 
   # Evaluate a single resource.
@@ -261,7 +261,7 @@ class Puppet::Transaction
       # See above. --daniel 2011-06-06
       unless suppress_report then
         s.failed_dependencies.each do |dep|
-          resource.notice "Dependency #{dep} has failures: #{resource_status(dep).failed}"
+          resource.notice _("Dependency #{dep} has failures: #{resource_status(dep).failed}")
         end
       end
     end
@@ -318,7 +318,8 @@ class Puppet::Transaction
     begin
       provider_class.prefetch(resources)
     rescue LoadError, Puppet::MissingCommand => detail
-      Puppet.log_exception(detail, "Could not prefetch #{type_name} provider '#{provider_class.name}': #{detail}")
+      #TRANSLATORS `prefetch` is a function name and should not be translated
+      Puppet.log_exception(detail, _("Could not prefetch #{type_name} provider '#{provider_class.name}': #{detail}"))
     end
     @prefetched_providers[type_name][provider_class.name] = true
   end
@@ -346,7 +347,7 @@ class Puppet::Transaction
       # like class and stage.  This is undesirable; while just skipping the
       # output isn't perfect, it is RC-safe. --daniel 2011-06-07
       unless resource.class == Puppet::Type.type(:whit) then
-        resource.warning "Skipping because of failed dependencies"
+        resource.warning _("Skipping because of failed dependencies")
       end
     elsif resource.virtual?
       resource.debug "Skipping because virtual"
--- a/lib/puppet/transaction.rb
+++ b/lib/puppet/transaction.rb
@@ -95,7 +95,7 @@ class Puppet::Transaction
 
     persistence.load if catalog.host_config?
 
-    Puppet.info _("Applying configuration version '#{catalog.version}'") if catalog.version
+    Puppet.info _("Applying configuration version '%{version}'") % { version: catalog.version } if catalog.version
 
     continue_while = lambda { !stop_processing? }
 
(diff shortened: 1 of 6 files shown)


lib/puppet/transaction.rb
```diff
@@ -157,11 +157,11 @@ def evaluate(&block)
...
                                 :teardown => teardown) do |resource|
       if resource.is_a?(Puppet::Type::Component)
-        Puppet.warning "Somehow left a component in the relationship graph"
+        Puppet.warning _("Somehow left a component in the relationship graph")
       else
-        resource.info "Starting to evaluate the resource" if Puppet[:evaltrace] and @catalog.host_config?
+        resource.info _("Starting to evaluate the resource") if Puppet[:evaltrace] and @catalog.host_config?
         seconds = thinmark { block.call(resource) }
-        resource.info "Evaluated in %0.2f seconds" % seconds if Puppet[:evaltrace] and @catalog.host_config?
+        resource.info _("Evaluated in %0.2f seconds") % seconds if Puppet[:evaltrace] and @catalog.host_config?
```
MosesMendoza: @LarissaLane is there a special way we're supposed to mark units of measure like time? ie here we have seconds

lib/puppet/transaction.rb
```diff
@@ -134,14 +134,14 @@ def evaluate(&block)
...
-        Puppet.err "Could not find a suitable provider for #{type}"
+        Puppet.err _("Could not find a suitable provider for #{type}")
       end
 
       post_evalable_providers.each do |provider|
         begin
           provider.post_resource_eval
         rescue => detail
-          Puppet.log_exception(detail, "post_resource_eval failed for provider #{provider}")
+          Puppet.log_exception(detail, _("post_resource_eval failed for provider #{provider}"))
```
MosesMendoza: @LarissaLane added a note to the doc https://docs.google.com/document/d/1NPa1uZYD8gqlMx7guuZlji6TheMfrQYRCjtJQxYg_10/edit to avoid translating `_` separated terms like this one

lib/puppet/transaction.rb
```diff
@@ -322,7 +322,7 @@ def prefetch(provider_class, resources)
     begin
       provider_class.prefetch(resources)
     rescue LoadError, Puppet::MissingCommand => detail
-      Puppet.log_exception(detail, "Could not prefetch #{type_name} provider '#{provider_class.name}': #{detail}")
+      Puppet.log_exception(detail, _("Could not prefetch #{type_name} provider '#{provider_class.name}': #{detail}"))
```
MosesMendoza: wow this is going to be a hard string to translate. 

`prefetch` is a specific function name / term in puppet and shouldn't be translated anywhere - but in this string it's not marked uniquely or separately from any other regular word to be translated. Is there some way we're supposed to notify the translator of these kinds of terms? Maybe a code comment that is pulled into the POT file? cc @LarissaLane

lib/puppet/transaction.rb
```diff
@@ -350,7 +350,7 @@ def skip?(resource)
       # like class and stage.  This is undesirable; while just skipping the
       # output isn't perfect, it is RC-safe. --daniel 2011-06-07
       unless resource.class == Puppet::Type.type(:whit) then
-        resource.warning "Skipping because of failed dependencies"
+        resource.warning _("Skipping because of failed dependencies")
       end
     elsif resource.virtual?
       resource.debug "Skipping because virtual"
```
MosesMendoza: Looks like this string was missed

MosesMendoza: Or is this only warnings, not debug output

lib/puppet/transaction.rb
Magisus: A comment would be good, I think. In your opinion, does this message need to be translated at all?

lib/puppet/transaction.rb
Magisus: We are not doing debug right now. You'll see lots of those skipped.

lib/puppet/transaction/event_manager.rb
```diff
@@ -147,19 +147,19 @@ def process_callback(resource, callback, events)
     resource.send(callback)
 
     if not resource.is_a?(Puppet::Type.type(:whit))
-      resource.notice "Triggered '#{callback}' from #{events.length} events"
+      resource.notice _("Triggered '#{callback}' from #{events.length} events")
```
MosesMendoza: I think for externalizing strings with pluralization, we should take a different form. In this case, I think it looks something like:

```ruby
resource.notice n_("Triggered '#{callback}' from %{count} event", "Triggered '#{callback}' from %{count} events", events.length) % { :count => events.length }
```

though strictly speaking that would change this string to include `event` singular or `events` plural even in English depending on the length of the `events` array... but I think thats ok given we're pushing this towards pup 5

lib/puppet/transaction/resource_harness.rb
```diff
@@ -235,9 +235,9 @@ def noop(event, param, current_value, audit_message)
   def sync(event, param, current_value, audit_message)
     param.sync
     if param.sensitive
-      event.message = param.format("changed %s to %s#{audit_message}", param.is_to_s(current_value), param.should_to_s(param.should))
+      event.message = param.format(_("changed %s to %s#{audit_message}"), param.is_to_s(current_value), param.should_to_s(param.should))
     else
-      event.message = "#{param.change_to_s(current_value, param.should)}#{audit_message}"
+      event.message = _("#{param.change_to_s(current_value, param.should)}#{audit_message}")
```
MosesMendoza: This string shouldn't get wrapped as its composed entirely of variables.

lib/puppet/transaction/event_manager.rb
Magisus: +1 to taking this opportunity to improve the English too.

Magisus:

lib/puppet/transaction/resource_harness.rb
```diff
@@ -144,17 +144,17 @@ def sync_if_needed(param, context)
 
       event = create_change_event(param, current_value, historical_value)
       event.status = "failure"
-      event.message = param.format("change from %s to %s failed: #{detail}",
+      event.message = param.format(_("change from %s to %s failed: "),
```
Magisus: So I think I've narrowed the test failures down to these lines (and the ones like them below). I've gone through a number of iterations trying to make this syntax happy, but no luck so far. Thoughts?

MosesMendoza: @Magisus you are on the right track. We can't just include `detail` in the string that's passed to `param.format` because it gets `[redacted]`. So it has to be appended later as you're doing...except we have to cast it as a string to do so, because the `+` operator won't do it for us, like string interpolation does. This is a bit of ruby being inconsistently helpful:

```
[5] pry(#<Puppet::Transaction::ResourceHarness>)> "foo" + detail
TypeError: no implicit conversion of StandardError into String
from (pry):5:in `+'
[6] pry(#<Puppet::Transaction::ResourceHarness>)> "foo #{detail}"
=> "foo StandardError"
[7] pry(#<Puppet::Transaction::ResourceHarness>)> "foo " + detail.to_s
=> "foo StandardError"
```
 i.e.
```ruby
      event.message = param.format(_("change from %s to %s failed: "),
                                   param.is_to_s(current_value),
                                   param.should_to_s(param.should)) + detail.to_s
```

Magisus: Thanks @MosesMendoza that makes sense!

Magisus:

lib/puppet/transaction/additional_resource_generator.rb
```diff
@@ -51,13 +51,13 @@ def generate_additional_resources(resource)
 
   def eval_generate(resource)
     return false unless resource.respond_to?(:eval_generate)
-    raise Puppet::DevError,"Depthfirst resources are not supported by eval_generate" if resource.depthfirst?
+    raise Puppet::DevError, "Depthfirst resources are not supported by eval_generate" if resource.depthfirst?
```
MosesMendoza: Missed wrap?

lib/puppet/transaction/additional_resource_generator.rb
```diff
@@ -51,13 +51,13 @@ def generate_additional_resources(resource)
...
     begin
       generated = replace_duplicates_with_catalog_resources(resource.eval_generate)
       return false if generated.empty?
     rescue => detail
       @resources_failed_to_generate = true
-      resource.log_exception(detail, "Failed to generate additional resources using 'eval_generate': #{detail}")
+      resource.log_exception(detail, _("Failed to generate additional resources using 'eval_generate': %{detail}") % { detail: detail })
```
MosesMendoza: Maybe add a translator comment to not translate `eval_generate` ?

lib/puppet/transaction/event_manager.rb
```diff
@@ -147,19 +147,19 @@ def process_callback(resource, callback, events)
     resource.send(callback)
 
     if not resource.is_a?(Puppet::Type.type(:whit))
-      resource.notice "Triggered '#{callback}' from #{events.length} events"
+      resource.notice n_("Triggered '#{callback}' from %{count} event", "Triggered '#{callback}' from %{count} events", events.length) % { :count => events.length }
```
MosesMendoza: Interpolation on `callback` is still using regular style vs externalized

lib/puppet/transaction/event_manager.rb
```diff
@@ -147,19 +147,19 @@ def process_callback(resource, callback, events)
...
+    resource.err _("Failed to call %{callback}: %{detail}") % { callback: callback, detail: detail }
 
     transaction.resource_status(resource).failed_to_restart = true
     resource.log_exception(detail)
     return false
   end
 
   def process_noop_events(resource, callback, events)
-    resource.notice "Would have triggered '#{callback}' from #{events.length} events"
+    resource.notice _("Would have triggered '%{callback}' from %{count} events") % { callback: callback, count: events.length }
```
MosesMendoza: This could use the same pluralization form as line 150

lib/puppet/transaction/additional_resource_generator.rb
Magisus: I was actually curious, do most of these `DevErrors` wind up in front of users?

MosesMendoza: I think they're intended for users writing their own puppet code, types/providers, and functions. I.e., in this case if someone wrote a type that they intended to generate resources, they might come across this.

lib/puppet/transaction/event_manager.rb
```diff
@@ -147,19 +147,19 @@ def process_callback(resource, callback, events)
...
+    resource.err _("Failed to call %{callback}: %{detail}") % { callback: callback, detail: detail }
 
     transaction.resource_status(resource).failed_to_restart = true
     resource.log_exception(detail)
     return false
   end
 
   def process_noop_events(resource, callback, events)
-    resource.notice "Would have triggered '#{callback}' from #{events.length} events"
+    resource.notice n_("Would have triggered '%{callback}' from %{count} event", "Would have triggered '%{callback}' from %{count} events", events.length) % { :count => events.length, callback: callback }
```
MosesMendoza: nit: This hash uses two different hash syntaxes for different keys - should probably just use one (seems like the newer form is dominant here)

PR #5572, opened 2017-01-31
(PUP-7042) Mark strings in face

Magisus:
This commit marks user-facing error and info strings in
`lib/puppet/face.rb` and `lib/puppet/face/*` for translation.

### DIFF
--- a/lib/puppet/face/ca.rb
+++ b/lib/puppet/face/ca.rb
@@ -6,7 +6,7 @@ Puppet::Face.define(:ca, '0.1.0') do
 
   summary _("Local Puppet Certificate Authority management.")
 
-  description _(<<-TEXT)
+  description <<-TEXT
     This provides local management of the Puppet Certificate Authority.
 
     You can use this subcommand to sign outstanding certificate requests, list
@@ -16,7 +16,7 @@ Puppet::Face.define(:ca, '0.1.0') do
   action :list do
     summary _("List certificates and/or certificate requests.")
 
-    description _(<<-TEXT)
+    description <<-TEXT
       This will list the current certificates and certificate signing requests
       in the Puppet CA.  You will also get the fingerprint, and any certificate
       verification failure reported.
@@ -41,7 +41,7 @@ Puppet::Face.define(:ca, '0.1.0') do
     option "--subject " + _("PATTERN") do
       summary _("Only list if the subject matches PATTERN.")
 
-      description _(<<-TEXT)
+      description <<-TEXT
         Only include certificates or requests where subject matches PATTERN.
 
         PATTERN is interpreted as a regular expression, allowing complex
@@ -50,8 +50,10 @@ Puppet::Face.define(:ca, '0.1.0') do
     end
 
     when_invoked do |options|
+      #TRANSLATORS "CA" stands for "certificate authority"
       raise _("Not a CA") unless Puppet::SSL::CertificateAuthority.ca?
       unless ca = Puppet::SSL::CertificateAuthority.instance
+        #TRANSLATORS "CA" stands for "certificate authority"
         raise _("Unable to fetch the CA")
       end
       Puppet::SSL::Host.ca_location = :only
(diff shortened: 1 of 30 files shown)


Magisus:

Magisus:

lib/puppet/face/ca.rb
```diff
@@ -50,9 +50,9 @@
     end
 
     when_invoked do |options|
-      raise "Not a CA" unless Puppet::SSL::CertificateAuthority.ca?
+      raise _("Not a CA") unless Puppet::SSL::CertificateAuthority.ca?
```
MosesMendoza: Thoughts on translator notes to ignore CA here and below? Maybe not necessary since it's an abbreviation?

lib/puppet/face/ca.rb
```diff
@@ -206,15 +206,15 @@
   end
 
   action :fingerprint do
-    summary "Print the DIGEST (defaults to the signing algorithm) fingerprint of a host's certificate."
-    option "--digest ALGORITHM" do
-      summary "The hash algorithm to use when displaying the fingerprint"
+    summary _("Print the DIGEST (defaults to the signing algorithm) fingerprint of a host's certificate.")
```
MosesMendoza: Can you add a translator note that DIGEST refers to a hash algorithm?

lib/puppet/face/certificate.rb
```diff
@@ -3,9 +3,9 @@
 
 Puppet::Indirector::Face.define(:certificate, '0.0.1') do
   copyright "Puppet Labs", 2011
-  license   "Apache 2 license; see COPYING"
+  license   _("Apache 2 license; see COPYING")
 
-  summary "Provide access to the CA for certificate management."
+  summary _("Provide access to the CA for certificate management.")
   description <<-EOT
```
MosesMendoza: The prior changes have the description field heredoc marker wrapped, but several (maybe all) of the remaining do not. Do you want to wrap those too?

lib/puppet/face/module/list.rb
```diff
@@ -140,7 +140,7 @@ def warn_unmet_dependencies(environment)
             parent_version     = dep[:parent][:version]
 
             msg = "'#{parent_name}' (#{parent_version})"
-            msg << " requires '#{dep_name}' (#{version_constraint})"
+            msg << _(" requires '%{dep_name}' (%{version_constraint})") % { dep_name: dep_name, version_constraint: version_constraint }
```
MosesMendoza: An you add a translator note that these two lines are concatenated to form one sentence w/example of what the output would be?

lib/puppet/face/module/uninstall.rb
```diff
@@ -69,9 +69,8 @@
         exit 1
       else
         mod = return_value[:affected_modules].first
-        "Removed '#{return_value[:module_name]}'" <<
-        (mod.version ? " (#{colorize(:cyan, mod.version.to_s.sub(/^(?=\d)/, 'v'))})" : '') <<
-        " from #{mod.modulepath}"
+        module_version = mod.version ? "(#{colorize(:cyan, mod.version.to_s.sub(/^(?=\d)/, 'v'))})" : ''
+        _("Removed '%{name}' %{module_version} from %{path}") % { name: return_value[:module_name], module_version: module_version, path: mod.modulepath }
```
MosesMendoza: Thanks for cleaning this up.

One minor change - I the previous lines accommodated the extra space in the dynamically added content so that regardless of how it came together there weren't extraneous spaces between words. Can you do the same?

lib/puppet/face/parser.rb
```diff
@@ -113,7 +113,7 @@
         if missing_files.empty?
           dumps
         else
-          dumps + "One or more file(s) specified did not exist:\n" + missing_files.collect { |f| "   #{f}" }.join("\n")
+          dumps + _("One or more file(s) specified did not exist:\n") + missing_files.collect { |f| "   #{f}" }.join("\n")
```
MosesMendoza: I expect translators won't mess with embedded newlines - does that seem right to you?

lib/puppet/face/ca.rb
Magisus: I think I might actually make a note that CA stands for certificate authority, and let them decide how to handle it. Might be nice to give it meaning, rather than not translate it.

lib/puppet/face/certificate.rb
Magisus: I was told to leave docs-like strings alone. So I should probably go back and unwrap the previous ones...

lib/puppet/face/module/list.rb
Magisus: Do you think we could just put that all in one string? I can't see any reason for it to be split up, and it may be untranslatable this way.

lib/puppet/face/parser.rb
Magisus: Yeah I would assume not.

Magisus:
Updated.

lib/puppet/face/module/list.rb
MosesMendoza: 👍

PR #5593, opened 2017-01-31
(PUP-7151) Read selinux module version as BINARY

Iristyle:
- Previously SELinux modules were parsed by using the Ruby File.new API
   without specifying 'binary' mode, nor using Ruby block form.

 - Update the code to use Puppet::FileSystem and to specify its a
   binary read.  Further, use the block format of the call to close
   the file handle as soon its done being read


MosesMendoza:
I think the commit message first bullet might have a couple types? Otherwise 👍

PR #5594, opened 2017-01-31
(PUP-7152) AIX encrypted user passwords are ASCII

Iristyle:
- According to AIX documentation for the /etc/security/passwd file
   it is in an ASCII encoding
 - Given -e is passed while calling chpasswd, which assumes an
   already encrypted password, the contents should be written to
   /etc/security/passwd directly by AIX.  Therefore, that value should
   also be ASCII

### DIFF
--- a/lib/puppet/provider/user/aix.rb
+++ b/lib/puppet/provider/user/aix.rb
@@ -216,7 +216,9 @@ Puppet::Type.type(:user).provide :aix, :parent => Puppet::Provider::AixObject do
 
   def open_security_passwd
     # helper method for tests
-    File.open("/etc/security/passwd", 'r')
+    # AIX reference indicates this file is ASCII
+    # https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/com.ibm.aix.files/passwd_security.htm
+    Puppet::FileSystem.open("/etc/security/passwd", nil, "r:ASCII")
   end
 
   #--------------------------------
@@ -255,7 +257,10 @@ Puppet::Type.type(:user).provide :aix, :parent => Puppet::Provider::AixObject do
     user = @resource[:name]
 
     # Puppet execute does not support strings as input, only files.
-    tmpfile = Tempfile.new('puppet_#{user}_pw')
+    # The password is expected to be in an encrypted format given -e is specified:
+    # https://www.ibm.com/support/knowledgecenter/ssw_aix_71/com.ibm.aix.cmds1/chpasswd.htm
+    # /etc/security/passwd is specified as an ASCII file per the AIX documentation
+    tmpfile = Tempfile.new("puppet_#{user}_pw", :encoding => Encoding::ASCII)
     tmpfile << "#{user}:#{value}\n"
     tmpfile.close()
 


lib/puppet/provider/user/aix.rb
```diff
@@ -255,7 +257,11 @@ def password=(value)
     user = @resource[:name]
 
     # Puppet execute does not support strings as input, only files.
-    tmpfile = Tempfile.new('puppet_#{user}_pw')
+    # The password is expected to be in an encrypted format given -e is specified:
+    # https://www.ibm.com/support/knowledgecenter/ssw_aix_71/com.ibm.aix.cmds1/chpasswd.htm
+    # /etc/security/passwd is specified as an ASCII file per the AIX documentation
+    # TODO: Is there a reason we don't simply pipe the output to chpasswd like `echo foo | chpasswd`
```
MosesMendoza: Good question. @branan you've been toolin around on an AIX box recently. Any chance you could check this?

lib/puppet/provider/user/aix.rb
```diff
@@ -216,7 +216,9 @@ def expiry_from_attr(value)
 
   def open_security_passwd
     # helper method for tests
-    File.open("/etc/security/passwd", 'r')
+    # AIX reference indicates this file is ASCII
+    # https://www.ibm.com/support/knowledgecenter/en/ssw_aix_72/com.ibm.aix.files/passwd_security.htm
+    Puppet::FileSystem.open("/etc/security/passwd", nil, "r:ASCII")
```
MosesMendoza: would be good to add a spec?

lib/puppet/provider/user/aix.rb
branan: piping the password would make it visible in the process table. The tempfile seems right, and the coercion to ASCII seems reasonable here :+1:

branan: We should probably validate that the tempfile has sane permissions, though (600)

Iristyle: I wish GitHub had the (facepalm) emoticon... thanks @branan - totally forgot about the process table.

If you check the perms I can update the PR to remove the comments.

branan: Permissions look good, although the line below has a bug

```
[0] [AIX] root@pe-aix-72-acceptance:~ # ls -l '/tmp/puppet_#{user}_pw20170201-8716668-n0747r'
-rw-------    1 root     system            9 Feb 01 13:59 /tmp/puppet_#{user}_pw20170201-8716668-n0747r
```

lib/puppet/provider/user/aix.rb
```diff
@@ -255,7 +257,11 @@ def password=(value)
     user = @resource[:name]
 
     # Puppet execute does not support strings as input, only files.
-    tmpfile = Tempfile.new('puppet_#{user}_pw')
+    # The password is expected to be in an encrypted format given -e is specified:
+    # https://www.ibm.com/support/knowledgecenter/ssw_aix_71/com.ibm.aix.cmds1/chpasswd.htm
+    # /etc/security/passwd is specified as an ASCII file per the AIX documentation
+    # TODO: Is there a reason we don't simply pipe the output to chpasswd like `echo foo | chpasswd`
+    tmpfile = Tempfile.new('puppet_#{user}_pw', :encoding => Encoding::ASCII)
```
branan: This needs to be double-quoted, or the variable doesn't expand.

Iristyle: Sweet - accidental existing bug discovery!

branan: Yeah, I don't even want to know how far back the `git blame` on that goes :). Would be nice to get it fixed up while we're in here, though.

lib/puppet/provider/user/aix.rb
Iristyle: A regular spec would be of dubious value I think - would only serve as double book-keeping.

I think there are 2 existing acceptances which probably cover this (at least indirectly) as they set a password:
* https://github.com/puppetlabs/puppet/blob/ebd96213cab43bb2a8071b7ac0206c3ed0be8e58/acceptance/tests/resource/user/should_modify_while_managing_home.rb
* https://github.com/puppetlabs/puppet/blob/ebd96213cab43bb2a8071b7ac0206c3ed0be8e58/acceptance/tests/resource/user/should_modify_when_not_managing_home.rb

lib/puppet/provider/user/aix.rb
Iristyle: Fixed

PR #5596, opened 2017-02-01
(PUP-7098) Updated acceptance test for utf-8 characters in catalog

seangriff:

### DIFF
--- a/acceptance/tests/utf8/utf8-in-catalog.rb
+++ b/acceptance/tests/utf8/utf8-in-catalog.rb
@@ -1,13 +1,26 @@
 test_name 'utf-8 characters in cached catalog' do
-  # utf8chars = "€‰ㄘ万竹ÜÖ"
+  confine :except, :platform => [
+    'windows',      # PUP-6983
+    'cumulus',      # PUP-7147
+    'cisco_ios',    # PUP-7150
+    'eos-4',        # PUP-7146
+    'aix',          # PUP-7194
+    'huawei',       # PUP-7195
+  ]
+
+  utf8chars_lit = "€‰ㄘ万竹ÜÖ"
   utf8chars = "\u20ac\u2030\u3118\u4e07\u7af9\u00dc\u00d6"
   file_content = "This is the file content. file #{utf8chars}"
-  tmpdir = master.tmpdir("code")
-  on(master, "rm -rf #{tmpdir}")
-  env_dir = "#{tmpdir}/environments"
+  codedir = master.tmpdir("code")
+  on(master, "rm -rf #{codedir}")
+  env_dir = "#{codedir}/environments"
   agents.each do |agent|
+    puts "agent name: #{agent.hostname}, platform: #{agent.platform}"
+    agent_vardir = agent.tmpdir("agent_vardir")
     agent_file = agent.tmpfile("file" + utf8chars) 
-
+    teardown do
+      on(agent, "rm -rf #{agent_vardir} #{agent_file}")
+    end
     step "Apply manifest" do
       on(
         agent,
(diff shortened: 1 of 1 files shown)


MosesMendoza:

seangriff:
@MosesMendoza

acceptance/tests/utf8/utf8-in-catalog.rb
```diff
@@ -0,0 +1,132 @@
...
+      'main' => {
+        'environmentpath' => "#{env_dir}",
+      },
+      'agent' => {
+        'use_cached_catalog' => 'true'
+      }
+    }
+    
+    with_puppet_running_on(master, master_opts, tmpdir) do 
+      step "puppet agent -t" do
```
MosesMendoza: It might be helpful if we're debugging test output if this step description included info about what the puppet run was intended to accomplish?

acceptance/tests/utf8/utf8-in-catalog.rb
```diff
@@ -0,0 +1,132 @@
...
+      end
+    
+      step "verify cached catalog" do
+        result = on(
+          agent,
+          puppet("config print vardir"),
+          :environment => {:LANG => "en_US.UTF-8"}
+        )
+        catalog_file_name =
+          "#{result.stdout.strip}/client_data/catalog/#{agent.hostname}.json"
```
MosesMendoza: Strictly speaking, puppet names the catalog file after the `node_name_value` fact, which may not always be the hostname. Beaker provides a way to access this, `agent.node_name`

acceptance/tests/utf8/utf8-in-catalog.rb
```diff
@@ -0,0 +1,132 @@
...
+            :acceptable_exit_codes => [0, 2],
+            :environment => {:LANG => "en_US.UTF-8"}
+          }
+        )
+      end
+    
+      step "verify cached catalog" do
+        result = on(
+          agent,
+          puppet("config print vardir"),
```
MosesMendoza: Rather than having to do an extra ssh call out to the agent to gets its `vardir`, you can just tell the agent what its vardir is with `--vardir` in the preceding `puppet("agent -t"...` call. Saves us both the extra time of the call and yet another possible ssh/network failure. 

An example of this being done is here:
https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/direct_puppet/catalog_uuid_correlates_catalogs_with_reports.rb#L8-L30

acceptance/tests/utf8/utf8-in-catalog.rb
```diff
@@ -0,0 +1,132 @@
...
+          :environment => {:LANG => "en_US.UTF-8"}
+        )
+        catalog_file_name =
+          "#{result.stdout.strip}/client_data/catalog/#{agent.hostname}.json"
+  
+        result = on(
```
MosesMendoza: Similar to my comment on https://github.com/puppetlabs/puppet/pull/5597#pullrequestreview-20651399, I think these would really benefit from using the block form of `#on`

acceptance/tests/utf8/utf8-in-catalog.rb
```diff
@@ -0,0 +1,132 @@
...
+  
+      step "apply cached catalog" do
+        on(
+          agent,
+          puppet("resource file #{agent_file} ensure=absent"),
+          :environment => {:LANG => "en_US.UTF-8"}
+        )
+        on(
+          agent,
+          puppet("catalog apply --terminus json"),
```
MosesMendoza: Neat. I like that this is doing this instead of `puppet agent --use_cached_catalog --one-time --no-daemonize` as I don't expect we have much acceptance coverage of the `catalog` face anywhere

acceptance/tests/utf8/utf8-in-catalog.rb
```diff
@@ -0,0 +1,132 @@
...
+    
+        result = on(
+          agent,
+          "cat #{agent_file}",
+          :environment => {:LANG => "en_US.UTF-8"}
+        )
+        assert_match(
+          /#{utf8chars}/,
+          result.stdout,
+          "result stdout did not contain"
```
MosesMendoza: should this string be appended with the missing content?

acceptance/tests/utf8/utf8-in-catalog.rb
seangriff: changed to "apply utf-8 catalog"

acceptance/tests/utf8/utf8-in-catalog.rb
seangriff: changed to using node_name method.

PR #5598, opened 2017-02-01
(PUP-7118) New Acceptance test: utf-8 characters in function args

seangriff:

### DIFF
--- a/acceptance/tests/utf8/utf8-in-function-args.rb
+++ b/acceptance/tests/utf8/utf8-in-function-args.rb
@@ -1,6 +1,12 @@
 test_name 'utf-8 characters in function parameters' do
-  confine :except, :platform => [ 'windows', 'ubuntu-16']   # PUP-6983
-  utf8chars = "€‰ㄘ万竹ÜÖ"
+  confine :except, :platform => [
+    'windows',      # PUP-6983
+    'eos-4',        # PUP-7146
+    'cumulus',      # PUP-7147
+    'cisco',        # PUP-7150
+  ]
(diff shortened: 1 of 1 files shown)


MosesMendoza:

seangriff:
@MosesMendoza

seangriff:
This PR will fix the failure on the Jenkins build today (PUP-7172)

acceptance/tests/utf8/utf8-in-function-args.rb
```diff
@@ -14,6 +20,7 @@
         result = on(
           agent,
           puppet("apply", "-e" "'alert(\"alert #{utf8chars}\")'"),
+          :environment => {:LANG => "en_UTF-8"}
```
MosesMendoza: @seangriff I think we want these to be `:LANG => "en_US.UTF-8"`? At least on Ubuntu 16.04 that's the case:

```ruby
☂ : LANG="en_UTF-8" ruby -e "puts Encoding.default_external"
US-ASCII
☂ : locale-gen "en_UTF-8"
Error: 'en_UTF-8' is not a supported language or locale
☂ : locale-gen "en_US.UTF-8"
Generating locales (this might take a while)...
  en_US.UTF-8... done
Generation complete.
☂ : LANG="en_US.UTF-8" ruby -e "puts Encoding.default_external"
UTF-8
```

Down below later in the test it looks like you do indeed switch over to `en_US.UTF-8` ie, line 62 and later

seangriff:
output of successfull run:
https://gist.github.com/seangriff/fe851317f79843d38db7dec086d61f98

PR #5599, opened 2017-02-01
(PUP-7119) New Acceptance test: utf-8 characters in 'puppet describe'

seangriff:

### DIFF
--- a/acceptance/tests/utf8/utf8-in-puppet-describe.rb
+++ b/acceptance/tests/utf8/utf8-in-puppet-describe.rb
@@ -1,0 +1,89 @@
+test_name 'utf-8 characters in module doc string, puppet describe' do
+  platforms = hosts.map {|val| val[:platform]}
+  if (platforms.any? { |val| /^eos-/ =~ val})
+    skip_test "Skipping because Puppet describe fails when the Arista module is installed (ARISTA-51)"
+  end 
+
+  # utf8chars = "€‰ㄘ万竹ÜÖ"
+  utf8chars = "\u20ac\u2030\u3118\u4e07\u7af9\u00dc\u00d6"
+
+  master_mod_dir = master.tmpdir("describe_master")
+  on(master, "chmod -R 755 #{master_mod_dir}");
+  teardown do
+    on(master, "rm -rf #{master_mod_dir}")
+  end
+  master_manifest = 
+<<MASTER_MANIFEST
+
+File {
+  ensure => directory,
+  mode => "0755",
+}
+
+file {
+  '#{master_mod_dir}/code':;
+  '#{master_mod_dir}/code/environments':;
+  '#{master_mod_dir}/code/environments/production':;
+  '#{master_mod_dir}/code/environments/production/modules':;
+  '#{master_mod_dir}/code/environments/production/modules/master_mytype_module':;
+  '#{master_mod_dir}/code/environments/production/modules/master_mytype_module/lib':;
+  '#{master_mod_dir}/code/environments/production/modules/master_mytype_module/lib/puppet':;
+  '#{master_mod_dir}/code/environments/production/modules/master_mytype_module/lib/puppet/type':;
+}
+
+file { '#{master_mod_dir}/code/environments/production/modules/master_mytype_module/lib/puppet/type/master_mytype.rb' :
+  ensure => file,
+  mode => '0755',
+  content => '
+Puppet::Type.newtype(:master_mytype) do
+  @doc = "Testing to see if puppet handles describe blocks correctly
+when they contain utf8 characters, such as #{utf8chars}
+"
+  newparam(:name) do
+    isnamevar
+    desc " name parameter for mytype, also with some utf8 chars #{utf8chars}"
+  end
+end
+',
+}
+
+MASTER_MANIFEST
+
+  step "Apply master manifest" do
+    apply_manifest_on(master, master_manifest)
+  end
+  master_opts = {
+    'main' => {
+       'environmentpath' => "#{master_mod_dir}/code/environments",
+    }
+  }
+
+  step "Start puppet server"
+  with_puppet_running_on(master, master_opts, master_mod_dir) do
+    agents.each do |agent|
+      puts "agent name: #{agent.hostname}, platform: #{agent.platform}"
+      step "Run puppet agent for plugin sync" do 
+        on(
+          agent, puppet("agent", "-t", "--server #{master.node_name}"),
+          :acceptable_exit_codes => [0, 2]
+        )
+      end
+
+      step "Puppet describe for master-hosted mytype" do 
+        on(agent, puppet("describe", "master_mytype")) do |result|
+          assert_match(
+            /master_mytype.*such as #{utf8chars}/m,
+            result.stdout,
+  "Main description of master_mytype did not match utf8 chars, '#{utf8chars}'"
+          )
+
+          assert_match(
+            /name parameter.*chars #{utf8chars}/,
+            result.stdout,
+            "Name parameter description of master_mytype did not match utf8 chars, '#{utf8chars}'"
+          )
+        end
+      end
+    end
+  end
+end


MosesMendoza:

seangriff:
@MosesMendoza

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,71 @@
...
+      result = on(
+        agent,
+        puppet(
+          "describe",
+          "describe",
+          "--environmentpath #{mod_dir}/environments"
+        ),
+        :enviornment => {:LANG => "en_US.UTF-8"}
+      )
+      assert_equal(result.exit_code, 0, "puppet describe failed.")
```
MosesMendoza: This line shouldn't be needed - [Beaker::DSL::Helpers::HostHelpers#on](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Helpers/HostHelpers#on-instance_method) defaults to `:acceptable_exit_codes => 0`

MosesMendoza: This line shouldn't be needed - [Beaker::DSL::Helpers::HostHelpers#on](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/DSL/Helpers/HostHelpers#on-instance_method) defaults to `:acceptable_exit_codes => 0`

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    step "puppet describe" do
+      result = on(
+        agent,
+        puppet(
+          "describe",
+          "describe",
+          "--environmentpath #{mod_dir}/environments"
+        ),
+        :enviornment => {:LANG => "en_US.UTF-8"}
```
MosesMendoza: Spelling - s/:enviornment/environment

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,71 @@
...
+PP
+
+    step "Apply manifest" do
+      apply_manifest_on(
+        agent,
+        manifest,
+        {
+          :acceptable_exit_codes => [0, 2],
+          :catch_failures => true,
+          :enviornment => {:LANG => "en_US.UTF-8"}
```
MosesMendoza: Spelling - s/:enviornment/environment

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,71 @@
+test_name 'utf-8 characters in module doc string, puppet describe' do
+  # utf8chars = "€‰ㄘ万竹ÜÖ"
+  utf8chars = "\u20ac\u2030\u3118\u4e07\u7af9\u00dc\u00d6"
+  agents.each do |agent|
+    mod_dir = agent.tmpdir("describe")
```
MosesMendoza: I think this test should have a teardown step to remove this directory

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,80 @@
+platforms = hosts.map {|val| val[:platform]}
+if (platforms.any? { |val| /^eos-/ =~ val})
+  skip_test "Skipping because Puppet describe fails when the Arista module is installed (ARISTA-51)"
+end
```
er0ck: this block needs to be inside `test_name`

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,80 @@
+platforms = hosts.map {|val| val[:platform]}
+if (platforms.any? { |val| /^eos-/ =~ val})
+  skip_test "Skipping because Puppet describe fails when the Arista module is installed (ARISTA-51)"
+end 
+
+test_name 'utf-8 characters in module doc string, puppet describe' do
+  # utf8chars = "€‰ㄘ万竹ÜÖ"
+  utf8chars = "\u20ac\u2030\u3118\u4e07\u7af9\u00dc\u00d6"
+  agents.each do |agent|
+    puts "agent name: #{agent.hostname}, platform: #{agent.platform}"
```
er0ck: why is this important?

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,80 @@
...
+    mod_dir = agent.tmpdir("describe")
+    on(agent, "chmod -R 777 #{mod_dir}")
+    teardown do
+      on(agent, "rm -rf #{mod_dir}")
```
er0ck: should the `mod_dir` only be created on the master?

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,80 @@
...
+      ) do
+        assert_match(
+          /#{utf8chars}/, stdout, "describe did not match utf8 chars"
+        )
+      end
+    end
+  end
+end
+
+
```
er0ck: extra newlines

acceptance/tests/utf8/utf8-in-puppet-describe.rb
seangriff: I was testing with a hosts.yaml file that specified ~40 different agent types.  A similar test failed on a small subset of them.  Adding that output made it easier to tell which platforms had passed, and how far along through the set of agents the test had come, and which platform failed.  I guess it is not that important here.  Should I remove it?

er0ck: no need to remove

seangriff:
I committed a new version of this test.  Differences:
1.  The skip-guard for Arista was move to within the test_name block.
2. The module definition containing the type under test was moved from the agent to the master
3.  Extra lines at the end of the file were removed.

puppet describe examines resource local to the machine in order to provide a description of
the type.  This includes core types, modules in the module path and artifacts received via
plugin sync as  a result of a puppet run.

I measured the performance running the test both ways and found they take essentially the
same amount of time to run.  This means the time required to do run 'puppet agent -t' on each
agent is about the same cost as running 'puppet apply' to install the module locally.

Defining the module on the master is how the we would expect the user to implement a type.

seangriff:
Don't merge this yet.

I checked in the wrong version of the test which implements type-modules on both master and agent.  I will fix this by replacing with the desired test, implementing the module on the master only.

New commit coming.

seangriff:
Now I've got the right file in place.  This one should is the merge candidate.  Gist of successful run is here:
https://gist.github.com/seangriff/87eaef1f43b7eeb1f10a3822c28fd217

er0ck:
@seangriff can you squash down your commits?  you can leave the arista filter commit separate if you want.  most of the others seem to be related to change requests in here against the original test.

acceptance/tests/utf8/utf8-in-puppet-describe.rb
er0ck: i disagree. i tend to leave these around as they are tmpdirs, created uniquely(-ish) per run.  leaving them around aids in test debugging, where required.  

but i wouldn't hold up the merge, either way

acceptance/tests/utf8/utf8-in-puppet-describe.rb
er0ck: true.  i hand-assert on exit codes to better aid debugging with custom assertion failure messaging and use `:accept_all_exit_codes => true`.  this also forces beaker errors to become failures.

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,89 @@
...
+  '#{master_mod_dir}/code/environments/production/modules':;
+  '#{master_mod_dir}/code/environments/production/modules/master_mytype_module':;
+  '#{master_mod_dir}/code/environments/production/modules/master_mytype_module/lib':;
+  '#{master_mod_dir}/code/environments/production/modules/master_mytype_module/lib/puppet':;
+  '#{master_mod_dir}/code/environments/production/modules/master_mytype_module/lib/puppet/type':;
+}
+
+file { '#{master_mod_dir}/code/environments/production/modules/master_mytype_module/lib/puppet/type/master_mytype.rb' :
+  ensure => file,
+  mode => '0755',
```
er0ck: this is in the defaults.  no big deal i guess

acceptance/tests/utf8/utf8-in-puppet-describe.rb
```diff
@@ -0,0 +1,89 @@
...
+        on(
+          agent, puppet("agent", "-t", "--server #{master.node_name}"),
+          :acceptable_exit_codes => [0, 2]
+        )
+      end
+
+      step "Puppet describe for master-hosted mytype" do 
+        on(agent, puppet("describe", "master_mytype")) do |result|
+          assert_match(
+            /master_mytype.*\n.*\n.*\n.*such as #{utf8chars}/,
```
er0ck: this regex might be somewhat brittle in changes to `puppet describe`

it should probably use `s/ /m` multi-line modifier and reduce the `.*\n.*\n.*\n.*` to `.*`

sorry i didn't catch this before

seangriff:
Commits squashed.  Regex fixed.  @er0ck  Could you have another look now?  Thanks.

er0ck:
interesting.  github associates changes due to change requests with commits, not with line changes.  so if you squash or otherwise retroactively change commit history, that functionality breaks.  

@MosesMendoza are you good to go with this?  i'm 👍

MosesMendoza:
yes

PR #5600, opened 2017-02-01
(PUP-6588) Fix --module-repository to use full remote URL

MikaelSmith:
Fixes the `--module-repository` argument when a remote URL includes a
path, as in `http://forge:8080/modules`. This allows the module tool to
work with repositories that use paths to separate different types of
artifacts.

I'd like to retarget this at stable after releases are cleaned up.

### DIFF
--- a/lib/puppet/forge/repository.rb
+++ b/lib/puppet/forge/repository.rb
@@ -45,7 +45,7 @@ class Puppet::Forge
     # Return a Net::HTTPResponse read for this +path+.
     def make_http_request(path, io = nil)
       Puppet.debug "HTTP GET #{@host}#{path}"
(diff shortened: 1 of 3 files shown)


turbodog:
@MikaelSmith @MosesMendoza what is standing in the way of this being merged?

MosesMendoza:
@turbodog no blockers. It would be nice to get a second set of eyes but I'm comfortable merging as-is if no one else pipes up this morning.

MikaelSmith:
Should this target stable instead?

MosesMendoza:
+1 to retarget at stable

MikaelSmith:
Done.

PR #5601, opened 2017-02-01
(PUP-7158) pkg_spec.rb failure on individual execution starting c5d97e

shawnferry:
Before this change attempts to execute only pkg_spec would fail while
mocking CHILD_STATUS. When executed with other specs that initialize
CHILD_STATUS in some way execution is successful

### DIFF
--- a/spec/unit/provider/package/pkg_spec.rb
+++ b/spec/unit/provider/package/pkg_spec.rb
@@ -9,6 +9,8 @@ describe Puppet::Type.type(:package).provider(:pkg) do
     # Get a pid for $CHILD_STATUS to latch on to
     command = "cmd.exe /c \"exit 0\""
     Puppet::Util::Execution.execute(command, {:failonfail => false})
+  else
+    Puppet::Util::Execution.execute('exit 0', {:failonfail => false})
   end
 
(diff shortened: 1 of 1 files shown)


adrienthebo:
I don't understand how c5d97e triggered this bug; do you know what happened?

shawnferry:
I really don't it's just that prior to it there isn't a problem. 

My first guess was that /something/ was pre-initializing CHILD_STATUS and it no longer does but I don't see the change.

MosesMendoza:
I actually found this to fail for me when run by itself going back pretty far, but it depended on ruby version. With ruby 2.3, it failed all the way back to 4.0.0. With ruby 2.1.8, it started failing between puppet 4.2.0 and 4.3.0 (per your commit title). Oddly enough there was no diff in this provider, spec, or the spec_helper between 4.2.0 and 4.3.0. Some aspects of Puppet::Util::Execution changed, but not in a way that affects this spec. Maybe a change in one of our testing dependencies perhaps?  In any case, I think this is fine as-is, given we do a similar thing for Windows already. I'm going to retarget at 4.10.x but otherwise 👍

PR #5604, opened 2017-02-02
(PUP-7164) Handle empty hiera yaml data file gracefully

thallgren:
This commit ensures that a hiera yaml file that parses correctly but
returns something other than a hash is accepted (as an empty hash) but
with a warning.

The commit also contains a fix that made type mismatch messages repeated
twice when the exception was logged.


hlindberg:
please target stable - master is now for 5.0.0

PR #5606, opened 2017-02-03
(PUP-7123) Allow use of beaker 3.x in acceptance tests

ScottGarman:
in particular, which was added in beaker 3.10).

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -12,7 +12,7 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.5.0')
+gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.10')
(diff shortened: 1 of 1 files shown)


acceptance/Gemfile
```diff
@@ -12,7 +12,7 @@ def location_for(place, fake_version = nil)
   end
 end
 
-gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.5.0')
+gem "beaker", *location_for(ENV['BEAKER_VERSION'] || '~> 3.5')
```
Iristyle: Can we change this to `'~> 3.10'` to indicate we need a minimum Beaker version?  IMHO that more clearly conveys the need for at least a min version to support Ubuntu 16.10 (per the ticket).

It should also make anyone with an already installed gemset locally blow up if they try to `bundle exec beaker` after updating to this code, should they have an old Beaker version written into `Gemfile.lock` (which is a good thing I think)

PR #5609, opened 2017-02-03
(PUP-7174) Add 'call' to the Puppet built in language functions

seanmil:
Per PUP-7174

### DIFF
--- a/spec/unit/functions/call_spec.rb
+++ b/spec/unit/functions/call_spec.rb
@@ -1,0 +1,77 @@
+require 'puppet'
+require 'spec_helper'
+require 'puppet_spec/compiler'
+
+require 'matchers/resource'
+
+describe 'the call method' do
+  include PuppetSpec::Compiler
+  include PuppetSpec::Files
+  include Matchers::Resource
+
+  context "should be callable as" do
+    let(:env_name) { 'testenv' }
+    let(:environments_dir) { Puppet[:environmentpath] }
+    let(:env_dir) { File.join(environments_dir, env_name) }
+    let(:env) { Puppet::Node::Environment.create(env_name.to_sym, [File.join(populated_env_dir, 'modules')]) }
+    let(:node) { Puppet::Node.new("test", :environment => env) }
+    let(:env_dir_files) {
+      {
+        'modules' => {
+          'test' => {
+            'functions' => {
+              'call_me.pp' => 'function test::call_me() { "called" }'
+            }
+          }
+        }
+      }
+    }
+
+    let(:populated_env_dir) do
+      dir_contained_in(environments_dir, env_name => env_dir_files)
+      PuppetSpec::Files.record_tmp(env_dir)
+      env_dir
+    end
+
+    it 'call on a built-in 4x Ruby API function' do
+      expect(compile_to_catalog(<<-CODE)).to have_resource('Notify[a]')
+          $a = call('split', 'a-b-c', '-')
+          notify { $a[0]: }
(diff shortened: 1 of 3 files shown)


hlindberg:
Tests are failing with this error:
```
the call method should be callable as call on a Puppet language function with no arguments
     Failure/Error: raise ArgumentError, "Function #{self.class.name}(): Unknown function: '#{function_name}'" unless func_3x
     Puppet::PreformattedError:
       Evaluation Error: Error while evaluating a Function Call, Function call(): Unknown function: 'call_me' at line 2:16 on node test
```

spec/unit/functions/call_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    it 'call on a built-in 4x Ruby API function' do
+      expect(compile_to_catalog(<<-CODE)).to have_resource('Notify[a]')
+          $a = call('split', 'a-b-c', '-')
+          notify { $a[0]: }
+        CODE
+    end
+
+    it 'call on a Puppet language function with no arguments' do
+      expect(compile_to_catalog(<<-CODE)).to have_resource('Notify[called]')
+          function call_me() { 'called' }
```
hlindberg: This does not seem to take - not sure why.

seanmil: That was the bug I was referring to on Slack that I thought I had found. I've tested the call function in a full environment and it works for namedspaced Puppet language functions. However, it does not work for a non-namespaced Puppet language function. And that seems like a bug in internal_call_function to me. Based on my reading of the code and the API notes it sounds like internal_call_function should work equally for all types of functions. The loader seems unable to find it, but how/why that occurs I don't know. I started digging into the pops code, but quickly got lost.

seanmil:
@hlindberg, I spent a bit more time looking into this one spec test failing and saw what I considered to be some surprising behavior. This is all tested on 2016.4.2

If I take an empty environment (without the `call` function patch) and define this in `manifests/site.pp`:
```
function mainfunc() { 'main manifest function' }
include test
```

and then create `modules/test/manifests/init.pp` with:
```
class test() {
    notify { mainfunc(): }
}
```

Then I get `Error 500 on SERVER: Server Error: Evaluation Error: Unknown function: 'mainfunc'` error.

According to https://docs.puppet.com/puppet/4.7/lang_write_functions_in_puppet.html#aside-writing-functions-in-the-main-manifest I believe that it should work. 

Also interestingly, if I do: `puppet apply -e "function foo() { 'FOO' } ; class test { notify { foo(): } } ; include test"` it works and I get:
```
Notice: FOO
Notice: /Stage[main]/Test/Notify[FOO]/message: defined 'message' as 'FOO'
```

It also works in the agent/master scenario if I call the function directly in `site.pp`, but not from the module.

I feel like I am hitting the same (or a very similar) issue in the spec test. Given how simple the `call` function is, and that I'm seeing some odd behavior here, I don't feel it is a problem in the function itself, but rather something else (maybe the loader?)

I think I've hit the limit of what I can diagnose. Any ideas?

seanmil:
I'm not sure why the appveyor build failed - it looks to be an unrelated problem. However, I have refactored the spec tests a bit and they are all passing for all the original test cases.

hlindberg:
@seanmil the appveyor tests are failing because of a bug in a recent version of AppVeyor (errors related to UNC paths) - there are 4 failing tests (which we ignore for now).

hlindberg:
Thanks @seanmil for the contribution!

PR #5610, opened 2017-02-03
(PUP-7173) server_set facts rb failing with undefined local variable or method `master opts

er0ck:
* fix the test
  * no longer need to test the setting
  * attempts at overwriting the facts are now errors
* had to rename the test file to use our tmp_environment abstraction.  sorry for the difficult to follow diff

[skip ci]


hlindberg:
Thanks @er0ck !

PR #5611, opened 2017-02-03
(PUP-7176) Handle merge of scalar or mixed types as unique

thallgren:
This commit ensures that a 'deep' merge that encounters a mixed set of
types or non mergable types defaults to a 'first found' strategy. This
is done by simply trusting the logic in the DeepMerge gem.


hlindberg:
good to merge when tests go green

PR #5614, opened 2017-02-05
(PUP-7179) Resolve Pcore types using the same loader that loaded the type

thallgren:
When resolving a type (Object, Alias, or TypeSet), it is essential that
the loader that initially loaded the type is used. This means that
if a type `A::X` in module `a` references type `B::Y` i module `b`, which
in in turn references a type `C::Z` in module `c`, then the loader for
module `a` must be used when resolving `A::X`, the loader for module `b`
must resovle `B::Y` and finally, the loader for module `c` resolves
`C::Z`.

In essence, this means that all loaders must be found by the loader names
adapted to the source of the AST that represents the type.

Fixing this resulted in cleaner code. It is now never necessary to pass
a `scope` to the TypeParser and only tests that lacks a fully configured
`Loaders` instance needs to pass a loader.

### DIFF
--- a/lib/puppet/functions/assert_type.rb
+++ b/lib/puppet/functions/assert_type.rb
@@ -50,7 +50,7 @@
 #
 # @since 4.0.0
 #
-Puppet::Functions.create_function(:assert_type, Puppet::Functions::InternalFunction) do
+Puppet::Functions.create_function(:assert_type) do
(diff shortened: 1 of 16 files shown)


hlindberg:
This is good to merge once appveyor is green.

PR #5617, opened 2017-02-06
(PUP-7175) lookup.rb failing automatic parameter lookup with module data binding

er0ck:
* automatic data-binding to functions in modules has been removed in puppet 5
* remove this portion of the test, retain the binding via metadata.json
* this change also adds do/end closures to test_name and step
[skip ci]


MosesMendoza:
@er0ck have you addressed the feedback from @thallgren? Is this ready for merge?

er0ck:
@MosesMendoza yes, feedback addressed.  please :shipit:

PR #5618, opened 2017-02-06
(PUP-7192) Implement state machine classes for SSL initialization

adrienthebo:
This pull request defines logic for building, composing, and running state machines. The intent is to use this for reimplementing how Puppet agents submit CSRs and fetch certificates.


lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,87 @@
...
+        if current.error?
+          return :errored
+        elsif current.final?
+          return :complete
+        end
+
+        event = current.event(result)
+
+        if event.nil?
+          raise "Cannot determine transition for state #{current.name.inspect}: no event was returned"
```
adrienthebo: What exception class should we raise when a critical error is encountered when running a state machine?

lib/puppet/state_machine/state.rb
```diff
@@ -0,0 +1,109 @@
...
+          [event, Array(target).dup.unshift(ns)]
+        end]
+
+        new_options = @options.dup
+        case new_options[:type]
+        when :final
+          new_event_cb = ->(_) { :'*final*' }
+          new_options.delete(:type)
+        when :error
+          new_event_cb = ->(_) { :'*error*' }
```
adrienthebo: Error state transitions are never followed, this isn't necessary.

lib/puppet/state_machine/state.rb
```diff
@@ -0,0 +1,109 @@
...
+      def namespace(ns)
+        new_name = Array(@name).dup.unshift(ns)
+        new_transitions = Hash[@transitions.map do |(event, target)|
+          [event, Array(target).dup.unshift(ns)]
+        end]
+
+        new_options = @options.dup
+        case new_options[:type]
+        when :final
+          new_event_cb = ->(_) { :'*final*' }
```
adrienthebo: When we namespace a state with the intent to compose multiple state machines, we remove the `final` state type from final states and add a default event, `:'*final*'`, so that we can transition from a final state of one machine to another machine. Is this confusing? Is the addition/use of `:'*final*' too magical/confusing and if so what would be a better option?

lib/puppet/state_machine/state.rb
```diff
@@ -0,0 +1,109 @@
...
+      # Generate a new state with namespaced transitions.
+      #
+      # In order to compose multiple state machines into a single machine, the states of an
+      # individual machine need to be namespaced to prevent name collisions across different
+      # machines and identify which states originated from which of the individual machines.
+      #
+      # @param ns [Symbol]
+      # @return [Puppet::StateMachine::State]
+      def namespace(ns)
+        new_name = Array(@name).dup.unshift(ns)
```
adrienthebo: Namespacing a state with a name like `:foo` converts the name from `Symbol` to `Array<Symbol>`. Is this too clever?

adrienthebo:
Per @branan's request, this is an example of how one would use this state machine logic.

```ruby
BOOTSTRAP_CA = Puppet::StateMachine.build("bootstrap ca") do |m|
  m.start_state(:start)

  # TODO
  #   * Lock and unlock SSL directory
  #   * Enable HTTP verification after CA cert has been downloaded
  #

  m.state :start,
    action: -> { Puppet::SSL::Host.ca_location = :remote },
    event: ->(_) { :read_ca_cert },
    transitions: {read_ca_cert: :read_ca_cert}

  m.state :read_ca_cert,
    action: -> { File.readable?(Puppet[:localcacert]) },
    event: ->(result) { result ? :ca_cert_present : :ca_cert_absent },
    transitions: {
      ca_cert_present: :read_ca_crl,
      ca_cert_absent: :fetch_ca_cert
    }

  m.state :fetch_ca_cert,
    action: -> do
      http = Puppet::Network::HttpPool.http_instance(Puppet[:ca_server], Puppet[:ca_port], true, false)
      resp = http.get('/puppet-ca/v1/certificate/ca?environment=production')
      if resp.code == '200'
        File.open(Puppet[:localcacert], 'w:ASCII') { |fh| fh.write(resp.body) }
      end
    end,
    event: ->(result) { result ? :ca_cert_fetched : :ca_cert_absent },
    transitions: {
      ca_cert_fetched: :read_ca_crl,
      ca_cert_absent: :ca_cert_unretrievable
    }

  m.state :read_ca_crl,
    action: -> { File.readable?(Puppet[:hostcrl]) },
    event: ->(result) { result ? :ca_crl_present : :ca_crl_absent },
    transitions: {
      ca_crl_present: :complete,
      ca_crl_absent: :fetch_ca_crl,
    }

  m.state :fetch_ca_crl,
    action: -> do
      http = Puppet::Network::HttpPool.http_instance(Puppet[:ca_server], Puppet[:ca_port], true, false)
      resp = http.get('/puppet-ca/v1/certificate_revocation_list/ca?environment=production')
      if resp.code == '200'
        File.open(Puppet[:hostcrl], 'w:ASCII') { |fh| fh.write(resp.body) }
      end
    end,
    event: ->(result) { result ? :ca_crl_fetched : :ca_crl_absent },
    transitions: {
      ca_crl_fetched: :complete,
      ca_crl_absent: :ca_crl_unretrievable
    }

  m.state :complete, type: :final
  m.state :error, type: :error
  m.state :ca_cert_unretrievable, type: :error
  m.state :ca_crl_unretrievable, type: :error
end
```

I can include another example for composing state machines later.

lib/puppet/state_machine.rb
branan: This seems to be what `Puppet::DevError` is for, right?

branan: In C++ I'd use a `logic_error` - the developer made a mistake here, and there is no environmental change that could make this work.

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,87 @@
...
+      result = :running
+      current = start_state
+      loop do
+        result = current.action
+
+        if current.error?
+          return :errored
+        elsif current.final?
+          return :complete
+        end
```
branan: Instead of returning any values here, should we just return quietly on success and raise an exception on an error state? Can the calling code provide better context for the error than the machine can?

**random thought here** Perhaps error states could/should encode more information (error messages and exception classes) that the machine can then `raise` if its hits that state?

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,240 @@
...
+    # The MachineBuilder class provides a DSL for implementing the states and transitions of a state machine,
+    # and validating that the machine is well behaved.
+    #
+    # ## Composing state machines
+    #
+    # Multiple state machines can be composed into a single unified state machine with the {#compose} method.
+    # When state machines are composed the state names and transitions are namespaced based on the originating
+    # state machine.
+    #
+    # ### Final states
```
branan: Should most of this doc be in `state_machine.rb` since `StateMachine::build` is the top-level way to build a machine with this class as an implementation detail?

lib/puppet/state_machine/state.rb
branan: The goal is that a state `:foo` in namespace `:bar` will end up being named `[:bar, :foo]`, right? That probably needs good docs (probably around the compose method), but is sane in my mind.

lib/puppet/state_machine/state.rb
```diff
@@ -0,0 +1,111 @@
...
+        new_transitions = Hash[@transitions.map do |(event, target)|
+          [event, Array(target).dup.unshift(ns)]
+        end]
+
+        new_type = @type
+        if new_type == :final
+          new_event_cb = ->(_) { :'*final*' }
```
branan: are `'` and `*` valid in symbols, or is this syntax doing something weird?

lib/puppet/state_machine/state.rb
```diff
@@ -0,0 +1,111 @@
...
+      # In order to compose multiple state machines into a single machine, the states of an
+      # individual machine need to be namespaced to prevent name collisions across different
+      # machines and identify which states originated from which of the individual machines.
+      #
+      # @param ns [Symbol]
+      # @return [Puppet::StateMachine::State]
+      def namespace(ns)
+        new_name = Array(@name).dup.unshift(ns)
+        new_transitions = Hash[@transitions.map do |(event, target)|
+          [event, Array(target).dup.unshift(ns)]
```
branan: This is doing the same munging operation as above. Maybe break this out into a `namespace_state_name` or similar method so it's a little  more clear what's going on?

lib/puppet/state_machine/state.rb
adrienthebo: The quotes are normal quotes; this creates a symbol with a string value of `*final*` and the quotes allow the use of the asterisks. In effect this makes a symbol that's more clearly marked as special, since we're using it for a special event transition.

Is this too clever?

branan: I'm generally pretty bad at Ruby, so I have no idea if this is a "well known" syntax. I do like the idea of trying to mark the even in a special way so we can't conflict with users' events. ¯\\\_(ツ)_/¯

branan: Could we perhaps hide this behind a constant, so we don't have to expose users to this funky event name in their composed machines, but can still namespace it? Then we can be as crazy as we want.

`FINAL_EVENT=:'***FINAL***MACHINE***EVENT***'` (maybe not actually this)

lib/puppet/state_machine.rb
adrienthebo: I've been planning on adding some sort of description to some states, especially error states, for this. How about an `errmsg` field whose value is the exception string thrown when the state is entered?

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,240 @@
...
+      # @note The source state must be defined before defining transitions for that state.
+      #
+      # @example
+      #   machine.transition(
+      #     source: :create_lockfile,
+      #     target: :run_agent,
+      #     on: :got_lock)
+      #
+      # @return [Hash<Event, Target>] The defined transitions for the named source state
+      def transition(opts)
```
hlindberg: Can I have a guard (logic preventing/allowing) the transition?

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,240 @@
...
+      #     m.start_state([:m1, :start])
+      #
+      #     m.transition(
+      #       source: [:m1, :final],
+      #       target: [:m2, :start],
+      #       on: :'*final*')
+      #
+      #     m.final([:m2, :final])
+      #   end
+      def compose(machines, &blk)
```
adrienthebo: The states are copied from the original state machine, namespaced, and then placed in the new state machine. For the case where m1 has states on and off, and m2 has multiple states for that on state, you would model this with the following.

```ruby
on_machine = Puppet::StateMachine.build("On states") do |m|
  m.start_state(:initializing)

  m.state(:initializing,
    action: -> { Puppet.notice("initializing") },
    event: ->(result) { :initialized },
    transitions: {initialized: :on}) # Assuming that initialization can't fail

  m.state(:operative,
    action: -> { Puppet.notice("operating") },
    event: ->(result) { result == :halt ? :shutdown : :run },
    transitions: {
      halt: :shutting_down,
      run: :operative # Self referential transition
    })

  m.state(:shutting_down,
    action: -> { Puppet.notice("Shutting down") },
    event: ->(result) { :halted },
    transitions: {halted: :final})

  m.state(:final,
    type: :final,
    action: -> { Puppet.notice("Halted") })
end

composed = Puppet::StateMachine.build(
  "Composed machine",
  {on: on_machine)
) do |m|
  m.start_state(:start)

  m.state(:start,
          action: -> { Puppet.notice("starting composed machine") },
          event: ->(_) { :turn_on },
          transitions: {turn_on: [:on, :initializing]})

  m.state(:final
          type: :final
          action: -> { Puppet.notice("finishing composed machine") })

  m.transition(
    source: [:on, :final],
    target: :final,
    on: Puppet::StateMachine::FINAL)
end

```

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,87 @@
...
+      @states[state_name]
+    end
+
+    # Generate a set of state machines with the given namespace added to the state name
+    # and transitions.
+    #
+    # @param ns [Symbol]
+    # @return [Hash<(Symbol, Array<Symbol>), Puppet::StateMachine::State>]
+    def namespaced_states(ns)
+      @states.values.inject({}) do |hash, state|
```
adrienthebo: I didn't realize we changed that convention, this has been fixed.

lib/puppet/state_machine.rb
adrienthebo: @hlindberg my original approach has been to model this somewhat like a DFA that takes some input and runs till completion or error. In this case we don't have an input string and the transitions are determined by the current state, but the model was to run a state machine to completion. One option for this is to extract the running logic to a separate class again, so if we want to model different types of behavior or have an async machine we can provide a different runner class that implements the desired behavior. What do you think of this approach?

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,240 @@
...
+          opts[:transitions].each do |event, target|
+            trans[event] = target
+          end
+        end
+        @states[state_name] = State.new(state_name, opts[:action], opts[:event], opts[:type], trans)
+      end
+
+      # Add a transition between states.
+      #
+      # @note The source state must be defined before defining transitions for that state.
```
adrienthebo: I thought I responded to this comment but it seems to have vanished. Anywho, transitions are defined as `{event => :state_name}`, so there's no reason this wouldn't work. That said I haven't had the need for self referencing transitions so this isn't under test. When this behavior is needed we can explore it further.

lib/puppet/state_machine/machine_builder.rb
adrienthebo: Per the above comment I'm trying to keep the first revision somewhat minimal; this functionality can be added but I'm deferring on that work.

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,197 @@
...
+    #       transitions: {
+    #         got_lock: :run_agent,
+    #         already_locked: :cancel_run,
+    #       })
+    #
+    #     m.state(:cancel_run,
+    #       type: :error,
+    #       action: -> { Puppet.error("Cancelling run.") }
+    #
+    #     m.state(:run_puppet,
```
MosesMendoza: I think this state should be `:run_agent` given line 26? Really quite minor but I'm a stickler for doc strings being accurate.

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,197 @@
...
+    #     # Transition from m2 final state to composed machine final state
+    #     m.transition(
+    #       source: [:m2, :m2_final],
+    #       target: :composed_final,
+    #       on: Puppet::StateMachine::FINAL)
+    #
+    #   end
+    #
+    def self.compose(name, machines, &block)
+      MachineBuilder.new(name).compose(machines, &block)
```
MosesMendoza: Perhaps a stylistic thing but maybe this would be more maintainable with fully qualified namespaces, ie `Puppet::StateMachine::MachineBuilder.new` ? maybe that's too verbose

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,197 @@
...
+    end
+
+    # @!attribute [r] name
+    #   @return [String] The description of this state machine
+    attr_reader :name
+
+    # @!attribute [r] states
+    #   @api private
+    attr_reader :states
```
MosesMendoza: Did you intentionally omit the `@return` tag here?

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,182 @@
...
+      #     action: -> { Puppet::Util::Lockfile.new("/var/lock/puppet.lock").lock },
+      #     event: ->(has_lock) { has_lock ? :got_lock : :already_locked },
+      #     transitions: {
+      #       got_lock: :run_agent,
+      #       already_locked: :cancel_run,
+      #     })
+      #
+      # @return [Puppet::StateMachine::State] The generated machine state
+      def state(state_name, opts = {})
+        unhandled = opts.keys - [:action, :event, :type, :transitions]
```
MosesMendoza: It seems odd to me that the various validations in this method (particularly knowledge of the supported state options) exist in the MachineBuilder class - why aren't they part of the State class?

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,182 @@
...
+      #     source: :create_lockfile,
+      #     target: :run_agent,
+      #     on: :got_lock)
+      #
+      # @return [Hash<Event, Target>] The defined transitions for the named source state
+      def transition(opts)
+        source = opts[:source]
+        target = opts[:target]
+        event = opts[:on]
+        source_state = @states[opts[:source]]
```
MosesMendoza: does the target state also have to exist in `@states` ?

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,182 @@
...
+            if @states[target].nil?
+              raise ArgumentError, "Unable to build state machine '#{@machine_name}': invalid transition from from state '#{state.name}' to missing state '#{target}'"
+            end
+          end
+        end
+      end
+
+      def verify_exiting_transitions!
+        @states.values.each do |state|
+          if !(state.error? || state.final?) && state.transitions.empty?
```
MosesMendoza: I'm guessing this is overkill but it seems like the first half of this, ie `!(state.error? || state.final?)` is really asking `is this is a terminal state?` which might be logic better encapsulated in a method on the State class, ie `#is_terminal_state?`

lib/puppet/state_machine.rb
MosesMendoza: Same for line 60 above

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,197 @@
...
+    #       action: -> { Puppet.notice("Starting machine 1") },
+    #       event: ->(result) { :started },
+    #       transitions: {started: :m1_final})
+    #
+    #     m.state(:m1_final,
+    #       action: -> { Puppet.notice("Finishing machine 1") },
+    #       type: :final)
+    #   end
+    #
+    #   machine2 = Puppet::StateMachine.build do |m|
```
MosesMendoza: This example is missing the machine name - could you add?

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,197 @@
...
+    def start_state
+      @states[@start_state]
+    end
+
+    def state(state_name)
+      @states[state_name]
+    end
+
+    # Generate a set of state machines with the given namespace added to the state name
+    # and transitions.
```
MosesMendoza: Is this comment right? It seems like it generates a set of machine states, rather than state machines.

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,182 @@
...
+      def build(&blk)
+        yield self
+        verify!
+        Puppet::StateMachine.new(@machine_name, @states, @start_state)
+      end
+
+      # Compose multiple state machines into a a single state machine.
+      #
+      # @example
+      #   machine.compose({machine1: m1, machine2: m2}) do |m|
```
MosesMendoza: I think the keys/values are reversed here

MosesMendoza:
A general comment - I personally find the explicit required namespacing of machines during a composition a little unintuitive. The namespaces are a new set of metadata I have to keep track of that are otherwise not relevant to the ultimate outcome, being internal machinery of the composition.

`Puppet::StateMachine::Machine#name` is a required attribute for creating a state machine (though admittedly not required to be globally unique, we could do collision detection in composition) - would it be simpler to use the #name as an implicit namespace, or would that just muddy the waters?

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,197 @@
...
+    #
+    # @param ns [Symbol]
+    # @return [Hash<(Symbol, Array<Symbol>), Puppet::StateMachine::State>]
+    def namespaced_states(ns)
```
MosesMendoza: what would you think about renaming this method to `namespace_states`, ie transforming it into a verb rather than a noun?

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,182 @@
...
+        end
+        build(&blk)
+      end
+
+      # Verify that the machine is well defined.
+      # @return [void]
+      def verify!
+        verify_start_state!
+        verify_transition_targets!
+        verify_exiting_transitions!
```
MosesMendoza: curious why these are suffixed with `!` - is it because they raise on failure of validation?

lib/puppet/state_machine.rb
```diff
@@ -0,0 +1,197 @@
...
+    #     m.transition(
+    #       source: [:m1, :m1_final],
+    #       target: [:m2, :m2_start],
+    #       on: Puppet::StateMachine::FINAL)
+    #
+    #     # Transition from m2 final state to composed machine final state
+    #     m.transition(
+    #       source: [:m2, :m2_final],
+    #       target: :composed_final,
+    #       on: Puppet::StateMachine::FINAL)
```
MosesMendoza: Is this supposed to reference `Puppet::StateMachine::FINAL_EVENT`? (same on 136)

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,182 @@
...
+      #     m.transition(
+      #       source: [:m1, :final],
+      #       target: [:m2, :start],
+      #       on: Puppet::StateMachine::FINAL_EVENT)
+      #
+      #     m.final([:m2, :final])
+      #   end
+      def compose(machines, &blk)
+        machines.each_pair do |ns, machine|
+          @states.merge!(machine.namespaced_states(ns))
```
MosesMendoza: Aside from my more meta comment re:namespaces, I think we should do namespace collision detection here or in StateMachine#compose, or if I'm not mistaken this will fail in interesting and inexplicable ways.

lib/puppet/state_machine/machine_builder.rb
```diff
@@ -0,0 +1,182 @@
...
+      #     m.start_state([:m1, :start])
+      #
+      #     m.transition(
+      #       source: [:m1, :final],
+      #       target: [:m2, :start],
+      #       on: Puppet::StateMachine::FINAL_EVENT)
+      #
+      #     m.final([:m2, :final])
+      #   end
+      def compose(machines, &blk)
```
MosesMendoza: Out of curiosity are there situations where MachineBuilder#compose would be used in any way other than as a Puppet::StateMachine constructor?

lib/puppet/state_machine/state.rb
MosesMendoza: It doesn't seem quite right to me to have two different data types as canonical forms for referencing a state (symbol and array). Maybe its unrealistic, overly complicated or actually just worse but are there other options we can consider for unifying the representation of namespaced and non-namespaced states?

adrienthebo:
@MosesMendoza @branan since I changed the diff that the namespacing conversation was associated with I can't comment there, so I'll continue the conversation here.

So yes, there is some weirdness with namespacing. Right now a simple name is just `Symbol`; namspacing that yields `Array<Symbol>`; namespacing that yields another `Array<Symbol>`. This makes simple state machines simple, is consistent when all states are namespaced, but gets odd otherwise.

One option is to always use arrays for state names. This makes things more consistent, with and without namespaces, but declaring states and transitions becomes more cumbersome.
```ruby
m.state([:run_agent], # namespaces to [:ns, :run_agent]
  transitions: {
    event1: [:transition1], # namespaces to [:ns, :transition1]
    event2: [:transition2], # namespaces to [:ns, :transition2]
  }
)
```

The other option is to use strings, and use double colons or some other separator for namespacing.

```ruby
m.state("run_agent", # namespaces to "ns::run_agent"
  transitions: {
    event1: "transition1", # namespaces to "ns::transition1"
    event2: "transition2", # namespaces to "ns::transition2"
  }
)
```

This unifies the object types in use, but I don't know if this is any clearer.

Anywho. I'm happy to change this, I don't have a lot of opinions. Namespacing is probably the least interesting/valuable part of the work I'm doing in terms of SSL, so I'm happy to implement whatever is clearest so that I can get to the really exciting things involving large prime numbers and elliptic curves.

MosesMendoza:
Can you see any future benefit to splitting `transition` into its own class (ie such that a Puppet::StateMachine::State instance may have/reference one or more Puppet::StateMachine::Transition instances) ?

MosesMendoza:
@adrienthebo this is pretty cool - nice work on this PR. I see us being able to leverage this quite successfully.

adrienthebo:
There are some cases where we could extract a Transition class, such as Henrik's interest in being able to add guards to transitions. However I don't foresee an immediate need for this, and if we need that functionality we should be able to add it later as the implementation of transitions (thus far) is internal only.

lib/puppet/state_machine/machine_builder.rb
adrienthebo: It looks like GitHub ate my comment on this, so there might be duplication here.

The MachineBuilder class is just a builder pattern, and `#compose` and `#build` methods are the two entry points. What else would you expect to use this class and these methods for aside from generating state machines?

lib/puppet/state_machine/machine_builder.rb
adrienthebo: Again, comment might have been eaten. Yes, these are suffixed as such since those methods exist to either silently return or throw an exception. I'm happy to remove them if this isn't consistent with the rest of the codebase.

lib/puppet/state_machine.rb
MosesMendoza: I also didn't know we had made that change.

lib/puppet/state_machine/machine_builder.rb
MosesMendoza: Doesn't bother me at all - I like highlighting that they're different. Just curious.

lib/puppet/state_machine/machine_builder.rb
MosesMendoza: I don't think I would use them for anything else.. I think where my mind was going with that original comment was trying to figure out if we expected people to be interacting with the `MachineBuilder` class directly or only indirectly via the `StateMachine` class, and if there was any way we would add extra docs/comments suggesting as much.

adrienthebo: I expect users to use the `#state` and `#transition` methods per the documentation, but I don't expect users to instantiate the builder class itself. I've marked the class private to emphasize this point.

lib/puppet/state_machine/machine_builder.rb
adrienthebo: My reply to this apparently got eaten because I didn't submit the review properly. In any case, I would use named arguments if we could use them but since we need to support Ruby 1.9 syntax for JRuby we can't. The State class uses positional parameters, which is sensible for that class but less understandable for this DSL. The builder class glues together the named argument via hashes, and then translates that into the more compact positional arguments.

As for why the validation occurs here as opposed to within the state, the individual classes assume that they're being used correctly and defer all validation to this machine. The MachineBuilder class builds up a state machine incrementally instead of requiring all transitions and states to be defined in a single shot, and validates everything before the machine is generated. In short this just implements the builder pattern.

In addition, a single state only contains the context associated with that single state and is not aware of the entire machine. In order to do this validation within the state class we'd need to add a connection to the machine. In addition the state machine class is supposed to be immutable; pushing this logic into the State so that it can peer into the state machine would require that the state machine would have be modifiable after creation. Instead, the MachineBuilder class is the mutable object that incrementally builds up things, validates them, and then generates an immutable, verified state machine.

Another benefit of extracting all of the validation logic here is that it allows us to turn it off if desired. Once a state machine has been implemented we don't necessarily need to validate it every time; these validations only really need to occur while the implementation of the states are in flux.

lib/puppet/state_machine/machine_runner.rb
```diff
@@ -0,0 +1,46 @@
...
+  class StateMachine
+    # Run a state machine until it enters a final or error state.
+    #
+    # This runner follows the behavior of a DFA in that the machine is run until completion or
+    # error. There are two notable differences from this runner and a pure DFA. First, there
+    # is no input string, and instead the state event callbacks determine the next transition.
+    # Second, instead of terminating the state machine successfully if the machine is in a final
+    # state when all input is consumed, entering a final state will immediately terminate the
+    # machine.
+    class MachineRunner
```
adrienthebo: I'm really enjoying how many comments GitHub is making me rewrite because of how intuitive and well behaved the review functionality works. In any case my immediate need is for a state machine that behaves like a DFA in that it runs to completion, but that's not a required property of a state machine. State machines could be used to model the [Solaris zone state machine](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/zone.rb#L10). Outside operations transition the state machine between states, and the machine itself doesn't run till completion - in fact it doesn't run at all. This machine running logic is only needed for a certain class of machines and the state machine implementation as it stands doesn't have to be constrained to that single use case.

lib/puppet/state_machine/machine_builder.rb
MosesMendoza: 👍

lib/puppet/state_machine/machine_runner.rb
MosesMendoza: ahh that makes sense. Thanks for explaining

MosesMendoza:
@adrienthebo thought more about the namespacing - I agree requiring arrays would be laborious and pendantic. And a quick glance suggests there's nothing that would prohibit a user from using either a symbol OR an array to name a "simple" non-namespaced state if one so desires? I'm 👍  on implementation as-is.

MosesMendoza:
will merge pending CI going green again

adrienthebo:
@branan @MosesMendoza are we still blocked on CI for this? It looks like integration tests are passing but internationalization tests are failing; is the latter a blocker for merging this?

MosesMendoza:
Actually waiting to get a green run on the puppet-agent/master puppet tests.

branan:
puppet/master looks green. I'm merging this

PR #5622, opened 2017-02-07
(PUP-7186) Switch AIX nim test to a simpler test fixture

branan:
The AIX nim provider tests previously relied on a real package. This package did not install cleanly on AIX 7.2. This updates the nim test to use a new test fixture created internally, which allows the test to function on all AIX versions.

### DIFF
--- a/acceptance/tests/aix/nim_package_provider.rb
+++ b/acceptance/tests/aix/nim_package_provider.rb
@@ -6,7 +6,7 @@ confine :to, :platform => "aix"
 
 teardown do
     test_apply('cdrecord', 'absent', '')
-    test_apply('bos.atm.atmle', 'absent', '')
+    test_apply('puppet.test.rte', 'absent', '')
 end
 
 def assert_package_version(package, expected_version)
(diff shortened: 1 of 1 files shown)


MosesMendoza:
@branan would you mind linking output of a passing run of this updated test on AIX 7.2?

branan:
@moses The updated test run against all 4 AIX versions https://gist.github.com/branan/db489f0fa3b159a4c8eec3f8078325e6

acceptance/tests/aix/nim_package_provider.rb
```diff
@@ -6,7 +6,7 @@
 
 teardown do
     test_apply('cdrecord', 'absent', '')
-    test_apply('bos.atm.atmle', 'absent', '')
+    test_apply('puppet.test.rte', 'absent', '')
```
MosesMendoza: Would you mind noting somewhere in the test that this uses a fixture that is for all intents and purposes hidden and locked away somewhere on our nim master?

acceptance/tests/aix/nim_package_provider.rb
```diff
@@ -48,21 +48,24 @@ def test_apply(package_name, ensure_value, expected_version)
   end
 end
 
+# These two packages live in an LPP source on the NIM master. Details
```
branan: @MosesMendoza Comment added here

PR #5625, opened 2017-02-08
(PUP-7188) Acceptance: Add AIX 7.2 static node config

geoffnichols:
Because AIX is not a vmpooler platform, add a static node config for AIX
7.2 acceptance testing.


geoffnichols:
This may not be needed after all, closing.

PR #5627, opened 2017-02-08
(PUP-7190) confine utf-8 test against aix and huawei

MosesMendoza:
Our acceptance hosts are not configured with UTF-8 locale files, resulting in a
failure of these tests. Eventually we may reconfigure the hosts to support this
but for now confine this test to resume passing.


seangriff:
LGTM

LuvCurves:
@MosesMendoza @seangriff Why are we using the confine block above this?

MosesMendoza:
@LuvCurves i felt inclined to specifically highlight these two separately - other than that no reason

PR #5630, opened 2017-02-08
(PUP-7159) Force destroy zfs / zpool

Iristyle:
- There have been intermittent test failures resulting from the
   inability (under certain circumstances) to properly tear down
   resources on Solaris 11 hosts.  Logs will typically show something
   like:

   rm -rf /ztstpool ||:
         rm: Unable to remove directory /ztstpool/mnt: Device busy
         rm: Unable to remove directory /ztstpool: Directory not empty

 - According to Oracle docs for ZFS file system:
   https://docs.oracle.com/cd/E19253-01/819-5461/gammq/index.html

   And Oracle docs for ZFS storage pools:
   https://docs.oracle.com/cd/E19253-01/819-5461/gammr/index.html

   Both the zfs and zpool commands have -f switches available to force
   removal when a device is busy.

### DIFF
--- a/acceptance/lib/puppet/acceptance/solaris_util.rb
+++ b/acceptance/lib/puppet/acceptance/solaris_util.rb
@@ -20,7 +20,7 @@ module Puppet
         lst = on(agent, "zfs list").stdout.lines.each do |l|
           case l
           when /rpool.tstzones/
-            on agent,"zfs destroy -r rpool/tstzones"
+            on agent,"zfs destroy -f -r rpool/tstzones"
           end
         end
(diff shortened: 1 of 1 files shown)


Iristyle:
Note that we're not seeing `device busy` when the `zpool` and `zfs` commands are executed (which is where the `-f` is used).  `zfs destroy` is showing `filesystem does not exist` which makes us think this is an environmental issue.

LuvCurves:
@Iristyle This was related to having an LDom image which had an aborted ZFS test run on it.  The image has been corrected. The problem should go away.

PR #5632, opened 2017-02-09
(PUP-7191) consider non-existent services on AIX to be stopped/disabled rather than failing

MosesMendoza:
Prior to this PR, attempting to manage a non-existent service would result in an error even when ensuring the service is stopped, ie:
```
$ puppet resource service foo ensure=stopped
Error: /Service[foo]: Could not evaluate: Cannot get status of foo, error was: Execution of '/usr/bin/lssrc -s foo' returned 1: 0513-085 The foo Subsystem is not on file.
Wrapped exception:
Execution of '/usr/bin/lssrc -s foo' returned 1: 0513-085 The foo Subsystem is not on file.
Error: Could not run: Cannot get status of foo, error was: Execution of '/usr/bin/lssrc -s foo' returned 1: 0513-085 The foo Subsystem is not on file.
Wrapped exception:
Execution of '/usr/bin/lssrc -s foo' returned 1: 0513-085 The foo Subsystem is not on file.
$ echo $?
1
```

This is in contrast to current and long-standing behavior on RHEL and Ubuntu
systems, where doing something like ensure => stopped on a non-existent service
is a no-op, and from the ral's perspective a non-existent service is considered
to be in the 'stopped' state. That said, this behavior is not necessarily
"standard" though it probably should be. For example, the launchd provider both
prints an error *and* considers the service stopped.

This commit updates the AIX behavior so that a non-existent service is
considered stopped rather than failing with an error and/or failing the puppet
run if ensuring the service stopped. Ral attempts to ensure=running on a
non-existent service still provide an error, but return correct puppet now
instead of just erroring.

With this PR, output when stopping a non-existent service is effectively a no-op:
```
$ puppet resource service foo ensure=stopped
service { 'foo':
  ensure => 'stopped',
}
$ echo $?
0
$ puppet apply -e " service { 'foo' : ensure => stopped } "
Notice: Compiled catalog for pe-aix-71-agent.delivery.puppetlabs.net in environment production in 1.12 seconds
Notice: Applied catalog in 0.19 seconds
$ echo $?
0
```

And attempting to start a non-existent service errors, but exits according to puppet's detailed exit codes policy:
```
$ puppet apply -e " service { 'foo' : ensure => running } "
Notice: Compiled catalog for pe-aix-71-agent.delivery.puppetlabs.net in environment production in 0.57 seconds
Error: Could not start Service[foo]: Execution of '/usr/bin/startsrc -s foo' returned 1: 0513-085 The foo Subsystem is not on file.
Error: /Stage[main]/Main/Service[foo]/ensure: change from stopped to running failed: Could not start Service[foo]: Execution of '/usr/bin/startsrc -s foo' returned 1: 0513-085 The foo Subsystem is not on file.
Notice: Applied catalog in 0.17 seconds
$ echo $?
0
$ puppet apply -e " service { 'foo' : ensure => running } " --detailed-exitcodes
Notice: Compiled catalog for pe-aix-71-agent.delivery.puppetlabs.net in environment production in 0.58 seconds
Error: Could not start Service[foo]: Execution of '/usr/bin/startsrc -s foo' returned 1: 0513-085 The foo Subsystem is not on file.
Error: /Stage[main]/Main/Service[foo]/ensure: change from stopped to running failed: Could not start Service[foo]: Execution of '/usr/bin/startsrc -s foo' returned 1: 0513-085 The foo Subsystem is not on file.
Notice: Applied catalog in 0.17 seconds
$ echo $?
4
```

There's a whitespace commit in this PR that makes this change appear large than it is. Diff suppressing whitespace change: https://github.com/puppetlabs/puppet/pull/5632/files?w=1

### DIFF
--- a/lib/puppet/provider/service/src.rb
+++ b/lib/puppet/provider/service/src.rb
@@ -138,7 +138,6 @@ Puppet::Type.type(:service).provide :src, :parent => :base do
       Puppet.debug("Service #{@resource[:name]} is #{args[-1]}")
       return state
     end
-    self.fail("No such service found")
   rescue Puppet::ExecutionFailure => detail
     self.debug(detail.message)
     return :stopped
--- a/lib/puppet/provider/service/src.rb
+++ b/lib/puppet/provider/service/src.rb
@@ -91,56 +91,56 @@ Puppet::Type.type(:service).provide :src, :parent => :base do
   end
 
   def restart
-      execute([command(:lssrc), "-Ss", @resource[:name]]).each_line do |line|
-        args = line.split(":")
+    execute([command(:lssrc), "-Ss", @resource[:name]]).each_line do |line|
+      args = line.split(":")
 
-        next unless args[0] == @resource[:name]
+      next unless args[0] == @resource[:name]
 
-        # Subsystems with the -K flag can get refreshed (HUPed)
-        # While subsystems with -S (signals) must be stopped/started
-        method = args[11]
-        do_refresh = case method
-          when "-K" then :true
-          when "-S" then :false
-          else self.fail("Unknown service communication method #{method}")
-        end
-
-        begin
-          if do_refresh == :true
-            execute([command(:refresh), "-s", @resource[:name]])
-          else
-            self.stop
-            self.start
-          end
-          return :true
-        rescue Puppet::ExecutionFailure => detail
-          raise Puppet::Error.new("Unable to restart service #{@resource[:name]}, error was: #{detail}", detail )
-        end
+      # Subsystems with the -K flag can get refreshed (HUPed)
+      # While subsystems with -S (signals) must be stopped/started
+      method = args[11]
+      do_refresh = case method
+        when "-K" then :true
+        when "-S" then :false
+        else self.fail("Unknown service communication method #{method}")
       end
-      self.fail("No such service found")
+
+      begin
+        if do_refresh == :true
+          execute([command(:refresh), "-s", @resource[:name]])
+        else
+          self.stop
+          self.start
+        end
+        return :true
+      rescue Puppet::ExecutionFailure => detail
+        raise Puppet::Error.new("Unable to restart service #{@resource[:name]}, error was: #{detail}", detail )
+      end
+    end
+    self.fail("No such service found")
   rescue Puppet::ExecutionFailure => detail
-      raise Puppet::Error.new("Cannot get status of #{@resource[:name]}, error was: #{detail}", detail )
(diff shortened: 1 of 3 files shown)


MosesMendoza:
@LuvCurves @McdonaldSeanp @branan  - I started an ad-hoc agent run with these changes on 'classic' jenkins so that I can get puppet-agent packages with these changes built and validate the new acceptance test before we merge this. That build is here: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_init-van-int_suite-manual-ad-hoc/

I know there's been some changes around how we build for AIX recently, so let me know if this is wrong. Just to make sure its clear - the goal is to be able to do a local beaker run against AIX SUTs against this change set. Not sure how I do that once I do have packages, but once step at a time. Any input is much appreciated.

MosesMendoza:
Marked as blocked pending a successful ad-hoc run of the acceptance test in this PR against AIX SUTs

MosesMendoza:
The updated acceptance test has passed a local run against AIX 7.1: https://gist.github.com/MosesMendoza/99c411d8b57e45be7ea101ced0536d99

MosesMendoza:
Removing blocked label - this is ready for review

Magisus:
This seems reasonable, strikes a good compromise, still reporting status without being obnoxious. It's nice to see effort being made to improve our consistency across platforms. Also, +1 whitespace cleanup!

Your branch name includes the word "stable" but this is targeted at master, so just want to verify: master is where you want this to land?

lib/puppet/provider/service/src.rb
```diff
@@ -91,56 +91,56 @@ def stop
...
+      # skip this so split doesn't work right
+      state = case args[-1]
+        when "active" then :running
+        when "inoperative" then :stopped
       end
-      self.fail("No such service found")
+      Puppet.debug("Service #{@resource[:name]} is #{args[-1]}")
+      return state
+    end
   rescue Puppet::ExecutionFailure => detail
```
branan: This catches all execution failures - are there other situations where lssrc could exit non-0 and cause an execution failure?

MosesMendoza: Interesting question. `lsssrc` should always be there, and it can be run by an un-privileged user. As far as I can tell, it will always exit 0 if a given subsystem exists regardless of its state (hence we parse the output string) and 1 if the subsystem does not exist.

MosesMendoza:
Ping @branan any other changes on this or can we push the button?

PR #5633, opened 2017-02-10
(pup-7021) establish central string encoding conversion

MosesMendoza:
This commit creates puppet/util/character_encoding, a single location to house
logic related to the management character encoding conversion logic in puppet.
Prior to this commit, conversion logic was just beginning to fan out and
duplicate throughout puppet - with all the associated risk and maintenance
burden. We will leverage this code in a separate effort adding a
character-encoding-aware wrapper for ruby's Etc module and rewiring our
providers to use that instead.

This commit actually only adds a single public method -
'Puppet::Util::CharacterEncoding::convert_to_utf8'.

This method applies a standard heuristic to attempt to convert a (copy of a)
passed in string to UTF-8. That heuristic is summarized and justified as
follows (much of this is left as code-comments in the commit as well):

Given a supplied string:
- If the string has UTF-8 encoding, return it as-is. We have no need to do
  anything.

otherwise, if the string is not UTF-8 already:
- Check if its bytes represent valid UTF-8 as currently constitued (in its
  non-UTF-8 encoding) and if so, limit ourselves to setting the external
  encoding of the string to UTF-8 rather than actually transcoding it. We do
  this to handle a couple scenarios:

The first scenario is that the string was originally valid UTF-8 but the
current puppet run is not in a UTF-8 environment. In this case, the string will
likely have invalid byte sequences (i.e., string.valid_encoding? == false), and
attempting to transcode will fail with Encoding::InvalidByteSequenceError,
referencing the invalid byte sequence in the original, pre-transcode, string.
We might have gotten here, for example, if puppet is run first in a user
context with UTF-8 encoding (setting the "is" value to UTF-8) and then later
run via cron without UTF-8 specified, resulting in running in POSIX or some
other default encoding on many systems. In this scenario we're effectively
best-guessing this string originated as UTF-8 and only set external encoding to
UTF-8 - transcoding would have failed anyway.

The second scenario (more rare/unlikely, I expect) is that this string does NOT
have invalid byte sequences (string.valid_encoding? == true), but would
transcode cleanly into different, also valid unicode.

Our example case is "\u16A0" - "RUNIC LETTER FEHU FEOH FE"
http://www.fileformat.info/info/unicode/char/16A0/index.htm
0xE1 0x9A 0xA0 / 225 154 160

These bytes are valid in ISO-8859-1 but the character they represent transcodes
(incorrectly, but cleanly) in ruby to *different* characters in UTF-8. That's
not what we want if the user intended the original string as UTF-8.  Since
we're effectively moving towards a universal UTF-8 standard, if the string is
valid UTF-8 as currently constituted, we default to assuming the string
originated in UTF-8 and do not transcode it - we only set external encoding.

Otherwise, if the string is neither currently considered UTF-8 nor would be
valid UTF-8 bytes:
- Check if the string is even transcodable, as best we can. We know at least
  one way to determine it is _not_ transcodable, which is if
  string.valid_encoding == false, which will always result in
  InvalidByteSequenceError. We may augment this check in the future if we have
  better ways. If it is transcodable, go ahead and attempt to transcode (byte
  changing conversion) to UTF-8.

If none of our conditions have been met thus far:
- the string has been confirmed to not be UTF-8, nor would be valid UTF-8 with
  its current bytes, nor can be transcoded to UTF-8. In this scenario, we fail,
  as this string cannot be converted to UTF-8.

### DIFF
--- a/lib/puppet/util/character_encoding.rb
+++ b/lib/puppet/util/character_encoding.rb
@@ -1,0 +1,96 @@
+# A module to centralize heuristics/practices for managing character encoding in Puppet
+require 'puppet/util'
+
+module Puppet::Util::CharacterEncoding
+  class << self
+    #TODO make encoding warnings a category/silenceable#.
+
+    # @api public
+    # @param [String] string a string to transcode / force_encode to utf-8
+    # @return [String] string if already utf-8, OR
+    #   a copy of string with external encoding set to utf-8 if bytes are valid utf-8 OR
+    #   a copy of string transcoded to utf-8
+    # @raise [Puppet::Error] a failure to legitimately set external encoding or
+    #   transcode string
+    def convert_to_utf_8(string)
+      return string if string.encoding == Encoding::UTF_8
+
+      value_to_encode = string.dup
+
+      begin
+        if valid_utf_8?(value_to_encode)
+          # Before we try to transcode the string, check if it is valid UTF-8 as
+          # currently constitued (in its non-UTF-8 encoding), and if it is, limit
+          # ourselves to setting the external encoding of the string to UTF-8
+          # rather than actually transcoding it. We do this to handle
+          # a couple scenarios:
+
+          # The first scenario is that the string was originally valid UTF-8 but
+          # the current puppet run is not in a UTF-8 environment. In this case,
+          # the string will likely have invalid byte sequences (i.e.,
+          # string.valid_encoding? == false), and attempting to transcode will
+          # fail with Encoding::InvalidByteSequenceError, referencing the
+          # invalid byte sequence in the original, pre-transcode, string. We
+          # might have gotten here, for example, if puppet is run first in a
+          # user context with UTF-8 encoding (setting the "is" value to UTF-8)
+          # and then later run via cron without UTF-8 specified, resulting in in
+          # EN_US (ISO-8859-1) on many systems. In this scenario we're
+          # effectively best-guessing this string originated as UTF-8 and only
+          # set external encoding to UTF-8 - transcoding would have failed
+          # anyway.
+
+          # The second scenario (more rare, I expect) is that this string does
+          # NOT have invalid byte sequences (string.valid_encoding? == true),
+          # but is *ALSO valid unicode*.
+          # Our example case is "\u16A0" - "RUNIC LETTER FEHU FEOH FE"
+          # http://www.fileformat.info/info/unicode/char/16A0/index.htm
+          # 0xE1 0x9A 0xA0 / 225 154 160
+          # These bytes are valid in ISO-8859-1 but the character they represent
+          # transcodes cleanly in ruby to *different* characters in UTF-8.
+          # That's not what we want if the user intended the original string as
+          # UTF-8. We can only guess, so if the string is valid UTF-8 as
+          # currently constituted, we default to assuming the string originated
+          # in UTF-8 and do not transcode it - we only set external encoding.
+          value_to_encode.force_encoding(Encoding::UTF_8)
+        elsif transcodable?(value_to_encode)
+          # If the string is not currently valid UTF-8 but it can be transcoded,
+          # we can guess this string was not originally unicode. We need it to
+          # be now, so transcode it to UTF-8. For strings with original
+          # encodings like SHIFT_JIS, this should be the final result.
+          value_to_encode.encode!(Encoding::UTF_8)
+        else
+          # If the string is neither valid UTF-8 as-is nor is transcodable, fail
+          # at this point. It requires user remediation.
+          raise EncodingError
+        end
+      rescue EncodingError => detail
+        # catch both our own self-determined failure to transcode as well as any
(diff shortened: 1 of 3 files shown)


MosesMendoza:
looks like a couple spec failures on ruby 1.9.3 - investigating

MosesMendoza:
@branan  @Iristyle you had some feedback on the original PR for this, https://github.com/puppetlabs/puppet/pull/5520. That feedback is incorporated in this PR.

lib/puppet/util/character_encoding.rb
```diff
@@ -0,0 +1,96 @@
...
+        else
+          # If the string is neither valid UTF-8 as-is nor is transcodable, fail
+          # at this point. It requires user remediation.
+          raise EncodingError
+        end
+      rescue EncodingError => detail
+        # catch both our own self-determined failure to transcode as well as any
+        # error on ruby's part, ie Encoding::InvalidByteSequenceError or
+        # Encoding::UndefinedConversionError.
+        raise Puppet::Error, "#{detail.inspect}: #{value_to_encode.dump} is not valid UTF-8 and cannot be transcoded by Puppet."
```
adrienthebo: Does this need to be tagged for externalization?

MosesMendoza: Yes! Thanks for the reminder. Will review PRs and ensure wrapping is done where needed.

lib/puppet/util/character_encoding.rb
```diff
@@ -0,0 +1,96 @@
+# A module to centralize heuristics/practices for managing character encoding in Puppet
+require 'puppet/util'
```
Iristyle: Was it necessary to have the circular `require` here?

lib/puppet/util/character_encoding.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    private
+
+    # Do our best to determine if a string is valid
+    # UTF-8 via String#valid_encoding?
+    # @api private
+    # @param [String] string a string to test
+    # @return [Boolean] whether we think the string is UTF-8 or not
+    def valid_utf_8?(string)
+      string.dup.force_encoding(Encoding::UTF_8).valid_encoding?
```
Iristyle: I'm not sure about this implementation, as if it becomes pervasive, this will generate quite a bit of string data on the Ruby heap that isn't really used, which could lead to perf issues.

If we could do an in-place scan against the bytes, that would be better.

* I checked for another implementation of `valid_encoding?` elsewhere in the Ruby code, and didn't see anything very obvious - under the hood `rb_enc_str_coderange` is called.  That function lives at https://github.com/ruby/ruby/blob/7de42daa21a47335a485265f74322533a2da4f1a/string.c#L619-L637
* One hacky way of just using Ruby would be to save the encoding, set encoding to UTF-8, check `valid_encoding?` and then reset back to the original.  Gross.
* UTF-8 bytes sequences are actually pretty easy to detect by scanning strings byte-by-byte, so we could write a quick implementation that does just that in-place on the existing string.  I was going to point you to the Wikipedia page, but I found a quick example of this at http://codereview.stackexchange.com/a/59439


Also, some more related discussion on detecting UTF-8 for reference:
https://stackoverflow.com/questions/8635578/how-to-check-whether-the-character-is-utf-8

Iristyle: Also - https://gist.github.com/brianmario/2992961

Iristyle: Doing a bit more reading, it looks like `dup` may actually share string data under the hood, so I think my previous comments are invalid.  TIL!

lib/puppet/util/character_encoding.rb
MosesMendoza: I suppose not. I think it was only in there as an aside during dev I think. The new `puppet/etc` does `require 'puppet/util/character_encoding'`, and its associated spec fails unless `puppet/util` is required first, but I can instead have `puppet/etc` require `puppet/util` instead of `puppet/util/character_encoding` which would effect the same thing and probably be less circuitous.

MosesMendoza: Addressed in a commit on https://github.com/puppetlabs/puppet/pull/5634

lib/puppet/util/character_encoding.rb
```diff
@@ -0,0 +1,96 @@
...
+        else
+          # If the string is neither valid UTF-8 as-is nor is transcodable, fail
+          # at this point. It requires user remediation.
+          raise EncodingError
+        end
+      rescue EncodingError => detail
+        # catch both our own self-determined failure to transcode as well as any
+        # error on ruby's part, ie Encoding::InvalidByteSequenceError or
+        # Encoding::UndefinedConversionError.
+        raise Puppet::Error, _("#{detail.inspect}: #{value_to_encode.dump} is not valid UTF-8 and cannot be transcoded by Puppet.")
```
Iristyle: Note that this `raise` loses the stack trace from `detail` based on how things are re-raised.  That is addressed in #5772

Iristyle:
Follow on PR is at #5799

PR #5634, opened 2017-02-10
(PUP-7021) Add UTF-8-aware Etc wrapper

MosesMendoza:
This PR includes the content of https://github.com/puppetlabs/puppet/pull/5633, and is separated out for ease of review. This PR content is the content of commit '(pup-7021) Add UTF-8-aware Etc wrapper'.

The Ruby Etc module will return various Struct objects containing strings in
the system's locale, or if they are invalid in the current locale, as
ASCII-8BIT (Binary). This leads to inconsistent and problematic behavior when
Puppet uses the Etc module in its providers, particularly when trying to
concatenate or compare strings which now may have different, and often
incompatible, encodings. The nameservice provider in particular makes heavy use
of Etc, which has resulted in numerous bugs reported regarding UTF-8.

Initial attempts to resolve the issues focused on the locations where string
comparison or concatenation was happening, but it became quickly clear that
there are too many places to do this, and the better solution would likely be
to control the problem where the strings first enter the system - at the Etc
module.

This commit adds a Puppet wrapper around ruby's Etc module that leverages
puppet/util/character_encoding to convert the values returned by Etc to UTF-8.
In order to avoid ever letting the original Etc values get into the system (and
with a side benefit of avoiding extraneous object creation) the Structs are
modified in-place.

It is true that this commit may introduce different failure scenarios, as
unavoidable failure will now likely be located in the string conversion process
(ie if we have truly UTF-8 incompatible strings), rather than higher in the
stack at comparison or concatenation for display, but it seems these were going
to fail anyway.

### DIFF
--- a/lib/puppet/etc.rb
+++ b/lib/puppet/etc.rb
@@ -1,0 +1,148 @@
+require 'puppet/util/character_encoding'
+# Wrapper around Ruby Etc module allowing us to manage encoding in a single
+# place.
+# This represents a subset of Ruby's Etc module, only the methods required by Puppet.
+# Etc returns strings in variable encoding depending on
+# environment. For Puppet we specifically want UTF-8 as our input from the Etc
+# module - which is our input for many resource instance 'is' values. The
+# associated 'should' value will basically always be coming from Puppet in
+# UTF-8. Etc is defined for Windows but calls to it return nil.
+#
+# We only use Etc for retrieving existing property values from the system. For
+# setting property values, providers leverage system tools (i.e., `useradd`)
+#
+# @api private
+module Puppet::Etc
+  class << self
+    # Etc::getgrent returns an Etc::Group struct object
+    # On first call opens /etc/group and returns parse of first entry. Each subsquent call
+    # returns new struct the next entry or nil if EOF. Call ::endgrent to close file.
+    def getgrent
+      if group_entry = ::Etc.getgrent
+        convert_field_values_to_utf8!(group_entry)
+      end
+      group_entry
+    end
+
+    # closes handle to /etc/group file
+    def endgrent
+      ::Etc.endgrent
+    end
+
+    # effectively equivalent to IO#rewind of /etc/group
+    def setgrent
+      ::Etc.setgrent
+    end
+
+    # Etc::getpwent returns an Etc::Passwd struct object
+    # On first call opens /etc/passwd and returns parse of first entry. Each subsquent call
+    # returns new struct for the next entry or nil if EOF. Call ::endgrent to close file.
+    def getpwent
+      if user_entry = ::Etc.getpwent
+        convert_field_values_to_utf8!(user_entry)
+      end
+      user_entry
+    end
+
+    # closes handle to /etc/passwd file
+    def endpwent
+      ::Etc.endpwent
+    end
+
+    #effectively equivalent to IO#rewind of /etc/passwd
+    def setpwent
+      ::Etc.setpwent
+    end
+
+    # Etc::getpwnam searches /etc/passwd file for an entry corresponding to
+    # username.
+    # returns an Etc::Passwd struct corresponding to the entry or raises
+    # ArgumentError if none
+    def getpwnam(username)
+      if user_entry = ::Etc.getpwnam(username)
+        convert_field_values_to_utf8!(user_entry)
+      end
+      user_entry
+    end
+
+    # Etc::getgrnam searches /etc/group file for an entry corresponding to groupname.
+    # returns an Etc::Group struct corresponding to the entry or raises
+    # ArgumentError if none
+    def getgrnam(groupname)
+      if group_entry = ::Etc.getgrnam(groupname)
+        convert_field_values_to_utf8!(group_entry)
+      end
+      group_entry
+    end
+
+    # Etc::getgrid searches /etc/group file for an entry corresponding to id.
+    # returns an Etc::Group struct corresponding to the entry or raises
+    # ArgumentError if none
+    def getgrgid(id)
+      if group_entry = ::Etc.getgrgid(id)
+        convert_field_values_to_utf8!(group_entry)
+      end
+      group_entry
+    end
+
+    # Etc::getpwuid searches /etc/passwd file for an entry corresponding to id.
+    # returns an Etc::Passwd struct corresponding to the entry or raises
+    # ArgumentError if none
+    def getpwuid(id)
+      if user_entry = ::Etc.getpwuid(id)
+        convert_field_values_to_utf8!(user_entry)
+      end
+      user_entry
+    end
+
+    private
+    # Utility method for converting the String values of a struct returned by
+    # the Etc module to UTF-8. Structs returned by the ruby Etc module contain
+    # members with fields of type String, Integer, or Array of Strings, so we
+    # handle these types. Otherwise ignore fields.
+    #
+    # NOTE: If a string cannot be converted to UTF-8, this leaves the original
+    # string string intact in the Struct.
+    #
+    # Warning! This is a destructive method - the struct passed is modified!
+    #
+    # @api private
+    # @param [Etc::Passwd or Etc::Group struct]
+    # @return [Etc::Passwd or Etc::Group struct] the original struct with values
+    #   converted to UTF-8 if possible, or the original value intact if not
+    def convert_field_values_to_utf8!(struct)
+      struct.each_with_index do |value, index|
+        if value.is_a?(String)
+          begin
+            struct[index] = Puppet::Util::CharacterEncoding.convert_to_utf_8(value)
+          rescue Puppet::Error
+            # struct[index] unmodified
+          end
+        elsif value.is_a?(Array) && value.all? { |elem| elem.is_a?(String) }
+          struct[index] = convert_array_values_to_utf8!(value)
+        end
+      end
+    end
+
+    # Helper method for ::convert_field_values_to_utf8!
+    #
+    # Warning! This is a destructive method - the array passed is modified!
+    #
+    # @api private
+    # @param [Array] object containing String values to convert to UTF-8
+    # @return [Array] original Array with String values converted to UTF-8 if
+    #   convertible, or original, unmodified values if not.
+    def convert_array_values_to_utf8!(string_array)
+      string_array.map! do |elem|
+        begin
+          Puppet::Util::CharacterEncoding.convert_to_utf_8(elem)
+        rescue Puppet::Error
+          elem # individual array element unmodified
+        end
+      end
+    end
+  end
+end
+
+
+
--- a/lib/puppet/etc.rb
+++ b/lib/puppet/etc.rb
@@ -18,10 +18,7 @@ module Puppet::Etc
     # On first call opens /etc/group and returns parse of first entry. Each subsquent call
     # returns new struct the next entry or nil if EOF. Call ::endgrent to close file.
     def getgrent
-      if group_entry = ::Etc.getgrent
-        convert_field_values_to_utf8!(group_entry)
-      end
-      group_entry
+      convert_field_values_to_utf8!(::Etc.getgrent)
     end
 
     # closes handle to /etc/group file
@@ -38,10 +35,7 @@ module Puppet::Etc
     # On first call opens /etc/passwd and returns parse of first entry. Each subsquent call
     # returns new struct for the next entry or nil if EOF. Call ::endgrent to close file.
     def getpwent
-      if user_entry = ::Etc.getpwent
-        convert_field_values_to_utf8!(user_entry)
-      end
-      user_entry
+      convert_field_values_to_utf8!(::Etc.getpwent)
     end
 
     # closes handle to /etc/passwd file
@@ -59,40 +53,28 @@ module Puppet::Etc
     # returns an Etc::Passwd struct corresponding to the entry or raises
     # ArgumentError if none
     def getpwnam(username)
-      if user_entry = ::Etc.getpwnam(username)
-        convert_field_values_to_utf8!(user_entry)
-      end
-      user_entry
+      convert_field_values_to_utf8!(::Etc.getpwnam(username))
     end
 
     # Etc::getgrnam searches /etc/group file for an entry corresponding to groupname.
     # returns an Etc::Group struct corresponding to the entry or raises
     # ArgumentError if none
     def getgrnam(groupname)
-      if group_entry = ::Etc.getgrnam(groupname)
-        convert_field_values_to_utf8!(group_entry)
-      end
-      group_entry
+      convert_field_values_to_utf8!(::Etc.getgrnam(groupname))
     end
 
     # Etc::getgrid searches /etc/group file for an entry corresponding to id.
     # returns an Etc::Group struct corresponding to the entry or raises
     # ArgumentError if none
     def getgrgid(id)
-      if group_entry = ::Etc.getgrgid(id)
-        convert_field_values_to_utf8!(group_entry)
-      end
-      group_entry
+      convert_field_values_to_utf8!(::Etc.getgrgid(id))
     end
 
     # Etc::getpwuid searches /etc/passwd file for an entry corresponding to id.
     # returns an Etc::Passwd struct corresponding to the entry or raises
     # ArgumentError if none
     def getpwuid(id)
-      if user_entry = ::Etc.getpwuid(id)
-        convert_field_values_to_utf8!(user_entry)
-      end
-      user_entry
+      convert_field_values_to_utf8!(::Etc.getpwuid(id))
     end
 
     private
@@ -111,14 +93,12 @@ module Puppet::Etc
     # @return [Etc::Passwd or Etc::Group struct] the original struct with values
     #   converted to UTF-8 if possible, or the original value intact if not
     def convert_field_values_to_utf8!(struct)
+      return nil if struct.nil?
       struct.each_with_index do |value, index|
         if value.is_a?(String)
-          begin
-            struct[index] = Puppet::Util::CharacterEncoding.convert_to_utf_8(value)
-          rescue Puppet::Error
-            # struct[index] unmodified
-          end
-        elsif value.is_a?(Array) && value.all? { |elem| elem.is_a?(String) }
+            converted = Puppet::Util::CharacterEncoding.convert_to_utf_8!(value)
+            struct[index] = converted if !converted.nil?
+        elsif value.is_a?(Array)
           struct[index] = convert_array_values_to_utf8!(value)
         end
       end
@@ -134,11 +114,8 @@ module Puppet::Etc
     #   convertible, or original, unmodified values if not.
     def convert_array_values_to_utf8!(string_array)
       string_array.map! do |elem|
-        begin
-          Puppet::Util::CharacterEncoding.convert_to_utf_8(elem)
-        rescue Puppet::Error
-          elem # individual array element unmodified
-        end
+        converted = Puppet::Util::CharacterEncoding.convert_to_utf_8!(elem)
+        converted.nil? ? elem : converted
       end
     end
   end
--- a/lib/puppet/util/character_encoding.rb
+++ b/lib/puppet/util/character_encoding.rb
@@ -1,24 +1,29 @@
 # A module to centralize heuristics/practices for managing character encoding in Puppet
-require 'puppet/util'
 
 module Puppet::Util::CharacterEncoding
   class << self
-    #TODO make encoding warnings a category/silenceable#.
-
+    # Warning! This is a destructive method - the string supplied is modified!
     # @api public
     # @param [String] string a string to transcode / force_encode to utf-8
     # @return [String] string if already utf-8, OR
-    #   a copy of string with external encoding set to utf-8 if bytes are valid utf-8 OR
-    #   a copy of string transcoded to utf-8
-    # @raise [Puppet::Error] a failure to legitimately set external encoding or
-    #   transcode string
-    def convert_to_utf_8(string)
-      return string if string.encoding == Encoding::UTF_8
-
-      value_to_encode = string.dup
+    #   the same string with external encoding set to utf-8 if bytes are valid utf-8 OR
+    #   the same string transcoded to utf-8 OR
+    #   nil upon a failure to legitimately set external encoding or transcode string
+    def convert_to_utf_8!(string)
+      currently_valid = string.valid_encoding?
 
       begin
-        if valid_utf_8?(value_to_encode)
+        if string.encoding == Encoding::UTF_8
+          if currently_valid
+            return string
+          else
+            # If a string is currently believed to be UTF-8, but is also not
+            # valid_encoding?, we have no recourse but to fail because we have no
+            # idea what encoding this string originally came from where it *was*
+            # valid - all we know is it's not currently valid UTF-8.
+            raise EncodingError
+          end
+        elsif valid_utf_8_bytes?(string)
           # Before we try to transcode the string, check if it is valid UTF-8 as
           # currently constitued (in its non-UTF-8 encoding), and if it is, limit
           # ourselves to setting the external encoding of the string to UTF-8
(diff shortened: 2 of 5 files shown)


MosesMendoza:
depends on/includes https://github.com/puppetlabs/puppet/pull/5633

spec/unit/etc_spec.rb
```diff
@@ -0,0 +1,155 @@
+require 'puppet/etc'
+require 'spec_helper'
+
+# The Ruby::Etc module is largely non-functional on Windows - many methods
+# simply return nil regardless of input, the Etc::Group struct is not defined,
+# and Etc::Passwd is missing fields
+describe Puppet::Etc, :if => !Puppet.features.microsoft_windows? do
```
MosesMendoza: Per the preceding code comment, the upstream (ruby) Etc module methods return nil on Windows, and the `Etc::Passwd` and `Etc::Group` structs are missing member fields, so I've gated this test to not run on Windows. The original code that used Etc in Puppet appears to all be posix-specific and Windows is thusly avoided which makes sense. I haven't gated loading of `Puppet::Etc` above behind some sort of Windows platform test, but if one does `require 'puppet/etc'` on Windows and uses the methods of `Puppet::Etc`, they'll just return `nil` like their upstream counterparts. That made the most sense to me at the time as I was trying to mimic upstream behavior as closely as possible (excepting the UTF-8 conversion) but I'm all ears if there's a more standard way to do this.

adrienthebo:
This probably needs to be rebased on master to incorporate the changes to #5633 that aren't mirrored in this PR.

spec/unit/etc_spec.rb
adrienthebo: IMO it's reasonable to surface the underlying behavior of `Etc` on Windows; in an ideal world the underlying code would raise a `NotImplementedError` if invoked on Windows but adding that here would be changing the contract. :+1: for this as it is.

MosesMendoza:
rebased

lib/puppet/etc.rb
```diff
@@ -0,0 +1,118 @@
...
+    #
+    # Warning! This is a destructive method - the struct passed is modified!
+    #
+    # Structs returned by the ruby Etc module contain members with
+    # fields of type String, Integer, or Array of Strings, so we handle these
+    # types. Otherwise ignore fields.
+    def convert_field_values_to_utf8!(struct)
+      struct.each_with_index do |value, index|
+        if value.is_a?(String)
+          struct[index] = Puppet::Util::CharacterEncoding.convert_to_utf_8(value)
```
branan: As written, if any data in the struct is not transcodable to UTF-8, we'll fail the entire Etc read. This could mean that if a user is managing a particular property of a User (or Group), and an unmanaged property is non-UTF-8, it will cause the resource to fail at prefetch. This may be the correct behavior, but I wanted to raise it for discussion

MosesMendoza: Thanks for pointing that out @branan. I'm not sure if this is the right thing to do.. I'm very slightly inclined to say it would be more surprising if puppet managed some properties of a user and failed to manage others (assuming these are supported properties under management) and as a user I would rather have puppet fail to manage the user than fail to manage some parts of it. The case I'm thinking of is something like group membership, where we might create a new user and add it some unprivileged UTF-8 named group, or perhaps want to remove an existing user from a privileged UTF-8 named group. I think I'd rather have the user not created than have it created and put in incorrect groups. Wouldn't mind a second opinion from @puppet-maintainers..

MosesMendoza:
this should be retargeted at 4.10.x

MosesMendoza:
rebased against 4.10.x

adrienthebo:
@MosesMendoza we good to merge this when CI passes?

MosesMendoza:
@adrienthebo I think so - as long as you read @branan's comment above and agree with the PR as-is?

lib/puppet/etc.rb
adrienthebo: Between partially managing a user's properties and failing to apply the resource entirely, the latter option is 1) visible and 2) less dangerous. Failing early on bad input means we won't leave a system in a partially managed, but mangled, state.

adrienthebo:
@MosesMendoza seems reasonable to me. Given that it's Friday afternoon I'm going to leave the merge to future me, so that present me can enjoy this sunshine for the few hours that it'll last. :)

MosesMendoza:
👍  sounds good @adrienthebo. good to merge then

lib/puppet/etc.rb
joshcooper: I would expect puppet to only raise an exception if it can't UTF-8 encode a string it's managing. I think it's pretty common to only manage a subset of the fields in the various structs, e.g. `gecos`.

lib/puppet/etc.rb
```diff
@@ -0,0 +1,118 @@
...
+# This represents a subset of Ruby's Etc module, only the methods required by Puppet.
+# Etc returns strings in variable encoding depending on
+# environment. For Puppet we specifically want UTF-8 as our input from the Etc
+# module - which is our input for many resource instance 'is' values. The
+# associated 'should' value will basically always be coming from Puppet in
+# UTF-8. Etc is defined for Windows but calls to it return nil.
+#
+# We only use Etc for retrieving existing property values from the system. For
+# setting property values, providers leverage system tools (i.e., `useradd`)
+module Puppet::Etc
```
joshcooper: Should this be `@api private`?

lib/puppet/etc.rb
MosesMendoza: Thanks @joshcooper. It wasn't clear to me the implications of this, but I see that logic - it would be confusing for a user to suddenly encounter failures for properties of resources that aren't managed at all. Despite @branan's saying just that I hadn't connected the dots.

The beauty of the approach in this PR was its position at a single point of ingress for a ton of data that enters as `is` values during a transaction. Unfortunately, the actual handling and manipulation of the specific properties under management (subsets of any given `Etc` struct) is spread out across myriad providers in a web of meta-programming and obfuscation (see `nameservice` and `useradd` providers), so scope expands quite a bit if we're going to fix this there (we've been trying to do this, and it's been error-prone whack-a-mole).

For the purposes of this PR, what if instead of failing, `Puppet::Etc` is augmented with a (non-!) method that warns on the inability to convert the string, and just leaves the original string intact? In some cases, this is just going to fail upstream in the provider, but then at least a) its only failing the managed resource and b) its the same failure that was happening before. Thoughts?

lib/puppet/etc.rb
MosesMendoza: @joshcooper ie, documented as private API, but not explicitly ruby private?

lib/puppet/etc.rb
joshcooper: 👍  to warning (maybe only when `strict` is enabled) and preserving the original string and its encoding.

lib/puppet/etc.rb
joshcooper: Yeah thinking we should specify it's a private API in docs so that we can make changes as needed in the future. The methods themselves don't need to be ruby private though, e.g. `private :somemethod`

lib/puppet/etc.rb
```diff
@@ -0,0 +1,118 @@
...
+  class << self
+    # Etc::getgrent returns an Etc::Group struct object
+    # On first call opens /etc/group and returns parse of first entry. Each subsquent call
+    # returns new struct the next entry or nil if EOF. Call ::endgrent to close file.
+    def getgrent
+      if group_entry = Etc.getgrent
```
Iristyle: Would it be more clear to use the scope operator here so it's clear that we're using the Ruby methods.

Also.. nit, but I usually try to refrain from assignment inside conditionals.

```ruby
if group_entry = ::Etc.getgrent
```

lib/puppet/etc.rb
adrienthebo: :+1:, along with :+1: on Josh's suggested changes.

Iristyle: Can we run a really quick set of perf numbers on this to get a feeling for cost?

Nothing fancy...

spec/unit/etc_spec.rb
```diff
@@ -0,0 +1,155 @@
+require 'puppet'
+require 'spec_helper'
+
+# The Ruby::Etc module is largely non-functional on Windows - many methods
+# simply return nil regardless of input, the Etc::Group struct is not defined,
+# and Etc::Passwd is missing fields
+describe Puppet::Etc, :if => !Puppet.features.microsoft_windows? do
+  let(:bin) { 'bin'.force_encoding(Encoding::ISO_8859_1) }
+  let(:root) { 'root'.force_encoding(Encoding::ISO_8859_1) }
```
Iristyle: Would it make sense to vary the encodings here to be something other than `Encoding::ISO_8859_1` for each property?

Also - should we bring in our standard multibyte UTF-8 here somewhere?  I'm not sure I see anything validating that UTF-8 still works as desired.

lib/puppet/etc.rb
MosesMendoza: 👍  to feedback. Will update to not fail. I'm not sure anymore about even warning though, as in this module we won't have knowledge necessarily of the resource being managed, so we'll really just be printing the value out of context. Maybe not that helpful? Also possibly information disclosure?

> Can we run a really quick set of perf numbers on this to get a feeling for cost?

Yes, for sure. Were you thinking adding benchmarks, or would some ad-hoc Etc read comparisons between this module and Ruby suffice?

lib/puppet/etc.rb
MosesMendoza: 👍

lib/puppet/etc.rb
MosesMendoza: 👍

spec/unit/etc_spec.rb
MosesMendoza: yes on both counts. missing additional utf-8 validation was an oversight.

MosesMendoza:
@joshcooper @Iristyle commits added responding to feedback

lib/puppet/etc.rb
```diff
@@ -0,0 +1,126 @@
...
+    # Warning! This is a destructive method - the struct passed is modified!
+    #
+    # @api private
+    # @param [Etc::Passwd or Etc::Group struct]
+    # @return [Etc::Passwd or Etc::Group struct] the original struct with values
+    #   converted to UTF-8 if possible, or the original value intact if not
+    def convert_field_values_to_utf8!(struct)
+      struct.each_with_index do |value, index|
+        if value.is_a?(String)
+          struct[index] = Puppet::Util::CharacterEncoding.convert_to_utf_8(value) rescue value
```
MosesMendoza: It's generally frowned upon to do a blanket `rescue` like this as opposed to a more explicit class rescue (in this case would be `Puppet::Error` but since we really want to ensure that the string is untouched it seemed appropriate here. Interested to hear what others think.

Iristyle: In the helper you're already rescuing `EncodingError` ... and then re-raising with it.  What else do you expect to catch?

I also noticed that you're actually eating the previous error, which you might want to fix in this PR - see https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/character_encoding.rb#L70

MosesMendoza: > In the helper you're already rescuing EncodingError ... and then re-raising with it. What else do you expect to catch?

Well, strictly speaking that's rescuing EncodingError and then raising a Puppet::Error with a new message. Since there are several different kinds of Ruby EncodingError errors, that's generally to ensure we have a consistent error message/class across conversion attempts, an error that we route through Puppet.

I'm not sure what you mean re: eating the previous error?

MosesMendoza: Or maybe you mean that we should be surfacing that error here? I intended for Puppet::Util::CharacterEncoding::convert_to_utf_8 to raise a Puppet::Error if it could not convert a string to UTF-8 and allow users to handle that error as they saw fit - in this case, per feedback we're hiding the failure to convert and just continuing?

MosesMendoza: Chatted with @Iristyle - this comment was in reference to the fact that the Puppet::Error constructor, here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/character_encoding.rb#L70, is not of the form that accepts an original exception to wrap. This should be updated to wrap the exception, so that original error context is maintained. 

We also agreed, as I expected, that we should not be doing the blank exception catch. Simply too much that could fail outside of our conversion attempt. Will refactor to explicit block form to rescue Puppet::Error only. Relevant is this ruby feature request for a one-line rescue syntax supporting exception classes: https://bugs.ruby-lang.org/issues/6739.

MosesMendoza: PR updated with new commits to reflect this. Will squash after approval

Iristyle: 👍

MosesMendoza:
Some basic basic benchmark comparisons between the ruby Etc and this wrapper:

Environment: Ubuntu 16.04 vmware VM with 4GB ram / 2 processor, running this code as a ruby gem on system ruby, version 2.3.1

**Tests**
* In both a Non-UTF-8 environment (ISO-8859-1) and a UTF-8 environment:
  * Look up a user 100,000 times
  * Look up a group with 5 members 100,000 times
  * Look up a group containing 1000 members 5000 times
  * Do so using `Etc::getpwnam` and `Etc::getgrnam`, respectively, and their Puppet::Etc equivalents

**Expectations**
* We expect the non-UTF-8 environment to be more resource-intensive, as the strings inside the struct are converted to UTF-8, while the UTF-8 environment should presumably only iterate over all strings but not transform them.
* We expect group lookup to take slightly longer as well, as we also iterate over a the group membership array contained in the struct
* As far as I can tell, the ruby `Etc::getgrnam` and `Etc::getpwnam` perform no caching - just delegate to the system POSIX getgrnam, getpwnam, ie http://man7.org/linux/man-pages/man3/getgrnam.3.html

**Results**
_UTF-8 Environment_
* User Test
```
[4] pry(main)> Benchmark.bm do |user_test|
[4] pry(main)*   user_test.report("ruby Etc user UTF-8") { 100000.times { Etc.getpwnam('foo') } }
[4] pry(main)*   user_test.report("Puppet::Etc user UTF-8") { 100000.times { Puppet::Etc.getpwnam('foo') } }
[4] pry(main)* end
ruby Etc user UTF-8  0.700000   0.490000   1.190000 (  1.188727)
Puppet::Etc user UTF-8  1.040000   0.520000   1.560000 (  1.572868)
```
Lookup of a user 100,000 times takes ~0.4 seconds wall-clock time longer in aggregate using `Puppet::Etc`

* Group lookup test - group with 5 members, 100,000 times
```
[4] pry(main)> Benchmark.bm do |test|
[4] pry(main)*   test.report("ruby Etc UTF-8") { 100000.times { Etc.getgrnam('agroup') } }
[4] pry(main)*   test.report("Puppet::Etc UTF-8") { 100000.times { Puppet::Etc.getgrnam('agroup') } }
[4] pry(main)* end
ruby Etc UTF-8  0.860000   0.500000   1.360000 (  1.381253)
Puppet::Etc UTF-8  1.260000   0.620000   1.880000 (  1.892280)
```
Lookup of a group with 5 members 100,000 times takes ~0.5 seconds wall-clock time longer in aggregate using `Puppet::Etc`

* Group lookup test - group with 1,000 members, 5,000 times
```
[11] pry(main)> Benchmark.bm do |test|
[11] pry(main)*   test.report("ruby Etc UTF-8") { 5000.times { Etc.getgrnam('bgroup') } }
[11] pry(main)*   test.report("Puppet::Etc UTF-8") { 5000.times { Puppet::Etc.getgrnam('bgroup') } }
[11] pry(main)* end
ruby Etc UTF-8  0.790000   0.030000   0.820000 (  0.823944)
Puppet::Etc UTF-8  1.710000   0.050000   1.760000 (  1.766623)
```

Lookup of a group containing 1,000 members 5,000 times takes ~1 second wall-clock time longer in aggregate using `Puppet::Etc`

_ISO-8859-1 Environment_
* User lookup test
```
[5] pry(main)> Benchmark.bm do |user_test|
[5] pry(main)*   user_test.report("ruby Etc user") { 100000.times { Etc.getpwnam('foo') } }
[5] pry(main)*   user_test.report("Puppet::Etc user") { 100000.times { Puppet::Etc.getpwnam('foo') } }
[5] pry(main)* end
ruby Etc user  0.720000   0.500000   1.220000 (  1.234082)
Puppet::Etc user  1.370000   0.550000   1.920000 (  1.931865)
```
Querying the user 100,000 times takes ~0.7 seconds wall-clock time longer in aggregate using `Puppet::Etc`

* Group lookup test - group with 5 members, 100,000 times
```
[3] pry(main)> Benchmark.bm do |test|
[3] pry(main)*   test.report("ruby Etc") { 100000.times { Etc.getgrnam('agroup') } }
[3] pry(main)*   test.report("Puppet::Etc") { 100000.times { Puppet::Etc.getgrnam('agroup') } }
[3] pry(main)* end
ruby Etc  0.860000   0.480000   1.340000 (  1.350584)
Puppet::Etc  1.850000   0.550000   2.400000 (  2.418090)
```
Querying the 5 member group 100,000 times takes ~1.1 seconds wall-clock time longer in aggregate using `Puppet::Etc`.

* Group lookup test - group with 1,000 members, 5,000 times
```
[5] pry(main)> Benchmark.bm do |test|
[5] pry(main)*   test.report("ruby Etc UTF-8") { 5000.times { Etc.getgrnam('bgroup') } }
[5] pry(main)*   test.report("Puppet::Etc UTF-8") { 5000.times { Puppet::Etc.getgrnam('bgroup') } }
[5] pry(main)* end
ruby Etc UTF-8  0.890000   0.020000   0.910000 (  0.925681)
Puppet::Etc UTF-8  4.780000   0.060000   4.840000 (  4.834927)
```
Querying the 1,000 member group 5,000 times takes ~4 seconds wall-clock time longer in aggregate using `Puppet::Etc`. This is the worst performance delta in the set of tests.

This implies there is certainly a delta. `self.instances` in the nameservice provider iterates over the entire `/etc/passwd` or `/etc/group` file via repeated calls `Etc.getgrent` and `Etc.getpwent`, each of which creates a new struct from the current line in the file, so a user with an `/etc/group` file 100,000 lines long composed of groups containing five users could expect `self.instances` to take about a second longer in a non-utf-8 environment and half a second longer in a UTF-8 environment. Likewise an environment containing with an `/etc/group` file containing 5,000 groups each with 1,000 members could expect to see a 4 second slow-down. When the number of lookups in that last test was reduced to 500, the aggregate lookup time was ~.5 seconds. At 10,000 lookups, the time was ~9.6 seconds - which implies the slow-down grows linearly with the number of groups, at a rate of .001 seconds per added group with 1000 members.

MosesMendoza:
@Iristyle interested to get your take on these numbers

MosesMendoza:
blocked pending perf investigation

lib/puppet/util/character_encoding.rb
```diff
@@ -13,12 +12,16 @@ class << self
...
+      return string if (string.encoding == Encoding::UTF_8 && string.valid_encoding?)
 
       begin
-        if valid_utf_8?(value_to_encode)
+        if already_invalid_utf_8?(string)
+          # If a string is currently believed to be UTF-8, but is also not
+          # valid_encoding?, we have no recourse but to fail because we have no
+          # idea what encoding this string originally came from where it *was*
+          # valid - all we know is it's not currently valid UTF-8.
+          raise EncodingError
```
MosesMendoza: Calling out these lines specifically as a new addition for review - change to the heuristic, not necessarily related to the Etc wrapper but discovered when trying to improve it.

MosesMendoza:
I was able to get the bm approx 33% faster in that large test just by avoiding extra string creation - going to refactor to add a destructive `!` form that manipulates the string.

with string object duplication:
```ruby
☂ : be ruby ~/user_bm.rb
ruby etc  2.970000   0.130000   3.100000 (  3.092196)
Puppet etc  6.630000   0.100000   6.730000 (  6.744819)
```
with manipulation of string to convert:
```ruby
☂ : be ruby ~/user_bm.rb
ruby etc  2.850000   0.100000   2.950000 (  2.958789)
Puppet etc  4.790000   0.070000   4.860000 (  4.863782)
```

MosesMendoza:
Refactored to modify strings in place rather than create duplicate objects. Updated numbers for this PR:

For groups containing a member name that does not convert cleanly (cannot be converted to UTF-8):
_UTF-8 Environment_
```
::Etc lookup group with 5 members 100000 times  0.820000   0.690000   1.510000 (  1.523810)
Puppet::Etc with 5 members 100000 times  2.520000   0.850000   3.370000 (  3.381769)
```

```
::Etc group with 1000 members 5000 times  0.790000   0.030000   0.820000 (  0.811397)
Puppet::Etc group with 1000 members 5000 times  1.450000   0.030000   1.480000 (  1.493500)
```

_ISO-8859-1 Environment_
```
::Etc lookup group with 5 members 100000 times  1.050000   0.630000   1.680000 (  1.693595)
Puppet::Etc with 5 members 100000 times  2.130000   0.840000   2.970000 (  2.975335)
```

```
::Etc group with 1000 members 5000 times  0.890000   0.050000   0.940000 (  0.950036)
Puppet::Etc group with 1000 members 5000 times  2.950000   0.040000   2.990000 (  3.000961)
```

For groups with strings that convert cleanly:
_UTF-8 Environment_
```
::Etc lookup group with 5 members 100000 times  1.080000   0.730000   1.810000 (  1.817930)
Puppet::Etc with 5 members 100000 times  1.470000   0.710000   2.180000 (  2.190792)
```
```
::Etc group with 1000 members 5000 times  0.820000   0.050000   0.870000 (  0.869856)
Puppet::Etc group with 1000 members 5000 times  1.620000   0.040000   1.660000 (  1.657790)
```

_ISO-8859-1 Environment_
```
::Etc lookup group with 5 members 100000 times  0.940000   0.670000   1.610000 (  1.624854)
Puppet::Etc with 5 members 100000 times  1.510000   0.710000   2.220000 (  2.227365)
```

```
::Etc group with 1000 members 5000 times  0.870000   0.060000   0.930000 (  0.924928)
Puppet::Etc group with 1000 members 5000 times  2.880000   0.040000   2.920000 (  2.927006)
```

The overhead at this point appears to be primarily related to generating/catching exception objects, and just the logic itself. I played around with reducing the objects etc and managed to shave about .5 seconds off the worst case but it felt like a pyrrhic victory given the cost to readability.

lib/puppet/etc.rb
```diff
@@ -0,0 +1,154 @@
...
+    # Etc::getgrent returns an Etc::Group struct object
+    # On first call opens /etc/group and returns parse of first entry. Each subsquent call
+    # returns new struct the next entry or nil if EOF. Call ::endgrent to close file.
+    def getgrent
+      group_entry = ::Etc.getgrent
+      if !group_entry.nil?
+        convert_field_values_to_utf8!(group_entry)
+      end
+      group_entry
+    end
```
joshcooper: we could simply all of the `get*` methods by making the `convert_*` method nil safe:

```ruby
def getgrent
  to_utf8 ::Etc.getgrent)
end

def to_utf8(entry)
  return nil if entry.nil?
  ...
end
```

lib/puppet/util/character_encoding.rb
joshcooper: I think this method could be simplified so we don't call `valid_encoding?` more than once (even though it looks like ruby will memoize the result, until the string is next mutated). The reason I mention that is `valid_encoding?` will at worst case scan the string multiple times (once to see if there are any non-US ASCII bytes, and then again looking for multi-byte sequences). So maybe something like?

```ruby
valid = string.valid_encoding?

if string.encoding == Encoding::UTF_8
  if valid
    string
  else
    raise EncodingError
  end
elsif valid_utf_8?(string)
  string.force_encoding(Encoding::UTF_8)
else
  string.encode!(Encoding::UTF_8)
end
```

That complexity is because we're trying to figure out if the string is already valid UTF-8 or could be force encoded. It would take a lot of string encodings for that to matter I think. For example, if I run a benchmark 10M times, it takes 4.2 seconds to create the various string objects. If I also try to UTF-8 encode a string already in UTF-8 encoding, it takes 7.57 seconds. So it's not a noop, but it takes about as long to "noop encode" the string as it does to create the string itself, probably due to making a copy of the string. But if I UTF-8 encode a string that's in ISO-8859-1 (even though the byte sequences are exactly the same), it takes over 10x times as long. So ruby is doing a decent job at short-circuiting the encoding logic when source and destination encodings are the same.

```
$ ruby enc.rb
            baseline  4.220000   0.070000   4.290000 (  4.279521)
      same encodings  7.570000   0.010000   7.580000 (  7.591500)
      diff encodings 81.410000   0.750000  82.160000 ( 82.362164)
```

If we wanted to simplify the logic, we could do something like:

```ruby
begin
  string.encode!(Encoding::UTF_8)
rescue EncodingError => e
  original_encoding = string.encoding
  string.force_encoding(Encoding::UTF_8)
  if string.valid_encoding?
    string
  else
    string.force_encoding(original_encoding)
    raise e
  end
end
```

lib/puppet/etc.rb
MosesMendoza: nice. I hadn't thought of that.

lib/puppet/util/character_encoding.rb
MosesMendoza: I tried to get some numbers on `valid_encoding?` and at first it seemed like basically a no-op compared with the other operations... But I suppose it adds up on multiple checks. For example this string:
`"foo bar baz f dsa fs a fsa fsd afdsa fdas fdsa fa a afdsa afa  \xfd\xf1 \u2603"` which contains both invalid byte sequences and valid unicode, running `valid_encoding?` 10,000,000 times takes about 1 second. Calling it 3x more times is 3 more seconds, so might as well as cut it back if we can!

> If we wanted to simplify the logic, we could do something like:
> begin
  string.encode!(Encoding::UTF_8)
rescue EncodingError => e
  original_encoding = string.encoding
  string.force_encoding(Encoding::UTF_8)
  if string.valid_encoding?
    string
  else
    string.force_encoding(original_encoding)
    raise e
  end
end

I actually want to avoid doing the `encode!` first because due to collisions we can end up transcoding a string (into another valid utf-8 string) when it was actually unicode bits to begin with. Part of the heuristic is to avoid transcoding what would already be valid utf-8 bits on the assumption that they started correct. For example, this character:

ᚠ (rune - 225, 154, 160) is valid UTF-8, but if read in on a system currently running in LATIN-1, will transcode cleanly to different UTF-8, but that's likely not what we want.
```
[21] pry(main)> s = [225, 154, 160].pack('C*').force_encoding(Encoding::ISO_8859_1)
=> "\xE1\x9A\xA0"
[22] pry(main)> s.encode!(Encoding::UTF_8)
=> "á\u009A "
[23] pry(main)> s.valid_encoding?
=> true
[24] pry(main)> s.bytes
=> [195, 161, 194, 154, 194, 160]
```

 (I actually added a test for it, here: https://github.com/puppetlabs/puppet/blob/master/spec/unit/util/character_encoding_spec.rb#L13-L28)

MosesMendoza:
@joshcooper updated per feedback

lib/puppet/util/character_encoding.rb
```diff
@@ -51,46 +57,39 @@ def convert_to_utf_8(string)
...
       rescue EncodingError => detail
-        # catch both our own self-determined failure to transcode as well as any
-        # error on ruby's part, ie Encoding::InvalidByteSequenceError or
-        # Encoding::UndefinedConversionError.
-        raise Puppet::Error, _("#{detail.inspect}: #{value_to_encode.dump} is not valid UTF-8 and cannot be transcoded by Puppet.")
+        # Catch both our own self-determined failure to transcode as well as any
+        # error on ruby's part, ie Encoding::UndefinedConversionError on a
+        # failure to encode!.
+        raise Puppet::Error.new(_("%{error}: %{value} is not valid UTF-8 and cannot be transcoded by Puppet.") %
+          { error: detail.inspect, value: string.dump }, detail)
```
joshcooper: We raise `Puppet::Error` as means of flow control (as in the caller rescues it, swallows it, and takes another code path). Perhaps would be better to warn about the error, and have the `convert_to_utf_8!` method return the converted string or `nil`. Then have the caller do something like:

```
map do |string|
  utf8 = convert_to_utf_8(string)
  utf8 ? utf8 : string
end
```

MosesMendoza: sounds good. I really really didn't like doing this raise/catch/re-raise business and it definitely is a huge part of the performance overhead (probably the biggest) but I couldn't think of another way to unify all the various Encoding errors we could raise into a single one with a clear message. I like that option - just don't raise any error at all :) This will certainly be much faster without it.

MosesMendoza: I tried out doing a Puppet.warning and returning nil, and then benchmarked by doing 100000 lookups of a group with five regular ascii test members, one valid unicode member, and one invalid unicode member and...
wow. It turns out adding `Puppet.warning` on failure to encode is almost 3x slower than not doing a warning (shave a second off that if we don't parse the preceding EncodingError), and almost 2x slower than just raising `Puppet::Error` and rescuing it in `Etc`.
With Puppet.warning instead of rescue/raise:
```
::Etc lookup group with 5 members 100000 times  0.980000   0.610000   1.590000 (  1.596053)
Puppet::Etc with 5 members 100000 times  5.080000   1.030000   6.110000 (  6.125440)
```
Without Puppet.warning (just return nil):
```
::Etc lookup group with 5 members 100000 times  0.920000   0.700000   1.620000 (  1.634029)
Puppet::Etc with 5 members 100000 times  1.830000   0.720000   2.550000 (  2.560253)
```
With exceptions for flow control (current PR):
```
::Etc lookup group with 5 members 100000 times  0.970000   0.590000   1.560000 (  1.558745)
Puppet::Etc with 5 members 100000 times  2.580000   0.880000   3.460000 (  3.478575)
```

I expect that's because these benchmarks just call `::Etc` and `Puppet::Etc` so in this PR the rescue means we never invoke any of the logging machinery, just create the superfluous exception objects. If we do Puppet.warning in `convert_to_utf_8!` that is always invoked since default log level is warning.

If I do `Puppet.debug` instead of `Puppet.warning` with log level at default (warning) the performance is roughly on par with raising the exceptions. Still a performance hit over not logging at all, but probably the best mix of user experience/performance? I have updated the PR on this assumption.
With Puppet.debug instead of rescue/raise:
```
::Etc lookup group with 5 members 100000 times  0.970000   0.670000   1.640000 (  1.642895)
Puppet::Etc with 5 members 100000 times  2.600000   0.870000   3.470000 (  3.470088)
```

MosesMendoza:
@joshcooper updated again per feedback, and squashed commits.

Iristyle:
Some additional work appears in #5799

PR #5547, opened 2017-01-25
(DO NOT MERGE)(maint) Minor file encoding changes

Iristyle:
These are leftovers that remain after #5545 and #5384 have been merged.

They probably should end up in separate tickets, but put up to gather additional feedback.

PR #5636, opened 2017-02-11
(maint) Explicitly specify encodings where missed

Iristyle:
Intended to be merged after:
- [x] #5384 
- [ ] #5547 
- [x] #5593

This cleans up remaining locations where an encoding should be specified, but was not.

Important commit is just the last one - currently https://github.com/puppetlabs/puppet/pull/5636/commits/30a2ef7375a98d9eef6c0db643d289ff6873ff90 with +57 diffstat

As noted in #5660, https://github.com/puppetlabs/puppet/pull/5636/commits/5cf9f31c21e29a14fb87ea1d94613c43c7cdc3a2 should fix calls to `StringIO.new` to `.encode` files passed in, rather than specifying an `:encoding`, which `StringIO` ignores.

### DIFF
--- a/lib/puppet/face/epp.rb
+++ b/lib/puppet/face/epp.rb
@@ -153,7 +153,7 @@ Puppet::Face.define(:epp, '0.0.1') do
       # Print to a buffer since the face needs to return the resulting string
       # and the face API is "all or nothing"
       #
-      buffer = StringIO.new
+      buffer = StringIO.new('', 'r+:UTF-8')
 
       if options[:e]
         buffer.print dump_parse(options[:e], 'command-line-string', options, false)
@@ -320,7 +320,7 @@ Puppet::Face.define(:epp, '0.0.1') do
       # Print to a buffer since the face needs to return the resulting string
       # and the face API is "all or nothing"
       #
-      buffer = StringIO.new
+      buffer = StringIO.new('', 'r+:UTF-8')
       status = true
       if options[:e]
         buffer.print render_inline(options[:e], compiler, options)
--- a/lib/puppet/file_bucket/file.rb
+++ b/lib/puppet/file_bucket/file.rb
@@ -95,7 +95,7 @@ class Puppet::FileBucket::File
     end
 
     def stream(&block)
-      s = StringIO.new(@contents)
+      s = StringIO.new(@contents, 'r+b')
       begin
         block.call(s)
       ensure
--- a/lib/puppet/file_system/memory_file.rb
+++ b/lib/puppet/file_system/memory_file.rb
@@ -41,7 +41,7 @@ class Puppet::FileSystem::MemoryFile
 
   def handle
     raise Errno::ENOENT unless exist?
-    StringIO.new(@properties[:content] || '')
+    StringIO.new(@properties[:content] || '', 'r+:UTF-8')
   end
 
   def duplicate_as(other_path)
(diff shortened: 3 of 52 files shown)


lib/puppet/network/http/compression.rb
```diff
@@ -18,9 +18,10 @@ module Active
     # return an uncompressed body if the response has been
     # compressed
     def uncompress_body(response)
+      # TODO: is response.body already a BINARY string?
       case response['content-encoding']
       when 'gzip'
-        return Zlib::GzipReader.new(StringIO.new(response.body)).read
```
Iristyle: Based on some debugging with @MosesMendoza on https://tickets.puppetlabs.com/browse/PUP-7251, we have determined that this change is not quite right.

When `response.body` is received, it is already an `Encoding::BINARY` string, which `StringIO` already understands when setting its encoding. Therefore additionally specifying `r+b` is actually meaningless.

The real problem here is that the value returned from `uncompress_body` is not in the same encoding as the value returned when compression is disabled.  With compression disabled, `response.body` is returned directly, and its encoding, as mentioned, is already `Encoding::BINARY`.

When using `Zlib::GzipReader`, the default encoding of the returned string is actually `Encoding.default_external` - which could be an array of things varying on the platform, like `Encoding::CP932` (in Japan) or `Encoding::ISO8859_1` (in Europe).  Fortunately, the `Zlib::GzipReader.new` initializer can accept options that specify `:encoding => Encoding::BINARY` so that we continue to return binary strings for PSON when compression is on, as we're doing when compression is off.

Therefore as part of [PUP-7251](https://tickets.puppetlabs.com/browse/PUP-7251) the fix for this will probably be:

```ruby
return Zlib::GzipReader.new(StringIO.new(response.body, 'r+b'), :encoding => Encoding::BINARY).read
```

lib/puppet/face/epp.rb
```diff
@@ -153,7 +153,7 @@
       # Print to a buffer since the face needs to return the resulting string
       # and the face API is "all or nothing"
       #
-      buffer = StringIO.new
+      buffer = StringIO.new('', 'r+:UTF-8')
```
Iristyle: `r+:UTF-8` is meaningless to `StringIO`.  Instead, ensure that `''` is in `UTF-8`

lib/puppet/file_bucket/file.rb
```diff
@@ -95,7 +95,7 @@ def initialize(content)
     end
 
     def stream(&block)
-      s = StringIO.new(@contents)
+      s = StringIO.new(@contents, 'r+b')
```
Iristyle:
```ruby
s = StringIO.new(@contents.dup.force_encoding(Encoding::BINARY))
```

lib/puppet/file_system/memory_file.rb
```diff
@@ -41,7 +41,7 @@ def each_line(&block)
 
   def handle
     raise Errno::ENOENT unless exist?
-    StringIO.new(@properties[:content] || '')
+    StringIO.new(@properties[:content] || '', 'r+:UTF-8')
```
Iristyle:
```ruby
StringIO.new((@properties[:content] || '').dup.force_encoding(Encoding::BINARY))
```

joshcooper:
Looks like more conflicts, probably due to i18n and json/compression changes.

melissa:
This is intended to be merged once work for https://tickets.puppetlabs.com/browse/PUP-9261 is finished. 

Once the work for that ticket is finished, we'll revisit this work to address the merge conflicts.

melissa:
There are some changes here that are important and we should carry forward, prioritization is challenging though.

@ehom is this something you can look into?

jtappa:
This needs a rebase

jtappa:
Looks like this PR has gone stale with no updates for many months. I'm going to close for now, but if anyone wishes to rebase and update it in the future, please feel free to reopen it.

PR #5637, opened 2017-02-11
(docs) exec's environment takes an array

igalic:
Intuitively, one may think that `exec`'s environment parameter takes a
Hash, however, it expects an Array, which was not documented anywhere.

### DIFF
--- a/lib/puppet/type/exec.rb
+++ b/lib/puppet/type/exec.rb
@@ -252,10 +252,11 @@ module Puppet
     end
 
     newparam(:environment) do
-      desc "Any additional environment variables you want to set for a
-        command.  Note that if you use this to set PATH, it will override
-        the `path` attribute.  Multiple environment variables should be
-        specified as an array."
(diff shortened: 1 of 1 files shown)


MikaelSmith:
@adrienthebo why is this waiting on contributor?

adrienthebo:
@igalic @MikaelSmith it looks like GitHub swallowed my review comment, sorry for the delay!

lib/puppet/type/exec.rb
```diff
@@ -252,10 +252,11 @@ def value=(*values)
     end
 
     newparam(:environment) do
-      desc "Any additional environment variables you want to set for a
-        command.  Note that if you use this to set PATH, it will override
-        the `path` attribute.  Multiple environment variables should be
-        specified as an array."
+      desc "An array of any additional environment variables you want to set for a
```
igalic: true… i should enhance it then…

MosesMendoza:
thanks @igalic

PR #5639, opened 2017-02-13
(PUP-7214) PUP-6663 causes a regression of PUP-535

heliocentric:
This fixes a regression introduced in PUP-6663 where
the package name contains a leading ' in some conditions

Strip out all uses of ' before applying the regex, as
according to most rpm guidelines ' is not a valid character.

See
https://fedoraproject.org/wiki/Packaging:Naming?rd=Packaging:NamingGuidelines

### DIFF
--- a/spec/unit/provider/package/rpm_spec.rb
+++ b/spec/unit/provider/package/rpm_spec.rb
@@ -7,12 +7,13 @@ describe provider_class do
 
   let (:packages) do
     <<-RPM_OUTPUT
-    cracklib-dicts 0 2.8.9 3.3 x86_64
+    'cracklib-dicts 0 2.8.9 3.3 x86_64
     basesystem 0 8.0 5.1.1.el5.centos noarch
     chkconfig 0 1.3.30.2 2.el5 x86_64
     myresource 0 1.2.3.4 5.el4 noarch
     mysummaryless 0 1.2.3.4 5.el4 noarch
     tomcat 1 1.2.3.4 5.el4 x86_64
+    '
     RPM_OUTPUT
   end
 
--- a/lib/puppet/provider/package/rpm.rb
+++ b/lib/puppet/provider/package/rpm.rb
@@ -19,7 +19,7 @@ Puppet::Type.type(:package).provide :rpm, :source => :rpm, :parent => Puppet::Pr
   # eventually become this Puppet::Type::Package::ProviderRpm class.
   # The query format by which we identify installed packages
   self::NEVRA_FORMAT = %Q{%{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\\n}
-  self::NEVRA_REGEX  = %r{^(\S+) (\S+) (\S+) (\S+) (\S+)$}
+  self::NEVRA_REGEX  = %r{^'?(\S+) (\S+) (\S+) (\S+) (\S+)$}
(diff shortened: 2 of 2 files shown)


adrienthebo:
First off, [PUP-533](https://tickets.puppetlabs.com/browse/PUP-533) doesn't look like it has anything to do with RPM and was closed as wontfix. Is this the ticket that you meant to link to?

Secondly, if this issue is fixing a regression then we really should have a test for it to prevent future regressions. The associated ticket does provide a reproduction case; do you think you could turn that into a unit test?

heliocentric:
Sorry, it was PUP-535 that I meant to reference.

Yes, I'll update the tests accordingly

heliocentric:
Tests updated, verified without the changes (failed on the exact behaviour we saw), and it succeeds with the patch

pkranenburg:
Cf. my comments in the PUP-6663 pull request.

May I suggest to fix the 'leading quote' issue by changing the regular expression that interprets the rpm query output as follows:

    self::NEVRA_REGEX  = %r{^'?(\S+) (\S+) (\S+) (\S+) (\S+)$}

(around line 24 in rpm.rb). I think it would be faster to do it here, rather then stripping quotes separately before applying the regexp. Also, the regexp is assigned the task of dealing with the textual format, so it would make sense to have it deal with the quoting characters too.

heliocentric:
process.each_line { |line|
          hash = nevra_to_hash(line)
          packages << new(hash) unless hash.empty?
        }

heliocentric:
yea, that's a better way, updated and updated the test to match what I was seeing so the regex would capture correctly.

spec/unit/provider/package/rpm_spec.rb
```diff
@@ -7,12 +7,13 @@
 
   let (:packages) do
     <<-RPM_OUTPUT
-    cracklib-dicts 0 2.8.9 3.3 x86_64
+    'cracklib-dicts 0 2.8.9 3.3 x86_64
```
MikaelSmith: So my understanding of this would be that rpm outputs a quoted list. Any chance the trailing quote would appear at the end of the last line, rather than by itself on a following line?

heliocentric: the quote is being added in the provider here:
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/rpm.rb#L134
which when passed to rpm via argv (where each argument is it's own string index) it gets retained in the output:

rpm -q --qf "'%{NAME} %|EPOCH?{%{EPOCH}}:{0}| %{VERSION} %{RELEASE} %{ARCH}\\n'" -p ~/c/vagrant-pe4-simp/rpms/simp-adapter-pe-0.0.1-0.noarch.rpm 
'simp-adapter-pe 0 0.0.1 0 noarch
'

pkranenburg: The final quote on a line by itself is expected because of the trailing new-line (\\n) in the rpm query formatting string NEVRA_FORMAT.

heliocentric: I'd have to get time to pry again and look at instances, I recall line having the same output style but I don't think I put it anywhere

MikaelSmith: Ok, makes sense. Thanks.

Iristyle:
This has been cherry-picked back to the stable branch as c4f21f40

PR #5641, opened 2017-02-14
(PUP-7212) Add a Pcore Annotation type and an annotate function

thallgren:
This PR adds the `Annotation` type as an `Object`. The actual implementation class is a subclass of `Puppet::Pops::Adaptable::Adapter` and may be extended from the Puppet Language as an `Object` type that uses `Annotation` as its parent.

The commit also adds the function `annotate` that can be used in several different ways (see documentation in ruby source for the function).

### DIFF
--- a/spec/unit/pops/types/p_object_type_spec.rb
+++ b/spec/unit/pops/types/p_object_type_spec.rb
@@ -1162,6 +1162,53 @@ describe 'The Object Type' do
       coll_descendants.each { |tc| expect(find_parent(tc, 'Pcore::CollectionType').name).to eq('Pcore::CollectionType') }
     end
   end
+
+  context 'when dealing with annotations' do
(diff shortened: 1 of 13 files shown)


spec/unit/pops/types/p_object_type_spec.rb
```diff
@@ -1162,6 +1162,53 @@ def find_parent(tc, parent_name)
...
+
+    it 'the Annotation type can be used as parent' do
+      code = <<-PUPPET
+        #{annotation}
+        notice(MyAdapter < Annotation)
+      PUPPET
+      expect(eval_and_collect_notices(code)).to eql(['true'])
+    end
+
+    it 'an annotation can added to an Object type' do
```
hlindberg: typo

hlindberg:
ok to merge once the small typo is corrected

PR #5647, opened 2017-02-18
(PUP-7224) Replace the Rgen AST model with one generated from Pcore

thallgren:
This PR contains a series of commits that, in the end, results use of an PL AST model that is fully generated (using a Rake task) from a Pcore TypeSet. The commits improves the generated Ruby code, prepares the Rgen model and Binder for transition, adds the rake task, performs the actual transition, and then adds some optimizations.


MosesMendoza:
@thallgren this is great. With this PR are we able to remove the vendored rgen from lib/puppet/vendor? Also this PR feels quite large to me.. for the sake of reviewers would it be possible to separate it into a few smaller discreet PRs that are still meaningful, even if with dependencies?

MosesMendoza:
Nevermind, I missed #5649

thallgren:
Given how easy it is to review the commits separately, I would not be in favor of splitting this into several PRs. Several tickets perhaps, but I already did that. I created the PUP-7223 epic to describe the full removal and then five tickets for the actual work.

hlindberg:
Reviewed in the context of #5648

lib/puppet/pops/validation/checker4_0.rb
```diff
@@ -610,7 +645,7 @@ def internal_check_illegal_assignment(o)
     else
       # recursively check all contents unless it's a lambda expression. A lambda may contain
       # local assignments
-      o.eContents.each {|model| internal_check_illegal_assignment(model) } unless o.is_a?(Model::LambdaExpression)
+      o._pcontents {|model| internal_check_illegal_assignment(model) } unless o.is_a?(Model::LambdaExpression)
```
hlindberg: Hm, the `_pcontent` is a somewhat strange name. Good to start with `_`, but is a `p` enough to denote pcore operations? I can imagine `_pcore_contents`, or `_pc_contents`  (to save three chars). IIRC there are not that many mehtods now (maybe just the `_pcontents`, but if we continue with others maybe it is best to include the full `pcore` in the name?

lib/puppet/pops/validation/checker4_0.rb
```diff
@@ -649,13 +684,13 @@ def check_ResourceBody(o)
   end
 
   def check_ResourceDefaultsExpression(o)
-    if o.form && o.form != :regular
+    if o.form != 'regular'
```
hlindberg: This is different from the original where 'form' was optional - is it now always a value ?

spec/unit/pops/benchmark_spec.rb
```diff
@@ -85,15 +67,15 @@ def json_load(string)
...
     puts "Marshal1: #{m}"
   end
 
   it "rgenjson", :profile => true do
     parser = Puppet::Pops::Parser::EvaluatingParser.new()
-    model = parser.parse_string(code).current
+    model = parser.parse_string(code).model
     dumped = json_dump(model)
     m = Benchmark.measure { 10000.times { json_load(dumped) }}
     puts "RGen Json: #{m}"
```
hlindberg: This mentions RGen - should say Pcore as the test now benchmarks the pcore serialization in comparison with a Ruby Marchal. (i.e. also change the `it "rgenjson"` to `it "pcore_json". Could also add a pcore msgpack variant here for completeness.

spec/unit/pops/evaluator/access_ops_spec.rb
```diff
@@ -24,115 +24,115 @@ def binary(s)
     # Note that the factory is not aware of Binary and cannot operate on a
     # literal binary. Instead, it must create a call to Binary.new() with the base64 encoded
     # string as an argument
-    CALL_NAMED(QREF("Binary"), true, [Base64.strict_encode64(s)])
+    CALL_NAMED(QREF("Binary"), true, [infer(Base64.strict_encode64(s))])
```
hlindberg: why is `infer` called here? the original seems to create an instance of Binary - this seems to create an instance of binary with an inferred type as the value ?

spec/unit/pops/evaluator/access_ops_spec.rb
```diff
@@ -24,115 +24,115 @@ def binary(s)
...
   end
 
   context 'The evaluator when operating on a String' do
     it 'can get a single character using a single key index to []' do
-      expect(evaluate(literal('abc')[1])).to eql('b')
+      expect(evaluate(literal('abc').test_access(1))).to eql('b')
```
hlindberg: not too happy about the name `test_access` as I keep reading it as a verb rather than a noun. Can we rename it to something that makes more sense? (I.e. I read it as something that tests access, not that is a special kind of access that is only performed by tests.

lib/puppet/pops/types/puppet_object.rb
```diff
@@ -11,6 +11,16 @@ module PuppetObject
   def _ptype
     self.class._ptype
   end
+
+  def _pall_contents(path, &block)
+  end
+
+  def _pcontents
```
hlindberg: not that happy with the name as it becomes part of the API of every AST node. Suggested `_pcore_contents` elsewhere.

lib/puppet/pops/types/puppet_object.rb
```diff
@@ -11,6 +11,16 @@ module PuppetObject
   def _ptype
     self.class._ptype
   end
+
+  def _pall_contents(path, &block)
```
hlindberg: Not too happy with the name as it becomes part of the API of every AST node. Maybe `_pcore_all_contents` ?

lib/puppet/pops/types/puppet_object.rb
```diff
@@ -11,6 +11,16 @@ module PuppetObject
   def _ptype
```
hlindberg: If the other `_p...` names change to `_pcore_...` then this should also change. i.e. `_pcore_type` instead of `_ptype`.
I know this is more painful - up till now the Pcore Objects have not been exposed anywhere where it is considered API, but now it does.

lib/puppet/pops/types/puppet_object.rb
```diff
@@ -11,6 +11,16 @@ module PuppetObject
...
     self.class._ptype
   end
+
+  def _pall_contents(path, &block)
+  end
+
+  def _pcontents
+  end
+
+  def i12n_hash
```
hlindberg: This is a nasty i14e name (incomprehensible). It is also part of every AST's API.

lib/puppet/pops/validation/checker4_0.rb
thallgren: Yes. The default value of the field is 'regular' and its type stipulates an Enum (not optional).

spec/unit/pops/evaluator/access_ops_spec.rb
thallgren: Because the expression needs a LiteralExpression, not a String. The Factory infers an expression, not a Type.

spec/unit/pops/evaluator/access_ops_spec.rb
thallgren: I'm out of ideas. The method `access` method can make assumptions that would be cumbersome to make while testing, but it is quicker to do so. So I added this method to do what `access` does, but with the extra creation of expressions etc. included. It's a method that does what `access` does, but for test purposes.

lib/puppet/pops/types/puppet_object.rb
thallgren: Changing that will break the netconf stuff. They use _ptype in several places. Using "_p" as a prefix works for me. My preference would be to keep it as is (including the other methods).

lib/puppet/pops/types/puppet_object.rb
thallgren: Yes, on an instance, this should be named "_pi12n_hash". The term "i12n_hash" is used throughout in the Pcore type system. It's rather late to change that now.

spec/unit/pops/evaluator/access_ops_spec.rb
hlindberg: we discussed an `access_at` as being different but more readable - the method really has no particular tie to "test" - it isn't testing anything. If anything it evaluates/performs/does an access operation - so `eval_access`, `eval_access_op`, `do_access`

lib/puppet/pops/model/ast.rb
```diff
@@ -4,19 +4,19 @@ module Puppet
 module Pops
 module Model
 class PopsObject
-  def self._plocation
-    @_plocation ||= (loc = Puppet::Util.path_to_uri("#{__FILE__}"); URI("#{loc}?line=#{__LINE__.to_i - 3}"))
+  def self._pcore_location
```
hlindberg: location is not really a pcore thing - it is part of what is modelled

hlindberg: unless we turn this into something generic, but then that generic thing is not "location" but some mechanism that provides a link back to document information in the root of the model instance

lib/puppet/pops/types/puppet_object.rb
hlindberg: this is now changed

lib/puppet/pops/types/puppet_object.rb
hlindberg: this is now changed

lib/puppet/pops/types/puppet_object.rb
hlindberg: this is now changed

lib/puppet/pops/model/ast.rb
```diff
@@ -0,0 +1,5169 @@
...
+  end
+
+  def eql?(o)
+    o.instance_of?(self.class)
+  end
+  alias == eql?
+end
+
+class Positioned < PopsObject
+  def self._pcore_location
```
hlindberg: This isn't really a pcore thing is it? We could generalize access to model root features, but then it would not be "location"

lib/puppet/pops/model/ast.rb
```diff
@@ -0,0 +1,5169 @@
...
+    @_pcore_location ||= (loc = Puppet::Util.path_to_uri("#{__FILE__}"); URI("#{loc}?line=#{__LINE__.to_i - 3}"))
```
hlindberg: Isn't this quite wasteful - one full `__FILE__` path URI for every class when it is really just `__LINE__` that is different?

lib/puppet/pops/validation/checker4_0.rb
hlindberg: changed later

lib/puppet/info_service/class_information_service.rb
```diff
@@ -105,7 +105,6 @@ def value_as_literal(value_expr)
 
   # Extracts the source for the expression
   def extract_value_source(value_expr)
-    position = Puppet::Pops::Adapters::SourcePosAdapter.adapt(value_expr)
-    position.extract_tree_text
+    value_expr.locator.extract_tree_text(value_expr)
   end
 end
```
hlindberg: fixed

lib/puppet/pops/types/puppet_object.rb
hlindberg: this is now changed

lib/puppet/pops/model/ast.rb
hlindberg: not sure how this ended up being a comment on an outdated change... entered it again on later variant

lib/puppet/pops/model/ast.rb
hlindberg: I mean, not all pcore objects come from a textual source, and if they do, then they need a mechanism like offset and length to be able to refer to parts of the text.
Also, if this is the location of the source, then this generated code should really reference the input source.

hlindberg:
In summary - this is coming along nicely. I don't really understand the purpose of the "location" stuff on the generated classes, and I still don't like the "test_access" method name.
Think we need to talk about "location" and "source" and "document" before deciding what to do.

lib/puppet/pops/model/ast.rb
thallgren: Not sure what you mean here. An URI is assigned to the @_pcore_location. Each such URI has to be a different URI because it contains the line. The @_pcore_location is what the loader stores as `origin` in the `NamedEntry` used for binding the type. There's no way to split that up unless we change the `NamedEntry`.

lib/puppet/pops/model/ast.rb
thallgren: I'm not so sure that appointing the input source is the right thing to do. It all depends on your use-case. If you're trying to find a problem, then it might be very helpful to see where the actual Ruby class is. After all, the loader doesn't load this type from the .pp file, so stating that is somewhat false.

thallgren:
Darn. I fixed the 'access_at' yesterday. Forgot to push it.

Regarding document, source, and location; I don't think that discussion should hold this PR back. It's more a discussion on how the Ruby generator works, and we should add tickets to it to improve it.

lib/puppet/pops/model/ast.rb
hlindberg: yeah, but the ruby implementation is being run by Ruby, and it knows where it loaded the code already - so what purpose does this duplicated information serve?
I understand that if they code was created dynamically, then it would be different - but then, would you want it to reference the ERB template, or the pcore source?

lib/puppet/pops/model/ast.rb
hlindberg: ah - ok - need to think about that, it is for the interface between puppet runtime loaders/type-system and pcore defined/loaded type

PR #5648, opened 2017-02-18
(PUP-7225) Remove the Puppet binder

thallgren:
This PR must be merged after [PR-5647](/puppetlabs/puppet/pull/5647). It contains the commits needed to remove the Puppet Injector/Binder and replace it with a much simpler mechanism for plugins that uses the Puppet context.


hlindberg:
While the change "Let the PP parser return the AST model instead of a factory" is the right thing to do, it modifies an API that others use and it will require updates in a couple of places; Cory Osman's "Puppet Debugger", (and most likely) "Puppet Strings". There is also documentation related to the Parser API that needs an update. I think these changes are containable, but it does require new releases and we want Puppet 5.0.0 to have as little churn as possible.

hlindberg:
ok, there are other AST changes that requires downstream to change (RGen Enums are Symbols vs. Pcore String values).

lib/puppet/pops/model/ast.pp
```diff
@@ -0,0 +1,687 @@
...
+        },
+        'char_offsets' => {
+          type => Boolean,
+          value => false
+        },
+        'locator' => Locator
+      }
+    }]
+  }
+}]
```
hlindberg: missing EOL

hlindberg:
The commit "Add Rake task for generating the AST model from Pcore" talks about "hen and egg" - in English it is "chicken and egg" :-)

lib/puppet/pops/adaptable.rb
```diff
@@ -146,19 +133,16 @@ def self.adapt_new(o, &block)
...
+      adapters = GLOBAL_ADAPTER_MAP[o.object_id]
+      adapters.nil? ? nil : adapters.delete(self)
     end
 
     # This base version creates an instance of the class (i.e. an instance of the concrete subclass
     # of Adapter). A Specialization may want to create an adapter instance specialized for the given target
     # object.
     # @param o [Adaptable] The object to adapt. This implementation ignores this variable, but a
     #   specialization may want to initialize itself differently depending on the object it is adapting.
+    #   It is imperative that the adapter does not bind this object! If it does, then that's a certain memory leak
```
hlindberg: Are we sure there are no such adapters? (I don't remember)

lib/puppet/plugins/configuration.rb
```diff
@@ -48,25 +42,15 @@ module Puppet::Plugins::Configuration
...
+      {
+        SyntaxCheckers::SYNTAX_CHECKERS_KEY => {
+          'json' => Puppet::SyntaxCheckers::Json.new,
+          'base64' => Puppet::SyntaxCheckers::Base64.new
+        }
+      }
     end
-end
+  end
+end
```
hlindberg: EOF without NL

lib/puppet/pops/model/model_meta.rb
```diff
@@ -100,7 +100,7 @@ class UnfoldExpression < UnaryExpression; end
   # An assignment expression assigns a value to the lval() of the left_expr.
   #
   class AssignmentExpression < BinaryExpression
-    has_attr 'operator', OpAssignment, :lowerBound => 1
+    has_attr 'operator', String, :lowerBound => 1
```
thallgren: Yes, in the Pcore model this is an Enum. See [ast.pp, line 83](https://github.com/puppetlabs/puppet/pull/5648/commits/0f0c8778e214ff5e5cc11fbee12a912c6addb5a0#diff-56bfa3148b52eaab040806043811b7bcR83)

lib/puppet/pops/adaptable.rb
thallgren: I've made sure that we don't.

lib/puppet/pops/model/ast.pp
thallgren: This comment is outdated. The EOL is present now.

PR #5650, opened 2017-02-20
(PUP-7102) Fail with error on access problems with environment directory

thallgren:
This PR ensures that access problems on a directory that is a
candidate to be an environment (reachable using the environmentpath)
or the parent of such a directory results in an error and that puppet
gives up.

### DIFF
--- a/lib/puppet/environments.rb
+++ b/lib/puppet/environments.rb
@@ -213,9 +213,34 @@ module Puppet::Environments
     end
 
     def valid_directory?(envdir)
-      name = Puppet::FileSystem.basename_string(envdir)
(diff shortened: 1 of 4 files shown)


hlindberg:
ok to merge when appveyor goes green

joshcooper:
@thallgren and I chatted IRL about this. If an ancestor directory has a permissions problem, the `Errno:EACESS` exception is raised and the exception message is returned to the client, including the path on disk. We should modify `indirected_routes.rb`, so that the call to `Puppet.lookup(:environments).get(environment)` rescues the exception and raises a generic HTTP 500 error.

PR #5652, opened 2017-02-21
(PUP-7235) Fix bug causing crash when using hiera.yaml glob pattern

thallgren:
This commit repairs the broken glob/globs implementation that was
introduced in hiera.yaml, version 5. Apparently there were no tests for
it, which of course was an oversight that had consequences. Tests are
also added in this commit.


hlindberg:

PR #5653, opened 2017-02-21
(PUP-7130) Unify ruby Integer types, ENV['HOME'] handling and HTTP time

thallgren:
This PR changes three things to make the Puppet code base execute without problems using a Ruby 2.4.0 runtime.

1. Changes all Fixnum references to Integer.
2. Fixes an ambiguous time representation in `HttpMetaData` uses UTC.
3. Conditions some tests that sets ENV['HOME'] to nil (no-op on 2.4).

### DIFF
--- a/lib/puppet/provider/augeas/augeas.rb
+++ b/lib/puppet/provider/augeas/augeas.rb
@@ -203,7 +203,7 @@ Puppet::Type.type(:augeas).provide(:augeas) do
 
   def is_numeric?(s)
     case s
-    when Fixnum
+    when Integer
       true
     when String
       s.match(/\A[+-]?\d+?(\.\d+)?\Z/n) == nil ? false : true
--- a/spec/unit/pops/types/type_calculator_spec.rb
+++ b/spec/unit/pops/types/type_calculator_spec.rb
@@ -1629,12 +1629,12 @@ describe 'The type calculator' do
       types_to_test.each {|t| expect(calculator.instance?(t::DEFAULT, :default)).to eq(false) }
     end
 
-    it 'should consider fixnum instanceof PIntegerType' do
+    it 'should consider integer instanceof PIntegerType' do
       expect(calculator.instance?(PIntegerType::DEFAULT, 1)).to eq(true)
     end
 
-    it 'should consider fixnum instanceof Fixnum' do
-      expect(calculator.instance?(Fixnum, 1)).to eq(true)
+    it 'should consider integer instanceof Integer' do
+      expect(calculator.instance?(Integer, 1)).to eq(true)
     end
 
     it 'should consider integer in range' do
@@ -1885,10 +1885,16 @@ describe 'The type calculator' do
   end
 
   context 'when converting a ruby class' do
-    it 'should yield \'PIntegerType\' for Integer, Fixnum, and Bignum' do
-      [Integer,Fixnum,Bignum].each do |c|
-        expect(calculator.type(c).class).to eq(PIntegerType)
-      end
+    it 'should yield \'PIntegerType\' for Fixnum' do
+      expect(calculator.type(Fixnum).class).to eq(PIntegerType)
+    end
+
+    it 'should yield \'PIntegerType\' for Bignum' do
+      expect(calculator.type(Bignum).class).to eq(PIntegerType)
+    end
+
+    it 'should yield \'PIntegerType\' for Integer' do
+      expect(calculator.type(Integer).class).to eq(PIntegerType)
     end
 
     it 'should yield \'PFloatType\' for Float' do
--- a/spec/unit/provider/user/directoryservice_spec.rb
+++ b/spec/unit/provider/user/directoryservice_spec.rb
@@ -157,7 +157,7 @@ describe Puppet::Type.type(:user).provider(:directoryservice) do
     "9377c46908a1c8ac2c3e45c0d44da8ad0fcd85ec5c14d9a59ffc40c9da31f0ec"
   end
 
-  # The below value is a Fixnum iterations value used in the PBKDF2
+  # The below value is an Integer iterations value used in the PBKDF2
   # key stretching algorithm
   let(:pbkdf2_iterations_value) do
     24752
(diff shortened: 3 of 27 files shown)


spec/unit/util/run_mode_spec.rb
```diff
@@ -2,6 +2,13 @@
 require 'spec_helper'
 
 describe Puppet::Util::RunMode do
+
+  # Discriminator for tests that attempts to unset HOME since that, for reasons currently unknown,
```
Iristyle: I'm not seeing a problem (with core Ruby `ENV`), at least on OSX / Ruby 2.4.0

```
irb(main):001:0> RUBY_VERSION
=> "2.4.0"
irb(main):002:0> ENV['HOME']
=> "/Users/Iristyle"
irb(main):003:0> ENV['HOME'] = nil
=> nil
irb(main):004:0> ENV['HOME']
=> nil

irb(main):009:0> ENV.keys.include?('HOME')
=> false
```

With our helpers:

```
[1] pry(main)> require 'puppet'
=> true
[2] pry(main)> Puppet::Util.get_env('HOME')
=> "/Users/Iristyle"
[3] pry(main)> Puppet::Util.set_env('HOME')
=> nil
[4] pry(main)> Puppet::Util.get_env('HOME')
=> nil
[5] pry(main)> RUBY_VERSION
=> "2.4.0"
[6] pry(main)> Puppet.version
=> "5.0.0"
```

Iristyle: Were there particular platforms you saw having a problem with unsetting `HOME` @thallgren ?

lib/puppet/provider/augeas/augeas.rb
```diff
@@ -203,7 +203,7 @@ def close_augeas
 
   def is_numeric?(s)
     case s
-    when Fixnum
+    when Integer,Float
```
hlindberg: This is not the same - augeas may not handle floats well.

spec/unit/pops/types/type_calculator_spec.rb
```diff
@@ -1885,10 +1885,8 @@ def foo(a)
   end
 
   context 'when converting a ruby class' do
-    it 'should yield \'PIntegerType\' for Integer, Fixnum, and Bignum' do
-      [Integer,Fixnum,Bignum].each do |c|
-        expect(calculator.type(c).class).to eq(PIntegerType)
-      end
+    it 'should yield \'PIntegerType\' for Integer' do
+      expect(calculator.type(Integer).class).to eq(PIntegerType)
```
hlindberg: The original test needs to be retained for < Ruby 2.4 as we otherwise remove the assertion of how Fixnum/Bignum are mapped to an Integer PType. Make the test depend on Ruby version, and add new test for >= 2.4.

spec/unit/provider/user/directoryservice_spec.rb
```diff
@@ -799,8 +799,8 @@ module Puppet::Util::Plist
     it "should accept a hash containing a PBKDF2 password hash, salt, and iterations value and return the correct iterations value" do
         expect(provider.class.get_salted_sha512_pbkdf2('iterations', pbkdf2_embedded_bplist_hash)).to eq(pbkdf2_iterations_value)
     end
-    it "should return a Fixnum value when looking up the PBKDF2 iterations value" do
-        expect(provider.class.get_salted_sha512_pbkdf2('iterations', pbkdf2_embedded_bplist_hash)).to be_a_kind_of(Fixnum)
+    it "should return a Integer value when looking up the PBKDF2 iterations value" do
```
hlindberg: "return an Integer"

spec/unit/util/run_mode_spec.rb
thallgren: That's because the tests you do doesn't reveal the problem. Try
```
File.expand_path('~/tmp')
```
and you will discover that it expands the path even when you have set the ENV['HOME'] to nil. Prior to 2.4.0, you'd get an exception and that exception is what the disabled tests rely on.

lib/puppet/provider/augeas/augeas.rb
thallgren: Take a look at the regular expression used for the String a couple of lines down. It accepts floats. Seems rather odd that the string form would, but the numeric form would not. One of them should be adjusted.

thallgren:
@hlindberg, regarding the change of HOME. I think the title of this ticket should change instead. To "Make it possible to run Puppet with Ruby 2.4". It makes no sense to me having separate tickets for what needs to be fixed since they wouldn't produce a functional code-base.

spec/unit/pops/types/type_calculator_spec.rb
thallgren: Isn't that a bit of an overkill? After all, both `Fixnum` and `Bignum` inherits from `Integer`.

lib/puppet/provider/augeas/augeas.rb
hlindberg: While I agree that it looks like a bug in the original, this is not a change that is required to make puppet run on Ruby 2.4.0. After this change, more values are allowed than earlier (floats in float form).

hlindberg: If also attempting to fix the potential bug, I would like that as a separate commit with rationale etc.

hlindberg:
It would be much better if the changes were split up into several commits. You list them as "1,2,3" already - plus add one for the possible bug-fixed auegas.

spec/unit/util/run_mode_spec.rb
Iristyle: I would have preferred we understand / solve this problem rather than papering over the test failures with exclusions.

Iristyle: FWIW - it looks like the change was made in https://github.com/ruby/ruby/commit/6b88dd2698760ef2e2306e3e79a0ed9fe159a12c 

The associated Ruby ticket is https://bugs.ruby-lang.org/issues/12695

thallgren: @Iristyle, Great job finding the actual issue. I tried but didn't succeed. I agree that it's a good thing to understand the problem. My intention was not to paper over it, but rather that the current tests could be made into acceptance tests instead. I suggest that in the commit comment.

PR #5657, opened 2017-02-22
(PUP-7245) Make configuration_interpolation test self-contained

branan:
Previously, this test would clobber the default hiera.yaml that other
tests depended on. It now backs up whatever hiera.yaml it finds, and
restores that file in a teardown.


acceptance/tests/lookup/configuration_interpolation.rb
```diff
@@ -5,19 +5,21 @@
   app_type        = File.basename(__FILE__, '.*')
   tmp_environment = mk_tmp_environment_with_teardown(master, app_type)
   fq_tmp_environmentpath  = "#{environmentpath}/#{tmp_environment}"
+  master_confdir = master.puppet('master')['confdir']
+
+  step "backup hiera config file" do
+    on master, "cp #{master_confdir}/hiera.yaml #{master_confdir}/hiera.bak}"
```
joshcooper: typo `hiera.bak}`

branan: ... why didn't that fial

branan: Now I have no confidence in any of this :/

acceptance/tests/lookup/configuration_interpolation.rb
```diff
@@ -46,7 +48,7 @@ class some_mod {
       SITE
 
       on(master, "chmod -R 775 #{fq_tmp_environmentpath}")
-      on(master, "chmod -R 775 #{confdir}")
+      on(master, "chmod -R 775 #{master_confdir}")
```
joshcooper: I know this test was written this way, but I don't like that it makes changes to the real `confdir`. Usually we create a temp confdir and point the master at that.

branan: I agree, but I was trying for minimal changes to make it work

acceptance/tests/lookup/configuration_interpolation.rb
branan: Oh, my restore checks for an existing backup (duh). Not sure that actually makes sense since the first step is to make that backup. How do you feel about making the restore unconditional (since if the backup broke, further tests will likely fail)

branan:
@joshcooper typo fixed, and I'm gonna take your thumb as agreement on the "minimal changes" route for this PR

acceptance/tests/lookup/configuration_interpolation.rb
kurtwall: :+1: for @joshcooper's comment, but can that be fixed in a separate PR?

johnduarte:
Completed full test run using SHA ac2c0d18fbe5ba05a95c74f669a3ae39a024300b on Ubuntu 16.04.
There were two unrelated failures in this run. All other tests passed.
* tests/language/binary_data_type.rb
* tests/resource/file/ticket_6448_file_with_utf8_source.rb

PR #5659, opened 2017-02-22
(PUP-7243) Fixes CI blocker

seangriff:
The previous version of this test created a temp directory for the
test environment.  This directory had the default permission of 0700
which causes Jenkins to fail.  This version changes the permissions
of that directory to 0777.

### DIFF
--- a/acceptance/tests/utf8/utf8-in-function-args.rb
+++ b/acceptance/tests/utf8/utf8-in-function-args.rb
@@ -4,13 +4,14 @@ test_name 'utf-8 characters in function parameters' do
     'eos-4',        # PUP-7146
     'cumulus',      # PUP-7147
     'cisco',        # PUP-7150
+    'aix',          # PUP-7194
+    'huawei',       # PUP-7195
   ]
-  confine :except, :platform => /^aix/           # PUP-7190
-  confine :except, :platform => /^huawei/
 
   # utf8chars = "€‰ㄘ万竹ÜÖ"
   utf8chars = "\u20ac\u2030\u3118\u4e07\u7af9\u00dc\u00d6"
   master_lookup_test_dir = master.tmpdir("lookup_test_dir")
+  on(master, "chmod -R 777 #{master_lookup_test_dir}")
   master_opts = {
     'main' => {
       'hiera_config' => "#{master_lookup_test_dir}/hiera.yaml",
@@ -327,7 +328,7 @@ LOOKUP_MANIFEST
           :environment => {:LANG => "en_US.UTF-8"}
         )
         assert_match(
-          /[[€‰ㄘ万竹ÜÖ], [€‰ㄘ万竹ÜÖ], , ]/,
+          /[[#{utf8chars}], [#{utf8chars}], , ]/,
           result.stdout,
           "match() result unexpected"
         )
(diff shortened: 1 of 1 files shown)


joshcooper:
Should this target stable? It looks like the test in stable has the same issue, but I thought it was passing CI there?

seangriff:
@joshcooper 
The stable pipeline is passing, just checked.  I don't know why.  Do they run as the same user or could stable be running as root?

Should I direct this PR to stable?  Happy to do so.

joshcooper:
Ah, ok, so the issue was introduced in https://tickets.puppetlabs.com/browse/PUP-7102, merged to master only, so I think it's ok to target master as well.

Previously, puppet was not sensitive to permissions issues for parent directories in the environmentpath, e.g. `/etc/puppetlabs`. So puppet would treat that situation as "the environment doesn't exist". The test doesn't fail in stable, because the test always runs `puppet apply` as `root`, even on the `master` host.

In puppet#master, the permissions issue causes puppetserver to fail to start:

```
06:01:52     Feb 21 06:01:21 rbjkogu91d1gagw systemd[1]: Starting puppetserver Service...
06:01:52     Feb 21 06:01:21 rbjkogu91d1gagw puppetserver[13389]: OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
06:01:52     Feb 21 06:01:50 rbjkogu91d1gagw puppetserver[13389]: Errno::EACCES: Permission denied - /tmp/lookup_test_dir.hZDf2k
06:01:52     Feb 21 06:01:50 rbjkogu91d1gagw puppetserver[13389]: valid_directory? at /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/environments.rb:242
```

Note we also create a [tmpdir for agents](https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/utf8/utf8-in-function-args.rb#L73), but it isn't an issue because agents run as `root`.

That said, this test only ever uses `puppet apply`, but not `puppet agent`. So we can probably drop the `with_master_running_on` bit.

acceptance/tests/utf8/utf8-in-function-args.rb
```diff
@@ -4,93 +4,93 @@
...
   # utf8chars = "€‰ㄘ万竹ÜÖ"
   utf8chars = "\u20ac\u2030\u3118\u4e07\u7af9\u00dc\u00d6"
   master_lookup_test_dir = master.tmpdir("lookup_test_dir")
+  on(master, "chmod -R 777 #{master_lookup_test_dir}")
   master_opts = {
     'main' => {
       'hiera_config' => "#{master_lookup_test_dir}/hiera.yaml",
       'environmentpath' => "#{master_lookup_test_dir}/environments",
     },
   }
```
joshcooper: I think we can drop lines 13-20, since we're not starting a master.

acceptance/tests/utf8/utf8-in-function-args.rb
```diff
@@ -4,93 +4,93 @@
...
+        puppet("apply", "-e", puppet_cmd),
+        :environment => {:LANG => "en_US.UTF-8"}
+      )
+      assert_match(/#{utf8chars}/, result.stdout, "filter() failed.")
+    end
+  
+    agent_lookup_test_dir =
+      (agent == master) ?
+      master_lookup_test_dir :
+      agent_lookup_test_dir = agent.tmpdir("lookup_test_dir")
```
joshcooper: This can be simplified to just `agent_lookup_test_dir = agent.tmpdir("lookup_test_dir")`

joshcooper:
I went ahead and updated the test so we can get CI green.

PR #5660, opened 2017-02-24
(PUP-7251) decompress gzip as binary

MosesMendoza:
When puppet server recently correctly started sending gzip compressed content
(TK-429) we saw regressions in puppet's handling of UTF-8 characters in the
catalog, specificaly the test ticket_6448_file_with_utf8_source.rb. It turned
out that after decompression in Puppet::Network::Http::Compression, the catalog
had mangled UTF-8 bytes. This is because Zlib::GzipReader will decompress
content as whatever Encoding.default_external is. PSON (currently default
catalog content type) is BINARY encoded per specification
https://docs.puppet.com/puppet/latest/http_api/pson.html#differences-from-json.
Zlib::GzipReader will accept an encoding option just like an IO object, and
thus this commit updates the decompression to Encoding::BINARY.

This is confirmed by analyzing the body returned when puppet server's webserver
is run with decompression disabled. In this case we fall through to the final
statement in the compression determination, here: which simply returns
response.body. With compression disabled, the encoding of response.body is
binary.

Note that a more complete fix will actually specify the charset parameter of
the Content-Type header, i.e.:

Content-Type: text/json; charset=utf-8

That fix is tracked via PUP-7261.

### DIFF
--- a/lib/puppet/network/http/compression.rb
+++ b/lib/puppet/network/http/compression.rb
@@ -20,7 +20,8 @@ module Puppet::Network::HTTP::Compression
     def uncompress_body(response)
       case response['content-encoding']
       when 'gzip'
-        return Zlib::GzipReader.new(StringIO.new(response.body)).read
+        # ZLib::GzipReader has an associated encoding, by default Encoding.default_external
+        return Zlib::GzipReader.new(StringIO.new(response.body), :encoding => Encoding::BINARY).read
       when 'deflate'
         return Zlib::Inflate.new.inflate(response.body)
       when nil, 'identity'
--- a/spec/unit/network/http/compression_spec.rb
+++ b/spec/unit/network/http/compression_spec.rb
@@ -86,17 +86,48 @@ describe "http compression" do
         expect(@uncompressor.uncompress_body(@response)).to eq("uncompresseddata")
       end
 
-      it "should use a GzipReader with 'gzip' content-encoding" do
-        @response.stubs(:[]).with('content-encoding').returns('gzip')
+      context "with 'gzip' content-encoding" do
+        it "should use a GzipReader" do
+          @response.stubs(:[]).with('content-encoding').returns('gzip')
 
-        io = stub 'io'
-        StringIO.expects(:new).with("mydata").returns io
+          io = stub 'io'
+          StringIO.expects(:new).with("mydata").returns io
 
-        reader = stub 'gzip reader'
-        Zlib::GzipReader.expects(:new).with(io).returns(reader)
-        reader.expects(:read).returns "uncompresseddata"
+          reader = stub 'gzip reader'
+          Zlib::GzipReader.expects(:new).with(io, :encoding => Encoding::BINARY).returns(reader)
+          reader.expects(:read).returns "uncompresseddata"
 
-        expect(@uncompressor.uncompress_body(@response)).to eq("uncompresseddata")
+          expect(@uncompressor.uncompress_body(@response)).to eq("uncompresseddata")
+        end
+
+        it "should correctly decompress PSON containing UTF-8 in Binary Encoding" do
+          # Simulate a compressed response body containing PSON containing UTF-8
+          # using different UTF-8 widths:
+
+          # \u06ff - ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
+          # \u16A0 - ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
+          # \u{2070E} - 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
+
+          pson = "foo\u06ff\u16A0\u{2070E}".to_pson # unicode expression eqivalent of "foo\xDB\xBF\xE1\x9A\xA0\xF0\xA0\x9C\x8E\" per above
+          writer = Zlib::GzipWriter.new(StringIO.new)
+          writer.write(pson)
+          compressed_body = writer.close.string
+
+          begin
+            default_external = Encoding.default_external
+            Encoding.default_external = Encoding::ISO_8859_1
+
+            @response.stubs(:[]).with('content-encoding').returns('gzip')
+            @response.stubs(:body).returns(compressed_body)
+
+            uncompressed = @uncompressor.uncompress_body(@response)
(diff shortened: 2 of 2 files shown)


lib/puppet/network/http/compression.rb
```diff
@@ -20,7 +20,8 @@ module Active
     def uncompress_body(response)
       case response['content-encoding']
       when 'gzip'
-        return Zlib::GzipReader.new(StringIO.new(response.body)).read
+        # ZLib::GzipReader has an associated encoding, by default Encoding.default_external
+        return Zlib::GzipReader.new(StringIO.new(response.body), :encoding => Encoding::BINARY).read
```
Iristyle: I know that it's superfluous, but what do you think about adding in the explicit encoding to the `StringIO` initializer like:

```ruby
return Zlib::GzipReader.new(StringIO.new(response.body, 'rb'), :encoding => Encoding::BINARY).read
``` 

A couple reasons:

* If we do find an analysis tool to scan for missing usage of explicit encodings (RubyMine inspections might be our best bet - need to look into it more), then this would be one less thing to worry about.
* If for some reason `response.body` changes from a `BINARY` string, this could all of a sudden start failing hard - which would be a clear signal of a behavior change.

spec/unit/network/http/compression_spec.rb
```diff
@@ -86,17 +86,48 @@ class HttpUncompressor
...
+          # \u16A0 - ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
+          # \u{2070E} - 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
+
+          pson = "foo\u06ff\u16A0\u{2070E}".to_pson # unicode expression eqivalent of "foo\xDB\xBF\xE1\x9A\xA0\xF0\xA0\x9C\x8E\" per above
+          writer = Zlib::GzipWriter.new(StringIO.new)
+          writer.write(pson)
+          compressed_body = writer.close.string
+
+          begin
+            default_external = Encoding.default_external
```
Iristyle: I assume things totally exploded if you instead used?

```ruby
Encoding.stubs(:default_external).returns(Encoding::ISO_8859_1)
```

That should remove the need for the `begin` / `ensure` blocks.

lib/puppet/network/http/compression.rb
Iristyle: Totally tangential, but FYI - I just checked RubyMine 2016.3, and out of the box there are no code inspections for not having explicitly specified an encoding.

That said, it looks like you can build custom inspections for IntelliJ based IDEs, so there's some hope we can do something easily - https://www.jetbrains.com/help/idea/2016.3/creating-custom-inspections.html.

MosesMendoza: 👍

MosesMendoza: Actually, for `StringIO.new`, it doesn't appear that specifying an encoding in the constructor doesn't do anything? It appears it will just take on the encoding of the string (first argument)?:
```ruby
[26] pry(main)> binary = "foo".force_encoding(Encoding::BINARY)
=> "foo"
[27] pry(main)> io = StringIO.new(binary, 'rb')
=> #<StringIO:0x007f9bfbadb800>
[28] pry(main)> io.external_encoding
=> #<Encoding:ASCII-8BIT>
[29] pry(main)> read = io.read
=> "foo"
[30] pry(main)> read.encoding
=> #<Encoding:ASCII-8BIT>
[31] pry(main)> utf8 = "foo".force_encoding(Encoding::UTF_8)
=> "foo"
[32] pry(main)> io = StringIO.new(utf8, 'rb')
=> #<StringIO:0x007f9bfb9985b0>
[33] pry(main)> io.external_encoding
=> #<Encoding:UTF-8>
[34] pry(main)> read = io.read
=> "foo"
[35] pry(main)> read.encoding
=> #<Encoding:UTF-8>
```

spec/unit/network/http/compression_spec.rb
MosesMendoza: I don't thinking stubbing Encoding.default_external actually accomplishes what we want. While we say that Zlib::GzipReader will decompress as Encoding.default_external by default, it doesn't actually call Encoding.default_external, but rather whatever ruby uses under the hood to identify it.

I.e., session first with Encoding.default_external stubbed as ISO_8859_1, and then with Encoding.default_external actually set:
```rspec
[1] pry(#<RSpec::ExampleGroups::HttpCompression::WhenZlibIsAvailable::WhenUncompressingResponseBody::WithGzipContentEncoding>)> Encoding.default_external
=> #<Encoding:ISO-8859-1 (autoload)>
[2] pry(#<RSpec::ExampleGroups::HttpCompression::WhenZlibIsAvailable::WhenUncompressingResponseBody::WithGzipContentEncoding>)> r = Zlib::GzipReader.new(StringIO.new(@response.body)).read
=> "\"fooۿᚠ𠜎\""
[3] pry(#<RSpec::ExampleGroups::HttpCompression::WhenZlibIsAvailable::WhenUncompressingResponseBody::WithGzipContentEncoding>)> r.encoding
[4] pry(#<RSpec::ExampleGroups::HttpCompression::WhenZlibIsAvailable::WhenUncompressingResponseBody::WithGzipContentEncoding>)> Encoding.default_external = Encoding::ISO_8859_1
[4] pry(#<RSpec::ExampleGroups::HttpCompression::WhenZlibIsAvailable::WhenUncompressingResponseBody::WithGzipContentEncoding>)> Encoding.default_external = Encoding::ISO_8859_1
=> #<Encoding:ISO-8859-1>
[5] pry(#<RSpec::ExampleGroups::HttpCompression::WhenZlibIsAvailable::WhenUncompressingResponseBody::WithGzipContentEncoding>)> r = Zlib::GzipReader.new(StringIO.new(@response.body)).read
=> "\"fooÛ¿á\x9A ð \x9C\x8E\""
[6] pry(#<RSpec::ExampleGroups::HttpCompression::WhenZlibIsAvailable::WhenUncompressingResponseBody::WithGzipContentEncoding>)> r.encoding
=> #<Encoding:ISO-8859-1>
```

Iristyle: Ah yes, that's correct - thanks for the reminder!  I recall @glennsarti and I running into that in the past.

branan: `ENV` works the same way. It's lots of fun.

lib/puppet/network/http/compression.rb
Iristyle: Yes, you're correct.

PR #5662, opened 2017-02-24
(PUP-7252) Stop puppet service with helper

johnduarte:
This commit updated the `puppet_mcollective_service_management`
test to use the beaker helper `stop_agent` to set up the precondition
of ensuring that the `puppet` service is not running.

Prior to this commit, attempting to stop the `puppet` service with a
call to `puppet apply` would intermittently fail due to an active
puppet run on the agent. This caused the `puppet apply` attempt to
fail.

### DIFF
--- a/acceptance/tests/resource/service/puppet_mcollective_service_management.rb
+++ b/acceptance/tests/resource/service/puppet_mcollective_service_management.rb
@@ -49,7 +49,13 @@ agents.each do |agent|
       assert_service_status_on_host(agent, service, {'ensure' => 'running'}) # Status should always be 'running' after starting
 
       step "Stopping the #{service} service while it is #{status}: it should be stopped"
-      set_service_initial_status(agent, service, status)
+      if service == 'puppet' and status == 'stopped' then
+        # using the beaker helper to set the initial state of puppet to
+        # prevent coliding with a puppet run in progress on the SUT (PUP-7252)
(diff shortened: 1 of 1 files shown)


Magisus:
Looking at that beaker helper, it definitely seems much more robust for handling special cases around the agent running 👍

Magisus:
@MikaelSmith is there a reason to prefer your alternative solution? (I don't have a lot of context on the pros and cons here)

johnduarte:
I have run this change 3 times sequentially against an OS X 10.10 system that was previously consistently failing on this test. All iterations passed.
```
tests/resource/service/puppet_mcollective_service_management.rb passed in 586.80 seconds
```

MikaelSmith:
The test was confirming that stopping a service specifically via `puppet apply` works. Since it's still being checked for mcollective, this is probably fine, just slightly circumvents the intention of the test.

johnduarte:
The confirmation that `puppet apply` can stop the `puppet` service is still tested on line 36 of this test.
This change should only impact the per-condition for 'stopping service while it is stopped' test case.

MikaelSmith:
Ah right. Then we probably could've used the beaker helper for set_service_initial_status everywhere.

joshcooper:
Does this need to be backported?

Magisus:
No, the change that caused the bug landed on master.

PR #5666, opened 2017-02-27
(PUP-7204) Add support for mapped paths to hiera.yaml

thallgren:
This commit adds support for a new hierarchy entry named `mapped_paths`.
It must be a three element array of strings where the first string
denotes a scope variable that should point to a collection of strings,
the second string is the variable name that will be mapped to each
element of the collection, and the third string is a template where that
variable can be used in interpolation expressions.

Paths are produced by iterating over the collection, mapping each entry
to the variable name in a local scope, and then resolve the template
using the local scope.

### DIFF
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -901,6 +901,19 @@ class Puppet::Parser::Scope
     find_global_scope.without_ephemeral_scopes(&block)
   end
 
+  # Execute given block with a ephemeral scope containing the given variables
+  # @api private
+  def with_local_scope(scope_variables)
+    local = LocalScope.new(@ephemeral.last)
+    scope_variables.each_pair { |k, v| local[k] = v }
+    @ephemeral.push(local)
+    begin
+      yield(self)
+    ensure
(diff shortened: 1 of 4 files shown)


hlindberg:
I think the tests needs to show that it is possible to also interpolate other variables (just like for any non mapped `path` entry), and that a choice of local var name that shadows wins over an outer scope unless the variable is interpolated with absolute top scope anchored name (e.g. `::foo`). Those tests show that the implemented scoping handling works as intended.

hlindberg:
I would also like to see a test that shows how a mapped_paths result is explained (how it arrived at the resulting path value fed to the function, or one that does not exist etc).

hlindberg:
@thallgren this needs a rebase - and since we are going to be doing a 4.10.0 release, this should target 4.10.x branch (I am going to change the ticket now).

hlindberg:
Still (or again) needs a rebase

thallgren:
Re-re-rebased.

hlindberg:
good to merge when tests go green (AppVeyor seems to be extremely slow today).

PR #5670, opened 2017-03-01
(PUP-7240) Prevent invocation inheritance when lookup adapters differs

thallgren:
The logic that controls how a lookup invocation is inherited did not
take into account that a completely different lookup instance could
execute a lookup in a nested fashion. This made a lookup for the special
`lookup_options` key fail for custom adapters.

This commit ensures that the inheritance of adapter class is assigned
to the class of the parent when no explicit class is given. It also
ensures that no inheritance takes place when the adapter classes differ.


thallgren:
This PR has been cherry-picked to branch 4.10.x

PR #5671, opened 2017-03-01
(PUP-7285) Add metric_id http option to http report processor

rlinehan:
Add a `metric_id` http option to the http request made by the http report
processor to report urls. This will allow Puppet Server to track the amount of
time it spends making these requests. The value for this metric id is
`[:puppet :report :http]`.

### DIFF
--- a/lib/puppet/reports/http.rb
+++ b/lib/puppet/reports/http.rb
@@ -14,7 +14,11 @@ Puppet::Reports.register_report(:http) do
   def process
     url = URI.parse(Puppet[:reporturl])
     headers = { "Content-Type" => "application/x-yaml" }
-    options = {}
+    # This metric_id option is silently ignored by Puppet's http client
+    # (Puppet::Network::HTTP) but is used by Puppet Server's http client
+    # (Puppet::Server::HttpClient) to track metrics on the request made to the
+    # `reporturl` to store a report.
+    options = { :metric_id => [:puppet, :report, :http] }
     if url.user && url.password
       options[:basic_auth] = {
         :user => url.user,
--- a/lib/puppet/network/http/connection.rb
+++ b/lib/puppet/network/http/connection.rb
@@ -62,9 +62,13 @@ module Puppet::Network::HTTP
     end
 
     # @!macro [new] common_options
-    #   @param options [Hash] options influencing the request made
+    #   @param options [Hash] options influencing the request made. Any
+    #   options not recognized by this class will be ignored - no error will
+    #   be thrown.
     #   @option options [Hash{Symbol => String}] :basic_auth The basic auth
-    #     :username and :password to use for the request
+    #     :username and :password to use for the request, :metric_id Ignored
+    #     by this class - used by Puppet Server only. The metric id by which
+    #     to track metrics on requests.
 
     # @param path [String]
     # @param headers [Hash{String => String}]
--- a/spec/unit/reports/http_spec.rb
+++ b/spec/unit/reports/http_spec.rb
@@ -39,6 +39,7 @@ describe processor do
   describe "when making a request" do
     let(:connection) { stub_everything "connection" }
     let(:httpok) { Net::HTTPOK.new('1.1', 200, '') }
+    let(:options) { {:metric_id => [:puppet, :report, :http]} }
 
     before :each do
       Puppet::Network::HttpPool.expects(:http_instance).returns(connection)
@@ -46,7 +47,7 @@ describe processor do
 
     it "should use the path specified by the 'reporturl' setting" do
       report_path = URI.parse(Puppet[:reporturl]).path
-      connection.expects(:post).with(report_path, anything, anything, {}).returns(httpok)
+      connection.expects(:post).with(report_path, anything, anything, options).returns(httpok)
 
       subject.process
     end
@@ -54,22 +55,24 @@ describe processor do
     it "should use the username and password specified by the 'reporturl' setting" do
       Puppet[:reporturl] = "https://user:pass@myhost.mydomain:1234/report/upload"
 
-      connection.expects(:post).with(anything, anything, anything, :basic_auth => {
-        :user => 'user',
-        :password => 'pass'
-      }).returns(httpok)
+      connection.expects(:post).with(anything, anything, anything,
+                                     {:metric_id => [:puppet, :report, :http],
+                                      :basic_auth => {
+                                        :user => 'user',
+                                        :password => 'pass'
+                                      }}).returns(httpok)
 
       subject.process
     end
 
     it "should give the body as the report as YAML" do
-      connection.expects(:post).with(anything, subject.to_yaml, anything, {}).returns(httpok)
+      connection.expects(:post).with(anything, subject.to_yaml, anything, options).returns(httpok)
 
       subject.process
     end
 
     it "should set content-type to 'application/x-yaml'" do
-      connection.expects(:post).with(anything, anything, has_entry("Content-Type" => "application/x-yaml"), {}).returns(httpok)
+      connection.expects(:post).with(anything, anything, has_entry("Content-Type" => "application/x-yaml"), options).returns(httpok)
 
       subject.process
     end


rlinehan:
looks like the appveyor `Environment: LOCALE=ja-JP; Platform: Ruby21-x64` job timed out. I'm not quite sure how to rerun it though.

rlinehan:
Realized a potential issue, please hold off on merging.

rlinehan:
Alright, so this is only going to work with (what will be) Puppet Server 5 - it needs this change: https://github.com/puppetlabs/puppetserver/commit/06ac9471a149ca1d26b7da2a823bf52f7026af8d#diff-5805ba6eb4857e236651f3c3fcdffeb0R75 which allows `get` to be called with 3 arguments (matching what Puppet has had in its http client for a long time).

I don't know if given version and compatibility things between puppet and puppet server, if we can say that this terminus will *only* be run under Puppet Server 5. If we can, then that's great, except that this will cause tests to fail because I think they're pinned to Puppet Server 2.7.2 (last released).

I can also do a similar terrible hack here as I did in PuppetDB: https://github.com/puppetlabs/puppetdb/pull/2179/commits/ad0905ab1837db8a0cec5dd7daaef322213e94c8#diff-b43043c3acd01a66d0886dd6f3b01d71R234. This is terrible, but it also means that we don't have to make some terrible compatibility requirements between PuppetDB and Puppet Server, which ultimately benefits our users more.

MosesMendoza:
@rlinehan I think we're trying pretty hard to avoid a change that would require server 5 with agent 5

camlow325:
> Alright, so this is only going to work with (what will be) Puppet Server 5 - it needs this change: puppetlabs/puppetserver@06ac947#diff-5805ba6eb4857e236651f3c3fcdffeb0R75 which allows get to be called with 3 arguments (matching what Puppet has had in its http client for a long time).

@rlinehan For this one, don't we only need to instrument the `:metric-id` option through the `post` method on the http client, not the `get` method.  We should be good already there for Puppet Server since we've been supporting that from the 1.x series on - including the version that Puppet core is currently being tested with, [2.7.2](https://github.com/puppetlabs/puppetserver/blob/2.7.2/src/ruby/puppetserver-lib/puppet/server/http_client.rb#L53).

lib/puppet/reports/http.rb
```diff
@@ -14,7 +14,7 @@
   def process
     url = URI.parse(Puppet[:reporturl])
     headers = { "Content-Type" => "application/x-yaml" }
-    options = {}
+    options = { :metric_id => [:puppet, :report, :http] }
```
MosesMendoza: @rlinehan based on my reading of https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/http/connection.rb#L25-L54 I think we may need to extend the `Puppet::Network::HTTP::Connection` class to be aware of the `:metric_id` option for this to not fail. I may certainly be missing something though..

MosesMendoza: nevermind, I'm wrong. that validation is only done when creating a new connection instance, which is already done for us via the http pool

MosesMendoza:
@rlinehan just so I understand the context, is this only intended for use with a (our) custom report processor? I.e., we don't also want to send `:metric_id` along by default whenever an agent sends its report to the server?

lib/puppet/reports/http.rb
MosesMendoza: Ok I'm confused again :) It looks like at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/http/connection.rb#L199-L205 the only option we propagate is `basic_auth` - nothing else is retained as the connection is executed, https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/http/connection.rb#L172. Does this use different connection logic somewhere? I'm afraid I'm not super familiar with the puppet server ruby http client and where it intersects with the puppet agent http client code.

camlow325:
@MosesMendoza The `:metric-id` we'd be adding here is something that Puppet Server's metrics implementation would be reading to track metrics for HTTP client requests made by the server itself.  While we could add this option for the agent report POST, I don't know of anything that would use it since that code wouldn't be running within Puppet Server.

lib/puppet/reports/http.rb
camlow325: Yeah, it's fine that the `:metric-id` won't be used by the `http_pool` connection in Ruby Puppet code.  As long as its presence in the `get` or `post` options doesn't cause any problems, should be fine.  Where this will actually be used is in Puppet Server's http connection class - an alternate implementation registered only when running under Puppet Server.  See [this code](https://github.com/puppetlabs/puppetserver/blob/bb21b3824532d4c52295c348e8caf231ee584d76/src/ruby/puppetserver-lib/puppet/server/http_client.rb#L98-L100) for reference.

MosesMendoza: Got it. Thanks for clarifying @camlow325! In that case, yeah I think this key will basically get silently dropped in the current implementation, which seems just fine given your explanation.

Thought it might be risky to implicitly rely on `:metric_id` getting dropped, since `Puppet::Network::HTTP::Connection#post` is a bit removed from here. I'm not sure what the implications would be if we suddenly started including `:metric_id` in all report posts on accident, but we probably don't want to? 

It's not the cleanest thing but what do you think about checking if our `Puppet::Network::HttpPool::connection_class` is `Puppet::Server::HttpClient` instead of `Puppet::Network::Http::Connection` (or some other way of detecting that we're in the right environment) and only then adding `:metric_id` to the options? Do we do that kind of check anywhere else? Failing that, could you add a code comment with some explanation/warning that `:metric_id` is only used by the Puppet Server http client and is ignored otherwise?

camlow325: > I'm not sure what the implications would be if we suddenly started including :metric_id in all report posts on accident, but we probably don't want to?

We shouldn't ever end up with the case where elements passed in on the options hash are opaquely translated into content in the request, though, should we - e.g., as a query param on the URL or multipart form data?  My understanding has been that the options hash is a mechanism for passing through settings which must be explicitly read/acted upon by the connection class.

> It's not the cleanest thing but what do you think about checking if our Puppet::Network::HttpPool::connection_class is Puppet::Server::HttpClient instead of Puppet::Network::Http::Connection (or some other way of detecting that we're in the right environment) and only then adding :metric_id to the options?

That might be appropriate for the code change on this PR.  It's worth keeping in mind, though, that this client is used for various extension points whose source code lives outside of the core Ruby Puppet repo - e.g., in the PuppetDB and Node Classifier terminii.  We're adding the use of the `:metric-id` to those terminii.  Non-Puppet owned extensions may be using this as well.  Embedding knowledge about the specific implementation class used in code that lives outside of the Puppet repo would seem dangerous - e.g., in the event that a future refactor were to change the name of the default implementation class which gets used.  Also, while this code wouldn't add anything which would use the `:metric-id` when running outside of Puppet Server, I could envision a time in the future where having the `:metric-id` acted upon might be useful.  For example, in the `Puppet::Network::HTTP::Connection`, we might want to wrap a Puppet profiler block around the HTTP get/post calls and use the supplied `:metric-id` as an identifier for those.  This could be nice for profiling http client calls made by the agent - in a way that can work fairly generically for extensions which live outside of the Ruby Puppet source.

I agree that it does seem a little distant to have an option set which is only acted on via code which lives in another repo.  I wonder if a decent compromise may be to expand the [option](https://github.com/puppetlabs/puppet/blob/4.9.3/lib/puppet/network/http/connection.rb#L66-L67) comments in the connection.rb class to mention the `:metric-id` as a possible option but that it is not actually used by the `Puppet::Network::Http::Connection` implementation.  We could also mention in these comments that any options not recognized by the `Connection` class will be ignored, that no error will be thrown.  We could also add a comment here and where used in the reports.rb code that that the `:metric-id` option is being added for Puppet Server's use.

@MosesMendoza - WDYT?

MosesMendoza: > We shouldn't ever end up with the case where elements passed in on the options hash are opaquely translated into content in the request, though, should we - e.g., as a query param on the URL or multipart form data? My understanding has been that the options hash is a mechanism for passing through settings which must be explicitly read/acted upon by the connection class.

Oh right. Good point.

> Embedding knowledge about the specific implementation class used in code that lives outside of the Puppet repo would seem dangerous - e.g., in the event that a future refactor were to change the name of the default implementation class which gets used

Agreed 100%. Even as I was writing that I was skeptical of it. The root concern is it just seems odd to add the facility into puppet when its not used/understood by puppet - eg it loosely feels like inappropriate knowledge, just going in the other direction (server => puppet). That said I'm not sure of a cleaner way and as you suggest `:metric_id` may be of expanded use in the future. I'm good with just a code comment explaining the use/background.

rlinehan:
@MosesMendoza @camlow325 thanks for the review! I updated the commit to add comments that this option is only used by Puppet Server.

MosesMendoza:
@rlinehan is it ok to remove the blocked label now?

rlinehan:
@MosesMendoza oops, yes, removed!

camlow325:
@MosesMendoza Looks like we have two approvals now, so I'm going to merge this.  Thanks for reviewing!

MosesMendoza:
👍

PR #5673, opened 2017-03-01
(PUP-7108) Add pseudo-localized PO file to puppet for testing

Magisus:
This commit adds a PO file to the acceptance directory, for use in smoke
testing the gettext machinery. Note it is currently unused.

What is a good spot to put this file? For now I just stuck it in the acceptance directory, but I'm not sure if that's where we ultimately want it. Unless we put it in the production locales directory (which I gather from the ticket is not something we want to do) I can't use it to test out CI workflows for generating binary MO files, which is the only consideration from my end.

It will currently NOT be automatically kept up to date with the POT file, though that can be added to the CI job that generates the POT if we want.

### DIFF
--- a/acceptance/puppet.po
+++ b/acceptance/puppet.po
@@ -1,0 +1,103 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) 2017 Puppet, Inc.
+# This file is distributed under the same license as the Puppet automation framework package.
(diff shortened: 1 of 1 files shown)


Magisus:
Closing this, going to put up a PR that actually includes a smoke test for the gettext tooling, which will make use of this file.

PR #5676, opened 2017-03-02
(PUP-7252) Ensure puppet agent not running when calling puppet apply

johnduarte:


Magisus:
This makes me sad. But I agree that it's probably the safest way to go about it.

Magisus:
Unless we just want to call the stop helper in between all the cases. I'm not really sure what the tradeoffs are. The Ubuntu failure was on a step that was not guarded with the special `if` block.

johnduarte:
The test itself is trying to verify that `puppet apply` can control the `puppet` and `mco` services. So, using the Beaker helper in certain spots would short circuit the intention of the test (e.g. https://github.com/johnduarte/puppet/blob/6de39ddd06b2f3d14adf9ae18f8ea7cadb012200/acceptance/tests/resource/service/puppet_mcollective_service_management.rb#L36).
The Ubuntu error occurred when attempting to ensure that the service was running and the Beaker helper does not cover this condition.

joshcooper:
I wanted to catch up with you both tomorrow about this before merging, because I'm chasing down a maybe(?) related issue on ubuntu 16.04 https://tickets.puppetlabs.com/browse/PE-19023

MikaelSmith:
Like I said in the previous PR, you could instead use a different statedir to avoid this whole class of problems.

joshcooper:
I reverted merge commit e0f280c8 (in 844320f) which was an attempt to work around the problems introduced by PUP-5609.

I've also reverted merge commit f68f54f5 (in 56eef4a) and reopened PUP-5609 due to all of the downstream CI issues we're seeing.

Since the underlying issue has been addressed (by reverting for now), I'm closing this PR.

PR #5677, opened 2017-03-02
(PUP-7286) Fix regression in Hiera version 3 merge behavior

thallgren:
An oversight caused the a merge behavior specified in the hiera.yaml
file to be applied to all hiera_xxx functions. The hiera_hash is the
only function that should use that definition.

This commit ensures that no other functions are affected by the
merge_behavior or merge_options in hiera.yaml.


hlindberg:

PR #5679, opened 2017-03-02
(PUP-7293) Ensure that Hiera 3 interpolated options are evaluated

thallgren:
Only the `datadir` option in a Hiera 3 backend specific options entry was evaluated. This PR ensures that all keys and values in the options hash gets interpolations evaluated.


hlindberg:

hlindberg:
I retarget this to stable

PR #5682, opened 2017-03-03
(PUP-7296) Ensure that hash key interpolations are evaluated.

thallgren:
The eyaml_lookup_key function used by Hiera did not interpolated
keys in hashes. This commit ensures that it does.


m4ce:
Will this solve https://tickets.puppetlabs.com/browse/PUP-6862 ?

m4ce:
@thallgren, ignore, I see this is only related to the eyaml backend.

hlindberg:
@m4ce PUP 6862 is per design - there will not be a fix for that

PR #5683, opened 2017-03-03
(PUP-7297) Remove bad test for lookup and Hiera 3 merge_behavior

thallgren:
This commit removes a test that verified that lookup made use of
merge_behavior and deep_merge_options defined in hiera.yaml. Althoug
this behavior was added in 5.0, it was found erranious. The behavior
configured in hiera.yaml must only be used by the `hiera_hash` function.
It should never be used by lookup. Consequently, that behavior was
removed as part of fixing PUP-7286.


johnduarte:
Confirmed that this allows the test to pass with the current stable puppet-agent build that is failing in CI.
```
SHA=69ae08451b1133d1e7688021ed583b6bd7fad935 SUITE_COMMIT=1.9.2.29.g69ae084 TEST_TARGET=ubuntu1404-64a be rake ci:test:aio TESTS=tests/lookup/v3_config_and_data.rb
...
tests/lookup/v3_config_and_data.rb passed in 29.01 seconds
```

er0ck:
👍

PR #5686, opened 2017-03-03
(PUP-7294) Performance benchmark for hiera_include

Iristyle:
- Similar to the hiera_function benchmark, this benchmark will call
   the hiera_include function over a large hiearchy.

   The dataset was chosen to demonstrate a Puppet 4.9 regression where
   the same benchmark runs approximately 5x slower than Puppet 4.8.2

   Additionally, note that if groups.yaml has an additional
   hierarchy added that it will run about 50% slower.  If the number
   of users nested under a particular group is doubled from 125 to
   250 that the benchmark will run 4x slower instead of 2x.

Paired-with: Josh Cooper <josh@puppet.com>

### DIFF
--- a/benchmarks/hiera_include/benchmarker.rb
+++ b/benchmarks/hiera_include/benchmarker.rb
@@ -1,0 +1,101 @@
+require 'fileutils'
+
+class Benchmarker
+  include FileUtils
+
+  def initialize(target, size)
+    @target = target
+    @size = size > 100 ? size : 100
+  end
+
+  def setup
+    require 'puppet'
+    @config = File.join(@target, 'puppet.conf')
+    Puppet.initialize_settings(['--config', @config])
+    envs = Puppet.lookup(:environments)
+    @node = Puppet::Node.new('testing', :environment => envs.get('benchmarking'))
+  end
+
+  def run(args=nil)
+    @compiler = Puppet::Parser::Compiler.new(@node)
+    @compiler.compile do |catalog|
+      scope = @compiler.topscope
+      scope['confdir'] = 'test'
+      @size.times do
+        100.times do |index|
+          hiera_func = @compiler.loaders.puppet_system_loader.load(:function, 'hiera_include')
+          hiera_func.call(scope, 'common_entry')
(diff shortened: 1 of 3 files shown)


benchmarks/hiera_include/benchmarker.rb
```diff
@@ -0,0 +1,101 @@
...
+      YAML
+    end
+
+    File.open(groups_yaml, 'w') do |f|
+      f.puts(<<-YAML)
+---
+puppet:
+  staff:
+    groups:
+      YAML
```
Iristyle: @thallgren if you add additional hierarchy, the benchmark runs slower.

benchmarks/hiera_include/benchmarker.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      0.upto(50).each do |i|
+        f.puts("      group#{i}:")
+        0.upto(125).each do |j|
```
Iristyle: @thallgren if you bump 125 to 250, the benchmark runs 4x slower instead of 2x slower

benchmarks/hiera_include/benchmarker.rb
```diff
@@ -0,0 +1,101 @@
...
+          f.puts("        - user#{j}")
+        end
+      end
+    end
+
+    File.open(dummy_class_manifest, 'w') do |f|
+      f.puts("class dummy_class { }")
+    end
+
+    File.open(common_yaml, 'w') do |f|
```
Iristyle: We can collapse the contents of `common.yaml` with `groups.yaml` and end up with the same result.  Happy to update the PR with those changes if it makes things clearer.

I was trying to find a way to eliminate the `dummy_class` to further reduce what's necessary to induce the pathological behavior, but couldn't find a way to do so - seems that there must be at one class defined somewhere to make the call to `hiera_include` like:

```ruby
hiera_func = @compiler.loaders.puppet_system_loader.load(:function, 'hiera_include')
hiera_func.call(scope, 'common_entry')
```

PR #5687, opened 2017-03-04
(PUP-7301) Avoid unnecessary type inference

thallgren:
Since type inference might potentially be very expensive (an instance
can be a very large hash or array), it should only be used unless there
are no cheaper alternatives.

This commit aims to limit the number of type inferences that are
performed during normal execution by:

1. Removing the need for inferring argument types in a call. Instead,
   the callable checks if the argument array is an instance of the
   parameter tuple and if the block is an instance of the block type.
2. Adding some `#instance?` methods where it is relatively easy to
   compute the result without inferring the type.

The rake task `benchmark:hiera_include` shows a very drastic improvement
with this commit in place.

### DIFF
--- a/lib/puppet/pops/types/types.rb
+++ b/lib/puppet/pops/types/types.rb
@@ -138,13 +138,22 @@ class PAnyType < TypedModelObject
     end
   end
 
-  # Returns `true` if this instance is a callable that accepts the given _args_
+  # Returns `true` if this instance is a callable that accepts the given _args_type_ type
   #
-  # @param args [PAnyType] the arguments to test
+  # @param args_type [PAnyType] the arguments to test
   # @param guard [RecursionGuard] guard against recursion. Only used by internal calls
   # @return [Boolean] `true` if this instance is a callable that accepts the given _args_
-  def callable?(args, guard = nil)
-    args.is_a?(PAnyType) && kind_of_callable? && args.callable_args?(self, guard)
+  def callable?(args_type, guard = nil)
+    args_type.is_a?(PAnyType) && kind_of_callable? && args_type.callable_args?(self, guard)
+  end
+
+  # Returns `true` if this instance is a callable that accepts the given _args_
+  #
+  # @param args [Array] the arguments to test
+  # @param block [Proc] block, or nil if not called with a block
+  # @return [Boolean] `true` if this instance is a callable that accepts the given _args_
+  def callable_with?(args,  block = nil)
+    false
   end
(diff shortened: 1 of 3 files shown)


lib/puppet/pops/types/types.rb
```diff
@@ -666,12 +677,21 @@ def _assignable?(o, guard)
...
   def self.register_ptype(loader, ir)
     create_ptype(loader, ir, 'AnyType')
   end
 
   def instance?(o, guard = nil)
-    assignable?(TypeCalculator.infer(o), guard)
+    if KNOWN_SCALARS.any? { |scalar_class| o.is_a?(scalar_class) }
+      true
+    elsif KNOWN_NON_SCALARS.any? { |scalar_class| o.is_a?(scalar_class) }
+      false
```
joshcooper: Could this be converted to a hash lookup instead of linear scan (albeit a short one), e.g

```ruby
KNOWN_SCALARS = { String => true, ...}
...
if KNOWN_SCALARS[scalar_class]
  true
...
```

thallgren: This can be improved for sure. There are two ways to do that:
1. Replace Integer, Float, Timespan, and Timestamp with just Numeric. They are all subclasses anyway so the #is_a? would cover them.
2. Do what you suggest and use a Hash. That will however require that the Hash contains all possible subclasses.
Given that the list will be 3 entries shorter if I change according to #1, I think that would be my preference. Do you agree?

joshcooper: Might also be able to do:

```
case scalar_class
when Numeric, FalseClass, TrueClass, ..
  true
when Array, Hash, PAnyType, NilClass
  false
else
  ...
end
```

since that will use triple equals `===`. I assume ruby builds a lookup table instead of doing a linear scan?

thallgren: In my experience, a case/when is generally slower.  And Ruby will do a linear scan here, testing each expression with ===. There is no key to use for lookup.

thallgren: I wrote [a performance test](/thallgren/ruby-performance/blob/master/case_versus_array_isa_match.rb) for this. The results surprised me a bit. The absolute fastest way to do this test is none of the ones we've been discussing. It's an if/elsif/else statement. I'll change the test accordingly.

Iristyle: 👍

PR #5689, opened 2017-03-04
[WIP] Add configuration option 'merge_facts_later'

nbarrientos:
To be documented much better and tests looked at if the idea goes forward.


nbarrientos:
Kind of superseded by the patch referenced in https://tickets.puppetlabs.com/browse/PUP-7302.

PR #5691, opened 2017-03-06
(PUP-7305) Add file cache to Hiera

thallgren:
This PR adds a file cache with environment life-cycle to the lookup context that is passed to all data provider functions. It also adds a new benchmark named `hiera_include_one` which recreates the compiler between each call to `hiera_include`. This benchmark should display a significant difference before and after the subsequent commits that adds the cache. The FileCache is enabled for the _hiera.yaml_ configuration and all bundled data provider functions (`yaml_data`, `json_data`, `hocon_data`, and `eyaml_lookup_key`).

### DIFF
--- a/lib/puppet/pops/lookup/context.rb
+++ b/lib/puppet/pops/lookup/context.rb
@@ -2,19 +2,73 @@ require_relative 'interpolation'
 
 module Puppet::Pops
 module Lookup
+# The EnvironmentContext is adapted to the current environment
+#
+class EnvironmentContext < Adaptable::Adapter
+  class FileData
+    attr_reader :data
+
+    def initialize(path, inode, mtime, size, data)
+      @path = path
+      @inode = inode
+      @mtime = mtime
+      @size = size
+      @data = data
+    end
+
+    def valid?(stat)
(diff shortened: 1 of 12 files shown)


lib/puppet/pops/lookup/context.rb
```diff
@@ -2,19 +2,73 @@
...
+  # @yieldparam content [String] the content that was read from the file
+  # @yieldreturn [Object] some result based on the content
+  # @return [Object] the content, or if a block was given, the return value of the block
+  #
+  def cached_file_data(path)
+    file_data = @file_data_cache[path]
+    stat = Puppet::FileSystem.stat(path)
+    unless file_data && file_data.valid?(stat)
+      Puppet.debug("File at '#{path}' was changed, reloading") if file_data
+      content = Puppet::FileSystem.read(path, :encoding => 'utf-8')
```
joshcooper: Should we be using `bom` in the encoding? /cc @Iristyle, @MosesMendoza

Iristyle: We have a ticket to change everything to `bom|utf-8`

PR #5692, opened 2017-03-06
(PUP-7309) acceptance: hiera merge strategies

er0ck:
test hiera3 and hiera5 merge strategies using hiera_hash and lookup
functions

edit: added a commit for hiera_array and merge_hash_arrays

### DIFF
--- a/acceptance/tests/lookup/merge_strategies.rb
+++ b/acceptance/tests/lookup/merge_strategies.rb
@@ -37,6 +37,8 @@ test_name 'C99903: merge strategies' do
   - "%{::kernel}"
   - "common"
 :merge_behavior: deeper
+:deep_merge_options:
+  :merge_hash_arrays: true
 HIERA
     on(master, "chown puppet:puppet #{master_confdir}/hiera.yaml")
 
@@ -52,10 +54,13 @@ profiles:
           - php
           - ssl
 arrayed_hash:
-  - array1:
-      key1: val1
-      key2: val2
-    YAML
+  the_hash:
+    - array1:
+        key1: val1
+        key2: val2
+array:
+  - foo
+YAML
 
     create_remote_file(master, "#{fq_tmp_environmentpath}/hieradata/profiles.yaml", <<-YAML)
 profiles:
@@ -68,18 +73,19 @@ profiles:
           - cgid
           - php
           - status
-arrayed_hash:
-  - array1:
-      key1: valB
-      key3: val3
+array:
+  - bar
 YAML
 
     create_sitepp(master, tmp_environment, <<-SITE)
 notify { "hiera_hash: ${hiera_hash ('profiles')['webserver']['apache']['httpd']['modules']}": }
 notify { "lookup1: ${lookup ('profiles')['webserver']['apache']['httpd']['modules']}": }
 notify { "lookup1b: ${lookup ({'name' => 'profiles', 'merge' => 'deep'})['webserver']['apache']['httpd']['modules']}": }
-notify { "hiera_arrayed_hash: ${hiera_array ('arrayed_hash')}": }
+notify { "hiera_merge_hash: ${hiera_hash ('arrayed_hash')}": }
 notify { "lookup_arrayed_hash: ${lookup ({'name' => 'arrayed_hash', 'merge' => {'strategy' => 'deep', 'merge_hash_arrays' => true}})}": }
+notify { "hiera-array: ${hiera ('array')}": }
+notify { "hiera_array: ${hiera_array ('array')}": }
+notify { "lookup-array: ${lookup ('array')}": }
     SITE
 
     on(master, "chmod -R 775 #{fq_tmp_environmentpath}")
@@ -110,9 +116,12 @@ profiles:
           - php
           - ssl
 arrayed_hash:
-  - array1:
-      key1: val1
-      key2: val2
+  the_hash:
+    - array1:
+        key1: val1
+        key2: val2
+array:
+  - foo
 lookup_options:
   'profiles':
     merge:
@@ -130,10 +139,8 @@ profiles:
           - cgid
           - php
           - status
-arrayed_hash:
-  - array1:
-      key1: valB
-      key3: val3
+array:
+  - bar
 lookup_options:
   'profiles':
     merge:
@@ -144,8 +151,11 @@ YAML
 notify { "hiera_hash: ${hiera_hash ('profiles')['webserver']['apache']['httpd']['modules']}": }
 notify { "lookup2: ${lookup ('profiles')['webserver']['apache']['httpd']['modules']}": }
 notify { "lookup2b: ${lookup ({'name' => 'profiles', 'merge' => 'first'})['webserver']['apache']['httpd']['modules']}": }
-notify { "hiera_arrayed_hash: ${hiera_array ('arrayed_hash')}": }
+notify { "hiera_merge_hash: ${hiera_hash ('arrayed_hash')}": }
 notify { "lookup_arrayed_hash: ${lookup ({'name' => 'arrayed_hash', 'merge' => {'strategy' => 'deep', 'merge_hash_arrays' => true}})}": }
+notify { "hiera-array: ${hiera ('array')}": }
+notify { "hiera_array: ${hiera_array ('array')}": }
+notify { "lookup-array: ${lookup ('array')}": }
     SITE
 
     on(master, "chmod -R 775 #{fq_tmp_environmentpath2}")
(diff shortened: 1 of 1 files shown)


acceptance/tests/lookup/merge_strategies.rb
```diff
@@ -0,0 +1,172 @@
...
+  fq_tmp_environmentpath  = "#{environmentpath}/#{tmp_environment}"
+  tmp_environment2 = mk_tmp_environment_with_teardown(master, app_type)
+  fq_tmp_environmentpath2  = "#{environmentpath}/#{tmp_environment2}"
+
+  confdir = master.puppet('master')['confdir']
+  codedir = master.puppet('master')['codedir']
+
+  teardown do
+    step "remove global hiera.yaml" do
+      on(master, "rm #{confdir}/hiera.yaml")
```
joshcooper: The global hiera.yaml is installed by packages, so we shouldn't remove it, and the test should use a temp hiera.yaml instead (or backup/restore).

er0ck: oh right.  thanks.  i'll backup/restore

acceptance/tests/lookup/merge_strategies.rb
```diff
@@ -0,0 +1,194 @@
...
+    create_sitepp(master, tmp_environment, <<-SITE)
+notify { "hiera_hash: ${hiera_hash ('profiles')['webserver']['apache']['httpd']['modules']}": }
+notify { "lookup1: ${lookup ('profiles')['webserver']['apache']['httpd']['modules']}": }
+notify { "lookup1b: ${lookup ({'name' => 'profiles', 'merge' => 'deep'})['webserver']['apache']['httpd']['modules']}": }
+notify { "hiera_arrayed_hash: ${hiera_array ('arrayed_hash')}": }
+notify { "lookup_arrayed_hash: ${lookup ({'name' => 'arrayed_hash', 'merge' => {'strategy' => 'deep', 'merge_hash_arrays' => true}})}": }
+    SITE
+
+    on(master, "chmod -R 775 #{fq_tmp_environmentpath}")
+    on(master, "chmod -R 775 #{master_confdir}")
```
joshcooper: Is changing the permissions on the real `confdir` going to affect/break later tests that run `puppet apply/agent` and assume there are no changes? Do we just need to make the newly generated `hiera.yaml` ownable by `puppet:puppet`?

er0ck: yeah, sloppy.  i'll be more restrictive

acceptance/tests/lookup/merge_strategies.rb
```diff
@@ -0,0 +1,194 @@
...
+  fq_tmp_environmentpath  = "#{environmentpath}/#{tmp_environment}"
+  tmp_environment2 = mk_tmp_environment_with_teardown(master, app_type)
+  fq_tmp_environmentpath2  = "#{environmentpath}/#{tmp_environment2}"
+
+  master_confdir = master.puppet('master')['confdir']
+  hiera_conf_backup = master.tmpfile(app_type)
+
+  teardown do
+    step "restore default global hiera.yaml" do
+      on(master, "mv #{hiera_conf_backup} #{master_confdir}/hiera.yaml")
```
johnduarte: This teardown step should ensure that the `hiera_conf_backup` file is present beforehand. If the backup step below should fail to create the backup file, then the teardown will erroneously fail.

acceptance/tests/lookup/merge_strategies.rb
```diff
@@ -0,0 +1,194 @@
...
+      httpd:
+        modules:
+          - mpm_prefork
+          - php
+          - ssl
+arrayed_hash:
+  - array1:
+      key1: val1
+      key2: val2
+    YAML
```
johnduarte: nitpick, this is the only indented heredoc terminator in this PR

acceptance/tests/lookup/merge_strategies.rb
```diff
@@ -0,0 +1,194 @@
...
+notify { "hiera_arrayed_hash: ${hiera_array ('arrayed_hash')}": }
+notify { "lookup_arrayed_hash: ${lookup ({'name' => 'arrayed_hash', 'merge' => {'strategy' => 'deep', 'merge_hash_arrays' => true}})}": }
+    SITE
+
+    on(master, "chmod -R 775 #{fq_tmp_environmentpath2}")
+  end
+
+  with_puppet_running_on(master,{}) do
+    agents.each do |agent|
+      step "agent lookups #{agent.hostname}, hiera3" do
```
johnduarte: Would this test be more readable and maintainable if it was split into two tests, one for v3 and one for v5?

acceptance/tests/lookup/merge_strategies.rb
er0ck: 😐 meh

i'm not clear why that's important.  in particular because i stole this from your test O:)
although you have acceptable exit codes of 0 and 1

acceptance/tests/lookup/merge_strategies.rb
er0ck: i believe the opposite.  putting it in one place makes it more discoverable, readable and maintainable, IMO

acceptance/tests/lookup/merge_strategies.rb
joshcooper: That's one reason why I slightly prefer generating temp files and pointing the test to the temp file. Less to worry about when cleaning up, and less likely to break test that follow.

er0ck: yeah, but you can't do that here without reloading/restarting the server.  i'm trying to avoid that in our tests so we can later remove `#with_puppet_running_on`.

i agree it's a tradeoff and we should probably make a decision on if you can touch the default install/config.  we might even make it difficult to do, in the tooling or on the SUT in the pre-suite steps.

we should do some profiling with the proposed beaker profiler

acceptance/tests/lookup/merge_strategies.rb
er0ck: it's just on the verge of being too long :shark:

acceptance/tests/lookup/merge_strategies.rb
joshcooper: I admire the direction, but I'm not sure that skipping restart (at least HUP) is safe, due to caching in hiera and/or the compiler. So I'm ok to merge this PR, but something to think about before removing `with_puppet_running_on`.

er0ck:
please do not merge until after PA 1.9.3 ships.

acceptance/tests/lookup/merge_strategies.rb
er0ck: the one argument i'm thinking of for splitting it in two is that it could be easier to remove the v3 testing when it comes time.  but this is still very easy in this test, and extremely difficult in other tests like lookup/lookup.rb :-\

joshcooper:
@er0ck could you verify this test works when there is an agent on the master?

er0ck:
```
    "HOSTS": {
        "r0pibn118h5p96h.delivery.puppetlabs.net": {
            "pe_dir": null,
            "pe_ver": null,
            "pe_upgrade_dir": null,
            "pe_upgrade_ver": null,
            "hypervisor": "vmpooler",
            "platform": "el-7-x86_64",
            "template": "redhat-7-x86_64",
            "roles": [
                "master",
                "agent",
                "default"
            ],
            "host_tags": {},
            "user": "root",
            "group": "puppet",
            "pathseparator": ":",
            "vmhostname": "r0pibn118h5p96h.delivery.puppetlabs.net",
            "puppetbindir": "/opt/puppetlabs/bin",
            "privatebindir": "/opt/puppetlabs/puppet/bin",
            "distmoduledir": "/etc/puppetlabs/code/modules",
            "sitemoduledir": "/opt/puppetlabs/puppet/modules"
        },
        "kmkbp73wgtc5uc0.delivery.puppetlabs.net": {
            "pe_dir": null,
            "pe_ver": null,
            "pe_upgrade_dir": null,
            "pe_upgrade_ver": null,
            "hypervisor": "vmpooler",
            "platform": "ubuntu-14.04-amd64",
            "template": "ubuntu-1404-x86_64",
            "roles": [
                "agent",
                "aio"
            ],
            "host_tags": {},
            "user": "root",
            "group": "puppet",
            "pathseparator": ":",
            "vmhostname": "kmkbp73wgtc5uc0.delivery.puppetlabs.net",
            "puppetbindir": "/opt/puppetlabs/bin",
            "privatebindir": "/opt/puppetlabs/puppet/bin",
            "distmoduledir": "/etc/puppetlabs/code/modules",
            "sitemoduledir": "/opt/puppetlabs/puppet/modules"
        }
    },

```
...
```
tests/lookup/merge_strategies.rb passed in 27.37 seconds
      Test Suite: tests @ 2017-03-13 21:21:56 +0000

       Total Suite Time: 27.37 seconds
      Average Test Time: 27.37 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
Pending Tests Cases:
```

PR #5693, opened 2017-03-07
(PUP-7312) Confine systemd provider to /lib/systemd/system

arioch:
/run/systemd/system does not exist when running Puppet in a chroot environment
/lib/systemd/system is always available irregardless of the environment and is therefore a safe alternative

### DIFF
--- a/lib/puppet/provider/service/systemd.rb
+++ b/lib/puppet/provider/service/systemd.rb
@@ -1,5 +1,7 @@
 # Manage systemd services using systemctl
 
+require 'puppet/file_system'
+
 Puppet::Type.type(:service).provide :systemd, :parent => :base do
   desc "Manages `systemd` services using `systemctl`.
 
@@ -9,14 +11,7 @@ Puppet::Type.type(:service).provide :systemd, :parent => :base do
 
   commands :systemctl => "systemctl"
 
-  if Facter.value(:osfamily).downcase == 'debian'
-    # With multiple init systems on Debian, it is possible to have
-    # pieces of systemd around (e.g. systemctl) but not really be
-    # using systemd.  We do not do this on other platforms as it can
-    # cause issues when running in a chroot without /run mounted
-    # (PUP-5577)
-    confine :exists => "/run/systemd/system"
-  end
+  confine :true => Puppet::FileSystem.exist?('/proc/1/exe') && Puppet::FileSystem.readlink('/proc/1/exe').include?('systemd')
 
   defaultfor :osfamily => [:archlinux]
   defaultfor :osfamily => :redhat, :operatingsystemmajrelease => ["7", "8"]
--- a/spec/integration/provider/service/systemd_spec.rb
+++ b/spec/integration/provider/service/systemd_spec.rb
@@ -15,6 +15,11 @@ describe test_title, unless: Puppet::Util::Platform.jruby? do
     expect(provider_class).to be_suitable
   end
 
+  it "should be considered suitable if /proc/1/exe is present and contains 'systemd'",
+    :if => File.exist?('/proc/1/exe') && File.open('/proc/1/exe').grep(/systemd/).size > 0
+    expect(provider_class).to be_suitable
+  end
+
   it "should not be cosidered suitable if systemctl is absent",
     :unless => (File.executable?('/bin/systemctl') or File.executable?('/usr/bin/systemctl')) do
     expect(provider_class).not_to be_suitable


MosesMendoza:
Hi @arioch, thanks for submitting this. I wonder if we should actually be checking if /sbin/init is a symlink to systemd, per the latest comment in PUP-5577. Seems to be consistent on some platforms. Also, I think the location of systemd might vary on some platforms.. On SLES 12, it's in /usr/lib..

```
shrhbklg760ugrk:~ # ls -l /usr/lib/systemd/systemd
-rwxr-xr-x 1 root root 1206688 Oct  8  2014 /usr/lib/systemd/systemd
shrhbklg760ugrk:~ # cat /etc/os-release
NAME="SLES"
VERSION="12"
VERSION_ID="12"
PRETTY_NAME="SUSE Linux Enterprise Server 12"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:12"
```

Thoughts?

arioch:
Hm. The plot thickens... 
I've whipped up a tiny script to check /sbin/init and a few related directories:

```
#!/bin/bash

for i in debian centos ubuntu fedora opensuse
do
  echo $i
  docker run -ti ${i}:latest ls -l /sbin/init
  for j in /lib/systemd/system /usr/lib/systemd/system /run/systemd/system
  do
    docker run -ti ${i}:latest test -d $j && echo "√ ${j}" || echo "- ${j}"
  done
  echo
done
```

```
debian
lrwxrwxrwx 1 root root 20 Jan  4 23:42 /sbin/init -> /lib/systemd/systemd
√ /lib/systemd/system
- /usr/lib/systemd/system
- /run/systemd/system

centos
lrwxrwxrwx 1 root root 22 Dec 14 14:58 /sbin/init -> ../lib/systemd/systemd
√ /lib/systemd/system
√ /usr/lib/systemd/system
- /run/systemd/system

ubuntu
lrwxrwxrwx 1 root root 20 Jan 18 22:04 /sbin/init -> /lib/systemd/systemd
√ /lib/systemd/system
- /usr/lib/systemd/system
- /run/systemd/system

fedora
lrwxrwxrwx 1 root root 22 Jan 16 12:13 /sbin/init -> ../lib/systemd/systemd
√ /lib/systemd/system
√ /usr/lib/systemd/system
- /run/systemd/system

opensuse
ls: cannot access '/sbin/init': No such file or directory
- /lib/systemd/system
√ /usr/lib/systemd/system
- /run/systemd/system
```

On the distros I've checked none of them have /run/systemd/system available on Docker, similar to the original problem as stated above. But they all do link /sbin/init to the same binary.

arioch:
Based on this I would suggest modifying the code as follows:

```
case Facter.value(:osfamily).downcase
when 'opensuse'
  confine :exists => "/usr/lib/systemd/system"
else
  confine :exists => "/lib/systemd/system"
end
```

What are your thoughts on this?

arioch:
I've spent hours scratching my head trying to figure out why the tests fail and I can't reproduce it on an actual box.

Anyways I think it's because /sbin/init isn't available on TravisCI/AppVeyor docker images.
Any advice how to abstract this?

adrienthebo:
`/run/systemd/system` may not be present when running Docker, since Systemd and Docker don't really play well together. Running these same checks via Vagrant or somesuch might be illuminating and can help show differences between Systemd in Docker and Systemd on a bare metal or fully virtualized system.

It's also quite possible that there is no difference between Systemd in Docker and Systemd on non-containerized environments, but it's worth thinking about. :)

MosesMendoza:
@arioch yeah its interesting - the travis test fails as if '/sbin/init' exists, but is not a symlink. Turns out if passed a non-symlink file, `File::readlink` raises Invalid argument, while if the file doesn't exist it raises No such file or directory.

```
[6] pry(main)> Puppet::FileSystem.readlink('/Users/moses/foo/foo_link.pp')
=> "foo_link_target.pp"
[7] pry(main)> Puppet::FileSystem.readlink('/Users/moses/foo/foo_hard_link')
Errno::EINVAL: Invalid argument @ rb_readlink - /Users/moses/foo/foo_hard_link
from /Users/moses/development/puppet/lib/puppet/file_system/file_impl.rb:132:in `readlink'
[8] pry(main)> Puppet::FileSystem.readlink('/Users/moses/foo/does_not_exist')
Errno::ENOENT: No such file or directory @ rb_readlink - /Users/moses/foo/does_not_exist
from /Users/moses/development/puppet/lib/puppet/file_system/file_impl.rb:132:in `readlink'
```

Maybe we need to check if its a symlink first to avoid raising? I.e., 
```
[31] pry(main)> !!(Puppet::FileSystem.symlink?('/Users/moses/foo/foo_link.pp') && (Puppet::FileSystem.readlink('/Users/moses/foo/foo_link.pp') =~ %r{foo_link_target.pp}))
=> true
```

Regarding suse, it looks like Suse Linux Enterprise Server does have '/sbin/init', but from your OpenSuse test, OpenSuse does not. That's probably fine because they have different os families, it appears. On SLES 12:
```ruby
irb(main):002:0> Facter.value('osfamily')
=> "Suse"
irb(main):004:0> Puppet::FileSystem.readlink('/sbin/init')
=> "../usr/lib/systemd/systemd"
```

So maybe a hybrid approach - just check for the existence of '/usr/lib/systemd/system' on OpenSuse and otherwise do the symlink test?

lutter:
Wouldn't it be simpler to just check the contents of `/proc/1/cmdline` ? Alternatively, `ps --no-headers -o comm 1` should also provide an answer (at the cost of requiring that ps be there)

arioch:
That's an interesting approach... 
Let's see what the test suite thinks about that.

I don't think `ps --no-headers -o comm 1` is an option in this particular case because not all distros install`procps` by default.

arioch:
... that's weird.
Either I misinterpreted the use of confine or confine is broken.

This works:
```
next unless Puppet::FileSystem.exist?('/proc/1/cmdline')
```

While these lines don't:
```
confine :exists => '/proc/1/cmdline'
confine :true => Puppet::FileSystem.exist?('/proc/1/cmdline')
```

joshcooper:
@arioch your PR is failing travis because the summary line for each commit needs to start with the ticket number `(PUP-7312) ...` Can you update those commits, and force push to this topic branch?

vinzent:
systemd can be run in a docker container. I've started a centos7 container and /proc/1/cmdline contained `/bin/bash`. wouldn't this confine to /proc/1/cmdline content prevent the systemd provider from working in this situation?

arioch:
@joshcooper as requested

joshcooper:
@vinzent sorry, missed this. The specs are failing on windows. I'd just skip all of the systemd tests on Windows by adding an rspec filter around the outermost describe block:

```ruby
describe '....', :unless => Puppet.features.microsoft_windows?
  ...
end
```

lib/puppet/provider/service/systemd.rb
```diff
@@ -1,5 +1,7 @@
 # Manage systemd services using systemctl
 
+require 'puppet/file_system/file_impl'
```
joshcooper: probably just want to `require 'puppet/file_system'` and let it figure out which impl to require.

arioch: Done.

lib/puppet/provider/service/systemd.rb
```diff
@@ -9,14 +11,8 @@
...
-  if Facter.value(:osfamily).downcase == 'debian'
-    # With multiple init systems on Debian, it is possible to have
-    # pieces of systemd around (e.g. systemctl) but not really be
-    # using systemd.  We do not do this on other platforms as it can
-    # cause issues when running in a chroot without /run mounted
-    # (PUP-5577)
-    confine :exists => "/run/systemd/system"
-  end
+  next unless Puppet::FileSystem.exist?('/proc/1/cmdline')
+  confine :true => Puppet::FileSystem.read('/proc/1/cmdline').include?('systemd')
```
branan: On my Ubuntu 18.04 system, `/proc/1/cmdline` is `/sbin/initsplash`, and doesn't contain the string `systemd`.

It does look like the `/proc/1/exe` symlink does point to `/lib/systemd/systemd`, though. Perhaps that's the better thing to try to confine on?

arioch: Not every distro has `/lib/systemd/systemd` as pointed out above.
However in my very unscientific tests the distro subset mentioned over a year ago all seem to have `/proc/1/exe` available, making it a valid candidate. Thanks for that!

Patch coming right up.

melissa:
It looks like we're still waiting for test changes

cc/ @branan

jtappa:
Could someone from @puppetlabs/platform-os please take a look at this?

pillarsdotnet:
[Test failures](https://ci.appveyor.com/project/puppetlabs/puppet/builds/25010332) appear unrelated to patch.

arioch:
> Should pass tests if [rebased and resubmitted](https://github.com/puppetlabs/puppet/pull/7545).

^ done

mihaibuzgau:
jenkins please test this on debian8-64a,debian8-32a,debian9-64a,debian9-32a,fedora28-64a,fedora29-64a,osx1012-64a,osx1013-64a,osx1014-64a,redhat5-64a,centos6-64a,centos6-32a,redhat7-64a,redhat8-64a,redhatfips7-64a,sles11-64a,sles11-32a,sles12-64a,sles12-POWERa,sles15-64a,solaris10-64a,solaris11-64a,solaris114-64a,ubuntu1404-64a,ubuntu1404-32a,ubuntu1604-64a,ubuntu1604-32a,ubuntu1804-64a,windows2016-64a,windows2019-64a

GabrielNagy:
Hi @arioch,

Can you update the tests in `spec/integration/provider/service/systemd_spec.rb` to check the changed functionality? (checking for `/proc/1/exe` and following the symlink)

Once this is done we can merge this.

Thanks!

mihaibuzgau:
fixed in this PR:  https://github.com/puppetlabs/puppet/pull/7628

PR #5694, opened 2017-03-07
(PUP-7313) Improve compiler initialization performance

thallgren:
This PR contains a series of commits that aim to improve the time it takes for the Puppet compiler to initialize. The main problem that it solves concerns initialization of the compiler scope from the Puppet settings. That slowness of that initialization gets worse as the number of settings increases.

### DIFF
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -748,7 +748,7 @@ class Puppet::Parser::Scope
   # to be reassigned.
   #   It's preferred that you use self[]= instead of this; only use this
   # when you need to set options.
-  def setvar(name, value, options = {})
+  def setvar(name, value, options = EMPTY_HASH)
     if name =~ /^[0-9]+$/
       raise Puppet::ParseError.new("Cannot assign to a numeric match result variable '$#{name}'") # unless options[:ephemeral]
     end
@@ -845,12 +845,12 @@ class Puppet::Parser::Scope
   #
   # @param varname [String] The variable name to which the value is assigned. Must not contain `::`
   # @param value [String] The value to assign to the given variable name.
-  # @param options [Hash] Additional options, not part of api.
+  # @param options [Hash] Additional options, not part of api and no longer used.
   #
   # @api public
   #
-  def []=(varname, value, options = EMPTY_HASH)
-    setvar(varname, value, options)
+  def []=(varname, value, _ = nil)
+    setvar(varname, value)
   end
 
   def append_value(bound_value, new_value)
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -588,7 +588,7 @@ class Puppet::Parser::Scope
   end
 
   def is_topscope?
-    compiler and self == compiler.topscope
+    @compiler && equal?(@compiler.topscope)
   end
 
   def lookup_qualified_variable(class_name, variable_name, position)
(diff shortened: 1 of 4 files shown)


lib/puppet/parser/scope.rb
```diff
@@ -820,8 +849,8 @@ def effective_symtable(use_ephemeral)
   #
   # @api public
   #
-  def []=(varname, value, options = {})
-    setvar(varname, value, options = {})
+  def []=(varname, value, options = EMPTY_HASH)
+    setvar(varname, value, options)
```
hlindberg: This is different - the original blocks the options from being passed! I think that should be preserved.
Can be done a better way.

lib/puppet/settings.rb
```diff
@@ -1046,26 +1048,33 @@ def values(environment, section)
...
-    if @cache[environment||"none"].has_key?(param)
-      return @cache[environment||"none"][param]
-    elsif bypass_interpolation
-      val = values(environment, self.preferred_run_mode).lookup(param)
-    else
-      val = values(environment, self.preferred_run_mode).interpolate(param)
-    end
+    cached_env = @cache[environment || NONE]
+    val = cached_env[param]
+    return val if val || cached_env.include?(param)
```
hlindberg: what if `val` is a Boolean `false`

thallgren: then `false` will be returned since `cached_env.include?(param)` will return `true`.

hlindberg: so that is a slower route - why not do `if !val.nil?`

thallgren:
So, in > 90% of the cases, the `val` will be different than `false` or `nil`. Hence, that test is done first. In case `val` is `false` or `nil`, only then is the control made whether or not the hash actually contains the key. I can't think of a faster route to do this.

hlindberg:
so you think `if val` is faster than `if !val.nil?` ?

hlindberg:
@joshcooper ok - assigning ticket to you.

thallgren:
@joshcooper, did you have a chance to look at this one yet? I'd like to have it merged before I commence work on PUP-7328 since it touches the same code.

lib/puppet/parser/scope.rb
```diff
@@ -816,12 +845,12 @@ def effective_symtable(use_ephemeral)
   #
   # @param varname [String] The variable name to which the value is assigned. Must not contain `::`
   # @param value [String] The value to assign to the given variable name.
-  # @param options [Hash] Additional options, not part of api.
+  # @param options [Hash] Additional options, not part of api and no longer used.
   #
   # @api public
   #
-  def []=(varname, value, options = {})
-    setvar(varname, value, options = {})
```
joshcooper: Despite what the comment says, this is a breaking change, right? Previously, you could do `scope['foo', 'bar', :append => true)` and change how the variables were defined. Now the `options` argument is ignored.

thallgren: The argument was ignored before this change. An empty array is unconditionally assigned in the call to {{setvar}}.

joshcooper: Oh! I completely missed the assignment! Yeah that's terrible.

lib/puppet/settings.rb
```diff
@@ -1046,26 +1048,37 @@ def values(environment, section)
...
+    cached_env = @cache[environment || NONE]
+
+    # Avoid two lookups in cache_env unless val is nil. When it is, it's important
+    # to check if the key is included so that further processing (that will result
+    # in nil again) is avoided.
+    val = cached_env[param]
+    return val if !val.nil? || cached_env.include?(param)
+
+    # Short circuit to nil for undefined settings.
+    return nil unless @config.include?(param)
```
joshcooper: Should this be `cached_env.include?(param)`?

thallgren: No, I don't think so. The code used to read:
```
    setting = @config[param]

    # Short circuit to nil for undefined settings.
    return nil if setting.nil?
```

lib/puppet/parser/scope.rb
joshcooper: I just saw @hlindberg's comment. Are you still discussing this issue?

One question is whether this was accidentally broken recently, perhaps in a bad merge? As in, may be this did work correctly in 4.7, but was broken in between? But if it's one of those things that has been broken for a very long time, then I'm 👍  to your change.

joshcooper: It looks like it's been broken since 57e1db0c in 2012, so nevermind.

PR #5697, opened 2017-03-07
(PUP-7310) Convert string interpolations to format strings in lib/puppet/face/help.rb

Magisus:
The gettext tooling does not work correctly with interpolated strings.
Any string marked for extraction into the POT file must therefore be
converted to use Ruby's string format syntax. This updates the extracted
strings in `lib/puppet/face/help.rb`, which was used as a proof of
concept for the gettext tooling.

This file was edited with the help of a [new script](https://github.com/Magisus/string-wrapper/blob/master/interpolation_extracter.rb) I wrote using the same Ruby AST rewriter I used to do the original extraction. It finds marked strings and performs the extraction, using a placeholder name for values that are not plain variables. I then went in and renamed the placeholder values to something more descriptive for the translator. This should in general avoid the need for extensive translator comments (and might actually improve clarity for the translators overall).

### DIFF
--- a/lib/puppet/face/help.rb
+++ b/lib/puppet/face/help.rb
@@ -78,11 +78,11 @@ Puppet::Face.define(:help, '0.0.1') do
   def render_application_help(applicationname)
     return Puppet::Application[applicationname].help
   rescue StandardError, LoadError => detail
-    msg = _(<<-MSG)
-Could not load help for the application #{applicationname}.
+    msg = _(<<-MSG) % { applicationname: applicationname, detail_msg: detail.message, }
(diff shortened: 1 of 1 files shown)


MosesMendoza:
I wonder why appveyor failed. It shows all passed, but then exited 1.
```
Finished in 9 minutes 36 seconds
21184 examples, 0 failures, 64 pending
Command exited with code 1

```

Magisus:
Rekick fixed Appveyor.

Magisus:
Closing this, it will get fixed in my update to https://github.com/puppetlabs/puppet/pull/5572

PR #5698, opened 2017-03-08
(maint) Remove gettext rake task wrappers

Magisus:
This commit removes the wrappers for the gettext rake tasks in favor of
squashing the Rake load error when the gettext-setup gem cannot be
found.


MosesMendoza:
also merged up to master at 23e9edb4e7afe458c916ecf8cbafefdbf43803fc

PR #5699, opened 2017-03-08
(PUP-4039) - Improve yum provider handling of new package names

werekraken:
The yum provider will fail to find new package names due to `yum list` not refreshing metadata. #2654 caused this regression, but also implemented a much needed fix.

This change keeps the fix provided by #2654 and fixes new package name issue without extra manifest code. It also only causes extra steps to be taken when the initial `yum list` fails. Performance should not be impacted unless the package would otherwise fail.

Details available in [PUP-4039](https://tickets.puppetlabs.com/browse/PUP-4039).

### DIFF
--- a/lib/puppet/provider/package/yum.rb
+++ b/lib/puppet/provider/package/yum.rb
@@ -142,13 +142,24 @@ Puppet::Type.type(:package).provide :yum, :parent => :rpm, :source => :rpm do
     'update'
   end
 
+  def makecache_fast
+    command = [command(:cmd), :makecache, :fast]
+    execute(command, :failonfail => false, :combine => false)
+  end
+
   def install
     wanted = @resource[:name]
     error_level = self.class.error_level
     update_command = self.class.update_command
     # If not allowing virtual packages, do a query to ensure a real package exists
     unless @resource.allow_virtual?
-      execute([command(:cmd), '-d', '0', '-e', error_level, '-y', install_options, :list, wanted].compact)
+      command = [command(:cmd), '-d', '0', '-e', error_level, '-y', install_options, :list, wanted].compact
+      output = execute(command, :failonfail => false, :combine => false)
+
+      if output.exitstatus != 0
+        makecache_fast
+        execute(command)
+      end
     end
 
     should = @resource.should(:ensure)
--- a/spec/shared_examples/rhel_package_provider.rb
+++ b/spec/shared_examples/rhel_package_provider.rb
@@ -147,10 +147,28 @@ shared_examples "RHEL package provider" do |provider_class, provider_name|
         Puppet::Util::Execution.expects(:execute).with(["/usr/bin/#{provider_name}", '-d', '0', '-e', error_level, '-y', ['-t', '-x=expackage'], :install, name])
         provider.install
       end
-      it 'allow virtual packages' do
+      it 'should allow virtual packages' do
         resource[:ensure] = :installed
         resource[:allow_virtual] = true
-        Puppet::Util::Execution.expects(:execute).with(["/usr/bin/#{provider_name}", '-d', '0', '-e', error_level, '-y', :list, name]).never
+        Puppet::Util::Execution.expects(:execute).with(["/usr/bin/#{provider_name}", '-d', '0', '-e', error_level, '-y', :list, name], {:failonfail => false, :combine => false}).never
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/package/yum.rb
```diff
@@ -142,13 +142,24 @@ def self.update_command
...
+  end
+
   def install
     wanted = @resource[:name]
     error_level = self.class.error_level
     update_command = self.class.update_command
     # If not allowing virtual packages, do a query to ensure a real package exists
     unless @resource.allow_virtual?
-      execute([command(:cmd), '-d', '0', '-e', error_level, '-y', install_options, :list, wanted].compact)
+      command = [command(:cmd), '-d', '0', '-e', error_level, '-y', install_options, :list, wanted].compact
```
adrienthebo: I'm a bit leery about duplicating this command string. If we need to change the invocation of this command it would be possible to only do it for one of the invocations. Could we move this command list into a variable and then just use that variable for both invocations?

werekraken: Sure. I had only duplicated it since I saw that in other package providers (e.g. dpkg.rb).

adrienthebo:
@werekraken how familiar are you with writing unit test, and RSpec tests in specific? We like to have tests for new features so that we don't accidentally regress on them.

werekraken:
@adrienthebo tests added. Sorry for not including them upfront. I had looked at yum_spec.rb, but had missed all the meat in shared_examples/rhel_package_provider.rb.

lib/puppet/provider/package/yum.rb
```diff
@@ -142,13 +142,24 @@ def self.update_command
...
+      output = execute(command, :failonfail => false, :combine => false)
+
+      if output.exitstatus != 0
```
branan: One concern I have here is that if `yum list` is going to fail consistently for some reason, we could potentially be running `yum makecache` for every package resource - this would cause a puppet run to be very slow and use a lot of resources just to ultimately fail. Is it possible to more precise about catching the failure we're expecting to see (the package isn't in the cache)?

werekraken: I can parse the output of `yum repolist enabled -v` and only run `yum makecache` when a repo is expired (comparing now to last plus metadata_expire seconds).
```
[root@localhost ~]# yum repolist enabled -v | grep 'Repo-name\|expire'
Repo-name    : CentOS-7 - Base
Repo-expire  : 21,600 second(s) (last: Wed Mar  8 19:01:59 2017)
Repo-name    : CentOS-7 - Extras
Repo-expire  : 21,600 second(s) (last: Wed Mar  8 19:02:00 2017)
Repo-name    : CentOS-7 - Updates
Repo-expire  : 21,600 second(s) (last: Wed Mar  8 19:02:01 2017)
```
I think we can even use `yum makecache fast --disablerepo='*' --enablerepo=<repo>` to target the specific repos which are expired.

Parsing output may have portability concerns. Pulling this programmatically via the yum python library was pretty easy.
```
[root@localhost ~]# python
Python 2.7.5 (default, Aug 18 2016, 15:58:25) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import yum
>>> y = yum.YumBase()
>>> for repo in y.repos.listEnabled():
...   print repo
...   print repo.metadataCurrent()
... 
Loaded plugins: fastestmirror
base
True
extras
True
localrepo
False
updates
True
```
Do either of these solutions sound reasonable? If the latter, is there a standard/desired way to include a python helper in a ruby project (I'm not aware of comprehensive yum library for ruby).

Magisus: @branan thoughts on this reply?

branan: We specifically *removed* our python shim - I think option 1 is the way to go here

melissa:
@werekraken Hey! This has been open for quite a while with no update from you. We do value your contribution, but we're trying to keep out backlog of pull requests a bit cleaner. If we don't hear back within two weeks, we will close this pull request. At that point, if you do get back to working on this, you can always reopen this PR or create a new one!

melissa:
Hey @werekraken. It's been a while since we've heard from you. I'm going to go ahead and close this pull request. If you do get the opportunity to circle back to this work, always feel free to either re-open this PR or create a new one. Thank you for taking the time to contribute to Puppet!

PR #5701, opened 2017-03-09
(PUP-7329) This change fixes a regex in describe.rb.

seangriff:
Before this fix a long line without spaces in a puppet type description
causes puppet describe to fail at formatting the description.  This is a
side effect of anchoring the match expression to ^, beginning of string,
when it should be anchored with \A, beginning of line.

For example, with this type, (described by modules/mytype_module/lib/puppet/type/mytype.rb):

    Puppet::Type.newtype(:mytype) do
      @doc = ". 12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890 nick@magpie.puppetlabs.lan
    **this part should not repeat!**
    "
      newparam(:name) do
        isnamevar
        desc " name parameter for mytype"
      end
    end
would give this result before the fix:

    [root@syqs5cacqx016cd type]# puppet describe mytype

    mytype
    ======
    .
    **this part should not
    **this part should not
    7890123456789012345678901234567890123456789012345678901234567890
    nick@magpie.puppetlabs.lan
    **this part should not repeat!**

    Parameters
    ----------

    - **name**
        name parameter for mytype
    [root@syqs5cacqx016cd type]#

    After the fix, the result is this :

    [root@syqs5cacqx016cd type]# puppet describe mytype

    mytype
    ======
    .
    1234567890123456789012345678901234567890123456789012345678901234567890123456
    7890123456789012345678901234567890 nick@magpie.puppetlabs.lan
    **this part should not repeat!**

    Parameters
    ----------

    - **name**
        name parameter for mytype
    [root@syqs5cacqx016cd type]#

### DIFF
--- a/lib/puppet/application/describe.rb
+++ b/lib/puppet/application/describe.rb
@@ -11,7 +11,7 @@ class Formatter
     work = (opts[:scrub] ? scrub(txt) : txt)
     indent = (opts[:indent] ? opts[:indent] : 0)
     textLen = @width - indent
-    patt = Regexp.new("^(.{0,#{textLen}})[ \n]")
+    patt = Regexp.new("\\A(.{0,#{textLen}})[ \n]")
     prefix = " " * indent
 
     res = []
--- a/spec/unit/application/describe_spec.rb
+++ b/spec/unit/application/describe_spec.rb
@@ -76,4 +76,23 @@ describe Puppet::Application::Describe do
       @describe.run_command
     end
   end
+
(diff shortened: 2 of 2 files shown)


joshcooper:
@seangriff could you break up the commit summary and description? Also what was the behavior previous when there was a long line?

seangriff:
@joshcooper : Is there anything else you think this PR needs before merging?

PR #5702, opened 2017-03-10
(PUP-6405) Use `pip freeze --all`

SamK:
https://tickets.puppetlabs.com/browse/PUP-6405

This pull request adds management of the "core" pip packages. The current special case only handles the pip package while this pull request will take care of all the others.

I would feel better if I receive confirmations about a few topics:

1. Someone double check the unit tests: I am not experienced in Ruby.
1. Make sure version 8.1.0 of pip is really the one that implemented the `freeze --all` option: the jira issue explains more why I am confused about this.

### DIFF
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -44,7 +44,8 @@ Puppet::Type.type(:package).provide :pip,
 
     # Pip can also upgrade pip, but it's not listed in freeze so need to special case it
     # Pip list would also show pip installed version, but "pip list" doesn't exist for older versions of pip (E.G v1.0)
-    if version = self.pip_version
+    # Not needed when "pip freeze --all" is available
+    if Puppet::Util::Package.versioncmp(self.pip_version, '8.1.0') == -1 && version = self.pip_version
       packages << new({:ensure => version, :name => File.basename(pip_cmd), :provider => name})
     end
 
--- a/lib/puppet/provider/package/pip.rb
+++ b/lib/puppet/provider/package/pip.rb
@@ -31,7 +31,11 @@ Puppet::Type.type(:package).provide :pip,
     packages = []
     pip_cmd = self.pip_cmd
     return [] unless pip_cmd
-    execpipe "#{pip_cmd} freeze" do |process|
+    command = [pip_cmd, 'freeze']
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/package/pip.rb
```diff
@@ -31,7 +31,11 @@ def self.instances
     packages = []
     pip_cmd = self.pip_cmd
     return [] unless pip_cmd
-    execpipe "#{pip_cmd} freeze" do |process|
+    command = ["#{pip_cmd}", 'freeze']
```
binford2k: Don't need to interpolate a string here. Just use `command = [pip_cmd, 'freeze']`

SamK: Can I rebase or should I create a new commit for this change?

Magisus: You can rebase, we generally do rebasing and force-pushing to forks to help keep the history cleaner.

SamK: Suggestion by binford2k has been implemented.

lib/puppet/provider/package/pip.rb
```diff
@@ -31,7 +31,11 @@ def self.instances
     packages = []
     pip_cmd = self.pip_cmd
     return [] unless pip_cmd
-    execpipe "#{pip_cmd} freeze" do |process|
+    command = [pip_cmd, 'freeze']
+    if Puppet::Util::Package.versioncmp(self.pip_version, '8.1.0') >= 0 # a >= b
+        command << '--all'
```
MosesMendoza: Minor nitpick - can you adjust indentation to two spaces rather than four?

SamK: Done.

SamK: OK I'm lost. After I happily implemented the minor nitpick, the Appveyor says there is a conflict but does not specify against witch branch there is a conflict. This is pretty confusing.

Magisus: It will be against the base branch, in this case master. Looks like the pip spec has changed, probably due to us merging https://github.com/puppetlabs/puppet/pull/5903 this morning.

SamK: I rebased on master, resolved the conflict and updated my commit to reflect the same changes as #5903.

PR #5704, opened 2017-03-10
(PUP-1334) overwrite corrupted contents_file in filebucket

MosesMendoza:
If Puppet receives a file to back up in filebucket that has the same checksum
as an existing backed up file, Puppet will detect this, validate that the
files are indeed identical by comparing size and content, and if so use the
existing backup for the new file.

If Puppet detects that the existing backup and the new file are NOT the same,
it will fail with an error "Got passed new contents for sum <shared checksum
value>". Originally, this somewhat opaque error was only expected to only be
raised in the event of an sum collision, which was expected to be very unlikely
and infrequent.

However, in practice, a filebucket contents_file can accidentally vary from its
expected state (and check sum value) a number of ways - for example, if puppet
unexpectedly fails during the backup write but has the file handle open, or if
the contents_file becomes corrupt on disk after write.

Prior to this commit, if this occurred, the same collision error was raised, as
Puppet never validated that the contents_file actually did, in fact, match the
expected check sum value. In this case, the error is both confusing and cannot
be recovered from without manually removing the contents_file (the path to
which is never mentioned).

With this commit, we:

* Add the ability to validate that the contents_file matches the expected checksum
* Since this does a checksum calculation which can be expensive, we only
  perform this check if we have received a new file with a checksum value that
  matches an existing backup but has a different size/contents (the previous
  failure case)
* If the existing backup does indeed have the expected checksum, we
  have more likely found a true sum collision - in this case we still fail, but
  offer a more explicit message to the user with the path to the existing
  backup contents_file.
* If the existing backup does not have the expected checksum, we assume it is
  corrupt, issue a warning to the user, and overwrite it with the new backup
  (not failing).

To accomplish this without a mess of nested conditionals, this commit added to
and refactored private methods within Puppet::FileBucket::File to be more
intuitive with their actions/return values.
* #verify_identical_file! is updated to simply return true/false for whether
  the two files are identical, rather than returning nil if identical and
  printing a message and raising an exception if not. Since it no longer raises
  an exception as a code path, it is renamed to #verify_identical_file
* #contents_file_matches_checksum? and #copy_bucket_file_to_contents_file are
  broken out as their own private utility methods to reduce the havoc inside #save_to_disk
* debug/warning output and exception raising is centralized in #save_to_disk
  rather than spread out amongst utility methods.
* yard tags are added and corrected for new/existing methods

### DIFF
--- a/lib/puppet/indirector/file_bucket_file/file.rb
+++ b/lib/puppet/indirector/file_bucket_file/file.rb
@@ -133,9 +133,17 @@ module Puppet::FileBucketFile
       return false
     end
 
-    # @param contents_file [Object] Opaque file path
-    # @param paths_file [Object] Opaque file path
-    #
+    # @param bucket_file [Puppet::FileBucket::File] IO object representing
+    #   content to back up
+    # @param files_original_path [String] Path to original source file on disk
+    # @param contents_file [Pathname] Opaque file path to intended backup
+    #   location
+    # @param paths_file [Pathname] Opaque file path to file containing source
+    #   file paths on disk
+    # @return [void]
+    # @raise [Puppet::FileBucket::BucketError] on possible sum collision between
+    #   existing and new backup
+    # @api private
     def save_to_disk(bucket_file, files_original_path, contents_file, paths_file)
       Puppet::Util.withumask(0007) do
         unless Puppet::FileSystem.dir_exist?(paths_file)
@@ -148,15 +156,28 @@ module Puppet::FileBucketFile
         # same way when the bytes are decoded as UTF-8, continue using system encoding
         Puppet::FileSystem.exclusive_open(paths_file, 0640, 'a+:external') do |f|
           if Puppet::FileSystem.exist?(contents_file)
-            verify_identical_file!(contents_file, bucket_file)
-            Puppet::FileSystem.touch(contents_file)
-          else
-            Puppet::FileSystem.open(contents_file, 0440, 'wb') do |of|
-              # PUP-1044 writes all of the contents
-              bucket_file.stream() do |src|
-                FileUtils.copy_stream(src, of)
-              end
+            if verify_identical_file(contents_file, bucket_file)
+              Puppet.info "FileBucket got a duplicate file #{bucket_file.checksum}"
+              Puppet::FileSystem.touch(contents_file)
+            elsif contents_file_matches_checksum?(contents_file, bucket_file.checksum_data, bucket_file.checksum_type)
+              # If the contents or sizes don't match, but the checksum does,
+              # then we've found a conflict (potential hash collision).
+              # Unlikely, but quite bad. Don't remove the file in case it's
+              # needed, but ask the user to validate.
+              # Note: Don't print the full path to the bucket file in the
+              # exception to avoid disclosing file system layout on server.
+              Puppet.err(_("Unable to verify existing FileBucket backup at '%{path}'.") % { path: contents_file.to_path })
+              raise Puppet::FileBucket::BucketError, _("Existing backup and new file have different content but same checksum, %{value}. Verify existing backup and remove if incorrect.") %
+                { value: bucket_file.checksum }
+            else
+              # PUP-1334 If the contents_file exists but does not match its
+              # checksum, our backup has been corrupted. Warn about overwriting
+              # it, and proceed with new backup.
+              Puppet.warning(_("Existing backup does not match its expected sum, %{sum}. Overwriting corrupted backup.") % { sum: bucket_file.checksum })
+              copy_bucket_file_to_contents_file(contents_file, bucket_file)
             end
+          else
+            copy_bucket_file_to_contents_file(contents_file, bucket_file)
           end
 
           unless path_match(f, files_original_path)
@@ -189,19 +210,45 @@ module Puppet::FileBucketFile
       Puppet::FileSystem.pathname(subfile ? ::File.join(basedir, subfile) : basedir)
     end
 
-    # @param contents_file [Object] Opaque file path
-    # @param bucket_file [IO]
-    def verify_identical_file!(contents_file, bucket_file)
-      if bucket_file.size == Puppet::FileSystem.size(contents_file)
-        if bucket_file.stream() {|s| Puppet::FileSystem.compare_stream(contents_file, s) }
-          Puppet.info "FileBucket got a duplicate file #{bucket_file.checksum}"
-          return
+    # @param contents_file [Pathname] Opaque file path to intended backup
+    #   location
+    # @param bucket_file [Puppet::FileBucket::File] IO object representing
+    #   content to back up
+    # @return [Boolean] whether the data in contents_file is of the same size
+    #   and content as that in the bucket_file
+    # @api private
+    def verify_identical_file(contents_file, bucket_file)
+      (bucket_file.size == Puppet::FileSystem.size(contents_file)) &&
+        (bucket_file.stream() {|s| Puppet::FileSystem.compare_stream(contents_file, s) })
+    end
+
+    # @param contents_file [Pathname] Opaque file path to intended backup
+    #   location
+    # @param expected_checksum_data [String] expected value of checksum of type
+    #   checksum_type
+    # @param checksum_type [String] type of check sum of checksum_data, ie "md5"
+    # @return [Boolean] whether the checksum of the contents_file matches the
+    #   supplied checksum
+    # @api private
+    def contents_file_matches_checksum?(contents_file, expected_checksum_data, checksum_type)
+      contents_file_checksum_data = Puppet::Util::Checksums.method(:"#{checksum_type}_file").call(contents_file.to_path)
+      contents_file_checksum_data == expected_checksum_data
+    end
+
+    # @param contents_file [Pathname] Opaque file path to intended backup
+    #   location
+    # @param bucket_file [Puppet::FileBucket::File] IO object representing
+    #   content to back up
+    # @return [void]
+    # @api private
+    def copy_bucket_file_to_contents_file(contents_file, bucket_file)
+      Puppet::FileSystem.open(contents_file, 0440, 'wb') do |of|
+        # PUP-1044 writes all of the contents
+        bucket_file.stream() do |src|
+          FileUtils.copy_stream(src, of)
         end
       end
-
-      # If the contents or sizes don't match, then we've found a conflict.
-      # Unlikely, but quite bad.
-      raise Puppet::FileBucket::BucketError, "Got passed new contents for sum #{bucket_file.checksum}"
     end
+
   end
 end
--- a/spec/unit/indirector/file_bucket_file/file_spec.rb
+++ b/spec/unit/indirector/file_bucket_file/file_spec.rb
@@ -52,11 +52,42 @@ describe Puppet::FileBucketFile::File, :uses_checksums => true do
                               0x2f94cc40,0x15a12deb,0xdc15f4a3,0x490786bb,
                               0x6d658673,0xa4341f7d,0x8fd75920,0xefd18d5a].pack("I" * 16)
 
-        save_bucket_file(first_contents, "/foo/bar")
+        checksum_value = save_bucket_file(first_contents, "/foo/bar")
 
+        # We expect Puppet to log an error with the path to the file
+        Puppet.expects(:err).with(regexp_matches(/Unable to verify existing FileBucket backup at '#{Puppet[:bucketdir]}.*#{checksum_value}\/contents'/))
+
+        # But the exception should not contain it
         expect do
           save_bucket_file(collision_contents, "/foo/bar")
-        end.to raise_error(Puppet::FileBucket::BucketError, /Got passed new contents/)
+        end.to raise_error(Puppet::FileBucket::BucketError, /\AExisting backup and new file have different content but same checksum, {md5}#{checksum_value}\. Verify existing backup and remove if incorrect\.\Z/)
+      end
+
+      # See PUP-1334
+      context "when the contents file exists but is corrupted and does not match the expected checksum" do
+        let(:original_contents) { "a file that will get corrupted" }
+        let(:bucket_file) { Puppet::FileBucket::File.new(original_contents) }
+        let(:contents_file) { "#{Puppet[:bucketdir]}/8/e/6/4/f/8/5/d/8e64f85dd54a412f65edabcafe44d491/contents" }
+
+        before(:each) do
+          # Ensure we're starting with a clean slate - no pre-existing backup
+          Puppet::FileSystem.unlink(contents_file) if Puppet::FileSystem.exist?(contents_file)
+          # Create initial "correct" backup
+          Puppet::FileBucket::File.indirection.save(bucket_file)
+          # Modify the contents file so that it no longer matches the SHA, simulating a corrupt backup
+          Puppet::FileSystem.unlink(contents_file) # bucket_files are read-only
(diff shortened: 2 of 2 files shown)


McdonaldSeanp:
👍

lib/puppet/indirector/file_bucket_file/file.rb
```diff
@@ -148,15 +156,28 @@ def save_to_disk(bucket_file, files_original_path, contents_file, paths_file)
...
+            elsif contents_file_matches_checksum?(contents_file, bucket_file.checksum_data, bucket_file.checksum_type)
+              # If the contents or sizes don't match, but the checksum does,
+              # then we've found a conflict (potential hash collision).
+              # Unlikely, but quite bad. Don't remove the file in case it's
+              # needed, but ask the user to validate.
+              # Note: Don't print the full path to the bucket file to avoid
+              # disclosing file system layout on server. Remove bucket_path from
+              # beginning of string.
+              raise Puppet::FileBucket::BucketError, _("Existing backup and new file have different content but same checksum, %{value}. Verify file '%{path}' inside the FileBucket and remove if incorrect.") %
+                { value: bucket_file.checksum, path: strip_bucket_path(contents_file, bucket_file) }
```
joshcooper: What about logging a message with the full path using `Puppet.error` (so we're not disclosing any file info) and raise an exception with only the `sum`, so that we don't need to `strip_bucket_path`?

spec/unit/indirector/file_bucket_file/file_spec.rb
```diff
@@ -53,10 +53,36 @@ def save_bucket_file(contents, path = "/who_cares")
...
           save_bucket_file(collision_contents, "/foo/bar")
-        end.to raise_error(Puppet::FileBucket::BucketError, /Got passed new contents/)
+        end.to raise_error(Puppet::FileBucket::BucketError, /Existing backup and new file have different content but same checksum.*(?!#{Puppet[:bucketdir]})/)
+      end
+
+      # See PUP-1334
+      context "when the contents file exists but is corrupted and does not match the expected checksum" do
+        let(:original_contents) { "a file that will get corrupted" }
+        let(:bucket_file) { Puppet::FileBucket::File.new(original_contents) }
+        let(:contents_file) { "#{Puppet[:bucketdir]}/8/e/6/4/f/8/5/d/8e64f85dd54a412f65edabcafe44d491/contents" }
```
joshcooper: Is `Puppet[:bucketdir]` a tempdir so we're not leaving behind test files? I assume it's relative to `Puppet[:vardir]` which should be a tempdir.

Iristyle: Its `bucket` under `$vardir`:

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L1175-L1182

lib/puppet/indirector/file_bucket_file/file.rb
MosesMendoza: 👍

spec/unit/indirector/file_bucket_file/file_spec.rb
MosesMendoza: Yes, this should be getting cleaned up per https://github.com/puppetlabs/puppet/blob/master/spec/spec_helper.rb#L155-L156

lib/puppet/indirector/file_bucket_file/file.rb
MosesMendoza: @joshcooper updated

johnduarte:
Tested on Ubuntu-16.04 and Windows 2012 agents using released puppet-agent version 1.9.3 for expected failure states.
```
SHA=1.9.3 TEST_TARGET=windows2012r2-64a-ubuntu1604-64a be rake ci:test:aio TESTS=tests/ticket_1334_clientbucket_corrupted.rb
...
       Total Suite Time: 24.99 seconds
      Average Test Time: 24.99 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```

Sample of expected failure conditions
```
  * manage file again

    fzqn3estk37h1fz.delivery.puppetlabs.net (fzqn3estk37h1fz.delivery.puppetlabs.net) 00:32:51$ mktemp -t apply_manifest.pp.XXXXXX
      /tmp/apply_manifest.pp.7Rd1C0

    fzqn3estk37h1fz.delivery.puppetlabs.net (fzqn3estk37h1fz.delivery.puppetlabs.net) executed in 0.00 seconds
    localhost $ scp /tmp/beaker20170321-11140-1yjgbj1 fzqn3estk37h1fz.delivery.puppetlabs.net:/tmp/apply_manifest.pp.7Rd1C0 {:ignore => }

    fzqn3estk37h1fz.delivery.puppetlabs.net (fzqn3estk37h1fz.delivery.puppetlabs.net) 00:32:51$ puppet apply --verbose /tmp/apply_manifest.pp.7Rd1C0
      Notice: Compiled catalog for fzqn3estk37h1fz.delivery.puppetlabs.net in environment production in 0.10 seconds
      Info: Applying configuration version '1490056373'
      Info: Computing checksum on file /tmp/c99977file.mfJUWC
      Error: Could not back up /tmp/c99977file.mfJUWC: Got passed new contents for sum {md5}f014bc47f0a4ea375aafae4074d090cf
      Error: Could not back up /tmp/c99977file.mfJUWC: Got passed new contents for sum {md5}f014bc47f0a4ea375aafae4074d090cf
      Error: /Stage[main]/Main/File[/tmp/c99977file.mfJUWC]/content: change from {md5}f014bc47f0a4ea375aafae4074d090cf to {md5}e0b1a605610f4d3196be32721050ae0d failed: Could not back up /tmp/c99977file.mfJUWC: Got passed new contents for sum {md5}f014bc47f0a4ea375aafae4074d090cf

    fzqn3estk37h1fz.delivery.puppetlabs.net (fzqn3estk37h1fz.delivery.puppetlabs.net) executed in 1.94 seconds
    An assertion was expected to fail, and did. This is probably due to a known product bug, and is probably not a problem. Additional info: 'no stdrr' Failed assertion: '--- expected
    +++ actual
    @@ -1 +1,4 @@
    -""
    +"\e[1;31mError: Could not back up /tmp/c99977file.mfJUWC: Got passed new contents for sum {md5}f014bc47f0a4ea375aafae4074d090cf\e[0m
    +\e[1;31mError: Could not back up /tmp/c99977file.mfJUWC: Got passed new contents for sum {md5}f014bc47f0a4ea375aafae4074d090cf\e[0m
    +\e[1;31mError: /Stage[main]/Main/File[/tmp/c99977file.mfJUWC]/content: change from {md5}f014bc47f0a4ea375aafae4074d090cf to {md5}e0b1a605610f4d3196be32721050ae0d failed: Could not back up /tmp/c99977file.mfJUWC: Got passed new contents for sum {md5}f014bc47f0a4ea375aafae4074d090cf\e[0m
    +"
    '

    fzqn3estk37h1fz.delivery.puppetlabs.net (fzqn3estk37h1fz.delivery.puppetlabs.net) 00:32:53$ cat /tmp/c99977file.mfJUWC
      unmanaged

    fzqn3estk37h1fz.delivery.puppetlabs.net (fzqn3estk37h1fz.delivery.puppetlabs.net) executed in 0.00 seconds
    An assertion was expected to fail, and did. This is probably due to a known product bug, and is probably not a problem. Additional info: 'file managed' Failed assertion: '--- expected
    +++ actual
    @@ -1,2 +1,2 @@
    -"managed
    +"unmanaged
     "
    '
```

lib/puppet/indirector/file_bucket_file/file.rb
```diff
@@ -189,19 +210,45 @@ def path_for(bucket_path, digest, subfile = nil)
...
+    #   location
+    # @param bucket_file [Puppet::FileBucket::File] IO object representing
+    #   content to back up
+    # @return [void]
+    # @api private
+    def copy_bucket_file_to_contents_file(contents_file, bucket_file)
+       Puppet::FileSystem.open(contents_file, 0440, 'wb') do |of|
+        # PUP-1044 writes all of the contents
+        bucket_file.stream() do |src|
+          FileUtils.copy_stream(src, of)
```
joshcooper: I think we want to use `Puppet::Util.replace_file` to atomically write the file. Otherwise if two agents try to backup the same file, then it's possible they will have overlapping writes, leading to corrupt file contents.

MosesMendoza: I was initially worried about that possibility as well. I was just more worried about changing more behavior than just fixing the empty filebucket problem, and accidentally introducing some unrelated regression - this is the same behavior that existed before my PR: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/indirector/file_bucket_file/file.rb#L154-L158. That said, when I looked at it before, I concluded we do actually protect against multiple simultaneous writes because we obtain a lock on the corresponding paths file before beginning the contents write: https://github.com/MosesMendoza/puppet/blob/c72777242b8571813bce4a2c55a9f1e78b4a4fbf/lib/puppet/indirector/file_bucket_file/file.rb#L157. I think through that exclusive lock is the only path by which we'd be writing to the contents file. However, it was much clearer that this is the case (and thus probably safer) before I broke this out into a separate method to reduce duplication.

However, it occurs to me now - I think that ruby's `::File::LOCK_EX` (used by our `Puppet::FileSystem::FileImpl#exclusive_open` implementation) allows you to restrict access to one process, which was safe in rack/passenger days, but I think Puppet Server rubies live in different threads in the same process.. so is this broken there?

joshcooper: >  we obtain a lock on the corresponding paths file

Ah yes, brings back memories https://github.com/puppetlabs/puppet/commit/8b0a4ec9cecb0c2ff0e79cd26966d279fecaa9c4

>  I think Puppet Server rubies live in different threads in the same process.. so is this broken there?

Yeah, I think you're right. If you believe the internet http://stackoverflow.com/a/9477276:

> In addition, flock locks don't 'stack', so if you try to acquire a lock you already hold, the flock call is a noop that returns immediately without blocking and without changing the lock state in any way. Since threads within a process share file descriptors, you can flock the file multiple times from different threads, and it won't block, as the lock is already held.

MosesMendoza: @joshcooper I filed https://tickets.puppetlabs.com/browse/PUP-7386 to separately tackle the atomicity issue

joshcooper: Ah the other reason for using `replace_file` is in case the server is stopped/killed in the process of writing the file.

MosesMendoza: +1. updated contents_file write to use `Puppet::Util::replace_file`

lib/puppet/indirector/file_bucket_file/file.rb
```diff
@@ -189,19 +210,45 @@ def path_for(bucket_path, digest, subfile = nil)
...
+    end
+
+    # @param contents_file [Pathname] Opaque file path to intended backup
+    #   location
+    # @param bucket_file [Puppet::FileBucket::File] IO object representing
+    #   content to back up
+    # @return [void]
+    # @api private
+    def copy_bucket_file_to_contents_file(contents_file, bucket_file)
+      Puppet::Util.replace_file(contents_file, 0440) do |of|
```
MosesMendoza: In this `Puppet::Util::replace_file` implementation, we no longer specify `wb` as our open mode option for the contents file. I think we should be OK as to whether we are still reading and writing files correctly in binary encoding, but I'm noting the thinking here for posterity.

`bucket_file.stream` here will either be doing a `FileContents#stream` or a `StringContents#stream`. `FileContents#stream` opens the file for reading in binary: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_bucket/file.rb#L128. `StringContents#stream` creates a StringIO object which preserves the encoding of the original string content (which I think is likely UTF-8). https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_bucket/file.rb#L98.

When we move to file writing inside the block, we're using `FileUtils.copy_stream`.. which is effectively an alias to `IO::copy_stream`. `IO::copy_stream` docs don't explicitly state how it handles encoding (https://ruby-doc.org/core-2.3.0/IO.html#method-c-copy_stream). However, [this bug](https://bugs.ruby-lang.org/issues/8767) implies that `IO::copy_stream` writes in binary mode. That fix was introduced in ruby 2.1.0 and backported to ruby 1.9.3. So I believe we're still good on encoding.

MosesMendoza: cc @Iristyle - can you validate my thinking?

Iristyle: That thinking seems legit based on my looking at the code.  

2 question that arise:

* Will `StringContents` always stream raw bytes of input string?  Can test with a variety of input strings on different Ruby versions with something like:

```ruby
FileUtils.touch('/tmp/foo')
Puppet::Util.replace_file('/tmp/foo', 0440) { |of| FileUtils.copy_stream(StringIO.new("\u1234"), of) }
```

* It looks like `replace_file` requires a file to exist first or else it bombs, so are we certain it's a complete drop-in replacement for the old code?

Iristyle: I was able to verify that the `StringIO.new` should be totally fine with respect to bytes being written properly to disk:

```ruby
kr = "\xFD\xF1".force_encoding(Encoding::EUC_KR)
Puppet::Util.replace_file('/tmp/foo', 0440) { |of| FileUtils.copy_stream(StringIO.new(kr), of) }

File.read('/tmp/foo').bytes
# => [253, 241]

File.read('/tmp/foo').force_encoding(Encoding::EUC_KR).encode(Encoding::UTF_8)
# => "希"
```

MosesMendoza: After investigation, it turns out that for the tempfile that it creates, `Puppet::Util::replace_file` will either

  a) use the mode of an existing target file as the mode of the tempfile
or
  b) use the supplied mode argument if the target file does not exist

But it chmods the tempfile to this mode _before_ yielding to the block defined by the caller.

This is unintuitive for a couple reasons. First, it makes more sense to me to have the supplied mode override the mode of an existing file (if I supply an explicit argument, it should be honored). Second, chmoding the tempfile before yielding means a read-only tempfile?

For the purposes of this PR, this then plays out thusly:

```ruby
Puppet::Util.replace_file('/tmp/foo', 0440) { |of| FileUtils.copy_stream(StringIO.new("\u1234"), of) }
```
If a `contents` file does not exist (which we expect to happen a substantial percentage of the time), `Puppet::Util::replace_file` will create a tempfile with a mode of 0440 (read only), and then supply this handle to the block. Since we can't actually write to this read-only tempfile, the copy fails.

If a `contents` file does exist (which will happen any time there is a duplicate entry, a substantial percentage also), it _should_ have a mode of 0440. So then `Puppet::Util::replace_file`, when seeking to overwrite the `content`s file with a new one, will create a tempfile with a mode of 0440, and we will fail to overwrite it.

I suppose we could wrap the call to `Puppet::Util::replace_file` with a preceding `chmod 0640` on the `contents` file and succeeding `chmod 0440` afterward, but this seems to be a dubious direction...what happens if we chmod 640 and then die during the write? It seems to me a better change would be to handle this use-case in `Puppet::Util::replace_file`.

I'd prefer not to hold up this PR for that so I'm inclined to remove the commit that adds replace_file from the PR.

MosesMendoza: https://tickets.puppetlabs.com/browse/PUP-7397 tracks updating this to use `replace_file`

PR #5705, opened 2017-03-11
(PUP-7283) Break up DefaultValidator

adrienthebo:
This pull request breaks up the DefaultValidator into separate validators with well defined responsibilities, and a factory method that will select the best validator available based on the present SSL files.


joshcooper:
👍  

Travis failed due to `(WIP)` in the commit summary but that's expected.

lib/puppet/ssl/validator/unauthenticated_validator.rb
```diff
@@ -0,0 +1,165 @@
...
+    peer_certs.any? do |peer_cert|
+      authz_certs.any? do |authz_cert|
+        peer_cert.verify(authz_cert.public_key)
+      end
+    end
+  end
+
+  protected
+
+  # Conditionally enable CRL checking for the SSL store if certificate
```
adrienthebo: This bit is not the greatest because it's a case where we're downgrading security. This is an improvement over the old implementation because previously we didn't even try to use the CRL until we had a host certificate, and this allows a user to pass a CRL in an out of band manner and have it be used. However any case where we potentially reduce our security is worth staring at for a few extra seconds, just in case.

lib/puppet/ssl/validator.rb
```diff
@@ -24,7 +24,42 @@ def self.no_validator()
   # @api public
   #
   def self.default_validator()
-    Puppet::SSL::Validator::DefaultValidator.new()
+    Puppet.deprecation_warning("Puppet::SSL::Validator.default_validator is deprecated. Use Puppet::SSL::Validator.cert_auth_validator if the host has a signed cert, or Puppet::SSL::Validator.best_validator if the host certificate signing status is unknown.")
```
MosesMendoza: Given this is now deprecated can/should we update its use in `Puppet::Network::HttpPool` ?

lib/puppet/ssl/validator/unauthenticated_validator.rb
lutter: It might be good to put something like this as a comment into the code to help future readers understand that (a) security downgrade might happen and (b) that it's intentional and the best we can do because of what you mentioned there.

lib/puppet/ssl/validator.rb
adrienthebo: Thanks for catching this; it turns out that this logging was breaking the appveyor builds as well. I've fixed it.

lib/puppet/ssl/validator/unauthenticated_validator.rb
adrienthebo: Comment added.

MosesMendoza:
can merge pending appveyor

PR #5708, opened 2017-03-13
ext/redhat/logrotate: really avoid 'service puppet reload' in favor o…

imz:
…f 'service puppet rotate' with the correct signal

It looks like there was an intention to introduce this fix in
793d713503d27600b4c64e4e0c331d6d7365e3c4
'(PUP-4918) Updated logrotate and init scripts',
where the "rotate" command was added to
ext/redhat/client.init specifically for this purpose.

This doesn't apply to ext/debian/puppet.logrotate because
ext/debian/puppet.init lacks a special command with this signal.

### DIFF
--- a/ext/redhat/logrotate
+++ b/ext/redhat/logrotate
@@ -15,7 +15,7 @@
   nocreate
   sharedscripts
   postrotate
-   ([ -x /etc/init.d/puppet ] && /etc/init.d/puppet reload > /dev/null 2>&1) ||
+   ([ -x /etc/init.d/puppet ] && /etc/init.d/puppet rotate > /dev/null 2>&1) ||
     ([ -x /usr/bin/systemctl ] && /usr/bin/systemctl kill -s USR2 puppet.service > /dev/null 2>&1) || true
   endscript
 }


branan:
This looks like a reasonable, quick change. Thanks!

Could you please file an appropriate ticket in Jira and update your commit message? That will fix the Travis failure. We use those commit tags as part of our release process.

In addition, I believe we would require the CLA to be signed for this (it's not "trivial" by our definitions).

MosesMendoza:
I'm going to amend the message to include PUP-4918, which this is a completer for. I also think this falls under the trivial patch exemption.

MosesMendoza:
thanks @imz. I've taken your contribution into https://github.com/puppetlabs/puppet/pull/6301.

MosesMendoza:
superseded by https://github.com/puppetlabs/puppet/pull/6301

imz:
Thanks @MosesMendoza ! Mentioning the old PUP is smarter than I could have done anyway. And I've been forgetting to make a copyright assignment for this trivial change, so I am glad this has been resolved now.

PR #5709, opened 2017-03-13
(PUP-7341) Fix crash related to regexp identified lookup_options

thallgren:
Before this commit, if lookup_options was identified using regular
expressions, the lookup would crash unless there was also at least one
lookup option that was identified with an explicit key. This commit
fixes the problem and adds a test for the case when only regexp's are
used.

### DIFF
--- a/spec/unit/functions/lookup_spec.rb
+++ b/spec/unit/functions/lookup_spec.rb
@@ -595,56 +595,59 @@ describe "The lookup function" do
           YAML
       end
 
+      let(:env_lookup_options) { <<-YAML.unindent }
+        lookup_options:
+          b:
+            merge: hash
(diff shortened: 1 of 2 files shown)


spec/unit/functions/lookup_spec.rb
```diff
@@ -705,6 +708,25 @@ def explain(key, options = {})
           expect { lookup('mod::a') }.to raise_error(Puppet::DataBinding::LookupError, /all lookup_options patterns must match a key starting with module name/)
         end
       end
+
+      context 'and there are no lookup options that do not use patterns' do
```
hlindberg: and "all lookup options use patterns" perhaps... to avoid the double negation. Not important enough to update the PR.

PR #5711, opened 2017-03-13
(maint) Allow use of beaker 3.x in acceptance tests

McdonaldSeanp:
in particular, which was added in beaker 3.10).


Magisus:
👍

PR #5713, opened 2017-03-13
(pup-7344) acceptance hiera5 config interpolated with various scoped and nested variables

er0ck:
@thallgren @hlindberg 
we used to have a partly unwritten rule that any regression should get an acceptance test.  i had written most of this one last week to reproduce the errors in PUP-7336.

this is partly covered here: https://github.com/puppetlabs/puppet/pull/5706/files#diff-2c12cbb4931cfd2102db8d688e73049eR367

i might be missing the coverage in unit tests for the other variable scoping and subkeys i've used here.  

do you have feelings on which of these should be acceptance tests, and which not?  keeping in mind we're trying to push as many tests closer to the developers' unit test feedback loop as possible.

### DIFF
--- a/acceptance/tests/lookup/config5_interpolation.rb
+++ b/acceptance/tests/lookup/config5_interpolation.rb
@@ -1,0 +1,119 @@
+test_name 'C99578: hiera5 lookup config with interpolated scoped nested variables' do
+  require 'puppet/acceptance/environment_utils.rb'
+  extend Puppet::Acceptance::EnvironmentUtils
+
+  app_type        = File.basename(__FILE__, '.*')
+  tmp_environment = mk_tmp_environment_with_teardown(master, app_type + '1')
+  fq_tmp_environmentpath  = "#{environmentpath}/#{tmp_environment}"
+
+  step "create environment hiera5.yaml and environment data" do
+
+    create_remote_file(master, "#{fq_tmp_environmentpath}/hiera.yaml", <<-HIERA)
+---
+version: 5
+defaults:
+  datadir: 'hieradata'
+  data_hash: yaml_data
+hierarchy:
+  - name: "Global settings"
+    path: "global.yaml"
+  - name: "Role specific settings"
+    paths:
+      - "roles/%{::roles.0}.yaml"
+  - name: "Other Role specific settings"
+    paths:
+      - "roles/%{roles2.0}.yaml"
+  - name: "scoped variable"
+    paths:
+      - "roles/%{::myclass::myvar.0}.yaml"
+  - name: "nested hash variable"
+    paths:
+      - "roles/%{::hash_array.key1.0}.yaml"
+    HIERA
+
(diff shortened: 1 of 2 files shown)


acceptance/tests/lookup/config5_interpolation.rb
```diff
@@ -0,0 +1,119 @@
...
+$data2 = lookup('data', Array[String], 'unique')
+notify{"data2: ${data2}":}
+$hiera_array_data2 = hiera_array('data')
+notify{"hiera_array_data2: ${hiera_array_data2}":}
+    SITE
+
+    on(master, "chmod -R 775 #{fq_tmp_environmentpath}")
+  end
+
+  with_puppet_running_on(master,{}) do
```
thallgren: I get errors when running this test on Ubuntu 14.04 (it installs Ruby 1.9.3). It passes when using Ubuntu 16.10.
```
Permission denied - /etc/puppetlabs/puppet/hiera.yaml at /etc/puppetlabs/code/environments/config5_interpolation1_u3jifxre/manifests/site.pp:8:10
```

er0ck: it works for me.  actually i think i developed this against ubuntu14.04.
```
   "HOSTS": {
        "kzbi51xkewniib6.delivery.puppetlabs.net": {
            "pe_dir": null,
            "pe_ver": null,
            "pe_upgrade_dir": null,
            "pe_upgrade_ver": null,
            "hypervisor": "vmpooler",
            "platform": "el-7-x86_64",
            "template": "redhat-7-x86_64",
            "roles": [
                "master",
                "default"
            ],
            "host_tags": {},
            "user": "root",
            "group": "puppet",
            "pathseparator": ":",
            "vmhostname": "kzbi51xkewniib6.delivery.puppetlabs.net",
            "puppetbindir": "/opt/puppetlabs/bin",
            "privatebindir": "/opt/puppetlabs/puppet/bin",
            "distmoduledir": "/etc/puppetlabs/code/modules",
            "sitemoduledir": "/opt/puppetlabs/puppet/modules"
        },
        "swf0gyhdp0064vk.delivery.puppetlabs.net": {
            "pe_dir": null,
            "pe_ver": null,
            "pe_upgrade_dir": null,
            "pe_upgrade_ver": null,
            "hypervisor": "vmpooler",
            "platform": "ubuntu-14.04-amd64",
            "template": "ubuntu-1404-x86_64",
```

```
tests/lookup/config5_interpolation.rb passed in 18.18 seconds
      Test Suite: tests @ 2017-04-05 16:27:35 +0000

       Total Suite Time: 18.18 seconds
      Average Test Time: 18.18 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1

      - Specific Test Case Status -
```

thallgren: Well, perhaps Mercury was in line with the moon when I tried it :-). I cannot reproduce now.

PR #5714, opened 2017-03-13
(PUP-7340) Fix context_spec failure

Iristyle:
- Environment timeouts were configured improperly for this test.
   They were passing when run in parallel because the
   :environment_timeout setting is global state, and tests on
   different threads may interfere with one another

### DIFF
--- a/spec/unit/pops/lookup/context_spec.rb
+++ b/spec/unit/pops/lookup/context_spec.rb
@@ -171,10 +171,6 @@ describe 'Puppet::Pops::Lookup::Context' do
 
       context 'and multiple compilations' do
 
-        after(:each) do
-          Puppet.lookup(:environments).clear_all
-        end
-
         it 'will reuse cached_file_data and not call block again' do
           Puppet.settings[:environment_timeout] = 'unlimited'
 
--- a/spec/unit/pops/lookup/context_spec.rb
+++ b/spec/unit/pops/lookup/context_spec.rb
@@ -171,13 +171,12 @@ describe 'Puppet::Pops::Lookup::Context' do
 
       context 'and multiple compilations' do
 
-        before(:each) { Puppet.settings[:environment_timeout] = 'unlimited' }
         after(:each) do
-          Puppet.settings[:environment_timeout] = 0
           Puppet.lookup(:environments).clear_all
         end
 
         it 'will reuse cached_file_data and not call block again' do
+          Puppet.settings[:environment_timeout] = 'unlimited'
 
           code1 = <<-PUPPET.unindent
             $ctx = Puppet::LookupContext.new(nil)
--- a/lib/puppet/test/test_helper.rb
+++ b/lib/puppet/test/test_helper.rb
@@ -160,6 +160,8 @@ module Puppet::Test
 
       Puppet.settings.send(:clear_everything_for_tests)
 
+      Puppet.lookup(:environments).clear_all
+
       Puppet::Util::Storage.clear
       Puppet::Util::ExecutionStub.reset
 


spec/unit/pops/lookup/context_spec.rb
```diff
@@ -171,13 +171,13 @@ module Lookup
 
       context 'and multiple compilations' do
 
-        before(:each) { Puppet.settings[:environment_timeout] = 'unlimited' }
         after(:each) do
           Puppet.settings[:environment_timeout] = 0
```
joshcooper: FWIW, it shouldn't be necessary to reset the env timeout at the end of the test as we create a new settings object for each test (to avoid tests leaking state)

Iristyle:
The failing test was being influenced by the `before` in the test group it shared, by `Puppet.settings[:environment_timeout] = 'unlimited'`

To see the change to the data file, the default `environment_timeout` must be set to ensure changes are picked up immediately.

spec/unit/pops/lookup/context_spec.rb
```diff
@@ -171,13 +171,13 @@ module Lookup
...
           Puppet.lookup(:environments).clear_all
```
joshcooper: Also if we should be cleaning up `environments` global state for each test, then we probably should put that in https://github.com/puppetlabs/puppet/blob/master/lib/puppet/test/test_helper.rb#L151

spec/unit/pops/lookup/context_spec.rb
Iristyle: Validated that is correct.  Having the code here has the positive of being more explicit though.

I'll update PR to remove that

joshcooper:
@Iristyle I forgot but I filed a ticket last week about this: https://tickets.puppetlabs.com/browse/PUP-7340

spec/unit/pops/lookup/context_spec.rb
```diff
@@ -171,13 +171,8 @@ module Lookup
...
       context 'and multiple compilations' do
 
-        before(:each) { Puppet.settings[:environment_timeout] = 'unlimited' }
-        after(:each) do
-          Puppet.settings[:environment_timeout] = 0
-          Puppet.lookup(:environments).clear_all
-        end
-
         it 'will reuse cached_file_data and not call block again' do
+          Puppet.settings[:environment_timeout] = 'unlimited'
```
thallgren: Assuming that this setting doesn't survive between tests, then this needs to happen in both tests in this context.

Iristyle: Having the second test default back to `Puppet.settings[:environment_timeout] = 0` instead is what allows the test to pass.  Otherwise, the cache is never refreshed / the file write isn't observed.

If you're sure that the file write should be seen, even if cache doesn't expire, then there's a bug there I believe.

thallgren: Please note that the test executed OK when the `environment_timeout` was set to "unlimited" before each test in the context.
Setting `environment_timeout` to 0 effectively destroys the assertion that the test is trying to make. With that setting, the cache is cleared between each compile so of course the test passes. The file is always loaded! The test is there to assert that the cache is refreshed due to a file change, not due to environment recreation. If that doesn't work, then the code that is tested is broken. The test executed OK before though, so I don't think that's the case.

thallgren: I tried your PR. It doesn't fail for me when I add the line to the second test as well.

Iristyle: Ok.. I had a feeling that my modification invalidated what you were trying to assert - but the code wasn't totally clear to me.  I also didn't realize that the file change reload and the environment reload were separate circumstances.  I thought the test might have been generating a false positive success for you because you were running the specs differently - but my assumption there seems incorrect.

Are you running on OSX?  This test failed for @joshcooper, @MosesMendoza and myself locally, which is why we were making adjustments. We're all running OSX - not sure if you're running Linux (the test seemingly passes in TravisCI in its prior form)?

joshcooper: I can repro running the test in isolation (ruby 2.2.5 on osx 10.11):

```bundle exec rspec spec/unit/pops/lookup/context_spec.rb
.................F...

Failures:

  1) Puppet::Pops::Lookup::Context an instance with cached_file_data and multiple compilations will invalidate cache if file changes
     Failure/Error: expect(logs).to eql(["{parsed => a: value a\n}", "{parsed => b: value b\n}"])

       expected: ["{parsed => a: value a\n}", "{parsed => b: value b\n}"]
            got: ["{parsed => a: value a\n}", "{parsed => a: value a\n}"]

       (compared using eql?)

       Diff:
       @@ -1,3 +1,3 @@
        {parsed => a: value a\n}
       -{parsed => b: value b\n}
       +{parsed => a: value a\n}

     # ./spec/unit/pops/lookup/context_spec.rb:228:in `block (5 levels) in <module:Lookup>'

Finished in 1.18 seconds (files took 1.31 seconds to load)
21 examples, 1 failure
```

thallgren:
Interesting. It fails on MacOS and succeeds on Fedora 25. That proves two
things. The line really must be there and the test revealed a platform
specific bug. I'll investigate why it doesn't work on osx.

spec/unit/pops/lookup/context_spec.rb
thallgren: Interesting. It fails on MacOS and succeeds on Fedora 25. That proves two
things. The line really must be there and the test revealed a platform
specific bug. I'll investigate why it doesn't work on osx.

thallgren: The plot thickens. Setting a debug breakpoint in the cached_file_data makes the test pass on osx.

thallgren: Found the problem. The time resolution is not high enough to detect the change between the two compiles and neither the inode nor the size of the file changes. I'll push a PR that changes the content so that the size differs.

thallgren:
Superseeded by [PR-5718](/puppetlabs/puppet/pull/5718).

PR #5716, opened 2017-03-14
(PUP-7350) Update minitar to 0.6.1

McdonaldSeanp:
We have updated the minitar component in the agent, and subsequently we will
need to update minitar for the puppet gem as well

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -41,7 +41,7 @@ gem_platform_dependencies:
       # Use of win32-security is deprecated
       win32-security: '= 0.2.5'
       win32-service: '= 0.8.8'
-      minitar: '~> 0.5.4'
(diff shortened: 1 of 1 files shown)


Magisus:
Are we okay having a PA ticket in a Puppet commit?

McdonaldSeanp:
Ah, probably not.

McdonaldSeanp:
@Magisus updated

Magisus:
👍

Iristyle:
So looks like we can close #5624 then

Iristyle:
Also - could someone file the ticket for vendoring `minitar` within our Solaris builds if that hasn't already been done?

Iristyle:
[PA-1019](https://tickets.puppetlabs.com/browse/PA-1019) has been filed to address the vendoring on Solaris

PR #5717, opened 2017-03-14
(maint) Address incorrect argument passing within Puppet

Iristyle:
These were discovered with the aid of RubyMines `Incorrect call argument count` inspection, which is just one of many inspections it supports in code analysis..  There were a lot of false positives that turned up, but these seem legit.

### DIFF
--- a/lib/puppet/util/instance_loader.rb
+++ b/lib/puppet/util/instance_loader.rb
@@ -41,7 +41,7 @@ module Puppet::Util::InstanceLoader
 
     # Use this method so they all get loaded
     loaded_instances(type).sort { |a,b| a.to_s <=> b.to_s }.each do |name|
-      mod = self.loaded_instance(name)
+      mod = self.loaded_instance(type, name)
       docs << "#{name}\n#{"-" * name.to_s.length}\n"
 
       docs << Puppet::Util::Docs.scrub(mod.doc) << "\n\n"
--- a/ext/upload_facts.rb
+++ b/ext/upload_facts.rb
@@ -102,7 +102,7 @@ HELP
     files.each do |file|
       facts = YAML.load_file(file)
 
-      request = Puppet::Indirector::Request.new(:facts, :save, facts)
+      request = Puppet::Indirector::Request.new(:facts, :save, facts, nil)
 
       # The terminus warns for us if we fail.
       if terminus.save(request)
--- a/lib/puppet/application.rb
+++ b/lib/puppet/application.rb
@@ -311,13 +311,13 @@ class Application
   # handling of this.
   option("--version", "-V") do |arg|
     puts "#{Puppet.version}"
-    exit
+    exit(0)
   end
 
   # Every app responds to --help
   option("--help", "-h") do |v|
     puts help
-    exit
+    exit(0)
   end
 
   def app_defaults()
--- a/lib/puppet/application/face_base.rb
+++ b/lib/puppet/application/face_base.rb
@@ -24,7 +24,7 @@ class Puppet::Application::FaceBase < Puppet::Application
     else
       puts Puppet::Face[:help, :current].help(face.name)
     end
-    exit
+    exit(0)
   end
 
   attr_accessor :face, :action, :type, :arguments, :render_as
--- a/lib/puppet/application/resource.rb
+++ b/lib/puppet/application/resource.rb
@@ -21,7 +21,7 @@ class Puppet::Application::Resource < Puppet::Application
       types << t.name.to_s
     end
     puts types.sort
-    exit
+    exit(0)
   end
 
   option("--param PARAM", "-p") do |arg|
--- a/lib/puppet/module_tool/dependency.rb
+++ b/lib/puppet/module_tool/dependency.rb
@@ -16,7 +16,7 @@ module Puppet::ModuleTool
       # TODO: add error checking, the next line raises ArgumentError when +full_module_name+ is invalid
       @username, @name = Puppet::ModuleTool.username_and_modname_from(full_module_name)
       @version_requirement = version_requirement
-      @repository = repository ? Puppet::Forge::Repository.new(repository) : nil
+      @repository = repository ? Puppet::Forge::Repository.new(repository, nil) : nil
     end
 
     # We override Object's ==, eql, and hash so we can more easily find identical
--- a/lib/puppet/network/http/compression.rb
+++ b/lib/puppet/network/http/compression.rb
@@ -32,7 +32,7 @@ module Puppet::Network::HTTP::Compression
     end
 
     def uncompress(response)
-      raise Net::HTTPError.new("No block passed") unless block_given?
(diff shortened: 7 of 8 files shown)


Iristyle:
@thallgren / @hlindberg would you mind double-checking 4ff80a6?  Thanks!

lib/puppet/bindings.rb
```diff
@@ -129,7 +129,7 @@ def [](name)
     end
 
     def has_name?(name)
-      @named_bindings.has_key?
+      @named_bindings.has_key?(name)
```
hlindberg: As I understand it, this code is slated for removal in Puppet 5.0.0 and merging this at this point just makes that work having to jump through one extra hoop.

lib/puppet/pops/binder/producers.rb
```diff
@@ -456,7 +456,7 @@ def producer(scope)
     # @api private
     #
     def internal_produce(scope, *args)
-      producer() unless value_producer
+      producer(scope) unless value_producer
```
hlindberg: Same here - slated for removal in Puppet 5.0.0

joshcooper:
@Iristyle nice! It looks like some of things need unit tests, e.g. match vs equal, and some of this must be dead code.

lib/puppet/bindings.rb
thallgren: Even if this certainly is wrong, I would prefer if it was left as is. Reason being that it's not used and we are removing the binder on master. Fixing this will create conflicts down the line.

lib/puppet/pops/binder/producers.rb
thallgren: Same comment as above. This is not used and fixing will create conflicts.

Iristyle:
Yeah, figured I'd get this up and get feedback from language before I go forward on tests, given some of the POPS bits might have been superseded in newer branches.

lib/puppet/pops/evaluator/compare_operator.rb
```diff
@@ -222,7 +222,7 @@ def match_Hash(hash, left, scope)
 
   def match_Symbol(symbol, left, scope)
     return true if symbol == :default
-    equals(left, default, scope)
```
Iristyle: This is still present in master at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/evaluator/compare_operator.rb#L248

lib/puppet/pops/binder/producers.rb
```diff
@@ -456,7 +456,7 @@ def producer(scope)
     # @api private
     #
     def internal_produce(scope, *args)
-      producer() unless value_producer
```
Iristyle: Still present in master: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/pops/binder/producers.rb#L459

thallgren: Yes, but not in the PR for PUP-7224.

lib/puppet/pops/evaluator/compare_operator.rb
```diff
@@ -222,7 +222,7 @@ def match_Hash(hash, left, scope)
...
+    match(left, default, scope)
```
hlindberg: I think the code is broken in other ways - there does not seem to be a value associated with `default` local variable - there is no such method either. I think this is a hidden bug because we never run into the situation that the evaluator ever matches/compares against any other symbol than `:default`. If it were, it should probably do an `equal`, but then doing an equals of `symbol` and `left`.

The only way to trigger the faulty original logic would be if there is faulty AST trying to compare some other symbol than `:default`.

I think the correct change here is to change line 225 to `equals(symbol, left, scope)` as that would make it correctly handle the `:undef` symbol that could be to only possible other relative thing to compare against. 

Ping @thallgren for second opinion.

lib/puppet/bindings.rb
Iristyle: Still present in master at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/bindings.rb#L132

hlindberg:
Indeed nice catch of some hidden problems. I think it is best to not fix the two problems related to `binder` - they will be fixed soon as I get to merge the PRs Thomas mentioned.
The evaluator issue is not that of `match` vs `equals` (it should call equals, but not with a non existing variable). The unit test to add would be to make sure legacy compare against `:undef` works in a case expression.

lib/puppet/pops/evaluator/compare_operator.rb
thallgren: Almost agree with @hlindberg , but it should say `equals(symbol, left)`, i.e. no `scope` parameter.

Iristyle: Ah, thanks for the info - I can update / correct.

hlindberg: Thanks thomas - yeah, no scope - my mistake.

Iristyle:
@hlindberg / @thallgren can you give me another hint about how to get `:undef` into a spec as you've mentioned?

I tried making some additions around https://github.com/puppetlabs/puppet/blob/master/spec/unit/pops/evaluator/evaluating_parser_spec.rb#L532-L545 but it seems that only the keyword `default` gets special treatment as a symbol.

hlindberg:
@Iristyle The only way to get an `:undef` symbol into data that the evaluator is processing is if a function returns it (directly or wrapped in array/hash, etc). There is no way to create an `:undef` value in the puppet language.

The rationale for still supporting checks against `:undef` is that there is 3.x logic out in the wild that can return it directly or as a value inside an array or hash. For performance reasons we do not have translation of 3.x ruby conventions to 4.x conventions.

A test would need to define a function returning :undef, or be a unit test of the lower level API of CompareOperator.

joshcooper:
Needs rebase

lib/puppet/bindings.rb
hlindberg: dead as a doornail - file should be removed

hlindberg: See https://github.com/puppetlabs/puppet/pull/6405

lib/puppet/util/instance_loader.rb
```diff
@@ -41,7 +41,7 @@ def instance_docs(type)
 
     # Use this method so they all get loaded
     loaded_instances(type).sort { |a,b| a.to_s <=> b.to_s }.each do |name|
-      mod = self.loaded_instance(name)
+      mod = self.loaded_instance(type, name)
```
joshcooper: this is terrifying

joshcooper:
@Iristyle there looks to be some good changes in here. It looks like the things @hlindberg and @thallgren mentioned have already been deleted (in puppet 5), so if you could rebase/fixup this PR, we can merge for 6

lib/puppet/util/instance_loader.rb
Iristyle: Fortunately only used for documentation generation.

PR #5718, opened 2017-03-14
(PUP-7340) Fix context_spec failure on OSX

thallgren:
A test that wrote new content with the exact same size to a file and
then checked if the file had changed by comparing inode, mtime, and
size, failed on OSX since the time resolution was too low for the rapid
create, read, change, check cycle and the inode stayed the same. This
commit ensures that the size of the file changes.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -161,6 +161,8 @@ RSpec.configure do |config|
   config.after :each do
     Puppet::Test::TestHelper.after_each_test()
 
+    Puppet.lookup(:environments).clear_all
+
     # TODO: would like to move this into puppetlabs_spec_helper, but there are namespace issues at the moment.
     PuppetSpec::Files.cleanup
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -161,8 +161,6 @@ RSpec.configure do |config|
   config.after :each do
     Puppet::Test::TestHelper.after_each_test()
 
-    Puppet.lookup(:environments).clear_all
-
     # TODO: would like to move this into puppetlabs_spec_helper, but there are namespace issues at the moment.
     PuppetSpec::Files.cleanup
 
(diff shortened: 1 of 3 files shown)


joshcooper:
If we change the test to always change the file size, how do we know the code works correctly when the size doesn't change?

joshcooper:
We should move https://github.com/thallgren/puppet/blob/a163cdb2ed0938c7991bf8a276b5a26780d525b9/spec/unit/pops/lookup/context_spec.rb#L177 to the test_helper so that environments global state is always cleared after each test. And we can drop https://github.com/thallgren/puppet/blob/a163cdb2ed0938c7991bf8a276b5a26780d525b9/spec/unit/pops/lookup/context_spec.rb#L176 since settings are cleared after each test.

Iristyle:
I'm with @joshcooper on this one - if changing the file size is necessary to get the test to pass, that seems like a bug to me.

Iristyle:
To fix the other recommendations from @joshcooper you can cherry-pick over 40680d51188e84189958da3191633b9b81265cf6

thallgren:
Thanks @Iristyle, I tried that commit but it was conflicting (didn't remove my environment_timeout = 0 setting). I did that move as a separate commit though.

spec/spec_helper.rb
```diff
@@ -161,6 +161,8 @@ module PuppetSpec
   config.after :each do
     Puppet::Test::TestHelper.after_each_test()
 
+    Puppet.lookup(:environments).clear_all
+
```
joshcooper: nit, this should probably go in `Puppet::Test::TestHelper.after_each_test()` as it's something that should be called by puppetlabs_spec_helper when testing modules, and not specific to puppet's spec tests.

Iristyle:
@thallgren so it looks like the cache is keyed by a composite of `length`, `mtime`, and `inode` - is that correct?  And what you're pointing out is that it's very unlikely to update a file in the wild and not have at least one of these elements of the cache key change?

PR #5720, opened 2017-03-16
(PUP-7360) Add a minor_version method to make docs links easier

binford2k:
Sometimes error messages link to versioned docs pages. The docs pages
are only versioned by minor numbers, such as `4.9`. Ergo, it would be
nice to have a method to generate those easily. This adds that method.

This "reimplements" some version parsing because attempts to load the
SemanticPuppet library became too heavy and fragile.

### DIFF
--- a/lib/puppet/version.rb
+++ b/lib/puppet/version.rb
@@ -5,7 +5,6 @@
 # The version can be set programmatically because we want to allow the
 # Raketasks and such to set the version based on the output of `git describe`
 
-
 module Puppet
   PUPPETVERSION = '4.10.0'
 
@@ -66,6 +65,11 @@ module Puppet
       @puppet_version = version
     end
     @puppet_version ||= PUPPETVERSION
+  end
+
+  # @return [String] containing the puppet version to minor specificity, e.g. "3.0"
+  def self.minor_version
+    self.version.split('.')[0..1].join('.')
   end
 
   def self.version=(version)
--- a/lib/hiera/puppet_function.rb
+++ b/lib/hiera/puppet_function.rb
@@ -61,7 +61,7 @@ class Hiera::PuppetFunction < Puppet::Functions::InternalFunction
   def lookup(scope, key, default, has_default, override, &default_block)
     unless Puppet[:strict] == :off
       Puppet.warn_once(:deprecation, self.class.name,
-        "The function '#{self.class.name}' is deprecated in favor of using 'lookup'. See https://docs.puppet.com/puppet/#{Puppet.version}/reference/deprecated_language.html")
+        "The function '#{self.class.name}' is deprecated in favor of using 'lookup'. See https://docs.puppet.com/puppet/#{Puppet.minor_version}/reference/deprecated_language.html")
     end
     lookup_invocation = Puppet::Pops::Lookup::Invocation.new(scope, {}, {})
     adapter = lookup_invocation.lookup_adapter
--- a/lib/puppet/parser/functions/hiera.rb
+++ b/lib/puppet/parser/functions/hiera.rb
@@ -78,7 +78,7 @@ The returned value's data type depends on the types of the results. In the examp
 above, Hiera matches the 'users' key and returns it as a hash.
 
 The `hiera` function is deprecated in favor of using `lookup` and will be removed in 6.0.0.
-See  https://docs.puppet.com/puppet/#{Puppet.version}/reference/deprecated_language.html.
+See  https://docs.puppet.com/puppet/#{Puppet.minor_version}/reference/deprecated_language.html.
 Replace the calls as follows:
 
 | from  | to |
--- a/lib/puppet/parser/functions/hiera_array.rb
+++ b/lib/puppet/parser/functions/hiera_array.rb
@@ -66,7 +66,7 @@ $allusers = hiera_array('users') | $key | { "Key \'${key}\' not found" }
 value is a hash, Puppet raises a type mismatch error.
 
 `hiera_array` is deprecated in favor of using `lookup` and will be removed in 6.0.0.
-See  https://docs.puppet.com/puppet/#{Puppet.version}/reference/deprecated_language.html.
+See  https://docs.puppet.com/puppet/#{Puppet.minor_version}/reference/deprecated_language.html.
 Replace the calls as follows:
 
 | from  | to |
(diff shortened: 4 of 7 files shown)


MosesMendoza:
Hey @binford2k, thanks for filing. I'm a little concerned that this would bind us to certain language/definitions in a public api but separate from our SemVer implementation (see 'lib/puppet/vendor/semantic_puppet'). For example, in Semantic, the terms `major`, `minor`, `patch`, etc have specific definitions we should adhere to - puppet's minor version would be a single integer, not the combination of the first and second integers separated by a period.

I wonder if for the purposes of this we could still leverage that library without loading all of puppet (which seems to be one of the goals for `puppet/version`). I.e., `require 'puppet/vendor/semantic_puppet/lib/semantic_puppet`.. though we were recently bitten by an external project reaching into puppet's vendored libs and expecting them not to change. So I'm not sure it's a good idea. BUT in that case we could do something like
```ruby
def self.parse_version
  @parsed ||= SemanticPuppet::Version.parse(@puppet_version)
end

def self.major
  @parsed.major
end
...
```

binford2k:
I forgot about that library. I didn't feel good about reinventing here. We do need to have the `x.y` version easily accessible though, even if it uses another name. Alternatively, we can add redirects on the docs page so the `x.y.z` form redirects appropriately to the correct section. (already proposed and linked to the ticket)

Thoughts about that?

thallgren:
Also, this PR needs to target the 4.10.x branch.

thallgren:
I like @MosesMendoza suggestion but would like to change it so that the `major` method doesn't assume that parsing has been done already, i.e.:
```
# @return [SemanticPuppet::Version] the puppet version
def self.parsed_version
  @parsed_version ||= SemanticPuppet::Version.parse(@puppet_version)
end

# @return [String] the minor version as a  "major.minor" string
def self.minor_version
  v = parsed_version
  "#{v.major}.#{v.minor}"
end
```
With this approach, there's a need to return the major version specifically since all individual version numbers can be reached using `Puppet.parsed_version.major`, `Puppet.parsed_version.minor`, etc.

binford2k:
@thallgren this should target `stable`? Can the docs at https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md#making-changes be updated then?

lib/puppet/version.rb
```diff
@@ -5,6 +5,11 @@
 # The version can be set programmatically because we want to allow the
 # Raketasks and such to set the version based on the output of `git describe`
 
+# Seems to be the minimum to get a vendored library without adding extra dependencies.
+vendor_dir = File.join([File.dirname(File.expand_path(__FILE__)), "vendor"])
+load "#{vendor_dir}/load_semantic_puppet.rb"
```
thallgren: Ouch. Didn't anticipate this and I don't like it. The proper way to load the vendored stuff is to use `Puppet::Vendor.load_vendored` so this seems a bit patchy. Using the proper way will however not work since it requires `Puppet.debug` which is instantiated much later. Based on that, I'd rather go back to the previous solution with split/join. @MosesMendoza , do you agree with this?

thallgren:
@binford2k, yes, this PR must target `stable` in order for it to be included in the 4.10.x release. The documentation says "usually the master branch" and I agree, that wording could perhaps be improved, given that targeting `stable` isn't really that unusual.

lib/puppet/version.rb
binford2k: I should have phrased my commit message better. That error log was requiring `puppet/vendor` and then calling `.load_vendored`. I don't like it either.

binford2k:
wow, that was a mess. I'll try to clean it up.

lib/puppet/version.rb
MosesMendoza: Yeah that was my concern... how bad would it feel to reach into vendor directly. Seems iffy at best, risky at worst. I think +1 to just split/join, but can you make sure the variable names match up with SemanticPuppet?

lib/puppet/version.rb
```diff
@@ -73,19 +67,19 @@ def self.version
...
-    "#{v.major}.#{v.minor}"
+  # @return [String] containing the puppet minor version, e.g. "0"
+  def self.version_minor
+    self.version.split('.')[1]
   end
 
-  def self.major_version
-    parsed_version.major.to_s
+  # @return [String] containing the puppet version to minor specificity, e.g. "3.0"
+  def self.version_to_minor
```
thallgren: I liked the previous naming better. This name `version_to_minor` suggest that some kind of transformation takes place, from version to minor.

lib/puppet/version.rb
```diff
@@ -73,19 +67,19 @@ def self.version
...
+  def self.version_major
+    self.version.split('.').first
   end
 
-  # @return [String] the minor version as a  "major.minor" string
-  def self.minor_version
-    v = parsed_version
-    "#{v.major}.#{v.minor}"
+  # @return [String] containing the puppet minor version, e.g. "0"
+  def self.version_minor
```
thallgren: Who needs this? I would assume that anyone that has the need to dissect a version would use the SemanticPuppet gem. I don't think this class should aspire to do what that gem does.

lib/puppet/version.rb
```diff
@@ -73,19 +67,19 @@ def self.version
     @puppet_version ||= PUPPETVERSION
   end
 
-  # @return [SemanticPuppet::Version] the puppet version
-  def self.parsed_version
-    @parsed_version ||= SemanticPuppet::Version.parse(version)
+  # @return [String] containing the puppet major version, e.g. "3"
+  def self.version_major
```
thallgren: I liked the old name `major_version` better. At the same time, I think this PR would be fine if this method was removed altogether. AFAIK, no one has requested a method that returns the major version number in string form.

lib/puppet/version.rb
binford2k: No, all I needed was the method that returns `MAJOR.MINOR`. @MosesMendoza suggested that it would be better if it were as close to SemVer nomenclature. I don't care either way.

lib/puppet/version.rb
binford2k: Do you have a suggestion that differentiates this from SemanticPuppet::Version.minor? (a single Integer digit)

thallgren: I think what you had before, `minor_version` works well. It's the same as version but restricted to minor granularity. Alternatively, perhaps `major_minor_version`.

lib/puppet/version.rb
thallgren: Ok, in that case, I'd suggest just removing it.

PR #5722, opened 2017-03-16
(PUP-7182) Improve error messages from Hiera 5

thallgren:
This commit contains an overhaul of the errors raised by the lookup
framework. Most errors are converted into proper issues and added to
lib/puppet/pops/issues.rb and, as such, allowed to be propagated all
the way to the log while retaining location information.

Many errors originating from the hiera.yaml will now also have line
information. This is accomplished by a combination of regexps and a
fairly naive parser that is aware of strings and comments.

The errors that were reported when a value was found that didn't conform
to the Puppet::LookupValue have been changed to explicitly state the
originating location and, when appropriate, the key of the culprit.

Unit tests were added to the unit/functions/lookup_spec.rb to cover all
the added issues.

### DIFF
--- a/lib/puppet/pops/lookup/hiera_config.rb
+++ b/lib/puppet/pops/lookup/hiera_config.rb
@@ -154,7 +154,9 @@ class HieraConfig
     when 3
       HieraConfigV3.new(config_root, config_path, loaded_config)
     else
-      raise Puppet::DataBinding::LookupError, "#{@config_path}: This runtime does not support #{CONFIG_FILE_NAME} version '#{version}'"
+      issue = Issues::HIERA_UNSUPPORTED_VERSION
+      raise Puppet::DataBinding::LookupError.new(
+        issue.format(:version => version),  config_path, nil, nil, nil, issue.issue_code)
     end
   end
 
@@ -173,6 +175,11 @@ class HieraConfig
     @data_providers = nil
   end
 
+  def fail(issue, args = EMPTY_HASH, line = nil)
+    raise Puppet::DataBinding::LookupError.new(
+      issue.format(args.merge(:label => self)),  @config_path, line, nil, nil, issue.issue_code)
+  end
+
   # Returns the data providers for this config
   #
   # @param lookup_invocation [Invocation] Invocation data containing scope, overrides, and defaults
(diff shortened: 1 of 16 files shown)


lib/puppet/pops/lookup/hiera_config.rb
```diff
@@ -173,6 +175,11 @@ def initialize(config_root, config_path, loaded_config)
     @data_providers = nil
   end
 
+  def fail(issue, args = EMPTY_HASH, line = nil)
```
MosesMendoza: Was it intentional to override/redefine ruby's `fail` here with our own (`fail` originaly being an alias to `raise`)? That seems like it could surprise someone.

hlindberg: We use this pattern elsewhere in the 4.x codebase (pops) - when that was added I was unaware that `fail` was an alias for `raise`. 
It is probably worth to modify that everywhere (new ticket) and use some other name for this - `raise_issue` or somesuch perhaps?

lib/puppet/pops/lookup/hiera_config.rb
```diff
@@ -323,7 +365,15 @@ def create_configured_data_providers(lookup_invocation, parent_data_provider)
     data_providers = {}
 
     [@config[KEY_BACKENDS]].flatten.each do |backend|
-      raise Puppet::DataBinding::LookupError, "#{@config_path}: Backend '#{backend}' defined more than once" if data_providers.include?(backend)
+      if data_providers.include?(backend)
+        first_line = find_line_matching(/[^\w]#{backend}(?:[^\w]|$)/)
```
MosesMendoza: Is it ok that these character classes inverse match only the ASCII character set for word ([a-zA-Z0-9_]) and not the unicode? I.e., `\w` instead of `[[:word:]]`

lib/puppet/pops/lookup/lookup_adapter.rb
```diff
@@ -85,6 +85,7 @@ def lookup_global(key, lookup_invocation, merge_strategy)
       end
     end
   rescue Puppet::DataBinding::LookupError => detail
+    raise detail unless detail.issue_code.nil?
```
MosesMendoza: I'm curious about this - is the test for `detail.issue_code.nil?` a way we're checking if this is one of our issues described above in `Issues` ?

thallgren: The reason for this test is to avoid that the exception is re-wrapped and thereby looses it's issue and location. Such re-wrapping will ultimately mean that a new location is assigned to the error. That new location is the file and line of the .pp file where the lookup (or APL) originated.
Ty my knowledge, all errors that have an issue_code originates {{Issues}}. If not, then something is wrong.

lib/puppet/pops/lookup/hiera_config.rb
thallgren: Yes, unless we think that backends will be named using such characters. Given that backends are on their way out, I don't think we need to worry about that.

lib/puppet/pops/lookup/hiera_config.rb
MosesMendoza: 👍

lib/puppet/pops/lookup/hiera_config.rb
MosesMendoza: ok

lib/puppet/pops/lookup/lookup_adapter.rb
MosesMendoza: got it, thanks

lib/puppet/pops/lookup/hiera_config.rb
hlindberg: Ping @thallgren - do you agree? Change `fail` here to something else, and then make the corresponding change elsewhere at some other point?

thallgren: I think it would be better to change all `fail` as a separate ticket. Otherwise, we have a situation in between where two conventions are used.

hlindberg: ok, fair - logged PUP-7314 for that.

PR #5723, opened 2017-03-16
(PUP-2469) Puppet emits error if numeric file mode attribute specified

kurtwall:
This test verifies that puppet emits an error if a file resource is declared using a numeric mode attribute.

### DIFF
--- a/acceptance/tests/resource/file/numeric_mode_error.rb
+++ b/acceptance/tests/resource/file/numeric_mode_error.rb
@@ -1,0 +1,34 @@
+test_name 'puppet errors if file resource specified with numeric mode'
+# TestRail test case C14937
+
+agents.each do |a|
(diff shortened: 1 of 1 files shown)


joshcooper:
So we have a unit test for this in https://github.com/puppetlabs/puppet/blob/master/spec/unit/type/file/mode_spec.rb#L13-L17. I'd prefer to not add beaker tests unless it's really necessary. What about adding an integration test to https://github.com/puppetlabs/puppet/blob/master/spec/integration/type/file_spec.rb which exercising applying file resources?

kurtwall:
@joshcooper Evidently I'm blind. I missed the unit test when I went scanned these files for a test that did this. There's no need to add another test that tests the same thing. I'll close this PR.

PR #5725, opened 2017-03-16
(PUP-7283) Disable CRL checking when fetching the CRL

adrienthebo:
When we're fetching the CRL for the first time and a CRL hasn't been provided out of band, we need to disable CRL checking until that request has completed.

I've also included a graphical representation of the indirector for anyone interested.

Ouroboros


lib/puppet/context.rb
```diff
@@ -40,6 +40,7 @@ def pop
     if @stack[-1][0] == 0
       raise(StackUnderflow, "Attempted to pop, but already at root of the context stack.")
     else
+      @id -= 1
```
MosesMendoza: Was this just a bug that already existed?

lib/puppet/indirector/certificate_revocation_list/rest.rb
```diff
@@ -7,4 +7,16 @@ class Puppet::SSL::CertificateRevocationList::Rest < Puppet::Indirector::REST
   use_server_setting(:ca_server)
   use_port_setting(:ca_port)
   use_srv_service(:ca)
+
+  def find(request)
```
MosesMendoza: Would we ever execute a find via the `file` terminus and run into the same problem?

lib/puppet/context.rb
adrienthebo: The `@id` field isn't directly exposed and you can only really observe it by stepping into the class via the debugger. I assumed the id was meant to indicate the context depth but it could be a unique identifier. Is that more helpful, identifying the context or stack depth?

But since this isn't exposed anywhere I can drop this commit and we can pretend this question never came up.

lib/puppet/indirector/certificate_revocation_list/rest.rb
adrienthebo: We have to have this guard clause because we might need to make network operations in order to configure SSL, which causes infinite recursion. With the file terminus the file is there or it's not, and we won't recurse.

lib/puppet/context.rb
MosesMendoza: ¯\_(ツ)_/¯. I looked further at it and agree its innocuous.

lib/puppet/indirector/certificate_revocation_list/rest.rb
MosesMendoza: 👍  that makes sense

MosesMendoza:
@adrienthebo are you able to do a local acceptance run against this?

adrienthebo:
@MosesMendoza I've run the tests locally and I don't get the infinite recursion that we did last time. However I was getting a failure on the autosign command tests but that was also failing against the 4.9.4 tag, so I don't know if it's related. I'm going to investigate more and see if the failure is related or due to some precondition that another test was satisfying.

adrienthebo:
@branan @MosesMendoza I've been getting a lot of failures when running beaker locally, and they generally seem unrelated to the changes I've made. The autosigning failure that I've seen is perplexing but it happens on master and my branch, so I'm baffled. May we try merging this and see if it works? If nothing else the infinite recursion looks fixed.

MosesMendoza:
@adrienthebo sounds good. puppet/master tests are currently red but good to merge when green.

adrienthebo:
Thank you!

lib/puppet/context.rb
joshcooper: I'm pretty sure `@id` is meant to be a unique id for things in the stack, not the stack depth. Although it isn't exposed outside of this class, I'm thinking we shouldn't modify it's behavior, especially since it's not meaningful to this PR.

lib/puppet/indirector/certificate_revocation_list/rest.rb
```diff
@@ -7,4 +7,16 @@ class Puppet::SSL::CertificateRevocationList::Rest < Puppet::Indirector::REST
...
+    if request.method == :find && !Puppet::FileSystem.exist?(Puppet[:hostcrl])
```
joshcooper: Is the `request.method == :find` check redundant?

lib/puppet/context.rb
adrienthebo: Removed.

lib/puppet/indirector/certificate_revocation_list/rest.rb
adrienthebo: Thanks for catching that, I originally implemented this in `#network` where this check was needed. When I shifted the implementation I didn't drop this check. It's gone now.

adrienthebo:
@MosesMendoza it looks like the puppet master component tests are passing, can this get merged today?

branan:
Since this has a thumb from @MosesMendoza pending a green master, master is green, and I am also :+1: : I'm hitting the big green button. Nobody can stop me. NOBODY! MUAHAHAHAHAHAHAHA!

PR #5726, opened 2017-03-16
(PUP-7382)(SPIKE) Customize YAML serialization

Iristyle:
- In Ruby 2.3, a change was introduced to how Hash is seralized as
   YAML. If Puppet resources are derived from Hash instead of the
   usual Puppet::Type, they will now try to serialize data that was
   previously not serialized (including Provider instances).

   In particular, use of the ACL module fails on Ruby 2.3 to properly
   serialize corrective changes YAML because Puppet::Type::Acl::Ace
   derives from Hash.

   While changes to the ACL module can be made to either implement
   `encode_with` directly OR to instead derive from something other
   than Hash, this may not handle any other such instances within the
   Puppet ecosystem.

   NOTE: other options are currently being explored to address this

### DIFF
--- a/lib/puppet/transaction/resource_harness.rb
+++ b/lib/puppet/transaction/resource_harness.rb
@@ -283,7 +283,24 @@ class Puppet::Transaction::ResourceHarness
       # We use instance_variable_get here because we want this process to bypass any
       # munging/unmunging or validation that the property might try to do, since those
       # operations may not be correctly implemented for custom types.
-      property.instance_variable_get(:@should)
+      # require 'pry'; binding.pry if property.name == :permissions
+      should_value = property.instance_variable_get(:@should)
+      # wrapping trick to ensure Enumerable
+      [should_value].flatten.each { |i| customize_yaml_generation(i) }
+      should_value
+    end
+  end
+
+  def customize_yaml_generation(instance, to_remove = ['provider'])
+    # require 'pry'; binding.pry if instance.class.name == 'Puppet::Type::Acl::Ace'
+    if !instance.respond_to?(:encode_with) && to_remove.any? { |n| instance.instance_variable_defined?("@#{n}") }
+      singleton_class = class << instance; self; end
+      singleton_class.send(:define_method, :encode_with) do |coder|
+        names = instance_variables.map { |i| i.to_s[1..-1] } - to_remove
+        names.each do |var|
+          coder[var] = instance.instance_variable_get("@#{var}")
+        end
+      end
     end
   end
 


lib/puppet/transaction/resource_harness.rb
```diff
@@ -283,7 +283,24 @@ def new_system_value(property, event, old_system_value)
...
+      # wrapping trick to ensure Enumerable
+      [should_value].flatten.each { |i| customize_yaml_generation(i) }
+      should_value
+    end
+  end
+
+  def customize_yaml_generation(instance, to_remove = ['provider'])
+    # require 'pry'; binding.pry if instance.class.name == 'Puppet::Type::Acl::Ace'
+    if !instance.respond_to?(:encode_with) && to_remove.any? { |n| instance.instance_variable_defined?("@#{n}") }
+      singleton_class = class << instance; self; end
```
Iristyle: Monkey-patching instances on the fly ... not great.

adrienthebo: This is hilariously evil. I like it.

Iristyle:
Previous YAML for an ACE looked like this in Ruby 2.2 and lower:

```yaml
--- !ruby/hash:Puppet::Type::Acl::Ace {}
```

With this change YAML for an ACE looks like this on Ruby 2.3:

```yaml
---
- !ruby/object:Puppet::Type::Acl::Ace
  identity: Administrators
  hash: 
  id: S-1-5-32-544

  mask: 
    rights:
    - :full
    perm_type: :allow
    child_types: :all
    affects: :all
    is_inherited: false
```

lib/puppet/transaction/resource_harness.rb
Iristyle: You laugh, but there is a precedent for this btw - https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/file.rb#L302-L317

lib/puppet/transaction/resource_harness.rb
```diff
@@ -283,7 +283,24 @@ def new_system_value(property, event, old_system_value)
...
+      should_value = property.instance_variable_get(:@should)
+      # wrapping trick to ensure Enumerable
+      [should_value].flatten.each { |i| customize_yaml_generation(i) }
+      should_value
+    end
+  end
+
+  def customize_yaml_generation(instance, to_remove = ['provider'])
+    # require 'pry'; binding.pry if instance.class.name == 'Puppet::Type::Acl::Ace'
+    if !instance.respond_to?(:encode_with) && to_remove.any? { |n| instance.instance_variable_defined?("@#{n}") }
```
branan: I wonder if instead of looking for property names, we could look for certain known-bad types? Make sure that property `should` values don't contain providers, catalogs, etc.

Iristyle: Yeah, I thought about type vs. name as well - not sure which is better, but if you had a property named `provider` that wasn't actually a `provider`, seems like it would be bad to exclude it.

Iristyle:
This is not really PUP-7381 but rather closely related.

joshcooper:
I think we can close this.

PR #5727, opened 2017-03-17
(PUP-7368) Add ability to declare function argument mismatch handlers

thallgren:
This commit adds an alternative `dispatch` to a function. It is named
`argument_mismatch`. The function that it calls must produce a String
that represents a message describing that the matched arguments are in
fact in error. An `ArgumentError` is then raised when the function returns
that string.

An `argument_mismatch` dispatcher will be excluded when the default
argument mismatch handler describes viable options. It is also excluded
from the list of dispatchers returned from the `Dispatcher`.

A function may have zero to many `argument_mismatch` declarations and
when none of them matches, the default mismatch describer still kicks
in.

### DIFF
--- a/lib/puppet/functions.rb
+++ b/lib/puppet/functions.rb
@@ -48,11 +48,11 @@
 #
 # Special dispatches designed to create error messages for an argument mismatch
 # can be added using the keyword `argument_mismatch` instead of `dispatch`. The
-# function appointed by an `argument_mismatch` will be called with arguments
-# just like a normal `dispatch` function would, but the function must produce a
-# string. The string is then used as the message in the `ArgumentError` that is
-# raised when the function returns. A block parameter can be given, but it is not
-# propagated in the function call.
+# method appointed by an `argument_mismatch` will be called with arguments
+# just like a normal `dispatch` would, but the method must produce a string.
+# The string is then used as the message in the `ArgumentError` that is raised
+# when the method returns. A block parameter can be given, but it is not
+# propagated in the method call.
 #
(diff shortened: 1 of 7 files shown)


lib/puppet/functions.rb
```diff
@@ -46,6 +46,14 @@
 # specifies the method that should be called for that signature. When a
 # matching signature is found, the corresponding method is called.
 #
+# Special dispatches designed to create error messages for an argument mismatch
+# can be added using the keyword `argument_mismatch` instead of `dispatch`. The
+# function appointed by an `argument_mismatch` will be called with arguments
```
hlindberg: "method appointed" instead of "function appointed"

lib/puppet/functions.rb
```diff
@@ -46,6 +46,14 @@
...
+# just like a normal `dispatch` function would, but the function must produce a
```
hlindberg: strike "function" in "function would", and change "but that function must" to "but the method must..."

lib/puppet/functions.rb
```diff
@@ -46,6 +46,14 @@
...
+# string. The string is then used as the message in the `ArgumentError` that is
+# raised when the function returns. A block parameter can be given, but it is not
```
hlindberg: change "function" to "method"

lib/puppet/functions.rb
```diff
@@ -46,6 +46,14 @@
...
+# propagated in the function call.
```
hlindberg: change "function call" to "method call"

PR #5729, opened 2017-03-17
(PUP-7273) Improve hiera function error message when path is missing

thallgren:
Before this commit, a missing path declaration in hiera.yaml resulted in
an error explaing that the options hash sent to a data_hash function that
required a path had an incorrect size (0 instead of 1). While correct,
that message was not very informative. This commit leverages the new
`argument_mismatch` dispatcher added by PUP-7368 to generate more useful
errors when the path is missing.

### DIFF
--- a/lib/puppet/functions/eyaml_lookup_key.rb
+++ b/lib/puppet/functions/eyaml_lookup_key.rb
@@ -19,6 +19,13 @@ Puppet::Functions.create_function(:eyaml_lookup_key) do
   def eyaml_lookup_key(key, options, context)
     return context.cached_value(key) if context.cache_has_key(key)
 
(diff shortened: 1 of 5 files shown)


lib/puppet/functions/eyaml_lookup_key.rb
```diff
@@ -19,6 +19,13 @@
   def eyaml_lookup_key(key, options, context)
     return context.cached_value(key) if context.cache_has_key(key)
 
+    # Can't do this with an argument_mismatch dispatcher since there is no way to declare a struct that at least
+    # contains some keys but may contain other arbitrary keys.
```
hlindberg: I think we need such a type for precisely this reason. (separate issue).

PR #5730, opened 2017-03-17
(PUP-6854) Add paths to skip for launchd

keeleysam:
On macOS 10.12 and higher, the file `/System/Library/LaunchDaemons/com.apple.jetsamproperties.Mac.plist` exists which is not actually a service.  This would generate the following warning on every single Puppet run:

```
Warning: The /System/Library/LaunchDaemons/com.apple.jetsamproperties.Mac.plist plist does not contain a 'label' key; Puppet is skipping it
```

As this file (and perhaps others in the future) will never be a service, this adds a list of files which we will now skip from reading at all.

### DIFF
--- a/lib/puppet/provider/service/launchd.rb
+++ b/lib/puppet/provider/service/launchd.rb
@@ -65,6 +65,17 @@ Puppet::Type.type(:service).provide :launchd, :parent => :base do
     ]
   end
 
+  # These are known bad files in the normal paths, skip them completely.
+  # There's only one right now, it's a file of memory limits and other
+  # data, it's not actually a service and would generate a warning on every run.
+  #
+  # @api private
+  def self.launchd_skip_paths
+    [
+      '/System/Library/LaunchDaemons/com.apple.jetsamproperties.Mac.plist'
+    ]
+  end
+
   # Gets the current Darwin version, example 10.6 returns 9 and 10.10 returns 14
   # See https://en.wikipedia.org/wiki/Darwin_(operating_system)#Release_history
   # for more information.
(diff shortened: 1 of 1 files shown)


daenney:
Could we get this merged please? We've got a fleet of Macs that are reporting that warning everywhere.

MosesMendoza:
Thanks @keeleysam for your contribution! We try to have test coverage for changes we add to puppet. Can you add a test for this change to https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/service/launchd_spec.rb? If you're not familiar with rspec we can help you out.

jgreen210:
Somehow asserting that Puppet#warning or above is not called by any test case (except in the one test case where warning is called) would give sufficient coverage I think, in combination with testing on enough mac OS platforms.

Alternatively, living without a small amount of coverage seems better to me (as a puppet user) than having unreadable logs if manage services on a recent mac OS.

MosesMendoza:
It seems like this might be effectively duplicating the fall-back case where a plist is malformed, so we issue a warning about it. It sounds like we now know there are always going to 1 or more of these plists, and we don't want the warning. Rather than adding a new hardcoded list of plists, could we just switch the fallback to logging at the debug level rather than the warning level? I think that would do the same thing with less code? Then for test it's just a matter of switching warning to debug here: https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/service/launchd_spec.rb#L295

octomike:
Is it really just those two lines? The message is sooo annoying.

MosesMendoza:
superseded by https://github.com/puppetlabs/puppet/pull/6273

PR #5731, opened 2017-03-17
(PUP-893) Deprecate `inspect` command and `audit` metaparameter

Magisus:
This PR:
1) Adds a utility for deprecating applications based on Moses's work in https://github.com/puppetlabs/puppet/pull/5449/commits/7fefceae40d5ace8df2bd85aefa8bdee3246d9e1
2) Deprecates the `inspect` command
3) Deprecates the `audit` metaparameter

### DIFF
--- a/lib/puppet/application/inspect.rb
+++ b/lib/puppet/application/inspect.rb
@@ -19,6 +19,7 @@ puppet-inspect(8) -- Send an inspection report
 
 SYNOPSIS
 --------
+Note: this command is deprecated
 
 Prepares and submits an inspection report to the puppet master.
 
@@ -74,6 +75,8 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
   end
 
   def setup
+    deprecate
+
     exit(Puppet.settings.print_configs ? 0 : 1) if Puppet.settings.print_configs?
 
     raise "Inspect requires reporting to be enabled. Set report=true in puppet.conf to enable reporting." unless Puppet[:report]
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -1692,11 +1692,21 @@ EOT
     :archive_files => {
         :type     => :boolean,
         :default  => false,
-        :desc     => "During an inspect run, whether to archive files whose contents are audited to a file bucket.",
+        :desc     => "During an inspect run, whether to archive files whose contents are audited to a file bucket. Note that the `inspect` command is deprecated.",
+        :hook => proc { |value|
+          if Puppet[:strict] != :off
+            Puppet.deprecation_warning(_("Setting 'archive_files' is deprecated. It will be removed in a future release along with the `inspect` command."))
+          end
+        }
     },
     :archive_file_server => {
         :default  => "$server",
-        :desc     => "During an inspect run, the file bucket server to archive files to if archive_files is set.",
+        :desc     => "During an inspect run, the file bucket server to archive files to if archive_files is set. Note that the `inspect` command is deprecated.",
+        :hook => proc { |value|
+          if Puppet[:strict] != :off
+            Puppet.deprecation_warning(_("Setting 'archive_file_server' is deprecated. It will be removed in a future release along with the `inspect` command."))
+          end
+        }
     }
   )
 
--- a/lib/puppet/type.rb
+++ b/lib/puppet/type.rb
@@ -1291,7 +1291,9 @@ class Type
   end
 
   newmetaparam(:audit) do
-    desc "Marks a subset of this resource's unmanaged attributes for auditing. Accepts an
+    desc "(This metaparameter is deprecated and will be ignored in a future release.)
+
+      Marks a subset of this resource's unmanaged attributes for auditing. Accepts an
       attribute name, an array of attribute names, or `all`.
 
       Auditing a resource attribute has two effects: First, whenever a catalog
@@ -1312,6 +1314,14 @@ class Type
       and the second run will log the edit made by Puppet.)"
 
     validate do |list|
+      if Puppet.settings[:strict] != :off
+        dep_warning = _("The `audit` metaparameter is deprecated and will be ignored in a future release.")
+        if file && line
+          puppet_deprecation_warning(dep_warning, { :line => line, :file => file })
+        else
+          Puppet.deprecation_warning(dep_warning)
+        end
+      end
       list = Array(list).collect {|p| p.to_sym}
       unless list == [:all]
         list.each do |param|
--- a/spec/unit/type/file_spec.rb
+++ b/spec/unit/type/file_spec.rb
@@ -1598,7 +1598,38 @@ describe Puppet::Type.type(:file) do
       catalog.apply
 
       expect(Puppet::FileSystem.exist?(path)).to be_truthy
-      expect(@logs).not_to be_any {|l| l.level != :notice }
+      expect(@logs).not_to be_any { |l|
+        # the audit metaparameter is deprecated and logs a warning
+        l.level != :notice && !(/deprecated/.match(l.message))
+      }
+    end
+
+    it "should not log a deprecation warning when `strict` logging is turned off" do
+      # with `strict` turned off
+      Puppet.settings[:strict] = :off
+
+      file[:audit]  = 'content'
+      file[:ensure] = 'present'
+      catalog.add_resource(file)
+
+      catalog.apply
+      expect(@logs).not_to be_any { |l|
+        /deprecated/.match(l.message)
+      }
+    end
+
+    it "should log a deprecation warning when `strict` is turned on" do
+      # with strict turned on
+      Puppet.settings[:strict] = :warning
+
+      file[:audit]  = 'content'
+      file[:ensure] = 'present'
+      catalog.add_resource(file)
+
+      catalog.apply
+      expect(@logs).to be_any { |l|
+        l.level == :warning && /deprecated/.match(l.message)
+      }
     end
   end
 
--- a/lib/puppet/application.rb
+++ b/lib/puppet/application.rb
@@ -322,6 +322,15 @@ class Application
     @options = {}
   end
 
+  # call in setup of subclass to deprecate an application
+  def deprecate
+    @deprecated = true
+  end
+
+  def deprecated?
+    @deprecated
+  end
+
   # Execute the application.
   # @api public
   # @return [void]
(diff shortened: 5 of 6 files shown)


MosesMendoza:
@Magisus looks like related specs failed. i'm not sure if it applies here, but when we started doing the face deprecations, I recall spec testing got a little squirrelly because of the lifecycle implications. I added https://github.com/puppetlabs/puppet/pull/5449/files which helped a bit. Different class (Interface vs Application) but if it's of any use.

Magisus:
Updated, specs now pass. Added a utility for deprecating applications, would like feedback on whether the approach seems reasonable, based it off Moses's work for faces linked above, but it was a little weirder. Also, does it make sense to add deprecation warnings to the settings? We do that in some places, but I know there was also a complaint of too much noise (e.g. around the server_list thing).

lib/puppet/type.rb
```diff
@@ -1316,6 +1318,7 @@ def pathbuilder
       and the second run will log the edit made by Puppet.)"
 
     validate do |list|
+      Puppet.deprecation_warning(_("The `audit` metaparameter is deprecated and will be ignored in a future release."))
```
joshcooper: I think this will cause one deprecation warning to be issue for each instance of `audit` in a resource. I'm wondering if this should be a docs only deprecation or a "warn once" type of deprecation. /cc @nfagerlund @ahpook

Magisus: I agree that this should definitely be at least `warn_once` if it's otherwise per instance in a resource; that sounds insane. Is docs-only enough to get people to remove it?

joshcooper: After talking with @ahpook the desired behavior is to:

1. Document that `audit` metaparam is deprecated (done)
2. By default do not emit deprecation warnings if they appear in a manifest
3. When running with `strict` mode enabled, emit deprecation warnings for each resource using `audit` including file and line numbers, so users know where to look to remove them.

/cc @nfagerlund @jtappa

Magisus: Sounds good, I'll update this.

ahpook: Exactly right @joshcooper 💯

ahpook: Thanks @Magisus !! 🎉

lib/puppet/defaults.rb
```diff
@@ -1692,11 +1692,17 @@ def self.default_diffargs
     :archive_files => {
         :type     => :boolean,
         :default  => false,
-        :desc     => "During an inspect run, whether to archive files whose contents are audited to a file bucket.",
+        :desc     => "During an inspect run, whether to archive files whose contents are audited to a file bucket. Note that the `inspect` command is deprecated.",
+        :hook => proc { |value|
+          Puppet.deprecation_warning(_("Setting 'archive_files' is deprecated. It will be removed in a future release along with the `inspect` command."))
```
Magisus: @joshcooper how should that decision around the audit parameter affect what's being done for these settings?

joshcooper: Good question @Magisus. We're not consistent in how we issue deprecation warnings for settings. Sometimes it's whenever the value is set, e.g. `trusted_server_facts`, or other times only in strict mode, e.g. `data_binding_terminus`.

Moving forward I think we adopt the latter approach. Don't spam users, but give them an opportunity to run with strict.

In this particular case, these settings are in the `:inspect` section, but nothing in puppet will `Puppet.settings.use(:inspect)` them. So they never end up in a settings catalog, like can happen with file/dir settings (which can cause deprecation warnings each time puppet runs).

AFAIK these settings only exist so that the `inspect` application can have default values, but override them on the command line. If you have these settings in `puppet.conf` and you run puppet (apply, agent, etc), does it issue a deprecation? If so, then I would make it only do that in `strict` mode.

Magisus: If the setting appears in the config file, a warning is issued. I will make it strict only.

spec/unit/type/file_spec.rb
```diff
@@ -1598,7 +1598,10 @@
       catalog.apply
 
       expect(Puppet::FileSystem.exist?(path)).to be_truthy
-      expect(@logs).not_to be_any {|l| l.level != :notice }
+      expect(@logs).not_to be_any { |l|
+        # the audit metaparameter is deprecated and logs a warning
+        l.level != :notice && !(/deprecated/.match(l.message))
+      }
```
MosesMendoza: What do you think about adding a spec that verifies a deprecation warning is logged? 

Per prior discussion, we'll probably want an assertion that a deprecation warning is logged in strict mode, and an assertion that none is logged in "non-strict" mode.

Magisus: Sounds good

lib/puppet/type.rb
```diff
@@ -1316,6 +1318,9 @@ def pathbuilder
       and the second run will log the edit made by Puppet.)"
 
     validate do |list|
+      if Puppet.settings[:strict] != :off
+        Puppet.deprecation_warning(_("You are attempting to use the `audit` metaparameter in %{resource} in %{manifest}. This parameter is deprecated and will be ignored in a future release.") % { resource: @resource.path, manifest: @resource.catalog.environment_instance.manifest })
```
Magisus: Is there a better way to say where this warning comes from? I was just using pry to poke around in the resource object, but there's a lot of stuff there...

lib/puppet/type.rb
```diff
@@ -1316,6 +1318,9 @@ def pathbuilder
       and the second run will log the edit made by Puppet.)"
 
     validate do |list|
+      if Puppet.settings[:strict] != :off
+        Puppet.deprecation_warning(_("You are attempting to use the `audit` metaparameter in %{resource}. This parameter is deprecated and will be ignored in a future release.") % { resource: @resource.path })
```
Magisus: Is there a way to say more specifically where this warning was triggered? Wasn't able to find what I was looking for poking around in the resource object with `pry`...

spec/unit/type/file_spec.rb
Magisus: Added two tests for this, @MosesMendoza

lib/puppet/type.rb
MosesMendoza: So there's actually a method in `Puppet::Util::Logging` called `puppet_deprecation_warning` ostensibly for this purpose (though I couldn't find it actually used anywhere): https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/logging.rb#L145-L153.

>   Logs a warning whose origin comes from Puppet source rather than somewhere internal within Puppet.  Otherwise the same as deprecation_warning()

This allows us to specify the origin file/line for the deprecation, rather than printing this file/line of lib/puppet/type.rb. Luckily, `Puppet::Type` includes `Puppet::Util::Logging` on line 82 so this method is available. Also, both `line` and `file` are available as methods on `self` here and I think they'll do what you want.

So this can be revised to something like:

```ruby
puppet_deprecation_warning("The audit metaparameter is deprecated and will be ignored in a future release", {:line => line, :file => file})
```
Worth a shot anyway?

MosesMendoza: It has the added nicety of prefixing the message with the resource ref:
```
☂ : be puppet apply ~/foo.pp
Notice: Compiled catalog for macbook-pro-2.local in environment production in 0.07 seconds
Warning: /File[/Users/moses/foo]/audit: The audit metaparameter is deprecated and will be ignored in a future release
   (at /Users/moses/foo.pp:1)
```

Magisus: Awesome, that looks perfect.

Magisus: Updated.

joshcooper:
The current behavior is slightly different in that `strict` defaults to `warning`. So if you have `archive_*` properties set, run `puppet inspect` or use the `audit` metaparam you will see the deprecation warnings, which can be turn off by specifying `strict = off` either on the command line or settings. Since puppet behaves this way for all deprecations, I think it's ok to follow that pattern.

lib/puppet/application/inspect.rb
```diff
@@ -74,6 +74,8 @@ def help
   end
 
   def setup
+    deprecate
```
MosesMendoza: @Magisus minor comment - can we also add `(Deprecated)` to the help output above?

I.e., so I get something like:
```
be puppet inspect --help

puppet-inspect(8) -- Send an inspection report (Deprecated)
...
```

Magisus: Updated.

MosesMendoza:
Confirmed deprecations are exhibited as expected:
Set `strict` off
```
☂ : be puppet config set strict off
```

Set `archive_files` and `archive_file_server`:
```
☂ : be puppet config set archive_files true
☂ : be puppet config set archive_file_server 'foo.bar'
```

Managed a resource with the `audit` parameter. Confirmed no deprecations exhibited:
```
☂ : cat ~/foo.pp
file { '/Users/moses/foo' :
  ensure => directory,
  audit => 'all',
}
☂ : be puppet apply ~/foo.pp
Notice: Compiled catalog for macbook-pro-2.local in environment production in 0.06 seconds
```

Confirmed the `inspect` application issues a warning:
```
☂ : be puppet inspect foo
Warning: `puppet inspect` is deprecated and will be removed in a future release.
   (at /Users/moses/foo/puppet/lib/puppet/application.rb:353:in `run')
```
As well as the help output:
```
puppet-inspect(8) -- Send an inspection report
========

SYNOPSIS
--------
Note: this command is deprecated
```

Turned strict on:
```
☂ : be puppet config set strict warning
```
Confirmed deprecations are exhibited as expected
```
☂ : be puppet apply ~/foo.pp
Warning: Setting 'archive_files' is deprecated. It will be removed in a future release along with the `inspect` command.
   (at /Users/moses/foo/puppet/lib/puppet/defaults.rb:1698:in `block in <module:Puppet>')
Warning: Setting 'archive_file_server' is deprecated. It will be removed in a future release along with the `inspect` command.
   (at /Users/moses/foo/puppet/lib/puppet/defaults.rb:1707:in `block in <module:Puppet>')
Notice: Compiled catalog for macbook-pro-2.local in environment production in 0.06 seconds
Warning: /File[/Users/moses/foo]/audit: The `audit` metaparameter is deprecated and will be ignored in a future release.
   (at /Users/moses/foo.pp:1)
```

MosesMendoza:
will merge when travis/appveyor pass

Magisus:
Closing and reopening to retrigger CI (it doesn't like it when you retarget branches).

PR #5733, opened 2017-03-20
(PUP-7370) Ensure value returned from provider is checked only once.

thallgren:
A data provider function that declares a return type, must use a type that is assignable to `Puppet::LookupValue` or, in case the function is a `data_hash`, to `Hash[Puppet::LookupKey,Puppet::LookupValue]`. This PR ensures that an error is raised if the return type is incorrect.

In this PR, it's also ensured that the lookup framework will rely on function to do return type validation when all of its dispatchers declares a correct return type.

### DIFF
--- a/spec/unit/functions/lookup_spec.rb
+++ b/spec/unit/functions/lookup_spec.rb
@@ -2183,55 +2183,57 @@ describe "The lookup function" do
       end
 
       context 'using a lookup_key that is a puppet function' do
+        let(:puppet_function) { <<-PUPPET.unindent }
+          function mod_a::pp_lookup_key(Puppet::LookupKey $key, Hash[String,String] $options, Puppet::LookupContext $context) >> Puppet::LookupValue {
+            case $key {
+              'mod_a::really_interpolated': { $context.interpolate("-- %{lookup('mod_a::a')} --") }
+              'mod_a::recursive': { lookup($key) }
+              default: {
+                if $context.cache_has_key(mod_a::a) {
+                  $context.explain || { 'reusing cache' }
+                } else {
+                  $context.explain || { 'initializing cache' }
+                  $context.cache_all({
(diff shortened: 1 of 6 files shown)


spec/unit/functions/lookup_spec.rb
```diff
@@ -2304,6 +2306,19 @@ def lookup(key, scope, order_override, resolution_type, context)
...
+        context 'with declared but incompatible return_type' do
+          let(:puppet_function) { <<-PUPPET.unindent }
+            function mod_a::pp_lookup_key(Puppet::LookupKey $key, Hash[String,String] $options, Puppet::LookupContext $context) >> Runtime['ruby','Symbol'] {
+              undef
+            }
+            PUPPET
+
+          it 'fails and reports error' do
+            expect{lookup('mod_a::a')}.to raise_error(
+              "Return type of 'lookup_key' function named 'mod_a::pp_lookup_key' is incorrect, expects a value of type Scalar, Sensitive, Type, Hash, or Array, got Runtime")
```
hlindberg: isn't Undef also allowed?

thallgren: Yes. And it's pretty weird that it doesn't show up in the message. Must be a bug in the mismatch describer. The type is defined as:
```
type Puppet::LookupValue =  Variant[
  Scalar,
  Undef,
  Sensitive,
  Type,
  Hash[Puppet::LookupKey, Puppet::LookupValue],
  Array[Puppet::LookupValue]]
```

hlindberg: want to fix in this PR, or separate?

thallgren: I added [PUP-7371](https://tickets.puppetlabs.com/browse/PUP-7371) as a blocker for this ticket.

hlindberg:
This is good to merge when tests go green.

PR #5735, opened 2017-03-20
(PUP-7373) Zypper is default package provider on SUSE

kurtwall:
PUP-2728 updated the default package provider to be `zypper` instead of `rug` on SUSE-family OSes. This PR, ticketed in PUP-7373, updates the package provider integration spec test to expect `zypper` as the default package provider on SUSE-family OSes 10.x and newer and `rug` on anything older.

### DIFF
--- a/spec/integration/type/package_spec.rb
+++ b/spec/integration/type/package_spec.rb
@@ -33,6 +33,12 @@ describe Puppet::Type.type(:package), "when choosing a default package provider"
       else
         :yum
       end
+    when 'Suse'
+      if Puppet::Util::Package.versioncmp(Facter.value(:operatingsystemmajrelease), '10') >= 0
+        :zypper
+      else
+        :rug
+      end
     when 'FreeBSD'
       :ports
     when 'OpenBSD'


kurtwall:
Passing spec run:
```
15:12 $ be rspec spec/integration/type/package_spec.rb --format d --color

Puppet::Type::Package when choosing a default package provider
  should have a default provider
  should choose the correct provider each platform

Puppet::Type::Package when packages with the same name are sourced
  with same title
    should give an error
  with different title
    should give an error
  from multiple providers
    when it should be present
      should do nothing if it is present
      should do nothing if it is latest
      should do nothing if it is 1.0
      should install if it is purged
      should install if it is absent

Puppet::Type::Package logging package state transitions
  installed package
    ensure => installed should log nothing
    ensure => absent should log removed
    ensure => purged should log purged
  absent package
    ensure => installed should log created
    ensure => absent should log nothing
    ensure => purged should log purged
  purged package
    ensure => installed should log created
    ensure => absent should log nothing
    ensure => purged should log nothing
```

ericldelaney:
looks ok based on my limited knowledge

Magisus:
Looks fine to me too.

joshcooper:
@kurtwall looks good, minor nit, can you change the commit summary to just `(PUP-7373) Zypper is default package provider on SUSE`? The PUP-2728 ticket was fixed awhile ago, so our ticket reconciliation scripts will get confused if that is in the commit summary.

kurtwall:
@joshcooper Thx. Done. I wasn't sure which one to reference.

kurtwall:
Merging based on approvals.

joshcooper:
Sorry I should have been more specific. We need the summary in the git commit to reference the ticket number. The summary in the PR is not as important. Does that make sense?

kurtwall:
@joshcooper Yeah, I realized that moments after I mashed the merge button. Sorry. :(

PR #5738, opened 2017-03-21
(PUP-1890) Change URI escaping to support UTF-8

Iristyle:
- URI is obsolete and has some undesirable behaviors:
    - URI.escape returns strings in US-ASCII instead of UTF-8
    - URI.parse returns strings in their source encoding instead of
      UTF-8
 - CGI behaves correctly with UTF-8, but returns more characters
   encoded, which might mess with URI.parse

### DIFF
--- a/acceptance/tests/reports/submission.rb
+++ b/acceptance/tests/reports/submission.rb
@@ -32,7 +32,7 @@ if master.is_pe?
       require "json"
 
       puppetdb_url = URI("http://localhost:8080/v3/reports")
-      puppetdb_url.query = URI.escape(%Q{query=["=","certname","#{agent}"]})
+      puppetdb_url.query = CGI.escape(%Q{query=["=","certname","#{agent}"]})
       result = Net::HTTP.get(puppetdb_url)
       json = JSON.load(result)
       puts json.first["receive-time"]
--- a/lib/puppet/configurer/fact_handler.rb
+++ b/lib/puppet/configurer/fact_handler.rb
@@ -32,6 +32,6 @@ module Puppet::Configurer::FactHandler
 
     text = facts.render(:pson)
 
-    {:facts_format => :pson, :facts => CGI.escape(text)}
+    {:facts_format => :pson, :facts => Puppet::Util.uri_query_encode(text)}
   end
 end
--- a/lib/puppet/file_serving/metadata.rb
+++ b/lib/puppet/file_serving/metadata.rb
@@ -31,7 +31,7 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
 
   def content_uri=(path)
     begin
-      uri = URI.parse(URI.escape(path))
+      uri = URI.parse(Puppet::Util.uri_encode(path))
     rescue URI::InvalidURIError => detail
       raise(ArgumentError, "Could not understand URI #{path}: #{detail}")
     end
--- a/lib/puppet/file_serving/metadata.rb
+++ b/lib/puppet/file_serving/metadata.rb
@@ -38,7 +38,7 @@ class Puppet::FileServing::Metadata < Puppet::FileServing::Base
     raise(ArgumentError, "Cannot use opaque URLs '#{path}'") unless uri.hierarchical?
     raise(ArgumentError, "Must use URLs of type puppet as content URI") if uri.scheme != "puppet"
 
-    @content_uri = path
+    @content_uri = path.encode(Encoding::UTF_8)
   end
 
   class MetaStat
--- a/lib/puppet/indirector/request.rb
+++ b/lib/puppet/indirector/request.rb
@@ -41,6 +41,8 @@ class Puppet::Indirector::Request
     end
   end
 
+  # @api private
+  # @deprecated used only internally to be removed in Puppet 5
   def escaped_key
     # Puppet::Network::HTTP::API::IndirectedRoutes is the only caller to escaped_key
     # and these fragments only appear in the path part of a URI
--- a/lib/puppet/indirector/request.rb
+++ b/lib/puppet/indirector/request.rb
@@ -42,7 +42,9 @@ class Puppet::Indirector::Request
   end
 
   def escaped_key
-    URI.escape(key)
+    # Puppet::Network::HTTP::API::IndirectedRoutes is the only caller to escaped_key
+    # and these fragments only appear in the path part of a URI
+    Puppet::Util.uri_encode(key)
   end
 
   # LAK:NOTE This is a messy interface to the cache, and it's only
@@ -143,7 +145,7 @@ class Puppet::Indirector::Request
 
   def encode_params(params)
     params.collect do |key, value|
-      "#{key}=#{CGI.escape(value.to_s)}"
+      "#{key}=#{Puppet::Util.uri_query_encode(value.to_s)}"
     end.join("&")
   end
 
@@ -243,9 +245,8 @@ class Puppet::Indirector::Request
   def set_uri_key(key)
     @uri = key
     begin
-      # calling URI.escape for UTF-8 characters will % escape them
-      # and the resulting string components of the URI are now ASCII
-      uri = URI.parse(URI.escape(key))
+      # calling uri_encode for UTF-8 characters will % escape them and keep them UTF-8
+      uri = URI.parse(Puppet::Util.uri_encode(key))
     rescue => detail
       raise ArgumentError, "Could not understand URL #{key}: #{detail}", detail.backtrace
     end
@@ -274,8 +275,6 @@ class Puppet::Indirector::Request
       @protocol = uri.scheme
     end
 
-    # The unescaped bytes are correct but in ASCII and must be treated
-    # as UTF-8 for the sake of performing string comparisons later
-    @key = URI.unescape(uri.path.sub(/^\//, '')).force_encoding(Encoding::UTF_8)
+    @key = URI.unescape(uri.path.sub(/^\//, ''))
   end
 end
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -416,13 +416,17 @@ module Util
   #   Note that with no specified scheme, authority or query parameter delimiter
   #  ? that a naked string will be treated as a path.
   #
+  # @param [Hash{Symbol=>String} opts] Options to alter encoding
+  # @option opts [Array<Symbol>] :allow_fragment defaults to false. When false
+  #   will treat # as part of a path or query and not a fragment delimiter
+  #
   # @return [String] a new string containing appropriate portions of the URI
   #   encoded per the rules of RFC3986.
   #   In particular,
   #   path will not encode +, but will encode space as %20
   #   query will encode + as %2B and space as %20
   #   fragment behaves like query
-  def uri_encode(path)
+  def uri_encode(path, opts = { :allow_fragment => false })
     raise ArgumentError.new('path may not be nil') if path.nil?
 
     # ensure string starts as UTF-8 for the sake of Ruby 1.9.3
@@ -441,7 +445,7 @@ module Util
     encoded += URI.escape(parts[:path]) unless parts[:path].nil?
 
     encoded += ('?' + uri_query_encode(parts[:query])) unless parts[:query].nil?
-    encoded += ('#' + uri_query_encode(parts[:fragment])) unless parts[:fragment].nil?
+    encoded += ((opts[:allow_fragment] ? '#' : '%23') + uri_query_encode(parts[:fragment])) unless parts[:fragment].nil?
 
     encoded
   end
--- a/lib/puppet/util.rb
+++ b/lib/puppet/util.rb
@@ -362,6 +362,38 @@ module Util
 
   public
 
+  # Percent-encodes a URI query parameter per RFC3986 - https://tools.ietf.org/html/rfc3986
+  #
+  # The output will correctly round-trip through URI.unescape
+  #
+  # @param [String query_string] A URI query string that may be in the form of:
+  #
+  #   query
+  #   query=foo
+  #   query=foo&query2=bar
+  #   query=foo+bar
+  #   query=hello world are you there
+  #
+  #   Note: Also usable by fragments, but not suitable for paths
+  #
+  # @return [String] a new string containing an encoded query string per the
+  #   rules of RFC3986.
+  #
+  #   In particular,
+  #   query will encode + as %2B and space as %20
+  def uri_query_encode(query_string)
+    return nil if query_string.nil?
+
+    # query can encode space to %20 OR +
+    # + MUST be encoded as %2B
+    # in RFC3968 both query and fragment are defined as:
+    # = *( pchar / "/" / "?" )
+    # CGI.escape turns space into + which is the most backward compatible
+    # however it doesn't roundtrip through URI.unescape which prefers %20
+    CGI.escape(query_string).gsub('+', '%20').gsub('%3D', '=').gsub('%26', '&')
+  end
+  module_function :uri_query_encode
+
   # Percent-encodes a URI string per RFC3986 - https://tools.ietf.org/html/rfc3986
   #
   # Properly handles escaping rules for paths, query strings and fragments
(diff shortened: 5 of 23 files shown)


Iristyle:
Just a quick spike for now after doing a little exploratory work on this ticket - will likely complete during next sprint.

joshcooper:
In general 👍 for using CGI.escape/unescape. I think the main question is whether puppetserver will correctly unescape them, and I believe that is the case, e.g. anytime it sees %XX, it will attempt to decode (otherwise the `%` would have had to been escaped %%XX.)

Iristyle:
URI.escape vs CGI.escape - https://gist.github.com/rcarver/66498

Iristyle:
This should be run through a git based acceptance prior to merge to verify nothing has incidentally been broken.

lib/puppet/file_serving/metadata.rb
```diff
@@ -31,14 +31,14 @@ def source_permissions=(source_permissions)
...
-      uri = URI.parse(URI.escape(path))
+      uri = URI.parse(Puppet::Util.uri_escape(path))
     rescue URI::InvalidURIError => detail
       raise(ArgumentError, "Could not understand URI #{path}: #{detail}")
     end
     raise(ArgumentError, "Cannot use opaque URLs '#{path}'") unless uri.hierarchical?
     raise(ArgumentError, "Must use URLs of type puppet as content URI") if uri.scheme != "puppet"
 
-    @content_uri = path
+    @content_uri = path.encode(Encoding::UTF_8)
```
joshcooper: Can `path` be nil?

lib/puppet/util.rb
```diff
@@ -348,6 +348,18 @@ def uri_to_path(uri)
   end
   module_function :uri_to_path
 
+  # Takes a string containing characters that should be encoded in a Uri
+  # The output is safe to pass to URI.parse or URI::Generic.build  because it
+  # unescapes characters / and : prior to producing a value
+  def uri_escape(path)
+    return unless path
```
joshcooper: This is a slight API change in that `URI.escape` and `CGI.escape` fail with a `nil` parameter, but our `uri_escape` does not. Does that matter?

spec/unit/indirector/rest_spec.rb
```diff
@@ -353,7 +353,7 @@ def save_request(key, instance, options={})
       it "calls get on the connection" do
         request = find_request('foo bar')
 
-        connection.expects(:get).with("#{url_prefix}/test_model/foo%20bar?environment=production&", anything).returns(mock_response('200', 'response body'))
+        connection.expects(:get).with("#{url_prefix}/test_model/foo+bar?environment=production&", anything).returns(mock_response('200', 'response body'))
```
joshcooper: Will this be backwards compatible if an older 4.x agent sends %20 encoded spaces to a 5.0 puppetserver and we unescape in JRuby?

Are there any puppetserver specific REST endpoints we need to worry about? Or will puppetserver/jetty decode both '+' and '%20' as a space? /cc @camlow325 @rlinehan

rlinehan: It looks like puppetserver/jetty does not decode `+`. It does decode `%20` as space.

I tested this on master branch of Puppet Server with both this PR and without (with cef0271da01a4212d6b5ab589cd10f8035efa768 of Puppet, which was HEAD last week) by doing `curl -k https://localhost:8140/puppet/v3/node/foo%20bar?environment=production` and `curl -k https://localhost:8140/puppet/v3/node/foo+bar?environment=production`.

With `foo%20bar` I got `{"name":"foo bar","environment":"production"}` as the response, and saw in the logs:

```
2017-04-04 11:30:20,267 DEBUG [qtp541735488-33] [p.r.core] Processing :get /puppet/v3/node/foo bar
...
2017-04-04 11:30:20,470 DEBUG [qtp541735488-33] [puppetserver] Puppet Did not match path ("/node/foo bar")
...
2017-04-04 11:30:20,500 INFO  [qtp541735488-33] [puppetserver] Puppet Caching node for foo bar
```

With `foo+bar` I got `{"name":"foo+bar","environment":"production"}` as the response, and saw in the logs:

```
2017-04-04 11:30:31,088 DEBUG [qtp541735488-33] [p.r.core] Processing :get /puppet/v3/node/foo+bar
...
2017-04-04 11:30:31,091 DEBUG [qtp541735488-33] [puppetserver] Puppet Did not match path ("/node/foo+bar")
...
2017-04-04 11:30:31,096 INFO  [qtp541735488-33] [puppetserver] Puppet Caching node for foo+bar
```

Given that I got the same results with and without this PR, it does seem like this should be backwards compatible though.

Iristyle: My assumption was that it should behave in a backwards compatible manner - thanks for verifying this in server @rlinehan !

joshcooper: I'm confused, if the agent start encoding spaces as `+`, but puppetserver doesn't decode it as a space, then isn't that backwards incompatible?

rlinehan: So actually, thinking this through a little more, maybe this is a problem just generally? This test seems to imply that the agent will now send `foo+bar` for something with a space, whereas previously it would send `foo%20bar`, but Puppet Server will decode `foo+bar` with a plus, rather than a space.

In the Rack handler, it seems like we're using `CGI.unescape` on the path - https://github.com/puppetlabs/puppet/pull/5738/files/626913a576da17d4823f084936940e040e5d7b2b#diff-031bcd92260c9c0f6e0440c6710552f5R90, which has the behavior of decoding `+` to space. Since this is only happening in the Rack handler, we might see different behavior in Puppet Server than in the unit tests here.

As far as I can tell, we're not explicitly doing any decoding of the path in Puppet Server itself. This might be automatically happening in Jetty or possibly in our Jetty library - I haven't managed to dig into it yet. In any case, the problem here seems not so much one of backwards compatibility, but rather that it's currently not compatible, at least if we expect to have the same behavior.

With that said though, is there anywhere where we can have a `key` that allows a space? It seems like files might cause a problem there, both in some of the `/file*` endpoints and also in a resource `find` with a file resource with a space in its path.

rlinehan: Yeah, basically what @joshcooper said. :P Apparently I didn't refresh.

Iristyle: Yeah, so I think in URIs, the <space> / `%20` / `+` *should* be treated equivalently as far as HTTP goes.  It's more of a concern when strings get parsed afterwards and used as string data - think hash keys and other string comparisons.

Looks like I'll have to do a little bit of investigation into the Jetty code or similar to see how URIs are canonicalized - but it does look based on your logs that the `+` is passed through, which is probably undesirable for backwards compat.

Iristyle: Upon further discussion, there *should* be a distinction for path vs query - thanks for pointing that out @rlinehan - some discussion at http://stackoverflow.com/a/29948396 and https://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape

Looks like I need to take a closer look at this PR to make sure I'm doing the best we can do.

lib/puppet/indirector/request.rb
```diff
@@ -42,7 +42,8 @@ def environment=(env)
   end
 
   def escaped_key
-    URI.escape(key)
+    # TODO: these keys are generally fragments so Puppet::Util.uri_escape isn't really appropriate
+    Puppet::Util.uri_escape(key)
   end
```
Iristyle: Need to determine if this is actually a problem or not... it doesn't make tests fail.

lib/puppet/util.rb
```diff
@@ -349,6 +356,52 @@ def uri_to_path(uri)
...
+    escaped += parts[:authority] unless parts[:authority].nil?
+
+    # path requires space to be encoded as %20 (NEVER +)
+    # + should be left unencoded
+    # URI::parse and URI::Generic.build don't like values from CGI.escape(path)
+    # URI.escape does not change / to %2F and : to %3A like CGI.escape
+    escaped += URI.escape(parts[:path]) unless parts[:path].nil?
+
+    escaped += ('?' + uri_query_escape(parts[:query])) unless parts[:query].nil?
+    # TODO: does fragment have same rules?
```
Iristyle: I believe fragment follows the same rules as query... and commits should be tweaked accordingly

Iristyle:
A few legit failures to deal with on Ruby 1.9.3... AppVeyor failures are the bug in their image affecting UNC - https://github.com/appveyor/ci/issues/1579

lib/puppet/util.rb
```diff
@@ -349,6 +356,52 @@ def uri_to_path(uri)
...
   module_function :uri_to_path
 
+  private
+
+  RFC_3986_URI_REGEX = /^(?<scheme>([^:\/?#]+):)?(?<authority>\/\/([^\/?#]*))?(?<path>[^?#]*)(\?(?<query>[^#]*))?(#(?<fragment>.*))?$/
+
+  public
+
+  # Will escape a query parameter or any portion of the query string
+  # so that it will remain unescapable with URI.unescape
```
MosesMendoza: Minor but any chance you wouldn't mind rewording this comment to something like `so that it can still be unescaped with URI.unescape`? I originally read this as saying the opposite of what's intended, ie "so that cannot be escaped with URI.unescape"

lib/puppet/configurer/fact_handler.rb
```diff
@@ -32,6 +32,6 @@ def facts_for_uploading
 
     text = facts.render(:pson)
 
-    {:facts_format => :pson, :facts => CGI.escape(text)}
+    {:facts_format => :pson, :facts => Puppet::Util.uri_query_escape(text)}
```
MosesMendoza: just fyi this code has changed in master so this might need some fixup on merge-up

Iristyle: Bummer... I was hoping to get this one in first ;0

`CGI.escape` is *mostly* correct, except when it comes to spaces - hence our own helper ;(

lib/puppet/file_serving/metadata.rb
Iristyle: Nope, not anymore - `uri_escape` will throw when `path` is `nil` (given I fixed it per your suggestion to match previous behavior)

lib/puppet/indirector/request.rb
MosesMendoza: I think the key is mostly (always?) part of the path and not the query so if I'm understanding `uri_escape` correctly this is the right thing to do.

MosesMendoza: Unless we're talking about different `key`s

Iristyle: I was just looking into this... the only caller to this method is at https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/http/api/indirected_routes.rb#L228 which means it always appears in the path of the URI.

I think I may remove this method and inline it into the single callsite...

lib/puppet/util.rb
```diff
@@ -349,6 +356,53 @@ def uri_to_path(uri)
...
+  # The output is safe to pass to URI.parse or URI::Generic.build  because it
+  # unescapes characters / and : prior to producing a value
+  def uri_escape(path, opts = { :allow_fragment => false })
+    raise ArgumentError.new('path may not be nil') if path.nil?
+
+    # ensure string starts as UTF-8 for the sake of Ruby 1.9.3
+    escaped = ''.encode!(Encoding::UTF_8)
+
+    # parse uri into named matches, then reassemble properly escaped
+    parts = path.match(RFC_3986_URI_REGEX)
```
MosesMendoza: Maybe a silly question but is there any chance `path` is going to contain an invalid byte sequence?

spec/integration/type/file_spec.rb
```diff
@@ -1074,6 +1074,35 @@ def build_path(dir)
...
+      resource = catalog.resources.first { |r| r.title == "File[#{dest}]" }
+      uri_path = resource.parameters[:source].uri.path
+
+      # note that Windows file:// style URIs get an extra / in front of c:/ like /c:/
+      source_prefix = Puppet.features.microsoft_windows? ? '/' : ''
+
+      # the URI can be round-tripped through unescape
+      expect(URI.unescape(uri_path)).to eq(source_prefix + source)
+      # and is properly UTF-8
+      expect(uri_path.encoding).to eq (Encoding::UTF_8)
```
MosesMendoza: nit: extra space

Iristyle:
Ok @MosesMendoza / @joshcooper - I think this is ready for a final pass.  If we decide that af45a27 is too much of a breaking change, we can remove it.

acceptance/tests/reports/submission.rb
```diff
@@ -32,7 +32,7 @@ def query_last_report_time_on(agent)
       require "json"
 
       puppetdb_url = URI("http://localhost:8080/v3/reports")
-      puppetdb_url.query = URI.escape(%Q{query=["=","certname","#{agent}"]})
+      puppetdb_url.query = CGI.escape(%Q{query=["=","certname","#{agent}"]})
```
Iristyle: I'm actually not sure if this is correct given `CGI.escape` also encodes `=` which is likely undesirable here (but depends on the puppetserver implementation):

```
[2] pry(main)> CGI.escape(%Q{query=["=","certname","foo.bar.com"]})
=> "query%3D%5B%22%3D%22%2C%22certname%22%2C%22foo.bar.com%22%5D"

[3] pry(main)> URI.escape(%Q{query=["=","certname","foo.bar.com"]})
=> "query=[%22=%22,%22certname%22,%22foo.bar.com%22]"
```

Though either way CGI.unescape produces the correct original value:

```
[8] pry(main)> query = %Q{query=["=","certname","foo.bar.com"]}
=> "query=[\"=\",\"certname\",\"foo.bar.com\"]"
[9] pry(main)> URI.unescape(URI.escape(query)) == query
=> true
[10] pry(main)> URI.unescape(CGI.escape(query)) == query
=> true
```

lib/puppet/util.rb
Iristyle: Inline docs completely overhauled

lib/puppet/util.rb
Iristyle: Shouldn't ... but I'm OK with an exception being generated here if it does.

lib/puppet/util.rb
```diff
@@ -349,6 +356,98 @@ def uri_to_path(uri)
...
+  # Percent-encodes a URI query parameter per RFC3986 - https://tools.ietf.org/html/rfc3986
+  #
+  # The output will correctly round-trip through URI.unescape
+  #
+  # @param [String query_string] A URI query string that may be in the form of:
+  #
+  #   query
+  #   query=foo
+  #   query=foo+query2=bar
+  #   query=hello world are you there
```
MosesMendoza: Can we also handle multiple query params, ie `query=foo&bar=baz` ?

lib/puppet/util.rb
```diff
@@ -349,6 +356,98 @@ def uri_to_path(uri)
...
+  # correctly round-trip through URI.unescape
+  #
+  # @param [String path] A URI string that may be in the form of:
+  #
+  #   http://foo.com/bar?query
+  #   //foo.com/bar?query
+  #   /bar?query
+  #   bar?query
+  #   bar
+  #   .
```
MosesMendoza: Worth noting handling of windows file uri?

MosesMendoza:
I'm find with removing `Puppet::Indirector::Request#escaped_key` in master (for 5.0) but maybe that should just be a deprecation for stable?

Iristyle:
Agreed on that final commit @MosesMendoza

MosesMendoza:
👍  - good for squash

MosesMendoza:
Note - this will require a separate PR for merge-up to master as some code paths have changed

Iristyle:
Merged up to master in https://github.com/puppetlabs/puppet/commit/70af0de51e56e4e95f3b421eeb40b0e205fc549b

lib/puppet/util.rb
```diff
@@ -349,6 +356,101 @@ def uri_to_path(uri)
...
+  def uri_query_encode(query_string)
+    return nil if query_string.nil?
+
+    # query can encode space to %20 OR +
+    # + MUST be encoded as %2B
+    # in RFC3968 both query and fragment are defined as:
+    # = *( pchar / "/" / "?" )
+    # CGI.escape turns space into + which is the most backward compatible
+    # however it doesn't roundtrip through URI.unescape which prefers %20
+    CGI.escape(query_string).gsub('+', '%20').gsub('%3D', '=').gsub('%26', '&')
```
Iristyle: The decoding of `%26` to `&` has caused The problem reported in https://tickets.puppetlabs.com/browse/PUP-7658.

`uri_query_encode` conceptually works over an entire query string - and unencodes any `&` that may have been encoded, though this may be a bit naive.  

Within `fact_handler.rb`, all of the fact data should be encoded, else it injects delimiters into values that end up being URI percent encoded OR form-encoded.  The correct helper to use there is actually [URI.encode_www_form_component(str)](https://ruby-doc.org/stdlib-1.9.3/libdoc/uri/rdoc/URI.html#method-c-encode_www_form_component)

PR #5740, opened 2017-03-21
(maint) Make shell scripts executable

kurtwall:
PR-5432 removed execute bits from a bunch of ruby test files. This PR restores them on two shell scripts I inadvertently modified as well in the original PR.


ericldelaney:
👍

PR #5741, opened 2017-03-21
(maint) Ensure that lookup CLI initializes loaders

thallgren:
Under some very rare circumstances, such as when the only available
data provider for Hiera is a custom backend written in Ruby, the
Pcore loaders would not be initialized by the lookup CLI command. This
commit ensures that they are.


joshcooper:
Is there a ticket number that this PR fixes (ideally the original ticket that introduced the failure, not the CI blocker ticket)?

lib/puppet/application/lookup.rb
```diff
@@ -339,6 +339,7 @@ def generate_scope
 
     Puppet[:code] = 'undef' unless options[:compile]
     compiler = Puppet::Parser::Compiler.new(node)
+    compiler.loaders # ensures that Pcore et.al. is initialized
```
joshcooper: Does this pattern need to be done elsewhere? Thinking of `puppet apply` or other applications that create a `Compiler`?

thallgren: No. The loaders are initialized as soon as any pp code is executed.

thallgren:
@joshcooper, I don't know if anyone has entered a ticket for the CI failure. You reported it to me in Hipchat. Did you enter at ticket?

joshcooper:
PUP-7330 was committed to stable in 12b8ab94cd75c3db9feb858c4eccf61d9b6940db, so doesn't this PR need to target the same? Can you git bisect to confirm what exact commit introduced the problem so we can target this PR correctly?

thallgren:
Oh, I thought that went into 4.10.x. My bad. I rebased this to stable. It is the commit you mentioned.

thallgren:
And of course, that rebase was a no-go. Closing this and reopening with another base branch.

PR #5742, opened 2017-03-21
(PUP-7330) Ensure that lookup CLI initializes loaders

thallgren:
Under some very rare circumstances, such as when the running the lookup
CLI command and the only available data provider for Hiera was a custom
backend written in Ruby, the Pcore would not be initialized by the
compiler since no loader was ever requested. This in turn, resulted in
that the `Puppet::LookupKey` and `Puppet::LookupValue` types never got
initialized.

This commit changes the compiler so that it performs the task of
initializing loaders and Pcore during its own initialization instead of
performing it lazily when the loaders are first requested.

The change in behavior broke a couple of tests that made assumptions on
the deferred loader initialization. This commit fixes those tests as
well.

The problem was triggered by commit 12b8ab94cd75c3db9feb858c4eccf61d9b6940db

### DIFF
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -217,7 +217,7 @@ class Puppet::Parser::Compiler
     {
       :current_environment => environment,
       :global_scope => @topscope,             # 4x placeholder for new global scope
-      :loaders  => lambda {|| loaders() },    # 4x loaders
+      :loaders  => @loaders,                  # 4x loaders
       :injector => lambda {|| injector() }    # 4x API - via context instead of via compiler
     }
   end
@@ -357,7 +357,7 @@ class Puppet::Parser::Compiler
     end
   end
 
-  # Evaluates each specified class in turn. If there are any classes that 
+  # Evaluates each specified class in turn. If there are any classes that
   # can't be found, an error is raised. This method really just creates resource objects
   # that point back to the classes, and then the resources are themselves
   # evaluated later in the process.
@@ -440,10 +440,6 @@ class Puppet::Parser::Compiler
   def injector
     create_injector if @injector.nil?
     @injector
-  end
-
-  def loaders
-    @loaders ||= Puppet::Pops::Loaders.new(environment)
   end
 
   def boot_injector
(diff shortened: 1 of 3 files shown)


thallgren:
Superseeds [PR-5741](/puppetlabs/puppet/pull/5741)

joshcooper:
@thallgren since this PR is part of the necessary changes for PUP-7330, could you add that to the git commit summary?

lib/puppet/application/lookup.rb
```diff
@@ -339,6 +339,7 @@ def generate_scope
 
     Puppet[:code] = 'undef' unless options[:compile]
     compiler = Puppet::Parser::Compiler.new(node)
+    compiler.loaders # ensures that Pcore et.al. is initialized
```
hlindberg: Doesn't this mean that you would always need to follow a call to `Compiler.new` with this call? 
And then, doesn't that mean that this potentially break other use cases of compiler? Like catalog preview (to mention one of ours) or puppet debug from the community?

hlindberg:
Agree with @joshcooper - this should not be a (maint) ticket. Either target PUP-7330 (probably best), or a separate ticket.

lib/puppet/application/lookup.rb
thallgren: No, it doesn't. The only code-path where this problem can occur is if ruby code is trying to access the loader without any pp code has been loaded. Catalog preview will not do that since it doesn't call lookup by itself.

thallgren:
Commit amended to target PUP-7330.

lib/puppet/application/lookup.rb
joshcooper: Is there a way to ensure pp code has been loaded when the compiler is created? It just seems strange that we create a compiler object, but it's not fully functional.

thallgren:
@joshcooper , @hlindberg , I found a better solution. It required some changes of tests that made assumptions about internal behavior but those were assumptions that they really should not rely on in the first place.

hlindberg:
There was (at least once upon a time) a very good reason for the loaders to be initialized lazily. See PUP-2176 where this was introduced 

> (PUP-2176) Use lazy initialization of context in compiler
> 
> This makes the compiler initialize its context lazily wrt. the
> injector and loader when using the future branch of the logic.
> 
> This is required because loaders and injector would otherwise
> need to be initialized twice; once for the initial import and
> a second time when doing the compilation. Now the values are retained.
> 
> The initial problem was that compiler's initial import is triggered
> before loaders and injector was initialized (from initvars). This is
> required for using logic from the loader subsystem. Specifically for
> being able to define a function on the command line using puppet apply.

Maybe other things have now changed so this is no longer needed.

hlindberg:
I looked at the concern I raised above - the branches of the logic that caused the problems fixed by PUP-2176 no longer exist for loaders.

hlindberg:
I think this is good to merge now. The comment @joshcooper made has been addressed (targeting PUP-7330). My concern that the lazy init could still be of value, or possible affect performance was unfounded. Merging.

joshcooper:
Thanks @thallgren and @hlindberg, changes look good and cleaner too.

PR #5744, opened 2017-03-21
(maint) Only ci:test:quick ruby files

joshcooper:
Previously, the quick task was trying to run `.sh` files as beaker tests
which were recently modified.

This commit ensures we only try to run files ending in `.rb`. If those
are not beaker tests then they need to be in a different directory
other than `tests`.

Depends on https://github.com/puppetlabs/puppet/pull/5743.

PR #5745, opened 2017-03-21
(maint) Only ci:test:quick ruby files

joshcooper:
Previously, the quick task was trying to run `.sh` files as beaker tests
which were recently modified.

This commit ensures we only try to run files ending in `.rb`. If those
are not beaker tests then they need to be in a different directory
other than `tests`.

Supersedes PR #5744.

[skip ci]


Magisus:
👍

PR #5748, opened 2017-03-22
(PUP-7237) Restore BEAKER_HOSTS env variable support

joshcooper:
Commit fb9ab3b18 removed the ability to specify a pre-generated beaker
hosts config file using the `BEAKER_HOSTS` environment variable which is
used in ci-job-configs. Instead the rake task was combining the default
master platform and the `TEST_TARGET`, which already contains the master
platform resulting in a BHG string like:

    redhat7-64m-redhat7-64m-redhat6-64a

However, this causes the beaker run to fail because there are two master
platforms specified.

This commit restores the ability to specify a pre-generated beaker hosts
config file using `BEAKER_HOSTS`.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -108,8 +108,10 @@ EOS
   tests_opt = "--tests=#{tests}" if tests
 
   overriding_options = ENV['OPTIONS'].to_s
+  # preserved hosts take precedence
+  hosts = options[:hosts] || ENV['HOSTS']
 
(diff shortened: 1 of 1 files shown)


joshcooper:
I've tested this with `ci:test:aio`, `ci:test:quick`, `ci:test:gem` and when specifying an external config file using `HOSTS`, `BEAKER_HOSTS`, and `CONFIG`. 'm going to merge this to unblock CI.

MosesMendoza:
wow gettin pretty squirrelly in the Rakefile.

👍

PR #5749, opened 2017-03-22
(PUP-7334) Add Hiera default_hierarchy to module hiera.yaml

thallgren:
This commit adds the ability to declare a second hierarchy named
`default_hierarchy` to a hiera.yaml file that resides in a module root.
This hierarchy is only allowed in modules and will be consulted when, and
only when, a normal lookup does not find a value.

The `default_hierarchy` is insensitive to `lookup_options` declared in
the regular `hierarchy` and vice versa.

The `default_hierarchy` is also insensitive to any merge directives
passed in as a parameter in the lookup call.

### DIFF
--- a/spec/unit/functions/lookup_spec.rb
+++ b/spec/unit/functions/lookup_spec.rb
@@ -282,6 +282,24 @@ describe "The lookup function" do
           end
         end
 
+        context 'with a declared default_hierarchy' do
+          let(:hiera_yaml) { <<-YAML.unindent }
+            version: 5
+            hierarchy:
+              - name: Common
+                path: common.yaml
+            default_hierarchy:
+              - name: Defaults
+                path: defaults.yaml
+            YAML
+
+          it 'fails and reports error' do
+            Puppet[:strict] = :error
+            expect { lookup('a') }.to raise_error(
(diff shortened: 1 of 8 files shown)


spec/unit/functions/lookup_spec.rb
```diff
@@ -2503,6 +2539,119 @@ def ruby_dig(segments, options, context)
...
+
+        it 'lookup_options from regular hierarchy does not effect values found in the default hierarchy' do
+          expect(lookup('mod_a::e')).to eql('a' => { 'a' => 'value mod_a::e.a.a (from module defaults)' })
+        end
+
+        it 'lookup_options from default hierarchy affects values found in the default hierarchy' do
+          expect(lookup('mod_a::g')).to eql('a' => { 'a' => 'value mod_a::g.a.a (from module defaults)', 'b' => 'value mod_a::g.a.b (from module defaults)'})
+        end
+
+        it 'lookup_options given as parameter overrides lookup_options from the default hierarchy' do
```
hlindberg: This is wrong - you should not be able to override the lookup_options in the defaults_hierarchy

hlindberg:
A test for explain output of a lookup that goes into a default hierarchy is needed.

PR #5750, opened 2017-03-22
(PUP-7372) Confine Hiera 3 interpolations to the global layer

thallgren:
This commit ensures that a data file that is reached using a Hiera 3
configuration and that in turn contains interpolation expressions that
use the 'hiera' or 'lookup' function, will have that lookup confined
to the global layer unless an environment data provider, version 5
is present. It also ensures that when an interpolation that originas
from Hiera 3 is calling out, that the special `Hiera::Scope` class
is unwrapped.

This was not the case earlier which resulted in that the special class
`Hiera::Scope` was passed on to the environment and module layers which
resulted in errors as they tried to access the `Scope#environment`
method which is not present in that class. Other side-effects of passing
that scope would be to make the special variables that it contains
available to data configured using Hiera 5.

### DIFF
--- a/lib/puppet/pops/lookup/global_data_provider.rb
+++ b/lib/puppet/pops/lookup/global_data_provider.rb
@@ -15,11 +15,16 @@ class GlobalDataProvider < ConfiguredDataProvider
       # Hiera version 3 needs access to special scope variables
       scope = lookup_invocation.scope
       unless scope.is_a?(Hiera::Scope)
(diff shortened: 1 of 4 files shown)


hlindberg:
interpolation needs to follow the same rules as the hiera_xxx calls wrt drilling down into other layers (an hiera.yaml in the env is taken as opt in to drill down)

hlindberg:
good to merge when tests go green

PR #5754, opened 2017-03-22
(PUP-7390) Prefer minitar for PMT when available

Iristyle:
- For platforms that ship a minitar gem (currently Windows, about to
   include Solaris, and will eventually be *all* platforms), prefer
   that to a local `tar` binary.

   Solaris 10 currently doesn't work properly with the `tar` binary
   and will be soon bundled with the minitar gem.  However, to use
   the minitar gem, the priority ordering needs to be adjusted.

 - Currently only Windows ships the minitar gem (as it has a platform
   specific gem), but eventually all platforms will reference the
   minitar gem for the same reason - consistency and robustness.

### DIFF
--- a/lib/puppet/module_tool/tar.rb
+++ b/lib/puppet/module_tool/tar.rb
@@ -6,10 +6,10 @@ module Puppet::ModuleTool::Tar
   require 'puppet/module_tool/tar/mini'
 
   def self.instance
-    if Puppet::Util.which('tar') && ! Puppet::Util::Platform.windows?
-      Gnu.new
-    elsif Puppet.features.minitar? && Puppet.features.zlib?
+    if Puppet.features.minitar? && Puppet.features.zlib?
       Mini.new
+    elsif Puppet::Util.which('tar') && ! Puppet::Util::Platform.windows?
+      Gnu.new
     else
       raise RuntimeError, 'No suitable tar implementation found'
     end
--- a/spec/unit/module_tool/tar_spec.rb
+++ b/spec/unit/module_tool/tar_spec.rb
@@ -3,21 +3,27 @@ require 'puppet/module_tool/tar'
 
 describe Puppet::ModuleTool::Tar do
 
-  it "uses tar when present and not on Windows" do
+  [
+    { :name => 'ObscureLinuxDistro', :win => false }, 
+    { :name => 'Windows', :win => true }
+  ].each do |os|
+    it "always prefers minitar if it and zlib are present, even with tar available" do
+      Facter.stubs(:value).with('osfamily').returns os[:name]
+      Puppet::Util.stubs(:which).with('tar').returns '/usr/bin/tar'
+      Puppet::Util::Platform.stubs(:windows?).returns os[:win]
+      Puppet.stubs(:features).returns(stub(:minitar? => true, :zlib? => true))
+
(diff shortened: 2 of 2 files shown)


McdonaldSeanp:
👍

Iristyle:
We'll probably want to validate this on a few platforms via acceptance / manual testing prior to merge given there aren't any real tests.

joshcooper:
Should this retarget master since PUP-7390 is scheduled for 5.y?

Iristyle:
I was thinking we'd be targeting the Solaris PMT install bugfix at the LTS release, hence stable @joshcooper

MosesMendoza:
Looks like needs a spec fixup. Can we retarget PUP-7390 to 4.10.0 if going for LTS? 👍  pending acceptance run

Iristyle:
@joshcooper / @MosesMendoza / @McdonaldSeanp are we shooting for next `4.10` series release on this one?

joshcooper:
If we make this change in 4.x, then it means minitar must be able to unpack any module that has previously been built using native tar (except for Windows which has only ever supported minitar). Do we know if minitar can handle it? The PAX header thing comes to mind (which we fixed), but I don't know if there are other issues like that. Unless we're sure, I'd suggest targeting puppet 5 instead.

Iristyle:
It's my understanding that for packages, we only ship minitar to Windows, and we haven't enabled minitar (yet) on Solaris packages.  For gems, **only** the Windows platform-specific gems refer to minitar (i.e. https://rubygems.org/gems/puppet/versions/4.9.4-x64-mingw32).

To answer the question about "native tar" - there isn't really such a thing.  There are various flavors of the file format (see https://en.wikipedia.org/wiki/Tar_(computing)#File_format or the notes from myself and Glenn on https://tickets.puppetlabs.com/browse/MODULES-3135).  We have had no guarantees previously in the forge about what format a given module tarball is in - even with Puppet modules, old modules prepared / shipped from local dev boxes will have been tar'd in a different format (OSX uses bsdtar) than ones shipped from jobs running on Linux.  The only tarball that I'm aware of that caused an issue was the one for DSC due to long filenames and the originating tar format - as you've noted, this has been fixed.

Given Windows has already been using `minitar` for a long time and we don't have reports of failures AND given that I don't believe we actually ship `minitar` anywhere else yet - I think we're OK on merging this PR.

Iristyle:
The reason I'm eager to get this in is so that we can make PMT on Solaris functional - see [PA-1019 Vendor minitar gem for Solaris packages](https://tickets.puppetlabs.com/browse/PA-1019) - this is currently targeted at agnet 1.10.1 and has a corresponding PE ticket / escalation ([PE-19583](https://tickets.puppetlabs.com/browse/PE-19583)).

joshcooper:
>  we only ship minitar to Windows, and we haven't enabled minitar (yet) on Solaris packages

Ah right, forgot about that: https://github.com/puppetlabs/puppet-agent/blob/a0cbf6ff51c3c78caa7b59b6e80f08429d6602ac/configs/projects/puppet-agent.rb#L233

> To answer the question about "native tar" - there isn't really such a thing. 

Yeah, I was using "native tar" to mean gnu tar, bsd tar, etc whatever tar happens to be in the PATH, with it's platform specific behaviors. There's a reason Java uses Jar files!

PR #5755, opened 2017-03-23
(PUP-7380) Skip augesu test on Cisco XR

LuvCurves:
We have found that the new Cisco eXR 6.1.3 VM image has /etc/hosts
as a mount point, so the augeus host test fails to rename the it.


ericldelaney:
👍

PR #5756, opened 2017-03-23
(PUP-893) Fix audit tests and revert revert

Magisus:
We recently deprecated the `audit` resource metaparameter, which broke a
test that was expecting no output on stderr. The test has been updated
to expect a deprecation but no errors.

This PR also reverts the revert, reinstating the deprecations of the `audit` metaparameter and the `inspect` command.

### DIFF
--- a/lib/puppet/application.rb
+++ b/lib/puppet/application.rb
@@ -322,6 +322,15 @@ class Application
     @options = {}
   end
 
+  # call in setup of subclass to deprecate an application
+  def deprecate
+    @deprecated = true
+  end
+
+  def deprecated?
+    @deprecated
+  end
+
   # Execute the application.
   # @api public
   # @return [void]
@@ -339,6 +348,11 @@ class Application
     exit_on_fail("initialize")                                   { preinit }
     exit_on_fail("parse application options")                    { parse_options }
     exit_on_fail("prepare for execution")                        { setup }
+
+    if deprecated?
+      Puppet.deprecation_warning(_("`puppet %{name}` is deprecated and will be removed in a future release.") % { name: name })
+    end
+
     exit_on_fail("configure routes from #{Puppet[:route_file]}") { configure_indirector_routes }
     exit_on_fail("log runtime debug info")                       { log_runtime_environment }
     exit_on_fail("run")                                          { run_command }
--- a/lib/puppet/application/inspect.rb
+++ b/lib/puppet/application/inspect.rb
@@ -19,6 +19,7 @@ puppet-inspect(8) -- Send an inspection report
 
 SYNOPSIS
 --------
+Note: this command is deprecated
 
 Prepares and submits an inspection report to the puppet master.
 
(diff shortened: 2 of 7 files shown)


acceptance/tests/direct_puppet/cached_catalog_remediate_local_drift.rb
```diff
@@ -89,16 +89,18 @@
...
+      on(agent, puppet("agent",
         "--use_cached_catalog",
         "--server", "no_such_host",
         "--report_server", master.hostname,
         "--onetime",
         "--no-daemonize",
         "--detailed-exitcodes",
         "--verbose"
-      )).stderr
-      assert_equal(r, "", "Fail: Did agent try to contact server?")
```
joshcooper: Why is a deprecation warning being issue if the catalog doesn't contain `audit`?

Magisus: So I was wondering that. I assumed it was something beaker was doing with some helper method that I didn't understand >.>

joshcooper: I'm wondering if another test left `archive_*` setting behind causing this test to fail.

Magisus: A basic manifest definitely does NOT issue one in my local manual testing.

Magisus: No, that warning message is different. This is definitely the one that comes from the parameter...

joshcooper: it looks like `puppet resource <type> <name>` for a resource that doesn't exist will cause the `resource` application to programmatically set the `audit` metaparam triggering the deprecation warning.

Magisus: That particular step of the test does not issue any deprecation warnings on Windows:
```
* Verify puppet run without drift does not make file request from server
  
  qkdrnn9879m6bnl.delivery.puppetlabs.net (windows2012r2-64-1) 18:41:43$ cmd.exe /c puppet agent --use_cached_catalog --server no_such_host --report_server x6rgarwz0fz5piz.delivery.puppetlabs.net --onetime --no-daemonize --detailed-exitcodes --verbose
    Info: Using cached catalog from environment 'production'
    Info: Applying configuration version '1490294499'
  
  qkdrnn9879m6bnl.delivery.puppetlabs.net (windows2012r2-64-1) executed in 9.28 seconds
  Warning: 
  the audit paramter should issue a deprecation warning.
  Expected /The `audit` metaparameter is deprecated/ to match # encoding: US-ASCII
  "".
```

I'm going to update this to just assert that there are no errors.

Magisus:
Closing this in favor of a PR that only issues the deprecation warning for `audit` when the param is used in a manifest.

PR #5759, opened 2017-03-23
(PUP-7392) Verify that stopping puppet svc on Windows stops it

kurtwall:
PUP-2987 removed some race conditions in code around stopping the `puppet` service on Windows platforms. This PR verifies that stopping the `puppet` service on Windows does, actually, eventually stop the service.

### DIFF
--- a/acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
+++ b/acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
@@ -1,0 +1,47 @@
+test_name '(C14941) Stopping puppet service on Windows stops puppet' do
+
+  confine :to, :platform => 'windows'
+  
+  opts = {
+    :acceptable_exit_codes => [0],
+  }
+    
+  # Wait for the puppet service to enter the requested state and return
+  # after at most 5 attempts 
+  def wait_for_windows_puppet_service_state(state, options=opts)
+    options ||= {
+      :acceptable_exit_codes => [0],
+    }
+
+    query = on(agent, 'sc query puppet', options)
+    unless query.stdout.include?(state)
+      repeat_fibonacci_style_for 5 do
+        result = on(agent, 'sc query puppet', options)
+        return result if result.stdout.include?(state)
+      end
+    end
+    query = on(agent, 'sc query puppet', options)
+    assert_match(/#{state}/, result.stdout)
+  end
+
+  agents.each do
+    step 'Verify the puppet service is running' do
+      # Exit code 32 means the service is already running, so add this
+      # value at block scope to the list of okay exit codes
+      opts[:acceptable_exit_codes] << 32
+      start = on(agent, 'sc start puppet', opts)
+      wait_for_windows_puppet_service_state('RUNNING', opts)
+    end
+   
+    step 'Stop the puppet service and verify it stopped' do
+      stop = on(agent, 'sc stop puppet', opts)
+      wait_for_windows_puppet_service_state('STOPPED')
+    end
+  
+    step 'Restart the puppet service and verify it restarted' do
+      restart = on(agent, 'sc start puppet', opts)
+      wait_for_windows_puppet_service_state('RUNNING')
+    end
+  end
+
+end


acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
```diff
@@ -0,0 +1,34 @@
...
+
+confine :to, :platform => 'windows'
+
+agents.each do
+  step 'Make sure puppet service is running' do
+    # returns 32 if service is already running
+    result = on(agent, 'sc start puppet', :acceptable_exit_codes => [0, 32])
+    if result.stdout.include?("START_PENDING")
+      sleep 5
+    end
```
joshcooper: I think there's a beaker helper to retry logic until a certain condition is true or a timeout is reached. /cc @er0ck

kurtwall: @joshcooper I made the change your suggested. It worked in 2 of 3 cases, but on the 3rd case I encountered a wrinkle in the way Beaker's `retry_on` method decides when to return. See the FIXME comment in the latest PR. I'm open to suggestions for something better than `sleep`.

acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
```diff
@@ -0,0 +1,44 @@
+test_name '(C14941) Stopping puppet service on Windows stops puppet'
```
er0ck: needs `do` `end` closure

acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
```diff
@@ -0,0 +1,44 @@
...
+
+opts = {
+  :max_retries => 2,
+  :retry_interval => 5,
+}
+
+agents.each do
+  step 'Make sure puppet service is running' do
+    query = on(agent, 'sc query puppet', :acceptable_exit_codes => [0])
+    if ! query.stdout.include?('RUNNING')
```
er0ck: this should probably be `unless`

acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  step 'Restart the puppet service and verify it restarted' do
+    restart = on(agent, 'sc start puppet', :acceptable_exit_codes => [0])
+    if ! restart.stdout.include?('RUNNING')
+      ### FIXME: Doesn't pass without the sleep; the query command succeeds
+      ### so retry_on returns before we get the status we want. The 
+      ### sc command will succeed while the service status remains
+      ### START_PENDING. As a result, the test fails. The hard sleep 
+      ### gives the service "enough" time to start
+      sleep 5
```
er0ck: can you fibonacci retry until `stdout.include?('RUNNING')`, instead of a hard-coded sleep?

er0ck: https://github.com/puppetlabs/beaker/blob/master/lib/beaker/shared/repetition.rb#L14

not sure if that helps your success conditions (might need better options for the various `repeat` methods in beaker).

kurtwall: I'll give that a look.

acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
kurtwall: Yeah. Fixed.

acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
kurtwall: Fixed.

kurtwall:
@er0ck Pushed some changes to address your review. Decided to use `assert_match` too. Changed the retry logic. Etc.

acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  confine :to, :platform => 'windows'
+  
+  opts = {
+    :acceptable_exit_codes => [0],
+  }
+  
+  agents.each do
+    step 'Verify the puppet service is running' do
+      opts[:acceptable_exit_codes] << 32
```
er0ck: i had to think about this for a second.  could probably use a comment to say we WANT to exit successfully on either 0 or 32 to confirm `RUNNING` later

acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
```diff
@@ -0,0 +1,48 @@
...
+      start = on(agent, 'sc query puppet', opts)
+      unless start.stdout.include?('RUNNING')
+        repeat_fibonacci_style_for 5 do
```
er0ck: are we sure 5 is enough all the time?

acceptance/tests/windows/stop_puppet_svc_stops_puppet.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+   
+    step 'Stop the puppet service and verify it stopped' do
+      stop = on(agent, 'sc stop puppet', opts)
+      unless stop.stdout.include?('STOPPED')
+        repeat_fibonacci_style_for 5 do
+          stop = on(agent, 'sc query puppet', opts)
+          break if stop.stdout.include?('STOPPED')
+        end
+      end
```
er0ck: this block is repeated across the three steps.  could we use a local method of `def wait_for_windows_puppet_run_state(state='STOPPED')` or similar?
which would be a good example of possible reuse when someone finds it...

joshcooper:
I just remembered we have a fairly comprehensive service test in https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/resource/service/puppet_mcollective_service_management.rb. It's currently [skipped on windows](https://github.com/puppetlabs/puppet/blob/master/acceptance/tests/resource/service/puppet_mcollective_service_management.rb#L3) though I'm not sure why we're not testing puppet and only skipping mco.

kurtwall:
@joshcooper Thanks for the linkage. There's some similarity here or at least some common code paths covered. The intent of my test is to catch a hang caused by a race condition. The test you mentioned seems to be about testing the ability of puppet to manage itself and mcollective as services. Related, to be sure, but different emphases.

I'd be happy to look at the other test and see about unskipping Windows, adding puppet, and so forth.

PR #5761, opened 2017-03-24
(PUP-7360) Strip patch version from deprecation message URL

thallgren:
The URL contained in the deprecation message from some functions and
also is present in many function API docs, incorrectly contained the
full major.minor.patch triplet. This commit ensures that the patch
number is no longer included.

### DIFF
--- a/lib/puppet/version.rb
+++ b/lib/puppet/version.rb
@@ -72,6 +72,10 @@ module Puppet
     @puppet_version = version
   end
 
+  def self.major_minor_version
+    self.version.sub(/\A(\d+\.\d+)\..*$/, '\1')
+  end
+
   ##
   # read_version_file reads the content of the "VERSION" file that lives in the
   # same directory as this source code file.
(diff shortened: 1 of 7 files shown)


lib/puppet/version.rb
```diff
@@ -72,6 +72,10 @@ def self.version=(version)
     @puppet_version = version
   end
 
+  def self.major_minor_version
+    self.version.sub(/\A(\d+\.\d+)\..*$/, '\1')
+  end
```
joshcooper: Can you add a unit test for this? Btw, there's a related PR in https://github.com/puppetlabs/puppet/pull/5720, with some discussion about using semantic puppet to parse the version string. Would be good to consolidate these PRs. /cc @binford2k @MosesMendoza

thallgren: I changed one of the tests for the hiera function. It checks that the URL contains only major and minor version. I think that is sufficient to test this.

hlindberg: There isn't much to consolidate - they are basically the same except that the name chosen for the method is different.
While we could use semantic puppet to parse, we also know that we are using a basic standard versioning scheme with three digits. If we change that then we need to change here as well.

thallgren: I reviewed [PR-5720](/puppetlabs/puppet/pull/5720) and added a couple of comments there. I suggest we close this one and use that instead as it contains all needed tests.

joshcooper:
Closing based on @thallgren comment

PR #5764, opened 2017-03-27
(PUP-6084) Make reference to Class[<upper case name>] be an error

hlindberg:
This removes the deprecated (since Puppet 4.6.0) construct of
referencing a class with an upper case name, for example Class[Foo] by
making it always be an error.

This is removed since Foo evaluates to a type reference, and it may be a
type alias or a Resource. It may also be a type that requires
parameters. The use of an upper case name thus creates unssesary type
loading, and potentially a mysterious error.

The correct reference is always a String, and this string is always in
lower case.

### DIFF
--- a/lib/puppet/pops/evaluator/access_operator.rb
+++ b/lib/puppet/pops/evaluator/access_operator.rb
@@ -616,6 +616,8 @@ class AccessOperator
     return result_type_array ? result : result.pop
   end
 
+  NS = '::'.freeze
+
   def access_PHostClassType(o, scope, keys)
     blamed = keys.size == 0 ? @semantic : @semantic.keys[0]
     keys_orig_size = keys.size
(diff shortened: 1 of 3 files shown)


MosesMendoza:
@hlindberg could this potentially force users to change their modules/manifest code?

hlindberg:
@MosesMendoza yes - they would have seen warnings since Puppet 4.6.0 though unless they turned them off. And if they have been using `strict=error` they would already have gotten and fixed the errors.

MosesMendoza:
I ask because @ahpook has stated in a few places that puppet 5 will require no changes to puppet code that worked under puppet 4:
I.e., 
https://groups.google.com/forum/#!topic/puppet-dev/-H1pHJM6NLE
> Q: Puppet 5, what the hell eric0?! I just spent a month updating my code to run under Puppet 4. 
A: No Puppet code that works under Puppet 4 needs changing[1] to work under 5. This is a semver major to release some backwards-incompatible changes that have stacked up, plus some additional feature work, but does not affect the language. 

Maybe we need to circle up on that?

thallgren:
@MosesMendoza, I think there might be reason to amend that statement from @ahpook. AFAIK, we intend to remove things in 5.0 that has been flagged as deprecated a while (with warnings in the log, and errors when strict). If that's not the case, then we have a problem since such removals have been made already on the master branch. There are also API changes on methods marked `@api private`, so if "Puppet code" includes stuff written in Ruby, that needs to be conveyed too.

ahpook:
@MosesMendoza   Note that my footnoote from the quoted mail says:

> [1] I'm reserving a tiny, tiny asterisk for some Ruby extensions that use internal APIs that may change, like pre-Puppet 4.9 lookup extensions.

Also, there's obviously some things that have been emitting deprecation warnings for a long time that are going away. What I meant was that "code that's working *cleanly* and in compliance under 4.x will not need to change under 5"

MosesMendoza:
Ah, got it @ahpook. Thanks for clarifying. I wonder if that's worth being explicit about to the lists? I interpreted your statement(s) as suggesting even deprecated language behavior would still function for puppet code in PUP 5.0. My understanding is this PR reflects changes to deprecated, but public language behavior/API (not private, covered by your asterisk). I suppose my interpretations can't be cast as a representative sample:) @thallgren @hlindberg thanks for your patience, feel free to proceed..

hlindberg:
Thanks @ahpook for the clarification.

PR #5768, opened 2017-03-30
(PUP-7412) Device, ability to pass single certname

tphoney:


tphoney:
@branan can you have a look over this, or recommend someone ?

spec/unit/application/device_spec.rb
```diff
@@ -265,12 +271,12 @@
     end
   end
 
-
   describe "when running" do
     before :each do
       @device.options.stubs(:[]).with(:fingerprint).returns(false)
       Puppet.stubs(:notice)
       @device.options.stubs(:[]).with(:detailed_exitcodes).returns(false)
+      @device.options.stubs(:[]).with(:node).returns(false)
```
branan: Above you ensure that it is init'd to `nil`, but here you stub with `false`. Should this be `nil` instead?

branan:
@tphoney Could you update the commit messages here to follow our [contributing guidelines](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md)? In addition, some more details in the associated ticket about why these changes are useful would be rad

lib/puppet/application/device.rb
```diff
@@ -169,8 +174,15 @@ def main
     # find device list
     require 'puppet/util/network_device/config'
     devices = Puppet::Util::NetworkDevice::Config.devices
+    if options[:node]
+      devices.select! { |key, value| key == options[:node] }
```
branan: This feels kind of nasty to me, as it's a fairly major modification of global state. I assume access to the `Puppet::Util::NetworkDevice::Config.deivces` hash is hard-coded all over the place, making a cleaner solution a major pain?

tphoney: so how this works is cruddy, assigning the global list of devices to a local variable is not great 🤢.  But there is only one calling function to self.devices and it is this code.  
So  we can leave it as is. 
OR 
I can rewrite this so the code lives inside lib/puppet/util/network_device/config.rb

spec/unit/application/device_spec.rb
tphoney: resolved this. WIll push once

tphoney:
@branan i am still unsure how to approach the remaining question you posed. RE:
This feels kind of nasty to me, as it's a fairly major modification of global state. I assume access to the Puppet::Util::NetworkDevice::Config.deivces hash is hard-coded all over the place, making a cleaner solution a major pain?

shermdog:
:+1:

lib/puppet/application/device.rb
MosesMendoza: I'm inclined to agree with @branan. 

A simpler solution than rewriting might just be to create a copy of `Puppet::Util::NetworkDevice::Config.devices` and then update later operations to operate on the copy variable?

I.e.,
```ruby
devices = Puppet::Util::NetworkDevice::Config.devices.dup
```
or
```ruby
selected_devices = devices.select { |key, value| key == options[:node] }
...
```

tphoney: devices is a local copy in the the device application of global state.

lib/puppet/application/device.rb
```diff
@@ -174,18 +178,17 @@ def main
     # find device list
     require 'puppet/util/network_device/config'
     devices = Puppet::Util::NetworkDevice::Config.devices
-    if options[:node]
-      devices.select! { |key, value| key == options[:node] }
+    if options[:target]
+      devices.select! { |key, value| key == options[:target] }
```
branan: The change from `node` to `target` lost the context on this, but here we go:

`devices` here is not a copy. It's a reference to the same devices hash that is stored in `Puppet::Utili::NetworkDevice::Config`. If you want a true copy, you'll want to use `.dup` as @MosesMendoza suggested, or just get rid of the `!` on the select.

My first comment presumed that you were intentionally trying to propagate the change through into the global state to force other parts of Puppet to use the filtered list. This was why I requested cleanups to avoid using global state in that way.

If we're just trying to filter locally, I'd go with a `dup` or a non-mutating `select`

tphoney:
@branan @MosesMendoza thanks for the clarification, apologies for being such a laggard. 

I have made the appropriate change to the code and made the appropriate changes to the unit tests.

branan:
I think this probably ought to be squished the first commit isn't really complete or useful on its own, and lacks a full commit message.

Would also still be nice to see the commit message cleaned up to match our contributing guidelines. Other than that, I think this is reasonable at this point :+1:

lib/puppet/application/device.rb
```diff
@@ -170,10 +180,17 @@ def main
...
-      devices = Puppet::Util::NetworkDevice::Config.devices
+      devices = Puppet::Util::NetworkDevice::Config.devices.dup
+      if options[:target]
+        devices.select! { |key, value| key == options[:target] }
+      end
       if devices.empty?
-        Puppet.err "No device found in #{Puppet[:deviceconfig]}"
-        exit(1)
+        if options[:target]
+          Puppet.err "Target device/certificate '#{options[:target]}' not found in #{Puppet[:deviceconfig]}"
```
Magisus: Please mark this string for externalization. See https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md#writing-translatable-code.

lib/puppet/application/device.rb
```diff
@@ -170,10 +180,17 @@ def main
...
+        else
+          Puppet.err "No device found in #{Puppet[:deviceconfig]}"
```
Magisus: Same with this one.

MosesMendoza:
Nice, thanks for the updates @tphoney. If you can get the strings wrapped per @Magisus's comment, i'm happy to take care of the rest re: commit squashing etc

tphoney:
Ok hopefully the final attempt with strings ready for translation. @branan @MosesMendoza

hlindberg:
The commit message needs some love - too wide and has typos.

DavidS:
FTFY.

PR #5770, opened 2017-03-31
(PUP-7402) Handle trailing slash in File resources

hlindberg:
This PR fixes bugs in the handling of resource references. Specifically, a resource such as File where it drops a trailing slash using a titlte_patterns that maps the title to the namevar path. This meant that a `/tmp/foo/` and a `/tmp/foo` would be considered as two different resource by the compiler/catalog, but not when later applying the catalog.

Over time several changes have accumulated to this problem, a problem that has gone undetected since there was no validation performed for relationships formed with meta parameters (and longer back, no validation for relationship operators). When validation was added, it was clear that handling of aliases was broken during compilation. Then, in PUP-7402, the title_patterns were found to not be involved when searching. This seems to have been lost in an attempt to optimize logic for application orchestration.

### DIFF
--- a/lib/puppet/resource/catalog.rb
+++ b/lib/puppet/resource/catalog.rb
@@ -82,6 +82,9 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
     tag(*classes)
   end
 
+  # Returns [typename, title] when given a String with "Type[title]". 
+  # Returns [nil, nil] if '[' ']' not detected.
+  #
   def title_key_for_ref( ref )
     s = ref.index('[')
     e = ref.rindex(']')
@@ -366,18 +369,14 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
       # an instance has to be created in order to construct the unique key used when
       # searching for aliases, or when app_management is active and nothing is found in
       # which case it is needed by the CapabilityFinder.
-# PUP-4947
       res = nil
-#      app_mgnt = Puppet[:app_management]
-#      if app_mgnt || !@aliases.empty?
       res = Puppet::Resource.new(type, title, { :environment => @environment_instance })
 
-        # No need to build the uniqueness key unless there are aliases
-      result = @resource_table[[type_name, res.uniqueness_key].flatten] unless @aliases.empty?
-# PUP-4947
-#      end
+      # Must check with uniqueness key because of aliases or if resource transforms title in title
+      # to attribute mappings.
+      result = @resource_table[[type_name, res.uniqueness_key].flatten]
 
-      if result.nil? # && app_mgnt
+      if result.nil?
         resource_type = res.resource_type
         if resource_type && resource_type.is_capability?
           # @todo lutter 2015-03-10: this assumes that it is legal to just
--- a/lib/puppet/resource/catalog.rb
+++ b/lib/puppet/resource/catalog.rb
@@ -369,7 +369,6 @@ class Puppet::Resource::Catalog < Puppet::Graph::SimpleGraph
       # an instance has to be created in order to construct the unique key used when
       # searching for aliases, or when app_management is active and nothing is found in
       # which case it is needed by the CapabilityFinder.
-      res = nil
       res = Puppet::Resource.new(type, title, { :environment => @environment_instance })
 
       # Must check with uniqueness key because of aliases or if resource transforms title in title
--- a/lib/puppet/parser/resource.rb
+++ b/lib/puppet/parser/resource.rb
@@ -190,7 +190,7 @@ class Puppet::Parser::Resource < Puppet::Resource
   alias []= set_parameter
 
   def to_hash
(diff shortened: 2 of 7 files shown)


joshcooper:
The spec test indicates the compile should succeed, but I'm getting an error (during application). I'm not sure if that's expected:

```
pr/5770 $ cat test.pp
file {"/Users/josh/tmp/foo":
  ensure => directory,
  require => File["/Users/josh/tmp/foo/"],
}
pr/5770 $ bundle exec puppet apply test.pp --trace
Notice: Compiled catalog for XXX.puppetlabs.net in environment production in 0.02 seconds
Error: Failed to apply catalog: Found 1 dependency cycle:
(File[/Users/josh/foo] => File[/Users/josh/foo])
Try the '--graph' option and opening the resulting '.dot' file in OmniGraffle or GraphViz
/Users/josh/work/puppet/lib/puppet/graph/simple_graph.rb:242:in `report_cycles_in_graph'
/Users/josh/work/puppet/lib/puppet/graph/relationship_graph.rb:106:in `traverse'
/Users/josh/work/puppet/lib/puppet/transaction.rb:154:in `evaluate'
/Users/josh/work/puppet/lib/puppet/resource/catalog.rb:231:in `block in apply'
...
```

lib/puppet/resource/catalog.rb
```diff
@@ -366,18 +369,14 @@ def resource(type, title = nil)
...
-#      if app_mgnt || !@aliases.empty?
       res = Puppet::Resource.new(type, title, { :environment => @environment_instance })
 
-        # No need to build the uniqueness key unless there are aliases
-      result = @resource_table[[type_name, res.uniqueness_key].flatten] unless @aliases.empty?
-# PUP-4947
-#      end
+      # Must check with uniqueness key because of aliases or if resource transforms title in title
+      # to attribute mappings.
+      result = @resource_table[[type_name, res.uniqueness_key].flatten]
```
thallgren: I see that all performance optimizations that I added in the past are now removed and the fairly expensive uniqueness_key calculation is always performed. Has any performance measurements been to check what impact this has?

hlindberg: No impact on performance done. Note that that only doing the uniqueness test if there are aliases is wrong since the uniqueness key is needed for all resources that have one that is different from the title. Not even if the uniqueness key contains just a single entry can you be sure that it is not different.

hlindberg:
@joshcooper it is supposed to fail when doing apply because of the circularity. There is another ticket about checking for simple circularity (statically not 100% possible and therefore not done now at compile time). If a circularity check is done at compile time, tests like this one will need to be changed.

lib/puppet/resource/catalog.rb
joshcooper: Oh, I get it now. It seems less desirable to compile the catalog only to fail later on the agent. In the future, should we fail compile too? If so, can we put a deprecation warning when strict mode is enabled? I'm not sure if we have enough information at compile time to detect the problem given it is `File` specific behavior?

lib/puppet/resource/catalog.rb
```diff
@@ -366,18 +369,14 @@ def resource(type, title = nil)
       # an instance has to be created in order to construct the unique key used when
       # searching for aliases, or when app_management is active and nothing is found in
       # which case it is needed by the CapabilityFinder.
-# PUP-4947
       res = nil
```
thallgren: This line is redundant.

lib/puppet/resource/catalog.rb
```diff
@@ -366,18 +369,14 @@ def resource(type, title = nil)
...
-#      app_mgnt = Puppet[:app_management]
-#      if app_mgnt || !@aliases.empty?
       res = Puppet::Resource.new(type, title, { :environment => @environment_instance })
 
-        # No need to build the uniqueness key unless there are aliases
-      result = @resource_table[[type_name, res.uniqueness_key].flatten] unless @aliases.empty?
-# PUP-4947
-#      end
```
thallgren: Suggest you remove all comments regarding PUP-4947.

lib/puppet/resource/catalog.rb
thallgren: I ran some benchmark tests. I can't see any degradation in performance so I'm fine with this change.

hlindberg:
@joshcooper IMO we should fail the compilation if we know for sure that it will not apply. I think that is a separate issue though.

lib/puppet/resource/catalog.rb
hlindberg: There are no comments about PUP-4947 afaict. Think you are commenting on the original where I removed those comments ??? PUP-4947 is only mentioned in a commit message as it states it cleans up commented out code from that ticket.

lib/puppet/resource/catalog.rb
hlindberg: removed

PR #5771, opened 2017-03-31
(docs) Document behavior when file mode is omitted

nfagerlund:
When this is merged, please resolve [DOCUMENT-661](https://tickets.puppetlabs.com/browse/DOCUMENT-661).

My source for the thing about 0644 for new files is this:

``` ruby
# lib/puppet/type/file.rb
  def write(property = nil)
    # ...
    else
      umask = mode ? 000 : 022
      Puppet::Util.withumask(umask) { ::File.open(self[:path], 'wb', mode_int ) { |f| property.write(f) if property } }
    end
    # .....
  end
```

### DIFF
--- a/lib/puppet/type/file/mode.rb
+++ b/lib/puppet/type/file/mode.rb
@@ -13,6 +13,10 @@ module Puppet
       notation. This value **must** be specified as a string; do not use
       un-quoted numbers to represent file modes.
 
+      If the mode is omitted (or explicitly set to `undef`), Puppet does not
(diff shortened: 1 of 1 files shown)


MosesMendoza:
This seems accurate.

MosesMendoza:
@nfagerlund have some work internally to fixup stable before merging more into it. will revisit once that's wrapped up

nfagerlund:
👍 works for me! Thank you.

PR #5772, opened 2017-03-31
(PUP-5756) remove dependency on eventlog

MosesMendoza:
This re-opens https://github.com/puppetlabs/puppet/pull/5509, which was reverted due to CI failures when ruby returned a binary string that originated in Encoding::Windows_31J that we attempted to report (see the last comments in PUP-5756).

This PR adds two new commits on top of the previous PR, one that improves our handling of conversion of BINARY strings in Puppet::Util::CharacterEncoding, and a second to leverage this in our EventLog reporting. A different approach would be to update the base Log class, but given CharacterEncoding hasn't been out in the wild yet I'm a little wary of patching everything to use it - one logging class felt like a reasonable test given this is where we saw the original error.

    (PUP-5756) (PUP-7021) better handling of BINARY string conversion

    Prior to this commit, we would transcode to UTF-8 a string not composed of
    valid UTF-8 bytes but valid in its current encoding. This presents a challenge
    for BINARY strings, which if originating in an encoding other to UTF-8 are
    otherwise valid, but cannot be transcoded (because we have no idea where they
    came from). In order to transcode successfully, we need an originating
    encoding to work with. Rather than just blindly call encode! with binary
    strings, we can take one more guess as to their origin - which is that they
    originated in `Encoding.default_external`. I.e., if a system is running in
    Encoding::Windows_31J, and a string is returned as BINARY, we can guess that
    the string might have been Encoding::Windows_31J. This perhaps substantially
    increases our odds that we'll be able to conver the (binary) string to UTF-8
    successfully.

and

    (PUP-5756) convert eventlog entries to UTF-8

    This commit updates the EventLog destination to utilize
    Puppet::Util::CharacterEncoding to convert messages to UTF-8. This sanitization
    will be particularly helpful in the case where we get back messages from ruby
    in BINARY that we can't convert cleanly to UTF-16LE without some intermediate
    conversions.

### DIFF
--- a/lib/puppet/util/log.rb
+++ b/lib/puppet/util/log.rb
@@ -170,6 +170,17 @@ class Puppet::Util::Log
   def Log.newmessage(msg)
     return if @levels.index(msg.level) < @loglevel
 
+    if msg.message.valid_encoding?
+      message = Puppet::Util::CharacterEncoding.convert_to_utf_8(msg.message)
+    else
+      message = _("Received a log message with invalid encoding:")
+      message << Puppet::Util::CharacterEncoding.convert_to_utf_8(msg.message.dump) << "\n"
+      # We only select the last 10 callers in the stack to avoid being spammy
+      message << _("Backtrace:") << "\n" << caller[0..10].join("\n")
+    end
+
+    msg.message = message
+
     queuemessage(msg) if @destinations.length == 0
 
     @destinations.each do |name, dest|
--- a/lib/puppet/util/log/destinations.rb
+++ b/lib/puppet/util/log/destinations.rb
@@ -212,34 +212,28 @@ Puppet::Util::Log.newdesttype :array do
 end
 
 Puppet::Util::Log.newdesttype :eventlog do
+  # Leaving these in for backwards compatibility - duplicates the same in
+  # Puppet::Util::Windows::EventLog
   Puppet::Util::Log::DestEventlog::EVENTLOG_ERROR_TYPE       = 0x0001
   Puppet::Util::Log::DestEventlog::EVENTLOG_WARNING_TYPE     = 0x0002
   Puppet::Util::Log::DestEventlog::EVENTLOG_INFORMATION_TYPE = 0x0004
 
   def self.suitable?(obj)
-    Puppet.features.eventlog?
+    Puppet.features.microsoft_windows?
   end
 
   def initialize
-    @eventlog = Win32::EventLog.open("Application")
+    @eventlog = Puppet::Util::Windows::EventLog.open("Puppet")
   end
 
   def to_native(level)
-    case level
-    when :debug,:info,:notice
-      [self.class::EVENTLOG_INFORMATION_TYPE, 0x01]
-    when :warning
-      [self.class::EVENTLOG_WARNING_TYPE, 0x02]
-    when :err,:alert,:emerg,:crit
-      [self.class::EVENTLOG_ERROR_TYPE, 0x03]
-    end
+    Puppet::Util::Windows::EventLog.to_native(level)
   end
 
   def handle(msg)
     native_type, native_id = to_native(msg.level)
 
     @eventlog.report_event(
-      :source      => "Puppet",
       :event_type  => native_type,
       :event_id    => native_id,
       :data        => (msg.source and msg.source != 'Puppet' ? "#{msg.source}: " : '') + msg.to_s
--- a/lib/puppet/util/log/destinations.rb
+++ b/lib/puppet/util/log/destinations.rb
@@ -236,7 +236,7 @@ Puppet::Util::Log.newdesttype :eventlog do
     @eventlog.report_event(
       :event_type  => native_type,
       :event_id    => native_id,
-      :data        => (msg.source and msg.source != 'Puppet' ? "#{msg.source}: " : '') + msg.to_s
+      :data        => (msg.source && msg.source != 'Puppet' ? "#{msg.source}: " : '') + msg.to_s
     )
   end
 
--- a/spec/unit/util/log_spec.rb
+++ b/spec/unit/util/log_spec.rb
@@ -251,20 +251,20 @@ describe Puppet::Util::Log do
 
   describe Puppet::Util::Log::DestEventlog, :if => Puppet.features.eventlog? do
     before :each do
-      Win32::EventLog.stubs(:open).returns(stub 'mylog')
-      Win32::EventLog.stubs(:report_event)
-      Win32::EventLog.stubs(:close)
+      Puppet::Util::Windows::EventLog.stubs(:open).returns(stub 'mylog')
+      Puppet::Util::Windows::EventLog.stubs(:report_event)
+      Puppet::Util::Windows::EventLog.stubs(:close)
       Puppet.features.stubs(:eventlog?).returns(true)
     end
 
-    it "should restrict its suitability" do
-      Puppet.features.expects(:eventlog?).returns(false)
+    it "should restrict its suitability to Windows" do
+      Puppet.features.expects(:microsoft_windows?).returns(false)
 
       expect(Puppet::Util::Log::DestEventlog.suitable?('whatever')).to eq(false)
     end
 
-    it "should open the 'Application' event log" do
-      Win32::EventLog.expects(:open).with('Application')
+    it "should open the 'Puppet' event log" do
+      Puppet::Util::Windows::EventLog.expects(:open).with('Puppet')
 
       Puppet::Util::Log.newdestination(:eventlog)
     end
@@ -272,7 +272,7 @@ describe Puppet::Util::Log do
     it "should close the event log" do
       log = stub('myeventlog')
       log.expects(:close)
-      Win32::EventLog.expects(:open).returns(log)
+      Puppet::Util::Windows::EventLog.expects(:open).returns(log)
 
       Puppet::Util::Log.newdestination(:eventlog)
       Puppet::Util::Log.close(:eventlog)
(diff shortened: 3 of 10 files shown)


Iristyle:
@MosesMendoza looks like we have some test failures on Travis

MosesMendoza:
Ha interesting. I'll dig into it. It appears at least from the way travis is rendering the output that it's getting what it expects but still failing on each test.
```
       expected: "\xFE\xF1"
            got: "\xFE\xF1"
```

Iristyle:
@MosesMendoza yeah, I noticed that... I wonder if the metadata on the string has a different encoding?

MosesMendoza:
Yeah, exactly. I found it - side-effect of the latest change affects the original string's external encoding rather than leaving it alone on failure

MosesMendoza:
Updated. Added a new commit c4e32cf that address previous side-effect. Will squash prior to merge, separate for review.

Also kicked off an ad-hoc build with all arch/version/locale windows test targets represented, will post results on completion

spec/unit/etc_spec.rb
```diff
@@ -50,39 +51,85 @@
       end
     end
 
-    context "given an original Etc::Group struct with field values that cannot be converted to UTF-8" do
+    context "given an original Etc::Group struct with BINARY field values" do
+      original_encoding = Encoding.default_external
+
+      after(:each) do
+        Encoding.default_external = original_encoding
+      end
```
MosesMendoza: We end up repeating this pattern a few times, and I wonder if it's going to come up more.

Might be worth putting this into a spec util helper function, something callable like
```
with_external_encoding(Encoding::EUC_KR) do
  let(:group) { Etc::Group.new }
  ...
```

Iristyle: I like that idea

lib/puppet/util/character_encoding.rb
```diff
@@ -62,13 +63,27 @@ def convert_to_utf_8!(string)
...
           return string.encode!(Encoding::UTF_8)
         else
           # If the string is neither valid UTF-8 as-is nor valid in its current
           # encoding, fail. It requires user remediation.
           raise EncodingError
         end
       rescue EncodingError => detail
+        # Ensure the string retains it original external encoding since
+        # we've failed
+        string.force_encoding(original_encoding)
```
joshcooper: This method modifies the `string` parameter and also returns a value. Can we make it do one or the other?

MosesMendoza: @joshcooper I'm sorry I'm not connecting the dots. Are you saying this line would end up modifying the string from its original state? My hope was that it was actually doing the opposite - restoring its original state after it was modified on line 76 in anticipation of calling `encode!`. I could modify line 86 to only do `force_encoding` if the same criteria are satisfied as we demand on line 76, ie only if the string's original encoding was BINARY and default external is not UTF-8? Is that what you mean?

joshcooper: More a general comment about the `convert_to_utf8!` method. We generally modify the string object that is passed in, and then we return that same object, e.g. [`return string.encode!(...)](https://github.com/MosesMendoza/puppet/blob/c4e32cfdb25a87da21048f19d65402a272b94b4a/lib/puppet/util/character_encoding.rb#L77). But we also `return nil`](https://github.com/MosesMendoza/puppet/blob/c4e32cfdb25a87da21048f19d65402a272b94b4a/lib/puppet/util/character_encoding.rb#L92). So the API sometimes modifies the string that was passed in, and sometimes returns a nil value. As a consumer of the API, I'm not sure which value I should use after the API is called (the string that was passed in or the return value). Can the method be simplified to either always return the "resulting" object (making a copy of the original), or always modify the passed in object and not returning a value?

MosesMendoza: Ah that makes sense, thanks. Yes I agree with your concern. In its first form the util never modified the original string and always returned a copy. The performance implications of always copying the string (in some cases multiple times) were not insignificant, so we adjusted to modifying the string in place. But, as you point out, we ended up with inconsistent return semantics. I'm inclined to modify the string in place and always return nil to keep the performance improvement but improve consistency.

MosesMendoza:
re-added blocked label as this is undergoing further revision based on feedback

MosesMendoza:
@Iristyle @joshcooper https://github.com/puppetlabs/puppet/pull/5799 has been opened separately to contain changes to encoding utility for isolated review. This PR now depends on that PR, both of which are ready for review.

MosesMendoza:
No longer blocked.

This (in combination with #5799) has passed ad-hoc acceptance run, here: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/all/job/platform_puppet-agent_intn-van-sys_suite-ad-hoc-vmpooler-puppet-ad-hoc/15/

against these targets:
* redhat7-64a
* ubuntu1604-64a
* windows10ent-32a
* windows2008r2-64a
* windows2012r2-64a
* windows2012r2-6432a
* windows2012r2_ja-64a
* windows2012r2_ja-6432a
* windows2016-64a

MosesMendoza:
this will need rebased when #5799 is merged. That one has been retargeted at stable, but perhaps this should still go into master?

lib/puppet/util/log.rb
```diff
@@ -170,6 +170,17 @@ def Log.with_destination(destination, &block)
   def Log.newmessage(msg)
     return if @levels.index(msg.level) < @loglevel
 
+    if msg.message.valid_encoding?
+      message = Puppet::Util::CharacterEncoding.convert_to_utf_8(msg.message)
+    else
+      message = _("Received a log message with invalid encoding:")
+      message << Puppet::Util::CharacterEncoding.convert_to_utf_8(msg.message.dump) << "\n"
+      # We only select the last 5 callers in the stack to avoid being spammy
```
Iristyle: 5 was pretty arbitrary... I wonder if we should go a little deeper than that?

spec/unit/util/log_spec.rb
```diff
@@ -18,6 +18,32 @@ def log_notice(message)
     expect(message).to eq("foo")
   end
 
+  context "given a message with invalid encoding" do
+    let(:logs) { [] }
+    let(:invalid_message) { "\xFDfoo".force_encoding(Encoding::UTF_8) }
```
Iristyle: Maybe better to test with an invalid byte of another multi-byte encoding, like Shift-JIS?

```
[4] pry(main)> "\xFD\xFB".force_encoding(Encoding::Shift_JIS).valid_encoding?
=> false
```

lib/puppet/util/log.rb
MosesMendoza: Sure. Maybe 10 lines? I really wouldn't want to go deeper than 10.. Also just realized definitely need to make sure we don't access out of bounds of the array. Classic.

spec/unit/util/log_spec.rb
MosesMendoza: sure

lib/puppet/util/log.rb
```diff
@@ -170,6 +170,17 @@ def Log.with_destination(destination, &block)
   def Log.newmessage(msg)
     return if @levels.index(msg.level) < @loglevel
 
+    if msg.message.valid_encoding?
```
Iristyle: In this case should we also bump up the `level`?

Our levels are:
* :debug
* :notice
* :info
* :warning
* :err

lib/puppet/util/log.rb
Iristyle: I was actually looking at the bounds problem initially... it's not actually a problem, but you could also use `.first` for the same thing:

```
[1] pry(main)> foo = ['hi']
=> ["hi"]
[2] pry(main)> foo[0..4]
=> ["hi"]
[3] pry(main)> foo[0..4].join("\n")
=> "hi"
```

MosesMendoza: i'm not sure whether to be grateful or dismayed

lib/puppet/util/log.rb
MosesMendoza: I was on the fence about that... Since we're still logging the original message instead of a separate one re: the encoding I was hesitant to change too much about it - mostly to avoid surprising people and/or changing expectations/log levels in a (possibly?) backwards-incompatible way. I figured just logging at initial level was least surprising.

Iristyle: I can buy that... for 💵

spec/unit/util/log_spec.rb
MosesMendoza: updated

lib/puppet/util/log/destinations.rb
```diff
@@ -212,37 +212,35 @@ def handle(msg)
...
-    end
+    Puppet::Util::Windows::EventLog.to_native(level)
   end
 
   def handle(msg)
     native_type, native_id = to_native(msg.level)
 
+    # Attempt to convert our message to UTF-8
+    # Just returns a copy of the original message on failure to convert
+    message = Puppet::Util::CharacterEncoding.convert_to_utf_8(msg.to_s)
```
MosesMendoza: We don't need this anymore since we're doing it higher up

Iristyle: 👍

lib/puppet/util/log/destinations.rb
```diff
@@ -212,6 +212,7 @@ def handle(msg)
 end
 
 Puppet::Util::Log.newdesttype :eventlog do
+  require 'puppet/util/character_encoding'
```
Iristyle: This `require` is no longer needed

MosesMendoza: truth

Iristyle:
The hunch here is that this also fixes [PUP-7063](https://tickets.puppetlabs.com/browse/PUP-7063)

Iristyle:
A follow-on ticket has been filed at https://tickets.puppetlabs.com/browse/PUP-7880 to address encoding incompatibilities between `source` and `message` attributes of the log object.

PR #5570, opened 2017-01-31
(PUP-7042) Mark strings in application

Magisus:
This commit marks user-facing error and info strings in
`lib/puppet/application.rb` and `lib/puppet/application/*` for
translation.

PR #5775, opened 2017-04-03
(PUP-7259) Support JSON and PSON

joshcooper:
The agent now supports JSON and PSON, and will prefer to receive JSON (catalog), but always send PSON (facts, reports). Later PRs will allow the agent to send facts and reports in JSON as well, as well as gracefully handle when the agent tries to send data in a format that the server doesn't accept (415).

PSON is deprecated in documentation only.

Remove `binary` as a supported format for models which don't actually support it, e.g. catalog. We use `binary` when transferring file content (both the file type and file buckets). We use `s` (aka `text/plain`) for certificate related  PEM files. And for all others we support `json, pson, yaml`, where `yaml` is excluded for REST requests. Additionally, if msgpack is enabled, then it only affects the third "generic" bucket (and will have a higher precedence than json). The complete list of models and supported formats (excluding msgpack) is:

```
$ bundle exec irb
true
irb(main):002:0> ObjectSpace.enum_for(:each_object).select do |obj| obj.respond_to?(:supported_formats) end.each do |obj| puts "OBJ: #{obj}"; puts "  #{obj.supported_formats}" end
OBJ: Puppet::Transaction::Event
  [:json, :pson, :yaml]
OBJ: Puppet::Util::Log
  [:json, :pson, :yaml]
OBJ: Puppet::Util::TagSet
  [:json, :pson, :yaml]
OBJ: Puppet::Resource::Catalog
  [:json, :pson, :yaml, :dot]
OBJ: Puppet::SSL::Host
  [:json, :pson, :yaml]
OBJ: Puppet::FileBucket::File
  [:binary]
OBJ: Puppet::SSL::Key
  [:s]
OBJ: Puppet::Node::Facts
  [:json, :pson, :yaml]
OBJ: Puppet::Relationship
  [:json, :pson, :yaml]
OBJ: Puppet::SSL::CertificateRevocationList
  [:s]
OBJ: Puppet::Resource::Type
  [:json, :pson, :yaml]
OBJ: Puppet::Status
  [:json, :pson, :yaml]
OBJ: Puppet::Resource
  [:json, :pson, :yaml]
OBJ: Puppet::Util::Metric
  [:json, :pson, :yaml]
OBJ: Puppet::Node
  [:json, :pson, :yaml]
OBJ: Puppet::SSL::Certificate
  [:s]
OBJ: Puppet::DataBinding
  [:yaml]
OBJ: Puppet::Transaction::Report
  [:json, :pson, :yaml]
OBJ: Puppet::Parser::Resource
  [:json, :pson, :yaml]
OBJ: Puppet::Resource::Status
  [:json, :pson, :yaml]
OBJ: Puppet::SSL::CertificateRequest
  [:s]
```

`DataBinding` only supports `yaml` because it doesn't implement the `from_data_hash` methods needed to support `json`, etc. It's not clear why that is.

Bump hiera minimum version to 3.2.1 to eliminate pure_json dependency

### DIFF
--- a/lib/puppet/application/apply.rb
+++ b/lib/puppet/application/apply.rb
@@ -338,7 +338,6 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
   def read_catalog(text)
     begin
       catalog = Puppet::Resource::Catalog.convert_from(Puppet::Resource::Catalog.default_format,text)
-      catalog = Puppet::Resource::Catalog.pson_create(catalog) unless catalog.is_a?(Puppet::Resource::Catalog)
     rescue => detail
       raise Puppet::Error, "Could not deserialize catalog from pson: #{detail}", detail.backtrace
     end
--- a/lib/puppet/application/apply.rb
+++ b/lib/puppet/application/apply.rb
@@ -336,10 +336,11 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
   private
 
   def read_catalog(text)
+    format = Puppet::Resource::Catalog.default_format
     begin
-      catalog = Puppet::Resource::Catalog.convert_from(Puppet::Resource::Catalog.default_format,text)
+      catalog = Puppet::Resource::Catalog.convert_from(format, text)
     rescue => detail
-      raise Puppet::Error, "Could not deserialize catalog from pson: #{detail}", detail.backtrace
+      raise Puppet::Error, "Could not deserialize catalog from #{format}: #{detail}", detail.backtrace
     end
 
     catalog.to_ral
--- a/lib/puppet/application/apply.rb
+++ b/lib/puppet/application/apply.rb
@@ -340,7 +340,7 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
     begin
       catalog = Puppet::Resource::Catalog.convert_from(format, text)
     rescue => detail
-      raise Puppet::Error, "Could not deserialize catalog from #{format}: #{detail}", detail.backtrace
+      raise Puppet::Error, _("Could not deserialize catalog from %{format}: %{detail}") % { format: format, detail: detail }, detail.backtrace
     end
 
     catalog.to_ral
--- a/lib/puppet/file_bucket/file.rb
+++ b/lib/puppet/file_bucket/file.rb
@@ -17,10 +17,6 @@ class Puppet::FileBucket::File
     [:binary]
   end
 
-  def self.default_format
-    :binary
-  end
-
   def initialize(contents, options = {})
     case contents
     when String
--- a/lib/puppet/file_bucket/file.rb
+++ b/lib/puppet/file_bucket/file.rb
@@ -74,15 +74,6 @@ class Puppet::FileBucket::File
     self.new(contents)
   end
 
-  def to_data_hash
-    # Note that this serializes the entire data to a string and places it in a hash.
-    { "contents" => contents.to_binary }
-  end
-
-  def self.from_data_hash(data)
-    self.new(data["contents"])
-  end
-
   private
 
   class StringContents
--- a/lib/puppet/indirector/json.rb
+++ b/lib/puppet/indirector/json.rb
@@ -4,8 +4,7 @@ require 'puppet/util'
 # The base class for JSON indirection terminus implementations.
 #
 # This should generally be preferred to the YAML base for any future
-# implementations, since it is ~ three times faster despite being pure Ruby
-# rather than a C implementation.
+# implementations, since it is faster and can load untrusted data safely.
 class Puppet::Indirector::JSON < Puppet::Indirector::Terminus
   def find(request)
     load_json_from_file(path(request.key), request.key)
--- a/lib/puppet/indirector/json.rb
+++ b/lib/puppet/indirector/json.rb
@@ -67,10 +67,10 @@ class Puppet::Indirector::JSON < Puppet::Indirector::Terminus
   end
 
   def from_json(text)
-    model.convert_from('pson', text)
+    model.convert_from('json', text)
   end
 
   def to_json(object)
-    object.render('pson')
+    object.render('json')
   end
 end
--- a/lib/puppet/network/format.rb
+++ b/lib/puppet/network/format.rb
@@ -70,7 +70,8 @@ class Puppet::Network::Format
   def render_multiple(instances)
     # This method implicitly assumes that all instances are of the same type.
     return instances[0].class.send(render_multiple_method, instances) if instances[0].class.respond_to?(render_multiple_method)
-    raise NotImplementedError, "#{instances[0].class} does not respond to #{render_multiple_method}; can not render multiple instances to #{mime}"
+    raise NotImplementedError, _("%{klass} does not respond to %{method}; can not render multiple instances to %{mime}") \
+                               % { klass: instances[0].class, method: render_multiple_method, mime: mime }
   end
 
   def required_methods_present?(klass)
--- a/lib/puppet/network/format.rb
+++ b/lib/puppet/network/format.rb
@@ -70,7 +70,7 @@ class Puppet::Network::Format
   def render_multiple(instances)
     # This method implicitly assumes that all instances are of the same type.
     return instances[0].class.send(render_multiple_method, instances) if instances[0].class.respond_to?(render_multiple_method)
-    raise NotImplementedError, "#{instances[0].class} does not respond to #{render_multiple_method}; can not intern multiple instances to #{mime}"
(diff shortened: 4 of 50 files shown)


joshcooper:
Acceptance against puppetserver 2.7.2 will fail, since the older server doesn't accept JSON reports. We will need to test against a version of puppetserver that includes these changes, or set `preferred_serialization_format=pson` in a presuite step for the duration of the run (until we can update puppetserver).

joshcooper:
This passed adhoc pipeline in https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent_intn-van-sys_suite-ad-hoc-vmpooler-puppet-ad-hoc/11/. Note this works because we're still sending facts and reports in PSON. Once these changes are merged and pulled into puppetserver packaging, then we can switch to JSON by default, but ensure `preferred_serialization_format=pson` will control both fact and report formats.

joshcooper:
@puppetlabs/agent-team ready for review

lib/puppet/application/apply.rb
```diff
@@ -333,11 +333,11 @@ def setup
   private
 
   def read_catalog(text)
+    format = Puppet::Resource::Catalog.default_format
     begin
-      catalog = Puppet::Resource::Catalog.convert_from(Puppet::Resource::Catalog.default_format,text)
-      catalog = Puppet::Resource::Catalog.pson_create(catalog) unless catalog.is_a?(Puppet::Resource::Catalog)
```
MosesMendoza: Out of curiosity, I tried to trace back why we were doing this, and it seem slike this line has been dead code since 0cb5e7d226934e74f8b53dc1f6425b47f536afab. Does that sound accurate?

lib/puppet/application/apply.rb
```diff
@@ -333,11 +333,11 @@ def setup
...
+      catalog = Puppet::Resource::Catalog.convert_from(format, text)
     rescue => detail
-      raise Puppet::Error, "Could not deserialize catalog from pson: #{detail}", detail.backtrace
+      raise Puppet::Error, "Could not deserialize catalog from #{format}: #{detail}", detail.backtrace
```
MosesMendoza: @Magisus has a separate PR open that wraps this string, https://github.com/puppetlabs/puppet/pull/5570 - could you update to follow that format? We're going to have conflicts either way depending on which is merged first

lib/puppet/file_bucket/file.rb
```diff
@@ -17,10 +17,6 @@ def self.supported_formats
     [:binary]
   end
 
-  def self.default_format
-    :binary
-  end
```
MosesMendoza: nice

lib/puppet/indirector/json.rb
```diff
@@ -67,10 +67,10 @@ def load_json_from_file(file, key)
   end
 
   def from_json(text)
-    model.convert_from('pson', text)
+    model.convert_from('json', text)
   end
 
   def to_json(object)
-    object.render('pson')
```
MosesMendoza: We can probably lose the comment at the top of this file about being faster as pure ruby JSON since we'll primarily be using native extension JSON now.

lib/puppet/application/apply.rb
joshcooper: Sure, I'll add a commit for any strings touched as part of this PR.

lib/puppet/network/format_support.rb
```diff
@@ -35,7 +35,9 @@ def support_format?(name)
     end
 
     def supported_formats
-      result = format_handler.formats.collect do |f|
+      result = format_handler.formats.reject do |f|
+        f == :binary
```
MosesMendoza: It feels a little odd to me to do this here. I'm still trying to trace through the implications but just to put it out there:

What if instead we removed this:
https://github.com/puppetlabs/puppet/blob/master/lib/puppet/network/formats.rb#L82-L84

And specified `:required_formats => :render_method, :intern_method` for the `:binary` format?

This is from Luke, circa 2008 3101ea2:

>  Adding a hackish raw format.
    As the comment in the file says, we don't really have enough
    data to know what a good design would look like, and I think
    this format will be a bit of a one-off, so I'm just throwing
    up some barriers to keep people from doing silly things with it.

Then it's still in the generic list of `supported_formats` but won't be available for `Puppet::Resource::Catalog` etc

lib/puppet/network/formats.rb
```diff
@@ -111,6 +113,31 @@ def data_to_instance(klass, data)
...
+  end
+
+  # JSON monkey-patches Array, so this works.
+  # https://github.com/ruby/ruby/blob/ruby_1_9_3/ext/json/generator/generator.c#L1416
+  def render_multiple(instances)
+    instances.to_json
+  end
+
+  # Unlike PSON, we do not need to unwrap the data envelope, because legacy 3.x agents
+  # have never supported JSON
```
MosesMendoza: If anyone else is curious what this pson envelope looks like, here's an example: https://github.com/puppetlabs/puppet/blob/3.8.7/lib/puppet/resource/catalog.rb#L388-L396

lib/puppet/network/format_support.rb
joshcooper: ah yeah, excellent idea

lib/puppet/util/log.rb
```diff
@@ -340,10 +340,6 @@ def to_structured_hash
     hash
   end
 
-  def to_pson(*args)
```
MosesMendoza: Are you concerned at all about other places using this? Seems like it might be a stretch but I can possibly imagine someone creating their own log destination that does to_pson. It's unlikely, but this class is also ubiquitous serving as the base for all log events.

spec/unit/file_serving/metadata_spec.rb
```diff
@@ -18,8 +18,13 @@
...
 
-  it "should support pson serialization" do
-    expect(Puppet::FileServing::Metadata.new(foobar)).to respond_to(:to_pson)
+  it "should default to json" do
+    expect(Puppet::FileServing::Metadata.default_format).to eq(:json)
+  end
+
+  it "should support json, pson, yaml" do
+    # msgpack is optional, so using include instead of eq
+    expect(Puppet::FileServing::Metadata.supported_formats).to include(:json, :pson, :yaml)
```
MosesMendoza: Just to make sure I understand - this support is the result of leveraging the indirector and defining the right `render` and `intern` methods? ie not necessarily explicitly specifying these in the class.

lib/puppet/util/log.rb
joshcooper: Since Puppet::Util::Log includes Puppet::Network::FormatSupport, and the latter implements `to_pson` (in the same way), then I think this is a noop.

spec/unit/file_serving/metadata_spec.rb
joshcooper: Yeah, it's specifically the `FormatSupport` module that all indirected types (Catalog, Facts, etc) need to include to support serialization consistently.

lib/puppet/indirector/json.rb
joshcooper: updated

lib/puppet/network/format_support.rb
joshcooper: updated.

lib/puppet/application/apply.rb
```diff
@@ -337,7 +337,7 @@ def read_catalog(text)
     begin
       catalog = Puppet::Resource::Catalog.convert_from(format, text)
     rescue => detail
-      raise Puppet::Error, "Could not deserialize catalog from #{format}: #{detail}", detail.backtrace
+      raise Puppet::Error, _("Could not deserialize catalog from #{format}: #{detail}") % { format: format, detail: detail }, detail.backtrace
```
Magisus: The interpolations in the string need to be prefixed with `%`, not `#`.

lib/puppet/network/format.rb
```diff
@@ -70,7 +70,8 @@ def render(instance)
   def render_multiple(instances)
     # This method implicitly assumes that all instances are of the same type.
     return instances[0].class.send(render_multiple_method, instances) if instances[0].class.respond_to?(render_multiple_method)
-    raise NotImplementedError, "#{instances[0].class} does not respond to #{render_multiple_method}; can not render multiple instances to #{mime}"
+    raise NotImplementedError, _("#{instances[0].class} does not respond to #{render_multiple_method}; can not render multiple instances to #{mime}") \
```
Magisus: The interpolations in the string need to be prefixed with `%`, not `#`. Also, the stuff inside the curly braces needs to correspond to the keys used in the following format hash.

joshcooper: doh, thanks!

joshcooper:
Rebased to resolved conflicting Gemfile changes.

PR #5776, opened 2017-04-03
(PUP-7410) Fix tests that rely on buggy VersionRange behavior

thallgren:
This commit adjusts some tests that relied on buggy behavior in the
`SemanticPuppet::VersionRange` class.

With this fix, the Puppet adjusted tests will function with both the
current buggy vendored semver_puppet gem, and the version that has been
fixed by PUP_6242/PUP-6373.


thallgren:
Marking this as blocked since it should wait until stable is open after 4.10.0.

PR #5777, opened 2017-04-03
(PUP-7345) Remove deprecated SemVer implementation

thallgren:
This commit removes the deprecated SemVer implementation that resides
in lib/semver.rb along with all tests and references.


lib/puppet/provider/service/upstart.rb
```diff
@@ -1,5 +1,3 @@
-require 'semver'
```
joshcooper: Do we know if any other modules on the forge `require 'semver'`? Ideally code written for 4.x will continue to work in 5.0. I wonder if providing an empty `semver.rb` would be useful (as in this case where it was required but never used)?

thallgren: I did consider to keep a tiny class that just delegated everything to `SemanticPuppet` but decided against it. It is deprecated now and 5.0 is the new major release where its removal is scheduled to take place. I don't think an empty file would be any better. After all, if some module requires the presence of the `semver.rb` file, isn't it very likely that it will break anyway? Why require it if nothing in it is referenced?

joshcooper: I was wondering if anyone else copy-pasted the `upstart` provider which requires `semver` but never uses it (or more likely used it at some point but that code's been removed). Do any other modules on the forge do that?

thallgren: Not sure how to find the answer to that question. Is there any way to efficiently search the code of all forge modules? I've been wanting to do such searches many times before.

hlindberg: We don't have to be that conservative when we have had a deprecation for a Ruby API. Still nice to reach out to module authors if we can find the modules that need an updated.

PR #5778, opened 2017-04-04
(PUP-7423) Make Pcore loader available for relevant CLI commands

thallgren:
This PR ensures that a `Puppet::Pops::Loaders` instance is created and adapted to the current environment when executing the `apply`, `device` or `resource` command. This enables subsequent ruby code to access the instance and load Pcore types dynamically.


hlindberg:
For those looking at this PR; this is an enabler for features that use pcore/puppet data types on the agent. This PR does not solve the larger feature of also making sure that data types are available and can be loaded using the "agent" command - it assumes that data types are laid out on disk the same way as for a compilation.

thallgren:
I've tested the changes with the acceptance tests under 'agent', 'apply', and 'resource'. They all run without errors. I found no acceptance tests for 'device'.

PR #5779, opened 2017-04-04
(PUP-7425) Refactor Gemfile / .gemspec for Bundler workflow

Iristyle:
- Without platform-specific gem dependencies marked as such inside the
   .gemspec, this can cause issues with a bundler based workflow that
   uses `bundle install --system --binstubs c:\ruby\bin` on Windows
   machines *even if* the given Ruby version already satisfies all of
   the Puppet platform-specific dependencies. This is due to how a
   Gem is installed by Bundler when it refers to a git repo and its
   .gemspec is rewritten by Bundler as its installed.

 - Note that all runtime dependencies are now present in the .gemspec
   and that any development dependencies are still defined in the
   Gemfile itself.

   The .gemspec is included in a `bundle install` by calling the
   `gemspec` method inside the Gemfile.  This is the intended inclusion
   mechanism for Bundler.

   The Gemfile continues to allow a few gem overrides with the
   environment variables:

   FACTER_LOCATION
   HIERA_LOCATION

   These overrides are only activated when the relevant environment
   variables are defined.

### DIFF
--- a/.gemspec
+++ b/.gemspec
@@ -38,5 +38,41 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency(%q<hiera>, [">= 2.0", "< 4"])
   # PUP-7115 - return to a gem dependency in Puppet 5
   # s.add_runtime_dependency(%q<semantic_puppet>, ['>= 0.1.3', '< 2'])
+  # i18n support (gettext-setup and dependencies)
   s.add_runtime_dependency(%q<gettext-setup>, [">= 0.10", "< 1"])
+  s.add_runtime_dependency(%q<locale>, "~> 2.1")
+  # Hiera has an unbound dependency on json_pure
+  # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
+  s.add_runtime_dependency(%q<json_pure>, "~> 1.8")
+  # net-ssh is a runtime dependency of Puppet::Util::NetworkDevice::Transport::Ssh
+  s.add_runtime_dependency(%q<net-ssh>, "~> 2.1")
+
+  # loads platform specific gems like ffi, win32 platform gems
+  # as additional runtime dependencies
+  require 'yaml'
+  data = YAML.load_file(File.join(File.dirname(__FILE__), 'ext', 'project_data.yaml'))
+  bundle_platforms = data['bundle_platforms']
+  x64_platform = Gem::Platform.local.cpu == 'x64'
+  data['gem_platform_dependencies'].each_pair do |gem_platform, info|
+    next if gem_platform == 'x86-mingw32' && x64_platform
+    next if gem_platform == 'x64-mingw32' && !x64_platform
+    if bundle_deps = info['gem_runtime_dependencies']
+      bundle_platform = bundle_platforms[gem_platform] or raise "Missing bundle_platform"
+      if bundle_platform == "all"
+        bundle_deps.each_pair do |name, version|
+          s.add_runtime_dependency(name, version)
+        end
+      else
+        # important to use .to_s and not .os for the sake of Windows
+        # .cpu  => x64
+        # .os   => mingw32
+        # .to_s => x64-mingw32
+        if Gem::Platform.local.to_s == gem_platform
+          bundle_deps.each_pair do |name, version|
+            s.add_runtime_dependency(name, version)
+          end
+        end
+      end
+    end
+  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
+gemspec
+
 def location_for(place, fake_version = nil)
   if place =~ /^(git[:@][^#]*)#(.*)/
     [fake_version, { :git => $1, :branch => $2, :require => false }].compact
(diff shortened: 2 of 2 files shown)


.gemspec
```diff
@@ -38,5 +38,43 @@ Gem::Specification.new do |s|
...
+    next if gem_platform == 'x64-mingw32' && !x64_platform
+    if bundle_deps = info['gem_runtime_dependencies']
+      bundle_platform = bundle_platforms[gem_platform] or raise "Missing bundle_platform"
+      if bundle_platform == "all"
+        bundle_deps.each_pair do |name, version|
+          s.add_runtime_dependency(name, version)
+          # gem(name, version, :require => false)
+        end
+      else
+        puts s.class
```
Iristyle: Temporarily left in while I continue to test this... so far, things are looking good.

Iristyle: Got this sorted out - needed to use `Gem::Platform.local.to_s` and the value of `gem_platform` for the sake of comparisons.

Iristyle:
dbebd0d seems to be causing some weird behavior on OSX when using `FACTER_LOCATION`, and should probably be moved to its own PR / removed altogether.  I thought that the behavior I was fixing was a new problem caused by this refactor, but it turns out that it's always been a problem.

Iristyle:
One thing to potentially look into in the future - `ext/mkrf_conf.rb` can supposedly be used to include platform specific gems at runtime as long as `gemspec.extensions = 'ext/mkrf_conf.rb'` is added to the gemspec.  This could make it so that Puppet does not need to have platform specific gems (and could remove the need for a separate gem building rake task that uses project_data.yaml).  Probably worth another ticket / investigation for a later date.

.gemspec
```diff
@@ -38,5 +38,39 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency(%q<hiera>, [">= 2.0", "< 4"])
   # PUP-7115 - return to a gem dependency in Puppet 5
   # s.add_runtime_dependency(%q<semantic_puppet>, ['>= 0.1.3', '< 2'])
+  # i18n support (gettext-setup and dependencies)
   s.add_runtime_dependency(%q<gettext-setup>, [">= 0.10", "< 1"])
+  s.add_runtime_dependency(%q<locale>, ["~> 2.1"])
+  s.add_runtime_dependency(%q<rake>, ["10.1.1"])
```
Iristyle: @McdonaldSeanp I noticed that our shipped gem at https://rubygems.org/gems/puppet doesn't actually reference rake, though our `Gemfile` seems to indicate it's a dep.

I *think* that it's actually been expressed wrong in the `Gemfile` and should probably be labeled as a development dependency here instead.  It doesn't really matter to what we ship as those gems are generated by build tooling / automation and this file is ignored, but might be worth correcting here.

McdonaldSeanp: I think you are correct, rake is usually used as a dev dep, so this could be changed.

Iristyle:
Interestingly I moved `rake` to a dev dependency, but now this has caused a failure in AppVeyor:

```
bundle exec rake parallel:spec[2]
C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/bundler-1.14.5/lib/bundler/rubygems_integration.rb:408:in `block in replace_bin_path': can't find executable rake (Gem::Exception)
	from C:/Ruby23-x64/lib/ruby/gems/2.3.0/gems/bundler-1.14.5/lib/bundler/rubygems_integration.rb:434:in `block in replace_bin_path'
	from C:/Ruby23-x64/bin/rake:22:in `<main>'
Command exited with code 1
```

The AppVeyor invocation uses the following Bundler invocation which doesn't jive:

```
bundle install --jobs 4 --retry 2 --without development extra
```

Interestingly TravisCI appears to do the same via `bundler_args: --without development` at --https://github.com/puppetlabs/puppet/blob/master/.travis.yml#L6 ... but manages to install `rake`!

```
Installing rake 10.1.1
```


Hmmmmmm

.gemspec
Iristyle: I've addressed the `rake` dependency situation separately in PR #5784 as it required AppVeyor / Travis changes - we can get that merged in and then I'll rebase and we can get this PR in.  Otherwise, this PR fails on AppVeyor when `rake` is moved to a `development` dependency.

Iristyle:
See https://tickets.puppetlabs.com/browse/PUP-7433 for the next phase, which will require a bit more coordination.

Iristyle:
Note that https://github.com/puppetlabs/puppet/pull/5781 which builds on this commit has cleared Travis and AppVeyor at https://github.com/puppetlabs/puppet/pull/5781 which is validation that this refactor is good in both of those environments.

.gemspec
```diff
@@ -38,5 +38,45 @@ Gem::Specification.new do |s|
...
   s.add_runtime_dependency(%q<gettext-setup>, [">= 0.10", "< 1"])
+  s.add_runtime_dependency(%q<locale>, "~> 2.1")
+  # Hiera has an unbound dependency on json_pure
+  # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
+  s.add_runtime_dependency(%q<json_pure>, "~> 1.8")
+  # net-ssh is a runtime dependency of Puppet::Util::NetworkDevice::Transport::Ssh
+  s.add_runtime_dependency(%q<net-ssh>, "~> 2.1")
+
+  # development dependencies
+  s.add_development_dependency(%q<rake>, "10.1.1")
```
Iristyle: Strictly speaking this PR does not need the above line and could wait for PUP-7433.

er0ck: any feelings on if this belongs here or in the Gemfile?  does anyone here use a `gem --dev` workflow?

Iristyle: That scenario probably isn't used.  Consuming the Puppet gem as a dependency in a dev workflow is almost certainly more common, but needing to do Puppet dev like things (e.g. run Puppet tests against Puppet) probably isn't used either.

FWIW, I think it *is* nice to see dev deps listed on rubygems.  It's also nice to have the `.gemspec` drive this, but you could argue its cleaner to keep runtime deps in `.gemspec` and dev deps in `Gemfile`.

But again, this `.gemspec` isn't used *yet* for shipping - the shipping `.gemspec` is built on the fly - so it's not super relevant until PUP-7433 / PA-1077 are addressed anyhow.

er0ck: so... if this isn't used for shipping, and no one uses `gem install --dev`,  why is it here?

Iristyle: That circles back to my first comment ;0

er0ck: i'm confused then.  if subsequent commits are going to remove it, why have it in this one?

Iristyle: The plan wasn't necessarily to remove in a future commit - it was to try and move all the development dependencies into the `.gemspec`.

But I'll remove it from here for now and it can be addressed later in PUP-7433

.gemspec
```diff
@@ -38,5 +38,41 @@ Gem::Specification.new do |s|
...
   # PUP-7115 - return to a gem dependency in Puppet 5
   # s.add_runtime_dependency(%q<semantic_puppet>, ['>= 0.1.3', '< 2'])
+  # i18n support (gettext-setup and dependencies)
   s.add_runtime_dependency(%q<gettext-setup>, [">= 0.10", "< 1"])
+  s.add_runtime_dependency(%q<locale>, "~> 2.1")
+  # Hiera has an unbound dependency on json_pure
+  # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
+  s.add_runtime_dependency(%q<json_pure>, "~> 1.8")
+  # net-ssh is a runtime dependency of Puppet::Util::NetworkDevice::Transport::Ssh
+  s.add_runtime_dependency(%q<net-ssh>, "~> 2.1")
```
joshcooper: Wondering if we should hold off on this part until puppet 5? It seems like it might cause dependency resolution to fail in cases where `puppet device` is never used, e.g. puppetserver?

joshcooper: We discussed in meatspace, and this dependency is only active when running puppet in bundler (either puppet itself or a module that has a dependency on puppet and it's pointing at to a checkout).

PR #5575, opened 2017-01-31
(PUP-7042) Mark strings in file_serving

Magisus:
This commit marks user-facing error and info strings in
`lib/puppet/file_server/*` for translation.

PR #5782, opened 2017-04-05
(PUP-7256) Use JSON to render console format

joshcooper:
* Add missing tests for the console format
* Use JSON instead of PSON to pretty print
* Use JSON instead of PSON when rendering console output for face-based applications

### DIFF
--- a/lib/puppet/application/master.rb
+++ b/lib/puppet/application/master.rb
@@ -172,7 +172,7 @@ Copyright (c) 2012 Puppet Labs, LLC Licensed under the Apache 2.0 License
         raise _("Could not compile catalog for %{node}") % { node: options[:node] }
       end
 
-      puts PSON::pretty_generate(catalog.to_resource, :allow_nan => true, :max_nesting => false)
+      puts JSON::pretty_generate(catalog.to_resource, :allow_nan => true, :max_nesting => false)
     rescue => detail
       Puppet.log_exception(detail, _("Failed to compile catalog for node %{node}: %{detail}") % { node: options[:node], detail: detail })
       exit(30)
--- a/lib/puppet/module_tool/applications/builder.rb
+++ b/lib/puppet/module_tool/applications/builder.rb
@@ -135,9 +135,8 @@ module Puppet::ModuleTool
           f.write(metadata.to_json)
         end
 
-        # PSON.pretty_generate is a BINARY string
         Puppet::FileSystem.open(File.join(build_path, 'checksums.json'), nil, 'wb') do |f|
-          f.write(PSON.pretty_generate(Checksums.new(build_path)))
+          f.write(JSON.pretty_generate(Checksums.new(build_path)))
         end
       end
 
--- a/spec/unit/network/formats_spec.rb
+++ b/spec/unit/network/formats_spec.rb
@@ -378,42 +378,64 @@ describe "Puppet Network Format" do
       end
     end
 
-    ["hello", 1, 1.0].each do |input|
-      it "should just return a #{input.inspect}" do
-        expect(console.render(input)).to eq(input)
+    context "when rendering ruby types" do
+      ["hello", 1, 1.0].each do |input|
+        it "should just return a #{input.inspect}" do
+          expect(console.render(input)).to eq(input)
+        end
+      end
+
+      { true  => "true",
+        false => "false",
+        nil   => "null",
+      }.each_pair do |input, output|
+        it "renders #{input.class} as '#{output}'" do
+          expect(console.render(input)).to eq(output)
+        end
+      end
+
+      it "renders an Object as its quoted inspect value" do
+        obj = Object.new
+        expect(console.render(obj)).to eq("\"#{obj.inspect}\"")
       end
     end
 
-    [true, false, nil, Object.new].each do |input|
-      it "renders #{input.class} using PSON" do
-        expect(console.render(input)).to eq(input.to_pson)
+    context "when rendering arrays" do
+      {
+        []                => "",
+        [1, 2]            => "1\n2\n",
+        ["one"]           => "one\n",
+        [{1 => 1}]        => "{1=>1}\n",
+        [[1, 2], [3, 4]]  => "[1, 2]\n[3, 4]\n"
+      }.each_pair do |input, output|
+        it "should render #{input.inspect} as one item per line" do
+          expect(console.render(input)).to eq(output)
+        end
       end
     end
 
-    [[1, 2], ["one"], [{ 1 => 1 }]].each do |input|
-      it "should render #{input.inspect} as one item per line" do
-        expect(console.render(input)).to eq(input.collect { |item| item.to_s + "\n" }.join(''))
+    context "when rendering hashes" do
+      {
+        {}                                   => "",
+        {1 => 2}                             => "1  2\n",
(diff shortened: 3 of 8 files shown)


MosesMendoza:
@joshcooper do you want to rebase this now that #5575 is merged?

joshcooper:
Rebased, also ran `puppet facts find` against stable and master to ensure the output format isn't broken, and I get the following expected differences:

```
diff facts-old.txt facts-new.txt
4c4
<     "puppetversion": "4.10.0",
---
>     "puppetversion": "5.0.0",
89c89
<     "sp_uptime": "up 13:20:55:49",
---
>     "sp_uptime": "up 13:20:54:36",
98c98
<     "memoryfree": "297.57 MB",
---
>     "memoryfree": "615.62 MB",
100c100
<     "swapfree": "635.00 MB",
---
>     "swapfree": "571.00 MB",
102c102
<     "swapfree_mb": "635.00",
---
>     "swapfree_mb": "571.00",
104c104
<     "memoryfree_mb": "297.57",
---
>     "memoryfree_mb": "615.62",
127c127
<       "seconds": 1198521,
---
>       "seconds": 1198448,
136c136
<     "uptime_seconds": 1198521,
---
>     "uptime_seconds": 1198448,
145c145
<     "clientversion": "4.10.0",
---
>     "clientversion": "5.0.0",
148,149c148,149
<   "timestamp": "2017-04-12T13:50:56.907555000-07:00",
<   "expiration": "2017-04-12T14:20:56.907882000-07:00"
---
>   "timestamp": "2017-04-12T13:49:43.920123000-07:00",
>   "expiration": "2017-04-12T14:19:43.920394000-07:00"
```

lib/puppet/module_tool/applications/builder.rb
```diff
@@ -135,9 +135,9 @@ def write_json
           f.write(metadata.to_json)
         end
 
-        # PSON.pretty_generate is a BINARY string
+        # JSON.pretty_generate is a BINARY string
```
joshcooper: I don't think this comment is true anymore.

lib/puppet/application/master.rb
```diff
@@ -172,7 +172,7 @@ def compile
         raise "Could not compile catalog for #{options[:node]}"
       end
 
-      puts PSON::pretty_generate(catalog.to_resource, :allow_nan => true, :max_nesting => false)
+      puts JSON::pretty_generate(catalog.to_resource, :allow_nan => true, :max_nesting => false)
```
joshcooper: I don't think this will work, JSON::pretty_generate only accepts arrays or hashes

lib/puppet/module_tool/applications/builder.rb
```diff
@@ -135,9 +135,9 @@ def write_json
...
         Puppet::FileSystem.open(File.join(build_path, 'checksums.json'), nil, 'wb') do |f|
-          f.write(PSON.pretty_generate(Checksums.new(build_path)))
+          f.write(JSON.pretty_generate(Checksums.new(build_path)))
```
joshcooper: This won't work, JSON.pretty_generate requires hash or array

lib/puppet/application/master.rb
MosesMendoza: I think it works because the catalog object has a `to_json` method on it that it gets from `Puppet::Network::FormatSupport`

joshcooper:
The presence of pure_json is allowing some things to work when they shouldn't. I want to get https://github.com/puppetlabs/puppet/pull/5800 merged first, and then I'll rebase this PR on that one (which should force me to fix the issues mentioned above).

lib/puppet/module_tool/applications/builder.rb
MosesMendoza: I think the same comment applies - the Checksums class mixes in `FormatSupport` and has a `:to_data_hash` method so the `to_json` method on `FormatSupport` actually lets this work:

```ruby
12] pry(main)> checksums = Puppet::ModuleTool::Checksums.new('/Users/moses/foo/bar')
=> #<Puppet::ModuleTool::Checksums:0x007facd0b90170 @path=#<Pathname:/Users/moses/foo/bar>>
[13] pry(main)> JSON::Parser
=> JSON::Ext::Parser
[14] pry(main)> JSON.pretty_generate(checksums)
=> "{\n  \"baz\": \"68b329da9893e34099c7d8ad5cb9c940\"\n}"
[15] pry(main)> checksums.to_json
=> "{\"baz\":\"68b329da9893e34099c7d8ad5cb9c940\"}"
```

Without `to_json` defined on FormatSupport, we get this in either case:
```ruby
[5] pry(main)> JSON.pretty_generate(checksums)
JSON::GeneratorError: only generation of JSON objects or arrays allowed
from /Users/moses/.rbenv/versions/2.3.1/lib/ruby/2.3.0/json/common.rb:286:in `generate'
```

(json_pure has been removed)

lib/puppet/module_tool/applications/builder.rb
joshcooper: PSON.pretty_generate can be ASCII_8BIT in ruby 1.9. In ruby 2.3, it's UTF-8. Also JSON.pretty_generate always returns UTF-8.

lib/puppet/application/master.rb
joshcooper: For future us, JSON calls [`to_json`](https://github.com/flori/json/blob/v1.8.3/ext/json/ext/generator/generator.c#L867-L870) on objects that are not JSON values (string, numeric, etc).

spec/unit/network/formats_spec.rb
```diff
@@ -378,65 +378,126 @@ def self.from_binary(data)
...
 
-    it "should render a hash nicely with a multi-line value" do
-      pending "Moving to PSON rather than PP makes this unsupportable."
-      hash = {
-        "number" => { "1" => '1' * 40, "2" => '2' * 40, '3' => '3' * 40 },
-        "text"   => { "a" => 'a' * 40, 'b' => 'b' * 40, 'c' => 'c' * 40 }
-      }
-      expect(console.render(hash)).to eq <<EOT
+      it "should render a hash nicely with a multi-line value" do
+        pending "Moving to PSON rather than PP makes this unsupportable."
```
MosesMendoza: Should we just remove this test?

joshcooper: Ah right, I forgot about that. Apparently puppet at one time would use `pp` to emit output like:

```
(byebug) pp datum
{"number"=>
  {"1"=>"1111111111111111111111111111111111111111",
   "2"=>"2222222222222222222222222222222222222222",
   "3"=>"3333333333333333333333333333333333333333"},
 "text"=>
  {"a"=>"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
   "b"=>"bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
   "c"=>"cccccccccccccccccccccccccccccccccccccccc"}}
```

I was thinking of leaving the pending test, and coming back to this issue if I had time. I could also delete the test, since puppet has worked this way since 3.0 at least.

MosesMendoza: Ah thanks. Given that context I lean towards removing the test just to clean up cruft

PR #5784, opened 2017-04-06
(maint) Update TravisCI / AppVeyor config

Iristyle:
This must be merged prior to #5779

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: ruby
 sudo: false
-before_install:
-  - gem install bundler -v 1.10
 bundler_args: --without development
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: ruby
 sudo: false
 before_install:
```
Magisus: Is this a required field? If not, can it be deleted since it's empty?

MikaelSmith: Definitely not required.

Iristyle: I left it in for documentation, but can remove it as well.

PR #5785, opened 2017-04-06
(PUP-7427) Store rich data as normal parameter of type Hash

thallgren:
This PR changes the way resource parameters with rich data values
are stored in the catalog. Instead of using a special 'ext_parameters'
entry, the value is stored as a two element hash containing a special
key in the normal 'parameters' entry. The special key is
'\_\_pcore_type\_\_' and its value is the name of the data type of the
value that is associated with the second key 'value' in the same hash.

Any client that reads this catalog and is aware of the special rich
data key, will deserialize the value correctly. Clients that are
unaware of the rich data will just see a hash.

The value will be the string representation of the original value
unless the original is an Array, Hash, or a Pcore Object, in which
case Pcore serialization will serialize the value into an array.

### DIFF
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -30,22 +30,19 @@ class Puppet::Resource
   TYPE_NODE  = 'Node'.freeze
   TYPE_SITE  = 'Site'.freeze
 
-  def self.from_data_hash(data, json_deserializer = nil)
+  PCORE_TYPE_KEY = '__pcore_type__'.freeze
+  VALUE_KEY = 'value'.freeze
+
+  def self.from_data_hash(data, rich_data_enabled = false)
     raise ArgumentError, "No resource type provided in serialized data" unless type = data['type']
     raise ArgumentError, "No resource title provided in serialized data" unless title = data['title']
 
     resource = new(type, title)
 
     if params = data['parameters']
-      params.each { |param, value| resource[param] = value }
-    end
-
-    if ext_params = data['ext_parameters']
-      raise Puppet::Error, 'Unable to deserialize non-Data type parameters unless a deserializer is provided' unless json_deserializer
-      reader = json_deserializer.reader
-      ext_params.each do |param, value|
-        reader.re_initialize(value)
-        resource[param] = json_deserializer.read
+      params.each do |param, value|
+        value = convert_rich_data_hash(value, rich_data_enabled) if value.is_a?(Hash) && value.include?(PCORE_TYPE_KEY)
+        resource[param] = value
       end
     end
 
@@ -66,11 +63,35 @@ class Puppet::Resource
     resource
   end
 
+  def self.convert_rich_data_hash(rich_data_hash, rich_data_enabled)
+    if rich_data_enabled
+      value = rich_data_hash[VALUE_KEY]
+      if value.is_a?(Array)
+        json_deserializer ||= Puppet::Pops::Serialization::Deserializer.new(Puppet::Pops::Serialization::JSON::Reader.new([]), nil)
+        reader = json_deserializer.reader
+        reader.re_initialize(value)
+        json_deserializer.read
+      else
+        pcore_type =  Puppet::Pops::Types::TypeParser.singleton.parse(rich_data_hash[PCORE_TYPE_KEY])
+        pcore_type.create(value)
+      end
+    else
+      strict = Puppet[:strict]
+      unless strict == :off
+        msg = "Unable to deserialize non-Data value for parameter #{resource.name}.#{param} unless rich data is enabled"
+        raise Puppet::Error, msg if strict == :error
+        Puppet.warn(msg)
+      end
+      rich_data_hash
+    end
+  end
+  private_class_method :convert_rich_data_hash
+
   def inspect
     "#{@type}[#{@title}]#{to_hash.inspect}"
   end
 
-  def to_data_hash(json_serializer = nil)
+  def to_data_hash(rich_data_enabled = false)
     data = ([:type, :title, :tags] + ATTRIBUTES).inject({}) do |hash, param|
       next hash unless value = self.send(param)
       hash[param.to_s] = value
@@ -87,7 +108,7 @@ class Puppet::Resource
         value = Puppet::Resource.value_to_pson_data(value)
         if is_json_type?(value)
           params[param] = value
-        elsif json_serializer.nil?
+        elsif !rich_data_enabled
           Puppet.warning("Resource '#{to_s}' contains a #{value.class.name} value. It will be converted to the String '#{value}'")
           params[param] = value
         else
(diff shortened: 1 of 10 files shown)


lib/puppet/resource.rb
```diff
@@ -66,11 +63,35 @@ def self.from_data_hash(data, json_deserializer = nil)
...
+        reader.re_initialize(value)
+        json_deserializer.read
+      else
+        pcore_type =  Puppet::Pops::Types::TypeParser.singleton.parse(rich_data_hash[PCORE_TYPE_KEY])
+        pcore_type.create(value)
+      end
+    else
+      strict = Puppet[:strict]
+      unless strict == :off
+        msg = "Unable to deserialize non-Data value for parameter #{resource.name}.#{param} unless rich data is enabled"
```
Magisus: Can you mark this for externalization? See https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md#writing-translatable-code for details.

Magisus: Can you mark this for externalization? See https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md#writing-translatable-code for details.

There is a PR pending do mark all the others in this part of the code, so I don't want these new ones to get missed.

lib/puppet/resource.rb
```diff
@@ -96,18 +117,35 @@ def to_data_hash(json_serializer = nil)
...
-        json_serializer.write(ext_params[key])
-        writer.finish
-        ext_params[key] = writer.to_a
+      tc = Puppet::Pops::Types::TypeCalculator.singleton
+      sc = Puppet::Pops::Types::StringConverter.singleton
+      ext_params.each_pair do |key, value|
+        pcore_type = tc.infer(value)
+
+        # Don't allow unknown runtime types to leak into the catalog
+        raise Puppet::Error, "No Puppet Type found for #{value.class.name}" if pcore_type.is_a?(Puppet::Pops::Types::PRuntimeType)
```
Magisus: Same here, can you mark for externalization?

joshcooper:
Are you sure this is backwards compatible? I've fairly certain lots of providers look at the `@parameters` hash, which will now contain something that isn't a desired value from the manifest.

thallgren:
@joshcooper, how is that relevant to this change? What happens here is that rich data is stored as a JSON hash in a resource. No one is using rich data today, and hashes are valid as parameter values, so how could that break backward compatibility?

lib/puppet/resource.rb
thallgren: I sure can. The reason I didn't do it here was taht I didn't realize that this was needed on stable. There's a lot of user facing messages in there that isn't converted.

joshcooper:
@thallgren it's the part about "Clients that are unaware of the rich data will just see a hash" that concerns me. Existing providers have been written such that they expect `Puppet::Type.parameters` to only contain values that have been defined in the type, e.g. newparam/newproperty. I'm assuming that the deserialized parameters in `Puppet::Resource` will be copied into the `Puppet::Type::<foo>` object as part of the `catalog.to_ral` call?

If so, then the rich data will be present in `Puppet::Type.parameters`. Are there providers that are not expecting essentially arbitrary key/values in that hash? For example, a provider may call `send(key)` expecting the getter to be present, which normally it would be because the type and provider are versioned together. But now we're injecting arbitrary data that is neither a param nor property.

thallgren:
@joshcooper all current resource parameter values out there are unaffected by this change. An old client will only see an uninterpreted hash in a resource if it reads a catalog produced by a newer client that has `rich_data = true` and resources with rich data values. When that happens, the old client will still be able to read the parameter. but being unaware of the special meaning of the hash, it will treat it just like any other hash. There is no arbitrary data.

hlindberg:
The key thing here is naturally that an agent / master needs to both be aware of the rich data in the catalog. Deserialization is then done the right way, and a resource created from the information in the catalog will then have the rich data in its properties/parameters. Most existing resource types would not know what to do with those and would error. In the event an agent with rich data turned off gets a catalog with rich data, it will (as Thomas points out) get a hash (with the special key). Any quality type/provider would probably error on that as it would not expect such a hash as the value of a particular parameter/property. If it does not error, it would be just as if the user had entered such an (unexpected) hash directly in their manifest.

joshcooper:
> Any quality type/provider would probably error on that as it would not expect such a hash as the value of a particular parameter/property. 

Yeah, that's my concern. If you enable rich data, then providers will receive data that they are not expecting. You could argue that's not a quality provider, but since the API is vague/underspecified, who knows. It could be a provider everyone is using and the rich data feature will be dead-on-arrival.

I retract what I said about it being backwards-incompatible. There can only be problems if you opt-into rich data, which isn't the default.

thallgren:
@joshcooper, how is what you describe different from when users assign other types of unexpected data to a provider? Say, a number where a boolean is expected, or a hash where a string is expected?

joshcooper:
> how is what you describe different from when users assign other types of unexpected data to a provider?

Because the puppet resource type has an opportunity to validate the newparam/provider. But in this PR we're injecting a specially named hash, so any construct like the following could fail in mysterious ways. This example is taken from nagios which is terrible, but it shows how `@parameters` is basically API:

```
    @parameters.each { |name,value|
      next if self.class.suppress.include?(name)
      ldapname = self.parammap(name)
      str += ldapname + ": #{value}\n"
    }
```

thallgren:
Why would that fail? The `value` would be a hash, and that hash would receive a `to_s`, which admittedly would produce a somewhat funny looking result, but that would happen if the user passed a hash as the value today too.

thallgren:
@joshcooper , also please note that we are not injecting a specially named hash. We are replacing an existing value with a hash.

hlindberg:
Ah, I think I understand Josh's concern; The __pcore_type__ key is in a hash that is a value of a parameter or a property. It is not a magic new parameter/property. (Which is what Thomas is also saying).

thallgren:
@joshcooper , are your concerns answered in this PR? Can we go ahead and merge it?

hlindberg:
I am merging this as the raised concern is about an extension of the schema when it is in fact just a magic key in a hash. The feature is also opt-in and experimental.

joshcooper:
I still have concerns as I believe the `@parameters` hash to be part of the provider API. Previously, a provider would be given data in the form:

```ruby
@parameters = { :command => '/some/sensitive/command', ... }
```

But with the feature enabled, we have (I think):

```ruby
@parameters = { :command => { '__pcore_type__' => :sensitive, value => '/some/sensitive/command' } }
```

Up until now the values in the `parameters` hash have always been the literal value (string, numeric, ...) but now it can take a different form, which seems like a breaking change.

I'm ok if that's behind a feature flag, but that doesn't provide an upgrade path (new servers talking to older agents). I would think the serialization format needs to be backwards compatible so that both sides don't need to be updated at the same time.

thallgren:
@joshcooper, you write "hash have always been the literal value (string, numeric, ...)". AFAIK, there's nothing stopping the user from putting literal hashes in there today. E.g.:
```
@parameters = { :adress => { 'host' => 'example.com', 'port' => 1234 } }
```
is legal and does not break anything. That's why I still don't understand why this would become a breaking change. What will break? Even if you enable rich data, there will be no such values unless they are explicitly and deliberately added somewhere. Now, if a user adds rich data to types that cannot handle it, then there will be problems of course, but that's no different from using unexpected types today.

PR #5788, opened 2017-04-07
(PUP-7402) Update test now that trailing slash alias works

joshcooper:
Previously, the test added two resources to the catalog where one had a
trailing slash. The test relied on `defined` returning false, as it did
before commit ae066f9.

As of ae066f9, puppet will now detect that the resource with the
trailing slash has the same name as the one without. So this test
started failing (since we never tried to add a duplicate resource).

This commit updates the test to unconditionally add the duplicate. It
also removes the app_management cases since that property is deprecated
and no longer controls any behavior.

    Error: Evaluation Error: Error while evaluating a Resource
    Statement, Cannot alias File[/tmp/create-dir.foyI4D/] ...

[skip ci]

### DIFF
--- a/acceptance/tests/resource/file/should_create_directory.rb
+++ b/acceptance/tests/resource/file/should_create_directory.rb
@@ -23,27 +23,15 @@ agents.each do |agent|
     file {$dir:
       ensure => directory,
     }
-    if !defined(File["${same_dir}"]) {
-      file { $same_dir:
-        ensure => directory,
-      }
+    file { $same_dir:
+      ensure => directory,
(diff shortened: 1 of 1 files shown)


joshcooper:
This should fix the CI failure mentioned earlier today: @puppetlabs/agent-team

McdonaldSeanp:
👍

Magisus:
Seems reasonable 👍

johnduarte:
:+1: 
Spot validated on Window2012r2 and Ubuntu1404-64

PR #5789, opened 2017-04-07
(maint) Fix Puppet Ruby version check comparison

Iristyle:
- Previously, the minimum version comparison check was performed on a
   string. This can lead to failures - for instance:

   0.0.10 < 0.0.2 # true!

   Version numbers should not be treated as strings for the sake of
   comparisons.  Note that while we do have access to `SemVer` inside
   of Puppet, its not yet loaded here, so return to the standard
   `Gem::Version` for comparisons.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -1,6 +1,6 @@
 require 'puppet/version'
 
-if RUBY_VERSION < "1.9.3"
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("1.9.3")
   raise LoadError, "Puppet #{Puppet.version} requires ruby 1.9.3 or greater."
 end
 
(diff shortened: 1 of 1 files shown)


Magisus:
Thanks for fixing this up 👍

lib/puppet.rb
```diff
@@ -1,6 +1,6 @@
 require 'puppet/version'
 
-if RUBY_VERSION < "1.9.3"
```
joshcooper: Is it always safe to reference `Gem` across different supported ruby versions? Not sure if rubygems has to be loaded first?

Iristyle: Good question... seems to be available always in my OSX based installs.

```
11:02:25  Iristyle@Niihau   ~  rbenv global 1.9.3-p551
 11:02:30  Iristyle@Niihau   ~  irb
irb(main):001:0> Gem::Version.new('1.2.3.4')
=> #<Gem::Version "1.2.3.4">
irb(main):002:0> exit

 11:02:41  Iristyle@Niihau   ~  rbenv global 2.1.9
 11:02:48  Iristyle@Niihau   ~  irb
irb(main):001:0> Gem::Version.new('1.2.3.4')
=> #<Gem::Version "1.2.3.4">
irb(main):002:0> exit

 11:02:58  Iristyle@Niihau   ~  rbenv global 2.3.3
 11:03:04  Iristyle@Niihau   ~  irb
irb(main):001:0> Gem::Version.new('1.2.3.4')
=> #<Gem::Version "1.2.3.4">
irb(main):002:0> exit

 11:03:11  Iristyle@Niihau   ~  rbenv global 2.4.0
 11:03:16  Iristyle@Niihau   ~  irb
irb(main):001:0> Gem::Version.new('1.2.3.4')
=> #<Gem::Version "1.2.3.4">
irb(main):002:0> exit
```

PR #5792, opened 2017-04-10
(maint) Merge stable to master

joshcooper:
* upstream/stable:
  (maint) Fix Puppet version check comparison
  (maint) Mark net-ssh as Gemfile runtime dependency
  (maint) Remove tzinfo gem from Gemfile
  (maint) Move rake to Gemfile development / test
  (maint) TravisCI should not bundle install extra
  (maint) Remove TravisCI gem install of bundler

Preserve `gem update --system --no-doc` needed for ruby 2.4 (df0b1d0)

Preserve `extra` bundler_arg to be consistent with appveyor (f32b8dc)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: ruby
 sudo: false
-before_install:
-  - gem update --system --no-doc
 bundler_args: --without development extra
 script:
   - "bundle exec rake $CHECK"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: ruby
 sudo: false
 before_install:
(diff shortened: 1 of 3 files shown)


joshcooper:
This passed my local `bx rake parallel:spec`, but want to confirm conflicts were resolved correctly.

Magisus:
This looks right to me based on my understanding of the changes 👍  A passing Travis/Appveyor will also give us a good indication that it's correct.

Magisus:
The `bundle install` works for me locally under Ruby 2.0.0...

Magisus:
The `bundle install` worked for me locally under both 2.0.0 and 1.9.3...

PR #5793, opened 2017-04-10
(SERVER-1766) Use tk-auth rules when running acceptance under puppetserver

adrienthebo:
/cc @rlinehan @jpinsonault @camlow325 

If we like this approach I can apply it to other instances where we need to modify auth.conf.

### DIFF
--- a/acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
+++ b/acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
@@ -12,20 +12,13 @@ end
 node_names.uniq!
 
 if @options[:is_puppetserver]
-  step "Ensure legacy auth.conf is disabled for test" do
+  step "Prepare for custom tk-auth rules" do
+    on master, 'cp /etc/puppetlabs/puppetserver/conf.d/auth.conf /etc/puppetlabs/puppetserver/conf.d/auth.bak'
     modify_tk_config(master, options['puppetserver-config'], {'jruby-puppet' => {'use-legacy-auth-conf' => false}})
   end
 
   teardown do
     modify_tk_config(master, options['puppetserver-config'], {'jruby-puppet' => {'use-legacy-auth-conf' => true}})
-  end
-
-  step "Backup the tk auth.conf file" do
-    on master, 'cp /etc/puppetlabs/puppetserver/conf.d/auth.conf /etc/puppetlabs/puppetserver/conf.d/auth.bak'
-  end
-
-  teardown do
-    # restore the original tk auth.conf file
     on master, 'cp /etc/puppetlabs/puppetserver/conf.d/auth.bak /etc/puppetlabs/puppetserver/conf.d/auth.conf'
   end
 
--- a/acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
+++ b/acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
@@ -11,9 +11,10 @@ end
 
 node_names.uniq!
 
-authfile = "#{testdir}/auth.conf"
-authconf = node_names.map do |node_name|
-  %Q[
+step "Setup auth.conf" do
+  authfile = "#{testdir}/auth.conf"
+  authconf = node_names.map do |node_name|
+    <<-MANIFEST
 path /puppet/v3/catalog/#{node_name}
 auth yes
 allow *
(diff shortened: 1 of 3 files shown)


acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
```diff
@@ -11,9 +11,109 @@
 
 node_names.uniq!
 
-authfile = "#{testdir}/auth.conf"
-authconf = node_names.map do |node_name|
-  %Q[
+if @options[:is_puppetserver]
+  step "Ensure legacy auth.conf is disabled for test" do
```
adrienthebo: This step is being done so we can modify one test at a time, once all of the tests are updated we can remove this setting from each test and change the default being set to true in the puppetserver pre-suite.

adrienthebo:
I'm re-running the entire acceptance test suite to confirm that this doesn't break anything under puppetserver. Once that passes I'll ensure that things behave correctly when puppet itself runs acceptance tests via git.

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
```diff
@@ -11,9 +11,109 @@
...
+    modify_tk_config(master, options['puppetserver-config'], {'jruby-puppet' => {'use-legacy-auth-conf' => false}})
+  end
+
+  teardown do
+    modify_tk_config(master, options['puppetserver-config'], {'jruby-puppet' => {'use-legacy-auth-conf' => true}})
```
camlow325: Seems like this step should just set the value back to whatever it was to start with, which might have been `false`, the new default.  Another option could be to just back up the `puppetserver-config` file in its entirety and restore it as part of the teardown.  I did something like that with the [collect_default_metrics](https://github.com/puppetlabs/puppetserver/blob/97be028c37709701a47abf10ead39d2a152bd32f/acceptance/suites/tests/metrics/collect_default_metrics.rb#L13-L19) test in Puppet Server.

camlow325: I suppose this step should also be gated with the `:is_puppetserver` option?

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
```diff
@@ -11,9 +11,109 @@
...
+  end
+
+  step "Backup the tk auth.conf file" do
+    on master, 'cp /etc/puppetlabs/puppetserver/conf.d/auth.conf /etc/puppetlabs/puppetserver/conf.d/auth.bak'
```
camlow325: I think it'd be good to add a `-p` to the `cp` here.  Seems like without that that you might end up with the file only being readable by root - and not `puppet` - which could lead Puppet Server to not being able to read the file after it has been restarted?

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
```diff
@@ -11,9 +11,109 @@
...
+  end
+
+  teardown do
+    # restore the original tk auth.conf file
+    on master, 'cp /etc/puppetlabs/puppetserver/conf.d/auth.bak /etc/puppetlabs/puppetserver/conf.d/auth.conf'
```
camlow325: Can you do multiple `teardown` steps in the same test file?  I don't know how to beaker.  😵

camlow325: Maybe a `mv` would be good here to avoid leaving around a .bak file at the end of the test?

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
```diff
@@ -11,9 +11,109 @@
...
+            allow: "*"
+            sort-order: 500
+            name: "puppetlabs file"
+        },
+    HEADER
+
+    tka_node_rules = node_names.map do |node_name|
+      <<-NODE_RULES
+        {
+            # Allow nodes to retrieve their own catalog
```
camlow325: These comment lines seem a bit misleading.  I think these are allowing all users to retrieve the `.../node_name` endpoint - as opposed to just the `node_name` user to retrieve its own endpoint.  Maybe could just take these comment lines out since the rules otherwise seem self-describing.

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
```diff
@@ -11,9 +11,109 @@
...
+          sort-order: 999
+          name: "puppetlabs deny all"
+        }
+    ]
+}
+    FOOTER
+
+    tk_auth = [tka_header, tka_node_rules, tka_footer].flatten.join("\n")
+
+    apply_manifest_on(master, <<-MANIFEST, :catch_failures => true)
```
camlow325: Could also set a variable for the auth.conf file and just have one implementation of `apply_manifest_on` apply the right auth.conf file for both the if and else case.  Seems fine how you have it, too, though.

camlow325:
[Travis failed](https://travis-ci.org/puppetlabs/puppet/jobs/220725894) but doesn't seem to be related to the changes.  Maybe transient?

```
$ bundle exec rake $CHECK
Document generation not available without yard. cannot load such file -- yard
Checking commits 4b09fbc293a7e02d8541331965e9570a154e0c90..9dafd41ca370259ba79b335f5e4b229a17399b8f
rake aborted!

		(SERVER-1766) Use tk-auth for node_name_fact test

	The commit summary (i.e. the first line of the commit message) should start with one of:
		(pup-<digits>) # this is most common and should be a ticket at tickets.puppetlabs.com
		(docs)
		(maint)
		(packaging)
	This test for the commit summary is case-insensitive.
/home/travis/build/puppetlabs/puppet/Rakefile:92:in `block (2 levels) in <top (required)>'
/home/travis/build/puppetlabs/puppet/Rakefile:88:in `each_line'

/home/travis/build/puppetlabs/puppet/Rakefile:88:in `block in <top (required)>'
/home/travis/.rvm/gems/ruby-2.1.9/bin/ruby_executable_hooks:15:in `eval'
/home/travis/.rvm/gems/ruby-2.1.9/bin/ruby_executable_hooks:15:in `<main>'
Tasks: TOP => commits
(See full trace by running task with --trace)
```

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
adrienthebo: Right now the puppetserver setup forcibly sets `use-legacy-auth-conf` to true so the original value can't vary. Once we have fixes to the rest of the tests I want to drop these setup steps entirely and assume that if we're running under puppetserver then we're using the tk-auth rules.

I'm fine with backing up the entire puppetserver configuration directory though.

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
adrienthebo: This is an exact copy of what we do here: https://github.com/puppetlabs/puppetserver/blob/master/acceptance/suites/tests/authorization/x509_auth.rb#L9-L16 . I can implement the behavior you've suggested but wanted to note that this would vary from the reference that I used.

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
adrienthebo: See the above comment, I think your approach is cleaner and agree with it but want to note that we're already doing this.

With respect to multiple teardowns I've run the tests locally and it looks like multiple teardown steps are respected.

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
adrienthebo: I can fix this.

adrienthebo:
Given the challenges I've had with getting changes merged and through CI successfully I'm very strongly in favor of incrementally getting chunks of work through; trying to do everything in one fell swoop has pretty consistently left a lot of work jammed up on CI and I'd like to avoid it this time around.

With respect to the travis failures this is due to me using SERVER ticket numbers; I might just convert the commit messages to use the `(maint)` descriptor.

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
camlow325: Sounds good.  If we're going to drop these steps completely then my comments about backing up the files before modifying them probably wouldn't apply anyway.

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
camlow325: Ah, okay.  I guess if that works, it may be fine.

camlow325:
(maint) tag seems fine to me.  Or if you want to move the SERVER ticket to a PUP one so that we have a ticket we can match up, that seems fine too.

acceptance/tests/allow_arbitrary_node_name_fact_for_agent.rb
adrienthebo: There's a little bit more duplication but it's minimal, and in exchange we have very well separated setup steps depending on if we're running under webrick or puppetserver. Since webrick support isn't incredibly long for this world I think we're better off keeping these separate for easier cleanup later.

camlow325:
Are these the only tests that we need to rework for the `use_legacy_auth_conf` default setting change?  Seems like there was at least one other test that needed to be updated but I'm not sure what that was now.

adrienthebo:
I thought that there were more test failures but haven't been able to confirm. I'm re-running all of the puppet acceptance tests to find any stragglers.

camlow325:
@adrienthebo When you re-run the full suite, you might ensure that `use_legacy_auth_conf` is set to `false` in the teardown for the tests on this PR so that we don't inadvertently mask any other tests that might be affected.

camlow325:
👍 on the additional `check_woy_cache_works.rb` test changes.  @rlinehan or @jpinsonault - WDYT?

jpinsonault:
Well as far as I understand this it looks sane to me 👍

PR #5794, opened 2017-04-10
Updated Gemfile fun

Iristyle:
Like #5792 but with less gem magic


Iristyle:
#5792 will handle this - the experiment here proved 4fd35e8 was feasible

PR #5795, opened 2017-04-11
(maint) Fix intermittent AppVeyor failures / rewrite AppVeyor test_script as PowerShell

Iristyle:
- AppVeyor is generating intermittent failures when executing
   `bundle exec rake parallel:spec[2]`, despite no spec tests
   actually failing.

   This has been narrowed down to a non-zero exitcode being returned
   from bundler when yard is not present due to:

   Document generation not available without yard. cannot load such file -- yard

   It's unclear why this only intermittently fails when executed
   through cmd, as it will always fail when executed within a
   PowerShell script (like the refactoring in the next commit)

 - Since `bundler` wants access to `yard`, move it out of the
   :developemnt group (which is currently not installed in
   AppVeyor / Travis) so that it's present in both the :test and
   :development groups - allowing it to be installed in those
   environments

- Previously test_script was a combination of individual batch
   commands and PowerShell code.  Unify the code as PowerShell given
   PowerShell is already being used for Get-WinSystemLocale.

   Be more explicit about exiting the script, by setting the PowerShell
   host exit status based on $LASTEXITCODE set by the execution of
   bundler.

 - This change combined with the prior move of the `yard` gem to a
   different gem group, should prevent spurious AppVeyor failures.

 - Note that sleep test_script used after a locale change is left as a
   distinct step (but moved from PowerShell to cmd) to ensure AppVeyor
   logs correctly.

 - Added 2 additional pieces of useful information while logging the
   AppVeyor environment -

   - gem version
   - bundler version

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -38,17 +38,23 @@ before_test:
 
 test_script:
   # The sleep is needed to allow the Appveyor tracing to start correctly after a reboot (locale change)
-  - ps: Start-Sleep -Seconds 5
-  - SET
-  - chcp
-  - ps: Get-WinSystemLocale
-  - SET PATH=C:\%PLATFORM%\bin;%PATH%
-  - SET LOG_SPEC_ORDER=true
-  - bundle install --jobs 4 --retry 2 --without development extra
-  - type Gemfile.lock
-  - ruby -v
-  - bundle exec rake parallel:spec[2]
-
+  # - ps: Start-Sleep -Seconds 5
+  - timeout 6
+  - ps: |
+      chcp
+      Get-WinSystemLocale
+      $Env:PATH = "C:\${Env:PLATFORM}\bin;${Env:PATH}"
+      $Env:LOG_SPEC_ORDER = 'true'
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -38,17 +38,23 @@ before_test:
...
+      bundle install --jobs 4 --retry 2 --without development extra
+      Get-Content Gemfile.lock
+      ruby -v
+      gem --version
+      bundle --version
+      bundle exec rake parallel:spec[2]
+      $exitCode = $LASTEXITCODE
+      Write-Output "Spec run exited with $exitCode"
+      # exit $exitCode
+      $host.SetShouldExit($exitCode)
```
Iristyle: So to get an exit status to propagate correctly from PowerShell code run in AppVeyor, we have to call against the `$host` like the above.  `exit #` does not work properly within their custom host as it does with the Puppet PowerShell module.

/cc @glennsarti @MikaelSmith @jpogran

Iristyle:
Ok @adrienthebo you shouldn't see those AppVeyor failures anymore now -- but in the event that you do for some reason, please let me know!

adrienthebo:
@Iristyle you're my hero, thank you!

PR #5798, opened 2017-04-12
(docs) Explain what node_terminus setting does; clarify external_nodes setting

nfagerlund:
When this is merged, please resolve [DOC-2939](https://tickets.puppetlabs.com/browse/DOC-2939).

### DIFF
--- a/lib/puppet/defaults.rb
+++ b/lib/puppet/defaults.rb
@@ -407,7 +407,29 @@ deprecated and has been replaced by 'always_retry_plugins'."
     :node_terminus => {
       :type       => :terminus,
       :default    => "plain",
-      :desc       => "Where to find information about nodes.",
+      :desc       => <<-'EOT',
+        Which node data plugin to use when compiling node catalogs.
+
+        When Puppet compiles a catalog, it combines two primary sources of info: the main manifest,
+        and a node data plugin (often called a "node terminus," for historical reasons). Node data
+        plugins provide three things for a given node name:
+
+        1. A list of classes to add to that node's catalog (and, optionally, values for their
+           parameters).
+        2. Which Puppet environment the node should use.
+        3. A list of additional top-scope variables to set.
+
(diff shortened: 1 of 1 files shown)


nfagerlund:
(The `classifier` terminus isn't part of base puppet, I'm leaving out the LDAP terminus, etc. etc., but since indirections are such a hairy extension point and the "node object" is such a weird construct, I'm trying to hit a solid 80/20 point with this description so people can understand what this setting is for without getting confused or going down a rabbit hole.)

lib/puppet/defaults.rb
```diff
@@ -407,7 +407,29 @@ def self.default_diffargs
...
+        plugins provide three things for a given node name:
+
+        1. A list of classes to add to that node's catalog (and, optionally, values for their
+           parameters).
+        2. Which Puppet environment the node should use.
+        3. A list of additional top-scope variables to set.
+
+        The three main node data plugins are:
+
+        * `plain` --- Returns no data, so that the main manifest controls all node configuration.
```
MosesMendoza: Is `main manifest` a common term for the manifests in a manifest-driven classification workflow (I assume that's what this refers to)?  What if we went even further with something like `...so that puppet manifests such as site.pp control...` ? Or is that too specific..

nfagerlund: @MosesMendoza Yeah, "main manifest" is the standard name we use for the "entry point" manifest — usually `<ENVIRONMENT>/manifests/**/*.pp`. The page about that is https://docs.puppet.com/puppet/latest/dirs_manifest.html. 

Do you think we should link to that docs page from here for additional context? I could go either way, since anyone changing their node_terminus is hopefully familiar with the main manifest anyway.

MosesMendoza: Ah thanks @nfagerlund. With that context, I think this is great as-is.

PR #5799, opened 2017-04-12
(PUP-7498) update char encoding util per feedback

MosesMendoza:
This PR splits out adjustments to the character encoding heuristics that were previously a part of #5509 and contains changes based on feedback on that PR.

The three primary changes are:
* Binary string handling
Prior to this PR, we would transcode to UTF-8 a string not composed of valid UTF-8 bytes but valid in its current encoding. This presents a challenge for BINARY strings, which if originating in an encoding other to UTF-8 are otherwise valid, but cannot be transcoded (because we have no idea where they came from). In order to transcode successfully, we need an originating encoding to work with. Rather than just blindly call encode! with binary strings, we can take one more guess as to their origin - which is that they originated in `Encoding.default_external`. I.e., if a system is running in
Encoding::Windows_31J, and a string is returned as BINARY, we can guess that the string might have been Encoding::Windows_31J. This perhaps substantially increases our odds that we'll be able to convert the (binary) string to UTF-8 successfully.

* Consistently return nil
Prior to this PR, we were modifying the supplied string in Puppet::Util::CharacterEncoding as a side-effect of the recent change that added `force_encoding(Encoding.default_external)` on binary strings. In the case where we fail to encode! that change leaves the original string with a
different external encoding than it started. This PR addresses that by ensuring any on failure to convert we set the string back to its original external encoding. A quirk was accounted for in that we don't set external_encoding if the default external encoding is UTF-8. The reason is we've already determined that this string doesn't represent valid UTF-8 bytes so doing a force_encoding to UTF-8 doesn't get us any closer to being able to transcode it to UTF-8, and ruby won't do anything on `encode!` to string already has as its default_external set to the encoding supplied to encode!.

* Separate out conversion from overriding of string encoding
Prior to this PR, the Puppet::Util::CharacterEncoding::convert_to_utf_8! was inaccurately named in that the method would both convert (transcode) and override (set external encoding) on a string in different circumstances. After review, it was determined that this behavior was potentially dangerous and also could lead to inadvertent destruction or loss of string fidelity unintentionally. This PR updates the method to split out the two different operations, each with an improved description for when to use it. Conversion should be our primary use-case, for when we believe that the encoding of a string is an accurate representation of that string. Overriding a string's encoding should be used sparingly, effectively only when we have reason to believe that the existing encoding is inaccurate/incorrect.

  This PR adjusts the Puppet::Etc wrapper to account for the modified heuristics in Puppet::Util::CharacterEncoding. We leverage the new distinction between converting and overriding by only "overriding" the encoding to UTF-8 when the string is returned in BINARY and otherwise defaulting to converting it.  This is a shift in default assumptions - whereas before we would assume the string encoding was not valid for that environment, now we do. We retain the preceding requirement that strings that cannot be converted are left unmodified in the Etc struct.

  In addition to updating the Etc wrapper to account for the preceding change to split out two methods in CharacterEncoding, prior to this PR the encoding management in our Etc wrapper incorrectly expected that ruby Etc would return BINARY encoded strings when the values on disk are invalid in the system encoding. This is incorrect - ruby Etc quite happily returns strings with invalid encoding. The only case where it appears to return binary is when the environment's encoding (Encoding.default_external) is ASCII (7-bit), and any individual character bit representation is equal to or greater than \x80 (binary 10000000), ie the first 8-bit value. We account for this change in expectation by fixing our spec tests to expect and test for the appropriate scenarios.

Finally, we also add a small spec helper for running a block in a different external encoding.

### DIFF
--- a/lib/puppet/application/resource.rb
+++ b/lib/puppet/application/resource.rb
@@ -136,14 +136,14 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
     resources = find_or_save_resources(type, name, params)
 
     if options[:to_yaml]
-      text = resources.
-        map { |resource| resource.prune_parameters(:parameters_to_include => @extra_params).to_hierayaml }.
-        join("\n")
+      text = resources.map do |resource|
+        resource.prune_parameters(:parameters_to_include => @extra_params).to_hierayaml.force_encoding(Encoding.default_external)
+      end.join("\n")
       text.prepend("#{type.downcase}:\n")
     else
-      text = resources.
-        map { |resource| resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest }.
-        join("\n")
+      text = resources.map do |resource|
+        resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest.force_encoding(Encoding.default_external)
+      end.join("\n")
     end
 
     options[:edit] ?
--- a/lib/puppet/etc.rb
+++ b/lib/puppet/etc.rb
@@ -2,11 +2,28 @@ require 'puppet/util/character_encoding'
 # Wrapper around Ruby Etc module allowing us to manage encoding in a single
 # place.
 # This represents a subset of Ruby's Etc module, only the methods required by Puppet.
-# Etc returns strings in variable encoding depending on
-# environment. For Puppet we specifically want UTF-8 as our input from the Etc
-# module - which is our input for many resource instance 'is' values. The
-# associated 'should' value will basically always be coming from Puppet in
-# UTF-8. Etc is defined for Windows but calls to it return nil.
+
+# On Ruby 2.1.0 and later, Etc returns strings in variable encoding depending on
+# environment. The string returned will be labeled with the environment's
+# encoding (Encoding.default_external), with one exception: If the environment
+# encoding is 7-bit ASCII, and any individual character bit representation is
+# equal to or greater than 128 - \x80 - 0b10000000 - signifying the smallest
+# 8-bit big-endian value, the returned string will be in BINARY encoding instead
+# of environment encoding.
+#
+# Barring that exception, the returned string will be labeled as encoding
+# Encoding.default_external, regardless of validity or byte-width. For example,
+# ruby will label a string containing a four-byte characters such as "\u{2070E}"
+# as EUC_KR even though EUC_KR is a two-byte width encoding.
+#
+# On Ruby 2.0.0 and earlier, Etc will always return string values in BINARY,
+# ignoring encoding altogether.
+#
+# For Puppet we specifically want UTF-8 as our input from the Etc module - which
+# is our input for many resource instance 'is' values. The associated 'should'
+# value will basically always be coming from Puppet in UTF-8 - and written to
+# disk as UTF-8. Etc is defined for Windows but the majority calls to it return
+# nil and Puppet does not use it.
 #
 # We only use Etc for retrieving existing property values from the system. For
 # setting property values, providers leverage system tools (i.e., `useradd`)
@@ -18,7 +35,7 @@ module Puppet::Etc
     # On first call opens /etc/group and returns parse of first entry. Each subsquent call
     # returns new struct the next entry or nil if EOF. Call ::endgrent to close file.
     def getgrent
-      convert_field_values_to_utf8!(::Etc.getgrent)
+      override_field_values_to_utf8!(::Etc.getgrent)
     end
 
     # closes handle to /etc/group file
@@ -35,7 +52,7 @@ module Puppet::Etc
     # On first call opens /etc/passwd and returns parse of first entry. Each subsquent call
     # returns new struct for the next entry or nil if EOF. Call ::endgrent to close file.
     def getpwent
-      convert_field_values_to_utf8!(::Etc.getpwent)
+      override_field_values_to_utf8!(::Etc.getpwent)
     end
 
     # closes handle to /etc/passwd file
@@ -53,69 +70,68 @@ module Puppet::Etc
     # returns an Etc::Passwd struct corresponding to the entry or raises
     # ArgumentError if none
     def getpwnam(username)
-      convert_field_values_to_utf8!(::Etc.getpwnam(username))
+      override_field_values_to_utf8!(::Etc.getpwnam(username))
     end
 
     # Etc::getgrnam searches /etc/group file for an entry corresponding to groupname.
     # returns an Etc::Group struct corresponding to the entry or raises
     # ArgumentError if none
     def getgrnam(groupname)
-      convert_field_values_to_utf8!(::Etc.getgrnam(groupname))
+      override_field_values_to_utf8!(::Etc.getgrnam(groupname))
     end
 
     # Etc::getgrid searches /etc/group file for an entry corresponding to id.
     # returns an Etc::Group struct corresponding to the entry or raises
     # ArgumentError if none
     def getgrgid(id)
-      convert_field_values_to_utf8!(::Etc.getgrgid(id))
+      override_field_values_to_utf8!(::Etc.getgrgid(id))
     end
 
     # Etc::getpwuid searches /etc/passwd file for an entry corresponding to id.
     # returns an Etc::Passwd struct corresponding to the entry or raises
     # ArgumentError if none
     def getpwuid(id)
-      convert_field_values_to_utf8!(::Etc.getpwuid(id))
+      override_field_values_to_utf8!(::Etc.getpwuid(id))
     end
 
     private
-    # Utility method for converting the String values of a struct returned by
+    # Utility method for overriding the String values of a struct returned by
     # the Etc module to UTF-8. Structs returned by the ruby Etc module contain
     # members with fields of type String, Integer, or Array of Strings, so we
     # handle these types. Otherwise ignore fields.
     #
-    # NOTE: If a string cannot be converted to UTF-8, this leaves the original
-    # string string intact in the Struct.
+    # NOTE: If a string cannot be overidden to UTF-8 because it would be invalid
+    # in that encoding, this leaves the original string intact and unmodified in
+    # the Struct.
     #
     # Warning! This is a destructive method - the struct passed is modified!
     #
     # @api private
     # @param [Etc::Passwd or Etc::Group struct]
     # @return [Etc::Passwd or Etc::Group struct] the original struct with values
-    #   converted to UTF-8 if possible, or the original value intact if not
-    def convert_field_values_to_utf8!(struct)
+    #   overidden to UTF-8 if possible, or the original value intact if not
+    def override_field_values_to_utf8!(struct)
       return nil if struct.nil?
       struct.each_with_index do |value, index|
         if value.is_a?(String)
-            converted = Puppet::Util::CharacterEncoding.convert_to_utf_8!(value)
-            struct[index] = converted if !converted.nil?
+          Puppet::Util::CharacterEncoding.override_encoding_to_utf_8!(value) if value.encoding != Encoding::UTF_8
         elsif value.is_a?(Array)
-          struct[index] = convert_array_values_to_utf8!(value)
+          override_array_values_to_utf8!(value)
         end
       end
     end
 
-    # Helper method for ::convert_field_values_to_utf8!
+    # Helper method for ::override_field_values_to_utf8!
     #
     # Warning! This is a destructive method - the array passed is modified!
     #
     # @api private
-    # @param [Array] object containing String values to convert to UTF-8
-    # @return [Array] original Array with String values converted to UTF-8 if
-    #   convertible, or original, unmodified values if not.
-    def convert_array_values_to_utf8!(string_array)
-      string_array.map! do |elem|
-        converted = Puppet::Util::CharacterEncoding.convert_to_utf_8!(elem)
-        converted.nil? ? elem : converted
+    # @param [Array] object containing String values to override to UTF-8
+    # @return [Array] original Array with String values overidden to UTF-8 if
+    #   they would be valid in UTF-8 or original, unmodified values if not.
+    def override_array_values_to_utf8!(string_array)
+      string_array.each do |elem|
+        Puppet::Util::CharacterEncoding.override_encoding_to_utf_8!(elem)
       end
     end
   end
--- a/lib/puppet/etc.rb
+++ b/lib/puppet/etc.rb
@@ -16,7 +16,7 @@ require 'puppet/util/character_encoding'
 # ruby will label a string containing a four-byte characters such as "\u{2070E}"
 # as EUC_KR even though EUC_KR is a two-byte width encoding.
 #
-# On Ruby 2.0.0 and earlier, Etc will always return string values in BINARY,
+# On Ruby 2.0.x and earlier, Etc will always return string values in BINARY,
 # ignoring encoding altogether.
 #
 # For Puppet we specifically want UTF-8 as our input from the Etc module - which
@@ -25,17 +25,26 @@ require 'puppet/util/character_encoding'
 # disk as UTF-8. Etc is defined for Windows but the majority calls to it return
 # nil and Puppet does not use it.
 #
+# That being said, we have cause to retain the original, pre-override string
+# values. `puppet resource user`
+# (Puppet::Resource::User.indirection.search('User', {})) uses self.instances to
+# query for user(s) and then iterates over the results of that query again to
+# obtain state for each user. If we've overridden the original user name and not
+# retained the original, we've lost the ability to query the system for it
+# later. Hence the Puppet::Etc::Passwd and Puppet::Etc::Group structs.
+#
 # We only use Etc for retrieving existing property values from the system. For
 # setting property values, providers leverage system tools (i.e., `useradd`)
 #
 # @api private
 module Puppet::Etc
   class << self
+
     # Etc::getgrent returns an Etc::Group struct object
     # On first call opens /etc/group and returns parse of first entry. Each subsquent call
     # returns new struct the next entry or nil if EOF. Call ::endgrent to close file.
     def getgrent
-      override_field_values_to_utf8!(::Etc.getgrent)
+      override_field_values_to_utf8(::Etc.getgrent)
     end
 
     # closes handle to /etc/group file
@@ -52,7 +61,7 @@ module Puppet::Etc
     # On first call opens /etc/passwd and returns parse of first entry. Each subsquent call
     # returns new struct for the next entry or nil if EOF. Call ::endgrent to close file.
     def getpwent
-      override_field_values_to_utf8!(::Etc.getpwent)
+      override_field_values_to_utf8(::Etc.getpwent)
     end
 
     # closes handle to /etc/passwd file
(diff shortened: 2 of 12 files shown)


lib/puppet/etc.rb
```diff
@@ -96,10 +113,9 @@ def convert_field_values_to_utf8!(struct)
       return nil if struct.nil?
       struct.each_with_index do |value, index|
         if value.is_a?(String)
-            converted = Puppet::Util::CharacterEncoding.convert_to_utf_8!(value)
-            struct[index] = converted if !converted.nil?
+          Puppet::Util::CharacterEncoding.override_encoding_to_utf_8!(value) if value.encoding != Encoding::UTF_8
```
MosesMendoza: @joshcooper @Iristyle This change is the crux of our discussion - This now basically says we are explicitly overriding the encoding of the string returned by Etc to UTF-8 (not converting it) if it is a) not already UTF-8 and b) would be valid UTF-8. We are starting from the assumption that it is incorrectly labeled with whatever encoding it has (ie ISO-8859-1) because we have reason to believe it is actually UTF-8 because Puppet set (or will set) the value to UTF-8.

Just to reiterate for PR posterity why I believe this is the correct behavior: Consider these two scenarios. In the first, we default to assuming the encoding is correct from Etc and do `encode!` and in the second we assume the encoding is incorrect and override it with `force_encoding`.

* Puppet agent is creating a new user with managed comment property. The host is running in ISO-8859-1 encoding. When puppet agent get's its `should` value via the catalog, the should value is UTF-8 and its bytes are written to disk as-is with no re-encoding/modification (we don't use Etc for writing, only reading). When Puppet runs next time, Ruby::Etc will read in these bytes and label them as ISO-8859-1, which will likely be valid in that encoding. At least, our commonly used `mixed_utf8` string in tests is "valid" ISO-8859-1. When we pipe these values through Puppet::Etc, they'll now be `encode!`d as UTF-8 - changing the byte representation. So our original UTF-8 has now been modified to something else. When Puppet compares this `is` value to the catalog `should` value, they won't be the same... so Puppet will consider the resource out of sync and write the `should` value to disk again - and will every time it runs.

```
[1] pry(main)> mixed_utf_8 = "A\u06FF\u16A0\u{2070E}"
=> "Aۿᚠ𠜎"
[2] pry(main)> mixed_utf_8.bytes
=> [65, 219, 191, 225, 154, 160, 240, 160, 156, 142]
[3] pry(main)> mixed_utf_8.force_encoding(Encoding::ISO_8859_1)
=> "A\xDB\xBF\xE1\x9A\xA0\xF0\xA0\x9C\x8E"
[4] pry(main)> mixed_utf_8.valid_encoding?
=> true
[5] pry(main)> mixed_utf_8.encode!(Encoding::UTF_8)
=> "AÛ¿á\u009A ð \u009C\u008E"
[6] pry(main)> mixed_utf_8.bytes
=> [65, 195, 155, 194, 191, 195, 161, 194, 154, 194, 160, 195, 176, 194, 160, 194, 156, 194, 142]
[7] pry(main)> orig_mixed_utf_8 = "A\u06FF\u16A0\u{2070E}"
=> "Aۿᚠ𠜎"
[8] pry(main)> mixed_utf_8 == orig_mixed_utf_8
=> false
```

`Etc` is a special case for us.  Consider the same scenario above - ie host in ISO-8859-1 managed by Puppet, but instead believing the non-UTF-8 encoding is inaccurate and doing `force_encoding!`. 

Puppet will be managing the properties of a user, but Puppet has not run yet on the host. The values on disk or accurate an accurate representation of their encoding, ie ISO-8859-1. When Puppet first runs, these values are read in and incorrectly re-labeled (overridden) as UTF-8 by Puppet::Etc. Puppet compares these values to the the UTF-8 `should` values, finds them out of sync, and overwrites them on the system with the UTF-8 values. The next time Puppet runs, the UTF-8 values are read in as ISO-8859-1, but correctly overridden by Puppet::Etc to UTF-8, and Puppet correctly finds them in sync with the catalog `should` values. Even if we had not incorrectly overridden the encoding on the very first Puppet run, the result would have been the same - Puppet would have found the values out of sync, as the ISO-8859-1 value would not equal the UTF-8 value, and corrected them.

spec/unit/util/character_encoding_spec.rb
```diff
@@ -1,88 +1,118 @@
...
-          expect(result.encoding).to eq(Encoding::UTF_8)
-        end
+      it "should be able to convert BINARY to UTF-8 by labeling as Encoding.default_external" do
+        # そ - HIRAGANA LETTER SO
+        # In Windows_31J: \x82 \xbb - 130 187
+        # In Unicode: \u305d - \xe3 \x81 \x9d - 227 129 157
+
+        # When received as BINARY are not transcodable, but by "guessing"
+        # Encoding.default_external can transcode to UTF-8
+        win_31j = [130, 187].pack('C*')
```
Iristyle: We should probably introduce a failure spec here that shows an incoming BINARY byte sequence that is *not* valid in `Encoding.default_external` ... i.e. it should raise an exception of `InvalidByteSequenceError`

MosesMendoza: Test added

lib/puppet/util/character_encoding.rb
```diff
@@ -3,79 +3,72 @@
...
+    #
     # @api public
-    # @param [String] string a string to transcode / force_encode to utf-8
-    # @return [String] string if already utf-8, OR
-    #   the same string with external encoding set to utf-8 if bytes are valid utf-8 OR
-    #   the same string transcoded to utf-8 OR
-    #   nil upon a failure to legitimately set external encoding or transcode string
+    # @param [String] string a string to transcode
+    # @return [nil] (string is modified in place)
     def convert_to_utf_8!(string)
```
Iristyle: We don't need to address it here, but it seems to me that `convert_to_utf_8` should return a copy of the initial string rather than mutating it...

lib/puppet/util/character_encoding.rb
```diff
@@ -3,79 +3,72 @@
...
+    # of the encoding of the string.
+    #
+    # @api public
+    # @param [String] string set external encoding (re-label) to utf-8
+    # @return [nil] (string is modified in place)
+    def override_encoding_to_utf_8!(string)
+      if valid_utf_8_bytes?(string)
+        string.force_encoding(Encoding::UTF_8)
+      else
+        Puppet.debug(_("%{value} is not valid UTF-8 and result of overriding encoding would be invalid.") % { value: string.dump })
```
Iristyle: Should we be emitting a string into debug output that we know isn't UTF-8, or will `dump` always produce UTF-8 due to its behavior? 

https://ruby-doc.org/core-2.2.0/String.html#method-i-dump

spec/unit/util/character_encoding_spec.rb
```diff
@@ -1,88 +1,118 @@
...
   end
+
+  describe "::override_encoding_to_utf_8!" do
+    context "given a string with bytes that represent valid UTF-8" do
+      it "should set the external encoding of the string to UTF-8" do
+        # ☃ - unicode snowman
+        # \u2603 - \xe2 \x98 \x83 - 226 152 131
+        snowman = [226, 152, 131].pack('C*')
+        Puppet::Util::CharacterEncoding.override_encoding_to_utf_8!(snowman)
+        expect(snowman).to eq("\u2603")
```
Iristyle: Should we add to the expectation:

```ruby
expect(snowman.encoding).to eq(Encoding::UTF_8)
```

lib/puppet/util/character_encoding.rb
```diff
@@ -3,79 +3,72 @@
...
+      end
+      nil
     end
 
     private
```
Iristyle: Can the `valid_bytes` implementation be simplified to?

```ruby
def valid_utf_8_bytes?(string)
  string.dup.force_encoding(Encoding::UTF_8).valid_encoding?
end
```

Given one of your prior PRs, we determined `.dup` doesn't copy a string, but just creates a new wrapper around the same underlying string data - making `dup` pretty cheap.  This lets us test bytes without having to do the save / revert of encoding on the original string.

spec/unit/etc_spec.rb
```diff
@@ -1,234 +1,325 @@
...
+      let(:overridden) {
+        PuppetSpec::CharacterEncoding.with_external_encoding(Encoding::EUC_KR) do
+          Puppet::Etc.send(subject, *params)
+        end
+      }
 
-      it "should convert convertible values in arrays to UTF-8" do
-        expect(converted.mem[0]).to eq("A\u06FF\u16A0\u{2070E}")
-        expect(converted.mem[0].encoding).to eq(Encoding::UTF_8) # just being explicit
+      it "should override EUC_KR-labeled values in arrays to UTF-8 if that would result in valid UTF-8" do
```
Iristyle: Note to re-examine these tests... not sure on the logic exactly.  Might be helpful to make a small map of inputs / outputs / expectations to get a better handle on it.

spec/unit/util/character_encoding_spec.rb
```diff
@@ -47,6 +47,25 @@
         expect(win_31j.bytes.to_a).to eq([227, 129, 157])
       end
 
+      context "that is BINARY encoded but invalid in Encoding.default_external" do
+        let(:invalid_win_31j) { [255, 254, 253].pack('C*') } # these bytes are not valid windows_31j
+
+        it "should leave the string umodified" do
+          PuppetSpec::CharacterEncoding.with_external_encoding(Encoding::Windows_31J) do
+            Puppet::Util::CharacterEncoding.convert_to_utf_8!(invalid_win_31j)
```
Iristyle: 👍

spec/unit/etc_spec.rb
MosesMendoza: These were invalid, and they were being masked because we were modifying the expected original values. Pushed a new commit that ensures we're acting on copies for our tests so that the originals go unmodified.

spec/unit/util/character_encoding_spec.rb
MosesMendoza: 👍  added

lib/puppet/util/character_encoding.rb
MosesMendoza: 👍  that seems like a reasonable optimization

MosesMendoza: added

lib/puppet/util/character_encoding.rb
MosesMendoza: Right - dump will print the escaped hex values of the string ie so the ultimate character output is, strictly speaking, valid UTF-8, even though it doesn't represent valid UTF-8.
```
[1] pry(main)> s = [255, 254, 253].pack('C*').force_encoding(Encoding::UTF_8)
=> "\xFF\xFE\xFD"
[2] pry(main)> s.valid_encoding?
=> false
[3] pry(main)> output = "bar #{s}"
=> "bar \xFF\xFE\xFD"
[4] pry(main)> puts output
bar ���
=> nil
[5] pry(main)> output = "bar #{s.dump}"
=> "bar \"\\xFF\\xFE\\xFD\""
```

spec/unit/etc_spec.rb
```diff
@@ -45,7 +49,7 @@
     group.name = euc_kr_as_utf_8
     # group passwd field is valid UTF-8
     group.passwd = mixed_utf_8
-    group
+    group = duplicate_struct_values(group)
```
Iristyle: Why not just:

```ruby
duplicate_struct_values(group)
```

MosesMendoza: thanks for pointing that out

Iristyle:
Everything is looking pretty good... One last thing to check before merge is:

* Configure a Linux system with Latin1
* Create a user with a name that contains a Latin1 accented character that is represented with different bytes in UTF-8
* Create a Puppet manifest that tries to create the same user, but with the name represented in UTF-8
* Run puppet apply on manifest / observe output
* Run puppet apply on manifest again to check for idempotency / observe output
* Run `puppet resource user` at the shell with Latin1 encoding - observe output
* Run `puppet resource user` at the shell with UTF-8 encoding - observe output


Make sure we get what we want here...

MosesMendoza:
Re-added blocked label. in running through manual testing with @Iristyle we determined that this works in a manifest driven workflow with agent in LATIN-1, but fails when running puppet resource directly on the agent when the resource values contain a mix of latin-1 and utf-8 strings.

lib/puppet/application/resource.rb
```diff
@@ -138,14 +138,16 @@ def main
...
+        end.join("\n")
         text.prepend("#{type.downcase}:\n")
       else
-        text = resources.
-          map { |resource| resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest }.
-          join("\n")
+        text = resources.map do |resource|
+          r = resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest
+          r.encoding == Encoding.default_external ? r : r.force_encoding(Encoding.default_external)
+         end.join("\n")
```
MosesMendoza: This change needs a test

lib/puppet/etc.rb
```diff
@@ -53,70 +79,68 @@ def setpwent
...
     # @api private
     # @param [Etc::Passwd or Etc::Group struct]
-    # @return [Etc::Passwd or Etc::Group struct] the original struct with values
-    #   converted to UTF-8 if possible, or the original value intact if not
-    def convert_field_values_to_utf8!(struct)
+    # @return [Puppet::Etc::Passwd or Puppet::Etc::Group struct] a new struct
+    #  object with the original struct values overidden to UTF-8. For each member
+    #  the struct also contains a corresponding :canonical_<member name> struct
+    #  member
+    def override_field_values_to_utf8(struct)
```
MosesMendoza: These tests need updated

lib/puppet/etc.rb
```diff
@@ -53,70 +79,68 @@ def setpwent
...
+    # Defines Puppet::Etc::Passwd struct class. Contains all of the
+    # original member fields of Etc::Passwd, and additional "canonical_"
+    # versions of these fields as well. API compatible with Etc::Passwd.
+    Puppet::Etc::Passwd = Struct.new(*Etc::Passwd.members, *Etc::Passwd.members.map { |member| "canonical_#{member}".to_sym })
+
+    # @api private
+    # Defines Puppet::Etc::Group struct class. Contains all of the original
+    # member fields of Etc::Group, and additional "canonical_" versions of these
+    # fields as well. API compatible with Etc::Group.
+    Puppet::Etc::Group = Struct.new(*Etc::Group.members, *Etc::Group.members.map { |member| "canonical_#{member}".to_sym })
```
MosesMendoza: should get coverage for these new struct objects

lib/puppet/etc.rb
```diff
@@ -35,7 +61,7 @@ def setgrent
     # On first call opens /etc/passwd and returns parse of first entry. Each subsquent call
     # returns new struct for the next entry or nil if EOF. Call ::endgrent to close file.
     def getpwent
-      convert_field_values_to_utf8!(::Etc.getpwent)
+      override_field_values_to_utf8(::Etc.getpwent)
```
MosesMendoza: should get coverage for the new struct objects via these methods

lib/puppet/util/character_encoding.rb
```diff
@@ -3,78 +3,84 @@
...
+    #
+    # This method is intended for situations where we do not believe that the
+    # encoding associated with a string is an accurate reflection of its actual
+    # bytes, i.e., effectively when we believe Ruby is incorrect in its assertion
+    # of the encoding of the string.
+    #
+    # @api public
+    # @param [String] string to set external encoding (re-label) to utf-8
+    # @return [String] a copy of string with external encoding set to utf-8
+    def override_encoding_to_utf_8(string)
```
MosesMendoza: per commit comment, ultimately kept this method (albeit with non-destructive form) because we have need to only override if the bytes are invalid when running `puppet resource user` in non-utf8 form. Otherwise, we then issue replacement characters in our now-invalid-utf8 string, which causes incorrect puppet resource user output.

lib/puppet/util/character_encoding.rb
```diff
@@ -3,78 +3,84 @@
...
+    # sequences in its current encoding replaced with "?"
+    # @param string a string to remove invalid byte sequences from
+    # @return a copy of string invalid byte sequences replaced by "?" character
+    # @note does not modify encoding, but new string will have different bytes
+    #   from original
+    def scrub(string)
+      if String.method_defined?(:scrub)
+        string.scrub
+      else
+        string.chars.map { |c| c.valid_encoding? ? c : "?"}.join
```
MosesMendoza: scrub isn't present in 1.9.3, and various iterations on encode with replacement characters failed to work. supplying utf-8 for the first two args to encode is a no-op, and supplying binary as the second arg causes any non-ascii character to be replaced regardless of the original encoding or validity of the string...

MosesMendoza:
TODO: need test fixups

@Iristyle this is an implementation using extended structs with canonical values. ultimately not sure it's worth the extra logic, but it seems cleaner and more future proof though in case there are other places in the codebase that need what we have done in nameservice (still need to audit). as-is though I think this now works in all the cases we were testing together previously.

lib/puppet/util/character_encoding.rb
MosesMendoza: Updated - no longer a destructive method

lib/puppet/util/character_encoding.rb
```diff
@@ -2,94 +2,88 @@
...
+    # sure our output is consistent across ruby versions and encodings, and
+    # because calling scrub on a non-UTF8 string with the unicode replacement
+    # character "\uFFFD" results in an Encoding::CompatibilityError.
+    # @param string a string to remove invalid byte sequences from
+    # @return a copy of string invalid byte sequences replaced by "?" character
+    # @note does not modify encoding, but new string will have different bytes
+    #   from original. Only needed for ruby 1.9.3 support.
+    def scrub(string)
+      if String.method_defined?(:scrub)
+        string.scrub("?")
```
Iristyle: Let's use `string.scrub` with default replacement character behavior here.

lib/puppet/util/character_encoding.rb
```diff
@@ -2,94 +2,88 @@
...
+      else
+        string.chars.map { |c| c.valid_encoding? ? c : "?" }.join
```
Iristyle: And lets select the replacement character based on input string encoding here...

* `"\uFFFD"` for UTF-8
* `?` for everything else

Iristyle: And lets select the replacement character based on input string encoding here...

* `"\uFFFD"` for UTF-8
* `?` for everything else

lib/puppet/util/character_encoding.rb
```diff
@@ -2,94 +2,88 @@
...
-          # in UTF-8 and do not transcode it - we only set external encoding.
-          return string.force_encoding(Encoding::UTF_8)
-        elsif currently_valid
-          # If the string is not currently valid UTF-8 but it can be transcoded
-          # (it is valid in its current encoding), we can guess this string was
-          # not originally unicode. Transcode it to UTF-8. For strings with
-          # original encodings like SHIFT_JIS, this should be the final result.
-          return string.encode!(Encoding::UTF_8)
+          # String is aleady valid UTF-8 - noop
+          return string
```
Iristyle: The API says you get a copy... if so, this should be `string.dup`

lib/puppet/util/character_encoding.rb
```diff
@@ -2,94 +2,88 @@
...
         end
       rescue EncodingError => detail
         # Catch both our own self-determined failure to transcode as well as any
         # error on ruby's part, ie Encoding::UndefinedConversionError on a
         # failure to encode!.
-        Puppet.debug(_("%{error}: %{value} is not valid UTF-8 and cannot be transcoded by Puppet.") %
+        Puppet.debug(_("%{error}: %{value} cannot be transcoded by Puppet.") %
           { error: detail.inspect, value: string.dump })
-        return nil
+        return string
```
Iristyle: The API says you get a copy... if so, this should be `string.dup`

lib/puppet/util/character_encoding.rb
```diff
@@ -2,94 +2,88 @@
...
+    # encoding associated with a string is an accurate reflection of its actual
+    # bytes, i.e., effectively when we believe Ruby is incorrect in its
+    # assertion of the encoding of the string.
+    #
+    # @api public
+    # @param [String] string to set external encoding (re-label) to utf-8
+    # @return [String] a copy of string with external encoding set to utf-8, or
+    # the original string if override would result in invalid encoding.
+    def override_encoding_to_utf_8(string)
+      return string if string.encoding == Encoding::UTF_8
```
Iristyle: The API says you get a copy... if so, this should be `string.dup`

lib/puppet/util/character_encoding.rb
```diff
@@ -2,94 +2,88 @@
...
+      dup = string.dup
+      if dup.force_encoding(Encoding::UTF_8).valid_encoding?
+        return dup
+      else
+        Puppet.debug(_("%{value} is not valid UTF-8 and result of overriding encoding would be invalid.") % { value: string.dump })
+        return string
```
Iristyle: The API says you get a copy... if so, this should be `string.dup`

lib/puppet/util/character_encoding.rb
MosesMendoza: 👍

lib/puppet/etc.rb
```diff
@@ -53,70 +79,77 @@ def setpwent
...
+     private
+
+    # @api private
+    # Defines Puppet::Etc::Passwd struct class. Contains all of the original
+    # member fields of Etc::Passwd, and additional "canonical_" versions of
+    # these fields as well. API compatible with Etc::Passwd. Because Struct.new
+    # defines a new Class object, we memozie to avoid superfluous extra Class
+    # instantiations.
+    def puppet_etc_passwd_class
+      @password_class ||= Struct.new(*Etc::Passwd.members, *Etc::Passwd.members.map { |member| "canonical_#{member}".to_sym })
```
Iristyle: Are these all strings?

lib/puppet/etc.rb
```diff
@@ -53,70 +79,77 @@ def setpwent
...
+    end
+
+    # @api private
+    # Defines Puppet::Etc::Group struct class. Contains all of the original
+    # member fields of Etc::Group, and additional "canonical_" versions of these
+    # fields as well. API compatible with Etc::Group. Because Struct.new
+    # defines a new Class object, we memoize to avoid superfluous extra Class
+    # instantiations.
+    def puppet_etc_group_class
+      @group_class ||= Struct.new(*Etc::Group.members, *Etc::Group.members.map { |member| "canonical_#{member}".to_sym })
```
Iristyle: Are these all strings?

lib/puppet/util/character_encoding.rb
MosesMendoza: 👍

lib/puppet/etc.rb
MosesMendoza: The member names will be strings - the values can vary

MosesMendoza: Well, strictly speaking, all the member names are returned here as an array of symbols, but the interpolation implicitly calls to_s on the symbol

lib/puppet/etc.rb
MosesMendoza: Same

lib/puppet/util/character_encoding.rb
MosesMendoza: updated to return a copy in all cases

MosesMendoza:
cleaned up whitespace. This ran a local ci:test:git run against ubuntu 1604 and appeared to only fail in expected ways - at least none apparently related to this PR

MosesMendoza:
Should be good to merge

lib/puppet/etc.rb
Iristyle: Ah sorry, what I meant was... aren't `canonical_` fields only useful for string values?

For instance, `canonical_expire` isn't really useful, is it?

MosesMendoza: Nope, not particularly useful. To avoid that, we could hard code the list of available attributes here? I thought about it but decided to go with this because there's a couple things that make that more annoying
- the list varies based on the struct class (and knowing ruby could change without warning) so we maintain different lists based on if we're doing a user or group struct
- the list actually varies based on the flags passed to [ruby during compilation](https://ruby-doc.org/stdlib-2.3.0/libdoc/etc/rdoc/Etc.html#label-The+following+members+below+are+optional-2C+and+must+be+compiled+with+special+flags-3A) . This means we'd need to conditionally check for the presence of the members that we expect to exist and have string values and only duplicate those. I feel like there's a tradeoff there in maintainability vs the completeness of just duplicating all the members
- A sidenote that doesn't necessarily impact this discussion but is worth noting is that the Etc struct members aren't actually typed - any field could just as easily return a string as an integer. So if for some reason the underlying system changes the return value for any of these we're going to be in trouble. We could guard by only trying to encode things that are strings rather than expecting specific fields to be, but again that felt like going to far. Happy to entertain alternatives there though.

Iristyle:
FYI - I think this should be targeted at stable... I've rebased and I'm testing that locally.

MosesMendoza:
@Iristyle rebased against stable

lib/puppet/application/resource.rb
```diff
@@ -136,14 +136,16 @@ def main
     resources = find_or_save_resources(type, name, params)
 
     if options[:to_yaml]
-      text = resources.
-        map { |resource| resource.prune_parameters(:parameters_to_include => @extra_params).to_hierayaml }.
-        join("\n")
+      text = resources.map do |resource|
+        r = resource.prune_parameters(:parameters_to_include => @extra_params).to_hierayaml
+        r.encoding == Encoding.default_external? ? r : r.force_encoding(Encoding.default_external)
```
Iristyle: Can this just be simplified to `r.force_encoding(Encoding.default_external)` - i.e. is the check roughly as costly as performing the effectual `no-op`?

MosesMendoza: good idea

MosesMendoza: updated

lib/puppet/util/character_encoding.rb
```diff
@@ -86,10 +100,7 @@ def convert_to_utf_8!(string)
     # @param [String] string a string to test
     # @return [Boolean] whether we think the string is UTF-8 or not
     def valid_utf_8_bytes?(string)
```
Iristyle: `valid_utf_8_bytes` was not in prior commit and should be torched

MosesMendoza: oh hell

Iristyle:
Merged to stable in 365d6a966dc77e70837b90f2f9d6265a48765ce0

PR #5801, opened 2017-04-12
(PUP-7425) Update required version of net-ssh

Iristyle:
- The version of net-ssh Puppet previously consumed was ~> 2.1
   but this version conflicts with Beaker and can cause gem installs,
   particularly for module testing, to fail.

   Beaker 3 moved to net-ssh 3.3.0beta1 to resolve some Unicode
   related issues (see BKR-40 for further detail). When net-ssh 4
   became available, Beaker 3.11.0 and higher began consuming it to
   resolve other Bundler workflow problems.

   net-ssh usage is limited to `puppet device` support and should be
   safe to generally bump to a newer version, especially one that
   resolves known UTF-8 issues.

 - To allow for Puppet to run with any version of Beaker, allow
   net-ssh 3 and 4 versions to be used by adjusting the .gemspec
   accordingly.

   To continue to allow the Puppet gem to be tested against Ruby
   1.9.3, exclude net-ssh for Ruby < 2

   Note that the .gemspec in the repository is not yet the one used
   for packaging / shipping gems. It is however consumed inside
   agent packages that bundle / vendor specific gem versions.

 - Continue to not make net-ssh a part of project_data.yaml which
   feeds into the shipping .gemspec through packaging automation
   tasks. puppet-agent packages do currently include net-ssh 2.9.2
   (which will be updated soon in packaging to 4.1.0). At that time,
   Puppet server should not be negatively impacted by bumping the
   version of net-ssh to one incompatible with its JRuby / Ruby
   compatibility (1.9.3) given net-ssh usage is guarded behind a
   feature inside Puppet. puppet-agent itself already vendors Ruby
   2.1.9 and would be compatible.

   The gem situation is a bit more complex.

   The Puppet gem is currently marked as compatible with a minimum
   Ruby of 1.9.3.  Bundler may not use the RUBY_VERSION specified by
   the gem when resolving dependencies which could result in an attempt
   to install an incompatible net-ssh gem under 1.9.3. This likely
   *wouldn't* be a problem in a Beaker acceptance scenario, as Ruby
   1.9.3 already requires the older Beaker 2, which includes a more
   specific net-ssh dependency like ~> 2.9.

   However, it is unclear what other situations could fail to install
   net-ssh when Puppet is consumed as a gem, so for the sake of
   preventing disruption, net-ssh as a shipped gem dependency will
   remain omitted, to be handled at a later date.

For reference, BKR-995 updated the Beaker `net-ssh` dependency for `3.11.0` at https://github.com/puppetlabs/beaker/commit/7fb98233b2fa989d9c3795fb8c15169d5eb93f4f

### DIFF
--- a/.gemspec
+++ b/.gemspec
@@ -45,7 +45,10 @@ Gem::Specification.new do |s|
   # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
   s.add_runtime_dependency(%q<json_pure>, "~> 1.8")
   # net-ssh is a runtime dependency of Puppet::Util::NetworkDevice::Transport::Ssh
-  s.add_runtime_dependency(%q<net-ssh>, "~> 2.1")
+  # Beaker 3.0.0 to 3.10.0 depends on net-ssh 3.3.0beta1
+  # Beaker 3.11.0+ depends on net-ssh 4.0+
+  # be lenient to allow module testing where Beaker and Puppet are in same Gemfile
+  s.add_runtime_dependency(%q<net-ssh>, [">= 3.0", "< 5"]) if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.0.0')
 
   # loads platform specific gems like ffi, win32 platform gems
   # as additional runtime dependencies


Iristyle:
Arg, net-ssh is incompatible with Ruby 1.9.3, which we're still testing in Travis.

.gemspec
```diff
@@ -45,7 +45,10 @@ Gem::Specification.new do |s|
   # json_pure 2.0.2+ officially requires Ruby >= 2.0, but should have specified that in 2.0
   s.add_runtime_dependency(%q<json_pure>, "~> 1.8")
   # net-ssh is a runtime dependency of Puppet::Util::NetworkDevice::Transport::Ssh
-  s.add_runtime_dependency(%q<net-ssh>, "~> 2.1")
+  # Beaker 3.0.0 to 3.10.0 depends on net-ssh 3.3.0beta1
+  # Beaker 3.11.0+ depends on net-ssh 4.0+
+  # be lenient to allow module testing where Beaker and Puppet are in same Gemfile
+  s.add_runtime_dependency(%q<net-ssh>, [">= 3.0", "< 5"]) if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.0.0')
```
Iristyle: Yup, I'm aware that `.gemspec` should not generally contain dynamic dependencies like this (one way to get around that is with `ext/mkrf_conf.rb`).  But note that this is not the main `.gemspec` for Puppet as a gem - that one is built during packaging by a rake task.  This `.gemspec` only influences a local `bundle install` (and is also shipped as part of agent packages which maintain fixed dependencies when built).  See https://tickets.puppetlabs.com/browse/PA-1077 for details on which files are used for what - it's a bit complicated currently.

I already tried to use the `[">= 2.1", "<5 "]` but unfortunately that still failed on Ruby 1.9.3.  You can see the commit at https://github.com/puppetlabs/puppet/commit/9e4cab2b34296b19b8ac5afc24bfff3bee3bf645 / CI result at https://travis-ci.org/puppetlabs/puppet/builds/221556267

I believe Bundler has a bug where it doesn't actually consider `RUBY_VERSION` when resolving dependencies without a version guard (they might have fixed it in newer versions, but I've definitely run into it before).  That *could* be addressed by adding even more dynamic resolution, but I think it's unnecessary (though that could be debated).  Another thing we can do is remove `net-ssh` altogether, but I believe that's misleading given there is code in Puppet that consumes `net-ssh`.

Iristyle:
Ok - looks like I misread how the `net-ssh` dependency was included.  It turns out that it *is* guarded by a feature at https://github.com/puppetlabs/puppet/blob/3a5dfed0fef8fc5cb1b6ec168766115de3e8f23b/lib/puppet/feature/ssh.rb#L3

That may change my plan of attack here.

Iristyle:
With fresh clone of ACL module under 


#### Ruby 2.1.9:

```
PUPPET_GEM_VERSION=file:///Users/Iristyle/source/puppet bundle install --path .bundle/gems

Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Resolving dependencies................

...

Installing net-ssh 4.1.0
Installing puppet-module-posix-default-r2.1 0.0.3
Using puppet 4.10.0 from source at `/Users/Iristyle/source/puppet`
Installing beaker 3.14.0
Installing puppet-module-posix-dev-r2.1 0.0.3
Installing puppet-module-posix-system-r2.1 0.0.3

...

Bundle complete! 13 Gemfile dependencies, 153 gems now installed.
Bundled gems are installed into ./.bundle/gems.
```

#### Ruby 2.3.3

```
PUPPET_GEM_VERSION=file:///Users/Iristyle/source/puppet bundle install --path .bundle/gems

The dependency puppet-module-win-default-r2.3 (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mswin32, x86-mingw32, x64-mingw32. To add those platforms to the bundle, run `bundle lock --add-platform x86-mswin32 x86-mingw32 x64-mingw32`.
The dependency puppet-module-win-dev-r2.3 (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mswin32, x86-mingw32, x64-mingw32. To add those platforms to the bundle, run `bundle lock --add-platform x86-mswin32 x86-mingw32 x64-mingw32`.
The dependency puppet-module-win-system-r2.3 (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mswin32, x86-mingw32, x64-mingw32. To add those platforms to the bundle, run `bundle lock --add-platform x86-mswin32 x86-mingw32 x64-mingw32`.
Fetching gem metadata from https://rubygems.org/........
Fetching version metadata from https://rubygems.org/.
Resolving dependencies......
Unable to use the platform-specific (universal-darwin) version of facter (2.4.6) because it has different dependencies from the  version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.

...

Installing net-ssh 4.1.0
Installing puppet-module-posix-default-r2.3 0.0.3
Using puppet 4.10.0 from source at `/Users/Iristyle/source/puppet`
Installing beaker 3.14.0
Installing puppet-module-posix-dev-r2.3 0.0.3
Installing puppet-module-posix-system-r2.3 0.0.3

...

Bundle complete! 13 Gemfile dependencies, 154 gems now installed.
Bundled gems are installed into ./.bundle/gems.
```


#### Ruby 1.9.3-p551 (with ACL module version 1.1.2)

This doesn't work anyway for other reasons... 

```
PUPPET_GEM_VERSION=file:///Users/Iristyle/source/puppet bundle install --path .bundle/gems

Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Resolving dependencies......

... 

Gem::InstallError: public_suffix requires Ruby version >= 2.0.

...

Installing net-ssh 2.9.4
Using bundler 1.12.5

...

An error occurred while installing public_suffix (2.0.5), and Bundler cannot continue.
Make sure that `gem install public_suffix -v '2.0.5'` succeeds before bundling.
```

.gemspec
Iristyle: The Bundler issue I mentioned above is documented at https://github.com/bundler/bundler/pull/4651 - some more info can be found at https://stackoverflow.com/questions/40348066/can-bundler-automatically-solve-ruby-version-dependencies

PR #5802, opened 2017-04-14
(maint) Use instance var for puppetserver config

adrienthebo:


joshcooper:
It looks like the instance variable and reader work on the same underlying hash:

```
(byebug) options.class
Beaker::Options::OptionsHash
(byebug) options == @options
true
(byebug) @options['puppetserver-config']
nil
(byebug) options['puppetserver-config']
nil
```

I think the problem is that there is something different about puppet's acceptance setup, which is causing the variable to not be set.

joshcooper:
I think we need this line https://github.com/puppetlabs/puppetserver/blob/fa34cedb5ba38f86ab156b40c007fe25fae600b1/acceptance/config/beaker/options.rb#L13. It looks like puppet acceptance has gotten away with not having to specify that variable, and instead building the path to it:

```ruby
        puppetserver_config = "#{master['puppetserver-confdir']}/puppetserver.conf"
        on master, "cp #{scriptdir}/puppetserver.conf.bak #{puppetserver_config}"
```

joshcooper:
Superseded by https://github.com/puppetlabs/puppet/pull/5803

PR #5800, opened 2017-04-12
(PUP-7258) Use native JSON parser and generator

joshcooper:
Commit 30e0f1bcd was supposed to remove the json_pure dependency, but
due to a conflict when rebasing wasn't. As a result, puppet was using
json_pure instead of native json in some cases. It also exposed other
places where json_pure was being used:

* Bundler within puppet uses Gemfile, which includes .gemspec
* Bundler for a module that depends on a source checkout of puppet uses .gemspec
* Packaging uses project_data.yaml when building puppet as a gem
* Git acceptance on Solaris (10 and 11) install json_pure to satisfy
  puppet's dependencies.

All traces of json_pure have been removed so that the native version
already present in ruby 1.9.3 and up is used.

PR #5804, opened 2017-04-17
(PUP-7258) Use native JSON parser and generator

joshcooper:
Commit 30e0f1bcd was supposed to remove the json_pure dependency, but
due to a conflict when rebasing wasn't. As a result, puppet was using
json_pure instead of native json in some cases. It also exposed other
places where json_pure was being used:

* Bundler within puppet uses Gemfile, which includes .gemspec
* Bundler for a module that depends on a source checkout of puppet uses .gemspec
* Packaging uses project_data.yaml when building puppet as a gem
* Git acceptance on Solaris (10 and 11) install json_pure to satisfy
  puppet's dependencies.

All traces of json_pure have been removed so that the native version
already present in ruby 1.9.3 and up is used.

These commits also restore the benchmark profiling rake tasks and adds a benchmark for comparing JSON and PSON serialization using a sample PE agent catalog.


joshcooper:
Supersedes PR #5800

PR #5805, opened 2017-04-18
Merge stable to master

hlindberg:


hlindberg:
The failure is due to the POT updating command as it commits with a non compliant message.

joshcooper:
This PR is also failing on ruby 1.9.3 due to net/ssh, thoughts @Iristyle?

```
Gem::InstallError: net-ssh requires Ruby version >= 2.0.
An error occurred while installing net-ssh (4.1.0), and Bundler cannot continue.
Make sure that `gem install net-ssh -v '4.1.0'` succeeds before bundling.
The command "eval bundle install --without development extra " failed 3 times.
```

joshcooper:
POT file issue should be fixed in https://github.com/puppetlabs/ci-job-configs/pull/2742, but won't solve files that have already been updated in puppet#stable.

joshcooper:
Looks like @thallgren merged stable->master in 9ce6842, closing this.

PR #5810, opened 2017-04-19
(PUP-7070) Remove stacktrace from HTTP error responses

joshcooper:
Previously, our REST error responses contained a `stacktrace` element,
and in 4.x this was changed to a deprecation warning so that the server
did not leak sensitive information to the caller.

This commit removes the `stacktrace` element from the response, and
related schema and API docs. Note the `stacktrace` element was optional
according to our schema, so this should not be backwards incompatible.

### DIFF
--- a/spec/unit/network/http/handler_spec.rb
+++ b/spec/unit/network/http/handler_spec.rb
@@ -95,9 +95,6 @@ describe Puppet::Network::HTTP::Handler do
       expect(res[:content_type_header]).to eq("application/json")
       expect(res_body["issue_kind"]).to eq(Puppet::Network::HTTP::Issues::RUNTIME_ERROR.to_s)
       expect(res_body["message"]).to eq("Server Error: the sky is falling!")
-      expect(res_body["stacktrace"].is_a?(Array) && !res_body["stacktrace"].empty?).to be_truthy
-      expect(res_body["stacktrace"][0]).to match(/The 'stacktrace' property is deprecated/)
-      expect(res_body["stacktrace"] & original_stacktrace).to be_empty
       expect(res[:status]).to eq(500)
     end
 
--- a/api/docs/http_api_index.md
+++ b/api/docs/http_api_index.md
@@ -134,8 +134,6 @@ error responses will uniformly be a JSON object with the following properties:
 
 * `message`: (`String`) A human readable message explaining the error.
 * `issue_kind`: (`String`) A unique label to identify the error class.
-* `stacktrace` (only for 5xx errors): (`Array<String>`) A stacktrace to where
-  the error occurred.
 
 A [JSON schema for the error objects](../schemas/error.json) is also available.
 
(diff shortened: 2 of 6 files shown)


spec/unit/network/http/handler_spec.rb
```diff
@@ -95,9 +95,6 @@ def respond(text)
       expect(res[:content_type_header]).to eq("application/json")
       expect(res_body["issue_kind"]).to eq(Puppet::Network::HTTP::Issues::RUNTIME_ERROR.to_s)
       expect(res_body["message"]).to eq("Server Error: the sky is falling!")
-      expect(res_body["stacktrace"].is_a?(Array) && !res_body["stacktrace"].empty?).to be_truthy
```
MosesMendoza: @joshcooper thoughts on is it worth keeping a negative test around to assert we're not including a stacktrace?

joshcooper: I assumed that the json schema validation would fail, but I didn't actually try that. Lemme check that.

joshcooper: I verified the schema validates correctly (by adding back 'stacktrace' to the error object):

```
  1) Puppet::Network::HTTP::Error Puppet::Network::HTTP::Error::HTTPError should serialize to JSON that matches the error schema
     Failure/Error: JSON::Validator.validate!(@schema, json)

     JSON::Schema::ValidationError:
       The property '#/' contains additional properties ["stacktrace"] outside of the schema when none are allowed in schema c296f061-a030-5759-ab86-12b403e61ef0#
     # ./.bundle/ruby/2.3.0/gems/json-schema-2.1.1/lib/json-schema/validator.rb:62:in `validation_error'
     # ./.bundle/ruby/2.3.0/gems/json-schema-2.1.1/lib/json-schema/attributes/additionalproperties.rb:29:in `validate'
     # ./.bundle/ruby/2.3.0/gems/json-schema-2.1.1/lib/json-schema/validator.rb:95:in `block in validate'
     # ./.bundle/ruby/2.3.0/gems/json-schema-2.1.1/lib/json-schema/validator.rb:93:in `each'
     # ./.bundle/ruby/2.3.0/gems/json-schema-2.1.1/lib/json-schema/validator.rb:93:in `validate'
     # ./.bundle/ruby/2.3.0/gems/json-schema-2.1.1/lib/json-schema/schema.rb:38:in `validate'
     # ./.bundle/ruby/2.3.0/gems/json-schema-2.1.1/lib/json-schema/validator.rb:213:in `validate'
     # ./.bundle/ruby/2.3.0/gems/json-schema-2.1.1/lib/json-schema/validator.rb:374:in `validate!'
     # ./spec/lib/matchers/json.rb:116:in `matches?'
     # ./spec/unit/network/http/error_spec.rb:17:in `block (3 levels) in <top (required)>'

Finished in 8.24 seconds (files took 1.06 seconds to load)
```

MosesMendoza: 👍

PR #5811, opened 2017-04-19
DO NOT MERGE *** (PUP-7357) Verfiy agent can parse a JSON catalog

kurtwall:
Now that JSON is preferred to PSON (PUP-7259), verify that that the agent can parse a JSON catalog. This is the first of several tests to against PUP-7357. Note the **Blocked** label -- I would especially appreciate review/suggestions.


kurtwall:
Tested on ubuntu1604-64, osx1012-64, solaris11-64, windows2012r2-64, centos7-64a, arista4-32a-sles12-64.

acceptance/tests/agent/agent_parses_json_catalog.rb
```diff
@@ -0,0 +1,31 @@
...
+
+step "Agent parses a JSON catalog" do
+  agents.each do |agent|
+    # Is it there?
+    json_catalog = File.join(agent.puppet['client_datadir'], 'catalog',
+                             "#{agent.puppet['certname']}.json")
+    on(agent, "[[ -f #{json_catalog} ]]", {:acceptable_exit_codes => [0]})
+
+    # Can we read it?
+    rc = on(agent, "cat #{json_catalog}", {:acceptable_exit_codes => [0]})
```
branan: This is going to dump the entire catalog to the beaker log, right? Should we redirect to `/dev/null` to keep the log cleaner, or is showing it to the user intentional? If intentional, can we add a comment to that effect?

branan:
Instead of pushing the entire JSON catalog across the wire, and making the baker log even harder to read with that stdout data, could we instead do this entirely agent-side? Something like

    cat #{catalog} | ruby -rjson -e 'JSON.parse(STDIN.read)'

This will exit 1 on invalid JSON, and keep everything local to the SUT

joshcooper:
I think there's a beaker or puppet helper to run a ruby command, since ruby isn't in the PATH, something like "cat #{catalog} | #{ruby_command(host)} -rjson -e ..."

kurtwall:
@branan @joshcooper How's this?

PR #5813, opened 2017-04-20
(PUP-6123) remove resource_types endpoint

rlinehan:
The `resource_types` endpoint was deprecated in Puppet 4, with the intention of
removing it for Puppet 5. This PR removes the indirection, termini, face,
application, code to transform to/from JSON, tests, docs, schema, and man page.


rlinehan:
ping @camlow325 @adrienthebo @jpinsonault for review.

rlinehan:
I ran acceptance tests on this via the "ad-hoc" jenkins pipeline and it was all green.

api/schemas/resource_type.json
```diff
@@ -1,43 +0,0 @@
-{
```
rlinehan: Oh wait, should I not have removed this? In puppet server it looks like we use this in our catalog schema, but here it looks like the [catalog.json](https://github.com/puppetlabs/puppet/blob/master/api/schemas/catalog.json#L33) just has all the same stuff in it.

camlow325: Probably okay since none of the other Puppet API schemas refer back to this file directly, it seems?

lib/puppet/resource/type.rb
```diff
@@ -64,49 +59,6 @@ class Puppet::Resource::Type
     define_method("#{t}?") { self.type == t }
   end
 
-  require 'puppet/indirector'
-  extend Puppet::Indirector
-  indirects :resource_type, :terminus_class => :parser
-
-  def self.from_data_hash(data)
```
camlow325: @joshcooper had said it was okay to remove these because they would only be called in the context of the resource_types indirection, right (and not in the context of serialization for any other indirection, like for a catalog response)?

rlinehan: Yeah - I should probably expand my commit message about these, but they're only used by the indirector or if you've included `Puppet::Network::FormatSupport`.

MosesMendoza: Right, you get `Puppet::Network::FormatSupport` from extending `Puppet::Indirector` so yeah strictly speaking I don't think we'll be able to call these methods anymore for serialization because we've stopped indirecting. My fear though is that they're being used outside of that for some other serialization, possibly externally? We never deprecated from_/to_data_hash for `puppet/resource/type` and they're not marked private. Maybe I'm being too cautious but it makes me nervous:)

MosesMendoza:
👍

PR #5820, opened 2017-04-25
(PUP-25) Ensure that resource defaults are evaluated before collections

thallgren:
This commit moves the evaluation of resource defaults so that it happens
before the evaluation of generators and collections. In addition, the
`add_defaults` method will now also set a tag if the name of the
parameter that receives a default value is `:tag`


hlindberg:
Ping @joshcooper - do you see any problems with this?
In general we have circular dependencies between collections/defaults/overrides which were resolved by "the order in which the compiler happened to have implemented them in", this changes the order. It probably does not affect anything, since collections would need to use override to set values, and they would then simply override the defaults. There may be a subtle difference though - hence my question for a second opinion.

hlindberg:
I think my concern was unfounded. Merging.

PR #5822, opened 2017-04-25
(PUP-7470) Use a feature check instead of inspecting `Gem.loaded_specs['gettext-setup']`

joshuaspence:
`Gem.loaded_specs['gettext_setup']` doesn't work with Bundler's standalone binstubs, which intentionally avoids a dependency on RubyGems and, instead, manipulates Ruby's `$LOAD_PATH` directly. As such, Puppet 4 doesn't currently work using Bundler's standalone binstubs (Puppet 3, however, works fine):

```
/etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/lib/puppet.rb:76:in `singleton class': undefined method `version' for nil:NilClass (NoMethodError)
    from /etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/lib/puppet.rb:55:in `<module:Puppet>'
    from /etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/lib/puppet.rb:49:in `<top (required)>'
    from /usr/local/rbenv/versions/2.1.5/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /usr/local/rbenv/versions/2.1.5/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/lib/puppet/util/command_line.rb:12:in `<top (required)>'
    from /usr/local/rbenv/versions/2.1.5/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /usr/local/rbenv/versions/2.1.5/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/bin/puppet:4:in `<top (required)>'
    from /etc/puppet/bin/puppet:14:in `load'
    from /etc/puppet/bin/puppet:14:in `<main>'
```

See some relevant context in bundler/bundler#5601.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -73,7 +73,7 @@ module Puppet
       end
 
       if locale_path
-        if Gem.loaded_specs['gettext-setup'].version < Gem::Version.new('0.8')
+        if GettextSetup.method(:initialize).parameters.count == 1
           # Will load translations from PO files only
           GettextSetup.initialize(locale_path)
         else


Gemfile
```diff
@@ -28,7 +28,7 @@ gem "hiera", *location_for(ENV['HIERA_LOCATION'] || ['>= 3.2.1', '< 4'])
 # PUP-7115 - return to a gem dependency in Puppet 5
 # gem "semantic_puppet", *location_for(ENV['SEMANTIC_PUPPET_LOCATION'] || ['>= 0.1.3', '< 2'])
 # i18n support (gettext-setup and dependencies)
-gem 'gettext-setup', '>= 0.10', '< 1.0', :require => false
+gem 'gettext-setup', '>= 0.10', '< 1.0', :github => 'joshuaspence/gettext-setup-gem', :branch => 'version', :require => false
```
joshuaspence: This is needed so that I can actually test this change, but ideally puppetlabs/gettext-setup-gem#40 would be merged and a new version released, in which case I can simply bump the minimum version constraint here.

lib/puppet.rb
```diff
@@ -74,7 +74,7 @@ class << self
       end
 
       if locale_path
-        if Gem.loaded_specs['gettext-setup'].version < Gem::Version.new('0.8')
+        if Gem::Version.new(GettextSetup::VERSION) < Gem::Version.new('0.8')
```
smcclellan: @Magisus — After reading the reason for this being added in the first place, can this conditional be removed altogether? The Gemfile specifies version >= 0.10. Is there still a situation where an earlier version of gettext-setup would run this code?

Magisus: Yes, in the context of Puppet within puppet-server. They vendor their own gems, and the older versions of it (e.g. less than 2.7.2, I think) have something like gettext-setup 0.6.

smcclellan: `GettextSetup::VERSION` won't exist in version 0.8. I wonder if there are any other ways to determine this. Any ideas, given the problem above with `loaded_specs`? Perhaps we could get the version number from the load path, though that sounds messy.

joshuaspence: Perhaps we can check if `GettextSetup` has the _functionality_ that is required, rather than checking for the version number.

smcclellan: One idea for achieving that:

```
GettextSetup.method(:initialize).parameters.count == 1
```

Magisus: So long as that correctly detects parameters with defaults, works for me.

joshuaspence: Thanks, that works

lib/puppet.rb
```diff
@@ -74,7 +74,7 @@ class << self
       end
 
       if locale_path
-        if Gem.loaded_specs['gettext-setup'].version < Gem::Version.new('0.8')
+        if GettextSetup.method(:initialize).parameters.count > 1
```
Magisus: This might need to be the opposite? As in, if it's 2 (or greater than 1), use the version that passes 2 arguments, otherwise just pass one.

joshuaspence: Oops, yep

lib/puppet.rb
```diff
@@ -74,7 +74,7 @@ class << self
       end
 
       if locale_path
-        if Gem.loaded_specs['gettext-setup'].version < Gem::Version.new('0.8')
+        if GettextSetup.method(:initialize).parameters.count < 2
```
smcclellan: I'm confused why this wouldn't be just `== 1`. It's called with one argument below. Perhaps a nit, but it just reads weird to have 2 here and 1 argument below.

joshuaspence: Yeah, fair enough. I guess `==1` is clearer.

smcclellan: And if this equals 0, the call on the next line will fail anyway.

smcclellan:
Thanks for discovering/working through this, @joshuaspence!

smcclellan:
Looks like the checks are failing because the commit doesn't start with `(PUP-###)` or `(maint)`. Filing a ticket [here](https://tickets.puppetlabs.com/browse/PUP) would be helpful, if you haven't already.

joshuaspence:
Filed PUP-7470.

Magisus:
@puppetlabs/puppet-agent-maintainers would you mind taking a look at this to see if it seems reasonable?

jarro2783:
Which version of Puppet would this be merged into? I would be interested in having this in Puppet 4.

Magisus:
This could theoretically go into stable. It's not a breaking change. If we want to do that, this commit will need a rebase and the PR will have to be retargeted.

lib/puppet.rb
```diff
@@ -74,7 +74,7 @@ class << self
       end
 
       if locale_path
-        if Gem.loaded_specs['gettext-setup'].version < Gem::Version.new('0.8')
+        if GettextSetup.method(:initialize).parameters.count == 1
```
MosesMendoza: If we're searching for the number of parameters, we can access that with `:arity`
ie
```
Foo.method(:bar).arity
=> 2
```

smcclellan: That won't work here because both arguments are optional. `GettextSetup.method(:initialize).arity` returns `-1`.

MosesMendoza: Ah cool, thanks

MosesMendoza:
👍

joshuaspence:
I have rebased this PR onto the `stable` branch.

MosesMendoza:
Hi @joshuaspence  thanks for the retarget. going to close and reopen to get a clean travis run.

Iristyle:
👍  to this... looks like the right approach.

joshuaspence:
bump

MosesMendoza:
Hi @joshuaspence  - thanks for the ping on this. I haven't forgotten about it, just working some things out on our end before merging stuff into stable.

joshuaspence:
Thanks for merging this. Is there a scheduled 4.x release anytime soon?

PR #5824, opened 2017-04-26
(PUP-7357) Verify agent can parse a JSON catalog

kurtwall:
Now that JSON is preferred to PSON (PUP-7259), verify that that the agent can parse a JSON catalog. This is the first of several tests for PUP-7357.

### DIFF
--- a/acceptance/tests/agent/agent_parses_json_catalog.rb
+++ b/acceptance/tests/agent/agent_parses_json_catalog.rb
@@ -1,0 +1,23 @@
+test_name "C99978: Agent parses a JSON catalog"
+tag 'risk:medium'
+
+require 'json'
+
+step "Refresh the catalog" do
+  on(agents, puppet("agent --test --server #{master}"))
+end
+
+step "Agent parses a JSON catalog" do
+  agents.each do |agent|
+    # Is it there and can we read it?
+    json_catalog = File.join(agent.puppet['client_datadir'], 'catalog',
+                             "#{agent.puppet['certname']}.json")
+    on(agent, "[[ -r #{json_catalog} ]]", {:acceptable_exit_codes => [0]})
+
+    # The catalog file should be parseable JSON
+    on(agent, "cat #{json_catalog} | ruby -rjson -e 'JSON.parse(STDIN.read)'")
+
+    # Can the agent parse it as JSON?
+    on(agent, puppet("catalog find --terminus json --server #{master} > /dev/null"))
+  end
+end


kurtwall:
This replaces the previous PR that @branan and @joshcooper already reviewed.

acceptance/tests/agent/agent_parses_json_catalog.rb
```diff
@@ -0,0 +1,20 @@
...
+
+step "Agent parses a JSON catalog" do
+  agents.each do |agent|
+    # Is it there and can we read it?
+    json_catalog = File.join(agent.puppet['client_datadir'], 'catalog',
+                             "#{agent.puppet['certname']}.json")
+    on(agent, "[[ -r #{json_catalog} ]]", {:acceptable_exit_codes => [0]})
+
+    # Can the agent parse it as JSON?
+    on(agent, puppet("catalog find --terminus json --server #{master} > /dev/null"))
```
johnduarte: Is this enough to ensure that a JSON catalog was retrieved? I am fine if the answer is, yes. I am just wondering if we should validate that an expected catalog element is contained in the JSON payload.

For example
```
on(agent, puppet("catalog find --terminus json --server #{master}")) do |result|
  require 'json'
  json = JSON.parse(result.stdout)
  assert_equal(true, json.has_key? 'name')
  assert_equal(#{agent}, json['name'])
end
```

kurtwall: Fixed.

acceptance/tests/agent/agent_parses_json_catalog.rb
```diff
@@ -0,0 +1,23 @@
...
+
+step "Agent parses a JSON catalog" do
+  agents.each do |agent|
+    # Is it there and can we read it?
+    json_catalog = File.join(agent.puppet['client_datadir'], 'catalog',
+                             "#{agent.puppet['certname']}.json")
+    on(agent, "[[ -r #{json_catalog} ]]", {:acceptable_exit_codes => [0]})
+
+    # The catalog file should be parseable JSON
+    on(agent, "cat #{json_catalog} | ruby -rjson -e 'JSON.parse(STDIN.read)'")
```
johnduarte: Does this make the call on line 15 redundant?

Using ruby on the SUT this way assumes it is on the PATH on the SUT which is not always the case, especially on Windows.

kurtwall: Line 15 just says we have read permission on the file. Line 18 says we can parse it as JSON. Didn't want to try to read something before verifying it is present and readable.

PR #5825, opened 2017-04-26
(PUP-7431) Upgrade vendored semantic_puppet gem from 0.1.4 to 1.0.0

thallgren:
This commit replaces the code beneath lib/puppet/vendor/semantic_puppet
to reflect the source of version 1.0.0.


lib/puppet/vendor/semantic_puppet/lib/semantic_puppet.rb
```diff
@@ -1,15 +1,7 @@
...
-  def _(msg)
-    msg
-  end
-end
+require 'gettext-setup'
 
 module SemanticPuppet
-  if defined?(GettextSetup)
-    GettextSetup.initialize(File.absolute_path('semantic_puppet/locales', File.dirname(__FILE__)))
-  end
```
MosesMendoza: I think we might need to keep this feature gate? It was introduced separately in https://github.com/puppetlabs/puppet/commit/0016250c

thallgren: If so, then we need to change the original semantic_puppet gem and release a 1.0.1 version of it. But that, in turn, gets awkward since the semantic_puppet gem has a hard dependency on gettext. Making that a loose dependency is complicated since semantic_puppet has no knowledge of features.

thallgren: One reason to avoid patching the vendored code is that we want to remove it in [PUP-7115](https://tickets.puppetlabs.com/browse/PUP-7115).

@MosesMendoza, do you feel that we need an 1.0.1 release of semantic_puppet to resolve this (one without a hard dependency on gettext) or should it be resolved in some other way?

hlindberg: But, just asking for `defined?(GettextSetup)` does not create a hard dependency - does it? It just checks if that name is defined...

thallgren: Yes, but that change, along with dropping the current hard dependency on the gettext gem must be made in the the semantic_puppet gem. It should not be done here. This is just a copy.

MosesMendoza: Strictly speaking, we don't need it in the gem, because the gem can/does express a dependency on gettext, so it should always be satisfied when running as a gem. The reason we need the gettext check specifically here in the vendored case is because this puppet code can be loaded from a Puppet Server install pre-2.7.2 that does not have gettext available - so both here, and in puppet (https://github.com/puppetlabs/puppet/blob/master/lib/puppet.rb#L28-L37) we do this check.  It was definitely not ideal to add it into the vendored lib as it represents a divergence from the upstream library but it seemed the least of all evils. We want to find a better solution, and have that ticketed. These issues: https://tickets.puppetlabs.com/browse/PA-1018, https://tickets.puppetlabs.com/browse/PUP-7114, and https://tickets.puppetlabs.com/browse/PUP-7116 have more context.

My recommendation here, as unfortunate as it is, is to just patch this vendored copy to do the previous behavior prior to this PR - ie check for gettext and only load if necessary. I don't think the same changes are needed in the gem.

thallgren: I patched the copy (reluctantly :-) )

lib/puppet/vendor/semantic_puppet/lib/semantic_puppet.rb
```diff
@@ -1,14 +1,8 @@
-begin
-  require 'gettext-setup'
-rescue LoadError
-  def _(msg)
-    msg
-  end
-end
+require 'gettext-setup'
```
MosesMendoza: Won't this still issue a load error though if the gem isn't available?

thallgren: It sure will. I changed the last commit to not require gettext-setup and instead rely on the constant {{Puppet::GETTEXT_AVAILABLE}}. When doing this, I ran some tests. Turns out that puppet cannot run without the gem anyway so I created ticket https://tickets.puppetlabs.com/browse/PUP-7480.

Magisus: I have a PR up for that now: https://github.com/puppetlabs/puppet/pull/5833

MosesMendoza: Ah and this works because Puppet stubs the methods in question

PR #5826, opened 2017-04-26
(PUP-7473) remove dead code in Puppet::Type::instances

MosesMendoza:
This block removed in this commit iterates over all of the properties of the
type, checks if the given provider supports the property, and returns an array
of all the supported property names.  Except we don't do anything with the
result.  Before commit 53241a91, which was a maint commit to remove superfluous
variables we never used, the result of this block was assigned to the variable
`all_properties` - which was never used. The commit removed the variable, but
not the associated block.  I believe the intention was to filter out properties
that weren't supported by the provider before we called `newattr` on them,
except `newattr` already does this filtering, and provides helpful debug output
that a provider doesn't support a property so we're not managing it. Hence we
remove it here to avoid useless confusion/object iteration/creation.

### DIFF
--- a/lib/puppet/type.rb
+++ b/lib/puppet/type.rb
@@ -1161,12 +1161,6 @@ class Type
     # Put the default provider first, then the rest of the suitable providers.
     provider_instances = {}
     providers_by_source.collect do |provider|
-      self.properties.find_all do |property|
-        provider.supports_parameter?(property)
-      end.collect do |property|
-        property.name
-      end
-
       provider.instances.collect do |instance|
         # We always want to use the "first" provider instance we find, unless the resource
         # is already managed and has a different provider set


Iristyle:
6 year old cruft!

There are some interesting notes in the commit at:

https://github.com/MosesMendoza/puppet/commit/d4e6c268aecabdbf36ece126f960ce2384576fcc

```
(#7624) Manually fetch all properties in instances.
When we removed the `:audit => :all` flag on creation of an instance from a
Puppet type, we stopped fetching all the properties.  This had flow-on
consequences that were visible from the outside; while some places did their
own work to ensure that properties were fetched, others didn't.

We now open-code the loop that creates and fetches those properties, to ensure
that we have the same data without going through the :audit machinery.

This resolves the excessive logging, and also eliminates the behavioural
change that we introduced in the previous commit.

Reviewed-By: Daniel Pittman <daniel@puppetlabs.com>
```

Original redmine is at https://projects.puppetlabs.com/issues/7624

It seems to imply that even though the results were never used, that the code was used to ensure things were loaded properly.

Iristyle:
Might want to double check the purge bug report in 7624 - if for some reason we still need the code you've deleted to fix that bug, should probably leave a comment for future maintainers instead of deleting it.

MosesMendoza:
I made these changes and replicated the setup described in that bug, but the issue is no longer reproducible. In the original diff from https://projects.puppetlabs.com/issues/7624,
```
+      all_properties = self.properties.find_all do |property|
+        provider.supports_parameter?(property)
+      end.collect do |property|
+        property.name
+      end
+
       provider.instances.collect do |instance|
         # We always want to use the "first" provider instance we find, unless the resource
         # is already managed and has a different provider set
@@ -886,7 +892,10 @@ class Type
         end
         provider_instances[instance.name] = instance
 
-        new(:name => instance.name, :provider => instance, :audit => :all)
+        result = new(:name => instance.name, :provider => instance)
+        properties.each { |name| result.newattr(name) }
+        result
```

I think the key is that we removed the line that specified the `:audit` parameter which had a side-effect of creating property instances, and replaced it with the explicitly calling `newattr` with all of the properties. Even in that original patch, as written the first block (the one removed in this PR) wasn't necessary to solve the bug, I think. I think the intention was to use the `all_properties` result from that block to supply the list of properties we would iterate over and call `newattr` with (ie, `all_properties.each { |name| result.newattr(name) }` , but instead in an oversight we used `properties.each {|name| result.newattr(name) }`. This was masked because there's a `:properties` accessor.

PR #5827, opened 2017-04-26
(PUP-7357) Verify agent can parse a JSON catalog

kurtwall:
Confirm that the agent can parse a JSON catalog and that the catalog request prefers JSON to PSON.


kurtwall:
Tested and passing on ubuntu1604-64a, fedora24-64a, windows2012r2-64a, and osx1012-64a.

PR #5830, opened 2017-04-27
(maint) Update locale detection to use gettext-setup wrapper

Magisus:
This commit switches to using `GettextSetup.candidate_locales` instead
of calling out to the Locale gem directly to detect the locale.
Internally gettext-setup uses the locale gem anyway, but the wrapper
methods provide additional sorting and fallback features. This also
allows us to remove our dependency on Locale gem.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -27,7 +27,6 @@ require 'puppet/external/pson/pure'
 # supplied by gettext-setup. Can be removed in Puppet 5. See PUP-7116.
 begin
   require 'gettext-setup'
-  require 'locale'
   Puppet::GETTEXT_AVAILABLE = true
 rescue LoadError
   def _(msg)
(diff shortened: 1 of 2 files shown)


lib/puppet.rb
```diff
@@ -80,7 +79,7 @@ class << self
         else
           GettextSetup.initialize(locale_path, :file_format => :mo)
         end
-        FastGettext.locale = GettextSetup.negotiate_locale(Locale.current.language)
+        GettextSetup.negotiate_locale!(GettextSetup.candidate_locales)
```
Iristyle: I assume `FastGettext` automatically picks up on `Gettextsetup` configuration?

I don't know how the various libs plug together.

Magisus: The method with the `!` just calls the non-mutating one and assigns the result to FastGettext the way we were doing it previously. The main difference for us is using `candidate_locales` instead of `Locale.current`.

MosesMendoza:
Reverted at 1628c6ecdcdf18e72768f47765e3cf123eb48944. Introduced failures in acceptance. 
/code

PR #5828, opened 2017-04-26
(Spike) Changes / new tests for #5799

Iristyle:

PR #5831, opened 2017-04-27
(Spike) track canonical resource username before encoding fixup in nameservice

MosesMendoza:
includes #5828

### DIFF
--- a/lib/puppet/application/resource.rb
+++ b/lib/puppet/application/resource.rb
@@ -138,14 +138,16 @@ Copyright (c) 2011 Puppet Labs, LLC Licensed under the Apache 2.0 License
       resources = find_or_save_resources(type, name, params)
 
       if options[:to_yaml]
-        text = resources.
-          map { |resource| resource.prune_parameters(:parameters_to_include => @extra_params).to_hierayaml }.
-          join("\n")
+        text = resources.map do |resource|
+          r = resource.prune_parameters(:parameters_to_include => @extra_params).to_hierayaml
+          r.encoding == Encoding.default_external? ? r : r.force_encoding(Encoding.default_external).scrub!
+        end.join("\n")
         text.prepend("#{type.downcase}:\n")
       else
-        text = resources.
-          map { |resource| resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest }.
-          join("\n")
+        text = resources.map do |resource|
+          r = resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest
+          r.encoding == Encoding.default_external ? r : r.force_encoding(Encoding.default_external).scrub!
+         end.join("\n")
       end
 
       options[:edit] ?
--- a/lib/puppet/provider/nameservice.rb
+++ b/lib/puppet/provider/nameservice.rb
@@ -25,9 +25,12 @@ class Puppet::Provider::NameService < Puppet::Provider
     def instances
       objects = []
       listbyname do |name|
-        objects << new(:name => name, :ensure => :present)
+        canonical_name = name.dup
+        if !name.valid_encoding?
+          name.scrub!
+        end
+        objects << new(:name => name, :ensure => :present, :canonical_name => canonical_name)
       end
-
       objects
     end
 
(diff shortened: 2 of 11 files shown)


lib/puppet/application/resource.rb
```diff
@@ -138,14 +138,16 @@ def main
...
+        end.join("\n")
         text.prepend("#{type.downcase}:\n")
       else
-        text = resources.
-          map { |resource| resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest }.
-          join("\n")
+        text = resources.map do |resource|
+          r = resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest
+          r.encoding == Encoding.default_external ? r : r.force_encoding(Encoding.default_external).scrub!
+         end.join("\n")
```
MosesMendoza: Need tests for these changes

lib/puppet/application/resource.rb
```diff
@@ -138,14 +138,16 @@ def main
...
+          r.encoding == Encoding.default_external? ? r : r.force_encoding(Encoding.default_external).scrub!
+        end.join("\n")
         text.prepend("#{type.downcase}:\n")
       else
-        text = resources.
-          map { |resource| resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest }.
-          join("\n")
+        text = resources.map do |resource|
+          r = resource.prune_parameters(:parameters_to_include => @extra_params).to_manifest
+          r.encoding == Encoding.default_external ? r : r.force_encoding(Encoding.default_external).scrub!
```
MosesMendoza: Can't use `scrub!` as it's not in ruby 1.9.3

Appears we can do this to replicate, even if the string begins in UTF-8
```
'string'.force_encoding(Encoding::UTF_8, Encoding::BINARY, :undef => :replace, :invalid => :replace)
```

lib/puppet/provider/nameservice.rb
```diff
@@ -276,6 +283,13 @@ def initialize(resource)
     super
     @custom_environment = {}
     @objectinfo = nil
+    if resource.is_a?(Hash) && !resource[:canonical_name].nil?
+      # we're instantiating from a hash, ie self.instances
+      @canonical_name = resource[:canonical_name]
+    else
+      # we're instantiating from a resource
+      @canonical_name = resource[:name]
```
MosesMendoza: In this case, we've created a new resource object directly and might be testing that it exists, for example the puppet service user. in this case, we haven't defined a canonical name and its ok to just use name because we're not gathering this info from Etc and haven't done any scrubbing of invalid bytes

lib/puppet/provider/nameservice.rb
```diff
@@ -276,6 +283,13 @@ def initialize(resource)
     super
     @custom_environment = {}
     @objectinfo = nil
+    if resource.is_a?(Hash) && !resource[:canonical_name].nil?
+      # we're instantiating from a hash, ie self.instances
+      @canonical_name = resource[:canonical_name]
```
MosesMendoza: In this case we've queried the system Etc value to get a struct of user data (and subsequently it has been scrubbed of invalid data) and transformed that user data into a hash, which we're going to use to construct this provider instance.

lib/puppet/provider/nameservice.rb
```diff
@@ -25,9 +25,12 @@ def initvars
     def instances
       objects = []
       listbyname do |name|
-        objects << new(:name => name, :ensure => :present)
+        canonical_name = name.dup
+        if !name.valid_encoding?
+          name.scrub!
+        end
+        objects << new(:name => name, :ensure => :present, :canonical_name => canonical_name)
```
MosesMendoza: This approach I think is not the one we're going to go with.

in discussion with @Iristyle I think we concluded extending the Etc::Passwd and Etc::Group structs to include fields that hold this "canonical" data in addition to the "scrubbed" data is a cleaner and more complete/correct solution

MosesMendoza:
got what we needed from this spike - going to push new changes to original PR #5799

PR #5832, opened 2017-04-27
(PUP-7475) Fix problem with multiple backslashes in regexp literal

thallgren:
Before this commit, it was not possible to declare a regexp that ended
with an escaped backslash because there was no logic to determine if
the backslash was preceded by a backslash.

This commit adds code to ensure that only an uneven number of subsequent
backslashes causes an escape of the token that follows.


hlindberg:
Fails on Ruby 1.9.3

PR #5833, opened 2017-04-28
(PUP-7480) Stub gettext pluralization function

Magisus:
Puppet should be able to run without the gettext-setup gem. We had
previously stubbed the basic `_` translation function to avoid errors
when the gem is missing, but later string externalization commits also
made use of the `n_` function for pluralization handling. This commit
adds a stub for that method.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -33,6 +33,14 @@ rescue LoadError
   def _(msg)
     msg
   end
+
+  def n_(*args, &block)
+    # assume two string args (singular and plural English form) and the count
+    # to pluralize on
+    plural = args[2] == 1 ? args[0] : args[1]
+    # if a block is passed, prefer that over the string selection above
+    block ? block.call : plural
+  end
   Puppet::GETTEXT_AVAILABLE = false
 end
 
--- a/lib/puppet/pops/loader/puppet_resource_type_impl_instantiator.rb
+++ b/lib/puppet/pops/loader/puppet_resource_type_impl_instantiator.rb
@@ -50,7 +50,7 @@ class PuppetResourceTypeImplInstantiator
     end
 
     unless statements.size == 1
-      raise ArgumentError, _("The code loaded from %{source_ref} must contain only the creation of resource type '%{typen_name}' - it has additional logic.") % { source_ref: source_ref, type_name: typed_name.name }
+      raise ArgumentError, _("The code loaded from %{source_ref} must contain only the creation of resource type '%{type_name}' - it has additional logic.") % { source_ref: source_ref, type_name: typed_name.name }
     end
 
     closure_scope = Puppet.lookup(:global_scope) { {} }


McdonaldSeanp:
👍

lib/puppet.rb
```diff
@@ -33,6 +33,10 @@
   def _(msg)
     msg
   end
+
+  def n_(msg_singular, msg_plural, count)
+    return count == 1 ? msg_singular : msg_plural
```
MosesMendoza: This works as long as we only use a single plural form.. but strictly speaking, the n_ definition we're using from fast_getext supports more (and a block form, actually): https://github.com/grosser/fast_gettext/blob/master/lib/fast_gettext/translation.rb#L28-L43

It might be safer not to assume the number of args and index on count?

Magisus: Okay that works. Should we handle the block form somehow too?

Magisus: So, actually, I'm unsure what you mean about indexing on count. Count is the bit of the sentence that determines which form to use based on linguistic rules, and is unrelated to how many args there are. In English, there are only two plural forms, ever, which means this method will only ever be called with two args plus the count as long as we're comfortable assuming it will only be called in an English context. The gettext version supports more because they need to be flexible enough to support other "native" languages which may have additional plural forms.

MosesMendoza: Yep, makes sense. When redefining a method I just try to lean as close to honoring the semantics of the original method as possible. Seems like it ends up being less surprising down the road sometimes. but you're right - what I meant by count was nonsensical because it's not like we're going to be needing to use or supplying additional pluralization rules to support multiple forms. So never mind.

Magisus: Yeah it does seem good to do in general. Sigh, gotta love Ruby >.> I'm happy to change the signature and then just keep the functionality as it is now, i.e. grab the first two args of the list and return the right one based on `count`. I'll also add the block.

Magisus:
I'm not really familiar enough with Puppet to know how to trigger the code paths that use `n_()` to make sure this works. Does someone have a suggestion?

thallgren:
@Magisus, when I tested, I went through the following steps:
1. Comment out the dependencies to gettext-setup and locale from the .gemspec and Gemfile
2. `gem uninstall gettext-setup` and `gem uninstall locale`
3. `bundle install`
4. bundle exec rake parallel:spec[22]

With this PR, everything works. Well almost. A [typo sneaked in](/puppetlabs/puppet/blob/master/lib/puppet/pops/loader/puppet_resource_type_impl_instantiator.rb#L53) when you fuxed PUP-7310. _typen_name_ instead of _type_name_ (perhaps you can fix that with a maint PR?). Apparently gettext doesn't detect missing keys but using `String#%` does.

Magisus:
Thanks for catching that, I'll add a maint commit to this.

PR #5834, opened 2017-04-29
(PUP-7484) Ensure that generated resources get defaults assigned

thallgren:
The fix in PUP-25, where the assignment of resource parameter default
values was moved to take place before resource generators was evaluated,
resulted in that no defaults were assigned to the generated resources.

This commit ensures that resources created during the evaluation of
generators will get their parameter defaults assigned.


hlindberg:
Typo in commit mesage "begin/rescure"

hlindberg:
ok to merge once the typo has been fixed

PR #5835, opened 2017-05-01
(PUP-996) Make relationship with empty middle set work

hlindberg:
This makes chains or relationships close a gap in a chain of relationships. 
It used to break the chain.

Thus:
`[File[a], File[b]] -> [ ] ~> [File[aa], File[bb]]`
results in
```
a ~> aa
a ~> bb
b ~> aa
b ~> bb
```
This works for both explicitly empty sets, and for collections returning empty sets. Note that the gap is closed with the rightmost operator with a non-empty right set.

### DIFF
--- a/lib/puppet/parser/relationship.rb
+++ b/lib/puppet/parser/relationship.rb
@@ -31,13 +31,6 @@ class Puppet::Parser::Relationship
   end
 
   def mk_relationship(source, target, catalog)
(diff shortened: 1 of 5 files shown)


hlindberg:
Note - this is uglier than it would have to be if the Catalog understood the Resource and Class data types directly.

PR #5837, opened 2017-05-02
(PUP-7485) Ensure that global functions and types are publicly available

thallgren:
Prior to this PR, a global function or type that was not declared so that it was available to the auto-loader, would not be reachable from a module. This PR changes this so that such a function becomes known to the public environment loader.

### DIFF
--- a/lib/puppet/pops/loaders.rb
+++ b/lib/puppet/pops/loaders.rb
@@ -170,7 +170,7 @@ class Loaders
   # @raise [Puppet::ParseError] if no loader can be found
   # @api private
   def find_loader(module_name)
-    if module_name.nil? || module_name == ''
+    if module_name.nil? || EMPTY_STRING == module_name
       # TODO : Later when fdefinition can be private, a decision is needed regarding what that means.
       #        A private environment loader could be used for logic outside of modules, then only that logic
       #        would see the definition.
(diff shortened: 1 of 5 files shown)


lib/puppet/pops/loaders.rb
```diff
@@ -170,15 +170,9 @@ def [](loader_name)
...
-    if module_name.nil? || module_name == ''
-      # TODO : Later when fdefinition can be private, a decision is needed regarding what that means.
-      #        A private environment loader could be used for logic outside of modules, then only that logic
-      #        would see the definition.
-      #
-      # Use the private loader, this definition may see the environment's dependencies (currently, all modules)
-      loader = private_environment_loader()
-      raise Puppet::ParseError, 'Internal Error: did not find public loader' if loader.nil?
-      loader
+    if module_name.nil? || EMPTY_STRING == module_name
```
hlindberg: is it better to compare that way (frozen constant first)? Easier to read the other way around.

thallgren: In general, it's better to place a known object first because you then dictate what `#eql?` method that is being used. Placing an unknown first might expose you to slow implementations of that method and sometimes you will even get errors due to bad implementations (we recently had such errors with SemanticPuppet). In this case `module_name` is whatever the user decides to send to the method, so who knows what gets called when doing it the other way around.

As for readability, I guess that's a matter of taste. Nowadays I prefer constants first. I think my preference changed when coding Java where sending `#equals` to the constant is a way avoid NPEs.

PR #5838, opened 2017-05-03
(PUP-3552) Reject invalid requests and unsupported content types

joshcooper:
PR #5836 should be merged first.

Previously, puppet returned HTTP 500 if the request was invalid due to missing `Content-Type` or parsing the request body failed, or if the Content-Type was unsupported.

In order to implement proper HTTP content-negotiation, the server should:

1. Respond with HTTP 400 if the Content-Type header is missing or we fail to parse the request body
1. Respond with HTTP 415 if the client sends data, e.g. PUT report, but the server or model doesn't support that format.
1. Respond with HTTP 406 if the client sends an `Accepts` header, but the server or model doesn't support any of the formats.

This commit adds tests for the HTTP 400 and 415 case, and updates the code accordingly. This is important so that the server is future compatible (it will reply with 415 for content-types it doesn't understand). The server already handled the 406 case correctly, so that behavior is unchanged.


lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -25,9 +25,12 @@ def extract_facts_from_request(request)
       # in Network::HTTP::Handler will automagically deserialize the value.
       if text_facts.is_a?(Puppet::Node::Facts)
         facts = text_facts
-      else
+      elsif format == 'pson'
         # We unescape here because the corresponding code in Puppet::Configurer::FactHandler escapes
+        # PSON is deprecated, but continue to accept from older agents
```
joshcooper: The ticket mentions we should issue a deprecation warning if we receive PSON. Given the likelihood of spamming the master's logs when dealing with mixed agent versions, I didn't add a deprecation warning.

MosesMendoza:
For the 406 should we respond with the formats we do support?

joshcooper:
@MosesMendoza good question. The RFC says we should, but it's up to the implementation to decide how to represent that info in the response. I couldn't find any examples in practice, so I didn't implement that part.

lib/puppet/network/http/request.rb
```diff
@@ -19,41 +19,42 @@ def route_into(prefix)
...
       format = Puppet::Network::FormatHandler.mime(header)
-      if format.nil?
-        #TRANSLATORS "mime-type" is a keyword and should not be translated
-        raise _("Client sent a mime-type (%{header}) that doesn't correspond to a format we support") % { header: header }
-      else
-        assert_supported_format(format)
-        return format.name.to_s if format.suitable?
-      end
+
+      return valid_network_format?(format) ? format : nil
```
MosesMendoza: Thoughts on renaming the local `format` variable here to something else, to avoid possible confusion/collision with the `format` method defined later?

lib/puppet/network/http/request.rb
```diff
@@ -19,39 +19,42 @@ def route_into(prefix)
     self.class.new(headers, params, method, path, routing_path.sub(prefix, ''), client_cert, body)
   end
 
-  def format
+  def formatter
```
MosesMendoza: so this returns a `Puppet::Network::Format` instance?

lib/puppet/network/http/api/indirected_routes.rb
```diff
@@ -190,9 +190,21 @@ def accepted_response_formatter_or_json_for(model_class, request)
...
+          _("The request body is invalid: %{message}") % { message: e.message })
+      end
+    end
 
-    format = request.format
-    model_class.convert_from(format, data)
+    #TRANSLATORS "mime-type" is a keyword and should not be translated
+    raise Puppet::Network::HTTP::Error::HTTPUnsupportedMediaTypeError.new(
+      _("Client sent a mime-type (%{header}) that doesn't correspond to a format we support") % { header: request.headers['content-type'] },
+      Puppet::Network::HTTP::Issues::UNSUPPORTED_MEDIA_TYPE)
```
MosesMendoza: This is probably a crazy question, because why would the answer possibly be yes, but is there any other kind of request we'd serve besides `PUT` / `save` where we'd expect/use the request's `Content-Type`? I only ask because...well you know. We do unusual things sometimes.

joshcooper: Good question. The `IndirectedRoutes` class only handles a subset of REST requests, and there are other routes for things that aren't indirected types, e.g. `/puppet/v3/environments`. While it appears we only examine the request body when `PUT`ing an indirected model, e.g. CSR, Report, it's not safe to assume that will always be the case.

If the agent sends an unknown `Content-Type`, then we should handle that more generally in `Request#formatter` as that logic is shared for all routes and all servers (webrick/puppetserver/rack). Then I can leave this check for indirected models that don't support the requested format, e.g. uploading a report in `application/octet-stream`.

lib/puppet/network/http/request.rb
joshcooper: yep, unfortunately puppet uses both format and formatter to refer to a `Format` object, and other times `format` refers to either the mime type or the registered name, e.g. `json`.

lib/puppet/network/http/request.rb
joshcooper: sure, let me update that

lib/puppet/network/http/api/indirected_routes.rb
```diff
@@ -192,7 +192,12 @@ def read_body_into_model(model_class, request)
     data = request.body.to_s
 
     format = request.format
-    model_class.convert_from(format, data)
+    begin
+      model_class.convert_from(format, data)
+    rescue => e
+      raise Puppet::Network::HTTP::Error::HTTPBadRequestError.new(
```
Iristyle: Unknown Content-Type should be a 415 per https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.16

Iristyle: Whether 400 vs 422 for the body has been a point of debate for a while - https://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data - currently 400 seems to be the way to go.

lib/puppet/network/http/api/indirected_routes.rb
```diff
@@ -190,14 +190,21 @@ def accepted_response_formatter_or_json_for(model_class, request)
...
+      begin
+        return model_class.convert_from(formatter.name.to_s, data)
+      rescue => e
+        raise Puppet::Network::HTTP::Error::HTTPBadRequestError.new(
+          _("The request body is invalid: %{message}") % { message: e.message })
+      end
     end
+
+    #TRANSLATORS "mime-type" is a keyword and should not be translated
+    raise Puppet::Network::HTTP::Error::HTTPUnsupportedMediaTypeError.new(
```
Iristyle: Oh look at that, this later commit responds with the 415 - disregard my earlier comment!

lib/puppet/network/http/api/indirected_routes.rb
MosesMendoza: To make sure I'm following - it looks like `Request#formatter` will return `nil` if the agent sends an unknown `Content-Type`. So we expect the caller to handle the `nil` case?

joshcooper: Commit 3b234a8 ensures that `formatter` always returns a valid `Format` or raises.

lib/puppet/network/http/request.rb
```diff
@@ -19,39 +19,48 @@ def route_into(prefix)
...
+      Puppet::Network::HTTP::Issues::UNSUPPORTED_FORMAT)
   end
 
-  def assert_supported_format(format)
-    if format.name == :yaml || format.name == :b64_zlib_yaml
-      raise Puppet::Error, _("YAML in network requests is not supported. See http://links.puppetlabs.com/deprecate_yaml_on_network")
-    end
+  private
+
+  def valid_network_format?(format)
```
MosesMendoza: thank you for renaming this

spec/unit/network/http/api/indirected_routes_spec.rb
```diff
@@ -347,17 +327,6 @@
...
-      data = Puppet::IndirectorTesting.new("my data")
-      indirection.save(data, "my data")
-      request = a_request_that_destroys(data, :accept_header => nil)
-
-      handler.call(request, response)
-
-      expect(response.body).to eq(data.render(:json))
-      expect(response.type).to eq(Puppet::Network::FormatHandler.format(:json))
-    end
-
```
joshcooper: I forgot to `git add` where these tests moved to. Updated PR.

lib/puppet/network/http/request.rb
```diff
@@ -19,39 +19,48 @@ def route_into(prefix)
...
 
-  def response_formatter_for(supported_formats, accepted_formats = headers['accept'])
+  def response_formatter_for(supported_formats, default_accepted_formats = nil)
+    accepted_formats = headers['accept'] || default_accepted_formats
+
+    if accepted_formats.nil?
+      raise Puppet::Network::HTTP::Error::HTTPBadRequestError.new(_("Missing required Accept header"), Puppet::Network::HTTP::Issues::MISSING_HEADER_FIELD)
+    end
+
     formatter = Puppet::Network::FormatHandler.most_suitable_format_for(
```
MosesMendoza: Now with formatter method rename similar request - can we change the name of this local variable to avoid confusion with `Puppet::Network::HTTP::Request#formatter` up above?

spec/unit/network/http/request_spec.rb
```diff
@@ -0,0 +1,107 @@
+#! /usr/bin/env ruby
```
MosesMendoza: We seriously had no request_spec? Wow

lib/puppet/network/http/request.rb
joshcooper: updated

spec/unit/network/http/request_spec.rb
joshcooper: Most of it was tested in `indirected_routes_spec`, but there are some routes, e.g. `puppet/v3/environments`, that create a `Request`, but are not routed to `indirected_routes`, since they don't handle indirected models. Since we're pushing more functionality into `Request` it seemed best to move the tests there too.

PR #5839, opened 2017-05-03
(PUP-7486) useradd provider silently does not set password

cloakedcode:
If password-related parameters are set on User resources the client
should alarm when unable to enforce those parameters on the host. When
`ruby-libshadow` is unavailable the provider won't/can't set the
password-related parameters, but until now there was no indication in
the output on a client or server that the parameters were not applied.

A side affect of this warning is the `:manages_passwords` and
`:manages_password_age` features are always "on" for the `useradd` provider
in order to alarm during the application of the catalog. The removal of
validating `Puppet.features.libshadow?` at compile time is exchanged for
a better user experience at runtime.

### DIFF
--- a/lib/puppet/provider/user/useradd.rb
+++ b/lib/puppet/provider/user/useradd.rb
@@ -96,7 +96,7 @@ Puppet::Type.type(:user).provide :useradd, :parent => Puppet::Provider::NameServ
   has_features :manages_homedir, :allows_duplicates, :manages_expiry
   has_features :system_users unless %w{HP-UX Solaris}.include? Facter.value(:operatingsystem)
 
-  has_features :manages_passwords, :manages_password_age if Puppet.features.libshadow?
+  has_features :manages_passwords, :manages_password_age
   has_features :manages_shell
 
   def check_allow_dup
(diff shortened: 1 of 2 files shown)


MosesMendoza:
Hi @cloakedcode - thanks for taking the time to put this together. As written I don't think this will work for us though. `puppet resource user` queries the state of the system and with this PR that will fail instead of returning the list of users. I left a note in the ticket to continue the conversation on the best way to approach this - I'm not sure about failing the resource is the best path forward.

cloakedcode:
@MosesMendoza - I have no problem with this being a useless PR. For my own edification, what code path is followed by `puppet resource user` to result in a full list of users?

joshcooper:
@cloakedcode Thanks for taking a stab at this. I left some comments on the ticket. I think we want to understand why the error is failing instead of removing the feature detection logic, otherwise, puppet can't install the dependency in the same run that it's managing the user, without an explicit dependency between those resources.

MosesMendoza:
A bit of discussion is on-going in PUP-7486 (though it's stalled at the moment). I don't think we've decided to keep this approach as-is, so let's keep discussion in the ticket and when we settle on the next steps we can re-open this or address in a new way. Thanks for reporting it @cloakedcode, sorry for the delay in working this.

PR #5840, opened 2017-05-03
(PUP-7484) Move resource defaults initialization to initializer

thallgren:
In PUP-25, an attempt was made to add another phase to the compiler that
was responsible for setting default values. That was not sufficient since
new resources were created after the fact.

The PR for PUP-7484 that went in prior to this one attempted to fix this
by setting resource defaults on all resources created after the phase
where defaults had been set. This wasn't good enough either, since in
some cases, resources are created during evaluate_main phase.

This commit uses a simpler and more straight forward approach. Defaults
are now assigned in the Puppet::Parser::Resource#initialize method. The
method has an additional and optional argument that enables this behavior
to be turned off (needed when creating resources for the sake of
overriding parameter defaults).

### DIFF
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -460,6 +460,17 @@ class Puppet::Parser::Scope
     values
   end
 
+  # Check if the given value is a known default for the given type
+  #
+  def is_default?(type, key, value)
+    defaults_for_type = @defaults[type]
+    unless defaults_for_type.nil?
+      default_param = defaults_for_type[key]
+      return true if !default_param.nil? && value == default_param.value
+    end
+    !parent.nil? && parent.is_default?(type, key, value)
+  end
+
   # Look up a defined type.
   def lookuptype(name)
     # This happens a lot, avoid making a call to make a call
--- a/spec/unit/parser/resource_spec.rb
+++ b/spec/unit/parser/resource_spec.rb
@@ -51,6 +51,7 @@ describe Puppet::Parser::Resource do
   it "should get its environment from its scope" do
     scope = stub 'scope', :source => stub("source"), :namespaces => nil
     scope.expects(:environment).returns("foo").at_least_once
+    scope.expects(:lookupdefaults).returns({})
     expect(Puppet::Parser::Resource.new("file", "whatever", :scope => scope).environment).to eq("foo")
   end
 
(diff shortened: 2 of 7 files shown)


lib/puppet/parser/scope.rb
```diff
@@ -460,6 +460,17 @@ def lookupdefaults(type)
     values
   end
 
+  # Check if the given value is a known default for the given type
+  #
+  def is_default?(type, key, value)
+    defaults_for_type = @defaults[type]
+    unless defaults_for_type.nil?
+      default_param = defaults_for_type[key]
+      return true if !default_param.nil? && value == default_param.value
```
hlindberg: Does this mean that if a value is first set with a default (for example 42), and then overridden with the same value (42), and then overridden again - that it then thinks that the value is default the second time (it isn't since it was explicitly set)?

spec/unit/parser/resource_spec.rb
```diff
@@ -283,44 +284,69 @@ class { alpha: stage => before }
       expect(edges).to include(['One[a]', 'Notify[bob says hello]'])
       expect(edges).to include(['One[b]', 'Notify[bill says hello]'])
     end
+
+    it 'should override default values with new defaults' do
```
hlindberg: Technically the override is not a "new defaults" is it? It is just an override - an assignment.

spec/unit/parser/resource_spec.rb
```diff
@@ -283,44 +284,69 @@ class { alpha: stage => before }
...
+      @scope.expects(:lookupdefaults).with('File').returns(:owner => param(:owner, 'replaced', @source))
 
-      expect(@resource[:owner]).to eq("oldvalue")
+      resource.set_parameter(:owner, 'newvalue')
+      expect(resource[:owner]).to eq('newvalue')
     end
 
-    it "should add a copy of each default, rather than the actual default parameter instance" do
-      newparam = param(:owner, "default", @resource.source)
+    it 'should add a copy of each default, rather than the actual default parameter instance' do
```
hlindberg: Why a shallow copy and not a deep copy? The result of modifying something deep is just as bad.

hlindberg:
The changed order of evaluation of defaults means that resources will have the default values during the evaluation of puppet code, possibly changed later. Earlier, if a resource values was looked up, it never got a value for an attribute that got its value from a type level default since that would always be evaluated after all user code finished evaluating. For example, users depending on something like `File['foo']['mode']` to return `undef` if not given will now instead see the default value (if set). This is something that has been requested in the past. 
I don't think the change is more wrong than the current behavior since the override and collector override has mutating capabilities - it is just different, and may silently cause a different catalog result.

spec/unit/parser/resource_spec.rb
thallgren: Good question that I don't know the answer to. Not part of this PR.

lib/puppet/parser/scope.rb
thallgren: Yes. You will be able to add an override as long as the value is the default value. Keeping track of if an override to the exact same value just to prevent that another override later changes it (or doesn't) didn't seem worth the added complexity.

spec/unit/parser/resource_spec.rb
thallgren: Correct. It's just an override.

spec/unit/parser/resource_spec.rb
thallgren: Looking closer at this, I think that all instances of `Puppet::Parser::Resource::Param` should be frozen. The only place where it ever mutates seems to be when sensitive data values are unwrapped. That could just as well happen by replacing the Param instance with a copy. I think that deserves a separate ticket though.

joshcooper:
Thanks @thallgren and @hlindberg. I think we wanted to tag these commits with PUP-25 since that's the original issue we were trying to fix (as opposed to the subsequent CI failures). Nothing to do now, but next time.

Given the potential for breakage, can you add info to the original ticket about what a manifest looks like that would break under the new behavior? I want to make sure that it's prominent in the release notes.

PR #5841, opened 2017-05-03
(PUP-7464) Make json serialization handle :undef Symbol

hlindberg:
Before this, an :undef symbol nested in an Array or Hash used as a
resource's parameter value would serialize the Ruby `:undef` symbol.

The support for rich data checks if the value is compliant with Json an
if not transforms the entire Array or Hash into a String.

This commit modifies the "to_pson" such that an `:undef` value is
transformed to `nil` (which can be represented in JSON). Prior to this
commit, only Array's where scanned for values in need of transformation,
now also Hashes are transformed.

### DIFF
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -184,15 +184,7 @@ class Puppet::Resource
   end
 
   def yaml_property_munge(x)
-    case x
-    when Hash
-      x.inject({}) { |h,kv|
-        k,v = kv
-        h[k] = self.class.value_to_pson_data(v)
-        h
-      }
-    else self.class.value_to_pson_data(x)
-    end
+    self.value.to_pson_data(x)
   end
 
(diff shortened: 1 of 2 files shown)


lib/puppet/resource.rb
```diff
@@ -168,10 +168,15 @@ def is_json_type?(value)
   end
 
   def self.value_to_pson_data(value)
-    if value.is_a? Array
+    case value
+    when Array
       value.map{|v| value_to_pson_data(v) }
-    elsif value.is_a? Puppet::Resource
+    when Hash
+      Hash[ value.map {|k, v| [value_to_pson_data(k), value_to_pson_data(v)] } ]
```
thallgren: This change renders the `case` statement in the method `yaml_property_munge(x)` below, obsolete.

hlindberg:
Merging - approved, and no longer blocked

PR #5842, opened 2017-05-04
(PUP-5975) Remove monkey patches from Ruby Range

thallgren:
This commit removes the monkey patched method `Range#intersection` and its
alias `Range#&`.


hlindberg:
As I understand it, this removal does not affect any Puppet project logic, but may affect 3d parties that used the old semver (which is to be / is removed in Puppet 5.0.0).

PR #5843, opened 2017-05-04
(PUP-7492) Prevent that an interpolated alias interpolates again

thallgren:
This commit fixes a problem that caused a resolved "%{alias('key')}" to
be subject to interpolation once again after the alias had been resolve
which meant that sometimes, an alias would return a different value than
the key that it was an alias for.


hlindberg:
Unblocked. Merging.

PR #5844, opened 2017-05-04

MikaelSmith:
Under selinux, when Puppet is invoked by another process with reduced
privileges, any sub-programs invoked by Puppet will not inherit Puppet's
selinux priveleges. This specifically causes silent failures when
invoking applications that don't normally have the ability to write
files - such as iptables-save or hostname - because Puppet redirects
their output to a temporary file.

subprocesses.

### DIFF
--- a/acceptance/tests/resource/exec/should_accept_large_output.rb
+++ b/acceptance/tests/resource/exec/should_accept_large_output.rb
@@ -1,0 +1,26 @@
+test_name "tests that puppet correctly captures large and empty output."
+
+agents.each do |agent|
+  testfile = agent.tmpfile('should_accept_large_output')
+
+  # Generate >64KB file to exceed pipe buffer.
+  lorem_ipsum = <<EOF
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna
+aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
+occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+EOF
+  create_remote_file(agent, testfile, lorem_ipsum*1024)
+
+  apply_manifest_on(agent, "exec {'cat #{testfile}': path => ['/bin', '/usr/bin', 'C:/cygwin32/bin', 'C:/cygwin64/bin'], logoutput => true}") do
+    fail_test "didn't seem to run the command" unless
+      stdout.include? 'executed successfully'
+    fail_test "didn't print output correctly" unless
+      stdout.lines.select {|line| line =~ /\/returns:/}.count == 4097
+  end
+
+  apply_manifest_on(agent, "exec {'echo': path => ['/bin', '/usr/bin', 'C:/cygwin32/bin', 'C:/cygwin64/bin'], logoutput => true}") do
+    fail_test "didn't seem to run the command" unless
+      stdout.include? 'executed successfully'
+  end
+end
--- a/lib/puppet/util/execution.rb
+++ b/lib/puppet/util/execution.rb
@@ -187,18 +187,31 @@ module Puppet::Util::Execution
     null_file = Puppet.features.microsoft_windows? ? 'NUL' : '/dev/null'
 
     begin
+      # We close stdin/stdout/stderr immediately after execution as there no longer needed.
+      # In most cases they could be closed later, but when stdout is the writer pipe we
+      # must close it or we'll never reach eof on the reader.
+      reader, writer = IO.pipe unless options[:squelch]
       stdin = Puppet::FileSystem.open(options[:stdinfile] || null_file, nil, 'r')
-      stdout = options[:squelch] ? Puppet::FileSystem.open(null_file, nil, 'w') : Puppet::FileSystem::Uniquefile.new('puppet')
+      stdout = options[:squelch] ? Puppet::FileSystem.open(null_file, nil, 'w') : writer
       stderr = options[:combine] ? stdout : Puppet::FileSystem.open(null_file, nil, 'w')
 
       exec_args = [command, options, stdin, stdout, stderr]
+      output = ''
 
       if execution_stub = Puppet::Util::ExecutionStub.current_value
-        return execution_stub.call(*exec_args)
+        child_pid = execution_stub.call(*exec_args)
+        [stdin, stdout, stderr].each {|io| io.close rescue nil}
+        return child_pid
       elsif Puppet.features.posix?
         child_pid = nil
         begin
           child_pid = execute_posix(*exec_args)
+          [stdin, stdout, stderr].each {|io| io.close rescue nil}
+          unless options[:squelch]
+            while !reader.eof?
+              output << reader.read
+            end
+          end
           exit_status = Process.waitpid2(child_pid).last.exitstatus
           child_pid = nil
         rescue Timeout::Error => e
@@ -216,6 +229,12 @@ module Puppet::Util::Execution
       elsif Puppet.features.microsoft_windows?
         process_info = execute_windows(*exec_args)
         begin
+          [stdin, stdout, stderr].each {|io| io.close rescue nil}
+          unless options[:squelch]
+            while !reader.eof?
+              output << reader.read
+            end
+          end
           exit_status = Puppet::Util::Windows::Process.wait_process(process_info.process_handle)
         ensure
           FFI::WIN32.CloseHandle(process_info.process_handle)
@@ -223,21 +242,15 @@ module Puppet::Util::Execution
         end
       end
 
-      [stdin, stdout, stderr].each {|io| io.close rescue nil}
-
-      # read output in if required
-      unless options[:squelch]
-        output = wait_for_output(stdout)
-        Puppet.warning _("Could not get output") unless output
-      end
-
       if options[:failonfail] and exit_status != 0
         raise Puppet::ExecutionFailure, _("Execution of '%{str}' returned %{exit_status}: %{output}") % { str: command_str, exit_status: exit_status, output: output.strip }
       end
     ensure
-      if !options[:squelch] && stdout
-        # if we opened a temp file for stdout, we need to clean it up.
-        stdout.close!
+      # Make sure all handles are closed in case an exception was thrown attempting to execute.
+      [stdin, stdout, stderr].each {|io| io.close rescue nil}
+      if !options[:squelch] && reader
+        # if we opened a pipe, we need to clean it up.
+        reader.close
       end
     end
 
@@ -325,35 +338,4 @@ module Puppet::Util::Execution
     end
   end
   private_class_method :execute_windows
-
-
-  # This is private method.
-  # @comment see call to private_class_method after method definition
-  # @api private
-  #
-  def self.wait_for_output(stdout)
-    # Make sure the file's actually been written.  This is basically a race
-    # condition, and is probably a horrible way to handle it, but, well, oh
-    # well.
-    # (If this method were treated as private / inaccessible from outside of this file, we shouldn't have to worry
-    #  about a race condition because all of the places that we call this from are preceded by a call to "waitpid2",
-    #  meaning that the processes responsible for writing the file have completed before we get here.)
-    2.times do |try|
-      if Puppet::FileSystem.exist?(stdout.path)
-        stdout.open
-        begin
-          return stdout.read
-        ensure
-          stdout.close
-          stdout.unlink
-        end
-      else
-        time_to_sleep = try / 2.0
-        Puppet.warning _("Waiting for output; will sleep %{time_to_sleep} seconds") % { time_to_sleep: time_to_sleep }
-        sleep(time_to_sleep)
-      end
-    end
-    nil
-  end
-  private_class_method :wait_for_output
 end
--- a/spec/unit/util/execution_spec.rb
+++ b/spec/unit/util/execution_spec.rb
@@ -29,6 +29,7 @@ describe Puppet::Util::Execution do
       end
     end
 
+
     describe "#execute_posix (stubs)", :unless => Puppet.features.microsoft_windows? do
       before :each do
         # Most of the things this method does are bad to do during specs. :/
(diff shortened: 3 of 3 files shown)


MikaelSmith:
I think this works, and it solves the problem I'm trying to fix. However, it's pretty fundamental and I'm still in the process of testing it.

MikaelSmith:
Test build available at http://builds.puppetlabs.lan/puppet-agent/f4d9c43839580f89c3313601d316ab77ef89c733.

MikaelSmith:
Added some testing via pxp-agent and mcollective: https://github.com/puppetlabs/pxp-agent/pull/583, https://github.com/puppetlabs/marionette-collective/pull/433

lib/puppet/util/execution.rb
```diff
@@ -188,7 +188,8 @@ def self.execute(command, options = NoOptionsSpecified)
 
     begin
       stdin = Puppet::FileSystem.open(options[:stdinfile] || null_file, nil, 'r')
-      stdout = options[:squelch] ? Puppet::FileSystem.open(null_file, nil, 'w') : Puppet::FileSystem::Uniquefile.new('puppet')
+      reader, writer = IO.pipe
```
nicklewis: I believe the child needs to close the reader and the parent needs to close the writer.

MikaelSmith: Line 226/227 should be closing the writer.

MikaelSmith: 64KB test fails, need to read output in a loop to clear the buffer before waiting on the process.

MikaelSmith:
I believe this is ready. Kicked off a build at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_pkg-van-ship_ad-hoc-vmpooler-ad-hoc/54/, which includes acceptance tests from this PR, the pxp-agent PR, and the mco PR.

MikaelSmith:
Manually ran mco tests on RedHat 7
```
      Test Suite: tests @ 2017-05-05 15:07:30 -0700

       Total Suite Time: 156.45 seconds
      Average Test Time: 26.08 seconds
              Attempted: 6
                 Passed: 5
                 Failed: 0
                Errored: 0
                Skipped: 1
                Pending: 0
                  Total: 6

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
    Test Case tests/mco_puppet_powershell.rb
Pending Tests Cases:
```

MikaelSmith:
pxp-agent and mcollective tests failed on a few platforms due to test setup. I've fixed them, but verifying they now where can be handled after merging this fix.

MikaelSmith:
pxp-agent's run_puppet_killed_puppet test seems to be a real issue caused by this change. Looking into it.

Iristyle:
Wow, looks like the original Windows / Posix split happened in https://github.com/puppetlabs/puppet/commit/cb53870a8546216a170db4c21b090d76d10361cb

Though the original Tempfile usage stems from https://github.com/puppetlabs/puppet/commit/f8c1b08edf1c83d719511d847ad3be2b139275b9

MikaelSmith:
The original Tempfile usage is actually https://github.com/puppetlabs/puppet/commit/30ebbc90d396fa929fa1072dfd157e380c4668b5.

Looks like they attempted to fix it, which led to https://projects.puppetlabs.com/issues/3025.

MikaelSmith:
Using pipes on Windows seems to cause problems if Puppet is killed when we test the lockfile. `Process.kill(0, pid)` continues to return 1 until the subprocess exits.

acceptance/tests/resource/exec/should_accept_large_output.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+  # Generate >64KB file to exceed pipe buffer.
+  create_remote_file(agent, testfile, <<-EOF)
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna
+aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
+occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+EOF
+  on(agent, "for i in {1..10}; do cat #{testfile} #{testfile} > #{testfile}.2 && mv #{testfile}.2 #{testfile}; done")
```
Iristyle: Why not just generate 64k+ content on the Ruby side? It will send more over the wire, but will remove the Bash-ism above. There will be a QA undertaking shortly to prune / rewrite tests where possible.  The less Bash shows up in tests, the better I think.

MikaelSmith: Ok, I can rewrite that.

lib/puppet/util/execution.rb
```diff
@@ -187,18 +187,30 @@ def self.execute(command, options = NoOptionsSpecified)
...
+        child_pid = execution_stub.call(*exec_args)
+        [stdin, stdout, stderr].each {|io| io.close rescue nil}
+        return child_pid
       elsif Puppet.features.posix?
         child_pid = nil
         begin
           child_pid = execute_posix(*exec_args)
+          [stdin, stdout, stderr].each {|io| io.close rescue nil}
+          unless options[:squelch]
+            while !reader.eof?
```
Iristyle: I believe `read` is normally performed within a Ruby `Thread` since it's blocking.

You can look at the `capture2` source for an example - https://github.com/ruby/ruby/blob/trunk/lib/open3.rb#L305-L322

Iristyle: You can also take a look at the PowerShell module which uses pipes - https://github.com/puppetlabs/puppetlabs-powershell/blob/master/lib/puppet_x/puppetlabs/powershell/powershell_manager.rb#L240-L307

MikaelSmith: Since we're only using a single pipe, I think that's unnecessary.

lib/puppet/util/execution.rb
```diff
@@ -187,18 +187,30 @@ def self.execute(command, options = NoOptionsSpecified)
     null_file = Puppet.features.microsoft_windows? ? 'NUL' : '/dev/null'
 
     begin
+      reader, writer = IO.pipe unless options[:squelch]
```
MikaelSmith: Using a pipe on Windows causes an issue in https://github.com/puppetlabs/pxp-agent/blob/master/acceptance/tests/pxp-module-puppet/run_puppet_killed_puppet.rb.

When that test kills the Puppet process, the child process is intentionally left running. However, the mechanism Puppet uses to determine whether a previous Puppet lockfile [is stale](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/pidlock.rb#L55) reads the Puppet run as active. This appears to be because it's terminated, but still has an open handle to the pipe with the child process (perhaps specifically that it's waiting to read from that pipe).

MikaelSmith: I'm not sure what to do about that. It's not ideal to be unable to run Puppet again if a previous run was killed but left a long-running or hung process behind, but I'm not sure when that would really come up.

MikaelSmith: Puppet uses Ruby's Process.kill to check whether the process is still running, and that states the process is running if OpenProcess succeeds. Which it appears to as long as the child process has an open handle to the shared pipe.

MikaelSmith: Full explanation of closing a process in https://msdn.microsoft.com/en-us/library/windows/desktop/ms686714(v=vs.85).aspx.

MikaelSmith: I don't think this is behavior people depend on it Puppet, so I modified the pxp-agent test to not expect it in https://github.com/puppetlabs/pxp-agent/pull/583.

MikaelSmith:
Updated, test pipeline at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_pkg-van-ship_ad-hoc-vmpooler-ad-hoc/57/.

lib/puppet/util/execution.rb
MikaelSmith: Specifically, capture2 is using a thread for read while it also writes to stdin. We're not handling anything except reading from a single pipe, so only a single thread needed.

MikaelSmith:
Solaris machine died during Puppet acceptance tests, but everything else passed.

lib/puppet/util/execution.rb
MikaelSmith: @Iristyle have any concerns here?

acceptance/tests/resource/exec/should_accept_large_output.rb
```diff
@@ -0,0 +1,33 @@
...
+
+@testfilename = master.tmpfile('should_accept_large_output').split('/')[-1]
+
+agents.each do |agent|
+  testfile =
+    if /windows/ =~ agent[:platform]
+      "C:/#{@testfilename}"
+    else
+      "/tmp/#{@testfilename}"
+    end
```
joshcooper: I think you can use `host.tmpfile` which IIRC will return a platform-specific path that you can use later in the test. Might also add a teardown for deleting the file.

spec/unit/util/execution_spec.rb
```diff
@@ -550,39 +521,19 @@ def stub_process_wait(exitstatus)
...
 
-      it "should delete the file used for output if squelch is false" do
-        stdout = Puppet::FileSystem::Uniquefile.new('test')
-        path = stdout.path
-        Puppet::FileSystem::Uniquefile.stubs(:new).returns(stdout)
-
-        Puppet::Util::Execution.execute('test command')
-
-        expect(Puppet::FileSystem.exist?(path)).to be_falsey
-      end
```
joshcooper: should we add a test to make sure both ends of the pipe are closed (during normal run and if an exception is raised)?

lib/puppet/util/execution.rb
```diff
@@ -187,18 +187,30 @@ def self.execute(command, options = NoOptionsSpecified)
...
       stderr = options[:combine] ? stdout : Puppet::FileSystem.open(null_file, nil, 'w')
 
       exec_args = [command, options, stdin, stdout, stderr]
+      output = ''
+      read_succeeded = false
 
       if execution_stub = Puppet::Util::ExecutionStub.current_value
-        return execution_stub.call(*exec_args)
+        child_pid = execution_stub.call(*exec_args)
+        [stdin, stdout, stderr].each {|io| io.close rescue nil}
```
joshcooper: Puppetserver installs an execution stub, e.g. when executing an autosigning policy. Will it work correctly if we close these descriptors? /cc @camlow325

MikaelSmith: We should only be closing them on the Puppet side, the forked process inherits the handles. I'm actually not sure we need to close them earlier than we did before, since nothing else about forking changed.

lib/puppet/util/execution.rb
```diff
@@ -216,28 +228,32 @@ def self.execute(command, options = NoOptionsSpecified)
...
         end
       end
 
-      [stdin, stdout, stderr].each {|io| io.close rescue nil}
-
-      # read output in if required
-      unless options[:squelch]
-        output = wait_for_output(stdout)
-        Puppet.warning "Could not get output" unless output
+      unless options[:squelch] || read_succeeded
```
joshcooper: Is `read_succeeded ` necessary? If `options[:squelch]` is `false`, then we always call (for posix and windows):

```ruby
while !reader.eof?
  output << reader.read
end
read_succeeded = true
```

which always sets `read_succeeded ` to `true` or will raise. But if it's the latter, then we never reach this line.

lib/puppet/util/execution.rb
```diff
@@ -216,28 +228,32 @@ def self.execute(command, options = NoOptionsSpecified)
...
       if options[:failonfail] and exit_status != 0
         raise Puppet::ExecutionFailure, "Execution of '#{command_str}' returned #{exit_status}: #{output.strip}"
       end
     ensure
-      if !options[:squelch] && stdout
-        # if we opened a temp file for stdout, we need to clean it up.
-        stdout.close!
+      if !options[:squelch] && reader
+        # if we opened a pipe, we need to clean it up.
+        reader.close
```
joshcooper: If an exception is raised between the time we open the pipe and when the `execute_{posix,windows}` call returns, then the writer is never closed.

lib/puppet/util/execution.rb
MikaelSmith: There doesn't seem to be a rescue, so I agree I don't see any circumstance where we'd reach this line but reading had failed.

lib/puppet/util/execution.rb
camlow325: We already do close them in Puppet Server, although it seems like it would have better if we didn't have to.  See [this code](https://github.com/puppetlabs/puppetserver/blob/2.7.2/src/ruby/puppetserver-lib/puppet/server/execution.rb#L8-L12).

If we add this here, we might get exceptions when the second set of close calls are made.  That should be okay with the swallowing `rescue`, though, right?  Not sure if anything would be written to the log when this happens?

MikaelSmith: rescue nil should handle that in both cases, and nothing should be logged. The execution doesn't change at all in this case, after execution_stub.call we still would have closed all the handles at old line 226.

lib/puppet/util/execution.rb
MikaelSmith: I think you're misreading what block this ensure relates to. It starts at line 189.

spec/unit/util/execution_spec.rb
MikaelSmith: I'll try, that seems useful.

lib/puppet/util/execution.rb
MikaelSmith: This might make more sense outside the last ensure block.

lib/puppet/util/execution.rb
joshcooper: If an exception is raised between lines 190 and 201 (posix) or 207 (windows), then I don't think we close the `writer`:

```ruby
begin
  reader, writer = IO.pipe unless options[:squelch]
  ...
  process_info = execute_windows(*exec_args) # if this raises
  begin
    [stdin, stdout, stderr].each {|io| io.close rescue nil} # this line is never reached
    ...
   end
ensure
  if !options[:squelch] && reader
    reader.close
  end
end
```

A good test is to try to execute a non-existing program with `squelch = false`. On Windows, I know `execute_windows` will raise, and I think the posix impl behaves the same.

MikaelSmith: That does seem like a loose end. However, as far as I can tell if we never fork and then close one end, the other end also closes. So in that scenario the pipe would still be closed.

I'm not sure if that leaves dangling handles for the files. We could still attempt to close all handles in the ensure block anyway.

MikaelSmith:
Updated and kicked off a new build.

Iristyle:
Do we want this on stable or master @MikaelSmith ?

MikaelSmith:
I was hoping stable.

joshcooper:
@MikaelSmith In talking this over with the team, it feels too risky of a change to make in a .z without some assurances it won't cause regressions on other platforms. Can we target this for master, and if there are no downstream issues, backport to stable?

MikaelSmith:
Sure. I'll rebase and retarget.

MikaelSmith:
Done. I'll retarget the pxp-agent and mco acceptance tests to master as well.

MikaelSmith:
Not sure why it pulled in extra commits for commit checking, but everything else was passing.

Magisus:
Yeah looks like it didn't update the job parameter for some reason... weird.

joshcooper:
I think the failure is because we're using double-dots in the commit check, so it's pulling in commits on the base branch that were made after you forked this branch, and it looks like the "Revert" commits are causing issues:

```
Checking commits 054989b458a3a48188551bf8efbf855cb767a201..6519360061e2a005b0b5913f2789ea15fe792842
```

I thought we had updated travis and appveyor to use triple dots?

Magisus:
We did: https://github.com/puppetlabs/puppet/blob/master/Rakefile#L86. I think it was just Github's PR re-targeting being kind of screwy with commit hooks. We've seen stuff like that a lot.

Iristyle:
Reverted in #5868

PR #5845, opened 2017-05-05
(PUP-7261) Include `charset` parameter in the Content-Type

joshcooper:
This PR marks text-based formats (json, text/plain) as supporting a `charset` parameter.

If the response body is a string and the serialization format supports charset, then Puppet will encode the body in that charset (if it isn't already), and include the `charset` parameter in the `Content-Type` header. In a future PR, Puppet's HTTP client will use that information to decode the response body, such as when the response is compressed.

The `charset` parameter will be omitted is the response body is a file or if the `Content-Type` is a binary format (`application/octet-stream`, `msgpack`), `pson`, or `yaml`.

We don't include it with `pson` because old agent's don't know about `charset`. We don't include it with `yaml` because it's not a network serialization format.

### DIFF
--- a/lib/puppet/network/format.rb
+++ b/lib/puppet/network/format.rb
@@ -5,7 +5,7 @@ require 'puppet/confiner'
 class Puppet::Network::Format
   include Puppet::Confiner
 
-  attr_reader :name, :mime
+  attr_reader :name, :mime, :charset
   attr_accessor :intern_method, :render_method, :intern_multiple_method, :render_multiple_method, :weight, :required_methods, :extension
 
   def init_attribute(name, default)
@@ -34,6 +34,7 @@ class Puppet::Network::Format
     init_attribute(:weight, 5)
     init_attribute(:required_methods, method_list.keys)
     init_attribute(:extension, name.to_s)
+    init_attribute(:charset, nil)
 
     method_list.each do |method, value|
       init_attribute(method, value)
@@ -58,6 +59,10 @@ class Puppet::Network::Format
 
   def mime=(mime)
     @mime = mime.to_s.downcase
+  end
+
+  def charset=(charset)
+    @charset = charset
   end
 
   def render(instance)
--- a/lib/puppet/network/format.rb
+++ b/lib/puppet/network/format.rb
@@ -5,8 +5,8 @@ require 'puppet/confiner'
 class Puppet::Network::Format
   include Puppet::Confiner
 
-  attr_reader :name, :mime, :charset
-  attr_accessor :intern_method, :render_method, :intern_multiple_method, :render_multiple_method, :weight, :required_methods, :extension
+  attr_reader :name, :mime
+  attr_accessor :intern_method, :render_method, :intern_multiple_method, :render_multiple_method, :weight, :required_methods, :extension, :charset
 
   def init_attribute(name, default)
     value = @options.delete(name)
@@ -59,10 +59,6 @@ class Puppet::Network::Format
 
   def mime=(mime)
     @mime = mime.to_s.downcase
-  end
-
-  def charset=(charset)
-    @charset = charset
   end
 
   def render(instance)
--- a/lib/puppet/network/format.rb
+++ b/lib/puppet/network/format.rb
@@ -9,11 +9,9 @@ class Puppet::Network::Format
   attr_accessor :intern_method, :render_method, :intern_multiple_method, :render_multiple_method, :weight, :required_methods, :extension
 
   def init_attribute(name, default)
-    if value = @options[name]
-      @options.delete(name)
-    else
-      value = default
-    end
+    value = @options.delete(name)
+    value = default if value.nil?
+
     self.send(name.to_s + "=", value)
   end
 
--- a/lib/puppet/network/http/response.rb
+++ b/lib/puppet/network/http/response.rb
@@ -5,7 +5,23 @@ class Puppet::Network::HTTP::Response
   end
 
   def respond_with(code, type, body)
-    @handler.set_content_type(@response, type)
+    format = Puppet::Network::FormatHandler.format_for(type)
+    mime = format.mime
+    charset = format.charset
+
+    if charset && body.is_a?(String)
+      # REMIND: not all charsets are valid ruby encodings, e.g. ISO-2022-KR
+      encoding = Encoding.find(charset)
+
+      if body.encoding != encoding
+        # REMIND this can raise if body contains invalid UTF-8
+        body.encode!(encoding)
+      end
+
+      mime += "; charset=#{charset}"
+    end
+
+    @handler.set_content_type(@response, mime)
     @handler.set_response(@response, body, code)
   end
 end
--- a/lib/puppet/network/http/response.rb
+++ b/lib/puppet/network/http/response.rb
@@ -9,13 +9,15 @@ class Puppet::Network::HTTP::Response
     mime = format.mime
     charset = format.charset
 
-    if charset && body.is_a?(String)
-      # REMIND: not all charsets are valid ruby encodings, e.g. ISO-2022-KR
-      encoding = Encoding.find(charset)
+    if charset
+      if body.is_a?(String)
+        # REMIND: not all charsets are valid ruby encodings, e.g. ISO-2022-KR
+        encoding = Encoding.find(charset)
 
-      if body.encoding != encoding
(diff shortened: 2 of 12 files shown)


joshcooper:
@Iristyle @MosesMendoza I'm not sure if additional changes are needed to honor the `charset` parameter when the client reads the response? For example, do we need to force_encoding the string that [`GZipReader(..).read`](https://github.com/puppetlabs/puppet/commit/4d2589f35ba0a6284c376dfd06b73707c6356f06#diff-a7c86d1dfe193edd8e64ee4919690595R23) returns?

joshcooper:
Looks like I have more work to do. We parse [strip the charset off and throw it away](https://github.com/puppetlabs/puppet/blob/eedec930595e7ef78eff90256dcfac31bf52e84b/lib/puppet/indirector/rest.rb#L300) when parsing the response body. I'm thinking if a `charset` is specified, then we should `force_encoding` on the string that `uncompress_body` returns. Otherwise, the returned string will be in whatever encoding GzipReader/Inflate returns.

nicklewis:
This appears to fix the issue I've been seeing with unicode characters not making it through correctly.

lib/puppet/network/http/response.rb
```diff
@@ -5,7 +5,18 @@ def initialize(handler, response)
   end
 
   def respond_with(code, type, body)
-    @handler.set_content_type(@response, type)
+    format = Puppet::Network::FormatHandler.format_for(type)
+    mime = format.mime
+
+    if format.supports_charset
+      if body.encoding != Encoding::UTF_8
+        body.encode!(Encoding::UTF_8)
```
Iristyle: I'm not sure if you'll ever get into this state (since I don't know what `format.supports_charset` does), but if you have a `BINARY` string here, calling `encode!` will crash.

Iristyle:
On your reference to https://github.com/puppetlabs/puppet/commit/4d2589f35ba0a6284c376dfd06b73707c6356f06#diff-a7c86d1dfe193edd8e64ee4919690595R23 I suggested initially that we actually change that code:

```ruby
# ZLib::GzipReader has an associated encoding, by default Encoding.default_external
return Zlib::GzipReader.new(StringIO.new(response.body), :encoding => Encoding::BINARY).read
```

To be more explicit like this:

```ruby
return Zlib::GzipReader.new(StringIO.new(response.body, "r+:#{Encoding::BINARY.name}"), :encoding => Encoding::BINARY).read
```

`StringIO.new` does support the mode string as a 2nd parameter in the initializer, which also accepts the weird Ruby syntax that supports the encoding name after a `:`.

However, what I found out is that the `StringIO.new` initializer will ignore the encoding passed and it will always use the encoding of the string being passed in (in this case `response.body.encoding`).  Gzip'd strings should always be a stream of bytes at this point - the most important part is how the encoding is specified to the `GzipReader` initializer - we need to make sure it won't transcode anything, hence we specify `BINARY` which tells Ruby to leave the byte sequence alone.

joshcooper:
@nicklewis I think [this explains the behavior you were seeing](https://github.com/flori/json/blob/master/ext/json/ext/parser/parser.c#L1687-L1694). JSON.parse assumes if the encoding label is binary, then the string can be force encoded to UTF-8. But if you pass a Shift-JIS string whose encoding label is binary, then it will incorrectly force encode the label to `UTF-8`, preventing the characters from being interpreted correctly.

MikaelSmith:
Why is this still blocked?

joshcooper:
When decompressing the gzipped content, the ruby string has a binary encoding, which isn't right. For reasons I'm not sure, puppet still outputs the snowman correctly, but we might be getting lucky, as I would expect puppet to fail if it tried to string concat or apply a regex on the binary string. I hope to have an updated version today.

lib/puppet/network/http/response.rb
```diff
@@ -5,7 +5,23 @@ def initialize(handler, response)
...
+    mime = format.mime
+    charset = format.charset
+
+    if charset && body.is_a?(String)
+      # REMIND: not all charsets are valid ruby encodings, e.g. ISO-2022-KR
+      encoding = Encoding.find(charset)
+
+      if body.encoding != encoding
+        # REMIND this can raise if body contains invalid UTF-8
+        body.encode!(encoding)
```
joshcooper: @Iristyle @MosesMendoza I wanted to sync up with tomorrow about this.

1. Can we assume if we're using `application/json` or `text/plain` that the response body is always valid UTF-8 or can be at least be transcoded to it?
2. I'm assuming charset is really only going to be `utf-8` or `nil`. Does that seem reasonable?

joshcooper:
I'm going to split this work into 2 PRs. This PR ensures we add `charset` to HTTP responses, e.g. catalog, for serialization formats that support it, e.g. json, but not binary. I will put up a separate PR to honor the charset (if present) when deserializing the response.

@nicklewis could you verify this still resolves the issue you were seeing?

nicklewis:
@joshcooper Yep, this still appears to fix my issue.

lib/puppet/network/http/response.rb
joshcooper: I believe raising if `encode!` fails is the correct thing to do. When sending `json`, the native JSON generator will ensure it emits valid utf-8 data, and will raise with invalid input:

```
irb(main):001:0> require 'json'
=> true
irb(main):002:0> "\355\274\267".to_json
JSON::GeneratorError: source sequence is illegal/malformed utf-8
...
```

So if `encode!` fails, then something went wrong between the time we rendered the result and here.

Note that `pson` will emit it as, but we don't attempt to encode `pson`:

```
irb(main):003:0> require 'puppet'
1=> true
irb(main):004:0> "\355\274\267".to_pson
=> "\"\xED\xBC\xB7\""
```

joshcooper: Add a test that `respond_with` raises if given data that can't be transcoded to UTF-8

lib/puppet/network/http/response.rb
```diff
@@ -5,7 +5,23 @@ def initialize(handler, response)
...
 
   def respond_with(code, type, body)
-    @handler.set_content_type(@response, type)
+    format = Puppet::Network::FormatHandler.format_for(type)
+    mime = format.mime
+    charset = format.charset
+
+    if charset && body.is_a?(String)
+      # REMIND: not all charsets are valid ruby encodings, e.g. ISO-2022-KR
+      encoding = Encoding.find(charset)
```
joshcooper: Should raise an appropriate HTTP error if the charset is unknown. In practice shouldn't happen since it's either `utf-8` or `nil`, but that could change in the future.

lib/puppet/network/format.rb
```diff
@@ -62,6 +61,10 @@ def mime=(mime)
     @mime = mime.to_s.downcase
   end
 
+  def charset=(charset)
+    @charset = charset
+  end
```
joshcooper: I'll update that

spec/unit/network/format_spec.rb
```diff
@@ -129,6 +129,15 @@ def to_my_format
     it "should support overriding the extension" do
       expect(Puppet::Network::Format.new(:foo, :extension => "bar").extension).to eq("bar")
     end
+
+    it "doesn't support charset by default" do
+      expect(Puppet::Network::Format.new(:foo).charset).to be_nil
+    end
+
+    it "allows charset to be set to 'utf-8'" do
+      expect(Puppet::Network::Format.new(:foo, :charset => 'utf-8').charset).to eq('utf-8')
```
joshcooper: I need to update this

joshcooper:
@MosesMendoza updated

lib/puppet/network/http/response.rb
```diff
@@ -5,7 +5,19 @@ def initialize(handler, response)
...
+
+    if charset
+      if body.is_a?(String) && body.encoding != charset
+        body.encode!(charset)
+      end
+
+      mime += "; charset=#{charset.name.downcase}"
+    end
+
+    @handler.set_content_type(@response, mime)
```
joshcooper: The good thing is that `format_to_mime` passes through the argument if the argument isn't a Format object. So this PR won't break puppetserver, as it will continue calling `format_to_mime("application/json; charset=utf-8")` and our `format_to_mime` will be a noop.

spec/unit/network/format_spec.rb
MosesMendoza: Ah yes. I missed that

PR #5847, opened 2017-05-06
(PUP-7110) Execute agent run in separate contract

mruzicka:
Before this commit, it was not possible to restart/stop a puppet agent service on Solaris from an agent run executed by that very service without killing the agent run along with the service.
This was because services on Solaris run in so called contracts - a kind of process groups - and during a service restart/stop all processes in the contract are killed. Under normal circumstances the contract is inherited during fork system call so the child process is a member of the same contract as its parent process. This was the case of the puppet run executed by the puppet agent service on Solaris too, and that is why the puppet run could not restart/stop the puppet agent service without killing itself.
So in this commit we simply make sure the agent run is executed in a different contract than its parent (the puppet agent service) which fixes the issue.

### DIFF
--- a/lib/puppet/agent.rb
+++ b/lib/puppet/agent.rb
@@ -1,5 +1,6 @@
 require 'puppet/application'
 require 'puppet/error'
+require 'puppet/util/at_fork'
 
 # A general class for triggering a run of another
 # class.
@@ -64,16 +65,26 @@ class Puppet::Agent
   def run_in_fork(forking = true)
     return yield unless forking or Puppet.features.windows?
 
-    child_pid = Kernel.fork do
-      $0 = _("puppet agent: applying configuration")
-      begin
-        exit(yield)
-      rescue SystemExit
-        exit(-1)
-      rescue NoMemoryError
-        exit(-2)
+    atForkHandler = Puppet::Util::AtFork.get_handler
+
+    atForkHandler.prepare
+
+    begin
+      child_pid = Kernel.fork do
+        atForkHandler.child
+        $0 = _("puppet agent: applying configuration")
+        begin
+          exit(yield)
+        rescue SystemExit
+          exit(-1)
+        rescue NoMemoryError
+          exit(-2)
+        end
       end
+    ensure
+      atForkHandler.parent
     end
+
     exit_code = Process.waitpid2(child_pid)
     case exit_code[1].exitstatus
     when -1
--- a/lib/puppet/util/at_fork.rb
+++ b/lib/puppet/util/at_fork.rb
@@ -1,0 +1,35 @@
+require 'puppet'
+
+# A module for building AtFork handlers. These handlers are objects providing
+# pre/post fork callbacks modeled after those registered by the `pthread_atfork`
+# function.
+# Currently there are two AtFork handler implementations:
+# - a noop implementation used on all platforms except Solaris (and possibly
+#   even there as a fallback)
+# - a Solaris implementation which ensures the forked process runs in a different
+#   contract than the parent process. This is necessary for agent runs started by
+#   the puppet agent service to be able to restart that service without being
+#   killed in the process as a consequence of running in the same contract as the
+#   service.
+module Puppet::Util::AtFork
+  @handler_class = loop do
+    if Facter.value(:operatingsystem) == 'Solaris'
+      begin
+        require 'puppet/util/at_fork/solaris'
+        # using break to return a value from the loop block
+        break Puppet::Util::AtFork::Solaris
+      rescue LoadError => detail
+        Puppet.log_exception(detail, _('Failed to load Solaris implementation of the Puppet::Util::AtFork handler. Child process contract management will be unavailable, which means that agent runs executed by the puppet agent service will be killed when they attempt to restart the service.'))
+        # fall through to use the no-op implementation
+      end
+    end
+
+    require 'puppet/util/at_fork/noop'
+    # using break to return a value from the loop block
+    break Puppet::Util::AtFork::Noop
+  end
+
+  def self.get_handler
+    @handler_class.new
+  end
+end
--- a/lib/puppet/util/at_fork/solaris.rb
+++ b/lib/puppet/util/at_fork/solaris.rb
@@ -1,0 +1,158 @@
+require 'puppet'
+require 'fiddle'
+
+# Early versions of Fiddle relied on the deprecated DL module and used
+# classes defined in the namespace of that module instead of classes defined
+# in the Fiddle's own namespace e.g. DL::Handle instead of Fiddle::Handle.
+# We don't support those.
+raise LoadError, _('The loaded Fiddle version is not supported.') unless defined?(Fiddle::Handle)
+
+# Solaris implementation of the Puppet::Util::AtFork handler.
+# The callbacks defined in this implementation ensure the forked process runs
+# in a different contract than the parent process. This is necessary in order
+# for the child process to be able to survive termination of the contract its
+# parent process runs in. This is needed notably for an agent run executed
+# by a puppet agent service to be able to restart that service without being
+# killed in the process as a consequence of running in the same contract as
+# the service, as all processes in the contract are killed when the contract
+# is terminated during the service restart.
+class Puppet::Util::AtFork::Solaris
+  private
+
+  {
+    'libcontract.so.1' => [
+      #function name,            return value type, parameter types, ...
+      [:ct_ctl_abandon,          Fiddle::TYPE_INT,  Fiddle::TYPE_INT],
+      [:ct_tmpl_activate,        Fiddle::TYPE_INT,  Fiddle::TYPE_INT],
+      [:ct_tmpl_clear,           Fiddle::TYPE_INT,  Fiddle::TYPE_INT],
+
+      [:ct_tmpl_set_informative, Fiddle::TYPE_INT,  Fiddle::TYPE_INT, Fiddle::TYPE_INT],
+      [:ct_tmpl_set_critical,    Fiddle::TYPE_INT,  Fiddle::TYPE_INT, Fiddle::TYPE_INT],
+      [:ct_pr_tmpl_set_param,    Fiddle::TYPE_INT,  Fiddle::TYPE_INT, Fiddle::TYPE_INT],
+      [:ct_pr_tmpl_set_fatal,    Fiddle::TYPE_INT,  Fiddle::TYPE_INT, Fiddle::TYPE_INT],
+
+      [:ct_status_read,          Fiddle::TYPE_INT,  Fiddle::TYPE_INT, Fiddle::TYPE_INT, Fiddle::TYPE_VOIDP],
+
+      [:ct_status_get_id,        Fiddle::TYPE_INT,  Fiddle::TYPE_VOIDP],
+      [:ct_status_free,          Fiddle::TYPE_VOID, Fiddle::TYPE_VOIDP],
+    ],
+  }.each do |library, functions|
+    libhandle = Fiddle::Handle.new(library)
+
+    functions.each do |f|
+      define_method f[0], Fiddle::Function.new(libhandle[f[0].to_s], f[2..-1], f[1]).method(:call).to_proc
+    end
+  end
+
+  CTFS_PR_ROOT = File.join('', %w(system contract process))
+  CTFS_PR_TEMPLATE = File.join(CTFS_PR_ROOT, %q(template))
+  CTFS_PR_LATEST = File.join(CTFS_PR_ROOT, %q(latest))
+
+  CT_PR_PGRPONLY = 0x4
+  CT_PR_EV_HWERR = 0x20
+
+  CTD_COMMON = 0
+
+  def raise_if_error(&block)
+    unless (e = yield) == 0
+      e = SystemCallError.new(nil, e)
+      raise e, e.message, caller
+    end
+  end
+
+  def activate_new_contract_template
+    begin
+      tmpl = File.open(CTFS_PR_TEMPLATE, File::RDWR)
+
+      begin
+        tmpl_fd = tmpl.fileno
+
+        raise_if_error { ct_pr_tmpl_set_param(tmpl_fd, CT_PR_PGRPONLY) }
+        raise_if_error { ct_pr_tmpl_set_fatal(tmpl_fd, CT_PR_EV_HWERR) }
+        raise_if_error { ct_tmpl_set_critical(tmpl_fd, 0) }
+        raise_if_error { ct_tmpl_set_informative(tmpl_fd, CT_PR_EV_HWERR) }
+
+        raise_if_error { ct_tmpl_activate(tmpl_fd) }
+      rescue
+        tmpl.close
+        raise
+      end
+
+      @tmpl = tmpl
+    rescue => detail
+      Puppet.log_exception(detail, _('Failed to activate a new process contract template'))
+    end
+  end
+
+  def deactivate_contract_template(parent)
+    return if @tmpl.nil?
+
+    tmpl = @tmpl
+    @tmpl = nil
+
+    begin
+      raise_if_error { ct_tmpl_clear(tmpl.fileno) }
+    rescue => detail
+      Puppet.log_exception(detail, parent \
+        ? _('Failed to deactivate process contract template in the parent process')
+        : _('Failed to deactivate process contract template in the child process')
+      )
+      exit(1)
+    ensure
+      tmpl.close
+    end
+  end
+
+  def get_latest_child_contract_id
+    begin
+      stat = File.open(CTFS_PR_LATEST, File::RDONLY)
+
+      begin
+        stathdl = Fiddle::Pointer.new(0)
+
+        raise_if_error { ct_status_read(stat.fileno, CTD_COMMON, stathdl.ref) }
+        ctid = ct_status_get_id(stathdl)
+        ct_status_free(stathdl)
+      ensure
+        stat.close
+      end
+
+      ctid
+    rescue => detail
+      Puppet.log_exception(detail, _('Failed to get latest child process contract id'))
+      nil
+    end
+  end
+
+  def abandon_latest_child_contract
+    ctid = get_latest_child_contract_id
+    return if ctid.nil?
+
+    begin
+      ctl = File.open(File.join(CTFS_PR_ROOT, ctid.to_s, %q(ctl)), File::WRONLY)
+
+      begin
+        raise_if_error { ct_ctl_abandon(ctl.fileno) }
+      ensure
+        ctl.close
+      end
+    rescue => detail
+      Puppet.log_exception(detail, _('Failed to abandon a child process contract'))
+    end
+  end
+
+  public
+
+  def prepare
+    activate_new_contract_template
+  end
+
+  def parent
+    deactivate_contract_template(true)
+    abandon_latest_child_contract
+  end
+
+  def child
+    deactivate_contract_template(false)
+  end
+end
--- a/lib/puppet/util/at_fork/noop.rb
+++ b/lib/puppet/util/at_fork/noop.rb
@@ -1,0 +1,18 @@
+# A noop implementation of the Puppet::Util::AtFork handler
+class Puppet::Util::AtFork::Noop
+  class << self
+    def new
+      # no need to instantiate every time, return the class object itself
+      self
+    end
+
+    def prepare
+    end
+
+    def parent
+    end
+
+    def child
+    end
+  end
+end
--- a/spec/unit/util/at_fork_spec.rb
+++ b/spec/unit/util/at_fork_spec.rb
@@ -1,0 +1,150 @@
+#! /usr/bin/env ruby
+require 'spec_helper'
+
+describe 'Puppet::Util::AtFork' do
+  EXPECTED_HANDLER_METHODS = [:prepare, :parent, :child]
+
+  before :each do
+    Puppet::Util.class_exec do
+      const_set(:AtFork, Module.new)
+    end
+  end
+
+  after :each do
+    Puppet::Util.class_exec do
+      remove_const(:AtFork)
+    end
+  end
+
+  describe '.get_handler' do
+    context 'when on Solaris' do
+      before :each do
+        Facter.expects(:value).with(:operatingsystem).returns('Solaris')
+      end
+
+      after :each do
+        Object.class_exec do
+          remove_const(:Fiddle) if const_defined?(:Fiddle)
+        end
+      end
+
+      def stub_solaris_handler(stub_noop_too = false)
+        Puppet::Util::AtFork.stubs(:require).with() do |lib|
+          if lib == 'puppet/util/at_fork/solaris'
+            load lib + '.rb'
+            true
+          elsif stub_noop_too && lib == 'puppet/util/at_fork/noop'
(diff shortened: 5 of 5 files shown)


lib/puppet/agent.rb
```diff
@@ -1,6 +1,167 @@
...
+# than the parent process. This is necessary in order for the child process
+# (an agent run) to be able to restart/stop the puppet agent service without
+# being killed (as a result of the termination of the parent process
+# contract during the service restart/stop).
+# The implementation uses Fiddle to make calls to functions defined in the
+# libcontract library.
+class AtFork
+  require 'fiddle'
+
+  raise LoadError, %q(The loaded fiddle version doesn't define 'Fiddle::Handle') unless defined?(Fiddle::Handle)
```
MikaelSmith: What Fiddle versions don't support this? How likely are we to encounter them in the wild?

If we expect people to run into this error, I'd be interested in making it more graceful (i.e. the AtFork functionality doesn't work but everything else works as before).

MikaelSmith: Oh, I guess the rescue handles old versions of Fiddle or no libcontract by falling back to noop.

joshcooper: If we're using an old version of fiddle, then just `require 'puppet'` will fail, even if we're not on Solaris. Can we move this into an `AtFork` method, so we only check if/when we're on Solaris and need to make a fiddle call?

lib/puppet/agent.rb
```diff
@@ -1,6 +1,167 @@
...
+      tmpl_fd = tmpl.fileno
+
+      raise if ct_pr_tmpl_set_param(tmpl_fd, CT_PR_PGRPONLY) != 0
+      raise if ct_pr_tmpl_set_fatal(tmpl_fd, CT_PR_EV_HWERR) != 0
+      raise if ct_tmpl_set_critical(tmpl_fd, 0) != 0
+      raise if ct_tmpl_set_informative(tmpl_fd, CT_PR_EV_HWERR) != 0
+
+      raise if ct_tmpl_activate(tmpl_fd) != 0
+    rescue
+      tmpl.close()
```
joshcooper: If someone adds code between the two begin/rescue blocks and that code raises, then close won't be called. Could simplify to:

```ruby
begin
  tmpl = File.open(..)
  begin
    tmpl_fd = tmpl.fileno
    ...
  ensure
    tmpl.close
  end
rescue => detail
  Puppet.error(_("Failed to activate contract template: %{detail}") % { detail: detail }
end
```

lib/puppet/agent.rb
MikaelSmith: It appears the whole class has a `rescue` block. I haven't worked with that before.

lib/puppet/agent.rb
```diff
@@ -1,6 +1,167 @@
...
+      Puppet.fatal 'Failed to configure / activate contract template'
+    end
+
+    @tmpl = tmpl
+  end
+
+  def deactivate_contract_template(where)
+    return 0 if @tmpl.nil?
```
joshcooper: Can we just `return if @tmpl.nil?` since the caller doesn't look at the return code?

lib/puppet/agent.rb
```diff
@@ -1,6 +1,167 @@
...
+    begin
+      stathdl = Fiddle::Pointer.new(0)
+
+      raise if ct_status_read(stat.fileno, CTD_COMMON, stathdl.ref) != 0
+      ctid = ct_status_get_id(stathdl)
+      ct_status_free(stathdl)
+    rescue
+      Puppet.fatal 'Unable to read contract status'
+    ensure
+      stat.close()
```
joshcooper: Similar comment about how `stat.close` might not be called if someone adds code on line 104.

lib/puppet/agent.rb
joshcooper: Ah... I think it'd be better to move the raise into the place we try to use it, since ruby will raise a NameError if you try to reference a non-existent constant:

```ruby
begin
  libhandle = Fiddle::Handle.new(library)
rescue NameError => e
  raise _("The loaded fiddle version doesn't define 'Fiddle::Handle'")
rescue Fiddle::DLError => e
  raise LoadError, e.message, e.backtrace
end
```

lib/puppet/agent.rb
```diff
@@ -1,6 +1,167 @@
...
+    rescue
+      Puppet.fatal 'Unable to access latest contract'
+    end
+
+    begin
+      stathdl = Fiddle::Pointer.new(0)
+
+      raise if ct_status_read(stat.fileno, CTD_COMMON, stathdl.ref) != 0
+      ctid = ct_status_get_id(stathdl)
+      ct_status_free(stathdl)
```
joshcooper: If  line 108 raises, then we never free the `stathdl`. A good way to handle this is to nest begin/rescues:

```ruby
stat = File.open(...)
begin
  stathdl = Fiddle::Pointer.new(0)
  begin
    raise if ct_status_read(stat.fileno, CTD_COMMON, stathdl.ref) != 0
    ...
  ensure
    ct_status_free(stathdl)
  end
rescue
  ...
ensure
  stat.close
end
```

Or create private helper methods to yield that param and ensure it's closed/released:

```ruby
def with_stat(&block)
  stat = File.open(...)
  yield stat
ensure
  stat.close
end
```

Then the caller can do:

```ruby
with_stat do |stat|
  with_fiddle_pointer do |stathdl|
    raise if ct_status_read(stat.fileno, CTD_COMMON, stathdl.ref) != 0
    ...
  end
end

File: lib/puppet/agent.rb (hunk 1 above)
Comment (joshcooper) [inline lib/puppet/agent.rb]: Ah, sorry I get it now... One concern is that if we are running on Solaris, and we fail to map one of the functions (maybe a newer solaris version changes a method signature?), then we raise a LoadError, catch it, and silently "downgrade" to noop methods. So we could regress on this issue and not know it.

Could we instead define two implementations, one for Solaris, and one for everyone else, something like we do for FileSystem:

```ruby
class AtFork
  @impl = if Facter.value(:operatingsystem) == 'Solaris'
            SolarisAtFork.new
          else
             NoopAtFork.new
          end

  def prepare
    @impl.prepare
  end
end

class NoopAtFork
  def prepare
  end
  ...
end

class SolarisAtFork
  begin
    libhandle = Fiddle::Handle.new(library)
  rescue NameError => e
    raise _("The loaded fiddle version doesn't define 'Fiddle::Handle'")
  rescue Fiddle::DLError => e
    raise LoadError, e.message, e.backtrace
  end
  
  def prepare
    activate_new_contract_template
  end
end
```

mruzicka: A bit of explanation, which hopefully answers some of your questions:
* Without this check travis was failing on ruby 1.9.3 which apparently includes an early version of Fiddle which doesn't define the `Fiddle::Handle` class and does define (and throws) different exceptions (e.g. `DL::DLError` instead of `Fiddle::DLError`) than more modern versions, so in order to avoid having to complicate the implementation to make it compatible with this older versions of fiddle I opted for adding the check.
* The check is really meant to ensure the code is executing with a sufficiently modern version of Fiddle - ideally I'd do something like:
```ruby
raise "Too old version of fiddle." unless Fiddle::Version >= x.y.z
```
and as such it felt reasonable to place it at the beginning.
* I agree that checking for whether the code is executing on Solaris is a more robust approach than the no-op fallback.

lib/puppet/agent.rb
mruzicka: I don't think you can avoid that problem (the code above has the same problem if someone adds code between the line with the `File.open` call and the `begin` on the the next line) unless at the expense of more clutter:
```
tmpl = nil
begin
  tmpl = File.open(...)
  ...
rescue
  tmpl.close() unless tmpl.nil?
end
```

lib/puppet/agent.rb
mruzicka: Absolutely!

lib/puppet/agent.rb
mruzicka: If line 108 raises, there is nothing to free. The `stathdl` itself is a ruby managed object and is garbage collected. Only when the call on line 108 succeeds that object points to an opaques structure, which needs to be freed by calling the `ct_status_free`. But as the `ct_status_get_id` on line 109 should not raise - there is no failure mode documented for it: http://docs.oracle.com/cd/E86824_01/html/E54768/ct-status-get-id-3contract.html, the code should always free it on line 110.

lib/puppet/agent.rb
mruzicka: A question: What should we do on Solaris when Fiddle is not available or too old?
Should we:
1. Fallback to no-op? Rationale: It's how it used to work and the chances are hight the code won't be restarting the service and thus won't be affected by the problem we are trying to solve.
2. Raise an exception? Rationale: It either works in all scenarios as it supposed to or not at all ...

lib/puppet/agent.rb
joshcooper: True, but I think nested it makes it self-evident that there is a dependency between the open and close. If you have two independent begin/rescues then it's not obvious that they are in fact dependent.

lib/puppet/agent.rb
joshcooper: Ah right. That said I prefer explicitly managing the pointer lifetimes when interacting with native calls and not relying on garbage collection to deallocate memory, e.g.

```ruby
stathdl = Fiddle::Pointer.new(0)
begin
  raise if ct_status_read(stat.fileno, CTD_COMMON, stathdl.ref) != 0
  begin
    ...
  ensure
    ct_status_free(stathdl)
  end
ensure
  stathdl.free
end
```

FFI has a nice block form that handles that automatically, e.g. https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/api_types.rb#L25-L30, wish Fiddle had something similar.

mruzicka:
@joshcooper I'm no sure sure if we really always want to only log the exceptions instead of failing the execution, since for instance leaving the contract template active in the child process would cause every process executed by the child process to run in its own contract, my guess is there is probably quite a few of those (consider facter) and that could potentially lead to exhaustion of some limit on the total number of contracts. So I'd rather fail to surface the problem early.
On the other hand if we fail to active the new contract template it may be desirable to log the exception and continue as in that case the worst we risk is a return to the behavior before this fix.

lib/puppet/util/at_fork/solaris.rb
```diff
@@ -0,0 +1,155 @@
...
+  def get_latest_child_contract_id()
+    begin
+      stat = File.open(CTFS_PR_LATEST, File::RDONLY)
+    rescue => detail
+      Puppet.log_exception(detail, _('Unable to access latest child process contract status'))
+      return nil
+    end
+
+    begin
+      stathdl = Fiddle::Pointer.new(0)
```
mruzicka: Sorry if I'm stating the obvious, but it seems there is a misunderstanding here:
This does not allocate any bit of memory which would not be managed by ruby. It creates a Fiddle equivalent of something which would in C syntax read:
```c
void *stathdl = NULL;
```
the storage space for the `stathdl` itself is managed by ruby.
The following `ct_status_read` call resets the pointer (note the use `stathdl.ref` - a Fiddle equivalent of the `&stathdl` C syntax) to point to an opaque structure which can be further examined by the  `ct_status_get_*` family of calls and eventually must be freed by the `ct_status_free` call. OTOH there is no need to do anything special about the free the `stathdl` itself as it is garbage collected by ruby automatically when it goes out of scope.

joshcooper: Yeah, makes sense. I assumed that fiddle allocated native memory for the ptr itself, even when the address is 0x0, but it doesn't look like it behaves that way. I assumed it worked like FFI, which does allocate memory:

```ruby
static VALUE
memptr_initialize(int argc, VALUE* argv, VALUE self)
{
    VALUE size = Qnil, count = Qnil, clear = Qnil;
    int nargs = rb_scan_args(argc, argv, "12", &size, &count, &clear);

    memptr_malloc(self, rbffi_type_size(size), nargs > 1 ? NUM2LONG(count) : 1,
        RTEST(clear) || clear == Qnil);
```

About the NULL business, it might be clearer to do `stathdl = Fiddle::NULL`

mruzicka: I'm not sure you if you can reset the Fiddle::NULL pointer constant. But if you can I agree it's a clearer approach, I'll investigate.

mruzicka: It turns out you indeed can reset the `Fiddle::NULL` constant:
```ruby
require 'fiddle'
strtol = Fiddle::Function.new(Fiddle::Handle.new['strtol'], [Fiddle::TYPE_VOIDP, Fiddle::TYPE_VOIDP, Fiddle::TYPE_INT], Fiddle::TYPE_LONG)
a = '12xy'
b = Fiddle::NULL
puts "Fiddle::NULL address: #{Fiddle::NULL.to_i}"
puts "Calling strtol function ..."
c = strtol.call(Fiddle::Pointer.to_ptr(a), b.ref, 10)
puts "Fiddle::NULL address: #{Fiddle::NULL.to_i}"
puts "Integer value: #{c}"
puts "Unconverted string: #{b.to_s}"
```
which outputs:
```
Fiddle::NULL address: 0
Calling strtol function ...
Fiddle::NULL address: 140539989261538
Integer value: 12
Unconverted string: xy
```
Which is not something with what to do, so I'll stick with the current implementation.

lib/puppet/util/at_fork/solaris.rb
```diff
@@ -0,0 +1,155 @@
...
+
+  CT_PR_PGRPONLY = 0x4
+  CT_PR_EV_HWERR = 0x20
+
+  CTD_COMMON = 0
+
+  def raise_if_error()
+    unless (e = yield) == 0
+      e = SystemCallError.new(nil, e)
+      raise e, e.message, caller
```
joshcooper: A couple of ruby-isms. We don't use parentheses for empty argument lists, so just `def raise_if_error`. However, if a method yields to a block, then you'll want to put that in the argument list, `def raise_if_error(&block)`. Ruby doesn't require it, but it makes the method more readable.

lib/puppet/util/at_fork/solaris.rb
```diff
@@ -0,0 +1,155 @@
...
+        raise_if_error { ct_pr_tmpl_set_param(tmpl_fd, CT_PR_PGRPONLY) }
+        raise_if_error { ct_pr_tmpl_set_fatal(tmpl_fd, CT_PR_EV_HWERR) }
+        raise_if_error { ct_tmpl_set_critical(tmpl_fd, 0) }
+        raise_if_error { ct_tmpl_set_informative(tmpl_fd, CT_PR_EV_HWERR) }
+
+        raise_if_error { ct_tmpl_activate(tmpl_fd) }
+      rescue
+        tmpl.close()
+        raise
+      end
```
joshcooper: You could change that to just:

```ruby
ensure
  tmpl.close
end
```
So that you don't need to re-raise the exception.

lib/puppet/util/at_fork/solaris.rb
mruzicka: ok

lib/puppet/util/at_fork/solaris.rb
mruzicka: No, I can't, because I need the handle to remain open in case no exception is thrown. In that case the handle is stored in an instance variable and later closed in the `deactivate_contract_template` method.

lib/puppet/util/at_fork.rb
```diff
@@ -0,0 +1,35 @@
+require 'puppet'
+
+# A module for building AtFork handlers. These handlers are objects providing
+# pre/post fork callbacks modeled after those registered by the `pthread_atfork`
+# function.
+# Currently there are two AtFork handler implementations:
+# - a noop implementation used on all platforms except Solaris (and possibly
+#   even there as a fallback)
+# - a Solaris implementation which ensures the forked process runs in a different
+#   contract than the parent process. This is necessary for agent runs stared by
```
MikaelSmith: Typo: `stared` should be `started`.

lib/puppet/util/at_fork.rb
```diff
@@ -0,0 +1,35 @@
...
+#   the puppet agent service to be able to restart that service without being
+#   killed in the process as a consequence of running in the same contract as the
+#   service.
+module Puppet::Util::AtFork
+  @handler_class = loop do
```
MikaelSmith: This is a crazy idiom. I suppose it avoids writing and calling a new function just to get simple fall-through and assignment.

lib/puppet/util/at_fork.rb
```diff
@@ -0,0 +1,35 @@
...
+    if Facter.value(:operatingsystem) == 'Solaris'
```
MikaelSmith: Same day using Facter here may bite us, but we currently use it in settings parsing so this isn't the first use in the service.

lib/puppet/util/at_fork.rb
mruzicka: Yes, exactly so, I wish there was a method similar to `loop` but which would have called the block only once.
I realize this is not a common use of the `loop` method that's why I added the comments on the break statements used to return values from the construct.

lib/puppet/util/at_fork.rb
mruzicka: Is there a better way? Perhaps RUBY_PLATFORM?
@joshcooper?

lib/puppet/util/at_fork.rb
MikaelSmith: This is the only comment I expect a change for. The others were just observations.

lib/puppet/util/at_fork.rb
joshcooper: Thor vendored into bundler does that: https://github.com/bundler/bundler/blob/09358d459565c39314eea933020c8f848b3cfce0/lib/bundler/vendor/thor/lib/thor/shell/basic.rb#L378

One issue is it doesn't work in JRuby (but that doesn't apply here). Or maybe [`RbConfig::CONFIG['host_os']`](http://stackoverflow.com/questions/11784109/detecting-operating-systems-in-ruby)?

lib/puppet/util/at_fork/solaris.rb
```diff
@@ -0,0 +1,155 @@
...
+  def deactivate_contract_template(where)
+    return if @tmpl.nil?
+
+    tmpl = @tmpl
+    @tmpl = nil
+
+    begin
+      raise_if_error { ct_tmpl_clear(tmpl.fileno) }
+    rescue => detail
+      Puppet.log_exception(detail, _("Failed to deactivate process contract template in the #{where} process"))
```
mruzicka: @joshcooper I suspect this might cause problems to the i18n's translatable string extraction, effectively those are two distinct messages:
```
Failed to deactivate process contract template in the parent process
```
and
```
Failed to deactivate process contract template in the child process
```
but figuring that out is probably beyond the string extraction ability?
What would be correct approach?

mruzicka:
@joshcooper @MikaelSmith I believe I've addressed all of your comments. The tests are now also included. Note that tests only exercise the functionality of the 'Puppet::Util::AtFork' module, but not of the handlers returned by the module. This is because the handler is either a noop or a very platform specific code which is difficult to test in a generic way beyond verifying it logs the expected messages when some of the system calls fails. I'm happy to add those tests if you feel they provide enough value.

lib/puppet/util/at_fork/solaris.rb
joshcooper: Ah, thanks for catching that. You can do something like:

```ruby
_("Failed to ... %{where} process") % { where: where }
```

mruzicka: But then the translators will only see:
```
Failed to ... %{where} process
```
which may be more difficult to translate than the two distinct non-parametrized phrases.
What I had on my mind was something like:
```ruby
Puppet.log_exception(detail, parent? ? _("... parent process") : _("... child process"))
```
does that make sense?

mruzicka:
And the tests are passing now too ... yay!

lib/puppet/util/at_fork/solaris.rb
joshcooper: ah yeah, just the parameter to a boolean and make the two cases constants strings. 👍

mruzicka: k

mruzicka:
Rebased on latest stable.

MikaelSmith:
I'm running this through a quick acceptance suite at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_intn-van-sys_suite-ad-hoc-vmpooler-puppet-ad-hoc/28/. If it passes, I'll merge.

MikaelSmith:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_intn-van-sys_suite-ad-hoc-vmpooler-puppet-ad-hoc/28/SLAVE_LABEL=beaker,TEST_TARGET=solaris10-64a/testReport/junit/(root)/tests_direct_puppet/cached_catalog_remediate_local_drift_rb/

A single failure. Maybe not 100% reliable?

mruzicka:
The problem is the ruby built by this pipeline (ruby 2.1.9) doesn't include fiddle - search for fiddle in [console output of the packaging build #69](https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppet-agent_pkg-van-ship_ad-hoc-vmpooler-ad-hoc/BUILD_TARGET=solaris-11-i386,SLAVE_LABEL=beaker/69/consoleFull).
FIddle is only included in ruby 2.3.3 which only seems to by built by the master pipeline.

So the question is: Should I rebase the PR on the master branch or should I rather file a ticket to make sure Fiddle is included in ruby 2.1.9 which we ship with older versions of puppet-agent?

MikaelSmith:
I'd discuss it with @McdonaldSeanp maybe, I think he's done a fair amount of the Ruby 2.3 work. If it's not very hard, I'd rather get that into the ruby 2.1.9 line.

mruzicka:
Rebased on the master branch.

MikaelSmith:
Is the fact that Ruby 2.3 is not default on master going to block merging this ticket?

Magisus:
Ruby 2.3 IS currently the default on the master branch...

MikaelSmith:
Did I do something wrong to kick it off then? https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_pkg-van-ship_ad-hoc-vmpooler-ad-hoc/BUILD_TARGET=solaris-10-i386,SLAVE_LABEL=beaker/71/consoleFull used 2.1.9.

Magisus:
Looks like you might need to rebase your branch, the project config there has 2.1.9, but in master we have 2.3.3: https://github.com/puppetlabs/puppet-agent/blob/master/configs/projects/puppet-agent.rb#L94-L95

mruzicka:
Yesterday I kicked off this build: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_init-van-int_suite-ad-hoc-vmpooler-ad-hoc/87/
And it has completed successfully.

PR #5848, opened 2017-05-08
(PUP-5635) Prevent qualified var access from becoming locally resolved

hlindberg:
See commits and ticket.

### DIFF
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -602,23 +602,23 @@ class Puppet::Parser::Scope
     @compiler && equal?(@compiler.topscope)
   end
 
-  def lookup_qualified_variable(class_name, variable_name, position)
+  def lookup_qualified_variable(class_name, variable_name, options)
     begin
       if lookup_as_local_name?(class_name, variable_name)
         if is_topscope?
           # This is the case where $::x is looked up from within the topscope itself, or from a local scope
           # parented at the top scope. In this case, the lookup must ignore local and ephemeral scopes.
           #
-          handle_not_found(class_name, variable_name, position) unless @symtable.include?(variable_name)
+          handle_not_found(class_name, variable_name, options) unless @symtable.include?(variable_name)
           @symtable[variable_name]
         else
-          lookupvar(variable_name)
+          lookupvar(variable_name, options)
         end
       else
-        qualified_scope(class_name).lookupvar(variable_name, position)
+        qualified_scope(class_name).lookupvar(variable_name, options)
       end
     rescue RuntimeError => e
-      handle_not_found(class_name, variable_name, position, e.message)
+      handle_not_found(class_name, variable_name, options, e.message)
     end
   end
 
--- a/lib/puppet/parser/scope.rb
+++ b/lib/puppet/parser/scope.rb
@@ -29,7 +29,7 @@ class Puppet::Parser::Scope
 
   include Puppet::Util::Errors
   attr_accessor :source, :resource
-  attr_accessor :compiler
+  attr_reader :compiler
   attr_accessor :parent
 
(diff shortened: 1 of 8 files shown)


lib/puppet/parser/scope.rb
```diff
@@ -615,7 +628,7 @@ def lookup_qualified_variable(class_name, variable_name, options)
           lookupvar(variable_name, options)
         end
       else
-        qualified_scope(class_name).lookupvar(variable_name, options)
+        qualified_scope(class_name).lookupvar(variable_name, options.merge({ :fqn_leaf => true }))
```
thallgren: The handling of `:fqn_leaf` seems a bit expensive. Potentially, there will be two extra hash lookups and then a merge, which in turn creates a new copy of the options hash.
Given that this code is very performance critical (I recently optimized it for that reason), I'd suggest either making `fqn_leaf` a third optional argument to the `#lookupvar` method, or to extract it into a variable once at the beginning of the method and then use that variable to avoid repeated hash lookups and also to make the merge optional.

lib/puppet/parser/scope.rb
```diff
@@ -560,7 +560,7 @@ def variable_not_found(name, reason=nil)
   # @see #[]=
   # @api public
   #
-  def [](varname, options={})
+  def [](varname, options = EMPTY_HASH)
```
thallgren: This change can be done to the `#initialize` method too, now that `:namespace` out of the way.

lib/puppet/parser/scope.rb
```diff
@@ -798,6 +798,19 @@ def setvar(name, value, options = EMPTY_HASH)
     else
       table[name] = value
     end
+    # Assign the qualified name in the environment
+    # Note that Settings scope sets source to Boolean true.
+    #
+    # require 'byebug'; debugger if name == "fooo"
```
thallgren: Please remove this comment

lib/puppet/parser/scope.rb
```diff
@@ -798,6 +798,19 @@ def setvar(name, value, options = EMPTY_HASH)
...
+    if source = self.source.is_a?(Puppet::Resource) && self.source.type == :hostclass
+      sourcename = source.name
```
thallgren: Unless I'm mistaking, the line above this one, assigns a boolean value to `self.source` (it's an `attr_accessor` so there's no difference between `source` and `self.source`). If that boolean is `true`, it's now asked for a name. That cannot work so I assume that there's no test that covers this piece of code.

hlindberg: uhoh  - commited wrong code - that should not have been there at all

lib/puppet/parser/scope.rb
```diff
@@ -579,47 +580,51 @@ def inherited_scope
...
-        if is_topscope?
-          # This is the case where $::x is looked up from within the topscope itself, or from a local scope
-          # parented at the top scope. In this case, the lookup must ignore local and ephemeral scopes.
-          #
-          handle_not_found(class_name, variable_name, position) unless @symtable.include?(variable_name)
-          @symtable[variable_name]
-        else
-          self[variable_name]
+  # @api private
+  def lookup_qualified_variable(fqn, options)
```
thallgren: This change isn't backward compatible. The method used to be public and the signature changes. Consider keeping the old method with a deprecation warning and name the new method something else.

hlindberg: The Scope class has a general message at the top that none of the methods are public except where explicitly stated that it is api. Maybe we should now mark all of them as private. (I can naturally also change the name of the method in question).

thallgren: I missed the message at the top. That's good enough IMO.

PR #5850, opened 2017-05-09
(PUP-7503) Fix hostname mismatch error reporting with Ruby 2.4

Magisus:
Ruby 2.4 introduced a change to OpenSSL that adds a step to
certificate preverification that checks that the hostname matches what
is in the certificate. While this is not enabled by default for most use
cases, any use of `set_params` to configure OpenSSL will enable it.
In the past, the hostname was verified in a post-connection verification
step and produced a specific error message. Now it simply emits a generic
SSL error. Since we still want to give a more informative error message,
this commit changes our error handling to explicitly check for a hostname
mismatch when we receive an SSL error and emit our custom descriptive
message. Note that in Ruby 2.4, this duplicates work already done by Ruby,
but it allows us to be consistent with our error reporting.

### DIFF
--- a/lib/puppet/ssl/validator/default_validator.rb
+++ b/lib/puppet/ssl/validator/default_validator.rb
@@ -69,12 +69,12 @@ class Puppet::SSL::Validator::DefaultValidator #< class Puppet::SSL::Validator
   # @api private
   #
   def call(preverify_ok, store_context)
+    current_cert = store_context.current_cert
+    @peer_certs << Puppet::SSL::Certificate.from_instance(current_cert)
+
     # We must make a copy since the scope of the store_context will be lost
     # across invocations of this method.
     if preverify_ok
-      current_cert = store_context.current_cert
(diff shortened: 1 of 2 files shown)


lib/puppet/ssl/validator/default_validator.rb
```diff
@@ -100,7 +100,6 @@ def call(preverify_ok, store_context)
           @verify_errors << error_string
         end
       else
-        current_cert = store_context.current_cert
```
Magisus: I discovered a strange thing happening here in our unit tests, where apparently at least with the way we are mocking this (https://github.com/Magisus/puppet/blob/master/spec/unit/ssl/validator_spec.rb#L35) if you query `store_context.current_cert` twice, it returns a different thing the second time (maybe the next one in the list?). I wasn't sure why this was happening, but removing the duplicate call fixed it. It seemed wrong to set `current_cert` twice anyway.

Magisus:
This passed a run of the full acceptance suite, fixing the original failing test and not introducing any new failures.

PR #5851, opened 2017-05-10
(PUP-7518) i18n module loading

eputnam:
This adds to Puppet::Module to "initialize" a module for translations. Each module must be added to FastGettext.translation_repositories, so that translations in the modules code will be available to Puppet. Two methods are added here, one to check if a module has been initialized and another to initialize. To initialize, we are just wrapping GettextSetup.initialize basically.

### DIFF
--- a/lib/puppet/module.rb
+++ b/lib/puppet/module.rb
@@ -68,6 +68,17 @@ class Puppet::Module
     load_metadata
 
     @absolute_path_to_manifests = Puppet::FileSystem::PathPattern.absolute(manifests)
+
+    # i18n initialization for modules
+    if Puppet::GETTEXT_AVAILABLE
+      begin
+        initialize_i18n
+      rescue Exception => e
+        Puppet.warning _("GettextSetup initialization for %{module_name} failed with: %{error_message}") % { module_name: name, error_message: e.message }
+      end
+    else
+      Puppet.warning _("GettextSetup is not available, skipping GettextSetup initialization for %{module_name}.") % { module_name: name }
+    end
   end
 
   # @deprecated The puppetversion module metadata field is no longer used.
@@ -353,7 +364,26 @@ class Puppet::Module
     self.environment == other.environment
   end
 
+  def initialize_i18n
+    module_name = @forge_name.gsub("/","-") if @forge_name
+    return if module_name.nil? || i18n_initialized?(module_name)
+
+    locales_path = File.absolute_path('locales', path)
+
+    begin
+      GettextSetup.initialize(locales_path)
+      Puppet.debug "#{module_name} initialized for i18n: #{GettextSetup.translation_repositories[module_name]}"
+    rescue
+      config_path = File.absolute_path('config.yaml', locales_path)
+      Puppet.debug "Could not find locales configuration file for #{module_name} at #{config_path}. Skipping i18n initialization."
+    end
+  end
+
   private
+
+  def i18n_initialized?(module_name)
+    GettextSetup.translation_repositories.has_key? module_name
+  end
 
   def wanted_manifests_from(pattern)
     begin
--- a/spec/unit/module_spec.rb
+++ b/spec/unit/module_spec.rb
@@ -440,6 +440,35 @@ describe Puppet::Module do
     end
   end
 
+
+  describe "initialize_i18n" do
+
+    let(:modpath) { tmpdir('modpath') }
+    let(:modname) { 'puppetlabs-i18n'}
+    let(:modroot) { "#{modpath}/#{modname}/" }
+    let(:config_path) { "#{modroot}/locales/config.yaml" }
+    let(:mod_obj) { PuppetSpec::Modules.create( modname, modpath, :metadata => { :dependencies => [] }, :env => env ) }
+
+    it "is expected to initialize an un-initialized module" do
+      expect(GettextSetup.translation_repositories.has_key? modname).to be false
+
+      FileUtils.mkdir_p("#{mod_obj.path}/locales")
+      config = {
+        "gettext" => {
+          "project_name" => modname
+        }
+      }
+      File.open(config_path, 'w') { |file| file.write(config.to_yaml) }
+
+      mod_obj.initialize_i18n
+
+      expect(GettextSetup.translation_repositories.has_key? modname).to be true
+    end
+    it "is expected return nil if module is intiailized" do
+      expect(mod_obj.initialize_i18n).to be nil
+    end
+  end
+
   describe "when managing supported platforms" do
     it "should support specifying a supported platform" do
       mod.supports "solaris"


Iristyle:
@eputnam the ticket here should be moved over to PUP, and commit message should be updated to correspond to the PUP ticket.  We don't use MODULES tickets inside of Puppet.


Thanks!

Iristyle:
Pinged @Magisus for a review as she's the resident gettext ninja.

lib/puppet/module.rb
```diff
@@ -67,6 +67,16 @@ def initialize(name, path, environment)
     load_metadata
 
     @absolute_path_to_manifests = Puppet::FileSystem::PathPattern.absolute(manifests)
+
+    # # i18n initialization for modules
```
glennsarti: nit: double hash

lib/puppet/module.rb
```diff
@@ -67,6 +67,16 @@ def initialize(name, path, environment)
...
+    begin
+      require 'gettext-setup'
+      require 'locale'
+
+      initialize_i18n
+    rescue
```
glennsarti: It's nice to actually output the error message too

lib/puppet/module.rb
```diff
@@ -351,6 +361,25 @@ def ==(other)
     self.environment == other.environment
   end
 
+  def initialize_i18n
+    locales_path = File.absolute_path('locales', path)
+    config_path = File.absolute_path('config.yaml', locales_path)
+    module_name = @forge_name.gsub("/","-") if @forge_name
+
+    nil if initialized_i18n? module_name
```
glennsarti: Should this be `return nil`?

eputnam:
i may also add gettext config creation [here](https://github.com/puppetlabs/puppet/blob/master/spec/lib/puppet_spec/modules.rb) in the future to reduce test clutter

lib/puppet/module.rb
```diff
@@ -67,6 +67,16 @@ def initialize(name, path, environment)
...
     @absolute_path_to_manifests = Puppet::FileSystem::PathPattern.absolute(manifests)
+
+    # # i18n initialization for modules
+    begin
+      require 'gettext-setup'
+      require 'locale'
+
+      initialize_i18n
+    rescue
+      Puppet.warning "GettextSetup initialization for #{@name} failed."
```
glennsarti: I assume this cannot be externalised as Gettext isn't setup yet.

spec/unit/module_spec.rb
```diff
@@ -440,6 +440,50 @@
...
 
+
+  describe "i18n" do
+
+    let(:modpath) { tmpdir('modpath') }
+    let(:modname) { 'puppetlabs-i18n'}
+
+    before do
+      modroot = "#{modpath}/#{modname}/"
+      @config_path = "#{modroot}/locales/config.yaml"
```
glennsarti: nit: These variables should probably all be `let` definitions.  None of them are used in `before(:after)` which is usually why `@` variables are used.

lib/puppet/module.rb
```diff
@@ -351,6 +361,25 @@ def ==(other)
...
+
+    begin
+      GettextSetup.initialize(locales_path) unless !File.file? config_path
+      Puppet.debug _("%{value0,} initialized for i18n: %{value1,}") % { value0: module_name, value1: GettextSetup.translation_repositories[module_name]}
+    rescue
+      Puppet.debug _("Could not find locales configuration file for %{value0,} at %{value1,}. Skipping i18n initialization.") % { value0: module_name, value1: config_path }
+    end
+  end
+
+  def initialized_i18n? module_name
```
glennsarti: nit: It's probably more readable as `i18n_initialized?`  e.g. "Is the door closed" vs "have you closed the door?"

lib/puppet/module.rb
```diff
@@ -351,6 +361,25 @@ def ==(other)
...
+  def initialize_i18n
+    locales_path = File.absolute_path('locales', path)
+    config_path = File.absolute_path('config.yaml', locales_path)
+    module_name = @forge_name.gsub("/","-") if @forge_name
+
+    nil if initialized_i18n? module_name
+
+    begin
+      GettextSetup.initialize(locales_path) unless !File.file? config_path
+      Puppet.debug _("%{value0,} initialized for i18n: %{value1,}") % { value0: module_name, value1: GettextSetup.translation_repositories[module_name]}
```
Magisus: We have been asked at least for now to not externalize debug messages.

lib/puppet/module.rb
Magisus: +1 to name change

lib/puppet/module.rb
Magisus: So I think since this is going to use Puppet's GettextSetup and not the module's, it can be. And it should be. We are trying to externalize all warnings and errors.

lib/puppet/module.rb
Magisus: Also for future reference, please try to replace the `value0` etc. with meaningful names. This will make it easier for the translator. Also as written, the commas will cause a mismatch of keys and this will explode.

Magisus:
Are modules with externalized strings able to function normally if GettextSetup is not present? See https://github.com/puppetlabs/puppet/blob/master/lib/puppet.rb#L25-L45 for how we are handling this in Puppet.

lib/puppet/module.rb
eputnam: agreed.

lib/puppet/module.rb
eputnam: explicit returns in Ruby are frowned upon?

lib/puppet/module.rb
eputnam: good catch, it should have been but i guess we aren't externalizing yet.

lib/puppet/module.rb
Magisus: Only at the ends of methods. This one is in the middle, and we _want_ to quit if this is true.

eputnam:
we haven't started externalizing yet fortunately. this is a great point @Magisus , thank you!

Magisus:
This is blocked on more extensive testing.

Magisus:
Also if this functionality is going to be needed/supported in the LTS, this needs to be retargeted at stable.

Magisus:
Closing and reopening to fix CI

Magisus:
Sorry to shuffle this again, but would you mind retargeting this at the 1.10.x branch? We're going to be deleting the stable branch in favor of that one soon.

eputnam:
Would like to get this merged before the next release. @Iristyle have you had a chance to take a look?

joshcooper:
Are the translations namespaced so that translations in the module(s) don't affect puppet, and vice-versa? Or do we need to rely/assume that the keys are unique across puppet and all modules?

eputnam:
they are namespaced inside the master_domain by "project_name" from config.yaml so there should not be any interference.

lib/puppet/module.rb
```diff
@@ -353,8 +364,27 @@ def ==(other)
...
+      GettextSetup.initialize(locales_path)
+      Puppet.debug "#{module_name} initialized for i18n: #{GettextSetup.translation_repositories[module_name]}"
+    rescue
+      Puppet.debug "Could not find locales configuration file for #{module_name} at #{config_path}. Skipping i18n initialization."
+    end
+  end
+
   private
 
+  def i18n_initialized? module_name
```
Iristyle: nit: Please use parens in method declaration

lib/puppet/module.rb
```diff
@@ -68,6 +68,17 @@ def initialize(name, path, environment)
     load_metadata
 
     @absolute_path_to_manifests = Puppet::FileSystem::PathPattern.absolute(manifests)
+
+    # i18n initialization for modules
+    if Puppet::GETTEXT_AVAILABLE
```
Iristyle: `GETTEXT_AVAILABLE` as a top-level constant for a feature test feels a little weird.  The ship has probably sailed on this decision, but wondering why this isn't in a feature?

lib/puppet/module.rb
```diff
@@ -353,8 +364,27 @@ def ==(other)
     self.environment == other.environment
   end
 
+  def initialize_i18n
+    locales_path = File.absolute_path('locales', path)
+    config_path = File.absolute_path('config.yaml', locales_path)
+    module_name = @forge_name.gsub("/","-") if @forge_name
+
+    return nil if i18n_initialized? module_name
```
Iristyle: nit: can just `return` rather than `return nil` given the return value of `initialize_i18n` is not used

Iristyle: I'm also not sure how the method is designed to fail here?  This method rescues ... and so does its caller. Can probably streamline the error-handling a bit.

lib/puppet/module.rb
Magisus: Loading gettext happens very early, and I seem to remember there being a reason we couldn't use Puppet features because of that. If that's not the case, this CAN probably change; the constant is only used internally.

lib/puppet/module.rb
eputnam: it's not designed to fail, it just warns that there won't be any translations. i didn't think it was worth blocking a puppet run because translations aren't running. it's not affecting system configuration or anything, just the language of the log output, which will be obvious.

also, i like your version of the function a lot. thanks for the review!

eputnam:
perhaps a little blurb above the `initialize_i18n` function is in order also? 

>   Wraps GettextSetup.initialize which adds a translation_repository to the master text_domain for a module. For successful initialization, GettextSetup requires a locales/ dir and a locales/config.yaml file. If the module already has a translation_repository in the master text_domain or the module has no name, this function returns nil.

eputnam:
doing some git shuffling now, don't merge yet.

eputnam:
ok, canoodling over, ready for merge.

Magisus:
I would go ahead and squash those last two commits. They're not really logically separable.

lib/puppet/module.rb
Iristyle: Fair enough... that sounds like a plausible explanation.

lib/puppet/module.rb
```diff
@@ -68,6 +68,17 @@ def initialize(name, path, environment)
     load_metadata
 
     @absolute_path_to_manifests = Puppet::FileSystem::PathPattern.absolute(manifests)
+
+    # i18n initialization for modules
+    if Puppet::GETTEXT_AVAILABLE
+      begin
+        initialize_i18n
```
Iristyle: What I meant here @eputnam is that if `initialize_i18n` never `raises` an exception based on its code being inside a `begin` / `rescue` ... that means there's no need for a `rescue` here.

Merged for now, but this could probably be tightened up a little.

PR #5852, opened 2017-05-10
(PUP-7516) Fix Fedora25 createrepo failures

Iristyle:
- On 5/3/2017 version 7.43.0-5.fc25 of python-pycurl was released to
   address issue https://bugzilla.redhat.com/show_bug.cgi?id=1446850

```
   ImportError: pycurl: libcurl link-time ssl backend (openssl) is different from compile-time ssl backend (nss)
```

 - Puppet performs some package tests with yum and dnf providers by
   installing / using createrepo.  Createrepo requires python and a
   number of python dependencies, including the python-pycurl bindings
   for curl.  These bindings depend on the curl package.

   Unfortunately on Fedora 25, the baseline curl package installed on
   images is curl-7.50.3-1.fc25.x86_64. This curl package in
   conjunction with the newer python-pycurl 7.43.0-5.fc25 will
   generate an error when calling createrepo:

```
   ImportError: pycurl: libcurl link-time version (7.50.3) is older than compile-time version (7.51.0)
```

 - One solution to this problem is to downgrade the python-pycurl
   package to the older working version with:

```
   yum downgrade python-pycurl-7.43.0-4.fc25
```

 - Another solution would be to not mirror the newer package.

 - Given both of these options are brittle, it's been decided that
   when installing createrepo, curl should be upgraded at the same
   time. In testing, the newer curl 7.51.0-6.fc25 has been found to
   be compatible with the updated python-pycurl, allowing createrepo
   to function again without crashing.

 - To be able to upgrade curl, which may already be installed on a
   target machine, requires that the `--best` parameter be passed
   to yum / dnf install (both package managers have the feature). It
   also requires memoizing whether or not the package has been
   previously installed in Puppet::Acceptance::RpmUtils#setup_rpm.
   That method currently checked if a package was installed and would
   skip installing it if it was already present. Since that prevents a
   desirable upgrade of curl, make it track whether or not an install
   / upgrade has been attempted.

   8c1751e defined the original setup
   method which is modified here

 - On Fedora 25, Bundler will outright crash as it has a hard
   dependency on io/console by way of the Thor CLI library.

   Bundler 1.14-6 vendors Thor 0.19.1

   https://github.com/bundler/bundler/blob/v1.14.6/lib/bundler/vendor/thor/lib/thor/version.rb
   https://github.com/erikhuda/thor/blob/v0.19.1/lib/thor/shell/basic.rb#L2

   The hard dependency has been removed in an as of yet unshipped
   version of Thor via erikhuda/thor#539
   meaning that at some point it will not be necessary to have
   io/console installed for Bundler.

   For now, it remains a hard dependency,
   without which calling Bundler will fail with an error like:

```
 /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- io/console (LoadError)
  from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
  from /usr/local/share/gems/gems/bundler-1.14.6/lib/bundler/vendor/thor/lib/thor/shell/basic.rb:2:in `<top (required)>'
  from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
  from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
  from /usr/local/share/gems/gems/bundler-1.14.6/lib/bundler/vendor/thor/lib/thor/shell/color.rb:1:in `<top (required)>'
  from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
  from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
  from /usr/local/share/gems/gems/bundler-1.14.6/lib/bundler/vendor/thor/lib/thor/shell.rb:17:in `shell'
  from /usr/local/share/gems/gems/bundler-1.14.6/lib/bundler/ui/shell.rb:15:in `initialize'
  from /usr/local/share/gems/gems/bundler-1.14.6/lib/bundler/cli.rb:13:in `new'
  from /usr/local/share/gems/gems/bundler-1.14.6/lib/bundler/cli.rb:13:in `rescue in start'
  from /usr/local/share/gems/gems/bundler-1.14.6/lib/bundler/cli.rb:17:in `start'
  from /usr/local/share/gems/gems/bundler-1.14.6/exe/bundle:32:in `block in <top (required)>'
  from /usr/local/share/gems/gems/bundler-1.14.6/lib/bundler/friendly_errors.rb:121:in `with_friendly_errors'
  from /usr/local/share/gems/gems/bundler-1.14.6/exe/bundle:24:in `<top (required)>'
  from /usr/local/bin/bundle:23:in `load'
  from /usr/local/bin/bundle:23:in `<main>'
```

### DIFF
--- a/acceptance/lib/puppet/acceptance/rpm_util.rb
+++ b/acceptance/lib/puppet/acceptance/rpm_util.rb
@@ -3,6 +3,7 @@ module Puppet
     module RpmUtils
       # Utilities for creating a basic rpm package and using it in tests
       @@defaults = {:repo => '/tmp/rpmrepo', :pkg => 'mypkg', :publisher => 'tstpub.lan', :version => '1.0'}
+      @@setup_packages = {}
 
       def rpm_provider(agent)
         has_dnf = on(agent, 'which dnf', :acceptable_exit_codes => [0,1]).exit_code
@@ -15,10 +16,11 @@ module Puppet
 
       def setup(agent)
         cmd = rpm_provider(agent)
-        required_packages = ['createrepo', 'rpm-build']
+        required_packages = ['createrepo', 'curl', 'rpm-build']
         required_packages.each do |pkg|
-          unless ((on agent, "#{cmd} list installed #{pkg}", :acceptable_exit_codes => (0..255)).exit_code == 0) then
-            on agent, "#{cmd} install -y #{pkg}"
+          unless @@setup_packages.has_key?(pkg) && ((on agent, "#{cmd} list installed #{pkg}", :acceptable_exit_codes => (0..255)).exit_code == 0) then
+            on agent, "#{cmd} install -y #{pkg} --best"
+            @@setup_packages[pkg] = true
           end
         end
       end
--- a/acceptance/setup/git/pre-suite/000_EnvSetup.rb
+++ b/acceptance/setup/git/pre-suite/000_EnvSetup.rb
@@ -12,6 +12,7 @@ PACKAGES = {
     'git',
     'ruby',
     'rubygem-json',
+    'rubygem-io-console'
   ],
   :debian => [
     ['git', 'git-core'],


MosesMendoza:
will merge pending travis/appveyor

Iristyle:
Waiting on the ad-hoc run at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_intn-van-sys_suite-ad-hoc-vmpooler-puppet-ad-hoc/ to finish.

I tried to run `ci:test:git` over Fedora on just this test with `GEM_SOURCE=http://rubygems.delivery.puppetlabs.net/ TEST_TARGET=fedora25-64a PUPPET_FORK='Iristyle' bundle exec rake ci:test:git SHA=`git lastref` TEST=./tests/resource/package/yum.rb` ... but bundler blows up on Fedora25.

Magisus:
👍

acceptance/lib/puppet/acceptance/rpm_util.rb
```diff
@@ -15,10 +16,11 @@ def rpm_provider(agent)
 
       def setup(agent)
         cmd = rpm_provider(agent)
-        required_packages = ['createrepo', 'rpm-build']
+        required_packages = ['createrepo', 'curl', 'rpm-build']
```
Iristyle: There is a larger question around whether or not Beaker should be upgrading `curl` - it performs a version check at the beginning of a run with https://github.com/puppetlabs/beaker/blob/8f660718f971c6663e28bf6b5cbf3d26157beae0/lib/beaker/host_prebuilt_steps.rb#L123-L124

Any thoughts @kevpl ?

acceptance/setup/git/pre-suite/000_EnvSetup.rb
```diff
@@ -12,6 +12,7 @@
     'git',
     'ruby',
     'rubygem-json',
+    'rubygem-io-console'
```
Iristyle: * Should `rubygem-io-console` already be baked into images? Thoughts @jcoconnor?
* Is `rubygem-io-console` available on all `:redhat` (RHEL) based distros?  I believe we're mostly concerned here with Fedora, but there will be some variation across different RHEL versions and the Ruby versions that they want to install by default as `ruby`

acceptance/lib/puppet/acceptance/rpm_util.rb
kevpl: @Iristyle similarly to the code you changed here, I think the code you referenced in beaker is doing an existence check and installing if it's not there, not a version check.

I think the utilities provided here are [used seldom enough](https://github.com/search?utf8=✓&q=org%3Apuppetlabs+send_rpm+extension%3Arb&type=Code&ref=advsearch&l=&l=) that they alone shouldn't create enough need for beaker to always be checking against the right versions, especially since for most things, package existence is all we really need.

Beaker does provide the ability to upgrade packages with:
- [Beaker::Host's `upgrade_package`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix/pkg.rb#L195)
- [DSL convenience method: `upgrade_package`](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/dsl/helpers/host_helpers.rb#L411)

So in general I would recommend using those for upgrades, but this file is already low-level enough that you might as well just skip to making the package install call yourself, as you are.

Another option would be to specify in the beaker code you referenced the specific versioned package of curl for fedora, like freebsd does [here](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host_prebuilt_steps.rb#L16) (although they allow a backup of the unversioned package).

acceptance/setup/git/pre-suite/000_EnvSetup.rb
jcoconnor: @Iristyle Do you mean is ```rubygem-io-console``` already in the image, or should we add it as part of the image creation ?

Should be no issue adding it if needed.

Iristyle: @jcoconnor nevermind on adding it to images - that doesn't make sense given images don't have Ruby and we're also installing Ruby as part of the setup in https://github.com/puppetlabs/puppet/blob/9ccdd7e690033ec93e87b9a8e2ab853f0c21bd7d/acceptance/setup/git/pre-suite/000_EnvSetup.rb#L10-L40

PR #5854, opened 2017-05-10
(PUP-7149) Add job_id to catalog compilation and reports

adreyer:
This allows the job_id to be set when the agent is run directly. This
will be used to allow pxp_module_puppet to set job_id and
transaction_uuid for agent runs it initiates.

### DIFF
--- a/spec/unit/configurer_spec.rb
+++ b/spec/unit/configurer_spec.rb
@@ -199,6 +199,17 @@ describe Puppet::Configurer do
       @agent.run
     end
 
+    it "should create report with passed transaction_uuid and job_id" do
+      @agent = Puppet::Configurer.new(Puppet::Configurer::DownloaderFactory.new, "test_tuuid", "test_jid")
+      @agent.stubs(:init_storage)
+
+      report = Puppet::Transaction::Report.new('apply', nil, "test", "aaaa")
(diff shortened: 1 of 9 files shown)


spec/unit/configurer_spec.rb
```diff
@@ -199,6 +199,17 @@
...
+      @agent = Puppet::Configurer.new(Puppet::Configurer::DownloaderFactory.new, "test_tuuid", "test_jid")
+      @agent.stubs(:init_storage)
+
+      report = Puppet::Transaction::Report.new('apply', nil, "test", "aaaa")
+      Puppet::Transaction::Report.expects(:new).with(anything, nil, anything, 'test_tuuid', 'test_jid').returns(report)
+      @agent.expects(:send_report).with(report)
+
+      @agent.run
+    end
+
```
MikaelSmith: I think we also want a spec that the job_id is part of indirection requests for the catalog (there's a similar test for transaction_uuid).

MikaelSmith:
Realized in discussion with Alex that we don't put the job_id in the catalog response sent back to the agent, so I don't think that schema needs updated. The changes to PuppetDB in https://github.com/puppetlabs/puppetdb/pull/2274 result in associating it with the catalog in PuppetDB as intended.

So just the docs need updates: reports for the new content, catalog for the optional job_id request parameter.

MikaelSmith:
Do the PDB changes need to go in first?

PR #5855, opened 2017-05-11
(PUP-7519) Enabling rubocop securitycop scans

jsane:
Background: Rubocop scans were not running as the various cops in .rubocop.yml were not kept uptodate with the latest definitions. This will enable only SecurityCop/YAMLLoad as a start to revive the scans in an effort to flag any inadvertent usages.  Additional security or other cop rules can/would be added in future.   
Rubocop do not have support for flagging YAML.load_file which would have been desirable. Likewise it is not necessary to mark the known instance of Yaml.load_file in Puppet::Util::Yaml to be ignored.

Note that there are a couple of other YAML.load instances elsewhere so this may generate violations. If that is not desired this PR be merged after all these have been resolved/changed appropriately.


jsane:
Huh! looks like the rubocop version used in Travis does not recognize Security/YAMLLoad(?). Will figure and push the needed changes again.

jsane:
Gemfile had to be updated to use the latest rubocop gem as the Security Cop being used seems to be available only in 0.48.xx onwards (in a more later version than 0.39.0). 
As noted above, this will cause builds to fail in the current state due to use of YAML.load at a couple of places.

joshcooper:
Thanks for getting this working @jsane. Unfortunately if we merge this, then all subsequent PRs will fail. So I'd like to resolve the usages of unsafe `YAML.load`. Ideally those places should be modified to use our safe yaml load methods, so instead of:

```ruby
    if config = YAML.load(text)
```

Instead do:

```ruby
config = Puppet::Resource::Catalog.from_data_hash(Puppet::Util::Yaml.load(text))
```

The `Puppet::Util::Yaml` module needs to be modified to allow safely parsing of in-memory strings, as opposed to only files.

@thallgren could you take a look at this as part of the safe yaml epic?

jsane:
@thallgren Understandably this might not be high on priority wrt other things but wanted to know if there are plans to take this up anytime in coming future.  /cc @joshcooper

thallgren:
@josh, I actually created a (maint) PR a couple of weeks ago that upgrades .rubocop to 0.49.  See https://github.com/puppetlabs/puppet/pull/6088. I think that should go in first. Then sure, enabling a check for Yaml load make sense.

Gemfile
```diff
@@ -50,7 +50,7 @@ group(:development, :test) do
   gem "multi_json", "1.7.7", :require => false, :platforms => [:ruby, :jruby]
   gem "json-schema", "2.1.1", :require => false, :platforms => [:ruby, :jruby]
 
-  gem "rubocop", "~> 0.39.0", :platforms => [:ruby]
+  gem "rubocop", "~> 0.48.1", :platforms => [:ruby]
```
thallgren: Please update this to 0.49.1.

thallgren:
Another problem that needs to be circumvented somehow is that newer versions of rubocop requires Ruby >= 2.0.0. When can we drop support for 1.9.3?

jsane:
I think we configured .travis to skip rubocop scans in ruby 1.9.3

kris-bosland:
Hi @jsane, with the 5.5.x branch being cut, please retarget that branch.  Thanks!

joshcooper:
@jsane this needs a rebase, and let's continue targeting master (after @thallgren PR lands)

jsane:
Ack. I will target master per your comment in Thomas's PR.

jsane:
Created a new PR: https://github.com/puppetlabs/puppet/pull/6732. Closing this one - decided to started with a new branch instead of dealing with rebasing old set of changes.

PR #5856, opened 2017-05-11
(PUP-7156) Fix Puppet::Acceptance::RpmUtils#setup

Iristyle:
#### (PUP-7156) Fix Puppet::Acceptance::RpmUtils#setup

- An assumption was made in 2a8ee90
   that all versions of yum and dnf have a --best option during
   install.

   It turns out this isn't accuracte, so yum / dnf install cannot be
   used to both install and upgrade.

   Separate install / upgrade into two discrete options

 - Furthermore, state tracking in RpmUtils for upgraded packages was
   being managed across all agents instead of on an agent by agent
   basis.

 - Note that depending on `yum install` or `yum upgrade` actions the
   yum cache may be considered current, so that when a new repo is
   added to /etc/yum.repos.d that it may not actually be queried
   despite having new content.

#### Fix Fedora ci:test:git via rubygem-rdoc

 - Some gem install tests require rubygem-rdoc
 - Added note about why ci:test:git won't work on RHEL6

Fixes #5852

### DIFF
--- a/acceptance/lib/puppet/acceptance/rpm_util.rb
+++ b/acceptance/lib/puppet/acceptance/rpm_util.rb
@@ -15,13 +15,23 @@ module Puppet
       end
 
       def setup(agent)
+        @@setup_packages[agent] ||= {}
         cmd = rpm_provider(agent)
         required_packages = ['createrepo', 'curl', 'rpm-build']
         required_packages.each do |pkg|
-          unless @@setup_packages.has_key?(pkg) && ((on agent, "#{cmd} list installed #{pkg}", :acceptable_exit_codes => (0..255)).exit_code == 0) then
-            on agent, "#{cmd} install -y #{pkg} --best"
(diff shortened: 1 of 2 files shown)


Iristyle:
Ad-hoc run triggered at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_init-van-int_suite-ad-hoc-vmpooler-ad-hoc/78/

Iristyle:
Some notes about the fixes here - https://gist.github.com/Iristyle/e7c9ff357e809545f1fe8f69bc775e99#file-pass-vs-fail-diff

Magisus:
👍

McdonaldSeanp:
👍

Iristyle:
Oops, commits were tagged PUP-7156 but should have been PUP-7516

PR #5859, opened 2017-05-11
(maint) correct apply_catalog code comment

MosesMendoza:
Once upon a time Puppet::Configurer#apply_catalog would retrieve a catalog if
not supplied, but this is no longer the case. Update the code comment to
reflect reality because its confusing.

### DIFF
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -160,8 +160,7 @@ class Puppet::Configurer
     nil
   end
 
-  # Retrieve (optionally) and apply a catalog. If a catalog is passed in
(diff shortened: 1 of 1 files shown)


joshcooper:
Can we target master for this?

MosesMendoza:
Sure! Mind if I ask why though? Seems like a trivial/innocuous thing... definitely applies in stable too?

joshcooper:
I prefer master since targeting stable adds noise to that branch (more builds, more ticket reconciliation), but doesn't affect outcome like tests, rake tasks, etc do

PR #5860, opened 2017-05-11
(MAINT) IMAGES-555 Tag Tests that break Image Acceptance

jcoconnor:
These tests break the Image Acceptance Test suite, so are being tagged
so that they can be excluded pending further investigation as part of
ticket PUP-7521

This was labelelled IMAGES-555 but Travis is unable to accept this as PR
prefix.

### DIFF
--- a/acceptance/tests/agent/agent_disable_lockfile.rb
+++ b/acceptance/tests/agent/agent_disable_lockfile.rb
@@ -1,4 +1,5 @@
 test_name "the agent --disable/--enable functionality should manage the agent lockfile properly"
+tag
 confine :except, :platform => 'cisco_nexus' #See BKR-749
 
 #
--- a/acceptance/tests/agent/fallback_to_cached_catalog.rb
+++ b/acceptance/tests/agent/fallback_to_cached_catalog.rb
@@ -1,4 +1,5 @@
 test_name "fallback to the cached catalog"
+tag
 
 step "run agents once to cache the catalog" do
   with_puppet_running_on master, {} do
--- a/acceptance/tests/aix/aix_package_provider.rb
+++ b/acceptance/tests/aix/aix_package_provider.rb
@@ -1,4 +1,5 @@
 test_name "aix package provider should work correctly"
+tag
 
 confine :to, :platform => /aix/
 
--- a/acceptance/tests/aix/nim_package_provider.rb
+++ b/acceptance/tests/aix/nim_package_provider.rb
@@ -1,4 +1,5 @@
 test_name "NIM package provider should work correctly"
+tag
 
 confine :to, :platform => "aix"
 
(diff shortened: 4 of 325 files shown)


johnduarte:
Hrm. The Beaker tag facility does not operate in the manner in which I anticipated. When testing the exclusion of these tests based on the tag, the tests are not skipped as expected.
Here is an example run where I have applied your patch and then removed all other tests except for 1. My expectation is that the one unaltered test should run and the other 4 with the 'broken:image' tag should be skipped. However, the result is that all 5 tests are executed.

Here is the list of tests directory
```
tests
├── reports
│   ├── cached_catalog_status_in_report.rb
│   ├── corrective_change_new_resource.rb
│   ├── corrective_change_outside_puppet.rb
│   └── corrective_change_via_puppet.rb
└── resource
    └── file
        └── ticket_6448_file_with_utf8_source.rb

3 directories, 5 files
```
Here is a list of the tests modified by the patch to add the 'broken:images' tag
```
 M tests/reports/corrective_change_new_resource.rb
 M tests/reports/corrective_change_outside_puppet.rb
 M tests/reports/corrective_change_via_puppet.rb
 M tests/resource/file/ticket_6448_file_with_utf8_source.rb
```
Here is the `rake` command used to execute the suite
```
SHA=1.10.1 TEST_TARGET=centos7-64a OPTIONS='--test-tag-exclude="broken:images"' bundle exec rake ci:test:aio
```
Here is the result summary
```
      Test Suite: tests @ 2017-05-11 16:14:27 -0700

       Total Suite Time: 112.81 seconds
      Average Test Time: 22.56 seconds
              Attempted: 5
                 Passed: 5
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 5
```

@kevpl does the `--test-tag-exclude` flag require that all of the tests have the `tag` method included in order to function correctly.

kevpl:
@johnduarte yes, all of the logic for tagging is in the `tag` method, so it needs to be in a particular test to select whether or not that test should be included or excluded. I'll put a note down to make a doc improvement so that that is clearer in the future.

jcoconnor:
@johnduarte @adrienthebo 
I've put in a large (MAINT) commit to add an empty ```tag``` to the remaining tests that don't have one.

Have run an ```AND``` test to demonstrate that only the ```broken:images``` tests are executed and am currently running a jenkins/imaging build to verify that the ```broken:images``` tests are screened out.

adrienthebo:
Uh... very strong -1 for adding an empty tag to every test. That seems like a really messy workaround for tags not working correctly. I think we should spend the time to make tagging work correctly rather than touch every file.

jcoconnor:
@adrienthebo I agree that I think we should fix tagging so that the default is that each test has an empty tag.

/cc: @kevpl

jcoconnor:
Closing in favour of #5884 which has been renamed to resolve ```appveyor``` naming validation issue.

jcoconnor:
Keep the branch in place to facilitate imaging testing.

PR #5863, opened 2017-05-11
(PUP-7523) Dup RUBY_VERSION in Gem::Version

joshcooper:
The following fails on some buggy versions of rubygems in ruby 1.9.3[1]:

    irb(main):001:0> Gem::Version.new(RUBY_VERSION)
    RuntimeError: can't modify frozen String

As a result, puppet will not run when installed as a gem, until rubygems
is updated.

This commits `dup`s the version string as is done elsewhere[2].

[1] https://github.com/bundler/bundler/issues/3187
[2] https://github.com/puppetlabs/modulesync_configs/commit/be386b44b329e95dc8732155e76bee4353e61216


Iristyle:
Oh Ruby...

PR #5864, opened 2017-05-12
(PUP-7381) Consistent use of YAML

thallgren:
This PR unifies how several Puppet structures, defined in Ruby, interact with the YAML/Psych processor. The objective is to make the code insensitive to ruby versions and also to, to the extent possible, make the data serialized with JSON the same as the data serialized with YAML. One key factor in that effort is to get rid of the YAML tags that causes the YAML.load to directly create object instances.

This PR gets rid of such tags but retains them at the top level. The tags are still needed there for two reasons.
- Puppets own yaml indirections rely on this tag when deserializing.
- Other clients, like the MCO PuppetAgentMgr also relies on such tags.

### DIFF
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -39,7 +39,7 @@ class Puppet::Node
   def to_data_hash
     result = {
       'name' => name,
-      'environment' => environment.name,
+      'environment' => environment.name.to_s,
     }
     result['classes'] = classes unless classes.empty?
     result['parameters'] = parameters unless parameters.empty?
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -27,7 +27,10 @@ class Puppet::Node
     @name       = data['name']       || (raise ArgumentError, _("No name provided in serialized data"))
     @classes    = data['classes']    || []
     @parameters = data['parameters'] || {}
-    @environment_name = data['environment']
+    env_name = data['environment']
+    env_name = env_name.intern unless env_name.nil?
+    @environment_name = env_name
+    environment = env_name
   end
 
   def self.from_data_hash(data)
--- a/lib/puppet/node.rb
+++ b/lib/puppet/node.rb
@@ -36,6 +36,10 @@ class Puppet::Node
     node
   end
 
+  def to_yaml_properties
+    [:@classes, :@environment, :@name, :@parameters]
+  end
+
   def to_data_hash
     result = {
       'name' => name,
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -8,10 +8,11 @@ require 'puppet/parameter'
 # @api public
 class Puppet::Resource
   include Puppet::Util::Tagging
+  include Puppet::Util::PsychSupport
 
   include Enumerable
   attr_accessor :file, :line, :catalog, :exported, :virtual, :strict
-  attr_reader :type, :title
+  attr_reader :type, :title, :parameters, :rich_data_enabled
 
   # @!attribute [rw] sensitive_parameters
   #   @api private
@@ -25,6 +26,7 @@ class Puppet::Resource
   extend Puppet::Indirector
   indirects :resource, :terminus_class => :ral
 
+  EMPTY_ARRAY = [].freeze
   EMPTY_HASH = {}.freeze
 
   ATTRIBUTES = [:file, :line, :exported].freeze
@@ -36,37 +38,46 @@ class Puppet::Resource
   VALUE_KEY = 'value'.freeze
 
   def self.from_data_hash(data, rich_data_enabled = false)
-    raise ArgumentError, "No resource type provided in serialized data" unless type = data['type']
-    raise ArgumentError, "No resource title provided in serialized data" unless title = data['title']
+    resource = self.allocate
+    resource.instance_variable_set(:@rich_data_enabled, true) if rich_data_enabled
+    resource.initialize_from_hash(data)
+    resource
+  end
 
-    resource = new(type, title)
+  def initialize_from_hash(data)
+    raise ArgumentError, _('No resource type provided in serialized data') unless type = data['type']
+    raise ArgumentError, _('No resource title provided in serialized data') unless title = data['title']
+    @type, @title = self.class.type_and_title(type, title)
 
     if params = data['parameters']
+      @parameters = {}
       params.each do |param, value|
-        value = convert_rich_data_hash(value, rich_data_enabled) if value.is_a?(Hash) && value.include?(PCORE_TYPE_KEY)
-        resource[param] = value
+        value = convert_rich_data_hash(value) if value.is_a?(Hash) && value.include?(PCORE_TYPE_KEY)
+        self[param] = value
       end
+    else
+      @parameters = EMPTY_HASH
     end
 
-    if sensitive_parameters = data['sensitive_parameters']
-      resource.sensitive_parameters = sensitive_parameters.map(&:to_sym)
+    if sensitives = data['sensitive_parameters']
+      @sensitive_parameters = sensitives.map(&:to_sym)
+    else
+      @sensitive_parameters = EMPTY_ARRAY
     end
 
     if tags = data['tags']
-      tags.each { |tag| resource.tag(tag) }
+      tags.each { |t| tag(t) }
     end
 
     ATTRIBUTES.each do |a|
       if value = data[a.to_s]
-        resource.send(a.to_s + "=", value)
+        send(a.to_s + "=", value)
       end
     end
-
-    resource
   end
 
-  def self.convert_rich_data_hash(rich_data_hash, rich_data_enabled)
-    if rich_data_enabled
+  def convert_rich_data_hash(rich_data_hash)
+    if instance_variable_defined?(:@rich_data_enabled)
       value = rich_data_hash[VALUE_KEY]
       if value.is_a?(Array)
         json_deserializer ||= Puppet::Pops::Serialization::Deserializer.new(Puppet::Pops::Serialization::JSON::Reader.new([]), nil)
@@ -87,17 +98,21 @@ class Puppet::Resource
       rich_data_hash
     end
   end
-  private_class_method :convert_rich_data_hash
+  private :convert_rich_data_hash
 
   def inspect
     "#{@type}[#{@title}]#{to_hash.inspect}"
   end
 
   def to_data_hash(rich_data_enabled = false)
-    data = ([:type, :title, :tags] + ATTRIBUTES).inject({}) do |hash, param|
-      next hash unless value = self.send(param)
-      hash[param.to_s] = value
-      hash
+    data = {
+      'type' => type,
+      'title' => title,
+      'tags' => tags.to_data_hash
+    }
+    ATTRIBUTES.each do |param|
+      value = send(param)
+      data[param.to_s] = value if value
     end
 
     data['exported'] ||= false
@@ -729,11 +744,5 @@ class Puppet::Resource
     else
       return { :name => title.to_s }
     end
-  end
-
-  def parameters
-    # @parameters could have been loaded from YAML, causing it to be nil (by
-    # bypassing initialize).
-    @parameters ||= {}
   end
 end
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -343,6 +343,7 @@ class Puppet::Resource
         # set the type name to the symbolic name
         type = type.name
       end
+      @exported = false
 
       # Set things like strictness first.
       attributes.each do |attr, value|
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -25,6 +25,8 @@ class Puppet::Resource
   extend Puppet::Indirector
   indirects :resource, :terminus_class => :ral
 
+  EMPTY_HASH = {}.freeze
+
   ATTRIBUTES = [:file, :line, :exported].freeze
   TYPE_CLASS = 'Class'.freeze
   TYPE_NODE  = 'Node'.freeze
@@ -98,7 +100,7 @@ class Puppet::Resource
       hash
     end
 
-    data["exported"] ||= false
+    data['exported'] ||= false
 
     ext_params = {}
     params = {}
@@ -286,7 +288,7 @@ class Puppet::Resource
   #   be the full resource name in the form of `"Type[Title]"`.
   #
   # @api public
-  def initialize(type, title = nil, attributes = {})
+  def initialize(type, title = nil, attributes = EMPTY_HASH)
     @parameters = {}
     @sensitive_parameters = []
     if type.is_a?(Puppet::Resource)
@@ -617,7 +619,7 @@ class Puppet::Resource
     raise Puppet::ParseError.new(_("no parameter named '%{name}'") % { name: name }, file, line) unless valid_parameter?(name)
   end
 
-  def prune_parameters(options = {})
+  def prune_parameters(options = EMPTY_HASH)
     properties = resource_type.properties.map(&:name)
 
     dup.collect do |attribute, value|
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -105,14 +105,15 @@ class Puppet::Resource
     self.to_hash.each_pair do |param, value|
       # Don't duplicate the title as the namevar
       unless param == namevar && value == title
+        name = param.to_s
         value = Puppet::Resource.value_to_json_data(value)
         if is_json_type?(value)
-          params[param] = value
+          params[name] = value
         elsif !rich_data_enabled
           Puppet.warning(_("Resource '%{resource}' contains a %{klass} value. It will be converted to the String '%{value}'") % { resource: to_s, klass: value.class.name, value: value })
-          params[param] = value
+          params[name] = value
         else
-          ext_params[param] = value
+          ext_params[name] = value
         end
       end
     end
--- a/lib/puppet/resource.rb
+++ b/lib/puppet/resource.rb
@@ -70,9 +70,8 @@ class Puppet::Resource
     end
 
     ATTRIBUTES.each do |a|
-      if value = data[a.to_s]
-        send(a.to_s + "=", value)
-      end
+      value = data[a.to_s]
+      send("#{a}=", value) unless value.nil?
     end
   end
 
@@ -112,7 +111,7 @@ class Puppet::Resource
     }
     ATTRIBUTES.each do |param|
       value = send(param)
-      data[param.to_s] = value if value
+      data[param.to_s] = value unless value.nil?
     end
 
     data['exported'] ||= false
--- a/lib/puppet/resource/status.rb
+++ b/lib/puppet/resource/status.rb
@@ -101,7 +101,7 @@ module Puppet
       YAML_ATTRIBUTES = %w{@resource @file @line @evaluation_time @change_count
                            @out_of_sync_count @tags @time @events @out_of_sync
                            @changed @resource_type @title @skipped @failed
-                           @containment_path}.
+                           @containment_path @corrective_change}.
         map(&:to_sym)
 
       def self.from_data_hash(data)
--- a/lib/puppet/resource/status.rb
+++ b/lib/puppet/resource/status.rb
@@ -191,14 +191,8 @@ module Puppet
         @failed = data['failed']
         @corrective_change = data['corrective_change']
         @events = data['events'].map do |event|
-          # in YAML (for reports) we serialize this as an object, but
-          # in PSON it becomes a hash. Depending on where we came from
-          # we might not need to deserialize it.
-          if event.class == Puppet::Transaction::Event
-            event
-          else
-            Puppet::Transaction::Event.from_data_hash(event)
-          end
+          # Older versions contain tags that causes Psych to create instances directly
+          event.is_a?(Puppet::Transaction::Event) ? event : Puppet::Transaction::Event.from_data_hash(event)
         end
       end
 
@@ -219,7 +213,7 @@ module Puppet
           'skipped' => @skipped,
           'change_count' => @change_count,
           'out_of_sync_count' => @out_of_sync_count,
-          'events' => @events,
+          'events' => @events.map { |event| event.to_data_hash },
           'corrective_change' => @corrective_change,
         }
       end
--- a/lib/puppet/transaction/event.rb
+++ b/lib/puppet/transaction/event.rb
@@ -57,7 +57,7 @@ class Puppet::Transaction::Event
       'desired_value' => @desired_value,
       'historical_value' => @historical_value,
       'message' => @message,
-      'name' => @name,
+      'name' => @name.nil? ? nil : @name.to_s,
       'status' => @status,
       'time' => @time.iso8601(9),
       'redacted' => @redacted,
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -35,6 +35,7 @@ require 'puppet/indirector'
 #
 # @api public
 class Puppet::Transaction::Report
+  include Puppet::Util::PsychSupport
   extend Puppet::Indirector
 
   indirects :report, :terminus_class => :processor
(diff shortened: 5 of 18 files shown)


joshcooper:
IIRC `to_yaml_properties` has been deprecated since ruby 1.9.3. Since puppet requires at least that version, can we drop `to_yaml_properties` and solely use `encode_with` for serialization? That way we avoid those getting out of sync in the future.

lib/puppet/resource/status.rb
```diff
@@ -101,7 +101,7 @@ def dependency_failed?
       YAML_ATTRIBUTES = %w{@resource @file @line @evaluation_time @change_count
                            @out_of_sync_count @tags @time @events @out_of_sync
                            @changed @resource_type @title @skipped @failed
-                           @containment_path}.
+                           @containment_path @corrective_change}.
```
joshcooper: Does this mean we were not serializing `corrective_change` in the YAML report? Or did it still work because ruby will prefer `encode_with`?

lib/puppet/node.rb
```diff
@@ -39,7 +39,7 @@ def self.from_data_hash(data)
   def to_data_hash
     result = {
       'name' => name,
-      'environment' => environment.name,
+      'environment' => environment.name.to_s,
```
joshcooper: Are we missing a test in `node_spec` that `environment` is a String?

lib/puppet/transaction/event.rb
```diff
@@ -57,7 +57,7 @@ def to_data_hash
       'desired_value' => @desired_value,
       'historical_value' => @historical_value,
       'message' => @message,
-      'name' => @name,
+      'name' => @name.nil? ? nil : @name.to_s,
```
joshcooper: Same comment about needing a test?

lib/puppet/resource.rb
```diff
@@ -343,6 +343,7 @@ def initialize(type, title = nil, attributes = {})
         # set the type name to the symbolic name
         type = type.name
       end
+      @exported = false
```
joshcooper: Did `@exported` default to `nil` previously, e.g. when written out in YAML? Do we need a test that it's always true/false but never `nil`?

lib/puppet/resource.rb
```diff
@@ -87,17 +98,21 @@ def self.convert_rich_data_hash(rich_data_hash, rich_data_enabled)
...
-      hash[param.to_s] = value
-      hash
+    data = {
+      'type' => type,
+      'title' => title,
+      'tags' => tags.to_data_hash
+    }
+    ATTRIBUTES.each do |param|
+      value = send(param)
+      data[param.to_s] = value if value
```
joshcooper: Looks like a bug in the old code. If `value` is `false` then we never set the value. Seems like that should be `unless value.nil?`

lib/puppet/transaction/report.rb
```diff
@@ -274,47 +275,52 @@ def initialize_from_hash(data)
...
+      'resource_statuses' => Hash[@resource_statuses.map { |key, rs| [key, rs.nil? ? nil : rs.to_data_hash] }],
       'corrective_change' => @corrective_change,
     }
+
+    # The following is include only when set
+    hash['master_used'] = @master_used if @master_used
+    hash['catalog_uuid'] = @catalog_uuid if @catalog_uuid
+    hash['code_id'] = @code_id if @code_id
+    hash['job_id'] = @job_id if @job_id
+    hash['cached_catalog_status'] = @cached_catalog_status if @cached_catalog_status
```
joshcooper: Some of these attributes are [required according to the schema](https://github.com/puppetlabs/puppet/blob/master/api/schemas/catalog.json#L135), even though the values may be `null`.

lib/puppet/util/tag_set.rb
```diff
@@ -16,6 +16,7 @@ def self.from_data_hash(data)
     self.new(data)
   end
 
+  # TODO: A method named #to_data_hash should not return an array
   def to_data_hash
     to_a
```
joshcooper: How does this even work? Is there an implicit conversation from Array to Hash?

lib/puppet/util/metric.rb
```diff
@@ -4,17 +4,24 @@
...
-    metric = new(data['name'], data['label'])
-    metric.values = data['values']
+    metric = allocate
+    metric.initialize_from_hash(data)
     metric
   end
 
+  def initialize_from_hash(data)
+    @name = data['name']
+    @label = data['label'] || self.class.labelize(@name)
```
joshcooper: Can `initialize_from_hash` and `initialize` be collapsed? Right now `initialize` does a `to_s` on the name, but `from_data_hash`/`initialize_from_hash` doesn't.

lib/puppet/node.rb
```diff
@@ -27,7 +27,10 @@ def initialize_from_hash(data)
     @name       = data['name']       || (raise ArgumentError, _("No name provided in serialized data"))
     @classes    = data['classes']    || []
     @parameters = data['parameters'] || {}
-    @environment_name = data['environment']
+    env_name = data['environment']
+    env_name = env_name.intern unless env_name.nil?
+    @environment_name = env_name
+    environment = env_name
```
joshcooper: This will cause an environment lookup to occur and fail if the environment doesn't exist (due to https://github.com/thallgren/puppet/blob/6d42c0e6326b3293eb9513726edbfa32fa3a7892/lib/puppet/node.rb#L78). Is that a problem?

lib/puppet/node.rb
thallgren: We did miss those tests. In this PR, tests are added that asserts that everything returned from the to_data_hash methods is `Data`. The `Data` type does not accept symbols.

lib/puppet/transaction/event.rb
thallgren: And the same reply.

lib/puppet/transaction/report.rb
thallgren: You link to the Catalog schema. None of the properties are r[equired according to the Report schema](https://github.com/puppetlabs/puppet/blob/master/api/schemas/report.json#L9-L21).

lib/puppet/util/metric.rb
thallgren: They could, but that would change API since the `Util::Metric` class is not marked as `@api private`. Also, if we consider that all data hashes passed to `initialize_from_hash` are now `Data`, then a `to_s` call there would be redundant.

lib/puppet/node.rb
thallgren: I haven't been able to find any situation when this fails. All tests pass, both the rspec tests and the acceptance tests (aside from the acceptance tests that failed due to problems with `ci:test:git`).

lib/puppet/resource.rb
thallgren: Actually, [it defaulted to false previously](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/resource.rb#L101) too.

lib/puppet/util/tag_set.rb
thallgren: Generally speaking, this does't work. If someone tries to serialize/deserialize a `TagSet` as a stand-alone entity, it will fail. The only reason it does cause problems now is that the array is embedded in serialization of a `Resource` and that class expects an array of tags when deserializing.

The resolution is to let the `Resource` continue to serialize the tags in an `Array` (by calling `TagSet#to_a` directly), and then let this method return a Hash that describes both its type and its set of tags (to make it possible to serialize tagsets in places where the owner doesn't fix the problem). I think that's beyond the scope of this ticket and would like to do that as part of implementing my proposed serialization.

lib/puppet/resource/status.rb
thallgren: This change is only there to ensure that the `#to_yaml_properties` and `#to_data_hash` uses the same property set. The YAML report is always produced using `#to_data_hash` (tested with Ruby 1.9.3 and 2.3.1) so it is unaffected by this change.

thallgren:
@joshcooper, do you feel that all your questions are answered on this PR?

lib/puppet/node.rb
joshcooper: What if you run `puppet agent` and the agent is configured with an environment name `foo` but the environment directory doesn't exist? Or if you run `puppet resource` and puppet.conf references a non-existent directory?

thallgren: The `puppet apply` case seems to be covered by the acceptance tests _environment/enc_nonexistent_directory_environment.rb_ and _environment/environment-scenario-non_exists.rb_ (covers config, module_install, and apply). Those tests pass. I don't think `puppet resource` does anything differently that would affect this ticket, but if it does, then CI should add a test for that case.

MosesMendoza:
@joshcooper +1 on replacing `to_yaml_properties` with `encode_with` - doesn't necessarily have to be this PR but we should stop using that method before its removed

lib/puppet/node.rb
MosesMendoza: It looks like we already fail both resource and apply when the environment directory doesn't exist because of https://github.com/puppetlabs/puppet/blob/master/lib/puppet/application_support.rb#L29 (we're in `run_mode` user)

lib/puppet/transaction/report.rb
```diff
@@ -274,47 +275,52 @@ def initialize_from_hash(data)
...
-      'metrics' => @metrics,
-      'resource_statuses' => @resource_statuses,
+      'logs' => @logs.map { |log| log.to_data_hash },
+      'metrics' => Hash[@metrics.map { |key, metric| [key, metric.to_data_hash] }],
+      'resource_statuses' => Hash[@resource_statuses.map { |key, rs| [key, rs.nil? ? nil : rs.to_data_hash] }],
       'corrective_change' => @corrective_change,
     }
+
+    # The following is include only when set
+    hash['master_used'] = @master_used if @master_used
```
MosesMendoza: can we do the same `unless...nil?` form for these instead of `if` to avoid not setting when false?

MosesMendoza: can we do the same unless...nil? form for these instead of if to avoid not setting when false?

lib/puppet/node.rb
joshcooper: Ah I think that's https://tickets.puppetlabs.com/browse/PUP-6739.

joshcooper:
Diffing reports between 4.10.1 and master looks correct. There is only a top-level ruby object tag:

```
$ grep object last_run_report.yaml.new
--- !ruby/object:Puppet::Transaction::Report
```

And previous "inner" tags have been stripped:

```diff
1a2,61
> host: xxx
> time: '2017-05-22T14:18:05.996912000-07:00'
> configuration_version: 1495487887
> transaction_uuid: 3a302bc8-1819-47c3-894d-0c86845eb87c
> report_format: 7
> puppet_version: 5.0.0
> kind: apply
> status: unchanged
> noop: false
> noop_pending: false
> environment: production
> logs:
> - level: info
>   message: Using configured environment 'production'
>   source: Puppet
>   tags:
>   - info
>   time: '2017-05-22T14:18:06.185648000-07:00'
>   file: 
>   line: 
> - level: info
>   message: Retrieving pluginfacts
>   source: Puppet
>   tags:
>   - info
>   time: '2017-05-22T14:18:06.186239000-07:00'
>   file: 
>   line: 
> - level: info
>   message: Retrieving plugin
>   source: Puppet
>   tags:
>   - info
>   time: '2017-05-22T14:18:06.233558000-07:00'
>   file: 
>   line: 
> - level: info
>   message: Caching catalog for xxx
>   source: Puppet
>   tags:
>   - info
>   time: '2017-05-22T14:18:07.993096000-07:00'
>   file: 
>   line: 
> - level: info
>   message: Applying configuration version '1495487887'
>   source: Puppet
>   tags:
>   - info
>   time: '2017-05-22T14:18:08.045080000-07:00'
>   file: 
>   line: 
> - level: notice
>   message: Applied catalog in 0.01 seconds
>   source: Puppet
>   tags:
>   - notice
>   time: '2017-05-22T14:18:08.056945000-07:00'
>   file: 
>   line: 
3c63
<   resources: !ruby/object:Puppet::Util::Metric
---
>   resources:
34c94
<   time: !ruby/object:Puppet::Util::Metric
---
>   time:
40c100
<       - 0.000461
---
>       - 0.000694
43c103
<       - 0.00028300000000000005
---
>       - 0.00041999999999999996
46c106
<       - 5.9e-05
---
>       - 6.4e-05
49c109
<       - 0.048120544000994414
---
>       - 0.11104146399884485
52,53c112,113
<       - 0.04892354400099441
<   changes: !ruby/object:Puppet::Util::Metric
---
>       - 0.11221946399884485
>   changes:
60c120
<   events: !ruby/object:Puppet::Util::Metric
---
>   events:
73,127d132
< logs:
< - !ruby/object:Puppet::Util::Log
<   level: :info
<   message: Using configured environment 'production'
<   source: Puppet
<   tags:
<   - info
<   time: '2017-05-22T14:20:29.507222000-07:00'
<   file: 
<   line: 
< - !ruby/object:Puppet::Util::Log
<   level: :info
<   message: Retrieving pluginfacts
<   source: Puppet
<   tags:
<   - info
<   time: '2017-05-22T14:20:29.507950000-07:00'
<   file: 
<   line: 
< - !ruby/object:Puppet::Util::Log
<   level: :info
<   message: Retrieving plugin
<   source: Puppet
<   tags:
<   - info
<   time: '2017-05-22T14:20:29.554353000-07:00'
<   file: 
<   line: 
< - !ruby/object:Puppet::Util::Log
<   level: :info
<   message: Caching catalog for xxx
<   source: Puppet
<   tags:
<   - info
<   time: '2017-05-22T14:20:31.255911000-07:00'
<   file: 
<   line: 
< - !ruby/object:Puppet::Util::Log
<   level: :info
<   message: Applying configuration version '1495488031'
<   source: Puppet
<   tags:
<   - info
<   time: '2017-05-22T14:20:31.298745000-07:00'
<   file: 
<   line: 
< - !ruby/object:Puppet::Util::Log
<   level: :notice
<   message: Applied catalog in 0.01 seconds
<   source: Puppet
<   tags:
<   - notice
<   time: '2017-05-22T14:20:31.309144000-07:00'
<   file: 
<   line: 
129c134
<   File[/tmp/dest.bin]: !ruby/object:Puppet::Resource::Status
---
>   File[/tmp/dest.bin]:
139c144
<     evaluation_time: 0.000461
---
>     evaluation_time: 0.000694
143c148
<     time: '2017-05-22T14:20:31.302391000-07:00'
---
>     time: '2017-05-22T14:18:08.049918000-07:00'
152c157
<   Schedule[puppet]: !ruby/object:Puppet::Resource::Status
---
>   Schedule[puppet]:
160c165
<     evaluation_time: 5.5e-05
---
>     evaluation_time: 6.0e-05
164c169
<     time: '2017-05-22T14:20:31.304147000-07:00'
---
>     time: '2017-05-22T14:18:08.051091000-07:00'
173c178
<   Schedule[hourly]: !ruby/object:Puppet::Resource::Status
---
>   Schedule[hourly]:
181c186
<     evaluation_time: 3.8e-05
---
>     evaluation_time: 6.1e-05
185c190
<     time: '2017-05-22T14:20:31.304270000-07:00'
---
>     time: '2017-05-22T14:18:08.051255000-07:00'
194c199
<   Schedule[daily]: !ruby/object:Puppet::Resource::Status
---
>   Schedule[daily]:
202c207
<     evaluation_time: 3.8e-05
---
>     evaluation_time: 0.0001
206c211
<     time: '2017-05-22T14:20:31.304407000-07:00'
---
>     time: '2017-05-22T14:18:08.051394000-07:00'
215c220
<   Schedule[weekly]: !ruby/object:Puppet::Resource::Status
---
>   Schedule[weekly]:
223c228
<     evaluation_time: 7.7e-05
---
>     evaluation_time: 0.000109
227c232
<     time: '2017-05-22T14:20:31.304502000-07:00'
---
>     time: '2017-05-22T14:18:08.051653000-07:00'
236c241
<   Schedule[monthly]: !ruby/object:Puppet::Resource::Status
---
>   Schedule[monthly]:
244c249
<     evaluation_time: 3.4e-05
---
>     evaluation_time: 5.5e-05
248c253
<     time: '2017-05-22T14:20:31.304640000-07:00'
---
>     time: '2017-05-22T14:18:08.051829000-07:00'
257c262
<   Schedule[never]: !ruby/object:Puppet::Resource::Status
---
>   Schedule[never]:
265c270
<     evaluation_time: 4.1e-05
---
>     evaluation_time: 3.5e-05
269c274
<     time: '2017-05-22T14:20:31.304729000-07:00'
---
>     time: '2017-05-22T14:18:08.051945000-07:00'
278c283
<   Filebucket[puppet]: !ruby/object:Puppet::Resource::Status
---
>   Filebucket[puppet]:
286c291
<     evaluation_time: 5.9e-05
---
>     evaluation_time: 6.4e-05
290c295
<     time: '2017-05-22T14:20:31.304837000-07:00'
---
>     time: '2017-05-22T14:18:08.052076000-07:00'
299,313d303
< host: xxx
< time: 2017-05-22 14:20:29.392055000 -07:00
< kind: apply
< report_format: 6
< puppet_version: 4.10.1
< configuration_version: 1495488031
< transaction_uuid: c46f26b9-0106-4d2f-845d-ef9749001bcb
< code_id: 
< catalog_uuid: 5d249424-6984-4edb-a84a-dbdf97cdd267
< cached_catalog_status: not_used
< master_used: 
< environment: production
< status: unchanged
< noop: false
< noop_pending: false
314a305,306
> catalog_uuid: e5c30cf4-924f-472c-b029-7a34283e3145
> cached_catalog_status: not_used
```

And new code can parse old and new reports:

```diff
1c1
< #<Puppet::Transaction::Report:0x007fc981378870
---
> #<Puppet::Transaction::Report:0x007fd4f997c360
3,4c3,4
<  @catalog_uuid="5d249424-6984-4edb-a84a-dbdf97cdd267",
<  @configuration_version=1495488031,
---
>  @catalog_uuid="e5c30cf4-924f-472c-b029-7a34283e3145",
>  @configuration_version=1495487887,
10c10
<   [#<Puppet::Util::Log:0x007fc9813704e0
---
>   [#<Puppet::Util::Log:0x007fd4f995ce98
15,16c15,16
<     @time=2017-05-22 14:20:29 -0700>,
<    #<Puppet::Util::Log:0x007fc982592510
---
>     @time=2017-05-22 14:18:06 -0700>,
>    #<Puppet::Util::Log:0x007fd4f995c420
21,22c21,22
<     @time=2017-05-22 14:20:29 -0700>,
<    #<Puppet::Util::Log:0x007fc982590cd8
---
>     @time=2017-05-22 14:18:06 -0700>,
>    #<Puppet::Util::Log:0x007fd4f9957a38
27,28c27,28
<     @time=2017-05-22 14:20:29 -0700>,
<    #<Puppet::Util::Log:0x007fc98258be90
---
>     @time=2017-05-22 14:18:06 -0700>,
>    #<Puppet::Util::Log:0x007fd4f9956cc8
33,34c33,34
<     @time=2017-05-22 14:20:31 -0700>,
<    #<Puppet::Util::Log:0x007fc98258b198
---
>     @time=2017-05-22 14:18:07 -0700>,
>    #<Puppet::Util::Log:0x007fd4f9955828
36c36
<     @message="Applying configuration version '1495488031'",
---
>     @message="Applying configuration version '1495487887'",
39,40c39,40
<     @time=2017-05-22 14:20:31 -0700>,
<    #<Puppet::Util::Log:0x007fc98258a130
---
>     @time=2017-05-22 14:18:08 -0700>,
>    #<Puppet::Util::Log:0x007fd4f9954158
45c45
<     @time=2017-05-22 14:20:31 -0700>],
---
>     @time=2017-05-22 14:18:08 -0700>],
48c48
<     #<Puppet::Util::Metric:0x007fc981378410
---
>     #<Puppet::Util::Metric:0x007fd4f995cf60
62c62
<     #<Puppet::Util::Metric:0x007fc981372510
---
>     #<Puppet::Util::Metric:0x007fd4f995cf38
66,70c66,70
<       [["file", "File", 0.000461],
<        ["schedule", "Schedule", 0.00028300000000000005],
<        ["filebucket", "Filebucket", 5.9e-05],
<        ["config_retrieval", "Config retrieval", 0.048120544000994414],
<        ["total", "Total", 0.04892354400099441]]>,
---
>       [["file", "File", 0.000694],
>        ["schedule", "Schedule", 0.00041999999999999996],
>        ["filebucket", "Filebucket", 6.4e-05],
>        ["config_retrieval", "Config retrieval", 0.11104146399884485],
>        ["total", "Total", 0.11221946399884485]]>,
72c72
<     #<Puppet::Util::Metric:0x007fc981371610
---
>     #<Puppet::Util::Metric:0x007fd4f995cf10
77c77
<     #<Puppet::Util::Metric:0x007fc981371070
---
>     #<Puppet::Util::Metric:0x007fd4f995cee8
86,87c86,87
<  @puppet_version="4.10.1",
<  @report_format=6,
---
>  @puppet_version="5.0.0",
>  @report_format=7,
90c90
<     #<Puppet::Resource::Status:0x007fc982588da8
---
>     #<Puppet::Resource::Status:0x007fd4f994f2c0
95c95
<      @evaluation_time=0.000461,
---
>      @evaluation_time=0.000694,
107c107
<      @time=2017-05-22 14:20:31 -0700,
---
>      @time=2017-05-22 14:18:08 -0700,
110c110
<     #<Puppet::Resource::Status:0x007fc982582e30
---
>     #<Puppet::Resource::Status:0x007fd4f994dc18
115c115
<      @evaluation_time=5.5e-05,
---
>      @evaluation_time=6.0e-05,
126c126
<      @time=2017-05-22 14:20:31 -0700,
---
>      @time=2017-05-22 14:18:08 -0700,
129c129
<     #<Puppet::Resource::Status:0x007fc9825815f8
---
>     #<Puppet::Resource::Status:0x007fd4f994d330
134c134
<      @evaluation_time=3.8e-05,
---
>      @evaluation_time=6.1e-05,
145c145
<      @time=2017-05-22 14:20:31 -0700,
---
>      @time=2017-05-22 14:18:08 -0700,
148c148
<     #<Puppet::Resource::Status:0x007fc98257be00
---
>     #<Puppet::Resource::Status:0x007fd4f994cac0
153c153
<      @evaluation_time=3.8e-05,
---
>      @evaluation_time=0.0001,
164c164
<      @time=2017-05-22 14:20:31 -0700,
---
>      @time=2017-05-22 14:18:08 -0700,
167c167
<     #<Puppet::Resource::Status:0x007fc9825791a0
---
>     #<Puppet::Resource::Status:0x007fd4f994c200
172c172
<      @evaluation_time=7.7e-05,
---
>      @evaluation_time=0.000109,
183c183
<      @time=2017-05-22 14:20:31 -0700,
---
>      @time=2017-05-22 14:18:08 -0700,
186c186
<     #<Puppet::Resource::Status:0x007fc9825721c0
---
>     #<Puppet::Resource::Status:0x007fd4f9947868
191c191
<      @evaluation_time=3.4e-05,
---
>      @evaluation_time=5.5e-05,
202c202
<      @time=2017-05-22 14:20:31 -0700,
---
>      @time=2017-05-22 14:18:08 -0700,
205c205
<     #<Puppet::Resource::Status:0x007fc982570b68
---
>     #<Puppet::Resource::Status:0x007fd4f9946e68
210c210
<      @evaluation_time=4.1e-05,
---
>      @evaluation_time=3.5e-05,
221c221
<      @time=2017-05-22 14:20:31 -0700,
---
>      @time=2017-05-22 14:18:08 -0700,
224c224
<     #<Puppet::Resource::Status:0x007fc98256b208
---
>     #<Puppet::Resource::Status:0x007fd4f9946508
229c229
<      @evaluation_time=5.9e-05,
---
>      @evaluation_time=6.4e-05,
240c240
<      @time=2017-05-22 14:20:31 -0700,
---
>      @time=2017-05-22 14:18:08 -0700,
243,244c243,244
<  @time=2017-05-22 14:20:29 -0700,
<  @transaction_uuid="c46f26b9-0106-4d2f-845d-ef9749001bcb">
---
>  @time=2017-05-22 14:18:05 -0700,
>  @transaction_uuid="3a302bc8-1819-47c3-894d-0c86845eb87c">
```

PR #5865, opened 2017-05-12
(PUP-7525) Move Hocon to runtime dependency

Iristyle:
- Move the hocon gem dependency from Gemfile to .gemspec, allowing
   for ci:test:git with some Hiera tests to work properly on platforms
   like Fedora 25.

### DIFF
--- a/.gemspec
+++ b/.gemspec
@@ -41,6 +41,8 @@ Gem::Specification.new do |s|
   # i18n support (gettext-setup and dependencies)
   s.add_runtime_dependency(%q<gettext-setup>, [">= 0.10", "< 1"])
   s.add_runtime_dependency(%q<locale>, "~> 2.1")
+  # hocon is an optional hiera backend shipped in puppet-agent packages
+  s.add_runtime_dependency(%q<hocon>, "~> 1.0")
   # net-ssh is a runtime dependency of Puppet::Util::NetworkDevice::Transport::Ssh
   # Beaker 3.0.0 to 3.10.0 depends on net-ssh 3.3.0beta1
   # Beaker 3.11.0+ depends on net-ssh 4.0+
--- a/Gemfile
+++ b/Gemfile
@@ -62,7 +62,6 @@ group(:development, :test) do
   gem 'addressable', '< 2.5.0'
   gem 'webmock', '~> 1.24'
   gem 'vcr', '~> 2.9'
-  gem "hocon", :require => false
   gem "hiera-eyaml", :require => false
 end
 


Magisus:
👍  after reading the ticket about why this also makes sense in a more general use case.

Iristyle:
One question I have here as a result of including hocon in all circumstances - do we need the hocon feature / guards anymore?

See:

* https://github.com/puppetlabs/puppet/blob/master/lib/puppet/feature/hocon.rb#L3
* https://github.com/puppetlabs/puppet/blob/master/lib/puppet/functions/hocon_data.rb#L4-L6

Magisus:
Doesn't look like it... that is a now a hard dependency, right?

Iristyle:
@Magisus removing guards would make it a hard dependency.  We can always follow up with that part in a separate PR if we want to stay risk averse on this one.

Magisus:
I guess I was asking whether having it in the gemspec as a runtime dependency means that something will fail if the gem can't be found, e.g. will the installation of the puppet gem itself fail?

Iristyle:
Ah @Magisus nope... it would be no different than any other gem dependency in that regard, where the list currently includes:

```
facter < 4, > 2.0
gettext-setup < 1, >= 0.10
hiera < 4, >= 2.0
json_pure ~> 1.8
locale ~> 2.1
```

Iristyle:
I took the @thallgren WIP PR at https://github.com/thallgren/puppet/commits/issue/pup-7381/use-encode_with ... rebased it on this PR, and triggered an acceptance run like:

```shell
GEM_SOURCE=http://rubygems.delivery.puppetlabs.net/ TEST_TARGET=fedora25-64a PUPPET_FORK='Iristyle' bundle exec rake ci:test:git SHA=`git lastref` TEST=./tests/lookup
```

Note that hocon properly installed during the `bundle install` step of `ci:test:git`:


```
rq3bccpeqbj8by6.delivery.puppetlabs.net (fedora25-64-1) 14:21:34$ cd /tmp/puppet.iGk8AH && bundle install --system --binstubs /usr/bin
    Don't run Bundler as root. Bundler can ask for sudo if it is needed, and
    installing your bundle as root will break this application for all non-root
    users on this machine.
    Fetching https://github.com/Iristyle/puppet.git
    /usr/share/gems/gems/psych-2.1.0/lib/psych.rb:227: warning: already initialized constant Psych::VERSION
    /usr/share/ruby/psych.rb:227: warning: previous definition of VERSION was here
    /usr/share/gems/gems/psych-2.1.0/lib/psych.rb:230: warning: already initialized constant Psych::LIBYAML_VERSION
    /usr/share/ruby/psych.rb:230: warning: previous definition of LIBYAML_VERSION was here
    /usr/share/gems/gems/psych-2.1.0/lib/psych.rb:232: warning: already initialized constant Psych::FALLBACK
    /usr/share/ruby/psych.rb:232: warning: previous definition of FALLBACK was here
    Fetching source index from http://rubygems.delivery.puppetlabs.net/
    Resolving dependencies...
    Installing CFPropertyList 2.2.8
    Installing facter 2.4.6
    Installing fast_gettext 1.1.0
    Installing locale 2.1.2
    Installing text 1.3.1
    Installing hiera 3.3.1
    Installing hocon 1.2.5
    Installing net-ssh 4.1.0
    Using bundler 1.14.6
    Installing gettext 3.2.2
    Installing gettext-setup 0.24
    Using puppet 5.0.0 from https://github.com/Iristyle/puppet.git (at 669f6f7@669f6f7)
    Bundle complete! 1 Gemfile dependency, 12 gems now installed.
    Use `bundle show [gemname]` to see where a bundled gem is installed.

  rq3bccpeqbj8by6.delivery.puppetlabs.net (fedora25-64-1) executed in 9.22 seconds
```

And the ultimate result of the test run was 6 pass, 1 fail, 1 error, where I presume the fail / error is a result of the PR being work in progress... not that there was an issue with hocon installation:

```
 Total Suite Time: 100.73 seconds
Average Test Time: 12.59 seconds
        Attempted: 8
           Passed: 6
           Failed: 1
          Errored: 1
          Skipped: 0
          Pending: 0
            Total: 8

- Specific Test Case Status -

Failed Tests Cases:
        Test Case ./tests/lookup/config5_interpolation.rb reported: #<Beaker::DSL::Outcomes::FailTest:
      agent lookup didn't exit properly: (1)
      >
          Test line: ./tests/lookup/config5_interpolation.rb:88:in `block (2 levels) in run_test'

Errored Tests Cases:
        Test Case ./tests/lookup/hiera3_custom_backend.rb reported: #<Beaker::Host::CommandFailure: Host 'lmkmw45hs2xt36c.delivery.puppetlabs.net' exited with 1 running:
       cat /etc/puppetserver/conf.d/puppetserver.conf
      Last 10 lines of output were:
        cat: /etc/puppetserver/conf.d/puppetserver.conf: No such file or directory>
          Test line: ./tests/lookup/hiera3_custom_backend.rb:11:in `block (2 levels) in run_test'
```

thallgren:
FWIW, the observed failures occurs even without my PR. I tested with:
```
TEST_TARGET=fedora25-64a MASTER_TEST_TARGET=fedora25-64m PUPPET_FORK=iristyle bundle exec rake ci:test:git SHA=3c443ddf130b31f52a7b718dfa77e31f491b58aa TESTS=./tests/lookup
```

Magisus:
So interestingly, the failed test case (`config5_interpolation`) passes if run in isolation. The `hiera3_custom_backend` test still errors if run alone. Could it be messing up the state somehow?

Iristyle:
As @Magisus mentioned, due to https://tickets.puppetlabs.com/browse/PA-1077, the `.gemspec` here is used only by:

* bundle installs from source (i.e. git clone, then `bundle install`)
* bundle installs (such as `ci:test:git`) that install Puppet as a gem (with Puppet resolved via github repo rather than gem version) 

The actual shipped gem is produced via another process that consumes https://github.com/puppetlabs/puppet/blob/master/ext/project_data.yaml#L19-L25 - arguably, we should fix that as well given all other package formats ship Hocon.

PR #5866, opened 2017-05-12
(PUP-7198) Use facts from compilation request to retrieve node

MosesMendoza:
(PUP-7198)(PUP-3438) use request facts for classification

Prior to this PR, the catalog compiler terminus would extract the facts from a given compilation request, save them, and then not use them when requesting classification for a node. Instead the node terminus would look up the facts for any given request. However, because fact submission can be asynchronous, it was possible that the facts saved by the compiler terminus would not ultimately be the ones used by the node terminus when retrieving classification - possibly instead retrieving stale/outdated facts for a node despite the correct facts having been submitted with the original compilation request. This race condition is most evident in environments using exported resurces with puppetdb that have not deployed (to every compile master) a routes.yaml which sets the facts cache specifically to yaml instead of puppetdb. The puppetdb module deploys a routes.yaml file by default, but in a multi-master environment, the routes.yaml file may only have been deployed to one of the masters.

This PR updates the catalog compiler terminus to instead thread the facts object that it extracts from the request through its logic in finding a node, so that we can ultimately pass this facts object on to the node request. We extract the facts object first instead of just passing the original request on because the compiler terminus performs deserialization on the received facts from the original request and it would be nice to not duplicate that elsewhere.

Depends on https://github.com/puppetlabs/classifier/pull/584

### DIFF
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -20,6 +20,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
       raise ArgumentError, _("Facts but no fact format provided for %{request}") % { request: request.key }
     end
 
+    facts = nil
     Puppet::Util::Profiler.profile(_("Found facts"), [:compiler, :find_facts]) do
       # If the facts were encoded as yaml, then the param reconstitution system
       # in Network::HTTP::Handler will automagically deserialize the value.
@@ -38,19 +39,21 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
       unless facts.name == request.key
         raise Puppet::Error, _("Catalog for %{request} was requested with fact definition for the wrong node (%{fact_name}).") % { request: request.key.inspect, fact_name: facts.name.inspect }
       end
-
-      options = {
-        :environment => request.environment,
-        :transaction_uuid => request.options[:transaction_uuid],
-      }
-
-      Puppet::Node::Facts.indirection.save(facts, nil, options)
     end
+    facts
+  end
+
+  def save_facts_from_request(facts, request)
+    Puppet::Node::Facts.indirection.save(facts, nil,
+                                         :environment => request.environment,
+                                         :transaction_uuid => request.options[:transaction_uuid])
   end
 
   # Compile a node's catalog.
   def find(request)
-    extract_facts_from_request(request)
+    facts = extract_facts_from_request(request)
+
+    save_facts_from_request(facts, request) if !facts.nil?
 
     node = node_from_request(request)
     node.trusted_data = Puppet.lookup(:trusted_information) { Puppet::Context::TrustedInformation.local(node) }.to_h
--- a/lib/puppet/indirector/catalog/compiler.rb
+++ b/lib/puppet/indirector/catalog/compiler.rb
@@ -55,7 +55,7 @@ class Puppet::Resource::Catalog::Compiler < Puppet::Indirector::Code
 
     save_facts_from_request(facts, request) if !facts.nil?
 
-    node = node_from_request(request)
+    node = node_from_request(facts, request)
     node.trusted_data = Puppet.lookup(:trusted_information) { Puppet::Context::TrustedInformation.local(node) }.to_h
 
     if catalog = compile(node, request.options)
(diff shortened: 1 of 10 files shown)


lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -292,22 +295,26 @@ def compile(node, options)
...
+    configured_environment = request.options[:configured_environment]
+
     Puppet::Util::Profiler.profile(_("Found node information"), [:compiler, :find_node]) do
       node = nil
       begin
         node = Puppet::Node.indirection.find(name, :environment => environment,
                                              :transaction_uuid => transaction_uuid,
-                                             :configured_environment => configured_environment)
+                                             :configured_environment => configured_environment,
+                                             :facts => facts)
```
branan: Should we update the various built-in node terminii to use this new parameter? At a glance, `plain` and `ldap` at least will query the facts terminus (using `Node::fact_merge`)

Magisus:
This solution makes a lot of sense to me, mostly because it cuts down on complexity. We seemed be to be already passing a lot of things around between these various classes, so this pattern doesn't seem too out of place. I am interested in Branan's comment about whether or not we can/should use this elsewhere. Since it seems easier to track and debug, I'd be in favor of that if possible.

lib/puppet/indirector/catalog/compiler.rb
joshcooper: 👍 Seems like we could add an optional parameter to `fact_merge(facts = nil)` and modify the implementation to use the facts if passed in, or fall back to the old behavior. That way we don't unnecessarily break custom node termini.

lib/puppet/node.rb
```diff
@@ -113,15 +113,25 @@ def initialize(name, options = {})
...
+  # @param[optional, Puppet::Node::Facts] facts to merge into node parameters.
+  #   Will query Facts indirection if not supplied.
+  # @raise[Puppet::Error] Raise on failure to retrieve facts if not supplied
+  # @return[nil]
+  def fact_merge(facts = nil)
+    if facts.nil?
+      begin
+        @facts = Puppet::Node::Facts.indirection.find(name, :environment => environment)
+      rescue => detail
+        error = Puppet::Error.new(_("Could not retrieve facts for %{name}: %{detail}") % { name: name, detail: detail })
```
pcarlisle: you can supply the original error as the second argument to the constructor to save it as a wrapped exception

joshcooper: Welcome back!

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -292,22 +295,26 @@ def compile(node, options)
     config
   end
 
-  # Turn our host name into a node object.
-  def find_node(name, environment, transaction_uuid, configured_environment)
+  # Use indirection to find the node associated with a given request
+  def find_node(facts, request)
```
pcarlisle: it was better with explicit arguments

pcarlisle: Especially since the one specific thing was also extracted from request.

lib/puppet/indirector/catalog/compiler.rb
```diff
@@ -20,6 +20,7 @@ def extract_facts_from_request(request)
       raise ArgumentError, _("Facts but no fact format provided for %{request}") % { request: request.key }
     end
 
+    facts = nil
```
pcarlisle: I confused myself reading this at first thinking it was unnecessary because the next big if expression was exhaustive. I think it would relieve my confusion if that expression were extracted into its own function, then you could just assign the result to facts directly.

lib/puppet/indirector/catalog/compiler.rb
MosesMendoza: interesting. ok.

MosesMendoza: i like my reply apparently

lib/puppet/indirector/catalog/compiler.rb
MosesMendoza: yes

MosesMendoza: `plain`, `ldap`, and `exec` node terminii updated to use facts from request

MosesMendoza:
I think I've addressed all feedback and this is ready for another review

geoffnichols:
@joshcooper https://github.com/puppetlabs/classifier/pull/584 was merged, so I think this is good to go.

lib/puppet/indirector/catalog/compiler.rb
joshcooper: If the `facts` from the request is nil (from https://github.com/puppetlabs/puppet/pull/5866/files#diff-a499fd5869d90e9967707b6fb6f445f3R45), we end up passing a `:facts => nil` param above. Is that ok? Should we instead omit that hash entry?

PR #5867, opened 2017-05-12
(PUP-3486) Disable puppetserver CA service for test

johnduarte:
This commit updates the
`acceptance/tests/external_ca_support/jetty_external_root_ca.rb` test
to disable the puppetserver CA service rather than using the `ca = false`
setting in the `puppet.conf` file. The `auth.conf` file is also updated to
used the HOCON syntax and proper location for use with puppetserver.

[skip ci]

### DIFF
--- a/acceptance/tests/external_ca_support/fixtures/auth.conf
+++ b/acceptance/tests/external_ca_support/fixtures/auth.conf
@@ -1,60 +1,149 @@
-# Puppet 4.0.0 auth.conf, modified to allow requests from example.org for
+# Puppetserver 2.2.0 auth.conf, modified to allow requests from example.org for
 # external ca testing.
 
-# allow nodes to retrieve their own catalog
-path ~ ^/puppet/v3/catalog/([^/]+)$
-method find
-allow *.example.org
-allow $1
-
-# allow nodes to retrieve their own node definition
-path ~ ^/puppet/v3/node/([^/]+)$
-method find
-allow *.example.org
-allow $1
-
-# allow all nodes to access the certificates services
-path /puppet-ca/v1/certificate_revocation_list/ca
-method find
-allow *
-
-# allow all nodes to store their own reports
-path ~ ^/puppet/v3/report/([^/]+)$
-method save
-allow *.example.org
-allow $1
-
-# Allow all nodes to access all file services; this is necessary for
-# pluginsync, file serving from modules, and file serving from custom
-# mount points (see fileserver.conf). Note that the `/file` prefix matches
-# requests to both the file_metadata and file_content paths. See "Examples"
-# above if you need more granular access control for custom mount points.
-path /puppet/v3/file
-allow *
-
-### Unauthenticated ACLs, for clients without valid certificates; authenticated
-### clients can also access these paths, though they rarely need to.
-
-# allow access to the CA certificate; unauthenticated nodes need this
-# in order to validate the puppet master's certificate
-path /puppet-ca/v1/certificate/ca
-auth any
-method find
-allow *
-
-# allow nodes to retrieve the certificate they requested earlier
-path /puppet-ca/v1/certificate/
-auth any
-method find
-allow *
-
-# allow nodes to request a new certificate
-path /puppet-ca/v1/certificate_request
-auth any
-method find, save
-allow *
-
-# deny everything else; this ACL is not strictly necessary, but
-# illustrates the default policy.
-path /
-auth any
+authorization: {
+    version: 1
+    rules: [
+        {
+            # Allow nodes to retrieve their own catalog
+            match-request: {
+                path: "^/puppet/v3/catalog/([^/]+)$"
+                type: regex
+                method: [get, post]
+            }
+            allow: [
+                "$1",
+                "*.example.org"
+            ]
+            sort-order: 500
+            name: "puppetlabs catalog"
+        },
+        {
+            # Allow nodes to retrieve the certificate they requested earlier
+            match-request: {
+                path: "/puppet-ca/v1/certificate/"
+                type: path
+                method: get
+            }
+            allow-unauthenticated: true
+            sort-order: 500
+            name: "puppetlabs certificate"
+        },
+        {
+            # Allow all nodes to access the certificate revocation list
+            match-request: {
+                path: "/puppet-ca/v1/certificate_revocation_list/ca"
+                type: path
+                method: get
+            }
+            allow-unauthenticated: true
+            sort-order: 500
+            name: "puppetlabs crl"
+        },
+        {
+            # Allow nodes to request a new certificate
+            match-request: {
+                path: "/puppet-ca/v1/certificate_request"
+                type: path
+                method: [get, put]
+            }
+            allow-unauthenticated: true
+            sort-order: 500
+            name: "puppetlabs csr"
+        },
+        {
+            match-request: {
+                path: "/puppet/v3/environments"
+                type: path
+                method: get
+            }
+            allow: "*"
+            sort-order: 500
+            name: "puppetlabs environments"
+        },
+        {
+            # Allow nodes to access all file services; this is necessary for
+            # pluginsync, file serving from modules, and file serving from
+            # custom mount points (see fileserver.conf). Note that the `/file`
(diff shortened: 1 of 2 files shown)


acceptance/tests/external_ca_support/jetty_external_root_ca.rb
```diff
@@ -93,19 +89,21 @@
...
-        'ca' => false,
         'certname' => fixtures.master_name,
         'ssl_client_header' => "HTTP_X_CLIENT_DN",
         'ssl_client_verify_header' => "HTTP_X_CLIENT_VERIFY"
     }
 }
 
+# disable CA service
+# https://github.com/puppetlabs/puppetserver/blob/master/documentation/configuration.markdown#service-bootstrapping
+create_remote_file master, "#{jetty_confdir}/../services.d/ca.conf", "puppetlabs.services.ca.certificate-authority-disabled-service/certificate-authority-disabled-service"
```
camlow325: I think this file needs to have a .cfg file extension, like ca.cfg, and not ca.conf.  I don't think the Trapperkeeper bootstrapping logic will load a file that has any file extension other than .cfg.

johnduarte: Thank you @camlow325 ! I missed that detail in the docs. I have pushed an updated commit fixing this oversight.

acceptance/tests/external_ca_support/jetty_external_root_ca.rb
```diff
@@ -20,18 +20,24 @@
...
 
 # Register our cleanup steps early in a teardown so that they will happen even
 # if execution aborts part way.
 teardown do
-  step "Restore /etc/hosts and webserver.conf"
-  on master, "cp -p '#{testdir}/hosts' /etc/hosts"
-  on master, "cp -p '#{testdir}/webserver.conf.orig' '#{jetty_confdir}/webserver.conf'"
+  step "Restore /etc/hosts and puppetserver configs"
+  on master, "cp -p '#{backupdir}/hosts' /etc/hosts"
+  on master, "cp -rp '#{backupdir}/puppetserver/' '#{jetty_confdir}/../'"
```
camlow325: This appears to copy files to the wrong directory.  On my VM, the files end up in `/etc/puppetlabs/puppetserver/puppetserver`, leaving the files modified while the test was running still in place - and any normal agent runs I do after that to fail.

Something like this might work:

```ruby
yes | cp -rp '#{backupdir}/puppetserver' '#{jetty_confdir}/../..'
```
I had to do the `yes` pipe in front in order to avoid the overwrite prompts I would get otherwise.

camlow325: It seems like sometime after the original configuration is restored that we should restart the server.  Otherwise, any subsequent test which runs without the server having been restarted might be using a dirty configuration.

acceptance/tests/external_ca_support/fixtures/auth.conf
```diff
@@ -1,60 +1,149 @@
...
-path /puppet-ca/v1/certificate_request
-auth any
-method find, save
-allow *
-
-# deny everything else; this ACL is not strictly necessary, but
-# illustrates the default policy.
-path /
-auth any
+authorization: {
```
camlow325: Going with tk-auth instead of the Ruby auth.conf makes sense to me.  For Puppet 5, Puppet Server will default to using tk-auth.  So long as these tests are only run with Puppet Server 5+ code, should be okay.  If these might run against older Puppet Server builds, though, (like 2.7.2, as I believe the agent pipeline currently does) you might want to have the test explicitly set `use-legacy-auth-conf` to `false`.  The older Puppet Server builds default the `use-legacy-auth-conf` setting to `true`.

You'll see lines like this in the `allow_arbitrary_node_name...` and `check_woy_cache_works` tests:

```ruby
modify_tk_config(master, options['puppetserver-config'], {'jruby-puppet' => {'use-legacy-auth-conf' => false}})
```

acceptance/tests/external_ca_support/jetty_external_root_ca.rb
camlow325: Oh, also, you'll need to have the test do a full service restart after making this change.  Bootstrap configuration file changes don't take effect with a service reload/HUP only, which is what `with_puppet_running_on` would do by default.

acceptance/tests/external_ca_support/fixtures/auth.conf
johnduarte: Thanks for the awesome background on this @camlow325 . I have pushed an update to set `use-legacy-auth-conf = false`.

acceptance/tests/external_ca_support/jetty_external_root_ca.rb
johnduarte: Thanks for the catch. That was a dumb mistake on my part. I have chosen to use \cp to ensure that they system alias that includes the `-i` flag is subverted. This will allow the `--force` to perform the overwrite without interruption.

acceptance/tests/external_ca_support/jetty_external_root_ca.rb
```diff
@@ -20,18 +20,25 @@
...
 
 # Register our cleanup steps early in a teardown so that they will happen even
 # if execution aborts part way.
 teardown do
-  step "Restore /etc/hosts and webserver.conf"
-  on master, "cp -p '#{testdir}/hosts' /etc/hosts"
-  on master, "cp -p '#{testdir}/webserver.conf.orig' '#{jetty_confdir}/webserver.conf'"
+  step "Restore /etc/hosts and puppetserver configs; Restart puppetserver"
+  on master, "cp -p '#{backupdir}/hosts' /etc/hosts"
+  on master, "\\cp -frp #{backupdir}/puppetserver/* #{jetty_confdir}/../"
```
joshcooper: typo `\\cp`?

johnduarte: @joshcooper this is not a typo. This is to ensure that the actual `cp` command is used rather than the `alias cp='cp -i'` which is present on most systems. If the alias is used the command will get executed in interactive mode which is not what we want here.

joshcooper: ahhhhh! Could you add a comment?

johnduarte: Comment added. Thanks @joshcooper

johnduarte:
@camlow325 I think I have addressed your concerns. Let me know if I missed anything.

joshcooper:
ping @camlow325

camlow325:
Sorry, will look at this again today....

acceptance/tests/external_ca_support/jetty_external_root_ca.rb
```diff
@@ -93,19 +98,21 @@
...
         'certname' => fixtures.master_name,
         'ssl_client_header' => "HTTP_X_CLIENT_DN",
         'ssl_client_verify_header' => "HTTP_X_CLIENT_VERIFY"
     }
 }
 
+# disable CA service
+# https://github.com/puppetlabs/puppetserver/blob/master/documentation/configuration.markdown#service-bootstrapping
+create_remote_file master, "#{jetty_confdir}/../services.d/ca.cfg", "puppetlabs.services.ca.certificate-authority-disabled-service/certificate-authority-disabled-service"
+
```
camlow325: I think we should put another line like this here:

```ruby
on(master, "service #{master['puppetservice']} restart")
```

Without this, the `certificate-authority-disabled-service` isn't loaded before subsequent steps are run.  After one of the bootstrap-related configuration files is changed, a full service restart is needed whereas the `with_puppet_running_on` function will only try to do a service reload (HUP) by default.

camlow325:
@johnduarte Other than adding the one extra service restart I suggested, I think this PR looks good.  Once you put that in place, I can merge this.  Thanks!

acceptance/tests/external_ca_support/jetty_external_root_ca.rb
johnduarte: Thanks @camlow325 . I have added this restart line.

PR #5868, opened 2017-05-12
Revert "(PUP-6675) Use pipes instead of temporary files for Puppet exec"

MikaelSmith:
Reverts puppetlabs/puppet#5844

This fails the sleepy_daemon test on Solaris. For some reason `reader.eof?` doesn't return when executing `/usr/bin/ctrun` as part of starting an init script.


Magisus:
👍

PR #5869, opened 2017-05-14
(PUP-7529) Handling of non-leading tildes in rpmvercmp

knqyf263:
`rpmvercmp` handles only the leading tilde.
So,  the following test fails.

```
it { expect(provider.rpmvercmp("~", "~~")).to eq(1) }
```

A tilde must sort before everything else.
https://github.com/rpm-software-management/rpm/issues/214


joshcooper:
Thanks for your contribution @knqyf263, this will be released in Puppet 5!

PR #5870, opened 2017-05-15

MikaelSmith:
Under selinux, when Puppet is invoked by another process with
reduced privileges, any sub-programs invoked by Puppet will not
inherit Puppet's selinux priveleges. This specifically causes silent
failures when invoking applications that don't normally have the
ability to write files - such as iptables-save or hostname -
because Puppet redirects their output to a temporary file.

subprocesses on POSIX systems. Poorly behaved scripts need special
handling, as a plain `foo &` will inherit stdout and not close it.
Read from the pipe without requiring EOF.

On Windows, we retain the old file-based IO to avoid breaking poorly
written scripts - `start /b ping 127.0.0.1` captures stdout - and to
retain the current behavior when Puppet is terminated - a new Puppet run
can start even if a subprocess was left running by the prior process.

### DIFF
--- a/acceptance/tests/resource/exec/should_accept_large_output.rb
+++ b/acceptance/tests/resource/exec/should_accept_large_output.rb
@@ -1,0 +1,26 @@
+test_name "tests that puppet correctly captures large and empty output."
+
+agents.each do |agent|
+  testfile = agent.tmpfile('should_accept_large_output')
+
+  # Generate >64KB file to exceed pipe buffer.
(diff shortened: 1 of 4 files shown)


MikaelSmith:
This is a travesty, but I think it'll work. I'd like to work up an RFC for puppet-dev to see if we can kill off supporting poorly-behaved scripts that capture stdout in Puppet 5.

MikaelSmith:
Ok, totally meant to merge a different PR.

PR #5871, opened 2017-05-15
Revert "(PUP-6675) Use pipes instead of temporary files for Puppet exec"

MikaelSmith:
Reverts puppetlabs/puppet#5870, which was accidentally merged.


joshcooper:
👍

PR #5872, opened 2017-05-15

MikaelSmith:
Under selinux, when Puppet is invoked by another process with
reduced privileges, any sub-programs invoked by Puppet will not
inherit Puppet's selinux priveleges. This specifically causes silent
failures when invoking applications that don't normally have the
ability to write files - such as iptables-save or hostname -
because Puppet redirects their output to a temporary file.

subprocesses on POSIX systems. Poorly behaved scripts need special
handling, as a plain foo & will inherit stdout and not close it.
Read from the pipe without requiring EOF.

On Windows, we retain the old file-based IO to avoid breaking poorly
written scripts - start /b ping 127.0.0.1 captures stdout - and to
retain the current behavior when Puppet is terminated - a new Puppet run
can start even if a subprocess was left running by the prior process.

### DIFF
--- a/acceptance/tests/resource/exec/should_run_bad_command.rb
+++ b/acceptance/tests/resource/exec/should_run_bad_command.rb
@@ -1,0 +1,67 @@
+test_name "tests that puppet can run badly written scripts that fork and inherit descriptors"
+
+def sleepy_daemon_script(agent)
+  if agent['platform'] =~ /win/
+    # Windows uses a shorter sleep, because it's expected to wait until the end.
+    return <<INITSCRIPT
+echo hello
+start /b ping.exe 127.0.0.1 -n 1
+INITSCRIPT
+  else
+    return <<INITSCRIPT
+echo hello
+/bin/sleep 60 &
+INITSCRIPT
+  end
+end
+
+# TODO: taken from pxp-agent, find common home
+def stop_sleep_process(targets, accept_no_pid_found = false)
+  targets = [targets].flatten
+
+  targets.each do |target|
+    case target['platform']
+    when /osx/
+      command = "ps -e -o pid,comm | grep sleep | sed 's/^[^0-9]*//g' | cut -d\\  -f1"
+    when /win/
+      command = "ps -efW | grep PING | sed 's/^[^0-9]*[0-9]*[^0-9]*//g' | cut -d ' ' -f1"
+    else
+      command = "ps -ef | grep 'bin/sleep ' | grep -v 'grep' | grep -v 'true' | sed 's/^[^0-9]*//g' | cut -d\\  -f1"
+    end
+
+    # A failed test may leave an orphaned sleep process, handle multiple matches.
+    pids = nil
+    on(target, command) do |output|
+      pids = output.stdout.chomp.split
+      if pids.empty? && !accept_no_pid_found
+        raise("Did not find a pid for a sleep process on #{target}")
+      end
+    end
+
+    pids.each do |pid|
+      target['platform'] =~ /win/ ?
+        on(target, "taskkill /F /pid #{pid}") :
+        on(target, "kill -s TERM #{pid}")
+    end
+  end
+end
+
+teardown do
+  # On Windows, Puppet waits until the sleep process exits before exiting
+  stop_sleep_process(agents.select {|agent| agent['platform'] =~ /win/}, true)
+  # Requiring a sleep process asserts that Puppet exited before the sleep process.
+  stop_sleep_process(agents.reject {|agent| agent['platform'] =~ /win/})
+end
+
+agents.each do |agent|
+  ext = if agent['platform'] =~ /win/ then '.bat' else '' end
+  daemon = agent.tmpfile('sleepy_daemon') + ext
+  create_remote_file(agent, daemon, sleepy_daemon_script(agent))
+  on(agent, "chmod +x #{daemon}")
+
+  apply_manifest_on(agent, "exec {'#{daemon}': logoutput => true}") do
+    fail_test "didn't seem to run the command" unless
+      stdout.include? 'executed successfully'
+  end
+end
+
--- a/lib/puppet/util/execution.rb
+++ b/lib/puppet/util/execution.rb
@@ -188,18 +188,69 @@ module Puppet::Util::Execution
 
     begin
       stdin = Puppet::FileSystem.open(options[:stdinfile] || null_file, nil, 'r')
-      stdout = options[:squelch] ? Puppet::FileSystem.open(null_file, nil, 'w') : Puppet::FileSystem::Uniquefile.new('puppet')
+      # On Windows, continue to use the file-based approach to avoid breaking people's existing
+      # manifests. If they use a script that doesn't background cleanly, such as
+      # `start /b ping 127.0.0.1`, we couldn't handle it with pipes as there's no non-blocking
+      # read available.
+      if options[:squelch]
+        stdout = Puppet::FileSystem.open(null_file, nil, 'w')
+      elsif Puppet.features.posix?
+        reader, stdout = IO.pipe
+      else
+        stdout = Puppet::FileSystem::Uniquefile.new('puppet')
+      end
       stderr = options[:combine] ? stdout : Puppet::FileSystem.open(null_file, nil, 'w')
 
       exec_args = [command, options, stdin, stdout, stderr]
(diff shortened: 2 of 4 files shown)


MikaelSmith:
Re-opened after reverting the accidental merge.

MikaelSmith:
Need to look into https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_intn-van-sys_suite-ad-hoc-vmpooler-pxp-agent-ad-hoc/39/.

lib/puppet/util/execution.rb
```diff
@@ -188,18 +188,65 @@ def self.execute(command, options = NoOptionsSpecified)
...
+            # forks and inherits stdout, as happens in `foo &`.
+            until results = Process.waitpid2(child_pid, Process::WNOHANG)
+
+              # If not done, wait for data to read with a timeout
+              # This timeout is selected to keep activity low while waiting on
+              # a long process, while not waiting too long for the pathological
+              # case where stdout is never closed.
+              ready = IO.select([reader], [], [], 0.1)
+              begin
+                output << reader.read_nonblock(4096) if ready
```
MikaelSmith: @joshcooper @Iristyle this can result in returning a string that is BINARY/ASCII-8BIT encoded. The read is binary, and if the result of that read includes non 7-bit ASCII characters it will not implicitly convert to the internal encoding (by assigning to output).

Previously we would read from a file, which would use whatever default encoding `File.open` uses.

My intuition is that we should force_encoding to external_encoding after finishing all reads, then encode to internal_encoding. Alternatively, we could use the `Puppet::Util::CharacterEncoding::convert_to_utf_8` helper to convert to UTF-8. Have any suggestions?

MikaelSmith: Pushed up a PR doing this. Fixes the pxp-agent issue. It may need some error handling.

lib/puppet/util/execution.rb
```diff
@@ -188,18 +188,65 @@ def self.execute(command, options = NoOptionsSpecified)
...
+        reader, stdout = IO.pipe
+      else
+        stdout = Puppet::FileSystem::Uniquefile.new('puppet')
+      end
       stderr = options[:combine] ? stdout : Puppet::FileSystem.open(null_file, nil, 'w')
 
       exec_args = [command, options, stdin, stdout, stderr]
+      output = ''
 
+      # We close stdin/stdout/stderr immediately after execution as there no longer needed.
```
mruzicka: typo: there -> they're

lib/puppet/util/execution.rb
```diff
@@ -188,18 +188,67 @@ def self.execute(command, options = NoOptionsSpecified)
...
+            begin
+              loop do
+                output << reader.read_nonblock(4096)
+              end
+            rescue Errno::EAGAIN
+            rescue EOFError
+            end
+
+            output.force_encoding(Encoding.default_external)
+            output.encode!
```
Iristyle: Not sure what this `.encode!` is for...

lib/puppet/util/execution.rb
Iristyle: I would use the new helper, which is designed around coercing byte streams into the correct encoding.

lib/puppet/util/execution.rb
MikaelSmith: It converts to `Encoding.default_internal`. I thought that was more generic than UTF-8, but if we assert that the default internal will always be UTF-8 then using `Puppet::Util::CharacterEncoding::convert_to_utf_8` definitely makes sense.

Iristyle: Yeah, we always want UTF-8 within our system - I don't see that changing as our encoding lingua franca.

MikaelSmith:
Build available at http://builds.puppetlabs.lan/puppet-agent/ac16ae4b0a2905f6ef1d0d52a9ed06ca9f0c775e/, ad-hoc pipeline's acting up so plan to manually validate some of the platforms.

MikaelSmith:
Now a Solaris failure in the pxp-agent test...

lib/puppet/util/execution.rb
```diff
@@ -188,18 +188,69 @@ def self.execute(command, options = NoOptionsSpecified)
...
+                output << reader.read_nonblock(4096)
+              end
+            rescue Errno::EAGAIN
+            rescue EOFError
+            end
+
+            # Force to external encoding to preserve prior behavior when reading a file.
+            # Wait until after reading all data so we don't encounter corruption when
+            # reading part of a multi-byte unicode character if default_external is UTF-8.
+            output.force_encoding(Encoding.default_external)
```
MikaelSmith: Switched to only forcing the encoding to `Encoding.default_external` to preserve the prior behavior with `File.open` (see `wait_for_output`).

MikaelSmith:
Build available at SHA=`5a32508a947d649d9bb25bd0fb245fa3a6c0216a` SUITE_VERSION=`1.8.2.1240.g5a32508`, doing acceptance runs on `redhat7-64a-windows2012r2-64a-ubuntu1604-64a-redhat5-64a-solaris11-64a` now.

lib/puppet/util/execution.rb
```diff
@@ -188,18 +188,69 @@ def self.execute(command, options = NoOptionsSpecified)
 
     begin
       stdin = Puppet::FileSystem.open(options[:stdinfile] || null_file, nil, 'r')
-      stdout = options[:squelch] ? Puppet::FileSystem.open(null_file, nil, 'w') : Puppet::FileSystem::Uniquefile.new('puppet')
+      # On Windows, continue to use the file-based approach to avoid breaking people's existing
+      # manifests. If they use a script that doesn't background cleanly, such as
+      # `start /b ping 127.0.0.1`, we couldn't handle it with pipes as there's no non-blocking
+      # read available.
+      if options[:squelch]
+        stdout = Puppet::FileSystem.open(null_file, nil, 'w')
```
mruzicka: Would it make sense to reimplement https://stackoverflow.com/questions/34504970/non-blocking-read-on-os-pipe-on-windows to get non-blocking IO on pipes even on windows and unify the implementations?

MikaelSmith: Possibly. It's currently not causing any issues to use a file on Windows.

lib/puppet/util/execution.rb
MikaelSmith: To fully mimic the `File.open` behavior, this should then encode to default_internal (which is normally unset, resulting in a no-op). That could often cause problems, so I'm not sure it's a behavior worth emulating.

MikaelSmith:
Re-running acceptance at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_intn-van-sys_suite-daily-puppet-ad-hoc/2/.

acceptance/tests/resource/exec/should_run_bad_command.rb
```diff
@@ -0,0 +1,67 @@
...
+    pids.each do |pid|
+      target['platform'] =~ /win/ ?
+        on(target, "taskkill /F /pid #{pid}") :
+        on(target, "kill -s TERM #{pid}")
+    end
+  end
+end
+
+teardown do
+  # On Windows, Puppet waits until the sleep process exits before exiting
```
MikaelSmith: Apparently Windows didn't behave the way I expected. This behavior isn't changed by my PR, but this test is different on Windows to take it into account.

MikaelSmith:
Passed Puppet acceptance: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_intn-van-sys_suite-daily-puppet-ad-hoc/5/

lib/puppet/util/execution.rb
mruzicka: ok

MosesMendoza:
@MikaelSmith sorry I didn't intend to close this - I deleted the `revert-5871-revert-5870-PUP-6675` branch from puppetlabs/puppet believing it was one of the left-over automatically created branches from a github web-based revert. Apologies - can you re-open?

MikaelSmith:
Reopened as https://github.com/puppetlabs/puppet/pull/5967.

PR #5875, opened 2017-05-16
(PUP-6566) Update copyright info in application help to "Puppet Inc."

nfagerlund:
This commit also updates a spec test that was allowing "Puppet Labs" as a
copyright name.


joshcooper:
It looks like something is still hardcoded to "Puppet Labs".

nfagerlund:
Nice, nice. P. sure I can find that.

nfagerlund:
found it.

PR #5878, opened 2017-05-17
(PUP-7558) Remove unmaintained editor extensions

Sharpie:
This commit removes two editor extensions that pre-date the Puppet 4 language:

  - ext/emacs/: No meaningful update since 2014
  - ext/vim/: No meaningful update since 2009

Extensions such as this make more sense as separate projects so that updates
can be shipped independently from the Puppet release cycle.


hlindberg:
I have wanted to do this for quite some time. Thanks @Sharpie !

PR #5882, opened 2017-05-18
(PUP-7155) Make Parameter#format_value_for_display use StringConverter

thallgren:
The `Parameter#format_value_for_display` method had many issues. Strings
containing quotes would be emitted without escaping the quotes. All
values would be quoted (including numbers, booleans, and objects). The
output, allegedly "formatted fo display to a human" did not include
newlines and indentation. Output was sorted alphabetically although
Puppet specifies hashes as ordered.

This commit replaces the body of the method with a call to the
`StringConverter` class which is designed to address all described
issues.

### DIFF
--- a/acceptance/tests/resource/user/utf8_user_comments.rb
+++ b/acceptance/tests/resource/user/utf8_user_comments.rb
@@ -26,7 +26,9 @@ test_name 'PUP-6777 Manage users with UTF-8 comments' do
   # 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
   # 4-byte 𠜎 - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
   mixed_utf8_0 = "A\u06FF"
+  reported_mixed_utf8_0 = '"A\\\\u\\{6FF\\}"'
   mixed_utf8_1 = "\u16A0\u{2070E}"
+  reported_mixed_utf8_1 = '"\\\\u\\{16A0\\}\\\\u\\{2070E\\}"'
 
   teardown do
     # remove user on all agents
@@ -55,7 +57,7 @@ test_name 'PUP-6777 Manage users with UTF-8 comments' do
     EOF
     # Note setting LANG='<encoding>' environment like this has no effect on Windows agents.
     apply_manifest_on(agents, set_comment_utf8, :expect_changes => true, :environment => {:LANG => "en_US.UTF-8"}) do |result|
-      assert_match(/changed '#{mixed_utf8_0}' to '#{mixed_utf8_1}'/, result.stdout, "failed to modify UTF-8 user comment in UTF-8 environment")
+      assert_match(/changed #{reported_mixed_utf8_0} to #{reported_mixed_utf8_1}/, result.stdout, "failed to modify UTF-8 user comment in UTF-8 environment")
     end
   end
 
@@ -79,7 +81,7 @@ test_name 'PUP-6777 Manage users with UTF-8 comments' do
     EOF
     # Since LANG=<'encoding'> has no effect, this test is redundant on Windows - exclude it.
     apply_manifest_on(agents - windows_agents, set_comment_utf8, :expect_changes => true, :environment => {:LANG => "en_US.ISO8859-1"}) do |result|
-      assert_match(/changed '#{mixed_utf8_0}' to '#{mixed_utf8_1}'/, result.stdout, "failed to modify UTF-8 user comment in ISO-8859-1 environment")
+      assert_match(/changed #{reported_mixed_utf8_0} to #{reported_mixed_utf8_1}/, result.stdout, "failed to modify UTF-8 user comment in ISO-8859-1 environment")
     end
   end
 
@@ -105,7 +107,7 @@ test_name 'PUP-6777 Manage users with UTF-8 comments' do
     # OS X is known broken in POSIX locale with UTF-8 chars on OS X, so exclude OS X here.
     # Also since LANG=<'encoding'> has no effect, this test is redundant on Windows - exclude it.
     apply_manifest_on(agents - osx_agents - windows_agents, set_comment_utf8, :expect_changes => true, :environment => {:LANG => "POSIX"}) do |result|
-      assert_match(/changed '#{mixed_utf8_0}' to '#{mixed_utf8_1}'/, result.stdout, "failed to modify UTF-8 user comment with POSIX environment")
+      assert_match(/changed #{reported_mixed_utf8_0} to #{reported_mixed_utf8_1}/, result.stdout, "failed to modify UTF-8 user comment with POSIX environment")
     end
   end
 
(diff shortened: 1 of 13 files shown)


acceptance/tests/resource/user/utf8_user_comments.rb
```diff
@@ -55,7 +57,7 @@
     EOF
     # Note setting LANG='<encoding>' environment like this has no effect on Windows agents.
     apply_manifest_on(agents, set_comment_utf8, :expect_changes => true, :environment => {:LANG => "en_US.UTF-8"}) do |result|
-      assert_match(/changed '#{mixed_utf8_0}' to '#{mixed_utf8_1}'/, result.stdout, "failed to modify UTF-8 user comment in UTF-8 environment")
+      assert_match(/changed #{reported_mixed_utf8_0} to #{reported_mixed_utf8_1}/, result.stdout, "failed to modify UTF-8 user comment in UTF-8 environment")
```
joshcooper: The current and desired values should remain quoted in the output.

joshcooper:
We need to make sure this doesn't break the puppetlabs-acl module, or at least understand what will break before we merge.

acceptance/tests/resource/user/utf8_user_comments.rb
thallgren: They are. The quote is just moved to the variables.

thallgren:
I'd appreciate some help testing with the puppetlabs-acl module. Not sure how to set that up.

joshcooper:
The beaker CI job is https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/acl/view/master/job/forge-windows_puppetlabs-acl_intn-sys_smoke-master/1/PLATFORM=windows2012r2-64default.a-redhat7-64mdca,WORKER_LABEL=beaker/

and it looks like you can configure the module to use a specific puppet-agent SHA by setting:

```
PUPPET_INSTALL_TYPE=agent
PUPPET_AGENT_SHA=<full SHA>
PUPPET_AGENT_SUITE_VERSION=<git describe>
```

That assumes you have a puppet-agent build with your changes. See https://github.com/puppetlabs/beaker-puppet_install_helper/blob/master/lib/beaker/puppet_install_helper.rb#L34-L38

thallgren:
How do I go about building my own puppet agent? Also, the instructions mention a "development repository". Where do I specify my fork?

joshcooper:
Unfortunately it looks like this breaks the acl module. To repro from your branch (on a Windows host):

```
C:\work\puppet>bundle install --path .bundle
C:\work\puppet>bundle exec puppet module install puppetlabs-acl
C:\work\puppet>bundle exec puppet resource acl c:\temp\foo
acl { 'c:\temp\foo':
  group                      => 'S-1-5-21-2397885826-1833024046-1055597067-513',
  inherit_parent_permissions => true,
  owner                      => 'S-1-5-32-544',
  permissions                => [
  {

  },
  {

  },
  {

  },
  {

  },
  {

  },
  {

  }],
}
```

It emits the following in puppet 4.9.4:

```
C:\work\puppet>bundle exec puppet --version
4.9.4
C:\work\puppet>bundle exec puppet resource acl c:\temp\foo
acl { 'c:\temp\foo':
  group                      => 'S-1-5-21-2397885826-1833024046-1055597067-513',
  inherit_parent_permissions => 'true',
  owner                      => 'S-1-5-32-544',
  permissions                => [{'identity' => 'BUILTIN\Users', 'rights' => ['full']}, {'identity' => 'BUILTIN\Administrators', 'is_inherited' => 'true', 'rights' => ['full']}, {'identity' => 'NT AUTHORITY\SYSTEM', 'is_inherited' => 'true', 'rights' => ['full']}, {'identity' => 'BUILTIN\Users', 'is_inherited' => 'true', 'rights' => ['read', 'execute']}, {'child_types' => 'containers', 'identity' => 'BUILTIN\Users', 'is_inherited' => 'true', 'mask' => '4', 'rights' => ['mask_specific']}, {'affects' => 'children_only', 'identity' => 'CREATOR OWNER', 'is_inherited' => 'true', 'rights' => ['full']}],
}
```

thallgren:
The fact that `'true'` becomes `true` is positive, and in line with the expectations. The empty hashes however, is very bad. It's caused by a class derived from `Hash` and reveals a bug in the type system inference. I created [PUP-7587](https://tickets.puppetlabs.com/browse/PUP-7587) to track that problem. Marking this PR as blocked until that has been fixed.

joshcooper:
The `Ace < Hash` logic also got us into trouble in ruby 2.4 as they changed how Hash subclasses are YAML serialized.

thallgren:
This PR is now rebased on top of PUP-7587 and no longer exhibits the problems with the `puppetlabs-acl` module. Well, aside from the difference that the boolean is no longer being quoted, that is, but that's intentional.

joshcooper:
I don't think it's quite right...

```
C:\work\puppet>bundle exec puppet resource acl c:\temp\foo
acl { 'c:\temp\foo':
  group                      => 'S-1-5-21-2397885826-1833024046-1055597067-513',
  inherit_parent_permissions => true,
  owner                      => 'S-1-5-32-544',
  permissions                => ["\n { identity => 'BUILTIN\\Users', rights => [\"full\"] }","\n { identity => 'BUILTIN\\Administrators', rights => [\"full\"], is_inherited => 'true' }","\n { identity => 'NT AUTHORITY\\SYSTEM', rights => [\"full\"], is_inherited => 'true' }","\n { identity => 'BUILTIN\\Users', rights => [\"read\", \"execute\"], is_inherited => 'true' }","\n { identity => 'BUILTIN\\Users', rights => [\"mask_specific\"], mask => '4', child_types => 'containers', is_inherited => 'true' }","\n { identity => 'CREATOR OWNER', rights => [\"full\"], affects => 'children_only', is_inherited => 'true' }"],
}
```

Note how it's emitting an array containing a single string with all of the nested data `["\n { identity ...]`, instead of an array of hashes `[ { 'identity' => ...`

thallgren:
It's not a single string. It's one string for each contained object.
```
[
  "\n { identity => 'BUILTIN\\Users', rights => [\"full\"] }",
  "\n { identity => 'BUILTIN\\Administrators', rights => [\"full\"], is_inherited => 'true' }",
  ...
```
The String representation is defined by `Puppet::Type::Acl::Ace#inspect` of which `#to_s` is an alias. The old code discovered that this was a hash and iterated the keys of that hash. That just happens to work because the method '#keys' is implemented as:
```
  def keys
    to_hash.keys
  end
```
Other methods, such as `#each`, `#each_pair`, or `map` are however not implemented (the act on the empty hash) so it doesn't really "quake" as a Hash.

The sane thing to do with an instance of an unknown class is to output a quoted version of its string representation. That's what happens now but perhaps it could be improved by adding a rule stating that:
1. When the object is an instance of a class that is derived from Hash and,
2. the class is not registered as a known Pcore Type and,
3. the object respond to '#to_hash`, then
4. tepresent the result of calling `#to_hash` rather than the object itself.

We could do the same thing with derived arrays and '#to_a`

I'll provide a commit with a fix for that to this PR.

joshcooper:
Looks good:

```
C:\work\puppet>bundle exec puppet resource acl c:\temp\foo
acl { 'c:\temp\foo':
  group                      => 'S-1-5-21-2397885826-1833024046-1055597067-513',
  inherit_parent_permissions => true,
  owner                      => 'S-1-5-32-544',
  permissions                => [
  {'identity' => "BUILTIN\\Users", 'rights' => ['full']},
  {'identity' => "BUILTIN\\Administrators", 'rights' => ['full'], 'is_inherited' => true},
  {'identity' => "NT AUTHORITY\\SYSTEM", 'rights' => ['full'], 'is_inherited' => true},
  {'identity' => "BUILTIN\\Users", 'rights' => ['read', 'execute'], 'is_inherited' => true},
  {'identity' => "BUILTIN\\Users", 'rights' => ['mask_specific'], 'mask' => '4', 'child_types' => 'containers', 'is_inhe
rited' => true},
  {'identity' => 'CREATOR OWNER', 'rights' => ['full'], 'affects' => 'children_only', 'is_inherited' => true}],
}
```

PR #5884, opened 2017-05-18
(PUP-7572) Tag Tests breaking Image Acceptance

jcoconnor:
These tests break the Image Acceptance Test suite, so are being tagged
so that they can be excluded pending further investigation as part of
ticket PUP-7521

This was labelelled IMAGES-555 but Travis is unable to accept this as PR
prefix.

This replaces #5860


jcoconnor:
@adrienthebo I have removed the mega-edit as suggested.
Was talking to @pinkypie who is going to look at the possibility of fixing beaker so that tags are not required in all tests in the suite.

jcoconnor:
For reference, the tagging ticket is [BKR-676](https://tickets.puppetlabs.com/browse/BKR-676)

PR #5887, opened 2017-05-18
(PUP-7554) Don't error on v3 hiera.yaml in module

reidmv:
A v3 hiera.yaml in the root of a module directory should be a warning,
not an error. It should be roughly equivalent to not having a hiera.yaml
file there at all. This is to allow customers who currently have
hiera.yaml files in modules to install new versions of Puppet without
having to first refactor all their modules and CI systems.


Iristyle:
Given this fix should go into LTS (currently `4.10.2`), lets retarget at `stable` branch.

@hlindberg / @thallgren can we get some eyes on this?

Thanks!

PR #5888, opened 2017-05-18
(PUP-7575) Add --job-id flag to `puppet agent`

nicklewis:
Commit 4e05e16 added internal handling of job id for agent catalog
requests and report submission, but had no facility to actually provide
the job id to the agent. This commit now adds a --job-id flag which will
simply thread the provided value (which is an arbitrary string) through
to the catalog request and the report. This will allow correlating node
run reports with orchestration jobs in PuppetDB.


lib/puppet/application/agent.rb
```diff
@@ -71,6 +71,10 @@ def preinit
     options[:waitforcert] = arg.to_i
   end
 
+  option("--job-id ID") do |arg|
+    options[:job_id] = arg
```
MosesMendoza: just to confirm - this is OK as any kind of string, ie UTF-8 will make through the system ok?

nicklewis: Yeah, UTF-8 should be fine. The contract is that it's an opaque string, though in practice it's just an integer [stringified] today and perhaps some day a uuid.

MosesMendoza: and no handlingdownstream ie in orchestrator will have any side-effects / security risks from taking in this string as-is?

nicklewis: This data isn't intended to be particularly trusted. Agents could spoof having been part of a job, but only in PuppetDB. Orchestrator is the official record for which jobs happened, and there's no way to forge that. Threading the data through PuppetDB is a convenience to expose it via PQL, etc.

The usage pattern mostly involves querying PuppetDB to find the job you care about (I'm looking at a report and want to know if it was part of a job) and then querying orchestrator to find out what happened in that job. If a node were to lie about being part of the job, you would learn that when asking orchestrator.

MosesMendoza: Generally +1 just want to confirm there's no way this string could somehow get serialized / deserialized unsafely by the master

PR #5889, opened 2017-05-18
(PUP-6060) set node cache terminus to nil

MosesMendoza:
(PUP-6060) stop setting node_cache_terminus

This PR stops overriding the default `node_cache_terminus` for lookup and master applications to `write_only_yaml` and removes the setting of Puppet::Node's indirection cache_class. In the case of the master application, with this, Puppet::Node's indirection won't cache by default and so won't use the write_only_yaml cache. Per PUP-6060, with the advent of PuppetDB, workflows using lists of nodes are better off querying it than these yaml files on the master, and since this was a `write_only` cache we were always querying the node terminus for nodes anyway. With respect to the lookup application, while we set the `node_cache_terminus` setting, we never set to the `Puppet::Node` indirection cache_class to its value, which meant it was never set at all - the code removed from lookup never had any effect. This PR is largely cleanup - for our current deployments, puppet server deploys its own version of the master application that also sets `node_cache_terminus`. A separate PR is raised in puppet-server to remove it there as well. Since there are no more callers to this terminus, we deprecate it.

### DIFF
--- a/lib/puppet/application/master.rb
+++ b/lib/puppet/application/master.rb
@@ -135,15 +135,8 @@ Copyright (c) 2012 Puppet Inc., LLC Licensed under the Apache 2.0 License
     HELP
   end
 
-  # Sets up the 'node_cache_terminus' default to use the Write Only Yaml terminus :write_only_yaml.
-  # If this is not wanted, the setting ´node_cache_terminus´ should be set to nil.
-  # @see Puppet::Node::WriteOnlyYaml
-  # @see #setup_node_cache
-  # @see puppet issue 16753
-  #
   def app_defaults
     super.merge({
-      :node_cache_terminus => :write_only_yaml,
       :facts_terminus => 'yaml'
     })
   end
@@ -256,11 +249,10 @@ Copyright (c) 2012 Puppet Inc., LLC Licensed under the Apache 2.0 License
     Puppet::SSL::Oids.load_custom_oid_file(Puppet[:trusted_oid_mapping_file])
   end
 
-  # Sets up a special node cache "write only yaml" that collects and stores node data in yaml
-  # but never finds or reads anything (this since a real cache causes stale data to be served
-  # in circumstances when the cache can not be cleared).
-  # @see puppet issue 16753
-  # @see Puppet::Node::WriteOnlyYaml
+  # Honor the :node_cache_terminus setting if users have specified it directly.
+  # We normally want this nil as use-cases for querying nodes should be going to
+  # PuppetDB.
+  # @see PUP-6060
   # @return [void]
   def setup_node_cache
     Puppet::Node.indirection.cache_class = Puppet[:node_cache_terminus]
--- a/lib/puppet/indirector/node/write_only_yaml.rb
+++ b/lib/puppet/indirector/node/write_only_yaml.rb
@@ -11,7 +11,12 @@ require 'puppet/indirector/yaml'
 # @api private
 #
 class Puppet::Node::WriteOnlyYaml < Puppet::Indirector::Yaml
-  desc "Store node information as flat files, serialized using YAML,
+  def initialize
+    Puppet.warn_once(:deprecation, 'Puppet::Node::WriteOnlyYaml', _('Puppet::Node::WriteOnlyYaml is deprecated and will be removed in a future release of Puppet.'))
+    super
+  end
+
+  desc "(Deprecated) Store node information as flat files, serialized using YAML,
     does not deserialize (write only)."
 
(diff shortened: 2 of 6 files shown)


MosesMendoza:
Relates to/review with https://github.com/puppetlabs/puppetserver/pull/1404

lib/puppet/application/master.rb
```diff
@@ -256,16 +249,6 @@ def setup_ssl
...
-  # Sets up a special node cache "write only yaml" that collects and stores node data in yaml
-  # but never finds or reads anything (this since a real cache causes stale data to be served
-  # in circumstances when the cache can not be cleared).
-  # @see puppet issue 16753
-  # @see Puppet::Node::WriteOnlyYaml
-  # @return [void]
-  def setup_node_cache
-    Puppet::Node.indirection.cache_class = Puppet[:node_cache_terminus]
-  end
-
```
joshcooper: If we remove this, then doesn't that mean you can't enable the `node_cache_terminus` (if you wanted the 4.x behavior)?

lib/puppet/indirector/node/write_only_yaml.rb
```diff
@@ -11,7 +11,12 @@
 # @api private
 #
 class Puppet::Node::WriteOnlyYaml < Puppet::Indirector::Yaml
-  desc "Store node information as flat files, serialized using YAML,
+  def initialize
+    Puppet.deprecation_warning('Puppet::Node::WriteOnlyYaml is deprecated and will be removed in a future release of Puppet.')
+    super
+  end
```
joshcooper: If we create a new WriteOnlyYaml terminus for each request, then this will spam the logs. Can we warn once (but also see comment below)?

lib/puppet/indirector/node/write_only_yaml.rb
```diff
@@ -11,7 +11,12 @@
...
+
+  desc "(Deprecated) Store node information as flat files, serialized using YAML,
```
joshcooper: Do we want to deprecate this?

lib/puppet/application/master.rb
MosesMendoza: Well, you wouldn't be able to do it just by by setting `node_cache_terminus`. But if you really wanted it, you could add it to routes.yaml:

```
master:
      facts:
        terminus: puppetdb
        cache: yaml
      node:
        cache: write_only_yaml
```

lib/puppet/indirector/node/write_only_yaml.rb
MosesMendoza: Sure! I figured since we're only loading this once per indirection/indirected class we wouldn't be too spammy, but agreed - best to ensure we only do it once.

lib/puppet/indirector/node/write_only_yaml.rb
MosesMendoza: I think so. My thinking was a) we're trying to move away from yaml serializing to disk (well really yaml entirely) and b) there are no users of this terminus in the codebase after this PR so why ship it?

lib/puppet/indirector/node/write_only_yaml.rb
MosesMendoza: Updated

adrienthebo:
Since we merged the Puppetserver version of this test we just got CI failures because of the write_only_yaml acceptance test. I've deleted that in https://github.com/puppetlabs/puppet/pull/5898.

lib/puppet/application/master.rb
MosesMendoza: Discussing with @joshcooper we're going to leave this method in to honor the setting value in the master application to enable a user to keep the value if they'd like to. That should be more discoverable/easier than routes.yaml.

MosesMendoza:
@joshcooper updated per discussion

lib/puppet/indirector/node/write_only_yaml.rb
```diff
@@ -11,7 +11,12 @@
 # @api private
 #
 class Puppet::Node::WriteOnlyYaml < Puppet::Indirector::Yaml
-  desc "Store node information as flat files, serialized using YAML,
+  def initialize
+    Puppet.warn_once(:deprecation, 'Puppet::Node::WriteOnlyYaml', _('Puppet::Node::WriteOnlyYaml is deprecated and will be removed in a future release of Puppet.'))
+    super
+  end
```
joshcooper: Verified only a single deprecation warning is emitted:

```
Warning: Puppet::Node::WriteOnlyYaml is deprecated and will be removed in a future release of Puppet.
   (file & line not available)
```

PR #5890, opened 2017-05-19
(PUP-7576) fix face_base arity check

MosesMendoza:
The exception raised when the wrong number of arguments is supplied to a face
command has an incorrect externalization that references an undefined variable,
`s`. This was a carry-over from a previous form that we replaced when it was
pluralized.


MosesMendoza:
filed pup-7576

PR #5891, opened 2017-05-19
(PUP-6517) Raise output and backtrace of rescued exception

joshcooper:
Previously if `execfail` method failed to execute the command, it would
try to access the `backtrace` method on the exception class passed into
the method instead of the backtrace of the rescued exception.

This commit marks the execfail method as deprecated, since it's
confusing more than anything, and modifies the method to rescue the
exception and re-raise the rescued exception message and backtrace. It
also adds tests for the execfail method as previously there were none.

### DIFF
--- a/lib/puppet/util/execution.rb
+++ b/lib/puppet/util/execution.rb
@@ -98,11 +98,11 @@ module Puppet::Util::Execution
   # @raise [exception] under same conditions as {execute}, but raises the given `exception` with the output as argument
   # @return (see execute)
   # @api public
+  # @deprecated
   def self.execfail(command, exception)
-    output = execute(command)
-    return output
-  rescue Puppet::ExecutionFailure
-    raise exception, output, exception.backtrace
+    execute(command)
+  rescue Puppet::ExecutionFailure => detail
+    raise exception, detail.message, detail.backtrace
   end
 
   # Default empty options for {execute}
--- a/spec/unit/util/execution_spec.rb
+++ b/spec/unit/util/execution_spec.rb
@@ -675,13 +675,42 @@ describe Puppet::Util::Execution do
     it "should fail if asked to fail, and the child does" do
       Puppet::Util::Execution.stubs(:open).with('| echo hello 2>&1').returns('error message')
       Puppet::Util::Execution.expects(:exitstatus).returns(1)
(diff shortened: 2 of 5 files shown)


spec/unit/util/execution_spec.rb
```diff
@@ -675,13 +675,35 @@ def stub_process_wait(exitstatus)
...
+      }.to raise_error(Puppet::Error, /failed to execute/)
+    end
+
+    it "raises exceptions that don't extend ExecutionFailure" do
+      Puppet::Util::Execution.stubs(:execute).raises(ArgumentError, "failed to execute")
+      expect {
+        Puppet::Util::Execution.execfail("this will fail", Puppet::Error)
+      }.to raise_error(ArgumentError, /failed to execute/)
+    end
+  end
```
Iristyle: What about when the `exception` argument is `nil`?

lib/puppet/util/execution.rb
```diff
@@ -98,11 +98,11 @@ def self.exitstatus
   # @raise [exception] under same conditions as {execute}, but raises the given `exception` with the output as argument
   # @return (see execute)
   # @api public
+  # @deprecated
```
Iristyle: Do you also want to deprecate the provider wrappers?
https://github.com/puppetlabs/puppet/blob/ebd96213cab43bb2a8071b7ac0206c3ed0be8e58/lib/puppet/provider.rb#L120-L129

In our code, looks like RPM provider (and its derivatives) are the only callers - https://github.com/puppetlabs/puppet/blob/8a6dd1b9034e0d7e403c524fe74b297825b6a93a/lib/puppet/provider/package/rpm.rb#L135

joshcooper: @Iristyle I deprecated the provider wrappers and modified the rpm provider to not call the deprecated method.

Iristyle: Two other quick notes:

* we agreed that it's fine to skip deprecation messages here for end users - given the inline docs should be enough (noting that `execfail` is an external API given the `@api public`)
* we'll add a ticket for PUP 6 removal

Iristyle: Removal ticket filed at https://tickets.puppetlabs.com/browse/PUP-7584

PR #5892, opened 2017-05-19
(PUP-7471) better error message

DavidS:


lib/puppet/pops/parser/egrammar.ra
```diff
@@ -656,6 +656,7 @@ classname
   | WORD
   | CLASSREF
   | CLASS { error val[0], "'class' is not a valid classname" }
+  | STRING { error val[0], "classnames must be bare words, not strings" }
```
hlindberg: This is not just for the name of a class - this rule accepts `Foo`, `Foo::Bar`, `foo`, `foo::bar`, `not-name-but-word` and it is used for classes, defines, functions, etc.
Suggest changing this to be more generic "A String was found where it is not accepted"

hlindberg: And while at it, also improve the error for the line above to be similar.

hlindberg: simpler "A quoted string is not valid as a name here"

DavidS:
Updated the error message as requested.

lib/puppet/pops/parser/egrammar.ra
```diff
@@ -655,7 +655,8 @@ classname
   : NAME
   | WORD
   | CLASSREF
-  | CLASS { error val[0], "'class' is not a valid classname" }
+  | CLASS { error val[0], "'class' keyword not allowed at this location" }
+  | STRING { error val[0], "a string was found where it is not accepted" }
```
hlindberg: you forgot to update the tests. Also, my second suggestion is better as it includes the information that the error is about a "name"

hlindberg: my suggestion was to use "A quoted string is not acceptable as a name" - for that it is clear that it is expecting a name but got a quoted string (as opposed to a bare word which is also a string).

DavidS:
Reading what you wrote would definitely have helped. Updated.

PR #5893, opened 2017-05-19
(PUP-7579) Allow tags to contain Unicode words

Iristyle:
- The regex for tags was originally /^\w[-\w:]*$/ in a Luke commit
    from 10 years ago f98be4a7198326b26f1072c401b3e337f340db40

  - Over the years it has received updates to modernize it and has
    shuffled around.

  - 0badfce72bafc767bf3099718c63324f51108996 changed it from

    /^\w[-\w:.]*$/

    to

    /^[0-9A-Za-z_][0-9A-Za-z_:.-]*$/

    for performance reasons.

  - The concat module depends on tags, and as such requires that tags
    can contain other valid content.  The regex here is too
    restrictive for that purpose and should be returned to its
    original intent - word characters.

    However, it now needs to be updated to accept word characters in
    any language.

  - This now allows for the follow manifest to not cause an error:

  puppet apply -e 'notify { "hi": tag => "_メインディレクトリ_醸造所"  }'

### DIFF
--- a/lib/puppet/util/tagging.rb
+++ b/lib/puppet/util/tagging.rb
@@ -1,7 +1,7 @@
 require 'puppet/util/tag_set'
 
 module Puppet::Util::Tagging
-  ValidTagRegex = /\A[0-9A-Za-z_][0-9A-Za-z_:.-]*\Z/
+  ValidTagRegex = /\A[[:alnum:]_][[:alnum:]_:.-]*\Z/
 
   # Add a tag to the current tag set.
   # When a tag set is used for a scope, these tags will be added to all of
--- a/spec/unit/util/tagging_spec.rb
+++ b/spec/unit/util/tagging_spec.rb
@@ -48,6 +48,61 @@ describe Puppet::Util::Tagging do
     expect { tagger.tag("good.tag") }.to_not raise_error
   end
 
+  # different UTF-8 widths
+  # 1-byte A
+  # 2-byte ۿ - http://www.fileformat.info/info/unicode/char/06ff/index.htm - 0xDB 0xBF / 219 191
+  # 3-byte ᚠ - http://www.fileformat.info/info/unicode/char/16A0/index.htm - 0xE1 0x9A 0xA0 / 225 154 160
+  # 4-byte ܎ - http://www.fileformat.info/info/unicode/char/2070E/index.htm - 0xF0 0xA0 0x9C 0x8E / 240 160 156 142
+  let (:mixed_utf8) { "A\u06FF\u16A0\u{2070E}" } # Aۿᚠ܎
+
+  it "should allow UTF-8 alphanumeric characters" do
+    expect { tagger.tag(mixed_utf8) }.not_to raise_error
+  end
+
+  # completely non-exhaustive list of a few UTF-8 punctuation characters
+  # http://www.fileformat.info/info/unicode/block/general_punctuation/utf8test.htm
+  [
+    "\u2020", # dagger †
+    "\u203B", # reference mark ※
+    "\u204F", # reverse semicolon ⁏
+    "!",
+    "@",
+    "#",
+    "$",
+    "%",
+    "^",
+    "&",
+    "*",
+    "(",
+    ")",
+    "-",
+    "+",
+    "=",
+    "{",
+    "}",
+    "[",
+    "]",
+    "|",
+    "\\",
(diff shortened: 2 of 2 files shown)


Iristyle:
Would be nice to have a few tests that ensured no regressions.

lib/puppet/util/tagging.rb
```diff
@@ -1,7 +1,7 @@
 require 'puppet/util/tag_set'
 
 module Puppet::Util::Tagging
-  ValidTagRegex = /\A[0-9A-Za-z_][0-9A-Za-z_:.-]*\Z/
+  ValidTagRegex = /\A[[:word:]][[[:word:]]:.-]*\Z/
```
thallgren: I'm not overly fond of [:word:] since it's a non-standard construct. Documentation states that it includes everything in the general Unicode category 'Mark' and that doesn't seems right. Also, I think there are one pair of brackets too many in the expression. How about this?
```
/\A[[:alnum:]_][[:alnum:]_:.-]*\z/
```

thallgren: Please add tests.

Iristyle: The `[[` / `]]` brackets look weird, but are correct - see https://ruby-doc.org/core-2.1.1/Regexp.html#class-Regexp-label-Character+Classes

I can give `:alnum:` a shot with some test data to see what happens.  We're mostly concerned with making tags compatible with the concat module for https://tickets.puppetlabs.com/browse/MODULES-4864.

thallgren: The extra brackets are redundant. The example you point at is already a character group. You don't need an extra set of brackets if you're already in one when the objective is to add more characters to the same character group.
The appointed ticket seems somewhat unrelated.

Iristyle: Ah ah, you were referring to the `[[[:word:]]:.-]` being rewritable as `[[:word:]:.-]` - I see. 
 Yes, that's fine.  

I thought you were referring to the first `[[:word:]]` which cannot have the double brackets removed: - i.e. 

```
[1] pry(main)> "abc" =~ /[:word:]/
=> nil
[2] pry(main)> "abc" =~ /[[:word:]]/
=> 0
```

As for the modules ticket, I've been told that concat makes extensive use of tags, and without tag names fully supporting UTF-8, concat explodes - so this is the root issue (at least with how concat is currently implemented).  I haven't looked at the concat code nor am I aware if it can be reasonably rewritten without tags.

thallgren: OK, but that's probably [PUP-4684](https://tickets.puppetlabs.com/browse/MODULES-4684), not 4864 ;-)

Iristyle: Yes!  Thanks

https://tickets.puppetlabs.com/browse/MODULES-4684

joshcooper: We'll need to update docs: https://docs.puppet.com/puppet/4.10/lang_reserved.html#tags

MosesMendoza:
PuppetDB also appears to validate tags against a similar schema: https://github.com/puppetlabs/puppetdb/blob/master/src/puppetlabs/puppetdb/catalogs.clj#L304-L320

lib/puppet/util/tagging.rb
Iristyle: With @MosesMendoza confirming that PuppetDB code uses RegEx for tag validation as well, I agree that we should not use the Ruby `[[:word:]]` @thallgren and should instead rely on the POSIX compatible `[[:alnum:]]`, which has a Java equivalent in https://docs.oracle.com/javase/6/docs/api/java/util/regex/Pattern.html as `\p{Alnum}`.

/cc @mullr @rbrw

Iristyle:
PR has been updated and should be ready for merge.  @puppetlabs/puppetdb should make corresponding changes to coincide with these.

Iristyle:
@MFredette if you could point me to the docs I can put up a pull request there as well.  Thanks!

joshcooper:
Is there a PDB ticket?

Iristyle:
Looks like the documentation updates should land at https://github.com/puppetlabs/puppet-docs/blob/master/source/puppet/5.0/lang_reserved.markdown

Iristyle:
Update doc PR in #6016

PR #5895, opened 2017-05-22
(PUP-1606) Deprecate virtual class

hlindberg:
This adds warning/error under the control of --strict for resource stye class inclusion and `create_resources` when an attempt is made to use `@class` or `@@class` as neither virtual nor exported works with classes.

### DIFF
--- a/lib/puppet/pops/issues.rb
+++ b/lib/puppet/pops/issues.rb
@@ -368,7 +368,7 @@ module Issues
     "Resource Defaults are not virtualizable"
   end
 
-  CLASS_NOT_VIRTUALIZEABLE = issue :CLASS_NOT_VIRTUALIZEABLE do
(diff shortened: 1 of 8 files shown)


lib/puppet/pops/issues.rb
```diff
@@ -368,6 +368,10 @@ def self.hard_issue(issue_code, *args, &block)
     "Resource Defaults are not virtualizable"
   end
 
+  CLASS_NOT_VIRTUALIZEABLE = issue :CLASS_NOT_VIRTUALIZEABLE do
```
thallgren: I think the spelling should be CLASS_NOT_VIRTUALIZABLE to be consistent with its own text (also, "virtualizable" gives me 22.200 hits on Google. "virtualizeable" just 1.420).

hlindberg: meh - fixing

PR #5896, opened 2017-05-22
(PUP-1795) Warn when numeric auto convert takes place

hlindberg:
See commit message

### DIFF
--- a/lib/puppet/pops/issues.rb
+++ b/lib/puppet/pops/issues.rb
@@ -506,6 +506,10 @@ module Issues
     "The value '#{value}' cannot be converted to Numeric."
   end
 
+  NUMERIC_COERCION = issue :NUMERIC_COERCION, :before, :after do
+    "The string '#{before}' was automatically coerced to the numerical value #{after}"
+  end
+
   UNKNOWN_FUNCTION = issue :UNKNOWN_FUNCTION, :name do
     "Unknown function: '#{name}'."
   end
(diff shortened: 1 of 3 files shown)


lib/puppet/pops/issues.rb
```diff
@@ -506,6 +506,10 @@ def self.hard_issue(issue_code, *args, &block)
     "The value '#{value}' cannot be converted to Numeric."
   end
 
+  NUMERIC_COERCION = issue :NUMERIC_COERCION, :before, :after do
+    "The string '#{before}' was automatically coerced to the numerical value #{after}"
```
Magisus: Hmm... So in master, all of these errors have been externalized. How do we want to handle making sure that new errors also get externalized? Is that something that should happen post-mergeup? And if so, how are we going to remember to do it?

thallgren:
Closing and reopening to make Travis choose correct base branch.

lib/puppet/pops/issues.rb
hlindberg: A very good question

hlindberg:
Note that the string in issues.rb must be externalized when merging up to master

hlindberg:
travis build was failed after 5s - restarted build to see if that helps

thallgren:
The travis build is back to thinking that this PR should be based on master. Not sure why that happens. I'm closing and reopening again to see if that has an effect (it worked yesterday).

PR #5898, opened 2017-05-22
(PUP-6060) Manually set node_cache_terminus for woy test

adrienthebo:
The write_only_yaml cache test needs to explicitly set the node terminus as the default value is changing in Puppet 5.

### DIFF
--- a/acceptance/tests/node/check_woy_cache_works.rb
+++ b/acceptance/tests/node/check_woy_cache_works.rb
@@ -120,6 +120,7 @@ master_opts = {
   'master' => {
     'rest_authconfig' => authfile,
     'yamldir' => temp_yamldir,
+    'node_cache_terminus' => 'write_only_yaml',
   }
 }
 


camlow325:
Should we delete the test outright or modify it to enable the write_only_yaml cache through settings - assuming that will still be possible?  For PUP-6060, I wasn't sure if we were removing the ability to use the write_only_yaml cache or just not configuring it to be used by default.

adrienthebo:
I assumed/hoped that we were deleting it outright, but you caught that - I didn't check. Looks like it's slated for removal in Puppet 6. I'll update the test accordingly.

adrienthebo:
Commit rewritten to set `'node_terminus' => 'write_only_yaml'`. I'm testing this against Puppetserver now and will comment on the results of that when it completes.

adrienthebo:
I've verified these changes locally against puppetserver; since this is explicitly setting a value that previously used a default it should be safe to merge before https://github.com/puppetlabs/puppet/pull/5889.

PR #5899, opened 2017-05-22
(PUP-7528) Update Windows ci:test:git to Ruby 2.4

Magisus:
Now that Ruby 2.4 builds have been added to puppet-win32-ruby, this
commit updates puppet to use Ruby 2.4 when running acceptance tests on
Windows via ci:test:git.

Note that some tests are still failing when run this way. These failures
are being investigated separately.


Iristyle:
@Magisus could we note in commit message that this only impacts Windows?  For other operating systems, we use the distro version (which is why some older OSes like RHEL6 no longer work, given their distro version is 1.8.7)

Magisus:
Updated.

PR #5903, opened 2017-05-23
(PUP-6895) Specify unique command for Python pip on Windows

tkishel:
On Windows:

The package provider for Python's pip fails when Perl's pip is installed
and Perl is first in the PATH, sincle both install a pip command.

This fix specifies a unique command (pip.exe) for Python pip on Windows.

See also:

https://www.penwatch.net/cms/pip_conflict/
https://bugzilla.redhat.com/show_bug.cgi?id=958377

### DIFF
--- a/spec/unit/provider/package/pip_spec.rb
+++ b/spec/unit/provider/package/pip_spec.rb
@@ -2,11 +2,11 @@
 require 'spec_helper'
 
 provider_class = Puppet::Type.type(:package).provider(:pip)
-osfamilies = { ['All', nil] => ['pip', 'pip-python'] }
+osfamilies = { 'windows' => ['pip.exe'], 'other' => ['pip', 'pip-python'] }
 
 describe provider_class do
 
-  before do
+  before do  
     @resource = Puppet::Resource.new(:package, "fake_package")
     @provider = provider_class.new(@resource)
     @client = stub_everything('client')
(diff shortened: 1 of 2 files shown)


tkishel:
I've tried adding 'pip.exe' to  pip_spec.rb. 
This succeeds:

```
  describe "cmd" do
    it "should return pip.exe on Windows" do
      Puppet.features.stubs(:microsoft_windows?).returns true
      expect(provider_class.cmd[0]).to eq('pip.exe')
    end

   ...
```

I have not been able to do the same for the `instances` and `lazy_pip` tests that use `osfamilies`.

Magisus:
I've been digging into this a bit too, it seems like what might be happening is the windows feature flag you've set up is still being evaluated in the test, and is coming up false. So it is choosing the second array as commands to look for, but the test stubs out `which` calls for these two to return `nil` here https://github.com/puppetlabs/puppet/blob/master/spec/unit/provider/package/pip_spec.rb#L55. So then it gets to this line https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/pip.rb#L33, `pip_cmd` is `nil`, and the method exits early, failing the test. I'm not sure what the right way forward is here, but apparently the Windows feature test you added doesn't work right in our testing environment (not sure about "in the real world"). @Iristyle thoughts?

tkishel:
I fixed the pip_spec tests.

Not sure about the security_spec tests which appear to be outside the scope of my changes.

Magisus:
It looks like Appveyor did its weird abort thing. I re-kicked it.

spec/unit/provider/package/pip_spec.rb
```diff
@@ -2,11 +2,11 @@
 require 'spec_helper'
 
 provider_class = Puppet::Type.type(:package).provider(:pip)
-osfamilies = { ['All', nil] => ['pip', 'pip-python'] }
+osfamilies = { 'windows' => ['pip.exe'], 'other' => ['pip', 'pip-python'] }
 
 describe provider_class do
 
-  before do
+  before do
```
SamK: are the extra spaces really needed?

tkishel: Of course they are not. Text editor noise.

PR #5909, opened 2017-05-24
(PUP-7590) Fix merge of PUP-1600 - symbols not used in AST

hlindberg:
When PUP-1600 was merged, the fact that the AST on master uses strings
for enums instead of symbols. This led to warning/error about attempts
to virtualize a class for perfectly valid code.

This went undetected because of a missing unit tests that tested that
the check put in place was not triggered by valid code. A test is now
added.


Magisus:
Thanks for adding the test! 👍

PR #5910, opened 2017-05-24
(PUP-7592) Disable unicode user comment test on AIX

branan:
The AIX providers have not yet been made unicode-aware,
so this test is not valid on that platform.


Magisus:
Thanks. We should really look into what it would take to actually fix this. 👍

branan:
the AIX user provider does some weird stuff - it's worth diving in to for sure, but I want to get stuff unblocked

branan:
@joshcooper updated to re-target stable

PR #5912, opened 2017-05-25
(PUP-7594) Hiera hash keys are being converted to strings recursively.

heliocentric:
Puppet 4.9+ currently converts hiera hash elements into strings unconditiionally. YAML.load_file() will deserialize say an Integer to a ruby Fixnum, but the function symkeys_to_string converts every key to a string, regardless of whether it is a symbol.

### DIFF
--- a/lib/puppet/pops/lookup/hiera_config.rb
+++ b/lib/puppet/pops/lookup/hiera_config.rb
@@ -106,7 +106,9 @@ class HieraConfig
   def self.symkeys_to_string(struct)
     case(struct)
     when Hash
-      Hash[struct.map { |k,v| [k.to_s, symkeys_to_string(v)] }]
+      map = {}
+      struct.each_pair {|k,v| map[ k.is_a?(Symbol) ? k.to_s : k] = symkeys_to_string(v) }
+      map
     when Array
       struct.map { |v| symkeys_to_string(v) }
     else
--- a/spec/unit/pops/lookup/lookup_spec.rb
+++ b/spec/unit/pops/lookup/lookup_spec.rb
@@ -57,6 +57,11 @@ describe 'The lookup API' do
           mod::c: mod::c (from module)
           mod::e: mod::e (from module)
           mod::f: mod::f (from module)
+          mod::g:
+            :symbol: value
+            key: value
+            6: value
+            2.7: value
           YAML
       }
     }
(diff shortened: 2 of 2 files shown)


lib/puppet/pops/lookup/hiera_config.rb
```diff
@@ -106,7 +106,18 @@ def self.config_exist?(config_root)
...
+        when "Symbol"
+          key = k.to_s
+        else
+          key = k
+        end
+        value = symkeys_to_string(v)
+        [key, value]
+      }
+
+      Hash[map]
```
hlindberg: The algorithm used is a slow way of mapping a hash (the original was too). It is faster to do as follows:
```
map = {}
struct.each {|k,v| map[ k.is_a?(Symbol) ? k.to_s : k] = symkeys_to_string(v) }
map
```
And - for the future note that a `case` with a value can match with its type against a class - in the PR it takes the class name as a String and then compares against `"Symbol"` instead of just matching the value against `Symbol`. 

We have found in benchmarks that it is faster to build up a hash key by key, than first creating an array of array hash entries and then creating a hash from that.

hlindberg: Ping @thallgren, not sure where exctly the `symkeys_to_string` is used - by its location it suggests that it is for hiera's configuration - but it seems to affect data. For the hiera.yaml config I wonder if it should keep `Integer` keys (the changed version would do that).

hlindberg:
The commit message should be written as an imperative statement (for example): "(PUP-7594) Change only Symbol hash keys when transforming hiera data"

hlindberg:
Thank you @heliocentric for reporting and submitting this PR. Are you up for making the requested changes? If so, please wait on input from @thallgren as well - there may be additional things that should be modified / things to consider.

hlindberg:
One thing that came to mind is that we probably want this for Puppet 4.10.2 - which means this PR should target the `stable` branch.

heliocentric:
yea, I can make the changes, I did it as 'correct' in terms of logic, but I wasn't familiar with performance considerations.

lib/puppet/pops/lookup/hiera_config.rb
thallgren: It does affect data as well. It is called from the `#yaml_data` and `#eyaml_lookup_key` functions and I suspect that if we remove those calls, then we will start seeing problems with users that have symbolic keys in their data. That's just a guess though. The change, optimized as suggested by @hlindberg, should be fine in any case. Perhaps we should consider removing the calls from `#yaml_data` and `#eyaml_lookup_key` too?

thallgren: Actually, I think this might be slightly faster still:
```
map = {}
struct.each_pair {|k,v| map[ k.is_a?(Symbol) ? k.to_s : k] = symkeys_to_string(v) }
map
```
Note the use of `#each_pair` instead of `#each`. This saves Ruby from creating and splashing an array in each iteration.

hlindberg: I am ok with transforming this way - to get rid of the symbols.

hlindberg:
@heliocentric awesome - do the optimized version thomas shows (using `each_pair`). Will you be adding a test as well - or do you want/need help with that?

heliocentric:
yea, I'll add a test case in, it shouldn't be an issue.

heliocentric:
Added a unit test, and verified it failed on the old behavior and passes with the changes. Rebased on top of stable branch. Implemented each_pair method.

hlindberg:
hm, it is still targeting master - will switch branch and see what happens.

PR #5913, opened 2017-05-25
(PUP-7508) improve mapped paths explain output - WIP - DO NOT MERGE

hlindberg:
This is W.I.P - DO NOT MERGE - PR is for review

See commit message for what this does.
The test I added does not work - the explain output that I produced does not show up in the output because the lookup_invocation in use does not have an explainer.
I thought it would have one - since the very similar test above tests that the output contains a "not found". Need help to resolve this. Ping @thallgren


hlindberg:
Closed because the approach is too simplistic - there is no explainer in action at the time the mapped paths are resolved. The information needs to propagate and we need to design how that should be done. Closing this rather than making a hacky implementation that only works for this one case. See ticket.

PR #5914, opened 2017-05-25
(PUP-3042) Make logged errors in report set status == failed

hlindberg:
Before this, a circular dependency would not alter the state of a report
for an agent run. This is bad because users are then not alerted to the
problem and believe that the catalog application was correct.

Now the report's status is set to 'failed' if there is any error
included in the logs of the report.

Logs are given to the report both before it is finalized and after. This
requires a check when computing the status at the time the report is
finalized, and when any additional runtime log entries are added to the
report.

A test is added for the specific error of circular dependencies. The
status will however be set to 'failed' for any logged error.

This alters two tests that assumed that it was ok to call << with a
string where it should have been an instance of Puppet::Util::Log. (The
`<<` method is api private.)

### DIFF
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -147,8 +147,10 @@ class Puppet::Transaction::Report
   end
 
   # @api private
-  def <<(msg)
-    @logs << msg
+  def <<(log_entry)
+    # runtime errors after completion of report are added after report has finalized
+    @status = 'failed' if log_entry.level == :err
+    @logs << log_entry
     self
   end
 
@@ -176,7 +178,7 @@ class Puppet::Transaction::Report
 
   # @api private
   def compute_status(resource_metrics, change_metric)
-    if resources_failed_to_generate || (resource_metrics["failed"] || 0) > 0
+    if resources_failed_to_generate || (resource_metrics["failed"] || 0) > 0 || @logs.any? { |lg| lg.level == :err }
       'failed'
     elsif change_metric > 0
       'changed'
--- a/spec/integration/transaction/report_spec.rb
+++ b/spec/integration/transaction/report_spec.rb
@@ -701,6 +701,16 @@ describe Puppet::Transaction::Report do
         expect(report.corrective_change).to eq(false)
         expect(get_cc_count(report)).to eq(0)
       end
+
+      it "a circular dependency is reported with status == failed" do
+        resources = [
+          Puppet::Type.type(:notify).new(:title => 'a', :before => 'Notify[b]'),
+          Puppet::Type.type(:notify).new(:title => 'b', :before => 'Notify[a]')
+          ]
+        report = run_catalogs(resources, resources, true, true)
+
+        expect(report.status).to eq('failed')
+      end
     end
   end
 end
--- a/spec/unit/transaction/report_spec.rb
+++ b/spec/unit/transaction/report_spec.rb
@@ -147,12 +147,14 @@ describe Puppet::Transaction::Report do
     end
 
     it "should add new logs to the log list" do
-      @report << "log"
-      expect(@report.logs[-1]).to eq("log")
+      lg = Puppet::Util::Log.new(:level => :err, :message => 'log')
+      @report << lg
+      expect(@report.logs[-1]).to eq(lg)
     end
 
     it "should return self" do
-      r = @report << "log"
+      lg = Puppet::Util::Log.new(:level => :err, :message => 'log')
+      r = @report << lg
       expect(r).to equal(@report)
     end
   end


Magisus:
Are we okay with the chance that this may cause additional situations to be flagged as errors that weren't before?

lib/puppet/transaction/report.rb
```diff
@@ -147,8 +147,10 @@ def as_logging_destination(&block)
   end
 
   # @api private
-  def <<(msg)
-    @logs << msg
+  def <<(log_entry)
+    # runtime errors after completion of report are added after report has finalized
+    @status = 'failed' if log_entry.level == :err
+    @logs << log_entry
```
joshcooper: One possible issue is https://tickets.puppetlabs.com/browse/PUP-1763. Currently, puppet can fail to pluginsync (because there isn't al least one module with a `lib` directory), and I think puppet will emit an error message as a result. But I think this change will make the report fail even though the agent run did succeed...

lib/puppet/transaction/report.rb
```diff
@@ -147,8 +147,10 @@ def as_logging_destination(&block)
   end
 
   # @api private
-  def <<(msg)
-    @logs << msg
+  def <<(log_entry)
+    # runtime errors after completion of report are added after report has finalized
+    @status = 'failed' if log_entry.level == :err
```
Iristyle: Agree with @Magisus to see what else sets `log_entry.level` to `:err` - I don't really know all the cases.

/cc @MosesMendoza

hlindberg:
On the ticket, several other issues that are also error states where commented on as possibly being solved by this fix.
As I understand the changes I made, this is only reporting the result as a failure, it does not actually fail the application of the catalog - it ran as much as it could. The failed status in the report will simply alert users (as I understand it).

lib/puppet/transaction/report.rb
```diff
@@ -176,7 +178,7 @@ def add_resource_status(status)
 
   # @api private
   def compute_status(resource_metrics, change_metric)
-    if resources_failed_to_generate || (resource_metrics["failed"] || 0) > 0
+    if resources_failed_to_generate || (resource_metrics["failed"] || 0) > 0 || @logs.any? { |lg| lg.level == :err }
```
Iristyle: So if you're making this change (which I guess is how we calculate status), being called from `finalize_report` at https://github.com/hlindberg/puppet/blob/36225c3c81c0f7a03f46d86802c6060814881d7f/lib/puppet/transaction/report.rb#L201

Then I don't believe the above changes to `def <<(msg)` are necessary, are they?

hlindberg: yes, very much needed - see the commit comment. There are additional things that happen after the report is finalized.

Iristyle: Ouch, we still allow logs after finalizing a report?  That seems... wrong.

I would expect that once finalize is called the report becomes immutable - maybe something to file a ticket in the `Improve Agent Lifecycle` epic to fix in the future.

joshcooper: This is a side-effect of the report being a log destination. So after we finialize the report, any log output gets written to the report. Also leads to issues where the last saved report on disk may not be the same as what we sent.

lib/puppet/transaction/report.rb
joshcooper: Yeah, I would hope that an exception is raised in the transaction, and we could rescue it/set the report status explicitly.

joshcooper:
We could alternatively do something like this, so that the configurer explicitly sets the failed state on the report, and then finalize_report only sets the status if it's not already set:

```diff
diff --git a/lib/puppet/configurer.rb b/lib/puppet/configurer.rb
index b0f2b63..fece88c 100644
--- a/lib/puppet/configurer.rb
+++ b/lib/puppet/configurer.rb
@@ -344,6 +344,7 @@ class Puppet::Configurer
       apply_catalog(catalog, options)
       report.exit_status
     rescue => detail
+      report.status = 'failed'
       Puppet.log_exception(detail, _("Failed to apply catalog: %{detail}") % { detail: detail })
       return nil
     ensure
diff --git a/lib/puppet/transaction/report.rb b/lib/puppet/transaction/report.rb
index 64143e7..b6e7659 100644
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -107,7 +107,7 @@ class Puppet::Transaction::Report
   # The status of the client run is an enumeration: 'failed', 'changed' or 'unchanged'
   # @return [String] the status of the run - one of the values 'failed', 'changed', or 'unchanged'
   #
-  attr_reader :status
+  attr_accessor :status
 
   # @return [String] The Puppet version in String form.
   # @see Puppet::version()
@@ -204,7 +204,7 @@ class Puppet::Transaction::Report
     change_metric = calculate_change_metric
     add_metric(:changes, {"total" => change_metric})
     add_metric(:events, calculate_event_metrics)
-    @status = compute_status(resource_metrics, change_metric)
+    @status ||= compute_status(resource_metrics, change_metric)
     @noop_pending = @resource_statuses.any? { |name,res| has_noop_events?(res) }
   end
 
```

MosesMendoza:
It feels odd to me to mark the report failed this way - based on if any report entries contain an error log level. It doesn't feel like we're particularly consistent about the level we choose to log at, and I'd guess it's not impossible that in some places we've issued a `Puppet.err('foo')` without expecting that log entry to fail the run or resource. Would it be possible to set the event status for the resource that triggers the cyclical dependency to failed, which would then bubble up to failing the run without changes to the report?

joshcooper:
> Would it be possible to set the event status for the resource that triggers the cyclical dependency to failed

Ideally that makes sense. I think the trouble is that events are only created when we start applying the catalog, and the cycle detection happens before that.

MosesMendoza:
Sure. Just a little brainstorming here - the cycle detection knows about the resources in the cycle, so a pretty easy option might be:

- make `Puppet::Graph::SimpleGraph#report_cycles_in_graph` warn with the cycle message instead return the list of resources in the cycle instead of raising
- define a new `graph_cycle_handler` handler in `Puppet::Transaction#evaluate` that accepts a list of resources in a cycle, generates a resource status for each and sets it to failed, and then fails the transaction
- pass the new handler to `Puppet::Graph::RelationshipGraph#traverse`, which will call it with the result of `report_cycles_in_graph`

I did a quick spike of it, here: https://gist.github.com/MosesMendoza/d36dea8a0aa48a7430ed5e412f4f0f79

joshcooper:
@MosesMendoza that makes sense. I like that the resources causing the cycle are marked as failed. Can you submit a PR and make sure that the overall report status is marked as `failed`?

Also I'm wondering if we should always mark the overall report status as failed in the case that an exception is raised, so that we don't have the same problem again, for slightly different reasons.

MosesMendoza:
@joshcooper yes i can do that

MosesMendoza:
filed https://github.com/puppetlabs/puppet/pull/5950

joshcooper:
Superseded by PR #5950

PR #5916, opened 2017-05-26
(PUP-7155) Expect double quoted strings

joshcooper:
puppet resource now emits double quotes for strings containing
backslashes, which must also be escaped. Update the acceptance tests to
match the new output.

### DIFF
--- a/acceptance/tests/resource/scheduled_task/should_create.rb
+++ b/acceptance/tests/resource/scheduled_task/should_create.rb
@@ -22,10 +22,10 @@ agents.each do |agent|
 
   step "verify task properties"
   on agent, puppet_resource('scheduled_task', name) do
-    assert_match(/command\s*=>\s*'c:\\windows\\system32\\notepad.exe'/, stdout)
+    assert_match(/command\s*=>\s*"c:\\\\windows\\\\system32\\\\notepad.exe"/, stdout)
     assert_match(/arguments\s*=>\s*'foo bar baz'/, stdout)
     assert_match(/enabled\s*=>\s*'true'/, stdout)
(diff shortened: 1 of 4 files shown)


joshcooper:
@thallgren This PR fixes the acceptance failures on Windows, but I don't think we want to emit double quoted strings (which causes us to have to escape backslashes). See my comments in https://tickets.puppetlabs.com/browse/PUP-7155?focusedCommentId=447135&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-447135.

I'm concerned that changing the format will break Windows module pipelines, which likely use `puppet resource` to verify that the provider made the expected changes.

If the double quotes are necessary, could you verify this PR and merge it? Otherwise, close this PR and update `Puppet::Parameter#format_value_for_display` as needed.

/cc @Iristyle

acceptance/tests/resource/scheduled_task/should_create.rb
```diff
@@ -22,10 +22,10 @@
 
   step "verify task properties"
   on agent, puppet_resource('scheduled_task', name) do
-    assert_match(/command\s*=>\s*'c:\\windows\\system32\\notepad.exe'/, stdout)
+    assert_match(/command\s*=>\s*"c:\\\\windows\\\\system32\\\\notepad.exe"/, stdout)
     assert_match(/arguments\s*=>\s*'foo bar baz'/, stdout)
     assert_match(/enabled\s*=>\s*'true'/, stdout)
```
joshcooper: The `enabled` property is stringified, because the `scheduled_task` type does not specify that it is boolean.

PR #5917, opened 2017-05-26
PUP(5912) Add Human Friendly Serialization of Rich Data values

thallgren:
This PR introduces two new classes under the `Puppet::Pops::Serialization` module. One that can convert an arbitrary `Object` into something that is guaranteed to be assignable to `Data` and one that can reverse that conversion. The intended use of these classes is when some value should be presented in a human readable form such as YAML or JSON.

The commit also changes the way a Resource containing rich data is serialized so that it uses the converters instead of the prior hard-coded approach.


thallgren:
Wrong ticket number!

lib/puppet/resource.rb
```diff
@@ -633,6 +561,12 @@ def validate_parameter(name)
     raise Puppet::ParseError.new(_("no parameter named '%{name}'") % { name: name }, file, line) unless valid_parameter?(name)
   end
 
+  # This method, together with #file and #line, makes it possible for a Resource to be a 'source_pos' in a reported issue.
+  # @return [Integer] Instances of this class will always return 0.
+  def pos
+    0
+  end
```
hlindberg: This means it will report a position - no? As opposed to `pos` being `nil` with the meaning that the position is unknown...

hlindberg: IIRC, the reporting looks at file, line and pos, and determines based on what is set how it will format the location (with or without line and pos depending on if they are `nil` or not).

lib/puppet/relationship.rb
```diff
@@ -19,10 +19,10 @@ def self.from_data_hash(data)
 
     args = {}
     if event = data["event"]
-      args[:event] = event
+      args[:event] = :"#{event}"
     end
     if callback = data["callback"]
-      args[:callback] = callback
+      args[:callback] = :"#{callback}"
```
thallgren: Well, `false` and `nil` will never get to this line. See `if` statement above. The `data` is supposed to be a `Hash produced by `to_data_hash` so other objects should really not be here unless someone is doing something that is very out of the ordinary.

hlindberg:
Merging - the 4 failures on Windows are due to bug in AppVeyor.

PR #5920, opened 2017-05-26
(PUP-7599) Fix authorized keys acceptance tests

LuvCurves:
New version of beaker exposed an oversight in the tests that
the user would be root, this is not the case on Cisco Nexus

### DIFF
--- a/acceptance/tests/resource/ssh_authorized_key/create.rb
+++ b/acceptance/tests/resource/ssh_authorized_key/create.rb
@@ -14,11 +14,12 @@ agents.each do |agent|
   #------- SETUP -------#
   step "(setup) backup #{auth_keys} file"
   on(agent, "cp #{auth_keys} /tmp/auth_keys", :acceptable_exit_codes => [0,1])
+  on(agent, "chown $LOGNAME #{auth_keys}")
 
   #------- TESTS -------#
   step "create an authorized key entry with puppet (present)"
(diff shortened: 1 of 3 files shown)


Magisus:
So log name is already getting set somewhere?

LuvCurves:
@Magisus It's set as part of the environment when you login.

johnduarte:
@Magisus yeah, I scratched my head on that for a sec too. `LOGNAME` is a POSIX env var that is available to the session.
@LuvCurves can you add a comment about the nature of `LOGNAME`? I don't know if this will cause problems with Windows when/if we migrate off of Cygwin.

LuvCurves:
@johnduarte This test is not run on windows, or it would have broken too (there is no root on Windows)

LuvCurves:
@johnduarte I'll push a comment to the commit shortly.

LuvCurves:
@johnduarte Added comment to commit about LOGNAME

LuvCurves:
@Magisus tested:
      Test Suite: tests @ 2017-05-26 21:16:32 +0000

       Total Suite Time: 6.24 seconds
      Average Test Time: 1.56 seconds
              Attempted: 4
                 Passed: 3
                 Failed: 0
                Errored: 0
                Skipped: 1
                Pending: 0
                  Total: 4

      - Specific Test Case Status -

PR #5921, opened 2017-05-26
(PUP-7260) Remove inspect command and audit metaparam

Magisus:
* Remove `puppet inspect` command
`puppet inspect` was deprecated in 4.x and is now being removed as a
command line option for Puppet 5. This commit also removes the
associated spec tests.

* Disable the audit metaparameter
The audit metaparameter was deprecated in Puppet 4.10. This commit
disables it, so that while it is still allowed to be present in
manifests, it will do nothing. 
This commit removes the mechanics of auditing, as well as references
throughout the codebase and in the tests. It also updates some details
of the report schema to stop referencing audit.

* Remove `audited` field from the report
Since the audit metaparameter is being disabled, this field is no longer
relevant. This commit removes it from the schema, serialization code,
and tests.

* Remove `historical_value` field from the report
This field was used to track previous values when auditing. Since
auditing is being disabled, this field is unnecessary. This commit
removes the field from the report schema, serialization code, and tests.


lib/puppet/type.rb
```diff
@@ -1289,65 +1289,18 @@ def pathbuilder
...
         if file && line
-          puppet_deprecation_warning(_("The `audit` metaparameter is deprecated and will be ignored in a future release."), { :line => line, :file => file })
-        end
-      end
-      list = Array(list).collect {|p| p.to_sym}
-      unless list == [:all]
-        list.each do |param|
-          next if @resource.class.validattr?(param)
-          fail "Cannot audit #{param}: not a valid attribute for #{resource}"
+          Puppet.warn(_("The `audit` metaparameter is ignored."), { :line => line, :file => file })
```
branan: Should this be a warn_once or perhaps still a deprecation warning?

lib/puppet/type.rb
```diff
@@ -1289,65 +1289,18 @@ def pathbuilder
...
     end
 
     munge do |args|
-      properties_to_audit(args).each do |param|
-        next unless resource.class.validproperty?(param)
-        resource.newattr(param)
-      end
-    end
-
-    def all_properties
```
branan: This might only be used for auditing internally, but it doesn't seem to be marked as private - does it have a use outside of our codebase?

lib/puppet/indirector/resource/ral.rb
```diff
@@ -12,7 +12,7 @@ def allow_remote_requests?
   def find( request )
     # find by name
     res   = type(request).instances.find { |o| o.name == resource_name(request) }
-    res ||= type(request).new(:name => resource_name(request), :audit => type(request).properties.collect { |s| s.name })
+    res ||= type(request).new(:name => resource_name(request))
```
branan: This looks like it originally was designed to audit everything for the requested resource. I suspect that this is intended to allow the `ral` indirection to work for providers that do not implement `self.instances`

I don't know if we have any solid tests for the `ral` indirection (does the `puppet resource` command use it?) It's possible that this may break the indirection for some such types.

Digging through the history of this, it looks like audit replaced a previous mechanism for querying resource properties. We may need to re-implement such a system if resource querying was dependent on audit.

Magisus: @joshcooper helped me walk through this one, including manual testing with the `puppet resource` command using `file`, which goes through this code path. We were not able to make it break. Do you have a concrete example of a case where this could cause trouble? (my grasp on how this stuff all works is still sorta weak)

branan: No concrete examples, just a general feeling of dread :)

If you've already walked through it and verified it all seems to work, I'm happy :+1:

lib/puppet/type.rb
Magisus: Warn once is probably good. I'll update.

lib/puppet/type.rb
Magisus: How might I find out?

branan: Grepping a forge dump is about the only way I know of to look in to this stuff. It's always a pain.

I think this would be easy enough for somebody to re-implement if they really need it - it's probably not worth going to the effort.

Magisus: Works for me.

lib/puppet/type.rb
```diff
@@ -1295,7 +1295,7 @@ def pathbuilder
       if Puppet.settings[:strict] != :off
         # Only warn if `audit` metaparam came from a manifest
         if file && line
-          Puppet.warn(_("The `audit` metaparameter is ignored."), { :line => line, :file => file })
+          Puppet.warn_once(_("The `audit` metaparameter is ignored."), { :line => line, :file => file })
```
joshcooper: `warn_once` is a little tricky, you have to tell it what kind of warning as the first parameter, and `key` as the second, which serves as the message identity (so we don't emit the same warning twice).

I don't think we have a category for this kind of thing (audit is ignored as opposed to deprecated), so perhaps something like:

```ruby
Puppet.warn_once(:audit, "ignore_metaparameter", _("The `audit` metaparameter..."), ...)
```

Iristyle:
I've been running through all the commits individually / running specs / etc.  I assume you know this, but in case not, I got to 700c6cf and that commit causes a large number of fails (that I presume are fixed in subsequent commits) - plan is to squash some of that down, correct?

Magisus:
Yes I plan to put these all into one I think. I removed uses before I removed the definition of the metaparameter itself, to help isolate failures. Once all those were gone I had nothing left but to rip that out, which introduced a certain test failure that I think corrupts the state for all tests run after it. https://github.com/puppetlabs/puppet/pull/5921/commits/d9db71ebcc69b7113b126790439389e3737e0d59 should make things better again.

Iristyle:
It looks like `api/schemas/report.json` should have the value / string `* `audit` - property was in sync, and was being audited.` removed at https://github.com/puppetlabs/puppet/blob/master/api/schemas/report.json#L631

api/schemas/report.json
```diff
@@ -623,7 +623,7 @@
                 },
 
                 "status": {
-                    "description": "One of the following strings:\n\n* `success` - property was out of sync, and was successfully changed to be in sync.\n* `failure`- property was out of sync, and couldn't be changed to be in sync due to an error.\n* `noop` - property was out of sync, and wasn't changed due to noop mode.\n* `audit` - property was in sync, and was being audited.\n\ndepending on the type of the event. Always `audit` in reports of `inspect` kind.\n",
+                    "description": "One of the following strings:\n\n* `success` - property was out of sync, and was successfully changed to be in sync.\n* `failure`- property was out of sync, and couldn't be changed to be in sync due to an error.\n* `noop` - property was out of sync, and wasn't changed due to noop mode.\n\ndepending on the type of the event.\n",
                     "enum": [
                         "success",
                         "failure",
```
Iristyle: I believe there's an `audit` value in the `enum` as well

Magisus: There's also one up in `events_metrics`, but I'm not sure how that one is used. Do we want to get rid of that one too?

Iristyle: If we can no longer generate an `audit` event (based on your code removals), then I would think we should remove it from the schema as a potential value.

Magisus: Pretty sure I removed that in here somewhere...

Magisus: See https://github.com/puppetlabs/puppet/pull/5921/commits/fdfcb56bffe840f3dab74dfca197f0d14cea45a3

Iristyle: Sorry - grepped the wrong directory!

Iristyle:
To answer your question about `audited` and `historical_value` - I'd normally say we should be careful about removing this because it might be breaking.  However, we're on a major version boundary, and report processors are supposed to be OK with a schema version bump as part of a new release, from what I understand.  /cc @joshcooper @hlindberg 

It looks like we've already added / bumped to version 7 as part of https://github.com/puppetlabs/puppet/commit/4e05e16306ec250555da43c33e5155bec0fd9346 so I think that making a few removals in `report.json` can piggy-back on version 7.

I've tested local removal of at least `audited` and can get tests passing:


```diff
diff --git a/api/schemas/report.json b/api/schemas/report.json
index ec7f0bf..0fa16cd 100644
--- a/api/schemas/report.json
+++ b/api/schemas/report.json
@@ -371,7 +371,6 @@
                                 "enum": [
                                     "success",
                                     "failure",
-                                    "audit",
                                     "noop",
                                     "total"
                                 ]
@@ -381,7 +380,6 @@
                                 "enum":  [
                                     "Success",
                                     "Failure",
-                                    "Audit",
                                     "Noop",
                                     "Total"
                                 ]
@@ -564,11 +562,6 @@
             "type":        "object",
             "properties": {
 
-                "audited": {
-                    "description": "True if this property is being audited, otherwise false. True in report of `inspect` kind.",
-                    "type":        "boolean"
-                },
-
                 "property": {
                     "description": "The property for which the event occurred.",
                     "oneOf": [
@@ -623,12 +616,11 @@
                 },
 
                 "status": {
-                    "description": "One of the following strings:\n\n* `success` - property was out of sync, and was successfully changed to be in sync.\n* `failure`- property was out of sync, and couldn't be changed to be in sync due to an error.\n* `noop` - property was out of sync, and wasn't changed due to noop mode.\n* `audit` - property was in sync, and was being audited.\n\ndepending on the type of the event. Always `audit` in reports of `inspect` kind.\n",
+                    "description": "One of the following strings:\n\n* `success` - property was out of sync, and was successfully changed to be in sync.\n* `failure`- property was out of sync, and couldn't be changed to be in sync due to an error.\n* `noop` - property was out of sync, and wasn't changed due to noop mode.\n\ndepending on the type of the event.\n",
                     "enum": [
                         "success",
                         "failure",
-                        "noop",
-                        "audit"
+                        "noop"
                     ]
                 },
                 "time": {
@@ -642,7 +634,6 @@
                 }
             },
             "required": [
-                "audited",
                 "property",
                 "message",
                 "name",
diff --git a/lib/puppet/transaction/event.rb b/lib/puppet/transaction/event.rb
index ffa7364..0853100 100644
--- a/lib/puppet/transaction/event.rb
+++ b/lib/puppet/transaction/event.rb
@@ -11,13 +11,13 @@ class Puppet::Transaction::Event
   include Puppet::Util::Logging
   include Puppet::Network::FormatSupport
 
-  ATTRIBUTES = [:name, :resource, :property, :previous_value, :desired_value, :historical_value, :status, :message, :file, :line, :source_description, :audited, :invalidate_refreshes, :redacted, :corrective_change]
-  YAML_ATTRIBUTES = %w{@audited @property @previous_value @desired_value @historical_value @message @name @status @time @redacted @corrective_change}.map(&:to_sym)
+  ATTRIBUTES = [:name, :resource, :property, :previous_value, :desired_value, :historical_value, :status, :message, :file, :line, :source_description, :invalidate_refreshes, :redacted, :corrective_change]
+  YAML_ATTRIBUTES = %w{@property @previous_value @desired_value @historical_value @message @name @status @time @redacted @corrective_change}.map(&:to_sym)
   attr_accessor *ATTRIBUTES
   attr_accessor :time
   attr_reader :default_log_level
 
-  EVENT_STATUSES = %w{noop success failure audit}
+  EVENT_STATUSES = %w{noop success failure}
 
   def self.from_data_hash(data)
     obj = self.allocate
@@ -26,7 +26,6 @@ class Puppet::Transaction::Event
   end
 
   def initialize(options = {})
-    @audited = false
     @redacted = false
     @corrective_change = false
 
@@ -35,7 +34,6 @@ class Puppet::Transaction::Event
   end
 
   def initialize_from_hash(data)
-    @audited = data['audited']
     @property = data['property']
     @previous_value = data['previous_value']
     @desired_value = data['desired_value']
@@ -51,7 +49,6 @@ class Puppet::Transaction::Event
 
   def to_data_hash
     {
-      'audited' => @audited,
       'property' => @property,
       'previous_value' => @previous_value,
       'desired_value' => @desired_value,
diff --git a/spec/unit/transaction/event_spec.rb b/spec/unit/transaction/event_spec.rb
index 98d6810..c604eae 100644
--- a/spec/unit/transaction/event_spec.rb
+++ b/spec/unit/transaction/event_spec.rb
@@ -41,7 +41,7 @@ describe Puppet::Transaction::Event do
     expect(event.status).to eq("success")
   end
 
-  it "should fail if the status is not to 'audit', 'noop', 'success', or 'failure" do
+  it "should fail if the status is not to 'noop', 'success', or 'failure" do
     event = Puppet::Transaction::Event.new
     expect { event.status = "foo" }.to raise_error(ArgumentError)
   end
@@ -54,12 +54,6 @@ describe Puppet::Transaction::Event do
     expect(Puppet::Transaction::Event.new.time).to be_instance_of(Time)
   end
 
-  describe "audit property" do
-    it "should default to false" do
-      expect(Puppet::Transaction::Event.new.audited).to eq(false)
-    end
-  end
-
   describe "when sending logs" do
     before do
       Puppet::Util::Log.stubs(:new)
@@ -164,7 +158,6 @@ describe Puppet::Transaction::Event do
 
       tripped = Puppet::Transaction::Event.from_data_hash(PSON.parse(event.to_pson))
 
-      expect(tripped.audited).to eq(event.audited)
       expect(tripped.property).to eq(event.property)
       expect(tripped.previous_value).to eq(event.previous_value)
       expect(tripped.desired_value).to eq(event.desired_value)
@@ -178,7 +171,6 @@ describe Puppet::Transaction::Event do
   it "should round trip an event for an inspect report through pson" do
       resource = Puppet::Type.type(:file).new(:title => make_absolute("/tmp/foo"))
       event = Puppet::Transaction::Event.new(
-        :audited => true,
         :source_description => "/my/param",
         :resource => resource,
         :file => "/foo.rb",
@@ -195,7 +187,6 @@ describe Puppet::Transaction::Event do
       expect(tripped.historical_value).to be_nil
       expect(tripped.name).to be_nil
 
-      expect(tripped.audited).to eq(event.audited)
       expect(tripped.property).to eq(event.property)
       expect(tripped.previous_value).to eq(event.previous_value)
       expect(tripped.message).to eq(event.message)
diff --git a/spec/unit/transaction/report_spec.rb b/spec/unit/transaction/report_spec.rb
index d4f2325..fd6cd69 100644
--- a/spec/unit/transaction/report_spec.rb
+++ b/spec/unit/transaction/report_spec.rb
@@ -606,7 +606,6 @@ describe Puppet::Transaction::Report do
 
   def generate_report
     event_hash = {
-      :audited => false,
       :property => 'message',
       :previous_value => 'absent',
       :desired_value => 'a resource',
```

Iristyle:
Also trashed historical_value in a local branch - pushed up to https://github.com/Iristyle/puppet/commits/pr/5921 - all tests pass locally

Magisus:
I've squashed the commits into meaningful chunks. I'd like to run a build of this through the puppet acceptance tests before it gets merged.

api/schemas/report.json
```diff
@@ -588,17 +580,7 @@
...
-                    "oneOf": [
-                        { "type": "string" },
-                        { "type": "array" },
-                        { "type": "object" },
-                        { "type": "null" }
-                    ]
-                },
-
-                "historical_value": {
-                    "description": "The audited value from a previous run of Puppet, if known. Absent in reports of `inspect` kind.",
```
joshcooper: I verified the puppetdb report processor doesn't make use of `historical_value`: https://github.com/puppetlabs/puppetdb/blob/master/puppet/lib/puppet/reports/puppetdb.rb

Magisus:
So it looks like there are a few acceptance failures: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/view/vmpooler/job/platform_puppet-agent_intn-van-sys_suite-daily-puppet-ad-hoc/1/ Going to dig through these today, any insight is appreciated!

lib/puppet/indirector/resource/ral.rb
Magisus: Jk, this is broken.

`puppet resource file /tmp/foo.txt` used to return all properties, now it only returns `ensure`.

Magisus:
The failure is caused by https://github.com/puppetlabs/puppet/pull/5921/files#diff-677fdc6172c0d09148bf5af07e8e9d79R15. We will need to figure out how to replace the functionality accomplished by audit here.

Magisus:
The above commit fixes the pluginsync test issue. I'll squash it once it's gotten reviewed and I verify whether or not we need something else to fix the service test.

Magisus:
With the new package, the service test now passes as well. It was most likely a fluke from the last package build, which would explain it not appearing under `ci:test:git`.

api/schemas/report.json
```diff
@@ -136,9 +136,8 @@
         },
 
         "kind": {
```
branan: We're updating the report schema version number for Puppet 5 anyway - does it make sense to just entirely rip out this enumeration?

lib/puppet/indirector/resource/ral.rb
```diff
@@ -12,7 +12,13 @@ def allow_remote_requests?
...
     # find by name
     res   = type(request).instances.find { |o| o.name == resource_name(request) }
-    res ||= type(request).new(:name => resource_name(request), :audit => type(request).properties.collect { |s| s.name })
+    if !res
+      res = type(request).new(:name => resource_name(request))
+      # Register all of the properties for data collection
+      type(request).properties.collect do |s|
+        res.newattr(s.name.to_sym)
+      end
+    end
```
branan: I was worried that this would be much more complicated. Glad it was easy

api/schemas/report.json
Magisus: I was wondering about that. Happy to do it if you think it won't break too much.

Magisus: Updated.

Magisus:
Now with updated specs.

api/schemas/report.json
```diff
@@ -135,14 +134,6 @@
             "type":        "string"
         },
 
-        "kind": {
```
Magisus: @mullr do you know if this is used in PuppetDB anywhere?

mullr: I don't think it is.

mullr: That is: I looked in the source code and in the protocol documentation, and I can't find it in either of those places.

Magisus: Awesome. Well, count this as a heads up if something ends up breaking... it doesn't make sense to have anymore, given we've reduced it to one option.

MosesMendoza:
do we need to bump the report format? Or was that done since we shipped last...

Magisus:
It was done already for Puppet 5.

MosesMendoza:
@Magisus would you mind updating the report example in `http_report.md` for the new report w/out kind?

Magisus:
Nice catch, updated.

MosesMendoza:
Yes that would be good I think? Also it looks like there's one last straggler use of `:audit` in specs, here: https://github.com/Magisus/puppet/blob/c15487c14a499db14be82ab2465f154ef189f6b9/spec/unit/util/storage_spec.rb#L36. Desired?

Magisus:
Updated both.

spec/unit/type_spec.rb
```diff
@@ -1154,151 +1150,107 @@ def destroy; end
...
-      Puppet::Type.type(:file).stubs(:key_attributes).returns [:path, :mode, :owner]
-      Puppet::Type.type(:file).stubs(:title_patterns).returns(
-        [ [ /(.*)/, [ [:path, lambda{|x| x} ] ] ] ]
-      )
-      myfile = make_absolute('/my/file')
-      res = Puppet::Type.type(:file).new( :title => myfile, :path => myfile, :owner => 'root', :content => 'hello' )
-      expect(res.uniqueness_key).to eq([ nil, 'root', myfile])
-    end
-  end
+  it "should handle proprieties correctly" do
```
MosesMendoza: nit (spelling)

Iristyle:
Love the diffstat! `+202 −1,139 `

PR #5923, opened 2017-05-26
(PUP-7487) Default facts and reports to JSON

joshcooper:
Submit facts and reports as json by default. Fallback to pson if `preferred_serialization_format=pson`.

### DIFF
--- a/lib/puppet/configurer/fact_handler.rb
+++ b/lib/puppet/configurer/fact_handler.rb
@@ -30,8 +30,10 @@ module Puppet::Configurer::FactHandler
   def facts_for_uploading
     facts = find_facts
 
-    text = facts.render(:pson)
-
-    {:facts_format => :pson, :facts => CGI.escape(text)}
+    if Puppet[:preferred_serialization_format] == "pson"
+      {:facts_format => :pson, :facts => CGI.escape(facts.render(:pson)) }
+    else
+      {:facts_format => 'application/json', :facts => facts.render(:json) }
+    end
   end
 end
--- a/lib/puppet/transaction/report.rb
+++ b/lib/puppet/transaction/report.rb
@@ -398,10 +398,6 @@ class Puppet::Transaction::Report
     status
   end
 
-  def self.default_format
-    :pson
-  end
-
   private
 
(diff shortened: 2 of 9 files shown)


MosesMendoza:
confirmed report default serialization format is `:json` with this change:
```
[10] pry(main)> Puppet::Transaction::Report.default_format
=> :json
```

And that facts are submitted as json by default:
```
[1] pry(#<Puppet::Resource::Catalog::Compiler>)> request.options[:facts_format]
request.options[:facts_format]
=> "application/json"
2] pry(#<Puppet::Resource::Catalog::Compiler>)> request.options[:facts]
request.options[:facts]
=> "{\"name\":\"ubuntu16.localdomain\",\"values\":{\"puppetversion\":\"5.0.0\"...
```
As of https://github.com/puppetlabs/puppet/commit/72589ad16cc95b6063823fd597e3069b47b04349 puppet will convert the JSON facts string to a facts instance.

MosesMendoza:
I'm 👍  for merge - but before that do we need to update any puppet server tests to make sure the server side loads an agent that accepts json facts @joshcooper @camlow325 ?

joshcooper:
Good question, the pin may need to be bumped. I verified that the agent works as expected when the server is using puppet-agent latest which passed all of our acceptance tests:

```
[root@jz47aepajldj6pq ~]# rpm -q puppet-agent
puppet-agent-4.99.0.199.g649992c-1.el7.x86_64
[root@jz47aepajldj6pq ~]# rpm -q puppetserver
puppetserver-5.0.0.master-0.1SNAPSHOT.2017.06.05T0956.el7.noarch
```

camlow325:
FWIW, I ran the Puppet Server unit/integration Clojure tests with the Puppet submodule updated with the changes from this PR and those passed.  It is possible that we may need to make other updates in the Puppet Server CI pipeline when we bump our Puppet submodule / puppet-agent pins for this later but it seems unlikely given that @joshcooper indicated that the full Puppet acceptance tests are still passing with the change.  I'd be fine with merging this and updating Puppet Server acceptance tests if any further changes are needed.

PR #5924, opened 2017-05-29
(PUP-7603) Make StringFormatter use single quotes more diligently

thallgren:
Prior to this commit, the StringFormatter would use double quotes for
all strings that contained control-, non-ascii-, dollar, or backslash
characters. Double quoting is however only needed for strings that
contain control- or non-ascii characters.

This commit changes how the detection is made so that single quotes are
used whenever possible.

### DIFF
--- a/lib/puppet/pops/types/string_converter.rb
+++ b/lib/puppet/pops/types/string_converter.rb
@@ -835,35 +835,60 @@ class StringConverter
   #
   # @api public
   def puppet_quote(str)
-    if str.ascii_only? && (str =~ /(?:'|\$|\p{Cntrl}|\\)/).nil?
-      "'#{str}'"
-    else
-      bld = '"'
-      str.codepoints do |codepoint|
-        case codepoint
-        when 0x09
-          bld << '\\t'
-        when 0x0a
-          bld << '\\n'
-        when 0x0d
-          bld << '\\r'
-        when 0x22
-          bld << '\\"'
-        when 0x24
-          bld << '\\$'
-        when 0x5c
-          bld << '\\\\'
+    # Assume that the string can be single quoted
+    bld = '\''
+    escaped = false
+    str.each_codepoint do |codepoint|
+      # Control characters and non-ascii characters cannot be present in a single quoted string
+      return puppet_double_quote(str) if codepoint > 0x7f || codepoint < 0x20
+
+      if escaped
+        bld << 0x5c << codepoint
+        escaped = false
+      else
+        if codepoint == 0x27
+          bld << 0x5c << codepoint
+        elsif codepoint == 0x5c
+          escaped = true
         else
-          if codepoint < 0x20 || codepoint > 0x7f
-            bld << sprintf('\\u{%X}', codepoint)
-          else
-            bld.concat(codepoint)
-          end
+          bld << codepoint
         end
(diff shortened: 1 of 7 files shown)


lib/puppet/pops/types/string_converter.rb
```diff
@@ -835,35 +835,60 @@ def string_PStringType(val_type, val, format_map, _)
...
-          bld << '\\\\'
+    # Assume that the string can be single quoted
+    bld = '\''
+    escaped = false
+    str.each_codepoint do |codepoint|
+      # Control characters and non-ascii characters cannot be present in a single quoted string
+      return puppet_double_quote(str) if codepoint > 0x7f || codepoint < 0x20
+
+      if escaped
+        bld << 0x5c << codepoint
```
hlindberg: this looks like it is unreachable because `escaped` must be `false` here

spec/unit/pops/types/string_converter_spec.rb
```diff
@@ -134,17 +134,29 @@
         expect(converter.convert("esc \u{1b}.", string_formats)).to eq('"esc \\u{1B}."')
       end
 
-      it 'escape for $' do
-        expect(converter.convert('escape the $ sign', string_formats)).to eq('"escape the \$ sign"')
+      it 'escape for $ in double quoted string' do
+        expect(converter.convert("escape the $ sign\n", string_formats)).to eq('"escape the \$ sign\n"')
```
hlindberg: At first I did not get that \n forces the string representation to be double quoted . Worth a comment.

joshcooper:
Could you update the scheduled_task and user managing home acceptance tests to expect single or double quotes?

lib/puppet/pops/types/string_converter.rb
thallgren: On the contrary. This traps the case when the above loop ends with a 0x5c and hence, leave `escape == true`.

thallgren:
@joshcooper I tested all acceptance tests under tests/resource/user and tests/resource/scheduled_task. They pass without modification so I guess they all are resilient against quote changes now. Perhaps you have other tests in mind?

joshcooper:
> They pass without modification so I guess they all are resilient against quote changes now.

I had to modify the tests to accept both single or double quotes (see https://github.com/puppetlabs/puppet/commit/fabc26d59c15b182d6c1683d3cfbb374e2435e8f
https://github.com/puppetlabs/puppet/commit/64ea8da09352a70df2de54a96f11617968999872) , and escaped backslashes, but now that the output is consistently single quotes with single backslashes, I'm thinking the tests should be reverted to how they were before (requiring single quotes with single backslashes)?

thallgren:
Ah, yes, that makes sense. I'll revert them to single quotes.

lib/puppet/pops/types/string_converter.rb
hlindberg: duh... there is and `each` there

thallgren: Yes, and I though you complained about the one after the each. Hence my weird reply. In any case, there's no unreachable code.

joshcooper:
I think the appveyor issues are https://github.com/appveyor/ci/issues/1579

PR #5925, opened 2017-05-30
(PUP-7382) Ensure that Transaction::Event is serializable

thallgren:
The `Transaction::Event` class contains _previous_, _desired_, and
_historical_ values that stems from the _current_ value of a provider.
These values in turn, may not be serializable using JSON or YAML since
they might not conform to the `Data` type.

This commit ensures that the hash returned by the `Event#to_data_hash`
method really is `Data` compliant by passing it through the
`ToDataConverter#convert` method. This means that the range of JSON/YAML
serializable values is extended from just being `Data` to instead be
`RichData`. It also means that instances of classes, unknown to the
puppet type system, will be serialized as strings, formatted using the
`StringConverter` which in turn will emit the same string as will be
produced by the `Parameter#format_value_for_display` method.


thallgren:
Please note that this PR is on top of PUP-5912. It should not be merged until that [PR-5917](/puppetlabs/puppet/pull/5917) is merged.

hlindberg:
I suspect that when :pson is used that the text on the wire is identical to current version (just as horrible). Need to confirm with @joshcooper if that is true.

Iristyle:
The AppVeyor issue is https://github.com/appveyor/ci/issues/1579

lib/puppet/pops/serialization/to_data_converter.rb
```diff
@@ -0,0 +1,284 @@
...
+
+    def non_string_keyed_hash_to_data(hash)
+      if @rich_data
+        to_key_extended_hash(hash)
+      else
+        result = {}
+        hash.each_pair do |key, value|
+          if key.is_a?(Symbol) && @symbol_as_string
+            key = key.to_s
+          elsif !key.is_a(String)
```
joshcooper: Should be `is_a?` and is missing tests.

thallgren: I created [PR-5951](/puppetlabs/puppet/pull/5951) with a fix and tests since this PR is merged and closed

PR #5926, opened 2017-05-30
(PUP-7583) Remove to_yaml_properties

thallgren:
This commit removes the `#to_yaml_properties` method everywhere in the
Puppet code-base and adds an include of `Puppet::Util::PsychSupport`
where needed.

Some tests that only asserted the returned value from the removed
method was also removed. Other tests were rewritten so that they use
the keys of the hash returned from the `#to_data_hash` methods instead.

### DIFF
--- a/lib/puppet/graph/simple_graph.rb
+++ b/lib/puppet/graph/simple_graph.rb
@@ -6,6 +6,7 @@ require 'set'
 class Puppet::Graph::SimpleGraph
   include Puppet::Util::PsychSupport
 
+  #
   # All public methods of this class must maintain (assume ^ ensure) the following invariants, where "=~=" means
   # equiv. up to order:
   #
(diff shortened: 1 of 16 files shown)


lib/puppet/graph/simple_graph.rb
```diff
@@ -4,6 +4,8 @@
 
 # A hopefully-faster graph class to replace the use of GRATR.
 class Puppet::Graph::SimpleGraph
+  include Puppet::Util::PsychSupport
```
joshcooper: Can you break this out into a separate PR for PUP-7381, since that one is about using `encode_with` consistently, and this PR is about removing duplication?

thallgren: Done. Please note that this PR is on top of the PR for PUP-7381.

thallgren: I verified that the puppetdb unit tests are green.

PR #5919, opened 2017-05-26
(PUP-7597) Add Hiera5 default yaml file

jcoconnor:
This is installed by PA instead of the previous hiera3 format file.

PR #5927, opened 2017-05-30
(PUP-7597) Add Hiera5 default yaml file

jcoconnor:
This is installed by PA instead of the previous hiera3 format file.

This is #5919 retargeted at ```stable```


McdonaldSeanp:
👍

Iristyle:
@hlindberg / @thallgren this make sense to you?

hlindberg:
I don't get this - why use a global hiera 5 hiera.yaml that uses a hieradata directory inside an environment? You would want a hiera.yaml inside that environment and a global that does not have any data at all from any of the environments.

This configuration in this PR does not make sense. It is just a straight translation from the old hiera 3 "single global layer" to  using hiera 5 format in the hiera.yaml.

jcoconnor:
@hlindberg & @branan This is the ```puppet``` side changes with the two files we discussed.
```puppet-agent``` repo changes to follow in the other PR.

File: conf/hiera.yaml [hunk 1]
```diff
@@ -0,0 +1,2 @@
+---
+# Hiera5 Global configuration file.
```
hlindberg: Think this should contain:
```
# Hiera 5 Global configuration file

version: 5

# defaults:
#   data_hash: yaml_data
# hierarchy:
#   - name: "Global Overrides"
#     path: "overrides.yaml"
#
```

The config is empty in the sense that it does not define any search. At the same time it communicates the intention that global overrides could be added in a file - suggested "overrides.yaml".

For second opinion: ping @DavidS , what would you like the default to be?

ext/hiera/hiera.yaml
```diff
@@ -0,0 +1,14 @@
+---
+version: 5
+defaults:
+  # datadir is empty here, so hiera uses its defaults:
+  # - /etc/puppetlabs/code/environments/%{environment}/hieradata on *nix
+  # - %CommonAppData%\PuppetLabs\code\environments\%{environment}\hieradata on Windows
+  # When specifying a datadir, make sure the directory exists.  datadir:
```
hlindberg: The information about `datadir` is not correct. By default it is a directory named ´data´ in the root directory of the environment (i.e. next to the `hiera.yaml` file).
The text that says `/etc/puppetlabs/code/environments/%{environment}/hieradata` is not correct since `/etc/puppetlabs/code/environments` depends on a setting. And ´hieradata´ is the version 3 standard. The Windows equiv has the same problem.

conf/hiera.yaml
jcoconnor: Have replace contents as suggested.

ext/hiera/hiera.yaml
jcoconnor: Have amended ```datadir``` to ```data```

Can you point me to info on the setting for ```/etc/puppetlabs/code/environments```

conf/hiera.yaml
DavidS: no idea. I dug a bit through the ticket backstory, and I have to admit that I still do not have enough context to understand how this **should** work.

ext/hiera/hiera.yaml
jcoconnor: @hlindberg Updated file as requested.

ext/hiera/hiera.yaml
```diff
@@ -0,0 +1,15 @@
+---
+version: 5
+defaults:
+  # The default value for "datadir" is "data" under the same directory as the hiera.yaml
+  # file (this file)
+  # When specifying a datadir, make sure the directory exists.
+  # See https://docs.puppet.com/puppet/4.10/environments.html for further details on environments.
+  datadir:
```
hlindberg: This is wrong. Did you perhaps intend it to be like this?
```
# datadir: data
```
Now it is a yaml syntax error (key without a value)

jcoconnor: Fixed.

jcoconnor:
Ping @thallgren Can you take a look at this PR please ?
Thanks - John

conf/hiera.yaml
```diff
@@ -0,0 +1,11 @@
+---
+# Hiera 5 Global configuration file
+
+version: 5
+
+# defaults:
+#   data_hash: yaml_data
+# hierarchy:
+#   - name: "Global Overrides"
+#     path: "overrides.yaml"
```
thallgren: This comment might be interpreted as a description of the default for the defaults. If that's intended, then it's not correct. The default hierarchy of a Version 5 hiera.yaml is always:
```
hierarchy:
  - name: Common
    data_hash: yaml_data
```

ext/hiera/hiera.yaml
```diff
@@ -0,0 +1,15 @@
+---
+version: 5
+defaults:
+  # The default value for "datadir" is "data" under the same directory as the hiera.yaml
+  # file (this file)
+  # When specifying a datadir, make sure the directory exists.
+  # See https://docs.puppet.com/puppet/4.10/environments.html for further details on environments.
+  # datadir: data
+  data_hash: yaml_data
```
thallgren: This is already the default so it's inconsistent to comment out `datadir: data` and leave this line uncommented.

jcoconnor: Done

conf/hiera.yaml
jcoconnor: Done

jcoconnor:
@thallgren I've pushed up changes to address you comments.

geoffnichols:
@jcoconnor I think we may want to retarget this against puppet#master, since https://github.com/puppetlabs/puppet-agent/pull/1076 is targeted against puppet-agent#master.

PR #5928, opened 2017-05-30
(PUP-7381) Ensure that SimpleGraph includes Puppet::Util::PsychSupport

thallgren:
This commit adds the `#encode_with` support provided by the
`Puppet::Util::PsychSupport` module to the `Puppet::Graph::SimpleGraph`
class and thereby also to its subclass  `Puppet::Resource::Catalog`.

This PR has been tested against puppetdb at 4a5092d (current HEAD on master)

### DIFF
--- a/lib/puppet/graph/simple_graph.rb
+++ b/lib/puppet/graph/simple_graph.rb
@@ -4,7 +4,8 @@ require 'set'
 
 # A hopefully-faster graph class to replace the use of GRATR.
 class Puppet::Graph::SimpleGraph
-  #
+  include Puppet::Util::PsychSupport
+
   # All public methods of this class must maintain (assume ^ ensure) the following invariants, where "=~=" means
   # equiv. up to order:
   #
@@ -295,6 +296,7 @@ class Puppet::Graph::SimpleGraph
   # since they have to specify what kind of edge it is.
   def add_edge(e,*a)
     return add_relationship(e,*a) unless a.empty?
+    e = Puppet::Relationship.from_data_hash(e) if e.is_a?(Hash)
     @upstream_from.clear
     @downstream_from.clear
     add_vertex(e.source)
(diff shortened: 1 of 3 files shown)


lib/puppet/graph/simple_graph.rb
```diff
@@ -481,70 +483,49 @@ class << self
...
-    attr_reader :vertex, :adjacencies
-    def initialize(vertex, adjacencies)
-      @vertex = vertex
-      @adjacencies = adjacencies
-    end
-
-    def inspect
-      { :@adjacencies => @adjacencies, :@vertex => @vertex.to_s }.inspect
+  def initialize_from_hash(hash)
+    initialize
```
joshcooper: Is this `initialize` call needed?

lib/puppet/graph/simple_graph.rb
```diff
@@ -481,70 +483,49 @@ class << self
...
         end
-        result
+        vname = vertex.to_s
+        result[vname] = { 'adjacencies' => adjacencies, 'vertex' => vname }
       end
-    else
-      super(v)
+      result
     end
+    Puppet::Pops::Types::TypeAsserter.assert_instance_of('SimpleGraph', Puppet::Pops::Types::TypeFactory.data, hash)
```
joshcooper: Should this be here, or were you testing?

spec/unit/graph/simple_graph_spec.rb
```diff
@@ -528,31 +528,31 @@ def empty_graph(graph)
     end
 
     def one_vertex_graph(graph)
-      graph.add_vertex(:a)
```
joshcooper: Are there any usages of SimpleGraph where we need to preserve symbols? I don't think the Catalog relies on that as the vertices are either `Puppet::Resource` objects in the resource (server) catalog, or `Puppet::Type::<name>` instances in the RAL (agent) catalog.

spec/unit/graph/simple_graph_spec.rb
```diff
@@ -605,7 +609,7 @@ def graph_to_yaml(graph, which_format)
             edge.keys.each { |x| expect(['source', 'target', 'callback', 'event']).to include(x) }
             %w{source target callback event}.collect { |x| edge[x] }
           end
-          expect(Set.new(actual_edge_tuples)).to eq(Set.new(expected_edge_tuples))
+          expect(Set.new(actual_edge_tuples)).to eq(Set.new(expected_edge_tuples.map { |tuple| tuple.map {|e| e.nil? ? nil : e.to_s }}))
```
joshcooper: Why do we need to special case the case where `e` is nil now, but we didn't need to previously? Does `Catalog.from_data_hash` now include `nil` edges (that would be strange)?

lib/puppet/graph/simple_graph.rb
thallgren: I tried without it. Doesn't work. I suspect that YAML just allocates, then calls initialize_from_hash.

lib/puppet/graph/simple_graph.rb
thallgren: It shouldn't be there. I'll remove it.

spec/unit/graph/simple_graph_spec.rb
thallgren: I didn't find any cases where Symbol preservation is needed. If it is, it should be done using rich_data semantics instead of relying on YAML. What if we serialize with JSON instead for instance?

spec/unit/graph/simple_graph_spec.rb
thallgren: The values are nil, both in the expected tuple and in the hash. The check for nil is there to prevent that a nil becomes an empty string when converting symbols to strings. The test is silly in a way, in that it compares the result of a to_data_hash with the real object. I felt it was easier to just add this conversion than to rewrite the whole test.

spec/unit/graph/simple_graph_spec.rb
```diff
@@ -615,24 +619,25 @@ def graph_to_yaml(graph, which_format)
...
                   # Since we already checked edges, just check consistency
                   # with edges.
                   desired_source = direction == :in ? adj_key : key
                   desired_target = direction == :in ? key : adj_key
                   expected_edges = edges.select do |edge|
                     edge['source'] == desired_source && edge['target'] == desired_target
                   end
-                  expect(adj_value).to be_a(Set)
-                  if object_ids(adj_value) != object_ids(expected_edges)
+                  expect(adj_value).to be_a(Array)
```
joshcooper: Why is this now an `Array`?

thallgren: Because `Set` is not `Data` and cannot be represented in YAML/JSON without special tags. The `Array` is unique though.

spec/unit/graph/simple_graph_spec.rb
```diff
@@ -681,16 +686,20 @@ def graph_to_yaml(graph, which_format)
       end
 
       it "should be able to serialize a graph where the vertices contain backreferences to the graph (#{which_format} format)" do
+        pending('Make SimpleGraph a known to the Puppet Type System')
```
joshcooper: Could we fix this test so it's not skipped (I assume by setting `rich_data => false`)? And add a new pending test for the rich data case?

thallgren: No, this test cannot run unless SimpleGraph as a PuppetObject. That's not related to `rich_data` but to the fact that the SimpleGraph object is not `Data`. I tried relying on special YAML tags here, but that fails when Ruby < 2.3 (hence my question earlier in HipChat). So if we really want this functionality, then we need to reintroduce `#to_yaml_properties` or defer the problem until later (if it is a problem).

joshcooper: Why does SimpleGraph need to be known to the type system? For example, the type system doesn't know about Reports, but `Report#to_data_hash` works as expected.

thallgren: None of those have tests where they are made to contain an object that in turn points back to themselves with the requirement that this self-reference is resolved during deserialization. In order to do that with Ruby < 2.3, we will either need to go back to the `#to_yaml_properties` approach, or use an alternative. The `#encode_with` seems to work well with self-references as long as they don't reference the top object. When that happens Rubies < 2.3 fails.

That leaves us with three alternatives:
1. Keep ` #to_yaml_properties` for the sole sake of being able to do this.
2. Use an alternative approach. Registering `SimpleGraph` in the Puppet Type system and use rich_data solves this.
3. Conclude that this self-reference capability is not needed and remove the test. That would be my take.

The Catalog rewrites the `#to_data_hash` completely and does not use the functionality in `SimpleGraph` at all. So who benefits from this self-referencing capability anyway?

hlindberg: The "backreference-test" was added in 2010 when optimization of SimpleGraph was performed. Does not seem related to a real requirement.

hlindberg:
Merging as I think @joshcooper's concerns have been addressed.

PR #5929, opened 2017-05-31
(PUP-7605) Ensure that resource title is a String (get rid of :main)

thallgren:
This commit ensures that a resource title is a String. A Symbol (like
:main) is still accepted but will be converted into a String.

### DIFF
--- a/lib/puppet/util/tagging.rb
+++ b/lib/puppet/util/tagging.rb
@@ -40,7 +40,7 @@ module Puppet::Util::Tagging
   # since that results in testing the same string twice
   #
   def tag_if_valid(name)
-    if name.is_a?(String) && name =~ ValidTagRegex && Puppet::Resource::MAIN != name
+    if name.is_a?(String) && name =~ ValidTagRegex
       name = name.downcase
       @tags ||= new_tags
       if @tags.add?(name) && name.include?('::')
--- a/lib/puppet/util/tagging.rb
+++ b/lib/puppet/util/tagging.rb
@@ -40,7 +40,7 @@ module Puppet::Util::Tagging
   # since that results in testing the same string twice
   #
   def tag_if_valid(name)
-    if name.is_a?(String) && name =~ ValidTagRegex
+    if name.is_a?(String) && name =~ ValidTagRegex && Puppet::Resource::MAIN != name
       name = name.downcase
       @tags ||= new_tags
       if @tags.add?(name) && name.include?('::')
--- a/spec/unit/transaction/resource_harness_spec.rb
+++ b/spec/unit/transaction/resource_harness_spec.rb
@@ -1,6 +1,7 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
 
+require 'puppet/transaction/resource_harness'
 require 'puppet/type/component'
 
 describe Puppet::Transaction::ResourceHarness do
--- a/spec/unit/transaction/resource_harness_spec.rb
+++ b/spec/unit/transaction/resource_harness_spec.rb
@@ -1,7 +1,7 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
 
-require 'puppet/transaction/resource_harness'
+require 'puppet/type/component'
 
 describe Puppet::Transaction::ResourceHarness do
   include PuppetSpec::Files
--- a/spec/unit/type/component_spec.rb
+++ b/spec/unit/type/component_spec.rb
@@ -42,10 +42,6 @@ describe component do
       expect(component.new(:name => "Class[foo]").pathbuilder).to eq(["Foo"])
     end
 
-    it "should produce the class name even for the class named main" do
-      expect(component.new(:name => "Class[main]").pathbuilder).to eq(['main'])
-    end
-
     it "should produce a resource reference if the component does not model a class" do
       expect(component.new(:name => "Foo[bar]").pathbuilder).to eq(["Foo[bar]"])
     end
--- a/spec/unit/type/component_spec.rb
+++ b/spec/unit/type/component_spec.rb
@@ -43,7 +43,7 @@ describe component do
     end
 
     it "should produce the class name even for the class named main" do
-      expect(component.new(:name => "Class[main]").pathbuilder).to eq(["Main"])
+      expect(component.new(:name => "Class[main]").pathbuilder).to eq(['main'])
     end
 
     it "should produce a resource reference if the component does not model a class" do
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -577,7 +577,7 @@ class Puppet::Parser::Compiler
     krt = environment.known_resource_types
     @main = krt.find_hostclass('') || krt.add(Puppet::Resource::Type.new(:hostclass, ''))
     @topscope.source = @main
-    @main_resource = Puppet::Parser::Resource.new('class', Puppet::Resource::MAIN, :scope => @topscope, :source => @main)
+    @main_resource = Puppet::Parser::Resource.new(Puppet::Resource::TYPE_CLASS, '', :scope => @topscope, :source => @main)
     @topscope.resource = @main_resource
 
     add_resource(@topscope, @main_resource)
--- a/lib/puppet/parser/compiler.rb
+++ b/lib/puppet/parser/compiler.rb
@@ -577,7 +577,7 @@ class Puppet::Parser::Compiler
     krt = environment.known_resource_types
     @main = krt.find_hostclass('') || krt.add(Puppet::Resource::Type.new(:hostclass, ''))
     @topscope.source = @main
-    @main_resource = Puppet::Parser::Resource.new('class', :main, :scope => @topscope, :source => @main)
+    @main_resource = Puppet::Parser::Resource.new('class', Puppet::Resource::MAIN, :scope => @topscope, :source => @main)
     @topscope.resource = @main_resource
 
     add_resource(@topscope, @main_resource)
@@ -637,7 +637,7 @@ class Puppet::Parser::Compiler
   end
 
   def add_resource_metaparams
-    unless main = catalog.resource(:class, :main)
+    unless main = catalog.resource(:class, Puppet::Resource::MAIN)
       #TRANSLATORS "main" is a function name and should not be translated
       raise _("Couldn't find main")
     end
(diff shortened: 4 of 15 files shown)


joshcooper:
Copying what I mentioned in hipchat, I'm not in favor of making this change now, because we have to account for things like https://github.com/puppetlabs/puppet/pull/5929/files#diff-ac9d5450f67274b038705194387c3907R43. Previously, `:main` was never a valid tag, because it was a Symbol, but if it's a String, then we have to add exclusions. And we're likely only going to identify those places if we have a test, the test is accurate, and it wasn't relying on the Symbol monkey patch we recently removed.

I'd rather fix the puppetdb terminus, since it's using the result of `to_data_hash` to call back into the catalog, in effect, mixing data from different "domains".

hlindberg:
So, @joshcooper you are basically saying that the fix here: https://github.com/puppetlabs/puppetdb/pull/2304 is the only thing needed?
I do agree that changing the special use of :main (which is special in more ways than one) into being a regular name 'main' is far from ideal. The real solution is probably to make the main class something different altogether.

spec/unit/type/component_spec.rb
```diff
@@ -43,7 +43,7 @@
     end
 
     it "should produce the class name even for the class named main" do
-      expect(component.new(:name => "Class[main]").pathbuilder).to eq(["Main"])
+      expect(component.new(:name => "Class[main]").pathbuilder).to eq(['main'])
```
hlindberg: Have no clue if this change in case makes a difference.

lib/puppet/util/tagging.rb
```diff
@@ -40,7 +40,7 @@ def tag(*ary)
   # since that results in testing the same string twice
   #
   def tag_if_valid(name)
-    if name.is_a?(String) && name =~ ValidTagRegex
+    if name.is_a?(String) && name =~ ValidTagRegex && Puppet::Resource::MAIN != name
```
hlindberg: This is the special case @joshcooper pointed out - when name is a String and 'main', it is special case to consider - we don't know all the places that may matter.

spec/unit/type/component_spec.rb
thallgren: The really funny thing about this test is that when creating the class this way (prior to this PR), the title becomes the string 'Main'. The same resource created with the symbol, does not capitalize the title. This is what happens:
```
2.3.1 :002 > x = Puppet::Resource.new('class', :main)
 => Class[main]{:name=>"main"} 
2.3.1 :003 > x.ref
 => "Class[main]" 
2.3.1 :004 > x = Puppet::Resource.new('class', 'main')
 => Class[Main]{:name=>"Main"} 
2.3.1 :005 > x.ref
 => "Class[Main]" 
```
IMHO, this doesn't make much sense.

lib/puppet/util/tagging.rb
thallgren: The only way to find out is probably to run every conceivable test that we have. A PEZ build would be a good start.

joshcooper: According to https://docs.puppet.com/puppet/4.10/lang_reserved.html#reserved-class-names, `main` is a reserved class name, though attempts to use it produce a duplicate resource declaration error, which isn't what I was expecting. I assume it's because the implicit `:main` class resolves to `Class[main]{:name=>"main"}`:

```puppet
cat main.pp
class main {
  notify { 'hi': }
}
include main
```

```
$ bx puppet apply main.pp
...
Error: Duplicate declaration: Class[Main] is already declared; cannot redeclare
```

spec/unit/type/component_spec.rb
joshcooper: According to https://docs.puppet.com/puppet/4.10/lang_reserved.html#reserved-class-names, `main` is a reserved class name, though attempts to use it produce a duplicate resource declaration error, which isn't what I was expecting. I assume it's because the implicit `:main` class resolves to `Class[main]{:name=>"main"}`:

```puppet
cat main.pp
class main {
  notify { 'hi': }
}
include main
```

```
$ bx puppet apply main.pp
...
Error: Duplicate declaration: Class[Main] is already declared; cannot redeclare
```

thallgren: If `main` is a reserved class name, then the above test should be removed. Instead, we should trap all attempts to create a class with that name. The only time a Class[main] is created is when the [compiler executes `#evaluate_main`](/puppetlabs/puppet/blob/master/lib/puppet/parser/compiler.rb#L580).

lib/puppet/util/tagging.rb
joshcooper: If we can prevent a user-defined class main from being created in the first place, then we shouldn't need this kind of change, right?

thallgren: Not sure what you mean. This change just prevents the title 'main' from becoming a tag. There will still be one class with that title.

spec/unit/transaction/resource_harness_spec.rb
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/env ruby
 require 'spec_helper'
 
-require 'puppet/transaction/resource_harness'
```
joshcooper: We need to preserve this require, otherwise, there's no guarantee `Puppet::Transaction::ResourceHarness` is defined.

lib/puppet/util/tagging.rb
joshcooper: As @hlindberg said, "we don't know all the places that may matter." That is exactly my concern. I don't want to play a game of whack-a-mole, trying to figure out all of the places we need to check if it's the `main` class or not.

However, if we can prevent a user from ever creating a main class (like you've done in commit 9771456, then do we need to add any checks at all? For example, can we drop the `&& Puppet::Resource::MAIN != name` conditional above, since a user can't add a `main` class, so we don't have to worry about it being tagged?

thallgren: We can remove that test and the one and only 'main' class that indeed is created (by the compiler), will then get tagged with 'main'. That in turn causes a unit test failure.

If you think that tagging the main class with main is OK, then I'll remove this special case and fix that unit test.

spec/unit/transaction/resource_harness_spec.rb
thallgren: But I'm also adding a `require 'puppet/type/component'` on the next line which in turn `require 'puppet/transaction'`. If someone fiddles with that chain of requirements, then all tests in this file will break.

We usually try to avoid redundant `require` declarations. Is there any specific reason you don't want that done here?

joshcooper: I'm of the opinion that a class should require its immediate dependencies, even if that means a file is required multiple times. Ruby keeps track of which files have already been required (`LOADED_FEATURES`) so there's no harm in requiring multiple times. It will only be loaded once.

In cases where there is a cluster of related functionality, e.g. `lib/puppet/network`, we often just require the parent, with the knowledge that it requires all "related" files.

But in this case a `component` shouldn't know anything about `transaction`. If anything the dependency should be the other way around, as a transaction applies resources, which can include components.
 
The other reason is if you require `component` with the knowledge that it requires `transaction`, then you're making assumptions about transitive dependencies. If you ever make changes to `component` so it no longer requires `transaction`, then you'll break everything that depends on `component`. That seems very fragile.

lib/puppet/util/tagging.rb
joshcooper: > If you think that tagging the main class with main is OK

TBH I'm not sure. For example, currently the `main` tag is exclude from `Stage[main]` and `Class[main]`, but `Class[settings]` is tagged with `settings`. I have no idea if that's intentional or just happens to work that way:

```json
    {
      "type": "Stage",
      "title": "main",
      "tags": [
        "stage"
      ],
      "exported": false
    },
    {
      "type": "Class",
      "title": "Settings",
      "tags": [
        "class",
        "settings"
      ],
      "exported": false
    },
    {
      "type": "Class",
      "title": "Main",
      "tags": [
        "class"
      ],
      "exported": false,
      "parameters": {
        "name": "main"
      }
    }
```

thallgren: In that case, I will remove the special case and adjust the unit-test accordingly, in anticipation that the aforementioned thorough test needed to verify this PR will encounter unknown problems.

spec/unit/transaction/resource_harness_spec.rb
thallgren: Ok, I'll add the requirement, although I suspect that we have a vast number of places where we indeed do rely on transitive dependencies.

joshcooper:
@thallgren, @hlindberg and I talked about this during standup and decided to close as won't fix. We're concerned that we can't identify all of the places this might break, and while running PE tests is useful, it won't be conclusive. For now, I'm going to close this PR.

joshcooper:
Something else I realized afterwards, the current behavior depends on whether we're dealing with a resource (server) or RAL (agent) catalog:

Resource/Server Catalog

```
(byebug) result.resource('class', 'Main').ref
"Class[Main]"
(byebug) result.resource('Class', 'main').ref
"Class[Main]"
(byebug) result.resource('Class', 'Main').ref
"Class[Main]"
(byebug) result.resource('class', :main).ref
*** NoMethodError Exception: undefined method `ref' for nil:NilClass

nil
(byebug) result.resource('Class', :main).ref
*** NoMethodError Exception: undefined method `ref' for nil:NilClass

nil
```

But all of the following work for a RAL/Agent catalog:

```
(byebug) result.to_ral.resource('class', 'main').ref
"Class[Main]"
(byebug) result.to_ral.resource('class', 'Main').ref
"Class[Main]"
(byebug) result.to_ral.resource('Class', 'main').ref
"Class[Main]"
(byebug) result.to_ral.resource('Class', 'Main').ref
"Class[Main]"
(byebug) result.to_ral.resource('class', :main).ref
"Class[Main]"
(byebug) result.to_ral.resource('Class', :main).ref
"Class[Main]"
```

The part that I don't understand is that puppetdb tries to save a resource/server catalog using `:main`, but the following returns `nil` :


```
(byebug) result.resource('class', :main)
nil
```

So how did the pdb terminus ever work? /cc @mullr

mullr:
> So how did the pdb terminus ever work?

You mean in general? ;)

I think in the terminus we have `'Class', :main`; does that work?

mullr:
Oh, I see above that it does not.

thallgren:
Copying my comment regarding closing this from the jira ticket:

I'm not at all happy closing this for the reason that we don't know if it might break something, somewhere. We do know that there is a problem in that a deserialzed resource is different from the original. We also know for sure that referencing a resource using the string 'main' as the title causes a problem. To me, it seems very wrong to not fix known problems because we're afraid of the unknown ones. If we don't rely on our tests, then let's improve the tests. We cannot have a situation where we are paralyzed by fear that something might break somewhere down the line.

joshcooper:
> We also know for sure that referencing a resource using the string 'main' as the title causes a problem.

I'm not sure that's the case, see my comments in https://github.com/puppetlabs/puppet/pull/5929#issuecomment-306299922.

thallgren:
But doesn't that just reveal yet another inconsistency that would be fixed by this ticket?

PR #5930, opened 2017-05-31
(maint) Ship locales/ in Puppet tarballs

domcleal:
0a9ba3e (PUP-6958) added locales to the gem file list, but not `files`,
so it's missing from the published Puppet tarballs.


joshcooper:
Thanks @domcleal! We can ignore the appveyor issues as that's a regression in the appveyor environment.

MosesMendoza:
Merged to master at 8fd3340f56ac4e8550a2da692966290b84db9ede

PR #5932, opened 2017-05-31
(PUP-7133) Remove deprecated `always_cache_features` setting

Magisus:
The `always_cache_features` setting has been deprecated in favor of
`always_retry_plugins` for some time. This commit removes it.


Magisus:
About https://github.com/puppetlabs/puppet/blob/master/ext/rack/config.ru#L26 Is this file something that is still used? If so, do we need to replace that arg with something?

Magisus:
Switched the rack config to disable `always_retry_plugins`, since that seems to be the behavior we were achieving with the old setting.

PR #5934, opened 2017-06-01
(PUP-7594) Allow dotted key with integers to navigate integer hash keys

thallgren:
Prior to this commit, Hiera would consider an unquoted integer in a
dotted key to be an index in an array. This is now changed so that an
unquoted integer key can be used for navigating a hash that uses integer
keys.


thallgren:
Wrong ticket number in comment

PR #5937, opened 2017-06-02
Change dnf to be default for Fedora

mhjacks:
As per https://tickets.puppetlabs.com/browse/PUP-7613


blackknight36:
This looks like a duplicate of #5451.  Also, I don't think dnf should be used as the default for all Fedora releases.  There are legacy systems in use that still need to use the yum provider which this commit would break.

mhjacks:
Agreed @blackknight36.  Your fix is better than mine in any case (because it avoids the breakage problem, which I agree is a risk).

PR #5938, opened 2017-06-02
(PUP-3940) Send proper mime types in the Accept header

joshcooper:
Previously, puppet sent the format names in the Accept header, e.g. `Accept: pson`. This modifies puppet to send the proper mime type, e.g. `Accept: application/json, text/pson`. It is backwards compatible with all puppet ruby REST endpoints, because it historically has always accepted either the format name or mime type. It is backwards compatible with puppetserver REST endpoints for CA and static_file_content, because puppetserver ignores the `Accept` header and always returns content as `text/plain` or `application/octet-stream` respectively.

### DIFF
--- a/lib/puppet/indirector/rest.rb
+++ b/lib/puppet/indirector/rest.rb
@@ -104,8 +104,9 @@ class Puppet::Indirector::REST < Puppet::Indirector::Terminus
     network_formats = model.supported_formats.reject do |format|
       [:yaml, :b64_zlib_yaml].include?(format)
     end
+    mime_types = network_formats.map { |f| model.get_format(f).mime }
     common_headers = {
-      "Accept"                                     => network_formats.join(", "),
+      "Accept"                                     => mime_types.join(', '),
       Puppet::Network::HTTP::HEADER_PUPPET_VERSION => Puppet.version
     }
 
--- a/lib/puppet/indirector/rest.rb
+++ b/lib/puppet/indirector/rest.rb
@@ -10,6 +10,7 @@ class Puppet::Indirector::REST < Puppet::Indirector::Terminus
   include Puppet::Network::HTTP::Compression.module
 
   IndirectedRoutes = Puppet::Network::HTTP::API::IndirectedRoutes
+  EXCLUDED_FORMATS = [:yaml, :b64_zlib_yaml, :dot]
 
   class << self
     attr_reader :server_setting, :port_setting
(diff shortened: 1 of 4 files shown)


camlow325:
I noticed that with this code in place that one of the formats the agent says it will accept for the catalog is "text/dot".  Is this expected?  Might be fine - I just wasn't familiar with it.

joshcooper:
> `text/dot`

It's expected that the full mime type is in the `Accept` header, but it really shouldn't be in the list of network formats, since the agent has no way to deserialize a catalog in that format:

```
$ bx puppet agent -t --preferred_serialization_format dot
Info: Retrieving pluginfacts
Info: Retrieving plugin
Error: Could not run: Puppet::Resource::Catalog does not respond to from_dot; can not intern instances from text/dot
```

The issue is puppet conflates serialization formats with renderers, like `console`. I can add a commit to ensure we don't include `dot`, similar to how we exclude `yaml` and `b64_zlib_yaml`.

/cc @thallgren

joshcooper:
@camlow325 updated:

```
<- "POST /puppet/v3/catalog/agent?environment=production HTTP/1.1\r\nAccept: application/json, application/x-msgpack, text/pson\r\nX-Puppet-Version: 5.0.0\r\nAccept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3\r\nUser-Agent: Puppet/5.0.0 Ruby/2.3.1-p112 (x86_64-darwin15)\r\nHost: chara.corp.puppetlabs.net:8140\r\nContent-Length: 6605\r\nContent-Type: application/x-www-form-urlencoded\r\n\r\n"
```

lib/puppet/indirector/rest.rb
```diff
@@ -102,10 +102,11 @@ def self.port
   def headers
     # yaml is not allowed on the network
     network_formats = model.supported_formats.reject do |format|
-      [:yaml, :b64_zlib_yaml].include?(format)
+      [:yaml, :b64_zlib_yaml, :dot].include?(format)
```
Iristyle: Can this just simply be:

```ruby
excluded_formats = [:yaml, :b64_zlib_yaml, :dot]
network_formats = model.supported_formats - excluded_formats
```

```
[4] pry(main)> excluded_formats = [:yaml, :b64_zlib_yaml, :dot]
=> [:yaml, :b64_zlib_yaml, :dot]
[5] pry(main)> formats = [:foo, :bar, :dot, :baz, :qux, :yaml, :b64_zlib_yaml]
=> [:foo, :bar, :dot, :baz, :qux, :yaml, :b64_zlib_yaml]
[6] pry(main)> formats - excluded_formats
=> [:foo, :bar, :baz, :qux]
```

joshcooper: ah yeah, let me fix that

joshcooper:
@Iristyle updated

PR #5940, opened 2017-06-02
(PUP-6359) Fail fileserver auth for mounts with non-global allow

camlow325:
The use of security directives in fileserver.conf mount points has been
deprecated for multiple releases in favor of the use of the auth.conf file.
This commit is a step toward complete removal of support for these
directives.

With this commit, requests will be denied access to a mount point at the
fileserver indirector layer, with a corresponding error message logged, if any
non-global allow (allow *) directives are present in the mount point definition.

Previously, requests made to mount points for which no security directives were
defined would fail authentication at the fileserver indirector layer.  With
this commit, these requests would be granted authentication at the fileserver
indirector layer.

### DIFF
--- a/lib/puppet/indirector/file_server.rb
+++ b/lib/puppet/indirector/file_server.rb
@@ -15,7 +15,15 @@ class Puppet::Indirector::FileServer < Puppet::Indirector::Terminus
 
     # If we're not serving this mount, then access is denied.
     return false unless mount
-    mount.allowed?(request.node, request.ip)
+
+    # If there are no auth directives or there is an 'allow *' directive, then
+    # access is allowed.
+    if mount.empty? || mount.globalallow?
+      return true
+    end
+
+    Puppet.err _("Denying %{method} request for %{desc} on fileserver mount '%{mount_name}'. Use of auth directives for 'fileserver.conf' mount points is no longer supported. Remove these directives and use the 'auth.conf' file instead for access control.") % { method: request.method, desc: request.description, mount_name: mount.name }
+    return false
   end
 
   # Find our key using the fileserver.
--- a/spec/integration/indirector/file_content/file_server_spec.rb
+++ b/spec/integration/indirector/file_content/file_server_spec.rb
@@ -70,6 +70,7 @@ describe Puppet::Indirector::FileContent::FileServer, " when finding files" do
 
     # Use a real mount, so the integration is a bit deeper.
     mount1 = Puppet::FileServing::Configuration::Mount::File.new("one")
+    mount1.stubs(:globalallow?).returns true
     mount1.stubs(:allowed?).returns true
     mount1.path = File.join(path, "%h")
 
--- a/spec/integration/indirector/file_metadata/file_server_spec.rb
+++ b/spec/integration/indirector/file_metadata/file_server_spec.rb
@@ -50,6 +50,7 @@ describe Puppet::Indirector::FileMetadata::FileServer, " when finding files" do
 
         # Use a real mount, so the integration is a bit deeper.
         mount1 = Puppet::FileServing::Configuration::Mount::File.new("one")
+        mount1.stubs(:globalallow?).returns true
         mount1.stubs(:allowed?).returns true
         mount1.path = File.join(env_path, "%h")
 
--- a/spec/unit/indirector/file_server_spec.rb
+++ b/spec/unit/indirector/file_server_spec.rb
@@ -249,6 +249,7 @@ describe Puppet::Indirector::FileServer do
       @configuration.stubs(:split_path).with(@request).returns([@mount, "rel/path"])
       @request.stubs(:node).returns("mynode")
       @request.stubs(:ip).returns("myip")
+      @mount.stubs(:name).returns "myname"
       @mount.stubs(:allowed?).with("mynode", "myip").returns "something"
     end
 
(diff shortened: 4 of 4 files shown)


lib/puppet/indirector/file_server.rb
```diff
@@ -15,7 +15,15 @@ def authorized?(request)
...
-    mount.allowed?(request.node, request.ip)
+
+    # If there are no auth directives or there is an 'allow *' directive, then
+    # access is allowed.
+    if mount.empty? || mount.globalallow?
+      return true
+    end
+
+    Puppet.err _("Denying %{method} request for %{desc} on fileserver mount '%{mount_name}'. Use of auth directives for 'fileserver.conf' mount points is no longer supported. Remove these directives and use the 'auth.conf' file instead for access control.") % { method: request.method, desc: request.description, mount_name: mount.name }
+    return false
```
camlow325: I could have instead gone with raising an error instead of the arguably more passive approach of just denying access.  From an information disclosure perspective, though, I was thinking that it would be best for the content which goes out over the wire in this case to only show that the request is not authorized rather than having any more specific details about the removal of fileserver.conf security directive support.  I figured that it would be more appropriate to give the extra level of detail in the server side log, where there's a smaller likelihood of the info being exploited for nefarious means.

lib/puppet/indirector/file_server.rb
```diff
@@ -15,7 +15,15 @@ def authorized?(request)
 
     # If we're not serving this mount, then access is denied.
     return false unless mount
-    mount.allowed?(request.node, request.ip)
+
+    # If there are no auth directives or there is an 'allow *' directive, then
+    # access is allowed.
+    if mount.empty? || mount.globalallow?
```
camlow325: As I mentioned in a [comment](https://tickets.puppetlabs.com/browse/PUP-6359?focusedCommentId=449247&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-449247) on the related JIRA ticket, I was a little on the fence about whether it's best to grant access when there are no security directives present in the mount point definition - where `mount.empty?` would be true - because on upgrade it would be possible for authentication to be granted to a mount point for which it had previously been denied.

Ultimately, though, I think this is best assuming that we'd at some point like to eliminate the presence / handling of security directives from the mount point definitions completely - presumably at the same time we would gut support for Ruby auth.conf from the Puppet code.

lib/puppet/indirector/file_server.rb
adrienthebo: My experience is that every time we log more information in remote requests someone figures out a way to use that for evil. :+1: for logging information locally only and just denying access.

adrienthebo:
Code changes look sensible. I'll :+1: this after I've verified the changes locally.

camlow325:
@adrienthebo Cool, thanks.  Also, I should mention that I ran the full Puppet acceptance tests against two node beaker configuration - 1 having a master and agent role and the other having just an agent role.  The existing tests all passed for me - without any modifications - under this configuration.

lib/puppet/indirector/file_server.rb
MosesMendoza: @camlow325 I disagree here - IMO we should retain the previous behavior of denying if no allow/deny directives are present. I think the only way access should be granted is if they've done `allow *`. Admittedly I don't know what the use case would be for having such a mount defined, but maybe one exists?

camlow325: @MosesMendoza ok, but if we do that, I think it means we have to support / require the use of `allow *` indefinitely - for as long as we have `fileserver.conf` support.  I'd like to get @joshcooper's take on this.

adrienthebo: Defaulting to deny and requiring an `allow *` for each mountpoint seems like the most secure option, but I'm not a huge fan of it. We're explicitly removing authorization rules from `fileserver.conf`, and removing authorization will mean that at some point we're going to change to default allow.

Do we have long term plans to replace the fileserver with another implementation or service? If so I'm okay-ish with requiring `allow *` on each mountpoint as a stopgap until we replace this service. However if we don't have such plans I dislike leaving this technical debt lying as cruft.

Do we have a feel for how widely custom fileserver.conf mountpoints are used?

adrienthebo:
@beergeek @BillWeiss you both have feelings about security, and almost certainly a better understanding of how widespread fileserver.conf rules include auth. Care to opine?

BillWeiss:
Hm. It's been a while since I was in there, I don't know what's common. Could we package something in the installer that looks for fileserver.conf rules and logs something appropriately urgent seeming?

camlow325:
> Could we package something in the installer that looks for fileserver.conf rules and logs something appropriately urgent seeming?

I'm not sure that we really have an installer per-se in the OSS puppet-agent packaging.  I suppose a script could be launched post-install which might do something like this but I've understood that RE has frowned upon doing anything sophisticated like that.  There's probably a more elegant solution which could be done for PE.

lib/puppet/indirector/file_server.rb
camlow325: @ahpook weighed in the question about moving to default allow instead of default deny on the JIRA ticket [here](https://tickets.puppetlabs.com/browse/PUP-6359?focusedCommentId=450960&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-450960).  He's in agreement with the idea of moving to default allow since it enables users to avoid having to have any security directives in a fileserver mount point definition, which is ultimately what we'd like to see happen.  He also said he convinced @MosesMendoza in an offline conversation to go with this as well.  Given that, I'm going to leave the implementation as-is on the PR.

PR #5941, opened 2017-06-02
(PUP-7617) Only use Windows locale path on Windows

Magisus:
Because the config file for locale tooling is a relative path, in some
situations the Windows path is a valid path on Unix systems, even though
it does not point to the right location. This causes Puppet to look for
the locales config file in the wrong place and fail. This commit adds a
check, so we only try to use the Windows path on Windows, even if it is
valid elsewhere.

### DIFF
--- a/lib/puppet.rb
+++ b/lib/puppet.rb
@@ -37,9 +37,9 @@ begin
 
   if File.exist?(local_locale_path)
     locale_path = local_locale_path
-  elsif File.exist?(win32_system_locale_path)
+  elsif Puppet::Util::Platform.windows? && File.exist?(win32_system_locale_path)
     locale_path = win32_system_locale_path
-  elsif File.exist?(posix_system_locale_path)
+  elsif !Puppet::Util::Platform.windows? && File.exist?(posix_system_locale_path)
     locale_path = posix_system_locale_path
   else
     # We couldn't load our locale data.


lib/puppet.rb
```diff
@@ -37,7 +37,7 @@
 
   if File.exist?(local_locale_path)
     locale_path = local_locale_path
-  elsif File.exist?(win32_system_locale_path)
+  elsif Puppet::Util::Platform.windows? && File.exist?(win32_system_locale_path)
     locale_path = win32_system_locale_path
   elsif File.exist?(posix_system_locale_path)
```
Iristyle: Should this similarly have a posix feature check?

joshcooper: maybe `!Puppet::Util::Platform.windows?`?

Magisus: Are those ever both false?

Magisus: Yeah I guess that makes sense.

Magisus:
Updated to include posix check

joshcooper:
@abottchen could you give this a try and verify it resolves the issue?

abottchen:
@joshcooper Yep, that worked for me on 2017.2.1.  Thanks to all involved for the quick turnaround!

